PR #1128, opened 2023-10-23
Hash getEmptyExtents() in DynamicTransformConcretizationInfo

jacobhinkle:
This would only improve latency when a Fusion has a lot of invocations with different patterns of emptiness. I would expect that to be very rare.


jacobhinkle:
!build

naoyam (approved):

PR #1135, opened 2023-10-24
Adding a CUPTI based fusion profiler to report stable GPU Kernel times

kevinstephano:
_Note: I intend to make a separate PR that includes a python specific profiling API that consumes a profile object 
 and prints the profiling (because it can be made prettier in python) rather than relying on C++'s stdout and providing the ability to query the profile object from python._

One intended consumer is pytest benchmarking in a manner like (_not implemented yet!_):
```python
out = fd.execute(inputs, profile_to_obj=True)
prof = fd.profile()
# or
out = fd.validate(inputs, ref_outputs, profile_to_obj=True)
prof = fd.profile()
```

# What is being profiled?

A Cuda Device Synchronize is executed prior to each measured fusion in order to expose the Host Time to measurement.  The first line of each profile includes the aggregate fusion information and 1st segment's information.  Successive segments have their aggregate fusion information omitted.

* Aggregate Fusion information
  * `Fus#` _(nocupti)_ - Fusion Number
  * `NSegs` _(nocupti)_ - Number of Segments
  * `CuEvtTime(ms)` _(nocupti)_ - The Cuda Event Time that wraps the entire fusion process including kernel execution.
  * `CmpTm(ms)` _(nocupti)_ - Is the Parallel Compilation time or the aggregate of serialized segment compilation depending on which option is used. Compilation time includes Kernel Lowering and NVRTC compilation.
  * `KerTime(ms)` - The aggregated Kernel Time of all the segments executed.
  * `EffBw(GB/S)` - The Effective Bandwidth of the inputs/outputs to the fusions which does not include the I/O specific to each segment.  Therefore, segmentation should penalize the aggregate Effective Bandwidth.
  * `%PeakBw` - The Percentage of the Peak Bandwidth of the Fusion with respect to the previous Effective Bandwidth. 
 This is also the `%SOL` of the Fusion.
  * `In(MB)` _( verbose)_ - The size of the fusion inputs in MBs.
  * `Out(MB)` _( verbose)_ - The size of the fusion outputs in MBs.
* Segment Information 
  * `Seg#` = The segment index found in the line
  * `S-KerName` - The Cuda Kernel's short name
  * `S-KerTm(ms)` - The CUPTI profiled Kernel Time for the given segment
  * `S-CmpTm(ms)` _( verbose)_ - The Compile Time for the given segment if kernels are compiled without Parallel Compilation.
  * `S-EffBw(GB/s)` - The effective bandwidth of the segment given the specific I/O of this segment's kernel.
  * `S-%PeakBw` - The percent of the Peak Bandwidth of this kernel.  The kernel's `%SOL` assuming the kernel is bandwidth limited.
  * `S-In(MB)` - The size of the kernel inputs in MBs.
  * `S-Out(MB)` - The size of the kernel outputs in MBs.
  * `S-Smem[Dyn,Stat]` -  The Dynamic and Static Shared Memory used by the kernel in bytes
  * `S-Regs` - The registers-per-thread used by the kernel
  * `S-Grid` - The grid dimensions of the kernel
  * `S-Block` - The block dimensions of the kernel
  * `S-Cluster` _( verbose)_ - The cluster dimensions of the kernel
  * `S-Dev` _( verbose)_ - The device id of the kernel execution.
  * `S-Stm` _(verbose)_ - The stream id of the kernel execution.
  * `S-DeviceName` _( verbose)_ - The market name of the device used for kernel execution.
  * `S-PeakBW(GB/s)` _( verbose)_ - The Peak Bandwidth of the device used for execution.
  
# ENV Variable `NVFUSER_PROF` Values to Enable Profiling
All environment variables minimally enable the Fusion Profiler that populates the `FusionProfile` object.

`NVFUSER_PROF` values:
* `enable` : Enables full profiling and records the activity in the `FusionProfile` object only.
* `enable.nocupti` : Only enables host timer profiling of fusion level activity and records the activity in the `FusionProfile` object only.  The "nocupti" object allows the user to see the host timing without CUPTI overhead that can be as significant as `10us` to `30us`.
* `print` : Enables full profiling and records the activity in the `FusionProfile` object and prints it to the console.
* `print.verbose` : Enables full profiling and records the activity in the `FusionProfile` object and prints it to the console.  Additional information is printed in extra columns that may cause line wrapping on smaller terminal windows.  The _(verbose)_ level indicates the set of additional columns printed, above.
* `print.nocupti` : Only enables host timer profiling of fusion level activity and records the activity in the `FusionProfile` object and print it to the console with an abbreviated set of columns for the limited information.  The _(nocupti)_ level indicates the set of columns printed, above.

# Profiling Output as an Object

You can collect statistics from profiling a specific fusion by querying the output object:
```cpp
FusionProfile fprof = FusionProfiler::get()->profile();
```

## FusionProfile Object

```cpp
struct FusionProfile {
  int64_t fusion_id{-1};
  int64_t segments{0};

  double cuda_evt_time_ms{0.0};
  double host_time_ms{0.0};
  double compile_time_ms{0.0};
  double kernel_time_ms{0.0};

  int64_t input_bytes{0};
  int64_t output_bytes{0};

  double effective_bandwidth_gbs{0.0};
  double percentage_peak_bandwidth{0.0};

  std::vector<KernelProfile> kernel_profiles{};
};
```
## KernelProfile Object
```cpp
struct KernelProfile {
  std::string name{};
  int64_t segment_id;
  int device{-1};
  uint32_t stream{0};
  uint32_t correlation_id{0};

  double compile_time_ms{0.0};
  double time_ms{0.0};
  double effective_bandwidth_gbs{0.0};
  double percentage_peak_bandwidth{0.0};

  std::array<int32_t, 3> grid{0, 0, 0};
  std::array<int32_t, 3> block{0, 0, 0};
  std::array<uint32_t, 3> cluster{0, 0, 0};

  int32_t dynamic_shared_mem{0};
  int32_t static_shared_mem{0};
  uint32_t registers{0};

  int64_t input_bytes{0};
  int64_t output_bytes{0};

  std::string device_name{};
  double peak_bandwidth_gbs{0.0};
};
```

# Profiling Output in C++ stdout

## Default - `NVFUSER_PROF=print`

```bash
Fus#  NSegs CuEvtTm(ms) HstTm(ms) CmpTm(ms) KerTm(ms) EffBw(GB/s) %PeakBw   Seg# S-KerName  S-KerTm(ms) S-EffBw(GB/s) S-%PeakBw S-In(MB)  S-Out(MB) S-Smem[Dyn,Stat] S-Regs S-Grid           S-Block
    0     1     328.582   328.529   277.940     0.059     1765.88     91.24    0    kernel1       0.059       1765.88     91.24    52.442    52.494        [512, 16]     41     [8192, 1, 1]      [128, 1, 1]
    1     1      95.395    95.345    93.025     0.057     1845.05     95.33    0    kernel2       0.057       1845.05     95.33    52.429    52.429           [0, 0]     16    [25600, 1, 1]      [128, 1, 1]
    2     1    1076.190  1076.054  1012.063     0.141     1583.82     81.84    0    kernel3       0.141       1583.82     81.84   104.870   118.030        [512, 16]     52     [8192, 1, 1]      [128, 1, 1]
    3     1     139.589   139.350   129.783     0.246     1707.12     88.21    0    kernel4       0.246       1707.12     88.21   209.715   209.715           [0, 0]     16   [102400, 1, 1]      [128, 1, 1]
    4     1     283.830   283.744   276.638     0.092     1855.98     95.90    0    kernel5       0.092       1855.98     95.90   104.858    65.536           [0, 4]     34    [25600, 1, 1]      [128, 1, 1]
    5     1     124.316   124.253   119.113     0.069     1716.20     88.68    0    kernel6       0.069       1716.20     88.68    65.536    52.429           [0, 0]     18    [25600, 1, 1]      [128, 1, 1]
    6     2     369.042   368.662   226.779     0.431     1582.54     81.77    0    kernel7       0.048       1085.88     56.11    52.429     0.006       [2048, 16]     32      [50, 17, 1]      [32, 16, 1]
    -     -           -         -         -         -           -         -    1    kernel8       0.382       1645.26     85.01   419.430   209.715           [0, 0]     22   [102400, 1, 1]      [128, 1, 1]
    7     2    2024.829  2024.633   978.544     0.338     1436.02     74.20    0    kernel9       0.132       1585.10     81.90   209.715     0.026       [2048, 16]     32      [50, 30, 1]       [64, 8, 1]
    -     -           -         -         -         -           -         -    1   kernel10       0.205       1340.02     69.24   170.466   104.870        [512, 16]    124      [1, 432, 1]       [2, 64, 1]
    8     2     197.137   197.089   193.144     0.101     1559.95     80.60    0   kernel12       0.049       1070.98     55.34    52.429     0.006       [2048, 16]     32      [50, 17, 1]      [32, 16, 1]
    -     -           -         -         -         -           -         -    1   kernel11       0.052       2021.47    104.45    52.429    52.429           [0, 0]     16    [25600, 1, 1]      [128, 1, 1]
    9     2    1729.789  1729.616   872.094     0.284     1294.23     66.87    0   kernel13       0.104       1506.07     77.82   157.286     0.019       [2048, 16]     32      [38, 31, 1]       [64, 8, 1]
    -     -           -         -         -         -           -         -    1   kernel14       0.179       1170.76     60.49   157.358    52.442        [512, 16]    126      [1, 432, 1]       [2, 64, 1]
   10     1       0.149     0.099     0.000     0.058     1809.74     93.51    0    kernel1       0.058       1809.74     93.51    52.442    52.494        [512, 16]     41     [8192, 1, 1]      [128, 1, 1]
```
## Verbose  - `NVFUSER_PROF=print.verbose`
This option gives added details in the far right columns that includes things like the device type profiled and the peak bandwidth of the device.
```base
Fus#  NSegs CuEvtTm(ms) HstTm(ms) CmpTm(ms) KerTm(ms) EffBw(GB/s) %PeakBw   Seg# S-KerName  S-KerTm(ms) S-CmpTm(ms) S-EffBw(GB/s) S-%PeakBw S-In(MB)  S-Out(MB) S-Smem[Dyn,Stat] S-Regs S-Grid           S-Block          S-Cluster        S-Dev S-Stm S-DeviceName         S-PeakBw(GB/s)
    0     1     328.880   328.828   277.962     0.059     1777.37     91.84    0    kernel1       0.059     276.450       1777.37     91.84    52.442    52.494        [512, 16]     41     [8192, 1, 1]      [128, 1, 1]        [0, 0, 0]     0     7   NVIDIA A100 80GB P        1935.36
    1     1      95.295    95.241    92.902     0.059     1791.61     92.57    0    kernel2       0.059      92.573       1791.61     92.57    52.429    52.429           [0, 0]     16    [25600, 1, 1]      [128, 1, 1]        [0, 0, 0]     0     7   NVIDIA A100 80GB P        1935.36
    2     1    1078.698  1078.567  1014.494     0.139     1599.47     82.64    0    kernel3       0.139    1013.166       1599.47     82.64   104.870   118.030        [512, 16]     52     [8192, 1, 1]      [128, 1, 1]        [0, 0, 0]     0     7   NVIDIA A100 80GB P        1935.36
    3     1     139.003   138.766   129.382     0.244     1719.43     88.84    0    kernel4       0.244     128.690       1719.43     88.84   209.715   209.715           [0, 0]     16   [102400, 1, 1]      [128, 1, 1]        [0, 0, 0]     0     7   NVIDIA A100 80GB P        1935.36
    4     1     283.905   283.822   276.647     0.092     1855.35     95.87    0    kernel5       0.092     276.213       1855.35     95.87   104.858    65.536           [0, 4]     34    [25600, 1, 1]      [128, 1, 1]        [0, 0, 0]     0     7   NVIDIA A100 80GB P        1935.36
    5     1     124.911   124.845   119.094     0.069     1706.67     88.18    0    kernel6       0.069     118.597       1706.67     88.18    65.536    52.429           [0, 0]     18    [25600, 1, 1]      [128, 1, 1]        [0, 0, 0]     0     7   NVIDIA A100 80GB P        1935.36
    6     2     368.721   368.344   226.877     0.428     1593.32     82.33    0    kernel7       0.048       0.000       1085.91     56.11    52.429     0.006       [2048, 16]     32      [50, 17, 1]      [32, 16, 1]        [0, 0, 0]     0     7   NVIDIA A100 80GB P        1935.36
    -     -           -         -         -         -           -         -    1    kernel8       0.379       0.000       1657.88     85.66   419.430   209.715           [0, 0]     22   [102400, 1, 1]      [128, 1, 1]        [0, 0, 0]     0     7   NVIDIA A100 80GB P        1935.36
    7     2    2031.692  2031.500   975.809     0.336     1441.90     74.50    0    kernel9       0.133       0.000       1571.43     81.20   209.715     0.026       [2048, 16]     32      [50, 30, 1]       [64, 8, 1]        [0, 0, 0]     0     7   NVIDIA A100 80GB P        1935.36
    -     -           -         -         -         -           -         -    1   kernel10       0.203       0.000       1356.72     70.10   170.466   104.870        [512, 16]    124      [1, 432, 1]       [2, 64, 1]        [0, 0, 0]     0     7   NVIDIA A100 80GB P        1935.36
    8     2     196.050   196.000   192.079     0.103     1533.67     79.24    0   kernel12       0.049       0.000       1073.08     55.45    52.429     0.006       [2048, 16]     32      [50, 17, 1]      [32, 16, 1]        [0, 0, 0]     0     7   NVIDIA A100 80GB P        1935.36
    -     -           -         -         -         -           -         -    1   kernel11       0.054       0.000       1952.80    100.90    52.429    52.429           [0, 0]     16    [25600, 1, 1]      [128, 1, 1]        [0, 0, 0]     0     7   NVIDIA A100 80GB P        1935.36
    9     2    1722.642  1722.485   865.384     0.266     1381.35     71.37    0   kernel13       0.103       0.000       1530.45     79.08   157.286     0.019       [2048, 16]     32      [38, 31, 1]       [64, 8, 1]        [0, 0, 0]     0     7   NVIDIA A100 80GB P        1935.36
    -     -           -         -         -         -           -         -    1   kernel14       0.163       0.000       1287.31     66.52   157.358    52.442        [512, 16]    126      [1, 432, 1]       [2, 64, 1]        [0, 0, 0]     0     7   NVIDIA A100 80GB P        1935.36
```

## No-Cupti - `NVFUSER_PROF=print.nocupti`
The "no-cupti" option omits kernel profiling and, therefore, you do not see any segment information.  The purpose of this option is to show you the Host Time `HstTm(ms)` without the overhead of CUPTI.  It empirically appears to be between `10us` and `30us` in the Host Time when the Compile Time `CmpTm(ms)` does not dominate the Host Time.
```bash
Fus#  NSegs CuEvtTm(ms) HstTm(ms) CmpTm(ms)
    0     1     328.874   328.818   278.639
    1     1      94.973    94.920    92.629
    2     1    1076.356  1076.217  1012.515
    3     1     139.532   139.288   129.602
    4     1     283.904   283.817   277.174
    5     1     124.174   124.110   119.053
    6     2     370.288   369.907   226.604
    7     2    2031.641  2031.442   977.103
    8     2     196.486   196.436   192.158
    9     2    1726.821  1726.661   870.991
```
# How CUPTI is used to capture the Kernel Time

## Enabled Activity Records through the [CUPTI Activity API](https://docs.nvidia.com/cupti/r_main.html#r_activity)
3 types of activity were recorded:
* `CONCURRENT_KERNEL` - Captures the kernel profile associated with a CUPTI defined `correlation-id`.
* `EXTERNAL_CORRELATION` - Records `correlation-id` associated with a pushed Id to associate segment kernels with specific `correlation-id`s.
* `DRIVER` - These records are never captured but are required to be enabled in order to see the `EXTERNAL_CORRELATION` records
## Associating Correlation-Ids with Kernels of Segments
There is a push and pop API used to associate correlation-ids of records with a given range of code.  For nvFuser, the range was the code associated with launching a kernel for a segment.  The push and pop methods were: `cuptiActivityPushExternalCorrelationId` and `cuptiActivityPopExternalCorrelationId`.
## Register Callbacks to Asynchronously capture Record Activity

Two functions need to be record: one to associate a buffer to capture a record `cupti_buffer_requested()` and the second reads the buffer and writes the data from the buffer into a data structure for consumption `cupti_buffer_completed()`.  The callback API is: `cuptiActivityRegisterCallbacks()`.

## Avoiding CUPTI Overhead in Host Time measurement

Despite flushing the recorded activity only after the Fusion Profiler `stop` method is called, CUPTI can generate a lot of host time overhead.  The documentation suggests the CUPTI activity is asynchronously handled on a separate thread but I did not verify that is true.  One thing that made a big difference was the Activity Buffer size.  The default size shown in the sample code is `8MB`.  When the buffer size was reduced to `4KB`,  there was a decrease in the overhead amount by about `150us`.  The remaining overhead appeared to be on the order of `10us` to `30us`.


CMakeLists.txt
kevinstephano, line 1: The changes to this file include "finding" the library to link CUPTI and adding `fusion_profiler.h/cpp` and a test file.

csrc/executor.cpp
```diff
@@ -1673,6 +1673,14 @@ std::vector<at::Tensor> FusionExecutor::runFusion(
...
   }
 
+  const bool measure_kernel_time = measure_kernel_time_ ||
+      isDebugDumpEnabled(DebugDumpOption::EffectiveBandwidth) ||
+      isDebugDumpEnabled(DebugDumpOption::PerfDebugVerbose);
+
+  if (measure_kernel_time) {
+    inputBytesProcessed(args);
+  }
+
```
kevinstephano, line 1683: I needed to move up where the input bytes were measured because the `args` variable captures all of a kernel's args in nvFuser that includes outputs if I wait to look at the `args` farther down.  Therefore, our effective bandwidth measurement was previously double counting outputs.

kevinstephano:
!build

csrc/cuda_utils.h
```diff
@@ -47,3 +48,20 @@
...
+
+#define NVFUSER_CUPTI_SAFE_CALL(x)                              \
+  do {                                                          \
+    CUptiResult _status = x;                                    \
+    if (_status != CUPTI_SUCCESS) {                             \
+      const char* pErrorString;                                 \
+      cuptiGetResultString(_status, &pErrorString);             \
+      fprintf(                                                  \
+          stderr,                                               \
+          "%s:%d: Error: Function %s failed with error: %s.\n", \
```
wujingyue, line 60:
```suggestion
          "%s:%d: Error: %s failed with error: %s.\n", \
```

csrc/cuda_utils.h
```diff
@@ -47,3 +48,20 @@
         " failed with error ",        \
         cudaGetErrorString(_result)); \
   } while (0)
+
+#define NVFUSER_CUPTI_SAFE_CALL(x)                              \
+  do {                                                          \
+    CUptiResult _status = x;                                    \
+    if (_status != CUPTI_SUCCESS) {                             \
+      const char* pErrorString;                                 \
```
wujingyue, line 56:
```suggestion
      const char* errorString;                                 \
```

csrc/executor.h
```diff
@@ -179,26 +179,89 @@ class FusionExecutor : public NonCopyable {
     return measure_kernel_time_ ? kernel_time_ms_ : 0;
   }
 
+  //! Returns the input bytes accessed for a kernel
+  //! \note It is important to sample the args struct prior to adding the
+  // 1    output to the args struct
+  int64_t inputBytesProcessed(KernelArgumentHolder& args) {
```
wujingyue, line 185: Can this be a `const KernelArgumentHolder&`?

csrc/executor.h
```diff
@@ -179,26 +179,89 @@ class FusionExecutor : public NonCopyable {
...
+    int64_t total_bytes = 0;
+    if (!bytes_processed_per_input_.has_value()) {
+      int64_t num_bytes = 0;
+      bytes_processed_per_input_ = std::optional<std::vector<int64_t>>(
```
wujingyue, line 189: I think it's fine to say `bytes_processed_per_input_ = std::vector<int64_t>(args.size(), 0);`, which is less verbose.

csrc/executor.h
```diff
@@ -179,26 +179,89 @@ class FusionExecutor : public NonCopyable {
...
+          std::vector<int64_t>(args.size(), 0));
+      // Figure how many bytes are inputs, outputs, and temporary buffers
+      for (auto i : c10::irange(args.size())) {
+        if (args[i]->is<at::Tensor>()) {
+          auto t = args[i]->as<at::Tensor>();
+          num_bytes = t.numel() *
```
wujingyue, line 195: I'm not sure this is the right measure for non-packed tensors. How about `t.storage().nbytes()`?

csrc/executor.h
```diff
@@ -179,26 +179,89 @@ class FusionExecutor : public NonCopyable {
...
+    } else {
+      for (auto bp : bytes_processed_per_input_.value()) {
+        total_bytes += bp;
+      }
+    }
+    return total_bytes;
+  }
+
+  //! Returns the output bytes accessed for a kernel
+  int64_t outputBytesProcessed(std::vector<at::Tensor>& outputs) {
```
wujingyue, line 210: I'd move non-trivial implementation to the `.cpp` file.

csrc/executor.h
```diff
@@ -179,26 +179,89 @@ class FusionExecutor : public NonCopyable {
...
+        if (args[i]->is<at::Tensor>()) {
+          auto t = args[i]->as<at::Tensor>();
+          num_bytes = t.numel() *
+              (int64_t)dataTypeSize(aten_to_data_type(t.scalar_type()));
+          bytes_processed_per_input_.value().at(i) = num_bytes;
+          total_bytes += num_bytes;
+        }
+      }
+    } else {
+      for (auto bp : bytes_processed_per_input_.value()) {
```
wujingyue, line 202: This doesn't look right. Are we going to cache `bytes_processed_per_input_` across `FusionExecutor::runFusion` calls? What if tensor sizes changed?

csrc/fusion_profiler.h
```diff
@@ -0,0 +1,319 @@
...
+  ProfilerState state_;
+};
+
+//! \struct DeviceDescriptor
+//! \brief This struct captures the GPU information necessary to calculate the
+//! the Peak Bandwidth of the specific GPU queried.
+struct DeviceDescriptor {
+  //! Queries the GPU to populate the struct's data members and calculates the
+  //! peak bandwidth
+  void generate(int device);
```
wujingyue, line 82: `generate` doesn't need to be an instance method. I'd say

```
static DeviceDescriptor generate(int device);
```

csrc/fusion_profiler.h
```diff
@@ -0,0 +1,319 @@
...
+#define _FP_ENABLE(code)                                            \
+  if (isDebugDumpEnabled(DebugDumpOption::FusionProfiler) ||        \
+      isDebugDumpEnabled(DebugDumpOption::FusionProfilerNocupti) || \
+      isDebugDumpEnabled(DebugDumpOption::FusionProfilerVerbose) || \
+      isOptionEnabled(EnableOption::FusionProfiler) ||              \
+      isOptionEnabled(EnableOption::FusionProfilerNocupti)) {       \
+    code;                                                           \
+  }
+
+//! Fusion Profiling Macros
```
wujingyue, line 273: Is there a good reason to not use functions instead?

wujingyue (changes requested):

csrc/executor.h
jjsjann123: second this. numel isn't right for expanded tensors e.t.c. storage is also deprecated btw. so `t.untyped_storage().nbytes()`

jacobhinkle: Thanks guys. I wrote those lines originally, but I didn't know how to get the actual storage.

csrc/executor.h
jacobhinkle: We only store it if the debug dump option is enabled, in which case we retrieve it afterward. So it's always referring to the last call that had the dump option enabled. Note that storing this requires a little bit of computation, so we only want to compute it when we're debugging.

wujingyue: The part I didn't get was line 187. It appears to me that we only compute `bytes_processed_per_input_` when it's empty; otherwise the previous computation is reused. Isn't that a problem or am I missing anything?

csrc/fusion_profiler.h
jjsjann123: Agree with Jingyue here.

It doesn't feel like the macro is saving us much code.

csrc/fusion_profiler.cpp
```diff
@@ -0,0 +1,737 @@
...
+// CUPTI buffer size 4.0 KB
+#define BUF_SIZE (1 * 4 * 1024)
+// 8-byte alignment for the buffers
+#define ALIGN_SIZE (8)
+#define ALIGN_BUFFER(buffer, align)                                 \
+  (((uintptr_t)(buffer) & ((align)-1))                              \
+       ? ((buffer) + (align) - ((uintptr_t)(buffer) & ((align)-1))) \
+       : (buffer))
+
+const char* get_demangled_name(const char* pName) {
```
jjsjann123, line 26: we also have `demangle` in `csrc/exceptions.cpp`

Should we just put them together. Also __cxa_demangle may fail. So either assert or just return the original there.

jjsjann123: https://github.com/NVIDIA/Fuser/blob/ae428309dc41c987c9ed7673e8e832a5828ad1a4/csrc/exceptions.cpp#L51

csrc/fusion_profiler.cpp
```diff
@@ -0,0 +1,737 @@
...
+    void* pUserData) {
+  CUpti_Activity* pRecord = nullptr;
+  CUptiResult status = CUPTI_SUCCESS;
+
+  do {
+    status = cuptiActivityGetNextRecord(pBuffer, validBytes, &pRecord);
+    if (status == CUPTI_SUCCESS) {
+      record_cupti_activity(pRecord, stdout);
+    } else if (status == CUPTI_ERROR_MAX_LIMIT_REACHED) {
+      break;
```
jjsjann123, line 93: Dumb question: What happens when max limit is reached? should we throw a warning here?

csrc/fusion_profiler.cpp
```diff
@@ -0,0 +1,737 @@
...
+void cupti_buffer_requested(
+    uint8_t** ppBuffer,
+    size_t* pSize,
+    size_t* pMaxNumRecords) {
+  uint8_t* pBuffer = FusionProfiler::get()->cuptiBufferPtr();
+  NVF_ERROR(pBuffer, "CUPTI Malloced buffer Pointer is null!");
+
+  *pSize = (size_t)BUF_SIZE;
+  *ppBuffer = ALIGN_BUFFER(pBuffer, ALIGN_SIZE);
+  *pMaxNumRecords = 0;
```
jjsjann123, line 109: another dumb question: why are we clearing max number records?

csrc/fusion_profiler.cpp
```diff
@@ -0,0 +1,737 @@
...
+  parallel_compile_timer_.reset();
+  segments_.clear();
+  kernel_profiles_.clear();
+  corrid_2_segid_.clear();
+}
+
+FusionProfiler* FusionProfiler::get() {
+  std::lock_guard<std::mutex> guard(singleton_lock_);
+  if (singleton_ == nullptr) {
+    singleton_ = new FusionProfiler();
```
jjsjann123, line 538: nitpick: maybe std::unique_ptr instead.

wujingyue: A static std::unique_ptr? There's a good reason not to do that: https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables

wujingyue: In the same spirit, I think we should move singleton_lock_ to this function as a local variable so the time of construction is predictable and thread-safe.

csrc/fusion_profiler.cpp
```diff
@@ -0,0 +1,737 @@
...
+  }
+  return os;
+}
+
+std::mutex FusionProfiler::singleton_lock_;
+FusionProfiler* FusionProfiler::singleton_ = nullptr;
+
+FusionProfiler::FusionProfiler()
+    : cupti_disabled_(false),
+      cupti_buffer_((size_t)4 * 1024),
```
Priya2698, line 503: Replace by `BUF_SIZE`?

csrc/fusion_profiler.cpp
```diff
@@ -0,0 +1,737 @@
...
+    }
+    case CUPTI_ACTIVITY_KIND_EXTERNAL_CORRELATION: {
+      CUpti_ActivityExternalCorrelation* pExternalCorrelationRecord =
+          (CUpti_ActivityExternalCorrelation*)pRecord;
+      FusionProfiler::get()->recordAsyncCorrIdActivity(
+          pExternalCorrelationRecord->externalId,
+          pExternalCorrelationRecord->correlationId);
+      break;
+    }
+    case CUPTI_ACTIVITY_KIND_DRIVER:
```
Priya2698, line 72: Is this case unhandled or not expected to occur? Should we raise some warning here?

csrc/executor.h
jacobhinkle: You're right that is a problem.

kevinstephano: I purposely tried to minimally modify the code that was already there instead of refactoring it.  A lot of this code can likely be removed once we have an alternative to the instrumentation in place.

csrc/fusion_profiler.cpp
kevinstephano: It walks the record in order to reach the limit.  It always reaches this point.

csrc/fusion_profiler.h
kevinstephano: Changed to functions.

csrc/fusion_profiler.cpp
kevinstephano: I added a NOTE to say that the DRIVER records are recorded in the Activity buffer but we just skip them since they don't have any relevant information to use.  They happen to exist because they are required to be in order to get EXTERNAL_CORRELATION records to appear.

csrc/fusion_profiler.cpp
kevinstephano: I added some more security to this loop so it cannot go into an infinite loop and I commented the `status == CUPTI_ERROR_MAX_LIMIT_REACHED` case.

csrc/fusion_profiler.cpp
kevinstephano: This is funny C overuse of a variable to communicate an additional piece of information. The `MaxNumRecords` value of 0 is meant to say there is not limit.  Consume all records!

csrc/fusion_profiler.cpp
kevinstephano: I made the suggested change.

csrc/fusion_profiler.cpp
kevinstephano: I decided to use a `static constexpr` for a data member under `FusionProfiler` to handle the `BUF_SIZE` consistently.

csrc/fusion_profiler.cpp
kevinstephano: I moved to use `demange()` from `csrc/exceptions.cpp`.  Therefore, I had to move the code out of the anonymous `namespace`.

csrc/cuda_utils.h
```diff
@@ -47,3 +48,20 @@
         " failed with error ",        \
         cudaGetErrorString(_result)); \
   } while (0)
+
+#define NVFUSER_CUPTI_SAFE_CALL(x)                              \
```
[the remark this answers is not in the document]
kevinstephano: Good point!  I'll move this.

csrc/fusion_profiler.h
kevinstephano: I made this change.

csrc/cuda_utils.h
kevinstephano: I made this change.

csrc/cuda_utils.h
kevinstephano: Made this change.

csrc/cuda_utils.h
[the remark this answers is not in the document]
kevinstephano: I moved the `NVFUSER_CUPTI_SAFE_CALL` to the `fusion_profiler.cpp` file to isolate the CUPTI usage.

csrc/executor.h
kevinstephano: `at::Tensor` does not appear to have the method `untyped_storage()` in the C++ API.  It does exist in the Python API.

csrc/executor.h
kevinstephano: Made the change.

csrc/executor.h
kevinstephano: Made the change.

csrc/executor.h
kevinstephano: Made the change.

csrc/executor.h
kevinstephano: The vector of tensor sizes should go away.  It does not make sense to maintain.  I just don't want to add refactoring away some infrastructure that depends on this vector to this PR.

CMakeLists.txt
```diff
@@ -331,6 +334,7 @@ if(BUILD_PYTHON)
     set_target_properties(${NVFUSER} PROPERTIES SUFFIX ".pyd")
   endif()
 
+  target_link_libraries(${NVFUSER} PRIVATE ${LIBCUPTI})
```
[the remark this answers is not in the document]
kevinstephano: Removed!

CMakeLists.txt
```diff
@@ -216,6 +218,7 @@ target_link_libraries(${NVFUSER_CODEGEN} PRIVATE flatbuffers)
 # For kernel_db, linking STL Filesystem Library for backward compatability with C++14
 target_link_libraries(${NVFUSER_CODEGEN} PRIVATE stdc++fs)
 target_link_libraries(${NVFUSER_CODEGEN} PRIVATE dynamic_type)
+target_link_libraries(${NVFUSER_CODEGEN} PRIVATE ${CUDA_NVRTC_LIB} ${LIBCUPTI})
```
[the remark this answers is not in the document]
kevinstephano: Changed!

CMakeLists.txt
```diff
@@ -75,6 +76,7 @@ list(APPEND NVFUSER_SRCS
   ${NVFUSER_SRCS_DIR}/evaluator_common.cpp
   ${NVFUSER_SRCS_DIR}/executor_utils.cpp
   ${NVFUSER_SRCS_DIR}/fusion.cpp
+  ${NVFUSER_SRCS_DIR}/fusion_profiler.cpp
```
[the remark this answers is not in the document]
kevinstephano: I added the conditional append for not MSVC as well as for the test.

kevinstephano:
!build

csrc/fusion_profiler.h
```diff
@@ -0,0 +1,308 @@
...
+//! A Fusion Profiler enabled check "decorator"
+template <class T>
+auto fusion_profiler_enabled(T&& func) {
+  // we create a closure below
+  auto new_function = [func = std::forward<T>(func)](auto&&... args) {
+    if (isDebugDumpEnabled(DebugDumpOption::FusionProfiler) ||
+        isDebugDumpEnabled(DebugDumpOption::FusionProfilerNocupti) ||
+        isDebugDumpEnabled(DebugDumpOption::FusionProfilerVerbose) ||
+        isOptionEnabled(EnableOption::FusionProfiler) ||
+        isOptionEnabled(EnableOption::FusionProfilerNocupti)) {
```
wujingyue, line 279: An easier and cleaner way is to let `FusionProfiler::get()` return `nullptr` when FusionProfiler is not enabled and then check just for that. This way, we only need to decide once whether FusionProfiler is enabled from EnableOptions.

csrc/fusion_profiler.h
```diff
@@ -0,0 +1,308 @@
...
+  void reset();
+  ProfilerState state() const;
+  SegmentProfiler& segment(size_t idx);
+
+  //! Profiling Methods
+  void start(bool cupti_disable = false);
+  void stop();
+  void createSegments(size_t num);
+  void startParallelCompile();
+  void stopParallelCompile();
```
wujingyue, line 218: How about `setParallelCompileWallTime`? This will let `FusionKernelRuntime` do the measure and make sure the start and the stop are called in pairs.

csrc/kernel_cache.cpp
```diff
@@ -1130,9 +1148,11 @@ void FusionKernelRuntime::compileFusionParallel(KernelArgumentHolder args) {
 
     if (num_groups == 1 || isOptionDisabled(DisableOption::ParallelCompile)) {
       FUSER_PERF_SCOPE("FusionKernelRuntime::compileFusionParallel");
+      segment_profiler_start_compile(group_id, args.getDeviceIndex());
```
wujingyue, line 1151: Call this inside `compileKernel`? Similar to my previous comments, I think it's better to let compileKernel measure the time and directly put it in the SegmentProfiler. Up to you!

wujingyue (changes requested):

csrc/kernel_cache.cpp
kevinstephano: If I start and stop inside `compileKernel`, then it will record sequential compile times in the parallel case.  The segment compile time is only recorded when the parallel compile is disabled.

csrc/fusion_profiler.h
kevinstephano: Currently, the timer is found in the Profiler singleton, therefore, recording the start and stop time at the same time as a pair would mean changing the location of the timer.  I would rather leave two functions in place as it does not seem guaranteed that the code will remain the same.

csrc/exceptions.cpp
```diff
@@ -144,6 +117,33 @@ std::optional<FrameInformation> parse_frame_information(
...
 
+std::string demangle(const char* name) {
+  int status = -1;
+
+  // This function will demangle the mangled function name into a more human
+  // readable format, e.g. _Z1gv -> g().
+  // More information:
+  // https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/libsupc%2B%2B/cxxabi.h
+  // NOTE: `__cxa_demangle` returns a malloc'd string that we have to free
+  // ourselves.
```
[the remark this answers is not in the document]
kevinstephano: Comment is moved!

csrc/executor.cpp
```diff
@@ -1845,6 +1836,52 @@ std::vector<at::Tensor> FusionExecutor::runFusion(
   return outputs;
 }
 
+int64_t FusionExecutor::inputBytesProcessed(const KernelArgumentHolder& args) {
+  int64_t total_bytes = 0;
+  if (!bytes_processed_per_input_.has_value()) {
+    int64_t num_bytes = 0;
+    bytes_processed_per_input_ =
+        std::vector<int64_t>(std::vector<int64_t>(args.size(), 0));
```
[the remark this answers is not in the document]
kevinstephano: Yikes, I did it in 2 places.  Thanks for the catch!

csrc/fusion_profiler.h
```diff
@@ -0,0 +1,308 @@
...
+//! includes aggregated times from the kernels generated by segments
+//! encapsulated by a Fusion.
+struct FusionProfile {
+  //! A static array to capture header strings for tables that print
+  //! the profiled information
+  static std::array<const char*, 25> column_strs;
+
+  void reset();
+
+  bool verbose{isDebugDumpEnabled(DebugDumpOption::FusionProfilerVerbose)};
```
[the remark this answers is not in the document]
kevinstephano: This is used by the `operator<<()` function for the `FusionProfile` object to determine when to display all of the possible columns of the profile output.  The verbose version might not fit on most people's screens without wrapping which is why the default is a smaller set.  The verbose output includes items that do not currently change from run to run in nvFuser at this time like the `cluster` configuration of a kernel and the Device Name.

csrc/fusion_profiler.h
wujingyue: That's fair. Nit: do you see a particular reason to call it `startParallelCompile` vs simply `startCompile`? The underlying implementation simply starts a wall timer, which is useful for non-parallel compilation too.

csrc/kernel_cache.cpp
wujingyue: > The segment compile time is only recorded when the parallel compile is disabled.

Got it. If you haven't, I'd like to see that written down somewhere (e.g. in stdout) so users are not confused why they are not seeing the compilation time for each segment. 

That being said, I think, even with parallel compilation, there's value to measure each segment's compile time. For example, to identify long tails that make parallelization less useful than expected. Up to you!

csrc/fusion_profiler.h
kevinstephano: I removed the prevous form of enablement. The new enablement is via:
```
isProfilerEnabled()
```

csrc/options.h
```diff
@@ -43,6 +43,9 @@ enum class DebugDumpOption {
   KernelArgs, //!< Print the runtime kernel arguments when launching kernels
   EffectiveBandwidth, //! Measure kernel performance and print effective
                       //! bandwidth
+  FusionProfiler, //! Measures fusion, compile, and kernel performance
+  FusionProfilerNocupti, //! Measures fusion, compile, and kernel performance
+  FusionProfilerVerbose, //! Measures fusion, compile, and kernel performance
```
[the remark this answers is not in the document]
kevinstephano: I entirely re-worked the options by adding a new environment variable: `NVFUSER_PROF`.  The exact values and descriptions are documented in the PR description.

csrc/fusion_profiler.h
kevinstephano: I think you are right.  I will just change the Fusion level compile time to reflect either the serial or parallel compilation time via a `startCompile` option.

csrc/kernel_cache.cpp
kevinstephano: I think you are correct with this assessment as well.  I will still record the segment time so we can determine the long-pole which was the original intention in adding the column.

kevinstephano:
!build

kevinstephano:
!build

csrc/fusion_profiler.h
```diff
@@ -0,0 +1,266 @@
...
+
+  //! Profiling Methods
+  void start(bool cupti_disable = false);
+  void stop();
+  void createSegments(size_t num);
+  void startCompile();
+  void stopCompile();
+  void inputBytesAccessed(int64_t bytes);
+  void outputBytesAccessed(int64_t bytes);
+  const FusionProfile& profile() const;
```
jacobhinkle, line 221: Could these all be made `static` and just call a private `get()` under the hood? Then we'll never be tempted to hold onto a `FusionProfiler*` since we won't be able to construct one.

csrc/fusion_profiler.cpp
```diff
@@ -0,0 +1,770 @@
...
+    "S-Out(MB)",
+    "S-Smem[Dyn,Stat]",
+    "S-Regs",
+    "S-Grid",
+    "S-Block",
+    "S-Cluster",
+    "S-Dev",
+    "S-Stm",
+    "S-DeviceName",
+    "S-PeakBw(GB/s)"};
```
jacobhinkle, line 421: Minor nitpick: `S-DeviceName` has spaces in it, which can make awking past it somewhat *awk*ward. Placing this field last might make it slightly easier to grab the `S-PeakBW` field.

csrc/fusion_profiler.cpp
```diff
@@ -0,0 +1,770 @@
...
+
+std::ostream& operator<<(std::ostream& os, const FusionProfile& fp) {
+  if (fp.fusion_id == 0) {
```
jacobhinkle, line 424: Optional CSV output would be fantastic here. For that we would probably want to break up the tuples into separate fields.

csrc/fusion_profiler.cpp
```diff
@@ -0,0 +1,770 @@
...
+std::array<const char*, 25> column_strs{
+    "Fus#",
+    "NSegs",
+    "CuEvtTm(ms)",
+    "HstTm(ms)",
+    "CmpTm(ms)",
+    "KerTm(ms)",
+    "EffBw(GB/s)",
+    "%PeakBw",
+    "Seg#",
```
jacobhinkle, line 405: Keeping with the theme,
```suggestion
    "S-Seg#",
```

csrc/fusion_profiler.cpp
kevinstephano: I can certainly change the default for `S-DeviceName`, however, I am getting back what is returned from the cuda query.  I can certainly change the order.  My hope is that you are not querying the stdout output as I am hoping I am giving you other ways to get the output.

csrc/fusion_profiler.h
kevinstephano: I can do that.  You are right there is no real reason to hold onto the `FusionProfiler*`.

csrc/fusion_profiler.cpp
jacobhinkle: Yeah true. Wrapping the underlying data is better.

jjsjann123 (approved):
minor comments. Thanks for cleaning up the headers.and build stuff.

csrc/options.h
```diff
@@ -114,6 +114,27 @@ enum class DisableOption {
...
+                 //! profiling inorder to measure true host time without
+                 //! overhead.
+  Print, //! Enables the profiler and prints the output to the console.
+  PrintNocupti, //! Enables the profiler, disables CUPTI specific
+                //! profiling inorder to measure true host time without
+                //! overhead, and prints the output to the console.
+  PrintVerbose, //! Enables the profiler and prints a complete set of columns
+                //! to the console.  WARNING: The output is will wrap on small
+                //! screens!
+  EndOfOption //! Placeholder for counting the number of elements
```
wujingyue, line 135: Is it used at all? I didn't manage to find where.

csrc/options.h
```diff
@@ -114,6 +114,27 @@ enum class DisableOption {
...
+  Enable, //! Enables the profiler.
+  EnableNocupti, //! Enables the profiler, but disables CUPTI specific
+                 //! profiling inorder to measure true host time without
+                 //! overhead.
+  Print, //! Enables the profiler and prints the output to the console.
+  PrintNocupti, //! Enables the profiler, disables CUPTI specific
+                //! profiling inorder to measure true host time without
+                //! overhead, and prints the output to the console.
+  PrintVerbose, //! Enables the profiler and prints a complete set of columns
+                //! to the console.  WARNING: The output is will wrap on small
```
wujingyue, line 133: I think in theory we are missing PrintNoCuptiVerbose? It's unfortunate we have to list all combinations. Is it possible to decouple enablement and printing so we have fewer knobs?

wujingyue (changes requested):

csrc/options.h
kevinstephano: `print.nocupti.verbose` does not make any sense as `.verbose` and `.nocupti` are distinct.  `.nocupti` prints a very limited subset of statistic only at the Fusion level that does not include any kernel specific timing.  `.verbose` is used to print out a larger set of of columns of statistics.  The goal with `.verbose` is to remove stats that don't change that and help to fit the set of columns on smaller screens.  You can see examples of how the 3 differ in 1st comment/description of the PR.

You can see the `.verbose` knob sets a variable in the `FusionProfile`: https://github.com/NVIDIA/Fuser/blob/baaacdf36e87416ca8692b0e9fb72192e08c6e89/csrc/fusion_profiler.h#L133

This is for use with the `operator<<()` for `FusionProfile`. 

https://github.com/NVIDIA/Fuser/blob/baaacdf36e87416ca8692b0e9fb72192e08c6e89/csrc/fusion_profiler.cpp#L529

wujingyue: I understood that -- that's why I said "in theory" we are missing PrintNoCuptiVerbose. My point was `Enable*` and `Print*` are orthogonal and therefore map more clearly to two separate flags. If we do that, `PrintNoCupti` doesn't even need to exist because it can be represented as `EnableNoCupti`+`Print`. 

Anyhow, the practical benefit of removing `PrintNoCupti` may be too small to justify further changes to this PR.

csrc/fusion_profiler.cpp
```diff
@@ -0,0 +1,787 @@
...
+}
+
+FusionProfiler* FusionProfiler::get() {
+  static std::mutex singleton_lock;
+  std::lock_guard<std::mutex> guard(singleton_lock);
+  if (singleton_ == nullptr) {
+    singleton_ = new FusionProfiler();
+  }
+  return singleton_;
+}
```
wujingyue, line 570: It appears that `singleton_` is only used in this function. I'd just define it here, getting rid of a non-local variable. 

```suggestion
  static std::mutex singleton_lock;
  static FusionProfiler* singleton = nullptr;
  
  std::lock_guard<std::mutex> guard(singleton_lock);
  if (singleton == nullptr) {
    singleton = new FusionProfiler();
  }
  return singleton;
}
```

csrc/kernel_cache.cpp
```diff
@@ -525,6 +534,15 @@ std::vector<at::Tensor> FusionExecutorCache::runFusionWithInputs(
     }
   }
   outputs.resize(new_size);
+
+  // NOTE: This should be the last code in the method to capture all host time
+  if (isProfilerEnabled()) {
+    FusionProfiler::stop();
+  }
+  if (isProfilerPrintingEnabled()) {
+    debug() << FusionProfiler::profile();
```
wujingyue, line 543: Do we need to print `std::endl` at the end?

wujingyue (approved):

csrc/kernel_cache.cpp
kevinstephano: Every row is terminated with a `std::endl` in the `operator<<()` function.
https://github.com/NVIDIA/Fuser/blob/baaacdf36e87416ca8692b0e9fb72192e08c6e89/csrc/fusion_profiler.cpp#L535

csrc/fusion_profiler.cpp
kevinstephano: Change made!

kevinstephano:
!build

PR #1138, opened 2023-10-24
[dynamic_type] Add binary size plot

zasdfgbnm:


naoyam (approved):

mmigdal-nv:
@zasdfgbnm seems that compile time and memory usage are O(n**2), I'm surprised that you don't see it here. I'm curious as to why. Thanks for the graph!

PR #1141, opened 2023-10-25
pushTensorProxy except for trivial forwarding.

wujingyue:
This surprisingly passes all unit tests. Is it crazy? From the code comment, it seems fine to just return proxy tensors. Do you remember why alias outputs were treated differently?

### DIFF
--- a/csrc/executor.cpp
+++ b/csrc/executor.cpp
@@ -1292,7 +1292,6 @@ KernelArgumentHolder FusionExecutor::inferOutputSizes(
   expr_eval.precomputedValues() = evaluator_precomputed_values.get();
 
   auto arg_index_type = args.getSmallestIndexTypeOfArguments();
-  const auto& output_to_input_aliases = fusion->getOutputToInputAliasIndices();
 
   KernelArgumentHolder ret;
   ret.setDeviceIndex(args.getDeviceIndex());
--- a/csrc/executor.cpp
+++ b/csrc/executor.cpp
@@ -1315,34 +1315,12 @@ KernelArgumentHolder FusionExecutor::inferOutputSizes(
           fusion->outputs()[out_i]->isA<TensorView>(),
           "Cannot allocate outputs that are not tensors.");
       auto output_tv = fusion->outputs()[out_i]->as<TensorView>();
-
-      auto alias_it = std::find_if(
-          output_to_input_aliases.begin(),
-          output_to_input_aliases.end(),
-          [&](const auto& output_to_input) {
-            return output_to_input.first == (int)out_i;
-          });
-
-      if (alias_it != output_to_input_aliases.end()) {
-        // When aliasing output to an input, we do not need to allocate a new
-        // output but still need to push an entry.
-        ret.push(PolymorphicValue(0L));
-      } else {
-        const auto& [sizes, strides] = inferShapeOfOutput(output_tv, expr_eval);
-        const auto dtype = (output_tv->dtype() == DataType::Index)
-            ? data_type_to_aten(arg_index_type)
-            : data_type_to_aten(output_tv->dtype());
-        ret.pushTensorProxy(sizes, strides, dtype);
-      }
+      const auto& [sizes, strides] = inferShapeOfOutput(output_tv, expr_eval);
+      const auto dtype = (output_tv->dtype() == DataType::Index)
+          ? data_type_to_aten(arg_index_type)
+          : data_type_to_aten(output_tv->dtype());
+      ret.pushTensorProxy(sizes, strides, dtype);
     }
-  }
-
-  for (const auto& [aliased_output_index, aliased_input_index] :
-       output_to_input_aliases) {
-    NVF_ERROR(
-        args[aliased_input_index]->is<at::Tensor>(),
-        "alias io only supports tensor");
-    *ret[aliased_output_index] = *args[aliased_input_index];
   }
   return ret;
 }
--- a/test/test_gpu3.cpp
+++ b/test/test_gpu3.cpp
@@ -5248,11 +5248,13 @@ TEST_F(NVFuserTest, FusionTrivialInputForwarding_CUDA) {
   FusionExecutorCache fec(std::move(fusion_ptr));
   auto cg_outputs = fec.runFusionWithInputs({t0, t1});
 
+  EXPECT_EQ(cg_outputs[0].data_ptr(), t0.data_ptr());
   testValidate(fusion, cg_outputs, {t0, t1}, {t0}, __LINE__, __FILE__);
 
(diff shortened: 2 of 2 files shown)


wujingyue:
!build

wujingyue:
!build

rdspring1:
From the comment, it seems you can pass TensorProxy.

@jjsjann123 Do you remember why we have special logic for alias outputs here?

csrc/executor.cpp
```diff
@@ -1315,35 +1314,13 @@ KernelArgumentHolder FusionExecutor::inferOutputSizes(
...
-        const auto dtype = (output_tv->dtype() == DataType::Index)
-            ? data_type_to_aten(arg_index_type)
-            : data_type_to_aten(output_tv->dtype());
-        ret.pushTensorProxy(sizes, strides, dtype);
-      }
+      const auto& [sizes, strides] = inferShapeOfOutput(output_tv, expr_eval);
+      const auto dtype = (output_tv->dtype() == DataType::Index)
+          ? data_type_to_aten(arg_index_type)
+          : data_type_to_aten(output_tv->dtype());
+      ret.pushTensorProxy(sizes, strides, dtype);
```
jjsjann123, line 1321: does this mean we are allocating memory for forwarded outputs then? We need to double check if the integration code is still blindly allocate buffer for every output, regardless of its alias status.

The old trick here is that we put a dummy output there so allocation wouldn't create any buffer there.

jjsjann123: I agree this is tricky to read and I'm onboard with us removing the code here.

Since we are at this topic, we can probably refactor the output allocation code and properly use aliased buffer there as well.

jjsjann123: This is where we skip this place holder in the existing code. https://github.com/NVIDIA/Fuser/blob/main/csrc/kernel_cache.cpp#L198C1-L198C1

csrc/executor.cpp
```diff
@@ -1315,35 +1314,13 @@ KernelArgumentHolder FusionExecutor::inferOutputSizes(
...
-        // output but still need to push an entry.
-        ret.push(PolymorphicValue(0L));
-      } else {
-        const auto& [sizes, strides] = inferShapeOfOutput(output_tv, expr_eval);
-        const auto dtype = (output_tv->dtype() == DataType::Index)
-            ? data_type_to_aten(arg_index_type)
-            : data_type_to_aten(output_tv->dtype());
-        ret.pushTensorProxy(sizes, strides, dtype);
-      }
+      const auto& [sizes, strides] = inferShapeOfOutput(output_tv, expr_eval);
```
jjsjann123, line 1317: The logic here looks fine.

Looking at allocOutputs part here, everything seems fine as well: https://github.com/NVIDIA/Fuser/blob/main/csrc/executor.cpp#L916

We are also skipping it here:
https://github.com/NVIDIA/Fuser/blob/main/csrc/kernel_cache.cpp#L198C1-L206

Since we are not only pushing TensorProxy here, maybe it's perfectly fine to push trivial forward as-is.

Can we double check that part with some simple examples?

jjsjann123 (approved):
Since we are now using a TensorProxy already, maybe it's fine to move forward with this change.

Can we double check that we are not allocating memory with trivial forwarding?

csrc/executor.cpp
wujingyue: > Can we double check that part with some simple examples?

Double check a trivially forwarded input stays an alias? Sure, I added more checks in `FusionTrivialInputForwarding*_CUDA`.

jjsjann123: I think more importantly, double checking that we are NOT allocating any memory for the trivial forwarded input and throwing it away later...
Having said that, I don't know what would be a reliable way to check that automatically though.

wujingyue: Phew, it took me a while to confirm this, but I managed to verify the number of calls to `cudaMalloc` equals the number of inputs. 

```
me @ swagtron : dev | /opt/pytorch/nvfuser (alias*)
$ PYTORCH_NO_CUDA_MEMORY_CACHING=1 nsys nvprof bin/nvfuser_tests --gtest_filter=*TrivialInputForwarding2_CUDA 2> /dev/null | grep cudaMalloc
      1.2            94359          1    94359.0    94359.0     94359     94359          0.0  cudaMalloc
me @ swagtron : dev | /opt/pytorch/nvfuser (alias*)
$ PYTORCH_NO_CUDA_MEMORY_CACHING=1 nsys nvprof bin/nvfuser_tests --gtest_filter=*TrivialInputForwarding_CUDA 2> /dev/null | grep cudaMalloc
      1.2           100611          2    50305.5    50305.5      6102     94509      62513.2  cudaMalloc
```

Two caveats though: 
1. I commented out `testValidate` which allocates more tensors for computing atol and rtol. 
2. Surprisingly, `at::empty` triggers a `cudaMalloc`! So I patched the following. This patch may be the right thing to do anyway. Trivial input forwarding is a special case of the aliasing I'm trying to support.

```
diff --git a/csrc/executor.cpp b/csrc/executor.cpp
index e1ad311..f9ebe93 100644
--- a/csrc/executor.cpp
+++ b/csrc/executor.cpp
@@ -943,6 +943,15 @@ std::vector<at::Tensor> allocOutputs(
           "alias io only supports tensor");
       outputs.emplace_back(*inputs[aliased_input_index]);
     } else if (kernel->outputs().at(output_idx)->isFusionInput()) {
+#if 1
+      const int input_idx = std::find(
+                                kernel->inputs().begin(),
+                                kernel->inputs().end(),
+                                kernel->outputs().at(output_idx)) -
+          kernel->inputs().begin();
+      NVF_ERROR(input_idx < (int)kernel->inputs().size());
+      outputs.emplace_back(*inputs[input_idx]);
+#else
       // pushing empty tensor for trivial forwarding. Since we handle this in
       // integration, see step 1 - note [trivial forwarding]
       auto alloc_dom =
@@ -954,6 +963,7 @@ std::vector<at::Tensor> allocOutputs(
           at::TensorOptions().dtype(at::kFloat).device(device);
       outputs.emplace_back(
           at::empty(std::vector<int64_t>(alloc_dom.size(), 0), tensor_options));
+#endif
     } else {
       auto alloc_tensor = at::native::empty_strided_cuda(
           buf_info.sizes,
```

jjsjann123: Thanks for double checking this~~

That's indeed a surprise, I thought `at::empty` with a zero sized-tensor should have got a short-cut to not trigger any allocation call there. :laughing: 

Also agree that trivial forwarding is just a special case of aliasing. Lots of clean up and code merge to be done :+1:

PR #1143, opened 2023-10-25
Make sure stdout capture is stopped at the end of each test

### LINKED ISSUE #1142: Stdout capture fails to be cleaned up
Gtest `CaptureStdout` needs to be reset at each test, however, that's not currently guaranteed. If an exception occurs during captured, the test would just exit without stopping the capture, which would result like:

```
[ FATAL ] /home/nmaruyama/nvfuser/debug3/third_party/googletest/googletest/src/gtest-port.cc:1200:: Only one stdout capturer can exist at a time.
```

naoyam:
Fixes #1142 

To make it easy to understand why this fix is necessary, I temporarily added two new tests just to expose the issue.

```
 ./bin/nvfuser_tests --gtest_filter="*CaptureTest*"
```

This should run `CaptureTest1` and then `CaptureTest2`. The former fails before stopping the capture, so with the current TOT, the issue of #1142 should happen. This PR makes sure the capture is stopped no matter if the test itself succeeds.

### DIFF
--- a/test/test_gpu3.cpp
+++ b/test/test_gpu3.cpp
@@ -8412,7 +8412,7 @@ TEST_F(NVFuserTest, FusionTestWarnRegisterSpill_CUDA) {
         __FILE__,
         "");
   }
-  std::string output = stopCaptureStdout();
+  std::string output = getCapturedStdout();
   NVF_CHECK(
       output.find("Register spill detected") != std::string::npos,
       "Register spill is not captured!");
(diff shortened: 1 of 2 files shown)


test/test_gpu3.cpp
```diff
@@ -9692,6 +9692,20 @@ TEST_F(NVFuserTest, PredicateRNGOps) {
...
+  NVF_ERROR(false);
+  stopCaptureStdout();
+}
+
+// Dummy tests to reproduce the issue. Will be deleted
+TEST_F(NVFuserTest, CaptureTest2) {
+  captureStdout();
+  stopCaptureStdout();
+}
+
```
naoyam, line 9708: Will be deleted before merging

jacobhinkle (approved):
Looks good to me. If `stopCaptureStdout` is called before `captureStdout` we get an empty string which might be confusing if we forget to call `captureStdout` since empty is a valid result. That is probably not a major concern though.

naoyam:
Slightly changed the interface. Trying to get a captured text without starting capturing should fail.

PR #1147, opened 2023-10-25
Implement `LoadStoreOp::evaluate` for `Set.Permute`.

wujingyue:


test/test_evaluator.cpp
```diff
@@ -535,4 +538,24 @@ TEST_F(ExprEvalTest, TernaryOps) {
   }
 }
 
+TEST_F(ExprEvalTest, Permute) {
+  Fusion fusion;
+  FusionGuard fg(&fusion);
+
+  TensorView* in =
+      TensorViewBuilder().shape({-1, 6}).dtype(DataType::Float).build();
```
zasdfgbnm, line 546: Could you test a higher dimensional tensor? For example, a tensor whose shape is `(1, 2, 3, 4, 5, 6, 7, 8, 9, 10)`, by using higher dimensional tensor, it will be more likely to find potential bugs in the implementation.

csrc/ir/nodes.cpp
```diff
@@ -2361,9 +2361,46 @@ LoadStoreOp::LoadStoreOp(
...
+std::vector<int64_t> Permutation(
+    const std::vector<T>& in,
+    const std::vector<T>& out) {
+  std::vector<int64_t> permutation;
+  permutation.reserve(out.size());
+  // O(n^2) is totally fine for the current use case of computing the
+  // root-to-rfactor permutation. If needed, this can be improved by requiring T
+  // to be hashable and/or comparable.
+  for (const T& out_element : out) {
+    permutation.push_back(
```
zasdfgbnm, line 2378: I think what you implemented is
```C++
out[i] = in[perm[i]]
```
which does not match with what you have documented
```C++
out[perm[i]]==in[i]
```

csrc/ir/nodes.cpp
```diff
@@ -2361,9 +2361,46 @@ LoadStoreOp::LoadStoreOp(
...
+        std::find(in.begin(), in.end(), out_element) - in.begin());
```
zasdfgbnm, line 2379: nit: `std::distance`

csrc/ir/nodes.cpp
```diff
@@ -2361,9 +2361,46 @@ LoadStoreOp::LoadStoreOp(
   addDataAttribute(cache_op);
 }
 
+namespace {
+// Returns the permutation from `in` to `out`, i.e., `out[perm[i]]==in[i]`.  As
+// a precondition, `out` must be a permutation of `in` per the definition of
+// std::is_permutation.
+template <typename T>
+std::vector<int64_t> Permutation(
```
zasdfgbnm, line 2369: nit: I think function names should start with lower case letters?

test/test_evaluator.cpp
wujingyue: Done. It's testing 4D now.

csrc/ir/nodes.cpp
wujingyue: Good catch!

csrc/ir/nodes.cpp
wujingyue: Done.

csrc/ir/nodes.cpp
wujingyue: Done.

test/test_evaluator.cpp
```diff
@@ -535,4 +538,24 @@ TEST_F(ExprEvalTest, TernaryOps) {
...
+  fusion.addOutput(out);
+
+  at::Tensor in_tensor =
+      at::rand({256}).cuda().as_strided({2, 3, 4, 6}, {128, 32, 8, 1});
+
+  ExpressionEvaluator evaluator;
+  evaluator.bind(in, in_tensor);
+  at::Tensor out_tensor = evaluator.evaluate(out).as<at::Tensor>();
+  EXPECT_THAT(out_tensor.sizes(), ElementsAre(2, 6, 3, 4));
+  EXPECT_THAT(out_tensor.strides(), ElementsAre(128, 1, 32, 8));
```
zasdfgbnm, line 558: Should we do:
```C++
std::vector<int64_t> new2old = {0, 3, 1, 2};
....
TensorView* out = permute(in, new2old);
...
auto expect = in_tensor.permute(new2old);
EXPECT_EQ(out_tensor.sizes(), expect.sizes());
EXPECT_EQ(out_tensor. strides(), expect. strides());
```

zasdfgbnm (approved):

test/test_evaluator.cpp
wujingyue: For the purpose of the mentioned issue, I prefer checking the aliasing effect of `Set.Permute` explicitly. In other words, if (I know it's rare) `at::permute` stopped producing an alias, I want the test to fail so I'm forced to change the alias analysis and/or the expression evaluator.

csrc/ir/nodes.cpp
```diff
@@ -2361,9 +2361,46 @@ LoadStoreOp::LoadStoreOp(
...
+        in.begin(), std::find(in.begin(), in.end(), out_element)));
+  }
+  return permutation;
+}
+} // namespace
+
 std::vector<PolymorphicValue> LoadStoreOp::evaluate(
     const ExpressionEvaluator& ee,
     const std::vector<PolymorphicValue>& inputs) const {
+  if (TensorView* out_tv = dynamic_cast<TensorView*>(out())) {
```
Priya2698, line 2388: How many `op_types` are there for `LoadStoreOp`. Should we split the evaluate method (for instance: a switch-case implementation) based on `op_type`?

wujingyue: I remember several. According to https://github.com/NVIDIA/Fuser/blob/a55c2df1d45f0fb5030b068ab7755d0527163e6b/csrc/ir/utils.cpp#L576, only those with an rfactor domain are used for permutation and others can simply be implemented as input forwarding. So I think it's better to not split based on `op_type`. Wdyt?

Priya2698: Got it, this is fine then.

Priya2698 (approved):

PR #1149, opened 2023-10-26
Finalize loop promotion map

### LINKED ISSUE #1146: IdGraph: Not all loop groups get promotion domains
If no inline promotion happens, no loop promotion is set, which means some loop groups have no promotion domain. That is a little inconvenient since the loop promotion will be used as the concrete ID in the current CompuAtMap graph.

naoyam:
Fixes #1146


naoyam:
Note that loop groups appearing before forwarded domains may still not have unique promotions. I'd argue that those groups themselves do not make much sense, however. I'll ignore them for now.

PR #1152, opened 2023-10-26
Validation method for python benchmarks

### LINKED ISSUE #1117: Variable validation thresholds in python benchmarks.
Benchmarks currently use a fixed threshold for verifying correctness of fusion definitions against the torch implementation. This threshold can be too high/low depending on the input size and datatype. 
The validation tests particularly fail for `bfloat16` with the current tolerances.

Priya2698:
- Moves tolerance computations out of `test`
- Binds a `getValTolerances` method to `FusionDefinition` to return validation thresholds based on reduction sizes.
- Adds a `validate(inputs, reference outputs)` method to Fusion definition: Validates nvfuser output against reference outputs using variable tolerances.

Note: Some sizes have been removed due to validation errors (For 4D inputs, `spatial sizes = 2` and for 2D inputs, `sizes = 2/4/8 and sizes > 1048576`). The larger sizes will be added after we update the tolerance values for those sizes. For smaller sizes, we plan on skipping validation but including them in benchmarking after all the benchmarks are ported. 

Closes #1117


kevinstephano:
I would just add a `validate()` method to the  python `FusionDefinition` so anyone can use the functionality out of just benchmarking.  I think it is reasonable to just bind the tolerances  from C++ and then make a `validate()` method in `nvfuser/__init__.py`. 

I would make it easy:
```
FusionDefinition.validate(inputs, ref_outputs, kwargs)
```

jacobhinkle:
> ```python
> FusionDefinition.validate(inputs, ref_outputs, kwargs)
> ```

Could we make it even less conspicuous?
```python
fd.execute(inputs, validate_against=ref_outputs)
```
then raise a well-defined exception if the validation fails.

csrc/validator_utils.h
```diff
@@ -0,0 +1,349 @@
...
+  // Tolerances generated from randn + add + sum fusion
+  // compared against double precision
+  std::array<std::array<double, 2>, 20> sum_tolerances_float = {
+      {{4, 1.68222e-06},      {8, 2.23704e-06},      {16, 2.95788e-06},
+       {32, 4.4778e-06},      {64, 6.75395e-06},     {128, 8.57934e-06},
+       {256, 1.30594e-05},    {512, 2.19122e-05},    {1024, 3.3451e-05},
+       {2048, 5.78476e-05},   {4096, 0.000108292},   {8192, 0.00012207},
+       {16384, 0.000136882},  {32768, 0.000248561},  {65536, 0.000407594},
+       {131072, 0.000500901}, {262144, 0.000923019}, {524288, 0.00156909},
+       {1048576, 0.00223107}, {2097152, 0.00343043}}};
```
jacobhinkle, line 35: I know this PR didn't introduce this table, but I expected to see a square root formula, and it seems like these do roughly follow a sqrt trend. Modelling rounding error as iid finite-variance noise, the variance of the sum should linearly scale with the number of items being reduced, so for a single sum operation we should use `sqrt(n)` times the epsilon for a given type. If we are doing reduction other than sum it could be more or less: e.g. a product of zero-mean variables should have variance raised to the power `n`, so should use `pow(epsilon, n/2)` for the tolerance.

python_benchmarks/global_params.py
```diff
@@ -41,8 +41,7 @@ def generate_input_sizes(dims: Union[int, List] = 2) -> List[Tuple]:
 
 
 # Datatypes to benchmark
-# TODO: Add torch.bfloat16 after adding support for variable thresholds
-FLOAT_DTYPES = [torch.float16, torch.float32]
+FLOAT_DTYPES = [torch.bfloat16, torch.float16, torch.float32]
```
jacobhinkle, line 44: Should we only add bfloat16 if we detect Ampere and above?

Priya2698:
> > ```python
> > FusionDefinition.validate(inputs, ref_outputs, kwargs)
> > ```
> 
> Could we make it even less conspicuous?
> 
> ```python
> fd.execute(inputs, validate_against=ref_outputs)
> ```
> 
> then raise a well-defined exception if the validation fails.

Are you suggesting combining the `execute` method with validation?

csrc/validator_utils.h
Priya2698: While there is no restriction on the reduction type, the comments indicate the tolerances are for `sum` reductions. We can modify this table and tolerance computations (preferably in a separate PR though), but it seems that the values have been working fine for the existing tests. What do you think?

jacobhinkle:
> Are you suggesting combining the `execute` method with validation? Right now, there is a separate `fd.validate(inputs, reference_outputs)`.

I was, but I think I misunderstood what validate is doing. It's probably better as you have it so you can ignore me. 😁

csrc/python_frontend/fusion_definition.h
```diff
@@ -135,6 +135,10 @@ class FusionDefinition : public FusionState {
   std::optional<std::string> getDebugOutput() const {
     return debug_output_;
   }
+
+  std::vector<std::pair<double, double>> get_val_tolerances(
```
naoyam, line 139: nit: getValTolerances

python_benchmarks/normalization.py
```diff
@@ -47,16 +47,17 @@ def norm_fwd_fusion(
         weight = fd.ops.cast(weight, dtype=DataType.Float)
         bias = fd.ops.cast(bias, dtype=DataType.Float)
 
-    var, mean = fd.ops.var_mean(input, axes=reduction_axes, correction=0, keepdim=False)
```
naoyam, line 50: Isn't this PR supposed to be just about the validation?

Priya2698: This was an error in the fusion definition due to the order of operations which was causing the validation tests to fail for normalization benchmarks. Should we have this in a separate PR?

naoyam (approved):
LGTM

csrc/python_frontend/fusion_definition.h
```diff
@@ -135,6 +135,10 @@ class FusionDefinition : public FusionState {
   std::optional<std::string> getDebugOutput() const {
     return debug_output_;
   }
+
```
naoyam, line 138: nit: Please leave some comment

python_benchmarks/global_params.py
```diff
@@ -12,8 +13,8 @@ def generate_input_sizes(dims: Union[int, List] = 2) -> List[Tuple]:
     # TODO: Add more input sizes.
     for dim in dims:
         if dim == 2:
-            range_outer = [2**i for i in range(1, 5)]
-            range_inner = [32 * 1024 * 2**i for i in range(11)]
+            range_outer = [2**i for i in range(4, 9)]  # {16, 256}
```
naoyam, line 16: All these range changes are due to the validation failures?

naoyam:
@Priya2698 As we discussed offline, there are several things we still need to do, e.g., extending the supported ranges of validations. Please create issues to keep track of the remaining work.

python_benchmarks/global_params.py
Priya2698: Yes, in `layernorm`.

naoyam: OK, please leave a TODO comment as we are planning to address it

naoyam (approved):
LGTM.

PR #1154, opened 2023-10-26
Cleanup sync_information.cpp

zasdfgbnm:
I understand that if it is a broadcast, it does not matter, but why are we interested in duplicate parallelizing it?


zasdfgbnm:
!build

PR #1158, opened 2023-10-26
Avoid copy.

wujingyue:


wujingyue:
!build

wujingyue:
This will break https://github.com/NVIDIA/Fuser/blob/7db623f2eed57a3870924fa6ed75d0736786c0aa/csrc/tensor_metadata.cpp#L26, which expects a non-const reference. It triggers my curiosity about the two flavors of `ExpressionEvaluator::evaluate`, const and non-const. AFAIU, the only difference is that the former doesn't cache the intermediate and final results and the latter does cache them. However, I'm a bit surprised by that `Expr::evaluate` takes a `const ExpressionEvaluator&` and thus doesn't cache. Isn't that too wasteful? 

cc @zasdfgbnm

wujingyue:
> However, I'm a bit surprised by that Expr::evaluate takes a const ExpressionEvaluator& and thus doesn't cache. Isn't that too wasteful?

I think I got this part now. We don't cache intermediate values computed by `Expr::evaluate`, but we do cache its outputs [here](https://github.com/NVIDIA/Fuser/blob/6751f27a8e128c427f6fd4aedd58732e37fb4013/csrc/expr_evaluator.cpp#L202-L204).

zasdfgbnm:
Your understanding is correct. const expr evaluator avoid copies, but are slower during runtime. If you `evaluate` two different `Val`s, the second `evaluate` will have to recompute everything, even if it some intermediate value has already been computed in the first `evaluate`. I honestly don't have any idea on which in practice will be better. It would be nice if you can grab some fusion with allocation domain and check the time for this function.

PR #1160, opened 2023-10-26
Fix crash in diff_report.py when ptx_diff_lines is empty

jacobhinkle:
Fixes a failure in a recent codegen diff CI job. `diff_lines` should never be None, but check it anyway.


xwang233 (approved):
Thanks for the fix!

PR #1161, opened 2023-10-26
Renaming IterDomainGraphs to IdModel

naoyam:
I prefer to have `IdGraph` and `IdModel`, rather than `IdGraph` and `IdGraphs`


naoyam:
CC: @csarofeen

PR #1162, opened 2023-10-26
Set autoincrementing index in FusionExecutor::compileRtc

jacobhinkle:
compileRtc is used only in a few tests (seven currently) to test basic compilation functionality. Currently, calling compileRtc does not use fusion_id_counter_, and does not set fusion_id_ until after calling getStructuredCode. This means that if we print the kernel or compiled PTX, we get a file named like `__tmp_kernel-1.{cu,ptx}`. Here -1 is the default value of `fusion_id_`. This has caused a problem in codegen diff, which expects us to not overwrite these files, and for them to follow a somewhat regular naming scheme.

This change simply sets the ID before `getStructuredCode` using `fusion_id_counter_` to mimic what happens in `compileFusion`.

### DIFF
--- a/csrc/executor.cpp
+++ b/csrc/executor.cpp
@@ -1865,13 +1865,15 @@ void FusionExecutor::compileRtc(
       index_type == PrimDataType::Int || index_type == PrimDataType::Int32 ||
           "Invalid index type: ",
       index_type);
+
+  fusion_id_ = ++fusion_id_counter_;
+
   std::string scode;
   if (!structured) {
     scode = getStructuredCode(code, index_type);
   } else {
     scode = code;
   }
-  fusion_id_ = 1;
 
   compiled_kernel_ =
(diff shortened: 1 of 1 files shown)


jacobhinkle:
!build

jacobhinkle:
!build --diff

jacobhinkle:
!build --diff

jacobhinkle:
!build --diff

jacobhinkle:
With the latest change in #1330 we no longer have stuff like `__tmp_kernel-1.cu`, but we instead get wraparound, so you will find `__tmp_kernel_32195.cu` instead when running the binary tests. They actually autoincrement from there just fine, but it's a bit strange. Anyway, I believe this PR will fix it so that it works as expected anyway..

naoyam:
@jacobhinkle Is this still relevant?

jacobhinkle:
> @jacobhinkle Is this still relevant?

It is not. I need to do some spring cleaning on all of my old draft PRs

Issue #643, opened 2023-07-24
Cleanup legacy methods in `Val`


zasdfgbnm:
Current, `Val` has the following methods:
```C++
  // If this Val is an integer with a direct constant value associated with it,
  // will return the value of that constant integer. If this integer has
  // defining expressions it will return a std::nullopt. Those values should be
  // infered using evaluateInt.
  std::optional<int64_t> getInt() const;

  // If this Val is a double with a direct constant value associated with it,
  // will return the value of that constant double. If this double has
  // defining expressions it will return a std::nullopt. Those values should be
  // infered using evaluateDouble.
  std::optional<double> getDouble() const;

  // If this Val is a bool with a direct constant value associated with it,
  // will return the value of that constant bool. If this bool has defining
  // expressions it will return a std::nullopt. Those values should be infered
  // using evaluateBool.
  std::optional<bool> getBool() const;

  // If this Val is a constant integer, and its history is comprised only of
  // constant values, will return the value of that constant integer. Cannot
  // make constant as expression evaluator takes non-constant Vals.
  int64_t evaluateInt();

  // If this Val is a constant double, and its history is comprised only of
  // constant values, will return the value of that constant double. Cannot
  // make constant as expression evaluator takes non-constant Vals.
  double evaluateDouble();

  // If this Val is a constant bool, and its history is comprised only of
  // constant values, will return the value of that constant bool. Cannot
  // make constant as expression evaluator takes non-constant Vals.
  bool evaluateBool();
```
This was a legacy before we have a `PolymorphicValue`. With `PolymorphicValue`, we only need a single
```C++
const PolymorphicValue& value() const;
```
and a single
```C++
PolymorphicValue evaluate();
```
we should replace all uses of the above 6 legacy methods with the two new methods.

zasdfgbnm:
We can get the usage of these legacy methods using `grep`. The grep result below is based on commit https://github.com/NVIDIA/Fuser/commit/08ed7241a53786d678e85adbe372b1fa690f015e

```C++
❯ grep "getInt()" -r .
./csrc/ops/alias.cpp:    auto id_size = id->extent()->getInt();
./csrc/ops/alias.cpp:    auto id_size = new_sizes.at(i)->getInt();
./csrc/ops/arith.cpp:    auto expanded_size_int = expanded_sizes[i]->getInt();
./csrc/ops/arith.cpp:      auto inp_id_size_int = inp_id->extent()->getInt();
[...]

PR #1164, opened 2023-10-27
Staged replacement of the use of evaluate* and get* with evaluate() and value() [4/N]

protonu:
Replaces uses of evaluate(Int, Bool, Double) with evaluate which returns a PolymorphicValue.

This also replaces the uses of get(Int, Bool, Double) with value().

This PR is the fourth and final one of them to fix the issue. In the final PR, the older get* and evaluate* methods will be removed.

Fixes https://github.com/NVIDIA/Fuser/issues/643


protonu:
!build

protonu:
!build

Issue #986, opened 2023-09-29
We shouldn't require allocation domain to be between root and leaf


zasdfgbnm:
Our current implementation for swizzle is very hacky and broken: https://github.com/NVIDIA/Fuser/issues/554

Assume that we have a fusion
```
Tswizzled = T0
T1 = Tswizzled
```
the indexing we want for swizzle is like this
```
Tswizzled[swizzled_index] = T0[index]
T1[index] = Tswizzled[swizzled_index]
```
The most natural way to implement this behavior is by making allocation domain not required to be between root and leaf. Instead, for `Tswizzled`, we can have:
```
allocation domain: [I1, I2]
I3, I4 = swizzle2d(I1, I2)
root domain: [I3, I4]
... transformations ...
leaf domain: [..., BIDx, TIDx, Vec]
```
Then when indexing `Tswizzled` as consumer, we are already indexing into allocation domain, we will automatically get the swizzled index. When indexing `Tswizzled` as producer, we first replay `Tswizzled` as `T1`, which is trivial in this case. We do not need any hacks like "skipSwizzles" to support this.

csarofeen:
I don't disagree that we might want to support allocation domains not between root and leaf, but I'm a bit confused on this example. I think swizzle is hacky, yes, but I don't know why it's hacky relative to allocation domain. I was mentioning swizzle with @naoyam recently in how we want to model it in IterDomain graph which demonstrates the very strange behavior of how swizzle is treated today.

zasdfgbnm:
> hacky relative to allocation domain

What do you mean by this? I think what I am trying to say is, by using allocation domain, we can make it non-hacky.

csarofeen:
Swizzle is hacky, but I'm not understanding how allocation domain could play a role in reducing the hackyness.

zasdfgbnm:
> Swizzle is hacky, but I'm not understanding how allocation domain could play a role in reducing the hackyness.

If you have a fusion:
```
T1 = set(T0)
T2 = set(T1)
```
and you want to swizzle `T1`, with the current design, it will be scheduled like this:
<img width="610" alt="image" src="https://github.com/NVIDIA/Fuser/assets/1032377/d0552d79-3fcf-4905-b3a7-e3f7251f18b7">
[...]

PR #1167, opened 2023-10-27
ExpressionEvaluator validates allocation domain for Set.Permute. 

wujingyue:
This was a fun exercise. It's half done -- see my TODO. But I'm curious to hear your thoughts before I continue. 
1. Should ExpressionEvaluator guarantee to produce tensors that always respect allocation domain and contiguity? 
2. If so, should it guarantee that for just fusion outputs or every intermediate tensor as well? 
3. Is an empty allocation domain compatible with any stride order or with only row-major?

### DIFF
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -2402,6 +2402,10 @@ std::vector<PolymorphicValue> LoadStoreOp::evaluate(
           out_tv->getRootDomain(), out_tv->getRFactorDomain()));
 
       inferAndValidateAllocationSizesAndStrides(out_tensor, out_tv, ee);
+      // TODO: When validation fails, relayout `out_tensor` to make it
+      // compatible with `out_tv`'s allocation domain and contiguity. I haven't
+      // quite figured out how to do that especially when the allocation domain
+      // is not a permutation of `out_tv`'s rfactor domain.
       return {out_tensor};
     }
   }
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -2364,7 +2364,8 @@ LoadStoreOp::LoadStoreOp(
 namespace {
 // Returns the permutation from `in` to `out`, i.e., `out[i]==in[perm[i]]`. As a
 // precondition, `out` must be a permutation of `in` per the definition of
-// std::is_permutation.
+// std::is_permutation. The return type is `vector<int64_t>` instead of
+// `vector<int>` so it can be converted to `at::IntArrayRef`.
 template <typename T>
 std::vector<int64_t> computePermutation(
     const std::vector<T>& in,
@@ -2380,6 +2381,56 @@ std::vector<int64_t> computePermutation(
   }
   return permutation;
 }
+
+std::vector<int64_t> strideOrder(const at::Tensor& tensor) {
+  std::vector<std::pair<int64_t, int64_t>> stride_dim;
+  stride_dim.reserve(tensor.dim());
+  for (int64_t i = 0; i < tensor.dim(); i++) {
+    stride_dim.emplace_back(tensor.stride(i), i);
+  }
+
+  std::sort(
+      stride_dim.begin(),
+      stride_dim.end(),
+      std::greater<std::pair<int64_t, int64_t>>());
+
+  std::vector<int64_t> stride_order;
+  stride_order.reserve(tensor.dim());
+  for (const auto& [stride, dim] : stride_dim) {
+    stride_order.push_back(dim);
+  }
+  return stride_order;
+}
+
+bool strideOrdersAreCompatible(const TensorView& tv, const at::Tensor& tensor) {
+  const std::vector<IterDomain*>& rfactor = tv.getMaybeRFactorDomain();
+  const std::vector<IterDomain*>& allocation = tv.getMaybeAllocationDomain();
+
+  std::unordered_set<int64_t> rfactor_indices_to_ignore;
+  for (size_t i = 0; i < rfactor.size(); i++) {
+    if (rfactor[i]->isReduction() || rfactor[i]->isBroadcast()) {
+      rfactor_indices_to_ignore.insert(i);
+    }
+  }
+
+  std::vector<int64_t> tv_stride_order =
+      computePermutation(rfactor, allocation);
+  std::vector<int64_t> tensor_stride_order = strideOrder(tensor);
+
+  auto filter = [](const std::unordered_set<int64_t>& to_filter_out,
+                   std::vector<int64_t>& v) -> void {
+    v.erase(
+        std::remove_if(
+            v.begin(),
+            v.end(),
+            [&](const int64_t e) { return to_filter_out.count(e); }),
+        v.end());
+  };
+  filter(rfactor_indices_to_ignore, tv_stride_order);
+  filter(rfactor_indices_to_ignore, tensor_stride_order);
+  return tv_stride_order == tensor_stride_order;
+}
+
 } // namespace
 
 std::vector<PolymorphicValue> LoadStoreOp::evaluate(
@@ -2398,6 +2449,19 @@ std::vector<PolymorphicValue> LoadStoreOp::evaluate(
       at::Tensor in_tensor = inputs[0].as<at::Tensor>();
       at::Tensor out_tensor = in_tensor.permute(computePermutation(
           out_tv->getRootDomain(), out_tv->getRFactorDomain()));
+
+      // Check that `out_tv`'s allocation domain and `out_tensor.stride()` infer
+      // the same major-to-minor order. Otherwise, we need to apply `flatten`
+      // and `as_strided` on `out_tensor`, which is left as a todo.
+      NVF_ERROR(
+          strideOrdersAreCompatible(*out_tv, out_tensor),
+          out_tv->toString(),
+          " and ",
+          out_tensor.toString(),
+          out_tensor.sizes(),
+          ",strides=",
+          out_tensor.strides(),
+          " have incompatible strider orders.");
       return {out_tensor};
     }
   }
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -17,6 +17,7 @@
 #include <kernel_ir.h>
 #include <ops/arith.h>
 #include <root_domain_map.h>
+#include <tensor_metadata.h>
 #include <transform_iter.h>
 #include <transform_rfactor.h>
 #include <transform_view.h>
@@ -2381,56 +2382,6 @@ std::vector<int64_t> computePermutation(
   }
   return permutation;
 }
-
-std::vector<int64_t> strideOrder(const at::Tensor& tensor) {
-  std::vector<std::pair<int64_t, int64_t>> stride_dim;
-  stride_dim.reserve(tensor.dim());
-  for (int64_t i = 0; i < tensor.dim(); i++) {
-    stride_dim.emplace_back(tensor.stride(i), i);
-  }
-
-  std::sort(
-      stride_dim.begin(),
-      stride_dim.end(),
-      std::greater<std::pair<int64_t, int64_t>>());
-
-  std::vector<int64_t> stride_order;
-  stride_order.reserve(tensor.dim());
-  for (const auto& [stride, dim] : stride_dim) {
-    stride_order.push_back(dim);
-  }
-  return stride_order;
-}
-
-bool strideOrdersAreCompatible(const TensorView& tv, const at::Tensor& tensor) {
-  const std::vector<IterDomain*>& rfactor = tv.getMaybeRFactorDomain();
-  const std::vector<IterDomain*>& allocation = tv.getMaybeAllocationDomain();
-
-  std::unordered_set<int64_t> rfactor_indices_to_ignore;
-  for (size_t i = 0; i < rfactor.size(); i++) {
-    if (rfactor[i]->isReduction() || rfactor[i]->isBroadcast()) {
-      rfactor_indices_to_ignore.insert(i);
-    }
-  }
-
-  std::vector<int64_t> tv_stride_order =
-      computePermutation(rfactor, allocation);
-  std::vector<int64_t> tensor_stride_order = strideOrder(tensor);
-
-  auto filter = [](const std::unordered_set<int64_t>& to_filter_out,
-                   std::vector<int64_t>& v) -> void {
-    v.erase(
-        std::remove_if(
-            v.begin(),
-            v.end(),
-            [&](const int64_t e) { return to_filter_out.count(e); }),
-        v.end());
-  };
-  filter(rfactor_indices_to_ignore, tv_stride_order);
-  filter(rfactor_indices_to_ignore, tensor_stride_order);
-  return tv_stride_order == tensor_stride_order;
-}
-
 } // namespace
 
 std::vector<PolymorphicValue> LoadStoreOp::evaluate(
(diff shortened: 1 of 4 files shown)


wujingyue:
!build

csrc/ir/nodes.cpp
```diff
@@ -2398,6 +2400,12 @@ std::vector<PolymorphicValue> LoadStoreOp::evaluate(
       at::Tensor in_tensor = inputs[0].as<at::Tensor>();
       at::Tensor out_tensor = in_tensor.permute(computePermutation(
           out_tv->getRootDomain(), out_tv->getRFactorDomain()));
+
+      inferAndValidateAllocationSizesAndStrides(out_tensor, out_tv, ee);
```
jjsjann123, line 2404: oh, so we are just totally not using allocDomain at all here. :crying_cat_face: 

So the validation here would just always fail when we have permuted/transformed alloc domain?

jjsjann123 (commented):
I think we should respect explicit alloc domain and contiguity, be it an intermediate or output. If fusion API has it and it's going to be used by scheduler as well as fusion definition, stride in those sense becomes semantic and is no longer an implementation detail.

> Is an empty allocation domain compatible with any stride order or with only row-major?

For empty allocation domain, I tend to think this is left to interpretation. i.e. if there's no constraint, we don't necessarily enforce anything there.
i.e. feel free to create aliases if that means we are returning tensor in arbitrary stride order.

csrc/tensor_metadata.cpp
```diff
@@ -259,6 +260,24 @@ void validateAllocationSizesAndStrides(
           [](auto c_flag) { return c_flag.has_value(); }),
       "The size of contiguity mismatch with the dimensionality of allocation domain");
 
+  // Validate that strides must be in decreasing order modulo reduction and
```
zasdfgbnm, line 263: Why strides must be decreasing? If we have a discontiguous tensor, can't it be arbitrary?

csrc/ir/nodes.cpp
zasdfgbnm: Generally, I think it does make sense to validate with allocation domain, but my concern are:
1. `inferAndValidateAllocationSizesAndStrides` is slow, and IIUC `LoadStoreOp::evaluate` will become a hot path because it will be executing at runtime. At some point, we might need to worry about CPU overhead here.
2. I don't know how the alias analysis works, but, for example, if I have rFactor `[I1, I2]` and allocation `[I2, I1]`, then `view(-1)` should be invalid, and we should do `reshape(-1)`, which is not an alias. I think the alias analysis should take this into consideration. I think we can not really have an accurate check unless that is built out. But depending on how you would like to incrementally approach, I have no preference on whether we should add this check first or add complete handling allocation first.

wujingyue:
> I think we should respect explicit alloc domain and contiguity, be it an intermediate or output. If fusion API has it and it's going to be used by scheduler as well as fusion definition, stride in those sense becomes semantic and is no longer an implementation detail.

Makes sense -- that's most strict and costly to implement. I think it implies most of our `XyzOp::evaluate`s are wrong, because AFAICT none of them (except this one I'm going to submit) look into the allocation domain at all.

csrc/tensor_metadata.cpp
wujingyue: IIUC, the input `sizes` and `strides` are rearranged according to the allocation domain and therefore in major-to-minor order. So I'm expecting `strides` to be in decreasing order modulo broadcast and reduction. Otherwise, it's not compatible with the allocation domain.

csrc/ir/nodes.cpp
wujingyue: > inferAndValidateAllocationSizesAndStrides is slow, and IIUC LoadStoreOp::evaluate will become a hot path because it will be executing at runtime. At some point, we might need to worry about CPU overhead here.

Good point -- I'm quite concerned about that too. @jjsjann123 and I happened to talk about this the other day. An alternative is to do more symbolic execution at compile time. For example, if fusion IR could represent strides (as `Val*`s) and propagate these symbolic strides through meta ops like reshapes and transposes, we could spend less run time. This would involve a much larger change (possibly redefine allocation domain) so I didn't take that path for now. 

cc @tfogal

> I think the alias analysis should take this into consideration.

It does, e.g., https://github.com/NVIDIA/Fuser/blob/6751f27a8e128c427f6fd4aedd58732e37fb4013/csrc/optimization/alias_analysis.cpp#L103. The invariant I've been enforcing is: if alias analysis reports `out_tv` is an alias of `in_tv`, then ExpressionEvaluator is able to bind `in_tv` to any valid `in_tensor` (i.e. matching allocation domain and contiguity etc) and evaluate `out_tv` as a `out_tensor` that's an alias of `in_tensor`.

wujingyue: > oh, so we are just totally not using allocDomain at all here. 😿
> 
> So the validation here would just always fail when we have permuted/transformed alloc domain?

Right. In fact, I don't even know how to handle an allocation domain that's not merely a permutation of the rfactor domain (see my TODO below). For example, let's say the rfactor domain is [N,H,W,C] and the allocation domain is [N,C/32,H,W,32]. `out_tensor` must still have the shape of [N,H,W,C] (because the rfactor domain dictates the logical shape), but what are going to be the strides?! 

PS: It's interesting that, at this point, fusion IR is more expressive than ATen ops in some cases (e.g. the allocation domain example above) and less expressive in some other cases (e.g. not being able to represent strides). I'm getting nervous about inconsistency like this will bite us at some point in the future.

jjsjann123: > less expressive in some other cases (e.g. not being able to represent strides)

This feels more like a implementation thing. Up until this point, we only deal with strides in our indexing. So the concept of stride isn't part of our fusion IR.
I'm not saying your concern isn't legit. I don't know if there's something in our indexing code that can be unified/repurposed to represent stride.

jjsjann123:
I'm wondering if it makes sense to limit some of our analysis on alloc_domain to be `permutation of rfactor`.

I know @zasdfgbnm has use case of transformed alloc_domain to represent swizzle, but that's mostly just a scheduling thing?!?! Alias analysis happening before the scheduler shouldn't need to worry about that, since we are looking at fusion before scheduling any way.

wujingyue:
> I'm wondering if it makes sense to limit some of our analysis on alloc_domain to be `permutation of rfactor`.

It's not all about alias analysis. We call `ExpressionEvaluator` after all compilation. We can't stop people from using `ExpressionEvaluator` on a `ViewOp` whose allocation is not a rfactor permutation. Can we?

csrc/ir/nodes.cpp
zasdfgbnm: > I'm getting nervous about inconsistency like this will bite us at some point in the future.

For the place where we need to interface with ATen, we should check consistency and disallow things like `[N, C/32, H, W, 32]` but still allow things like `[N, H, W, C/32, 32]`. For intermediate tensors, I don't think we should have any restriction on what allocation domain can be.

csrc/tensor_metadata.cpp
zasdfgbnm: I think:
1. Fusion executor cache should reorder it as you described.
2. Even if fusion executor does not reorder it, as long as it is marked as discontinuity, the rest of our system should support it.

So I do think we need such a test, but not here.

csrc/ir/nodes.cpp
wujingyue: > For the place where we need to interface with ATen

Due to the power of `ExpressionEvaluator`, it's too hard to differentiate such places. A user can ask `ExpressionEvaluator` to turn any TensorView (including intermediate ones) into an `at::Tensor`. Furthermore, even if the user asks `ExpressionEvaluator` for a fusion output TensorView, `ExpressionEvaluator` internally computes all dependent `TensorView`s, many of which are intermediate, as `at::Tensor`s. 

AFAICS, for this particular case, it's probably best to proceed with the assumption that allocation must be an rfactor permutation. Because I don't see another good way around. Do you all agree? We should probably have another chat about what's the right assumption between allocation and rfactor in general.

csrc/tensor_metadata.cpp
wujingyue: I think we might have talked past each other. The reorder happens [here](https://github.com/NVIDIA/Fuser/blob/b4ef869c36de5b36c67f3f118c59012d618b434f/csrc/tensor_metadata.cpp#L318) in `inferAndValidateAllocationSizesAndStrides`. `validateAllocationSizesAndStrides` is an internal function whose input `sizes` and `strides` are guaranteed to be reordered (and thus my expectation). Does that make sense?

csrc/ir/nodes.cpp
zasdfgbnm: > AFAICS, for this particular case, it's probably best to proceed with the assumption that allocation must be an rfactor permutation.

This is definitely better than just ignoring the allocation domain and produce silent wrong result. I think we can start from here.

csrc/tensor_metadata.cpp
zasdfgbnm: Are we talking about the same reorder? I think there are two different "reorder"s we can think about. One reorder is to reorder the computed sizes and strides to be consistent with the allocation domain. This reorder does not change allocation domain itself. For this reordering, it happens in the code you refers to. But the "reorder" I was referring to is to set the allocation domain so that the order of the allocation domain matches with the stride order of the ATen tensor. This is a process similar to how we set the contiguity of a tensor.

wujingyue: Got it. This goes back to my question 3 in OP. AFAICS, @jjsjann123 proposed that an empty allocation domain means no constraint, which I agree. So here I shouldn't expect sizes and strides to be in certain order when the allocation domain is empty. Sgty?

zasdfgbnm: > I shouldn't expect sizes and strides to be in certain order in certain order when the allocation domain is empty.

I think we shouldn't expect sizes and strides to be in certain order in all cases. For example, if my allocation domain is `[I1, I2]` and my stride is `[1, 10]`, although this is not performant, I don't see any reason why we should reject this. This can be handled as a general discontiguous tensor. It's not the most performant kernel, but it is functionally correct. I think our lowering and execution should be tolerant to all good and bad schedules, having a non-decreasing stride order is a bad schedule, but not wrong. This is just like C++ should support all good and bad coding styles, passing by value instead of by `const &` can be slow, but the C++ compiler should not reject the program.

zasdfgbnm: I consider nvFuser as two parts:
1. Fusion, lowering, executor - a programming language like C++
2. Schedulers, executor cache - an automatic code generator that generates programs written in the above language

wujingyue: > For example, if my allocation domain is [I1, I2] and my stride is [1, 10], although this is not performant, I don't see any reason why we should reject this. 

I'm confused. What's the purpose of allocation domain if it doesn't dictate a physical memory layout (and therefore a stride order)?

jjsjann123: Xiang seems to be suggesting us to relax what `allocDomain` means. i.e. `allocDomain` is only semantically meaningful when we have the corresponding contiguity flag set as `true`?

I felt this makes some sense from the design of indexing. nitpicking on this topic,  it felt a bit sloppy conceptually. Given that `allocDomain` and `contiguity` are two separate field. But our API only allows the mutation of both field together, so maybe we are not that far away from the reality that `allocDomain` and `contiguity` should come as a single package :shrug: 

both protocol sound fine to me.

wujingyue: > allocDomain is only semantically meaningful when we have the corresponding contiguity flag set as true?

Interesting. Does that mean `allocation_domain[i]` is completely useless when `contiguity[i]==false`?

tfogal:
I have a lot to catch up on here, so apologies for not commenting technically.

At a more logistical level: it seems we're discussing semantics (and a possible change in semantics) of core components of nvFuser's IR. I suggest we don't have these discussions inside a github PR, or issue, etc., but rather as a group at, say, the Tuesday meeting series.

csrc/tensor_metadata.cpp
zasdfgbnm: > What's the purpose of allocation domain if it doesn't dictate a physical memory layout (and therefore a stride order)?

It does dictate the physical memory layout, but "dictate the physical memory layout" can not be simply translated to "require the stride order to be decreasing". For example, if I have a tensor whose rFactor domain is `[I1, I2, I3]`, shape is `(4, 5, 6)`, and stride is `(6, 2, 1)`, then it might makes more sense to have allocation domain `[I2, I1, I3]` and contiguity `(false, true, true)`, instead of having allocation domain `[I1, I2, I3]` and contiguity `(false, false, true)`.

zasdfgbnm: > maybe we are not that far away from the reality that `allocDomain` and `contiguity` should come as a single package 🤷

I think they are already a single package, no?

jjsjann123:
> I have a lot to catch up on here, so apologies for not commenting technically.
> 
> At a more logistical level: it seems we're discussing semantics (and a possible change in semantics) of core components of nvFuser's IR. I suggest we don't have these discussions inside a github PR, or issue, etc., but rather as a group at, say, the Tuesday meeting series.

Good point!

We do have tomorrow's meeting scheduled for discussion on `allocDomain`, I'll put a slide on this topic and get some input from the group on this.

csrc/tensor_metadata.cpp
jjsjann123: Re @zasdfgbnm 

> I think they are already a single package, no?

Conceptually I think yes. But looking at the implementation we have, it's likely a no... hence the confusion that Jingyue is having regarding the topic.

Re @wujingyue 

> >  allocDomain is only semantically meaningful when we have the corresponding contiguity flag set as true?
> 
> Interesting. Does that mean allocation_domain[i] is completely useless when contiguity[i]==false?

Sorta yes. You would also need `contiguity[i-1] == false`.
But this isn't a definitive answer there. As @zasdfgbnm points out, we can still rely on the order for scheduler, but that's should be just a performance hint.
I think Xiang's point here is that, we shouldn't assert on stride order when corresponding contiguity flags were set to false.
I agree with this ^^^. Even though in general, I'm not a big fan of  "be conservative in what you do, be liberal in what you accept from others" :laughing:

wujingyue:
This PR needs more work. 

#986 explains why memory swizzling requires allocation domain to be more than an rfactor permutation. This means that `ExpressionEvaluator::evaluate(TensorView*)` might not always be able to return a valid `at::Tensor`, e.g., when the input `TensorView` has a rfactor domain of `[i0,i1]` and an allocation domain of `[i0/2,i1/2,4]`. 

This leaves `ExpressionEvaluator` two options to treat allocation domains. 
1. The strict approach. `ExpressionEvaluator::evaluate(TensorView*)` errors out when the allocation domain of the input TensorView is beyond what strides can represent. This applies to intermediate `TensorView`s as well because `ExpressionEvaluator::evaluate` is recursive. 
2. The loose approach. `ExpressionEvaluator::evaluate(TensorView*)` takes the best effort to produce an `at::Tensor` that complies with the allocation domain. This way, the user of `ExpressionEvaluator::evaluate(TensorView*)` is responsible to check the allocation domain required (e.g. by alias analysis or by Framework) is compatible with the output strides. This check likely only needs to happen at fusion outputs. 

I haven't decided which option to take. Chances are, in the short term, they only matter in terms of implementation complexity.

PR #1171, opened 2023-10-27
overload set to accept cache op

liqiangxl:
add another overload of `set` function to accept cache op.
separated from #1173.

### DIFF
--- a/csrc/ops/alias.cpp
+++ b/csrc/ops/alias.cpp
@@ -22,8 +22,18 @@ Val* set(Val* v) {
   return out;
 }
 
+Val* set(Val* v, CacheOp cache_op) {
+  Val* out = ops::newValLike(v, v->getDataType().value());
+  IrBuilder::create<LoadStoreOp>(LoadStoreOpType::Set, out, v, cache_op);
+  return out;
+}
+
 TensorView* set(TensorView* tv) {
   return set(tv->as<Val>())->as<TensorView>();
+}
+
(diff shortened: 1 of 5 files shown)


csrc/ops/alias.cpp
```diff
@@ -22,10 +22,20 @@ Val* set(Val* v) {
   return out;
 }
 
+Val* set(Val* v, CacheOp cache_op) {
```
wujingyue, line 25: How about giving `cache_op` here the default value `CacheOp::Unspecified` and remove the other flavor of `set` that doesn't take `cache_op`?

wujingyue (changes requested):

naoyam:
Why do we need to have this? I'm a bit nervous about exposing scheduling options through the user-facing interface.

liqiangxl:
> Why do we need to have this? I'm a bit nervous about exposing scheduling options through the user-facing interface.

Just for convenience, allows us to do `auto x1 = set(x, CacheOp::AllLevels);` instead of doing extra code to change the cache op.
```
  TensorView* tv1 =
      ops::newValLike(tv0, tv0->getDataType().value())->as<TensorView>();
  IrBuilder::create<LoadStoreOp>(LoadStoreOpType::Set, tv1, tv0, cache_op);
```
If this is a concern, I can close this PR.

naoyam:
The ops interface is the one we provide to the user and should not have options like this. Otherwise, we would need to deal with user-specified caching options.

PR #1176, opened 2023-10-27
Add simple pass manager for `GpuLower`

zasdfgbnm:
`GpuLower`'s ctor now only does analysis. Users needs to call `GpuLower::run` to run passes. Passes are `std::vector<std::pair<std::string, std::function<std::vector<Expr*>(const std::vector<Expr*>&)>>`, and are exposes to the public as `GpuLower::passes()`. Users can modify it. The main use case for this mechanism is for testing purposes.

One demo is the test added in `test_gpu3.cpp`:

```C++
TEST_F(NVFuserTest, LoweringHook) {
  Fusion fusion;
  FusionGuard fg(&fusion);

  fusion.addInput(tv0);
  auto tv1 = set(tv0);
  fusion.addOutput(tv1);

  GpuLower gpulw(&fusion);
  bool executed = false;
  gpulw.passes().push_back(
      {"test",
       [&executed](const std::vector<Expr*>& exprs) -> std::vector<Expr*> {
         executed = true;
         return exprs;
       }});
  EXPECT_FALSE(executed);
  gpulw.run();
  EXPECT_TRUE(executed);
}
```

This should be just refactoring, no codegen change.


zasdfgbnm:
!build

zasdfgbnm:
!build --diff

zasdfgbnm:
!build --diff

zasdfgbnm:
!build --diff

jacobhinkle:
You may want to update and rerun `!build --diff` since I just merged #1214 which should squelch all those false positive PTX diffs.

zasdfgbnm:
!build --diff

zasdfgbnm:
ping @naoyam

csrc/device_lower/lower2device.h
```diff
@@ -51,16 +52,17 @@ class GpuLower : public NonCopyable {
  public:
   GpuLower() = delete;
 
+  using Pass = std::pair<
```
naoyam, line 55: nit: some comment about `Pass` would be great to have, e.g., what the string means, what the function input and output are.

csrc/device_lower/lower2device.h
```diff
@@ -71,6 +73,9 @@ class GpuLower : public NonCopyable {
   //! Query if lowering is in progress
   static bool hasCurrent();
 
+  //! Actually run the lowering
```
naoyam, line 76:
```suggestion
  //! Actually run the lowering by executing the passes in the order
```

csrc/device_lower/lower2device.h
```diff
@@ -230,6 +243,9 @@ class GpuLower : public NonCopyable {
   // Lowered Kernel IR
   std::unique_ptr<kir::Kernel> kernel_;
 
+  // Passes to lower kernel
```
naoyam, line 246: Please also mention they are executed in this order

naoyam (approved):
LGTM. Thanks for the refactoring!

csrc/device_lower/lower2device.h
zasdfgbnm: added

csrc/device_lower/lower2device.h
zasdfgbnm: added

csrc/device_lower/lower2device.h
zasdfgbnm: added

zasdfgbnm:
!build

zasdfgbnm:
!build

test/test_gpu3.cpp
```diff
@@ -8413,8 +8444,10 @@ TEST_F(NVFuserTest, AlignedSyncReduction1_CUDA) {
 
   scheduler_utils::parallelizeAllLike(tv2);
 
-  const std::string kernel_string =
-      codegen::generateCudaKernel(GpuLower(&fusion).kernel());
+  GpuLower gpulw(&fusion);
+  gpulw.run();
```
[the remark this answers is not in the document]
zasdfgbnm: done

PR #1179, opened 2023-10-27
Fix TMATest by handling constant Array type

protonu:
Fix the failing MATest.LoadCompleteTensor*D test by handling constant Array types.

### DIFF
--- a/csrc/codegen.cpp
+++ b/csrc/codegen.cpp
@@ -444,6 +444,58 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
     code_ << gen(pred->value());
   }
 
+  void stringify(const PolymorphicValue& value, const DataType dtype) {
+    if (value.is<bool>()) {
+      code_ << (value ? "true" : "false");
+    } else if (value.is<int64_t>()) {
+      code_ << value << getLiteralSuffix(dtype);
+    } else if (value.is<double>()) {
+      auto val = value.as<double>();
+      // note: default inf/nan doesn't work and should be replaced with macros
+      // `NAN`, `POS_INFINITY` and `NEG_INFINITY` instead.
+      if (std::isinf(val)) {
+        if (val > 0) {
+          code_ << "POS_INFINITY";
+        } else {
+          code_ << "NEG_INFINITY";
+        }
+      } else if (std::isnan(val)) {
+        code_ << "NAN";
+      } else {
+        setPrecision(code_, dtype);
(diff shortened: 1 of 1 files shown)


protonu:
!build

csrc/codegen.cpp
```diff
@@ -493,6 +493,19 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
...
+        NVF_ERROR(value.is<std::vector>());
+        auto dims = static_cast<int>(value.as<std::vector>().size());
+
+        // print out the vector of numbers.
+        code_ << to_str(s->dtype()) << " {";
+        for (auto i = 0; i < dims; i++) {
+          if (i > 0) {
+            code_ << ", ";
+          }
+          code_ << value[i];
```
zasdfgbnm, line 506: Should we make this section of code a separate function, say, for example `stringify`, probably at `type.cpp` and use `stringify(value[i])`?

protonu: Sure! I felt like this should have been available in some library somewhere :)

protonu: A qq, do you mean stringify'ing the vector of values or just that one line `code_ << X'

zasdfgbnm: I mean to have a `std::string stringify(PolymorphicValue, DataType)` which recursively calls itself for container types.

protonu: I see - would it make sense to close this PR, and land the other PR which use Polymorphic Vals, and then I can add this  function to clean it up.

csrc/codegen.cpp
```diff
@@ -493,6 +493,9 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
           NVF_ERROR(dtype == DataType::ComplexDouble);
           code_ << "std::complex<double>" << value;
         }
+      } else if (std::holds_alternative<ArrayType>(s->dtype().type)) {
+        // print out the vector of numbers.
+        code_ << to_str(s->dtype()) << stringifyPolymorphicValAsVector(value);
```
zasdfgbnm, line 498: Should we put the entire thing, starting from line 469 into this function?

protonu:
!build

csrc/codegen.cpp
protonu: I had missed this comment. This makes sense. Let me update the PR.

protonu:
!build

zasdfgbnm (approved):

protonu:
!build

PR #838, opened 2023-09-05
Minor cleanup in scalar hoisting

zasdfgbnm:

PR #1183, opened 2023-10-28
Recover test_scalar_hoisting.

wujingyue:
I'm not sure why it wasn't added or was removed from CMakeLists.txt.

### DIFF
--- a/test/test_scalar_hoisting.cpp
+++ b/test/test_scalar_hoisting.cpp
@@ -300,7 +300,7 @@ TEST_F(ScalarHoistTest, IndexHoist3) {
   const std::string expected_kernel = R"(
 __global__ void CUDAGeneratedKernel(Tensor<float, 2, 2> T0, Tensor<float, 2, 2> T2) {
   nvfuser_index_t i0;
-  i0 = ((nvfuser_index_t)threadIdx.x) + (256 * ((nvfuser_index_t)blockIdx.x));
+  i0 = ((nvfuser_index_t)threadIdx.x) + (256LL * ((nvfuser_index_t)blockIdx.x));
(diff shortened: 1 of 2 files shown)


test/test_scalar_hoisting.cpp
```diff
@@ -300,33 +300,33 @@ TEST_F(ScalarHoistTest, IndexHoist3) {
   const std::string expected_kernel = R"(
 __global__ void CUDAGeneratedKernel(Tensor<float, 2, 2> T0, Tensor<float, 2, 2> T2) {
   nvfuser_index_t i0;
-  i0 = ((nvfuser_index_t)threadIdx.x) + (256 * ((nvfuser_index_t)blockIdx.x));
+  i0 = ((nvfuser_index_t)threadIdx.x) + (256LL * ((nvfuser_index_t)blockIdx.x));
```
wujingyue, line 303: We'd better get rid of this. It tests implementation not behavior...

naoyam:
Looks like the test file was added in #838 but the CMake file was not updated. CC: @zasdfgbnm

test/test_scalar_hoisting.cpp
naoyam: The intention here is to verify hoisting of a scalar is indeed performed, so the literal matching is sufficient but not necessary. Some more intelligent matching would reduce false alarms, but we shouldn't just remove this check.

wujingyue: Sure. I'll leave this test case as is, but hope someone will fix it later.

wujingyue:
!build

zasdfgbnm (approved):

PR #1186, opened 2023-10-30
avoid inline gmem data load and usage

liqiangxl:
**Purpose:** improve performance of softmax

**Method:**
(1) Previously, `inlineMost` was used by default to inline all computations, but tests showed that separating data loading from computation can increase performance. 
(2) This PR adjusts the inner persistent scheduler behavior by introducing the `is_inline_all_tvs` parameter. `is_inline_all_tvs` defaults to `true`, and is set to `false` for vectorized cases with persistent batch sizes larger than one (already separated for persistent batch size of one).
(3) When `is_inline_all_tvs` is false, the scheduler adopts an all-or-nothing approach to inlining: it will inline cached input tvs at a higher level only if they or their consumers are persistent tvs, thereby separating data loading from computation without increasing register usage for persistent bufffers. If any tv does not fulfill this condition, the default inlining is applied, as the performance impact of partial separation is minimal.

**Results:**
(1) current implementation only influence softmax forward/backward out of all the cpp benchmarks. see codegen diff.
(2) Due to limited coverage of the cpp benchmarks, very few noticable changes. The following figures showed all softmax forward/backward cases in the cpp benchmarks tested on A100 and H100.
A100:
image
H100:
image

(3) More python benchmarks are tested.
Performance change on **A100, batch = 2K**, feature = [64, 128, ..., 32K]:
image

Performance change on **A100, batch = 32K**, feature = [64, 128, ..., 32K]:
image

Performance change on **H100, batch = 2K**, feature = [64, 128, ..., 32K]:
image

Performance change on **H100, batch = 32K**, feature = [64, 128, ..., 32K]:
image


liqiangxl:
!build --diff

liqiangxl:
!build --diff

liqiangxl:
!build --diff

liqiangxl:
Hi @xwang233 , looks like codediff lists many false positive cases, e.g. [out of listed 138 cases, only 2 cases are changed](https://dl.gitlab-master-pages.nvidia.com/pytorch/fuser-gh-mirror//nvfuser_github_ci/codegen_diff_p10518179_j73007876_1698728846533237756_codediff_17e4d4f_b4ef869_custom_command_20231030_212134.html)

xwang233:
> Hi @xwang233 , looks like codediff lists many false positive cases, e.g. [out of listed 138 cases, only 2 cases are changed](https://dl.gitlab-master-pages.nvidia.com/pytorch/fuser-gh-mirror//nvfuser_github_ci/codegen_diff_p10518179_j73007876_1698728846533237756_codediff_17e4d4f_b4ef869_custom_command_20231030_212134.html)

cc @jacobhinkle

jacobhinkle:
> > Hi @xwang233 , looks like codediff lists many false positive cases, e.g. [out of listed 138 cases, only 2 cases are changed](https://dl.gitlab-master-pages.nvidia.com/pytorch/fuser-gh-mirror//nvfuser_github_ci/codegen_diff_p10518179_j73007876_1698728846533237756_codediff_17e4d4f_b4ef869_custom_command_20231030_212134.html)
> 
> cc @jacobhinkle

Ah yes this is the PTX `callseq` comment non-repeatability that Xiao and I noticed a few days ago. We should strip comments in the PTX when computing the diff. I had thought this would be more rare, but apparently it's happening a lot. I'll update the script.

liqiangxl:
!build  --diff-bench

jacobhinkle:
@liqiangxl as a workaround until #1195 is merged, if you didn't change anything in `runtime/` then you can just scan down through the diff and look at only the rows where the `CUDA` button is yellow, since the color indicates diff.

jacobhinkle:
!build --diff

jacobhinkle:
Oh @liqiangxl you may need to update the branch before running code diff again to see the fix.

liqiangxl:
> Oh @liqiangxl you may need to update the branch before running code diff again to see the fix.

Thanks for the quick patch.

liqiangxl:
!build --diff-bench

liqiangxl:
!build --diff-bench

liqiangxl:
!build --diff --diff-bench

liqiangxl:
!build --diff --diff-bench

liqiangxl:
!build --diff --diff-bench

liqiangxl:
!build --diff --diff-bench

liqiangxl:
!build --diff --diff-bench

liqiangxl:
perf was improved by #1408

PR #1193, opened 2023-10-31
Don't predicate trivial splits

jacobhinkle:
This small change just avoids adding predicates when an `IterDomain` is split by a constant factor of 1. The predicate is not needed in this case since 1 evenly divides any extent. It also moves a repeated independent check outside a loop.

### DIFF
--- a/csrc/device_lower/analysis/predicate_elimination.cpp
+++ b/csrc/device_lower/analysis/predicate_elimination.cpp
@@ -159,6 +159,11 @@ class PredicateAnalyzer : public OptOutDispatch {
       return;
     }
 
+    if (factor.value() == 1) {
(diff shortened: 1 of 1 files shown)


jacobhinkle:
!build

jacobhinkle:
!build --diff

jacobhinkle:
!build --diff

zasdfgbnm (approved):

jacobhinkle:
!build --diff

jacobhinkle:
@xwang233 heads up that the CI died at the `dnf` step of the docker build. See https://gitlab-master.nvidia.com/dl/pytorch/fuser-gh-mirror/-/jobs/73067209

xwang233:
yeah, that step is always flaky, I restarted the job

PR #1200, opened 2023-11-01
Add test `AllocationDomainTest.VectorizeOverlappingTensor`

zasdfgbnm:
Adding an example showing that why the order of allocation domain does not always match with the order of stride. Please let me know if it makes sense to you @wujingyue @jjsjann123 

Generated code:
```CUDA
__global__ void kernel1(Tensor<float, 3, 3> T0, Tensor<float, 3, 3> T2) {
  NVFUSER_DEFINE_MAGIC_ZERO;
  nvfuser_index_t i0;
  i0 = 4LL * ((nvfuser_index_t)threadIdx.x);
  nvfuser_index_t i1;
  i1 = i0 / T0.logical_size[2LL];
  nvfuser_index_t i2;
  i2 = T0.logical_size[2LL] * T0.logical_size[1LL];
  bool b3;
  b3 = (3LL + i0) < ((T0.logical_size[1LL] * T0.logical_size[0LL]) * T0.logical_size[2LL]);
  Array<float, 4LL, 4> T1;
  if (b3) {
    loadGlobalToLocal<float, /*vec_size=*/4, /*is_volatile=*/false, CacheOp::Streaming>(&T1[0LL],  &T0[(((T0.alloc_stride[0LL] * (i1 / T0.logical_size[0LL])) + (T0.logical_size[2LL] * (i1 % T0.logical_size[0LL]))) + (i0 % T0.logical_size[2LL]))]);
  }
  #pragma unroll
  for(nvfuser_index_t i4 = 0; i4 < 4LL; ++i4) {
    nvfuser_index_t i5;
    i5 = i0 + (i4 + nvfuser_zero);
    nvfuser_index_t i6;
    i6 = i5 / T0.logical_size[2LL];
    if (b3) {
      T2[(((i2 * (i6 % T0.logical_size[0LL])) + (T0.logical_size[2LL] * (i6 / T0.logical_size[0LL]))) + (i5 % T0.logical_size[2LL]))]
         = T1[i4];
    }
  }
  NVFUSER_UPDATE_MAGIC_ZERO;
}
```

### DIFF
--- a/test/test_allocation_domain.cpp
+++ b/test/test_allocation_domain.cpp
@@ -1195,6 +1195,15 @@ TEST_F(NVFuserTest, AllocationDomainContiguityForExplicitBroadcast) {
   testValidate(fusion, outputs, {t0}, {t1}, __LINE__, __FILE__);
 }
 
+// Test that allocation domain can be used to vectorize overlapping tensors,
+// by making the allocation domain deviate from the stride order. Note that
+// this test is only a demo "hey, we can do this", instead of checking for
+// a real use case. Supporting overlapping tensor is a gray area for framework,
+// and we are not actively using the trick in this test to generate a better
+// kernel for overlapping tensors. The only reason why this test exists is
+// because I think it is a good sign that we have a good design (a good design
+// automatically supports all kinds of use cases, even those that we don't have
+// an active plan to support on).
 TEST_F(AllocationDomainTest, VectorizeOverlappingTensor) {
   auto fusion_ptr = std::make_unique<Fusion>();
   Fusion& fusion = *fusion_ptr.get();
@@ -1227,6 +1236,8 @@ TEST_F(AllocationDomainTest, VectorizeOverlappingTensor) {
   }
   tv1->axis(1)->parallelize(ParallelType::Vectorize);
 
+  // Note that the stride of the second dimension of the input tensor must be a
+  // multiple of 4, otherwise we will have misaligned address access.
   at::Tensor t0 =
       at::randn({4 * 5 * 7}).cuda().as_strided({4, 5, 7}, {7, 4, 1});
 
--- a/test/test_allocation_domain.cpp
+++ b/test/test_allocation_domain.cpp
@@ -1214,17 +1214,21 @@ TEST_F(AllocationDomainTest, VectorizeOverlappingTensor) {
       {tv0->axis(1), tv0->axis(0), tv0->axis(2)}, {false, true, true});
 
   for (auto tv : {tv2, tv1}) {
+    // [I0, I1, I2]
     tv->reorder({{0, 1}});
+    // [I1, I0, I2]
     tv->merge(0);
+    // [I1*I0, I2]
     tv->merge(0);
+    // [I1*I0*I2]
     tv->split(0, 4);
+    // [I1*I0*I2/4, 4]
     tv->axis(0)->parallelize(ParallelType::TIDx);
   }
   tv1->axis(1)->parallelize(ParallelType::Vectorize);
 
-  auto options = at::TensorOptions().dtype(at::kFloat).device(at::kCUDA, 0);
   at::Tensor t0 =
-      at::randn({4 * 5 * 7}, options).as_strided({4, 5, 7}, {7, 4, 1});
+      at::randn({4 * 5 * 7}).cuda().as_strided({4, 5, 7}, {7, 4, 1});
 
   FusionExecutor fe;
   fe.compileFusion(fusion_ptr.get(), {t0});
--- a/test/test_allocation_domain.cpp
+++ b/test/test_allocation_domain.cpp
@@ -1195,4 +1195,42 @@ TEST_F(NVFuserTest, AllocationDomainContiguityForExplicitBroadcast) {
   testValidate(fusion, outputs, {t0}, {t1}, __LINE__, __FILE__);
 }
 
+TEST_F(AllocationDomainTest, VectorizeOverlappingTensor) {
+  auto fusion_ptr = std::make_unique<Fusion>();
+  Fusion& fusion = *fusion_ptr.get();
+  FusionGuard fg(&fusion);
+
+  auto tv0 = makeContigTensor(3);
+  fusion.addInput(tv0);
+  auto tv1 = set(tv0);
+  auto tv2 = set(tv1);
+  fusion.addOutput(tv2);
+
+  // According to the stride order below, the allocation domain should be the
+  // same as the root domain. However, here we intentionally make the allocation
+  // domain [axis(1), axis(0), axis(2)] because doing so allows us to vectorize
+  // by 4.
+  tv0->setAllocationDomain(
+      {tv0->axis(1), tv0->axis(0), tv0->axis(2)}, {false, true, true});
+
+  for (auto tv : {tv2, tv1}) {
+    tv->reorder({{0, 1}});
+    tv->merge(0);
+    tv->merge(0);
+    tv->split(0, 4);
+    tv->axis(0)->parallelize(ParallelType::TIDx);
+  }
+  tv1->axis(1)->parallelize(ParallelType::Vectorize);
+
+  auto options = at::TensorOptions().dtype(at::kFloat).device(at::kCUDA, 0);
+  at::Tensor t0 =
+      at::randn({4 * 5 * 7}, options).as_strided({4, 5, 7}, {7, 4, 1});
+
+  FusionExecutor fe;
+  fe.compileFusion(fusion_ptr.get(), {t0});
+  auto cg_outputs = fe.runFusion({t0});
+
+  testValidate(&fusion, cg_outputs, {t0}, __LINE__, __FILE__);
+}
+
 } // namespace nvfuser


zasdfgbnm:
!build

test/test_allocation_domain.cpp
```diff
@@ -1195,4 +1195,42 @@ TEST_F(NVFuserTest, AllocationDomainContiguityForExplicitBroadcast) {
...
+
+  // According to the stride order below, the allocation domain should be the
+  // same as the root domain. However, here we intentionally make the allocation
+  // domain [axis(1), axis(0), axis(2)] because doing so allows us to vectorize
+  // by 4.
+  tv0->setAllocationDomain(
+      {tv0->axis(1), tv0->axis(0), tv0->axis(2)}, {false, true, true});
+
+  for (auto tv : {tv2, tv1}) {
+    tv->reorder({{0, 1}});
```
wujingyue, line 1217: This shows my ignorance about nvFuser :), but can you clarify what this reorder is doing? I don't quite follow the comment there. 

```
// Reorder axes according to old2new[old_pos] = new_pos
```

test/test_allocation_domain.cpp
```diff
@@ -1195,4 +1195,42 @@ TEST_F(NVFuserTest, AllocationDomainContiguityForExplicitBroadcast) {
...
+    tv->merge(0);
+    tv->merge(0);
+    tv->split(0, 4);
+    tv->axis(0)->parallelize(ParallelType::TIDx);
+  }
+  tv1->axis(1)->parallelize(ParallelType::Vectorize);
+
+  auto options = at::TensorOptions().dtype(at::kFloat).device(at::kCUDA, 0);
+  at::Tensor t0 =
+      at::randn({4 * 5 * 7}, options).as_strided({4, 5, 7}, {7, 4, 1});
```
wujingyue, line 1227: Just to confirm my understanding, the `4` in `{7,4,1}` can be changed to any value between 1 and 7, exclusively, without affecting the behavior of the test. Correct?

test/test_allocation_domain.cpp
```diff
@@ -1195,4 +1195,42 @@ TEST_F(NVFuserTest, AllocationDomainContiguityForExplicitBroadcast) {
...
+  for (auto tv : {tv2, tv1}) {
+    tv->reorder({{0, 1}});
+    tv->merge(0);
+    tv->merge(0);
+    tv->split(0, 4);
+    tv->axis(0)->parallelize(ParallelType::TIDx);
+  }
+  tv1->axis(1)->parallelize(ParallelType::Vectorize);
+
+  auto options = at::TensorOptions().dtype(at::kFloat).device(at::kCUDA, 0);
```
wujingyue, line 1225: I saw some of your tests write `at::randn(...).cuda().as_strided(...)` without creating a verbose `TensorOptions`. I sightly prefer that for simplicity.

wujingyue (commented):
Thanks, Xiang

test/test_allocation_domain.cpp
zasdfgbnm: Yes, except that it must be compatible with the vectorization factor. If you still want to vectorize 4, then it must be 4 or 0. If you want to vectorize 2, then it can be 0, 2, 4, 6

test/test_allocation_domain.cpp
zasdfgbnm: I added comment to before or after each step to the code:
```C++
  for (auto tv : {tv2, tv1}) {
    // [I0, I1, I2]
    tv->reorder({{0, 1}});
    // [I1, I0, I2]
    tv->merge(0);
    // [I1*I0, I2]
    tv->merge(0);
    // [I1*I0*I2]
    tv->split(0, 4);
    // [I1*I0*I2/4, 4]
    tv->axis(0)->parallelize(ParallelType::TIDx);
  }
```

Basically, how I think through the reorder is like this, for example, if I have a 5d tensor
```C++
[I0, I1, I2, I3, I4]
```
and I do `tv->reorder({{0, 1}, {1, 2}, {2, 3}})`, then I will think like this:

Let me start from
```
old: [I0, I1, I2, I3, I4]
new: [??, ??, ??, ??, ??]
```
Let me traverse the list one by one, the first item says "what was at 0 is moved to 1", then I will do:
```
old: [__, I1, I2, I3, I4]
new: [??, I0, ??, ??, ??]
```
Then the next item says "what was at 1 is moved to 2", which is:
```
old: [__, __, I2, I3, I4]
new: [??, I0, I1, ??, ??]
```
Then the next item says "what was at 2 is moved to 3", which is:
```
old: [__, __, __, I3, I4]
new: [??, I0, I1, I2, ??]
```
Now I have traversed all `old2new`, I will fill the remaining IterDomains from old to new in order:
```
old: [__, __, __, __, __]
new: [I3, I0, I1, I2, I4]
```

test/test_allocation_domain.cpp
wujingyue: > If you still want to vectorize 4, then it must be 4 or 0.

How so? ({4,5,7},{7,x,1}) can be viewed as ({5,28},{x,1}). Because 28 is a multiple of 4, we can always vectorize the tensor by 4 regardless of `x`. Correct?

test/test_allocation_domain.cpp
wujingyue: > Now I have traversed all old2new, I will fill the remaining IterDomains from old to new in order:

Got it! That's what I missed.

wujingyue:
AFAICT from this PR, when Framework asks nvFuser for a TensorView of shape (4,5,7) and stride order (0,2,1) and default contiguity (i.e. non-contiguous), they can end up getting `(4,5,7),(7,4,1)` as long as nvFuser likes it (e.g. for less storage or for an output being an input alias). Is my understanding correct? Maybe a question to @jjsjann123 : Is this behavior acceptable to Framework? (I know probably 99% of the time Framework will ask for a contiguous tensor, but this is still a valid use case)

test/test_allocation_domain.cpp
zasdfgbnm: Assume the tensor's base address is 0, and x is 1, then the slice `[:, 1, :]`'s base address will be 1. This is not aligned to 4, and you will get a "misaligned address" error if you try to vectorize 4.

zasdfgbnm:
> Is my understanding correct?

I think it is correct. And the reason for liking it is not necessarily for aliasing or for saving storage, it can be just because doing so is faster.

For example, if the tensor has shape (10240000000, 5, 7) and stride (7, 4, 1), the best way to do the computation is of course to identify that the tensor is repeating itself and avoid duplicate computation. However, because this case is just so rare, we do not want to implement a overlapping analysis. But on the other hand, a quick and simple way to support this better without overlapping analysis is to just mark dim 0 as contiguous with dim 2, so that we can vectorize it, which will be much faster than unvectotized load.

test/test_allocation_domain.cpp
wujingyue: For posterity, can you add a comment explaining `stride=[1]` has to be a multiple of 4 for alignment?

wujingyue (approved):
I'd wait for Jie's approval too. Also added @naoyam in case he has an opinion.

jjsjann123:
I'm fine with us having the example in the test.

Just want to point it out that, overlapping tensor is a grey topic in framework. I think for the context of nvfuser where in-place update isn't a "real" feature and we don't promise anything on race condition, we could safely assume reading from them being safe. But on framework, it is not. Thinking about an in-place update on this tensor. It's going to be implementation-specific and could be non-deterministic.

> when Framework asks nvFuser for a TensorView of shape (4,5,7) and stride order (0,2,1) 

I'm not sure this is something we want to support? overlapping tensor on output is ill-defined. I guess we can ignore the overlapping part and treat it as-if it's just an ordinary tensor.

Similarly, I'm cool with @zasdfgbnm 's argument here that, we should relax our checks and allow the order of allocation domain to diverge from the real order, as long as it's compliant to the contiguity flags.

test/test_allocation_domain.cpp
```diff
@@ -1195,4 +1195,46 @@ TEST_F(NVFuserTest, AllocationDomainContiguityForExplicitBroadcast) {
...
+  auto tv1 = set(tv0);
+  auto tv2 = set(tv1);
+  fusion.addOutput(tv2);
+
+  // According to the stride order below, the allocation domain should be the
+  // same as the root domain. However, here we intentionally make the allocation
+  // domain [axis(1), axis(0), axis(2)] because doing so allows us to vectorize
+  // by 4.
+  tv0->setAllocationDomain(
+      {tv0->axis(1), tv0->axis(0), tv0->axis(2)}, {false, true, true});
```
jjsjann123, line 1214: Also want to point out that, there's not plan for us to plumb logics like this in integration. Interleaved/overlapped tensor is such a exotic topic. :)

jjsjann123 (approved):
stamping since I'm not against having test coverage for corner cases.

test/test_allocation_domain.cpp
zasdfgbnm: I added a comment before the test describing the situation.

PR #1201, opened 2023-11-01
Reorder compare codegen

jacobhinkle:
I previously put the order backwards so that the left-hand (should be old) commit was actually the new commit. That means that currently we show improvements in red and regressions in blue or green, for example. This change just reverses those to have more sane behavior.


tools/codediff/templates/command_env_info.html
```diff
@@ -11,7 +11,6 @@
                 {% if run1.gpu_names != run2.gpu_names %}{{ run1.name }}{% endif %}
                 GPUs:
                 <pre><code>{{run1.gpu_names | e}}</code></pre>
-                %}
```
jacobhinkle, line 14: Unrelated, but this cleans up the extra marker after GPU list

xwang233 (approved):

PR #1203, opened 2023-11-01
Generate an empty kernel when all outputs are aliases.

wujingyue:
Summary of changes: 
1. Made `Fusion::isNoOp` return true when all outputs are aliases. When `Fusion::isNoOp` returns true, device lowering emits no expressions, leading to an empty kernel. 
2. Made `AliasAnalysisResult` a class with `add` and `findRoot`. 
3. Added a pre-segmentation pass (`MarkAliasPass`) to call `aliasOutputToInput` for aliases detected by `AliasAnalysis`. 
4. Added unit tests.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -184,6 +184,7 @@ list(APPEND NVFUSER_SRCS
   ${NVFUSER_SRCS_DIR}/optimization/add_axioms.cpp
   ${NVFUSER_SRCS_DIR}/optimization/alias_analysis.cpp
   ${NVFUSER_SRCS_DIR}/optimization/consecutive_cast.cpp
+  ${NVFUSER_SRCS_DIR}/optimization/mark_alias.cpp
   ${NVFUSER_SRCS_DIR}/optimization/pre_segmenter.cpp
   ${NVFUSER_SRCS_DIR}/optimization/remove_empty.cpp
 )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -399,6 +399,7 @@ if(BUILD_TEST)
     ${NVFUSER_ROOT}/test/test_tutorial.cpp
     ${NVFUSER_ROOT}/test/test_alias_analysis.cpp
     ${NVFUSER_ROOT}/test/test_scalar_hoisting.cpp
+    ${NVFUSER_ROOT}/test/test_no_op.cpp
   )
 
   if(BUILD_PYTHON)
--- a/csrc/fusion.cpp
+++ b/csrc/fusion.cpp
@@ -313,10 +313,33 @@ std::vector<Expr*> Fusion::exprs() {
   return StmtSort::getExprs(this);
 }
 
+namespace {
+
+bool allOutputsArePointerCasts(Fusion* fusion) {
+  for (Val* out : fusion->outputs()) {
+    const auto& [in, info] = fusion->getOutputAlias(out);
+    if (in == nullptr) {
+      return false;
+    }
+    NVF_ERROR(info != nullptr);
+    if (info->type != AliasType::PointerCast) {
+      return false;
+    }
+  }
+  return true;
+}
+
+} // namespace
+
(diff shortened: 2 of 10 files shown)


csrc/fusion.cpp
```diff
@@ -313,23 +313,45 @@ std::vector<Expr*> Fusion::exprs() {
...
-        size_zero = true;
-        break;
-      }
-    }
+    const std::vector<IterDomain*>& root_dom =
+        TensorDomain::noReductions(out_tv->getMaybeRFactorDomain());
+    const bool size_zero =
+        std::any_of(root_dom.begin(), root_dom.end(), [](IterDomain* id) {
+          return id->extent()->isConstScalar() && id->extent()->evaluate() == 0;
+        });
```
wujingyue, line 349: Just a refactor to avoid using a mutable variable and `break`.

wujingyue:
!build

CMakeLists.txt
```diff
@@ -399,6 +400,7 @@ if(BUILD_TEST)
     ${NVFUSER_ROOT}/test/test_tutorial.cpp
     ${NVFUSER_ROOT}/test/test_alias_analysis.cpp
     ${NVFUSER_ROOT}/test/test_scalar_hoisting.cpp
+    ${NVFUSER_ROOT}/test/test_no_op.cpp
```
jjsjann123, line 403: lol. did we forget to build the test the whole time?!

csrc/optimization/alias_analysis.cpp
```diff
@@ -104,25 +104,54 @@ void findAliasesFromExpr(Expr* expr, AliasAnalysisResult& alias_to_source) {
...
+  const TensorView*& old_source = alias_to_source_[alias];
+  NVF_ERROR(
+      old_source == nullptr,
+      "The current implementation of alias analysis shouldn't find two sources for an alias. However, it's trying to make ",
+      alias->toString(),
+      " an alias of ",
+      source->toString(),
+      " while it's already an alias of ",
+      old_source->toString());
+  old_source = source;
```
jjsjann123, line 126: nitpick, should we instead us `old_source = findRoot(source)`?

jjsjann123: I only suggested that since the existing AliasAnalysisResult doesn't have any way to mutate existing alias. Would understand if we are going to expand that in the future and want to store the primitive raw alias relationship instead.

csrc/optimization/mark_alias.h
```diff
@@ -0,0 +1,20 @@
+// clang-format off
```
jjsjann123, line 1: nitpick. Leaning towards merge `mark_alias` with `alias_analysis`.

jjsjann123 (approved):
LGTM. some minor nitpick.

CMakeLists.txt
wujingyue: Fortunately, not the whole time. My recent #1205 missed that...

csrc/optimization/mark_alias.h
wujingyue: Merge two files or merge the two classes? To the former, I general prefer creating one file per class to avoid unnecessary build dependency. To the latter, I want to separate analysis (which doesn't mutate a fusion) and passes (which mutate a fusion) and I do test the analysis separately.

csrc/optimization/mark_alias.cpp
```diff
@@ -0,0 +1,29 @@
...
+ */
+// clang-format on
+#include <ir/utils.h>
+#include <optimization/alias_analysis.h>
+#include <optimization/mark_alias.h>
+
+namespace nvfuser::optimization {
+
+void MarkAliasPass::runPass(Fusion* fusion) {
+  fusion->print();
```
naoyam, line 15: Leftover debug print?

wujingyue: Sorry... #1219

PR #1206, opened 2023-11-01
Minor cleanups to the Python API for normalization and its test.

wujingyue:


nvfuser/contrib/nn/normalization.py
```diff
@@ -113,14 +113,11 @@ def norm_fusion_forward(
     is_spatial_dim = [True] * num_dims
     is_spatial_or_batch_dim = [True] * num_dims
 
-    num_stats = fd.define_scalar(1)
```
jjsjann123, line 116: wait. this doesn't look right....

why are we not using num_stats?

jjsjann123: wasn't num_stats used when momentum is missing?!?! Looks like a functional mismatch here.

jjsjann123: https://github.com/pytorch/pytorch/blob/8191fb3e060f600090c839d6b86c28dbb687b679/torch/nn/modules/batchnorm.py#L157

jjsjann123 (approved):
not saying that it really matters.
Looking at the code, I guess our kernel don't even support `None` for momentum. So we don't have functional parity in the first place and this PR isn't making it worse. Stamping.

nvfuser/contrib/nn/normalization.py
wujingyue: I've no idea... I attempted to test `fd.ops.batch_norm` and realized the logic here isn't the easiest to follow :) I may give up and create a new test.

wujingyue:
!build

PR #1210, opened 2023-11-02
Set the allocation domain of mma output as leaf domain and remove `accumulator_stride`

zasdfgbnm:
Matmul support was initially developed before we invented the concept "allocation domain". Without the concept of "allocation domain", our codegen has lots of limitations. One limitation is, the allocation is always based on the order of the rFactor domain, even for tensors allocated on shared memory or local memory. For example, if I have a tenor `[M, N]`, and I do some splits and reorder to get `[BIDx{M/2}, TIDx{N/3/5}, 3, 2, 5]` and the memory layout of the register will always be `[2, 3, 5]`, regardless of how we reorder these three numbers in leaf domain.

Due to this limitation, scheduling the layout of MMA output was fundamentally impossible. For example, in the [memory layout for Volta MMA](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#matrix-fragments-for-mma-m8n8k4-with-f16-floating-point-type):

the memory layout of register should be:
```
N --(splits)--> [N/8, 2*, 2**, 2***]
M --(splits)--> [M/4, 2', 2'']
register layout: [2*, 2', 2***]
```
which has interleaved M and N.

The workaround? If you look at the code in `tensorcore.cu`, each MMA op has a template argument `acc_stride`. For example, in this function:
```CUDA
template <int acc_stride>
DEVICE_INLINE void initM16N8K16TN(Array<float, 4, 4>* accumulator) {
  float* _C = reinterpret_cast<float*>(accumulator);
  _C[0] = 0;
  _C[1] = 0;
  _C[acc_stride] = 0;
  _C[acc_stride + 1] = 0;
}
```
we accept an array[4] as a parameter, however, we are intentionally doing out of boundary access inside this function, instead of using the four items passed as arguments. This approach in practice works, but is definitely not the correct way of doing engineering.

In the original design, there would be a "transpose swizzle" which takes care of this, but that never got implemented. Today, we have a better solution: we can just set the allocation domain to the desired order we want, and the `accumulator` passed in will automatically contain the elements we want. The `accumulator_stride` hack is no longer needed.

The major change of this PR is actually just three lines, see the change in `csrc/tensor_view.cpp::TensorView::applyMmaSwizzle`, the rest of this PR are just clean ups. These three lines just set the allocation domain to leaf domain, which has the correct memory layout.

cc: @jjsjann123 @wujingyue who may be interested in this PR because this is the first time that allocation domain get used in a scheduler to do some serious work. Feel free to ask questions or leave comments.
cc: @naoyam as always

### DIFF
--- a/csrc/index_compute.cpp
+++ b/csrc/index_compute.cpp
@@ -57,9 +57,6 @@ int getProducerHaloOffset(
   // producer tensors, where reduction axes are skipped, producer_id
   // should never be a reduction axis.
   if (it == p2c.end()) {
-    TORCH_WARN(
-        "getProducerHaloOffset p2c mapping has failed. See "
-        "https://github.com/NVIDIA/Fuser/issues/1122");
     return 0;
   }
   IterDomain* consumer_id = it->second;
--- a/csrc/tensor_view.cpp
+++ b/csrc/tensor_view.cpp
@@ -1379,7 +1379,9 @@ void TensorView::applyMmaSwizzle(MmaOptions options) {
   switch (options.operand) {
     case MmaOptions::Operand::Accumulator:
       mma_utils::WarpMmaSwizzler::scheduleMmaWarpOutput(this, options);
-      setAllocationDomain(getLeafDomain(), true);
+      if (definition()->isA<MmaOp>()) {
+        setAllocationDomain(getLeafDomain(), true);
+      }
       break;
     case MmaOptions::Operand::A:
     case MmaOptions::Operand::B:
--- a/csrc/tensor_view.cpp
+++ b/csrc/tensor_view.cpp
@@ -1379,6 +1379,7 @@ void TensorView::applyMmaSwizzle(MmaOptions options) {
   switch (options.operand) {
     case MmaOptions::Operand::Accumulator:
       mma_utils::WarpMmaSwizzler::scheduleMmaWarpOutput(this, options);
+      setAllocationDomain(getLeafDomain(), true);
       break;
     case MmaOptions::Operand::A:
     case MmaOptions::Operand::B:
--- a/csrc/tensor_view.cpp
+++ b/csrc/tensor_view.cpp
@@ -928,7 +928,7 @@ TensorView* TensorView::rFactor(const std::vector<int>& axes) {
         this_mma->inA(),
         this_mma->inB(),
         this_mma->init(),
-        this_mma->options(),
+        this_mma->macro(),
         this_mma->layout());
 
     // Remaining reduction that can be scheduled cross
--- a/csrc/codegen.cpp
+++ b/csrc/codegen.cpp
@@ -1094,10 +1094,6 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
           "via ldmatrix.trans for fp16 or explicitly for other types.");
     }
     ss << toString(mma_layout_opt.value());
-    // TODO: additional parameter could be removed by swizzling iterdomain
-    auto acc_stride = mma->accStride();
-    NVF_ERROR(acc_stride > 0);
-    ss << "<" << acc_stride << ">";
     return ss.str();
   }
 
--- a/csrc/codegen.cpp
+++ b/csrc/codegen.cpp
@@ -1076,18 +1076,18 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
 
   std::string genMmaOp(const MmaOp* mma, bool init = false) {
     std::stringstream ss;
-    auto options = mma->options();
-    ss << genArchString(options.macro) << "::";
+    auto macro = mma->macro();
+    ss << genArchString(macro) << "::";
     if (init) {
       ss << "init";
     }
-    ss << toString(options.macro);
+    ss << toString(macro);
 
     // clang-tidy: bugprone-unchecked-optional-access
     // clang-tidy assumes that function result is unstable, so we need a copy.
     auto mma_layout_opt = mma->layout();
     NVF_ERROR(mma_layout_opt.has_value(), "mma unknown input layout");
-    if (isTuring(options.macro) || isAmpere(options.macro)) {
+    if (isTuring(macro) || isAmpere(macro)) {
       NVF_ERROR(
           mma_layout_opt == MmaOptions::MmaLayout::TN,
           "MMAs in Turing and Ampere are TN only, transpose is handled either "
@@ -1099,43 +1099,43 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
 
   void genMmaOperands(const MmaOp* mma) {
     std::stringstream ss;
-    auto options = mma->options();
+    auto macro = mma->macro();
     auto in_a = mma->inA()->as<kir::TensorIndex>()->view();
     auto dtype = in_a->getDataType().value();
     indent() << kTab << "&(reinterpret_cast<Array<" << dtype << ","
-             << getInputARegisterSize(options.macro) << ","
-             << getInputARegisterSize(options.macro) << ">*>(&"
+             << getInputARegisterSize(macro) << ","
+             << getInputARegisterSize(macro) << ">*>(&"
              << genVariableName(mma->inA()->as<kir::TensorIndex>()->view())
              << ")[" << genInline(mma->inA()->as<kir::TensorIndex>()->index())
              << "])"
              << ",\n";
     indent() << kTab << "&(reinterpret_cast<Array<" << dtype << ","
-             << getInputBRegisterSize(options.macro) << ","
-             << getInputBRegisterSize(options.macro) << ">*>(&"
+             << getInputBRegisterSize(macro) << ","
+             << getInputBRegisterSize(macro) << ">*>(&"
              << genVariableName(mma->inB()->as<kir::TensorIndex>()->view())
              << ")[" << genInline(mma->inB()->as<kir::TensorIndex>()->index())
              << "])";
   }
 
   void genMmaInitialization(const MmaOp* mma, const LoadStoreOp* ldst) {
-    auto options = mma->options();
+    auto macro = mma->macro();
 
     indent() << genMmaOp(mma, true) << "(reinterpret_cast<Array<"
              << mma->out()->getDataType().value() << ","
-             << getOutputRegisterSize(options.macro) << ","
-             << getOutputRegisterSize(options.macro) << ">*>"
+             << getOutputRegisterSize(macro) << ","
+             << getOutputRegisterSize(macro) << ">*>"
              << "(&" << gen(ldst->out()) << "));\n";
   }
 
   void handle(const MmaOp* mma) final {
-    auto options = mma->options();
+    auto macro = mma->macro();
     auto out = mma->out()->as<kir::TensorIndex>();
     indent() << genMmaOp(mma) << "(\n";
     indent() << kTab << "reinterpret_cast<Array<"
              << out->view()->getDataType().value() << ","
-             << getOutputRegisterSize(options.macro) << ","
-             << getOutputRegisterSize(options.macro) << ">*>(&"
-             << gen(mma->out()) << "),\n";
+             << getOutputRegisterSize(macro) << ","
+             << getOutputRegisterSize(macro) << ">*>(&" << gen(mma->out())
+             << "),\n";
     genMmaOperands(mma);
     code_ << ");\n";
   }
--- a/csrc/codegen.cpp
+++ b/csrc/codegen.cpp
@@ -1093,7 +1093,9 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
           "MMAs in Turing and Ampere are TN only, transpose is handled either "
           "via ldmatrix.trans for fp16 or explicitly for other types.");
     }
-    ss << toString(mma_layout_opt.value());
+    if (!init) {
+      ss << toString(mma_layout_opt.value());
+    }
     return ss.str();
   }
 
--- a/csrc/device_lower/pass/index.cpp
+++ b/csrc/device_lower/pass/index.cpp
@@ -1377,7 +1377,7 @@ void IndexLowering::handle(const MmaOp* mma) {
   const auto b = lowerSrcIndex(mma->inB(), mma->out());
   const auto out = lowerDstIndex(mma->out());
   auto mma_indexed = IrBuilder::create<MmaOp>(
-      out, a, b, mma->init(), mma->options(), mma->layout());
+      out, a, b, mma->init(), mma->macro(), mma->layout());
   pushBack(mma_indexed);
   GpuLower::current()->propagateExprInfo(mma, back());
 }
--- a/csrc/device_lower/utils.cpp
+++ b/csrc/device_lower/utils.cpp
@@ -605,7 +605,7 @@ class ReplaceExprInput : private kir::ExprMutator {
           replaced_inputs->at(node->inA()),
           replaced_inputs->at(node->inB()),
           node->init(),
-          node->options(),
+          node->macro(),
           node->layout());
       registerReplaceWithPredicate(node, replacement);
     }
--- a/csrc/device_lower/validation.cpp
+++ b/csrc/device_lower/validation.cpp
@@ -1066,7 +1066,7 @@ void validateMma(Fusion* fusion) {
     if (auto mma = dynamic_cast<MmaOp*>(expr)) {
       validateMmaTensors(mma);
 
-      switch (mma->options().macro) {
+      switch (mma->macro()) {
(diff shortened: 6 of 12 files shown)


zasdfgbnm:
!build

csrc/tensor_view.cpp
```diff
@@ -1379,6 +1379,9 @@ void TensorView::applyMmaSwizzle(MmaOptions options) {
   switch (options.operand) {
     case MmaOptions::Operand::Accumulator:
       mma_utils::WarpMmaSwizzler::scheduleMmaWarpOutput(this, options);
+      if (definition()->isA<MmaOp>()) {
+        setAllocationDomain(getLeafDomain(), true);
```
jjsjann123, line 1383: :partying_face:

csrc/index_compute.cpp
```diff
@@ -57,9 +57,6 @@ int getProducerHaloOffset(
   // producer tensors, where reduction axes are skipped, producer_id
   // should never be a reduction axis.
   if (it == p2c.end()) {
-    TORCH_WARN(
-        "getProducerHaloOffset p2c mapping has failed. See "
-        "https://github.com/NVIDIA/Fuser/issues/1122");
```
jacobhinkle, line 62: Is this unrelated?

jacobhinkle (commented):
This is great.

wujingyue:
> I do some splits and reorder to get `[BIDx{M/2}, TIDx{N/3/5}, 3, 2, 5]` and the memory layout of the register will always be `[2, 3, 5]`, regardless of how we reorder these three numbers in leaf domain.

`[BIDx{M/2}, TIDx{N/3/5}, 3, 2, 5]` is the rfactor domain or leaf? I only saw in our tests splits and reorder used to generate leaf domains. But I guess they are more general? 

Is `[2, 3, 5]` a typo? Do you mean `[3, 2, 5]` (the same order as in `[BIDx{M/2}, TIDx{N/3/5}, 3, 2, 5]`)?

jjsjann123:
> > I do some splits and reorder to get `[BIDx{M/2}, TIDx{N/3/5}, 3, 2, 5]` and the memory layout of the register will always be `[2, 3, 5]`, regardless of how we reorder these three numbers in leaf domain.
> 
> `[BIDx{M/2}, TIDx{N/3/5}, 3, 2, 5]` is the rfactor domain or leaf? I only saw in our tests splits and reorder used to generate leaf domains. But I guess they are more general?
> 
> Is `[2, 3, 5]` a typo? Do you mean `[3, 2, 5]` (the same order as in `[BIDx{M/2}, TIDx{N/3/5}, 3, 2, 5]`)?

I think `[BIDx{M/2}, TIDx{N/3/5}, 3, 2, 5]` is marking the leaf domain. In that case, layout of register would be in the original split order, which is `[2, 5, 3]`, but this is contradicting the `[2, 3, 5]` order from Xiang's description.

i.e. `M, N`
-> split preserves the original layout `M/2, 2, ((N/3)/5), 5, 3`
-> reorder to `M/2, ((N/3)/5), 3, 2, 5]`, but does not change the layout.
So the layout should be in the order after split, which is `2, 5, 3`

csrc/index_compute.cpp
naoyam: I suppose yes. @zasdfgbnm can you please leave a comment?

wujingyue:
> we are intentionally doing out of boundary access inside this function, instead of using the four items passed as arguments.

Lol, beautiful! Great motivation for future allocation domain improvements.

wujingyue:
> Lol, beautiful! Great motivation for future allocation domain improvements.

I misunderstood. This PR gets rid of the acc_stride hack already.

csrc/tensor_view.cpp
wujingyue: Is this still non ideal? We shouldn't rely on leaf domain to dictate register layout, should we? My mental model for a leaf domain has been a loop nest. We shouldn't need to "loop index" into a register.

naoyam: There's nothing wrong to make the allocation layout the same as the loop structure. The leaf domain of a tensor does dictate its loop nest. In this case, it happens to be that its allocation layout is the same as the loop structure.

naoyam:
> > > I do some splits and reorder to get `[BIDx{M/2}, TIDx{N/3/5}, 3, 2, 5]` and the memory layout of the register will always be `[2, 3, 5]`, regardless of how we reorder these three numbers in leaf domain.
> > 
> > 
> > `[BIDx{M/2}, TIDx{N/3/5}, 3, 2, 5]` is the rfactor domain or leaf? I only saw in our tests splits and reorder used to generate leaf domains. But I guess they are more general?
> > Is `[2, 3, 5]` a typo? Do you mean `[3, 2, 5]` (the same order as in `[BIDx{M/2}, TIDx{N/3/5}, 3, 2, 5]`)?
> 
> I think `[BIDx{M/2}, TIDx{N/3/5}, 3, 2, 5]` is marking the leaf domain. In that case, layout of register would be in the original split order, which is `[2, 5, 3]`, but this is contradicting the `[2, 3, 5]` order from Xiang's description.
> 
> i.e. `M, N` -> split preserves the original layout `M/2, 2, ((N/3)/5), 5, 3` -> reorder to `M/2, ((N/3)/5), 3, 2, 5]`, but does not change the layout. So the layout should be in the order after split, which is `2, 5, 3`

I think @jjsjann123 is right. The register buffer itself is allocated based on the leaf domain, so the buffer size is calculated as `3 * 2 * 5`. Indexing is done with the allocation domain, so before this PR, since the allocation domain is `[M, N]`, we generate index expressions for both `M` and `N`. For the domain of `M`, since `M/2` is parallelized, only the `2` part is relevant. Similarly, for `N`, only `3` and `5` are relevant. Since `N` is first split by `3` and then `5`, the domain of `3` is inner than `5`. So the overall index should consist of `[2, 5, 3]`.

naoyam (approved):
LGTM.

csrc/tensor_view.cpp
wujingyue: I understood there's nothing wrong. If we look at the MMA instruction alone (`T9_l` in the following example), its leaf domain has the part `iMMA53{( ceilDiv(8, 4) )}, iMMA49{( ceilDiv(4, 2) )}, iMMA56{2}` that dictates how `T9_l` should index **into** the MMA register. If that part were in a different order (e.g. `iMMA56, iMMA53, iMMA49`) for whatever reasons, I believe that would be an invalid fusion IR. This leads me to think that, ideally, an MMA instruction should collapse **the part of** the leaf domain that dictates how it loop-indexes **into** the 2x2x2 MMA register, because the MMA instruction produces the register as a whole and can't control the loop nest **inside** the register. 

```
[ RUN      ] NVFuserTest.FusionVoltaMMATN_CUDA

%kernel {
T2_l[ iS4{16}, bS5{1}, iS6{4} ]
   = broadcast( T0_g[ iS0{16}, iS1{4} ] )
T5_s[ iS13{16}, bS14{1}, iS15{4} ]
   = Set( T2_l[ iS4{16}, bS5{1}, iS6{4} ], cache_op=Streaming )
T6_l[ ithreadIdx.x35{( ( ( ( ceilDiv(16, 4) ) * ( ceilDiv(16, 8) ) ) * ( ceilDiv(1, 16) ) ) * 4 )}, bS32{8}, iS18{4} ]
   = Set( T5_s[ iS13{16}, bS14{1}, iS15{4} ], cache_op=Streaming )
T3_l[ bS7{1}, iS8{16}, iS9{4} ]
   = broadcast( T1_g[ iS2{16}, iS3{4} ] )
T7_s[ bS19{1}, iS20{16}, iS21{4} ]
   = Set( T3_l[ bS7{1}, iS8{16}, iS9{4} ], cache_op=Streaming )
T8_l[ bS36{( ceilDiv(1, 16) )}, bS39{8}, ithreadIdx.x46{( ( ( ( ceilDiv(8, 4) ) * ( ceilDiv(16, 8) ) ) * ( ceilDiv(16, 8) ) ) * 4 )}, iS24{4} ]
   = Set( T7_s[ bS19{1}, iS20{16}, iS21{4} ], cache_op=Streaming )
T9_l[ ithreadIdx.x59{( ( ( ( ceilDiv(16, 4) ) * ( ceilDiv(16, 8) ) ) * ( ceilDiv(4, 2) ) ) * 2 )}, iMMA53{( ceilDiv(8, 4) )}, iMMA49{( ceilDiv(4, 2) )}, iMMA56{2}, rMMA12{4} ]
   = mma(T6_l[ ithreadIdx.x35{( ( ( ( ceilDiv(16, 4) ) * ( ceilDiv(16, 8) ) ) * ( ceilDiv(1, 16) ) ) * 4 )}, bS32{8}, iS18{4} ],
         T8_l[ bS36{( ceilDiv(1, 16) )}, bS39{8}, ithreadIdx.x46{( ( ( ( ceilDiv(8, 4) ) * ( ceilDiv(16, 8) ) ) * ( ceilDiv(16, 8) ) ) * 4 )}, iS24{4} ])
T4_g[ ithreadIdx.x72{( ( ( ( ceilDiv(16, 4) ) * ( ceilDiv(16, 8) ) ) * ( ceilDiv(4, 2) ) ) * 2 )}, iS66{( ceilDiv(8, 4) )}, iS62{( ceilDiv(4, 2) )}, iS69{2} ]
   = Set( T9_l[ ithreadIdx.x59{( ( ( ( ceilDiv(16, 4) ) * ( ceilDiv(16, 8) ) ) * ( ceilDiv(4, 2) ) ) * 2 )}, iMMA53{( ceilDiv(8, 4) )}, iMMA49{( ceilDiv(4, 2) )}, iMMA56{2}, rMMA12{4} ], cache_op=Streaming )
```

```
T9_l[ ithreadIdx.x59{( ( ( ( ceilDiv(16, 4) ) * ( ceilDiv(16, 8) ) ) * ( ceilDiv(4, 2) ) ) * 2 )}, iMMA53{( ceilDiv(8, 4) )}, iMMA49{( ceilDiv(4, 2) )}, iMMA56{2}, rMMA12{4} ]
 root domain : (iS10{16}, iS11{16}, rMMA12{4})
 allocation domain : (ithreadIdx.x59{( ( ( ( ceilDiv(16, 4) ) * ( ceilDiv(16, 8) ) ) * ( ceilDiv(4, 2) ) ) * 2 )}, iMMA53{( ceilDiv(8, 4) )}, iMMA49{( ceilDiv(4, 2) )}, iMMA56{2}, rMMA12{4})
 contiguity: t t t t n
  Split: iS10{16} by factor 4 -> iS47{( ceilDiv(16, 4) )}, iS48{4}, start offset: 0, stop offset: 0
  Split: iS11{16} by factor 8 -> iS51{( ceilDiv(16, 8) )}, iS52{8}, start offset: 0, stop offset: 0
  Merge: iS47{( ceilDiv(16, 4) )} and iS51{( ceilDiv(16, 8) )} -> iS57{( ( ceilDiv(16, 4) ) * ( ceilDiv(16, 8) ) )}
  Split: iS52{8} by factor 4 -> iMMA53{( ceilDiv(8, 4) )}, iS54{4}, start offset: 0, stop offset: 0
  Split: iS54{4} by factor 2 -> iS55{( ceilDiv(4, 2) )}, iMMA56{2}, start offset: 0, stop offset: 0
  Merge: iS57{( ( ceilDiv(16, 4) ) * ( ceilDiv(16, 8) ) )} and iS55{( ceilDiv(4, 2) )} -> iS58{( ( ( ceilDiv(16, 4) ) * ( ceilDiv(16, 8) ) ) * ( ceilDiv(4, 2) ) )}
  Split: iS48{4} by factor 2 -> iMMA49{( ceilDiv(4, 2) )}, iS50{2}, start offset: 0, stop offset: 0
  Merge: iS58{( ( ( ceilDiv(16, 4) ) * ( ceilDiv(16, 8) ) ) * ( ceilDiv(4, 2) ) )} and iS50{2} -> ithreadIdx.x59{( ( ( ( ceilDiv(16, 4) ) * ( ceilDiv(16, 8) ) ) * ( ceilDiv(4, 2) ) ) * 2 )}
 leaf domain : (ithreadIdx.x59{( ( ( ( ceilDiv(16, 4) ) * ( ceilDiv(16, 8) ) ) * ( ceilDiv(4, 2) ) ) * 2 )}, iMMA53{( ceilDiv(8, 4) )}, iMMA49{( ceilDiv(4, 2) )}, iMMA56{2}, rMMA12{4})
```

zasdfgbnm: @wujingyue Are you talking about merging `iMMA53{( ceilDiv(8, 4) )}, iMMA49{( ceilDiv(4, 2) )}, iMMA56{2}` into a single ID during schedule? Conceptually this does make more sense to me, but in practice, this makes no difference but, because loops parallelized as `MMA` is by definition "trivial", therefore will never have index.

zasdfgbnm: > We shouldn't rely on leaf domain to dictate register layout, should we?

I think it is the opposite. In general, we should use leaf domain as leaf domain as much as possible. For example, if I have
```
T1 = T0 (input);
T2 (output) = T1;
```
Assume that the leaf domain for `T1` and `T2` are both `[BIDx, TIDx, 8]`, then when indexing `T1` as a consumer, because the leaf domain is already allocation domain, so the indexing is trivial: the index is just the loop variable for the dimension of size 8. If instead, we use rFactor domain as allocation, we need to traverse the transformation from rFactor to leaf to back propagate the index. When indexing `T1` as a producer, we need to replay `T1` as `T2`, because the leaf domain of `T2` is already identical to the allocation domain of `T1`, we can then just trivially use the loop index again as the index of `T1`.

csrc/index_compute.cpp
zasdfgbnm: This warning is issued in every matmul kernel, which is very annoying. The fundamental problem for this is because our halo analysis assumes we are indexing into rFactor domain, and when rFactor domain mismatch with allocation domain, the halo analysis will just fail. However, halo is not something we prioritize right now, so I am just doing whatever easiest hack to unblock, as long as in practice the code works. We need to rewrite indexing and abandon this code anyway.

zasdfgbnm:
@jjsjann123 and @naoyam Thanks for the explanation! Yes, what you described matches with my understanding.

I will merge this PR as-is to unblock my future work. Feel free to continue discussion.

csrc/tensor_view.cpp
wujingyue: I agree with you, but one key question remains not answered: I understand `T9_l`'s leaf domain is currently `ithreadIdx.x59, iMMA53, iMMA49, iMMA56, rMMA12`. If it were instead `ithreadIdx.x59, iMMA56, iMMA53, iMMA49, rMMA12` instead, should codegen (1) error out or (2) produce the same kernel or (2) produce a functional but slower kernel? Note that I'm asking what it **should** do (the contract), not what it *does* (the implementation). Understanding the contract helps me program with the right mental model.

wujingyue: A side question: `rMMA12{4}` in `T9_l`'s root/allocation/leaf domain means a reduction? I don't see reduction ops around it. Does it bear a special meaning?

zasdfgbnm: > codegen (1) error out or (2) produce the same kernel or (2) produce a functional but slower kernel?

Ideally, it should error out, but unfortunately there is no good way to reliablely detect errors without duplicating the implementation another time. Generally, I would like to consider our lowering an a programming language like C++, and our schedulers as some automated C++ code generator. In this mindset, then the mma schedules are inline assembly `asm` to C++. Ideally, C++ compilers should check the correctness of inlined assmebly, but in practice, it is not checked. Our hardware just definitely this special weird memory format for mma, and we should schedule consistent with it. If not, the behavior is undefined. It can be silent wrong result, it can also be misaligned address access.

zasdfgbnm: > A side question: `rMMA12{4}` in `T9_l`'s root/allocation/leaf domain means a reduction? I don't see reduction ops around it. Does it bear a special meaning?

Yes, it is a reduction. Mma operators is a fused mul + sum.

csrc/index_compute.cpp
naoyam: Yes, it'll be rewritten, and for that I'd like to keep track of what's working and what isn't. A code comment like you just mentioned would be greatly helpful for when we would revisit here.

zasdfgbnm: Oh I see what you mean. I will include some description in my next PR.

zasdfgbnm: description added in https://github.com/NVIDIA/Fuser/pull/1176

PR #1211, opened 2023-11-02
adding FusionDefinition and FusionDefinition to the stub file

crcrpar:


jacobhinkle:
Could we possibly generate these automatically during build using https://github.com/sizmailov/pybind11-stubgen?

PR #1214, opened 2023-11-02
Standardize mangled kernel name in ptx

jacobhinkle:
This should reduce false positives like those seen in http://nv/e5M/nvfuser_github_ci/codegen_diff_p10551865_j73238023_1698902733422536803_codediff_f450ece_ed879b8_custom_command_20231101_213723.html associated with #1176.

### DIFF
--- a/tools/codediff/diff_report.py
+++ b/tools/codediff/diff_report.py
@@ -633,6 +633,23 @@ class TestDiff:
     kernel_diffs: list[KernelDiff] | None = None
 
 
+def sanitize_ptx_lines(lines: list[str]) -> list[str]:
+    """Remove comments and translate kernel38 to kernelN"""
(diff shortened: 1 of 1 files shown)


xwang233 (approved):
Should we use some kind of C++ demangler instead of changing from one mangled format to another? I know linux has `c++filt` tool. Maybe there are some python libraries can do this?

jacobhinkle:
> Should we use some kind of C++ demangler instead of changing from one mangled format to another? I know linux has `c++filt` tool. Maybe there are some python libraries can do this?

Interesting idea. I did a quick search and it seems that the options available in python currently just do string->string translation instead of converting to some sort of structured format. Using something like that we'd still need to do a regex on the resulting string, and I think staying as close to the original as possible is desirable.

PR #1217, opened 2023-11-03
Add simple benchmark plot linking to test

jacobhinkle:
I looked up d3.js last night and figured out how to automatically plot benchmark runtimes. Here is a truncated example from a comparison I ran yesterday: 
[diff.html.gz](https://github.com/NVIDIA/Fuser/files/13251353/diff.html.gz)

This also silences some very verbose output that snuck in from #1214.

### DIFF
--- a/tools/codediff/diff_report.py
+++ b/tools/codediff/diff_report.py
@@ -646,7 +646,6 @@ def sanitize_ptx_lines(lines: list[str]) -> list[str]:
         # Remove comments. This is important for
         l = re.sub(r"//.*$", "", l)
         sanitary_lines.append(l)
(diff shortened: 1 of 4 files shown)


tools/codediff/diff_report.py
```diff
@@ -646,7 +646,6 @@ def sanitize_ptx_lines(lines: list[str]) -> list[str]:
         # Remove comments. This is important for
         l = re.sub(r"//.*$", "", l)
         sanitary_lines.append(l)
-        print("Sanitized:", l.rstrip())
```
jacobhinkle, line 649: Oops

xwang233 (approved):

PR #1220, opened 2023-11-03
Transpose benchmark

Priya2698:
Ports the `transpose` benchmark at: https://github.com/NVIDIA/Fuser/blob/main/benchmark/transpose.cpp
As compared to the CPP benchmark, the intermediate output is transposed instead of the input. 
The fusion definition when one of the inputs/output is transposed does not include the `transpose` op in the fusion due to optimizations.

### DIFF
--- a/python_benchmarks/test_transpose.py
+++ b/python_benchmarks/test_transpose.py
@@ -1,0 +1,78 @@
+import pytest
+from nvfuser import FusionDefinition, DataType
+from nvfuser.pytorch_utils import torch_dtype_to_nvfuser_dtype
+from .core import run_benchmark
+import torch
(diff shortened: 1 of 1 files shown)


naoyam:
@jjsjann123 Please review

jjsjann123 (approved):
LGTM

jjsjann123:
> the intermediate output is transposed instead of the input. The fusion definition when one of the inputs/output is transposed does not include the `transpose` op in the fusion due to optimizations.

I'm not following the part where you were saying that. inputs/outputs `transpose` isn't included in the fusion?!
I'm not aware that we have such optimization in the code.

PR #1218, opened 2023-11-03
Remove unused debug dump and enable options

jacobhinkle:
This changes cleans up options that apparently no longer have any effect. The following are removed:
- `DebugDumpOption::AssertMemoryViolation`
- `EnableOption::Complex`
- `EnableOption::ConvDecomposition`
- `EnableOption::GraphOp`
- `EnableOption::LinearDecomposition`

PR #1222, opened 2023-11-04
Fix a regression due to #1218

naoyam:
The TOT fails as:

```
  what():  available_options.size() == static_cast<int>(OptionEnum::EndOfOption) INTERNAL ASSERT FAILED at "/raid/nmaruyama/debug2/csrc/options.cpp":24, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. Invalid available option map
Exception raised from parseEnvOptions at /raid/nmaruyama/debug2/csrc/options.cpp:24 (most recent call first):
```

Unused options were removed in #1218. Those options need to be done with the structs as well.


jacobhinkle (approved):
Thanks for fixing my mess.

PR #973, opened 2023-09-28
Simplify extents in ops

jacobhinkle:
This PR uses `SimplifyingIrBuilder::maybeCastExpr` to simplify scalars in `maybeCastOp`, which is used in ops to simplify extents and other scalars at definition. This prevents introducing new cast expressions when those scalars are constant, instead just creating a new constant with the same value but a new dtype. This has a big effect when we use a constant `Int` as an `Index`, since those casts generally prevent some simplifications.

As an example, the incomplete fusion from #963 used to print like this
```
Inputs:                                                                                                                                  
  T0_g[ iS0{i0}, iS1{i2} ], float                                   
Outputs:                                                                                                                                                                                                                                                                          
%kernel_math {                                                                                                                           
T1_l[ iS2{( (nvfuser_index_t)(9) )}, iS3{( (nvfuser_index_t)(5) )} ]                                                                     
   = Set( T0_g[ iS0{i0}, iS1{i2} ], cache_op=Streaming )            
i7 = (nvfuser_index_t)(5);                                                                                                               
i16 = i7 + 2;                                                                                                                            
i22 = i7 + 3;                                                                                                                            
i35 = i7 - 2;                                                                                                                            
i24 = fmin(i7, 3);                                                                                                                       
i26 = fmax(2, i24);                                                                                                                      
i29 = -i7;                                                                                                                               
i31 = i26 + i29;
i37 = i35 + i31;                                                                                                                         
Resize: iS5{( (nvfuser_index_t)(5) )}rf by -2 and ( ( fmax(2, ( fmin(( (nvfuser_index_t)(5) ), 3) )) ) + ( -( (nvfuser_index_t)(5) ) ) ) 
-> bS6{( ( ( (nvfuser_index_t)(5) ) - 2 ) + ( ( fmax(2, ( fmin(( (nvfuser_index_t)(5) ), 3) )) ) + ( -( (nvfuser_index_t)(5) ) ) ) )}rf
i5 = (nvfuser_index_t)(9);                                                                                                               
T2_l[ iS4{( (nvfuser_index_t)(9) )}, bS6{( ( ( (nvfuser_index_t)(5) ) - 2 ) + ( ( fmax(2, ( fmin(( (nvfuser_index_t)(5) ), 3) )) ) + ( -(
 (nvfuser_index_t)(5) ) ) ) )}rf ]                                                                                                       
[...]

PR #1223, opened 2023-11-04
Evaluate extent when validating sum over broadcast dimension

### LINKED ISSUE #1221: Reducing a broadcast dimension created from a Pad fails while creating Fusion
Error:
```bash
RuntimeError: Cannot reduce an axis that is marked as broadcasted as it has an undetermined size. Tried to reduce ID = bS140{( ( ( (nvfuser_index_t)(2) ) - 1 ) + ( ( fmax(1, ( fmin(( (nvfuser_index_t)(2) ), 2) )) ) + ( -( (nvfuser_index_t)(2) ) ) ) )}rf of tensor T32_l[ ?S135{( (nvfuser_index_t)(12) )}rf, ?S136{( (nvfuser_index_t)(128) )}rf, ?S137{( (nvfuser_index_t)(25) )}rf, ?S138{( (nvfuser_index_t)(32) )}rf, bS140{( ( ( (nvfuser_index_t)(2) ) - 1 ) + ( ( fmax(1, ( fmin(( (nvfuser_index_t)(2) ), 2) )) ) + ( -( (nvfuser_index_t)(2) ) ) ) )}rf ]
Exception raised from newForReduction at /opt/pytorch/nvfuser/csrc/ops/arith.cpp:1089 (most recent call first):
```

Reproduction:
```python
import torch

    T18 = fd.define_tensor(shape=[-1, -1, -1, -1], contiguity=[True, True, True, True], dtype=DataType.Float, is_cpu=False)
    S91 = fd.define_scalar(12, dtype=DataType.Int)
    S92 = fd.define_scalar(128, dtype=DataType.Int)
    S93 = fd.define_scalar(25, dtype=DataType.Int)
    S94 = fd.define_scalar(32, dtype=DataType.Int)
    S95 = fd.define_scalar(2, dtype=DataType.Int)
    V96 = fd.define_vector([S91, S92, S93, S94, S95], dtype=DataType.Int)
    T97 = fd.ops.reshape(T18, new_shape=V96)
    T98 = fd.ops.slice(T97, start_indices=[0, 0, 0, 0, 0], end_indices=[12, 128, 25, 32, 1], strides=[1, 1, 1, 1, 1])
    T89 = fd.ops.sum(T98, axes=[4], keepdim=False, dtype=DataType.Null)
    fd.add_output(T89)

    nvfuser_fusion_id0(fd)
```

naoyam:
Fixes #1221.

This does a deep as opposed to shallow evaluation of the extent of a broadcast dimension when summing. Additionally, when we determine that a slice results in a broadcast, we rewrite the extent as `oneVal()` instead of leaving the slice extent expression.


naoyam:
The error was:

```
RuntimeError: Cannot reduce an axis that is marked as broadcasted as it has an undetermined size. Tried to reduce ID = bS18{( ( (nvfuser_index_t)(2) ) + ( ( fmax(0, ( fmin(( (nvfuser_index_t)(2) ), 1) )) ) + ( -( (nvfuser_index_t)(2) ) ) ) )}rf of tensor T2_l[ ?S13{( (nvfuser_index_t)(12) )}rf, ?S14{( (nvfuser_index_t)(128) )}rf, ?S15{( (nvfuser_index_t)(25) )}rf, ?S16{( (nvfuser_index_t)(32) )}rf, bS18{( ( (nvfuser_index_t)(2) ) + ( ( fmax(0, ( fmin(( (nvfuser_index_t)(2) ), 1) )) ) + ( -( (nvfuser_index_t)(2) ) ) ) )}rf ]
```

This is due to the fact that `isOneInt()` fails to detect the extent is size one. Using `evaluateInt() == 1` instead work arounds this.

@jacobhinkle, I wonder if we should replace the extent with `1` when we concretize the domain. This particular domain is concretized as a broadcast domain, so the concretize knows its extent is 1. What do you think?

jacobhinkle:
We could do that. I wonder why we have this check though. What does it mean to have an unexpanded broadcast ID with an extent that is not one. Is that ever valid?

jacobhinkle:
This particular case might also be fixed by #973 by simplifying the extent to satisfy `isOneInt()`.

naoyam:
> We could do that. I wonder why we have this check though. What does it mean to have an unexpanded broadcast ID with an extent that is not one. Is that ever valid?

I don't think that would happen with root domains, but it's possible, for example, to split a size-1 broadcast domain by a factor of, say, 4, and then the resulting domains would be `ceilDiv(i, 4)` and `4`, and both would be non-expanded broadcast domains. Since this part of the code only deals with root domains, I don't think it's a real concern here, but I suppose the check is just a sanity check.

I think it's valid to ignore the check, but simplifying the extent to 1 seems preferred still.

jacobhinkle:
> wonder if we should replace the extent with `1` when we concretize the domain.

Note that this particular error is happening at definition, on a symbolic Fusion, so we have not yet had a chance to concretize. The domain is marked broadcast during definition by the `slice` op since the input size is known to be 2 and the slice range is known to be `0:1`.

naoyam:
> > wonder if we should replace the extent with `1` when we concretize the domain.
> 
> Note that this particular error is happening at definition, on a symbolic Fusion, so we have not yet had a chance to concretize. The domain is marked broadcast during definition by the `slice` op since the input size is known to be 2 and the slice range is known to be `0:1`.

Oh, I didn't realize that, but then at the definition of the slice op, we should create the broadcast domain with extent 1, shouldn't we?

jacobhinkle:
>  at the definition of the slice op, we should create the broadcast domain with extent 1, shouldn't we?

Oh, yeah that's a good idea. That should be another one-liner so maybe I will add it to this PR along with a test.

naoyam (commented):
Thanks @jacobhinkle completing the fix. Looks like I can't approve as I started this PR, but anyone please give a stamp.

tfogal (approved):

PR #1228, opened 2023-11-06
Error out when `ExpressionEvaluator` doesn't comply with alias analysis.

wujingyue:
I.e. when it fails to produce an alias when alias analysis thinks it does.


wujingyue:
!build

PR #1230, opened 2023-11-06
Remove a redundant anonymous namespace.

wujingyue:
This is why I prefer `static` for local functions :) It's hard to figure out whether a function is in an anonymous namespace when that namespace spans across hundreds of lines.


naoyam (approved):
LGTM

PR #1238, opened 2023-11-07
Don't produce for-loops on sharded dimensions

samnordmann:
After this patch, the sharded dimensions (i.e. axis parallelized with respect to `DIDx`) won't result in a for-loop in the kernel. The index is set to be 0. This solves the issue of garbage computations.

- [1st commit](https://github.com/NVIDIA/Fuser/commit/3017c518849baba10a957c2e3b0c3dca1f626800): adding two helper functions, `bool IterDomain::isDevice() const` and `bool TensorView::isSharded() const`
- [2nd commit](https://github.com/NVIDIA/Fuser/commit/090db0c1d011bd59e77837f749e5bb7b7fa94282) don't produce for loops and set the index to 0. Update the lowering and the tests.

_Note:  However, the sharded axis will still be fully allocated -- this will be fixed in a future pr_

## Example:
Let us consider the following simple fusion:
```
Fusion fusion
FusionGuard fg(&fusion)

auto tv0 = makeConcreteTensor({4, 32});
auto tv1 = add(tv0, tv0);
fusion.addInput(tv0);
fusion.addOutput(tv1);

tv0->setDeviceMesh({0,1,2,3});
tv1->setDeviceMesh({0,1,2,3});
tv0->axis(0)->parallelize(ParallelType::DIDx);
tv1->axis(0)->parallelize(ParallelType::DIDx);
```
Before this patch, the generated kernel would be something like:
```
float[4, 32] tv1;
int i,j;
for (i=0 ; i < 4; i++) {
    for (j=0 ; j < 32; j++) {
        tv1[i][j] = tv0[i][j] + tv0[i][j];
    }
}
```

With this patch, the generated kernel is:
```
float[4,32] tv1;
int j;
for (j=0 ; j < 32; j++) {
    tv1[0][j] = tv0[0][j] + tv0[0][j];
}
```


cowanmeg:
LGTM!

csrc/compute_at_map.cpp
```diff
@@ -806,12 +806,24 @@ void ComputeAtMap::allocateIndexVariables() {
...
+    // We don't allocate any index variable for domains which
+    // are parallelized accross devices
+    auto result = std::find_if(
+        loop_disjoint_set->vector().begin(),
+        loop_disjoint_set->vector().end(),
+        [](IterDomain* id) { return id->isDevice(); });
+    if (result != loop_disjoint_set->vector().end()) {
+      loop_index_variable_map_[loop_disjoint_set.get()] = fusion_->zeroVal();
+      continue;
+    }
```
naoyam, line 819:
```suggestion
    if (auto result = std::find_if(
        loop_disjoint_set->vector().begin(),
        loop_disjoint_set->vector().end(),
        [](IterDomain* id) { return id->isDevice(); });
        result != loop_disjoint_set->vector().end()) {
      loop_index_variable_map_[loop_disjoint_set.get()] = fusion_->zeroVal();
      continue;
    }
```

csrc/compute_at_map.cpp
```diff
@@ -806,12 +806,24 @@ void ComputeAtMap::allocateIndexVariables() {
...
+
     // first allocate thread and grid parallel indices:
     //  The validation pass will check that the parallel bindings within the
     //  loop nodes are consistent so all the loops within this disjoint set
     //  will be realized implicitly using parallel index variables.
 
-    auto result = std::find_if(
+    result = std::find_if(
```
naoyam, line 826:
```suggestion
    if (auto result = std::find_if(
```
(same as the above)

csrc/ir/interface_nodes.h
```diff
@@ -527,6 +527,11 @@ class TensorView : public Val {
     return promote_reuse_;
   }
 
+  // Returns whether a TensorView has its first non-reduction axis parallelized
+  // on Didx
+  // Checks that the other non-reduction axis are not parallelized on Didx
+  bool isSharded() const;
```
naoyam, line 533: nit: I'd prefer this to be put in some util header for multidevice to make the definition of `TensorView` concise.

csrc/ir/internal_base_nodes.h
```diff
@@ -229,6 +229,10 @@ class IterDomain : public Val {
     return (isBlockDim() || isThreadDim());
   }
 
+  bool isDevice() const {
```
naoyam, line 232: To make the naming consistent as `isThreadDim`, `isBlockDim`:

```suggestion
  bool isDeviceDim() const {
```

naoyam (commented):
Mostly looks good, just some minor changes I'd like to have

csrc/ir/interface_nodes.h
samnordmann: done! I created the files `multidevice/utils.(h|cpp)` which will be populated with other functions as we progress

csrc/ir/internal_base_nodes.h
samnordmann: I renamed to isDeviceDim()

About renaming `isThread` to `isParallelized` I think it makes sense. 

What do you mean though by "include `isBlockDim`"? Did you rather mean  "include `isDeviceDim`" ? If yes, I agree with you in principle, but on the other hand I find it safer for now to just add isDeviceDim() wherever needed

samnordmann:
!build

csrc/ir/internal_base_nodes.h
naoyam: Sorry, yes, I meant `isDeviceDim`. As you needed to change some parts of the conditions to include both `isThread()` and `isDeviceDim()`, if the former included `isDeviceDim`, it didn't need any change. We would need to review all the uses of `isThread`, so I think it's fine for now, but something to consider later.

naoyam (approved):
LGTM

PR #1239, opened 2023-11-07
add multidevice debug utilities: print and barrier

samnordmann:
Add two debug utilities to be used in gtest. They are controlled through environment variables
- **debug print:** activated by setting `NVFUSER_MULTIDEVICE_DEBUG_PRINT=1`
- **do barrier at the end of each test:** activated by setting `NVFUSER_MULTIDEVICE_DEBUG_BARRIER=1`

*Note: when using mpirun, setting an environment variable is done with the "-x" option, e.g.,*
```
mpirun -x NVFUSER_MULTIDEVICE_DEBUG_PRINT=1 -np 6 $BUILD_DIRECTORY/nvfuser_tests --gtest_filter=*PipelineTest*
```


zasdfgbnm:
How about using this?
https://github.com/NVIDIA/Fuser/blob/7c7fdc2bbedb54a9a88687a8ad4e5955985febeb/csrc/options.h#L24

test/multidevice.cpp
```diff
@@ -21,6 +21,12 @@ auto multidevice_env = static_cast<MultiDeviceEnvironment*>(
 
 void MultiDeviceEnvironment::SetUp() {
   communicator_ = std::make_unique<Communicator>();
+  if (getenv("NVFUSER_MULTIDEVICE_DEBUG_PRINT")) {
```
naoyam, line 24: Please use `getNvFuserEnv`

naoyam (approved):
LGTM

samnordmann:
> How about using this?
> 
> https://github.com/NVIDIA/Fuser/blob/7c7fdc2bbedb54a9a88687a8ad4e5955985febeb/csrc/options.h#L24

Thanks for the good suggestion. However these options only deal with tests, and not with nvFuser lib, so maybe it is not worth adding them in this struct, right ?

zasdfgbnm:
> > How about using this?
> > https://github.com/NVIDIA/Fuser/blob/7c7fdc2bbedb54a9a88687a8ad4e5955985febeb/csrc/options.h#L24
> 
> Thanks for the good suggestion. However these options only deal with tests, and not with nvFuser lib, so maybe it is not worth adding them in this struct, right ?

Sure, if it is just test, then let's use `getNvFuserEnv`

test/multidevice.h
```diff
@@ -25,15 +25,28 @@ class MultiDeviceEnvironment : public testing::Environment {
     return communicator_.get();
   }
 
+  bool debug_print() const {
```
zasdfgbnm, line 28: nit: keep the same coding style:
```suggestion
  bool debugPrint() const {
```

zasdfgbnm (approved):

PR #1241, opened 2023-11-07
add reduction collectives

samnordmann:
In this PR, we add the following reduction MPI collectives:
- Reduce
- Allreduce
- ReduceScatter

Unary tests are provided


csrc/multidevice/communication.cpp
```diff
@@ -203,7 +206,77 @@ c10::intrusive_ptr<c10d::Work> Scatter::post(Communicator& comm) {
...
+c10::intrusive_ptr<c10d::Work> Reduce::post(Communicator& comm) {
+  if (comm.deviceId() == params_.root) {
+    assertBufferCount(params_.dst_bufs, 1);
+  } else {
+    assertBufferCount(params_.dst_bufs, 0);
+  }
+  post_common(*this, comm);
+  auto backend = comm.getBackendForTeam(params_.team);
+#ifdef USE_C10D_NCCL
+  auto nccl_backend = dynamic_cast<c10d::ProcessGroupNCCL*>(backend.get());
```
zasdfgbnm, line 224: Why hard code NCCL, instead of using:
https://github.com/pytorch/pytorch/blob/98564d2d7a8bc116674198e192dff957c3d59fc6/torch/csrc/distributed/c10d/ProcessGroup.hpp#L194

samnordmann: The only `reduce` collective exposed from the ProcessGroup is in-place (i.e. input_buffer != output_buffer at the root).
If we request an out-of-place reduction, then we need to add a local copy, as you can see below.

However NCCL PG exposes an out-of-place collective `_reduce_oop`, so I thought it was worth using it.
Does it make sense ?

naoyam (approved):
LGTM

samnordmann:
!build

samnordmann:
!build

PR #1244, opened 2023-11-07
fix allocation of intermediate tensor in pipelines

samnordmann:
We provide a new function 
```
std::unordered_map<Val*, c10::IValue> allocatePipelineIntermediateBuffers(
    Pipeline* pipeline,
    DeviceIdxType my_device_index,
    std::vector<c10::IValue> global_inputs_IValues);
```
used to allocate intermediate tensors when executing Pipeline in multidevice setting.

Before this patch, we would allocate every output of each pipeline stage (including global outputs), even if we don't run the stage. This results in a lot of garbage allocation.

This function allocates only the following intermediate tensors, that is, tensors that will be received from other devices through a network communication and used as a subsequent stage's input. The allocation of global or intermediate outputs are taken care of by the fusion executor.

we also provide a patch to fix the test after these changes


csrc/multidevice/allocator.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+  // We copy the original fusion and set the outputs to be the tensors to be
+  // allocated. This way we can directly use FusionExecutor::allocOutputSpace
+  auto [fusion_copy, copy_to_original_map] =
+      copyFusionAndChangeOutputs(pipeline->originalFusion(), vals_to_allocate);
+  if (fusion_copy->outputs().empty()) {
+    return {};
+  }
+  FusionExecutor fe;
+  fe.compileFusion(fusion_copy.get(), global_inputs_IValues);
```
naoyam, line 95: This means the copied fusion is lowered and compiled by nvrtc. And the compiled kernel is just discarded. Why do we need to do that just to allocate interemediates?

samnordmann: It is just the easiest way I found to allocate a given set of tensors. Do you have another suggestion?

naoyam: I suppose there's a way to allocate outputs without compiling the kernel and it should be just a matter of factoring out the allocation interface to work with a `Fusion` rather than a `Kernel`.

naoyam: I'm not sure how cumbersome the refactoring would be, but compiling a fusion just for allocating outputs seems to be quite wasteful.

samnordmann: I think the refactoring will be substantial and belongs to another PR. There is also a need to change the `FusionExecutor` interface since the `compile` method is the only way I see (except `runFusion`) to set the Fusion.
In the meantime I really see no other option than doing like that. Let me know what you suggest

naoyam (approved):
LGTM. As we discussed offline, please create an issue to keep track of the issue.

PR #1252, opened 2023-11-07
Fix cupti build

zasdfgbnm:
`libcupti.so` does not exist in `${CUDA_TOOLKIT_ROOT_DIR}/lib64/`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,7 +47,7 @@ add_subdirectory(${CMAKE_CURRENT_LIST_DIR}/lib/dynamic_type)
 
 # TODO: fix MSVC
 if(NOT MSVC)
-  find_library(LIBCUPTI libcupti.so PATHS ${CUDA_TOOLKIT_ROOT_DIR}/lib64/)
+  find_library(LIBCUPTI libcupti.so PATHS ${CUDA_TOOLKIT_ROOT_DIR}/extras/CUPTI/lib64//)
   find_library(LIBNVTOOLSEXT libnvToolsExt.so PATHS ${CUDA_TOOLKIT_ROOT_DIR}/lib64/)
 endif()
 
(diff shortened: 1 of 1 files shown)


zasdfgbnm:
!build

zasdfgbnm:
Errrr, the failure does not make sense. I freshly downloaded a run file to install cuda, and there is no `/path/to/cuda/lib64/libcupti.so`. In my archlinux, the path also follow the run file. Why ubuntu move files around?

zasdfgbnm:
!build

jjsjann123:
I'm totally out of context here.

Are we just struggling to find cupti at runtime?
We can try get cupti from pypi instead. https://pypi.org/project/nvidia-cuda-cupti-cu12/
Similar to how we are getting nvrtc. We can add this relative path to RPATH. cc'ing @xwang233

zasdfgbnm:
> Are we just struggling to find cupti at runtime?

Yes, exactly.



> We can try get cupti from pypi instead. https://pypi.org/project/nvidia-cuda-cupti-cu12/
> Similar to how we are getting nvrtc. We can add this relative path to RPATH. cc'ing @xwang233

Yeah, this is also possible. My current PR should work with runfile-installed CUDA and on archlinux. But I don't have a preference on which is better.

zasdfgbnm:
> But I don't have a preference on which is better.

Which makes more sense in terms of packaging wheel?

jjsjann123:
> > But I don't have a preference on which is better.
> 
> Which makes more sense in terms of packaging wheel?

oh it definitely is the pypi thing makes more sense. i.e. people can pip install torch and pip install nvfuser to have it running. Nobody want to mess with CUDA installation :laughing:

jjsjann123:
I think the only thing you need to change here is just the `CMakeLists.txt` to add the proper RPATH.
We'll need to update our CI script in the manylinux build to pass the "proper" cupti dependency per our cuda version. Which maybe @xwang233 could help with that?!

zasdfgbnm:
I will close this PR and let @xwang233 help.

zasdfgbnm:
Had some offline chat with @xwang233 . Let's use our local CUDA for local development, and use pypi for wheel release. Reopen this PR.

xwang233 (approved):
If this can fix your local build on arch, go for it.

PR #1263, opened 2023-11-08
Simplify index generation of ldmatrix and mma ops by allowing `kir::TensorIndex` to view the memory differently, 

zasdfgbnm:
The dtype of `kir::TensorIndex` is no longer required to be the same as its `view()`. If they are different, then a `reinterpret_cast` is generated to convert the type. `ldmatrix` and the output of mma is changed to utilize this feature. This PR pave the way towards lowering `MmaOp` as `kir::Asm`.

This PR only adjusts the style of generated code, there should be no behavioral change.


zasdfgbnm:
!build

zasdfgbnm:
!build

zasdfgbnm:
!build

zasdfgbnm:
!build

zasdfgbnm:
!build

naoyam:
> The dtype of kir::TensorIndex is no longer required to be the same as its view().

Can you tell a bit more why this is necessary? It seems unintuitive to me that `TensorIndex` could change a type.

zasdfgbnm:
> > The dtype of kir::TensorIndex is no longer required to be the same as its view().
> 
> Can you tell a bit more why this is necessary? It seems unintuitive to me that `TensorIndex` could change a type.

This is required for vectorization and MMA parallelized TVs. For example, if a tensor is parallelized as `iV{8}`, by default, the allocation may be `__half T1[32]`, but it actually needs to be viewed as `Array<__half, 8, 8>[4]` or `Array<__half, 8, 1>[4]` (for register tensors, the alignment does not really matter). Same applies for MMA parallelized tensors: we may want to view `__half T1[32]` as `Array<unsigned, 4, 1> T1[4]`.

zasdfgbnm:
> > The dtype of kir::TensorIndex is no longer required to be the same as its view().
> 
> Can you tell a bit more why this is necessary? It seems unintuitive to me that `TensorIndex` could change a type.

I think the change here is mostly from "specializing by expr" to "specializing by tensor index". In the past, in `codegen.cpp`, we say "this is a vectorized load store op,  it is a special op that we need to treat its inputs and outputs as arrays". But for now, it is changed to "this is a special tensor index that should be viewed as an array".

naoyam:
Thanks. It's getting clearer. One more question, I don't see any change in codegen.cpp for normal vectorizations. If this PR is required for vectorization, I wonder if there should be some change?

zasdfgbnm:
> is required for vectorization

It is not *required* for vectorization. It can be used for vectorization, but not mandatory. There are two ways to do things, the new way, and the legacy way. There is nothing wrong with the legacy way, it is just the new way is more convenient for exprs that will be lowered to `kir::Asm`

naoyam (approved):
LGTM

PR #1266, opened 2023-11-08
Add UCC support for Communicator

cowanmeg:
- Communicator supports UCC and NCCL backends. The Communicator is created with a default backend that can be changed with `setDefaultBackend`.
- Communications can select which backend to use, otherwise the default backend is used.
- Add tests for UCC collectives. 

Run NCCL collective tests with: `mpirun -np <n> /opt/pytorch/nvfuser/nvfuser_tests --gtest_filter="*CommunicationTest*/0"`
Run UCC collective tests with: `mpirun -np <n> /opt/pytorch/nvfuser/nvfuser_tests --gtest_filter="*CommunicationTest*/1"`


csrc/multidevice/communicator.h
```diff
@@ -31,8 +31,8 @@ namespace nvfuser {
 
 using RankType = DeviceIdxType;
 
-// Supported backends. TODO: only tested with nccl for now
-enum class CommunicatorBackend { nccl, ucc, gloo };
+// Supported backends. TODO: gloo untested
+enum class CommunicatorBackend { nccl, ucc, gloo, none };
```
samnordmann, line 35: What about renaming `none` with `default`?
Actually, maybe we don't need to add a new entry to CommunicatorBackend, but rather to make the backend type argument optional inside the methods that use it
What do you think ?

test/test_multidevice_communications.cpp
```diff
@@ -177,14 +180,19 @@ TEST_F(CommunicationTest, Communication_SendRecvToSelf) {
...
 
     auto obtained = params.dst_bufs.at(0);
     auto ref = at::arange(tensor_size, tensor_options) + j;
     validate(obtained, ref);
   }
 }
 
+INSTANTIATE_TEST_SUITE_P(CommunicatorBackend,
+                         CommunicationTest,
+                         testing::Values(CommunicatorBackend::nccl,
```
samnordmann, line 193: Should we skip the test if ucc or nccl is not available?

csrc/multidevice/communicator.h
```diff
@@ -31,8 +31,8 @@ namespace nvfuser {
...
 
 constexpr CommunicatorBackend comm_backend_default = CommunicatorBackend::nccl;
```
samnordmann, line 37: What about falling back to ucc if nccl is not available ? This can still be determined at compile time, with something like:
```
#ifdef USE_C10D_NCCL
constexpr CommunicatorBackend comm_backend_default = CommunicatorBackend::nccl;
#elif USE_C10D_UCC
constexpr CommunicatorBackend comm_backend_default = CommunicatorBackend::ucc;
#endif
```

samnordmann (commented):
Thanks Meghan ! This looks really great :)

samnordmann:
Could we test the new backend in the tests from `test/test_multidevice_pipeline.cpp` as well?
It should work as is hopefully

csrc/multidevice/communicator.h
cowanmeg: I think optional arguments will be the cleanest. Looks like Fuser mostly uses std::optional instead of c10::optional so I'll use std.

samnordmann (approved):

csrc/multidevice/communicator.cpp
```diff
@@ -173,15 +179,20 @@ Communicator::Communicator(
...
-  // creates the world's backend
-  std::vector<RankType> all_ranks(size_);
-  std::iota(all_ranks.begin(), all_ranks.end(), 0);
-  world_ = getBackendForTeam(all_ranks);
+#if defined(USE_C10D_UCC) && defined(NVFUSER_BUILD_WITH_UCC)
+  ucc_available_ = true;
+#endif
+
+#ifdef USE_C10D_NCCL
+  nccl_available_ = false;
```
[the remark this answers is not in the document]
cowanmeg: haha good catch

cowanmeg:
!build

naoyam:
I'm not familiar with UCC, but as long as @samnordmann approves, please go ahead.

samnordmann:
!build

cowanmeg:
!build

PR #1267, opened 2023-11-08
Replace Serialization-only Enum with cast to underlying type

rdspring1:
This PR removes the Datatype Enum from the FlatBuffers schema. It is replaced with a cast to the original Enum's underlying type. During deserialization, we cast the underlying type back to the correct Enum.


rdspring1:
!build

csrc/type.h
```diff
@@ -995,7 +995,7 @@ inline PolymorphicValue castToDtype(
 
 // Convert an enum to its underlying type.
 template <typename E>
-constexpr auto toUnderlying(E e) noexcept {
+constexpr auto castEnumToUnderlyingType(E e) noexcept {
```
jjsjann123, line 998: Just double check my understanding here.

I see that we are using `castEnumToUnderlyingType` to replace `serde::mapToSerdeDtype`, which has a few overloads on different dtypes.

I guess our assumption here is that, we'll always serialize the one object to a int and deserialize the int back to the same object. So even though the `long dtype` thing could carry different objects/types in the serialized form. The deserialized object would still be consistent so we are all good.

jjsjann123 (approved):
`+58 - 179` :clap:

csrc/type.h
jjsjann123: > It is replaced with a cast to the NvFuser Enum's underlying type. During deserialization, we cast the underlying type back to the correct NvFuser enum.

So this is not just the NvFuser enum, since we are also handling the ScalarType from aten? Might want to update the PR description.

csrc/type.h
```diff
@@ -995,7 +995,7 @@ inline PolymorphicValue castToDtype(
 
 // Convert an enum to its underlying type.
 template <typename E>
-constexpr auto toUnderlying(E e) noexcept {
```
jacobhinkle, line 998: The new name is more descriptive. Note however that originally I named it after the C++23 feature https://en.cppreference.com/w/cpp/utility/to_underlying. Because of this it might be best to rename it `std::to_underlying` and place it in a C++23 directory like we are temporarily doing with C++20 features. Paging @zasdfgbnm for his opinion on shadowing C++23 functionality.

rdspring1:
!build

rdspring1:
!build

csrc/type.h
rdspring1: I found a name conflict with `toUnderlying` with pytorch. It is defined for the `c10::ScalarType`. https://github.com/pytorch/pytorch/blob/main/c10/core/ScalarType.h#L363

rdspring1: I switched to back to the original name but specified the `nvfuser` namespace to avoid conflicts.

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

jacobhinkle (approved):

PR #1271, opened 2023-11-09
Add variants of `slice` that takes `int64_t`s.

wujingyue:
This is easier for C++ unit tests to use and is closer to the Python API. I also moved some error checking code from `python_bindings.cpp` to the new variants.

### DIFF
--- a/csrc/python_frontend/python_bindings.cpp
+++ b/csrc/python_frontend/python_bindings.cpp
@@ -2344,8 +2344,8 @@ void initNvFuserPythonBindings(PyObject* module) {
       "slice",
       [](FusionDefinition::Operators& self,
          Tensor arg,
-         std::vector<int64_t>& start_indices,
-         std::vector<int64_t>& end_indices,
+         const std::vector<int64_t>& start_indices,
+         const std::vector<int64_t>& end_indices,
          // NOTE: Tried to use std::reference_wrapper to a vector and during
          // testing, I was not getting the proper value back.  It was like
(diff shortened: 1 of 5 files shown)


csrc/python_frontend/python_bindings.cpp
```diff
@@ -2366,50 +2366,10 @@ void initNvFuserPythonBindings(PyObject* module) {
               opt_strides.value().size());
           strides.assign(
               opt_strides.value().begin(), opt_strides.value().end());
+        } else {
+          strides.resize(start_indices.size(), 1);
         }
-        NVF_CHECK(
```
jjsjann123, line 2370: Can we keep these checks here?

There's a difference doing them here vs doing them in the `operator()`. i.e.

```
  ...
  fd.ops.slice([illegal_slice). # currently we throw error here.
  ...
```

Moving the checks into operator(), we only throw error when we construct fusion, i.e. when we exit the context manager.
Not a deal breaker, but would prefer to have semantic check at the place when the python API is called.

jjsjann123: alternatively, we can also do checks in `SliceOpRecord`:ctor.

wujingyue: Good point! Done.

naoyam (approved):
LGTM. That's indeed simpler and easier.

jjsjann123 (approved):
LGTM

PR #1272, opened 2023-11-09
Fix invisible uses in DeadCodeRemover traversal

### LINKED ISSUE #1246: Inferring output shape from an operation that consumes a Zero-element Tensor Fails
Error:
```
RuntimeError: inferred_val.hasValue() INTERNAL ASSERT FAILED at "/workspace/Fuser/csrc/executor.cpp":525, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. Could not launch kernel as program could not infer i1(i1) for the buffer T6_g[ bS33{1}, iS34{i1}, iS35{i2}, iS36{i3} ]
Exception raised from inferShape at /workspace/Fuser/csrc/executor.cpp:525 (most recent call first):
```

Repro:
```python
import torch

    T0 = fd.define_tensor(shape=[1, -1, -1, -1], contiguity=[None, True, True, True], dtype=DataType.Float, is_cpu=False)
    T1 = fd.define_tensor(shape=[1, -1, -1, -1], contiguity=[None, True, False, True], dtype=DataType.Float, is_cpu=False)
    S2 = fd.define_scalar(2.00000, dtype=DataType.Double)
    T3 = fd.ops.mul(T0, S2)
    T4 = fd.ops.cat([T3, T1], dim=-1)
    fd.add_output(T4)

    nvfuser_fusion_id0(fd)

inputs = [
    torch.randn((8388608,), dtype=torch.float32, device='cuda:0').as_strided((1, 32, 2048, 128), (8388608, 262144, 128, 1)),
    torch.randn((0,), dtype=torch.float32, device='cuda:0').as_strided((1, 32, 2048, 0), (8388608, 262144, 128, 1)),
]
fd.execute(inputs)
```
One thing to note is that if the `mul` operation is removed and there is not an intermediate, the `cat` is fine with an zero-element tensor.

The Fusion IR for the Kernel that is about to be launched looks empty?

```
Inputs:
  T0_g[ bS0{1}, iS1{i1}, iS2{i2}, iS3{i3} ], float
  T1_g[ bS4{1}, iS5{i6}, iS6{i7}, iS28{0} ], float
Outputs:
  T6_g[ bS31{1}, iS32{i1}, iS33{i2}, iS34{i3} ], float

%kernel_math {
}
```

Strangely, the Fusion IR for the unscheduled Fusion shows a `pad` in addition to the `cat`.  `cat` can also be represented as `pad` + `add`.

```python
Inputs:
  T0_g[ bS0{1}, iS1{i1}, iS2{i2}, iS3{i3} ], float
  T1_g[ bS4{1}, iS5{i6}, iS6{i7}, iS7{i8} ], float
Outputs:
  T5_g[ bS22{1}, iS23{i6}, iS24{i7}, ?S25{( i8 + i3 )} ], float

%kernel_math {
i18 = i8 + i3;
i20 = i18 - i8;
T3_l[ bS12{1}, iS13{i6}, iS14{i7}, ?S16{( i8 + i3 )}rf ]
   = pad( T1_g[ bS4{1}, iS5{i6}, iS6{i7}, iS7{i8} ], {0, 0, 0, 0, 0, 0, 0, i20} )
T2_l[ bS8{1}, iS9{i1}, iS10{i2}, iS11{i3} ]
   = T0_g[ bS0{1}, iS1{i1}, iS2{i2}, iS3{i3} ]
   * double(2);
i22 = 0 + i8;
T4_l[ bS17{1}, iS18{i1}, iS19{i2}, ?S21{( i3 + ( 0 + i8 ) )}rf ]
   = pad( T2_l[ bS8{1}, iS9{i1}, iS10{i2}, iS11{i3} ], {0, 0, 0, 0, 0, 0, i22, 0} )
T5_g[ bS22{1}, iS23{i6}, iS24{i7}, ?S25{( i8 + i3 )} ]
   = cat( T3_l[ bS12{1}, iS13{i6}, iS14{i7}, ?S16{( i8 + i3 )}rf ], T4_l[ bS17{1}, iS18{i1}, iS19{i2}, ?S21{( i3 + ( 0 + i8 ) )}rf ], 3 )
}
```

jacobhinkle:
This fixes #1246 which was a bug in `DeadCodeRemover`, which is used in the `EmptyTensorRemover` pre-segmentation pass. In the fusion from that issue, `T2` is erroneously removed from the Fusion, causing the Fusion to appear empty. That removal occurs after the backward traversal, during which it is marked dead despite the fact that it was earlier used to define the new tensor `T6 = set(T2)` which was also earlier registered to replace `T5`.

The underlying issue was that `Val::uses()` only returns uses that lead to Fusion outputs (see `Fusion::resetTvUses()`). It will not just return all the `Expr`s for which that `Val` is an input. So during the backward traversal, even though `T2` should've been kept alive by `T6`'s definition, that use was not returned so `allUsesDead()` returned `true`.

This PR tracks future live uses of `Val`s when they are marked as live in order to avoid this overzealous pruning.


python_tests/test_python_frontend.py
```diff
@@ -2708,6 +2708,46 @@ def fusion_func(fd: FusionDefinition) -> None:
...
+                S2 = fd.define_scalar(2.00000, dtype=DataType.Double)
+                T3 = fd.ops.mul(T0, S2)
+                T4 = fd.ops.cat([T3, T1], dim=-1)
+                if final_mul:
+                    # NOTE: original repro does not have this final op
+                    S3 = fd.define_scalar(1.00000, dtype=DataType.Double)
+                    T5 = fd.ops.mul(T4, S3)
+                    fd.add_output(T5)
+                else:
+                    fd.add_output(T4)
```
jacobhinkle, line 2745: Without the `final_uses_` part of this PR, the `final_mul==False` branch will succeed but the `final_mul==True` branch will fail.

jacobhinkle:
!build

csrc/iter_visitor.cpp
```diff
@@ -469,8 +469,9 @@ void BackwardVisitor::traverseTo(
 
   {
     size_t pos = 0;
-    for (auto expr : exprs)
+    for (auto expr : exprs) {
```
jjsjann123, line 472: :clap:

csrc/iter_visitor.cpp
```diff
@@ -679,8 +680,9 @@ class DependentVals : public IterVisitor {
   std::unordered_set<Val*> boundary_;
 
   std::vector<Statement*> next(Val* v) override {
-    if (boundary_.find(v) != boundary_.end())
+    if (boundary_.find(v) != boundary_.end()) {
```
jjsjann123, line 683: :clap:

csrc/iter_visitor.cpp
```diff
@@ -1102,9 +1107,9 @@ void DeadCodeRemover::markLiveRecursive(Statement* stmt) {
     return;
   }
   markLive(stmt);
-  if (stmt->isVal() && stmt->asVal()->definition()) {
-    markLiveRecursive(stmt);
-  } else {
+  if (stmt->isVal() && stmt->asVal()->definition() != nullptr) {
+    markLiveRecursive(stmt->asVal()->definition());
```
jjsjann123, line 1111: is this just trying to avoid marking stmt live again, since we already called `markLive(stmt)`?

csrc/iter_visitor.h
```diff
@@ -480,6 +488,21 @@ class DeadCodeRemover : BackwardVisitor {
   //! Mark a single Statement as being alive.
   inline void markLive(Statement* stmt) {
     live_statements_.insert(stmt);
+    if (auto e = dynamic_cast<Expr*>(stmt)) {
+      // Check if this expression is already in uses() for each of its inputs
+      // and if not, record it in future_uses_
+      for (Val* inp : e->inputs()) {
+        if (std::find(inp->uses().begin(), inp->uses().end(), e) ==
+            inp->uses().end()) {
+          auto fu_it = future_uses_.find(inp);
```
jjsjann123, line 497: how did we enter here?

so `e` has `inp` as its inputs but `inp` doesn't have `e` in its uses? we have a invalid graph here?

csrc/iter_visitor.cpp
jacobhinkle: This was a bug where we were actually not recursing into the definition actually!

csrc/iter_visitor.h
jacobhinkle: The graph is valid, but `inp->uses()` only gives us uses **that lead to outputs**. In this function `stmt` will often be a replacement `Val` which is temporarily "dead" in that it hasn't yet been hooked up to a path leading to outputs.

jacobhinkle: IDK whether we should leave this behavior of `uses()` or possibly change it since it's slightly surprising. It has been here a long time though: https://github.com/csarofeen/pytorch/commit/cfaecaf40bd6cabd3f4e0ef0d8c7252655349b61#diff-6e4fc28533e1790003083738b2994e3a44e45a05ff5264193d199357ff78baebR465-R488.

jjsjann123: Ha, thanks for bearing with my ignorance. `StmtSort::getStmts` adds an artificial end point `fusion->getTerminationOutputs()`.

jjsjann123 (approved):
LGTM

jacobhinkle:
!build

csrc/iter_visitor.h
```diff
@@ -480,6 +488,21 @@ class DeadCodeRemover : BackwardVisitor {
...
+          if (fu_it == future_uses_.end()) {
+            future_uses_.emplace(inp, std::unordered_set<Expr*>({e}));
+          } else {
+            fu_it->second.insert(e);
+          }
```
naoyam, line 502: nit: this could be just:

```
future_uses_[inp].insert(e);
```

csrc/iter_visitor.h
```diff
@@ -462,6 +462,14 @@ class DeadCodeRemover : BackwardVisitor {
 
   //! Check whether all uses have been marked dead
   inline bool allUsesDead(Val* val) const {
+    auto fu_it = future_uses_.find(val);
+    if (fu_it != future_uses_.end() && !fu_it->second.empty()) {
+      // Regardless of whether current uses are marked dead, this appears in a
+      // replacement expression, so it has a future live use and we should keep
+      // it.
+      return false;
+    }
```
naoyam, line 471: Uses in `future_uses_` may be really dead, so aren't we going to miss real dead uses? In other words, does this change mean this analysis is now less precise and more conservative?

naoyam:
IIUC, this issue happened only when there's an active replacement while trying to remove dead code, which seems to be really tricky. Would it be possible to separate those two processes apart? Could we first do the replacement and apply the DCE later?

csrc/iter_visitor.h
jacobhinkle: Future uses holds Exprs that are used in replacements, so although they are currently dead they will not be after this pass is done. Previously we were not conservative enough and wound up removing Vals which were used only in such future uses.

jacobhinkle: I don't think we will miss removing actual dead code as long as everything registered as a replacement winds up alive. If we register two different replacements for the same Val this might be violated I suppose, but like OptOutMutator, this class assumes we're not doing that.

naoyam: > Future uses holds Exprs that are used in replacements

Is this guaranteed in general?

naoyam: It seems that the code here assumes the particular use case of `DeadCodeRemover`, but I'm assuming the class itself is designed to be more generic as it's defined in iter_visitor.h.

jacobhinkle:
> IIUC, this issue happened only when there's an active replacement while trying to remove dead code, which seems to be really tricky. Would it be possible to separate those two processes apart? Could we first do the replacement and apply the DCE later?

That seems preferable: have an OptOutMutator that replaces empty tensors then a separate DCE pass at the end of this preseg stage. I'll look at that possibility in more detail.

naoyam:
> > IIUC, this issue happened only when there's an active replacement while trying to remove dead code, which seems to be really tricky. Would it be possible to separate those two processes apart? Could we first do the replacement and apply the DCE later?
> 
> That seems preferable: have an OptOutMutator that replaces empty tensors then a separate DCE pass at the end of this preseg stage. I'll look at that possibility in more detail.

Thanks for confirming my understanding. Since `DeadCodeRemover` is only used there, I think it's fine to leave it as is with a TODO note to clean it up.

PR #1274, opened 2023-11-09
`SliceOp::in` and `::out` return `TensorView*` instead of `Val*`.

wujingyue:
I believe they are always `TensorView*` given how `SliceOp`s are created:
https://github.com/search?q=repo%3ANVIDIA%2FFuser+create%3Csliceop%3E&type=code


wujingyue:
!build

zasdfgbnm (approved):
Not sure if it can also be `kir::TensorIndex`, but if CI is happy with it, then this won't be a problem.

PR #1280, opened 2023-11-11
Fix dangling dead uses in Fusion::removeVal

### LINKED ISSUE #1270: Bug in Segmentation of reductions with empty tensors?
There appears to be a segfault while copying a Fusion object.  I think this is for segmentation as there happens to be an inner and outer reduction.  If I remove one or the other, the fusion is just fine.

Repro:

```python
import torch

    T0 = fd.define_tensor(shape=[-1, -1], contiguity=[True, None], dtype=DataType.BFloat16, is_cpu=False)
    T1 = fd.define_tensor(shape=[-1, -1], contiguity=[None, True], dtype=DataType.BFloat16, is_cpu=False)
    T2 = fd.ops.cast(T1, dtype=DataType.Float)
    S3 = fd.define_scalar(1.00000, dtype=DataType.Double)
    T4 = fd.ops.full(fill_value=S3, shape=[5, 0], dtype=DataType.BFloat16)
    T5 = fd.ops.cast(T4, dtype=DataType.Float)
    T6 = fd.ops.mul(T2, T5)
    T7 = fd.ops.cast(T0, dtype=DataType.Float)
    T8 = fd.ops.mul(T7, T5)
    T24 = fd.ops.sum(T6, axes=[1], keepdim=False, dtype=DataType.Null)
    T11 = fd.ops.sum(T8, axes=[0], keepdim=False, dtype=DataType.Null)
    fd.add_output(T24)
    fd.add_output(T11)

inputs = [
         torch.randn(5, 0, device='cuda', dtype=torch.bfloat16),
         torch.randn(5, 0, device='cuda', dtype=torch.bfloat16),
        ]

    nvfuser_fusion_id0(fd)

out = fd.execute(inputs)
```

jacobhinkle:
Fixes #1270.

When a `Val` is removed from the `Fusion`, we currently first loop over `val->uses()` and remove them all from the Fusion. This seems correct at first glance, but it is incomplete since `val->uses()` actually only gives all **live** uses. When there is dead code in the `Fusion` that includes some uses of a val that is to be removed, we can wind up with an expression that holds an invalid pointer to the removed value in its `inputs()`. In #1270 this caused a segfault when the fusion was cloned since that will clone not only live objects but also these dangerous dangling dead ones.

This PR iterates over `exprs_` in `Fusion::removeVal()` in order to definitively find uses, dead or alive. Note that we still do not check attributes for the presence of an `attributeVal` equal to `val` so we could probably still contrive a case that would lead to dangling references. If we do check such attributes we should maybe produce an error instead of silenlty removing the `Expr`. I am not sure how prevalent `Val` attributes are or what would be preferred in those cases.

### DIFF
--- a/csrc/fusion.cpp
+++ b/csrc/fusion.cpp
@@ -205,7 +205,6 @@ void Fusion::removeVal(Val* val) {
     removeExpr(orig);
   }
 
-  const std::unordered_set<Expr*> exprs = exprs_;
   std::vector<Expr*> exprs_to_remove;
   for (Expr* e : exprs_) {
     if (!inContainer(e)) {
--- a/csrc/fusion.cpp
+++ b/csrc/fusion.cpp
@@ -205,7 +205,14 @@ void Fusion::removeVal(Val* val) {
     removeExpr(orig);
   }
 
-  // We previously first looped over val->uses() and removed them all from the Fusion. This seems correct at first glance, but it is incomplete since `val->uses()` actually only gives all live uses. When there is dead code in the Fusion that includes some uses of a val that is to be removed, we can wind up with an expression that holds an invalid pointer to the removed value in its inputs(). In https://github.com/NVIDIA/Fuser/issues/1270 this caused a segfault when the fusion was cloned since that will clone not only live objects but also these dangerous dangling dead ones.
+  // We previously first looped over val->uses() and removed them all from the
+  // Fusion. This seems correct at first glance, but it is incomplete since
+  // `val->uses()` actually only gives all live uses. When there is dead code in
+  // the Fusion that includes some uses of a val that is to be removed, we can
+  // wind up with an expression that holds an invalid pointer to the removed
+  // value in its inputs(). In https://github.com/NVIDIA/Fuser/issues/1270 this
+  // caused a segfault when the fusion was cloned since that will clone not only
+  // live objects but also these dangerous dangling dead ones.
   std::vector<Expr*> exprs_to_remove;
   for (Expr* e : exprs_) {
     if (!inContainer(e)) {
--- a/csrc/fusion.cpp
+++ b/csrc/fusion.cpp
@@ -205,6 +205,7 @@ void Fusion::removeVal(Val* val) {
     removeExpr(orig);
   }
 
+  // We previously first looped over val->uses() and removed them all from the Fusion. This seems correct at first glance, but it is incomplete since `val->uses()` actually only gives all live uses. When there is dead code in the Fusion that includes some uses of a val that is to be removed, we can wind up with an expression that holds an invalid pointer to the removed value in its inputs(). In https://github.com/NVIDIA/Fuser/issues/1270 this caused a segfault when the fusion was cloned since that will clone not only live objects but also these dangerous dangling dead ones.
   std::vector<Expr*> exprs_to_remove;
   for (Expr* e : exprs_) {
     if (!inContainer(e)) {
--- a/csrc/fusion.cpp
+++ b/csrc/fusion.cpp
@@ -205,8 +205,20 @@ void Fusion::removeVal(Val* val) {
     removeExpr(orig);
   }
 
-  for (Expr* use : unordered_uses(val)) {
-    removeExpr(use);
+  const std::unordered_set<Expr*> exprs = exprs_;
+  std::vector<Expr*> exprs_to_remove;
+  for (Expr* e : exprs_) {
+    if (!inContainer(e)) {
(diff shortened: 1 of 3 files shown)


jacobhinkle:
!build

wujingyue:
Thanks for the detailed explanation! Pardon my ignorance of the code base. I'm not sure we are fixing the right place. 

> it is incomplete since val->uses() actually only gives all live uses.

Is this supposed to happen or is `val->uses` supposed to capture dead uses as well? 

> When there is dead code in the Fusion

Is this supposed to happen? As an alternative, I would expect dead code are gone (e.g. by `DeadCodeRemover`) or at least unlinked so they are not referring to anything.

jacobhinkle:
> Thanks for the detailed explanation! Pardon my ignorance of the code base. I'm not sure we are fixing the right place. 
> 
> > it is incomplete since val->uses() actually only gives all live uses.
> 
> Is this supposed to happen or is `val->uses` supposed to capture dead uses as well? 

It was surprising to me too but it seems intentional as both `Val::uses()` and `Fusion::exprs()` restrict between inputs and outputs. One reason may be to improve performance e.g. when processing a small segment in a large Fusion.

> > When there is dead code in the Fusion
> 
> Is this supposed to happen? As an alternative, I would expect dead code are gone (e.g. by `DeadCodeRemover`) or at least unlinked so they are not referring to anything. 

It's tough to know when to do this DCE though. Dead code exists at definition until `addOutput` is called. In this case it happens even after that due to some cleverness in recognizing a trivial reduction. So we will have some unused Statments in the Fusion at times. As long as we dont have dangling pointers this is fine. This PR just tries to isolate the operation that could lead to a dangling invalid pointer and ensure that dead code is not ignored instead of an explicit DCE pass. DeadCodeRemover is only instantiated currently for RemoveEmptyPass btw: we could do a generic DCE pass with it as a presegmentation pass if we need to.

jacobhinkle:
I'd also note that the way we do segmentation is to create a segment by just changing inputs and outputs of the fusion. So everything outside the segment becomes dead. DCE then would be doing a lot of movement outside the segment which might not be necessary.

csrc/fusion.cpp
```diff
@@ -205,8 +205,20 @@ void Fusion::removeVal(Val* val) {
     removeExpr(orig);
   }
 
-  for (Expr* use : unordered_uses(val)) {
-    removeExpr(use);
+  const std::unordered_set<Expr*> exprs = exprs_;
```
wujingyue, line 208: I don't think `exprs` is used anywhere. 

```suggestion
```

csrc/fusion.cpp
```diff
@@ -205,8 +205,20 @@ void Fusion::removeVal(Val* val) {
...
+  std::vector<Expr*> exprs_to_remove;
+  for (Expr* e : exprs_) {
```
wujingyue, line 210: Having to scan the whole expression list to remove just one `Val` isn't ideal. We may need to later discuss broadly about the contracts around dead code. However, this PR seems to be the fastest and practical way to fix the bug. So LGTM!

wujingyue (approved):
Thank you!

csrc/fusion.cpp
jacobhinkle: Ah thanks for catching this. I was previously copying the whole set to prevent altering it inside the loop.

csrc/fusion.cpp
jacobhinkle: Yeah I agree. I considered adding another member to track this but ultimately decided it wouldn't be worth maintaining since Val removal is not done all that often.

csrc/fusion.cpp
```diff
@@ -205,8 +205,19 @@ void Fusion::removeVal(Val* val) {
     removeExpr(orig);
   }
 
-  for (Expr* use : unordered_uses(val)) {
-    removeExpr(use);
+  std::vector<Expr*> exprs_to_remove;
```
wujingyue, line 208: Add a comment for posterity. 

```suggestion
  // We previously first looped over val->uses() and removed them all from the Fusion. This seems correct at first glance, but it is incomplete since `val->uses()` actually only gives all live uses. When there is dead code in the Fusion that includes some uses of a val that is to be removed, we can wind up with an expression that holds an invalid pointer to the removed value in its inputs(). In https://github.com/NVIDIA/Fuser/issues/1270 this caused a segfault when the fusion was cloned since that will clone not only live objects but also these dangerous dangling dead ones.
  std::vector<Expr*> exprs_to_remove;
```

PR #460, opened 2023-06-06
Clip slice range expressions

jacobhinkle:
This PR normalizes the inputs to `slice` in order to mimic the semantics of numpy/PyTorch slicing. For an axis with extent `ext`, if we receive a slice of `(start, stop, step)` we normalize it to `(norm_start, norm_stop, step)` where
```
norm_start = max(0, start < 0 ? start + ext : start);
norm_stop = max(norm_start, min(ext, stop < 0 ? stop + ext : stop));
```

Specific changes in this PR:
- Form the above expressions in the `slice` op.
- Add shmoo tests that test various scenarios with constant and input size slices.

The simple Fusion in the input range test prints like this:
```
Inputs:
  T0_g[ iS0{9} ], float
  i3, nvfuser_index_t
  i4, nvfuser_index_t
Outputs:
  T1_g[ ?S2{( ( ( -( fmax(0, ( where(( i3 < 0 ), ( i3 + 9 ), i3) )) ) ) + 9 ) + ( ( fmax(( fmax(0, ( where(( i3 < 0 ), ( i3 + 9 ), i3) )) ), ( fmin(9, ( where(( i4 < 0 ), ( i4 + 9 ), i4) )) )) ) - 9 ) )}rf ], float

%kernel_math {
b7 = i3 < 0;
i5 = i3 + 9;
i9 = where(b7, i5, i3);
i11 = fmax(0, i9);
b15 = i4 < 0;
i13 = i4 + 9;
i17 = where(b15, i13, i4);
i19 = fmin(9, i17);
i21 = fmax(i11, i19);
T1_g[ ?S2{( ( ( -( fmax(0, ( where(( i3 < 0 ), ( i3 + 9 ), i3) )) ) ) + 9 ) + ( ( fmax(( fmax(0, ( where(( i3 < 0 ), ( i3 + 9 ), i3) )) ), ( fmin(9, ( where(( i4 < 0 ), ( i4 + 9 ), i4) )) )) ) - 9 ) )}rf ]
   = slice( T0_g[ iS0{9} ], { {i11, i21, 1} } )
}

T0_g[ iS0{9} ]
 root domain : (iS0{9})
 contiguity: f
 leaf domain : (iS0{9})
[...]

PR #1281, opened 2023-11-11
AliasAnalysis handles slices.

wujingyue:


wujingyue:
!build

wujingyue:
!build

wujingyue:
!build

csrc/optimization/alias_analysis.cpp
```diff
@@ -180,25 +251,94 @@ void AliasFinder::handle(const LoadStoreOp* permute) {
...
+
+void AliasFinder::handle(const SliceOp* slice) {
+  TensorView* in = slice->in();
+  TensorView* out = slice->out();
+
+  const std::vector<IterDomain*>& in_rfactor = in->getMaybeRFactorDomain();
+  const std::vector<IterDomain*>& out_rfactor = out->getMaybeRFactorDomain();
+
+  const auto rank = in_rfactor.size();
+  NVF_ERROR(rank == out_rfactor.size());
```
wujingyue, line 265: This apparently broke some tests. Looking...

naoyam:
Can this be split to two PRs, one for slices and another for reshapes?

wujingyue:
> Can this be split to two PRs, one for slices and another for reshapes?

Sure. Found more problems though: apparently, the pointwise scheduler changes the output contiguity, breaking `ResizeTest.SliceAndReshape2`. Looking...

wujingyue:
> > Can this be split to two PRs, one for slices and another for reshapes?
> 
> Sure. Found more problems though: apparently, the pointwise scheduler changes the output contiguity, breaking `ResizeTest.SliceAndReshape2`. Looking...

This is the apparent root cause: https://github.com/NVIDIA/Fuser/blob/2d2a2a47e11e96fc4582f6e1fb23cc45f38cdea0/csrc/tensor_view.cpp#L1210 doesn't copy allocation domain and contiguity.

wujingyue:
> > Sure. Found more problems though: apparently, the pointwise scheduler changes the output contiguity, breaking `ResizeTest.SliceAndReshape2`. Looking...
> 
> This is the apparent root cause:
> 
> https://github.com/NVIDIA/Fuser/blob/2d2a2a47e11e96fc4582f6e1fb23cc45f38cdea0/csrc/tensor_view.cpp#L1210
> 
> doesn't copy allocation domain and contiguity.

Will be fixed by #1291

wujingyue:
!build

csrc/optimization/alias_analysis.cpp
```diff
@@ -247,6 +248,81 @@ void AliasFinder::handle(const LoadStoreOp* permute) {
...
+    }
+
+    if (next_non_broadcast_is_non_contiguous) {
+      out_layout.contiguity[i] = false;
+      next_non_broadcast_is_non_contiguous = false;
+    } else {
+      out_layout.contiguity[i] = in_layout.contiguity[i];
+    }
+
+    std::vector<Expr*> dependencies = DependencyCheck::getAllExprsBetween(
```
jacobhinkle, line 312: You can use `StmtSort::getExprsBetween`.

jacobhinkle (approved):
This PR looks good to me. Note that a slice can also be achieved with a PadOp. In that case, we cannot assume the pads are negative, and if any are positive we must materialize. Otherwise we can alias just as you've done here.

wujingyue:
> This PR looks good to me. Note that a slice can also be achieved with a PadOp. In that case, we cannot assume the pads are negative, and if any are positive we must materialize. Otherwise we can alias just as you've done here.

Is the reverse also true? I.e. a `SliceOp` may take a negative offset, making it essentially a pad. That would break my analysis because a real pad has to be materialized.

jjsjann123:
Should I review this one or wait until we split the PR into two per Naoya's suggestion?

wujingyue:
> Should I review this one or wait until we split the PR into two per Naoya's suggestion?

Already split

csrc/optimization/alias_analysis.cpp
```diff
@@ -247,6 +248,81 @@ void AliasFinder::handle(const LoadStoreOp* permute) {
...
 }
 
+void AliasFinder::handle(const SliceOp* slice) {
+  TensorView* in = slice->in();
+  TensorView* out = slice->out();
+
+  const std::vector<IterDomain*>& in_rfactor = in->getMaybeRFactorDomain();
+  const std::vector<IterDomain*>& out_root = out->getRootDomain();
+  const std::vector<IterDomain*>& out_rfactor = out->getMaybeRFactorDomain();
+
```
jjsjann123, line 258: same here, we should check for allocationdomain on outputs. Either give us when we see anything, or double check it after the propagation that the preferred layout is consistent with it.

csrc/optimization/alias_analysis.cpp
```diff
@@ -247,6 +248,81 @@ void AliasFinder::handle(const LoadStoreOp* permute) {
...
+    if (!in_rfactor_to_out_root.count(in_allocation_id)) {
+      // `in_allocation_id` is a reduction product.
+      continue;
+    }
+    IterDomain* out_root_id = in_rfactor_to_out_root.at(in_allocation_id);
+    out_layout.allocation_domain.push_back(out_root_to_rfactor.at(out_root_id));
+  }
+
+  // Scan through the allocation domain in minor-to-major order. If an
+  // IterDomain is sliced, the next non-broadcast IterDomain has to be marked
```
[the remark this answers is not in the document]
jjsjann123: Agree with Jacob. i.e. if we later decide to merge a trivially sliced dimension, we should recognize that and merge it back I guess.

But this can only be done opportunistically (only handle static trivial slice), I guess?! Since padding value can be a runtime value and we can't know for sure that it'll be a trivial slice or not until runtime.

test/test_alias.cpp
```diff
@@ -285,4 +285,85 @@ TEST_F(AliasTest, ViewPermute) {
...
+      __LINE__,
+      __FILE__);
+}
+
+TEST_F(AliasTest, NotAllOutputsAlias) {
+  auto fusion = std::make_unique<Fusion>();
+  FusionGuard fg(fusion.get());
+
+  TensorView* in = makeContigConcreteTensor({2, 3});
+  TensorView* slice_out = slice(in, {0, 0}, {2, 2});
```
jjsjann123, line 347: nitpick, can we try a static trivial slice with a view to merge the sliced dimension?

jacobhinkle:
> > This PR looks good to me. Note that a slice can also be achieved with a PadOp. In that case, we cannot assume the pads are negative, and if any are positive we must materialize. Otherwise we can alias just as you've done here.
> 
> Is the reverse also true? I.e. a `SliceOp` may take a negative offset, making it essentially a pad. That would break my analysis because a real pad has to be materialized. 

No. Slice will never pad. See #460.

csrc/optimization/alias_analysis.cpp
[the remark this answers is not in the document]
wujingyue: > Is it possible to encounter a trivial SliceOp here? Where both left and right pad are zero? We should treat such case as the identity on that axis.

https://github.com/NVIDIA/Fuser/blob/2e2f3c7c3c07545cd673ac95174d124892a520e7/csrc/ops/alias.cpp#L792 and https://github.com/NVIDIA/Fuser/blob/2e2f3c7c3c07545cd673ac95174d124892a520e7/csrc/ops/alias.cpp#L772 have already done a good job filtering out trivial slices and trivial slice dimensions. So I'd rather keep the logic here simple.

wujingyue:
> This PR looks good to me. Note that a slice can also be achieved with a PadOp. In that case, we cannot assume the pads are negative, and if any are positive we must materialize. Otherwise we can alias just as you've done here.

Left a note on `AliasFinder:handle(const SliceOp* slice)` for future improvement.

csrc/optimization/alias_analysis.cpp
wujingyue: Mark this as resolved and leave the discussion to the other PR.

test/test_alias.cpp
wujingyue: Done.

jjsjann123 (approved):
LGTM

wujingyue:
!build

PR #1283, opened 2023-11-13
Modified `uniform` and `normal`'s Python APIs to fully support Dynamic Shape Definitions

kevinstephano:
I modified the two APIs that have a `shape` parameter to take either a `Vector`, `py::list`, or `py::tuple` such that the user can specify the shape as a combination of `Scalar`s, Python Integers, or a `Vector` in order to utilize the output of nvFuser's shape ops.  The API was partially already there by allowing the user to specify the output shape with just a List of `Scalar`s.  I also cleaned up the API such that each Random Distribution has its own Record Type instead of relying on distinguishing each type by a string.  I also made the code more modular so the same code is used to handle each distribution.

Old syntax:
```python
ops.uniform(minval: Scalar, maxval: Scalar, shape: List[Scalar], dtype: DataType, *, rng_seed: Scalar, rng_offset: Scalar) -> Tensor
ops.normal(mean: Scalar, std: Scalar, shape: List[Scalar], dtype: DataType, *, rng_seed: Scalar, rng_offset: Scalar) -> Tensor
```
New Syntax:
```python
ops.uniform(minval: Scalar, maxval: Scalar, shape: Union[Vector, List[Union[Scalar, int]]], *, rng_seed: Scalar, rng_offset: Scalar, dtype: DataType) -> Tensor
ops.normal(mean: Scalar, std: Scalar, shape: Union[Vector, List[Union[Scalar, int]]], *, rng_seed: Scalar, rng_offset: Scalar, dtype: DataType) -> Tensor
```

### DIFF
--- a/csrc/python_frontend/fusion_record.h
+++ b/csrc/python_frontend/fusion_record.h
@@ -2657,9 +2657,9 @@ struct RandomDistOpRecord : RecordFunctor {
     static_assert(
         (RType == serde::RecordType::NormalDistOp) ||
         (RType == serde::RecordType::UniformDistOp));
-    if constexpr (RType == serde::RecordType_UniformDistOp) {
+    if constexpr (RType == serde::RecordType::UniformDistOp) {
       name_ = "ops.uniform";
-    } else if constexpr (RType == serde::RecordType_NormalDistOp) {
+    } else if constexpr (RType == serde::RecordType::NormalDistOp) {
       name_ = "ops.normal";
     }
     setArgName(2, "shape");
@@ -2729,7 +2729,7 @@ struct RandomDistOpRecord : RecordFunctor {
   std::pair<serde::RecordData, flatbuffers::Offset<void>> recordData(
       flatbuffers::FlatBufferBuilder& builder) const final {
     return {
-        serde::RecordData_TensorCreationSymbolic,
+        serde::RecordData::TensorCreationSymbolic,
         serde::CreateTensorCreationSymbolic(builder, toUnderlying(dtype_))
             .Union()};
   }
--- a/csrc/python_frontend/fusion_record.h
+++ b/csrc/python_frontend/fusion_record.h
@@ -2645,28 +2645,35 @@ struct IotaOpRecord : RecordFunctor {
 };
 
 //! Specialized Record Functors for random ops.
-struct RandomOpRecord : RecordFunctor {
-  RandomOpRecord(
+template <serde::RecordType RType>
+struct RandomDistOpRecord : RecordFunctor {
+  RandomDistOpRecord(
       std::vector<State> _args,
       std::vector<State> _outputs,
-      std::string name,
-      serde::RecordType record_type,
       PrimDataType dtype)
       : RecordFunctor(
             std::move(_args),
             std::move(_outputs),
-            name,
-            record_type),
+            "",
+            RType),
         dtype_(dtype) {
+    static_assert(
+      (RType == serde::RecordType_NormalDistOp) ||
+      (RType == serde::RecordType_UniformDistOp));
+    if constexpr (RType == serde::RecordType_UniformDistOp) {
+      name_ = "ops.uniform";
+    } else if constexpr (RType == serde::RecordType_NormalDistOp) {
+      name_ = "ops.normal";
+    }
     setArgName(2, "shape");
     if (args_.size() == 5) {
       setArgName(3, "rng_seed");
       setArgName(4, "rng_offset");
     }
   }
-  ~RandomOpRecord() override = default;
+  ~RandomDistOpRecord() override = default;
   RecordFunctor* clone() final {
-    return new RandomOpRecord(*this);
+    return new RandomDistOpRecord(*this);
   }
 
   //! Child specific hash function in lower 32 bits.
@@ -2679,7 +2686,7 @@ struct RandomOpRecord : RecordFunctor {
 
   bool operator==(const RecordFunctor& other) const final {
     auto result = false;
-    if (auto child_ptr = dynamic_cast<const RandomOpRecord*>(&other)) {
+    if (auto child_ptr = dynamic_cast<const RandomDistOpRecord*>(&other)) {
       result = RecordFunctor::operator==(other);
       result = result && (dtype_ == child_ptr->dtype_);
     }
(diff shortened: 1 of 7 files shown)


kevinstephano:
!build

kevinstephano:
!build

kevinstephano:
!build

csrc/python_frontend/python_bindings.cpp
```diff
@@ -149,6 +149,46 @@ Tensor reshape_fn(
...
+template <class ShapeType, serde::RecordType RType>
+Tensor random_dist_op_fn(
+    FusionDefinition::Operators& self,
+    Scalar arg1,
+    Scalar arg2,
+    ShapeType generic_new_shape,
+    std::optional<Scalar> rng_seed,
+    std::optional<Scalar> rng_offset,
+    PrimDataType dtype) {
+  static_assert(
```
jjsjann123, line 161: nitpick. The assert here looks a bit strange.
RType is passed to RandomDistOpRecord and isn't used here. We can just rely on the assert in the template instead.

csrc/python_frontend/fusion_record.h
```diff
@@ -2645,107 +2645,80 @@ struct IotaOpRecord : RecordFunctor {
...
-      setArgName(2, "rng_seed");
-      setArgName(3, "rng_offset");
+    static_assert(
+        (RType == serde::RecordType_NormalDistOp) ||
+        (RType == serde::RecordType_UniformDistOp));
+    if constexpr (RType == serde::RecordType_UniformDistOp) {
+      name_ = "ops.uniform";
+    } else if constexpr (RType == serde::RecordType_NormalDistOp) {
+      name_ = "ops.normal";
+    }
```
jjsjann123, line 2663: maybe assert in the else block for an unrecognized template parameter, instead of the static_assert above.

csrc/python_frontend/python_bindings.cpp
```diff
@@ -149,6 +149,46 @@ Tensor reshape_fn(
...
+  FusionDefinition* fd = self.fusion_definition;
+  Vector new_shape = ShapeAsVector(generic_new_shape, *fd);
+
+  Tensor output = fd->defineTensor(new_shape.size);
+  std::vector<State> arg_states = {
+      fd->recordingState(arg1()),
+      fd->recordingState(arg2()),
+      fd->recordingState(new_shape()),
+  };
+  if (rng_seed.has_value()) {
```
jjsjann123, line 178: nitpick. should we raise error when rng_offset is provided but rng_seed is missing?

csrc/python_frontend/python_bindings.cpp
```diff
@@ -1954,6 +1994,48 @@ void initNvFuserPythonBindings(PyObject* module) {
   NVFUSER_PYTHON_BINDING_CAST_OP("cast", castOp)
 #undef NVFUSER_PYTHON_BINDING_CAST_OP
 
+#define NVFUSER_PYTHON_BINDING_RANDOM_DIST_OP( \
```
jjsjann123, line 1997: nitpick: maybe a nested macro to avoid boiler plate.

something like
```
#define NVFUSER_ALL_VECTOR_TYPES(fn, ...) \
  fn(Vector, __VA_ARGS__); \
  fn(py::list, __VA_ARGS__); \
  fn(py::tuple, __VA_ARGS__);

#define NVFUSER_RANDOM_DIST_OP_HELPER(vec_type, op_str, op_type, arg1_str, arg2_str)       \
  nvf_ops.def(                                 \
      op_str,                                  \
      random_dist_op_fn<vec_type, op_type>,      \
      py::arg(arg1_str),                       \
      py::arg(arg2_str),                       \
      py::arg("shape"),                        \
      py::kw_only(),                           \
      py::arg("rng_seed") = py::none(),        \
      py::arg("rng_offset") = py::none(),      \
      py::arg("dtype") = DataType::Float,      \
      py::return_value_policy::reference);

#define NVFUSER_PYTHON_BINDING_RANDOM_DIST_OP(...)       \
  NVFUSER_ALL_VECTOR_TYPES(NVFUSER_RANDOM_DIST_OP_HELPER, __VA_ARGS__)
```

jjsjann123 (commented):
minor comments on macro~

csrc/python_frontend/fusion_record.h
```diff
@@ -2645,107 +2645,80 @@ struct IotaOpRecord : RecordFunctor {
...
-            serde::RecordType_RandomOp),
-        output_shape_(std::move(output_shape)),
+      : RecordFunctor(std::move(_args), std::move(_outputs), "", RType),
         dtype_(dtype) {
-    if (args_.size() == 4) {
-      // seed and offset were provided in addition to the usual 2 arguments
-      setArgName(2, "rng_seed");
-      setArgName(3, "rng_offset");
+    static_assert(
+        (RType == serde::RecordType_NormalDistOp) ||
```
rdspring1, line 2657: Note: I updated the Enum syntax for serialization.
```suggestion
        (RType == serde::RecordType::NormalDistOp) ||
```

csrc/python_frontend/fusion_record.h
```diff
@@ -2645,107 +2645,80 @@ struct IotaOpRecord : RecordFunctor {
...
+        (RType == serde::RecordType_UniformDistOp));
```
rdspring1, line 2658:
```suggestion
        (RType == serde::RecordType::UniformDistOp));
```

rdspring1 (commented):
LGTM.

Just FYI, the serialization syntax for Enum types has changed.
It should be changed from `serde::RecordType_NormalDistOp` to `serde::RecordType::NormalDistOp`.
Ditto for `serde::RecordType_UniformDistOp`

kevinstephano:
> LGTM.
> 
> Just FYI, the serialization syntax for Enum types has changed. It should be changed from `serde::RecordType_NormalDistOp` to `serde::RecordType::NormalDistOp`. Ditto for `serde::RecordType_UniformDistOp`

I just figured out I needed to change all the syntax to match the new way. :-)

kevinstephano:
!build

csrc/python_frontend/python_bindings.cpp
kevinstephano: RType is a template parameter and this should be a compile time assert.   I believe there is a better way to do this with C++ 20 but I am not inclined to use C++20 until we officially decide to use it.

csrc/python_frontend/python_bindings.cpp
kevinstephano: Good idea!

csrc/python_frontend/fusion_record.h
kevinstephano: This should be covered by the compile time `static_assert` at the top of the function.

csrc/python_frontend/python_bindings.cpp
kevinstephano: Clever!

kevinstephano:
!build

kevinstephano:
!build

csrc/python_frontend/fusion_record.h
```diff
@@ -2646,107 +2646,80 @@ struct IotaOpRecord : RecordFunctor {
...
+      } else if (args_.size() == 5) { // provided seed and offset
+        auto seed = fd.getFusionState(args_.at(3).index);
+        auto offset = fd.getFusionState(args_.at(4).index);
         output = uniform(output_shape, arg1, arg2, dtype_, seed, offset);
       }
-    } else if (name_.compare("ops.normal") == 0) {
-      if (args_.size() == 2) { // stochastic normal
+    } else if constexpr (RType == serde::RecordType::NormalDistOp) {
+      if (args_.size() == 3) { // stochastic normal
         output = normal(output_shape, arg1, arg2, dtype_);
```
jacobhinkle, line 2710: Very minor, feel free to ignore: I guess we could factor this slightly more by replacing this and the corresponding `uniform` call with a method called `backendOp` or something that is specialized for the distribution type. Of course this is not an issue unless we plan to start adding more distributions.

jacobhinkle (approved):
LGTM

csrc/python_frontend/python_bindings.cpp
```diff
@@ -1955,6 +1996,35 @@ void initNvFuserPythonBindings(PyObject* module) {
...
+      py::kw_only(),                               \
+      py::arg("rng_seed") = py::none(),            \
+      py::arg("rng_offset") = py::none(),          \
+      py::arg("dtype") = DataType::Float,          \
+      py::return_value_policy::reference);
+
+#define NVFUSER_PYTHON_BINDING_RANDOM_DIST_OP(...) \
+  NVFUSER_ALL_VECTOR_TYPES(NVFUSER_RANDOM_DIST_OP_HELPER, __VA_ARGS__)
+
+  NVFUSER_PYTHON_BINDING_RANDOM_DIST_OP(
```
jjsjann123, line 2021: Thanks for cleaning up the macro!

jjsjann123 (approved):
stamping since all my concerns are resolved.

csrc/python_frontend/python_bindings.cpp
kevinstephano: I put the `static_assert` in the `else` statement.

csrc/python_frontend/fusion_record.h
kevinstephano: I also put `static_assert` in `else` statement.

csrc/python_frontend/fusion_record.h
kevinstephano: I think it would be fine to have a `random distribution op` in the codegen but I don't think this is the PR to make that change.

kevinstephano:
!build

PR #1287, opened 2023-11-13
Print fusion IR with transforms after concretization.

wujingyue:
This is useful for debugging pre-segmenter optimizations.


jacobhinkle (approved):
Thanks. I often wind up changing this during debugging :-D.

PR #1291, opened 2023-11-14
TransformReplay::replayCasP sets the right contiguity.

### LINKED ISSUE #1290: The pointwise scheduler changes contiguity of fusion outputs.
This happened when a fusion output is also used by other `Expr`s. This triggered https://github.com/NVIDIA/Fuser/blob/b33fcf9b86966b2032a15260c2497b0fdba816da/csrc/scheduler/utils.cpp#L1057, which failed to set up the right contiguity flags for the forked TensorView.

wujingyue:
Fixes #1290

### DIFF
--- a/csrc/transform_replay.cpp
+++ b/csrc/transform_replay.cpp
@@ -517,49 +517,17 @@ std::pair<TensorDomain*, size_t> TransformReplay::replayPasC(
     }
   }
 
-  if (!opt.replay_allocation) {
-    TensorDomain* replayed = IrBuilder::create<TensorDomain>(
-        producer->container(),
-        producer->getRootDomain(),
-        producer->getRFactorDomain(),
-        producer->getAllocationDomain(),
-        new_IDs,
-        producer->domain()->contiguity());
-    return {replayed, producer_pos};
-  }
+  NVF_ERROR(
+      !opt.replay_allocation,
+      "replayAllocation is not implemented yet for TransformReplay::replayPasC");
 
   TensorDomain* replayed = IrBuilder::create<TensorDomain>(
       producer->container(),
       producer->getRootDomain(),
       producer->getRFactorDomain(),
-      std::vector<IterDomain*>{},
+      producer->getAllocationDomain(),
       new_IDs,
       producer->domain()->contiguity());
-
-  if (consumer->hasAllocation()) {
-    auto replay_PasC = BestEffortReplay(
-        new_IDs,
-        consumer->getLeafDomain(),
-        root_map.mapConsumerToProducer(consumer->domain(), replayed));
-    const auto& c2p_map = replay_PasC.getReplay();
-    std::vector<IterDomain*> new_allocation_domain;
-    new_allocation_domain.reserve(consumer->getAllocationDomain().size());
-    for (auto id : consumer->getAllocationDomain()) {
-      auto it = c2p_map.find(id);
-      NVF_CHECK(
-          it != c2p_map.end(),
-          "Unable to replayPasC: can not map ",
-          id->toString(),
-          " in the allocation domain of consumer tensor ",
-          consumer->toString(),
-          " to producer tensor ",
-          producer->toString());
-      new_allocation_domain.push_back(it->second);
-    }
-    replayed->setAllocationDomain(
-        std::move(new_allocation_domain), consumer->getContiguity());
-  }
-
   return {replayed, producer_pos};
 }
 
--- a/csrc/transform_replay.cpp
+++ b/csrc/transform_replay.cpp
@@ -572,8 +572,9 @@ std::pair<TensorDomain*, size_t> TransformReplay::replayCasP(
 
   // If this is a reduction operation, we may call transform_replay on the same
   // tensor view. When this happens, just return thet target view.
-  if (consumer == producer)
+  if (consumer == producer) {
     return {consumer->domain(), consumer->nDims()};
+  }
 
   if (producer_pos < 0) {
     producer_pos += (int64_t)producer->nDims() + 1;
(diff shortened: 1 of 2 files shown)


wujingyue:
!build

wujingyue:
!build

wujingyue:
!build

csrc/transform_replay.cpp
```diff
@@ -829,7 +830,8 @@ std::pair<TensorDomain*, size_t> TransformReplay::replayCasP(
           consumer->toString());
       new_allocation_domain.emplace_back(it->second);
     }
-    replayed->setAllocationDomain(std::move(new_allocation_domain), true);
+    replayed->setAllocationDomain(
+        std::move(new_allocation_domain), producer->getContiguity());
```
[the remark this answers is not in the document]
zasdfgbnm: I think @naoyam's concern makes sense. I think in the loop above, we should do something like:
```C++
    for (auto [id, contig] : producer->getAllocationDomainAndContiguity()) {
      auto it = p2c_map.find(id);
      NVF_CHECK(
          it != p2c_map.end(),
          "Unable to replayCasP: can not map ",
          id->toString(),
          " in the allocation domain of producer tensor ",
          producer->toString(),
          " to consumer tensor ",
          consumer->toString());
      new_allocation_domain.emplace_back(it->second);
      new_contiguity.emplace_back(contig);
    }
```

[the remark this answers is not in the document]
wujingyue: Done.

csrc/transform_replay.cpp
```diff
@@ -554,9 +554,10 @@ std::pair<TensorDomain*, size_t> TransformReplay::replayPasC(
           consumer->toString(),
           " to producer tensor ",
           producer->toString());
-      new_allocation_domain.emplace_back(it->second);
+      new_allocation_domain.push_back(it->second);
     }
-    replayed->setAllocationDomain(std::move(new_allocation_domain), true);
+    replayed->setAllocationDomain(
+        std::move(new_allocation_domain), consumer->getContiguity());
```
wujingyue, line 560: I made this change as @jjsjann123 recommended. But I didn't manage to find a good way to test this. Unlike CasP which is used by `TensorView::cache*`, PasC is rarely used by the public API.

wujingyue: More specifically, if `producer` and `consumer` have different ranks (e.g., when they are the input and the output of a reduction), `replayed` will get a wrong allocation domain. Can this happen at all? If so, I don't even know what allocation domain to assign -- where do we place the size-1 reduction IterDomain? (I guess the place of the size-1 dimension doesn't matter much in practice, but we'll still have to make a reasonable decision...)

wujingyue:
!build

csrc/transform_replay.cpp
wujingyue: I'll look more into this tomorrow. I deliberately added a `NVF_CHECK` asserting producer's rfactor rank == consumer's root rank and noticed several tests including `DynamicTransformIssue418_CUDA` are failing. I can probably revise one of them to get a regression test...

zasdfgbnm: Should we follow the same protocol as I commented in https://github.com/NVIDIA/Fuser/pull/1291#discussion_r1393149562 for this?

jjsjann123: Xiang's comment makes sense to me.

replay on either direction should handle the same issue consistently.

wujingyue: This is apparently dead code... (I'm running our unit tests to verify) It appears that never call `TransformReplay::replayPasC` with `replayAllocation` on. Therefore, the function early exits at line 528.

wujingyue: All right. I removed the dead code so it's not a concern **at this moment**.

wujingyue:
!build

wujingyue:
Can someone review this again? It's blocking #1281, #1295 and #1321.

zasdfgbnm (approved):

wujingyue:
!build

csrc/transform_replay.cpp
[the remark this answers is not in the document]
naoyam: > I think @naoyam's concern makes sense. I think in the loop above, we should do something like:
> 
> ```c++
>     for (auto [id, contig] : producer->getAllocationDomainAndContiguity()) {
>       auto it = p2c_map.find(id);
>       NVF_CHECK(
>           it != p2c_map.end(),
>           "Unable to replayCasP: can not map ",
>           id->toString(),
>           " in the allocation domain of producer tensor ",
>           producer->toString(),
>           " to consumer tensor ",
>           consumer->toString());
>       new_allocation_domain.emplace_back(it->second);
>       new_contiguity.emplace_back(contig);
>     }
> ```

@zasdfgbnm Hmm, what about consumer broadcast domains, like the case as I mentioned above? `replayCasP` will ignore such consumer broadcast domains, so they should just remain in the leaf domain. It won't affect the allocation, so conceptually it should have no effect but mechanically aren't we assuming root broadcast domains remain in the allocation domain?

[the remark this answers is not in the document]
wujingyue: Fortunately, `TransformReplay::replayCasP` is **currently** only used in `cacheBefore`, `cacheAfter` and `cacheFork`, so `consumer` is always a `set` of `producer`. The only tricky case at this moment is when `producer` is the product of a reduction, which I added a test for. 

However, your concern is valid. I actually don't know what the right behavior should be. As you said, the allocation order of consumer broadcast domains is ambiguous.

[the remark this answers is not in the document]
naoyam: Can you please add an assertion just in case?

Or, actually, I *think* we can just put a broadcast contiguity value in any position as it doesn't matter for the allocation. We would just need it for the consistency of the allocation domain.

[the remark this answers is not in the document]
wujingyue: > Can you please add an assertion just in case?

Done.

PR #1292, opened 2023-11-14
Move implementation to cpp.

wujingyue:
This saves build time when we change only the implementation.


zasdfgbnm (approved):
Moving implementation to cpp is not always good. Because compiler will be unable to inline it, which introduces more function call. For this specific case, because the constructor is only used once for each lowering, I think the perf is not an issue.

PR #1295, opened 2023-11-14
AliasAnalysis handles non-packed reshapes.

wujingyue:
Partially taken from #1281 to keep that PR small.

### DIFF
--- a/csrc/optimization/alias_analysis.cpp
+++ b/csrc/optimization/alias_analysis.cpp
@@ -197,12 +197,6 @@ void AliasFinder::handle(const LoadStoreOp* permute) {
     return;
   }
 
-  // Another lazy move: we could check compatibility and only give up when
-  // the allocation domain is incompatible with what we prefer for aliasing.
-  if (out->hasAllocation()) {
-    return;
-  }
-
   TensorView* in = permute->in()->as<TensorView>();
   // Look at the preferred layout not `in`'s current layout.
   Layout in_layout = analysis_.preferredLayout(in);
@@ -253,8 +247,8 @@ void AliasAnalysisResult::add(
     const TensorView* alias,
     const TensorView* source,
     Layout&& layout) {
-  auto [i, inserted] =
-      alias_to_source_.emplace(alias, std::make_pair(source, layout));
+  auto [i, inserted] = alias_to_source_.emplace(
+      alias, std::make_pair(source, std::move(layout)));
   NVF_ERROR(
       inserted,
       "The current implementation of alias analysis shouldn't find two sources for an alias. However, it's trying to make ",
--- a/csrc/optimization/alias_analysis.cpp
+++ b/csrc/optimization/alias_analysis.cpp
@@ -13,74 +13,13 @@
 #include <ir/interface_nodes.h>
 #include <ir/internal_base_nodes.h>
 #include <ir/utils.h>
+#include <linked_hash_map.h>
 #include <optimization/alias_analysis.h>
+#include <root_domain_map.h>
 
 namespace nvfuser::optimization {
 
 namespace {
-
-bool isContiguous(const std::vector<std::optional<bool>>& contiguity) {
-  for (const auto& id_contiguity : contiguity) {
-    // We skip std::nullopt contiguity. It represents a broadcast or reduction
-    // dimension, which is of size 1 and always contiguous.
-    if (id_contiguity.has_value() && id_contiguity.value() == false) {
-      return false;
-    }
-  }
-  return true;
-}
-
-// Returns whether the input TensorView is allocated contiguously.
-bool isContiguous(const TensorView& tv) {
-  return isContiguous(tv.getContiguity());
-}
-
-// Whether a ViewOp transforms any expanded broadcast IterDomain in the input.
-// This is a corner case in which we can't always turn `out` into an alias.
-//
-// For example, given
-//
-//   t0 = makeContigConcreteTensor({4, 5});
-//   t1 = broadcast(t0, {false, false, true});
-//   t2 = expand(t1, {4, 5, 6});
-//
-// `reshape(t2, {40, 3})` and `reshape(t2, {4, 30})` because both merge the
-// expanded broadcast IterDomain (6) or a subspace of it with preceding
-// IterDomains.  However, the output of `reshape(t2, {20, 6})` can simply be an
-// alias because the expanded broadcast IterDomain is forwarded not transformed.
-//
-// As a future improvement, when an expanded broadcast dimension is only split,
-// the output of the reshape can be an alias. However, nvFuser currently decides
-// to materialize the expansion, making the output not an alias (#1126).
-//
-// Obviously, this function assumes `in` and `out` are the input and output
-// TensorView of the same ViewOp.
-bool transformsExpandedBroadcastIterDomain(TensorView* in, TensorView* out) {
-  const std::vector<IterDomain*>& in_rfactor = in->getMaybeRFactorDomain();
-  const std::vector<IterDomain*>& out_root = out->getRootDomain();
-  const std::vector<IterDomain*>& out_rfactor = out->getMaybeRFactorDomain();
-
-  std::unordered_set<Val*> expanded_broadcast_dims;
-  for (size_t i = 0, size = in_rfactor.size(); i < size; i++) {
-    IterDomain* id = in_rfactor[i];
-    if (id->isBroadcast() && id->hasExpandedExtent()) {
-      expanded_broadcast_dims.insert(out_root[i]);
-    }
-  }
-
-  const std::vector<Expr*> transforms = DependencyCheck::getAllExprsBetween(
-      {out_root.begin(), out_root.end()},
-      {out_rfactor.begin(), out_rfactor.end()});
-
-  for (const auto* transform : transforms) {
-    for (Val* input : transform->inputs()) {
-      if (expanded_broadcast_dims.count(input)) {
-        return true;
-      }
-    }
-  }
-  return false;
-}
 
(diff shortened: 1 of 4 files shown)


wujingyue:
!build

wujingyue:
!build

csrc/optimization/alias_analysis.cpp
```diff
@@ -103,27 +42,151 @@ class AliasFinder : public OptOutConstDispatch {
...
+    IterDomain* out_root_id = in_rfactor_to_out_root.at(in_allocation_id);
+    allocation_to_contiguity.pushBack(out_root_id, in_layout.contiguity[i]);
+  }
+
+  // Replay `Expr`s from `out`'s root to `out`'s rfactor on `out`'s root.
+  // Stop when an `Expr` requires a data copy; otherwise generate the allocation
+  // order of `out`'s rfactor domain and the corresponding contiguity flags.
+  for (Expr* transform : DependencyCheck::getAllExprsBetween(
+           {out_root.begin(), out_root.end()},
+           {out_rfactor.begin(), out_rfactor.end()})) {
```
jacobhinkle, line 153: You can use `StmtSort::getExprsBetween` for this.

csrc/optimization/alias_analysis.cpp
```diff
@@ -103,27 +42,151 @@ class AliasFinder : public OptOutConstDispatch {
...
+      const auto [mergeable, contiguity] = mergeContiguity(
+          merge->outer(), outer_contiguity, merge->inner(), inner_contiguity);
+      if (!mergeable) {
+        return;
+      }
+      allocation_to_contiguity.insert(merge_i, merge->out(), contiguity);
+    } else {
+      NVF_ERROR(
+          false, "Expect Split or Merge, but found: ", transform->toString());
+    }
```
jacobhinkle, line 181: The other transform you might expect to find here is a Resize. When there is a resize we are either padding or truncating an axis. I think padding will always require materialization. Truncation followed by a merge with something other than broadcast will require materialization also. If you truncate then split that is fine, but if those split IDs later merge with IDs from another ID then we need to materialize I think.

jacobhinkle (commented):
Some initial comments.

csrc/optimization/alias_analysis.cpp
jacobhinkle: Note that in general we can't tell the sign of the pad values in a Resize. If we can prove both pads are non-positive and there is no downstream merge that has a dependency other than the Resize input we don't have to materialize. This would be the case for a simple slice.

jacobhinkle: Nevermind! I forgot this is a `ViewOp` so it's correct as is. No need to look for `Resize` here. I'll look at #1281 for that.

csrc/optimization/alias_analysis.cpp
```diff
@@ -103,27 +42,151 @@ class AliasFinder : public OptOutConstDispatch {
...
+std::pair<bool, std::optional<bool>> mergeContiguity(
+    const IterDomain* outer_id,
+    const std::optional<bool>& outer_contiguity,
+    const IterDomain* inner_id,
+    const std::optional<bool>& inner_contiguity) {
+  // Statuses `b` and `e` are represented in the IR with isBroadcast() and
+  // hasExpandedExtent(). Status `C` means stops propagating because we know we
+  // can't alias at that point.
+  //
+  // o\i | t  f  b  e
```
jjsjann123, line 79: Nice comment!

jjsjann123 (approved):
minor comments on some c++ nitpick. lgtm overall~~~

csrc/optimization/alias_analysis.cpp
```diff
@@ -103,27 +42,151 @@ class AliasFinder : public OptOutConstDispatch {
...
+}
+
 void AliasFinder::handle(const ViewOp* view) {
   TensorView* in = view->in();
   TensorView* out = view->out();
 
+  const std::vector<IterDomain*>& in_rfactor = in->getMaybeRFactorDomain();
+  const std::vector<IterDomain*>& out_root = out->getRootDomain();
+  const std::vector<IterDomain*>& out_rfactor = out->getMaybeRFactorDomain();
+
```
jjsjann123, line 124: I think we forgot to check the AllocationDomain on `out` here.

jjsjann123 (changes requested):
Only realized it when I happen to see the `hasAllocation` check on `LoadStoreOp`

csrc/kernel.cpp
```diff
@@ -10,6 +10,7 @@
 #include <expr_evaluator.h>
 #include <instrumentation.h>
 #include <ir/iostream.h>
+#include <ir/utils.h>
```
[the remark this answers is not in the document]
wujingyue: I separated this change to a different PR.

csrc/optimization/alias_analysis.cpp
```diff
@@ -167,38 +230,39 @@ void AliasFinder::handle(const LoadStoreOp* permute) {
...
 } // namespace
 
 void AliasAnalysisResult::add(
     const TensorView* alias,
     const TensorView* source,
-    const Layout& layout) {
-  std::pair<const TensorView*, Layout>& old_source = alias_to_source_[alias];
+    Layout&& layout) {
+  auto [i, inserted] =
+      alias_to_source_.emplace(alias, std::make_pair(source, layout));
```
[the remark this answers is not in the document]
wujingyue: Done. I'm surprised that https://clang.llvm.org/extra//clang-tidy/checks/cppcoreguidelines/rvalue-reference-param-not-moved.html didn't kick in. 🤷

csrc/optimization/alias_analysis.cpp
wujingyue: Credits go to @jacobhinkle !

csrc/optimization/alias_analysis.cpp
wujingyue: Good catch -- we should at least be consistent. We chatted about this offline. I have been on the fence whether AliasAnalysis or MarkAliasPass should respect existing allocation domains. For implementation convenience, I'm leaning towards letting the user resolve incompatibility. I added a comment around `findAliases` and adjusted some tests to reflect that.

csrc/optimization/alias_analysis.cpp
wujingyue: What's the benefit of that? I also saw `Dependencies::getAllExprs`. I'm confused when to use which...

wujingyue:
!build

csrc/optimization/mark_alias.cpp
```diff
@@ -17,6 +17,13 @@ void MarkAliasPass::runPass(Fusion* fusion) {
   const AliasAnalysisResult alias_analysis = findAliases(fusion);
   for (TensorView* out :
        ir_utils::filterByType<TensorView>(fusion->outputs())) {
+    // Lazy move: we could check compatibility and only give up when
```
jjsjann123, line 20: :laughing:  we are moving a lazy move around

csrc/optimization/mark_alias.cpp
```diff
@@ -17,6 +17,13 @@ void MarkAliasPass::runPass(Fusion* fusion) {
...
+    // the allocation domain is incompatible with what we prefer for
+    // aliasing.
+    if (out->hasAllocation()) {
+      continue;
```
jjsjann123, line 24: nitpick: Can we add a debug print here explaining why we are ditching the aliasing.

jjsjann123 (approved):
minor comment. LGTM

csrc/optimization/mark_alias.cpp
wujingyue: Yep!

wujingyue:
!build

csrc/optimization/alias_analysis.cpp
wujingyue: @jacobhinkle I merged my PR but would still appreciate your thought on this. I found at least three functions doing the similar work :) I'm not sure when to use which.

csrc/optimization/mark_alias.cpp
wujingyue: Yep, "like a rolling stone" :)

csrc/optimization/alias_analysis.cpp
jacobhinkle: Great question. It seems to me like `DependencyCheck` and `StmtSort` are redundant; they are used roughly equally often and appear to provide the same functionality, with `DependencyCheck` maybe slightly more complicated. `DependencyCheck` has been around since the beginning of the project, while `StmtSort` evolved from `ExprSort` (https://github.com/csarofeen/pytorch/pull/1376). `ExprSort` was introduced in https://github.com/pytorch/pytorch/pull/45218/files#diff-49787cacb3e793473c24bc615113fb2c782086cbe552c63ac75ffede45f417a1R240-R254. I have been mostly using `StmtSort` which has options for controlling whether we traverse members and attributes. We should make clear in the comments when to use each. CC @naoyam @jjsjann123

naoyam: IIUC, `Dependencies` has been used from a very beginning of the project, and later `IterVisitor::traverseBetween` and `StmtSort` were added. I believe `StmtSort` basically reproduces the same functionality as `Dependencies` but more concisely using `IterVisitor::traverseBetween`. I think it should be valid to replace all `DependencyCheck::getAllExprs/ValsBetween` with `StmtSort::getExprs/Vals`, but should be confirmed by @csarofeen

csarofeen: Yes, that's correct.

PR #1299, opened 2023-11-14
Evaluate method for `FullOp`, `EyeOp`, `IotaOp`.

Priya2698:


Priya2698:
!build

Priya2698:
!build

csrc/ir/nodes.cpp
```diff
@@ -67,6 +67,27 @@ std::string FullOp::toInlineString(int indent_size) const {
...
+  for (auto i : c10::irange(inputs.size() - 1)) {
+    shape.push_back((int)inputs.at(i));
+  }
+  DataType dtype = getFillValue()->getDataType().value();
+  const auto options = at::TensorOptions().device(at::kCUDA);
+  if (dtype == DataType::ComplexFloat) {
+    return {at::full(shape, (c10::complex<float>)inputs.back(), options)};
+  } else if (dtype == DataType::ComplexDouble) {
+    return {at::full(shape, (c10::complex<double>)inputs.back(), options)};
+  } else {
```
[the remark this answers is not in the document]
Priya2698: Re: conversion from `PolymorphicValue` to `c10::Scalar` -- Something similar to https://github.com/NVIDIA/Fuser/blob/31573bccf7fb224feb87a90685eaaff4ead75838/csrc/polymorphic_value.h#L336?

csrc/polymorphic_value.h
```diff
@@ -373,6 +373,21 @@ inline PolymorphicValue toTensor(
       false, "PolymorphicValue toTensor not implemented for ", x.type().name());
 }
 
+// Convert PolymorphicValue to c10::Scalar
+inline c10::Scalar toScalar(const PolymorphicValue& x) {
```
zasdfgbnm, line 377: Can this be just
```C++
if (x.is<std::complex<double>>()) {
  return (c10::complex<double>)x;
} else {
  return x;
}
```

Priya2698: That gives a conversion error from `PolymorphicValue` to `c10::Scalar`

zasdfgbnm: Does this work?
```C++
if (x.is<std::complex<double>>()) {
  return (c10::complex<double>)x;
} else {
  return (c10::Scalar)x;
}
```

Priya2698: Yes, that works. Although in that case, we will not be checking if the input can actually be converted to a Scalar (i.e, if x is one of `at::Tensor, <std::vector>, StructHandle, Pointer, Opaque`) so I prefer the explicit checks. We can keep this implementation and be careful about the usage as well.

Additionally, can we make this more general? It would be useful in the `TernaryOp where` when the condition input is a tensor instead of a scalar. Right now, the `evaluate` method does not work for tensors: Example test: https://github.com/NVIDIA/Fuser/blob/e30b6bbc353f1595c8b5c3337626b8a90c80b16b/test/test_gpu_view.cpp#L2177
In these cases, `at::where` expects the second and third argument to either be a tensor or scalar.
```
inline std::variant<at::Tensor, c10::Scalar> toTensorOrScalar(const PolymorphicValue& x)
```

zasdfgbnm: Sounds good, let's do it!

Priya2698: Sticking to `toScalar` since `at::where` will require explicit conversion to either a tensor/scalar so the evalaute method will be just as verbose. Will modify `toScalar` if needed in a separate PR.

Priya2698:
!build

zasdfgbnm (approved):

PR #1303, opened 2023-11-14
Refactor ldmatrix and mma input scheduling

zasdfgbnm:
This PR is stacked on https://github.com/NVIDIA/Fuser/pull/1311

Similar to https://github.com/NVIDIA/Fuser/pull/1210, this PR schedules the allocation domain of mma inputs and ldmatrix. Similar to the situation of https://github.com/NVIDIA/Fuser/pull/1210, the modified piece of code was initially implemented prior to the invention of allocation, therefore there were some hacks around there.

For the case of ldmatrix, if you look at the official doc https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#warp-level-matrix-instructions-ldmatrix, you will find that, there are basically two correct but different schedules:
1. If you look at the memory format of which thread own which part of the tensor, from this information, you will able to derive one schedule. Let's call this memory-layout-based schedule.
2. If you look at which thread should pass the address of which item as an operand to this instruction, from this information, you will be able to derive another schedule. Let's call this indexing-based schedule.

Unfortunately, these two schedules does not match. Prior to this PR, we were uniquely using the indexing-based schedule. This is fine with ldmatrix, but the index of mma inputs is basically incorrect, and therefore a hacked index was used.

I believe the most natural way to implement these two separate schedules is as follows: Assume we have a fusion:
```C++
Tregister = ldmatrix(Tsmem)
Tregister2 = broadcast(Tregister)
... = mma(Tregister2, ...)
```
then the allocation domain of `Tregister2` and `Tregister` must be scheduled as memory-layout-based schedule, the leaf domain of `Tregister` must be scheduled as the indexing-based schedule. The leaf domain of `Tregister2` should be scheduled similar to memory-layout-based schedule.

In this PR, I refactored the mma swizzler for mma inputs to implement the above design. In the refactored code, `scheduleTuringOperandRead` schedules the memory-layout-based schedule, and `scheduleLdMatrix` starts from the memory-layout based schedule, and generates the indexing-based schedule on top of it.

Due to this change, the codegen of `MmaOp` no longer needs special handling the index, it is now just a naive:
```C++
  void handle(const MmaOp* mma) final {
    indent() << genMmaOp(mma) << "(\n";
    indent() << kTab << gen(mma->out()) << ",\n";
    indent() << kTab << gen(mma->inA()) << ",\n";
    indent() << kTab << gen(mma->inB());
    code_ << ");\n";
  }
```

However, our current sync analysis and indexing infrastructure is not capable of analyzes this style of scheduling, and therefore, I would need to add additional hacking points to them:

- In our sync analysis, currently, it always assume that the parallelization of the leaf domain determines which thread owns which item. With the allocation domain support, it is actually the parallelization of the *allocation* domain determines which thread owns which item. In a perfect world, for exprs other than ldmatrix, warp shuffling, and mma op, the analysis based on leaf domain should match with the analysis based on the allocation, because, for example, if you do `y = sin(x)`, there is no way that `y` get the result from other threads, but our system should not assume that the analysis based on leaf domain always match with the analysis based on the allocation generally. Unfortunately, at the current state, our system can not handle this correctly. So I added one hack to it: if the expr is ldmatrix, assume the schedule is correct and just give up sync analysis. I think for now, this is a valid solution, because we will be rewriting our sync analysis with `IdModel` anyway.
- In our indexing of `MmaOp` producer, we are incorrectly marking some of allocation domain as zero domain, where it should not be. Similar to the above point, the mma op implies warp shuffling of data. During indexing, it does make sense to replay consumer as producer, but only to the block level. Within a warp, it makes no sense to replay the producer as consumer. I think we should carefully think about the design when we are rewriting our indexing with `IdModel`, but for this PR, I just manually set the last three `IterDomain`s of mma input as special and not treat them as zero domain.

I believe this PR is an improvement compared to the previous approach, because it has less special handling in the schedule itself, but I still think we are far from supporting ldmatrix and mma ops elegantly. In the future, I think we should:
1. Reconsider the design of different domains, leaf vs allocation, etc. and figure out what is the best way to schedule ldmatrix.
2. When we rewriting, we should keep the indexing of ldmatrix and mma in mind, and make sure that it can be supported without any hack in the new system.


csrc/codegen.cpp
```diff
@@ -1098,60 +1099,11 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
     return ss.str();
   }
 
-  static int getInputARegisterSize(MmaOptions::MacroType macro) {
```
zasdfgbnm, line 1101: Moved to `csrc/device_lower/pass/index.cpp`

zasdfgbnm:
In this PR, I didn't update the scheduling of Volta mma, because I don't think it worth my time. Let's just remove Volta completely: https://github.com/NVIDIA/Fuser/pull/1334

zasdfgbnm:
!build

zasdfgbnm:
also cc @wujingyue @jjsjann123 who might be interested in discussing allocation domain design.

csrc/device_lower/validation.cpp
```diff
@@ -876,10 +876,14 @@ namespace {
 //!  specialization of tidx as lane id.
 void validateMmaTensors(MmaOp* mma) {
   bool tidx_validated = false;
-  std::vector<TensorView*> to_validate = {
-      mma->inA()->as<TensorView>(),
-      mma->inB()->as<TensorView>(),
-      mma->out()->as<TensorView>()};
+  std::vector<TensorView*> to_validate = {mma->out()->as<TensorView>()};
```
wujingyue, line 879: It appears to me that `MmaOp::out`, `MmaOp::inA` and `MmaOp::inB` always return a `TensorView`. I suggest (in a separate PR) we make the return type `TensorView*` instead of `Val*`.

wujingyue (commented):
I like the general idea -- it makes a lot of sense to separate memory schedules and index schedules, and allocation domain is designed for that.

csrc/device_lower/validation.cpp
zasdfgbnm: They can actually be `kir::TensorIndex` as well, for example, in https://github.com/NVIDIA/Fuser/blob/d243121980debe620caea898801336928ea4a121/csrc/codegen.cpp#L1151, it should return a `kir::TensorIndex`. Generally, for tensor op, there in order to know if its input or output is always a `TensorView`, you can look at https://github.com/NVIDIA/Fuser/blob/d243121980debe620caea898801336928ea4a121/csrc/device_lower/pass/fusion_simplifier.cpp#L20 and https://github.com/NVIDIA/Fuser/blob/main/csrc/device_lower/pass/index.cpp, if it is replaced with a `LoadStoreOp` in `UnaryOpInserter` or if it is lowered as a `LoadStoreOp` in indexing lowering, then the operands and outputs of such an expr must be `TensorView`. Otherwise, it can also be `kir::TensorIndex`.

zasdfgbnm:
!build

zasdfgbnm:
!build

csrc/scheduler/matmul.cpp
```diff
@@ -974,9 +978,23 @@ void scheduleMatmul(Fusion* fusion, const MatmulParams& params) {
...
+
+    //  -5  -4   -3   -2   -1          or          -5  -4   -3   -2   -1
+    //[8mi, 4k, 2ko, 2mo, 2ki]                   [8ni, 4k, 2ko, 1no, 2ki]
+    for (auto tv : {ab, bb}) {
+      tv->merge(-5);
+      tv->axis(-4)->parallelize(ParallelType::TIDx);
+    }
+    propagate_mma_input_schedule_to(acr, bcr);
+  } else {
+    acr->axis(-1)->parallelize(ParallelType::Vectorize);
```
drzejan2, line 995: With recent removal of Volta support, the `else` block won't called, at least until Hopper support for MMA is added. Perhaps, for the time being, we can replace content with:
```
NVF_ERROR(false, "Unsupported MMA macro, only Turing and Ampere architectures are currently supported.");
```

csrc/scheduler/mma_utils.h
```diff
@@ -170,12 +170,44 @@ class WarpMmaSwizzler {
...
+  //!   +--------+--------+
+  //!   |        |        |
+  //!   |  8x8   |  8x8   |
+  //!   |        |        |
+  //!   +--------+--------+
+  //! If mn_major is true, these 8x8 matrices are visited in the order of:
+  //! top left -> top right -> bottom left -> bottom right.
+  //! If mn_major is false, these 8x8 matrices are visited in the order of:
+  //! top left -> bottom left -> top right -> bottom right.
+  static void scheduleLdMatrix(TensorView* tv, bool mn_major = false);
```
drzejan2, line 203: Based on the changes in `matmul.cpp`, the `mn_major` swizzle pattern is applied for loading `B` input of `MmaOp`.

Can you briefly share the background for this, or it will be enough to revisit the links you shared in the documentation?

drzejan2 (approved):
Changes look good, I mostly asked about some additional clarifications of the proposed changes.

csrc/scheduler/mma_utils.cpp
```diff
@@ -584,16 +584,11 @@ void WarpMmaSwizzler::scheduleMmaWarpOutput(
...
-        setWarpMapped(tv, 4);
-      }
-      break;
     case MmaOptions::MacroType::Turing_16_16_16:
     case MmaOptions::MacroType::Ampere_16_16_16:
-      scheduleTuringM16N16K16MmaWarpOutput(tv, options);
+      scheduleTuringMmaWarpOutput(tv, options);
       if (tv->definition()->isA<MmaOp>()) {
-        setWarpMapped(tv, 4);
+        setWarpMapped(tv, 7);
```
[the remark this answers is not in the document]
zasdfgbnm: Actually no. The `setWarpMapped` does nothing but prevent accidental editing. It just says "the innermost 7 dims are done scheduling, don't touch it."

test/test_gpu_tensorcore.cpp
```diff
@@ -693,7 +709,7 @@ TEST_F(NVFuserTest, FusionAmpereMatmul_CUDA) {
   // Keep multiples of 8 to keep vectorizable.
   int M = 504, N = 136, K = 248;
 
-  for (auto layout : kAllSupportedMatmulLayout) {
+  for (auto layout : {MatmulLayout::TN}) {
```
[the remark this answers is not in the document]
zasdfgbnm: Oh sorry. I changed this for convenient debugging, but forget to change them back...

csrc/scheduler/matmul.cpp
zasdfgbnm: removed this branch completely

csrc/scheduler/mma_utils.h
zasdfgbnm: This is a very good question. I updated the code to add a paragraph of comments explaining the situation. Cross posting it here:
```C++
  //! In principle, only `mn_major = false` should be needed. But unfortunately,
  //! we are taking advantage of the ldmatrix large load in a pretty hacky way.
  //! For example, for Turing, only m16n8k8 is supported by hardware. But we are
  //! also using a fake m16n8k16 and m16n16k16, which uses a single large
  //! ldmatrix to load data to register, and run multiple mma instructions to
  //! consume these data. In the future, we should only keep the m16n8k8 macro,
  //! and schedule m16n8k16 and m16n16k16 more correctly than this current way.
```

zasdfgbnm:
!build

csrc/scheduler/mma_utils.h
drzejan2: Thank you for adding this piece of comment, now everything makes sense!

PR #1310, opened 2023-11-16
Duplicated outputs patch

### LINKED ISSUE #1305: Fails when multiple outputs: RuntimeError: group_outputs.size() == group_runtime_outputs.size() INTERNAL ASSERT FAILED Output size does not match.
To reproduce:
```
import torch

    S0 = fd.define_scalar(1.00000, dtype=DataType.Double)
    T1 = fd.ops.full(fill_value=S0, shape=[4, 4], dtype=DataType.BFloat16)
    fd.add_output(T1)
    fd.add_output(T1)

    nvfuser_fusion_id0(fd)

inputs = [
]
fd.execute(inputs)

```

Error message:
```
  File "/opt/pytorch/nvfuser/nvfuser/__init__.py", line 123, in execute
    result = self._execute(
RuntimeError: group_outputs.size() == group_runtime_outputs.size() INTERNAL ASSERT FAILED at "/opt/pytorch/nvfuser/cs
rc/kernel_cache.cpp":191, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues.
 Output size does not match.
Exception raised from addOutputsToArgsAndTensorMap at /opt/pytorch/nvfuser/csrc/kernel_cache.cpp:191 (most recent cal
l first):
frame #0: nvfuser::nvfCheckFail(char const*, char const*, unsigned int, std::string const&) + 0x8d (0x7f37f120d901 in
 /opt/pytorch/nvfuser/nvfuser/lib/libnvfuser_codegen.so)
frame #1: nvfuser::nvfErrorFail(char const*, char const*, unsigned int, char const*, std::string const&) + 0x53 (0x7f
37f134acc3 in /opt/pytorch/nvfuser/nvfuser/lib/libnvfuser_codegen.so)
frame #2: nvfuser::FusionKernelRuntime::compileFusionParallel(nvfuser::KernelArgumentHolder) + 0x1299 (0x7f37f1499cb9
 in /opt/pytorch/nvfuser/nvfuser/lib/libnvfuser_codegen.so)
frame #3: nvfuser::FusionExecutorCache::runFusionWithInputs(c10::ArrayRef<c10::IValue> const&, std::optional<nvfuser:
:PrimDataType>, std::optional<signed char>) + 0xac3 (0x7f37f14a46d3 in /opt/pytorch/nvfuser/nvfuser/lib/libnvfuser_co
degen.so)
frame #4: nvfuser::python_frontend::FusionDefinition::execute(c10::ArrayRef<c10::IValue> const&, bool, bool, std::opt
ional<signed char>) const + 0x161 (0x7f37f1792c51 in /opt/pytorch/nvfuser/nvfuser/lib/libnvfuser_codegen.so)
frame #5: <unknown function> + 0xef31e (0x7f37f1ad431e in /opt/pytorch/nvfuser/nvfuser/_C.cpython-310-x86_64-linux-gn
u.so)
frame #6: <unknown function> + 0x170227 (0x7f37f1b55227 in /opt/pytorch/nvfuser/nvfuser/_C.cpython-310-x86_64-linux-g
nu.so)

```

jjsjann123:
Fixes #1305 
Patches duplicated outputs handling in segmenter.
Adding tests to verify proper behavior with aliasing handling for duplicated outputs.

### DIFF
--- a/csrc/fusion_segmenter.cpp
+++ b/csrc/fusion_segmenter.cpp
@@ -1554,7 +1554,8 @@ std::unique_ptr<Fusion> SegmentedFusion::makeFusion(SegmentedGroup* sg) {
     }
   }
 
-  // note, we would want to keep output consistent and not artificially drop duplicates.
+  // note, we would want to keep output consistent and not artificially drop
+  // duplicates.
   for (auto out : sg->output_vals) {
     fusion_segment->addOutput(complete_to_segment_map.clone(out));
   }
--- a/csrc/fusion_segmenter.cpp
+++ b/csrc/fusion_segmenter.cpp
@@ -1554,7 +1554,8 @@ std::unique_ptr<Fusion> SegmentedFusion::makeFusion(SegmentedGroup* sg) {
     }
   }
 
-  for (auto out : getAllOutputs(sg)) {
+  // note, we would want to keep output consistent and not artificially drop duplicates.
+  for (auto out : sg->output_vals) {
     fusion_segment->addOutput(complete_to_segment_map.clone(out));
   }
 
--- a/test/test_alias.cpp
+++ b/test/test_alias.cpp
@@ -292,11 +292,11 @@ TEST_F(AliasTest, DuplicateOutputs) {
         !fec.getMostRecentKernelRuntime()->isSegmented(),
         "segmentation is not supposed to happen");
 
-    at::Tensor out_tensor = in_tensor.add(3.141);
+    at::Tensor expected_out_tensor = in_tensor.add(3.141);
     // Verify output values.
     testValidate(
         fec.fusion(),
-        {out_tensor, out_tensor},
+        {expected_out_tensor, expected_out_tensor},
         {in_tensor},
         __LINE__,
         __FILE__);
--- a/test/test_alias.cpp
+++ b/test/test_alias.cpp
@@ -285,8 +285,12 @@ TEST_F(AliasTest, DuplicateOutputs) {
     at::Tensor out_tensor_0 = out_tensors[0];
     at::Tensor out_tensor_1 = out_tensors[1];
 
-    // TODO: Verify aliasing among duplicated outputs
-    // TODO: Verify no segmentation
+    // Verify aliasing among duplicated outputs
+    EXPECT_TRUE(out_tensor_0.is_alias_of(out_tensor_1));
+    // Verify no segmentation
+    NVF_CHECK(
+        !fec.getMostRecentKernelRuntime()->isSegmented(),
+        "segmentation is not supposed to happen");
 
     at::Tensor out_tensor = in_tensor.add(3.141);
     // Verify output values.
@@ -294,7 +298,7 @@ TEST_F(AliasTest, DuplicateOutputs) {
   }
 
   {
-    // testing segmented fusion
+    // testing duplicated output in segmented fusion
     auto fusion = std::make_unique<Fusion>();
     FusionGuard fg(fusion.get());
 
@@ -306,6 +310,8 @@ TEST_F(AliasTest, DuplicateOutputs) {
     TensorView* segment_tv = segment_set(intermediate_tv);
     TensorView* out = mul(segment_tv , IrBuilder::create<Val>(2.0));
     
+    fusion->addOutput(intermediate_tv);
+    fusion->addOutput(intermediate_tv);
     fusion->addOutput(out);
     fusion->addOutput(out); // duplicated outputs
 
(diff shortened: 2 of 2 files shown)


jjsjann123:
!build

test/test_alias.cpp
```diff
@@ -263,4 +263,95 @@ TEST_F(AliasTest, ViewPermute) {
...
+    // Verify no segmentation
+    NVF_CHECK(
+        !fec.getMostRecentKernelRuntime()->isSegmented(),
+        "segmentation is not supposed to happen");
+
+    at::Tensor out_tensor = in_tensor.add(3.141);
+    // Verify output values.
+    testValidate(
+        fec.fusion(),
+        {out_tensor, out_tensor},
```
wujingyue, line 299:
```suggestion
    at::Tensor expected_out_tensor = in_tensor.add(3.141);
    // Verify output values.
    testValidate(
        fec.fusion(),
        {expected_out_tensor, expected_out_tensor},
```

test/test_alias.cpp
```diff
@@ -263,4 +263,95 @@ TEST_F(AliasTest, ViewPermute) {
...
+        {in_tensor},
+        __LINE__,
+        __FILE__);
+  }
+
+  {
+    // testing duplicated output in segmented fusion
```
wujingyue, line 306: Can this be a separate test? It looks very independent and there are many benefits from small tests in general, e.g., isolation and parallelization.

test/test_alias.cpp
```diff
@@ -263,4 +263,95 @@ TEST_F(AliasTest, ViewPermute) {
...
+    at::Tensor out_tensor_2 = out_tensors[2];
+    at::Tensor out_tensor_3 = out_tensors[3];
+
+    // Verify aliasing among duplicated outputs
+    EXPECT_TRUE(out_tensor_0.is_alias_of(out_tensor_1));
+    EXPECT_TRUE(out_tensor_2.is_alias_of(out_tensor_3));
+    // Verify segmentation
+    NVF_CHECK(
+        fec.getMostRecentKernelRuntime()->isSegmented(),
+        "segmentation didn't happen");
```
wujingyue, line 339: Is this check redundant given the next one?

wujingyue (approved):
Thanks for the cleanup!

test/test_alias.cpp
jjsjann123: good catch :)

csrc/fusion_segmenter.cpp
```diff
@@ -1554,7 +1554,9 @@ std::unique_ptr<Fusion> SegmentedFusion::makeFusion(SegmentedGroup* sg) {
     }
   }
 
-  for (auto out : getAllOutputs(sg)) {
+  // note, we would want to keep output consistent and not artificially drop
+  // duplicates.
+  for (auto out : sg->output_vals) {
```
naoyam, line 1559: Do we need to do something similar for inputs?

jjsjann123: We wouldn't want that for inputs. i.e. if we have duplicated inputs, we would want to converge those instead.

naoyam: Hmm, I'm not 100% clear why we would want to have duplicated inputs or outputs, but if that's what the given fusion has, wouldn't we need to preserve that, no matter if it's an input or output?

naoyam: Also, I wonder why we would need to do duplicate filtering for outputs. There should be no duplicated output unless there's a duplication in the original fusion, which should be preserved.

It seems to me that inputs are different. We don't want to artificially duplicate inputs but there could be duplications in the original fusion, so we want to only filter duplications due to segmentation but don't want to filter original duplications.

jjsjann123: Ha, now I see what you mean... I thought you were referring to the segmentation logic that might introduce multiple use of a single tensor. But I think you meant to refer to c++ API `add_input` instead.

I think you are right then -> using c++ API, people could have done that as well. I'll try to patch that.

Meanwhile, this isn't a real problem with our python API.
We expose `add_output` only, so user have no-way to duplicate inputs. :laughing:

naoyam: >  this isn't a real problem with our python API.

So, does that mean that even if a user gives a single tensor as two aliased inputs, we would consider they are different tensors and not considered duplication?

jjsjann123: > so we want to only filter duplications due to segmentation but don't want to filter original duplications.

We are only filtering it out from the generated kernel.

In integration, we do have a "matching" signature and return value, where our added alias support would recognize the duplication and just aliasing the corresponding return value. So we aren't paying anything extra for that.

I'm not sure if we have the same alias analysis logic for fusion input... Which we happen to need for another optimization that we were just discussing earlier. I'll open an issue to track this one.

naoyam: > In integration, we do have a "matching" signature and return value, where our added alias support would recognize the duplication and just aliasing the corresponding return value. So we aren't paying anything extra for that.

Does that mean we can filter duplicated outputs from generated kernels as the integration would take care of them?

jjsjann123: > Does that mean we can filter duplicated outputs from generated kernels as the integration would take care of them?

That's supposed to happen. Let me double check that one, I don't think I have tested it thoroughly with different outputs including duplication in arbitrary order.

PR #1317, opened 2023-11-16
Fix `NVFUSER_EXTERNAL_SRC` after kernel name change

jacobhinkle:
This functionality was broken. Note that `fusion_id_` is commonly 0 despite having a non-zero global fusion count. This uses the global count to search through the comma-separated string instead.

### DIFF
--- a/csrc/executor.cpp
+++ b/csrc/executor.cpp
@@ -101,7 +101,7 @@ std::string getStructuredCodeFromExternalFiles(const int64_t fusion_id) {
     return "";
   }
   std::string all_external_code_paths(external_code_path);
-  if (all_external_code_paths.empty() || fusion_id < 1) {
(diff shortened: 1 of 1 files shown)


csrc/executor.cpp
```diff
@@ -101,7 +101,7 @@ std::string getStructuredCodeFromExternalFiles(const int64_t fusion_id) {
     return "";
   }
   std::string all_external_code_paths(external_code_path);
-  if (all_external_code_paths.empty() || fusion_id < 1) {
+  if (all_external_code_paths.empty() || fusion_id < 0) {
```
jacobhinkle, line 104: @rdspring1 This check on `fusion_id` could possibly be removed. I changed it here to check that `fusion_id != -1`.

rdspring1 (approved):
LGTM

PR #1324, opened 2023-11-17
Hide symbols by default.

tfogal:
... and require explicit marking to make them public. This gives us explicit control over our ABI.

I recommend starting your review with `doc/dev/visibility.md`. This describes what the purpose of the PR is and how we might update it in the future.

The CMake changes are really the ones to scrutinize. There's lots of code changes, but they are really just adding in `NVF_API` annotations to a bunch of places, or some minor include/forward declaration changes.

Todo:

- [x] Restore RPATH on libnvfuser.so (`nvfuser` library in cmake)
- [x] Add `${CMAKE_SOURCE_DIR}/csrc` as a `PUBLIC` include dir for the `nvfuser` target, and check to make that that this means the proper dir is included for `sinh_libtorch`.
- [x] Don't `set_target_properties` on `nvf_py_internal` twice.
- [x] Add reference to visibility.md in visibility.h
- [x] Investigate using $<> syntax for codegen_internal instead of adding it in the target_link_libraries


tfogal:
!build

tfogal:
!build

tfogal:
!build

tfogal:
!build

tfogal:
!build

tfogal:
!build

CMakeLists.txt
```diff
@@ -441,13 +472,28 @@ if(BUILD_TEST)
   set(NVFUSER_TESTS_KERNELS "${NVFUSER_TESTS}_kernels")
 
   add_library(${NVFUSER_TESTS_KERNELS} SHARED ${NVFUSER_ROOT}/test/rng_kernels.cu)
-  # CUDA 11 does not support C++20, so hard code C++17 here
-  set_property(TARGET ${NVFUSER_TESTS_KERNELS} PROPERTY CXX_STANDARD 17)
+  set_target_properties(${NVFUSER} PROPERTIES
+    C_STANDARD 17
```
zasdfgbnm, line 476: Should we use the following?
```suggestion
    C_STANDARD ${NVFUSER_CPP_STANDARD}
```

tfogal:
!build

CMakeLists.txt
tfogal: C standards and C++ standards are not necessarily equivalent (they happen to have coincided in years, lately, but there's no guarantee there).

tfogal:
!build

tfogal:
!build

tfogal:
!build

tfogal:
!build

tfogal:
!build

tfogal:
!build

tfogal:
!build

tfogal:
!build

tfogal:
@xwang233 sent me some patches yesterday and I forgot to apply them 🙄. mea culpa.

Just pushed it now. 🤞

!build

tfogal:
!build

tfogal:
!build

tfogal:
!build

tfogal:
!build

tfogal:
!build

CMakeLists.txt
```diff
@@ -220,51 +221,87 @@ if(BUILD_PYTHON)
...
+# that symbols are all visible. This allows us to test internal components
+# inside our test or benchmark binaries, even if we do not want said components
+# to be visible to the outside. If we used only shared libraries, then any API
+# we invoked from test binaries would need to be marked as public, even if we
+# did not want to expose it to users.
+#
+# Note technically we create an "OBJECT" library instead of a "STATIC" library.
+# This is just a CMake quirk; an OBJECT library is a better way to implement a
+# "private" (not installed) static library.
+add_library(codegen_internal OBJECT ${NVFUSER_SRCS})
```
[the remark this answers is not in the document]
tfogal: Yep! It's just a virtual thing. automake calls them "[convenience libraries](https://www.gnu.org/software/automake/manual/html_node/Libtool-Convenience-Libraries.html)" FWIW. 🤷

CMakeLists.txt
```diff
@@ -344,34 +399,51 @@ if(BUILD_PYTHON)
...
+    ${LIBNVTOOLSEXT}
+    "${TORCH_INSTALL_PREFIX}/lib/libtorch_python.so"
+    pybind11::pybind11 pybind11::headers
+  )
 
-  target_link_libraries(${NVFUSER} PRIVATE ${TORCH_LIBRARIES})
-  target_link_libraries(${NVFUSER} PRIVATE "${TORCH_INSTALL_PREFIX}/lib/libtorch_python.so")
-  target_link_libraries(${NVFUSER} PRIVATE dynamic_type)
-  target_link_libraries(${NVFUSER} PRIVATE pybind11::pybind11)
-  set_target_properties(${NVFUSER} PROPERTIES INSTALL_RPATH
```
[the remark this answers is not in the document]
tfogal: It's an oversight. I don't actually know if it's relevant---if it's working w/o it probably not 😆---but I'll restore this in the interest of shrinking the logical difference (this patch has started to suffer a bit from a 'boil the ocean' mindset already).

- [ ] Restore RPATH on libnvfuser.so (`nvfuser` library in CMake)

examples/sinh_libtorch/CMakeLists.txt
```diff
@@ -16,3 +16,8 @@ target_link_libraries(sinh_example "-Wl,-no-as-needed ${TORCH_INSTALL_PREFIX}/li
 
 target_link_libraries(sinh_example ${TORCH_LIBRARIES})
 target_link_libraries(sinh_example nvfuser_codegen)
+target_include_directories(sinh_example SYSTEM PRIVATE
+  ../../csrc
```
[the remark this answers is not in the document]
tfogal: Definitely, the correct solution is to add a new exported variable to our generated+installed CMake config file, and then picking up on that for the build of this project.

It's not the source files that are needed, just the headers. So no need to change the package.

I think the problem here is that we never set a `target_include_directories(nvfuser PUBLIC ${CMAKE_SOURCE_DIR}/csrc)`, so it doesn't know that the csrc header files are a public dependency of nvFuser as opposed to a private one. That is, I think CMake should be automagically creating + using your hypothetical `NVFUSER_INCLUDE_DIRECTORIES` variable.

Note also that we're really making way too many headers public, and similarly packaging way too many headers. As per my earlier comment though, I want to step back from boiling the ocean here ;-)

Looking at this again I also see that I've `set_target_properties` *twice* on `nvf_py_internal`, which is harmless but confusing. I'll fix that too.

- [ ] Add `${CMAKE_SOURCE_DIR}/csrc` as a `PUBLIC` include dir for the `nvfuser` target, and check to make that that this means the proper dir is included for `sinh_libtorch`.
- [ ] Don't `set_target_properties` on `nvf_py_internal` twice.

doc/dev/visibility.md
zasdfgbnm, line 1: Thank you for preparing this doc! Could you add a pointer to it in `csrc/visibility.h` so that people know where to look at?

zasdfgbnm (commented):
The changes on C++ codes LGTM. Will let @xwang233 approve, as I don't have enough expertise judging on cmake files.

CMakeLists.txt
```diff
@@ -220,51 +221,87 @@ if(BUILD_PYTHON)
...
+# We create both static and shared libraries.
+#
+# Shared libraries are what ships, but a large advantage of static libraries is
+# that symbols are all visible. This allows us to test internal components
+# inside our test or benchmark binaries, even if we do not want said components
+# to be visible to the outside. If we used only shared libraries, then any API
+# we invoked from test binaries would need to be marked as public, even if we
+# did not want to expose it to users.
+#
+# Note technically we create an "OBJECT" library instead of a "STATIC" library.
```
jjsjann123, line 233: neat~

CMakeLists.txt
```diff
@@ -220,51 +221,87 @@ if(BUILD_PYTHON)
...
-set_target_properties(${NVFUSER_CODEGEN} PROPERTIES INSTALL_RPATH
-  "$ORIGIN/../../nvidia/cuda_runtime/lib:$ORIGIN/../../nvidia/cuda_nvrtc/lib:$ORIGIN/../../nvidia/nvtx/lib:$ORIGIN/../../nvidia/cuda_cupti/lib:$ORIGIN/../../torch/lib")
-install(TARGETS ${NVFUSER_CODEGEN} EXPORT NvfuserTargets DESTINATION lib)
+add_library(nvfuser_codegen SHARED $<TARGET_OBJECTS:codegen_internal>)
+target_link_libraries(nvfuser_codegen PRIVATE
+  flatbuffers
+  ${CUDA_NVRTC_LIB}
+  ${LIBNVTOOLSEXT}
+  ${LIBCUPTI}
+  ${TORCH_LIBRARIES}
```
jjsjann123, line 290: I thought nvfuser_codegen would just inherit the dependency from the target. Do we need to have `LIBCUPT` and `TORCH_LIBRARIES` here, since they are already in `codegen_internal` as public.

CMakeLists.txt
```diff
@@ -439,7 +511,17 @@ function(add_test TEST_NAME TEST_SRC ADDITIONAL_LINK)
...
+  target_include_directories(${TEST_NAME} PRIVATE
+    ${CMAKE_SOURCE_DIR}/csrc
+  )
+  target_link_libraries(${TEST_NAME} PRIVATE
+    codegen_internal
+    ${ADDITIONAL_LINK}
+    dynamic_type
+    GTest::gtest_main GTest::gmock_main
+    flatbuffers
+    ${TORCH_LIBRARIES}
```
jjsjann123, line 523: ditto

CMakeLists.txt
```diff
@@ -344,34 +399,51 @@ if(BUILD_PYTHON)
...
+  add_dependencies(nvf_py_internal gen_nvfuser_version)
 
-  target_compile_options(${NVFUSER} PRIVATE "-DTORCH_CUDA_BUILD_MAIN_LIB")
-
-  # NB: This must be target_compile_definitions, not target_compile_options,
-  # as the latter is not respected by nvcc
-  target_compile_definitions(${NVFUSER} PRIVATE "-DTORCH_CUDA_BUILD_MAIN_LIB")
+  target_compile_definitions(nvf_py_internal PRIVATE
+    "-DTORCH_CUDA_BUILD_MAIN_LIB"
+    "-DC10_BUILD_MAIN_LIB=1"
```
jjsjann123, line 406: I'm not confident here. I thought we are setting `C10_BUILD_MAIN_LIB` and ` TORCH_CUDA_BUILD_MAIN_LIB` because we are using c10/torch macro for export/import. But we might also need this since we are still linking some of their headers?

CMakeLists.txt
```diff
@@ -220,51 +221,87 @@ if(BUILD_PYTHON)
...
-  ${CMAKE_SOURCE_DIR}/third_party/flatbuffers/include
+target_compile_definitions(codegen_internal PRIVATE "-DTORCH_CUDA_BUILD_MAIN_LIB")
+target_include_directories(codegen_internal SYSTEM PUBLIC
   ${CMAKE_SOURCE_DIR}/third_party/gloo # TODO: guard this on usage
+  ${CMAKE_SOURCE_DIR}/third_party/flatbuffers/include
+	PRIVATE
   ${CUDA_TOOLKIT_ROOT_DIR}/extras/CUPTI/include
   ${CUDA_INCLUDE_DIRS}
 )
-target_include_directories(${NVFUSER_CODEGEN} PUBLIC
```
jjsjann123, line 240: so this was the part where we were packaging the old header files. looks like they are no longer exposed through adding `codegen_internal` as part of the dependency?

I'm referring to @xwang233 's comment below on the sample cmake example

jjsjann123 (commented):
LGTM, a few nitpick.

doc/dev/visibility.md
tfogal: That's a good idea!

- [ ] Add ref to visibility.md in visibility.h

CMakeLists.txt
tfogal: Yeah, I thought so too. I tried it though, it doesn't work. 🤷

CMakeLists.txt
tfogal: Yeah, we are setting this because we are using some symbols from their headers. We haven't used their export macros for a while. And, something is seriously messed up in the whole c10 library situation. There's some `c10::Error` symbols (for example) that are undefined if we do not use these macros.

We should dig into this at some point, but as you can see this patch doesn't try to change the situation, and fixing every weird build sketchiness I encountered is an exercise in ocean boiling. I'll write up an issue, though.

CMakeLists.txt
tfogal: definitions and include directories don't impact packaging in any way. Packaging in CMake comes from the `install` command[s], but note that we also have do packaging things inside our setup.py that cmake knows nothing about.

As I understand it, because codegen_internal is a `PRIVATE` dependency its includes and such do not propagate. It might apply transitively if we use the $<> syntax and add the objects directly to nvfuser_codegen; let me investigate.


- [ ] Investigate using $<> syntax for codegen_internal instead of adding it in the target_link_libraries

CMakeLists.txt
jjsjann123: yeah torch packaging isn't bulletproof neither.

CMakeLists.txt
jjsjann123: Yes!

https://github.com/NVIDIA/Fuser/blob/1603f39bab8c1bbe12e38f2b5de53dec3b7cc373/CMakeLists.txt#L240-L243
https://github.com/NVIDIA/Fuser/blob/1603f39bab8c1bbe12e38f2b5de53dec3b7cc373/CMakeLists.txt#L297-L312

This is the part I was talking about. we were doing some questionable things, but somehow did dump everything under `./nvfuser/`, and later the setup.py just blindly package everything in the same hierarchy, so somehow a pip installed nvfuser package happen to still work with the cmake example. (at least it used to).

jjsjann123: Having said that, we don't have to fix everything in a single PR. Thanks for putting a TODO/TASK for this.

tfogal: Yeah we definitely want to be way more selective about what headers we install. But, also yes: I'll avoid that for this PR.

tfogal: Looked into this. Just using `$<TARGET_OBJECTS:library>` is not enough.

What I found is that CMake does not propagate `PUBLIC` include directories when using the `TARGET_OBJECTS` method of adding a library. Which kind of makes sense, because it's just pulling the built objects and not the library target itself. In our case, it means that our directories that you mention in line 240 do not propagate and thus there are build failures due to the missing include paths.

As per the other thread I'll still look into making the proper directories `PUBLIC` include paths on the installed target.

examples/sinh_libtorch/CMakeLists.txt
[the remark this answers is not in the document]
tfogal: I was clearly confused when writing the above. The exact thing I said doesn't work, because it doesn't make sense ;-)

This isn't needed. the `target_link_libraries(sinh_example nvfuser_codegen)` will implicitly pull in the appropriate directories, because we:

```cmake
target_include_directories(codegen_internal PUBLIC
  "$<BUILD_INTERFACE:${NVFUSER_SRCS_DIR}>"
  "$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/nvfuser>"
)
```

in the main cmake.

I think my confusion was thinking that we needed to link the `nvfuser` library instead of the `nvfuser_codegen` library.

Anyway, just going to remove this part. Thanks for calling it to my attention, Xiao!

tfogal:
!build

CMakeLists.txt
tfogal: Fixed in latest version.

tfogal:
!build

tfogal:
Thanks for all the comments! This is finally ready for review again, sorry for the delay.

xwang233 (approved):
LGTM. Thanks for working on this!

CMakeLists.txt
```diff
@@ -231,51 +231,91 @@ if(BUILD_PYTHON)
...
+target_include_directories(codegen_internal PUBLIC
   "$<BUILD_INTERFACE:${NVFUSER_SRCS_DIR}>"
   "$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/nvfuser>"
 )
-set_property(TARGET ${NVFUSER_CODEGEN} PROPERTY CXX_STANDARD_REQUIRED ON)
-set_property(TARGET ${NVFUSER_CODEGEN} PROPERTY CXX_STANDARD ${NVFUSER_CPP_STANDARD})
+set_target_properties(codegen_internal PROPERTIES
+  C_STANDARD 17
+  CUDA_STANDARD 17 # CUDA 11 does not support C++20
```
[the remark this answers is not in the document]
tfogal: > maybe a global variable for C/CUDA_STANDARD

it's good that you flagged this because IIRC we recently *did* create a variable for this. It's probably I missed this while merging. I will fix that now.

> if just use a global flag with set_property would work easier.

CMake's global flags are broken and really should never be used. the probably is that they apply to include()d subdirs and project()s that aren't necessarily expecting something to come in and change things on them. https://cliutils.gitlab.io/modern-cmake/chapters/intro/dodonot.html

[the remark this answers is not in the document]
tfogal: >> maybe a global variable for C/CUDA_STANDARD

> it's good that you flagged this because IIRC we recently did create a variable for this. It's probably I missed this while merging. I will fix that now.

okay, 😌 , I *did* properly merge in the use of `NVFUSER_CPP_STANDARD`; this is specifically about CUDA_STANDARD, which I missed upon first reading your comment.

A variable is still a good idea for these settings too, but because this patch has so many conflicts && is hard to get in, I'll wait for a follow-up PR unless CI is unhappy or similar.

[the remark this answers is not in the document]
jjsjann123: Thanks for educating me on CMake's sharp edges regarding global flags. (But really which part of CMake is smooth?!)

Yeah let's merge it first and polish afterwards. I have another PR merged right before yours and I figure my api SHOULD have been exposed as well and I likely have my code base broken.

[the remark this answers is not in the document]
tfogal: https://github.com/NVIDIA/Fuser/pull/1802

PR #1325, opened 2023-11-17
Modified `squeeze`s Python APIs to fully support Dynamic Shape Definitions

kevinstephano:
Old syntax:
```python
ops.squeeze(arg: Tensor, original_shape: List[int], dims: List[int]) -> Tensor
```
New Syntax:
```python
ops.squeeze(arg: Tensor, dims: List[int]) -> Tensor
```

The new syntax removes baking in the original tensor shape with a constant shape so it can be dynamic.

This PR also includes:
* Squeeze Python API input testing and error testing
* Replaces old syntax in `test_python_frontend.py`
* Bumps version number to `0.1.5`


kevinstephano:
!build

rdspring1 (approved):
LGTM

csrc/python_frontend/fusion_record.h
```diff
@@ -582,84 +582,66 @@ struct SqueezeOpRecord : RecordFunctor {
...
+      if (dim < 0) {
+        dim = static_cast<int64_t>(squeeze_dims.size()) + dim;
+      }
+      NVF_CHECK(
+          (dim >= 0) && (static_cast<size_t>(dim) < squeeze_dims.size()),
+          "Squeeze dim is outside of Tensor size! Tensor Size: ",
+          squeeze_dims.size(),
+          " Dim: ",
+          dim);
+      squeeze_dims[dim] = true;
```
jjsjann123, line 635: I think we need to skip `squeeze_dims[dim]` if `inp_dom[dim]` is not a size-1 dimension.

python_tests/pytest_input_generators.py
```diff
@@ -1233,6 +1233,58 @@ def slice_error_generator(
...
+        ((5, 1, 1), (-1, -2)),
+        ((1, 5, 1), (0, 2)),
+        ((1, 5, 1), (-3, -1)),
+        ((1, 1, 5), (0, 1)),
+        ((1, 1, 5), (-3, -2)),
+        ((5, 5, 5), ()),
+        ((1, 1, 1), ()),
+        ((1, 1, 1), (0, 1, 2)),
+        ((1, 1, 1), (-3, -2, -1)),
+    )
```
jjsjann123, line 1257: can we add a couple test cases:
1. squeeze a non-size-1 dimension.
2. squeeze an expanded dimension. We'll need to adjust the test setup.

jjsjann123 (changes requested):
I think our support needs to be updated.

```
>>> x.shape
torch.Size([5, 1, 1])
>>> x.squeeze(0).shape
torch.Size([5, 1, 1])
```

csrc/python_frontend/fusion_record.h
kevinstephano: We cannot skip dimensions of a specific size from the python API as we are only constructing the fusion object.  We do not have the runtime sizes.

python_tests/pytest_input_generators.py
kevinstephano: This may not be obvious but the error checking in the `pytest` tests are only for the Python API.  Therefore, the testing just makes sure the checks in the python API are working as expected given the fusion is not actually run.  The checks that exist based on the Fusion IR would be checked by a C++ test, at this time.   The reason this is the case, is that the op is getting validated in isolation since it is hard to generate a valid Fusion with an invalid Python API usage.  This is why the check for trying to squeeze a dimension that is not 1 is not done here.

kevinstephano:
> I think our support needs to be updated.
> 
> ```
> >>> x.shape
> torch.Size([5, 1, 1])
> >>> x.squeeze(0).shape
> torch.Size([5, 1, 1])
> ```

I made a conscious choice not to expand the Tensor based method support beyond simple queries like `size()` and `shape()` and there is some support for simple math operators like `+, -, *, /`.  The reason was that it added a fair amount of extra binding support that required extra specific maintenance any time changes were made that looked error prone.  You should have all the features you need in the API.

```
t1 = fd.ops.squeeze(t0, [0])
v0 = t1.shape()
```

csrc/python_frontend/fusion_record.h
jjsjann123: This is the `operator()`. So at this phase, we at least have access to its TensorDomain.

i.e. we can check for `isBroadcast` and `isExpanded` to figure out if we can skip the squeeze of a certain domain.
Without doing this, we'll run into asserts later, since our cpp API `squeeze(TensorView*, const std::vector<bool>&)` asserts on that.

Alternatively, we can update the `squeeze` API and have it clean up the squeeze dimension there.

python_tests/pytest_input_generators.py
jjsjann123: IIUC, the statement above only applies to `error checking`.

`squeeze a dimension that is not 1` is NOT an error though and we should be able to execute it. I believe our implementation here isn't doing the right thing, hence my suggestion to fix that and add a corresponding test here.

jjsjann123 (commented):
sorry about the previous confusing comment.

I'm happy with the API here as-is. I was merely suggesting that we should support squeeze on non-size-1 dimension.

jjsjann123:
Oops. the version got messed up.

Let's bump version to 0.1.4

csrc/python_frontend/fusion_record.h
jacobhinkle: I agree, we should only have a `SqueezeOp` node if `inp_dom[dim]->isBroadcast() && inp_dom[dim]->hasExpandedExtent()`. But maybe that is best placed in the `squeeze()` C++ ops?

csrc/ops/alias.cpp
```diff
@@ -246,94 +265,45 @@ TensorView* squeeze(TensorView* x, const std::vector<bool>& to_squeeze) {
...
+    if (to_squeeze[idx]) {
+      if (!id->isSymbolic()) {
+        NVF_CHECK(
+            id->isBroadcast(),
+            "Can not squeeze non-broadcasting dimension(s).");
+        NVF_CHECK(
+            !id->hasExpandedExtent(), "Can not squeeze expanded dimension(s).");
+        NVF_CHECK(
+            id->extent()->isConstScalar() && id->extent()->evaluate() == 1,
+            "Can not squeeze dimension(s) with size != 1.");
```
jacobhinkle, line 292: Could we just skip the squeeze in these conditions? Then we could ignore trivial squeezes in the python frontend.

kevinstephano: The Python Frontend calls the other `squeeze` that skips non-broadcasted dimensions that are selected to be squeezed.

csrc/ops/alias.cpp
```diff
@@ -205,26 +205,45 @@ TensorView* flatten(TensorView* x, int64_t start_dim, int64_t end_dim) {
...
   for (const auto idx : c10::irange(ndims)) {
     auto id = x_dom[idx];
     if (to_squeeze[idx]) {
       if (!id->isSymbolic()) {
-        NVF_CHECK(
-            id->isBroadcast(),
-            "Can not squeeze non-broadcasting dimension(s).");
+        // If a squeeze is attempted on a non-broadcast dimension
+        // just don't do it!  This confirms with Pytorch.
+        if (!id->isBroadcast()) {
```
jjsjann123, line 243: I'm wondering what happens when we try to squeeze a reshaped size-1 dimension. I think we'll accidentally miss that squeeze, since here we only squeeze a dimension if it's not broadcast.

Can we verify if it indeed gives us a rank-2 tensor in the example below?

```
        inputs = [
            torch.randn((3,), dtype=torch.float16, device="cuda:0").as_strided(
                (3, 4), (1, 0)
            ),
        ]
        def fusion_func(fd: FusionDefinition) -> None:
            T2 = fd.define_tensor(
                shape=[-1, -1],
                contiguity=[None, True],
                dtype=DataType.Half,
                is_cpu=False,
            )
            S7 = fd.define_scalar(3, dtype=DataType.Int)
            S8 = fd.define_scalar(4, dtype=DataType.Int)
            S9 = fd.define_scalar(1, dtype=DataType.Int)
            V10 = fd.define_vector([S7, S8, S9], dtype=DataType.Int)
            T11 = fd.ops.reshape(T2, new_shape=V10)
            T12 = fd.ops.squeeze(T11, (2))

        nvf_out, _ = self.exec_nvfuser(fusion_func, inputs)
```

jjsjann123: tagging issue #1393 

I'm expecting the squeeze would fail the same thing. I'm wondering if we are doing the right thing with our dynamic reshape, if we need to patch everything c++ API to consider the symbolic dimension....

jjsjann123 (commented):
My old concerns were addressed. But now I'm wondering if this would suffer from the same issue from dynamic reshape.

jjsjann123:
Let's not forget about this PR. we can discuss it when we get back from vacation. :wink:

csrc/ops/alias.cpp
kevinstephano: There seems to be more than one issue in this case related to just expanding. 
```
import torch

inputs = [
        torch.randn((3,), dtype=torch.float16, device="cuda:0").as_strided( (3, 4), (1, 0) ),
         ]

def fusion_func(fd: FusionDefinition) -> None:
    T2 = fd.define_tensor( shape=[-1, -1], contiguity=[None, True], dtype=DataType.Half, is_cpu=False, )
    S7 = fd.define_scalar(3, dtype=DataType.Int)
    S8 = fd.define_scalar(4, dtype=DataType.Int)
    S9 = fd.define_scalar(1, dtype=DataType.Int)
    V10 = fd.define_vector([S7, S8, S9], dtype=DataType.Int)
    T11 = fd.ops.reshape(T2, new_shape=V10)
    #T12 = fd.ops.squeeze(T11, (2,))
    fd.add_output(T11)

    fusion_func(fd)

out = fd.execute(inputs)
```

Error:
```
  File "/opt/pytorch/nvfuser/nvfuser/__init__.py", line 137, in execute
    result = self._execute(
RuntimeError: t.size(i) == 1 || t.stride(i) == 0 INTERNAL ASSERT FAILED at "/opt/pytorch/nvfuser/csrc/expr_evaluator.cpp":130, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. IterDomain bS0{i0} in TensorView T0_g[ bS0{i0}, iS1{i1} ] has expanded extent but input tensor has size 3 and stride 1 in dimension 0
```

kevinstephano:
This PR is now doing a few things.  
1. Originally their were 4 C++ `squeeze` Alias operations. This removes 3 of those interfaces and adds an interface specifically compliant to Pytorch.  I left one of the historical interface as it matches the IR Node while the other 3 were never used and fed the fully specified boolean squeeze dimension.
2. This adds functionality to the interface like allowing you the user to attempt to squeeze a non-broadcasted dimension which results in not doing a squeeze versus tripping an error.
3. Python based tests were added to test the API errors and correctness.

@jjsjann123 notes there are some issues with expanded tensors and squeezes.  One problem is that, expands from `reshape` operations seem to have some issues.

kevinstephano:
!build

csrc/ops/alias.cpp
jjsjann123: sorry this is a busted example that I put there.
Contiguity should be `[True, None]` instead. And then there seems to be an issue with dynamic reshape not considering the expanded extent when validating consistent numel.

I think Jingyue(?!) has already fixed that and I see the updated code in main.

jjsjann123 (approved):
My concerns have all been cleared. stamp!

jjsjann123:
!build

PR #1327, opened 2023-11-17
Clean `test_gpu_view.cpp`.

Priya2698:
1. Add `rsrt`, `cos`, `bit_cast` to `UnaryOp::evaluate`.
2. PolymorphicValue `abs` for `at::Tensor` type.

### DIFF
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -459,7 +459,7 @@ std::vector<PolymorphicValue> UnaryOp::evaluate(
       return {in.as<at::Tensor>().cos()};
       break;
     case UnaryOpType::BitCast:
-      if (isComplexType(input(0)->dtype()) && !isComplexType(out()->dtype())){
+      if (isComplexType(input(0)->dtype()) && !isComplexType(out()->dtype())) {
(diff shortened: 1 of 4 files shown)


csrc/ir/nodes.cpp
```diff
@@ -455,6 +455,19 @@ std::vector<PolymorphicValue> UnaryOp::evaluate(
     case UnaryOpType::Sin:
       return {in.as<at::Tensor>().sin()};
       break;
+    case UnaryOpType::Cos:
+      return {in.as<at::Tensor>().cos()};
+      break;
+    case UnaryOpType::BitCast:
+      if (isComplexType(input(0)->dtype()) && !isComplexType(out()->dtype())) {
+        return {at::view_as_real(in.as<at::Tensor>())};
```
zasdfgbnm, line 463: I don't think this is correct. What if we want to bit cast `complex<float>` as `int64_t`? I think we should check that if the output dtype is something not supported by ATen, and implement it as a no-op. It makes more sense to call `at::view_as_real` inside `ViewAsScalar::evaluate`.

zasdfgbnm (approved):

PR #1174, opened 2023-10-27
Preserve expanded broadcasts in reshape

jacobhinkle:
Currently, we always "realize" expanded broadcasts (replace them with `Iteration` domains) during reshape if they appear in a split or merge transform. This is not always necessary, for example we never need to do this for a `SplitTransform`, and we can avoid it for a `MergeTransform` if both inner and outer IDs are `Broadcast`. This PR makes that a reality.

Fixes #1126.

PR #1329, opened 2023-11-17
Unify `IterDomain::split` with `SplitTransform` and similar for merge

jacobhinkle:
Currently, `MergeTransform` and `SplitTransform`, which are the classes used to perform reshape operations, do not re-use the splitting and merging operations used in scheduling, namely the static methods `IterDomain::{split,merge}`. This leads to needing to maintain changes in both as I recently found out when debugging #1174. We can actually already notice they have diverged with respect to `IndexType::GatherScatter` handling.

This PR simply calls the static `IterDomain` method from each of the reshape `*Transform` classes. Since those classes also need to set the rfactor-product flag on the produced `IterDomain`s, I added a flag to the static methods that is only used during reshape.

### DIFF
--- a/csrc/ir/internal_base_nodes.h
+++ b/csrc/ir/internal_base_nodes.h
@@ -120,8 +120,8 @@ class IterDomain : public Val {
   static std::vector<IterDomain*> clone(
       const std::vector<IterDomain*>& domains);
 
-  // When `rfactor_domain` is true, also set the `is_rfactor_domain_` flag of
-  // the result IterDomain.
+  //! When `rfactor_domain` is true, also set the `is_rfactor_domain_` flag of
+  //! the result IterDomain.
   static IterDomain* merge(
       IterDomain* outer,
       IterDomain* inner,
@@ -129,6 +129,8 @@ class IterDomain : public Val {
 
   //! start_offset and stop_offset defines partial split. Only root
   //! domains are allowed to have non-zero start and stop offsets.
+  //! When `rfactor_domain` is true, also set the `is_rfactor_domain_` flag of
+  //! both result IterDomains.
   static std::pair<IterDomain*, IterDomain*> split(
       IterDomain* in,
       Val* factor,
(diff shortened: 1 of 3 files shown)


jacobhinkle:
!build

csrc/transform_view.cpp
```diff
@@ -277,23 +271,13 @@ class SplitTransform final : public ViewTransform {
...
-            .build();
-
-    IrBuilder::create<Split>(factor_id, remainder_id, id, factor, false);
+    auto [factor_id, remainder_id] = IterDomain::split(
+        id,
+        factor,
+        /*inner_split*/ false,
+        /*start_offset*/ nullptr,
+        /*stop_offset*/ nullptr,
+        /*rfactor_domain*/ true);
```
wujingyue, line 280: https://clang.llvm.org/extra/clang-tidy/checks/bugprone/argument-comment.html

`/*param_name=*/` has the advantage of being checked by clang-tidy. I'm not sure `/*param_name*/` has the same effect. Please ignore if it does. 

```suggestion
        /*inner_split=*/false,
        /*start_offset=*/nullptr,
        /*stop_offset=*/nullptr,
        /*rfactor_domain=*/true);
```

csrc/ir/internal_base_nodes.h
```diff
@@ -120,7 +120,10 @@ class IterDomain : public Val {
   static std::vector<IterDomain*> clone(
       const std::vector<IterDomain*>& domains);
 
-  static IterDomain* merge(IterDomain* outer, IterDomain* inner);
+  static IterDomain* merge(
```
wujingyue, line 123:
```suggestion
  // When `rfactor_domain` is true, also set the `is_rfactor_domain_` flag of the result IterDomain. 
  static IterDomain* merge(
```

wujingyue (approved):

csrc/transform_view.cpp
jacobhinkle: That's good to know! Thanks.

jacobhinkle:
!build --diff

jacobhinkle:
!build --diff

PR #1330, opened 2023-11-18
Fix codediff ptx

jacobhinkle:
The change of default kernel IDs broke the link between PTX and CUDA dumped filenames, e.g. for a given kernel we might have `__tmp_kernel5.cu` and `__tmp_kernel_pointwise_f0_c1_r0_g0.ptx`. Also this re-uses kernel names so we wound up with far fewer PTX files than .cu files when running code diffs. This can be avoided by setting `NVFUSER_ENABLE=static_fusion_count` which will revert to name very similar to the old behavior. This PR sets that in `run_command.sh` in order to fix the code diff script, and ensures the filenames match, i.e. `__tmp_kernel5.cu` is now `__tmp_kernel_5.cu`.

### DIFF
--- a/csrc/executor.cpp
+++ b/csrc/executor.cpp
@@ -171,7 +171,7 @@ std::string FusionExecutor::getStructuredCode(
   if (isDebugDumpEnabled(DebugDumpOption::CudaToFile) ||
       isDebugDumpEnabled(DebugDumpOption::DebugInfo)) {
     std::stringstream file_name;
-    file_name << "__tmp_kernel" << getGlobalFusionCount() << ".cu";
+    file_name << "__tmp_kernel_" << kernel_id_ << ".cu";
     debug() << "PRINTING: " << file_name.str() << std::endl;
     std::ofstream out(file_name.str());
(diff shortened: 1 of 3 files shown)


xwang233 (approved):

csrc/executor.cpp
```diff
@@ -171,7 +171,7 @@ std::string FusionExecutor::getStructuredCode(
   if (isDebugDumpEnabled(DebugDumpOption::CudaToFile) ||
       isDebugDumpEnabled(DebugDumpOption::DebugInfo)) {
     std::stringstream file_name;
-    file_name << "__tmp_kernel" << getGlobalFusionCount() << ".cu";
+    file_name << "__tmp_kernel_" << kernel_id_ << ".cu";
```
xwang233, line 174: Would it be easier to keep the old `kernel934` style without the trailing underscore? I think the current special handling on diff-bench relies on this naming without trailing underscore, but we can definitely have the trailing underscore and fix diff-bench later.

rdspring1 (approved):
LGTM

PR #1334, opened 2023-11-19
Remove Volta matmul support

zasdfgbnm:


zasdfgbnm:
!build

zasdfgbnm:
!build

drzejan2 (approved):
With this change the support for Volta architecture will be dropped - will there be a architecture support matrix?

zasdfgbnm:
> With this change the support for Volta architecture will be dropped - will there be a architecture support matrix?

I think we mainly care about Ampere and Hopper. But supporting Turing is not much additional work, so also Turing.

PR #1337, opened 2023-11-20
Make the `FusionProfiler` printing mechanical.

kevinstephano:
The point of this PR is to add print attributes per column in table and then to mechanically print each line by putting each line into a Tuple.  The hope is this also is valuable for printing via Python where the expectation is to add lines into a DataFrame.

The printed output should not change!

Changes:
* Mechanical Printing
  * Created a table of attributes per Profiler Column
  * Created Tuples for Cupti based lines and nocupti based lines
  * Replaced `operator<<()` definition with a mechanical function that uses tuples
* Added a function to query and turn off CPP printing to disable the `operator<<` based printing when used from python. NOTE: the python specific printing has not been added, yet.


kevinstephano:
!build

csrc/fusion_profiler.cpp
```diff
@@ -387,146 +443,156 @@ void FusionProfile::reset() {
...
+
+namespace {
+// The operator* overloads are to satisfy the compiler and should not be called!
+double operator*(const std::basic_string<char>& a, double b) {
+  NVF_ERROR(false, "This types operator* overload should not be called!");
+  return 0.0;
+}
+
+template <typename T, size_t I>
+std::string to_string(const std::array<T, I>& cont) {
```
jacobhinkle, line 495: Nit: `toString` following camelCase for C++ functions (ignore this if you are instead trying to keep consistent with a python snake_case interface).

csrc/fusion_profiler.cpp
```diff
@@ -387,146 +443,156 @@ void FusionProfile::reset() {
...
+      out += ", ";
+    }
+    out += std::to_string(elem);
+  }
+  out += "]";
+  return out;
+}
+
+template <bool NOCUPTI = false, size_t I = 0, typename... Ts>
+constexpr std::ostream& print_tuple(
```
jacobhinkle, line 511: Same comment as above re camelCase.

csrc/fusion_profiler.cpp
```diff
@@ -378,6 +379,61 @@ uint32_t SegmentProfiler::segmentId() const {
...
+  NVF_CHECK(
+      !prof.kernel_profiles.empty(),
+      "Cannot convert FusionProfile to a tuple containing CUPTI gathered stats!");
+  NVF_CHECK(
+      seg_id < prof.kernel_profiles.size(),
+      "Invalid seg_id for FusionProfile. Segments: ",
+      prof.kernel_profiles.size(),
+      " seg_id: ",
+      seg_id);
+  auto& kp = prof.kernel_profiles[seg_id];
```
jacobhinkle, line 392: Is `const auto&` safe here? It changes the return type but I think the tuple type is never explicitly stated anyway. If it doesn't work, please ignore.

jacobhinkle (approved):
LGTM

csrc/fusion_profiler.cpp
```diff
@@ -675,15 +749,21 @@ void FusionProfiler::stop() {
           fp->segments_[kp_idx].state() == ProfilerState::Finished,
           "SegmentProfiler ProfilerState is not Finished!",
           fp->segments_[kp_idx].state());
+      kprof.segment_id = static_cast<size_t>(kp_idx);
       kprof.input_bytes = segment(kp_idx).inputBytes();
       kprof.output_bytes = segment(kp_idx).outputBytes();
-      kprof.effective_bandwidth_gbs =
-          (double)(kprof.input_bytes + kprof.output_bytes) / kprof.time_ms *
-          mb_divider;
+      kprof.effective_bandwidth_gbs = (double)(segment(kp_idx).inputBytes() +
```
jjsjann123, line 755: nitpick, what's the reason for code change here? wasn't `segment(kp_idx).inputBytes()` exactly what we assigned to `kprof.input_bytes` earlier?

jjsjann123 (approved):

csrc/fusion_profiler.cpp
```diff
@@ -387,146 +443,156 @@ void FusionProfile::reset() {
...
-    "Fus#",      "NSegs",       "CuEvtTm(ms)",  "HstTm(ms)",
-    "CmpTm(ms)", "KerTm(ms)",   "EffBw(GB/s)",  "%PeakBw",
-    "S-Seg#",    "S-KerTm(ms)", "S-CmpTm(ms)",  "S-EffBw(GB/s)",
-    "S-%PeakBw", "S-In(MB)",    "S-Out(MB)",    "S-Smem[Dyn,Stat]",
-    "S-Regs",    "S-Grid",      "S-Block",      "S-Cluster",
-    "S-Dev",     "S-Stm",       "S-PkBw(GB/s)", "S-DeviceName",
-    "S-KerName"};
+const size_t FusionProfile::first_cupti_idx = 5;
+const std::vector<ProfileAttrDescriptor> FusionProfile::profile_attr_descs{
+    // column_header, verbose, segment, list, column_width, number,
```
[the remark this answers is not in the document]
kevinstephano: The `column_width` does not always match the `column_header` size.  For instance, `S-Grid` is an example where it's string is of size 6 and the `column_width` is 16.

csrc/fusion_profiler.cpp
```diff
@@ -387,146 +443,156 @@ void FusionProfile::reset() {
...
-std::array<const char*, 25> column_strs{
-    "Fus#",      "NSegs",       "CuEvtTm(ms)",  "HstTm(ms)",
-    "CmpTm(ms)", "KerTm(ms)",   "EffBw(GB/s)",  "%PeakBw",
-    "S-Seg#",    "S-KerTm(ms)", "S-CmpTm(ms)",  "S-EffBw(GB/s)",
-    "S-%PeakBw", "S-In(MB)",    "S-Out(MB)",    "S-Smem[Dyn,Stat]",
-    "S-Regs",    "S-Grid",      "S-Block",      "S-Cluster",
-    "S-Dev",     "S-Stm",       "S-PkBw(GB/s)", "S-DeviceName",
-    "S-KerName"};
+const size_t FusionProfile::first_cupti_idx = 5;
+const std::vector<ProfileAttrDescriptor> FusionProfile::profile_attr_descs{
```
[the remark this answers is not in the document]
kevinstephano: I originally tried to use a `std::array` but the compiler does not like the use of a `struct`, unfortunately.

csrc/fusion_profiler.cpp
```diff
@@ -387,146 +443,156 @@ void FusionProfile::reset() {
...
+    {"S-Cluster", true, true, true, 16, false, 0, std::nullopt},
+    {"S-Dev", true, true, false, 5, true, 0, std::nullopt},
+    {"S-Stm", true, true, false, 5, true, 0, std::nullopt},
+    {"S-PkBw(GB/s)", true, true, false, 12, true, 3, std::nullopt},
+    {"S-DeviceName", true, true, false, 20, false, 0, std::nullopt},
+    {"S-KerName", false, true, false, 20, false, 0, std::nullopt}};
+
+namespace {
+// The operator* overloads are to satisfy the compiler and should not be called!
+double operator*(const std::basic_string<char>& a, double b) {
```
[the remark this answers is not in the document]
kevinstephano: I am noting the error in a comment in the code:

```cpp
        if (desc.unit_multiplier.has_value()) {
          // NOTE: The "operator*(const std::basic_string<char>& a, double b)"
          // that is defined in this anonymous namespace is used to prevent a
          // compiler error in the following line as some tuple values are
          // strings and trigger this overload even though it does not make
          // sense to execute.
          os << static_cast<double>(
              std::get<I>(tup) * desc.unit_multiplier.value());
        } else {
          os << std::get<I>(tup);
        }
```

csrc/fusion_profiler.cpp
kevinstephano: I have no idea why I changed this.  I changed it back.

kevinstephano:
!build

PR #1350, opened 2023-11-20
Check gcc version and raise error if not supported

zasdfgbnm:


zasdfgbnm:
!build

tfogal (commented):
I don't think we should explicitly check versions; this sets us up for a bad path of having a 'supported versions' list for compilers.

There's a CXX_STANDARD_REQUIRED setting we can use. I start using it in a patch i'm preparing (https://github.com/NVIDIA/Fuser/pull/1324/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR479 for example), so if we could, I'd like us to wait to avoid creating more diffs for me to deal with (it's a painful patch as-is!).

zasdfgbnm:
> There's a CXX_STANDARD_REQUIRED setting we can use.

The reason for checking compiler version is not for C++20. Earlier compilers already start a bit support on C++20. The reason for checking is mostly because earlier `gcc`s has compiler bugs that does not compile nvFuser.

zasdfgbnm:
> if we could, I'd like us to wait to avoid creating more diffs for me to deal with (it's a painful patch as-is!).

I understand. If you think it makes sense to reject gcc < 11 and tell the users about the compiler bug, could you copy-paste the following code to somewhere in your PR, and I will close this PR?

```C++
if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
    if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 11.4)
        message(FATAL_ERROR "GCC < 11.4 has compiler bugs and can not compiler nvFuser")
    endif()
endif()
```

tfogal:
>>  There's a CXX_STANDARD_REQUIRED setting we can use.
>
> The reason for checking compiler version is not for C++20. [...] earlier `gcc`s has compiler bugs [...]
>
> > if we could, I'd like us to wait to avoid creating more diffs for me to deal with (it's a painful patch as-is!).
> 
>  If you think it makes sense [...] could you copy-paste the following code to somewhere in your PR, and I will close this PR?

No, I misunderstood the situation; let's get this in, I'll just rebase. Some day I'd like to delve deeper on these compiler bugs, but I can't reasonably make time now and it's not reasonable to hold off until I find time for it.

tfogal (approved):

PR #1356, opened 2023-11-20
Disable `modernize-loop-convert`

zasdfgbnm:
This is asking to convert loops like
```C++
for (auto it = exprs.rbegin(); it != exprs.rend(); ++it)
```
into
```C++
for (auto expr : std::ranges::reverse_view(exprs))
```
However, seems that clang <16 can not compile the latter code

So disable this check for now to unblock. If we want to upgrade clang versions, we can re-enable it.


jjsjann123 (commented):
:shipit:

xwang233 (approved):
😂

PR #1357, opened 2023-11-20
Roll forward "AliasAnalysis handles slices. (#1281)". 

wujingyue:


wujingyue:
!build

csrc/optimization/alias_analysis.cpp
```diff
@@ -295,16 +295,17 @@ void AliasFinder::handle(const SliceOp* slice) {
...
     if (out_layout.allocation_domain[i]->isBroadcast()) {
       out_layout.contiguity[i] = std::nullopt;
-      continue;
-    }
-
-    if (next_non_broadcast_is_non_contiguous) {
+      // But don't reset next_non_broadcast_is_non_contiguous. Say we slice a
+      // TV of shape [4,1,7],[t,t,t] to a TV of shape [4,1,5]. The output
+      // contiguity should be [f,n,t].
+    } else if (next_non_broadcast_is_non_contiguous) {
```
jjsjann123, line 301: The change here does make sense to me. But I'm not following the comment here.

> TV of shape [4,1,7],[t,t,t]
> don't reset next_non_broadcast_is_non_contiguous.

The old code calls `continue` for broadcast dimension, so we are not skipping anything here are we?

But rather the broadcast dimension here is not viewed as such, since its contiguity flag is marked as `True`?

test/test_alias.cpp
```diff
@@ -368,6 +368,50 @@ TEST_F(AliasTest, DuplicateOutputs) {
...
+
+  TensorView* in = makeContigConcreteTensor({4, 1, 7});
+  fusion->addInput(in);
+  TensorView* out = slice(in, {0, 0, 0}, {4, 1, 5});
+  fusion->addOutput(out);
+
+  FusionExecutorCache fec(std::move(fusion));
+  at::Tensor in_tensor = at::randn({4, 1, 7}).cuda();
+  at::Tensor out_tensor = fec.runFusionWithInputs({in_tensor})[0];
+  EXPECT_TRUE(out_tensor.is_alias_of(in_tensor));
```
tfogal, line 405: You would know better than me, but this confused me. I'm assuming that `7` is the fastest dimension in `{4,1,7}`, and it might be that I just have this backwards :-)

But if I have it the right way, I think of two tensors as aliases when they use the same base address for the data, but wouldn't the `out` tensor need a stride of `7` (despite the dim being `5`) in the fastest dim for that to be true? Does the alias analysis / usage perhaps guarantee that `out` indeed *does* get a stride `7`? If not, where did my logic go awry?

tfogal (approved):
LGTM.

Would love if you could take some time to educate me on how the alias analysis is used, though! see comment

csrc/optimization/alias_analysis.cpp
tfogal: Note this is over the `[f,n,t]` of the *output* tensor, Jie, not the `[t,t,t]` of the input tensor.
(but still a valid point, seems like the code is the same)

test/test_alias.cpp
jjsjann123: Think about this in the python indexing. Any basic indexing (slice included) can be done via a meta update as a `view` into the original storage and you don't need copy.

after the slice, the stride on each dimension remains the same and we are only updating the shape (and offset when it applies)

jjsjann123:
```
import numpy as np
arr = np.arange(28).reshape(4, 1, 7)
b = arr[:,:,0:5]
```

```
>>> arr.strides
(56, 56, 8)
>>> b.strides
(56, 56, 8)
>>> arr.ctypes.data
94247211410336
>>> b.ctypes.data
94247211410336
```

csrc/optimization/alias_analysis.cpp
jjsjann123: I guess I see how the first example below when we are slicing a broadcast dimension, how this code could fix it.

But the second example where the comment is referring doesn't make sense to me at all.

wujingyue: I added this comment because I almost wrote something like
```
    if (out_layout.allocation_domain[i]->isBroadcast()) {
      out_layout.contiguity[i] = std::nullopt;
      next_non_broadcast_is_non_contiguous = false;
    } else ...
```
which is wrong. We shouldn't reset `next_non_broadcast_is_non_contiguous` here when deciding the contiguity of a broadcast dimension. 

After a second thought, I feel it's pretty obvious because the variable name says it all, so I may just remove this comment.

test/test_alias.cpp
wujingyue: > I'm assuming that 7 is the fastest dimension in {4,1,7}

Right. 

> I think of two tensors as aliases when they use the same base address for the data

No, they alias when they share the storage buffer but the base addresses don't have to be the same. I believe @jjsjann123 answered this question. See also `AliasTest.SliceViewPermute` where all the three outputs are alias but only the first one uses the same base address.

csrc/optimization/alias_analysis.cpp
wujingyue: > But the second example where the comment is referring doesn't make sense to me at all.

How so? You think the contiguity ought to be something else?

jjsjann123: Ha, now I see where the comment it coming from.

I somehow was reading too much into the example and how those were translating to the code. NVM my poor reading.

jjsjann123 (approved):

wujingyue:
!build

PR #1362, opened 2023-11-22
Encode mma macro enum as `uint64_t` and define hopper macros

zasdfgbnm:
`MacroType` is a enum that specifies which mma instruction to use. In the past, the underlying type for this enum is just the default given by C++. This PR changes this to `uint64_t`, and explicitly specify a value for each macro, so that the first two bits represent arch, the second two bits represent `m`. This makes it possible to pull the information of an enum without using a `switch`. I am trying to avoid `switch` because hopper have a lot of macros, and having to write all of them down is so boring...

This PR also adds hopper macros, although they are not implemented yet.

### DIFF
--- a/csrc/codegen.cpp
+++ b/csrc/codegen.cpp
@@ -1071,7 +1071,23 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
     std::stringstream ss;
     auto macro = mma->macro();
     ss << genArchString(macro) << "::";
-    ss << toString(macro);
+
+    switch (macro) {
+      case MmaOptions::MacroType::NoMMA:
+        ss << "NoOp";
+        break;
+      case MmaOptions::MacroType::Turing_16_8_16:
+      case MmaOptions::MacroType::Ampere_16_8_16:
+        ss << "M16N8K16";
+        break;
+      case MmaOptions::MacroType::Turing_16_16_16:
+      case MmaOptions::MacroType::Ampere_16_16_16:
+        ss << "M16N16K16";
+        break;
+      default:
+        NVF_ERROR(false, "undefined mma type");
+        break;
+    }
 
     // clang-tidy: bugprone-unchecked-optional-access
     // clang-tidy assumes that function result is unstable, so we need a copy.
--- a/csrc/mma_type.h
+++ b/csrc/mma_type.h
@@ -83,7 +83,7 @@ struct MatMulTileOptions {
 
 enum class MmaMacro : uint64_t;
 
-struct MmaMacroUnderlying {
+struct MmaMacroEncode {
   enum class Arch { NoMma, Volta, Turing, Ampere, Hopper } arch : 16;
   unsigned m : 16;
   unsigned n : 16;
@@ -101,13 +101,13 @@ struct MmaMacroUnderlying {
 
   constexpr operator MmaMacro();
 
-  constexpr MmaMacroUnderlying(MmaMacro macro);
+  constexpr MmaMacroEncode(MmaMacro macro);
 
-  constexpr MmaMacroUnderlying(Arch arch, unsigned m, unsigned n, unsigned k)
+  constexpr MmaMacroEncode(Arch arch, unsigned m, unsigned n, unsigned k)
       : arch(arch), m(m), n(n), k(k) {}
 };
 
-static_assert(sizeof(MmaMacroUnderlying) == sizeof(uint64_t));
+static_assert(sizeof(MmaMacroEncode) == sizeof(uint64_t));
 
 //! Type of mma instrinsic macro to use
 //!  This will translate to which mma intrinsic from runtime string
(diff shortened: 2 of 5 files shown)


csrc/codegen.cpp
```diff
@@ -1071,7 +1071,23 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
     std::stringstream ss;
     auto macro = mma->macro();
     ss << genArchString(macro) << "::";
-    ss << toString(macro);
+
+    switch (macro) {
```
zasdfgbnm, line 1075: I understand that it makes more sense to keep this as a separate function, but this entire `genMmaOp` will be removed soon, so I don't really care about code quality.

zasdfgbnm:
!build

csrc/mma_type.h
```diff
@@ -71,26 +81,119 @@ struct MatMulTileOptions {
...
+  return static_cast<MmaMacro>(static_cast<uint64_t>(*this));
+#endif
+}
+
+constexpr MmaMacroUnderlying::MmaMacroUnderlying(MmaMacro macro) {
+#if IS_CPP20 && !defined(__clang__)
+  // std::bit_cast for bit field is not supported by clang yet
+  *this = std::bit_cast<MmaMacroUnderlying>(macro);
+#else
+  k = (uint64_t)macro & 0xFFFF; // NOLINT
```
zasdfgbnm, line 187: clang-tidy wants me to put this into member initializer, but I want it to stay on function body so it can be more conveniently switched by the `#if IS_CPP20` guard.

zasdfgbnm:
!build

csrc/mma_type.h
```diff
@@ -71,26 +81,119 @@ struct MatMulTileOptions {
...
-  //!  latency bound kernels.
-  enum class MacroType {
-    NoMMA = 0,
-    Ampere_16_8_16,
-    Ampere_16_16_16,
-    Turing_16_8_16,
-    Turing_16_16_16,
-    Ampere_16_8_8 // place holder for tf32
-  };
+  using MacroType = MmaMacro;
```
zasdfgbnm, line 196: I might need to do a global string replacement `MmaOptions::MacroType` -> `MmaMacro`, but this will conflict with my other PRs. So will do this clean up in a followup when my related outstanding PRs are merged.

zasdfgbnm:
!build

csrc/mma_type.h
wujingyue, line 187: Instead of disabling all warnings universally, consider `NOLINT(a-specific-error)`.

wujingyue, line 187: Consider https://github.com/NVIDIA/Fuser/blob/cee1010634c77a69328762f91bc55cea5f431f95/csrc/type.h#L1002

csrc/mma_type.h
```diff
@@ -71,26 +81,119 @@ struct MatMulTileOptions {
...
+//!  one can use 16_8_8 or 16_8_16.
+//! Will consider adding more choices that the scheduler can pick from
+//!  when our perf target becomes more fine grained, which is more likely in
+//!  latency bound kernels.
+
+#define MACRO(arch, m, n, k) \
+  arch##_##m##_##n##_##k =   \
+      MmaMacroUnderlying(MmaMacroUnderlying::Arch::arch, m, n, k)
+
+enum class MmaMacro : uint64_t {
```
wujingyue, line 127: I must be missing the big picture because I haven't read your subsequent PRs. Why do we need this enum? In other words, what can this achieve while `MmaMacroUnderlying` can't?

zasdfgbnm: I think the biggest benefit is for being able to check that we are using a valid macro at compilation time. For example, if in your heuristics, you wrote something like
```C++
if (blabla) {
  macro = MmaMacro::Turing_16_7_16;
}
```
you will get a compilation error at exactly that line of code, because `MmaMacro::Turing_16_7_16` is not a valid macro. 

However, without being an enum, and if you did:
```C++
if (blabla) {
  macro = MmaMacroUnderlying(Turing, 16, 7, 16);
}
```
your code will compile correctly, and during lowering, you likely get ptxas error. The former is much easier to debug.

wujingyue: I see -- the enum defines a set of **valid** <arch,m,n,k> combinations. 

Naming nit: MmaMacroUnderlying is a bit confusing because the underlying type of MmaMacro is really uint64_t. Now that I understand the enum defines a set of valid combinations, you may want to `s/MmaMacro/ValidMmaMacro` and `s/MmaMacroUnderlying/MmaMacro`. However, if you are going to use the enum form way more often, calling it `MmaMacro` makes more sense.

wujingyue (approved):

csrc/mma_type.h
zasdfgbnm: changed

zasdfgbnm: Changed to `// NOLINT(*-member-init)`, I am not using the full name because it is too long and will enforc line break, which looks ugly.

csrc/mma_type.h
zasdfgbnm: I do think we will use the enum form more often. I will keep its name `MmaMacro`. And I renamed `MmaMacroUnderlying` to `MmaMacroEncode`, hope this makes it less confusing.

zasdfgbnm:
!build

zasdfgbnm:
!build

zasdfgbnm:
!build

zasdfgbnm:
!build

drzejan2:
Checked post-mortem.

Thank you for making this improvement and sharing the nice way of handling enums!

PR #1363, opened 2023-11-22
Alloc domain patch concretization

jjsjann123:
DynamicTransformConcretizer should preserve allocation domain

### DIFF
--- a/csrc/dynamic_transform.cpp
+++ b/csrc/dynamic_transform.cpp
@@ -753,6 +753,9 @@ void DynamicTransformConcretizer::mutate(TensorDomain* td) {
       ? updateIdVec(td->maybeRFactor())
       : std::vector<IterDomain*>();
   std::vector<IterDomain*> domain = updateIdVec(td->leaf());
+  std::vector<IterDomain*> alloc_dom = td->hasAllocation()
+      ? updateIdVec(td->maybeAllocation())
+      : std::vector<IterDomain*>();
 
   if (!mutated) {
     return;
@@ -761,8 +764,12 @@ void DynamicTransformConcretizer::mutate(TensorDomain* td) {
   // Update the contiguity vector. Drop the contig val if mutated to broadcast
   auto contig = td->contiguity();
 
-  for (const auto i : c10::irange(td->maybeRFactor().size())) {
-    auto original_id = td->maybeRFactor().at(i);
+  const auto& maybe_alloc = td->hasAllocation() ? alloc_dom
+      : td->hasRFactor()                        ? rfactor_dom
+                                                : root_dom;
+
+  for (const auto i : c10::irange(maybe_alloc.size())) {
+    auto original_id = maybe_alloc.at(i);
     if (original_id->getIterType() != IterType::Symbolic) {
       continue;
     }
@@ -772,7 +779,7 @@ void DynamicTransformConcretizer::mutate(TensorDomain* td) {
         "Unexpected to have a non-contig symbolic domain: ",
         original_id->toString());
 
-    auto updated_id = td->hasRFactor() ? rfactor_dom.at(i) : root_dom.at(i);
+    auto updated_id = maybe_alloc.at(i);
 
     // If the concretized ID is a broadcast domain, drop the contig val
     if (updated_id->isBroadcast()) {
@@ -781,7 +788,7 @@ void DynamicTransformConcretizer::mutate(TensorDomain* td) {
   }
 
   Val* mutated_val = IrBuilder::create<TensorDomain>(
-      td->container(), root_dom, rfactor_dom, domain, contig);
+      td->container(), root_dom, rfactor_dom, alloc_dom, domain, contig);
   registerConcretization(td, mutated_val);
 }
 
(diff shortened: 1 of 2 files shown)


jjsjann123:
!build

python_tests/test_python_frontend.py
```diff
@@ -2584,6 +2584,29 @@ def dynamic_reshape(fd: FusionDefinition) -> None:
         self.assertEqual(y.shape, torch.Size([3, 2, 2]))
         self.assertEqual(x.flatten(), y.flatten())
 
+    def test_allocation_domain_concretization(self):
+        inputs = [
+            torch.randn((0,), dtype=torch.float64, device="cuda:0").as_strided(
+                (1, 0, 1, 1), (0, 1, 1, 1)
```
jacobhinkle, line 2590: Why does the input need to be empty here?

python_tests/test_python_frontend.py
```diff
@@ -2584,6 +2584,29 @@ def dynamic_reshape(fd: FusionDefinition) -> None:
...
+                contiguity=[True, None, None, None],
+                dtype=DataType.Double,
+                is_cpu=False,
+                stride_order=[0, 3, 2, 1],
+            )
+            S1 = fd.define_scalar(2.0, dtype=DataType.Double)
+            T2 = fd.ops.mul(T1, S1)
+            fd.add_output(T2)
+
+        nvf_out, _ = self.exec_nvfuser(fusion_func, inputs)
```
jacobhinkle, line 2606: Does this check strides?

jacobhinkle (approved):
Looks great. Thanks for the update!

python_tests/test_python_frontend.py
jjsjann123: I don't think the stride matters here. i.e. we are not specifying stride on output, so codegen is free to do what it sees fits.

python_tests/test_python_frontend.py
jjsjann123: I thought I needed this in order to trigger concretization, i.e. `concretizeEmptyExtents` in order to enter the code path.

jjsjann123:
looks like we accidentally added bfloat16 on python test, which is now failing on v100. I'll patch that.

jjsjann123:
Patched in #1365.

This PR should be good by itself. I'm merging with broken CI.

python_tests/test_python_frontend.py
wujingyue: I had the same question :) I think it's worth at least a code comment. It'd be even better if we can verify concretization is indeed triggered. This way, **if** later we no longer concretize empty extents, we'll get notified and change the test to trigger concretization in a different way.

csrc/dynamic_transform.cpp
```diff
@@ -753,6 +753,9 @@ void DynamicTransformConcretizer::mutate(TensorDomain* td) {
       ? updateIdVec(td->maybeRFactor())
       : std::vector<IterDomain*>();
   std::vector<IterDomain*> domain = updateIdVec(td->leaf());
```
wujingyue, line 755:
```suggestion
  std::vector<IterDomain*> leaf_domain = updateIdVec(td->leaf());
```

csrc/dynamic_transform.cpp
```diff
@@ -753,6 +753,9 @@ void DynamicTransformConcretizer::mutate(TensorDomain* td) {
...
+  std::vector<IterDomain*> alloc_dom = td->hasAllocation()
+      ? updateIdVec(td->maybeAllocation())
```
wujingyue, line 757:
```suggestion
      ? updateIdVec(td->allocation())
```

csrc/dynamic_transform.cpp
```diff
@@ -753,6 +753,9 @@ void DynamicTransformConcretizer::mutate(TensorDomain* td) {
       ? updateIdVec(td->maybeRFactor())
```
wujingyue, line 753:
```suggestion
      ? updateIdVec(td->rfactor())
```

csrc/dynamic_transform.cpp
```diff
@@ -761,8 +764,16 @@ void DynamicTransformConcretizer::mutate(TensorDomain* td) {
   // Update the contiguity vector. Drop the contig val if mutated to broadcast
   auto contig = td->contiguity();
 
-  for (const auto i : c10::irange(td->maybeRFactor().size())) {
-    auto original_id = td->maybeRFactor().at(i);
+  const auto& new_alloc = td->hasAllocation() ? alloc_dom
```
wujingyue, line 767:
```suggestion
  const auto& new_maybe_alloc = td->hasAllocation() ? alloc_dom
```

to distinguis it from `alloc_dom`, the real new allocation domain.

wujingyue (approved):

python_tests/test_python_frontend.py
jacobhinkle: > I thought I needed this in order to trigger concretization, i.e. `concretizeEmptyExtents` in order to enter the code path.

I see. Yes that works. From C++ we could use a dynamic reshape but this way exercises the python alloc domain stuff too.

PR #1364, opened 2023-11-22
Support boolean inputs for `kir::Asm` and lower `cp.async` as `kir::Asm`

zasdfgbnm:
Inline PTX does not have a builtin constraint for boolean values, we have to convert the boolean to int first, and use `setp.ne.b32` to put this int to predicate register. With this new feature added, we can now lower `cp.async` as `kir::Asm`. This feature is also needed for supporting Hopper MMA.

### DIFF
--- a/csrc/codegen.cpp
+++ b/csrc/codegen.cpp
@@ -2864,7 +2864,8 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
       for (auto input : asm_->inputs()) {
         if (input->dtype() == DataType::Bool) {
           indent() << "\"  .reg .pred p" << boolean_counter << "; \\n\"\n";
-          indent() << "\"  setp.ne.b32 p" << boolean_counter << ", %" << counter << ", 0;\\n\"\n";
+          indent() << "\"  setp.ne.b32 p" << boolean_counter << ", %" << counter
+                   << ", 0;\\n\"\n";
           boolean_counter++;
         }
         if (std::holds_alternative<ArrayType>(input->dtype().type)) {
--- a/csrc/codegen.cpp
+++ b/csrc/codegen.cpp
@@ -577,28 +577,6 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
     code_ << genVariableName(tv);
   }
 
-  // Utility function to emit a cp.async intrinsic
-  void genCpAsync(const LoadStoreOp* ldst, size_t vec_size) {
-    auto dtype = ldst->in()->getDataType().value();
-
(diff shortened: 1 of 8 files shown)


zasdfgbnm:
!build

zasdfgbnm:
!build

naoyam (approved):
LGTM

csrc/codegen.cpp
```diff
@@ -2860,12 +2828,38 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
...
 
-    code_ << "\"" << asm_->code();
+    if (asm_->hasBooleanInput()) {
+      code_ << "\"{\\n\"\n";
+      int64_t boolean_counter = 0;
+      int64_t counter = 0;
+      for (auto input : asm_->inputs()) {
+        if (input->dtype() == DataType::Bool) {
+          indent() << "\"  .reg .pred p" << boolean_counter << "; \\n\"\n";
+          indent() << "\"  setp.ne.b32 p" << boolean_counter << ", %" << counter
```
drzejan2, line 2838: Sorry for late review of these changes.

I try to wrap my mind around this change. Can you share some explanation to this?

My understanding so far is:

1st use case:

If the first input object is of bool type then predicate will be set to false (`boolean_counter` and `counter` will be both `0`). Then, if there are any Boolean inputs that follows then all of them will have predicate set to true, as `counter` is not increased (stays `0`) while `boolean_counter` is changed hence `ne` check will set predicated to true.

2nd use case:

If after first input, that is of Boolean type, the second input is non-Boolean input and third is Boolean type then all Boolean inputs will have predicates set to false:
- 1st input, Boolean: `counter` (0), `boolean_counter` (0) -> `false` predicate
- 2nd input, Int: `counter` (0), `boolean_counter` (1) -> skip
- 3rd input, Boolean: `counter` (1), `boolean_counter` (1) -> `false` predicate

Is this an expected scenario, or I missing some crucial piece here?

zasdfgbnm: We are creating one predicate for each boolean inputs. The `boolean_counter` is the predicate name, the `counter` is variable name. For example, you might need to do `setp.ne.b32 p1 %3`, for this case, `1` is the `boolean_counter`, and `3` is the counter.

drzejan2: Oh, you are right, I assumed there is a space between `p` and `boolean_counter` in
```
indent() << "\"  setp.ne.b32 p" << boolean_counter << ", %" << counter
```
but now, as you pointer it out, things are clear. Earlier I tried to understand how a single predicated could be used for multiple parameters.

Sorry for the noise!

PR #1373, opened 2023-11-24
More cleanup on matmul schedule

zasdfgbnm:
There should be no behavioral change, just code movement and removal of unused code.
- `scheduleTuringMmaWarpOutput` is removed. Its code is moved to `scheduleMmaWarpOutput`. Because after reading the documentation of Hopper MMAs, I feel that Hopper can share the same schdule.
- `scheduleTuringOperandRead` is removed, and its code copied to `scheduleOperandRead`, similar as above.
- `scheduleLdMatrix` now takes an `MmaOptions`.
- `TensorView::applyMmaSwizzle` will call `WarpMmaSwizzler::scheduleLdMatrix` if the tensor is the consumer of ldmatrix, this removes the need for manually calling `WarpMmaSwizzler::scheduleLdMatrix` in unit tests.


zasdfgbnm:
!build

PR #1375, opened 2023-11-24
Make mma and ldmatrix unit tests really unit

zasdfgbnm:
Currently, our mma "unit" tests uses ldmatrix to load data to registers. This makes the "unit" test not really unit, because if we scheduled both ldmatrix and mma wrongly, as long as they are wrong in a consistent way, we can still get the correct result. Before https://github.com/NVIDIA/Fuser/pull/1303, we were indeed intentionally scheduling it wrongly but consistently to workaround the lack of allocation domain. But today, we should have everything we need to schedule it correctly, so in the unit test of mma, we should no longer use `ldmatrix` to read data to registers. Instead, we should read from gmem to registers directly, so that we can test if the memory format we scheduled is correct.

On the other hand, we need to add a unit test for ldmatrix. Similarly, we should not validate its results by using it with mma op. Instead, we should use `ldmatrix` to do a copy kernel, there should be no matmul involved.

This PR also cleans up some deprecated code.


csrc/scheduler/mma_utils.cpp
```diff
@@ -691,37 +691,6 @@ void WarpMmaSwizzler::scheduleLdMatrix(TensorView* tv, MmaOptions options) {
     //[2ko, 2mno, 8k, 8mni]                 [2ko, 1mno, 8k, 8mni]
   }
 
-  // ldmatrix loads multiple 8x8 matrices from shared memory to registers in a
```
zasdfgbnm, line 694: According to the unit test result, this should be handled in the allocation domain instead of the leaf domain

zasdfgbnm:
!build

zasdfgbnm:
!build

csrc/device_lower/validation.cpp
```diff
@@ -950,68 +950,6 @@ void validateMmaTensors(MmaOp* mma) {
       "MMA op cannot directly take double buffered register input, put a set stage before.");
 }
 
-//! Note and TODO:
-//!   Currently relying on ldmatrix to
-//!     obtain the correct data layout for turing/ampere
-//!     mma's.
-//!   This restriction will eventually not
-//!    be necessary once the scatter swizzle is ready.
-void validateTuringMmaInput(TensorView* tv) {
```
zasdfgbnm, line 959: These validation does not make sense. Although ldmatrix and mma ops are often used together, they are not married. They can be used alone without the other.

drzejan2 (approved):
Changes look good - I left some minor comments regarding changes in the comments/documentation.

PR #1381, opened 2023-11-27
Fix logic error in PromoteReuseSyncModifier

jacobhinkle:
This fixes an issue that can appear when using smem for matmul epilogue. It is possible to request promotion of prologue reuse and not wind up re-using it currently. The reason is that we promote by inserting block syncs in between lifetimes of smem allocations. The logic for this sync insertion step currently assumes that there is an expression that corresponds to each "position" as determined by `ScopeMap`. However, `ScopeMap` actually increments the position at the end of each for loop, in order to distinguish outer last reads from the last inner expression position in the corresponding loop nest. We can wind up with an inter-lifetime sync interval that exists entirely within these skipped endfor positions, in which case the current code will not insert syncs, meaning we do not re-use memory.

See the `SkippedSyncInterval` test, which fails on `main`, for an example showing how parallelization of outer dimensions can shrink outer lifetime intervals such that they get lost in the skipped endfor region.

This PR fixes this logic error by ensuring that any skipped position corresponding to a last read (i.e. the start of an inter-lifetime interval) is processed properly, translating the end of that interval to the next non-skipped position in case the interval also ended within the skipped range. It also checks that the ends of sync intervals are not skipped.


jacobhinkle:
!build --diff

jacobhinkle:
The errors are unrelated #1407. Codediff of binary tests shows new tests, but that is because one of the commits died prematurely with OOM.

csrc/device_lower/pass/alias_memory.cpp
```diff
@@ -1925,16 +1925,50 @@ class PromoteReuseSyncModifier : private kir::ExprMutator {
   using kir::ExprMutator::dispatch;
 
   void dispatch(Expr* expr) final {
-    auto position = allocation_info_map_.getScopeMap().getExprPos(expr);
+    int prev_position = position_;
+    position_ = allocation_info_map_.getScopeMap().getExprPos(expr);
+
+    if (position_ - prev_position > 1) {
+      for (auto [start, end] : sync_intervals_) {
+        if (start >= prev_position && start < position_ - 1) {
```
naoyam, line 1933: Trying to remember the overall logic here, but shouldn't this be:

```
if (start >= prev_position && end< position_ - 1) {
```

I'm assuming this condition corresponds to:
> If a sync interval is contained entirely within the skipped expression range

naoyam:
Can you please add a unit test?

csrc/device_lower/pass/alias_memory.cpp
jacobhinkle: Sorry it's not very clear. I will update the comment.

This condition is just for whether we have skipped the beginning of the sync interval; this block is meant to ensure they are added to `upcoming_first_writes_`. In the next block, we cover cases where we skipped the end of the interval, whether or not we also skipped the beginning, and we translate those so that they are caught in the subsequent (existing) checks.

test/test_smem_reuse.cpp
```diff
@@ -415,4 +417,90 @@ TEST_F(SmemReuseTest, MultiplePromoteReuse) {
...
+  // By parallelizing the first loop, we remove its appearance in the kernel.
+  //
+  //   6  FOR threadIdx.x in ithreadIdx.x3{5}:
+  //   7    FOR i85 in iS4{6}:
+  //   8      FOR i86 in iS5{7}:
+  //   9        T0_s[ ithreadIdx.x0{5}, iS1{6}, iS2{7} ] ca_pos( 3 )
+  //      = full({5, 6, 7}, ( (float)(1) ));
+  //   10        T1_l[ ithreadIdx.x3{5}, iS4{6}, iS5{7} ] produce_pos( 3 )
+  //      = -T0_s[ ithreadIdx.x0{5}, iS1{6}, iS2{7} ] ca_pos( 3 );
+  //   14  FOR threadIdx.x in ithreadIdx.x6{5}:
```
zasdfgbnm, line 460: Why this is 14 instead of 11? Is it because there are three `END` for the three for loop? Could you explicitly write them out for clarity?

jacobhinkle: Yes you got it. Good idea, I will add in the "endfor" positions for clarity.

test/test_smem_reuse.cpp
```diff
@@ -415,4 +417,98 @@ TEST_F(SmemReuseTest, MultiplePromoteReuse) {
...
+  //   22    FOR i83 in iS11{6}:
+  //   23      FOR i84 in iS12{8}:
+  //   24        T3_g[ ithreadIdx.x10{5}, iS11{6}, iS12{8} ]
+  //      = Set( T2_s[ ithreadIdx.x6{5}, iS7{6}, iS9{8}rf ], cache_op=Streaming)
+  //   25      ENDFOR i84
+  //   26    ENDFOR i83
+  //   27  ENDFOR threadIdx.x
+  //
+  // Parallelized loops still occupy a "position" in the expression list; i.e.
+  // the ENDFOR of the ithreadIdx.x3{5} loop is at position 13 (not shown).
```
zasdfgbnm, line 482: nit:
```suggestion
  // the ENDFOR of the ithreadIdx.x3{5} loop is at position 13.
```

zasdfgbnm (commented):
IIUC, we are missing a sync because we do not have `handle(ENDFOR)`? If this is the case, will it be more natural to just override `handle(kir::ForLoop)` to emulate a `handle(ENDFOR)`?

jacobhinkle:
> IIUC, we are missing a sync because we do not have `handle(ENDFOR)`? If this is the case, will it be more natural to just override `handle(kir::ForLoop)` to emulate a `handle(ENDFOR)`?

Good point. Originally I was also a little suspicious that earlier passes might have removed some exprs or replaced them without replacing their entries in the `AllocationInfoMap`. However, I think this actually can only occur because of ENDFOR positions. I will try an alternative where I fake an ENDFOR and report back.

jacobhinkle:
> I will try an alternative where I fake an ENDFOR and report back.

The latest commit implements this and I think it's more clear. I also added more comments. This implementation checks that positions are processed sequentially with no gaps. If gaps do appear then they are unexpected and should be considered a bug. Thanks for the suggestion!

jacobhinkle:
!build

zasdfgbnm (approved):
LGTM, thanks for fixing!

PR #1382, opened 2023-11-27
Call maybeClearAllocator in FusionReshapeVectorize

jacobhinkle:
This avoids an OOM when running this test in CI on V100, which allocates 1GB of memory. See here for a job that failed in this way: https://github.com/NVIDIA/Fuser/actions/runs/7005955828/job/19056753333.


jacobhinkle:
!build

xwang233 (approved):

PR #1383, opened 2023-11-27
Fix PTX false positives in codediff

jacobhinkle:
I think this should fix the false positives like those in https://nv/e2E/75523191

### DIFF
--- a/tools/codediff/diff_report.py
+++ b/tools/codediff/diff_report.py
@@ -650,6 +650,9 @@ def sanitize_ptx_lines(lines: list[str]) -> list[str]:
         l = re.sub(
             r"(_tmp_kernel|nvfuser)(_[a-z]+)?_(\d+|f\d+_c\d*_r\d+_g\d+)", "kernel", l
         )
+        # This part standardizes the _ZN76_ part which can change when more
+        # digits are needed for later parts of the mangled name
+        l = re.sub(r"_ZN\d+_", "_ZN76_", l)
         # This part removes the hash and timestamp of the cu file
         l = re.sub(r"kernel_cu_[0-9a-z]{8}_\d{5}", "kernel_cu_", l)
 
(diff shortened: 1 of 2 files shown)


jacobhinkle:
!build --diff

xwang233 (approved):

jacobhinkle:
!build --diff

jacobhinkle:
!build --diff

jacobhinkle:
@xwang233 any idea why the code diff jobs keep dying? Could this be OOM also?

jacobhinkle:
!build --diff

xwang233:
> @xwang233 any idea why the code diff jobs keep dying? Could this be OOM also?

The codegen-diff log is too long, so the extra log is dumped in the job artifact named `extra_log_dump_*.txt`. Sorry for the inconvenience.

E.g. http://nv/e6K. This is the error message. Is that related to your change?
```
================= 852 passed, 4 warnings in 567.53s (0:09:27) ==================
Previous HEAD position was 3864970 Print prefix to distinguish from other schedulers' log. (#1392)
Switched to branch 'temp_branch'
  File "/opt/pytorch/nvfuser/tools/codediff/diff_report.py", line 908, in <module>
    td = TestDifferences(
  File "<string>", line 9, in __init__
  File "/opt/pytorch/nvfuser/tools/codediff/diff_report.py", line 759, in __post_init__
    sanitize_ptx_lines(kern1.ptx.splitlines()),
  File "/opt/pytorch/nvfuser/tools/codediff/diff_report.py", line 655, in sanitize_ptx_lines
    namelen = int(d["namelen"])
KeyError: 'namelen'
```

xwang233:
After the kernel dumped file naming changed from `__tmp_kernel*.cu` to `__tmp_kernel_*.cu`, without the trailing underscore, there might be some issue in the CI setup for `--diff-bench`, but `--diff` should still work.

jacobhinkle:
> E.g. http://nv/e6K. This is the error message. Is that related to your change?

It is!! Thank you for pointing this out! OK I can fix this.

jacobhinkle:
!build --diff

jacobhinkle:
I think the only failures now are due to known OOM issue in binary tests, along with some legitimate code differences due to not having the branch updated (we should compare against `git merge-base origin/main HEAD` instead of `origin/main` probably. Also the H100 binary test failed to launch but I think that happens sometimes. I'll go ahead and merge since I think the codediff script itself is working now.

PR #1388, opened 2023-11-27
AliasAnalysis handles basic `Set` (no permute).

wujingyue:


wujingyue:
!build

wujingyue:
Some unit tests are failing because they assume `consumer=set(producer)` triggers the pointwise scheduler, which is no longer true after this PR. Fixing those tests...

wujingyue:
!build

test/test_alias.cpp
```diff
@@ -108,6 +108,29 @@ TEST_F(AliasAnalysisTest, View_MergeNonContiguous) {
...
+TEST_F(AliasAnalysisTest, Set) {
+  Fusion fusion;
+  FusionGuard fg(&fusion);
+
+  const std::vector<int64_t> in_shape({2, 3, 5});
+
+  TensorView* in = makeContigConcreteTensor(in_shape);
+  fusion.addInput(in);
+  TensorView* out = set(in);
+  fusion.addOutput(out);
```
jjsjann123, line 120: nitpick: can we also add a test where we have output stride set that's not consistent/compatible with the input stride and would block aliasing?

jjsjann123 (approved):
Thanks for cleaning-up and separating the tests. :bow:

test/test_alias.cpp
wujingyue: Done. Note that AliasAnalysis itself still produces the alias with a preferred allocation domain; it's MarkAliasPass that checks the allocation domain is compatible.

wujingyue:
!build

jjsjann123:
:shipit:

PR #1390, opened 2023-11-28
Suppress NVRTC warnings about unused functions

jacobhinkle:
When we encounter a compilation error, or if we use `NVFUSER_DUMP=ptxas_verbose`, we get a ton of warnings printed to screen like 
```
__tmp_kernel_none_f0_c0_r0_g0.cu(2586): warning #177-D: function "<unnamed>::max(int, <unnamed>::int64_t)" was declared but
 never referenced                                                                                                          
  __device__ constexpr int64_t max(int a, int64_t b) {
```
There are of course a ton of unused functions in the preamble so these are not particularly useful warnings. This PR just turns them off.


naoyam (approved):
LGTM

PR #1394, opened 2023-11-28
Expand patch on symbolic iterdomain

### LINKED ISSUE #1393: dynamic reshape API failed to properly set broadcast dimensions in reshape.
```
TEST_F(NVFuserTest, Playground) {
  FusionGuard fg(fusion);

  auto tv0 = TensorViewBuilder()
                 .ndims(2)
                 .shape({-1, -1})
                 .contiguity({true, std::nullopt})
                 .expanded({false, true})
                 .build();
  auto tv1 = TensorViewBuilder()
                 .ndims(2)
                 .shape({-1, -1})
                 .contiguity({std::nullopt, true})
                 .expanded({true, false})
                 .build();
  fusion->addInput(tv0);
  fusion->addInput(tv1);

  auto tv2 = add(tv0, tv1);
  auto s0 = IrBuilder::create<Val>(3);
  auto s1 = IrBuilder::create<Val>(4);
  auto s2 = IrBuilder::create<Val>(1);
  auto s3 = IrBuilder::create<Val>(5);
  auto tv3 = reshape(tv2, {s0, s1, s2});
  // auto tv3 = reshape(tv2, {3, 4}, {3, 4, 1});
  auto tv4 = expand(tv3, {s0, s1, s3});
  fusion->addOutput(tv4);

  FusionExecutorCache fec(std::move(fusion_ptr));

  at::Tensor t0 = at::randn({3}, options).as_strided({3, 4}, {1, 0});
  at::Tensor t1 = at::randn({4}, options).as_strided({3, 4}, {0, 1});
  auto ref = t0.add(t1).as_strided({3, 4, 5}, {4, 1, 0});

  std::vector<c10::IValue> aten_inputs({t0, t1});
  auto outputs = fec.runFusionWithInputs(aten_inputs);

  testValidate(fusion, outputs, {t0, t1}, {ref}, __LINE__, __FILE__);
}
```

I'm hitting an assert here
```
C++ exception with description "Invalid expand size, 5, for ?S10{( (nvfuser_index_t)(1) )}rf
Exception raised from expand at /opt/pytorch/nvfuser/csrc/ops/arith.cpp:1501 (most recent call first):
frame #0: nvfuser::nvfCheckFail(char const*, char const*, unsigned int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) + 0xf2 (0x7fe910da3c1c in /opt/pytorch/nvfuser/build/libnvfuser_codegen.so)
frame #1: nvfuser::expand(nvfuser::TensorView*, std::vector<nvfuser::Val*, std::allocator<nvfuser::Val*> > const&) + 0x9c6 (0x7fe91119af36 in /opt/pytorch/nvfuser/build/libnvfuser_codegen.so)
frame #2: <unknown function> + 0x4a0c45 (0x55c9ad94cc45 in ./build/nvfuser_tests)
frame #3: <unknown function> + 0x8149a1 (0x55c9adcc09a1 in ./build/nvfuser_tests)
frame #4: <unknown function> + 0x805a05 (0x55c9adcb1a05 in ./build/nvfuser_tests)
frame #5: <unknown function> + 0x806155 (0x55c9adcb2155 in ./build/nvfuser_tests)
frame #6: <unknown function> + 0x806c81 (0x55c9adcb2c81 in ./build/nvfuser_tests)
frame #7: <unknown function> + 0x80869c (0x55c9adcb469c in ./build/nvfuser_tests)
frame #8: <unknown function> + 0x806638 (0x55c9adcb2638 in ./build/nvfuser_tests)
frame #9: <unknown function> + 0x1f615c (0x55c9ad6a215c in ./build/nvfuser_tests)
frame #10: <unknown function> + 0x29d90 (0x7fe8daffdd90 in /usr/lib/x86_64-linux-gnu/libc.so.6)
frame #11: __libc_start_main + 0x80 (0x7fe8daffde40 in /usr/lib/x86_64-linux-gnu/libc.so.6)
frame #12: <unknown function> + 0x2490e5 (0x55c9ad6f50e5 in ./build/nvfuser_tests)
```

It looks like we are just not recognizing the "symbolic" broadcast from the `reshape` operation properly.

jjsjann123:
fixes #1393


jjsjann123:
!build

csrc/ops/arith.cpp
```diff
@@ -1492,6 +1492,17 @@ TensorView* expand(TensorView* inp, const std::vector<Val*>& expanded_sizes) {
       // Note that expansion to 1 just means its extent becomes 1 and
       // does not mean the ID becomes a broadcast.
       out_id_builder.extent(maybeCastOp(DataType::Index, expanded_sizes[i]));
+    } else if (
```
jjsjann123, line 1495: IIRC, the dynamic reshape API is added by @jacobhinkle

Does this change look reasonable to you?

jjsjann123:
!build

jjsjann123:
!build

jjsjann123:
:disappointed_relieved:  failure is out of memory, which doesn't seem to be related to the changes. I'm marking this as ready to review.

csrc/ops/arith.cpp
jacobhinkle: This is good for now since it fixes, but we should probably file an issue with a test that has dynamic shapes like 
```c++
  fusion->addInput(tv0);

  auto s0 = IrBuilder::create<Val>(DataType::Index);
  auto s1 = IrBuilder::create<Val>(DataType::Index);
  auto s2 = IrBuilder::create<Val>(DataType::Index);
  fusion->addInput(s0);
  fusion->addInput(s1);
  fusion->addInput(s2);
  fusion->addInput(s3);

  auto tv1 = reshape(tv0, {s0, s1});
  auto tv2 = expand(tv1, {s0, s2});

  fusion->addOutput(tv2);
```
The long-term solution would be to accept `Symbolic` inputs here in the definition op and also add an implementation for the method `ExpandOp::checkConcretization(Val* old_val, Val* new_val)` which checks that `old_val` is an `IterDomain` in the input that satisfies the necessary conditions.

naoyam: Why do we need to create a symbolic IterDomain with a concrete size of 1?

jjsjann123: Opened #1398 .

Would appreciate a review as well :stuck_out_tongue_winking_eye:

jjsjann123: > Why do we need to create a symbolic IterDomain with a concrete size of 1 in the first place?

That's a good question. We can probably do a clean up on the `reshape` logic when new sizes are just static.

naoyam: > That's a good question. We can probably do a clean up on the reshape logic when new sizes are just static.

Yeah, I think we could just create a broadcast domain if the extent is known to be 1, although this only matters with static reshape, so not a common case. The more fundamental fix would be to support symbolic domains with expand, as Jacob mentioned.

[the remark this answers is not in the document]
jjsjann123: yes it should. There's only 1 line of code change here where I was setting IterType to broadcast. But that should just be a no-op for that clause anyway.
`out_id_builder.iter_type(IterType::Broadcast);`

I left it intentionally unmerged, hoping it's easier to read. I'll change it.

jacobhinkle: Yeah, I think it's fine to create it as `Broadcast` for constant 1. We just need to be sure _some_ `IterDomain`s are `Symbolic` in the output in order to detect a dynamic reshape, but these broadcasts should be safe I think.

naoyam: > We just need to be sure some IterDomains are Symbolic in the output in order to detect a dynamic reshape

Yeah, that's what I also suspected. This would not work if all the shape extents of a reshape op are static. We would not need to create symbolic domains since we knew the actual sizes, but then we would miss the reshape would need to be concretized. I guess we would need to have a different algorithm to detect such a case, which shouldn't be too much of an issue, but again since this is a special case, I don't think we would need to prioritize.

csrc/ir/utils.cpp
```diff
@@ -858,6 +858,7 @@ class ValidateDomainEquivalence : private IterVisitor {
               derived_domain.end(),
               [&](auto id) {
                 return id->getIterType() == IterType::Symbolic ||
+                    id->getIterType() == IterType::Broadcast ||
```
jjsjann123, line 861: @jacobhinkle I don't think I'm doing the right thing here. But it asserts here with the changes... I think I missed something else.

csrc/ops/alias.cpp
```diff
@@ -148,13 +149,15 @@ TensorView* reshape(TensorView* inp_tv, const std::vector<Val*>& new_sizes) {
...
+    } else if (new_size->isConstScalar() && new_size->evaluate() == 1) {
+      iter_type = IterType::Broadcast;
     }
     if (new_size->dtype() != DataType::Index) {
       new_size = castOp(DataType::Index, new_size);
     }
     auto rf_id =
         IterDomainBuilder(FusionGuard::getCurFusion()->zeroVal(), new_size)
-            .iter_type(IterType::Symbolic)
+            .iter_type(iter_type)
```
jjsjann123, line 160: I'm now trying to mark the reshape output's iterdomain as broadcast when we see a static size-1 here.
But this triggered an assert in `ValidateDomainEquivalence`...

I'll try to dive into the call stack to figure out what else I missed.

csrc/ops/arith.cpp
jjsjann123: > We just need to be sure some IterDomains are Symbolic in the output in order to detect a dynamic reshape

> This would not work if all the shape extents of a reshape op are static. 

Thanks. I tried to special case on the constant 1 dimension for reshape. I think it's safe, as long as we have "some other scalar in the reshape that's not constant 1", we'll always have a Symbolic Iterdomain on the output, so a dynamic reshape will be detected.

jjsjann123:
!build

csrc/ops/arith.cpp
jjsjann123: Meanwhile, I'm tracking the general dynamic reshape + broadcast issue here: https://github.com/NVIDIA/Fuser/issues/1398

jjsjann123:
!build

csrc/ops/arith.cpp
jjsjann123: I thought about this earlier... if we are reshaping a tensor to `[1, 1, 1, ..., 1]`, that implies the input tensor would also needs to be in a `compatible` shape. So it needs to be a size-1 tensor anyway.
Given that we specialize tensor dimension with size-1 as broadcast, the extent would be a constant and the corresponding reshape would become a static reshape. https://github.com/NVIDIA/Fuser/blob/61092461a6abc66746947d98ede565d090e28267/csrc/ops/alias.cpp#L115-L118

Let me know if I missed anything.

jjsjann123:
!build

csrc/ops/alias.cpp
```diff
@@ -157,19 +167,23 @@ TensorView* reshape(TensorView* inp_tv, const std::vector<Val*>& new_sizes) {
...
-          rfactor_domain,
-          TensorDomain::getContiguityFilledWith(rfactor_domain, true)),
+          view_rfactor_domain,
+          view_rfactor_domain,
+          TensorDomain::getContiguityFilledWith(view_rfactor_domain, true)),
       inp_tv->dtype());
 
   IrBuilder::create<ViewOp>(inp_tv->container(), out_tv, inp_tv);
 
+  if (found_static_broadcast) {
```
jjsjann123, line 183: @jacobhinkle I think I did the right thing here now.

csrc/ops/arith.cpp
jjsjann123: > symbolic input domains are indeed just broadcast domains,

I *think* the goal is to mark any `static symbolic` domain that is a broadcast **as a broadcast domain**, instead of a symbolic domain.
So this PR only handles reshape. If we apply similar changes to `pad` & `slice`, we should be alright?

Let me see if I can make up some script to break `tryStaticReshape`.

If we are generally ok with the approach we are taking, do you mind if we follow up other operations with symbolic domains in separate issues? just so we can unblock the failing cases for now.

> I think this propagation of static shapes is possible and preferrable, but it might not be that trivial.

yeah going forward there will be more work needed to patch other operations that could give us symbolic domains.
We still need to support `dynamic` symbolic domains, once that is in, this `static` symbolic domain special handling for broadcasting becomes only an "optimization"?

jjsjann123: I was just referring to a symbolic IterDomain where the extent is just a constant.

csrc/ops/alias.cpp
jacobhinkle: Yeah I like this.

jjsjann123:
!build

naoyam (approved):
LGTM

jjsjann123:
There seems to be networks issue or something in the log? blindly retry CI...

jjsjann123:
!build

jjsjann123:
!build

jjsjann123:
!build

jjsjann123:
CI failure doesn't look like related.... As I'm also seeing the same error on another PR, which is totally not related here.

Issue #1386, opened 2023-11-27
Over estimation of persistent buffer size when project persistent buffer to its persistent producers


liqiangxl:
In normalization scheduler, if `project_persistent_buffers = True`, `PersistentBufferProjector` will project persistent tvs to inputs, otherwise it trys to project persistent tvs to its immediate persistent producers. See #419 and #343

However, the buffer size was over estimated in `scheduler_utils::persistentBufferSize`. This overestimation occurs because the potential to project persistent tvs onto their immediate persistent producers isn't factored into the calculations.

For example, in the following fusion, T7 and T11 are persistent, both of them can be projected to inputs. The current calculated `persistent_buffer_size` is based on the reduction dim of T7 and T11. The `projected_persistent_buffer_size`  is based on the correponding dim of inputs T0, T1, and T2.  If the inputs are fp16, the `projected_persistent_buffer_size` is smaller than `persistent_buffer_size` and `project_persistent_buffers = True`.
However, since T11 can be projected back to T7, the correct value of `persistent_buffer_size`  should only include the reduction dim of T7. Then, the `projected_persistent_buffer_size` is larger than `persistent_buffer_size` and the correct value of `project_persistent_buffers` should be `False`.
![image](https://github.com/NVIDIA/Fuser/assets/116412316/39cffca9-34d6-411c-9d33-c1c7449019e9)

liqiangxl:
fixed.

PR #1395, opened 2023-11-28
adjust persistent buffer size if can be projected to immediate producers

liqiangxl:
Issue: #1386
In normalization scheduler, if `project_persistent_buffers = True`, `PersistentBufferProjector` will project persistent tvs to inputs, otherwise it trys to project persistent tvs to its immediate persistent producers.
However, the buffer size was over estimated in `scheduler_utils::persistentBufferSize`. This overestimation occurs because the potential to project persistent tvs onto their immediate persistent producers isn't factored into the calculations.

Fix:
This PR fixed this overestimation by taking out the size of buffers which can be projected to its producers.

Influence:
The buffer size is corrected. Lead to smaller maxregcount generated by the heuristics, but no influence on generated code of the cpp benchmarks/tests (checked with `!build --diff --diff-bench`) since the current persistent batch size depends on elements count rather than buffer size.


liqiangxl:
!build --diff --diff-bench

naoyam:
!build --diff --diff-bench

csrc/scheduler/utils.cpp
```diff
@@ -915,6 +932,19 @@ PersistentBufferSizeReturn persistentBufferSize(
         std::max(max_proj_persistence_size, projected_buffer_size);
   }
 
+  // reduce max_persistence_size if a persistent buffer can be projected to
+  // other persistent tvs.
+  std::unordered_set<TensorView*> persistent_buffer_set(
```
naoyam, line 937: Persistent buffers are scoped, and the buffer size is the largest among multiple scopes. This reduction of the projection to immediate producers needs to be done with each scope.

Try to create an example with multiple scopes.

naoyam (changes requested):

liqiangxl:
!build --diff --diff-bench

csrc/scheduler/utils.cpp
liqiangxl: Good point. In the newly added test case, the largest buffer size comes from a scope which doesn't have a projecatable buffer. The other scope has a projecatable buffer. So the previous implementation overestimated the buffer size. In the revised version, the `persistent_mask` is used in the buffer size calculation. It is set to `true` when 
```
    if (!canProjectToPersistentProducer(
            buffer, producers, persistent_buffer_set)) {
      persistent_mask[buffer_i] = true;
    }
```
when the mask is `false`, the size is not counted.

liqiangxl:
For `nvfuser-ci/jit_codegen_diff_bench_17_5/5 — Failing (w/msg) `, the different PTX is expected since the maxregcount passed to NVRTC is changed.

test/test_gpu3.cpp
```diff
@@ -9409,6 +9413,90 @@ TEST_F(NVFuserTest, LoweringHook) {
...
+  auto t0 = at::randn({batch_size, hidden_size}, options);
+  auto t1 = at::randn({batch_size, hidden_size}, options);
+  auto t2 = at::randn({batch_size, hidden_size}, options);
+  std::vector<c10::IValue> inputs{t0, t1, t2};
+
+  // The persistent buffers in this fusion are: tv3, tv7, tv12, and tv17. Note
+  // that tv7 can be projected back to its producer, tv3. When calculating the
+  // total size of persistent buffers ([persistent_buffer_size]), it's important
+  // to consider the active scopes of these buffers. Simply subtracting the
+  // buffer size of tv7 from the total may lead to an overestimation. This is
```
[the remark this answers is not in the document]
liqiangxl: oh, you are right, underestimation.

naoyam (approved):
LGTM

PR #1397, opened 2023-11-28
Skip stride_order when it is a canonical row-major storage.

jjsjann123:
This would allow us to temporarily ignore the performance issue resulted from the lack of support on alloc_domain in vectorization support.

### DIFF
--- a/csrc/python_frontend/fusion_record.h
+++ b/csrc/python_frontend/fusion_record.h
@@ -1256,7 +1256,7 @@ struct TensorRecord : RecordFunctor {
     // stride_order_, but currently alloc_domain support isn't ideal and could
     // prevent vectorization. Adding this workaround to restore performance.
     std::vector<int64_t> stride_order;
-    for (auto i = c10::range(stride_order_.size())) {
+    for (auto i : c10::range(stride_order_.size())) {
       if (stride_order_[i] != stride_order_.size() - i - 1) {
         // detect permutation in stride_order_, apply stride_order and break;
         stride_order = stride_order_;
--- a/csrc/python_frontend/fusion_record.h
+++ b/csrc/python_frontend/fusion_record.h
@@ -1257,7 +1257,7 @@ struct TensorRecord : RecordFunctor {
     // prevent vectorization. Adding this workaround to restore performance.
     std::vector<int64_t> stride_order;
     for (auto i : c10::irange(stride_order_.size())) {
-      if (stride_order_[i] != stride_order_.size() - i - 1) {
+      if (stride_order_[i] != (int64_t)(rank - 1 - i)) {
         // detect permutation in stride_order_, apply stride_order and break;
         stride_order = stride_order_;
         break;
--- a/csrc/python_frontend/fusion_record.h
+++ b/csrc/python_frontend/fusion_record.h
@@ -1251,13 +1251,26 @@ struct TensorRecord : RecordFunctor {
       is_expand[index] = is_broadcast && has_non_broadcast_size;
     }
 
+    // skip stride_order if it's alloc_domain is in the same order as with root
+    // domain. We don't need this and we should be able to just use
+    // stride_order_, but currently alloc_domain support isn't ideal and could
+    // prevent vectorization. Adding this workaround to restore performance.
+    std::vector<int64_t> stride_order;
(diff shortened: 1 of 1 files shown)


jjsjann123:
!build

csrc/python_frontend/fusion_record.h
```diff
@@ -1251,13 +1251,26 @@ struct TensorRecord : RecordFunctor {
       is_expand[index] = is_broadcast && has_non_broadcast_size;
     }
 
+    // skip stride_order if it's alloc_domain is in the same order as with root
```
wujingyue, line 1254:
```suggestion
    // skip stride_order if its alloc_domain is in the same order as with rfactor
```

csrc/python_frontend/fusion_record.h
```diff
@@ -1251,13 +1251,26 @@ struct TensorRecord : RecordFunctor {
...
+    // domain. We don't need this and we should be able to just use
+    // stride_order_, but currently alloc_domain support isn't ideal and could
```
wujingyue, line 1256: Can you point to the problematic code? Maybe some code is over nervous about allocation domain is specified at all rather than figure out it's actually trivial and has the same effect as an empty allocation domain?

wujingyue:
Also, how hard is it to write a unit test?

csrc/python_frontend/fusion_record.h
wujingyue: > I'll open an issue to track this when I get some better ideas.

Sure -- I'm fine with a TODO. Regardless whether you manage to root cause this, I'd open the issue first. This way, I can say `TODO(<issue id>)` here so readers have something to refer to :)

wujingyue:
> > Also, how hard is it to write a unit test?
> 
> 😛 you caught me being a slacker again.
> 
> It's actually a bit awkward to test this. I can do string comparison here relying on `NVFUSER_DUMP`. It's a bit ugly, since we need to have the test in python, meanwhile, we don't have API on the python side to query some fusion IR precisely, other than relying on the debug output with `FusionDefinition::_debug_output`.

I guess how to test this will be more clear once you figure out the root cause. Since this is related to vectorization, I suspect something like https://github.com/NVIDIA/Fuser/blob/9d73fa6e30d4751f98782188ddc939db5223c6c6/test/test_pointwise.cpp#L23 may work.

csrc/python_frontend/fusion_record.h
```diff
@@ -1251,13 +1251,26 @@ struct TensorRecord : RecordFunctor {
...
+    // prevent vectorization. Adding this workaround to restore performance.
+    std::vector<int64_t> stride_order;
```
wujingyue, line 1258: Given `stride_order_` must be a permutation, you can simply use https://en.cppreference.com/w/cpp/algorithm/adjacent_find to check it's sorted in descending order. It's semantically easier to follow and probably leads to shorter code.

wujingyue (approved):
I'm not a big fan of submitting code without test coverage. But if this is urgent enough, I don't mind you merging the PR first and root-causing the bug as the immediate next step.

jacobhinkle:
> It's actually a bit awkward to test this. 

Could this be implemented in `TensorViewBuilder` instead and tested in C++?

csrc/python_frontend/fusion_record.h
```diff
@@ -1251,13 +1251,26 @@ struct TensorRecord : RecordFunctor {
...
+    }
+
     auto tv = TensorViewBuilder()
                   .ndims(shape_.size())
                   .contiguity(contiguity_)
                   .shape(shape_)
                   .dtype(dtype_)
                   .expanded(std::move(is_expand))
-                  .strideOrder(stride_order_)
+                  .strideOrder(stride_order)
```
jacobhinkle, line 1273: Could we just accept `stride_order_` here (i.e. undo changes to this file), then perform the check above inside `TensorViewBuilder::build()` instead?

jjsjann123: good point. Let me do that instead.

jjsjann123:
Oops. turns out we have coordination issue and this PR now becomes critical for per measurement. I'm gonna merge it as-is and open up another issue (to track root-causing) and PR (address review comments here).

Issue #1389, opened 2023-11-28
Matmul nvfuser auto splitk benchmarks are too slow


xwang233:
... and caused many CI timeouts

Reproduce command 

```
/opt/pytorch/nvfuser/bin/nvfuser_bench --benchmark_repetitions=1 --benchmark_min_time=0.00000001 '--benchmark_filter=^NvFuserScheduler_Matmul/nvfuser_a.*' | ts -s
```

note that we only have 1 iteration for everything

some examples on A100
```
00:00:22 --------------------------------------------------------------------------------------------------------------------------------------------------
00:00:22 Benchmark                                                                                                        Time             CPU   Iterations
00:00:22 --------------------------------------------------------------------------------------------------------------------------------------------------
00:00:22 NvFuserScheduler_Matmul/nvfuser_auto_splitk_TT/M:128/N:128/K:65536/warps:4/stages:3/manual_time               3237 us         3463 us            1 /Launch_Parameters[block(2/2/32)/grid(108/1/1)/49664]
00:00:42 NvFuserScheduler_Matmul/nvfuser_auto_splitk_TT/M:128/N:256/K:65536/warps:4/stages:3/manual_time               2432 us         2643 us            1 /Launch_Parameters[block(2/2/32)/grid(54/2/1)/49664]
00:01:02 NvFuserScheduler_Matmul/nvfuser_auto_splitk_TT/M:128/N:384/K:65536/warps:4/stages:3/manual_time               1926 us         2134 us            1 /Launch_Parameters[block(2/2/32)/grid(36/3/1)/49664]
00:01:23 NvFuserScheduler_Matmul/nvfuser_auto_splitk_TT/M:128/N:512/K:65536/warps:4/stages:3/manual_time               1974 us         2183 us            1 /Launch_Parameters[block(2/2/32)/grid(27/4/1)/49664]
00:01:43 NvFuserScheduler_Matmul/nvfuser_auto_splitk_TT/M:128/N:768/K:65536/warps:4/stages:3/manual_time               1918 us         2130 us            1 /Launch_Parameters[block(2/2/32)/grid(18/6/1)/49664]
00:02:03 NvFuserScheduler_Matmul/nvfuser_auto_splitk_TT/M:128/N:1152/K:65536/warps:4/stages:3/manual_time              1818 us         2024 us            1 /Launch_Parameters[block(2/2/32)/grid(12/9/1)/49664]
[...]

PR #1408, opened 2023-11-29
adjust when should the persistent buffuers being projected to inputs

liqiangxl:
Issue:
In normalization scheduler, persistent buffuers are only allowed to project to inputs if the recalculation doesn't require reduction ops. With this restriction, the persistent buffer in softmax is not being projected to inputs and fp32 is required to store the buffter.

Fix: for details see the design [doc](https://docs.google.com/document/d/1hmdaoHnz3QSqmiZ-cymScsqobKHTTy7JBh5h97rVp6I/edit#heading=h.9ljkdipwwpi).

(1) **Removal of the existing limitation**.
(2) To minimize redundant recalculations, the buffer is now **projected onto the broadcasted tensor** following the initial reduction.
(3) Despite these changes facilitating the use of fp16 for persistent buffers without necessitating extra reductions, a performance decline has been observed in scenarios with smaller hidden sizes. Consequently, the third alteration in this PR is the **deactivation of projection in situations where the benefits of reducing buffer size do not outweigh the drawbacks of recalculations**. Given the challenge in precisely measuring these benefits and drawbacks prior to actual application, the approach adopted in this PR is based purely on empirical data derived from softmax experiments. It is implemented in a standalone funciton and can be extended to handle other cases, e.g. RNG ops.

Performance:
(1) 50% increase for cases with hidden size large than 20K.
(2) 5% decrease for cases with hidden size in range of 10~12K. (hope can be removed after revising the inner persistent heuristics)
image


liqiangxl:
!build --diff --diff-bench

liqiangxl:
!build --diff --diff-bench

liqiangxl:
1. Codegen diff is expected.
2. Failure on V100 is out of memory, not related to this PR.
```
GpuViewTest.FusionReshapeVectorize\n\nunknown file\nC++ exception with description \"CUDA out of memory. Tried to allocate 256.00 MiB. GPU 0 has a total capacity of 15.78 GiB of which 20.75 MiB is free.
```

xwang233:
There were some issues in codegen diff-bench CI scripts. I've fixed these. Feel free to try those tomorrow if needed.

liqiangxl:
!build --diff --diff-bench

naoyam:
Could you do the same experiments on H100?

liqiangxl:
> Could you do the same experiments on H100?

Similar results on H100.
image

csrc/scheduler/reduction_utils.cpp
```diff
@@ -732,8 +733,18 @@ class PersistentBufferProjector {
                 buffer) == projectable_persistent_buffers.end()) {
           continue;
         }
+        std::vector<Val*> vals_project_to = fusion_->inputs();
```
naoyam, line 736: Please add some more comments here. What the issue is, how it is addressed, any potential concern? An example would be also very helpful.

naoyam: Also, please be more explicit what are assumed. One important assumption here is that since this is just for normalizations, a reduction must be always followed by a broadcast.

naoyam: Also, if there's a path from a reduction output to the persistent buffer without having a broadcast, we would end up recomputing the reduction. I suspect this  pattern would still be a valid persistent fusion, although it should be uncommon in practice. I think we should check if all the paths between a reduction output to a persistent buffer contains a broadcast.

csrc/scheduler/reduction_utils.cpp
```diff
@@ -732,8 +733,18 @@ class PersistentBufferProjector {
...
+        auto dep_vals = DependencyCheck::getAllValsBetween(
+            {reduction_tvs.begin(), reduction_tvs.end()}, {buffer});
+        for (auto val : dep_vals) {
+          if (auto tv = dynamic_cast<TensorView*>(val)) {
+            if (tv->hasBroadcast()) {
+              vals_project_to.push_back(val);
```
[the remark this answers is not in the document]
liqiangxl: I added a case with three sequential normalization patterns, see `ProjectPersistentBufferToInputsAndBroadcastTvs`. The  broadcast tensor in the first normalization is not duplicated in the list. This list is built separately for each buffer.

liqiangxl:
!build --diff --diff-bench

liqiangxl:
!build --diff --diff-bench

liqiangxl:
Looks like the codgen_diff script still has some issues. see the err msg from `
nvfuser-ci/jit_codegen_diff_bench_17_3/5` and `nvfuser-ci/jit_codegen_diff_bench_17_5/5`. @xwang233 
```
+ python /workspace/pyt_nightly_ci_utils/webpage_deploy_wrapper.py
  File "/workspace/pyt_nightly_ci_utils/webpage_deploy_wrapper.py", line 8, in <module>
    url = deploy_html_to_internal_pages(pathlib.Path(sys.argv[1]).resolve().absolute().__str__())
IndexError: list index out of range
```

xwang233:
Sorry the error message seems confusing.

For diff-bench 3/5 job, there is actually a failure, which was fixed (reverted) in the latest main branch. Rebase with main would resolve that.

```python
  what():  !info.contig_alloc_ids.empty() INTERNAL ASSERT FAILED at "/opt/pytorch/nvfuser/csrc/executor_utils.cpp":372, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. No root ID found for vectorization with T51_g[ ithreadIdx.x498{( ceilDiv(( ceilDiv(( ceilDiv(( (( (( getMetaData(T0) )).logical_size ))[2] ), 4) ), 2) ), 1) )}_p, iS495{2}, iUS497{1}, iV494{4} ] ca_pos( 3 ) produce_pos( 3 ) and T52_l[ ithreadIdx.x492{( ceilDiv(( ceilDiv(( ceilDiv(( (( (( getMetaData(T0) )).logical_size ))[2] ), 4) ), 2) ), 1) )}_p, iS489{2}, iUS491{1}, iS488{4} ] ca_pos( 3 ) produce_pos( 4 )
Exception raised from validateAlignedVectorizeExtents at /opt/pytorch/nvfuser/csrc/executor_utils.cpp:372 (most recent call first):
frame #0: nvfuser::nvfCheckFail(char const*, char const*, unsigned int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) + 0xf3 (0x7f81ab20243b in /opt/pytorch/nvfuser/build/libnvfuser_codegen.so)
frame #1: nvfuser::nvfErrorFail(char const*, char const*, unsigned int, char const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) + 0x53 (0x7f81ab3475b3 in /opt/pytorch/nvfuser/build/libnvfuser_codegen.so)
frame #2: nvfuser::executor_utils::validateVectorizedTensors(nvfuser::kir::Kernel*, nvfuser::KernelArgumentHolder const&, std::vector<at::Tensor, std::allocator<at::Tensor> > const&, nvfuser::executor_utils::caching::ExecutorCompileTimeInfoCache*, nvfuser::ExpressionEvaluator&) + 0x1578 (0x7f81ab3a0c28 in /opt/pytorch/nvfuser/build/libnvfuser_codegen.so)
frame #3: nvfuser::FusionExecutor::initializeExecutorEntry(nvfuser::FusionExecutor::ExecutorEntry&, nvfuser::KernelArgumentHolder const&, nvfuser::LaunchParams const&, nvfuser::CompileParams const&, std::vector<at::Tensor, std::allocator<at::Tensor> > const&, nvfuser::DataType) + 0x1eb (0x7f81ab35657b in /opt/pytorch/nvfuser/build/libnvfuser_codegen.so)
frame #4: nvfuser::FusionExecutor::runFusion(nvfuser::KernelArgumentHolder&, nvfuser::LaunchParams const&, nvfuser::CompileParams, std::vector<at::Tensor, std::allocator<at::Tensor> >) + 0x4f5 (0x7f81ab35a0a5 in /opt/pytorch/nvfuser/build/libnvfuser_codegen.so)
frame #5: nvfuser::FusionKernelRuntime::runKernelWithInput(nvfuser::KernelArgumentHolder&, nvfuser::SegmentedGroup*) + 0x23a (0x7f81ab49e21a in /opt/pytorch/nvfuser/build/libnvfuser_codegen.so)
frame #6: nvfuser::FusionKernelRuntime::runSegmentsWithInputs(nvfuser::KernelArgumentHolder&) + 0xc47 (0x7f81ab4a3687 in /opt/pytorch/nvfuser/build/libnvfuser_codegen.so)
frame #7: nvfuser::FusionKernelRuntime::runWithInputs(nvfuser::KernelArgumentHolder&) + 0xa9 (0x7f81ab4a4179 in /opt/pytorch/nvfuser/build/libnvfuser_codegen.so)
frame #8: nvfuser::FusionExecutorCache::runFusionWithInputs(c10::ArrayRef<c10::IValue> const&, std::optional<nvfuser::PrimDataType>, std::optional<signed char>) + 0x406 (0x7f81ab4acc16 in /opt/pytorch/nvfuser/build/libnvfuser_codegen.so)
frame #9: <unknown function> + 0xfea86 (0x559f0e175a86 in /opt/pytorch/nvfuser/bin/nvfuser_bench)
frame #10: <unknown function> + 0xeee83 (0x559f0e165e83 in /opt/pytorch/nvfuser/bin/nvfuser_bench)
frame #11: <unknown function> + 0x150603 (0x559f0e1c7603 in /opt/pytorch/nvfuser/bin/nvfuser_bench)
frame #12: <unknown function> + 0x12f7bb (0x559f0e1a67bb in /opt/pytorch/nvfuser/bin/nvfuser_bench)
frame #13: <unknown function> + 0x1313c5 (0x559f0e1a83c5 in /opt/pytorch/nvfuser/bin/nvfuser_bench)
frame #14: <unknown function> + 0x1319e9 (0x559f0e1a89e9 in /opt/pytorch/nvfuser/bin/nvfuser_bench)
frame #15: <unknown function> + 0x12033b (0x559f0e19733b in /opt/pytorch/nvfuser/bin/nvfuser_bench)
frame #16: <unknown function> + 0x12277e (0x559f0e19977e in /opt/pytorch/nvfuser/bin/nvfuser_bench)
frame #17: <unknown function> + 0x122a1e (0x559f0e199a1e in /opt/pytorch/nvfuser/bin/nvfuser_bench)
frame #18: <unknown function> + 0x5da54 (0x559f0e0d4a54 in /opt/pytorch/nvfuser/bin/nvfuser_bench)
frame #19: <unknown function> + 0x29d90 (0x7f8175537d90 in /usr/lib/x86_64-linux-gnu/libc.so.6)
frame #20: __libc_start_main + 0x80 (0x7f8175537e40 in /usr/lib/x86_64-linux-gnu/libc.so.6)
frame #21: <unknown function> + 0x7a285 (0x559f0e0f1285 in /opt/pytorch/nvfuser/bin/nvfuser_bench)
```

For diff-bench 5/5 job, there is a failure in codegen-diff script, maybe @jacobhinkle can help take a look. Full job log http://nv/e6V 
```python
WARNING: Test NvFuserScheduler_Softmax_Inner_fp16___GRAPH/NvFuserScheduler_Softmax_Inner_fp16/40960/8192/manual_time has different number of kernels in /opt/pytorch/nvfuser/codegen_comparison/6b9e5b6/custom_command_20231206_043215 than in /opt/pytorch/nvfuser/codegen_comparison/cd8481f/custom_command_20231206_043215. Not showing diffs for this test.
  File "/opt/pytorch/nvfuser/tools/codediff/diff_report.py", line 908, in <module>
    td = TestDifferences(
  File "<string>", line 9, in __init__
  File "/opt/pytorch/nvfuser/tools/codediff/diff_report.py", line 751, in __post_init__
    kern2 = self.run2.get_kernel(testname, kernel_num, strip_preamble=True)
  File "/opt/pytorch/nvfuser/tools/codediff/diff_report.py", line 570, in get_kernel
    kern = self.kernel_map[test_name].kernels[kernel_number]
IndexError: list index out of range
```

naoyam:
Is it correct that this PR should not have any impact with non-softmax fusions?

csrc/scheduler/normalization_utils.cpp
```diff
@@ -813,6 +813,47 @@ void checkReductionTvForScheduling(Fusion* fusion, TensorView* ref_red_tv) {
...
+      has_exp_op = true;
+    }
+  }
+  if (!has_exp_op) {
+    return true;
+  }
+
+  // For cases with expensive ops, don't project if buffer size is 'small',
+  if (persistent_buffer_size_info.persistent_buffer_size <=
+      scheduler_utils::small_buffer_size_threshold) {
```
naoyam, line 850: Can we avoid using the magic value? The underlying issue is that if the non-projected persistent buffer is small enough, i.e., not affecting the occupancy, projecting back to the inputs isn't buying us anything. Can we infer when it becomes small enough?

naoyam: Also, can we just do this even without expensive operations? The perf difference would be small but it shouldn't result in lower performances, right?

naoyam:
I'm wondering why we need to find broadcast. The main factor here is we don't want to redo a reduction, so why not just caching the output of the reduction? Wouldn't it make more sense?

naoyam:
Does the persistent buffer size include the size of the cached broadcast tensor? Or is it just ignored?

naoyam:
> > Is it correct that this PR should not have any impact with non-softmax fusions?
> 
> If only consider the benchmark cases, then the answer is yes. Only cases with softmax are influenced by this PR. Generally, the answer is no. This PR allows the projection of persistent buffers whose recalculation requires reduction op, and the projection will happen if it can reduce buffer size, e.g. the added test case `ProjectPersistentBufferToInputsAndBroadcastTvs`.

Yeah, that was I wanted to confirm. So all the codegen diffs are with softmax tests and benchmarks?

csrc/scheduler/normalization_utils.cpp
liqiangxl: > Can we avoid using the magic value? The underlying issue is that if the non-projected persistent buffer is small enough, i.e., not affecting the occupancy, projecting back to the inputs isn't buying us anything. Can we infer when it becomes small enough?

(1) Probabally no. To calculate occupancy, we need to estimate register usage which is based on magic number (24 or 40 registers as overhead).
(2) Calculate occupancy is also non-trivial, we need to run the heuristics with project_to_input = True / False, then compare the change of occupancy.

liqiangxl: > Also, can we just do this even without expensive operations? The perf difference would be small but it shouldn't result in lower performances, right?

we can do that. You are right, the perf difference would be small. Maybe we just don't need to change it if the difference is small. we can leave it for a following work and this PR introduces this function as a placehoder and only improve softmax without causing regressions on other cases.

liqiangxl:
> I'm wondering why we need to find broadcast. The main factor here is we don't want to redo a reduction, so why not just caching the output of the reduction? Wouldn't it make more sense?

Only thread-0 has the right answer of the reduction results, it needs a broadcast for all threads to have the right answer. Also, it is just a float and already automatically cached in a register .

csrc/scheduler/normalization_utils.cpp
naoyam: > (1) Probabally no. To calculate occupancy, we need to estimate register usage which is based on magic number (24 or 40 registers as overhead).

Why does this matter? That magic number already exists. Why not avoid adding another if not strictly required?

liqiangxl:
> > > Is it correct that this PR should not have any impact with non-softmax fusions?
> > 
> > 
> > If only consider the benchmark cases, then the answer is yes. Only cases with softmax are influenced by this PR. Generally, the answer is no. This PR allows the projection of persistent buffers whose recalculation requires reduction op, and the projection will happen if it can reduce buffer size, e.g. the added test case `ProjectPersistentBufferToInputsAndBroadcastTvs`.
> 
> Yeah, that was I wanted to confirm. So all the codegen diffs are with softmax tests and benchmarks?

Yes. but here `nvfuser-ci/jit_codegen_diff_17_4/4` also has several other diffs because the baseline is old. In benchmark, a previously segmented softmax case is changed to unsegmented in this PR and `codegen-diff script` didn't generate the diff correctly.

naoyam:
> > I'm wondering why we need to find broadcast. The main factor here is we don't want to redo a reduction, so why not just caching the output of the reduction? Wouldn't it make more sense?
> 
> Only thread-0 has the right answer of the reduction results, it needs a broadcast for all threads to have the right answer. Also, it is just a float and already automatically cached in a register .

So, we also want to avoid redoing the broadcast as well since the broadcast here is not just a local register copy but involves an inter-thread communication. Could you leave comments describing this?

naoyam:
> > Does the persistent buffer size include the size of the cached broadcast tensor? Or is it just ignored?
> 
> Just ignored. This PR doesn't influence how the reduction output or broadcast tensor is stored. It is just one register.

Is it always just one register? In outer normalizations, IIRC, there can be multiple independent normalizations, so the allocation size of each intermediate tensor should be something like `vectorization_factor * batch_factor`.

liqiangxl:
> > > I'm wondering why we need to find broadcast. The main factor here is we don't want to redo a reduction, so why not just caching the output of the reduction? Wouldn't it make more sense?
> > 
> > 
> > Only thread-0 has the right answer of the reduction results, it needs a broadcast for all threads to have the right answer. Also, it is just a float and already automatically cached in a register .
> 
> So, we also want to avoid redoing the broadcast as well since the broadcast here is not just a local register copy but involves an inter-thread communication. Could you leave comments describing this?

added in comments.

liqiangxl:
!build --diff --diff-bench

naoyam:
> > > > Does the persistent buffer size include the size of the cached broadcast tensor? Or is it just ignored?
> > > 
> > > 
> > > Just ignored. This PR doesn't influence how the reduction output or broadcast tensor is stored. It is just one register.
> > 
> > 
> > Is it always just one register? In outer normalizations, IIRC, there can be multiple independent normalizations, so the allocation size of each intermediate tensor should be something like `vectorization_factor * batch_factor`.
> 
> You are right. For outer reduction the broadcast tv needs `vectorization_factor`, e.g. if vectorized by 4, it needs ` float[4].` But this buffer for broadcast tv was already there before this PR and was ignored when calculating the buffer size. I'll create an issue tracking this problem.

It's ignored because it isn't persistent, but now this PR may change that. In other words, the lifetime of the variable will be different, and when it becomes persistent, it needs to remain live in addition to the input persistent buffer. The impact may not be negligible, for example, when vectorized by 8 with a batch size of 4.

liqiangxl:
!build --diff --diff-bench

xwang233:
!build --diff --diff-bench

liqiangxl:
> !build --diff --diff-bench

@xwang233  Is the CI script updated? Looks like it still can't handle the case where number of segments is changed. I tested a fix and works fine on my local test. https://github.com/NVIDIA/Fuser/pull/1495

xwang233:
The CI script for binary tests was upgraded because of the binary test split by https://github.com/NVIDIA/Fuser/pull/1465 a few days ago. I restarted your pipeline because there was a script failure and I just patched.

For codegen diff, we didn't change any internal scripts, the `diff_report.py` change needs to be merged in your PR branch to take effect. You may rebase that change into your current PR for CI testing if that blocks you. Once that PR gets merged into main, the codegen diff changes won't show up in your PR diff's, I think.

liqiangxl:
!build --diff --diff-bench

liqiangxl:
> The CI script for binary tests was upgraded because of the binary test split by #1465 a few days ago. I restarted your pipeline because there was a script failure and I just patched.
> 
> For codegen diff, we didn't change any internal scripts. The `diff_report.py` change needs to be merged in your PR branch to take effect. You may rebase that change into your current PR for CI testing if that blocks you. Once that PR gets merged into main, the codegen diff changes won't show up in your PR diff's, I think.

Thanks for the info.
(1) The fix for `jit_codegen_diff_bench` is merged in main and works as expected.
(2) The failure in `nvfuser-ci/jit_codegen_diff_17_4/4 — Failing after 55 minutes https://nv/e2E/76932410` is due to
```
remote: GitLab: File "public_compressed/codegen_diff_p11385688_j76932410_1702232183433250832_codediff_f1741a9_55b64bd_custom_command_20231210_172237.html.tar.gz" is larger than the allowed size of 5 MiB. Use Git LFS to manage this file.
To https://gitlab-master.nvidia.com/dl/pytorch/fuser-gh-mirror/
 ! [remote rejected] web -> web (pre-receive hook declined)
error: failed to push some refs to 'https://gitlab-master.nvidia.com/dl/pytorch/fuser-gh-mirror/'
```
(3) The failure in `jit_codegen_diff_bench_17_TNVF 4/5` is due to `ERROR: Job failed: execution took longer than 1h45m0s seconds`. can we increase the allowed time?
(4) The failure in `jit_codegen_diff_bench_17_TNVF 5/5` is caused by code change in this PR and is expected.

xwang233:
2. If the compressed webpage size is still larger than 5 MB, we should reduce the number of kernels rendered on the webpage. The uncompressed webpage is likely much larger than 5 MB. Even if it's uploaded, browsers may have trouble viewing it. cc @jacobhinkle 

3. The nvfuser_bench 4/5 shard timeout is a known issue at here https://github.com/NVIDIA/Fuser/issues/1389. We should mark those benchmarks as "slow" in the source code instead of blindly increasing timeout limit, so that we can easier filter slow benchmarks in separate jobs. cc @zasdfgbnm 
    The timeout limit was already increased from 1h to 1h45m, and I think it's not a good idea to keep increasing that.

liqiangxl:
As discussed, added restriction to apply buffer threshold only to inner persistent scheduler. So, for outer softmax with FP16 inputs, the size of the persistent buffer is effectively halved, corresponding to the projection to input.
Up to 1.69x speedup is observed for reduction size of 1024 and segmentation is avoided for reduction size of 2048 (para is marked as NaN)  but no speedup is observed for this case. The tests were carried on A100.
image

Regression happens for cases with extremally small reduction sizes (2, 8, 16). This can be avoided by further tuning a buffer threshold for outer reductions, tracked in https://github.com/NVIDIA/Fuser/issues/1508
image

naoyam:
> As discussed, added restriction to apply buffer threshold only to inner persistent scheduler. So, for outer softmax with FP16 inputs, the size of the persistent buffer is effectively halved, corresponding to the projection to input. Up to 1.69x speedup is observed for reduction size of 1024 and segmentation is avoided for reduction size of 2048 (para is marked as NaN) but no speedup is observed for this case. The tests were carried on A100. image
> 
> Regression happens for cases with extremally small reduction sizes (2, 8, 16). This can be avoided by further tuning a buffer threshold for outer reductions, tracked in #1508 image

Thanks for the measurement. That seems to make sense to me.

naoyam (approved):
LGTM

PR #1409, opened 2023-11-29
Initial support for Hopper MMA instructions

zasdfgbnm:
This PR adds support for Hopper MMA instructions. Current limitations are:
1. Each smem tensor must be exactly one tile.
2. Synchronization are hard coded before and after the MMA instructions, therefore it is indeed synchronous. 
3. Only support swizzle mode "none", which is pretty different from other swizzle modes, but conceptually a bit simpler to start with.

In terms of testing, two new test suites are added: `MmaTest/Hopper.RS` (where A is on register and B on smem) and `MmaTest/Hopper.RS` (where both A and B are on smem). These two test suites will be instantiated for all Hopper mma instructions, for all layouts, for all swizzle modes, and for all dtypes (half and bf16), so there will be a great number of new tests added, but I believe this exhaustive test is necessary.

For review, I recommend briefly read through [PTX doc](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#asynchronous-warpgroup-level-matrix-multiply-accumulate-instructions) first before reading this PR.

Also, be warned that, the smem memory format is not straightforward at all (actually less straigntforward than register memory format in my opinion). Especially, I was originally expecting that the instruction tile for the no-swizzle mode will be just a simple row major or column major matrix, but this seems to be not the case. For the case of no-swizzle mode, you need to divide the instruction tile into 8x8 core matrices, and put each core matrix into contiguous memory, which is something like:
```C++
[M, K] -> [M/8, 8, K/8, 8] -> [M/8, K/8, 8, 8]
```
Also, another surprise to me is, the `imm-trans-A` and `imm-trans-B` is actually ignored by the `wgmma` instruction for no-swizzle mode. So regardless of your input layout, in smem, you need to manually transpose it to the K-major layout (`NT`, `[K, M] x [K, N]`) in smem before calling the `wgmma` instruction.

This PR is only initial support for Hopper MMA. Remaining things are:
- Support multi-tile
- Support other swizzle mode
- Be asynchronous

### DIFF
--- a/test/test_mma.cpp
+++ b/test/test_mma.cpp
@@ -270,7 +270,7 @@ TEST_P(Hopper, RS) {
   tv0b->axis(1)->parallelize(ParallelType::TIDx);
 
   tv1b->setMemoryType(MemoryType::Shared);
-  tv1b->applyMmaSwizzle(swizzle);
+  tv1b->applyMmaSwizzle(swizzle, transpose_b);
 
   if (!transpose_b) {
     // [M, K, N] -> [M, N, K]
@@ -402,9 +402,9 @@ TEST_P(Hopper, SS) {
   auto tv2c = tv2->cacheBefore();
 
   tv0b->setMemoryType(MemoryType::Shared);
-  tv0b->applyMmaSwizzle(swizzle);
+  tv0b->applyMmaSwizzle(swizzle, transpose_a);
   tv1b->setMemoryType(MemoryType::Shared);
-  tv1b->applyMmaSwizzle(swizzle);
+  tv1b->applyMmaSwizzle(swizzle, transpose_b);
 
   tv2c->applyMmaSwizzle(MmaOperand::Accumulator);
   tv2->applyMmaSwizzle(MmaOperand::Accumulator);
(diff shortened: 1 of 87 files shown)


zasdfgbnm:
!build

test/test_mma.cpp
```diff
@@ -163,4 +162,334 @@ INSTANTIATE_TEST_SUITE_P(
...
+  tv1b->setMemoryType(MemoryType::Shared);
+  tv1b->applyMmaSwizzle(swizzle, transpose_b);
+
+  if (transpose_a) {
+    // TODO: Why do we need to transpose B if A is transposed? I don't really
+    // understand why, but empirically it works...
+    auto alloc = tv1b->getAllocationDomain();
+    std::swap(alloc[alloc.size() - 1], alloc[alloc.size() - 2]);
+    tv1b->setAllocationDomain(alloc, true);
+  }
```
zasdfgbnm, line 412: Maybe it makes more sense to put this section to `applyMmaSwizzle`, but because I don't understand why we need this, I will leave it here for now. In the future, I may move code around as I get deeper understanding of how tensor core works.

zasdfgbnm:
!build

zasdfgbnm:
!build

zasdfgbnm:
!build

zasdfgbnm:
!build

PR #1413, opened 2023-11-29
s/DependencyCheck::getAllExprsBetween/StmtSort::getExprsBetween

wujingyue:
Per https://github.com/NVIDIA/Fuser/pull/1295#discussion_r1408614357.

### DIFF
--- a/csrc/iter_visitor.cpp
+++ b/csrc/iter_visitor.cpp
@@ -126,7 +126,7 @@ void IterVisitor::dispatch(Val* v) {
 
 // Implementation details:
 // We start with an entry in stmt_stack that is the outputs we want to
-// process. We cannot process these outputs untill all Stmts in their history
+// process. We cannot process these outputs until all Stmts in their history
 // have been processed, as those Stmts contain all dependencies to produce
 // these values. What we will do is traverse towards inputs until we hit a
 // leaf node. Once we hit a leaf node that node will be visited, then we will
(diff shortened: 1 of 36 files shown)


wujingyue:
!build

wujingyue:
!build

naoyam (commented):
Looks like the main change is actually removal of a `Fusion` parameter from the traverse functions of `IterVisitor`. I think that's fine as I don't think of any case where we would have a different `Fusion` than the `Fusion` where a given Statement is contained.

@zasdfgbnm Any concern?

csrc/iter_visitor.cpp
```diff
@@ -987,22 +953,22 @@ std::vector<Val*> InputsOf::outputs(
     Fusion* fusion,
     const std::vector<Val*>& outputs_) {
   InputsOf io;
-  io.traverseTo(fusion, outputs_, false);
+  io.traverseTo(outputs_, false);
```
naoyam, line 956: It's a bit strange that the function has a `Fusion` parameter but it isn't used.

naoyam:
!build --diff --diff-bench

zasdfgbnm:
I don't see any use case where we want to use a different fusion either.

csrc/iter_visitor.cpp
wujingyue: Done.

naoyam (approved):
LGTM. Could you make sure the codegen diffs don't look anything suspicious? Since the PR changes the sorting API, it's expected that some expressions and variables may be ordered differently, but I don't expect no other change should happen.

wujingyue:
> LGTM. Could you make sure the codegen diffs don't look anything suspicious? Since the PR changes the sorting API, it's expected that some expressions and variables may be ordered differently, but I don't expect no other change should happen.

The [only codegen diff](https://dl.gitlab-master-pages.nvidia.com/pytorch/fuser-gh-mirror//nvfuser_github_ci/codegen_diff_p11152977_j75945727_1701374506504128869_codediff_6109246_c5f2f6e_custom_command_20231130_191144.html) is in integer multiplication order, which is fine. (I'd worry more about floating point multiplication order).

PR #1415, opened 2023-11-30
Disabling duplicated inputs

jjsjann123:
1. disabling duplicated inputs (user registering input multiple times). It's scary to think about what the expectation is, when user register a single input tensor multiple times. Disabling it for now.
2. adding a test case to verify the proper behavior for duplicated output, which is supported by codegen.

### DIFF
--- a/csrc/executor.cpp
+++ b/csrc/executor.cpp
@@ -984,6 +984,8 @@ std::vector<at::Tensor> allocateOutputs(
 
   for (const auto output_idx : c10::irange(output_info.size())) {
     Val* out = kernel->outputs()[output_idx];
+    // TODO: remove the else block and outputs_map when output aliasing is
+    // handled properly
     auto iter = outputs_map.find(out);
     if (iter == outputs_map.end()) {
       auto [aliased_in, alias_info] = kernel->getOutputAlias(out);
--- a/csrc/executor.cpp
+++ b/csrc/executor.cpp
@@ -979,26 +979,36 @@ std::vector<at::Tensor> allocateOutputs(
 
   std::vector<at::Tensor> outputs;
   outputs.reserve(output_info.size());
+
+  std::unordered_map<Val*, at::Tensor> outputs_map;
+
   for (const auto output_idx : c10::irange(output_info.size())) {
     Val* out = kernel->outputs()[output_idx];
-    auto [aliased_in, alias_info] = kernel->getOutputAlias(out);
-    at::Tensor aliased_in_tensor;
-    if (aliased_in != nullptr) {
-      const PolymorphicValue& aliased_in_val =
-          *inputs[IndexOfFusionInput(aliased_in, kernel)];
-      NVF_ERROR(
-          aliased_in_val.is<at::Tensor>(),
-          "Alias io only supports tensor. Found ",
-          PolymorphicValue_functions::toString(aliased_in_val));
-      aliased_in_tensor = aliased_in_val.as<at::Tensor>();
+    auto iter = outputs_map.find(out);
+    if (iter == outputs_map.end()) {
+      auto [aliased_in, alias_info] = kernel->getOutputAlias(out);
+      at::Tensor aliased_in_tensor;
+      if (aliased_in != nullptr) {
+        const PolymorphicValue& aliased_in_val =
+            *inputs[IndexOfFusionInput(aliased_in, kernel)];
+        NVF_ERROR(
+            aliased_in_val.is<at::Tensor>(),
+            "Alias io only supports tensor. Found ",
+            PolymorphicValue_functions::toString(aliased_in_val));
+        aliased_in_tensor = aliased_in_val.as<at::Tensor>();
+      }
+      auto output = allocateOutput(
+          output_info[output_idx],
+          aliased_in,
+          alias_info,
+          aliased_in_tensor,
+          device,
+          ee);
+      outputs_map[out] = output;
+      outputs.push_back(output);
+    } else {
+      outputs.push_back(iter->second);
     }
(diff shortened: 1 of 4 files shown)


csrc/fusion.cpp
```diff
@@ -242,6 +242,11 @@ void Fusion::addInput(Val* input) {
         "Immediate scalar value cannot be added as an input. It is not necessary to pass it as an input.");
   }
 
+  NVF_CHECK(
+      std::find(inputs_.begin(), inputs_.end(), input) == inputs_.end(),
```
wujingyue, line 246: I think this is equivalent and shorter. 

```suggestion
      input->isFusionInput(),
```

test/test_alias.cpp
```diff
@@ -588,4 +588,55 @@ TEST_F(AliasTest, Set_NoAliasForIncompatibleLayout) {
...
+
+  // Verify aliases among outputs.
+  EXPECT_TRUE(out_tensors[0].is_alias_of(out_tensors[1]));
+  EXPECT_FALSE(out_tensors[0].is_alias_of(out_tensors[2]));
+  EXPECT_TRUE(out_tensors[0].is_alias_of(out_tensors[3]));
+
+  // Verify output values.
+  testValidate(
+      fec.fusion(),
+      {inc_5_tensor, inc_5_tensor, inc_1_tensor, inc_5_tensor},
```
wujingyue, line 622:
```suggestion
      out_tensors,
```

test/test_alias.cpp
```diff
@@ -588,4 +588,55 @@ TEST_F(AliasTest, Set_NoAliasForIncompatibleLayout) {
...
+  fusion->addOutput(out);
+  TensorView* out1 = add(in, IrBuilder::create<Val>(1.0));
+  fusion->addOutput(out1);
+  // duplicated output
+  fusion->addOutput(out);
+
+  FusionExecutorCache fec(std::move(fusion));
+  at::Tensor in_tensor = at::randn({2, 3, 5}).cuda();
+  auto inc_5_tensor = in_tensor.add(5.0);
+  auto inc_1_tensor = in_tensor.add(1.0);
```
wujingyue, line 609: Thanks to @Priya2698 , we don't have to pass in reference outputs as long as `ExpressionEvaluator` can compute them. 

```suggestion
```

test/test_alias.cpp
```diff
@@ -588,4 +588,55 @@ TEST_F(AliasTest, Set_NoAliasForIncompatibleLayout) {
...
+
+TEST_F(AliasTest, DuplicatedInputs) {
+  EXPECT_THAT(
+      []() {
+        auto fusion = std::make_unique<Fusion>();
+        FusionGuard fg(fusion.get());
+
+        TensorView* in = makeContigConcreteTensor({2, 3, 5});
+        fusion->addInput(in);
+        fusion->addInput(in);
```
wujingyue, line 636: I'd wrap only the second `addInput` to be precise.

wujingyue (approved):

test/test_alias.cpp
jjsjann123: Oops, I didn't realize we have an overload there. Thanks.

csrc/fusion.cpp
jjsjann123: I was debating on this earlier. `isFusionInput` only checks on the `is_fusion_input_` flag and I figured that we are really wanting to avoid duplicate in the `inputs_` vector.

Admittedly it's supposed to be equivalent. :man_shrugging: I'll switch for cleaner code.

jjsjann123:
Oh no. The duplicated outputs are totally not working after I fixed the tests.... :face_with_head_bandage: 
I'll take another look at this.

csrc/executor.cpp
```diff
@@ -979,26 +979,38 @@ std::vector<at::Tensor> allocateOutputs(
...
+          output_info[output_idx],
+          aliased_in,
+          alias_info,
+          aliased_in_tensor,
+          device,
+          ee);
+      outputs_map[out] = output;
+      outputs.push_back(output);
+    } else {
+      outputs.push_back(iter->second);
```
jjsjann123, line 1012: cc'ing @wujingyue 

I didn't realize that we don't have output aliasing output supports yet. I added this part so that we re-use buffer for duplicated outputs. We would want to remove this when your alias update on output is plumbed through.

If you are ok with this I'll move forward with this PR then.

jjsjann123:
!build

jjsjann123:
Pinging @naoyam on this one.

Are you ok with us disabling duplicated inputs? I don't see a good reason to have in the first place, also it's a bit confusing to what the protocol is to me.

Also, this PR fixed the duplicated output issue we discussed in the old PR. You can look at the added example for some reference

jjsjann123:
!build

naoyam:
> Pinging @naoyam on this one.
> 
> Are you ok with us disabling duplicated inputs? I don't see a good reason to have in the first place, also it's a bit confusing to what the protocol is to me.
> 
> Also, this PR fixed the duplicated output issue we discussed in the old PR. You can look at the added example for some reference

Yes, I think this is fine as I don't think of any reason that we would need to support duplicated inputs.

csrc/executor.cpp
wujingyue: > so that we re-use buffer for duplicated outputs

I'd remove this change from the PR. AFAICT, code at HEAD is functionally correct with duplicated outputs, and I'll add the proper output-to-output alias support soon. The value of adding a special case of output-to-output alias isn't very useful at this moment.

jjsjann123:
!build

jjsjann123:
!build

PR #1418, opened 2023-11-30
Reduce size of input in FusionReshapeVectorize test

### LINKED ISSUE #1407: OOM error
There's a persisting OOM error recently. See for example https://github.com/NVIDIA/Fuser/actions/runs/7032815711, which was a CI test for PR #1319. The specific test failing is `GpuViewTest.FusionReshapeVectorize`. There was an attempt to fix the error (#1382), but apparently it was not enough.

The error message seems to suggest there's something occupying a large chunk of GPU memory aside from the PyTorch itself:

```
CUDA out of memory. Tried to allocate 256.00 MiB. GPU 0 has a total capacity of 15.78 GiB of which 20.75 MiB is free. Process 3304 has 15.76 GiB memory in use. Of the allocated memory 8.76 GiB is allocated by PyTorch, and 55.88 MiB is reserved by PyTorch but unallocated. If reserved but unallocated memory is large try setting PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True to avoid fragmentation. See documentation for Memory Management (https://pytorch.org/docs/stable/notes/cuda.html#environment-variables)
```

Here, PyTorch itself only used 8.75 GB, but the total usage was 15.76 GB, and we don't know where the remaining 7 GB went.

jacobhinkle:
This temporarily fixes #1407 but a longer-term solution would be to alter our cmake config so that each test file corresponds to a different test executable, in order to keep the CUDA context size small as we generate thousands of kernels that are not removed until the process terminates.

### DIFF
--- a/test/test_gpu_view.cpp
+++ b/test/test_gpu_view.cpp
@@ -1191,12 +1191,7 @@ TEST_F(GpuViewTest, FusionReshapeVectorize) {
 
   auto options = at::TensorOptions().dtype(at::kFloat).device(at::kCUDA, 0);
 
-  // This test allocates about 1GB of memory, so in order to avoid an OOM during
-  // this test, we manually clear the allocator after it's reached a certain
-  // threshold.
-  maybeClearAllocator();
-
-  at::Tensor input = at::randn({256, 1024, 1024}, options);
+  at::Tensor input = at::randn({256, 256, 256}, options);
 
   auto lparams = schedulePointwise(&fusion, {input});
 


jacobhinkle:
!build

test/test_gpu_view.cpp
```diff
@@ -1191,12 +1191,7 @@ TEST_F(GpuViewTest, FusionReshapeVectorize) {
 
   auto options = at::TensorOptions().dtype(at::kFloat).device(at::kCUDA, 0);
 
-  // This test allocates about 1GB of memory, so in order to avoid an OOM during
-  // this test, we manually clear the allocator after it's reached a certain
-  // threshold.
-  maybeClearAllocator();
```
jacobhinkle, line 1197: This workaround, introduced in #1382, is no longer necessary. If >4GiB are allocated in the PyTorch allocator at the beginning of the test, it will be cleared anyway. This PR reduces the size required by a factor of 16, so even if we are at 4GiB, this test should fit easily.

naoyam (approved):
LGTM. Thanks for the fix

PR #1419, opened 2023-11-30
Combine a mul-sum pair to a mma op

protonu:
Adding a stand alone pass to combine a mul-sum pair to a mma op.
There's an unit test case to try it out.

The next step will be to invoke this pass from the matmul scheduler.

### DIFF
--- a/csrc/scheduler/mma_utils.cpp
+++ b/csrc/scheduler/mma_utils.cpp
@@ -1229,6 +1229,163 @@ RolesMapOpt getTensorsRoles(Fusion* fusion) {
   return roles_map;
 }
 
+namespace {
+
+void addMMAOp(Fusion* fusion_, std::vector<MulSumAsMmaProps>& props) {
+  auto* init = IrBuilder::create<Val>(0.0);
+  for (auto prop : props) {
+    IrBuilder::create<MmaOp>(prop.out, prop.a, prop.b, init);
+  }
+}
+
+// Check the val (in) is the output of broadcast.
+// Then check the output of the broadcast is 3D (4D for bmm).
+bool hasValidBroadcastOp(TensorView* bcast_out) {
+  // First check the tensorsview is 3D (4D)
+  // and has one broadcast dim.
+  auto dims = bcast_out->domain()->nDims();
+  if (!((dims == 3 || dims == 4) &&
+        bcast_out->domain()->noBroadcasts().size() == dims - 1)) {
+    return false;
+  }
+
+  // Check if the definition is a broadcast op.
+  if (dynamic_cast<BroadcastOp*>(bcast_out->definition())) {
+    return true;
+  }
+
+  return false;
+}
+
+// This function checks if the mul-sum can be replace with a mma op. The checks
+// are:
+// 1. The inputs to the muls are broadcast ops.
+// 2. The broadcasts have 2D or 3D(bmm) inputs.
+// 3. The broadcasts only broadcast one dim and the dims are different for the 2
+// muls.
+// 4. There is a single reduction dim, and that dim that is not either of the
+// broadcast dims.
+bool broadcastsAreValid(
+    TensorView* left,
+    TensorView* right,
+    unsigned int reduction_axis) {
+  if (!(hasValidBroadcastOp(left) && hasValidBroadcastOp(right))) {
+    return false;
+  }
+
+  auto bcast_l = dynamic_cast<BroadcastOp*>(left->definition());
+  auto bcast_r = dynamic_cast<BroadcastOp*>(right->definition());
+
+  // Ensure that only one dim is getting broadcast.
+  auto bcastFlags_l = bcast_l->getBroadcastDimFlags();
+  auto bcastFlags_r = bcast_r->getBroadcastDimFlags();
+  auto count_l = std::count(bcastFlags_l.begin(), bcastFlags_l.end(), true);
+  auto count_r = std::count(bcastFlags_r.begin(), bcastFlags_r.end(), true);
+  if ((count_l != 1) || (count_l != count_r)) {
+    return false;
+  }
+
+  // Also ensure that it's not the same dim for the two muls. that's
+  // getting broadcast.
+  auto idx_l = std::find(bcastFlags_l.begin(), bcastFlags_l.end(), true) -
+      bcastFlags_l.begin();
+  auto idx_r = std::find(bcastFlags_r.begin(), bcastFlags_r.end(), true) -
+      bcastFlags_r.begin();
+  if (idx_l == idx_r) {
+    return false;
+  }
+
+  // Also ensure that the reduction dim is not either of the broadcast dim.
+  if (reduction_axis == idx_l || reduction_axis == idx_r) {
+    return false;
+  }
+
+  // Check different dimensions are the broadcast dims.
+  return true;
+}
+
+// If the tensorview is a output of a cast operation, then
+// return the input to the cast operation, else return the tensorview.
+TensorView* getTensorviewPriorToCast(TensorView* in) {
+  if (auto uCastOp = dynamic_cast<UnaryOp*>(in->definition());
+      uCastOp && uCastOp->getUnaryOpType() == UnaryOpType::Cast) {
+    return static_cast<TensorView*>(uCastOp->in());
+  }
+  return in;
+}
+
+// Check if the Mul-Sum pair represents a matmul. If so, add the properties
+// of the mma op which can be a tentatice substitue. This checks that the output
+// of sum has on reduction axis, and the inputs to mul are valid broadcasts.
+std::optional<MulSumAsMmaProps> getMulSumInsOutsBcasts(
+    BinaryOp* mop,
+    ReductionOp* redop) {
+  auto a = getTensorviewPriorToCast(static_cast<TensorView*>(mop->lhs()));
+  auto b = getTensorviewPriorToCast(static_cast<TensorView*>(mop->rhs()));
+
+  // Get the dimension of the reduction in the output. If not present, bail.
+  // Also ensure there is only only reduction axis.
+  auto red_axis = static_cast<TensorView*>(redop->out())->getReductionAxis();
+  auto num_reduction_dims =
+      static_cast<TensorView*>(redop->out())->domain()->nDims() -
+      static_cast<TensorView*>(redop->out())->domain()->noReductions().size();
+  if (!red_axis.has_value() || num_reduction_dims > 1) {
+    return std::nullopt;
+  }
+
+  if (broadcastsAreValid(a, b, *red_axis)) {
+    return MulSumAsMmaProps(
+        mop,
+        redop,
+        a,
+        b,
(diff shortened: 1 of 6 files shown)


csrc/scheduler/mma_utils.cpp
```diff
@@ -1229,6 +1229,86 @@ RolesMapOpt getTensorsRoles(Fusion* fusion) {
...
 }
 
+namespace {
+
+void add_mma_op_helper(TensorView* tv_a, TensorView* tv_b, TensorView* out) {
+  auto* init = IrBuilder::create<Val>(0.0);
+
+  NVF_CHECK(
+      tv_a->getDataType().value() == DataType::Half ||
+      tv_a->getDataType().value() == DataType::BFloat16);
```
jacobhinkle, line 1239: I think it would be better to let the `MmaOp` handle checks like this. Also we'll want single precision, and potentially others in the future.

csrc/scheduler/mma_utils.cpp
```diff
@@ -1229,6 +1229,86 @@ RolesMapOpt getTensorsRoles(Fusion* fusion) {
...
+};
+} // namespace
+
+
+void CombineMulSum::handle(ReductionOp* stmt) {
+  // Check if operation is a sum.
+  if (stmt->getReductionOpType() == BinaryOpType::Add) {
+    auto* inputOfSum = stmt->in();
+    if (inputOfSum != nullptr) {
+      auto* producer = inputOfSum->definition();
```
zasdfgbnm, line 1288: I think the term `producer` usually refers to `Val`s, instead of expressions?

csrc/scheduler/mma_utils.cpp
```diff
@@ -1229,6 +1229,86 @@ RolesMapOpt getTensorsRoles(Fusion* fusion) {
...
+      // Then check if the prodcer of the sum is a mul.
+      if (producer && dynamic_cast<BinaryOp*>(producer)) {
+        auto* bOp = dynamic_cast<BinaryOp*>(producer);
```
zasdfgbnm, line 1291: nit:
```suggestion
      if (auto bOp = dynamic_cast<BinaryOp*>(producer)) {
```

csrc/scheduler/mma_utils.cpp
```diff
@@ -1229,6 +1229,86 @@ RolesMapOpt getTensorsRoles(Fusion* fusion) {
...
+        if (bOp->getBinaryOpType() == BinaryOpType::Mul) {
+          sum_mul_pairs_.push_back({stmt, bOp});
+        }
+      }
+    }
+  }
+};
+
+bool CombineMulSum::run() {
+  if (sum_mul_pairs_.size() > 0) {
```
zasdfgbnm, line 1302: nit: use `sum_mul_pairs_.empty` (otherwise clang-tidy will complain I believe).

csrc/scheduler/mma_utils.cpp
```diff
@@ -1229,6 +1229,86 @@ RolesMapOpt getTensorsRoles(Fusion* fusion) {
...
+  // Check if operation is a sum.
+  if (stmt->getReductionOpType() == BinaryOpType::Add) {
+    auto* inputOfSum = stmt->in();
+    if (inputOfSum != nullptr) {
+      auto* producer = inputOfSum->definition();
+      // Then check if the prodcer of the sum is a mul.
+      if (producer && dynamic_cast<BinaryOp*>(producer)) {
+        auto* bOp = dynamic_cast<BinaryOp*>(producer);
+        // If it'a mul followed by a sum, put this in a list.
+        if (bOp->getBinaryOpType() == BinaryOpType::Mul) {
```
zasdfgbnm, line 1293: Should we actually check the domains of these tensors and only add it to `sum_mul_pairs_` if it is really a matmul? Not all mul-sum are matmuls, it could be just a pointwise mul followed by a sum.

protonu: @naoyam said something similar. I am working on that.

csrc/scheduler/mma_utils.cpp
protonu: Thanks, removed.

csrc/scheduler/mma_utils.cpp
protonu: Thanks, I added checks, please take a look.

naoyam: Shouldn't we also check the two inputs should have a broadcast in a different position, and the reduction domain should be the common non-broadcast domain?

csrc/scheduler/mma_utils.cpp
```diff
@@ -1229,6 +1229,114 @@ RolesMapOpt getTensorsRoles(Fusion* fusion) {
...
+  if (!((dims == 3 || dims == 4) &&
+        bcast_out->domain()->noBroadcasts().size() == dims - 1)) {
+    return false;
+  }
+
+  // Check if the inputs to bcastExpt is 2D (or 3D for bmm)
+  // and the output of Bcast is 3D or 4D
+  if (auto bcastExpr = dynamic_cast<BroadcastOp*>(bcast_out->definition())) {
+    auto input = static_cast<TensorView*>(bcastExpr->in());
+    if (input->nDims() == dims - 1) {
```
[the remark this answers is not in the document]
protonu: Thanks! I tried to take care of it. Please take a look.

csrc/scheduler/mma_utils.cpp
```diff
@@ -1229,6 +1229,158 @@ RolesMapOpt getTensorsRoles(Fusion* fusion) {
...
+    if (uCastOp && uCastOp->getUnaryOpType() == UnaryOpType::Cast) {
+      auto* aTV = static_cast<TensorView*>(uCastOp->in());
+      NVF_CHECK(
+          (aTV->getDataType() == DataType::BFloat16 ||
+           aTV->getDataType() == DataType::Half),
+          "The output of the cast must be half or BF16");
+      ret = aTV;
+    }
+  }
+  return ret;
```
jacobhinkle, line 1318: So this function is expected to return nullptr if the input is a `Double` tensor. Is that right?

csrc/scheduler/mma_utils.cpp
```diff
@@ -1229,6 +1229,158 @@ RolesMapOpt getTensorsRoles(Fusion* fusion) {
...
+
+std::optional<MulSumAsMmaProps> getMulSumInsOutsBcasts(
+    BinaryOp* mop,
+    ReductionOp* redop) {
+  auto a = getTensorviewPriorToCast(static_cast<TensorView*>(mop->lhs()));
+  auto b = getTensorviewPriorToCast(static_cast<TensorView*>(mop->rhs()));
+
+  // Get the dimension of the reduction in the output. If not present, bail.
+  auto red_axis =
+      static_cast<TensorView*>(redop->out())->domain()->getReductionAxis();
```
jacobhinkle, line 1329: Note this will give the _first_ reduction axis, but there may be others.

csrc/scheduler/mma_utils.cpp
```diff
@@ -1229,6 +1229,158 @@ RolesMapOpt getTensorsRoles(Fusion* fusion) {
...
+        }
+      }
+    }
+  }
+};
+
+std::vector<MulSumAsMmaProps> CombineMulSum::generateMulSumCanidates() {
+  if (!mul_sum_props_.empty()) {
+    return mul_sum_props_;
+  }
```
jacobhinkle, line 1371: Does this mean this function is caching its result? If so, please indicate that in the doxygen comment.

jacobhinkle: Also, if this is needed to prevent repeated traversals, keep in mind that in the vast majority of cases this vector will always be empty, so you might want to use a flag to explicitly indicate whether it has been processed. The above-mentioned comment should make it more clear how likely we are to be making redundant calls to this method.

csrc/scheduler/mma_utils.h
```diff
@@ -300,6 +300,60 @@ std::pair<bool, bool> generateSharedMemoryEpilogueHeuristics(
...
+//! in two phases. It can go through the graph and find the mul-sum
+//! pairs which can be replaced by a mma op. This phase returns a vector
+//! of properties of the mma op (MulSumAsMmaProps) which would replace the
+//! the mul-sum pair. It then exposes a function to replace with mma ops.
+class CombineMulSum : public IterVisitor {
+ public:
+  CombineMulSum(Fusion* fusion) : IterVisitor(), fusion_(fusion){};
+
+  //! Goes through the fusion to find mul-sum pairs.
+  std::vector<MulSumAsMmaProps> generateMulSumCanidates();
```
jacobhinkle, line 340: This feels like it could be private.

test/utils.cpp
```diff
@@ -338,6 +338,80 @@ TensorView* matmul(
...
+  }
+  std::vector<bool> bcast_dims(a->nDims() + 1, false);
+  bcast_dims.at(bcast_dims.size() - 2) = true;
+  tv0b = broadcast(tv0t, bcast_dims);
+  bcast_dims.at(bcast_dims.size() - 2) = false;
+  bcast_dims.at(bcast_dims.size() - 3) = true;
+  tv1b = broadcast(tv1t, bcast_dims);
+  tv2 = mul(tv0b, tv1b);
+  tv3 = sum(tv2, {-1});
+  // tv2 = fusedMultiplySum(tv0b, tv1b, {-1});
```
jacobhinkle, line 402: Since this is the extent of this change, could it be implemented as a flag in `matmul` instead of a separate function in order to reduce code duplication?

jacobhinkle (commented):
I think this is almost correct but I'm still unsure what happens if there is more than on reduction axis, and if the dtype is double, complex, bool, int, etc.

csrc/scheduler/mma_utils.h
protonu: The plan is to use this in the mma scheduler in a follow-up diff. If I don't use it, I'll hide it.

test/utils.cpp
protonu: More than happy to. I wasn't sure if we wanted to mix implementations.

naoyam:
> I think this is almost correct but I'm still unsure what happens if there is more than on reduction axis, and if the dtype is double, complex, bool, int, etc.

I think the translation itself should support those data types as well as long as it makes sense as matmul, and it's the responsibility of the scheduler to reject non-supported types.

csrc/scheduler/mma_utils.cpp
protonu: Removing type checks here.

naoyam (approved):
LGTM.

There may be some further changes needed to plum this into the scheduler, but that's beyond the scope of this initial PR.

@zasdfgbnm, @jacobhinkle , I think this looks reasonable. Please chime in if anything is still concerning for you.

protonu:
!build

jacobhinkle (approved):
LGTM. Thanks for addressing my comments.

PR #1423, opened 2023-11-30
Use `90a` target for hopper

zasdfgbnm:
Without this change, we can not use mma instructions for hopper.

# 90 vs 90a:
See https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#ptx-module-directives-target

> Target architectures with suffix “a”, such as `sm_90a`, include architecture-accelerated features that are supported on the specified architecture only, hence such targets do not follow the onion layer model. Therefore, PTX code generated for such targets cannot be run on later generation devices. Architecture-accelerated features can only be used with targets that support these features.


zasdfgbnm:
!build

PR #1427, opened 2023-11-30
L shift and R shift: `IrBuilder` and `evaluate`

zasdfgbnm:


zasdfgbnm:
!build

wujingyue (approved):
In general, I very much prefer shipping a change with tests **even if** it means the change has to be merged into a larger change. For example, `newArithmeticExpr` **could** have problems handling `BinaryOpType::Lshift`. A test would boost confidence and therefore make the change easier to review.

zasdfgbnm:
> In general, I very much prefer shipping a change with tests **even if** it means the change has to be merged into a larger change. For example, `newArithmeticExpr` **could** have problems handling `BinaryOpType::Lshift`. A test would boost confidence and therefore make the change easier to review.

I think it makes sense. I just manually added a test locally:

```C++
TEST_F(NVFuserTest, LRShift) {
  Fusion fusion;
  FusionGuard fg(&fusion);

  auto two = IrBuilder::create<Val>(2);
  auto one = IrBuilder::create<Val>(1);

  ExpressionEvaluator ee;
  EXPECT_EQ(ee.evaluate(IrBuilder::lShiftExpr(two, one)), 4);
  EXPECT_EQ(ee.evaluate(IrBuilder::rShiftExpr(two, one)), 1);
}
```

and check that this test is actually working. But I will not commit this test because I don't think we need to run this over and over again on every PR in the future to keep the CI time short.

naoyam:
> > In general, I very much prefer shipping a change with tests **even if** it means the change has to be merged into a larger change. For example, `newArithmeticExpr` **could** have problems handling `BinaryOpType::Lshift`. A test would boost confidence and therefore make the change easier to review.
> 
> I think it makes sense. I just manually added a test locally:
> 
> ```c++
> TEST_F(NVFuserTest, LRShift) {
>   Fusion fusion;
>   FusionGuard fg(&fusion);
> 
>   auto two = IrBuilder::create<Val>(2);
>   auto one = IrBuilder::create<Val>(1);
> 
>   ExpressionEvaluator ee;
>   EXPECT_EQ(ee.evaluate(IrBuilder::lShiftExpr(two, one)), 4);
>   EXPECT_EQ(ee.evaluate(IrBuilder::rShiftExpr(two, one)), 1);
> }
> ```
> 
> and check that this test is actually working. But I will not commit this test because I don't think we need to run this over and over again on every PR in the future to keep the CI time short.

Also, the shift ops are tested in `FusionBinaryOps`, although it doesn't use the IR builder nor the evaluator.

We should probably need to update `FusionBinaryOps` and `FusionUnaryOps` as we have more op types.

drzejan2 (commented):
Reviewed post-mortem - changes look good.
Thank you!

PR #1430, opened 2023-11-30
Move `TensorMapSwizzle` to `mma_type.h` and rename as `MmaInputSmemSwizzle`

zasdfgbnm:
Because it is used not only in TMA


zasdfgbnm:
!build

naoyam (approved):
LGTM

PR #1436, opened 2023-12-01
IdModel: Check self mappings

naoyam:
Mostly just copied from ComputeAtMap

### DIFF
--- a/csrc/id_model/id_model.cpp
+++ b/csrc/id_model/id_model.cpp
@@ -49,13 +49,34 @@ void mapThroughLoopSwizzles(ValGraph& graph) {
 
 } // namespace
 
-IdModel::IdModel(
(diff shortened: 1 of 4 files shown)


csrc/id_model/id_model.cpp
```diff
@@ -49,13 +49,34 @@ void mapThroughLoopSwizzles(ValGraph& graph) {
 
 } // namespace
 
+void IdModel::assertNoSelfMapping() {
+  if (hasSelfMapping()) {
```
zasdfgbnm, line 53: Do we need this `if`?

zasdfgbnm (approved):

csrc/id_model/id_model.cpp
naoyam: Hahaha, no. Will remove.

PR #1440, opened 2023-12-03
Serialize Fusion Segmentation

rdspring1:
### Summary:
NvFuser has heuristic schedulers to specify how to turn fusions into CUDA kernels. Fusions may need to be divided into smaller fusions to find a compatible scheduler. In very large fusions, this segmentation step takes a significant amount of time. This PR serializes the result of segmentation to avoid this costly step.

### Benchmark: 
A pathological case where 100 additions are performed in a single fusion. Overall running time decreases from 9 to 6 seconds because the segmentation cost is eliminated during deserialization.

CPU: AMD EPYC 7742
GPU: A100 40 GB

| NVRTC Marker Name  | Duration |
| ------------- | ------------- |
| Finding valid fusion segment solutions  | 3.436 s  |
| SegmentedFusion::deserialize  | 804.586 μs  |

### Nsight trace with full fusion segmentation.
pathological_full_compile

### Nsight Trace where segmentation is serialized in flatbuffers.
pathological_serde


rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
> How do you expect a Python user to use this feature?

Users can save the entire cache hierarchy manually via `FusionCache.serialize(filename)` or use the automatic serialization enabled via `nvfuser.enable_automatic_serialization()`. Users don't need to change anything to use this optimization.

> What's the entry point for segmentation serialization?

The `segmented_fusion` is stored inside the `FusionKernelRuntime`, which is a part of the cache hierarchy. At the moment, if a flatbuffer for the segmented fusion is passed to the `FusionKernelRuntime` constructor, we'll deserialize the `segmented_fusion`.

Reference: https://github.com/NVIDIA/Fuser/blob/main/csrc/kernel_cache.cpp#L982-L983

Also, all serialization is tested in `python_tests/test_python_frontend.py`. Majority of the tests use `exec_nvfuser`, which is decorated with `serde_check` function. The `serde_check` function exercises the serialization functionality. Automatic serialization is invoked in the `setUpModule` function.

References: 
- https://github.com/NVIDIA/Fuser/blob/main/python_tests/test_python_frontend.py#L63-L104
- https://github.com/NVIDIA/Fuser/blob/main/python_tests/test_python_frontend.py#L52-L60

wujingyue:
> Also, all serialization is tested in `python_tests/test_python_frontend.py`. Majority of the tests use `exec_nvfuser`, which is decorated with `serde_check` function. The `serde_check` function exercises the serialization functionality. Automatic serialization is invoked in the `setUpModule` function.
> 
> References:
> 
> * https://github.com/NVIDIA/Fuser/blob/main/python_tests/test_python_frontend.py#L63-L104
> * https://github.com/NVIDIA/Fuser/blob/main/python_tests/test_python_frontend.py#L52-L60

Thanks! That answered why we don't ship a test with this PR.

csrc/fusion_segmenter.cpp
```diff
@@ -26,8 +26,129 @@ namespace {
 
 using GroupSet = VectorOfUniqueEntries<SegmentedGroup*>;
 
+// This helper function converts pointer values into a one-to-one integer id.
+// It is used during serialization to define the container's state.
+template <typename ContainerT, typename T>
+std::vector<int64_t> convertPointerToInteger(
```
wujingyue, line 32: This helper doesn't leverage the fact that the key is a pointer, so I think we can make this more general with almost the same amount of code. 

```
template <typename K, typename V, typename ContainerK>
std::vector<V> convertContainer(const ContainerK& container, const std::unordered_map<K, V>& map) {
```

csrc/fusion_segmenter.cpp
```diff
@@ -316,16 +437,149 @@ std::unique_ptr<SegmentedFusion> SegmentedFusion::fromCompleteFusion(
...
+  const std::unordered_map<Expr*, int64_t>& exprs_map =
+      completeFusion()->deterministic_exprs_map();
+  const std::unordered_map<SegmentedGroup*, int64_t>& groups_map =
+      impl_.deterministic_groups_map();
+  const std::unordered_map<SegmentedEdge*, int64_t>& edges_map =
+      impl_.deterministic_edges_map();
+
+  std::vector<flatbuffers::Offset<serde::SegmentedEdge>> edges_fb;
+  edges_fb.reserve(edges_.size());
+  for (auto se : edges_) {
```
wujingyue, line 469:
```suggestion
  for (SegmentedEdge* se : edges_) {
```

csrc/fusion_segmenter.cpp
```diff
@@ -316,16 +437,149 @@ std::unique_ptr<SegmentedFusion> SegmentedFusion::fromCompleteFusion(
...
+  for (auto sg : groups_) {
+    groups_fb.push_back(
+        sg->serialize(builder, vals_map, exprs_map, groups_map, edges_map));
+  }
+
+  std::vector<int64_t> force_fp16_tv_fb;
+  force_fp16_tv_fb.reserve(force_fp16_tv_set_.size());
+  for (auto tv : force_fp16_tv_set_) {
+    force_fp16_tv_fb.push_back(vals_map.at(tv));
+  }
```
wujingyue, line 484: I think this can be simplified using the more general `convertContainer` I suggested early on.

csrc/fusion_segmenter.cpp
```diff
@@ -316,16 +437,149 @@ std::unique_ptr<SegmentedFusion> SegmentedFusion::fromCompleteFusion(
...
+      groups_map.at(edge->to),
+      vals_map.at(edge->val));
+}
+
+nvfuser::SegmentedEdge SegmentedFusion::deserialize(
+    const serde::SegmentedEdge* buffer,
+    const std::deque<Val*>& vals) {
+  FUSER_PERF_SCOPE("SegmentedEdge::deserialize");
+  NVF_ERROR(buffer != nullptr, "serde::SegmentedEdge is nullptr.");
+  return {
```
wujingyue, line 577: I'd `NVF_ERROR` that `groups_` is populated before calling this function, so the contract of order is clear.

csrc/kernel_cache.cpp
```diff
@@ -755,6 +755,7 @@ FusionKernelRuntime* FusionExecutorCache::getKernelRuntimeFor(
     kernel_runtimes.emplace_back(std::make_unique<FusionKernelRuntime>(
         std::move(conc_fusion),
         args,
+        nullptr /* serde_buffer */,
```
wujingyue, line 758:
```suggestion
        /*serde_buffer=*/nullptr,
```

`/*param_name=*/` is checked by clang-tidy, so is safer.

csrc/kernel_cache.h
```diff
@@ -92,6 +92,7 @@ class FusionKernelRuntime {
   explicit FusionKernelRuntime(
       std::unique_ptr<Fusion> fusion,
       const KernelArgumentHolder& inputs,
+      const serde::FusionKernelRuntime* serde_buffer = nullptr,
```
wujingyue, line 95: Can you add a code comment explaining what this means?

csrc/kernel_cache.cpp
```diff
@@ -997,6 +1021,11 @@ FusionKernelRuntime::FusionKernelRuntime(
   // Pre-compute the executor order so that the run time path
   //  would go directly to kernel launch.
   prepareRuntimeOrder();
+
+  // Restore State from Flatbuffers
+  if (serde_buffer != nullptr) {
+    deserialize(serde_buffer);
+  }
```
wujingyue, line 1028: I'm surprised this block is not at line 1008. Is there a good reason to put it here? Logically, it makes more sense to 

```
if (buffer is null) {
  do the expensive work
} else {
  restore from the buffer
}
```

csrc/fusion_segmenter.h
```diff
@@ -414,6 +449,10 @@ class SegmentedFusion {
     SegmentedGroup* makeFusionInputGroup();
     SegmentedEdge* makeEdge(SegmentedGroup* from, SegmentedGroup* to, Val* val);
     void cleanUnused();
+    const std::unordered_map<SegmentedGroup*, int64_t> deterministic_groups_map()
```
jjsjann123, line 452: nitpick: I'm usually against returning const value...

Wondering if @wujingyue has some opinion here. :stuck_out_tongue_closed_eyes:

rdspring1: I didn't put much thought into making the return value const.

I must have copied it from here.
https://github.com/NVIDIA/Fuser/blob/main/csrc/ir/container.h#L55-L64

csrc/fusion_segmenter.cpp
```diff
@@ -26,8 +26,129 @@ namespace {
...
+
+  // -1 corresponds with a nullptr value
+  int64_t merge_with_segmented_group = -1;
+  if (merge_with_ != nullptr) {
+    merge_with_segmented_group = groups_map.at(merge_with_);
+  }
+
+  // -1 corresponds with a nullptr value
+  int64_t merge_through_segmented_edge = -1;
+  if (merge_with_ != nullptr) {
```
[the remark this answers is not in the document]
rdspring1: I'm blind.

csrc/fusion_segmenter.cpp
```diff
@@ -26,8 +26,129 @@ namespace {
...
+    merge_with_ = groups.at(buffer->merge_with_segmented_group());
+  }
+
+  // -1 corresponds with a nullptr value
+  if (buffer->merge_through_segmented_edge() != -1) {
+    merge_through_ = edges.at(buffer->merge_through_segmented_edge());
+  }
+
+  merged_ = buffer->merged();
+  is_fusion_input_ = buffer->merged();
```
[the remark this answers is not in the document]
rdspring1: I'm blind again.

csrc/fusion_segmenter.cpp
```diff
@@ -370,6 +624,38 @@ void SegmentedFusion::Impl::cleanUnused() {
       edges_.end());
 }
 
+//! Return mapping from SegmentedGroup to integer id
+const std::unordered_map<SegmentedGroup*, int64_t> SegmentedFusion::Impl::
+    deterministic_groups_map() const {
```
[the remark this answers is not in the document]
rdspring1: I removed the `deterministic_` modifier. It isn't need here.

rdspring1:
!build

jjsjann123 (approved):
Looks like we still have a few minor code quality improvement nitpicks. But overall I don't see functional issues any more. Stamping since my concerns are resolved now.

csrc/kernel_cache.cpp
wujingyue: Gotcha! This is `FusionKernelRuntime::deserialize` not `SegmentedFusion::deserialize`. 

A follow-up question: as you said, `FusionKernelRuntime::deserialize` creates `FusionExecutor`s from the flatbuffer. So aren't we supposed to have an if-else branch that creates `FusionExecutor`s in a different way? Conceptually, it makes more sense to 

```
if (serialization is on) {
  create something from flatbuffer;
} else {
  create something from void;
}
```

But I don't see the if-else branch in your code.

rdspring1:
!build

csrc/kernel_cache.cpp
rdspring1: The `else` branch occurs at https://github.com/NVIDIA/Fuser/blob/main/csrc/kernel_cache.cpp#L476-L478. There isn't an obvious reason why it couldn't occur in the constructor.

In any case, I moved the if-branch back to its original position at
https://github.com/NVIDIA/Fuser/blob/main/csrc/kernel_cache.cpp#L924-L926.

rdspring1:
!build

csrc/fusion_segmenter.cpp
```diff
@@ -316,16 +426,152 @@ std::unique_ptr<SegmentedFusion> SegmentedFusion::fromCompleteFusion(
...
+      initial_exprs_size_,
+      &edges_fb,
+      &groups_fb,
+      &force_fp16_tv_fb,
+      toUnderlying(std::get<PrimDataType>(force_half_precision_type_.type)));
+}
+
+void SegmentedFusion::deserialize(const serde::SegmentedFusion* buffer) {
+  FUSER_PERF_SCOPE("SegmentedFusion::deserialize");
+  NVF_ERROR(buffer != nullptr, "serde::SegmentedFusion is nullptr.");
```
wujingyue, line 488: You can avoid this check by passing in a const ref.

csrc/kernel_cache.cpp
```diff
@@ -979,8 +982,34 @@ FusionKernelRuntime::FusionKernelRuntime(
...
+        [](const serde::SegmentedGroup* sg) {
+          auto heuristic = static_cast<ScheduleHeuristic>(sg->heuristic());
+          return heuristic == ScheduleHeuristic::InnerPersistent ||
+              heuristic == ScheduleHeuristic::OuterPersistent ||
+              heuristic == ScheduleHeuristic::InnerOuterPersistent;
+        });
+
+    bool has_welford_ops = ir_utils::hasOpsOfType<WelfordOp>(fusion.get());
+    if (has_welford_ops && has_persistent_heuristic) {
+      SegmentCandidateFinder::translateWelfordInFusion(fusion.get(), args);
```
wujingyue, line 1008: We modify Fusion here because we don't serialize Fusion? Is there value (in the future) to serialize/deserialize fusion IR so we can cache fusion modifications as well?

wujingyue (approved):

csrc/fusion_segmenter.cpp
rdspring1: All FlatBuffer fields are optional and are returned as pointers, so I don't think we can avoid a check.

csrc/kernel_cache.cpp
rdspring1: Serializing the fusion IR is planned but this welford transformation is the top priority.

Issue #1406, opened 2023-11-29
Failed chain projections when project persistent buffer to its persistent producers


liqiangxl:
In normalization scheduler, persistent tvs can be projected to its immediate persistent producers if `project_persistent_buffers = False`.
However, current implentation fails if a chain projection is required. Consider the following case where T1, T2, T3 are persistent buffers, ideally, T2 and T3 should be projected to T1.
```
      T1 = f(T0); Tx = add(T1, broadcast(sum(T1)));
      T2 = f(T1); Ty = add(T2, broadcast(sum(T2)));
      T3 = f(T2); Tz = add(T3, broadcast(sum(T3)));
```
If T2 is projected to T1 before T3 is projected, we get:
```
      T1 = f(T0); Tx = add(T1, broadcast(sum(T1)));
      T2 = f(T1); Ty = add(f(T1), broadcast(sum(T2)));
      T3 = f(f(T1)); Tz = add(T3, broadcast(sum(T3)));
```
Then, T3's producer is changed to f(T1) which is not a persistent tensor. So T3 can't be projected back to T1.

To fix this issue, we should: set a projection order based on dependencies. In this case, the visiting order should be [T3, T2, T1]. After project T3 to its producers, we have:
```
 Tz = add(f(T2),broadcast(sum(T3)));
```
Then,  continue project T2 to its producers, we have:
```
      Tz = add(f(f(T1)),broadcast(sum(T3)));
      Ty = add(f(T1), broadcast(sum(T2)));
```

The only left persistent buffer is T1.

naoyam:
Just to clarify, the T2 and T3 inputs to the sum ops should also be replaced, right?

liqiangxl:
> Just to clarify, the T2 and T3 inputs to the sum ops should also be replaced, right?

why? For example:
After project T3 to its producers, in the fusion we have:
`T3 = f(T2); Tz = add(f(T2),broadcast(sum(T3)));`
Here `sum(T3)` is consuming T3 generated by `T3 = f(T2)`.

Then, continue project T2 to its producers, we have:
```
T2 = f(T1);  Ty = add(f(T1), broadcast(sum(T2)));
T3 = f(f(T1));  Tz = add(f(f(T1)),broadcast(sum(T3)));
```
similarly, `sum(T2)` is consuming T2 generated by `T2 = f(T1)`.

naoyam:
[...]

PR #1441, opened 2023-12-04
set projection order when project persistent buffer to its persistent producers

liqiangxl:
Issue: #1406
Fix: set a projection order based on dependencies when project persistent buffer to its persistent producers.
Example:
```
      // visit consumer before producer. e.g.
      // T1 = f(T0); Tx = add(T1, broadcast(sum(T1)));
      // T2 = f(T1); Ty = add(T2, broadcast(sum(T2)));
      // T3 = f(T2); Tz = add(T3, broadcast(sum(T3)));
      // T1, T2, T3 are persistent buffers.
      // The visiting order should be [T3, T2, T1].
      // After project T3 to its producers, we have:
      // Tz = add(f(T2),broadcast(sum(T3)));
      // After project T2 to its producers, we have:
      // Tz = add(f(f(T1)),broadcast(sum(T3)));
      // Ty = add(f(T1), broadcast(sum(T2)));
      // At last, the only persistent buffer is T1.
      // For a solid case, see NVFuserTest.ChainProjectionToPersistentProducer.
```


liqiangxl:
!build --diff --diff-bench

liqiangxl:
The CI failures are not related to the changes in this PR.

csrc/scheduler/reduction_utils.cpp
```diff
@@ -736,18 +736,38 @@ class PersistentBufferProjector {
...
+      for (auto buffer_i : visiting_order) {
         auto buffer = persistent_buffers[buffer_i];
         const auto& producers = ir_utils::producerTvsOf(buffer);
         if (scheduler_utils::canProjectToPersistentProducer(
                 buffer, producers, persistent_buffer_set)) {
           projectToInputOrImmediatePersistentProducer(
               (int)buffer_i,
               std::vector<Val*>(producers.begin(), producers.end()));
-          // "buffer" is no longer a persistent buffer
-          persistent_buffer_set.erase(buffer);
```
naoyam, line 750: Why is this removed?

csrc/scheduler/reduction_utils.cpp
```diff
@@ -736,18 +736,38 @@ class PersistentBufferProjector {
             (int)buffer_i, fusion_->inputs());
       }
     } else {
+      // visit consumer before producer. e.g.
```
naoyam, line 739: This `project` function is getting bigger. Can you do some cleanup? Just separating it to `projectToInputs()` and `projectToProducers` would be enough.

csrc/scheduler/reduction_utils.cpp
```diff
@@ -736,18 +736,38 @@ class PersistentBufferProjector {
...
+      // After project T3 to its producers, we have:
+      // Tz = add(f(T2),broadcast(sum(T3)));
+      // After project T2 to its producers, we have:
+      // Tz = add(f(f(T1)),broadcast(sum(T3)));
+      // Ty = add(f(T1), broadcast(sum(T2)));
+      // At last, the only persistent buffer is T1.
+      // For a solid case, see NVFuserTest.ChainProjectionToPersistentProducer.
+      std::vector<int> visiting_order(persistent_buffers.size());
+      std::iota(visiting_order.begin(), visiting_order.end(), 0);
+      std::sort(
```
naoyam, line 754: Couldn't this be a source of non-determinism? Shouldn't we use `std::stable_sort`?

csrc/scheduler/reduction_utils.cpp
liqiangxl: Once a "buffer" has been projected, it is no longer a persistent buffer. Nonetheless, its removal from the buffer set is unnecessary due to the established visiting order. This order guarantees that buffers accessed prior to the current one cannot be the producer of the current buffer. Therefore, additional cleanup is not required.

csrc/scheduler/reduction_utils.cpp
liqiangxl: changed.

csrc/scheduler/reduction_utils.cpp
liqiangxl: changed to `std::stable_sort`

naoyam (approved):
LGTM

PR #1442, opened 2023-12-04
add api to return kernel occupancy and register spills

liqiangxl:
when tuning heuristics, needs a function to get kernel occupancy and register spills.
usage:
```
FusionExecutor fe;
...
auto occupancy =  fe.getKernelOccupancy();
auto register_spills = fe.getKernelRegisterSpills();
```


csrc/executor.h
```diff
@@ -205,6 +205,19 @@ class FusionExecutor : public NonCopyable {
     return measure_kernel_time_ ? kernel_time_ms_ : 0;
   }
 
+  //! get occupancy of the last kernel execution
+  float getKernelOccupancy() const {
+    return kernel_occupancy_;
```
naoyam, line 210: nit: `NVF_ERROR(kernel_occupancy > 0, "Occupancy unknown");`

naoyam (approved):
LGTM

csrc/executor.h
liqiangxl: added

PR #1448, opened 2023-12-05
fix detection of nccl backend in reduce_oop_

samnordmann:
Since the CommunicatorBackend argument is optional, it can be null, in which case the default backend is returned. Then we need to dynamic cast the backend to check if it is nccl


cowanmeg (approved):

samnordmann:
@naoyam is it ok to merge ?

naoyam (approved):
LGTM

PR #1451, opened 2023-12-05
add reduction comms

samnordmann:
Add a lowering path for reduction collective.
More precisely, before this patch: in a multi-device pipeline, the expr in-between two stages was required to be a "set". With this patch, it can also be a reduction. This expression will be lowered to MPI collectives of the type `Reduce`, `Allreduce` or `ReduceScatter`.
We augment our test units to cover those scenarios.

Some restrictions:
- Only proper reduce-scatter without futher resharding are supported for now (i.e. I/O of a reduce scatter must be sharded on the same device mesh. If further resharding is desired, the user must manually insert a set before/after the reduction to take care of the device mesh change)
- Only reducing one axis at a time is supported
- If a tensor is reduced in-between stage, it must be sharded on the producer side (i.e., the reduction must occurs accross devices ; otherwise the operation boils down to a local reduction + a non-reduction resharding, and can be obtained by inserting a "set" after the reduction op). Similarily, the sharding must be non-trivial, i.e., on an axis of size > 1, otherwise it is automatically replaced by a "squeeze" and thus will not be supported.

Those restrictions will be relaxed later and handled automatically, but it is more convenient and generic to handle those corner cases by modifying the Fusion itself by adding "Set" exprs before or after the reduction. This will be the object of a future PR.


samnordmann:
!build

csrc/multidevice/lower_communication.cpp
```diff
@@ -36,6 +83,12 @@ inline at::Tensor createDummyTensor(at::Tensor reference) {
   return at::empty_like(reference, reference.options());
 }
 
+inline at::Tensor createDummyTensor(
+    at::Tensor reference,
+    BinaryOpType op_type) {
+  return createDummyTensor(reference).fill_(getInitialValue<float>(op_type));
```
crcrpar, line 89: out of curiosity, why is `getInitialValue` called with `float`?

samnordmann: There is no particular reason, if I'm correct the type doesn't matter here since the value inside `fill_` will be casted to the tensor type

Do you have a better suggestion ?

cowanmeg (approved):

csrc/multidevice/lower_communication.cpp
crcrpar: no, I was just curious

csrc/multidevice/lower_communication.h
```diff
@@ -14,6 +14,8 @@
 
 namespace nvfuser {
 
+bool isLowerableToCommunication(Expr* expr);
```
naoyam, line 17: Please add a comment

csrc/multidevice/lower_communication.cpp
```diff
@@ -291,40 +455,97 @@ std::vector<std::shared_ptr<Communication>> lowerCommunication(
...
 
+bool isLowerableToCommunication(Expr* expr) {
+  if (expr->isA<ReductionOp>()) {
+    auto out = expr->as<ReductionOp>()->out();
+    NVF_ERROR(out->isA<TensorView>(), "output is not a TensorView");
+    auto out_tv = out->as<TensorView>();
+    NVF_ERROR(
+        out_tv->domain()->nDims() ==
+            TensorDomain::noReductions(out_tv->getMaybeRFactorDomain()).size() +
+                1,
```
naoyam, line 541:
```suggestion
        std::count_if(out_tv->getMaybeRFactorDomain().begin(),
                  out_tv->getMaybeRFactorDomain().end(),
                   [] (IterDomain* id) {
                        return id->isReduction();
                    }) == 1
```

csrc/multidevice/lower_communication.cpp
```diff
@@ -291,40 +455,97 @@ std::vector<std::shared_ptr<Communication>> lowerCommunication(
...
   }
   return comms;
 }
 
+bool isLowerableToCommunication(Expr* expr) {
+  if (expr->isA<ReductionOp>()) {
+    auto out = expr->as<ReductionOp>()->out();
+    NVF_ERROR(out->isA<TensorView>(), "output is not a TensorView");
+    auto out_tv = out->as<TensorView>();
+    NVF_ERROR(
```
naoyam, line 538: Shouldn't this just return `false`?

csrc/multidevice/lower_communication.cpp
```diff
@@ -291,40 +455,97 @@ std::vector<std::shared_ptr<Communication>> lowerCommunication(
...
+        out_tv->domain()->nDims() ==
+            TensorDomain::noReductions(out_tv->getMaybeRFactorDomain()).size() +
+                1,
+        "only reducing one-axis at a time is supported");
+    return true;
+  }
+  return expr->isA<LoadStoreOp>() &&
+      (expr->as<LoadStoreOp>()->opType() == LoadStoreOpType::Set);
```
naoyam, line 546:
```suggestion
  } else {
    return expr->isA<LoadStoreOp>() &&
      (expr->as<LoadStoreOp>()->opType() == LoadStoreOpType::Set);
  }
```

csrc/multidevice/device_mesh.h
```diff
@@ -40,6 +40,10 @@ class DeviceMesh final {
     return std::find(vector_.begin(), vector_.end(), device) != vector_.end();
   }
 
+  bool operator==(const DeviceMesh& other) const {
```
naoyam, line 43: Is it necessary to explicitly define this?

csrc/multidevice/lower_communication.cpp
```diff
@@ -11,11 +11,57 @@
...
+inline T getInitialValue(BinaryOpType op) {
+  switch (op) {
+    case BinaryOpType::Add:
+    case BinaryOpType::BitwiseOr:
+    case BinaryOpType::BitwiseXor:
+      return 0;
+    case BinaryOpType::Mul:
+      return 1;
+    case BinaryOpType::Min:
+      return std::numeric_limits<T>::min();
```
naoyam, line 31: Shouldn't this be `std::numeric_limits<T>::max()`?

csrc/multidevice/lower_communication.cpp
```diff
@@ -11,11 +11,57 @@
...
+    case BinaryOpType::Max:
+    case BinaryOpType::BitwiseAnd:
+      return std::numeric_limits<T>::max();
```
naoyam, line 34: Here too. Also why BitwiseAnd use `max`? What would type `T` be?

csrc/multidevice/lower_communication.cpp
samnordmann: right

csrc/multidevice/device_mesh.h
samnordmann: not necessary, it was for convenience, but let me remove it

csrc/multidevice/lower_communication.cpp
samnordmann: Indeed, `BinaryOpType::Max:` needs `std::numeric_limits<T>::min()`

For `BitwiseAnd`, what do you suggest? The neutral element should be 111....1 bitwise, so this is max, right? Let me know if I am wrong

About type T, I think it can be anything for all the ops except BitwiseAnd, for which there will actually be a problem...

samnordmann: The best would be to choose T according to the tensor's type, but I couldn't find how to translate torch's type into c++ native type

naoyam: Since `T` for bitwise operations should be an integer type, why not just something like `~(T)0`?

naoyam: > I couldn't find how to translate torch's type into c++ native type

Does `AtenTypeToNativeType` work? See type.h.

naoyam (approved):
LGTM. I had some comments but nothing critical.

csrc/multidevice/lower_communication.cpp
samnordmann: > Since `T` for bitwise operations should be an integer type, why not just something like `~(T)0`?

Right, I didn't think about that, thx 


> Does AtenTypeToNativeType work? See type.h.

Something like:
`getInitialValue<c10::impl::ScalarTypeToCPPType<reference.scalar_type()>::t>(op_type)`
should work, or alternatively replacing `ScalarTypeToCPPType` with `AtenTypeToNativeType` as you suggest.
But we still have the problem of "non-`constexpr`ness" of the template argument.
We should write a dispatch function or something like that. I don't know what's the best way to handle that. As far as I understand there is no c++ torch call to get the limits from the ScalarType, even though there is a python pytorch way to do it, through finfo.

For now, I just assert that the type is kFloat32 and write a TODO. Lmk if it is ok for you

PR #1405, opened 2023-11-29
Grid reduction with serialized blocks

jacobhinkle:
This PR introduces kernel IR nodes for the syncs that need to occur before and after a loop containing a serial grid reduction. That grid reduction can be inlined with other computation in a loop nest, and the sync nodes will be placed around the outer loop in the generated kernel. The `kir::GridReduction` node itself is modified to have an attribute available via `bool kir::GridReduction::isSerial() const` indicating whether this is a serial grid reduction. This PR tests that codegen is correct.

Default CUDA kernel for the included test:
```
__global__ void nvfuser_none_f0_c0_r0_g0(Tensor<float, 2, 2> T0, Tensor<float, 1, 1> T1, Tensor<float, 1, 1> T4, Tensor<int64_t, 1, 1> T5) {
  alignas(16) extern __shared__ char array[];
  void* shared_mem = array;
  NVFUSER_DEFINE_MAGIC_ZERO;
  nvfuser_index_t i0;
  i0 = 32LL * ((nvfuser_index_t)threadIdx.y);
  nvfuser_index_t i1;
  i1 = 32768LL * ((nvfuser_index_t)blockIdx.y);
  nvfuser_index_t i2;
  i2 = 262144LL * ((nvfuser_index_t)blockIdx.x);
  nvfuser_index_t i3;
  i3 = ((((2097152LL * ((nvfuser_index_t)blockIdx.z)) + ((nvfuser_index_t)threadIdx.x)) + i0) + i1) + i2;
  nvfuser_index_t i4;
  i4 = ((((nvfuser_index_t)threadIdx.x) + i0) + i1) + i2;
  nvfuser_index_t i5;
  i5 = (((-2097152LL + ((nvfuser_index_t)threadIdx.x)) + i0) + i1) + i2;
  bool b6;
  b6 = ((nvfuser_index_t)blockIdx.z) == (((nvfuser_index_t)gridDim.z) + -1LL);
  // Allocate global tensor T4
  // Allocate global tensor T5
  float T2[128LL];
  #pragma unroll
  for(nvfuser_index_t i7 = 0; i7 < 32LL; ++i7) {
    #pragma unroll
    for(nvfuser_index_t i8 = 0; i8 < 4LL; ++i8) {
      T2[(i7 + (32LL * i8))] = 0.000000000e+00f;
    }
  }
  NVFUSER_UPDATE_MAGIC_ZERO;
  #pragma unroll
[...]

PR #1456, opened 2023-12-06
Scheduling/lowering for serial grid reduction

jacobhinkle:
This change enables `ReductionOp`s to be lowered as serial reductions (see #1405) if requested during scheduling.

1. At scheduling, a `ReductionOp` is modified by calling its `requestSerialGridReduction()` method. The output tensor can be scheduled before or after this method call, and should result in the op having all its reduction axes parallelized as grid dimensions.
2. Early in lowering, we find `ReductionOp`s having `serialGridReductionRequested() == true`, and we place syncs around their outer loop. At this point, we also analyze that outer loop to determine if there are any conflicting expressions, such as conflicting grid reductions.
3. Later in lowering, during the indexing pass, we convert the `ReductionOp` to a `GridReduction` that has its serial buffer set. The serial buffer is a temporary `TensorIndex` indexed like a global memory version of the reduction output tensor.

The generated kernel looks like this:
```c++
  // Allocate global tensor T4
  grid_sync::blockSerializeWait<false, false, true>(&T4[index_utils::maskedOffset<true, true, false>(blockIdx, gridDim)]);
  #pragma unroll
  for(nvfuser_index_t i13 = 0; i13 < 4LL; ++i13) {
    nvfuser_index_t i14;
    i14 = 8LL * i13;
    nvfuser_index_t i15;
    i15 = 2048LL * i13;
    nvfuser_index_t i16;
    i16 = i4 + i15;
    nvfuser_index_t i17;
    i17 = -i15;
    #pragma unroll
    for(nvfuser_index_t i18 = 0; i18 < 8LL; ++i18) {
      nvfuser_index_t i19;
      i19 = 256LL * (i18 + nvfuser_zero);
      nvfuser_index_t i20;
      i20 = i16 + i19;
      float T3[1LL];
      T3[0LL] = 0.000000000e+00f;
      // Allocate global tensor T5
      reduction::serialReductionStep(
        T3[0LL],
        T2[(i14 + i18)],
        0.000000000e+00f,
        T5[i20],
        true,
        true);
      if ((b6 && (i5 < (i17 - i19)))) {
        T1[i20]
           = T3[0LL];
      }
    }
  }
  NVFUSER_UPDATE_MAGIC_ZERO;
  grid_sync::blockSerializeRelease<false, false, true>(&T4[index_utils::maskedOffset<true, true, false>(blockIdx, gridDim)]);
```
Notice that the index `i20` now matches between the output `T1` and the intermediate `T5`. In another PR, I will attempt to extend our buffer reuse machinery to recognize this as a chance to use `T1` in place of `T5` (i.e. inner aliasing, in-place reduction).

Also notice that I have not yet hoisted the sync flags index, or the `first_block` and `last_block` predicates.


csrc/device_lower/lower2device.cpp
```diff
@@ -266,13 +266,17 @@ GpuLower::GpuLower(Fusion* fusion, const CompileParams& cparams)
...
            {"insertRawThreadSynchronization", insertRawThreadSynchronization},
+           // NOTE: the smem/register reuse pass occurs here
            {"reuseMemoryAllocations", reuseMemoryAllocations},
            {"insertWarThreadSynchronization", insertWarThreadSynchronization},
            {"DoubleBufferPass", DoubleBufferPass::run},
            {"rotateLoops", rotateLoops},
            {"UnrollPass", UnrollPass::runPass},
            {"processMisalignedVectorization", processMisalignedVectorization},
+           // NOTE: serial GridReduction introduced here. New syncs can be
+           // introduced here which could impact smem reuse
```
jacobhinkle, line 277: We should probably factor out the logic for detecting serial grid reductions and insert those syncs much earlier than this point, so that they are available to `reuseMemoryAllocations`. This would let us re-use the wait sync instead of inserting a new one when we re-use prologue memory for epilogue.

csrc/kernel_ir.cpp
```diff
@@ -149,7 +149,7 @@ Allocate::Allocate(
     NVF_ERROR(buffer->isA<TensorView>());
     NVF_ERROR(buffer->as<TensorView>()->getMemoryType() == memory_type);
     const auto domain = buffer->as<TensorView>()->domain();
-    for (auto axis : domain->noReductions()) {
+    for (auto axis : TensorDomain::noReductions(domain->maybeAllocation())) {
```
jacobhinkle, line 152: This was done to work around the error described in https://github.com/NVIDIA/Fuser/pull/1456/commits/910ff09862387b3783af1837c82d5d600dc8b893

csrc/device_lower/pass/insert_syncs.cpp
```diff
@@ -369,32 +369,6 @@ class ValidatePlacementAfterWrites : private kir::IrVisitor {
   const std::unordered_set<Expr*>& writes_;
 };
 
-namespace {
-
-Val* getGridSyncBufferSize(const ParallelTypeBitmap& ptb) {
```
jacobhinkle, line 374: Moved to `lower_utils::`

csrc/codegen.cpp
```diff
@@ -1584,17 +1584,17 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
     const auto data_type = grop->out()->dtype();
     const auto op_type = grop->getReductionOpType();
 
+    if (grop->isSerial()) {
+      generateSerialGridReduction(grop);
+      return;
+    }
+
```
jacobhinkle, line 1591: This block is moved earlier to avoid dereferencing null pointers for reduction and sync buffer, which are not set for serial reductions.

csrc/device_lower/pass/index.cpp
```diff
@@ -611,10 +612,117 @@ void IndexLowering::handleBlockReduction(
...
+void IndexLowering::handleSerialGridReduction(
+    const ReductionOp* rop,
+    Val* out,
+    Val* in) {
+  const auto out_tv = out->as<kir::TensorIndex>()->view();
+  const auto out_domain = out_tv->domain();
+
+  // If we do a grid reduction we can't have a reduction axis that is not bound
+  // to a grid or block dim.
+  NVF_ERROR(
```
jacobhinkle, line 624: This is carried over from the original `GridReduction`. We might be able to relax this constraint for serial reductions at some point.

jacobhinkle:
!build --diff

jacobhinkle:
the `Pipeline_CUDA` test is failing, which seems to be real. I'm looking into it now.

jacobhinkle:
> the iterdomain extents are bumped starting in the second reduction

This is because now we create an additional `bool` `Val*` when we create a `ReductionOp`, corresponding to the `serialGridReductionRequested()` attribute. I was surprised at first because the attribute is actually not a `Val` attribute, but `Expr::addDataAttribute(PolymorphicValue attr)` actually creates a `Val` with the given `value` and adds that using `addAttribute(Statement* attr)`. The proper fix then seems to be to just update the val names in the failing test.

test/test_pipeline.cpp
```diff
@@ -21,6 +21,26 @@ namespace nvfuser {
...
+  bool first = true;
+  for (auto line : lines) {
+    if (!first) {
+      output << std::endl;
+    }
+    first = false;
+    output << line;
+  }
+  return output.str();
+}
```
jacobhinkle, line 42: Just cleaning up the TODO in this test while I'm here. This makes the error message provide a useful diff.

test/test_pipeline.cpp
```diff
@@ -81,7 +101,7 @@ TEST_F(NVFuserTest, Pipeline_CUDA) {
       " PipelineVal representing Val T0_g[ iS0{i0}, iS1{i2} ] on stage " +
       std::to_string(stage0.unique_id) +
       "\n"
-      " PipelineVal representing Val T4_g[ iS6{i13}, iS7{i14}, iS8{i15} ] on stage " +
+      " PipelineVal representing Val T4_g[ iS6{i15}, iS7{i16}, iS8{i17} ] on stage " +
```
jacobhinkle, line 104: These changes were necessitated by the addition of an attribute in `ReductionOp`. A `Val*` is created when each attribute is added, incrementing the scalar name counter.

jacobhinkle:
!build

csrc/kernel_ir.cpp
naoyam: Can you give more details? Does `T5` in the example in the commit have an allocation domain?

csrc/device_lower/pass/index.cpp
```diff
@@ -611,10 +612,111 @@ void IndexLowering::handleBlockReduction(
...
+          }),
+      "Found a reduction stage that has both a non-parallelized ",
+      "reduction and a grid reduction. This is not supported, ",
+      "please use rfactor to do the serialized reduction first, ",
+      "then the grid reduction. ",
+      rop->toString());
+
+  NVF_ERROR(!rop->isAllreduce(), "Serial grid allReduce is not implemented");
+
+  // Allocate global work buffer TensorIndex.
```
naoyam, line 640: Why does the work buffer need to be scheduled? Can't it be simply just a buffer with the leaf domains of the reduction output tensor?

csrc/device_lower/pass/grid_serialization.cpp
```diff
@@ -0,0 +1,177 @@
...
+      } else {
+        cur_expr_sync_pattern_ = sync_pattern;
+      }
+    }
+  }
+
+  void dispatch(Expr* expr) override {
+    if (auto loop = dynamic_cast<kir::ForLoop*>(expr);
+        cur_top_level_expr_ || (loop && loop->isTrivial())) {
+      // Never sync around trivial loops. Also avoid redefining
```
naoyam, line 100: Placing sync around a trivial loop results in an error?

jacobhinkle: `cur_top_level_expr_` is the expression we'll place syncs around in `insertSyncs`. This check just bypasses setting `cur_top_level_expr_`, so if there are trivial loops, we will still dispatch into them, but they won't count as top-level expr.

csrc/device_lower/pass/index.cpp
jacobhinkle: I think you're right. It doesn't need to be scheduled explicitly, we just need to use the leaf domains of the output to determine how to index it (indexing as if the output were in global mem).

jacobhinkle: Just pushed a change to simplify this by reusing `out_tv->domain()` instead of replaying to recreate it.

naoyam: This will share the same IterDomains with multiple TensorDomains, which I think we should avoid since we implicitly assume each tensor has unique IterDomains.

I think what we should do here would be something like:

```
std::vector<IterDomain*> work_buffer_ids(out_tv->nDims());
for (IterDomain* out_id : out_tv->leaf()) {
   work_buffer_ids.push_back(IterDomainBuilder(out_id).build());
}
auto work_buffer_domain = IrBuilder::create<TensorDomain>(work_buffer_ids);
```

jacobhinkle: Ah OK. I see what you mean now. I was trying to preserve the transforms originally so that we could potentially re-use a a fusion output buffer for the work buffer in the future. However, that is probably low priority since it only applies when the output has the same type as the mma accumulator and we commonly have single precision accumulator and half-precision output tensors. For simplicity's sake I'll do as you suggest and just use the leaf domain of the mma output as the work buffer's root/leaf. I'll leave a comment to remind myself to revisit this if we implement global buffer re-use in the future.

naoyam: Sorry I wasn't clear. Yes, this should be enough for now.

naoyam (approved):
LGTM

PR #1469, opened 2023-12-07
Allocate intermediate global buffers the same way we allocate outputs

jacobhinkle:
This fixes a subtle bug that was exposed by #973. In the test `ExpandedBroadcastGlobalIntermediateTest.TheTest_CUDA`, which currently passes, we are binding an unexpanded output tensor of size (2, 1, 2) to the intermediate `T1_g` even though that tensor should have an expanded size of `1 << 60 = 1152921504606846976`.
```
%kernel {
i10 = (nvfuser_index_t)(1152921504606846976);
T1_g[ iblockIdx.x3{2}, bS4{( (nvfuser_index_t)(1152921504606846976) )}, ithreadIdx.x5{2} ] = expand( T0_g[ iS0{2}, bS1{1}, iS2{2} ], {2, i10, 2} )
T2_g[ iblockIdx.x6{2}, bS7{( (nvfuser_index_t)(1152921504606846976) )}, ithreadIdx.x8{2} ]
   = Set( T1_g[ iblockIdx.x3{2}, bS4{( (nvfuser_index_t)(1152921504606846976) )}, ithreadIdx.x5{2} ], cache_op=Streaming )
```
The expression evaluator currently doesn't complain that we are binding the value 1 to a constant int with value 1152921504606846976 since in fact the expanded extent of that axis has a cast, i.e. it is the non-immediate constant integer `(nvfuser_index_t)1152921504606846976`. That is the actual `Val*` being bound, so the `ExpressionEvaluator` does not complain that the values conflict, but once the cast is simplified away in #973, we see the error.

This PR fixes this problem by allocating intermediate buffers in the same way we allocate outputs, meaning we will also allocate expanded tensors. `ExpressionEvaluator` is updated to check for expanded extents when binding.


jacobhinkle:
!build --diff

jacobhinkle:
!build

naoyam:
I'll update this PR for shift

naoyam:
FYI, here's an example of the error you can see with compute-sanitizer.

```
[ RUN      ] ShiftTest.Shift2
========= Invalid __global__ write of size 4 bytes
=========     at 0x920 in <unnamed>::nvfuser_none_f0_c0_r0_g0(<unnamed>::Tensor<float, (int)2, (int)2>, <unnamed>::Tensor<float, (int)2, (int)2>, <unnamed>::Tensor<float, (int)2, (int)2>, <unnamed>::Tensor<float, (int)2, (int)2>, <unnamed>::Tensor<float, (int)2, (int)2>, <unnamed>::Tensor<float, (int)2, (int)2>, <unnamed>::Tensor<float, (int)2, (int)2>, <unnamed>::Tensor<float, (int)2, (int)2>, <unnamed>::Tensor<float, (int)2, (int)2>, <unnamed>::Tensor<float, (int)2, (int)2>, <unnamed>::Tensor<float, (int)2, (int)2>, <unnamed>::Tensor<float, (int)2, (int)2>)
=========     by thread (0,0,0) in block (0,0,0)
=========     Address 0x7eff3fa0098c is out of bounds
=========     and is 1 bytes after the nearest allocation at 0x7eff3fa00800 of size 396 bytes
=========     Saved host backtrace up to driver entry point at kernel launch time
=========     Host Frame: [0x305c18]
=========                in /usr/lib/x86_64-linux-gnu/libcuda.so.1
=========     Host Frame:/home/nmaruyama/nvfuser/debug1/csrc/executor.cpp:1780:nvfuser::FusionExecutor::runFusion(nvfuser::KernelArgumentHolder&, nvfuser::LaunchParams const&, nvfuser::CompileParams, std::vector<at::Tensor, std::allocator<at::Tensor> >) [0x14f66c1]
=========                in /home/nmaruyama/nvfuser/debug1/build/libnvfuser_codegen.so
=========     Host Frame:/home/nmaruyama/nvfuser/debug1/csrc/executor.h:133:nvfuser::FusionExecutor::runFusion(c10::ArrayRef<c10::IValue> co
```

jacobhinkle:
> FYI, here's an example of the error you can see with compute-sanitizer.

Ah, ok thanks.

naoyam:
The python test still works fine:

```
 python python_tests/test_python_frontend.py  -k test_issue1270                                                                                                ~/nvfuser/debug1
.
----------------------------------------------------------------------
Ran 1 test in 2.861s

OK
```

naoyam:
!build

naoyam (approved):
@jacobhinkle I think this should be good to merge. Please check my last commit and merge the PR if it looks good to you.

jacobhinkle:
> @jacobhinkle I think this should be good to merge. Please check my last commit and merge the PR if it looks good to you.

After looking at the commit, I see what you mean now. Thank you very much for contributing a fix!

Issue #1470, opened 2023-12-07
Fusion with ConsecutiveOuterWelford failed in nvrtc compile


liqiangxl:
With current [12/07/2023] main branch, the following fusion failed.
```
TEST_F(
    NVFuserTest, ConsecutiveOuterWelford) {
  std::unique_ptr<Fusion> fusion_ptr = std::make_unique<Fusion>();
  auto fusion = fusion_ptr.get();
  FusionGuard fg(fusion);

  const int batch_size = 8192;
  const int hidden_size = 1024;
  DataType input_dtype = DataType::Half;
  auto tv0 = makeContigTensor(2, input_dtype);
  fusion->addInput(tv0);
  auto tv1 = castOp(DataType::Float, tv0);
  auto tv2 = add(tv1, tv1);
  auto tv3 = Welford(tv2, {0});
  auto tv4 = broadcast(tv3.avg, {true, false});
  auto tv5 = div(tv2, tv4);

  auto tv6 = exp(tv5);
  auto tv7 = Welford(tv6, {0});
  auto tv8 = broadcast(tv7.avg, {true, false});
  auto tv9 = div(tv6, tv8);

  fusion->addOutput(tv5);
  fusion->addOutput(tv9);

  auto options = at::TensorOptions()
                     .dtype(data_type_to_aten(input_dtype))
                     .device(at::kCUDA, 0);
  auto t0 = at::randn({batch_size, hidden_size}, options);
  std::vector<c10::IValue> inputs{t0};

reinterpret_cast<float*>(shared_mem), &T31[0]);  
  FusionExecutorCache fec(std::move(fusion_ptr));
  auto cg_outputs = fec.runFusionWithInputs(inputs);
}
```

err msg:
```
  // failed, 
  // CUDA NVRTC compile error: __tmp_kernel_outer_persistent_f0_c1_r0_g0.cu(9643): error: no instance of overloaded function "<unnamed>::fused_reduction::ParallelReduce<X_BLOCK, Y_BLOCK, Z_BLOCK, X_THREAD, Y_THREAD, Z_THREAD, PERSISTENT_REDUCTION, BROADCAST>::welfordGroupOuter [with X_BLOCK=1, Y_BLOCK=0, Z_BLOCK=3, X_THREAD=1, Y_THREAD=0, Z_THREAD=3, PERSISTENT_REDUCTION=true, BROADCAST=true]" matches the argument list
[...]

PR #1475, opened 2023-12-07
avoid mixed usage of Array<T, N> and T[N] when passing to template function

liqiangxl:
Issue: https://github.com/NVIDIA/Fuser/issues/1470
In the generated code of the case posted in the issue page:
`nvfuser_index_t T5[4]` is aliased as `Array<float, 4> T9.` `float T4[4]` is aliased as `auto& T10 = T4`.
Then T9 and T10 are used to call `welfordGroupOuter` and causes a compilation error due to type mismatch: T9 is a custom array type, while T10 is a native float array.

Fix: always pass the regular array of registers to template function.


liqiangxl:
!build --diff --diff-bench

liqiangxl:
!build --diff --diff-bench

liqiangxl:
codediff in matmul cases are caused by `Cleanup matmul test utilities (#1804)`

csrc/codegen.cpp
```diff
@@ -246,6 +249,24 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
     return val_to_name_.at(v);
   }
 
+  // If the variable is an aligned array, append ".array" to use the reguar
+  // array. This avoid the type mismatch in template functions when one of the
+  // arguments is an aligned array (Array<T,N>) while the other is a regular
+  // array T[N].
+  std::string genVarForTemplateFunction(Val* v) {
```
zasdfgbnm, line 256: Should this be named something like `maybeConvertAlignedArrayToRawArray`?

zasdfgbnm (approved):

csrc/codegen.cpp
liqiangxl: I rename to `genVariableNameConvertAlignedArray` to reflect its two functions: (1) call `genVariableName` and (2) ConvertAlignedArrayToRawArray

PR #1476, opened 2023-12-07
Unify commit-wait style sync

zasdfgbnm:
`kir::CpAsyncWait` and `kir::CpAsyncCommit` is now renamed as `kir::AsyncWait` and `kir::AsyncCommit`, and they are now also used for synchronizing `cp.async.bulk` and `wgmma`. Using a unified IR node reduces the number of code, and therefore in the future it is easier to add another type of sync.

### DIFF
--- a/csrc/kernel_ir.h
+++ b/csrc/kernel_ir.h
@@ -572,6 +572,9 @@ class AsyncCommit final : public Expr {
   std::string toInlineString(int indent_size = 0) const override;
 
   const char* ptx() const;
+
+  //! Returns if the corresponding PTX needs a `:memory` in the end, this value
+  //! will be used to set AsmOptions::memory when lowering to inline PTX.
   bool memory() const;
 
   AsyncOpType asyncOpType() const {
--- a/csrc/type.h
+++ b/csrc/type.h
@@ -1008,7 +1008,4 @@ constexpr auto toUnderlying(E e) noexcept {
 
 enum class AsyncOpType { CpAsync, CpAsyncBulk, WgMma };
 
-const char* getAsyncOpCommitPTX(AsyncOpType op_type);
-const char* getAsyncOpWaitPTX(AsyncOpType op_type);
-
 } // namespace nvfuser


csrc/device_lower/pass/inline_ptx.cpp
```diff
@@ -264,25 +246,6 @@ class LowerToInlinePtx : public kir::ExprMutator {
                 /*volatile=*/true,
                 /*memory=*/false,
                 /*readable_outputs=*/{0}}));
-
```
zasdfgbnm, line 267: Sync insertion is now moved to indexing, still not good, but at least `LowerToInlinePtx` becomes cleaner.

zasdfgbnm:
!build

csrc/type.h
```diff
@@ -1006,4 +1006,9 @@ constexpr auto toUnderlying(E e) noexcept {
   return static_cast<std::underlying_type_t<E>>(e);
 }
 
+enum class AsyncOpType { CpAsync, CpAsyncBulk, WgMma };
+
+const char* getAsyncOpCommitPTX(AsyncOpType op_type);
+const char* getAsyncOpWaitPTX(AsyncOpType op_type);
```
naoyam, line 1012: Are they actually used?

csrc/kernel_ir.h
```diff
@@ -521,84 +519,64 @@ class MBarrierWait final : public Expr {
...
-
-class CpAsyncBulkS2GCommit final : public Expr {
- public:
-  using Expr::Expr;
-
-  explicit CpAsyncBulkS2GCommit(IrBuilderPasskey passkey);
-
-  NVFUSER_DECLARE_CLONE_AND_CREATE
+  const char* ptx() const;
+  bool memory() const;
```
naoyam, line 575: What does this function mean?

naoyam (approved):
LGTM. Just a few minor comments.

csrc/type.h
zasdfgbnm: No, they are dead code. Removed.

csrc/kernel_ir.h
zasdfgbnm: Added a comment in the code. It basically tells if the ptx should have a `:memory` in the end. It is strange that some PTX needs it, some not....

drzejan2: Side note, similar comment / documentation should be added to `AsyncWait` definition.

Issue #1401, opened 2023-11-29
Alias analysis on inputs and intermediate tensors


jjsjann123:
Currently our alias analysis only handles output to input / output to output.

This is slightly different from what bookend optimization does, which seems to give better perf in some attention block. Bookend optimization would strip meta-operation (view/squeeze/slice e.t.c.) on the bookend of a fusion and pass in tensor with mutated strides.
This should be doable given the alias analysis information. We just need some mechanics to plumb this through our segmenter to know *when* and *where* to stop merging segments, as well as *when* to initiate segmentation even though the entire fusion could be scheduled as a single kernel.

For example, when we handle meta operations on inputs, we might need to consider whether stripping aliases are beneficial, even though they give a slightly smaller fusion IR.

```
  auto tv0 = makeSymbolicTensor(2);
  auto s0 = IrBuilder::create<Val>(DataType::Index);
  auto s1 = IrBuilder::create<Val>(DataType::Index);
  auto tv1 = unsqueeze(tv0, -1);
  auto tv2 = expand(tv1, {s0, s1});
  // ... some fusion uses both tv0 and tv2
```

If we decided to strip the unsqueeze and expand from the actual codegen kernel, we are going to pass two tensor inputs to the kernel `tv0` and `tv2` here.
Even though the two tensors are just aliasing and cache would help us, it might still be worth keeping these operations just so we have a single input to the fusion.

jjsjann123:
cc'ing @wujingyue

wujingyue:
We are going to talk about this soon, but just a preliminary thought: for this particular example, why do we **need** alias analysis to generate optimal code? Even without alias analysis, I expect nvFuser to already inline `unsqueeze` and `expand` (into downstream ops) and make them cheap index calculation.

wujingyue:
@jjsjann123 and I looked at some other cases. Here's an interesting one: 

```
def nvfuser_fusion_id23(fd : FusionDefinition) -> None :
    S0 = fd.define_scalar(None, dtype=DataType.Double)
    S1 = fd.define_scalar(None, dtype=DataType.Double)
    S2 = fd.define_scalar(None, dtype=DataType.Double)
    T3 = fd.define_tensor(shape=[-1, -1], contiguity=[True, True], dtype=DataType.BFloat16, is_cpu=False)
    T4 = fd.define_tensor(shape=[-1, -1, -1], contiguity=[True, True, True], dtype=DataType.BFloat16, is_cpu=False)
[...]

PR #1478, opened 2023-12-07
Mark aliases during scheduling. 

wujingyue:
This is one of the many steps to solve #1401.

Currently, we run alias analysis and mark aliases before segmentation. This can't handle the output-to-output alias patterns we observed in the nanogpt2 benchmark, because some schedulers fork outputs and therefore break the aliasing chain observed before segmentation (and thus before scheduling). https://github.com/NVIDIA/Fuser/issues/1401#issuecomment-1840137527 shows an example for curious minds. 

This PR changes the algorithm so it runs alias analysis both before segmentation and during scheduling. The former is used to change layouts to enable aliases, and the latter is used to detect and mark aliases in each segment. This allows segmentation to split out meta-op-only regions so the rest is easier to optimize. 

This PR doesn't solve the aforementioned output-to-output alias patterns. I'll need to add logic to segment out meta-op-only output-to-output regions in later PRs.


csrc/device_lower/pass/expr_sort.cpp
```diff
@@ -1749,17 +1749,18 @@ std::vector<Expr*> ExprSegmentationSorter::getExprs() const {
...
+  // this check is needed at this moment.
   if (fusion->isNoOp()) {
     return {};
   }
-  NVF_ERROR(fusion != nullptr);
+
   ExprSegmentationSorter sorter(fusion);
   sorter.sort();
-  auto sorted_exprs = sorter.getExprs();
-  NVF_ERROR(
```
wujingyue, line 1759: `ExprSegmentationSorter` skips generating code for output aliases, so the expression list can be empty.

wujingyue:
!build

csrc/optimization/alias_analysis.cpp
```diff
@@ -132,6 +132,8 @@ void AliasFinder::handle(const ViewOp* view) {
 
   std::unordered_map<IterDomain*, IterDomain*> in_rfactor_to_out_root =
       PairwiseRootDomainMap(in, out).mapBroadcast(true).mapProducerToConsumer();
+  std::unordered_map<IterDomain*, IterDomain*> out_root_to_in_rfactor =
```
jjsjann123, line 135: errr. This is coming from #1489 right?

Something doesn't look right on github's code diff.

jjsjann123 (commented):
errr. code diff looks wrong on github GUI. I'll try to do a local diff and review that way.

jjsjann123:
Ha I think it's just the conflicts that's making it extra confusing to github. @wujingyue would you mind resolving that first and I'll review afterwards :bow:

wujingyue:
!build

csrc/fusion.cpp
```diff
@@ -21,6 +21,7 @@
 #include <iter_visitor.h>
 #include <kernel.h>
 #include <ops/arith.h>
+#include <optimization/alias_analysis.h>
```
jjsjann123, line 24: is this one an accidental addition?

csrc/optimization/alias_analysis.h
```diff
@@ -22,7 +22,9 @@ struct Layout {
 
   // Returns whether this layout is compliant with `required`. This is
   // uni-directional. For example, `contiguity=[t,t]` is compliant with
-  // `contiguity=[f,f]` but not vice versa.
+  // `contiguity=[f,f]` but not vice versa. As a special case,
+  // an empty `required.allocation` indicates no requirements, i.e., the method
+  // always returns true.
```
jjsjann123, line 27: Thanks for the nice comment here!

csrc/scheduler/mark_aliases.cpp
```diff
@@ -0,0 +1,56 @@
...
+       ir_utils::filterByType<TensorView>(fusion->outputs())) {
+    const Val* in = analysis.getAliasedInput(out);
+    if (in == nullptr) {
+      continue;
+    }
+
+    fusion->aliasOutputToInput(
+        out,
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
+        const_cast<Val*>(in),
```
jjsjann123, line 46: nitpick, not your problem, but I should probably have updated the signature in `aliasOutputToInput` to take `const Val*` instead.

jjsjann123 (approved):
LGTM

csrc/fusion.cpp
wujingyue: Done.

naoyam (approved):
LGTM as well

wujingyue:
!build

PR #32, opened 2023-03-19
WIP: IterDomain Graphs

csarofeen:
Build out of Iter domain graphs as infrastructure. I kept these concepts separate from the compute_at_map as that may need to be reimplemented later based on this new concept/infrastructure.

This new concept of IterDomainGraphs will eventually replace all our index and parallelization logic. This infrastructure is to make it easier to work with iter domain graphs for processes like accurate broadcast resolution/promotion. IdGraph or IterDomainGraphs could also directly replace BestEffortReplay and similar mappings across producer-consumers.

I added a couple interesting tests that I want to make work but still don't.

PR #1479, opened 2023-12-08
Minor cleanup of transform_iter

naoyam:
Taken out of #32. The `ForwardingInfo` class is going to be used from `IdModel`, so it's moved to the header file


naoyam:
!build --diff --diff-bench

zasdfgbnm (approved):

PR #1480, opened 2023-12-08
Use separate swizzzle for A and B for hopper MMA

zasdfgbnm:


zasdfgbnm:
!build

drzejan2 (approved):
Sorry for blocking this PR. The prepared changes make sense, I forgot myself that we should cover test case where inputs have different swizzle modes.

Thanks!

PR #1199, opened 2023-10-31
Activation benchmarks

Priya2698:
- `Gelu` forward and backward: Ports https://github.com/NVIDIA/Fuser/blob/main/benchmark/gelu_backward.cpp
- `Scale-bias-relu` forward and backward: Ports SBR benchmark at https://github.com/NVIDIA/Fuser/blob/main/benchmark/scale_bias_relu.cpp.

PR #1486, opened 2023-12-08
Switch to 2D tensors in `gelu` and `scale-bias-relu` benchmarks.

Priya2698:
The current benchmarks use 3D and 4D tensor (for parity with CPP benchmarks), however for benchmarking purposes, there is no specific requirements for the inputs to be 3D/4D. 
Switching to 2D tensors will also reduce time as we start to increase the number of inputs as pointed by @jacobhinkle earlier (https://github.com/NVIDIA/Fuser/pull/1199#discussion_r1387385506).


jacobhinkle (approved):
LGTM. Makes sense to me that these pointwise ops need at most 2D tests.

PR #1488, opened 2023-12-08
Update the input sizes in Python Benchmarks

Priya2698:
This updates the input sizes to be better reflective of the actual sizes seen in LLMs. 
`step_size` in 2D inputs was adjusted such that runtime of the benchmarks was limited to few hours (< 2hrs ideally).


Priya2698:
@kevinstephano The 2D input sizes are the ones derived from LLM sizes. For weekly, we can switch to a step size of 8. For nightly, I have 64 as the step size right now. If this increases the runtime to > 2hrs, we can switch to running only the actual LLM sizes or a larger step size ( + the actual LLM sizes -- some LLM sizes will be skipped with step sizes > 64).

python_benchmarks/global_params.py
```diff
@@ -2,6 +2,41 @@
 from typing import Union, List, Tuple
 from nvfuser import DataType
 from .core import DEVICE_PROPERTIES
+import numpy as np
+
+# Model Parameters from LLMs (GPT2/3, PaLM, LLama)
+
+# Input sequence length
+SEQ_LENGTH_MIN = 1024
+SEQ_LENGTH_MAX = 16384
```
wujingyue, line 11: Are they used anywhere?

Priya2698: No, they are not. I was hoping to use them in attention benchmarks.
I'll remove them for now and add as needed.

wujingyue: (Same for other `*_MIN` and `*_MAX` variables)

python_benchmarks/global_params.py
```diff
@@ -11,49 +46,63 @@ def generate_input_sizes(dims: Union[int, List] = 2) -> List[Tuple]:
...
+                )
 
             inputs.extend(
                 [
-                    (i, j, k, l)
-                    for i in batch_range
-                    for j in channel_range
-                    for k in spatial_range
-                    for l in spatial_range
+                    (N, C, HW, HW)
```
wujingyue, line 93: https://peps.python.org/pep-0008/#function-and-variable-names

wujingyue (approved):
LGTM with comments

python_benchmarks/global_params.py
```diff
@@ -2,6 +2,37 @@
...
+    768,
+    1024,
+    1280,
+    1536,
+    1600,
+    2048,
+    2560,
+    4096,
+    5120,
+    5140,
```
kevinstephano, line 25: Do you know where 5140 came from?  I think Liqiang discovered it was a typo.

Priya2698: GPT-3 13B model.

wujingyue: Since you raised this question, do you two think it's useful to annotate (in code comment) which model variation a size comes from?

Priya2698:
Update: Pytest memory usage increases with the number of tests. PR #1532 was aimed at reducing the number of tests in each file. I am removing the weekly coverage from this PR since at this point it will run into OOM killing the pytest process, while I investigate what contributes to this increasing memory usage.
Possibly, it is the kernel cache entries: https://github.com/NVIDIA/Fuser/pull/1532#issuecomment-1857349564.
Pytest is known to become slow with increasing number of tests as well: https://stackoverflow.com/a/20934950 but it seems unlikely that it alone will consume ~40GB for ~12.5K tests. Tracking this in Issue #1537.
CC: @wujingyue @naoyam

python_benchmarks/global_params.py
Priya2698: Done

Priya2698:
**Update**: PR #1543 fixes the GPU OOM errors. 
There are runtime errors in softmax benchmarks for a subset of hidden size (Issue #1548)

Priya2698:
Update: Issue #1548 was resolved by @liqiangxl in PR #1561.

python_benchmarks/global_params.py
```diff
@@ -11,49 +35,76 @@ def generate_input_sizes(dims: Union[int, List] = 2) -> List[Tuple]:
...
+            spatial_range = [7 * 2**i for i in range(5)]  # {7, 112}
+            input_ranges.append((batch_range, channel_range, spatial_range))
+
+            for batch_range, channel_range, spatial_range in input_ranges:
+                inputs.extend(
+                    [
+                        (n, c, hw, hw)
+                        for n in batch_range
+                        for c in channel_range
+                        for hw in spatial_range
```
wujingyue, line 89: Consider https://docs.python.org/3/library/itertools.html#itertools.product

python_benchmarks/global_params.py
```diff
@@ -11,49 +35,76 @@ def generate_input_sizes(dims: Union[int, List] = 2) -> List[Tuple]:
...
             inputs.extend(
                 [
-                    (i, j, k, l)
-                    for i in batch_range
-                    for j in channel_range
-                    for k in spatial_range
-                    for l in spatial_range
+                    (n, c, hw, hw)
+                    for n in [128, 256]
+                    for (c, hw) in [
```
wujingyue, line 97: ditto

wujingyue (approved):

Issue #1126, opened 2023-10-23
Split/merged expanded broadcast dimensions can remain broadcast. 


wujingyue:
Spinoff of https://github.com/NVIDIA/Fuser/pull/1124#discussion_r1368077711. 

In the following example, nvFuser could have made the size-2 and size-3 dimensions a broadcast dimension to save space and compute. 

```
TEST_F(AliasAnalysisTest, View_SplitExpandedBroadcast) {
  Fusion fusion;
  FusionGuard fg(&fusion);

  TensorView* in = makeContigConcreteTensor({4, 5});
  fusion.addInput(in);
  TensorView* out = broadcast(in, {false, false, true});
  out = expand(
      out,
      {IrBuilder::create<Val>(4),
       IrBuilder::create<Val>(5),
       IrBuilder::create<Val>(6)});
  out = reshape(out, {4, 5, 6}, {4, 5, 2, 3});
  fusion.addOutput(out);

  FusionExecutor fe;
  at::Tensor in_tensor =
      at::randn({4, 5}, at::dtype(at::kFloat).device(at::kCUDA, 0));
  fe.compileFusion(&fusion, {in_tensor});
  at::Tensor out_tensor = fe.runFusion({in_tensor})[0];
  EXPECT_THAT(out_tensor.strides(), ElementsAre(5, 1, 0, 0));
}
```

https://github.com/NVIDIA/Fuser/commit/085f6a71d8d48a73018c34da69ef27742809a4e8 added a similar reproducer for merge. 
```[tasklist]
### Tasks
```

jacobhinkle:
This is actually a bug right? `Split` doesn't look at expanded extent so I'd be surprised if this executes or gives the proper output shape. It should be a pretty simple fix though.

jacobhinkle:
We have tests that cover this issue (it applies to `Merge` as well), but they only perform the check after the broadcast is resolved. Adding the stride checks there causes a failure as expected. A proper fix here will require us to insert both `BroadcastOp` and `ExpandOp` in the graph. I'll work on it soon.

PR #1489, opened 2023-12-08
Fix a bug related to expanded broadcast IterDomains.

wujingyue:
My #1295 changed some tests in the wrong way so this bug has been hidden since then.

### DIFF
--- a/csrc/optimization/alias_analysis.cpp
+++ b/csrc/optimization/alias_analysis.cpp
@@ -132,6 +132,8 @@ void AliasFinder::handle(const ViewOp* view) {
 
   std::unordered_map<IterDomain*, IterDomain*> in_rfactor_to_out_root =
       PairwiseRootDomainMap(in, out).mapBroadcast(true).mapProducerToConsumer();
+  std::unordered_map<IterDomain*, IterDomain*> out_root_to_in_rfactor =
+      PairwiseRootDomainMap(in, out).mapBroadcast(true).mapConsumerToProducer();
 
   // Collect the allocation order of `in`'s rfactor domain and thus `out`'s root
   // domain.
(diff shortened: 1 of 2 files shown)


wujingyue:
!build

test/test_alias.cpp
```diff
@@ -172,7 +172,7 @@ TEST_F(AliasAnalysisTest, View_SplitExpandedBroadcast) {
 
   optimization::AliasAnalysisResult alias_analysis =
       optimization::findAliases(&fusion);
-  EXPECT_EQ(alias_analysis.findRoot(out), expand_out);
+  EXPECT_EQ(alias_analysis.findRoot(out), out);
```
jacobhinkle, line 175: Why is `findRoot(out)` not equal to `in` in this case? Is this just because of #1126? If so, maybe we should add a comment that ideally we would alias the input here, and refer to that issue.

wujingyue:
> A bit more explanation would be much appreciated. As I understand it, this applies when the input of a `ViewOp` has an allocation domain that is a permutation of its rfactor. The input's allocation domain is used to determine the contiguity of the output **at its root domain** (as opposed to `.contiguity()` which refers to `out`'s allocation domain). That is then propagated from root to rfactor to check if we require a copy. The change seems reasonable to me but it's not clear to me yet how this fails currently.

Your understanding described above is all correct. The code currently fails because due to #1126 (or a similar issue that your #1174 is trying to fix) nvFuser doesn't propagate the **has-expanded-extent** flag from `in_rfactor` to `out_root`. As a result, the expressions between `out_root` and `out_rfactor` see `out_root`'s IterDomains as a `t` or `f` instead of an `e`: https://github.com/NVIDIA/Fuser/blob/4697acead4f741d2dc3ed720fbe205b9ff00de49/csrc/optimization/alias_analysis.cpp#L86.

test/test_alias.cpp
wujingyue: > Is this just because of #1126?

No, no-alias is the right answer here. `in`'s shape is `{4,5},stride={5,1}` and `out`'s shape is `{40,3}`. No matter what strides we pick for `out`, there has to be a memcpy. 

Unlike this case, `View_ForwardExpandedBroadcast` can be made an alias as long as `out`'s shape is `{20,6},strides={1,0}`.

jacobhinkle: Ah ok, got it. Of course. Thanks for the explanation.

wujingyue:
!build

csrc/optimization/alias_analysis.cpp
```diff
@@ -182,9 +194,14 @@ void AliasFinder::handle(const ViewOp* view) {
     }
   }
 
+  auto try_map_to_out_root = [&in_rfactor_to_out_root](IterDomain* id) {
+    const auto i = in_rfactor_to_out_root.find(id);
+    return i == in_rfactor_to_out_root.end() ? id : i->second;
+  };
+
```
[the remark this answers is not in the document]
wujingyue: Done!

csrc/optimization/alias_analysis.cpp
```diff
@@ -132,6 +132,8 @@ void AliasFinder::handle(const ViewOp* view) {
 
   std::unordered_map<IterDomain*, IterDomain*> in_rfactor_to_out_root =
       PairwiseRootDomainMap(in, out).mapBroadcast(true).mapProducerToConsumer();
+  std::unordered_map<IterDomain*, IterDomain*> out_root_to_in_rfactor =
```
jjsjann123, line 135: nitpick: Please add a comment here explaining why we need to map everything to in_rfactor to detect the correct expand flag. A link to the issue would be great as well!

jjsjann123 (approved):
LGTM

csrc/optimization/alias_analysis.cpp
wujingyue: The comment you are looking for was at Line 151.

PR #1492, opened 2023-12-09
Relax the allocation domain constraint. 

wujingyue:
Currently, `MarkAliasPass` skips an output when its allocation domain is set at all. This PR makes it smarter to check compliance.

This PR prepares #1478 for landing. Before segmentation, we'll run alias analysis on the un-segmented fusion and mark layouts for aliasing. During scheduling, we'll run alias analysis again on segmented fusions and have to recognize the preferred layout is compliant with the previously marked layout.

### DIFF
--- a/csrc/optimization/alias_analysis.cpp
+++ b/csrc/optimization/alias_analysis.cpp
@@ -381,7 +381,7 @@ void AliasAnalysisResult::finalize(Fusion* fusion) {
 
     const Layout preferred_layout = preferredLayout(out);
     if (out->hasAllocation() &&
-        !preferred_layout.isCompatibleWith(
+        !preferred_layout.isCompliantWith(
             {out->getAllocationDomain(), out->getContiguity()})) {
       continue;
     }
@@ -452,10 +452,30 @@ std::string Layout::toString(const int indent_size) const {
   return ss.str();
 }
 
-bool Layout::isCompatibleWith(const Layout& other) const {
-  // Most basic for now.
-  return allocation_domain == other.allocation_domain &&
-      contiguity == other.contiguity;
+namespace {
+bool contiguityIsCompliant(
+    const std::optional<bool>& actual,
+    const std::optional<bool>& required) {
+  if (actual == true && required == false) {
+    return true;
+  }
+  return actual == required;
+}
+} // namespace
+
+bool Layout::isCompliantWith(const Layout& required) const {
+  if (allocation_domain != required.allocation_domain) {
+    // This can be relaxed by allowing broadcast dimensions to be ordered
+    // differently.
+    return false;
+  }
+
+  for (const auto i : c10::irange(allocation_domain.size())) {
+    if (!contiguityIsCompliant(contiguity[i], required.contiguity[i])) {
+      return false;
+    }
+  }
+  return true;
 }
 
 } // namespace nvfuser::optimization
--- a/csrc/optimization/alias_analysis.cpp
+++ b/csrc/optimization/alias_analysis.cpp
@@ -374,15 +374,15 @@ const TensorView* AliasAnalysisResult::getAliasedInput(
 void AliasAnalysisResult::finalize(Fusion* fusion) {
   for (TensorView* out :
        ir_utils::filterByType<TensorView>(fusion->outputs())) {
-    // Lazy move: we could check compatibility and only give up when
-    // the allocation domain is incompatible with what we prefer for
-    // aliasing.
-    if (out->hasAllocation()) {
+    const Val* in = findRoot(out);
+    if (!in->isFusionInput()) {
       continue;
     }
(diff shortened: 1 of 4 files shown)


wujingyue:
!build

csrc/optimization/alias_analysis.cpp
```diff
@@ -374,15 +374,15 @@ const TensorView* AliasAnalysisResult::getAliasedInput(
 void AliasAnalysisResult::finalize(Fusion* fusion) {
   for (TensorView* out :
        ir_utils::filterByType<TensorView>(fusion->outputs())) {
-    // Lazy move: we could check compatibility and only give up when
-    // the allocation domain is incompatible with what we prefer for
-    // aliasing.
-    if (out->hasAllocation()) {
+    const Val* in = findRoot(out);
+    if (!in->isFusionInput()) {
```
jjsjann123, line 378: what does this mean to output->output aliasing -> DuplicateOutputs.

We do have those in our cpp tests and I thought they are resolved via aliasing?!

csrc/optimization/alias_analysis.cpp
```diff
@@ -392,7 +392,8 @@ void AliasAnalysisResult::finalize(Fusion* fusion) {
 
 Layout AliasAnalysisResult::preferredLayout(const Val* v) const {
   const TensorView* tv = dynamic_cast<const TensorView*>(v);
-  NVF_CHECK(
+  NVF_ERROR(v != nullptr); // Just to make clang-tidy happy.
```
jjsjann123, line 395: nitpick: remove this and later in the error message we can do `v != nullptr ? "nullptr" : v->toSTring()`.
Does that keep clang happy as well?

csrc/optimization/alias_analysis.cpp
```diff
@@ -452,4 +453,10 @@ std::string Layout::toString(const int indent_size) const {
   return ss.str();
 }
 
+bool Layout::isCompatibleWith(const Layout& other) const {
+  // Most basic for now.
```
jjsjann123, line 457: :laughing: 

nitpick: what's more advanced than the most basic? I know we can relax contiguity and that's about it?

The function name doesn't suggest the direction of which one between this/other can be relaxed. We might as well just rename it as `isSameAs` to stick better to what it's doing for now.

test/test_alias.cpp
```diff
@@ -299,6 +299,33 @@ TEST_F(AliasTest, View) {
...
+
+  FusionExecutorCache fec(std::move(fusion));
+  at::Tensor in_tensor =
+      at::randn({60}).cuda().as_strided({2, 3, 4}, {2, 20, 5});
+  std::vector<at::Tensor> out_tensors = fec.runFusionWithInputs({in_tensor});
+  ASSERT_EQ(out_tensors.size(), 1);
+  at::Tensor out_tensor = out_tensors[0];
+  testValidate(fec.fusion(), {out_tensor}, {in_tensor}, __LINE__, __FILE__);
+
+  EXPECT_TRUE(out_tensor.is_alias_of(in_tensor));
```
jjsjann123, line 326: Nice!

cc'ing @kevinstephano since you were asking about support like this earlier.

jjsjann123 (approved):
LGTM

csrc/optimization/alias_analysis.cpp
jjsjann123: I'm stamping this, since if tests are passing this is likely just my misunderstanding of the code base. :laughing:

wujingyue: Aren't they currently resolved by `outputs_map` you added? That code remains there in `csrc/executor.cpp` -- this PR doesn't touch it.

csrc/optimization/alias_analysis.cpp
wujingyue: > What's more advanced than the most basic?

I was thinking about broadcast dimensions can be placed differently. 

> I know we can relax contiguity and that's about it?

Thanks -- I totally forgot about that. Yes, it makes sense to have this method be uni-directional. Let me do that in this PR.

csrc/optimization/alias_analysis.cpp
jjsjann123: sorry, I was dumb... yeah duplicated ouptuts isn't related here. I was thinking about common subexpression... but we don't have any support there neither at this point.

csrc/optimization/alias_analysis.cpp
wujingyue: Done.

csrc/optimization/alias_analysis.cpp
wujingyue: Done.

wujingyue:
!build

PR #1493, opened 2023-12-09
Duplicate the code of `Swizzle2D` and name it as `Swizzle`

zasdfgbnm:
The current implementation of `Swizzle2D` is so weird that I don't want to use it for implementing the smem layout of Hopper. So I selectively copy-pasted its code and name the new op `Swizzle`. In the future, I am expecting to deprecate `Swizzle2D`. Except for the newly added test, this entire PR is copy-pasted from existing `Swizzle2D`. But not all of `Swizzle2D` are duplicated. I want the new `Swizzle` op to remain simple and stupid, that is,  it will behave similar to `Split` and `Merge`:
- It is not handled separately in indexing, `IndexSwizzle` only applies to `Swizzle2D` and is unrelated to the new `Swizzle`. The new `Swizzle` is just indexed by the most common back propagation like `Split` and `Merge`.
- In `TransformReplay` and `BestEffortReplay`, there is no mechanism for skipping it
- Unlike `Swizzle2D`, the inputs and outputs of `Swizzle` are not mapped in compute at map

And  I want the new `Swizzle` to remain like this in the future.

Currently, this new `Swizzle` is not capable of replacing the existing `Swizzle2D` for supporting Turing and Ampere matmul, because the lack of the above mentioned mechanism that I intentionally kept out. In the long future, `Swizzle` in combination with allocation domain will be used to support Turing and Ampere, in a way as described in https://github.com/NVIDIA/Fuser/issues/986, but this requires us to be able to do forward propagation for ID ops for indexing. In the near future, `Swizzle` will be used to support swizzling smem for Hopper, at least for supporting tests in https://github.com/NVIDIA/Fuser/blob/main/test/test_mma.cpp.

I understand that, in general, code reuse should be preferred compared to copy-pasting code, but for this specific case, I prefer to just copy past the code. Because the `Swizzle2D` will be removed completely, and in the intermediate state today, I do not want to share any code with it, so that I will not have to validate if my changes in `Swizzle` is compatible with the legacy `Swizzle2D`. I want to keep `Swizzle2D` completely untouched, until one day I completely remove it. I believe copy-pasting has the lowest maintenance cost.


test/test_swizzle.cpp
```diff
@@ -651,4 +651,43 @@ TEST_F(SwizzleTest, TransformPropagatorSkipSwizzleOnTarget) {
...
+  fusion->addInput(tv0);
+  auto tv1 = set(tv0);
+  tv1->split(1, 8);
+  tv1->split(0, 8);
+  tv1->reorder({{2, 1}});
+  tv1->swizzle(SwizzleType::XOR, 2, 3);
+  tv1->reorder({{2, 1}});
+  tv1->merge(0);
+  tv1->merge(1);
+  tv1->commitLeafToRFactor();
```
zasdfgbnm, line 667: I can not think of a better way than just create a "view-like" op and check the input and output.

wujingyue:
This PR is beyond my knowledge to review. But I'll keep myself CC'ed because I'm curious about where the swizzling work is going!

naoyam:
I haven't looked at the actual changes yet, but could you be more explicit about what points should be focused for this review? IIUC, the new interface is not a replacement of the existing one yet, so what are supposed to do at this point?

zasdfgbnm:
> I haven't looked at the actual changes yet, but could you be more explicit about what points should be focused for this review? IIUC, the new interface is not a replacement of the existing one yet, so what are supposed to do at this point?

Almost the entire PR (except for the new test) is copy-pasted from `Swizzle2D`, so I don't think there is much interesting code to read except for the test. I think reviewing this PR is mostly reviewing if the design (what I intentionally didn't copy-paste) and plan makes sense.

The design for now is: I want to have a swizzle op that is not special handled in `TransformReplay` and `BestEffortReplay`, and `ComputeAtMap`. During indexing, when we see it, we just generated an XOR and move on. They are not mapped in CA map or skipped in replays. There is no "index, update leaf, index again" thing like `Swizzle2D`. The motivation for adding this  now is, I am trying to use swizzle for Hopper mma, I put the swizzle op between allocation domain and root domain, but the generated index does not have an XOR as I expected. I am not interested in digging into why the currently does not work, and think about how to hack it differently to make it work. I just want a new op that is just completely clean and trivial.

The future plan is well described by Jingyue's comment https://github.com/NVIDIA/Fuser/issues/986#issuecomment-1789344648, we will schedule fusions like that figure and we will index that schedule.

naoyam (approved):
LGTM.

PR #1495, opened 2023-12-10
fix codegen diff failure when number of segments in a fusion is changed

liqiangxl:
Issue: When number of segments in a fusion is reduced from 2 to 1, the codegen diff script tries to compare 2 kernels.
Fix: only compare the first N kernels, where `N = min(number of kernels in old code, number of kernels in new code)`


jacobhinkle (approved):
LGTM. Thanks for the improvement.

PR #1501, opened 2023-12-12
Hopper MMA RS B 32byte swizzle

zasdfgbnm:


zasdfgbnm:
!build

zasdfgbnm:
!build

zasdfgbnm:
!build

zasdfgbnm:
!build

zasdfgbnm:
!build

zasdfgbnm:
!build

csrc/device_lower/pass/index.cpp
```diff
@@ -1458,20 +1472,21 @@ void IndexLowering::handle(const MmaOp* mma) {
...
     // smem.
-    auto base_addr = IrBuilder::baseAddressExpr(mma->inA()->as<TensorView>());
-    int stride_bytes =
-        /*8x8 items each core matrix*/ 64 * /*bytes per item*/ 2;
-    int leading_bytes = /*8x8 items each core matrix*/ 64 *
-        /*number of core matrices*/ (getM(mma->macro()) / 8) *
-        /*bytes per item*/ 2;
+    auto tv = mma->inA()->as<TensorView>();
+    auto base_addr = IrBuilder::baseAddressExpr(tv);
+    int64_t stride_bytes =
```
drzejan2, line 1477: For A tensor shouldn't it be like this:
```
int64_t stride_bytes =  /*8x8 items each core matrix*/ 64L *
        /*number of core matrices*/ (getM(mma->macro()) / 8L) *
        /*bytes per item*/ 2L;
int64_t leading_bytes = /* 16 bytes in a row */ 16L;
```
(based on my understanding of documentation: [link](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#asynchronous-warpgroup-level-matrix-shared-memory-layout-strides)) 

Why for `B` tensor you take into consideration swizzle period but for `A` it's not the case?

zasdfgbnm: You may be right. I need to test it. But for this PR, I am only focusing on B of the RS variant. The change here is just `int` to `int64_t`, the existing logic is not changed. SS's A and B is not working yet. It needs more work, so I decided to leave it for a separate PR. Maybe there needs another `std::swap(stride_bytes, leading_bytes)` like for B?

drzejan2: Ok, it makes sense then - sorry for the noise!

drzejan2 (approved):

PR #1503, opened 2023-12-12
Handle aliases between fusion outputs. 

wujingyue:
This is (roughly) the second-to-last step of solving the example in https://github.com/NVIDIA/Fuser/issues/1401#issuecomment-1832822602. We'll also need to fix the CSE in the framework. 

This PR improves the OptimizeLayout pass to mark aliases between outputs and inserts segment hints. This way, output aliases are more likely segmented out and become a no-op kernel.


wujingyue:
!build

csrc/executor.cpp
```diff
@@ -982,21 +986,51 @@ std::vector<at::Tensor> allocateOutputs(
...
+  // non_alias_out = ...
+  // alias_out_0 = reshape(non_alias_out, ...)
+  // alias_out_1 = reshape(alias_out_0, ...)
+  // ```
+  //
+  // It's fine to compute `alias_out_1` before computing `alias_out_0`: when we
+  // compute `alias_out_1`, `alias_out_0` will be recursively
+  // `ExpressionEvaluator::evaluate`ed. However, `non_alias_out` must be
+  // allocated first.
+  std::vector<std::pair<int64_t, Val*>> sorted_outs;
```
naoyam, line 1004: Just want to make sure I understand this correctly, this sorting is not sorting the outputs themselves but just the order of their allocations, right?

naoyam (approved):
LGTM

csrc/optimization/alias_analysis.h
```diff
@@ -61,14 +61,14 @@ class AliasAnalysisResult {
 
   std::string toString(int indent_size) const;
 
-  // Gets the aliased fusion input of a fusion output. Returns nullptr
-  // when `fusion_out` is not a fusion output or does not alias a fusion input.
-  TensorView* getAliasedInput(const TensorView* fusion_out) const;
+  // Gets the nearest aliased fusion input of a `fusion_out` other than
+  // `fusion_out` itself. Returns null if that doesn't exist.
+  TensorView* getNearestAliasedIo(const TensorView* fusion_out) const;
```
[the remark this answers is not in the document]
wujingyue: Technically, it doesn't have to, but in practice this method is only called with fusion outputs.

csrc/executor.cpp
wujingyue: Yes!

csrc/executor.cpp
```diff
@@ -925,13 +925,18 @@ at::Tensor allocateOutput(
     return ee.evaluate(out_tv).as<at::Tensor>();
   }
 
-  if (aliased_in != nullptr) {
-    const PolymorphicValue& aliased_in_val = ee.evaluate(aliased_in);
+  if (aliased_io != nullptr) {
     NVF_ERROR(
-        aliased_in_val.is<at::Tensor>(),
+        aliased_io->isFusionInput() || aliased_io->isFusionOutput(),
+        "Otherwise, `ee.evaluate` ",
```
jjsjann123, line 931: nitpick, should we mention `aliased_io needs to be an input/output`. I guess the code is self-explanatory, maybe that's not necessary.

jjsjann123: nitpick: Is it fair to say the GPU computation is involved when we alias intermediate tensors? ee just falls to torch, so technically if we are aliasing intermediate, that intermediate could have been just a reshaped input and technically the assert message here isn't valid.

Note I understand why we don't want to support that yet, but let's fix the error message.

csrc/executor.cpp
jjsjann123: nitpick: maybe a note there for us mortals that don't get this quickly enough, that non-aliased outputs needs to be allocated since allocation of aliased tensor needs to reference them.

csrc/kernel_cache.cpp
```diff
@@ -196,17 +196,20 @@ class ArgumentManager {
...
+      PolymorphicValue group_runtime_output;
       if constexpr (std::is_pointer_v<
                         decltype(group_runtime_outputs[group_out_i])>) {
-        fusion_args_.push(*group_runtime_outputs[group_out_i]);
+        group_runtime_output = *group_runtime_outputs[group_out_i];
       } else {
-        fusion_args_.push(group_runtime_outputs[group_out_i]);
+        group_runtime_output = group_runtime_outputs[group_out_i];
       }
+      fusion_args_.push(std::move(group_runtime_output));
```
jjsjann123, line 212: out of curiosity, is this pattern better than the previous push to fusion_args_ directly?

csrc/optimization/alias_analysis.h
```diff
@@ -61,14 +61,14 @@ class AliasAnalysisResult {
 
   std::string toString(int indent_size) const;
 
-  // Gets the aliased fusion input of a fusion output. Returns nullptr
-  // when `fusion_out` is not a fusion output or does not alias a fusion input.
-  TensorView* getAliasedInput(const TensorView* fusion_out) const;
+  // Gets the nearest aliased fusion input of a `fusion_out` other than
```
jjsjann123, line 64:
```suggestion
  // Gets the nearest aliased fusion input/output of a `fusion_out` other than
```
We can also have output -> output aliases now right?

csrc/optimization/optimize_layout.cpp
```diff
@@ -12,21 +12,36 @@
...
+    debug() << "Fusion before OptimizeLayoutPass:" << std::endl;
+    fusion->printMath();
     debug() << "Alias analysis result:" << std::endl;
     debug() << analysis.toString(/*indent_size=*/1) << std::endl;
   }
 
+  // Fusion outputs that are (1) aliased by others and (2) not aliases
+  // themselves. We will later add `segment_set` before them so aliases are
+  // separated from non-aliases and more likely to be accepted by the no-op
+  // scheduler.
```
jjsjann123, line 29: solid plan!

csrc/utils.h
```diff
@@ -23,6 +23,7 @@
 #include <string>
 #include <type_traits>
 #include <typeinfo>
+#include <unordered_map>
```
jjsjann123, line 26: why would we need this header?

jjsjann123 (approved):
Looks great! some minor nitpicks on comments~~

csrc/executor.cpp
wujingyue: Done.

csrc/kernel_cache.cpp
wujingyue: I forgot to revert this. It was a side effect of a lateral change that I later reverted.

csrc/utils.h
wujingyue: I added this for a change that I later reverted. However, I keep it here for IWYU -- the current file already uses `std::unordered_map`.

wujingyue:
!build

wujingyue:
!build

wujingyue:
!build

PR #1507, opened 2023-12-13
Modify the matmul scheduler to allow combining a valid mul sum pair to a mma op.

protonu:
This PR modifies the matmul scheduler to:

1. If there's no mma op present, it detects if there's a possible mul-sum pair that can be replaced by a mma op.
2. Extracts the properties of the mul-sum pair or generated these properties from the existing mma op and checks if the mma op's properties means that it's a valid candidate for the mma scheduler
3. If it was a mul-sum pair and the checks to see if it can be scheduled as a mma op passed, convert the mul-sum pair of ops to a mma op.


naoyam:
Probably this is a question to @zasdfgbnm: why we have `mma_utils.cpp` and `matmul_utils.cpp`. What's the difference?

naoyam:
Do we need to keep both the existing `MmaOp` based code and the new `MulSumAsMmaProps`? If we can fill the property struct from a `MmaOp`, can't we just change the existing code to always work with `MulSumAsMmaProps` only?

csrc/scheduler/matmul_utils.cpp
```diff
@@ -187,10 +195,148 @@ std::string isMatmulFusionDefinitionSupported(
...
+      return "MMA output TV has no reduction domain";
+    }
+  }
+
+  // Quick checks - Fusion
+  {
+    // Fusion should contain at least two inputs (for now)
+    if (minimal_number_of_inputs > fusion_inputs.size()) {
+      return "Fusion inputs contain at least one non-TensorView object";
+    }
```
jacobhinkle, line 226: Do we not support squaring square matrices?

csrc/scheduler/matmul_utils.cpp
```diff
@@ -187,10 +195,148 @@ std::string isMatmulFusionDefinitionSupported(
...
+
+    // MmaOp input is a result of broadcast op with input being fusion input
+    for (const auto* mma_in : mma_inputs) {
+      if (!areMmaOpInputDependeciesValid(mma_in)) {
+        return "MmaOp input has unsupported dependency";
+      }
+    }
+  }
+
+  return "";
```
jacobhinkle, line 311: Nitpick: would `std::optional<std::string>` make a better error type?

jacobhinkle (commented):
Just a quick first pass. I'll have more comments later.

csrc/scheduler/mma_utils.h
```diff
@@ -239,6 +239,32 @@ class DataWrapperOpt {
...
+      TensorView* out_t = nullptr,
+      BroadcastOp* bc_a = nullptr,
+      BroadcastOp* bc_b = nullptr)
+      : mop(m),
+        redop(r),
+        a(at),
+        b(bt),
+        out(out_t),
+        bcast_a(bc_a),
+        bcast_b(bc_b){};
```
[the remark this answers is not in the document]
protonu: I removed it.

csrc/scheduler/matmul_utils.cpp
protonu: This was a bit of existing code - and this is at multiple locations. I don't want to change it in this PR, since this may also mean modifying code that handles the return.

csrc/scheduler/matmul_utils.cpp
protonu: This was existing logic (and code) I moved around.

protonu:
> Do we need to keep both the existing `MmaOp` based code and the new `MulSumAsMmaProps`? If we can fill the property struct from a `MmaOp`, can't we just change the existing code to always work with `MulSumAsMmaProps` only?

@naoyam Do you want to take another look? I tried to address this issues, without modifying function signatures of methods which can be used directly without going through the scheduler class.

zasdfgbnm:
> Probably this is a question to @zasdfgbnm: why we have `mma_utils.cpp` and `matmul_utils.cpp`. What's the difference?

It is mostly historical reason. I don't think we have a clear rule on which should go to `matmul_utils.cpp` and which should go to `mma_utils.cpp`. Right now things used by segmenter that determines whether to accept the fusion and which heuristics to use are in `matmul_utils.cpp`, and things used by `scheduleMatmul` are mostly in `mma_utils.cpp`. But I don't consider this as a strict rule to follow

csrc/scheduler/matmul_utils.cpp
```diff
@@ -307,32 +309,63 @@ std::string getMatmulCompileTimeRejectReason(Fusion* fusion) {
   // 1. check if there is exactly one MmaOp defined in the fusion
   // 2. check if MmaOp inputs match any of supported inputs layout
   // 3. check if fusion represents expressions that are recognized by matmul
-  //    scheduler
+  // scheduler In addition to the above, it may be the case that there is no mma
```
zasdfgbnm, line 312: As an incremental step, it might makes sense to allow preexisting `MmaOp`s. But after things are built up, I prefer to make the pass you added the only source that creates `MmaOp`s, so at that time, we should have "it will always be the case that" instead of "it may be the case that"

csrc/scheduler/matmul_utils.cpp
```diff
@@ -307,32 +309,63 @@ std::string getMatmulCompileTimeRejectReason(Fusion* fusion) {
...
-        return support_status;
-      }
+    auto support_status = isMatmulFusionDefinitionSupported(
+        fusion,
+        mma_exprs.empty() ? mma_from_mul_sums.front().insouts : mma_inputs_out);
+    if (!support_status.empty()) {
+      return support_status;
+    }
+    if (!mma_from_mul_sums.empty()) {
+      combiner.replaceWithMmaOp();
```
zasdfgbnm, line 368: I think `getMatmulCompileTimeRejectReason` should just say "yes I will accept the fusion" or "No, I don't want to take this fusion because blabla". It should not modify the fusion.

protonu: Yeah, I completely agree. 
But, how about the function, MatmulScheduler::canScheduleCompileTime (which invokes `getMatmulCompileTimeRejectReason`)?
We could also add a new member function to MatmulScheduler `MatmulScheduler::combineMulSum` to modify the fusion, though that may go again the mul sum being the default.

zasdfgbnm: I think `canScheduleCompileTime` should also be ready-only. Does it make sense to do it at the very beginning of `scheduleMatmul`?

protonu: I was thinking of:
1. Adding a private flag in `MatmulScheduler` called `combineMulSum` which I can set inside `MatmulScheduler::canScheduleCompileTime`. Then in `MatmulScheduler::canScheduleCompileTime` run the combination pass before calling `MatmulScheduler::canScheduleCompileTime()`.

The only problem that I can think of is, will `computeHeuristics` be called before I call `MatmulScheduler::canScheduleCompileTime()` since `computeHeuristics` searches for a Mmm Op and isn't a member function.

csrc/scheduler/matmul_utils.cpp
naoyam: Actually, it seems it's always the case? This function is called from `getMatmulCompileTimeRejectReason`, which fills a proper `props` value even when an existing `MmaOp` is used. 

https://github.com/NVIDIA/Fuser/pull/1507/files#diff-b979e55bd012809d32c2fe0641c25ba3c15b8b447deea459d80a13bfb0a1fadaR361

So, it seems we should always just use the given `props` parameter.

csrc/scheduler/matmul_utils.cpp
```diff
@@ -307,32 +309,63 @@ std::string getMatmulCompileTimeRejectReason(Fusion* fusion) {
...
+         << mma_exprs.size();
+      return ss.str();
+    }
+  }
+
+  if (!mma_exprs.empty()) {
+    mma_inputs_out = {
+        static_cast<TensorView*>(mma_exprs.front()->inA()),
+        static_cast<TensorView*>(mma_exprs.front()->inB()),
+        static_cast<TensorView*>(mma_exprs.front()->out())};
```
naoyam, line 346: nit: 
```suggestion
  mma_utils::MmaOrMulSumProperties::InputsOutputs mma_inputs_out =
    mma_exprs.empty() ? mma_from_mul_sums.front().insouts :
      {
        static_cast<TensorView*>(mma_exprs.front()->inA()),
        static_cast<TensorView*>(mma_exprs.front()->inB()),
        static_cast<TensorView*>(mma_exprs.front()->out())};
```

And replace all the `mma_exprs.empty() ? ...` below with just `mma_inputs_out`

naoyam:
It seems `getMatmulHeuristics` still assumes `MmaOp`. I don't think we should need `MmaOp` just to get matmul heuristics. IIUC, this is related to Xiang's comment https://github.com/NVIDIA/Fuser/pull/1507#discussion_r1430666929.

csrc/scheduler/matmul_utils.cpp
naoyam: The scheduler should convert a mul-sum pattern to `MmaOp` if found. It shouldn't do anything if there's already an `MmaOp` expr. I'm not sure why we would need a flag.

csrc/scheduler/matmul_utils.cpp
```diff
@@ -307,32 +309,63 @@ std::string getMatmulCompileTimeRejectReason(Fusion* fusion) {
...
-  //    scheduler
+  // scheduler In addition to the above, it may be the case that there is no mma
+  // op, but a mul-sum pair which may be replaced by a mma op. In that case we
+  // run a pass through the IR to find such a pair of mul-sum, and the
+  // properties of that pair. We then test if the mma op that would replace the
+  // mul-sum pair can be scheduled.
 
   // #1
   auto mma_exprs = ir_utils::getOpsOfType<MmaOp>(fusion);
+  // Initializing the machinery to check if there's a Mul-Sum pair
```
drzejan2, line 320: How about changing the API of the `mma_utils::CombineMulSum` in a way to hide implementation details, the fact that there is `mul-sum` pattern or `MmaOp`? In future it will make it much simpler to remote `MmaOp` from handling.

Something like this:
```
// The analysis is based on mul-sum pair pattern, detected in the provided fusion
//  definition. If detected and properties of an instance of such pattern are valid
//  then it will be later replaced with MmaOp in fusion definition.
// For the time being a direct instance of MmaOp is also accepted and handled by
//  the analysis.
 
// The plan:
// 1. check if there is exactly one mul-sum pattern in the fusion, an instance of
//     MmaOp works too
// 2. check if mul-sum inputs match any of supported inputs layout
// 3. check if fusion represents expressions that are recognized by matmul scheduler

// #1
const mma_utils::CombineMulSum combiner(fusion);

// of the record notes: isValid() will check if fusion contains mul-sum pair (1st step) or if there is a MmaOp (1nd step) 
if (!combiner.isValid()) {
  return "Failed to find mul-sum pattern in the provided fusion";
  // of the record notes: there can be `getValidityError()` which will provide more detailed explanation why fusion does not meet requirements, e.g. no pattern / no MmaOp
}
const auto mul_sum_candidates = combiner.generateMulSumCanidates();
if (mul_sum_candidates.size() != 1) {
  std::stringstream ss;
  ss  << "Matmul scheduler supports fusions with only single mul-sum pattern, got: " << mul_sum_candidates.size();
  return ss.str();
}

// #2 
{
  // of the record notes: combiner returns inputs/outputs for any supported source, be it mul-sum pattern or MmaOp
  //    this should be done during creation of mma_utils::CombineMulSum
  const auto input_layout_opt = mma_utils::getMmaLayout(fusion, mul_sum_candidates.front().insouts());
  if (!input_layout_opt.isValid()) {
    return input_layout_opt.getErrorMsg();
  }
}

// #3
{
  auto support_status = isMatmulFusionDefinitionSupported(fusion, mul_sum_candidates.front().insouts());
  if (!support_status.empty()) {
    return support_status;
  }
}
```

csrc/scheduler/mma_utils.h
```diff
@@ -239,6 +239,34 @@ class DataWrapperOpt {
   }
 };
 
+// This struct hold properties of a Mul and Sum pair
+// which can possibly be replaced a Mma op. This struct
+// can be be created (partially) from a Mma op.
+struct MmaOrMulSumProperties {
```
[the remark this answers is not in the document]
drzejan2: I'm for `MulSumProperties` name for this structure. `MatmulProperties` seems to be too broad.

csrc/scheduler/matmul_utils.cpp
```diff
@@ -137,10 +137,9 @@ inline bool initCoreHeuristics(
 
 //! A helper for getting problem shape from fusion and runtime info.
 ProblemShape getProblemShape(
-    Fusion* fusion,
-    const MmaOp* mma_expr,
+    mma_utils::MulSumProperties::InputsOutputs props,
```
naoyam, line 140: nit:
```suggestion
    const mma_utils::MulSumProperties::InputsOutputs props&,
```

csrc/scheduler/mma_utils.h
```diff
@@ -261,6 +289,9 @@ using DependenciesMap = std::map<TensorView*, DomainsDesc>;
 //!  transposition of inputs in mma instructions, while other (e.g. Turing,
 //!  Ampere) the only supported transposition is TN which means that mma
 //!  instruction first input is transposed, the second input is non-transposed.
+MatmulProblemLayoutOpt getMmaLayout(
```
naoyam, line 292: Please mention that the other `getMmaLayout` is just a wrapper of this `getMmaLayout` for convenience.

naoyam (commented):
LGTM. Thanks Protonu. I'll leave a final stamp to @drzejan2 and @jacobhinkle.

csrc/scheduler/matmul.cpp
```diff
@@ -700,18 +700,26 @@ void scheduleMatmul(Fusion* fusion, const MatmulParams& params) {
...
+  mma_utils::CombineMulSum combiner(fusion);
+  std::vector<mma_utils::MulSumProperties> mulSum =
+      combiner.generateMulSumCanidates();
 
   auto mma_ops = ir_utils::getOpsOfType<MmaOp>(fusion);
+  if (combiner.isValid() && mma_ops.empty()) {
+    combiner.replaceWithMmaOp();
+  }
+
+  mma_ops = ir_utils::getOpsOfType<MmaOp>(fusion);
```
drzejan2, line 712: With the proposed changes in public API of `CombineMulSum` there won't be a need for `mul_sum` object.

```suggestion
  mma_utils::CombineMulSum combiner(fusion);

  auto mma_ops = ir_utils::getOpsOfType<MmaOp>(fusion);
  if (mma_ops.empty()) {
    // NOTE: the combiner object has been already validated during
    //  previous checks performed on the current fusion
    combiner.replaceWithMmaOp();
  }

  mma_ops = ir_utils::getOpsOfType<MmaOp>(fusion);
```

drzejan2 (commented):
I checked proposed changes while I focused on `CombineMulSum` tool, and I think we could improve its API / usage. @protonu please check my comments, if you think these are overkill then let me know.

Functional-wise, changes look ok to me.

csrc/scheduler/mma_utils.h
```diff
@@ -345,13 +355,20 @@ class CombineMulSum : public IterVisitor {
...
       bool use_cached_results = false);
 
   //! Replaces the candidate mul-sum pairs with mma ops.
   //! Please not this will run generateMulSumCandidates again.
   void replaceWithMmaOp();
 
+  //! Check if the fusion has a mul-sum pair
+  //! than can be replaced by a mma op, or has a single mma op.
+  //! This should be run after generaateMulSumCandidates is run.
+  bool isValid() {
```
jacobhinkle, line 368: A more descriptive name might be helpful here since "valid" could be interpreted as "this is a valid input for which we can combine mul-sum to produce an MmaOp", which is not true if there is already a single MmaOp.

test/test_combine_mul_sum.cpp
```diff
@@ -187,4 +187,37 @@ TEST_F(CombineMulSumAsMmaTest, AmpereMulSumToMatmul_Schedule) {
...
+  ASSERT_TRUE(ir_utils::getOpsOfType<MmaOp>(fusion.get()).empty());
+
+  auto t0 = matmulAtInput(layout, TensorMatmulPos::A, at::kHalf, M, N, K);
+  auto t1 = matmulAtInput(layout, TensorMatmulPos::B, at::kHalf, M, N, K);
+  auto tref = atMatmul(t0, t1, layout);
+
+  FusionExecutorCache executor_cache(std::move(fusion));
+  auto outputs = executor_cache.runFusionWithInputs({t0, t1});
+
+  NVF_CHECK(
```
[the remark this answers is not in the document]
jacobhinkle: I think this is not actually resolved.  You can inspect this with `executor_cache.getMostRecentKernelRuntime()->schedulerHeuristics()->heuristicsList()`. The first entry there will be a `std::unique_ptr<SchedulerEntry>` that you can try to `dynamic_cast` to `MatmulScheduler`.

test/test_combine_mul_sum.cpp
```diff
@@ -187,4 +187,44 @@ TEST_F(CombineMulSumAsMmaTest, AmpereMulSumToMatmul_Schedule) {
...
+  auto outputs = executor_cache.runFusionWithInputs({t0, t1});
+  // Ensure there's a mma op.
+  ASSERT_FALSE(
+      ir_utils::getOpsOfType<MmaOp>(executor_cache.getMostRecentKernelRuntime()
+                                        ->executors()
+                                        .at(0)
+                                        .kernel())
+          .empty());
+
+  NVF_CHECK(
```
jacobhinkle, line 222: nit: use `EXPECT_FALSE` for these two checks.

test/test_combine_mul_sum.cpp
```diff
@@ -187,4 +187,44 @@ TEST_F(CombineMulSumAsMmaTest, AmpereMulSumToMatmul_Schedule) {
   }
 }
 
+TEST_F(CombineMulSumAsMmaTest, UseMatmulScheduler) {
+  // Keep multiples of 8 to keep vectorizable.
+  int M = 504, N = 136, K = 248;
+  const auto layout = MmaLayout::TT;
```
jacobhinkle, line 193: Is there any value in looping over `kAllSupportedMmaLayout` in order to test the pattern matching in presence of various patterns?

jacobhinkle (approved):
LGTM. Just some minor comments.

csrc/scheduler/matmul_utils.cpp
```diff
@@ -302,37 +303,49 @@ std::string getMatmulRunTimeRejectReason(
   return "";
 }
 
+// The analysis is based on mul-sum pair pattern, detected in the provided
+// fusion definition. If detected and properties of an instance of such
+// pattern are valid then it will be later replaced with MmaOp in
+// fusion definition. For the time being a direct instance of MmaOp is
+// also accepted and handled by the analysis.
```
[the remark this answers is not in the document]
protonu: Thanks, will modify.
Is there any (historical) reason for this? Can we not just single spacing and add empty lines when require?

csrc/scheduler/mma_utils.h
```diff
@@ -345,13 +355,20 @@ class CombineMulSum : public IterVisitor {
   //! Goes through the fusion to find mul-sum pairs.
   //! If user sets the caching flags and properties have been previously
   //! computed, then just return cached results.
-  std::vector<MulSumAsMmaProps> generateMulSumCanidates(
+  std::vector<MulSumProperties> generateMulSumCanidates(
```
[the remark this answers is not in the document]
protonu: This makes a lot of sense. I made these modifications. Let me know if you want further changes.

csrc/scheduler/matmul.cpp
protonu: Thanks!
I made most of those changes. 
In addition, I used:

```
  auto mma_ops = ir_utils::getOpsOfType<MmaOp>(fusion);
  if (combiner.isValid() && mma_ops.empty()) {
    combiner.replaceWithMmaOp();
    mma_ops = ir_utils::getOpsOfType<MmaOp>(fusion);
  }
```

I kept the `isValid` check to be safe, since I think function can be used directly.
I also moved the (second) getOps to inside the if-condidion.

test/test_combine_mul_sum.cpp
jacobhinkle: `EXPECT_FALSE` in particular, as opposed to `ASSERT_FALSE` or `NVF_CHECK`, will allow the test to proceed, but will give an error message and cause the test to fail after it runs to completion. This is more informative for conditions that are not guarding downstream checks. For example, you'd use `ASSERT_*` to check the size of a vector if you are going to extract an element from that vector later. Since nothing else depends on this condition, I think it's better as `EXPECT`. Note that of course this is not very important since it doesn't affect whether or not the test fails :-D

test/test_combine_mul_sum.cpp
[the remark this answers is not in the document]
jacobhinkle: I believe this is still unresolved.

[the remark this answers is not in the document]
protonu: In this case I was checking if there's a mma op. If the fusion got scheduled, and there's a mma op, can I not infer that the matmul scheduler was invoked?

test/test_combine_mul_sum.cpp
protonu: Changed it to `EXPECT_FALSE` -- but I assert when checking for a mma op.

test/test_combine_mul_sum.cpp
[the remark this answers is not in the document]
protonu: I check for both now -- the op and the scheduling pass.

test/test_combine_mul_sum.cpp
```diff
@@ -187,4 +187,51 @@ TEST_F(CombineMulSumAsMmaTest, AmpereMulSumToMatmul_Schedule) {
...
+                         .kernel())
+                     .empty());
+    // Ensure that the matmul scheduler ran.
+    EXPECT_TRUE(
+        dynamic_cast<MatmulScheduler*>(
+            executor_cache.getMostRecentKernelRuntime()
+                ->schedulerHeuristics()
+                ->heuristicsList()
+                .at(0)
+                .get()) != nullptr);
```
jacobhinkle, line 228: Nice!

protonu:
!build

csrc/scheduler/matmul_utils.cpp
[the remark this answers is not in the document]
drzejan2: This formatting has been around when I joined the project, so, unfortunately, I don't know the reason.
I simply followed this for the changes I have prepared.

csrc/scheduler/mma_utils.h
```diff
@@ -271,10 +305,15 @@ MatmulProblemLayoutOpt getMmaLayout(Fusion* fusion);
 //!  be gathered.
 //!  TODO: 4th domain must be added for batch gemm support.
 ProblemIterDomainsOpt getProblemIterDomains(Fusion* fusion);
+ProblemIterDomainsOpt getProblemIterDomains(
+    mma_utils::MulSumProperties::InputsOutputs props);
```
drzejan2, line 309: Nitpick:

`mma_utils::MulSumProperties::InputsOutputs` is a simple structure but perhaps there is no need to make a copy of it when this helper function is called?

csrc/scheduler/mma_utils.h
```diff
@@ -271,10 +305,15 @@ MatmulProblemLayoutOpt getMmaLayout(Fusion* fusion);
...
 
 //! Returns wrapped collection of TensorView roles in fusion.
 //!  An error message is stored in retruned object if valid data cannot
 //!  be gathered.
+RolesMapOpt getTensorsRoles(
+    Fusion* fusion,
+    mma_utils::MulSumProperties::InputsOutputs props);
```
drzejan2, line 316: Nitpick:

`mma_utils::MulSumProperties::InputsOutputs` is a simple structure but perhaps there is no need to make a copy of it when this helper function is called?

drzejan2 (approved):
I left a minor comments about the API (possibly unnecessary copy of passed data).

Otherwise, changes look good. Thanks!

csrc/scheduler/matmul_utils.cpp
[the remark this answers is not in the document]
naoyam: Just FYI, I'm not aware of this convention in the overall codebase.

protonu:
!build

PR #1509, opened 2023-12-13
Handle duplicated outputs using ExpressionEvaluator.

wujingyue:
I also tried using output-to-output aliases to handle this. It didn't work well because `getOutputAlias` couldn't distinguish the same `Val*` at different indices. To allocate a tensor for the first output but not the second, I would still end up with something like `outputs_map`.

This PR removes the ad-hoc `outputs_map` in favor of `ExpressionEvaluator`. It also moves trivial-input-forwarding tests to `test_alias.cpp` to make `test_gpu3.cpp` (slightly) smaller.

### DIFF
--- a/csrc/executor.cpp
+++ b/csrc/executor.cpp
@@ -921,6 +921,9 @@ at::Tensor allocateOutput(
     const c10::Device& device,
     ExpressionEvaluator& ee) {
   TensorView* out_tv = out_info.tv;
+  if (ee.isKnown(out_tv)) {
+    return ee.evaluate(out_tv).as<at::Tensor>();
+  }
 
   if (aliased_in != nullptr) {
     const PolymorphicValue& aliased_in_val = ee.evaluate(aliased_in);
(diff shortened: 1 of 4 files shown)


wujingyue:
!build

zasdfgbnm (approved):

csrc/executor.cpp
```diff
@@ -981,22 +985,14 @@ std::vector<at::Tensor> allocateOutputs(
...
-    if (iter == outputs_map.end()) {
-      auto [aliased_in, alias_info] = kernel->getOutputAlias(out);
-      auto output = allocateOutput(
-          output_info[output_idx], aliased_in, alias_info, device, ee);
-      outputs_map[out] = output;
-      outputs.push_back(output);
-    } else {
-      outputs.push_back(iter->second);
-    }
+    auto [aliased_in, alias_info] = kernel->getOutputAlias(out);
```
jjsjann123, line 990: Thanks for the clean up.

jjsjann123 (approved):

PR #1510, opened 2023-12-13
Enable serial grid reduction for split-K in matmul scheduler

jacobhinkle:
Stacked on #1456.

This simply enables serial reduction in the matmul scheduler when split-K is used.


csrc/ops/arith.cpp
```diff
@@ -2542,7 +2542,7 @@ TensorView* fusedMultiplySum(
     const std::vector<int>& axes,
     Val* init) {
   if (init == nullptr) {
-    init = IrBuilder::create<Val>(0.0);
+    init = IrBuilder::create<Val>(0.0, DataType::Float);
```
jacobhinkle, line 2545: Using incorrect dtype `Double` instead of `Float` caused the `init` value in the reduction to be printed as double, leading to a type error in compilation. This `init` variable seems to only be used by rfactor, so maybe we should just set it (properly) there instead of holding it as an `MmaOp` attribute. Would we ever use a non-zero `init` in an `MmaOp`?

csrc/scheduler/mma_utils.cpp
```diff
@@ -1276,8 +1276,8 @@ RolesMapOpt getTensorsRoles(Fusion* fusion) {
 namespace {
 
 void addMMAOp(Fusion* fusion_, std::vector<MulSumAsMmaProps>& props) {
-  auto* init = IrBuilder::create<Val>(0.0);
   for (auto prop : props) {
+    auto* init = IrBuilder::create<Val>(0.0, prop.out->getDataType().value());
```
jacobhinkle, line 1280: See above comment about init dtype.

jacobhinkle:
I had to change some surrounding code to fix the following error, which was caused by using a `Double` scalar for the `init` value of the reduction in the rfactored `MmaOp`:
```c++
  // Allocate global tensor T13
  grid_sync::blockSerializeWait<false, false, true>(&T13[index_utils::maskedOffset<true, true, false>(blockIdx, gridDim)]);
  #pragma unroll
  for(nvfuser_index_t i93 = 0; i93 < 4LL; ++i93) {
    nvfuser_index_t i94;
    i94 = 32LL * i93;
    nvfuser_index_t i95;
    i95 = i35 + (i30 * i93);
    nvfuser_index_t i96;
    i96 = -(16LL * i93);
    #pragma unroll
    for(nvfuser_index_t i97 = 0; i97 < 8LL; ++i97) {
      nvfuser_index_t i98;
      i98 = i94 + (4LL * i97);
      nvfuser_index_t i99;
      i99 = i95 + (8LL * i97);
      bool b100;
      b100 = i41 < (-(8LL * (i97 + nvfuser_zero)));
      bool b101;
      b101 = b43 && b100;
      #pragma unroll
      for(nvfuser_index_t i102 = 0; i102 < 2LL; ++i102) {
        nvfuser_index_t i103;
        i103 = i98 + (2LL * i102);
        nvfuser_index_t i104;
        i104 = i99 + (i36 * i102);
        nvfuser_index_t i105;
        i105 = i102 + nvfuser_zero;
        bool b106;
        b106 = i42 < (i96 - (8LL * i105));
        bool b107;
        b107 = b100 && b106;
        Array<float, 2LL, 2> T7;
        #pragma unroll
        for(nvfuser_index_t i108 = 0; i108 < 2LL; ++i108) {
          T7[i108] = 0.00000000000000000e+00;
        }
        #pragma unroll
        for(nvfuser_index_t i108 = 0; i108 < 2LL; ++i108) {
          // Allocate global tensor T14
          reduction::serialReductionStep(
            T7[i108],
            T12[(i103 + i108)],
            0.00000000000000000e+00,
            T14[(i104 + (i108 + nvfuser_zero))],
            b107,
            b107);
        }
        if ((b101 && b106)) {
          loadLocalToGlobal<float, /*vec_size=*/2, /*is_volatile=*/false>( &T6[(i99 + (i36 * i105))], &T7[0LL]);
        }
      }
    }
  }
  NVFUSER_UPDATE_MAGIC_ZERO;
  grid_sync::blockSerializeRelease<false, false, true>(&T13[index_utils::maskedOffset<true, true, false>(blockIdx, gridDim)]);
}
}
/*
CUDA NVRTC compile error: __tmp_kernel_none_f0_c0_r0_g0.cu(9791): error: no instance of function template "<unnamed>::reduction::serialReductionStep" matches the argument list
            argument types are: (float, float, double, float, lambda [](float &, float)->void, __nv_bool, __nv_bool, __nv_bool, __nv_bool)
            reduction::serialReductionStep(
            ^

1 error detected in the compilation of "__tmp_kernel_none_f0_c0_r0_g0.cu".
*/
```

csrc/ops/arith.cpp
naoyam: Can you show the code that caused the compilation error?

jacobhinkle: Sure!
```c++
          reduction::serialReductionStep(                                                                                                              
            T6[i115],                                                                                                                                  
            T12[(i110 + i115)],                                                                                                                        
            0.00000000000000000e+00,                                                                                                                   
            T14[(i111 + (i115 + nvfuser_zero))],                                                                                                       
            b114,                                                                                                                                      
            b114);                                                                                                                                     
        }                                                                                                                                              
```                                                               
The error is
```
CUDA NVRTC compile error: __tmp_kernel_none_f0_c0_r0_g0.cu(9806): error: no instance of function template "<unnamed>::reduction::serialReductionStep" matches the argument list
            argument types are: (float, float, double, float, lambda [](float &, float)->void, __nv_bool, __nv_bool, __nv_bool, __nv_bool)
            reduction::serialReductionStep(
```

jacobhinkle: Changing init to `Float` makes it print properly:
```c++
          reduction::serialReductionStep(
            T6[i115],
            T12[(i110 + i115)],
            0.000000000e+00f,
            T14[(i111 + (i115 + nvfuser_zero))],
            b114,
            b114);
```

naoyam: Looks like it's a template matching issue. Doesn't that mean the type just needs to be the same as the other parameters? That's most likely `float`, but theoretically could be other FP types as well, right?

jacobhinkle: Oh good point, maybe I could cast `grop->init()` to `grop->in()->dtype()` instead possibly in `codegen.cpp`.

naoyam: Why not just using the dtype of `tv_a`?

jacobhinkle: Ah, but `init` needs to be in the accumulator type. The inputs might be half precision.

jacobhinkle: This function (`fusedMultiplySum`) calls `newForMma` to set up the output, so it is currently always defaulting to `Float`. It would be better to set up `init` after we get `out`, so that we can get the output type from there.

jacobhinkle: I just pushed a change that uses `out->dtype()` as the dtype for `init`.

naoyam (approved):
LGTM. Congrats on enabling split-K!

PR #1511, opened 2023-12-13
Fix mapping propagation through backward merge

naoyam:
While working on comparing the Permissive graph with the CA Permissive map, I found one thing that I think should be considered inconsistency. When backward propagating mappings through a Merge expr, we require at least one of the input domain pairs should be already mapped or have equal extents, which makes sense, but the current code only checks the extents of one ID out of each ID group, which is fine with the Exact graph but in the case of the Permissive graph, each ID group may consist of domains that have different extents, so I think we would also need to check all domains in the group, and if there's one matching pair, we should allow propagation. 

This PR also adds the validation of the Permissive graph. It does not pass yet as the CA map does not map the compliment IDs.

### DIFF
--- a/csrc/id_model/id_model.cpp
+++ b/csrc/id_model/id_model.cpp
@@ -1072,6 +1072,10 @@ void IdModel::build(
   }
 
   buildPermissiveMap(tv_exprs);
+  if (validate) {
+    validator->checkPermissiveGraphEquivalence(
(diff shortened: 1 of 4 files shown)


naoyam:
!build

naoyam:
!build

naoyam:
The error with Indexing19 is known.

PR #1513, opened 2023-12-13
Clean up the shift tests

naoyam:


xwang233 (approved):
LGTM. If the build works, it should be fine.

PR #1516, opened 2023-12-13
Hopper MMA SS B swizzle

zasdfgbnm:


zasdfgbnm:
!build

test/test_mma.cpp
```diff
@@ -501,15 +489,7 @@ TEST_P(HopperSS, SingleTile) {
   // Hopper tensor core assumes K major, so we are using !transpose_a here.
   tv0b->applyMmaSwizzle(swizzle_a, !transpose_a);
   tv1b->setMemoryType(MemoryType::Shared);
-  tv1b->applyMmaSwizzle(swizzle_b, transpose_b);
-
-  if (transpose_a) {
```
zasdfgbnm, line 506: moved to `applyMmaSwizzle`

csrc/device_lower/pass/index.cpp
```diff
@@ -1451,17 +1451,18 @@ static Val* constructMatrixDescriptor(
...
-      8;
-  if (num_core_matrices == 1) {
-    return MmaInputSmemSwizzle::None;
+  const auto& alloc_domain = tv->getRootDomain();
+  const auto& leaf_domain = tv->getLeafDomain();
+  auto exprs = StmtSort::getExprsBetween(
+      {alloc_domain.begin(), alloc_domain.end()},
+      {leaf_domain.begin(), leaf_domain.end()});
+  for (auto expr : exprs) {
+    if (auto swizzle = dynamic_cast<const Swizzle*>(expr)) {
```
drzejan2, line 1460: Do we plan to handle anyhow situations where by mistake there are two swizzle ops applied on one of mma inputs?

Currently we search for first instance, and return its swizzle, while we could put a strong requirement to have a single instance in the expr set. Something like this:

```
std::vector<const Swizzle*> swizzle_exprs = [](){
  std::vector<const Swizzle*> results;
  for (auto expr: exprs) {
    if (auto swizzle = dynamic_cast<const Swizzle*>(expr)) {
      results.push_back(swizzle);
    }
  }
  return results;
}();
NVF_ERROR(swizzle_exprs.size() < 2, "expected 2 or less swizzle expressions in mma input, got ", swizzle_exprs.size());

if (swizzle_exprs.empty()) {
  swizzle_exprs.front()->inX()->extent()->evaluate().as<int64_t>() * 16);
}
else {
  return MmaInputSmemSwizzle::None;
}
```

The next step could be checking if there are more than a single swizzle expr but all of them have the same type of swizzle (effectively may mean that second and following exprs could be no-ops)

csrc/device_lower/pass/index.cpp
```diff
@@ -1506,7 +1507,8 @@ void IndexLowering::handle(const MmaOp* mma) {
         roundUpToMultiple(getN(mma->macro()) / 8L,
                           getBytesFromSwizzle(swizzle) / 16L) *
         /*bytes per item*/ 2L;
-    if (swizzle != MmaInputSmemSwizzle::None) {
+    if (swizzle != MmaInputSmemSwizzle::None &&
+        (mma->layout() == MmaLayout::TT || mma->layout() == MmaLayout::TN)) {
```
drzejan2, line 1511: Do you have some theory why this change is needed? To be honest, I'm lost here.

We handle B tensor here, but it looks like A tensor layout impacts the processing 🤔

drzejan2 (approved):

zasdfgbnm:
!build

csrc/device_lower/pass/index.cpp
zasdfgbnm: Added

csrc/device_lower/pass/index.cpp
zasdfgbnm: Yes, this is super strange. I don't have a theory :(

zasdfgbnm:
!build

zasdfgbnm:
!build

zasdfgbnm:
!build

csrc/device_lower/pass/index.cpp
drzejan2: Your implementation looks better than the suggestion - I learnt something new. Thanks!

drzejan2:
Checked updated sources - looks good to me.

Thanks!

PR #1517, opened 2023-12-13
Add `tests` target to CMake so that all the tests can be built as `ninja tests`

naoyam:


zasdfgbnm (approved):
Thanks for adding this!

PR #1528, opened 2023-12-14
Vectorized serial grid reduction

jacobhinkle:
This change allows us to use vectorized loads/stores in `serialReductionStep`. The generated kernel now looks like
```c++
  NVFUSER_UPDATE_MAGIC_ZERO;                                        
  grid_sync::blockSerializeWait<false, false, true>(&T5[index_utils::maskedOffset<true, true, false>(blockIdx, gridDim)]);
  #pragma unroll                                                                                                                         
  for(nvfuser_index_t i16 = 0; i16 < 4LL; ++i16) {                                                                                           nvfuser_index_t i17;                                                                                                                 
    i17 = 32LL * i16;                                                                                                                        nvfuser_index_t i18;                                                                                                                 
    i18 = 4096LL * i16;                                                                                                                  
    nvfuser_index_t i19;                                                                                                                 
    i19 = i5 + i18;                                                                                                                      
    nvfuser_index_t i20;                                                                                                                 
    i20 = -i18;                                                                                                                          
    #pragma unroll                                                                                                                       
    for(nvfuser_index_t i21 = 0; i21 < 8LL; ++i21) {                                                                                     
      nvfuser_index_t i22;                                                                                                               
      i22 = 512LL * (i21 + nvfuser_zero);                                                                                                
      Array<float, 4LL, 4> T3;                                                                                                           
      T3.set(float(0.000000000e+00f));                                                                                                   
      reduction::serialReductionStep</*vec_size=*/4>(                                                                                    
        &T3[0LL],                                                                                                                        
        &T2[(i17 + (4LL * i21))],                                                                                                        
        0.000000000e+00f,                                                                                                                
        &T6[(i19 + i22)],                                                                                                                
        true,                                                                                                                                    true);                                                                                                                           
      if ((b7 && (i6 < (i20 - i22)))) {                                                                                                  
        loadLocalToGlobal<float, /*vec_size=*/4, /*is_volatile=*/false>( &T1[(i19 + i22)], &T3[0LL]);                                    
      }                                                                                                                                  
    }                                                                                                                                    
  }                                                                                                                                      
  grid_sync::blockSerializeRelease<false, false, true>(&T5[index_utils::maskedOffset<true, true, false>(blockIdx, gridDim)]);            
  NVFUSER_UPDATE_MAGIC_ZERO;       
```


test/test_serial_gridreduce.cpp
```diff
@@ -40,6 +40,7 @@ using SerialGridReductionTest = NVFuserTest;
 TEST_F(SerialGridReductionTest, CodegenNodes) {
   for (bool serial : {true, false}) {
     for (int64_t num_warps : {4, 8}) {
+    for (int64_t vec_size : {4, 2, 1}) {
```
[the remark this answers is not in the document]
jacobhinkle: Fixed as of 0ef6204

test/test_serial_gridreduce.cpp
```diff
@@ -75,21 +76,23 @@ TEST_F(SerialGridReductionTest, CodegenNodes) {
...
+        tv3->split(0, num_warps); // blocks_x*blocks_y*A*B, num_warps, 32, vec_size, H
+        tv3->split(0, B); // blocks_x*blocks_y*A, B, num_warps, 32, vec_size, H
+        tv3->split(0, A); // blocks_x*blocks_y, A, B, num_warps, 32, vec_size, H
+        tv3->split(0, blocks_y); // blocks_x, blocks_y, A, B, num_warps, 32, vec_size, H
         tv3->axis(0)->parallelize(ParallelType::BIDx);
         tv3->axis(1)->parallelize(ParallelType::BIDy);
         tv3->axis(4)->parallelize(ParallelType::TIDy);
         tv3->axis(5)->parallelize(ParallelType::TIDx);
-        tv3->axis(6)->parallelize(ParallelType::BIDz);
+        tv3->axis(6)->parallelize(ParallelType::Vectorize);
```
jacobhinkle, line 94: `ParallelType::Group` might be a better choice here to signal vectorized serial reduction. I'll experiment with that. Originally I worried that lowering would detect it and convert the node to a `GroupedGridReduction` but  I think that might be the proper way to handle it actually.

csrc/device_lower/validation.cpp
```diff
@@ -579,9 +579,10 @@ void validateAndCollectVectorizeInfo(Fusion* fusion) {
       }
     }
     if (has_vectorize_dim) {
+      Expr* def = tv->definition();
       NVF_ERROR(
-          tv->definition() == nullptr || tv->definition()->isA<LoadStoreOp>() ||
-              tv->definition()->isA<SliceOp>(),
+          def == nullptr || def->isA<LoadStoreOp>() ||
+              def->isA<SliceOp>() || (def->isA<ReductionOp>()), // && def->as<ReductionOp>()->SerialGridReductionRequested(),
```
naoyam, line 585: Is the `&&` part supposed to be commented out?

runtime/grid_reduction.cu
```diff
@@ -646,12 +646,24 @@ __device__ void serialReductionStep(
   if (!write_pred) {
     return;
   }
-  out = read_pred ? in : init;
+  if (read_pred) {
+    loadGeneric<T, vec_size>(out, in);
```
naoyam, line 650: Out of curiosity, does it matter to use `loadGeneric` here? Both `out` and `in` are registers, so I'd assume just a normal for-loop would be the same.

naoyam (approved):
I think this is fine but also wonder if we could represent the serial grid reduction code with KIR expressions. The serial reduction part looks like just a normal reduction with writes to a global buffer. We would need a special sync IR node after that, but it seems possible to represent the computation part in KIR. If that's the case, we would just need to vectorize the loads and stores as we would do normally.

naoyam:
Oh, maybe I should have reviewed #1456 first. Will do.

csrc/device_lower/validation.cpp
jacobhinkle: That needs to be uncommented once #1456 is merged, since that introduces the method `SerialGridReductionRequested()`. For now it is just allowing reductions.

I am starting to doubt using vectorize for this though. It would be cleaner to use `Group` for the vectorized reduction output. Then as you say, if I lower to kernel IR instead of using a helper function, we could convert that to vectorized LoadStoreOps with `CacheOp::Global` then the default validation checks would work out I think. I will mark this draft while I give that a shot.

naoyam: > lower to kernel IR instead of using a helper function

Just want to make sure that may be desirable but not necessary. Prototyping is typically easier with helper runtime functions than generating in KIR>

runtime/grid_reduction.cu
jacobhinkle: Yeah for loop would work fine too.

csrc/device_lower/validation.cpp
jacobhinkle: I think it is desirable, since we'd get a lot of stuff for free like vectorization, inserting allocations, and hoisting predicates. I've been trying a bit to get it working as a single lowering pass but generating IR for the work buffer is challenging because it appears we need to introduce a cycle into the graph; we load from the work buffer then write back to it. We could try making the input and output global work buffers into separate `TensorView`s but we'd then need to figure out how to force those to be aliased.

test/test_serial_gridreduce.cpp
jacobhinkle: I am actually going to keep this as `Vectorize` since there is another good use for `Group` in this context. When we use split-K in a **bmm**, the batch dimension is no longer parallelized as it would be without split-K. In this case we have an outer loop in the batch dim that includes a separate main loop for each row in the batch dimension. Our current approach of syncing around _all_ non-trivial loops for serial reduction is insufficient in that case since it means we are actually serializing the entire kernel instead of just each output stage. We could explicitly mark the inner dimensions as `ParallelType::Group` in order to signal that we actually want to sync inside that outer loop.

jacobhinkle:
!build

naoyam (approved):
LGTM

jacobhinkle:
This is the speedup in `nvfuser_bench` benchmarks when comparing this PR with main, after merging #1719 into each:
image
These benchmarks include split-K factors of 2 thru 6. I missed this speedup previously by looking at just a couple of problem sizes, and also I was doing HSS instead of HSH. So even without unswizzling (#1534) or reduced precision reduction (#1719), this PR seems to have a pretty big impact. Unswizzling will let us use a vectorization size of 4 instead of 2, and reduced precision will cut the amount of split-K IO in half, so I expect some further speedups.

jacobhinkle:
!build --diff

jacobhinkle:
Python test failure is just a rare event due to low tolerance.

Batch matmul splitk tests show increase in register usage with this change vs. no change for non-batch matmul tests (see `nvfuser-ci/jit_codegen_diff_17_4/4`).

PR #1530, opened 2023-12-15
avoid self projection when project to broadcast tvs

liqiangxl:
Redo #1408 with fix in #1533

### DIFF
--- a/csrc/scheduler/reduction_utils.cpp
+++ b/csrc/scheduler/reduction_utils.cpp
@@ -906,8 +906,12 @@ class PersistentBufferProjector {
       // `T2->T1->T7->T6->T4->T5` which has all root domain information.
       // See FusionBroadcastPersistentReduction_CUDA for an example
       dummy_outputs_.emplace_back(add(buffer_replicate, buffer));
-      ir_utils::replaceValInExprInputs(
-          use->definition(), buffer, buffer_replicate);
+      // avoid replacing the use with itself, see
+      // https://github.com/NVIDIA/Fuser/issues/1533
+      if (buffer != buffer_replicate) {
+        ir_utils::replaceValInExprInputs(
+            use->definition(), buffer, buffer_replicate);
+      }
     }
   }
 };
--- a/csrc/scheduler/reduction_utils.cpp
+++ b/csrc/scheduler/reduction_utils.cpp
@@ -740,7 +740,6 @@ class PersistentBufferProjector {
   const bool project_to_inputs_;
 
   void projectToInputs() {
-    fusion_->printMath();
     // Iterate through projected buffers, tracking which index it corresponds
     // too since there's a resolution point entry for every buffer.
     const auto& reduction_tvs = scheduler_utils::getReductionTvs(fusion_);
--- a/csrc/scheduler/reduction_utils.cpp
+++ b/csrc/scheduler/reduction_utils.cpp
@@ -772,7 +772,6 @@ class PersistentBufferProjector {
       // reduciton and broadcast from input t0 to t3. The broadcast here is not
       // just a local register copy but involves an inter-thread communication.
       std::vector<Val*> vals_project_to = fusion_->inputs();
-<<<<<<< HEAD
       const auto& [can_project, broadcast_tvs] =
           scheduler_utils::canProjectToInputsWithoutReduction(
               reduction_tvs, buffer);
@@ -781,14 +780,6 @@ class PersistentBufferProjector {
             vals_project_to.end(), broadcast_tvs.begin(), broadcast_tvs.end());
       }
 
-=======
-      const auto& dep_vals = DependencyCheck::getAllValsBetween(
-          {reduction_tvs.begin(), reduction_tvs.end()}, {buffer});
-      const auto& broadcast_tvs =
-          scheduler_utils::getBufferProjectableBroadcastsTvs(dep_vals, buffer);
-      vals_project_to.insert(
-          vals_project_to.end(), broadcast_tvs.begin(), broadcast_tvs.end());
->>>>>>> eb41fcb53970c4e825f403af5237ff505a0d7f63
       projectToInputOrImmediatePersistentProducer(
           (int)buffer_i, vals_project_to);
     }
--- a/csrc/scheduler/reduction_utils.cpp
+++ b/csrc/scheduler/reduction_utils.cpp
@@ -721,17 +721,11 @@ class PersistentBufferProjector {
         project_to_inputs_(project_to_inputs) {}
 
   const std::vector<TensorView*>& project() {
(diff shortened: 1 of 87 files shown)


liqiangxl:
!build --diff

liqiangxl:
!build --diff

csrc/scheduler/utils.cpp
```diff
@@ -473,12 +473,38 @@ class PersistentBufferResolution : public IterVisitor {
...
+      if (!tv->definition()->isA<BroadcastOp>() || tv == persistent_buffer) {
+        continue;
+      }
+      const auto& producers = ir_utils::producerTvsOf(tv);
+      NVF_ERROR(
+          producers.size() == 1,
+          "Tv defined by BroadcastOp must have one produer! Got: ",
+          producers.size());
+      auto producer = producers.at(0);
+      if (!producer->definition()->isA<ReductionOp>()) {
```
naoyam, line 494: The scheduler already knows which tensors are the reduction outputs of the normalization pattern. Shouldn't we check this producer is one of them?

csrc/scheduler/utils.cpp
```diff
@@ -473,12 +473,38 @@ class PersistentBufferResolution : public IterVisitor {
...
+        continue;
+      }
+      // Each reduction dimension the producer, must be mapped to a broadcast
+      // dimension in the consumer, otherwise it is not a valid broadcast after
+      // reduction.
+      bool is_broadcast_after_reduction = true;
+      for (auto i : c10::irange(producer->nDims())) {
+        if (producer->axis(i)->isReduction() && !tv->axis(i)->isBroadcast()) {
```
naoyam: I don't think this is strictly correct as the number of dimensions may be different. Not common, but the normalization scheduler may accept a pattern like:

```
tv1 = sum(tv0, {0, 1});
tv2 = broadcast(tv1, {true});
tv3 = broadcast(tv2, {true, false});
```

csrc/scheduler/utils.cpp
```diff
@@ -558,10 +584,13 @@ PersistentBufferInfo persistentBuffers(Fusion* fusion) {
     // If the buffer doesn't depends on reduction tvs or there is a broadcasted
     // tv between reduction and persistent buffer. The re-calculation of the
     // persistent buffer doesn't need additional reductions and this buffer is
-    // projecatable.
+    // projecatable. When searching for broadcasted tvs, exclude the buffer
+    // itself.
     const auto& dep_vals = DependencyCheck::getAllValsBetween(
```
naoyam, line 590: nit: Move this into `getBufferProjectableBroadcastsTvs`

csrc/scheduler/utils.h
```diff
@@ -197,8 +197,13 @@ bool canProjectToPersistentProducer(
     const std::vector<TensorView*>& producers,
     const std::unordered_set<TensorView*>& persistent_buffer_set);
 
-// Returns all broadcast tvs.
-std::vector<TensorView*> getBroadcastTvs(const std::vector<Val*>& tvs);
+// Returns all broadcast tvs the persistent buffer tv can be projected to.
```
naoyam, line 200: Please give more information rather than just what it does. Why is this done? Why broadcast tensors?

naoyam (commented):
I'm not sure why self projection needs to be avoided. If the resulting size is smaller, then why not? Isn't the real issue is how the affected path of exprs are modified?

csrc/scheduler/utils.cpp
liqiangxl: Changed the algorithm a little bit. For each buffer, will get all the reduction tvs it depends on, then for each reduction tv, get the broadcast tv which broadcasts the reduction dims. So, if the buffer depends on `N reduction tvs`, there must be `N broadcast tvs`, otherwise, not allowed to project.

csrc/scheduler/utils.cpp
liqiangxl: Good point. IN this case, we should project to `tv3`.  The revised algorithm solves this problem and a test case is added.

csrc/scheduler/utils.cpp
liqiangxl: changed.

csrc/scheduler/utils.h
liqiangxl: added more comments.

liqiangxl:
> I'm not sure why self projection needs to be avoided. If the resulting size is smaller, then why not? Isn't the real issue is how the affected path of exprs are modified?

Will try to allow self projection in a following commit.

liqiangxl:
> > I'm not sure why self projection needs to be avoided. If the resulting size is smaller, then why not? Isn't the real issue is how the affected path of exprs are modified?
> 
> Will try to allow self projection in a following commit.

Made self projection to a no-op.
```
      if (buffer != buffer_replicate) {
        ir_utils::replaceValInExprInputs(
            use->definition(), buffer, buffer_replicate);
      }
```
With the fix in this PR, didn't find a test to trigger this self projection. But it was tested using main branch before #1408 was reverted and the failed script in #1533 runs without error.

liqiangxl:
!build --diff

csrc/scheduler/reduction_utils.cpp
```diff
@@ -876,8 +900,12 @@ class PersistentBufferProjector {
       // `T2->T1->T7->T6->T4->T5` which has all root domain information.
       // See FusionBroadcastPersistentReduction_CUDA for an example
       dummy_outputs_.emplace_back(add(buffer_replicate, buffer));
-      ir_utils::replaceValInExprInputs(
-          use->definition(), buffer, buffer_replicate);
+      // avoid replacing the use with itself, see
+      // https://github.com/NVIDIA/Fuser/issues/1533
+      if (buffer != buffer_replicate) {
```
naoyam, line 905: Is this the right place to handle this uncommon case? `RecomputeTv::recompute` returns the input buffer, which means nothing is recomputed? At least in that case it shouldn't be necessary to add the dummy output, should it?

liqiangxl: (1) Yes. The direct cause is self-replacement, here it is directly avoided. We can also handle this in other places e.g. (a) before `auto buffer_replicate = RecomputeTv::recompute(buffer, producers);`, we make sure `buffer` is not in `producers`. or (b) within `replaceValInExprInputs(Expr* expr, Val* reference, Val* substitute)` make sure `reference` is different from `substitute`

(2) Yes, `RecomputeTv::recompute(buffer, producers)` will put `producers` to the clone map, since `buffer` is also one of the `producers`, it is already in the clone map. So nothing is recomputed, just returns the original input.

(3) You are right. Revised to make sure dummpy output is only added when the replacement actually happens.

naoyam (approved):
LGTM. Thanks for fixing the bug

PR #1535, opened 2023-12-15
Handle dynamic ExpandOp during concretization

### LINKED ISSUE #1398: symbolic iterdomain needs special handling on broadcast check.
Our new reshape API introduces symbolic iterdomain, which leaves `isBroadcast()` on symbolic iterdomain non-determined until concretization.

There's issue #1393, which our c++ expand API asserts when we try to expand a symbolic iterdomain, even though it has a static extent `1`.

@jacobhinkle gave a more general example than the one in the original issue here: https://github.com/NVIDIA/Fuser/pull/1394#discussion_r1408359746

```
  fusion->addInput(tv0);

  auto s0 = IrBuilder::create<Val>(DataType::Index);
  auto s1 = IrBuilder::create<Val>(DataType::Index);
  auto s2 = IrBuilder::create<Val>(DataType::Index);
  fusion->addInput(s0);
  fusion->addInput(s1);
  fusion->addInput(s2);
  fusion->addInput(s3);

  auto tv1 = reshape(tv0, {s0, s1});
  auto tv2 = expand(tv1, {s0, s2});

  fusion->addOutput(tv2);
```

Since the reshape here is symbolic and dynamic, we can only determine the semantic after concretization. So two things needs to be done:
1. c++ API needs to relax the check on `isBroadcast()` for operations like `expand` / `squeeze` e.t.c.. We can probably have a new `maybeBroadcast()` on `TensorView`/`TensorDomain`.
2. we also needs to add a `checkConcretization` as suggested https://github.com/NVIDIA/Fuser/pull/1394#discussion_r1408359746

### LINKED ISSUE #1468: Reshape then expand results in unexpanded output
While looking into #1398, I noticed that the following test exhibits unexpected behavior:
```c++
// See https://github.com/NVIDIA/Fuser/issues/1398
TEST_F(NVFuserTest, SymbolicExpand) {
  FusionGuard fg(fusion);

  fusion->addInput(tv0);

  auto s0 = IrBuilder::create<Val>(DataType::Index);
  auto s1 = IrBuilder::create<Val>(DataType::Index);
  auto s2 = IrBuilder::create<Val>(DataType::Index);
  auto s3 = IrBuilder::create<Val>(DataType::Index);
  fusion->addInput(s0);
  fusion->addInput(s1);
  fusion->addInput(s2);
  fusion->addInput(s3);

  auto tv1 = reshape(tv0, {s0, s1});
  auto tv2 = expand(tv1, {s2, s3});

  fusion->addOutput(tv2);

  FusionExecutorCache fec(std::move(fusion_ptr));

  at::Tensor t0 = at::randn({3, 2}, options);
  std::vector<c10::IValue> valid_inputs = {t0, 6, 1, 6, 5};
  // An invalid input has a second dimension that cannot be expanded
  std::vector<c10::IValue> invalid_inputs = {t0, 2, 3, 2, 5};

  auto outputs = fec.runFusionWithInputs(valid_inputs);

  std::cout << outputs.at(0) << std::endl;
  std::cout << "sizes: " << outputs[0].sizes() << std::endl;
  std::cout << "strides: " << outputs[0].strides() << std::endl;

  auto ref = at::reshape(t0, {6, 1}).expand({6, 5});

  std::cout << ref << std::endl;
  std::cout << "ref sizes: " << ref.sizes() << std::endl;
  std::cout << "ref strides: " << ref.strides() << std::endl;
  /*
-0.9247 -0.4253 -2.6438  0.1452 -0.1209
-0.5797     nan     nan     nan     nan
    nan     nan     nan     nan     nan
    nan     nan     nan     nan     nan
    nan     nan     nan     nan     nan
    nan     nan     nan     nan     nan
[ CUDAFloatType{6,5} ]
sizes: [6, 5]
strides: [5, 1]
-0.9247 -0.9247 -0.9247 -0.9247 -0.9247
-0.4253 -0.4253 -0.4253 -0.4253 -0.4253
-2.6438 -2.6438 -2.6438 -2.6438 -2.6438
 0.1452  0.1452  0.1452  0.1452  0.1452
-0.1209 -0.1209 -0.1209 -0.1209 -0.1209
-0.5797 -0.5797 -0.5797 -0.5797 -0.5797
[ CUDAFloatType{6,5} ]
ref sizes: [6, 5]
ref strides: [1, 0]
  */
}
```
As we can see, it appears the kernel correctly writes as if the output is expanded, but the strides are computed in a way that is ignoring the expand.

Note that this does not happen with a simple fusion consisting of only an expand:
```c++
TEST_F(NVFuserTest, ExpandedOutput) {
  FusionGuard fg(fusion);

  auto tv0 = makeSymbolicTensor(1);
  fusion->addInput(tv0);

  auto tv1 = broadcast(tv0, {false, true});
  auto tv2 = expand(tv1, {IrBuilder::create<Val>(6, DataType::Index),
      IrBuilder::create<Val>(5, DataType::Index)});

  fusion->addOutput(tv2);

  FusionExecutorCache fec(std::move(fusion_ptr));

  at::Tensor t0 = at::randn({6}, options);
  std::vector<c10::IValue> valid_inputs = {t0};

  auto outputs = fec.runFusionWithInputs(valid_inputs);

  std::cout << outputs.at(0) << std::endl;
  std::cout << "sizes: " << outputs[0].sizes() << std::endl;
  std::cout << "strides: " << outputs[0].strides() << std::endl;

  // We must add one here since nvfuser's output realizes the expand
  auto ref = t0.unsqueeze(1).expand({6, 5});

  std::cout << ref << std::endl;
  std::cout << "ref sizes: " << ref.sizes() << std::endl;
  std::cout << "ref strides: " << ref.strides() << std::endl;
  /*
-0.9247 -0.9247 -0.9247 -0.9247 -0.9247
-0.4253 -0.4253 -0.4253 -0.4253 -0.4253
-2.6438 -2.6438 -2.6438 -2.6438 -2.6438
 0.1452  0.1452  0.1452  0.1452  0.1452
-0.1209 -0.1209 -0.1209 -0.1209 -0.1209
-0.5797 -0.5797 -0.5797 -0.5797 -0.5797
[ CUDAFloatType{6,5} ]
sizes: [6, 5]
strides: [1, 0]
-0.9247 -0.9247 -0.9247 -0.9247 -0.9247
-0.4253 -0.4253 -0.4253 -0.4253 -0.4253
-2.6438 -2.6438 -2.6438 -2.6438 -2.6438
 0.1452  0.1452  0.1452  0.1452  0.1452
-0.1209 -0.1209 -0.1209 -0.1209 -0.1209
-0.5797 -0.5797 -0.5797 -0.5797 -0.5797
[ CUDAFloatType{6,5} ]
ref sizes: [6, 5]
ref strides: [1, 0]
  */
  testValidate(fusion, outputs, valid_inputs, {ref}, __LINE__, __FILE__);
}
```
Switching the reshape to a static reshape `auto tv1 = reshape(tv0, {3, 2}, {6, 1});` restores the expected behavior, so this seems to be a concretization bug.

jacobhinkle:
This adds `ExpandOp` to the possible dynamic transforms we explicitly handle during concretization, beside `ViewOp` and `Resize`.

`ExpandOp::checkConcretization` is not useful actually since the output of the expand is not replaced until after that method is called. That means we have no way to tell which axes are actually expanded, so we cannot test whether the inputs should be `Broadcast` at that point. Instead, this check is done at `analyzeExpands`, and the `SymbolicExpand` test exercises it by passing in invalid sizes.

Since this adds a member to both `DynamicTransformInitialInfo` and `DynamicTransformConcretizationInfo`, there are a number of mechanical changes to cloning, comparison operator, hash(), etc.  The important changes are in `arith.cpp`, and in `dynamic_transform.cpp` in `analyzeExpands` and `concretizeExpand`.

Fixes #1398. Fixes #1468.


csrc/ir/nodes.cpp
```diff
@@ -2891,10 +2891,10 @@ std::string IterDomain::toString(int indent_size) const {
...
     ss << stop()->toInlineString() << " : ";
   }
-  if (isBroadcast() && hasExpandedExtent()) {
-    ss << expandedExtent()->toInlineString();
-  } else {
-    ss << extent()->toInlineString();
+  // TODO: before merging, undo this debug style or hide behind an option
+  ss << extent()->toInlineString();
+  if (hasExpandedExtent()) {
+    ss << " expanded to " << expandedExtent()->toInlineString();
```
jacobhinkle, line 2897: Changes `toString()` to look like
```
%kernel {                                                                                                                                
T4_l[ iS14{( i0 * i2 )}rf ] = view( T0_g[ iS0{i0}, iS1{i2} ] )      
T5_l[ iS15{( i0 * i2 )}, bS16{1} ]                                                                                                       
   = broadcast( T4_l[ iS14{( i0 * i2 )}rf ] )                       
T2_l[ iS18{i5}rf, bS19{1 expanded to i6}rf ] = expand( T5_l[ iS15{( i0 * i2 )}, bS16{1} ], {i5, i6} )
T3_g[ iS20{i5}, bS21{1 expanded to i6} ]                            
   = T2_l[ iS18{i5}rf, bS19{1 expanded to i6}rf ]                   
   + T2_l[ iS18{i5}rf, bS19{1 expanded to i6}rf ];    
```
This was useful for debugging since I was seeing expanded extents alongside non-unity regular extents. I have seen that before in other issues and it can be confusing. I am still leaning toward reverting this back to the existing behavior but I figured I'd leave it up for debate for now.
I was planning to res

csrc/ir/utils.h
```diff
@@ -172,6 +172,10 @@ std::vector<int> normalizeOld2New(
 //! Warning: Creates new Expr defining substitute.
 Expr* replaceValInExprInputs(Expr* expr, Val* reference, Val* substitute);
 
+//! Replace old_val with new_val in all active uses as well as in fusion
+//! outputs.
+void replaceVal(Val* old_val, Val* new_val);
```
jacobhinkle, line 177: I use this twice, and it may be useful more generally.

jacobhinkle:
!build --diff

jacobhinkle:
The change to `test_nanogpt_mha_dpa` looks like this at the IR level:
```diff
--- /root/nanogpt_mha_dpa_main	2023-12-15 20:25:28.937126605 +0000
+++ /root/nanogpt_mha_dpa_after1535	2023-12-15 20:44:42.630721370 +0000
@@ -1,264 +1,252 @@
 Fusion before concretization:
 Inputs:
   T0_g[ iS0{i0}, iS1{i1}, iS2{i2}, iS3{i3} ], float
   T1_g[ bS4{1}, bS5{1}, iS6{i6}, iS7{i7} ], float
 Outputs:
   T25_g[ iS102{16}, iS103{16}, iS104{128}, iS105{128} ], float
 
 %kernel_math {
 i53 = fmin(i6, 128);
 i55 = fmax(0, i53);
 i69 = fmin(i7, 128);
 i71 = fmax(0, i69);
 T3_l[ bS12{1}rf, bS13{1}rf, ?S15{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}rf, ?S17{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )}rf ]
    = slice( T1_g[ bS4{1}, bS5{1}, iS6{i6}, iS7{i7} ], { {0, 1, 1} {0, 1, 1} {0, i55, 1} {0, i71, 1} } )
 T4_l[ bS18{1}, bS19{1}, ?S20{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}, ?S21{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )} ]
    = double(0)
    == T3_l[ bS12{1}rf, bS13{1}rf, ?S15{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}rf, ?S17{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )}rf ];
 T5_l[ bS22{1}, bS23{1}, ?S24{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}, ?S25{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )} ]
    = Set( T4_l[ bS18{1}, bS19{1}, ?S20{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}, ?S21{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )} ], cache_op=Streaming )
 i97 = (nvfuser_index_t)(16);
 i99 = (nvfuser_index_t)(16);
-i58 = -i6;
-i60 = i55 + i58;
-i63 = i6 + i60;
-i74 = -i7;
-i76 = i71 + i74;
-i79 = i7 + i76;
-T6_l[ bS26{( (nvfuser_index_t)(16) )}, bS27{( (nvfuser_index_t)(16) )}, ?S28{( (nvfuser_index_t)(128) )}, ?S29{( (nvfuser_index_t)(128) )} ] = expand( T5_l[ bS22{1}, bS23{1}, ?S24{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}, ?S25{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )} ], {i97, i99, i63, i79} )
+i101 = (nvfuser_index_t)(128);
+i103 = (nvfuser_index_t)(128);
+T6_l[ bS26{( (nvfuser_index_t)(16) )}, bS27{( (nvfuser_index_t)(16) )}, ?S28{( (nvfuser_index_t)(128) )}, ?S29{( (nvfuser_index_t)(128) )} ] = expand( T5_l[ bS22{1}, bS23{1}, ?S24{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}, ?S25{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )} ], {i97, i99, i101, i103} )
 T2_l[ iS8{i0}, iS9{i1}, iS10{i2}, iS11{i3} ]
    = T0_g[ iS0{i0}, iS1{i1}, iS2{i2}, iS3{i3} ]
    * double(0.125);
 T7_l[ iS30{i0}, iS31{i1}, iS32{128}, iS33{128} ]
    = where(T6_l[ bS26{( (nvfuser_index_t)(16) )}, bS27{( (nvfuser_index_t)(16) )}, ?S28{( (nvfuser_index_t)(128) )}, ?S29{( (nvfuser_index_t)(128) )} ]
   , double(-inf)
   , T2_l[ iS8{i0}, iS9{i1}, iS10{i2}, iS11{i3} ]);
 T8_l[ iS34{i0}, iS35{i1}, iS36{128}, rS37{128} ]
    = reduction( T7_l[ iS30{i0}, iS31{i1}, iS32{128}, iS33{128} ], op = fmax, initial value = double(-inf), allreduce = false )
 T9_l[ iS38{i0}, iS39{i1}, iS40{128}, bS41{1} ]
    = broadcast( T8_l[ iS34{i0}, iS35{i1}, iS36{128}, rS37{128} ] )
 T10_l[ iS42{( (nvfuser_index_t)(16) )}, iS43{( (nvfuser_index_t)(16) )}, iS44{128}, bS45{1} ]
    = Set( T9_l[ iS38{i0}, iS39{i1}, iS40{128}, bS41{1} ], cache_op=Streaming )
 T11_l[ iS46{( (nvfuser_index_t)(16) )}, iS47{( (nvfuser_index_t)(16) )}, iS48{128}, bS49{1} ]
    = Set( T10_l[ iS42{( (nvfuser_index_t)(16) )}, iS43{( (nvfuser_index_t)(16) )}, iS44{128}, bS45{1} ], cache_op=Streaming )
 i129 = (nvfuser_index_t)(16);
 i131 = (nvfuser_index_t)(16);
 i147 = (nvfuser_index_t)(128);
 T12_l[ iS50{( (nvfuser_index_t)(16) )}, iS51{( (nvfuser_index_t)(16) )}, iS52{128}, bS53{( (nvfuser_index_t)(128) )} ] = expand( T11_l[ iS46{( (nvfuser_index_t)(16) )}, iS47{( (nvfuser_index_t)(16) )}, iS48{128}, bS49{1} ], {i129, i131, 128, i147} )
 T13_l[ iS54{16}, iS55{16}, iS56{128}, iS57{128} ]
    = T7_l[ iS30{i0}, iS31{i1}, iS32{128}, iS33{128} ]
    - T12_l[ iS50{( (nvfuser_index_t)(16) )}, iS51{( (nvfuser_index_t)(16) )}, iS52{128}, bS53{( (nvfuser_index_t)(128) )} ];
 T14_l[ iS58{16}, iS59{16}, iS60{128}, iS61{128} ]
    = expf(T13_l[ iS54{16}, iS55{16}, iS56{128}, iS57{128} ]);
 T15_l[ iS62{16}, iS63{16}, iS64{128}, rS65{128} ]
    = reduction( T14_l[ iS58{16}, iS59{16}, iS60{128}, iS61{128} ], op = add, initial value = float(0), allreduce = false )
 T16_l[ iS66{16}, iS67{16}, iS68{128}, bS69{1} ]
    = broadcast( T15_l[ iS62{16}, iS63{16}, iS64{128}, rS65{128} ] )
 T17_l[ iS70{16}, iS71{16}, iS72{128}, bS73{1} ]
    = Set( T16_l[ iS66{16}, iS67{16}, iS68{128}, bS69{1} ], cache_op=Streaming )
 T18_l[ iS74{16}, iS75{16}, iS76{128}, bS77{1} ]
    = Set( T17_l[ iS70{16}, iS71{16}, iS72{128}, bS73{1} ], cache_op=Streaming )
 i195 = (nvfuser_index_t)(128);
 T19_l[ iS78{16}, iS79{16}, iS80{128}, bS81{( (nvfuser_index_t)(128) )} ] = expand( T18_l[ iS74{16}, iS75{16}, iS76{128}, bS77{1} ], {16, 16, 128, i195} )
 T20_l[ iS82{16}, iS83{16}, iS84{128}, iS85{128} ]
    = T14_l[ iS58{16}, iS59{16}, iS60{128}, iS61{128} ]
    / T19_l[ iS78{16}, iS79{16}, iS80{128}, bS81{( (nvfuser_index_t)(128) )} ];
 T21_l[ iS86{16}, iS87{16}, iS88{128}, iS89{128} ]
    = rng_uniform_range({16, 16, 128, 128}, double(0), double(1), float);
 T22_l[ iS90{16}, iS91{16}, iS92{128}, iS93{128} ]
    = T21_l[ iS86{16}, iS87{16}, iS88{128}, iS89{128} ]
    < double(1);
 T23_l[ iS94{16}, iS95{16}, iS96{128}, iS97{128} ]
    = (float)(T22_l[ iS90{16}, iS91{16}, iS92{128}, iS93{128} ]);
 T24_l[ iS98{16}, iS99{16}, iS100{128}, iS101{128} ]
    = T20_l[ iS82{16}, iS83{16}, iS84{128}, iS85{128} ]
    * T23_l[ iS94{16}, iS95{16}, iS96{128}, iS97{128} ];
 T25_g[ iS102{16}, iS103{16}, iS104{128}, iS105{128} ]
    = T24_l[ iS98{16}, iS99{16}, iS100{128}, iS101{128} ]
    * double(1);
 }
 
 Concretized Fusion:
 
 %kernel {
 i53 = fmin(i6, 128);
 i55 = fmax(0, i53);
 i69 = fmin(i7, 128);
 i71 = fmax(0, i69);
 T3_l[ bS12{1}rf, bS13{1}rf, iS106{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}rf, iS107{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )}rf ]
    = slice( T1_g[ bS4{1}, bS5{1}, iS6{i6}, iS7{i7} ], { {0, 1, 1} {0, 1, 1} {0, i55, 1} {0, i71, 1} } )
-T4_l[ bS18{1}, bS19{1}, iS108{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}, iS109{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )} ]
+T4_l[ bS18{1}, bS19{1}, iS112{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}, iS113{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )} ]
    = double(0)
    == T3_l[ bS12{1}rf, bS13{1}rf, iS106{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}rf, iS107{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )}rf ];
-T5_l[ bS22{1}, bS23{1}, iS110{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}, iS111{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )} ]
-   = Set( T4_l[ bS18{1}, bS19{1}, iS108{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}, iS109{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )} ], cache_op=Streaming )
-i97 = (nvfuser_index_t)(16);
-i99 = (nvfuser_index_t)(16);
-i58 = -i6;
-i60 = i55 + i58;
-i63 = i6 + i60;
-i74 = -i7;
-i76 = i71 + i74;
-i79 = i7 + i76;
-T6_l[ bS26{( (nvfuser_index_t)(16) )}, bS27{( (nvfuser_index_t)(16) )}, iS112{( (nvfuser_index_t)(128) )}, iS113{( (nvfuser_index_t)(128) )} ] = expand( T5_l[ bS22{1}, bS23{1}, iS110{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}, iS111{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )} ], {i97, i99, i63, i79} )
+T5_l[ bS22{1}, bS23{1}, iS114{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}, iS115{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )} ]
+   = Set( T4_l[ bS18{1}, bS19{1}, iS112{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}, iS113{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )} ], cache_op=Streaming )
+T26_l[ bS108{1}, bS109{1}, iS116{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}, iS117{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )} ]
+   = Set( T5_l[ bS22{1}, bS23{1}, iS114{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}, iS115{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )} ], cache_op=Streaming )
 T2_l[ iS8{i0}, iS9{i1}, iS10{i2}, iS11{i3} ]
    = T0_g[ iS0{i0}, iS1{i1}, iS2{i2}, iS3{i3} ]
    * double(0.125);
 T7_l[ iS30{i0}, iS31{i1}, iS32{128}, iS33{128} ]
-   = where(T6_l[ bS26{( (nvfuser_index_t)(16) )}, bS27{( (nvfuser_index_t)(16) )}, iS112{( (nvfuser_index_t)(128) )}, iS113{( (nvfuser_index_t)(128) )} ]
+   = where(T26_l[ bS108{1}, bS109{1}, iS116{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}, iS117{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )} ]
   , double(-inf)
   , T2_l[ iS8{i0}, iS9{i1}, iS10{i2}, iS11{i3} ]);
 T8_l[ iS34{i0}, iS35{i1}, iS36{128}, rS37{128} ]
    = reduction( T7_l[ iS30{i0}, iS31{i1}, iS32{128}, iS33{128} ], op = fmax, initial value = double(-inf), allreduce = false )
 T9_l[ iS38{i0}, iS39{i1}, iS40{128}, bS41{1} ]
    = broadcast( T8_l[ iS34{i0}, iS35{i1}, iS36{128}, rS37{128} ] )
 T10_l[ iS42{( (nvfuser_index_t)(16) )}, iS43{( (nvfuser_index_t)(16) )}, iS44{128}, bS45{1} ]
    = Set( T9_l[ iS38{i0}, iS39{i1}, iS40{128}, bS41{1} ], cache_op=Streaming )
 T11_l[ iS46{( (nvfuser_index_t)(16) )}, iS47{( (nvfuser_index_t)(16) )}, iS48{128}, bS49{1} ]
    = Set( T10_l[ iS42{( (nvfuser_index_t)(16) )}, iS43{( (nvfuser_index_t)(16) )}, iS44{128}, bS45{1} ], cache_op=Streaming )
 i129 = (nvfuser_index_t)(16);
 i131 = (nvfuser_index_t)(16);
 i147 = (nvfuser_index_t)(128);
 T12_l[ iS50{( (nvfuser_index_t)(16) )}, iS51{( (nvfuser_index_t)(16) )}, iS52{128}, bS53{( (nvfuser_index_t)(128) )} ] = expand( T11_l[ iS46{( (nvfuser_index_t)(16) )}, iS47{( (nvfuser_index_t)(16) )}, iS48{128}, bS49{1} ], {i129, i131, 128, i147} )
 T13_l[ iS54{16}, iS55{16}, iS56{128}, iS57{128} ]
    = T7_l[ iS30{i0}, iS31{i1}, iS32{128}, iS33{128} ]
    - T12_l[ iS50{( (nvfuser_index_t)(16) )}, iS51{( (nvfuser_index_t)(16) )}, iS52{128}, bS53{( (nvfuser_index_t)(128) )} ];
 T14_l[ iS58{16}, iS59{16}, iS60{128}, iS61{128} ]
    = expf(T13_l[ iS54{16}, iS55{16}, iS56{128}, iS57{128} ]);
 T15_l[ iS62{16}, iS63{16}, iS64{128}, rS65{128} ]
    = reduction( T14_l[ iS58{16}, iS59{16}, iS60{128}, iS61{128} ], op = add, initial value = float(0), allreduce = false )
 T16_l[ iS66{16}, iS67{16}, iS68{128}, bS69{1} ]
    = broadcast( T15_l[ iS62{16}, iS63{16}, iS64{128}, rS65{128} ] )
 T17_l[ iS70{16}, iS71{16}, iS72{128}, bS73{1} ]
    = Set( T16_l[ iS66{16}, iS67{16}, iS68{128}, bS69{1} ], cache_op=Streaming )
 T18_l[ iS74{16}, iS75{16}, iS76{128}, bS77{1} ]
    = Set( T17_l[ iS70{16}, iS71{16}, iS72{128}, bS73{1} ], cache_op=Streaming )
 i195 = (nvfuser_index_t)(128);
 T19_l[ iS78{16}, iS79{16}, iS80{128}, bS81{( (nvfuser_index_t)(128) )} ] = expand( T18_l[ iS74{16}, iS75{16}, iS76{128}, bS77{1} ], {16, 16, 128, i195} )
 T20_l[ iS82{16}, iS83{16}, iS84{128}, iS85{128} ]
    = T14_l[ iS58{16}, iS59{16}, iS60{128}, iS61{128} ]
    / T19_l[ iS78{16}, iS79{16}, iS80{128}, bS81{( (nvfuser_index_t)(128) )} ];
 T21_l[ iS86{16}, iS87{16}, iS88{128}, iS89{128} ]
    = rng_uniform_range({16, 16, 128, 128}, double(0), double(1), float);
 T22_l[ iS90{16}, iS91{16}, iS92{128}, iS93{128} ]
    = T21_l[ iS86{16}, iS87{16}, iS88{128}, iS89{128} ]
    < double(1);
 T23_l[ iS94{16}, iS95{16}, iS96{128}, iS97{128} ]
    = (float)(T22_l[ iS90{16}, iS91{16}, iS92{128}, iS93{128} ]);
 T24_l[ iS98{16}, iS99{16}, iS100{128}, iS101{128} ]
    = T20_l[ iS82{16}, iS83{16}, iS84{128}, iS85{128} ]
    * T23_l[ iS94{16}, iS95{16}, iS96{128}, iS97{128} ];
 T25_g[ iS102{16}, iS103{16}, iS104{128}, iS105{128} ]
    = T24_l[ iS98{16}, iS99{16}, iS100{128}, iS101{128} ]
    * double(1);
 
 TransformPrinter :
 T1_g[ bS4{1}, bS5{1}, iS6{i6}, iS7{i7} ]
  root domain : (bS4{1}, bS5{1}, iS6{i6}, iS7{i7})
  contiguity: n n t t
  leaf domain : (bS4{1}, bS5{1}, iS6{i6}, iS7{i7})
 T3_l[ bS12{1}rf, bS13{1}rf, iS106{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}rf, iS107{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )}rf ]
  root domain : (bS12{1}rf, bS13{1}rf, iS14{i6}rf, iS16{i7}rf)
   Resize: iS14{i6}rf by 0 and ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) -> iS106{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}rf
   Resize: iS16{i7}rf by 0 and ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) -> iS107{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )}rf
  rfactor domain : (bS12{1}rf, bS13{1}rf, iS106{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}rf, iS107{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )}rf)
  contiguity: n n t t
  leaf domain : (bS12{1}rf, bS13{1}rf, iS106{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}rf, iS107{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )}rf)
-T4_l[ bS18{1}, bS19{1}, iS108{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}, iS109{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )} ]
- root domain : (bS18{1}, bS19{1}, iS108{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}, iS109{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )})
+T4_l[ bS18{1}, bS19{1}, iS112{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}, iS113{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )} ]
+ root domain : (bS18{1}, bS19{1}, iS112{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}, iS113{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )})
  contiguity: n n t t
- leaf domain : (bS18{1}, bS19{1}, iS108{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}, iS109{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )})
-T5_l[ bS22{1}, bS23{1}, iS110{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}, iS111{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )} ]
- root domain : (bS22{1}, bS23{1}, iS110{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}, iS111{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )})
+ leaf domain : (bS18{1}, bS19{1}, iS112{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}, iS113{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )})
+T5_l[ bS22{1}, bS23{1}, iS114{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}, iS115{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )} ]
+ root domain : (bS22{1}, bS23{1}, iS114{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}, iS115{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )})
  contiguity: n n t t
- leaf domain : (bS22{1}, bS23{1}, iS110{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}, iS111{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )})
-T6_l[ bS26{( (nvfuser_index_t)(16) )}, bS27{( (nvfuser_index_t)(16) )}, iS112{( (nvfuser_index_t)(128) )}, iS113{( (nvfuser_index_t)(128) )} ]
- root domain : (bS26{( (nvfuser_index_t)(16) )}, bS27{( (nvfuser_index_t)(16) )}, iS112{( (nvfuser_index_t)(128) )}, iS113{( (nvfuser_index_t)(128) )})
+ leaf domain : (bS22{1}, bS23{1}, iS114{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}, iS115{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )})
+T26_l[ bS108{1}, bS109{1}, iS116{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}, iS117{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )} ]
+ root domain : (bS108{1}, bS109{1}, iS116{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}, iS117{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )})
  contiguity: n n t t
- leaf domain : (bS26{( (nvfuser_index_t)(16) )}, bS27{( (nvfuser_index_t)(16) )}, iS112{( (nvfuser_index_t)(128) )}, iS113{( (nvfuser_index_t)(128) )})
+ leaf domain : (bS108{1}, bS109{1}, iS116{( i6 + ( ( fmax(0, ( fmin(i6, 128) )) ) + ( -i6 ) ) )}, iS117{( i7 + ( ( fmax(0, ( fmin(i7, 128) )) ) + ( -i7 ) ) )})
 T0_g[ iS0{i0}, iS1{i1}, iS2{i2}, iS3{i3} ]
  root domain : (iS0{i0}, iS1{i1}, iS2{i2}, iS3{i3})
  contiguity: t t t t
  leaf domain : (iS0{i0}, iS1{i1}, iS2{i2}, iS3{i3})
 T2_l[ iS8{i0}, iS9{i1}, iS10{i2}, iS11{i3} ]
  root domain : (iS8{i0}, iS9{i1}, iS10{i2}, iS11{i3})
  contiguity: t t t t
  leaf domain : (iS8{i0}, iS9{i1}, iS10{i2}, iS11{i3})
 T7_l[ iS30{i0}, iS31{i1}, iS32{128}, iS33{128} ]
  root domain : (iS30{i0}, iS31{i1}, iS32{128}, iS33{128})
  contiguity: t t t t
  leaf domain : (iS30{i0}, iS31{i1}, iS32{128}, iS33{128})
 T8_l[ iS34{i0}, iS35{i1}, iS36{128}, rS37{128} ]
  root domain : (iS34{i0}, iS35{i1}, iS36{128}, rS37{128})
  contiguity: t t t n
  leaf domain : (iS34{i0}, iS35{i1}, iS36{128}, rS37{128})
 T9_l[ iS38{i0}, iS39{i1}, iS40{128}, bS41{1} ]
  root domain : (iS38{i0}, iS39{i1}, iS40{128}, bS41{1})
  contiguity: t t t n
  leaf domain : (iS38{i0}, iS39{i1}, iS40{128}, bS41{1})
 T10_l[ iS42{( (nvfuser_index_t)(16) )}, iS43{( (nvfuser_index_t)(16) )}, iS44{128}, bS45{1} ]
  root domain : (iS42{( (nvfuser_index_t)(16) )}, iS43{( (nvfuser_index_t)(16) )}, iS44{128}, bS45{1})
  contiguity: t t t n
  leaf domain : (iS42{( (nvfuser_index_t)(16) )}, iS43{( (nvfuser_index_t)(16) )}, iS44{128}, bS45{1})
 T11_l[ iS46{( (nvfuser_index_t)(16) )}, iS47{( (nvfuser_index_t)(16) )}, iS48{128}, bS49{1} ]
  root domain : (iS46{( (nvfuser_index_t)(16) )}, iS47{( (nvfuser_index_t)(16) )}, iS48{128}, bS49{1})
  contiguity: t t t n
  leaf domain : (iS46{( (nvfuser_index_t)(16) )}, iS47{( (nvfuser_index_t)(16) )}, iS48{128}, bS49{1})
 T12_l[ iS50{( (nvfuser_index_t)(16) )}, iS51{( (nvfuser_index_t)(16) )}, iS52{128}, bS53{( (nvfuser_index_t)(128) )} ]
  root domain : (iS50{( (nvfuser_index_t)(16) )}, iS51{( (nvfuser_index_t)(16) )}, iS52{128}, bS53{( (nvfuser_index_t)(128) )})
  contiguity: t t t n
  leaf domain : (iS50{( (nvfuser_index_t)(16) )}, iS51{( (nvfuser_index_t)(16) )}, iS52{128}, bS53{( (nvfuser_index_t)(128) )})
 T13_l[ iS54{16}, iS55{16}, iS56{128}, iS57{128} ]
  root domain : (iS54{16}, iS55{16}, iS56{128}, iS57{128})
  contiguity: t t t t
  leaf domain : (iS54{16}, iS55{16}, iS56{128}, iS57{128})
 T14_l[ iS58{16}, iS59{16}, iS60{128}, iS61{128} ]
  root domain : (iS58{16}, iS59{16}, iS60{128}, iS61{128})
  contiguity: t t t t
  leaf domain : (iS58{16}, iS59{16}, iS60{128}, iS61{128})
 T15_l[ iS62{16}, iS63{16}, iS64{128}, rS65{128} ]
  root domain : (iS62{16}, iS63{16}, iS64{128}, rS65{128})
  contiguity: t t t n
  leaf domain : (iS62{16}, iS63{16}, iS64{128}, rS65{128})
 T16_l[ iS66{16}, iS67{16}, iS68{128}, bS69{1} ]
  root domain : (iS66{16}, iS67{16}, iS68{128}, bS69{1})
  contiguity: t t t n
  leaf domain : (iS66{16}, iS67{16}, iS68{128}, bS69{1})
 T17_l[ iS70{16}, iS71{16}, iS72{128}, bS73{1} ]
  root domain : (iS70{16}, iS71{16}, iS72{128}, bS73{1})
  contiguity: t t t n
  leaf domain : (iS70{16}, iS71{16}, iS72{128}, bS73{1})
 T18_l[ iS74{16}, iS75{16}, iS76{128}, bS77{1} ]
  root domain : (iS74{16}, iS75{16}, iS76{128}, bS77{1})
  contiguity: t t t n
  leaf domain : (iS74{16}, iS75{16}, iS76{128}, bS77{1})
 T19_l[ iS78{16}, iS79{16}, iS80{128}, bS81{( (nvfuser_index_t)(128) )} ]
  root domain : (iS78{16}, iS79{16}, iS80{128}, bS81{( (nvfuser_index_t)(128) )})
  contiguity: t t t n
  leaf domain : (iS78{16}, iS79{16}, iS80{128}, bS81{( (nvfuser_index_t)(128) )})
 T20_l[ iS82{16}, iS83{16}, iS84{128}, iS85{128} ]
  root domain : (iS82{16}, iS83{16}, iS84{128}, iS85{128})
  contiguity: t t t t
  leaf domain : (iS82{16}, iS83{16}, iS84{128}, iS85{128})
 T21_l[ iS86{16}, iS87{16}, iS88{128}, iS89{128} ]
  root domain : (iS86{16}, iS87{16}, iS88{128}, iS89{128})
  contiguity: t t t t
  leaf domain : (iS86{16}, iS87{16}, iS88{128}, iS89{128})
 T22_l[ iS90{16}, iS91{16}, iS92{128}, iS93{128} ]
  root domain : (iS90{16}, iS91{16}, iS92{128}, iS93{128})
  contiguity: t t t t
  leaf domain : (iS90{16}, iS91{16}, iS92{128}, iS93{128})
 T23_l[ iS94{16}, iS95{16}, iS96{128}, iS97{128} ]
  root domain : (iS94{16}, iS95{16}, iS96{128}, iS97{128})
  contiguity: t t t t
  leaf domain : (iS94{16}, iS95{16}, iS96{128}, iS97{128})
 T24_l[ iS98{16}, iS99{16}, iS100{128}, iS101{128} ]
  root domain : (iS98{16}, iS99{16}, iS100{128}, iS101{128})
  contiguity: t t t t
  leaf domain : (iS98{16}, iS99{16}, iS100{128}, iS101{128})
 T25_g[ iS102{16}, iS103{16}, iS104{128}, iS105{128} ]
  root domain : (iS102{16}, iS103{16}, iS104{128}, iS105{128})
  contiguity: t t t t
  leaf domain : (iS102{16}, iS103{16}, iS104{128}, iS105{128})
 }
```
This might be a logic error at definition where we are erroneously replacing a real expand with a `set`. Interesting that it reduces register use so much and that the test still passes...

jjsjann123:
> The important changes are in alias.h, and in dynamic_transform.cpp in analyzeExpands and concretizeExpand.

I'm actually not seeing any diffs in `alias.h`.

csrc/ir/utils.cpp
```diff
@@ -198,6 +198,16 @@ Expr* replaceValInExprInputs(Expr* expr, Val* reference, Val* substitute) {
       expr, reference, substitute);
 }
 
+void replaceVal(Val* old_val, Val* new_val) {
```
jjsjann123, line 201: nitpick: We have `replaceValue`, which is a more general version? That one does require a hash map for the replacement.... maybe we can switch the implementation to use that for consistency.

csrc/ir/nodes.cpp
jjsjann123: nitpick: `expanded to` feels a bit too elaborate, maybe simpler form as `->` or some simpler mark.

jacobhinkle:
> > The important changes are in alias.h, and in dynamic_transform.cpp in analyzeExpands and concretizeExpand.
> 
> I'm actually not seeing any diffs in `alias.h`.

Sorry that was a typo. Check alias.cpp

csrc/ops/arith.cpp
```diff
@@ -1478,11 +1479,10 @@ TensorView* expand(TensorView* inp, const std::vector<Val*>& expanded_sizes) {
...
     } else if (
-        (inp_id->isBroadcast() ||
-         // special patch for Symbolic IterDomain with a static size-1 extent
-         // See Issue: https://github.com/NVIDIA/Fuser/pull/1393
-         (inp_id->isSymbolic() && inp_id->extent()->isConstInt() &&
-          inp_id->extent()->evaluate() == 1)) &&
+        // special patch for Symbolic IterDomain with a static size-1 extent
+        // since we know it will become broadcast at concretization
+        // See Issue: https://github.com/NVIDIA/Fuser/pull/1393
+        (inp_id->extent()->isConstInt() && inp_id->extent()->evaluate() == 1) &&
```
jjsjann123, line 1485: Does this mean we need to update our `isBroadcast()` implementation to use this notion `(inp_id->extent()->isConstInt() && inp_id->extent()->evaluate() == 1)` instead?

With symbolic_iterDomain, clearly the `isBraodcast()` is no longer sufficient any more.

csrc/dynamic_transform.h
```diff
@@ -213,6 +234,11 @@ class DynamicTransformConcretizationInfo {
   //! with its concretized IterType
   std::vector<std::pair<size_t, IterType>> resize_itertypes_;
 
+  //! Holds the index of the expanded TensorView in the vector returned by
+  //! initial_info_->getDynamicExpandedTensorViews(), and a corresponding vector
+  //! of bools indicating whether each axis is in fact expanded.
+  std::vector<std::pair<size_t, std::vector<bool>>> expand_axes_;
```
jjsjann123, line 240: nitpicking on the mapping here using index... but I see other structure in concretization info uses similar mapping. :shrug:

csrc/dynamic_transform.cpp
```diff
@@ -330,6 +375,53 @@ void DynamicTransformConcretizationInfo::analyzeResizes(
...
+            inp_id->toString(),
+            " has possibly expanded extent ",
+            in_size,
+            " which is incompatible with expansion to size ",
+            out_size,
+            ". Note that already-expanded axes may not themselves be expanded.");
+        expand_axes.push_back(false);
+      }
+    }
+    expand_axes_.emplace_back(tv_index, expand_axes);
```
jjsjann123, line 421: having said that, it's so tempting to just replace `tv_index` here with `out_tv` for the mapping. :)

csrc/dynamic_transform.h
jacobhinkle: Yes, using the raw pointer seems at first like it would be more convenient right; that is how this was originally implemented. However, it gets complicated when we need to clone the dynamic Fusion for concretization: at that point we need the mapping from the original pointers in the dynamic definition to the new cloned versions. We also need to be able to compare concretization info objects and hash them, since they act as cache keys. The solution we use is to attach `DynamicTransformInitialInfo` to the fusion in the form of managed data, so that it is automatically cloned when the Fusion is cloned. Then we refer to objects held by it inside the concretization info, which lets us hold those outside the Fusion, and still apply them to any Fusion clone we want.

csrc/ir/utils.cpp
jacobhinkle: Some reworking or renaming is in order here for sure. The problem is that while `replaceValue` is more general in the sense that it can do multiple replacements in a single traversal, it doesn't replace any vals in `fusion->outputs()`. This utility is meant to replace it in all uses and also in the outputs list. It could use a better name though at the very least.

csrc/ops/arith.cpp
jacobhinkle: Until #1393 I thought we would not have `Symbolic` ID with size-1 extent, but then we did. It seems like expanding a symbolic domain to constant size 1 is the only way for that to happen, so we should probably more carefully check that particular case here and ensure the output will be set as `Broadcast` (I don't think it will be in the current version here).

I don't think we should modify `isBroadcast()`. `Symbolic` can still be either `Iteration` or `Broadcast` and `isBroadcast` only means it's guaranteed to be broadcast.

csrc/dynamic_transform.cpp
jacobhinkle: I feel you.

csrc/ops/arith.cpp
jjsjann123: Yeah I agree with you on keeping `isBroadcast` as-is (a guaranteed broadcast).

Does it make sense to add a `maybeBroadcast` as a simpler API to address the ambiguity introduced by symbolic iterDomain?

jacobhinkle: Something like the following?
```c++
namespace ir_utils {
bool maybeBroadcast(IterDomain* id) {
  return id->isBroadcast() || id->isSymbolic();
}
} // namespace ir_utils
```
Sounds good but the name in this case is not quite as precise as the explicit condition, so it might be a little confusing :man_shrugging: ?

jjsjann123: I'm bad with names. if you think it's a confusing name, I'll trust your judgement on that one :laughing: 

please just ignore the previous request. sorry about the noise.

csrc/ir/utils.cpp
naoyam: I think this needs to be more explicit and clear. For example, why just replacing outputs but not inputs? Would it make more sense to name it like `replaceAllExprInputsAndOutputs`?

csrc/dynamic_transform.cpp
```diff
@@ -63,12 +70,16 @@ std::string DynamicTransformInitialInfo::toString() const {
   ss << "DynamicTransformInitialInfo\n";
   std::string indent = "  ";
   ss << indent << "Dynamic reshaped TensorViews:\n";
-  for (const auto& op : dynamic_reshaped_tvs_) {
-    ss << indent << indent << op->toString() << "\n";
+  for (const auto& tv : dynamic_reshaped_tvs_) {
+    ss << indent << indent << tv->toString() << "\n";
```
naoyam, line 74: Related to the above comment, if `tv` can be `nullptr`, should we have the same guard here?

naoyam (approved):
LGTM. Great work!

csrc/dynamic_transform.cpp
```diff
@@ -26,15 +27,21 @@ DynamicTransformInitialInfo DynamicTransformInitialInfo::clone(
   DynamicTransformInitialInfo cloned_info(
       static_cast<Fusion*>(ir_cloner.container()));
   cloned_info.dynamic_reshaped_tvs_.reserve(dynamic_reshaped_tvs_.size());
-  for (const auto op : dynamic_reshaped_tvs_) {
-    if (op) {
-      cloned_info.dynamic_reshaped_tvs_.push_back(ir_cloner.clone(op));
+  for (const auto tv : dynamic_reshaped_tvs_) {
+    if (tv) {
```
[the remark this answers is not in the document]
jacobhinkle: Ah good catch. None of these should be `nullptr`. I think that's a holdover from a long time ago. I'll clean it up by removing all these checks.

csrc/dynamic_transform.cpp
jacobhinkle: Should never be null.

csrc/ir/utils.cpp
jacobhinkle: Went with the verbose but explicit `replaceValInAllExprInputsAndFusionOutputs`.

csrc/ir/nodes.cpp
jacobhinkle: Great point. Changed to `ex`, e.g.
```
%kernel {                                                                                                                                
T4_l[ iS14{( i0 * i2 )}rf ] = view( T0_g[ iS0{i0}, iS1{i2} ] )      
T5_l[ iS15{( i0 * i2 )}, bS16{1} ]                                                                                                       
   = broadcast( T4_l[ iS14{( i0 * i2 )}rf ] )                       
T2_l[ iS18{i5}rf, bS19{1 ex i6}rf ] = expand( T5_l[ iS15{( i0 * i2 )}, bS16{1} ], {i5, i6} )
T3_g[ iS20{i5}, bS21{1 ex i6} ]                            
   = T2_l[ iS18{i5}rf, bS19{1 ex i6}rf ]                   
   + T2_l[ iS18{i5}rf, bS19{1 ex i6}rf ]; 
```

PR #1536, opened 2023-12-15
Updated the usage of TensorMap in generated kernels

drzejan2:
The scope of introduced changes:
- modify the string representation of TensorMap object type for kernel
  lowering phase,
- fix the type of TensorMap local variables in kernel (aliases),
- add local helper tool to detect TensorMap data type,

### DIFF
--- a/csrc/codegen.cpp
+++ b/csrc/codegen.cpp
@@ -34,6 +34,21 @@ std::string ptrType(DataType dt) {
   return ss.str();
 }
 
+bool isTmaType(const DataType& dtype) {
+  return std::visit(
+      [](auto&& dtype) -> bool {
+        using T = std::decay_t<decltype(dtype)>;
+        if constexpr (std::is_same_v<T, PointerType>) {
+          return isTmaType(*dtype.type);
+        }
+        if constexpr (std::is_same_v<T, OpaqueType>) {
+          return 0 == dtype.name.compare("TensorMap");
+        }
+        return false;
+      },
+      dtype.type);
+}
+
 //! Utility class to build an argument list
 class ArgumentBuilder {
  public:
@@ -275,7 +290,12 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
         }
       } else {
         NVF_ERROR(param->isScalar()); // NOLINT (LLVM bug 48525)
-        code_ << param->dtype() << " " << var_name_ss.str();
+        if (isTmaType(param->dtype())) {
+          code_ << "const __grid_constant__ " << param->dtype() << " "
+                << var_name_ss.str();
+        } else {
+          code_ << param->dtype() << " " << var_name_ss.str();
+        }
       }
 
       if (i + 1 != kernel_->parameters().size()) {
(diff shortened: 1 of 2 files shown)


drzejan2:
I checked the following test cases with the proposed changes on h100 device:
```
TMATest.LoadCompleteTensor1D
TMATest.LoadCompleteTensor2D
TMATest.LoadCompleteTensor3D
TMATest.LoadCompleteTensor4D
TMATest.LoadCompleteTensor5D
TMATest.StoreCompleteTensor1D
TMATest.StoreCompleteTensor2D
TMATest.StoreCompleteTensor3D
TMATest.StoreCompleteTensor4D
TMATest.StoreCompleteTensor5D
TMATest.DisableIndexHoisting
```
there were no issues, all tests passed.

---

The issue surfaced during the work on Double Buffering / Circular buffer support for Hopper, where the kernel initially looked like this:

```
__global__ void nvfuser_none_f0_c0_r0_g0(Tensor<float, 1, 1> T0, const __grid_constant__ TensorMap var0, Tensor<float, 1, 1> T2) {
  alignas(16) extern __shared__ char array[];
  const unsigned smem_offset = 0;
  NVFUSER_DEFINE_MAGIC_ZERO;
  nvfuser_index_t i1;
  i1 = ceilDiv(T0.logical_size[0LL], 32);
  const __grid_constant__ TensorMap* ptr2;
  ptr2 = &var0;
...
...
CUDA NVRTC compile error: __tmp_kernel_none_f0_c0_r0_g0.cu(9451): error: attribute "grid_constant" does not apply here
    const __grid_constant__ TensorMap* ptr2;
          ^
```

With the change (the kernel still does not compile, but due to other issues):
```
__global__ void nvfuser_none_f0_c0_r0_g0(Tensor<float, 1, 1> T0, const __grid_constant__ TensorMap var0, Tensor<float, 1, 1> T2) {
  alignas(16) extern __shared__ char array[];
  const unsigned smem_offset = 0;
  NVFUSER_DEFINE_MAGIC_ZERO;
  nvfuser_index_t i1;
  i1 = ceilDiv(T0.logical_size[0LL], 32);
  const TensorMap* ptr2;
  ptr2 = &var0;
...
```

drzejan2:
!build

csrc/codegen.cpp
```diff
@@ -34,6 +34,25 @@ std::string ptrType(DataType dt) {
...
+        } else if constexpr (std::is_same_v<T, OpaqueType>) {
+          if (0 == dtype.name.compare("TensorMap")) {
+            return true;
+          } else {
+            return false;
+          }
+        } else {
+          return false;
+        }
+      },
```
wujingyue, line 52:
```suggestion
      [](auto&& dtype) -> bool {
        using T = std::decay_t<decltype(dtype)>;
        if constexpr (std::is_same_v<T, PointerType>) {
          return isTmaType(*dtype.type);
        }
        if constexpr (std::is_same_v<T, OpaqueType>) {
          return 0 == dtype.name.compare("TensorMap");
        }
        return false;
      },
```

wujingyue (commented):
A drive-by comment

zasdfgbnm (approved):
Jingyue's suggestion does make sense to me. Other than that, LGTM!

csrc/codegen.cpp
drzejan2: Thanks for suggestion, @wujingyue - you are right, this should be simplified. I will update the code. Thanks!

I made it like this as I wanted to keep it aligned with how `data_type2string(...)` is implemented, but that's an overkill.

drzejan2:
!build

PR #1540, opened 2023-12-16
Binding for matmul

kevinstephano:


kevinstephano:
No need to keep this open.

Issue #1537, opened 2023-12-15
OOM with increasing number of tests in Python Benchmarks


Priya2698:
PR #1488 ran into the pytest process being killed due to OOM with increase in number of tests in a single file. Possible reasons:

1. Kernel cache entries: https://github.com/NVIDIA/Fuser/pull/1532#issuecomment-1857349564
2. Pytest collects and maintains a test status for each test to generate a report at the end. However it is unlikely to alone consume ~30GB memory for 12.5k tests.

naoyam:
Is it CPU or GPU memory?

Priya2698:
> Is it CPU or GPU memory?

CPU memory.

naoyam:
#1407 was an issue with GPU memory. Compiled kernels are loaded into GPU memory, and eventually it took a significant portion of the available memory.

I don't think I have seen CPU OOM errors. Can you get more detailed information such as the physical memory capacity, how much was used for pytest, etc?

Priya2698:
Update: PR #1543 resolves the GPU OOM errors seen in PR #1488. The GPU memory usage pattern does not indicate that the errors are due to kernels being cached since the memory is eventually released. However, larger tests run consecutively still fail if memory allocated on the GPU in a previous test is not released before the next test executes.

jacobhinkle:
So collecting garbage then clearing cache reduces the host memory enough to prevent OOM? If so then I think that implies the underlying cause is not the kernel cache. In a failing test how many Fusions/kernels do we have when it fails?

csarofeen:
I wonder if it's the pytorch caching allocator failing due to fragmentation.

Priya2698:
> Update: PR #1543 resolves the GPU OOM errors seen in PR #1488. The GPU memory usage pattern does not indicate that the errors are due to kernels being cached since the memory is eventually released. However, larger tests run consecutively still fail if memory allocated on the GPU in a previous test is not released before the next test executes.

Each benchmark has only one fusion being executed for a range of input sizes. So the kernel cache size should not be large or result in OOM.

csarofeen:
Okay, yeah, if adding `torch.cuda.empty_cache()` resolves it then it's likely the PyTorch caching allocator.

naoyam:
Should this be closed?

PR #1543, opened 2023-12-19
Clear CUDA cache before running a test

Priya2698:
This PR clears CUDA cache at the start of each test instead of at the end. This avoids CUDA OOM errors in tests due to any remaining allocated memory from the previous test.

Cache is only cleared when reserved memory is greater than 80% of GPU memory or there is allocated memory at the start of the test. This significantly reduces the overhead associated with gc collection/clearing cache.
Credit: @jacobhinkle.

Resolves GPU OOM errors in Issue #1537.

### DIFF
--- a/python_benchmarks/core.py
+++ b/python_benchmarks/core.py
@@ -5,6 +5,7 @@ from typing import List, Callable, Union, Tuple
 from torch.autograd import DeviceType
 import gc
 
+
 def get_device_properties() -> Tuple[int, float]:
     """
     Computes device properties using ctypes and cuda.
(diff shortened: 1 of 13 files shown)


jacobhinkle (approved):
LGTM. One thing to consider is whether there is a measurable delay when doing gc or clearing the cache. If it is noticeable, then we may want to only do it when usage is above a certain threshold. You can check by just measuring wall time with and without this change I guess, for a set of small tests.

Priya2698:
> LGTM. One thing to consider is whether there is a measurable delay when doing gc or clearing the cache. If it is noticeable, then we may want to only do it when usage is above a certain threshold. You can check by just measuring wall time with and without this change I guess, for a set of small tests.

The overhead is not significant right now. I'll revisit this for weekly benchmarks, it may become a noticeable delay as I increase the number of input sizes.

Priya2698:
Thanks for the suggestion!
There is an overhead associated with collected garbage and clearing cache ( > 40%). It will not be a significant time delay with the current number of tests but may become noticeable with increase in number of tests.
Setting the threshold brings down the overhead to < 10% (as low as 1% for some subsets).
The threshold may have to be varied with change in input sizes in the future.

PR #1544, opened 2023-12-19
Fix bug in codediff for pytest

jacobhinkle:
This should fix the `assert self.current_test is not None` errors we see in CI sometimes.

Also, the default reference git ref was changed from `origin/main` to `git merge-base origin/main HEAD`. This can help when a branch is not up-to-date with upstream, since it prevents comparing against commits that have not yet been merged into the branch.

I also added a warning when the number of kernels in a test changes. We still compare the first N kernels where N is the minimum of the number of kernels between the two runs, but now if that number differs we give a warning. This helps alert us that we might not be comparing apples to apples and we should look at that specific test in more detail manually.

### DIFF
--- a/tools/codediff/compare_codegen.sh
+++ b/tools/codediff/compare_codegen.sh
@@ -6,11 +6,12 @@
 # compare_codegen.sh - compare generated CUDA kernels between git commits
 #
 # This script is made to compare generated kernels when making changes to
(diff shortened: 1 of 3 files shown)


jacobhinkle:
!build --diff

jacobhinkle:
!build --diff

xwang233 (approved):

PR #1546, opened 2023-12-19
Set index_type properly in matmul benchmarks

jacobhinkle:
This fixes a benchmark failure with large inputs due to hardcoded Int32 index type.


drzejan2:
Prepared changes look good - thanks for preparing them!

drzejan2 (approved):

PR #1550, opened 2023-12-21
Single reduction benchmark

Priya2698:
Benchmark inner and outer `sum` reduction.


jacobhinkle (approved):
LGTM. This reminds me though that in the C++ benchmarks we disable kernel reuse to ensure we don't use suboptimal kernels accidentally. See https://github.com/NVIDIA/Fuser/pull/563/files#diff-8d7acd33992d95340a01891d6fc419908ba005b9201ccc9c7eaeef0898478a1aR62. I am not sure whether we could currently do that but it's probably a good idea to run these Python benchmarks with NVFUSER_DISABLE=kernel_reuse.

Issue #633, opened 2023-07-21
`keepdim=False` then `broadcast_in_dim` removes explicit alias registers and shared memory can't be reused.


liqiangxl:
When working on #582 and test softmax, I found that if FusionDefinition use `keepdim=False` then `broadcast_in_dim`, the generated code failed to explicitly use alias registers. There is no influence on performance since the compiler can auto reuse. But this disables shared memory reuse when the register buffers are changed to shared memroy buffers in #582 .
Related to #221 
```
import torch
from nvfuser import FusionDefinition, DataType

def nvfuser_fusion_id0(fd : FusionDefinition) -> None :
    T0 = fd.define_tensor(symbolic_sizes=[-1, -1], contiguity=[True, True], dtype=DataType.Half, is_cpu=False)
    T1 = fd.ops.cast(T0, dtype=DataType.Float)
    T2 = fd.ops.max(T1, axes=[1], keepdim=False, dtype=DataType.Null)
    T3 = fd.ops.broadcast_in_dim(T2, output_shape=[8192, 1], broadcast_dims=[0])
    T4 = fd.ops.broadcast_in_dim(T3, output_shape=[8192, 10240], broadcast_dims=[0, 1])
    T5 = fd.ops.sub(T1, T4)
    T6 = fd.ops.exp(T5)
    T7 = fd.ops.sum(T6, axes=[1], keepdim=False, dtype=DataType.Null)
    T8 = fd.ops.broadcast_in_dim(T7, output_shape=[8192, 1], broadcast_dims=[0])
    T9 = fd.ops.broadcast_in_dim(T8, output_shape=[8192, 10240], broadcast_dims=[0, 1])
    T10 = fd.ops.div(T6, T9)
    T11 = fd.ops.cast(T10, dtype=DataType.Half)
    fd.add_output(T11)

with FusionDefinition() as fd:
    nvfuser_fusion_id0(fd)

inputs = [
    torch.randn((8192, 10240), dtype=torch.float16, device='cuda:0').as_strided((8192, 10240), (10240, 1)),
]
fd.execute(inputs)

```

If `keepdim=True`, alias register is explicitly used and also shared memory.
```
import torch
from nvfuser import FusionDefinition, DataType

def nvfuser_fusion_id0(fd : FusionDefinition) -> None :
    T0 = fd.define_tensor(symbolic_sizes=[-1, -1], contiguity=[True, True], dtype=DataType.Half, is_cpu=False)
[...]

PR #1561, opened 2023-12-26
allow expand op and adjust vectorization check in memory alias

liqiangxl:
Fix shared memory reuse in softmax by:
(1) allow `expand` op and treat it similar to broadcast, that is `require matched iterdomains for reuse`
(2) Allow non-vectorized tv to reuse vectorized tv.

see https://github.com/NVIDIA/Fuser/issues/633 and https://github.com/NVIDIA/Fuser/issues/1562

### DIFF
--- a/csrc/device_lower/pass/alias_memory.cpp
+++ b/csrc/device_lower/pass/alias_memory.cpp
@@ -1205,14 +1205,20 @@ class ReusableAllocationFinder : private kir::IrVisitor {
 
           // Vectorized allocations require correct alignment so if [this_tv]
           // is vectorized, the [reuse_tv] must be vectorized with the same
-          // factor.
+          // or smaller factor.
           // No need to check shared memory since it is always aligned to 16
           // Bytes which is also the maximum vectorization width.
           if (this_tv->getMemoryType() == MemoryType::Local) {
             const auto& va = GpuLower::current()->vectorizedAccesses();
-            if ((va.find(this_tv) != va.end())) {
-              if ((va.find(reuse_tv) == va.end()) ||
-                  va.at(this_tv) != va.at(reuse_tv)) {
+            bool this_tv_vectorized = va.find(this_tv) != va.end();
+            if (this_tv_vectorized) {
+              bool reuse_tv_vectorized = va.find(reuse_tv) != va.end();
+              if (!reuse_tv_vectorized) {
+              return false;
+              }
+              int this_tv_alignment = va.at(this_tv);
+              int reuse_tv_alignment = va.at(reuse_tv);
+              if (this_tv_alignment > reuse_tv_alignment) {
                 return false;
               }
             }
--- a/csrc/device_lower/pass/alias_memory.cpp
+++ b/csrc/device_lower/pass/alias_memory.cpp
@@ -1202,20 +1202,19 @@ class ReusableAllocationFinder : private kir::IrVisitor {
           auto this_tv = alloc_info->alloc_expr->buffer()->as<TensorView>();
           auto reuse_tv =
               alloc_to_reuse->alloc_expr->buffer()->as<TensorView>();
-          // Vectorized allocations require correct alignment so if [this_tv]
-          // is vectorized, the [reuse_tv] must also be vectorized.
-          const auto& va = GpuLower::current()->vectorizedAccesses();
-          if ((va.find(this_tv) != va.end()) &&
-              (va.find(reuse_tv) == va.end())) {
-            return false;
-          }
 
-          // Shared memory is all aligned to 128 bits, local memory might not be
-          if (this_tv->getMemoryType() == MemoryType::Local &&
-              va.find(this_tv) != va.end()) {
-            // Make sure alignment matches
-            if (va.at(this_tv) != va.at(reuse_tv)) {
-              return false;
+          // Vectorized allocations require correct alignment so if [this_tv]
+          // is vectorized, the [reuse_tv] must be vectorized with the same
+          // factor.
+          // No need to check shared memory since it is always aligned to 16
+          // Bytes which is also the maximum vectorization width.
+          if (this_tv->getMemoryType() == MemoryType::Local) {
+            const auto& va = GpuLower::current()->vectorizedAccesses();
+            if ((va.find(this_tv) != va.end())) {
+              if ((va.find(reuse_tv) == va.end()) ||
+                  va.at(this_tv) != va.at(reuse_tv)) {
+                return false;
+              }
             }
           }
         }
--- a/csrc/device_lower/pass/alias_memory.cpp
+++ b/csrc/device_lower/pass/alias_memory.cpp
@@ -1385,7 +1385,7 @@ class ReusableAllocationFinder : private kir::IrVisitor {
     }
   }
 
-  // Utility to capture broadcast and expand ops
+  // Utility to capture broadcast ops
   bool isBroadcastTvOp(const Expr* expr) {
     if (!ir_utils::isTvOp(expr)) {
       return false;
--- a/csrc/device_lower/pass/alias_memory.cpp
+++ b/csrc/device_lower/pass/alias_memory.cpp
@@ -1344,8 +1344,8 @@ class ReusableAllocationFinder : private kir::IrVisitor {
           continue;
         }
         if (!ir_utils::isPointwiseTvOp(tv_def) &&
-            !ir_utils::isReductionTvOp(tv_def)) {
-          if (isBroadcastExpandTvOp(tv_def)) {
+            !ir_utils::isReductionTvOp(tv_def) && !tv_def->isA<ExpandOp>()) {
+          if (isBroadcastTvOp(tv_def)) {
             info.has_broadcast_between = true;
           } else {
             info.has_unsupported_op = true;
@@ -1386,11 +1386,11 @@ class ReusableAllocationFinder : private kir::IrVisitor {
   }
 
   // Utility to capture broadcast and expand ops
-  bool isBroadcastExpandTvOp(const Expr* expr) {
+  bool isBroadcastTvOp(const Expr* expr) {
     if (!ir_utils::isTvOp(expr)) {
       return false;
     }
-    return expr->isA<BroadcastOp>() || expr->isA<ExpandOp>();
+    return expr->isA<BroadcastOp>();
   }
 
  private:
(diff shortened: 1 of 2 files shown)


liqiangxl:
!build --diff --diff-bench

liqiangxl:
The change of `Allow non-vectorized tv to reuse vectorized tv` added explicity register alias in some of the matmul [cases](https://dl.gitlab-master-pages.nvidia.com/pytorch/fuser-gh-mirror//nvfuser_github_ci/codegen_diff_p11711578_j78194308_1703573108667311342_codediff_a3d1e61_4acaf81_custom_command_20231226_055718.html), and `Softmax_Inner_fp32`, see codegen diffs.

naoyam:
Is the alignment check necessary for smem? I thought we always align smem allocations at 16B boundaries.

liqiangxl:
> Is the alignment check necessary for smem? I thought we always align smem allocations at 16B boundaries.

You are right. We can skip this check for smem since the largest vectorization is 16 Bytes.

csrc/device_lower/pass/alias_memory.cpp
```diff
@@ -1202,21 +1202,19 @@ class ReusableAllocationFinder : private kir::IrVisitor {
...
+          // Vectorized allocations require correct alignment so if [this_tv]
+          // is vectorized, the [reuse_tv] must be vectorized with the same
+          // factor.
+          // No need to check shared memory since it is always aligned to 16
+          // Bytes which is also the maximum vectorization width.
+          if (this_tv->getMemoryType() == MemoryType::Local) {
+            const auto& va = GpuLower::current()->vectorizedAccesses();
+            if ((va.find(this_tv) != va.end())) {
+              if ((va.find(reuse_tv) == va.end()) ||
+                  va.at(this_tv) != va.at(reuse_tv)) {
```
naoyam, line 1215: Let's make the code simpler and more obvious:

```suggestion
bool this_tv_vectorized = va.find(this_tv) != va.end();
if (this_tv_vectorized) {
  bool reuse_tv_vectorized = va.find(reuse_tv) != va.end();
  if (!reuse_tv_vectorized) {
   return false;
  }
  int this_tv_alignment = va.at(this_tv);
  int reuse_tv_alignment = va.at(reuse_tv);
  if (this_tv_alignment != reuse_tv_alignment) {
    return false;
  }
}
```

Also can we relax the alignment check as below??
```
  if (this_tv_alignment > reuse_tv_alignment) {
    return false;
  }
```

naoyam:
Please add unit tests

naoyam:
Also, it would be good if we could have better protection or error detection from the wrong estimation of smem usage. I understand that we need to estimate the smem usage before actually scheduling a fusion, but can we actually validate the estimate after the scheduling (but before lowering)?

csrc/device_lower/pass/alias_memory.cpp
liqiangxl: changed.

liqiangxl:
> Please add unit tests

added 2 tests, one for smem reuse, the other for register reuse with varied vectorization factors.

liqiangxl:
> estimation of smem usage.

smem/register reuse before scheduling is only based on lifescope analysis. However the actual reuse in `reuseMemoryAllocations` depends on how the fusion is actually scheduled. It seems difficult if possible to validate the estimated memory usage before lowering. Maybe create another issue?

liqiangxl:
> Also, it would be good if we could have better protection or error detection from the wrong estimation of smem usage. I understand that we need to estimate the smem usage before actually scheduling a fusion, but can we actually validate the estimate after the scheduling (but before lowering)?

smem/register reuse before scheduling is only based on lifescope analysis. However the actual reuse in reuseMemoryAllocations depends on how the fusion is actually scheduled. It seems difficult if possible to validate the estimated memory usage before lowering. Maybe create another issue?

test/test_smem_reuse.cpp
```diff
@@ -511,4 +511,122 @@ TEST_F(SmemReuseTest, SkippedSyncInterval) {
...
+    testValidate(fusion.get(), cg_outputs, {t0}, __LINE__, __FILE__);
+  };
+
+  // test different vectorization factors, max is 4 since float is used.
+  for (int vect_factor_1 : {1, 2, 4}) {
+    for (int vect_factor_2 : {1, 2, 4}) {
+      testRegisterReuse(vect_factor_1, vect_factor_2);
+    }
+  }
+}
```
jacobhinkle, line 631: Can we add a test with an `ExpandOp` that interferes with re-use here?

jacobhinkle (commented):
The changes make sense to me. `ExpandOp` is being grouped with `BroadcastOp` here, but I think it's really a no-op wrt alias analysis since it cannot introduce broadcasts on its own. I could be wrong but my understanding is that `ExpandOp` really is just constraining the extents of other `TensorView`s that might resolve a broadcast, but I don't think it actually changes the indexing of its argument, so it should basically be ignored in this alias analysis.

naoyam:
> > Also, it would be good if we could have better protection or error detection from the wrong estimation of smem usage. I understand that we need to estimate the smem usage before actually scheduling a fusion, but can we actually validate the estimate after the scheduling (but before lowering)?
> 
> smem/register reuse before scheduling is only based on lifescope analysis. However the actual reuse in reuseMemoryAllocations depends on how the fusion is actually scheduled. It seems difficult if possible to validate the estimated memory usage before lowering. Maybe create another issue?

Hmm, we already know which tensors are cached on the shared memory, so isn't it just checking their leaf domains and the computeAt positions?

liqiangxl:
> > > Also, it would be good if we could have better protection or error detection from the wrong estimation of smem usage. I understand that we need to estimate the smem usage before actually scheduling a fusion, but can we actually validate the estimate after the scheduling (but before lowering)?
> > 
> > 
> > smem/register reuse before scheduling is only based on lifescope analysis. However the actual reuse in reuseMemoryAllocations depends on how the fusion is actually scheduled. It seems difficult if possible to validate the estimated memory usage before lowering. Maybe create another issue?
> 
> Hmm, we already know which tensors are cached on the shared memory, so isn't it just checking their leaf domains and the computeAt positions?

If just check leaf domains and the computeAt positions, we can do that after schedule. I am not sure if this is equivalent to the checks in reuseMemoryAllocations. 
I added `validateSmemPersistentBufferSize` at the end of `schedulePersistentKernel`, It works as follows:
```
// If there are multiple persistent buffers
// and not all of them are active at the same program point,
// It indicates shared memory reuse and needs to validate the reuse.
// Find the program point where the number of active buffers is maximum,
// then check the reuse of inactive buffers by active buffers.
// The reuse pair must have the same leaf domains (except for vectorization)
// and compute at position.
```

test/test_smem_reuse.cpp
liqiangxl: added

csrc/device_lower/pass/alias_memory.cpp
```diff
@@ -1341,7 +1345,7 @@ class ReusableAllocationFinder : private kir::IrVisitor {
         }
         if (!ir_utils::isPointwiseTvOp(tv_def) &&
             !ir_utils::isReductionTvOp(tv_def)) {
```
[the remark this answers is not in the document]
liqiangxl: changed.

naoyam:
> > > > Also, it would be good if we could have better protection or error detection from the wrong estimation of smem usage. I understand that we need to estimate the smem usage before actually scheduling a fusion, but can we actually validate the estimate after the scheduling (but before lowering)?
> > > 
> > > 
> > > smem/register reuse before scheduling is only based on lifescope analysis. However the actual reuse in reuseMemoryAllocations depends on how the fusion is actually scheduled. It seems difficult if possible to validate the estimated memory usage before lowering. Maybe create another issue?
> > 
> > 
> > Hmm, we already know which tensors are cached on the shared memory, so isn't it just checking their leaf domains and the computeAt positions?
> 
> If just check leaf domains and the computeAt positions, we can do that after schedule. I am not sure if this is equivalent to the checks in reuseMemoryAllocations. I added `validateSmemPersistentBufferSize` at the end of `schedulePersistentKernel`, It works as follows:
> 
> ```
> // If there are multiple persistent buffers
> // and not all of them are active at the same program point,
> // It indicates shared memory reuse and needs to validate the reuse.
> // Find the program point where the number of active buffers is maximum,
> // then check the reuse of inactive buffers by active buffers.
> // The reuse pair must have the same leaf domains (except for vectorization)
> // and compute at position.
> ```

Oh, I see. Yeah, you're right, the actual used size depends on the reuse logic, which is still unknown at the point of the scheduling.

Thanks for adding the validation routine, but it doesn't seem add much value as the check is mostly just about what the schedule does. Let's not add the check for now.

naoyam:
The PR almost looks good to me. See my previous comment (https://github.com/NVIDIA/Fuser/pull/1561#issuecomment-1874973026). I'll let @jacobhinkle give a final stamp.

liqiangxl:
> validate reuse after schedule before lowering



> > > > > Also, it would be good if we could have better protection or error detection from the wrong estimation of smem usage. I understand that we need to estimate the smem usage before actually scheduling a fusion, but can we actually validate the estimate after the scheduling (but before lowering)?
> > > > 
> > > > 
> > > > smem/register reuse before scheduling is only based on lifescope analysis. However the actual reuse in reuseMemoryAllocations depends on how the fusion is actually scheduled. It seems difficult if possible to validate the estimated memory usage before lowering. Maybe create another issue?
> > > 
> > > 
> > > Hmm, we already know which tensors are cached on the shared memory, so isn't it just checking their leaf domains and the computeAt positions?
> > 
> > 
> > If just check leaf domains and the computeAt positions, we can do that after schedule. I am not sure if this is equivalent to the checks in reuseMemoryAllocations. I added `validateSmemPersistentBufferSize` at the end of `schedulePersistentKernel`, It works as follows:
> > ```
> > // If there are multiple persistent buffers
> > // and not all of them are active at the same program point,
> > // It indicates shared memory reuse and needs to validate the reuse.
> > // Find the program point where the number of active buffers is maximum,
> > // then check the reuse of inactive buffers by active buffers.
> > // The reuse pair must have the same leaf domains (except for vectorization)
> > // and compute at position.
> > ```
> 
> Oh, I see. Yeah, you're right, the actual used size depends on the reuse logic, which is still unknown at the point of the scheduling.
> 
> Thanks for adding the validation routine, but it doesn't seem add much value as the check is mostly just about what the schedule does. Let's not add the check for now.

The commit adding validation routine is being removed.

liqiangxl:
!build --diff --diff-bench

csrc/device_lower/pass/alias_memory.cpp
```diff
@@ -1381,7 +1385,7 @@ class ReusableAllocationFinder : private kir::IrVisitor {
     }
   }
 
-  // Utility to capture broadcast ops
+  // Utility to capture broadcast and expand ops
```
jacobhinkle, line 1388:
```suggestion
  // Utility to capture broadcast ops
```

jacobhinkle (approved):
LGTM!

PR #1569, opened 2023-12-31
IdModel: Add an initial loop graph (no promotion yet)

naoyam:
This PR adds an initial loop graph. It should be effectively equivalent to the current ComputeAt loop map except that non-leaf domains are also mapped. That may not be strictly necessary but is useful for subsequent loop promotion analysis, which will come after this PR.

I did not add any validation as we already validate the permissive graph and this initial loop graph is pretty straightforward sub graph of the permissive graph.


naoyam:
!build

naoyam:
!build

naoyam:
!build

csrc/val_graph.h
```diff
@@ -120,6 +120,19 @@ class ValGraph {
 
   bool hasUses(const ValGroup& val_group) const;
 
+  // Supports one to many mappings, uses the Valgraph to produce
+  // mappings between from and to. If multiple Vals in to map
+  // to a single Val in from, the order of the Vals in value of
+  // the map is preserved to be the order provided in to.
+  std::unordered_map<Val*, VectorOfUniqueEntries<Val*>> buildMapBetween(
```
zasdfgbnm, line 127: Am I understanding correct that, if my valgraph is:
```C++
{I1, I2, I3, I4}
{I5, I6, I7, I8}
```
my `from` is `{I1, I2, I8}`, my `to` is `{I2, I3, I7}`, then I will get
```
I1 -> {I2, I3}
I2 -> {I2, I3}
I8 -> {I7}
```
?

It would be nice if you can add an example like this, otherwise it is hard to tell what this function is doing without reading its code.

zasdfgbnm (approved):

csrc/val_graph.h
naoyam: Yes, that is correct. I'll add some example. Thanks for the suggestion.

PR #1571, opened 2024-01-02
A test that surrounds a sharding expression with `segment_set`s.

wujingyue:
As a follow-up to our previous meeting about segmentation for multi-device, I wrote this PR showing how to surround a sharding expression with `segment_set`s. This way, a sharding expression becomes its own segment without changing the current segmentation algorithm. I expect a pre-segmenter pass to add such `segment_set`s. See [`MarkAliasPreparePass`](https://github.com/NVIDIA/Fuser/blob/101530ec09a7eab9937f277715be0e3ecc1478b6/csrc/optimization/mark_aliases_prepare.cpp) for an example. I believe this is good enough for segmenting out CPU-initiated communication.

### DIFF
--- a/test/test_segmentation.cpp
+++ b/test/test_segmentation.cpp
@@ -154,4 +154,56 @@ TEST_F(SegmentationTest, SegmentHintOnNonTerminatingOutput) {
   EXPECT_EQ(runtime->fusionSegments()->groups().size(), 2);
 }
 
+TEST_F(SegmentationTest, EnforceSegmentationByCachingBeforeAndAfter) {
+  auto fusion = std::make_unique<Fusion>();
+  FusionGuard fg(fusion.get());
+
+  TensorView* tv0 = makeContigTensor(2);
+  TensorView* tv1 = sum(tv0, {0});
+  TensorView* tv2 = div(tv0, tv1);
+  fusion->addInput(tv0);
+  fusion->addOutput(tv2);
+
+  // A fake proxy for the real isSharding check.
+  auto is_sharding = [](Expr* expr) -> bool {
+    return expr->isA<ReductionOp>();
+  };
+
+  // I'd put this in a pre-segmenter pass.
+  std::vector<Expr*> sharding_exprs;
+  for (Expr* expr : fusion->exprs()) {
+    if (is_sharding(expr)) {
+      sharding_exprs.push_back(expr);
+    }
+  }
+  for (Expr* sharding_expr : sharding_exprs) {
+    for (TensorView* in_tv :
+         ir_utils::filterByType<TensorView>(sharding_expr->inputs())) {
+      if (!in_tv->isFusionInput()) {
+        in_tv->cacheBefore(LoadStoreOpType::SegmenterSet);
+      }
+    }
+    for (TensorView* out_tv :
+         ir_utils::filterByType<TensorView>(sharding_expr->outputs())) {
+      if (!out_tv->isFusionOutput()) {
+        out_tv->cacheAfter(LoadStoreOpType::SegmenterSet);
+      }
+    }
+  }
+
+  FusionExecutorCache fec(std::move(fusion));
+  at::Tensor in_tensor = at::randn({2, 3}).cuda();
+  std::vector<at::Tensor> out_tensors = fec.runFusionWithInputs({in_tensor});
+  testValidate(
+      fec.fusion(),
+      out_tensors,
+      {in_tensor},
+      {in_tensor / in_tensor.sum({0})},
+      __LINE__,
+      __FILE__);
+
+  FusionKernelRuntime* runtime = fec.getMostRecentKernelRuntime();
+  EXPECT_EQ(runtime->fusionSegments()->groups().size(), 2);
+}
+
 } // namespace nvfuser


naoyam (approved):
Nice example. Thanks for adding this.

samnordmann:
Thank you very much! That's very interesting, I wasn't aware of the `Segment_Set` op type. 

I think this way of doing is simple and has the advantage of leaving the segmenter untouched.
However, I think in the future it won't be sufficient and we will need to modify the segmenter anyway (when we want to merge communications together, or comms + compute). So, later we will need to build upon the modifications I proposed.

Another minor drawback of this method is that it will add a lot of trivial Exprs (i.e. Segmenter_Set) and Tvs which need to be discarded so it may make the analysis more cumbersome.
What do you think ?

In a near future, I will open a PR with the segmentation for multidevice, we will have to decide what method to use.
Which one do you think is preferable?

samnordmann (approved):

wujingyue:
> However, I think in the future it won't be sufficient and we will need to modify the segmenter anyway (when we want to merge communications together, or comms + compute)

The current implementation of segments heavily assumes that one segment maps to one CUDA kernel, and we don't nest one kernel into another. So I prefer not making segments hierarchical at this moment. This saves all of us development and maintenance. 

That being said, I agree with you that we'll probably need to represent an expression group better without assuming how it'll be translated and executed. Are you back from your vacation? Can we talk about more about the future use cases next week? I'm very curious about use cases of merging communications themselves and merging communications and compute. We should definitely get nvFuser to support them. 

> Another minor drawback of this method is that it will add a lot of trivial Exprs (i.e. Segmenter_Set) and Tvs which need to be discarded so it may make the analysis more cumbersome.

Yes. That's usually solved by making analysis handle `Segmenter.Set` and/or by running a clean-up pass immediately after segmentation (because `Segmenter.Set` is useless afterwards).

wujingyue:
Here's the summary of today's meeting. 

We (@cowanmeg , @naoyam , @samnordmann , and @wujingyue) talked about the current state of [@samnordmann 's draft nvFuser fork](https://github.com/samnordmann/Fuser) for multi-device and why certain things are implemented in certain ways. This is mainly to help me understand how segmentation needs to be improved for the multi-device use case. 

Facts about the code: 
1. [MultiDeviceExecutor](https://github.com/samnordmann/Fuser/blob/main/csrc/multidevice/executor.h#L25) (instead of FusionExecutorCache or FusionExecutor) is the entry point of compilation and execution, for multi-device. 
2. MultiDeviceExecutor does a [partial segmentation](https://github.com/samnordmann/Fuser/blob/main/csrc/multidevice/executor.cpp#L102), which segments only sharding ops. 
3. Sharding-op-only segments are compiled and executed by [postCommunication](https://github.com/samnordmann/Fuser/blob/main/csrc/multidevice/executor.cpp#L181) and other segments by [postKernel](https://github.com/samnordmann/Fuser/blob/main/csrc/multidevice/executor.cpp#L144). The current implementation of postKernel has a bug where it uses FusionExecutor (not FusionExecutorCache) to compile and execution a fusion. This will fail for some large fusions that have to be segmented. 

My takeaways: 
1. Device meshes are at this moment annotated by the user of nvFuser. We will want annotation to be done automatically. However, this will need to be done before segmentation and therefore not by (the current) scheduling, because device meshes affect segmentation decisions. 
2. I don't see a case for nested `SegmentedGroup`s. A flat DAG of `SegmentedGroup`s suffices for both CPU-initiated and GPU-initiated multi-device communication. 
3. We **may** need a multi-level segmentation. The current implementation applies segmentation in two levels (as mentioned above about the code). I **think** we could avoid even that for simplicity. One idea floating around in my head is to create a `CommunicationScheduler` to accept sharding-op-only Fusions and to let lowering lower such a fusion to a vector of communications. This way, we can reuse Executor for multi-device execution. Also, this plays better with GPU-initiated communication, which has to be lowered to code inside a CUDA kernel.

naoyam:
If we do a multi-level segmentation, wouldn't it make sense to have `SegmentedGroup`s to be further segmented? Wouldn't it make sense to have some form of nested groups?

wujingyue:
> If we do a multi-level segmentation, wouldn't it make sense to have SegmentedGroups to be further segmented? 

Yes, it's done by [calling `makeFusion`](https://github.com/samnordmann/Fuser/blob/main/csrc/multidevice/executor.cpp#L170) and then potentially calling `FusionExecutorCache::runFusionWithInputs`. 

> Wouldn't it make sense to have some form of nested groups?

I doubt that. What's the practical benefit? Do we need to traverse from a parent SegmentedGroup to a child or from child to parent?

naoyam:
> > Wouldn't it make sense to have some form of nested groups?
> 
> I doubt that. What's the practical benefit? Do we need to traverse from a parent SegmentedGroup to a child or from child to parent?

I don't have any concrete use case, but I'm just feeling that since there's nested structures it seems to make sense to represent the nesting in some way. That said, I also agree that we shouldn't make things overly complicated than necessary.

samnordmann:
Thank you very much @wujingyue ! Very useful discussion and summary

About nested SegmentedGroups, I must say I have the same intuition as Naoya, I think it would be a natural and useful thing to have. If we are thinking of complex pipelinings, we would probably need to represent different segments of the fusion at different scales.

Since we are already using several levels of segmentations, why would it be overcomplicated to use nested SegmentedGroup? For now, we need to transform the SegmentedGroup to a Fusion before segmenting it again, which seems to me unnatural

naoyam:
As I think more about this, I'm getting feeling that separating inter-device and intra-device segmentations as two separate processes may not be what we eventually would like to do. I think that, for a given fusion, how we should distribute tensors is likely to depend on how we could schedule computations within a device. That is, the intra-device segmentation/scheduling would affect the intra-kernel segmentation. While there's certainly conceptual hierarchy between the two segmentations, in reality the hierarchy may be blurred as we may need to think about the inter and intra device segmentations as a whole, so the enforcing the hierarchy between the two may not make much sense.

PR #1575, opened 2024-01-03
Prevent replacing constant extents with non-const in concretization

### LINKED ISSUE #1572: `lessThan` ran into infinite recursion.
```
$ git fetch origin wjy/recurse
$ git checkout wjy/recurse
$ _bn && NVFUSER_DISABLE=parallel_compile python repro.py
```

`NVFUSER_DISABLE=parallel_compile` is there to display the call stack better. 

To make debugging easier, I limited recursion of `lessEqual` to 100 levels. The reproducer is still valid but slower with that check removed.

jacobhinkle:
Fixes #1572

In that case, `concretizeReshape` is replacing some extents in the new shape, such as `(nvfuser_index_t)16` with non-constant values like `i0`. This leads to a downstream expression that originally was constant `0` to be non-constant, so it cannot be picked up by `RemoveEmptyPass`. This PR avoids concretizing constant extents with non-constants. This should be safe since even if those tensors are segmentation edges, we will not lose any scalars because the extent is constant so there are no free scalars to lose.

### DIFF
--- a/test/test_dynamic_transform.cpp
+++ b/test/test_dynamic_transform.cpp
@@ -1295,4 +1295,51 @@ TEST_F(NVFuserTest, SymbolicSqueeze) {
           " must concretize to IterType::Broadcast but found")));
 }
 
+// Test that constant zero extents are not overwritten during concretization
+// with non-constant extents.
+// See https://github.com/NVIDIA/Fuser/issues/1572
+TEST_F(NVFuserTest, ConcretizeConstantExtents) {
+  std::unique_ptr<Fusion> fusion_ptr = std::make_unique<Fusion>();
+  Fusion* fusion = fusion_ptr.get();
+  FusionGuard fg(fusion);
+
+  auto tv0 = makeSymbolicTensor(2);
+  fusion->addInput(tv0);
+
+  // Explicitly cast Int to Index, so that these extents are not immediate
+  // constants
+  auto tv1 = reshape(
+      tv0,
+      {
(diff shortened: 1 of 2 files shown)


wujingyue (approved):
LGTM - pending a unit test. Thanks for the quick fix, @jacobhinkle

test/test_dynamic_transform.cpp
```diff
@@ -1295,4 +1295,51 @@ TEST_F(NVFuserTest, SymbolicSqueeze) {
...
+          castOp(DataType::Index, IrBuilder::create<Val>(4096, DataType::Int)),
+          castOp(DataType::Index, IrBuilder::create<Val>(128, DataType::Int)),
+      });
+  // Note this slice has zero extent in last dimension. RemoveEmptyPass should
+  // recognize this and replace with full()
+  auto tv5 = slice(tv4, {0, 0, 0}, {32, 4096, 0});
+
+  fusion->addOutput(tv5);
+
+  FusionExecutorCache fec(std::move(fusion_ptr));
```
jacobhinkle, line 1334: Without the change to `dynamic_transform.cpp` this hangs with infinite recursion due to contradiction `0 < blockDim.x < 0`.

jacobhinkle:
!build --diff

jacobhinkle:
> LGTM - pending a unit test. Thanks for the quick fix, @jacobhinkle

Happy to help! Thanks for finding this subtle bug.

xwang233:
!build --diff

jacobhinkle:
!build --diff

PR #1576, opened 2024-01-03
Check for obvious contradicting assumptions. 

wujingyue:
Helps #1572

### DIFF
--- a/csrc/expr_simplifier.cpp
+++ b/csrc/expr_simplifier.cpp
@@ -191,6 +191,8 @@ class Context {
     }
 
     // Check for obvious contradictions. Could have helped debugging #1572.
+    // Let me know if there's a subquadratic way to check this. `Val*` equality
+    // was't sufficient, so I had to call `sameAs` instead.
     for (const auto& [a, b] : getKnownLessThan()) {
       for (const auto& [x, y] : getKnownLessEqual()) {
         // a < b && b <= a is impossible.
--- a/csrc/expr_simplifier.cpp
+++ b/csrc/expr_simplifier.cpp
@@ -23,7 +23,6 @@
 #include <memory>
 #include <numeric>
 #include <regex>
-#include <set>
 #include <sstream>
 #include <string>
 #include <unordered_set>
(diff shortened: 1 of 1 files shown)


wujingyue:
!build

csrc/expr_simplifier.cpp
```diff
@@ -188,6 +189,26 @@ class Context {
         assumptions.push_back(bop->rhs());
       }
     }
+
+    // Check for obvious contradictions. Could have helped debugging #1572.
+    // Let me know if there's a subquadratic way to check this. `Val*` equality
+    // was't sufficient, so I had to call `sameAs` instead.
+    for (const auto& [a, b] : getKnownLessThan()) {
+      for (const auto& [x, y] : getKnownLessEqual()) {
+        // a < b && b <= a is impossible.
```
jacobhinkle, line 198: We could add an inner loop to check `a < b && b < a`.

csrc/expr_simplifier.cpp
```diff
@@ -23,6 +23,7 @@
 #include <memory>
 #include <numeric>
 #include <regex>
+#include <set>
```
jacobhinkle, line 26: Is this needed?

jacobhinkle (commented):
Makes sense to me, but I'll let @zasdfgbnm stamp. Note we may want to factor this into a `bool isConsistent() const` method.

xwang233:
!build

zasdfgbnm (commented):
Thanks for fixing this! Other than Jacob's comment, this PR makes sense to me. Will let Jacob stamp.

wujingyue:
!build

csrc/expr_simplifier.cpp
wujingyue: Done.

csrc/expr_simplifier.cpp
wujingyue: Removed.

jacobhinkle (approved):
LGTM!

csrc/expr_simplifier.cpp
```diff
@@ -188,6 +188,45 @@ class Context {
         assumptions.push_back(bop->rhs());
       }
     }
+
+    validateConsistency();
+  }
+
+  void validateConsistency() {
```
jacobhinkle, line 195: Noticed a nit too late: `const` :-D.

PR #1578, opened 2024-01-03
Print fusion IR before scheduling.

wujingyue:
Guarded by NVFUSER_DUMP=fusion_ir_presched. The printing in `FusionDefinition::finalizeDefinition` is guarded instead by `NVFUSER_DUMP=fusion_ir_original` to avoid confusion.


csrc/python_frontend/fusion_definition.cpp
```diff
@@ -87,10 +87,6 @@ void FusionDefinition::finalizeDefinition() {
     }
 
     buildFusionIr(preschedFusion());
-
-    if (isDebugDumpEnabled(DebugDumpOption::FusionIrPresched)) {
```
rdspring1, line 91: This prints the original unscheduled fusion generated by the python frontend.

csrc/kernel_cache.cpp
```diff
@@ -1329,6 +1329,9 @@ void FusionKernelRuntime::compileKernel(
   // Running a segment group as a single kernel,
   // make a fusion to run from segmented fusion
   auto fusion_to_run = segmented_fusion_->makeFusion(sg);
+  if (isDebugDumpEnabled(DebugDumpOption::FusionIrPresched)) {
```
rdspring1, line 1332: This prints the concretized and segmented fusion before it is scheduled during compilation.

csrc/python_frontend/fusion_definition.cpp
rdspring1: Perhaps this should be `DebugDumpOption::FusionIrOriginal`

rdspring1 (commented):
It makes sense to move the `FusionIrPresched` dump option to `FusionKernelR`::compileKernel`. I think we should keep the original dump option but rename it to something like FusionIrOriginal`.

csrc/kernel_cache.cpp
rdspring1: Please update the `FusionIrPresched` option comment in https://github.com/NVIDIA/Fuser/blob/main/csrc/options.h so we know that we're printing the concretized, segmented fusion.

kevinstephano:
I think you are asking for something different.  This flag is specifically printing the Fusion IR as it is built based on the Python Definition and nothing more.  It's helpful for debugging the translation between the Python Definition and the Fusion IR prior to it being modified by any step.

csrc/python_frontend/fusion_definition.cpp
wujingyue: I see -- the word "schedule" is overloaded. How about printing `FusionIrOriginal` inside `FusionExecutorCache::runFusionWithInputs`? AFAICT from [code](https://github.com/NVIDIA/Fuser/blob/78343694da4c7795e035363bfbe867d10a3a327a/csrc/python_frontend/fusion_definition.cpp#L186), both Python and C++ go through that method. So this will reduce divergence between the two APIs.

kevinstephano: Sounds reasonable!

wujingyue:
!build

rdspring1 (approved):
LGTM

PR #1582, opened 2024-01-04
Improve debug printing.

wujingyue:
Previously, it was unclear why an output doesn't alias an input even when there's a chain between them. With this PR,
AliasAnalysisResult::toString also prints out the allocation and rfactor domain of each output.

### DIFF
--- a/csrc/alias_analysis.cpp
+++ b/csrc/alias_analysis.cpp
@@ -364,8 +364,7 @@ TensorView* AliasAnalysisResult::findNearestAliasedIo(
 
 TensorView* AliasAnalysisResult::getNearestAliasedIo(
     const TensorView* fusion_out) const {
-  const auto i = out_to_root_.find(fusion_out);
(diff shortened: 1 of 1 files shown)


csrc/alias_analysis.cpp
```diff
@@ -385,8 +384,12 @@ void AliasAnalysisResult::finalize(
     const bool can_override_empty_allocation_domain) {
   for (TensorView* out :
        ir_utils::filterByType<TensorView>(fusion->outputs())) {
-    TensorView* root = findNearestAliasedIo(out);
-    if (root == nullptr) {
+    // Even a non-aliasing output gets a null entry in the map for
+    // toString() to debug print.
+    TensorView*& root = out_to_root_[out];
```
jacobhinkle, line 389: It took me a second to realize `root` will be default initialized here.

jacobhinkle (approved):
LGTM

Priya2698 (approved):

PR #1583, opened 2024-01-04
AliasAnalysis handles BroadcastOp.

wujingyue:


csrc/alias_analysis.cpp
```diff
@@ -332,6 +333,49 @@ void AliasFinder::handle(const SliceOp* slice) {
...
+  Layout in_layout = analysis_.preferredLayout(in);
+  if (!ir_utils::computePermutation(
+           in->getMaybeRFactorDomain(), in_layout.allocation_domain)
+           .has_value()) {
+    // Give up when `in`'s allocation domain is not an rfactor permutation.
+    return;
+  }
+
+  std::unordered_map<IterDomain*, IterDomain*> in_rfactor_to_out_root =
+      PairwiseRootDomainMap(in, out).mapBroadcast(true).mapProducerToConsumer();
```
jacobhinkle, line 353: I don't think you need `mapBroadcast(true)` here since you don't expect any bcast/non-bcast pairings.

jacobhinkle (approved):
LGTM

csrc/alias_analysis.cpp
wujingyue: You are right! I blindly copied it from other places and never tried to really understand how it's used: https://github.com/NVIDIA/Fuser/blob/841a7da804ed849cea9b8f1fd9daaeeb2a276c6e/csrc/root_domain_map.cpp#L181-L186. 

Does that mean the `mapBroadcast(true)` in `handle(ViewOp)` and `handle(LoadStoreOp)` are also unnecessary because these ops don't resize?

jacobhinkle: Yes, `mapBroadcast` would be useful if you want to find a producer broadcast dimension which is resolved in a binary op to produce an `Iteration` output domain, for example.

wujingyue:
!build

PR #1584, opened 2024-01-04
AliasAnalysis handles SqueezeOp.

wujingyue:

PR #1587, opened 2024-01-05
Factor out code that maps input layout to output root.

wujingyue:
This is a follow-up to https://github.com/NVIDIA/Fuser/pull/1584#discussion_r1441862342. 

This also comes with a refactor to `mergeContiguity` so `handle(ViewOp)` can use `mapInLayoutToOutRoot`. I can split that out if you find it hard to review.

### DIFF
--- a/csrc/alias_analysis.cpp
+++ b/csrc/alias_analysis.cpp
@@ -296,17 +296,16 @@ void AliasFinder::handle(const SliceOp* slice) {
   }
 
   // Inherit the allocation order from the input. However, refine the
-  // contiguity flags.
-  // Scan through the allocation domain in minor-to-major order. If an
-  // IterDomain is sliced, the next non-broadcast IterDomain has to be marked
-  // non-contiguous. For example,
+  // contiguity flags. This is done by scanning through the allocation domain in
+  // minor-to-major order. If an IterDomain is sliced, the next non-broadcast
+  // IterDomain has to be marked non-contiguous. For example,
   //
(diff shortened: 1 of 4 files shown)


wujingyue:
!build

csrc/alias_analysis.cpp
```diff
@@ -113,114 +128,123 @@ std::pair<bool, std::optional<bool>> mergeContiguity(
...
   LinkedHashMap<IterDomain*, std::optional<bool>> allocation_to_contiguity;
-  for (const auto i : c10::irange(in_layout.allocation_domain.size())) {
-    IterDomain* in_allocation_id = in_layout.allocation_domain[i];
-    if (in_allocation_id->isReduction()) {
-      // Reduction IterDomains won't appear in `out_root`.
-      continue;
-    }
+  for (const auto i : c10::irange(out_root_layout->size())) {
     allocation_to_contiguity.pushBack(
-        in_allocation_id, in_layout.contiguity[i]);
```
wujingyue, line 146: Context: to work around #1174, `allocation_to_contiguity` may contain IterDomains both from the input TV and from the output. This however is unnecessary. The new code only puts output IterDomains in `allocation_to_contiguity` and looks up `out_root_to_in_rfactor` for expanded extents. It has the equivalent behavior but is more intuitive and doesn't need the `in_rfactor_to_out_root` map.

Priya2698 (approved):
LGTM

PR #1593, opened 2024-01-06
Update validation constants

Priya2698:
The updated validation constants were generated on `A100-asus` node on `dlcluster`.

**Generation:** The computation `(a * b).sum()` was repeated `10^8` times and the base threshold was set to the maximum error observed between the output in the desired dtype and the next higher precision. The base threshold was multiplied by a safety factor of 3 to obtain the final thresholds.

```
import numpy as np
import torch

sizes = [2**i for i in range(1, 22)] # {2, 2097152}

def compute_max_error(size, dtype, ref_dtype):
    a, b = [torch.randn(size, dtype=dtype, device="cuda") for _ in range(2)]
    out = (a * b).sum()
    ref = (a.to(ref_dtype) * b.to(ref_dtype)).sum()
    max_error = out.sub(ref).abs().max().to(torch.double).item()
    return max_error

validation_consts = {}
dtype_to_ref_dtypes = {torch.float16: torch.float32, torch.float32: torch.float64}
for dtype, ref_dtype in dtype_to_ref_dtypes.items():
    tolerances = {}
    for size in sizes:
        print(f'size:{size}')
        val_tolerance = torch.finfo(torch.double).min
        for i in range(10**8):
            val_tolerance = max(val_tolerance, compute_max_error(size, dtype, ref_dtype))
        tolerances[size] = val_tolerance
    validation_consts[dtype] = tolerances
```
The relevant files are present at: https://gitlab-master.nvidia.com/prmishra/scripts.

The graphs show the existing threshold (blue), the new base threshold (orange), and the final proposed threshold (green).

float16
float32
bfloat16


wujingyue:
Can you include also the generation script? In the future, people may want to reproduce the results or debug divergence if the threshold no longer works.

naoyam:
Shouldn't the ref type be always float64?

> out = (a * b).sum()

Is this done in fp16 when `dtype` is fp16? Or is it automatically converted to `float`?

Priya2698:
> Shouldn't the ref type be always float64?
@csarofeen generated the original validation constants using the next higher precision as reference.
> 
> > out = (a * b).sum()
> 
> Is this done in fp16 when `dtype` is fp16? Or is it automatically converted to `float`?
The computation is done in fp16 and the reference computation in fp32.

naoyam:
> > Shouldn't the ref type be always float64?
> 
> @csarofeen generated the original validation constants using the next higher precision as reference, so right now using that. We will likely have smaller max errors (tighter tolerances) for fp16 when compared with fp32 than fp64.

Why smaller errors? FP64 should have more stable results with less variance, so shouldn't FP64 provide smaller errors?

> 
> > > out = (a * b).sum()
> > 
> > 
> > Is this done in fp16 when `dtype` is fp16? Or is it automatically converted to `float`?
> 
> The computation is done in fp16 and the reference computation in fp32.

OK, thanks. Should we do the same with bf16 too?

Priya2698:
> > > Shouldn't the ref type be always float64?
> > 
> > 
> > @csarofeen generated the original validation constants using the next higher precision as reference, so right now using that. We will likely have smaller max errors (tighter tolerances) for fp16 when compared with fp32 than fp64.
> 
> Why smaller errors? FP64 should have more stable results with less variance, so shouldn't FP64 provide smaller errors?
>

You're right, I miswrote. Tolerances will be smaller when compared with fp64. Since we are updating all the constants, we can switch to using fp64 as the reference in all cases. @csarofeen was there any particular reason why you chose to use the next higher precision as reference?

> > > > out = (a * b).sum()
> > > 
> > > 
> > > Is this done in fp16 when `dtype` is fp16? Or is it automatically converted to `float`?
> > 
> > 
> > The computation is done in fp16 and the reference computation in fp32.
> 
> OK, thanks. Should we do the same with bf16 too?
Yes, I'll compute for `bf16` as well.

Priya2698:
!build

naoyam (approved):
LGTM.

Doesn't need to be part of this PR, but please add the script to the nvfuser repository.

Priya2698:
@xwang233 helped me run the python benchmarks with this PR but we are still hitting some validation errors. The errors are only with `fp32`. Should we bump up the safety factor (currently 3) for `fp32`?

PR #1602, opened 2024-01-09
Attention benchmarks

Priya2698:
Attention benchmarks:

- HuggingFace Attention: https://github.com/Lightning-AI/lightning-thunder/blob/main/thunder/tests/hf_bart_self_attn.py#L73-L83

- NanoGpt attention: https://github.com/Lightning-AI/lightning-thunder/blob/d3da8517bff02a913fd149b4d6559f6b5a4c6c7f/thunder/tests/nanogpt_model.py#L102-L106

The current input sizes for attention benchmarks are the actual LLM configs. The batch size and seq lengths are set according to maximum size that fits on a GPU.

Question: Are there other sizes of interest for the attention benchmarks?


python_benchmarks/test_huggingface_attn_fwd.py
```diff
@@ -0,0 +1,105 @@
...
+from nvfuser import FusionDefinition, DataType
+from nvfuser.pytorch_utils import torch_dtype_to_nvfuser_dtype
+from .core import run_benchmark, clear_cuda_cache
+import torch
+from .global_params import generate_attn_inputs, FLOAT_DTYPES, PROMOTE_DTYPES
+
+
+# Fusion from huggingface attention implementation
+# https://github.com/Lightning-AI/lightning-thunder/blob/main/thunder/tests/hf_bart_self_attn.py#L73-L83
+def huggingface_attn_fwd_fusion(
```
wujingyue, line 11: The name suggests it's the full MHA module, but the code seems to be just the non-matmul portion of it? Maybe change the name so readers don't get the wrong impression.

wujingyue (approved):

python_benchmarks/test_huggingface_attn_fwd.py
Priya2698: Right. The matmul portion does not show up in the nvfuser definitions.

PR #1609, opened 2024-01-10
Set device index during deserialization.

rdspring1:
This PR sets the `CompileOptions` in the `FusionExecutor` during deserialization to the appropriate device. It fixes errors encountered in a distributed setting.

### DIFF
--- a/python_tests/test_python_frontend.py
+++ b/python_tests/test_python_frontend.py
@@ -111,7 +111,9 @@ class TestNvFuserFrontend(TestCase):
     # definition based on the FusionDefinition is executable and matches the
     # original definition
     @serde_check
-    def exec_nvfuser(self, fusion_func, inputs, *, new_fusion_expected=True):
+    def exec_nvfuser(
+        self, fusion_func, inputs, *, new_fusion_expected=True, device=None
+    ):
         inputs_cap = deepcopy(inputs)
         fc = FusionCache.get()
         before_fusions = fc.num_fusions()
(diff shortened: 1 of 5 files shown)


rdspring1:
!build

jjsjann123 (commented):
I'm imagining you have different devices on machine which triggers the error here?!

I don't know if we have CI nodes with heterogeneous GPUs, but it would still be a good idea to put in some test the would replicate the failure on your local setup.

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
I modified `exec_nvfuser` to accept a device argument, so `test_selected_device` will test serialization with multiple devices.

python_tests/test_python_frontend.py
```diff
@@ -2234,23 +2236,24 @@ def schedule(self):
     )
     def test_selected_device(self):
         """
-        Run the same Fusion as in test_scalar_only_inputs, but on device 1
+        Run the Fusion on device 1
         """
+        inputs = [
```
jjsjann123, line 2241: bear with my ignorance, why do we need to change it to tensors here instead of just using scalar inputs?

rdspring1: Previously, this test was a copy of `test-scalar-inputs`. The `exec-nvfuser` function checks if a new fusion is generated. There isn't a way to check if `test-scalar-inputs` is run prior to this test, so the new-fusion-generated check may fail. An easy fix was to slightly modify the fusion in this test so it is unique.

jjsjann123: got'ya. when seeing that we switched from scalar to tensor, I was under the impression that the device index here might not be tested as expected... I'm paranoid like I usually is :)

jjsjann123 (approved):

rdspring1:
!build

Issue #1607, opened 2024-01-10
`Cannot find scheduler` in `layernorm + dropout + add` benchmarks


Priya2698:
The following error occurs for several input sizes in PR #1606.

Repro:
```
ERROR    nvfuser:__init__.py:196 An error occurred while executing nvFuser FusionDefinition 0.
If you believe this is a bug or need assistance, please file an issue at https://github.com/NVIDIA/Fuser/issues/new
Here's a script to reproduce the error:


import torch
from nvfuser import FusionDefinition, DataType

def nvfuser_fusion_id0(fd : FusionDefinition) -> None :
    T0 = fd.define_tensor(shape=[-1, -1], contiguity=[True, True], dtype=DataType.Bool, is_cpu=False)
    T1 = fd.define_tensor(shape=[-1], contiguity=[True], dtype=DataType.Float, is_cpu=False)
    T2 = fd.define_tensor(shape=[-1, 1], contiguity=[True, None], dtype=DataType.Float, is_cpu=False)
    T3 = fd.define_tensor(shape=[-1, -1], contiguity=[True, True], dtype=DataType.Float, is_cpu=False)
    T4 = fd.define_tensor(shape=[-1], contiguity=[True], dtype=DataType.Float, is_cpu=False)
    T5 = fd.define_tensor(shape=[-1, -1], contiguity=[True, True], dtype=DataType.Float, is_cpu=False)
    T6 = fd.ops.mul(T5, T0)
    S7 = fd.define_scalar(1.11111, dtype=DataType.Double)
    T8 = fd.ops.mul(T6, S7)
    T9 = fd.ops.add(T5, T8)
    S10 = fd.ops.size(T5, dim=0)
    S11 = fd.define_scalar(1, dtype=DataType.Int)
    V12 = fd.define_vector([S10, S11], dtype=DataType.Int)
    T13 = fd.ops.broadcast_in_dim(T1, shape=V12, broadcast_dims=[0])
    V14 = fd.ops.shape(T5)
    T15 = fd.ops.broadcast_in_dim(T13, shape=V14, broadcast_dims=[0, 1])
    T16 = fd.ops.sub(T9, T15)
    T17 = fd.ops.broadcast_in_dim(T2, shape=V14, broadcast_dims=[0, 1])
    T18 = fd.ops.mul(T16, T17)
    T19 = fd.ops.broadcast_in_dim(T4, shape=V14, broadcast_dims=[1])
[...]

Issue #1301, opened 2023-11-14
Inferring output shape failure


kiya00:
To reproduce the error:
```
import torch
from nvfuser import FusionDefinition, DataType

def nvfuser_fusion_id1(fd : FusionDefinition) -> None :
    T0 = fd.define_tensor(shape=[-1, -1, -1], contiguity=[True, True, True], dtype=DataType.Float, is_cpu=False)
    T1 = fd.define_tensor(shape=[-1, -1, -1], contiguity=[True, True, None], dtype=DataType.Float, is_cpu=False)
    T2 = fd.define_tensor(shape=[-1, -1, -1], contiguity=[None, None, True], dtype=DataType.BFloat16, is_cpu=False)
    T3 = fd.define_tensor(shape=[-1, -1, -1], contiguity=[True, True, True], dtype=DataType.BFloat16, is_cpu=False)
    T4 = fd.define_tensor(shape=[-1, -1, 1], contiguity=[True, True, None], dtype=DataType.Float, is_cpu=False)
    T5 = fd.define_tensor(shape=[-1, -1, -1], contiguity=[True, True, True], dtype=DataType.BFloat16, is_cpu=False)
    T6 = fd.ops.cast(T5, dtype=DataType.Float)
    S7 = fd.define_scalar(16, dtype=DataType.Int)
    S8 = fd.define_scalar(128, dtype=DataType.Int)
    S9 = fd.define_scalar(768, dtype=DataType.Int)
    V10 = fd.define_vector([S7, S8, S9], dtype=DataType.Int)
    T11 = fd.ops.broadcast_in_dim(T4, shape=V10, broadcast_dims=[0, 1, 2])
    T12 = fd.ops.mul(T0, T11)
    T13 = fd.ops.cast(T2, dtype=DataType.Float)
    T14 = fd.ops.cast(T3, dtype=DataType.Float)
    T15 = fd.ops.cast(T3, dtype=DataType.Float)
    T16 = fd.ops.sum(T15, axes=[0, 1], keepdim=False, dtype=DataType.Null)
    T17 = fd.ops.cast(T16, dtype=DataType.BFloat16)
    T18 = fd.ops.mul(T14, T13)
    T19 = fd.ops.mul(T14, T12)
    T20 = fd.ops.sum(T19, axes=[0, 1], keepdim=False, dtype=DataType.Null)
    T21 = fd.ops.cast(T20, dtype=DataType.BFloat16)
[...]

PR #1616, opened 2024-01-11
only forward inputs with one use to avoid introducing additional persistent buffer

liqiangxl:
See https://github.com/NVIDIA/Fuser/issues/1607

### DIFF
--- a/csrc/fusion_segmenter.cpp
+++ b/csrc/fusion_segmenter.cpp
@@ -3784,11 +3784,8 @@ void SegmentCandidateFinder::forwardInputs() {
       // (1) Fragmentation of the DAG, increased segments, see test in #1301.
       // (2) Miss detection of persistent buffers, see issue #1607.
       const auto& input_uses = inp->uses();
-      if (input_uses.size() != 1) {
-        continue;
-      }
-      // Add the use of input if it is a UnaryOp
-      if (input_uses.at(0)->isA<UnaryOp>()) {
+      // Add single-use input if it is a UnaryOp
+      if (input_uses.size() == 1 && input_uses.at(0)->isA<UnaryOp>()) {
         to_visit.push_back(input_uses.at(0)->as<UnaryOp>());
       }
     }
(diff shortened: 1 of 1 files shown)


liqiangxl:
!build --diff --diff-bench

liqiangxl:
!build --diff --diff-bench

naoyam:
> Just a comment. Looking at [this diff](http://nv/e5M/nvfuser_github_ci/codegen_diff_p12046756_j79498625_1705030929968697353_codediff_7c8b098_931601a_custom_command_20240112_032916.html) it looks like the test for #1301 is reduced from three segments to two with this PR (test is misnamed, oops). The other diffs are just reordering. That test contains three reductions and on main it is producing three separate reduction kernels, two grid reductions and a block reduce, while in this PR both the grid reductions are put into a single reduction kernel which is probably good.

I think that's because stripping out multi-use inputs would make a DAG disjoint, so we would end up having separate segments. 

Another side effect of excluding multi-use inputs is what happened with the original bug, i.e., missing to detect a normalization pattern. 

@liqiangxl This change makes sense to me. Please add more comments to the code why we shouldn't exclude multi-use inputs.

csrc/fusion_segmenter.cpp
```diff
@@ -3779,14 +3779,17 @@ void SegmentCandidateFinder::forwardInputs() {
...
+      // Stripping out multi-used inputs can lead to:
+      // (1) Fragmentation of the DAG, increased segments, see test in #1301.
+      // (2) Miss detection of persistent buffers, see issue #1607.
+      const auto& input_uses = inp->uses();
+      if (input_uses.size() != 1) {
+        continue;
+      }
+      // Add the use of input if it is a UnaryOp
+      if (input_uses.at(0)->isA<UnaryOp>()) {
+        to_visit.push_back(input_uses.at(0)->as<UnaryOp>());
```
jacobhinkle, line 3792:
```suggestion
      // Add single-use input if it is a UnaryOp
      if (input_uses.size() == 1 && input_uses.at(0)->isA<UnaryOp>()) {
        to_visit.push_back(input_uses.at(0)->as<UnaryOp>());
```

jacobhinkle (approved):
LGTM

Issue #1502, opened 2023-12-12
Make concat no-op. 


wujingyue:
While it's in general beneficial for nvFuser to generate a fast concat kernel, sometimes concats can be made a no-op when we allocate tensors carefully. Thanks @tfogal for mentioning this opportunity. 

Here's a pattern from the nanogpt gpt2 grad benchmark: 

```
t6698, t6699, t6700 = scaled_dot_product_attention_backprop(...)  # bf16[16,12,128,64], bf16[16,12,128,64], bf16[16,12,128,64]

t6703 = transpose(t6698, [0, 2, 1, 3])  # [16, 128, 12, 64]
t6704 = reshape(t6703, [16, 128, 768])  # [16, 128, 768]
t6705 = transpose(t6699, [0, 2, 1, 3])  # [16, 128, 12, 64]
t6706 = reshape(t6705, [16, 128, 768])  # [16, 128, 768]
t6701 = transpose(t6700, [0, 2, 1, 3])  # [16, 128, 12, 64]
t6702 = reshape(t6701, [16, 128, 768])  # [16, 128, 768]

t6707 = concat(t6704, t6706, t6702, axis=-1)  # [16, 128, 2304]

t6715 = cast(t6707, f32)  # [16, 128, 2304]
t6716 = sum(t6715, [0, 1])  # [2304]
# Output 0
t6717 = cast(t6716, bf16)  # [2304]
# Output 1
t6708 = view(t6707, [2048, 2304])
# Output 2
t6712 = transpose(t6708, [1, 0])  # [2034, 2048]
```

To me, the hardest piece of the puzzle is buffer allocation. The best outcome is to make the concat's input buffers slices of its output buffer. The current buffer allocation algorithm in `csrc/executor.cpp` follows the execution order, so it has to allocate the concat's input buffers (i.e. producers' output buffers) before the output. 

[...]

PR #1617, opened 2024-01-12
Fix a bug that potentially blocks #1502.

wujingyue:


wujingyue:
!build

csrc/scheduler/reduction.cpp
```diff
@@ -1225,6 +1226,12 @@ void scheduleReduction(Fusion* fusion, const ReductionParams& rparams) {
       cached_outputs);
 
   scheduler_utils::promoteProducerMemoryTypes(fusion, cached_inputs);
+
+  // TODO(#1401): We could let segmentation split a partially alias-producing
+  // fusion into an alias-only segment and the rest. This way, the rest of the
+  // fusion (which has fewer expressions) can potentially find a better
+  // scheduler and we need to call markAliases only in NoOpScheduler.
+  markAliases(fusion);
```
jjsjann123, line 1234: Am I understanding this correctly that `markAliases` should potentially be called in every scheduler? should we have added it in maybe transpose scheduler as well?

wujingyue: Yes! I'm adding it incrementally to reduce the blast radius. I was also hoping (as the comment says) segmentation is improved so only NoOpScheduler needs to mark aliases, but I haven't prioritized that yet without a concrete need.

jjsjann123: Got'ya.

jjsjann123 (approved):

Issue #1570, opened 2024-01-02
ExpressionEvaluator cannot compute `N2at6TensorE != i`. 


wujingyue:
```
$ git fetch origin wjy/reduce_bug
$ git checkout wjy/reduce_bug
$ _bn && bin/nvfuser_tests --gtest_filter=*SumDiv
```

```
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from ExprEvalTest
[ RUN      ] ExprEvalTest.SumDiv
unknown file: Failure
C++ exception with description "false INTERNAL ASSERT FAILED at "/opt/pytorch/nvfuser/lib/dynamic_type/src/dynamic_type/dynamic_type.h":697, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. Cannot compute N2at6TensorE != i : incompatible type
Exception raised from operator!= at /opt/pytorch/nvfuser/lib/dynamic_type/src/dynamic_type/dynamic_type.h:697 (most recent call first):
frame #0: nvfuser::nvfCheckFail(char const*, char const*, unsigned int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) + 0xf3 (0x7fad695e29e3 in /opt/pytorch/nvfuser/build/libnvfuser_codegen.so)
frame #1: nvfuser::nvfErrorFail(char const*, char const*, unsigned int, char const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) + 0x53 (0x7fad69731d53 in /opt/pytorch/nvfuser/build/libnvfuser_codegen.so)
frame #2: bool dynamic_type::operator!=<dynamic_type::DynamicType<dynamic_type::Containers<std::vector>, nvfuser::StructHandle, nvfuser::Pointer, nvfuser::Opaque, at::Tensor, std::complex<double>, double, long, bool>, int, void>(dynamic_type::DynamicType<dynamic_type::Containers<std::vector>, nvfuser::StructHandle, nvfuser::Pointer, nvfuser::Opaque, at::Tensor, std::complex<double>, double, long, bool> const&, int const&) + 0x110 (0x7fad6977ae00 in /opt/pytorch/nvfuser/build/libnvfuser_codegen.so)
[...]

PR #1619, opened 2024-01-12
[DynamicType] Add named comparison ops returning dynamic type

zasdfgbnm:
Fixes https://github.com/NVIDIA/Fuser/issues/1570

This PR adds a variant of comparison operators that returns `DynamicType`. This new variant is not implemented as operator overloading, instead, users need to explicitly call its function. The reason for not overloading it as operators is because it makes more sense to make comparison operators return bool, so that code like
```C++
PolymorphicValue pv = 2;
if (pv > 0) {
  blabla();
}
```
will seamlessly work.

### DIFF
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -585,7 +585,6 @@ std::vector<PolymorphicValue> BinaryOp::evaluate(
       return {lhs * rhs};
       break;
     case BinaryOpType::Div:
-      NVF_CHECK(rhs != 0);
       return {lhs / rhs};
       break;
     case BinaryOpType::Mod:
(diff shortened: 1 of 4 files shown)


zasdfgbnm:
!build

csrc/ir/nodes.cpp
```diff
@@ -585,7 +585,6 @@ std::vector<PolymorphicValue> BinaryOp::evaluate(
       return {lhs * rhs};
       break;
     case BinaryOpType::Div:
-      NVF_CHECK(rhs != 0);
```
zasdfgbnm, line 588: This needs to be removed, because lhs or rhs might be tensors.

zasdfgbnm:
!build

lib/dynamic_type/src/dynamic_type/dynamic_type.h
```diff
@@ -573,18 +573,31 @@ constexpr bool is_dynamic_type_v = is_dynamic_type<T>::value;
...
+DEFINE_BINARY_OP(land, &&, operator&&);
+DEFINE_BINARY_OP(lor, ||, operator||);
+DEFINE_BINARY_OP(lshift, <<, operator<<);
+DEFINE_BINARY_OP(rshift, >>, operator>>);
+
+// Not defining comparison operators that returns DynamicType as operator
+// overloading, because we want to leave the operator overloading for comparison
+// operators that returns bool. Instead, we give each operator a function name,
+// so that users can use the function name to call the operator. That is:
+//   dt1 < dt2 --> returns a bool
```
wujingyue, line 593: (Trust by verify) 

Where is this defined? I only saw `op_name` (thus `lt`) defined above by `DEFINE_BINARY_OP`.

zasdfgbnm: It is defined by `DEFINE_COMPARE_OP`, I will update this PR mentioning that.

wujingyue (approved):

PR #1621, opened 2024-01-12
[WIP] Translate serial grid reduction to kernel IR

jacobhinkle:
This removes the need for special handling of serial grid reductions in codegen.cpp and the `serialReductionStep` runtime helper function. Instead, we translate those expressions into regular IR expressions and let lowering passes handle allocation of the work buffer and indexing.

### DIFF
--- a/csrc/device_lower/pass/grid_serialization.cpp
+++ b/csrc/device_lower/pass/grid_serialization.cpp
@@ -13,6 +13,7 @@
 #include <ir/utils.h>
 #include <kernel_ir.h>
 #include <kernel_ir_dispatch.h>
+#include <ops/arith.h>
 
 #include <unordered_set>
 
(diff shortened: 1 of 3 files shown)


csrc/device_lower/pass/grid_serialization.cpp
```diff
@@ -53,45 +54,156 @@ class GridSerializationSyncInserter : kir::ExprMutator {
...
+    // work_buffer_out is dead code (it has no uses). This means we won't
+    // automatically insert new Allocate nodes for it in subsequent lowering
+    // passes, so it is safe for us to create one here.
+    TensorView* work_buffer_out =
+        result->cacheAfter(LoadStoreOpType::Set, CacheOp::Global);
+    work_buffer_out->setMemoryType(MemoryType::Global);
+    // TODO: we cannot force aliasing work_buffer_out => work_buffer_in yet,
+    // since that is done with Allocate. Since we _do_ want automatic insertion
+    // of work_buffer_in's Allocate, it is unclear how we can effectively set
+    // this alias.
```
jacobhinkle, line 191: @naoyam I think this is the biggest challenge to using IR to represent `serialReductionStep`. We could try and manually set that alias in a later pass, but once we have done this translation, it might be difficult to recognize this as a reduction step.

PR #1622, opened 2024-01-12
Add an tutorial case for demonstrating IdModel for analyzing reshape

naoyam:
See the added test, which demonstrates how IdModel can be used to analyze reshape ops

Tagging @wujingyue, @Priya2698, @csarofeen

### DIFF
--- a/test/test_tutorial.cpp
+++ b/test/test_tutorial.cpp
@@ -9,6 +9,7 @@
 #include <gtest/gtest.h>
 
 #include <fusion.h>
+#include <id_model/id_model.h>
 #include <inlining.h>
 #include <ir/utils.h>
 #include <ops/alias.h>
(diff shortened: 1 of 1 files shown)


naoyam:
@wujingyue @Priya2698 Let me know if it makes sense.

test/test_tutorial.cpp
```diff
@@ -630,4 +631,70 @@ TEST_F(Tutorial, Reshape) {
...
+  // two inputs are the same as there's no op allowing inference of
+  // equivalence (e.g., tv0 + tv1)
+  auto tv2 = reshape(tv0, {10, 20}, {20, 10});
+  auto tv3 = reshape(tv1, {10, 20}, {20, 10});
+
+  fusion.addOutput(tv2);
+  fusion.addOutput(tv3);
+
+  IdModel id_model(&fusion);
+  ValGraph& exact_graph = id_model.idGraph(IdMappingMode::EXACT);
```
wujingyue, line 656: (Just a question; not requesting for changes)

Any IdMappingMode should work for this use case because we map input values explicitly and only rely on IdModel/IdGraph to propagate. Correct?

wujingyue (approved):

Priya2698 (approved):

test/test_tutorial.cpp
naoyam: The other mapping modes such as `PERMISSIVE` and `LOOP` are mostly for analyzing loop-nests and indexing. Those two graphs will have relaxed mappings with broadcast domains to represent inlining of broadcast domains to non-broadcast domains. Since in your use case there's no inlining done yet as the fusion is not scheduled, the other graphs won't be helpful.

There's also `ALMOSTEXACT`, which also relaxes mappings when a domain is merged with a size-1 domain. It shouldn't matter for your use case either.

Or, I think we could also create a new graph from scratch. `IdModel::initializeIdGraph`, which is currently not public but can be changed, can be used to create an empty graph, and then we could add whatever mappings we could assume.

wujingyue: > Or, I think we could also create a new graph from scratch. IdModel::initializeIdGraph, which is currently not public but can be changed, can be used to create an empty graph, and then we could add whatever mappings we could assume.

Yeah, that sounds cleaner. IdModel implicitly maps certain things (between in rfactor and out root), which may come as a surprise to users who expect mappings to be explicit. 

Anyway, don't worry about this yet. Priya or I can do the refactor when we get there...

PR #1625, opened 2024-01-13
Vectorization Factor patch for computeInfoC2P with Broadcast in mapped IterDomain

### LINKED ISSUE #1567: vectorize_factor is reduced to 1 with allocation domain
This is the repro script:

```
import torch
import nvfuser
import functools
from torch import Tensor, HalfTensor, BoolTensor
from typing import Tuple

DEBUG = True
if DEBUG:
    WARMUP_ITERS = 0
    ITERS = 1
else:
    WARMUP_ITERS = 10
    ITERS = 100

#REPRO = False
REPRO = True

if REPRO:
    FORMAT = (3, 0, 2, 1)
else:
    FORMAT = (3, 2, 1, 0)

def print_tensors(tensors, tensor_names):
    for i, t in enumerate(tensors):
        print(tensor_names[i] + ":", t.shape, t.stride(), t.dtype, t.requires_grad)

class bn_relu_jit(torch.autograd.Function):
    @staticmethod
    @torch.cuda.amp.custom_fwd(cast_inputs=torch.half)
    def forward(ctx, input, scale, bias):
        bn_relu_out, relu_mask = fwd_bn_relu_nvfuser(input, scale, bias)

        # Debugging prints
        if DEBUG:
            print_tensors([input, scale, bias, bn_relu_out, relu_mask],
                          ["fwd_input", "fwd_scale", "fwd_bias", "fwd_bn_relu_out", "fwd_relu_mask"])

        ctx.save_for_backward(scale, relu_mask)
        return bn_relu_out

    @staticmethod
    @torch.cuda.amp.custom_bwd
    def backward(ctx, grad_output):
        raise NotImplemented


### nvFuser Python frontend

def partially_contig_tensor(
    fd: "nvfuser.FusionDefinition",
    x: torch.Tensor,
) -> "nvfuser.Tensor":
    return fd.define_tensor(
        sizes=x.size(),
        strides=x.stride(),
        dtype=torch_dtype_to_nvfuser_dtype(x.dtype),
    )
    #return fd.define_tensor(
    #    shape=[-1] * x.ndim,
    #    contiguity=nvfuser.compute_contiguity(x.size(), x.stride()),
    #    dtype=torch_dtype_to_nvfuser_dtype(x.dtype),
    #)

def fwd_bn_relu_nvfuser(input: HalfTensor, scale: HalfTensor, bias: HalfTensor) -> Tuple[HalfTensor, BoolTensor]:
    tensors = [input, scale, bias]

    with nvfuser.FusionDefinition() as fd:
        x = partially_contig_tensor(fd, tensors[0])
        s = partially_contig_tensor(fd, tensors[1])
        b = partially_contig_tensor(fd, tensors[2])
        z = fd.define_scalar(0)

        T0 = fd.ops.mul(x, s)
        T1 = fd.ops.add(T0, b)
        T2 = fd.ops.relu(T1)
        if REPRO:
            T3 = fd.ops.cast(T2, dtype=nvfuser.DataType.Half)
        else:
            T3 = fd.ops.set(T2)
        T4 = fd.ops.gt(T1, z)

        fd.add_output(T3, FORMAT)
        fd.add_output(T4, FORMAT)

    bn_relu, relu_mask = fd.execute(tensors)

    return bn_relu, relu_mask

### Inputs
inputs = [
  torch.randn(32, 1024, 25, 25, device="cuda").as_strided((32, 1024, 25, 25), (640000, 1, 25600, 1024)).half().requires_grad_(),
  torch.randn(1, 1024, 1, 1, device="cuda").as_strided((1, 1024, 1, 1), (1024, 1, 1024, 1024)).half(),
  torch.randn(1, 1024, 1, 1, device="cuda").as_strided((1, 1024, 1, 1), (1024, 1, 1024, 1024)).half(),
]

### Repro code
model = bn_relu_jit()

for i in range(WARMUP_ITERS):
    o = model.apply(*inputs)

torch.cuda.profiler.start()
for i in range(ITERS):
    o = model.apply(*inputs)
torch.cuda.profiler.stop()
```

with stride_order/alloc_domain properly plumbed through nvfuser codegen IR, we are actually seeing plummeting performance. :cry: 
running with `NVFUSER_DUMP=python_definition,dump_eff_bandwidth,scheduler_params,segmenter_logging,fusion_ir python repro.py`

I'm getting
```
    T0 = fd.define_tensor(shape=[-1, -1, -1, -1], contiguity=[True, True, True, True], dtype=DataType.Half, is_cpu=False, stride_order=[3, 0, 2, 1])
    T1 = fd.define_tensor(shape=[1, -1, 1, 1], contiguity=[None, None, None, True], dtype=DataType.Half, is_cpu=False, stride_order=[3, 0, 2, 1])
    T2 = fd.define_tensor(shape=[1, -1, 1, 1], contiguity=[None, None, None, True], dtype=DataType.Half, is_cpu=False, stride_order=[3, 0, 2, 1])
    S3 = fd.define_scalar(0, dtype=DataType.Int)
    T4 = fd.ops.mul(T0, T1)
    T5 = fd.ops.add(T4, T2)
    T6 = fd.ops.relu(T5)
    T7 = fd.ops.cast(T6, dtype=DataType.Half)
    T8 = fd.ops.gt(T5, S3)
    fd.add_output(T7, stride_order=[3, 0, 2, 1])
    fd.add_output(T8, stride_order=[3, 0, 2, 1])

===== Pointwise Stats ========
num_elems: 20480000
elem_counts: 32 1024 25 25
max_input_dtype_size: 2
vectorize_factor: 1
broadcast_byte_multiples: (0, 9), (5, 9), (9, 5), (9, 5), LHS elems: 32768 RHS elems: 625


===== Pointwise Parameters ========
Tag: Pointwise heuristics Pointwise Characteristics:
 Gridx: 1 BlckY: 1 BlckX: 79
2D Schedule
  Bcast break point: 2
Unroll, Factor: 8
```

```
kernelpointwise_f0_c1_r0_g0 run in 1.13971 ms, achieved: 89.8509 GB/s
```

I'll take a initial look at this to see if I can fix the vectorization analysis first.

jjsjann123:
Fixes #1567

This PR patches vectorization factor in `ContiguousInnerDimensionsMapper::computeInfoC2P`.

Handling of resolved broadcast dimension should be made on mapped consumer tensors' from_ids, instead of the root_domain order. Added a few tests per @zasdfgbnm 's suggestion:

```
Case 0:
T2[1024, 2, 512] = T0[1024, 2, 1] + T1[1024, 2, 512]
allocation = rfactor
--> T0 has no vectorization

Case 1:
T2[1024, 512, 2] = T0[1024, 1, 2] + T1[1024, 512, 2]
allocation = rfactor
--> T0 has vectorization 2

Case 2:
T2[1024, 512, 2] = T0[1024, 1, 2] + T1[1024, 512, 2];
T3[512, 1024, 2] = transpose(T2[1024, 512, 2])
allocation = rfactor
*except T1 has stride_order {1, 2, 0}
--> T0 has vectorization 4

Case 3:
T2[512, 1024, 2] = T0[1, 1024, 2] + T1[512, 1024, 2]
T3[1024, 512, 2] = transpose(T2[512, 1024, 2])
allocation = rfactor
--> T0 has vectorization 2
```


jjsjann123:
!build

jjsjann123:
Note: This is NOT a general fix. We should discuss how we would want to handle vectorization analysis with allocation domain. The logic felt a bit messy, i.e. how `scheduler`s are going to mutate the fusion per allocation domain is going to be dependent on each individual scheduler, does the vectorization analysis now needs to be scheduler dependent?!

This is already the case for `Transpose` and `Pointwise`, does this mean we also need to have future special vectorization analysis for matmul/reduction/normalization?!

jjsjann123:
I'm marking this `ready for review` as a small patch, since I think it's doing the right thing here.
But in general, allocation domain support in vectorization I think needs more work and thought. Even with just a simple case on pointwise scheduler, the bias/scale inputs can be fed in many different ways (expanded tensor / vector tensor with all the variants on stride order). I'll try to put in more concrete test and we can investigate those later.

jjsjann123:
!build

jjsjann123:
!build

jjsjann123:
!build

csrc/scheduler/vectorize_helper.cpp
```diff
@@ -484,6 +484,29 @@ ContiguousInnerDimensionsMapper::computeInfoC2P(
...
+    // inner most and we should still map it for vectorization. see issue:
+    // https://github.com/NVIDIA/Fuser/issues/1567#issuecomment-1894605385
+    bool clear_pos_in_mapped_root = false;
+    for (auto i : c10::irange(from_ids.size())) {
+      if (clear_pos_in_mapped_root) {
+        if (auto iter = consumer_ids_to_clear.find(from_ids[i]);
+            iter != consumer_ids_to_clear.end()) {
+          consumer_ids_to_clear.erase(iter);
+        }
+      } else if (from_ids[i] == from->getRootDomain()[clear_pos]) {
```
naoyam, line 506: Shouldn't we have a guard like `clear_pos >= 0`?

jjsjann123: good catch! I should have kept it inside the true block for `clear_pos >= 0`

I'll try to add a test case for that!

csrc/scheduler/vectorize_helper.cpp
```diff
@@ -484,6 +484,29 @@ ContiguousInnerDimensionsMapper::computeInfoC2P(
           from->getRootDomain().begin(),
           from->getRootDomain().begin() + clear_pos + 1);
     }
+
+    // A special case is that, if the last broadcast dimension resolved in
+    // consumers root domain is mapped for vectorization, then merge order in
+    // the vectorization axes matters. T0[i0, i1] T1[i0, i1, b2] = broadcast(T0)
+    // T2[i0, i1, i3]
+    // T3[i0, i1, i2] = T1 + T2
```
jacobhinkle, line 492:
```suggestion
    // the vectorization axes matters.
    // T0[i0, i1]
    // T1[i0, i1, b2] = broadcast(T0)
    // T2[i0, i1, i3]
    // T3[i0, i1, i2] = T1 + T2
```

csrc/scheduler/vectorize_helper.cpp
jjsjann123: put it under the block. I did verify the invalid access on a simple repro but don't see a good reason to put a test.
The existing code didn't break any test, looks like that `-1` accidentally just give us a false on this `from_ids[i] == from->getRootDomain()[clear_pos]`...

jjsjann123:
!build

csrc/scheduler/vectorize_helper.cpp
```diff
@@ -483,6 +483,33 @@ ContiguousInnerDimensionsMapper::computeInfoC2P(
...
+      // T0[i0, i1]
+      // T1[i0, i1, b2] = broadcast(T0)
+      // T2[i0, i1, i3]
+      // T3[i0, i1, i2] = T1 + T2
+      //
+      // If the mapped ids are {i0, i2, i1}, when propagating from T3 to T1, the
+      // previous "inner most resolved broadcast" would eliminate all `i0, i1,
+      // i2` from vectorization, but the reality is that the merged axes has i1
+      // in the inner most and we should still map it for vectorization. see
+      // issue:
```
naoyam, line 500: I'm still trying to understand the reasoning here. 

Is vectorizable because the domains are ordered as `{i0, i2, i1}` and the broadcast domain of `i2` only interferes with `i0`?

In this case, is the allocation domains of `T3` are also ordered as `{i0, i2, i1}`?

jjsjann123: logging some offline conversation: Naoya pointed out a similar issue with the broadcast propagation, tracking that via #1654

There's also a concern on the side effect from this change, i.e. how safe it is regarding the use of vectorization factor in schedulers. tagging @xwang233 @jacobhinkle for comments.

csrc/scheduler/vectorize_helper.cpp
```diff
@@ -477,17 +483,14 @@ ContiguousInnerDimensionsMapper::computeInfoC2P(
...
-    if (clear_pos >= 0) {
-      consumer_ids_to_clear.insert(
-          from->getRootDomain().begin(),
-          from->getRootDomain().begin() + clear_pos + 1);
-    }
   }
 
   std::vector<IterDomain*> producer_rfactor_ids;
-  for (auto from_id : from_ids) {
+  for (int64_t i : c10::irange(from_ids.size())) {
```
zasdfgbnm, line 489:
```suggestion
  for (int64_t i : c10::irange(clear_pos, from_ids.size())) {
```

csrc/scheduler/vectorize_helper.cpp
```diff
@@ -445,29 +448,32 @@ ContiguousInnerDimensionsMapper::computeInfoC2P(
...
   // concretization.
   std::unordered_set<IterDomain*> consumer_ids_to_clear;
+  int clear_pos = -1;
   if (to->hasBroadcast()) {
-    // Find the last broadcast dimension resolved in consumers root domain
-    int clear_pos = -1;
-    for (auto i : c10::irange(from->getRootDomain().size())) {
-      auto c_id = from->getRootDomain()[i];
+    // Find the last broadcast dimension resolved in consumers through from_ids
+    for (auto i : c10::irange(from_ids.size())) {
```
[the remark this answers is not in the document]
jjsjann123: :face_with_head_bandage: Good catch! I don't know why I was just mindlessly copying the old behavior.

csrc/scheduler/vectorize_helper.cpp
jjsjann123: I guess I really needed some sleep at this point :laughing:

jjsjann123:
!build

jjsjann123:
!build

jjsjann123:
Wondering if I can get a review on this @zasdfgbnm

jjsjann123:
!build

test/test_pointwise.cpp
```diff
@@ -201,4 +201,173 @@ TEST_F(PointwiseTest, VectorizeAllocationDomain) {
...
+
+  // NOTE: force pointwise scheduler here just for testing purpose
+  auto params = getPointwiseHeuristics(fusion, aten_inputs);
+  auto lparams = schedulePointwise(fusion, aten_inputs);
+  FusionExecutor fe;
+  fe.compileFusion(fusion, aten_inputs, lparams);
+  auto cg_outputs = fe.runFusion(aten_inputs, lparams);
+
+  EXPECT_EQ(params->vectorize, true);
+  EXPECT_EQ(params->unroll_factor, 4);
```
zasdfgbnm, line 275: I think we should check that, in the scheduled fusion, tv0's input cache is not vectorized?

test/test_pointwise.cpp
```diff
@@ -201,4 +201,173 @@ TEST_F(PointwiseTest, VectorizeAllocationDomain) {
...
+  fusion->addOutput(tv2);
+
+  FusionExecutorCache fec(std::move(fusion_ptr));
+  fec.profile(true);
+
+  auto options = at::TensorOptions().dtype(at::kFloat).device(at::kCUDA, 0);
+  at::Tensor t0 = at::randn({1024, 1, 2}, options);
+  at::Tensor t1 = at::randn({1024, 512, 2}, options);
+  auto cg_outputs = fec.runFusionWithInputs({t0, t1});
+  EXPECT_EQ(getVecSizeForPointwise(fec), 2);
```
zasdfgbnm, line 306: I think we should also check that the input cache of tv0 is vectorized?

test/test_pointwise.cpp
```diff
@@ -201,4 +201,173 @@ TEST_F(PointwiseTest, VectorizeAllocationDomain) {
...
+  fusion->addOutput(tv3);
+
+  FusionExecutorCache fec(std::move(fusion_ptr));
+  fec.profile(true);
+
+  auto options = at::TensorOptions().dtype(at::kFloat).device(at::kCUDA, 0);
+  at::Tensor t0 = at::randn({1024, 1, 2}, options);
+  at::Tensor t1 = at::empty_strided({1024, 512, 2}, {2, 2048, 1}, options);
+  auto cg_outputs = fec.runFusionWithInputs({t0, t1});
+  EXPECT_EQ(getVecSizeForPointwise(fec), 4);
```
zasdfgbnm, line 338: I think we should also check that the input cache of tv0 is vectorized?

test/test_pointwise.cpp
```diff
@@ -201,4 +201,173 @@ TEST_F(PointwiseTest, VectorizeAllocationDomain) {
...
+  fusion->addOutput(tv3);
+
+  FusionExecutorCache fec(std::move(fusion_ptr));
+  fec.profile(true);
+
+  auto options = at::TensorOptions().dtype(at::kFloat).device(at::kCUDA, 0);
+  at::Tensor t0 = at::randn({1, 1024, 2}, options);
+  at::Tensor t1 = at::randn({512, 1024, 2}, options);
+  auto cg_outputs = fec.runFusionWithInputs({t0, t1});
+  EXPECT_EQ(getVecSizeForPointwise(fec), 2);
```
zasdfgbnm, line 369: I think we should also check that the input cache of tv0 is vectorized?

test/test_pointwise.cpp
```diff
@@ -201,4 +201,173 @@ TEST_F(PointwiseTest, VectorizeAllocationDomain) {
...
+  fusion->addOutput(tv3);
+
+  FusionExecutorCache fec(std::move(fusion_ptr));
+  fec.profile(true);
+
+  auto options = at::TensorOptions().dtype(at::kFloat).device(at::kCUDA, 0);
+  at::Tensor t0 = at::empty_strided({1024, 128, 25}, {128*25, 1, 128}, options);
+  at::Tensor t1 = at::empty_strided({1, 128, 1}, {128, 1, 128}, options);
+  auto cg_outputs = fec.runFusionWithInputs({t0, t1});
+  EXPECT_EQ(getVecSizeForPointwise(fec), 4);
```
zasdfgbnm, line 239: I think we should also check that the input cache of tv0 and tv1 is vectorized? Because it is possible that, if not done correctly, only the output tv is vectorized, and inputs are just unrolled.

zasdfgbnm (approved):
Left some comments on improving tests to be more specific on who is vectorized. But the changes in vectorization helper and the overall idea of these tests make sense to me.

naoyam (approved):
Makes a lot of sense!

jjsjann123:
!build

Issue #1590, opened 2024-01-05
`Allocations must be based on constant integers for local memory` assertion Error.


nikitaved:
The following code
```python                                                                                                                                                                                                                                     
import torch                                                                                                                                                                                                                                  
from nvfuser import FusionDefinition, DataType                                                                                                                                                                                                
                                                                                                                                                                                                                                              
def nvfuser_fusion_id42(fd : FusionDefinition) -> None :                                                                                                                                                                                      
    T0 = fd.define_tensor(shape=[-1, -1, -1, -1, -1], contiguity=[True, True, None, True, None], dtype=DataType.BFloat16, is_cpu=False, stride_order=[4, 1, 2, 3, 0])                                                                         
    T1 = fd.define_tensor(shape=[-1, -1, -1, -1, -1], contiguity=[True, None, True, True, True], dtype=DataType.BFloat16, is_cpu=False, stride_order=[4, 3, 2, 0, 1])                                                                         
    T2 = fd.ops.cast(T0, dtype=DataType.Float)                                                                                                                                                                                                
    S3 = fd.define_scalar(1.00000, dtype=DataType.Double)                                                                                                                                                                                     
    S4 = fd.define_scalar(2, dtype=DataType.Int)                                                                                                                                                                                              
    S5 = fd.define_scalar(8, dtype=DataType.Int)                                                                                                                                                                                              
    S6 = fd.define_scalar(8, dtype=DataType.Int)                                                                                                                                                                                              
    V7 = fd.define_vector([S4, S5, S6], dtype=DataType.Int)                                                                                                                                                                                   
    T8 = fd.ops.full(shape=V7, fill_value=S3, dtype=DataType.BFloat16)                                                                                                                                                                        
    S9 = fd.define_scalar(2, dtype=DataType.Int)                                                                                                                                                                                              
    S10 = fd.define_scalar(8, dtype=DataType.Int)                                                                                                                                                                                             
    S11 = fd.define_scalar(8, dtype=DataType.Int)                                                                                                                                                                                             
    S12 = fd.define_scalar(512, dtype=DataType.Int)                                                                                                                                                                                           
    S13 = fd.define_scalar(512, dtype=DataType.Int)                                                                                                                                                                                           
    V14 = fd.define_vector([S9, S10, S11, S12, S13], dtype=DataType.Int)                                                                                                                                                                      
[...]

PR #1626, opened 2024-01-13
avoid reorder constant root domain

liqiangxl:
Fix https://github.com/NVIDIA/Fuser/issues/1590

### DIFF
--- a/csrc/scheduler/reduction_utils.cpp
+++ b/csrc/scheduler/reduction_utils.cpp
@@ -576,8 +576,11 @@ int idPos(const IterDomain* id) {
   // Reduction, constant and:
   // id must be the factor used in a split, otherwise, the split output may be
   // constant in the reference tensor but dynamic in others. See issue 1590.
-  bool is_split_factor = id->isDefinitionType<Split>() &&
-      (id == dynamic_cast<Split*>(id->definition())->factor());
+  bool is_split_factor = false;
+  if (auto split = dynamic_cast<Split*>(id->definition())) {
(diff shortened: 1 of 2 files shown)


liqiangxl:
!build --diff --diff-bench

liqiangxl:
!build --diff --diff-bench

liqiangxl:
!build --diff --diff-bench

liqiangxl:
!build --diff --diff-bench

liqiangxl:
!build --diff --diff-bench

liqiangxl:
The codediff is because the order of the IDs are changed, e.g. in `NVFuserTest.CombinedReduction_CUDA`, main branch has
`T9_l[ ... iS99{27}, ithreadIdx.y100{( ceilDiv(216, 27) )} ] ca_pos( 5 )`, 
it is changed to 
`T9_l[ ..., ithreadIdx.y100{( ceilDiv(216, 27) )}, iS99{27} ] ca_pos( 5 )`
 becuase `ithreadIdx.y100{( ceilDiv(216, 27) )}` is not a const constructed from split factors. So it is being considered as `i-thread dims` instead of `i-constant sized`, which seems more reasonable since it is parallelized by `threadIdx.y`.

Other changes are due to the same reasons. e.g `NVFuserTest.FusionRepro2241_CUDA`, `FusionReshapeMagicSchedule8`,  `FusionReshapePersistentShmoo`, and `TestNvFuserFrontend::test_nanogpt_mha_dpa`. These are cases with constant root domains.

liqiangxl:
Fixed by https://github.com/NVIDIA/Fuser/pull/1642

PR #1627, opened 2024-01-14
IdModel: Fix inconsistent graph issue 

### LINKED ISSUE #1636: IdModel: Inconsistent graph
Just realized there's a test failure with `test_instance_norm[float32-True-True-False-channels_last-1-7-3-True]`

```
RuntimeError: entry_it != disjointSetMap().end() INTERNAL ASSERT FAILED at "/home/nmaruyama/nvfuser/debug1/csrc/disjoint_set.h":412, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. Strict mapp
ing failed on element: Merge: bS61{1} and iS230{( (( (( getMetaData(T0) )).logical_size ))[4] )} -> iS135{( 1 * ( (( (( getMetaData(T0) )).logical_size ))[4] ) )}
 either an error occurred, or non strict mapping should have been used.
Exception raised from strictAreMapped at /home/nmaruyama/nvfuser/debug1/csrc/disjoint_set.h:412 (most recent call first):
frame #0: nvfuser::nvfCheckFail(char const*, char const*, unsigned int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) + 0x7a (0x7f2a8009c0fa in /opt/conda/pytorch/lib/python3.11/site-packages/torch/lib/libnvfuser_codegen.so)
frame #1: nvfuser::nvfErrorFail(char const*, char const*, unsigned int, char const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) + 0x5d (0x7f2a8009c36d in /opt/conda/pytorch/lib/python3.11/site-packages/torch/lib/libnvfuser_codegen.so)
frame #2: nvfuser::DisjointSets<nvfuser::Expr*, std::hash<nvfuser::Expr*> >::strictAreMapped(nvfuser::Expr*, nvfuser::Expr*) const + 0xcd (0x7f2a8074d06d in /opt/conda/pytorch/lib/python3.11/site-packages/torch/lib/libnvfuser_codegen.so)
frame #3: nvfuser::ValGraph::mapExprs(nvfuser::Expr*, nvfuser::Expr*) + 0x51 (0x7f2a807489a1 in /opt/conda/pytorch/lib/python3.11/site-packages/torch/lib/libnvfuser_codegen.so)
frame #4: nvfuser::ValGraph::maybeMapThroughExprs(nvfuser::Expr*, nvfuser::Expr*, bool) + 0x6f (0x7f2a8074862f in /opt/conda/pytorch/lib/python3.11/site-packages/torch/lib/libnvfuser_codegen.so)
frame #5: nvfuser::IdModel::addReplayAs(std::vector<nvfuser::IterDomain*, std::allocator<nvfuser::IterDomain*> >, nvfuser::Expr*) + 0x15f0 (0x7f2a80183230 in /opt/conda/pytorch/lib/python3.11/site-packages/torch/lib/libnvfuser_codegen.so)
```

The error output is cryptic. The error is reported at the point where we replay some domain for loop promotion, but the real root cause seems to start showing up even when the almost exact graph is created. 

What happens seems to be that the def and use maps becomes inconsistent when merging expr groups. All val groups should be updated to point to the new merged expr group, but that doesn't seem to be the case in some cases. The above test actually exposed the inconsistency but there may be other tests that also result in similar inconsistent graphs with no visible side effect.

Here's an example scenario where this error happens. 

Inconsistent Graph-2

It has two input tensors, which are outer-split by a factor of 1 as illustrated. Then, suppose IDs 4 and 7 and 3 and 8 are mapped, respectively (Step 2). At this point, only no other IDs are mapped and each expr should be an expr group by its own. Then, suppose IDs 1 and 2 are mapped, perhaps because of the almost exact map rule, which should then map the first and second splits of the left tensor (i.e., e1 and e2) since the inputs are mapped and the splits have the same attributes. At this point the definition map should look like as shown in the image. At this point, notice that the val groups of `{5}` and `{3, 8}` are not mapped yet but they share a defining expr group, `{e1, e2}`. The next thing to do is to propagate mappings to their output val groups, and suppose IDs 2 and 4 are mapped, which will create a val group of `{1, 2, 4, 7}`. Now that IDs 2 and 4 are mapped, suppose their definition exprs are mapped next, that is the first split of the left tensor and the only split of the right tensor (i.e., e1 and e3). Note that this is a back propagation, and since the split outputs are already mapped the expr groups are merged into `{e1, e2, e3}`. Since e1 and e3 are merged, their output val groups will be updated to point to the new merged expr group. The output groups are: `{1, 2, 4, 7}, {3, 8}`. Notice that the group of ID 5 is not included since since it's not an output of e1 nor e3. Because of this, the definition map will still have the mapping from `{5}` to `{e1, e2}`, however at this point there's no val group of just e1 and e2 as they are merged with e3, so the definition expr group of `{5}` is a dangling expr group, which will be kept as is even after ID 5 is merged with the other val groups, making the expr groups no longer disjoint.

I tried to reproduce the error with a simple fusion but gave up as it depends on the ordering of vals and exprs, some of which are difficult to control with the current infrastructure. 

To fix this issue, we should change this part:

https://github.com/NVIDIA/Fuser/blob/main/csrc/val_graph.cpp#L446-L471

A WIP PR is here: #1627

naoyam:
Fixes #1636 

Previously when exprs are merged, their input and output groups are updated to point to the new merged group. However, this is not enough as there may be val groups that are used or defined by the pre-merged expr groups but they may not be inputs or outputs of the exprs. See #1636 for a concrete example.

Here, we check all mapping entries of `unique_definitions_` and `unique_uses_` so all uses and definitions are correctly updated.

### DIFF
--- a/csrc/disjoint_set.h
+++ b/csrc/disjoint_set.h
@@ -410,7 +410,8 @@ class DisjointSets {
         "Strict mapping failed on element: ",
         abstractToString(entry0),
         " either an error occurred, or non strict mapping should have been used.",
-        " ", entry0->name());
+        " ",
+        entry0->name());
     return entry_it->second->has(entry1);
   }
 
--- a/csrc/disjoint_set.h
+++ b/csrc/disjoint_set.h
@@ -409,7 +409,8 @@ class DisjointSets {
         entry_it != disjointSetMap().end(),
         "Strict mapping failed on element: ",
         abstractToString(entry0),
-        " either an error occurred, or non strict mapping should have been used.");
+        " either an error occurred, or non strict mapping should have been used.",
+        " ", entry0->name());
     return entry_it->second->has(entry1);
   }
 
--- a/csrc/id_model/id_model.cpp
+++ b/csrc/id_model/id_model.cpp
@@ -1256,7 +1256,7 @@ void IdModel::build(
   // debug_print(IdMappingMode::EXACT);
 
   buildAlmostExactMap();
-  if (false && validate) {
+  if (validate) {
     validator->checkAlmostExactGraphEquivalence(
         idGraph(IdMappingMode::ALMOSTEXACT));
   }
--- a/csrc/id_model/id_model.cpp
+++ b/csrc/id_model/id_model.cpp
@@ -410,6 +410,7 @@ Expr* IdModel::addReplayAs(std::vector<IterDomain*> new_inputs, Expr* expr) {
     for (IterDomain* inp : new_inputs) {
       if (const ExprGroups* uses = graph.getUses(graph.toGroup(inp)); uses) {
         for (const ExprGroup& use_group : *uses) {
+          NVF_ERROR(!use_group->empty());
           representative_uses.pushBack(use_group->front());
         }
       }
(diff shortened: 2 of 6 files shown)


naoyam:
!build

naoyam:
!build

naoyam:
!build

naoyam:
!build

naoyam:
!build

PR #1632, opened 2024-01-16
Automatic Resharding Set insertion 

samnordmann:
Add the multidevice utility function `void insertReshardings(Fusion* fusion)` that runs through the fusion and inserts a resharding Set Op before any resharding Expr that is not directly lowerable to a series of communications

### DIFF
--- a/csrc/multidevice/utils.cpp
+++ b/csrc/multidevice/utils.cpp
@@ -128,6 +128,7 @@ void reshardBefore(Expr* expr, Fusion* fusion) {
         return val->as<TensorView>();
       });
   std::vector<TensorView*> new_inputs;
+  // if the expr is not resharding, the following for loop is empty
   for (auto input : haveDifferentSharding(output, inputs)) {
     // TODO: reuse cacheAfter?
     // TODO: here we should add a mechanism to potentially reuse the inserted
@@ -148,6 +149,7 @@ void insertReshardings(Fusion* fusion) {
   auto exprs = fusion->exprs();
   for (auto expr : exprs) {
     if (!isLowerableToCommunication(expr)) {
+      // if the expr is not resharding, reshardBefore will not modify it
       reshardBefore(expr, fusion);
     }
   }
--- a/csrc/multidevice/utils.cpp
+++ b/csrc/multidevice/utils.cpp
@@ -49,7 +49,7 @@ std::vector<IterDomain*> getShardedIterDomains(TensorView* tv) {
 
 } // namespace
 
-template<typename TvIterator>
+template <typename TvIterator>
 std::unordered_set<TensorView*> getTvsWithDifferentSharding(
     TensorView* ref,
     TvIterator tvs) {
@@ -114,35 +114,34 @@ void shardAllLike(TensorView* ref, std::vector<TensorView*> tvs) {
   scheduler_utils::parallelizeAllLike(ref, tvs, {ParallelType::DIDx});
 }
 
-void maybeReshardInputs(Expr* expr, Fusion* fusion) {
-  NVF_ERROR(
-      expr->outputs().size() == 1,
-      "multi-output expressions are not supported");
-  auto output = expr->outputs().at(0)->as<TensorView>();
-  std::vector<TensorView*> new_inputs;
-  for (auto input : getTvsWithDifferentSharding(output, ir_utils::filterByType<TensorView>(expr->inputs()))) {
-    // TODO: reuse cacheAfter?
-    // TODO: here we should add a mechanism to potentially reuse the inserted
-    // resharding accross all the consumer of the resharded tensor. This way we
-    // could avoid wasteful resharding set insertion.
-    TensorView* new_input = set(input);
-    expr = ir_utils::replaceValInExprInputs(expr, input, new_input);
-    new_inputs.push_back(new_input);
-  }
-  if (!new_inputs.empty()) {
-    shardAllLike(output, new_inputs);
-  }
-}
-
 } // namespace
 
 void insertReshardings(Fusion* fusion) {
   auto exprs = fusion->exprs();
   for (auto expr : exprs) {
-    if (!isLowerableToCommunication(expr)) {
-      NVF_ERROR(ir_utils::isTvOp(expr), "Non-tv op is not supported yet: ", expr->toString());
-      maybeReshardInputs(expr, fusion);
+    if (isLowerableToCommunication(expr)) {
+      continue;
     }
+    NVF_ERROR(
+        ir_utils::isTvOp(expr),
+        "Non-tv op is not supported yet: ",
+        expr->toString());
+    NVF_ERROR(
+        expr->outputs().size() == 1,
+        "multi-output expressions are not supported");
+    auto output = expr->outputs().at(0)->as<TensorView>();
+    std::vector<TensorView*> new_inputs;
+    for (auto input : getTvsWithDifferentSharding(
+            output, ir_utils::filterByType<TensorView>(expr->inputs()))) {
+      // TODO: reuse cacheAfter?
+      // TODO: here we should add a mechanism to potentially reuse the inserted
+      // resharding accross all the consumer of the resharded tensor. This way we
+      // could avoid wasteful resharding set insertion.
+      TensorView* new_input = set(input);
+      new_inputs.push_back(new_input);
+      expr = ir_utils::replaceValInExprInputs(expr, input, new_input);
+    }
+    shardAllLike(output, new_inputs);
   }
 }
 
--- a/csrc/multidevice/utils.cpp
+++ b/csrc/multidevice/utils.cpp
@@ -47,7 +47,7 @@ std::vector<IterDomain*> getShardedIterDomains(TensorView* tv) {
 
 } // namespace
 
-std::unordered_set<TensorView*> haveDifferentSharding(
+std::unordered_set<TensorView*> getTvsWithDifferentSharding(
     TensorView* ref,
     std::unordered_set<TensorView*> tvs) {
   std::unordered_set<TensorView*> ret;
@@ -96,7 +96,7 @@ bool isResharding(Expr* expr) {
   }
   auto tv_ref = *tvs.begin();
   tvs.erase(tv_ref);
-  return !haveDifferentSharding(tv_ref, tvs).empty();
+  return !getTvsWithDifferentSharding(tv_ref, tvs).empty();
 }
 
 namespace {
@@ -129,7 +129,7 @@ void reshardBefore(Expr* expr, Fusion* fusion) {
       });
   std::vector<TensorView*> new_inputs;
   // if the expr is not resharding, the following for loop is empty
-  for (auto input : haveDifferentSharding(output, inputs)) {
+  for (auto input : getTvsWithDifferentSharding(output, inputs)) {
     // TODO: reuse cacheAfter?
     // TODO: here we should add a mechanism to potentially reuse the inserted
     // resharding accross all the consumer of the resharded tensor. This way we
--- a/csrc/multidevice/utils.cpp
+++ b/csrc/multidevice/utils.cpp
@@ -132,11 +132,11 @@ void insertReshardings(Fusion* fusion) {
     auto output = expr->outputs().at(0)->as<TensorView>();
     std::vector<TensorView*> new_inputs;
     for (auto input : getTvsWithDifferentSharding(
-            output, ir_utils::filterByType<TensorView>(expr->inputs()))) {
+             output, ir_utils::filterByType<TensorView>(expr->inputs()))) {
       // TODO: reuse cacheAfter?
       // TODO: here we should add a mechanism to potentially reuse the inserted
-      // resharding accross all the consumer of the resharded tensor. This way we
-      // could avoid wasteful resharding set insertion.
+      // resharding accross all the consumer of the resharded tensor. This way
+      // we could avoid wasteful resharding set insertion.
       TensorView* new_input = set(input);
       new_inputs.push_back(new_input);
       expr = ir_utils::replaceValInExprInputs(expr, input, new_input);
--- a/csrc/multidevice/utils.cpp
+++ b/csrc/multidevice/utils.cpp
@@ -9,6 +9,9 @@
 #include <compute_at_map.h>
 #include <ir/internal_base_nodes.h>
 #include <ir/utils.h>
+#include <scheduler/utils.h>
+#include <ops/all_ops.h>
+#include <multidevice/lower_communication.h>
 #include <multidevice/utils.h>
 
 #include <c10/util/irange.h>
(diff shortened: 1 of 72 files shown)


samnordmann:
!build

csrc/multidevice/utils.cpp
```diff
@@ -96,4 +99,58 @@ bool isResharding(Expr* expr) {
...
+            "the expression's input is not a TensorView");
+        return val->as<TensorView>();
+      });
+  std::vector<TensorView*> new_inputs;
+  for (auto input : haveDifferentSharding(output, inputs)) {
+    // TODO: reuse cacheAfter?
+    // TODO: here we should add a mechanism to potentially reuse the inserted
+    // resharding accross all the consumer of the resharded tensor. This way we
+    // could avoid wasteful resharding set insertion.
+    TensorView* new_input = set(input);
```
cowanmeg, line 136: I think this will break for something like add(tv0, tv0) since it will try to reshard tv0 twice?

csrc/multidevice/utils.cpp
```diff
@@ -96,4 +99,58 @@ bool isResharding(Expr* expr) {
...
+  }
+}
+
+} // namespace
+
+void insertReshardings(Fusion* fusion) {
+  auto exprs = fusion->exprs();
+  for (auto expr : exprs) {
+    if (!isLowerableToCommunication(expr)) {
+      reshardBefore(expr, fusion);
```
cowanmeg, line 151: There is a small bug here - isLowerableToCommunication marks all reductions as lowerable, but that is only true if the reduction is over a device axis otherwise it needs a reshard expression.

csrc/multidevice/utils.cpp
samnordmann: Actually the scenario you mention is supported in the current version. I added this scenario in the test to support my claim.

Let me try to explain: in the situation you describe, where an Expr has twice the same input `tv0`, then:
1. in the surrounding for loop `auto input : haveDifferentSharding(output, inputs)`, `tv0` will be iterated on only once since `haveDifferentSharding` returns a set (without duplicates)
2. Then, `replaceValInExprInputs` will indeed substitute both "occurences" of `tv0` with the new resharded tv.

csrc/multidevice/utils.cpp
samnordmann: fixed

samnordmann:
!build

samnordmann:
I noticed that `parallelizeAllLike` silently discards the input tvs which are global input of the Fusion. On the other hand, some scheduler like Transpose do indeed call parallelizeAllLike with input tensors. I just wanted to point this out here. Lmk if I should open an issue, write a warning in the comment, or just leave it like that

samnordmann:
the failure in CI seems unrelated

cowanmeg (approved):

csrc/multidevice/utils.cpp
naoyam, line 151: Just want to clarify a couple of things here. It seems a bit strange to do resharding before any expression that doesn't directly correspond to communication ops. For example, certainly scalar, non-tensor ops should not need resharding, and it seems it's implicitly assumed as there's assertions in `resharedBefore`. 

It seems that `insertReshardings` would fail if such ops exist in a given fusion. Shouldn't we make it work anyway? Likely just ignoring non-relevant ops.

csrc/multidevice/utils.cpp
```diff
@@ -96,4 +99,58 @@ bool isResharding(Expr* expr) {
...
+      expr->inputs().end(),
+      std::inserter(inputs, inputs.end()),
+      [](Val* val) {
+        NVF_ERROR(
+            val->isA<TensorView>(),
+            "the expression's input is not a TensorView");
+        return val->as<TensorView>();
+      });
+  std::vector<TensorView*> new_inputs;
+  for (auto input : haveDifferentSharding(output, inputs)) {
```
naoyam, line 131:
```suggestion
  TensorView* output = expr->outputs().at(0)->as<TensorView>();
  std::vector<TensorView*> new_inputs;
  for (auto input_tv: ir_utils::filterByType<TensorView>(expr->inputs())) {
  // Why haveDifferentSharding needs to have all tensor inputs? Can it be just the output and one input tensor like this?
  if (haveDifferentSharding(output, input_tv)) {
```

samnordmann: Ok for filterByType, I actually used this one before, but wanted to assert that all the inputs are TensorViews (mainly for simplicity, I'm always assuming I am in this case because we have not encountered a situation where it is not the case so far).
But alright I replace as you suggest

> // Why haveDifferentSharding needs to have all tensor inputs? Can it be just the output and one input tensor like this?

Because we need run through all the inputs. Note that `haveDifferentSharding` will return the set of all the inputs which sharding is different than the output

csrc/multidevice/utils.cpp
samnordmann: > It seems a bit strange to do resharding before any expression that doesn't directly correspond to communication ops. 

Actually that is quite the opposite: the point of this function is precisely to  insert a set before an expr that doesn't directly correspond to a communication op.

Consider the example:
```
tv0 = makeContigTensor(1);
tv0->setDeviceMesh({0});
tv1 = abs(tv0)
tv0->setDeviceMesh({1});
fusion.addInput(tv0);
fusion.addOutput(tv1);
```

Here, the unary op `abs` doesn't correspond directly to a communication op, because it embedds both a compute (taking the absolute value) AND a communication (send/recv from device 0 to device 1)

In this case, `insertReshardings` will add a intermediate tensor between `tv0` and `tv1`, i.e., it will transform the Fusion to be equivalent to:
```
tv0 = makeContigTensor(1);
tv0->setDeviceMesh({0});
tv0_ = set(tv0);
tv0_->setDeviceMesh({1});
tv1 = abs(tv0_)
tv0->setDeviceMesh({1});
fusion.addInput(tv0);
fusion.addOutput(tv1);
```

This way, we have "separated" the communication and the compute:
- `tv0_ = set(tv0)` corresponds directly to a communication between device 0 and 1
- `tv1 = abs(tv0_)` corresponds to a compute that is local to device 1, not involving any comm


> For example, certainly scalar, non-tensor ops should not need resharding, and it seems it's implicitly assumed as there's assertions in `resharedBefore`

Probably that some cases will need a special treatment, but we have not encoutered them so far. The case you are mentionning, that is "scalar, non-tensor ops", is a case we are not considering because we assert that all Vals are tensors. I think we should not bother with those cases until a concrete example comes to us.
Anyway, if we want (in this PR or later), to exclude some cases, we just need to add conditions on the line you mention. Let me know what condition(s) you would like me to add there.

naoyam: My comment is mostly about clarification rather than technical concerns.

From the function name, it seems to me that `isLowerableToCommunication` implies the given expr can be converted to communication ops, so it makes sense to say that if it returns `true`, it shouldn't need to reshard inputs. However, the fact that it returns `false` doesn't seem to me it requires sharding. A simple example is scalar ops, as I mentioned. 

So, to me more explicit, I think it would be more clear if the code would look like:

```
if (requireSharding(expr) && !isLowerableToCommunication(expr)) {
  reshardBefore(expr);
}
```

Does it make sense?

> Probably that some cases will need a special treatment, but we have not encoutered them so far.

Aren't we going to apply this routine to any given fusions in the future? It makes sense to have some assumptions for short term progress, but it'd be helpful to make it clear what such short-term assumptions are and make them clearly separated from more fundamental limitations.

csrc/multidevice/utils.cpp
naoyam: > Don't you think it is worth adding those asserts ?

I don't think we have any non-tensor input for tensor ops ATM, but even there's one, we could just ignore non-tensor input as it shouldn't require any communication (or, in other words, we don't support distributing non-tensor values).

> Note that haveDifferentSharding will return the set of all the inputs which sharding is different than the output

I looked at the function, and I'm sorry my previous review was sloppy. I have some concerns, but let's leave it as is for now. I'll open a separate issue.

naoyam: #1656

csrc/multidevice/utils.cpp
samnordmann: > My comment is mostly about clarification rather than technical concerns.

alright no problem

> From the function name, it seems to me that `isLowerableToCommunication` implies the given expr can be converted to communication ops, so it makes sense to say that if it returns `true`, it shouldn't need to have a separate reshard op for the inputs.

I agree

> However, the fact that it returns `false` doesn't seem to me it requires sharding. A simple example is scalar ops, as I mentioned.
> 
> So, to me more explicit, I think it would be more clear if the code would look like:
> 
> ```
> if (requireSharding(expr) && !isLowerableToCommunication(expr)) {
>   reshardBefore(expr);
> }
> ```
> 
> Does it make sense?

Ok I understand. Actually, there is the util function `isResharding` that can exactly fulfill this. The only reason I removed it is because, internally, `isResharding` calls `haveSameSharding` and checks if the returned set is empty. Since `haveSameSharding` is called anyway inside `reshardBefore`, this is wasteful.

So I did that for performance, over clarity. If you think we'd rather add `isResharding` for clarity let me know. Alternatively, I can also add a comment.

> > Probably that some cases will need a special treatment, but we have not encoutered them so far.
> 
> Aren't we going to apply this routine to any given fusions in the future? It makes sense to have some assumptions for short term progress, but it'd be helpful to make it clear what such short-term assumptions are and make them clearly separated from more fundamental limitations.

The routine proposed here is as general as I can imagine for now. The only assumption, made for simplicity, is that the Vals are TensorViews, and I think it is a reasonable assumption.

Let me know what do to with that

csrc/multidevice/utils.cpp
samnordmann: > > Don't you think it is worth adding those asserts ?
> 
> I don't think we have any non-tensor input for tensor ops ATM, but even there's one, we could just ignore non-tensor input as it shouldn't require any communication (or, in other words, we don't support distributing non-tensor values).

ok so I will remove the assert and use filterByType

> > Note that haveDifferentSharding will return the set of all the inputs which sharding is different than the output
> 
> I looked at the function, and I'm sorry my previous review was sloppy. I have some concerns, but let's leave it as is for now. I'll open a separate issue.

Alright I will take a look at the issue

samnordmann: Actually I don't think filterByType is suitable here because I want to construct a std::set<TensorView*> to feed the function `haveDifferentSharding`. 

I could, as you suggest, remove the assertion, but I also would rather be extra careful and get an error the day we encounter a case we didn't think about -- in this case if the outputs are not tensors.

So, personally, I wouldn't change anything here. Let me know what you suggest

csrc/multidevice/utils.cpp
samnordmann: I added a comment for clarity in the last commit. I resolve this conversation but feel free to open it again if needed

csrc/multidevice/utils.cpp
samnordmann: The issue https://github.com/NVIDIA/Fuser/issues/1656 you opened is addressed in a separate PR https://github.com/NVIDIA/Fuser/pull/1674

csrc/multidevice/utils.cpp
```diff
@@ -96,4 +99,60 @@ bool isResharding(Expr* expr) {
...
+    shardAllLike(output, new_inputs);
+  }
+}
+
+} // namespace
+
+void insertReshardings(Fusion* fusion) {
+  auto exprs = fusion->exprs();
+  for (auto expr : exprs) {
+    if (!isLowerableToCommunication(expr)) {
```
naoyam, line 151: Insert this:
```
NVF_ERROR(ir_utils::isTvOp(expr), "Non-tv op is not supported yet: ", expr->toString());
```

csrc/multidevice/utils.cpp
naoyam: > The routine proposed here is as general as I can imagine for now. The only assumption, made for simplicity, is that the Vals are TensorViews, and I think it is a reasonable assumption.

The assumption is fine, though I'm not sure why. The problem I had when reviewing the code, and still not clear, is that I didn't know there's this assumption. I added a comment to improve the clarity.

naoyam: > Ok I understand. Actually, there is the util function isResharding that can exactly fulfill this. The only reason I removed it is because, internally, isResharding calls haveSameSharding and checks if the returned set is empty. Since haveSameSharding is called anyway inside reshardBefore, this is wasteful.
> 
> So I did that for performance, over clarity. If you think we'd rather add isResharding for clarity let me know. Alternatively, I can also add a comment.

Again, my comment is for improving the code clarity by separating semantic concerns more clearly. What I'd expect the code should look like is:

```
for each expr in fusion:
  if needResharding(expr):
     insertReshardingBefore(expr)
```

`insertReshardings` exactly looks like this, but what's confusing is `reshardBefore` actually does additional conditional check and may not actually does resharding, and I don't understand why it needs to be written in that way. It's just clearing separating semantic concerns. The first thing is to check the condition, and the next thing is to do what needs to be done if the condition is met. 

I don't understand your performance concern. Of course, we don't want to call `haveSameSharding` twice. We just need to restructure the code, not adding additional things to do.

samnordmann: > but what's confusing is reshardBefore actually does additional conditional check and may not actually does resharding

I renamed the function `reshardBefore` to `maybeReshardInputs`, perhaps this will make the code more readable.


> The first thing is to check the condition, and the next thing is to do what needs to be done if the condition is met.
I don't understand your performance concern. Of course, we don't want to call haveSameSharding twice.

On the one hand, the "condition" whether a "set insertion" is needed must be checked separately at each input -- on the other hand, `parallelizeAllLike` is more efficient if called on all the inputs at once.

That is why it is not so natural to divide it into two functions. Please look at the last version and let me know if it is better now.

samnordmann:
!build

csrc/multidevice/utils.cpp
```diff
@@ -97,7 +102,47 @@ bool isResharding(Expr* expr) {
...
+        ir_utils::isTvOp(expr),
+        "Non-tv op is not supported yet: ",
+        expr->toString());
+    NVF_ERROR(
+        expr->outputs().size() == 1,
+        "multi-output expressions are not supported");
+    auto output = expr->outputs().at(0)->as<TensorView>();
+    std::vector<TensorView*> new_inputs;
+    for (auto input : getTvsWithDifferentSharding(
+             output, ir_utils::filterByType<TensorView>(expr->inputs()))) {
```
naoyam, line 135: I'm feeling a strong uncertainty with this loop. In the loop body, `expr` is being replaced with a new expr that has a new input. Replacing here means the old expr is removed, IIRC, so it seems `expr` in this line becomes a dangling pointer, which should also mean the input vector of the expr becomes invalid.

csrc/multidevice/utils.cpp
naoyam: Thanks. Looks good now.

csrc/multidevice/utils.cpp
samnordmann: I see. So should I revert back to just producing the container of the TensorView inputs that we iterate over ?

naoyam: Sorry, actually not a concern. `getTvsWithDifferentSharding` is called here, so the iteration should be over the returned set, which shouldn't be affected by the replacement inside the loop body. Sorry for the noise.

naoyam (approved):
LGTM

samnordmann:
!build

PR #1637, opened 2024-01-17
[IdModle] Val graph cleanup

naoyam:
Simplify `getDefinitions()` and `getUses()` of `ValGraph` as they always have mappings for any valid ValGroups.


naoyam:
!build

PR #1638, opened 2024-01-18
[IdModel] valgraph cleanup

naoyam:
This is just code simplification as `getDefinitions` and `getUses` don't need to return pointers as there should always be mapped `ExprGroup`.

PR #1639, opened 2024-01-18
[IdModel] mapvals fix

naoyam:
Stacked on top of #1638 

Fixes a memory usage error of using stale references.

Also added sanity checks just in case

### DIFF
--- a/csrc/val_graph.cpp
+++ b/csrc/val_graph.cpp
@@ -335,10 +335,13 @@ void ValGraph::mapVals(Val* val0, Val* val1) {
   const ValGroup orig_val_group0 = toGroup(val0);
   const ValGroup orig_val_group1 = toGroup(val1);
 
-  const ExprGroups& orig_defs0 = getDefinitions(orig_val_group0);
-  const ExprGroups& orig_defs1 = getDefinitions(orig_val_group1);
-  const ExprGroups& orig_uses0 = getUses(orig_val_group0);
-  const ExprGroups& orig_uses1 = getUses(orig_val_group1);
+  // Note that getDefinitions and getUses return references, which
+  // will be invalidated once unique_definitions_ and unique_uses_ are
+  // updated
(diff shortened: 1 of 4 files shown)


naoyam:
!build

naoyam:
!build

naoyam:
!build

naoyam:
!build

csrc/val_graph.cpp
```diff
@@ -335,10 +335,13 @@ void ValGraph::mapVals(Val* val0, Val* val1) {
...
 
-  const ExprGroups& orig_defs0 = getDefinitions(orig_val_group0);
-  const ExprGroups& orig_defs1 = getDefinitions(orig_val_group1);
-  const ExprGroups& orig_uses0 = getUses(orig_val_group0);
-  const ExprGroups& orig_uses1 = getUses(orig_val_group1);
+  // Note that getDefinitions and getUses return references, which
+  // will be invalidated once unique_definitions_ and unique_uses_ are
+  // updated
+  const ExprGroups orig_defs0 = getDefinitions(orig_val_group0);
+  const ExprGroups orig_defs1 = getDefinitions(orig_val_group1);
```
jacobhinkle, line 342: I believe you put this here for a reason, but I wonder what the effect would be if we put these lines down below after the "uses" loop, so that the "defs" loop would use updated definitions instead of the outdated ones.

jacobhinkle (approved):
LGTM

csrc/val_graph.cpp
naoyam: I think we could reorder them as long as each refers to the correct mapping. The interface is not the most robust design as there are a couple of things that can be easily missed. For example, `disjoint_vals_.mapEntries(val0, val1)` invalidates the references returned by `toGroup` above (that's why they need to copy the returned references). Also, the update of `unique_definitions_` and `unique_uses_` invalidates the references returned by `getDefinitions` and `getUses`. I already had nasty bugs a couple of times 😃

PR #1641, opened 2024-01-18
Block API use on USE_DISTRIBUTED.

tfogal:


tfogal:
!build

tfogal:
@samnordmann, `isLowerableToCommunication` is only available when USE_DISTRIBUTED is set.

Alternative thoughts might be to drop multidevice/pipeline.cpp from the build entirely when the define is off, or to make the function available always but have it return some 'default' value when distributed is off. Or maybe something else entirely? Thoughts?

What I mean is: consider this more an "RFC" than a "please rubber stamp this"!

samnordmann (approved):

PR #1643, opened 2024-01-18
More bandaids for #1126. 

### LINKED ISSUE #1126: Split/merged expanded broadcast dimensions can remain broadcast.

wujingyue:
Unfortunately, but I didn't manage to find a better way to unblock myself. This makes alias analysis less effective but is at least safe for expanded broadcasts.

### DIFF
--- a/test/test_alias.cpp
+++ b/test/test_alias.cpp
@@ -928,7 +928,8 @@ TEST_F(AliasTest, MergeTwoExpandedBroadcasts) {
   at::Tensor out_tensor = fec.runFusionWithInputs({in_tensor})[0];
   testValidate(fec.fusion(), {out_tensor}, {in_tensor}, __LINE__, __FILE__);
 
-  EXPECT_TRUE(out_tensor.is_alias_of(in_tensor));
+  // TODO(#1126): This should become an alias when #1126 is fixed.
+  // EXPECT_TRUE(out_tensor.is_alias_of(in_tensor));
 }
 
 TEST_F(AliasTest, MergeBroadcastsBetweenConcretes) {
--- a/test/test_alias.cpp
+++ b/test/test_alias.cpp
@@ -944,6 +944,7 @@ TEST_F(AliasTest, MergeBroadcastsBetweenConcretes) {
                        .build();
   fusion->addInput(in);
   TensorView* out = reshape(in, {2, 3, 5, 7}, {2, -1, 7});
+  out = reshape(out, {2, 15, 7}, {30, 7});
   fusion->addOutput(out);
 
   FusionExecutorCache fec(std::move(fusion));
@@ -951,8 +952,6 @@ TEST_F(AliasTest, MergeBroadcastsBetweenConcretes) {
       at::randn({2 * 7}).cuda().as_strided({2, 3, 5, 7}, {7, 0, 0, 1});
   at::Tensor out_tensor = fec.runFusionWithInputs({in_tensor})[0];
   testValidate(fec.fusion(), {out_tensor}, {in_tensor}, __LINE__, __FILE__);
-
-  EXPECT_TRUE(out_tensor.is_alias_of(in_tensor));
 }
 
 TEST_F(AliasTest, Squeeze) {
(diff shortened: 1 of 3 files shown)


test/test_alias.cpp
```diff
@@ -908,6 +908,27 @@ TEST_F(AliasTest, Broadcast) {
   EXPECT_EQ(out_tensor.data_ptr(), in_tensor.data_ptr());
 }
 
+TEST_F(AliasTest, MergeTwoExpandedBroadcasts) {
```
zasdfgbnm, line 911: Could you add a test that has both concrete and broadcasting dims? For example, contiguity
```
t n n t n n t
```
Because I am concerned about whether in the above code you can just `contiguity = true;`. 

For example, if a tensor has shape (4, 5, 6) and stride (6, 0, 1). Ideally, we should mark dim 1 as broadcasting, and have contiguity `(t, n, t)`. But if instead we are unable to mark dim 1 as broadcasting, we can not just replace the `n` as `t` to get `(t, t, t)`, instead, we should have `(f, f, t)`. Can this case be correctly handled?

wujingyue: > Could you add a test that has both concrete and broadcasting dims?

Done and the test seems to work fine. 

> we can not just replace the `n` as `t` to get `(t, t, t)`

Can you clarify what would break? I'm also not quite confident about this change and feel there can be a problem.

wujingyue:
!build

test/test_alias.cpp
zasdfgbnm: > Can you clarify what would break? I'm also not quite confident about this change and feel there can be a problem.

For example, if the tensor has shape (4, 5, 6) and stride (6, 0, 1), and the contiguity is `(t, t, t)`, then during lower, nvFuser will likely index the tensor like:
```python
for  i in range(4*5*6):
  tensor[0, 0, i]
```
which is wrong. The above indexing is correct only when the stride is `(30, 6, 1)`.

zasdfgbnm (approved):
Not 100% confident if this is safe, but if the added test is green, I am happy with taking this PR to unblock.

test/test_alias.cpp
wujingyue: Alas, I made a slightly more complicated test that breaks this... 

The root problem is that merging `e` is more restrictive than `t` **and even** `f`. For example, according to [this table](https://github.com/NVIDIA/Fuser/blob/0107a7ad24959b575662e94d699ebbc729a9e209/csrc/alias_analysis.cpp#L93-L98), `t,e=>C` but `t,f=>f` and `t,t`=>t. 

I'll think about this more. Chances are we'll have to properly fix #1126, @jacobhinkle

wujingyue:
!build

wujingyue:
!build

PR #1648, opened 2024-01-20
Refactoring to allow more flexible testing

naoyam:
Previously IdModel constructor builds all graphs at once. Now each graph can be built individually. This refactoring is done in preparation for loop promotion analysis to allow testing of each step of the analysis. It would also be useful as sometimes not all graphs would be necessary to build.

Stacked on #1627

### DIFF
--- a/csrc/id_model/id_model.cpp
+++ b/csrc/id_model/id_model.cpp
@@ -360,6 +360,8 @@ void IdModel::buildExactGraph(const std::vector<Expr*>& exprs) {
     // TODO: Revisit if we really should map domains in the exact map
     mapThroughLoopSwizzles(idGraph(IdMappingMode::EXACT));
   }
+
+  idGraph(IdMappingMode::EXACT).validateConsistency();
 }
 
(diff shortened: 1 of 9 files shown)


naoyam:
!build

csrc/id_model/id_model.cpp
```diff
@@ -541,82 +591,93 @@ void IdModel::initializeLoopMap(const StatefulInliningInfo& info) {
...
-  if (validate) {
+  buildExactGraph();
+  if (validate_) {
     validator->checkExactGraphEquivalence(idGraph(IdMappingMode::EXACT));
   }
 
-  buildAlmostExactMap();
-  if (validate) {
+  // Make sure there's no self mapping in TensorView's during lowering
+  // that would invalidate lowering assumptions.
```
jacobhinkle, line 634: It might be helpful to mention that this only checks for Exact self-mappings.

jacobhinkle (approved):
LGTM

PR #1566, opened 2023-12-28
[WIP] Fusion IR Serialization Tracking

rdspring1:
The goal of this PR is to serialize the IR nodes, so we can store the concretized fusion, the scheduled fusion, and the lowered kernel.

PR #1649, opened 2024-01-21
Refactor creation and update of `FusionHeuristics` in `FusionKernelRuntime`

rdspring1:
This PR seeks to avoid accumulating all statements in the segmenter's complete fusion. Instead of pseudo-segmenting the complete fusion, we create a separate fusion for each segmented group. Any new statements are stored in the segmented group's fusion. For `test_issue1277`, serialization handles ~147K values and ~71K expressions with current ToT. With this PR, we only serialize 28K values and 7K expressions. On my local machine, the running time of `test_python_frontend` drops by 28% from ~124s to ~89s.

`SegmentedFusion::makeInitialHeuristic` creates the heuristic for each segmented group. It uses the `FusionSegmentGuard` to pseudo-segment the complete fusion. The values and expressions created when making the heuristics are all stored in the same fusion. For complex fusions with several segments, a significant amount of statements can be accumulated.

`test_issue1277` in `test_python_frontend.py` is a pathological case. The concretized fusion contains 1.5K values and 300 expressions. After creating the heuristics, the complete fusion has 10K values and 5K expressions. In `FusionKernelRuntime::compileKernel`, we create a segmented fusion that copies all the values and expressions from the complete fusion. In this test, segmentation creates 14 segmented groups, so serialization handles ~147K values and ~71K expressions.

Related to: https://github.com/NVIDIA/Fuser/pull/1566


rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

wujingyue:
Sorry for the delay -- I'm still reviewing. The problem makes sense and the numbers look great. It would be great to add to the PR description an overview of the **solution** as well. Thank you!

csrc/fusion_segmenter.cpp
```diff
@@ -510,7 +530,7 @@ void SegmentedFusion::deserialize(const serde::SegmentedFusion* buffer) {
       buffer->num_vals(),
       " values.");
   NVF_ERROR(
-      complete_fusion_->exprs().size() <= buffer->num_exprs(),
+      complete_fusion_->unordered_exprs().size() <= buffer->num_exprs(),
```
jjsjann123, line 533: Is this a necessary change? I thought even after the clone, we would still only serialize expressions leading to outputs, but not necessarily everything in the IR.

csrc/kernel_cache.h
```diff
@@ -222,7 +222,8 @@ class FusionKernelRuntime {
   using HeuristicsPtr = std::unique_ptr<FusionHeuristics>;
   std::optional<HeuristicsPtr> getMaybeHeuristicsFor(
       const KernelArgumentHolder& args,
-      std::optional<PrimDataType> forced_index_type = std::nullopt);
+      std::optional<PrimDataType> forced_index_type,
+      bool initial_heuristics);
```
jjsjann123, line 226: nitpick, maybe add default for `forced_index_type` as well as `initial_heuristics` so we don't have to go change the benchmark files.

csrc/kernel_cache.h
```diff
@@ -222,7 +222,8 @@ class FusionKernelRuntime {
   using HeuristicsPtr = std::unique_ptr<FusionHeuristics>;
   std::optional<HeuristicsPtr> getMaybeHeuristicsFor(
       const KernelArgumentHolder& args,
-      std::optional<PrimDataType> forced_index_type = std::nullopt);
```
jjsjann123, line 225: nitpick: HeuristicsPtr can be used for declaration of `heuristics_`

csrc/kernel_cache.cpp
```diff
@@ -975,14 +975,11 @@ FusionKernelRuntime::FusionKernelRuntime(
...
-
-  // Run segmentation on the copied fusion
+  // SchedulerRuntimeInfo modifies the fusion, so it is required for both
+  // compile paths.
+  std::vector<TensorView*> all_tvs = ir_utils::allTvs(fusion.get());
   SchedulerRuntimeInfo runtime_info(
-      fusion.get(), args, nullptr, all_tvs_, forced_index_type);
-
-  // Initialize the evaluator simplifer
-  precomputed_values_ = std::make_unique<PrecomputedValues>(fusion.get());
```
jjsjann123, line 985: QQ: I'm wondering what are we losing here from now caching `precomputed_values_` for each segments.

wujingyue:
> Sorry for the delay -- I'm still reviewing. The problem makes sense and the numbers look great. It would be great to add to the PR description an overview of the **solution** as well. Thank you!

AFAIU, with this PR, we will clone each segmented group twice: once via `SegmentedGroup::getFusion` when making initial heuristics, and once via `SegmentedFusion::makeFusion`. Currently, we clone it only once (via `SegmentedFusion::makeFusion`). Do you see a time win without serialization? I'm curious about the overhead of this extra cloning.

rdspring1:
Perhaps the solution wasn't clear in the first paragraph. I'll try to expand on the solution.

> Instead of pseudo-segmenting the complete fusion, we create a separate fusion for each segmented group. Any new statements are stored in the segmented group's fusion.

Currently in fusion segmentation, we create a `FusionSegmentGuard` that temporarily narrows a fusion by changing its inputs and outputs. A new `IrContainer` isn't actually created, but the original fusion acts like the segmented fusion.

The solution in this PR is to create a new `IrContainer` for each segmented fusion, which is stored in each `SegmentedGroup` object. When creating the scheduler heuristics, any new values are stored in the separate `IrContainer` rather than the original fusion.

A key change for this solution is that we need to generate the argument metadata for each segment. This PR takes a similar approach to `FusionKernelRuntime::compileFusionParallel`. We use expression evaluator to get the metadata for the segment's outputs and pass that as the next segments inputs. See https://github.com/NVIDIA/Fuser/blob/main/csrc/kernel_cache.cpp#L1255-L1303 for details.

wujingyue:
> Perhaps the solution wasn't clear in the first paragraph. I'll try to expand on the solution.
> 
> > Instead of pseudo-segmenting the complete fusion, we create a separate fusion for each segmented group. Any new statements are stored in the segmented group's fusion.

Thanks for clarifying! I thought that was describing the problem... My bad.

rdspring1:
> AFAIU, with this PR, we will clone each segmented group twice: once via SegmentedGroup::getFusion when making initial heuristics, and once via SegmentedFusion::makeFusion. Currently, we clone it only once (via SegmentedFusion::makeFusion). Do you see a time win without serialization? I'm curious about the overhead of this extra cloning.

ToT running time without serialization is 14.120s. This PR running time without serialization is 13.054s. About 7.5% improvement.

The second clone was necessary because the scheduler heuristics require an unscheduled fusion. During compilation, the cloned fusion is scheduled before being lowered to the CUDA kernel. With dynamic inputs, we compare heuristics to determine if we can reuse a compiled kernel.

wujingyue:
> The second clone was necessary because the scheduler heuristics require an unscheduled fusion

Yes I understood that. I was worried about the first clone by `SegmentedGroup::getFusion`, which this PR adds. It's now less worrisome given you saw a 7.5% win even without serialization. 

> The cost of the second clone is minimized because complete fusion won't contain so many values.

Get it. That probably explains the 7.5% win. `SegmentedFusion::makeFusion` copies the *complete* fusion and redirects the inputs and outputs. So a smaller complete fusion makes the second clone cheaper.

wujingyue:
> The values and expressions created when making the heuristics are all stored in the same fusion.

@rdspring1 , since you debugged this all through, can you point me to where in code these values and expressions are created? I'm trying to understand the problem better.

naoyam:
> a significant amount of statements can be accumulated.

Is this because the schedulers add vals and exprs during the `canSchedule` check?

rdspring1:
@wujingyue The functions that generate majority of the statements are [getContigMergeOfInnerSize](https://github.com/NVIDIA/Fuser/blob/main/csrc/scheduler/vectorize_helper.cpp#L686-L687) and [simplifyExpr](https://github.com/NVIDIA/Fuser/blob/main/csrc/scheduler/vectorize_helper.cpp#L760). They're related to [getVectorizationFactor](https://github.com/NVIDIA/Fuser/blob/main/csrc/scheduler/vectorize_helper.cpp#L797-L802) for the scheduler heuristics.

> Is this because the schedulers add vals and exprs during the `canSchedule` check?

In this case, statements are being added when making the scheduler entry in [SchedulerEntry::makeEntry](https://github.com/NVIDIA/Fuser/blob/main/csrc/scheduler/registry.cpp#L216-L220). The heuristic is already determined for each `SegmentedGroup`.

csrc/kernel_cache.cpp
```diff
@@ -1585,37 +1586,102 @@ void FusionKernelRuntime::updateHeuristicsLaunchParams(
...
+    for (auto input : group_to_run->inputs()) {
+      group_runtime_inputs.push(*args_manager.checkTensorMap(input));
     }
-    auto scheduler_entry = std::move(maybe_scheduler_entry.value());
-    if (!scheduler_entry->sameAs(
-            heuristics_->heuristicsList()[group_index].get())) {
-      return std::nullopt;
+
+    // Create PrecomputedValues initialized with original fusion inputs
+    std::unique_ptr<PrecomputedValues> evaluator_precomputed_values =
```
wujingyue, line 1632:
```suggestion
    auto evaluator_precomputed_values =
```

csrc/kernel_cache.cpp
```diff
@@ -1585,37 +1586,102 @@ void FusionKernelRuntime::updateHeuristicsLaunchParams(
...
+
+  // ArgumentManager manipulates the KernelArgumentHolder argument.
+  // We make another metadata copy for the PrecomputedValues.
+  KernelArgumentHolder complete_fusion_metadata_args(args_metadata);
+
+  ArgumentManager args_manager(
+      args_metadata, runtime_workspace_, segmented_fusion_->inputs());
+
+  // Follow group run order
+  for (int64_t group_id = 0; group_id < num_groups; ++group_id) {
```
wujingyue, line 1617: Consider c10::irange

csrc/kernel_cache.cpp
```diff
@@ -1585,37 +1586,102 @@ void FusionKernelRuntime::updateHeuristicsLaunchParams(
...
+    auto group_to_run = runtime_workspace_.group_run_order.at(group_id);
+
+    // Create fusion for this segmented group
+    Fusion* fusion_to_run = group_to_run->getFusion();
```
wujingyue, line 1621: Cleanup suggestions: given this cloning is for `makeInitialSchedulerEntry` and `getMaybeSchedulerEntry`, can we move cloning to those methods? This way, we don't need to add `fusion_to_run` to the interface of SegmentedFusion.

csrc/fusion_segmenter.h
```diff
@@ -236,6 +242,12 @@ class SegmentedGroup {
 
   //! SegmentedFusion this group belongs to
   SegmentedFusion* segmented_fusion_;
+
+  //! The cloned segmented fusion
+  std::unique_ptr<Fusion> cloned_fusion_;
```
wujingyue, line 247: Cleanup suggestions: can we move this and `original_inputs_in_cloned_fusion_` to SegmentedFusion? SegmentedGroup has been a minimal representation for a subgraph view into the original fusion. Therefore, SegmentedFusion owns a list of SegmentedGroups but a SegmentedGroup only refers to the original fusion. With this change as is, a SegmentedGroup owns also a cloned fusion, a bit too complicated for what it's designed for.

wujingyue (commented):
Thanks again for providing enough context, @rdspring1 ! 

I'm sending out my first round of reviews. Hopefully, addressing them will make it easier for me to review this PR further.

csrc/fusion_segmenter.cpp
rdspring1: In this case, we're not serializing the `complete_fusion_` but the mapping from `complete_fusion_` to segmented group. The deserialized `complete_fusion_` should match the original up to the expected number of expressions.

`IrContainer::deterministic_exprs_map` should match `IrContainer::unordered_exprs` rather than `Fusion::exprs`.

csrc/fusion_segmenter.h
```diff
@@ -265,6 +284,13 @@ class FusionHeuristics {
   FusionHeuristics(const FusionHeuristics&) = delete;
   FusionHeuristics& operator=(const FusionHeuristics&) = delete;
 
+  SchedulerEntryOwningPtr& at(int index) {
+    NVF_ERROR(is_segmented_);
+    NVF_ERROR(index >= 0);
+    NVF_ERROR(index < (int)heuristics_.size());
```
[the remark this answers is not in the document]
rdspring1: I'll remove it. Sometimes I'm lazy to run `gdb` with `catch throw`

rdspring1:
!build

csrc/kernel_cache.cpp
rdspring1: `fusion_to_run` is required for `PrecomputedValues`, `SchedulerRuntimeInfo`, and `FusionExecutor::inferOutputSizes`, which are outside of `makeInitialSchedulerEntry`.

csrc/kernel_cache.cpp
rdspring1: We probably have to recompute values common to each segmented fusion.

csrc/kernel_cache.cpp
```diff
@@ -1584,37 +1585,102 @@ void FusionKernelRuntime::updateHeuristicsLaunchParams(
...
+        group_runtime_inputs,
+        evaluator_precomputed_values.get(),
+        all_tvs_for_fusion_to_run,
+        forced_index_type);
+
+    if (initial_heuristics) {
+      // Add new scheduler entry for this segmented group
+      opt_heuristics.value()->at(group_to_run->groupId()) =
+          segmented_fusion_->makeInitialSchedulerEntry(
+              fusion_to_run, group_to_run, fusion_to_run_info);
```
wujingyue, line 1653: Can `makeInitialSchedulerEntry` and `getMaybeSchedulerEntry` below access `fusion_to_run` via `fusion_to_run_info->fusion()`? This way, we don't need to change their interface to take a Fusion.

rdspring1: I also moved `getMaybeSchedulerEntry` back to `SegmentedGroup`.

rdspring1:
!build

csrc/executor.h
```diff
@@ -68,6 +68,13 @@ class FusionExecutor : public NonCopyable {
       Fusion* fusion,
       const KernelArgumentHolder& args);
 
+  //! Only a single PrecomputedValues is active for a fusion. Consequently, it
+  //! can be necessary to pass PrecomputedValues within a given scope.
+  KernelArgumentHolder inferOutputSizes(
+      Fusion* fusion,
+      const KernelArgumentHolder& args,
+      PrecomputedValues* evaluator_precomputed_values);
```
wujingyue, line 76: Cleanup suggestion: we can probably combine this and the flavor before by allowing this parameter to be nullptr.

csrc/fusion_segmenter.cpp
```diff
@@ -1861,7 +1882,12 @@ std::unique_ptr<Fusion> SegmentedFusion::makeFusion(SegmentedGroup* sg) {
   // new Vals so that they can be bound to the segment inputs.
   convertInputRfactorsToRoots(fusion_segment.get());
 
-  return fusion_segment;
+  return std::make_pair(complete_to_segment_map, std::move(fusion_segment));
+}
+
+std::unique_ptr<Fusion> SegmentedFusion::makeFusion(SegmentedGroup* sg) {
```
wujingyue, line 1888: Nit: Given `makeFusion` is not any faster than `makeFusionWithCloner`, I'd just keep one version, name it `makeFusion` as is, and let it always return a pair. The user can easily 

```
auto&& [ir_cloner, fusion] = makeFusion(sg);
```

or 

```
makeFusion(sg)[1];
```

This avoids an unnecessary overloading, and makes it clear to the user that `makeFusion` creates an IrCloner even if it's not used.

csrc/kernel_cache.cpp
```diff
@@ -1584,37 +1585,102 @@ void FusionKernelRuntime::updateHeuristicsLaunchParams(
...
-    auto group = segmented_fusion_->groups()[group_index];
-
-    auto maybe_scheduler_entry = group->getMaybeSchedulerEntry(runtime_info);
-    if (!maybe_scheduler_entry.has_value()) {
-      return std::nullopt;
+
+  // The runtime group run order is different from the segmented_fusion group
+  // order. Instead of using FusionHeuristics::emplaceBack, we initialize
+  // FusionHeuristics with the desired number of groups.
+  std::optional<FusionKernelRuntime::HeuristicsPtr> opt_heuristics;
```
wujingyue, line 1595: The `std::optional` wrapper doesn't look necessary. All uses expect it to have a value.

csrc/kernel_cache.cpp
```diff
@@ -1012,7 +1009,9 @@ FusionKernelRuntime::FusionKernelRuntime(
     segmented_fusion_->deserialize(serde_buffer->segmented_fusion());
   }
 
-  heuristics_ = segmented_fusion_->makeInitialHeuristics(args, runtime_info);
+  // Pre-compute the executor order so that the run time path
+  //  would go directly to kernel launch.
+  prepareRuntimeOrder();
```
wujingyue, line 1014: I see this is moved before making initial heuristics. What's the reason for that?

csrc/fusion_segmenter.cpp
rdspring1: `Fusion::exprs` doesn't contain all expressions necessary for the fusion. We need `SegmentCandidateFinder::resolveScalarsInGroup` to find expressions related to the fusion's scalars.

csrc/kernel_cache.cpp
rdspring1: In `getMaybeHeuristicsFor`, we're following the `group_run_order` created in `prepareRuntimeOrder` so it needs to occur first before generating the heuristics.

Reference:
```cpp
// Follow group run order
for (int64_t group_id : c10::irange(num_groups)) {
    auto group_to_run = runtime_workspace_.group_run_order.at(group_id);
    ...
}
```

csrc/fusion_segmenter.h
rdspring1: I see your point about keeping `SegmentedGroup` a simple class.  However, it is more cumbersome to serialize. 

Your suggestion would mean we'd have a `std::unordered_map<SegmentedGroup*, ClonedFusionData>` in `SegmentedFusion`. For serialization, we'd separate the keys and values into separate lists because maps aren't serializable. The pointer values would be mapped to an integer id that we'd check for during deserialization. It is much easier to have data contained in the class it is associated with rather than mapping through pointers.

rdspring1:
!build

wujingyue:
Do you want me to review this again? You didn't click the cycle arrow that says "re-request review", so I assumed you are still working on the PR.

rdspring1:
@wujingyue The PR is ready for review. 

The CI failures are unrelated to the PR. `RuntimeError: CUDNN_BACKEND_ENGINEHEUR_DESCRIPTOR: GetAttribute CUDNN_ATTR_ENGINEHEUR_RESULTS Failed cudnn_status: CUDNN_STATUS_SUCCESS`

rdspring1:
!build

jjsjann123:
What's with the failing examples?

csrc/fusion_segmenter.h
```diff
@@ -266,6 +295,11 @@ class FusionHeuristics {
   FusionHeuristics(const FusionHeuristics&) = delete;
   FusionHeuristics& operator=(const FusionHeuristics&) = delete;
 
+  SchedulerEntryOwningPtr& at(int index) {
+    NVF_ERROR(is_segmented_);
```
jjsjann123, line 299: nitpick: is it necessary to assert here?

For complete fusion, `heuristics_` still has one entry here right?!

jjsjann123: in other words, if we do intend to insert, maybe put an error message would make more sense.

csrc/kernel_cache.cpp
```diff
@@ -1586,37 +1589,101 @@ void FusionKernelRuntime::updateHeuristicsLaunchParams(
...
-    auto maybe_scheduler_entry = group->getMaybeSchedulerEntry(runtime_info);
-    if (!maybe_scheduler_entry.has_value()) {
-      return std::nullopt;
+
+  // The runtime group run order is different from the segmented_fusion group
+  // order. Instead of using FusionHeuristics::emplaceBack, we initialize
+  // FusionHeuristics with the desired number of groups.
+  FusionKernelRuntime::HeuristicsPtr heuristics;
+  const int64_t num_groups = (int64_t)runtime_workspace_.group_run_order.size();
+  heuristics = std::make_unique<FusionHeuristics>(num_groups);
```
jjsjann123, line 1601: nitpick, move line 1599 down here? `auto heuristics = std::make_unique<FusionHeuristics>(num_groups);`

csrc/kernel_cache.cpp
```diff
@@ -1586,37 +1589,101 @@ void FusionKernelRuntime::updateHeuristicsLaunchParams(
...
+
+  // Store metadata copy of arguments for ArgumentManager
+  KernelArgumentHolder args_metadata;
+  std::transform(
+      args.cbegin(),
+      args.cend(),
+      args_metadata.getBackInserter(),
+      convertMetadataArg);
```
jjsjann123, line 1609: nitpick: `convertMetadataArg` is only used in combination with `std::transform` and later `setDeviceIndex`.

Might as well wrap them together in a util `void copyMetadataArg(KernelArgumentHolder& dst, const KernelArgumentHolder& src)`

jjsjann123 (commented):
minor comment.

I'm still curious on what's the failing CI. I don't spot anything affecting the build of examples.... I guess is there any API changes that causes that??!!

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

csrc/fusion_segmenter.h
rdspring1: I agree that the assertion isn't necessary.

rdspring1:
!build

rdspring1:
@jjsjann123 I fixed the build examples. I needed to add `NVF_API` macros for https://github.com/NVIDIA/Fuser/pull/1324.

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

csrc/kernel_cache.h
```diff
@@ -222,7 +222,8 @@ class FusionKernelRuntime {
...
+      std::optional<PrimDataType> forced_index_type = std::nullopt,
+      bool initial_heuristics = false);
```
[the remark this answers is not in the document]
wujingyue: It would be even better if getMaybeHeuristicsFor can tell by itself whether it's the initial attempt. This might make the internal logic a little complicated but get rid of an external knob, which sounds like a win.

csrc/fusion_segmenter.h
jjsjann123: > making multiple fusions for the same SegmentedGroup

Wondering where's *making multiple fusions for the same segmentedGroup* happening? Weren't there only one SegmentedGroup instance per group and they only make one copy of fusion inside?

wujingyue: There are SegmentedFusion::makeFusion, which always creates a fusion, and SegmentedGroup::getFusion, which yes tries to reuse the cached.

rdspring1:
!build

csrc/fusion_segmenter.h
```diff
@@ -124,6 +124,19 @@ class SegmentedGroup {
     return exprs_;
   }
 
+  //! Returns the complete fusion inputs mapped to this segmented group's fusion
+  const auto& getCompleteFusionInputs() const {
```
wujingyue, line 128: Note to myself: complete fusion inputs are used by FusionKernelRuntime::getMaybeHeuristicsFor, which holds a PrecomputedValues for the complete fusion. FusionKernelRuntime::getMaybeHeuristicsFor is on the critical path when not making the initial heuristics.

csrc/kernel_cache.cpp
```diff
@@ -1598,35 +1601,92 @@ std::optional<FusionKernelRuntime::HeuristicsPtr> FusionKernelRuntime::
...
-    if (!scheduler_entry->sameAs(
-            heuristics_->heuristicsList()[group_index].get())) {
-      return std::nullopt;
+
+    // Create PrecomputedValues initialized with original fusion inputs
+    auto evaluator_precomputed_values =
+        std::make_unique<PrecomputedValues>(fusion_to_run);
+    evaluator_precomputed_values->bindInputs(group_runtime_inputs);
+    evaluator_precomputed_values->bindValues(
+        group_to_run->getCompleteFusionInputs(), complete_fusion_metadata_args);
```
wujingyue, line 1642: evaluator_precomputed_values now describes the cloned fusion segment. Why do we bind the complete fusion's inputs?

wujingyue (commented):
@rdspring1 can you teach me how to best test this change? Run just the unit tests? (I'm experimenting some changes by myself based on this PR).

csrc/multidevice/executor.cpp
```diff
@@ -262,7 +260,7 @@ std::ostream& MultiDeviceExecutor::print() {
       communication_counter++;
     } else {
       debug() << "Compute segment " << compute_segment_counter << ":{\n";
-      auto fusion = staged_fusion_->makeFusion(group);
+      auto&& [cloner, fusion] = staged_fusion_->makeFusion(group);
```
jjsjann123, line 263: nitpick: I see a few places where we are merely trying to print the group.

`makeFusion` clones the fusion, while if we do `group->getFusion()`, we are going to be using a cached one instead?
My question is, is every `group` supposed to have a cached fusion at this/some point and will we be saving some fusion copy then?

csrc/fusion_segmenter.cpp
```diff
@@ -141,6 +141,25 @@ void SegmentedGroup::deserialize(
...
+  // Map inputs for original fusion to the segmented fusion through IrCloner
+  const std::vector<Val*>& complete_inputs =
+      segmented_fusion_->completeFusion()->inputs();
+  original_inputs_in_cloned_fusion_.reserve(complete_inputs.size());
+  std::transform(
+      complete_inputs.begin(),
+      complete_inputs.end(),
+      std::back_inserter(original_inputs_in_cloned_fusion_),
+      [&complete_to_segment_map = ir_cloner](Val* v) {
+        return complete_to_segment_map.clone(v);
```
jjsjann123, line 159: IIUC, `fusion_segment` returned from this function isn't really changed by this clone. since we should have already cloned it in our segment?

Does it mean we can/should be using `group->getFusion()` instead of `SegmentedFusion::makeFusion(group)` across the board?

csrc/kernel_cache.cpp
```diff
@@ -1598,35 +1601,92 @@ std::optional<FusionKernelRuntime::HeuristicsPtr> FusionKernelRuntime::
...
+      if (!maybe_scheduler_entry.has_value()) {
+        return std::nullopt;
+      }
+      // Check if this scheduler entry matches the previous entry for this
+      // segmented group. If no match, then return std::nullptr
+      auto scheduler_entry = std::move(maybe_scheduler_entry.value());
+      if (!scheduler_entry->sameAs(
+              heuristics_->at(group_to_run->groupId()).get())) {
+        return std::nullopt;
+      }
```
jjsjann123, line 1675: nitpick: don't see much point in the move in the middle. maybe combine the two if checks together.

jjsjann123 (commented):
Have some more nitpicks on coding.

@wujingyue seems to have some more insights regarding the overall pipeline. I'll let him stamp.

csrc/kernel_cache.cpp
rdspring1: It is a rare corner case, but there can still be dependencies on the original fusion's arguments.

To reproduce:
1. Disable this line
2. Run `python test/python/test_python_frontend.py -k test_issue1277`

csrc/fusion_segmenter.cpp
rdspring1: > Does it mean we can/should be using group->getFusion() instead of SegmentedFusion::makeFusion(group) across the board?

Only if the downstream function does not modify the segmented fusion. It can be a replacement for print functions but not if you need to modify the function.

wujingyue:
> After creating the heuristics, the complete fusion has 10K values and 5K expressions.

Nice find! I saw that too. 

AFAICT from [this line](https://github.com/NVIDIA/Fuser/blob/86c254dc7a997647a3fa09837f464f87ef8cbdcd/csrc/fusion_segmenter.cpp#L2413), these added Exprs are not even between the original fusion's inputs and outputs and are wasteful to keep anyway. Is it possible to remove them when the guard is destructed?

csrc/kernel_cache.cpp
wujingyue: > We probably have to recompute values common to each segmented fusion.

And [that](https://github.com/NVIDIA/Fuser/pull/1649/files/08cffeea3446f760a47cbfae59cf02bd100036df#diff-69db587a93b7b0fefd57640c68f0f1faaca8ece64173d757d17d8962e502ecd6R1639) is on the critical path, correct? Do you see an increase on the critical path latency with this change?

csrc/fusion_segmenter.cpp
wujingyue: What specifically modifies the fusion and has to be done on the copy? AFAIU, the end state is going to be: after segmentation, 

```
FKR::ctor:
  for each segment:
    create a subfusion for that segment
    heuristics = make initial heuristics for that subfusion

FKR::compileFusionParallel
  for each segment: 
    schedule the associated subfusion according to the initial heuristics
    compile the associated subfusion
```

It'll be unfortunate if we can't reuse the same subfusion for scheduling and compiling and have to copy the fusion.

csrc/kernel_cache.cpp
```diff
@@ -1149,7 +1150,7 @@ void FusionKernelRuntime::deserialize(
     NVF_ERROR(
         !sg || scheduler_entry->heuristic() == sg->heuristic(),
         "Heuristics do not match.");
-    std::unique_ptr<Fusion> fusion_to_run = segmented_fusion_->makeFusion(sg);
+    auto&& [ir_cloner, fusion_to_run] = segmented_fusion_->makeFusion(sg);
```
wujingyue, line 1153: `ir_cloner` doesn't seem to be used, so how about
```suggestion
    auto fusion_to_run = segmented_fusion_->makeFusion(sg).second;
```

wujingyue: Several other places in this PR may be simplified like this.

rdspring1:
!build

wujingyue (approved):
LGTM!

This PR unfortunately introduced several complexities to work around **pre-existing** problems in nvFuser. The following problems are worth investigating and resolving them will probably lead to cleanups: 
1. A potentially easier alternative is to speed up makeFusion, e.g., by copying Exprs only between the segment's inputs and outputs. 
2. It's a leak of abstraction to bind complete fusion inputs in order to schedule and compile a fusion segment: https://github.com/NVIDIA/Fuser/pull/1649/files#r1535101958. Some tests failed when we tried that, but those feel like bugs in other components. 
3. getFusion can be used more extensively to avoid copy, e.g., https://github.com/NVIDIA/Fuser/pull/1649/files#diff-69db587a93b7b0fefd57640c68f0f1faaca8ece64173d757d17d8962e502ecd6R1348. The practical benefit is unclear at this moment, because this PR made fusion segments much smaller.

kevinstephano:
> > After creating the heuristics, the complete fusion has 10K values and 5K expressions.
> 
> Nice find! I saw that too.
> 
> AFAICT from [this line](https://github.com/NVIDIA/Fuser/blob/86c254dc7a997647a3fa09837f464f87ef8cbdcd/csrc/fusion_segmenter.cpp#L2413), these added Exprs are not even between the original fusion's inputs and outputs and are wasteful to keep anyway. Is it possible to remove them when the guard is destructed?

Were you able to remove the Expr and Vals as Jingyue suggested when the guard is destructed?  I could not find if there was a direct response.

rdspring1:
> Were you able to remove the Expr and Vals as Jingyue suggested when the guard is destructed? 

IIRC `FusionSegmentGuard` removes any casts from the Fusion and reverts the modified inputs and outputs back to fusion's originals. Removing the Exprs and Vals added by the scheduler heuristics would requires some extra bookkeeping. IMO those changes are beyond the scope of this PR. It is a good idea for a follow-up PR.

csrc/kernel_cache.h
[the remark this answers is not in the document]
rdspring1: We determine `initial_heuristics=True` when `heuristics_ == nullptr`. The argument was removed.

csrc/kernel_cache.cpp
```diff
@@ -1596,37 +1600,95 @@ void FusionKernelRuntime::updateHeuristicsLaunchParams(
 std::optional<FusionKernelRuntime::HeuristicsPtr> FusionKernelRuntime::
     getMaybeHeuristicsFor(
         const KernelArgumentHolder& args,
-        std::optional<PrimDataType> forced_index_type) {
+        std::optional<PrimDataType> forced_index_type,
+        bool initial_heuristics) {
```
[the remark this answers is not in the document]
rdspring1: `initial_heuristics` was removed.

csrc/evaluator_common.h
```diff
@@ -142,12 +142,21 @@ class NaiveValueMachine {
 class PrecomputedValues {
  public:
   PrecomputedValues() = delete;
+  NVF_API PrecomputedValues(PrecomputedValues&&) = default;
 
   NVF_API explicit PrecomputedValues(Fusion* fusion);
 
-  //! Bind concrete values from fusion runtime inputs
+  NVF_API ~PrecomputedValues();
```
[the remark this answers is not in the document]
rdspring1: I was seeing this `undefined symbol: _ZN7nvfuser17PrecomputedValuesD1Ev` in the build tests without adding `NVF_API`.

rdspring1:
!build

rdspring1:
!build

PR #1650, opened 2024-01-21
IdModel: Loop Promotion Analysis Step 1: Build a map of broadcast resolutions for root domains

naoyam:
This is the first step of the [loop promotion analysis](https://github.com/NVIDIA/Fuser/blob/IterDomainGraphs/csrc/id_model/id_model.cpp#L1032). It may not be very clear what this analysis is intended to do at this point, but in short, the root resolution information is propagated to intermediate and leaf ID groups in the IEL graph, which is then projected back to the loop graph. I thought before diving into the full promotion analysis it would make reviewing easier to split this part off as the first sub-PR with unit tests.

There's a detailed design doc for the overall loop promotion analysis. DM me.


naoyam:
!build

naoyam:
!build

naoyam:
!build

naoyam:
!build

naoyam:
!build

csrc/id_model/id_model.cpp
```diff
@@ -540,6 +558,41 @@ void IdModel::buildPermissiveGraph() {
   idGraph(IdMappingMode::PERMISSIVE).validateConsistency();
 }
 
+namespace {
+
+// Returns the root producer iteration domains that are resolved by provided
+// consumer
+std::unordered_map<IterDomain*, IterDomain*> resolvedRootBroadcasts(
```
zasdfgbnm, line 565: If the only use for this map is to iterate through it, should we instead use a `std::vector<std::pair<IterDomain*, IterDomain*>>` which is both faster and deterministic?

csrc/id_model/id_model.cpp
```diff
@@ -611,16 +671,187 @@ void IdModel::buildLoopGraph() {
...
+          << "Invalid multiple broadcast resolution within shared loops detected, group:\n  "
+          << iel_group->toString() << "\nIs being broadcasted to:";
+      for (const ValGroup& entry : loop_exact_resolved_intersection) {
+        err_msg << "\n  " << entry->toString();
+      }
+      NVF_ERROR(false, err_msg.str());
+    }
+
+    const ValGroup& exact_resolution_group =
+        loop_exact_resolved_intersection.front();
```
zasdfgbnm, line 820: Should this be just:

```C++
if (resolved_exact_groups.size() > 1) {
  NVF_ERROR(false, err_msg.str());
}
const ValGroup& exact_resolution_group = resolved_exact_groups.front();
```
?

csrc/val_graph.h
```diff
@@ -89,6 +90,32 @@ class ValGraph {
...
+          std::is_base_of<Expr, ElementType>::value>>
+  typename std::conditional<
+      std::is_base_of<Val, ElementType>::value,
+      ValGroups,
+      ExprGroups>::type
+  toGroups(const ContainerType& entries) const {
+    using RetType = typename std::conditional<
+        std::is_base_of<Val, ElementType>::value,
+        ValGroups,
+        ExprGroups>::type;
```
zasdfgbnm, line 111: nit, instead of copy-pasting this definition of `RetType` twice, could we just:

```C++
template <typename ContainerType, typename ElementType = ..., typename RetType = ..., typename = SFINAE>
RetType toGroups(const ContainerType& entries) const {
```

zasdfgbnm (approved):
Finished reading, don't see any blocking issue.

csrc/id_model/id_model.cpp
naoyam: Not necessarily, I believe. `resolved_exact_groups` may contain non-broadcast domains where the broadcast domain is not inlined. The loop graph is used to exclude non-inlined cases.

When `loop_exact_resolved_intersection.size()` is 1, `resolved_exact_groups.size()` should also be 1. However, `loop_exact_resolved_intersection.size()` is 0, `resolved_exact_groups.size()` can be greater than 1. For example:

```
auto t0 = [b0];
fusion.addInput(t0);
auto t1 = [i1];
fusoin.addInput(t1);
auto t2 = [i2];
fusion.addInput(t2);
// i1 != i2

auto t3 = t0 + t1
fusion.addOutput(t3);

auto t4 = t0 + t2
fusion.addOutput(t4);
```

This is a valid program. The `b0` broadcast domain is a non-unique broadcast domain. `resolved_exact_groups` would have two groups: `{i1}` and `{i2}`, but  `loop_exact_resolved_intersection` should be empty.

PR #1651, opened 2024-01-22
print has segment hints

liqiangxl:
when segmented due to `hasSegmentHints`, print out why.


jjsjann123 (approved):
Thanks for adding the debug print

PR #1661, opened 2024-01-24
transpose scheduler fix: reduction IterDomain on input tensors

### LINKED ISSUE #1659: Transpose scheduler merges IterDomains with different iteration types.
repro script
```
import torch

    T0 = fd.define_tensor(shape=[-1, -1, 1], contiguity=[True, True, None], dtype=DataType.Float, is_cpu=False, stride_order=[2, 1, 0])
    T1 = fd.define_tensor(shape=[-1, 1, -1], contiguity=[True, None, True], dtype=DataType.Float, is_cpu=False, stride_order=[2, 1, 0])
    T2 = fd.ops.sum(T0, axes=[1], keepdim=False, dtype=DataType.Null)
    T3 = fd.ops.sum(T1, axes=[1], keepdim=False, dtype=DataType.Null)
    T4 = fd.ops.mul(T2, T3)
    fd.add_output(T4)

    nvfuser_fusion_id0(fd)

inputs = [
    torch.randn((524288,), dtype=torch.float32, device='cuda:0').as_strided((1024, 512, 1), (512, 1, 1)),
    torch.randn((524288,), dtype=torch.float32, device='cuda:0').as_strided((1024, 1, 512), (512, 512, 1)),
]
fd.execute(inputs)
```

Running into issues:
```
  File "/opt/pytorch/nvfuser/nvfuser/__init__.py", line 137, in execute
    result = self._execute(
RuntimeError: Merging IterDomains requires that their iteration types match. Outer: iS69{32}, Inner: rS7{i1}
Exception raised from merge at /opt/pytorch/nvfuser/csrc/ir/nodes.cpp:2692 (most recent call first):
frame #0: nvfuser::nvfCheckFail(char const*, char const*, unsigned int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) + 0xf3 (0x7fd2d0278a43 in /opt/pytorch/nvfuser/nvfuser/lib/libnvfuser_codegen.so)
frame #1: nvfuser::IterDomain::merge(nvfuser::IterDomain*, nvfuser::IterDomain*, bool) + 0x39b (0x7fd2d04fd27b in /opt/pytorch/nvfuser/nvfuser/lib/libnvfuser_codegen.so)
frame #2: nvfuser::TensorDomain::merge(int, int) + 0xc9 (0x7fd2d04fd3d9 in /opt/pytorch/nvfuser/nvfuser/lib/libnvfuser_codegen.so)
frame #3: nvfuser::TensorView::merge(int, int) + 0xdb (0x7fd2d07cffcb in /opt/pytorch/nvfuser/nvfuser/lib/libnvfuser_codegen.so)
frame #4: nvfuser::scheduleTranspose(nvfuser::Fusion*, nvfuser::TransposeParams) + 0x109b (0x7fd2d071fb5b in /opt/pytorch/nvfuser/nvfuser/lib/libnvfuser_codegen.so)
frame #5: nvfuser::TransposeScheduler::schedule(nvfuser::Fusion*) + 0xe0 (0x7fd2d0721db0 in /opt/pytorch/nvfuser/nvfuser/lib/libnvfuser_codegen.so)
frame #6: nvfuser::FusionKernelRuntime::compileKernel(nvfuser::KernelArgumentHolder const&, nvfuser::SegmentedGroup*) + 0x151 (0x7fd2d0533ab1 in /opt/pytorch/nvfuser/nvfuser/lib/libnvfuser_codegen.so)
frame #7: nvfuser::FusionKernelRuntime::compileFusionParallel(nvfuser::KernelArgumentHolder) + 0x41c (0x7fd2d0539b9c in /opt/pytorch/nvfuser/nvfuser/lib/libnvfuser_codegen.so)
frame #8: nvfuser::FusionExecutorCache::runFusionWithInputs(c10::ArrayRef<c10::IValue> const&, std::optional<nvfuser::PrimDataType>, std::optional<signed char>) + 0xa43 (0x7fd2d0545523 in /opt/pytorch/nvfuser/nvfuser/lib/libnvfuser_codegen.so)
frame #9: nvfuser::python_frontend::FusionDefinition::execute(c10::ArrayRef<c10::IValue> const&, bool, bool, std::optional<signed char>) const + 0x16c (0x7fd2d0854a2c in /opt/pytorch/nvfuser/nvfuser/lib/libnvfuser_codegen.so)
frame #10: <unknown function> + 0xf152e (0x7fd2d0ba252e in /opt/pytorch/nvfuser/nvfuser/_C.cpython-310-x86_64-linux-gnu.so)
frame #11: <unknown function> + 0x177488 (0x7fd2d0c28488 in /opt/pytorch/nvfuser/nvfuser/_C.cpython-310-x86_64-linux-gnu.so)
<omitting python frames>
frame #27: <unknown function> + 0x29d90 (0x7fd4f73bbd90 in /usr/lib/x86_64-linux-gnu/libc.so.6)
frame #28: __libc_start_main + 0x80 (0x7fd4f73bbe40 in /usr/lib/x86_64-linux-gnu/libc.so.6)
```

jjsjann123:
Fixes #1659 

Reorders trivial reduction IterDomain so it won't interfere with scheduling tiling from transpose scheduler.


jjsjann123:
!build

jjsjann123:
!build

zasdfgbnm:
Why are there still trivial reductions in the fusion? Shouldn't trivial reduction be automatically converted into squeeze?

zasdfgbnm:
> Why are there still trivial reductions in the fusion? Shouldn't trivial reduction be automatically converted into squeeze?

I think you are referring to non-trivial reductions that appears in fusion inputs?

jjsjann123:
> > Why are there still trivial reductions in the fusion? Shouldn't trivial reduction be automatically converted into squeeze?
> 
> I think you are referring to non-trivial reductions that appears in fusion inputs?

oops, thanks for correcting me. I think I mean that.

So it's a real reduction from its producer segment. but in the consumer segment, it's just a dangling reduction iterdomain that doesn't map to anything and wouldn't participate in indexing neither. :stuck_out_tongue: I thought I can call it trivial since it's not doing anything. Totally forgot that we can actually just reduce a size-1 dimension and turn that into a squeeze, aka trivial reduction :facepalm:

csrc/scheduler/transpose.cpp
```diff
@@ -127,6 +127,26 @@ void TransposeScheduler::computeHeuristics(
...
+void cleanInnerNDLeafDomain(TensorView* tv, int n) {
+  if (!tv->isFusionInput()) {
+    return;
+  }
+
+  std::unordered_map<int, int> old2new;
+
+  for (int i = 0; i < n; i++) {
+    if (tv->axis(-1 - i)->isReduction()) {
+      old2new[-1 - i] = i;
```
zasdfgbnm, line 143: Instead of making it `i`, should we have a counter of the number of reductions we have seen, and use that counter as index?

jjsjann123: oh I think that's what I was trying to do, but then forgot that when I change the code to take `n` as an argument :face_with_head_bandage:

jjsjann123:
!build

csrc/scheduler/transpose.cpp
```diff
@@ -1236,6 +1257,7 @@ void scheduleTranspose(Fusion* fusion, TransposeParams params) {
 
   int pos = (int)reference2->nDims() - 2;
   // [..., tile1, tile2]
+  cleanInnerNDLeafDomain(reference2, 2);
```
jacobhinkle, line 1260: Might this interfere with `inlineMost`? If you have a setting like this:
```
Inputs:
    T0[ i0 i1 ]
    T1[ i2 ]
Outputs:
    T4[ i0 ]
T2[ i0 i1 ] = - T0[ i0 i1 ]
T3[ i0 r1 ] = sum(T2[ i0 i1 ])
T4[ i0 ] = T3[ i0 r1 ] * T1[ i2 ]
```
I'd expect this to generate these loops:
```
FOR i0
  // Allocate T3
  FOR i1
    T2[ i0 i1 ] = - T0[ i0 i1 ]
    T3[ i0 r1 ] += T2[ i0 i1 ]
  ENDFOR i1
  T4[ i0 ] = T3[ i0 r1 ] * T1[ i2 ]
ENDFOR i0
```
When you reorder it as `T2[ r3 i1 i2 ]` will it break the inlining into multiple outer loops like below?
```
// Allocate T3
FOR i0
  // Allocate T3
  FOR i1
    T2[ i0 i1 ] = - T0[ i0 i1 ]
  ENDFOR i1
ENDFOR i0
FOR i0
  FOR i1
    T3[ i0 r1 ] += T2[ i0 i1 ]
  ENDFOR i1
ENDFOR i0
FOR i0
  T4[ i0 ] = T3[ i0 r1 ] * T1[ i2 ]
ENDFOR i0
```

jacobhinkle (commented):
I have one question about the effect of this reordering on inlining.

csrc/scheduler/transpose.cpp
jacobhinkle: Sorry I know this wasn't a transpose example. Is there something about the fusions this scheduler accepts that means we can ignore this?

jjsjann123: :stuck_out_tongue_closed_eyes: I don't know enough about how indexing works to answer this one.

```
void cleanInnerNDLeafDomain(TensorView* tv, int n) {
  if (!tv->isFusionInput()) {
    return;
  }
```

IIUC, we are only re-ordering this one if `it's a fusion input`. And my impression here is also that, in protocol it's also called on reference tensor and then propagated out. I was expecting that'll give us a consistent ordering for its consumers as well, even though this iterdomain doesn't map to anything afterwards.

jacobhinkle: Ah right. Of course, since it's the reference tensor it will be propagated out so you don't need to worry about this. Thanks, this looks good to me.

csrc/scheduler/transpose.cpp
```diff
@@ -127,6 +127,27 @@ void TransposeScheduler::computeHeuristics(
 
 namespace {
 
+// propagation could miss trivial reduction iterdomain on input tensors, since
```
naoyam, line 130: As Xiang also commented, trivial reduction could be confusing as it would imply squeeze. Can you please rephrase the comment?

csrc/scheduler/transpose.cpp
```diff
@@ -127,6 +127,27 @@ void TransposeScheduler::computeHeuristics(
...
+// that doesn't have any dependency and doesn't map to anything, we can naively
+// just reorder them so they won't interfere with tiling.
+// See https://github.com/NVIDIA/Fuser/issues/1659#issuecomment-1907053830
+void cleanInnerNDLeafDomain(TensorView* tv, int n) {
```
naoyam, line 134: What does `ND` mean here?

naoyam (commented):
The changes make sense. Just a couple of things to clarify.

csrc/scheduler/transpose.cpp
jjsjann123: I meant to say N-Dimension... I guess I should have just left it as `cleanInnerLeafDomain`

jjsjann123:
!build

test/test_gpu_transpose.cpp
```diff
@@ -1335,4 +1335,56 @@ TEST_F(TransposeTest, TransposeSplitAggregatedVectorizationWidth) {
   NVF_CHECK(ref.equal(cg_outputs.at(0)));
 }
 
+// Testing transpose scheduler to handle trivial reduction IterDomain produced
+// by segmented fusion, see issue 1659 for details
+TEST_F(TransposeTest, TrivialReductionIterDomainOnInputsIssueRepro1659) {
```
zasdfgbnm, line 1340: Please also change the comment here to avoid the term "trivial reduction"

csrc/scheduler/transpose.cpp
zasdfgbnm: Should this be called `moveReductionsOut`?

csrc/scheduler/transpose.cpp
```diff
@@ -127,6 +127,31 @@ void TransposeScheduler::computeHeuristics(
 
 namespace {
 
+// propagation could miss reduction iterdomain on input tensors, which could
+// happen for segmented fusion. Since reduction iterdomains on inputs don't have
```
zasdfgbnm, line 131:
```C++
// If a fusion is segmented, the segmenter will create fusions whose inputs contain reduction IterDomains. These reduction IterDomains on input TensorViews does not have any meaning, and should just be left untouched.
```

zasdfgbnm (approved):
The change makes sense to me. But please improve naming and term usage.

jjsjann123:
!build

naoyam (approved):
LGTM as well

jacobhinkle (approved):

PR #1663, opened 2024-01-24
[Hopper] MBarrier storage scope update

drzejan2:
The purpose of this change is to update the expected scope in which smem placeholder for mbarrier objects are alive before the memory can be reused. Life of such variable starts with its initialization, and ends with its invalidation.

The core change is switching from last write to last read for mbarrier invalidate helper function call.


drzejan2:
!build

zasdfgbnm (approved):
Looks like all test passes.

PR #1666, opened 2024-01-24
fix persistent buffer resolution points detection

### LINKED ISSUE #1665: resolution points of persistent buffers are not correctly identified
**Issue:** In the following fusion we have two resolution points, t4 and t5. Current main branch only detects t4. After project to inputs, tv1 remains a persistent buffer as it cannot be inlined with t5.
image

**Reason:** outputs of expression with resolution point are still on reduction path. **Current code only puts the outputs to persistent path.**

**Fix:** Put outputs ( of expression with resolution point ) to **both reduction path and persistent path**

liqiangxl:
Fix #1665

### DIFF
--- a/csrc/scheduler/utils.cpp
+++ b/csrc/scheduler/utils.cpp
@@ -427,18 +427,17 @@ class PersistentBufferResolution : public IterVisitor {
 
     bool input_on_persistent_buffer_path =
         input_on_persitent_buffer_path_it != expr->inputs().end();
-
     if (input_on_reduction_path && input_on_persistent_buffer_path) {
       // Expression has inputs on both a reduction and persistent buffer path,
       // this is a resolution.
       auto out_tvs = ir_utils::filterByType<TensorView>(expr->outputs());
-
       // Add resolution point
       resolution_points_.insert(
           resolution_points_.end(), out_tvs.begin(), out_tvs.end());
 
-      // Outputs are still on a persistent path
+      // Outputs are still on reduction path and persistent path
       for (auto out : expr->outputs()) {
+        on_reduction_path_.emplace(out);
         on_persitent_buffer_path_.emplace(out);
       }
     } else if (input_on_reduction_path) {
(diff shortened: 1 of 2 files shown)


liqiangxl:
!build --diff

liqiangxl:
!build --diff

liqiangxl:
codegen diffs:
(1) `NVFuserTest.FusionPersistentBufferProjection_CUDA`, due to an additional resolution point detected for buffer t1. In main branch only t6 is t1's resolution point. in this PR, t8 is also a resolution point of t1. No change in scheduler paras. This fusion is also used in `FusionPersistentBufferCalculation4_CUDA` where we need to change the reference value.
image

(2) `NVFuserTest.ProjectToInputsAndBroadcastTvs3`, similar to (1), an additional resolution point is detected. No change in scheduler paras.
(3) `NVFuserTest.CombinedSchedulerLayerNormBackward_CUDA`, same cuda but PTX diff.

liqiangxl:
!build -diff

liqiangxl:
!build --diff

liqiangxl:
!build --diff

liqiangxl:
1 python test failed due to bad luck.
```
00:11:55 E       Mismatched elements: 1 / 1048576 (0.0%)
00:11:55 E       Greatest absolute difference: 0.0013810679083690047 at index (327, 272) (up to 0.001 allowed)
00:11:55 E       Greatest relative difference: 8.137622131698663e-08 at index (327, 272) (up to 0 allowed)
```

mattwalsh:
According to Claude (which I asked "is this PR ready to merge OR redundant OR no longer relevant")

**Assessment**

  This PR is already merged! 🎉

  PR #1666 is showing as "CONFLICTING" and still open, but the changes from it were already committed directly to main on March 6, 2024 as commit ef5d5d7d1 by the same author (Liqiang Lu).

  Comparing the PR diff with the current code in csrc/scheduler/utils.cpp:349-489, I can confirm the changes are identical:
  - The getResolutionPointsOf function now takes persistent_buffers parameter (lines 349-352)
  - The constructor uses it to initialize persistent_buffers_ member (lines 367-374)
  - The logic to skip persistent buffers when marking outputs (lines 450-453)
  - The new member variable persistent_buffers_ (line 488)

  Recommendation: The PR should be closed since its changes are already in main. The PR shows as conflicting because the branch hasn't been updated since the direct commit was made. The work is complete and
  already part of the codebase.

Issue #1668, opened 2024-01-25
`eager` and `torch.compile` benchmarks


Priya2698:
Currently, the python benchmarks only support nvfuser executions. For weekly benchmark coverage, we need support to run `eager` and `torch.compile` as well. This issue is to track progress.

Priya2698:
The existing benchmarks now include eager/torchcompile baselines.
Closing the issue.

PR #1670, opened 2024-01-25
`torch.compile` and `eager` benchmarks for `softmax`

Priya2698:
Adds `torch.compile` and `eager` baseline benchmarks to be used in weekly benchmark runs.
Issue #1668.

Update:
Building off of @jacobhinkle's suggestion:
1. `torch.compile` and `eager` mode are combined in a single benchmark function using parametrization.
2. By default, only nvfuser benchmarks are run. To enable eager/compile, the flags `--benchmark-eager` and `benchmark-torchcompile` are used. 

I'll update the wiki to describe and detail benchmarking options.

Future changes:
In a separate PR, I plan on renaming `disable-benchmarking` to `disable-nvf-benchmarking` since we are adding more executors.

### DIFF
--- a/python_benchmarks/conftest.py
+++ b/python_benchmarks/conftest.py
@@ -45,6 +45,11 @@ def pytest_benchmark_update_machine_info(config, machine_info):
 
 
 def pytest_collection_modifyitems(session, config, items):
+    """
+    The baseline benchmarks use `compile` parameter:
+        compile = false: Eager mode benchmark
+        compile = true: torch.compile benchmark
+    """
     run_eager = config.getoption("--benchmark-eager")
     run_torchcompile = config.getoption("--benchmark-torchcompile")
 
--- a/python_benchmarks/conftest.py
+++ b/python_benchmarks/conftest.py
@@ -15,6 +15,19 @@ def pytest_addoption(parser):
         default=False,
         help="Disable benchmarking.",
     )
+    parser.addoption(
+        "--benchmark-eager",
+        action="store_true",
+        default=False,
+        help="Benchmarks torch eager mode.",
+    )
+
+    parser.addoption(
+        "--benchmark-torchcompile",
+        action="store_true",
+        default=False,
+        help="Benchmarks torch.compile mode.",
+    )
 
 
 @pytest.fixture
@@ -33,3 +46,27 @@ def pytest_make_parametrize_id(val):
 
 def pytest_benchmark_update_machine_info(config, machine_info):
     machine_info.update(DEVICE_PROPERTIES)
+
+
+def pytest_collection_modifyitems(session, config, items):
+    run_eager = config.getoption("--benchmark-eager")
+    run_torchcompile = config.getoption("--benchmark-torchcompile")
+
+    if not run_eager:
+        skip_eager = pytest.mark.skip(reason="need --benchmark-eager option to run")
+        for item in items:
+            # If the benchmark has compile=False parameter (eager mode), skip it.
+            if (
+                "compile" in item.callspec.params
+                and not item.callspec.params["compile"]
+            ):
+                item.add_marker(skip_eager)
+
+    if not run_torchcompile:
+        skip_torchcompile = pytest.mark.skip(
+            reason="need --benchmark-torchcompile option to run"
+        )
+        for item in items:
+            # If the benchmark has compile=True parameter (torch.compile mode), skip it.
+            if "compile" in item.callspec.params and item.callspec.params["compile"]:
+                item.add_marker(skip_torchcompile)
--- a/python_benchmarks/conftest.py
+++ b/python_benchmarks/conftest.py
@@ -6,26 +6,22 @@ def pytest_addoption(parser):
     parser.addoption(
         "--disable-validation",
         action="store_true",
-        default=False,
         help="Disable output validation in benchmarks.",
     )
     parser.addoption(
         "--disable-benchmarking",
         action="store_true",
-        default=False,
         help="Disable benchmarking.",
     )
     parser.addoption(
         "--benchmark-eager",
         action="store_true",
-        default=False,
         help="Benchmarks torch eager mode.",
     )
 
     parser.addoption(
         "--benchmark-torchcompile",
         action="store_true",
-        default=False,
         help="Benchmarks torch.compile mode.",
     )
 
--- a/python_benchmarks/test_softmax_bwd.py
+++ b/python_benchmarks/test_softmax_bwd.py
@@ -73,8 +73,8 @@ def test_softmax_bwd_nvf_benchmark(
     clear_cuda_cache()
 
     inputs = [
-        torch.randn(*size, device="cuda", dtype=dtype, requires_grad=True),
-        torch.randn(*size, device="cuda", dtype=dtype),
+        torch.randn(size, device="cuda", dtype=dtype, requires_grad=True),
+        torch.randn(size, device="cuda", dtype=dtype),
     ]
 
     with FusionDefinition() as fd:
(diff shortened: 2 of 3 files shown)


python_benchmarks/test_softmax_fwd.py
```diff
@@ -63,8 +67,36 @@ def test_softmax_fwd_benchmark(
...
+@pytest.mark.parametrize("reduction_axis", [0, 1])
+def test_softmax_fwd_compile_benchmark(
+    benchmark,
+    size: tuple,
+    dtype: torch.dtype,
+    reduction_axis: int,
+):
+    clear_cuda_cache()
+    input = torch.randn(*size, device="cuda", dtype=dtype)
+    run_benchmark(benchmark, torch.compile(softmax_fwd_fn), [input, reduction_axis])
```
jacobhinkle, line 102: Feel free to ignore, but since these tests are nearly identical is it practical to parametrize a single test here over whether `torch.compile` is used or not? I am thinking something like
```python
@pytest.mark.parametrize("size", generate_input_sizes(dims=2))
@pytest.mark.parametrize("dtype", FLOAT_DTYPES)
@pytest.mark.parametrize("reduction_axis", [0, 1])
@pytest.mark.parametrize("compile", [False, True], ids=["eager", "compile"])
def test_softmax_fwd_compile_benchmark(
    benchmark,
    size: tuple,
    dtype: torch.dtype,
    reduction_axis: int,
    compile: bool,
):
    clear_cuda_cache()
    input = torch.randn(*size, device="cuda", dtype=dtype)
    run_benchmark(
        benchmark,
        torch.compile(softmax_fwd_fn) if compile else softmax_fwd_fn,
        [input, reduction_axis])
```

jacobhinkle: Same for bwd tests above.

jacobhinkle (approved):
LGTM, just minor comments.

python_benchmarks/test_softmax_fwd.py
Priya2698: I considered this. The main reason I split them is to be able to run them independently using filtering flags (`-k eager, compile`). I am looking into if I can filter on the parametrization flag in a clean way.
The other option could be to merge and use flags similar to `disable-validation` and `disable-benchmarking.`

python_benchmarks/test_softmax_bwd.py
```diff
@@ -54,10 +54,15 @@ def softmax_bwd_fusion(
     fd.add_output(T19)
 
 
+def softmax_bwd_fn(inputs: list):  # [in_tensor, output, grads]
+    inputs[1].backward(inputs[2], retain_graph=True)
+    return inputs[0].grad
```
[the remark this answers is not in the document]
Priya2698: Good point, will rename, thanks!

python_benchmarks/test_softmax_fwd.py
jacobhinkle: You can use `-k` with the `ids` that print I believe. In this case the test prints like `test_softmax_fwd_compile_benchmark[eager-0-torch.float32-(16, 768)]`. I guess you'd want to rename the function to `test_softmax_fwd_benchmark` instead, but point being that the `ids` argument to `parametrize` helps to make these filterable.

jacobhinkle: On that topic, I actually had some trouble filtering on shapes because  of how they print with commas. For example try filtering down to only tests that run with size `(16, 6400)`. I still don't know how to use `-k` to do that without modifying the ids. In cases like that, you can use something like this though
```python
@pytest.mark.parametrize("size", generate_input_sizes(dims=2), ids=lambda s: '_'.join(str(si) for si in s))
def test_foo(
    ...
```
This will print names like `python_benchmarks/test_softmax_fwd.py::test_foo[16_6400]` which you can filter to using `pytest foo.py -k "foo and 16_6400]"`.

Priya2698: It is filterable on ids. But I am having to filter the entire test ID: `-k test_softmax_fwd_compile_benchmark[eager-0-torch.float32-(16, 768)]` instead of something like `-k eager`. Although, this can work too (`test_softmax_fwd_compile_benchmark[*eager*]`), my only reluctance is that it is quite verbose and test name dependent. I believe I can simplify it by adding parser options.

Priya2698: > On that topic, I actually had some trouble filtering on shapes because of how they print with commas. For example try filtering down to only tests that run with size `(16, 6400)`. I still don't know how to use `-k` to do that without modifying the ids. In cases like that, you can use something like this though
> 
> ```python
> @pytest.mark.parametrize("size", generate_input_sizes(dims=2), ids=lambda s: '_'.join(str(si) for si in s))
> def test_foo(
>     ...
> ```
> 
> This will print names like `python_benchmarks/test_softmax_fwd.py::test_foo[16_6400]` which you can filter to using `pytest foo.py -k "foo and 16_6400]"`.

That's a good point, let me see if I can change this without needing significant changes in the CI. @xwang233 has scripts to process the outputs generated by pytest and assumes the sizes to be integers at the moment.

python_benchmarks/conftest.py
```diff
@@ -33,3 +46,27 @@ def pytest_make_parametrize_id(val):
...
+                item.add_marker(skip_eager)
+
+    if not run_torchcompile:
+        skip_torchcompile = pytest.mark.skip(
+            reason="need --benchmark-torchcompile option to run"
+        )
+        for item in items:
+            # If the benchmark has compile=True parameter (torch.compile mode), skip it.
+            if "compile" in item.callspec.params and item.callspec.params["compile"]:
+                item.add_marker(skip_torchcompile)
```
jacobhinkle, line 72: So this will skip any tests with a `compile=True` or `compile=False` option. Is our convention then that any eager mode benchmarks should have a `compile=False` option? We should document those conventions in the docstring of this function or maybe in the docstring of this module.

python_benchmarks/test_softmax_fwd.py
```diff
@@ -63,8 +67,28 @@ def test_softmax_fwd_benchmark(
...
+@pytest.mark.parametrize("reduction_axis", [0, 1])
+def test_softmax_fwd_baseline_benchmark(
+    benchmark,
+    size: tuple,
+    dtype: torch.dtype,
+    reduction_axis: int,
+    compile: bool,
+):
+    clear_cuda_cache()
+    input = torch.randn(*size, device="cuda", dtype=dtype)
```
jacobhinkle, line 89: I think `torch.randn` can accept a tuple of ints for the shape parameter, so no asterisk is needed here.

python_benchmarks/test_softmax_bwd.py
```diff
@@ -82,3 +87,25 @@ def test_softmax_bwd_benchmark(
...
+def test_softmax_bwd_baseline_benchmark(
+    benchmark,
+    size: tuple,
+    dtype: torch.dtype,
+    reduction_axis: int,
+    compile: bool,
+):
+    clear_cuda_cache()
+    input = torch.randn(*size, device="cuda", dtype=dtype, requires_grad=True)
+    grads = torch.randn(*size, device="cuda", dtype=dtype)
```
jacobhinkle, line 105: See comment below about asterisks.

python_benchmarks/conftest.py
```diff
@@ -15,6 +15,19 @@ def pytest_addoption(parser):
         default=False,
         help="Disable benchmarking.",
     )
+    parser.addoption(
+        "--benchmark-eager",
+        action="store_true",
+        default=False,
```
jacobhinkle, line 21: Minor: I think `default=False` is not needed if `action="store_true"` is given.

jacobhinkle (approved):
LGTM. Just a few suggestions.

PR #1671, opened 2024-01-25
Remove unnecessary code.

wujingyue:
I don't know why it was added, but it looks unnecessary. An alias TV (whether it's a pointer arithmetic or an in-place update) should be connected to some input TV via Exprs.


wujingyue:
!build

naoyam (approved):
Makes sense. I don't think this is necessary either.

PR #1673, opened 2024-01-25
Insert squeeze before trivial reductions at concretization

### LINKED ISSUE #1667: Numbers mismatch between Torch and nvFuser.
https://github.com/NVIDIA/Fuser/blob/wjy/mismatch/repro.py is the reproducer. To run it,

```shell
git fetch origin wjy/mismatch
git checkout wjy/mismatch
python repro.py
```

~I suspect it's an alias analysis bug related to expanded broadcasts.~ No, it seems to have nothing to do with alias analysis. I managed to simplify the reproducer to reshape+sum.

jacobhinkle:
This change inspects expressions at concretization to check for reductions and Welford ops. When finding those, we look for unexpanded broadcasts in the axes being reduced. If we find any, we re-run the appropriate op with the concretized input, to ensure that the output is squeezed exactly as it would be were the fusion not dynamic.

Fixes #1667

### DIFF
--- a/csrc/device_lower/validation.cpp
+++ b/csrc/device_lower/validation.cpp
@@ -1291,22 +1291,24 @@ void validateResize(Fusion* fusion) {
 }
 
 void validateReductions(Fusion* fusion) {
-  for (auto expr : fusion->exprs()) {
-    if (auto rop = dynamic_cast<ReductionOp*>(expr)) {
-      const std::vector<IterDomain*> in_rfactor = TensorDomain::noReductions(
-          rop->in()->as<TensorView>()->getMaybeRFactorDomain());
-      const std::vector<IterDomain*>& out_root =
-          rop->out()->as<TensorView>()->getRootDomain();
-
-      NVF_ERROR(in_rfactor.size() == out_root.size());
-      for (auto i : c10::irange(in_rfactor.size())) {
-        if (out_root[i]->isReduction()) {
-          IterDomain* in_id = in_rfactor[i];
-          NVF_ERROR(
-              !in_id->isBroadcast() || in_id->hasExpandedExtent(),
-              "Reductions of unexpanded broadcast domains should be ",
-              "converted to squeeze before lowering.");
-        }
+  for (auto rop : ir_utils::getOpsOfType<ReductionOp>(fusion)) {
+    auto in = rop->in()->as<TensorView>();
+    auto out = rop->out()->as<TensorView>();
+    PairwiseRootDomainMap c2p_map(in, out);
+    c2p_map.mapBroadcast(true);
+    auto c2p = c2p_map.mapConsumerToProducer();
+    for (auto out_id : out->getRootDomain()) {
+      if (out_id->isReduction()) {
+        auto in_it = c2p.find(out_id);
+        NVF_ERROR(
+            in_it != c2p.end(),
+            "Could not find producer IterDomain mapped to ",
+            out_id->toString());
+        IterDomain* in_id = in_it->second;
+        NVF_ERROR(
+            !in_id->isBroadcast() || in_id->hasExpandedExtent(),
+            "Reductions of unexpanded broadcast domains should be ",
+            "converted to squeeze before lowering.");
       }
     }
   }
--- a/csrc/device_lower/validation.cpp
+++ b/csrc/device_lower/validation.cpp
@@ -1290,4 +1290,26 @@ void validateResize(Fusion* fusion) {
   }
 }
 
+void validateReductions(Fusion* fusion) {
+  for (auto expr : fusion->exprs()) {
+    if (auto rop = dynamic_cast<ReductionOp*>(expr)) {
+      const std::vector<IterDomain*> in_rfactor = TensorDomain::noReductions(
+          rop->in()->as<TensorView>()->getMaybeRFactorDomain());
+      const std::vector<IterDomain*>& out_root =
+          rop->out()->as<TensorView>()->getRootDomain();
+
+      NVF_ERROR(in_rfactor.size() == out_root.size());
+      for (auto i : c10::irange(in_rfactor.size())) {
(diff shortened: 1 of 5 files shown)


jacobhinkle:
!build --diff

test/test_dynamic_transform.cpp
```diff
@@ -1385,4 +1385,79 @@ TEST_F(NVFuserTest, ConcretizeConstantExtents) {
   testValidate(fusion, outputs, inputs, __LINE__, __FILE__);
 }
 
+// Test that dynamic reductions that should result in squeezes are handled
+// properly.
+// See https://github.com/NVIDIA/Fuser/issues/1667
+TEST_F(NVFuserTest, DynamicSqueezeTrivialReduction) {
```
wujingyue, line 1391: As a separate cleanup, consider s/NVFuserTest/DynamicTransformTest. This makes it easy to tell which file a test is in.

test/test_dynamic_transform.cpp
```diff
@@ -1385,4 +1385,79 @@ TEST_F(NVFuserTest, ConcretizeConstantExtents) {
...
+  std::unique_ptr<Fusion> fusion_ptr = std::make_unique<Fusion>();
```
wujingyue, line 1392:
```suggestion
  auto fusion_ptr = std::make_unique<Fusion>();
```

Nit: using auto here is fine and encouraged because the return type is clear from the context. Even for people outside nvFuser, `std::make_unique<T>` is a well-known standard function that returns `std::unique<T>`.

csrc/device_lower/validation.cpp
```diff
@@ -1290,4 +1290,26 @@ void validateResize(Fusion* fusion) {
...
+void validateReductions(Fusion* fusion) {
+  for (auto expr : fusion->exprs()) {
+    if (auto rop = dynamic_cast<ReductionOp*>(expr)) {
+      const std::vector<IterDomain*> in_rfactor = TensorDomain::noReductions(
+          rop->in()->as<TensorView>()->getMaybeRFactorDomain());
+      const std::vector<IterDomain*>& out_root =
+          rop->out()->as<TensorView>()->getRootDomain();
+
+      NVF_ERROR(in_rfactor.size() == out_root.size());
+      for (auto i : c10::irange(in_rfactor.size())) {
```
wujingyue, line 1302: Is it easier to use PairwiseRootDomainMap?

csrc/dynamic_transform.cpp
```diff
@@ -943,6 +946,90 @@ void DynamicTransformConcretizer::mutate(TensorDomain* td) {
...
+void DynamicTransformConcretizer::mutate(Expr* expr) {
+  const auto mutateAndReplaceInOutputs = [this](Val* old_val, Val* new_val) {
+    registerMutation(old_val, new_val);
+    if (old_val->isFusionOutput()) {
+      old_val->fusion()->replaceOutput(old_val, new_val);
+    }
+  };
+
+  if (ReductionOp* rop = dynamic_cast<ReductionOp*>(expr); rop) {
+    auto in = rop->in()->as<TensorView>();
```
wujingyue, line 960:
```suggestion
    auto* in = rop->in()->as<TensorView>();
```

csrc/dynamic_transform.cpp
```diff
@@ -943,6 +946,90 @@ void DynamicTransformConcretizer::mutate(TensorDomain* td) {
...
+    for (size_t i : c10::irange(in_rfactor.size())) {
+      if (!orig_out_root[i]->isReduction()) {
+        continue;
+      }
+      reduction_axes.push_back((int)i);
+      IterDomain* in_id = in_rfactor[i];
+      if (in_id->isBroadcast() && !in_id->hasExpandedExtent()) {
+        has_trivial_reduction = true;
+      }
+    }
```
wujingyue, line 977: This part is duplicated between the two branches. Make it a helper? E.g. 

```
bool hasTrivialReduction(TensorView* in, TensorView* out, std::vector<int>& reduction_axes);
```

wujingyue (approved):
LGTM. Given the bug blocks another PR and this PR is a strict improvement, I don't mind merging it before @naoyam takes a look.

csrc/dynamic_transform.cpp
```diff
@@ -943,6 +946,90 @@ void DynamicTransformConcretizer::mutate(TensorDomain* td) {
   registerConcretization(td, mutated_val);
 }
 
+// Maybe insert SqueezeOps on inputs of ReductionOp, to simplify trivial
+// reductions.
+void DynamicTransformConcretizer::mutate(Expr* expr) {
+  const auto mutateAndReplaceInOutputs = [this](Val* old_val, Val* new_val) {
+    registerMutation(old_val, new_val);
+    if (old_val->isFusionOutput()) {
+      old_val->fusion()->replaceOutput(old_val, new_val);
```
naoyam, line 955: I'm not sure if this is a real concern, but we are delaying the mutation after finishing the traversal is done since it's really error prone to modify a fusion while traversing it. Since `replaceOutput` is also a modification, shouldn't it also be delayed?

jacobhinkle: Great question. This pass actually already modifies the fusion by rewriting `Expr`s; that's what `mutate(Expr*)` does. However, there's a little bit of a subtlety here because if we change outputs, that can actually change `uses()` during traversal. I don't think that's actually important in this case, but in order to avoid tricky bugs, I think I'll add a stage at the end of concretization where we loop over outputs and replace them if they're registered for mutation.

csrc/device_lower/validation.cpp
jacobhinkle: Good advice. Thanks.

jacobhinkle:
!build

naoyam (approved):
LGTM

Issue #1656, opened 2024-01-23
haveDifferentSharding


naoyam:
https://github.com/NVIDIA/Fuser/blob/da7c4e97e3b90267a5465de856afc5e8d64fc56a/csrc/multidevice/utils.cpp#L47

- Does it need to look at all tensor inputs together?
- What happens if there's no mapped ID, i.e., line 70 never returns true? There can be still cases requiring resharding but looks like we would miss that. 

It seems to me that supporting all potential patterns of parallelization would effectively require the same analysis we do for TID/BID (sync_information.cpp). It would be fine to have some simplified analysis for now, but I think we should at least make it clear what are assumed and have proper assertions.

samnordmann:
> https://github.com/NVIDIA/Fuser/blob/da7c4e97e3b90267a5465de856afc5e8d64fc56a/csrc/multidevice/utils.cpp#L47
> 
> * Does it need to look at all tensor inputs together?

it doesn't strictly have to, but by doing so we can save to recompute the computeAtMap and the IDs at every tensor. I also thought it is strictly more general to accept a set of tensors instead of a single one, and probably makes the code a bit nicer.

Do you want me to change that?

> * What happens if there's no mapped ID, i.e., line 70 never returns true? There can be still cases requiring resharding but looks like we would miss that.

I didn't think about this case. Do you have an example in mind?

> It seems to me that supporting all potential patterns of parallelization would effectively require the same analysis we do for TID/BID (sync_information.cpp). It would be fine to have some simplified analysis for now, but I think we should at least make it clear what are assumed and have proper assertions.

I did the best I could with my understanding of the analysis, and this was already not so easy for me... :) This function works at least on our tests cases. But I'd be very happy if we can improve it! I would need some indications though.

Same for the assumptions, Can you help me figure out what are the implicit assumptions here? and what could make the function break?
One thing I can say is that, for now, we don't allow any split/merge and only allow the first axis to be parallelized accross devices. Assertions could be added to ensure that, like the ones inside `isSharded`

naoyam:
> > https://github.com/NVIDIA/Fuser/blob/da7c4e97e3b90267a5465de856afc5e8d64fc56a/csrc/multidevice/utils.cpp#L47
> > 
> > 
> > * Does it need to look at all tensor inputs together?
> 
[...]

PR #1674, opened 2024-01-25
Add asserts in haveDifferentSharding

samnordmann:
Adresses issue https://github.com/NVIDIA/Fuser/issues/1656

### DIFF
--- a/csrc/multidevice/utils.cpp
+++ b/csrc/multidevice/utils.cpp
@@ -20,7 +20,8 @@ bool isSharded(TensorView* tv) {
   for (IterDomain* id : TensorDomain::noReductions(tv->getLeafDomain())) {
     is_sharded.push_back(id->isDeviceDim());
   }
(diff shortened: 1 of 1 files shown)


naoyam (approved):
LGTM

csrc/multidevice/utils.cpp
```diff
@@ -60,6 +61,9 @@ std::unordered_set<TensorView*> haveDifferentSharding(
   }
 
   for (auto tv : tvs) {
+    // isSharded asserts that there are no split/merge and that only the outmost
```
naoyam, line 64: Actually, let's move the assumptions to the beginning of this function as it also affects how we can use `ComputeAtMap`.

naoyam: Also, shouldn't we check `ref` as well?

samnordmann:
!build

Issue #1669, opened 2024-01-25
Default scheduling for matmul in nvFuser


Priya2698:
This issue is intended to track the development of default scheduling support for `matmul` and design options.

By default, whenever nvFuser encounters a matmul configuration that is not supported, we want it to be evaluated through the expression evaluator using PyTorch's `aten::matmul`.

CC: @protonu  @wujingyue @kevinstephano

wujingyue:
@zasdfgbnm raised an important implementation challenge. There's apparently a plan to change nvFuser's interface so `fd.ops.matmul` will become `mul+sum` (and optionally broadcast and transpose) in the input FusionDefinition. The MatMulScheduler will be instead responsible to fuse that (or part of that) into an `MmaOp` with low-level details. This means `MmaOp` will only appear in MatMulScheduler::schedule and afterwards (e.g. device lowering). 

And thanks @zasdfgbnm for also explaining the motivations behind this plan of change:
1. MmaOp had been designed as a container for low level detail that user’s shouldn’t care. These details include Mma macros (like Ampere_16_8_16 ), layout (TT, TN, NT, NN), accumulator_stride  (a hack used to workaround a poor design in generated code, this hack no longer exist any more). It sounds weird to require users to provide these information. It makes more sense to have some simpler IR for the user to use, and let the matmul scheduler figure out these details and create the MmaOp.
2. The matmul scheduler is primarily for compute-bound matmuls. But depending on problem shape, matmul is not necessarily a compute bound kernel. In such senario, it would good to just let the matmul scheduler reject the fusion so that nvFuser can schedule it like a reduction, which can potentially fuse more ops.
3. At that time, there were no mechanism for nvFuser to invoke eager mode to evaluate tensor expressions. In that context, when users define a matmul, it is clearly a better option to schedule it like a reduction, compared with just fail and raise an error.
4. This design is kind of consistent with how we handle normalization: we do not have a NormalizationOp . Instead, normalization is defined as a reduction-broadcast-pointwise. Following the same scheme, it is natural to define matmul as mul-sum.

@protonu made a Monday meeting for us to talk about this. We'll clarify problems and hopefully converge to a solution that addresses the above concerns and this feature request.

wujingyue:
Here are the meeting notes. cc attendees: @jacobhinkle , @kevinstephano , @Priya2698 , @protonu , @zasdfgbnm 

TL;DR: the recommendation so far is to pursue option 4 with option 5 as an intermediate step. 

[...]

Issue #1628, opened 2024-01-15
Extra S->S copy introduced by tmaLoad->broadcast->mma. 


zasdfgbnm:
When we define matmul, we have inputs of shape `[M, K]` and `[K, N]`, and we broadcast it to `[M, K, 1]` and `[1, K, N]`, then we do a `mul` and `sum` to obtain the matmul result. From `[M, K]` to `[M, K, 1]`, there must be a `BroadcastOp`. Unfortunately, this does not work with Hopper, because for fast Hopper matmul kernel, we must use TMA for global -> shared memory loading and use MMA instructions to directly pull data from shared mem. If we are unable to fuse the `global -> shared` with broadcasting op, we will end up having a kernel like:
```C++
T1_s = tmaLoad(T0_g);
T2_s = broadcast(T1_s);
... = mma(T2_s, ...);
```
That is, we need to double the smem usage to store the same data twice (`T1 ` and `T2`), one for TMA consumer, another for MMA producer, and do a S->S copy between these two memory.

I am still learning but according to my current understanding, the smem data written by TMA is supposed to be used by MMA directly, so this extra S->S copy introduced by the `BroadcastOp` does not make sense.

What I am proposing is to go through the same process as I did in the past for `TransposeOp`: In the past, transposes must be done in a `TransposeOp`, but after my change, any op can have a fused transpose by specifying the order in its rFactor domain. Most commonly, the op having a fused transpose is a `LoadStoreOp`, but it is not required to be so.

I want broadcasting (and to be consistent, squeeze as well) to follow the same contract: There will no longer be a `BroadcastOp` or `SqueezeOp`, instead, broadcasting `IterDomain`s is allowed to be created or annihilated at any `TensorView`. (The term "created" and "annihilated" is borrowed from particle physics, because I am feeling that broadcasting `IterDomain` is like a Boson, it can be created from nowhere, and it can just disappear as it wants). For example, if we want to do a broadcast, we can define:
```C++
T1 = LoadStoreOp(T0)

T0: root domain [I1, I2]
T1: root domain [I3, I4]
    rFactor domain [I3, b5, I4]
```
we can also do fused squeeze-broadcast
```C++
T1 = LoadStoreOp(T0)

T0: root domain [I1, b3, I2]
T1: root domain [I4, b5, I6]
    rFactor domain [I4, b7, I6]
// squeeze b5, broadcast b7
```

[...]

PR #1675, opened 2024-01-26
`MmaOp::evaluate` method

Priya2698:
Implements the `evaluate` method for `MmaOp`. 
The `evaluate` method assumes:

1. `MmaOp` is preceded by `broadcast`.
2. The inputs (non-broadcasted) are 2D.
3. The inputs to the `matmul` API and the broadcasts will be `[M, K] x [K, N]` which is the same as PyTorch.

This is intended to be used in default matmul scheduling .
(Issue #1669)

### DIFF
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -2066,6 +2066,7 @@ void MmaOp::setMacro(MmaMacro macro) {
 std::vector<PolymorphicValue> MmaOp::evaluate(
     const ExpressionEvaluator& ee,
     const std::vector<PolymorphicValue>& inputs) const {
+
   auto& a = inputs.at(0).as<at::Tensor>();
   auto& b = inputs.at(1).as<at::Tensor>();
   NVF_CHECK(
@@ -2075,7 +2076,9 @@ std::vector<PolymorphicValue> MmaOp::evaluate(
       "Must have either zero or one batch dimensions");
 
   // Squeeze the inputs to remove the broadcasted dimensions.
-  // TODO: This is assuming the broadcast dimensions but may be variable?
+  // TODO: This is assuming the broadcast dimensions to be the last dim for the first operand
+  //      and the first dim for the second operand. If this is variable, using the inputs of the
+  //      broadcast instead of the immediate inputs may be better.
   auto in_a = a.squeeze(-1);
   auto in_b = b.squeeze(0);
 
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -2075,6 +2075,7 @@ std::vector<PolymorphicValue> MmaOp::evaluate(
       "Must have either zero or one batch dimensions");
 
   // Squeeze the inputs to remove the broadcasted dimensions.
+  // TODO: This is assuming the broadcast dimensions but may be variable?
   auto in_a = a.squeeze(-1);
   auto in_b = b.squeeze(0);
 
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -2081,7 +2081,7 @@ std::vector<PolymorphicValue> MmaOp::evaluate(
   // fusedMultiplySum assumes [B, M, K] [B, N, K]
   // Transpose them to aten::matmul format.
   if (in_a.dim() == 3) { // bmm
-    return {in_a.transpose(0, 1).matmul(in_b.transpose(0, 1).transpose(1, 2))};
+    return {in_a.matmul(in_b.transpose(1, 2))};
   } else {
     return {in_a.matmul(in_b.t())};
   }
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -2066,26 +2066,24 @@ void MmaOp::setMacro(MmaMacro macro) {
 std::vector<PolymorphicValue> MmaOp::evaluate(
     const ExpressionEvaluator& ee,
     const std::vector<PolymorphicValue>& inputs) const {
-  // MmaOp is preceded by broadcast. For aten::matmul use the non-broadcasted
-  // inputs.
-  const auto& a =
-      ee.evaluate(input(0)->definition()->input(0)).as<at::Tensor>();
-  const auto& b =
-      ee.evaluate(input(1)->definition()->input(0)).as<at::Tensor>();
-
+  auto& a = inputs.at(0).as<at::Tensor>();
+  auto& b = inputs.at(1).as<at::Tensor>();
   NVF_CHECK(
       a.dim() == b.dim(), "Either both or none of A and B should be batch");
   NVF_CHECK(
       a.dim() == 2 || a.dim() == 3,
       "Must have either zero or one batch dimensions");
 
-  // The inputs to broadcast will be in the TN format.
-  // (M, B, K) x (N, B, K) or (M, K) x (N, K).
+  // Squeeze the inputs to remove the broadcasted dimensions.
+  auto in_a = a.squeeze(-1);
+  auto in_b = b.squeeze(0);
+
+  // fusedMultiplySum assumes [B, M, K] [B, N, K]
   // Transpose them to aten::matmul format.
-  if (a.dim() == 3) { // bmm
-    return {a.transpose(0, 1).matmul(b.transpose(0, 1).transpose(1, 2))};
+  if (in_a.dim() == 3) { // bmm
+    return {in_a.transpose(0, 1).matmul(in_b.transpose(0, 1).transpose(1, 2))};
   } else {
-    return {a.matmul(b.t())};
+    return {in_a.matmul(in_b.t())};
   }
 }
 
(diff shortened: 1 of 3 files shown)


Priya2698:
@protonu @zasdfgbnm from the existing mma tests, it appeared that the inputs to the `broadcast` preceding `MmaOp` will be in the `TN` format. Is this assumption correct?

wujingyue:
> @protonu @zasdfgbnm from the existing mma tests, it appeared that the inputs to the `broadcast` preceding `MmaOp` will be in the `TN` format. Is this assumption correct?

Can you clarify what you mean by `TN` format? Are you talking about physical layouts or logic shapes?

Priya2698:
> > @protonu @zasdfgbnm from the existing mma tests, it appeared that the inputs to the `broadcast` preceding `MmaOp` will be in the `TN` format. Is this assumption correct?
> 
> Can you clarify what you mean by `TN` format? Are you talking about physical layouts or logic shapes?

Logic shapes:` [M, B, K] @ [N, B, K] -> [B, M, N]`

test/test_evaluator.cpp
```diff
@@ -673,4 +673,59 @@ TEST_F(ExprEvalTest, SumDiv) {
...
+  FusionGuard fg(&fusion);
+
+  std::vector<MmaLayout> all_mma_layouts{
+      MmaLayout::NT, MmaLayout::NN, MmaLayout::TT, MmaLayout::TN};
+
+  for (auto layout : all_mma_layouts) {
+    bool transpose_a = (layout == MmaLayout::NT || layout == MmaLayout::NN);
+    bool transpose_b = (layout == MmaLayout::TT || layout == MmaLayout::NT);
+
+    int64_t M = 2, K = 3, N = 4;
```
wujingyue, line 687: nvFuser's naming convention seems to be lower case for variable names.

csrc/ir/nodes.cpp
```diff
@@ -2063,6 +2063,32 @@ void MmaOp::setMacro(MmaMacro macro) {
...
+  // inputs.
+  const auto& a =
+      ee.evaluate(input(0)->definition()->input(0)).as<at::Tensor>();
+  const auto& b =
+      ee.evaluate(input(1)->definition()->input(0)).as<at::Tensor>();
+
+  NVF_CHECK(
+      a.dim() == b.dim(), "Either both or none of A and B should be batch");
+  NVF_CHECK(
+      a.dim() == 2 || a.dim() == 3,
```
wujingyue, line 2079: Interesting -- I didn't know our MmaOp takes 2D (without broadcasting). IIUC, your test doesn't test this case. So either remove this feature or test it.

Priya2698: I don't have a test case for 3D yet, I'll add one for that.

csrc/ir/nodes.cpp
```diff
@@ -2063,6 +2063,34 @@ void MmaOp::setMacro(MmaMacro macro) {
...
+      "Must have either zero or one batch dimensions");
+
+  // Squeeze the inputs to remove the broadcasted dimensions.
+  // TODO: This is assuming the broadcast dimensions to be the last dim for the first operand
+  //      and the first dim for the second operand. If this is variable, using the inputs of the
+  //      broadcast instead of the immediate inputs may be better.
+  auto in_a = a.squeeze(-1);
```
Priya2698, line 2082: This is based on the broadcast dimensions in the `matmul` API on @protonu's branch. However, if it doesn't always hold, we can bypass and use the inputs to the broadcast directly.

csrc/ir/nodes.cpp
```diff
@@ -2063,6 +2063,32 @@ void MmaOp::setMacro(MmaMacro macro) {
...
 
+std::vector<PolymorphicValue> MmaOp::evaluate(
+    const ExpressionEvaluator& ee,
+    const std::vector<PolymorphicValue>& inputs) const {
+  // MmaOp is preceded by broadcast. For aten::matmul use the non-broadcasted
+  // inputs.
+  const auto& a =
+      ee.evaluate(input(0)->definition()->input(0)).as<at::Tensor>();
+  const auto& b =
+      ee.evaluate(input(1)->definition()->input(0)).as<at::Tensor>();
```
jacobhinkle: You should check that these definitions are non-null and that they are `BroadcastOp` if that is what is assumed. There is discussion now whether it will always be the case that MmaOp will always be preceded by `BroadcastOp` as the interface is evolving.

jacobhinkle: It seems more robust to me to instead try and squeeze and matmul without pattern matching the definitions of `MmaOp` inputs. For max generality this could be done with https://github.com/pytorch/pytorch/blob/8ff55c7e689d4c6f89f806112e04ff56dbfb6a3f/aten/src/ATen/native/Linear.cpp#L137 but I think that is not exposed. `einsum` is though, if you wanted to go that route.

csrc/ir/nodes.cpp
```diff
@@ -2063,6 +2063,32 @@ void MmaOp::setMacro(MmaMacro macro) {
...
+
+  NVF_CHECK(
+      a.dim() == b.dim(), "Either both or none of A and B should be batch");
+  NVF_CHECK(
+      a.dim() == 2 || a.dim() == 3,
+      "Must have either zero or one batch dimensions");
+
+  // The inputs to broadcast will be in the TN format.
```
jacobhinkle: TN format refers to memory layout. Instead, I would just say "in the following format".

csrc/ir/nodes.cpp
```diff
@@ -2063,6 +2063,34 @@ void MmaOp::setMacro(MmaMacro macro) {
   attribute<MmaMacro>(ATTR_POS_MACRO) = macro;
 }
 
+std::vector<PolymorphicValue> MmaOp::evaluate(
+    const ExpressionEvaluator& ee,
+    const std::vector<PolymorphicValue>& inputs) const {
+
+  auto& a = inputs.at(0).as<at::Tensor>();
```
wujingyue, line 2070: `const auto&` seems good enough. (Same for `b`)

csrc/ir/nodes.cpp
wujingyue: I'm confused. The test you added is for 3D not 2D. The `broadcast`s there make the MmaOp's inputs 3D.

csrc/ir/nodes.cpp
wujingyue: > This is assuming the broadcast dimensions to be the last dim for the first operand and the first dim for the second operand.

I'd explicitly `NVF_CHECK` these assumptions by examining `inputs()[0]`'s rfactor domain and `inputs()[1]`'s. 

> If this is variable, using the inputs of the broadcast instead of the immediate inputs may be better.

This would be my last resort. As Jacob said, it's more robust to avoid assuming inputs are broadcasts. Also, MmaOp, as other ops (maybe Cat is an exception?), should be standalone and have a clear semantics.

csrc/ir/nodes.cpp
Priya2698: The non-broadcasted inputs can be 2D or 3D. The checks for the non-broadcasted inputs and I need to update the checks if I am using the broadcasted inputs or have them after removing those dimensions.

csrc/ir/nodes.cpp
```diff
@@ -2063,6 +2063,45 @@ void MmaOp::setMacro(MmaMacro macro) {
...
+    tv_a->nDims() == tv_b->nDims(), 
+    "Either both or none of A and B should be batch");
+  // Verify that the broadcasted size is 3. 
+  NVF_CHECK(
+    tv_a->nDims() == 3,
+    "MmaOp::evaluate is not implemented for size: ", tv_a->nDims());
+
+  // Assumptions:
+  //    Currently, the evaluate method assumes that the MmaOp is preceded by a
+  //    broadcast. The inputs to MmaOp are broadcasted as the last dim for the
```
wujingyue, line 2082: Do we care about the MmaOp being preceded by a BroadcastOp? Is the `isBroadcast` check sufficient?

Priya2698: I added the first check to verify that these broadcasts resulted from the `broadcastOp` preceding `fusedMatmulSum`. It is possible for the inputs to have been broadcasted at some other point and not for matmul.

Priya2698: It is mainly to guard against this change so we are clear on what we expect in the `evaluate` method.

wujingyue: > In that case, we may be removing dimensions which just happen to be broadcasted but by some other op.

You are probably referring to #1628, for which one of the solutions (probably the POR) is to have `MatmulScheduler` reject the preceding `BroadcastOp`s and therefore insert `Set` for tmaLoad between `BroadcastOp` and `MmaOp`. In that case, your code as is will work just fine **except** this `BroadcastOp` check :) That's why I think removing this check is better. 

Anyhow, leaving the check as is should be fine for now. I was mainly verifying the motivation behind it.

wujingyue (approved):

PR #1679, opened 2024-01-26
Convert reduction of expanded dims to squeeze

### LINKED ISSUE #1678: Reduction on expanded tensor produces result with incorrect shape
Here's an example of two almost identical fusions with the only difference being the contiguity of T0, the output of the fusion with `contiguity=[None, None]` produces the result with incorrect shape:
```py
import torch

# The only difference between two fusions is "contiguity" of T0.
    T0 = fd.define_tensor(shape=[-1, -1], contiguity=[None, None], dtype=DataType.Float, is_cpu=False, stride_order=[1, 0])
    T1 = fd.ops.sum(T0, axes=[0], keepdim=False, dtype=DataType.Null)
    fd.add_output(T1)

    T0 = fd.define_tensor(shape=[-1, -1], contiguity=[True, True], dtype=DataType.Float, is_cpu=False, stride_order=[1, 0])
    T1 = fd.ops.sum(T0, axes=[0], keepdim=False, dtype=DataType.Null)
    fd.add_output(T1)

t0_contiguous = torch.ones((2, 3), dtype=torch.float32, device='cuda')
t0_expanded = torch.tensor(1.0, device='cuda').as_strided((2, 3), (0, 0))

with FusionDefinition() as fd0:
    nvfuser_fusion_id0(fd0)

with FusionDefinition() as fd1:
    nvfuser_fusion_id1(fd1)

out0 = fd0.execute([t0_expanded])
out1 = fd1.execute([t0_contiguous])

# AssertionError: The values for attribute 'shape' do not match: torch.Size([1]) != torch.Size([3]).
torch.testing.assert_allclose(out0[0], out1[0])
```

jacobhinkle:
See comment in arith.cpp for details.

One controversial change here is to allow squeezing expanded dimensions, both in our IR's `SqueezeOp` and in the user-facing functions `squeeze`. This results in actually removing those dimensions. This behavior diverges from PyTorch, whose `squeeze` command will ignore requested squeezes if the size is not 1 regardless of whether that dimension is expanded. I'm happy to discuss this change and potentially take another course, but I think we do need to be able to remove expanded axes (see https://github.com/NVIDIA/Fuser/issues/1174#issuecomment-1851998818 for another case where I encountered this limitation).

Fixes #1678

### DIFF
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -1309,8 +1309,6 @@ SqueezeOp::SqueezeOp(
           id->isBroadcast() || id->isSymbolic(),
           "Squeeze dimension should be either Symbolic or Broadcast. Found ",
           id->getIterType());
-      NVF_ERROR(
-          !id->hasExpandedExtent(), "Can not squeeze expanded dimension(s).");
       if (id->isBroadcast()) {
         // Check concrete broadcast extent here. For Symbolic inputs, this check
         // will be deferred to concretization. See dynamic_transform.cpp
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -2101,7 +2101,7 @@ std::vector<PolymorphicValue> ExpandOp::evaluate(
   for (auto i : c10::irange(1, inputs.size())) {
     expanded_size.push_back((int64_t)inputs.at(i));
   }
-  return {at::expand_copy(in, expanded_size)};
+  return {in.expand(expanded_size)};
 }
 
 NVFUSER_DEFINE_CLONE_AND_CREATE(ExpandOp)
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -1354,12 +1354,20 @@ std::vector<PolymorphicValue> SqueezeOp::evaluate(
   NVF_ERROR(
       (int64_t)is_squeeze_dims.size() == in.dim(),
       "The dimensions of input tensor and does not match with is_squeeze_dims");
+  at::Tensor out = in;
   for (int64_t i : c10::irange((int64_t)is_squeeze_dims.size())) {
-    if (!is_squeeze_dims[i]) {
+    if (is_squeeze_dims[i]) {
+      if (in.stride(i) == 0) {
+        // If the input dimension is expanded in this dimension, undo the expand
+        // by slicing. This ensures that any broadcast dimensions will be
+        // unexpanded when we do the final call to view()
+        out = out.slice(i, 0, 1);
+      }
+    } else {
       out_shape.push_back(in.sizes()[i]);
     }
   }
-  return {in.view(out_shape)};
+  return {out.view(out_shape)};
 }
 
 void SqueezeOp::checkConcretization(Val* old_val, Val* new_val) const {
--- a/csrc/ops/alias.cpp
+++ b/csrc/ops/alias.cpp
@@ -245,8 +245,6 @@ TensorView* squeeze(TensorView* x, const std::vector<int64_t>& dims) {
           continue;
         }
         NVF_CHECK(
-            !id->hasExpandedExtent(), "Can not squeeze expanded dimension(s).");
-        NVF_CHECK(
             id->extent()->isConstScalar() && id->extent()->evaluate() == 1,
             "Can not squeeze dimension(s) with size != 1.");
       }
@@ -291,8 +289,6 @@ TensorView* squeeze(TensorView* x, const std::vector<bool>& to_squeeze) {
         NVF_CHECK(
             id->isBroadcast(),
             "Can not squeeze non-broadcasting dimension(s).");
-        NVF_CHECK(
-            !id->hasExpandedExtent(), "Can not squeeze expanded dimension(s).");
         NVF_CHECK(
             id->extent()->isConstScalar() && id->extent()->evaluate() == 1,
             "Can not squeeze dimension(s) with size != 1.");
--- a/csrc/ops/alias.cpp
+++ b/csrc/ops/alias.cpp
@@ -204,7 +204,10 @@ TensorView* flatten(TensorView* x, int64_t start_dim, int64_t end_dim) {
   return out;
 }
 
-TensorView* squeeze(TensorView* x, const std::vector<int64_t>& dims) {
+TensorView* squeeze(
+    TensorView* x,
+    const std::vector<int64_t>& dims,
+    bool squeeze_expanded) {
   NVF_ERROR(x != nullptr, "Input is invalid.");
   auto x_dom = x->domain()->noReductions();
   const auto ndims = static_cast<int>(x_dom.size());
@@ -232,44 +235,13 @@ TensorView* squeeze(TensorView* x, const std::vector<int64_t>& dims) {
     to_squeeze[dim] = true;
   }
 
-  std::vector<IterDomain*> out_domain;
-  for (const auto idx : c10::irange(ndims)) {
-    auto id = x_dom[idx];
-    if (to_squeeze[idx]) {
-      if (!id->isSymbolic()) {
-        // If a squeeze is attempted on a non-broadcast dimension
-        // just don't do it!  This conforms with Pytorch.
-        if (!id->isBroadcast()) {
-          to_squeeze[idx] = false;
-          out_domain.push_back(id->cloneWithoutRFactor());
-          continue;
-        }
-        NVF_CHECK(
-            id->extent()->isConstScalar() && id->extent()->evaluate() == 1,
-            "Can not squeeze dimension(s) with size != 1.");
-      }
-    } else {
-      out_domain.push_back(id->cloneWithoutRFactor());
-    }
-  }
-
-  auto out = IrBuilder::create<TensorView>(
-      IrBuilder::create<TensorDomain>(
-          out_domain, TensorDomain::getContiguityFilledWith(out_domain, true)),
-      *x->getDataType());
-
-  std::vector<bool> all_false(to_squeeze.size(), false);
-  // If a squeeze does not perform a squeeze, create a no-op
-  if (to_squeeze == all_false) {
-    IrBuilder::create<LoadStoreOp>(LoadStoreOpType::Set, out, x);
-  } else {
-    IrBuilder::create<SqueezeOp>(x->container(), out, x, to_squeeze);
-  }
-
-  return out;
+  return squeeze(x, to_squeeze, squeeze_expanded);
 }
 
-TensorView* squeeze(TensorView* x, const std::vector<bool>& to_squeeze) {
+TensorView* squeeze(
+    TensorView* x,
+    const std::vector<bool>& to_squeeze,
+    bool squeeze_expanded) {
   NVF_ERROR(x != nullptr, "Input is invalid.");
   auto x_dom = x->domain()->noReductions();
   const auto ndims = static_cast<int>(x_dom.size());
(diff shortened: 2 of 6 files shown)


jacobhinkle:
!build --diff

csrc/ops/arith.cpp
```diff
@@ -1291,32 +1291,74 @@ TensorView* reductionOp(
...
+  std::vector<bool> is_broadcast_reduction(ndims, false);
+  bool expand_reductions_are_trivial = reduction_op_type != BinaryOpType::Add &&
+      reduction_op_type != BinaryOpType::Mul &&
+      reduction_op_type != BinaryOpType::BitwiseXor;
   int offset = 0;
   for (unsigned int axis : uint_axes) {
     auto id = tv_root[axis];
-    is_trivial_reduction[axis] = id->isBroadcast() &&
-        !id->hasExpandedExtent() && id->extent()->isConstInt() &&
-        id->extent()->evaluate().as<int64_t>() == 1;
```
jacobhinkle, line 1301: We were checking that extent is 1. I don't think we should need to do that, since we should not have any way to create a broadcast dim with unexpanded extent that's not 1.

jacobhinkle:
There are currently two test failures. The first is
```
unknown file: Failure                                                                                            
C++ exception with description "shape '[8]' is invalid for input of size 96" thrown in the test body.            
To reproduce: NVFUSER_TEST_RANDOM_SEED=1706286828 NVFUSER_TEST_ATEN_RANDOM_SEED=0 nvfuser_tests --gtest_filter='NVFuserTest.FusionExpandReduce_CUDA'                                                                              
[  FAILED  ] NVFuserTest.FusionExpandReduce_CUDA (360 ms)  
```
The fusion is defined like this:
```c++
  auto tv0 = makeConcreteTensor({1, 8});
  fusion->addInput(tv0);

  auto tv1 =
      expand(tv0, {IrBuilder::create<Val>(12L), IrBuilder::create<Val>(8L)});

  auto tv2 = sum(tv1, {0});
  fusion->addOutput(tv2);
```
resulting in this IR:
```
Inputs:
  T0_g[ bS0{1}, iS1{8} ], float
Outputs:
  T3_g[ iS5{8} ], float

%kernel_math {
T1_l[ bS2{1 ex 12}, iS3{8} ] = expand( T0_g[ bS0{1}, iS1{8} ], {12, 8} )
T2_l[ iS4{8} ]
   = squeeze( T1_l[ bS2{1 ex 12}, iS3{8} ] )
T3_g[ iS5{8} ]
   = T2_l[ iS4{8} ]
   * float(12);
}
```
It is properly computed, but the `evaluate` method of `SqueezeOp` translates to `at::squeeze` which does not squeeze the expanded axis. If we go this route we can check for expanded axes and first slice them before squeezing.

jacobhinkle:
The other test failure is
```
unknown file: Failure
C++ exception with description "Splitting an axis of non-Serial parallel type is not supported at this time. Parallelization strategy must be set after calling split.. Tensor: T3_g[ iblockIdx.y8{gridDim.y}, iS9{( ceilDiv(( ceilDiv(4, blockDim.y) ), gridDim.y) )}, ithreadIdx.y7{blockDim.y} ]
Exception raised from split at /opt/pytorch/nvfuser/csrc/tensor_view.cpp:671 (most recent call first):
frame #0: nvfuser::nvfCheckFail(char const*, char const*, unsigned int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) + 0x7f (0x7f50df36577e in /opt/pytorch/nvfuser/build/libnvfuser_codegen.so)
frame #1: nvfuser::TensorView::split(int, nvfuser::Val*, bool, bool) + 0x3fe (0x7f50dfa84c22 in /opt/pytorch/nvfuser/build/libnvfuser_codegen.so)
frame #2: <unknown function> + 0x79024a (0x55e21c44324a in build/nvfuser_tests)
frame #3: <unknown function> + 0xc982cd (0x55e21c94b2cd in build/nvfuser_tests)
frame #4: <unknown function> + 0xc9157f (0x55e21c94457f in build/nvfuser_tests)
frame #5: <unknown function> + 0xc66be2 (0x55e21c919be2 in build/nvfuser_tests)
frame #6: <unknown function> + 0xc67670 (0x55e21c91a670 in build/nvfuser_tests)
frame #7: <unknown function> + 0xc67f77 (0x55e21c91af77 in build/nvfuser_tests)
frame #8: <unknown function> + 0xc77e89 (0x55e21c92ae89 in build/nvfuser_tests)
frame #9: <unknown function> + 0xc991ac (0x55e21c94c1ac in build/nvfuser_tests)
frame #10: <unknown function> + 0xc92543 (0x55e21c945543 in build/nvfuser_tests)
frame #11: <unknown function> + 0xc765f1 (0x55e21c9295f1 in build/nvfuser_tests)
frame #12: <unknown function> + 0xc4a93c (0x55e21c8fd93c in build/nvfuser_tests)
frame #13: <unknown function> + 0xc4a8b5 (0x55e21c8fd8b5 in build/nvfuser_tests)
frame #14: <unknown function> + 0x29d90 (0x7f50a7f05d90 in /usr/lib/x86_64-linux-gnu/libc.so.6)
frame #15: __libc_start_main + 0x80 (0x7f50a7f05e40 in /usr/lib/x86_64-linux-gnu/libc.so.6)
frame #16: <unknown function> + 0x3ebf45 (0x55e21c09ef45 in build/nvfuser_tests)
" thrown in the test body.
To reproduce: NVFUSER_TEST_RANDOM_SEED=1706286737 NVFUSER_TEST_ATEN_RANDOM_SEED=0 nvfuser_tests --gtest_filter='NVFuserTest.FusionExpandReduce2_CUDA'
```
That test uses a similar fusion but is doing some manual scheduling, which assumes we have not translated the reduction to a squeeze already, so it breaks before compilation.

csrc/ir/nodes.cpp
```diff
@@ -2103,7 +2109,7 @@ std::vector<PolymorphicValue> ExpandOp::evaluate(
   for (auto i : c10::irange(1, inputs.size())) {
     expanded_size.push_back((int64_t)inputs.at(i));
   }
-  return {at::expand_copy(in, expanded_size)};
+  return {in.expand(expanded_size)};
```
jacobhinkle, line 2112: I believe this change is warranted regardless of whether we go forward with squeezing expanded ops. It was necessary in this case to fix broken test `ExpandReduce_CUDA`. cc @Priya2698

test/test_gpu3.cpp
```diff
@@ -3496,58 +3496,6 @@ TEST_F(NVFuserTest, FusionExpandReduce_CUDA) {
...
-  tv2->axis(1)->parallelize(ParallelType::TIDy);
-  tv2->split(0, NamedScalar::getParallelDim(ParallelType::BIDy), false);
-  tv2->axis(0)->parallelize(ParallelType::BIDy);
-  tv2->split(-1, NamedScalar::getParallelDim(ParallelType::TIDx));
-  tv2->axis(-1)->parallelize(ParallelType::TIDx);
-  tv2->axis(-2)->parallelize(ParallelType::BIDx);
-  // [rBIDy, rO, rTIDy, iBIDx, iTIDx]
-  tv2->reorder({{-2, 0}, {-1, 1}, {2, 2}});
-  // [iBIDx, iTIDx, rTIDy, rBIDy, rO]
-  auto tv3 = tv2->rFactor({-1});
```
jacobhinkle, line 3524: This test was meant to exercise reduction of expanded domains, which can always be translated to pointwise ops. I removed it because it seems no longer relevant to me.

jacobhinkle:
Tests are passing now. One was fixed by updating the `evaluate` methods in `SqueezeOp` and `ExpandOp` and the other test was removed.

naoyam:
> This behavior diverges from PyTorch

Does it mean end results would not match if expand and squeeze are used?

> I think we do need to be able to remove expanded axes (see https://github.com/NVIDIA/Fuser/pull/1174#issuecomment-1851998818)

Could you please explain the reasoning again? Looked at the linked comment, still not really clear to me.

jacobhinkle:
> Does it mean end results would not match if expand and squeeze are used?

Possibly, which is the main downside I see to the current approach in this PR. To see the torch behavior, consider this example:
```
[nav] In [1]: import torch
[nav] In [2]: x = torch.randn((2, 1)).expand([2, 3])
[nav] In [3]: x.squeeze(1)
Out[3]: 
tensor([[2.0809, 2.0809, 2.0809],
        [2.1042, 2.1042, 2.1042]])
```
In this case, torch silently refuses to squeeze the second axis because it has extent > 1. In order to remove that axis you need to slice it instead.
```
[ins] In [4]: x[:, 0]
Out[4]: tensor([2.0809, 2.1042])
```
If we stuck with that behavior in our IR, we would probably want special handling for `SliceOp` when operating on `Broadcast` input axes. Instead, this PR is currently just allowing the first behavior to squeeze an expanded axis. I suppose another way to go would be to introduce an `UnexpandOp` which could undo expansions, then that could be followed by `SqueezeOp` when we want to remove an expanded broadcast.

A middle ground might be to keep this behavior in `squeeze`, but  hide it behind a default-disabled option which would still allow us to use it when needed like `squeeze(tv0, {1, 2}, /*squeeze_expanded=*/true)`.

> > I think we do need to be able to remove expanded axes (see [#1174 (comment)](https://github.com/NVIDIA/Fuser/pull/1174#issuecomment-1851998818))
> 
> Could you please explain the reasoning again? Looked at the linked comment, still not really clear to me.

In #1174, we are encountering reshapes that need to split an expanded dimension. It's unnecessary to try and track a split expand back to the original broadcast. We could just forget about the original expanded broadcast and introduce two new ones. This seems simplest to me, and matches our current treatment of unexpanded broadcasts in reshape; we currently squeeze all pre-existing broadcasts and use a `BroadcastOp` after the `ViewOp` to introduce new ones. We could do the same with expand if we were able to remove the expanded broadcasts.

csrc/ops/arith.cpp
```diff
@@ -1291,32 +1291,74 @@ TensorView* reductionOp(
...
+  //   Eq -> squeeze
+  //   Gcd -> squeeze
+  // Other op-types are non-commutative, so we ignore them here as they should
+  // not be used in reductions. We can see that the only two common ops that
+  // require special consideration are Add and Mul. Currently Xor is not
+  // supported for expanded reduction. We treat all others as trivial (i.e.
+  // squeeze).
   std::vector<int> reduction_axes;
-  std::vector<bool> is_trivial_reduction(ndims, false);
+  std::vector<bool> is_broadcast_reduction(ndims, false);
```
naoyam, line 1312: nit: 
```suggestion
  std::vector<bool> is_squeeze(ndims, false);
```

csrc/ops/alias.cpp
```diff
@@ -244,8 +244,6 @@ TensorView* squeeze(TensorView* x, const std::vector<int64_t>& dims) {
           out_domain.push_back(id->cloneWithoutRFactor());
           continue;
         }
-        NVF_CHECK(
```
naoyam, line 247: Please leave a comment about the deviation of behavior

naoyam (commented):
Can you please add the repro as a test?

jacobhinkle:
> Can you please add the repro as a test?

I can. I had left it out since it is essentially the same as this existing test https://github.com/NVIDIA/Fuser/blob/1603f39bab8c1bbe12e38f2b5de53dec3b7cc373/test/test_gpu3.cpp#L3477

csrc/ops/alias.cpp
jacobhinkle: Noted in the docstrings for both signatures of `squeeze` in `alias.h`.

naoyam (approved):
LGTM

PR #1685, opened 2024-01-26
[WIP] Two kernel split-k

jacobhinkle:
WIP status. Mma segment is rejected bc it has leaf axis that doesn't map to input, so we can't find layout. Might need to either explicitly handle that case in `mma_utils::getMmaLayout` or perform the split in the definition i.e. commit it from leaf to rfactor and propagate backward. The second involves commiting a non-divisible split, so I prefer the alternative hack of special-casing in `mma_utils::getMmaLayout`.


jacobhinkle:
I think at this point a more realistic future path for two-kernel split-K is more of a megakernel persistent warp specialized approach.

PR #1689, opened 2024-01-28
print bandwidth when perf_debug_verbose is true

liqiangxl:
simple change, will print bandwidth when `perf_debug_verbose` is true.


jacobhinkle (approved):
Thanks for the fix. I thought it was already printing when `perf_debug_verbose` is given, but you're right that it's not. I have just been using `NVFUSER_PROF=print` instead lately so I didn't notice.

PR #1692, opened 2024-01-29
Accomodate Reduction IterDomains when concretizing reshape extents

### LINKED ISSUE #1691: `reshape` fails with `RuntimeError: old_rfactor.size() == new_rfactor.size() INTERNAL ASSERT FAILED at "/workspace/Fuser/csrc/dynamic_transform.cpp":652`
The following code
```python                                                                                                              
import torch                                                                                                           
                                                                                                                       
def nvfuser_fusion_id9(fd : FusionDefinition) -> None :                                                                                                                                                                                       
    T0 = fd.define_tensor(shape=[1, -1, -1], contiguity=[None, True, True], dtype=DataType.Float, is_cpu=False, stride_order=[2, 1, 0])
    T1 = fd.define_tensor(shape=[-1, -1], contiguity=[True, True], dtype=DataType.Float, is_cpu=False, stride_order=[1, 0])
    T2 = fd.ops.sum(T1, axes=[1], keepdim=False, dtype=DataType.Null)                                                                                                                                                                         
    T3 = fd.ops.sum(T0, axes=[1, 0], keepdim=False, dtype=DataType.Null)                                                                                                                                                                      
    S4 = fd.define_scalar(4, dtype=DataType.Int)                                                                                                                                                                                              
    V5 = fd.define_vector([S4], dtype=DataType.Int)                                                                                                                                                                                           
    T6 = fd.ops.reshape(T2, new_shape=V5)                                                                                                                                                                                                     
    S7 = fd.define_scalar(4, dtype=DataType.Int)                                                                       
    V8 = fd.define_vector([S7], dtype=DataType.Int)                                                                                                                                                                                           
    T9 = fd.ops.reshape(T3, new_shape=V8)                                                                              
    T10 = fd.ops.mul(T6, T9)                                                                                                                                                                                                                  
    T11 = fd.ops.sum(T10, axes=[0], keepdim=False, dtype=DataType.Null)                                                
    fd.add_output(T11)                                                                                                                                                                                                                        
                                                                                                                       
    nvfuser_fusion_id9(fd)                                                                                                                                                                                                                    
                                                                                                                                                                                                                                              
inputs = [                                                                                                                                                                                                                                        torch.randn((12,), dtype=torch.float32, device='cuda:0').as_strided((1, 3, 4), (12, 4, 1)),            
    torch.randn((12,), dtype=torch.float32, device='cuda:0').as_strided((4, 3), (3, 1)),                                                                                                                                                      
]                                                                                                                      
fd.execute(inputs)
```
fails with
```
RuntimeError: old_rfactor.size() == new_rfactor.size() INTERNAL ASSERT FAILED at "/workspace/Fuser/csrc/dynamic_transform.cpp":652, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. Concretized re
shape rfactor size does not match symbolic rfactor

```

jacobhinkle:
We register extents for concretization when we concretize reshape. In order to do that, we line up `IterDomain`s in the symbolic reshaped TV and the new, concretized one. In cases where the concretized reshape is trivial, such as when the output shape is the same as the input, we do not create a new TV. In those cases, we will have the input to the original `ViewOp` as the concretized output. That input TV might have reduction domains, as in the provided test, in which case we need to filter those out when doing this alignment. This small PR just implements that filtering.

Fixes #1691.

### DIFF
--- a/csrc/dynamic_transform.cpp
+++ b/csrc/dynamic_transform.cpp
@@ -647,6 +647,32 @@ void DynamicTransformConcretizer::concretizeReshape() {
     // Extent expressions often change when concretizing a reshape. Here we
     // replace these in all downstream expressions so that the Fusion looks just
     // like it would have if we had used a static reshape instead.
+    //
+    // Note that Reduction IterDomains might be present in the concretized
+    // reshape. For example, suppose we are given the following dynamic Fusion
+    //
+    //   Inputs:
+    //     T0
+    //   Outputs:
+    //     T3
+    //   T1[ iS2{i0} rS3{i1} ] = sum(T0[ iS0{i0} iS1{i1} ])
+    //   T2[ ?S4{i2} ] = view(T1[ iS2{i0} rS3{i1} ])
+    //   T3[ ?S4{i2} ] = -T2[ ?S4{i2} ]
+    //
+    // Then we will concretize this as
+    //
+    //   Inputs:
+    //     T0
+    //   Outputs:
+    //     T3
+    //   T1[ iS2{i0} rS3{i1} ] = sum(T0[ iS0{i0} iS1{i1} ])
+    //   T3[ iS4{i0} ] = -T1[ iS2{i0} rS3{i1} ]
+    //
+    // Notice here that the ViewOp is gone since we recognized that there is no
+    // transformation to perform. Instead, T1 is used directly in place of T2.
+    // We also replace the extent i2 from the dynamic reshape output T2 with i0,
+    // which is what the code below implements. Since T1 includes a Reduction
+    // IterDomain, we must ignore it in order to match ?S4{i2} with iS2{i0}.
     auto old_rfactor =
         TensorDomain::noReductions(incomplete_out_tv->getMaybeRFactorDomain());
     auto new_rfactor = TensorDomain::noReductions(
--- a/csrc/dynamic_transform.cpp
+++ b/csrc/dynamic_transform.cpp
@@ -647,8 +647,10 @@ void DynamicTransformConcretizer::concretizeReshape() {
     // Extent expressions often change when concretizing a reshape. Here we
     // replace these in all downstream expressions so that the Fusion looks just
     // like it would have if we had used a static reshape instead.
-    auto old_rfactor = incomplete_out_tv->getMaybeRFactorDomain();
-    auto new_rfactor = concrete_reshape_out_tv->getMaybeRFactorDomain();
+    auto old_rfactor =
+        TensorDomain::noReductions(incomplete_out_tv->getMaybeRFactorDomain());
+    auto new_rfactor = TensorDomain::noReductions(
+        concrete_reshape_out_tv->getMaybeRFactorDomain());
     NVF_ERROR(
         old_rfactor.size() == new_rfactor.size(),
         "Concretized reshape rfactor size does not match symbolic rfactor");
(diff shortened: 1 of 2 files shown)


jacobhinkle:
!build

naoyam:
> We register extents for concretization when we concretize reshape. In order to do that, we line up `IterDomain`s in the symbolic reshaped TV and the new, concretized one. In cases where the concretized reshape is trivial, such as when the output shape is the same as the input, we do not create a new TV. In those cases, we will have the input to the original `ViewOp` as the concretized output. That input TV might have reduction domains, as in the provided test, in which case we need to filter those out when doing this alignment. This small PR just implements that filtering.
> 
> Fixes #1691.

Could you give a concrete example? 

> In cases where the concretized reshape is trivial, such as when the output shape is the same as the input, we do not create a new TV. In those cases, we will have the input to the original `ViewOp` as the concretized output. 

This is a bit unexpected for me. Are you saying the reshape would look like: `tv0 = ViewOp(tv0, ...)`? That seems like violating the SSA assumption in the Fusion IR.

jacobhinkle:
> a concrete example

Ah, sorry I should've been more clear. Say we have a dynamic reshape like
```
Inputs:
  T0
Outputs:
  T2
T1[ ?S2{i2} ?S3{i3} ] = view(T0[ iS0{i0} iS1{i1} ])
T2[ ?S4{i2} ?S5{i3} ] = -T1[ ?S2{i2} ?S3{i3} ]
```
Then if it turns out that `i2==i0` and `i3==i1`, then this is actually concretized as
```
Inputs:
  T0
Outputs:
  T2
T2[ iS4{i2} iS5{i3} ] = -T0[ iS0{i0} iS1{i1} ]
```
What happened is we replaced `T1[ ?S2{i2} ?S3{i3} ]` with `T0[ iS0{i0} iS1{i1} ]` in all uses. We did not just replace the definition of `T1` with `set(T0)`.

Now imagine we have instead
```
Inputs:
  T0
Outputs:
  T3
T1[ iS2{i0} rS3{i1} ] = sum(T0[ iS0{i0} iS1{i1} ])
T2[ ?S4{i2} ] = view(T1[ iS2{i0} rS3{i1} ])
T3[ ?S4{i2} ] = -T2[ ?S4{i2} ]
```
This is like the case in the issue/test. Now, when we concretize we have
```
Inputs:
  T0
Outputs:
  T3
T1[ iS2{i0} rS3{i1} ] = sum(T0[ iS0{i0} iS1{i1} ])
T3[ iS4{i2} ] = -T1[ iS2{i0} rS3{i1} ]
```
Here we have replaced `T2[ ?S4{i2} ]` with `T1[ iS2{i0} rS3{i1} ]`.

naoyam:
Thanks. Could you add that to the code?

csrc/dynamic_transform.cpp
```diff
@@ -647,8 +647,10 @@ void DynamicTransformConcretizer::concretizeReshape() {
     // Extent expressions often change when concretizing a reshape. Here we
     // replace these in all downstream expressions so that the Fusion looks just
     // like it would have if we had used a static reshape instead.
-    auto old_rfactor = incomplete_out_tv->getMaybeRFactorDomain();
-    auto new_rfactor = concrete_reshape_out_tv->getMaybeRFactorDomain();
+    auto old_rfactor =
+        TensorDomain::noReductions(incomplete_out_tv->getMaybeRFactorDomain());
+    auto new_rfactor = TensorDomain::noReductions(
+        concrete_reshape_out_tv->getMaybeRFactorDomain());
```
naoyam, line 653: Now I understand `concrete_reshape_out_tv` can have reduction domains. When is that also the case with `incomplete_out_tv`?

jacobhinkle: That should not happen, since that domain is always created by the `reshape` op. I'll remove that filter.

naoyam (approved):
LGTM. Thanks for the fix.

PR #1695, opened 2024-01-30
Check that the reduced axis is sharded on producer in isLowerableToCommunication

samnordmann:
Currently, a reduction is lowerable to a communication iff only one axis is reduced and this axis is sharded across devices on the **producer** side.
Before this patch, we would mistakenly check that the axis is sharded on **consumer** side, which led to some runtime assert error.

### DIFF
--- a/csrc/multidevice/lower_communication.cpp
+++ b/csrc/multidevice/lower_communication.cpp
@@ -549,8 +549,8 @@ bool isLowerableToCommunication(Expr* expr) {
     // get the reduced axis
     std::vector<IterDomain*> reduction_axis;
     std::copy_if(
-        out->getMaybeRFactorDomain().begin(),
-        out->getMaybeRFactorDomain().end(),
+        out->getRootDomain().begin(),
+        out->getRootDomain().end(),
         std::back_inserter(reduction_axis),
         [](IterDomain* id) { return id->isReduction(); });
     // check whether the reduction involves only one axis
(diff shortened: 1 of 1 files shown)


naoyam (approved):
LGTM

csrc/multidevice/lower_communication.cpp
```diff
@@ -538,18 +539,28 @@ std::vector<std::shared_ptr<Communication>> lowerCommunication(
...
-    NVF_ERROR(out->isA<TensorView>(), "output is not a TensorView");
-    auto out_tv = out->as<TensorView>();
+    auto in = expr->as<ReductionOp>()->in()->as<TensorView>();
+    auto out = expr->as<ReductionOp>()->out()->as<TensorView>();
+    // get the reduced axis
     std::vector<IterDomain*> reduction_axis;
     std::copy_if(
-        out_tv->getMaybeRFactorDomain().begin(),
-        out_tv->getMaybeRFactorDomain().end(),
+        out->getMaybeRFactorDomain().begin(),
```
naoyam, line 552:
```suggestion
        out->getRootDomain().begin(),
```

It doesn't actually matter as there should be no rfactor here, but semantically when a consumer tensor is matched up with a producer tensor, the root domain of the consumer should be used.

samnordmann: ok thanks that's helpful for my understanding

PR #1699, opened 2024-01-30
Associate each fusion cache with its local rank in distributed setting.

rdspring1:
### Problem:
Currently, automatic serialization saves a single cache regardless of the number of devices. In a distributed setting, each process restores its fusion cache from the same common workspace. However, this workspace only contains the CUDA kernels for a single device. The remaining processes must recompile the kernels for their devices.

### Solution:
A separate process is created for each device with `ddp` or `fsdp` and each process contains a separate `FusionCache`. This PR associates each fusion cache with its local rank in a distributed setting, allowing automatic serialization to create a separate workspace for each device. During deserialization, each process loads the workspace associated with its local rank.


rdspring1:
!build

jjsjann123:
sorry for jumping on this late. looks like we now have conflicts on this PR.

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

PR #1703, opened 2024-01-31
`eager` and `torch.compile` benchmarks for `layernorm`

Priya2698:
Issue #1668.

Adds IOByte computation and benchmarking for Layernorm baselines (eager & torchcompile)

### DIFF
--- a/benchmarks/python/test_layernorm_bwd.py
+++ b/benchmarks/python/test_layernorm_bwd.py
@@ -93,6 +93,7 @@ def layernorm_bwd_fusion(
 
 
 def layernorm_bwd_iobytes(size: tuple, dtype: torch.dtype):
+    # Manual IOBytes computation since nvfuser input/outputs (in_tensor, grad_out, mean, invstd, weigts) differ from baselines (out, grad_out)
     # Total IO bytes = in_tensor (size, dtype) + grad_out (size, dtype) + mean (size[0], float) +
     #       invstd (size[0], float) + weights (size[1], dtype) +
     #       grad_in (size, dtype) + grad_weights (size[1], dtype)+ grad_bias (size[1], dtype)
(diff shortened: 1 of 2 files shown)


naoyam (commented):
I'll let @jacobhinkle review this PR.

benchmarks/python/test_layernorm_bwd.py
```diff
@@ -91,9 +92,19 @@ def layernorm_bwd_fusion(
     fd.add_output(T28)
 
 
+def layernorm_bwd_iobytes(size: tuple, dtype: torch.dtype):
+    # Total IO bytes = in_tensor (size, dtype) + grad_out (size, dtype) + mean (size[0], float) +
+    #       invstd (size[0], float) + weights (size[1], dtype) +
+    #       grad_in (size, dtype) + grad_weights (size[1], dtype)+ grad_bias (size[1], dtype)
```
jacobhinkle, line 98: Could you comment on how this might differ from the default computation in the eager case?

benchmarks/python/test_layernorm_fwd.py
```diff
@@ -51,9 +52,27 @@ def layernorm_fwd_fusion(
     fd.add_output(T14)
 
 
+def layernorm_fwd_fn(inputs: list):  # [in_tensor, weights, bias]
```
jacobhinkle, line 55: Suggestion: drop `_fn` from the name here

benchmarks/python/test_layernorm_fwd.py
```diff
@@ -64,7 +83,7 @@ def test_layernorm_fwd_benchmark(
     clear_cuda_cache()
```
jacobhinkle, line 84: Unwrapping the `size` tuple might help with clarity here since `size[1]` is a little tough to decipher at a glance.

jacobhinkle (approved):
LGTM other than, like in #2038, comments and reference to #1725 would really help to walk a new reader through the issue and the need for iobytes manual calculation.

PR #1708, opened 2024-01-31
Weekly Benchmarks Input Range

Priya2698:
For weekly benchmarks, we will focus on 3 states:

1. Steady-State - full machine
2. Just filled the machine
3. Latency limited

The batch sizes are selected to cover these 3 states and the hidden sizes will be varied at a finer granuarity of `step_size=8`.

No major changes are required in our setup for weekly benchmarks. However, pytest is memory intensive and we may need to split the tests in each file to avoid CPU OOM.
~As a solution, we can use `pytest-split` when integrating these into the CI.~
CI uses `pytest-shard` which should work well.

**Running weekly benchmark coverage:**
`BENCHMARK_MODE=weekly pytest <options> <benchmark_file.py>`

CC: @xwang233 @kevinstephano

### DIFF
--- a/python_benchmarks/global_params.py
+++ b/python_benchmarks/global_params.py
@@ -4,6 +4,12 @@ from nvfuser import DataType
 from .core import DEVICE_PROPERTIES
 import numpy as np
 import itertools
+import os
+
+# BENCHMARK_MODE = weekly/nightly.
+BENCHMARK_MODE = os.getenv("BENCHMARK_MODE")
+if not BENCHMARK_MODE:
+    BENCHMARK_MODE = "nightly"
 
 # Datatypes to benchmark
 FLOAT_DTYPES = [torch.float32]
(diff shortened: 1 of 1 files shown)


python_benchmarks/global_params.py
```diff
@@ -44,19 +50,35 @@
...
 # Utility function to generate input sizes for benchmarks
 def generate_input_sizes(dims: Union[int, List] = 2) -> List[Tuple]:
+    """
+    The weekly vs nightly input ranges only differ for 2D inputs currently.
+    Nightly input range:
+        Batch size: [16->16384] Hidden size: [768, 4*18432] (step size = 256)
+    Weekly input range:
+        Batch size:
+            [16]: Latency bound state
+            [512, 1024]: Just filled the machine
```
liqiangxl, line 60: what's the meaning of `just filled the machine`?

Priya2698: Just filled the machine -> 1 full wave.
From the SOL vs hidden size graphs for different batch sizes, these are the batch sizes where we just start to be bandwidth bound instead of latency bound.

liqiangxl (approved):
LGTM.

naoyam:
I'm not sure how these sizes are determined. Wouldn't they depend on problems as well as device properties such as the number of SMs?

@Priya2698 I'd like to review the current status and plan. I'll schedule a meeting.

PR #1711, opened 2024-02-02
NVFUSER_DISTRIBUTED instead of USE_DISTRIBUTED

cowanmeg:
Adds `NVFUSER_DISTRIBUTED` macro and updates multidevice #ifdef guards to use `NVFUSER_DISTRIBUTED` instead of `USE_DISTRIBUTED`.

This will decouple nvfuser's distributed support from from Pytorch's `USE_DISTRIBUTED` macro so that our CI can build nvfuser without multidevice support.  `NVFUSER_DISTRIBUTED` is dependent `USE_DISTRIBUTED` being defined. 

From `setup.py` add `--build-without-distributed` to build without multidevice support.

FYI @samnordmann

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,10 +15,11 @@ option(NVFUSER_BUILD_WITH_ASAN "Build nvFuser with asan" OFF)
 
 include(CMakeDependentOption)
 cmake_dependent_option(NVFUSER_DISTRIBUTED "" ON
-	"$ENV{USE_DISTRIBUTED}" OFF)
+	"USE_DISTRIBUTED" OFF)
 if (NVFUSER_DISTRIBUTED)
   add_compile_definitions(NVFUSER_DISTRIBUTED)
 endif()
+message(STATUS "Setting NVFUSER_DISTRIBUTED=${NVFUSER_DISTRIBUTED}")
 
 if(NOT NVFUSER_CPP_STANDARD)
   set(NVFUSER_CPP_STANDARD 20)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,13 @@ set(NVFUSER_THIRD_PARTY_DIR "${NVFUSER_ROOT}/third_party")
 option(NVFUSER_STANDALONE_BUILD_WITH_UCC "" OFF)
 option(NVFUSER_BUILD_WITH_ASAN "Build nvFuser with asan" OFF)
 
+include(CMakeDependentOption)
+cmake_dependent_option(NVFUSER_DISTRIBUTED "" ON
+	"$ENV{USE_DISTRIBUTED}" OFF)
+if (NVFUSER_DISTRIBUTED)
+  add_compile_definitions(NVFUSER_DISTRIBUTED)
+endif()
+
 if(NOT NVFUSER_CPP_STANDARD)
   set(NVFUSER_CPP_STANDARD 20)
 endif()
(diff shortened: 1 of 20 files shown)


tfogal (approved):
Great to see this! Thank you, Meghan.

CMakeLists.txt
```diff
@@ -13,6 +13,13 @@ set(NVFUSER_THIRD_PARTY_DIR "${NVFUSER_ROOT}/third_party")
 option(NVFUSER_STANDALONE_BUILD_WITH_UCC "" OFF)
 option(NVFUSER_BUILD_WITH_ASAN "Build nvFuser with asan" OFF)
 
+include(CMakeDependentOption)
+cmake_dependent_option(NVFUSER_DISTRIBUTED "" ON
+	"$ENV{USE_DISTRIBUTED}" OFF)
```
xwang233, line 18: I'm a bit worried about dependency on `USE_DISTRIBUETD` from environment variable. Although we build pytorch from source with or without this variables, building against a pytorch wheel (with distributed package) won't have this environment variable set by default.

`USE_DISTRIBUTED` definition is inherited from linking libtorch. Is it possible to parse that from CMake side? If not, we can add something similar like this https://github.com/NVIDIA/Fuser/blob/c1ac1196ea546fdb77cf6732936aa4fd4848b209/tools/gen_nvfuser_version.py#L32-L46, which runs a `python -c 'import torch; torch._C._show_config()` subprocess to find out if libtorch has distributed package.

This can be a mid-term plan and doesn't need to be fixed now.

jjsjann123: strangely I don't see `USE_DISTRIBUTED` in `torch._C._show_config()` inside our container.

setup.py
```diff
@@ -321,6 +328,8 @@ def cmake(install_prefix: str = "./nvfuser"):
         cmd_str.append("-DBUILD_NVFUSER_BENCHMARK=ON")
     if BUILD_WITH_ASAN:
         cmd_str.append("-DNVFUSER_BUILD_WITH_ASAN=ON")
+    if BUILD_WITHOUT_DISTRIBUTED:
+        cmd_str.append("-DNVFUSER_DISTRIBUTED=OFF")
```
jjsjann123, line 332: I'm wondering how this would interact with the envvar in our CMakeLists.txt. i.e. what if the two disagrees against each other?

Also, looks like via setuptools, we can only force disable the flag. We still need to go through environment variable to mark it as `enabled`. This feels like something that worth putting down in build doc.

https://github.com/NVIDIA/Fuser/wiki/Developer-guide

CMakeLists.txt
cowanmeg: I also do not see the `USE_DISTRIBUTED` inside our container.
I think we can parse it from `torch.distributed.is_available()`?

setup.py
cowanmeg: The dependent option will not let two flags disagree with each other.
If USE_DISTRIBUTED is off then NVFUSER_DISTRIBUTED is forced to off and cannot be changed by the user.
If USE_DISTRIBUTED is on then NVFUSER_DISTRIBUTED is an optional flag that defaults to on. Using`--no-distributed` will set NVFUSER_DISTRIBUTED to off.

If we properly parse USE_DISTRIBUTED from pytorch, then we shouldn't run into any problems.

CMakeLists.txt
xwang233: We can use `torch._C._has_distributed()`. 

https://github.com/pytorch/pytorch/blob/200108c6e611875f9d4e3a1c7a1cec282b57f031/torch/csrc/Module.cpp#L526-L533

setup.py
jjsjann123: Got'ya. I was a bit nervous earlier with our usage of env var `USE_DISTRIBUTED`. Looks like we are query the proper torch config. :bow:

jjsjann123 (approved):
LGTM

PR #1712, opened 2024-02-02
Make nvfuser matmul benchmarks HSH instead of HSS

jacobhinkle:
This matches the `at::matmul` baselines.

This PR also adds a few more problem sizes, and runs each eagermode baseline with and without FP16 reduction allowed.

### DIFF
--- a/benchmark/matmul.cpp
+++ b/benchmark/matmul.cpp
@@ -593,7 +593,7 @@ static std::vector<std::tuple<T, T, T>> sizeProduct(
 static void MatmulShapeEager(
     benchmark::internal::Benchmark* b,
     std::vector<std::tuple<long int, long int, long int>> sizes) {
-  b->ArgNames({"M", "N", "K", "FP16Reduction"});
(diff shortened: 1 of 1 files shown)


drzejan2:
The changes look good to me, thanks for adding casting the output to desired type!

@mmigdal-nv are the test shapes added by @jacobhinkle in the PR the only ones that should be added, or it would be best to add some additional?

drzejan2 (approved):

jacobhinkle:
Per offline convo with Michel, we may want to overhaul the set of problem sizes for benchmarking. In particular, we might want to track problem sizes that influence things like bandwidth calculation in internal project. So for now, I'll merge as is and we can revisit problem sizes separately.

jacobhinkle:
Per offline convo with Michel, we may want to overhaul the set of problem sizes for benchmarking. In particular, we might want to track problem sizes that influence things like bandwidth calculation in internal project. So for now, I'll merge as is and we can revisit problem sizes separately.

PR #1713, opened 2024-02-02
Add a visitor for ValGraph

naoyam:
Used in the loop promotion analysis. Extracted from #32


naoyam:
!build

csrc/val_graph_visitor.cpp
```diff
@@ -0,0 +1,118 @@
...
+  // ValGroups: {{i0, i0*1}, {1}}
+  //
+  // Then, {i0, i0*1} and {1} would be visited first, then the merge
+  // expr group would be visited. {i0, i0*1} is also an output group
+  // of the merge but since it's already in the visited set, it would
+  // not be visited again.
+  //
+  // See also IdModelTest.ValGraphStmtSort3 for a concrete example.
+  auto is_val_ready = [&](const ValGroup& val_group) -> bool {
+    const ExprGroups& unique_defs = graph().getDefinitions(val_group);
```
wujingyue, line 50: Can you remind me when a `val_group` can have multiple defining `ExprGroup`s?

csrc/val_graph_visitor.cpp
```diff
@@ -0,0 +1,118 @@
...
+    return std::all_of(
+        unique_defs.begin(), unique_defs.end(), [&](ExprGroup expr_group) {
+          return expr_group->empty() || visited_exprs.has(expr_group) ||
+              terminating_inputs.has(val_group) ||
+              graph().isTrivialExprGroup(expr_group);
+        });
+  };
+
+  while (!to_visit_ids.empty() || !to_visit_exprs.empty()) {
```
wujingyue, line 59: Is it OK to split the two conditions to two loops? It feels to me that we can visit all `to_visit_exprs` first and then visit all `to_visit_ids`.

csrc/val_graph.h
```diff
@@ -199,6 +202,11 @@ class ValGraph {
   // be the only call in ValGraph to mapThroughExpr.
   void maybeMapThroughExprs(Expr* expr0, Expr* expr1, bool forward);
 
+  // Returns if the expression group has an input id group that matches an
+  // output id group. This means traversing on this expression doesn't actually
```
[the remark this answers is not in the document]
naoyam: Thanks for pointing this. The comment is actually stale. The stmt sort visitor does visit trivial expr groups as well.

csrc/val_graph_visitor.cpp
naoyam: For example, suppose we have a 1D tensor and a 2D tensor, `t0` and `t1` and:

```
t0: [i0]
t1: [i1, i2, i3]

t2 = reshape(t0, {i4, i5})
// t2: [i4, i5]

t3 = reshape(t1, {i6, i7})
// t3: [i6, i7]

t4 = add(t2, t3)
// t4: [i8, i9]
```

From `t4 = add(t2, t3)`, we would create a val group of `{i4, i6, i8}` as well as `{i5, i7, i9}`. 

The definition of `i4` should be a split of `i0`, whereas `i6` may be something like merge of `i1`, `i2`, `i3` and then split. So the defining expression of `i6` is different from that of `i4`. And since `i4` and `i6` compose a single group, `{i4, i6, i8}`, that group has multiple different definitions. Same is true for the group of `{i5, i7, i9}`.

csrc/disjoint_set.h
```diff
@@ -92,6 +92,15 @@ class VectorOfUniqueEntries {
     return any_added;
   }
 
+  // Remove and returns the first element in vector. Note that this is
+  // a costly operation as the underlying container is std::vector.
```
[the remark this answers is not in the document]
naoyam: I just realized we can easily replace with `std::deque` when `popFront` is currently used. Will remove.

csrc/val_graph_visitor.cpp
naoyam: Actually, that's what the two inner loops do. Since visiting exprs or ids may add more ids or exprs to visit, respectively, these two inner loops need to be repeated. Or are you thinking something different?

naoyam:
!build

csrc/val_graph_visitor.cpp
```diff
@@ -0,0 +1,126 @@
...
+    const ExprGroups& unique_defs = graph().getDefinitions(val_group);
+    return std::all_of(
+        unique_defs.begin(), unique_defs.end(), [&](ExprGroup expr_group) {
+          return expr_group->empty() || visited_exprs.has(expr_group) ||
+              terminating_inputs.has(val_group) ||
+              graph().isTrivialExprGroup(expr_group);
+        });
+  };
+
+  while (!to_visit_vals.empty() || !to_visit_exprs.empty()) {
```
wujingyue, line 60: The outer loop can be simplified to

```
bool something_was_processed;
do {
  something_was_processed = false;
  ...
} while (something_was_processed);
// outside the loop
NVF_ERROR(to_visit_vals.empty() && to_visit_exprs.empty());
```

The behavior remains the same, because when the two to-visit lists are empty there's nothing to process and `something_processed` will trivially become false.

csrc/val_graph_visitor.cpp
```diff
@@ -0,0 +1,126 @@
...
+    // Process expressions first as all definitions of vals have to be
+    // processed before we can process that val.
+
+    // Detect if nothing has been processed which would put us in an infinite
+    // loop
+    bool something_was_processed = false;
+    std::deque<ExprGroup> still_to_visit_exprs;
```
wujingyue, line 67: I think we can get rid of `still_to_visit_exprs`. Let me know if the following makes sense and it's certainly fine to do it in a separate PR. 

All we need is to guarantee `to_visit_*` contains only the ready ones. An unready expr/val can be attempted multiple times to be added to `to_visit_*`, but only when its last-to-visit operand is handled it's actually added to `to_visit_*`. This probably saves some time because an expr/val don't need to be bounced between `to_visit_*` and `still_to_visit_*` potentially many times.

csrc/val_graph_visitor.cpp
```diff
@@ -0,0 +1,126 @@
...
+  // not be visited again.
+  //
+  // See also IdModelTest.ValGraphStmtSort3 for a concrete example.
+  auto is_val_ready = [&](const ValGroup& val_group) -> bool {
+    const ExprGroups& unique_defs = graph().getDefinitions(val_group);
+    return std::all_of(
+        unique_defs.begin(), unique_defs.end(), [&](ExprGroup expr_group) {
+          return expr_group->empty() || visited_exprs.has(expr_group) ||
+              terminating_inputs.has(val_group) ||
+              graph().isTrivialExprGroup(expr_group);
```
wujingyue, line 56: This seems wrong. IIUC, an ExprGroup is considered trivial when its input ValGroups and its output ValGroups have at least one common element. However, even when that happens, that ExprGroup still needs to be `handle`d before being treated as visited, doesn't it?

csrc/val_graph_visitor.cpp
naoyam: Thanks. Makes sense. Applied the change: e33a005

csrc/val_graph_visitor.cpp
naoyam: I think you're right. Let me clean it up.

wujingyue: (As I said, it's certainly fine to do it in a separate PR. I mainly wanted to run this idea through you to make sure I understand your code correctly.)

csrc/val_graph_visitor.cpp
```diff
@@ -0,0 +1,142 @@
...
+      }
+    }
+
+    std::swap(to_visit_vals, still_to_visit_vals);
+
+  } while (something_was_processed);
+
+  if (!to_visit_vals.empty()) {
+    std::stringstream ss;
+    ss << "Remaining Vals to visit:";
```
wujingyue, line 125:
```suggestion
    ss << "The graph has an infinite loop. The following Vals should be visited but are never ready:";
```

csrc/val_graph_visitor.cpp
```diff
@@ -0,0 +1,142 @@
...
+    for (const ValGroup& vg : to_visit_vals) {
+      ss << " " << nvfuser::toString(vg);
+    }
+    NVF_ERROR(false, ss.str());
+  }
+
+  if (!to_visit_exprs.empty()) {
+    std::stringstream ss;
+    ss << "Remaining Exprs to visit:";
```
wujingyue, line 134:
```suggestion
    ss << "The graph has an infinite loop. The following Exprs should be visited but are never ready:";
```

csrc/val_graph_visitor.cpp
naoyam: Updated: b92ec79cfa013dbc8512afc1e0157f6c36f723cd

csrc/val_graph_visitor.cpp
naoyam: I think I see what you mean but struggled to create a case that actually fails. 

Updated the code. Does this make sense? b891e36

test/test_id_model.cpp
```diff
@@ -37,4 +38,350 @@ TEST_F(IdModelTest, DetectSelfMapping) {
...
+  // exprg{21}: Split iS30
+  // exprg{55 67}: Split iS57
+  // exprg{13}: Split iS24
+  // exprg{29}: Split iS36
+  // exprg{37 49}: Split iS42
+  // exprg{23}: Split iS31
+  // exprg{15}: Split iS25
+  // exprg{31}: Split iS37
+
+  std::vector<Expr*> ref_order;
```
wujingyue, line 358: IIUC, `ref_order` is **a** correct total order. Other total orders may be correct as well but just not produced due to the way IdModel is implemented. This may not be a big deal for now, but I can imagine people who later change the implementation of IdModel may run into such different-from-ref-but-still-correct problems. 

Have you considered to check only the partial order? E.g. for each def-use, check the def is ordered before the use. This would be the minimal requirement.

wujingyue (approved):
Sorry for the delay. I wanted to finish reviewing on Friday but got sick.

csrc/val_graph_visitor.cpp
```diff
@@ -0,0 +1,150 @@
...
+        unique_defs.begin(), unique_defs.end(), [&](ExprGroup expr_group) {
+          if (expr_group->empty() || visited_exprs.has(expr_group) ||
+              terminating_inputs.has(val_group)) {
+            return true;
+          }
+          // Handle trivial expr groups. This expr_group is not
+          // visited yet, which means there're input ValGroups that
+          // are not yet visited. If those not-visited inputs are
+          // actually the same as val_group, visit val_group at this
+          // point to resolve the circular dependency.
```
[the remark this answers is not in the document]
naoyam: Exactly. At this point, the order doesn't matter as the traversal in the loop promotion analysis should not have these cycles. This may matter in the future when generating different mappings and graphs, but at this point we could also just make it fail if such cycles were detected. I'll add more notes.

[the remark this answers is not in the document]
wujingyue: > at this point we could also just make it fail if such cycles were detected

That would be my preference. Instead of adding non-tested, arguable features, it's usually better to fail explicitly for now (if it simplifies).

test/test_id_model.cpp
naoyam: Right, in fact that was my initial validation, but it's also important to have a deterministic order, so I switched to the current version. That's how I debugged the non-deterministic issue of #1714. 

We could also do the partial order validation when a failure with the total order validation is failed. That way, the failure message could say something like "an unexpected but valid order is detected" rather than just saying "an unexpected order is detected". I'll consider that if these tests create false alarms frequently.

PR #1719, opened 2024-02-03
[WIP] Half-precision reduction for split-K

jacobhinkle:
This cange implements a non-default option to use a reduced-precision work buffer for the split-K grid reduction. Note that this does not mean the accumulator precision is reduced: register buffers are still `Float`. However, for split-K we might have say 5 segments reduced in single precision that need to be grid-reduced. That grid reduction requires global writes and reads, and this change lets us reduce the precision just for that IO.

Note that this mimics the default behavior of cublas and PyTorch/ATen.


csrc/device_lower/pass/index.cpp
```diff
@@ -650,7 +650,7 @@ void IndexLowering::handleSerialGridReduction(
   }
   auto work_buffer_domain = IrBuilder::create<TensorDomain>(work_buffer_root);
   auto work_buffer_tv = IrBuilder::create<TensorView>(
-      work_buffer_domain, out_tv->dtype(), MemoryType::Global);
+      work_buffer_domain, DataType::Half, MemoryType::Global);
```
jacobhinkle, line 653: Placeholder. This will be removed once we update the interface for requesting serial grid reduction to also specify the precision.

runtime/grid_reduction.cu
```diff
@@ -667,4 +667,62 @@ __device__ void serialReductionStep(
   }
 }
 
+// Same as above, but cast to lower precision for work buffer IO
+template <int64_t vec_size, typename Twork, typename Func>
+__device__ void serialReductionStep(
+    float* out,
+    float* in,
+    float init,
+    volatile Twork* work,
```
jacobhinkle, line 676: This template's redundant since `Twork = float` might match this if we're not careful. Instead it might be best to just check `is_same` for the types of `out` and `work` and dispatch from there to separately-named helper functions.

runtime/grid_reduction.cu
```diff
@@ -667,4 +667,62 @@ __device__ void serialReductionStep(
...
+    loadGlobalToLocal<Twork, vec_size, true, CacheOp::Global>(work_reg, work);
+#pragma unroll
+    for (int i = 0; i < vec_size; ++i) {
+      float work_float = NAN;
+      if constexpr (std::is_same<Twork, __half>::value) {
+        work_float = __half2float(work_reg[i]);
+      } else if constexpr (std::is_same<Twork, __bfloat>::value) {
+        work_float = __bfloat2float(work_reg[i]);
+      } else {
+        // static_assert(false);
```
jacobhinkle, line 708: Shouldn't be needed since I assert at the start, but I don't know why this caused compile to fail unless, as mentioned above, this template is also matching the `Twork=float` case...

runtime/type_conversion.cu
```diff
@@ -0,0 +1,32 @@
...
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+// clang-format on
+
+
+// This should be specialized for each combination of types that cannot be
+// automatically cast by static_cast
+template <typename TO, typename FROM>
+__device__ __inline__ TO castFloating(FROM x) {
```
jacobhinkle, line 13: TODO: we could support vectorized casts here by adding a `vec_size` template arg then specializing to the usual set of vectorization widths and using `__half22float2` and friends.

jacobhinkle:
A note about vectorization with half reduction:

We currently schedule vectorized reduction by vectorizing the `ReductionOp` output (see #1528). However, with this PR, we might have single precision output and half precision reduction buffer. This means that while we would ideally use a vectorization width of 8 for the half-precision reduction, we will be bound by the single-precision output type. If we try and vectorize at width 8, we will hit an error. We could either special-case for this error, or introduce some other way to indicate vectorization of the temporary buffer. Currently we don't schedule that temporary buffer: we allocate it according to the leaf domain of the (single-precision) output at lowering (`index.cpp`). A more flexible way might be to actually create a global `TensorView` at scheduling and attach it as an attribute to the `ReductionOp`. That tensor's leaf domain would equal its root/allocation domain and could hold vectorization, grouping, dtype, etc.

PR #1722, opened 2024-02-05
Add the missing `clang-format on` and reformat.

wujingyue:


naoyam (approved):
Thanks for catching it

PR #1725, opened 2024-02-06
Set IOBytes for benchmarks

Priya2698:
For uniformity of IO bytes, and compairson between nvFuser and baselines of `eager` and `torch.compile` mode, set the IO bytes instead of computing based on inputs / outputs of the benchmarked function. 

The inputs / outputs to the torch functions may not always be the same as nvFuser.
For instance, in `layernorm fwd`, nvFuser definition returns the `mean` and `invstd`, but `torch.nn.functional.layer_norm` does not.

### DIFF
--- a/python_benchmarks/test_softmax_fwd.py
+++ b/python_benchmarks/test_softmax_fwd.py
@@ -77,9 +77,7 @@ def test_softmax_fwd_nvf_benchmark(
         fd.validate(inputs, [eager_output])
 
     if not disable_benchmarking:
-        run_benchmark(
-            benchmark, fd.execute, inputs, iobytes=softmax_fwd_iobytes(size, dtype)
-        )
+        run_benchmark(benchmark, fd.execute, inputs)
 
 
 @pytest.mark.parametrize("compile", [False, True], ids=["eager", "compile"])
--- a/python_benchmarks/test_softmax_fwd.py
+++ b/python_benchmarks/test_softmax_fwd.py
@@ -4,6 +4,7 @@ from nvfuser.pytorch_utils import torch_dtype_to_nvfuser_dtype
 from .core import run_benchmark, clear_cuda_cache
 import torch
 from .global_params import generate_input_sizes, FLOAT_DTYPES, PROMOTE_DTYPES
+import numpy as np
 
 
 def softmax_fwd_fusion(
@@ -48,6 +49,11 @@ def softmax_fwd_fn(inputs: list):  # [in_tensor, reduction_axis]
     return torch.nn.functional.softmax(inputs[0], inputs[1])
 
 
+def softmax_fwd_iobytes(size: tuple, dtype: torch.dtype):
+    # Total IO bytes = input + output
+    return int(np.prod(size) * dtype.itemsize * 2)
+
+
 @pytest.mark.parametrize("size", generate_input_sizes(dims=2))
 @pytest.mark.parametrize("dtype", FLOAT_DTYPES)
 @pytest.mark.parametrize("reduction_axis", [0, 1])
@@ -71,7 +77,9 @@ def test_softmax_fwd_nvf_benchmark(
         fd.validate(inputs, [eager_output])
 
     if not disable_benchmarking:
-        run_benchmark(benchmark, fd.execute, inputs)
+        run_benchmark(
+            benchmark, fd.execute, inputs, iobytes=softmax_fwd_iobytes(size, dtype)
+        )
 
 
 @pytest.mark.parametrize("compile", [False, True], ids=["eager", "compile"])
@@ -87,8 +95,10 @@ def test_softmax_fwd_baseline_benchmark(
 ):
     clear_cuda_cache()
     input = torch.randn(size, device="cuda", dtype=dtype)
+
     run_benchmark(
         benchmark,
         torch.compile(softmax_fwd_fn) if compile else softmax_fwd_fn,
         [input, reduction_axis],
+        iobytes=softmax_fwd_iobytes(size, dtype),
     )
--- a/python_benchmarks/core.py
+++ b/python_benchmarks/core.py
@@ -240,7 +240,7 @@ class NVFBenchmark:
     ) -> None:
         """
         Utility function to compute metrics for the target function.
-        
+
         Args:
             inputs: Inputs to the target function
             outputs: Outputs of the target function
--- a/python_benchmarks/core.py
+++ b/python_benchmarks/core.py
@@ -240,6 +240,13 @@ class NVFBenchmark:
     ) -> None:
         """
         Utility function to compute metrics for the target function.
+        
+        Args:
+            inputs: Inputs to the target function
+            outputs: Outputs of the target function
+            iobytes (Optional): When given, IO bytes computation is skipped
+                and this is used to compute the metrics.
+
         Current metrics:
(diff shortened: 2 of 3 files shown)


jacobhinkle:
How are we discovering inputs/outputs for these functions? For a `torch.autograd.Function`, the `.forward` will typically save some tensors for the backward pass. Are we counting those? For example, `F.instance_norm` might actually be saving the mean and invstd, but it's just not returning them in the return value.

Priya2698:
> How are we discovering inputs/outputs for these functions? For a `torch.autograd.Function`, the `.forward` will typically save some tensors for the backward pass. Are we counting those? For example, `F.instance_norm` might actually be saving the mean and invstd, but it's just not returning them in the return value.

I am computing the IO bytes externally based on the nvFuser definitions and passing them when running the benchmarks. So the IO bytes will not be computed based on the outputs returned but instead use the passed value.

The PR is addressing the issue you are mentioning. While these intermediate outputs are returned from nvFuser definitions, torch does not return them although it holds them for the backward pass.

python_benchmarks/test_softmax_fwd.py
```diff
@@ -87,8 +95,10 @@ def test_softmax_fwd_baseline_benchmark(
 ):
     clear_cuda_cache()
     input = torch.randn(size, device="cuda", dtype=dtype)
+
     run_benchmark(
         benchmark,
         torch.compile(softmax_fwd_fn) if compile else softmax_fwd_fn,
         [input, reduction_axis],
+        iobytes=softmax_fwd_iobytes(size, dtype),
```
Priya2698, line 103: @jacobhinkle I am using the IO bytes computation to pass the IO bytes for the benchmark. This will skip the IO bytes computation and use this value instead.

python_benchmarks/test_softmax_fwd.py
```diff
@@ -48,6 +49,11 @@ def softmax_fwd_fn(inputs: list):  # [in_tensor, reduction_axis]
     return torch.nn.functional.softmax(inputs[0], inputs[1])
 
 
+def softmax_fwd_iobytes(size: tuple, dtype: torch.dtype):
+    # Total IO bytes = input + output
+    return int(np.prod(size) * dtype.itemsize * 2)
```
Priya2698, line 54: These functions will be added to the benchmarks to manually compute IO bytes to avoid mismatch in IOBytes since some intermediate outputs will not be returned with torch but are returned in nvfuser definitions.

kevinstephano:
This does not have to be part of this PR but the intention of creating the nvFuser profiler was so that you can just grab the I/O bytes for the fusion and directly compare them without having to figure anything out.

Priya2698:
> This does not have to be part of this PR but the intention of creating the nvFuser profiler was so that for benchmarking you do not need to create special logic per benchmark, you should be able to grab the I/O bytes for the fusion and directly compare them without having to figure anything out. 

>This does minimally assume you are profiling nvFuser in addition to eager mode and whatever else you would like to compare.

This is the key reason behind manual computation since it makes the baselines independent of the nvFuser benchmarking.
If we were to run all the baselines in the same process as the nvFuser (and always run nvFuser execution before), we could easily just return the IO bytes computed in the nvFuser execution to be used for the baselines as well. This is preferable to manual computation.

I can organize the benchmarks so that this is true, but that removes the ability to independently run only baselines.

What do you think? @kevinstephano @jacobhinkle 
We could go with either of the designs: For weekly runs, it does not make a difference since we will run all executors. For local benchmarking, users may wish to independently run the baselines?

naoyam:
> The inputs / outputs to the torch functions may not always be the same as nvFuser.
For instance, in layernorm fwd, nvFuser definition returns the mean and invstd, but torch.nn.functional.layer_norm does not.

So, are we not evaluating the same configuration?

Priya2698:
> > The inputs / outputs to the torch functions may not always be the same as nvFuser.
> > For instance, in layernorm fwd, nvFuser definition returns the mean and invstd, but torch.nn.functional.layer_norm does not.
> 
> So, are we not evaluating the same configuration?

What do you mean by the same configuration?

The difference in the automatically computed IO bytes comes from the difference in what torch returns vs nvFuser. While nvFuser fwd definition returns `mean` and `invstd` for `layernorm fwd`, `torch.nn.functional.layer_norm` does not explicitly return them but they are stored for backward pass. 

So in these cases where the intermediate tensors are returned explicitly from `nvFuser` definitions but not torch does not leads to a mismatch in the IOBytes computed. This will lead to differences when we compute bandwidth and may lead to inaccurate comparison.

naoyam:
Oh, I see. Looks like I misunderstood. I thought the nvfuser version stores some data that is not stored by the eager version. Looks like both do the same stores, but it's just the interface that is different between the two.

naoyam:
> I can organize the benchmarks so that this is true, but that removes the ability to independently run only baselines.
> 
> What do you think? @kevinstephano @jacobhinkle
>We could go with either of the designs: For weekly runs, it does not make a difference since we will run all executors. For local benchmarking, users may wish to independently run the baselines?

Assuming that baselines refer to the eager mode, I don't think it's important to support just running the baselines. So, I'd prefer simplifying the benchmark design with the fusion profiler. As Kevin mentioned, it doesn't need to be done in this PR.

Priya2698:
> > I can organize the benchmarks so that this is true, but that removes the ability to independently run only baselines.
> > What do you think? @kevinstephano @jacobhinkle
> > We could go with either of the designs: For weekly runs, it does not make a difference since we will run all executors. For local benchmarking, users may wish to independently run the baselines?
> 
> Assuming that baselines refer to the eager mode, I don't think it's important to support just running the baselines. So, I'd prefer simplifying the benchmark design with the fusion profiler. As Kevin mentioned, it doesn't need to be done in this PR.

Sounds reasonable.
I'll make the change in this PR to avoid adding IO compute functions (`softmax_fwd_iobytes`) since they will be removed anyway.

Priya2698:
Options I tried to share the IObytes between benchmarks:
1. Benchmarking multiple functions in the same call is not supported by `pytest-benchmarks`.
2. Storing variables computed by one benchmark can be done using `fixtures`, however there is no clear way to guarantee that the nvfuser and baseline benchmarks for a given configuration end up in the same shard. We would need to modify the sharding logic
[here](https://github.com/AdamGleave/pytest-shard/blob/64610a08dac6b0511b6d51cf895d0e1040d162ad/pytest_shard/pytest_shard.py#L56)
(something like run different tests using the same set of parameters (size, shape) etc in the same shard). I'm skeptical of making sharding test-structure dependent since that may change with future updates or if we use any other utility to split tests, it won't work.

I am leaning towards continuing with the approach in the current PR and manually set the IOBytes for baselines.
What do you think? @naoyam @kevinstephano

naoyam:
So, the nvfuser and corresponding eager runs need to be separate benchmarks in pytest-benchmarks, and there's no (easy) way to run them together when sharded. The simplest solution is to just manually set the IO bytes. Am I right?

Priya2698:
> So, the nvfuser and corresponding eager runs need to be separate benchmarks in pytest-benchmarks, and there's no (easy) way to run them together when sharded. The simplest solution is to just manually set the IO bytes. Am I right?

Yes.

naoyam:
OK, makes sense.

naoyam (approved):
LGTM

PR #1727, opened 2024-02-06
Avoid unused smem workspace for serial grid reductions

jacobhinkle:
GridReduction can be lowered to either `gridReduce` or `serialReductionStep`. `gridReduce` requires a smem workspace in order to use multiple threads to aggregate partial sums. However, `serialReductionStep` does not coordinate among threads and has no use for a workspace. This change simply disables allocating that little bit of extra shared memory if our only grid reductions are serial, which currently only happens in split-K GEMM.

This reduces the smem allocated in a simple test from 16896 B to 16384 B (about 97%). More importantly, this makes the computation in `mma_utils::generateSharedMemoryEpilogueHeuristics()` more accurate. Tests are updated to check that this computation is accurate.

The change in `kernel.cpp` is responsible for reducing actual smem usage for split-K. The changes to `mma_utils` and `test_gpu_tensorcore.cpp` are needed for adding testing that our expected smem usage matches the actual usage.

### DIFF
--- a/csrc/kernel.cpp
+++ b/csrc/kernel.cpp
@@ -128,7 +128,11 @@ class KernelIrScanner : private IrVisitor {
   }
 
   void handle(GridReduction* grid_reduction) final {
-    summary_.has_grid_reductions = true;
+    // summary.has_grid_reductions is used to determine whether we need a
+    // reduction workspace. Serial grid reductions do not require this
+    // workspace.
+    summary_.has_grid_reductions =
+        grid_reduction->serialReductionTensor() == nullptr;
     if (grid_reduction->isAllreduce()) {
       summary_.has_cooperative_grid_reduction = true;
     }
--- a/csrc/scheduler/mma_utils.cpp
+++ b/csrc/scheduler/mma_utils.cpp
@@ -36,22 +36,14 @@ inline mma_utils::MmaDataTypes getMmaDataTypes(
   return mma_utils::MmaDataTypes{a_type, b_type, c_type};
 }
 
-std::pair<bool, bool> generateSharedMemoryEpilogueHeuristics(
+//! Return sizes of smem_a, smem_b, smem_c
+std::tuple<size_t, size_t, size_t> computeSharedMemorySizes(
     const MatMulTileOptions& gemm_tile,
     const int smem_double_buffer_stage,
-    const MmaDataTypes& data_types,
-    bool smem_a_reuse_guaranteed,
-    bool smem_b_reuse_guaranteed,
-    bool ignore_occupancy_drop) {
+    const MmaDataTypes& data_types) {
   const auto properties = at::cuda::getCurrentDeviceProperties();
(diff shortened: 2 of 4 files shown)


jacobhinkle:
cc @mmigdal-nv hopefully this will address the discrepancy in smem computation.

jacobhinkle:
!build

jacobhinkle:
Interestingly, it seems like we are currently overestimating the smem size for split-K (or underallocating?). This is on my 3090Ti:
```

MMA macro: Ampere_16_8_16
DoubleBufferOptions:
  double_buffer_smem_write: false
  double_buffer_smem_read: false
  smem_double_buffer_stage: 2
MatMulTileOptions: instruction tile [16, 8, 16], warp tile [64, 64, 32], CTA tile [128, 128, 32]
Async global mem load: false
Indexing mode: unavailable
Tile rastrization order: row-major
Grid swizzle factor: 1
Use shared memory epilogue: 0
Promote re-use of prologue shared memory: 0
Split-K factor: 2

computeExpectedSharedMemoryUsage
  params.use_smem_epilogue = 0
  params.promote_prologue_smem_reuse = 0
  smem_a = 16384
  smem_b = 16384
  smem_c = 65536
/opt/pytorch/nvfuser/test/test_gpu_tensorcore.cpp:2540: Failure
Expected equality of these values:
  estimated_smem
    Which is: 32768
  actual_smem
    Which is: 16384
```
I'm investigating now, but this check might have exposed an actual bug...

jacobhinkle:
!build --diff-bench

csrc/kernel.cpp
```diff
@@ -128,7 +128,11 @@ class KernelIrScanner : private IrVisitor {
   }
 
   void handle(GridReduction* grid_reduction) final {
-    summary_.has_grid_reductions = true;
+    // summary.has_grid_reductions is used to determine whether we need a
+    // reduction workspace. Serial grid reductions do not require this
+    // workspace.
+    summary_.has_grid_reductions =
+        grid_reduction->serialReductionTensor() == nullptr;
```
jacobhinkle, line 135: This change is what reduces the split-K smem usage by a few hundred bytes.

naoyam: I think this is fine for now, but I wonder if we should create an allocation node for the reduction work buffer as well, rather than handling it separately.

jacobhinkle: Yes, that would be ideal. Then we can finally get rid of `smem_offset`.

jacobhinkle:
!build

naoyam (approved):
LGTM

jacobhinkle:
I believe the CI failure is unrelated as it is a low tolerance match error in `thunder/tests/test_cudnn_executor.py::test_vjp_correctness_sdpa_cudnnex_manual_grad_forward_scaled_dot_product_attention_nvfuser_cuda_float16`.

Issue #1706, opened 2024-01-31
Deserialization fails during segmentation with half precision inputs.


rdspring1:
The following error occurs: `IndexError: deque::_M_range_check: __n (which is 334)>= this->size() (which is 107)`.

Reproducer script:
```
    def test_fp16_fusion_segentation(self):
        # deserialization fails during segmentation with half inputs
        inputs = [
            1e-6,
            10,
            4096,
            4096,
            torch.randn((1, 4096, 4096,), dtype=torch.bfloat16, device="cuda:0"),
            torch.randn((10, 32), dtype=torch.bfloat16, device="cuda:0"),
            torch.randn((1, 4096, 4096,), dtype=torch.bfloat16, device="cuda:0"),
            torch.randn((1, 4096, 1,), dtype=torch.bfloat16, device="cuda:0"),
            torch.randn((1, 1, 4096,), dtype=torch.bfloat16, device="cuda:0").expand(1, 4096, 4096),
        ]

        def fusion_func(fd : FusionDefinition) -> None :
            S0 = fd.define_scalar(None, dtype=DataType.Double)
            S1 = fd.define_scalar(None, dtype=DataType.Int)
            S2 = fd.define_scalar(None, dtype=DataType.Int)
            S3 = fd.define_scalar(None, dtype=DataType.Int)
            T4 = fd.define_tensor(shape=[1, -1, -1], contiguity=[None, True, True], dtype=DataType.BFloat16, is_cpu=False)
            T5 = fd.define_tensor(shape=[-1, -1], contiguity=[True, True], dtype=DataType.BFloat16, is_cpu=False)
            T6 = fd.define_tensor(shape=[1, -1, -1], contiguity=[None, True, True], dtype=DataType.BFloat16, is_cpu=False)
            T7 = fd.define_tensor(shape=[1, -1, 1], contiguity=[None, True, None], dtype=DataType.BFloat16, is_cpu=False)
            T8 = fd.define_tensor(shape=[1, -1, -1], contiguity=[None, None, True], dtype=DataType.BFloat16, is_cpu=False)
            T9 = fd.ops.cast(T6, dtype=DataType.Float)
            T10 = fd.ops.cast(T6, dtype=DataType.Float)
            T11 = fd.ops.cast(T7, dtype=DataType.Float)
            T12 = fd.ops.rsqrt(T11)
            T13 = fd.ops.cast(T12, dtype=DataType.BFloat16)
[...]

PR #1731, opened 2024-02-06
Check if all statements exist in original fusion before serializing fusion segmentation.

rdspring1:
Workaround for #1706

**Problem:** 
`SegmentCandidateFinder::findSegments` can generate new statements when finding valid sub-fusions. Therefore, `SegmentedGroup` can reference statements that do not exist in the original fusion. It is difficult to generate those statements during deserialization without merging `SegmentedGroups` in the original order.

**Solution:**
During serialization, check if all statements are present in the initial fusion passed to `SegmentedFusion`. If not, fallback to `SegmentCandidateFinder::findSegments` during deserialization.


rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

csrc/fusion_segmenter.cpp
```diff
@@ -503,23 +554,22 @@ void SegmentedFusion::deserialize(const serde::SegmentedFusion* buffer) {
...
+      complete_fusion_->unordered_exprs().size() <= buffer->num_exprs(),
       "The complete fusion has ",
-      exprs.size(),
+      complete_fusion_->unordered_exprs().size(),
       " expressions while serialization expected ",
       buffer->num_exprs(),
       " expressions.");
-
+  const std::deque<Val*>& vals = complete_fusion_->deterministic_vals();
+  const std::deque<Expr*>& exprs = complete_fusion_->deterministic_exprs();
```
jjsjann123, line 572: nitpick: wasn't `deterministic_xxx()` variants return container with the same size as with `unordered_xxx`?

Can't we just use `vals`/`exprs` in the check above?

csrc/fusion_segmenter.cpp
```diff
@@ -503,23 +554,22 @@ void SegmentedFusion::deserialize(const serde::SegmentedFusion* buffer) {
   // NOTE SchedulerEntry::proposeHeuristics can add values and expressions to
   // the fusion. We relax the constraints here because we already know the
   // proposed scheduler for each segmented group.
-  const std::deque<Val*>& vals = complete_fusion_->deterministic_vals();
-  const std::deque<Expr*>& exprs = complete_fusion_->deterministic_exprs();
   NVF_ERROR(
       complete_fusion_->vals().size() <= buffer->num_vals(),
       "The complete fusion has ",
-      vals.size(),
+      complete_fusion_->vals().size(),
```
jjsjann123, line 560: nitpick: error message seems to suggest an `equal` assert, but in reality it's a `less-than`. Might as well update that.

csrc/serde/fusion_cache.fbs
```diff
@@ -387,6 +387,7 @@ table SegmentedGroup {
 
 // The SegmentedFusion represents a segmented fusion graph that owns the segmented groups.
 table SegmentedFusion {
+  valid: bool;
```
jjsjann123, line 390: nitpick: can we actually put a comment here explaining why we have the `valid` field. I imagine it would be very confusing to folks going through this.
Maybe a link the issue 1706 would suffice.

csrc/fusion_segmenter.cpp
```diff
@@ -449,11 +450,39 @@ SegmentedFusion::SegmentedFusion(std::unique_ptr<Fusion> fusion)
...
+    const std::unordered_map<Expr*, int64_t>& exprs_to_id_map,
+    int64_t initial_vals_size,
+    int64_t initial_exprs_size) {
+  auto check_value = [&](Val* v) {
+    return vals_to_id_map.at(v) < initial_vals_size;
+  };
+  auto check_expr = [&](Expr* e) {
+    return exprs_to_id_map.at(e) < initial_exprs_size;
+  };
+  bool all_compatible_inputs =
```
jjsjann123, line 474: nitpick:?

```
bool compatible = std::all_of(...);
compatible = compatible && std::all_of(...);
...
return compatible;
```

csrc/fusion_segmenter.cpp
```diff
@@ -449,11 +450,39 @@ SegmentedFusion::SegmentedFusion(std::unique_ptr<Fusion> fusion)
...
+      std::all_of(sg->inputs().begin(), sg->inputs().end(), check_value);
+  bool all_compatible_outputs =
+      std::all_of(sg->outputs().begin(), sg->outputs().end(), check_value);
```
jjsjann123, line 477: This is assuming all intermediate `Val*` would have an entry that's smaller than the largest outputs. I guess we have incremental update so that has to be safe for now.

I also lean towards checking all vals leading to output just to be "future-proof"... not a strong opinion since it has no impact for today.

jjsjann123 (approved):
LGTM. a few nitpick here and there.

:shipit:

csrc/fusion_segmenter.cpp
rdspring1: It is but `deterministic_xxx` has to make a new container first, while `unordered_xxx` uses an already existing one.

jjsjann123: I would assume the assert is an uncommon thing.

So if we bring 571/572 up a bit, we can just use `vals.size()` and `exprs.size()` instead. i.e. we need to build the container anyway.

csrc/fusion_segmenter.cpp
rdspring1: The `input_vals` and `output_vals` are the only `Val*` in `SegmentedGroup`, so there aren't any intermediates to check for serialization. This function checks if we can serialize correctly. It doesn't guarantee that the segmented group is correct.

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

jjsjann123:
!build

jjsjann123:
!build

jjsjann123:
Looks like your last CI still grabbed the old container. I'll restarting it.

PR #1687, opened 2024-01-27
Move `aliased_io` into `AliasInfo`. 

wujingyue:
This makes `getOutputAlias` easier to use. This will eventually help #1669 represent outputs that require evaluation but are not an alias. For that, we'll need to
1. Rename `Alias*` to `Allocation*`.
2. Rename `AllocationType::PointerArithmetic` to `AllocationType::Evaluate`. 
3. When a MmaOp segment needs ATen evaluation, set its output allocation to `.type=AllocationType::Evaluate,aliased_io=nullptr`.

PR #1732, opened 2024-02-07
Rename `AliasType` -> `AllocationType`

Priya2698:
This PR is strictly for renaming `AliasType` -> `AllocationType`.
Follows PR #1687.

The next PR will add `AllocationType::Evaluate` to mark Mma outputs to be evaluated through expression evaluator.

### DIFF
--- a/csrc/fusion.h
+++ b/csrc/fusion.h
@@ -86,7 +86,7 @@ class FusionGuard {
 
 // Set the enum base to `int` so it can be safely serialized as a part of
 // serde::InputOutputAlias.
-enum class AliasType : int {
+enum class AllocationType : int {
   NoAlias,
   // For example, the tensor storing BatchNorm's running mean. The output EMA is
   // updated in place.
@@ -98,7 +98,7 @@ enum class AliasType : int {
 };
 
 struct AliasInfo {
-  AliasType type;
+  AllocationType type;
   Val* aliased_io;
   // Whether integration should hide the output from users. This is currently
   // only used for InplaceUpdate.
@@ -248,7 +248,7 @@ class Fusion : public IrContainer {
   // the input tensor to the section where output is produced. Currently,
   // aliases of type `PointerArithmetics` are marked after segmentation, but
   // those of type `InplaceUpdate` are marked in fusion definitions.
-  void aliasOutputToInput(Val* output, Val* input, AliasType type);
+  void aliasOutputToInput(Val* output, Val* input, AllocationType type);
 
   //! Returns the aliased input of a given output along with an `AliasInfo`
   //! describing how they alias. Returns <nullptr,nullptr> when `output` is not
--- a/csrc/device_lower/pass/expr_sort.cpp
+++ b/csrc/device_lower/pass/expr_sort.cpp
@@ -1516,7 +1516,7 @@ void ExprSegmentationSorter::sort() {
       std::back_inserter(non_pointer_arithmetic_outs),
       [this](Val* out) {
         return fusion_->getOutputAlias(out).type !=
-            AliasType::PointerArithmetic;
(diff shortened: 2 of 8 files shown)


csrc/fusion.h
```diff
@@ -86,7 +86,7 @@ class FusionGuard {
 
 // Set the enum base to `int` so it can be safely serialized as a part of
 // serde::InputOutputAlias.
-enum class AliasType : int {
+enum class AllocationType : int {
   NoAlias,
```
wujingyue, line 90: How about the following naming for AllocationType? 

```
New,  // allocate a new buffer
ReuseBuffer,  // reuse the buffer allocated for `aliased_io`
Evaluate,  // evaluated by ExpressionEvaluator
```

wujingyue (approved):

csrc/fusion.h
wujingyue: This wasn't resolved though. What's your thought on this?

Priya2698: Sounds reasonable when merging `PointerArithmetic` and the current `Evaluate` but some of their requirements are different. `PointerArithmetic` requires a non-null `aliased_io` so that check should be refactored away from `allocateOutput`.

wujingyue: Yes. I'd skip those checks when `aliased_io` is null. Wdyt?

Priya2698: I'd still like to have some check for that (unless you think that we will never violate this condition). Let me look at if we can move it to some other place where this alias is discovered.

wujingyue: > some check for that

Check for what? [These checks](https://github.com/NVIDIA/Fuser/blob/ea1bff3596fa263402c376c1ca720c5a5af53878/csrc/executor.cpp#L984-L990) only make sense when `aliased_io` is not null, so 
```
if (aliased_io != nullptr) {
  do these checks;
}
```
sounds pretty reasonable to me.

Priya2698: The check for `aliased_io` itself to be non-null when it is of PointerArithmetic type. The rest of the checks can be reorganized easily. 
From a quick look at the aliasing files though, it does not seem we will run into the issue of null `aliased_io` pointers in type PointerArithmetic though.

wujingyue: > The check for aliased_io itself to be non-null when it is of PointerArithmetic type.

Get it! 

> From a quick look at the aliasing files though, it does not seem we will run into the issue of null aliased_io pointers in type PointerArithmetic though.

That's right. That's why I think it's pretty safe to use whether `aliased_io==nullptr` to distinguish between PointerArithmetic and Evaluate. This way, we can remove `PointerArithmetic` whose logic is identical to `Evaluate` except the alias check. Wdyt?

Priya2698: Yes, that works nicely. My only concern was regarding `PointerArithmetic` somehow ending up with null `aliased_io` which is unlikely to be the case!

wujingyue: From @Priya2698:

> I actually prefer NoAlias and InplaceUpdate to New and ReuseBuffer. The former set seems more explicit to read.
> Wdyt?

SGTM -- combining the other two types is more important. 

I suggested New and ReuseBuffer because they describe an allocation, e.g., allocate a new buffer for this output vs reuse the existing buffer allocated for `aliased_io`.  AllocationType::NoAlias and AllocationType::InplaceUpdate sound a bit weird to me -- what does an inplace-update allocation mean?

Priya2698: That is true. We have Alias and Allocation terminology being merged now so they don't fit well together.
I don't have a strong preference so we can move to `New` and `ReuseBuffer`. I'll add comments to explain the types instead.

PR #1733, opened 2024-02-07
Make `MmaOp::evaluate` return output of the same dtype as `MmaOp`

Priya2698:
`MmaOp::evaluate` should return output with the same dtype as the `MmaOp->out()` for consistency.

### DIFF
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -2111,7 +2111,11 @@ std::vector<PolymorphicValue> MmaOp::evaluate(
 
   // After removing the broadcast dimensions, the format should be
   // [M, K] x [K, N] compatible with aten::matmul format.
-  return {in_a.matmul(in_b)};
+  auto output = in_a.matmul(in_b);
+  if (tv_a->getDataType() != out()->getDataType().value()) {
+    output = output.to(data_type_to_aten(out()->getDataType().value()));
+  }
+  return {output};
 }
 
 NVFUSER_DEFINE_CLONE_AND_CREATE(MmaOp)
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -2115,6 +2115,11 @@ std::vector<PolymorphicValue> MmaOp::evaluate(
 
   // ATen preserves the input dtype whereas MmaOP generates float outputs.
   // Cast to the dtype of the MmaOp output for consistency.
+  // NOTE: MmaOp returns the float output, whereas in the evaluate method,
+  //      we are casting from float -> input_dtype -> float. This will lead
+  //      to loss of precision.
+  //      MmaOp::evaluate should be modified to effectively handle cast(MmaOp(H,
+  //      H), H) This will avoid the above cast chain and precision issue.
   if (tv_a->getDataType() != out()->getDataType().value()) {
     output = output.to(data_type_to_aten(out()->getDataType().value()));
   }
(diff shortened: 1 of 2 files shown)


test/test_evaluator.cpp
```diff
@@ -703,5 +703,8 @@ TEST_F(ExprEvalTest, MmaOp) {
   evaluator.bind(tv1, in_b);
   at::Tensor out = evaluator.evaluate(tv2).as<at::Tensor>();
   NVF_CHECK(out_ref.equal(out));
+
+  auto ref_dtype = data_type_to_aten(tv2->getDataType().value());
+  NVF_CHECK(out.dtype() == ref_dtype);
```
wujingyue, line 708:
```suggestion
  EXPECT_EQ(out.dtype(), ref_dtype);
```

wujingyue: Even better, I think we can do 

```
EXPECT_TRUE(at::allclose(out, out_ref));
```

test/test_evaluator.cpp
```diff
@@ -696,12 +696,13 @@ TEST_F(ExprEvalTest, MmaOp) {
 
   at::Tensor in_a = at::ones(a_shape, at::kHalf).cuda();
   at::Tensor in_b = at::ones(b_shape, at::kHalf).cuda();
-  at::Tensor out_ref = k * at::ones(out_shape, at::kHalf).cuda();
+  auto out_dtype = data_type_to_aten(tv2->getDataType().value());
+  at::Tensor out_ref = k * at::ones(out_shape, out_dtype).cuda();
```
wujingyue, line 700: Why did you make this change? The old code looks better. In general, make the reference implementation as straightforward as possible, and avoid reusing stuff from the actual implementation. This is because the part you reuse could be buggy.

test/test_evaluator.cpp
```diff
@@ -696,12 +696,13 @@ TEST_F(ExprEvalTest, MmaOp) {
 
   at::Tensor in_a = at::ones(a_shape, at::kHalf).cuda();
   at::Tensor in_b = at::ones(b_shape, at::kHalf).cuda();
-  at::Tensor out_ref = k * at::ones(out_shape, at::kHalf).cuda();
```
wujingyue, line 699: How about `at::full`?

csrc/ir/nodes.cpp
```diff
@@ -2111,7 +2111,14 @@ std::vector<PolymorphicValue> MmaOp::evaluate(
 
   // After removing the broadcast dimensions, the format should be
   // [M, K] x [K, N] compatible with aten::matmul format.
-  return {in_a.matmul(in_b)};
+  auto output = in_a.matmul(in_b);
+
+  // ATen preserves the input dtype whereas MmaOP generates float outputs.
+  // Cast to the dtype of the MmaOp output for consistency.
+  if (tv_a->getDataType() != out()->getDataType().value()) {
+    output = output.to(data_type_to_aten(out()->getDataType().value()));
```
wujingyue, line 2119: Note that this is still a wrong implementation. I believe `at::matmul (bfloat->bfloat) ==> at::cast(bfloat->float)` is less precise than `MmaOp (bfloat->float)`. `MmaOp (bfloat->float)` accumulates in float and never casts it back to the lower precision `bfloat`. However, `at::matmul ==> at::cast` (as in this implementation) introduces an **extra** float->bfloat->float roundtrip and therefore loses some precision. 

I'm fine with this code change as is because it's a strict improvement -- MmaOp::evaluate didn't even get the output type right. However, I'd put this warning in code comment and fix it soonish.

Priya2698: Right. There will some precision lost. Will add a comment.
Since `at::matmul` does not allow HH->S matmuls, the nvFuser matmul API will also most likely not support it (@protonu is this assumption valid?).

Once the evaluate method here is updated to handle `MmaOp + castOp` (which is what we should get from the frontend), evaluate can simply return the original output and without any such round trip casting.

wujingyue (approved):

Issue #1597, opened 2024-01-08
Improve performance on RoPE (and code around it). 


wujingyue:
I'm separating this from #1502. While we can get rid of `cat` in some cases, improving nvFuser's codegen for `slice` and `cat` still benefits the RoPE module and the QKV split around it.

wujingyue:
```
$ NVFUSER_DUMP=segmented_fusion pytest qkv_split_rope.py -s > stdout 2>&1 

$ grep 'g{(' stdout 
g{(no_op)
g{(pointwise)
g{(no_op)
g{(no_op)
g{(pointwise)
g{(pointwise)
g{(pointwise)
g{(pointwise)
```

Currently, the fusion is segmented into eight kernels. The alias stuff made three of them no-op but the rest five contain non-meta ops like cast, neg, mul, and add. I think we'll have to fix our segmentation around slicing.

wujingyue:
nvFuser enforces resizing ops like Slice and Cat to be fusion inputs: https://github.com/NVIDIA/Fuser/blob/cb4bd59e1a75e1c6bcaf19ee18a6eb53bf0d8de3/csrc/scheduler/registry_utils.cpp#L193. That explains the many segments. (However, I'd expect 4 not 5)

In another thread, @naoyam explained why resizing ops have to be inputs at this moment, which I copied below: 

By default, we only allow those ops with inputs being fusion inputs. The reason is that in our current parallelization scheme, an input to a silce op and its output, when parallelized, may require communications between threads, which can be between threads of different thread blocks, which in turn requires grid synchronizations. For example:

```
TensorView* t0 = makeConcreteTensor({128});
auto t1 = set(t0);
auto t2 = t1[1:-2];

t2->split(0, 32);
propagateTransformationFrom(t2);

t2->axis(0)->parallelize(BIDx);
t2->axis(1)->parallelize(TIDx);
propagateParallelizationFrom(t2);
```

After this the `t1` and `t2` tensors would look like:

```
t1: [BIDx(128/32), TIDx(32)]
t2: [BIDx(126/32), TIDx(32)]
```

[...]

PR #1739, opened 2024-02-07
Add benchmarks for RoPE.

wujingyue:
This PR adds two implementations of the RoPE module and benchmarks them for #1597.

`rope_with_cat_fusion` mimics the Hugging Face implementation. `rope_without_cat_fusion` implements an idea from @nikitaved to avoid concatenation. Even though it looks difficult for the compiler to do it all automatically, it's still useful to keep a record of the idea. 

As a side change, I made `fd.define_tensor` to accept empty contiguity.

### DIFF
--- a/csrc/python_frontend/python_bindings.cpp
+++ b/csrc/python_frontend/python_bindings.cpp
@@ -738,7 +738,7 @@ void initNvFuserPythonBindings(PyObject* module) {
                 if (dim_size == 1) {
                   contiguity.emplace_back(std::nullopt);
                 } else {
-                  contiguity.emplace_back(true);
+                  contiguity.emplace_back(false);
                 }
               }
             }
--- a/csrc/python_frontend/python_bindings.cpp
+++ b/csrc/python_frontend/python_bindings.cpp
@@ -713,10 +713,10 @@ void initNvFuserPythonBindings(PyObject* module) {
       .def(
           "define_tensor",
           [](FusionDefinition& self,
-             std::vector<int64_t>& shape,
-             std::vector<std::optional<bool>>& contiguity,
-             PrimDataType dtype = DataType::Float,
-             bool is_cpu = false,
+             const std::vector<int64_t>& shape,
+             std::vector<std::optional<bool>> contiguity = {},
+             const PrimDataType dtype = DataType::Float,
+             const bool is_cpu = false,
              std::vector<int64_t> stride_order = {}) -> Tensor {
             FUSER_PERF_SCOPE("FusionDefinition.define_tensor (default)");
             NVF_CHECK(
@@ -733,6 +733,16 @@ void initNvFuserPythonBindings(PyObject* module) {
                   " was neither symbolic(-1), zero_element(0), broadcast(1), or static(>1).");
             }
 
+            if (contiguity.empty()) {
+              for (const auto dim_size : shape) {
+                if (dim_size == 1) {
+                  contiguity.push_back(std::nullopt);
+                } else {
+                  contiguity.emplace_back(true);
+                }
+              }
+            }
+
             Tensor out = self.defineTensor(shape.size());
             self.defineRecord(new TensorRecord(
                 {self.recordingState(out())},
@@ -745,7 +755,7 @@ void initNvFuserPythonBindings(PyObject* module) {
             return out;
           },
           py::arg("shape"),
-          py::arg("contiguity"),
+          py::arg("contiguity") = py::list(),
           py::arg("dtype") = DataType::Float,
           py::arg("is_cpu") = false,
           py::arg("stride_order") = py::list(),
--- a/csrc/python_frontend/python_bindings.cpp
+++ b/csrc/python_frontend/python_bindings.cpp
@@ -736,7 +736,7 @@ void initNvFuserPythonBindings(PyObject* module) {
             if (contiguity.empty()) {
               for (const auto dim_size : shape) {
                 if (dim_size == 1) {
-                  contiguity.push_back(std::nullopt);
+                  contiguity.emplace_back(std::nullopt);
                 } else {
                   contiguity.emplace_back(true);
                 }
--- a/python_benchmarks/test_rope.py
+++ b/python_benchmarks/test_rope.py
@@ -1,0 +1,94 @@
+import pytest
+from nvfuser import FusionDefinition, DataType
+from .core import run_benchmark, clear_cuda_cache
+import torch
+
+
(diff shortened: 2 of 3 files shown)


wujingyue:
!build

csrc/python_frontend/python_bindings.cpp
```diff
@@ -733,6 +733,16 @@ void initNvFuserPythonBindings(PyObject* module) {
                   " was neither symbolic(-1), zero_element(0), broadcast(1), or static(>1).");
             }
 
+            if (contiguity.empty()) {
+              for (const auto dim_size : shape) {
+                if (dim_size == 1) {
+                  contiguity.emplace_back(std::nullopt);
```
jjsjann123, line 739: Interestingly surprised...
I saw the clang-tidy failure earlier and I somehow never associate a `std::nullopt` to call a constructor :face_with_head_bandage:

csrc/python_frontend/python_bindings.cpp
```diff
@@ -713,10 +713,10 @@ void initNvFuserPythonBindings(PyObject* module) {
       .def(
           "define_tensor",
           [](FusionDefinition& self,
-             std::vector<int64_t>& shape,
-             std::vector<std::optional<bool>>& contiguity,
-             PrimDataType dtype = DataType::Float,
-             bool is_cpu = false,
+             const std::vector<int64_t>& shape,
+             std::vector<std::optional<bool>> contiguity = {},
+             const PrimDataType dtype = DataType::Float,
```
jjsjann123, line 718: nitpick: I guess this is not an exposed c++ api so it's likely no big deal, but const-qualified value argument sounds like a bad practice. Feel free to discard my personal style~~

python_benchmarks/test_rope.py
```diff
@@ -0,0 +1,184 @@
...
+# Mimic the Hugging Face implementation:
+# https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py#L216
+def rope_with_cat_fusion(
+    fd: FusionDefinition,
+    batch_size: int,
+    seq_len: int,
+    num_heads: int,
+    features_per_head: int,
+) -> None:
+    q = fd.define_tensor(
```
jjsjann123, line 16: FYI, you can also do `fd.from_pytorch(pytorch_tensor)`, which does require you to pass in each tensor args... But that also sorted out the stride_order thing for you, which admittedly doesn't apply here.

jjsjann123 (approved):
Thanks for the API change!

csrc/python_frontend/python_bindings.cpp
```diff
@@ -733,6 +733,16 @@ void initNvFuserPythonBindings(PyObject* module) {
                   " was neither symbolic(-1), zero_element(0), broadcast(1), or static(>1).");
             }
 
+            if (contiguity.empty()) {
+              for (const auto dim_size : shape) {
+                if (dim_size == 1) {
+                  contiguity.emplace_back(std::nullopt);
+                } else {
+                  contiguity.emplace_back(true);
```
jjsjann123, line 741: nitpick: should we default to `false` though....

python_benchmarks/test_rope.py
```diff
@@ -0,0 +1,184 @@
...
+
+    cos = fd.ops.broadcast_in_dim(
+        cos, shape=[1, 1, seq_len, features_per_head], broadcast_dims=[2, 3]
+    )
+    sin = fd.ops.broadcast_in_dim(
+        sin, shape=[1, 1, seq_len, features_per_head], broadcast_dims=[2, 3]
+    )
+
+    out = q * cos + q_rotated * sin
+    out = fd.ops.cast(out, DataType.BFloat16)
```
Priya2698, line 58: Is this needed? All the inputs are `BFloat16`.

jjsjann123: we don't do fp16 math. I imagine without the explicit cast, we'll be generating fp32 output and that's gonna be a surprise.

python_benchmarks/test_rope.py
```diff
@@ -0,0 +1,184 @@
...
+    )
+
+    cos_sin_matrix = fd.ops.broadcast_in_dim(
+        cos_sin_matrix,
+        shape=[batch_size, num_heads, seq_len, 2, features_per_head // 2, 2],
+        broadcast_dims=[2, 3, 4, 5],
+    )
+
+    out = fd.ops.sum(q * cos_sin_matrix, [-1])
+    out = fd.ops.cast(out, DataType.BFloat16)
```
Priya2698, line 116: Is this cast necessary -- same as for `rope_with_cat` above?

wujingyue: Yes. nvFuser does implicit type promotion. Without this, the output type would be Float.

Priya2698 (approved):
LGTM.

python_benchmarks/test_rope.py
wujingyue: I see -- That's probably why we haven't been annoyed enough by having to specify the contiguity.

python_benchmarks/test_rope.py
Priya2698: Oh right. OOC, is there a difference performance-wise between explicitly casting all the inputs to float vs letting nvFuser implicitly cast inputs? (Context: we explicitly cast inputs to float in other benchmark fusions)

csrc/python_frontend/python_bindings.cpp
wujingyue: Good point -- that'll be consistent with https://github.com/NVIDIA/Fuser/blob/5cb0bb4a37e1950cf1a4dda809aa0b1744bc178a/csrc/ir/nodes.cpp#L3267.

csrc/python_frontend/python_bindings.cpp
wujingyue: > const-qualified value argument sounds like a bad practice

You are right about that for function **declarations**. For function **definitions**, const-qualified value arguments are fine: https://abseil.io/tips/109. 

However, this case is a bit tricky because it's both a declaration and a definition, so 🤷

wujingyue:
!build

python_benchmarks/test_rope.py
wujingyue: I have no opinions between the two at this moment. In nvFuser, implicit casts are done where they are needed (e.g. for mul but not for reshape), so using implicit cast can be more performant in some situations. On the other hand, I doubt everyone on the team knows where implicit casts are inserted, so using explicit casts probably makes code easier to read.

PR #1740, opened 2024-02-08
io_alias_ const update

jjsjann123:
1. changing the key type from `Val*` to `const Val*` in `std::unordered_map<const Val*, AliasInfo> Fusion::io_alias_`;

2. changing the signature from `Val*` to `const Val*` in `const AliasInfo& getOutputAlias(const Val* output) const;`

### DIFF
--- a/csrc/fusion.cpp
+++ b/csrc/fusion.cpp
@@ -814,7 +814,7 @@ void Fusion::aliasOutputToInput(
   }
 }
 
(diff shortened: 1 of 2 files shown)


jjsjann123:
!build

wujingyue:
Why? The key of a map can usually stay const unless you want to `for each entry`, and even that can usually be avoided by `for each fusion output: getOutputAlias`.

wujingyue (approved):

PR #1741, opened 2024-02-08
Rename axes= to dims= in frontend

jacobhinkle:
Currently we accept `axes=` for some ops like `fd.ops.sum` and `dims=` for others like `fd.ops.squeeze`.

This is a small attempt to make the frontend arguments more consistent. This change renames the `axis=` kwarg to `dim=` and the same for `axes=` -> `dims=`.

I think we're free to set our own convention, but for reference:
- PyTorch uses `dim=` in most places and accepts either a single dim or multiple using that same argument name, where applicable.
- Numpy uses `axis=` and, like PyTorch, accepts a list where applicable.
- `jax.lax` uses `dimensions=`


jacobhinkle:
!build

jjsjann123 (approved):
:clap:

version.txt
```diff
@@ -1 +1 @@
-0.1.5
```
jjsjann123, line 1: Thanks for remembering to do this. :yum:

jacobhinkle:
!build

version.txt
jjsjann123: We need to automate this part... or at least have a CI check to assert when we do break API... something maybe @xwang233 would be interested?!

jacobhinkle: Oh that's a good idea.

xwang233: I'll check pytorch's CI implementation on detecting BC breaking changes. 😉

PR #1743, opened 2024-02-08
Matmul default scheduling [1]

Priya2698:
Initial PR for Issue #1669.

1. Adds a `EnableOption::MatmulExprEval` to turn on expression evaluation for matmul while the API in under progress.
2. Currently, we only evaluate the `MmaOp`. The next PRs will amend this to _look ahead_ and evaluate Mma + Cast, which is what we should see in the fusion definitions. See discussion [here](https://github.com/NVIDIA/Fuser/issues/1669#issuecomment-1918383315). In the absence of this we may have casts such as `bfloat->float->bfloat.`

### DIFF
--- a/csrc/device_lower/pass/expr_sort.cpp
+++ b/csrc/device_lower/pass/expr_sort.cpp
@@ -1516,7 +1516,9 @@ void ExprSegmentationSorter::sort() {
       std::back_inserter(non_pointer_arithmetic_outs),
       [this](Val* out) {
         return fusion_->getOutputAlias(out).type !=
-            AllocationType::PointerArithmetic;
+            AllocationType::PointerArithmetic &&
+            fusion_->getOutputAlias(out).type !=
+            AllocationType::Evaluate;
       });
 
   // Not putting the exprs between fusion inputs and allKnownVals() here
--- a/csrc/device_lower/pass/expr_sort.cpp
+++ b/csrc/device_lower/pass/expr_sort.cpp
@@ -1506,14 +1506,15 @@ void ExprSegmentationSorter::sort() {
   // Need this for initialization of the DAG that is processed
   std::unordered_map<Expr*, ExprGroup*> expr2group;
 
-  // We skip generating code that computes only pointer-arithmetic outputs.
+  // We skip generating code that computes only pointer-arithmetic outputs
+  // or any output marked for expression evaluator (AllocationType::Evaluate).
   // Those outputs will be computed by ExpressionEvaluator.
-  std::vector<Val*> non_pointer_arithmetic_outs;
-  non_pointer_arithmetic_outs.reserve(fusion_->outputs().size());
+  std::vector<Val*> non_expr_evaluator_outs;
+  non_expr_evaluator_outs.reserve(fusion_->outputs().size());
   std::copy_if(
       fusion_->outputs().begin(),
       fusion_->outputs().end(),
-      std::back_inserter(non_pointer_arithmetic_outs),
+      std::back_inserter(non_expr_evaluator_outs),
       [this](Val* out) {
         return (fusion_->getOutputAlias(out).type !=
             AllocationType::PointerArithmetic &&
@@ -1524,7 +1525,7 @@ void ExprSegmentationSorter::sort() {
   // Not putting the exprs between fusion inputs and allKnownVals() here
   // because they are computed using the expr evaluator.
   auto all_exprs = StmtSort::getExprsBetween(
-      GpuLower::current()->allKnownVals(), non_pointer_arithmetic_outs);
+      GpuLower::current()->allKnownVals(), non_expr_evaluator_outs);
 
   // Figure out all the values used as inputs to the expressions we're sorting
   // (to find terminating expressions). There could be branches of expressions
--- a/csrc/device_lower/pass/expr_sort.cpp
+++ b/csrc/device_lower/pass/expr_sort.cpp
@@ -1509,12 +1509,12 @@ void ExprSegmentationSorter::sort() {
   // We skip generating code that computes only pointer-arithmetic outputs
   // or any output marked for expression evaluator (AllocationType::Evaluate).
   // Those outputs will be computed by ExpressionEvaluator.
-  std::vector<Val*> non_expr_evaluator_outs;
-  non_expr_evaluator_outs.reserve(fusion_->outputs().size());
+  std::vector<Val*> outs_requiring_codegen;
+  outs_requiring_codegen.reserve(fusion_->outputs().size());
   std::copy_if(
       fusion_->outputs().begin(),
       fusion_->outputs().end(),
-      std::back_inserter(non_expr_evaluator_outs),
+      std::back_inserter(outs_requiring_codegen),
       [this](Val* out) {
         return (
             fusion_->getOutputAlias(out).type !=
@@ -1525,7 +1525,7 @@ void ExprSegmentationSorter::sort() {
   // Not putting the exprs between fusion inputs and allKnownVals() here
   // because they are computed using the expr evaluator.
   auto all_exprs = StmtSort::getExprsBetween(
-      GpuLower::current()->allKnownVals(), non_expr_evaluator_outs);
+      GpuLower::current()->allKnownVals(), outs_requiring_codegen);
 
   // Figure out all the values used as inputs to the expressions we're sorting
   // (to find terminating expressions). There could be branches of expressions
--- a/csrc/device_lower/pass/expr_sort.cpp
+++ b/csrc/device_lower/pass/expr_sort.cpp
@@ -1516,10 +1516,10 @@ void ExprSegmentationSorter::sort() {
       fusion_->outputs().end(),
       std::back_inserter(non_expr_evaluator_outs),
       [this](Val* out) {
-        return (fusion_->getOutputAlias(out).type !=
-            AllocationType::PointerArithmetic &&
+        return (
             fusion_->getOutputAlias(out).type !=
-            AllocationType::Evaluate);
+                AllocationType::PointerArithmetic &&
+            fusion_->getOutputAlias(out).type != AllocationType::Evaluate);
       });
 
   // Not putting the exprs between fusion inputs and allKnownVals() here
--- a/csrc/device_lower/pass/expr_sort.cpp
+++ b/csrc/device_lower/pass/expr_sort.cpp
@@ -1515,10 +1515,10 @@ void ExprSegmentationSorter::sort() {
       fusion_->outputs().end(),
       std::back_inserter(non_pointer_arithmetic_outs),
       [this](Val* out) {
-        return fusion_->getOutputAlias(out).type !=
+        return (fusion_->getOutputAlias(out).type !=
             AllocationType::PointerArithmetic &&
             fusion_->getOutputAlias(out).type !=
-            AllocationType::Evaluate;
+            AllocationType::Evaluate);
       });
 
   // Not putting the exprs between fusion inputs and allKnownVals() here
--- a/csrc/fusion.h
+++ b/csrc/fusion.h
@@ -96,7 +96,7 @@ enum class AllocationType : int {
   // cheaply compute the output tensor.
   PointerArithmetic,
   // To evaluate outputs which are not aliases.
-  // TODO: This can be potentially merged with PointerArithmetic.
+  // TODO: This will merged with PointerArithmetic in the next PR.
   //       PointerArithmetic requires aliased_io != nullptr while for evaluating
   //       a non-aliased output, aliased_io = nullptr. So keeping them separate
   //       initially.
--- a/csrc/fusion.h
+++ b/csrc/fusion.h
@@ -95,6 +95,11 @@ enum class AllocationType : int {
   // input.  In this case, we use `ExpressionEvaluator` (instead of a kernel) to
   // cheaply compute the output tensor.
   PointerArithmetic,
+  // To evaluate outputs which are not aliases.
+  // TODO: This can be potentially merged with PointerArithmetic.
+  //       PointerArithmetic requires aliased_io != nullptr while for evaluating
+  //       a non-aliased output, aliased_io = nullptr. So keeping them separate initially.
+  Evaluate,
 };
 
 struct AliasInfo {
@@ -255,6 +260,10 @@ class Fusion : public IrContainer {
   //! aliased.
   const AliasInfo& getOutputAlias(Val* output) const;
 
+  // Marks an output (by adding to io_alias_ map) to be evaluated through 
+  // expression evaluator.
+  void markOutputForEvaluation(Val* output, const AllocationType type);
+  
   // mark input at index to be permuted by permutation
   void setPermutationOnInput(int index, std::vector<int64_t> permutation) {
     permuted_input_map_.insert({index, permutation});
--- a/csrc/fusion.h
+++ b/csrc/fusion.h
@@ -260,9 +260,9 @@ class Fusion : public IrContainer {
   //! aliased.
   const AliasInfo& getOutputAlias(Val* output) const;
 
-  // Marks an output (by adding to io_alias_ map) to be evaluated through 
+  // Marks a non-aliased output to be evaluated through 
   // expression evaluator.
-  void markOutputForEvaluation(Val* output, const AllocationType type);
+  void markOutputForEvaluation(Val* output);
   
   // mark input at index to be permuted by permutation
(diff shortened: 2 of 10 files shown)


csrc/scheduler/matmul.cpp
```diff
@@ -711,6 +711,17 @@ void scheduleMatmul(Fusion* fusion, const MatmulParams& params) {
       mma_ops.size() == 1,
       "scheduleMatmul supports fusion with single mma op in definition, got ",
       mma_ops.size());
+
+  // Skip scheduling if Matmul will be expression evaluated.
+  if (isOptionEnabled(EnableOption::MatmulExprEval)) {
```
naoyam, line 716: Does this mean any matmul pattern including mma + epilogue will be handled by the expression evaluator? Shouldn't it only take care of the mma part?

wujingyue: There are two problems for taking care of just the mma: 
1. at::matmul doesn't do HH->S. We could plug in another backend that supports HH->S for EE though. 
2. MMA is never alone in these GPT models (e.g. LLaMA). It's always part of a linear layer or an SDPA. nvFuser doesn't do SDPA well and we will have to offload it to another executor for quite some time, so scratch that. A linear layer however comes with this MMA->BiasAdd->Relu pattern. In order for its performance to be on par with framework-not-giving-nvFuser-the-matmul, we have to execute MMA+epilogue in one call. 

Wdyt, @naoyam? Extending EE to do MMA+epilogue is the most obvious way to me to solve the above problems. But I could definitely be wrong.

naoyam: > Extending EE to do MMA+epilogue

How would it do that? Does Aten support matmul with some epilogue op?

wujingyue: Oh yeah, it's called `at::addmm`. I realized I was wrong about Relu. [torch.nn.Linear](https://pytorch.org/docs/stable/generated/torch.nn.Linear.html) doesn't do Relu, so the pattern would be matmul+biasadd which is what `at::addmm` does.

naoyam: So, how would you handle matmul+epilogue patterns that are accepted by the nvFuser matmul scheduler but there's no corresponding aten version? Would that end up doing mma and the epilogue op separately?

wujingyue: > Would that end up doing mma and the epilogue op separately?

Yes. However, we should probably change MatmulScheduler::canSchedule to avoid pulling in any non-fusible epilogue in the first place (so it has a chance to be fused with the following segment). Although the exact heuristics are unclear to me, I imagine MatmulScheduler::canSchedule will first decide whether to do matmul through codegen or EE, and then pull in as many epilogue ops as **that** backend (i.e. codegen or EE) can fuse.

naoyam: That makes sense. Or, maybe we might want to have a separate scheduler for EE? We try the native matmul scheduler first, and then the EE matmul scheduler, and then the other schedulers. No particular preference but just my two cents.

wujingyue: Yes, that's a certainly great option to consider. It makes things more composable at a risk of being harder to share logics with MatmulScheduler. I hope the preference will be more obvious when we know what the heuristics look like!

csrc/device_lower/pass/expr_sort.cpp
```diff
@@ -1506,23 +1506,26 @@ void ExprSegmentationSorter::sort() {
   // Need this for initialization of the DAG that is processed
   std::unordered_map<Expr*, ExprGroup*> expr2group;
 
-  // We skip generating code that computes only pointer-arithmetic outputs.
+  // We skip generating code that computes only pointer-arithmetic outputs
+  // or any output marked for expression evaluator (AllocationType::Evaluate).
   // Those outputs will be computed by ExpressionEvaluator.
-  std::vector<Val*> non_pointer_arithmetic_outs;
-  non_pointer_arithmetic_outs.reserve(fusion_->outputs().size());
+  std::vector<Val*> non_expr_evaluator_outs;
```
wujingyue, line 1512:
```suggestion
  std::vector<Val*> outs_requiring_codegen;
```

to be more concise and understandable in the context of device lowering.

test/test_matmul_default.cpp
```diff
@@ -0,0 +1,115 @@
...
+#include <test/validator.h>
+
+namespace nvfuser {
+
+class MatmulDefaultTest : public NVFuserTest {};
+
+TEST(MatmulDefaultTest, SingleMmaOp) {
+  auto fusion = std::make_unique<Fusion>();
+  FusionGuard fg(fusion.get());
+  EnableOptionsGuard::getCurOptions().set(EnableOption::MatmulExprEval);
```
wujingyue, line 26: You missed saving and restoring the previous state. See https://github.com/NVIDIA/Fuser/blob/72198a74661bc7d0c4e6185a5b302f9128ff978b/test/test_alias.cpp#L1113-L1114 and https://en.cppreference.com/w/cpp/language/raii.

wujingyue (commented):
Nice work! The PR currently shows as a draft. I'll hold off my reviews when it's ready.

csrc/scheduler/matmul.cpp
Priya2698: Adding on to @wujingyue's comments, the plan for the next PRs is:

1. Support Mma + Cast -> avoid roundtrip casting (`half-> float -> half`) by checking `mma_out->uses()` and if it is either a castOp or pointwise ops with inputs of the same type (`half`), then skip casting back the output of `at::matmul`. This will not execute `matmul + bias` in a single call.
2. Handle common epilogue fusions: We will need to pattern match and evaluate within the MmaOp accordingly. `test_matmul_scheduler.cpp` currently has a few cases that I will start with: `mma + bias`, `mma + bias + relu/gelu`, `mma + relu`.
3. Epilogue fusions that are not be supported: They can still be computed through EE but should ideally not be plumbed down to the matmul scheduler.

wujingyue: Do 2 only when it's needed. You should double check this but I think Llama for example uses linear with bias off and none of the linear layers in our benchmarks do relu or gelu

Priya2698: Llama2 has `bias=False` in the linear layers but some GPT configs have `bias=True`.

csrc/fusion.h
```diff
@@ -95,6 +95,12 @@ enum class AllocationType : int {
   // input.  In this case, we use `ExpressionEvaluator` (instead of a kernel) to
   // cheaply compute the output tensor.
   PointerArithmetic,
+  // To evaluate outputs which are not aliases.
+  // TODO: This can be potentially merged with PointerArithmetic.
```
wujingyue, line 99:
```suggestion
  // TODO: This will be merged with PointerArithmetic in the next PR.
```

I'd clean this up sooner than later. Personally, I'd do this in the next PR, so we don't keep a easy-to-fix tech debt for too long in the codebase. 

See also: https://github.com/NVIDIA/Fuser/pull/1732#discussion_r1480853889

csrc/scheduler/matmul.cpp
```diff
@@ -711,6 +711,17 @@ void scheduleMatmul(Fusion* fusion, const MatmulParams& params) {
...
+    NVF_CHECK(fusion->outputs().size() == 1)
+    fusion->markOutputForEvaluation(fusion->outputs()[0]);
+    scheduler_debug_utils::log(
+        __FUNCTION__,
```
wujingyue, line 720: Nice! This is an easy way to add a prefix to distinguish debug messages from different schedulers. 

Nit: Would `__FILE__` be better? We might want to move this call around as part of a refactoring for example. If it's moved into a function whose name doesn't contain matmul and/or whose name is not unique across the nvFuser code base, it's going to be hard to tell which file this message comes from. To that extent, I think `__FILE__` is slightly better as a way to locate the source of a message.

wujingyue (commented):
LGTM with some comments to be resolved.

csrc/scheduler/matmul.cpp
Priya2698: Good point: Will change it to `__FILE__ : __LINE__`.

wujingyue: That's even better. FWIW, for the future, we could create a macro around `scheduler_debug_utils::log` so we won't have to type `__FILE__` and `__LINE__` for convenience.

test/test_matmul_default.cpp
```diff
@@ -0,0 +1,121 @@
...
+#include <mma_type.h>
+#include <ops/all_ops.h>
+#include <scheduler/all_schedulers.h>
+#include <scheduler/mma_utils.h>
+#include <test/utils.h>
+#include <test/validator.h>
+
+namespace nvfuser {
+
+class MatmulDefaultTest : public NVFuserTest {};
```
jacobhinkle, line 21: Nit: maybe we could call these `MatmulATenEvaluationTest` or something a bit more clear.

test/test_matmul_default.cpp
```diff
@@ -0,0 +1,121 @@
...
+  fusion->addInput(tv0);
+  fusion->addInput(tv1);
+  fusion->addOutput(tv2);
+
+  at::Tensor t0 = at::ones(a_shape, at::kHalf).cuda();
+  at::Tensor t1 = at::ones(b_shape, at::kHalf).cuda();
+  at::Tensor out_ref = at::full(out_shape, k, at::kFloat).cuda();
+
+  FusionExecutorCache fec(std::move(fusion));
+  auto out = fec.runFusionWithInputs({t0, t1});
```
jacobhinkle, line 48: Is there a way for us to validate that this was indeed set up to evaluate with ATen? It might be hard to evaluate _directly_, but we could check the executor using `fec.getMostRecentKernelRuntime()->executors()`. We can check that that is size one and that the executor's fusion has the expected `io_alias_` entry.

csrc/fusion.h
jacobhinkle: Thanks for this note. On first pass I was wondering what is the difference since both signify that the output will be evaluated with `ExpressionEvaluator`.

test/test_matmul_default.cpp
wujingyue:
```suggestion
using MatmulDefaultTest = NVFuserTest;
```
to be concise, regardless of which name you and Jacob agree on.

test/test_matmul_default.cpp
```diff
@@ -0,0 +1,121 @@
...
+  int64_t m = 2, k = 3, n = 4;
+  std::vector<int64_t> a_shape{m, k}, b_shape{k, n}, out_shape{m, n};
+
+  auto tv0 = makeConcreteTensor(a_shape, DataType::Half);
+  auto tv1 = makeConcreteTensor(b_shape, DataType::Half);
+  auto tv0b = broadcast(tv0, {false, false, true}); // [M, K, 1]
+  auto tv1b = broadcast(tv1, {true, false, false}); // [1, K, N]
+  auto tv2 = fusedMultiplySum(tv0b, tv1b, {1});
+  auto tv3 = castOp(DataType::Half, tv2);
+  auto tv4 = makeConcreteTensor({m}, DataType::Half);
```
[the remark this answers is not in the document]
Priya2698: You're right. Bias should be cast instead.

drzejan2:
Thanks for preparing this change, they look good to me after caught up to the latest discussions around matmul scheduler.

test/test_matmul_aten_evaluation.cpp
```diff
@@ -0,0 +1,145 @@
...
+  at::Tensor t0 = at::ones(a_shape, at::kHalf).cuda();
+  at::Tensor t1 = at::ones(b_shape, at::kHalf).cuda();
+  at::Tensor t2 = at::randn({m}, at::kHalf).cuda();
+  at::Tensor out_ref =
+      at::full(out_shape, k, at::kHalf).cuda() + t2.unsqueeze(-1);
+
+  FusionExecutorCache fec(std::move(fusion));
+  auto out = fec.runFusionWithInputs({t0, t1, t2});
+
+  EXPECT_TRUE(fec.getMostRecentKernelRuntime()->executors().size() == 1);
```
wujingyue, line 134:
```suggestion
  EXPECT_EQ(fec.getMostRecentKernelRuntime()->executors().size(), 1);
```

test/test_matmul_aten_evaluation.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+  // Verify that the io_alias_ set has the correct entry
+  auto kernel = fec.getMostRecentKernelRuntime()->executors().at(0).kernel();
+  EXPECT_TRUE(
+      kernel->getOutputAlias(kernel->outputs()[0]).type ==
+      AllocationType::Evaluate);
```
wujingyue, line 140:
```suggestion
  EXPECT_EQ(
      kernel->getOutputAlias(kernel->outputs()[0]).type,
      AllocationType::Evaluate);
```

test/test_matmul_aten_evaluation.cpp
wujingyue: The practical benefit is that EXPECT_EQ prints out the actual value and the expected for better debugging.

Priya2698:
> Nit: Will it make sense to make this C++ test file a part of the `test_matmul` target? https://github.com/NVIDIA/Fuser/blob/9fac12bdd98a63fdc88dde265b8add6a0e3f41cf/CMakeLists.txt#L480C1-L489C41

Thanks for the suggestion! Moved the test file.

jjsjann123:
FYI, this pr seems to break CI on V100.

Priya2698:
Thanks for pointing this out. Let's revert this until I identify the patch.

jjsjann123:
FYI, these are the failing tests from CI:
```
00:00:43 [  FAILED  ] 3 tests, listed below:
00:00:43 [  FAILED  ] MatmulATenEvaluationTest.SingleMmaOp
00:00:43 [  FAILED  ] MatmulATenEvaluationTest.MmaOpAndCast
00:00:43 [  FAILED  ] MatmulATenEvaluationTest.MatmulWithBias
```

Priya2698:
Yes, these are the tests I added to check functionality. 
Looking into it.

wujingyue:
> Thanks for pointing this out. Let's revert this until I identify the patch.

Thank you! I can't say enough great things about revert-and-debug-later!

Priya2698:
Are we looking to support V100 through the default path? CC: @kevinstephano 
Else, `NVFUSER_TEST_CUDA_ARCH_RANGE_GUARD` can be used in the tests to fix the error. The error surfaced because `matmulScheduler` does not support `V100` (See: https://github.com/NVIDIA/Fuser/blob/77caa576e52f4e9f55956db29ad519f035355041/csrc/scheduler/matmul_utils.cpp#L44)

wujingyue:
> Are we looking to support V100 through the default path?

I'd do it. This is actually something that can be supported with **much less** effort in fallback mode than in codegen. Sounds like a low-hanging fruit to me.

wujingyue:
> Are we looking to support V100 through the default path? CC: @kevinstephano `NVFUSER_TEST_CUDA_ARCH_RANGE_GUARD` can be used in the tests to fix the error. The error surfaced because `matmulScheduler` does not support `V100` (See:
> 
> https://github.com/NVIDIA/Fuser/blob/77caa576e52f4e9f55956db29ad519f035355041/csrc/scheduler/matmul_utils.cpp#L44
> 
> ).
> If we wish to support V100, we can have appropriate checks in the heuristic verification to allow other architectures when we are using expression evaluator.

FYI, I suspect it's not just V100. https://github.com/NVIDIA/Fuser/actions/runs/7910657123/job/21593586846 seems to be the same error but for H100.

protonu:
On option could be use to use something like this (https://github.com/NVIDIA/Fuser/blob/88727dc828684f5a62d7f1837a610b7589f629d1/test/test_combine_mul_sum.cpp#L40C1-L57C3) to reduce the machines these tests run on, in case you guys don't plan on adding support for V100/H100.

Priya2698:
> On option could be use to use something like this (https://github.com/NVIDIA/Fuser/blob/88727dc828684f5a62d7f1837a610b7589f629d1/test/test_combine_mul_sum.cpp#L40C1-L57C3) to reduce the machines these tests run on, in case you guys don't plan on adding support for V100/H100.

I am moving forward with supporting any architecture.
For this PR, the fix will be to skip the heuristic computation when the `MatmulExprEval` is set. In the next PRs, I'll refactor some code to always use the default scheduling for any unsupported architecture.

PR #1744, opened 2024-02-09
Layout propagation

jjsjann123:
Stacked PRs:
#1755 enabling layout propagation through runtime 
#1744 adding layout inference pass **_<- this one_**

What's in this PR:
The pass we have works on a Fusion IR:
It summaries MemoryFormat on inputs by looking at each TensorView's allocation_domain and rfactor_domain;
It uses a predefined rule (MemoryFormatInferencer) to propagate MemoryFormat from inputs to the entire fusion;

Note that the pass itself doesn't mutate the fusion IR. It's just a utility function that suggests ways to specify allocation domain to be used by other optimization passes.

- [x] adding some simple rule to propagate layout through fusion IR;
- [x] adding cpp test to verify propagation rule;

Quick design doc: https://github.com/NVIDIA/Fuser/issues/1756

Future Work:
* expanding propagation rule to cover more operation;


jjsjann123:
!build

test/utils.h
```diff
@@ -70,6 +70,14 @@ inline TensorView* makeSymbolicTensor(
   return TensorViewBuilder().shape(shape).dtype(dtype).build();
 }
 
+// allows overload resolution with size-1 initializer list
+inline TensorView* makeSymbolicTensor(
+    std::initializer_list<int64_t> shape,
```
jjsjann123, line 75: This API is just added to allow overload of `makeSymbolicTensor({-1}, ...)`, which would otherwise be called into `makeSymbolicTensor(size_t, ...)`

csrc/optimization/layout_inference.cpp
```diff
@@ -0,0 +1,206 @@
...
+//   lhs TV0 [i0, i1, b2]
+//             2   0   1
+//   rhs TV0 [i3, i4, b5]
+//             2   1   0
+//   if we go from innermost to outermost order:
+//     TV0 has i1 -> b2 -> i0
+//     TV1 has b5 -> i4 -> i3
+//   we see that TV0 encounters a non-broadcast iter domain first, so TV0 is the
+//   dominating tensor. We'll produce an output with stride order identical to
+//   that of TV0 in the record.
```
jjsjann123, line 82: @kevinstephano This was what I was describing on the propagation rule for binary operations.

jjsjann123:
:face_exhaling:  I realized I went with my old implementation and the memory order permutation is inconsistent with our `stride_order` in our python API.

Let me refactor that... :scream:

test/utils.h
jjsjann123: note to myself. Do a quick clean up for other APIs as well!

csrc/optimization/layout_inference.cpp
```diff
@@ -0,0 +1,238 @@
...
+#include <ir/all_nodes.h>
+#include <ir/utils.h>
+#include <optimization/layout_inference.h>
+
+namespace nvfuser {
+
+namespace {
+
+// move this to util maybe?
+std::vector<int64_t> ascendingAxes(const std::vector<int64_t>& permutation) {
```
wujingyue, line 17: Code comment on what it does?

test/test_layout_inference.cpp
```diff
@@ -0,0 +1,176 @@
...
+#include <test/validator.h>
+
+#include <torch/torch.h>
+
+namespace nvfuser {
+
+using testing::_;
+using testing::ElementsAre;
+
+class LayoutInferenceTest : public NVFuserTest {};
```
wujingyue, line 27:
```suggestion
using LayoutInferenceTest;
```

wujingyue (commented):
I like the test/code ratio! Thanks

jjsjann123:
!build

csrc/optimization/layout_inference.cpp
jjsjann123: sorry this was another confusion from switching back and forth. we don't need this util any more. I'll remove it.

test/test_layout_inference.cpp
jjsjann123: I think there's a typo in the suggestion. But I got what you meant.

jjsjann123:
!build

naoyam:
@jjsjann123 I'm a bit lost with what's addressed in this PR. According to your design doc, what'll be done are:

> 1. It looks up the permutation from rfactor_dom to allocation_dom on input TensorViews and record the permutation as MemoryFormat for those tensors;
> 2. The pass traverse the fusion to propagate MemoryFormat. It uses a set of propagation rules, where it compute & record MemoryFormat of outputs from the recorded MemoryFormat of inputs;
> 3. Lastly, the pass iterates through all output tensors and try to specify their allocation domain as per recorded MemoryFormat.

Am I correct that this PR does items 1 and 2? 

Also, what are propagation rules?
> It uses a set of propagation rules, where it compute & record MemoryFormat of outputs from the recorded MemoryFormat of inputs;

csrc/optimization/layout_inference.cpp
```diff
@@ -0,0 +1,236 @@
...
+class MemoryFormatInferencer : public OptOutConstDispatch {
+ public:
+  MemoryFormatInferencer(
+      std::unordered_map<const TensorView*, MemoryFormat>& format_map)
+      : format_map_(format_map) {}
+
+ private:
+  void handle(const UnaryOp*) override;
+  void handle(const BinaryOp*) override;
+  void handle(const BroadcastOp*) override;
```
jjsjann123, line 25: @naoyam Propagation rules are specified here per operation. I'll add a note on the commit description.

jjsjann123:
> Am I correct that this PR does items 1 and 2?

Yes. This PR just adds `MemoryFormatInferencer` which propagates memory format from inputs to the entire fusion.

jjsjann123:
Failing test seems to be coming from https://github.com/NVIDIA/Fuser/pull/1743.
cc'ing @Priya2698

csrc/optimization/layout_inference.cpp
```diff
@@ -0,0 +1,236 @@
...
+//   that.
+//   2. When both tensor have recorded memory format. It breaks tie based on the
+//   innermost dimension. whichever operand has a "better match" dominates the
+//   output format, where a "better match" meaning "less broadcast dimensions on
+//   the inner dimension"
+//
+// e.g.
+//   lhs TV0 rfactor_dom [i0, i1, b2]
+//                         0   2   1
+//   rhs TV0 rfactor_dom [i3, i4, b5]
```
wujingyue, line 67:
```suggestion
//   rhs TV1 rfactor_dom [i3, i4, b5]
```

csrc/optimization/layout_inference.cpp
```diff
@@ -0,0 +1,236 @@
...
+//                         0   1   2
+//   if we go from innermost to outermost order:
+//       TV0 has i1 -> b2 -> i0
+//       TV1 has b5 -> i4 -> i3
+//   we see that TV0 encounters a non-broadcast iter domain first, so TV0 is the
+//   dominating tensor. We'll produce an output with stride order identical to
```
wujingyue, line 73: > so TV0 is the dominating tensor

Why are we in favor of the memory format that first hits a non-broadcast? (I suspect it's something about vectorization, but the comment wasn't clear about that)

csrc/optimization/layout_inference.cpp
```diff
@@ -0,0 +1,236 @@
...
+//   that of TV0 in the record.
+//   In the event of a tie, we'll just propagate the memory format of lhs.
+void MemoryFormatInferencer::handle(const BinaryOp* op) {
+  TensorView* out = dynamic_cast<TensorView*>(op->out());
+  if (out == nullptr) {
+    return;
+  }
+  TensorView* lhs = dynamic_cast<TensorView*>(op->lhs());
+  TensorView* rhs = dynamic_cast<TensorView*>(op->rhs());
+  if (lhs == nullptr) {
```
wujingyue, line 83: OOC, when would this happen? `out` is a TensorView but `lhs` or `rhs` isn't?

csrc/optimization/layout_inference.cpp
```diff
@@ -0,0 +1,236 @@
...
+//                         0   2   1
+//   rhs TV0 rfactor_dom [i3, i4, b5]
+//                         0   1   2
+//   if we go from innermost to outermost order:
+//       TV0 has i1 -> b2 -> i0
+//       TV1 has b5 -> i4 -> i3
+//   we see that TV0 encounters a non-broadcast iter domain first, so TV0 is the
+//   dominating tensor. We'll produce an output with stride order identical to
+//   that of TV0 in the record.
+//   In the event of a tie, we'll just propagate the memory format of lhs.
```
wujingyue, line 75: Is
```
i->b->i
i->i->b
```
considered a tie? I.e., do you care about just the first non-broadcast or the first difference in which case the non-broadcast wins? Either case, why?

wujingyue (commented):
If it's convenient for you to `git rebase` and `git add -p`, I'd suggest separate the BinaryOp change to a different PR. That would reduce the size a lot to make review easier.

naoyam:
Can you please define what exactly the memory format means?

csrc/optimization/layout_inference.h
```diff
@@ -0,0 +1,24 @@
...
+namespace nvfuser {
+
+using MemoryFormat = std::vector<int64_t>;
+
+// Propagate memory format from input to the entire fusion. It does NOT modify
+// any fusion IR, but instead stores the propagated memory format as an
+// unordered_map from TensorView to permutation.
+//
+// See details in Note [ Memory Format Propagation ]
+std::unordered_map<const TensorView*, MemoryFormat> inferenceMemoryFormat(
```
naoyam, line 21: nit
```suggestion
std::unordered_map<const TensorView*, MemoryFormat> inferMemoryFormat(
```

csrc/optimization/layout_inference.cpp
```diff
@@ -0,0 +1,236 @@
...
+  // Since it only handles permutation from a rfactor domain to allocation
+  // domain, it can be interpreted as:
+  //
+  // e.g. TV0 rfactor domain [i0, i1, i2]
+  //            alloc domain [i0, i2, i1]
+  //           memory format   0,  2,  1
+  std::unordered_map<const TensorView*, MemoryFormat>& format_map_;
+};
+
+// UnaryOp propagation forward memory format from input to output
```
naoyam, line 43: What if the output has an allocation domain? Shouldn't the permutation be calculated here too?

naoyam:
> Can you please define what exactly the memory format means? Does it just mean the allocation domain?

I found a definition for tensors with an allocation domain:

```
// TensorView with allocatoin
//   domain that's a permutation of its corresponding rfactor domain and record
//   it as the memory format of the tensor;
```

What about tensors with no allocation domain?

naoyam:
Do we just want to infer a preferred allocation domain of each output tensor? 

How would you propagate a inferred format through reshape?

csrc/optimization/layout_inference.h
zasdfgbnm, line 1: Why is this file placed inside `csrc/optimization`? Is the layout inference an "optimization"? Should we just call it `passes` or something like that?

csrc/optimization/layout_inference.cpp
zasdfgbnm: I have the same question. Why is this better than just use `lhs`? @jjsjann123 Could you add the explanation to the code here as comment?

csrc/optimization/layout_inference.cpp
zasdfgbnm: Something like `T1 = T0 + 3`?

csrc/optimization/layout_inference.cpp
zasdfgbnm: I somehow feel that, what we should do is: if this binary op contains a broadcast concretization, then respect the one with most number of concrete IDs, otherwise, just use lhs. cc @naoyam

csrc/optimization/layout_inference.cpp
```diff
@@ -0,0 +1,236 @@
...
+std::unordered_map<const TensorView*, MemoryFormat> inferenceMemoryFormat(
+    Fusion* fusion) {
+  std::unordered_map<const TensorView*, MemoryFormat> memory_format_map;
+
+  // Note: we only consider simple permutation of allocation domain to rfactor
+  // domain.
+  for (auto tv : ir_utils::filterByType<TensorView>(fusion->inputs())) {
+    std::optional<MemoryFormat> permutation = ir_utils::computePermutation(
+        TensorDomain::noReductions(tv->getMaybeRFactorDomain()),
+        tv->getMaybeAllocationDomain());
```
zasdfgbnm, line 220: Should this be `TensorDomain::noReductions(tv->getMaybeAllocationDomain())`? IIRC allocation domain do have these reductions, although it makes no sense to do so.

zasdfgbnm: Also, should we make sure that reductions in the allocation domain are correctly handled?

test/utils.h
zasdfgbnm: C++ syntax is weird...

csrc/optimization/layout_inference.cpp
jacobhinkle: If we are propagating only in the forward direction, it seems to me that we can't really know what will be the most advantageous stride order. For example if we later do a sum on some outer dimension then it might wind up that we would have preferred that dimension to be allocated inner-most, but we would need to propagate that information backwards. If we're sticking with forward-only, why not just use the first input's stride order for the output and call it a day? If we want to chase more optimality we could consider doing an iterative optimization on the segmented fusion, allowing the schedulers to specify weighted preferences for the allocation orderings of their inputs and propagating changes to the outputs using simple rules like the one here, but that optimization is a bigger change to tackle.

csrc/optimization/layout_inference.cpp
```diff
@@ -0,0 +1,236 @@
...
+  }
+}
+
+} // namespace
+
+// Note [ Memory Format Propagation ]
+//
+// The propagation tries to propagate memory format from inputs to the entire
+// fusion:
+//   1. Iterates through all inputs, looking for TensorView with allocatoin
```
jacobhinkle, line 206:
```suggestion
//   1. Iterates through all inputs, looking for TensorView with allocation
```

csrc/optimization/layout_inference.h
jacobhinkle: One could argue this is an optimization, but I support changing the name since some other passes are not necessarily optimizing. `passes` might be too generic as there is already `device_lower/pass`. The debug dump option is `fusion_ir_preseg` and these are really the last thing before segmentation, so what about `preseg_passes`?

csrc/optimization/layout_inference.cpp
```diff
@@ -0,0 +1,236 @@
...
+      }
+      // go from innermost to outermost until we find the first one that's
+      // non-broadcast
+      auto rank = lhs_iter->second.size();
+      NVF_ERROR(
+          rank == rhs_iter->second.size(),
+          "expect binary op operands to have same length of memory format");
+      for (auto i : c10::irange(rank)) {
+        if (!lhs_iter->first
+                 ->getMaybeRFactorDomain()[lhs_iter->second[rank - 1 - i]]
```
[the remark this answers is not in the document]
jjsjann123: I was using that earlier, but backed out later. Not sure if I did a good job at cleaning up.

I think that ` TensorDomain::noReductions(getMaybeRFactorDomain())` only applies to segmented fusion?
I'm trying to limit the scope to be only running on entire fusion IR. But that indeed doesn't feel like it's simplifying the logic by much.

jacobhinkle (commented):
I am trying to understand how propagation can be more useful than the default (or arbitrary rules like using the first input's stride order) if we are only propagating in the forward direction.

csrc/optimization/layout_inference.h
```diff
@@ -0,0 +1,24 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+// clang-format on
+#include <unordered_map>
+
+#include <fusion.h>
+
+namespace nvfuser {
+
+using MemoryFormat = std::vector<int64_t>;
```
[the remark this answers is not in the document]
jjsjann123: There is a messy topic.

I was avoiding the term `StrideOrder`, because that's used in our python API. I want our python API to match what thunder's semantic of `StrideOrder` is. (which is, nhwc tensor would be written as [3, 0, 2, 1]).

Meanwhile, the format notation used in codegen would mark nhwc tensor as [0, 2, 3, 1]. The reason we want that is so that it looks more consistent with our setAllocationDomain API.
```
tv0->setAllocationDomain({tv0->axis(0), tv->axis(2), tv->axis(3), tv->axis(1)}, true);
```

csrc/optimization/layout_inference.cpp
jjsjann123: > Why are we in favor of the memory format that first hits a non-broadcast? (I suspect it's something about vectorization, but the comment wasn't clear about that)

This one should have been updated. I was doing this earlier when I use a different propagation rule for broadcast, so I needed this trick to propagate nhwc.

```
tv0 = [i0 i1 i2 i3] @ {0 2 3 1}
bias0 = [i4] @ {0} -> broadcast_bias0 [b5 i4 b6 b7] @ {0 1 2 3}
```
But now I feel @zasdfgbnm's suggestion makes a lot more sense instead.

> I somehow feel that, what we should do is: if this binary op contains a broadcast concretization, then respect the one with most number of concrete IDs, otherwise, just use lhs.

This makes a lot more sense to me. i.e. favoring larger tensor (hopefully more concrete IDs would lead to a larger tensor). I'll update.

csrc/optimization/layout_inference.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+ private:
+  // format_map_ records the stride order (memory format) of each TensorView.
+  // Since it only handles permutation from a rfactor domain to allocation
+  // domain, it can be interpreted as:
+  //
+  // e.g. TV0 rfactor domain [i0, i1, i2]
+  //            alloc domain [i0, i2, i1]
+  //           memory format   0,  2,  1
+  std::unordered_map<const TensorView*, MemoryFormat>& format_map_;
```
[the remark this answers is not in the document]
jjsjann123: I haven't touched that yet.

But I think it should be 3d here. i.e. we'll want to exclude reduction iterdomain, since it doesn't help `resolve` propagation with a binary op. We can probably just leave the reduction iterdomain on the left of allocation domain... or better yet, maybe we should just remove it from allocation domain since it doesn't carry any real meaning.

csrc/optimization/layout_inference.cpp
```diff
@@ -0,0 +1,236 @@
...
+// clang-format on
+#include <ir/all_nodes.h>
+#include <ir/utils.h>
+#include <optimization/layout_inference.h>
+
+namespace nvfuser {
+
+namespace {
+
+class MemoryFormatInferencer : public OptOutConstDispatch {
```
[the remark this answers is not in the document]
jjsjann123: Definitely should have used that one instead. Thanks :bow:

csrc/optimization/layout_inference.cpp
jjsjann123: I make the decision to limit the scope of the pass to only propagate from inputs to outputs. So any intermediate tensor with an allocation domain would just be ignored.

Now I felt @zasdfgbnm 's comment about `is this just a pass or an actual optimization thing?` is quite on point. A real optimization run should have considered existing allocation domain on intermediates.

csrc/optimization/layout_inference.cpp
```diff
@@ -0,0 +1,236 @@
...
+  }
+  TensorView* in = op->in()->as<TensorView>();
+  if (auto iter = format_map_.find(in); iter != format_map_.end()) {
+    format_map_[out] = iter->second;
+  }
+}
+
+// BinaryOp propagation tries to merge the memory format of both inputs
+//
+//   1. when there's only one operand has a recorded memory format, it forwards
```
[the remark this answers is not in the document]
jjsjann123: For inputs without allocation domain, we're leaving them as empty, which sounds like a bad idea.

Meanwhile, this could still happen to tensors created with factory method. Since we are only recording memory format of input tensors and I don't want that to affect the output memory format.

This resonates with @jacobhinkle 's other comment on `should we have a backward propagation as well`.

csrc/optimization/layout_inference.h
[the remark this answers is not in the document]
zasdfgbnm: Interesting... Could you add this note to the code as a comment?

csrc/optimization/layout_inference.h
zasdfgbnm: `preseg_passes` works for me. Or even simpler, just `preseg`. I have no preference over `preseg_passes` vs `preseg`.

csrc/optimization/layout_inference.cpp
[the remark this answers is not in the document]
zasdfgbnm: I this this also happens for fusions with `ReductionOp`?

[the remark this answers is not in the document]
jjsjann123: Touche~

I'll try to add this back on a few other places as well. Probably won't do a very good job there until later I started adding propagation rules for reduction op to actually verify it. Thanks for pointing it out. :bow:

csrc/optimization/layout_inference.cpp
wujingyue: Sure. I thought 3 would be a rank-0 TensorView, but looks like it's going to be a Scalar.

jjsjann123:
closing this PR since we are handling this one in #1788 #1790 #1792

Issue #1748, opened 2024-02-09
Build failure due to `c10::DeviceIndex`.


naoyam:
https://gitlab-master.nvidia.com/dl/pytorch/update-scripts/-/jobs/82329135

See also the Nightly-CI teams chat.

PR #390, opened 2023-05-22
Periodically clear c10 memory allocator in nvfuser_tests

jacobhinkle:
The `c10::cuda::CUDACachingAllocator` tends to accumulate memory leading to >20GB usage when running `nvfuser_tests`. This causes OOM on some CI jobs. This PR just empties the allocator every time 100 tensors have been allocated using the `make*Tensor` functions in `test/utils.h`.

PR #1749, opened 2024-02-09
Issue #1748

Priya2698:
Closes Issue #1748.
Apart from `c10::cuda::GetDevice`, no other functionality seems affected.

### DIFF
--- a/test/utils.h
+++ b/test/utils.h
@@ -402,7 +402,7 @@ inline bool maybeClearAllocator(int64_t max_bytes = ((int64_t)1 << 32)) {
     cudaGetDevice(&device);
 #endif
 
-    auto device_stats = allocator->getDeviceStats(0);
+    auto device_stats = allocator->getDeviceStats(device);
     // allocated_bytes[] holds multiple statistics but the first is sum across
     // both small and large blocks
     if (uint64_t(device_stats.reserved_bytes[0].current) >
--- a/test/utils.h
+++ b/test/utils.h
@@ -397,7 +397,9 @@ inline bool maybeClearAllocator(int64_t max_bytes = ((int64_t)1 << 32)) {
     // in order to properly handle new CUDA 112 behavior
     // c10::cuda uses DeviceIndex instead of int
     // https://github.com/pytorch/pytorch/pull/119142
-    c10::cuda::GetDevice(reinterpret_cast<c10::DeviceIndex*>(&device));
+    c10::DeviceIndex device_index;
+    c10::cuda::GetDevice(&device_index);
+    device = static_cast<int>(device_index);
 #else
     cudaGetDevice(&device);
 #endif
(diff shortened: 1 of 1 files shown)


test/utils.h
```diff
@@ -395,7 +395,9 @@ inline bool maybeClearAllocator(int64_t max_bytes = ((int64_t)1 << 32)) {
 #if TORCH_VERSION_GREATER(2, 0, 1)
     // GetDevice was introduced in https://github.com/pytorch/pytorch/pull/94864
     // in order to properly handle new CUDA 112 behavior
-    c10::cuda::GetDevice(&device);
+    // c10::cuda uses DeviceIndex instead of int
+    // https://github.com/pytorch/pytorch/pull/119142
+    c10::cuda::GetDevice(reinterpret_cast<c10::DeviceIndex*>(&device));
```
naoyam, line 400: Hmm, why this is needed? It seems we are not using `device`. 

@jacobhinkle I remember this part was originally written by you. Looks like this is a related comment.

https://github.com/NVIDIA/Fuser/pull/390#discussion_r1202728217

I wonder `device` should be used at line 405 instead of `0`?

Priya2698: Looks like getDeviceStats still uses `int`.
https://github.com/pytorch/pytorch/blob/3372aa51b4797b69db97ba124b880e891ad5a85e/c10/cuda/CUDACachingAllocator.h#L233

jacobhinkle: You're right it should be used instead of 0. That was my mistake.

Priya2698: I'll add the change to this PR.

test/utils.h
```diff
@@ -395,12 +395,14 @@ inline bool maybeClearAllocator(int64_t max_bytes = ((int64_t)1 << 32)) {
...
```
naoyam, line 400: Is this `reinterpret_cast` safe? Looks like `c10::DeviceIndex` is `int8_t`.

https://github.com/pytorch/pytorch/blob/main/c10/core/Device.h#L19

I suppose what this `reinterpret_cast` would mean is we are only using one byte out of the int `device` variable. This may work, but seems fragile.

naoyam:
```suggestion
    c10::DeviceIndex device_index;
    c10::cuda::GetDevice(&device_index);
    int device = device_index; // or maybe static_cast<int>(device_index);
```

Generally try to avoid using `reintepret_cast` unless really necessary.

Priya2698: Another option is to declare device as type `DeviceIndex`.

naoyam: Yeah, but an annoying thing is that you still need to have an int-type device index for `getDeviceStats`, right?

Priya2698: Yes, the commit has only changed the type at some locations so far.

Priya2698: > Another option is to declare device as type `DeviceIndex`.
I was referring to the same thing you said, since we need an `int` device.

naoyam (approved):
LGTM. Thanks for the fix.

PR #1752, opened 2024-02-11
Device index patch

jjsjann123:
Fixes #1748 

guard c10::cuda::GetDevice API change on TORCH_VERSION

with this change, it ensures that we can build against stable release `< 2.2.0`, as well as TOT after  https://github.com/pytorch/pytorch/pull/119142

For 2.3.0 nightly, if someone accidentally checkout a commit before the patch, the build will still fail.

PR #1753, opened 2024-02-12
fixing multidevice build

jjsjann123:
API change coming from https://github.com/pytorch/pytorch/pull/119421

### DIFF
--- a/csrc/multidevice/communication.cpp
+++ b/csrc/multidevice/communication.cpp
@@ -236,7 +236,8 @@ c10::intrusive_ptr<c10d::Work> Reduce::post(
 #ifdef USE_C10D_NCCL
   auto nccl_backend = dynamic_cast<c10d::ProcessGroupNCCL*>(team_backend.get());
   if (nccl_backend) {
-    return nccl_backend->_reduce_oop(buf, params_.src_bufs, options);
+    return nccl_backend->_reduce_oop(
+        buf.at(0), params_.src_bufs.at(0), options);
   }
 #endif
   if (comm.deviceId() == params_.root) {


jjsjann123:
patching issue mentioned in: https://github.com/NVIDIA/Fuser/pull/1752#issuecomment-1937897678

My naive understanding going through the code base is that, we are always using buffer count of 1 any way from the assert, so the API change here really shouldn't have changed anything. cc'ing @samnordmann

jjsjann123:
!build

kwen2501 (approved):
LGTM!
Sorry about the break :)

naoyam (approved):

naoyam:
The build still seems failing. Does it also need the version guard?

naoyam:
```
/opt/pytorch/nvfuser/csrc/multidevice/communication.cpp
#12 63.49 /opt/pytorch/nvfuser/csrc/multidevice/communication.cpp: In member function ‘virtual c10::intrusive_ptr<c10d::Work> nvfuser::Reduce::post(nvfuser::Communicator&, std::optional<nvfuser::CommunicatorBackend>)’:
#12 63.49 /opt/pytorch/nvfuser/csrc/multidevice/communication.cpp:240:15: error: cannot convert ‘__gnu_cxx::__alloc_traits<std::allocator<at::Tensor>, at::Tensor>::value_type’ {aka ‘at::Tensor’} to ‘std::vector<at::Tensor>&’
#12 63.49   240 |         buf.at(0), params_.src_bufs.at(0), options);
```

jjsjann123:
> The build still seems failing. Does it also need the version guard?

I believe internal failure are because we are building against our container image.. which has an outdated code base. cc'ing @xwang233 to confirm that.

xwang233:
Yes, the internal image has pytorch from last night, which has the PR reverted and not-yet relanded.

jjsjann123:
> The build still seems failing. Does it also need the version guard?

I don't think we can version guard anything here. Since it's API change on ToT...

jjsjann123:
Merging as-is since nightly build succeeded.

PR #1755, opened 2024-02-13
Layout propagation (Part 2) - Enable

jjsjann123:
Stacked PRs:
==== #1755 enabling layout propagation through runtime <- **_this one_**
#1792 propagation rule for broadcast
#1790 propagation rule for binary op
#1788 adding layout inference pass

What's in this PR:
Enabling the MemoryFormat optimization pass in runtime. The pass is run as part of pre_segment optimization pass.
Adding cpp test to verify optimization behavior

Quick design doc: https://github.com/NVIDIA/Fuser/issues/1756

TODOs:
- [ ] rebase with the new inference rule PRs.


jjsjann123:
`48d3ee162b1cc7b2fe9796869e3a19dad3e36c52` local failing tests:
```
[  FAILED  ] 3 tests, listed below:
[  FAILED  ] NVFuserTest.FusionAvoidRedundantWrite_CUDA
[  FAILED  ] AliasTest.TrivialInputForwarding_ScalarTensor
[  FAILED  ] NoOpTest.FusionNullScheduler3
```

jjsjann123:
!build

csrc/optimization/layout_optimization.cpp
```diff
@@ -0,0 +1,45 @@
...
+#include <optimization/layout_inference.h>
+#include <optimization/layout_optimization.h>
+
+namespace nvfuser::optimization {
+
+void LayoutOptimizationPass::runPass(Fusion* fusion) {
+  std::unordered_map<const TensorView*, MemoryFormat> stride_mapping =
+      inferenceMemoryFormat(fusion);
+
+  for (auto out_val : fusion->outputs()) {
```
wujingyue, line 17:
```suggestion
  for (Val* out_val : fusion->outputs()) {
```

csrc/optimization/layout_optimization.cpp
```diff
@@ -0,0 +1,45 @@
...
+    auto out_tv = dynamic_cast<TensorView*>(out_val);
```
wujingyue, line 18:
```suggestion
    auto* out_tv = dynamic_cast<TensorView*>(out_val);
```

csrc/optimization/layout_optimization.cpp
```diff
@@ -0,0 +1,45 @@
...
+    // skip:
+    //   1. non-tensor output;
+    //   2. tensor output with allocation specified, assuming everything is
+    //   semantical
+    //   3. tensor output that's aliasing (Does aliased src matter?)
```
wujingyue, line 23: > Does aliased src matter?

No. That's mainly for error checking -- in allocateOutput, we check the aliasing output is indeed an alias of the aliased input.

csrc/optimization/layout_optimization.cpp
```diff
@@ -0,0 +1,45 @@
...
+    if (mapped_entry == stride_mapping.end() || mapped_entry->second.empty()) {
+      continue;
+    }
+
+    auto rfactor_dom = out_tv->getMaybeRFactorDomain();
+
+    auto rank = rfactor_dom.size();
+    std::vector<IterDomain*> allocation_domain(rank, nullptr);
+    for (auto i : c10::irange(rank)) {
+      allocation_domain.at(i) = rfactor_dom.at(mapped_entry->second.at(i));
```
wujingyue, line 39:
```suggestion
      allocation_domain[i] = rfactor_dom.at(mapped_entry->second[i]);
```

Is there a particular reason to use `.at(i)`? `[i]` is less typing.

test/test_layout_inference.cpp
```diff
@@ -173,4 +173,31 @@ TEST_F(LayoutInferenceTest, BinaryOpPropagation) {
   }
 }
 
+TEST_F(LayoutInferenceTest, MemoryFormatOptimizationTest) {
```
wujingyue, line 176:
```suggestion
TEST_F(LayoutOptimizationTest, MemoryFormat) {
```

test/test_layout_inference.cpp
wujingyue, line 1: Rename the file to `test_layout.cpp`, because now it's both for the analysis and the optimization?

wujingyue: I'm also not sure about the practical benefit of separating analysis and optimization. It makes sense for alias analysis because it's used by both MarkAliasesPrepare (pre segmentation) and markAliases (in scheduling).

csrc/optimization/layout_optimization.cpp
zasdfgbnm, line 17: Why are we only setting the allocation domain of the output of the unsegmented fusion? If my fusion is:
```C++
input: t0
output: t2
t1 = segmenter_set(t0)
t2 = segmenter_set(t1)
```
and `t0` is NHWC, then `t2` will be set as NHWC correctly, but `t1` will remain NCHW? Then both segments will use transpose scheduler instead of pointwise scheduler? This behavior doesn't make sense to me. Should we just set all tensor's allocation domain here, and inside the `scheduleXXX` (where `XXX` = `Pointwise`, `Transpose`, etc.) functions, reset the intermediate tensor's allocation before doing any real schedule?

test/test_layout_inference.cpp
jjsjann123: I'm expecting this would be the same. i.e. we might want to have scheduler use the pass in the next step, instead of at pre-segmentation phase.
Seems to be just a small refactor later, so I'll just merge the source files. :bow:

csrc/optimization/layout_optimization.cpp
jjsjann123: Should I not rely on `getOutputAlias` to see if an output is just an alias? I'm just trying to avoid changing allocation domain on tensor that could have been just an alias

csrc/optimization/layout_optimization.cpp
jjsjann123: `at` gives an assert for out of range, while [] operator is just an UB.

But I do see the point of changing indexing into `allocation_domain` here and I'll change that one.

test/test_layout_inference.cpp
jjsjann123: renamed this to `EnableInRuntime`

csrc/optimization/layout_optimization.cpp
jjsjann123: That's a very reasonable ask. i.e. not just for `segmenter_set`, we might as well want to support general segmented fusion.

But it's currently not in the scope for initial PRs. I'm leaning towards that we'd want to push this pass to each scheduler to handle, which we'll naturally be able to handle segment boundaries.

csrc/optimization/layout_optimization.cpp
wujingyue: Sorry, my earlier comment was confusing. 

There's nothing wrong with this code as is, and you don't need to check `aliased_io`.

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -99,4 +99,36 @@ std::unordered_map<const TensorView*, AllocationOrder> inferenceAllocationOrder(
...
+    //   2. tensor output with allocation specified, assuming everything is
+    //   semantical
+    //   3. tensor output that's aliasing (Does aliased src matter?)
+    if (out_tv == nullptr || out_tv->hasAllocation() ||
+        fusion->getOutputAlias(out_val).type != AllocationType::NoAlias) {
+      continue;
+    }
+
+    auto mapped_entry = stride_mapping.find(out_tv);
+    if (mapped_entry == stride_mapping.end() || mapped_entry->second.empty()) {
```
wujingyue, line 119: What's about `mapped_entry->second.empty()`?

test/test_allocation_order_inference.cpp
```diff
@@ -43,4 +43,31 @@ TEST_F(AllocationOrderInferenceTest, UnaryOpPropagation) {
...
+  at::Tensor in_tensor = at::randn({2, 4, 8, 8}, options);
+  at::Tensor in_nhwc =
+      in_tensor.as_strided({2, 4, 8, 8}, {4 * 8 * 8, 1, 4 * 8, 4});
+  FusionExecutorCache fec(std::move(fusion));
+
+  auto cg_outputs = fec.runFusionWithInputs({in_nhwc});
+  auto ref_out = in_nhwc.relu();
+
+  ASSERT_TRUE(cg_outputs[0].is_contiguous(at::MemoryFormat::ChannelsLast));
+  ASSERT_TRUE(ref_out.allclose(cg_outputs[0]));
```
wujingyue, line 70:
```suggestion
  EXPECT_TRUE(cg_outputs[0].is_contiguous(at::MemoryFormat::ChannelsLast));
  EXPECT_TRUE(ref_out.allclose(cg_outputs[0]));
```

wujingyue (approved):

csrc/preseg_passes/allocation_order_inference.cpp
wujingyue: (To be more verbose to reduce roundtrip) 

I was trying to see if removing this check is all fine. It looks like this only happens for rank-0 tensors and even without this check the code below is general enough to handle this special case.

jjsjann123: I think you are right.

I did forget to call `noReduction` on rfactor_domain of `out_tv`, which I think would make it safe for a rank-0 tensor from a reduction. Will update.

csrc/optimization/layout_optimization.cpp
zasdfgbnm: Sure, it don't have to be done in this PR. But why do you think
```
we'd want to push this pass to each scheduler
```
? Schedulers needs to know the stride order to decide if it can schedule the fusion or not.

jjsjann123: > ? Schedulers needs to know the stride order to decide if it can schedule the fusion or not.

Sure, if we do want a certain tensor to return in a specific stride order, that needs to be set prior to passing things to scheduler.

But for tensor with an empty allocation domain, it felt natural for a scheduler to come in and return something like. "i can schedule this fusion, meanwhile, I'm going to fill in empty allocation domain to something I can generate efficient kernel for"

jjsjann123:
!build

csrc/optimization/layout_optimization.cpp
zasdfgbnm: > I'm going to fill in empty allocation domain

Isn't this information ambiguous? For example, for a segment `T2 = set(T1)` where neither the allocation domain of `T1` and `T2` is set, the pointwise scheduler can fill either NHWC or NCHW, or WNHC, etc., it is hard to infer which one is preferred without the global stride order information from propagation.

jjsjann123: >  the global stride order information from propagation.

I'm not saying that the decision should be made merely within a `segment`. You should still see information propagated from your producer segment.

i.e. `T2 = set(T1)` would have a producer segment that hopefully will try to mark allocation domain on T1. So within each segment we are propagating only from inputs to outputs and that shouldn't leave any ambiguity in the rule. I think this works better with smaller segments using a small set of propagation rules.

I think what you were proposing is more aligned with @jacobhinkle's suggestion, an overall pass the does both forward and backward propagation. Which I'm very open minded as a future refactor.

jjsjann123:
failure is unrelated in CI. merging as-is.

PR #1761, opened 2024-02-14
Automatic multidevice fusion creation

samnordmann:
Make Multidevice Fusion creation and execution automatic.

# What

Before this patch, one had to manually schedule a pipeline and indicate where communications need to be inserted, i.e., add each Val to a `PipelineStage`, and then instantiate an object `Pipeline` that represents the "aggregate DAG" of `PipelineStage` and `PipelineCommunication`

With this patch, the API becomes much simpler, and we only have to define the Fusion in a standard way, and then to set tensors sharding through the multidevice API (i.e. `tv->setDeviceMesh` and `tv->axis(0)->parallelize(ParallelType::DIDx)`). Then, the user can simply instantiate the `MultideviceExecutor` and call its method `runWithInput`. The following steps are then automatically executed
- network communication (representing at the symbolic level by resharding "Set" Exprs) are automatically inserted in the fusion (see the function`insertReshardings`)
- the Fusion is segmented into segments which can be of two types: 1) compute segments and 2) communications. This uses our recently implemented multidevice segmentater
- the runtime order of execution of the different segments is inferred and stored
- necessary recv buffers are allocated
- The executor posts the comms/kernels
- Cache mechanism is used to avoid compilation/set-up overhead

# Why

- the implementation and API is **much** simpler and cleaner. Now the fusion definition and "pipeline" scheduling are clearly separated. We got rid of all the previous pipeline infrastructure: the IRs (`PipelineVal`, `PipelineCommunication`, `PipelineStage`), the container `Pipeline`
- The multidevice fusion definition is more user friendly since we automate the stage creation
- using the segmenter allows to support much more situations than before
- using insertResharding allows to support any resharding expr besides simple set and reduce

# Example
## "Allreduce" fusion, between devices 0,1,2,3

```
DeviceMesh mesh ({0,1,2,3});
std::vecotr<int> input_sizes = {4, 16};

TensorView* tv0 = makeConcreteTensor(input_sizes);
TensorView* tv1 = sum(tv0, {0});
fusion->addInput(tv0);
fusion->addOutput(tv1);

for (auto tv: {tv0,tv1}) {
  tv->setDeviceMesh(mesh);
}

inputs = {at::randn(input_sizes, tensor_options)};

MultiDeviceExecutor runtime(std::move(fusion_ptr), *communicator);
if (std::string error_msg = runtime.validate(); error_msg != "") {
  NVF_ERROR(false, error_msg);
}
auto outputs = runtime.runWithInput(inputs);
```

## "Allgather" fusion, between devices 0,1,2,3

```
DeviceMesh mesh ({0,1,2,3});
std::vecotr<int> input_sizes = {4, 16};

TensorView* tv0 = makeConcreteTensor(input_sizes);
TensorView* tv1 = set(tv0);
fusion->addInput(tv0);
fusion->addOutput(tv1);

for (auto tv: {tv0,tv1}) {
  tv->setDeviceMesh(mesh);
}
tv0->axis(0)->parallelize(ParallelType::DIDx);

inputs = {at::randn(input_sizes, tensor_options)};

MultiDeviceExecutor runtime(std::move(fusion_ptr), *communicator);
if (std::string error_msg = runtime.validate(); error_msg != "") {
  NVF_ERROR(false, error_msg);
}
auto outputs = runtime.runWithInput(inputs);
```

## "Reduce-scatter" fusion, between devices 0,1,2,3

```
DeviceMesh mesh ({0,1,2,3});
std::vecotr<int> input_sizes = {4, 4, 16};

TensorView* tv0 = makeConcreteTensor(input_sizes);
TensorView* tv1 = sum(tv0, {0});
fusion->addInput(tv0);
fusion->addOutput(tv1);

for (auto tv: {tv0,tv1}) {
  tv->setDeviceMesh(mesh);
}
tv0->axis(0)->parallelize(ParallelType::DIDx);
tv1->axis(1)->parallelize(ParallelType::DIDx);

inputs = {at::randn(input_sizes, tensor_options)};

MultiDeviceExecutor runtime(std::move(fusion_ptr), *communicator);
if (std::string error_msg = runtime.validate(); error_msg != "") {
  NVF_ERROR(false, error_msg);
}
auto outputs = runtime.runWithInput(inputs);
```

## "Pipeline parallelism"

```
DeviceMesh mesh1 ({0,1});
DeviceMesh mesh2 ({2,3});
std::vecotr<int> input_sizes = {2, 8, 16};

TensorView* tv0 = makeConcreteTensor(input_sizes);
TensorView* tv2 = set(tv1);
TensorView* tv3 = sum(tv2, {1});
fusion->addInput(tv0);
fusion->addOutput(tv3);

for (auto tv: {tv0,tv1,tv2,tv3}) {
  tv0->axis(0)->parallelize(ParallelType::DIDx);
}
tv0->setDeviceMesh(mesh1);
tv1->setDeviceMesh(mesh1);
tv2->setDeviceMesh(mesh2);
tv3->setDeviceMesh(mesh2);

inputs = {at::randn(input_sizes, tensor_options)};

MultiDeviceExecutor runtime(std::move(fusion_ptr), *communicator);
if (std::string error_msg = runtime.validate(); error_msg != "") {
  NVF_ERROR(false, error_msg);
}
auto outputs = runtime.runWithInput(inputs);
```


csrc/tensor_view.cpp
```diff
@@ -18,6 +18,7 @@
 #include <ir/interface_nodes.h>
 #include <ir/internal_nodes.h>
 #include <ir/iostream.h>
+#include <ir/printer.h>
```
cowanmeg, line 21: nit: one more

csrc/multidevice/executor.cpp
```diff
@@ -6,133 +6,234 @@
...
+    }
+    if (val_to_IValue_.find(output_val) != val_to_IValue_.end()) {
+      output_tensor = val_to_IValue_.at(output_val).toTensor();
+    }
     communications_.emplace(
-        c,
+        group,
         lowerCommunication(
-            runtime_.comm_.deviceId(), c, input_tensor, output_tensor));
+            comm_.deviceId(), expr, input_tensor, output_tensor));
```
cowanmeg, line 192: Does this cache input_tensor and output_tensor?

cowanmeg (commented):
Thanks Sam! This PR makes the multidevice interface much cleaner and unblocks many PRs!

csrc/multidevice/executor.cpp
```diff
@@ -6,133 +6,234 @@
...
+        !is_resharding_[group] || group->exprs().size() == 1,
+        "Communications cannot be fused");
+    auto expr = group->exprs().at(0);
+    should_run_[group] = involvedDevices(expr).count(comm_.deviceId());
   }
-  return should_run_[stage];
+  // prepare the order in which to launch the kernels/comms
+  RuntimeWorkSpace workspace;
+  prepareRuntimeOrder(staged_fusion_.get(), workspace);
+  group_run_order_ = std::move(workspace.group_run_order);
```
naoyam, line 127: I'd just have a `RuntimeWorkSpace` in `MultiDeviceExecutor` rather than just `group_run_order`. The other vector would be eventually required.

csrc/multidevice/executor.h
```diff
@@ -8,56 +8,75 @@
...
-
-  // Cache results of shouldRun method
-  std::unordered_map<PipelineStage*, bool> should_run_;
-
-  // MultiDeviceRuntime to be executed
-  MultiDeviceRuntime& runtime_;
+  // Cache whether a SegmentedGroup should be run by the current device
+  std::unordered_map<SegmentedGroup*, bool> should_run_;
+  // Cache whether a SegmentedGroup requires inter-device communication
+  std::map<SegmentedGroup*, bool> is_resharding_;
```
naoyam, line 79: Any particular reason using `std::map` instead of `std::unordered_map`?

csrc/multidevice/executor.h
```diff
@@ -8,56 +8,75 @@
...
+#include <multidevice/multidevice.h>
 
 namespace nvfuser {
 
-// Runtime Executor for Pipelines
-// This class inherits from IterVisitor because the execution
-// is ordered by the traversal of the Pipeline seen as a DAG
-class PipelineExecutor : public IterVisitor {
+/*
+  The MultiDeviceExecutor executes a Fusion on a multi-device setting.
```
naoyam, line 22: Can you please add more details here? Something like you have as the PR description.

csrc/multidevice/executor.cpp
naoyam: I suppose Meghan is asking if the aten tensors are reused. I think they should not be reused, but looks like they are indeed reused, right?

csrc/multidevice/executor.cpp
```diff
@@ -6,133 +6,234 @@
...
 namespace nvfuser {
 
-bool PipelineExecutor::shouldRun(PipelineStage* stage) {
-  if (should_run_.find(stage) == should_run_.end()) {
-    should_run_.emplace(
-        stage, stage->descriptor()->mesh.has(runtime_.comm_.deviceId()));
+namespace {
+
+std::pair<std::unique_ptr<Fusion>, std::unordered_map<Val*, Val*>>
+copyFusionAndChangeOutputs(Fusion* fusion, std::unordered_set<Val*> outputs) {
```
naoyam, line 23:
```suggestion
copyFusionAndChangeOutputs(Fusion* fusion, const std::unordered_set<Val*>& outputs) {
```

csrc/multidevice/executor.cpp
```diff
@@ -6,133 +6,234 @@
...
 
 namespace nvfuser {
 
-bool PipelineExecutor::shouldRun(PipelineStage* stage) {
-  if (should_run_.find(stage) == should_run_.end()) {
-    should_run_.emplace(
-        stage, stage->descriptor()->mesh.has(runtime_.comm_.deviceId()));
+namespace {
+
+std::pair<std::unique_ptr<Fusion>, std::unordered_map<Val*, Val*>>
```
naoyam, line 22: Please add a comment. I don't understand what this function is for.

csrc/multidevice/executor.cpp
```diff
@@ -6,133 +6,234 @@
...
+  auto original_outputs = fusion_copy->outputs();
+
+  // Remove original outputs
+  std::for_each(
+      original_outputs.begin(), original_outputs.end(), [&](auto& output) {
+        fusion_copy->removeOutput(output);
+      });
+
+  // Add new outputs
+  std::for_each(outputs.begin(), outputs.end(), [&](Val* const& output) {
```
naoyam, line 38: I'm not exactly sure what this function is meant to do, but since `outputs` is an unordered map, the order of the outputs will be non-deterministic, which should be avoided.

csrc/multidevice/executor.h
```diff
@@ -8,56 +8,75 @@
...
+  // Run the fusion on several devices with the given global inputs
   std::vector<at::Tensor> runWithInput(const std::vector<c10::IValue>& inputs);
 
+  // Returns the Communicator
+  Communicator* comm() const {
+    return &comm_;
+  }
+
+  // Returns the Fusion
+  auto fusion() const {
```
naoyam, line 38: I'd call this `completeFusion` as it's often really confusing which fusion we are talking about.

csrc/ir/utils.h
```diff
@@ -7,6 +7,7 @@
 // clang-format on
 #pragma once
 
+#include <disjoint_set.h>
```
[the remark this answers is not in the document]
samnordmann: it's needed

csrc/alias_analysis.cpp
```diff
@@ -13,6 +13,7 @@
 #include <fusion.h>
 #include <ir/interface_nodes.h>
 #include <ir/internal_base_nodes.h>
+#include <ir/iostream.h>
```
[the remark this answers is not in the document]
samnordmann: needed also

csrc/multidevice/executor.cpp
samnordmann: > I suppose Meghan is asking if the aten tensors are reused. I think they should not be reused, but looks like they are indeed reused, right?

Yes, they are reused. I agree with you that this is not ideal (and there might be a bug if we execute the same fusion multiple times, I just thought about it). However, this is the only way to use the cache for now because, at a lower level, the communications (in `multidevice/communications.h`) are instantiated with Aten buffers, instead of symbolic values.

If this is a concern we can simply turn off this basic caching mechanism for now.

csrc/multidevice/utils.cpp
```diff
@@ -111,9 +111,10 @@ void shardAllLike(TensorView* ref, std::vector<TensorView*> tvs) {
   for (auto tv : tvs) {
     tv->setDeviceMesh(ref->getDeviceMesh());
   }
-  scheduler_utils::parallelizeAllLike(ref, tvs, {ParallelType::DIDx});
+  if (!tvs.empty()) {
+    scheduler_utils::parallelizeAllLike(ref, tvs, {ParallelType::DIDx});
```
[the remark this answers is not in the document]
samnordmann: no, it is just that if tvs is empty, the function is applied to all tvs of the fusion. It is a misleading API in my opinion, but I didn't want to make the decision to change it

csrc/kernel_cache.cpp
```diff
@@ -362,6 +362,65 @@ InputsIdLookup::IdLookupReturn InputsIdLookup::lookupId(
   return ret;
 }
 
+void prepareRuntimeOrder(
```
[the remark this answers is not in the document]
samnordmann: no, it is just copy pasted and de-indented

csrc/multidevice/executor.cpp
samnordmann: no problem, I add it

csrc/multidevice/executor.cpp
samnordmann: I actually removed the communication lowering's caching, I think it is faulty, and for sure it should be made better to be interesting. Will work on it on a separate PR

csrc/kernel_cache.cpp
```diff
@@ -362,6 +362,65 @@ InputsIdLookup::IdLookupReturn InputsIdLookup::lookupId(
...
+  for (const size_t i : c10::irange(segmented_fusion->inputs().size())) {
+    auto input_val = segmented_fusion->inputs()[i];
+    available_input.insert(input_val);
+
+    if (auto input_tv = dynamic_cast<TensorView*>(input_val)) {
+      auto root_dom = TensorDomain::noReductions(input_tv->getRootDomain());
+      for (const size_t dim : c10::irange(root_dom.size())) {
+        const auto extent = root_dom[dim]->getMaybeExpandedExtent();
+        available_input.insert(extent);
+        runtime_workspace.group_extent_binding_order.push_back(extent);
```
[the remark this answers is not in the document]
samnordmann: I am not sure... I just copy-pasted it

csrc/multidevice/executor.h
samnordmann: no reason, I change as you suggest

csrc/multidevice/executor.h
samnordmann: sure, let me know if it is ok like that

csrc/multidevice/executor.cpp
samnordmann: let me know if it is clearer this way. I also added a "TODO" comment for `allocateRecvBuffers`

csrc/multidevice/executor.cpp
samnordmann: For our case it shouldn't matter that the order is non-deterministic. Also since the output container of both the original and the copy are unordered, I don't see why this is a problem.

This is mimicked on `SegmentedFusion::makeFusion`

I am ok to change according to your suggestion but I don't see how. Let me know.

csrc/multidevice/utils.cpp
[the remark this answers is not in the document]
naoyam: Hmm, that's indeed confusing.

csrc/kernel_cache.cpp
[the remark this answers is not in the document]
naoyam: It seems it passes some extents used in fusion segments.

csrc/multidevice/executor.h
naoyam: Looks good. Let's also clarify any assumptions about the input fusion. At this point, it must be manually annotated with `DeviceMesh` and DID parallelization. Anything else?

csrc/multidevice/executor.cpp
naoyam: `Fusion::outputs()` returns a deterministically ordered `std::vector`. 

Doesn't the non-determinism here mean generated code can be non-deterministic? That shouldn't be a problem mathematically, but we do not want unnecessary non-determinism.

samnordmann: ok, I changed to type of `ouputs` (the argument of `copyFusionAndChangeOutputs`) from `unordered_set` to `set`. Does it address your concern ?

samnordmann:
build

csrc/multidevice/executor.h
samnordmann: I added explanations and details on the assumptions. Let me know how it looks to you

csrc/multidevice/executor.cpp
naoyam: No, I don't think so since the order in `std::set` should be just defined by object addresses, so it won't be deterministic.

csrc/multidevice/executor.cpp
naoyam: OK, I see. I did understand what it does, but I didn't why.

csrc/multidevice/executor.cpp
naoyam: Just use `std::vector`

naoyam (approved):
LGTM. Great to see code size reduction!

PR #1762, opened 2024-02-14
fix compute at root domain map

### LINKED ISSUE #1631: Runtime Error running llama2 `RMSNorm` module in benchmarks
The following happens when running the benchmarks for RMSNorm with input size 4096:
 ```
RuntimeError: !fallback_mode_enabled_ INTERNAL ASSERT FAILED at "/opt/pytorch/nvfuser/csrc/device_lower/pass/expr_sort.cpp":1096, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. Couldn't succcessfully sort out the fusion expressions. There are remaining connections of the heirarchical segmentation which should have been flattened to a single ordered group, or disjoint ordered groups.
 ```
 Repro:
 
 ```python
import torch

    T0 = fd.define_tensor(shape=[-1], contiguity=[True], dtype=DataType.BFloat16, is_cpu=False, stride_order=[0])
    T1 = fd.define_tensor(shape=[-1], contiguity=[True], dtype=DataType.BFloat16, is_cpu=False, stride_order=[0])
    S2 = fd.define_scalar(1, dtype=DataType.Int)
    S3 = fd.define_scalar(4096, dtype=DataType.Int)
    V4 = fd.define_vector([S3], dtype=DataType.Int)
    T5 = fd.ops.full(shape=V4, fill_value=S2, dtype=DataType.BFloat16)
    T6 = fd.ops.cast(T1, dtype=DataType.Float)
    T7 = fd.ops.mul(T6, T6)
    T8 = fd.ops.sum(T7, axes=[0], keepdim=False, dtype=DataType.Null)
    S9 = fd.define_scalar(1, dtype=DataType.Int)
    V10 = fd.define_vector([S9], dtype=DataType.Int)
    T11 = fd.ops.broadcast_in_dim(T8, shape=V10, broadcast_dims=[])
    S12 = fd.define_scalar(4096.00, dtype=DataType.Double)
    S13 = fd.ops.reciprocal(S12)
    T14 = fd.ops.mul(T11, S13)
    S15 = fd.define_scalar(1.00000e-05, dtype=DataType.Double)
    T16 = fd.ops.add(T14, S15)
    T17 = fd.ops.rsqrt(T16)
    T18 = fd.ops.cast(T17, dtype=DataType.BFloat16)
    S19 = fd.define_scalar(4096, dtype=DataType.Int)
    V20 = fd.define_vector([S19], dtype=DataType.Int)
    T21 = fd.ops.broadcast_in_dim(T18, shape=V20, broadcast_dims=[0])
    T22 = fd.ops.cast(T21, dtype=DataType.Float)
    T23 = fd.ops.mul(T6, T22)
    T24 = fd.ops.cast(T0, dtype=DataType.Float)
    T25 = fd.ops.cast(T5, dtype=DataType.Float)
    T26 = fd.ops.mul(T25, T23)
    T27 = fd.ops.mul(T25, T24)
    T28 = fd.ops.cast(T26, dtype=DataType.BFloat16)
    T29 = fd.ops.mul(T27, T22)
    T30 = fd.ops.mul(T27, T6)
    T31 = fd.ops.sum(T30, axes=[0], keepdim=False, dtype=DataType.Null)
    T32 = fd.ops.cast(T31, dtype=DataType.BFloat16)
    S33 = fd.define_scalar(1, dtype=DataType.Int)
    V34 = fd.define_vector([S33], dtype=DataType.Int)
    T35 = fd.ops.broadcast_in_dim(T32, shape=V34, broadcast_dims=[])
    T36 = fd.ops.cast(T35, dtype=DataType.Float)
    S37 = fd.define_scalar(-0.500000, dtype=DataType.Double)
    T38 = fd.ops.mul(S37, T36)
    S39 = fd.define_scalar(3.00000, dtype=DataType.Double)
    T40 = fd.ops.pow(T17, S39)
    T41 = fd.ops.mul(T38, T40)
    S42 = fd.define_scalar(4096.00, dtype=DataType.Double)
    S43 = fd.ops.reciprocal(S42)
    T44 = fd.ops.mul(T41, S43)
    T45 = fd.ops.sum(T44, axes=[0], keepdim=False, dtype=DataType.Null)
    S46 = fd.define_scalar(1, dtype=DataType.Int)
    V47 = fd.define_vector([S46], dtype=DataType.Int)
    T48 = fd.ops.broadcast_in_dim(T45, shape=V47, broadcast_dims=[])
    S49 = fd.define_scalar(4096, dtype=DataType.Int)
    V50 = fd.define_vector([S49], dtype=DataType.Int)
    T51 = fd.ops.broadcast_in_dim(T48, shape=V50, broadcast_dims=[0])
    T52 = fd.ops.mul(T51, T6)
    T53 = fd.ops.add(T29, T52)
    T54 = fd.ops.add(T53, T52)
    T55 = fd.ops.cast(T54, dtype=DataType.BFloat16)
    fd.add_output(T28)
    fd.add_output(T55)

    nvfuser_fusion_id1(fd)

inputs = [
    torch.randn((4096,), dtype=torch.bfloat16, device='cuda:0').as_strided((4096,), (1,)),
    torch.randn((4096,), dtype=torch.bfloat16, device='cuda:0').as_strided((4096,), (1,)),
]
fd.execute(inputs)
```

### PR DESCRIPTION
**Pull request by liqiangxl**
First step to fix #1631 
For more details, see the [doc](https://docs.google.com/document/d/1tZoJKmyX6EFf1OZipaIg2Ts3fm7LN6BuuyfQa94_oQg/edit?usp=sharing).

**Issue:** expr sort failed in #1631 due to the wrong computeAt position, which is due to the wrong compute at root domain map, which is due to the bug in `isReductionOutputMapped()`. The bug happens when map `11` to its consumers `13` and `12` during the build of `ComputeAtRootDomainMap`. As shown in the figure, we already have two sets:
- Post-reduction set: {4, 5, 12, 6, 8}
- Pre-reduction set: {2, 14, 13, 7}

Consumer `12` is in the post-reduction set while consumer `13` is in the pre-reduction set. So, this map is illegal!
Current check in `isReductionOutputMapped` just ignores this scenario since neither `13` nor `12` is a producer of the reduction op.
image

**Fix:** 
1. should further check if any of the consumers (`12` and `13`) is mapped to any of the domains in the reduction input set {2, 14, 13, 7}.
2. After step-1, `T11` becomes a persistent buffer, another better fix is to recompute `T13` from `T10`. Then, `T13` is no longer in the pre-reduction set. It will be put into the post-reduction set. see #1676

### DIFF
--- a/csrc/root_domain_map.cpp
+++ b/csrc/root_domain_map.cpp
@@ -447,12 +447,32 @@ bool UnmappableReductionDomains::isReductionOutputMapped(
               input_keys.end();
         });
     // None of the consumer domains is used for the reduction
-    // domain. They should be safe with respect to this reduction
-    // domain
+    // domain.
+    if (it == consumer_domains.end()) {
+      // Further check if any consumer domain is mapped with the reduction input
+      // domains
+      it = std::find_if(
+          consumer_domains.begin(),
+          consumer_domains.end(),
+          [&](const auto& consumer_domain) {
+            return std::any_of(
+                input_keys.begin(),
+                input_keys.end(),
+                [&](const auto& input_key) {
+                  return root_map.canMap(
+                      consumer_domain.td(),
+                      consumer_domain.id(),
+                      input_key.td(),
+                      input_key.id());
+                });
+          });
+    }
+    // None of the consumer domains is used for the reduction.
+    // None of the consumer domains is mapped with the reduction input domains.
+    // Safe to map.
     if (it == consumer_domains.end()) {
       continue;
     }
-
     // A consumer domain that is an input to the reduction domain
     const DomainKey& input_to_reduction = *it;
 
--- a/test/test_gpu3.cpp
+++ b/test/test_gpu3.cpp
@@ -8738,6 +8738,48 @@ TEST_F(NVFuserTest, AvoidCachingSliceInput) {
     }
   }
 }
+
+// Step-1 to fix https://github.com/NVIDIA/Fuser/issues/1631
+// Needs to check consumer mapped with reduction inputs in
+// isReductionOutputMapped(), otherwise compute at root domain map
+// is wrong then leads to wrong compute at position and finally
+// expr sort failed.
+// After fix, there are two persistent buffers and can be further
+// reduced to one with a following step-2 to fix the issue in resolution
+// points detection.
+TEST_F(NVFuserTest, CaRootDomainMapConsumerMappedWithReductionInput) {
+  std::unique_ptr<Fusion> fusion_ptr = std::make_unique<Fusion>();
+  auto fusion = fusion_ptr.get();
+  FusionGuard fg(fusion);
+
+  const int batch_size = 128;
+  const int hidden_size = 10240;
+  DataType input_dtype = DataType::Half;
+  auto tv0 = makeContigTensor(2, input_dtype);
+  auto tv1 = makeContigTensor(2, DataType::Float);
+  fusion->addInput(tv0);
+  fusion->addInput(tv1);
+  auto tv2 = castOp(DataType::Float, tv0);
+  auto tv3 = sum(tv2, {1});
+  auto tv4 = broadcast(tv3, {false, true});
+  auto tv5 = div(tv2, tv4);
+  auto tv6 = add(tv1, tv5);
+  auto tv7 = add(tv2, tv1);
+  fusion->addOutput(tv6);
+  fusion->addOutput(tv7);
+
+  auto options = at::TensorOptions()
+                     .dtype(data_type_to_aten(input_dtype))
+                     .device(at::kCUDA, 0);
+  auto options_fp32 = at::TensorOptions()
+                          .dtype(data_type_to_aten(DataType::Float))
+                          .device(at::kCUDA, 0);
+  auto t0 = at::randn({batch_size, hidden_size}, options);
+  auto t1 = at::randn({batch_size, hidden_size}, options_fp32);
+  FusionExecutorCache fec(std::move(fusion_ptr));
+  auto cg_outputs = fec.runFusionWithInputs({t0, t1});
+}
+
 // Test file size should be up to 10K LoC. Create a new file for more tests.
 
 } // namespace nvfuser

### CODE REVIEW

**Comment by liqiangxl** [comment]
!build

**Comment by liqiangxl** [comment]
!build

**Comment by jjsjann123** [comment]
Note to myself: I'll let Naoya stamp this one but I'll be shadow reviewing it as well :sunglasses:

File: csrc/root_domain_map.cpp [hunk 1]
```diff
@@ -447,12 +447,39 @@ bool UnmappableReductionDomains::isReductionOutputMapped(
...
+          [&](const auto& consumer_domain) {
+            return std::any_of(
+                input_keys.begin(),
+                input_keys.end(),
+                [&](const auto& input_key) {
+                  // check if input_key is concretized
+                  if (input_key.id()->isBroadcast()) {
+                    if (!root_map.isConcretized(
+                            input_key.td(), input_key.id())) {
+                      return false;
```
naoyam, line 466: Please add why this case should be ignored.

naoyam:
!build --diff --diff-benchmark

naoyam:
!build --diff --diff-bench

liqiangxl:
The only code diff at `MBarrierTest.Simple` is due to the recent change in `Sync after mbarrier init`

csrc/root_domain_map.cpp
liqiangxl: added.
```
                  // canMap check requires concretized IDs for broadcast domain.
                  // For example, in softmax there are two consecutive
                  // reductions, one of the inputs to the 2nd reduction has a
                  // broadcast domain and it is not concretized until its
                  // consumer is visited where it will be concretized to its
                  // consumer's non-broadcast domain.
```

naoyam: That is correct, but that means the broadcast ID is not yet processed. Not just such broadcast domains, but there's generally no guarantee that reduction input domains are already processed. If not, `canMap` just returns false. Is the analysis still valid?

liqiangxl: The analaysis can only make a decision based on the current visited exprs. Let's still use `tv11` and its two consumers `tv12` and `tv13` as the example. If we are trying to do this mapping before `tv2 --> {tv13, tv14}` is visited. (This is similar to what you mensioned, where we didn't notice one of the consumer can be mapped with a reduction input) We will find it is OK to map  `tv11` with its two consumers.

Then the analysis continues,  it later finds a tensor `tv2` is an input to reduction domain and also used by `tv13`, this means we can't map `tv2` with its consumers. and `tv2` becomes a persistent tensor.

The generated code may be different if the visiting order is different. This is actually an example I put in the [doc](https://docs.google.com/document/d/1tZoJKmyX6EFf1OZipaIg2Ts3fm7LN6BuuyfQa94_oQg/edit#heading=h.5g2dpdnbafwu).

liqiangxl: if we want to guarantee the commutative law in fusion definition, e.g. `auto tv6 = add(tv1, tv5);` is always equivalent to ` auto tv6 = add(tv5, tv1);`. Looks like there is something needs to be done.

naoyam: > The analaysis can only make a decision based on the current visited exprs. Let's still use `tv11` and its two consumers `tv12` and `tv13` as the example. If we are trying to do this mapping before `tv2 --> {tv13, tv14}` is visited. (This is similar to what you mensioned, where we didn't notice one of the consumer can be mapped with a reduction input) We will find it is OK to map `tv11` with its two consumers.
> 
> Then the analysis continues, it later finds a tensor `tv2` is an input to reduction domain and also used by `tv13`, this means we can't map `tv2` with its consumers. and `tv2` becomes a persistent tensor.
> 
> The generated code may be different if the visiting order is different. This is actually an example I put in the [doc](https://docs.google.com/document/d/1tZoJKmyX6EFf1OZipaIg2Ts3fm7LN6BuuyfQa94_oQg/edit#heading=h.5g2dpdnbafwu).

That's correct. Please add that to the code as a comment.

csrc/root_domain_map.cpp
```diff
@@ -447,12 +447,44 @@ bool UnmappableReductionDomains::isReductionOutputMapped(
               input_keys.end();
         });
     // None of the consumer domains is used for the reduction
-    // domain. They should be safe with respect to this reduction
-    // domain
+    // domain.
+    if (it == consumer_domains.end()) {
+      // Further check if any consumer domain is mapped with the reduction input
+      // domains
+      it = std::find_if(
```
naoyam, line 454: Let's merge this `std::find_if` with the above `std::find`

test/test_gpu3.cpp
```diff
@@ -8738,6 +8738,48 @@ TEST_F(NVFuserTest, AvoidCachingSliceInput) {
...
+
+  auto options = at::TensorOptions()
+                     .dtype(data_type_to_aten(input_dtype))
+                     .device(at::kCUDA, 0);
+  auto options_fp32 = at::TensorOptions()
+                          .dtype(data_type_to_aten(DataType::Float))
+                          .device(at::kCUDA, 0);
+  auto t0 = at::randn({batch_size, hidden_size}, options);
+  auto t1 = at::randn({batch_size, hidden_size}, options_fp32);
+  FusionExecutorCache fec(std::move(fusion_ptr));
```
naoyam, line 8779: Please do a more explicit validation. Here, the CA position should be validated. Or, even better is just building `ComputeAtRootDomain` and check the relevant domains are not mapped.

Also, `runFusionWithInputs` does not guarantee to reproduce the original issue. Can we just manually schedule the fusion? Or at least the normalization scheduler should be directly used.

csrc/root_domain_map.cpp
liqiangxl: merged.

test/test_gpu3.cpp
liqiangxl: checked with `ComputeAtRootDomainMap`. I just manually projected tv2 in `auto tv6 = div(tv2, tv4);` to input to trigger the bug.

liqiangxl:
!build

test/test_gpu3.cpp
```diff
@@ -8738,6 +8738,82 @@ TEST_F(NVFuserTest, AvoidCachingSliceInput) {
...
+  fusion->addOutput(tv8);
+
+  // |--5------------------|
+  // 0 -> 2 -> r3 -> b4 -> 6 -> 7    8
+  // 1--------------------------|
+  // |-------------------------------|
+  //      |--------------------------|
+  // tv1 has two consumers, tv7 and tv8.
+  // tv7 is a consumer of the reduction output.
+  // tv8 is not a reduction input but it is mapped with a reduction input, tv2.
```
naoyam, line 8780: Shouldn't this be:

```
If tv8 is mapped with `tv2`, we can't map tv7 and tv8...
```

test/test_gpu3.cpp
```diff
@@ -8738,6 +8738,82 @@ TEST_F(NVFuserTest, AvoidCachingSliceInput) {
...
+        "Should not be mappable: ",
+        tva->getRootDomain().at(axis),
+        " of ",
+        tva,
+        " and ",
+        tvb->getRootDomain().at(axis),
+        " of ",
+        tvb);
+  };
+  shouldNotMapCheck(tv1, tv7, 1);
```
naoyam, line 8802: `tv1` is an input. I don't remember `ComputeAtRootDomainMap` does anything special for inputs, but mapping of inputs doesn't make much sense as there's no expression to produce them. Maybe just do `cacheAfter`?

test/test_gpu3.cpp
liqiangxl: revised to:
```

  // |--5------------------|
  // 0 -> 2 -> r3 -> b4 -> 6 -> 8    9
  // 1-------> 7----------------|
  //           |---------------------|
  //      |--------------------------|
  // tv7 has two consumers, tv8 and tv9.
  // tv8 is a consumer of the reduction output.
  // if tv9 is mapped with tv2, we can't map tv8 and tv9.
  // Because tv9 is in the pre-reduction set and tv8 is in the post-reduction set.
```

test/test_gpu3.cpp
```diff
@@ -8738,6 +8738,83 @@ TEST_F(NVFuserTest, AvoidCachingSliceInput) {
...
+  // |--5------------------|
+  // 0 -> 2 -> r3 -> b4 -> 6 -> 8    9
+  // 1-------> 7----------------|
+  //           |---------------------|
+  //      |--------------------------|
+  // tv7 has two consumers, tv8 and tv9.
+  // tv8 is a consumer of the reduction output.
+  // if tv9 is mapped with tv2, we can't map tv8 and tv9.
+  // Because tv9 is in the pre-reduction set and tv8 is in the post-reduction
+  // set.
```
naoyam, line 8783:
```suggestion
  // If tv9 is mapped with tv2, we can't map tv8 and tv9 because tv9 is in the pre-reduction set through tv2 and tv8 is in the post-reduction
  // set.
```

test/test_gpu3.cpp
```diff
@@ -8738,6 +8738,83 @@ TEST_F(NVFuserTest, AvoidCachingSliceInput) {
...
+        tva->getRootDomain().at(axis),
+        " of ",
+        tva,
+        " and ",
+        tvb->getRootDomain().at(axis),
+        " of ",
+        tvb);
+  };
+  shouldNotMapCheck(tv7, tv8, 1);
+  shouldNotMapCheck(tv7, tv9, 1);
```
naoyam, line 8804: Also assert that `tv9` is indeed mapped with `tv2`.

test/test_gpu3.cpp
```diff
@@ -8738,6 +8738,83 @@ TEST_F(NVFuserTest, AvoidCachingSliceInput) {
...
+
+  auto options = at::TensorOptions()
```
naoyam, line 8806: Let's remove below here. The above check should be sufficient.

naoyam (approved):
LGTM. Thanks for the fix.

csrc/root_domain_map.cpp
```diff
@@ -437,22 +437,47 @@ bool UnmappableReductionDomains::isReductionOutputMapped(
...
+                    return false;
+                  }
+                }
+                return root_map.canMap(
+                    consumer_domain.td(),
+                    consumer_domain.id(),
+                    input_key.td(),
+                    input_key.id());
+              });
+          return is_reduction_input || is_mapped_with_reduction_input;
```
jjsjann123, line 472: nitpick: we can early return here, instead of computing both flag ahead of time.

test/test_gpu3.cpp
```diff
@@ -8738,6 +8738,74 @@ TEST_F(NVFuserTest, AvoidCachingSliceInput) {
...
+  // 1-------> 7----------------|
+  //           |---------------------|
+  //      |--------------------------|
+  // tv7 has two consumers, tv8 and tv9.
+  // tv8 is a consumer of the reduction output.
+  // If tv9 is mapped with tv2, we can't map tv8 and tv9 because tv9 is in the
+  // pre-reduction set through tv2 and tv8 is in the post-reduction set.
+  ComputeAtRootDomainMap root_map;
+  root_map.build();
+  auto shouldMapCheck =
```
jjsjann123, line 8783: nice test!

nitpick: this feels pretty big for a lambda. Would be nice to have a separate test file for `ComputeAtRootDomainMap` and having this as a separate util function. But we don't have to do that in this PR.

jjsjann123 (commented):
:clap:

test/test_gpu3.cpp
liqiangxl: I think it looks big just because of the format of err msg. The actual code is just calling `root_map.canMap()`. Will consider a separate test file for `ComputeAtRootDomainMap` when we have more tests, or maybe moving all the existing related tests.

liqiangxl: I'll do this in a seperate PR.

PR #1763, opened 2024-02-14
Rename `AllocationTypes`

Priya2698:
This PR renames the `AllocationTypes`:

1. `NoAlias` -> `New`
2. `InplaceUpdate` -> `ReuseBuffer`
~3. `PointerArithmetic` and `Evaluate`-> `Evaluate`~

~Logic for `PointerArithmetic` and `Evaluate` is merged which removes additional code (`markOutputForEvaluation`) used for adding outputs of type `Evaluate` in PR #1743.~. This was done in PR #1775

### DIFF
--- a/csrc/fusion.cpp
+++ b/csrc/fusion.cpp
@@ -772,8 +772,8 @@ void Fusion::aliasOutputToInput(
     Val* input,
     const AllocationType type) {
   NVF_CHECK(
-      type != AllocationType::NoAlias,
-      "NoAlias is returned automatically for a missing key. Don't add it explicitly.");
+      type != AllocationType::New,
+      "New is returned automatically for a missing key. Don't add it explicitly.");
 
   if (type == AllocationType::Evaluate) {
     NVF_CHECK(
@@ -784,7 +784,7 @@ void Fusion::aliasOutputToInput(
     return;
   }
 
-  NVF_ERROR(type == AllocationType::InplaceUpdate);
+  NVF_ERROR(type == AllocationType::ReuseBuffer);
   // `input` can be a cast of a fusion input.
   if (!input->isFusionInput()) {
     auto input_expr = input->definition();
@@ -823,7 +823,7 @@ void Fusion::aliasOutputToInput(
 
 const AliasInfo& Fusion::getOutputAlias(const Val* output) const {
   static AliasInfo no_alias_info{
-      .type = AllocationType::NoAlias,
+      .type = AllocationType::New,
       .aliased_io = nullptr,
       .hide_output = false};
   if (auto search = io_alias_.find(output); search != io_alias_.end()) {
--- a/csrc/fusion.cpp
+++ b/csrc/fusion.cpp
@@ -823,9 +823,7 @@ void Fusion::aliasOutputToInput(
 
 const AliasInfo& Fusion::getOutputAlias(const Val* output) const {
   static AliasInfo no_alias_info{
-      .type = AllocationType::New,
-      .aliased_io = nullptr,
-      .hide_output = false};
+      .type = AllocationType::New, .aliased_io = nullptr, .hide_output = false};
   if (auto search = io_alias_.find(output); search != io_alias_.end()) {
     return search->second;
   }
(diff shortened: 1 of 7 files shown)


csrc/fusion.cpp
```diff
@@ -797,28 +797,37 @@ void Fusion::aliasOutputToInput(
...
-      "The input and output values are not alias-compatible.");
-  // Let integration hide any output that wasn't a fusion output when
-  // `aliasOutputToInput` was called. For example, running mean and var for
-  // batch norm.
-  io_alias_[output] = AliasInfo{
-      .type = type,
-      .aliased_io = input,
-      .hide_output = !output->isFusionOutput()};
+  // AllocationType::Evaluate
+  if (!input) { // Non-aliased outputs marked for expression evaluator.
```
wujingyue, line 801:
```suggestion
  if (input == nullptr) { // Non-aliased outputs marked for expression evaluator, e.g., a matmul running in fallback mode.
```
so it's immediately clear from this line that we are checking a pointer not a boolean value.

csrc/fusion.cpp
```diff
@@ -797,28 +797,37 @@ void Fusion::aliasOutputToInput(
...
+    io_alias_[output] =
+        AliasInfo{.type = type, .aliased_io = nullptr, .hide_output = false};
+  }
 
-  // TODO: output should be marked at the end of fusion definition #1488
-  if (!output->isFusionOutput()) {
-    addOutput(output);
+  else { // Pointer Arithmetic Outputs
+    NVF_ERROR(
+        isAliasCompatible(input, output),
```
wujingyue, line 811: Nit: if it helps to simplify, feel free to call this only for `ReuseBuffer`, which is what it was written for anyway. I tried to remove some checks there to accommodate pointer-arithmetic aliases, but these checks are minimal and can break in the future, e.g., we could bitcast a tensor to a different data type.

csrc/fusion.cpp
```diff
@@ -797,28 +797,37 @@ void Fusion::aliasOutputToInput(
...
-      isAliasCompatible(input, output),
-      "The input and output values are not alias-compatible.");
-  // Let integration hide any output that wasn't a fusion output when
-  // `aliasOutputToInput` was called. For example, running mean and var for
-  // batch norm.
-  io_alias_[output] = AliasInfo{
-      .type = type,
-      .aliased_io = input,
-      .hide_output = !output->isFusionOutput()};
+  // AllocationType::Evaluate
```
wujingyue, line 800: This is not accurate. `::ReuseBuffer` reaches here too, and that's intended.

csrc/fusion.cpp
```diff
@@ -797,28 +797,37 @@ void Fusion::aliasOutputToInput(
...
+  if (!input) { // Non-aliased outputs marked for expression evaluator.
+    NVF_CHECK(
+        output->isFusionOutput(),
+        "Only fusion outputs can be expression evaluated.");
+    io_alias_[output] =
+        AliasInfo{.type = type, .aliased_io = nullptr, .hide_output = false};
+  }
```
wujingyue, line 807: The current logic looks like:
```
if (reuse) {
  ...
}
if (alias) {
  change io_alias_
} else {
  change io_alias_
}
```

I suggest the following simplification:
```
if (reuse) {
  everything we need for reuse, e.g., hide_output, alias compatibility check, and change io_alias_. 
  return;  // early exit so the reader has only one type to reason about for the rest of the function 
}

assert type is evaluate;
assert output is a fusion output; 
// As I mentioned in the other comment, you could check isAliasCompatible, but IMHO the benefits of the minimal and maybe fragile check over there is not worth the added code complexity. 
add {type, input, false};
```

wujingyue (commented):
Nice! LGTM with comments

csrc/fusion.cpp
Priya2698: Right. Will move it.

csrc/fusion.cpp
Priya2698:
```
if (!output->isFusionOutput()) {
      addOutput(output);
    }
``` 
What cases does this cover?

Priya2698:
Converting to draft until PR #1743 is patched.

wujingyue:
!build

PR #1764, opened 2024-02-14
patch setAllocationDomain on rank-0 TensorView

jjsjann123:
Setting empyt allocatoin domain on rank-0 tensor shouldn't trigger assert.

### DIFF
--- a/csrc/ir/utils.cpp
+++ b/csrc/ir/utils.cpp
@@ -959,6 +959,10 @@ std::vector<Statement*> next(Statement* stmt) {
 void validateDomainEquivalence(
     const std::vector<IterDomain*>& initial_domain,
     const std::vector<IterDomain*>& derived_domain) {
+  // empty domain are equivalent.
+  if (initial_domain.empty() && derived_domain.empty()) {
+    return;
+  }
   ValidateDomainEquivalence(initial_domain, derived_domain);
 }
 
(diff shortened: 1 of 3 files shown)


jjsjann123:
run into this surprising assert in #1755

wujingyue (approved):
Thank you! Can you do me a favor to remove https://github.com/NVIDIA/Fuser/blob/a0cb47aa1be4432f00292325d1b50a35e7169aaa/csrc/optimization/mark_aliases_prepare.cpp#L52-L56, which I believe is unnecessary with this PR?

csrc/ir/utils.cpp
```diff
@@ -959,6 +959,10 @@ std::vector<Statement*> next(Statement* stmt) {
 void validateDomainEquivalence(
     const std::vector<IterDomain*>& initial_domain,
     const std::vector<IterDomain*>& derived_domain) {
+  // empty domain are equivalent.
+  if (initial_domain.empty() && derived_domain.empty()) {
```
naoyam, line 963: Thanks for the fix. Could you do this at around line 830 instead?

naoyam (approved):
Thanks for the fix!

jjsjann123:
!build

jjsjann123:
!build

Issue #1765, opened 2024-02-15
Compute validation references in `double` in python benchmarks


Priya2698:
To reduce validation errors, all the reference computations should be done in `double`, particularly for `backward` benchmarks.
CC: @naoyam

PR #1766, opened 2024-02-15
Compute validation reference in `double` [1]

Priya2698:
Issue #1765.

Compute references in `double`, starting with the benchmarks where we have existing validation errors in CI.

### DIFF
--- a/python_benchmarks/normalization.py
+++ b/python_benchmarks/normalization.py
@@ -340,23 +340,23 @@ def norm_bwd_benchmark(
         # PyTorch expects running mean and variance to be of same type as input.
         if norm == "batch_norm":
             eager_output = torch.nn.functional.batch_norm(
-                at_inputs,
-                running_mean.to(dtype),
-                running_var.to(dtype),
-                weight=weight,
-                bias=bias,
+                at_inputs.to(torch.double),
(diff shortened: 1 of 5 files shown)


naoyam (approved):
LGTM. Do these change help reduce validation errors?

Priya2698:
> LGTM. Do these change help reduce validation errors?

It does get rid of some errors on my local, but then, the errors are somewhat unreliable and higher on the CI. 
I have requested @xwang233  to run this PR through the CI.

Note that, this may (most likely) not get rid of all validation errors since benchmarks such as `layernorm_bwd` already have references in fp64 but we get these errors. But this is a simple first step, in reducing the test failures on the CI.

Priya2698:
> > LGTM. Do these change help reduce validation errors?
> 
> It does get rid of some errors on my local, but then, the errors are somewhat unreliable and higher on the CI. I have requested @xwang233 to run this PR through the CI.
> 
> Note that, this may (most likely) not get rid of all validation errors since benchmarks such as `layernorm_bwd` already have references in fp64 but we get these errors. But this is a simple first step, in reducing the test failures on the CI.

We still have existing validation errors in the CI. However, this PR does reduce the number of instances.

PR #1775, opened 2024-02-16
Matmul default scheduling 

Priya2698:
PR #1743 was reverted due to the following issues:
1. Matmul scheduler does not support all architectures: This caused some tests to fail on V100. This is fixed by returning early in `getMatmulHeuristics` and only setting the required parameters.
2. Errors in `GpuLower::analysis`: `validateMma` and `PredicateElimination` are modified to skip expressions and outputs marked for expression evaluator.

### DIFF
--- a/csrc/device_lower/analysis/predicate_elimination.cpp
+++ b/csrc/device_lower/analysis/predicate_elimination.cpp
@@ -863,6 +863,8 @@ class PredicateChcker : public IterVisitor {
 } // namespace
 
 PredicateElimination::PredicateElimination(Fusion* fusion) {
+  // To avoid errors in analysis when using ATen evaluation for matmul, only use
+  // outputs that require codegen. See PR # 1775 and Issue #1812
   traverseTo(fusion->getFusionOutputsRequiringCodegen());
 }
 
--- a/csrc/device_lower/analysis/predicate_elimination.cpp
+++ b/csrc/device_lower/analysis/predicate_elimination.cpp
@@ -863,18 +863,7 @@ class PredicateChcker : public IterVisitor {
 } // namespace
 
 PredicateElimination::PredicateElimination(Fusion* fusion) {
-  // TODO: Common code as in expr_sort.cpp.
-  // Move to a helper function
-  std::vector<Val*> outs_requiring_codegen;
-  outs_requiring_codegen.reserve(fusion->outputs().size());
-  std::copy_if(
-      fusion->outputs().begin(),
-      fusion->outputs().end(),
-      std::back_inserter(outs_requiring_codegen),
-      [&fusion](Val* out) {
-        return (fusion->getOutputAlias(out).type != AllocationType::Evaluate);
-      });
-  traverseTo(outs_requiring_codegen);
+  traverseTo(fusion->getFusionOutputsRequiringCodegen());
 }
 
 bool PredicateElimination::needsPredicate(Expr* expr) const {
--- a/csrc/device_lower/analysis/predicate_elimination.cpp
+++ b/csrc/device_lower/analysis/predicate_elimination.cpp
@@ -865,7 +865,7 @@ class PredicateChcker : public IterVisitor {
 PredicateElimination::PredicateElimination(Fusion* fusion) {
   // To avoid errors in analysis when using ATen evaluation for matmul, only use
   // outputs that require codegen. See PR # 1775 and Issue #1812
-  traverseTo(fusion->getFusionOutputsRequiringCodegen());
+  traverseTo(lower_utils::getFusionOutputsRequiringCodegen(fusion));
 }
 
 bool PredicateElimination::needsPredicate(Expr* expr) const {
--- a/csrc/device_lower/analysis/predicate_elimination.cpp
+++ b/csrc/device_lower/analysis/predicate_elimination.cpp
@@ -863,7 +863,18 @@ class PredicateChcker : public IterVisitor {
 } // namespace
 
 PredicateElimination::PredicateElimination(Fusion* fusion) {
-  traverseTo(fusion->outputs());
+  // TODO: Common code as in expr_sort.cpp.
+  // Move to a helper function
+  std::vector<Val*> outs_requiring_codegen;
+  outs_requiring_codegen.reserve(fusion->outputs().size());
+  std::copy_if(
+      fusion->outputs().begin(),
+      fusion->outputs().end(),
+      std::back_inserter(outs_requiring_codegen),
+      [&fusion](Val* out) {
+        return (fusion->getOutputAlias(out).type != AllocationType::Evaluate);
+      });
+  traverseTo(outs_requiring_codegen);
 }
 
 bool PredicateElimination::needsPredicate(Expr* expr) const {
--- a/csrc/device_lower/pass/expr_sort.cpp
+++ b/csrc/device_lower/pass/expr_sort.cpp
@@ -1516,7 +1516,9 @@ void ExprSegmentationSorter::sort() {
       std::back_inserter(non_pointer_arithmetic_outs),
       [this](Val* out) {
         return fusion_->getOutputAlias(out).type !=
-            AllocationType::PointerArithmetic;
+            AllocationType::PointerArithmetic &&
+            fusion_->getOutputAlias(out).type !=
+            AllocationType::Evaluate;
       });
 
   // Not putting the exprs between fusion inputs and allKnownVals() here
--- a/csrc/device_lower/pass/expr_sort.cpp
+++ b/csrc/device_lower/pass/expr_sort.cpp
@@ -1506,14 +1506,15 @@ void ExprSegmentationSorter::sort() {
   // Need this for initialization of the DAG that is processed
   std::unordered_map<Expr*, ExprGroup*> expr2group;
 
-  // We skip generating code that computes only pointer-arithmetic outputs.
+  // We skip generating code that computes only pointer-arithmetic outputs
+  // or any output marked for expression evaluator (AllocationType::Evaluate).
   // Those outputs will be computed by ExpressionEvaluator.
-  std::vector<Val*> non_pointer_arithmetic_outs;
-  non_pointer_arithmetic_outs.reserve(fusion_->outputs().size());
+  std::vector<Val*> non_expr_evaluator_outs;
+  non_expr_evaluator_outs.reserve(fusion_->outputs().size());
   std::copy_if(
       fusion_->outputs().begin(),
       fusion_->outputs().end(),
-      std::back_inserter(non_pointer_arithmetic_outs),
+      std::back_inserter(non_expr_evaluator_outs),
       [this](Val* out) {
         return (fusion_->getOutputAlias(out).type !=
             AllocationType::PointerArithmetic &&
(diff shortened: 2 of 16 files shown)


Priya2698:
!build

csrc/device_lower/analysis/predicate_elimination.cpp
```diff
@@ -863,7 +863,18 @@ class PredicateChcker : public IterVisitor {
 } // namespace
 
 PredicateElimination::PredicateElimination(Fusion* fusion) {
-  traverseTo(fusion->outputs());
+  // TODO: Common code as in expr_sort.cpp.
```
Priya2698, line 866: @wujingyue I want to move this to a common function in `fusion.cpp`. What do you think about merging `PointerArithmetic` and `Evaluate` in this PR itself?

csrc/device_lower/pass/expr_sort.cpp
```diff
@@ -1506,23 +1506,26 @@ void ExprSegmentationSorter::sort() {
   // Need this for initialization of the DAG that is processed
   std::unordered_map<Expr*, ExprGroup*> expr2group;
 
-  // We skip generating code that computes only pointer-arithmetic outputs.
+  // We skip generating code that computes only pointer-arithmetic outputs
+  // or any output marked for expression evaluator (AllocationType::Evaluate).
   // Those outputs will be computed by ExpressionEvaluator.
-  std::vector<Val*> non_pointer_arithmetic_outs;
-  non_pointer_arithmetic_outs.reserve(fusion_->outputs().size());
+  std::vector<Val*> outs_requiring_codegen;
```
Priya2698, line 1512: This segment of the code can be moved to a utility function in `fusion.cpp` to be used here and in `PredicateElimination`.

csrc/device_lower/analysis/predicate_elimination.cpp
```diff
@@ -863,7 +863,18 @@ class PredicateChcker : public IterVisitor {
...
+  std::vector<Val*> outs_requiring_codegen;
+  outs_requiring_codegen.reserve(fusion->outputs().size());
+  std::copy_if(
+      fusion->outputs().begin(),
+      fusion->outputs().end(),
+      std::back_inserter(outs_requiring_codegen),
+      [&fusion](Val* out) {
+        return (fusion->getOutputAlias(out).type != AllocationType::Evaluate);
+      });
+  traverseTo(outs_requiring_codegen);
```
Priya2698, line 877: This removes outputs marked for EE from predicate elimination to avoid errors in lowering analysis.

csrc/scheduler/matmul.cpp
```diff
@@ -30,6 +30,17 @@ MatmulScheduler::MatmulScheduler(
 
 void MatmulScheduler::schedule(Fusion* fusion) {
   FUSER_PERF_SCOPE("Schedule Matmul Fusion");
+  // Skip scheduling if Matmul will be expression evaluated.
```
Priya2698, line 33: Moving this from `scheduleMatmul` to `schedule`.

csrc/scheduler/matmul_utils.cpp
```diff
@@ -361,6 +361,13 @@ std::shared_ptr<MatmulParams> getMatmulHeuristics(
   (void)runtime_info;
   auto params = std::make_shared<MatmulParams>();
 
+  // Set kernel index mode
+  params->cparams.index_type = runtime_info.getIndexType();
```
Priya2698, line 365: `index_type` is needed in `compileFusion` and needs to be set before returning. This will avoid errors on architectures not supported by matmul scheduler (see `getMmaOp` function.)

Priya2698:
!build

csrc/options.h
```diff
@@ -86,6 +86,7 @@ enum class EnableOption {
   MemoryPromotion, //! Enable promotion of memory types for non-pointwise ops
   StaticFusionCount, //! Enable using single static count in kernel name
   WarnRegisterSpill, //! Enable warnings of register spill
+  MatmulExprEval, //! Enable ATen evaluation for Matmul
```
jjsjann123, line 89: I think we need to make a louder statement here regarding the impact of this flag.

Enabling `MatmulExprEval` means that we are running the entire fusion containing a matmul with expression evaluation, not just the matmul portion.

csrc/scheduler/matmul.cpp
```diff
@@ -30,6 +30,17 @@ MatmulScheduler::MatmulScheduler(
...
+  if (isOptionEnabled(EnableOption::MatmulExprEval)) {
+    NVF_CHECK(fusion->outputs().size() == 1)
```
jjsjann123, line 35: So right now we are asserting on a single output from the fusion? cc'ing @protonu This might be a sharp edge that you'd want to keep in mind.

Priya2698:
!build

Priya2698:
!build

Priya2698:
!build

csrc/fusion.h
```diff
@@ -261,8 +261,16 @@ class NVF_API Fusion : public IrContainer {
...
+  //! Returns the expressions that require codegen.
+  //! Expressions where all the outputs are marked to be computed through
+  //! expression evaluator are filtered out.
   std::vector<Expr*> getExprsToCodegen();
 
+  //! Returns the fusion outputs that require codegen.
+  //! The fusion outputs to be computed through expression evaluator are
+  //! filtered out.
+  std::vector<Val*> getFusionOutputsRequiringCodegen();
+
```
wujingyue: Move these two methods outside the class, i.e., 

```
std::vector<Val*> getExprsToCodegen(Fusion* fusion);
std::vector<Val*> getFusionOutputsRequireingCodegen(Fusion* fusion);
```

Reason: these two functions don't use and shouldn't need to use any private fields of Fusion. Making them class methods allows future writers to misuse private fields, making it harder to remove/change these fields in the future.

csrc/fusion.cpp
```diff
@@ -828,4 +837,34 @@ bool Fusion::hasDynamicTransform() {
...
+  for (auto expr : exprs()) {
+    bool is_expr_codegen = std::none_of(
+        expr->outputs().begin(), expr->outputs().end(), [this](Val* out) {
+          return this->getOutputAlias(out).type != AllocationType::Evaluate;
+        });
+
+    if (is_expr_codegen) {
+      exprs_requiring_codegen.emplace_back(expr);
+    }
+  }
```
wujingyue, line 853: I don't follow the implementation. Shouldn't we return all exprs between inputs and codegen outputs like https://github.com/NVIDIA/Fuser/blob/925acc63e5870718ab83de911ec7f85d6126f729/csrc/device_lower/pass/expr_sort.cpp#L1524-L1525?

csrc/executor.cpp
```diff
@@ -977,10 +977,16 @@ at::Tensor allocateOutput(
     return alloc_tensor;
   }
 
+  if (alias_info.type == AllocationType::Evaluate && !alias_info.aliased_io) {
+    // Compute non-aliased outputs that were marked for expr evaluator.
+    at::Tensor out_tensor = ee.evaluate(out_tv).as<at::Tensor>();
```
wujingyue, line 982:
```suggestion
    auto out_tensor = ee.evaluate(out_tv).as<at::Tensor>();
```

`as<at::Tensor>` is already clear about the type, so using `auto` as the return type is fine and encouraged.

csrc/executor.cpp
```diff
@@ -977,10 +977,16 @@ at::Tensor allocateOutput(
     return alloc_tensor;
   }
 
+  if (alias_info.type == AllocationType::Evaluate && !alias_info.aliased_io) {
```
wujingyue, line 980: The handling of `AllocationType::Evaluate` spread at two places, which is a bit hard to follow. Wdyt about the following? 

```
optional<at::Tensor> aliased_io_tensor = std::nullopt;
if (alias_info.aliased_io != nullptr) {
  aliased_io_tensor = ...
}

switch (alias_info.type) {
  case NoAlias:
    allocate the tensor and return it;
  case InplaceUpdate:  // I thought we wanted to rename it to ReuseBuffer, but fine with a separate PR to address that. 
    return *aliased_io_tensor;
  case Evaluate:
    out_tensor = ...
    if (aliased_io_tensor.has_value()) {
      verify indeed an alias
    }
    return out_tensor;
}
```

csrc/fusion.cpp
wujingyue: I don't think this has been addressed in the latest version. IIUC, we need to return all exprs between inputs and codegen outputs instead of all exprs that define a codegen output. Wdyt?

csrc/scheduler/matmul.cpp
protonu: Thanks!

csrc/fusion.cpp
wujingyue: There's a difference between an Expr producing an EE output directly (as in your current code) and an Expr producing that indirectly (which is I think the right way). E.g. in

```
b = e0(a) // say codegen out
c = e1(b) // intermediate
d = e2(c) // say ee out
```

we do want to filter out `e1` even though its output is not **directly** marked for EE.

Priya2698: Got it, thanks for the example.

Priya2698:
!build

csrc/executor.cpp
```diff
@@ -963,56 +963,60 @@ at::Tensor allocateOutput(
...
+            "ExpressionEvaluator failed to evaluate ",
+            out_tv->toString(),
+            " as an alias of ",
+            aliased_io->toString());
+        inferAndValidateAllocationSizesAndStrides(out_tensor, out_tv, ee);
+      }
+      return out_tensor;
+    }
+    default:
+      NVF_ERROR(false, "Unrecognized AllocationType.");
```
wujingyue, line 1018: I doubt this is needed. Compilers today should be smart enough to figure out the above cases have covered all possible options.

csrc/fusion.cpp
```diff
@@ -794,23 +794,32 @@ void Fusion::aliasOutputToInput(
...
+        .hide_output = !output->isFusionOutput()};
+
+    // TODO: output should be marked at the end of fusion definition #1488
+    if (!output->isFusionOutput()) {
+      addOutput(output);
+    }
+    return;
   }
+
+  // AllocationType::Evaluate
```
wujingyue, line 816:
```suggestion
```

The NVF_ERROR check below is pretty clear about what this region tries to handle.

csrc/fusion.cpp
```diff
@@ -794,23 +794,32 @@ void Fusion::aliasOutputToInput(
...
+  NVF_ERROR(type == AllocationType::Evaluate);
+  NVF_CHECK(
+      output->isFusionOutput(),
+      "Only fusion outputs can be expression evaluated.");
+  io_alias_[output] =
```
wujingyue, line 821: Nit: I think code will be cleaner if it tries to handle Evaluate first. That is, instead of 

```
if (inplace-update) {
  long code
  return;
}
short code
```
do 
```
if (evaluate) {
  short code
  return;
}
long code
```

This way, readers quickly cross out one case after reading "short code" instead of "long code".

csrc/scheduler/matmul.cpp
```diff
@@ -30,6 +30,18 @@ MatmulScheduler::MatmulScheduler(
...
+    fusion->aliasOutputToInput(
+        fusion->outputs()[0], nullptr, AllocationType::Evaluate);
```
wujingyue, line 37:
```suggestion
        fusion->outputs()[0], /*input=*/nullptr, AllocationType::Evaluate);
```

wujingyue (commented):
LGTM with comments

csrc/executor.cpp
Priya2698: Added this to avoid the error: `error: control reaches end of non-void function` even though the switch-case is completely handled.

wujingyue (approved):

Priya2698:
!build

Issue #1768, opened 2024-02-15
Remove unnecessary concatenation using the zipping approach. 


wujingyue:
A spin-off from https://github.com/NVIDIA/Fuser/issues/1502#issuecomment-1870837878. Created for tracking progress. 

# Problem

Below is a common pattern in nanoGPT's backprop. 

```
dQ, dK, dV = scaled_dot_product_attention_backprop(...)  # bf16[16,12,128,64], bf16[16,12,128,64], bf16[16,12,128,64]
dQ = transpose(dQ, [0, 2, 1, 3])  # [16, 128, 12, 64]
dQ = reshape(dQ, [16, 128, 768])  # [16, 128, 768]
dK = the product of the same transpose and reshape on dK
dV = the product of the same transpose and reshape on dV

concatenated = cat([dQ, dK, dV], axis=-1)

dQKV_sum = sum(concatenated, ...)  # omitting a round trip to float
dQKV_view = reshape(concatenated, [B*S, H*D*3])
dQKV_permute = transpose(dQKV_view, [1, 0])

return dQKV_sum, dQKV_view, dQKV_permute
```

Because nvFuser doesn't take sdpa_backward and therefore sees three unconnected input tensors (dQ, dK, and dV), it has to materialize `dQKV_view` and `dQKV_permute`. 

# Solution

TL;DR: change Thunder's cudnnex to feed nvFuser a concatenated tensor that contains dQ, dK and dV, so nvFuser realizes that the existing cat is unnecessary and removes it. 

1. cudnnex will convert the SDPA backward op into a cudnn `spda_backward` kernel (which outputs one dQKV tensor) followed by a split. 
2. cudnnex will give that split to nvFuser, so nvFuser will see the following pattern:
   ```
   dQKV = fd.ops.define_tensor([B, S, H, D*3])
   dQ = fd.ops.slice(dQKV, ...)
   dQ = fd.ops.view(dQ, ...)
   dQ = fd.ops.permute(dQ, ...)
   dK = ...the same slice-view-permute pattern
   dV = ...the same slice-view-permute pattern

   concatenated = fd.ops.cat([dQ, dK, dV], axis=-1)

   dQKV_sum = fd.ops.sum(concatenated, ...)  # omitting a round trip to float
   dQKV_view = fd.ops.view(concatenated, [B*S, H*D*3])
   dQKV_permute = fd.ops.permute(dQKV_view, [1, 0])
   ```
[...]

PR #1782, opened 2024-02-17
Optimize split->permute->cat. 

wujingyue:
With this PR, MoveSplitCatPass can cancel the <split,cat> pair with `permute`s in between and horizontally merge those `permute`s. See code comments for details. 

For #1768.

### DIFF
--- a/test/test_resize.cpp
+++ b/test/test_resize.cpp
@@ -3352,4 +3352,55 @@ TEST_F(ResizeTest, CancellableSplitAndCat_PermuteInBetween) {
   EXPECT_TRUE(out_tensors[0].is_alias_of(in_tensor));
 }
 
+TEST_F(ResizeTest, NoncancellableSplitAndCat_SomeArePermuted) {
+  auto fusion = std::make_unique<Fusion>();
+  FusionGuard fg(fusion.get());
+
+  TensorView* in = makeContigConcreteTensor({2, 2, 10});
+  TensorView* s0 = slice(in, {0, 0, 0}, {2, 2, 2});
+  TensorView* s1 = slice(in, {0, 0, 2}, {2, 2, 5});
+  TensorView* s2 = slice(in, {0, 0, 5}, {2, 2, 10});
+  s0 = permute(s0, {1, 0, 2});
+  s2 = permute(s2, {1, 0, 2});
+  TensorView* out = cat({s0, s1, s2}, /*dim=*/-1);
+
+  fusion->addInput(in);
+  fusion->addOutput(out);
+
+  auto options = at::TensorOptions().dtype(at::kFloat).device(at::kCUDA, 0);
+  at::Tensor in_tensor = at::randn({2, 2, 10}, options);
+
+  FusionExecutorCache fec(std::move(fusion));
+  auto out_tensors = fec.runFusionWithInputs({in_tensor});
+  testValidate(fec.fusion(), out_tensors, {in_tensor}, __LINE__, __FILE__);
+
+  EXPECT_FALSE(out_tensors[0].is_alias_of(in_tensor));
+}
+
+TEST_F(ResizeTest, NoncancellableSplitAndCat_PermutedDifferently) {
+  auto fusion = std::make_unique<Fusion>();
+  FusionGuard fg(fusion.get());
+
+  TensorView* in = makeContigConcreteTensor({2, 2, 2, 10});
+  TensorView* s0 = slice(in, {0, 0, 0, 0}, {2, 2, 2, 2});
+  TensorView* s1 = slice(in, {0, 0, 0, 2}, {2, 2, 2, 5});
+  TensorView* s2 = slice(in, {0, 0, 0, 5}, {2, 2, 2, 10});
+  s0 = permute(s0, {2, 1, 0, 3});
+  s1 = permute(s1, {1, 0, 2, 3});
+  s2 = permute(s2, {2, 1, 0, 3});
+  TensorView* out = cat({s0, s1, s2}, /*dim=*/-1);
+
+  fusion->addInput(in);
+  fusion->addOutput(out);
+
+  auto options = at::TensorOptions().dtype(at::kFloat).device(at::kCUDA, 0);
+  at::Tensor in_tensor = at::randn({2, 2, 2, 10}, options);
+
+  FusionExecutorCache fec(std::move(fusion));
+  auto out_tensors = fec.runFusionWithInputs({in_tensor});
+  testValidate(fec.fusion(), out_tensors, {in_tensor}, __LINE__, __FILE__);
+
+  EXPECT_FALSE(out_tensors[0].is_alias_of(in_tensor));
+}
+
 } // namespace nvfuser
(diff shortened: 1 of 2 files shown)


wujingyue:
!build

csrc/optimization/move_split_cat.cpp
```diff
@@ -122,14 +231,30 @@ TensorView* findCancelingSplit(CatOp* cat) {
...
+
+    TensorView* merged_out = split_in;
+    for (auto i = use_def_chain.rbegin(), end = use_def_chain.rend(); i != end;
+         i++) {
+      Expr* to_replay = *i;
+      if (to_replay->isA<LoadStoreOp>()) {
+        auto* set_out = to_replay->output(0)->as<TensorView>();
+        std::vector<int64_t> permutation = *ir_utils::computePermutation(
+            set_out->getRootDomain(), set_out->getMaybeRFactorDomain());
+        merged_out = permute(merged_out, permutation);
```
jacobhinkle, line 250: Honest Q: would something like this be better or worse? This will transfer definitions as you propagate forward, which means undefining the old values, so if you do this you should double check that there are no additional uses dangling around. Note that you don't necessarily need to use `transferDefinitionToNewOutputs` either: if you look at that function it shows how to use `to_replay->newObjectFunc()` to create another definition that would work. The key is to build the output TensorView properly.

```suggestion
        old_out = to_replay->output(0)->as<TensorView>();
        std::vector<IterDomain*> new_root;
        new_root.reserve(old_out->getRootDomain().size());
        for (auto* id : old_out->getRootDomain()) {
          new_root.push_back(
            // You'll need to initialize ca_map earlier and find the original axis from the split so that you can map it here
            ca_map.areMapped(id, original_split_axis, IdMappingMode::EXACT) ?
            IterDomainBuilder(id).extent(catted_extent).build() :
            id);          
        }
        auto* new_out = IrBuilder::create<TensorView>(new_root, old_out->contiguity());
        ir_utils::transferDefinitionToNewOutputs(to_replay, {new_out});
        merged_out = new_out;
```

test/test_resize.cpp
```diff
@@ -3326,4 +3326,81 @@ TEST_F(ResizeTest, NoncancellableSplitAndCat_CatOnlySubsetOfSplitOutputs) {
...
 }
 
+TEST_F(ResizeTest, CancellableSplitAndCat_PermuteInBetween) {
+  auto fusion = std::make_unique<Fusion>();
+  FusionGuard fg(fusion.get());
+
+  TensorView* in = makeContigConcreteTensor({2, 3, 10});
+  TensorView* s0 = slice(in, {0, 0, 0}, {2, 3, 2});
+  TensorView* s1 = slice(in, {0, 0, 2}, {2, 3, 5});
+  TensorView* s2 = slice(in, {0, 0, 5}, {2, 3, 10});
```
jacobhinkle, line 3336: You might also want a test where the split/cat axis is permuted, to ensure that you are tracking that properly.

jacobhinkle (approved):

csrc/optimization/move_split_cat.cpp
```diff
@@ -15,11 +15,96 @@
...
+  // 3. don't resize the split axis.
+
+  if (std::adjacent_find(
+          frontier.begin(), frontier.end(), [](Expr* lhs, Expr* rhs) {
+            return !lhs->sameOp(rhs);
+          }) != frontier.end()) {
+    return false;
+  }
+
+  if (auto* set = dynamic_cast<LoadStoreOp*>(frontier[0])) {
```
jjsjann123, line 43: qq: what happens when we encounter a reduction on the split_axis?

csrc/optimization/move_split_cat.cpp
```diff
@@ -30,52 +115,76 @@ namespace {
...
+    if (!horizontallyMergeable(frontier, std::ref(split_axis))) {
       return nullptr;
     }
-    slices.push_back(slice);
+    use_def_chain.push_back(frontier[0]);
+
+    // Advance the frontier in lockstep.
+    for (Expr*& e : frontier) {
+      NVF_ERROR(
+          e->inputs().size() == 1,
```
jjsjann123, line 156: ha, nvm. I see that we are only handling unary ops.

jjsjann123: Wouldn't squeeze/unsqueeze/broadcast still falls in here? I think we'll need to compensate split_axis for those as well... Otherwise, we might falsely cancel out split/concat on different axis.

test/test_resize.cpp
```diff
@@ -3326,4 +3326,81 @@ TEST_F(ResizeTest, NoncancellableSplitAndCat_CatOnlySubsetOfSplitOutputs) {
   EXPECT_FALSE(out_tensors[0].is_alias_of(in_tensor));
 }
 
+TEST_F(ResizeTest, CancellableSplitAndCat_PermuteInBetween) {
```
jjsjann123, line 3329: Can we add a test where we have:

```
TensorView* in = makeContigConcreteTensor({2, 2, 4});
TensorView* s0 = broadcast(slice(in, {0, 0, 0}, {2, 2, 2}), {false, true, false, false});
TensorView* s1 = broadcast(slice(in, {0, 0, 2}, {2, 2, 4}), {false, true, false, false});
TensorView* out = cat({s0, s1, s2}, /*dim=*/2);
```

csrc/optimization/move_split_cat.cpp
wujingyue: This function will skip the op and return false, which is a conservative but correct answer. Or are you saying LoadStoreOp can reduce an axis? (I wasn't aware of that).

csrc/optimization/move_split_cat.cpp
wujingyue: `horizontallyMergeable` returns false for anything that's not a set, so this optimization bails out when seeing squeeze/unsqueeze/broadcast between the split and the cat. Is that still a problem?

test/test_resize.cpp
wujingyue: Done. I assume this is to verify the optimization correctly bails out when seeing an unsupported op between the split and the cat?

csrc/optimization/move_split_cat.cpp
jjsjann123: Ha, sorry I was blind and didn't see there was a `return false` in the end.

you are right that `LoadStoreOp*` here should be pretty safe.

test/test_resize.cpp
jjsjann123: Thanks for adding the test. I was wrong about the horizontal merge logic, but I'm happy with adding that test.

jjsjann123 (approved):

csrc/optimization/move_split_cat.cpp
```diff
@@ -122,14 +230,30 @@ TensorView* findCancelingSplit(CatOp* cat) {
...
+
+    TensorView* merged_out = split_in;
+    for (auto i = use_def_chain.rbegin(), end = use_def_chain.rend(); i != end;
+         i++) {
+      Expr* to_replay = *i;
+      if (to_replay->isA<LoadStoreOp>()) {
+        auto* set_out = to_replay->output(0)->as<TensorView>();
+        std::vector<int64_t> permutation = *ir_utils::computePermutation(
+            set_out->getRootDomain(), set_out->getMaybeRFactorDomain());
+        merged_out = permute(merged_out, permutation);
```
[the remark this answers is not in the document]
wujingyue: Added a comment about this and will resolve this in a later PR.

wujingyue:
!build

PR #1784, opened 2024-02-17
Target arch validation

### LINKED ISSUE #1760: Validate use of bf16 (and other relatively new hardware features)
(Originally in #1758)

The bf16 type is only supported on sm80 and later. Currently, nvFuser will just fail when we compile a generated kernel with nvrtc. While we do not care too much about pre-A100 generations, we should at least do:

- More informative error messages rather than just nvrtc compilation failure
- Explicitly define the required minimum architecture version

When we lower a `Fusion` to a `Kernel`, we should define the minimum required version. The default should be sm70, i.e., the V100 version. When the `Kernel` has `bf16` vals, it should at least be `sm80`. Other recent architecture features such as memcpy_async and TMA should also set the minimum version accordingly.

This minimum version should then be checked before the `Kernel` is compiled by nvrtc. Now that we have the information about the required minimum version, we should be able to give more helpful information.

jacobhinkle:
This introduces a minimum device version attribute to the kernel summary and populates it by traversing the fusion early in lowering to find ops or data types introduced after sm70. This version constraint is checked in FusionExecutor::compileFusion and if the target arch doesn't satisfy the constraint then an informative error message is given. Currently BFloat16, cp.async and cp.async.bulk are checked, as are the various MMA macros.

Fixes #1760

### DIFF
--- a/csrc/device_lower/analysis/device_version.cpp
+++ b/csrc/device_lower/analysis/device_version.cpp
@@ -22,7 +22,7 @@ void MinimumDeviceVersion::dispatch(Val* val) {
 
 void MinimumDeviceVersion::handle(MmaOp* mma_op) {
   if (isTuring(mma_op->macro())) {
-    ensureVersion({7, 0}, "Fusion contains a Turing MMA macro");
+    ensureVersion({7, 5}, "Fusion contains a Turing MMA macro");
   } else if (isAmpere(mma_op->macro())) {
     ensureVersion({8, 0}, "Fusion contains an Ampere MMA macro");
   } else if (isHopper(mma_op->macro())) {
--- a/csrc/device_lower/analysis/device_version.cpp
+++ b/csrc/device_lower/analysis/device_version.cpp
@@ -11,12 +11,13 @@
 
 namespace nvfuser {
 
-void MinimumDeviceVersion::handle(Val* val) {
+void MinimumDeviceVersion::dispatch(Val* val) {
   if (val->dtype() == DataType::BFloat16) {
     ensureVersion(
         {8, 0},
         "Fusion contains BFloat16 values which was introduced in Ampere (8.0)");
   }
+  IterVisitor::dispatch(val);
 }
 
 void MinimumDeviceVersion::handle(MmaOp* mma_op) {
--- a/csrc/device_lower/analysis/device_version.cpp
+++ b/csrc/device_lower/analysis/device_version.cpp
@@ -1,0 +1,38 @@
+// clang-format off
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2023-present NVIDIA CORPORATION & AFFILIATES.
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+// clang-format on
+#include <device_lower/analysis/device_version.h>
+#include <device_lower/lower2device.h>
+#include <mma_type.h>
+
+namespace nvfuser {
+
+void MinimumDeviceVersion::handle(Val* val) {
+  if (val->dtype() == DataType::BFloat16) {
+    ensureVersion({7, 0});
+  }
+}
+
+void MinimumDeviceVersion::handle(MmaOp* mma_op) {
+  if (isTuring(mma_op->macro())) {
+    ensureVersion({7, 0});
+  } else if (isAmpere(mma_op->macro())) {
+    ensureVersion({8, 0});
+  } else if (isHopper(mma_op->macro())) {
+    ensureVersion({9, 0});
+  } else {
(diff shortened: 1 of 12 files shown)


jacobhinkle:
!build

jacobhinkle:
!build

jacobhinkle:
!build

jacobhinkle:
!build

jacobhinkle:
!build

jacobhinkle:
!build

jacobhinkle:
!build

csrc/device_lower/analysis/device_version.cpp
```diff
@@ -0,0 +1,60 @@
...
+        "LoadStoreOpType::CpAsyncBulkTensorTile requires Hopper (9.0) or newer");
+  }
+}
+
+void MinimumDeviceVersion::ensureVersion(
+    std::pair<int, int> version,
+    std::string reason) {
+  if (version > min_version_) {
+    min_version_ = version;
+    reason_ = reason;
```
jacobhinkle, line 56: There can be multiple reasons for ensuring the same version, but it would be slightly more complicated to list them all, so we will only include the latest reason in the error message.

jacobhinkle:
!build

jacobhinkle:
!build

jacobhinkle:
!build

test/test_gpu3.cpp
```diff
@@ -7745,7 +7765,11 @@ TEST_F(NVFuserTest, AllInputDtypes) {
     output = IrBuilder::addExpr(output, idx);
     output = IrBuilder::addExpr(output, i32);
     output = IrBuilder::addExpr(output, b);
-    output = IrBuilder::addExpr(output, castOp(DataType::Double, bf16));
+    if (at::cuda::getCurrentDeviceProperties()->major >= 8) {
+      auto bf16 = IrBuilder::create<Val>(DataType::BFloat16);
+      fusion->addInput(bf16);
+      output = IrBuilder::addExpr(output, castOp(DataType::Double, bf16));
+    }
```
jacobhinkle, line 7772: This test worked previously on V100, even though the fusion contains BFloat16.

jacobhinkle:
Should we also check cuda version?

csrc/executor.cpp
```diff
@@ -331,6 +331,29 @@ void FusionExecutor::compileFusion(
...
   kir::Kernel* kernel = lowered_->kernel();
+
+  const kir::KernelSummary& kernel_summary = kernel->summary();
+
+  // TODO: this replicates the target GPU version computation from
+  // executor_utils.
+  std::pair<int, int> target_arch;
+  bool compile_to_sass = false;
+  executor_utils::queryTargetGPUVersion(
+      properties, target_arch.first, target_arch.second, compile_to_sass);
```
wujingyue, line 342:
```suggestion
  executor_utils::queryTargetGPUVersion(
      properties, std::ref(target_arch.first), std::ref(target_arch.second), /*compile_to_sass=*/false);
```

std::ref makes it clear to the reader that this variable can be modified by the callee.

csrc/device_lower/analysis/device_version.h
```diff
@@ -0,0 +1,59 @@
...
+#include <vector>
+
+namespace nvfuser {
+
+class LoopIndexing;
+class ComputeAtMap;
+
+//! Traverses a Fusion to find the minimum supported CUDA compute capability
+//! that will be able to run the generated kernel.
+class MinimumDeviceVersion : IterVisitor {
```
wujingyue, line 27:
```suggestion
class MinimumDeviceVersion : private IterVisitor {
```

I rarely used private inheritance so I forgot about the default :) 

Nit: I'm not a big fan of using inheritance when it's not an "is-a" relation. I'd probably do composition instead -- adding a public IterVisitor as a member.

wujingyue (approved):

csrc/executor.cpp
```diff
@@ -331,6 +331,29 @@ void FusionExecutor::compileFusion(
...
+
+  NVF_CHECK(
```
zasdfgbnm, line 344: Could you move this code down after `getStructuredCode` before nvRTC? Doing so is helpful because, for example, when I am developing Hopper, I don't want to always grab a Hopper machine: sometimes I can just use whatever machine and using `DebugDumpOption::CudaKernel` to check is sufficient for me.

zasdfgbnm (approved):

csrc/executor.cpp
jacobhinkle: I don't have much experience with `std::ref` but that sounds good to me.

csrc/executor.cpp
jacobhinkle: Great suggestion. I moved it and verified that the generated kernel is now printed before the error.

PR #1789, opened 2024-02-19
Refactor so CancelSplitCat becomes a class.

wujingyue:
This makes it convenient to use an IdModel as a class member without having to pass it through many functions.

I examined the NVFUSER_TRACE. FusionKernelRuntime::FusionKernelRuntime is bottlenecked by
"Finding valid fusion segment solutions" not pre-segmenter passes. I added the FUSER_PERF_SCOPE for pre-segmenter passes anyway. 

For #1768

### DIFF
--- a/csrc/preseg_passes/move_split_cat.cpp
+++ b/csrc/preseg_passes/move_split_cat.cpp
@@ -21,9 +21,48 @@ namespace nvfuser::preseg_passes {
 
 namespace {
 
-// Returns true when Exprs in the frontier can be horizontally merged and
-// applied on the unsplit tensor.
-bool horizontallyMergeable(
+class CancelSplitCat {
+ public:
+  CancelSplitCat(Fusion* fusion) : fusion_(fusion) {}
+
+  // Finds all cancellable <split,cat> pairs, cancels them and horizontallly
+  // merges ops in between.
+  void run();
+
+ private:
+  // Returns true when Exprs in the frontier can be horizontally merged and
+  // applied on the unsplit tensor.
+  bool horizontallyMergeable(
+      const std::vector<Expr*>& frontier,
+      int64_t& split_axis);
+
+  // Finds the canceling split of `cat` and returns the input TensorView of the
+  // split. A split (implemented as multiple `slice`s) and a cat cancel when
+  // they work on the same dimension. For example, when
+  //
+  //   s0 = in[:, :5]
+  //   s1 = in[:, 5:]
+  //   out = cat([s0, s1], dim=-1)
+  //
+  // findCancelingSplit(out) returns `in`.
+  //
+  // `cat` doesn't have to immediately follow the split. For example, when
+  //
+  //   s0 = in[:, :5]
+  //   s1 = in[:, 5:]
+  //   t0 = permute(s0)
+  //   t1 = permute(s1)
+  //   out = cat([t0, t1], dim=0)
+  //
+  // In addition to returning `in`, findCancelingSplit(out) puts `t0`'s defining
+  // `permute` into `use_def_chain` so the caller can reconstruct `out` by
+  // replaying `use_def_chain` (in reverse order) on `in`.
+  TensorView* findCancelingSplit(CatOp* cat, std::vector<Expr*>& use_def_chain);
+
+  Fusion* fusion_;
+};
+
+bool CancelSplitCat::horizontallyMergeable(
     const std::vector<Expr*>& frontier,
     int64_t& split_axis) {
   NVF_ERROR(!frontier.empty());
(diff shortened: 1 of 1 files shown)


wujingyue:
!build

csrc/preseg_passes/move_split_cat.cpp
```diff
@@ -21,9 +22,52 @@ namespace nvfuser::preseg_passes {
...
 namespace {
 
-// Returns true when Exprs in the frontier can be horizontally merged and
-// applied on the unsplit tensor.
-bool horizontallyMergeable(
+class CancelSplitCat {
+ public:
+  CancelSplitCat(Fusion* fusion)
+      : fusion_(fusion),
+        id_model_(fusion, /*build_graphs=*/true, /*allow_self_mapping=*/true) {}
```
naoyam, line 29: We may want to consider not building all the graphs by default. If all we need is just the exact graph, we can skip generating the other graphs, which may be much more costly than just building the exact graph.

csrc/preseg_passes/move_split_cat.cpp
```diff
@@ -21,9 +22,52 @@ namespace nvfuser::preseg_passes {
...
+  //   out = cat([t0, t1], dim=0)
+  //
+  // In addition to returning `in`, findCancelingSplit(out) puts `t0`'s defining
+  // `permute` into `use_def_chain` so the caller can reconstruct `out` by
+  // replaying `use_def_chain` (in reverse order) on `in`.
+  TensorView* findCancelingSplit(CatOp* cat, std::vector<Expr*>& use_def_chain);
+
+  Fusion* fusion_;
+
+  IdModel id_model_;
```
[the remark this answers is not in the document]
wujingyue: That's right -- I'm doing this to keep https://github.com/NVIDIA/Fuser/pull/1799 small.

csrc/preseg_passes/move_split_cat.cpp
wujingyue: Totally agree -- https://github.com/NVIDIA/Fuser/pull/1799/files#diff-4e65ceb031d001c04bbef22544b699ffd74f1e4ed0dcea24230b18b77860d4f0R79-R80. Do you want to take that?

naoyam: I think it's fine to leave that part in the follow-up PR.

naoyam (approved):
LGTM.

Out of curiosity, have you noticed any significant latency increase? I haven't done any performance profiling, but some parts of the IdModel analyses could by expensive. Just building the exact graph would be no worse than the current `ComputeAtMap` but the loop promotion analysis, which is automatically done if a loop graph is also built, might be very costly. Note that ToT doesn't have much yet, but I have pending RPs to expand that. (#1777, #1830)

wujingyue:
> LGTM.
> 
> Out of curiosity, have you noticed any significant latency increase? I haven't done any performance profiling, but some parts of the IdModel analyses could by expensive. Just building the exact graph would be no worse than the current `ComputeAtMap` but the loop promotion analysis, which is automatically done if a loop graph is also built, might be very costly. Note that ToT doesn't have much yet, but I have pending RPs to expand that. (#1777, #1830)

Good point. No, I haven't measured the compile time impact either. How do we measure that? cc @rdspring1

That being said, I believe there are lots of low hanging fruits to make IdModel run faster. For example, our disjoint set algorithm can be made to approximately O(1) per operation by using path compression and tracking elements in a set using a linked list.

naoyam:
> > LGTM.
> > Out of curiosity, have you noticed any significant latency increase? I haven't done any performance profiling, but some parts of the IdModel analyses could by expensive. Just building the exact graph would be no worse than the current `ComputeAtMap` but the loop promotion analysis, which is automatically done if a loop graph is also built, might be very costly. Note that ToT doesn't have much yet, but I have pending RPs to expand that. (#1777, #1830)
> 
> Good point. No, I haven't measured the compile time impact either. How do we measure that? cc @rdspring1
> 

The easiest way would be to use the nvtx makers we embed. Unless it's disabled by `NVFUSER_DISABLE=nvtx`, the runtime automatically inserts nvtx markers so that nsys and other tools can understand and visualize the timeline. I haven't tried it myself, but running a test with nsys should generate a trace file that has timestamps of annotated regions by `FUSER_PERF_SCOPE`. The generated trace file can be visualized, but there should also be a command line tool to interpret it.

> That being said, I believe there are lots of low hanging fruits to make IdModel run faster. For example, our disjoint set algorithm can be made to approximately O(1) per operation by using path compression and tracking elements in a set using a linked list.

Yes, nothing has been done for its efficiency yet, and I won't be surprised to see some slow results. I don't think we would need to prioritize that at this moment, but it's been in my TODO list that at least we would make sure nothing gets intolerably slow.

rdspring1:
I use the nvtx markers to measure compile time latency. For visualization, I use either Nsight systems or built-in tracing infrastructure in chrome and the `NVFUSER_TRACE` flag.

Reference: https://github.com/NVIDIA/Fuser/blob/main/csrc/instrumentation.h#L23-L38

PR #1788, opened 2024-02-19
Layout propagation

jjsjann123:
Stacked PRs:
#1755 enabling layout propagation through runtime
#1792 propagation rule for broadcast
#1790 propagation rule for binary op
==== #1788 adding layout inference pass **_<- this one_**

What's in this PR:
inferenceAllocationOrder pass that works on an entire Fusion:
It computes AllocationOrder on inputs by looking at each TensorView's allocation_domain and rfactor_domain;
It uses a predefined rule (in AllocationOrderInferencer) to traverse and propagate AllocationOrder from inputs to the entire fusion;

Note that the pass itself doesn't mutate the fusion IR. It's just a utility function that suggests ways to specify allocation domain to be used by other optimization passes.

- [x] adding inferenceAllocationOrder pass function;
- [x] adding propagate rule for unary op;
- [x] adding cpp test to verify propagation rule;

Quick design doc: https://github.com/NVIDIA/Fuser/issues/1756

Future Work:
* expanding propagation rule to cover more operation;

PR #1790, opened 2024-02-19
adding propagation rule and test for binary ops

jjsjann123:
Stacked PRs:
#1755 enabling layout propagation through runtime
#1792 propagation rule for broadcast
==== #1790 propagation rule for binary op **_<- this one_**
#1788 adding layout inference pass

What's in this PR:

BinaryOp propagation tries to merge the allocation order of both inputs:
* when there's only one operand is a tensor, we just forward the recorded allocation order
* when both operands are tensors, we resolve it by:
    i. prioritize the tensor with less broadcast iterdomain;
    ii. otherwise, we just propagate the allocation order of lhs.

Propagation rule for binary operation, 

- [x] adding propagate rule for binary op;
- [x] handling two scalar;
- [x] handling intermediate tensors (factory tensor);
- [x] adding cpp test to verify propagation rule;


jjsjann123:
!build

wujingyue (approved):

jjsjann123:
!build

jjsjann123:
tests passed locally and previous CI cleared before merging `main`. I'm merging this PR.

PR #1793, opened 2024-02-19
Migrate more matmul tests to 3D

zasdfgbnm:
Stacked on https://github.com/NVIDIA/Fuser/pull/1776, just migrating more tests.
Partially fixes: https://github.com/NVIDIA/Fuser/issues/1628


zasdfgbnm:
!build

csrc/index_compute.cpp
zasdfgbnm, line 1: I don't know how correct this is generally. But for the case of matmul, this seems to be working. And I am not interested in improving the general correctness of legacy swizzle indexing. 

cc: @naoyam

csrc/index_compute.cpp
```diff
@@ -1101,26 +1101,67 @@ class UpdateLeafIndices : public IterVisitor {
...
+    if (index_map_.find(out_id) != index_map_.end()) {
+      return;
+    }
+
+    if (outer_id->isBroadcast()) {
+      if (!index_map_.count(inner_id)) {
+        // Reduction axes on producer side could be visited on forward
+        //  propagation pass and current implementation does not yet
+        //  support reduciton on swizzled iterdomains, so un-indexed
+        //  reduction iterdomains are just ignored for now.
```
drzejan2, line 1115: Perhaps we should also mention in comments that the same assumptions are made for `bcast` iterdomains, so it reflects the check in code?
```suggestion
        //  reduction iterdomains are just ignored for now. The same applies to BroadcastOp.
```

drzejan2 (approved):
Changes look ok, although I suggest to wait with promotion for one more reviewer to check this PR. I don't have enough knowledge to evaluate changes introduced to `index_compute.cpp`.

zasdfgbnm:
!build

PR #1796, opened 2024-02-20
don't project to inputs if calculate persistent buffer requires rng op

liqiangxl:
This PR stops project persistent buffer to inputs if the recalculaiton requires expensive rng op.
It improves performance of `test_dropout_layernorm_fwd_benchmark` in our current python benchmarks.
For case `[2048, 31744] with bf16` on A100, reduced from 0.92 ms to 0.37 ms.

I was thinking push this PR after the inner peristent heuristic revision is done. But since we already have `test_dropout_layernorm_fwd_benchmark` in our current python benchmarks, it is better to merge it early to save some bechmark time.


csrc/scheduler/normalization_utils.cpp
```diff
@@ -841,7 +842,15 @@ bool projectBufferToInputs(
         expr->as<UnaryOp>()->getUnaryOpType() == UnaryOpType::Exp) {
       has_exp_op = true;
     }
+    if (expr->isA<RNGOp>()) {
+      has_rng_op = true;
+    }
+  }
+  // don't project if calculate persistent buffer requires very expensive rng op
+  if (has_rng_op) {
+    return false;
```
jjsjann123, line 851: In my monkey brain, this felt like a straightforward update. i.e. recomputing dropout would be expensive....

Do we care about performance measurement across the benchmark?
^^^ FYI, this is a genuine question, not a suggestion that we should. But at the same time, a quick benchmark run on normalization shouldn't hurt :stuck_out_tongue:

liqiangxl: we should test so we can include in the top-5 saying the perf is increased by 2x to 3x. :laughing: 

image

jjsjann123 (approved):
That's some smoking number you have!

csrc/scheduler/normalization_utils.cpp
naoyam: Could you give some more details about the speedups? For example, I assume this would result in using a larger persistent buffer, so I'd expect it would hit the segmentation cliff sooner than before, but here I only see uniform speedups. What's the impact of the size of the persistent buffer?

liqiangxl: You are right. Persistent buffer is changed from `fp16` to `fp32`, so cutoff changes from 64K to 32K. I need to test do we want to segment into 2 kernerls or project to inputs.

liqiangxl: Add more tests. There is a bug in the current implementation. When disable project to inputs, the buffer size is increased, need to check if we still have enough register or smem to store the buffers. Otherwise, should still project to inputs or reject early in `canScheduleRunTime` .
image

liqiangxl: (1) This graph is taken from this PR + some extra fix to  ensure we have enough register or smem to store the buffers when disable project to input.
(2) I am thinking revert then redo the PR, a patch is also fine.

naoyam: Since it can result in a crash due to insufficient smem, we should revert the PR for now and work on the fixed version. I approved the revert PR.

jjsjann123: > There is a bug in the current implementation. When disable project to inputs, the buffer size is increased, need to check if we still have enough register or smem to store the buffers.

So before the code change here that disables the projection, we would still have a cliff where we won't have enough register/smem for the buffer. Why is it a launch failure after the change, what was the graceful fallback in existing implementation?

liqiangxl: In `canScheduleRunTime`, it ensures `avilable reg/smem` is larger than `min(buffer_not_project, buffer_project_to_input)`, otherwise will reject and segment. There may have a scenario where  `buffer_project_to_input < avilable reg/smem < buffer_not_project`, in this case we must project to inputs, otherwise there is not enough register/smem for the buffer.

In this PR (before fix), I just disabled project to inputs without checking whether we have enough register/smem or not.

PR #1797, opened 2024-02-20
Add swizzle support for TMA load

zasdfgbnm:
This PR adds support for swizzling for TMA load by scheduling the allocation domain of the smem tensor. The swizzle mode (None, 32B, 64B, 128B) is determined by the schedule, in specific, the extent of the IterDomain that is an input/output of the `Swizzle` expr.

### DIFF
--- a/csrc/index_compute.cpp
+++ b/csrc/index_compute.cpp
@@ -3256,21 +3256,23 @@ Val* Index::cpAsyncBulkIndex(
   NVF_ERROR(
       gmem_tv->getMemoryType() == MemoryType::Global,
       "cpAsyncBulkIndex is only for global memory tensors");
-  NVF_ERROR(
-      gmem_tv->getMaybeRFactorDomain() == gmem_tv->getLeafDomain(),
-      "not supported yet");
-  NVF_ERROR(
-      gmem_tv->getMaybeAllocationDomain() == gmem_tv->getLeafDomain(),
-      "not supported yet");
-  for (auto id : consumer->getMaybeRFactorDomain()) {
-    NVF_ERROR(
-        id->isBulk(),
-        "cpAsyncBulkIndex only support whole tensor copy for now.");
-  }
 
   int64_t dim = (int64_t)gmem_tv->nDims();
   NVF_ERROR(dim > 0);
   int64_t itemsize = dataTypeSize(gmem_tv->dtype());
+
+  int64_t swizzle_size = 1;
+  auto exprs = DependencyCheck::getAllExprsBetween(
+      {consumer->getMaybeRFactorDomain().begin(),
+       consumer->getMaybeRFactorDomain().end()},
+      {consumer->getMaybeAllocationDomain().begin(),
+       consumer->getMaybeAllocationDomain().end()});
+  for (auto expr : exprs) {
+    if (auto s = dynamic_cast<Swizzle*>(expr)) {
+      swizzle_size = s->inX()->extent()->evaluate().as<int64_t>();
+      break;
+    }
+  }
 
   auto metadata = IrBuilder::metadataExpr(gmem_tv);
   auto global_address = IrBuilder::getAttrExpr(metadata, "data");
@@ -3310,7 +3312,7 @@ Val* Index::cpAsyncBulkIndex(
       box_dim,
       element_strides,
       TensorMapInterleave::NoInterleave,
-      MmaInputSmemSwizzle::None,
+      getSwizzleFromBytes(swizzle_size * 16),
       TensorMapL2Promotion::NoL2Promotion,
       TensorMapFloatOOBFill::NoOOBFill);
 
--- a/test/test_memory.cpp
+++ b/test/test_memory.cpp
@@ -172,20 +172,109 @@ TEST_F(MemoryTest, RefineCachePolicy) {
 }
 
 class TMATest : public NVFuserTest {
+ protected:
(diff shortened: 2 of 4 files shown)


zasdfgbnm:
!build

zasdfgbnm:
!build

csrc/index_compute.cpp
```diff
@@ -3256,22 +3256,24 @@ Val* Index::cpAsyncBulkIndex(
...
-  NVF_ERROR(
-      gmem_tv->getMaybeRFactorDomain() == gmem_tv->getLeafDomain(),
-      "not supported yet");
-  NVF_ERROR(
-      gmem_tv->getMaybeAllocationDomain() == gmem_tv->getLeafDomain(),
-      "not supported yet");
-  for (auto id : consumer->getMaybeRFactorDomain()) {
-    NVF_ERROR(
-        id->isBulk(),
-        "cpAsyncBulkIndex only support whole tensor copy for now.");
```
zasdfgbnm, line 3268: We still only support whole tensor copy. It is just becoming harder to check this with the swizzle support. So I decided to stop checking it, as it is only a temporary solution.

test/test_memory.cpp
```diff
@@ -172,20 +172,109 @@ TEST_F(MemoryTest, RefineCachePolicy) {
...
+  void SetUp() override {
+    TMATest::SetUp();
+    swizzle = std::get<0>(GetParam());
+    dtype = std::get<1>(GetParam());
+  }
+};
+
+// Assuming tv is currently scheduled as a 1D flat tensor, schedule the TMA
+// swizzle for it.
+void scheduleTMASwizzle(TensorView* tv, int64_t swizzle_size) {
```
zasdfgbnm, line 261: For now, this is just a function in this test file. In the future, I will refactor to make it a scheduler utility function, and this utility function will be used to schedule mma inputs.

test/test_memory.cpp
```diff
@@ -197,19 +286,20 @@ TEST_F(TMATest, LoadCompleteTensor1D) {
...
   tv1->axis(0)->parallelize(ParallelType::Bulk);
 
-  auto options = at::TensorOptions().dtype(at::kFloat).device(at::kCUDA, 0);
-  auto t0 = at::randn({32}, options);
+  auto options =
+      at::TensorOptions().dtype(data_type_to_aten(dtype)).device(at::kCUDA, 0);
+  auto t0 = at::randn({innerDimSize()}, options);
   FusionExecutor fe;
-  fe.compileFusion(&fusion, {t0}, {}, {DataType::Int32});
+  fe.compileFusion(&fusion, {t0}, {}, matmul_cparams);
```
zasdfgbnm, line 293: The `matmul_cparams` is to make sure the `nvfuser_zero` is disabled, so smem allocation is well aligned.

test/test_memory.cpp
```diff
@@ -172,20 +172,109 @@ TEST_F(MemoryTest, RefineCachePolicy) {
...
+    if (cudaArchGuardShouldSkip(9, 0)) {
       GTEST_SKIP() << "skipping tests on pre-Hopper GPUs";
     }
     NVFuserTest::SetUp();
   }
 };
 
-TEST_F(TMATest, LoadCompleteTensor1D) {
+// Check that there is an xor "^" somewhere in the kernel
+class XorFinder : private kir::IrVisitor {
```
naoyam, line 185: Didn't we have something like `getAllExprs` with a kernel IR?

naoyam (approved):
LGTM

test/test_memory.cpp
zasdfgbnm: Are you referring to `ir_utils::flattenScopedExprs`? This might simplify the logic a little bit, but I don't think it will make this piece of code significantly simpler.

The complication here is, for a swizzled kernel, due to index hoisting, we might have something like:
```C++
i2 = i1 ^ i0;
T0[i3] = T0[i2];
```
or something like:
```C++
T0[i1 ^ i0] = T0[i2]
```
so we need to make sure that we are checking both the "scoped exprs" and the exprs in tensor index for xor.

Also, exprs in tensor index must be checked recursively. For example, if we have `i1 + (i2 - (i3 * i4 ^ i5))`, we will have to `handle(+)`, `handle(-)`, `handle(*)` before reaching `handle(^)`, so we need to implement this recursive algorithm anyway.

naoyam: Maybe next time we need something like this we should add a util function to grab all exprs recursively.

test/test_memory.cpp
drzejan2: A minor suggestion, instead of using literal/magic number (`16`) maybe we should introduce a constant?

csrc/index_compute.cpp
```diff
@@ -3310,7 +3312,7 @@ Val* Index::cpAsyncBulkIndex(
       box_dim,
       element_strides,
       TensorMapInterleave::NoInterleave,
-      MmaInputSmemSwizzle::None,
+      getSwizzleFromBytes(swizzle_size * 16),
```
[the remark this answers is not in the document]
zasdfgbnm: Fixed in https://github.com/NVIDIA/Fuser/pull/1818

test/test_memory.cpp
zasdfgbnm: Fixed in https://github.com/NVIDIA/Fuser/pull/1818

PR #1800, opened 2024-02-20
add iteration domain grouped block reduction (vectorized blk reduction)

liqiangxl:
**In this PR:** added `blockIterGroupedReduce`. It is used in outer reduction where the iteration domain is vectorized to reduce number of block syncs by moving the `for-loop` over vectorization factors from outside of the `blockReduce` function to the inside of the reduce function. For example, the following figure shows a case where `[vect] = 8`, the original implementation needs to call `blockReduce` 8 times, while the vectorized version only needs to call `blockIterGroupedReduce` once. It moved the 8 parallel reductions into the reduction function without adding additional synchronizations.

image

**Why doing this**:
This is the first step to optimize outer reduciton, it is required by `gridIterGroupedReduce`.

**Performance**:
Iteration domain vectorized block reduction is rarelly seen in the current outer reduction heuristics.
A test of a case with size of 16K x 2K is enforced to use `UNROLL=64 BDX=8 VEC=8`, shows the latency is reduced from 0.034 ms to 0.030 ms after this change. Will do more tests after `gridIterGroupedReduce` is added and heuristics revised.

**TODO:**
 (1) Add `gridIterGroupedReduce` which extends `gridReduce` using a similar method
 (2) optimize heuristics
 (3) unify code path in runtime file `blockIterGroupedReduce` and `blockReduce`


liqiangxl:
!build --diff --diff-bench

runtime/block_reduction.cu
```diff
@@ -111,3 +111,134 @@ __device__ void blockReduce(
...
+// smem. (2) do N * bdimx parallel reductions in smem.
+template <
+    bool X_REDUCE,
+    bool Y_REDUCE,
+    bool Z_REDUCE,
+    bool Aligned,
+    int N, // Number of elements per input array
+    typename T,
+    typename Func>
+__device__ void blockIterGroupedReduce(
```
jacobhinkle, line 129: Could this be unified with `blockReduce` instead of having a separate name? It seems like it's a pretty straight-forward extension. Is there any part of this that you'd expect a slowdown for `N=1` compared to `blockReduce`?

liqiangxl:
@jacobhinkle @xwang233 Do we want to hide the codegen diffs due to runtime function changes? e.g. [link](https://dl.gitlab-master-pages.nvidia.com/pytorch/fuser-gh-mirror//nvfuser_github_ci/codegen_diff_p13005701_j83564367_1708627533401400796_codediff_de82377_049f8ee_custom_command_20240222_182356.html)

jacobhinkle:
Good question. It would be nice to have the option. I think the tool has a flag to enable that. It was originally this way so thatI could see the effect on registers due to changing runtime files.

csrc/codegen.cpp
```diff
@@ -2659,9 +2659,91 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
...
+    // used for outer reduction with vectorized iteration domain.
+    if (num_grouped_iterations > 1 && num_grouped_exprs == 1) {
+      const auto output = grouped_rop->output(0)->as<kir::TensorIndex>();
+      const auto input = grouped_rop->input(0)->as<kir::TensorIndex>();
+      const auto op_type = grouped_rop->getReductionOpType(0);
+      const auto domain = output->view()->domain();
+      const bool has_block_reduce = domain->hasBlockReduction();
+      const bool has_grid_reduce = domain->hasGridReduction();
+      NVF_ERROR(
+          !has_grid_reduce, "IterGroupedGridReduction not implemented yet");
```
jjsjann123, line 2732: I'm assuming this one is fine, since the old code path also has it. So somewhere else it must have already been a branch that protects us.

csrc/device_lower/validation.cpp
```diff
@@ -1174,18 +1174,6 @@ void validateAndConvertIterDomainGrouping(Fusion* fusion) {
...
-          "Invalid use of ParallelType::Group.",
-          " Only enabled for allreduce reductions: ",
-          rop->toString());
-
-      NVF_CHECK(
-          tv->domain()->hasGridReduction(),
-          "Invalid use of ParallelType::Group.",
-          " Only enabled for grid reductions: ",
-          rop->toString());
-
```
jjsjann123, line 1188: Is the old checks just being on the conservative side?

csrc/executor.cpp
```diff
@@ -1247,11 +1247,22 @@ LaunchParams FusionExecutor::computeLaunchParams(
     const int welford_factor =
         kernel_summary.has_block_welford || kernel_summary.has_grid_welford ? 3
                                                                             : 1;
+    // in outer reduction, may group iteration domain, e.g. when vectorized.
+    const int grouped_iter_factor = kernel_summary.has_iter_grouped_reductions
```
jjsjann123, line 1251: wasn't the default just 1? don't see why we needed to branch here.

Meanwhile, is it safe to assert `kernel_summary.has_iter_grouped_reductions || grouped_iter_factor == 1`?

csrc/kernel.cpp
```diff
@@ -118,6 +118,19 @@ class KernelIrScanner : private IrVisitor {
...
+  void handle(GroupedReductionOp* grouped_rop) final {
+    summary_.has_iter_grouped_reductions = true;
+    int num_grouped_iterations = 1;
+    auto out_tv = ir_utils::getTvOutput(grouped_rop);
+    for (auto axis : out_tv->getLeafDomain()) {
+      if (axis->getParallelType() == ParallelType::Group) {
+        num_grouped_iterations *= (int)axis->extent()->value();
+      }
+    }
+    summary_.num_grouped_iterations =
```
jjsjann123, line 130: so `num_grouped_iterations` could still be 1 when we have `has_iter_gropued_reduction`?

jjsjann123 (changes requested):

jjsjann123:
I put a request for changes just to prevent accidental merge.

I'm mostly just concerned about the last loop out of range access. We don't necessarily need to have a clean guard there, but anything that prevents a silent IMA would be good :wink:

runtime/block_reduction.cu
```diff
@@ -111,3 +111,135 @@ __device__ void blockReduce(
...
+
+  // Adjust shared memory offset for array processing
+  unsigned int smem_offset =
+      (reduction_idx * reduction_size + reduction_tid) * N;
+  if (read_pred) {
+    if constexpr (sizeof(T) * N <= 16) {
+      loadGeneric<T, N>(shared_mem + smem_offset, const_cast<T*>(inp_val));
+    } else {
+      // may larger than 16 bytes, e.g. input fp16 vectorized by 8
+      // but calculation is fp32/fp64 vectorized by 8
```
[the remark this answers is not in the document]
liqiangxl: Then it means the comment is not clear! what about the this version?
```
    // This section calculates the number of vectorized load operations required
    // to fetch all elements of an array into shared memory, assuming each load
    // can transfer up to 16 bytes. For example, with fusion input vectorized by
    // 8 and computations in fp32 (sizeof(T) = 4 bytes), for N = 8 elements, the
    // total data size is 4 * 8 = 32 bytes, necessitating 32 / 16 = 2 load
    // transactions.
```

runtime/block_reduction.cu
liqiangxl: make sense. will consider merge them after `gridIterGoupedReduce`

csrc/executor.cpp
liqiangxl: Good point! I think I was following the welford factor, but we don't need to.
Corrected.
assert is changed to 
```
    NVF_CHECK(
        !(kernel_summary.has_iter_grouped_reductions && welford_factor == 3),
        "can't have welford and iter grouped reductions at the same time! Should be handled by grouped welford!");

```

csrc/kernel.cpp
liqiangxl: it can't be 1 if we have `has_iter_gropued_reduction` becuase we only group if vectorized and only vectorized if vectorization factor is larger than 1.
added check:
```
    NVF_ERROR(
        num_grouped_iterations == 2 || num_grouped_iterations == 4 ||
            num_grouped_iterations == 8 || num_grouped_iterations == 16,
        "Iteration grouped reduction only support grouping 2, 4, 8, or 16 iterations, but found ",
        num_grouped_iterations);
```

liqiangxl:
For `last loop out of range access`, are you indicating 
> for (unsigned int i = 0; i < total_loads; ++i) ?

Needs to ensure these two divisions are divisible.
```
      constexpr unsigned int total_loads = sizeof(T) * N / 16;
      constexpr unsigned int elements_per_load = 16 / sizeof(T);
```

added check:
```
      static_assert(
          sizeof(T) * N == 16 * total_loads,
          "This combination of vectorization factor and data type is not supported!");
      static_assert(
          sizeof(T) * elements_per_load == 16,
          "This data type is not supported!");
```

liqiangxl:
!build

liqiangxl:
!build

jjsjann123:
> Good question. It would be nice to have the option. I think the tool has a flag to enable that. It was originally this way so thatI could see the effect on registers due to changing runtime files.

cc'ing @xwang233 ^^^

runtime/block_reduction.cu
```diff
@@ -111,3 +111,135 @@ __device__ void blockReduce(
...
+      constexpr unsigned int total_loads = sizeof(T) * N / 16;
+      constexpr unsigned int elements_per_load = 16 / sizeof(T);
+#pragma unroll
+      for (unsigned int i = 0; i < total_loads; ++i) {
+        loadGeneric<T, elements_per_load>(
```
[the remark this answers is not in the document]
jjsjann123: nitpick: should we exlucde 32? i.e. fp8 is coming right?!

maybe just a naive check that N is a power of 2 would be a better way. I think functionally, that's enough to ensure the correctness of grouped reduction.

xwang233:
> Good question. It would be nice to have the option. I think the tool has a flag to enable that. It was originally this way so thatI could see the effect on registers due to changing runtime files.

cc @jacobhinkle can you send me the flags on `tools/codediff/compare_codegen.sh` to enable that feature?

runtime/block_reduction.cu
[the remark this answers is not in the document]
liqiangxl: with fp8, we will vectorize by 16, right?

csrc/codegen.cpp
```diff
@@ -2659,9 +2659,91 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
...
+      NVF_ERROR(
+          has_block_reduce,
```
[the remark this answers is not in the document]
jjsjann123: I think it's the mixed term that's very confusing to read.

so `num_grouped_iterations > 1` is really the flag here. Does this mean the code referenced should really be renamed from `use_grouped_reduction` -> `use_iter_grouped_reduction` instead?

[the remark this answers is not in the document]
jjsjann123: FYI: we shouldn't reject code change because it makes the code ugly. But in this case if it's supposed to assert here, then I agree we might prefer to have an error throw rather than silently back out from iter_gropued_reduction.

csrc/scheduler/reduction.cpp
```diff
@@ -1209,18 +1209,20 @@ void scheduleReduction(Fusion* fusion, const ReductionParams& rparams) {
...
-  const bool is_outer_grid_persistence = rparams.persistent_kernel &&
-      rparams.cross_grid_inner_reduction && !rparams.fastest_dim;
-  NVF_ERROR(
-      !is_outer_grid_persistence,
-      "is_outer_grid_persistence should be false in scheduleReduction.");
+
+  // allow iter domain grouped reduction for block outer reduction.
+  // the var name is confusing, should rename [cross_grid/block_inner_reduction]
+  // to [cross_grid/block_reduction]
+  const bool use_grouped_reduction = !rparams.cross_grid_inner_reduction &&
```
jjsjann123, line 1216: *** So this flag is really indicating whether we are `use_iter_grouped_reduction`?

Otherwise, we shouldn't have seen the genSerialReduction entry at all. Please rename this one if that's the case.

csrc/scheduler/reduction.cpp
```diff
@@ -1209,18 +1209,20 @@ void scheduleReduction(Fusion* fusion, const ReductionParams& rparams) {
...
   const bool vectorize =
       rparams.vectorize_inner_reduction || rparams.vectorize_iter_dom;
-  const bool is_outer_grid_persistence = rparams.persistent_kernel &&
-      rparams.cross_grid_inner_reduction && !rparams.fastest_dim;
-  NVF_ERROR(
-      !is_outer_grid_persistence,
-      "is_outer_grid_persistence should be false in scheduleReduction.");
+
+  // allow iter domain grouped reduction for block outer reduction.
+  // the var name is confusing, should rename [cross_grid/block_inner_reduction]
```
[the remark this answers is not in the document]
liqiangxl: created https://github.com/NVIDIA/Fuser/issues/1863

csrc/codegen.cpp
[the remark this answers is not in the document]
liqiangxl: renamed `use_grouped_reduction` to `use_iter_grouped_reduction`.

csrc/scheduler/reduction.cpp
liqiangxl: yes. renamed.

csrc/kernel.cpp
jjsjann123: I'm fine with it as-is for now.

But should we have maybe add another operation with `IterGroupedReductionOp` instead?

csrc/scheduler/reduction_utils.cpp
```diff
@@ -357,7 +357,7 @@ void multiReductionInliner(
     TensorView* reference_tv,
     const bool unroll,
     const bool vectorize,
-    const bool is_outer_grid_persistence,
+    const bool use_grouped_reduction,
```
jjsjann123, line 360: nitpick: these names should also be updated IIUC.

jjsjann123: resolved offline. No need to update since it could be either iter_grouped or expr_grouped.

csrc/kernel.cpp
jjsjann123: resolved offline. we decided to keep it as-is in case we would want a future hybrid mode.

jjsjann123 (approved):
stamping to unblock. Please double check with CI and follow up on requested comment updates.

:bow:

liqiangxl:
!build

jacobhinkle:
> > Good question. It would be nice to have the option. I think the tool has a flag to enable that. It was originally this way so thatI could see the effect on registers due to changing runtime files.
> 
> cc @jacobhinkle can you send me the flags on `tools/codediff/compare_codegen.sh` to enable that feature?

Hmm actually that was the motivation behind this flag https://github.com/NVIDIA/Fuser/blob/main/tools/codediff/diff_report.py#L896. It shouldn't have included a diff if the cuda and ptx were unchanged..

naoyam:
> A test of a case with size of 16K x 2K is enforced to use UNROLL=64 BDX=8 VEC=8, shows the latency is reduced from 0.034 ms to 0.030 ms after this change.

This sounds a little underwhelming. Is this expected? What other cases have you evaluated? It would be interesting to evaluate just the grouped reduction kernel only in isolation from nvfuser. See my internal gitlab repo named `outer_reduction`, which contains some benchmark code that I used when I worked on the grouped welford functions.

runtime/block_reduction.cu
```diff
@@ -111,3 +111,154 @@ __device__ void blockReduce(
...
+    typename T,
+    typename Func>
+__device__ void blockIterGroupedReduce(
+    T out[N],
+    const T inp_val[N],
+    Func reduction_op,
+    T* shared_mem,
+    bool read_pred,
+    bool write_pred,
+    T init_val) {
```
naoyam, line 140: Have you confirmed, for example, there's no bank conflicts? The store looks suspicious to me...

runtime/block_reduction.cu
```diff
@@ -111,3 +111,154 @@ __device__ void blockReduce(
...
+    if (reduction_tid < factor) {
+#pragma unroll
+      for (int i = 0; i < N; ++i) {
+        reduction_op(
+            shared_mem[smem_offset + i],
+            shared_mem[smem_offset + factor * N + i]);
+      }
+    }
+    block_sync::sync<Aligned>();
+  }
```
naoyam, line 221: Suppose `N=8` and the number of threads per thread block is 256, i.e., 8 warps. Once all the initial inputs are uploaded to the shared memory, the rest can be just done as warp reductions by letting each warp take care of one of `N` reductions. That would eliminate all the synchreads. Furthermore, unlike this implementation, the ratio of threads being idle is much smaller.

That's how the grouped welford is implemented. That was important in the case of welford particularly because it has three values to reduce. I'm not sure if it's also the case here, but I suppose that's true to some extent.

liqiangxl:
> > A test of a case with size of 16K x 2K is enforced to use UNROLL=64 BDX=8 VEC=8, shows the latency is reduced from 0.034 ms to 0.030 ms after this change.
> 
> This sounds a little underwhelming. Is this expected? What other cases have you evaluated? It would be interesting to evaluate just the grouped reduction kernel only in isolation from nvfuser. See my internal gitlab repo named `outer_reduction`, which contains some benchmark code that I used when I worked on the grouped welford functions.

I'll double check that. The block reduction doesn't change much becuase the block sync cost is small. For grid reduction, the difference is obvious, see results at https://github.com/NVIDIA/Fuser/pull/1867

naoyam:
> The block reduction doesn't change much becuase the block sync cost is small.

Let's make a quantitative, objective discussion with supporting data.

runtime/block_reduction.cu
liqiangxl: Sounds reasonable. Let me do somes tests of using warp reduction. Just curious, if it is more efficient why the original block reduction is not doing this?

naoyam: Because `N` is just 1, there's not much we can do. I tried the warp reduction once `factor` becomes 32. It was long time ago, but the difference was negligible.

naoyam: And before trying to implement it, let's gather more data. I'm pretty sure if it's more efficient, but I don't know if it's necessary.

PR #1802, opened 2024-02-21
Use cache variables for language standards.

tfogal:
So that users can change all of these at once without wading through the cmake for it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,9 +22,14 @@ if (NVFUSER_DISTRIBUTED)
 endif()
 message(STATUS "Setting NVFUSER_DISTRIBUTED=${NVFUSER_DISTRIBUTED}")
 
-if(NOT NVFUSER_CPP_STANDARD)
-  set(NVFUSER_CPP_STANDARD 20)
-endif()
+# We try to update which C++ standard we use together in lockstep across all
+# built libraries, and these variables control which that is. Generally we are
+# on C++20, but we still support a version of CUDA (11) that does not recognize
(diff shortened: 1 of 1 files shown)


tfogal:
!build

CMakeLists.txt
```diff
@@ -22,9 +22,14 @@ if (NVFUSER_DISTRIBUTED)
...
-  set(NVFUSER_CPP_STANDARD 20)
-endif()
+# We try to update which C++ standard we use together in lockstep across all
+# built libraries, and these variables control which that is. Generally we are
+# on C++20, but we still support a version of CUDA (11) that does not recognize
+# C++20 and so we drop back to 17 there. Also, we allow all of these to be
+# overridden by the user.
+set(NVFUSER_C_STANDARD 17 CACHE STRING "C standard to use for C code")
+set(NVFUSER_CPP_STANDARD 20 CACHE STRING "C++ standard to use for C++ code")
+set(NVFUSER_CUDA_STANDARD 17 CACHE STRING "C++ standard to use for CUDA code")
```
jjsjann123, line 32: Please add a quick comment here about using `set_target_properties` with global variables instead a global `set_property`, shielding us from potential sharp edges from global flags in cmake.

Because that's going to be the first thing a naive dev like me is going to refactor global flags.

jjsjann123 (approved):
LGTM

CMakeLists.txt
tfogal: 😂 will do

PR #1813, opened 2024-02-22
Fix a bug in input forwarding. 

wujingyue:
It causes unnecessary edges between segments and confusing debugging output. For example, 

```
$ NVFUSER_DUMP=segmenter_logging bin/nvfuser_tests --gtest_filter=SegmentationTest.InputForwardingViaUnaryOps

...
**Segmenter** Considering fusion:

***Accepted*** as: no_op
...
```

The empty fusion considered is produced by merging an unnecessary edge between two segments before forwarded inputs.

### DIFF
--- a/csrc/fusion_segmenter.cpp
+++ b/csrc/fusion_segmenter.cpp
@@ -3889,10 +3889,13 @@ void SegmentCandidateFinder::forwardInputs() {
       } else {
         // If there are either no more uses, more than one use, or one use that
         // is not a UnaryOp, then we cannot extend the chain of unary Ops. In
-        // these cases we finalize this chain by saving the uop and its output.
-        excluded_inp_unary_exprs_.pushBack(uop);
+        // these cases we finalize this chain by saving its output as a
+        // forwarded input.
         forwarded_inputs.pushBack(uop->out());
       }
+      // Either way, `uop` is excluded from merging until `resolveInputsInGroup`
+      // adds it back to one of the segments.
+      excluded_inp_unary_exprs_.pushBack(uop);
     }
   }
 
--- a/csrc/fusion_segmenter.cpp
+++ b/csrc/fusion_segmenter.cpp
@@ -2483,6 +2483,12 @@ std::optional<ScheduleHeuristic> tryMerge(
     SegmentedGroup* b = nullptr) {
   FusionSegmentGuard fsg(segmented_fusion, a, b);
 
+  NVF_ERROR(
+      !segmented_fusion->completeFusion()->unordered_exprs().empty(),
+      "We shouldn't attempt to merge empty fusions. "
+      "This might not indicate a bug, "
+      "but it's definitely a change of world view that we should be aware of.");
+
   scheduler_debug_utils::canScheduleMessage(
       "\n**Segmenter** Considering fusion:\n",
       segmented_fusion->completeFusion());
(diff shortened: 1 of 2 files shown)


wujingyue:
!build

test/test_segmentation.cpp
```diff
@@ -227,4 +227,31 @@ TEST_F(SegmentationTest, SetAllocationDomainOnSegmentBoundary) {
...
+
+  FusionExecutorCache fec(std::move(fusion));
+
+  auto options = at::TensorOptions().dtype(at::kBFloat16).device(at::kCUDA, 0);
+  at::Tensor in_tensor = at::randn({2, 3}, options);
+  std::vector<at::Tensor> out_tensors = fec.runFusionWithInputs({in_tensor});
+  testValidate(fec.fusion(), out_tensors, {in_tensor}, __LINE__, __FILE__);
+
+  FusionKernelRuntime* runtime = fec.getMostRecentKernelRuntime();
+  EXPECT_EQ(runtime->fusionSegments()->groups().size(), 1);
```
naoyam, line 254: Isn't this also fine with the current main? Not that I can think of any other good way to test the fix, though.

naoyam (approved):
LGTM. Thanks for the fix.

jacobhinkle (changes requested):
Here is a repro for one of the internal test failures:
```python
import torch


def nvfuser_fusion_id2(fd: FusionDefinition) -> None:
    T0 = fd.define_tensor(
        shape=[-1, -1, -1], contiguity=[True, True, True], dtype=DataType.Float, is_cpu=False, stride_order=[1, 2, 0]
    )
    T1 = fd.define_tensor(
        shape=[-1, -1, -1], contiguity=[True, True, True], dtype=DataType.Float, is_cpu=False, stride_order=[0, 1, 2]
    )
    T2 = fd.ops.tanh(T0)
    T3 = fd.ops.sin(T2)
    T4 = fd.ops.sin(T1)
    T5 = fd.ops.tanh(T4)
    fd.add_output(T3)
    fd.add_output(T5)


    nvfuser_fusion_id2(fd)

inputs = [
    torch.randn((30,), dtype=torch.float32, device="cuda:0").as_strided((3, 2, 5), (5, 15, 1)),
    torch.randn((30,), dtype=torch.float32, device="cuda:0").as_strided((5, 3, 2), (1, 5, 15)),
]
fd.execute(inputs)
```
This fails inferring the extent of one of the segments. In this case there are only uops so I think we're just skipping them all and maybe never marking any outputs for that group?

It might have something to do with this: https://github.com/NVIDIA/Fuser/blob/b5e55b6c07d273a630b02cf99b4cde6441fbc34f/csrc/fusion_segmenter.cpp#L3712-L3716

Maybe we need to check that the outputs of `expr` are not also Fusion outputs at line 3679?

wujingyue:
!build

csrc/fusion_segmenter.cpp
```diff
@@ -3850,6 +3850,35 @@ void SegmentCandidateFinder::findSegments() {
...
+  if (unary_use == nullptr) {
+    return nullptr;
+  }
+
+  // Don't forward an input to an output yet. Doing that would lead to an empty
+  // group that ought to work in theory but doesn't work in practice with the
+  // downstream logic. See #1813 for an example.
+  if (unary_use->out()->isFusionOutput()) {
+    return nullptr;
+  }
```
wujingyue, line 3877: This is the only behavior change since last snapshot. Other changes are refactors so we have only one function, shouldForward, to decide whether to forward an input.

test/test_segmentation.cpp
wujingyue: That's right. I didn't find a good way to test the fix either -- the most direct test would be to examine the initial segments, but those are not exposed. It may be fine to assert tryMerge only considers non-empty fusion. Would that check be too strong?

naoyam: I won't worry too much about validating this fix, but asserting non-emptiness sounds good to me as well since that shouldn't occur and if it does, I'd like to understand why.

jacobhinkle (approved):
Looks like the CI failures are unrelated. Thanks for the cleanup!

PR #1817, opened 2024-02-22
Remove function test/utils.h/matmul

zasdfgbnm:
This PR removes the `matmul` function in `test/utils.h`. In all unit tests and benchmarks, I am replacing code like
```C++
auto c = matmul(a, b, layout);
```
with
```C++
a = canonicalizeInputToBMNK(a, layout, MmaOperand::A);
b = canonicalizeInputToBMNK(b, layout, MmaOperand::B);
auto c = fusedMultiplySum(a, b, {-1});
```
I believe the latter is better because it is more explicit about how the fusion looks like, therefore more readable.

Besides, with this new coding style, it will be easier to define supported prologue fusions for matmul. For example, if I want to do `sin(A) x sin(B)`, I can just do:
```
a = canonicalizeInputToBMNK(a, layout, MmaOperand::A);
b = canonicalizeInputToBMNK(b, layout, MmaOperand::B);
auto c = fusedMultiplySum(sin(a), sin(b), {-1});
```
In the past, this was not possible (`auto c = matmul(sin(a), sin(b), layout)` will not work, because of our current limitation that the transposes must be acting on the fusion input tensor, because this transpose will be converted to `ldmatrix.trans`)

### DIFF
--- a/benchmark/matmul.cpp
+++ b/benchmark/matmul.cpp
@@ -48,7 +48,9 @@ void setupMatmul(Fusion* fusion, MmaLayout layout, MatmulParams params) {
   auto a = makeContigTensor(2, DataType::Half);
   auto b = makeContigTensor(2, DataType::Half);
 
-  auto c = matmul(a, b, layout);
+  a = canonicalizeInputToBMNK(a, layout, MmaOperand::A);
+  b = canonicalizeInputToBMNK(b, layout, MmaOperand::B);
+  auto c = fusedMultiplySum(a, b, {-1});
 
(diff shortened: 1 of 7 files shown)


zasdfgbnm:
!build

drzejan2 (approved):
Thank you @zasdfgbnm for preparing these changes.

I agree with you, the helper function limited what we could express in fusion.

zasdfgbnm:
!build

zasdfgbnm:
!build

PR #1821, opened 2024-02-22
multidevice debug disableSkip option

samnordmann:
# What
option to disable GTEST_SKIP() in multidevice tests if the multidevice setting is not available
This option doesn't modify nvFuser itself but only the multidevice tests

# How
this option is switched on by setting the env variable `NVFUSER_MULTIDEVICE_DISABLE_SKIP=1`

# Why
It is a very useful tool for debugging since it allows to run the test from gdb on a single process. Many of the debug deal with the IR lowering and thus do not involve multi-device operation per say

### DIFF
--- a/test/multidevice.cpp
+++ b/test/multidevice.cpp
@@ -28,6 +28,9 @@ void MultiDeviceEnvironment::SetUp() {
   if (getNvFuserEnv("MULTIDEVICE_DEBUG_BARRIER")) {
     do_barrier_at_test_ = true;
   }
+  if (getNvFuserEnv("MULTIDEVICE_DISABLE_SKIP")) {
+    disable_skip_ = true;
+  }
 }
 
 void MultiDeviceEnvironment::TearDown() {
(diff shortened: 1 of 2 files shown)


naoyam:
I'm wondering if these multi-device tests should not be disabled at all. They should work fine even with just one device.

naoyam:
> > I'm wondering if these multi-device tests should not be disabled at all. They should work fine even with just one device.
> 
> I am not sure to understand We need device indices in the communicator to match the one appearing in the device mesh, otherwise it makes no sense to run it, no ? What would mean resharding a tensor on [0,1] if at runtime we only have one device ?

What I was thinking about may be more related to making the distribution symbolic so that it can work with N devices, which can be 1. Think of typical MPI programs. The number of processors isn't typically hard coded but queried with the MPI routines.

I remember we discussed symbolic shaping of device meshes, but I don't remember if it's currently supported. If yes, running all tests even with just one device would be a great way to exercise a boundary condition.

samnordmann:
> > > I'm wondering if these multi-device tests should not be disabled at all. They should work fine even with just one device.
> > 
> > 
> > I am not sure to understand We need device indices in the communicator to match the one appearing in the device mesh, otherwise it makes no sense to run it, no ? What would mean resharding a tensor on [0,1] if at runtime we only have one device ?
> 
> What I was thinking about may be more related to making the distribution symbolic so that it can work with N devices, which can be 1. Think of typical MPI programs. The number of processors isn't typically hard coded but queried with the MPI routines.
> 
> I remember we discussed symbolic shaping of device meshes, but I don't remember if it's currently supported. If yes, running all tests even with just one device would be a great way to exercise a boundary condition.

I see what you mean, and I agree. So, for now, we do not support symbolic devices meshes, nor automatic setting of the devices (from the scheduler for example). So the device needs to be explicitely set with concrete values at the test level for now. In some tests, we impose a given device mesh -- those tests need to be skipped if the runtime doesn't match. In other tests, we simply take as many devices as available -- those tests should in theory not be skipped and be runnable with one single device. It is not supported right now but I note it as a TODO and I'll add a small patch for that soon.

PR #1823, opened 2024-02-22
multidevice print utils

samnordmann:
add/fixes some print utilities, notably `MultiDeviceExecutor::print()`

### DIFF
--- a/csrc/multidevice/communicator.h
+++ b/csrc/multidevice/communicator.h
@@ -35,6 +35,8 @@ using RankType = DeviceIdxType;
 // Supported backends. TODO: gloo untested
 enum class CommunicatorBackend { nccl, ucc, gloo };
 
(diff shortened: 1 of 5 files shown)


csrc/multidevice/communicator.h
```diff
@@ -35,6 +35,8 @@ using RankType = DeviceIdxType;
 // Supported backends. TODO: gloo untested
 enum class CommunicatorBackend { nccl, ucc, gloo };
 
+std::ostream& operator<<(std::ostream& out, const CommunicatorBackend& cb);
```
naoyam, line 38: Missing implementation?

samnordmann: implementation is already here https://github.com/NVIDIA/Fuser/blob/3434695576448f2dc5534deec967a60e295b27e6/csrc/multidevice/communicator.cpp#L30

It was just missing in the header

naoyam (approved):
LGTM

PR #1824, opened 2024-02-22
Allocate sharded memory

cowanmeg:
Allocates sharded tensors using their sharded size. Until this PR sharded tensorviews were bound to unsharded aten tensors. 
- Update allocation domain of device parallel (sharded) axes to be 1.
- Update index computations to skip device parallel axes since they are not allocated.
- Communication collectives updated to operate on sharded tensors. 
- Update input binding to expect device parallel axes are bound to aten tensor axes with size 1.
- Update relevant multidevice tests to accept aten tensors with sharded size. 

For example,a concrete TensorViews with unsharded size, (2,3,4), will have an allocation domain of (1,3,4). 
```
TensorView* tv = makeConcrete({2, 3, 4});
tv->mesh = DeviceMesh({0, 1});
```

Currently sharded TensorViews only support concrete TensorViews with no splits/merges on device parallel axes.


test/test_multidevice_sharding.cpp
```diff
@@ -0,0 +1,98 @@
...
+  TensorView* tv0 = makeConcreteTensor(input_size);
+  TensorView* tv1 = add(tv0, tv0);
+  TensorView* tv2 = set(tv1);
+  TensorView* tv3 = add(tv2, tv2);
+  TensorView* tv4 = set(tv3);
+  TensorView* tv5 = sum(tv4, {0});
+  fusion->addInput(tv0);
+  fusion->addOutput(tv5);
+
+  // TODO: split
```
samnordmann, line 42: remove comments ?

samnordmann (approved):
LGTM!

test/test_multidevice_sharding.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include <multidevice/utils.h>
+#include <ops/all_ops.h>
+#include <test/multidevice.h>
+#include <test/validator.h>
+
+namespace nvfuser {
+
+class ShardingTest : public MultiDeviceTest {};
+
+TEST_F(ShardingTest, UnshardedGlobalInput) {
```
naoyam, line 23: nit: please have at least one-line description about what this test is meant to verify and exercise.

test/test_multidevice_sharding.cpp
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  auto x = at::randn(unsharded_input_size, tensor_options);
+  std::vector<c10::IValue> inputs = {
+      shardTensor(x, mesh, communicator->deviceId())};
+  auto ref_outputs = x * 2;
+
+  MultiDeviceExecutor runtime(std::move(fusion), *communicator);
+  auto outputs = runtime.runWithInput(inputs);
+  testValidate(
```
naoyam, line 95: Can we also validate the sizes of the actual allocations?

csrc/device_lower/pass/allocation.cpp
```diff
@@ -163,11 +163,12 @@ class AllocationInserter : public kir::ExprMutator {
     std::vector<Val*> alloc_dims;
 
     for (const auto id : maybe_rfactor_domain) {
-      if (id->isReduction() || id->isStride() || id->isDeviceDim()) {
+      if (id->isReduction() || id->isStride()) {
         continue;
-      } else if (id->isBroadcast()) {
+      } else if (id->isBroadcast() || id->isDeviceDim()) {
```
naoyam, line 168: The change makes sense to me, but I also wonder if this should be also done with the other IterTypes, i.e., `isReduction()` and `isStride()`. For example, if a tensor only has a reduction domain, it seems that means it won't have any allocation dimension, which seems wrong. The existing schedulers should never put reduction outputs on global memory, so it may be that we haven't seen any issue with this. 

Unless I missed anything, could you try creating a simple repro and see if this is a real issue?

csrc/expr_evaluator.cpp
```diff
@@ -141,6 +141,11 @@ void ExpressionEvaluator::bind_(
             i);
         bind_(
             rfactor_domain[i]->expandedExtent(), t.size(i), evaluate_validate);
+      } else if (rfactor_domain[i]->isDeviceDim()) {
+        // Device dimensions extents will always be 1.
+        // Ignore concrete extents because they hold the unsharded extents.
```
naoyam, line 146: I'm not sure if we should just always ignore this case. At this point, we should know that `rfactor_domain[i]->extent()` is equal to the extent of `DIDx`.

cowanmeg: DIDx holds the unsharded extent. This case handles global sharded inputs: if an axis is sharded then we ignore the unsharded extent and check that it's size 1.

csrc/device_lower/pass/allocation.cpp
cowanmeg: Are there coverage metrics that can check if line 166 gets executed?

csrc/expr_evaluator.cpp
naoyam: Right, so we know the unsharded extent at this point, and that should be equal to `rfactor_domain[i]->extent()`, right?

naoyam: Well, actually, that isn't guaranteed. The only thing we can say is: `number-of-devices-in-DIDx >= rfactor_domain[i]->extent()`. So, I don't think we can bind anything in this case.

cowanmeg: This is the binding gets triggered when we have a sharded concrete tensorview input. We create the tensorview with the unsharded extents, so the unsharded extent is known. If the axis is sharded then the tensor's axis should be 1 otherwise it should be the unsharded size.

Since we don't handle split yet `number-of-devices-in-DIDx == rfactor_domain[i]->extent()`

cowanmeg: Ahh if that is the case we have a much bigger problem since all the multidevice logic currently expects if an axis is parallelized with DIDx then the extent of that axis == number of devices.

naoyam: It's okay to leave it as is for now.

naoyam: Could you leave a short summary of this conversation as a code comment?

naoyam (approved):
LGTM

csrc/device_lower/pass/allocation.cpp
cowanmeg: I ran the nvfuser_tests locally and the if branch on 166 is not triggered in any of the single device tests, which aligns with your assumption that the existing schedulers never put reductions on global memory. 

It does get executed in the multidevice_tests. Changing  `isReduction()` and `isStride()` axes to have an allocation size of 1 causes the tests to fail. 

We need to look into whether it's necessary for sharded tensors or if it will be fixed with proper memory allocation since the multidevice executor puts all TVs in global memory. (https://github.com/NVIDIA/Fuser/blob/main/csrc/multidevice/executor.cpp#L46)

cowanmeg: C++ exception with description "Expected T1_g[ bS4{1}, iS5{2}, iS6{3}, rS7{5} ] (DeviceMesh{0, 2, 3, }) to be bound to a tensor of rank 3, but got a tensor of rank 4
Exception raised from validateValWithConcreteValue at Fuser/csrc/expr_evaluator.cpp:38

naoyam: Ah, I see. That makes sense. I thought here it's just calculating the allocation size, but it's actually the shape of the allocation, so for reductions there should be no domain at all. Sorry for the noise.

PR #1830, opened 2024-02-24
IdModel: Loop promotion analysis step 3

naoyam:
Step 3 of the loop promotion analysis. 

Stacked on top of #1777.

After this, there'll be PRs for steps 4 and 5. They would be relatively smaller PRs as they are mostly just repeating steps 2 and 3.


naoyam:
!build

naoyam:
The JIT binary tests on H100 failure is due to a system problem:

```
nvidia-container-cli: detection error: nvml error: unknown error: unknown.
```

naoyam:
!build

test/test_id_model.cpp
```diff
@@ -57,14 +57,34 @@ Expr* getParentExpr(Val* val, int n) {
...
+    NVF_ERROR(!id->uses().empty());
+    NVF_ERROR(id->uses().front()->output(sibling_idx)->isA<IterDomain>());
+    id = id->uses().front()->output(sibling_idx)->as<IterDomain>();
+  }
+  NVF_ERROR(id != nullptr);
+  return id;
+};
+
+template <typename ValType>
+ValType* getValByName(const std::vector<ValType*>& vals, StmtNameType name) {
```
zasdfgbnm, line 82: I am wondering how stable the val names will be as we evolve our system. Will this become as fragile as the tests that uses
```C++
// assert that the given fusion lowers to the given CUDA kernel
void assertCUDAKernel(Fusion* fusion, const std::string& expected_kernel);
```
and we need to update this test again and again?

naoyam: Generally yes, but for these tests, I'm just using this function only for the tensor views originally defined in the initial fusion, so scheduling or ID replays won't affect their names. So, I believe they should be stable.

naoyam:
!build

csrc/id_model/id_model.cpp
```diff
@@ -1116,4 +1127,207 @@ Expr* IdModel::addReplayAs(std::vector<IterDomain*> new_inputs, Expr* expr) {
...
+      // consumer terminal. Also if there's no output groups it's id consumer
+      // terminal.
+      bool all_outs_in_loop_group = uses_it->second.empty() ? false : true;
+      for (auto use : uses_it->second) {
+        for (auto out_id : ir_utils::filterByType<IterDomain>(use->outputs())) {
+          if (group != idGraph(IdMappingMode::LOOP).toGroup(out_id)) {
+            all_outs_in_loop_group = false;
+          }
+        }
+      }
```
zasdfgbnm, line 1323:
```suggestion
            all_outs_in_loop_group = false;
            goto out;  // unfortunately C++ doesn't have a way to break nested loops.
          }
        }
      }
      out:
```

csrc/id_model/id_model.cpp
```diff
@@ -1116,4 +1127,207 @@ Expr* IdModel::addReplayAs(std::vector<IterDomain*> new_inputs, Expr* expr) {
...
+      if (uses_it == id_uses_.end()) {
+        terminal_loop_ids.pushBack(loop_id->as<IterDomain>());
+        continue;
+      }
+
+      // If there's an output group that is not in the same group, then it's id
+      // consumer terminal. Also if there's no output groups it's id consumer
+      // terminal.
+      bool all_outs_in_loop_group = uses_it->second.empty() ? false : true;
+      for (auto use : uses_it->second) {
```
zasdfgbnm, line 1317:
```suggestion
      if (all_outs_in_loop_group) {
        for (auto use : uses_it->second) {
```

csrc/id_model/id_model.cpp
```diff
@@ -1116,4 +1127,207 @@ Expr* IdModel::addReplayAs(std::vector<IterDomain*> new_inputs, Expr* expr) {
...
+
+    // Don't select producer iter domains
+    for (auto loop_id : *group) {
+      if (info.p2c_ca_permissive_maps.find(loop_id->as<IterDomain>()) !=
+          info.p2c_ca_permissive_maps.end()) {
+        continue;
+      }
+
+      auto uses_it = id_uses_.find(loop_id->as<IterDomain>());
+      if (uses_it == id_uses_.end()) {
```
zasdfgbnm, line 1308:
```suggestion
      if (uses_it == id_uses_.end() || uses_it->second.empty()) {
```

csrc/id_model/id_model.cpp
```diff
@@ -1116,4 +1127,207 @@ Expr* IdModel::addReplayAs(std::vector<IterDomain*> new_inputs, Expr* expr) {
...
+        terminal_loop_ids.pushBack(loop_id->as<IterDomain>());
+        continue;
+      }
+
+      // If there's an output group that is not in the same group, then it's id
+      // consumer terminal. Also if there's no output groups it's id consumer
+      // terminal.
+      bool all_outs_in_loop_group = uses_it->second.empty() ? false : true;
```
zasdfgbnm, line 1316:
```suggestion
      bool all_outs_in_loop_group = true;
```

zasdfgbnm (commented):
Not done reading yet. Just posting some minor comment so far.

csrc/id_model/id_model.cpp
```diff
@@ -1116,4 +1127,207 @@ Expr* IdModel::addReplayAs(std::vector<IterDomain*> new_inputs, Expr* expr) {
...
+  std::unordered_map<ValGroup, IterDomain*> promotion_map;
+
+  // Grab all the (potentially promoted) terminal iter domains in this group.
+  // Save the exact group and the iter domain in this vector.
+  std::vector<std::pair<ValGroup, IterDomain*>> exact_promoted_terminal_ids;
+  for (auto loop_id : *loop_group) {
+    // If not a terminal id in the group skip
+    if (!terminal_loop_ids.has(loop_id->as<IterDomain>())) {
+      continue;
+    }
```
zasdfgbnm, line 1237: Instead of computing the terminal loop ids for all loop groups altogether before calling `findPromotionOfLoopGroup` and use it to filter inside `findPromotionOfLoopGroup`, would it make the code more readable if we just compute the terminal ids for `loop_group` here? Something like:
```suggestion
  for (auto loop_id : computeTerminalIdsOfGroup(loop_group)) {
```

csrc/id_model/id_model.cpp
```diff
@@ -1116,4 +1127,207 @@ Expr* IdModel::addReplayAs(std::vector<IterDomain*> new_inputs, Expr* expr) {
...
+
+  return loop_promotion_map;
+}
+
+IterDomain* IdModel::findPromotionOfLoopGroup(
+    const ValGroup& loop_group,
+    const ValGraph& iel_graph,
+    const std::unordered_map<ValGroup, IterDomain*>& iel_promotion_map,
+    const std::unordered_map<ValGroup, ValGroups>& exact_covered_ids,
+    const VectorOfUniqueEntries<IterDomain*>& terminal_loop_ids) {
```
zasdfgbnm, line 1225: Why do we only consider terminal ids? Is it because of forwarding and we only want to consider the most forwarded ID?

csrc/id_model/id_model.cpp
```diff
@@ -1116,4 +1127,207 @@ Expr* IdModel::addReplayAs(std::vector<IterDomain*> new_inputs, Expr* expr) {
...
+  const ValGraph& exact_graph = idGraph(IdMappingMode::EXACT);
+
+  std::unordered_map<ValGroup, IterDomain*> promotion_map;
```
zasdfgbnm, line 1228: Is this used?

test/test_id_model.cpp
```diff
@@ -567,36 +630,43 @@ TEST_F(IdModelTest, LoopPromotion1) {
...
+        tester.s1_root_resolution_map);
 
     // Check Step 2 results
     // Nothing to propagate in this fusion, so iel_promotion_map
     // should be equivalent to root_resolution_map
-    ASSERT_EQ(root_resolution_map, iel_promotion_map)
+    ASSERT_EQ(tester.s1_root_resolution_map, tester.s2_iel_promotion_map)
         << "Unexpected IEL promotion map";
+
+    // Check Step 3 results. See the design doc for the expected results
```
zasdfgbnm, line 660: Instead of saying "see the design doc", should we instead make the nvFuser repo self-contained? Options are:
1. open source the design doc, reformat it as a markdown file in `doc/dev`
2. put the drawings somewhere, for example `doc/tests/xxx.png`, and mention the filename in the comments.

test/test_id_model.cpp
```diff
@@ -849,9 +974,66 @@ TEST_F(IdModelTest, LoopPromotion5) {
...
+               tv2->getRootDomain().at(2),
+               tv3->getRootDomain().at(3),
+               tv4->getRootDomain().at(3)},
+           tv4->getRootDomain().at(3)},
+          // 32, 26, 19 -> 19
+          {std::unordered_set<Val*>{
+               getParentId(tv2->axis(0), 3),
+               getParentId(tv3->axis(0), 3),
+               getParentId(tv4->axis(0), 3)},
+           getParentId(tv4->axis(0), 3)},
```
zasdfgbnm, line 1011: nit: would it be more readable if we use:
```suggestion
               getChildId(tv2->getRootDomain().at(0), 1),
               getChildId(tv3->getRootDomain().at(2), 1),
               getChildId(tv3->getRootDomain().at(2), 1)},
           getChildId(tv3->getRootDomain().at(2), 1)},
```

test/test_id_model.cpp
```diff
@@ -935,9 +1123,113 @@ TEST_F(IdModelTest, LoopPromotion6) {
...
+  auto id79 =
+      getValByName(ir_utils::consumerValsOf(tv9->getRootDomain().at(2)), 79)
+          ->as<IterDomain>();
+  ASSERT_NE(id79, nullptr) << "IterDomain 79 not found";
+  auto id80 =
+      getValByName(ir_utils::consumerValsOf(tv9->getRootDomain().at(2)), 80)
+          ->as<IterDomain>();
+  ASSERT_NE(id80, nullptr) << "IterDomain 80 not found";
+  auto id81 = getChildId(id79, 1);
+  ASSERT_EQ(id81->name(), 81);
```
zasdfgbnm, line 1139: Thank you for adding this check.

test/test_id_model.cpp
```diff
@@ -935,9 +1123,113 @@ TEST_F(IdModelTest, LoopPromotion6) {
...
+  auto id87 = getChildId(id83, 1);
+  ASSERT_EQ(id87->name(), 87);
+  auto id88 = getChildId(id83, 1, 1);
+  ASSERT_EQ(id88->name(), 88);
+
+  // Check Step 3 results. See the design doc for the expected results
+  std::vector<std::pair<std::unordered_set<Val*>, IterDomain*>>
+      s3_reference_map = {
+          // 1, 2, 3, 46, 6, 7, 41, 8, 9, 10, 35, 36, 11, 12, 61, 15, 16, 56,
+          // 17, 18, 29, 30, 79, 80 -> 79
```
zasdfgbnm, line 1159: Does it also include 19?

zasdfgbnm: Also, should it be `->80` instead of `->79`?

test/test_id_model.cpp
```diff
@@ -935,9 +1123,113 @@ TEST_F(IdModelTest, LoopPromotion6) {
...
+               tv8->axis(0),
+               tv9->axis(0),
+               id85,
+               id87},
+           id87},
+          // 48 -> 43
+          {std::unordered_set<Val*>{tv2->axis(2)}, tv4->axis(2)},
+          // 50 -> 45
+          {std::unordered_set<Val*>{tv2->axis(1)}, tv4->axis(1)},
+          // 40, 88 -> 88
```
zasdfgbnm, line 1215: Why 88 is grouped with 40, instead of 34? Is this arbitrary?

This makes me feel generally worried about doing a `mapThroughExpr` on the loop graph. For example, if we have:

```
I0   I1        I5   I6
 \  /           \   /
  I2      ->     I7
 /  \           /  \
I3 | I4        I8  I9
```
Then `I2` and `I7` will be in the same loop group. If we do a `mapThroughExpr` on this loop group forward, will we map `I4` and `I9`, because they are equivalent exprs?

I think mathematically, `mapThroughExpr` is assuming the "congruence" property: if x == y, then f(x) == f(y), but unfortunately loop graph is not a graph with congruence property, so now I am deeply worried about this line: https://github.com/NVIDIA/Fuser/blob/4df003a526561987b1be085c5dd286ddd964d87d/csrc/id_model/id_model.cpp#L1112

BTW: I learnt the word "congruence" from https://doi.org/10.1145/3434304, thanks @jacobhinkle for sharing this paper with me.

zasdfgbnm (commented):
Not done yet, but I believe I should post early, because I have a real worry to discuss.

test/test_id_model.cpp
```diff
@@ -1094,9 +1416,60 @@ TEST_F(IdModelTest, LoopPromotion8) {
...
-      iel_graph,
+      tester.iel_graph,
       tester.idGraph(IdMappingMode::EXACT),
-      iel_promotion_map);
+      tester.s2_iel_promotion_map);
+
+  // Check Step 3 results. See the design doc for the expected results
+  std::vector<std::pair<std::unordered_set<Val*>, IterDomain*>>
+      s3_reference_map = {
+          // 1, 2, 3, 20, 6, 7, 17, 8, 9, 26, 14, 15, 29
```
zasdfgbnm, line 1426:
```suggestion
          // 1, 2, 3, 20, 6, 7, 17, 8, 9, 26, 14, 15, 29 -> 29
```

zasdfgbnm: Also, the doc for this needs to be updated. The arrow for this group is missing in the doc.

test/test_id_model.cpp
```diff
@@ -1094,9 +1416,60 @@ TEST_F(IdModelTest, LoopPromotion8) {
...
+               tv4->getRootDomain().at(1),
+               getChildId(tv4->getRootDomain().at(0), 1),
+               tv5->getRootDomain().at(0),
+               tv5->getRootDomain().at(1),
+               getChildId(tv5->getRootDomain().at(0), 1),
+               tv7->getRootDomain().at(0),
+               tv7->getRootDomain().at(1),
+               getChildId(tv7->getRootDomain().at(0), 1)},
+           getChildId(tv7->getRootDomain().at(0), 1)},
+          // 35, 21, 18, 27, 30 -> 30
```
zasdfgbnm, line 1442: Arrow for this is missing in the doc.

test/test_id_model.cpp
```diff
@@ -1094,9 +1416,60 @@ TEST_F(IdModelTest, LoopPromotion8) {
...
+          {std::unordered_set<Val*>{
+               tv1->axis(0),
+               tv2->axis(0),
+               tv4->axis(0),
+               tv5->axis(0),
+               tv7->axis(0)},
+           tv7->axis(0)},
+          // 28, 10, 39, 31, 16, 42 -> 42
```
zasdfgbnm, line 1450: Arrow for this is missing in the doc.

zasdfgbnm (approved):
Finished reading. Don't see any blocker.

test/test_id_model.cpp
zasdfgbnm: > Does this address your concern?

Yes, it does. Thank you!

So this is the line that adds the replayed ID to some loop group?
https://github.com/NVIDIA/Fuser/blob/3281aa10727af3e01c974f6f13288d0c157bb776/csrc/id_model/id_model.cpp#L996-L997

csrc/id_model/id_model.cpp
naoyam: If there's no output, it should be considered a terminal ID, so this flag should be false. The below loop won't do anything if there's no output, so there must be something to set this flag to false when there's no output.

naoyam: Ah, I didn't read your above comment. This change makes sense now. Thanks for the suggestion.

csrc/id_model/id_model.cpp
naoyam: If an ID is not terminal, that means there's an consumer ID that's also in the same loop group, and that in turn means the non-terminal ID should never be the representative ID of this loop group. So, yes, we should only look for most forwarded IDs.

That said, I believe it is not required to consider only terminal IDs since the check around line 1278 should exclude non-terminal IDs anyway.

csrc/id_model/id_model.cpp
naoyam: It would definitely make the code simpler. Let me update the code.

test/test_id_model.cpp
naoyam: Yes.

test/test_id_model.cpp
```diff
@@ -849,9 +974,66 @@ TEST_F(IdModelTest, LoopPromotion5) {
...
+
+  // Check Step 3 results. See the design doc for the expected results
+  std::vector<std::pair<std::unordered_set<Val*>, IterDomain*>>
+      s3_reference_map = {
+          // 7, 10, 11, 25, 14, 15, 18 -> 18
+          {std::unordered_set<Val*>{
+               tv2->getRootDomain().at(0),
+               tv3->getRootDomain().at(0),
+               tv3->getRootDomain().at(1),
+               getParentId(tv3->axis(0), 4),
```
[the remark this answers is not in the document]
naoyam: Agree it'd be simpler, but since a replay may add a new expr using the same domain, traversing from parents to children may not always work.

test/test_id_model.cpp
naoyam: Yeah, I'll think about it. I'd wait until all the steps of the loop promotion analysis are completed.

csrc/id_model/id_model.cpp
naoyam: I'll do that later. https://github.com/NVIDIA/Fuser/issues/1924

csrc/id_model/id_model.cpp
naoyam: Simplified the loop structure.

naoyam:
!build

naoyam:
Thanks for the review @zasdfgbnm. I'll merge the PR once the tests are done. I'll update the doc as well.

PR #1831, opened 2024-02-26
Correct typo in MultiDeviceExecutor::print

samnordmann:
fix build error
```
csrc/multidevice/executor.cpp:268:7: error: variable 'communication_counter' set but not used [-Werror,-Wunused-but-set-variable]
  int communication_counter = 0;
      ^
```

Thanks @naoyam


naoyam (approved):
LGTM

PR #1833, opened 2024-02-26
Multidevice tests with unsharded inputs

samnordmann:
In this pr, we simplify the multidevice test validation implementation and make it run faster.

Before, we would allocate a shard of the inputs at each rank. Then, for validation, we needed to gather at a tester rank all the inputs and outputs shards, and compare them with the unsharded output obtained from running the Fusion on a single device with the unsharded inputs. This flow could be summarized as follows:
```
execution:
for all RANK k
- allocate input's shard k
- run Fusion on multidevice with input's shard k  --> get output's shard k

validation:
TESTER RANK:
- gather all input shards from other ranks 
- run Fusion on single device with unsharded input --> get unsharded output
- gather all outputs shards slices from other ranks
- compare all output shards with unsharded outputs
```
This flow is somehow cumbersome to implement and is not scalable because it requires a lot of network communications and synchronizations.

With this patch, we get rid of all network communication at test validation. All ranks initialize the same unsharded input, even though only their corresponding slice is used in the Multidevice Execution. Then they compare the output obtained through multidevice execution with the one obtained through single device execution with unsharded input. This way, validation requires no network communication, which improve the logic and the perf by a lot. 
```
execution:
for all RANK k
allocate global input (identitical accross ranks)
run Fusion with input's shard k --> get output's shard k

validation:
for all RANK k
- run Fusion on a single device with unsharded input --> get unsharded output
- compare output shard k and with a slice of unsharded outputs
```
The only "drawback" of this method is that we make the assumption that, in the tests, all ranks initialize identical inputs. However, that sounds like a much reasonable assumption. This assumption is ensured by taking the same random seed accross ranks.

### DIFF
--- a/csrc/multidevice/lower_communication.cpp
+++ b/csrc/multidevice/lower_communication.cpp
@@ -443,13 +443,6 @@ std::vector<std::shared_ptr<Communication>> lowerCommunication(
           static_cast<size_t>(output_tensor.size(0)) == 1,
       "Sharded dimension should have allocation size 1, but is ",
       output_tensor.size(0));
-  NVF_ERROR(!sender_mesh.vector().empty(), "sender mesh is empty");
-  NVF_ERROR(!receiver_mesh.vector().empty(), "receiver mesh is empty");
-
-  if (!isDeviceInvolved(my_device_index, sender_mesh, receiver_mesh)) {
-    return {};
-  }
-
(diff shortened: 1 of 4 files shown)


cowanmeg (approved):
LGTM!

naoyam (approved):
LGTM

PR #1834, opened 2024-02-26
Respect device smem constraint in `generateSharedMemoryEpilogueHeuristics`

jacobhinkle:
`mma_utils::generateSharedMemoryEpilogueHeuristics` is used to estimate whether it is possible and advantageous to use smem for epilogue unswizzling. Smem epilogue might increase the required memory even if we can re-use the operand buffer space, which can put us past the shared memory limit per thread. Even if it fits, it might not be advantageous as the shared memory per SM limit might cause a decrease in occupancy.

In settings where it is advantageous to use smem for the epilogue, we also estimate whether we should "promote" (i.e. enable) smem reuse. Since re-using operand smem buffer space requires an additional block sync, we don't enable reuse unless we either have to due to device constraints or if we detect that it will lead to an occupancy increase.

There is currently a bug wherein we correctly detect that we should use smem for the epilogue, but we incorrectly choose not to promote smem reuse. This leads to an execution error where we request more than the available amount of smem.

The problem is that when not ignoring occupancy drop, we _only_ look at occupancy and forgot to check that unreused smem epilogue would fit within device constraints.


liqiangxl (approved):
LGTM. You mensioned `This leads to an execution error`, we already have this test case?

jacobhinkle:
> LGTM. You mensioned `This leads to an execution error`, we already have this test case?

I will add the test case that fails on A100. It's device specific.

jacobhinkle:
Converting to draft until I can test more extensively, to prevent myself from merging by accident.

jacobhinkle:
!build

jacobhinkle:
!build

PR #1837, opened 2024-02-26
revise target occupancy in inner persistent heuristics

liqiangxl:


jjsjann123:
As discussed offline, we'll shelve this PR as a backup last resort quick patch in case needed.

Otherwise, we'll revisit the heuristics after updating the logic to compute block_size from persistent batch size, which should help iron out some corner cases where our current occupancy was undermined by bad block_size choices.

PR #1840, opened 2024-02-26
Dynamic input shapes with sharding

cowanmeg:
Adds support for dynamic shape on sharded tensors. Previously, if a global tensor was sharded we could only support concrete sized inputs. 

- Updates tensor metadata to store logical size as the unsharded tensor size when the tensor is sharded. 
- Sharding tests checks both concrete and symbolic sized input tensors.


csrc/multidevice/utils.cpp
```diff
@@ -192,4 +192,19 @@ std::set<DeviceIdxType> involvedDevices(Expr* expr) {
...
+    TensorView* tv,
+    c10::IntArrayRef sharded_sizes) {
+  std::vector<int64_t> unsharded_sizes;
+  std::copy(
+      sharded_sizes.begin(),
+      sharded_sizes.end(),
+      std::back_inserter(unsharded_sizes));
+  if (isSharded(tv)) {
+    int num_devices = tv->getDeviceMesh().vector().size();
+    unsharded_sizes[0] = num_devices;
```
naoyam, line 205: Please leave a note that:

- The outermost dimension is assumed to be the only sharded domain if the tensor is sharded
- The extent of the sharded dimension is assumed to be equal to the number of devices of the mesh

Both are not necessarily true in general.

naoyam (approved):
LGTM

samnordmann (approved):

cowanmeg:
!build

cowanmeg:
!build

cowanmeg:
!build

PR #1841, opened 2024-02-27
Python API (FusionDefinition::execute) to support tuple for vector inputs

jjsjann123:
Changing the FusionDefinition's execute method to support both`List` and `Tuple` for vector of scalars.

### DIFF
--- a/csrc/python_frontend/python_bindings.cpp
+++ b/csrc/python_frontend/python_bindings.cpp
@@ -543,7 +543,8 @@ void initNvFuserPythonBindings(PyObject* module) {
             std::vector<c10::IValue> inputs;
             for (py::handle obj : iter) {
               // Allows for a Vector of Sizes to be inputed as a list
-              if (py::isinstance<py::list>(obj)) {
+              if (py::isinstance<py::list>(obj) ||
+                  py::isinstance<py::tuple>(obj)) {
                 for (py::handle item : obj) {
                   inputs.push_back(
                       torch::jit::toIValue(item, c10::AnyType::get()));
--- a/csrc/python_frontend/python_bindings.cpp
+++ b/csrc/python_frontend/python_bindings.cpp
@@ -542,7 +542,7 @@ void initNvFuserPythonBindings(PyObject* module) {
              bool capture_debug_output) {
             std::vector<c10::IValue> inputs;
             for (py::handle obj : iter) {
-              // Allows for a Vector of Sizes to be inputed as a list
+              // Allows for a Vector of Sizes to be inputed as a list/tuple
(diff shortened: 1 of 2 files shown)


python_tests/test_python_frontend.py
```diff
@@ -539,6 +539,32 @@ def fusion_func(fd: FusionDefinition):
...
+            t2 = fd.ops.sum(t1, dims=[0])
+
+            fd.add_output(t2)
+
+        eager_out = torch.sum(inputs_with_list[0].reshape(new_shape), dim=0)
+
+        nvf_out, _ = self.exec_nvfuser(fusion_func, inputs_with_list)
+        self.assertEqual(eager_out, nvf_out[0])
+
+        inputs_with_tuple = [tensor, tuple(new_shape)]
```
tfogal, line 563: i realize it's not really related to this patch, but if the inputs themselves are tuples:
```python
inputs_with_tuple = (tensor, tuple(new_shape))
```
or even:
```python
inputs_with_tuple = (tensor, new_shape)
```
would we expect that to work? does it need a test?

tfogal (approved):
LGTM, thanks. Question on a potential follow-up.

csrc/python_frontend/python_bindings.cpp
```diff
@@ -543,7 +543,8 @@ void initNvFuserPythonBindings(PyObject* module) {
             std::vector<c10::IValue> inputs;
             for (py::handle obj : iter) {
               // Allows for a Vector of Sizes to be inputed as a list
-              if (py::isinstance<py::list>(obj)) {
+              if (py::isinstance<py::list>(obj) ||
```
tfogal, line 546: oh, also, please update the comment on line 545, too.

python_tests/test_python_frontend.py
jjsjann123: input here is a python iterable. so any container should work.

https://github.com/NVIDIA/Fuser/blob/af04aadb12b4b2e51eec5e17266f34e53b2a4430/csrc/python_frontend/python_bindings.cpp#L539

I wouldn't worry about it.

jjsjann123:
!build

kevinstephano:
It would be good to note that you are changing the FusionDefinition's execute method in the PR comment.

jjsjann123:
H100 failure is persistent, ignoring.

The stable wheel build failure looks to be a machine issue as well. I'm merging as-is. cc'ing @xwang233 
```
Cloning into 'Fuser'...
error: 566 bytes of body are still expected
fetch-pack: unexpected disconnect while reading sideband packet
```

PR #1843, opened 2024-02-27
Fix MultiDeviceExecutor::requestedNumberOfDevices

samnordmann:
A small fix to avoid some hangs in the tests in case the runtime doesn't match the required multi-device setting.

More precisely, imagine we shard a tensor on the device mesh [0,2], and that, at runtime, the communicator is of size 2, then, there will be no device with index 2, so the runtime is incompatible.

Before this patch, the above situation would not be detected as an error.


cowanmeg (approved):
LGTM!

naoyam (approved):

Issue #1553, opened 2023-12-22
Failure to segment cat-based fusion


jacobhinkle:
```python
import torch
from nvfuser import FusionDefinition, DataType

def nvfuser_fusion_id1(fd : FusionDefinition) -> None :
    T0 = fd.define_tensor(shape=[-1, -1, -1, -1], contiguity=[True, True, False, True], dtype=DataType.Float, is_cpu=False, stride_order=[3, 1, 2, 0])
    T1 = fd.define_tensor(shape=[-1, -1, -1, -1], contiguity=[True, True, False, True], dtype=DataType.Float, is_cpu=False, stride_order=[3, 1, 2, 0])
    T2 = fd.define_tensor(shape=[-1, -1, -1, -1], contiguity=[True, True, False, True], dtype=DataType.Float, is_cpu=False, stride_order=[3, 1, 2, 0])
    T3 = fd.define_tensor(shape=[-1, -1, -1, -1], contiguity=[None, None, True, True], dtype=DataType.Float, is_cpu=False, stride_order=[3, 2, 1, 0])
    T4 = fd.define_tensor(shape=[-1, -1, -1, -1], contiguity=[None, None, True, True], dtype=DataType.Float, is_cpu=False, stride_order=[3, 2, 1, 0])
    T5 = fd.define_tensor(shape=[-1, -1, -1, -1], contiguity=[True, True, False, True], dtype=DataType.Float, is_cpu=False, stride_order=[3, 1, 2, 0])
    T6 = fd.define_tensor(shape=[-1, -1, -1, -1], contiguity=[True, True, False, True], dtype=DataType.Float, is_cpu=False, stride_order=[3, 1, 2, 0])
    T7 = fd.define_tensor(shape=[-1, -1, -1, -1], contiguity=[True, True, False, True], dtype=DataType.Float, is_cpu=False, stride_order=[3, 1, 2, 0])
    T8 = fd.define_tensor(shape=[-1, -1, -1, -1], contiguity=[None, None, True, True], dtype=DataType.Float, is_cpu=False, stride_order=[3, 2, 1, 0])
    T9 = fd.define_tensor(shape=[-1, -1, -1, -1], contiguity=[None, None, True, True], dtype=DataType.Float, is_cpu=False, stride_order=[3, 2, 1, 0])
[...]

PR #1844, opened 2024-02-27
Option to skip scheduling in FusionExecutorCache

samnordmann:
This PR is comprised of mainly two dependent parts:

### Skip scheduling

We add the ability in `FusionExecutorCache` to skip scheduling the kernel with the heuristics. 

This feature allows the user to use manual scheduling instead of necessarily using the heuristic suggested by 
the auto-scheduler.

This feature is enabled by passing a bool argument `auto_scheduling` to `FusionExecutorCache`. This argument is optionnal and set to `true` by default, thus the default behavior is not changed. We want to keep this less permissive behavior as default to avoid uncaught errors.

### Use correct executor in `MultiDeviceExecutor`

Currently, we are using manual multi-device scheduling. Therefore the above feature is necessary, since auto-scheduler is not compatible with manual scheduling.

As a work around, we were using so far`FusionExecutor` in the `MultiDeviceExecutor`. However, using `FusionExecutor`, is faulty here (as pointed out by @wujingyue), and we need to use `FusionExecutorCache` instead (which is anyway the main entry point for nvFuser runtime). Thanks to the first feature of this pr, we are able to use `FusionExecutorCache` inside `MultiDeviceExecutor`. Moreover, this patch allows benefiting from the cache and other goodies that accelerate the flow.

### TODOs
Later, we might wanna 1) modify all the existing scheduler to integrate the multidevice aspect, and 2) get rid of `MultiDeviceExecutor` and integrate it into `FusionExecutorCache` itself. In the meantime this patch is relatively simple and allows us to go forward.


samnordmann:
!build

csrc/multidevice/executor.cpp
```diff
@@ -155,12 +155,13 @@ void MultiDeviceExecutor::postKernel(SegmentedGroup* group) {
...
   // Check if the executor has been cached. If not, create and cache it
-  if (fe_.find(group) == fe_.end()) {
-    fe_.emplace(group, std::make_unique<FusionExecutor>());
-    fusions_.emplace(group, staged_fusion_->makeFusion(group));
-    fe_[group]->compileFusion(fusions_.at(group).get(), group_input_IValues);
+  if (fec_.find(group) == fec_.end()) {
+    fec_.emplace(
+        group,
+        std::make_unique<FusionExecutorCache>(
+            staged_fusion_->makeFusion(group), 0, auto_schedule_));
```
wujingyue, line 162: Is `auto_schedule_` here always false?

wujingyue, line 162: Can you clarify the reason for disabling auto scheduling? I guess: because of multi-level segmentation, `staged_fusion` may have to be further segmented before it can be auto-scheduled.

csrc/multidevice/executor.cpp
```diff
@@ -86,24 +86,24 @@ std::unordered_map<Val*, c10::IValue> MultiDeviceExecutor::allocateRecvBuffers(
   if (fusion_copy->outputs().empty()) {
     return {};
   }
-  // TODO: Not working with FusionExecutorCache
-  FusionExecutor fe;
-  fe.compileFusion(fusion_copy.get(), global_inputs_IValues);
-  auto buffers = fe.allocOutputSpace(global_inputs_IValues);
+
+  FusionExecutorCache fec(std::move(fusion_copy), 0, false);
```
cowanmeg, line 90: Can we also save this fec in the executor so it is only compiled once?

cowanmeg (approved):
LGTM. This will speedup the tests a lot!

csrc/multidevice/executor.cpp
samnordmann: not necessarily, but for now we only have multidevice tests that set it to false. This is because we first need a patch on the scheduler to support multi-device. We are gonna open PRs soon about that

samnordmann: We disable auto-scheduling because there is only manual scheduling available for multi-device, which is not compatible with auto-scheduling (I tried to clarify this is in the PR description)

csrc/multidevice/executor.cpp
samnordmann: sure, good idea

samnordmann: Actually the best would be to simply cache the necessary allocations sizes, this is even more lightweight. I will do it on a separate PR.

Note that in the (hopefully near :) ) future we will melt `MultiDeviceExecutor` into the main path of `FusionExecutorCache`. Then, the allocation will be taken care of automatically there, and so, in my opinion, let's not worry to much about the allocation flow we have here.

csrc/multidevice/executor.cpp
samnordmann: see https://github.com/samnordmann/Fuser/tree/staged_reduction_tests_upstream for an example where we turn auto-scheduling on combined with manual multi-device scheduling. I will open that pr once this one is merged

csrc/kernel_cache.cpp
```diff
@@ -1337,7 +1344,9 @@ void FusionKernelRuntime::compileKernel(
     fusion_to_run->printMath();
   }
   FusionGuard fg(fusion_to_run.get());
-  scheduler_entry->schedule(fusion_to_run.get());
+  if (auto_schedule) {
```
naoyam, line 1347: Is this the only place disabled when `auto_schedule` is false?

csrc/kernel_cache.h
```diff
@@ -756,6 +762,8 @@ class FusionExecutorCache {
   // ID of fusion in python frontend fusion cache, which maps to a single
   // FusionExecutorCache.
   int64_t fusion_id_ = -1;
+
+  bool auto_schedule_ = true;
```
naoyam, line 766: Please add a comment.

csrc/kernel_cache.cpp
samnordmann: yes, this is the only place where this flag is used, otherwise it is just passed around through the interfaces.

naoyam: Including the segmentation?

samnordmann: yes. The segmentation happens before.

naoyam: I'm trying to understand what `auto_schedule` means here since the segmentation is effectively almost part of the scheduling.

I think what's not very clear is what is assumed to be done with the given fusion. Since `schedule` is not called, the fusion is assumed to be scheduled, but since the segmenter kicks in, I'm a bit surprised that the segmenter actually works with a pre-scheduled fusion.

csrc/multidevice/executor.cpp
wujingyue: Question to @naoyam for my education: regardless of multi-device, how does FusionExecutorCache behave when the input fusion is already (partially) scheduled (e.g. using `tv2->split`)? Does it overwrite those schedules or try to preserve them?

csrc/kernel_cache.cpp
```diff
@@ -534,7 +537,7 @@ std::vector<at::Tensor> FusionExecutorCache::runFusionWithInputs(
   }
 
   if (!kernel_runtime->isCompiled()) {
-    kernel_runtime->compileFusionParallel(args);
+    kernel_runtime->compileFusionParallel(args, auto_schedule_);
```
wujingyue, line 540: Instead of passing `auto_schedule` everywhere, I think it's more convenient for you and the readers to embed auto_schedule as a field of FusionKernelRuntime (as how you added auto_schedule_ to FusionExecutorCache). getKernelRuntimeFor should be responsible to construct (when cache miss) a FusionKernelRuntime with the right auto_schedule config.

csrc/multidevice/executor.h
```diff
@@ -115,13 +118,15 @@ class MultiDeviceExecutor {
...
-  std::unordered_map<SegmentedGroup*, std::unique_ptr<Fusion>> fusions_;
+  // Cache FusionsExecutorCache
+  std::unordered_map<SegmentedGroup*, std::unique_ptr<FusionExecutorCache>>
+      fec_;
   // Cache whether a SegmentedGroup should be run by the current device
   std::unordered_map<SegmentedGroup*, bool> should_run_;
   // Cache whether a SegmentedGroup requires inter-device communication
   std::unordered_map<SegmentedGroup*, bool> is_resharding_;
+  // Whether to apply auto-scheduling in FusionExecutorCache
+  bool auto_schedule_ = false;
```
wujingyue, line 129:
```suggestion
  const bool auto_schedule_ = false;
```

test/multidevice.h
```diff
@@ -99,6 +99,7 @@ class PipelineTest : public MultiDeviceTest {
   std::vector<c10::IValue> inputs;
   std::vector<c10::IValue> unsharded_inputs;
   std::vector<at::Tensor> outputs;
+  bool auto_schedule_in_execute = false;
```
wujingyue, line 102: Does `_in_execute` bear a special meaning?

csrc/multidevice/executor.cpp
naoyam: It assumes no scheduling is done, so I'm actually a bit surprised it doesn't fail 😆

wujingyue, line 162: I doubt the use of unique_ptr is necessary, which adds execution overhead and indirection for readers. I *think* something like 
```suggestion
    fec_.try_emplace(
        group,
        staged_fusion_->makeFusion(group), 0, auto_schedule_);
```
should just work. (Of course, you'd need to change fec_'s definition as well)

wujingyue (approved):
LGTM with comments. I'd also wait for @naoyam 's approval.

naoyam:
Thinking about this more closely, it seems there's some fundamental inconsistency. First of all, scheduling dictates how kernels are cached by `FusionExectutorCache` as the scheduling parameters are used as a key to the caching map. In this PR, however, since the scheduling is done before `FusionExecutorCache`, it seems that `FusionExecutorCache` needs to extract the corresponding scheduling parameters to build the correct key to the cache. However, IIUC, the segmentation is still done even with a prescheduled fusion, so there's no guarantee that the scheduling parameters obtained by the segmenter is indeed equal to the scheduling parameters corresponding to the pre-scheduled fusion. This also would mean that the kernel cache would use an incorrect caching key for the pre-scheduled fusion.

Does this make sense?

wujingyue:
> Thinking about this more closely, it seems there's some fundamental inconsistency. First of all, scheduling dictates how kernels are cached by `FusionExectutorCache` as the scheduling parameters are used as a key to the caching map. In this PR, however, since the scheduling is done before `FusionExecutorCache`, it seems that `FusionExecutorCache` needs to extract the corresponding scheduling parameters to build the correct key to the cache. However, IIUC, the segmentation is still done even with a prescheduled fusion, so there's no guarantee that the scheduling parameters obtained by the segmenter is indeed equal to the scheduling parameters corresponding to the pre-scheduled fusion. This also would mean that the kernel cache would use an incorrect caching key for the pre-scheduled fusion.
> 
> Does this make sense?

Yes, it makes sense. In practice, what does the so-called "pre-scheduling" (for which I believe the code is outside this PR) do? IIUC, for the short/medium term, pre-scheduling is done by the user of nvFuser and it only annotates the device IDs. After that, the "high-level" segmentation kicks in and divide the fusion at ops that require re-sharding. If that's the case, does pre-scheduling need to be done by `FusionExecutorCache(auto_schedule=true)` at all? 

PS: I think, when we later consolidate the two executors, we won't need multi-level segmentation and therefore won't need two `FusionExecutorCache`s, which can simplify a lot. Alas, I doubt I'll have time to implement that before GTC...

test/multidevice.h
samnordmann: Originally, for distinguishing with "_in_validate", but I will remove it

samnordmann:
(The text I ended up writing is very long, sorry about that!  I'm doing my best to make it as clear as possible. But let's meet to talk about it!)

What has been said so far makes a lot of sense to me and triggers some thoughts. Let me try to summarize/comment the different topics.

### 1. Cache
> scheduling dictates how kernels are cached

> the kernel cache would use an incorrect caching key for the pre-scheduled fusion

I think that this problem is intrinsic to the current cache mechanism, and not related to this PR per say. IIUC, the key for the cache looks at the heuristic type and the schedule params, instead of looking at the actual scheduling of the Fusion. So this cache mechanism becomes faulty whenever there is the slightest manual modification. For example, nothing prevents a user today from doing:

    i. defining a fusion
    ii. execute through FusionExecutorCache with concrete inputs
    iii. modify the fusion's schedule manually
    iv. execute again through the same instance of FusionExecutorCache with same input shapes

In this case the second execution will probably simply discard the schedule modifications. This is, in my opinion, similar to the inconsistency you are pointing at @naoyam. I think that the cache would be greatly improved if it could look at the actual scheduling of the fusion (rather than on the params), so it wouldn't rely entirely on the automatic segmentation/scheduling happening under the hood.

### 2. manual scheduling
#### 2.a. current state

Today, if I manually schedule a fusion: sometimes it works without reporting any error, and in other cases I run into [this assert error](https://github.com/NVIDIA/Fuser/blob/71d9b555b401655c0d664b94e8abb7021ce29637/csrc/tensor_view.cpp#L665) that a parallelized axis cannot be split. Note that this happens during scheduling (i.e. exactly what the flag `auto_schedule` turns off in the PR), not during segmentation nor cache lookup. So the error happens "by chance", and in some situations, there is no error at all. There is no thorough error checking that the scheduling is correct, nor a brute-force check that the fusion is not pre-scheduled.

For now it is not clear to what extent nvFuser allows the user to manually schedule. On the one hand, we heavily rely on auto-scheduling, on the other hand, the scheduling utilities are parts of the API and I feel that we focus a lot on those aspects during tutorial/presentations and present nvFuser as a framework for parallelization.


#### 2.b. what to do in the future


In my opinion, it is important to support in the future manual scheduling, and ideally composing manual and auto-scheduling. It would make both development and utilization more flexible and rich. As far as multidevice development is concerned, I have been needing support for manual scheduling from day one and am spending a lot of effort coming up with workarounds like this one. Let's also keep in mind that multidevice sharding changes the actual layout of the I/O, so the user needs to be aware of it anyway. 

#### how to do in the meantime?

If not using `FusionExecutorCache`, then, how is the good way to execute a pre-scheduled function ? I am currently using `FusionExecutor` [in the multidevice executor](https://github.com/NVIDIA/Fuser/blob/71d9b555b401655c0d664b94e8abb7021ce29637/csrc/multidevice/executor.cpp#L161) but @wujingyue told me it was faulty there (I don't remember the reason btw...?)

What I can also do is disabling the segmentation when `auto_schedul=false`, if that helps.

### 3. Other questions/comments

> Yes, it makes sense. In practice, what does the so-called "pre-scheduling" (for which I believe the code is outside this PR) do? 

For now, we mostly parallelize the outmost axis with `ParallelType::DiDx`. But later there will be other scheduling of course.

> IIUC, for the short/medium term, pre-scheduling is done by the user of nvFuser and it only annotates the device IDs. After that, the "high-level" segmentation kicks in and divide the fusion at ops that require re-sharding. If that's the case, does pre-scheduling need to be done by `FusionExecutorCache(auto_schedule=true)` at all?

Yes, we might wanna rely on auto-scheduling for the inter-device scheduling (aka composing manual and auto scheduling).
This is done in https://github.com/NVIDIA/Fuser/pull/1853

> PS: I think, when we later consolidate the two executors, we won't need multi-level segmentation and therefore won't need two `FusionExecutorCache`s, which can simplify a lot. Alas, I doubt I'll have time to implement that before GTC...

I agree that it will simplify. I'll soon start melting `MultiDeviceExecutor` into `FusionExecutorCache`. However, I don't think it will solve any of the other issues mentioned here, and the present patch (i.e. allowing  manual scheduling) is still needed.

csrc/kernel_cache.cpp
samnordmann: > I'm a bit surprised that the segmenter actually works with a pre-scheduled fusion.

The segmenter works because the error we might encounter with a pre-scheduled fusion happens down the line when we want to actually apply the scheduling.

So, in this patch, we "dupe" the segmenter: it thinks each segment will be scheduled wit ha given heuristic but we actually bypass the application of that heuristic

csrc/multidevice/executor.cpp
samnordmann: Now, there is no check that the fusion is not pre-scheduled, it is only implicitely assumed. The scheduler doesn't try to preserve pre-scheduling, it will just blindly try to apply scheduling, and might stumble by mistake on an assert error such as [this one](https://github.com/NVIDIA/Fuser/blob/71d9b555b401655c0d664b94e8abb7021ce29637/csrc/tensor_view.cpp#L665).

csrc/kernel_cache.cpp
samnordmann: done!

csrc/multidevice/executor.cpp
samnordmann: I agree with you on principle, and this is what I tried initially, but the compiler complains:

```
FAILED: CMakeFiles/codegen_internal.dir/csrc/multidevice/executor.cpp.o
/usr/bin/c++ -DNVFUSER_DISTRIBUTED -DTORCH_CUDA_BUILD_MAIN_LIB -DUSE_C10D_GLOO -DUSE_C10D_MPI -DUSE_C10D_NCCL -DUSE_C10D_UCC -DUSE_DISTRIBUTED -DUSE_RPC -DUSE_TENSORPIPE -I/opt/pytorch/Fuser/csrc -I/home/snordmann/nvfuser_build/include -I/opt/pytorch/Fuser/lib/dynamic_type/src -isystem /opt/pytorch/Fuser/third_party/gloo -isystem /opt/pytorch/Fuser/third_party/flatbuffers/include -isystem /usr/local/cuda/extras/CUPTI/include -isystem /usr/local/cuda/include -isystem /opt/p
ytorch/pytorch/torch/include -isystem /opt/pytorch/pytorch/torch/include/torch/csrc/api/include -Wno-psabi -D_GLIBCXX_USE_CXX11_ABI=1 -g -std=gnu++17 -fPIC -fvisibility=hidden -fvisibility-inlines-hidden -Wall -Wno-unused-function -Werror -D_GLIBCXX_USE_CXX11_ABI=1 -MD -MT CMakeFiles/codegen_internal.dir/csrc/multidevice/executor.cpp.o -MF CMakeFiles/codegen_internal.dir/csrc/multidevice/executor.cpp.o.d -o CMakeFiles/codegen_internal.dir/csrc/multidevice/executor.cpp.o -c
 /opt/pytorch/Fuser/csrc/multidevice/executor.cpp
In file included from /usr/include/x86_64-linux-gnu/c++/11/bits/c++allocator.h:33,
                 from /usr/include/c++/11/bits/allocator.h:46,
                 from /usr/include/c++/11/unordered_map:40,
                 from /usr/include/c++/11/functional:61,                                                                                                                                                                                                                                                                                                                                                                                                                                                      from /opt/pytorch/pytorch/torch/include/c10/core/Allocator.h:5,
                 from /opt/pytorch/pytorch/torch/include/ATen/cuda/CUDAContextLight.h:24,
                 from /opt/pytorch/pytorch/torch/include/ATen/cuda/CUDAContext.h:3,
                 from /opt/pytorch/Fuser/csrc/multidevice/executor.cpp:9:
/usr/include/c++/11/ext/new_allocator.h: In instantiation of ‘void __gnu_cxx::new_allocator<_Tp>::construct(_Up*, _Args&& ...) [with _Up = std::pair<nvfuser::SegmentedGroup* const, nvfuser::FusionExecutorCache>; _Args = {nvfuser::SegmentedGroup*&, nvfuser::FusionExecutorCache}; _Tp = std::__detail::_Hash_node<std::pair<nvfuser::SegmentedGroup* const, nvfuser::FusionExecutorCache>, false>]’:
/usr/include/c++/11/bits/alloc_traits.h:516:17:   required from ‘static void std::allocator_traits<std::allocator<_Tp1> >::construct(std::allocator_traits<std::allocator<_Tp1> >::allocator_type&, _Up*, _Args&& ...) [with _Up = std::pair<nvfuser::SegmentedGroup* const, nvfuser::FusionExecutorCache>; _Args = {nvfuser::SegmentedGroup*&, nvfuser::FusionExecutorCache}; _Tp = std::__detail::_Hash_node<std::pair<nvfuser::SegmentedGroup* const, nvfuser::FusionExecutorCache>, false
>; std::allocator_traits<std::allocator<_Tp1> >::allocator_type = std::allocator<std::__detail::_Hash_node<std::pair<nvfuser::SegmentedGroup* const, nvfuser::FusionExecutorCache>, false> >]’
/usr/include/c++/11/bits/hashtable_policy.h:1878:36:   required from ‘std::__detail::_Hashtable_alloc<_NodeAlloc>::__node_type* std::__detail::_Hashtable_alloc<_NodeAlloc>::_M_allocate_node(_Args&& ...) [with _Args = {nvfuser::SegmentedGroup*&, nvfuser::FusionExecutorCache}; _NodeAlloc = std::allocator<std::__detail::_Hash_node<std::pair<nvfuser::SegmentedGroup* const, nvfuser::FusionExecutorCache>, false> >; std::__detail::_Hashtable_alloc<_NodeAlloc>::__node_ptr = std::allocator<std::__detail::_Hash_node<std::pair<nvfuser::SegmentedGroup* const, nvfuser::FusionExecutorCache>, false> >::value_type*]’
/usr/include/c++/11/bits/hashtable.h:304:35:   required from ‘std::_Hashtable<_Key, _Value, _Alloc, _ExtractKey, _Equal, _Hash, _RangeHash, _Unused, _RehashPolicy, _Traits>::_Scoped_node::_Scoped_node(std::_Hashtable<_Key, _Value, _Alloc, _ExtractKey, _Equal, _Hash, _RangeHash, _Unused, _RehashPolicy, _Traits>::__hashtable_alloc*, _Args&& ...) [with _Args = {nvfuser::SegmentedGroup*&, nvfuser::FusionExecutorCache}; _Key = nvfuser::SegmentedGroup*; _Value = std::pair<nvfuse
r::SegmentedGroup* const, nvfuser::FusionExecutorCache>; _Alloc = std::allocator<std::pair<nvfuser::SegmentedGroup* const, nvfuser::FusionExecutorCache> >; _ExtractKey = std::__detail::_Select1st; _Equal = std::equal_to<nvfuser::SegmentedGroup*>; _Hash = std::hash<nvfuser::SegmentedGroup*>; _RangeHash = std::__detail::_Mod_range_hashing; _Unused = std::__detail::_Default_ranged_hash; _RehashPolicy = std::__detail::_Prime_rehash_policy; _Traits = std::__detail::_Hashtable_t
raits<false, false, true>; std::_Hashtable<_Key, _Value, _Alloc, _ExtractKey, _Equal, _Hash, _RangeHash, _Unused, _RehashPolicy, _Traits>::__hashtable_alloc = std::_Hashtable<nvfuser::SegmentedGroup*, std::pair<nvfuser::SegmentedGroup* const, nvfuser::FusionExecutorCache>, std::allocator<std::pair<nvfuser::SegmentedGroup* const, nvfuser::FusionExecutorCache> >, std::__detail::_Select1st, std::equal_to<nvfuser::SegmentedGroup*>, std::hash<nvfuser::SegmentedGroup*>, std::__d
etail::_Mod_range_hashing, std::__detail::_Default_ranged_hash, std::__detail::_Prime_rehash_policy, std::__detail::_Hashtable_traits<false, false, true> >::__hashtable_alloc]’
/usr/include/c++/11/bits/hashtable.h:1966:15:   required from ‘std::pair<typename std::__detail::_Insert<_Key, _Value, _Alloc, _ExtractKey, _Equal, _Hash, _RangeHash, _Unused, _RehashPolicy, _Traits>::iterator, bool> std::_Hashtable<_Key, _Value, _Alloc, _ExtractKey, _Equal, _Hash, _RangeHash, _Unused, _RehashPolicy, _Traits>::_M_emplace(std::true_type, _Args&& ...) [with _Args = {nvfuser::SegmentedGroup*&, nvfuser::FusionExecutorCache}; _Key = nvfuser::SegmentedGroup*; _V
alue = std::pair<nvfuser::SegmentedGroup* const, nvfuser::FusionExecutorCache>; _Alloc = std::allocator<std::pair<nvfuser::SegmentedGroup* const, nvfuser::FusionExecutorCache> >; _ExtractKey = std::__detail::_Select1st; _Equal = std::equal_to<nvfuser::SegmentedGroup*>; _Hash = std::hash<nvfuser::SegmentedGroup*>; _RangeHash = std::__detail::_Mod_range_hashing; _Unused = std::__detail::_Default_ranged_hash; _RehashPolicy = std::__detail::_Prime_rehash_policy; _Traits = std:
:__detail::_Hashtable_traits<false, false, true>; typename std::__detail::_Insert<_Key, _Value, _Alloc, _ExtractKey, _Equal, _Hash, _RangeHash, _Unused, _RehashPolicy, _Traits>::iterator = std::__detail::_Insert_base<nvfuser::SegmentedGroup*, std::pair<nvfuser::SegmentedGroup* const, nvfuser::FusionExecutorCache>, std::allocator<std::pair<nvfuser::SegmentedGroup* const, nvfuser::FusionExecutorCache> >, std::__detail::_Select1st, std::equal_to<nvfuser::SegmentedGroup*>, std::hash<nvfuser::SegmentedGroup*>, std::__detail::_Mod_range_hashing, std::__detail::_Default_ranged_hash, std::__detail::_Prime_rehash_policy, std::__detail::_Hashtable_traits<false, false, true> >::iterator; typename _Traits::__constant_iterators = std::__detail::_Hashtable_traits<false, false, true>::__constant_iterators; std::true_type = std::integral_constant<bool, true>]’
/usr/include/c++/11/bits/hashtable.h:915:21:   required from ‘std::_Hashtable<_Key, _Value, _Alloc, _ExtractKey, _Equal, _Hash, _RangeHash, _Unused, _RehashPolicy, _Traits>::__ireturn_type std::_Hashtable<_Key, _Value, _Alloc, _ExtractKey, _Equal, _Hash, _RangeHash, _Unused, _RehashPolicy, _Traits>::emplace(_Args&& ...) [with _Args = {nvfuser::SegmentedGroup*&, nvfuser::FusionExecutorCache}; _Key = nvfuser::SegmentedGroup*; _Value = std::pair<nvfuser::SegmentedGroup* const
, nvfuser::FusionExecutorCache>; _Alloc = std::allocator<std::pair<nvfuser::SegmentedGroup* const, nvfuser::FusionExecutorCache> >; _ExtractKey = std::__detail::_Select1st; _Equal = std::equal_to<nvfuser::SegmentedGroup*>; _Hash = std::hash<nvfuser::SegmentedGroup*>; _RangeHash = std::__detail::_Mod_range_hashing; _Unused = std::__detail::_Default_ranged_hash; _RehashPolicy = std::__detail::_Prime_rehash_policy; _Traits = std::__detail::_Hashtable_traits<false, false, true
>; std::_Hashtable<_Key, _Value, _Alloc, _ExtractKey, _Equal, _Hash, _RangeHash, _Unused, _RehashPolicy, _Traits>::__ireturn_type = std::_Hashtable<nvfuser::SegmentedGroup*, std::pair<nvfuser::SegmentedGroup* const, nvfuser::FusionExecutorCache>, std::allocator<std::pair<nvfuser::SegmentedGroup* const, nvfuser::FusionExecutorCache> >, std::__detail::_Select1st, std::equal_to<nvfuser::SegmentedGroup*>, std::hash<nvfuser::SegmentedGroup*>, std::__detail::_Mod_range_hashing,
std::__detail::_Default_ranged_hash, std::__detail::_Prime_rehash_policy, std::__detail::_Hashtable_traits<false, false, true> >::__ireturn_type]’
/usr/include/c++/11/bits/unordered_map.h:389:23:   required from ‘std::pair<typename std::_Hashtable<_Key, std::pair<const _Key, _Tp>, _Alloc, std::__detail::_Select1st, _Pred, _Hash, std::__detail::_Mod_range_hashing, std::__detail::_Default_ranged_hash, std::__detail::_Prime_rehash_policy, std::__detail::_Hashtable_traits<std::__not_<std::__and_<std::__is_fast_hash<_Hash>, std::__is_nothrow_invocable<const _Hash&, const _Key&> > >::value, false, true> >::iterator, bool>
std::unordered_map<_Key, _Tp, _Hash, _Pred, _Alloc>::emplace(_Args&& ...) [with _Args = {nvfuser::SegmentedGroup*&, nvfuser::FusionExecutorCache}; _Key = nvfuser::SegmentedGroup*; _Tp = nvfuser::FusionExecutorCache; _Hash = std::hash<nvfuser::SegmentedGroup*>; _Pred = std::equal_to<nvfuser::SegmentedGroup*>; _Alloc = std::allocator<std::pair<nvfuser::SegmentedGroup* const, nvfuser::FusionExecutorCache> >; typename std::_Hashtable<_Key, std::pair<const _Key, _Tp>, _Alloc, s
td::__detail::_Select1st, _Pred, _Hash, std::__detail::_Mod_range_hashing, std::__detail::_Default_ranged_hash, std::__detail::_Prime_rehash_policy, std::__detail::_Hashtable_traits<std::__not_<std::__and_<std::__is_fast_hash<_Hash>, std::__is_nothrow_invocable<const _Hash&, const _Key&> > >::value, false, true> >::iterator = std::__detail::_Insert_base<nvfuser::SegmentedGroup*, std::pair<nvfuser::SegmentedGroup* const, nvfuser::FusionExecutorCache>, std::allocator<std::pa
ir<nvfuser::SegmentedGroup* const, nvfuser::FusionExecutorCache> >, std::__detail::_Select1st, std::equal_to<nvfuser::SegmentedGroup*>, std::hash<nvfuser::SegmentedGroup*>, std::__detail::_Mod_range_hashing, std::__detail::_Default_ranged_hash, std::__detail::_Prime_rehash_policy, std::__detail::_Hashtable_traits<false, false, true> >::iterator]’
/opt/pytorch/Fuser/csrc/multidevice/executor.cpp:159:17:   required from here
/usr/include/c++/11/ext/new_allocator.h:162:11: error: no matching function for call to ‘std::pair<nvfuser::SegmentedGroup* const, nvfuser::FusionExecutorCache>::pair(nvfuser::SegmentedGroup*&, nvfuser::FusionExecutorCache)’
  162 |         { ::new((void *)__p) _Up(std::forward<_Args>(__args)...); }
      |           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

wujingyue:
> If not using FusionExecutorCache, then, what is the good way to execute a pre-scheduled function ? 

FusionExecutor. Plenty of our unit tests manually schedule a Fusion and run it via FusionExecutor. 

> but @wujingyue told me it was faulty there (I don't remember the reason btw...?)

FusionExecutor doesn't segment, so a region between two sharding ops would fail to run when it can't run as one segment.

naoyam:
We'll talk more in an offline meeting soon, but the important assumptions behind the design of `FusionExecutorCache` are:

- A user has a program to pass to nvFuser. The user first defines the program using the Fusion IR with no scheduling (or, the user defines a program in Python and the Python interface converts the Python program to a Fusion IR program)
- The Fusion IR program is given to `FusionExecutorCache`, which manages the execution of the program for any inputs but only for the given particular program
- Whenever a user has a different program, a new `FusionExecutorCache` needs to be used. This is a responsibility of the user not the cache itself.

> I think that this problem is intrinsic to the current cache mechanism, and not related to this PR per say. 

I think now you'll understand that this is not a problem but by design.

samnordmann:
> FusionExecutor. Plenty of our unit tests manually schedule a Fusion and run it via FusionExecutor.
> FusionExecutor doesn't segment, so a region between two sharding ops would fail to run when it can't run as one segment.

Thanks! But I am not sure to understand, is it safe to use `FusionExecutor` or can it fail? In what case it would typically fail? I can't figure out any example


> I think now you'll understand that this is not a problem but by design.

I am not sure to understand. Let's consider the example of the same Fusion definition but changing the scheduling. If the scheduling is not allowed to change, then why do we need to cache it? And why would it be inconsistent in this PR that the cached key doesn't correspond to the actual schedule?

csrc/multidevice/executor.cpp
wujingyue: I believe you missed the `try_` part. Anyhow, I tried the following on **your** branch and it builds. 

```
diff --git a/csrc/multidevice/executor.cpp b/csrc/multidevice/executor.cpp
index 52120d2..e3e0226 100644
--- a/csrc/multidevice/executor.cpp
+++ b/csrc/multidevice/executor.cpp
@@ -155,13 +155,8 @@ void MultiDeviceExecutor::postKernel(SegmentedGroup* group) {

   // Compile the group and execute it with FusionExecutor
   // Check if the executor has been cached. If not, create and cache it
-  if (fec_.find(group) == fec_.end()) {
-    fec_.emplace(
-        group,
-        std::make_unique<FusionExecutorCache>(
-            staged_fusion_->makeFusion(group), 0, auto_schedule_));
-  }
-  outputs = fec_[group]->runFusionWithInputs(group_input_IValues);
+  fec_.try_emplace(group, staged_fusion_->makeFusion(group), 0, auto_schedule_);
+  outputs = fec_.at(group).runFusionWithInputs(group_input_IValues);

   // Store the outputs in the context
   for (auto output_idx : c10::irange(outputs.size())) {
diff --git a/csrc/multidevice/executor.h b/csrc/multidevice/executor.h
index 49fe0b7..fe25418 100644
--- a/csrc/multidevice/executor.h
+++ b/csrc/multidevice/executor.h
@@ -119,8 +119,7 @@ class MultiDeviceExecutor {
   // Stores the order in which the pipeline's stage should be executed
   RuntimeWorkSpace workspace;
   // Cache FusionsExecutorCache
-  std::unordered_map<SegmentedGroup*, std::unique_ptr<FusionExecutorCache>>
-      fec_;
+  std::unordered_map<SegmentedGroup*, FusionExecutorCache> fec_;
   // Cache whether a SegmentedGroup should be run by the current device
   std::unordered_map<SegmentedGroup*, bool> should_run_;
   // Cache whether a SegmentedGroup requires inter-device communication
```

wujingyue: (I would have committed this directly if I had been added to your repo!)

wujingyue:
> Thanks! But I am not sure to understand, is it safe to use `FusionExecutor` or can it fail? In what case it would typically fail? I can't figure out any example

It can fail and crash the program, e.g., https://github.com/NVIDIA/Fuser/issues/1553. It typically manifests as a "nvFuser can't run this fusion unsegmented" error message.

naoyam (approved):
Stamping as we discussed today. Please make sure to have a clear warning that the use of the flag breaks the assumption of FusionExecutorCache.

samnordmann:
Quick summary of the discussion:

1) skipping auto-scheduling in FusionExecutorCache "abuses" the design and should be done with extreme care. It is a temporary hack that should only be used for development. Warning have been expressed about using FusionExecutorCache this way.
2) This hack is useful to gradually use FusionExecutorCache with our pre-scheduled Fusion, and also to start merging the MultiDeviceExecutor with FusionExecutorCache.
3) Alternatively, we would need to stick to FusionExecutor but to rewrite a lot of the logic of FusionExecutorCache inside MultiDEviceExecutor
4) Ideally, FusionExecutorCache should be refactored to enable a "clean" disabling of auto-scheduling, or, the schedulers should be composable between them or with user's constraint.

In an attempt to get "best of both world", in this pr, we keep both options by introducing a param struct, see multidevice/executor.h:
```
struct MultiDeviceExecutorParams {
  // Experimental: whether to use FusionExecutorCache rather than
  // FusionExecutor.
  bool use_fusion_executor_cache = false;
  // Experimental: whether to apply auto-scheduling in FusionExecutorCache if
  // use_fusion_executor_cache=true. WAR: temporary hack mainly use for
  // development
  bool skip_auto_scheduling = false;
  // Experimental: whether to cache fusion executor. WAR: avoid recompilation
  // but implicitely assumes that the input shape don't change over iterations
  bool cache_fusion_executor = false;
};
```
We'll try to maintain both executor modes as much as possible, or switch from one to the other depending on the case. For now, both work as is with our current test suites.

PR #1845, opened 2024-02-27
Fix an embarrassing bug.

wujingyue:
Previously, the NVF_ERROR did nothing but fallthrough.


wujingyue:
!build

zasdfgbnm (approved):

jjsjann123 (approved):
:laughing:

jacobhinkle:
I find myself almost doing this sometimes. I think it's because the name `NVF_ERROR` makes it sound like it will always error, as opposed to something like `NVF_ASSERT`.

PR #1846, opened 2024-02-28
Add utility replayExprWithNewInput for #1768.

wujingyue:


test/test_replay.cpp
```diff
@@ -0,0 +1,57 @@
...
+  TensorView* r1 = reshape(s1, {4, 3}, {2, 2, 3});
+  TensorView* out = cat({r0, r1}, /*dim=*/-1);
+
+  std::vector<IterDomain*> new_out_root = IterDomain::clone(
+      TensorDomain::noReductions(in->getMaybeRFactorDomain()));
+  TensorView* new_out = IrBuilder::create<TensorView>(
+      IrBuilder::create<TensorDomain>(new_out_root), *out->getDataType());
+  new_out->setDomain(
+      TransformReplay::fullSelfReplay(new_out->domain(), r0->domain()));
+  // FIXME: how about allocation domain and contiguity?
```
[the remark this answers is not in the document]
wujingyue: I think that'll work. I'm fixing the `r0->getContiguity()` part -- should be easy. Code as is fails at https://github.com/samnordmann/Fuser/blob/28c36b6fe69dc67486cc08628a25269ce855aff2/csrc/ir/nodes.cpp#L3191, because before replay the maybeAllocationDomain is 2D not 3D. 

That being said, how would you go around the validation that I disabled? **Maybe** TransformReplay::fullSelfReplay is the wrong API and I should somehow publicize ReplayTransforms and use that instead?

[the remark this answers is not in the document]
wujingyue: To be clear, the validation in fullSelfReplay fails at https://github.com/NVIDIA/Fuser/blob/71d9b555b401655c0d664b94e8abb7021ce29637/csrc/transform_replay.cpp#L205-L206 because r0's root domain has a rfactor product and IterDomain::clone removes rfactor product.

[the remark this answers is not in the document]
wujingyue: I could also weaken that check 🤷

csrc/transform_replay.cpp
```diff
@@ -200,6 +200,7 @@ TensorDomain* TransformReplay::fullSelfReplay(
   {
     size_t i = 0;
     for (auto id : self->root()) {
+#if 0
```
naoyam, line 203: I think what fails here is because some of the domains in `self` is marked as an rfactor product, but not in the `new_self_root`. IIRC, root domains transformed by a reshape op are marked as rfactor product.

csrc/ir/interface_nodes.h
```diff
@@ -545,7 +545,6 @@ class NVF_API TensorView : public Val {
     return !mesh_.vector().empty();
   }
 
- protected:
```
naoyam, line 548: Is this for `setDomain`? Is it used?

csrc/transform_replay.cpp
wujingyue: That's exactly right! What's the right fix?

csrc/ir/interface_nodes.h
wujingyue: Not any more after @jacobhinkle's change.

test/test_replay.cpp
```diff
@@ -0,0 +1,64 @@
...
+  TensorDomain* r_domain = TransformReplay::fullSelfReplay(
+      IrBuilder::create<TensorDomain>(r_root), r0->domain());
+  TensorView* r = IrBuilder::create<TensorView>(r_domain, *out->getDataType());
+
+  Expr* to_replay = r0->definition();
+  auto create_fn = to_replay->newObjectFunc();
+  create_fn(to_replay->container(), {in}, {r}, to_replay->attributes());
+  // To preserve the output allocation domain, we create a Set between `r` and
+  // `out` instead of replacing the fusion output.
+  IrBuilder::create<LoadStoreOp>(LoadStoreOpType::Set, out, r);
```
jacobhinkle, line 47: Does `out->cacheBefore()` work in this case?

jacobhinkle: Your way is pretty much equivalent though, and doesn't create a tensor `c` that just gets thrown away, so you should probably just keep it as is.

csrc/transform_replay.cpp
naoyam: A bit more manual, but using `IterDomainBuilder` directly should work. You just don't want to call `resetRFactor()`.

naoyam: To be clear, I'm referring to the code about cloning iter domains.

wujingyue: The source of the cloning, `in->getMaybeRFactorDomain()`, has no rfactor products. So changing cloning doesn't help. It's `r0->domain()`, the domain to replay, that has rfactor products.

naoyam: Yeah, but we know which domain should be rfactor products, so we could do `IterDomainBuilder::is_rfactor_domain(true)` for those domains.

wujingyue: Yes, we could, but is it safe? 

In this particular case, the slice's output TV (and thus the original reshape's input) has an `rf` because `slice` explicitly adds it: https://github.com/NVIDIA/Fuser/blob/71d9b555b401655c0d664b94e8abb7021ce29637/csrc/ops/alias.cpp#L752. When we create the new reshape on the non-sliced input, the root domain of the new reshape shouldn't have the `rf` because those dimensions are not sliced in the first place.

(A bigger question from me is what the `rf` flag means...)

wujingyue: Move the bigger, `rf` flag question to https://github.com/NVIDIA/Fuser/issues/1857.

naoyam (approved):
LGTM

wujingyue:
!build

Issue #1819, opened 2024-02-22
fp8 plumbing


jjsjann123:
Just want a place to track my fp8 plumbing work.

I'll likely just follow our fp16 story and plumb similar runtime support for both fp8 formats.

From ptx doc (https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#fundamental-types).
```
e4m3 data format:
This data format is an 8-bit floating point format with 4 bits for exponent and 3 bits for mantissa. The e4m3 encoding does not support infinity and NaN values are limited to 0x7f and 0xff. A register variable containing e4m3 value must be declared using bit-size type.

e5m2 data format:
This data format is an 8-bit floating point format with 5 bits for exponent and 2 bits for mantissa. A register variable containing e5m2 value must be declared using bit-size type.
```

PyTorch fp8 format contains this. I haven't got much details on what those formats mean, will ask upstream regarding this.
```
torch.float8_e4m3fn    torch.float8_e4m3fnuz  torch.float8_e5m2      torch.float8_e5m2fnuz
```

PR #1847, opened 2024-02-28
Adding fp8 dtypes

jjsjann123:
Fixes: #1819 

Two fp8 dtype `Float8_e4m3fn` and `Float8_e5m2` added to runtime files
* These two are what [cuda supported](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#alternate-floating-point-data-formats)

Supports casting between [e4m3/e5m2] <-> [double/float/half/bfloat32].

Note:
for runtime file we are currently going without marking the suffix `__e4m3` & `__e5m2`, since those are already cuda default.

TODO:
- [x] adding test; [in this PR]
- [x] adding more casting rules; [Can be done in future PRs]

FUTURE PR:
benchmark to verify performance. I'm not sure if my inline PTX is done appropriately. When I look at some compiled PTX from fp8 math APIs, I'm seeing a lot of bit shifting, which looks too explicit to my untrained eyes and I am not sure if it matter. Since this PR is already growing big, I'm leaning towards leaving that for a follow up PR.


csrc/type.cpp
```diff
@@ -111,12 +111,14 @@ bool isInclusiveType(const DataType& base_type, const DataType& wider_type) {
...
-       base_type == DataType::Half || base_type == DataType::BFloat16)) {
+       base_type == DataType::Half || base_type == DataType::BFloat16 ||
+       base_type == DataType::Float8_e4m3fn || base_type == DataType::Float8_e5m2)) {
     return true;
   }
   if ((wider_type == DataType::Float || wider_type == DataType::ComplexFloat) &&
       (base_type == DataType::Float || base_type == DataType::Half ||
-       base_type == DataType::BFloat16)) {
+       base_type == DataType::BFloat16 ||
+       base_type == DataType::Float8_e4m3fn || base_type == DataType::Float8_e5m2)) {
```
naoyam, line 121: Do we also need a case where the wider type is 16-bit and the base type is 8-bit?

runtime/fp8_support.cu
```diff
@@ -0,0 +1,99 @@
+// clang-format off
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2023-present NVIDIA CORPORATION & AFFILIATES.
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+// clang-format on
+
+#define __NVFUSER_HALF_TO_US(var) *(reinterpret_cast<unsigned short*>(&(var)))
```
naoyam, line 9: We already have this in the fp16 file. Doesn't this result in a name collision? Don't we need to undef it?

csrc/validator_utils.h
```diff
@@ -136,6 +136,8 @@ std::pair<double, double> getTolerance(
         return {abs_tol, abs_tol * 0.01};
       }
     }
+    case DataType::Float8_e4m3fn:
+    case DataType::Float8_e5m2:
```
naoyam, line 140: @Priya2698 extended the validator to better support fp16. We may need to do something for these new types. I'm not sure if the PR is already merged (is it, @Priya2698?)

Nothing needs to be done for now, but maybe leave a TODO comment here?

naoyam (commented):
Looks good so far. Just left a couple of minor comments.

csrc/validator_utils.h
Priya2698: Are you referring to the regenerated tolerances? (https://github.com/NVIDIA/Fuser/pull/1593). We did not merge this since it wasn't resolving the CI validation errors.

jjsjann123:
!build

jjsjann123:
!build

jjsjann123:
!build

jjsjann123:
> Can you please also add simple C++ tests as well? It should be also possible to generate a Kernel IR and make sure it has a proper cast op even on non-Hopper machines.

Let me clarify this.

So we are just trying to look at the kernel IR and check that it will generate the correct cast op, even though we cannot run the kernel since the hardware doesn't have support.

naoyam:
> > Can you please also add simple C++ tests as well? It should be also possible to generate a Kernel IR and make sure it has a proper cast op even on non-Hopper machines.
> 
> Let me clarify this.
> 
> So we are just trying to look at the kernel IR and check that it will generate the correct cast op, even though we cannot run the kernel since the hardware doesn't have support.

Yes, not a big deal, but this should be trivial now given the recent PR on the arch version by @jacobhinkle (#1784), so why not? 😄

jjsjann123:
!build

test/test_gpu1.cpp
```diff
@@ -2737,44 +2737,64 @@ TEST_F(NVFuserTest, FusionCompoundOps_CUDA) {
...
+      {DataType::Float8_e4m3fn, DataType::Float8_e5m2});
+  std::vector<DataType> cast_targets(
+      {DataType::Double, DataType::Float, DataType::BFloat16, DataType::Half});
+
+  for (const auto& fp8_type : fp8_variants) {
+    for (const auto& src_type : cast_targets) {
+      Fusion fusion;
+      FusionGuard fg(&fusion);
 
-  TensorView* tv0 = makeSymbolicTensor(2, DataType::Half);
```
jjsjann123, line 2744: The old test does a `half->float->half`, which would be optimized away already. So I just removed the test.

test/test_gpu1.cpp
```diff
@@ -2737,44 +2737,64 @@ TEST_F(NVFuserTest, FusionCompoundOps_CUDA) {
...
 
-  NVF_CHECK(
-      outputs[0].equal(ref_output),
-      "\nOp Type: -- ",
-      "cast FP16->FP32->FP16",
-      " -- had a mismatch.\n",
-      "\nABS MAX DIFF: ",
-      outputs[0].sub(ref_output).abs().max(),
-      "\n");
+      if (!deviceMajorMinorCheck(9)) {
```
jjsjann123, line 2777: @naoyam does this look good enough?

naoyam (approved):
LGTM

PR #1849, opened 2024-02-28
select persistent batch size based on occupancy, register usage, block size, and divisible split

liqiangxl:
**For details see the [[doc](https://docs.google.com/document/d/1oKG3Tb_abbNnqV7EUMU5GveUUwjK6DvON0jrudKTPp8/edit?usp=sharing)]**.
**Issue:** Current heuristic derive the persistent batch size from `bdimx = 128`, if it is larger than max allowed value, bdimx will be doubled until the new persistent batch size is lower than max allowed. This leads to low occupancy and performance at many hidden sizes.

**Fix:** Generate a heuristic for all the allowed persistent batch sizes and select the best one considering occupancy, register usage, block size and divisible split. More specifically:
```
// (1) If the maximum occupancy is less than the target occupancy, use the batch
//     leads to the largest occupancy.
// (2) Otherwise, sort the heuristics to prioritize occupancy.
//     (a) Prefer occupancy larger than target.
//     (b) Prefer divisible by persistent batch size.
//     (c) Prefer large register overhead.
//     (d) Prefer large occupancy.
//     (e) Tiebreaker, use large persistent batch size.
// (3) For softmax, a special rule is used to reduce inter-thread communication
//     (a) Prefer bdimx_val equals power of 2
//     (b) Prefer occupancy larger than target.
//     (c) Tiebreaker, use small bdimx
```
**Results:**
image
image


liqiangxl:
!build

liqiangxl:
!build

csrc/scheduler/normalization_inner.cpp
```diff
@@ -252,6 +263,384 @@ int64_t getMaxPersistentBatch(
...
+    score = compare(ha.n_persistent_tails == 0, hb.n_persistent_tails == 0);
+    if (score != 0) {
+      return score > 0;
+    }
+
+    // Large register overhead leaves more registers for the compiler to
+    // optimize generated SASS code and avoids register spills. But don't want
+    // to use a very large block size which increaes inter-thread communication.
+    // This condition influence dropout_layer_norm at 7K and softmax_dropout at
+    // 6K and 16K.
```
liqiangxl, line 396: @jjsjann123

csrc/scheduler/normalization_inner.cpp
```diff
@@ -252,6 +263,384 @@ int64_t getMaxPersistentBatch(
...
+    score = compare(ha.occupancy, hb.occupancy);
+    if (score != 0) {
+      return score > 0;
+    }
+
+    // Tiebreaker, use large persistent batch size so more registers are used
+    // for the persistent buffer.
+    return ha.persistent_batch_size > hb.persistent_batch_size;
+
+  } else {
```
liqiangxl, line 417: special for softmax, @jjsjann123

jjsjann123:
One quick question regarding benchmark. I haven't seen any A100 numbers... is that something we should check as well?

liqiangxl:
> One quick question regarding benchmark. I haven't seen any A100 numbers... is that something we should check as well?

yes, we should. I'll add.

csrc/scheduler/normalization_inner.cpp
```diff
@@ -203,19 +204,29 @@ int64_t getMaxRegisterCountPerThread(
...
-      blocks_per_sm_max * threads_per_block);
+  int64_t blocks_per_sm_max = scheduler_utils::safeDiv(
+      getThreadsPerSMGivenRegPerThread(register_per_thread_min),
+      threads_per_block);
+  threads_per_sm =
+      std::min(blocks_per_sm_max * threads_per_block, max_threads_per_sm);
+
+  return {
+      getRegPerThreadGivenThreadsPerSM(threads_per_sm),
+      blocks_per_sm_max * threads_per_block / threads_per_warp};
```
jjsjann123, line 229: we shouldn't be using `blocks_per_sm_max` here any more. If it's capped by `max_threads_per_sm` earlier, our occupancy is also lowered by that as well.

Shouldn't the occupancy just be `threads_per_sm/threads_per_warp`

csrc/scheduler/normalization_inner.cpp
```diff
@@ -252,6 +263,384 @@ int64_t getMaxPersistentBatch(
...
+    const bool has_exp_op,
+    const bool has_rng_op) {
+  // Define two free parameters used in this heuristic.
+  // register_overhead is all registers except those for the persistent
+  // buffers. The register in each thread = register_overhead +
+  // persistent_buffer_size / bytes_per_register
+  // Current values are based on tests of sofmax, layer_norm, softmax_dropout,
+  // dropout_layer_norm on A100 & H100. It directly affects maxregcount passed
+  // to NVRTC and influences the occupancy.
+  const int64_t register_overhead = has_exp_op ? 32l : 16l;
```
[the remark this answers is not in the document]
liqiangxl: Here it is also used to detect `softmax_dropout` or any fusion has exp op.

csrc/scheduler/normalization_inner.cpp
liqiangxl: nice catch!

csrc/scheduler/normalization_inner.cpp
```diff
@@ -252,6 +263,384 @@ int64_t getMaxPersistentBatch(
...
+
+  // Sort the heuristics and select the best one.
+  // If no persistent batch size can achieve the target occupancy, and
+  HeuristicParams best_heuristic;
+  if (current_max_occupancy < target_warps_per_sm) {
+    best_heuristic = all_heuristics.at(idx_max_occupancy);
+  } else {
+    // Prioritize occupancy for all fusions except softmax.
+    // Not sure why softmax has regression (about 10%) if prioritize occupancy.
+    // So condition [!has_exp_op] is added to filter out softmax.
```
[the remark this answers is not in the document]
liqiangxl: Yes, it is used to identify softmax. It also identifies any fusion with exp op and without rng_op, which are not tested, hopefully the perf diff is around 10%.

csrc/scheduler/normalization_inner.cpp
```diff
@@ -203,19 +204,29 @@ int64_t getMaxRegisterCountPerThread(
       register_overhead;
 
   // (1) use register calculated from target occupancy
-  int64_t register_per_thread_target = getRegPerThreadGivenThreadsPerSM(
-      target_blocks_per_sm * threads_per_block);
+  int64_t threads_per_sm =
+      std::min(threads_per_block * target_blocks_per_sm, max_threads_per_sm);
```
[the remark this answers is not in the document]
liqiangxl: I think `threads_per_sm = std::min(threads_per_block * target_blocks_per_sm, max_threads_per_sm);` guarantees `threads_per_sm` won't beyond the allowed resident threads per SM.

csrc/scheduler/normalization_inner.cpp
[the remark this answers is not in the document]
liqiangxl: updated comments

csrc/scheduler/normalization_inner.cpp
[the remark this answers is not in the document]
jjsjann123: sorry I was reading this wrong earlier and didn't realize here we are just rounding up the target_threads_per_sm.

Looking at it here, we are basically doing:
```
int64_t threads_per_sm = 
ceilDiv(target_warps_per_sm * threads_per_warp, threads_per_block) * threads_per_block;
```
And then followed by `threads_per_sm = std::min(threads_per_sm, max_threads_per_sm);`

The cap here with `max_threads_per_sm` doesn't look right. Should that be a round down to a multiple of `threads_per_block` instead?

csrc/scheduler/normalization_inner.cpp
```diff
@@ -252,6 +261,386 @@ int64_t getMaxPersistentBatch(
...
+      std::min(target_warps_per_sm * threads_per_warp, max_threads_per_sm);
+  const int64_t pbs_max_2 = getMaxPersistentBatch(
+      buffer_bytes_per_batch, target_threads_per_sm, register_overhead);
+  const int64_t batches_per_block_inner_reduction_max = std::max(
+      batches_per_block_inner_reduction_min, std::min(pbs_max_1, pbs_max_2));
+
+  // Compute maximum number of reductions we could do in the same kernel based
+  // on persistent buffer size. Bounded by the wave count for utilization of
+  // SMs.
+  const int64_t max_multi_reduction_factor = std::min(
```
jjsjann123, line 524: nitpick: not a super big fan of duplicating some of the logic here with the old code path.

Including: `register_overhead`, `targets_warps_per_sm`, computation of `buffer_bytes_per_batch`, `getMaxPersistentBatch`, `max_multi_reduction_factor`.

Do you think it make sense to unify those?

csrc/scheduler/normalization_inner.cpp
```diff
@@ -252,6 +261,386 @@ int64_t getMaxPersistentBatch(
...
+
+  // set the max persistent batch size to avoid low occupancy
+  // don't increase persistent batch size until we have enough threads
+  const int64_t pbs_max_1 =
+      ceilDiv(parallel_after_vectorize, min_threads_per_block);
+  // derived the maximum persistent batch size from the target occupancy
+  const int64_t buffer_bytes_per_batch = max_persistent_buffer_size /
+      total_reduction_numel * (int64_t)vectorize_factor;
+  const int64_t target_threads_per_sm =
+      std::min(target_warps_per_sm * threads_per_warp, max_threads_per_sm);
```
jjsjann123, line 515: sorry I think this is what I meant earlier.

i.e. if we have `target_warps_per_sm * threads_per_warp > max_threads_per_sm`, it should be an assert because we are setting an unrealistic `target_warps_per_sm`...

csrc/scheduler/normalization_inner.cpp
```diff
@@ -252,6 +261,386 @@ int64_t getMaxPersistentBatch(
...
+  }
+
+  // Sort the heuristics and select the best one.
+  // If no persistent batch size can achieve the target occupancy, and
+  HeuristicParams best_heuristic;
+  if (current_max_occupancy < target_warps_per_sm) {
+    best_heuristic = all_heuristics.at(idx_max_occupancy);
+  } else {
+    // Prioritize occupancy for all fusions except softmax.
+    // Not sure why softmax has regression (about 10%) if prioritize occupancy.
```
jjsjann123, line 562: I know you are already on top of this. Let's either:
1. figure out why before we merge the PR, or
2. open an issue and follow up on that in a later PR.

csrc/scheduler/normalization_inner.cpp
```diff
@@ -252,6 +261,386 @@ int64_t getMaxPersistentBatch(
...
+    // Large register overhead leaves more registers for the compiler to
+    // optimize generated SASS code and avoids register spills. But don't want
+    // to use a very large block size which increaes inter-thread communication.
+    // This condition influence dropout_layer_norm at 7K and softmax_dropout at
+    // 6K and 16K.
+    constexpr int64_t opt_max_threads_per_block = 512l;
+    score = compare(
+        ha.register_overhead > target_register_overhead &&
+            ha.padded_bdimx <= opt_max_threads_per_block,
+        hb.register_overhead > target_register_overhead &&
```
jjsjann123, line 399: `register_overhead` is it more like `allowed_register_overhead`? so the higher the better?

More importantly, line 397 & 399 should have the same compare operator.

csrc/scheduler/normalization_inner.cpp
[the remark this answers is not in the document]
liqiangxl: make sense! I moved up the check:
```
  // ensure higher than target by round up
  int64_t target_blocks_per_sm =
      ceilDiv(target_warps_per_sm * threads_per_warp, threads_per_block);
  // ensure lower than hardware limit by round down
  if(threads_per_block * target_blocks_per_sm > max_threads_per_sm){
    target_blocks_per_sm = max_threads_per_sm / threads_per_block;
  }
```

This change won't influence the perf because our current `target_warps_per_sm * threads_per_warp = 28 * 32 = 896`, the worst case is round up from 895 threads to 1790 threads which is smaller than `max_threads_per_sm`

csrc/scheduler/normalization_inner.cpp
```diff
@@ -348,6 +739,25 @@ std::shared_ptr<ReductionParams> innerPersistentHeuristic(
         index_type);
   }
 
+  // 2D reduction
+  if (total_reduction_numel == inner_most_dimension_numel) {
```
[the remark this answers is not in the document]
liqiangxl: the assert is redundant and is removed.

csrc/scheduler/normalization_inner.cpp
liqiangxl: Make sense.
The plan is to separate 2D inner persistent `(e.g. layerNorm, [Iter,Redu])` and 3D inner persistent `(e.g. batchNorm, [Redu, Iter, Redu])`, so these duplicated variables may change in the 3D inner persistent heuristic (not revised yet).

The solution I have is we can refactor this `namespace` into a class, and these vars are converted to priviate members so we don't need to duplicate or passing them around different functions. I'll make these changes in a single commit, if you feel the change is too big, we can drop it and redo in another PR.

csrc/scheduler/normalization_inner.cpp
liqiangxl: yes. `    // Large register overhead leaves more registers for the compiler to optimize generated SASS code and avoids register spills.`

line 397 `ha.register_overhead > target_register_overhead` & 399  `hb.register_overhead > target_register_overhead` have the same compare operator, right? both are `>`. 
Or you mean `ha.padded_bdimx <= opt_max_threads_per_block`? Here these two conditions are merged with `and`. we want to use high register overhead but don't want to achieve it at the cost of using a very large block size.

Maybe we should rename `register_overhead` to `spare_registers` or `free_registers` to more precisly refelecting their purpose. They are specifically set aside for transient or temporary usage rather than storing persistent buffers.

csrc/scheduler/normalization_inner.cpp
liqiangxl: made a [draft](https://github.com/NVIDIA/Fuser/pull/1873/files/452d5d9dc65556c4ab1823a3e43eaa7475c3db20..7d032b66cba034fdddec655efe96eb21d9f97f09) to solve this issue:

csrc/scheduler/normalization_inner.cpp
liqiangxl: @naoyam softmax is also causing trouble in this PR.

naoyam: What do you mean?

liqiangxl: Not sure why softmax has regression (about 10%) if prioritize occupancy. 
All other tested fusions show improvement using the following method to select the best persistent batch sizes.
```
// (1) If the maximum occupancy is less than the target occupancy, use the batch
//     leads to the largest occupancy.
// (2) Otherwise, sort the heuristics to prioritize occupancy.
//     (a) Prefer occupancy larger than target.
//     (b) Prefer divisible by persistent batch size.
//     (c) Prefer large register overhead.
//     (d) Prefer large occupancy.
//     (e) Tiebreaker, use large persistent batch size.
// (3) For softmax, a special rule is used to reduce inter-thread communication
//     (a) Prefer bdimx_val equals power of 2
//     (b) Prefer occupancy larger than target.
//     (c) Tiebreaker, use small bdimx
```

liqiangxl: some explanations: https://docs.google.com/document/d/1oKG3Tb_abbNnqV7EUMU5GveUUwjK6DvON0jrudKTPp8/edit#heading=h.rp1ug14ktzrc

csrc/scheduler/normalization_inner.cpp
jjsjann123: :pushpin: ^^^ This change hasn't been applied yet.

csrc/scheduler/normalization_inner.cpp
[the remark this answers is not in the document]
jjsjann123: I prefer to have it here since 8.6~8.9 has an odd 1536 cap... :vomiting_face:

csrc/scheduler/normalization_inner.cpp
jjsjann123: 📌 ^^^ This hasn't been updated yet.

csrc/scheduler/normalization_inner.cpp
jjsjann123: oh sorry I must have been blind... the compare operator looks correct here. Somehow I read them wrong earlier.

csrc/scheduler/normalization_inner.cpp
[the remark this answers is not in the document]
liqiangxl: new knowledge!

csrc/scheduler/normalization_inner.cpp
liqiangxl: updated.

csrc/scheduler/normalization_inner.cpp
liqiangxl: I added some explanations on why softmax is so special and creasted an issue to track: https://github.com/NVIDIA/Fuser/issues/1876

jjsjann123 (commented):
1. For code refactor, we are tracking that in #1873 . Which given that it further added more code, I'm somewhat not so happy with it. We can visit that later.

2. The softmax issue that we have to specialize the heuristic for is tracked in #1876 . I don't think we have a good reasoning behind why yet, but since our special case can cover the perf regression, we can just merge this PR as-is and we'll follow up with the discussion in the separate issue.

I think this PR is good-to-go besides those, tagging @naoyam for a sanity check.

csrc/scheduler/normalization_inner.cpp
```diff
@@ -252,6 +263,335 @@ int64_t getMaxPersistentBatch(
   return std::min(max_batches_per_block, batch_from_register);
 }
 
+// calculate bdimx, bdimy, occupancy, given a persistent batch size
```
jjsjann123, line 266: the comment here is meant for `getHeuristicParamsGivenPerisisentBatchSize`

csrc/scheduler/normalization_inner.cpp
```diff
@@ -252,6 +263,335 @@ int64_t getMaxPersistentBatch(
...
+  auto reg_occ = getMaxRegisterCountPerThreadAndOccupancy(
+      persistent_buffer_size,
+      threads_per_block,
+      target_warps_per_sm,
+      free_registers);
+  params.register_per_thread = reg_occ.first;
+  params.occupancy = reg_occ.second;
+  params.free_registers = params.register_per_thread -
+      persistent_buffer_size / scheduler_utils::bytes_per_register;
+  // (4) Calculate other quantities reflecting the quality of the heuristic.
```
jjsjann123, line 337: nitpick, what the `(4)` here?

jjsjann123 (commented):
LGTM.

csrc/scheduler/normalization_inner.cpp
```diff
@@ -252,6 +263,335 @@ int64_t getMaxPersistentBatch(
...
+//     (a) Prioritize occupancy exceeding target.
+//         Ensures minimum required occupancy is surpassed.
+//     (b) Favor divisibility by persistent batch size.
+//         Aims for even workload distribution.
+//     (c) Opt for more free registers.
+//         Maximizes compiler optimization potential.
+//     (d) Seek larger occupancy.
+//         Exceeds the target minimum for better performance.
+//     (e) Use large persistent batch size as a tiebreaker.
+//         Use more registers for persistent buffers.
```
jjsjann123, line 418: One last word here is that, I don't think we have gone through great length in discussing the criterion here. But our end performance looked pretty good so I'm fine with us moving forward with this. :stuck_out_tongue_winking_eye:

jjsjann123 (approved):
:ship:

csrc/scheduler/normalization_inner.cpp
liqiangxl: refering to `n_persistent_tails`, it is not a heuristic para but it is an indicator of the quality of the heuristic. In other words, it is a para used in the sorting.

liqiangxl:
!build

csrc/scheduler/normalization_inner.cpp
```diff
@@ -252,6 +263,335 @@ int64_t getMaxPersistentBatch(
...
+  }
+
+  // prefer reduction count after vectorization is divisible by persistent
+  // batch size
+  score = compare(ha.n_persistent_tails == 0, hb.n_persistent_tails == 0);
+  if (score != 0) {
+    return score > 0;
+  }
+
+  // Large register overhead leaves more registers for the compiler to
```
naoyam, line 379: Please use consistent naming. register overhead -> free registers

csrc/scheduler/normalization_inner.cpp
```diff
@@ -252,6 +263,335 @@ int64_t getMaxPersistentBatch(
...
+  // optimize generated SASS code and avoids register spills. But don't want
+  // to use a very large block size which increaes inter-thread communication.
```
naoyam, line 381: I don't disagree that larger block sizes would have larger overheads, but why should this particular condition be used? Why not, for example, just having something like `compare(ha.padded_bdimx < hb.padded_bdimx)`? 

If there's any backing reasoning, please leave it as a comment. If not, then please mention that it is experimentally found to work well.

csrc/scheduler/normalization_inner.cpp
```diff
@@ -252,6 +263,335 @@ int64_t getMaxPersistentBatch(
...
+    const PrimDataType index_type,
+    const bool has_exp_op) {
+  // Define two free parameters used in this heuristic.
+  // free_registers is all registers except those for the persistent
+  // buffers. The register in each thread = free_registers +
+  // persistent_buffer_size / bytes_per_register
+  // Current values are based on tests of sofmax, layer_norm, softmax_dropout,
+  // dropout_layer_norm on A100 & H100. It directly affects maxregcount passed
+  // to NVRTC and influences the occupancy.
+  const int64_t free_registers = has_exp_op ? 32l : 16l;
```
naoyam, line 439: Since there are a lot of parameters, it's very important to make naming be accurate and precise. `free_registers` is defined as "all registers except those for the persistent buffers." Obviously, `free_registers` here does not mean that but it just means the minimum free registers we should enforce.

naoyam (commented):
Some comments for improving the readability

csrc/scheduler/normalization_inner.cpp
liqiangxl: what about rename `free_registers` to `min_non_buffer_registers`?

naoyam: Why not just overhead? Then, the logic would be that we would want to ensure the free register count is larger than the overhead, which seems to be more clear.

csrc/scheduler/normalization_inner.cpp
liqiangxl: @naoyam this dicussion. we were using `register_overhead`, but the sort algorithm prefers a larger value, then the logic doesn't seem right since we usually want to have a small overhead instead of a large one.

naoyam: Not sure what you mean. It actually doesn't prefer a larger value but just prefers a heuristics that has a larger free registers than the set minimum count. If both heuristics have enough registers, there's no preference made. Am I understanding it correctly?

liqiangxl: yes. your understanding is correct.

csrc/scheduler/normalization_inner.cpp
jjsjann123: Agree on this one: https://github.com/NVIDIA/Fuser/pull/1849#discussion_r1523952768

csrc/scheduler/normalization_inner.cpp
jjsjann123: I guess this variable can be still named as `register_overhead`.

I was complaining earlier on the member variable inside `HeuristicParams`, the two doesn't have to be the same name (and they probably shouldn't since they mean two different things).

liqiangxl:
Thanks @naoyam and @jjsjann123 for comments.
I created a PR to revise code comments and var names.
https://github.com/NVIDIA/Fuser/pull/1948

PR #1853, opened 2024-02-29
Multidevice staged reduction with composed manual and auto-scheduling

samnordmann:
This PR implements the first composition of **manual interdevice scheduling** and **automatic intradevice scheduling.**

Besides being useful in itself, we also hope that this could serve as an example to help the team modify other schedulers to support inter-/intra- device scheduling in a wider range of scenario.

## What

Let us focus on the test implemented in `test/test_multidevice_pipeline.cpp`. There, we consider a simple fusion which expresses a multi-stage reduction:

```
  int num_devices = communicator->size();
  
  TensorView* tv0 = makeConcreteTensor([num_devices, 8, 32]);
  TensorView* tv_out = sum(tv1, {0});
  fusion->addInput(tv0);
  fusion->addOutput(tv_out);
```

_(Note that, for now, we only support reduction which are either non-resharding (i.e. not reducing a sharded dimension) or resharding but affecting only one axis. On the contrary, as far as intra-device scheduling is concerned, Fuser typically requests consecutive reductions to be merged. This inconsistency is only superficial, as we can always translate one case to another by splitting/merging consecutive RedOps)_

We decide to manually shard the tensors on their outmost axis:

```
  std::vector<int64_t> devices(num_devices);
  std::iota(devices.begin(), devices.end(), 0);
  DeviceMesh mesh(devices);
  for (auto tv : {tv0, tv1, tv_out}) {
    tv->setDeviceMesh(mesh);
  }
  for (auto tv : {tv0, tv1}) {
  }
```

This constitutes our multidevice scheduling.

Then, we test composing this interdevice scheduling with different modes of intra-device scheduling:

- No scheduling: which is the only mode supported before this patch
- Manual scheduling: where we try a typical fully featured reduction scheduling. Supporting this mode only required very minor changes.
- Automatic scheduling using `scheduleReduction` and `FusionExecutor`: this required careful modification of the reduction scheduler.
- Fully automated segmentation+scheduling using `FusionExecutorCache`: which is our ultimate target.

## How

The core modifications of the reduction scheduler can be simply explained as follow. Before this patch, the scheduler would first transform the underlying TensorDomain into the canonical form

```
{all_iteration_domains_merged, all_reduction_domains_merged}
```

Note that `all_iteration_domains_merged`  can possibly be empty. We modify this canonical form to be

```
{possibly_one_sharded_outmost_axis, all_iteration_domains_merged, all_reduction_domains_merged}
```

Then, we only need to possibly offset the axis we operate upon, in order to leave untouch the outmost sharded axis.

We made sure that the behavior of the the scheduler remains unchanged in the absence of multidevice scheduling.

### Other  modifications:

- MultiDeviceExecutor: handle passing the launch params, and “bookkeeping” of the inputs (since now some symbolic NamedScalars such as gridDim can be inputs)
- PrepareRuntimeOrder: add symbolic extents of the LeafDomain
- multidevice Test: add the option to validate using prescribed tensors.


csrc/multidevice/executor.cpp
```diff
@@ -87,13 +87,14 @@ std::unordered_map<Val*, c10::IValue> MultiDeviceExecutor::allocateRecvBuffers(
     return {};
   }
 
-  FusionExecutorCache fec(std::move(fusion_copy), 0, false);
-  auto buffers = fec.allocOutputSpace(global_inputs_IValues);
+  FusionExecutor fe;
```
samnordmann, line 90: this change will be overwritten by https://github.com/NVIDIA/Fuser/pull/1900

naoyam:
If not too much of work, could you split the PR? I suppose at least the third and last modes can be done as a separate PR.

> No scheduling: which is the only mode supported before this patch
> Manual scheduling: where we try a typical fully featured reduction scheduling. Supporting this mode only required very minor changes.
> Automatic scheduling using scheduleReduction and FusionExecutor: this required careful modification of the reduction scheduler.
> Fully automated segmentation+scheduling using FusionExecutorCache: which is our ultimate target.

test/test_multidevice_pipeline.cpp
```diff
@@ -304,6 +304,135 @@ INSTANTIATE_TEST_SUITE_P(
...
+    case SchedulingMode::reductionScheduler: {
+      auto reduction_params = getReductionHeuristics(
+          fusion.get(), {at::empty(input_sizes, tensor_options)});
+      NVF_CHECK(reduction_params, "Reduction schedule was not generated!");
+      l_params = reduction_params->lparams;
+      scheduleReduction(fusion.get(), *reduction_params);
+      break;
+    }
+    case SchedulingMode::manualScheduling: {
+      // clang-format off
```
wujingyue, line 383: If it's just about clang-format messing up the comments, put (multiple) `clang-format off` and `on` around the comments, so other real code is still checked.

test/multidevice.cpp
```diff
@@ -93,12 +93,13 @@ void CommunicationTest::resetDstBuffers() {
 // It compares the given (possibly sharded) output with the result of the Fusion
 // run on a single device with the given (possibly sharded) inputs
 void PipelineTest::validate() {
-  // execute the fusion on one device without pipeline scheduling
-  auto fusion_copy = std::make_unique<Fusion>(*runtime->completeFusion());
-  unshard(fusion_copy.get());
-  FusionExecutorCache unsharded_fec(std::move(fusion_copy));
-  auto ref_unsharded_outputs =
-      unsharded_fec.runFusionWithInputs(unsharded_inputs);
+  if (ref_unsharded_outputs.empty()) {
```
wujingyue, line 96: This adds complexity because ref_unsharded_outputs may be filled outside this function and whether it's filled affects the logic inside the function. Is it possible to avoid this added complexity in this PR?

wujingyue:
> If not too much of work, could you split the PR? I suppose at least the third and last modes can be done as a separate PR.
> 
> > No scheduling: which is the only mode supported before this patch
> > Manual scheduling: where we try a typical fully featured reduction scheduling. Supporting this mode only required very minor changes.
> > Automatic scheduling using scheduleReduction and FusionExecutor: this required careful modification of the reduction scheduler.
> > Fully automated segmentation+scheduling using FusionExecutorCache: which is our ultimate target.

That would make it much easier to review!

samnordmann:
> > If not too much of work, could you split the PR? I suppose at least the third and last modes can be done as a separate 
> That would make it much easier to review!

Sure no problem, here is the part 1
https://github.com/NVIDIA/Fuser/pull/1925

test/multidevice.cpp
samnordmann: porting to other pr:
https://github.com/NVIDIA/Fuser/pull/1925/files#r1523869069

test/test_multidevice_pipeline.cpp
samnordmann: ok!
https://github.com/NVIDIA/Fuser/pull/1925/files#r1523870318

samnordmann:
@naoyam @wujingyue @cowanmeg and others, this PR is ready for review, now that part 1 https://github.com/NVIDIA/Fuser/pull/1925 has been merged

tests/cpp/test_multidevice_pipeline.cpp
```diff
@@ -307,14 +307,20 @@ INSTANTIATE_TEST_SUITE_P(
 enum class SchedulingMode {
   InterDeviceOnly,
   Manual,
+  ReductionScheduler,
+  Automatic,
 };
```
wujingyue, line 312:
```suggestion
enum class SchedulingMode {
  Manual,
  InterDeviceOnly,
  ReductionScheduler,
  Automatic,
};
```
I'd like to sort them from most manual to most automatic. Correct me if the order is wrong.

Also, I'd add verbose comments on what each means, especially now that we have some options between fully manual and fully automatic.

wujingyue (commented):
Still reviewing... Thanks for splitting things into multiple PRs!

tests/cpp/test_multidevice_pipeline.cpp
samnordmann: The way it was sorted before was the right order. I've added some comments!

csrc/kernel_cache.cpp
```diff
@@ -377,10 +377,13 @@ void prepareRuntimeOrder(
 
     if (auto input_tv = dynamic_cast<TensorView*>(input_val)) {
       auto root_dom = TensorDomain::noReductions(input_tv->getRootDomain());
-      for (const size_t dim : c10::irange(root_dom.size())) {
-        const auto extent = root_dom[dim]->getMaybeExpandedExtent();
-        available_input.insert(extent);
-        runtime_workspace.group_extent_binding_order.push_back(extent);
+      auto leaf_dom = TensorDomain::noReductions(input_tv->getLeafDomain());
+      for (const auto& dom : {root_dom, leaf_dom}) {
```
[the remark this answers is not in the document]
samnordmann: You're probably right, my test doesn't complain, let's see eif CI does

samnordmann:
!build

csrc/multidevice/executor.cpp
```diff
@@ -182,7 +203,8 @@ void MultiDeviceExecutor::postCommunication(SegmentedGroup* group) {
 }
 
 std::vector<at::Tensor> MultiDeviceExecutor::runWithInput(
-    const std::vector<c10::IValue>& inputs) {
+    const std::vector<c10::IValue>& inputs,
+    LaunchParams l_params) {
```
wujingyue, line 207: Is there a problem with const ref?

csrc/kernel_cache.cpp
```diff
@@ -376,9 +376,8 @@ void prepareRuntimeOrder(
     available_input.insert(input_val);
 
     if (auto input_tv = dynamic_cast<TensorView*>(input_val)) {
-      auto root_dom = TensorDomain::noReductions(input_tv->getRootDomain());
-      for (const size_t dim : c10::irange(root_dom.size())) {
-        const auto extent = root_dom[dim]->getMaybeExpandedExtent();
+      for (auto id : TensorDomain::noReductions(input_tv->getLeafDomain())) {
```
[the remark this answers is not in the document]
samnordmann: I'm hitting an assert without this change, mainly because the NamedScalar "gridDim.x" is not contained in the root domain but is only in the leaf.
I suspect we never encountered this issue before because the segmenter was always used before the scheduler, but in `SchedulingMode::ReductionScheduler` it is used after.
Also, in theory, I think that leaf domain makes more sense here because this is indeed the domains tha the kernel sees to creates the loop, but I might be wrong. Let me know your thoughts

tests/cpp/test_multidevice_pipeline.cpp
```diff
@@ -304,17 +304,31 @@ INSTANTIATE_TEST_SUITE_P(
         ::testing::Values(true),
         ::testing::Bool()));
 
+// Different scheduling modes used in
+// PipelineTestStagedReduction.StagedReduction
 enum class SchedulingMode {
+  // Manual interdevice scheduling, no intra-device scheduling
```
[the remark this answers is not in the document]
samnordmann: I am not sure to understand. Why all alias op? Here the fusion features a reduction for example.
This mode simply means that we don't apply scheduling primitive (`split`, `merge`, `parallelize`, `setMemoryType`, `inlineAt`, etc) besides the fusion definition, so everything will be "by default", and basically all axis will be `ParallelType::Serial`.
What do you mean by most fusion "won't work"? It always work in all of our tests

tests/cpp/test_multidevice_pipeline.cpp
```diff
@@ -304,17 +304,31 @@ INSTANTIATE_TEST_SUITE_P(
...
   InterDeviceOnly,
+  // Manual inter-/intra-device scheduling
   Manual,
+  // Manual inter-device scheduling, composed with ReductionScheduler
+  // intra-device schedule
+  ReductionScheduler,
+  // Manual inter-device scheduling, composed with fully automated intra-device
+  // scheduling (through FusionExecutorCache)
```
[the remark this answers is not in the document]
samnordmann: Indeed! This PR is the first towards this goal

csrc/multidevice/executor.cpp
samnordmann: Let me change that

tests/cpp/test_multidevice_pipeline.cpp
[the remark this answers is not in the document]
wujingyue: > basically all axis will be ParallelType::Serial

Get it. So these unscheduled intra-device fusions do "work", just super slow because they are executed by one thread in loops. Is my understanding right?

csrc/multidevice/executor.h
```diff
@@ -111,8 +113,9 @@ class MultiDeviceExecutor {
 
  private:
   // execute locally a SegmentedGroup that does not involve inter-device
-  // communication
-  void postKernel(SegmentedGroup* group);
+  // communication. Launch Params are used only if
+  // params_.use_fusion_executor_cache = true
```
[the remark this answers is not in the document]
samnordmann: Right, it's the opposite (need to change "true" to "false"), thanks!

tests/cpp/test_multidevice_pipeline.cpp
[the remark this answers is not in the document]
samnordmann: correct!

csrc/kernel_cache.cpp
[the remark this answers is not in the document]
samnordmann: but it might that in theory we need both leaf and root domain, but at least we don't encounter such a case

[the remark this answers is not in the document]
cowanmeg: Just a clarification so I can understand better. There are two pathways
(1) Automatic scheduling using scheduleReduction and FusionExecutor: this required careful modification of the reduction scheduler.
(2) Fully automated segmentation+scheduling using FusionExecutorCache: which is our ultimate target.

Is this change needed for both or just one of the paths?

tests/cpp/test_multidevice_pipeline.cpp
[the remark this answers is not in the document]
wujingyue: So back to my initial question -- what's the practical benefit of this mode? AFAICS, ReductionScheduler mode also generates functional but slow fusions, but it's strictly better because the fusions at least do reduction fast?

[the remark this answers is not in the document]
samnordmann: by practical benefit do you mean "performance"?
The practical benefit here is just that we need to support all these situations.

> ReductionScheduler mode also generates functional but slow fusions

why slow fusion?

> but it's strictly better because the fusions at least do reduction fast?

yes

csrc/kernel_cache.cpp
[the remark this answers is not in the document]
samnordmann: This change is needed only for path (1). I would say this is part of additional "book keeping" we need to handle -- in this case it is handling the named scalars `gridDim` and `blockDim`

tests/cpp/test_multidevice_pipeline.cpp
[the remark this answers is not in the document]
wujingyue: > by practical benefit do you mean "performance"?

That or fixing functional bugs or improving developer velocity. 

> Or more precisely, the "InterDeviceOnly" mode is the only thing we support before this patch, and it is the sinmplest situation, but the slowest

That matches my understanding. I think the practical benefit of this (testing) mode is developer velocity. If we see a test failing in this mode, we are certain that some inter-device stuff (e.g. communication) is broken. 

> why slow fusion?

Non-reduction intra-device segments will run as slow as not scheduled -- serialized to one thread.

tests/cpp/test_multidevice_pipeline.cpp
```diff
@@ -304,17 +304,31 @@ INSTANTIATE_TEST_SUITE_P(
...
+// Different scheduling modes used in
+// PipelineTestStagedReduction.StagedReduction
 enum class SchedulingMode {
+  // Manual interdevice scheduling, no intra-device scheduling
   InterDeviceOnly,
+  // Manual inter-/intra-device scheduling
   Manual,
+  // Manual inter-device scheduling, composed with ReductionScheduler
+  // intra-device schedule
+  ReductionScheduler,
```
wujingyue, line 316:
```suggestion
  ReductionOnly,
```

Or `Predefined` in case you'll soon apply more than just the reduction scheduler.

tests/cpp/test_multidevice_pipeline.cpp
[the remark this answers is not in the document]
samnordmann: right!

tests/cpp/test_multidevice_pipeline.cpp
samnordmann: No problem to change. But actually I think the word "auto" or "automatic" should be present for clarity. What about "AutoReductionSchedulingOnly"?

wujingyue: Up to you. 

I don't think any of these names (ReductionScheduler, ReductionOnly, AutoReductionSchedulingOnly, or ReductionAutoScheduler) is ambiguous, so I'd optimize for conciseness and therefore proposed ReductionOnly. "Only" is to emphasize we are not testing running other schedulers at this moment. I'll be probably fine with simply `Reduction`. 

"Auto" has different meanings for different people. For me, applying just the reduction scheduler is a "manual" selection, and running FEC with segmentation and all schedulers turned on is "auto". But I'm sure you want it to be present for good reasons.

csrc/multidevice/executor.cpp
```diff
@@ -98,10 +98,29 @@ MultiDeviceExecutor::MultiDeviceExecutor(
       copyFusionAndChangeOutputs(completeFusion(), vals_to_allocate_);
 }
 
-void MultiDeviceExecutor::postKernel(SegmentedGroup* group) {
+void MultiDeviceExecutor::postKernel(
+    SegmentedGroup* group,
+    const LaunchParams& l_params) {
```
wujingyue, line 103:
```suggestion
    const LaunchParams& launch_params) {
```

The characters saved don't seem to be worth the confusion created for first-time readers.

csrc/multidevice/executor.cpp
```diff
@@ -98,10 +98,29 @@ MultiDeviceExecutor::MultiDeviceExecutor(
...
   if (!should_run_.at(group)) {
     return;
   }
+
+  if (!params_.use_fusion_executor_cache) {
+    // erase from group's input extent corresponding to grid and block dims. Set
+    // them through launchparams.
+    std::vector<Val*> group_input_vals;
+    for (auto input : group->inputs()) {
+      if (input->isA<NamedScalar>()) {
```
wujingyue, line 113:
```suggestion
      if (input->isScalar()) {
```
Is there a reason to treat named ones specially?

csrc/multidevice/executor.cpp
```diff
@@ -98,10 +98,29 @@ MultiDeviceExecutor::MultiDeviceExecutor(
...
+        NVF_ERROR(input->as<NamedScalar>()->getParallelDim().has_value());
+        NVF_ERROR(l_params.hasDim(
+            input->as<NamedScalar>()->getParallelDim().value()));
+        continue;
```
wujingyue, line 117: Is this safe to `continue`? A fusion segment may take scalars IIUC.

tests/cpp/test_multidevice_pipeline.cpp
samnordmann: I changed to `ReductionOnly` as you suggested

csrc/multidevice/executor.cpp
samnordmann: yes, this is on purpose.
First of all, let me recall that we don't support scalar inputs for now (just for simplicity). Here, the NamedScalar correspond to the parallel dimension "gridDim" and "blockDim", which are not "real" inputs of the Fusion, but rather symbolic extent inserted by the ReductionScheduler that need to be binded to concrete values at runtime. This is why we treat them in a special path. That is why we "`continue`": we don't know need to treat them as actual inputs, and the `FusionExecutor` how to bind those `Val` with the concrete values contained in the `launch_params`. Let me recall the two asserts in the code ensuring all of the above:
```
NVF_ERROR(input->as<NamedScalar>()->getParallelDim().has_value());
NVF_ERROR(l_params.hasDim(
    input->as<NamedScalar>()->getParallelDim().value()));
```

Now, Let me point out that this change is only necessary for `SchedulingMode::ReductionOnly`, in which we use `FusionExecutor` and not `FusionExecutorCache`. So I would call that as well the "additional bookkeeping" annoying but necessary in this case, as we already talked about multiple times.
This remark also holds for the changes in `kernel_cache.cpp` (see [this comment](https://github.com/NVIDIA/Fuser/pull/1853#discussion_r1534688986))

So, this change is exactly the kind of hack around that I don't know how to do better when using FusionExecutor. This falls back to our previous discussions.

csrc/multidevice/executor.cpp
samnordmann: this should also answer this comment
https://github.com/NVIDIA/Fuser/pull/1853#discussion_r1535722505

samnordmann:
By the way, the review is pretty advanced by now, but if it helps I can further divide the PR by separating out 1) `SchedulingMode::Automatic` containing the patch on the scheduler, and 2) `SchedulingMode::ReductionOnly` containing the additional patch on the runtime inside `MultiDeviceExecutor::postKernel` and `prepareRuntimeOrder` (from `kernel_cache.cpp`).

Let me know

csrc/multidevice/executor.cpp
wujingyue: > we don't support scalar inputs for now (just for simplicity)

Can you clarify what support is missing? AFAICS from [this test](https://github.com/NVIDIA/Fuser/blob/5817da992a052d6005fd22621f4140fbc3bcfa6c/tests/cpp/test_scalar_hoisting.cpp#L347-L349), FusionExecutor does support scalar inputs. So I doubt simplicity is the real reason.

samnordmann: I think nothing fundamental is missing, this is just a simplifying assumption we make (among others). But someone could go and add this support :) It's one of our TODO. I think it should be pretty straightforward but we haven't done it and it's not a priority in my opinion.

samnordmann: (when I say "we don't support" I'm talking exclusively about the multidevice module)

csrc/kernel_cache.cpp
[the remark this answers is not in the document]
wujingyue: > I suspect we never encountered this issue before because the segmenter was always used before the scheduler, but in SchedulingMode::ReductionScheduler it is used after.

Yes, very likely. 

> but it might that in theory we need both leaf and root domain, but at least we don't encounter such a case

@naoyam , wdyt? Does it make more sense to add the rfactor domain instead? AFAIU, this code is trying to add input dependencies for consuming ops. Those use rfactor (not leaf, unless it happens to be the same) extents of their producers.

wujingyue (commented):
LGTM with comments.

csrc/kernel_cache.cpp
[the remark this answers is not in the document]
samnordmann: it does not! :)

csrc/multidevice/executor.cpp
samnordmann: > it's same code complexity but covers more cases (allowing non-grid/block-dim scalars).

IIUC you suggest to move the asserts into the "if" condition, which I did.

> FusionExecutor already have problems taking grid/block dims and therefore we have to erase them?

It is not that FusionExecutor has a problem taking those inputs. It's more that `blockDim` and `gridDim` need to be passed to  `FusionExecutor::runFusion` (at least to match the expected number of inputs!), while being of a different nature than the other inputs because they come from the scheduler's `launch_params`.

I'm trying to clarify the situation as much as I can here:
1. First, the scheduler is used on the Fusion and includes some symbolic extents `gridDim` and `blockDim` in the tensor.
2. Then, the segmenter is used, and it treats these symbolic extents as group's I/O, and add them in the group's `input_vals`. So the group's inputs are `{input 1, ..., input n, blockDim, gridDim}`. But we see here that we are mixing two inputs of different natures: real inputs and inputs coming from the scheduler's `launch_params`. Let me point out here that this situation was not encountered before because this is the first time we use the segmenter AFTER scheduling. 
Whether the segmenter should add `gridDim` and `blockDim` as group's input is a question in itself, I think it makes sense in terms of dependency, but it requires a special handling later on (as the one I implemented here), so that's debatable. I chose not to change the scheduler and to handle it manually.
3. Inside `MultiDeviceExecutor::postKernel`, the segmented group is compiled into a kernel, and we need to create the kernel's concrete input called `group_input_IValues` to pass it to `FusionExecutor::runFusion`. We populate `group_input_IValues` by retrieving concrete values from the global context of previously stored values. However, `blockDim` and `gridDim` should not be retrieved from global context per say but rather from the`launch_params` (passed from the scheduler to our MultiDeviceExecutor). The question is: "what to do with `blockDim` and `gridDim`". We could either:
  a. retrieve the concrete values manually from `launch_params` and add then to `group_input_IValues`
  b. remove them from group's input, pass the `launch_params` to `FusionExecutor::runFusion` and rely on that function for handling it (which is already implemented).
I chose solution b. because a. is redundant with `FusionExecutor::runFusion` already handling the `launch_params`.

I hope this detailed explanation makes things clearer. Let me know if it's still unclear, or if you have a better suggestion.

samnordmann:
Thanks @wujingyue, I adressed all your comments.
@naoyam let me know your feedbacks.

naoyam:
Will take a look.

csrc/kernel_cache.cpp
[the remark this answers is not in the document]
naoyam: What assert is this?

> I'm hitting an assert without this change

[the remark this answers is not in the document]
naoyam: Could you please give more details why that assertion fails too?

csrc/multidevice/executor.cpp
naoyam: > So the group's inputs are {input 1, ..., input n, blockDim, gridDim}.

Maybe the real problem is the scheduler adding `blockDim` and `gridDim` as inputs. They are not inputs but intrinsic variables automatically defined in CUDA. What do you think, @jacobhinkle ?

csrc/scheduler/utils.cpp
```diff
@@ -243,6 +244,7 @@ size_t mergeReduction(TensorView* tv) {
 }
 
 size_t mergeNonReduction(TensorView* tv) {
+  bool has_outmost_device_dim = false;
```
naoyam, line 247:
```suggestion
  bool has_outermost_device_dim = false;
```

csrc/scheduler/utils.cpp
```diff
@@ -252,6 +254,10 @@ size_t mergeNonReduction(TensorView* tv) {
     if (tv->axis(i)->isReduction()) {
       continue;
     }
+    if (tv->axis(i)->isDeviceDim()) {
```
naoyam, line 257: `i` is not guaranteed to be the outermost position, right? I suspect you're just assuming that's the case? If so, let's assert that:

```
NVF_ERROR(i == 0, "Invalid device parallelization detected at ", i);
```

csrc/multidevice/executor.cpp
```diff
@@ -98,10 +98,29 @@ MultiDeviceExecutor::MultiDeviceExecutor(
...
+void MultiDeviceExecutor::postKernel(
+    SegmentedGroup* group,
+    const LaunchParams& l_params) {
   if (!should_run_.at(group)) {
     return;
   }
+
+  if (!params_.use_fusion_executor_cache) {
+    // erase from group's input extent corresponding to grid and block dims. Set
+    // them through launchparams.
```
wujingyue:
```suggestion
    // Erase from group's input extent corresponding to grid and block dims. Set
    // them through launchparams. When use_fusion_executor_cache is off, we schedule before segmentation. The segmenter treats `blockDim` and `gridDim` as group's I/O, and add them in the group's `input_vals`. So the group's `input_vals` are `{input 1, ..., input n, blockDim, gridDim}`. However, `FusionExecutor::runFusion` expects `inputs` to be `{input 1, ..., input n}` and `launch_constraints` to contain `blockDim` and `gridDim`. 
```

csrc/multidevice/executor.cpp
wujingyue: Thanks -- that clarifies lots of my questions and is the level of details I wanted to hear. I put some of your words into the code comment for posterity.

csrc/scheduler/reduction_utils.cpp
```diff
@@ -30,9 +31,11 @@ TensorView* scheduleReductionTV(
   // Inner here though is only relative to the other axis. When
   // rparams.fastest_dim == false, the reduction axis is logically outside the
   // iteration axis.
-  const int iter_axis = 0;
+  const int has_outmost_dim_sharded = isSharded(reduction_tv);
```
naoyam: nit: we use "outer" more often than "out" when referring to positions in a domain.

```suggestion
  const int has_outermost_dim_sharded = isSharded(reduction_tv);
```

csrc/multidevice/executor.cpp
wujingyue: > First, the scheduler is used on the Fusion and includes some symbolic extents gridDim and blockDim in the tensor.

How so? I thought we do only inter-device scheduling before segmentation, so gridDim and blockDim shouldn't be there at least conceptually.

csrc/scheduler/reduction_utils.cpp
naoyam: Please add a comment about this. For example, we should note that we assume only the outermost domain can be parallelized with DIDx at this point and in that case this reduction scheduler only schedules the remaining domains while leaving the DIDx domain unchanged.

csrc/scheduler/reduction_utils.cpp
```diff
@@ -30,9 +31,11 @@ TensorView* scheduleReductionTV(
...
+  const int iter_axis = has_outmost_dim_sharded ? 1 : 0;
   const int outer_reduce_axis = rparams.schedule_3D ? 1 : 0;
-  const int inner_reduce_axis = rparams.schedule_3D ? 2 : has_iter_axis ? 1 : 0;
+  const int inner_reduce_axis =
+      rparams.schedule_3D ? 2 : has_outmost_dim_sharded + has_iter_axis;
```
naoyam, line 38: When `rparams.schedule_3D` is true, shouldn't this be 3 when sharded?

csrc/multidevice/executor.cpp
samnordmann: > Maybe the real problem is the segmenter adding `blockDim` and `gridDim` as inputs. They are not inputs but intrinsic variables automatically defined in CUDA. What do you think, @jacobhinkle ?

From the CPU perspective they are actually launch parameters, so user can control it, in this sense it is an input.

> How so? I thought we do only inter-device scheduling before segmentation, so gridDim and blockDim shouldn't be there at least conceptually.

No, doing intra-device scheduling is precisely the goal of this pr. See the test and `SchedulingMode::ReductionOnly`

samnordmann:
we assert in the code that we don't mix 3D reduction and multidevice. I mainly assume this for simplicity. Is it a problem?
Btw: why is 3D important in general?

csrc/scheduler/reduction_utils.cpp
samnordmann: we assert in the code that we don't mix 3D reduction and multidevice. I mainly assume this for simplicity. Is it a problem?
Btw: why is 3D important in general?

samnordmann: I add an assert at this place too

csrc/kernel_cache.cpp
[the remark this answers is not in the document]
samnordmann: > What assert is this?
> 
> > I'm hitting an assert without this change

It's the assert at the end of that function body, `csrc/kernel_cache.cpp:420`

This assert makes sense, since `gridDim` is an input of the group, that dependency needs to be satisfied. That's why we need to add this extent in `available_input`, and therefore iterate on the leaf domain.

However, let me point out that if we decide to remove `gridDim` from the segmented group's input (as suggested in https://github.com/NVIDIA/Fuser/pull/1853#discussion_r1537941767), this modification will probably not be needed anymore.

csrc/multidevice/executor.cpp
samnordmann: > SchedulingMode::Auto implements flow 2.

yes



> SchedulingMode::ReductionOnly implements flow 1

not exactly, I would say

Inter AND INTRA-device scheduling => inter-device only segmentation (which means we just segment out the resharding ops)  => FE::runFusion these scheduled intra-device segments (without further segmentation).

wujingyue: > AND INTRA

Is this necessary? Inter-device segmentation needs only DID schedules. If we can move the intra-device scheduling after segmentation, I think it will save us some unnecessary complexity (e.g. this one)

samnordmann: In my opinion, this is necessary, and the steps are done in the right order, which is:
1) Fusion definition
2) scheduling
3) execution

the user should be able to schedule a Fusion (as done in the test), and then rely on the MultiDeviceExecutor to automatically lower/compile/execute it. Here what you call the "inter-device segmentation" is part of the lowering process, because we need to insert the necessary communications (what I mean: the segmentation step here doesn't make any decision on the parallelization strategy, it just does what is necessary to execute the scheduled Fusion).

does it make sense?

> I think it will save us some unnecessary complexity

I don't think this is an unnecessary complexity, especially if we want to make the segmenter, scheduler, and FusionExecutor a bit more modulable than what it currently is, and understand what implicit assumption they rely on.

This complexity is the type of additional "bookkeeping" necessary when not using FusionExecutorCache.

wujingyue: Sorry if my question wasn't clear. I was asking about intra-device scheduling. Why does that have to happen before inter-device segmentation? 

(And I understood we are talking about a flow that will eventually go away.)

samnordmann: I understood your question correctly, and thought I addressed it, sorry if I haven't.
I think that we shouldn't treat differently inter and intra device scheduling. Both belongs to the same step, and should happen strictly before lowering and execution.
 
If that's still unclear maybe it's better to have this high-level discussion during a call?

samnordmann: Why do you think it should be done in the order you describe?
The "avoided complexity" I can see relates to implicit assumption made in the implementation of the segmenter/scheduler

samnordmann: > (And I understood we are talking about a flow that will eventually go away.)

Sorry I forgot this question.

I think that it's pretty unclear at this point. cf our previous discussion. It will depend on how the segmenter, scheduler, segmenter will be refactored. Here, despite adding implementation complexity, I make the effort to implement two paths (FusionExecutor and FusionExecutorCache) to better anticipate future change, as you advocate during our previous discussions

naoyam: > From the CPU perspective they are actually launch parameters, so user can control it, in this sense it is an input.

But if you leave them as segment inputs, since they are never "produced", that segment would never be ready to run, right?

naoyam: Since this is a temporary workaround, I don't have any strong preference. I think this is just a matter of:

> Maybe the real problem is the segmenter adding blockDim and gridDim as inputs. They are not inputs but intrinsic variables automatically defined in CUDA.

naoyam: I suspect this issue is also automatically fixed:

https://github.com/NVIDIA/Fuser/pull/1853/files/40160a398a71fda65a7af48485430510c92386d9#r1538108135

samnordmann:
@naoyam do you have other comments?

samnordmann:
> > @naoyam do you have other comments?
> 
> I don't think these are resolved yet.
> 
> [#1853 (comment)](https://github.com/NVIDIA/Fuser/pull/1853#discussion_r1535066244) [#1853 (comment)](https://github.com/NVIDIA/Fuser/pull/1853#discussion_r1534552175)

Oh, let me just check that I understand correctly then: do you suggest modifying the segmenter to remove the named scalars from the segmented group's input?

naoyam:
> > > @naoyam do you have other comments?
> > 
> > 
> > I don't think these are resolved yet.
> > [#1853 (comment)](https://github.com/NVIDIA/Fuser/pull/1853#discussion_r1535066244) [#1853 (comment)](https://github.com/NVIDIA/Fuser/pull/1853#discussion_r1534552175)
> 
> Oh, let me just check that I understand correctly then: do you suggest modifying the segmenter to remove the named scalars from the segmented group's input?

I suspect the real problem is in the segmenter, and these changes are just hiding the real problem.

samnordmann:
> I suspect the real problem is in the segmenter, and these changes are just hiding the real problem.

I claim that if we modify the segmenter so that the named scalars are not added to the segmented group's input, then we will solve our issues (more precisely the changes in kernel_cache.cpp and in multidevice/executor.cpp).
Whether the segmenter's current behavior is a "problem" is imo more a matter of convention.

So, if you suggest that we should change the segmenter, I can go and change it.

This being said, I have the feeling it would help to further divide the PR as suggested [here](https://github.com/NVIDIA/Fuser/pull/1853#issuecomment-2015285073), because the important part about the scheduler is ready to be merged

naoyam:
> > I suspect the real problem is in the segmenter, and these changes are just hiding the real problem.
> 
> I claim that if we modify the segmenter so that the named scalars are not added to the segmented group's input, then we will solve our issues (more precisely the changes in kernel_cache.cpp and in multidevice/executor.cpp). Whether the segmenter's current behavior is a "problem" is imo more a matter of convention.
> 
> So, if you suggest that we should change the segmenter, I can go and change it.
> 
> This being said, I have the feeling it would help refining those question to further divide the PR as suggested [here](https://github.com/NVIDIA/Fuser/pull/1853#issuecomment-2015285073), because the important part about the scheduler is independent of those considerations and ready to be merged

Well, if my suspect is right, the segmenter would just need a change of one line and some of the workarounds in this PR can be removed. If not, I'd be happy to leave it as a TODO item, but I'd like to confirm first. I asked this question offline too, but what's the command to reproduce the error of this one?

https://github.com/NVIDIA/Fuser/pull/1853#discussion_r1534552175

samnordmann:
I have opened another PR https://github.com/NVIDIA/Fuser/pull/2035/ to further divide this one into two parts. The new open PR is independent to the blocking discussion 

> Well, if my suspect is right, the segmenter would just need a change of one line and some of the workarounds in this PR can be removed.

I think I see what you mean (but all patches I can think of are a bit brutal, i.e., handling NamedScalar as a special case in the segmenter). Let me know.

> If not, I'd be happy to leave it as a TODO item, but I'd like to confirm first. I asked this question offline too, but what's the command to reproduce the error of this one?

ok! to reproduce: checkout to [this branch](https://github.com/samnordmann/Fuser/tree/staged_reduction_tests_upstream_bug_reproduce) and execute `mpirun -np 2 test_multidevice --gtest_filter=SchedulingModes/PipelineTestStagedReduction.StagedReduction/2`

samnordmann:
@naoyam thanks a lot for that patch! It allows us to get rid of the modifications we needed in `prepareRuntimeOrder` and `MultiDeviceExecutor::runWithInput` which raised concerns.

Let me know if it looks good now. I think it still make sense to first merge https://github.com/NVIDIA/Fuser/pull/2035 to make this one smaller but it is up to you -- we can also merge this one directly if you think it's ok

csrc/scheduler/reduction_utils.cpp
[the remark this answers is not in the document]
naoyam: Looks good to me

naoyam (approved):
LGTM.

PR #1855, opened 2024-02-29
Add DEBUG_SERDE flag to python tests

rdspring1:
This PR adds a `DEBUG_SERDE` environment flag to the python tests to debug serialization failures.
1) It disables automatically saving `FusionCache` upon program exit.
2) It resets the `FusionCache` after each test, which is useful for isolating failures. Note, some failures only occur when running multiple tests together and accumulating fusions in the cache.
3) It keeps the temporary files that are created during `serde_check`. Normally, these files are deleted after each test.


rdspring1:
!build

wujingyue:
How about a per-test knob similar to skip_serde_check as an example? Something like: 

```
def test_foo(...):
  ...
  self.exec_nvfuser(fusion_func, inputs, debug_serde=True)
```

Environment variables are more coarse-grained and harder to reason about by readers.

python_tests/test_python_frontend.py
```diff
@@ -34,6 +35,19 @@
...
+# 1) It disables automatically saving FusionCache upon program exit.
+#
+# 2) It resets the FusionCache after each test, which is useful for isolating
+# failures. Note, some failures only occur when running multiple tests
+# together and accumulating fusions in the cache.
+#
+# 3) It keeps the temporary files that are created during serde_check.
+# Normally, these files are deleted after each test.
+env_var_debug_serde = os.getenv("DEBUG_SERDE", None)
+debug_serde: bool = env_var_debug_serde in ("true", "1")
```
jjsjann123, line 49: nitpick: If this is only a helper for debug, wondering if it makes sense to have a finer control here? using debug_serde to turn on/off each of the three features.

i.e. if debug_serde & 4 -> disables automatically saving FusionCache upon program exit.

jjsjann123 (commented):
I wasn't even aware of skip_serde_check can be used at per-test level. Sounds like a good suggestion from Jingyue.

rdspring1:
I went with an environment variable because we need to disable cache saving on exit in `setUpModule` before running any test.

Just my two cents. I'd prefer to run `DEBUG_SERDE=1 python test_python_frontend.py -k test_foo` rather than manually setting `debug_serde=True/False` for each test.

jjsjann123:
> I went with an environment variable because we need to disable cache saving on exit in `setUpModule` before running any test.
> 
> Just my two cents. I'd prefer to run `DEBUG_SERDE=1 python test_python_frontend.py -k test_foo` rather than manually setting `debug_serde=True/False` for each test.

I'm neutral on this. If this is just used as a debug option. i.e. we are not supposed to tweak this per test to skip any test, maybe it doesn't matter to have a finer control per test? wdyt @wujingyue 

One more nitpick is `DEBUG_SERDE` -> `NVFUSER_DEBUG_SERDE` maybe?

jjsjann123 (approved):
I'm convinced.

python_tests/test_python_frontend.py
```diff
@@ -34,6 +35,19 @@
 
 RUN_NVFUSER = RUN_CUDA and not TEST_WITH_ROCM
 
+# This DEBUG_SERDE environment flag is used to debug serialization failures.
+#
+# 1) It disables automatically saving FusionCache upon program exit.
```
wujingyue, line 40:
```suggestion
# 1) It disables automatically saving FusionCache upon program exit. Therefore, it has to be a global flag not per-test. 
```

python_tests/test_python_frontend.py
rdspring1: That make sense. I think there are three options:

1. Disable automatic saving FusionCache on program exit. --- Option Name: "disable_atexit_cache_saving"
2. Disable accumulating fusions in FusionCache. --- Option Name: "single_fusion_only"
3. Both Options 1 and 2. --- Option Name: ["all", "true", "1"]

python_tests/test_python_frontend.py
```diff
@@ -34,6 +35,19 @@
...
+#
+# 2) It resets the FusionCache after each test, which is useful for isolating
+# failures. Note, some failures only occur when running multiple tests
+# together and accumulating fusions in the cache.
+#
+# 3) It keeps the temporary files that are created during serde_check.
+# Normally, these files are deleted after each test.
+env_var_debug_serde = os.getenv("DEBUG_SERDE", None)
```
wujingyue, line 48:
```suggestion
env_var_debug_serde = os.getenv("DEBUG_SERDE")
```

Per https://docs.python.org/3/library/os.html#os.getenv, the default is None. If you really want to specify the default value explicitly, I'd say `os.getenv("DEBUG_SERDE", default=None)`.

python_tests/test_python_frontend.py
```diff
@@ -50,10 +64,12 @@ def is_pre_ampere():
 
 
 def setUpModule():
-    from nvfuser import enable_automatic_serialization
+    if not debug_serde:
+        from nvfuser import enable_automatic_serialization
+
+        # Turn on default serialization upon program exit
+        enable_automatic_serialization()
```
wujingyue, line 71: Instead of creating a hook to do that upon program exit, how about simply serialize in tearDownModule for less indirection?

python_tests/test_python_frontend.py
```diff
@@ -70,8 +86,12 @@ def inner_fn(*args, **kwargs):
         self, fusion_func, inputs = args
 
         # NOTE: For debug purposes, clear FusionCache before running first test
-        # if ("new_fusion_expected" not in kwargs) or kwargs["new_fusion_expected"]:
-        #    FusionCache.reset()
+        is_new_fusion_expected = ("new_fusion_expected" not in kwargs) or kwargs[
+            "new_fusion_expected"
+        ]
```
wujingyue, line 91:
```suggestion
        is_new_fusion_expected = kwargs.get("new_fusion_expected", true)
```

python_tests/test_python_frontend.py
```diff
@@ -70,8 +86,12 @@ def inner_fn(*args, **kwargs):
         self, fusion_func, inputs = args
 
         # NOTE: For debug purposes, clear FusionCache before running first test
```
wujingyue, line 88:
```suggestion
        # NOTE: For debug purposes, clear FusionCache before running first test so the behavior is more deterministic (#1848). 
```

python_tests/test_python_frontend.py
```diff
@@ -70,8 +86,12 @@ def inner_fn(*args, **kwargs):
...
-        # if ("new_fusion_expected" not in kwargs) or kwargs["new_fusion_expected"]:
-        #    FusionCache.reset()
+        is_new_fusion_expected = ("new_fusion_expected" not in kwargs) or kwargs[
+            "new_fusion_expected"
+        ]
+        if debug_serde and is_new_fusion_expected:
```
wujingyue, line 92: For my education, why `is_new_fusion_expected` imply a cache reset? From the naming, they seem to be orthogonal.

python_tests/test_python_frontend.py
```diff
@@ -86,16 +106,27 @@ def inner_fn(*args, **kwargs):
...
+                FusionCache.reset()
 
-            # Get new FusionCache because the previous one was destroyed by the reset call.
-            fc = FusionCache.get()
-            fc.deserialize(tmp.name)
+                # Get new FusionCache because the previous one was destroyed by the reset call.
+                fc = FusionCache.get()
+                fc.deserialize(tmp.name)
+            except Exception as e:
+                if not debug_serde:
```
wujingyue, line 122: Instead of 
```
if not a:
  false_branch
else:
  true_branch
```
do 
```
if a:
  true_branch
else:
  false_branch
```

python_tests/test_python_frontend.py
```diff
@@ -86,16 +106,27 @@ def inner_fn(*args, **kwargs):
...
+                    raise RuntimeError(
+                        "***** Use DEBUG_SERDE=true to debug serialization failure."
+                    )
+                else:
+                    raise RuntimeError(
+                        f"***** {tmp.name} contains the serialized binary for this failure."
```
wujingyue, line 128: When debug_serde is on, this temp file won't even exist. I guess you wanted to print this when debugging is off.

wujingyue (approved):
LGTM with comments

python_tests/test_python_frontend.py
rdspring1: It is to test the automatic serialization/deserialization API. 
https://github.com/NVIDIA/Fuser/blob/main/csrc/python_frontend/fusion_cache.h#L212-L220

python_tests/test_python_frontend.py
rdspring1: Good catch! It should be `tempfile.NamedTemporaryFile(delete=(not debug_serde))`

python_tests/test_python_frontend.py
rdspring1: When `is_new_fusion_expected` is True, the test expects a new fusion to be added to the cache. Otherwise, an existing fusion is reused.

For serialization, we're allowed to reset the cache only when `is_new_fusion_expected` is True.

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

PR #1860, opened 2024-03-01
Decide whether slices form a split by checking the slice ranges. 

wujingyue:
Currently, the code checks the slicing amounts are the same as the padding amounts. This is hard to get to work when the split/cat axis is `Split`ed or `Merge`ed. The new way works for those cases easily. 

For example,
```
# in: a tensor of shape [6,2]
s0, s1 = in.split([3, 3])  # [3,2], [3,2]
v0 = s0.view(-1)  # [6]
v1 = s1.view(-1)  # [6]
out = torch.cat([v0, v1])
```
The slicing amount of `s0` is 3 and the padding amount of `out` is 6 (because of the merge by 2). The split and cat are still cancellable, but with the current approach we would need to compute the matching amount along the branch.


wujingyue:
!build

wujingyue:
FYI, unit tests are passing: https://gitlab-master.nvidia.com/dl/pytorch/fuser-gh-mirror/-/pipelines/13192696

csrc/preseg_passes/move_split_cat.cpp
```diff
@@ -105,6 +104,87 @@ std::pair<std::vector<PadOp*>, int64_t> getCatInputsAndAxis(CatOp* cat) {
...
+      if ((j == split_axis) != sliced) {
+        return nullptr;
+      }
+    }
+
+    // Collect the slice range for the split axis.
+    split_ranges.push_back(slice_ranges[split_axis]);
+  }
+
+  if (!split_ranges.front().start->isZero()) {
```
Priya2698, line 163: IIUC, the implicit assumption here is that the slice expression ranges are in sorted order?

Priya2698: Is it possible for the slice expressions to not be in the order of their slice ranges? For e.g. if the order of slice expressions is inverted`s0 = in[:, 5:], s1 = in[:, :5]`

Priya2698: > Yes, it's possible. What's the problem with that? Do you worry about that the optimization may misfire or may miss opportunities? 

That it may miss opportunities, although seems less probable.

> FWIW, `exprs` in `exprsFormSplit` are in the order they are concatenated.

Got it.

Priya2698 (approved):

wujingyue:
!build

PR #1867, opened 2024-03-02
add iter grouped grid reduction

liqiangxl:
Following #1800 
Add iteration domain grouped grid reduction.
**TODO:**
(1) revise outer reduction heuristics
(2) combine grouped with regular grid/block reduction function.

**Performance of outer reduction:**
(1) Reduction dim = 2048, iteration dim = 1 to 32K
image
(2) Reduction dim = 16K, iteration dim = 1 to 32K
image

More results avialable in the [doc](https://docs.google.com/document/d/1YI0KX_vmBMJurxooN2_Ya89Niki6zvYCvGDwURTEVCs/edit#heading=h.6behzp60hcgm).

### DIFF
--- a/csrc/scheduler/reduction.cpp
+++ b/csrc/scheduler/reduction.cpp
@@ -835,7 +835,7 @@ std::shared_ptr<ReductionParams> outerReductionHeuristic(
     debug() << "\n===== Reduction Stats ========\n"
             << "total_reduction_numel: " << total_reduction_numel << "\n"
             << "total_iteration_numel: " << total_iteration_numel << "\n"
-            << "vectorize_factor: " << vectorize_factor << "\n"
+            << "vectorize_factor: " << iter_unroll_factor << "\n"
             << "n_tensor_inputs: " << n_tensor_inputs << "\n"
             << "max_input_dtype_size: " << max_input_dtype_size << "\n"
             << "block(" << bdimx << ", " << bdimy << ", 1)" << std::endl;
--- a/csrc/scheduler/reduction.cpp
+++ b/csrc/scheduler/reduction.cpp
@@ -1214,6 +1214,8 @@ void scheduleReduction(Fusion* fusion, const ReductionParams& rparams) {
   // TODO: the var name is confusing, should rename
   // [cross_grid/block_inner_reduction] to [cross_grid/block_reduction], see
   // https://github.com/NVIDIA/Fuser/issues/1863
+  // grouped welford is only enabled for grid persistent.
+  // see validateAndConvertIterDomainGrouping
   const bool has_welford = ir_utils::hasOpsOfType<WelfordOp>(fusion);
   const bool use_iter_grouped_reduction = !rparams.fastest_dim &&
       (has_welford
--- a/csrc/scheduler/reduction.cpp
+++ b/csrc/scheduler/reduction.cpp
@@ -1210,12 +1210,15 @@ void scheduleReduction(Fusion* fusion, const ReductionParams& rparams) {
   const bool vectorize =
       rparams.vectorize_inner_reduction || rparams.vectorize_iter_dom;
 
-  // allow iter domain grouped reduction for block outer reduction.
+  // allow iter domain grouped reduction for block and grid outer reductions.
   // TODO: the var name is confusing, should rename
   // [cross_grid/block_inner_reduction] to [cross_grid/block_reduction], see
   // https://github.com/NVIDIA/Fuser/issues/1863
-  const bool use_iter_grouped_reduction = !rparams.cross_grid_inner_reduction &&
-      !rparams.fastest_dim && rparams.cross_block_inner_reduction;
+  const bool has_welford = ir_utils::hasOpsOfType<WelfordOp>(fusion);
+  const bool use_iter_grouped_reduction = !rparams.fastest_dim &&
+      (has_welford
+           ? rparams.cross_grid_inner_reduction && rparams.persistent_kernel
+           : rparams.cross_block_inner_reduction);
 
   reduction_scheduler_utils::multiReductionInliner(
       fusion,
--- a/runtime/grid_reduction.cu
+++ b/runtime/grid_reduction.cu
@@ -707,7 +707,10 @@ __device__ void iterGroupedGridReduceLastBlock(
       index_utils::maskedOffset<!X_THREAD, !Y_THREAD, !Z_THREAD>(
           threadIdx, blockDim);
 
-  // index into iteration dim
+  // index into iteration dim.
+  // Its calculation is same to that in [iterGroupedGridReduce]. Becuase when
+  // [iterGroupedGridReduceLastBlock] is called from [iterGroupedGridReduce],
+  // X_THREAD, Y_THREAD, Z_THREAD are flipped.
   const auto thread_offset =
       index_utils::maskedOffset<X_THREAD, Y_THREAD, Z_THREAD>(
           threadIdx, blockDim);
@@ -768,6 +771,16 @@ __device__ void iterGroupedGridReduceLastBlock(
   }
 }
 
+// Main algorithm is same to gridReduce: start with block reduce then write
+// results to gmem, the last block load from gmem and finalize with a block
+// reduction. Main differences:
+// (1) each thread in the iter dim does [vec_size] reductions instead of 1.
+// (2) using [blockIterGroupedReduce] instead of [blockReduce].
+// (3) ensures vectorized load/store to gmem.
+// Specifically, the new para [vec_size] is the vecotrization factor in the
+// iteration dimension. It is used in outer reduction to reduce calling this
+// grid reduction from [vec_size] times to only 1 time. Its value is limited
+// to 1, 2, 4, 8, 16 based on the hardware support and input data type.
 template <
     bool X_BLOCK,
     bool Y_BLOCK,
@@ -789,9 +802,7 @@ __device__ void iterGroupedGridReduce(
     T* shared_buf,
     bool read_pred,
     bool write_pred,
-    T init_val,
-    const nvfuser_index_t entrance_ind,
-    const nvfuser_index_t n_entrances) {
+    T init_val) {
   // inp or block reduction results
   T block_reduction_val[vec_size];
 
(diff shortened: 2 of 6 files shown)


liqiangxl:
!build

liqiangxl:
!build

liqiangxl:
!build

liqiangxl:
!build

liqiangxl:
!build

test/test_gpu_outer_reduction.cpp
```diff
@@ -2288,17 +2288,21 @@ TEST_F(OuterReductionTest, OuterReductionMagicScheduler) {
...
+    auto t0 = at::ones(shape, options);
     std::vector<c10::IValue> inputs({t0});
     FusionExecutorCache executor_cache(std::move(fusion));
     auto cg_outputs = executor_cache.runFusionWithInputs(inputs);
     auto ref = t0.to(at::kFloat).sum({0});
     testValidate(
         executor_cache.fusion(), cg_outputs, inputs, {ref}, __LINE__, __FILE__);
   };
+
+  test(2048, 32768);
```
jacobhinkle, line 2300: This is covered by the following loops I believe so you can remove before merging.

test/test_gpu_outer_reduction.cpp
```diff
@@ -2288,17 +2288,21 @@ TEST_F(OuterReductionTest, OuterReductionMagicScheduler) {
...
+
   maybeClearAllocator(0);
   for (int dim0 = 1024; dim0 <= 32768; dim0 *= 2) {
     for (int dim1 = 1024; dim1 <= 32768; dim1 *= 2) {
+      printf("Testing dim0=%d, dim1=%d\n", dim0, dim1);
```
jacobhinkle, line 2305: Same re. remembering to remove before merge.

test/test_serial_gridreduce.cpp
```diff
@@ -36,8 +36,8 @@ namespace nvfuser {
 using SerialGridReductionTest = NVFuserTest;
 
 TEST_F(SerialGridReductionTest, Scheduling) {
-  for (bool serial : {true, false}) {
-    for (int64_t num_warps : {4, 8}) {
+  for (bool serial : {true}) {
+    for (int64_t num_warps : {4}) {
```
jacobhinkle, line 40: Don't forget to remove before merging.

runtime/grid_reduction.cu
```diff
@@ -667,4 +667,243 @@ __device__ void serialReductionStep(
   }
 }
 
+// vectorized reduction
+template <
+    bool X_THREAD,
+    bool Y_THREAD,
+    bool Z_THREAD,
+    bool Aligned,
+    int VECT_FACTOR,
```
jacobhinkle, line 676: FYI this template parameter is called `vec_size` in other places.

test/test_gpu_outer_reduction.cpp
liqiangxl: thanks! Looks like I pushed some debug changes by accident.

runtime/grid_reduction.cu
liqiangxl: changed.

runtime/grid_reduction.cu
```diff
@@ -667,4 +667,243 @@ __device__ void serialReductionStep(
...
+      (Y_THREAD || threadIdx.y == 0) && (Z_THREAD || threadIdx.z == 0);
+  if (should_write && write_pred) {
+#pragma unroll
+    for (int i = 0; i < vec_size; i++) {
+      reduction_op(out[i], inp_tmp[i]);
+    }
+  }
+}
+
+template <
```
jjsjann123, line 771: nitpick: understandably we can refer to `gridReduce` for the main part of documentation. We should put a quick note here on how this method differs from that.

jjsjann123: Especially comment on what the added argument does. i.e. `vec_size`.

runtime/grid_reduction.cu
```diff
@@ -667,4 +667,243 @@ __device__ void serialReductionStep(
...
+    const T* inp_val,
+    Func reduction_op,
+    volatile T* work_buf,
+    int64_t* sync_flags,
+    T* shared_buf,
+    bool read_pred,
+    bool write_pred,
+    T init_val,
+    const nvfuser_index_t entrance_ind,
+    const nvfuser_index_t n_entrances) {
```
jjsjann123, line 794: nitpick: n_entrances is just not used at all here. Should we remove it?

More interestingly, n_entrances doesn't seem to be used in gridReduce neither... Is that obsolete code?

runtime/grid_reduction.cu
```diff
@@ -667,4 +667,243 @@ __device__ void serialReductionStep(
...
+    constexpr int total_bytes = vec_size * sizeof(T);
+    constexpr int n_loads = total_bytes <= 16 ? 1 : total_bytes / 16;
+    constexpr int n_elements = total_bytes <= 16 ? vec_size : 16 / sizeof(T);
+#pragma unroll
+    for (int i = 0; i < n_loads; i++) {
+      loadLocalToGlobal<T, n_elements, true>(
+          &work_buf[work_buf_offset + i * n_elements],
+          &block_reduction_val[i * n_elements]);
+    }
+  }
```
jjsjann123, line 864: I notice that this kernel is slightly different from gridReduction:

We are not updating work_buf here. `thread_offset` computed in this function is also different from the one in the last block. For my own curiosity, wondering what's the reason behind the change?

runtime/grid_reduction.cu
jjsjann123: fyi, wea re also not using `entrance_ind` here at all...

test/test_gpu_outer_reduction.cpp
```diff
@@ -2205,4 +2205,104 @@ TEST_F(OuterReductionTest, IterGroupedBlockReduction) {
...
+    auto cg_outputs = executor_cache.runFusionWithInputs(inputs);
+    auto ref = t0.to(at::kFloat).sum({0});
+    testValidate(
+        executor_cache.fusion(), cg_outputs, inputs, {ref}, __LINE__, __FILE__);
+  };
+
+  maybeClearAllocator(0);
+  for (int dim0 = 1024; dim0 <= 32768; dim0 *= 2) {
+    for (int dim1 = 1024; dim1 <= 32768; dim1 *= 2) {
+      test(dim0, dim1);
```
jjsjann123, line 2303: hmmm. this is a lot of tests... I'm all in for tests, having said that, should we worry about exploding our test coverage? I'm wondering if we should be a bit picky on what we put in as `functional` test here.

Not suggesting that we need to change anything here for now.

jjsjann123 (commented):
A few minor nitpick, Let's get some comment on the runtime file~~

csrc/scheduler/reduction.cpp
```diff
@@ -835,7 +835,7 @@ std::shared_ptr<ReductionParams> outerReductionHeuristic(
     debug() << "\n===== Reduction Stats ========\n"
             << "total_reduction_numel: " << total_reduction_numel << "\n"
             << "total_iteration_numel: " << total_iteration_numel << "\n"
-            << "vectorize_factor: " << vectorize_factor << "\n"
+            << "vectorize_factor: " << iter_unroll_factor << "\n"
```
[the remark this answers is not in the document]
liqiangxl: Here `vectorize_factor` is just the input to this function, it is actually the maximum allowed vectorization factor based on inner most dim. We should change it and we did the same thing in all the reduction/norm schedulers.

runtime/grid_reduction.cu
liqiangxl: Nice catch! I removed `entrance_ind` and `n_entrances` in grouped version. 
For the original gridReduce, I'll do another PR to clean `n_entrances`.

runtime/grid_reduction.cu
liqiangxl: I changed how the block reduction result is written to gmem. It ensures contiguous vectorized access.
`work_buf` is the base pointer, `work_buf_offset` is different for each thread to ensure we are writing to the correct gmem. I added additional comments.

liqiangxl: For `thread_offset`, added comments:
```  
// Its calculation is same to that in [iterGroupedGridReduce]. Becuase when
  // [iterGroupedGridReduceLastBlock] is called from [iterGroupedGridReduce],
  // X_THREAD, Y_THREAD, Z_THREAD are flipped.
``
We can also make it work without doing the flip in the function call but I didn't have a good reason do that change.

File: test/test_gpu_outer_reduction.cpp (hunk 8 above)
Comment (liqiangxl) [reply test/test_gpu_outer_reduction.cpp:2303]: This test is for validation purpose. Ideally we should test all the possible values but that seems impossible.
Looks like a lot but not very much. Only 36 cases.  {1,2,4,8,16,32} x {1,2,4,8,16,32}.

File: runtime/grid_reduction.cu (hunk 5 above)
Comment (liqiangxl) [reply runtime/grid_reduction.cu:771]: added

File: csrc/scheduler/reduction.cpp [hunk 10]
```diff
@@ -1210,12 +1210,15 @@ void scheduleReduction(Fusion* fusion, const ReductionParams& rparams) {
...
       rparams.vectorize_inner_reduction || rparams.vectorize_iter_dom;
 
-  // allow iter domain grouped reduction for block outer reduction.
+  // allow iter domain grouped reduction for block and grid outer reductions.
   // TODO: the var name is confusing, should rename
   // [cross_grid/block_inner_reduction] to [cross_grid/block_reduction], see
   // https://github.com/NVIDIA/Fuser/issues/1863
-  const bool use_iter_grouped_reduction = !rparams.cross_grid_inner_reduction &&
-      !rparams.fastest_dim && rparams.cross_block_inner_reduction;
+  const bool has_welford = ir_utils::hasOpsOfType<WelfordOp>(fusion);
```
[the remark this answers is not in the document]
liqiangxl: `welford` is a special reduction, grouped welford is only enabled for grid persistent.
Before this PR, we allow block reduce to `use_iter_grouped_reduction`, it doesn't cause error because we don't have a vectorized block welford outer reduction in our tests.
In this PR we want to extend to `grid reduce`, so needs to do special check for welford.

liqiangxl:
!build

liqiangxl:
!build

csrc/scheduler/reduction.cpp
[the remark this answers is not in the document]
jjsjann123: Does that mean we need to change the same thing on `innerReductionHeuristics`, which is still printing the old `vectorize_factor`.

We can do that in a separate PR as clean up.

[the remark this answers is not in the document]
liqiangxl: Yes. we should revise these debug print in all reduction/normalization schedulers. `vectorize_factor` is helpful it shows the maximum allowed factor but at the same time we also want to know the actual used factor.
I agree we should do this clean/refactor in a separate PR.

csrc/scheduler/reduction.cpp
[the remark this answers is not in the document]
jjsjann123: Thanks, I understand it as this is an intended behavior change.

runtime/grid_reduction.cu
jjsjann123: Ha I didn't realized that we flipped the flags. No worries.

I think the old logic is that, if we increment `work_buf` in the original function with `thread_offset`, then we don't have to compute it again in `last block`. But that's probably nitpicking too much in the code. We can revisit this when we merge the two functions.

jjsjann123 (approved):
LGTM.

There's some clean up that I would want. but let's do that when merging the block/nonblock runtime helper.

PR #1869, opened 2024-03-03
Fix ReplayTransformations to properly set isRFactorProduct. 

### LINKED ISSUE #1857: ReplayTransformations don't set isRFactorProduct correctly.
E.g. https://github.com/NVIDIA/Fuser/blob/71d9b555b401655c0d664b94e8abb7021ce29637/csrc/transform_iter.cpp#L53-L54 uses the default `rfactor_domain=false`. I guess this has been fine for now because we only use it to replay leaf transformations. However, this caused my pending https://github.com/NVIDIA/Fuser/pull/1846 to fail when replaying a pointwise unary op. 

I added a reproducer in #1846. The error message is
```
C++ exception with description "id_a->definition() == nullptr || id_a->isRFactorProduct() INTERNAL ASSERT FAILED at "/opt/pytorch/nvfuser/csrc/root_domain_map.cpp":546, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. Non-root domain is not supported: iS37{2}
```

Apparently, when scheduling a pointwise kernel, ComputeAtMap relies on isRFactorProduct to skip mapping certain IterDomains.

wujingyue:
This PR adds a knob to copy isRFactorProduct from the IterDomain being replayed. I hate knobs but couldn't find a better way. Plenty of ReplayTransformations users (e.g. https://github.com/NVIDIA/Fuser/blob/a684e2a220ed6daa14673d4c492b8eb80c299ce0/csrc/scheduler/utils.cpp#L2171 triggered by https://github.com/NVIDIA/Fuser/blob/8226e61a2a843e8e4ad8c6c2801b019f654828d1/test/test_gpu3.cpp#L7320) replay rfactor transforms on a TensorView without rfactor. Therefore, ReplayTransformations cannot copy isRFactorProduct unconditionally. 

Fixes #1857.

### DIFF
--- a/csrc/transform_iter.cpp
+++ b/csrc/transform_iter.cpp
@@ -57,7 +57,7 @@ void ReplayTransformations::handle(Split* s) {
       s->innerSplit(),
       s->startOffset(),
       s->stopOffset(),
-      s->outer()->isRFactorProduct());
+      replay_rfactor_ && s->outer()->isRFactorProduct());
   // Remove mapped from the leaf IDs
   leaf_ids_.erase(mapped);
 
@@ -128,7 +128,9 @@ void ReplayTransformations::handle(Merge* m) {
 
   // Replay the merge operation
   auto out = IterDomain::merge(
-      id_outer_mapped, id_inner_mapped, m->out()->isRFactorProduct());
+      id_outer_mapped,
+      id_inner_mapped,
+      replay_rfactor_ && m->out()->isRFactorProduct());
 
   // Remove inputs from the leaf IDs
   leaf_ids_.erase(id_outer_mapped);
@@ -172,6 +174,7 @@ void ReplayTransformations::handle(Swizzle* swizzle) {
 
   if (replay_swizzle_) {
     // Replay the swizzle onto mapped
+    // FIXME: isRFactorProduct
     outs = IterDomain::swizzle(swizzle->swizzleType(), mapped_x, mapped_y);
 
     // Remove mapped from the leaf IDs
@@ -219,6 +222,7 @@ void ReplayTransformations::handle(Swizzle2D* swizzle_2d) {
 
   if (replay_swizzle_) {
     // Replay the swizzle onto mapped
+    // FIXME: isRFactorProduct
     outs = IterDomain::swizzle(swizzle_2d->swizzleType(), mapped_x, mapped_y);
 
     // Remove mapped from the leaf IDs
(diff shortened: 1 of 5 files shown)


wujingyue:
!build

csrc/transform_iter.cpp
```diff
@@ -50,8 +50,14 @@ void ReplayTransformations::handle(Split* s) {
...
   // Replay the split onto mapped
+  NVF_ERROR(s->outer()->isRFactorProduct() == s->inner()->isRFactorProduct());
   auto outs = IterDomain::split(
-      mapped, s->factor(), s->innerSplit(), s->startOffset(), s->stopOffset());
+      mapped,
+      s->factor(),
+      s->innerSplit(),
+      s->startOffset(),
+      s->stopOffset(),
+      replay_rfactor_ && s->outer()->isRFactorProduct());
```
naoyam, line 60: Suppose `mapped` is an rfactor product, this would result in an non-rfactor product output if `replay_rfactor_` is false. Should it be considered an error?

wujingyue: > Suppose mapped is an rfactor product, this would result in an non-rfactor product output if replay_rfactor_ is false.

That's right. 

> Should it be considered an error?

That's already the current behavior and I'm not making it worse at least :)

AFAICT, it's not necessarily an error. Users of ReplayTransformations may want to replay rfactor-to-leaf transformations on a new rfactor domain. The results of those transformations shouldn't be marked `rf`. Correct?

wujingyue: Below is a real scenario that happens ToT. It's not for split per say, but for merge, which I think is close enough to show the idea. It can be reproduced with `bin/nvfuser_tests --gtest_filter=*ForIncompliantLayout`. 

The pointwise scheduler tries to replay the consumer's root-to-leaf transformations on the producer's rfactor domain. Dimension 0 of the producer's rfactor doesn't have `rf` but dimension 1 has `rf`. 

image

csrc/transform_iter.cpp
```diff
@@ -253,7 +262,7 @@ void ReplayTransformations::handle(Resize* exp) {
         mapped,
         exp->leftExpand(),
         exp->rightExpand(),
-        mapped->isRFactorProduct());
```
naoyam, line 256: I suspect this should have been `false` always. It would have resulted in an `rfactor` domain outside of the root-rfactor transformation.

naoyam (approved):
LGTM.

csrc/transform_iter.cpp
wujingyue: Agreed. I did try to test this line of change alone and found all tests passed. Probably a lack of test coverage

wujingyue:
!build

PR #1874, opened 2024-03-06
Evaluate `MmaOp + Cast` in a single call

Priya2698:
Since PyTorch matmul maintains the input data type, we expect to see fusions of type `MmaOp + Cast`. To avoid roundtrip casts, we look ahead and skip casting the ATen output to float whenever MmaOp is followed by cast.

Nsys profiles for the test: `MmaOpAndCast`:

Main branch:
```
Start (ns)  Duration (ns)  CorrId  GrdX  GrdY  GrdZ  BlkX  BlkY  BlkZ  Reg/Trd  StcSMem (MB)  DymSMem (MB)  Bytes (MB)  Throughput (MB/s)  SrcMemKd  DstMemKd            Device             Ctx  GreenCtx  Strm                                                  Name                                                
 ----------  -------------  ------  ----  ----  ----  ----  ----  ----  -------  ------------  ------------  ----------  -----------------  --------  --------  ---------------------------  ---  --------  ----  ----------------------------------------------------------------------------------------------------
  586509701            640     136                                                                                0.000             18.750  Pageable  Device    NVIDIA GeForce RTX 4090 (0)    1               7  [CUDA memcpy Host-to-Device]                                                                        
  586557445            353     150                                                                                0.000             67.989  Pageable  Device    NVIDIA GeForce RTX 4090 (0)    1               7  [CUDA memcpy Host-to-Device]                                                                        
  586584709            384     164                                                                                0.000             41.667  Pageable  Device    NVIDIA GeForce RTX 4090 (0)    1               7  [CUDA memcpy Host-to-Device]                                                                        
  688414636           2272    1631     1     1     1   128     1     1       57         0.000         0.005                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void cutlass::Kernel2<cutlass_75_wmma_tensorop_f16_s161616gemm_f16_32x32_32x1_nn_align1>(T1::Params)
  702497302           2048    1644     1     1     1   128     1     1       32         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::unrolled_elementwise_kernel<at::native::direct_copy_kernel_cuda(at::TensorIterator…
  702586486           1248    1658     1     1     1   128     1     1       32         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::unrolled_elementwise_kernel<at::native::direct_copy_kernel_cuda(at::TensorIterator…
  713771070           1184    1671     1     1     1   128     1     1       22         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::BinaryFunctor<c10::Half, c10::Ha…
  718786401           1056    1681     1     1     1   128     1     1       17         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::AUnaryFunctor<c10::Half, c10::Ha…
  720615906           1120    1701     1     1     1   128     1     1       18         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::AbsFunctor<c10::Half>, at::detai…
  728717384           1152    1711     1     1     1   128     1     1       20         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::CUDAFunctorOnSelf_add<c10::Half>…
  728761672           1056    1721     1     1     1   128     1     1       22         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::CUDAFunctor_add<c10::Half>, at::…
  728775048           1024    1741     1     1     1   128     1     1       18         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::AbsFunctor<c10::Half>, at::detai…
  728794536            992    1761     1     1     1   128     1     1       18         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::AbsFunctor<c10::Half>, at::detai…
  728834440           1056    1771     1     1     1   128     1     1       19         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::AUnaryFunctor<c10::Half, c10::Ha…
  728844232           1088    1781     1     1     1   128     1     1       22         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::BinaryFunctor<c10::Half, c10::Ha…
  728866184           1216    1791     1     1     1   128     1     1       20         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::BinaryFunctor<bool, bool, bool, …
  733362859           1312    1801     1     1     1   128     1     1       22         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::<unnamed>::CompareFunctor<c10::H…
  740977616           1216    1807     1     1     1   128     1     1       20         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::BinaryFunctor<bool, bool, bool, …
  741023184           1088    1813     1     1     1   128     1     1       23         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::BinaryFunctor<bool, bool, bool, …
  748730166           2112    1827     1     1     1     8     1     1       32         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::reduce_kernel<(int)512, (int)1, at::native::ReduceOp<bool, at::native::func_wrappe…
  749799223           1727    1839                                                                                0.000              0.579  Device    Pinned    NVIDIA GeForce RTX 4090 (0)    1               7  [CUDA memcpy Device-to-Host]                                                                        
```

This branch:
```
Start (ns)  Duration (ns)  CorrId  GrdX  GrdY  GrdZ  BlkX  BlkY  BlkZ  Reg/Trd  StcSMem (MB)  DymSMem (MB)  Bytes (MB)  Throughput (MB/s)  SrcMemKd  DstMemKd            Device             Ctx  GreenCtx  Strm                                                  Name                                                
 ----------  -------------  ------  ----  ----  ----  ----  ----  ----  -------  ------------  ------------  ----------  -----------------  --------  --------  ---------------------------  ---  --------  ----  ----------------------------------------------------------------------------------------------------
  573374330            769     136                                                                                0.000             15.605  Pageable  Device    NVIDIA GeForce RTX 4090 (0)    1               7  [CUDA memcpy Host-to-Device]                                                                        
  573429690            353     150                                                                                0.000             67.989  Pageable  Device    NVIDIA GeForce RTX 4090 (0)    1               7  [CUDA memcpy Host-to-Device]                                                                        
  573457787            352     164                                                                                0.000             45.455  Pageable  Device    NVIDIA GeForce RTX 4090 (0)    1               7  [CUDA memcpy Host-to-Device]                                                                        
  676276859           2272    1631     1     1     1   128     1     1       57         0.000         0.005                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void cutlass::Kernel2<cutlass_75_wmma_tensorop_f16_s161616gemm_f16_32x32_32x1_nn_align1>(T1::Params)
  691458602           1856    1643     1     1     1   128     1     1       22         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::BinaryFunctor<c10::Half, c10::Ha…
  696694638           1057    1653     1     1     1   128     1     1       17         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::AUnaryFunctor<c10::Half, c10::Ha…
  698517552           1120    1673     1     1     1   128     1     1       18         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::AbsFunctor<c10::Half>, at::detai…
  706948024           1152    1683     1     1     1   128     1     1       20         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::CUDAFunctorOnSelf_add<c10::Half>…
  707000024           1024    1693     1     1     1   128     1     1       22         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::CUDAFunctor_add<c10::Half>, at::…
  707017464           1056    1713     1     1     1   128     1     1       18         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::AbsFunctor<c10::Half>, at::detai…
  707037624           1024    1733     1     1     1   128     1     1       18         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::AbsFunctor<c10::Half>, at::detai…
  707088408           1088    1743     1     1     1   128     1     1       19         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::AUnaryFunctor<c10::Half, c10::Ha…
  707100248           1088    1753     1     1     1   128     1     1       22         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::BinaryFunctor<c10::Half, c10::Ha…
  707120568           1248    1763     1     1     1   128     1     1       20         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::BinaryFunctor<bool, bool, bool, …
  711613949           1312    1773     1     1     1   128     1     1       22         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::<unnamed>::CompareFunctor<c10::H…
  719774596           1280    1779     1     1     1   128     1     1       20         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::BinaryFunctor<bool, bool, bool, …
  719823204           1184    1785     1     1     1   128     1     1       23         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::BinaryFunctor<bool, bool, bool, …
  730395694           2464    1799     1     1     1     8     1     1       32         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::reduce_kernel<(int)512, (int)1, at::native::ReduceOp<bool, at::native::func_wrappe…
  731443855           1696    1811                                                                                0.000              0.590  Device    Pinned    NVIDIA GeForce RTX 4090 (0)    1               7  [CUDA memcpy Device-to-Host]                                                                        
```

For reference, the nsys profile for the MmaOp alone (SingleMmaOp test):
```
Start (ns)  Duration (ns)  CorrId  GrdX  GrdY  GrdZ  BlkX  BlkY  BlkZ  Reg/Trd  StcSMem (MB)  DymSMem (MB)  Bytes (MB)  Throughput (MB/s)  SrcMemKd  DstMemKd            Device             Ctx  GreenCtx  Strm                                                  Name                                                
 ----------  -------------  ------  ----  ----  ----  ----  ----  ----  -------  ------------  ------------  ----------  -----------------  --------  --------  ---------------------------  ---  --------  ----  ----------------------------------------------------------------------------------------------------
  583502109            672     136                                                                                0.000             17.857  Pageable  Device    NVIDIA GeForce RTX 4090 (0)    1               7  [CUDA memcpy Host-to-Device]                                                                        
  583549533            352     150                                                                                0.000             68.182  Pageable  Device    NVIDIA GeForce RTX 4090 (0)    1               7  [CUDA memcpy Host-to-Device]                                                                        
  583577949            352     164                                                                                0.000             90.909  Pageable  Device    NVIDIA GeForce RTX 4090 (0)    1               7  [CUDA memcpy Host-to-Device]                                                                        
  684288833           2272    1631     1     1     1   128     1     1       57         0.000         0.005                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void cutlass::Kernel2<cutlass_75_wmma_tensorop_f16_s161616gemm_f16_32x32_32x1_nn_align1>(T1::Params)
  698291457           2112    1644     1     1     1   128     1     1       32         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::unrolled_elementwise_kernel<at::native::direct_copy_kernel_cuda(at::TensorIterator…
  708948825           1056    1657     1     1     1   128     1     1       24         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::BinaryFunctor<float, float, bool…
  714311589           1024    1667     1     1     1   128     1     1       16         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::AUnaryFunctor<float, float, floa…
  716180425           1024    1687     1     1     1   128     1     1       16         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::AbsFunctor<float>, at::detail::A…
  724816349           1024    1697     1     1     1   128     1     1       16         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::CUDAFunctorOnSelf_add<float>, at…
  724860989           1024    1707     1     1     1   128     1     1       22         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::CUDAFunctor_add<float>, at::deta…
  724875549           1024    1727     1     1     1   128     1     1       16         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::AbsFunctor<float>, at::detail::A…
  724895965           1024    1747     1     1     1   128     1     1       16         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::AbsFunctor<float>, at::detail::A…
  724935133           1024    1757     1     1     1   128     1     1       16         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::AUnaryFunctor<float, float, bool…
  724945885           1056    1767     1     1     1   128     1     1       24         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::BinaryFunctor<float, float, bool…
  724966845           1248    1777     1     1     1   128     1     1       20         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::BinaryFunctor<bool, bool, bool, …
  729533799           1248    1787     1     1     1   128     1     1       22         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::<unnamed>::CompareFunctor<float>…
  738022394           1185    1793     1     1     1   128     1     1       20         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::BinaryFunctor<bool, bool, bool, …
  738068187           1088    1799     1     1     1   128     1     1       23         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::BinaryFunctor<bool, bool, bool, …
  745858252           2144    1813     1     1     1     8     1     1       32         0.000         0.000                                                     NVIDIA GeForce RTX 4090 (0)    1               7  void at::native::reduce_kernel<(int)512, (int)1, at::native::ReduceOp<bool, at::native::func_wrappe…
  746932079           1536    1825                                                                                0.000              0.651  Device    Pinned    NVIDIA GeForce RTX 4090 (0)    1               7  [CUDA memcpy Device-to-Host]                                                                        
```

The two elementwise kernels following the CUTLASS matmul kernel are no longer present with this update in the test `MmaOpAndCast`.

Issue #1669

### DIFF
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -2105,6 +2105,30 @@ std::vector<PolymorphicValue> MmaOp::evaluate(
       tv_b->getRootDomain().front()->isBroadcast(),
       "Expected first dimension to be broadcasted for second operand.");
 
+  // ATen preserves the dtype of MmaOp inputs whereas MmaOp generates float
+  // outputs. To preserve numerical equivalence and precision, the output of
+  // ATen matmul should be the same as MmaOp out `eventually`. Supported cases:
+  //  1. MmaOp->out() and MmaOp->input() are the same dtype.
+  //  2. MmaOp->out() is followed by a CastOp() to the MmaOp->input() dtype.
+  // NOTE: Currently MmaOp only accepts Half and BFloat16 so case (1) will not
+  // occur.
+
+  const std::vector<Expr*>& uses = out()->uses();
+  // Check if the only use of MmaOp is a single castOp.
+  bool is_use_single_cast = [&uses] {
+    return (
+        uses.size() == 1 && uses.front()->isA<UnaryOp>() &&
+        uses.front()->as<UnaryOp>()->getUnaryOpType() == UnaryOpType::Cast);
+  }();
+
+  // Check if we eventually convert to the ATen output dtype.
+  // See https://github.com/NVIDIA/Fuser/pull/1874#discussion_r1516991574
+  NVF_CHECK(
+      tv_a->getDataType() == out()->getDataType() ||
+      (is_use_single_cast &&
+       tv_a->getDataType() ==
+           uses.front()->as<UnaryOp>()->out()->getDataType()));
+
   // Squeeze the inputs to remove the broadcasted dimensions.
   const auto in_a = inputs.at(0).as<at::Tensor>().squeeze(-1);
   const auto in_b = inputs.at(1).as<at::Tensor>().squeeze(0);
@@ -2113,34 +2137,7 @@ std::vector<PolymorphicValue> MmaOp::evaluate(
   // [M, K] x [K, N] compatible with aten::matmul format.
   at::Tensor output = in_a.matmul(in_b);
 
-  // ATen preserves the input dtype whereas MmaOP generates float outputs.
-  // If the MmaOp is followed by a cast, skip casting back the input to avoid
-  // roundtrip casts. For eg: MmaOp (H, H) + CastOp (F, H) will result in
-  // F->H->F->H impacting precision and performance. Note: This potentially
-  // binds the MmaOp output in the expression evaluator to a tensor
-  //        of different datatype.
-  const std::vector<Expr*> uses = out()->uses();
-
-  // Check if the only use of MmaOp is a single castOp.
-  bool is_use_single_cast = [&uses] {
-    return (
-        uses.size() == 1 && uses.front()->isA<UnaryOp>() &&
-        uses.front()->as<UnaryOp>()->getUnaryOpType() == UnaryOpType::Cast);
-  }();
-
-  if (is_use_single_cast) {
-    // We currently don't support evaluating Cast (MmaOp (H, H), Bfloat16) since
-    // this loses precision due to ATen matmul converting Float to Half.
-    // CastOp should eventually convert to the MmaOp out() or ATen output dtype.
-    auto cast_out_dtype = uses.front()->as<UnaryOp>()->out()->getDataType();
-    NVF_CHECK(
-        cast_out_dtype == tv_a->getDataType() ||
-        cast_out_dtype == out()->getDataType());
-    return {output};
-  }
-
-  // Convert to float in other cases.
-  return {output.to(data_type_to_aten(out()->getDataType().value()))};
+  return {output};
 }
 
 NVFUSER_DEFINE_CLONE_AND_CREATE(MmaOp)
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -2114,16 +2114,19 @@ std::vector<PolymorphicValue> MmaOp::evaluate(
   auto output = in_a.matmul(in_b);
 
   // ATen preserves the input dtype whereas MmaOP generates float outputs.
-  // Cast to the dtype of the MmaOp output for consistency.
-  // NOTE: MmaOp returns the float output, whereas in the evaluate method,
-  //      we are casting from float -> input_dtype -> float. This will lead
-  //      to loss of precision.
-  //      MmaOp::evaluate should be modified to effectively handle cast(MmaOp(H,
-  //      H), H) This will avoid the above cast chain and precision issue.
-  if (tv_a->getDataType() != out()->getDataType().value()) {
-    output = output.to(data_type_to_aten(out()->getDataType().value()));
+  // If the MmaOp is followed by a cast, skip casting back the input to avoid
+  // roundtrip casts. For eg: MmaOp (H, H) + CastOp (F, H) will result in
+  // F->H->F->H impacting precision and performance. Note: This potentially
+  // binds the MmaOp output in the expression evaluator to a tensor
+  //        of different datatype.
+  auto uses = out()->uses();
+  if ((tv_a->getDataType() == out()->getDataType().value()) ||
+      (uses.size() == 1 && uses.front()->isA<UnaryOp>() &&
+       uses.front()->as<UnaryOp>()->getUnaryOpType() == UnaryOpType::Cast)) {
+    return {output};
+  } else {
+    return {output.to(data_type_to_aten(out()->getDataType().value()))};
   }
-  return {output};
 }
 
 NVFUSER_DEFINE_CLONE_AND_CREATE(MmaOp)
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -2111,7 +2111,7 @@ std::vector<PolymorphicValue> MmaOp::evaluate(
 
   // After removing the broadcast dimensions, the format should be
   // [M, K] x [K, N] compatible with aten::matmul format.
-  auto output = in_a.matmul(in_b);
+  at::Tensor output = in_a.matmul(in_b);
 
   // ATen preserves the input dtype whereas MmaOP generates float outputs.
   // If the MmaOp is followed by a cast, skip casting back the input to avoid
@@ -2119,10 +2119,17 @@ std::vector<PolymorphicValue> MmaOp::evaluate(
   // F->H->F->H impacting precision and performance. Note: This potentially
   // binds the MmaOp output in the expression evaluator to a tensor
   //        of different datatype.
-  auto uses = out()->uses();
+  const std::vector<Expr*> uses = out()->uses();
+
+  // Check if the only use of MmaOp is a single castOp.
+  bool is_use_single_cast = [&uses] {
+    return (
+        uses.size() == 1 && uses.front()->isA<UnaryOp>() &&
+        uses.front()->as<UnaryOp>()->getUnaryOpType() == UnaryOpType::Cast);
+  }();
+
   if ((tv_a->getDataType() == out()->getDataType().value()) ||
-      (uses.size() == 1 && uses.front()->isA<UnaryOp>() &&
-       uses.front()->as<UnaryOp>()->getUnaryOpType() == UnaryOpType::Cast)) {
+      is_use_single_cast) {
     return {output};
   } else {
     return {output.to(data_type_to_aten(out()->getDataType().value()))};
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -2128,12 +2128,19 @@ std::vector<PolymorphicValue> MmaOp::evaluate(
         uses.front()->as<UnaryOp>()->getUnaryOpType() == UnaryOpType::Cast);
   }();
 
-  if ((tv_a->getDataType() == out()->getDataType().value()) ||
-      is_use_single_cast) {
+  if (is_use_single_cast) {
+    // We currently don't support evaluating Cast (MmaOp (H, H), Bfloat16) since
+    // this loses precision due to ATen matmul converting Float to Half.
+    // CastOp should eventually convert to the MmaOp out() or ATen output dtype.
+    auto cast_out_dtype = uses.front()->as<UnaryOp>()->out()->getDataType();
+    NVF_CHECK(
+        cast_out_dtype == tv_a->getDataType() ||
+        cast_out_dtype == out()->getDataType());
     return {output};
-  } else {
-    return {output.to(data_type_to_aten(out()->getDataType().value()))};
   }
+
+  // Convert to float in other cases.
+  return {output.to(data_type_to_aten(out()->getDataType().value()))};
 }
 
 NVFUSER_DEFINE_CLONE_AND_CREATE(MmaOp)
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -2113,21 +2113,21 @@ std::vector<PolymorphicValue> MmaOp::evaluate(
   // NOTE: Currently MmaOp only accepts Half and BFloat16 so case (1) will not
   // occur.
 
-  const std::vector<Expr*>& uses = out()->uses();
-  // Check if the only use of MmaOp is a single castOp.
-  bool is_use_single_cast = [&uses] {
-    return (
-        uses.size() == 1 && uses.front()->isA<UnaryOp>() &&
-        uses.front()->as<UnaryOp>()->getUnaryOpType() == UnaryOpType::Cast);
-  }();
+  auto used_as_dtype = [](Val* out) -> DataType {
+    const std::vector<Expr*>& uses = out->uses();
+    if (uses.size() == 1) {
+      if (auto* unary = dynamic_cast<UnaryOp*>(uses.front())) {
+        if (unary->getUnaryOpType() == UnaryOpType::Cast) {
+          return unary->out()->getDataType().value();
+        }
+      }
+    }
+    return out->getDataType().value();
+  };
 
   // Check if we eventually convert to the ATen output dtype.
   // See https://github.com/NVIDIA/Fuser/pull/1874#discussion_r1516991574
-  NVF_CHECK(
-      tv_a->getDataType() == out()->getDataType() ||
-      (is_use_single_cast &&
-       tv_a->getDataType() ==
-           uses.front()->as<UnaryOp>()->out()->getDataType()));
+  NVF_CHECK(used_as_dtype(out()) == tv_a->getDataType().value());
 
   // Squeeze the inputs to remove the broadcasted dimensions.
   const auto in_a = inputs.at(0).as<at::Tensor>().squeeze(-1);
--- a/test/test_matmul_aten_evaluation.cpp
+++ b/test/test_matmul_aten_evaluation.cpp
@@ -33,44 +33,6 @@ class MatmulATenEvaluationTest : public NVFuserTest {
       guard_;
 };
 
-TEST_F(MatmulATenEvaluationTest, SingleMmaOp) {
-  auto fusion = std::make_unique<Fusion>();
-  FusionGuard fg(fusion.get());
-
-  EnableOptionsGuard enable_guard;
-  EnableOptionsGuard::getCurOptions().set(EnableOption::MatmulExprEval);
-
-  int64_t m = 2, k = 3, n = 4;
-  std::vector<int64_t> a_shape{m, k}, b_shape{k, n}, out_shape{m, n};
-
-  auto tv0 = makeConcreteTensor(a_shape, DataType::Half);
-  auto tv1 = makeConcreteTensor(b_shape, DataType::Half);
-  auto tv0b = broadcast(tv0, {false, false, true}); // [M, K, 1]
-  auto tv1b = broadcast(tv1, {true, false, false}); // [1, K, N]
-  auto tv2 = fusedMultiplySum(tv0b, tv1b, {1});
-
-  fusion->addInput(tv0);
-  fusion->addInput(tv1);
-  fusion->addOutput(tv2);
-
-  at::Tensor t0 = at::ones(a_shape, at::kHalf).cuda();
-  at::Tensor t1 = at::ones(b_shape, at::kHalf).cuda();
-  at::Tensor out_ref = at::full(out_shape, k, at::kFloat).cuda();
-
-  FusionExecutorCache fec(std::move(fusion));
-  auto out = fec.runFusionWithInputs({t0, t1});
-
-  EXPECT_EQ(fec.getMostRecentKernelRuntime()->executors().size(), 1);
-
-  // Verify that the io_alias_ set has the correct entry
-  auto kernel = fec.getMostRecentKernelRuntime()->executors().at(0).kernel();
-  EXPECT_EQ(
-      kernel->getOutputAlias(kernel->outputs()[0]).type,
-      AllocationType::Evaluate);
-
-  EXPECT_TRUE(at::allclose(out[0], out_ref));
-}
-
 TEST_F(MatmulATenEvaluationTest, MmaOpAndCast) {
   auto fusion = std::make_unique<Fusion>();
   FusionGuard fg(fusion.get());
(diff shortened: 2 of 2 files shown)


Priya2698:
!build

csrc/ir/nodes.cpp
```diff
@@ -2114,16 +2114,19 @@ std::vector<PolymorphicValue> MmaOp::evaluate(
...
-  if (tv_a->getDataType() != out()->getDataType().value()) {
-    output = output.to(data_type_to_aten(out()->getDataType().value()));
-  }
-  return {output};
+  // If the MmaOp is followed by a cast, skip casting back the input to avoid
+  // roundtrip casts. For eg: MmaOp (H, H) + CastOp (F, H) will result in
+  // F->H->F->H impacting precision and performance. Note: This potentially
+  // binds the MmaOp output in the expression evaluator to a tensor
+  //        of different datatype.
+  auto uses = out()->uses();
```
wujingyue, line 2122:
```suggestion
  const std::vector<Expr*>& uses = out()->uses();
```

csrc/ir/nodes.cpp
```diff
@@ -2114,16 +2114,19 @@ std::vector<PolymorphicValue> MmaOp::evaluate(
   auto output = in_a.matmul(in_b);
```
wujingyue, line 2114:
```suggestion
  at::Tensor output = in_a.matmul(in_b);
```

csrc/ir/nodes.cpp
```diff
@@ -2114,16 +2114,19 @@ std::vector<PolymorphicValue> MmaOp::evaluate(
...
-  return {output};
+  // If the MmaOp is followed by a cast, skip casting back the input to avoid
+  // roundtrip casts. For eg: MmaOp (H, H) + CastOp (F, H) will result in
+  // F->H->F->H impacting precision and performance. Note: This potentially
+  // binds the MmaOp output in the expression evaluator to a tensor
+  //        of different datatype.
+  auto uses = out()->uses();
+  if ((tv_a->getDataType() == out()->getDataType().value()) ||
+      (uses.size() == 1 && uses.front()->isA<UnaryOp>() &&
+       uses.front()->as<UnaryOp>()->getUnaryOpType() == UnaryOpType::Cast)) {
```
wujingyue, line 2125: Is it necessary to also check that the single use casts to `tv_a->getDataType()`?

wujingyue, line 2125: There are a lot of compounded conditions here. Consider to put them into a lambda so we can early exit to make the conditions easier to understand.

Priya2698: I don't think it is. We can let the cast handle the final dtype.
The cases where we won't cast back to float are `half, bfloat16` for matmul. So even if the single use was casting to float instead of the tv_a dtype, it would work.

wujingyue: For example, `BFloat(MmaOp(Half,Half))` is not equivalent to `Bfloat(Half(MmaOp(Half,Half)))`, what `at::matmul` + `at::Tensor::to` would compute. Because `Half` has more mantissa bits and `Bfloat` has more exponent bits, the extra conversion to `Half` is going to lose precision or even worse clip large values to `inf`. 

What options do we have to precisely expression-evaluate `BFloat(MmaOp(Half,Half))`? Making it `mul+sum+cast` is the only option I can think of. If we don't want to include that option in this particular PR, I'd still make this function hard-fail on things like `BFloat(MmaOp(Half,Half))`. Because it's much easier to root-cause a hard-fail than a numeric mismatch.

wujingyue: If you agree with above, I'd simplify the code to

```
NVF_CHECK(out() is eventually used as output.dtype());
return output;
```

Priya2698: I considered that case, but didn't think it occured in practice. OOC, are there models/cases where we see this?
I would cast back to float in this case instead of erroring out, wdyt? So we skip the cast if and only if it is a single cast to the same dtype.

wujingyue: > OOC, are there models/cases where we see this?

No. FYI, for cases that haven't occurred, assuming similar code complexity, my general preference is
1. to generate functional and performant code 
2. to generate functional but slow code
3. to hard fail during compilation (earlier is better)
4. to generate code that hard fails (earlier is better)
5. to generate code that produces wrong numbers

> I would cast back to float in this case instead of erroring out, wdyt? 

Can you clarify? 

Do you mean `BFloat(MmaOp(Half,Half))` becomes `at::matmul + to(float)`? Wouldn't that cause the fusion to output a different dtype than torch? 

Or, do you mean `BFloat(MmaOp(Half,Half))` becomes `at::matmul + to(float) + to(bfloat)`? That IIUC would have the same precision loss problem as `Bfloat(Half(MmaOp(Half,Half)))`, because of the cast to `half` (done by `at::matmul`).

Priya2698: > Or, do you mean `BFloat(MmaOp(Half,Half))` becomes `at::matmul + to(float) + to(bfloat)`? That IIUC would have the same precision loss problem as `Bfloat(Half(MmaOp(Half,Half)))`, because of the cast to `half` (done by `at::matmul`).

You're right, we would run into the precision issue here, although we are no less precise than pytorch in this case. The cast to `float` is however unnecessary.

> What options do we have to precisely expression-evaluate BFloat(MmaOp(Half,Half))? Making it mul+sum+cast is the only option I can think of. 

Will this be done in the matmul scheduler?

>If we don't want to include that option in this particular PR, I'd still make this function hard-fail on things like BFloat(MmaOp(Half,Half)). Because it's much easier to root-cause a hard-fail than a numeric mismatch.

Will error out in this PR.

Priya2698: > If you agree with above, I'd simplify the code to
> 
> ```
> NVF_CHECK(out() is eventually used as output.dtype());
> return output;
> ```

What do you mean? If the cast is to the correct input dtype, then the output.dtype() is different than out().

wujingyue: By "eventually", I meant the dtype after the cast. For example, in `Half(MmaOp(H,H))`, MmaOp's out is used as a half after the cast, so it's safe to simply return output. (Of course, the single-use check is still necessary)

wujingyue: > Will this be done in the matmul scheduler?

No, I was thinking of doing it in MmaOp::evaluate. Actually, we might not need to decompose. Can you confirm `MmaOp(H,H)` is numerically equivalent to `at::matmul(Float(H),Float(H))`? 

(Since we are converging to let such cases error out, the above question is not so relevant any more)

Priya2698: > By "eventually", I meant the dtype after the cast. For example, in `Half(MmaOp(H,H))`, MmaOp's out is used as a half after the cast, so it's safe to simply return output. (Of course, the single-use check is still necessary)

We should also allow cast to `float`. In the current setting, if we see a cast to double, that will also error out. Is this a case we we want to support? 
IMO cast to double following MmaOp will be unlikely in practice so I am fine with not increasing the list of conditions on the cast output type. Wdyt?

wujingyue: > We should also allow cast to float

I don't think so. Casting to float or double should error out as well for the same reason. `Double(MmaOp(H,H))` is not equivalent to `at::matmul(H,H).double()`. `Float(MmaOp(H,H))` is not equivalent to `at::matmul(H,H).float()`.

csrc/ir/nodes.cpp
```diff
@@ -2111,19 +2111,36 @@ std::vector<PolymorphicValue> MmaOp::evaluate(
...
-  if (tv_a->getDataType() != out()->getDataType().value()) {
-    output = output.to(data_type_to_aten(out()->getDataType().value()));
-  }
-  return {output};
+  // If the MmaOp is followed by a cast, skip casting back the input to avoid
+  // roundtrip casts. For eg: MmaOp (H, H) + CastOp (F, H) will result in
+  // F->H->F->H impacting precision and performance. Note: This potentially
+  // binds the MmaOp output in the expression evaluator to a tensor
+  //        of different datatype.
+  const std::vector<Expr*> uses = out()->uses();
```
wujingyue, line 2122:
```suggestion
  const std::vector<Expr*>& uses = out()->uses();
```

Otherwise there's an extra copy.

csrc/ir/nodes.cpp
wujingyue: I think the confusion comes from that we use different definitions for "precision". Sorry, I should have caught that earlier. I'm not talking about difference being less than a pre-defined threshold. I'm talking about numeric equivalence where `a+b == b+a` but `float(half(a)) != a`. 

Numeric equivalence is important, because I saw and debugged too many cases where subtle numeric inequivalence between two implementations gets amplified by many layers of a network, leading to vastly different end results and hurting model accuracy. 

That being said, numeric equivalent alone doesn't guarantee bit exactness. Other things like accumulation order can be different between implementations, leading to different bitwise results. For example, a CPU implementation of a matmul likely adopts a different partition strategy than a GPU one. However, numeric equivalence rules out many other sources of differences and often makes the difference between implementations manageable (e.g. the auto threshold thing you worked on in testValidate). 

Now back to the practical stuff -- what we should do with the existing tests (SingleMmaOp and MatmulWithBias). I'd change SingleMmaOp to test mma+cast instead (and therefore probably change the test name as well). I'd disable MatmulWithBias until we have at::addmm hooked up with ExpressionEvaluator. PS: to disable the test, you can just change the name to `DISABLED_MatmulWithBias` and googletest does the magic to skip it and show it in the disabled test list.

Priya2698: Got it, thanks for the explanation. I see your point now.

csrc/ir/nodes.cpp
```diff
@@ -2105,24 +2105,38 @@ std::vector<PolymorphicValue> MmaOp::evaluate(
...
+  }();
+
+  // Check if we eventually convert to the ATen output dtype.
+  // See https://github.com/NVIDIA/Fuser/pull/1874#discussion_r1516991574
+  NVF_CHECK(
+      tv_a->getDataType() == out()->getDataType() ||
+      (is_use_single_cast &&
+       tv_a->getDataType() ==
+           uses.front()->as<UnaryOp>()->out()->getDataType()));
+
```
wujingyue, line 2131:
```suggestion
  auto used_as_dtype = [](Val* v) -> DataType {
    const std::vector<Expr*>& uses = v->uses();
    if (uses.size() == 1) {
      if (auto* unary = dynamic_cast<UnaryOp*>(uses.front())) {
        if (unary->getUnaryOpType() == UnaryOpType::Cast) {
          return unary->out()->getDataType();
        }
      }
    }
    return v->getDataType();
  };
  NVF_CHECK(used_as_dtype(out()) == tv_a->getDataType());
```

(The above is just to show the idea -- I didn't even try to compile it)

wujingyue (approved):

PR #1878, opened 2024-03-07
A Python API for Matmul

protonu:
We add a new Python API in Fuser, which will accept tensors of the shape (M,K), (N,K). Currently we will execute the Matmul using the Aten expression evaluator, but will move to Fuser's scheduling and lowering in the future.
This API will only allow tensors with the shape (M,K) (K,N), but the tensors may be strided, thus allowing different layouts.


csrc/ops/composite.cpp
```diff
@@ -106,6 +107,25 @@ TensorView* _matmul_tt(TensorView* a, TensorView* b) {
   return tv2;
 }
 
+TensorView* matmul(TensorView* a, TensorView* b) {
+  NVF_CHECK(a->nDims() == b->nDims());
+  NVF_CHECK(a->nDims() == 2 || a->nDims() == 3);
+  NVF_CHECK(a->getDataType().value() == b->getDataType().value());
+
+  // This flag being set means the matul is executed via ATen.
```
drzejan2, line 115: Nitpick:
```suggestion
  // This flag being set means the matmul is executed via ATen.
```

drzejan2 (commented):
Prepared changes look reasonable, but to be sure, another pair of eyes should check this PR.

Thanks!

csrc/ops/composite.cpp
```diff
@@ -106,6 +107,25 @@ TensorView* _matmul_tt(TensorView* a, TensorView* b) {
   return tv2;
 }
 
+TensorView* matmul(TensorView* a, TensorView* b) {
+  NVF_CHECK(a->nDims() == b->nDims());
+  NVF_CHECK(a->nDims() == 2 || a->nDims() == 3);
+  NVF_CHECK(a->getDataType().value() == b->getDataType().value());
```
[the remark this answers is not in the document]
rdspring1: You would add this opinfo definition to https://github.com/NVIDIA/Fuser/blob/main/python_tests/pytest_opinfos.py
```python
matmul_opinfo = OpInfo(
    lambda fd: fd.ops.matmul,
    "matmul",
    sample_input_generator=matmul_generator,
    error_input_generator=matmul_error_generator,
    reference=torch.matmul,
)
shape_ops.append(matmul_opinfo)
```
Then, define `matmul_generator` and `matmul_error_generator` in https://github.com/NVIDIA/Fuser/blob/main/python_tests/pytest_input_generators.py.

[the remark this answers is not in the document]
rdspring1: For `matmul_input_generator`, I usually take the sample input generator from pytorch.

https://github.com/pytorch/pytorch/blob/main/torch/testing/_internal/common_methods_invocations.py#L6490-L6515

csrc/ops/composite.cpp
```diff
@@ -106,6 +107,25 @@ TensorView* _matmul_tt(TensorView* a, TensorView* b) {
   return tv2;
 }
 
+TensorView* matmul(TensorView* a, TensorView* b) {
+  NVF_CHECK(a->nDims() == b->nDims());
+  NVF_CHECK(a->nDims() == 2 || a->nDims() == 3);
```
Priya2698, line 112: Are we supporting `a->nDims() == 3` in this API version? If not, we should remove that condition

python_tests/test_python_frontend.py
```diff
@@ -2408,6 +2408,27 @@ def fusion_func(fd: FusionDefinition, inps, matmul_fn) -> None:
                 fc = FusionCache.get()
                 fc.reset()
 
+    @unittest.skipUnless(
```
Priya2698, line 2411: ATen evaluation should be able to work with all architectures. Do we need this then?

protonu: I was in two minds about putting this here. Eventually we'll need this (once it goes through Fuser). I am happy remove it for now.

Priya2698: I prefer removing it for now, so that any errors on other architectures are captured properly when using ATen evaluation.

csrc/ops/composite.cpp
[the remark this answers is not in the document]
protonu: Thanks!  I think I got most of that in. I left the error checking out for now.

protonu:
!build

Priya2698 (approved):
LGTM.

csrc/ops/composite.cpp
```diff
@@ -106,6 +107,30 @@ TensorView* _matmul_tt(TensorView* a, TensorView* b) {
...
+      a->nDims() == b->nDims(),
+      "The number of dimension of A and B do not match");
+  // TODO: We'll need to suppor nDims == 3 for bmm.
+  NVF_CHECK(
+      a->nDims() == 2,
+      "Only 2-D Tensors are supported, in the future we'll support 3-D as well!");
+  NVF_CHECK(a->getDataType().value() == b->getDataType().value());
+
+  // This flag being set means the matmul is executed via ATen.
+  EnableOptionsGuard::getCurOptions().set(EnableOption::MatmulExprEval);
```
jjsjann123, line 121: This feels like a hack. Are we expecting people to use nvfuser matmul via python API to `always` execute it via aten?

python_tests/test_python_frontend.py
```diff
@@ -2408,6 +2408,25 @@ def fusion_func(fd: FusionDefinition, inps, matmul_fn) -> None:
...
+
+        def fusion_func(fd: FusionDefinition) -> None:
+            t0 = fd.from_pytorch(inputs[0])
+            t1 = fd.from_pytorch(inputs[1])
+            t2 = fd.ops.matmul(t0, t1)
+            t3 = fd.ops.cast(t2, dtype=DataType.Half)
+            fd.add_output(t3)
+
+        nvf_out, _ = self.exec_nvfuser(fusion_func, inputs)
+        # breakpoint()
```
jjsjann123, line 2425: what's breakpoint()?

python_tests/test_python_frontend.py
```diff
@@ -2408,6 +2408,25 @@ def fusion_func(fd: FusionDefinition, inps, matmul_fn) -> None:
                 fc = FusionCache.get()
                 fc.reset()
 
+    def test_unified_matmul(self):
```
jjsjann123, line 2411: what does unified_matmul mean?

I think this is test our aten execution? can we make it explicit in the test name as well.

python_tests/pytest_opinfos.py
```diff
@@ -1103,6 +1105,24 @@ def torch_reshape_sym_fn(input_tensor, output_shaped_tensor):
...
+matmul_ops = []
+
+
+def matmul_wrapper(fd, a, b):
+    t1 = fd.ops.matmul(a, b)
+    t2 = fd.ops.cast(t1, dtype=DataType.Half)
+    return t2
+
+
+matmul_opinfo = OpInfo(
```
jjsjann123, line 1117: This is not a general matmul test, regarding the output dtype cast. Let's name it properly for that as well.

csrc/ops/composite.cpp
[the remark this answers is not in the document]
jjsjann123: Would be nice to have a python API check.... But we don't have to push it for this PR.

python_tests/pytest_opinfos.py
protonu: I was I saw this was, this was the general matmul test. When we swap out the aten::matmul for Fuser's implementation, the user will not be aware of that. Please let me know if you think I have that wrong.

python_tests/test_python_frontend.py
protonu: thanks for catching that - will remove.

csrc/ops/composite.cpp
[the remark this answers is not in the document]
protonu: I'll put a TODO in the code for that. Thanks!

python_tests/pytest_input_generators.py
```diff
@@ -1478,3 +1478,25 @@ def vector_at_error_generator(
...
+):
+    make_arg = partial(
+        make_tensor,
+        dtype=torch.float16,
+        device="cuda",
+        low=None,
+        high=None,
+        requires_grad=requires_grad,
+    )
+    test_cases = (
```
rdspring1, line 1494: Just a suggestion. Feel free to leave it for a future PR. A cartesian product would give better coverage of matrix shapes without much manual effort.

```python
import itertools
import math

def multiply_range(maximum, step):
    assert maximum % step == 0
    num_steps = int(math.log(maximum, step))
    return tuple(map(pow, itertools.repeat(step, num_steps), range(1, num_steps + 1)))

# Ranges of tensor sizes: 8, 64, 512, 4096, 32768
# Use a Cartesian product to create a wide range of matrix shapes
M, N, K = itertools.repeat(multiply_range(32768, 8), 3)
for M, N, K in itertools.product(M, N, K):
    lhs_shape = (M, K)
    rhs_shape = (K, N)
    yield SampleInput(lhs, rhs)

**Review by jacobhinkle — COMMENTED** [review]
A few comments I thought I had posted last week...

File: csrc/ops/composite.cpp (hunk 3 above)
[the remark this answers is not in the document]
Comment (jacobhinkle) [reply csrc/ops/composite.cpp]: I see that you put a comment here indicating this will be supported in the future.

File: python_tests/test_python_frontend.py [hunk 10]
```diff
@@ -2408,6 +2408,27 @@ def fusion_func(fd: FusionDefinition, inps, matmul_fn) -> None:
                 fc = FusionCache.get()
                 fc.reset()
 
+    @unittest.skipUnless(
+        torch.cuda.get_device_properties(torch.cuda.current_device() == 8),
+        "Only the Ampere MMA Op is currently supported!",
+    )
+    def test_unified_matmul(self):
```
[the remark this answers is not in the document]
jacobhinkle: I see that bmm is not supported yet.

tests/python/pytest_opinfos.py
```diff
@@ -1103,6 +1105,24 @@ def torch_reshape_sym_fn(input_tensor, output_shaped_tensor):
 )
 tensor_creation_ops.append(uniform_opinfo)
 
+matmul_ops = []
+
+
+def matmul_wrapper(fd, a, b):
+    t1 = fd.ops.matmul(a, b)
+    t2 = fd.ops.cast(t1, dtype=DataType.Half)
```
Priya2698, line 1113: Should this also include bfloat16?

csrc/ops/composite.cpp
protonu: This should be resolved after this PR which made execution via ATen default.

python_tests/pytest_input_generators.py
protonu: Thanks, incorporated

tests/python/pytest_opinfos.py
protonu: Thanks, I added this, but the type deduction is clumsy. I'd need further PR(s) to unify the tests.

protonu:
!build

tests/python/pytest_opinfos.py
```diff
@@ -1103,6 +1105,33 @@ def torch_reshape_sym_fn(input_tensor, output_shaped_tensor):
 )
 tensor_creation_ops.append(uniform_opinfo)
 
+matmul_ops = []
+
+
+def matmul_wrapper(fd, data_type, a, b):
+    t1 = fd.ops.matmul(a, b)
+    t2 = fd.ops.cast(t1, dtype=DataType.Half)
```
jacobhinkle, line 1113: `data_type` is unused. I think you mean to replace `DataType.Half` with it`.

tests/python/pytest_opinfos.py
```diff
@@ -1103,6 +1105,33 @@ def torch_reshape_sym_fn(input_tensor, output_shaped_tensor):
...
+    return t2
+
+
+matmulfp16_opinfo = OpInfo(
+    lambda fd: partial(matmul_wrapper, fd, DataType.Half),
+    "matmul_float16",
+    dtypes=(torch.float16,),
```
jacobhinkle, line 1120: Are we able to access an `nvfuser.Tensor`'s dtype from python? If so then you could just add the dtypes in this list and then use `a.dtype()` instead of an explicit parameter right? This is a minor nitpick of course.

tests/python/pytest_input_generators.py
```diff
@@ -1478,3 +1478,31 @@ def vector_at_error_generator(
...
+
+    def multiply_range(maximum, step):
+        assert maximum % step == 0
+        num_steps = int(math.log(maximum, step))
+        return tuple(
+            map(pow, itertools.repeat(step, num_steps), range(1, num_steps + 1))
+        )
+
+    # Ranges of tensor sizes: 8, 64, 512, 4096, 32768
+    # Use a Cartesian product to create a wide range of matrix shapes
```
jacobhinkle, line 1503: Does this use the same `FusionDefinition` to run all the cases or will it be compiling 125 separate kernels?

tests/python/pytest_opinfos.py
protonu: Unfortunately, we can't access the dtype. I'd like to expose that, but I think that should be a separate PR.

tests/python/pytest_opinfos.py
protonu: Thanks for catching that!

tests/python/pytest_input_generators.py
protonu: It only compiles once (twice - once for fp16 and again for bf16).

protonu:
!build

tests/python/pytest_input_generators.py
rdspring1: We are creating a new `FusionDefinition` for each case, but they should map to the same `FusionExecutorCache` in the python frontend `FusionCache`.

https://github.com/NVIDIA/Fuser/blob/main/tests/python/pytest_ops.py#L47-L48

The total number of cuda kernels would depend on NvFuser's heuristics. e.g., Some kernels could be reused for some matrix shapes.

tests/python/pytest_opinfos.py
```diff
@@ -1103,6 +1105,33 @@ def torch_reshape_sym_fn(input_tensor, output_shaped_tensor):
 )
 tensor_creation_ops.append(uniform_opinfo)
 
+matmul_ops = []
+
+
+def matmul_wrapper(fd, data_type, a, b):
```
rdspring1, line 1111: We should have a quick comment indicating why we need separate OpInfo definitions for `fp16` and `bf16`.

rdspring1 (approved):
LGTM. Thanks for adding the OpInfo tests!

tests/python/pytest_input_generators.py
protonu: Thanks! My answer was based on debugging via prints. :)

protonu:
!build

protonu:
!build

PR #1879, opened 2024-03-07
Simplify some tests in test_id_model.cpp.

wujingyue:


wujingyue:
!build

test/test_id_model.cpp
```diff
@@ -938,6 +938,16 @@ TEST_F(IdModelTest, LoopPromotion8) {
   }
 }
 
+namespace {
+bool iterDomainsAreMapped(
+    const IdModel& id_model,
```
naoyam, line 943: nit: since there are multiple graphs in an `IdModel`, it might make more sense to receive a graph as a parameter rather than a `IdModel`.

naoyam (approved):
LGTM

test/test_id_model.cpp
wujingyue: Sorry, I missed this comment before I merged. Let's do that when we need this helper function to analyze other IdGraphs.

PR #1886, opened 2024-03-07
Update known values correctly

Priya2698:
This fixes a bug which would cause the intermediate values to be updated to `known_values`.


zasdfgbnm (approved):
Thanks for fixing!

PR #1889, opened 2024-03-08
Disable castInputOutputToLowerPrecision by default. 

### LINKED ISSUE #1742: nvFuser produces (slightly) wrong numbers in test_rope.py.
I was able to work around by increasing the threshold, but I think this is still worth attention. I believe there's a real bug rather than a floating-point representation error. 

To reproduce, run the following
```
$ git fetch origin wjy/mismatch
$ git checkout wjy/mismatch
$ pytest python_benchmarks/test_rope.py -k with_cat -s
...
>           torch.testing.assert_close(nvf_out, [ref_out], atol=0, rtol=0)
E           AssertionError: Tensor-likes are not equal!
E
E           Mismatched elements: 121664043 / 536870912 (22.7%)
E           Greatest absolute difference: 0.03125 at index (0, 0, 2444, 88)
E           Greatest relative difference: inf at index (0, 0, 122, 124)
E
E           The failure occurred for item [0]
```

`atol=0, rtol=0` may appear to be too strict but it's really not. `out = q * cos + q_rotated * sin` is the only reduction-ish op. Since it's adding two numbers piecewise, I expect a bit-exact match between the nvFuser output and the eager output. 

Now comes the most interesting part. If I remove the cast to bfloat at the end of the fusion and [that in the reference implementation](https://github.com/NVIDIA/Fuser/blob/wjy/mismatch/python_benchmarks/test_rope.py#L174), the two outputs match bit-exact. So I went to print the element at (0,0,122,124) before the cast and after the cast. 

```
python_benchmarks/test_rope.py before cast: 0
after cast in ref: 0
after cast in nvf: -0.0004119873046875
```

The eager output clearly makes more sense: `float2bfloat(0)` should be 0 not -0.0004119873046875. 

My first suspicion was the cast op. However, a smaller fusion I wrote just doing the cast failed to reproduce the problem.

### LINKED ISSUE #1872: INTERNAL ASSERT with funny casting
apparently, nvfuser does not like the following program, I'm not sure it should be invalid:
```
import torch

def nvfuser_fusion_id7(fd : FusionDefinition) -> None :
    S0 = fd.define_scalar(1.00000, dtype=DataType.Double)
    S1 = fd.define_scalar(5, dtype=DataType.Int)
    V2 = fd.define_vector([S1], dtype=DataType.Int)
    T3 = fd.ops.full(shape=V2, fill_value=S0, dtype=DataType.Float)
    T4 = fd.ops.slice(T3, start_indices=[0], end_indices=[2], strides=[1])
    T5 = fd.ops.cast(T4, dtype=DataType.BFloat16)
    T6 = fd.ops.slice(T3, start_indices=[2], end_indices=[5], strides=[1])
    T7 = fd.ops.cast(T6, dtype=DataType.Half)
    fd.add_output(T5)
    fd.add_output(T7)

    nvfuser_fusion_id7(fd)

inputs = [
]
fd.execute(inputs)

```

```
RuntimeError: h.has_value() INTERNAL ASSERT FAILED at "/home/tv/data/firma/grid/thunder/Fuser/csrc/fusion_segmenter.cpp":3640, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. Can not find a scheduler to schedule fusion segment
```

wujingyue:
At this very moment, `castInputOutputToLowerPrecision` seems to bring more problems than benefits. So I think we'd better turn it off by default. It has caused precision loss (#1742) and program crash (#1872). It was added for performance, but the current framework benchmarks are neutral to disabling it (the results are put in `1889_benchmark_results.txt` in the nvFuser team's folder). 

Properly fixing `castInputOutputToLowerPrecision` will take some time. To properly fix #1742, it'll have to be made more conservative, i.e., adding a cast only when it preserves numeric equivalence. To properly fix #1872, it'll have to stop adding a cast before a resizing op. 

Because this PR disables `castInputOutputToLowerPrecision` by default, it has to add option guards to some unit tests that exercise the function. 

Fixes #1742.

Fixes #1872.


wujingyue:
!build

wujingyue:
!build

test/test_segmentation.cpp
```diff
@@ -329,4 +329,208 @@ TEST_F(SegmentationTest, ForwardedExprsAreReplicated) {
   testValidate(fec.fusion(), out_tensors, {in_tensor}, __LINE__, __FILE__);
 }
 
+TEST_F(SegmentationTest, ForceFp16Simple) {
+  EnableOptionsGuard opt_guard;
```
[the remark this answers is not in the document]
wujingyue: To reset the flag when the test completes.

test/test_segmentation.cpp
```diff
@@ -329,4 +329,208 @@ TEST_F(SegmentationTest, ForwardedExprsAreReplicated) {
...
+  EnableOptionsGuard::getCurOptions().set(EnableOption::IoToLowerPrecision);
```
jjsjann123, line 334: wondering if there is style preference here? static method goes through class scope rather than `opt_guard.getCurOptions()`?

jjsjann123 (approved):
LGTM on code change.

I'm wondering if we have concern over perf regression here. Should we make the option an opt-out instead?

test/test_segmentation.cpp
wujingyue: > static method goes through class scope rather than opt_guard.getCurOptions()?

That seems to be the right thing to do. Otherwise, readers may mistaken it as an instance method.

wujingyue:
> I'm wondering if we have concern over perf regression here. 

The framework benchmarks are neutral to this change, so I don't worry too much about perf at this moment. 

> Should we make the option an opt-out instead?

Let users opt out from `castInputOutputToLowerPrecision` when they know the error is due to that? Even for nvFuser developers, it's pretty hard to debug "can not find a scheduler to schedule fusion segment" as in #1872 and to debug a mismatch in numbers as in #1742.

jjsjann123:
That's fair. if there's no framework regression, there's no good reason to enable this.

wujingyue:
!build

wujingyue:
!build

wujingyue:
!build

PR #1891, opened 2024-03-08
Propagation rule reduction

jjsjann123:


jjsjann123:
Marking failing ci tests for myself
```
[  FAILED  ] 5 tests, listed below:
[  FAILED  ] NVFuserTest.FusionMagicSchedulerInstanceNormalizationBackward_CUDA
[  FAILED  ] NVFuserTest.FusionBNBackwardRepro_CUDA
[  FAILED  ] NVFuserTest.FusionPredicateElimination8_CUDA
[  FAILED  ] NVFuserTest.Reduction3DConstantIterationDomain
[  FAILED  ] ExprEvalTest.MmaOp
```

jjsjann123:
!build

jjsjann123:
!build

jjsjann123:
!build

jjsjann123:
failure is real and patched here: https://github.com/NVIDIA/Fuser/pull/1918

jjsjann123:
!build

jjsjann123:
!build

jjsjann123:
marking failing test:
test_matmul `CombineMulSumAsMmaTest.UseMatmulScheduler`

jjsjann123:
!build

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -68,7 +104,7 @@ class AllocationOrderInferencer : public IterVisitor {
   bool propagateAllocationOrder(TensorView* src, TensorView* dst) {
```
wujingyue, line 104: Is it a pre-condition that src-dst must be producer-consumer?

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -68,7 +104,7 @@ class AllocationOrderInferencer : public IterVisitor {
...
     if (auto iter = alloc_order_map_.find(src);
         iter != alloc_order_map_.end()) {
-      alloc_order_map_[dst] = iter->second;
+      alloc_order_map_[dst] = adjustAllocationOrder(iter->first, iter->second);
```
wujingyue, line 107:
```suggestion
      alloc_order_map_[dst] = adjustAllocationOrder(src, iter->second);
```

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -15,30 +15,66 @@ namespace nvfuser::preseg_passes {
...
 
+// returns allocation order to propagate adjusted for `tv`
+// Right now it merely removes all the entry corresponding to reduction iter
+// domains in tv,
+AllocationOrder adjustAllocationOrder(
+    const TensorView* tv,
+    const AllocationOrder& alloc_order) {
+  AllocationOrder ret = alloc_order;
+
+  int64_t tv_rank = static_cast<int64_t>(tv->nDims());
```
wujingyue, line 26:
```suggestion
  auto tv_rank = static_cast<int64_t>(tv->nDims());
```

Nit: `static_cast<int64_t>` makes the explicit type redundant. Using auto here is less error prone -- people may forget to change the receiver type when they change static_cast.

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -15,30 +15,66 @@ namespace nvfuser::preseg_passes {
...
+  auto& rf_dom = tv->getMaybeRFactorDomain();
```
wujingyue, line 27:
```suggestion
  const auto& rf_dom = tv->getMaybeRFactorDomain();
```

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -15,30 +15,66 @@ namespace nvfuser::preseg_passes {
 
 namespace {
 
+// returns allocation order to propagate adjusted for `tv`
+// Right now it merely removes all the entry corresponding to reduction iter
+// domains in tv,
+AllocationOrder adjustAllocationOrder(
+    const TensorView* tv,
+    const AllocationOrder& alloc_order) {
```
wujingyue, line 23: I'm no longer sure I follow the code. Is `alloc_order` the allocation order of `tv` or that of its producer?

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -15,30 +15,66 @@ namespace nvfuser::preseg_passes {
...
+  // by decrement indices higher than the removed entry.
+  while (!idx_stack.empty()) {
+    auto idx = idx_stack.top();
+    idx_stack.pop();
+    for (auto i = ret.begin(); i != ret.end(); i++) {
+      if (*i > idx) {
+        --(*i);
+      }
+    }
+  }
```
wujingyue, line 54: It will take me or any reader more than 10 minutes to review this :) Wdyt about the following alternative: 

1. construct the allocation domain from allocation order, e.g. `[r1,i2,i3,i0]` for the example in the PR description. 
2. map that `vector<IterDomain*>` to the consumer via PairwiseRootMap. I believe `r1` automatically disappear, so you have `[i2,i3,i0]`. 
3. compute the allocation order as `computePermutation(consumer's rfactor, [i2,i3,i0])`. 

Although this converts between allocation order and allocation domain, it makes the intent of filtering out reduction dimensions clearer, and avoids the index/offset calculation.

csrc/preseg_passes/allocation_order_inference.cpp
jjsjann123: alloc_order is the allocation order of `tv`. I see this as a failure in the comment. :bow:

csrc/preseg_passes/allocation_order_inference.cpp
jjsjann123: > map that vector<IterDomain*> to the consumer via PairwiseRootMap.

I'm slightly uncomfortable with this change here, because it bind the logic here as a `producer->consumer` mapping. Which makes it a bit cumbersome to wield.

But I see your point of making the logic easier to read.

I'll refactor it with PairwiseRootDomainMap and remove the util function. (since broadcast propagation also uses it but needs a slightly different behavior.)

tests/cpp/test_allocation_order_inference.cpp
```diff
@@ -222,6 +222,34 @@ TEST_F(AllocationOrderInferenceTest, TernaryOpPropagation) {
...
+
+  std::vector<IterDomain*> tv0_order = {
+      tv0->axis(1), tv0->axis(2), tv0->axis(3), tv0->axis(0)};
+  tv0->setAllocationDomain(tv0_order, true);
+
+  const auto inferred_layout = preseg_passes::inferenceAllocationOrder(&fusion);
+  EXPECT_THAT(inferred_layout.at(tv2), ElementsAre(1, 2, 3, 0));
+  EXPECT_THAT(inferred_layout.at(tv3), ElementsAre(1, 2, 0));
+  EXPECT_THAT(inferred_layout.at(tv4), ElementsAre(1, 0));
+  EXPECT_THAT(inferred_layout.at(tv5), ElementsAre(0, 3, 2, 1));
```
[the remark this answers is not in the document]
jjsjann123: yes you are correct. So this PR doesn't change any of the propagation logic for broadcast. It's here to validate that broadcast + reduction would work properly.

I'll put a comment on that.

jjsjann123:
!build

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -56,19 +45,67 @@ class AllocationOrderInferencer : public IterVisitor {
...
+  //
+  // we construct allocation domain on producer, filtering out reduction, apply
+  // root domain map from producer to consumer.
+  //   [r0', i0', i2', i1'] -> [i0', i2', i1'] -> [i0, i2, i1]
+  // so the function would return [i0, i2, i1]
+  std::vector<IterDomain*> mapAllocDomainNoReductionP2C(
+      TensorView* producer,
+      TensorView* consumer) {
+    // constructing alloc_domain for producer from its root domain, while
+    // filtering out reduction.
```
wujingyue, line 74:
```suggestion
    // filtering out reduction because they won't appear in consumer's domain. 
```

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -56,19 +45,67 @@ class AllocationOrderInferencer : public IterVisitor {
...
+  //    Operation
+  //    |
+  //    v
+  //   consumer rfactor dom [..., i0, ..., i1, ..., i2, ...]
+  //
+  // we construct allocation domain on producer, filtering out reduction, apply
+  // root domain map from producer to consumer.
+  //   [r0', i0', i2', i1'] -> [i0', i2', i1'] -> [i0, i2, i1]
+  // so the function would return [i0, i2, i1]
+  std::vector<IterDomain*> mapAllocDomainNoReductionP2C(
```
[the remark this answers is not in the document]
wujingyue: Naming nit: propagateAllocationDomain, to be more like natural language and be consistent with propagateAllocationOrder.

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -107,12 +144,15 @@ TensorView* AllocationOrderInferencer::resolveAllocationOrder(
   TensorView* src = nullptr;
   size_t non_bc_high_water_mark = 0;
 
-  // helper utils to count the number of non broadcast iterdomain
-  auto countNonBroadcastID = [](const TensorView* tv) -> size_t {
+  // helper utils to count the number of non broadcast / non reduction
+  // iterdomain
+  auto countLoopID = [](const TensorView* tv) -> size_t {
```
wujingyue, line 149:
```suggestion
  auto countLoopIterDomains = [](const TensorView* tv) -> size_t {
```

wujingyue (approved):

jjsjann123:
!build

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -56,24 +45,80 @@ class AllocationOrderInferencer : public IterVisitor {
...
+  //   consumer rfactor dom [..., i0, ..., i1, ..., i2, ...]
+  //
+  // we construct allocation domain on producer, filtering out reduction, apply
+  // root domain map from producer to consumer.
+  //   [r0', i0', i2', i1'] -> [i0', i2', i1'] -> [i0, i2, i1]
+  // so the function would return [i0, i2, i1]
+  std::vector<IterDomain*> propagateAllocationDomain(
+      TensorView* producer,
+      TensorView* consumer) {
+    // constructing alloc_domain for producer from its root domain, while
```
naoyam, line 73: "from its rfactor domain"?

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -56,24 +45,80 @@ class AllocationOrderInferencer : public IterVisitor {
...
+  //    Operation
+  //    |
+  //    v
+  //   consumer rfactor dom [..., i0, ..., i1, ..., i2, ...]
+  //
+  // we construct allocation domain on producer, filtering out reduction, apply
+  // root domain map from producer to consumer.
+  //   [r0', i0', i2', i1'] -> [i0', i2', i1'] -> [i0, i2, i1]
+  // so the function would return [i0, i2, i1]
+  std::vector<IterDomain*> propagateAllocationDomain(
```
naoyam, line 70:
```
propagateAllocationDomainFromProducerToConsumer
```

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -56,24 +45,80 @@ class AllocationOrderInferencer : public IterVisitor {
...
+        constructAllocationDomain(producer, alloc_order_map_.at(producer)));
+    // creating producer to consumer root domain map
+    std::unordered_map<IterDomain*, IterDomain*> p2c_map =
+        PairwiseRootDomainMap(producer, consumer).mapProducerToConsumer();
+    // map alloc_domain to consumer
+    std::transform(
+        alloc_domain.cbegin(),
+        alloc_domain.cend(),
+        alloc_domain.begin(),
+        [&p2c_map](IterDomain* id) { return p2c_map.at(id); });
```
naoyam, line 85: Note that the root domain map may not map all domains.

csrc/preseg_passes/allocation_order_inference.cpp
naoyam: Actually, it's a little confusing to have this and the below function both named as "propagate". This one doesn't update any state, but the other does update. Maybe like `getConsumerAllocationDomainFromProducer`.

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -56,24 +45,80 @@ class AllocationOrderInferencer : public IterVisitor {
...
+  //    v
+  //   consumer rfactor dom [..., i0, ..., i1, ..., i2, ...]
+  //
+  // we construct allocation domain on producer, filtering out reduction, apply
+  // root domain map from producer to consumer.
+  //   [r0', i0', i2', i1'] -> [i0', i2', i1'] -> [i0, i2, i1]
+  // so the function would return [i0, i2, i1]
+  std::vector<IterDomain*> propagateAllocationDomain(
+      TensorView* producer,
+      TensorView* consumer) {
```
naoyam, line 72:
```
TensorView* consumer) const {
```

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -56,24 +45,80 @@ class AllocationOrderInferencer : public IterVisitor {
...
+    return alloc_domain;
+  }
+
+  // propagate allocation order from producer to consumer. Returns true when
+  // producer has a recorded allocation order, false otherwise. This function
+  // assumes that all root domain in consumer can be mapped to producer.
+  bool propagateAllocationOrder(
+      TensorView* producer,
+      TensorView* consumer,
+      const std::vector<IterDomain*>& permutation_ref) {
```
naoyam, line 95: Can you please add comment what `permutation_ref` means?

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -240,6 +288,14 @@ void AllocationOrderInferencer::handle(TernaryOp* op) {
 }
 
 void AllocationOrderInferencer::handle(PadOp* op) {
+  auto* out = dynamic_cast<TensorView*>(op->out());
+  auto* in = dynamic_cast<TensorView*>(op->in());
+  // Note: `out` from pad has rfactor domain that cannot be mapped back to
+  // `in`'s domain. Hence we use `out`'s root domain to match permutation.
+  propagateAllocationOrder(in, out, out->getRootDomain());
```
naoyam, line 295: I don't understand why this is necessary. 

> `out` from pad has rfactor domain that cannot be mapped back to `in`'s domain

This is true, but why does it matter? The other usage of `propagateAllocationOrder` passes the rfactor domain of the out tensor. Why does pad need to pass the root domain?

csrc/preseg_passes/allocation_order_inference.cpp
naoyam: Actually, I don't seem to understand how this would work. `alloc_domain` is effectively a (potentially reordered) list of iter domains of the producer rfactor domain. Why could that be the allocation domain of the consumer? For example, how would reshape work?

csrc/preseg_passes/allocation_order_inference.cpp
jjsjann123: usually we have:
producer_rfactor_domain: {i0, i1, i2}
with allocation order, we'll construct a producer alloc domain as `permutation({i0, i1, i2})`

with a pointwise, we'll have consumer_rfactor_domain: {i0, i1, i2}, so it can be mapped to the above permutation to get consumer alloc domain

with pad, what we have here is that:
consumer_root_domain: {i0, i1, i2}
consumer_rfactor_domain: {(i0+-pad), (i1+-pad), (i2+-pad)}

We can no longer map consumer_rfactor_domain to its producer alloc domain, hence we need to special case on pad.

PR #1897, opened 2024-03-11
Index compute with non-outermost sharding

cowanmeg:
Updates strided index computation to properly handle non-outermost device parallel axes.
Specifically, it updates stride calculations so that the sharded extents are treated as size 1 instead of the unsharded extents.

This PR only updates index computations. A later PR will update communication lowering to handle non-outermost sharded axes.

### DIFF
--- a/csrc/index_compute.cpp
+++ b/csrc/index_compute.cpp
@@ -1222,6 +1222,8 @@ Val* getHaloExtentOfRootAxis(IterDomain* id, Val* normal_extent = nullptr) {
         SimplifyingIrBuilder::create<Val>(
             (int64_t)halo.width(), DataType::Index));
     return halo_extent;
+  } else if (id->isDeviceDim()) {
+    return GpuLower::current()->kernel()->oneVal();
   } else {
     return normal_extent;
   }
@@ -1601,12 +1603,15 @@ std::vector<Val*> Index::getGlobalProducerStridedIndices(
     for (const auto i : c10::irange(alloc_dom.size())) {
       if (alloc_dom[i]->isReduction()) {
         strides[i] = GpuLower::current()->kernel()->oneVal();
-        continue;
+      } else if (alloc_dom[i]->isDeviceDim()) {
+        strides[i] = GpuLower::current()->kernel()->oneVal();
+        stride_i++;
+      } else {
+        strides[i] = IrBuilder::getItemExpr(
+            IrBuilder::getAttrExpr(
+                IrBuilder::metadataExpr(producer_tv), "alloc_stride"),
+            (int64_t)stride_i++);
       }
-      strides[i] = IrBuilder::getItemExpr(
-          IrBuilder::getAttrExpr(
-              IrBuilder::metadataExpr(producer_tv), "alloc_stride"),
-          (int64_t)stride_i++);
     }
   }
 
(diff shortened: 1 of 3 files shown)


cowanmeg:
!build

csrc/index_compute.cpp
```diff
@@ -1222,6 +1222,8 @@ Val* getHaloExtentOfRootAxis(IterDomain* id, Val* normal_extent = nullptr) {
         SimplifyingIrBuilder::create<Val>(
             (int64_t)halo.width(), DataType::Index));
     return halo_extent;
+  } else if (id->isDeviceDim()) {
```
samnordmann, line 1225: just out of curiosity: when is "Halo" involved and what does it mean?

csrc/index_compute.cpp
```diff
@@ -1601,12 +1603,15 @@ std::vector<Val*> Index::getGlobalProducerStridedIndices(
     for (const auto i : c10::irange(alloc_dom.size())) {
       if (alloc_dom[i]->isReduction()) {
```
samnordmann, line 1604: a one-line patch should be sufficient here, right?
```suggestion
      if (alloc_dom[i]->isReduction() || alloc_dom[i]->isDeviceDim()) {
```

csrc/index_compute.cpp
```diff
@@ -1979,11 +1984,14 @@ std::vector<Val*> Index::getStrides(TensorView* tv) {
     for (const auto i : c10::irange(alloc_dom.size())) {
       if (alloc_dom[i]->isReduction() || alloc_dom[i]->isStride()) {
```
samnordmann, line 1985: Same here, correct me if I'm wrong but I think that a one-line change does the job
```suggestion
      if (alloc_dom[i]->isReduction() || alloc_dom[i]->isStride() || alloc_dom[i]->isDeviceDim()) {
```

test/test_multidevice_sharding.cpp
```diff
@@ -18,6 +18,46 @@
 
 namespace nvfuser {
 
+// Test single device computation with device parallelized axis is handled
```
samnordmann, line 21: Why not testing also on non-trivial meshes ?

csrc/index_compute.cpp
cowanmeg: An explanation would also be helpful for me. I've seen halo mentioned in papers before...
@naoyam is it the same halo exchange for data format as mentioned in the gspmd paper? (https://arxiv.org/pdf/2105.04663.pdf)

csrc/index_compute.cpp
cowanmeg: Not quite. The `isReduction` branch doesn't increment `stride_i` since reduction iter domains are not included in allocation domain (someone please correct me if I'm wrong), but the `isDeviceDim` branch increments `stride_i` since sharded iter domains are in the allocation domain.

test/test_multidevice_sharding.cpp
cowanmeg: Good point, will update to a non-trivial mesh.

csrc/index_compute.cpp
naoyam: Halo is just a common word in parallel scientific computing. It's a boundary dependent region when a problem is partitioned on disjoint memory spaces. https://wgropp.cs.illinois.edu/courses/cs598-s15/lectures/lecture25.pdf

naoyam: No need to worry about the current halo support. I'll most likely remove it soon as it should be completely redesigned from scratch.

csrc/index_compute.cpp
```diff
@@ -1979,11 +1984,14 @@ std::vector<Val*> Index::getStrides(TensorView* tv) {
...
         strides[i] = GpuLower::current()->kernel()->oneVal();
-        continue;
+      } else if (alloc_dom[i]->isDeviceDim()) {
+        strides[i] = GpuLower::current()->kernel()->oneVal();
```
naoyam, line 1988: Does this matter? Indexing of DID parallelized allocation domains should be always zero, right? If so, strides shouldn't matter.

cowanmeg: Actually, you are right there shouldn't be a special case for DeviceDim; it should be handled in the else branch. 
These were just bandaids to avoid the actual problem in expr_evaluator that Sam fixed in https://github.com/NVIDIA/Fuser/pull/1900. 
I will close this PR and clean it up.

PR #1900, opened 2024-03-11
Multidevice: Avoid compiling when allocating

samnordmann:
This patch accelerates the multidevice allocation which was the main overhead in our runtime.

Indeed, previously we needed to compile and run a Fusion just for allocation. This path gets rid of this step by simply computing the size of the allocation and allocating directly.

Replaces https://github.com/NVIDIA/Fuser/pull/1854

possible TODO (not sure it's worth it): cache the compute of the allocation size depending on the input's sizes.


wujingyue:
cc @Priya2698 who's trying to avoid compilation for fusions that are marked for ExpressionEvaluator (e.g. when all outputs are aliases).

csrc/multidevice/executor.cpp
```diff
@@ -129,6 +76,23 @@ MultiDeviceExecutor::MultiDeviceExecutor(
   }
   // prepare the order in which to launch the kernels/comms
   prepareRuntimeOrder(staged_fusion_.get(), workspace);
+
+  // Allocator setup
+  for (auto group : staged_fusion_->groups()) {
```
wujingyue, line 81:
```suggestion
  for (SegmentedGroup* group : staged_fusion_->groups()) {
```

csrc/executor.cpp
```diff
@@ -1384,6 +1374,19 @@ std::vector<FusionExecutor::GlobalBufferInfo> FusionExecutor::
   return outputs;
 }
 
+std::vector<at::Tensor> allocOutputSpace(
+    const at::ArrayRef<c10::IValue>& inputs,
+    kir::Kernel* kernel,
```
wujingyue, line 1379: Question, not request for change: I guess `Fusion` should be good enough here (of course, refactors required). allocOutputSpace doesn't use anything that's only in the lowered Kernel but not in the Fusion. Correct? 

cc @Priya2698

csrc/multidevice/executor.h
```diff
@@ -139,6 +135,10 @@ class MultiDeviceExecutor {
   std::unordered_map<SegmentedGroup*, bool> should_run_;
   // Cache whether a SegmentedGroup requires inter-device communication
   std::unordered_map<SegmentedGroup*, bool> is_resharding_;
+  // Cached objects used for MultiDevice allocation
+  std::unique_ptr<kir::Kernel> allocator_kernel_;
+  std::vector<Val*> vals_to_allocate_;
```
wujingyue, line 140: Can you add a code comment for `vals_to_allocate`? I didn't review the initial PR where you created it, so I'm lost on its meaning.

samnordmann:
WAR: I'm realizing the present patch breaks the multidevice support for symbolic sizes, which is only featured in the test suite `ShardingTest` in `test/test_multidevice_sharding.cpp`. Indeed neither in #1840 or in the present pr, we provide a way to bind in the ExprEvaluator the extents of the sharded axis with the extent of the DeviceMesh.

This should be relatively easy once we make the DeviceMesh symbolic. This will be done in a future PR soon (I note it as a TODO).

Any other idea? If not, in my opinion, this should not prevent us from merging this PR.

@cowanmeg

samnordmann:
> WAR: I'm realizing the present patch breaks the multidevice support for symbolic sizes, which is only featured in the test suite `ShardingTest` in `test/test_multidevice_sharding.cpp`. Indeed neither in #1840 or in the present pr, we provide a way to bind in the ExprEvaluator the extents of the sharded axis with the extent of the DeviceMesh.
> 
> This should be relatively easy once we make the DeviceMesh symbolic. This will be done in a future PR soon (I note it as a TODO).
> 
> Any other idea? If not, in my opinion, this should not prevent us from merging this PR.
> 
> @cowanmeg

Nevermind, I actually fixed it, by directly binding the sharded extent to the concrete Value (see the [last commit](https://github.com/NVIDIA/Fuser/pull/1900/commits/34a767985d080325c039639c8777791846ef5b54))

samnordmann:
@wujingyue Thanks for your comments, I have addressed them

csrc/expr_evaluator.cpp
```diff
@@ -149,6 +149,10 @@ void ExpressionEvaluator::bind_(
         // Ignore concrete extents because they hold the unsharded extents.
         NVF_CHECK(
             1 == t.size(i), "Tried to bind a constant value 1 as ", t.size(0));
+        bind_(
```
naoyam, line 152: I wonder why we need this for DIDx but not for TIDx/BIDx

samnordmann: I am not sure to understand. Tid/Bid are not needed to compute the allocation size (at least in our examples)

The extent of the shared axis can only be deduced through the device mesh's size

naoyam: > The extent of the shared axis can only be deduced through the device mesh's size

Generally, this is not true. When a domain is annotated with, e.g., TIDx, it only means `blockDim.x` is greater than or equal to the extent of the domain. 

It also seems strange that the extent of a domain is determined by the number of devices. Shouldn't it be the other way around?

csrc/executor.cpp
```diff
@@ -1342,34 +1342,25 @@ std::vector<FusionExecutor::GlobalBufferInfo> FusionExecutor::
...
-
-std::vector<FusionExecutor::GlobalBufferInfo> FusionExecutor::
-    getOutputBufferInfo(
-        const KernelArgumentHolder& args,
-        ExpressionEvaluator& expr_eval,
-        DataType index_dtype) {
+//! Return information necessary for allocating output tensors. Input
+//! and output tensors are allowed to alias each other, which is
+//! specified by the list of int pairs of input and output indices
+std::vector<FusionExecutor::GlobalBufferInfo> getOutputBufferInfo(
```
naoyam, line 1348: Should this be in an anonymous namespace? Looks like it's dropped from the header file.

csrc/executor.cpp
```diff
@@ -1342,34 +1342,25 @@ std::vector<FusionExecutor::GlobalBufferInfo> FusionExecutor::
...
-  const auto kernel = lowered_->kernel();
-  std::vector<GlobalBufferInfo> outputs;
+  std::vector<FusionExecutor::GlobalBufferInfo> outputs;
+  outputs.reserve(kernel->outputs().size());
   NVF_ERROR(
       args.size() == kernel->inputs().size(),
       "kernel arguments length does not match runtime arguments.");
   for (const auto out_i : c10::irange(kernel->outputs().size())) {
     auto out_val = kernel->outputs()[out_i];
     auto output = out_val->as<TensorView>();
```
naoyam, line 1361: Looks like we have `output`, which is a TensorView pointer, and also we do dynamic_cast with `out_val`. 

Maybe just delete `output` and replace it with `info.tv`.

csrc/expr_evaluator.cpp
cowanmeg: > When a domain is annotated with, e.g., TIDx, it only means blockDim.x is greater than or equal to the extent of the domain.

If BlockDim.x > TIDx extent does this imply that some threads in will not allocate local memory/perform any computation related to that tensor?

For sharding, we want the constraint that if a tensor is sharded over a mesh, that at least some part of the tensor is allocated on those devices in the mesh. 
Then we can say if a tensor of size [1, 4] is sharded on dimension 0 across mesh{0,1,2}, we know the unsharded size must be [3,4]

csrc/multidevice/executor.cpp
```diff
@@ -39,66 +39,13 @@ copyFusionAndChangeOutputs(Fusion* fusion, const std::vector<Val*>& outputs) {
...
   });
 
-  for (auto tv : ir_utils::filterByType<TensorView>(fusion_copy->vals())) {
-    tv->setMemoryType(MemoryType::Global);
-    for (auto i : c10::irange(tv->domain()->nDims())) {
-      if (!tv->axis(static_cast<int>(i))->isDeviceDim()) {
-        tv->axis(static_cast<int>(i))->parallelize(ParallelType::Serial);
-      }
-    }
-  }
```
naoyam, line 52: Why this was necessary before?

csrc/expr_evaluator.cpp
samnordmann: > It also seems strange that the extent of a domain is determined by the number of devices. Shouldn't it be the other way around?

We can see it in both way, the values just need to match. In our case since the DeviceMesh are not yet symbolic they hold the concrete value (which is not necessarily the number of devices)

I will soon make the DeviceMesh symbolic and those considerations will be made clearer hopefully

csrc/multidevice/executor.cpp
```diff
@@ -226,7 +190,12 @@ std::vector<at::Tensor> MultiDeviceExecutor::runWithInput(
       inputs.size() == staged_fusion_->inputs().size(),
       "Wrong number of inputs");
 
-  val_to_IValue_ = allocateRecvBuffers(inputs);
+  auto allocations =
```
naoyam, line 193: IIUC, we are allocating all the recv buffers separately even when, e.g., one buffer could be reused, right? 

Have you considered reusing buffers?

naoyam (commented):
Looks good overall. This one is my only true concern: https://github.com/NVIDIA/Fuser/pull/1900/files/34a767985d080325c039639c8777791846ef5b54#r1522106849

csrc/expr_evaluator.cpp
naoyam: > If BlockDim.x > TIDx extent does this imply that some threads in will not allocate local memory/perform any computation related to that tensor?

Yes

> For sharding, we want the constraint that if a tensor is sharded over a mesh, that at least some part of the tensor is allocated on those devices in the mesh.

Yes, having a constraint like this is fine. I just wanted to make sure you're aware of the deviation from TIDx/BIDx.

samnordmann: > Generally, this is not true. When a domain is annotated with, e.g., TIDx, it only means `blockDim.x` is greater than or equal to the extent of the domain.

We only support the case where there is strict equality between the axis's and DeviceMesh's extent. I don't even know what it would mean if we have just an inequality (especially in our case where we forbid split/merge)

More fundamentally, the Fusion needs to know __somehow__ the extents of the sharded axis, which cannot be deduced by the tensor's concrete shape (since sharded dimensions are not allocated)

All these are important points but they are not related to this PR though :)

naoyam: > I will soon make the DeviceMesh symbolic

So, then how would you determine the extent?

naoyam: And back to my original question, I'm trying to understand why we need this. We don't need to infer tensor shapes from blockDim.x or gridDim.x, but somehow we need for DIDx parallelized domains. I'm not saying we should not make [this assumption](https://github.com/NVIDIA/Fuser/pull/1900#discussion_r1522139313), and I understand that the assumption allows us to do the inference of extents. My question is why it's necessary only for DIDx parallelized domains.

cowanmeg: Oh I see your question.
This needed when a global input is a sharded. Since the input to the program will be the sharded tensor, we need to calculate unsharded extents.

naoyam: Thanks @samnordmann , I think I now understand why.

> since sharded dimensions are not allocated

So, the fundamental issue is the input to nvFuser, i.e., an aten tensor, does not convey anything about the distributed allocation, so we need to make this assumption atm. Ideally, sharding information should be also given to nvFuser, which should also tell the overall logical shape.

Does this make sense?

naoyam: @samnordmann Could you please add some more comments to the code?

csrc/executor.cpp
naoyam: Let's change the parameter to `Fusion*` (assuming the change is trivial). Having `Kernel` indicates it needs lowering.

csrc/expr_evaluator.cpp
samnordmann: Yep, that make sense. Let me add some comment

csrc/executor.cpp
samnordmann: correct!

csrc/multidevice/executor.cpp
samnordmann: I think it was just some dirty hack to try to avoid compilation/lowering error when using FusionExecutorCache due to previous scheduling (while keeping multi-device scheduling because it affects the allocation shape). It is a bit similar to https://github.com/NVIDIA/Fuser/blob/597c9b0c67ae09712688f34aa99da2aaf725f60f/test/multidevice.cpp#L98

This is not needed anyway with this cleaner method

csrc/multidevice/executor.cpp
samnordmann: That's correct! No, there is no support for reusing buffers.
We also do not support inplace collective for now, so there is allocation also when the receiver is the sender as well.

These optimizations could be done later-on, and then the patch will be quit orthogonal to this pr, only affecting the content of `vals_to_allocate_`

samnordmann: I added a comment and a TODO in the `MultiDeviceExecutor`'s constructor

csrc/executor.cpp
samnordmann: That makes sense! I didn't do it upfront bc I was lazy, but it is done now! :)

samnordmann:
!build

naoyam (approved):
LGTM. Thanks for addressing my concerns.

csrc/multidevice/executor.h
```diff
@@ -116,10 +116,6 @@ class MultiDeviceExecutor {
   // execute a SegmentedGroup representing inter-device communication
   void postCommunication(SegmentedGroup* group);
 
-  // allocate inter-device communication recv buffers
-  std::unordered_map<Val*, c10::IValue> allocateRecvBuffers(
-      std::vector<c10::IValue> global_inputs_IValues);
-
   // Stores concrete computed values,
   std::unordered_map<Val*, c10::IValue> val_to_IValue_;
```
wujingyue, line 120: (I understand this doesn't come from this PR.)

Why IValue instead of at::Tensor? Reading [its current usage](https://github.com/search?q=repo%3ANVIDIA%2FFuser%20val_to_ivalue_&type=code), `at::Tensor` seems to be good enough and probably simplifies (no need to toTensor etc).

csrc/multidevice/executor.h
```diff
@@ -139,6 +135,12 @@ class MultiDeviceExecutor {
   std::unordered_map<SegmentedGroup*, bool> should_run_;
   // Cache whether a SegmentedGroup requires inter-device communication
   std::unordered_map<SegmentedGroup*, bool> is_resharding_;
+  // Cached objects used for MultiDevice allocation
+  std::unique_ptr<Fusion> allocator_fusion_;
+  // Cache the tensors that need to be allocated at runtime, which correspond to
+  // the destination buffers of interdevice communications.
+  std::vector<Val*> vals_to_allocate_;
```
wujingyue, line 142: With this PR, we'll have both `val_to_IValue_` and `vals_to_allocate`. Is that a necessary complexity? According to https://github.com/NVIDIA/Fuser/pull/1900/files#diff-d783a79fdc189059065a81ba9acb6d6c3d3ecc5c70188706febc97d7fea5aa36R199-R201, the former seems to subsume the latter.

wujingyue (commented):
LGTM! My comments may lead to cleanups to be addressed in this PR or a separate PR.

csrc/multidevice/executor.h
samnordmann: I think it is not possible, at least in the present state. The problem would come from using multiple time the executor.
After being used at least once, `val_to_IValue_` will contain more keys than the one allocated, and so the second call to  https://github.com/NVIDIA/Fuser/pull/1900/files#diff-d783a79fdc189059065a81ba9acb6d6c3d3ecc5c70188706febc97d7fea5aa36R199-R201 will be wrong

csrc/multidevice/executor.h
samnordmann: Thanks for the suggestion. I actually agree. I will open a new short PR to address that

PR #1905, opened 2024-03-12
Disable options via e.g. fd.execute(inputs, disable=['index_hoist'])

jacobhinkle:


csrc/options.cpp
```diff
@@ -164,27 +164,57 @@ std::unordered_map<EnableOption, std::vector<std::string>> Options<
   return parseEnvOptions("ENABLE", available_options);
 }
 
+const std::unordered_map<std::string, std::pair<DisableOption, std::string>>&
+getDisableOptionDescriptions() {
```
jacobhinkle, line 168: Included the comments as description strings here so that we can fill out an informative docstring.

kevinstephano:
We need to discuss exposure of options in python.  In theory, most of our exposure should be through `FusionDefinition.execute()`.

jacobhinkle:
Irrelevant after #3270

PR #1911, opened 2024-03-12
Add `NVFUSER_DUMP=predicate_elimination`

zasdfgbnm:


zasdfgbnm:
!build

zasdfgbnm:
!build

csrc/device_lower/analysis/predicate_elimination.cpp
```diff
@@ -1035,13 +1036,27 @@ Val* PredicateElimination::getInitValue(TensorView* tv) const {
...
+        non_predicated_tvs.pushBack(ti->view());
+      } else if (auto tv = dynamic_cast<TensorView*>(out)) {
+        non_predicated_tvs.pushBack(tv);
+      } else {
+        NVF_ERROR(
+            false,
+            "Unexpected output ",
+            out->toString(),
+            " in ",
+            expr->toString());
```
wujingyue, line 1052:
```suggestion
            expr);
```

csrc/device_lower/analysis/predicate_elimination.cpp
```diff
@@ -1053,6 +1068,10 @@ std::string PredicateElimination::toString() const {
     }
   }
   ss << "\n";
+  ss << "Non-predicated expressions:";
+  for (auto expr : non_predicated_exprs_) {
+    ss << " " << expr->toString();
```
wujingyue, line 1073:
```suggestion
    ss << " " << expr;
```

csrc/device_lower/analysis/predicate_elimination.cpp
```diff
@@ -1035,13 +1036,27 @@ Val* PredicateElimination::getInitValue(TensorView* tv) const {
 
 std::string PredicateElimination::toString() const {
   std::stringstream ss;
-  ss << "Tensors that do not need predication:";
+  VectorOfUniqueEntries<TensorView*> non_predicated_tvs;
```
wujingyue, line 1039: I guess you didn't use `set<TensorView*>` because you like to print them in the order that they appear in the fusion?

wujingyue (approved):

csrc/device_lower/analysis/predicate_elimination.cpp
zasdfgbnm: Yes

csrc/device_lower/analysis/predicate_elimination.cpp
```diff
@@ -1035,13 +1036,27 @@ Val* PredicateElimination::getInitValue(TensorView* tv) const {
...
-      ss << " T" << out->name();
+      if (auto ti = dynamic_cast<kir::TensorIndex*>(out)) {
+        non_predicated_tvs.pushBack(ti->view());
+      } else if (auto tv = dynamic_cast<TensorView*>(out)) {
+        non_predicated_tvs.pushBack(tv);
+      } else {
+        NVF_ERROR(
+            false,
+            "Unexpected output ",
+            out->toString(),
```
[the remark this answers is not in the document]
zasdfgbnm: Why? I saw weird behavior for using pointers directly like suggested. Sometimes it prints the pointer address, but sometimes it prints the nice string as generated in `toString`. I believe the root cause is because by default, C++ already has `operator<<(std::ostream&, T*)` for pointers that print address, but we are defining our own `operator<<(std::ostream&, T*)` for our own IR node in some header. So depending on whether this header is included or not, the print of pointers will be different, and I feel this behavior confusing, that's why I always manually `toString`. Is this problem resolved? Do we have a general guideline on what we should prefer in our coding style?

PR #1912, opened 2024-03-12
Rename `PredicateAnalyzer` -> `ProducerConsumerPairAnalyzer`

zasdfgbnm:
Right now, we have `PredicateChcker` and `PredicateAnalyzer`. From their name, we can not tell which is doing what. After reading the code, we can tell that, `PredicateChcker` is the global entry point for predicate elimination, and `PredicateAnalyzer` is a helper class for `predicateProducerConsumerPair`, which is only one condition of `PredicateChcker`: https://github.com/NVIDIA/Fuser/blob/638bfe00a29f582a337042c04a63c13410b8b17f/csrc/device_lower/analysis/predicate_elimination.cpp#L222
So I do this rename to indicate this fact.

### DIFF
--- a/csrc/device_lower/analysis/predicate_elimination.cpp
+++ b/csrc/device_lower/analysis/predicate_elimination.cpp
@@ -59,7 +59,7 @@ bool isExactParallelSharedMemAccess(TensorView* tv) {
   return true;
 }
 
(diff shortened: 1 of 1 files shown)


naoyam (approved):
Yes, that was a bad unclear name. Thanks for improving it!

PR #1913, opened 2024-03-12
options should be thread local

protonu:
This PR makes the options flags thread_local.
Once such use case would be where tests are run in parallel and they need to have these flags set with different values.


protonu:
!build

jjsjann123 (approved):
cc'ing @jacobhinkle

jacobhinkle (commented):
Nice catch @protonu!

PR #1917, opened 2024-03-13
Fix stages=0 smem heuristic

jacobhinkle:
This simply fixes an integer division by zero error in `generateSharedMemoryEpilogueHeuristics()` when `params.double_buffer_options.smem_double_buffer_stage == 0`. The modified test fails on main with
```
[ RUN      ] NVFuserTest.FusionAmpereMatmulSmemEpilogue_CUDA
[penny:1500 :0:1500] Caught signal 8 (Floating point exception: integer divide by zero)
==== backtrace (tid:   1500) ====
 1 0x00000000009f384f nvfuser::mma_utils::generateSharedMemoryEpilogueHeuristics()  /opt/pytorch/nvfuser/csrc/scheduler/mma_utils.cpp:158
 2 0x0000000000be3b90 nvfuser::NVFuserTest_FusionAmpereMatmulSmemEpilogue_CUDA_Test::TestBody()  /opt/pytorch/nvfuser/test/test_gpu_tensorcore.cpp:2392
 3 0x0000000000c5fdd9 testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void>()  /opt/pytorch/nvfuser/third_party/googletest/googletest/src/gtest.cc:2612
 4 0x0000000000c592c5 testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>()  /opt/pytorch/nvfuser/third_party/googletest/googletest/src/gtest.cc:2648
 5 0x0000000000c33ef4 testing::Test::Run()  /opt/pytorch/nvfuser/third_party/googletest/googletest/src/gtest.cc:2687
 6 0x0000000000c34a0f testing::TestInfo::Run()  /opt/pytorch/nvfuser/third_party/googletest/googletest/src/gtest.cc:2836
 7 0x0000000000c35415 testing::TestSuite::Run()  /opt/pytorch/nvfuser/third_party/googletest/googletest/src/gtest.cc:3015
 8 0x0000000000c4593c testing::internal::UnitTestImpl::RunAllTests()  /opt/pytorch/nvfuser/third_party/googletest/googletest/src/gtest.cc:5920
 9 0x0000000000c60d66 testing::internal::HandleSehExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>()  /opt/pytorch/nvfuser/third_party/googletest/googletest/src/gtest.cc:2612
10 0x0000000000c5a3c9 testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>()  /opt/pytorch/nvfuser/third_party/googletest/googletest/src/gtest.cc:2648
11 0x0000000000c43f3d testing::UnitTest::Run()  /opt/pytorch/nvfuser/third_party/googletest/googletest/src/gtest.cc:5484
12 0x0000000000c203a8 RUN_ALL_TESTS()  /opt/pytorch/nvfuser/third_party/googletest/googletest/include/gtest/gtest.h:2317
13 0x0000000000c20321 main()  /opt/pytorch/nvfuser/third_party/googletest/googletest/src/gtest_main.cc:64
14 0x0000000000029d90 __libc_init_first()  ???:0
15 0x0000000000029e40 __libc_start_main()  ???:0
16 0x000000000007f895 _start()  ???:0
=================================
Floating point exception (core dumped)
```

### DIFF
--- a/csrc/scheduler/mma_utils.cpp
+++ b/csrc/scheduler/mma_utils.cpp
@@ -104,8 +104,10 @@ std::pair<bool, bool> generateSharedMemoryEpilogueHeuristics(
 
   // Create a temporary DoubleBufferOptions with full double buffering, for
   // estimating shared memory size.
+  // Note that we clip smem_double_buffer_stage to 1 since we will always load
+  // operands to smem even if stages=0
   MatmulParams::DoubleBufferOptions double_buffer_options{
-      true, true, smem_double_buffer_stage};
+      true, true, std::min(1, smem_double_buffer_stage)};
 
   const auto [smem_a, smem_b, smem_c] =
       computeSharedMemorySizes(gemm_tile, double_buffer_options, data_types);
@@ -150,12 +152,9 @@ std::pair<bool, bool> generateSharedMemoryEpilogueHeuristics(
   // total_without_smem_epilogue can be 0, i.e. smem_a == smem_b == 0, if
   // stages == 0. In such case, occupancy will not be bound by shared memory so
   // we use the register bound instead.
-  const auto blocks_per_sm_without_smem_epilogue =
-      total_without_smem_epilogue == 0
-      ? blocks_per_sm_by_register
-      : std::min(
-            shared_memory_available / total_without_smem_epilogue,
-            (size_t)blocks_per_sm_by_register);
+  const auto blocks_per_sm_without_smem_epilogue = std::min(
+      shared_memory_available / total_without_smem_epilogue,
+      (size_t)blocks_per_sm_by_register);
   const auto blocks_per_sm_with_reused_smem_epilogue = std::min(
       shared_memory_available / total_with_reused_smem_epilogue,
       (size_t)blocks_per_sm_by_register);
--- a/csrc/scheduler/mma_utils.cpp
+++ b/csrc/scheduler/mma_utils.cpp
@@ -91,7 +91,7 @@ int64_t computeExpectedSharedMemoryUsage(
 
 std::pair<bool, bool> generateSharedMemoryEpilogueHeuristics(
     const MatMulTileOptions& gemm_tile,
-    const int smem_double_buffer_stage,
+    int smem_double_buffer_stage,
     const MmaDataTypes& data_types,
     bool smem_a_reuse_guaranteed,
     bool smem_b_reuse_guaranteed,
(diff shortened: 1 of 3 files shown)


jacobhinkle:
!build

zasdfgbnm:
Why is stages=0 a valid number? Isn't the non-double-buffered case stage=1 instead of stages=0?

jacobhinkle:
We generate the same kernel for stages=0 and stages=1. That might be fine for now, but it could be confusing and we'd probably need to maintain it in the future as the scheduler evolves. Instead, if we drop stages=0 support then should we assert `smem_double_buffer_stage > 0` here in `generateSharedMemoryEpilogueHeuristics`? This is a little bit of a strange place to put such assertion since this is a helper function to be used by heuristics and it might not be called before `scheduleMatmul`. Rejecting in `scheduleMatmul` is too late to prevent the division by zero encountered here in case `generateSharedMemoryEpilogueHeuristics` is called first. How about this: keep the fix here so that we don't crash heuristics even if they choose weird parameters, then assert that stages > 0 in `scheduleMatmul` with an informative error message?

jacobhinkle:
The scheduler ignores the stages param if double buffering is off, effectively treating it as `1`. I just pushed a change that clips the stages to 1 to mimic this. That way the smem estimate is still accurate, and we'll get the expected kernel. We could print a warning that stages=0 should be set as stages=1, as a signal that heuristics should be updated.

drzejan2:
@jacobhinkle the fix and additional comments are good, thank you for adding them!

A side note, question
the situation where stages are 0, does not mean it that params are incorrectly defined?
The process of creating `MatmulParams` and `DoubleBufferOptions` included in it should make sure that `params.double_buffer_options.smem_double_buffer_stage` is 2. I'm trying to understand what lead to this situation.

jacobhinkle:
> I'm trying to understand what led to this situation.

This configuration was created by our internal heuristics generator. I have patched that tool to no longer suggest these types of kernels but you're right that it would be better to guard against creating invalid configs. Currently `MatmulParams` is struct-like in that all its attributes are public, but we could create constructors for it instead.

csrc/scheduler/mma_utils.cpp
```diff
@@ -147,6 +149,9 @@ std::pair<bool, bool> generateSharedMemoryEpilogueHeuristics(
       warp_dims.m * warp_dims.n * warp_dims.k * properties->warpSize;
   const auto threads_per_sm = getThreadsPerSMGivenRegPerThread(255);
   const auto blocks_per_sm_by_register = threads_per_sm / threads_per_block;
+  // total_without_smem_epilogue can be 0, i.e. smem_a == smem_b == 0, if
```
zasdfgbnm, line 152: Is this comment still related?

jacobhinkle: Removed

test/test_gpu_tensorcore.cpp
```diff
@@ -2350,141 +2350,143 @@ TEST_F(NVFuserTest, FusionAmpereMatmulSmemEpilogue_CUDA) {
   constexpr bool ignore_occupancy_drop = true;
   // Keep multiples of 8 to keep vectorizable.
   int M = 4096, N = 4096, K = 4096;
-  for (auto layout : kAllSupportedMmaLayout) {
-    Fusion fusion;
-    FusionGuard fg(&fusion);
+  for (int num_stages : {0, 2}) {
```
zasdfgbnm, line 2353: Could you add a comment here and in the definition of `MatmulParams::DoubleBufferOptions` about what `num_stages <= 1` means?

zasdfgbnm (approved):

test/test_gpu_tensorcore.cpp
jacobhinkle: Added comments.

PR #1896, opened 2024-03-10
avoid assert when consumer_innermost_ndims == 0

jjsjann123:
Fixes #1895

In `getVectorizationBreakPointOfReductionProducer`, the existing logic walks through mapped innermost iterdomain on reduction producer to find break_point. However, it doesn't account for case where no innermost iterdomain is mapped (i.e. `consumer_innermost_ndims == 0`) and triggers a false assert.

PR #1918, opened 2024-03-13
Patch out of bound access in `getVectorizationFactor`

jjsjann123:
Follow up on a previous fix #1896 

When break_point is at the boundary, added a short-cut to return 1 for vectorization factor.
Since existing vectorization analysis assumed to have a non-trivial break_point, which result in an out of bound access.

### DIFF
--- a/test/test_allocation_domain.cpp
+++ b/test/test_allocation_domain.cpp
@@ -1420,8 +1420,7 @@ TEST_F(NVFuserTest, ReductionVectorization) {
   FusionExecutorCache executor_cache(std::move(fusion));
   auto cg_outputs = executor_cache.runFusionWithInputs(inputs);
 
-  testValidate(
-      executor_cache.fusion(), cg_outputs, inputs, __LINE__, __FILE__);
+  testValidate(executor_cache.fusion(), cg_outputs, inputs, __LINE__, __FILE__);
 }
 
 } // namespace nvfuser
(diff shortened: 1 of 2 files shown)


jjsjann123:
!build

zasdfgbnm (approved):

test/test_allocation_domain.cpp
```diff
@@ -1384,4 +1384,45 @@ TEST_F(NVFuserTest, ReductionSchedulerIssue1895) {
...
+  auto t0 = at::randn({x, 1, z}, options);
+  auto t1 = at::randn({x, y, z}, options).as_strided({x, y, z}, {1, x, x * y});
+  std::vector<c10::IValue> inputs({t0, t1});
+
+  FusionExecutorCache executor_cache(std::move(fusion));
+  auto cg_outputs = executor_cache.runFusionWithInputs(inputs);
+
+  auto ref = t0.mul(t1).sum({2});
+  testValidate(
+      executor_cache.fusion(), cg_outputs, inputs, {ref}, __LINE__, __FILE__);
```
wujingyue, line 1425:
```suggestion
  testValidate(
      executor_cache.fusion(), cg_outputs, inputs, __LINE__, __FILE__);
```

Or is there a particular reason to compute ref by hand?

wujingyue (approved):

test/test_allocation_domain.cpp
jjsjann123: There really isn't. I just kept forgetting that when copying tests and didn't put too much thought into it during clean up.

PR #1921, opened 2024-03-13
Evaluate multiple ops together

Priya2698:
This PR is intended to add support for evaluating multiple ops together and skip evaluating intermediate inputs.
Current use cases:

1. CatOp: Currently, CatOp::evaluate uses the unpadded inputs, however the padded inputs are still evaluated.
2. Matmul Fallback path: We need the expression evaluator to evaluate patterns such as `MmaOp + Cast`, `MmaOp + Bias + Cast` in a single call instead of one op at a time.

### DIFF
--- a/csrc/expr_evaluator.cpp
+++ b/csrc/expr_evaluator.cpp
@@ -199,6 +199,11 @@ PolymorphicValue ExpressionEvaluator::evaluate(const Val* value) const {
   return evaluateHelper(value, known_values);
 }
 
+PolymorphicValue ExpressionEvaluator::evaluate(const Val* value, 
+  std::unordered_map<const Val*, PolymorphicValue>& known_values) const {
+  return evaluateHelper(value, known_values);
+}
+
 const PolymorphicValue& ExpressionEvaluator::evaluateHelper(
     const Val* value,
     std::unordered_map<const Val*, PolymorphicValue>& known_values) const {
@@ -213,16 +218,7 @@ const PolymorphicValue& ExpressionEvaluator::evaluateHelper(
   if (!maybe_concrete_value.get().hasValue()) {
     if (auto def = value->definition()) {
       FUSER_PERF_SCOPE("ExpressionEvaluator::evaluate");
-      std::vector<PolymorphicValue> inputs;
-      inputs.reserve(def->inputs().size());
-      for (auto i : def->inputs()) {
-        const auto& eval_i = evaluateHelper(i, known_values);
-        if (!eval_i.hasValue()) {
-          return null_;
-        }
-        inputs.emplace_back(eval_i);
-      }
-      auto outputs = def->evaluate(*this, inputs);
+      auto outputs = def->evaluate(*this, known_values);
       for (auto i : c10::irange(def->outputs().size())) {
         known_values[def->output(i)] = std::move(outputs[i]);
       }
--- a/csrc/expr_evaluator.cpp
+++ b/csrc/expr_evaluator.cpp
@@ -199,8 +199,9 @@ PolymorphicValue ExpressionEvaluator::evaluate(const Val* value) const {
   return evaluateHelper(value, known_values);
 }
 
-PolymorphicValue ExpressionEvaluator::evaluate(const Val* value, 
-  std::unordered_map<const Val*, PolymorphicValue>& known_values) const {
+PolymorphicValue ExpressionEvaluator::evaluate(
+    const Val* value,
+    std::unordered_map<const Val*, PolymorphicValue>& known_values) const {
   return evaluateHelper(value, known_values);
 }
 
--- a/csrc/expr_evaluator.cpp
+++ b/csrc/expr_evaluator.cpp
@@ -191,21 +191,15 @@ const PolymorphicValue& ExpressionEvaluator::evaluate(ParallelType pt) {
 }
 
 const PolymorphicValue& ExpressionEvaluator::evaluate(const Val* value) {
-  return evaluateHelper(value, known_values_);
+  return evaluate(value, known_values_);
 }
 
 PolymorphicValue ExpressionEvaluator::evaluate(const Val* value) const {
   std::unordered_map<const Val*, PolymorphicValue> known_values;
-  return evaluateHelper(value, known_values);
+  return evaluate(value, known_values);
 }
 
-PolymorphicValue ExpressionEvaluator::evaluate(
-    const Val* value,
-    std::unordered_map<const Val*, PolymorphicValue>& known_values) const {
-  return evaluateHelper(value, known_values);
-}
-
-const PolymorphicValue& ExpressionEvaluator::evaluateHelper(
+const PolymorphicValue& ExpressionEvaluator::evaluate(
     const Val* value,
     std::unordered_map<const Val*, PolymorphicValue>& known_values) const {
   if (precomputed_values_ && precomputed_values_->ready()) {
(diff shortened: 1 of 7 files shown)


Priya2698:
!build

Priya2698:
Tests are passing: https://gitlab-master.nvidia.com/dl/pytorch/fuser-gh-mirror/-/pipelines/13482167.

csrc/expr_evaluator.cpp
```diff
@@ -199,6 +199,12 @@ PolymorphicValue ExpressionEvaluator::evaluate(const Val* value) const {
   return evaluateHelper(value, known_values);
 }
 
+PolymorphicValue ExpressionEvaluator::evaluate(
```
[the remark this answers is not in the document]
Priya2698: Yes, that is equivalent so I am okay either way.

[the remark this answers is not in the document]
Priya2698: Done.

csrc/expr_evaluator.h
```diff
@@ -59,6 +59,11 @@ class ExpressionEvaluator {
   //! Try to evaluate a value using const evaluator ref
```
wujingyue, line 59:
```suggestion
  //! Try to evaluate a value without using known_values_.
```

csrc/ir/base_nodes.h
```diff
@@ -524,6 +524,10 @@ class NVF_API Expr : public Statement {
       const ExpressionEvaluator& ee,
       const std::vector<PolymorphicValue>& inputs) const;
 
+  virtual std::vector<PolymorphicValue> evaluate(
```
wujingyue, line 527: There should be a code comment here about why this version of `evaluate` (and thus the added complexity) is needed and how it's different from the version above.

csrc/ir/base_nodes.cpp
```diff
@@ -401,6 +401,21 @@ std::vector<PolymorphicValue> Expr::evaluate(
...
 
+std::vector<PolymorphicValue> Expr::evaluate(
+    const ExpressionEvaluator& ee,
+    std::unordered_map<const Val*, PolymorphicValue>& known_values) const {
+  std::vector<PolymorphicValue> expr_inputs;
+  expr_inputs.reserve(inputs().size());
+  for (auto inp : inputs()) {
+    const auto& eval_i = ee.evaluate(inp, known_values);
+    if (!eval_i.hasValue()) {
+      return {null_};
```
wujingyue, line 412: Expr::null_ is probably unnecessary complexity. IIRC, all `std::monostate`s are equal. So you can simply say `return {std::monostate{}}` here. **If** the overhead of constructing these `monostate`s is indeed significant, you can store it in a static local variable, so it's only constructed once.

test/test_evaluator.cpp
```diff
@@ -695,4 +695,33 @@ TEST_F(ExprEvalTest, SumDiv) {
...
+  auto t1 = at::randn({3, 2}, options);
+
+  ExpressionEvaluator evaluator;
+  evaluator.bind(tv0, t0);
+  evaluator.bind(tv1, t1);
+
+  at::Tensor out = evaluator.evaluate(tv2).as<at::Tensor>();
+
+  for (auto padded_in : tv2->definition()->inputs()) {
+    EXPECT_FALSE(evaluator.isKnown(padded_in));
```
wujingyue, line 721: Nice test!

csrc/ir/nodes.cpp
```diff
@@ -4459,14 +4459,16 @@ Val* CatOp::getPred(int input_idx) const {
...
-  std::vector<at::Tensor> in;
+    std::unordered_map<const Val*, PolymorphicValue>& known_values) const {
+  // CatOp is preceded by a PadOp internally.
+  // For ATen evaluation, directly compute the unpadded inputs.
+  std::vector<at::Tensor> unpadded_inputs;
   int64_t concat_dim = concatenatedDim();
-  for (auto i : c10::irange(inputs.size())) {
-    auto unpadded_inp = ee.evaluate(input(i)->definition()->input(0));
-    in.push_back(unpadded_inp.as<at::Tensor>());
+  for (auto inp : inputs()) {
```
wujingyue, line 4467:
```suggestion
  for (Val* inp : inputs()) {
```

csrc/ir/nodes.cpp
```diff
@@ -4459,14 +4459,16 @@ Val* CatOp::getPred(int input_idx) const {
 
 std::vector<PolymorphicValue> CatOp::evaluate(
     const ExpressionEvaluator& ee,
-    const std::vector<PolymorphicValue>& inputs) const {
-  std::vector<at::Tensor> in;
+    std::unordered_map<const Val*, PolymorphicValue>& known_values) const {
+  // CatOp is preceded by a PadOp internally.
+  // For ATen evaluation, directly compute the unpadded inputs.
+  std::vector<at::Tensor> unpadded_inputs;
```
wujingyue, line 4465:
```suggestion
  std::vector<at::Tensor> unpadded_inputs;
  unpadded_inputs.reserve(inputs().size());
```

I tend to call vector::reserve extensively to minimize space and time, probably a PTSD debugging a slowness that turned out to be simply vector resizing during push_back.

csrc/ir/base_nodes.cpp
zasdfgbnm, line 412: I think we can just return `{std::monostate{}}`. We do not want to have a `null_` member for each IR node.

csrc/expr_evaluator.cpp
```diff
@@ -213,16 +213,7 @@ const PolymorphicValue& ExpressionEvaluator::evaluateHelper(
   if (!maybe_concrete_value.get().hasValue()) {
     if (auto def = value->definition()) {
       FUSER_PERF_SCOPE("ExpressionEvaluator::evaluate");
-      std::vector<PolymorphicValue> inputs;
-      inputs.reserve(def->inputs().size());
-      for (auto i : def->inputs()) {
-        const auto& eval_i = evaluateHelper(i, known_values);
-        if (!eval_i.hasValue()) {
-          return null_;
```
zasdfgbnm, line 221: I think `ExpressionEvaluator::null_` is no longer used after this change. Could you remove it?

csrc/ir/nodes.cpp
```diff
@@ -4459,14 +4459,16 @@ Val* CatOp::getPred(int input_idx) const {
...
   int64_t concat_dim = concatenatedDim();
-  for (auto i : c10::irange(inputs.size())) {
-    auto unpadded_inp = ee.evaluate(input(i)->definition()->input(0));
-    in.push_back(unpadded_inp.as<at::Tensor>());
+  for (auto inp : inputs()) {
+    auto eval_i = ee.evaluate(inp->definition()->input(0), known_values);
```
zasdfgbnm, line 4468: nit: should we `NVF_ERROR` that the `inp->definition()` is a `PadOp`?

wujingyue (commented):
LGTM with comments

csrc/ir/nodes.cpp
Priya2698: Yes, would guard against future potential changes.

csrc/ir/base_nodes.h
Priya2698: Added. Should be more clear now.

Priya2698:
!build

csrc/ir/base_nodes.h
wujingyue: Can you fix the format of the comment? It shows some unnecessary indentation.

wujingyue (approved):
🚢

Priya2698:
!build

PR #1922, opened 2024-03-13
Use `getRaw` in `isExtentEqualToMaxParallelTypeExtent`

zasdfgbnm:
Because `get` will return things like `blockDim.x` instead of the actual extent unless it is constant:
https://github.com/NVIDIA/Fuser/blob/4621129399550b8a6f3994d5092ddd0d581a6dc0/csrc/parallel_dimension_map.cpp#L150-L156
which is not desired behavior we want here.


zasdfgbnm:
!build

drzejan2 (approved):
Thank you for preparing this change!

PR #1923, opened 2024-03-13
Baseline benchmarks for GeLU fwd, bwd and reduction.

Priya2698:
Part of resolving #1668.
Adds IObytes computation for bwd and bwd-reduction benchmarks for parity with nvFuser.


naoyam (approved):
LGTM

PR #1925, opened 2024-03-13
Multidevice staged reduction with inter/intra-device manual scheduling

samnordmann:
part 1 of https://github.com/NVIDIA/Fuser/pull/1853
see the description there for more details

This PR implements the first composition of **manual interdevice scheduling** and **manual intradevice scheduling.**

Let us focus on the test implemented in `test/test_multidevice_pipeline.cpp`. There, we consider a simple fusion which expresses a multi-stage reduction:
```
  int num_devices = communicator->size();
  
  TensorView* tv0 = makeConcreteTensor([num_devices, 8, 32]);
  TensorView* tv_out = sum(tv1, {0});
  fusion->addInput(tv0);
  fusion->addOutput(tv_out);
```

_(Note that, for now, we only support reduction which are either non-resharding (i.e. not reducing a sharded dimension) or resharding but affecting only one axis. On the contrary, as far as intra-device scheduling is concerned, Fuser typically requests consecutive reductions to be merged. This inconsistency is only superficial, as we can always translate one case to another by splitting/merging consecutive RedOps)_

We decide to manually shard the tensors on their outmost axis:

```
  std::vector<int64_t> devices(num_devices);
  std::iota(devices.begin(), devices.end(), 0);
  DeviceMesh mesh(devices);
  for (auto tv : {tv0, tv1, tv_out}) {
    tv->setDeviceMesh(mesh);
  }
  for (auto tv : {tv0, tv1}) {
  }
```
This constitutes our multidevice scheduling. Then, we test composing this interdevice scheduling with manual scheduling, where we try a typical fully featured reduction scheduling. Supporting this mode only required very minor changes.

We also add in the multidevice test the option to validate using prescribed tensors.


csrc/multidevice/executor.cpp
```diff
@@ -87,13 +87,14 @@ std::unordered_map<Val*, c10::IValue> MultiDeviceExecutor::allocateRecvBuffers(
     return {};
   }
 
-  FusionExecutorCache fec(std::move(fusion_copy), 0, false);
-  auto buffers = fec.allocOutputSpace(global_inputs_IValues);
+  FusionExecutor fe;
```
samnordmann, line 90: these changes will be overwritten by
https://github.com/NVIDIA/Fuser/pull/1900

test/multidevice.cpp
```diff
@@ -93,12 +93,13 @@ void CommunicationTest::resetDstBuffers() {
 // It compares the given (possibly sharded) output with the result of the Fusion
 // run on a single device with the given (possibly sharded) inputs
 void PipelineTest::validate() {
-  // execute the fusion on one device without pipeline scheduling
-  auto fusion_copy = std::make_unique<Fusion>(*runtime->completeFusion());
-  unshard(fusion_copy.get());
-  FusionExecutorCache unsharded_fec(std::move(fusion_copy));
-  auto ref_unsharded_outputs =
-      unsharded_fec.runFusionWithInputs(unsharded_inputs);
+  if (ref_unsharded_outputs.empty()) {
```
samnordmann, line 96: https://github.com/NVIDIA/Fuser/pull/1853#discussion_r1523694537

Your comment makes a lot of sense. 
Checking if it's empty is the way to signal that we want to validate with prescribed tensors. Do you suggest having an explicit flag?

test/test_multidevice_pipeline.cpp
```diff
@@ -304,6 +304,116 @@ INSTANTIATE_TEST_SUITE_P(
...
+    tv->axis(0)->parallelize(ParallelType::DIDx);
+  }
+
+  // Intra-device reduction scheduling for the first reduction:
+  switch (scheduling_mode) {
+    case SchedulingMode::noIntraDeviceScheduling:
+      break;
+    case SchedulingMode::manualScheduling: {
+      // clang-format off
+      // inspired from NVFuserTest.FusionReduction1_CUDA
```
samnordmann, line 367: https://github.com/NVIDIA/Fuser/pull/1853#discussion_r1523687892

test/test_multidevice_pipeline.cpp
```diff
@@ -304,6 +304,122 @@ INSTANTIATE_TEST_SUITE_P(
...
+  }
+
+  // Intra-device reduction scheduling for the first reduction:
+  switch (scheduling_mode) {
+    case SchedulingMode::noIntraDeviceScheduling:
+      break;
+    case SchedulingMode::manualScheduling: {
+      // inspired from NVFuserTest.FusionReduction1_CUDA
+      // tv0[I0{A}, I1{B}, I2{C}]
+      tv1->split(2, 128);
```
cowanmeg, line 368: Axis 2 is size 32 in the test, should it test something greater like 256?

csrc/multidevice/utils.cpp
```diff
@@ -38,7 +38,9 @@ bool isSharded(TensorView* tv) {
   // allow split/merge if tv is sharded.
   auto tensor_sharded = !is_sharded.empty() && is_sharded.at(0);
   NVF_ERROR(
-      !tensor_sharded || tv->getMaybeRFactorDomain() == tv->getLeafDomain());
+      !tensor_sharded ||
+      (TensorDomain::noReductions(tv->getMaybeRFactorDomain()).at(0) ==
+       TensorDomain::noReductions(tv->getLeafDomain()).at(0)));
```
wujingyue, line 43: This is probably subjective. I think it's more intuitive to say

```
if (tensor_sharded) {
  NVF_ERROR(...);
}
```

wujingyue, line 43: Can you add a code comment? I guess you are trying to make sure all DID transforms are between root and rfactor, but I (and probably other readers) am not sure.

test/multidevice.h
```diff
@@ -99,7 +99,11 @@ class PipelineTest : public MultiDeviceTest {
   std::vector<c10::IValue> inputs;
   std::vector<c10::IValue> unsharded_inputs;
   std::vector<at::Tensor> outputs;
+  std::vector<at::Tensor> ref_unsharded_outputs;
   MultiDeviceExecutorParams multi_device_executor_params;
+  // if true, the validation is performed with prescribed reference data stored
+  // in ref_unsharded_outputs. Otherwise, this data is computed automatically
+  bool validate_with_prescribed_values = false;
```
wujingyue, line 106: How do I reason about these "global" (to the test) variables? They can be modified and read anywhere in the test. What are the common recipes?

test/multidevice.h
```diff
@@ -99,7 +99,11 @@ class PipelineTest : public MultiDeviceTest {
   std::vector<c10::IValue> inputs;
   std::vector<c10::IValue> unsharded_inputs;
   std::vector<at::Tensor> outputs;
+  std::vector<at::Tensor> ref_unsharded_outputs;
   MultiDeviceExecutorParams multi_device_executor_params;
+  // if true, the validation is performed with prescribed reference data stored
+  // in ref_unsharded_outputs. Otherwise, this data is computed automatically
```
wujingyue, line 105: What does "this data" refer to?

test/test_multidevice_pipeline.cpp
```diff
@@ -304,6 +304,122 @@ INSTANTIATE_TEST_SUITE_P(
         ::testing::Values(true),
         ::testing::Bool()));
 
+enum class SchedulingMode {
+  noIntraDeviceScheduling,
+  manualScheduling,
```
wujingyue, line 309:
```suggestion
  InterDeviceOnly,
  Manual,
```

For conciseness. "Scheduling" is unnecessary because the enum class will be used as `SchedulingMode::xyz`. I prefer `InterDeviceOnly` because it's positive and direct. (Of course, let me know if I misunderstand the meanings here).

test/test_multidevice_pipeline.cpp
```diff
@@ -304,6 +304,122 @@ INSTANTIATE_TEST_SUITE_P(
...
+};
+
+std::ostream& operator<<(std::ostream& out, const SchedulingMode& mode) {
```
wujingyue, line 312: Sounds like a useful utility that should belong to csrc/type.h?

test/test_multidevice_pipeline.cpp
```diff
@@ -304,6 +304,122 @@ INSTANTIATE_TEST_SUITE_P(
...
+      NVF_ERROR(false);
+  }
+  return out;
+}
+
+using PipelineTestStagedReductionParams = std::tuple<SchedulingMode>;
+class PipelineTestStagedReduction
+    : public PipelineTest,
+      public ::testing::WithParamInterface<PipelineTestStagedReductionParams> {
+};
```
wujingyue, line 328:
```suggestion
class PipelineTestStagedReduction
    : public PipelineTest,
      public ::testing::WithParamInterface<SchedulingMode> {
};
```

Is there a good reason for a single-element tuple (and therefore indirection)?

test/test_multidevice_pipeline.cpp
```diff
@@ -304,6 +304,122 @@ INSTANTIATE_TEST_SUITE_P(
...
+  ref_unsharded_outputs = {at::sum(
+      unsharded_inputs.at(0).toTensor(), at::OptionalIntArrayRef({0, 2}))};
+
+  executeAndValidate();
+}
+
+INSTANTIATE_TEST_SUITE_P(
+    SchedulingModes,
+    PipelineTestStagedReduction,
+    ::testing::Combine(::testing::Values(
```
wujingyue, line 419: Why testing::Combine?

csrc/tensor_view.cpp
```diff
@@ -937,6 +937,8 @@ TensorView* TensorView::rFactor(const std::vector<int>& axes) {
   TensorView* producer =
       IrBuilder::create<TensorView>(producer_domain, getDataType().value());
 
+  producer->setDeviceMesh(mesh_);
```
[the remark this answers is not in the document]
samnordmann: Consider the fusion:
```
DeviceMesh mesh({0,1});

TensorView* tv0 = makeConcreteTensor(2);
TensorView* tv1 = sum(tv0, {0});

tv0->setDeviceMesh(mesh);
tv1->setDeviceMesh(mesh);

tv1->split(0, 4);
TensorView* tv2 = tv1->rFactor({0});
```
Then, `tv2`'s device mesh should be set to be the one of `tv1`.

This is why we need this line

csrc/multidevice/utils.cpp
samnordmann: overwritten by [Sharded tensor metadata](https://github.com/NVIDIA/Fuser/commit/26c4ca69d584559231467359188915f86c2556ac)

samnordmann: overwritten by [Sharded tensor metadata](https://github.com/NVIDIA/Fuser/commit/26c4ca69d584559231467359188915f86c2556ac)

The idea here is to forbid splt/merge on sharded axis

test/test_multidevice_pipeline.cpp
wujingyue: Don't you expect it to be shortly used elsewhere? For those types, I tend to move them to a common header, but I also don't mind keeping them local for now and refactor later.

csrc/tensor_view.cpp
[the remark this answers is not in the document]
naoyam: I think this is fine, but I wonder if we could do some basic inference to automatically set a proper device mesh.

naoyam (commented):
LGTM. I'll let @wujingyue give a stamp.

test/test_multidevice_pipeline.cpp
samnordmann: I don't know exactly how it could be used later, so I'd rather keep it here and refactor when needed

csrc/tensor_view.cpp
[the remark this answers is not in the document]
naoyam: Yes, it's not about this PR, but I was wondering if we could do something to make it more robust rather than manually annotating all tenors individually. Think of `parallelizeAllLike`.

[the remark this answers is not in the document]
samnordmann: it turns out we have a function [shardAllLike](https://github.com/NVIDIA/Fuser/blob/b9d3c116aa46dff33bcec3a00fbd2d8c7b2d0ffe/csrc/multidevice/utils.cpp#L115), but I'm not sure how it could help here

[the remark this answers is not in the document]
naoyam: Nothing urgent. I'm not sure what automation would be the most desirable, but I believe it would become gradually more obvious as we gain more experiences with multi-GPU examples.

tests/cpp/test_multidevice_pipeline.cpp
```diff
@@ -304,6 +304,117 @@ INSTANTIATE_TEST_SUITE_P(
...
+
+  // Intra-device reduction scheduling for the first reduction:
+  switch (scheduling_mode) {
+    case SchedulingMode::InterDeviceOnly:
+      break;
+    case SchedulingMode::Manual: {
+      // inspired from NVFuserTest.FusionReduction1_CUDA
+      // tv0[I0{A}, I1{B}, I2{C}]
+      tv1->split(2, 32);
+      // tv1[I0{A}, I1{B}, R2o{C/128}, R2i{128}] = tv0[I0{A}, I1{B}, I2{C}]
```
wujingyue, line 366: IIUC, the comment is not consistent with the code. I believe you intended to say
```suggestion
      // tv1[I0{A}, I1{B}, R2o{C/32}, R2i{32}] = tv0[I0{A}, I1{B}, I2{C}]
```
If so, we should fix other comments too.

tests/cpp/test_multidevice_pipeline.cpp
```diff
@@ -304,6 +304,117 @@ INSTANTIATE_TEST_SUITE_P(
...
+  }
+  return out;
+}
+class PipelineTestStagedReduction
+    : public PipelineTest,
+      public ::testing::WithParamInterface<SchedulingMode> {};
+
+// 1D staged reduction
+// Inputs: X[A,B,C]
+TEST_P(PipelineTestStagedReduction, staged_reduction) {
```
wujingyue, line 329: Nit: style convention. 

```suggestion
TEST_P(PipelineTestStagedReduction, StagedReduction) {
```

csrc/tensor_view.cpp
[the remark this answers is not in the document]
wujingyue: I now understand it's being used by the manual scheduling test. My concern was that: we have too many scheduling methods creating TensorView (seven occurrences in tensor_view.cpp alone). It's weird to have some but not all of them propagating the device mesh. A bigger question is: at the end of the day, it's unclear to me whether they should be responsible to propagate the device mesh or their users should. 

I'll review that test to see if there's a way to around.

tests/cpp/test_multidevice_pipeline.cpp
samnordmann: right

csrc/tensor_view.cpp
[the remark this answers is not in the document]
samnordmann: I see. In my opinion it depends on the considered function that creates the tensor. In my opinion it makes sense to add this change in factor because:
1) rFactor is really a scheduling tool, that replay the scheduling of the argument to the Tensor it cresates -- and the device mesh is an intrinsic part of that
2) The intention of rFactor is to divide a reduction into several steps, and so we want the intermediate tensor to be sharded on the save device mesh. In other word, in my opinion, in all cases where rFactor is used we will want to set the device mesh this way.

Those two above points might be different for other TensorView constructors.

This being said, it doesn't really matter for me to set manually the rFactor domain in the test. So let me know, if it is a concern let me simply move this to the tests -- it will just result in more verbose tests but that's not a problem for me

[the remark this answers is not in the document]
wujingyue: The current change is good enough! And I don't want to bikeshed too much before we have a clear plan for our Python API. 

I agree with both your points. My concern was about who should be responsible to propagate/infer the device mesh -- rFactor is likely too low-level an interface for **that** purpose, and thus my hesitation.

test/multidevice.h
wujingyue: > The flag is set in the test body and read in the method validate(). 

That's the problem I want to avoid. Instead of communicating with `validate` via some global-to-class variables, I think it's better to pass the same information in via the parameters of `validate`. This makes information flow more explicit. I'll read the test more closely to see whether it's possible to do so.

csrc/tensor_view.cpp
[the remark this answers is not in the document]
samnordmann: ok so, if that's ok with you, let me resolve this conversation and keep this point as a future discussion

test/multidevice.h
samnordmann: ok, I removed `validate_with_prescribed_values` as a class member and now pass this argument around the method

tests/cpp/multidevice.cpp
```diff
@@ -170,12 +166,12 @@ void PipelineTest::execute() {
...
   }
+  outputs = runtime->runWithInput(inputs);
   if (debug_print) {
     if (!communicator->deviceId()) {
       runtime->print();
     }
   }
-  outputs = runtime->runWithInput(inputs);
 
   if (debug_print) {
```
wujingyue, line 176:
```suggestion
```

wujingyue: The change suggestion doesn't work very well in this case. I meant to merge the two `if (debug_print)`.

tests/cpp/multidevice.h
```diff
@@ -88,17 +88,21 @@ class CommunicationTest
 class PipelineTest : public MultiDeviceTest {
  protected:
   void SetUp() override;
-  void validate();
+  // Run the multidevice fusion and stores the (sharded) results in "outputs"
   void execute();
-  void executeAndValidate() {
+  // Utility function used for validation in the tests. It compares the (sharded) outputs with ref_unsharded_outputs.
+  // if validate_with_prescribed_values= true, ref_unsharded_outputs is assumed to be set manually in the test body. Otherwise, ref_unsharded_outputs is computed by running a Fusion on a single device with the unsharded_inputs
```
wujingyue, line 94: Thanks for the comments! 

```suggestion
  // if validate_with_prescribed_values is true, ref_unsharded_outputs is assumed to be set manually in the test body. Otherwise, ref_unsharded_outputs is computed by running a Fusion on a single device with the unsharded_inputs
```

wujingyue (approved):
Great work! 🚢

samnordmann:
!build

PR #1927, opened 2024-03-13
matmul via ATen: converting enable option to diable

protonu:
Our default behavior for the Matmul API in design (in the stacked PR: ) is to evaluate via ATen/Expression  Evaluator (EE). To make the stacked a bit cleaner and to express the default behavior better, we have a  disable flag to to stop execution via EE. Thus, in the usual case it'll execute via EE.

I changed the constructor of our NVFuserTests to disable this options so that we can keep testing Fuser's matmul scheduling and lowering.


protonu:
!build

test/utils.h
```diff
@@ -441,6 +442,11 @@ class NVFuserTest : public ::testing::Test {
     // random seed. Otherwise, use zero. If a test fails, this seed will be
     // printed.
     std::srand(getCRandomSeed());
+
+    // For the NVFuser tests, we generally want to test Fuser's matmul
+    // scheduling and lowering. Thus disabling execution via expression
+    // evaluator.
+    DisableOptionsGuard::getCurOptions().set(DisableOption::MatmulExprEval);
```
wujingyue, line 449: This creates a divergence between running nvFuser through bin/nvfuser_tests and through python_tests or just an arbitrary Python script. 

If reviewers agree to make MatmulExprEval opt out, I'd only disable it in test fixtures that we want it disabled. If there turn out to be too many fixtures to disable, that's probably an indicator that it should be opt-in instead :)

wujingyue, line 449: Regardless of my other comment, we should save the previous state and recover that in TearDown.

wujingyue:
I'd like to hear opinions from @jacobhinkle and @zasdfgbnm , who do most of the real matmul codegen work. How would this PR affect your flow?

test/utils.h
jjsjann123: vote on @jacobhinkle 's suggestion. Let's use a guard and not leave a corrupted flag after tests.

jjsjann123 (commented):
How did we pass CI? I thought we have python matmul in CI that's supposed to also have the flag disabled as well.

And we also need to do the same option guard there in python, similar to how @jacobhinkle suggested in the cpp tests here.

test/utils.h
jacobhinkle: Could we instead do this just for the Matmul test subclass?

jacobhinkle:
I would just disable it in the matmul test class, not in the main NvFuserTest class. Also use the DisableGuard by instantiating it instead of using only the static method. Otherwise I think the idea is sound.

test/utils.h
wujingyue: > Create a DisableGuard object that will be deconstructed in Teardown(). You can use unique_ptr to do this safely. That way it will automatically restore settings.

That's magic I'm not aware of! For my education, how do you make Setup() create an object that is automatically destructed in TearDown()? (I only know the manual way: https://github.com/NVIDIA/Fuser/blob/195a3f75d728739d8b386520e00540d59766ce3e/test/test_gpu_transpose.cpp#L45-L57)

protonu:
Thanks for the comments everyone. I went with Jacob's suggestion of creating a unique_ptr, turning on the disable flag in the SetUp function of the required tests, and letting the pointer go out of scope to handle cleanup.

All the C++ tests pass on my system, but this PR I realized isn't ready. The Python tests also need to turn on the disable flag. I'll add that in another commit in this PR.

test/utils.h
wujingyue: Never mind -- I figured out what you mean and created https://github.com/NVIDIA/Fuser/pull/1940 to simplify. unique_ptr is not necessary and we should do things [in the constructor](https://github.com/google/googletest/blob/main/docs/faq.md#should-i-use-the-constructordestructor-of-the-test-fixture-or-setupteardown-ctorvssetup) whenever possible.

jjsjann123: ^^^ Thanks I actually thought `setup/teardown` is the preferred way :exploding_head:

jacobhinkle: > That's magic I'm not aware of! For my education, how do you make Setup() create an object that is automatically destructed in TearDown()? (I only know the manual way:

Sorry for the slow response. Yes, I meant the manual way of putting `options_guard_ptr_ = nullptr;` in `Teardown()`;

jacobhinkle: > would each TEST be considered to _always_ start from a fresh environment?

I believe so, yes. Are you thinking that there might be a case where we have two nested guards and we lose the middle state when we destroy the inner-most guard? In that case, the inner-most guard will be restoring the state it was initialized with, which should be whatever it was at initialization. This can only go wrong if an outer guard modifies its state during the lifetime of an inner guard, which will result in losing the updates when the inner guard is destroyed. Confusing enough?

wujingyue: > would each TEST be considered to always start from a fresh environment?

According to [this section](https://github.com/google/googletest/blob/main/docs/primer.md#test-fixtures-using-the-same-data-configuration-for-multiple-tests-same-data-multiple-tests):

> For each test defined with TEST_F(), GoogleTest will create a fresh test fixture at runtime, immediately initialize it via SetUp(), run the test, clean up by calling TearDown(), and then delete the test fixture.

so constructor should be as fresh as setup.

jjsjann123: ^^^ Thanks. I think I was just paranoid.

The doc JIngyue reference does seem to read like each test is an independent piece and ctor/setup considered as the singularity point where everything starts. So I shouldn't worry about restoring anything beyond that point..

jacobhinkle: Yes, thank you @wujingyue for pointing to that doc. I had not made any assumptions on when the test object was created but it seems like it's actually preferred to use the ctor/dtor unless there is a good reason not to. That should work just fine with the guard. We can just default-initialize the options guard then set the option in the ctor using `DisableOptionsGuard::getCurOptions().set(DisableOption::MatmulExprEval);`

protonu:
!build

protonu:
Thanks @jacobhinkle. @wujingyue and @jjsjann123. Based on Jacob's and Jigyue's comments, I created a member variable for the guard and disabled in the constructor. The unique_ptr going out of scope should handle cleanup of the guard. Please let me know what you think.
Please note that with a previous PR (checked in), Python tests for temporary matmul APIs which needed the Fuser's scheduler/lowering have been removed.

tests/cpp/test_combine_mul_sum.cpp
```diff
@@ -54,6 +60,8 @@ class CombineMulSumAsMmaTest : public NVFuserTest {
     }
     NVFuserTest::SetUp();
   }
+  std::unique_ptr<DisableOptionsGuard> disable_options_ptr =
+      std::make_unique<DisableOptionsGuard>();
```
wujingyue, line 64: Similar to https://github.com/NVIDIA/Fuser/pull/1940, we don't need a unique ptr. 

```suggestion
 private:
  DisableOptionsGuard opt_guard_;
```

tests/cpp/test_combine_mul_sum.cpp
```diff
@@ -38,6 +39,11 @@
 namespace nvfuser {
 
 class CombineMulSumAsMmaTest : public NVFuserTest {
+ protected:
+  CombineMulSumAsMmaTest() {
+    disable_options_ptr->getCurOptions().set(DisableOption::MatmulExprEval);
```
wujingyue, line 44:
```suggestion
    DisableOptionsGuard::getCurOptions().set(DisableOption::MatmulExprEval);
```
because it's a static method. It's technically possible to use it on an instance, but that gives readers the wrong impression.

tests/cpp/test_combine_mul_sum.cpp
wujingyue: The same comment goes to other uses of DisableOptionsGuard in this PR.

wujingyue (commented):
LGTM with comments. I'll take a final look before stamping it.

tests/cpp/test_matmul_scheduler.cpp
```diff
@@ -43,11 +46,16 @@ class PrecisionParametrizedTest
...
+  PrecisionParametrizedTest() : optimization_guard_(false) {
+    DisableOptionsGuard::getCurOptions().set(DisableOption::MatmulExprEval);
+  }
 
  private:
   preseg_passes::OptimizationPassGuard<preseg_passes::AllocationDomainPass>
       optimization_guard_;
+
+ private:
+  DisableOptionsGuard opt_guard_;
```
wujingyue, line 58:
```suggestion
  DisableOptionsGuard option_guard_;
```

There's already a `private` before this. Also, I'd rename this to option_guard_ to distinguish it from optimization_guard_.

wujingyue (approved):
:shipit:

tests/cpp/test_gpu_tensorcore.cpp
```diff
@@ -52,11 +52,22 @@
...
 namespace nvfuser {
+namespace {
+class GPUTTensorCoreTest : public NVFuserTest {
+ protected:
+  GPUTTensorCoreTest() {
+    DisableOptionsGuard::getCurOptions().set(DisableOption::MatmulExprEval);
+  }
+
+ private:
+  DisableOptionsGuard opt_guard_;
```
jacobhinkle, line 63: A comment might be useful here, because at a glance it looks like this is an unused variable. Instead, it's important that this is default initialized before the constructor body is run so that `DisableOptionsGuard::getCurOptions()` refers to the new options and not the previous ones, which are held in `options_guard_`.

jacobhinkle (approved):
A comment would be useful I think, but otherwise LGTM.

protonu:
!build

protonu:
!build

protonu:
!build

PR #1929, opened 2024-03-13
MultiDeviceExecutor: use at::Tensor instead of c10::IValue

samnordmann:
Change suggested by @wujingyue in https://github.com/NVIDIA/Fuser/pull/1900#discussion_r1523675221

I have one doubt though: for now we don't support Scalars input, but once we do, will it be ok with this patch?
in general, what is the benefit of using IValue over Tensor?

### DIFF
--- a/csrc/multidevice/executor.cpp
+++ b/csrc/multidevice/executor.cpp
@@ -113,7 +113,7 @@ void MultiDeviceExecutor::postKernel(SegmentedGroup* group) {
         input,
         " for handling group ",
         toString(group));
-    NVF_ERROR(val_to_IValue_.at(input).isTensor());
+    NVF_ERROR(val_to_IValue_.at(input).numel());
     group_input_IValues.push_back(val_to_IValue_.at(input));
   }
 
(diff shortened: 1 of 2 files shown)


wujingyue:
> I have one doubt though: for now we don't support Scalars input, but once we do, will it be ok with this patch? in general, what is the benefit of using IValue over Tensor?

Good question. @zasdfgbnm, wdyt? I wonder whether/how PolymorphicValue come into play here.

zasdfgbnm:
> Good question. @zasdfgbnm, wdyt? I wonder whether/how PolymorphicValue come into play here.

Regarding `PolymorphicValue` vs `IValue`, I think we should always prefer `PolymorphicValue`. `IValue` should only be used when we are interfacing with PyTorch. `PolymorphicValue` is a class defined in our own repository and we have better control of.

wujingyue:
> Regarding `PolymorphicValue` vs `IValue`, I think we should always prefer `PolymorphicValue`. `IValue` should only be used when we are interfacing with PyTorch. `PolymorphicValue` is a class defined in our own repository and we have better control of.

Thanks for the input! 

Anyhow, it's my mistake missing 
https://github.com/NVIDIA/Fuser/pull/1929/files#diff-d783a79fdc189059065a81ba9acb6d6c3d3ecc5c70188706febc97d7fea5aa36R206. I thought val_to_IValue_ contains only the allocated outputs. But this line clearly indicates it can also contain inputs and nvFuser does support scalar inputs.

Issue #1812, opened 2024-02-21
Skip compiling fusion segments that are marked for EE


Priya2698:
PR #1743 ran into errors in `GpuLower::analysis`. In the short term, the workaround is to skip Mma specific validation in `analysis`.

Ideally if intend to use the EE, then we should not be compiling a fusion / ~launching a kernel~ (already skipped after #1723) . This will require us to modify some downstream calls and `FusionExecutor` interface to integrate using EE alongside nvFuser kernels seamlessly. 

CC: @jacobhinkle @wujingyue

wujingyue:
(Nit: I modified the title to focus on the problem not a solution. There could be other valid solutions -- I don't want to discount them immediately)

wujingyue:
FYI, kernel launching is already skipped after https://github.com/NVIDIA/Fuser/pull/1723. It's the unnecessary compilation that we want to avoid here.

PR #1930, opened 2024-03-13
Skip compiling fusion segments when using expression evaluator

Priya2698:
This PR resolves Issue #1812.
1. Removes the workarounds added to support ATen evaluation for matmuls in PR #1775 (in `predicate_elimination.cpp` and `validation.cpp` (see comments in the PR)
2. `FusionExecutor::fusion_` is initialized only when compilation is skipped. So only one of `lowered_` or `fusion_` is non-null.
3. `FusionKernelRuntime/FusionExecutorCache/FusionExecutor::isCompiled()` indicates whether compilation was attempted (either marked for EE or via nvFuser).

### DIFF
--- a/csrc/device_lower/analysis/predicate_elimination.cpp
+++ b/csrc/device_lower/analysis/predicate_elimination.cpp
@@ -890,9 +890,7 @@ class PredicateChcker : public IterVisitor {
 } // namespace
 
 PredicateElimination::PredicateElimination(Fusion* fusion) {
-  // To avoid errors in analysis when using ATen evaluation for matmul, only use
-  // outputs that require codegen. See PR # 1775 and Issue #1812
-  traverseTo(lower_utils::getFusionOutputsRequiringCodegen(fusion));
+  traverseTo(fusion->outputs());
 }
 
 bool PredicateElimination::needsPredicate(Expr* expr) const {
--- a/csrc/device_lower/validation.cpp
+++ b/csrc/device_lower/validation.cpp
@@ -1029,12 +1029,7 @@ void validateSizeMemoryOp(LoadStoreOp* ldst) {
 //! Validate data format and GPU arch compatibility of scheduled
 //!  mma operators on the fusion.
 void validateMma(Fusion* fusion) {
-  // To avoid errors in analysis when using ATen evaluation for matmul, only
-  // validate expressions that require codegen. See PR # 1775 and Issue #1812
-  std::vector<Val*> outs_requiring_codegen =
-      lower_utils::getFusionOutputsRequiringCodegen(fusion);
-  auto exprs = StmtSort::getExprsBetween(
-      GpuLower::current()->allKnownVals(), outs_requiring_codegen);
+  auto exprs = StmtSort::getExprs(fusion);
 
   for (auto expr : exprs) {
     if (auto mma = dynamic_cast<MmaOp*>(expr)) {
--- a/csrc/executor.h
+++ b/csrc/executor.h
@@ -172,7 +172,10 @@ class FusionExecutor : public NonCopyable {
 
   // Function to query whether compilation was attempted for a `FusionExecutor`
   bool isCompiled() const {
-    return hasCompiledKernel() || isCompilationSkipped();
+    // Check atmost one of fusion_ and lowered_ is null.
+    NVF_ERROR(fusion_  == nullptr || !lowered_, "Expected GPU lowering to be skipped.");
+    NVF_ERROR(lowered_ == nullptr || !fusion_, "fusion_ should only be initializaed when using expression evaluator.");
+    return fusion_ || lowered_;
   };
 
   // function to query whether a `FusionExecutor` has a compiled kernel to
@@ -180,6 +183,7 @@ class FusionExecutor : public NonCopyable {
   bool hasCompiledKernel() const {
     if (compiled_kernel_ != nullptr) {
       NVF_ERROR(compiled_kernel_->function != nullptr);
+      NVF_ERROR(!fusion_, "fusion_ should only be initializaed when using expression evaluator.");
     }
     return validKernelId() && lowered_ && compiled_kernel_ != nullptr;
   };
@@ -415,7 +419,7 @@ class FusionExecutor : public NonCopyable {
       int64_t group_id);
 
   //! Check if compilation was skipped (fusion segment marked for EE).
-  bool isCompilationSkipped() const {
+  bool isExprEval() const {
     NVF_ERROR(
         fusion_ == nullptr || !lowered_,
         "Expected GPU lowering to be skipped.");
--- a/csrc/executor.h
+++ b/csrc/executor.h
@@ -412,7 +412,7 @@ class FusionExecutor : public NonCopyable {
   //! Check if compilation was skipped (fusion segment marked for EE).
   bool isCompilationSkipped() const{
     if (!fusion_) {
-      NVF_ERROR(!lowered_, "Expected a lowered kernel to be initialized.");
+      NVF_ERROR(lowered_, "Expected a lowered kernel to be initialized.");
       return false;
     }
     NVF_ERROR(!lowered_, "Expected GPU lowering to be skipped.");
--- a/csrc/executor.h
+++ b/csrc/executor.h
@@ -411,9 +411,7 @@ class FusionExecutor : public NonCopyable {
 
   //! Check if compilation was skipped (fusion segment marked for EE).
   bool isCompilationSkipped() const {
-    if (fusion_ != nullptr) {
-      NVF_ERROR(!lowered_, "Expected GPU lowering to be skipped.");
-    }
+    NVF_ERROR(fusion_ == nullptr || !lowered_, "Expected GPU lowering to be skipped.");
     return fusion_ != nullptr;
   }
 
--- a/csrc/executor.h
+++ b/csrc/executor.h
@@ -206,7 +206,7 @@ class FusionExecutor : public NonCopyable {
   }
 
   Fusion* fusion() const {
-    NVF_ERROR(lowered_ || fusion_);
+    NVF_ERROR(lowered_ ^ fusion_);
     return lowered_ ? lowered_->kernel()->as<Fusion>() : fusion_.get();
   }
 
--- a/csrc/executor.h
+++ b/csrc/executor.h
@@ -123,9 +123,11 @@ class FusionExecutor : public NonCopyable {
         concrete_id);
   }
 
-  std::vector<at::Tensor> runAtenFusion(
+  //! Computes fusion outputs through expression evaluator.
+  std::vector<at::Tensor> evaluateFusionOutputs(
     KernelArgumentHolder& args,
-    std::vector<at::Tensor> outputs);
+    std::vector<at::Tensor> outputs,
+    ExpressionEvaluator& expr_eval);
 
   NVF_API std::vector<at::Tensor> runFusion(
       KernelArgumentHolder& args,
@@ -407,6 +409,16 @@ class FusionExecutor : public NonCopyable {
       int64_t runtime_id,
       int64_t group_id);
 
+  //! Check if compilation was skipped (fusion segment marked for EE).
+  bool isCompilationSkipped() {
+    if (!fusion_) {
+      NVF_ERROR(!lowered_, "Expected a lowered kernel to be initialized.");
+      return false;
+    }
+    NVF_ERROR(!lowered_, "Expected GPU lowering to be skipped.");
+    return true;
+  }
+
  private:
   LaunchParams computeLaunchParams(
       const LaunchParams& launch_constraints,
@@ -498,15 +510,6 @@ class FusionExecutor : public NonCopyable {
 
   //! Clear the cached properties of the compiled kernel
   void resetCompiledKernelProperties();
-
-  bool isCompilationSkipped() {
-    if (!fusion_) {
-      NVF_ERROR(!lowered_, "Expected a lowered kernel to be initialized.");
-      return false;
-    }
-    NVF_ERROR(!lowered_, "Expected GPU lowering to be skipped.");
-    return true;
-  }
 
  private:
   CompileOptions options_;
--- a/csrc/executor.h
+++ b/csrc/executor.h
@@ -123,6 +123,10 @@ class FusionExecutor : public NonCopyable {
         concrete_id);
   }
 
+  std::vector<at::Tensor> runAtenFusion(
+    KernelArgumentHolder& args,
+    std::vector<at::Tensor> outputs);
+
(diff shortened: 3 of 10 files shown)


Priya2698:
!build

Priya2698:
!build

Priya2698:
!build

Priya2698:
!build

Priya2698:
!build

Priya2698:
!build

Priya2698:
!build

csrc/device_lower/analysis/predicate_elimination.cpp
```diff
@@ -890,9 +890,7 @@ class PredicateChcker : public IterVisitor {
 } // namespace
 
 PredicateElimination::PredicateElimination(Fusion* fusion) {
```
Priya2698, line 892: This was a WAR. Removing now, since we are skipping compilation.

csrc/device_lower/validation.cpp
```diff
@@ -1030,12 +1030,7 @@ void validateSizeMemoryOp(LoadStoreOp* ldst) {
 //! Validate data format and GPU arch compatibility of scheduled
 //!  mma operators on the fusion.
 void validateMma(Fusion* fusion) {
```
Priya2698, line 1032: This was a WAR. Removing now, since we are skipping compilation.

csrc/executor.h
```diff
@@ -199,6 +205,11 @@ class FusionExecutor : public NonCopyable {
     return lowered_->kernel();
   }
 
+  Fusion* fusion() const {
+    NVF_ERROR(lowered_ || fusion_);
```
wujingyue, line 209:
```suggestion
    NVF_ERROR(lowered_ ^ fusion_);
```

Xor or Or?

csrc/serde/fusion_cache.fbs
```diff
@@ -345,6 +345,7 @@ table CudaKernel {
 
 // Each Fusion Executor maps to a lowered and compiled kernel.
 table FusionExecutor {
+  is_compilation_skipped : bool;
```
wujingyue, line 348: Why is this needed? isCompilationSkipped is a method that describes the state of the FusionExecutor. Making it a field in `table FusionExecutor` seems to suggest it's part of the state itself. cc @rdspring1

csrc/kernel_cache.cpp
```diff
@@ -1379,7 +1380,9 @@ std::pair<LaunchParams, CompileParams> FusionKernelRuntime::getKernelConfig(
 
   // Check that the heuristics are matched, in the case of segmented fusion
   NVF_ERROR(!sg || scheduler_entry->heuristic() == sg->heuristic());
-  NVF_ERROR(executors_.at(group_id).isCompiled());
+  NVF_ERROR(
+      executors_.at(group_id).isCompiled() ||
+      executors_.at(group_id).isCompilationSkipped());
```
wujingyue, line 1385: Why does this check exist in the first place? The function itself doesn't seems to use `fusion_` or `kernel_`. It seems to belong to something like FusionExecutor::runFusion.

tests/cpp/test_resize.cpp
```diff
@@ -1900,6 +1902,9 @@ TEST_F(ResizeTest, FusionSliceForNanoGPT2) {
 
 // C++ version of TestNvFuserFrontend.test_nanogpt_split_mha_linears
 TEST_F(ResizeTest, FusionSliceForNanoGPT3) {
+  preseg_passes::OptimizationPassGuard<preseg_passes::MarkAliasesPreparePass>
+      optimization_guard(false);
```
jjsjann123, line 1906: naive question, why are we skipping alias pass on this?

tests/cpp/test_gpu3.cpp
```diff
@@ -31,6 +31,8 @@
 #include <kernel_ir.h>
 #include <kernel_ir_dispatch.h>
 #include <ops/all_ops.h>
+#include <preseg_passes/mark_aliases_prepare.h>
+#include <preseg_passes/optimization_pass.h>
```
jjsjann123, line 35: nitpick: Don't see any reason to add these headers.

csrc/kernel_cache.h
```diff
@@ -124,7 +124,7 @@ class FusionKernelRuntime {
     std::lock_guard<std::mutex> guard(mutex_);
     return std::all_of(
         executors_.begin(), executors_.end(), [](const auto& executor) {
-          return executor.isCompiled();
+          return executor.isCompiled() || executor.isCompilationSkipped();
```
jjsjann123, line 127: nitpick: This feels a little bit confusing to me.

i.e. `FusionKernelRuntime::isCompiled` means something different from `FusionExecutor::isCompiled`. (similarly we have `FusionExecutorCache::isCompiled`.

We don't have to do it in this PR, but maybe another renaming PR to clean up the `isCompiled` vs `isCompilationSkipped` in `FusionExecutor`.

csrc/executor.h
```diff
@@ -398,6 +409,14 @@ class FusionExecutor : public NonCopyable {
       int64_t runtime_id,
       int64_t group_id);
 
+  //! Check if compilation was skipped (fusion segment marked for EE).
+  bool isCompilationSkipped() const {
+    if (fusion_ != nullptr) {
+      NVF_ERROR(!lowered_, "Expected GPU lowering to be skipped.");
+    }
```
jjsjann123, line 416:
```suggestion
    NVF_ERROR(fusion_ == nullptr || !lowered_, "Expected GPU lowering to be skipped.");
```

nitpicking is all I do.

csrc/kernel_cache.cpp
Priya2698: I only see `scheduler_entry->params()->cparams.index_type` being updated in `compileFusion`.
Since this function is only called here before `runFusion`, we can move the checks inside `runFusion`.

Only if this function is used in isolation (which should probably not happen) it can have `index_type` vary based on compilation status

tests/cpp/test_resize.cpp
Priya2698: I'll add a comment to explain.

This test is checking 
```
NVF_CHECK(
      !kernel->summary().has_cooperative_grid_reduction,
      "Grid sync should not be used as slicing input should avoid input caching");
```
If we don't skip aliasing, the compilation will be skipped, and this property cannot be verified.
So I am skipping this pass, so the test can actually verify this behavior.

tests/cpp/test_gpu3.cpp
Priya2698: Right, this is a remnant from previous changes. Thanks for catching!

csrc/kernel_cache.h
Priya2698: I should rename it to `FusionKernelRuntime::isCompiledOrCompilationSkipped`.
This function is used to check if compilation was already attempted which is true if either the executors are all compiled or compilation was skipped for them.

Priya2698: Any other naming suggestions for this function?

tests/cpp/test_resize.cpp
Priya2698: We could also modify the test itself, to have other operators to avoid the outputs being aliased.

csrc/kernel_cache.h
rdspring1: 1. isCompilationSkipped() --- Returns True when we use the Aten function and `FusionExecutor` is never compiled.
2. isCompiled() --- Returns False when `FusionExecutor` has yet to be compiled.

Maybe we should rename `isCompilationSkipped()` to `isAtenExecuted()` or `isEagerModeExecutor()`

Priya2698: I agree with Ryan's suggestion. We can keep `isCompiled` to denote if compilation was attempted -- either a kernel was compiled or we are using Aten evaluation. Additionally, `isCompilationSkipped` is renamed to `isEagerModeExecutor`.

In this naming scheme, `isCompiled` always means whether or not we attempted to compile a fusion.

Wdyt @jjsjann123?

rdspring1: When `isEagerModeExecutor` is `false`, we're using the nvfuser executor.

`isCompiled` implies that we have a cubin for the given fusion. There is an implicit assumption that the nvfuser executor is being used. 

I'd consider renaming `isCompiled` to `isCubinCompiled`, but leaving it alone is also fine.

Priya2698: Okay, so the renaming is 
1. `FusionExecutor::isCompiled` -> `FusionExecutor::isCubinCompiled`
2. `FusionExecutor::isCompilationSkipped` -> `FusionExecutor::isEagerModeExecutor`
3. `FusionKernelRuntime/FusionExecutorCache::isCompiled` -> no change. Returns true if either of the above are true indicating that compilation was attempted.

I'll go with this naming if it clears the confusion. @rdspring1 @jjsjann123

csrc/serde/fusion_cache.fbs
wujingyue: Sorry for the delay -- I missed the notification. 

I understood how it's used [here](https://github.com/NVIDIA/Fuser/pull/1930/files#diff-e72d395ea8a3ab3d2a909dfb3677beac1bbd89336b54ee858867f652b48be4f0R2339) to avoid unnecessary compilation. However, can we replace that check with a check on the fusion itself instead of introducing yet another flag? If works, this can avoid an additional state that has to be consistent with some existing state of the fusion.

csrc/kernel_cache.h
jjsjann123: sounds very reasonable to me.

Priya2698:
!build

Priya2698:
Tests pass here: https://gitlab-master.nvidia.com/dl/pytorch/fuser-gh-mirror/-/pipelines/14499669

csrc/executor.h
```diff
@@ -164,11 +170,26 @@ class FusionExecutor : public NonCopyable {
...
+    NVF_ERROR(
+        lowered_ == nullptr || !fusion_,
+        "fusion_ should only be initializaed when using expression evaluator.");
+    return fusion_ || lowered_;
+  };
+
   // function to query whether a `FusionExecutor` has a compiled kernel to
   // execute
-  bool isCompiled() const {
+  bool hasCompiledKernel() const {
```
jjsjann123, line 187: :+1:

csrc/executor.h
```diff
@@ -164,11 +170,26 @@ class FusionExecutor : public NonCopyable {
     post_lowering_hooks_.push_back(std::move(hook));
   }
 
+  // Function to query whether compilation was attempted for a `FusionExecutor`
+  bool isCompiled() const {
```
jjsjann123, line 174: looks like we changed the plan of renaming here.

csrc/executor.h
```diff
@@ -164,11 +170,26 @@ class FusionExecutor : public NonCopyable {
...
+    // Check atmost one of fusion_ and lowered_ is null.
```
jjsjann123, line 175:
```suggestion
    // Check at most one of fusion_ and lowered_ is null.
```

csrc/executor.h
```diff
@@ -164,11 +170,26 @@ class FusionExecutor : public NonCopyable {
...
+    NVF_ERROR(
+        fusion_ == nullptr || !lowered_,
+        "Expected GPU lowering to be skipped.");
+    NVF_ERROR(
+        lowered_ == nullptr || !fusion_,
+        "fusion_ should only be initializaed when using expression evaluator.");
```
jjsjann123, line 181: out of curiosity, `!x` is just to check `x != nullptr`, right?

I don't see how the above check is associated with the error message. We are only checking that there's one initialized at most, so that's a `!(fusion_ && lowered_)`

csrc/executor.h
jjsjann123: I'm not nitpicking on this BTW.

Priya2698: Yes, I decided to keep `isCompiled` the same at every place. At `FusionExecutorCache / FusionKernelRuntime` level, we should not need to distinguish between nvFuser vs ATen evaluation generally, so trying to limit that detail within `FusionExecutor`

csrc/executor.h
```diff
@@ -199,6 +220,11 @@ class FusionExecutor : public NonCopyable {
     return lowered_->kernel();
   }
 
+  Fusion* fusion() const {
+    NVF_ERROR((lowered_ && !fusion_) || (!lowered_ && fusion_));
```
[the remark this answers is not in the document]
Priya2698: Yes, but XOR does not work on these objects.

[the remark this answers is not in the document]
jjsjann123: TIL. `explicit operator bool()`

jjsjann123 (approved):
Thanks for addressing my concerns.

Priya2698:
!build

csrc/executor.h
```diff
@@ -199,6 +215,13 @@ class FusionExecutor : public NonCopyable {
     return lowered_->kernel();
   }
 
+  Fusion* fusion() const {
+    NVF_ERROR(
+        (lowered_ && !fusion_) || (!lowered_ && fusion_),
```
wujingyue, line 220: I think it's clearer to say 
```
(lowered_ == nullptr) != (fusion_ == nullptr)
```
which essentially use `!=` to do a logical xor.

csrc/executor.h
```diff
@@ -398,6 +421,14 @@ class FusionExecutor : public NonCopyable {
       int64_t runtime_id,
       int64_t group_id);
 
+  //! Check if compilation was skipped (fusion segment marked for EE).
+  bool isExprEval() const {
```
wujingyue, line 425: How about isExpressionEvaluated for less ambiguity?

tests/cpp/test_resize.cpp
```diff
@@ -1900,6 +1902,11 @@ TEST_F(ResizeTest, FusionSliceForNanoGPT2) {
 
 // C++ version of TestNvFuserFrontend.test_nanogpt_split_mha_linears
 TEST_F(ResizeTest, FusionSliceForNanoGPT3) {
+  // To verify input caching condition in this test, disable aliasing as that
+  // will skip compilation and no kernel will exist.
+  preseg_passes::OptimizationPassGuard<preseg_passes::MarkAliasesPreparePass>
+      optimization_guard(false);
+
   auto fusion_ptr = std::make_unique<Fusion>();
```
wujingyue, line 1910: I know you didn't add this, but since you are here, do you mind doing a minor, side refactor? 

1. Change this line to
   ```
   ```
2. Remove the next line says `auto& fusion=*fusion_ptr`. 
3. Change existing uses of `fusion.` to `fusion->`. 

The gist is that `unique_ptr` comes with syntax sugar to make it as if it's a raw pointer. It's not necessary to create another variable just to hold its reference.

PR #1931, opened 2024-03-13
Use negative axis for `scheduleTMASwizzle`

zasdfgbnm:
This change should be a noop. But using negative axis is required for supporting tiling.


zasdfgbnm:
!build

drzejan2 (approved):
Changes look good, thanks you!

PR #1935, opened 2024-03-14
Multidevice python fork

jjsjann123:
plumbing some multidevice API to python:
1. `nvfuser.ParallelType.mesh_x` enum parallel type;
2. `FusionDefinition.sched._create_device_mesh`;
3. `FusionDefinition.sched._set_device_mesh`;
4. `FusionDefinition.sched._parallelize`;

Nothing is supposed to be final on API yet. The runtime has also an ugly flag `multidevice_flag` to hijack the executor to `multi_device_executor`. This flag should go away after we have proper integration with FusionExecutorCache.


jjsjann123:
surprisingly the rebase just worked. Some minor clean up to accommodate the api update from multidevice.

@samnordmann @cowanmeg wondering if the python test looks like reasonable?

python_tests/mpi_test.py
```diff
@@ -0,0 +1,61 @@
...
+
+class MultiDeviceModel(FusionDefinition):
+
+    def definition(self):
+        # dynamic shape isn't working? at least I'm getting asserts
+        #self.t0 = self.from_pytorch(inputs[0])
+
+        self.t0 = self.define_tensor((2, 4), (False, False), dtype=DataType.Float)
+
+        # looks like I cannot have scalar inputs to any expression, hitting an assert
```
samnordmann, line 31: scalars input not supported for now. This is not a hard restriction, we just didnt take the time to add it yet, and this is probably not a priority for now

python_tests/mpi_test.py
```diff
@@ -0,0 +1,61 @@
...
+        #self.s0 = self.define_constant(2.0)
+        #self.t1 = self.ops.mul(self.t0, self.s0)
+
+        # relu seems to also be complaining, I assume sharding operation has to be a set
```
samnordmann, line 35: fixed with this one-line patch https://github.com/jjsjann123/Fuser/pull/2/commits/51e0bdc32bf729c5525e39a17f9046a5ab5f66e1

samnordmann (commented):
Thanks a lot for this Python API :) It's really great to see those tests working.
I opened a PR on your branch to propose fixes/discussions to the couple of bugs you identified. (I found opening another PR was nicer than just pushing to this branch, but let me know if you prefer to do otherwise)
https://github.com/jjsjann123/Fuser/pull/2

csrc/python_frontend/python_bindings.cpp
```diff
@@ -2730,6 +2743,47 @@ void initNvFuserPythonBindings(PyObject* module) {
...
+        FusionDefinition* fd = self.fusion_definition;
+	      fd->setMultiDevice();
+        auto tv =
+            fd->getFusionState(arg.index)->template as<TensorView>();
+        tv->setDeviceMesh(mesh);
+      },
+      py::arg("arg"),
+      py::arg("mesh"));
+  nvf_sched.def(
+      "parallelize",
```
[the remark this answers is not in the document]
cowanmeg: maybe then DIDx --> mesh_x since we are binding the axis to the x dimension of the device mesh?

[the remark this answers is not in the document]
jjsjann123: Sounds like a good idea. Thought this is a larger topic, but I'm very willing to pick it up after this initial PR.

jjsjann123:
!build

jjsjann123:
!build

jjsjann123:
Looks like build is failing on CI. Let me try to patch that.

samnordmann (approved):
LGTM! thanks

csrc/python_frontend/fusion_definition.h
```diff
@@ -240,6 +250,17 @@ class NVF_API FusionDefinition : public FusionState {
 
  private:
   mutable std::optional<std::string> debug_output_ = std::nullopt;
+
+#ifdef NVFUSER_DISTRIBUTED
+  //! DO NOT COMMIT TO THESE CHANGES!
```
wujingyue, line 255: Sounds like a scary comment. Did you intend to fix it?

csrc/python_frontend/fusion_definition.cpp
```diff
@@ -170,6 +175,26 @@ std::vector<at::Tensor> FusionDefinition::execute(
 
   auto scheds = fusionCache()->queryFusionSchedules(id().value());
 
+#ifdef NVFUSER_DISTRIBUTED
+  static std::unique_ptr<Communicator> comm = nullptr;
```
wujingyue, line 179:
```suggestion
  static Communicator* comm = new Communicator();
```

unique_ptr here doesn't provide benefits -- it'll be deleted when the program exits anyway. However, a static, non-trivially destructible object can potentially create order problems. 

https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables

csrc/python_frontend/fusion_definition.cpp
```diff
@@ -170,6 +175,26 @@ std::vector<at::Tensor> FusionDefinition::execute(
 
   auto scheds = fusionCache()->queryFusionSchedules(id().value());
 
+#ifdef NVFUSER_DISTRIBUTED
```
wujingyue, line 178: Doesn't seem to be necessary? In general, I'd like to reduce the use of the macro, which makes single- and multi-device build diverge.

csrc/python_frontend/fusion_definition.cpp
```diff
@@ -187,8 +212,10 @@ std::vector<at::Tensor> FusionDefinition::execute(
     }
   }
 
-  outputs = scheds->auto_gen_schedules->runFusionWithInputs(
-      inputs, std::nullopt, selected_device);
+  if (outputs.empty()) {
```
wujingyue, line 215: Why is this check needed?

csrc/python_frontend/fusion_definition.cpp
jjsjann123: We are using `multi_device_executor` which is `std::unique_ptr<MultiDeviceExecutor>`. `MultiDeviceExecutor` is hidden behind the same macro `NVFUSER_DISTRIBUTED`.

single/multi-device build has to diverge, since our mult-device build relies on PyTorch package.
PyTorch could be shipped without multi-device support and we still want nvfuser to be able to built against that.

csrc/python_frontend/fusion_definition.cpp
jjsjann123: touche.

Thanks for catching this one!

csrc/python_frontend/fusion_definition.cpp
jjsjann123: I think I fixed the code. sorry for the ugly patch. I should have done a better job.

If you look at line 200-213. When `!override_user_schedule` is true, we shouldn't run automatic scheduled fusion. I'll clean it up. :bow:

csrc/python_frontend/fusion_definition.cpp
wujingyue: > MultiDeviceExecutor is hidden behind the same macro NVFUSER_DISTRIBUTED.

You are right -- I missed that! Aha, I can see we'll soon check NVFUSER_DISTRIBUTED "everywhere" :) It's so easy to have to propagate it. 

I'll look more into this. It may be possible to confine the use in just the communicator -- the communicator may be able to return the same tensor when the world size is one instead of calling c10d (which is guarded behind NVFUSER_DISTRIBUTED AFAIK). 

> single/multi-device build has to diverge

Yeah, my point was to reduce the build divergence because more build divergence means more likely to see build-succeeds-on-my-machine-but-fails-on-CI.

samnordmann: > You are right -- I missed that! Aha, I can see we'll soon check NVFUSER_DISTRIBUTED "everywhere" :) It's so easy to have to propagate it.
> 
> I'll look more into this. It may be possible to confine the use in just the communicator

You are right! It's been on our TODO list with @cowanmeg for quite some time now :) Some non-trivial refactor and interface will be needed though in communicator.cpp/.h and communication.cpp/.h
If it becomes higher priority let me know

csrc/python_frontend/fusion_definition.h
jjsjann123: If you carry on with your multi device executor integration, it would just naturally go away after that :stuck_out_tongue_winking_eye:

jjsjann123:
!build

jjsjann123:
@wujingyue wondering if you have any more concern on this PR?
Not rush you to review, but just want to make sure I get your stamp before merging. :bow:

csrc/multidevice/utils.cpp
```diff
@@ -132,6 +132,7 @@ void shardAllLike(TensorView* ref, std::vector<TensorView*> tvs) {
 } // namespace
 
 void insertReshardings(Fusion* fusion) {
+  FusionGuard fg(fusion);
```
wujingyue, line 135:
```suggestion
  // Remove this after we refactor this as a pre-segmenter pass. 
  FusionGuard fg(fusion);
```

csrc/python_frontend/fusion_definition.cpp
```diff
@@ -133,11 +134,15 @@ void FusionDefinition::finalizeSchedule(
   FusionGuard::setCurFusion(prev_fusion_);
   prev_fusion_ = nullptr;
 
-  user_sched_->executor->compileFusion(
-      user_sched_->schedule.get(),
-      inputs,
-      user_sched_->fusion_id_,
-      user_sched_->device_id_);
+#ifdef NVFUSER_DISTRIBUTED
```
wujingyue, line 137: Is this line needed? Are anything in the block only declared under NVFUSER_DISTRIBUTED?

csrc/python_frontend/fusion_definition.h
wujingyue: What does "DO NOT COMMIT TO THESE CHANGES!" mean? Are we not supposed to commit the following lines in this PR?

csrc/python_frontend/fusion_definition.h
```diff
@@ -240,6 +250,17 @@ class NVF_API FusionDefinition : public FusionState {
...
+  //! The reason we have these is due to the lack of cache for multidevice
+  //! executor
+  bool multidevice_flag = false;
```
wujingyue, line 258:
```suggestion
  bool multidevice_flag_ = false;
```

csrc/python_frontend/fusion_definition.h
```diff
@@ -240,6 +250,17 @@ class NVF_API FusionDefinition : public FusionState {
...
+  //! DO NOT COMMIT TO THESE CHANGES!
+  //! The reason we have these is due to the lack of cache for multidevice
+  //! executor
+  mutable std::unique_ptr<MultiDeviceExecutor> multi_device_executor = nullptr;
```
wujingyue, line 262:
```suggestion
  mutable std::unique_ptr<MultiDeviceExecutor> multi_device_executor_ = nullptr;
```

csrc/python_frontend/fusion_definition.h
wujingyue, line 258: I smelled redundancy between multidevice_flag_ and multi_device_executor_. I understand they are set at different times. However, before any fusion is executed, `multi_device_executor_` is not null iff `multidevice_flag` is true. Correct? 

If so, we may be able to set multi_device_executor directly in setMultiDevice instead of using multidevice_flag_ as a proxy.

wujingyue: There's also `multidevice_*` vs `multi_device_*`. I slightly prefer the former because multi-device (despite of the hypen) is an English word. But we should at least make the names in this file consistent.

csrc/python_frontend/fusion_definition.h
jjsjann123: I use too harsh a comment there. I'll rename it to `//! following code is to be refactored when multidevice executor is properly integrated`

csrc/python_frontend/fusion_definition.h
jjsjann123: That's a good call! Let me refactor it.

csrc/python_frontend/fusion_definition.cpp
jjsjann123: `multidevice_flag` and `multi_device_executor` is hidden behind the flag.

If i'm to refactor `multi_device_executor` to replace `multidevice_flag`, I'll need this macro I think.

csrc/python_frontend/fusion_definition.h
jjsjann123: Need to have this one first. #2051

samnordmann: For now, `MultiDeviceExecutor` constructor takes a `Fusion` as an argument (similarily to `FusionExecutorCache` and `FusionExecutor`) so it is not possible to instantiate it before the fusion is defined

tests/python/mpi_test.py
```diff
@@ -0,0 +1,44 @@
...
+# scalar inputs isn't supported;
+class MultiDeviceModel(FusionDefinition):
+    def definition(self):
+        self.t0 = self.define_tensor((2, 4), (False, False), dtype=DataType.Float)
+        self.t1 = self.ops.relu(self.t0)
+        self.t2 = self.ops.add(self.t1, self.t1)
+        self.add_output(self.t2)
+
+    def schedule(self):
+        self.mesh = self.sched._create_device_mesh((0, 1))
```
wujingyue, line 30: Why is `mesh` a field instead of a local variable?

csrc/python_frontend/python_bindings.cpp
```diff
@@ -2726,6 +2735,48 @@ void initNvFuserPythonBindings(PyObject* module) {
...
+      [](FusionDefinition::SchedOperators& self, std::vector<int64_t> devices) {
+        FUSER_PERF_SCOPE("SchedOperators._create_device_mesh");
+        return DeviceMesh(devices);
+      },
+      py::arg("devices"),
+      py::return_value_policy::reference);
+  //! experimental API for multidevice support
+  nvf_sched.def(
+      "_set_device_mesh",
+      [](FusionDefinition::SchedOperators& self, Tensor arg, DeviceMesh mesh) {
```
wujingyue, line 2750: Does the name `arg` mean anything special? For example, does it have to be a fusion input? If not, we could call it `tensor` or `t` for conciseness and less confusion.

wujingyue, line 2750: I saw many APIs in this file take values instead of references. Is that intentional? IOW, can we do `const DeviceMesh& mesh` here?

csrc/python_frontend/python_bindings.cpp
```diff
@@ -2726,6 +2735,48 @@ void initNvFuserPythonBindings(PyObject* module) {
   py::class_<FusionDefinition::SchedOperators> nvf_sched(
       fusion_def, "SchedOperators");
   nvf_sched.def(py::init<FusionDefinition*>());
+  //! experimental API for multidevice support
+  nvf_sched.def(
+      "_create_device_mesh",
+      [](FusionDefinition::SchedOperators& self, std::vector<int64_t> devices) {
+        FUSER_PERF_SCOPE("SchedOperators._create_device_mesh");
```
wujingyue, line 2742: Why FUSER_PERF_SCOPE? Is the constructor of DeviceMesh expensive enough to justify the overhead of scoping?

tests/python/mpi_test.py
```diff
@@ -0,0 +1,44 @@
...
+# dynamic shape isn't supported;
+# scalar inputs isn't supported;
+class MultiDeviceModel(FusionDefinition):
+    def definition(self):
+        self.t0 = self.define_tensor((2, 4), (False, False), dtype=DataType.Float)
+        self.t1 = self.ops.relu(self.t0)
+        self.t2 = self.ops.add(self.t1, self.t1)
+        self.add_output(self.t2)
+
+    def schedule(self):
```
wujingyue, line 29: Can you remind me where `schedule` gets used? I'm trying to understand why TensorView isn't available during `definition` but is during `schedule`.

csrc/python_frontend/fusion_definition.h
```diff
@@ -182,6 +183,15 @@ class NVF_API FusionDefinition : public FusionState {
   //! Gets a Record State object
   NVF_API State recordingState(size_t index) const;
 
+  //! Experimental API we'll redo this later!
+  void setMultiDevice() {
+#ifdef NVFUSER_DISTRIBUTED
```
cowanmeg, line 188: We can use the `distributedEnabled()` helper function to clean things up a bit.
https://github.com/NVIDIA/Fuser/blob/main/csrc/multidevice/utils.h#L18

tests/python/mpi_test.py
wujingyue: https://github.com/NVIDIA/Fuser/blob/9eba1ad122550bf598f4a8d453d003cea989e4fe/nvfuser/__init__.py#L141

Thanks for the pointer, @jjsjann123!

csrc/python_frontend/fusion_definition.h
jjsjann123: Awesome. Thanks for that.

tests/python/mpi_test.py
jjsjann123: good call. This one doesn't matter for the example here.

csrc/python_frontend/python_bindings.cpp
jjsjann123: That's just a boilerplate variable name for pybind functions, since all pybind args is declared with py::arg.

I actually don't know if there's a convention here. But I agree that tensor is probably more specific and a better choice. I'll update it for added functions here.

csrc/python_frontend/fusion_definition.cpp
```diff
@@ -130,14 +131,38 @@ void FusionDefinition::setupSchedule(const at::ArrayRef<c10::IValue>& inputs) {
 void FusionDefinition::finalizeSchedule(
     const at::ArrayRef<c10::IValue>& inputs) {
   FUSER_PERF_SCOPE("FusionDefinition::finalizeSchedule");
+#ifdef NVFUSER_DISTRIBUTED
```
jjsjann123, line 134: note: realized that it's easier to read having it separate out.

jjsjann123:
!build --dist

csrc/python_frontend/fusion_definition.cpp
```diff
@@ -130,14 +131,38 @@ void FusionDefinition::setupSchedule(const at::ArrayRef<c10::IValue>& inputs) {
...
+  // TODO: remove when multidevice executor integration is done natively
+  Fusion* fusion = user_sched_->schedule.get();
+  std::vector<Val*> fusion_inputs = InputsOf::outputs(fusion->outputs());
+  std::vector<Val*> vals = DependencyCheck::getAllValsBetween(
+      {fusion_inputs.begin(), fusion_inputs.end()}, fusion->outputs());
+  static Communicator* comm = new Communicator();
+  if (std::any_of(vals.begin(), vals.end(), [](Val* v) {
+        return v->isA<TensorView>() && v->as<TensorView>()->hasDeviceMesh();
+      })) {
+    multidevice_executor_ = std::make_unique<MultiDeviceExecutor>(
```
jjsjann123, line 144: FYI: my assumption here is that MultiDeviceExecutor wouldn't care about input "device". So there's no need to create one instances per device.

I'm actually quite curious how we'll handle this in integrated. Currently executor is fixed to one device. And support across different devices are handled by kernel_runtimes.
I imagine this would look very differently on multidevice support?

cc'ing @wujingyue @cowanmeg @samnordmann

jjsjann123:
!build --dist

csrc/python_frontend/fusion_definition.cpp
```diff
@@ -130,14 +131,38 @@ void FusionDefinition::setupSchedule(const at::ArrayRef<c10::IValue>& inputs) {
 void FusionDefinition::finalizeSchedule(
     const at::ArrayRef<c10::IValue>& inputs) {
   FUSER_PERF_SCOPE("FusionDefinition::finalizeSchedule");
+#ifdef NVFUSER_DISTRIBUTED
+  // TODO: remove when multidevice executor integration is done natively
+  Fusion* fusion = user_sched_->schedule.get();
+  std::vector<Val*> fusion_inputs = InputsOf::outputs(fusion->outputs());
+  std::vector<Val*> vals = DependencyCheck::getAllValsBetween(
```
wujingyue, line 138: Would this work? https://github.com/NVIDIA/Fuser/blob/618c4b79a053692bf83ebb32aa19dcc907ebb534/csrc/ir/utils.cpp#L406

csrc/python_frontend/fusion_definition.cpp
```diff
@@ -130,14 +131,38 @@ void FusionDefinition::setupSchedule(const at::ArrayRef<c10::IValue>& inputs) {
...
+  if (std::any_of(vals.begin(), vals.end(), [](Val* v) {
+        return v->isA<TensorView>() && v->as<TensorView>()->hasDeviceMesh();
+      })) {
+    multidevice_executor_ = std::make_unique<MultiDeviceExecutor>(
+        std::make_unique<Fusion>(*fusion), *comm);
+  }
+
+  FusionGuard::setCurFusion(prev_fusion_);
+  prev_fusion_ = nullptr;
+  if (!multidevice_executor_) {
```
wujingyue, line 150:
```suggestion
  if (multidevice_executor_ == nullptr) {
```

wujingyue: I guess you don't like the duplication on Line 158 either :) I think, if we declare (but not define) MultiDeviceExecutor when `!NVFUSER_DISTRIBUTED`, we should be able to use std::unique_ptr<MultiDeviceExecutor> as long as we don't construct the executor. 

That being said, once we hopefully localize NVFUSER_DISTRIBUTED to Communicator, we'll stop worry about all these.

wujingyue (approved):

csrc/python_frontend/fusion_definition.cpp
jjsjann123: Good call. Let me try that

csrc/python_frontend/fusion_definition.cpp
cowanmeg: We are using one process per device, so yes the executor is essentially fixed to one device.
> And support across different devices are handled by kernel_runtimes.

What do you mean by this? 

> I imagine this would look very differently on multidevice support?

Like one process per node and one thread per device?

wujingyue: > We are using one process per device

Wow, can you show me how this is done in code? For example, do we launch multiple bin/nvfuser_tests processes to run a multidevice test when there are multiple GPUs?

jjsjann123: > We are using one process per device, so yes the executor is essentially fixed to one device.

When I say `executor is fixed to one device`, I was referring to `FusionExecutor` and I was hoping that `MultiDeviceExecutor` would be agnostic for that.
Sounds like I'm totally wrong then. So I need to create a vector of `MultiDeviceExecutor` here to handle input on different devices as well...

Also, how would a broadcast look like?

i.e. input tensor is in cuda:0 and later sharded to both `cuda:0` and `cuda:1`? Do we need to feed input to both program?

cowanmeg: Yeah, we run all the multidevice tests with `mpirun -np <num_devices>` so nothing special  is done in the executor.

jjsjann123: FYI, I realize this is a bigger question and somewhat orthogonal to this PR. (given that device support on `schedule` is somewhat broken in tree).
I'm going to merge this PR without properly patching this topic.

cowanmeg: > So I need to create a vector of MultiDeviceExecutor here to handle input on different devices as well...

Oops I think my comment is causing confusion. No changes needed in your code since mpirun is creating the processes! 

>  input tensor is in cuda:0 and later sharded to both cuda:0 and cuda:1? Do we need to feed input to both program?

We probably need to think about how to handle sharded inputs better... Right now in our unit tests are pretty naive, we allocate the whole tensor, shard it, then pass it to the executor.

cowanmeg: > FYI, I realize this is a bigger question and somewhat orthogonal to this PR. (given that device support on `schedule` is somewhat broken in tree). I'm going to merge this PR without properly patching this topic.

Yes sorry, for derailing the PR. All my comments are side conversation :D

jjsjann123: > Oops I think my comment is causing confusion. No changes needed in your code since mpirun is creating the processes!

no I need that changes for a different case, where input changes on device across iterations. Actually I'm not totally sure if MultiDeviceExecutor supports that yet. We'll find out.

> Yes sorry, for derailing the PR.

Don't be! your comments are super helpful and I was the one asking for comments in the first place. I was just trying to communicate and clarify the scope of this PR is not going to address issues we discussed in this thread.
:bow:

csrc/python_frontend/fusion_definition.cpp
jjsjann123: Capture off line conversation.
An incomplete type `MultiDeviceExecutor` won't work here with `std::unique_ptr<MultiDeviceExecutor>`. `~unique_ptr` requires calls `sizeof` on the instance, and we do not have a definition of it at all :cry: 

We'll live with the duplicated code.

jjsjann123:
!build --dist

PR #1938, opened 2024-03-14
Calculate the buffer size for a persistent tensor in a separate function

liqiangxl:
(1) Moved the code used to calculate the buffer size for a persistent tensor into a separate function.
(2) Changed the logic to detect inputs from index based to more general `std::find` so it can be used in other places (will be used in a following PR revising inner outer combined scheduler).


jjsjann123 (approved):
stamping mechanical changes.

out of curiosity, is this code going to be re-used later? wondering the motivation behind the refactor.

jjsjann123:
!build

liqiangxl:
> stamping mechanical changes.
> 
> out of curiosity, is this code going to be re-used later? wondering the motivation behind the refactor.

Yes. I extracted this change from the PR for layer norm backward. Will be used there.

PR #1939, opened 2024-03-14
Shard non-outermost axis

cowanmeg:
Currently, we restrict sharding to the outermost axis. This PR enables non-outermost axis to be sharded:

```
TensorView* a = makeContigTensor(2);
a->axis(1)->parallelize(ParallelType::DIDx);
```

When lowering to collectives where input/output tensors are different shape (i.e. scatter/gather comms) the sharded axis cannot be read/written contiguously which some ProcessGroup collectives do not support. 
We handle this by copying the tensor into a layout where the sharded axis is the outermost axis before the communication op, then copy back into the original layout.

For gather operations:
```
TensorView* a = makeContigTensor(2);  // [i0, DIDx(i1)]
a->axis(1)->parallelize(ParallelType::DIDx);
TensorView* b = set(a); // [i0, i1]
```
we insert permutations that push the sharded axis forward before the set (gather) then permute the axis back to the proper index.
```
TensorView* a = makeContigTensor(2); // [i0, DIDx(i1)]
a->axis(1)->parallelize(ParallelType::DIDx);
TensorView* a_ = permute(a, {1, 0}) // [DIDx(i1), i0]
a_->axis(0)->parallelize(ParallelType::DIDx);
TensorView* b_ = set(a); // [i1, i0]
TensorView* b = permute(b_, {0, 1}); // [i0, i1]
```

For scatter operations:
```
TensorView* a = makeContigTensor(2);  // [i0, i1]
TensorView* b = set(a); [i0, DIDx(i1)]
b->axis(1)->parallelize(ParallelType::DIDx);
```
we insert permutations that push the sharded axis forward before the set (scatter) then permute the axis back to the proper index.
```
TensorView* a = makeContigTensor(2); // [i0, i1]
TensorView* a_ = permute(a, {1, 0}) // [i1, i0]
TensorView* b_ = set(a); // [DIDx(i1), i0]
b->axis(0)->parallelize(ParallelType::DIDx);
TensorView* b = permute(b_, {0, 1}); // [i0, DIDx(i1)]
b->axis(1)->parallelize(ParallelType::DIDx);
```

This is implemented as a pass that runs `insertReshardings` so that all resharding expressions are known to be `set` or `reduce` operations.


wujingyue:
Nice! I understand this is a strict improvement because something used to fail now can run. To help me understand the problem better, is it an nccl limitation that can only scatter/gather the outermost dimension so nvFuser has to work around by relayout (which is not free)?

csrc/multidevice/lower_communication.cpp
```diff
@@ -170,7 +170,7 @@ void lowerToGather(
     at::Tensor output_tensor,
     std::vector<std::shared_ptr<Communication>>& comms) {
   // we create as many 'Gathers' as there are devices in the receiver mesh
-  auto sender_mesh = input_tv->getDeviceMesh();
+  const auto& sender_mesh = input_tv->getDeviceMesh();
```
wujingyue, line 173:
```suggestion
  const DeviceMesh& sender_mesh = input_tv->getDeviceMesh();
```

Style nit: I tend to not use auto unless the type is clear from the lines around it. That leads to more typing but is general worthwhile for future readers (especially newcomers) to easily understand the code. https://google.github.io/styleguide/cppguide.html#Type_deduction

csrc/multidevice/lower_communication.cpp
```diff
@@ -397,10 +391,19 @@ void lowerToReduceScatter(
...
+    shard_axis++;
+  }
   for (auto i : c10::irange(mesh.vector().size())) {
-    auto sliced_buf = input_tensor.index(
-        {at::indexing::Slice(0, 1), static_cast<int>(i), "..."});
-    params.src_bufs.push_back(sliced_buf);
+    std::vector<at::indexing::TensorIndex> indices(
+        input_tensor.dim(), at::indexing::Slice());
+    indices[shard_axis] = at::indexing::Slice(i, i + 1);
+    auto slice = input_tensor.index(indices).squeeze(red_axis);
```
wujingyue, line 405: I think it can be simplified using the same trick as https://github.com/NVIDIA/Fuser/pull/1946#discussion_r1532569253

csrc/multidevice/lower_communication.cpp
```diff
@@ -397,10 +391,19 @@ void lowerToReduceScatter(
   params.redOp = getC10dReduceOpType(op_type);
   params.team = mesh.vector();
   params.dst_bufs = {output_tensor};
+  // TODO: Clean up once allocation domain for sharded tensors is
+  // better supported.
+  auto red_axis = output_tv->getReductionAxis().value();
```
wujingyue, line 396:
```suggestion
  auto reduction_axis = output_tv->getReductionAxis().value();
```

The typing saved isn't worth the confusion.

csrc/multidevice/lower_communication.cpp
```diff
@@ -397,10 +391,19 @@ void lowerToReduceScatter(
...
+  auto shard_axis = getShardedAxis(output_tv);
```
wujingyue, line 397:
```suggestion
  auto sharded_axis = getShardedAxis(output_tv);
```

csrc/multidevice/lower_communication.cpp
```diff
@@ -397,10 +391,19 @@ void lowerToReduceScatter(
...
+  if (red_axis <= shard_axis) {
+    shard_axis++;
+  }
```
wujingyue, line 400: I'm sure this and code around it are doing something cool, but I cannot immediately understand them! Can you add some code comment?

csrc/multidevice/utils.cpp
```diff
@@ -13,22 +13,93 @@
...
 #include <scheduler/utils.h>
 
 #include <c10/util/irange.h>
 
 namespace nvfuser {
 namespace {
 
-std::vector<IterDomain*> getShardedIterDomains(TensorView* tv) {
-  std::vector<IterDomain*> sharded_ids;
+const std::unordered_set<IterDomain*> getShardedIterDomains(TensorView* tv) {
```
wujingyue, line 24:
```suggestion
std::unordered_set<IterDomain*> getShardedIterDomains(TensorView* tv) {
```

I'll try to find some reference, but IIRC marking the return object (not ref) as const doesn't provide practical benefits.

wujingyue: https://stackoverflow.com/questions/8716330/purpose-of-returning-by-const-value

csrc/multidevice/utils.cpp
```diff
@@ -13,22 +13,93 @@
...
+// contiguously. when a IterDomain's sharding is flipped. This occurs when the
+// IterDomain being unsharded/sharded is the outermost in the allocation domain.
+// e.g. [DIDx(i0), r1, i2, i3]
+// i0 => true, outermost axis
+// i1 => true, reduction axis are not allocated
+// i2 => true, reduction and device axis are not allocated
+// i3 => false, i2 is allocated.
+bool isContiguousShard(TensorView* tv, IterDomain* changed_id) {
+  bool not_allocated = true;
+  for (auto id : tv->getLeafDomain()) {
```
wujingyue, line 44: I assume this is another place that we'll have to fix later when allocation domain supports multidevice better? I'm curious what supports are missing. Do you have a bug about it? I'm sure @jjsjann123 is all ears for this.

csrc/multidevice/utils.cpp
```diff
@@ -13,22 +13,93 @@
...
+    if (id == changed_id) {
+      return not_allocated;
+    }
+    not_allocated = not_allocated && (id->isDeviceDim() || id->isReduction());
+  }
```
wujingyue, line 49:
```suggestion
    if (id == changed_id) {
      return true;
    }
    if (!id->isDeviceDim() && !id->isReduction()) {
      return false;
    }
  }
```

csrc/multidevice/utils.cpp
```diff
@@ -13,22 +13,93 @@
...
+  NVF_ERROR(
+      false,
+      "Id",
+      changed_id->toString(),
+      " is not in TensorView ",
+      tv->toString());
+  return false;
+}
+
+// For a sharding expression, either a set or reduce, returns root IDs
```
wujingyue, line 59: Question about naming: I saw sometimes we say resharding expressions (or ops) vs sharding expressions. Do they have different meanings or are used interchangeably?

csrc/multidevice/utils.cpp
```diff
@@ -13,22 +13,93 @@
...
+  NVF_ERROR(
+      expr->outputs().size() == 1,
+      "Resharding operations can only have one output");
+  NVF_ERROR(
+      expr->inputs().size() == 1,
+      "Resharding operations can have only one input");
+  auto output = expr->outputs().at(0)->as<TensorView>();
+  auto input = expr->inputs().at(0)->as<TensorView>();
+
+  auto sharded_ids_input = getShardedIterDomains(input);
```
wujingyue, line 78: I couldn't connect the implementation with the function comment. Why do we look at `input` to decide whether a root ID of `output` is sharded? Maybe a different way of asking this is what IterDomain is considered to be sharded? (I thought we just need to check its parallel type)

csrc/multidevice/utils.cpp
cowanmeg: Yes, this helper function will need to be updated. imo the allocation domain needs to support multidevice better when we allow sharding on merged/split axis. 

Some background: Currently, sharding is only the RFactor domain. We have the constraint that `extent of a sharded axis == number of devices`, which guarantees the `sharded size of the axis = 1` or the actual allocation size of that axis = 1.

Eventually, we want to shard a split axis:
```
D = # of devices contained in DIDx
tv = makeContigTensor(2); [i0, i1]
tv->split(0, D, outer=true); //[D, i0//D, i1]
tv->axis(0)->parallelize(DIDx);
```
The allocation domain should be [i0//D, i1]
I might be wrong, but since we don't explicitly set the allocation domain for a tv with device parallelization (maybe we should?), the allocation domain is inferred from the rfactor domain?

As an aside sharding a split axis makes a lot of sense to me...but I don't know whether we should allow sharding a merged axis...

csrc/multidevice/utils.cpp
cowanmeg: I used them interchangeably. I can make this more consistent if there is a preferred naming for nvfuser?

csrc/multidevice/utils.cpp
cowanmeg: This function is identifying which axes _change_ sharding. So both input and output need to be checked. 
[DIDx(i0) DIDy(i1)] -> [i0 DIDy(i1)] deletions: i0 
[i0 DIDx(i1)] -> [DIDx(i0) DIDy(i1)] additions: i0

I will fix the comment to be more clear!

csrc/multidevice/utils.cpp
naoyam: As part of indexing refactoring, I'll probably explicitly set the allocation domain for all tensors. 

> As an aside sharding a split axis makes a lot of sense to me...but I don't know whether we should allow sharding a merged axis...

Maybe not common, but we may do that with `BIDx` as long as the extent is smaller than the maximum `gridDim.x`. Probably unlikely for `DIDx`.

csrc/multidevice/utils.cpp
```diff
@@ -151,6 +222,108 @@ void insertReshardings(Fusion* fusion) {
...
+      TensorView* output_permute = set(input_permute);
+      TensorView* new_output = permute(output_permute, {{0, idx}});
+      ir_utils::replaceValInAllExprInputsAndFusionOutputs(output, new_output);
+      for (int i = 0; i < static_cast<int>(output->nDims()); i++) {
+        new_output->axis(i)->parallelize(output->axis(i)->getParallelType());
+      }
+      new_output->setDeviceMesh(output->getDeviceMesh());
+      shardAllLike(input, {input_permute});
+      shardAllLike(new_output, {output_permute});
+      fusion->removeExpr(expr);
```
naoyam, line 275: In the Fusion IR, it is not necessary to manage live vals and exprs manually. Unused exprs should be automatically removed or ignored. In this case, at this point, there should be no remaining expr that uses `output`, so while `expr` still exists, it should not show up in, e.g., `fusion->exprs()` anymore since it has no dependency with the fusion outputs.

csrc/multidevice/utils.cpp
```diff
@@ -151,6 +222,108 @@ void insertReshardings(Fusion* fusion) {
...
+    }
+    // For scatter operations i.e. ID goes from unsharded to sharded
+    // Update input to push the scattered axis to the front -> collective ->
+    // permute the sharded axis to the proper location.
+    // Example: [i0 i1 DIDx(i2)] -> [i0 DIDx(i1) r(DIDx(i2))]
+    // Rewritten to: [i0 i1 DIDx(i2)] -> [i1 i0 DIDx(i2)] ->
+    //                    [DIDx(i1) i0 r(DIDx(i2))] -> [i0 DIDx(i1)]
```
naoyam, line 282: This is a bit confusing as we do not use the same parallel type for multiple domains of the same tensor. 

Is this a reduce-scatter pattern?

csrc/multidevice/utils.cpp
```diff
@@ -13,22 +13,93 @@
...
+
+// Returns whether a IterDomain in a TensorView can be read/written
+// contiguously. when a IterDomain's sharding is flipped. This occurs when the
+// IterDomain being unsharded/sharded is the outermost in the allocation domain.
+// e.g. [DIDx(i0), r1, i2, i3]
+// i0 => true, outermost axis
+// i1 => true, reduction axis are not allocated
+// i2 => true, reduction and device axis are not allocated
+// i3 => false, i2 is allocated.
+bool isContiguousShard(TensorView* tv, IterDomain* changed_id) {
```
naoyam, line 42: This seems a little confusing to me. IIUC, looks like what you're answering here is whether each per-device tensor corresponds to a logically contiguous part of the overall tensor. If that's the case, no packing/permutation is necessary, so you can skip the insertion of the permutation below.

Does my understanding make sense? If so, this is a property of a tensor, not an IterDomain of the tensor, so having a parameter like `changed_id` doesn't seem necessary.

csrc/multidevice/utils.cpp
```diff
@@ -13,22 +13,93 @@
...
+
+  auto sharded_ids_input = getShardedIterDomains(input);
+  auto sharded_ids_output = getShardedIterDomains(output);
+  std::vector<IterDomain*> shard_additions;
+  std::vector<IterDomain*> shard_deletions;
+  auto rootmap = PairwiseRootDomainMap(input, output);
+
+  const auto c2p_map = rootmap.mapConsumerToProducer(&sharded_ids_output);
+  for (auto [id1, id2] : c2p_map) {
+    if (!id1->sameAs(id2)) {
```
naoyam, line 86: Since we are using the root map, these two domains are guaranteed to have almost the same properties except for parallel types. So, it should be just fine to check their parallel types.

Also, by default, the root map maps broadcast and corresponding non-broadcast domains. IIRC, there should be no broadcast change in sharding expressions, so this shouldn't matter, though. Maybe to be more explicit, do `PairwiseRootDomainMap(input, output).mapBroadcast(false)`.

csrc/multidevice/utils.cpp
```diff
@@ -13,22 +13,93 @@
...
+      shard_additions.push_back(id1);
+    }
+  }
+
+  const auto p2c_map = rootmap.mapProducerToConsumer(&sharded_ids_input);
+  for (auto [id1, id2] : p2c_map) {
+    // Ignore sharded reductions i.e.
+    // DIDx(i0) -> r(i0) or DIDx(i0)->r(DIDx(i0))
+    // since they don't affect allocation.
+    if (!id1->sameAs(id2) && !id2->isReduction()) {
```
naoyam, line 96: We only have `DIDx` for now, but what if `id1`'s parallel type is `DIDx` and the reduction `id2` is `DIDy`?

csrc/multidevice/utils.cpp
naoyam: > if sharded extent of i0 == 1.

Isn't it always the case?

cowanmeg: Sorry, my example wasn't very clear. I was thinking when we support sharding on a split axis we can have a scenario like this:

For a 2D tensor sharded on DeviceMesh [Dx, Dy] where axis0 is ahrded on Dx, axis1 sharded on Dy.

root [i0, i1] 
leaf [DIDx(Dx), i0/Dx, DIDy(Dy), i1/Dy] 
allocation [i0/Dx, i1/Dy]

if axis 1 is unsharded:
root [i0, i1] 
leaf [DIDx(Dx), i0/Dx, Dy, i1/Dy]
allocation [i0/Dx, i1]

To check if we need a permutation for this pattern, we could either check leaf domain index 3 is the outermost non-device/non-reduce axis (false). 
Or check from that index 1 on the allocation domain is the outermost (false). This check is a bit cleaner.

naoyam: Thanks. That makes sense.

naoyam: Is `changed_id` always parallelized with DID?

csrc/multidevice/utils.cpp
naoyam: Can you update the example to follow the nvFuser convention, i.e., don't use the same parallel type for multiple domains in the same tensor.

csrc/multidevice/utils.cpp
cowanmeg: No, I assume it is either serial or DID.

csrc/multidevice/utils.cpp
cowanmeg: Yes, I can do that. In fact the code assumes that the reduced ID is not device parallel. So lowering to reduce scatter is looking for patterns of this form:
[I0 DIDx(i1) i2] -> [DIDx(i0) r(i1) i2]

We had some earlier discussions about how to represent reduction of a Device parallel iter domain and whether it should be `DIDx(i1) -> r(i1)` vs. `DIDx(i1) -> r(DIDx(i1))`. 
There were some cases where the second one was preferred. Since we pattern match onto a collective, we don't control how that reduction is implemented so these two examples are lowered onto the same collective. 
It sounds like we should use the first option to avoid two axes with DIDx in reduce scatter scenarios?

csrc/multidevice/utils.cpp
cowanmeg: To clarify why the parallel type of `changed_id` isn't important, we just want to know if `changed_id` is the outermost allocated axis. 
Sorry, the comment and function name are very bad. I should call this function `isOutermostAllocatedId(TensorView* tv, IterDomain* id)`

csrc/multidevice/utils.cpp
naoyam: The latter is the correct parallel reduction pattern in nvFuser. I don't think we can express the reduce-scatter pattern as a single nvFuser op. What we could do instead is to have a parallel reduction op followed by another parallel scatter op. Then when they were lowered, we could lower them to a single reduce-scatter op.

Generally speaking, we prefer to have only primitive ops in the Fusion IR, so for example, we don't have a "all-reduce" op, but it's expressed as a reduction followed by a broadcast. This pattern can be lowered to an all-reduce Kernel IR op when it's possible. See for example, https://github.com/NVIDIA/Fuser/blob/main/csrc/kernel_ir.h#L1004.

csrc/multidevice/utils.cpp
wujingyue: I don't have a strong preference. Resharding seems to be more precise -- it says the fact that the sharding is changed after this op. Anyhow, I worried about consistency more than which name.

csrc/multidevice/utils.cpp
cowanmeg: Makes sense, thanks for the explanation. 
I agree as primitives, reduce scatter would be expressed as parallel reduction op + parallel scatter op. 

For now, I'll assume it's ok that we have `set` with one to one lowering to "fused" MPI collectives. Later we will probably update as we improve the communication lowering analysis.

csrc/multidevice/utils.cpp
cowanmeg: Yes, I updated sharding to resharding and went through the PR to make certain the naming was consistent.

csrc/multidevice/utils.cpp
```diff
@@ -13,22 +13,93 @@
...
+
+  auto sharded_ids_input = getShardedIterDomains(input);
+  auto sharded_ids_output = getShardedIterDomains(output);
+  std::vector<IterDomain*> shard_additions;
+  std::vector<IterDomain*> shard_deletions;
+  auto rootmap = PairwiseRootDomainMap(input, output);
+
+  const auto c2p_map = rootmap.mapConsumerToProducer(&sharded_ids_output);
+  for (auto [id1, id2] : c2p_map) {
+    if (!id1->sameAs(id2)) {
```
wujingyue: I don't follow the implementation here and I'm sure I'm missing something. For example, 
```
[DIDx(i0) DIDy(i1)] -> [i0 DIDy(i1)] deletions: i0
```
The c2p_map will contain both i0 and i1. Because the parallel type changed, i0 will be added to shard_additions even though it's a deletion?

csrc/multidevice/utils.cpp
```diff
@@ -13,22 +13,93 @@
...
+  auto input = expr->inputs().at(0)->as<TensorView>();
+
+  auto sharded_ids_input = getShardedIterDomains(input);
+  auto sharded_ids_output = getShardedIterDomains(output);
+  std::vector<IterDomain*> shard_additions;
+  std::vector<IterDomain*> shard_deletions;
+  auto rootmap = PairwiseRootDomainMap(input, output);
+
+  const auto c2p_map = rootmap.mapConsumerToProducer(&sharded_ids_output);
+  for (auto [id1, id2] : c2p_map) {
```
wujingyue, line 86: To improve determinism for testing and debugging, I'd suggest not iterate through an unordered_map like this. Instead, you can iterate through the keys in a deterministic order (e.g. `for id : out->getRootDomain()`) and use the unordered_map just for lookup.

csrc/multidevice/utils.cpp
cowanmeg: When checking for additions I build the mapConsumerToProducer for only the sharded IterDomains of the output.
The c2p map contains (DIDy(i1), DIDy(i1)), so there are no sharding additions.

For deletions we use the mapProducertoConsumer map for only the sharded IterDomains of the input. 
The p2c map contains (DIDx(i0), i0) and (DIDy(i1), DIDy(i1)), so we can detect i0 is unsharded / it's sharding was deleted.

wujingyue: I was blind -- thank you!

csrc/multidevice/utils.cpp
```diff
@@ -13,22 +13,101 @@
...
+  auto output = expr->outputs().at(0)->as<TensorView>();
+  auto input = expr->inputs().at(0)->as<TensorView>();
+
+  auto sharded_ids_input = getShardedIterDomains(input);
+  auto sharded_ids_output = getShardedIterDomains(output);
+  std::vector<IterDomain*> shard_additions;
+  std::vector<IterDomain*> shard_deletions;
+  auto rootmap = PairwiseRootDomainMap(input, output).mapBroadcast(false);
+
+  const auto c2p_map = rootmap.mapConsumerToProducer(&sharded_ids_output);
```
wujingyue, line 85: I **think** we can simplify the code

```
for (IterDomain* out_root : output->getRootDomain()) {
  IterDomain* in_rfactor = c2p_map.at(out_root);
  // Not sure if out_root can be missing. Looks impossible for set and reduce. 
  if (in_rfactor is device and out_root is serial) {
    shard_deletions.push_back(out_root);
  }
  if (in_rfactor is serial and out_root is device) {
    // I think I'm missing a corner case about reduction. Feel free to change the condition here. 
    sharded_additions.push_back(out_root);
  }
}
```

This way, we
1. map once instead of two
2. do one loop instead of two
3. don't need getShardedIterDomains helper, which iterates all IterDomains anyway so it's not any faster.

cowanmeg: This will work with the corner case reduction checks. I agree with your comment that out_root will not be missing any ids. 

It will also address your non-determinism comment above.

wujingyue (commented):
Almost done -- insertShardedAxisReordering is the last piece I'm yet to review!

csrc/multidevice/utils.cpp
```diff
@@ -151,6 +217,106 @@ void insertReshardings(Fusion* fusion) {
...
+      TensorView* input_permute = permute(input, {{idx, 0}});
+      TensorView* output_permute = set(input_permute);
+      TensorView* new_output = permute(output_permute, {{0, idx}});
+      ir_utils::replaceValInAllExprInputsAndFusionOutputs(output, new_output);
+      for (int i = 0; i < static_cast<int>(output->nDims()); i++) {
+        new_output->axis(i)->parallelize(output->axis(i)->getParallelType());
+      }
+      new_output->setDeviceMesh(output->getDeviceMesh());
+      shardAllLike(input, {input_permute});
+      shardAllLike(new_output, {output_permute});
```
wujingyue, line 269: Did you try / Do you mind trying to simply `shardAllLike(input, {created tvs})`? It'd be a lesson for me if `parallelAllLike` fails to propagate parallel dimensions across `permute`.

csrc/multidevice/utils.cpp
```diff
@@ -151,6 +217,106 @@ void insertReshardings(Fusion* fusion) {
...
+
+    // For gather operations i.e. ID goes from sharded to unsharded
+    // this will rematerialize a sharded axis.
+    // ProcessGroup expects contiguous tensors.
+    // Update input to push the rematerialized axis to the front -> collective
+    // -> permute the rematerizlied axis to the proper location Example: [i0
+    // DIDx(i1)] -> [i0 i1] Rewritten to: [DIDx(i1) i0] -> [i1 i0] -> [i0 i1]
+    // Note: there are no reduction based collectives that materializes an axis
+    // so expr is guaranteed to be a set.
+    if (!shard_deletions.empty()) {
```
wujingyue, line 254: Is it possible to have `[DID(i0),i1] => [r0,i1]`? I think it's called [AllReduce](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/operations.html#allreduce) and the code below in this if-then branch won't be able to handle it by doing just `set`.

csrc/multidevice/utils.cpp
```diff
@@ -151,6 +217,106 @@ void insertReshardings(Fusion* fusion) {
...
+      if (isOutermostAllocatedId(output, id)) {
+        continue;
+      }
+
+      TensorView* input_permute = permute(input, {{idx, 0}});
+      TensorView* output_permute = nullptr;
+      // For reduce scatter, determine if the reduction axis shifted to the
+      // right by 1.
+      auto red_axis = output->getReductionAxis();
+      int offset =
```
wujingyue, line 289: It took me a while to understand what this does. I'd add an example in the code comment or modify [your previous example](https://github.com/NVIDIA/Fuser/pull/1939/files/fee8cfa42b43cebd94eb62f81c21e9c7789e5e5f#diff-4739b2bc50e3c078541f1ea6665370ea6dd853d1fa0b4d1085b95d7c4791d96dR274) to something like
```
i0, D(i1), i2 => i0, r1, D(i2)
```
to demonstrate why computing the offset.

wujingyue, line 289:
```suggestion
      int num_reduction_axes_before_sharding_axis =
```

csrc/multidevice/utils.cpp
```diff
@@ -151,6 +217,106 @@ void insertReshardings(Fusion* fusion) {
...
+          (red_axis.has_value() && idx > static_cast<int>(red_axis.value()))
+          ? 1
+          : 0;
+      if (expr->isA<ReductionOp>()) {
+        int raxis = static_cast<int>(red_axis.value()) + offset;
```
wujingyue, line 294:
```suggestion
        int reduction_axis_after_permute  = static_cast<int>(red_axis.value()) + offset;
```

csrc/multidevice/utils.cpp
```diff
@@ -151,6 +217,106 @@ void insertReshardings(Fusion* fusion) {
...
+    }
+    // For scatter operations i.e. ID goes from unsharded to sharded
+    // Update input to push the scattered axis to the front -> collective ->
+    // permute the sharded axis to the proper location.
+    // Example: [i0 i1 DIDx(i2)] -> [i0 DIDx(i1) r2]
+    // Rewritten to: [i0 i1 DIDx(i2)] -> [i1 i0 DIDx(i2)] ->
+    //                    [DIDx(i1) i0 r2] -> [i0 DIDx(i1)]
+    else if (!shard_additions.empty()) {
+      auto id = shard_additions[0];
+      int idx = static_cast<int>(output->domain()->posOf(id));
```
wujingyue, line 279:
```suggestion
      int sharding_axis = static_cast<int>(output->domain()->posOf(id));
```

I believe `axis` and `idx` here are used interchangeably, but better be consistent. I added `sharding_` because code below also talks about reduction axis, and `sharding_`  makes the name less ambiguous.

wujingyue (commented):
LGTM with comments. Great work! I'll take another look before submit.

csrc/multidevice/utils.cpp
cowanmeg: This worked! It will get tested more thoroughly when we have DIDy, DIDz shardings to propagate later.

csrc/multidevice/utils.cpp
cowanmeg: Updated example

csrc/multidevice/utils.cpp
```diff
@@ -13,22 +13,88 @@
...
+
+  for (IterDomain* out_root : output->getRootDomain()) {
+    IterDomain* in_root = c2p_map.at(out_root);
+    // Ignore sharded reductions on the output
+    // ex. DIDx(i0) -> r(i0) or DIDx(i0) -> r(DIDx(i0))
+    // since they don't affect allocation.
+    if (in_root->isDeviceDim() && !out_root->isParallelized() &&
+        !out_root->isReduction()) {
+      shard_deletions.push_back(in_root);
+    } else if (!in_root->isParallelized() && out_root->isDeviceDim()) {
```
naoyam, line 82: Does this mean this function assumes no TIDX/BIDX parallelization is yet done?

csrc/multidevice/utils.cpp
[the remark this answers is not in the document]
cowanmeg: Good catch, that could cause an error in the future. Updated to `rootPosOf`

csrc/multidevice/utils.cpp
cowanmeg: Correct.
Could something like this be allowed? `TID(i0)` -> `DIDx(i0)`
We could update to `(!in_root->isDeviceDim() && out_root->isDeviceDim())`

naoyam: That should have no problem.

naoyam (approved):
LGTM. Please make sure to get a stamp by @wujingyue as well.

csrc/multidevice/utils.h
```diff
@@ -45,4 +44,16 @@ void unshard(TensorView*);
...
+// Assumes all resharding ops are either a set or reduction.
+// For each resharding operation that requires communication
+// over a noncontiguous slices of the tensor, this pass
+// inserts permutations necessary to push the device parallel axis
+// to the front so that communication operations are contiguous.
+void insertShardedAxisReordering(Fusion* fusion);
+
+// Returns the index of the a sharded axis if none return -1.
+// TODO: Assumes no merges/splits on sharded axis.
+int64_t getShardedAxis(TensorView*);
```
wujingyue, line 57:
```suggestion
// Returns the index of the first sharded axis or -1 if none.
// TODO: Assumes no merges/splits on sharded axis.
int64_t getFirstShardedAxis(TensorView*);
```

wujingyue (approved):

csrc/multidevice/utils.h
wujingyue: Nit: I believe we will support multiple sharded axes in the not too distant future. When that happens, it's useful to remember by name this function returns merely the first sharded axis.

samnordmann: If I'm correct we should only accept one axis assigned to one parallel type, i.e., two axis of a given tensor cannot be both parallelized on DIDx, but rather one can be parallelized DIDx and the other on DIDy. I think there is the same reasonable asumption for TID, BID.
So I think the comment shouldn't mention "first", but rather "DIDx", and later it could take the parallel type as an argument

csrc/multidevice/utils.cpp
```diff
@@ -151,6 +217,116 @@ void insertReshardings(Fusion* fusion) {
...
+      TensorView* input_permute = permute(input, {{sharding_axis, 0}});
+      TensorView* output_permute = nullptr;
+      // Calculate the number of reduction axes before the sharding axis.
+      // After permuting the sharding axis to the front, the reduction axis
+      // will be offset by this amount.
+      auto reduction_axis = output->getReductionAxis();
+      int num_reduction_axes_before_sharding_axis =
+          (reduction_axis.has_value() &&
+           sharding_axis > static_cast<int>(reduction_axis.value()))
+          ? 1
```
samnordmann, line 297: I guess we implicitely assume here that there can be maximum one reduction axis?
Can you please make this explicity by adding an assert ? see https://github.com/NVIDIA/Fuser/blob/9c2359fe3f7655a0e4726a06ba186ecb58a5c2e4/csrc/scheduler/mma_utils.cpp#L1444

csrc/multidevice/utils.cpp
```diff
@@ -151,6 +217,116 @@ void insertReshardings(Fusion* fusion) {
...
+            static_cast<int>(reduction_axis.value()) +
+            num_reduction_axes_before_sharding_axis;
+        auto red_expr = dynamic_cast<ReductionOp*>(expr);
+        output_permute = reductionOp(
+            red_expr->getReductionOpType(),
+            {reduction_axis_after_permute},
+            red_expr->init(),
+            input_permute);
+      } else {
+        // Note this is a no-op and is moving a device parallel axis back
```
samnordmann, line 310: I am not sure to understand this comment.
It is not a no-op since it embedds a scatter, right ?

csrc/multidevice/utils.cpp
```diff
@@ -151,6 +217,116 @@ void insertReshardings(Fusion* fusion) {
...
+      IterDomain* shard_deleted_id = shard_deletions[0];
+      int sharding_axis =
+          static_cast<int>(input->domain()->rootPosOf(shard_deleted_id));
+      if (isOutermostAllocatedId(input, shard_deleted_id)) {
+        continue;
+      }
+      TensorView* input_permute = permute(input, {{sharding_axis, 0}});
+      TensorView* output_permute = set(input_permute);
+      TensorView* new_output = permute(output_permute, {{0, sharding_axis}});
+      ir_utils::replaceValInAllExprInputsAndFusionOutputs(output, new_output);
```
samnordmann, line 264: I am afraid there is a problem here if a tensor is the input of two resharding expr. In this case the second expr will be replaced by a new one after this line, and so the for loop will not iterate over it. Am I correct?

Same below for the other branch

csrc/multidevice/utils.cpp
```diff
@@ -151,6 +217,116 @@ void insertReshardings(Fusion* fusion) {
   }
 }
 
+void insertShardedAxisReordering(Fusion* fusion) {
+  auto exprs = fusion->exprs();
+  std::vector<Expr*> reshard_exprs;
+  for (auto expr : exprs) {
+    if (isResharding(expr)) {
```
samnordmann, line 224: could we add another check at this point to verify whether the expr needs axis reordering?
I understand that this check is somwhow done later by calling `isOutermostAllocatedId`, but it would be nice to have one method returning a bool wether an I/O sharded axis is non-outermost.

This way, we could add an assert in `isLowerableToCommunication` to make sure at the symbolic level that there is no communications on non-outermost domain.

tests/cpp/test_multidevice_pipeline.cpp
```diff
@@ -239,7 +236,7 @@ INSTANTIATE_TEST_SUITE_P(
         testing::Values(true),
         testing::Values(false),
         testing::Values(false),
-        testing::Values(0),
+        testing::Values(0, 1),
```
samnordmann, line 239: Could you also add different non-outermost resharding in this test https://github.com/NVIDIA/Fuser/blob/9c2359fe3f7655a0e4726a06ba186ecb58a5c2e4/tests/cpp/test_multidevice_resharding.cpp#L230C1-L230C46 ?

This test doesn't run the multidevice fusion (we should probably change that btw) but checks many more scenarios at the symbolic level. If `isLowerableToCommunication` is updated as I suggest in another comment, this test should spot if there are issues (possibly with `replaceValInAllExprInputsAndFusionOutputs`)

samnordmann (commented):
Thanks a lot for this PR!

As a note here for the record, I want to highlight that the present implemented strategy for handling non-outermost could be refered to as `bcopy`, since we induce extra copies in/out to allow the communication to operate on one contiguous buffer. This is a good strategy for small msg sizes.

The `zcopy` algorithm consists in posting as many collectives as there are contiguous buffers.  The `zcopy` algorithm will be needed for large msg sizes, but also for multi-dimensional parallelism: e.g., to pass from `[DIDx(i0), DIDx(i1)]` to `[DIDx(i0), i1]`, as is done in SUMMA matmul. Therefore, AFAICS, `zcopy` is somehow a pre-requesite to 2D parallelism.

Eventually we will need both `zcopy` and `bcopy` algos, and a mechanism to (manually or automatially) choose between them.

csrc/multidevice/utils.cpp
```diff
@@ -13,22 +13,88 @@
...
       tv->getLeafDomain().end(),
-      std::back_inserter(sharded_ids),
+      std::inserter(sharded_ids, sharded_ids.begin()),
       [](auto id) { return id->isDeviceDim(); });
   return sharded_ids;
 }
+
+// Returns whether a IterDomain in a TensorView is the outermost
+// allocated IterDomain in the TensorView.
+bool isOutermostAllocatedId(TensorView* tv, IterDomain* id) {
```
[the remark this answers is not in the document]
cowanmeg: Will skip broadcasts in the function. I am not certain what you mean by inlining?

csrc/multidevice/utils.cpp
```diff
@@ -13,22 +13,88 @@
...
+    // ex. DIDx(i0) -> r(i0) or DIDx(i0) -> r(DIDx(i0))
+    // since they don't affect allocation.
+    if (in_root->isDeviceDim() && !out_root->isParallelized() &&
+        !out_root->isReduction()) {
+      shard_deletions.push_back(in_root);
+    } else if (!in_root->isParallelized() && out_root->isDeviceDim()) {
+      shard_additions.push_back(out_root);
+    } else if (in_root->isDeviceDim() && out_root->isDeviceDim()) {
+      NVF_ERROR(
+          in_root->getParallelType() == out_root->getParallelType(),
```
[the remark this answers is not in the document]
cowanmeg: This is just temporary. If we add a lowering pathway that maps this to a collective then it will be removed.

csrc/multidevice/utils.cpp
```diff
@@ -13,22 +13,88 @@
...
+// i.e. sharded IterDomains that are present in the input, but not the output.
+// TODO: Analyze leaf domain for unsharded/sharded IDs and return their
+// parent root IDs.
+std::pair<std::vector<IterDomain*>, std::vector<IterDomain*>>
+getShardingChanges(Expr* expr) {
+  NVF_ERROR(
+      expr->outputs().size() == 1,
+      "Resharding operations can only have one output");
+  NVF_ERROR(
+      expr->inputs().size() == 1,
```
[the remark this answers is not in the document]
cowanmeg: Correct, this analysis depends on insert resharding ops pass to be applied on the fusion first.

csrc/multidevice/utils.cpp
```diff
@@ -151,6 +217,116 @@ void insertReshardings(Fusion* fusion) {
...
+        continue;
+      }
+      TensorView* input_permute = permute(input, {{sharding_axis, 0}});
+      TensorView* output_permute = set(input_permute);
+      TensorView* new_output = permute(output_permute, {{0, sharding_axis}});
+      ir_utils::replaceValInAllExprInputsAndFusionOutputs(output, new_output);
+
+      // Propagate shardings from input and manually apply sharding deletions.
+      shardAllLike(input, {input_permute, output_permute, new_output});
+      output_permute->axis(0)->parallelize(ParallelType::Serial);
```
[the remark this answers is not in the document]
cowanmeg: Since this is a gather, we know the axis 0's parallel type must be serial here since it will be moving from DeviceDim to Serial. 
We have to explicitly set it to serial, because the propagation from input preserves the parallel types.

csrc/multidevice/utils.cpp
cowanmeg: good catch, the permute after this is the view change

csrc/multidevice/utils.cpp
cowanmeg: The expression that take `output` as input will not be replaced, instead `new_output` will be the input tv instead of `output`. `new_output` is also identical to `output` just generated by a different expression.

csrc/multidevice/utils.cpp
```diff
@@ -151,6 +217,116 @@ void insertReshardings(Fusion* fusion) {
...
+    // Note: there are no reduction based collectives that materializes an axis
+    // so expr is guaranteed to be a set.
+    if (!shard_deletions.empty()) {
+      IterDomain* shard_deleted_id = shard_deletions[0];
+      int sharding_axis =
+          static_cast<int>(input->domain()->rootPosOf(shard_deleted_id));
+      if (isOutermostAllocatedId(input, shard_deleted_id)) {
+        continue;
+      }
+      TensorView* input_permute = permute(input, {{sharding_axis, 0}});
```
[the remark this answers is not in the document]
cowanmeg: This permute is moving a DeviceDim axis to position 0, so it's actually guaranteed to be DeviceDim. We know that in the input this axis must be DeviceDim and in the output it must be Serial since the expression is unsharding that axis.

csrc/multidevice/utils.cpp
```diff
@@ -151,6 +217,116 @@ void insertReshardings(Fusion* fusion) {
...
+        "Resharding operations can only have one output");
+    NVF_ERROR(
+        expr->inputs().size() == 1,
+        "Resharding operations can have only one input");
+    auto output = expr->outputs().at(0)->as<TensorView>();
+    auto input = expr->inputs().at(0)->as<TensorView>();
+    auto [shard_additions, shard_deletions] = getShardingChanges(expr);
+    NVF_ERROR(
+        shard_additions.size() + shard_deletions.size() <= 1,
+        "Resharding expr can only support one axis")
```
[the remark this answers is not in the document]
cowanmeg: Probably temporary, the check is in place just because that is all we can support currently.

PR #1946, opened 2024-03-14
Multidevice test cleanup

cowanmeg:
Refactoring and cleanup of multidevice tests.

1. test_pipeline.cpp -> test_multidevice_resharding.cpp and include in the test_multidevice executable.
2. Update shardTensor helper function.
3. Parameterizing sharding test and two stage pipeline tests with sharded dimension in preparation for sharding non-outermost axis.
4. Skip PipelineTest.Pipeline when there are less than 6 devices.


wujingyue:
> Skip PipelineTest when there are less than 6 devices.

What's the reason behind? I'm asking because that would make testing even harder. How many dlcluster nodes have >=6 GPUs?

tests/cpp/multidevice.h
```diff
@@ -48,15 +49,22 @@ class MultiDeviceTest : public NVFuserTest {
  public:
   static at::Tensor shardTensor(
```
wujingyue, line 50: Can you add a comment on what this function does and what each argument means?

cowanmeg:
> > Skip PipelineTest when there are less than 6 devices.
> 
> What's the reason behind? I'm asking because that would make testing even harder. How many dlcluster nodes have >=6 GPUs?

That particular test uses hardcoded non-standard meshes like {4, 5} and {1, 3, 4}.
There are a few older V100 nodes with 6 GPUs in dlcluster, but we can re-write this test to 4 GPUs if it becomes an issue for testing.

All other tests require up to 4 GPUs or use a mesh of the MPI world size.

tests/cpp/multidevice.cpp
```diff
@@ -121,11 +121,10 @@ void PipelineTest::validate() {
     if (!output_tv->getDeviceMesh().has(communicator->deviceId())) {
       continue;
     }
-    auto ref_output = isSharded(output_tv) ? shardTensor(
-                                                 ref_unsharded_outputs.at(i),
-                                                 output_tv->getDeviceMesh(),
-                                                 communicator->deviceId())
-                                           : ref_unsharded_outputs.at(i);
+    auto ref_output = isSharded(output_tv)
```
wujingyue, line 124: With your PR, shardTensor already returns the original tensor when the tv is not sharded. So should we simplify?

tests/cpp/test_multidevice_pipeline.cpp
```diff
@@ -52,6 +52,10 @@ using namespace at::indexing;
 */
 
 TEST_F(PipelineTest, Pipeline) {
+  if (communicator->size() < 6 || torch::cuda::device_count() < 6) {
```
wujingyue, line 55: Should `communicator->size()` and `torch::cuda::device_count()` be the same? Iow, what would break if I say 

```
const auto device_count = torch::cuda::device_count();
NVF_ERROR(communicator->size() == device_count);
if (device_count < 6) {
```

tests/cpp/test_multidevice_pipeline.cpp
```diff
@@ -132,9 +136,16 @@ TEST_F(PipelineTest, Pipeline) {
...
+// stage sharded?, do_reduction?, sharded dimension, use_fusion_executor_cache?)
+using PipelineTestTwoStagesParams = std::tuple<
+    CommunicatorBackend,
+    DeviceMesh,
+    DeviceMesh,
+    bool,
+    bool,
+    bool,
+    int,
+    bool>;
```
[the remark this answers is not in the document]
cowanmeg: This is a good suggestion, this test's params keep growing as we add new functionality. I'll add it now :D

tests/cpp/test_multidevice_pipeline.cpp
cowanmeg: It should be `communicator->size() <=  torch::cuda::device_count()` since it's fine to use less than the available number of devices. I'll add this check the `MutiDeviceEnvinroment::SetUp()` since if we try to create a communicator thats greater than the number of devices everything breaks.  Good catch!

samnordmann: It is not the same because communicator->size is the number of process spwaned, not number of GPU. Also (to be confirmed) torch::cuda::device_count() only counts the devices on the current node

cowanmeg: oh I forgot device_count() only returns the number of devices on the current node. I will not add that check then.

samnordmann: I think it is not necessary to add those tests, which are all already included in `MultiDeviceExecutor::validate()`. This function is called inside `PipelineTest::execute()` -- the only missing part is to propagate the skip outside the function to skip executing `PipelineTest::validate()`

does it make sens? :)

tests/cpp/multidevice.h
```diff
@@ -48,15 +49,22 @@ class MultiDeviceTest : public NVFuserTest {
...
       DeviceIdxType deviceId) {
-    int i = 0;
-    auto devices = mesh.vector();
-    auto it = find(devices.begin(), devices.end(), deviceId);
-    if (it != devices.end()) {
-      i = std::distance(devices.begin(), it);
+    if (isSharded(tv)) {
+      auto sharded_dim = 0;
+      int i = 0;
+      auto devices = tv->getDeviceMesh().vector();
```
samnordmann, line 57: nit: `const auto&`

tests/cpp/multidevice.h
```diff
@@ -48,15 +49,22 @@ class MultiDeviceTest : public NVFuserTest {
...
+      auto it = std::find(devices.begin(), devices.end(), deviceId);
+      if (it != devices.end()) {
+        i = std::distance(devices.begin(), it);
+      }
+      std::vector<at::indexing::TensorIndex> indices(
+          tensor.dim(), at::indexing::Slice());
+      indices[sharded_dim] = at::indexing::Slice(i, i + 1);
+      return tensor.index(indices).contiguous();
```
samnordmann, line 65: For my own understanding: is it really necessary to call `contiguous()`? I guess this is done in order to make the contiguity match between this reference tensor match and the obtained sharded tensor. So this is correct if and only if the sharded tensor is also contiguous, right?

tests/cpp/test_multidevice_pipeline.cpp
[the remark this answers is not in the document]
samnordmann: It totally make sense to define structs here, but as far as I can see this won't avoid things like `testing::Values(false)` in the test instantiation
But correct me if Im wrong !

tests/cpp/test_multidevice_pipeline.cpp
wujingyue: > the only missing part is to propagate the skip outside the function

$0.02: to simplify, it may even be possible to GTEST_SKIP inside execute(), although I haven't tried this and [the doc](https://google.github.io/googletest/advanced.html#skipping-test-execution) is unclear.

cowanmeg: I don't think this is possible. Since `execute()` is calling GTEST macros it has to be a void function, so we can't return the error message from `execute` back to the test.

tests/cpp/multidevice.h
cowanmeg: `index` returns a view to avoid copying the tensor. `contiguous()` just ensures if the slice isn't contiguous it will create a new tensor and copy data.

samnordmann: got it!

tests/cpp/test_multidevice_pipeline.cpp
samnordmann: > I don't think this is possible. Since `execute()` is calling GTEST macros it has to be a void function, so we can't return the error message from `execute` back to the test.

I would use a boolean class data member

tests/cpp/test_multidevice_pipeline.cpp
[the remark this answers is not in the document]
wujingyue: I was thinking about:
```
struct Config {
  knob1;
  knob2;
};

testing::Values({.knob1=..., .knob2=...}, {.knob1=..., .knob2=...})
```
or even
```
testing::ValuesIn(AFunctionThatGeneratesAVectorOfConfigs());
```

If `testing::Combine` is too handy, another way to enforce names is to use enum (e.g. `enum class ShardingType`) instead of booleans. After all, my main concern is that constants are floating around without names and types, so it's very easy to put `testing::Values(...)` in the wrong order.

tests/cpp/test_multidevice_pipeline.cpp
cowanmeg: We could do that, but imo it's cleaner to perform the check in the test rather than add a new class field.

samnordmann: The check should be perform in all the tests, not only this one. So in my opinion it is more convenient to have it automatically done at one place instead of duplicating it everywhere with hardcoded values, even if it is at the cost of one supplmentary class member.

Another good solution would be to merge `execute` and `validate` into one function -- they are always used together for now anyway, and it would save us the method executeAndValidate

tests/cpp/test_multidevice_pipeline.cpp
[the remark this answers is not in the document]
samnordmann: In my opinion Combine is very convenient here so I would keep it even if I admit that it is at the cost of readability when instantiating the test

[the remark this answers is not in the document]
samnordmann: Another drawback of using structs or enum class is that we will need to overload the `<<` operator for each created type for readable prints, whereas `Combine` is a native gtest and verything gets printed nicely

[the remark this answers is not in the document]
cowanmeg: Googletest supports the struct parameters and `testing::combine`. It is not the cleanest design. I've pushed some code, but it doesn't get rid of `testing::Values(false)` since we are still using combine. 
To improve the readability we can add comments to the current test instantiations like:
```
testing::Values(true), // first_stage_sharded
testing::Values(true), // second_stage_sharded
testing::Values(false), // do_reduction
```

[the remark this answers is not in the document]
cowanmeg: We can revert back to using the original tuple parameters, since I don't think this improved the readability that much.

tests/cpp/test_multidevice_pipeline.cpp
cowanmeg: Unfortunately, GTEST_SKIP only skips tests when called in either the enivronment's SetUp or the unit test itself, so combining execute and validate into one function and calling GTEST_SKIP will not skip the test. 

I can save the error message from validate and check it's results after each call executeValidate, but we will have to make certain to check it after executeValidate in each test otherwise the message won't get set.

tests/cpp/multidevice.cpp
```diff
@@ -121,11 +121,8 @@ void PipelineTest::validate() {
...
     }
-    auto ref_output = isSharded(output_tv) ? shardTensor(
-                                                 ref_unsharded_outputs.at(i),
-                                                 output_tv->getDeviceMesh(),
-                                                 communicator->deviceId())
-                                           : ref_unsharded_outputs.at(i);
+    auto ref_output = shardTensor(
+        ref_unsharded_outputs.at(i), output_tv, communicator->deviceId());
     auto obtained_output = outputs.at(i);
     GTEST_EXPECT_TRUE(torch::allclose(ref_output, obtained_output))
```
wujingyue, line 127: I know it's pre-existing, but why `GTEST_ASSERT/EXPECT_*` instead of simply `ASSERT/EXPECT_*`?

tests/cpp/test_multidevice_pipeline.cpp
wujingyue: > I can save the error message from validate and check it's results after each call executeValidate, but we will have to make certain to check it after executeValidate in each test otherwise the message won't get set.

We could just let execute or executeValidate return the error message (or "") to the caller (i.e. the test case itself), which is then responsible to call `GTEST_SKIP`. Would that work? 

That being said, if easy, I prefer GTEST_SKIP up front in the test so it's immediately clear to the reader why a test is skipped. Currently, it's unclear to the reader (like me) why MultiDeviceExecutor::validate would fail. If the skip condition is indeed as simple as device_count<6, GTEST_SKIP that up front brings lots of clarity.

tests/cpp/test_multidevice_pipeline.cpp
```diff
@@ -147,6 +177,7 @@ TEST_P(PipelineTestTwoStages, Communication) {
        is_stage0_sharded,
        is_stage1_sharded,
        do_reduction,
+       sharded_dim,
        use_fusion_executor_cache] = GetParam();
   if (!disable_skip && !communicator->isBackendAvailable(backend)) {
```
wujingyue, line 182: What's `disable_skip` for? Is it a duplicate of the existing `--gtest_also_run_disabled_tests`?

tests/cpp/test_multidevice_pipeline.cpp
```diff
@@ -157,24 +188,22 @@ TEST_P(PipelineTestTwoStages, Communication) {
     mesh1 = mesh0;
   }
 
-  int first_axis_extent = 3;
+  std::vector<int64_t> unsharded_input_sizes = {3, 2, 3, 5};
   if (is_stage0_sharded) {
-    first_axis_extent = mesh0.vector().size();
-  } else if (is_stage1_sharded) {
-    first_axis_extent = mesh1.vector().size();
+    unsharded_input_sizes[sharded_dim] = mesh0.vector().size();
```
wujingyue, line 193: I don't understand this sentence. We are replacing the extent of an axis with a mesh rank? They don't seem to be logically connected. 

(This is for my education and not a problem with a PR. The old code is as mysterious to me :)

tests/cpp/test_multidevice_sharding.cpp
```diff
@@ -147,20 +143,21 @@ TEST_P(ShardingTest, UnshardedGlobalInput) {
 // Test memory allocation of multidevice fusion with sharded input
 // and replicated intermediates and output.
 TEST_P(ShardingTest, ShardGlobalInput) {
-  auto concreteTv = GetParam();
-  int sharded_dim = 0;
+  auto [concreteTv, sharded_dim] = GetParam();
```
wujingyue, line 146:
```suggestion
  auto [creates_concrete_tensor, sharded_dim] = GetParam();
```

tests/cpp/test_multidevice_sharding.cpp
```diff
@@ -147,20 +143,21 @@ TEST_P(ShardingTest, UnshardedGlobalInput) {
...
+  unsharded_input_size[sharded_dim] = num_devices;
 
-  TensorView* tv0 = concreteTv ? makeConcreteTensor(unsharded_input_size)
-                               : makeContigTensor(3);
+  TensorView* tv0 = concreteTv
+      ? makeConcreteTensor(unsharded_input_size)
+      : makeSymbolicTensor(unsharded_input_size.size());
   TensorView* tv1 = set(tv0);
-  TensorView* tv2 = add(tv1, tv1);
+  TensorView* tv2 = sum(tv1, {1});
```
wujingyue, line 160: This seems to be a behavior change not a cleanup as the title suggested. Is there a reason behind this?

wujingyue: (The same comment goes to line 175)

wujingyue (commented):
LGTM with comments. Thanks for the cleanup! Although I sensed more cleanups, the PR alone is a strict improvement in several ways. I'll take another look after you address the comments.

tests/cpp/test_multidevice_pipeline.cpp
wujingyue: I'm asking this because tests should be skipped for good reasons, e.g., the test requires 6 GPUs but only 2 found. I don't understand why one would attempt to disable the skipping. If a test currently fails because of a bug, the test should be marked as `DISABLED_` and `--gtest_also_run_disabled_tests` is the right way to enforce to run them.

tests/cpp/test_multidevice_pipeline.cpp
cowanmeg: It's replacing the extent of the exist with the extent of the mesh.  We have a constraint that the extent of a device parallel axis == number of devices.
When we support split it will be less clunky.

tests/cpp/test_multidevice_sharding.cpp
cowanmeg: yeah, these are changes from some random testing - will revert back.

tests/cpp/test_multidevice_pipeline.cpp
cowanmeg: It was a flag used for debugging, so that we could run tests with 1 device and only see 1 processes' print statements.
I didn't know about this flag in gtest, so we can use that instead and remove the disable_skip! FYI @samnordmann just be aware of this update

wujingyue: > we could run tests with 1 device and only see 1 processes' print statements.

Interesting. In that case, does that process run the program on sharded tensors? If yes, how does stuff like allgather run? If no, should we make one-device a test that's always enabled?

cowanmeg: The program will not run correctly! 
This is helpful for debugging the test when the bug is in compilation before it executes.

cowanmeg: > If no, should we make one-device a test that's always enabled?
I think we could benefit from some more one-device only tests that check the behavior of compilation without execution. We have a couple, but as a TODO item we should try and improve the one-device test coverage where possible.

tests/cpp/multidevice.h
```diff
@@ -46,17 +47,29 @@ class MultiDeviceEnvironment : public testing::Environment {
...
+      int i = 0;
+      const auto& devices = tv->getDeviceMesh().vector();
+      auto it = std::find(devices.begin(), devices.end(), deviceId);
+      if (it != devices.end()) {
+        i = std::distance(devices.begin(), it);
+      }
+      std::vector<at::indexing::TensorIndex> indices(
+          tensor.dim(), at::indexing::Slice());
+      indices[sharded_dim] = at::indexing::Slice(i, i + 1);
+      return tensor.index(indices).contiguous();
```
wujingyue, line 65:
```suggestion
      return tensor.slice(sharded_dim, i, i + 1).squeeze(sharded_dim).contiguous();
```
and then you don't need `indices`. I think this is simpler code that does the same thing. `at::index` is not very convenient when dealing with slices.

tests/cpp/test_multidevice_pipeline.cpp
samnordmann: This flag is a debug utility for when we need to debug something happening during lowering (which is often the case), and to allow us to use gdb (because debuggin distributed program is very annoying). So, no network communication will be executed, the program will crash at some point, but it's still a good debugging tool that I'm using on a regular basis

wujingyue (approved):

tests/cpp/test_multidevice_pipeline.cpp
cowanmeg: Resolved offline with Sam. GTEST_SKIP is still in the helper function `executeAndValidate` and working properly now.

tests/cpp/test_multidevice_pipeline.cpp
samnordmann: > I don't understand this sentence. We are replacing the extent of an axis with a mesh rank? They don't seem to be logically connected.

The axis and device mesh extent are logically eauqlly. It is exactly the same as binding some axis extent to gridDim.x or blockDim.x. But here block/grid is replaced by the device mesh. Does it make sense ?

tests/cpp/test_multidevice_pipeline.cpp
```diff
@@ -147,8 +154,9 @@ TEST_P(PipelineTestTwoStages, Communication) {
        is_stage0_sharded,
        is_stage1_sharded,
        do_reduction,
+       sharded_dim,
        use_fusion_executor_cache] = GetParam();
-  if (!disable_skip && !communicator->isBackendAvailable(backend)) {
+  if (!communicator->isBackendAvailable(backend)) {
```
samnordmann, line 159: why removing `disable_skip`?

tests/cpp/multidevice.h
```diff
@@ -33,30 +34,37 @@ class MultiDeviceEnvironment : public testing::Environment {
...
-  bool disable_skip_ = false;
 };
 
 class MultiDeviceTest : public NVFuserTest {
  public:
+  // Given an aten tensor, TensorView the tensor is bound to, and deviceId
+  // returns a shard of the tensor according the sharding annotation in tv
+  // for the deviceId. If tensor is not sharded returns the original tensor.
+  // TODO: If deviceId is not part of the mesh this should return an empty
+  // tensor currently, we don't support this, so for now it returns a slice.
```
samnordmann, line 49: for what reason don't we support that?
also, nit: add a "." before currently maybe ?

tests/cpp/multidevice.h
```diff
@@ -88,18 +95,13 @@ class CommunicationTest
...
   // (sharded) outputs with ref_unsharded_outputs. if
   // validate_with_prescribed_values is true, ref_unsharded_outputs is assumed
   // to be set manually in the test body. Otherwise, ref_unsharded_outputs is
   // computed by running a Fusion on a single device with the unsharded_inputs
   void validate(bool validate_with_prescribed_values = false);
-  void executeAndValidate(bool validate_with_prescribed_values = false) {
-    execute();
-    validate(validate_with_prescribed_values);
-  }
+  void executeAndValidate(bool validate_with_prescribed_values = false);
```
samnordmann, line 104: thanks! :')

tests/cpp/multidevice.h
cowanmeg: Global input tensor sizes are used for concretization. We don't have a way to specify global input size separate from the input tensor yet.

samnordmann: Oh right! I see

samnordmann (approved):

PR #1936, opened 2024-03-14
Check for SliceOp before inserting casts

jacobhinkle:
We already check for select-based ops because they must be the first op in the segment. This PR just adds SliceOp to the list of types to check.

This current approach is flawed, since it will cause some segments to not cast at all, leading to increased IO. Instead, we should probably forward the casts through incompatible ops instead of skipping inserting casts altogether.

Fixes #1953.

Issue #1557, opened 2023-12-22
Compilation failure using `segment_set` with half precision inputs


jacobhinkle:
This test fails to compile:
```python
    def test_simple_slice_fusion_bfloat16(self):
        inputs = [torch.randn((10,), dtype=torch.bfloat16, device="cuda:0")]

        def fusion_func(fd: FusionDefinition) -> None:
            T0 = fd.define_tensor(shape=[-1], contiguity=[True], dtype=DataType.BFloat16)
            T1 = fd.ops.neg(T0)
            T2 = fd.ops.segment_set(T1)
            T3 = fd.ops.slice(T2, start_indices=[0], end_indices=[5])
            fd.add_output(T3)

        nvf_out, _ = self.exec_nvfuser(fusion_func, inputs)

        ref = -inputs[0][:5]

        torch.testing.assert_close(nvf_out[0], ref)
```
Generated kernel and error:
```c++
__global__ void nvfuser_pointwise_f0_c1_r0_g0(Tensor<__bfloat, 1, 1> T0, Tensor<__bfloat, 1, 1> T2) {
  nvfuser_index_t i0;
  i0 = ((nvfuser_index_t)threadIdx.x) + (128LL * ((nvfuser_index_t)blockIdx.x));
  if ((i0 < T0.logical_size[0LL])) {
    __bfloat T4[1];
    T4[0] = 0;
    T4[0]
       = T0[i0];
    __bfloat T1[1];
    T1[0]
       = -T4[0];
    __bfloat T5[1];
    T5[0]
       = T1[0];
    T2[i0]
       = T5[0];
  }
}
}

CUDA NVRTC compile error: __tmp_kernel_pointwise_f0_c1_r0_g0.cu(9455): error: no operator "-" matches these operands
            operand types are: - <unnamed>::__bfloat
         = -T4[0];
           ^
```
This is the first segment which should just be negating the bfloat inputs. It looks like casts are not inserted properly.

In the test, we are placing a `segment_set` to prevent "forwarding" the input through unary ops. In this case the overloaded terminology "forwarded" means the segmenter is pretending the output of a chain of unary ops from an input is itself the input. It then replays all the unary ops that were forwarded in segments that use that forwarded input. This is usually a good thing but in this case the replaying interferes with the `slice` op, which requires its inputs to be fusion inputs. So here we insert a `segment_set`, ensuring that the slice is a segment input.

Related to #1553. This was exposed when testing #1556.

[...]

PR #1949, opened 2024-03-15
[WIP] Skip casting inputs to ops that operate on segment inputs

jacobhinkle:
THIS IS INCOMPLETE

This is a better alternative to #1936, which merely skipped casting for a slice segment. That had the effect of removing all downcasting for that segment. Instead, we should allow some ops to operate on segment inputs but still cast their outputs, if needed. This PR implements that. If we encounter ops other than slice or select-based ops, we place a cast just before them, updating the dtypes of any intermediate tensors between the incoming segmentation edge and the op whose inputs we are upcasting.

Note that in the future we could use this to avoid upcasting for other lossless half-precision ops, like neg and abs (see https://github.com/NVIDIA/Fuser/issues/1557#issuecomment-1957833263).

### DIFF
--- a/csrc/ir/interface_nodes.h
+++ b/csrc/ir/interface_nodes.h
@@ -434,6 +434,7 @@ class NVF_API TensorView : public Val {
   friend OptOutMutator;
   friend class InlineBatchingGuard;
   friend class ir_utils::TVDomainGuard;
+  friend class SegmentedFusion;
 
   // Inline the computation of this tensor into its consumer at the given
   // position. If this tensor is already inlined in a higher position, then this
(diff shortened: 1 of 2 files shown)


csrc/ir/interface_nodes.h
```diff
@@ -550,6 +551,10 @@ class NVF_API TensorView : public Val {
     domain_ = td;
   }
 
+  void setDataType(DataType dtype) {
+    dtype_ = dtype;
+  }
+
```
jacobhinkle, line 557: We might have a chain of ops that we skip upcasting for. To avoid replaying that whole chain, I am just changing the dtypes of those intermediate tensors. Currently we have a closed collection of ops, so this is safe. If instead we switched to only upcasting for known ops that require it like arithmetic, then we should be careful that differences in input type only affect the output type and not other aspects of an Expr.

jacobhinkle: Alternatively, we can do the replay and mark replacements. That might be cleaner but there is still the possibility in the future for an op to require more care when changing the dtypes of its inputs.

PR #1950, opened 2024-03-15
Options cannot be thread local

protonu:
I had incorrectly added `thread_local` to the various options. They need to be shared across threads (for now). For example, when running from the Python frontend, the Pybind code (and segmenter) runs in a different thread from the scheduler code. 
I have added mutexes to enable thread safety when threads will modify options.


protonu:
!build

protonu:
!build

jacobhinkle:
Hmm does this mean that one thread is able to modify the options that another thread will see?

Was the problem that thread_local variables are initialized from scratch when the thread is spawned? It seems like we might need to manually set the options within each worker thread when we spawn the threads, so that they "inherit" the main thread options settings.

jjsjann123:
> Was the problem that thread_local variables are initialized from scratch when the thread is spawned? It seems like we might need to manually set the options within each worker thread when we spawn the threads, so that they "inherit" the main thread options settings.

^^^ This sounds like a cleaner approach. Meanwhile, do we really care about having a thread_local flag? I think we need that to justify plumbing that work in the first place. :)

jjsjann123:
!build

protonu:
I'll close this PR. We'll go with keeping these as thread local. The thread local options will inherit their values from the main/spawning thread.

Issue #1275, opened 2023-11-09
Validation tolerances


Priya2698:
Python benchmarks may have flaky validation errors due to small validation tolerances.
For example: https://github.com/NVIDIA/Fuser/pull/1265#issuecomment-1804659119

We may want to relax the tolerance values at https://github.com/NVIDIA/Fuser/blob/333f48632a130ee3db353ca86c1402d4978da2a0/csrc/validator_utils.h#L25 if the errors persist on nightly CI.

Priya2698:
The test errors in PR #1405 also require relaxed tolerances: https://github.com/NVIDIA/Fuser/pull/1405#discussion_r1412837509.

Priya2698:
Some efforts such as computing references in double improved the error rate. However, we have not been able to completely resolve this issue by recomputing error thresholds etc.
The recent benchmarking dashboard should cluster numerical errors separate from functional issue to reduce noise, so I am closing this as unplanned.

PR #1954, opened 2024-03-18
use both shared memory and regiser to store persistent buffer in innerOuter scheduler

liqiangxl:
For details see the [doc](https://docs.google.com/document/d/1A_hooa2wYZA-qlZ1sWFEWHi3Fnj42ceLBRmZdRnFTR4/edit?usp=sharing).
**Issue:** inner_outer scheduler segments when buffer size is larger than register size (for layer norm backward fp16, segments at 18K) , it leads to low performance.

**Fix:** When buffer size is larger than the available register space, move 1 or more buffers from register to shared memory. 
Prioritizing moving the buffer leads to the smallest memory access latency, e.g. weights in layer norm (and RMS norm) backward.

**Results:**: Bandwidth increased about 1.5x to 2.5x when using shared memory can avoid segmentation.


csrc/scheduler/normalization_inner_outer.h
```diff
@@ -26,6 +26,9 @@ class HeuristicSummary;
 
 class InnerOuterPersistentKernelScheduler : public SchedulerEntry {
  public:
+  constexpr static int64_t threads_per_block_min = 128l;
+  constexpr static int64_t threads_per_block_max = 256l;
```
jjsjann123, line 30: where does this 256 coming from?

csrc/scheduler/reduction_heuristic.h
```diff
@@ -133,8 +135,15 @@ class ReductionParams : public HeuristicParams {
   // block_dim_inner_reduction_extra (usually TIDy)
   ParallelType block_dim_inner_reduction_extra = ParallelType::Serial;
 
-  // use shared memory for persistent buffer, if false, will use registers
+  // Use shared memory for persistent buffer, if false, will use registers.
```
jjsjann123, line 138: nitpick: We might want update this comment, as well as the variable name.

The old logic, iiuc, is toggling between `use only shmem` vs `use only register`. But this PR is updating it to extend persistent buffer to both shmem and register.

csrc/scheduler/reduction_heuristic.h
```diff
@@ -133,8 +135,15 @@ class ReductionParams : public HeuristicParams {
...
+  // For innerOuterPersistentHeuristic, only the persistent tensors in the
```
jjsjann123, line 139: This is a TODO. Let's mark it as a TODO.

csrc/scheduler/reduction_heuristic.h
```diff
@@ -310,7 +319,8 @@ class ReductionParams : public HeuristicParams {
   }
 
   std::shared_ptr<HeuristicParams> clone() const override {
-    return std::make_shared<ReductionParams>(*this);
+    auto ptr = std::make_shared<ReductionParams>(*this);
+    return std::static_pointer_cast<HeuristicParams>(ptr);
```
jjsjann123, line 323: out of curiosity, is there a linter or a compiler that's complaining on this?

csrc/scheduler/utils.cpp
```diff
@@ -2490,6 +2492,31 @@ std::unordered_set<TensorView*> getAllTvsFrom(
   return tv_group;
 }
 
+int64_t getSharedMemoryOverheadPerBlock(
+    Fusion* fusion,
+    const std::vector<TensorView*>& reduction_tvs) {
+  const auto& dev_prop = at::cuda::getCurrentDeviceProperties();
+  const int64_t max_threads_per_block = dev_prop->maxThreadsPerBlock;
```
jjsjann123, line 2499: we are always using max threads per block here for computing?

jjsjann123 (commented):
minor comments/questions. Haven't really gone through the normalization scheduler changes yet.

csrc/scheduler/normalization_inner_outer.h
liqiangxl: It comes from the previous implementation. In this PR I just gave these magic numbers some names without changing them.
why 256 ? It is because this scheudler has very high register pressure due to extra registers to store intermediate outer reduction results. So prefer to allow 255 registers per thread and then the max threads per block is 256. (64K/256=256). May be revised in the heuristic revision PR.

csrc/scheduler/reduction_heuristic.h
liqiangxl: yes, clangtidy

csrc/scheduler/utils.cpp
liqiangxl: No. when this function is being called to do runtime check, we don't know how many threads will be used in the heuristics. So the max is used to be safe. I can put `max_threads_per_block` as a para, if the caller knows the value it can pass the exact value, otherwise, we will use the hardware max.

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -182,11 +182,239 @@ bool InnerOuterPersistentKernelScheduler::canScheduleCompileTime(
...
-std::pair<int64_t, int64_t> getPersistentBufferSize(
+// The roundup is due to the fact that the shared memory buffer is allocated
+// as: ceilDiv(dim_size / vectorize_factor, threads_per_block).
+// Let after_vect = dim_size / vectorize_factor;
+// n_batch = ceilDiv(after_vect, threads_per_block);
+// Then the shared memory buffer size is n_batch * vectorize_factor *
+// threads_per_block * data_type_size. This function returns the maximum
+// possible shared memory buffer size considering all possible block sizes.
+int64_t roundUpSharedMemory(
+    TensorView* tv,
```
jjsjann123, line 193: nitpick: not really using tv here, but rather just `int64_t data_type_size`

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -182,11 +182,239 @@ bool InnerOuterPersistentKernelScheduler::canScheduleCompileTime(
...
+      ref_broadcast_mask.reserve(root.size());
+      for (const auto i : c10::irange(root.size())) {
+        ref_broadcast_mask.push_back(!root.at(i)->isReduction());
+      }
+      break;
+    }
+  }
+  NVF_ERROR(!ref_broadcast_mask.empty(), "ref_broadcast_mask is empty!");
+
+  // find the brodcast tensor whose broadcast makse is same to the reference
```
jjsjann123, line 235:
```suggestion
  // find the broadcast tensor whose broadcast mask is same to the reference
```

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -182,11 +182,239 @@ bool InnerOuterPersistentKernelScheduler::canScheduleCompileTime(
...
+      }
+      auto id_size = runtime_info.expressionEvaluator().evaluate(id->extent());
+      NVF_ERROR(id_size.hasValue(), "Could not infer persistent buffer size.");
+      if (buffer_size == -1) {
+        buffer_size = id_size.as<int64_t>();
+      } else {
+        buffer_size *= id_size.as<int64_t>();
+      }
+    }
+    buffer_size = (buffer_size == -1) ? 0
```
jjsjann123, line 274: qq: what does `buffer_size == 0` mean? A full reduction/broadcast tensor? shouldn't that still give `buffer_size == 1`

jjsjann123: basically I'm saying, why not start with buffer_size = 1. so we can get rid of the if condition in line 268

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -182,11 +182,239 @@ bool InnerOuterPersistentKernelScheduler::canScheduleCompileTime(
...
+// data type size, and the pre-defined schedule structures. Assume each
+// global memory load has latency 100, each shared memory load has latency 10,
+// latency is proportional to the data type size. There is no write from
+// register to shared memory, so this type of latency is not considered. The
+// magic number 100 and 10 are not critical for our current major use case, e.g.
+// layer norm backward, and rms norm backward. The outer broadcast tensor (the
+// weight tensor) is almost always moved to shared memory before other buffers
+// unless the outer dimension is very small, e.g. less than 1 sm count or other
+// buffers are used 10 times more than the outer broadcast tensor.
+std::vector<TensorView*> sortPersistentBuffers(
```
jjsjann123, line 294: nitpick: the function is called `sortPersistentBuffers` and there's an argument of `persistent_buffers`. But it's instead sorting `projectable_buffe_inputs`.

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -199,31 +427,98 @@ std::pair<int64_t, int64_t> getPersistentBufferSize(
...
+      acc_regs_buffer_sizes[i] =
+          acc_regs_buffer_sizes[i - 1] + tv_buffer_size_regs;
+      acc_smem_buffer_sizes[i] =
+          acc_smem_buffer_sizes[i - 1] + tv_buffer_size_smem;
+    }
+
+    // Determine the least number of buffers to transfer to shared memory
+    // to ensure the register buffer size doesn't exceed the available limit.
+    int64_t n_smem_buffer = -1;
+    for (int i = 1; i <= n_buffers; i++) {
```
jjsjann123, line 491: nitpick: this for loop could go inside `if (buffer_params.regs_buffer_size > available_regs) {`

jjsjann123: ^^^ wait.. I probably read this one wrong. The logic here means that we'll *always* give at least 1 buffer to smem. Is that intended? if `buffer_params.regs_buffer_size <= available_regs`, shouldn't all buffer go to register instead?

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -199,31 +427,98 @@ std::pair<int64_t, int64_t> getPersistentBufferSize(
...
+
+  auto total_buffer_size = buffer_params.project_to_input
+      ? persistent_buffer_size_info.projected_persistent_buffer_size
+      : persistent_buffer_size_info.persistent_buffer_size;
+  total_buffer_size +=
+      partialOuterReductionBufferSize(reduction_tvs, runtime_info);
+
+  const auto dev_prop = at::cuda::getCurrentDeviceProperties();
+  int64_t smem_overhead =
+      scheduler_utils::getSharedMemoryOverheadPerBlock(fusion, reduction_tvs);
```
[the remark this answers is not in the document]
liqiangxl: You are right! I didn't update the caller after the function is enhanced.

csrc/scheduler/normalization_inner_outer.cpp
liqiangxl: revised interface.

csrc/scheduler/normalization_inner_outer.cpp
liqiangxl: It was in the original code to calculate buffer size so I just used it without change.
I think the code only accounts for arrays with size larger than 1 as persistent buffer. For example, almost all the broadcast Tvs can't inline with their consumers, but they are not counted as persistent buffer. It avoids unfriendly values, e.g. when we have a persistent buffer of size 1024, then if plus another 1 for broadcast, it becomes 1025. The definition of persistent buffer is on us and it is only used to estimate register/smem usage, so I belive a slight change should be fine.
I'll investigate this part later and check whether my guess is correct or we can also start with `buffer_size = 1` if nothing bad happens.

csrc/scheduler/normalization_inner_outer.cpp
liqiangxl: I think you may read it wrong. This part is nested inside `if (buffer_params.regs_buffer_size > available_regs) {`, so if condition is false, all buffers go to register.

csrc/scheduler/normalization_inner_outer.cpp
liqiangxl: I agree, it is a little bit confusing. The reason is because the current innerOuter scheduler always project to inputs. I renamed this function and added another one to process cases we won't project to inputs (when there is no out broadcast tv, see the newly added test case)

tests/cpp/test_combined_inner_outer_reduction.cpp
```diff
@@ -1061,4 +1061,43 @@ TEST_F(NVFuserTest, CombinedSchedulerInnerOuterMismatch) {
   test({0});
 }
 
+// innerOuter scheduler projects buffer to inputs when there is one or more
+// outer broadcast tvs, e.g. in layer norm backward and RMS norm backward.
+// This test covers the branch where the outer broadcast tensor is not exist
+// and data type is fp32, so the buffer is not projected to inputs.
```
jjsjann123, line 1067: note to myself. go back and revisit the buffer project to input thing.

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -182,11 +182,251 @@ bool InnerOuterPersistentKernelScheduler::canScheduleCompileTime(
...
+    bool is_reused = std::any_of(
+        outer_broadcast_tvs.begin(),
+        outer_broadcast_tvs.end(),
+        [&input_buffer](TensorView* tv) {
+          return DependencyCheck::isDependencyOf(input_buffer, tv);
+        });
+    // for each load from gmem to smem, latency += 100
+    latency += 100 * (is_reused ? 1 : n_outer_loop);
+    // for each load from smem to regs, latency += 10
+    latency += 10 * n_outer_loop * n_proj_buffer_uses.at(idx);
```
jjsjann123, line 373: This look a bit different from what we have in the doc.

here it is really
((is_outer_broadcast_tvs? 1 : S) * 100 + S * N * 10) * dtypeSize

Should the n_outer_loop in line 373 be replaced?

liqiangxl: The code is correct, I will revise the doc.
The code structure is:
```
Tb in smem = load from gmem
for i = 0 , n_outer_loop:
    load Tb to register
```
The outer broadcast tv only needs to do one load from gmem to smem, but it still needs to be loaded from gmem to register when it is being used in the loop.

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -182,11 +182,251 @@ bool InnerOuterPersistentKernelScheduler::canScheduleCompileTime(
...
+// is likely the first candidate to be moved to shared memory when the register
+// space runs low.
+std::vector<TensorView*> getOuterBroadcastTvs(
+    Fusion* fusion,
+    const std::vector<TensorView*>& reduction_tvs) {
+  // set reference broadcast mask using the first inner reduction tv
+  std::vector<bool> ref_broadcast_mask;
+  for (auto tv : reduction_tvs) {
+    if (scheduler_utils::isFastestDimReduction(tv)) {
+      const auto& root = tv->getMaybeAllocationDomain();
```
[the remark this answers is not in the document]
liqiangxl: changed to `getMaybeRFactorDomain`

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -199,31 +439,111 @@ std::pair<int64_t, int64_t> getPersistentBufferSize(
...
-            persistent_buffer_size_info.projected_persistent_buffer_size);
+  // Project to inputs when there is at least one outer broadcast tensor or
+  // projected persistent buffer size is smaller. When projecting to inputs, the
+  // outer broadcast tensor is reused in the loop over the iteration dimension,
+  // test shows it is faster than the non-projected version which requires
+  // reload from gmem for each iteration.
+  // Note: in current use cases (layer norm bwd and RMS norm bwd), there are
+  // outer broadcast tvs and always project to inputs.
+  const auto& outer_broadcast_tvs = getOuterBroadcastTvs(fusion, reduction_tvs);
+  bool can_project = ir_utils::getViewOps(fusion).empty() &&
```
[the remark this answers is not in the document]
liqiangxl: Let me write an issue to track it and investigate later. It has been there for a long time, all inner persistent, outer persistent schedulers don't project to inputs if view op is detected.
I also changed the check to use `normalization_scheduler_utils::isProjectBufferToInputs`

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -199,31 +439,111 @@ std::pair<int64_t, int64_t> getPersistentBufferSize(
...
+      persistent_buffer_size_info.projected_persistent_buffer_size > 0;
+  buffer_params.project_to_input = can_project &&
+      (!outer_broadcast_tvs.empty() ||
+       persistent_buffer_size_info.projected_persistent_buffer_size <
```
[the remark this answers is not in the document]
liqiangxl: it is calculated in `scheduler_utils::persistentBufferSize` based on the life scope of the persistent buffers.
It can be smaller than `persistent_buffer_size` and `0` indicates no buffer can be projected back to inputs.

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -199,31 +439,111 @@ std::pair<int64_t, int64_t> getPersistentBufferSize(
...
+      partialOuterReductionBufferSize(reduction_tvs, runtime_info);
+
+  const auto dev_prop = at::cuda::getCurrentDeviceProperties();
+  int64_t smem_overhead = scheduler_utils::getSharedMemoryOverheadPerBlock(
+      fusion,
+      reduction_tvs,
+      InnerOuterPersistentKernelScheduler::threads_per_block_max);
+  int64_t available_smem =
+      (int64_t)dev_prop->sharedMemPerMultiprocessor - smem_overhead;
+  int64_t available_regs = scheduler_utils::register_file_size_full;
```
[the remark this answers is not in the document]
liqiangxl: yes. we are using too many to store the buffers.

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -199,31 +439,111 @@ std::pair<int64_t, int64_t> getPersistentBufferSize(
...
+              fusion,
+              persistent_buffer_info.persistent_buffers,
+              persistent_buffer_info.projectable_buffer_inputs,
+              outer_broadcast_tvs,
+              reduction_tvs,
+              runtime_info)
+        : sortPersistentBuffers(persistent_buffer_info.persistent_buffers);
+    // calculate the accumulated buffer size of the first N buffers
+    const int64_t n_buffers = (int64_t)sorted_candidate_tvs.size();
+    std::vector<int64_t> acc_regs_buffer_sizes(n_buffers + 1, 0);
```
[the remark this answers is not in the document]
liqiangxl: revised these two from vectors to scalars.

csrc/scheduler/normalization_utils.cpp
```diff
@@ -853,29 +855,36 @@ bool isProjectBufferToInputs(
...
   if (!can_project) {
     return false;
   }
 
-  // don't project if can't reduce buffer size
-  if (persistent_buffer_size_info.projected_persistent_buffer_size >=
-      persistent_buffer_size_info.persistent_buffer_size) {
+  // don't project if can't reduce buffer size unless it's innerOuter with
+  // outer broadcast where project to inputs reduces gmem access.
+  if (!is_inner_outer_with_outer_bcast &&
```
jjsjann123, line 869: nitpick: rename `is_inner_outer_with_outer_bcast` to `check_projected_buffer_size` or something like that, and put the comment `unless it's innerOuter with outer broadcast where project to inputs reduces gmem access.` to the place where it's called.

The util here is shared across normalization scheduler and I *think* that's a more readable argument name.

jjsjann123 (commented):
LGTM.

Not stamping yet, since we discussed offline that we'd want to verify the (lack of) performance gain on larger sizes. Currently we are suspecting it's a combination of higher latency from shared memory and register spill. But we don't yet know to what extend each of those are contributing, or if there's some other issue lurking underneath.

liqiangxl:
> LGTM.
> 
> Not stamping yet, since we discussed offline that we'd want to verify the (lack of) performance gain on larger sizes. Currently we are suspecting it's a combination of higher latency from shared memory and register spill. But we don't yet know to what extend each of those are contributing, or if there's some other issue lurking underneath.

It's updated in the doc, the best I can get after revising the heuristics to minimize register spills is the white line. [1], [2], [3] indicates locations we are moving 1/2/3 buffers to shared memory.
image

jjsjann123:
> > LGTM.
> > Not stamping yet, since we discussed offline that we'd want to verify the (lack of) performance gain on larger sizes. Currently we are suspecting it's a combination of higher latency from shared memory and register spill. But we don't yet know to what extend each of those are contributing, or if there's some other issue lurking underneath.
> 
> It's updated in the doc, the best I can get after revising the heuristics to minimize register spills is the white line. [1], [2], [3] indicates locations we are moving 1/2/3 buffers to shared memory.

Thanks for updating. Sorry I wasn't paying enough attention to the doc update.
Looking at the graph it's just register spill that's punishing the perf when we have 3 buffer in shared memory. But as the performance is still better than main branch where we are simply not using persistent memory, looks like the decision is going to be a bit blurry as to when do we opt-in persistent buffer.

jjsjann123 (commented):
Not sure if this is one of the PR that we'd want to shift to our new workflow.
But at least it looked good in my naive eyes. Pulling in @naoyam for a sanity check

tests/cpp/test_combined_inner_outer_reduction.cpp
```diff
@@ -1066,4 +1066,43 @@ TEST_F(NVFuserTest, CombinedSchedulerInnerOuterMismatch) {
...
+
+  auto heuristic =
+      getInnerOuterPersistentHeuristics(fusion_ptr.get(), aten_inputs);
+  NVF_CHECK(heuristic, "InnerOuterPersistentHeuristics was not generated!");
+  NVF_CHECK(
+      !heuristic->project_persistent_buffers,
+      "Shouldn't project persistent buffers to inputs!");
+
+  FusionExecutorCache fec(std::move(fusion_ptr));
+  auto cg_outputs = fec.runFusionWithInputs(aten_inputs);
```
naoyam, line 1100: Does this guarantee the inner-outer scheduler is indeed used through `FusionExecutorCache`? IIRC, the `getHeuristics` interface does not do the compile time check, does it?

csrc/scheduler/utils.h
```diff
@@ -646,5 +647,17 @@ int64_t getPersistentBufferSizeOfTensor(
     SchedulerRuntimeInfo& runtime_info,
     const PersistentBufferInfo& persistent_buffer_info);
 
+//! Calculates the shared memory overhead per block, including CUDA driver
```
naoyam, line 650: I don't understand what this comment means. Could you please define what the shared memory overhead is?

csrc/scheduler/reduction_heuristic.h
```diff
@@ -133,8 +135,22 @@ class ReductionParams : public HeuristicParams {
   // block_dim_inner_reduction_extra (usually TIDy)
   ParallelType block_dim_inner_reduction_extra = ParallelType::Serial;
 
-  // use shared memory for persistent buffer, if false, will use registers
+  // If false, only register is used to store persistent buffers.
```
naoyam, line 138: Why do we need two properties? Why not just `smem_persistent_tvs` and let it mean tensors in the vector are stored in shared memory?

csrc/scheduler/normalization_utils.h
```diff
@@ -311,6 +311,8 @@ int64_t getMaxRegOrSharedMemorySizeForPersistentBuffer(
 // inputs are cached instead of the persistent buffers. The decision of
 // projection is primarily based on the required sizes of the two cases --
 // projection is done if projecting to the inputs results in a smaller size.
+// If the scheduler is innerOuter with outer broadcast, projection is allowed
+// even it leads to a larger buffer size.
```
naoyam, line 315: Why?

tests/cpp/test_combined_inner_outer_reduction.cpp
liqiangxl: You are right, `getHeuristics` won't do compile/runtime checks. I changed to use `scheduleInnerOuterPersistentKernel + compileFusion + runFusion`, which is more straightfoward showing which scheudler is used.

we can also directly check which scheduler is used by `fec.getMostRecentKernelRuntime()                            ->schedulerHeuristics()`, either way seems fine to me.

csrc/scheduler/reduction_heuristic.h
liqiangxl: > Why do we need two properties?

It's not strictly required. They are there because the current **inner persistent scheduler must move all or none** of the persistent buffers to smem. So there was only one `bool` indicating moving all or nothing, which is more efficient than using a vector of tensors. However, for this innerOuter scheudler, we needs to identify which one is being moved, so `smem_persistent_tvs` is added.

> Why not just smem_persistent_tvs and let it mean tensors in the vector are stored in shared memory?

we can do that. It means we need to change the inner persistent scheduler. I was trying to make this PR focus on the innerOuter persistent scheduler and prefer to leave the changes to other schedulers to other PRs if possible.

Of course, I can also add the changes into this PR if you feel that is more appropriate. :)

Added a TODO comment:
```
  // TODO: For innerPersistentHeuristic, all or none of the persistent buffers
  // should be stored in shared memory. The code can be extended to allow moving
  // some of the persistent buffers to shared memory while others still in
  // registers, similar to what we did in the innerOuterPersistentHeuristic.
```

csrc/scheduler/normalization_utils.h
liqiangxl: It is the logic exist in the current main branch. We need to re-evaluate this logic, for example if there is an extreme case where projecting to inputs increased the buffer size by 10 times, then we probabally don't want to project to inputs.
added comments:
```
// If the scheduler is innerOuter with outer broadcast, projection is allowed
// even it leads to a larger buffer size becuase the scheduled kernel allows the
// reuse of the outer broadcast Tv when iterating over the outer reduction
// dimension and leads to higher performance ( TODO: needs re-evaluate, may not
// true if the buffer size is increased a lot when projecting to inputs). See
// https://github.com/NVIDIA/Fuser/issues/402
```

csrc/scheduler/reduction_heuristic.h
naoyam: > So there was only one bool indicating moving all or nothing, which is more efficient than using a vector of tensors. 
I doubt the efficiency difference is significant, especially since the vector length, i.e., the number of persistent buffers, should be quite small. It'd be more important to make the heuristics as much simple as possible. It should be more preferable to extend an existing parameter rather than adding a new parameter.

csrc/scheduler/normalization_utils.h
naoyam: I'm still confused. If it's the current behavior, why does this function need to change?

jjsjann123: We are changing the logic.

This is the old logic:
> // ... The decision of 
// projection is primarily based on the required sizes of the two cases --
// projection is done if projecting to the inputs results in a smaller size.

The new code where Liqiang added a knob here that allowed projection to happen regardless of that restriction.
> // If the scheduler is innerOuter with outer broadcast, projection is allowed

csrc/scheduler/reduction_heuristic.h
jjsjann123: If I read Liqiang's comment correctly, I think the bigger impact from removing `shared_mem_persistent_buffer` is that vvv
> It means we need to change the inner persistent scheduler.

though I"m not sure how many LOC that means... If it's small let's do it here. Otherwise we can open a separate PR for clean up.

naoyam: Shouldn't it be just replacing `shared_mem_persistent_buffer` with `!smem_persistent_tvs.empty()`?

jjsjann123: Yeah for its uses in this PR it's straightforward.

For inner persistent scheduler, we need to throw something in `smem_persistent_tvs`, and I'm assuming we'd want to put something reasonable there as well. That's where I'm not sure how much change is needed. Maybe it's just a couple lines and in that case we should squeeze it in this PR as well.

liqiangxl: I added a commit deleting `shared_mem_persistent_buffer`. The section of code to move buffer to smem is put into a utility function shared by both inner persistent and innerOuter persistent.

csrc/scheduler/normalization_utils.h
naoyam: Thanks @jjsjann123.

tests/cpp/test_combined_inner_outer_reduction.cpp
naoyam: What did you change?

csrc/scheduler/utils.h
naoyam: Did you add anything?

Still the overall comment is unclear to me. 

> Defaults to using maximum threads per block (threads_per_block = -1),

What does it mean? 

> The callers can pass other values if they are sure about the max values used at the runtime. 

What values are you talking about and why?

csrc/scheduler/normalization_utils.h
naoyam: Please update the comment

csrc/scheduler/normalization_utils.cpp
```diff
@@ -1262,6 +1271,44 @@ bool compileTimeCheck(Fusion* fusion, ScheduleHeuristic schedule_heuristic) {
...
+      bool use_smem = isSharedMemoryPersistent(tv);
+      // If project to inputs, the buffer inputs are stored in
+      // [smem_persistent_buffers], after projection, the cached input becomes
+      // the new persistent buffer.
+      if (!use_smem && rparams.project_persistent_buffers &&
+          std::find(cached_inputs.begin(), cached_inputs.end(), tv) !=
+              cached_inputs.end()) {
+        auto input_tv = ir_utils::producerTvsOf(tv).at(0);
+        use_smem = isSharedMemoryPersistent(input_tv);
+      }
```
naoyam, line 1304: Can you please clarify what's going on here? Why does `isSharedMemoryPersistent` need to be used twice?

liqiangxl: Revised comments:
```
      // Persistent buffers are categorized into two types:
      // (1) Cached input tensors.
      //     For these, [smem_persistent_buffers] holds the original input
      //     tensors, not the cached.
      // (2) Intermediate tensors: Other tensors used throughout computation.

      // If a buffer is absent from [smem_persistent_buffers], it may be a
      // cached input. In such cases, verify if the original input tensor is
      // stored in [smem_persistent_buffers]. So, we may need to call
      // isSharedMemoryPersistent() twice, one for the buffer iteself and the
      // other for the buffer's input tensor if the buffer is a cached input
      // and it is not in [smem_persistent_buffers].
```
Condition`rparams.project_persistent_buffers` is not required to trigger this additional check since input tensors can be persistent buffer even without projecting to inputs.

csrc/scheduler/normalization_utils.h
liqiangxl: updated, looks like some of my local changes are not pushed to remote...

tests/cpp/test_combined_inner_outer_reduction.cpp
liqiangxl: updated, previous local change was not pushed to remote. sorry for the confusion.

csrc/scheduler/utils.h
liqiangxl: this change was also not updated to remote.
Refering to `reduction workspace which depends on the number of threads per block specified by the parameter threads_per_block`
revised:
```
//! The required shared memory size for a block inclues two parts: (1) smem
//! for persistent buffers and (2) overhead. The overhead includes space
//! reserved by the CUDA driver and reduction workspace which depends on the
//! number of threads per block specified by the parameter threads_per_block.
//! By default, the function uses the maximum allowed number of threads per
//! block (threads_per_block = -1) to calculate the overhead. The caller can
//! specify a different value if they are sure about the max value used at
//! runtime.
```

csrc/scheduler/normalization_utils.cpp
liqiangxl: Changed, thanks!

liqiangxl:
!build

csrc/scheduler/normalization_utils.cpp
```diff
@@ -1262,6 +1272,53 @@ bool compileTimeCheck(Fusion* fusion, ScheduleHeuristic schedule_heuristic) {
...
 }
 
+void movePersistentBufferToSmem(
+    Fusion* fusion,
+    const ReductionParams& rparams,
+    const std::vector<TensorView*>& cached_inputs) {
+  // Transfer the persistent buffer tensors to shared memory. These tensors are
+  // housed in smem_persistent_buffers. If a candidate tensor is input, move its
+  // associated cached tensors.
+  if (!rparams.smem_persistent_buffers.empty()) {
```
naoyam, line 1282: nit: 
```suggestion
  if (rparams.smem_persistent_buffers.empty()) {
     return;
   }
```

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -182,11 +182,252 @@ bool InnerOuterPersistentKernelScheduler::canScheduleCompileTime(
...
+  }
+  return sorted_candidate_tvs;
+}
+
+// Sort persistent buffers based on the number of uses from smallest to largest.
+// Used when buffers are not projected to inputs.
+std::vector<TensorView*> sortPersistentBuffers(
+    std::vector<TensorView*> persistent_buffers) {
+  // Directly sort the persistent_buffers vector based on the number of uses.
+  std::sort(
```
naoyam, line 399: Should use stable_sort?

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -182,11 +182,252 @@ bool InnerOuterPersistentKernelScheduler::canScheduleCompileTime(
...
+      persistent_buffers.begin(),
+      persistent_buffers.end(),
+      [](TensorView* a, TensorView* b) {
+        return a->uses().size() < b->uses().size();
```
naoyam, line 403: The number of uses only reflects the static count and is not necessarily correlated with the actual use count. I suppose you found this static info is enough for the layernorm case?

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -182,11 +182,252 @@ bool InnerOuterPersistentKernelScheduler::canScheduleCompileTime(
...
+          auto id_size =
+              runtime_info.expressionEvaluator().evaluate(id->extent());
+          NVF_ERROR(id_size.hasValue(), "Could not infer outer dim size.");
+          if (outer_dim_size == -1) {
+            outer_dim_size = id_size.as<int64_t>();
+          } else {
+            outer_dim_size *= id_size.as<int64_t>();
+          }
+        }
+      }
```
naoyam, line 348:
```suggestion
      int64_t outer_dim_size = 1;
      for (auto id : buffer->getMaybeRFactorDomain()) {
        if (id->isReduction()) {
          auto id_size =
              runtime_info.expressionEvaluator().evaluate(id->extent());
          NVF_ERROR(id_size.hasValue(), "Could not infer outer dim size.");
            outer_dim_size *= id_size.as<int64_t>();
        }
      }
```

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -182,11 +182,252 @@ bool InnerOuterPersistentKernelScheduler::canScheduleCompileTime(
...
+      int gdimy = sm_count;
+      int n_loops = (int)ceilDiv(outer_dim_size, gdimy);
+      return n_loops;
+    }
+    NVF_ERROR(false, "Could not infer outer loop count.");
+    return 1;
```
naoyam, line 354: No need to have this line

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -182,11 +182,252 @@ bool InnerOuterPersistentKernelScheduler::canScheduleCompileTime(
...
+  // for n_outer_loop = ceilDiv(outer_dim_size, gdimy) times. gdimy is a
+  // heuristic para, here we assume gdimy = sm_count. For outer broadcast tv,
+  // e.g. weight tensor, it is reused in every outer loop, when calculate the
+  // latency of global memory load, 1 is used instead of n_outer_loop. For other
+  // tensors, each loop is loading a different row, so n_outer_loop is used.
+  std::vector<int> buffer_latency(n_buffer_inputs, 0);
+  int sm_count =
+      (int)at::cuda::getCurrentDeviceProperties()->multiProcessorCount;
+  int n_outer_loop = [&sm_count, &reduction_tvs, &runtime_info]() -> int {
+    for (auto buffer : reduction_tvs) {
```
naoyam, line 332: Are all of the outer reduction tensors in `reduction_tvs` guaranteed to have the same shape? Asking because it seems there can be other tensors that have broadcast domains that correspond to the reduction domains in these `reduction_tvs`.

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -199,31 +440,108 @@ std::pair<int64_t, int64_t> getPersistentBufferSize(
...
+  int64_t available_regs = scheduler_utils::register_file_size_full;
+  buffer_params.smem_overhead = smem_overhead;
+
+  // Put all the persistent tensors in registers
+  buffer_params.regs_buffer_size = total_buffer_size;
+  buffer_params.smem_buffer_size = 0;
+
+  // Relocate buffers to shared memory until the buffer size in registers is
+  // within the allowable limit.
+  if (buffer_params.regs_buffer_size > available_regs) {
```
naoyam, line 482: Code cleanup suggestion: try to minimize conditional nesting. You'd need to remember all the active`if` conditions while reading this function, which is particularly problematic for long functions like this.

```suggestion
  if (buffer_params.regs_buffer_size <= available_regs) {
     ...;
     return;
  }
  
  // Reg+Smem specific code follows
```

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -182,11 +182,252 @@ bool InnerOuterPersistentKernelScheduler::canScheduleCompileTime(
...
+  // is parallelized by BIDy, each thread loops over the outer dimension
+  // for n_outer_loop = ceilDiv(outer_dim_size, gdimy) times. gdimy is a
+  // heuristic para, here we assume gdimy = sm_count. For outer broadcast tv,
+  // e.g. weight tensor, it is reused in every outer loop, when calculate the
+  // latency of global memory load, 1 is used instead of n_outer_loop. For other
+  // tensors, each loop is loading a different row, so n_outer_loop is used.
+  std::vector<int> buffer_latency(n_buffer_inputs, 0);
+  int sm_count =
+      (int)at::cuda::getCurrentDeviceProperties()->multiProcessorCount;
+  int n_outer_loop = [&sm_count, &reduction_tvs, &runtime_info]() -> int {
```
naoyam, line 331: Why does this need to be a lambda?

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -199,31 +440,108 @@ std::pair<int64_t, int64_t> getPersistentBufferSize(
...
+        ? sortProjectableBufferInputs(
+              fusion,
+              persistent_buffer_info.persistent_buffers,
+              persistent_buffer_info.projectable_buffer_inputs,
+              outer_broadcast_tvs,
+              reduction_tvs,
+              runtime_info)
+        : sortPersistentBuffers(persistent_buffer_info.persistent_buffers);
+
+    // determine the number of buffers to move to shared memory
```
naoyam, line 493: Give more details how the sorted tensor list is going to be used to figure out this. Looks like the algorithm is to visit each tensor from the largest to the smallest and try to store as many largest tensors as possible in registers. Is that right?

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -199,31 +440,108 @@ std::pair<int64_t, int64_t> getPersistentBufferSize(
...
+    const int64_t n_buffers = (int64_t)sorted_candidate_tvs.size();
+    int64_t acc_regs_buffer_size = 0, acc_smem_buffer_size = 0;
```
naoyam, line 495: Please comment these variables

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -182,11 +182,252 @@ bool InnerOuterPersistentKernelScheduler::canScheduleCompileTime(
...
+// move one or more buffers to shared memory. Since shared memory
+// is slower than registers, we should move the buffers that are
+// less frequently accessed to shared memory.
+struct PersistentBufferStorageParams {
+  std::vector<TensorView*> smem_persistent_buffers;
+  int64_t smem_buffer_size = -1;
+  int64_t regs_buffer_size = -1;
+  int64_t smem_overhead = -1;
+  bool has_enough_regs_and_smem = false;
+  bool project_to_input = false;
```
naoyam, line 420: Please add comments to each member

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -199,31 +440,108 @@ std::pair<int64_t, int64_t> getPersistentBufferSize(
...
+          vectorize_factor,
+          InnerOuterPersistentKernelScheduler::threads_per_block_min,
+          InnerOuterPersistentKernelScheduler::threads_per_block_max,
+          dev_prop->warpSize);
+
+      acc_regs_buffer_size += tv_buffer_size_regs;
+      acc_smem_buffer_size += tv_buffer_size_smem;
+      // check if we have enough registers after moving the first-i buffers to
+      // shared memory
+      if (buffer_params.regs_buffer_size - acc_regs_buffer_size <=
```
naoyam, line 515: Unclear what's being done here. `regs_buffer_size` is the total buffer size (btw, shouldn't it be renamed to to just `total_buffer_size` as it doesn't necessarily indicate the register usage). `acc_regs_buffer_size` is the sum of the register sizes of all the tensors visited so far. `regs_buffer_size - acc_regs_buffer_size` seems to indicate the necessary shmem size if the first i buffers are moved to shared memory. Am I following the code correctly? If so, what does `<= available_regs` mean?

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -199,31 +440,108 @@ std::pair<int64_t, int64_t> getPersistentBufferSize(
...
+  int64_t available_smem =
+      (int64_t)dev_prop->sharedMemPerMultiprocessor - smem_overhead;
+  int64_t available_regs = scheduler_utils::register_file_size_full;
+  buffer_params.smem_overhead = smem_overhead;
+
+  // Put all the persistent tensors in registers
+  buffer_params.regs_buffer_size = total_buffer_size;
+  buffer_params.smem_buffer_size = 0;
+
+  // Relocate buffers to shared memory until the buffer size in registers is
```
naoyam, line 480: Do you have data on performance impacts by this buffer selection?

csrc/scheduler/normalization_inner_outer.cpp
liqiangxl: make sense. changed

csrc/scheduler/normalization_inner_outer.cpp
liqiangxl: not required. I used it to make a code block for clarity. Changed to regular code block using `{}`

csrc/scheduler/normalization_inner_outer.cpp
liqiangxl: changed to regular code block.

csrc/scheduler/normalization_inner_outer.cpp
liqiangxl: yes, `compile time can schedule` check ensures they have the same shape.

csrc/scheduler/normalization_inner_outer.cpp
liqiangxl: Added comment
```
// TODO: Currently, this function is unused as all existing cases (LayerNorm bwd
// and RMS norm bwd) involve projecting buffers to inputs. Consider this
// function a placeholder for scenarios where buffers are not projected to
// inputs. Refer to the test CombinedSchedulerInnerOuterNoOuterBroadcastTv for a
// concrete example of such cases. Add NoOuterBroadcast cases to our benchmarks
// and test the performance of this function.
```

csrc/scheduler/normalization_inner_outer.cpp
liqiangxl: added

csrc/scheduler/normalization_inner_outer.cpp
liqiangxl: ` acc_regs_buffer_size and acc_smem_buffer_size` are removed.

csrc/scheduler/normalization_inner_outer.cpp
liqiangxl: The visiting order is from small to large. I think maybe the `i-1` index is confusing and it is revised.
Added more comments and revised the code.

Two steps:
```
  // (3.1) Sort the candidate persistent buffers, based on its read/write usage.
  // Put the buffer that is less frequently used in the front.
```
and 
```
  // (3.2) Before this loop, all buffers are in registers.
  // Try to move buffer from register to shared memroy. The visiting order is
  // from the first buffer to the last buffer in the sorted_candidate_tvs.
  // After one buffer is moved to shared memory, the buffer size in registers
  // and shared memory are updated accordingly. Break if required register and
  // shared memory are lower than limit or shared memory exceeds the limit.
```

csrc/scheduler/normalization_inner_outer.cpp
liqiangxl: Here is the diff between using sorted buffer selection and default buffer selection (the outer broadcast tv is not prioritized) for layer norm backward. There is no difference for the 2nd and 3rd buffers.
image

csrc/scheduler/normalization_inner_outer.cpp
naoyam: Sorry, I'm not sure how to interpret your comment. Is this enough or not?

csrc/scheduler/normalization_inner_outer.cpp
naoyam: Can you please add a comment?

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -199,31 +459,129 @@ std::pair<int64_t, int64_t> getPersistentBufferSize(
...
+      (int64_t)dev_prop->sharedMemPerMultiprocessor - smem_overhead;
+  int64_t available_regs = scheduler_utils::register_file_size_full;
+  buffer_params.smem_overhead = smem_overhead;
+
+  // (1) init the buffer_params by putting all the persistent tensors in
+  // registers
+  buffer_params.regs_buffer_size = total_buffer_size;
+  buffer_params.smem_buffer_size = 0;
+
+  // (2) If the avilable register is larger than current register buffer size,
```
naoyam, line 500:
```suggestion
  // (2) If the available register is larger than current register buffer size,
```

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -182,11 +182,271 @@ bool InnerOuterPersistentKernelScheduler::canScheduleCompileTime(
...
+  // Total number of bytes occupied by all persistent buffers stored in
+  // registers.
+  int64_t regs_buffer_size = -1;
+
+  // Additional shared memory usage per block that is not associated with
+  // persistent buffers. This includes memory for driver overhead and workspace
+  // for reductions.
+  int64_t smem_overhead = -1;
+
+  // Flag indicating whether there are sufficient registers and shared memory
```
naoyam, line 433: What happens when this flag is false? Is the scheduler going to be rejected?

csrc/scheduler/normalization_inner_outer.cpp
naoyam: Yeah, when I saw that and also `<= N` instead of usual `< N`, I thought the order was reversed. Sorry for my confusion.

csrc/scheduler/normalization_inner_outer.cpp
naoyam: Now it's much more clear. Thanks!

csrc/scheduler/normalization_inner_outer.cpp
liqiangxl: yes, false means we don't have sufficient registers and shared memory.

csrc/scheduler/normalization_inner_outer.cpp
liqiangxl: It is enough for layernorm case.

liqiangxl:
Checked results if move all buffers from register to smem, just tell us the relative speed between smem & register.
image

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -199,31 +459,129 @@ std::pair<int64_t, int64_t> getPersistentBufferSize(
...
+
+    // calculate the size of this buffer & reduce the register buffer size
+    int64_t tv_buffer_size_regs =
+        scheduler_utils::getPersistentBufferSizeOfTensor(
+            current_tv, runtime_info, persistent_buffer_info);
+    buffer_params.regs_buffer_size -= tv_buffer_size_regs;
+
+    // round up the buffer size to shared memory & increase the shared memory
+    // buffer size
+    int64_t tv_buffer_size_smem = roundUpSharedMemory(
```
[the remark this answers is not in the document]
naoyam: Actually, I guess the register usage is underestimated as it doesn't seem to account for the redundant nature of broadcast allocation.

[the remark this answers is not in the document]
naoyam: Talked with Liqiang offline. This was my misunderstanding. Since the tensor itself doesn't have broadcast domains, there's no concern of underestimation.

csrc/scheduler/normalization_inner_outer.cpp
liqiangxl: added.
```
      // Use the first outer reduction tensor to infer the outer dim size.
      // Compile time check ensures they have the same iter and reduction axes,
      // connected with inner reduction tv. Similar to the first inner reduction
      // tv is used to derive the reduction properties. Here the first outer
      // reduction tv is used to derive the outer dim size.
```

liqiangxl:
@naoyam as suggested, code for buffer sorting is removed. performance is as expected and we have consistent improvement over the main branch.

liqiangxl:
!build --pybench

naoyam (approved):
LGTM. Why did the Python benchmarks fail?

liqiangxl:
> LGTM. Why did the Python benchmarks fail?

(1) validation error, e.g. `test_dropout_layernorm_bwd_nvf_benchmark[dtype=torch.float32-size=[8192_66048]]`
`AssertionError: Max error: 0.0001220703125,                     Absolute tolerance: 0.000136882, Relative tolerance: 1.36882e-06`
(2) out of memory, e.g. `test_matmul_nvf_benchmark[dtype=torch.bfloat16-445896_5752_245784_NN] `
`torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 204.14 GiB. GPU 0 has a total capacity of 39.39 GiB of which 38.87 GiB is free.`

The validation issue is a known issue and tracked at https://github.com/NVIDIA/Fuser/issues/1275
The OOM seems new. @xwang233

xwang233:
> > LGTM. Why did the Python benchmarks fail?
> 
> (1) validation error, e.g. `test_dropout_layernorm_bwd_nvf_benchmark[dtype=torch.float32-size=[8192_66048]]` `AssertionError: Max error: 0.0001220703125, Absolute tolerance: 0.000136882, Relative tolerance: 1.36882e-06` (2) out of memory, e.g. `test_matmul_nvf_benchmark[dtype=torch.bfloat16-445896_5752_245784_NN] ` `torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 204.14 GiB. GPU 0 has a total capacity of 39.39 GiB of which 38.87 GiB is free.`
> 
> The validation issue is a known issue and tracked at #1275 The OOM seems new. @xwang233

cc @Priya2698

PR #1955, opened 2024-03-18
remove temp matmul APIs

protonu:
In preparation for a unified Python matmul API, this PR removes the temporary matmul APIs (and associated tests).


protonu:
!build

protonu:
!build

protonu:
!build

xwang233:
cc @jjsjann123 , there are python API changes, so `version.txt` likely needs to be bumped, that's why `python_bc_advisory` detected failures

jjsjann123:
> Just a note that I am landing this PR with a backward compatibility test failure. This arises from the fact that I am removing temporary Python APIs.

I'm guessing matmul API is fine since there's no user on this. Our protocol is that, if there's update to python API (especially when we are removing things), we would bump `version.txt` so framework tests can condition on that.

PR #1957, opened 2024-03-18
Communication lowering refactor

cowanmeg:
Small refactor of `multidevice/lower_communications.cpp`. Moving arguments of `DeviceMesh` to `TensorView*`.  Current lowerings only need the DeviceMesh, but more complex lowerings will need to analyze both. 

Creating a separate PR for this refactor to make an upcoming PR a little cleaner (https://github.com/NVIDIA/Fuser/pull/1939)

### DIFF
--- a/csrc/multidevice/lower_communication.cpp
+++ b/csrc/multidevice/lower_communication.cpp
@@ -146,7 +146,7 @@ void lowerToScatter(
     at::Tensor output_tensor,
     std::vector<std::shared_ptr<Communication>>& comms) {
   // we arbitrarily choose the first device of the sender mesh to be the root
-  auto receiver_mesh = output_tv->getDeviceMesh();
+  const auto& receiver_mesh = output_tv->getDeviceMesh();
   auto root = input_tv->getDeviceMesh().vector().at(0);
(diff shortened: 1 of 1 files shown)


csrc/multidevice/lower_communication.cpp
```diff
@@ -139,18 +140,19 @@ CommParams createParamsForGatherScatter(
...
-    const DeviceMesh& sender_mesh,
-    const DeviceMesh& receiver_mesh,
+    TensorView* input_tv,
+    TensorView* output_tv,
     at::Tensor input_tensor,
     at::Tensor output_tensor,
     std::vector<std::shared_ptr<Communication>>& comms) {
   // we arbitrarily choose the first device of the sender mesh to be the root
-  auto root = sender_mesh.vector().at(0);
+  auto receiver_mesh = output_tv->getDeviceMesh();
```
samnordmann, line 149: nit: `const auto&` here and everywhere it is a DeviceMesh, so we avoid copying a vector

samnordmann (approved):
Looks good!

For my own understanding, can you just detail a little bit the motivation? But anyway I find this change quite natural, so I approve it!

PR #1885, opened 2024-03-07
ensure has enought register or smem before disable project buffer to inputs

liqiangxl:
**redo**: https://github.com/NVIDIA/Fuser/pull/1796
**added**: ensure has enought register or smem before disable project buffer to inputs
**Performance:** posted in previous [PR](https://private-user-images.githubusercontent.com/116412316/311030075-93279f4c-4bb3-40c5-b8b8-125cecd6c8ca.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MDk5MjgyMDcsIm5iZiI6MTcwOTkyNzkwNywicGF0aCI6Ii8xMTY0MTIzMTYvMzExMDMwMDc1LTkzMjc5ZjRjLTRiYjMtNDBjNS1iOGI4LTEyNWNlY2Q2YzhjYS5wbmc_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjQwMzA4JTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI0MDMwOFQxOTU4MjdaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT03MTcyMTQ3NzlhNTUwNjE3ZWU5NDBjNDQ2MjdiNWZjYjJlYTEwYTE3MGYwNmQ1NTNmODA1ZjY2MTdkZDYxMzFkJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZhY3Rvcl9pZD0wJmtleV9pZD0wJnJlcG9faWQ9MCJ9.QtYUZgprP8w_oE67VkL09rzTYiSaSzdFM1m3FJC5T5Y).

PR #1961, opened 2024-03-19
add comments and check to func projectBufferToInputs

liqiangxl:
(1) Improve clarity by adding comments:
```
//  Evaluates whether to project a persistent buffer to inputs.
//  1. Rejects projection if it doesn't reduce buffer size.
//  2. Requires projection if the unprojected buffer size exceeds
//     available register/shared memory.
//  3. Checks operations between inputs and buffer:
//     - if recompute requires RNG operations, reject projection.
//     - if recompute requires Exp operations, reject if buffer size is lower
//       than a threshold.
//     - allow project for all other cases.
```

(2) error check
```
  NVF_ERROR(
      max_available_buffer >=
          persistent_buffer_size_info.projected_persistent_buffer_size,
      "There is not enough register/shared memory to store the projected buffer!",
      " max_available_space: ",
      max_available_buffer,
      " projected_buffer_size: ",
      persistent_buffer_size_info.projected_persistent_buffer_size);
```


csrc/scheduler/normalization_utils.cpp
```diff
@@ -864,6 +871,14 @@ bool projectBufferToInputs(
   // memory to store the buffers
   int64_t max_available_buffer = getMaxRegOrSharedMemorySizeForPersistentBuffer(
       runtime_info, persistent_buffer_info.persistent_buffers);
+  NVF_ERROR(
```
naoyam, line 874: Actually, when we return `false` here, we should make sure `persistent_buffer_info.persistent_buffers` is smaller than `max_available_buffer`. Do we have that check as well?

csrc/scheduler/normalization_utils.cpp
```diff
@@ -845,8 +845,15 @@ int64_t getMaxRegOrSharedMemorySizeForPersistentBuffer(
   return available_persistent_buffer_size;
 }
 
-// Returns true if the gains of reducing buffer size is larger than the pains of
-// recalculations. We don't know the real answer until we run it.
+//  Evaluates whether to project a persistent buffer to inputs.
```
naoyam, line 848: Most useful comments do not just describe what the corresponding code does but give the important context and reasoning for what the code does. For example, here, it would be more informative to have something like:

```
Returns true if persistent buffers are projected to inputs, meaning the inputs are cached instead of the persistent buffers.

The decision of projection is primarily based on the required sizes of the two cases -- projection is done if projecting to the inputs results in a smaller size. However, we experimentally found that certain relatively expensive operations should not be projected even when that would require a larger buffer size. Specifically,

- rng: should never be projected no matter how much larger the buffer would consume
- exp in inner normalization: only allowed to get projected if the buffer is smaller than a certain size

Otherwise, as long as the projected inputs are smaller than the original persistent buffers, this function returns `true`.



```

liqiangxl: will follow this style

csrc/scheduler/normalization_utils.cpp
liqiangxl: that is guranteed in `canScheduleRunTime`, otherwise will be rejected by the scheduler.

naoyam: Then why did the original PR fail?

liqiangxl: The original PR fails becase when `projected_persistent_buffer_size <= max_available_buffer < persistent_buffer_size`, we should project to inputs to lower the buffer size but the original PR used the larger buffer size due to rng op.

`canScheduleRunTime` ensures `max_available_buffer >= std::min(projected_persistent_buffer_size, persistent_buffer_size)`, when `projected_persistent_buffer_size >= persistent_buffer_size`, we can derive `max_available_buffer >= persistent_buffer_size`

naoyam: > canScheduleRunTime ensures max_available_buffer >= std::min(projected_persistent_buffer_size, persistent_buffer_size)

Why doesn't it do `max_available_buffer >= projectBufferToInputs(...) ?projected_persistent_buffer_size : persistent_buffer_size`?

liqiangxl: I tried to use this function to check whether we can project to inputs in run time check. But got an issue since the segmenter will simply peel off the rng op from the original fusion. Here is the original post, [link](https://github.com/NVIDIA/Fuser/pull/1885#issuecomment-1984964113).

naoyam: >  the segmenter will simply peel off the rng op from the original fusion

What do you mean? Could you give more specific info?

liqiangxl: The segmenter generates a fusion with a single rng op. No input to the fusion.

Error msg: ` Could not launch kernel as program could not infer i0(i0) for the buffer T5_g[ iS8{i0}, iS9{i1} ] `

The Fusion:
```
Inputs:
Outputs:
  T5_g[ iblockIdx.x576{( ceilDiv(( ceilDiv(( ceilDiv(( i0 * i1 ), 8) ), 1) ), 128) )}, iUS575{1}, iV573{8}, ithreadIdx.x577{128} ] ca_pos( 2 ) produce_pos( 2 ), bool

%kernel_math {
T4_l[ iblockIdx.x590{( ceilDiv(( ceilDiv(( ceilDiv(( i0 * i1 ), 8) ), 1) ), 128) )}, iUS589{1}, iS587{8}, ithreadIdx.x591{128} ] ca_pos( 4 )
   = rng_uniform_range({i0, i1}, double(0), double(1), float);
T296_l[ iblockIdx.x583{( ceilDiv(( ceilDiv(( ceilDiv(( i0 * i1 ), 8) ), 1) ), 128) )}, iUS582{1}, iS580{8}, ithreadIdx.x584{128} ] ca_pos( 2 ) produce_pos( 4 )
   = T4_l[ iblockIdx.x590{( ceilDiv(( ceilDiv(( ceilDiv(( i0 * i1 ), 8) ), 1) ), 128) )}, iUS589{1}, iS587{8}, ithreadIdx.x591{128} ] ca_pos( 4 )
   < double(0.80000000000000004);
T5_g[ iblockIdx.x576{( ceilDiv(( ceilDiv(( ceilDiv(( i0 * i1 ), 8) ), 1) ), 128) )}, iUS575{1}, iV573{8}, ithreadIdx.x577{128} ] ca_pos( 2 ) produce_pos( 2 )
   = Set( T296_l[ iblockIdx.x583{( ceilDiv(( ceilDiv(( ceilDiv(( i0 * i1 ), 8) ), 1) ), 128) )}, iUS582{1}, iS580{8}, ithreadIdx.x584{128} ] ca_pos( 2 ) produce_pos( 4 ), cache_op=Streaming )
}
```

do we allow fusion without inputs in the segmenter? If not we may need to fix the segmenter, if allowed we need to figure out how to use other fusion-segments to derive the size of T5.

naoyam: I did read your link, but why is this related? Using `projectBufferToInputs` should not change segmentation, right? If so, we should have seen this issue no matter if `projectBuffertoInputs` was used.

liqiangxl: It actually does. Here is how it works if we use `projectBufferToInputs` in `canScheduleRunTime`:
(1) It detects `rng` op and will set `project_to_inputs = false`, and when `max_available_buffer < non_project_buffer_size`, the fusion segment.
(2) The segmenter start from stratch, and merged all the exprs except the one with `rng` op, up to here, the `canScheduleRunTime` returns true because `project_to_inputs = true` and `max_available_buffer > projected_buffer_size`. However, the segmenter can't merge the `rng` op back with other exprs since it will lead to situation (1).

naoyam: I'm very confused. Are you saying we are making different projection decisions between `canScheduleRuntime` and wherever others that use `projectBufferToInputs`? Isn't it considered inconsistent?

liqiangxl: Previsosly it is inconssitent and that is why the original PR failed in some cases.
The current logic is:
(1) `canScheduleRuntime` assumes we can project to inputs as long as projected buffer size is smaller.
(2)  Then `projectBufferToInputs` did a further check which may disable project (as long as we still have enough space to host the buffer).

liqiangxl: Ideally, we should use `projectBufferToInputs` in `canScheduleRuntime`, but it caused the aforementioned segmentation issue.

naoyam: These two seem inconsistent to me:

>  It detects rng op and will set project_to_inputs = false, and when max_available_buffer < non_project_buffer_size, the fusion segment.

and 

> Then projectBufferToInputs did a further check which may disable project (as long as we still have enough space to host the buffer).

liqiangxl: Your first reference is the old logic in `canScheduleRuntime` and it leads to the aforementioned segmentation issue.
The second reference is the revised logic in `projectBufferToInputs`
They do inconsistgent but they do NOT exist in the same branch.

I retried the revised version of `projectBufferToInputs` and it can be directly used in `canScheduleRuntime` without leading to the segmentation issue. 
The code is updated. 
In this version, project to input or not soley comes from the function `isProjectBufferToInputs` (renamed from `projectBufferToInputs`).
Thanks for the comments!

csrc/scheduler/normalization_utils.cpp
```diff
@@ -846,42 +846,33 @@ int64_t getMaxRegOrSharedMemorySizeForPersistentBuffer(
...
   }
 
   // must project to inputs otherwise don't have enough register or shared
-  // memory to store the buffers
+  // memory to store the buffers. Even after projecting, may still not have
+  // enough register or shared memory, then canScheduleRunTime will return
+  // false.
   int64_t max_available_buffer = getMaxRegOrSharedMemorySizeForPersistentBuffer(
       runtime_info, persistent_buffer_info.persistent_buffers);
-  NVF_ERROR(
```
liqiangxl: this check is removed. since it is also used in canScheduleRuntime check.

csrc/scheduler/normalization_utils.cpp
```diff
@@ -1011,19 +1002,16 @@ PersistentKernelProperties getPersistentKernelProperties(
   // which requires more advanced analysis is batch norm backwards.
   // TODO: Fix projected persistent buffers with view
   // https://github.com/csarofeen/pytorch/issues/2054
-  // If projected persistent buffers are smaller, they will be used.
-  bool can_project = ir_utils::getViewOps(fusion).empty() &&
```
liqiangxl, line 997: view op check is merged into the function.

csrc/scheduler/normalization_utils.cpp
liqiangxl: The segmentation issue is gone because previously I used `bool has_rng_op = ir_utils::hasOpsOfType<RNGOp>(fusion);` to check rng op. I should only check the ops from inputs to the buffer as what was done in the `projectBufferToInputs`

naoyam: > Your first reference is the old logic in canScheduleRuntime and it leads to the aforementioned segmentation issue.

What do you mean? That is just from your comment above: https://github.com/NVIDIA/Fuser/pull/1961#discussion_r1532523189. It isn't about the old logic. It should be what would happen if `projecBufferToInputs` is used.

liqiangxl: Sorry for the confusion.  Let me do a summary of the four different approaches. Same colored blocks indicating they are calling the same version of `projecBufferToInputs` or other functions.

image

naoyam: Thanks for the clarification. 

Since this does involve a change of logic, please run the CI tests. 

Are there any existing tests/benchmarks that result in different segmentation due to this change?

liqiangxl:
!build --diff --diff-bench

liqiangxl:
Failed tests:
[jit_binary_tests_17_V100] IndexingOpTest.TorchGatherAllRankAllSelectedDim_CUDA\n\nunknown file\nC++ exception with description \"CUDA out of memory. Tried to allocate 1.59 GiB. GPU

liqiangxl:
Hi @xwang233 do you know why codegen diff CI shows the following err mgs?
```
ls: cannot access 'codegen_comparison/codediff_*.html': No such file or directory
+ python /workspace/pyt_nightly_ci_utils/webpage_deploy_wrapper.py
  File "/workspace/pyt_nightly_ci_utils/webpage_deploy_wrapper.py", line 8, in <module>
    url = deploy_html_to_internal_pages(pathlib.Path(sys.argv[1]).resolve().absolute().__str__())
IndexError: list index out of range
```

xwang233:
I think this is the actual error message:
```python
00:03:42 Traceback (most recent call last):
00:03:42   File "/opt/pytorch/nvfuser/tools/codediff/diff_report.py", line 924, in <module>
00:03:42     TestRun(args.dir1),
00:03:42   File "<string>", line 5, in __init__
00:03:42   File "/opt/pytorch/nvfuser/tools/codediff/diff_report.py", line 520, in __post_init__
00:03:42     self.compute_kernel_map()
00:03:42   File "/opt/pytorch/nvfuser/tools/codediff/diff_report.py", line 539, in compute_kernel_map
00:03:42     parser = LogParserPyTest(logfile)
00:03:42   File "/opt/pytorch/nvfuser/tools/codediff/diff_report.py", line 253, in __init__
00:03:42     self.parse(log_file)
00:03:42   File "/opt/pytorch/nvfuser/tools/codediff/diff_report.py", line 273, in parse
00:03:42     self.parse_line(line)
00:03:42   File "/opt/pytorch/nvfuser/tools/codediff/diff_report.py", line 426, in parse_line
00:03:42     self.finalize_test(True)
00:03:42   File "/opt/pytorch/nvfuser/tools/codediff/diff_report.py", line 287, in finalize_test
00:03:42     assert self.current_test is not None
00:03:42 AssertionError
```

cc @jacobhinkle

xwang233:
It's likely because the test info line is too verbose?

```python
Running 4 items in this shard: tests/python/test_schedule_ops.py::TestScheduleOps::test_merge_op, tests/python/test_schedule_ops.py::TestScheduleOps::test_reduction_factor_op, tests/python/test_schedule_ops.py::TestScheduleOps::test_reorder_op, tests/python/test_schedule_ops.py::TestScheduleOps::test_split_op

[2024-03-27 18:38:15] tests/python/test_schedule_ops.py::TestScheduleOps::test_merge_op [W327 18:38:16.260557159 fusion_cache.cpp:507] Warning: You are overwriting the current user schedule for a definition! (function createUserSchedule)
[W327 18:38:16.262093127 fusion_cache.cpp:507] Warning: You are overwriting the current user schedule for a definition! (function createUserSchedule)
[W327 18:38:16.263710179 fusion_cache.cpp:507] Warning: You are overwriting the current user schedule for a definition! (function createUserSchedule)
[W327 18:38:16.265022630 fusion_cache.cpp:507] Warning: You are overwriting the current user schedule for a definition! (function createUserSchedule)
PRINTING: __tmp_kernel_1.cu
ptxas info    : Overriding maximum register limit 256 for '_ZN56_GLOBAL__N__00000000_17___tmp_kernel_1_cu_930d841f_374419nvfuser_1ENS_6TensorIfLi3ELi3EEENS0_IfLi2ELi2EEE' with  255 of maxrregcount option
ptxas info    : Compiling entry function '_ZN56_GLOBAL__N__00000000_17___tmp_kernel_1_cu_930d841f_374419nvfuser_1ENS_6TensorIfLi3ELi3EEENS0_IfLi2ELi2EEE' for 'sm_80'
ptxas info    : Function properties for _ZN56_GLOBAL__N__00000000_17___tmp_kernel_1_cu_930d841f_374419nvfuser_1ENS_6TensorIfLi3ELi3EEENS0_IfLi2ELi2EEE
ptxas info    : Used 18 registers, 4 bytes smem, 448 bytes cmem[0]

PRINTING: __tmp_kernel_1.ptx
PRINTING: __tmp_kernel_2.cu
ptxas info    : Overriding maximum register limit 256 for '_ZN56_GLOBAL__N__00000000_17___tmp_kernel_2_cu_fd819f5e_374419nvfuser_2ENS_6TensorIfLi3ELi3EEENS0_IfLi2ELi2EEE' with  255 of maxrregcount option
ptxas info    : Compiling entry function '_ZN56_GLOBAL__N__00000000_17___tmp_kernel_2_cu_fd819f5e_374419nvfuser_2ENS_6TensorIfLi3ELi3EEENS0_IfLi2ELi2EEE' for 'sm_80'
ptxas info    : Function properties for _ZN56_GLOBAL__N__00000000_17___tmp_kernel_2_cu_fd819f5e_374419nvfuser_2ENS_6TensorIfLi3ELi3EEENS0_IfLi2ELi2EEE
ptxas info    : Used 17 registers, 408 bytes cmem[0]

PRINTING: __tmp_kernel_2.ptx
[W327 18:38:18.635457484 fusion_cache.cpp:507] Warning: You are overwriting the current user schedule for a definition! (function createUserSchedule)
PRINTING: __tmp_kernel_3.cu
ptxas info    : Overriding maximum register limit 256 for '_ZN56_GLOBAL__N__00000000_17___tmp_kernel_3_cu_edd60f52_374419nvfuser_3ENS_6TensorIfLi3ELi3EEENS0_IfLi2ELi2EEE' with  255 of maxrregcount option
ptxas info    : Compiling entry function '_ZN56_GLOBAL__N__00000000_17___tmp_kernel_3_cu_edd60f52_374419nvfuser_3ENS_6TensorIfLi3ELi3EEENS0_IfLi2ELi2EEE' for 'sm_80'
ptxas info    : Function properties for _ZN56_GLOBAL__N__00000000_17___tmp_kernel_3_cu_edd60f52_374419nvfuser_3ENS_6TensorIfLi3ELi3EEENS0_IfLi2ELi2EEE
ptxas info    : Used 18 registers, 4 bytes smem, 448 bytes cmem[0]

PRINTING: __tmp_kernel_3.ptx
PASSED
```

jacobhinkle:
Ah yes, `[2024-03-27 18:38:15] tests/python/test_schedule_ops.py::TestScheduleOps::test_merge_op [W327 18:38:16.260557159 fusion_cache.cpp:507] Warning: You are overwriting...`. This is probably screwing up the regex that detects a new test. I am not sure what to do about that other than try to disable that warning inside codediff.

jacobhinkle:
> Ah yes, `[2024-03-27 18:38:15] tests/python/test_schedule_ops.py::TestScheduleOps::test_merge_op [W327 18:38:16.260557159 fusion_cache.cpp:507] Warning: You are overwriting...`. This is probably screwing up the regex that detects a new test. I am not sure what to do about that other than try to disable that warning inside codediff. I guess a simpler workaround would be to just disable these `TestScheduleOps` tests in codediff. I wonder why we are just now seeing this...

liqiangxl:
Thanks @xwang233 and @jacobhinkle. Let me check if I can fix the regex check in the ci script.

csrc/scheduler/normalization_utils.cpp
liqiangxl: No change in the existing benchmarks.
I am also not aware of tests, but CI has some tech issues to run, will double check after fixed.

liqiangxl:
!build --diff

liqiangxl:
CI is fixed and no generated code change in test and benckmark cases. @naoyam

PR #1962, opened 2024-03-19
Indexing of TMA

zasdfgbnm:
This PR adds the indexing of TMA, please start the review from note `[Schedule and indexing of TMA]` to get an idea on how the schedule looks like and what is the algorithm for indexing.

With this PR, we are no longer limited to loading the entire tensor as a tile. We can arbitrarily view the tensor and create tiles, and arbitrarily schedule the tile and non-tile IterDomains.

Other limitations still apply, for example, we are still limited to only one thread per block. And to make this PR small and incremental, I decided to leave some important tasks for the future:
- Validation, test of invalid cases
- Investigation of some bugs


zasdfgbnm:
!build

zasdfgbnm:
!build

zasdfgbnm:
!build

csrc/index_compute.cpp
```diff
@@ -3291,81 +3291,572 @@ Val* Index::eye(
...
+// elements to to be transfered. If this is the case, then the originating bulk
+// IterDomains must be the inner of the split that defines it, and the outer of
+// the split must be non-bulk. For this case, the extents of the originating
+// bulk IterDomains are the tile sizes (the `boxDim` parameter in
+// `cuTensorMapEncodeTiled`).
+
+// However, it is also possible to load/store tiles in a strided manner, that
+// is, for example, when we have a tile size (8, 8), we want every other element
+// to be transfered, that is, the items at (0, 0), (0, 2), (0, 4), ..., (2, 0),
+// (2, 2), (2, 4), ..., (6, 6). In this senario, the `elementStrides` in
```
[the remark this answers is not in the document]
zasdfgbnm: This is indeed `(6, 6)`. I think it is mostly a matter of the definition of "tile size". Here, tile size 8 does not mean you have in total 8 items, but instead, you have a span of 8, and you select odd numbers that is >=0 and < 8.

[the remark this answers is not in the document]
zasdfgbnm: Is "tile size" confusing? Should we think of another name like "box dim" (just like what the driver API is using)?

csrc/index_compute.cpp
```diff
@@ -3291,81 +3291,572 @@ Val* Index::eye(
...
+//
+// Definition 2: A bulk IterDomain is "originating" if it has no parent or its
+// parents are not "bulk".
+//
+// In the above diagram, I4 and I7 are originating bulk IterDomains.
+//
+// The number of originating bulk IterDomains is the dimensionality of the
+// tensor in the eyes of TMA.
+//
+// Note that an originating bulk IterDomains must be the output of a split from
```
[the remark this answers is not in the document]
zasdfgbnm: You are right. It's wrong to say `if it has no parent`.  This was from my previous design.

csrc/index_compute.cpp
```diff
@@ -3291,81 +3291,572 @@ Val* Index::eye(
...
+  // IterDomain is inner and which is outer. The actual order must be determined
+  // by propagating from the allocation domain.
+  VectorOfUniqueEntries<IterDomain*> originating_bulk_ids;
+  for (auto id : pending) {
+    if (id == nullptr) {
+      continue;
+    }
+    originating_bulk_ids.pushBack(id);
+    auto def = id->definition();
+    NVF_ERROR(
```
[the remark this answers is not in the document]
zasdfgbnm: changed

csrc/index_compute.cpp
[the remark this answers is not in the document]
zasdfgbnm: changed

tests/cpp/test_memory.cpp
```diff
@@ -255,21 +256,34 @@ class TMALdstTest : public TMATest,
...
     dim = std::get<2>(GetParam());
 
+    // TODO: test less-nice shapes, for example 128 + 1 instead of 128
+
+    // TODO: I was originally trying large shapes, but unfortunately I see
+    // silent wrong results in the SimpleStore test. Only store tests have such
+    // problem, load tests works fine. I have no idea why, but I guess it is a
+    // synchronization issue instead of an indexing issue, so I decide to cheat
+    // by using small shapes for now, and leave the investigation to future
+    // work.
```
zasdfgbnm, line 266: I can confirm that, if I add a
```C++
asm volatile("nanosleep.u32 10000;");
```
before the `Hopper::cpAsyncBulkTensorTileS2G`, the test will pass.

zasdfgbnm: Adding a `asm volatile("fence.proxy.async.shared::cta;");` also resolve the problem.

zasdfgbnm: fixed by https://github.com/NVIDIA/Fuser/pull/1967

zasdfgbnm:
!build

csrc/index_compute.cpp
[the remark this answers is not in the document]
zasdfgbnm: Sure, this is certainly a better description. What do you mean by
> for two outer-most ranks in the structure.

isn't there 2 ranks in total, so it should be like below?

> for two ranks in the structure.

csrc/index_compute.cpp
```diff
@@ -3291,81 +3291,571 @@ Val* Index::eye(
...
+    // This replay has to be consistent with compute at index map.
+    BestEffortReplay replay_producer_as_consumer(
+        producer_tv->getLeafDomain(),
+        consumer_tv->getLeafDomain(),
+        c2p_root_map);
+
+    const auto& c2p_map = replay_producer_as_consumer.getReplay();
+
+    consumer_to_gmem = [=](IterDomain* id) -> IterDomain* {
+      return c2p_map.at(id);
```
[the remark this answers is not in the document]
zasdfgbnm: Good question. I have no idea how C++'s `[=]` works with reference. Looks like it automatically dereference:

```C++
#include <iostream>
#include <functional>
#include <vector>

class A {
public:
    A() {
        std::cout << "ctor " << this << std::endl;
    }
    A(const A& a) {
        std::cout << "copy ctor " << &a << " -> " << this << std::endl;
    }
    ~A() {
        std::cout << "dtor " << this << std::endl;
    }
    void print() const {
        std::cout << "A::print " << this << std::endl;
    }
};

int main() {
    std::function<void()> f;
    std::cout << "enter scope" << std::endl;
    {
        std::vector<A> av(1);
        const A& a = av[0];
        f = [=]() {
            a.print();
        };
    }
    std::cout << "leave scope" << std::endl;
    f();
}
```

```
enter scope
ctor 0x986ec0
copy ctor 0x986ec0 -> 0x7ffc00b40667
copy ctor 0x7ffc00b40667 -> 0x986ee0
dtor 0x7ffc00b40667
dtor 0x986ec0
leave scope
A::print 0x986ee0
dtor 0x986ee0
```

csrc/index_compute.cpp
```diff
@@ -3291,81 +3291,571 @@ Val* Index::eye(
...
+//
+// The number of originating bulk IterDomains is the dimensionality of the
+// tensor in the eyes of TMA.
+//
+// Note that an originating bulk IterDomains must be the output of a split from
+// another IterDomain. Most commonly, we load/store tiles "contiguously", that
+// is, for example, when we have a tile size (8, 8), we want all the 8*8=64
+// elements to to be transfered. If this is the case, then the originating bulk
+// IterDomains must be the inner of the split that defines it, and the outer of
+// the split must be non-bulk. For this case, the extents of the originating
```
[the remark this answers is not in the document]
zasdfgbnm: I added a few paragraphs before `Definition 1` explaining how the schedule looks like https://github.com/NVIDIA/Fuser/pull/1962/commits/7718921da6b430dfd94d2b0003ed35586388290c. Hope that will help answer you question. For your question, if the TensorView is 1D, but you want to use a 2D TMA, then you will need to first view it as 2D then create tile. So you will have something like:

```
  I1 -> split -> I2 -> split -> I4 (Bulk)
            |               \-> I5 (non-Bulk)
             \-> I3 -> split -> I6 (Bulk)
                            \-> I7 (non-Bulk)
```

csrc/index_compute.cpp
```diff
@@ -3291,81 +3291,571 @@ Val* Index::eye(
...
+    bool all_outputs_are_bulk = true;
+    for (auto out : def->outputs()) {
+      if (bulk_ids.count(out->as<IterDomain>()) == 0) {
+        all_outputs_are_bulk = false;
+        break;
+      }
+    }
 
+    if (all_outputs_are_bulk) {
+      // All outputs of def are bulk IterDomains, so all inputs of def are bulk
```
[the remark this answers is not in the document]
zasdfgbnm: > inputs will be bulk due to the fact that we are replaying transformations from consumer to producer? 

No, this has nothing to do with replaying producer as consumer. We replay producer as consumer because the consumer dictate the loop structure of an expression. For example, if we have `T1[I1*I2] = T0[I1, I2]`, we will have loop structure

```python
for i in I1*I2:
  T1[i] = T0[i / size2, i % size2]
```

instead of

```python
for i1 in I1:
  for i2 in I2:
    T1[i1 * size2 + i2] = T0[i1, i2]
```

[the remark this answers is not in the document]
zasdfgbnm: I think the confusion here is the concept "bulk IterDomain" vs "IterDomain parallelized as `ParallelType::Bulk`", they are different concepts. An "IterDomain parallelized as `ParallelType::Bulk`" is always a "bulk IterDomain", but not the reverse. We only parallelize leaf domains, so in your first diagram, it is an error if you parallelize `I2` as `ParallelType::Bulk`. It is a "bulk IterDomain" but not an "IterDomain parallelized as `ParallelType::Bulk`".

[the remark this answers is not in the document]
zasdfgbnm: I added some texts explaining these two concepts: https://github.com/NVIDIA/Fuser/pull/1962/commits/bfa3ddbcca8d24587b38b3427d0a1076c35d256a

csrc/index_compute.cpp
```diff
@@ -3291,81 +3291,571 @@ Val* Index::eye(
...
+  // IterDomain. Similarily, the order of the `global_ids` has no meaning.
+  // We are using a std::vector<Val*> just to make the algorithm deterministic.
+
+  std::vector<Val*> global_ids;
+  std::unordered_map<IterDomain*, IterDomain*> global_id_to_tile_id;
+  std::unordered_map<IterDomain*, IterDomain*> global_id_to_orig_bulk_id;
+  std::unordered_map<IterDomain*, IterDomain*> global_id_to_inner_id;
+  for (auto id : originating_bulk_ids) {
+    auto def = dynamic_cast<Split*>(id->definition());
+    NVF_ERROR(
```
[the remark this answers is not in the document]
zasdfgbnm: What if we have an invalid fusion like this?
```
// [I1, I2]
tv->swizzle(XOR, 0, 1);
// [I1', I2']
tv->axis(1)->parallelize(ParallelType::Bulk);
// [I1', I2' (Bulk) ]
```
Then `I2'` is an "originating bulk IterDomain", but this schedule is invalid.

csrc/index_compute.cpp
```diff
@@ -3291,81 +3291,571 @@ Val* Index::eye(
...
+        //     in,
+        //     " must be divisible by ",
+        //     split->factor());
+      }
+      frontier.insert(
+          in_it,
+          std::make_tuple(
+              split->outer(),
+              true,
+              SimplifyingIrBuilder::mulExpr(
```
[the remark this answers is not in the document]
zasdfgbnm: > with multiple `split` Exprs the stride of outer most entry in `frontier` may be several split factors larger than allocation domain stride.

What is the problem of this? For example, if we have an 1D tensor with shape `(102400,)`, and stride `(1,)`, and we split the tensor as `(100, 1024)`, shouldn't have stride `(1024, 1)`?

csrc/index_compute.cpp
```diff
@@ -3291,81 +3291,598 @@ Val* Index::eye(
...
+// the user. For example, it is possible that there is a 1D tensor of shape
+// 1000000 and the user wants to view it as (1000, 1000) and load a tile of (8,
+// 8) from it. Another example is, the user has a tensor of shape (1000, 1000),
+// and as long as the tensor is contiguous, the user can view it as a flattened
+// shape 1000000 tensor, and load a tile of 8 from it. The dimensionality of the
+// tensor in the eyes of TMA is determined by the schedule of the tensor.
+// However, the dimensionality of the tensor in the eyes of TMA can not be
+// smaller than the physical dimensionality of the tensor. For example, if the
+// user has a tensor of shape (128, 128, 128, 128, 128, 128), and the tensor is
+// fully discontiguous, then it is not possible to view it as a <5D tensor, thus
```
naoyam, line 3308: nit: `<= 5D`

csrc/index_compute.cpp
```diff
@@ -3291,81 +3291,598 @@ Val* Index::eye(
...
+//                         |    |       |    |
+//                        ...   |      ... split
+//                              |          /   |
+//                              |         I7   I8
+//                              |         |    |
+//                              |         |   ...
+//                              |         |
+//                         IterDomain Expressions
+//                          |    |    |    |    |
+//                         Bulk Bulk Bulk Bulk Bulk
```
naoyam, line 3336: Just a quick question. This part appears to indicate `I4` and `I7` are used together to generate the five "Bulk" domains. Is that correct? Or, are `I4` and `I7` transformed independently?

zasdfgbnm: They can be independent or together, it makes no difference.

csrc/index_compute.cpp
zasdfgbnm: fixed

csrc/index_compute.cpp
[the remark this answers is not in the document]
drzejan2: Thanks for checking this! I should have done it by myself, sorry for that.

It's good to know what is the behaviour, especially in cases where copy ctor might be expensive.

csrc/index_compute.cpp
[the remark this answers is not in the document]
drzejan2: Thanks for the example of scheduling 1D TV into 2d tiled bulk memory copy one. Now, that I see it, it obvious that it should be done this way.

The changes in the description look also good. Thank you!

csrc/index_compute.cpp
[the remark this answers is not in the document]
drzejan2: > What do you mean by
> > for two outer-most ranks in the structure.
>
> isn't there 2 ranks in total, so it should be like below?
> > for two ranks in the structure.

You are right, my bad, I wrote this part of comment at early stage of review, before I reread the documentation.


To be double check, if I understood everything correctly, based on the description you have added, the relationship between IterDomains looks like this:
```
I1 -> split -> I2 (non-bulk) -> split -> I4 (non-bulk) -> split -> I8 (non-bulk)
          |                         |                          \-> I9 (bulk)
          |                         \--> I5 (non-bulk)
          |
          \--> I3 (non-bulk) -> split -> I6 (non-bulk) -> split -> I10 (non-bulk)
                                    |                          \-> I11 (bulk)
                                    \--> I7 (non-bulk)
```

Here the originating bulk IterDomains: `I9` and `I11`
  - both originating bulk IterDomains are outer products of split
  - their sibling Ids are non-bulk
Parents of originating IterDomains: `I4` and `I6`
  - both parents are iner products of split
  - parents' sibling IDs are non-bulk

Tile size: extends of `I9` and `I11`
Box size: extends of `I4` and `I6`


If that's correct, then this thread can be resolved, thanks!

csrc/index_compute.cpp
```diff
@@ -3291,81 +3291,598 @@ Val* Index::eye(
...
+//
+// Also note that global IterDomains can not have dependencies between each
+// other, that is, no one global IterDomains can be an ancestor or descendant of
+// another global IterDomains. That is, schedules like this is not allowed:
+//   I1 -> split -> Bulk (inner)
+//              \-> I2 (outer) -> split -> Bulk (inner)
+//                                     \-> I3 (outer)
+// Also note that the set of all global IterDomains must be "equivalent" to the
+// allocation domain of the tensor (ignoring unrelated IterTypes like broadcast
+// and reduction), as defined in `ir_utils::validateDomainEquivalence`.
```
drzejan2, line 3444: A minor suggestion, given the number of new terms introduced in this comment, I would find it really useful to see a set of examples with asii-art diagrams followed by explanation which IterDomains are originating bulk, box, global - and what will be passed to `cuTensorMapEncodeTiled`.

We could have examples for:
- 1D tensor - tiled to 2d and scheduled as 2d dense TMA transaction (box == tile)
- 1D tensor - tiled to 2d and scheduled as 2d sparse TMA transaction (box != tile)
- 2D tensor - tiled, but with merge expres included, scheduled as 2d dense TMA transaction

@zasdfgbnm , what do you think about this idea? Something similar to what I included in my [comment](https://github.com/NVIDIA/Fuser/pull/1962#discussion_r1533581104)

csrc/index_compute.cpp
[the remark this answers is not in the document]
drzejan2: Thank you @zasdfgbnm for explanation, especially for adding explanation about the difference between "bulk IterDomain" and "IterDomain parallelized as Bulk".

In the first diagram, I though this operation is legal, and the aim of such operation was to set parallelization type once, and have it automatically propagate to ancestor IterDomains, created when we apply merge/split.

csrc/index_compute.cpp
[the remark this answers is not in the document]
zasdfgbnm: Yes, your understanding is correct.

csrc/index_compute.cpp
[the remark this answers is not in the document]
drzejan2: In this invalid fusion, won't the definition of `I2'` be `Swizzle2D` expression, and due to this definition, `I2'` will be rejected when we populate `originating_bulk_ids`, several lines above?

This is not a blocker, I was surprised see the same / similar check - feel free to resolve this thread :)

csrc/index_compute.cpp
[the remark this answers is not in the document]
drzejan2: Ok, now I understand my mistake. The inner most dimension, if tensor is dense, will have stride of `1` - for some reason I assumed it is the extend of the inner most dimension.

A fundamental mistake - sorry for the noise!

csrc/index_compute.cpp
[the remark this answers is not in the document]
zasdfgbnm: Oh, I see what you mean. Yes, you are right. This check is redundant.

csrc/index_compute.cpp
zasdfgbnm: Good suggestion! I will add examples.

tests/cpp/test_memory.cpp
```diff
@@ -255,21 +256,27 @@ class TMALdstTest : public TMATest,
     dtype = std::get<1>(GetParam());
     dim = std::get<2>(GetParam());
 
+    // TODO: test less-nice shapes, for example 128 + 1 instead of 128
```
zasdfgbnm, line 259: I guess this should be already working. But I still want to leave this for the future to check. I am trying to keep the mental effort for this PR limited to the design and implementation of the schedule.

csrc/index_compute.cpp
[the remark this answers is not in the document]
zasdfgbnm: removed

tests/cpp/test_memory.cpp
```diff
@@ -434,12 +437,201 @@ INSTANTIATE_TEST_SUITE_P(
...
+}
+
+TEST_F(TMAIndexingTest, Advanced) {
+  Fusion fusion;
+  FusionGuard fg(&fusion);
+
+  auto tv0 = TensorViewBuilder()
+                 .ndims(8)
+                 .dtype(DataType::Float)
+                 .contiguity({true, true, false, true, true, false, true, true})
```
drzejan2, line 557: Just a though:

In the test it looks as if IterDomains' contiguity doesn't play any role, but I'm not sure how we could check if finalized TMA structures correctly handled strides and indices.

drzejan2 (approved):
I checked all changes, indexing seems to be reasonable, given my limited experience with that part of the system.

@zasdfgbnm Thank you for preparing this MR.

zasdfgbnm:
I will wait for another approval from @naoyam. No hurry. I am not blocked on this.

zasdfgbnm:
Converting this to draft to avoid accidental merge. But this is ready for review!

tests/cpp/test_memory.cpp
zasdfgbnm: If the result is correct, isn't it sufficient to convince us that the strides and indices are correctly handled?

csrc/index_compute.cpp
```diff
@@ -3291,81 +3291,593 @@ Val* Index::eye(
...
+// diagram, TMA view the tensor as a 2D tensor [I1, I2]. Note that the
+// IterDomain expressions between the allocation domain and [I1, I2] must be
+// compatible with the allocation domain, for example, we can not merge
+// discontiguous IterDomains, and we can not have indivisible splits either.
+//
+// Second, we need to tile each dimension of TMA's view of the tensor. If we
+// want a contiguous box dim (to be described below), then we do one split to
+// get the box. If we want a strided box dim, then we do two splits, the first
+// split is to get the box, and the second split is to get the stride. In the
+// above diagram, I1 has a contiguous box dim, and I2 has a strided box dim.
```
naoyam, line 3350: Where's the definition of the box dim and the strided box dim?

zasdfgbnm: It is in Definition 3: "boxId"

zasdfgbnm: Updated the comment to refer to Definition 3

csrc/index_compute.cpp
```diff
@@ -3291,81 +3291,594 @@ Val* Index::eye(
...
+// consumer, must be scheduled like this:
+//
+//                         ---------------------
+//                         | Allocation Domain |
+//                         ---------------------
+//                           | | | | | | | | |
+//                         IterDomain Expressions
+//                             | | | | | | |
+//                         ---------------------
+//                         |  I1           I2  |
```
naoyam, line 3321: Should we define this part of the domains, i.e., `{I1, I2}`, as the TMA domain? Then, we could define the box domains as the inner split output domains of the TMA domain.

IIUC, the global domains defined below is the same as the TMA domain, right?

tests/cpp/test_memory.cpp
zasdfgbnm: I eventually decided to do it now. And as expected, it is working.

csrc/index_compute.cpp
zasdfgbnm: > IIUC, the global domains defined below is the same as the TMA domain, right?

Yes

> Should we define this part of the domains, i.e., `{I1, I2}`, as the TMA domain? Then, we could define the box domains as the inner split output domains of the TMA domain.

I do think we should give `{I1, I2}` a name, and TMA domain do sound like a good name. But conceptually, is it possible to **define** the box domain as a split of TMA domain? I think we can only **understand** the box domain as a split of TMA domain, but not **define** it this way. We need to define "global ID" first, because otherwise, how would you describe what IDs are included in TMA domain?

tests/cpp/test_memory.cpp
drzejan2: Ok, that's good enough proof that tested functionality works. I was just thinking about additional set of checks, but again, I don't have now a suggestion how it can be done.

I will mark this thread as resolved.

csrc/index_compute.cpp
```diff
@@ -3291,81 +3291,594 @@ Val* Index::eye(
...
-      break;
+      return s->inX()->extent()->evaluate().as<int64_t>();
     }
   }
+  return 1;
+}
+
+} // namespace
+
+std::pair<Val*, Val*> Index::getCpAsyncBulkGmemIndex(
```
naoyam, line 3464: Could you give here an overview of the approach taken in this function? Would be helpful before diving into the function.

csrc/index_compute.cpp
naoyam: I just meant to say we name the split domain of a TMA domain as the box domain. Yes, we would first need to identify what domains should be included in the TMA domain.

csrc/index_compute.cpp
zasdfgbnm: Added in https://github.com/NVIDIA/Fuser/pull/1962/commits/62dd8b9fb72872b21336b4db128892c7fb6dae1e

csrc/index_compute.cpp
zasdfgbnm: updated in https://github.com/NVIDIA/Fuser/pull/1962/commits/62dd8b9fb72872b21336b4db128892c7fb6dae1e

zasdfgbnm:
!build

naoyam (approved):
LGTM

PR #1963, opened 2024-03-19
Guard matmul test on devices with too little smem

jacobhinkle:
The test in question failed on 3090Ti, which is Ampere, due to insufficient memory. In the future we could add this macro call to other tests if needed.


jacobhinkle:
!build

csrc/utils.h
```diff
@@ -544,4 +544,6 @@ V getOrDefault(const std::unordered_map<K, V>& map, const K& key) {
   return i == map.end() ? V() : i->second;
 }
 
+size_t deviceAvailableSharedMemory();
```
wujingyue, line 547:
```suggestion
// In bytes.
size_t deviceAvailableSharedMemory();
```

tests/cpp/test_gpu_tensorcore.cpp
```diff
@@ -2582,6 +2596,10 @@ TEST_F(NVFuserTest, FusionAmpereMatmulSmemEpiloguePromotionRequiredA100_CUDA) {
     GTEST_SKIP()
         << "Test conducted without utilizing shared memory epilogue due to the device's constrained shared memory capacity.";
   }
+  if (!params.promote_prologue_smem_reuse) {
+    GTEST_SKIP()
+        << "Test conducted with shared memory epilogue but without promoting prologue smem re-use.";
+  }
```
wujingyue, line 2602: Should they be moved before `scheduleMatmul`, or am I missing some magic?

tests/cpp/test_gpu_tensorcore.cpp
```diff
@@ -2549,6 +2561,8 @@ TEST_F(NVFuserTest, FusionAmpereMatmulSmemEpiloguePromotionRequiredA100_CUDA) {
           /*ignore_occupancy_drop=*/false);
   scheduleMatmul(&fusion, params);
 
+  SKIP_IF_INSUFFICIENT_SMEM(params, data_types);
```
wujingyue, line 2564: Same question: why is this not before `scheduleMatmul`?

wujingyue (approved):

tests/cpp/test_gpu_tensorcore.cpp
jacobhinkle: This lets us detect a failure in case one occurs before this point. However, we still mark it as skipped since it is not checking what we are supposed to be checking in this case. Placing it before `scheduleMatmul` would speed things up but might reduce coverage slightly.

wujingyue: I got the "coverage" part -- makes sense. But these checks are at the **end** of the test. So either they don't matter for this test and should be removed or the test should be made stronger to catch real errors caused by `!promote_prologue_smem_reuse`. Correct?

tests/cpp/test_gpu_tensorcore.cpp
```diff
@@ -2472,6 +2484,8 @@ TEST_F(
...
```
wujingyue, line 2487:
```suggestion
  // FusionExecutor::compileFusion would fail otherwise.
  SKIP_IF_INSUFFICIENT_SMEM(params, data_types);
```

tests/cpp/test_gpu_tensorcore.cpp
jacobhinkle: Really this test just exists to exercise the path where we know we'll need to promote prologue smem reuse. That is, if we do not set `promote_prologue_smem_reuse`, we can expect an OOM smem error on A100. This skip actually is meant to skip on devices where it's _not_ needed.

jacobhinkle: Yeah, maybe `GTEST_SKIP` is not the right mechanism. I wanted to give some feedback that this test does not actually test what is intended, even if it succeeds. That would happen on devices other than A100. On all these devices we expect the fusion to succeed, so it's probably fine to just remove this skip and let it succeed since the test name already indicates that it's specifically for A100.

wujingyue: I see -- I think using GTEST_SKIP to give that feedback is reasonable. In addition, you may want to consider adding an ASSERT like
```
if (A100) {
  ASSERT_TRUE(params.promote_prologue_smem_reuse) << "...";
}
```
to make sure the promote_prologue_smem_reuse path is indeed triggered.

tests/cpp/test_gpu_tensorcore.cpp
```diff
@@ -2482,6 +2482,19 @@ TEST_F(
...
           data_types,
           /*ignore_occupancy_drop=*/false);
+
+  // Find GPU name to check if this is an A100
+  int dev_idx = 0;
+  NVFUSER_CUDA_RT_SAFE_CALL(cudaGetDevice(&dev_idx));
+  cudaDeviceProp prop;
+  NVFUSER_CUDA_RT_SAFE_CALL(cudaGetDeviceProperties(&prop, dev_idx));
+  std::string gpu_name(prop.name);
+  if (gpu_name.find("A100") != gpu_name.npos) {
```
wujingyue, line 2492: Are prop.major and prop.minor more reliable than names? 

https://docs.nvidia.com/cuda/cuda-runtime-api/structcudaDeviceProp.html#structcudaDeviceProp_164490976c8e07e028a8f1ce1f5cd42d6

jacobhinkle: You're right. I was thinking I needed to be more specific, but every 8.0 device supports the same max smem per cta.

jacobhinkle: Pushed change that uses `deviceMajorMinorCheck` instead of this.

PR #1965, opened 2024-03-19
Guard test_issue1953 for Ampere+

jacobhinkle:
We could alternatively use Half to cover Volta as well, but since the original issue uses BFloat16, I decided to keep it intact and just skip the test on Volta.


jacobhinkle:
!build

zasdfgbnm (approved):

rdspring1 (approved):
LGTM.

jacobhinkle:
Test failure is known and unrelated. Python tests passing on V100.

PR #1969, opened 2024-03-20
ensure threadds spawned inherit options from spawing thread

protonu:


jacobhinkle (commented):
Looks pretty good to me. You should add some tests where you enable and disable options globally (using top-level guards in the test), then you compile in parallel with a `FusionExecutorCache` and verify that the options were preserved. For debug dumps, you can do this silently within a test by creating a `DebugStreamGuard` and checking that the output is written to it. Of course, that might also mean handling `DebugStreamGuard` the way you do with the options guards here.

protonu:
Closing this PR since we aren't going to be using it.

PR #1825, opened 2024-02-23
Fix a segmentation bug with input forwarding. 

wujingyue:
This PR takes code from @naoyam's #1559 and restructures it. 

Before this PR, nvFuser resolves forwarded inputs by unconditionally pulling in their dependencies into the group. This was problematic (#1541) because a group may require a resize op to only appear as an input. This PR fixes #1541 by checking the dependencies are indeed mergeable before pulling them in.

PR #1971, opened 2024-03-20
Avoid duplication in consumer merge in resolveNonscalarForwardedInput

### LINKED ISSUE #1970: segmenter tries to merge not-connected groups.
Errors encountered 
```
  File "/opt/pytorch/nvfuser/nvfuser/__init__.py", line 146, in execute
    result = self._execute(
RuntimeError: areDirectlyConnected(group1, group2) INTERNAL ASSERT FAILED at "/opt/pytorch/nvfuser/csrc/fusion_segmenter.cpp":3621, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. only support testing immediate producer-consumer groups
Exception raised from codeGenSupportedMerge at /opt/pytorch/nvfuser/csrc/fusion_segmenter.cpp:3621 (most recent call first):
```

These are some quick debug prints. Looks like we are trying to merge two connected group, but the second group dropped its input (I think this is coming from input forwarding). So looks like a real bug.

```
(gdb) print detailGroupPrint(std::cout, group1)
g{
inputs: 
T0_g[ iS0{i0}, iS1{i1}, iS2{i2} ] __half
outputs:
 
 
T3_g[ iS5{i0}, iS6{i1}, iS7{i2} ]
   = __half2float(T0_g[ iS0{i0}, iS1{i1}, iS2{i2} ]);
(0)
}
 
$6 = void
(gdb) print detailGroupPrint(std::cout, group2)
g{(pointwise)
inputs: 
outputs: 
T15_g[ iS27{i0}, iS28{i1}, iS29{i2} ] __half
 
 
T14_g[ iS24{i0}, iS25{i1}, iS26{i2} ]
   = T3_g[ iS5{i0}, iS6{i1}, iS7{i2} ]
   + T3_g[ iS5{i0}, iS6{i1}, iS7{i2} ];
(20)
T15_g[ iS27{i0}, iS28{i1}, iS29{i2} ]
   = __float2half(T14_g[ iS24{i0}, iS25{i1}, iS26{i2} ]);
(21)
}
 
$7 = void
```

repro script vvv
```
import torch

    T0 = fd.define_tensor(shape=[-1, -1, -1], contiguity=[True, True, True], dtype=DataType.Half, is_cpu=False, stride_order=[2, 1, 0])
    T1 = fd.define_tensor(shape=[-1], contiguity=[True], dtype=DataType.Half, is_cpu=False, stride_order=[0])
    T2 = fd.define_tensor(shape=[-1], contiguity=[True], dtype=DataType.Half, is_cpu=False, stride_order=[0])
    T3 = fd.ops.cast(T0, dtype=DataType.Float)
    T4, T5 = fd.ops.var_mean(T3, dims=[0, 2], correction=0, keepdim=False)
    T6 = fd.ops.cast(T1, dtype=DataType.Float)
    T7 = fd.ops.add(T6, T5)
    T8 = fd.ops.cast(T7, dtype=DataType.Half)
    T9 = fd.ops.cast(T2, dtype=DataType.Float)
    T10 = fd.ops.add(T9, T4)
    T11 = fd.ops.cast(T10, dtype=DataType.Half)
    fd.add_output(T8, T1)
    fd.add_output(T11, T2)
    T12 = fd.ops.add(T3, T3)
    T13 = fd.ops.cast(T12, dtype=DataType.Half)
    fd.add_output(T13)

    nvfuser_fusion_id0(fd)

inputs = [
    torch.randn((36,), dtype=torch.float16, device='cuda:0').as_strided((3, 4, 3), (12, 3, 1)),
    torch.randn((4,), dtype=torch.float16, device='cuda:0').as_strided((4,), (1,)),
    torch.randn((4,), dtype=torch.float16, device='cuda:0').as_strided((4,), (1,)),
]
fd.execute(inputs)
```

jjsjann123:
fixes #1970


jjsjann123:
Oops. cpp test doesn't look right...

wujingyue:
> input forwarding removes connection between the groups

When does input forwarding do that? If this can indeed happen, I'm surprised existing tests like https://github.com/NVIDIA/Fuser/blob/c7db538a096097de77cedcab34ea97da235c3b58/tests/cpp/test_segmentation.cpp#L294 didn't capture this bug.

jacobhinkle:
So in the included test the groups were only connected through `tv3` which is a cast of the input `tv_bias`. The cast is a `UnaryOp` so it gets "forwarded", i.e. replayed in each segment so that instead of `tv3` as an input, each segment has `tv_bias` as an input.

Probably an ideal solution would perform forwarding during segmentation instead of just before finalization.  That would match how we handle half-precision casting of intermediates. I'm unsure of the complexity of such an approach without diving in further. cc @wujingyue and @naoyam as they've also been looking at uop forwarding recently.

jjsjann123:
> > input forwarding removes connection between the groups
> 
> When does input forwarding do that? If this can indeed happen, I'm surprised existing tests like
> 
> https://github.com/NVIDIA/Fuser/blob/c7db538a096097de77cedcab34ea97da235c3b58/tests/cpp/test_segmentation.cpp#L294
> 
> didn't capture this bug.

Thanks for catching this one! You are absolutely right about this one.

Went back and double checked the code, the issue actually shows up here when we have duplicated edges https://github.com/NVIDIA/Fuser/blob/f483d9190a8b45a70a673f5f3a284ebdc10808df/csrc/fusion_segmenter.cpp#L4246-L4251

I'll patch that instead.

Meanwhile, I'm guessing that you'd prefer explicit error and assert. I'll revert the relaxed check in this PR.

tests/cpp/test_segmentation.cpp
```diff
@@ -578,4 +578,33 @@ TEST_F(SegmentationTest, SliceSegmentCasts) {
...
+  auto tv_bias = makeSymbolicTensor(1, DataType::Half);
+  fusion->addInput(tv_bias);
+
+  auto* tv1 = castOp(DataType::Float, tv0);
+  auto* tv2 = sum(tv1, {0, 2});
+  auto* tv3 = castOp(DataType::Float, tv_bias);
+  auto* tv4 = add(tv2, tv3);
+  auto* tv5 = castOp(DataType::Half, tv4);
+  fusion->addOutput(tv5);
+  auto* tv6 = add(tv3, tv3);
```
jjsjann123, line 596: The trick here is that the segment has two uses of forwarded inputs.

csrc/fusion_segmenter.cpp
```diff
@@ -4243,10 +4243,11 @@ void SegmentCandidateFinder::resolveNonscalarForwardedInput(
   SegmentedGroup* aux_group = input2group_.at(forwarded_input);
   NVF_ERROR(aux_group->producer_edges.empty());
 
-  std::vector<SegmentedGroup*> consumers;
+  // use unordered_set to avoid duplicated group in consumers.
+  std::unordered_set<SegmentedGroup*> consumers;
   consumers.reserve(aux_group->consumer_edges.size());
   for (SegmentedEdge* edge : aux_group->consumer_edges) {
-    consumers.push_back(edge->to);
+    consumers.insert(edge->to);
```
jjsjann123, line 4250: duplicated entry in `consumer` would make use call `codeGenSupportedMerge(input_group, consumer)` twice. Where the second time the connection has already been severed by `mergeNodes()`.

jjsjann123:
!build

csrc/fusion_segmenter.cpp
```diff
@@ -4243,10 +4243,11 @@ void SegmentCandidateFinder::resolveNonscalarForwardedInput(
   SegmentedGroup* aux_group = input2group_.at(forwarded_input);
   NVF_ERROR(aux_group->producer_edges.empty());
 
-  std::vector<SegmentedGroup*> consumers;
+  // use unordered_set to avoid duplicated group in consumers.
+  std::unordered_set<SegmentedGroup*> consumers;
```
wujingyue, line 4247: How about VectorOfUniqueEntries instead? I'm worried that iterating a unordered_set is non-deterministic, which can lead to a serde problem like #1825. cc @rdspring1

csrc/fusion_segmenter.cpp
wujingyue: Thanks for getting to the bottom of this! Can you write down this explanation in code comment instead of PR for posterity?

tests/cpp/test_segmentation.cpp
wujingyue: Thanks! Can the test be simplified? Maybe a chain of unary followed by `add(x,x)` with a segment_set at the end? The segment_set is to skip the initial attempt to schedule the whole fusion unsegmented, a trick used by other tests in this file.

wujingyue (approved):
LGTM with comments! Please fix the title of the PR. It currently says why but not what.

csrc/fusion_segmenter.cpp
jjsjann123: :bow:

jjsjann123:
!build

PR #1973, opened 2024-03-20
Add `GpuLower::validations()`

zasdfgbnm:
Now `KernelSummary::validations` is just a copy of `GpuLower::validations()`. The validations for non-divisible split is now added by `NonDivisibleSplitInfo`, instead of in `KernelIrScanner`. By refactoring this way, we can add other validations during lowering.


zasdfgbnm:
!build

naoyam (approved):
LGTM

PR #1978, opened 2024-03-21
Prevent matmul benchmarks from using ATen eval

jacobhinkle:
These were broken by #1927, but the fix is simple.


jacobhinkle:
!build --diff-bench

protonu:
LGTM!

protonu (approved):

Issue #1829, opened 2024-02-24
Reusable zeroed memory


jacobhinkle:
We currently need zeroed global memory buffers for cross-cta communication. Our current executor calls `at::zeros` to initialize this before each launch of our nvfuser kernel, adding a handful of microseconds. Instead, each executor, kernel runtime, or maybe each process could hold one zeroed buffer (per device) and reuse it without the memset. This would require us to always clean up our semaphores after each use like we do for persistent kernels.

jacobhinkle:
For example, we see the following trace
```
$ nsys nvprof --print-gpu-trace build/nvfuser_bench --benchmark_filter=NvFuserScheduler_Matmul_Manual/nvfuser_splitk_TN/M:1024/N:2048/K:4096/warps:4/stages:3/splitk_factor:2/manual_time
 Start (ns)  Duration (ns)  CorrId  GrdX   GrdY  GrdZ  BlkX  BlkY  BlkZ  Reg/Trd  StcSMem (MB)  DymSMem (MB)  Bytes (MB)  Throughput (MB/s)  SrcMemKd  DstMemKd           Device            Ctx  GreenCtx  Strm                                                  Name
 ----------  -------------  ------  -----  ----  ----  ----  ----  ----  -------  ------------  ------------  ----------  -----------------  --------  --------  -------------------------  ---  --------  ----  ----------------------------------------------------------------------------------------------------
...
 7216730945           2208  410409      1     1     1   128     1     1       16         0.000         0.000                                                     NVIDIA A100 80GB PCIe (0)    1               7  void at::native::vectorized_elementwise_kernel<(int)4, at::native::FillFunctor<long>, at::detail::A…
 7216756801         147583  410422      8    16     2    32     2     2      240         0.000         0.049                                                     NVIDIA A100 80GB PCIe (0)    1               7  <unnamed>::nvfuser_none_f0_c0_r0_g0(<unnamed>::Tensor<<unnamed>::__half, (int)2, (int)2>, <unnamed>…
...
```
The first kernel is `at::zeros()`. Including the latency between these two launches, this introduces 7216756801 - 7216730945 = 25856 ns, which is 17.5% of the runtime of the actual kernel (147.6 us). This is a particularly small problem and this is where the latency will have the biggest impact. It's worth noticing that cuBLAS reuses zeroed workspace memory, so their single-kernel split-K traces include a single kernel:
```
[...]

PR #1984, opened 2024-03-22
Reusable zeroed memory

jacobhinkle:
This introduces a thread-local global memory allocator for each device and uses it whenever there is an intermediate tensor needed which requires zero-initialization.

To enable use `NVFUSER_ENABLE=reuse_zeroed_memory`. You can monitor the allocator using `NVFUSER_DUMP=global_zeroed_memory`.

Before we enable this feature by default, we need to have high confidence that every kernel using zero-initialized memory will always clean up their semaphores. This is currently only the case for serial grid reductions, as far as I know.

This enables the basic functionality of #1829. However, it does not modify existing algorithms to clean up their memory. See `NVFUSER_ENABLE=reuse_zeroed_memory NVFUSER_DUMP=global_zeroed_memory build/nvfuser_tests --gtest_filter=SerialGridReductionTest.Scheduling`, which succeeds when using serial grid reduction, but fails (in debug mode) when using `gridReduce` (note that this test is updated to behave differently in this PR):
```
# NVFUSER_ENABLE=reuse_zeroed_memory NVFUSER_DUMP=global_zeroed_memory build/nvfuser_tests --gtest_filter=SerialGridReductionTest.Scheduling                                                       
Running main() from /opt/pytorch/nvfuser/third_party/googletest/googletest/src/gtest_main.cc
Note: Google Test filter = SerialGridReductionTest.Scheduling
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from SerialGridReductionTest
[ RUN      ] SerialGridReductionTest.Scheduling
[global zeroed memory] Resizing arena to 512 bytes
[global zeroed memory] Allocating byte range: 0 to 512 bytes
[global zeroed memory] Resetting allocated bytes to 0
[global zeroed memory] Allocating byte range: 0 to 512 bytes
[global zeroed memory] Resetting allocated bytes to 0
[global zeroed memory] Resizing arena to 16384 bytes
[global zeroed memory] Allocating byte range: 0 to 16384 bytes
[global zeroed memory] Resetting allocated bytes to 0
[global zeroed memory] Allocating byte range: 0 to 16384 bytes
unknown file: Failure
C++ exception with description "nnz.equal(0) INTERNAL ASSERT FAILED at "/opt/pytorch/nvfuser/csrc/global_allocator.cpp":88, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. Global memory arena was not properly zeroed. Found 2048 bytes that are not zero
Exception raised from checkZeroed at /opt/pytorch/nvfuser/csrc/global_allocator.cpp:88 (most recent call first):
frame #0: <unknown function> + 0x2fde9e (0x556cdb95de9e in build/nvfuser_tests)
frame #1: <unknown function> + 0x2fe0df (0x556cdb95e0df in build/nvfuser_tests)
frame #2: <unknown function> + 0x3f3720 (0x556cdba53720 in build/nvfuser_tests)
frame #3: <unknown function> + 0x3f33df (0x556cdba533df in build/nvfuser_tests)
frame #4: <unknown function> + 0x3f38ed (0x556cdba538ed in build/nvfuser_tests)
frame #5: <unknown function> + 0x315e67 (0x556cdb975e67 in build/nvfuser_tests)
frame #6: <unknown function> + 0x7c5780 (0x556cdbe25780 in build/nvfuser_tests)
frame #7: <unknown function> + 0x7c5877 (0x556cdbe25877 in build/nvfuser_tests)
frame #8: <unknown function> + 0x138f8cc (0x556cdc9ef8cc in build/nvfuser_tests)
frame #9: <unknown function> + 0x1457f0b (0x556cdcab7f0b in build/nvfuser_tests)
frame #10: <unknown function> + 0x14519fd (0x556cdcab19fd in build/nvfuser_tests)
frame #11: <unknown function> + 0x142de24 (0x556cdca8de24 in build/nvfuser_tests)
frame #12: <unknown function> + 0x142e93f (0x556cdca8e93f in build/nvfuser_tests)
frame #13: <unknown function> + 0x142f345 (0x556cdca8f345 in build/nvfuser_tests)
frame #14: <unknown function> + 0x143f86c (0x556cdca9f86c in build/nvfuser_tests)
frame #15: <unknown function> + 0x1458e98 (0x556cdcab8e98 in build/nvfuser_tests)
frame #16: <unknown function> + 0x1452ac7 (0x556cdcab2ac7 in build/nvfuser_tests)
frame #17: <unknown function> + 0x143de6d (0x556cdca9de6d in build/nvfuser_tests)
frame #18: <unknown function> + 0x1407ca0 (0x556cdca67ca0 in build/nvfuser_tests)
frame #19: <unknown function> + 0x1407c19 (0x556cdca67c19 in build/nvfuser_tests)
frame #20: <unknown function> + 0x29d90 (0x7f616c7d4d90 in /usr/lib/x86_64-linux-gnu/libc.so.6)
frame #21: __libc_start_main + 0x80 (0x7f616c7d4e40 in /usr/lib/x86_64-linux-gnu/libc.so.6)
frame #22: <unknown function> + 0x11e9d5 (0x556cdb77e9d5 in build/nvfuser_tests)
" thrown in the test body.

To reproduce: NVFUSER_TEST_RANDOM_SEED=1711120799 NVFUSER_TEST_ATEN_RANDOM_SEED=0 nvfuser_tests --gtest_filter='SerialGridReductionTest.Scheduling'
[  FAILED  ] SerialGridReductionTest.Scheduling (5669 ms)
[----------] 1 test from SerialGridReductionTest (5669 ms total)
```
This test runs with serial grid reduction, then with `gridReduce`. Each time it runs two grid reductions. Both serial grid reductions succeed because the semaphore buffer is properly zeroed. The `gridReduce` succeeds the first time since the memory pool calls `at::zeros` again to request a larger buffer size (`gridReduce` requires more semaphores since there is one per thread segment vs one for each each block segment). However, the second call to `gridReduce` fails because it has not cleaned up its semaphores. Hacking that function to force `PERSISTENT=1` would clean up the semaphores resulting in success in this case. I'm leaving those kind of modifications for a follow-up.

### DIFF
--- a/csrc/executor.cpp
+++ b/csrc/executor.cpp
@@ -14,6 +14,7 @@
 #include <driver_api.h>
 #include <executor_kernel_arg.h>
 #include <executor_utils.h>
+#include <global_allocator.h>
 #include <instrumentation.h>
 #include <ir/all_nodes.h>
 #include <ir/utils.h>
@@ -1809,9 +1810,14 @@ std::vector<at::Tensor> FusionExecutor::runFusion(
       }
       at::Tensor intermediate_buffer;
       if (buf_info.zero_init) {
-        intermediate_buffer = at::zeros(
-            unexpanded_sizes,
-            at::TensorOptions().dtype(buf_info.type).device(options_.device));
+        if (isOptionEnabled(EnableOption::ReuseZeroedMemory)) {
+          intermediate_buffer = contigZeroTensor(
+              unexpanded_sizes, buf_info.type, options_.device);
+        } else {
+          intermediate_buffer = at::zeros(
+              unexpanded_sizes,
+              at::TensorOptions().dtype(buf_info.type).device(options_.device));
+        }
       } else {
         intermediate_buffer = at::native::empty_cuda(
             unexpanded_sizes,
@@ -1959,6 +1965,10 @@ std::vector<at::Tensor> FusionExecutor::runFusion(
       debug() << "kernel" << kernel_id_ << " run in " << kernel_time_ms_
               << " ms, achieved: " << gb_per_s << " GB/s" << std::endl;
     }
+  }
+
+  if (isOptionEnabled(EnableOption::ReuseZeroedMemory)) {
+    releaseZeroedMemory();
   }
 
   if (isOptionEnabled(EnableOption::KernelProfile)) {
--- a/csrc/global_allocator.cpp
+++ b/csrc/global_allocator.cpp
@@ -22,6 +22,8 @@ namespace {
 // high-water mark for their particular device until the thread terminates.
 class Arena {
  public:
+  // Mark allocated_bytes_ as 0, allowing all available zeroed memory to be
+  // reused on subsequent calls to getTensor().
   void reset() {
     if (isDebugDumpEnabled(DebugDumpOption::GlobalZeroedMemory)) {
       debug() << "[global zeroed memory] Resetting allocated bytes to 0"
@@ -121,6 +123,8 @@ at::Tensor contigZeroTensor(
   return arenas[device_num].getTensor(sizes, aten_dtype, device);
 }
 
+// Note that this does not free allocated zeroed memory, but rather it marks all
+// zeroed memory as available for re-use.
 void releaseZeroedMemory() {
   for (Arena& a : arenas) {
     a.reset();
--- a/csrc/global_allocator.cpp
+++ b/csrc/global_allocator.cpp
@@ -108,10 +108,12 @@ at::Tensor contigZeroTensor(
     const c10::ScalarType& aten_dtype,
     const c10::Device& device) {
   NVF_ERROR(device.is_cuda(), "contigZeroTensor requires CUDA device");
-  int64_t device_num = device.index();
+  // Intermediate cast from int8_t to uint8_t for clarity:
+  // https://clang.llvm.org/extra/clang-tidy/checks/bugprone/signed-char-misuse.html
+  size_t device_num = (uint8_t)device.index();
 
   // get arena from device number, resizing arenas if needed
-  if ((size_t)device_num >= arenas.size()) {
+  if (device_num >= arenas.size()) {
     arenas.resize(device_num + 1);
   }
 
--- a/csrc/global_allocator.cpp
+++ b/csrc/global_allocator.cpp
@@ -1,0 +1,96 @@
+// clang-format off
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2023-present NVIDIA CORPORATION & AFFILIATES.
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+// clang-format on
+
+#include <global_allocator.h>
+#include <type.h>
+
+#include <ATen/ATen.h>
+
+namespace nvfuser {
+
+namespace {
+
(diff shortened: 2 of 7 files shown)


csrc/global_allocator.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+  // request tensor from arena
+  return arenas[device_num].getTensor(sizes, aten_dtype, device);
+}
+
+void releaseZeroedMemory() {
+  for (Arena& a : arenas) {
+    a.reset();
+  }
+}
```
jacobhinkle, line 125: A more complicated alternative would be to return a custom subclass of `at::Tensor` whose destructor marks the corresponding region as free. But given our intended usage pattern where it's always safe to release all the memory after each kernel launch, I decided to keep things simple.

csrc/global_allocator.cpp
```diff
@@ -0,0 +1,130 @@
...
+  // get arena from device number, resizing arenas if needed
+  if (device_num >= arenas.size()) {
+    arenas.resize(device_num + 1);
+  }
+
+  // request tensor from arena
+  return arenas[device_num].getTensor(sizes, aten_dtype, device);
+}
+
+void releaseZeroedMemory() {
```
naoyam, line 124: Does it actually release the memory? It looks like it only sets the buffer size to zero.

naoyam: I guess it's just me got confused with some of the names. "release" here does not really release the allocated memory but return the allocation to the Arena.

csrc/executor.cpp
```diff
@@ -1809,9 +1810,14 @@ std::vector<at::Tensor> FusionExecutor::runFusion(
       }
       at::Tensor intermediate_buffer;
       if (buf_info.zero_init) {
-        intermediate_buffer = at::zeros(
-            unexpanded_sizes,
-            at::TensorOptions().dtype(buf_info.type).device(options_.device));
+        if (isOptionEnabled(EnableOption::ReuseZeroedMemory)) {
```
naoyam, line 1813: Would there be any way to remove this option and automatically decide which to use? We know that the serial reduction can use this, so we should be able to tell the executor to do so.

jacobhinkle: Good idea. We could do that by adding a flag to kir::Allocate I think and plumbing it through. Would it ever be preferable to launch a memset kernel rather than cleaning the semaphore? Even though a kernel might not itself be persistent, the fusion will likely be run multiple times.

csrc/global_allocator.cpp
jacobhinkle: Exactly. It is confusing but I wasn't sure what exactly to name it. I am open to suggestions of course.

naoyam: No need to change the name but a quick comment would be helpful.

csrc/executor.cpp
naoyam: > Would it ever be preferable to launch a memset kernel rather than cleaning the semaphore? 
I think it's highly unlikely.

naoyam (approved):
LGTM.

This doesn't need to be in this PR:

https://github.com/NVIDIA/Fuser/pull/1984/files#r1536246159

PR #1987, opened 2024-03-22
Distributed matmul

cowanmeg:
This PR enables 1D distributed matmuls with manual tensor shardings (i.e. `ParallelType::DIDx`) that use the 
Matmul scheduler and utils are modified to ignore sharded/`DeviceDim` axes so that they schedule local matmuls. 

**Modifications:**
MmaScheduler:
 MmaDetails and MatmulScheduler are updated to ignore DeviceDims when analyzing if a mul-sum pair can be replaced with an Mma op and updates the axes that intra-device matmul schedules. 
Both assume the `DeviceDim` axes are located on the outermost dimensions and only work when sharding is on the the RFactor domain.
  
**Distributed Matmul tests:**
For matmul(A,B) = C 

| Layout | A's sharded axis | B's sharded axis | C's sharded axis | Communication|
| -- | -- | -- | -- | -- |
| TN | M | x | M | None|
| TN | M | x |  x | AllGather|
| NT | K | K |  x | AllReduce|
| NT | K | K |  M | ReduceScatter|

**Follow up:**

- Batched distributed matmul
- LayoutNT_ReduceScatter requires a hack with `segment_set`


cowanmeg:
!build --dist

tests/cpp/test_multidevice_pipeline.cpp
```diff
@@ -576,6 +581,304 @@ INSTANTIATE_TEST_SUITE_P(
...
+    return std::make_tuple(a, b, c);
+  }
+
+ private:
+  preseg_passes::OptimizationPassGuard<preseg_passes::AllocationDomainPass>
+      optimization_guard_;
+  DisableOptionsGuard option_guard_;
+};
+
+TEST_F(DistributedMatmul, LayoutTN_NoComms) {
```
samnordmann, line 642: Is it possible to refactor the four different layouts in one test depending on a "layout param"? This would reduce duplication a lot and will be easier to read maintain

tests/cpp/test_multidevice_pipeline.cpp
```diff
@@ -576,6 +581,304 @@ INSTANTIATE_TEST_SUITE_P(
         SchedulingMode::Manual,
         SchedulingMode::ReductionOnly,
         SchedulingMode::Automatic));
+
+class DistributedMatmul : public MultiDeviceTest {
```
samnordmann, line 585: is there a reason not to use `PipelineTest` as base class? We could reuse the data members and the method executeAndValidate

samnordmann (commented):
That looks great! Thank you for the pr! This is an important milestone for us :)

csrc/multidevice/utils.cpp
```diff
@@ -223,11 +223,38 @@ void shardAllLike(TensorView* ref, std::vector<TensorView*> tvs) {
     scheduler_utils::parallelizeAllLike(ref, tvs, {ParallelType::DIDx});
   }
 }
+
+// TODO: Very simple sharding propagation pass.
+// Auto inserted ops like castOps don't propagate shardings
```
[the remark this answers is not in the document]
samnordmann: Can you describe the example for which this is needed?

csrc/multidevice/utils.cpp
```diff
@@ -223,11 +223,38 @@ void shardAllLike(TensorView* ref, std::vector<TensorView*> tvs) {
...
+    }
+  }
+  shardAllLike(input_with_mesh[0], outputs_without_mesh);
+}
 } // namespace
 
 void insertReshardings(Fusion* fusion) {
   auto exprs = fusion->exprs();
   for (auto expr : exprs) {
+    propagateShardings(expr);
```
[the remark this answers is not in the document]
wujingyue: I second this. That will also help to split up the PR, which I would do. For example, one PR implements shard propagation and tests that check leaf domains are propagated correctly, and the other PR changes the matmul scheduler and tests a sharded matmul example end-to-end.

csrc/multidevice/utils.cpp
```diff
@@ -240,17 +267,32 @@ void insertReshardings(Fusion* fusion) {
...
+      // For expressions with > 1 input, insert the set before the expr
+      // for each input (input [set] new_input) [expr] output
+      for (auto input : inputs) {
+        // TODO: reuse cacheAfter?
+        // TODO: here we should add a mechanism to potentially reuse the
+        // inserted resharding accross all the consumer of the resharded tensor.
+        // This way we could avoid wasteful resharding set insertion.
+        TensorView* new_input = set(input);
+        new_inputs.push_back(new_input);
+        expr = ir_utils::replaceValInExprInputs(expr, input, new_input);
```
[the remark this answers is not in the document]
cowanmeg: No, only expr's input can be safely updated. Since `input` can be an input to multiple expressions we don't know if `new_input` has the right shardings for the other expressions.

csrc/multidevice/utils.cpp
```diff
@@ -240,17 +267,32 @@ void insertReshardings(Fusion* fusion) {
...
-      // resharding accross all the consumer of the resharded tensor. This way
-      // we could avoid wasteful resharding set insertion.
-      TensorView* new_input = set(input);
-      new_inputs.push_back(new_input);
-      expr = ir_utils::replaceValInExprInputs(expr, input, new_input);
+    auto inputs = getTvsWithDifferentSharding(
+        output, ir_utils::filterByType<TensorView>(expr->inputs()));
+    // Insert resharding expression after the expr when there is only one input.
+    // input [expr] output [set] new_output
+    if (!inputs.empty() && expr->inputs().size() == 1) {
```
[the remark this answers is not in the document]
cowanmeg: The motivation was just simplicity for this PR. For an expression with more than input, if we add the resharding after the expr, we will have to decide which input we want to shardLike and potentially reshard some inputs, because the inputs might not have consistent sharding.

csrc/multidevice/utils.cpp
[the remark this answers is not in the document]
cowanmeg: No, the castOp's that are automatically inserted (for this example, it is binaryOp multiply) happen before manual scheduling so no TensorView has any DeviceMesh or parallelization yet.

tests/cpp/test_multidevice_pipeline.cpp
cowanmeg: The matmul tests need some extra constraints, some of which are temporary, like disabling the allocation presegmentation pass, ampere architecture or newer, increased validation tolerance, etc.

tests/cpp/test_multidevice_pipeline.cpp
```diff
@@ -576,6 +581,304 @@ INSTANTIATE_TEST_SUITE_P(
...
+ protected:
+  DistributedMatmul() : optimization_guard_(false) {
+    DisableOptionsGuard::getCurOptions().set(DisableOption::MatmulExprEval);
```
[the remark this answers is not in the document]
cowanmeg: There is a pathway to run matmuls through expression evaluator instead of applying the scheduler, so this flag disables that option to ensure the teset uses the scheduler.

tests/cpp/test_multidevice_pipeline.cpp
```diff
@@ -576,6 +581,304 @@ INSTANTIATE_TEST_SUITE_P(
...
+    auto b =
+        matmulAtInput2D(layout, TensorMatmulPos::B, type, M, N, K, 0, device);
+    auto c =
+        atMatmul(a.to(at::kDouble), b.to(at::kDouble), layout).to(at::kFloat);
+    return std::make_tuple(a, b, c);
+  }
+
+ private:
+  preseg_passes::OptimizationPassGuard<preseg_passes::AllocationDomainPass>
+      optimization_guard_;
```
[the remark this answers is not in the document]
cowanmeg: yes

tests/cpp/test_multidevice_pipeline.cpp
```diff
@@ -576,6 +581,304 @@ INSTANTIATE_TEST_SUITE_P(
...
+  DisableOptionsGuard option_guard_;
```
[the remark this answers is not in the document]
cowanmeg: yes

tests/cpp/test_multidevice_pipeline.cpp
cowanmeg: I would argue that would made the tests harder to read since some of the tests diverge and we would need to add many if statements to check which layout and which shardings.

tests/cpp/test_multidevice_pipeline.cpp
```diff
@@ -576,6 +581,304 @@ INSTANTIATE_TEST_SUITE_P(
...
+    MultiDeviceTest::SetUp();
+    num_devices = communicator->size();
+  }
+
+  // use_fusion_executor cache=true, skip_auto_scheduling=false
+  // cache_fusion_executor=false
+  MultiDeviceExecutorParams executor_params{true, false, false};
+  int num_devices;
+
+  DeviceMesh createDeviceMesh() {
```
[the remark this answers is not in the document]
cowanmeg: Good idea

csrc/scheduler/matmul.cpp
```diff
@@ -900,11 +900,14 @@ void scheduleMatmul(Fusion* fusion, const MatmulParams& params) {
...
   mma_utils::canonicalizeMmaTvOrdering(mma_result);
+  int nLocalDims = TensorDomain::noDevices(mma_result->getLeafDomain()).size();
   NVF_ERROR(
-      mma_result->nDims() == 3 || mma_result->nDims() == 4,
+      nLocalDims == 3 || nLocalDims == 4,
       "Currently, we only support B, M, N and K being a single dimension.",
       " More general tensor contraction is not supported yet.");
-  const int num_batch_dims = (int)mma_result->nDims() - 3;
+  const int offset_local_dim = (int)mma_result->nDims() - 3;
+  const bool has_batch_dim = (nLocalDims - 3 == 1);
```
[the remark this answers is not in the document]
cowanmeg: My bad, I did not document the constraints. This PR only modifies matmul scheduler to support non-batch shardings, and also assumes the sharded dimension is outermost. 
Will update with checks that verify this.

csrc/multidevice/utils.cpp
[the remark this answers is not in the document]
cowanmeg: Good idea.

csrc/multidevice/utils.cpp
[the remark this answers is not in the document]
samnordmann: right, nevermind!

csrc/multidevice/utils.cpp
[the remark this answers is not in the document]
samnordmann: right, nevermind!

naoyam:
@jacobhinkle Could you also take a look at the mma scheduler changes?

csrc/scheduler/matmul.cpp
```diff
@@ -900,11 +900,14 @@ void scheduleMatmul(Fusion* fusion, const MatmulParams& params) {
...
+  const int batch_dim = (int)mma_result->nDims() - nLocalDims;
```
jacobhinkle, line 910: Could we place a comment indicating what are the dimensions at this point? I think we are assuming device dims are outermost, so this would be something like
```c++
  // [ (device dims), (batch dims), M, N, K ]
```
Instead of assuming devices are outermost, perhaps we could reorder to make that so adding the following to `mma_utils::canonicalizeMmaTvOrdering`:
```c++
  std::unordered_map<int, int> old2new;
  int next_device_dim = 0;
  for (int i : c10::irange(tv->nDims()) {
    if (tv->axis(i)->isDeviceDim()) {
      old2new[i] = next_device_dim++;
    }
  }
  tv->reorder(old2new);
}
```
That should ensure that the commented dims actually do match reality regardless of input.

jacobhinkle (commented):
The changes look good! I had some minor suggestions, particularly reordering the device dims to outermost instead of trusting that they will already be in that order. I'll let the other reviewers approve as I have only looked at `csrc/scheduler/*` changes.

csrc/multidevice/communicator.h
```diff
@@ -108,6 +108,11 @@ class Communicator {
     return rankToDiD(rank_);
   }
 
+  // returns local rank
+  RankType local_rank() const {
+    return local_rank_;
```
wujingyue, line 113: Do you mind code-comment `local_rank_`? E.g. how is it different from `rank_`? I suspect it's related to the concept of "node", but I failed to find what node means in this file either.

tests/cpp/test_multidevice_pipeline.cpp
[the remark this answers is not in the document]
wujingyue: Can you at least code-comment why? Is it because of https://github.com/NVIDIA/Fuser/pull/2044#discussion_r1554143448, which needs to be fixed?

tests/cpp/test_multidevice_pipeline.cpp
```diff
@@ -576,6 +581,302 @@ INSTANTIATE_TEST_SUITE_P(
...
+    MultiDeviceTest::SetUp();
+    num_devices = communicator->size();
+  }
+
+  // use_fusion_executor cache=true, skip_auto_scheduling=false
+  // cache_fusion_executor=false
+  MultiDeviceExecutorParams executor_params{true, false, false};
+  int num_devices;
+
+  ValidationConstants getTolerances() {
```
wujingyue, line 604: It's a bug if we have to manually tune these tolerances. The auto-computed threshold by testValidate should have already considered these errors caused by reduction order. I don't see why multi-device should be an exception. 

Can you verify the single-device version can pass without customizing the threshold? If so, we (cc @Priya2698) should try to understand why the threshold for multi-device has to be larger.

wujingyue:
> Both assume the DeviceDim axes are located on the outermost dimensions and only work when sharding is on the the RFactor domain.

I got conflicting signals about what domain is sharded, so I'd appreciate your clarification. My impression so far is that DID has been used in only leaf domain (and therefore not rfactor). Is this PR the first to DID-parallelize rfactor domain?

cowanmeg:
> > Both assume the DeviceDim axes are located on the outermost dimensions and only work when sharding is on the the RFactor domain.
> 
> I got conflicting signals about what domain is sharded, so I'd appreciate your clarification. My impression so far is that DID has been used in only leaf domain (and therefore not rfactor). Is this PR the first to DID-parallelize rfactor domain?

Currently, we only DID parallelize the rfactor domain and do not allow split/merge, so that restrictions ensures the same DID parallelized iter domain is in both rfactor and leaf (correct me if my assumption is wrong!). 

We haven't fully discussed this plan, but from various PR discussions it sounds like we are waiting for IdModels to support split/merge + DID parallelize.

csrc/scheduler/matmul.cpp
cowanmeg: Thanks! This is a much better solution!

csrc/scheduler/matmul.cpp
```diff
@@ -900,11 +900,14 @@ void scheduleMatmul(Fusion* fusion, const MatmulParams& params) {
...
   // ------------------------------------------------------------------
   mma_utils::canonicalizeMmaTvOrdering(mma_result);
+  int nLocalDims = TensorDomain::noDevices(mma_result->getLeafDomain()).size();
   NVF_ERROR(
-      mma_result->nDims() == 3 || mma_result->nDims() == 4,
+      nLocalDims == 3 || nLocalDims == 4,
       "Currently, we only support B, M, N and K being a single dimension.",
       " More general tensor contraction is not supported yet.");
-  const int num_batch_dims = (int)mma_result->nDims() - 3;
+  const int offset_local_dim = (int)mma_result->nDims() - 3;
```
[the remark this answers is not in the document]
cowanmeg: Updated this code to create the variables you suggested. Let me know if this is more clear.

csrc/multidevice/communicator.h
cowanmeg: Do you think "rank within a machine/node" more clear?

csrc/multidevice/utils.cpp
```diff
@@ -267,17 +274,33 @@ void insertReshardings(Fusion* fusion) {
...
-      // we could avoid wasteful resharding set insertion.
-      TensorView* new_input = set(input);
-      new_inputs.push_back(new_input);
-      expr = ir_utils::replaceValInExprInputs(expr, input, new_input);
+    auto inputs = getTvsWithDifferentSharding(
+        output, ir_utils::filterByType<TensorView>(expr->inputs()));
+
+    // Insert resharding set after the expr when there is only one input.
+    // input [expr] output [set] new_output
+    if (!inputs.empty() && expr->inputs().size() == 1) {
```
wujingyue, line 282: Does it make sense to unit-test the change in insertReshardings, e.g., in tests/cpp/test_multidevice_sharding.cpp? These tests should run just fine on a single-device system so regressions are easier to discover.

tests/cpp/test_multidevice_pipeline.cpp
```diff
@@ -576,6 +581,362 @@ INSTANTIATE_TEST_SUITE_P(
         SchedulingMode::Manual,
         SchedulingMode::ReductionOnly,
         SchedulingMode::Automatic));
+
+class DistributedMatmulTest : public MultiDeviceTest {
```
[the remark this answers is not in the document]
cowanmeg: I had the same feeling and moved it to a separate file with some uncommitted work. We will have more distributed matmul tests in the future, so it makes sense for a separate file.

csrc/multidevice/utils.cpp
cowanmeg: Yeah I should probably test this... I think I'll pull this change out into a separate PR with some tests.

naoyam:
[celebrate]     Naoya Maruyama reacted to your message:
________________________________

naoyam:
[0]     Naoya Maruyama reacted to your message:
________________________________

tests/cpp/test_multidevice_pipeline.cpp
cowanmeg: One of the failing tests is LayoutTN_NoComms which doesn't involve any collective communication.
The matmul in that test is of size M=1024, N=512, K=256.
On D devices, the local matmul size is M=512/D, N=512, K=256.

The error is not off by a lot. For example, on 2 devices I see a slightly higher error.
Validation error in output 0 on line 145 in file /home/mcowan/Fuser/tests/cpp/test_multidevice_matmul.cpp.
Detected abs error of: 6.86646e-05
absolute tolerance was set to 2.19122e-05
and relative tolerance set to 2.19122e-07

Distributing across 4 results in a bit more error, but still not much:
Validation error in output 0 on line 145 in file /home/mcowan/Fuser/tests/cpp/test_multidevice_matmul.cpp.
Detected abs error of: 7.62939e-05
absolute tolerance was set to 2.19122e-05
and relative tolerance set to 2.19122e-07

tests/cpp/test_multidevice_pipeline.cpp
```diff
@@ -576,6 +581,362 @@ INSTANTIATE_TEST_SUITE_P(
...
+  TensorView* b = makeContigTensor(3, DataType::Half); // (Ko,Ki,N)
+  TensorView* a_t = transpose(a, 1, 2); // (Ko, M, Ki)
+  TensorView* b_t = transpose(b, 1, 2); // (Ko, N, Ki)
+  TensorView* a_b = broadcast(a_t, {false, false, true, false}); // (Ko,M,b,Ki)
+  TensorView* b_b = broadcast(b_t, {false, true, false, false}); // (Ko,b,N,Ki)
+  TensorView* ab = mul(a_b, b_b); // (Ko,M,N,Ki)
+  TensorView* c0 = sum(ab, {-1}); // (Ko,M,N,r)
+  c0 = segment_set(c0);
+  // TODO: Reshape works using sharded sizes. Should use unsharded size to
+  // stay consistent.
```
wujingyue, line 895: This sounds like a bug. How hard is it to fix that? Indeed it creates inconsistency, e.g., I'd expect `c1->axis(0)->parallelize(DIDx)` somewhere.

cowanmeg: > it may actually be `c1->axis(1)->parallelize(DIDx)`

This should be `c1->axis(0)->parallelize(DIDx)`, I should have inlucded it in the list of sharded_tvs (will also rename to k_sharded_tvs to not confuse with c which is sharded on M). 
That might fix the reshape problem.

wujingyue: > This should be c1->axis(0)->parallelize(DIDx)

You are right -- the reshape is intra-device (so dim 0 should be parallelized on DIDx), and the sum is cross-device (so dim 1 DIDx).

csrc/multidevice/utils.cpp
cowanmeg: Moved to a separate PR. This one is updated with explicit sets so that it doesn't need the changes in `insertReshardings`

tests/cpp/multidevice.cpp
```diff
@@ -63,6 +63,12 @@ void MultiDeviceTest::TearDown() {
   NVFuserTest::TearDown();
 }
 
+DeviceMesh MultiDeviceTest::createDeviceMesh() {
```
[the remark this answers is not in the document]
cowanmeg: updated to make size an argument.

tests/cpp/test_multidevice_pipeline.cpp
Priya2698: > The auto-computed threshold by testValidate should have already considered these errors caused by reduction order. I don't see why multi-device should be an exception.

Reduction order can play a role in the higher errors, however torch baselines don't have a way to tune the reduction order. The current method only looks at the reduction size.

For python benchmarking, where we see validation errors, we are considering relaxing tolerances by ~10%. CC: @naoyam 

> Can you verify the single-device version can pass without customizing the threshold? If so, we (cc @Priya2698) should try to understand why the threshold for multi-device has to be larger.

The error constants were generated by executing a large number of sum reductions and noting the max error. These experimentally generated constants were from single-device reductions. This may also be a reason for differing error constants.

Priya2698: The reference computation should also be in `double`.

wujingyue: > we are considering relaxing tolerances by ~10%

Good to know. But doesn't seem large enough to tolerate this case though (2.19e-5 vs 6.86e-5). 

> The error constants were generated by executing a large number of sum reductions and noting the max error.

Got it -- it's still empirical and not intended to capture the reduction order that triggers the largest difference. 

@Priya2698 what do you recommend are the next steps? Doing nothing will probably lead to more manual threshold customizations like this for multidevice tests.

Priya2698: For single-device cases, I had regenerated validation constants in the past (https://github.com/NVIDIA/Fuser/pull/1593, closed unmerged) but it hadn't helped completely resolve the issue.

For multi-device, we should run a similar experiment to see if the experimentally generated constants differ significantly from existing constants with the number of devices (2/4/8). We may need two sets of constants

cowanmeg: Small update - I've (1) updated the computation to be done in double, then cast back to float and (2) shrunk K to 64. 
Surprisingly, the allreduce and reduce scatter tests are passing.I thought they would have a larger error since the reduction is both local and across devices, but it is the allgather and no communication (local matmul where we either gather shards from different devices or the output stays sharded) are still slightly off. 

Detected abs error of: 1.14441e-05
    absolute tolerance was set to 8.57934e-06
    and relative tolerance set to 8.57934e-08
    
It looks like we would need even more relaxation then 10%, maybe 2x?

Priya2698: Yes, this seems like a much larger error. 
It looks like the experimentally generated error constants for single device are not applicable to multi-device.

cowanmeg: > We may need two sets of constants

This is probably true... the ring allreduce and reduce-scatter (default implementations for these collectives within a node) are known to be less numerically stable. It won't be an issue on 2 or 4 devices since the number of devices is low, but by 8 devices we might need more tolerance.

Priya2698: The above error numbers are for how many devices? 
Are you running into errors on single device as well with `double` reference computations?

cowanmeg: Above errors were for 2 devices. M=1024,N=512,K=64
I also tried a single device test, with double reference computations, and it was failing.

But, M=256,N=64,K=64 is passing for distributed and single device tests. I tested with 2 and 4 devices so looks promising. I guess the large non-reduction axis sizes were increasing the chances of one element being off by too much?

cowanmeg:
!build --dist

cowanmeg:
!build --dist

cowanmeg:
!build --dist

csrc/multidevice/utils.h
```diff
@@ -19,6 +19,9 @@ namespace nvfuser {
 // Checks that the other non-reduction axis are not parallelized on Didx
 NVF_API bool isSharded(TensorView*);
 
+// Returns number of device dimensions in a TensorView's leaf domain.
+NVF_API int64_t numDeviceDims(TensorView*);
```
wujingyue, line 23: https://github.com/NVIDIA/Fuser/blob/main/doc/dev/visibility.md#should-i-mark-my-new-method-or-class-as-nvf_api

(We can clean it up in a separate PR like https://github.com/NVIDIA/Fuser/pull/2161)

tests/cpp/test_multidevice_matmul.cpp
```diff
@@ -0,0 +1,302 @@
...
+  DistributedMatmulTest() : optimization_guard_(false) {
+    DisableOptionsGuard::getCurOptions().set(DisableOption::MatmulExprEval);
+  }
+
+  void SetUp() {
+    if (!deviceMajorMinorCheck(8)) {
+      GTEST_SKIP() << "Distributed matmul tests require Ampere or newer";
+    }
+    MultiDeviceTest::SetUp();
+    num_devices = communicator->size();
```
wujingyue, line 48: With https://github.com/NVIDIA/Fuser/pull/2126/, this can now be moved to the constructor and be made `const`.

tests/cpp/test_multidevice_matmul.cpp
```diff
@@ -0,0 +1,302 @@
...
+ protected:
+  DistributedMatmulTest() : optimization_guard_(false) {
+    DisableOptionsGuard::getCurOptions().set(DisableOption::MatmulExprEval);
+  }
+
+  void SetUp() {
+    if (!deviceMajorMinorCheck(8)) {
+      GTEST_SKIP() << "Distributed matmul tests require Ampere or newer";
+    }
+    MultiDeviceTest::SetUp();
```
wujingyue, line 47: Nit: with https://github.com/NVIDIA/Fuser/pull/2126/, SetUp merely skips tests, so I'd **slightly** prefer calling Superclass::SetUp first. The checks in the superclass tend to be more coarse-grained and permissive, sometimes making the fine-grained checks in subclass easier. I know the order doesn't matter to this particular case, but I think it's still a good general practice. 

(For similar reasons, if we have TearDown, I'd call SuperClass::TearDown last). 


are usually moreht  (more coarse-grained checks) first.

tests/cpp/test_multidevice_matmul.cpp
```diff
@@ -0,0 +1,302 @@
...
+    num_devices = communicator->size();
+  }
+
+  MultiDeviceExecutorParams executor_params{
+      .use_fusion_executor_cache = true,
+      .skip_auto_scheduling = false,
+      .cache_fusion_executor = false};
+  int num_devices;
```
wujingyue, line 55:
```suggestion
```

Doesn't seem to be used.

tests/cpp/test_multidevice_matmul.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+  std::tuple<at::Tensor, at::Tensor, at::Tensor> getAtenInputOutputs(
```
wujingyue, line 57:
```suggestion
  std::tuple<at::Tensor, at::Tensor, at::Tensor> getInputsAndReferenceOutputs(
```

tests/cpp/test_multidevice_matmul.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+  // Sharding M dimension
+  auto all_sharded_tvs = {a, a_b, b_b, ab, c};
+  for (auto tv : all_sharded_tvs) {
+    tv->axis(0)->parallelize(ParallelType::DIDx);
+    tv->setDeviceMesh(mesh);
+  }
+  b->setDeviceMesh(mesh);
+
+  auto [a_, b_, c_] = getAtenInputOutputs(MmaLayout::TN, M, N, K);
```
wujingyue, line 113: Just a note: the codebase has been using `name_` for class fields, so this had confused me until I realized the underscore is to indicate the tensor corresponding to TensorView `name`. This is OK unless you figure out a better way of naming.

wujingyue (approved):

tests/cpp/test_multidevice_matmul.cpp
wujingyue: Ignore the above -- I just realized this is to declare a field.

wujingyue, line 55:
```suggestion
  int num_devices_;
```

Also, can this and `executor_params` be moved to private?

tests/cpp/test_multidevice_matmul.cpp
```diff
@@ -0,0 +1,302 @@
...
+      .skip_auto_scheduling = false,
+      .cache_fusion_executor = false};
+  int num_devices;
+
+  std::tuple<at::Tensor, at::Tensor, at::Tensor> getAtenInputOutputs(
+      MmaLayout layout,
+      int M,
+      int N,
+      int K) {
+    at::manual_seed(0);
```
wujingyue, line 62: Is there a particular reason for this? Note that, NVFuserTest, the base test fixture, explicitly sets seed for determinism already.

tests/cpp/test_multidevice_matmul.cpp
wujingyue: But that does motivate the naming convention of `field_` in the codebase :)

csrc/ir/nodes.cpp
```diff
@@ -2374,8 +2374,12 @@ std::vector<PolymorphicValue> ViewOp::evaluate(
...
   out_shape.reserve(out_rfactor.size());
   for (IterDomain* id : out_rfactor) {
-    out_shape.push_back(
-        ee.evaluate(id->getMaybeExpandedExtent()).as<int64_t>());
+    if (id->isDeviceDim()) {
+      out_shape.push_back(1);
+    } else {
+      out_shape.push_back(
+          ee.evaluate(id->getMaybeExpandedExtent()).as<int64_t>());
+    }
```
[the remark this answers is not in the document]
cowanmeg: That is a good question. I don't quite know the answer, but my guess is the `ViewOp` is evaluated by FusionExecutor because the reshape is the output of a segment, since the sum is over a DID axis (maps to reduce scatter)
```
  std::vector<int64_t> orig_size = {K, M, N};
  std::vector<int64_t> new_size = {K, Mo, Mi, N};
  TensorView* c1 = reshape(c0, orig_size, new_size);
  TensorView* c = sum(c1, {0});
```

cowanmeg:
!build --dist

cowanmeg:
!build --dist

PR #1989, opened 2024-03-23
TMA: validate view and size of transfer

zasdfgbnm:


zasdfgbnm:
!build

zasdfgbnm:
!build

zasdfgbnm:
!build

csrc/device_lower/lower2device.h
```diff
@@ -238,6 +240,17 @@ class GpuLower : public NonCopyable {
     return passes_;
   }
 
+  template <typename... Args>
+  void validate(Val* val, Args... args) {
+    auto sv = simplifyExpr(val);
+    if (sv->isTrue()) {
```
drzejan2, line 246: We don't need to store in validation `Val*` objects that are simplified to bool True?

Just to be sure, Val* is ok for us as long as it's not evaluated to bool False, that is, we will think about Val with division results as valid, only because the stored value is not bool False?

tests/cpp/test_memory.cpp
```diff
@@ -952,8 +950,56 @@ TEST_F(TMARuntimeInvalidTest, SizeOfTransfer) {
...
+      at::TensorOptions().dtype(data_type_to_aten(dtype)).device(at::kCUDA, 0);
+  auto t0_valid = at::randn({10240}, options);
+  FusionExecutor fe;
+  fe.compileFusion(&fusion, {t0_valid}, {}, matmul_cparams);
+  auto cg_outputs = fe.runFusion({t0_valid});
+  testValidate(&fusion, cg_outputs, {t0_valid}, {t0_valid}, __LINE__, __FILE__);
+
+  EXPECT_THAT(
+      [&]() {
+        auto t0_inval = at::randn({10249}, options);
```
drzejan2, line 998: For future reference, can you add a comment why this tensor is invalid for given scheduling / what rule does it violates?

drzejan2 (approved):
I left comments, which are about some minor details, otherwise changes look good.

Thanks!

csrc/device_lower/lower2device.h
zasdfgbnm: If it is simplified to true, then this condition is always satisfied regardless of the input of the kernel. For this case, we do not want any check in the runtime. For example, we are not interested in checking that `3 < 4` each time we launch the kernel. We know that it is always true, and just skip it. I will update the code mentioning that.

zasdfgbnm: added a comment in the code

tests/cpp/test_memory.cpp
zasdfgbnm: added

csrc/device_lower/lower2device.h
```diff
@@ -238,6 +240,21 @@ class GpuLower : public NonCopyable {
     return passes_;
   }
 
+  template <typename... Args>
+  void validate(Val* val, Args... args) {
```
naoyam, line 244:
```suggestion
  // Register a boolean Val as a predicate to validate at the run time. Optional validation error messages can be given as args.
  template <typename... Args>
  void validate(Val* validation_condition, Args... args) {
```

naoyam (approved):
LGTM. Just added a suggestion to improve clarity

csrc/device_lower/lower2device.h
drzejan2: Thank you for explanation!

tests/cpp/test_memory.cpp
drzejan2: Thanks!

PR #1991, opened 2024-03-24
Define box by compositing

zasdfgbnm:
Add support for "define box by compositing".


zasdfgbnm:
!build

zasdfgbnm:
!build

zasdfgbnm:
!build

csrc/index_compute.cpp
```diff
@@ -3460,12 +3464,14 @@ Val* Index::eye(
 //           \               \-> I4
 //            \-> I2 -> split -> I5 (Bulk)
 //                           \-> I6
+// I7
 //
 // Originating bulk IterDomains: { I3, I5 }
 // Box IterDomains: { I3, I5 }
 // Inner IterDomains: {}
 // TMA-global IterDomains: { I1, I2 }
 // TMA domain: [I2, I1].
```
zasdfgbnm, line 3473:
```suggestion
// TMA domain: [I7, I2, I1].
```

csrc/index_compute.cpp
```diff
@@ -3338,18 +3338,19 @@ Val* Index::eye(
...
-// We call [I1, I2] the "TMA domain" of the tensor.
+// We call [I0, I1, I2] the "TMA domain" of the tensor.
 //
 // Second, we need to tile each dimension of the TMA domain of the tensor. If we
 // want a contiguous box dim (see Definition 3: "boxId" below), then we do one
 // split to get the box. If we want a strided box dim, then we do two splits,
 // the first split is to get the box, and the second split is to get the stride.
-// In the above diagram, I1 has a contiguous box dim, and I2 has a strided box
-// dim.
+// Note that if the tile size is 1, we can skip tiling, for this case, we call
```
naoyam, line 3351: To clarify, does this mean there's only one tile or there may be multiple tiles, each of which is a size-1 tile? I suppose it's the former.

zasdfgbnm: I am referring to the later. There can be multiple tiles, but the tile size must be one. For this case, we do not need to create a Bulk ID. For the former case, where there is only one tile, what we need is to parallelize the entire dim as `Bulk`, instead of doing a split and parallelize the inner as `Bulk`. The former case is done in https://github.com/NVIDIA/Fuser/pull/2004, which is currently WIP.

naoyam: Hmm, does the size-1 tile mean a 1D tile with dimension being 1 or just a 1D tile? It should be the latter since otherwise there's no point of TMA, right?

zasdfgbnm: Oh, I see the confusion here. What I want to say is, "if the size of a box dimension is 1". So it is not necessarily a 1D tile. It can be a tile like, for example, `(1, 4)`.

zasdfgbnm: For example, without this PR, if I have a tensor of shape `[1024, 1000, 128]`, and I want to load with tile size `[1, 16, 4]`, then I have to schedule the fusion like `[1024/1, 1 (bulk), 1000/16, 16 (bulk), 128/4, 4 (bulk)]`. With this PR, I can omit the first split, and leave the fusion as `[1024, 1000/16, 16 (bulk), 128/4, 4 (bulk)]`

naoyam: Thanks. That clarifies my confusion indeed. I think it'd be more clear to say "a dimension of a tile is 1".

csrc/index_compute.cpp
```diff
@@ -3460,12 +3464,14 @@ Val* Index::eye(
 //           \               \-> I4
```
drzejan2, line 3464: Maybe we could improve readability of the code:
```
// -> I0 -> split -> I1 -> split -> I3 (Bulk)
//              \               \-> I4
//               \-> I2 -> split -> I5 (Bulk)
//                              \-> I6
// -> I7
```

At the beginning I though that `I7` is some kind of artifact.

drzejan2 (dismissed):
Changes look good, although I will need some more time to wrap my mind around merging non-TMA-global IDs.

Thanks!

csrc/index_compute.cpp
zasdfgbnm: I changed it to `a box dimension has size 1` in https://github.com/NVIDIA/Fuser/pull/2004 in https://github.com/NVIDIA/Fuser/pull/2004/commits/095101f267109ef52cd8a2a78d46c7f5e5e0e5e3.
I will try to leave this PR as-is (as long as there is no functional error), and resolve all PR reviews in https://github.com/NVIDIA/Fuser/pull/2004. This way, I don't need to deal with merge conflicts.

csrc/index_compute.cpp
```diff
@@ -3878,38 +3885,98 @@ std::pair<Val*, Val*> Index::getCpAsyncBulkGmemIndex(
 
   int64_t itemsize = dataTypeSize(gmem_tv->dtype());
 
+  // Loop through the frontier from inner to outer, handle TMA-global ID and
```
naoyam, line 3888: Could you also give a simple illustrative example?

csrc/index_compute.cpp
zasdfgbnm: Added in https://github.com/NVIDIA/Fuser/pull/2004 in https://github.com/NVIDIA/Fuser/pull/2004/commits/f7869f4de4527b0a449470d84bf31d65fa0e2763. I am resolving reviews in followup PRs so that I do not need to handle merge conflicts.

csrc/index_compute.cpp
zasdfgbnm: Added a few examples in https://github.com/NVIDIA/Fuser/pull/2004/commits/43a43d181144417b52c94449d1195320142e8961
Example 1 and 2 are relavent to this PR.

csrc/index_compute.cpp
```diff
@@ -3878,38 +3885,98 @@ std::pair<Val*, Val*> Index::getCpAsyncBulkGmemIndex(
...
+  //               '-> [START] --'
+  //                   /    ^
+  //   non-TMA-global /      \ TMA-global
+  //        ID       |        |   ID
+  //                  \      /
+  //                   v    /
+  //              .-> [PENDING] --.
+  //              |               |
+  //              '---------------'
+  //              non-TMA-global ID
```
naoyam, line 3909: Not sure how to interpret this state diagram.  What does the state imply? Wonder what it means when the state returns to START. 

As for the arrow, I suppose the label refers to the outer domain of the current domain, correct? The next state means the state of the domain when two domains are merged?

zasdfgbnm: Looks like the doc needs a lot more clarification. I did it in https://github.com/NVIDIA/Fuser/pull/2004/commits/989092638145aa5f2bfc7b7171d381165495ea4b, hopefully it will be more understandable. Every time when I use a state machine, I feel that it does make the code clean and neat, but it also makes the idea very difficult to explain. This happened a few times in the past... Hopefully, I answered all your questions in the updated doc.

zasdfgbnm:
TODO: check dim of TMA

zasdfgbnm:
!build

zasdfgbnm:
!build

zasdfgbnm:
@naoyam This is ready for review again

csrc/index_compute.cpp
```diff
@@ -3696,38 +3675,112 @@ std::pair<Val*, Val*> Index::getCpAsyncBulkGmemIndex(
...
+  //
+  // There can only be four types of IterDomains in the TMA domain:
+  // -  P: partitioned IterDomain
+  // -  C: coordinate IterDomain
+  // - SB: strided box IterDomain
+  // - CB: contiguous box IterDomain
+  //
+  // For the example of the Figure 6 in doc/dev/tma.md, the TMA domain is
+  // [I1, I2, I3, I4, I5, I6, I7, I8, I9], and the types of these IDs are
+  // [ C,  B,  P,  C,  B,  B,  C,  B,  B]
```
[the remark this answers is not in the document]
zasdfgbnm: Oh sorry, you are right. I was originally thinking that we only need one type of `B`, but later I realized we need to distinguish `B` and `CB`, but I forget to update this.

csrc/index_compute.cpp
```diff
@@ -3696,38 +3675,112 @@ std::pair<Val*, Val*> Index::getCpAsyncBulkGmemIndex(
...
+  //
+  // The algorithm works as follows: We run a 3-state machine. The state machine
+  // is initialized as START. After setting the initial state, we loop through
+  // the TMA domain from inner to outer. During the loop, for each IterDomain we
+  // see, we take an action and change the state of the machine. The action and
+  // target state depend on the current state of the machine, and the type and
+  // contiguity of the IterDomain we encounter. The actions and transition of
+  // states are shown in the following diagram:
```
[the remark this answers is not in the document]
zasdfgbnm: Yes, for example, if you have a tensor [I1, I2, I3] with contiguity [F, F, T], and you schedule like:
```C++
coord3, tile3 = split(I3, 4);
// I1 and I2 are left untouched
```
then after visit `I2`, the machine will be in state `PENDING COORD`, and it will then visit `I1`, which is a discontiguous coordinate ID.

csrc/index_compute.cpp
```diff
@@ -3696,38 +3675,112 @@ std::pair<Val*, Val*> Index::getCpAsyncBulkGmemIndex(
...
+  //
```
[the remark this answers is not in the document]
zasdfgbnm: Yes, in the example https://github.com/NVIDIA/Fuser/pull/1991#discussion_r1574259005, after visiting `I3`, the machine will be in state `START`. Then when visiting `I2`, it will hit `C`. Note that getting a `C` in `START` always indicate that there is a new dimension which has implicit box size 1.

csrc/index_compute.cpp
```diff
@@ -3659,29 +3632,35 @@ std::pair<Val*, Val*> Index::getCpAsyncBulkGmemIndex(
...
+      tma_id_to_stride_id.count(std::get<0>(tma_domain.back())) == 0,
+      "When interleave is CU_TENSOR_MAP_INTERLEAVE_NONE ",
+      "(this is always the case for nvFuser now)",
+      ", the first element of elementStrides must be one.");
+
+  // Validate that tma_domain is a superset of tma_ids, otherwise there is
+  // something wrong in the schedule.
   std::unordered_set<IterDomain*> seen;
-  for (auto tuple : frontier) {
+  std::unordered_set<Val*> pending_tma_ids(tma_ids.begin(), tma_ids.end());
```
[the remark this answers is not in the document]
zasdfgbnm: fixed

csrc/index_compute.cpp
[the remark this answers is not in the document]
zasdfgbnm: changed

csrc/index_compute.cpp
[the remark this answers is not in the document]
naoyam: I see. Makes sense, but does the split of `I3` matter?

csrc/index_compute.cpp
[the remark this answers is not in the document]
naoyam: Oh, I see, so it's for an implicit size-one box.

csrc/index_compute.cpp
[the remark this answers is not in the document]
naoyam: OK, I see. Thanks!

naoyam (approved):
LGTM. Thanks for explaining all the details!

PR #1993, opened 2024-03-25
Evaluate `Matmul+Bias`

Priya2698:
Adds ATen evaluation for Matmul and Matmul + Bias. Based on PR #1921, when evaluating a `castOp`, we _look back_ to see if there is a preceding MmaOp and evaluate them together.

Issue #1775.

### DIFF
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -398,8 +398,18 @@ std::vector<PolymorphicValue> UnaryOp::evaluate(
   // Broadcast(bias)) -> Cast) If not, evaluate UnaryOp::CastOp along with the
   // other types by evaluating the immediate input.
 
-  if ((getUnaryOpType() == UnaryOpType::Cast) &&
-      input(0)->definition() != nullptr) {
+  // Check if the unary op is a cast from fp32 to lower precision.
+  auto is_downcast = [this]() -> bool {
+    if (getUnaryOpType() != UnaryOpType::Cast) {
+      return false;
+    }
+    auto in_dtype = input(0)->getDataType().value();
+    return (
+        in_dtype == DataType::Float &&
+        isInclusiveType(*(out()->getDataType()), in_dtype));
+  };
+
+  if (is_downcast() && input(0)->definition() != nullptr) {
     Val* mma_lhs = nullptr;
     Val* mma_rhs = nullptr;
 
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -392,12 +392,17 @@ std::vector<PolymorphicValue> UnaryOp::evaluate(
     std::unordered_map<const Val*, PolymorphicValue>& known_values) const {
   using namespace PolymorphicValue_functions;
 
-  // If the UnaryOp is CastOp, check if the preceding pattern of 
-  // operators matches with matmul (MmaOp(Broadcast (A), Broadcast(B)) -> Cast) or matmul + bias 
-  // (BinaryOp::Add (MmaOp(Broadcast (A), Broadcast(B), Broadcast(bias)) -> Cast)
-  // If not, evaluate UnaryOp::CastOp along with the other types by evaluating the immediate input. 
+  // If the UnaryOp is CastOp, check if the preceding pattern of
+  / operators matches with matmul (MmaOp(Broadcast (A), Broadcast(B)) -> Cast) 
+  // r matmul + bias
+ inaryOp::Add (MmaOp(Broadcast (A), Broadcast(B), B
+  // oadcast(bias)) -> Cast)
+   not, evaluate UnaryOp::CastOp along with the o
+  // her types by evaluating the immediate input.
 
-  auto has_matmul_and_bias = [](Val* in) -> bool {
+
+
+  o has_matmul_and_bias = [](Val* in) -> bool {
     if (auto* binary = dynamic_cast<BinaryOp*>(in->definition())) {
       if (binary->getBinaryOpType() == BinaryOpType::Add) {
         if (binary->input(0)->definition()->isA<MmaOp>()) {
@@ -408,18 +413,24 @@ std::vector<PolymorphicValue> UnaryOp::evaluate(
     return false;
   };
 
-  if ((getUnaryOpType() == UnaryOpType::Cast) && input(0)->definition() != nullptr){ 
-    // Case 1: MmaOp + Cast (Matmul)
+  if ((getUnaryOpType() == UnaryOpType::Cast) && in
+      ut(0)->definition() != nullptr){
+   
+    ase 1: MmaOp + Cast (Matmul)
     if (auto* mma = dynamic_cast<MmaOp*>(input(0)->definition())) {
       MmaOpUtils::verifyMmaOpForEvaluation(mma, out()->getDataType().value());
-      
-      std::vector<at::Tensor> mma_inputs;
-      for (Val* inp: mma->inputs()) {
+
+  
+
+      r<at::Tensor> mma_inputs;
+      for (Val* inp: mma->inp uts()) {
         auto eval_i = ee.evaluate(inp, known_values);
         mma_inputs.push_back(eval_i.as<at::Tensor>());
       }
-      
-      const auto a = mma_inputs.at(0).squeeze(-1);
+
+      co
+
+      ma_inputs.at(0).squeeze(-1);
       const auto b = mma_inputs.at(1).squeeze(0);
 
       // After removing the broadcast dimensions, the format should be
@@ -433,29 +444,38 @@ std::vector<PolymorphicValue> UnaryOp::evaluate(
       MmaOp* mma = binary->input(0)->definition()->as<MmaOp>();
 
       MmaOpUtils::verifyMmaOpForEvaluation(mma, out()->getDataType().value());
-      MmaOpUtils::verifyBiasForEvaluation(binary->input(1), out()->getDataType().value());
-      
-      // BinaryOp <- Broadcast <- CastOp <- Bias
-      const Val* bias = binary->input(1)->definition()->input(0)->definition()->input(0);
-      const auto bias_tensor = ee.evaluate(bias, known_values).as<at::Tensor>().unsqueeze(-1);
+      MmaOpUtils::verifyBiasForEvaluation(binary->input(1)
+          , out()->getDataType().value());
+
+      // Binar
+
+      st <- CastOp <- Bias
+      const Val* bias = binary->input(1)->def
+          nition()->input(0)->definition()->input(0);
+      const auto bias_tensor = ee.evaluate(bias, kno
+          n_values).as<at::Tensor>().unsqueeze(-1);
 
       std::vector<at::Tensor> mma_inputs;
       for (Val* inp: mma->inputs()) {
-        auto eval_i = ee.evaluate(inp, known_values);
+         auto eval_i = ee.evaluate(inp, known_values);
         mma_inputs.push_back(eval_i.as<at::Tensor>());
       }
-      
-      const auto a = mma_inputs.at(0).squeeze(-1);
+
+      const auto a =
+
+      (0).squeeze(-1);
       const auto b = mma_inputs.at(1).squeeze(0);
 
       return {at::addmm(bias_tensor, a, b)};
     }
   }
 
-  const auto& in = ee.evaluate(inputs().at(0), known_values);
+  // If there is not a preceding MmaOp, evaluate immediate inputs and compute the output for unary ops.
+ 
+  // const auto& in = ee.evaluate(inputs().at(0), known_values);
   if (!in.hasValue()) {
-      return {std::monostate{}};
-  }
+      return {std::monostat
+     }
 
   switch (getUnaryOpType()) {
     case UnaryOpType::Neg:
@@ -4470,3 +4490,4 @@ std::vector<PolymorphicValue> CatOp::evaluate(
 }
 
 } // namespace nvfuser
+                            
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -398,61 +398,34 @@ std::vector<PolymorphicValue> UnaryOp::evaluate(
   // Broadcast(bias)) -> Cast) If not, evaluate UnaryOp::CastOp along with the
   // other types by evaluating the immediate input.
 
-  auto has_matmul_and_bias = [](Val* in) -> bool {
-    if (auto* binary = dynamic_cast<BinaryOp*>(in->definition())) {
-      if (binary->getBinaryOpType() == BinaryOpType::Add) {
-        if (binary->input(0)->definition()->isA<MmaOp>()) {
-          return true;
-        }
-      }
-    }
-    return false;
-  };
-
   if ((getUnaryOpType() == UnaryOpType::Cast) &&
       input(0)->definition() != nullptr) {
+    Val* mma_lhs = nullptr;
+    Val* mma_rhs = nullptr;
+
     // Case 1: MmaOp + Cast (Matmul)
-    if (auto* mma = dynamic_cast<MmaOp*>(input(0)->definition())) {
-      MmaOpUtils::verifyMmaOpForEvaluation(mma, out()->getDataType().value());
-
-      std::vector<at::Tensor> mma_inputs;
-      for (Val* inp : mma->inputs()) {
-        auto eval_i = ee.evaluate(inp, known_values);
-        mma_inputs.push_back(eval_i.as<at::Tensor>());
-      }
-
-      const auto a = mma_inputs.at(0).squeeze(-1);
-      const auto b = mma_inputs.at(1).squeeze(0);
-
-      // After removing the broadcast dimensions, the format should be
-      // [M, K] x [K, N] compatible with aten::matmul format.
+    if (MmaOpUtils::matchMatmulCast(
+            this, std::ref(mma_lhs), std::ref(mma_rhs))) {
+      // Inputs to MmaOp are of the shape [M, K, 1] x [1, K, N]
+      const auto a =
+          ee.evaluate(mma_lhs, known_values).as<at::Tensor>().squeeze(-1);
+      const auto b =
+          ee.evaluate(mma_rhs, known_values).as<at::Tensor>().squeeze(0);
       return {a.matmul(b)};
     }
 
+    Val* bias = nullptr;
     // Case 2: Matmul + Bias
-    if (has_matmul_and_bias(input(0))) {
-      BinaryOp* binary = input(0)->definition()->as<BinaryOp>();
-      MmaOp* mma = binary->input(0)->definition()->as<MmaOp>();
-
-      MmaOpUtils::verifyMmaOpForEvaluation(mma, out()->getDataType().value());
-      MmaOpUtils::verifyBiasForEvaluation(
-          binary->input(1), out()->getDataType().value());
-
-      // BinaryOp <- Broadcast <- CastOp <- Bias
-      const Val* bias =
-          binary->input(1)->definition()->input(0)->definition()->input(0);
+    if (MmaOpUtils::matchMatmulBiasCast(
+            this, std::ref(mma_lhs), std::ref(mma_rhs), std::ref(bias))) {
+      // Inputs to MmaOp are of the shape [M, K, 1] x [1, K, N]
+      const auto a =
+          ee.evaluate(mma_lhs, known_values).as<at::Tensor>().squeeze(-1);
+      const auto b =
+          ee.evaluate(mma_rhs, known_values).as<at::Tensor>().squeeze(0);
+      // Bias is of shape [M,].
       const auto bias_tensor =
           ee.evaluate(bias, known_values).as<at::Tensor>().unsqueeze(-1);
-
-      std::vector<at::Tensor> mma_inputs;
-      for (Val* inp : mma->inputs()) {
-        auto eval_i = ee.evaluate(inp, known_values);
-        mma_inputs.push_back(eval_i.as<at::Tensor>());
-      }
-
-      const auto a = mma_inputs.at(0).squeeze(-1);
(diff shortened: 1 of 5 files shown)


Priya2698:
!build

Priya2698:
Tests (apart from known issues in the CI) pass here: https://gitlab-master.nvidia.com/dl/pytorch/fuser-gh-mirror/-/pipelines/13764968

csrc/ir/nodes.cpp
```diff
@@ -389,9 +389,81 @@ UnaryOp::UnaryOp(IrBuilderPasskey passkey, UnaryOpType type, Val* out, Val* in)
...
+      BinaryOp* binary = input(0)->definition()->as<BinaryOp>();
+      MmaOp* mma = binary->input(0)->definition()->as<MmaOp>();
+
+      MmaOpUtils::verifyMmaOpForEvaluation(mma, out()->getDataType().value());
+      MmaOpUtils::verifyBiasForEvaluation(
+          binary->input(1), out()->getDataType().value());
+
+      // BinaryOp <- Broadcast <- CastOp <- Bias
+      const Val* bias =
+          binary->input(1)->definition()->input(0)->definition()->input(0);
```
wujingyue, line 443: Is this safe? has_matmul_and_bias doesn't seem to check the bias side.

csrc/ir/nodes.cpp
```diff
@@ -389,9 +389,81 @@ UnaryOp::UnaryOp(IrBuilderPasskey passkey, UnaryOpType type, Val* out, Val* in)
...
+        if (binary->input(0)->definition()->isA<MmaOp>()) {
+          return true;
+        }
+      }
+    }
+    return false;
+  };
+
+  if ((getUnaryOpType() == UnaryOpType::Cast) &&
+      input(0)->definition() != nullptr) {
```
jacobhinkle, line 413: We could also check that the cast dtype is reduced precision and the input dtype is Float, to reduce pattern matching.

csrc/ir/nodes.cpp
```diff
@@ -389,9 +389,81 @@ UnaryOp::UnaryOp(IrBuilderPasskey passkey, UnaryOpType type, Val* out, Val* in)
...
+      const auto a = mma_inputs.at(0).squeeze(-1);
+      const auto b = mma_inputs.at(1).squeeze(0);
+
+      // After removing the broadcast dimensions, the format should be
+      // [M, K] x [K, N] compatible with aten::matmul format.
+      return {a.matmul(b)};
+    }
+
+    // Case 2: Matmul + Bias
+    if (has_matmul_and_bias(input(0))) {
```
wujingyue, line 433: In this version, the cast+broadcast+add+matmul+cast pattern we are looking for is checked in has_matmul_and_bias and is assumed outside (line 434 and 435). This creates room for error in the future when someone changes one place but forgets to change the other. I think it's going to be cleaner and more robust to combine matching the pattern and assigning the matched results into one function. This is remotely analogous to `re.match`ing a pattern with groups -- `re.match` both matches the pattern and associates the matched results with the group ID. 

```
Val* mma_lhs = nullptr;
Val* mma_rhs = nullptr;
if (matchMatmulCast(this, std::ref(mma_lhs), std::ref(mma_rhs))) {
  ...
  return ...;
}

Val* bias = nullptr;
if (matchMatmulBiasCast(this, std::ref(mma_lhs), std::ref(mma_rhs), std::ref(bias))) {
  ...
  return ...;
}

Evaluate the immediate inputs and this op. 
```

Yes, the root CastOp will be checked twice, but at this moment I'd prefer simplicity to performance unless we prove that the extra check is significant overhead (I don't think it is).

csrc/ir/nodes.cpp
jacobhinkle, line 413: We should probably use NVF_ERROR to check that definition is not null instead of skipping this branch in that unexpected case.

csrc/ir/nodes.cpp
Priya2698:
```
 MmaOpUtils::verifyMmaOpForEvaluation(mma, out()->getDataType().value());
      MmaOpUtils::verifyBiasForEvaluation(
          binary->input(1), out()->getDataType().value());
```

checks that the necessary conditions for the MmaOp and Bias are true.

csrc/ir/utils.cpp
```diff
@@ -1310,4 +1310,102 @@ MmaOpDetails getMmaOpDetails(
...
+    if (auto* unary = dynamic_cast<UnaryOp*>(in->definition())) {
+      if (unary->getUnaryOpType() == UnaryOpType::Cast) {
+        return unary->input(0)->getDataType().value() == expected_dtype;
+      }
+    }
+    return false;
+  };
+
+  NVF_CHECK(is_previous_castop(bcast_in, final_out_dtype),
+    "Expected the bias to be preceded by castOp before broadcasting and to be originally of the same type as the final output dtype .");
```
jacobhinkle, line 1409: Does addmm support different bias dtype? If so the can we make this a return value instead of an assertion, then handle this case gracefully?

csrc/ir/nodes.cpp
wujingyue: It may simplify to merge (some of) `verify*` into `match*` as well.

csrc/ir/nodes.cpp
```diff
@@ -389,9 +389,81 @@ UnaryOp::UnaryOp(IrBuilderPasskey passkey, UnaryOpType type, Val* out, Val* in)
...
+      MmaOp* mma = binary->input(0)->definition()->as<MmaOp>();
+
+      MmaOpUtils::verifyMmaOpForEvaluation(mma, out()->getDataType().value());
+      MmaOpUtils::verifyBiasForEvaluation(
+          binary->input(1), out()->getDataType().value());
+
+      // BinaryOp <- Broadcast <- CastOp <- Bias
+      const Val* bias =
+          binary->input(1)->definition()->input(0)->definition()->input(0);
+      const auto bias_tensor =
```
jacobhinkle, line 444: Do you think you could go one step further here and try to match the beta parameter?

csrc/ir/nodes.cpp
Priya2698: That is a valid case. For eg: a standalone castOp. In that case, it will be evaluated regularly with the other UnaryOp. Why do you think this case should not occur?

csrc/ir/utils.cpp
Priya2698: No. All the inputs have to be of the same type.

csrc/ir/nodes.cpp
Priya2698: Sure. Will add that too. I was thinking of doing it in a separate PR but can add that here as well.

csrc/ir/nodes.cpp
Priya2698: I am somewhat unclear on your concern. We do check that the bias side satisfies all the required conditions in `verifyBiasForEvaluation` similar to how the required assumptions for MmaOp are enforced in `verifyMmaForEvaluation.`

Are you suggesting merging the `has_matmul_and_bias` with assignment of bias? Or the verification should be done here as well instead of another call?

wujingyue: > Are you suggesting merging the has_matmul_and_bias with assignment of bias? Or the verification should be done here as well instead of another call?

Both. Let me clarify a bit more: I'd love to see everything about that pattern (including matching the pattern, retrieving the matched results, and probably verification about that pattern) to be localized, so it's harder to make oops-I-forgot-to-also-change-this type of programming errors and easier to read the code without jumping in and out of the `has_matmul_and_bias` function. 

I'm happy to chat more about it after 2:30 today!

Priya2698: Okay, I think I get your point. I still prefer keeping the verification of `MmaOp` modular since that will be reused in all patterns.
Let me refactor this, and we can discuss more.

wujingyue: > the verification of MmaOp modular since that will be reused in all patterns.

That makes sense. I guess we can call that function from both matchMatmulCast and matchMatmulBiasCast. Anyhow, I'm looking forward to the next version so the discussion is more concrete!

csrc/ir/nodes.cpp
jacobhinkle: Oh you're right. Sorry I was just confused. Thanks for the explanation.

csrc/ir/nodes.cpp
jacobhinkle: Yeah you can do it in another PR. Best to keep the PRs small if possible.

csrc/ir/utils.cpp
```diff
@@ -1310,4 +1310,111 @@ MmaOpDetails getMmaOpDetails(
...
+        Val* upcast_bias =
+            bcast_bias->definition()->as<BroadcastOp>()->input(0);
+
+        // The bias tensor and matmul inputs should be of the same dtype.
+        // The bias tensor is upcasted to fp32.
+        if (auto* unary = dynamic_cast<UnaryOp*>(upcast_bias->definition())) {
+          if (unary->getUnaryOpType() == UnaryOpType::Cast) {
+            NVF_CHECK(
+                unary->input(0)->getDataType().value() == final_out_dtype,
+                "Expected the bias to be originally of the same type as the final output dtype.");
```
wujingyue, line 1406: Is an `if` any safer? Nothing prevents the input fusion from having a different cast type. When that happens, is it better to hard fail or bail out? 

A similar comment goes to line 1413.

csrc/ir/utils.cpp
```diff
@@ -1310,4 +1310,111 @@ MmaOpDetails getMmaOpDetails(
...
+// 1. MmaOp is preceded by a broadcast.
+// 2. The inputs to MmaOp are broadcasted as the last dim for the first operand
+// and the first dim for the second operand. The inputs of MmaOp will be [M, K,
+// 1] x [1, K, N].
+void verifyMmaOpForEvaluation(MmaOp* mma_op, DataType final_out_dtype) {
+  const Val* in_a = mma_op->inA();
+  const Val* in_b = mma_op->inB();
+
+  NVF_ERROR(
+      in_a->as<TensorView>()->nDims() == in_b->as<TensorView>()->nDims(),
```
wujingyue, line 1323: `in_a/b->as<TensorView>` is used several times in this function. How about "caching" them (e.g. `in_a/b_tv = in_a/b->as<TensorView>`) before this line?

csrc/ir/utils.cpp
Priya2698: I am not clear. Are you saying we should not hard fail here?
At this point, we have discovered a bias and MmaOp so the bias must fullfill all the conditions we expect under the current matmul structure.

Priya2698: If there is not a castOp preceding the broadcasted bias, that implies the bias and the MmaOp inputs are of different types which is not supported in at::addmm, so we should hard fail. That implies we have an unexpected case / matmul fusion assumptions have changed and these need to be updated.

wujingyue: > Expected the bias to be originally of the same type as the final output dtype.

Let me ask this in a different way: when you say "expect" here, which of the following do you mean
1. If the bias is not of the same type, it's a programming error in nvFuser and we should fix that programming error.
2. If the bias is not of the same time, it's still a valid input fusion (e.g. can be constructed using fd.ops) and we should evaluate these ops in a different way (e.g. evaluate them one at a time). 
3. If the bias is not of the same time, it's still a valid input fusion but we don't have an alternative to evaluate them, so let's hard fail. 

I thought it was 2 but after a second thought it is likely 3.

Priya2698: I mean 3. similar to the case in `matmul + cast`. There is no corresponding ATen call.

Priya2698: I can rephrase the error message.

csrc/ir/utils.cpp
```diff
@@ -1310,4 +1310,111 @@ MmaOpDetails getMmaOpDetails(
...
+  // ATen matmul should be the same as MmaOp out `eventually`.
+  // See https://github.com/NVIDIA/Fuser/pull/1874#discussion_r1516991574
+  // Supported cases:
+  //  1. MmaOp->out() and MmaOp->input() are the same dtype.
+  //  2. MmaOp->out() is followed by a CastOp() to the MmaOp->input() dtype.
+  // NOTE: Currently MmaOp only accepts Half and BFloat16 so case (1) does not
+  // occur.
+
+  NVF_ERROR(
+      in_a->as<TensorView>()->getDataType().value() == final_out_dtype,
```
wujingyue:
```suggestion
      *(in_a_tv->getDataType()) == final_out_dtype,
```
less typing without being ambiguous.

wujingyue: Should we check in_b's dtype as well?

csrc/ir/utils.cpp
```diff
@@ -1310,4 +1310,111 @@ MmaOpDetails getMmaOpDetails(
   return details;
 }
 
+// Verifies the assumptions made when evaluating a fusion containing MmaOp:
+// 1. MmaOp is preceded by a broadcast.
+// 2. The inputs to MmaOp are broadcasted as the last dim for the first operand
+// and the first dim for the second operand. The inputs of MmaOp will be [M, K,
+// 1] x [1, K, N].
+void verifyMmaOpForEvaluation(MmaOp* mma_op, DataType final_out_dtype) {
```
wujingyue, line 1318:
```suggestion
void verifyMmaOpForEvaluation(MmaOp* mma_op, const DataType expected_in_dtype) {
```

I understand where `final_out_dtype` comes from, but I think `expected_in_dtype` is a better name for the context of this function.

csrc/ir/utils.cpp
```diff
@@ -1310,4 +1310,111 @@ MmaOpDetails getMmaOpDetails(
   return details;
 }
 
+// Verifies the assumptions made when evaluating a fusion containing MmaOp:
+// 1. MmaOp is preceded by a broadcast.
+// 2. The inputs to MmaOp are broadcasted as the last dim for the first operand
+// and the first dim for the second operand. The inputs of MmaOp will be [M, K,
+// 1] x [1, K, N].
```
wujingyue, line 1317: Can you add a comment regarding `final_out_dtype` or whatever you like to rename it to?

csrc/ir/utils.h
```diff
@@ -58,7 +58,9 @@ MmaOpDetails getMmaOpDetails(
     TensorView* in_a,
     TensorView* in_b);
 
-void verifyMmaOpForEvaluation(MmaOp* mma_op, DataType castop_out_dtype);
+void verifyMmaOpForEvaluation(
+    MmaOp* mma_op,
+    const DataType expected_input_dtype);
```
wujingyue, line 63:
```suggestion
    DataType expected_input_dtype);
```

https://abseil.io/tips/109

csrc/ir/utils.cpp
```diff
@@ -1310,4 +1310,121 @@ MmaOpDetails getMmaOpDetails(
   return details;
 }
 
+// Verifies the assumptions made when evaluating a fusion containing MmaOp:
+// 1. MmaOp is preceded by a broadcast.
+// 2. The inputs to MmaOp are broadcasted as the last dim for the first operand
+// and the first dim for the second operand.
+// The inputs of MmaOp will be [M, K, 1] x [1, K, N].
+// Additionally, the inputs to the MmaOp should of the `expected_input_dtype`.
```
wujingyue, line 1318: For posterity, can you explain why this is required?

wujingyue, line 1318:
```suggestion
// Additionally, the inputs to the MmaOp should be of `expected_input_dtype`.
```

csrc/ir/utils.cpp
```diff
@@ -1310,4 +1310,121 @@ MmaOpDetails getMmaOpDetails(
...
+}
+
+bool matchMatmulBiasCast(
+    const UnaryOp* cast_op,
+    Val*& mma_lhs,
+    Val*& mma_rhs,
+    Val*& bias) {
+  if (auto* binary = dynamic_cast<BinaryOp*>(cast_op->input(0)->definition())) {
+    if (binary->getBinaryOpType() == BinaryOpType::Add) {
+      if (auto* mma = dynamic_cast<MmaOp*>(binary->input(0)->definition())) {
```
wujingyue, line 1387:
```
auto* binary = ...
if (binary == nullptr) {
  return false;
}
if (binary is not add) {
  return false;
}

auto* mma = ...;
if (mma == nullptr) {
  return false;
}
```

Human brains are bad at reasoning about branches, so the earlier we exit (even for some code paths) the better. This also helps to reduce indentation.

csrc/ir/utils.cpp
```diff
@@ -1310,4 +1310,121 @@ MmaOpDetails getMmaOpDetails(
...
+
+        // Verify assumptions for MmaOp hold. Assign the values to Mma operands.
+        MmaOpUtils::verifyMmaOpForEvaluation(mma, final_out_dtype);
+        mma_lhs = mma->inA();
+        mma_rhs = mma->inB();
+
+        Val* bcast_bias = binary->input(1);
+        // Bias is casted to fp32 and broadcasted from shape [M,] to [M, 1] in
+        // biasEpilogue.
+        NVF_ERROR(
```
wujingyue, line 1398: I'm not sure about NVF_CHECK vs NVF_ERROR. But this check and several others down the road seems to be of the same nature as those in verifyMmaOpForEvaluation. Should all of them use the same macro?

csrc/ir/utils.cpp
```diff
@@ -1310,4 +1310,121 @@ MmaOpDetails getMmaOpDetails(
...
+    if (binary->getBinaryOpType() == BinaryOpType::Add) {
+      if (auto* mma = dynamic_cast<MmaOp*>(binary->input(0)->definition())) {
+        DataType final_out_dtype = cast_op->out()->getDataType().value();
+
+        // Verify assumptions for MmaOp hold. Assign the values to Mma operands.
+        MmaOpUtils::verifyMmaOpForEvaluation(mma, final_out_dtype);
+        mma_lhs = mma->inA();
+        mma_rhs = mma->inB();
+
+        Val* bcast_bias = binary->input(1);
```
wujingyue, line 1395: I think the code can simplify using dynamic_cast and reusing the same variable `bias`. The latter is a bit subjective because one could argue for SSA, but I tend to trade that for having to come up with different names for bias at different stages. 

For example,
```
bias = binary->input(1);
auto* bcast = dynamic_cast<BroadCastOp*>(bias->definition());  // btw, dynamic_cast a nullptr returns a nullptr. 
NVF_ERROR(bcast != nullptr, ...);
Check the last dimension is broadcasted;

bias = bcast->input(0);
auto* cast = dynamic_cast<UnaryOp*>(bias->definition());
check cast is not nullptr;
check cast is indeed of cast. 

bias = cast->input(0);

return true;
```
A slight downside is that bias may be non-null when match returns false, but callers tend to not read bias at all when the matching failed, so it's fine. If you really want to, you can add a comment saying these output references may be changed even when the matching failed.

wujingyue (commented):
LGTM with comments! Make sure you address Jacob's as well.

csrc/ir/utils.cpp
Priya2698: I am using NVF_ERROR for all. Will make sure to replace any NVF_CHECK with NVF_ERROR. My reasoning, is that this is a case that should not have occured under our current matmul assumptions.

csrc/ir/utils.cpp
Priya2698: There is a comment on line 1351 reasoning about the dtypes. Do you want to move it upwards?

csrc/ir/nodes.cpp
Priya2698: Is there any other way to reason about reduced precision, other than explicitly checking that the final types are fp16 / bf16?

Priya2698: `isInclusiveType` should work.

csrc/ir/utils.cpp
```diff
@@ -1310,4 +1310,116 @@ MmaOpDetails getMmaOpDetails(
...
+
+bool matchMatmulCast(const UnaryOp* cast_op, Val*& mma_lhs, Val*& mma_rhs) {
+  if (auto* mma = dynamic_cast<MmaOp*>(cast_op->input(0)->definition())) {
+    MmaOpUtils::verifyMmaOpForEvaluation(
+        mma, cast_op->out()->getDataType().value());
+    mma_lhs = mma->inA();
+    mma_rhs = mma->inB();
+    return true;
+  }
+  return false;
```
wujingyue, line 1377: Let this code path exit early. 

```
auto* mma = ...;
if (mma == nullptr) {
  return false;
}
the rest;
```

wujingyue (approved):

Priya2698:
!build

Priya2698:
!build

csrc/ir/nodes.cpp
```diff
@@ -389,9 +389,64 @@ UnaryOp::UnaryOp(IrBuilderPasskey passkey, UnaryOpType type, Val* out, Val* in)
...
+      const auto a =
+          ee.evaluate(mma_lhs, known_values).as<at::Tensor>().squeeze(-1);
+      const auto b =
+          ee.evaluate(mma_rhs, known_values).as<at::Tensor>().squeeze(0);
+      return {a.matmul(b)};
+    }
+
+    Val* bias = nullptr;
+    // Case 2: Matmul + Bias
+    if (MmaOpUtils::matchMatmulBiasCast(
```
protonu, line 429: I am not sure how to address this here, but what worries me is torch.nn.functional.linear takes x.A(T)+bias. So unless we transpose A, we won't have [M,K,1] [1,N,K]. 

https://pytorch.org/docs/stable/generated/torch.nn.functional.linear.html

We may need to write/use a routine to detect the shape.

Priya2698: I'll address this is in the next PR. The assumed shapes are the same as that in matmul, so I was assuming that the transpose would already have been done when needed. We can discuss more today.

PR #1995, opened 2024-03-25
Add kir::Allocate::resetsToZero()

jacobhinkle:
As suggested by @naoyam, this will allow us to incrementally enable reuse of zeroed memory (see #1984) on a case-by-case basis. To do so, when creating an `Allocate` node, if it is guaranteed to be "cleaned up" (i.e. reset to all zero values) by the end of the kernel, then you can pass `true` for the `resets_to_zero` argument in the `kir::Allocate` constructor. This will automatically enable reuse of zeroed memory for this buffer and avoid a memset call, regardless of whether
`EnableOption::ReuseZeroedMemory` is enabled or not. Eventually this will let us get rid of that option entirely, once all of our global memory semaphores are guaranteed to be reset.

### DIFF
--- a/csrc/device_lower/pass/grid_serialization.cpp
+++ b/csrc/device_lower/pass/grid_serialization.cpp
@@ -127,7 +127,8 @@ class GridSerializationSyncInserter : kir::ExprMutator {
     kir::Allocate* alloc = lower_utils::allocGlobalBufferForGridComm(
         lower_utils::getGridSyncBufferSize(cur_expr_sync_pattern_.value()),
         DataType::Int,
-        true);
+        /*zero_init=*/true,
+        /*resets_to_zero=*/true);
     auto wait = IrBuilder::create<kir::BlockSerializeWait>(
         cur_expr_sync_pattern_.value(), alloc->buffer());
     registerInsertBefore(cur_top_level_expr_, alloc);
--- a/tests/cpp/test_serial_gridreduce.cpp
+++ b/tests/cpp/test_serial_gridreduce.cpp
@@ -37,110 +37,91 @@ using SerialGridReductionTest = NVFuserTest;
 
 TEST_F(SerialGridReductionTest, Scheduling) {
   for (bool serial : {true, false}) {
-    // For serial grid reduction we test that re-using semaphore buffers works
-    // properly. This will cause a failure if serial grid reduction does not
-    // properly clean up its semaphores.
-    std::vector<bool> reuse_zeroed_memory_values{false};
-    if (serial) {
-      reuse_zeroed_memory_values.push_back(true);
-    }
-    for (bool reuse_zeroed_memory : reuse_zeroed_memory_values) {
-      EnableOptionsGuard eog;
-      if (reuse_zeroed_memory) {
-        EnableOptionsGuard::getCurOptions().set(
-            EnableOption::ReuseZeroedMemory);
-      } else {
-        EnableOptionsGuard::getCurOptions().unset(
-            EnableOption::ReuseZeroedMemory);
-      }
-      for (int64_t num_warps : {4, 8}) {
-        // B is size of inner serial loop. Outer loop is hardcoded at A=4
-        // Here we set B to a small value of 8 instead of 32 (i.e. 128 elements
-        // per thread), so that the non-serial compilation does not take too
-        // long.
-        for (int64_t B : {8}) {
-          std::unique_ptr<Fusion> fusion_ptr = std::make_unique<Fusion>();
-          auto fusion = fusion_ptr.get();
-          FusionGuard fg(fusion);
+    for (int64_t num_warps : {4, 8}) {
+      // B is size of inner serial loop. Outer loop is hardcoded at A=4
+      // Here we set B to a small value of 8 instead of 32 (i.e. 128 elements
+      // per thread), so that the non-serial compilation does not take too
+      // long.
+      for (int64_t B : {8}) {
+        std::unique_ptr<Fusion> fusion_ptr = std::make_unique<Fusion>();
+        auto fusion = fusion_ptr.get();
+        FusionGuard fg(fusion);
 
-          int64_t blocks_x = 8;
-          int64_t blocks_y = 8;
-          int64_t blocks_z = 5;
-          int64_t A = 4; // Size of outer serial loop
-          int64_t H = blocks_z;
-          int64_t W = A * B * blocks_x * blocks_y * num_warps * 32;
+        int64_t blocks_x = 8;
+        int64_t blocks_y = 8;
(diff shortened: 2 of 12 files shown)


csrc/device_lower/pass/grid_serialization.cpp
```diff
@@ -127,7 +127,8 @@ class GridSerializationSyncInserter : kir::ExprMutator {
     kir::Allocate* alloc = lower_utils::allocGlobalBufferForGridComm(
         lower_utils::getGridSyncBufferSize(cur_expr_sync_pattern_.value()),
         DataType::Int,
-        true);
+        /*zero_init=*/true,
+        /*resets_to_zero=*/true);
```
jacobhinkle, line 131: This is the idiomatic way to create a semaphore buffer that will be reused. This should only be done if we can guarantee that the buffer is clean. Note that for debugging you can enable `NVFUSER_DUMP=global_zeroed_memory` to test whether the semaphores are actually cleaned, as that will count non-zero bytes in the global zeroed memory pool.

tests/cpp/test_serial_gridreduce.cpp
```diff
@@ -37,110 +37,91 @@ using SerialGridReductionTest = NVFuserTest;
...
-    }
-    for (bool reuse_zeroed_memory : reuse_zeroed_memory_values) {
-      EnableOptionsGuard eog;
-      if (reuse_zeroed_memory) {
-        EnableOptionsGuard::getCurOptions().set(
-            EnableOption::ReuseZeroedMemory);
-      } else {
-        EnableOptionsGuard::getCurOptions().unset(
-            EnableOption::ReuseZeroedMemory);
-      }
```
jacobhinkle, line 55: This change just removes any explicit mention of zeroed memory. Using `NVFUSER_DUMP=global_zeroed_memory`, we can confirm that the zeroed memory pool was used.

jacobhinkle:
!build --diff-bench

tests/cpp/test_serial_gridreduce.cpp
jacobhinkle:
```
# ninja -C build -j10 nvfuser_tests && NVFUSER_DUMP=global_zeroed_memory build/nvfuser_tests --gtest_filter=SerialGridReductionTest.Scheduling
ninja: Entering directory `build'
ninja: no work to do.
Running main() from /opt/pytorch/nvfuser/third_party/googletest/googletest/src/gtest_main.cc
Note: Google Test filter = SerialGridReductionTest.Scheduling
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from SerialGridReductionTest
[ RUN      ] SerialGridReductionTest.Scheduling
[global zeroed memory] Resizing arena to 512 bytes
[global zeroed memory] Allocating byte range: 0 to 512 bytes
[global zeroed memory] Resetting allocated bytes to 0
[global zeroed memory] Allocating byte range: 0 to 512 bytes
[global zeroed memory] Resetting allocated bytes to 0
[global zeroed memory] Resetting allocated bytes to 0
[global zeroed memory] Resetting allocated bytes to 0
[       OK ] SerialGridReductionTest.Scheduling (7429 ms)
[----------] 1 test from SerialGridReductionTest (7429 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (7429 ms total)
[  PASSED  ] 1 test.
```

tests/cpp/test_serial_gridreduce.cpp
```diff
@@ -37,110 +37,91 @@ using SerialGridReductionTest = NVFuserTest;
...
-
-          auto input = at::randn(
-              {H, W},
-              at::TensorOptions().dtype(at::kFloat).device(at::kCUDA, 0));
-          auto outputs = fe.runFusion({input});
-
-          if (serial) {
-            testValidate(fusion, outputs, {input}, __LINE__, __FILE__);
-          }
+    for (int64_t num_warps : {4, 8}) {
```
naoyam, line 40: This test doesn't seem to add much value unless it validates no explicit zero clear is done after each call. Otherwise, it seems like just a normal grid reduction test.

tests/cpp/test_serial_gridreduce.cpp
```diff
@@ -37,110 +37,91 @@ using SerialGridReductionTest = NVFuserTest;
...
+        if (serial) {
+          tv3->definition()->as<ReductionOp>()->requestSerialGridReduction();
+        }
+        fe.compileFusion(fusion);
+
+        auto input = at::randn(
+            {H, W}, at::TensorOptions().dtype(at::kFloat).device(at::kCUDA, 0));
+        auto outputs = fe.runFusion({input});
+
+        if (serial) {
```
naoyam, line 123: Why only the serial case is validated?

jacobhinkle: You're right, we should not have this condition. That's a hold-over that I will remove.

tests/cpp/test_serial_gridreduce.cpp
jacobhinkle: You're right that this test has less value now that serial grid reduction is used in matmul split-K. We could remove it. It's difficult to test whether a memset kernel was launched. I could test that there is some zeroed memory allocated after this call, but that will only really be informative when this test is run in isolation since other tests that are run earlier could cause that allocation also.

naoyam: Maybe capturing the `global_zeroed_memory` output and doing some string match? I remember gtest has some utility function to do that.

jacobhinkle: Added

jacobhinkle:
CI failures are real and are due to the change in `kir::Allocate` constructor signature, namely when we create aliasing allocations. Fix is incoming shortly.

naoyam (approved):
LGTM

jacobhinkle:
!build --diff-bench

liqiangxl:
LGTM.
One question:
I didn't see a change in runtime functions. Does the current serial grid reduce implementation already reset the sync flags to zero?

jacobhinkle:
> Does the current serial grid reduce implementation already reset the sync flags to zero?

Yes it already does: https://github.com/NVIDIA/Fuser/blob/2e7ec1b98b1008ce383fecc052f20c65c1b8bb5e/runtime/grid_sync.cu#L212-L213.

> Can we add a comment about why we need to reset to zero and benefits compared with set to zero before kernel launch (the way it used to be).

Good idea. Added comments on the `zeroInit()` and `resetsToZero()` methods in `kir::Allocate`.

liqiangxl (approved):
Thanks for adding the code comments.

PR #1998, opened 2024-03-25
Add note about "Merge then split vs split then merge"

zasdfgbnm:
Just adding a reading material. This question comes to my mind when I was thinking through the design of TMA schedule. I don't think it is difficult for people to reach the same conclusion by themselves if they think into it, but still, I believe explicitly speaking this out could help someone understand our system better.

### DIFF
--- a/doc/reading/merge-split-vs-split-merge.md
+++ b/doc/reading/merge-split-vs-split-merge.md
@@ -1,0 +1,30 @@
+# Merge then split vs split then merge
+
+If I have a tensor `T0[I1, I2]`, are the following two schedules equivalent?
+
+- **Schedule 1:** `[I1, I2] -- split -> [I1, I2/4, 4] -- merge -> [I1*(I2/4), 4]`.
+- **Schedule 2:** `[I1, I2] -- merge -> [I1*I2] -- split -> [(I1*I2)/4, 4]`.
+
+Where the divisions above are all ceildiv. The answer is no.
+Why? Let's see an example where `I1` has extent `2`, and `I2` has extent `5`.
+
+For schedule 1, after schedule, the extents of the leaf domain will be `[2*2, 4]`.
+So for this schedule, we will be iterating the tensor as:
+
+```python
+T[0, 0], T[0, 1], T[0, 2] , T[0, 3]
+T[0, 4], T[0, 5], T[0, 6] , T[0, 7]
+T[1, 0], T[1, 1], T[1, 2] , T[1, 3]
+T[1, 4], T[1, 5], T[1, 6] , T[1, 7]
+```
+
+For schedule 2, after schedule, the extents of the leaf domain will be `[3, 4]`.
+So for this schedule, we will be iterating the tensor as:
+
+```python
+T[0, 0], T[0, 1], T[0, 2] , T[0, 3]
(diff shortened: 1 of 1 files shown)


doc/reading/merge-split-vs-split-merge.md
```diff
@@ -0,0 +1,30 @@
...
+For schedule 2, after schedule, the extents of the leaf domain will be `[3, 4]`.
+So for this schedule, we will be iterating the tensor as:
+
+```python
+T[0, 0], T[0, 1], T[0, 2] , T[0, 3]
+T[0, 4], T[1, 0], T[1, 1] , T[1, 2]
+T[1, 3], T[1, 4], T[2, 0] , T[2, 1]
+```
+
+They are clearly not equivalent.
```
zasdfgbnm, line 30: What this means for TMA is, schedule 1 should be treated as 2D TMA, but scheduled 2 should be treated as 1D TMA.

jjsjann123: any reason that the implication for TMA is not in the reading material?

jjsjann123 (approved):

doc/reading/merge-split-vs-split-merge.md
```diff
@@ -0,0 +1,30 @@
...
+- **Schedule 1:** `[I1, I2] -- split -> [I1, I2/4, 4] -- merge -> [I1*(I2/4), 4]`.
+- **Schedule 2:** `[I1, I2] -- merge -> [I1*I2] -- split -> [(I1*I2)/4, 4]`.
+
+Where the divisions above are all ceildiv. The answer is no.
+Why? Let's see an example where `I1` has extent `2`, and `I2` has extent `5`.
+
+For schedule 1, after schedule, the extents of the leaf domain will be `[2*2, 4]`.
+So for this schedule, we will be iterating the tensor as:
+
+```python
```
drzejan2, line 14: What exactly is the indexing in this example? I expected that indexing of the tensor would look like this:
```
T[0, 0], T[0, 1], T[0, 2], T[0, 3]
T[0, 4], T[1, 0], T[1, 1], T[1, 2]
T[1, 3], T[1, 4], T[2, 0], T[2, 1]
T[2, 2], T[2, 3], T[2, 4], T[3, 0]
```
where the real data is from T[0,0] to T[1,4] with `I1`'s extend being `2` and `I2`'s extend being `5`. Anything above will be ignored/handled by indexing.

zasdfgbnm: The tensor will be iterated like this:

```python
for i in range(4):
  for j in range(4):
    i1 = i / 2
    i2 = (i % 2) * 4 + j
    if i1 < 2 and i2 < 5:
      T[i1, i2]
```

so it is indeed as shown in the doc. Just out of bound accesses will be predicated away.

doc/reading/merge-split-vs-split-merge.md
zasdfgbnm: Added to note

doc/reading/merge-split-vs-split-merge.md
drzejan2: I will need to process this example, and how things are processed exactly.

Nevertheless, thanks for the explanation!

Issue #2000, opened 2024-03-26
Matmul circular buffering interferes with epilogue smem reuse


jacobhinkle:
This command often fails on A100 and H100:
 ```
$ bin/nvfuser_bench --benchmark_filter='NvFuserScheduler_Matmul_Manual/nvfuser_splitk_NT/M:1024/N:2048/K:4096/warps:8/stages:3/splitk_factor:5/smem_epilogue:1/manual_time'
...
  what():  Fusion returns wrong results! The result tensor has shape [..., 1024,2048]. Mismatch happens at region result[...,5:784,0:1544]
Exception raised from checkMatch at /opt/pytorch/nvfuser/benchmarks/cpp/matmul.cpp:104
 ```
Disabling index hoisting and expression simplification makes this go away, but I am 99% confident this is not a hoisting or simplification bug. Rather it seems to be a data race. The inputs are fixed (fixed seeds, I've verified the inputs and outputs are stable other than the mismatches). There are about 10 mismatches per million in the outputs, but when running in compute-sanitizer, or when slowed down via NVFUSER_DISABLE options, there are no mismatches. Running with
NVFUSER_DUMP=debug_info compute-sanitizer --tool racecheck indicates a data hazard in the circular buffering code that uses cp.async.

The hazard warning goes away for `smem_epilogue:0` in which case we also don't get any checkMatch errors. For `splitk_factor:1/smem_epilogue:1` we don't get checkMatch errors but the racecheck warnings are present. If I add `asm volatile("cp.async.wait_group %0;\n"::"n"(0LL));` (or `wait_all`) after the main loop and just before the smem unswizzling loop, then the race warnings and checkMatch failures go away in all cases I've tested.
 
I believe this is due to smem reuse being unaware of cp.async, and potentially due to some additional copies that go unused during circular buffering. The smem aliasing logic looks at the lifetime of the smem tensor as defined by the last read expression, and inserts a `__syncthreads()` to ensure that all threads in the block have reached that last read before allowing aliased writes to that chunk of smem. However, there could still be writes to smem from the previous lifetime when using `cp.async`. I am not sure how to programmatically check for this, and I'd also be surprised that we are doing `cp.async.cg.shared.global` transfers that go unused.

[...]

PR #2001, opened 2024-03-26
Hack to ensure cp.async is waited before smem reuse

### LINKED ISSUE #1996: Some matmul nvfuser_splitk benchmark fails
on A100 and H100

```python
root@65627d6877e4:/opt/pytorch/nvfuser# bin/nvfuser_bench --benchmark_filter='NvFuserScheduler_Matmul_Manual/nvfuser_splitk_NT/M:1024/N:2048/K:4096/warps:8/stages:3/splitk_factor:5/smem_epilogue:1/manual_time'
......
  what():  Fusion returns wrong results! The result tensor has shape [..., 1024,2048]. Mismatch happens at region result[...,526:783,0:328]
Exception raised from checkMatch at /opt/pytorch/nvfuser/benchmarks/cpp/matmul.cpp:104 (most recent call first):
frame #0: <unknown function> + 0x7e55f (0x5604dbe8b55f in bin/nvfuser_bench)
frame #1: <unknown function> + 0x6f653c (0x5604dc50353c in bin/nvfuser_bench)
frame #2: <unknown function> + 0x6fb670 (0x5604dc508670 in bin/nvfuser_bench)
frame #3: <unknown function> + 0x763ef8 (0x5604dc570ef8 in bin/nvfuser_bench)
frame #4: <unknown function> + 0x7434cb (0x5604dc5504cb in bin/nvfuser_bench)
frame #5: <unknown function> + 0x7450d5 (0x5604dc5520d5 in bin/nvfuser_bench)
frame #6: <unknown function> + 0x7456f9 (0x5604dc5526f9 in bin/nvfuser_bench)
frame #7: <unknown function> + 0x73482b (0x5604dc54182b in bin/nvfuser_bench)
frame #8: <unknown function> + 0x736c6e (0x5604dc543c6e in bin/nvfuser_bench)
frame #9: <unknown function> + 0x736f0e (0x5604dc543f0e in bin/nvfuser_bench)
frame #10: <unknown function> + 0x11acb4 (0x5604dbf27cb4 in bin/nvfuser_bench)
frame #11: <unknown function> + 0x29d90 (0x7f87a25d4d90 in /usr/lib/x86_64-linux-gnu/libc.so.6)
frame #12: __libc_start_main + 0x80 (0x7f87a25d4e40 in /usr/lib/x86_64-linux-gnu/libc.so.6)
frame #13: <unknown function> + 0x137355 (0x5604dbf44355 in bin/nvfuser_bench)

Aborted (core dumped)
```

several other nvfuser_bench tests fail with similar error messages

cc @naoyam @jacobhinkle

jacobhinkle:
This is a work-around for #2000.

It seems to address the issue in the only current use case for smem reuse: matmul with `params.use_smem_epilogue == true`. It is not ideal: for example it will insert a `cp.async.wait_all` instruction even if circular buffering is not used in the kernel.

Fixes #1996 but since this is a hack, I will not mark #2000 as fixed yet.

### DIFF
--- a/csrc/device_lower/pass/alias_memory.cpp
+++ b/csrc/device_lower/pass/alias_memory.cpp
@@ -1971,6 +1971,20 @@ class PromoteReuseSyncModifier : private kir::ExprMutator {
         debug() << "Inserting block sync before position " << position
                 << std::endl;
       }
+      {
+        // TODO: This is a temporary HACK to work around
+        // https://github.com/NVIDIA/Fuser/issues/2000
+        // Instead, we should only insert these wait statements when we detect
+        // that there are corresponding unsynced async operations involving the
+        // buffers in question. We should also update dispatch(Expr*) to check
+        // not only hasBlockSync but also check if there already exist AsyncWait
+        // expressions in the interval (or in some cases before the interval but
+        // after the last write?).
+        auto new_async_wait =
+            IrBuilder::create<kir::AsyncWait>(AsyncOpType::CpAsync);
+        registerInsertBefore(expr, new_async_wait);
+      }
+
       auto new_sync = IrBuilder::create<kir::BlockSync>();
       inserted_syncs_.insert(new_sync);
       registerInsertBefore(expr, new_sync);


jacobhinkle:
!build --diff-bench

PR #2003, opened 2024-03-26
IdModel: Step 4 of the loop promotion analysis

naoyam:
Step 4 of the loop promotion analysis. This is for fully propagating promotion mappings including partially inlined domains.


naoyam:
!build

csrc/id_model/id_model.h
```diff
@@ -192,7 +192,35 @@ class IdModel : public PolymorphicBase {
   // input is promoted, the output needs to be promoted too. If
   // there's already an equivalent expr that uses the promoted inputs,
   // create a mapping from the outputs of the IEL expr to the outputs
-  // of the equivalent expr.
+  // of the equivalent expr. When require_loop_mapped_promotion is
+  // true, the equivalent expr needs to be already loop mapped. If no
+  // such expr is found, the IEL expr is replayed iwth the promoted
```
zasdfgbnm, line 197:
```suggestion
  // such expr is found, the IEL expr is replayed with the promoted
```

csrc/id_model/id_model.h
```diff
@@ -281,4 +309,11 @@ class IdModel : public PolymorphicBase {
   std::unordered_map<ValGroup, IterDomain*> loop_promotion_map_;
 };
 
+// A utility function to update a map of ValGroups to ID from an old
+// Valgraph to a new ValGraph. The new graph must be a superset of the
+// old graph.
+std::unordered_map<ValGroup, IterDomain*> updateValGroupIdMap(
```
zasdfgbnm, line 315: Is this only used in unit tests?

csrc/id_model/id_model.cpp
```diff
@@ -925,17 +946,87 @@ Expr* findMatchingExpr(
       continue;
     }
 
+    // For the final loop promotion map, we want to find
+    // promotions within the same loop groups. Note that that's
+    // guaranteed when a new domain is replayed instead of reusing an
+    // existing domain.
+    if (require_loop_mapped_promotion) {
```
zasdfgbnm, line 953: Is there any reason why we don't just do the following?
```C++
    if (true) {
```

csrc/id_model/id_model.h
```diff
@@ -192,7 +192,35 @@ class IdModel : public PolymorphicBase {
...
+  // the former has the precedence. This is because when this function
+  // is used for step 4, the given iel_promotion_map starts as an
+  // empty map and gets populated during this propagation, so any
+  // mapping in the map is guaranteed to be the correct final mapping,
+  // whereas the loop graph may have invalid mappings for partially
+  // inlined domains.
+  void propagatePromotionsInIELGraph(
+      const ValGraph& iel_graph,
+      std::unordered_map<ValGroup, IterDomain*>& iel_promotion_map,
+      const ValGraph& loop_graph,
```
zasdfgbnm, line 217: Is this parameter needed? Can't we just call `idGraph(IdMappingMode::LOOP)` inside this function?

csrc/id_model/id_model.cpp
```diff
@@ -627,10 +627,29 @@ std::unordered_map<ValGroup, IterDomain*> IdModel::buildLoopPromotionMap(
...
+  // propagate back to the loop groups in Step 5. Unlike Step 2, the
+  // initial IEL promotion map is empty and is populated with the loop
+  // promotion map as we traverse down the IEL graph.
+  std::unordered_map<ValGroup, IterDomain*> final_iel_promotion_map;
+  propagatePromotionsInIELGraph(
+      iel_graph,
+      final_iel_promotion_map,
+      idGraph(IdMappingMode::LOOP),
+      loop_promotion_map,
+      true);
```
zasdfgbnm, line 647: Does it make sense to do the following instead?:
```C++
for (auto entry : iel_promotion_map) {
  ValGroup iel_group = entry->first;
  ValGroup loop_group = idGraph(IdMappingMode::LOOP).toGroup(iel_group->front());
  auto it = loop_promotion_map.find(loop_group);
  if (it != loop_promotion_map.end()) {
    entry->second = it->second;
  }
}
propagatePromotionsInIELGraph(iel_graph, iel_promotion_map, require_loop_mapped_promotion=true);
```
I think my biggest problem with `IdModel` is this is so complicated that I can not fit it into my mind. IIUC, changing it to the above code is equivalent to the current approach, but the mental model will be easier.

csrc/id_model/id_model.h
naoyam: ATM, yes. It will be used in the next step as well.

csrc/id_model/id_model.cpp
```diff
@@ -925,17 +946,87 @@ Expr* findMatchingExpr(
...
 
+// When propagating loop promotions from inputs to outputs of an IEL
+// expr, we can't blindly apply loop promotion when all of the input
+// domains are loop mapped with the outputs.
+//
+// i.e. if we have the inlined domains from:
+// T2[i0*i1] pa(1) = T0[i0*b1]ca(1) + T1[i0*i1]ca(1)
+// The inlined loop group would be:
+//
+// i0, i1, b1, i0*i1, b0*i1
```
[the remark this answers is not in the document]
naoyam: Thanks, you're right. The example is not correct. Let me update.

csrc/id_model/id_model.cpp
zasdfgbnm: Hmm... I understand this would avoid more replay. But wouldn't you need to pay back most of these "saved" replays in step 4 anyway?

csrc/id_model/id_model.cpp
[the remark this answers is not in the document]
zasdfgbnm: I do not have a concrete counter-example, but to me this sounds like "I found this unexpected behavior when there is this pattern. So I just do a pattern matching to special handle it." This sounds like just hiding the problem deeper, instead of solving the problem from its root. I am gaining more confidence that we should remove forwarding and find a better way to handle broadcast IDs created in the fusion.

To be clear, I am not asking any change for now. I am more talking about what we should do in the long term.

csrc/id_model/id_model.cpp
zasdfgbnm: I am feeling that this "optimization" is just like you borrowed some money in step 2, and pay them back in step 4. Not a strong ask because the cost of updating all docs/tests can be high, but I would prefer `if (true) {` here, because this simplifies our mental model about loop promotion: if you print the diff of the IEL promotion map of each step, the diff between step 3 and step 4 can be noisy due to the change in `require_loop_mapped_promotion `, removing this "optimization" could make it clearer to understand what step 4 does.

csrc/id_model/id_model.cpp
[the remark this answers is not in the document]
naoyam: Updated the comment. Please let me know if it makes sense.

I completely agree that this is one of the ugly complications of the Loop/Permissive graphs.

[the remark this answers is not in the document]
zasdfgbnm: It makes sense and very clear! Thank you for writing the explanation.

csrc/id_model/id_model.cpp
naoyam: Yeah, I agree with you. I'll create a cleanup issue for now.

naoyam: #2027

csrc/id_model/id_model.cpp
naoyam: This is actually related to https://github.com/NVIDIA/Fuser/pull/2003#discussion_r1548658814.

Using the Step 3 results needs to consider the condition checked by `hasUniqueInputLoopGraphs`, so the suggested code would result in the double propagation. We could avoid that by selectively updating `iel_promotion_map`, but that would mean we would need to look at inputs and outputs and propagate `loop_promotion_map` only to outputs in some cases. I'd say that would be almost equally complicated as the current version.

zasdfgbnm: I see, thanks for the explanation. Could you open an issue for https://github.com/NVIDIA/Fuser/pull/2003#discussion_r1548658814?

csrc/id_model/id_model.h
naoyam: Generally, yes, but may not be necessary for our use cases. Note that `loop_promotion_map` may have been built with an old version of `idGraph(IdMappingMode::LOOP)`. There's no guarantee that the current loop graph, `idGraph(IdMappingMode::LOOP)`, is the same as the graph used to build `loop_promotion_map`. Replaying domains could create a new val group, and if it's merged with an existing group, the domains in the existing group would be moved to a new group. Replay should not change the logical structure of the graph, but the pointer values of val groups could change, which means querying with `loop_promotion_map` using the current loop graph could fail if the map was originally built with an old version of the loop graph.

I'll file an issue and revisit this later, probably with the final Step 5 PR.

naoyam: #2029

tests/cpp/test_id_model.cpp
```diff
@@ -895,9 +950,26 @@ TEST_F(IdModelTest, LoopPromotion4) {
...
-      tester.s3_loop_promotion_map,
-      s3_reference_map);
+      tester.s3_loop_graph, tester.s3_loop_promotion_map, s3_reference_map);
+
+  auto id10 = getParentId(tv4->axis(0), 1);
+  ASSERT_EQ(id10->name(), 10);
+  auto id32 =
+      getValByName(ir_utils::consumerValsOf(id10), 32)->as<IterDomain>();
+  auto id33 =
+      getValByName(ir_utils::consumerValsOf(id10), 33)->as<IterDomain>();
```
zasdfgbnm, line 960: Use `getChildIdByName`

csrc/id_model/id_model.cpp
naoyam: > Could you open an issue for https://github.com/NVIDIA/Fuser/pull/2003#discussion_r1548658814?

What issue are you referring to? The broadcast forwarding?

zasdfgbnm: Yes

zasdfgbnm (approved):

csrc/id_model/id_model.cpp
naoyam: Added: #2030

naoyam:
!build

PR #2004, opened 2024-03-26
Bulk entire dim

zasdfgbnm:
This PR is stacked on https://github.com/NVIDIA/Fuser/pull/1991.

Besides https://github.com/NVIDIA/Fuser/pull/1991, another point that I feel very inconvenient when using TMA is, I am forced to do a split even if I want to treat the entire dimension as a box.

For example, if I want to load a shape `[100000000, 4]` tensor using a `[8, 4]` box, currently, I am required to schedule the tensor like `[100000000/8, 8 (bulk), 4/4, 4 (bulk)]`. This is inconvenient. This PR allows me to instead more conveniently schedule the fusion like `[100000000/8, 8 (bulk), 4 (bulk)]`.

Enabling this feature creates some complexity to the analysis in indexing. For example, if we have a contiguous tensor of shape `[1024, 4, 8, 16]`, if I schedule it like `[1024, 4, 8, 16(bulk)]`, I should be able to know that I wanted a 1D TMA. But if I instead scheduled like `[1024, 4(bulk), 8, 16(bulk)]`, I should be able to know that I need a 2D TMA. And if I scheduled it like `[1024, 4(bulk), 8(bulk), 16(bulk)]`, then I should be able to figure out that the inner 3 dims are mergable, so I will be using a 1D TMA. This requires me to extend the 2-state state machine built in https://github.com/NVIDIA/Fuser/pull/1991 into 3-state. Besides this, I also need to change the definition of `boxId` and `tmaGlobalId` to handle box without a split.

I suggest reviewing this PR hiding whitespaces:
image

### DIFF
--- a/csrc/index_compute.cpp
+++ b/csrc/index_compute.cpp
@@ -3311,6 +3311,7 @@ Val* Index::eye(
 // The gmem tensor of TMA store, or the gmem tensor of TMA load when replayed as
 // consumer, must be scheduled like this:
 //
+// Diagram 1:
 //                         ---------------------
 //                         | Allocation Domain |
 //                         ---------------------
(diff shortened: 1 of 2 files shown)


zasdfgbnm:
!build

zasdfgbnm:
merge this to https://github.com/NVIDIA/Fuser/pull/1991 and will update there.

PR #2005, opened 2024-03-26
[WIP] Peel off epilogue loop for circular buffering

jacobhinkle:
Currently, when a `TensorView` is circular buffered, we convert a loop like
```
for i in 0..N
  for j in ...
    ... = y[i, j]
```
to 
```
allocate x[S*D] // if S is the single-buffered size of the TensorView
for i in 0..D-1 // prologue
  for j in ...
    if pred:
      x[i*S + j] = y[i, j]
  cp.async.commit
cp.async.wait D-2 // Block until all but the last D-2 transactions complete
for i in 0..N // main loop
  for j in ...
    if pred:
      x[(((i+D-1)%D)*S+j)] = y[i+D-1, j]
    ... = x[((i%D)*S+j)]
    cp.async.commit
    cp.async.wait D-2
    __syncthreads()
```
As mentioned in #2000, the way we do the predicated load for circular buffering means that we actually have "in-flight" committed `cp.async` calls which will store zeros into smem. Since we allow `D-2` many of those to remain in flight after the end of this flight, they are never waited on, leading to the data race encountered in #2000.

This PR instead adds an epilogue loop, so that we now lower a circular buffered TV like this:
```
allocate x[S*D] // if S is the single-buffered size of the TensorView
for i in 0..D-1 // prologue
  for j in ...
    x[i*S + j] = y[i, j]
  cp.async.commit
cp.async.wait D-2 // Block until all but the last D-2 transactions complete
for i in 0..(N-(D-1)) // main loop
  for j in ...
    if pred:
      x[(((i+D-1)%D)*S+j)] = y[i+D-1, j]
    ... = x[((i%D)*S+j)]
    cp.async.commit
    cp.async.wait D-2
    __syncthreads()
for i in (N-(D-1))..N // epilogue
  cp.async.wait N-1-i
  for j in ...
    ... = x[((i%D)*S+j)]
```
Note that we have shortened the main loop and peeled it into the new "epilogue" loop which is identical to the main loop but without any additional async loads.

### Challenges encountered so far

This is going reasonably well, but there is a limitation I didn't notice at first: as mentioned [in the PTX docs](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#data-movement-and-conversion-instructions-cp-async-wait-group) the `N` argument in `cp.async.wait_group N` must be integer constant.

### DIFF
--- a/csrc/kernel_ir.h
+++ b/csrc/kernel_ir.h
@@ -631,7 +631,16 @@ class AsyncWait final : public Expr {
   explicit AsyncWait(
       IrBuilderPasskey passkey,
       AsyncOpType async_op_type,
-      int64_t keep_stages = 0);
+      Val* keep_stages);
+
+  explicit AsyncWait(
+      IrBuilderPasskey passkey,
+      AsyncOpType async_op_type,
+      int64_t keep_stages = 0)
+      : AsyncWait(
+            passkey,
+            async_op_type,
+            IrBuilder::create<Val>(keep_stages, DataType::Int)) {}
 
   NVFUSER_DECLARE_CLONE_AND_CREATE
 
@@ -651,8 +660,8 @@ class AsyncWait final : public Expr {
 
   //! Returns the remaining number of stages that are not synchronized
   //!  after this op.
-  int64_t keepStages() const {
-    return attribute<int64_t>(1);
+  Val* keepStages() const {
+    return attributeVal(1);
   }
 };
 
--- a/csrc/kernel_ir.h
+++ b/csrc/kernel_ir.h
@@ -636,11 +636,7 @@ class AsyncWait final : public Expr {
   explicit AsyncWait(
       IrBuilderPasskey passkey,
       AsyncOpType async_op_type,
-      int64_t keep_stages = 0)
-      : AsyncWait(
-            passkey,
-            async_op_type,
-            IrBuilder::create<Val>(keep_stages, DataType::Int)) {}
+      int64_t keep_stages = 0);
 
   NVFUSER_DECLARE_CLONE_AND_CREATE
 
--- a/csrc/device_lower/pass/double_buffer.cpp
+++ b/csrc/device_lower/pass/double_buffer.cpp
@@ -206,28 +206,21 @@ class DoubleBufferLoopCloner : public kir::IrVisitor {
         double_buffer_loop_->iter_domain());
 
     if (loop_type_ == DoubleBufferLoopStage::Prolog) {
-      std::cout << "Prolog" << std::endl;
       NVF_ERROR(start->isZeroInt());
       stop = SimplifyingIrBuilder::create<Val>(
           int64_t(stage_depth - 1), DataType::Index);
     } else if (
         loop_type_ == DoubleBufferLoopStage::Main &&
         requireEpilogue(double_buffer_load_exprs_)) {
-      std::cout << "Main" << std::endl;
       stop = IrBuilder::subExpr(
           double_buffer_loop_->stop(),
           IrBuilder::create<Val>(int64_t(stage_depth - 1), DataType::Index));
     } else if (loop_type_ == DoubleBufferLoopStage::Epilog) {
-      std::cout << "Epilog" << std::endl;
       NVF_ERROR(requireEpilogue(double_buffer_load_exprs_));
       start = IrBuilder::subExpr(
           double_buffer_loop_->stop(),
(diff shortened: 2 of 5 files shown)


csrc/kernel_ir.h
```diff
@@ -651,8 +656,8 @@ class AsyncWait final : public Expr {
 
   //! Returns the remaining number of stages that are not synchronized
   //!  after this op.
-  int64_t keepStages() const {
-    return attribute<int64_t>(1);
+  Val* keepStages() const {
+    return attributeVal(1);
```
jacobhinkle, line 660: I attempted to make `keepStages()` a `Val*`. The issue is that this must be a compile time constant argument for the inline asm instruction. I think it still might work if we are able to unroll the epilogue loop, but that might not always be preferable/acceptable. So instead, we could also have a runtime function/kir node that calls a runtime helper function that wraps `cp.async.wait_group N` for variable in and handles values up to say 5 or 6 inside a switch statement.

jacobhinkle: UPDATE: the switch statement helper approach seems to work and doesn't require unrolling the epilogue, but it means we need to set an upper limit on the number of unsynched stages. We could set that to something high like 10. We only need it to be `num_stages`, i.e. we don't need a switch statement with 10 cases if we have only 3 circular buffering stages. However, the requirement to have a constant `N` in `cp.async.wait_group`, along with the requirement for inline asm to have string literal inputs has stumped me. I tried all kinds of combinations of templates and macros but got nowhere.

jacobhinkle: If I'm understanding, you mean to use the unrolled loop index as the argument to `wait_group`. That fails since even if the loop variable is the actual argument we get
```
CUDA NVRTC compile error: __tmp_kernel_none_f0_c0_r0_g0.cu(10249): error: an asm operand must be an integral constant expression.
      asm volatile("cp.async.wait_group %0;"::"n"(ii)); 
```
That is, in PTX it is seeing this as a non-constant argument. I have tried interpolating it into that string, but for inline assembly the command must be a string literal...

jacobhinkle: @mmigdal-nv helped come up with this solution:
```c++
template <int num_stages>
__inline__ __device__ void cpAsyncPartialBarrier(int keep_stages) {
  if constexpr (num_stages < 0) {
    return;
  }
  if (keep_stages == num_stages) {
    asm volatile("cp.async.wait_group %0;"::"n"(num_stages));
  } else {
    cpAsyncPartialBarrier<num_stages - 1>(keep_stages);
  }
}

template <>
__inline__ __device__ void cpAsyncPartialBarrier<-1>(int keep_stages) {
}

...
  #pragma unroll
  for(nvfuser_index_t i13 = 12793; i13 < 12800LL; ++i13) {
    ...
    __syncthreads();
    cpAsyncPartialBarrier<8>((12800 - 2) - i13);
  }
  NVFUSER_UPDATE_MAGIC_ZERO;
```
We can replace `((12800 - 2) - i13)` with the `Val*` we have currently. The compiler will evaluate the recursive template and prune the dead branches. I think the only downside to this is that we need to unroll the new loop, will probably hurt compilation time.

zasdfgbnm: Fun C++...

```C++
template<int i>
void f() {}

int main() {
    #pragma unroll
    for (int i = 0; i < 10; i++) {
        f<i>();
    }
    return 0;
}
```

```C++
<source>: In function 'int main()':
<source>:7:13: error: no matching function for call to 'f<i>()'
    7 |         f<i>();
      |         ~~~~^~
<source>:2:6: note: candidate: 'template<int i> void f()'
    2 | void f() {}
      |      ^
<source>:2:6: note:   template argument deduction/substitution failed:
<source>:7:13: error: the value of 'i' is not usable in a constant expression
    7 |         f<i>();
      |         ~~~~^~
<source>:6:14: note: 'int i' is not const
    6 |     for (int i = 0; i < 10; i++) {
      |              ^
<source>:7:13: note: in template argument for type 'int'
    7 |         f<i>();
      |         ~~~~^~
```

jacobhinkle: BTW I looked to see what CUTLASS does. It seems they do not peel off an epilogue loop. Instead they just `wait_group 0` after the main loop. https://github.com/NVIDIA/cutlass/blob/c4e3e122e266644c61b4af33d0cc09f4c391a64b/include/cutlass/gemm/threadblock/mma_multistage.h

jacobhinkle:
Closing in favor of #2008.

PR #2008, opened 2024-03-27
Drain cp.async jobs after main loop

### LINKED ISSUE #2000: Matmul circular buffering interferes with epilogue smem reuse

jacobhinkle:
This just places a `cp.async.wait_group 0` instruction immediately after any circular buffer main loop which is the approach taken by CUTLASS for pipelining GEMMs: (see [mma_multistage.h#L664-L665](https://github.com/NVIDIA/cutlass/blob/c4e3e122e266644c61b4af33d0cc09f4c391a64b/include/cutlass/gemm/threadblock/mma_multistage.h#L664-L665)). The previous fix for #2000, #2001, is reverted.

This is an alternative to #2005.

Fixes #2000

### DIFF
--- a/csrc/device_lower/pass/alias_memory.cpp
+++ b/csrc/device_lower/pass/alias_memory.cpp
@@ -1972,20 +1972,6 @@ class PromoteReuseSyncModifier : private kir::ExprMutator {
         debug() << "Inserting block sync before position " << position
                 << std::endl;
       }
-      {
-        // TODO: This is a temporary HACK to work around
-        // https://github.com/NVIDIA/Fuser/issues/2000
-        // Instead, we should only insert these wait statements when we detect
-        // that there are corresponding unsynced async operations involving the
-        // buffers in question. We should also update dispatch(Expr*) to check
-        // not only hasBlockSync but also check if there already exist AsyncWait
-        // expressions in the interval (or in some cases before the interval but
-        // after the last write?).
-        auto new_async_wait =
-            IrBuilder::create<kir::AsyncWait>(AsyncOpType::CpAsync);
-        registerInsertBefore(expr, new_async_wait);
-      }
-
       auto new_sync = IrBuilder::create<kir::BlockSync>();
       inserted_syncs_.insert(new_sync);
       registerInsertBefore(expr, new_sync);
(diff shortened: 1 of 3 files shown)


jacobhinkle:
!build --diff-bench

jacobhinkle:
Fixed binary test. Python test is complex div correctness (unrelated). Code diffs look normal. I'm not sure why codediff 1/5 failed to upload but the artifact has a normal looking html. I'll merge once clang-build passes.

PR #2010, opened 2024-04-01
Fix diff report can't find test name

liqiangxl:
Happened at [link](https://github.com/NVIDIA/Fuser/pull/1961#issuecomment-2024154066).
when there is a time stamp before the test name, current diff report failed to capture the test name, e.g. 
`# [2024-03-27 18:40:08] tests/python/test_schedule_ops.py::TestScheduleOps::test_reduction_factor_op`
It can be solved by further search through double colon if test name is not found in the first try.

### DIFF
--- a/tools/codediff/diff_report.py
+++ b/tools/codediff/diff_report.py
@@ -387,6 +387,10 @@ class LogParserPyTest(LogParser):
 
         self.wildcard_testname_re = re.compile(r"^(?P<testname>\S+::\S+) (?P<line>.*)$")
 
+        self.extra_wildcard_testname_re = re.compile(
+            r".*?(?P<testname>\S+::\S+::\S+)\s+\[.*?\]\s+(?P<line>.*)"
+        )
+
         self.all_test_names: list[str] | None = None
 
     def parse_line(self, line):
@@ -396,7 +400,6 @@ class LogParserPyTest(LogParser):
                 # grab the test list
                 self.all_test_names = m.groups()[0].split(", ")
                 return True
-
         if self.all_test_names is not None:
             # Try to match a line like this:
             #
@@ -407,12 +410,25 @@ class LogParserPyTest(LogParser):
             # mark the beginning of a new test, and process the remainder of the
             # line as a separate line.
             testrest = line.split(maxsplit=1)
-            if len(testrest) > 0 and testrest[0] in self.all_test_names:
-                self.current_test = testrest[0]
-                if len(testrest) > 1:
-                    line = testrest[1]
+            if len(testrest) > 0:
+                if testrest[0] in self.all_test_names:
+                    self.current_test = testrest[0]
+                    if len(testrest) > 1:
+                        line = testrest[1]
+                    else:
+                        return True
                 else:
-                    return True
+                    # process cases with a time stamp, e.g.
+                    #
+                    # [2024-03-27 18:40:08] tests/python/test_schedule_ops.py::TestScheduleOps::test_reduction_factor_op
+                    #
+                    # Less reliable: match any line having at least one double colon
+                    # and interpret that as test name
+                    m = re.match(self.extra_wildcard_testname_re, line)
+                    if m is not None:
+                        d = m.groupdict()
+                        self.current_test = d["testname"]
+                        line = d["line"]
         else:
             # Less reliable: match any line having at least one double colon
             # and interpret that as test name
@@ -426,7 +442,6 @@ class LogParserPyTest(LogParser):
             self.finalize_test(True)
         elif line == "FAILED" and self.current_test is not None:
             self.finalize_test(False)
-
         if super().parse_line(line):
             return True
 
@@ -530,7 +545,6 @@ class TestRun:
             raise RuntimeError(
                 f"Input directory {self.directory} contains no file named 'stdout'"
             )
-
         if self.command_type == CommandType.GOOGLETEST:
             parser = LogParserGTest(logfile)
         elif self.command_type == CommandType.GOOGLEBENCH:
@@ -541,7 +555,6 @@ class TestRun:
             # The base class provides a parser that groups everything into a
             # single "test" called "Ungrouped Kernels"
             parser = LogParser(logfile)
-
         self.kernel_map = parser.kernel_map
 
     def find_preamble(self):


liqiangxl:
!build --diff

tools/codediff/diff_report.py
```diff
@@ -407,12 +411,25 @@ def parse_line(self, line):
...
+                    self.current_test = testrest[0]
+                    if len(testrest) > 1:
+                        line = testrest[1]
+                    else:
+                        return True
                 else:
-                    return True
+                    # process cases with a time stamp, e.g.
+                    #
+                    # [2024-03-27 18:40:08] tests/python/test_schedule_ops.py::TestScheduleOps::test_reduction_factor_op
```
jacobhinkle, line 424: In the [failing CI case](https://app.jobtread.com/jobs/22Nghu2FqYhc/documents/22NhqwKwdvZk?emailAddress=jacob.hinkle%40gmail.com&grantKey=22SQiGVCMqgdbJbmGffivGyx9T4dx7vzv5) there was a line like this:
```
[2024-03-27 18:38:15] tests/python/test_schedule_ops.py::TestScheduleOps::test_merge_op [W327 18:38:16.260557159 fusion_cache.cpp:507] Warning: You are overwriting the current user schedule for a definition! (function createUserSchedule)
```
This happened because of concurrent writes to stdout by pytest and `TORCH_WARN` in our codebase. So I think this re fix might prevent some errors, but ultimately I think we should just capture the warnings for tests that are known to provide them. To do this, we would need to replace `TORCH_WARN` and `TORCH_WARN_ONCE` with our own macro that uses `debug()` instead of `std::cerr` or `std::cout`. Then we'd want to use `capture_debug_output=True` for all python tests that we expect to have some output.

liqiangxl: Sounds good! Maybe let's do that change in another PR?

tools/codediff/diff_report.py
```diff
@@ -387,6 +387,10 @@ def compile_regex(self):
 
         self.wildcard_testname_re = re.compile(r"^(?P<testname>\S+::\S+) (?P<line>.*)$")
 
+        self.extra_wildcard_testname_re = re.compile(
+            r".*?(?P<testname>\S+::\S+::\S+)\s+\[.*?\]\s+(?P<line>.*)"
```
jacobhinkle, line 391: Do all of our tests and benchmarks use a fixture or class, so taht there are _two_ pairs of colons?

jacobhinkle (approved):
LGTM since I don't think this will break anything. Mentioned in the comments that we should work on piping `TORCH_WARN*` through to a capturable stream in anotehr PR.

jacobhinkle:
I looked at the artifacts for the currently failing codediff job (2/4). It does seem that the warnings are printed to a separate file `stderr` and that `stdout` looks mostly fine. It does include some lines like this:
```
[2024-04-01 16:34:10] tests/python/test_schedule_ops.py::TestScheduleOps::test_merge_op PRINTING: __tmp_kernel_1.cu
ptxas info    : Overriding maximum register limit 256 for '_ZN56_GLOBAL__N__00000000_17___tmp_kernel_1_cu_ffd1328e_745489nvfuser_1ENS_6TensorIfLi3ELi3EEENS0_IfLi2ELi2EEE' with  255 of maxrregcount option
ptxas info    : Compiling entry function '_ZN56_GLOBAL__N__00000000_17___tmp_kernel_1_cu_ffd1328e_745489nvfuser_1ENS_6TensorIfLi3ELi3EEENS0_IfLi2ELi2EEE' for 'sm_80'
ptxas info    : Function properties for _ZN56_GLOBAL__N__00000000_17___tmp_kernel_1_cu_ffd1328e_745489nvfuser_1ENS_6TensorIfLi3ELi3EEENS0_IfLi2ELi2EEE
ptxas info    : Used 18 registers, 4 bytes smem, 448 bytes cmem[0]
```
The `PRINTING __tmp_kernel_1.cu` part should already be handled by the existing regex, but this could be due to the added timestamp before the test name? That could've come from a changed pytest version or something.

tools/codediff/diff_report.py
jacobhinkle: Did you mean to put that first question mark there?

liqiangxl:
> I looked at the artifacts for the currently failing codediff job (2/4). It does seem that the warnings are printed to a separate file `stderr` and that `stdout` looks mostly fine. It does include some lines like this:
> 
> ```
> [2024-04-01 16:34:10] tests/python/test_schedule_ops.py::TestScheduleOps::test_merge_op PRINTING: __tmp_kernel_1.cu
> ptxas info    : Overriding maximum register limit 256 for '_ZN56_GLOBAL__N__00000000_17___tmp_kernel_1_cu_ffd1328e_745489nvfuser_1ENS_6TensorIfLi3ELi3EEENS0_IfLi2ELi2EEE' with  255 of maxrregcount option
> ptxas info    : 3 bytes gmem, 24 bytes cmem[4]
> ptxas info    : Compiling entry function '_ZN56_GLOBAL__N__00000000_17___tmp_kernel_1_cu_ffd1328e_745489nvfuser_1ENS_6TensorIfLi3ELi3EEENS0_IfLi2ELi2EEE' for 'sm_80'
> ptxas info    : Function properties for _ZN56_GLOBAL__N__00000000_17___tmp_kernel_1_cu_ffd1328e_745489nvfuser_1ENS_6TensorIfLi3ELi3EEENS0_IfLi2ELi2EEE
> ptxas         .     0 bytes stack frame, 0 bytes spill stores, 0 bytes spill loads
> ptxas info    : Used 18 registers, 4 bytes smem, 448 bytes cmem[0]
> ```
> 
> The `PRINTING __tmp_kernel_1.cu` part should already be handled by the existing regex, but this could be due to the added timestamp before the test name? That could've come from a changed pytest version or something.

`compare_codegen.sh` with the revised regex to process the time stamp in this PR works fine when I ran it locally. why failed again here? Does the CI use the script in this PR or it still uses the one in the main branch? @xwang233

tools/codediff/diff_report.py
liqiangxl: I revised the regex `re.compile(r".*?(?P<testname>\S+::\S+) (?P<line>.*)$")`. Basically a copy of the original one with `.*?` to skip the prefix time stamp.

liqiangxl:
!build --diff

liqiangxl:
> > I looked at the artifacts for the currently failing codediff job (2/4). It does seem that the warnings are printed to a separate file `stderr` and that `stdout` looks mostly fine. It does include some lines like this:
> > ```
> > [2024-04-01 16:34:10] tests/python/test_schedule_ops.py::TestScheduleOps::test_merge_op PRINTING: __tmp_kernel_1.cu
> > ptxas info    : Overriding maximum register limit 256 for '_ZN56_GLOBAL__N__00000000_17___tmp_kernel_1_cu_ffd1328e_745489nvfuser_1ENS_6TensorIfLi3ELi3EEENS0_IfLi2ELi2EEE' with  255 of maxrregcount option
> > ptxas info    : 3 bytes gmem, 24 bytes cmem[4]
> > ptxas info    : Compiling entry function '_ZN56_GLOBAL__N__00000000_17___tmp_kernel_1_cu_ffd1328e_745489nvfuser_1ENS_6TensorIfLi3ELi3EEENS0_IfLi2ELi2EEE' for 'sm_80'
> > ptxas info    : Function properties for _ZN56_GLOBAL__N__00000000_17___tmp_kernel_1_cu_ffd1328e_745489nvfuser_1ENS_6TensorIfLi3ELi3EEENS0_IfLi2ELi2EEE
> > ptxas         .     0 bytes stack frame, 0 bytes spill stores, 0 bytes spill loads
> > ptxas info    : Used 18 registers, 4 bytes smem, 448 bytes cmem[0]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The `PRINTING __tmp_kernel_1.cu` part should already be handled by the existing regex, but this could be due to the added timestamp before the test name? That could've come from a changed pytest version or something.
> 
> `compare_codegen.sh` with the revised regex to process the time stamp in this PR works fine when I ran it locally. why failed again here? Does the CI use the script in this PR or it still uses the one in the main branch? @xwang233

never mind, looks like the regex is not correct.

liqiangxl:
!build --diff

tools/codediff/diff_report.py
jacobhinkle: `.*` is enough to match anything including an empty string. Maybe the question mark is confusing it.

liqiangxl: `.*?: Matches any character (.) zero or more times (*), as few times as possible (?). This part is non-greedy or lazy, meaning it matches the shortest possible sequence.` [generated by AI]

I did a simple test using string `[2024-04-01 16:32:16] tests/python/test_python_frontend.py::TestNvFuserFrontend::test_input_scalar `

If using `pattern = re.compile(r".*?(?P<testname>\S+::\S+) (?P<line>.*)$")`, the captured test name is:
`Test Name: tests/python/test_python_frontend.py::TestNvFuserFrontend::test_input_scalar`

If drop `?` and use `pattern = re.compile(r".*(?P<testname>\S+::\S+) (?P<line>.*)$")`, the captured test name is:
`d::test_input_scalar`

jacobhinkle: Ah ok right. The `?` in this case is making `.*`, which is normally greedy, be lazy instead, so that you pick up as much of the test name as possible (with the greedy `\S+` included int he `testname` pattern). Makes sense to me. Have you tried running the `diff_report.py` script locally on the downloaded artifact?

jacobhinkle: Also, this regex will require a square bracket expression to follow the test name won't it since `\s+\[.*?\]\s+` will match " [] " at minimum? But we usually do not expect to have those bracket expressions following the test name.

liqiangxl: You are right. that is removed in the revised version, which is `re.compile(r".*?(?P<testname>\S+::\S+) (?P<line>.*)$")`
I believe when I tested it locally, I copied some results from CI where the time stamp is mixed with other msg, so there was `[]` when I wrote the regex.

liqiangxl:
Needs to check why there is code diff in `ShiftTest.Shift5ptStencilParallel1DThreadBlock`

PR #2011, opened 2024-04-01
make castOp part of matmul API

protonu:
This commit refactors the matmul API so that:
1. The CastOp is part of the matmul interface. (based on discussions with @kevinstephano, @Priya2698 and @jacobhinkle )
2. The Matmul Impl function will be reused by the API for linear. (follow-up commit)

### DIFF
--- a/csrc/ops/composite.cpp
+++ b/csrc/ops/composite.cpp
@@ -57,8 +57,9 @@ namespace {
 
 // This function will be used by both matmul and linear.
 // Matmul will add a castOp to the output of this function
-// whereas linear will skip the cast and add the bias.
-TensorView* matmulImpl(TensorView* a, TensorView* b) {
+// whereas linear will skip the cast (set cast flag as false) 
+// and add the bias.
+TensorView* matmulImpl(TensorView* a, TensorView* b, bool cast = true) {
   NVF_CHECK(
       a->nDims() == b->nDims(),
       "The number of dimension of A and B do not match");
@@ -66,9 +67,6 @@ TensorView* matmulImpl(TensorView* a, TensorView* b) {
   NVF_CHECK(
       a->nDims() == 2,
       "Only 2-D Tensors are supported, in the future we'll support 3-D as well!");
-  NVF_CHECK(
-      a->getDataType().value() == b->getDataType().value(),
-      "data types of inputs to matmul don't match");
 
   std::vector<bool> bcast_dims(a->nDims() + 1, false);
   // A: [M, K, Bcast]
@@ -78,14 +76,22 @@ TensorView* matmulImpl(TensorView* a, TensorView* b) {
   bcast_dims.at(bcast_dims.size() - 1) = false;
   bcast_dims.at(bcast_dims.size() - 3) = true;
   auto* tv1b = broadcast(b, bcast_dims);
+
+  NVF_CHECK(
+      a->getDataType().value() == b->getDataType().value(),
+      "data types of inputs to matmul don't match");
+  if (cast) {
+    // For matmul, the output dtype should match input.
+    return maybeCastOp(
+        a->getDataType().value(), fusedMultiplySum(tv0b, tv1b, {-2}));
+  }
   return fusedMultiplySum(tv0b, tv1b, {-2});
 }
 
 } // namespace
 
 TensorView* matmul(TensorView* a, TensorView* b) {
-  auto* tv = matmulImpl(a, b);
-  return castOp(a->getDataType().value(), tv);
+  return matmulImpl(a, b);
 }
 
 LstmResult lstm(
--- a/csrc/ops/composite.cpp
+++ b/csrc/ops/composite.cpp
@@ -53,7 +53,12 @@ TensorView* dropout_backward(TensorView* dy, TensorView* mask, Val* scale) {
   return dx;
 }
 
-TensorView* matmul(TensorView* a, TensorView* b) {
+namespace {
+
+// This function will be used by both matmul and linear.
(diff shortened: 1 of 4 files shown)


protonu:
!build

protonu:
!build

csrc/ops/composite.cpp
```diff
@@ -74,6 +81,13 @@ TensorView* matmul(TensorView* a, TensorView* b) {
   return fusedMultiplySum(tv0b, tv1b, {-2});
 }
 
+} // namespace
+
+TensorView* matmul(TensorView* a, TensorView* b) {
+  auto* tv = matmulImpl(a, b);
+  return castOp(a->getDataType().value(), tv);
```
jjsjann123, line 88: nitpick: it reads cleaner to me to have the
```
  NVF_CHECK(
      a->getDataType().value() == b->getDataType().value(),
      "data types of inputs to matmul don't match");
```

before casting output to one of the input types.

So can we move this cast inside `matmulImpl`?

jjsjann123 (approved):
LGTM

csrc/ops/composite.cpp
```diff
@@ -53,15 +53,22 @@ TensorView* dropout_backward(TensorView* dy, TensorView* mask, Val* scale) {
   return dx;
 }
 
-TensorView* matmul(TensorView* a, TensorView* b) {
+namespace {
+
+// This function will be used by both matmul and linear.
+// Matmul will add a castOp to the output of this function
+// whereas linear will skip the cast and add the bias.
+TensorView* matmulImpl(TensorView* a, TensorView* b) {
```
jacobhinkle, line 61: Suggestion: instead of an "impl" function, what if we accept an additional optional `DataType output_dtype` arg that defaults to `DataType::Null`. If that default is given then we replace it with the dtype of `a`. Then we can use `SimplifyingIrBuilder::maybeCastExpr` to cast to that dtype. That way we could still use the familiar `matmul()` interface to create matmuls with single-precision outputs when we want to by providing `/*output_dtype=*/DataType::Float`, but the default would be to downcast.

csrc/ops/composite.cpp
protonu: Thanks! Please let me know if I was not able to address your comment.

csrc/ops/composite.cpp
protonu: Thanks - I switched to using `maybeCastOp`, but I think I'd like to keep the Impl function and have a very thin matmul function. I'd prefer not to have that flag directly in matmul itself (thought I don't feel very strongly about it).

protonu:
!build

csrc/ops/composite.cpp
jacobhinkle: My reasoning was so that we could actually create a matmul with single-precision output from outside this compilation unit. As is currently in this PR, that functionality is hidden even in C++ is it not?

jacobhinkle: We don't need to expose this function in the python frontend, where I agree that we should keep the surface small. But on the C++ side it seems useful to be able to access single-precision output `matmul`, for example when creating matmuls with epilogues that don't fit the `linear` template.

protonu:
!build

protonu:
!build

protonu:
!build

csrc/ops/composite.cpp
protonu: Thanks for pointing that out. To enable using this from the C++ side without the cast I overloaded the matmul. Currently the python binding implementation is a bit messy which made calling the same matmul function with default args a bit tricky.

csrc/ops/composite.cpp
```diff
@@ -71,7 +73,20 @@ TensorView* matmul(TensorView* a, TensorView* b) {
...
+  auto* output = fusedMultiplySum(tv0b, tv1b, {-2});
+  if (cast_output_to_input) {
+    // For matmul, the output dtype should match input.
+    return maybeCastOp(a->getDataType().value(), output);
+  }
+  return output;
+}
+
+TensorView* matmul(TensorView* a, TensorView* b) {
+  return matmul(a, b, true /* cast output to input dtype */);
```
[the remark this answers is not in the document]
jjsjann123: I think the static cast thing here might be easily changed... https://github.com/NVIDIA/Fuser/blob/866f10a186c697ab8e915455325085794899372d/csrc/python_frontend/python_bindings.cpp#L1140

But I don't know if serde `Binary_TV` for matmul might have more static casts down the line.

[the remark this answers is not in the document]
protonu: As @jjsjann123 mentioned the problem came from creating an OpRecord where the function being called doesn't match the exact signature of the args. I think it can be solved, but I didn't think it was worth it at this point.

tests/python/pytest_opinfos.py
```diff
@@ -1107,37 +1106,20 @@ def torch_reshape_sym_fn(input_tensor, output_shaped_tensor):
...
-    lambda fd: partial(matmul_wrapper, fd, DataType.BFloat16),
-    "matmul_Bfloat16",
-    dtypes=(torch.bfloat16,),
+matmul_opinfo = OpInfo(
+    lambda fd: fd.ops.matmul,
+    "matmul",
+    # bf16 needs Ampere or newer.
+    dtypes=(
+        (torch.float16, torch.bfloat16)
+        if torch.cuda.get_device_properties(torch.cuda.current_device()).major >= 8
```
Priya2698, line 1115: nit: You may be able to use `is_pre_ampere()` here.

Priya2698 (approved):

PR #2013, opened 2024-04-01
Propagation rule for full

jjsjann123:
Refactoring:
1. allocation_order_map_ allows `empty` allocation order for tensors created by factory method, which would yield to input tensor with a meaningful allocation order.
2. Updated `resolveAllocationOrder` to be more conservative (used by propagation rule of Binary/Ternary operations). Instead of ignoring candidates without allocation order, we'll stop propagation when any candidates doesn't have recorded allocation order. This is to prevent the optimization pass aggressively setting allocation domain on output, where the larger tensor's allocation order isn't propagated through (i.e. operations not yet covered by propagation rule is in the data flow). This could lead to un-optimized output allocation domain configuration.

Minor changes:
1. Added propagation rule for factory method full.
3. Added tests


jjsjann123:
!build

jjsjann123:
!build

jjsjann123:
Marking a local failure.
`./build/nvfuser_tests --gtest_filter="TMACompileTimeInvalidTest.SizeOfTransfer"`. Not sure if it's real though, since it also repros on main for me but doesn't pop up in CI?! :confused: 

```
xpected: throws an exception which is a nvfuser::nvfError which contains .what() that has substring "The expected bytes must be a multiple of 16 bytes, but 8 is not."
  Actual: 16-byte object <10-8F 2E-6E FD-7F 00-00 60-8E 2E-6E FD-7F 00-00>, throws an exception which is a nvfuser::nvfError which contains .what() (of value = !sv->isFalse() INTERNAL ASSERT FAILED at "/opt/pytorch/nvfuser/csrc/device_lower/lower2device.h":256, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. The expected bytes must be a multiple of 16 bytes, but 0x7fb1485e1240 is not.
Exception raised from validate at /opt/pytorch/nvfuser/csrc/device_lower/lower2device.h:256 (most recent call first):
```

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -95,6 +96,12 @@ class AllocationOrderInferencer : public IterVisitor {
       const std::vector<IterDomain*>& permutation_ref) {
     if (auto iter = alloc_order_map_.find(producer);
         iter != alloc_order_map_.end()) {
+      // early termination to propagate empty allocation order
+      if (iter->second.empty()) {
+        alloc_order_map_[consumer] = {};
+        return true;
+      }
+
```
wujingyue, line 104: Is this needed? If the following code is general enough to correctly handle rank-0 (which seems so to me), I'd rather remove the unnecessary if-then.

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -123,8 +130,12 @@ class AllocationOrderInferencer : public IterVisitor {
   //
   // It scans through each candidate to find the first one that:
   //   1. is a TensorView
-  //   2. has an entry in alloc_order_map_
-  //   3. has the highest number of non_broadcast IterDomain
+  //   2. has the highest number of non_broadcast IterDomain
```
wujingyue, line 133:
```suggestion
  //   2. has the most non-broadcast IterDomains
```

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -40,6 +40,7 @@ class AllocationOrderInferencer : public IterVisitor {
  protected:
```
wujingyue, line 40: Can you add a code comment that the ctor takes a partially populated alloc_order_map that contains the ground truth (I guess allocation order of inputs/outputs) and the inferencer tries to fill the rest? This bit wasn't immediately clear to me and caused some unnecessary confusion.

csrc/preseg_passes/allocation_order_inference.cpp
jjsjann123: This bit is necessary.

I think my PR comment wasn't clear vvv

> allocation_order_map_ allows empty allocation order for tensors created by factory method

So it's not really trying to handle rank-0 tensor, but instead it's trying to handle cases where a factory method produces a tensor, where we record it with an empty allocation order.

The line below:
```
      std::vector<IterDomain*> alloc_domain =
          propagateAllocationDomain(producer, consumer);
```
Would try to compute the alloc_domain from the allocation order.

(I mean, I can move this short-cut to `propagateAllocationDomain` if you think that looks cleaner.

wujingyue: Can you remind me of the contract? Doesn't `alloc_order_map_[tv]` (if it exists in the map) need to have the same size as `tv`'s rfactor/allocation domain?

jjsjann123: ^^^ That was the contract and that's what I changed in this PR.

I failed at my PR description.

There's three cases now, `alloc_order_map_[tv]` should either:

1. have the same size as the rank of `tv`'s rfactor domain;
    we have a preferred allocation order for the entry and it should participate in propagation.
2. it could be an empty array;
    it's a wild card, it shouldn't dictate output allocation order, but it still tells that propagation is not interrupted.
3. not exist in the map.
    this is the case where it says propagation has stopped in the branch. (likely due to lack of allocation order on inputs or operation not yet covered by propagation rule).

The reason is coming from two parts:
1. I don't want to blindly ignore other operands that doesn't have an allocation order when resolving multiple candidates: imagine we have something like vvv
```
nvfusion(tv0, tv1):
  tv2 = op0(tv0) # not covered by propagation
  tv3 = add(tv2, tv1)
  return tv3
```
If tv0 is a bigger tensor than tv1, ideally we should propagate output tv3 to be in the same memory format as in tv0. Because op0 is not covered by our propagation rule yet, we won't see the allocation order for tv2.
So when we try to resolve the propagation for operations that takes multiple operands, we can really only make the decision when we know for sure that every thing coming

2. tensors created by factory method should carry a wild-card and not affecting propagation. Because those tensors are not going to affect vectorization.

So with those two considerations, I decided to expand the alloc_order_map_ to allow an `empty` entry to disambiguate wild-card tensor from tensor where propagation has failed.

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -170,8 +181,21 @@ TensorView* AllocationOrderInferencer::resolveAllocationOrder(
       continue;
     }
 
+    auto iter = alloc_order_map_.find(tv);
     // skip entry that doesn't have an allocation order
```
wujingyue, line 185: This comment is no longer consistent with the new code -- we return null instead of "skipping" the entry.

wujingyue: Btw, are you sure this change of behavior is fine?

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -170,8 +181,21 @@ TensorView* AllocationOrderInferencer::resolveAllocationOrder(
...
-    if (alloc_order_map_.count(tv) == 0) {
+    if (iter == alloc_order_map_.end()) {
+      return nullptr;
+    }
+
+    // skip entry that has an empty allocation order
+    if (iter->second.empty()) {
```
wujingyue, line 191: Will we be able to get rid of this if-then (therefore simplifies) if non_bc_high_water_mark is initialized to -inf?

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -95,6 +96,12 @@ class AllocationOrderInferencer : public IterVisitor {
       const std::vector<IterDomain*>& permutation_ref) {
```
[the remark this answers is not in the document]
jjsjann123: This is an overload function.
`permutation_ref` is used to compute the allocation order for consumer.

1. construct alloc_domain_producer for producer given its allocation_order.
2. map the alloc_domain_producer to consumer as alloc_domain.
3. compute permutation from `alloc_domain` to `permutation_ref` and store that as the allocation_order for consumer.

Usually we call `propagateAllocationOrder(producer, consumer)`, we'll by default use `consumer->getMaybeRFactorDomain()` as `permutation_ref`.

But this won't work for `pad`. since its output rfactor_domain is a resize and won't map to its producer any more. Hence we have this overload and would use its root_domain as reference instead: https://github.com/NVIDIA/Fuser/blob/79862aec572a9149fda7e464c2cc74dbd59bdb79/csrc/preseg_passes/allocation_order_inference.cpp#L295

csrc/preseg_passes/allocation_order_inference.cpp
jjsjann123: I don't think so. We still won't want to make sure factory-created tensor view yield to allocation order propagation from inputs where they have a real impact on vectorization.

i.e. initialization of non_bc_high_water_mark wouldn't change that behavior.

csrc/preseg_passes/allocation_order_inference.cpp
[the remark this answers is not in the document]
jjsjann123: ^^^ I think this is another inline comment I should have added. 📓

csrc/preseg_passes/allocation_order_inference.cpp
jjsjann123: Good catch. We are just stopping the propagation. I'll update the comment.

> are you sure this change of behavior is fine?
I'd prefer to have propagation failure, which motivates us to put in more propagation rule. Rather than having a propagation suggesting an allocation order on output that causes a perf regression. :crossed_fingers:

jjsjann123:
oops. sorry accidentally requested review @wujingyue  :bow: 
I'll let you know when I updated the PR.

jjsjann123:
@wujingyue request for more comment addressed.

jjsjann123:
!build

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -240,6 +323,12 @@ void AllocationOrderInferencer::handle(BroadcastOp* op) {
     return;
   }
 
+  // propagate empty allocation order;
+  if (iter->second.empty()) {
+    alloc_order_map_[out] = {};
+    return;
+  }
```
jacobhinkle, line 330: More generally, if all inputs have empty allocation orders, we should propagate as empty, right? If so, then maybe this could be done in `dispatch(Expr*)` instead of handling for each expr type.

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -114,6 +172,7 @@ class AllocationOrderInferencer : public IterVisitor {
     return false;
```
wujingyue, line 172: Now that the if-then branch gets very long, can we `return false` early? I.e.

```
if (found in alloc_order_map_) {
  long code;
  return true;
}
return false;
```
==> 
```
if (not found) {
  return false;
}
long code;
return true;
```

This reduces indentation and gets rid of a branch from readers' minds early.

csrc/preseg_passes/allocation_order_inference.cpp
jjsjann123: That's a good point. So `dispatch(Expr*)` should hijack the dispatch.

I'll start another PR playing with that. :+1:

csrc/preseg_passes/allocation_order_inference.cpp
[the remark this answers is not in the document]
wujingyue: As discussed offline, I think I understand the problem now: while `computePermutation` at line 159 needs rfactor IDs, `propagateAllocationDomain` returns root IDs not necessarily rfactor IDs. So, as a workaround for ops that transform root to rfactor (e.g. pad), we construct `permutation_ref` by hand so that it contains rfactor IDs. 

When we want to handle ViewOp (I guess soon?), we may need to compute permutation_ref  in a more sophisticated way like https://github.com/NVIDIA/Fuser/blob/main/csrc/alias_analysis.cpp#L163. 

SGTM if that's your understanding as well!

csrc/preseg_passes/allocation_order_inference.cpp
wujingyue: > inputs where they have a real impact on vectorization.

But won't those inputs have a positive non_bc_count and therefore "dominate"? I failed to see how an all-broadcast allocation domain can affect vectorization. 

Anyhow, if you think we need to distinguish empty allocation domain differently from all-broadcast allocation domain, you can **optionally** simplify the code a little bit

```
src = nullptr;
src_loop_domain_count = -inf;
src_allocation_order_size = -inf;

for each candidate {
  skip non TensorView or those that don't exist in alloc_order_map_. 

  loop_domain_count = countLoopIterDomains(candidate);
  allocation_order_size = candidate's allocation order size;
  if (make_pair(loop_domain_count, allocation_order_size) > make_pair(src_loop_domain_count, src_allocation_order_size)) {
    src = candidate;
    src_loop_domain_count = loop_domain_count;
    src_allocation_order_size = allocation_order_size;
  }
}
```

This better clarifies the intention of picking the candidate that has the most loop IterDomains and if there's a tie prefer the largest allocation_order size. The current pattern
```
if (...) {
  if (...) {
  }
  continue;
}
```
is a bit hard to follow.

wujingyue (approved):

jacobhinkle (approved):
LGTM!

csrc/preseg_passes/allocation_order_inference.cpp
jjsjann123: > I failed to see how an all-broadcast allocation domain can affect vectorization.

So factory method doesn't give an all-broadcast domain. Admittedly, `full` probably should, but we can also have random operation that shouldn't do that.

Quick example here:

```
T1_l[ iS2{16}, iS3{32} ]
   = full({16, 32}, float(1));
T2_g[ iS4{16}, iS5{32} ]
   = T0_g[ iS0{i0}, bS1{1} ]
   + T1_l[ iS2{16}, iS3{32} ];

TransformPrinter :
T0_g[ iS0{i0}, bS1{1} ]
 root domain : (iS0{i0}, bS1{1})
 allocation domain : (bS1{1}, iS0{i0})
T1_l[ iS2{16}, iS3{32} ]
 root domain : (iS2{16}, iS3{32})
 allocation domain : (iS2{16}, iS3{32})
T2_g[ iS4{16}, iS5{32} ]
 root domain : (iS4{16}, iS5{32})
```

i.e. T1 here is a tensor created by factory method, it has a higher loop_domain_count, but it shouldn't propagate its allocation order to T2.
We would want T0's allocation order to propagate to T2, which would allow both IO tensor to have the same innermost dimension -> vectorized load/store on both buffer.

(this is a bad example, since even when propagation failed, transpose scheduler should kick in and still give us vectorized load/store. But I hope it carries the argument across)

jjsjann123:
!build

jjsjann123:
will merge after CI clears. :crossed_fingers:

csrc/preseg_passes/allocation_order_inference.cpp
wujingyue: > T1 here is a tensor created by factory method, it has a higher loop_domain_count

I see -- countLoopIterDomains looks at the TensorView not `iter->second`. So the code I suggested would fail.

wujingyue (approved):

PR #2019, opened 2024-04-02
correct io bytes calculation in gelu bwd cpp benchmark

liqiangxl:
correct io bytes calculation in gelu bwd cpp benchmark.
Before: 1 `2D tensor` and  1 `1D tensor`
After: 3 `2D tensors`( 2 for inputs, 1 for output) and  1 `1D tensor`


jacobhinkle (approved):
Looks right to me. `t26r` is smaller than the other three inputs/outputs by a factor of `reduction_size`.

PR #1972, opened 2024-03-20
Add naive cache for provers

zasdfgbnm:
Before:
```C++
[       OK ] GPUTTensorCoreTest.FusionAmpereMatmul_CUDA (4571 ms)
[       OK ] NVFuserTest.FusionMagicSchedulerBatchNormalization_CUDA (2089 ms)
[       OK ] GpuViewTest.FusionReshapeReductionShmoo (17323 ms)
```

After:
```C++
[       OK ] GPUTTensorCoreTest.FusionAmpereMatmul_CUDA (3151 ms)
[       OK ] NVFuserTest.FusionMagicSchedulerBatchNormalization_CUDA (2044 ms)
[       OK ] GpuViewTest.FusionReshapeReductionShmoo (16693 ms)
```

PR #2020, opened 2024-04-02
Perform cancellation in SimplifyingIrBuilder::addExpr

jacobhinkle:
This change modifies `SimplifyingIrBuilder::addExpr` to cancel negative terms with positive ones. For example, the following expressions will all be simplified to `x + y`
- `x - (-y)`
- `(x + z) - ((-y) + z)`
- `(x + y) + (a + (b + (-(b + a)))`
This is a more powerful alternative to #2017, but it does introduce more complexity in `SimplifyingIrBuilder`.

Note that if cancellation is performed, terms might be reordered from the order given. However, if no cancellation is performed then `IrBuilder::addExpr(lhs, rhs)` is returned, so no reordering will take place.

### DIFF
--- a/csrc/ir/builder.cpp
+++ b/csrc/ir/builder.cpp
@@ -355,9 +355,132 @@ Val* SimplifyingIrBuilder::addExpr(Val* lhs, Val* rhs) {
     return addExpr(rhs, lhs->value());
   } else if (rhs->isConst()) {
     return addExpr(lhs, rhs->value(), rhs->dtype());
-  } else {
+  }
+  // Flatten nested additions. For example, suppose we have
+  //   lhs = a + ( b + c )
+  //   rhs = d + ( ( -b ) + e )
+  // Then we would flatten this to a sum of the terms
+  //   a b c d ( -b ) e
+  // Now suppose we have
+  //   lhs = a + b
+  //   rhs = d + ( -( b + e ) )
+  // We can flatten this to the sum of
+  //   a b d ( -b ) ( -e )
+  // We track a b d as positive terms and b e as negative terms, after which
+  // we can cancel b easily
+  std::vector<Val*> pos_terms;
+  std::vector<Val*> neg_terms;
+
+  // We'll do a stack-based search, using two stacks: one for positive and one
+  // for negative terms. Whenever the negative term stack is non-empty, we
+  // draw from it first.
+  std::vector<Val*> pos_term_stack{lhs, rhs};
+  std::vector<Val*> neg_term_stack;
+  while (!pos_term_stack.empty() || !neg_term_stack.empty()) {
+    Val* term = nullptr;
+    bool term_is_neg = false;
+    if (!neg_term_stack.empty()) {
+      term = neg_term_stack.back();
+      neg_term_stack.pop_back();
+      term_is_neg = true;
+    } else {
+      term = pos_term_stack.back();
+      pos_term_stack.pop_back();
+    }
+    if (auto* uop = dynamic_cast<UnaryOp*>(term->definition());
+        uop && uop->getUnaryOpType() == UnaryOpType::Neg) {
+      Val* unneg = uop->in();
+      if (term_is_neg) {
+        // This is a term like -(-a) or (-b) within -(a + (-b) ), so we treat
+        // it as a positive term
+        pos_term_stack.push_back(unneg);
+      } else {
+        neg_term_stack.push_back(unneg);
+      }
+      continue;
+    } else if (auto* bop = dynamic_cast<BinaryOp*>(term->definition())) {
+      if (bop->getBinaryOpType() == BinaryOpType::Add) {
+        if (term_is_neg) {
+          neg_term_stack.push_back(bop->lhs());
+          neg_term_stack.push_back(bop->rhs());
+        } else {
+          pos_term_stack.push_back(bop->lhs());
+          pos_term_stack.push_back(bop->rhs());
+        }
+        continue;
+      } else if (bop->getBinaryOpType() == BinaryOpType::Sub) {
+        if (term_is_neg) {
+          neg_term_stack.push_back(bop->lhs());
+          pos_term_stack.push_back(bop->rhs());
+        } else {
+          pos_term_stack.push_back(bop->lhs());
+          neg_term_stack.push_back(bop->rhs());
+        }
+        continue;
+      }
+    }
+    if (term_is_neg) {
+      neg_terms.push_back(term);
+    } else {
+      pos_terms.push_back(term);
+    }
+  }
+  // Now do cancellation
+  std::vector<bool> cancelled_pos(pos_terms.size(), false);
+  std::vector<bool> cancelled_neg(neg_terms.size(), false);
+  bool performed_cancellation = false;
+  for (size_t j : c10::irange(neg_terms.size())) {
+    if (cancelled_neg[j]) {
+      continue;
+    }
+    for (size_t i : c10::irange(pos_terms.size())) {
+      if (cancelled_pos[i]) {
+        continue;
+      }
+      if (neg_terms[j]->sameAs(pos_terms[i])) {
+        // Mark this pair of terms as cancelled
+        cancelled_pos[i] = true;
+        cancelled_neg[j] = true;
+        performed_cancellation = true;
+        break;
+      }
+    }
+  }
+  // If we didn't perform any cancellations, then perform the simple sum of
+  // lhs and rhs
+  if (!performed_cancellation) {
     return IrBuilder::addExpr(lhs, rhs);
   }
+
+  // If we did cancellations, then unflatten to produce simplified sum. We
+  // unflatten all the positives then all the negatives, then we negate the
+  // unflattened negative sum and add it to the positive sum.
+  Val* pos_sum = nullptr;
+  for (size_t i : c10::irange(pos_terms.size())) {
+    if (cancelled_pos[i]) {
+      continue;
+    }
+    pos_sum = pos_sum == nullptr ? pos_terms[i]
+                                 : IrBuilder::addExpr(pos_sum, pos_terms[i]);
+  }
+  Val* neg_sum = nullptr;
+  for (size_t j : c10::irange(neg_terms.size())) {
+    if (cancelled_neg[j]) {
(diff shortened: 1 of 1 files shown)


jacobhinkle:
!build --diff --diff-bench

jacobhinkle:
!build --diff --diff-bench

csrc/ir/builder.cpp
```diff
@@ -355,9 +355,132 @@ Val* SimplifyingIrBuilder::addExpr(Val* lhs, Val* rhs) {
...
+      neg_terms.push_back(term);
+    } else {
+      pos_terms.push_back(term);
+    }
+  }
+  // Now do cancellation
+  std::vector<bool> cancelled_pos(pos_terms.size(), false);
+  std::vector<bool> cancelled_neg(neg_terms.size(), false);
+  bool performed_cancellation = false;
+  for (size_t j : c10::irange(neg_terms.size())) {
```
jacobhinkle, line 432: Iterate over negative terms first since they are likely more rare. If no negative terms were found, we'll quickly see that no cancellation is possible and return the simple sum.

naoyam:
I wonder this level of simplification should be done by the expr simplifier. 

Some historical context: SimplifyingIrBuilder existed before the expr simplifier. We could replace SimplifyingIrBuilder entirely with the expr simplifier as the latter is more powerful. However, it's also much more costly than the simple and quick simplification done by SimplifyingIrBuilder. We can use SimplifyingIrBuilder without worrying about added overheads but that's not the case with the expr simplifier.

naoyam:
> I wonder this level of simplification should be done by the expr simplifier.
> 
> Some historical context: SimplifyingIrBuilder existed before the expr simplifier. We could replace SimplifyingIrBuilder entirely with the expr simplifier as the latter is more powerful. However, it's also much more costly than the simple and quick simplification done by SimplifyingIrBuilder. We can use SimplifyingIrBuilder without worrying about added overheads but that's not the case with the expr simplifier.

So, to me, SimplifyingIrBuilder is for quick low-hanging fruits only and thus can be used without worrying about overheads, whereas the expr simplifier is a heavy-weight process that's much more comprehensive, so it should be used selectively. 

Any thoughts? @zasdfgbnm

naoyam:
> I agree with this statement. However, it is not clear to me what should be the boundary. What is considered a "low-hanging fruits" and what should be considered as "heavy-weight"? We need a clear definition here.

Maybe, no recursive simplification? 

> An alternative idea: I think the idea of ["optimization fuel"](http://blog.ezyang.com/2011/06/debugging-compilers-with-optimization-fuel/) that I learnt from @wujingyue's talk is really interesting. Should we just totally kill SimplifyingIrBuilder and add a double fuel argument to simplifyExpr guarding how much effort we should spend on simplification? simplifyExpr(..., fuel=0.0) means no simplification at all, simplifyExpr(..., fuel=inf) means try the best to simplify. We can use a small fuel value if we don't want to worry about overheads.

Or super fast expr simplifier? 😉

zasdfgbnm:
In order to use `simplifyExpr` and get some useful simplifications, I think the minimum price we have to pay is `assoc_comm::flatten`+`eliminateTrivialComputation`, which is slightly higher than this PR.

jacobhinkle:
If `simplifyExpr` could cache its simplifications, at least for a given set of assumptions, then it could be used without assumptions in a practically non-recursive fashion and should be pretty fast. We could keep a thread-local `Context` that has no assumptions and use `simplifyExpr(IrBuilder::addExpr(lhs, rhs))` in place of `SimplifyingIrBuilder::addExpr(lhs, rhs)`. Then we could cache simplified expressions so that `recurseDown` stops when it recognizes an already-simplified expression.  That's similar to what @zasdfgbnm did for inequality proofs in #1972 .

zasdfgbnm:
Did a quick experiment in https://github.com/NVIDIA/Fuser/pull/2022, which uses `simplifyExpr` (enabling only `eliminateTrivialComputation` disabling all proves, with the cache mentioned by @jacobhinkle  enabled) to replace `SimplifyingIrBuilder`, the time for `GPUTTensorCoreTest.FusionAmpereMatmul_CUDA` goes up from `4012 ms` to `10686 ms` if I just mindlessly replace all `SimplifyingIrBuilder` with `simplifyExpr`. If I am being smarter in the sense that if I know this expr will eventually be simplified, I use `IrBuilder` instead of `SimplifyingIrBuilder`, I am bring the test time back to `4482 ms`, still a noticeable 10% slow down compared with the previous `SimplifyingIrBuilder`.

jacobhinkle:
On my workstation, I see `FusionAmpereMatmul_CUDA` go from 3460 ms on main (3686a961) to 3618 ms (+4.5%) with this PR. But there's a lot of error in this measurement. There doesn't seem to be a consistently large perf hit due to this PR.
```
withPR2020.out                                                                                                             
63:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmul_CUDA (3618 ms)                 
69:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulBFloat16_CUDA (3416 ms)         
71:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulPipelineGmem_CUDA (4554 ms)   
75:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulRegDoubleBuffer_CUDA (6883 ms) 
83:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulTNcpAsync_CUDA (1327 ms)    
89:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulTNSwizzled_CUDA (1643 ms)                       
91:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulLargeLoad_CUDA (3492 ms)                                              
95:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulTileCheck4warp_CUDA (47156 ms)                                        
97:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulTileCheck8warp_CUDA (36610 ms)
99:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulTileCheck6warp_CUDA (10190 ms)
101:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulLargeLoadLargeK_CUDA (3134 ms)
105:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulSmemEpilogue_CUDA (4495 ms)   
130:[  FAILED  ] GPUTTensorCoreTest.FusionAmpereMatmulSmemEpiloguePromotionRequiredA100_CUDA (620 ms)                      
132:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulSmemEpilogueCast_CUDA (4296 ms)                                      
134:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulSmemEpilogueRelu_CUDA (4354 ms)
136:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulSplitK_CUDA (5968 ms)        
138:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulSplitKBias_CUDA (6993 ms)    
140:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulBatchSplitK_CUDA (6754 ms)   
142:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulBatchSplitKBias_CUDA (8305 ms)
                                                                                                                           
main.out                                                                                                                   
63:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmul_CUDA (3460 ms)                 
69:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulBFloat16_CUDA (3441 ms)         
71:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulPipelineGmem_CUDA (4581 ms)  
75:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulRegDoubleBuffer_CUDA (6974 ms) 
83:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulTNcpAsync_CUDA (1285 ms)    
89:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulTNSwizzled_CUDA (1618 ms)                       
91:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulLargeLoad_CUDA (3127 ms)                                                           
95:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulTileCheck4warp_CUDA (48088 ms)                                                     
97:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulTileCheck8warp_CUDA (35263 ms)                                                     
99:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulTileCheck6warp_CUDA (9878 ms)                                                      
101:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulLargeLoadLargeK_CUDA (3637 ms)                                                    
105:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulSmemEpilogue_CUDA (4866 ms)                                                       
130:[  FAILED  ] GPUTTensorCoreTest.FusionAmpereMatmulSmemEpiloguePromotionRequiredA100_CUDA (673 ms)                                                                     
132:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulSmemEpilogueCast_CUDA (4862 ms)                                                                                     
134:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulSmemEpilogueRelu_CUDA (4290 ms)                                                                                     
136:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulSplitK_CUDA (6633 ms)                                                                                               
138:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulSplitKBias_CUDA (7207 ms)                                                                                           
140:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulBatchSplitK_CUDA (6819 ms)                                                                                          
142:[       OK ] GPUTTensorCoreTest.FusionAmpereMatmulBatchSplitKBias_CUDA (8366 ms)
```

jacobhinkle:
I'm going to merge main now that @naoyam has merged #2017 then check codediff to see if there is any value in adding this now.

jacobhinkle:
!build --diff

jacobhinkle:
Diff used old commit. Retrying.

jacobhinkle:
!build --diff

jacobhinkle:
!build --diff

PR #2021, opened 2024-04-02
Avoid contradicting assumptions from empty loops

### LINKED ISSUE #1808: Error compiling simple GEMM with specific warp/gemm tile sizes
To reproduce, modify `FusionAmpereMatmul_CUDA` in `test_gpu_tensorcore.cpp` as follows:
```diff
diff --git a/test/test_gpu_tensorcore.cpp b/test/test_gpu_tensorcore.cpp
index d29f7d2d..76d9f742 100644
--- a/test/test_gpu_tensorcore.cpp
+++ b/test/test_gpu_tensorcore.cpp
@@ -77,8 +77,8 @@ TEST_F(NVFuserTest, FusionAmpereMatmul_CUDA) {
     fusion.addOutput(tv2);
 
     MatMulTileOptions gemm_tile;
-    gemm_tile.cta_tile = GemmTile(128, 128, 32);
-    gemm_tile.warp_tile = GemmTile(64, 64, 32);
+    gemm_tile.cta_tile = GemmTile(160, 144, 16);
+    gemm_tile.warp_tile = GemmTile(80, 24, 16);
     gemm_tile.instruction_tile = GemmTile(16, 8, 16);
 
     MatmulParams params;
```
This hits the following inside `hoistScalar` during `IndexLowering`:

> C++ exception with description "!(x->sameAs(b) && y->sameAs(a)) INTERNAL ASSERT FAILED at "/opt/pytorch/nvfuser/csrc/expr_simplifier.cpp":215, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. Found two contradicting assumptions: i684 <= 0 and 0 < i684 both exist.

jacobhinkle:
This detects when a loop has zero iterations and avoids contradictions when simplifying included expressions.

Fixes #1808

### DIFF
--- a/csrc/device_lower/pass/scalar_hoist.cpp
+++ b/csrc/device_lower/pass/scalar_hoist.cpp
@@ -347,7 +347,7 @@ std::vector<Val*> getAssumptions(const std::vector<kir::ForLoop*>& loops) {
       // important to simplify its index. However, it is important that we avoid
       // contradicting assumptions, so we omit the index < stop assumption in
       // these cases.
-      TORCH_WARN(
+      TORCH_WARN_ONCE(
           "Encountered loop with no iterations. Stop value ",
           stop->toInlineString(),
           " is same as start value ",
--- a/csrc/device_lower/pass/scalar_hoist.cpp
+++ b/csrc/device_lower/pass/scalar_hoist.cpp
@@ -342,11 +342,19 @@ std::vector<Val*> getAssumptions(const std::vector<kir::ForLoop*>& loops) {
     Val* start = loop->start();
     assumptions.push_back(IrBuilder::geExpr(loop->index(), start));
     Val* stop = loop->simplifiedStop();
-    if (!stop->sameAs(start)) {
+    if (stop->sameAs(start)) {
       // If stop = start, then this loop will not be computed, so it's not
       // important to simplify its index. However, it is important that we avoid
       // contradicting assumptions, so we omit the index < stop assumption in
       // these cases.
+      TORCH_WARN(
+          "Encountered loop with no iterations. Stop value ",
+          stop->toInlineString(),
+          " is same as start value ",
+          start->toInlineString(),
+          ". This could indicate a suboptimal schedule such as double-buffering a ",
+          "loop that has only a single iteration.");
+    } else {
       assumptions.push_back(IrBuilder::ltExpr(loop->index(), stop));
     }
   }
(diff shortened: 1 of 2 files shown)


jacobhinkle:
!build --diff --diff-bench

zasdfgbnm (approved):

jacobhinkle:
I'm considering printing a warning in these cases since provably empty loops likely indicate a suboptimal schedule. In the matmul case it is a double buffered loop with a single iteration, for example.

csrc/device_lower/pass/scalar_hoist.cpp
```diff
@@ -339,9 +339,24 @@ std::vector<Val*> getAssumptions(const std::vector<kir::ForLoop*>& loops) {
...
+      // important to simplify its index. However, it is important that we avoid
+      // contradicting assumptions, so we omit the index < stop assumption in
+      // these cases.
+      TORCH_WARN_ONCE(
+          "Encountered loop with no iterations. Stop value ",
+          stop->toInlineString(),
+          " is same as start value ",
+          start->toInlineString(),
+          ". This could indicate a suboptimal schedule such as double-buffering a ",
+          "loop that has only a single iteration.");
```
jacobhinkle, line 356: Warning looks like 
```
[W scalar_hoist.cpp:350] Warning: Encountered loop with no iterations. Stop value 0 is same as start value 0. This could indicate a suboptimal schedule such as double-buffering a loop that has only a single iteration. (function operator())
```

csrc/device_lower/pass/scalar_hoist.cpp
```diff
@@ -342,11 +342,19 @@ std::vector<Val*> getAssumptions(const std::vector<kir::ForLoop*>& loops) {
     Val* start = loop->start();
     assumptions.push_back(IrBuilder::geExpr(loop->index(), start));
     Val* stop = loop->simplifiedStop();
-    if (!stop->sameAs(start)) {
+    if (stop->sameAs(start)) {
       // If stop = start, then this loop will not be computed, so it's not
       // important to simplify its index. However, it is important that we avoid
       // contradicting assumptions, so we omit the index < stop assumption in
       // these cases.
+      TORCH_WARN(
```
zasdfgbnm: No need to be this PR: should we add a `NVF_WARN`?

jacobhinkle: We definitely should, if for no other reason than to capture the output in python using debug().

PR #2024, opened 2024-04-02
Unify matmul pattern matching

Priya2698:
1. Adds alpha, beta parameters in matmul pattern matching
2. Unifies matching for matmul and matmul + bias.

Issue #1669.

### DIFF
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -444,6 +444,7 @@ std::vector<PolymorphicValue> UnaryOp::evaluate(
 
       if (bias_tensor.dim() != a.dim()) {
         // Bias is of shape [M,].
+        NVF_ERROR(bias_tensor.dim() == a.dim() - 1);
         bias_tensor = bias_tensor.unsqueeze(-1);
       }
 
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -412,23 +412,11 @@ std::vector<PolymorphicValue> UnaryOp::evaluate(
   if (is_downcast() && input(0)->definition() != nullptr) {
     Val* mma_lhs = nullptr;
     Val* mma_rhs = nullptr;
-
-    // Case 1: MmaOp + Cast (Matmul)
-    if (MmaOpUtils::matchMatmulCast(
-            this, std::ref(mma_lhs), std::ref(mma_rhs))) {
-      // Inputs to MmaOp are of the shape [M, K, 1] x [1, K, N]
-      const auto a =
-          ee.evaluate(mma_lhs, known_values).as<at::Tensor>().squeeze(-1);
-      const auto b =
-          ee.evaluate(mma_rhs, known_values).as<at::Tensor>().squeeze(0);
-      return {a.matmul(b)};
-    }
-
     Val* bias_val = nullptr;
     Val* alpha_val = nullptr;
     Val* beta_val = nullptr;
-    // Case 2: Matmul + Bias
-    if (MmaOpUtils::matchMatmulBiasCast(
+
+    if (MmaOpUtils::matchMatmulPatterns(
             this,
             std::ref(mma_lhs),
             std::ref(mma_rhs),
@@ -440,16 +428,18 @@ std::vector<PolymorphicValue> UnaryOp::evaluate(
           ee.evaluate(mma_lhs, known_values).as<at::Tensor>().squeeze(-1);
       const auto b =
           ee.evaluate(mma_rhs, known_values).as<at::Tensor>().squeeze(0);
-      auto bias = ee.evaluate(bias_val, known_values).as<at::Tensor>();
+      const c10::Scalar alpha =
+          alpha_val ? toScalar(ee.evaluate(alpha_val, known_values)) : 1;
 
+      if (bias_val == nullptr) {
+        return {alpha * a.matmul(b)};
+      }
+      auto bias = ee.evaluate(bias_val, known_values).as<at::Tensor>();
       if (bias.dim() != a.dim()) {
         // Bias is of shape [M,].
         NVF_ERROR(bias.dim() == a.dim() - 1);
         bias = bias.unsqueeze(-1);
       }
-
-      const c10::Scalar alpha =
-          alpha_val ? toScalar(ee.evaluate(alpha_val, known_values)) : 1;
       const c10::Scalar beta =
           beta_val ? toScalar(ee.evaluate(beta_val, known_values)) : 1;
 
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -425,18 +425,34 @@ std::vector<PolymorphicValue> UnaryOp::evaluate(
     }
 
     Val* bias = nullptr;
+    Val* alpha = nullptr;
+    Val* beta = nullptr;
     // Case 2: Matmul + Bias
     if (MmaOpUtils::matchMatmulBiasCast(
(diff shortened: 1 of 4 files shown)


Priya2698:
!build

wujingyue (commented):
LGTM! I'll defer approval to Jacob.

csrc/ir/utils.cpp
```diff
@@ -1400,22 +1417,37 @@ bool matchMatmulBiasCast(
...
+  }
 
-  bias = bcast->input(0); // Bias tensor in fp32
   auto* bias_cast = dynamic_cast<UnaryOp*>(bias->definition());
 
   // The bias tensor and matmul inputs should be of the same dtype.
   NVF_ERROR(
       bias_cast == nullptr || bias_cast->getUnaryOpType() == UnaryOpType::Cast,
-      "Expected the bias tensor to be preceded by castOp before broadcasting.");
+      "Expected the bias tensor to casted to Float.");
```
jacobhinkle, line 1450: I might have already asked this, but does the `bias` tensor have to be same type as `A` and `B` for `at::addmm`?

csrc/ir/nodes.cpp
```diff
@@ -425,18 +425,34 @@ std::vector<PolymorphicValue> UnaryOp::evaluate(
...
           ee.evaluate(mma_lhs, known_values).as<at::Tensor>().squeeze(-1);
       const auto b =
           ee.evaluate(mma_rhs, known_values).as<at::Tensor>().squeeze(0);
-      // Bias is of shape [M,].
-      const auto bias_tensor =
-          ee.evaluate(bias, known_values).as<at::Tensor>().unsqueeze(-1);
-      return {at::addmm(bias_tensor, a, b)};
+      auto bias_tensor = ee.evaluate(bias, known_values).as<at::Tensor>();
+
+      if (bias_tensor.dim() != a.dim()) {
```
jacobhinkle, line 445: Technically, we could have a bias that is 3D and matmul output that is 2D and add them, relying on implicit broadcasting. So we should probably just explicitly check the case that `bias_tensor.dim() == a.dim()` or `bias_tensor.dim() == a.dim() - 1` else fail the pattern.

csrc/ir/utils.h
```diff
@@ -66,7 +66,9 @@ bool matchMatmulBiasCast(
     const UnaryOp* cast_op,
     Val*& mma_lhs,
     Val*& mma_rhs,
-    Val*& bias);
+    Val*& bias,
+    Val*& alpha,
+    Val*& beta);
```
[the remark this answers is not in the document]
Priya2698: Yes, I have a separate PR for that but I can also have the changes here itself. Might simplify a few things.

csrc/ir/nodes.cpp
```diff
@@ -425,18 +425,34 @@ std::vector<PolymorphicValue> UnaryOp::evaluate(
...
+        // Bias is of shape [M,].
+        bias_tensor = bias_tensor.unsqueeze(-1);
+      }
+
+      const auto alpha_tensor =
+          alpha ? toScalar(ee.evaluate(alpha, known_values)) : 1;
+      const auto beta_tensor =
+          beta ? toScalar(ee.evaluate(beta, known_values)) : 1;
```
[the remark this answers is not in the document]
Priya2698: Ahh right, that's an error.

csrc/ir/utils.cpp
```diff
@@ -1378,17 +1378,34 @@ bool matchMatmulCast(const UnaryOp* cast_op, Val*& mma_lhs, Val*& mma_rhs) {
...
+    Val*& bias,
+    Val*& alpha,
+    Val*& beta) {
   auto* binary = dynamic_cast<BinaryOp*>(cast_op->input(0)->definition());
   if (binary == nullptr || binary->getBinaryOpType() != BinaryOpType::Add) {
     return false;
   }
 
-  auto* mma = dynamic_cast<MmaOp*>(binary->input(0)->definition());
+  // Check for alpha in first input: alpha * (MmaOp(A, B))
```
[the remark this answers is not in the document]
Priya2698: You're right. This pattern won't get matched.
Will make the change.

csrc/ir/utils.cpp
Priya2698: Yes, the dtypes need to be the same

csrc/ir/nodes.cpp
Priya2698: Only the second condition should occur, right? Matmul operands will be 3D and bias will be 2D.

csrc/ir/utils.cpp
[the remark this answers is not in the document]
Priya2698: This also goes for the first `BinaryOpType::Add`. biasEpilogue uses one format, but using `add` we could swap bias and mma.

csrc/ir/nodes.cpp
jacobhinkle: OK yeah I forgot about the broadcast. but we could then have a 4D or 5D "bias" being added to the matmul output. This current pattern would match it and unsqueeze to 5 or 6 dimensions, whereas it should just fail to match in those cases.

csrc/ir/utils.cpp
```diff
@@ -1366,46 +1366,52 @@ void verifyMmaOpForEvaluation(
...
+// 6. alpha * A x B
+// Note: We assume the first operand to be the MmaOp output
+bool matchMatmulPatterns(
     const UnaryOp* cast_op,
     Val*& mma_lhs,
     Val*& mma_rhs,
     Val*& bias,
     Val*& alpha,
     Val*& beta) {
+  // Check if there may be a bias present.
```
jacobhinkle, line 1384: You might want to either initialize all the outputs to `nullptr` here, or (possibly better) return a struct containing these values. That way you can be sure that checking for `nullptr` really means that element was not matched.

jacobhinkle: Another advantage of using a struct is that you don't have to update every call site if you extend the pattern in the future.

jacobhinkle (approved):
Looking good! Just one suggestion. Nice work.

csrc/ir/utils.cpp
Priya2698: Great point, added `MatmulInputs` to hold all the variables which are initialized to `nullptr`.

Priya2698:
!build

csrc/ir/utils.h
```diff
@@ -60,13 +60,20 @@ MmaOpDetails getMmaOpDetails(
 
 void verifyMmaOpForEvaluation(MmaOp* mma_op, DataType expected_input_dtype);
 
-bool matchMatmulCast(const UnaryOp* cast_op, Val*& mma_lhs, Val*& mma_rhs);
+struct MatmulInputs {
```
wujingyue, line 63: Is LinearLayerInputs a better name? MatmulInputs gives me the wrong impression that it's just a and b.

csrc/ir/utils.cpp
```diff
@@ -1366,29 +1366,46 @@ void verifyMmaOpForEvaluation(
...
-  if (mma == nullptr) {
+// Possible combinations:
+// 1. A x B + C
+// 2. alpha * A x B + C
+// 3. A x B + beta * C
+// 4. alpha * A x B  + beta * C
+// 5. A x B
+// 6. alpha * A x B
+// Note: We assume the first operand to be the MmaOp output
+bool matchMatmulPatterns(const UnaryOp* cast_op, MatmulInputs* matmul_inp) {
```
wujingyue, line 1377: MatmulInputs&

PR #2031, opened 2024-04-03
Add a unit test to try transpose + matmul + cast with EE evaluator

protonu:
This adds a test to try out cast(a.dtype,  Matmul(a, b.transpose()))


protonu:
!build

tests/cpp/test_matmul_aten_evaluation.cpp
```diff
@@ -68,6 +68,44 @@ TEST_F(MatmulATenEvaluationTest, MmaOpAndCast) {
   EXPECT_TRUE(at::allclose(out[0], out_ref));
 }
 
+TEST_F(MatmulATenEvaluationTest, TransposeMmaOpAndCast) {
+  auto fusion = std::make_unique<Fusion>();
+  FusionGuard fg(fusion.get());
+
+  int64_t m = 32, n = 64, k = 128;
+  std::vector<int64_t> a_shape{m, k}, b_shape{n, k}, out_shape{m, n};
```
wujingyue, line 76:
```suggestion
  std::vector<int64_t> a_shape{m, k}, b_shape{n, k};
```

wujingyue (approved):

protonu:
!build

protonu:
!build

tests/cpp/test_matmul_aten_evaluation.cpp
```diff
@@ -68,6 +68,44 @@ TEST_F(MatmulATenEvaluationTest, MmaOpAndCast) {
...
 
+TEST_F(MatmulATenEvaluationTest, TransposeMmaOpAndCast) {
+  auto fusion = std::make_unique<Fusion>();
+  FusionGuard fg(fusion.get());
+
+  int64_t m = 32, n = 64, k = 128;
+  std::vector<int64_t> a_shape{m, k}, b_shape{n, k};
+
+  auto tv0 = makeConcreteTensor(a_shape, DataType::Half);
+  auto tv1 = makeConcreteTensor(b_shape, DataType::Half);
```
jacobhinkle, line 79: Is there a reason to make these concrete? Actually, what are the contiguity requirements for `at::matmul` and `at::addmm`? If there are none, then it would be nice to have tests that have discontiguous inputs. Also having some with inputs that have M, K, and N not divisible by 8 would let us test cases with unaligned rows (currently not supported in nvfuser matmul).

tests/cpp/test_matmul_aten_evaluation.cpp
```diff
@@ -68,6 +68,44 @@ TEST_F(MatmulATenEvaluationTest, MmaOpAndCast) {
...
+
+  fusion->addInput(tv0);
+  fusion->addInput(tv1);
+  fusion->addOutput(tv3);
+
+  at::Tensor t0 = at::randn(a_shape, at::kHalf).cuda();
+  at::Tensor t1 = at::randn(b_shape, at::kHalf).cuda();
+  at::Tensor out_ref = at::matmul(t0, t1.transpose(0, 1));
+
+  FusionExecutorCache fec(std::move(fusion));
```
jacobhinkle, line 94: Using the `FusionExecutorCache` here means you are also exercising the segmenter. That's fine if you don't suspect work will be needed on the segmenter, but in the future when exercising `linear`, I suggest creating a test that creates a `FusionExecutor` and marks aliasing for the outputs directly instead, so that you can avoid issues with segmentation and focus on getting ATen execution of linear working.

tests/cpp/test_matmul_aten_evaluation.cpp
```diff
@@ -68,6 +68,44 @@ TEST_F(MatmulATenEvaluationTest, MmaOpAndCast) {
...
+  auto out = fec.runFusionWithInputs({t0, t1});
+
+  const std::vector<FusionExecutor>& executors =
+      fec.getMostRecentKernelRuntime()->executors();
+  EXPECT_EQ(executors.size(), 1);
```
jacobhinkle, line 99: `ASSERT_EQ` to avoid exception in `executors.front()`.

jacobhinkle (approved):
LGTM

tests/cpp/test_matmul_aten_evaluation.cpp
protonu: I think these are great points. These apply to the tests for ATen execution, and not just this test. I created a GH issue for this (https://github.com/NVIDIA/Fuser/issues/2032).

jjsjann123 (approved):
so cpp test works fine but yesterday the python example was triggering an assert that no scheduler is taking the fusion?!
That's interesting...

protonu:
!build

PR #2033, opened 2024-04-03
Move doc for expr simplifier to Markdown

zasdfgbnm:
Mostly reformatting. Also contain small rephrases in a few places.

I am doing so because when writing the doc for TMA, I sometimes feel useful to cross reference some theorems in expr simplifier, but it is awkward to link to a theorem in a `.h` file...

I believe this part of math should be not just "a doc for expr simplifier", but instead "part of our project's common mental model", so I promote it to the `doc` directory.


naoyam (approved):
LGTM

PR #2036, opened 2024-04-03
Torch baselines for RMSNorm

Priya2698:
Issue #1668


liqiangxl (approved):
LGTM. Similar to prevous PR, add comments about why need to manuually calculate IO bytes, what's the difference.

liqiangxl (approved):
LGTM.

PR #2037, opened 2024-04-03
Torch baselines for `batchnorm` and `instance_norm`

Priya2698:
Issue #1668

### DIFF
--- a/benchmarks/python/normalization.py
+++ b/benchmarks/python/normalization.py
@@ -2,7 +2,8 @@ from nvfuser import FusionDefinition, DataType
 from .global_params import PROMOTE_DTYPES
 from nvfuser.pytorch_utils import torch_dtype_to_nvfuser_dtype
 import torch
-from .core import run_benchmark, clear_cuda_cache
+from .core import run_benchmark, clear_cuda_cache, unary_bwd_torch
+import numpy as np
 
 
 def norm_fwd_fusion(
@@ -188,7 +189,7 @@ def norm_bwd_fusion(
     fd.add_output(grad_bias)
 
 
-def norm_fwd_benchmark(
+def norm_fwd_nvf_benchmark(
     benchmark,
     size: tuple,
     dtype: torch.dtype,
@@ -209,7 +210,7 @@ def norm_fwd_benchmark(
     # Size is assumed to be in the order N, C, ...
     num_dims = len(size)
 
-    at_inputs = torch.randn(*size, device="cuda", dtype=dtype, requires_grad=True)
+    at_inputs = torch.randn(size, device="cuda", dtype=dtype, requires_grad=True)
     weight = torch.randn(size[1], device="cuda", dtype=dtype)
     bias = torch.randn(size[1], device="cuda", dtype=dtype)
 
@@ -275,7 +276,7 @@ def norm_fwd_benchmark(
         )
 
 
-def norm_bwd_benchmark(
+def norm_bwd_nvf_benchmark(
     benchmark,
     size: tuple,
     dtype: torch.dtype,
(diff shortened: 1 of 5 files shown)


Priya2698:
> can you add comments about which dim is normalized and which scheduler is being used?

Is there a DEBUG_DUMP option to print the scheduler used or I'll need to debug statements to infer?

liqiangxl:
> > can you add comments about which dim is normalized and which scheduler is being used?
> 
> Is there a DEBUG_DUMP option to print the scheduler used or I'll need to debug statements to infer?

you can use NVFUSER_DUMP=scheduler_params
For normalization we have: inner persistent (2D or 3D), outer persistent, innerOuter persistent
IIRC, instance norm is outer persistent and batch norm is inner persistent 3D.

Priya2698:
> > > can you add comments about which dim is normalized and which scheduler is being used?
> > 
> > 
> > Is there a DEBUG_DUMP option to print the scheduler used or I'll need to debug statements to infer?
> 
> you can use NVFUSER_DUMP=scheduler_params For normalization we have: inner persistent (2D or 3D), outer persistent, innerOuter persistent IIRC, instance norm is outer persistent and batch norm is inner persistent 3D.

Thanks.
While looking into the schedulers used for layernorm, it used reduction kernel for larger sizes.

liqiangxl:
> > > > can you add comments about which dim is normalized and which scheduler is being used?
> > > 
> > > 
> > > Is there a DEBUG_DUMP option to print the scheduler used or I'll need to debug statements to infer?
> > 
> > 
> > you can use NVFUSER_DUMP=scheduler_params For normalization we have: inner persistent (2D or 3D), outer persistent, innerOuter persistent IIRC, instance norm is outer persistent and batch norm is inner persistent 3D.
> 
> Thanks. While looking into the schedulers used for layernorm, it used reduction kernel for larger sizes. The inner outer kernel was used in the smaller ranges.

If segmented, it uses a reduction kernel followed by a pointwise.

Priya2698:
@liqiangxl 
For a subset of the inputs, I saw the following schedulers:
1. Reduction with tag Outer persistent
2. Reduction with no tags
3. Reduction + Pointwise

Since there are large number of inputs, I haven't verified it for all input sizes.
What is the motivation behind listing the scheduler? It is not feasible to do it atleast manually.

liqiangxl:
> @liqiangxl For a subset of the inputs, I saw the following schedulers:
> 
> 1. Reduction with tag Outer persistent
> 2. Reduction with no tags
> 3. Reduction + Pointwise
> 
> Since there are large number of inputs, I haven't verified it for all input sizes. What is the motivation behind listing the scheduler? It is not feasible to do it atleast manually.

It seems that we need to clean the scheduler tags. The motivation is to know which scheduler is being used so we can quickly locate the issue if there is a regression and we may also add a feature to select a subset of benchmarks based on the scheduler. For example, if I modified the inner persistent scheduler, I probably only want to run benchmarks using the inner persistent scheduler.
Since this info is also missing in other benchmark cases, we can discuss it later and add in another PR. @naoyam what do you think about ` add a feature to select a subset of benchmarks based on the scheduler`?

liqiangxl (approved):
LGTM. discuss adding comments and tags about which scheduler is being used.

Priya2698:
> > @liqiangxl For a subset of the inputs, I saw the following schedulers:
> > 
> > 1. Reduction with tag Outer persistent
> > 2. Reduction with no tags
> > 3. Reduction + Pointwise
> > 
> > Since there are large number of inputs, I haven't verified it for all input sizes. What is the motivation behind listing the scheduler? It is not feasible to do it atleast manually.
> 
> It seems that we need to clean the scheduler tags. The motivation is to know which scheduler is being used so we can quickly locate the issue if there is a regression and we may also add a feature to select a subset of benchmarks based on the scheduler. For example, if I modified the inner persistent scheduler, I probably only want to run benchmarks using the inner persistent scheduler. Since this info is also missing in other benchmark cases, we can discuss it later and add in another PR. @naoyam what do you think about ` add a feature to select a subset of benchmarks based on the scheduler`?

How would such filtering work? How can we know upfront which inputs will use a given scheduler?

PR #2038, opened 2024-04-03
Torch baselines for attention benchmarks

Priya2698:
Issue #1668

The benchmarks use `dropout_p=0.2` for benchmarking. For validating in forward pass, to ensure the same dropout_mask is applied, this is set to 0.

### DIFF
--- a/benchmarks/python/test_huggingface_attn_bwd.py
+++ b/benchmarks/python/test_huggingface_attn_bwd.py
@@ -56,6 +56,8 @@ def huggingface_attn_bwd_fusion(
 
 
 def huggingface_attn_bwd_iobytes(size: tuple, dtype: torch.dtype):
+    # Manual IOByte computation is required since nvFuser input/outputs (grad_out, attn, dropout_mask, grad_input]) differ from baseline input/outputs (output, grad_output).
+
     # Total IO bytes = grad_output ([bs, nh, seq_len, seq_len], dtype) + attn ([bs*nh, seq_len, seq_len], dtype) + dropout_mask ([bs*nh, seq_len, seq_len], bool) + grad_input ([bs, nh, seq_len, seq_len], dtype)
     bs, seq_len, nh, n_embd = size
     return int(bs * nh * seq_len * seq_len * (dtype.itemsize * 3 + torch.bool.itemsize))
@@ -74,7 +76,7 @@ def test_huggingface_attn_bwd_nvf_benchmark(
 
     batch_size, seq_len, nh, n_embd = size
 
-    dropout_p = 0.2 
+    dropout_p = 0.0 
     inputs = torch.randn(
         batch_size, nh, seq_len, seq_len, device="cuda", dtype=dtype, requires_grad=True
     )
@@ -117,7 +119,7 @@ def test_huggingface_attn_bwd_baseline_benchmark(
 
     batch_size, seq_len, nh, n_embd = size
     dropout_p = 0.2
-    inputs = torch.randn(batch_size, nh, seq_len, seq_len, device="cuda", dtype=dtype)
+    inputs = torch.randn(batch_size, nh, seq_len, seq_len, device="cuda", dtype=dtype, requires_grad=True)
     attention_mask = torch.zeros(
         batch_size, nh, seq_len, seq_len, device="cuda", dtype=dtype
     )
--- a/benchmarks/python/test_huggingface_attn_bwd.py
+++ b/benchmarks/python/test_huggingface_attn_bwd.py
@@ -76,7 +76,7 @@ def test_huggingface_attn_bwd_nvf_benchmark(
 
     batch_size, seq_len, nh, n_embd = size
 
-    dropout_p = 0.0 
+    dropout_p = 0.0
     inputs = torch.randn(
         batch_size, nh, seq_len, seq_len, device="cuda", dtype=dtype, requires_grad=True
     )
@@ -98,7 +98,7 @@ def test_huggingface_attn_bwd_nvf_benchmark(
 
     if not disable_validation:
         # Use dropout_mask instead of torch.nn.functional.dropout for validating results.
-        out = 1 / (1 - dropout_p) * dropout_mask * attn
+        out = torch.nn.functional.dropout(attn, p=dropout_p)
         out.backward(grads)
         fd.validate([grads, attn, dropout_mask], [inputs.grad])
 
@@ -118,7 +118,7 @@ def test_huggingface_attn_bwd_baseline_benchmark(
     clear_cuda_cache()
 
     batch_size, seq_len, nh, n_embd = size
-    dropout_p = 0.2
+    dropout_p = 0.0
     inputs = torch.randn(batch_size, nh, seq_len, seq_len, device="cuda", dtype=dtype, requires_grad=True)
     attention_mask = torch.zeros(
         batch_size, nh, seq_len, seq_len, device="cuda", dtype=dtype
--- a/benchmarks/python/test_huggingface_attn_bwd.py
+++ b/benchmarks/python/test_huggingface_attn_bwd.py
@@ -76,7 +76,7 @@ def test_huggingface_attn_bwd_nvf_benchmark(
 
     batch_size, seq_len, nh, n_embd = size
 
-    dropout_p = 0.0
+    dropout_p = 0.2
     inputs = torch.randn(
         batch_size, nh, seq_len, seq_len, device="cuda", dtype=dtype, requires_grad=True
     )
@@ -98,7 +98,7 @@ def test_huggingface_attn_bwd_nvf_benchmark(
 
     if not disable_validation:
         # Use dropout_mask instead of torch.nn.functional.dropout for validating results.
-        out = torch.nn.functional.dropout(attn, p=dropout_p)
+        out = attn * dropout_mask * 1 / (1 - dropout_p)
         out.backward(grads)
         fd.validate([grads, attn, dropout_mask], [inputs.grad])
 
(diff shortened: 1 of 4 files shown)


naoyam (approved):
LGTM

benchmarks/python/test_huggingface_attn_bwd.py
```diff
@@ -1,7 +1,7 @@
 import pytest
 from nvfuser import FusionDefinition, DataType
 from nvfuser.pytorch_utils import torch_dtype_to_nvfuser_dtype
-from .core import run_benchmark, clear_cuda_cache
+from .core import run_benchmark, clear_cuda_cache, unary_bwd_torch
```
liqiangxl, line 4: where does this `unary_bwd_torch` defined? It doesn't exist in `core.py`.

benchmarks/python/test_huggingface_attn_bwd.py
```diff
@@ -95,3 +101,34 @@ def test_huggingface_attn_bwd_benchmark(
...
+    attn = torch.nn.functional.softmax(attn, dim=-1)
+    output = torch.nn.functional.dropout(attn, p=dropout_p)
+
+    grads = torch.randn(batch_size * nh, seq_len, seq_len, device="cuda", dtype=dtype)
+
+    run_benchmark(
+        benchmark,
+        torch.compile(unary_bwd_torch) if compile else unary_bwd_torch,
+        [output, grads],
+        iobytes=huggingface_attn_bwd_iobytes(size, dtype),
```
liqiangxl, line 133: can you comment why we need to set `iobytes`? what's the difference is using default.

benchmarks/python/test_huggingface_attn_fwd.py
```diff
@@ -71,9 +71,26 @@ def huggingface_attn_fwd_fusion(
     fd.add_output(T47)
 
 
+def huggingface_attn_fwd_fn(
```
liqiangxl, line 74: do we have references for this and all the other baseline implementations?

benchmarks/python/test_huggingface_attn_fwd.py
```diff
@@ -103,3 +120,28 @@ def test_huggingface_attn_fwd_benchmark(
...
+def test_huggingface_attn_fwd_baseline_benchmark(
+    benchmark,
+    size: tuple,
+    dtype: torch.dtype,
+    compile: bool,
+):
+    clear_cuda_cache()
+
+    batch_size, seq_len, nh, n_embd = size
+    dropout_p = 0.0
```
liqiangxl, line 137: is there an influence on perf if dropout is larger than 0, e.g. 0.2? If has, we can use a more practical value and set to 0 if validation is turned on.

liqiangxl: If its `0`, the predication is trivial and may influence the perf.
```
   S9 = fd.define_scalar(1 - dropout_p, dtype=DataType.Double)
    T10 = fd.ops.lt(T8, S9)
```

benchmarks/python/test_huggingface_attn_bwd.py
```diff
@@ -55,9 +55,15 @@ def huggingface_attn_bwd_fusion(
     fd.add_output(T35)
 
 
+def huggingface_attn_bwd_iobytes(size: tuple, dtype: torch.dtype):
```
jacobhinkle, line 58: tiny nit: when annotating collections, specify the element types. For tuple, annotate all of them like `size: tuple[int, int, int, int]`.

jacobhinkle: As @liqiangxl already mentioned, if this just implements the default behavior i.e. `sum(i.nbytes for i in inputs) + sum(o.nbytes for o in outputs)` then we should remove this helper.

benchmarks/python/test_huggingface_attn_bwd.py
```diff
@@ -95,3 +101,34 @@ def test_huggingface_attn_bwd_benchmark(
...
     if not disable_benchmarking:
         run_benchmark(benchmark, fd.execute, [grads, attn, dropout_mask])
+
+
+@pytest.mark.parametrize("compile", [False, True], ids=["eager", "compile"])
+@pytest.mark.parametrize("size", generate_attn_inputs())
+@pytest.mark.parametrize("dtype", FLOAT_DTYPES)
+def test_huggingface_attn_bwd_baseline_benchmark(
+    benchmark,
+    size: tuple,
```
jacobhinkle, line 111: Nit: `size: tuple[int, int, int, int]`

benchmarks/python/test_huggingface_attn_fwd.py
```diff
@@ -71,9 +71,26 @@ def huggingface_attn_fwd_fusion(
     fd.add_output(T47)
 
 
+def huggingface_attn_fwd_fn(
+    inputs: list,
+):  # [attention_mask, in_tensor, size, dropout_p]
```
jacobhinkle, line 76: Comment should be a docstring if providing more details, or omitted since it's clear from the following line.

benchmarks/python/test_huggingface_attn_fwd.py
jacobhinkle, line 74: Might I suggest
```suggestion
def huggingface_attn_fwd(
```
And related changes further down in this file.

benchmarks/python/test_huggingface_attn_bwd.py
Priya2698: It is defined here: https://github.com/NVIDIA/Fuser/blob/e10c4714aede47a41a1d71a879681d3f94c00df0/benchmarks/python/core.py#L142

benchmarks/python/test_huggingface_attn_bwd.py
Priya2698: This is for uniformity of IObytes between nvFuser and torch when the inputs/outputs differ between the two. See the discussion at: https://github.com/NVIDIA/Fuser/pull/1725. I'll add a reference to this PR in a comment.

benchmarks/python/test_huggingface_attn_fwd.py
Priya2698: I'll add links to the thunder code: They were removed earlier. 
Is that what you're referring to?

benchmarks/python/test_huggingface_attn_fwd.py
Priya2698: Right. Let me make that change. Thanks!

benchmarks/python/test_huggingface_attn_bwd.py
Priya2698: This is for uniformity of IObytes between nvFuser and torch when the inputs/outputs differ between the two. See the discussion at: https://github.com/NVIDIA/Fuser/pull/1725. I'll add a reference to this PR in a comment.

Priya2698: > tiny nit: when annotating collections, specify the element types. For tuple, annotate all of them like `size: tuple[int, int, int, int]`.

I think for size, the element types are understood. So I prefer this to reduce verbosity. Not a strong preference, though. I may have a bias being more familiar with the benchmarks.

jacobhinkle: Does the IOBytes issue affect this computation though? It seems like the helper method is just replicating default behavior. If that's the case, then maybe putting argument `iobytes=None,` would be the clearest. As you wish on the tuple annotation.

Priya2698: The nvFuser input/outputs are grad_out, attention_mask, dropout_mask and grad_input.
The eager input/outputs are grad_out and output. So there will be a difference in the computed IObytes.

jacobhinkle: Oh ok, so it does affect this case due to recomputation. Could you place a comment in the code describing that? Thanks.

benchmarks/python/test_huggingface_attn_fwd.py
Priya2698: This would although mean that we benchmark different things based on whether validation is turned on.

liqiangxl: can you check the influence on bandwidth? I am not sure how much it may have, maybe very small.

Priya2698: What do you think about having another fusion definition with 0.0 dropout in validation and 0.2 otherwise. That would ensure we always benchmark with dropout (0.2) and it is consistent with torch baseline benchmarks.

Priya2698: Dropout: 0.2
Screenshot 2024-04-04 at 11 53 57 AM

Dropout: 0.0
Screenshot 2024-04-04 at 11 53 49 AM

The difference for eager is significant.

liqiangxl: > What do you think about having another fusion definition with 0.0 dropout in validation and 0.2 otherwise. That would ensure we always benchmark with dropout (0.2) and it is consistent with torch baseline benchmarks.

sounds good to me.

liqiangxl: > Dropout: 0.2 Screenshot 2024-04-04 at 11 53 57 AM
> 
> Dropout: 0.0 Screenshot 2024-04-04 at 11 53 49 AM
> 
> The difference for eager is significant.

This seems difficult to explain. can you create an issue to investigate further when we have time?

jacobhinkle: Eager mode is probably skipping the dropout call altogether when `dropout=0` but i would guess torch.compile is either not skipping dropout, or it's just io-bound so skipping dropout makes no difference.

liqiangxl: that's very smart!

Priya2698:
Updated the PR description.
We use a non-zero dropout in benchmarking now. For fwd calls, I set it to 0 for validation.

Discovered a bug when using non-zero dropout in nanogpt bwd pass that is now fixed.

PR #2039, opened 2024-04-04
Always validate dynamic smem usage for matmuls

jacobhinkle:
This introduces a mechanism for schedulers to declare the expected dynamic shared memory usage by optionally calling
`fusion->setExpectedDynamicSmemBytes(n)`. During execution, when validating that the shared memory fits within device limits, we will also check that the dynamic smem matches this value.

This is debugging information, but since it's quick to check I have enabled it unconditionally. This will give us more confidence that `mma_utils::computeExpectedSharedMemoryUsage` is accurate. If that function is inaccurate, we would not necessarily hit an error but we might have suboptimal heuristics and occasional errors where we specify MatmulParams that cannot be launched on the current device.

Note that the smem is dynamic, so it could instead be a `Val*`, but it is compile-time constant in the case of matmul. If needed, we can extend this to be a `Val*` that we evaluate at execution, but there might be performance implications in such cases.

Also note that static smem (which is compile-time constant) is not checked currently. We could of course handle that in a similar manner if needed.

### DIFF
--- a/csrc/executor.cpp
+++ b/csrc/executor.cpp
@@ -1674,6 +1674,17 @@ void FusionExecutor::validateDynamicSmemSize(int64_t dynamic_smem_size) {
       getStaticSmemSize() + dynamic_smem_size,
       ". Device limit size: ",
       device_smem_limit_);
+  // If specified, check that dynamic smem size matches what the scheduler
+  // expects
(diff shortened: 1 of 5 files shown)


jacobhinkle:
This has surfaced some inaccuracies in internal testing, which I will fix in a separate PR.

zasdfgbnm (approved):

PR #2041, opened 2024-04-04
Documentation of TMA: part 2

zasdfgbnm:
This PR does the following:
- Added a few more texts to `doc/dev/tma.md` describing how to create box.
- Largely extents `doc/reading/divisibility-of-split.md` discussing indivisible splits.
- Added `doc/reading/tma-modeling-in-depth.md` discussing deeper and more formally about predication and correctness of TMA.

Suggested order of reading:
1. https://github.com/NVIDIA/Fuser/blob/tma-step-2/doc/reading/divisibility-of-split.md
2. https://github.com/NVIDIA/Fuser/blob/tma-step-2/doc/dev/tma.md
3. https://github.com/NVIDIA/Fuser/blob/tma-step-2/doc/reading/tma-modeling-in-depth.md


doc/reading/tma-modeling-in-depth.md
```diff
@@ -0,0 +1,153 @@
...
+- For an unpredicated TMA expression, if we do not initialize the buffer, what correctness model can we achieve?
+- In the case where strong correctness is needed, but we have only achieved weak correctness,
+  is there any way to upgrade to strong correctness by doing something?
+
+As we can see from CodeBlock 1, TMA has builtin predicates checking that the indices of all partitioned IterDomains are in bound.
+That is, TMA will never do out-of-boundary access on global memory even if the indices of
+some IterDomains may be out of boundary.
+Therefore, we have:
+
+**Theorem 1:** TMA provides weak correctness.
```
zasdfgbnm, line 98: Currently, we eliminated all predicates of TMA, and not initializing the output smem buffer. So we are assuming weak correctness. I believe in the future, we need to check the correctness model we need based on the fusion.

doc/reading/divisibility-of-split.md
zasdfgbnm, line 1: I have been thinking about making our ID modeling more formal and mathematically sound, I think a good starting point is to make the divisibility of split and TMA formal.

doc/dev/tma.md
```diff
@@ -72,6 +72,28 @@ Instead, it is a virtual domain that only exists in the user's mind.
...
+
+#### Define box by partitioning
+
+Defining box by partitioning is as simple as: select an IterDomain in the TMA domain, then
+inner split that IterDomain by the box size of that dimension.
+
+We call this split expression a "*boxing split*", the input of this split a "*partitioned IterDomain*",
+the inner output of this split a "*box IterDomain*", and the outer output of this split a "*coordinate IterDomain*".
+
+TODO: currently "partitioned IterDomain" is called "TMA-global IterDomain" in the code. I need to update the code.
```
zasdfgbnm, line 88: I will update the code later in a separate PR

doc/reading/divisibility-of-split.md
```diff
@@ -4,24 +4,249 @@
...
+$$i_2 / N < P$$
+According to "Rule 1" in `[Simplification of boolean predicates]` in `csrc/expr_simplifier.h`,
+(TODO: move this theorem to a md file)
+$$i_2 / N < P \Leftrightarrow i_2 < Q$$
+$\square$
+
+</details>
+
+**Theorem 3** Suppose that there is a resize `I1 = Resize(I0, L, R)`.
+Then "the index of `I0` is in bound" implies "the index of `I1` is in bound" if $R >= 0$.
```
naoyam, line 158: When you say `I0` is in bound, does that also mean `I0 >= -L`?

zasdfgbnm: Good catch! So far I have assumed that we only need to predicate upper bound, and all integers are automatically >= 0. Looks like this may not be the case for `Resize`. Let me thinking through it.

doc/reading/divisibility-of-split.md
```diff
@@ -4,24 +4,249 @@
...
-Indivisible splits has many interesting properties that deserve us thinking.
+> [!NOTE]
+> We use $\div$ for true division, and $/$ for integer division. For example, $5\div 2 = 2.5$, $5/2=2$.
 
-Let's begin the discussion with a simple question:
+## Introduction
 
-If I have a tensor `T0[I1, I2]`, are the following two schedules equivalent?
+In nvFuser, `Split` is an IterDomain expression that partitions the original IterDomain into nested sub-IterDomains,
+where the outer IterDomain iterates over the quotient and the inner IterDomain iterates over the remainder of the original extent divided by the split factor.
```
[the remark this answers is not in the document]
zasdfgbnm: Rephrased this part at https://github.com/NVIDIA/Fuser/pull/2041/commits/473b59b3f5591609d30cb44073fde1ed1d61d7e4

doc/reading/divisibility-of-split.md
zasdfgbnm: I resolved this in https://github.com/NVIDIA/Fuser/pull/2060, it does make it more complicated, but the conclusion regarding TMA does not change.

doc/reading/tma-modeling-in-depth.md
```diff
@@ -0,0 +1,153 @@
...
+Strong correctness means that, when an IterDomain expression create holes (indivisible split, resize),
+the holes must be filled with a certain value.
+
+Understanding strong correctness of TMA requires the following definition:
+
+**Definition 1 (TMA-protected IterDomain):** An IterDomain is TMA-protected if it satisfies any of the following conditions:
+
+1. It is a partitioned IterDomain.
+2. It is the outer output of a split of a TMA-protected IterDomain.
+3. It is the output of a merge whose outer input is a TMA-protected IterDomain.
```
naoyam, line 122: Could you give an example where the inner input is not protected but the output is protected due to the protected outer input?

doc/dev/tma.md
zasdfgbnm, line 1: Note to myself: create index for all documentations.

doc/reading/tma-modeling-in-depth.md
zasdfgbnm: See my latest commit: https://github.com/NVIDIA/Fuser/pull/2041/commits/036a495f10836e9e984f7c9ad43c730e57a28d79

zasdfgbnm: I think the analysis you are referring to is exactly the analysis described here. In this language, `I1` is TMA-protected, but `I2` is not, so `I2` needs to be predicated if there is an indivisible on it.

Being `TMA-protected` has nothing to do with whether it can go out of boundary or not. They are orthogonal concepts. Being `TMA-protected` means that, if it can go out of boundary, then it will be protected by the hardware.

|                        | TMA-protected  | not TMA-protected    |
| ---------------------- | -------------- | -------------------- |
| **can go out of bound**    | no predication | should be predicated |
| **cannot go out of bound** | no predication | no predication       |

zasdfgbnm: Added a new commit trying to explain why we want this definition: https://github.com/NVIDIA/Fuser/pull/2041/commits/0959cf0dd90e8fe0b9350df4614b3b7f44c961ea

naoyam: Thanks. That makes sense.

doc/reading/divisibility-of-split.md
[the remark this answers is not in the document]
zasdfgbnm: See https://github.com/NVIDIA/Fuser/pull/2041/commits/3846d486872c4a8b8a1f788042f59626ee148a36,
I unified the terms "data", "in-boundary items", and "valid items" into "valid items" and mentioned what it is in the beginning. I also changed a few wording that I think will make it clearer.

naoyam (approved):
Thanks for the clarifications. LGTM.

PR #2044, opened 2024-04-05
Disable failing multidevice unit tests

cowanmeg:
Disabling failing multidevice unit tests until they are fixed.

1. PipelineTestTwoStage test with UCC backend.
`mpirun -np N test_multidevice_test --gtest_filter='*DISABLED_UCC_*' --gtest_also_run_disabled_tests` where N > 4
These tests cause UCC barrier to hang when the tests are run with more than 4 processes. Still investigating the cause and whether the bug is potentially in UCCProcessGroup. 

2. Distributed reduction tests that execute with FusionExecutorCache. 
`mpirun -np 4 test_multidevice_test --gtest_filter='*DISABLED_FusionExecutorCache_*' --gtest_also_run_disabled_tests`
These tests are failing because of the new `AllocationDomainPass`. @jjsjann123  Hopefully, this is a simple fix to ignore reordering reductions on a DeviceDim axis.

### DIFF
--- a/tests/cpp/test_multidevice_pipeline.cpp
+++ b/tests/cpp/test_multidevice_pipeline.cpp
@@ -302,7 +302,7 @@ INSTANTIATE_TEST_SUITE_P(
         testing::Values(false),
         testing::Values(true),
         testing::Values(0, 1),
-        testing::Bool()));
+        testing::Values(false)));
 
 INSTANTIATE_TEST_SUITE_P(
     ReduceScatter,
@@ -315,8 +315,38 @@ INSTANTIATE_TEST_SUITE_P(
         testing::Values(true),
         testing::Values(true),
         testing::Values(0, 1),
-        testing::Bool()));
+        testing::Values(false)));
 
+// TODO: Distributed reduction tests using fusion executor cache are failing
+// AllocationDomainPass might be re-ordering compute
+INSTANTIATE_TEST_SUITE_P(
+    DISABLED_FusionExecutorCache_Reduce,
+    PipelineTestTwoStages,
+    testing::Combine(
+        testing::Values(CommunicatorBackend::nccl),
+        all_nontrivial_meshes,
+        all_meshes,
+        testing::Values(true),
+        testing::Values(false),
+        testing::Values(true),
+        testing::Values(0, 1),
+        testing::Values(true)));
+
+INSTANTIATE_TEST_SUITE_P(
+    DISABLED_FusionExecutorCache_ReduceScatter,
+    PipelineTestTwoStages,
+    testing::Combine(
+        testing::Values(CommunicatorBackend::nccl),
+        all_nontrivial_meshes,
+        testing::Values(mesh_null), // the same mesh is used for all tensors
+        testing::Values(true),
+        testing::Values(true),
+        testing::Values(true),
+        testing::Values(0, 1),
+        testing::Values(true)));
+
+// TODO: UCC PipelineTestTwoStages are hanging in UCC barrier
+// when number of processes > number of gpus required by test.
 INSTANTIATE_TEST_SUITE_P(
     DISABLED_UCC_Gather,
     PipelineTestTwoStages,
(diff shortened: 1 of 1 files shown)


cowanmeg:
!build --dist

jjsjann123:
> These tests are failing because of the new AllocationDomainPass

Is this the new pass where it's propagating allocation order to outputs? to work around that, you can either: 1. disable the pass for your test; *OR*, 2. set allocation order on outputs in your test.

tests/cpp/test_multidevice_pipeline.cpp
```diff
@@ -295,20 +295,141 @@ INSTANTIATE_TEST_SUITE_P(
...
+        all_nontrivial_meshes,
+        testing::Values(mesh_null), // the same mesh is used for all tensors
+        testing::Values(true),
+        testing::Values(true),
+        testing::Values(true),
+        testing::Values(0, 1),
+        testing::Values(false)));
+
+// TODO: Distributed reduction tests using fusion executor cache are failing
+// AllocationDomainPass might be re-ordering compute
```
jjsjann123, line 321: We can disable the optimization pass here like here. This is probably the easiest approach
https://github.com/NVIDIA/Fuser/blob/cf63eb6b6cc1d91aa6bdbfb0e3ac729ecebb5e58/tests/cpp/test_matmul_scheduler.cpp#L28

jjsjann123 (approved):
stamp for cleaning up CI

tests/cpp/test_multidevice_pipeline.cpp
jjsjann123: I think a *cleaner* approach is to specify allocation domain on output. Alternatively, if we think any computation coming in with a device mesh is too sensitive to be touched by optimization passes, I should probably also skip those in my optimization passes.

I think similar thing would apply to alias analysis as well. wondering if @wujingyue 's pass would do the same.

xwang233:
!build --dist

xwang233:
Seems like the distributed job wasn't setup correctly. Let me try it again.

tests/cpp/test_multidevice_pipeline.cpp
cowanmeg: Thanks for the pointers! I imagine for a while the multidevice tests will be sensitive to many optimizations until we are more stable. It might be better for us to disable optimization passes on the multidevice test. Later we can make the pass composable with multidevice. 

I think @naoyam mentioned he is setting the allocation domain for all tensors as part of index refactoring. Ideally, this will help make things more robust.

jjsjann123: > I think @naoyam mentioned he is setting the allocation domain for all tensors as part of index refactoring. Ideally, this will help make things more robust.

Yeah that's indeed useful. Though I'm imagining that is going to happen after the initial optimization pass... Conceptually I think it makes sense (at least for allocation order pass) to avoid touching fusion related to multi device. Given that we know such constraints applied to tensors involved with multidevice at the time. I'll try to patch that.

xwang233 (approved):
The distributed test passed and all looks good!

cowanmeg:
> The distributed test passed and all looks good!

🎉 Thanks again!!

tests/cpp/test_multidevice_pipeline.cpp
wujingyue: > I think a cleaner approach is to specify allocation domain on output. 

Correct. I think [insertShardingAxisReordering](https://github.com/NVIDIA/Fuser/blob/cf63eb6b6cc1d91aa6bdbfb0e3ac729ecebb5e58/csrc/multidevice/utils.cpp#L257) should set allocation domain explicitly. An empty allocation domain means a wildcard. 

> I think it makes sense (at least for allocation order pass) to avoid touching fusion related to multi device. 

That should yield to the correct solution, IMO. AllocationOrderPass and MarkAliasesPrepare are and will be still useful for non-DIDs. Wdyt?

cowanmeg: Yes I think these passes are still useful. Setting the allocation domain insertShardingAxisReordering is a great suggestion. I'll find some time to add this.

PR #2049, opened 2024-04-05
Matmul heuristic plugin interface

jacobhinkle:
This change introduces a plugin interface for user-provided matmul heuristics. To use it, set the environment variable `$NVFUSER_MATMUL_HEURISTIC_PLUGIN` to the location of a dynamic library to use for matmul heuristic computation (this variable can also be the base filename of a library in `$LD_LIBRARY_PATH`).  See `examples/matmul_heuristic_plugin` for a simple example.

TODO:
- build example plugin as part of CI


jacobhinkle:
!build

jacobhinkle:
!build

csrc/scheduler/matmul_utils.cpp
```diff
@@ -393,12 +394,36 @@ std::shared_ptr<MatmulParams> getMatmulHeuristics(
...
-  // Set whether to use shared memory for epilogue
   const auto& roles_map_opt =
       mma_utils::getTensorsRoles(fusion, mulSum.front().insouts);
   NVF_ERROR(roles_map_opt.isValid(), "Tensor roles map in mma is not valid.");
-
   const auto roles_map = roles_map_opt.getData();
+
+  const mma_utils::MatmulProblemLayoutOpt layout_opt =
+      mma_utils::getMmaLayout(fusion);
+  NVF_ERROR(layout_opt.isValid(), layout_opt.getErrorMsg());
```
jacobhinkle, line 404: This fails for `CombineMulSumAsMmaTest.UseMatmulScheduler` currently. I'm looking into it.

csrc/scheduler/matmul_utils.cpp
```diff
@@ -393,12 +394,34 @@ std::shared_ptr<MatmulParams> getMatmulHeuristics(
...
 
-  // Set whether to use shared memory for epilogue
   const auto& roles_map_opt =
       mma_utils::getTensorsRoles(fusion, mulSum.front().insouts);
   NVF_ERROR(roles_map_opt.isValid(), "Tensor roles map in mma is not valid.");
-
   const auto roles_map = roles_map_opt.getData();
+
+  const mma_utils::MatmulProblemLayoutOpt layout_opt =
+      mma_utils::getMmaLayout(fusion, mulSum.front().insouts);
```
jacobhinkle, line 403: @protonu I added the second argument to `getMmaLayout`. It's not strictly related to this PR but I think it was wrong before and since I needed the layout for scheduling, I had to fix it. Can you confirm this is the correct pattern?

jacobhinkle:
This is working for C++ programs but I currently hit a segfault upon program termination with python scripts, e.g. `NVFUSER_MATMUL_PLUGIN=examples/matmul_heuristic_plugin/build/libmatmul_heuristic_plugin.so python -m pytest tests/python/test_python_frontend.py -k matmul`. Perhaps we need to link the plugin against `libnvfuser.so` in addition to `libnvfuser_codegen.so`...

csrc/scheduler/matmul_utils.cpp
protonu: That looks like the correct usage.
https://github.com/NVIDIA/Fuser/blob/52692ed5347dc308b93a7f35acdf2c4123c10c1f/csrc/scheduler/matmul_utils.cpp#L361-L368

jacobhinkle: Right, that's what I was looking at. I just wanted to check that I was not modifying intentionally different behavior here.

csrc/scheduler/matmul_heuristic_plugin.h
```diff
@@ -0,0 +1,77 @@
...
+
+namespace nvfuser {
+
+namespace matmul_heuristic_plugin {
+
+//! This is the information available to the plugin to determine the kernel
+//! configuration. For API stability, these should not be accessed directly when
+//! implementing plugins, but rather through these accessors.
+//! matmul_heuristic_plugin_api.h
+struct NVF_API ProblemDescription;
```
zasdfgbnm, line 24: Why are we duplicating code in `matmul_heuristic_plugin_api.h`? Is it for `NVF_API`? Does it has to be `NVF_API`?

csrc/scheduler/matmul_utils.cpp
```diff
@@ -393,12 +394,34 @@ std::shared_ptr<MatmulParams> getMatmulHeuristics(
...
+  NVF_ERROR(layout_opt.isValid(), layout_opt.getErrorMsg());
+  const MmaLayout layout = layout_opt.getData();
+
+  if (matmul_heuristic_plugin::hasPlugin()) {
+    // Fill in proper values using plugin
+    matmul_heuristic_plugin::updateMatmulParams(
```
zasdfgbnm, line 409: Do we still need the `initCoreHeuristics` above?

csrc/scheduler/matmul_heuristic_plugin_api.h
```diff
@@ -0,0 +1,76 @@
...
+//!  T = BFloat16
+//!  H = Float16
+//!  F = TensorFloat32
+//!  S = Float32
+//!  D = Float64
+//!  C = complex<float>
+//!  Z = complex<double>
+const char* getProblemPrecision(const ProblemDescription* problem);
+
+//! This is the return type of a HeuristicFunc (defined below) implemented in a
```
zasdfgbnm, line 55: `HeuristicFunc` is not defined below?

csrc/scheduler/matmul_heuristic_plugin.cpp
```diff
@@ -0,0 +1,287 @@
...
+
+    NVF_ERROR(config != nullptr);
+    NVF_ERROR(problem != nullptr);
+
+    (*func_)(config, problem);
+  }
+
+ private:
+  char* filepath_ = nullptr;
+  void* handle_ = nullptr;
```
zasdfgbnm, line 72: nit:
```suggestion
  void* library_handle_ = nullptr;
```

csrc/scheduler/matmul_heuristic_plugin.cpp
```diff
@@ -0,0 +1,287 @@
...
+namespace nvfuser {
+
+namespace matmul_heuristic_plugin {
+
+namespace {
+
+//! Defines HeuristicFuncPtr as type of the "getConfig" symbol
+typedef void (*HeuristicFuncPtr)(KernelConfig*, const ProblemDescription*);
+
+thread_local class PluginInterface {
```
zasdfgbnm, line 27: Should this inherit this class?
https://github.com/NVIDIA/Fuser/blob/618c4b79a053692bf83ebb32aa19dcc907ebb534/csrc/utils.h#L90-L103

csrc/scheduler/matmul_heuristic_plugin.cpp
```diff
@@ -0,0 +1,287 @@
...
+  params.double_buffer_options.double_buffer_smem_write =
+      config.load_stages > 1;
+
+  // async load only for circular buffering (stages > 2)
+  params.async_gmem_load_operands = config.load_stages > 2;
+
+  return true;
+}
+
+} // namespace matmul_heuristic_plugin
```
zasdfgbnm, line 285: Shouldn't all this belong to the plugin? That is, nvfuser does not assume the structure of `KernelConfig` and `ProblemDescription`, it just treat them as opaque data. The plugin should provide API to create and free these objects, and get/set data of these objects, and nvFuser just call these APIs without knowing anything about the internal representation of such objects.

csrc/scheduler/matmul_heuristic_plugin.cpp
zasdfgbnm, line 27: Does this mean we will initialize an object everytime we create a thread?

csrc/scheduler/matmul_heuristic_plugin.h
jacobhinkle: Yes. It needs to be NVF_API and it's convenient to keep the includes minimal in the external header. I tried to be really conservative at first but I could probably include `visibility.h` without problems and just make these `NVF_API` in the external header. I'll try it.

jacobhinkle: If it's not NVF_API then we can have missing symbols when importing the plugin. I hit this in an early version.

csrc/scheduler/matmul_heuristic_plugin.cpp
jacobhinkle: Good idea. If it gets copied we could call dlclose twice, which would be fine if we also called dlopen twice since dlopen is reference counted but the copy constructor would not call dlopen. I'll make the change.

jacobhinkle: Yes we will have one of these objects per thread, but they are pretty light-weight. Without thread_local I was hitting segfaults at dlclose when running pytest. The alternative is to guard the dlopen and dlclose parts with `std::lock_guard`; I'll try that out.

csrc/scheduler/matmul_heuristic_plugin.cpp
jacobhinkle: Good point. I think there's a way to address this along with your other comment about code duplication. We could get rid of all the getters and setters by just declaring a struct with a virtual `configure` method in the api header. Then the plugin should implement a subclass of that class and implement `configure`. That can be private to the plugin i.e. opaque to nvfuser; they only need to export a factory function that creates one of these objects. I think that will simplify the nvfuser code quite a bit without burdening the plugin developers.

csrc/scheduler/matmul_heuristic_plugin.h
jacobhinkle: Code is no longer duplicated. Plugins subclass a class that defines the problem and config.

csrc/scheduler/matmul_heuristic_plugin.cpp
jacobhinkle: Switched to keeping `plugin` global and using a mutex to guard whenever we modify it.

jacobhinkle:
!build

csrc/scheduler/matmul_utils.cpp
jacobhinkle: Moved this call so that we only run it if we don't specify a plugin (after the warning).

jacobhinkle:
After the recent changes to the interface, the python test `test_python_frontend.py::test_matmul` is now failing with
> RuntimeError: mma_layout == MmaLayout::TN INTERNAL ASSERT FAILED at "/opt/pytorch/nvfuser/csrc/scheduler/matmul.cpp":895, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. MMAs in Turing and Ampere are TN only, transpose is handled either via ldmatrix.trans for fp16 or explicitly for other types. 

I'm looking into why this is happening.

csrc/scheduler/matmul_heuristic_plugin_api.h
```diff
@@ -0,0 +1,90 @@
...
+
+namespace nvfuser {
+
+namespace matmul_heuristic_plugin {
+
+//! This is intended as a minimal interface for enabling matmul heuristics.
+//! In order to plug in your own custom heuristic, create a dynamic library
+//! defining a subclass of KernelConfig, overriding the `configure` method. This
+//! class does not need to be exported from the dll, but you should export a
+//! std::unique_ptr<KernelConfig> makeConfig() function that returns a
```
zasdfgbnm, line 22: `examples/matmul_heuristic_plugin/matmul_heuristic_plugin.cpp` is `KernelConfig* makeConfig()` instead of using `unique_ptr`. Also, should we declare this function in this header?

csrc/scheduler/matmul_heuristic_plugin_api.h
```diff
@@ -0,0 +1,90 @@
...
+//! unique_ptr to an object of that type. The `configure` method will be called
+//! on that object by nvfuser in order to fill the correct values in the class.
+//!
+//! If that library is located at /path/to/libfoo.so you can set
+//! NVFUSER_MATMUL_HEURISTIC_PLUGIN=/path/to/libfoo.so to use the plugin to
+//! determine matmul parameters automatically.
+
+//! This is the return type of a HeuristicFunc (defined below) implemented in a
+//! plugin. This is used to set values in MatmulParams
+struct KernelConfig {
```
zasdfgbnm, line 32: Possibly related doc for subclassing a C++ class in python:

https://pybind11.readthedocs.io/en/stable/advanced/classes.html

csrc/scheduler/matmul_heuristic_plugin.cpp
```diff
@@ -0,0 +1,190 @@
...
+          dlerror());
+    }
+
+    return (*factory_func_ptr_)();
+  }
+
+ private:
+  char* filepath_ = nullptr;
+  void* library_handle_ = nullptr;
+  ConfigFactoryFuncPtr factory_func_ptr_ = nullptr;
```
zasdfgbnm, line 78: Should we pull this out of `PluginInterface` and add a setter `setMatmulConfigFactory` so that we can provide the factory function from a different source? For example, in a unit test, we can
```C++
TEST_F(NVFuserTest, TestMatmulHeuristicsPlugin){ 
  struct MyTestPlugin : public matmul_heuristic_plugin::KernelConfig {
    ....
    void configure() override {
      EXPECT_TRUE("the problem desc we get matches with what we set");
      set kernel config
    }
  };

  auto f = [&executed]() { return new MyTestPlugin() };
  setMatmulConfigFactory(f);
  ...
  EXPECT_TRUE("the kernel config we get matches with what we set");
}
```
we can even set the factory function as a python function.

csrc/scheduler/matmul_heuristic_plugin.cpp
```diff
@@ -0,0 +1,190 @@
...
+      return 2;
+    case MmaLayout::TT:
+      return 3;
+    default:
+      return 255;
+  }
+}
+
+//! Utility to standardize conversion of MmaLayout to uint8_t
+uint8_t layoutToByte(MmaLayout layout);
```
zasdfgbnm, line 98: Is this redundant?

csrc/scheduler/matmul_heuristic_plugin_api.h
jacobhinkle: Thanks. I need to update the comment. I originally made `makeConfig` return a `unique_ptr` in which case it didn't need to be `extern "C"`, but instead now I am returning a raw pointer and converting that to `unique_ptr` on the nvfuser side.

I don't think we can declare the function here since it is the symbol that we will need to load with `dlsym`.

csrc/scheduler/matmul_heuristic_plugin.cpp
jacobhinkle: Oops. Thanks

csrc/scheduler/matmul_heuristic_plugin.cpp
jacobhinkle: Good idea. Actually I wonder an `AbstractPluginInterface` class should be declared public with a virtual `makeConfig`. We could then accept a `AbstractPluginInterface* plugin_interface = nullptr` argument in `updateMatmulParams()`. It would just be a little more work then to implement a test I think:
```c++
TEST_F(NVFuserTest, TestMatmulHeuristicsPlugin){ 
  struct MyTestPlugin : public matmul_heuristic_plugin::KernelConfig {
    ....
    void configure() override {
      EXPECT_TRUE("the problem desc we get matches with what we set");
      set kernel config
    }
  };
  class MyPluginInterface : public AbstractPluginInterface {
    KernelConfig* makeConfig() override {
      return static_cast<KernelConfig*>(new MyTestPlugin);
    }
  } test_plugin;

  updateMatmulParams(..., /*plugin_interface=*/test_plugin);
  ...
  EXPECT_TRUE("the kernel config we get matches with what we set");
}
```

jacobhinkle: I see what you mean about python functions though. With my approach in the last comment, it would be awkward to plumb in a python heuristic, compared to being able to just set it globally as in your example.

zasdfgbnm: Yeah, the point is to have some context specifying that function, and the ability to update this context.

csrc/scheduler/matmul_heuristic_plugin.cpp
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  ~PluginInterface() {
+    if (library_handle_ != nullptr) {
+      std::lock_guard<std::mutex> lock(plugin_mutex);
+      dlclose(library_handle_);
+      library_handle_ = nullptr;
+      factory_func_ptr_ = nullptr;
+    }
+  }
```
zasdfgbnm, line 52: Would be nice to unify this with
https://github.com/NVIDIA/Fuser/blob/1110b0aff240e46604283ac4f9fcfe27bc276ff8/csrc/driver_api.cpp#L19-L47

Maybe move both to this file? https://github.com/NVIDIA/Fuser/blob/main/csrc/sys_utils.cpp

csrc/scheduler/matmul_heuristic_plugin.cpp
```diff
@@ -0,0 +1,210 @@
...
+    MatmulParams& params,
+    int64_t m,
+    int64_t n,
+    int64_t k,
+    int64_t batch_size,
+    MmaLayout layout,
+    const mma_utils::RolesMap& roles_map) {
+  if (!hasPlugin()) {
+    return false;
+  }
```
zasdfgbnm, line 126: This shouldn't return false if `!hasPlugin()` but the user manually specified a function?

csrc/scheduler/matmul_heuristic_plugin.h
```diff
@@ -0,0 +1,62 @@
...
+
+namespace nvfuser {
+
+namespace matmul_heuristic_plugin {
+
+//! Try to load plugin whose location is provided by the environment variable
+//! NVFUSER_MATMUL_HEURISTIC_PLUGIN and return whether or not we succeed.
+bool hasPlugin();
+
+//! If loading the plugin fails (see hasPlugin()) we return false. Otherwise, we
```
zasdfgbnm, line 25: This needs update: https://github.com/NVIDIA/Fuser/pull/2049/files#r1561522002

csrc/scheduler/matmul_heuristic_plugin.h
```diff
@@ -0,0 +1,62 @@
...
+//! value.
+class KernelConfigFactoryGuard {
+ public:
+  explicit KernelConfigFactoryGuard(KernelConfigFactoryPointer func);
+  ~KernelConfigFactoryGuard();
+
+ private:
+  KernelConfigFactoryPointer prev_factory_;
+
+  static thread_local KernelConfigFactoryPointer active_factory_;
```
zasdfgbnm, line 57: This is not used?

examples/matmul_heuristic_plugin/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023-present NVIDIA CORPORATION & AFFILIATES.
+# All rights reserved.
+# SPDX-License-Identifier: BSD-3-Clause
+cmake_minimum_required(VERSION 3.10 FATAL_ERROR)
+project(nvfuser_matmul_heuristic_plugin LANGUAGES CXX)
+
+set(CMAKE_CXX_STANDARD 17)
+
+set(nvfuser_include_dir /opt/pytorch/nvfuser/csrc CACHE FILEPATH REQUIRED)
+set(nvfuser_lib_dir /opt/pytorch/nvfuser/build CACHE FILEPATH REQUIRED)
```
zasdfgbnm, line 10: Building the plugin does not require nvFuser, correct? I assume people can just download the `csrc/scheduler/matmul_heuristic_plugin_api.h` file and use it. If this is the recommended way for people to write plugin, should we not hardcode anything about nvFuser here, and just create a symlink here pointing to `csrc/scheduler/matmul_heuristic_plugin_api.h`?

csrc/scheduler/matmul_heuristic_plugin.h
```diff
@@ -0,0 +1,62 @@
...
+//! object like this:
+//!
+//!   KernelConfigFactoryGuard kfg([]() { return (KernelConfig*)(new
+//!   MyKernelConfig);});
+//!
+//! When kfg passes out of scope, the config factory will be reset to its prior
+//! value.
+class KernelConfigFactoryGuard {
+ public:
+  explicit KernelConfigFactoryGuard(KernelConfigFactoryPointer func);
```
zasdfgbnm, line 51: nit: Should we use `std::function` instead of a function pointer so that it can be a lambda?

https://stackoverflow.com/questions/9054774/difference-between-stdfunction-and-a-standard-function-pointer

https://pybind11.readthedocs.io/en/stable/advanced/cast/functional.html

examples/matmul_heuristic_plugin/CMakeLists.txt
jacobhinkle: Actually, you're right that we can compile by just symlinking the header here but we also need to link the library. If we don't link against `nvfuser_codegen` then we cannot `dlsym` the function as there will be missing symbols. I had thought this would work by finding the symbols in the executable, but it does not work that way apparently. I think ideally we'd install a `FindnvFuser.cmake` somewhere so that we could just `find_package(Nvfuser REQUIRED);` then use `${NVFUSER_INCLUDE_DIR}` and `${NVFUSER_LIBS}` etc. Currently `find_package(Nvfuser)` works but it does not export the installation include or lib dirs or the `nvfuser_codegen` lib as variables (you have to hard-code the lib name). That is probably a topic for another issue.

jacobhinkle: Ah, I figured this out. The `sinh_libtorch` example has a proper cmake setup that should work for this example as well. I was hitting problems with missing headers but that is because I had not re-run `python setup.py install`, which populates the include dir `/opt/pytorch/nvfuser/nvfuser/include/nvfuser/`. I think once we install the package it works as expected.

csrc/scheduler/matmul_heuristic_plugin.cpp
jacobhinkle: Right. Updated

csrc/scheduler/matmul_heuristic_plugin.cpp
jacobhinkle: I took a stab at unifying them into a `LibraryLoader` class in the latest commit.

examples/matmul_heuristic_plugin/CMakeLists.txt
zasdfgbnm: Why does the plugin has to link nvFuser? I am doing experiment here: https://github.com/zasdfgbnm/dlopen-polymorphic-class
seems that it is not required.

Plugins linking with nvFuser could have versioning issues? For example, if I created my `libplugin.so` and link with nvFuser 1.1.1, will nvFuser 1.1.2 be able to use that plugin?

zasdfgbnm (approved):
Left more comments, nothing blocking.

csrc/scheduler/matmul_heuristic_plugin.cpp
```diff
@@ -0,0 +1,235 @@
...
+      config->double_buffer_smem_read;
+  params.rotate_ldmatrix_out_of_main_loop =
+      config->rotate_ldmatrix_out_of_main_loop;
+
+  // enable double buffering or circular buffering if configured
+  params.double_buffer_options.double_buffer_smem_write =
+      config->load_stages > 1;
+
+  // async load only for circular buffering (stages > 2)
+  params.async_gmem_load_operands = config->load_stages > 2;
```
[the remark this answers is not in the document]
jacobhinkle: Yes I was thinking we could avoid any cp.async overhead in these cases. Do you think this is not safe?

csrc/scheduler/matmul_heuristic_plugin.cpp
```diff
@@ -0,0 +1,235 @@
...
+          ". Expected 0 (row-major) or 1 (column-major)");
+  }
+  params.double_buffer_options.double_buffer_smem_read =
+      config->double_buffer_smem_read;
+  params.rotate_ldmatrix_out_of_main_loop =
+      config->rotate_ldmatrix_out_of_main_loop;
+
+  // enable double buffering or circular buffering if configured
+  params.double_buffer_options.double_buffer_smem_write =
+      config->load_stages > 1;
```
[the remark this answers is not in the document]
jacobhinkle: Also when would we choose to disable double buffered reads into registers? I guess a heuristic should be able to set it directly.

csrc/scheduler/matmul_heuristic_plugin.cpp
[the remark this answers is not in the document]
zasdfgbnm: Honestly I don't know.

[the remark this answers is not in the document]
zasdfgbnm: Tried one shape `int M = 50400, N = 13600, K = 248;` with `smem_double_buffer_stage = 2` and different values of `async_gmem_load_operands`. Looks like `true` is faster:

```C++
true:
kernelnone_f0_c0_r0_g0 run in 4.224 ms, achieved: 656.606 GB/s
kernelnone_f0_c0_r0_g0 run in 4.34294 ms, achieved: 638.623 GB/s
kernelnone_f0_c0_r0_g0 run in 4.224 ms, achieved: 656.606 GB/s
kernelnone_f0_c0_r0_g0 run in 4.2711 ms, achieved: 649.365 GB/s


fasle:
kernelnone_f0_c0_r0_g0 run in 4.7913 ms, achieved: 578.863 GB/s
kernelnone_f0_c0_r0_g0 run in 4.34176 ms, achieved: 638.797 GB/s
kernelnone_f0_c0_r0_g0 run in 4.99405 ms, achieved: 555.362 GB/s
kernelnone_f0_c0_r0_g0 run in 4.61312 ms, achieved: 601.221 GB/s
```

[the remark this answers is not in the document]
jacobhinkle: Thanks for testing! That's surprising. Yes, since it's not so straightforward Ill leave it up to the heuristic instead of hardcoding it.

jacobhinkle:
!build

Issue #2032, opened 2024-04-03
Tests for Matmul execution via the EE Evaluator (ATen execution) should cover more than concrete tensors


protonu:
Current tests for ATen execution of matmuls are in:
https://github.com/NVIDIA/Fuser/blob/main/tests/cpp/test_matmul_aten_evaluation.cpp

These tests have good coverage but miss a few things pointed out by Jacob Hinkle.

"Is there a reason to make these concrete? Actually, what are the contiguity requirements for at::matmul and at::addmm? If there are none, then it would be nice to have tests that have discontiguous inputs. Also having some with inputs that have M, K, and N not divisible by 8 would let us test cases with unaligned rows (currently not supported in nvfuser matmul)."

jacobhinkle:
@protonu I believe this file now contains some symbolic tests, so I believe this is resolved. If I'm wrong then please re-open.

PR #2050, opened 2024-04-05
ATen evaluation for linear primitive

Priya2698:
1. `fd.ops.linear` will take input shapes `[M, K] x [N, K]`. This is evaluated through `at::linear` for parity with PyTorch.
2. Fix the segmenter issue with Linear + Bias. See #2057.

Issue #1669

### DIFF
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -424,21 +424,27 @@ std::vector<PolymorphicValue> UnaryOp::evaluate(
           ? toScalar(ee.evaluate(matmul_inp.alpha, known_values))
           : 1;
 
+      // Matmul/Addmm: n_pos=2, k_pos=1
+      // Linear: n_pos=1, k_pos=2
+      const int k_pos =
+          matmul_inp.mma_dims_pos.value()[(size_t)MatmulDomain::K];
+      const int n_pos =
+          matmul_inp.mma_dims_pos.value()[(size_t)MatmulDomain::N];
+
       if (matmul_inp.bias == nullptr) {
-        auto out = matmul_inp.input_layout == MmaLayout::TT
-            ? alpha * a.matmul(b)
-            : at::linear(a, b);
+        auto out = k_pos < n_pos ? alpha * a.matmul(b) : at::linear(a, b);
         return {out};
       }
 
       auto bias = ee.evaluate(matmul_inp.bias, known_values).as<at::Tensor>();
 
       // Linear takes 1D bias. Unsqueeze for 1D bias in matmul/addmm.
-      if (bias.dim() != a.dim() && matmul_inp.input_layout == MmaLayout::TT) {
+      if (bias.dim() != a.dim() && (k_pos < n_pos)) {
         // Unsqueeze the broadcast dimensions.
         // For 2D inputs to the pattern, bias is of shape [M,1]/[1,N]
-        for (int64_t dim : c10::irange(matmul_inp.bias_bcast_flags.size())) {
-          if (matmul_inp.bias_bcast_flags[dim]){
+        for (auto dim :
+             c10::irange((int64_t)matmul_inp.bias_bcast_flags.size())) {
+          if (matmul_inp.bias_bcast_flags[dim]) {
             bias = bias.unsqueeze(dim);
           }
         }
@@ -448,9 +454,8 @@ std::vector<PolymorphicValue> UnaryOp::evaluate(
           ? toScalar(ee.evaluate(matmul_inp.beta, known_values))
           : 1;
 
-      auto out = matmul_inp.input_layout == MmaLayout::TT
-          ? at::addmm(bias, a, b, beta, alpha)
-          : at::linear(a, b, bias);
+      auto out = k_pos < n_pos ? at::addmm(bias, a, b, beta, alpha)
+                               : at::linear(a, b, bias);
       return {out};
     }
   }
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -413,17 +413,21 @@ std::vector<PolymorphicValue> UnaryOp::evaluate(
     MmaOpUtils::MatmulInputs matmul_inp;
 
     if (MmaOpUtils::matchMatmulPatterns(this, &matmul_inp)) {
-      // Inputs to MmaOp are of the shape [M, K] x [K, N] (matmul) / [M, K] x [N, K] (linear)
-      const auto a = ee.evaluate(matmul_inp.mma_lhs, known_values)
-                         .as<at::Tensor>();
-      const auto b = ee.evaluate(matmul_inp.mma_rhs, known_values)
-                         .as<at::Tensor>();
+      // Inputs to MmaOp are of the shape [M, K] x [K, N] (matmul) / [M, K] x
+      // [N, K] (linear).
+      // Note: alpha, beta parameters are not used when linear pattern is
+      // detected (TN layout).
+      const auto a =
+          ee.evaluate(matmul_inp.mma_lhs, known_values).as<at::Tensor>();
+      const auto b =
+          ee.evaluate(matmul_inp.mma_rhs, known_values).as<at::Tensor>();
       const c10::Scalar alpha = matmul_inp.alpha
           ? toScalar(ee.evaluate(matmul_inp.alpha, known_values))
           : 1;
 
       if (matmul_inp.bias == nullptr) {
-        auto out = matmul_inp.mma_layout == MmaLayout::TT ? alpha * a.matmul(b) : at::linear(a, b);
+        auto out = matmul_inp.mma_layout == MmaLayout::TT ? alpha * a.matmul(b)
+                                                          : at::linear(a, b);
         return {out};
       }
       auto bias = ee.evaluate(matmul_inp.bias, known_values).as<at::Tensor>();
@@ -436,7 +440,9 @@ std::vector<PolymorphicValue> UnaryOp::evaluate(
           ? toScalar(ee.evaluate(matmul_inp.beta, known_values))
           : 1;
 
-      auto out = matmul_inp.mma_layout == MmaLayout::TT ? at::addmm(bias, a, b, beta, alpha) : at::linear(a, b, bias);
+      auto out = matmul_inp.mma_layout == MmaLayout::TT
+          ? at::addmm(bias, a, b, beta, alpha)
+          : at::linear(a, b, bias);
       return {out};
     }
   }
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -427,9 +427,9 @@ std::vector<PolymorphicValue> UnaryOp::evaluate(
       // Matmul/Addmm: n_pos=2, k_pos=1
       // Linear: n_pos=1, k_pos=2
       const int k_pos =
-          matmul_inp.mma_dims_pos.value()[(size_t)MatmulDomain::K];
+          std::get<(size_t)MatmulDomain::K>(matmul_inp.mma_dims_pos);
       const int n_pos =
-          matmul_inp.mma_dims_pos.value()[(size_t)MatmulDomain::N];
+          std::get<(size_t)MatmulDomain::N>(matmul_inp.mma_dims_pos);
 
       if (matmul_inp.bias == nullptr) {
         auto out = k_pos < n_pos ? alpha * a.matmul(b) : at::linear(a, b);
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -413,9 +413,9 @@ std::vector<PolymorphicValue> UnaryOp::evaluate(
     MmaOpUtils::MatmulInputs matmul_inp;
 
     if (MmaOpUtils::matchMatmulPatterns(this, &matmul_inp)) {
-      // Inputs to MmaOp are of the shape [M, K] x [K, N] (matmul) / [M, K] x
-      // [N, K] (linear).
-      // Note: alpha, beta parameters are nullptr for linear.
+      // Inputs to the pattern are of the shape [M, K] x [K, N] (matmul) / [M,
+      // K] x [N, K] (linear). Note: alpha, beta parameters are nullptr for
+      // linear.
       const auto a =
           ee.evaluate(matmul_inp.mma_lhs, known_values).as<at::Tensor>();
       const auto b =
@@ -435,8 +435,12 @@ std::vector<PolymorphicValue> UnaryOp::evaluate(
 
       // Linear takes 1D bias. Unsqueeze for 1D bias in matmul/addmm.
       if (bias.dim() != a.dim() && matmul_inp.input_layout == MmaLayout::TT) {
-        bias = bias.unsqueeze(
-            *matmul_inp.bias_bcast_axis); // Bias is of shape [M,1]/[1,N]
+        // Unsqueeze the broadcast dimensions.
+        // For 2D inputs to the pattern, bias is of shape [M,1]/[1,N]
+        for (auto dim : c10::irange(matmul_inp.bias_bcast_flags.size())) {
+          if (matmul_inp.bias_bcast_flags[dim])
+            bias = bias.unsqueeze(dim);
+        }
       }
 
       const c10::Scalar beta = matmul_inp.beta
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -437,9 +437,10 @@ std::vector<PolymorphicValue> UnaryOp::evaluate(
       if (bias.dim() != a.dim() && matmul_inp.input_layout == MmaLayout::TT) {
         // Unsqueeze the broadcast dimensions.
         // For 2D inputs to the pattern, bias is of shape [M,1]/[1,N]
-        for (auto dim : c10::irange(matmul_inp.bias_bcast_flags.size())) {
-          if (matmul_inp.bias_bcast_flags[dim])
+        for (int64_t dim : c10::irange(matmul_inp.bias_bcast_flags.size())) {
+          if (matmul_inp.bias_bcast_flags[dim]){
             bias = bias.unsqueeze(dim);
+          }
         }
       }
 
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -415,8 +415,7 @@ std::vector<PolymorphicValue> UnaryOp::evaluate(
     if (MmaOpUtils::matchMatmulPatterns(this, &matmul_inp)) {
       // Inputs to MmaOp are of the shape [M, K] x [K, N] (matmul) / [M, K] x
       // [N, K] (linear).
-      // Note: alpha, beta parameters are not used when linear pattern is
-      // detected (TN layout).
+      // Note: alpha, beta parameters are nullptr for linear.
       const auto a =
           ee.evaluate(matmul_inp.mma_lhs, known_values).as<at::Tensor>();
       const auto b =
@@ -426,21 +425,24 @@ std::vector<PolymorphicValue> UnaryOp::evaluate(
           : 1;
 
       if (matmul_inp.bias == nullptr) {
-        auto out = matmul_inp.mma_layout == MmaLayout::TT ? alpha * a.matmul(b)
-                                                          : at::linear(a, b);
+        auto out = matmul_inp.input_layout == MmaLayout::TT
+            ? alpha * a.matmul(b)
+            : at::linear(a, b);
         return {out};
       }
+      
       auto bias = ee.evaluate(matmul_inp.bias, known_values).as<at::Tensor>();
-      if (bias.dim() != a.dim()) {
+      if (bias.dim() != a.dim() && matmul_inp.input_layout == MmaLayout::TT) {
         // Bias is of shape [M,].
         NVF_ERROR(bias.dim() == a.dim() - 1);
         bias = bias.unsqueeze(-1);
       }
+
       const c10::Scalar beta = matmul_inp.beta
           ? toScalar(ee.evaluate(matmul_inp.beta, known_values))
           : 1;
 
-      auto out = matmul_inp.mma_layout == MmaLayout::TT
+      auto out = matmul_inp.input_layout == MmaLayout::TT
           ? at::addmm(bias, a, b, beta, alpha)
           : at::linear(a, b, bias);
       return {out};
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -430,12 +430,12 @@ std::vector<PolymorphicValue> UnaryOp::evaluate(
             : at::linear(a, b);
         return {out};
       }
-      
+
       auto bias = ee.evaluate(matmul_inp.bias, known_values).as<at::Tensor>();
+
+      // Linear takes 1D bias. Unsqueeze for 1D bias in matmul/addmm. 
       if (bias.dim() != a.dim() && matmul_inp.input_layout == MmaLayout::TT) {
-        // Bias is of shape [M,].
-        NVF_ERROR(bias.dim() == a.dim() - 1);
-        bias = bias.unsqueeze(-1);
+        bias = bias.unsqueeze(*matmul_inp.bias_bcast_axis); // Bias is of shape [M,1]/[1,N]
       }
 
       const c10::Scalar beta = matmul_inp.beta
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -433,9 +433,10 @@ std::vector<PolymorphicValue> UnaryOp::evaluate(
 
       auto bias = ee.evaluate(matmul_inp.bias, known_values).as<at::Tensor>();
 
-      // Linear takes 1D bias. Unsqueeze for 1D bias in matmul/addmm. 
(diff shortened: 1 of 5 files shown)


csrc/ir/utils.cpp
```diff
@@ -1335,18 +1335,15 @@ void verifyMmaOpForEvaluation(
       "MmaOp::evaluate is not implemented for size: ",
       tv_a->nDims());
 
+  NVF_ERROR(
+      *mma_op->layout() == MmaLayout::TT || *mma_op->layout() == MmaLayout::TN);
```
wujingyue, line 1339: Interesting -- why did you choose to give up NT and NN?

csrc/ir/nodes.cpp
```diff
@@ -413,19 +413,22 @@ std::vector<PolymorphicValue> UnaryOp::evaluate(
...
-      const auto a = ee.evaluate(matmul_inp.mma_lhs, known_values)
-                         .as<at::Tensor>()
-                         .squeeze(-1);
-      const auto b = ee.evaluate(matmul_inp.mma_rhs, known_values)
-                         .as<at::Tensor>()
-                         .squeeze(0);
+      // Inputs to MmaOp are of the shape [M, K] x [K, N] (matmul) / [M, K] x
+      // [N, K] (linear).
+      // Note: alpha, beta parameters are not used when linear pattern is
+      // detected (TN layout).
```
wujingyue, line 419: I don't follow this comment. Why is it not a problem to ignore alpha and beta?

jacobhinkle: I think this is because linear does not have alpha and beta params, so they should be nullptr in the match object.

csrc/ir/nodes.cpp
```diff
@@ -437,7 +440,10 @@ std::vector<PolymorphicValue> UnaryOp::evaluate(
           ? toScalar(ee.evaluate(matmul_inp.beta, known_values))
           : 1;
 
-      return {at::addmm(bias, a, b, beta, alpha)};
+      auto out = matmul_inp.mma_layout == MmaLayout::TT
+          ? at::addmm(bias, a, b, beta, alpha)
+          : at::linear(a, b, bias);
```
jacobhinkle, line 445: Why does the memory layout dictate which function is called?

csrc/ir/utils.cpp
```diff
@@ -1414,8 +1411,10 @@ bool matchMatmulPatterns(const UnaryOp* cast_op, MatmulInputs* matmul_inp) {
 
   // Verify assumptions for MmaOp hold. Assign the values to Mma operands.
   MmaOpUtils::verifyMmaOpForEvaluation(mma, final_out_dtype);
-  matmul_inp->mma_lhs = mma->inA();
-  matmul_inp->mma_rhs = mma->inB();
+  // Get the non-broadcasted values to avoid inferring squeeze dimensions.
+  matmul_inp->mma_lhs = mma->inA()->definition()->input(0);
+  matmul_inp->mma_rhs = mma->inB()->definition()->input(0);
+  matmul_inp->mma_layout = mma->layout().value();
```
jacobhinkle, line 1417: This might better be done with `mma_utils::getMmaLayout(fusion);` than looking at what's stored in the MmaOp.

csrc/ir/utils.h
```diff
@@ -66,13 +66,17 @@ struct MatmulInputs {
...
 };
 
 //! Matches the following matmul patterns.
 //! Matmul: A x B, alpha * A x B
 //! Matmul + Bias: A x B + C,  alpha * A x B + C, A x B + beta * C,
 //!   alpha * A x B  + beta * C
-//! Note: For simplicity, we assume the MmaOp to be in the first operand.
+//! Assumptions:
+//! 1. For simplicity, we assume the MmaOp to be in the first operand.
+//! 2. If mma layout is TN ([M, K], [N, K]), alpha, beta parameters are ignored
```
jacobhinkle, line 78: Can you expand in a comment on the reasoning here? I thought we were going to match for a transpose on the B operand instead of looking at layout.

tests/cpp/test_matmul_aten_evaluation.cpp
```diff
@@ -334,4 +334,83 @@ TEST_F(MatmulATenEvaluationTest, MatmulBiasAlpha) {
...
+
+TEST_F(MatmulATenEvaluationTest, Linear) {
+  auto fusion = std::make_unique<Fusion>();
+  FusionGuard fg(fusion.get());
+
+  int64_t m = 32, n = 64, k = 128;
+  std::vector<int64_t> a_shape{m, k}, b_shape{n, k}, out_shape{m, n};
+
+  auto tv0 = makeConcreteTensor(a_shape, DataType::Half);
+  auto tv1 = makeConcreteTensor(b_shape, DataType::Half);
```
jacobhinkle, line 346: Is there any reason to make these tensors concrete?

tests/cpp/test_matmul_aten_evaluation.cpp
```diff
@@ -334,4 +334,83 @@ TEST_F(MatmulATenEvaluationTest, MatmulBiasAlpha) {
...
+  auto tv0b = broadcast(tv0, {false, true, false}); // [M, 1, K]
+  auto tv1b = broadcast(tv1, {true, false, false}); // [1, N, K]
```
jacobhinkle, line 348: If this is how we plan to implement linear op, then could you pattern match the location of the K axis in order to determine whether its matmul or linear? Since matmul will have K in the middle position instead.

tests/cpp/test_matmul_aten_evaluation.cpp
```diff
@@ -334,4 +334,83 @@ TEST_F(MatmulATenEvaluationTest, MatmulBiasAlpha) {
...
+  auto tv6 = castOp(DataType::Half, tv5);
+
+  fusion->addInput(tv0);
+  fusion->addInput(tv1);
+  fusion->addInput(tv3);
+  fusion->addOutput(tv6);
+
+  at::Tensor t0 = at::randn(a_shape, at::kHalf).cuda();
+  at::Tensor t1 = at::randn(b_shape, at::kHalf).cuda();
+  at::Tensor t2 = at::randn({m}, at::kHalf).cuda();
```
protonu, line 399: These probably need to be changed as well.

csrc/ir/utils.h
protonu: I think based on our last meeting on Thursday, we had decided that we don't need a transpose.  We just need to setup the fusedMultiplySum with tensors of appropriate shapes and plumb it down to at::linear. Please correct me if I am wrong. cc @kevinstephano.

csrc/ir/utils.cpp
Priya2698: `fd.ops.matmul` will plumb down (M,K x K,N) and `fd.ops.linear` will plumb down `M,K x N,K` shapes.

However, now I think `MmaLayout` is not logical shapes but memory layout, which would render this condition incorrect.

csrc/ir/nodes.cpp
Priya2698: I'll rephrase the comment. alpha, beta will always be null for linear primitive.

csrc/ir/utils.cpp
Priya2698: So this function also returns `MmaLayout`. Are we using MmaLayout for two different definitions?

tests/cpp/test_matmul_aten_evaluation.cpp
```diff
@@ -334,4 +334,83 @@ TEST_F(MatmulATenEvaluationTest, MatmulBiasAlpha) {
...
+
+  int64_t m = 32, n = 64, k = 128;
+  std::vector<int64_t> a_shape{m, k}, b_shape{n, k}, out_shape{m, n};
+
+  auto tv0 = makeConcreteTensor(a_shape, DataType::Half);
+  auto tv1 = makeConcreteTensor(b_shape, DataType::Half);
+  auto tv0b = broadcast(tv0, {false, true, false}); // [M, 1, K]
+  auto tv1b = broadcast(tv1, {true, false, false}); // [1, N, K]
+  auto tv2 = fusedMultiplySum(tv0b, tv1b, {2});
+  auto tv3 = makeConcreteTensor({m}, DataType::Half);
```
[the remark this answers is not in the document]
Priya2698: I see. What about with matmul?

tests/cpp/test_matmul_aten_evaluation.cpp
Priya2698: No. We can go with symbolic or other types too.
We need to expand the tests (Issue #2032) which can be done in this or another PR.

csrc/ir/utils.cpp
wujingyue: > now I think MmaLayout is not logical shapes but memory layout

How so? I must have been missing something because https://github.com/NVIDIA/Fuser/blob/9eba1ad122550bf598f4a8d453d003cea989e4fe/tests/cpp/utils.cpp#L332-L347 seems to set up different logical shapes for different so-called `MmaLayout`.

Priya2698: Yeah, that's why I also assumed that these govern the logical shapes. But @jacobhinkle pointed out that these are memory layouts in another comment.

@jacobhinkle may be able to offer more clarity.
`mma->layout()` and `getMmaLayout(fusion)`, both return `MmaLayout but it seems they have different definitions at the two places?

tests/cpp/test_matmul_aten_evaluation.cpp
Priya2698: Using `getMmaLayout` will be the same, right?

csrc/ir/utils.cpp
Priya2698: This comment points to some differences: https://github.com/NVIDIA/Fuser/blob/9eba1ad122550bf598f4a8d453d003cea989e4fe/csrc/scheduler/mma_utils.h#L285

jacobhinkle: There are at least three notions related to layout that need clarifying here I think:
1. The`MmaOp` describes a tensor-core instruction, whose inputs can be transposed in registers or not. There are restrictions on the allowed layouts that those instructions take. During scheduling, this should be set accurately so that `codegen.cpp` generats the proper PTX. This is (I think) always `TN` currently, but it could change in future architectures.
2. The layout returned by `getMmaLayout(fusion)` is the problem layout: i.e. is one of the operands transposed in global memory. There are no restrictions in layouts here. At scheduling, we copy tiles to smem then we load the smem into registers using either `ldmatrix` or `ldmatrix.trans` such that the data is no longer transposed in registers, i.e. we achieve an `MmaOp::layout` equal to `TN`.
3. The test utils are meant to exercise different problem layouts (notion 2). They allocate contiguous arrays with different logical shapes, because we then transpose them to pass them into a mul-sum or matmul op. The matmul op still sees logical shapes of `(M, K) @ (K, N)`.
@zasdfgbnm and @drzejan2 please correct me if any of that is innacurate.

wujingyue: > They allocate contiguous arrays with different logical shapes, because we then transpose them to pass them into a mul-sum or matmul op. The matmul op still sees logical shapes of (M, K) @ (K, N).

Thanks! That's what I missed.

csrc/ir/utils.cpp
```diff
@@ -1441,20 +1451,22 @@ bool matchMatmulPatterns(const UnaryOp* cast_op, MatmulInputs* matmul_inp) {
...
-    NVF_ERROR(
-        matmul_inp->bias->as<TensorView>()
-            ->getRootDomain()
-            .back()
-            ->isBroadcast(),
-        "Expected last dimension to be broadcasted for bias tensor.");
+  if (bcast != nullptr) {
+    // Bias of shape [M, 1] / [1, N]
+    // TODO: Allow multiple bcast axes for batched GEMMs
+    matmul_inp->bias_bcast_axis = bcast->isBroadcastDim(0) ? 0 : 1;
```
Priya2698, line 1466: batched GEMMs can have multiple broadcast axes in the bias. For simplicity, I have set the bcast axes for the bias to be either 0 (`bias = [1, N]`) / 1 (`bias = [M, 1]`). A more general approach would be to get the broadcastDimFlags and unsqueeze all the dims which have the flag set to `true`.

Any preference here?

Note: This change was made to support any bias shape with `addmm`. Earlier, we only supported `[M,]` or `[M,N]`.

tests/cpp/test_matmul_aten_evaluation.cpp
Priya2698: @jacobhinkle I'll pin this for the next PR. Do you have any preference?

csrc/scheduler/mma_utils.cpp
```diff
@@ -1236,12 +1236,8 @@ RolesMapOpt getTensorsRoles(
         roles_map[MatmulRole::INPUT_B].push_back(entry.first);
         continue;
       }
-      if (has_m && has_n && !has_k) {
-        roles_map[MatmulRole::INPUT_C].push_back(entry.first);
-        continue;
-      }
       // Bias vectors are assigned to INPUT_C role
-      if (has_m && !has_n && !has_k) {
+      if (!has_k) {
```
Priya2698, line 1240: Fixed the segmenter issue with `linear + bias`.

Priya2698:
!build

tests/cpp/test_matmul_aten_evaluation.cpp
```diff
@@ -334,4 +340,84 @@ TEST_F(MatmulATenEvaluationTest, MatmulBiasAlpha) {
 
   EXPECT_TRUE(at::allclose(out[0], out_ref));
 }
+
+// fd.ops.linear(a, b) where a = [M,K], b = [N,K]
+TEST_F(MatmulATenEvaluationTest, Linear) {
```
protonu, line 345: I think it might be useful to have another test where we call at::linear with only 2 arguments (no bias).

tests/cpp/test_matmul_aten_evaluation.cpp
jacobhinkle: Separate PR sounds good. Keeping the PRs small is best for velocity.

csrc/ir/utils.h
```diff
@@ -66,13 +66,20 @@ struct MatmulInputs {
   Val* bias = nullptr;
   Val* alpha = nullptr;
   Val* beta = nullptr;
+  // Input layout based on ordering of axes M,N,K, valid values.
+  // Determined based on broadcast axes.
```
jjsjann123, line 70: nitpick: no longer inferred from broadcast axes.

Priya2698: It is determined based on which axes was broadcasted for the first Mma operand.

csrc/ir/utils.cpp
```diff
@@ -1414,15 +1418,21 @@ bool matchMatmulPatterns(const UnaryOp* cast_op, MatmulInputs* matmul_inp) {
...
   // Verify assumptions for MmaOp hold. Assign the values to Mma operands.
   MmaOpUtils::verifyMmaOpForEvaluation(mma, final_out_dtype);
-  matmul_inp->mma_lhs = mma->inA();
-  matmul_inp->mma_rhs = mma->inB();
+
+  // Get the non-broadcasted values to avoid inferring squeeze dimensions.
+  matmul_inp->mma_lhs = mma->inA()->definition()->input(0);
+  matmul_inp->mma_rhs = mma->inB()->definition()->input(0);
+  matmul_inp->input_layout =
+      mma->inA()->as<TensorView>()->getRootDomain().back()->isBroadcast()
```
jjsjann123, line 1426: Ha, I see what you mean by that now.

csrc/ir/utils.cpp
wujingyue: > A more general approach would be to get the broadcastDimFlags and unsqueeze all the dims which have the flag set to true.

How hard is this? If it's 5 lines of code as it sounds like, I'd totally do this in this PR to make code more robust. Code as is will fail quietly when batched gemms kick in.

csrc/ir/nodes.cpp
```diff
@@ -413,31 +413,40 @@ std::vector<PolymorphicValue> UnaryOp::evaluate(
...
+          ee.evaluate(matmul_inp.mma_rhs, known_values).as<at::Tensor>();
       const c10::Scalar alpha = matmul_inp.alpha
           ? toScalar(ee.evaluate(matmul_inp.alpha, known_values))
           : 1;
 
       if (matmul_inp.bias == nullptr) {
-        return {alpha * a.matmul(b)};
+        auto out = matmul_inp.input_layout == MmaLayout::TT
+            ? alpha * a.matmul(b)
+            : at::linear(a, b);
```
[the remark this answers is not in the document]
Priya2698: In `verifyMmaOpForEvaluation`, I check the broadcast axes are these two cases only. 
https://github.com/NVIDIA/Fuser/blob/2604e6ad63d216df78063778a59f805fc0c98ab6/csrc/ir/utils.cpp#L1346

csrc/ir/utils.cpp
Priya2698: It was to avoid this issue:https://github.com/NVIDIA/Fuser/issues/2057#issuecomment-2046133253
We also discussed this case offline.

Priya2698:
Tests pass here: https://gitlab-master.nvidia.com/dl/pytorch/fuser-gh-mirror/-/pipelines/14120900

csrc/ir/nodes.cpp
```diff
@@ -413,31 +413,40 @@ std::vector<PolymorphicValue> UnaryOp::evaluate(
...
+          ee.evaluate(matmul_inp.mma_lhs, known_values).as<at::Tensor>();
+      const auto b =
+          ee.evaluate(matmul_inp.mma_rhs, known_values).as<at::Tensor>();
       const c10::Scalar alpha = matmul_inp.alpha
           ? toScalar(ee.evaluate(matmul_inp.alpha, known_values))
           : 1;
 
       if (matmul_inp.bias == nullptr) {
-        return {alpha * a.matmul(b)};
+        auto out = matmul_inp.input_layout == MmaLayout::TT
```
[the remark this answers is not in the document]
Priya2698: How about `input_layout->input_shape`? If we are to avoid using `MmaLayout`, another enum called `MatrixShape`/`MmaShape`?

csrc/ir/utils.cpp
```diff
@@ -1414,15 +1418,21 @@ bool matchMatmulPatterns(const UnaryOp* cast_op, MatmulInputs* matmul_inp) {
 
   // Verify assumptions for MmaOp hold. Assign the values to Mma operands.
   MmaOpUtils::verifyMmaOpForEvaluation(mma, final_out_dtype);
-  matmul_inp->mma_lhs = mma->inA();
-  matmul_inp->mma_rhs = mma->inB();
+
+  // Get the non-broadcasted values to avoid inferring squeeze dimensions.
+  matmul_inp->mma_lhs = mma->inA()->definition()->input(0);
+  matmul_inp->mma_rhs = mma->inB()->definition()->input(0);
+  matmul_inp->input_layout =
```
protonu, line 1425: Might be useful to add a comment here

csrc/ir/nodes.cpp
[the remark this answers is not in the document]
wujingyue: `lhs/rhs->getMaybeRFactorDomain()` is technically the input shape and we don't need yet another input shape. I think what we are missing are the semantics of the dimensions -- e.g., are they contracting dimensions? Are they non-contracting and if so how are they mapped between lhs and rhs and the output? That's why I considered rhs_contracting_axis, which is on the way of adding more semantics to the dimensions.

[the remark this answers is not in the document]
jacobhinkle: I agree that the rfactor domains of the inputs is all that matters. You could factor this piece of code out into a helper function then re-use it here:
https://github.com/NVIDIA/Fuser/blob/1110b0aff240e46604283ac4f9fcfe27bc276ff8/csrc/scheduler/mma_utils.cpp#L965-L989

Then you could use those positions to squeeze the broadcasted dims (N for A and M for B) and transpose A or B if K < M or K < N.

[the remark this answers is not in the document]
Priya2698: Thanks @jacobhinkle. Modified the code to use positions of M,N,K to differentiate between matmul/addmm and linear.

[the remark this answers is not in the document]
Priya2698: On a separate note, there does seem to be redundancy in `mma_utils.cpp`.
For instance, we have two enums `MmaDimension` and `MatmulDomains`. Similarly, `getProblemIterDomains` and `getMmaDomainSet` are both identifying problem iterdomains, so there is possibility of unifying these.

I will open an issue for this.
CC: @zasdfgbnm

Priya2698:
!build

csrc/ir/utils.cpp
```diff
@@ -1310,6 +1311,40 @@ MmaOpDetails getMmaOpDetails(
   return details;
 }
 
+namespace {
+std::vector<int> getMmaDimsPositions(MmaOp* mma) {
```
wujingyue, line 1315:
```suggestion
std::tuple<int, int, int> getMmaDimsPositions(MmaOp* mma) {
```

Also, add a code comment that explains the return value.

csrc/ir/utils.cpp
```diff
@@ -1310,6 +1311,40 @@ MmaOpDetails getMmaOpDetails(
...
+  auto mma_domains = mma_utils::getProblemIterDomains(mma->fusion());
+  NVF_ERROR(mma_domains.isValid(), mma_domains.getErrorMsg());
+
+  const auto domains_data = mma_domains.getData();
+  const auto m_id = domains_data[(size_t)MatmulDomain::M];
+  const auto n_id = domains_data[(size_t)MatmulDomain::N];
+  const auto k_id = domains_data[(size_t)MatmulDomain::K];
+
+  std::optional<int> m_pos, n_pos, k_pos = std::nullopt;
```
wujingyue, line 1324: Nit: I'd simply use -1 to represent not-found. It's pretty clear that -1 isn't a valid index. 

```suggestion
  int m_pos = -1;
  int n_pos = -1;
  int k_pos = -1;
```

csrc/ir/utils.h
```diff
@@ -66,13 +66,23 @@ struct MatmulInputs {
   Val* bias = nullptr;
   Val* alpha = nullptr;
   Val* beta = nullptr;
+  // Ordering of dimensions M,N,K in inputs.
+  // Determined based on position of iterdomains.
+  // For addmm/matmul ([M,K] x [K,N]): M=0, N=2, K=1
+  // For linear ([M,K] x [N,K]): M=0, N=1, K=2
+  // mma_dims_pos = {m_pos, n_pos, k_pos}
+  std::optional<std::vector<int>> mma_dims_pos = std::nullopt;
```
wujingyue, line 74: Besides uninitialized, when is this nullopt?

csrc/ir/utils.h
```diff
@@ -66,13 +66,23 @@ struct MatmulInputs {
...
+  std::vector<bool> bias_bcast_flags = {};
```
wujingyue, line 75: Can you add a code comment for this field?

wujingyue (commented):
Almost LGTM! Thanks for the cleanup!

csrc/ir/utils.h
Priya2698: No other case. Only when the pattern matching fails or when the positions cannot be properly assigned to M,N,K iterdomains.

Priya2698: Removed `std::optional`

Priya2698:
CI: https://gitlab-master.nvidia.com/dl/pytorch/fuser-gh-mirror/-/pipelines/14202941

csrc/ir/utils.h
```diff
@@ -66,13 +66,26 @@ struct MatmulInputs {
   Val* bias = nullptr;
   Val* alpha = nullptr;
   Val* beta = nullptr;
+  // Ordering of dimensions M,N,K in inputs.
+  // Determined based on position of iterdomains.
```
wujingyue, line 70: Why is `n_pos` or `k_pos` 2? For example, in `M,K x N,K`, `K` is dim 1 in lhs and again dim 1 in rhs. Does `k_pos=2` mean anything? Or is it an arbitrary way to distinguish `[K,N]` and `[N,K]` so `k_pos=10` should work equally well? 

To complicate this even more, while I understand we don't support `K,M x N,K` at this moment, how do you expect the following notation to be extended for that? 

To be more constructive, it **may** be less ambiguous and more extensible if we split k_pos into lhs_k_pos and rhs_k_pos, because k is a contracting dimension that appears in both operands. Does that make any sense?

csrc/ir/utils.h
```diff
@@ -66,13 +66,26 @@ struct MatmulInputs {
   Val* bias = nullptr;
   Val* alpha = nullptr;
   Val* beta = nullptr;
+  // Ordering of dimensions M,N,K in inputs.
```
wujingyue, line 69:
```suggestion
  // Ordering of dimensions M,N,K in MmaOp's output TensorView's root domain. 
```

csrc/ir/utils.h
wujingyue: Resolved in an offline discussion.

wujingyue (approved):

protonu (approved):

PR #2055, opened 2024-04-08
Canonicalize `dims` argument in `variance_mean` when `keepdim=True`

rdspring1:
**Problem:** The `variance_mean` function uses the `dims` argument directly, which can be a negative number. 
**Solution:** This PR canonicalizes the reduction dimensions when `keepdim=True`, and moves `canonicalizeAxes` function to utils.h


rdspring1:
!build

jjsjann123 (approved):
Thanks for clean up!

PR #2061, opened 2024-04-10
Fix serialization errors when skipping compilation in `FusionExecutor`. 

rdspring1:
Serialization assumes that `FusionExecutor` always contains a compiled kernel. This PR updates serialization to skip compilation correctly.


rdspring1:
!build

rdspring1:
!build

csrc/executor.cpp
```diff
@@ -2306,6 +2309,17 @@ void FusionExecutor::deserialize(
   // See table definition for FusionExecutor in serde/fusion_cache.fbs
 
   NVF_ERROR(buffer != nullptr, "serde::FusionExecutor is nullptr.");
+
+  // TODO Should we set fusion_id, concrete_id, runtime_id, and group_id when we
+  // skip compilation?
+  if (buffer->is_compilation_skipped()) {
+    fusion_ = std::make_unique<Fusion>(*fusion);
+    NVF_ERROR(
+        !isCompiled() && isCompilationSkipped(),
```
Priya2698, line 2318: `!isCompiled()` should be sufficient, why both the conditions?

csrc/executor.cpp
```diff
@@ -2364,7 +2378,9 @@ void FusionExecutor::deserialize(
   compiled_kernel_ = executor_utils::getCompiledKernel(
       buffer->compiled_kernel(), compile_params);
 
-  NVF_ERROR(isCompiled(), "Failed to deserialize FusionExecutor");
+  NVF_ERROR(
+      isCompiled() && !isCompilationSkipped(),
```
Priya2698, line 2382: Same as above, do we need both?

Priya2698:
General question: How does nvFuser handle if only one out of the many segments is not compiled? Do we skip serialization/deserialization for the entire fusion or is it segment-wise?

rdspring1:
> General question: How does nvFuser handle if only one out of the many segments is not compiled? Do we skip serialization/deserialization for the entire fusion or is it segment-wise?

We're skipping on a `FusionExecutor` basis, so it is segment-wise. Serialization follows the same behavior.

csrc/executor.cpp
rdspring1: The assertion wasn't originally within the if-statement, and I didn't update it.

Priya2698:
There are some new issues on my branch. I prefer either merging this to main or first fixing the issues in my branch so the root cause is clear. Wdyt?

Priya2698:
!build

rdspring1:
This PR can wait until the issues in your branch are fixed.

Priya2698:
> This PR can wait until the issues in your branch are fixed.

@rdspring1 We can merge this into my branch now. Could you resolve the conflicts, and finalize this PR? Thanks for the fix!

rdspring1:
!build

csrc/executor.cpp
```diff
@@ -2178,8 +2178,13 @@ flatbuffers::Offset<serde::FusionExecutor> FusionExecutor::serialize(
     executor_entry_lookup_values_fb.push_back(serialize(builder, value));
   }
 
+  if (isCompilationSkipped()) {
```
Priya2698, line 2181: Please add a comment here explaining the if condition

Priya2698 (approved):
LGTM. Added minor comments.

PR #2065, opened 2024-04-11
Add c10d mocks and localize NVFUSER_DISTRIBUTED in communicator. 

wujingyue:
This is the first step towards limiting the scope of NVFUSER_DISTRIBUTED to just communicator. If you think it's the right thing to do, more PRs will come your way!

### DIFF
--- a/csrc/multidevice/communicator.h
+++ b/csrc/multidevice/communicator.h
@@ -17,39 +17,10 @@
 #include <torch/csrc/distributed/c10d/Backend.hpp>
 #include <torch/csrc/distributed/c10d/TCPStore.hpp>
 #include <torch/csrc/distributed/c10d/Work.hpp>
+#else
+#include <multidevice/c10d_mock.h>
 #endif
 #include <visibility.h>
-
-#if !defined(NVFUSER_DISTRIBUTED)
-namespace c10d {
-class Work : public torch::CustomClassHolder {
- public:
-  void wait() {}
-};
-
-class Backend : public torch::CustomClassHolder {
- public:
-  c10::intrusive_ptr<c10d::Work> barrier() {
-    return c10::make_intrusive<c10d::Work>();
-  }
-  c10::intrusive_ptr<c10d::Work> send(
-      std::vector<at::Tensor>& tensors,
-      int dstRank,
-      int tag) {
-    return c10::make_intrusive<c10d::Work>();
-  }
-  c10::intrusive_ptr<Work> recv(
-      std::vector<at::Tensor>& tensors,
-      int srcRank,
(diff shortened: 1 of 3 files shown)


wujingyue:
!build

wujingyue:
!build

wujingyue:
!build --dist

csrc/multidevice/communicator.h
```diff
@@ -6,15 +6,51 @@
...
+  c10::intrusive_ptr<c10d::Work> barrier() {
+    return c10::make_intrusive<c10d::Work>();
+  }
+  c10::intrusive_ptr<c10d::Work> send(
+      std::vector<at::Tensor>& tensors,
+      int dstRank,
+      int tag) {
+    return c10::make_intrusive<c10d::Work>();
+  }
+  c10::intrusive_ptr<Work> recv(
```
cowanmeg, line 41:
```suggestion
  c10::intrusive_ptr<c10d::Work> recv(
```

nit: just to stay consistent with the other return types

cowanmeg (approved):
LGTM! I assume this is the start of code changes that will make our build more robust against non-distributed pytorch and hopefully get rid of NVFUSER_DISTRIBUTED for USE_DISTRIBUTED?

wujingyue:
Thanks for reviewing this! 

> make our build more robust against non-distributed pytorch 

To some degree, yes. We already have non-distributed CI tests, so I don't worry about people breaking those post-submit :) But limiting the scope of NVFUSER_DISTRIBUTED makes it hard to make pre-submit mistakes that break only non-distributed. This makes people more confident touching code around multidevice. 

Another motivation is housekeeping so we don't have to ifdef NVFUSER_DISTRIBUTED in files we don't need to. ifdef is the least flexible version of "if" because the compiler sees one path at a time. https://github.com/NVIDIA/Fuser/blob/332fc6f7a46069deb05089c5a0703080c0d40067/csrc/python_frontend/fusion_definition.cpp#L146-L163 is an example of duplicating code just because ifdef is around. 

> hopefully get rid of NVFUSER_DISTRIBUTED for USE_DISTRIBUTED?

This is orthogonal. AFAICT, USE_DISTRIBUTED seems better but comes with the extra work of changing the non-distributed CI to use non-distributed pytorch, which I'll talk to @xwang233 about.

csrc/multidevice/communicator.h
wujingyue: Thanks! I instead made others `Work` because it's under the c10d namespace. I also moved the mocks to a separate header file.

wujingyue:
!build --dist

Issue #2063, opened 2024-04-11
reduction with pointwise epilogue (tests, code optimizaiton, remove limitation)


liqiangxl:
Current reduction scheduler limites types of epilogue pointwise ops can be fused through `SchedulerTopologyChecker`. 
It needs further works in the following areas:
(1) the tests are missing
(2) the generated code may not optimal, e.g. outer reduciton + non-broadcast pointwise is allowed but the loading of the additional inputs for the non-broadcast pointwise is not predicated. The current generated code is:
```
  float T5[1];
  T5[0] = 0;
  if (b7) {
    T5[0]
       = T1[i6];
  }
  float T6[1];
  T6[0]
    = T2[0]
    + T5[0];
  if ((((((nvfuser_index_t)blockIdx.y) == (((nvfuser_index_t)gridDim.y) + -1)) && (((nvfuser_index_t)threadIdx.y) == 0)) && b7)) {
    T3[i6]
       = T6[0];
  }
```
The additional input tensor can be loaded during the final if condition check.
```

  if ((((((nvfuser_index_t)blockIdx.y) == (((nvfuser_index_t)gridDim.y) + -1)) && (((nvfuser_index_t)threadIdx.y) == 0)) && b7)) {
    float T5[1];
    T5[0] = 0;
    if (b7) {
      T5[0]
         = T1[i6];
    }
    float T6[1];
    T6[0]
      = T2[0]
      + T5[0];  
    T3[i6]
         = T6[0];
  }
```
(3) some of the limitations may be lifted if the scheduler is revised (needs to confirm).

liqiangxl:
Avoid redundant load from gmem has no noticable influence on performance.
![image](https://github.com/NVIDIA/Fuser/assets/116412316/02e8f301-0388-49cd-afa9-c46712a0fbba)

liqiangxl:
Inner reduction with non-bcast epilogue can be fused into one kernel and faster than segmented version if the kernel launch latency is also acounted. 
In fused version: the most common case is, each block load 1 element of the non-bcast epilogue tensor.
In segmented version: reduction result is dumped to gmem and the 2nd kernel do pointwise op on the reduction result and non-bcast epilogue tensor.
![image](https://github.com/NVIDIA/Fuser/assets/116412316/9bd57dd5-0deb-4632-856c-45867524cfdf)
![image](https://github.com/NVIDIA/Fuser/assets/116412316/022151fe-c465-4f55-8066-eaff5a15302c)

[...]

PR #2067, opened 2024-04-12
add tests for reduction with pointwise epilogue

liqiangxl:
First step of #2063 
Added tests for:
(1) inner reduction + non-broadcast epilogue
(2) outer reduction + non-broadcast epilogue
(3) inner reduction + broadcast epilogue
(4) outer reduction + broadcast epilogue

### DIFF
--- a/tests/cpp/test_reduction_pointwise.cpp
+++ b/tests/cpp/test_reduction_pointwise.cpp
@@ -11,8 +11,8 @@ namespace nvfuser {
 
 using PointwiseFusedReductionTest = NVFuserTest;
 
-// NO:  inner reduction + non-broadcast epilogue
-// YES: outer reduction + non-broadcast epilogue
+// inner reduction + non-broadcast epilogue, can't be fused
+// outer reduction + non-broadcast epilogue, can be fused
 // checked by: SchedulerTopologyChecker::supportedPostReductionFusion
 namespace {
 void ReductionNonBroadcast(const int reduction_dim) {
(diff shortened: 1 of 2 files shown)


tests/cpp/test_reduction_pointwise.cpp
```diff
@@ -0,0 +1,122 @@
...
+      {t_ref},
+      __LINE__,
+      __FILE__);
+};
+} // namespace
+TEST_F(PointwiseFusedReductionTest, InnerReductionNonBroadcast) {
+  const int reduction_dim = 1;
+  ReductionNonBroadcast(reduction_dim);
+}
+TEST_F(PointwiseFusedReductionTest, OuterReductionNonBroadcast) {
```
jjsjann123, line 65: so even though we are adding all four tests, the `OuterReductionNonBroadcast` is really the only one that the entire fusion goes through a single kernel. The other three would just segment into `reduction` + `pointwise`?

jjsjann123:
One more question here is that, this is only for single arith operation behind a reduction, so we are not yet considering use this to verify the other plan we have that uses `# of fusion ops to cap unroll_factor`?

tests/cpp/test_reduction_pointwise.cpp
liqiangxl: You are right. I believe all other three cases are segmented due to performance reason and we need to revisit this decision by optimizing the generated code and add corresponding benchmark cases in the followed PRs.

liqiangxl:
> One more question here is that, this is only for single arith operation behind a reduction, so we are not yet considering use this to verify the other plan we have that uses `# of fusion ops to cap unroll_factor`?

Not yet. This work is checking how to schedule reduction with epilogue ops. The heuristic revision primarily focuses on prelogue ops. Need to check the influence of epilogue ops on optimal heuristics after the prelogue ops are investigated. Then, combine them, looks even more difficult to have one works for all!

jjsjann123 (approved):
Thanks for adding the test.

PR #2075, opened 2024-04-15
TMA doc: part 3

zasdfgbnm:
- `tma.md`
  - Define box by compositing
  - Step 3: define tile 
- `tma-modeling-in-depth.md`
  - The unachievability of strong correctness for indivisible element stride
  - The lowering strategy


doc/reading/tma-modeling-in-depth.md
```diff
@@ -168,6 +176,298 @@ Another example is the following Figure 4:
...
+
+$\square$
+
+</details>
+
+Now, let's go back to the example in Figure 5.
+In this example, all the conditions of Theorem 6 are satisfied for the outer dimension,
+so strong correctness is unachievable. What if some condition is violated?
+
+If the element stride divided the box size, then the striding split would not create any hole at all.
```
naoyam, line 424: What's the "striding split'?

zasdfgbnm: It is in `tma.md` `Step 3: define tile`

```
For strided tile, we do an inner-split on the box IterDomain by the *element stride*.
We call this split "*striding split*", the inner output of this split "*stride IterDomain*",
```

doc/dev/tma.md
```diff
@@ -86,12 +90,87 @@ We call this split expression a "*boxing split*", the input of this split a "*pa
...
 For the case of Figure 1, if both box dimensions are defined by partitioning,
-the schedule should look like the Figure 3 below:
+the schedule should look like the Figure 4 below:
 
-![Figure 3: Boxing by partitioning](tma/box-by-partitioning.svg)
+![Figure 4: Boxing by partitioning](tma/box-by-partitioning.svg)
 
 Please note that, although in the above example, the split is divisible, this does not have to be the case in general.
 
 #### Define box by compositing
```
zasdfgbnm, line 99: This is supported in https://github.com/NVIDIA/Fuser/pull/1991.

doc/dev/tma.md
```diff
@@ -86,12 +90,87 @@ We call this split expression a "*boxing split*", the input of this split a "*pa
...
+The TMA domain of the tensor has nine IterDomains.
+The contiguities of these nine IterDomains are `(T, T, T, F, T, T, T, T, T)`.
+We scheduled a 5D TMA operation, where in these five dimensions, one of them defines box by partitioning, and others define box by compositing.
+Green color means box IterDomain and brown color means coordinate IterDomain.
+The first dimension is `[I1, I2]`, where `I1` is coordinate IterDomain and `I2` is box IterDomain.
+The second dimension is `I3`, it is the only dimension that defines box by partitioning.
+`I3` is the partitioned IterDomain, `I11` is the box IterDomain, and `I10` is the coordinate IterDomain.
+The third dimension is `I4`, it has no box IterDomain, therefore, the box size is implicitly one.
+The fourth dimension is `[I5, I6]`, it has no coordinate IterDomain, therefore this dimension only has one box,
+whose size is the produce of the extents of `I5` and `I6`.
```
naoyam, line 143:
```suggestion
whose size is the product of the extents of `I5` and `I6`.
```

doc/dev/tma.md
```diff
@@ -86,12 +90,87 @@ We call this split expression a "*boxing split*", the input of this split a "*pa
...
+The fifth dimension is `[I7, I8, I9]`, where `I8` and `I9` are box IterDomain, and `I7` is coordinate IterDomain.
+The imaginary TMA domain has five IterDomains, they are `merge(I1, I2)`, `I3`, `I4`, `merge(I5, I6)`, and `merge(I7, I8, I9)`.
+
+### Step 3: define tile
+
+TMA's tile can be dense or strided.
+Most commonly, we use dense tile.
+For dense tile, we define the *tile IterDomain* as the box IterDomain,
+and use the word *tile* and *box* interchangably.
```
naoyam, line 152:
```suggestion
and use the word *tile* and *box* interchangeably.
```

doc/dev/tma.md
```diff
@@ -86,12 +90,87 @@ We call this split expression a "*boxing split*", the input of this split a "*pa
...
 
-![Figure 3: Boxing by partitioning](tma/box-by-partitioning.svg)
+![Figure 4: Boxing by partitioning](tma/box-by-partitioning.svg)
 
 Please note that, although in the above example, the split is divisible, this does not have to be the case in general.
 
 #### Define box by compositing
 
-TODO: this documentation is under construction
+Define box by compositing is to select a slice of IterDomains from the TMA domain and consider the innermost IterDomain(s) in the slice as box.
```
naoyam, line 101: If the `I5` TMA domain is scheduled, then can we replace this way of defining boxes by the above partitioning method? I wonder why we need this method in addition to the partitioning.

zasdfgbnm: Yes, we can always replace this way of defining boxes by partitioning. This way exists primarily for convenience. For example, if I have a discontiguous tensor [I1, I2], then you will be forced to partition both I1 and I2. But what often happen is, you only want to partition I2, and use box size 1 for I1. With this way of defining box, we will be able to automatically consider this as a size 1 box. Another example is, when you want to write a unit test, you often want to treat the entire dim as box, with only partitioning, you are forced to split this dim by its extent to get a [1, N]. But with this mechanism, you will be able to just parallelize this dim as `ParallelType::Bulk`.

doc/reading/tma-modeling-in-depth.md
```diff
@@ -168,6 +176,298 @@ Another example is the following Figure 4:
...
+If the the tensor shape was `(4, 2)` instead of `(8, 2)`,
+which violated the condition that the box size is smaller than the tensor size,
+then the red items in Figure 6 would not exist, because items from `8` to `15` would now be out of boundary.
+
+If the element stride was `5` instead of `3`,
+the red items in Figure 6 would still exist without extra protextion,
+because the example of $i_3 = 0$, $i_4 = 4$, and $i_1 = 4$ mentioned above is still valid.
+However, by lowering smartly, we can add a check for the range of `I6` (i.e. $i_6 < 4$) to get rid of the red items.
+
+## The lowering strategy
```
naoyam, line 435: This part is still difficult for me to follow. Some examples would be helpful.

doc/reading/tma-modeling-in-depth.md
```diff
@@ -168,6 +176,298 @@ Another example is the following Figure 4:
...
+
+Theorem 6 is so powerful that it deserve a fancier name "**The fundamental theorem of TMA correctness**",
+or **FTTC** in short.
+It is powerful because the condition in it is both necessary and sufficient,
+and it directly tells us the correct lowering strategy:
+
+We should check the condition of FTTC.
+If the condition is true, then it is impossible to achieve strong correctness.
+So raise an error if strong correctness is required.
+If strong correctness is not required, generate code like below:
```
naoyam, line 445: If the strong correctness is not required, it means the weak correctness is sufficient, right? If so, why does it need additional predicates? Doesn't TMA automatically guarantee the weak correctness?

doc/reading/tma-modeling-in-depth.md
```diff
@@ -168,6 +176,298 @@ Another example is the following Figure 4:
...
+
+```C++
+if (predicates_for_all_non_TMA_protected_IterDomains_except_box_IterDomains) {
+  // Do TMA normally
+  tma(...);
+}
+```
+
+If the condition of FTTC is false, then we always achieve strong correctness by generate the following code:
```
naoyam, line 454: Also here, FTTC is false, which means the strong correctness is guaranteed. Why are these predicates necessary?

doc/reading/tma-modeling-in-depth.md
```diff
@@ -168,6 +176,298 @@ Another example is the following Figure 4:
...
+
+![Figure 6: Indivisible strided tile data](tma-modeling-in-depth/no-strong-correctness-data.svg)
+
+From the above figure, we see that there are two tiles where the first half tile is blue and the second half tile is red.
+However, the granularity on TMA that programmer has control of is per tile.
+There is no way for a programmer to tell the hardware that they want part of a tile to be filled with data but part of it filled with zero even if they are not out of the boundary of global memory.
+
+The above observation leads to the following theorem:
+
+**Theorem 6:** It is impossible to achieve strong correctness if and only if there exist a dimension that satisfies all of the following conditions:
```
naoyam, line 203: nit: perhaps easier to remember to rephrase this as:
 "The strong correctness is achieved only if the following conditions are satisfied."

doc/reading/tma-modeling-in-depth.md
zasdfgbnm: It does automatically guarantee the weak correctness, but if there is a way to save some memory traffic, why not do it?

naoyam: I thought the only concern here was correctness.

doc/reading/tma-modeling-in-depth.md
zasdfgbnm: This is not what Theorem 6 is saying. What Theorem 6 says is, if the condition is satisfied, then no matter what you do, you can not achieve strong correctness. If the condition is not satisfied, then you can achieve strong correctness by doing something.

So there are three levels:
- If there is no indivisible splits on TMA-unprotected IDs, then strong correctness is automatically achieved without needing to do anything
- If there are indivisible splits on TMA-unprotected IDs, but the condition of FTTC is not satisfied, then strong correctness is achievable by predicating these TMA-unprotected IDs.
- If the condition of FTTC is satisfied, then you have no choice but to raise an error.

naoyam: What I wanted to suggest was also to change the conditions because it sounded like a double negative. For example, the following levels would be phrased like:

> but the condition of FTTC is satisfied, then strong correctness is achievable by predicating these TMA-unprotected IDs.

> If the condition of FTTC is not satisfied, then you have no choice but to raise an error.

Just a nit suggestion

doc/reading/tma-modeling-in-depth.md
zasdfgbnm: Modified the doc to be more clear on that

doc/reading/tma-modeling-in-depth.md
zasdfgbnm: Added a table clarifying

doc/reading/tma-modeling-in-depth.md
zasdfgbnm: I am not sure, but if rephrased like "Strong correctness is achievable if no dim satisfies both of the following conditions: 1, 2", then when we say "the condition of theorem 6", it is hard to tell if it refers to "no dim satisfies 1 and 2" or refers to "1, 2".

naoyam (approved):
Thanks for the clarifications. Much more clear to me.

PR #2077, opened 2024-04-15
Introduce Features as a replacement for `{Enable,Disable}Option`

jacobhinkle:
This is the first PR in a series that will do the following:

- Unify `EnableOption` and `DisableOption` into a single `Feature` enum.
- Remove existing guard classes.
- Add attributes to `FusionExecutor` and `FusionKernelRuntime` to hold `FeatureSet`s.
- Include `FeatureSet`s in cache keys and serialization.
- Plumb features into python frontend (proper update or alternative to #1905)

The current PR only introduces the `Feature` enum and the `FeatureSet` class which wraps a `std::bitset`. `EnableOption` and `DisableOption` are preserved, so there should be no change in behavior.

The next PR will plumb this into the caching system.


jacobhinkle:
I still think something like this would be better than using global state and guards, but I don't have concrete plans to revive it soon, so I'm closing this for now.

PR #2079, opened 2024-04-15
add benchmarks for redu pointwise epilogue

liqiangxl:
related to #2063 
added python benchmarks for redu pointwise epilogue testing:
`outer reduction + non-broadcast epilogue.` 
this is the only allowed epilogue in the current scheduler, can extend to include other cases if scheduler is revised.

Will add redu pointwise prolog with varied ops in another PR. 
These benchmarks help tuning the reduction heuristics.


benchmarks/cpp/reduction_epilogue.cpp
```diff
@@ -0,0 +1,123 @@
...
+  benchmark_state.SetBytesProcessed(
+      int64_t(benchmark_state.iterations()) *
+      (iter_size * reduction_size + epilogue_size + iter_size) *
+      int64_t(dataTypeSize(dtype)));
+}
+
+// define 8 benchmarks for different combinations of data type, reduction dim
+// and epilogue.
+#define NVFUSER_REDUCTION_POINTWISE(                    \
+    NAME_SUFFIX, DATA_TYPE, REDU_DIM, BCAST_ELOG)       \
```
jjsjann123, line 88: Apology in advance for nitpicking on macro :laughing: 

`NAME_SUFFIX` could be inferred from DATA_TYPE, REDU_DIM, BCAST_ELOG and I think that could avoid mistakes like typo on test names.

benchmarks/cpp/reduction_epilogue.cpp
```diff
@@ -0,0 +1,123 @@
...
+  at::Tensor aten_input_epilogue = is_broadcast
+      ? at::randn(input_shape, options)
+      : at::randn({iter_size}, options);
+
+  std::vector<c10::IValue> aten_inputs = {aten_input_x, aten_input_epilogue};
+
+  runBenchmarkIterations(benchmark_state, fusion_executor_cache, aten_inputs);
+
+  // inputs: input tensor [I*R] + epilogue tensor [I] or [I*R]
+  // outputs: output_of_reduction [I]
```
jjsjann123, line 77: output size doesn't look right here. If epilogue tensor is [I*R], we are writing out a full size [I*R].

benchmarks/cpp/reduction_epilogue.cpp
```diff
@@ -0,0 +1,123 @@
...
+  auto epilogue_size = is_broadcast ? iter_size * reduction_size : iter_size;
+  benchmark_state.SetBytesProcessed(
+      int64_t(benchmark_state.iterations()) *
+      (iter_size * reduction_size + epilogue_size + iter_size) *
```
jjsjann123, line 81:
```suggestion
      (iter_size * reduction_size + epilogue_size * 2) *
```
^^^ So here it should be like this?

liqiangxl:
there is a validation error:
`benchmarks/python/test_reduction_epilogue.py::test_reduction_epilogue_nvf_benchmark[0-torch.float32-(8192, 63744)] - AssertionError: Max error: 0.0001220703125,                     Absolute tolerance: 0.000136882,`, do you know how to bump the tolerence? @Priya2698

benchmarks/python/test_reduction_epilogue.py
```diff
@@ -0,0 +1,89 @@
...
+    benchmark,
+    size: tuple,
+    dtype: torch.dtype,
+    reduction_axis: int,
+    compile: bool,
+):
+    clear_cuda_cache()
+
+    x = torch.randn(*size, device="cuda", dtype=dtype)
+    epilogue = torch.randn([reduction_axis - 1], device="cuda", dtype=dtype)
```
Priya2698, line 81: `torch.randn(size[reduction_axis - 1], device="cuda", dtype=dtype)`

Priya2698: You can verify the torch baselines by adding the flags `--benchmark-eager --benchmark-torchcompile`. They are skipped by default

benchmarks/python/test_reduction_epilogue.py
```diff
@@ -0,0 +1,89 @@
...
+def test_reduction_epilogue_baseline_benchmark(
+    benchmark,
+    size: tuple,
+    dtype: torch.dtype,
+    reduction_axis: int,
+    compile: bool,
+):
+    clear_cuda_cache()
+
+    x = torch.randn(*size, device="cuda", dtype=dtype)
```
Priya2698, line 80: nit: `*size` -> `size`

benchmarks/python/test_reduction_epilogue.py
```diff
@@ -0,0 +1,89 @@
...
+def test_reduction_epilogue_nvf_benchmark(
+    benchmark,
+    size: tuple,
+    dtype: torch.dtype,
+    reduction_axis: int,
+    disable_validation: bool,
+    disable_benchmarking: bool,
+):
+    clear_cuda_cache()
+    x = torch.randn(*size, device="cuda", dtype=dtype)
```
Priya2698, line 50: nit: `*size` -> `size`

benchmarks/python/test_reduction_epilogue.py
```diff
@@ -0,0 +1,89 @@
...
+
+def reduction_epilogue_fwd_fn(
+    inputs: list,
+):  # in_tensor, epilogue_tensor, reduction_axis
+    return torch.sum(inputs[0], dim=inputs[2]) + inputs[1]
+
+
+@pytest.mark.parametrize("size", generate_input_sizes(dims=2))
+@pytest.mark.parametrize("dtype", FLOAT_DTYPES)
+@pytest.mark.parametrize("reduction_axis", [0])
```
Priya2698, line 40: If the `reduction_axis` is fixed, do we need this as a parameter? Unless you want to test it in the future.

Priya2698 (changes requested):

Priya2698:
> there is a validation error: `benchmarks/python/test_reduction_epilogue.py::test_reduction_epilogue_nvf_benchmark[0-torch.float32-(8192, 63744)] - AssertionError: Max error: 0.0001220703125, Absolute tolerance: 0.000136882,`, do you know how to bump the tolerence? @Priya2698

This is a known issue in reduction and normalization benchmarks. You can merge the PR if these seem to be just a case of low tolerance. We are discussing ways on avoiding these errors. Bumping up our validation constants did not necessarily work for all input sizes in the past.

benchmarks/python/test_reduction_epilogue.py
liqiangxl: yes, that's the intention, will add inner reduction when scheduler/heuristic is changed. we can do it now but not very useful, so prefer later.

Priya2698 (approved):

PR #2081, opened 2024-04-16
Reset semaphore in grid sync

jacobhinkle:
This is a follow-up to #1984. Previously we only reused semaphores for serial grid reductions. This change extends reusable zeroed memory to all grid syncs and grid reductions including welfords. Grid reductions and grid Welfords only interact with the semaphore through `grid_sync::sync`, so we only need to take care to reset the semaphore in that function.

After this change, I believe all global semaphores in nvFuser will be reused without needing a memset.

### DIFF
--- a/runtime/grid_sync.cu
+++ b/runtime/grid_sync.cu
@@ -75,6 +75,10 @@ __device__ void sync(
       }
 #endif
     }
+    if (last_block) {
+      // Clean up semaphore for re-use in next iteration
+      semaphore = 0ULL;
+    }
   }
 
   // Sync block to make sure all other threads are waiting on the sync
@@ -141,6 +145,8 @@ __device__ void sync(
         }
 #endif
       }
+      // Clean up semaphore for reuse
+      semaphore = 0ULL;
     } else {
       auto old = atomicAdd(reinterpret_cast<uint64_t*>(&semaphore), 1);
     }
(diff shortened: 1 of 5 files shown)


jacobhinkle:
!build

jacobhinkle:
!build

runtime/grid_sync.cu
```diff
@@ -75,6 +75,10 @@ __device__ void sync(
       }
 #endif
     }
+    if (last_block) {
+      // Clean up semaphore for re-use in next iteration
+      semaphore = 0ULL;
```
jacobhinkle, line 80: We can't do this without causing deadlocks. The problem is that when `PERSISTENT=true` then all blocks will hang around waiting for the last block to finish finish syncing. This is done by watching for the first bit to flip, which is done by the last block. If the last block flips that first bit then immediately resets the semaphore to zero, then those other blocks will see zero instead of the flipped bit which means they will keep waiting indefinitely. It's an interesting way to create a reusable semaphore, but it's unclear to me how we can automatically clean this without just introducing an expression that resets the semaphore at the end of the kernel when we detect it's needed.

jacobhinkle: For now, I think it is actually safe to reset the semaphore after syncing as long as the PERSISTENT flag is false, since this should indicate that the semaphore won't be re-used.

PR #2087, opened 2024-04-17
insertResharding option for sharding after compute

cowanmeg:
Updates `insertResharding` pass so that any non-lowerable resharding expressions with a single input and output, are split into the local compute expression followed by a resharding expression. 
Example:
```
T1 [i0 i1]  = relu(T0 [DIDx(i0) i1])
---> Insert set after the relu
T2 [DIDx(i0) i1] = relu(T0 [DIDx(i0) i1])
T3 [i0 i1] = set(T2 [DIDx(i0) i1])
```

Resharding expressions with multiple inputs continue to insert sharding expression before the local compute expression.
Example:
```
T2 [i0, i1] = add(T0 [DIDx(i0) i1], T1 [i0, DIDx(i1)]
---> Insert set(s) before add
T3 [i0, i1] = set(T0 [DIDx(i0) i1])
T4 [i0, i1] = set(T1 [i0, DIDx(i1))
T5 [i0, i1] = add(T3 [i0, i1], T4 [i0, i1])
```

Testing:
- Move all multidevice tests that don't require execution into `test_multidevice_reshardings.cpp`
- Add unit test for `insertShardedAxisReording`


tests/cpp/multidevice.cpp
```diff
@@ -16,6 +16,14 @@
 
 namespace nvfuser {
 
+void checkSameShardings(TensorView* ref, TensorView* tv) {
+  ASSERT_EQ(ref->nDims(), tv->nDims());
+  EXPECT_TRUE(ref->getDeviceMesh() == tv->getDeviceMesh());
```
wujingyue, line 21:
```suggestion
  EXPECT_EQ(ref->getDeviceMesh(), tv->getDeviceMesh());
```

tests/cpp/multidevice.cpp
```diff
@@ -16,6 +16,14 @@
...
+  for (auto i : c10::irange(ref->nDims())) {
+    EXPECT_TRUE(ref->axis(i)->getParallelType() == tv->axis(i)->getParallelType());
```
wujingyue, line 23: ditto

CMakeLists.txt
```diff
@@ -522,6 +522,7 @@ list(APPEND JIT_TEST_SRCS
   ${NVFUSER_ROOT}/tests/cpp/test_swizzle.cpp
   ${NVFUSER_ROOT}/tests/cpp/test_tensor_factories.cpp
   ${NVFUSER_ROOT}/tests/cpp/test_unary.cpp
+  ${NVFUSER_ROOT}/tests/cpp/multidevice.cpp
```
wujingyue, line 525: IIUC, this line is added because this PR adds some test utils into `multidevice.cpp` that are used for tests running on single-device. 

It may be better to split the file into 
1. `multidevice.*` (or `sharding.*`), which contains test utils that are useful for both single-device and multi-device, and 
2. `multidevice_test.*`, which contains the declaration and definition of `MultiDeviceTest`. 

This way, MultiDeviceTest and its dependencies don't get unnecessarily pulled into `bin/nvfuser_tests` which only run single-device tests. 

Wdyt?

wujingyue (commented):
Some quick suggestions. Still reviewing...

CMakeLists.txt
samnordmann: I see your point -- keeping separated multi-device test and single device tests. I think bottom-line here it would just mean move `checkSameShardings` away from `multidevice.cpp` (and place it for example in `utils.*`), right?

csrc/multidevice/utils.cpp
```diff
@@ -254,7 +257,13 @@ void insertReshardings(Fusion* fusion) {
   // Remove this after we refactor this as a pre-segmenter pass.
   FusionGuard fg(fusion);
   auto exprs = fusion->exprs();
+  // Replacing expression inputs creates a new expression. Map the old
+  // expression to new expression, so that we can continue iterating.
+  std::unordered_map<Expr*, Expr*> replacement_map = {};
   for (auto expr : exprs) {
+    if (replacement_map.find(expr) != replacement_map.end()) {
+      expr = replacement_map[expr];
```
samnordmann, line 265: I understand the need for this. However, correct me if Im wrong, but I'm afraid there could be cases with multiple replacements composed, right?
e.g. `expr0` -> `expr1` -> `expr2` if the pass changes both the inputs and outputs of the expr. In this case, we would need here to apply the map in a while loop.

Otherwise, maybe it is simpler/more general to make `insertReshardings` recursive

csrc/multidevice/utils.cpp
```diff
@@ -266,40 +275,65 @@ void insertReshardings(Fusion* fusion) {
...
+        output, ir_utils::filterByType<TensorView>(expr->inputs()));
+
+    // Insert resharding set after the expr when there is only one input.
+    // input [expr] output [set] new_output
+    if (!inputs.empty() && expr->inputs().size() == 1) {
+      auto input = *inputs.begin();
+      TensorView* new_output = set(output);
+      auto new_output_map = ir_utils::replaceValInAllExprInputsAndFusionOutputs(
+          output, new_output);
+      replacement_map.merge(new_output_map);
```
samnordmann, line 289: I think it's ok but just checking: can there be key collisions here?

tests/cpp/multidevice.cpp
```diff
@@ -16,6 +16,14 @@
 
 namespace nvfuser {
 
+void checkSameShardings(TensorView* ref, TensorView* tv) {
```
samnordmann, line 19:
```suggestion
void checkStrictlySameShardings(TensorView* ref, TensorView* tv) {
```
Or smth like that, just to convey that the function is not robust to any axis change/transform (e.g. to broadcast, reduction, transpose, view, split, etc), and to disambiguate from functions like `getTvsWithDifferentSharding`.

tests/cpp/test_resharding.cpp
```diff
@@ -224,6 +225,97 @@ TEST_F(ReshardingTest, Detection) {
...
+  a->axis(0)->parallelize(ParallelType::DIDx);
+  b->axis(1)->parallelize(ParallelType::DIDx);
+
+  insertReshardings(&fusion);
+  std::vector<Val*> outputs = fusion.outputs();
+
+  b = outputs[0]->as<TensorView>();
+  Expr* expr = b->definition();
+  EXPECT_TRUE(expr->isA<LoadStoreOp>());
+  EXPECT_TRUE(expr->as<LoadStoreOp>()->opType() == LoadStoreOpType::Set);
```
samnordmann, line 280:
```suggestion
  EXPECT_EQ(expr->as<LoadStoreOp>()->opType(), LoadStoreOpType::Set);
```

tests/cpp/multidevice.cpp
samnordmann: Looking at the tests, I think it would be actually better to have a more robust function, otherwise the tests will break when they'll become more complex. Why not using `getTvsWithDifferentSharding` directly?

samnordmann (commented):
Those unit tests are great to have!! Thanks
What do you think about adding tests where we compose resharding before + resharding after?
and resharding + axis reordering?
It's not a strong request, strenghening those tests could be done later if you want to move on to the matmuls :)

csrc/multidevice/utils.cpp
cowanmeg: Yes, there is definitely an issue. I don't quite see the one you have mentioned, but I do see this one:
```
tv0=expr0(tv)
tv1=expr1(tv)
tv2 = expr2(tv0, tv1)
```

If all these expressions are resharding and require sets, then the heuristic would insert a set after expr0 for tv0, after expr1 for tv1, and before expr2 for tv0/tv1.

If we 'shard after' expr0 and expr1 to form tv0' and tv1' then both iterations would replace expr2 with`ir_utils::replaceValInAllExprInputsAndFusionOutputs` but`replacement_map` will map the original expr2 to the first replacement.

Ideally I would like to just mutate expr2 instead of replace it, because tv0' and tv1' will be identical to the original tv0 and tv1. 
@wujingyue maybe there is a better way to do this kind of mutation?

CMakeLists.txt
wujingyue: That's right. I suggested the name `multidevice_test.*` because it matches the class name `MultideviceTest`.

wujingyue (commented):
Some questions: 

1. Does this work or how do you plan to extend it to work with reduction? E.g. 

```
[x(Serial)] = reduce_sum([x(Serial), y(DIDx), z(Serial)], axis=[1, 2])
```

2. While I understood decomposing these ops to local computation and communication makes them compile, where do they come from in the first place? I'm asking because I suspect it's going to be costly to cover all cases. But it's getting hopeful if we have a small surface to target.

tests/cpp/multidevice.cpp
samnordmann: > BTW is there a reason getTvsWithDifferentSharding is a templated function?

If I remember correctly this was because it was used with `std::vector` and `std::unordered_set` and I grew tired of modifying it every time :) But I am not sure it's relevant anymore so feel free to simplify it

wujingyue:
> from the workload

That's what I was asking about. Consider the end state -- there's a scheduler that automatically splits/marks IterDomains to/as DID. For real workload, do you expect the auto scheduler to introduce DID inconsistency (therefore requiring resharding)
1. potentially anywhere in the model so we need a generic way to resolve DID inconsistency that can't assume certain op types, or 
2. just for certain patterns so we have a smaller target? 

I guess the answer depends on what models to support and partitioning strategies to adopt.

(I well understood this is too far ahead for this PR, thus just a question).

csrc/multidevice/utils.cpp
```diff
@@ -87,10 +87,12 @@ std::pair<std::vector<IterDomain*>, std::vector<IterDomain*>> getShardingChanges
     // Ignore sharded reductions on the output
     // ex. DIDx(i0) -> r(i0) or DIDx(i0) -> r(DIDx(i0))
     // since they don't affect allocation.
-    if (in_root->isDeviceDim() && !out_root->isDeviceDim() &&
-        !out_root->isReduction()) {
+    if (in_root->isDeviceDim() && !in_root->isBroadcast() &&
```
wujingyue, line 90: Can you add a code comment about the isBroadcast check?

csrc/multidevice/utils.h
```diff
@@ -21,12 +22,43 @@ NVF_API bool distributedEnabled();
...
-// Returns the subset of tvs which elements have the same multi-device sharding
-// as ref
+// Returns the subset of tvs which elements have the different multi-device
+// sharding as ref
 template <typename TvIterator>
 std::unordered_set<TensorView*> getTvsWithDifferentSharding(
     TensorView* ref,
-    TvIterator tvs);
+    TvIterator tvs) {
+  std::unordered_set<TensorView*> ret;
```
wujingyue, line 31: Is this just a move or should I also review the implementation?

csrc/multidevice/utils.h
```diff
@@ -21,12 +22,43 @@ NVF_API bool distributedEnabled();
 // Checks that the other non-reduction axis are not parallelized on Didx
 NVF_API bool isSharded(TensorView*);
 
-// Returns the subset of tvs which elements have the same multi-device sharding
-// as ref
+// Returns the subset of tvs which elements have the different multi-device
+// sharding as ref
 template <typename TvIterator>
```
wujingyue, line 27: I understand this is to have one implementation that supports both vector and FilteredView. However, it comes with the downside of increased compile time whenever this implementation is changed and multidevice/utils.h is a quite common header. You may want to consider explicit instantiation to keep the implementation in the `.cpp` file. We can also add an auto conversion from vector<T> to FilteredView<T, T::iterator>, which is likely more work.

csrc/multidevice/utils.cpp
```diff
@@ -266,40 +235,65 @@ void insertReshardings(Fusion* fusion) {
...
-      new_inputs.push_back(new_input);
-      expr = ir_utils::replaceValInExprInputs(expr, input, new_input);
+
+    auto inputs = getTvsWithDifferentSharding(
+        output, ir_utils::filterByType<TensorView>(expr->inputs()));
+
+    // Insert resharding set after the expr when there is only one input.
+    // input [expr] output [set] new_output
+    if (!inputs.empty() && expr->inputs().size() == 1) {
+      auto input = *inputs.begin();
```
wujingyue, line 245:
```suggestion
      TensorView* input = inputs.front();
```

csrc/multidevice/utils.cpp
```diff
@@ -266,40 +235,65 @@ void insertReshardings(Fusion* fusion) {
...
-      // resharding accross all the consumer of the resharded tensor. This way
-      // we could avoid wasteful resharding set insertion.
-      TensorView* new_input = set(input);
-      new_inputs.push_back(new_input);
-      expr = ir_utils::replaceValInExprInputs(expr, input, new_input);
+
+    auto inputs = getTvsWithDifferentSharding(
+        output, ir_utils::filterByType<TensorView>(expr->inputs()));
+
+    // Insert resharding set after the expr when there is only one input.
```
wujingyue, line 242: Why is this better than inserting the resharding before? It sounds like we are lacking some heuristics to look at tensor sizes?

csrc/multidevice/utils.cpp
```diff
@@ -254,7 +217,13 @@ void insertReshardings(Fusion* fusion) {
   // Remove this after we refactor this as a pre-segmenter pass.
   FusionGuard fg(fusion);
   auto exprs = fusion->exprs();
+  // Replacing expression inputs creates a new expression. Map the old
+  // expression to new expression, so that we can continue iterating.
```
wujingyue, line 221: `fusion->exprs()` creates a new list: https://github.com/NVIDIA/Fuser/blob/9780f666e55ef6a2d76b8da1396a2227c8c37e69/csrc/fusion.h#L196. So why was continuing iterating `exprs` a problem?

cowanmeg:
> > from the workload
> 
> That's what I was asking about. Consider the end state -- there's a scheduler that automatically splits/marks IterDomains to/as DID. For real workload, do you expect the auto scheduler to introduce DID inconsistency (therefore requiring resharding)
> 
> 1. potentially anywhere in the model so we need a generic way to resolve DID inconsistency that can't assume certain op types, or
> 2. just for certain patterns so we have a smaller target?
> 
> I guess the answer depends on what models to support and partitioning strategies to adopt.

^ Yup I think that is the key take away. IMO we should also not limit ourselves to existing strategies because we want to be flexible and support new ones.  

> (I well understood this is too far ahead for this PR, thus just a question).

csrc/multidevice/utils.h
cowanmeg: Just a move!

csrc/multidevice/utils.cpp
cowanmeg: Yes, we need to add better heuristics of where to insert the sets. Same for your below comment. Right now we don't look at minimizing communication size or what the effect is on the surrounding computation.

csrc/multidevice/utils.cpp
cowanmeg: `ir_utils::replaceValInAllExprInputsAndFusionOutputs` deletes and creates new expressions, so `fusion->exprs()` list holds pointers to deleted expression. We can work around it, by updating the expression pointers as we mutate them, but it would be nice if we could simply mutate the expression

csrc/multidevice/utils.cpp
wujingyue: Got it. Can you add a code comment to say heuristics need to be retuned in the future? While no heuristics are perfect, the decisions as they are seem a bit too arbitrary and can cause confusion without a code comment.

wujingyue:
> ^ Yup I think that is the key take away. IMO we should also not limit ourselves to existing strategies because we want to be flexible and support new ones.

That's fair. It means we will have to think about a generic way to handle all these possibilities; otherwise it's going to be endless hole-patching.

csrc/multidevice/utils.cpp
```diff
@@ -266,40 +235,65 @@ void insertReshardings(Fusion* fusion) {
...
+
+    auto inputs = getTvsWithDifferentSharding(
+        output, ir_utils::filterByType<TensorView>(expr->inputs()));
+
+    // Insert resharding set after the expr when there is only one input.
+    // input [expr] output [set] new_output
+    if (!inputs.empty() && expr->inputs().size() == 1) {
+      auto input = *inputs.begin();
+      TensorView* new_output = set(output);
+      auto new_output_map = ir_utils::replaceValInAllExprInputsAndFusionOutputs(
```
wujingyue, line 259: `new_output->definition()` is also a use of `output` at the point. Does replaceValInAllExprInputsAndFusionOutputs also replace `output` in **that** set, leading to a self-loop?

csrc/multidevice/utils.cpp
wujingyue: Instead of using `replacement_map`, can we do replacement in two passes? The first pass only replaces the inputs and therefore can be done in topological order as in `fusion->exprs()`. This is the same as before this PR. The second pass only replaces the outputs and therefore can be done in the reverse order of `fusion->exprs()`.

csrc/multidevice/utils.cpp
cowanmeg: Nope! I printed out what expressions get modified and it did not include the set. 

I think a TensorView's `uses()` doesn't get recalculated while updating use info? https://github.com/NVIDIA/Fuser/blob/72bc243dfa5e4a7130d80e04d76fed8b4cf5f842/csrc/ir/base_nodes.cpp#L92

TBH, I don't know if that check was intended to cover my use case, or I just got lucky and it does?

csrc/multidevice/utils.cpp
```diff
@@ -186,6 +187,18 @@ void shardAllLike(TensorView* ref, std::vector<TensorView*> tvs) {
         ref, tvs, {ParallelType::DIDx, ParallelType::Serial});
   }
 }
+
+void updateExprList(
```
samnordmann, line 191: nice!

csrc/multidevice/utils.cpp
cowanmeg: This worked! I updated all the passes that changed downstream operations to iterate through fusion->exprs() backwards, so no more tracking replaced exprs.

csrc/multidevice/utils.cpp
wujingyue: That's bizarre :) Let me run a quick test and then update this thread.

wujingyue: During the reset, https://github.com/NVIDIA/Fuser/blob/dad05c3f5fc6eee85057380cdd0643ee84bb0475/csrc/fusion.cpp#L617 only looks at "reachable" expressions so the use of `output` in `new_output->definition()` (the `set`) is not replaced. So it's safe to use `replaceValInAllExprInputsAndFusionOutputs` here. TIL!

csrc/multidevice/utils.cpp
```diff
@@ -220,9 +183,92 @@ void shardAllLike(TensorView* ref, std::vector<TensorView*> tvs) {
...
+    scheduler_utils::parallelizeAllLike(
+        ref, tvs, {ParallelType::DIDx, ParallelType::Serial});
+  }
+}
+
+// TODO: We can either reshard the inputs of a resharding expression or
+// the outputs. Currently, we reshard the outputs when there is only one
+// input, otherwise we reshard the inputs. This heuristic should be smarter
+// and attempt to minimize communication.
+bool reshardAfter(Expr* expr) {
```
wujingyue, line 195:
```suggestion
bool shouldReshardAfter(Expr* expr) {
```

Naming nit: `reshardAfter` gives readers the wrong impression that the function actually inserts resharding.

tests/cpp/test_resharding.cpp
```diff
@@ -224,6 +224,100 @@ TEST_F(ReshardingTest, Detection) {
...
+  c->axis(1)->parallelize(ParallelType::DIDx);
+
+  insertReshardings(&fusion);
+  int num_inner_reshardings = 0;
+  for (auto expr : fusion.exprs()) {
+    if (isResharding(expr) && isInnerResharding(expr)) {
+      num_inner_reshardings++;
+    }
+  }
+  EXPECT_TRUE(num_inner_reshardings > 0);
```
wujingyue, line 311:
```suggestion
  EXPECT_GT(num_inner_reshardings, 0);
```

tests/cpp/test_resharding.cpp
```diff
@@ -224,6 +224,100 @@ TEST_F(ReshardingTest, Detection) {
...
+
+  insertReshardings(&fusion);
+  std::vector<Val*> outputs = fusion.outputs();
+
+  b = outputs[0]->as<TensorView>();
+  Expr* expr = b->definition();
+  EXPECT_TRUE(expr->isA<LoadStoreOp>());
+  EXPECT_EQ(expr->as<LoadStoreOp>()->opType(), LoadStoreOpType::Set);
+  std::vector<TensorView*> tvs = {expr->inputs()[0]->as<TensorView>()};
+  EXPECT_TRUE(getTvsWithDifferentSharding(a, tvs).empty());
```
wujingyue, line 283:
```suggestion
  EXPECT_THAT(getTvsWithDifferentSharding(a, tvs), IsEmpty());
```

https://google.github.io/googletest/reference/matchers.html

The practical benefit is that the error message will print out the content of the vector if the check fails.

wujingyue (approved):

cowanmeg:
!build --dist

cowanmeg:
!build --dist

PR #2088, opened 2024-04-17
Matmul Python Benchmarks

### LINKED ISSUE #2082: Adding Python Benchmarks for Matmuls
Ask Jacob to get the set of problem sizes for matmul.

Priya2698:
Fixes #2082


benchmarks/python/test_matmul.py
```diff
@@ -0,0 +1,55 @@
...
+    benchmark,
+    config: tuple,
+    dtype: torch.dtype,
+    disable_validation: bool,
+    disable_benchmarking: bool,
+):
+    clear_cuda_cache()
+    m, n, k, layout = config
+
+    # a_dtype, b_dtype, out_dtype = [precision_types[prec] for prec in precision]
```
jacobhinkle, line 38: Good idea but we can stick with single input/output dtype for now, so we don't need this line or `precision_types`.

benchmarks/python/test_matmul.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+precision_types = {"H": torch.float16, "S": torch.float32, "T": torch.bfloat16}
+
+
+def load_matmul_problems():
+    with open(os.path.join(os.path.dirname(__file__), "matmul_problems.csv"), "r") as f:
+        reader = csv.reader(f)
+        next(reader, None)  # skip header row
+        return list((int(m), int(n), int(k), layout) for m, n, k, layout in reader)
```
jacobhinkle, line 33: We could also add batch size to this.

benchmarks/python/test_matmul.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+@pytest.mark.parametrize("config", load_matmul_problems())
```
jacobhinkle, line 36: We might also want to add the `ids=` kwarg here so that the benchmark names have easily filterable names.

benchmarks/python/test_matmul.py
jacobhinkle: It would also be interesting to see if there's a perf difference between a 2D matmul and a bmm with size 1 in our system. Of course as far as parametrization, we also do not particularly want to have 32,000 benchmarks run automatically necessarily.

Priya2698: Batched GEMMs are not supported by the API rn. We only accept 2D matmuls

Priya2698: I can open an issue to start development for this,

benchmarks/python/test_matmul.py
```diff
@@ -0,0 +1,55 @@
+import pytest
+from nvfuser import FusionDefinition, DataType
+from nvfuser.pytorch_utils import torch_dtype_to_nvfuser_dtype
+from .core import run_benchmark, clear_cuda_cache
+import torch
+from .global_params import generate_input_sizes, FLOAT_DTYPES, PROMOTE_DTYPES
+
+def matmul_fusion(fd: FusionDefinition, dtype: DataType) -> None:
+    # Decide contiguity based on layout
+    a = fd.define_tensor(
```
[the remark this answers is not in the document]
Priya2698: So, `fd.from_pytorch`, should be able to allocated the inputs correctly?

benchmarks/python/test_matmul.py
Priya2698: Created Issue #2092

Priya2698:
For future, we should probably have a separate benchmark for baselines (eager / torchcompile). 
While we run through ATen right not, those will be required later. Wdyt? @jacobhinkle. I can add them to this PR.

jacobhinkle:
Maybe we should do that in another PR since the current one is enough to benchmark aten vs nvfuser using env vars.

benchmarks/python/test_matmul.py
jacobhinkle: Sounds good. We should remember to enable this once it's supported. Thanks for opening the issue.

Priya2698:
@jacobhinkle Should we merge this PR?

jacobhinkle (approved):
Full speed ahead! Thanks for the changes.

PR #2090, opened 2024-04-17
Python API for linear and tests

### LINKED ISSUE #2057: Implement `ops.linear` Primitive
[Could we enumerate how this gets plumbed from Interface to IR]

We are working on plumbing down a call to linear from Python to our backends. For now, the default execution will rely on the aten linear implementation via our expression evaluator. This work is in the spirit as the implementation of the mamtul API in Python.

Linear[@Priya2698's PR in flight: https://github.com/NVIDIA/Fuser/pull/2050 ]

Issues to discuss:
1. Our logic for dispatching to at::matmul, admm, or linear could be (filled in by Priya below):

at::linear -> If the detected shape is [M,K] x [N, K].
at::addmm -> If the detected shape is [M,K] x [N, K] and bias != nullptr
at::matmul -> If the detected shape is [M,K] x [N, K] and bias == nullptr

I also agree with this. It should be able to cover the cases:
A. We call ops.matmul 
B. We call ops.linear (with 2 or 3 operands)
C. User creates a fusedMultiplySum followed by bias (admm).

2. The next question is how do we detect shape.  Priya mentioned:
While pattern matching, we look at the shapes of the unbroadcasted inputs to the MmaOp.

@jacobhinkle  also mentioned using getProblemIterDomains (https://github.com/NVIDIA/Fuser/blob/52692ed5347dc308b93a7f35acdf2c4123c10c1f/csrc/scheduler/mma_utils.cpp#L1056-L1089) to achieve this. I was also leaning towards this approach, and I was wondering if we can just call getMmaLayout (https://github.com/NVIDIA/Fuser/blob/52692ed5347dc308b93a7f35acdf2c4123c10c1f/csrc/scheduler/mma_utils.cpp#L1056-L1089) to get the shapes on the inputs. Thought the function name mentions layout, it's very much just the logical shapes of the tensors.

3. Segmentation doesn't work properly when the bias is of dimension {n} as is the case for linear.
This is probably solved. @jjsjann123  noticed this code:
(https://github.com/NVIDIA/Fuser/blob/618c4b79a053692bf83ebb32aa19dcc907ebb534/csrc/scheduler/mma_utils.cpp#L1239C1-L1247C8). It should be modified to just `!has_k`

protonu:
This add the Python API for linear and tests.

Fixes #2057


protonu:
!build

csrc/ops/composite.cpp
```diff
@@ -89,6 +89,40 @@ TensorView* matmul(TensorView* a, TensorView* b) {
...
+    NVF_CHECK(bias->nDims() == 1, "Only 1-D Bias are supported in Linear!");
+    auto* bias_with_cast = maybeCastOp(output->getDataType().value(), bias);
+    auto* bcast_bias = broadcast(bias_with_cast, {true, false});
+    auto* bias_output = add(output, bcast_bias);
+    return maybeCastOp(a->getDataType().value(), bias_output);
+  }
+  return maybeCastOp(a->getDataType().value(), output);
+}
+
+TensorView* linear(TensorView* a, TensorView* b) {
```
jacobhinkle, line 122: Why is this not handled by optional argument `bias = nullptr` in the function above?

csrc/ops/composite.h
```diff
@@ -57,6 +57,15 @@ TensorView* matmul(TensorView* a, TensorView* b);
 // can be used to create mamtuls without a cast operation following it.
 TensorView* matmul(TensorView* a, TensorView* b, bool cast_output_to_input);
 
+// Linear functions which takes in two tensors of shapes A[M,K] and
+// B[N,K]. Takes in a options bias of shape [N] and performs
+// out = A * B_Transpose + bias. The output dtype matches the dtype
+// ofthe inputs which should match.
+TensorView* linear(TensorView* a, TensorView* b, TensorView* bias);
+// This is an implemenation details to reflect when linear is called
```
jacobhinkle, line 65:
```suggestion
// This is an implemtenation details to reflect when linear is called
```
Also please explain why this is needed (is it?).

csrc/python_frontend/python_bindings.cpp
```diff
@@ -1153,6 +1153,45 @@ void initNvFuserPythonBindings(PyObject* module) {
...
+                  static_cast<
+                      TensorView* (*)(TensorView*, TensorView*, TensorView*)>(
+                      linear)));
+        } else {
+          fd->defineRecord(new OpRecord<TensorView*, TensorView*, TensorView*>(
+              {fd->recordingState(arg1()), fd->recordingState(arg2())},
+              {fd->recordingState(output())},
+              ("ops.linear"),
+              serde::RecordType::Binary_TV,
+              static_cast<TensorView* (*)(TensorView*, TensorView*)>(linear)));
```
jacobhinkle, line 1186: Can you use a lambda here and make the `bias=nullptr` default arg?
```suggestion
              static_cast<TensorView* (*)(TensorView*, TensorView*)>([](TensorView* a, TensorView* b) { return linear(a, b);})));
```

tests/python/pytest_input_generators.py
```diff
@@ -1499,11 +1499,17 @@ def multiply_range(maximum, step):
             map(pow, itertools.repeat(step, num_steps), range(1, num_steps + 1))
         )
 
+    is_linear = True if op.name == "linear" else False
```
jacobhinkle, line 1502:
```suggestion
    is_linear = op.name == "linear"
```

tests/python/test_python_frontend.py
```diff
@@ -2402,6 +2402,55 @@ def fusion_func(fd: FusionDefinition, inps) -> None:
...
+                nvf_out, _ = self.exec_nvfuser(
+                    partial(fusion_func, inputs=input), input
+                )
+                eager_out = (
+                    F.linear(inputs[0], inputs[1], bias)
+                    if use_bias
+                    else F.linear(inputs[0], inputs[1])
+                )
+                fp16_nvf_out = nvf_out[0]
+                self.assertEqual(eager_out, fp16_nvf_out)
```
jacobhinkle, line 2452: Just a suggestion to make this a little more readable:
```suggestion
        def fusion_func(fd: FusionDefinition, inp: torch.Tensor, wt: torch.Tensor, bias: torch.Tensor | None) -> None:
            t0 = fd.from_pytorch(inp)
            t1 = fd.from_pytorch(wt)
            if bias is not None:
                t2 = fd.from_pytorch(bias)
                t_out = fd.ops.linear(t0, t1, t2)
            else:
                t_out = fd.ops.linear(t0, t1)
            fd.add_output(t_out)

        for inp, wt in [inputs_mk_nk, inputs_mk_kn, inputs_km_nk, inputs_km_kn]:
            for use_bias in [True, False]:
                input_tensors = (inp, wt, bias) if use_bias else (inp, wt)
                nvf_out, _ = self.exec_nvfuser(
                    partial(fusion_func, inp=inp, wt=wt, bias=bias), input_tensors
                )
                eager_out = F.linear(input=inp, weight=wt, bias=bias)
                fp16_nvf_out = nvf_out[0]
                self.assertEqual(eager_out, fp16_nvf_out)
```

csrc/python_frontend/python_bindings.cpp
```diff
@@ -1153,6 +1153,45 @@ void initNvFuserPythonBindings(PyObject* module) {
...
+            self.validUse(), "Attempting to add to a completed definition!");
+        FusionDefinition* fd = self.fusion_definition;
+        Tensor output = fd->defineTensor(arg1.dims);
+
+        if (bias.has_value()) {
+          fd->defineRecord(
+              new OpRecord<TensorView*, TensorView*, TensorView*, TensorView*>(
+                  {fd->recordingState(arg1()),
+                   fd->recordingState(arg2()),
+                   fd->recordingState(bias.value()())},
```
jacobhinkle, line 1173: Is it not possible to have a missing value for bias here? i.e. can we just insert `nullptr` here instead of having two separate `OpRecord` definitions?

jacobhinkle: Nevermind. Of course these are `State`s, so there's no way to just say this argument should be nullptr.

protonu: Not quite a nullptr, but something like this:

```
 auto bias_state = bias.has_value()
            ? fd->recordingState(bias.value()())
            : State(0, serde::StateType::None);
```

And then we handle the None state. This pattern exists in our current `python_bindings.cpp`, and to make it work we'll need to create a new class like `OpRecord` for example, `LinearOpRecord`. Creating another class was seen as not really desirable. (https://github.com/NVIDIA/Fuser/blob/5c674cde66101a497bbaeed097624a60eefae07a/csrc/python_frontend/python_bindings.cpp#L2699-L2704)


The limitations we hit are:

```
struct OpRecord : RecordFunctor {
  OpRecord(
      std::vector<State> _args,
      std::vector<State> _outputs,
      std::string _name,
      serde::RecordType record_type,
      std::function<OutType(ArgTypes...)> fusion_op)
      : RecordFunctor(
            std::move(_args),
            std::move(_outputs),
            _name,
            record_type),
        fusion_op_(fusion_op)
```
Where the fusion_op has the same number input args as the template args. We thought about getting around this with `std::optional` but this doesn't quite work out because of:
https://github.com/NVIDIA/Fuser/blob/5c674cde66101a497bbaeed097624a60eefae07a/csrc/python_frontend/fusion_record.h#L302-L307

Another point to note is that, the two OpRecords will have different `serde::RecordType`, Binary and Ternary.

csrc/python_frontend/python_bindings.cpp
protonu: Let me try this out. Might need a few more modification to make a lambda work.

csrc/ops/composite.cpp
protonu: The main reasons for this are here:
https://github.com/NVIDIA/Fuser/pull/2090#discussion_r1569150816

But I had tried std::optional, but didn't consider a lambda as you mentioned below. I'll try out the lambda.

csrc/python_frontend/python_bindings.cpp
jacobhinkle: Actually there is `StateType::None`:
https://github.com/NVIDIA/Fuser/blob/9c220186f35a8478dbcd95f178f3749ebfaa1369/csrc/serde/fusion_cache.fbs#L16-L22
What happens if you just create one of these objects for the null bias if `!bias.hasValue()`? (don't merge this change)
```suggestion
                   fd->recordingState(bias.has_value() ? bias.value()() : State{.index=0, .stype=serde::StateType::None())},
```

jacobhinkle: (you beat me to it). I don't understand why it requires a new class. `StateType::None()` should just imply we pass a `nullptr` for that argument right? So the number of args should match.

jacobhinkle: It seems like this would be fixed if `fd.getFusionState` would return `nullptr` whenever `stype=None`.

protonu: @jjsjann123, @rdspring1 , @kevinstephano Tagging you guys since you may worked on this.
What would be the best way to pass in optional arguments to an Op like Linear.
My ideas were:
1. Create a new Record, like LinearOpRecord (cons: more work and code bloat)
2. Different `linear` functions (possibly via a lambda as @jacobhinkle pointed).
3. Is there another way?

I don't want to change the general way we handle default/optional arguments in our python fronted in this PR though.

protonu: > It seems like this would be fixed if fd.getFusionState would return nullptr whenever stype=None.
I am going to try this out.

tests/python/test_python_frontend.py
```diff
@@ -2402,6 +2402,51 @@ def fusion_func(fd: FusionDefinition, inps) -> None:
...
+            t1 = fd.from_pytorch(wt)
+            if bias is not None:
+                t2 = fd.from_pytorch(bias)
+                t_out = fd.ops.linear(t0, t1, t2)
+            else:
+                t_out = fd.ops.linear(t0, t1)
+            fd.add_output(t_out)
+
+        for inp, wt in [inputs_mk_nk, inputs_mk_kn, inputs_km_nk, inputs_km_kn]:
+            for use_bias in [True, False]:
```
wujingyue, line 2441: Instead of for loops, I'd parameterize the test: https://docs.pytest.org/en/7.1.x/how-to/parametrize.html. This way, when an error shows up, it's clear from the test name which config is failing.

wujingyue: I'm no longer sure we run these tests via pytest. If we don't, unittest has a way to create subtests, replacing iterations: https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests

tests/python/test_python_frontend.py
```diff
@@ -2402,6 +2402,51 @@ def fusion_func(fd: FusionDefinition, inps) -> None:
...
+                input_tensors = (inp, wt, bias) if use_bias else (inp, wt)
+                nvf_out, _ = self.exec_nvfuser(
+                    partial(fusion_func, inp=inp, wt=wt, bias=bias), input_tensors
+                )
+                eager_out = F.linear(input=inp, weight=wt, bias=bias)
+                fp16_nvf_out = nvf_out[0]
+                self.assertEqual(eager_out, fp16_nvf_out)
```
wujingyue, line 2448: I'm surprised assertEqual works out. I thought we'd need torch.testing.assert_close to tolerate some accumulation-order differences.

jacobhinkle: It probably works in this case because the problem size is small enough that it is never potentially using any kind of nondeterministic grid reduction; it can probably compute everything in a single tile and it's too small to use split-k. So then it is run to run repeatable. Bigger inputs might break then. Better to just use assert_close.

csrc/ops/composite.cpp
```diff
@@ -89,6 +89,40 @@ TensorView* matmul(TensorView* a, TensorView* b) {
...
+  bcast_dims.at(bcast_dims.size() - 3) = true;
+  auto* tv1b = broadcast(b, bcast_dims);
+
+  NVF_CHECK(
+      a->getDataType().value() == b->getDataType().value(),
+      "data types of inputs to matmul don't match");
+
+  auto* output = fusedMultiplySum(tv0b, tv1b, {-1});
+  if (bias) {
+    NVF_CHECK(bias->nDims() == 1, "Only 1-D Bias are supported in Linear!");
```
Priya2698, line 113: We should also check that the shape of bias is `N,`

Priya2698: Also, that the bias is of the same type as the inputs.

csrc/python_frontend/python_bindings.cpp
kevinstephano: I know it is slightly ugly to have two `linear` functions on the C++ side but I would like to move forward with just using a 2 argument and a 3 argument signature so we can move forward.

csrc/ops/composite.cpp
protonu: > We should also check that the shape of bias is N,
What would be the way to do this for symbolic tensors?

csrc/python_frontend/python_bindings.cpp
jjsjann123: Not against @kevinstephano 's suggestion on keeping it as-is for faster movement. @jacobhinkle 's suggestion doesn't seem to suggest lot more work... We should at least follow up with that as a clean up.

csrc/ops/composite.cpp
jjsjann123: can we actually check the shape of a tensor?

jjsjann123: nitpick: is `bias->nDims() == 1` a hard requirement? shouldn't we be able to just do normal broadcasting on `bias`?

I know torch doc suggests that bias should be shape `(out_features)` https://pytorch.org/docs/stable/generated/torch.nn.functional.linear.html#torch.nn.functional.linear

But implementation seems to disagree: 
```
>>> a = torch.randn(8, 2, 3, 4, device="cuda")
>>> b = torch.randn(5, 4, device="cuda")
>>> bias = torch.randn(8, 2, 3, 5, device="cuda")
>>> torch.ops.aten.linear(a, b, bias).shape
torch.Size([8, 2, 3, 5])
>>> torch.nn.functional.linear(a, b, bias).shape
torch.Size([8, 2, 3, 5])
```

csrc/ops/composite.cpp
```diff
@@ -89,6 +89,40 @@ TensorView* matmul(TensorView* a, TensorView* b) {
   return matmul(a, b, true /* cast output to input dtype */);
 }
 
+TensorView* linear(TensorView* a, TensorView* b, TensorView* bias) {
+  NVF_CHECK(
+      a->nDims() == b->nDims(),
+      "The number of dimension of A and B do not match");
+  NVF_CHECK(a->nDims() == 2, "Only 2-D A/B Tensors are supported in Linear!");
```
jjsjann123, line 96: similarly, can we add a TODO here as well to expand the support?

protonu: I went based on the API (as you saw as well). I can a TODO here.

protonu: >Not against @kevinstephano 's suggestion on keeping it as-is for faster movement. @jacobhinkle 's suggestion doesn't seem to suggest lot more work... We should at least follow up with that as a clean up.

 I don't mind doing - superficially it's not a lot of work - but this gets stuck in SerDe. We can chat about it.

csrc/ops/composite.cpp
```diff
@@ -89,6 +89,40 @@ TensorView* matmul(TensorView* a, TensorView* b) {
   return matmul(a, b, true /* cast output to input dtype */);
 }
 
+TensorView* linear(TensorView* a, TensorView* b, TensorView* bias) {
+  NVF_CHECK(
+      a->nDims() == b->nDims(),
+      "The number of dimension of A and B do not match");
```
jacobhinkle, line 95: This actually does not match [F.linear](https://pytorch.org/docs/stable/generated/torch.nn.functional.linear.html#torch.nn.functional.linear). `a` merely needs at least one dimension and `b` must have either 1 or 2 dimensions.

csrc/ops/composite.cpp
jacobhinkle: > We should also check that the shape of bias is `N,`

We can't check that at definition. It will be assumed to be true then when we receive inputs, if it is untrue we will display a somewhat cryptic error message in `ExpressionEvaluator::propagateBoundValuesThroughExactMaps`. :face_with_spiral_eyes:

jacobhinkle: > But implementation seems to disagree:

Interesting. So this should just be "torch-broadcastable" to match the output, i.e. bias can be any dimension up to `a->nDims()` and we should do something like `ops::maybeBroadcast({a, bias})`:
https://github.com/NVIDIA/Fuser/blob/356a59d7f9b99238cd6c330b00c883903fbee8f8/csrc/ops/utils.cpp#L296. But whereas `maybeBroadcast` inserts the broadcasts as new inner dimensions, we should instead add broadcasts in the outer dims. BTW @jjsjann123 do you happen to know why our implicit broadcasting works differently from pytorch?
```python
[ins] In [1]: import torch
[nav] In [2]: x = torch.randn([4, 5])
[nav] In [3]: y = torch.randn([5])
[ins] In [4]: x + y
Out[4]: 
tensor([[-1.8200,  2.8528, -1.2850, -2.3024, -0.1085],
        [-2.7299, -0.3150,  0.2120,  1.2288, -0.4108],
        [-0.8917, -0.1539, -0.3934,  0.6995, -0.1939],
        [-1.7878,  0.4470, -2.2318,  0.1037,  0.4602]])
```

csrc/python_frontend/python_bindings.cpp
rdspring1: Can we have optional arguments in our cpp Fusion Definition? 

I was seeing `RuntimeError: kernel->inputs().size() <= args.size()` with this PRs ToT.
Here was the failing case:
```python
import torch

    T0 = fd.define_tensor(shape=[-1, -1], contiguity=[True, True], dtype=DataType.Half, is_cpu=False, stride_order=[1, 0])
    T1 = fd.define_tensor(shape=[-1, -1], contiguity=[True, True], dtype=DataType.Half, is_cpu=False, stride_order=[1, 0])
    T2 = fd.define_tensor(shape=[-1], contiguity=[True], dtype=DataType.Half, is_cpu=False, stride_order=[0])
    T3 = fd.ops.linear(T0, T1, T2)
    fd.add_output(T3)

    nvfuser_fusion_id0(fd)

inputs = [
    torch.randn((192,), dtype=torch.float16, device='cuda:0').as_strided((24, 8), (8, 1)),
    torch.randn((128,), dtype=torch.float16, device='cuda:0').as_strided((16, 8), (8, 1)),
]
fd.execute(inputs)
```

jjsjann123: You are missing bias in `inputs` :wink:

jjsjann123: you defined three input tensor in your FusionDefinition.

jacobhinkle: That error looks legit. If we provide a bias in the fusion definition, we need to also supply it to `fd.execute`.

protonu: That errors is fixed, let me update the PR. This showed since I merged some of Jacob's comments on the browser

tests/python/test_python_frontend.py
protonu: Thank @wujingyue, please take another look at the test.

csrc/python_frontend/python_bindings.cpp
protonu: @rdspring1 can you try it again. It should be fixed now.

csrc/ops/composite.cpp
kevinstephano: Let's move forward with 2 C++ functions for `linear` and get the PR merged!

csrc/ops/composite.cpp
jacobhinkle: Oh, I think I misread this code! It is actually broadcasting the outer dims.

jjsjann123: forwarding Kevin's comment, looks like we are just matching thunder's checks: https://github.com/NVIDIA/Fuser/issues/2057#issuecomment-2064838196

jacobhinkle (approved):
Changes look good. See below.

tests/python/test_python_frontend.py
```diff
@@ -2407,6 +2407,8 @@ def test_linear(self):
         n = 16
         k = 8
         bias = torch.randn(n, device="cuda", dtype=torch.float16)
+        bias2d = torch.rand(m, n, device="cuda", dtype=torch.float16)
```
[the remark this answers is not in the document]
protonu: That's a great comment. I tried it out and ran into error in our segmentation logic which can't detect a 0-dim bias. I'll close this PR, and in follow-up PR(s) handle the higher than 2D A matrix and the 0-dim bias.

protonu:
!build

jjsjann123 (approved):

tests/python/test_python_frontend.py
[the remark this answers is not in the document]
jjsjann123: Can we get an issue tracking this, along with the other unsupported shapes/sizes?

to be fair, our support isn't really covering all `F.linear` configuration and that's probably fine. We can also reject unsupported configuration via checker function in registration. So non of these are really blocking issues.

csrc/ops/composite.cpp
```diff
@@ -89,6 +90,44 @@ TensorView* matmul(TensorView* a, TensorView* b) {
...
+      a->getDataType().value() == b->getDataType().value(),
+      "data types of inputs to matmul don't match");
+
+  auto* output = fusedMultiplySum(tv0b, tv1b, {-1});
+  if (bias) {
+    NVF_CHECK(
+        (bias->nDims() <= a->nDims()), "bias should be broadcastable to A");
+    NVF_CHECK(
+        a->getDataType().value() == bias->getDataType().value(),
+        "bias doesn't match input/weight");
```
Priya2698, line 118:
```suggestion
Expected bias to be of the same dtype as inputs/weights
```

Priya2698:
Can we also update the version? @protonu @jjsjann123

Priya2698 (approved):
LGTM

protonu:
!build

protonu:
!build

protonu:
!build

protonu:
!build

PR #2096, opened 2024-04-17
Move some tests from test_multidevice_sharding to test_sharding. 

wujingyue:
Those tests don't need multiple GPUs and can run without MPI.

With this PR, test_sharding.cpp is built into nvfuser_tests and runs without MPI. test_multidevice_sharding.cpp remains in test_multidevice and runs with MPI.


wujingyue:
!build --dist

cowanmeg (approved):
Looks good! Let's merge in test_multidevice_resharding into this file as well? It also doesn't require execution.

wujingyue:
> Looks good! Let's merge in test_multidevice_resharding into this file as well? It also doesn't require execution.

Are you sure? 
https://github.com/NVIDIA/Fuser/blob/356a59d7f9b99238cd6c330b00c883903fbee8f8/tests/cpp/test_multidevice_sharding.cpp#L154 seems to say it needs mpirun; otherwise Communicator wouldn't be able to read its device ID from mpi env vars.

wujingyue:
!build --dist

cowanmeg:
I meant this test file! https://github.com/NVIDIA/Fuser/blob/main/tests/cpp/test_multidevice_resharding.cpp
(Also, I wasn't thinking when I named these files sharding and resharding...)

PR #2097, opened 2024-04-17
Update version.txt

Priya2698:
We did not bump up the version when we added the `fd.ops.matmul` API. This will be used to check the nvFuser version in Thunder when enabling matmul,
Thanks to @jjsjann123 for catching this.


jjsjann123 (approved):
:shipit:

PR #2100, opened 2024-04-17
Fix comments.

wujingyue:


cowanmeg (approved):
FYI, NVFUSER_DISTRIBUTED is a dependent option that defaults to true when pytorch is built with USE_DISTRIBUTED

tests/cpp/test_multidevice_pipeline.cpp
```diff
@@ -42,13 +42,11 @@ namespace nvfuser {
...
-   Then simply run the tests on several processes, for example using mpirun
-   on a node having at least 6 GPUs,
-   e.g.: mpirun -np 6 build/nvfuser_tests
-   --gtest_filter=PipelineTest.Pipeline
-*/
+// To run the following tests on several devices, pytorch must be installed with
+// the flag USE_DISTRIBUTED=1 and nccl support. With that, nvFuser is built by
+// default with NVFUSER_DISTRIBUTED defined.  Then simply run the tests on
+// several processes. For example, run `mpirun -np 6 build/test_multidevice
+// --gtest_filter=PipelineTwoStages*` on a node having at least GPUs.
```
samnordmann, line 49: The "6" seems to be gone
```suggestion
// --gtest_filter=PipelineTwoStages*` on a node having at least 6 GPUs.
```
If there are 6 processes, it needs to have 6 GPU otherwise it will complain.
Also on multi node there is more requirements, and more generally, if not using mpirun then the user needs to set some environment variables manually. 

If we want to be precise here, we also need to add the following requirements:
```
/*
- Set automatically by mpirun (so there’s probably no need to manually set them):
     o	OMPI_COMM_WORLD_RANK
     o	OMPI_COMM_WORLD_SIZE
     o	OMPI_COMM_WORLD_LOCAL_RANK
     o	OMPI_COMM_WORLD_LOCAL_SIZE
- Optional for single-node system:
     o	MASTER_ADDR containing the address of one of the node (can be chosen arbitrarily at this point) which will be the server for the c10d::TCPStore
- Optional:
     o	MASTER_PORT port to be used by the c10d::TCPStore. If not set, we will use by default c10d::TCPStoreOptions::kDefaultPort (which equals 29500 currently)

*/
```

wujingyue: My bad -- I added the 6 back. 

> we also need to add the following requirements:

LGTM! Since you know these requirements better than I do, I'll let you add them to the right place. test_multidevice_pipeline.cpp might not be the best place for them because we have other `test_multidevice_*.cpp`. CMakeLists.txt could be a better place.

samnordmann: No problem, I'll add that

However, "a node" has been erased in the sentence, but I think it is an important detail and so it should be reverted.

Actually I would replace 
> Then simply run the tests on several processes. For example, run`

By 

> Then, on a node with at least 6 GPUs, run the test using mpirun:

(conveying the idea that `mpirun` _needs_ to be used)

PR #2103, opened 2024-04-18
Update`dropout-layernorm/rmsnorm` benchmarks

Priya2698:
1. Update the fusion definition to have 2 distinct inputs: `layernorm/rmsnorm (inp2 + dropout(inp1))`.
2. Add torch baselines
3. Add iobyte computation based on input/output dict listing size and dtype. This should make the computation clearer to the users and less error-prone

Issue #1668

### DIFF
--- a/benchmarks/python/core.py
+++ b/benchmarks/python/core.py
@@ -8,7 +8,7 @@ import torch
 from torch.autograd import DeviceType
 from torch.profiler import profile, ProfilerActivity
 from typing import List, Callable, Union, Tuple
-
+import numpy as np
 
 def get_device_properties() -> Tuple[int, float]:
     """
@@ -144,6 +144,20 @@ def clear_cuda_cache() -> None:
 # Backward function for torch baseline benchmarks.
 def unary_bwd_torch(inputs: List):  # [output, grad_out]
     inputs[0].backward(inputs[1], retain_graph=True)
+
+
+def compute_total_iobytes(tensor_props: dict):
+    '''
+    Compute IObytes for baselines from given description:
+    Tensor_props has entries of the form: {'tensor_id': (size: tuple, dtype: torch.dtype)}
+    '''
+    iobytes = 0
+    for _, tensor_prop in tensor_props.items():
+        size, dtype = tensor_prop[0], tensor_prop[1]
+        if not isinstance(size, tuple):
+            size = (size)
+        iobytes += np.prod(size) * dtype.itemsize
+    return int(iobytes)
 
 
 class NVFBenchmark:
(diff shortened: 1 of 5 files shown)


liqiangxl (approved):
LGTM.

benchmarks/python/core.py
```diff
@@ -146,6 +147,21 @@ def unary_bwd_torch(inputs: List):  # [output, grad_out]
     inputs[0].backward(inputs[1], retain_graph=True)
 
 
+def compute_total_iobytes(tensor_props: dict):
```
jacobhinkle, line 150: Based on docstring this could be
```suggestion
def compute_total_iobytes(tensor_props: dict[str, tuple[tuple[int, ...], torch.dtype]]):
```

benchmarks/python/test_dropout_layernorm_bwd.py
```diff
@@ -4,7 +4,12 @@
 import pytest
 from nvfuser import FusionDefinition, DataType
 from nvfuser.pytorch_utils import torch_dtype_to_nvfuser_dtype
-from .core import run_benchmark, clear_cuda_cache
+from .core import (
+    run_benchmark,
+    clear_cuda_cache,
+    unary_bwd_torch,
+    compute_total_iobytes,
+)
```
jacobhinkle, line 12: You could just do `import core` then qualify the functions at the call sites. Not really necessary, but if the import list gets bigger you could consider switching.

benchmarks/python/test_dropout_rmsnorm_bwd.py
```diff
@@ -128,10 +158,40 @@ def test_rmsnorm_bwd_benchmark(
...
+):
+    clear_cuda_cache()
+    dropout_p = 0.2
+    input1 = torch.randn(size, device="cuda", dtype=dtype, requires_grad=True)
+    input2 = torch.randn(size, device="cuda", dtype=dtype, requires_grad=True)
+    grads = torch.randn(size, device="cuda", dtype=dtype)
+    weights = torch.randn(size[1], device="cuda", dtype=dtype)
+
+    x = input2 + torch.nn.functional.dropout(input1, p=dropout_p)
+    output = weights * torch.rsqrt(x.pow(2).mean(-1, keepdim=True) + 1e-5)
```
jacobhinkle, line 190: Since only `output` and `grads` are needed, why not just generate them with `torch.randn` instead of the actual forward?

jacobhinkle (approved):
Changes look good to me. Just some minor comments.

benchmarks/python/test_dropout_rmsnorm_bwd.py
Priya2698: I don't see how that would still have the same performance. The backward pass will look different in this case since there wouldn't be any actual computation graph, most likely producing different performance numbers.

jacobhinkle: Oh of course. I missed that this uses `unary_bwd_torch` instead of a direct implementation of the backward. Nevermind!

Issue #682, opened 2023-08-03
`FusionMatmulSchedulerEpilogueBias_CUDA` is failing with misaligned address on RTX4090


zasdfgbnm:
```
❯ ./build/nvfuser_tests --gtest_filter=*./build/nvfuser_tests --gtest_filter=*NVFuserTest.FusionMatmulSchedulerEpilogueBias_CUDA*
Note: Google Test filter = *NVFuserTest.FusionMatmulSchedulerEpilogueBias_CUDA*
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from NVFuserTest
[ RUN      ] NVFuserTest.FusionMatmulSchedulerEpilogueBias_CUDA
unknown file: Failure
C++ exception with description "CUDA error: misaligned address
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1.
Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.

Exception raised from c10_cuda_check_implementation at ../c10/cuda/CUDAException.cpp:44 (most recent call first):
frame #0: c10::Error::Error(c10::SourceLocation, std::string) + 0x57 (0x7f2f6f3a8477 in /home/gaoxiang/.local/lib/python3.11/site-packages/torch/lib/libc10.so)
frame #1: c10::detail::torchCheckFail(char const*, char const*, unsigned int, std::string const&) + 0x64 (0x7f2f6f36489b in /home/gaoxiang/.local/lib/python3.11/site-packages/torch/lib/libc10.so)
frame #2: c10::cuda::c10_cuda_check_implementation(int, char const*, char const*, int, bool) + 0x118 (0x7f2f6d538298 in /home/gaoxiang/.local/lib/python3.11/site-packages/torch/lib/libc10_cuda.so)
frame #3: void at::native::gpu_kernel_impl<at::native::BinaryFunctor<float, float, bool, at::native::(anonymous namespace)::CompareEqFunctor<float> > >(at::TensorIteratorBase&, at::native::BinaryFunctor<float, float, bool, at::native::(anonymous namespace)::CompareEqFunctor<float> > const&) + 0xc9f (0x7f2f0bc3187f in /home/gaoxiang/.local/lib/python3.11/site-packages/torch/lib/libtorch_cuda.so)
[...]

PR #2105, opened 2024-04-18
Support misaligned matmul using vectorization analysis

### LINKED ISSUE #983: Correctly set the vectorization size for matmul scheduler
Currently we are hardcoding 8. This does not always work depending on the alignments of addresses and strides. We should change that and set the correct vectorization size.

### LINKED ISSUE #2083: Matmul Schedules requires vectorization Analysis, it is currently hard coded
#682 Issue related to vectorization
@drzejan2  had a PR: https://github.com/NVIDIA/Fuser/pull/807

jacobhinkle:
This change collects the maximum supported vectorization size for A, B, and "epilogue" and attaches these to `MatmulParams`.  In case any "C" role tensors like bias are provided, they are considered when computing epilogue vectorization. This information becomes available to heuristic plugins also.

Fixes #2083. Fixes #983. Related to existing PR #807. (Related to issue #682 but I'm not sure yet if this approach will fix it).


tests/cpp/test_matmul_scheduler.cpp
```diff
@@ -2248,6 +2248,93 @@ TEST_F(MatmulSchedulerTest, StridedBatchEpilogueSingleBias) {
...
+          ") and layout inferred from fusion definition (",
+          toString(fusion_layout.getData()),
+          ")");
+
+      FusionExecutorCache executor_cache(std::move(fusion));
+
+      for (const auto& [M, N, K, alignA, alignB, alignBias] :
+           std::vector<std::tuple<int, int, int, int, int, int>>{
+               {504, 136, 248, 16, 16, 16},
+               {505, 131, 249, 16, 16, 16},
```
jacobhinkle, line 2310: This case currently fails. We select the heuristic properly (no plugin, just `initCoreHeuristics`) I think:
```

MMA macro: Ampere_16_16_16
DoubleBufferOptions:
  double_buffer_smem_write: true
  double_buffer_smem_read: true
  smem_double_buffer_stage: 2
SupportedVectorization:
  a: 1
  b: 1
  epilogue: 4
MatMulTileOptions: instruction tile [16, 16, 16], warp tile [64, 64, 32], CTA tile [256, 64, 32]
Async global mem load: false
Indexing mode: int32_t
Tile rastrization order: row-major
Grid swizzle factor: 1
Use shared memory epilogue: 0
Promote re-use of prologue shared memory: 1
Split-K factor: 1
```
However, execution fails with
>  what():  _result == CUDA_SUCCESS INTERNAL ASSERT FAILED at "/opt/pytorch/nvfuser/csrc/executor_utils.cpp":1152, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. CUDA error: CUDA_ERROR_MISALIGNED_ADDRESS failed with error misaligned address
Generated kernel:
```c++
__global__ void nvfuser_matmul_f0_c1_r0_g0(Tensor<__half, 2, 2> T0, Tensor<__half, 2, 2> T1, Tensor<float, 2, 2> T5) {
  alignas(16) extern __shared__ char array[];
  const unsigned smem_offset = 0;
  nvfuser_index_t i0;
  i0 = ceilDiv(T0.logical_size[1LL], 32);
  nvfuser_index_t i1;
  i1 = ((nvfuser_index_t)threadIdx.z) % 2;
  nvfuser_index_t i2;
  i2 = ((nvfuser_index_t)threadIdx.x) / 8;
  nvfuser_index_t i3;
  i3 = (i2 + (4 * ((nvfuser_index_t)threadIdx.y))) + (4 * i1);
  nvfuser_index_t i4;
  i4 = ((nvfuser_index_t)threadIdx.z) / 2;
  nvfuser_index_t i5;
  i5 = ((nvfuser_index_t)threadIdx.x) % 8;
  nvfuser_index_t i6;
  i6 = (i5 + (64LL * i4)) + (64LL * (i3 / 8));
  nvfuser_index_t i7;
  i7 = i3 % 8;
  nvfuser_index_t i8;
  i8 = 32 * ((nvfuser_index_t)threadIdx.y);
  nvfuser_index_t i9;
  i9 = 32 * i1;
  nvfuser_index_t i10;
  i10 = T1.logical_size[1LL] * i4;
  nvfuser_index_t i11;
  i11 = 64LL * ((nvfuser_index_t)blockIdx.y);
  nvfuser_index_t i12;
  i12 = (((nvfuser_index_t)threadIdx.x) + i8) + i9;
  nvfuser_index_t i13;
  i13 = (i12 + i10) + i11;
  nvfuser_index_t i14;
  i14 = 2 * T1.logical_size[1LL];
  nvfuser_index_t i15;
  i15 = 32LL * ((nvfuser_index_t)threadIdx.y);
  nvfuser_index_t i16;
  i16 = 32LL * ((nvfuser_index_t)threadIdx.z);
  nvfuser_index_t i17;
  i17 = (((32LL * (i2 / 4)) + i5) + i15) + i16;
  nvfuser_index_t i18;
  i18 = i2 % 4;
  nvfuser_index_t i19;
  i19 = T0.logical_size[1LL] * ((nvfuser_index_t)threadIdx.y);
  nvfuser_index_t i20;
  i20 = T0.logical_size[1LL] * ((nvfuser_index_t)threadIdx.z);
  nvfuser_index_t i21;
  i21 = (256LL * T0.logical_size[1LL]) * ((nvfuser_index_t)blockIdx.x);
  nvfuser_index_t i22;
  i22 = ((((nvfuser_index_t)threadIdx.x) + i19) + i20) + i21;
  nvfuser_index_t i23;
  i23 = 4 * T0.logical_size[1LL];
  nvfuser_index_t i24;
  i24 = i2 / 2;
  nvfuser_index_t i25;
  i25 = i2 % 2;
  __half* T7 = reinterpret_cast<__half*>(array + smem_offset + 32768LL);
  __half* T6 = reinterpret_cast<__half*>(array + smem_offset + 0);
  unsigned i26;
  i26 = (toSmem(T6) + (512 * i25)) + (64 * i5);
  nvfuser_index_t i27;
  i27 = (i5 / 2) % 4;
  nvfuser_index_t i28;
  i28 = 4096 * ((nvfuser_index_t)threadIdx.z);
  unsigned i29;
  i29 = ((i26 + (64 * (i24 / 4))) + (16 * (i27 ^ (i24 % 4)))) + i28;
  unsigned i30;
  i30 = toSmem(T7);
  nvfuser_index_t i31;
  i31 = 128 * i5;
  nvfuser_index_t i32;
  i32 = 1024 * i25;
  nvfuser_index_t i33;
  i33 = 128 * ((nvfuser_index_t)threadIdx.y);
  unsigned i34;
  i34 = ((i30 + i31) + i32) + i33;
  nvfuser_index_t i35;
  i35 = i5 % 8;
  nvfuser_index_t i36;
  i36 = (((((32 * T1.logical_size[1LL]) + ((nvfuser_index_t)threadIdx.x)) + i8) + i9) + i10) + i11;
  nvfuser_index_t i37;
  i37 = 32LL * T1.logical_size[1LL];
  nvfuser_index_t i38;
  i38 = 32 + ((nvfuser_index_t)threadIdx.x);
  nvfuser_index_t i39;
  i39 = ((i38 + i19) + i20) + i21;
  nvfuser_index_t i40;
  i40 = 2LL + i24;
  unsigned i41;
  i41 = ((i26 + (64 * (i40 / 4))) + (16 * (i27 ^ (i40 % 4)))) + i28;
  unsigned i42;
  i42 = (((2048 + i30) + i31) + i32) + i33;
  nvfuser_index_t i43;
  i43 = ((nvfuser_index_t)threadIdx.x) / 4;
  nvfuser_index_t i44;
  i44 = 16LL * T1.logical_size[1LL];
  nvfuser_index_t i45;
  i45 = 2LL * (((nvfuser_index_t)threadIdx.x) % 4);
  nvfuser_index_t i46;
  i46 = 64LL * ((nvfuser_index_t)threadIdx.y);
  nvfuser_index_t i47;
  i47 = i43 / 8;
  nvfuser_index_t i48;
  i48 = i43 % 8;
  nvfuser_index_t i49;
  i49 = (((((i45 + (i44 * i47)) + (T1.logical_size[1LL] * i48)) + i46) + ((64LL * T1.logical_size[1LL]) * ((nvfuser_index_t)threadIdx.z))) + i11) + ((256LL * T1.logical_size[1LL]) * ((nvfuser_index_t)blockIdx.x));
  nvfuser_index_t i50;
  i50 = 8LL * T1.logical_size[1LL];
  bool b51;
  b51 = (i12 + i11) < T1.logical_size[1LL];
  nvfuser_index_t i52;
  i52 = (((-(64 * T0.logical_size[1LL])) + ((nvfuser_index_t)threadIdx.x)) + i15) + i16;
  bool b53;
  b53 = ((nvfuser_index_t)threadIdx.x) < T0.logical_size[1LL];
  nvfuser_index_t i54;
  i54 = -T0.logical_size[0LL];
  nvfuser_index_t i55;
  i55 = 256LL * ((nvfuser_index_t)blockIdx.x);
  nvfuser_index_t i56;
  i56 = ((i54 + ((nvfuser_index_t)threadIdx.y)) + ((nvfuser_index_t)threadIdx.z)) + i55;
  nvfuser_index_t i57;
  i57 = (32 - T0.logical_size[1LL]) + i4;
  nvfuser_index_t i58;
  i58 = (((1 - T1.logical_size[1LL]) + i45) + i46) + i11;
  nvfuser_index_t i59;
  i59 = (((i54 + (16LL * i47)) + i48) + (64LL * ((nvfuser_index_t)threadIdx.z))) + i55;
  Array<float, 128, 2> T8;
  #pragma unroll
  for(nvfuser_index_t i60 = 0; i60 < 4; ++i60) {
    nvfuser_index_t i61;
    i61 = 32 * i60;
    #pragma unroll
    for(nvfuser_index_t i62 = 0; i62 < 4; ++i62) {
      ((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i61 + (8 * i62))]))).set(0);
    }
  }
  #pragma unroll
  for(nvfuser_index_t i63 = 0; i63 < 16; ++i63) {
    T7[((i6 + (128 * i63)) + (8LL * (i7 ^ ((i4 + (2 * i63)) % 8))))] = 0;
  }
  #pragma unroll
  for(nvfuser_index_t i63 = 0; i63 < 16; ++i63) {
    if ((b51 && (i52 < (-(128LL * i63))))) {
      T7[((i6 + (128 * i63)) + (8LL * (i7 ^ ((i4 + (2 * i63)) % 8))))]
         = T1[(i13 + (i14 * i63))];
    }
  }
  #pragma unroll
  for(nvfuser_index_t i64 = 0; i64 < 64; ++i64) {
    T6[((i17 + (128 * i64)) + (8LL * (i18 ^ ((i4 + (2 * i64)) % 4))))] = 0;
  }
  #pragma unroll
  for(nvfuser_index_t i64 = 0; i64 < 64; ++i64) {
    if ((b53 && (i56 < (-(4 * i64))))) {
      T6[((i17 + (128 * i64)) + (8LL * (i18 ^ ((i4 + (2 * i64)) % 4))))]
         = T0[(i22 + (i23 * i64))];
    }
  }
  __syncthreads();
  Array<__half, 64, 8> T9;
  Array<__half, 64, 8> T3;
  #pragma unroll
  for(nvfuser_index_t i65 = 0; i65 < 4; ++i65) {
    asm volatile(
      "ldmatrix.sync.aligned.x4.m8n8.shared.b16 {%0, %1, %2, %3}, [%4];\n"
      :"=r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T9[(8LL * i65)]))[0]),
       "=r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T9[(8LL * i65)]))[1]),
       "=r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T9[(8LL * i65)]))[2]),
       "=r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T9[(8LL * i65)]))[3])
      :"r"((uint32_t)((i29 + (1024 * i65))))
    );
  }
  #pragma unroll
  for(nvfuser_index_t i66 = 0; i66 < 4; ++i66) {
    nvfuser_index_t i67;
    i67 = i24 + (2LL * i66);
    asm volatile(
      "ldmatrix.sync.aligned.x4.trans.m8n8.shared.b16 {%0, %1, %2, %3}, [%4];\n"
      :"=r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T3[(8LL * i66)]))[0]),
       "=r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T3[(8LL * i66)]))[1]),
       "=r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T3[(8LL * i66)]))[2]),
       "=r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T3[(8LL * i66)]))[3])
      :"r"((uint32_t)(((i34 + (128 * (i67 / 8))) + (16 * (i35 ^ (i67 % 8))))))
    );
  }
  #pragma unroll 1
  for(nvfuser_index_t i68 = 0; i68 < i0; ++i68) {
    nvfuser_index_t i69;
    i69 = (1 + i68) % 2;
    nvfuser_index_t i70;
    i70 = i6 + (2048 * i69);
    nvfuser_index_t i71;
    i71 = i36 + (i37 * i68);
    nvfuser_index_t i72;
    i72 = i17 + (8192 * i69);
    nvfuser_index_t i73;
    i73 = 32LL * i68;
    nvfuser_index_t i74;
    i74 = i39 + i73;
    nvfuser_index_t i75;
    i75 = i68 % 2;
    unsigned i76;
    i76 = i41 + (16384 * i75);
    unsigned i77;
    i77 = i42 + (4096 * i75);
    unsigned i78;
    i78 = i29 + (16384 * i69);
    unsigned i79;
    i79 = i34 + (4096 * i69);
    nvfuser_index_t i80;
    i80 = i57 + i73;
    bool b81;
    b81 = (i38 + i73) < T0.logical_size[1LL];
    #pragma unroll
    for(nvfuser_index_t i63 = 0; i63 < 16; ++i63) {
      T7[((i70 + (128 * i63)) + (8LL * (i7 ^ ((i4 + (2 * i63)) % 8))))] = 0;
    }
    #pragma unroll
    for(nvfuser_index_t i63 = 0; i63 < 16; ++i63) {
      nvfuser_index_t i82;
      i82 = 2 * i63;
      if ((b51 && (i80 < (-i82)))) {
        T7[((i70 + (128 * i63)) + (8LL * (i7 ^ ((i4 + i82) % 8))))]
           = T1[(i71 + (i14 * i63))];
      }
    }
    #pragma unroll
    for(nvfuser_index_t i64 = 0; i64 < 64; ++i64) {
      T6[((i72 + (128 * i64)) + (8LL * (i18 ^ ((i4 + (2 * i64)) % 4))))] = 0;
    }
    #pragma unroll
    for(nvfuser_index_t i64 = 0; i64 < 64; ++i64) {
      if ((b81 && (i56 < (-(4 * i64))))) {
        T6[((i72 + (128 * i64)) + (8LL * (i18 ^ ((i4 + (2 * i64)) % 4))))]
           = T0[(i74 + (i23 * i64))];
      }
    }
    #pragma unroll
    for(nvfuser_index_t i65 = 0; i65 < 4; ++i65) {
      asm volatile(
        "ldmatrix.sync.aligned.x4.m8n8.shared.b16 {%0, %1, %2, %3}, [%4];\n"
        :"=r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T9[(32 + (8LL * i65))]))[0]),
         "=r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T9[(32 + (8LL * i65))]))[1]),
         "=r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T9[(32 + (8LL * i65))]))[2]),
         "=r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T9[(32 + (8LL * i65))]))[3])
        :"r"((uint32_t)((i76 + (1024 * i65))))
      );
    }
    __half T2[32];
    #pragma unroll
    for(nvfuser_index_t i83 = 0; i83 < 4; ++i83) {
      nvfuser_index_t i84;
      i84 = 8 * i83;
      #pragma unroll
      for(nvfuser_index_t i85 = 0; i85 < 2; ++i85) {
        nvfuser_index_t i86;
        i86 = i84 + (4LL * i85);
        #pragma unroll
        for(nvfuser_index_t i87 = 0; i87 < 2; ++i87) {
          nvfuser_index_t i88;
          i88 = i86 + (2LL * i87);
          #pragma unroll
          for(nvfuser_index_t i89 = 0; i89 < 2; ++i89) {
            T2[(i88 + i89)] = 0.000000000e+00f;
          }
        }
      }
    }
    #pragma unroll
    for(nvfuser_index_t i83 = 0; i83 < 4; ++i83) {
      nvfuser_index_t i90;
      i90 = 8 * i83;
      #pragma unroll
      for(nvfuser_index_t i85 = 0; i85 < 2; ++i85) {
        nvfuser_index_t i91;
        i91 = i90 + (4LL * i85);
        #pragma unroll
        for(nvfuser_index_t i87 = 0; i87 < 2; ++i87) {
          nvfuser_index_t i92;
          i92 = i91 + (2LL * i87);
          #pragma unroll
          for(nvfuser_index_t i89 = 0; i89 < 2; ++i89) {
            nvfuser_index_t i93;
            i93 = i92 + i89;
            T2[i93]
               = T9[i93];
          }
        }
      }
    }
    #pragma unroll
    for(nvfuser_index_t i66 = 0; i66 < 4; ++i66) {
      nvfuser_index_t i94;
      i94 = i24 + (2LL * i66);
      asm volatile(
        "ldmatrix.sync.aligned.x4.trans.m8n8.shared.b16 {%0, %1, %2, %3}, [%4];\n"
        :"=r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T3[(32 + (8LL * i66))]))[0]),
         "=r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T3[(32 + (8LL * i66))]))[1]),
         "=r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T3[(32 + (8LL * i66))]))[2]),
         "=r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T3[(32 + (8LL * i66))]))[3])
        :"r"((uint32_t)(((i77 + (128 * (i94 / 8))) + (16 * (i35 ^ (i94 % 8))))))
      );
    }
    __half T4[32];
    #pragma unroll
    for(nvfuser_index_t i95 = 0; i95 < 4; ++i95) {
      nvfuser_index_t i96;
      i96 = 8 * i95;
      #pragma unroll
      for(nvfuser_index_t i97 = 0; i97 < 2; ++i97) {
        nvfuser_index_t i98;
        i98 = i96 + (4 * i97);
        #pragma unroll
        for(nvfuser_index_t i99 = 0; i99 < 2; ++i99) {
          nvfuser_index_t i100;
          i100 = i98 + (2LL * i99);
          #pragma unroll
          for(nvfuser_index_t i101 = 0; i101 < 2; ++i101) {
            T4[(i100 + i101)] = 0.000000000e+00f;
          }
        }
      }
    }
    #pragma unroll
    for(nvfuser_index_t i95 = 0; i95 < 4; ++i95) {
      nvfuser_index_t i102;
      i102 = 8 * i95;
      #pragma unroll
      for(nvfuser_index_t i97 = 0; i97 < 2; ++i97) {
        nvfuser_index_t i103;
        i103 = i102 + (4 * i97);
        #pragma unroll
        for(nvfuser_index_t i99 = 0; i99 < 2; ++i99) {
          nvfuser_index_t i104;
          i104 = i103 + (2LL * i99);
          #pragma unroll
          for(nvfuser_index_t i101 = 0; i101 < 2; ++i101) {
            nvfuser_index_t i105;
            i105 = i104 + i101;
            T4[i105]
               = T3[i105];
          }
        }
      }
    }
    #pragma unroll
    for(nvfuser_index_t i60 = 0; i60 < 4; ++i60) {
      nvfuser_index_t i106;
      i106 = 8 * i60;
      nvfuser_index_t i107;
      i107 = 32 * i60;
      #pragma unroll
      for(nvfuser_index_t i62 = 0; i62 < 4; ++i62) {
        nvfuser_index_t i108;
        i108 = 8 * i62;
        asm(
          "mma.sync.aligned.m16n8k16.row.col.f32.f16.f16.f32 {%0, %1, %2, %3}, {%4, %5, %6, %7}, {%8, %9}, {%10, %11, %12, %13};\n"
          :"=f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i107 + i108)])))))[0LL][0]),
           "=f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i107 + i108)])))))[0LL][1]),
           "=f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i107 + i108)])))))[0LL][2]),
           "=f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i107 + i108)])))))[0LL][3])
          :"r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T2[i106]))[0]),
           "r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T2[i106]))[1]),
           "r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T2[i106]))[2]),
           "r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T2[i106]))[3]),
           "r"((*reinterpret_cast<Array<Array<uint32_t, 2, 1>, 2, 1>*>(&((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T4[i108])))))[0LL][0]),
           "r"((*reinterpret_cast<Array<Array<uint32_t, 2, 1>, 2, 1>*>(&((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T4[i108])))))[0LL][1]),
           "f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i107 + i108)])))))[0LL][0]),
           "f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i107 + i108)])))))[0LL][1]),
           "f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i107 + i108)])))))[0LL][2]),
           "f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i107 + i108)])))))[0LL][3])
        );
        asm(
          "mma.sync.aligned.m16n8k16.row.col.f32.f16.f16.f32 {%0, %1, %2, %3}, {%4, %5, %6, %7}, {%8, %9}, {%10, %11, %12, %13};\n"
          :"=f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i107 + i108)])))))[1LL][0]),
           "=f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i107 + i108)])))))[1LL][1]),
           "=f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i107 + i108)])))))[1LL][2]),
           "=f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i107 + i108)])))))[1LL][3])
          :"r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T2[i106]))[0]),
           "r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T2[i106]))[1]),
           "r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T2[i106]))[2]),
           "r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T2[i106]))[3]),
           "r"((*reinterpret_cast<Array<Array<uint32_t, 2, 1>, 2, 1>*>(&((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T4[i108])))))[1LL][0]),
           "r"((*reinterpret_cast<Array<Array<uint32_t, 2, 1>, 2, 1>*>(&((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T4[i108])))))[1LL][1]),
           "f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i107 + i108)])))))[1LL][0]),
           "f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i107 + i108)])))))[1LL][1]),
           "f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i107 + i108)])))))[1LL][2]),
           "f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i107 + i108)])))))[1LL][3])
        );
      }
    }
    #pragma unroll
    for(nvfuser_index_t i83 = 0; i83 < 4; ++i83) {
      nvfuser_index_t i109;
      i109 = 8 * i83;
      #pragma unroll
      for(nvfuser_index_t i85 = 0; i85 < 2; ++i85) {
        nvfuser_index_t i110;
        i110 = i109 + (4LL * i85);
        #pragma unroll
        for(nvfuser_index_t i87 = 0; i87 < 2; ++i87) {
          nvfuser_index_t i111;
          i111 = i110 + (2LL * i87);
          #pragma unroll
          for(nvfuser_index_t i89 = 0; i89 < 2; ++i89) {
            T2[(i111 + i89)] = 0.000000000e+00f;
          }
        }
      }
    }
    #pragma unroll
    for(nvfuser_index_t i83 = 0; i83 < 4; ++i83) {
      nvfuser_index_t i112;
      i112 = 8 * i83;
      nvfuser_index_t i113;
      i113 = 32 + i112;
      #pragma unroll
      for(nvfuser_index_t i85 = 0; i85 < 2; ++i85) {
        nvfuser_index_t i114;
        i114 = 4LL * i85;
        nvfuser_index_t i115;
        i115 = i112 + i114;
        nvfuser_index_t i116;
        i116 = i113 + i114;
        #pragma unroll
        for(nvfuser_index_t i87 = 0; i87 < 2; ++i87) {
          nvfuser_index_t i117;
          i117 = 2LL * i87;
          nvfuser_index_t i118;
          i118 = i115 + i117;
          nvfuser_index_t i119;
          i119 = i116 + i117;
          #pragma unroll
          for(nvfuser_index_t i89 = 0; i89 < 2; ++i89) {
            T2[(i118 + i89)]
               = T9[(i119 + i89)];
          }
        }
      }
    }
    #pragma unroll
    for(nvfuser_index_t i95 = 0; i95 < 4; ++i95) {
      nvfuser_index_t i120;
      i120 = 8 * i95;
      #pragma unroll
      for(nvfuser_index_t i97 = 0; i97 < 2; ++i97) {
        nvfuser_index_t i121;
        i121 = i120 + (4 * i97);
        #pragma unroll
        for(nvfuser_index_t i99 = 0; i99 < 2; ++i99) {
          nvfuser_index_t i122;
          i122 = i121 + (2LL * i99);
          #pragma unroll
          for(nvfuser_index_t i101 = 0; i101 < 2; ++i101) {
            T4[(i122 + i101)] = 0.000000000e+00f;
          }
        }
      }
    }
    #pragma unroll
    for(nvfuser_index_t i95 = 0; i95 < 4; ++i95) {
      nvfuser_index_t i123;
      i123 = 8 * i95;
      nvfuser_index_t i124;
      i124 = 32 + i123;
      #pragma unroll
      for(nvfuser_index_t i97 = 0; i97 < 2; ++i97) {
        nvfuser_index_t i125;
        i125 = 4 * i97;
        nvfuser_index_t i126;
        i126 = i123 + i125;
        nvfuser_index_t i127;
        i127 = i124 + i125;
        #pragma unroll
        for(nvfuser_index_t i99 = 0; i99 < 2; ++i99) {
          nvfuser_index_t i128;
          i128 = 2LL * i99;
          nvfuser_index_t i129;
          i129 = i126 + i128;
          nvfuser_index_t i130;
          i130 = i127 + i128;
          #pragma unroll
          for(nvfuser_index_t i101 = 0; i101 < 2; ++i101) {
            T4[(i129 + i101)]
               = T3[(i130 + i101)];
          }
        }
      }
    }
    #pragma unroll
    for(nvfuser_index_t i60 = 0; i60 < 4; ++i60) {
      nvfuser_index_t i131;
      i131 = 8 * i60;
      nvfuser_index_t i132;
      i132 = 32 * i60;
      #pragma unroll
      for(nvfuser_index_t i62 = 0; i62 < 4; ++i62) {
        nvfuser_index_t i133;
        i133 = 8 * i62;
        asm(
          "mma.sync.aligned.m16n8k16.row.col.f32.f16.f16.f32 {%0, %1, %2, %3}, {%4, %5, %6, %7}, {%8, %9}, {%10, %11, %12, %13};\n"
          :"=f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i132 + i133)])))))[0LL][0]),
           "=f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i132 + i133)])))))[0LL][1]),
           "=f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i132 + i133)])))))[0LL][2]),
           "=f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i132 + i133)])))))[0LL][3])
          :"r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T2[i131]))[0]),
           "r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T2[i131]))[1]),
           "r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T2[i131]))[2]),
           "r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T2[i131]))[3]),
           "r"((*reinterpret_cast<Array<Array<uint32_t, 2, 1>, 2, 1>*>(&((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T4[i133])))))[0LL][0]),
           "r"((*reinterpret_cast<Array<Array<uint32_t, 2, 1>, 2, 1>*>(&((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T4[i133])))))[0LL][1]),
           "f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i132 + i133)])))))[0LL][0]),
           "f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i132 + i133)])))))[0LL][1]),
           "f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i132 + i133)])))))[0LL][2]),
           "f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i132 + i133)])))))[0LL][3])
        );
        asm(
          "mma.sync.aligned.m16n8k16.row.col.f32.f16.f16.f32 {%0, %1, %2, %3}, {%4, %5, %6, %7}, {%8, %9}, {%10, %11, %12, %13};\n"
          :"=f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i132 + i133)])))))[1LL][0]),
           "=f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i132 + i133)])))))[1LL][1]),
           "=f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i132 + i133)])))))[1LL][2]),
           "=f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i132 + i133)])))))[1LL][3])
          :"r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T2[i131]))[0]),
           "r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T2[i131]))[1]),
           "r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T2[i131]))[2]),
           "r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T2[i131]))[3]),
           "r"((*reinterpret_cast<Array<Array<uint32_t, 2, 1>, 2, 1>*>(&((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T4[i133])))))[1LL][0]),
           "r"((*reinterpret_cast<Array<Array<uint32_t, 2, 1>, 2, 1>*>(&((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T4[i133])))))[1LL][1]),
           "f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i132 + i133)])))))[1LL][0]),
           "f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i132 + i133)])))))[1LL][1]),
           "f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i132 + i133)])))))[1LL][2]),
           "f"((*reinterpret_cast<Array<Array<float, 4, 1>, 2, 1>*>(&((*reinterpret_cast<Array<float, 8, 1>*>(&T8[(i132 + i133)])))))[1LL][3])
        );
      }
    }
    __syncthreads();
    #pragma unroll
    for(nvfuser_index_t i65 = 0; i65 < 4; ++i65) {
      asm volatile(
        "ldmatrix.sync.aligned.x4.m8n8.shared.b16 {%0, %1, %2, %3}, [%4];\n"
        :"=r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T9[(8LL * i65)]))[0]),
         "=r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T9[(8LL * i65)]))[1]),
         "=r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T9[(8LL * i65)]))[2]),
         "=r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T9[(8LL * i65)]))[3])
        :"r"((uint32_t)((i78 + (1024 * i65))))
      );
    }
    #pragma unroll
    for(nvfuser_index_t i66 = 0; i66 < 4; ++i66) {
      nvfuser_index_t i134;
      i134 = i24 + (2LL * i66);
      asm volatile(
        "ldmatrix.sync.aligned.x4.trans.m8n8.shared.b16 {%0, %1, %2, %3}, [%4];\n"
        :"=r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T3[(8LL * i66)]))[0]),
         "=r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T3[(8LL * i66)]))[1]),
         "=r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T3[(8LL * i66)]))[2]),
         "=r"((*reinterpret_cast<Array<uint32_t, 4, 1>*>(&T3[(8LL * i66)]))[3])
        :"r"((uint32_t)(((i79 + (128 * (i134 / 8))) + (16 * (i35 ^ (i134 % 8))))))
      );
    }
  }
  #pragma unroll
  for(nvfuser_index_t i135 = 0; i135 < 4; ++i135) {
    nvfuser_index_t i136;
    i136 = 32 * i135;
    nvfuser_index_t i137;
    i137 = i49 + (i44 * i135);
    nvfuser_index_t i138;
    i138 = -(16LL * i135);
    #pragma unroll
    for(nvfuser_index_t i139 = 0; i139 < 4; ++i139) {
      nvfuser_index_t i140;
      i140 = i136 + (8 * i139);
      nvfuser_index_t i141;
      i141 = 16LL * i139;
      nvfuser_index_t i142;
      i142 = i137 + i141;
      nvfuser_index_t i143;
      i143 = -i141;
      #pragma unroll
      for(nvfuser_index_t i144 = 0; i144 < 2; ++i144) {
        nvfuser_index_t i145;
        i145 = i140 + (4LL * i144);
        nvfuser_index_t i146;
        i146 = 8LL * i144;
        nvfuser_index_t i147;
        i147 = i142 + i146;
        bool b148;
        b148 = i58 < (i143 - i146);
        #pragma unroll
        for(nvfuser_index_t i149 = 0; i149 < 2; ++i149) {
          if ((b148 && (i59 < (i138 - (8LL * i149))))) {
            loadLocalToGlobal<float, /*vec_size=*/2, /*is_volatile=*/false>( &T5[(i147 + (i50 * i149))], &T8[(i145 + (2LL * i149))]);
          }
        }
      }
    }
  }
}
```

jacobhinkle: Oh ok, the problem is in the output. Currently we are fully vectorizing the output, but that is not always possible depending on M and N.

tests/cpp/test_matmul_scheduler.cpp
```diff
@@ -2248,6 +2248,104 @@ TEST_F(MatmulSchedulerTest, StridedBatchEpilogueSingleBias) {
...
+          toString(fusion_layout.getData()),
+          ")");
+
+      FusionExecutorCache executor_cache(std::move(fusion));
+
+      for (const auto& [M, N, K, alignA, alignB, alignBias] :
+           std::vector<std::tuple<int, int, int, int, int, int>>{
+               {504, 136, 248, 16, 16, 16}, // all fully vectorizable in all
+                                            // layouts
+               {504, 136, 249, 16, 16, 16}, // odd K, operands not vectorizable
```
jacobhinkle, line 2311: This case has odd K, which is the inner dimension for A in the TT and TN cases, and is the inner dimension for B in the NN and TN cases. So I expect either A or B to have a maximum supported vectorization of 1 and the output to have max vec of 4 for HSS and 8 for HSH. I am seeing 8 8 4 instead, in the TT case, so it seems there is more work to do on the analysis side.

jacobhinkle: NN should be transposing `B`, but it seems `vectorize_helper` is not aware of that. Also, it doesn't assume we'll be doing vectorized loads starting at each row, instead of merging all contiguous dims then looping over that 1D dim. I think it's best to directly evaluate the stride of the M, N, and K dimensions in each operand (as well as the tensor pointer alignment and any outer strides as well).

jacobhinkle: I think this means that we can look at each operand, find the dimension with the smallest stride, assert that that stride is 1, then find the largest power of two that's at most 16 and which divides evenly into all other strides. If we know the tensor is contiguous then I guess we could just do this for the smallest stride which is not 1 since all others are a multiple of it.

jacobhinkle: Fixed in latest push. I no longer use `vectorize_helper` at all.

jacobhinkle:
This seems to be working, but I haven't updated the tests to include misaligned `data_ptr` in the operands, and the test does not include bias yet. However, for odd sized inputs in any layout, with and without downcasting the output, the kernels run.

I wonder if we should print a warning when we detect misaligned matmuls, since this is possibly a large perf penalty. We could have a disable option for it, maybe with args so that we could name specific perf warnings we want to suppress...

tests/cpp/test_matmul_scheduler.cpp
```diff
@@ -2248,6 +2248,107 @@ TEST_F(MatmulSchedulerTest, StridedBatchEpilogueSingleBias) {
...
+                                            // in TN. output fully vectorizable
+               {504, 137, 248, 16, 16, 16}, // A fully vectorizable, B fully
```
jacobhinkle, line 2313: Most of the tests are passing but this one is failing in NT layout, with `downcast_output=false`. It makes an unaligned read of A somehow.
```
                                                                                                                           
MMA macro: Ampere_16_16_16                                                                                                 
DoubleBufferOptions:                                                                                                       
  double_buffer_smem_write: true                                                                                           
  double_buffer_smem_read: true                                                                                            
  smem_double_buffer_stage: 2                                                                                              
SupportedVectorization:                                                                                                    
  a: 8                                                                                                                     
  b: 1                                                                                                                     
  epilogue: 1                                                                                                              
MatMulTileOptions: instruction tile [16, 16, 16], warp tile [64, 64, 32], CTA tile [256, 64, 32]                           
Async global mem load: false                                                                                               
Indexing mode: int32_t                                                                                                     
Tile rastrization order: row-major                                                                                         
Grid swizzle factor: 1                                                                                                     
Use shared memory epilogue: 0                                                                                              
Promote re-use of prologue shared memory: 1                                                                                
Split-K factor: 1                                                                                                          
Arguments for kernel0:                                                                                                     
Inputs:                                                                                                                    
  Tensor(sizes=[248, 504], stride=[504, 1], dtype=c10::Half, device=cuda:0, data_ptr=0x7eb2f7a00000)                       
  Tensor(sizes=[248, 137], stride=[137, 1], dtype=c10::Half, device=cuda:0, data_ptr=0x7eb2f7a3d200)                       
Outputs:                                                                                                                   
  Float [504, 137] (strides = [137, 1], address = 0x7eb2f7a4dc00)                                                          
Intermediate global buffers:  
```
```c++
  #pragma unroll
  for(nvfuser_index_t i66 = 0; i66 < 8; ++i66) {
    arraySet<__half, 8>(&T7[((i17 + (1024 * i66)) + (8LL * (i5 ^ (i18 + (4 * (i66 % 2))))))], (__half)0);
  }
  #pragma unroll
  for(nvfuser_index_t i66 = 0; i66 < 8; ++i66) {
    if ((b54 && (i55 < (-(1024LL * i66))))) {
      loadGeneric<__half, 8>( &T7[((i17 + (1024 * i66)) + (8LL * (i5 ^ (i18 + (4 * (i66 % 2))))))],  &T0[(i23 + (i24 * i66))]);
    }
  }
```

jacobhinkle: `T0.logical_size[1LL]` is 504, which is divisible by 8. And the `data_ptr` for T0 is 0x7eb2f7a00000. The error is
> ========= Invalid __global__ read of size 16 bytes
=========     at void <unnamed>::loadGeneric<<unnamed>::__half, (int)8>(T1 *, T1 *)+0x1730 in /opt/pytorch/nvfuser/__tmp_kernel_2.cu:2383
=========     by thread (1,0,1) in block (0,0,0)
=========     Address 0x72c4d3a00402 is misaligned
=========     and is inside the nearest allocation at 0x72c4d3a00000 of size 2097152 bytes
=========     Device Frame:<unnamed>::nvfuser_2(<unnamed>::Tensor<<unnamed>::__half, (int)2, (int)2>, <unnamed>::Tensor<<unnamed>::__half, (int)2, (int)2>, <unnamed>::Tensor<float, (int)2, (int)2>)+0x1730 in /opt/pytorch/nvfuser/__tmp_kernel_2.cu:10175

jacobhinkle: The offending address 0x7eb2f7a013ba is aligned at 2 bytes (`bin(0x7eb2f7a013ba) == '0b11111101011001011110111101000000001001110111010'`). I'm trying to figure out how this can happen.

jacobhinkle: Thread (0, 0, 1) in block (0, 0, 0) should be trying to access these three elements:
- `i3 = 0` : `T0.logical_size[1LL]` = 504
- `i3 = 1` : `T0.logical_size[1LL] + 4 * T0.logical_size[1LL]` = 5 * 504 = 2520
- `i3 = 2` : 9 * 504 = 4536
Yet the offending offset is at 0x7eb2f7a013ba = 0x7eb2f7a00000 + 5050 bytes. 5050 bytes = 2525 `__half` elements. so it appears there's some problem at `i3=1`?

jacobhinkle: Also suspicious: this doesn't happen if I isolate the test. That is, if I comment out the other problem sizes here, it goes away. I can set the layout to only `NT` and preserve the error..

jacobhinkle: OK I believe this is due to kernel re-use in `FusionExecutorCache`.

jacobhinkle: I was just missing a comparison for the new attribute `supported_vec_size` in `MatmulParams::sameAs`.

csrc/scheduler/utils.h
```diff
@@ -42,12 +42,24 @@ constexpr int64_t y_grid_limit = 65535;
...
-  return next_vector_size >> 1;
+  // For example
+  //   n               = b101101000
+  //           n - 1   = b101100111
+  //        ~ (n - 1)  = b010011000
+  //   n & (~ (n - 1)) = b000001000
+  // The key is that subtracting one flips all trailing 0s as well as the least
+  // significant 1, so all of the other bits will fail the &, leaving
+  // only that 1.
+  return n & (~(n - 1));
```
jacobhinkle, line 62: It's not strictly necessary to do this cleanup, but the original code was returning 0 when given a size equal to zero. When using this on a relative address, that can be problematic as it was causing us to set supported vectorization for some operands to 0.

csrc/scheduler/matmul_utils.cpp
```diff
@@ -294,6 +315,98 @@ std::string isMatmulFusionDefinitionSupported(
...
+//
+// If the argument has no contiguous dimensions, then a vectorization width of 1
+// is returned.
+//
+// These rows can start at any multiple of the non-contiguous strides, so we
+// seek the largest power of 2 that divides all those other dimensions (capped
+// to 16) as well as the data pointer.
+int64_t maxRowVectorization(const at::Tensor& tens) {
+  const int64_t data_ptr_int =
+      static_cast<int64_t>(reinterpret_cast<std::uintptr_t>(tens.data_ptr()));
```
jacobhinkle, line 330: For some reason `tens.data_ptr()` is always 0 currently. I am looking into this since it means we can't detect misaligned inputs currently (see failing test)

jacobhinkle: Actually that will require more involved changes. The issue is here: https://github.com/NVIDIA/Fuser/blob/2f8961a983dc1e136958f94a7658912a465cf57f/csrc/kernel_cache.cpp#L1613-L1614
That calls this piece of code which ensures that we do not hold tensors indefinitely that would reserve their data waste memory:https://github.com/NVIDIA/Fuser/blob/2f8961a983dc1e136958f94a7658912a465cf57f/csrc/kernel_cache.cpp#L34-L70
However, it also means the heuristic does not have information about pointer alignment, even though it is available prior to this. It is actually acknowledged as well since pointer alignment is used to create our fast path's cache key: https://github.com/NVIDIA/Fuser/blob/2f8961a983dc1e136958f94a7658912a465cf57f/csrc/kernel_cache.cpp#L314-L316
Changing that is probably beyond the scope of this PR, but it should be noted that inputs that are misaligned due to slicing might cause errors until we fix this.

jacobhinkle: Until then I will disable the failing tests and post an issue to track this.

jacobhinkle:
!build --diff

csrc/scheduler/matmul_utils.cpp
```diff
@@ -294,6 +315,98 @@ std::string isMatmulFusionDefinitionSupported(
...
+      break;
+    }
+    vec_size =
+        std::min(vec_size, scheduler_utils::maxVectorizationWidth(stride_i));
+  }
+
+  return vec_size;
+}
+
+MatmulParams::SupportedVectorization getSupportedVectorization(
```
zasdfgbnm, line 358: Could you split this PR into two?
1. Adding SupportedVectorization to heuristics and modify matmul scheduler to respect this field in heuristics
2. Computing SupportedVectorization

tests/cpp/test_matmul_scheduler.cpp
```diff
@@ -2248,6 +2248,126 @@ TEST_F(MatmulSchedulerTest, StridedBatchEpilogueSingleBias) {
...
+                                           // output fully vectorizable
+                 {505, 137, 248, 8, 8, 8}, // none vectorizable
+
+                 // The following cases are failing due to
+                 // https://github.com/NVIDIA/Fuser/issues/2118
+                 // Cases with vectorizable strides but misaligned base pointers
+                 //{504, 136, 248, 2, 8, 8}, // A not vectorizable due to offset
+                 //{504, 136, 248, 8, 2, 8}, // B not vectorizable due to offset
+                 //{504, 136, 248, 8, 8, 2}, // epilogue not vectorizable due to
+                 // offset
```
jacobhinkle, line 2317: Given that none of our other schedulers seem to support misaligned inputs (in the sense of having a base pointer not aligned to 16 bytes), that we should leave this for another PR.

jacobhinkle: Note however that I did try replacing this https://github.com/NVIDIA/Fuser/blob/43370c6564c74995332f2642497563172fb3f2c8/csrc/kernel_cache.cpp#L1613-L1614
with 
```c++
 KernelArgumentHolder args_metadata = args; 
```
and these cases succeed. So I think once #2118 is addressed we can re-enable these cases.

jacobhinkle: Waiting for #2123 to enable these.

jacobhinkle:
!build

jacobhinkle:
!build

zasdfgbnm:
cc: @drzejan2

tests/cpp/test_matmul_scheduler.cpp
```diff
@@ -2248,6 +2248,126 @@ TEST_F(MatmulSchedulerTest, StridedBatchEpilogueSingleBias) {
...
+                                           // vectorizable unless transposed,
+                                           // output not vectorizable
+                 {505, 136, 248, 8, 8, 8}, // B fully vectorizable, A
+                                           // vectorizable unless transposed,
+                                           // output fully vectorizable
+                 {505, 137, 248, 8, 8, 8}, // none vectorizable
+
+                 // The following cases are failing due to
+                 // https://github.com/NVIDIA/Fuser/issues/2118
+                 // Cases with vectorizable strides but misaligned base pointers
```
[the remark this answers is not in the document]
jacobhinkle: Good catch. Removed comment.

csrc/scheduler/matmul_utils.cpp
```diff
@@ -294,6 +315,98 @@ std::string isMatmulFusionDefinitionSupported(
...
+      static_cast<int64_t>(reinterpret_cast<std::uintptr_t>(tens.data_ptr()));
+  int64_t vec_size = scheduler_utils::maxVectorizationWidth(data_ptr_int);
+  vec_size = std::min(vec_size, (int64_t)(16l));
+  vec_size /= (int64_t)tens.element_size();
+  vec_size = std::max(vec_size, 1l);
+  if (vec_size == 1l) {
+    return vec_size;
+  }
+
+  std::vector<int64_t> strides = tens.strides().vec();
```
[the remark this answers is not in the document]
jacobhinkle: it will. For an ordinary loop we might vectorize but peel off the last iteration and predicate. However we ar enot doing that for matmul, so we should return 1 in that case correct? If so then we should pass the inner dimension in addition to the non-unit strides to `scheduler_utils::maxVectorizationWidth()`.

csrc/scheduler/matmul_utils.cpp
```diff
@@ -294,6 +315,98 @@ std::string isMatmulFusionDefinitionSupported(
...
+    }
+    return vec_size;
+  };
+  MatmulParams::SupportedVectorization supported_vec_size;
+  supported_vec_size.a = getMinVectorization(MatmulRole::INPUT_A);
+  supported_vec_size.b = getMinVectorization(MatmulRole::INPUT_B);
+  // If there is no C tensor, then we will use full vectorization of the
+  // outputs. So find the max vectorization of the output tensors, then
+  // possibly reduce it if it is not supported by a input C tensor.
+  supported_vec_size.epilogue = getMinVectorization(MatmulRole::INPUT_C);
```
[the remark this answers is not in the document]
jacobhinkle: This is a great point. We should support as general an epilogue as possible and that includes transposes. I had implicitly assumed we would schedule each aux input like the epilogue since that's what we currently do:
https://github.com/NVIDIA/Fuser/blob/eb75ab282dfe285cb8db7bcd5f8470f07ec889bb/csrc/scheduler/matmul.cpp#L647-L650
Perhaps we could use an issue to track at least that case of a transposed 2D bias. I could imagine eventually supporting normalizations and reductions in the epilogue as well, for example a loss function in an MLP.

Regarding the options you presented:

1. I believe so far the epilogue is ignored entirely in `canScheduleCompileTime`: https://github.com/NVIDIA/Fuser/blob/eb75ab282dfe285cb8db7bcd5f8470f07ec889bb/csrc/scheduler/matmul_utils.cpp#L313
All that is checked is that there is a mul-sum pattern or `MmaOp`. That function guards which fusions we will optionally accept as nvfuser matmuls once we have a new ATen matmul scheduler (soon). Since we know there are cases that are likely to fail (transposes, reductions, normalizations) we should at least filter those out. I think it makes sense to lock this down to only pointwise ops with inputs that have the same stride order as the output. The stride order requirement could be lifted in a separate PR.
2. We should definitely have a complete vectorization analysis accepting a broader collection of epilogues. I will have a look at #807 again now that I'm more familiar with this problem.

I see now that epilogue vectorization is not so simple. As you say, it does make sense to move it to a separate PR. I will strip the current PR down to just inferring A and B vectorization, leaving epilogue vectorization set to the maximum supported by the dtype. I'll file an issue to track epilogue analysis.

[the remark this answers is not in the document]
jacobhinkle: Opened issues #2167 and #2169

csrc/scheduler/matmul_utils.cpp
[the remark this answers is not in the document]
jacobhinkle: I added a test that checks that both stride and determine vectorizability.

tests/cpp/test_matmul_scheduler.cpp
```diff
@@ -2238,16 +2234,363 @@ TEST_F(MatmulSchedulerTest, StridedBatchEpilogueSingleBias) {
...
+          auto tv0 = TensorViewBuilder()
+                         .ndims(2)
+                         .contiguity({false, true})
+                         .dtype(DataType::Half)
+                         .build();
+          auto tv1 = TensorViewBuilder()
+                         .ndims(2)
+                         .contiguity({false, true})
+                         .dtype(DataType::Half)
+                         .build();
```
jacobhinkle, line 2440: Currently, our matmul scheduler will accept a matmul segment whose inputs are marked discontigous. Then if contiguous inputs are given, we set the vectorization factor and wind up vectorizing. This leads to a validation error in the fusion since the `TensorView` was not marked contiguous. It is rare that this would happen in practice, but we could check that the fusion's input TensorViews have contiguous inner dimensions when inferring vectorization for new inputs.

csrc/scheduler/matmul_heuristic.h
```diff
@@ -67,13 +67,13 @@ class MatmulParams : public HeuristicParams {
   //! This is the maximum vectorization supported by the inputs.
   struct SupportedVectorization {
     // operands
-    int64_t a = 8;
-    int64_t b = 8;
+    int64_t a = 16;
```
zasdfgbnm, line 70: Is this number of items or number of bytes?

tests/cpp/test_matmul_scheduler.cpp
jacobhinkle: I just pushed a change that does this and tests it. I believe all inputs are properly supported now, keeping in mind the restriction that epilogue vectorization is not yet handled correctly (#2169).

csrc/scheduler/matmul_utils.cpp
```diff
@@ -353,6 +478,24 @@ std::string getMatmulCompileTimeRejectReason(Fusion* fusion) {
...
+  }
+  // Use cp.async for loading operands if vec size is compatible
+  const auto& validCpAsyncVecSize = [](int64_t dtype_size,
+                                       int64_t vec_size) -> bool {
+    int64_t cp_bytes = dtype_size * vec_size;
+    return cp_bytes == 16 || cp_bytes == 8 || cp_bytes == 4;
+  };
+  return params->double_buffer_options.smem_double_buffer_stage > 1 &&
+      validCpAsyncVecSize(dtype_size_a, params->supported_vec_size.a) &&
+      validCpAsyncVecSize(dtype_size_b, params->supported_vec_size.b);
```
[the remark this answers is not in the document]
jacobhinkle: I think so. In most cases these will be the same but I don't think there's any harm in making them independent.

csrc/scheduler/matmul_utils.cpp
```diff
@@ -294,6 +315,110 @@ std::string isMatmulFusionDefinitionSupported(
...
+  }
+
+  // Find innermost dimension
+  bool contiguous = false;
+  for (int64_t i : c10::irange(tens.ndimension())) {
+    int64_t stride = tens.stride(i);
+    int64_t size = tens.size(i);
+    if (stride == 1) {
+      // Innermost dimension
+      contiguous = true;
```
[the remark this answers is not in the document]
jacobhinkle: Discontiguous (in the Fusion) tensors are supported properly now as of the latest push.
 
The overlapping case is more interesting. For the case of multiple dimensions with stride 1 and corresponding sizes that are _not_ 1, I think the vectorization should always be 1 since there is definitely a row that's equal to another offset by 1.  However, if the outer size is also 1 in that case then we could possibly still vectorize. Since it is somewhat difficult to determine which row will be vectorized at this point in the code, I think we should play it safe and set vectorization to 1 if there are multiple stride=1 dims.

csrc/scheduler/matmul_heuristic.h
jacobhinkle: Number of items. I'll make that clear in a comment. I set it to default to the largest possible vectorization for convenience so that we can just use `std::min` to update it.

zasdfgbnm: Isn't the largest possible vectorization 8 items?
https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#data-movement-and-conversion-instructions-cp-async

zasdfgbnm: Oh, you are saying "the largest possible vectorization for all possible dtypes"?

csrc/scheduler/matmul_utils.cpp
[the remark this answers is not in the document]
jacobhinkle: Would it make sense to have a different number of load stages for the two operands? If A cannot be asynchronously loaded due to too small vec size then it might only be able to double buffer, while B could still be circular buffered. I will try this out to see if the resulting kernel runs.

csrc/scheduler/matmul_heuristic.h
jacobhinkle: Yeah exactly. We could potentially have an FP8 operand at some point.

csrc/scheduler/matmul_utils.cpp
```diff
@@ -294,6 +315,118 @@ std::string isMatmulFusionDefinitionSupported(
...
+      if (contig.has_value() && !contig.value()) {
+        // If TensorView is marked discontiguous in inner dimension, we cannot
+        // vectorize regardless of input.
+        return 1l;
+      }
+      // Innermost dimension
+      contiguous = true;
+      vec_size =
+          std::min(vec_size, scheduler_utils::maxVectorizationWidth(size));
+    } else {
```
zasdfgbnm, line 357: Are you assuming rfactor domain == allocation domain? I don't think this works with @protonu's allocation domain support. I think we should check `contiguity[-1]` and `allocation_domain[-1]`'s extent?

jacobhinkle: This code is working on the `at::Tensor` input that comes from expression evaluator, so we just look at sizes and strides.  I had in mind that these would be fusion inputs, but that might not always be the case in a segmented fusion, so this means we'd actually be evaluating preceding segments with the expression evaluator to get that tensor.

Instead, I suppose I should use `SchedulerRuntimeInfo::input_discontig_strides_` and evaluate the sizes. That also has the limitation that we will only have strides of input tensors.

csrc/scheduler/matmul_utils.cpp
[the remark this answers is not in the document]
jacobhinkle: This change could potentially get large so I am thinking I will wait to do that in a follow-up PR.  See #2211 for tracking issue.

csrc/scheduler/matmul_utils.cpp
jacobhinkle: The allocation domain should match `tens.sizes()` and contiguity should reflect that and `tens.strides()` shouldn't it? I am not referring to rfactor or leaf domains anywhere here.

zasdfgbnm: > Instead, I suppose I should use `SchedulerRuntimeInfo::input_discontig_strides_` and evaluate the sizes. That also has the limitation that we will only have strides of input tensors.

I actually checked this, unfortunately I don't think it works. Due to the nature of how we schedule, we should effectively be only checking the innermost dim. For example, if the tensor has shape `(1024, 3)`, we can not vectorize, but IIUC `input_discontig_strides_` will give vectorization `8` because the `1024` is contiguous to the `3`.

csrc/scheduler/matmul_utils.cpp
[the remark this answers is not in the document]
zasdfgbnm: I don't think this will work. IIRC, currently our double buffering lowering pass assumes all double/circular buffers at the same level has the same stage.

csrc/scheduler/matmul_utils.cpp
jacobhinkle: > This is scary.

I agree.

> Should we make `ExpressionEvaluator::getValue` public so that we can avoid using `evaluate` if we just want to get an existing value? This can be in a separate PR.

That's one way to go. It might be simpler though to just expose `input_discontig_strides_` with an accessor so that we can use it here, along with using `ExpressionEvaluator` on the allocation domain extents to get the sizes. That functionality does not use the expression evaluator at all so it is safe; it records strides directly in the `SchedulerRuntimeInfo` ctor. I think I can fit that change into this PR and it will make things much cleaner.

We also need to think about what happens when we have a segmented fusion. I suppose when a `TensorView` is not a complete Fusion input then we should have a special case where we assume its allocation domain matches its rfactor.

csrc/scheduler/registry.cpp
```diff
@@ -56,24 +57,7 @@ SchedulerRuntimeInfo::SchedulerRuntimeInfo(
...
-          continue;
-        }
-        auto stride = alloc_strides.at(dim);
-        if (stride != expected_stride) {
-          input_discontig_strides_[fusion_inp].push_back(stride * dtype_size);
-          expected_stride = stride;
-        }
-        expected_stride *= size;
-      }
+      input_strides_elts_[fusion_inp] = alloc_strides.vec();
```
zasdfgbnm, line 60: I think for pointwise fusion, we still want to use `input_discontig_strides_`, because for a float tensor of shape `(102400000, 3)`, we want to vectorize 4. But this change will make it un-vectorizable?

jacobhinkle: You're right. We need to preserve that and also record the unmodified stride for use on its own.

jacobhinkle: Pushed a less invasive change that preserves previous behavior but adds a new stride member and accessor.

csrc/scheduler/matmul_utils.cpp
jacobhinkle: I see what you're saying.

So for example given an `at::Tensor x` with
```
x.sizes() = 2 3 4
x.strides() = 15 5 1
```
This is equivalent in nvFuser to `TensorView* tv0` with
```
tv0->getMaybeRFactorDomain() = [ iS0{2} iS1{3} iS2{4} ]
tv0->getMaybeAllocationDomain() = [ iS0{2} iS1{3} iS2{4} ]
tv0->getContiguity() = [ t f t ]
```
Now if we have an `at::Tensor y` with
```
y.sizes() = 2 3 4
y.strides() = 16 1 4
```
then this is equivalent to `TensorView* tv1` with
```
tv1->getMaybeRFactorDomain() = [ iS0{2} iS1{3} iS2{4} ]
tv1->getMaybeAllocationDomain() = [ iS0{2} iS2{4} iS1{3} ]
tv1->getContiguity() = [ t f t ]
```
We see that `y.strides()` and `tv1->getContiguity()` do not match.

For this utility I don't really need the strides to be ordered as long as they are in the same order as sizes. But I think determining whether the Fusion has the dimension marked as contiguous might require a utility similar to the one you posted.

jacobhinkle: @zasdfgbnm if I evaluate the sizes of the non-reduction IterDomains in `tv->getMaybeAllocationDomain()` and then I reverse sort the strides from the `at::Tensor`, won't those always correspond to one another? That means the only change I'd need to make is to add the stride sorting step.

jacobhinkle:
!build

jacobhinkle:
!build

jacobhinkle:
!build

jacobhinkle:
!build

csrc/scheduler/matmul_utils.cpp
zasdfgbnm: > @zasdfgbnm if I evaluate the sizes of the non-reduction IterDomains in `tv->getMaybeAllocationDomain()` and then I reverse sort the strides from the `at::Tensor`, won't those orders always correspond to one another and to that of `tv->getContiguity()`? That means the only change I'd need to make is to add the stride sorting step.

I think the only way that is guaranteed to be correct is to look at the permutation relationship of `getMaybeAllocationDomain` and `getMaybeRFactorDomain`. Looking at numbers does not always provide correct information. For example, for the same tensor of shape `(4, 5)` and stride `(5, 1)`, it can be interpreted as two different options:
```
allocation domain: I0{4}, I1{5}
contiguity: T, T
```
and
```
allocation domain: I1{5}, I0{4}
contiguity: F, F
```
Both are correct, although the resulting performance can be significantly different. Our code should handle both cases.

csrc/scheduler/matmul_heuristic.h
```diff
@@ -64,16 +64,18 @@ class MatmulParams : public HeuristicParams {
     }
   };
 
-  //! This is the maximum vectorization supported by the inputs.
+  //! This is the maximum vectorization supported by the inputs. This refers to
+  //! the number of data elements loaded simultaneously, not the number of
+  //! bytes.
```
zasdfgbnm, line 69: Could you add a comment describing why the default value is 16? This can be confusing.

csrc/scheduler/registry.h
```diff
@@ -69,6 +69,18 @@ class SchedulerRuntimeInfo : public NonCopyable {
   //!  return max_alignment_size_in_byte.
   size_t getAlignmentSize(TensorView* tv);
 
+  //! Returns strides of tensor, in elements instead of bytes. Only works for
+  //! complete Fusion inputs.
+  const std::vector<int64_t>& getInputStrides(TensorView* tv) {
```
zasdfgbnm, line 74: Can we return the stride in the language of allocation domain instead of rfactor domain?

zasdfgbnm (approved):
I still think `maxRowVectorization` needs change, but consider that:
1. The rest of this PR LGTM
2. Even as-is, this PR is a strict improvement on what we have on main
3. I will be off tomorrow and next week, I don't want to block this for one week.
I am OK with merging this as-is and fix `maxRowVectorization` in a future PR.

csrc/scheduler/matmul_heuristic.h
jacobhinkle: I will add that comment; I don't think there's a really compelling reason to change it. I agree that it's confusing, and I've had to set it manually to fix a lot of the benchmarks and tests, but now that it's set I think it might be nice because it is a bit more explicit. With automatic scheduling it should always be set to a sane value. One reason I set it higher to a value that I knew was invalid was as a signal that we've set it properly in each of those cases.

csrc/scheduler/registry.h
jacobhinkle: Yeah I can do that. Essentially it's just a reverse sort but we can do that when saving it instead of at the call site.

csrc/scheduler/matmul_utils.cpp
jacobhinkle: Ah I see. Yeah you are right that my logic only works out if the tensor is contiguous. I'll change it!

jacobhinkle:
!build

csrc/scheduler/registry.cpp
```diff
@@ -57,6 +57,23 @@ SchedulerRuntimeInfo::SchedulerRuntimeInfo(
...
+        // Save the strides in order of allocation domain in case the
+        // allocation domain is a permutation of RFactor domain
+        std::vector<int64_t> orig_strides = alloc_strides.vec();
+        std::vector<int64_t> ordered_strides;
+        ordered_strides.reserve(orig_strides.size());
+        NVF_ERROR(orig_strides.size() == alloc_perm_opt->size());
+        for (int64_t i : alloc_perm_opt.value()) {
+          ordered_strides.push_back(orig_strides[i]);
+        }
+        input_strides_elements_[fusion_inp] = ordered_strides;
```
zasdfgbnm, line 74: nit:
```suggestion
        input_strides_elements_[fusion_inp] = std::move(ordered_strides);
```

csrc/scheduler/matmul_utils.cpp
```diff
@@ -294,6 +315,161 @@ std::string isMatmulFusionDefinitionSupported(
...
+    }
+    for (TensorView* tv : it->second) {
+      // TODO: handle the case when tv is not a Fusion input by filling default
+      // contiguous strides based on tv->getMaybeAllocationDomain() and data
+      // pointer aligned to 16 bytes.
+
+      const std::vector<IterDomain*>& alloc = tv->getMaybeAllocationDomain();
+
+      // Get sizes for non-reduction dims in allocation domain
+      std::vector<int64_t> sizes;
```
zasdfgbnm, line 411: Could we also move the size to `runtime_info` to share the permutation code of stride?

jacobhinkle: Makes sense. I was just using expression evaluator to get that but it's more compact to just load it directly.

csrc/scheduler/matmul_utils.cpp
```diff
@@ -294,6 +315,161 @@ std::string isMatmulFusionDefinitionSupported(
...
+        // element. In such cases we cannot vectorize. An exception is when an
+        // outer dimension has both stride==1 and size==1, in which case we
+        // could ignore its stride. Currently we ignore that case and just
+        // return 1 whenever we detect multiple stride==1 dimensions.
+        return 1l;
+      }
+      inner_contiguous = true;
+      vec_size =
+          std::min(vec_size, scheduler_utils::maxVectorizationWidth(size));
+    } else {
```
zasdfgbnm, line 379: Would it be simpler if we just do
```suggestion
    if (!*noreductions_contig.back()) {
      return 1;
    }
    if (i == sizes.size() - 1) {
      vec_size =
          std::min(vec_size, scheduler_utils::maxVectorizationWidth(size));
    } else {
```

jacobhinkle: Good call on moving the inner contiguity check out of the loop. I simplified this code a bit in the latest push.

jacobhinkle:
!build

jacobhinkle:
!build

Issue #1693, opened 2024-01-29
Make python benchmarks filterable on sizes


Priya2698:
Since the sizes print as `(64, 1024)` it is difficult to filter the benchmarks on sizes. We can modify the ids for the `size` parameter so that the printed benchmark names are easier to filter.
Thanks to @jacobhinkle  for bringing this up: https://github.com/NVIDIA/Fuser/pull/1670#discussion_r1466939679.

PR #2108, opened 2024-04-18
Add CLI options for benchmarking, update test name printing

Priya2698:
1. Adds CLI options for controlling number of `warmup_rounds/rounds`: This will be useful when we run python benchmarks on the Github CI to control the total time required to run the benchmark suite.
2. Updates how complete test names are printed to allow for better filtering: Closes Issue #1693. The new format of test names is:
`test_name[argname=value]`. 
For example: `test_transpose_nvf_benchmark[axes=[0_1]-dtype=torch.float32-size=[2_2_16]]`. 
The tuple values are `_` joined to allow filtering.

### DIFF
--- a/benchmarks/python/conftest.py
+++ b/benchmarks/python/conftest.py
@@ -28,15 +28,17 @@ def pytest_addoption(parser):
         help="Benchmarks torch.compile mode.",
     )
 
+    # pytest-benchmark does not have CLI options to set rounds/warmup_rounds for benchmark.pedantic.
+    # The following two options are used to overwrite the default values through CLI.
     parser.addoption(
-        "--rounds",
+        "--benchmark-rounds",
         action="store",
         default=10,
         help="Number of rounds for each benchmark.",
     )
 
     parser.addoption(
-        "--warmup-rounds",
+        "--benchmark-warmup-rounds",
         action="store",
         default=1,
         help="Number of warmup rounds for each benchmark.",
@@ -64,8 +66,10 @@ def pytest_benchmark_update_machine_info(config, machine_info):
 
 
 def pytest_configure(config):
-    BENCHMARK_CONFIG["rounds"] = int(config.getoption("--rounds"))
-    BENCHMARK_CONFIG["warmup_rounds"] = int(config.getoption("--warmup-rounds"))
+    BENCHMARK_CONFIG["rounds"] = int(config.getoption("--benchmark-rounds"))
+    BENCHMARK_CONFIG["warmup_rounds"] = int(
+        config.getoption("--benchmark-warmup-rounds")
+    )
 
 
 def pytest_collection_modifyitems(session, config, items):
--- a/benchmarks/python/conftest.py
+++ b/benchmarks/python/conftest.py
@@ -2,7 +2,7 @@
 # All rights reserved.
 # SPDX-License-Identifier: BSD-3-Clause
 import pytest
-from .core import DEVICE_PROPERTIES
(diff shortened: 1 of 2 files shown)


benchmarks/python/conftest.py
```diff
@@ -28,6 +28,20 @@ def pytest_addoption(parser):
...
+        help="Number of rounds for each benchmark.",
+    )
+
+    parser.addoption(
+        "--warmup-rounds",
+        action="store",
+        default=1,
+        help="Number of warmup rounds for each benchmark.",
+    )
+
```
naoyam, line 44: Looks like pytest-benchmark itself has some related options:

https://pytest-benchmark.readthedocs.io/en/latest/usage.html#commandline-options

Do we need these our own options? Also, pytest-benchmark uses "rounds" and "iterations", and I'm not sure what the difference is.

Priya2698: There is no command line option to control the `rounds/iterations` in the pedantic mode (https://pytest-benchmark.readthedocs.io/en/latest/pedantic.html). They are passed directly to `benchmark.pedantic`. These CLI options are essentially a utility to update those numbers. 

Each round has a given number of iterations. When we use a setup function (which is our case to clear L2 cache), iterations=1.

naoyam: Thanks. What do the round and the iteration mean?

Priya2698: We execute a given target function for `n` rounds with each round itself consists of `k` iterations -- that is within each round, the target function is executed `k` times. So in total, we execute the benchmark `nk`. The statistics of the benchmarks are based on the duration of these rounds (mean, median..).  The motivation behind iterations seem to be able to measure functions that have smaller duration than the precision of the timer.

Reference: https://github.com/ionelmc/pytest-benchmark/issues/186

naoyam: Can you please leave a comment that the default pytest-benchmark options are not available as we use the pedantic mode?

naoyam (approved):
LGTM

benchmarks/python/conftest.py
```diff
@@ -39,14 +53,21 @@ def disable_benchmarking(request):
     return request.config.getoption("--disable-benchmarking")
 
 
-def pytest_make_parametrize_id(val):
-    return repr(val)
+def pytest_make_parametrize_id(val, argname):
+    if isinstance(val, tuple):
+        return f'{argname}=[{"_".join(str(v) for v in val)}]'
+    return f"{argname}={repr(val)}"
```
jacobhinkle, line 59: Cool. I didn't know about this hook. In the future we might want to extend it further but this looks good to me for now.

benchmarks/python/conftest.py
```diff
@@ -28,6 +28,20 @@ def pytest_addoption(parser):
...
+        default=10,
+        help="Number of rounds for each benchmark.",
+    )
+
+    parser.addoption(
+        "--warmup-rounds",
+        action="store",
+        default=1,
+        help="Number of warmup rounds for each benchmark.",
+    )
```
jacobhinkle, line 43: I think these options should be prefixed with `benchmark-`, like `--benchmark-rounds`. This matches the other default options from pytest-benchmark such as `--benchmark-min-rounds`.

jacobhinkle (approved):
LGTM. Just a note about the argument names.

benchmarks/python/conftest.py
Priya2698: Great point!

PR #2110, opened 2024-04-18
Validate with relaxed tolerances for failed cases

Priya2698:


nvfuser/__init__.py
```diff
@@ -360,15 +360,22 @@ def validate(
...
                 fusion_output
             ):
-                assert torch.allclose(
+                try:
+                    assert torch.allclose(
                     fusion_output, reference_output, atol=atol, rtol=rtol
-                ), f"Max error: {torch.abs(torch.max(fusion_output - reference_output))}, \
+                ), f"Max error nvFuser output[{inx}]: {torch.abs(torch.max(fusion_output - reference_output))}, \
+                    Absolute tolerance: {atol}, Relative tolerance: {rtol}"
+                except:
```
xwang233, line 368: Isn't this try-except block equivalent to `assert torch.allclose(xxx, yyy, 1.1*atol, 1.1*rtol)` ?

Priya2698: Yes, it's not the final code (just a quick draft for verification). It will actually be based on whether the function argument specifies testing using relaxed tolerances.
But I can remove it in this version

PR #2111, opened 2024-04-19
Use proper dtype in generateSharedMemoryEpilogueHeuristics

jacobhinkle:
This prevents underestimation of smem epilogue use in the default heuristic, which is rarely used.


jacobhinkle:
!build --diff

jacobhinkle:
!build --diff

zasdfgbnm (approved):

PR #2119, opened 2024-04-19
TMA doc part 4

zasdfgbnm:


doc/dev/tma.md
```diff
@@ -171,6 +171,245 @@ and the former is not equivalent to the latter as discussed in [Divisibility of
...
+The Figure 8 below shows some invalid schedule on the shared memory tensor:
+
+![Figure 8: Some invalid smem schedule](tma/invalid-smem-discontiguous.svg)
+
+These examples have the same reason for being invalid: The tile in shared memory is not contiguous.
+
+In the example (a), the allocated IterDomains are `[I3, I7, I8, I6]`
+(note that `I5` is parallelized on `blockIdx.x`, it is not allocated on shared memory).
+The tile IterDomains are `I7` and `I6`, there is an `I8` between them,
+makeing the tile not contiguous in memory.
```
naoyam, line 196: typo: making

doc/dev/tma/valid-smem-contig.svg
naoyam, line 1: Some domains are colored with green and light blue. Does that indicate it's a box and tile domain? IIRC, a box is equivalent to a tile if not strided.

zasdfgbnm: yes

doc/dev/tma.md
```diff
@@ -171,6 +171,245 @@ and the former is not equivalent to the latter as discussed in [Divisibility of
...
+In the example (b), the allocated IterDomains are `[(partial) I1, I2]`,
+where `I2` contains not only `I6` but also `I5`,
+so `I4` and `I6` is not contiguous in memory.
+Similar argument applies to example (d).
+
+In the example (c), the allocated IterDomains are `[I3, I4, I5, I6]`.
+The `I5` between `I4` and `I6` is parallelized on `TIDx`, therefore allocated.
+So `I4` and `I6` is not contiguous in memory.
+Similar argument applies to example (e).
+
```
naoyam, line 207: Could you also comment on (d) and (e)?

doc/dev/tma/arbitrary-view.svg
naoyam, line 1: In the case of (a), are all of `i4`, `i6`, `i7`, `i8` and `i9` are considered tile domains? Or just `i9` and `i8`?

zasdfgbnm: Good question: I sometimes feel that it is convenient to call all of them "tile ID", sometime I just want to call `I4` and `I6` "tile ID". And I don't have a clear definition on how should they be called...

Probably invent more names? tile ID, original tile ID, allocation tile ID, leaf tile ID...

doc/dev/tma.md
zasdfgbnm: fixed

doc/dev/tma.md
zasdfgbnm: added

doc/dev/tma.md
```diff
@@ -171,6 +171,251 @@ and the former is not equivalent to the latter as discussed in [Divisibility of
...
+we must make sure that the allocation domain of our shared memory tensor is consistent with the memory format defined by TMA.
+
+Let `s` be swizzle size (`bytes of swizzle / 16 bytes`).
+The overall structure of the schedule of the allocation domain of the shared memory tensor looks like:
+
+```
+tile IterDomains => view as ([-1, s, 8/s, s, 16B]) => swizzle(1, 3) => arbitrary view
+```
+
+The Figure 12 below shows an example schedule of the consumer of a TMA load of a `fp16` tensor with 64 byte swizzle:
```
naoyam, line 377: Here, the tile domain looks like a 2D domain of `[i5*i7, i9]`, which is a square matrix. I suppose this doesn't need to be a square, right?

doc/dev/tma.md
```diff
@@ -171,6 +171,251 @@ and the former is not equivalent to the latter as discussed in [Divisibility of
...
+</details>
+
+When we use the automatic swizzling feature of TMA,
+we must make sure that the allocation domain of our shared memory tensor is consistent with the memory format defined by TMA.
+
+Let `s` be swizzle size (`bytes of swizzle / 16 bytes`).
+The overall structure of the schedule of the allocation domain of the shared memory tensor looks like:
+
+```
+tile IterDomains => view as ([-1, s, 8/s, s, 16B]) => swizzle(1, 3) => arbitrary view
```
naoyam, line 374: Isn't `8/s` always supposed to be 2?

doc/dev/tma.md
zasdfgbnm: Right, it doesn't have to be square

doc/dev/tma.md
zasdfgbnm: For 32 byte swizzle, this is always 4. For 64 byte swizzle, this is always 2. For 128 byte swizzle, this is always 1 (so we can just omit this and schedule a `[-1, s, s, 16B]`).

naoyam: OK, I see. When `n = 32`, the xor swizzle steps is only done once since `32 / 16 = 2`, right?

doc/dev/tma.md
zasdfgbnm: Modified the test to not use a square box

doc/dev/tma.md
zasdfgbnm: Yes

zasdfgbnm: Updated the doc mentioning the concrete values for each swizzle mode.

naoyam (approved):
LGTM.

PR #2120, opened 2024-04-20
Matmul vectorization heuristic

jacobhinkle:
This introduces `supported_vec_size`, a struct attribute of `MatmulParams` that indicates the maximum vectorization supported by operands A and B as well as epilogue tensors (which will all be vectorized at the same width).

Note that we expect a temporary performance regression since we hardcode vec_size of 4 for the output, which is suboptimal for half-precision outputs. This will be remedied when we start inferring the proper vectorization (see #2105).


jacobhinkle:
!build

jacobhinkle:
!build

csrc/scheduler/matmul.cpp
```diff
@@ -1014,8 +1034,8 @@ void scheduleMatmul(Fusion* fusion, const MatmulParams& params) {
   // Schedule prolog:
   //   TODO: this section needs more configurability.
   // ------------------------------------------------------------------
-  scheduleProlog(acw_smem, params);
-  scheduleProlog(bcw_smem, params);
+  scheduleProlog(acw_smem, (int)params.supported_vec_size.a, params);
+  scheduleProlog(bcw_smem, (int)params.supported_vec_size.b, params);
```
zasdfgbnm, line 1038: nit: use `int64_t` as parameter of `scheduleProlog` so we do not need to cast

zasdfgbnm (approved):

Issue #2121, opened 2024-04-21
Gather/PipelineTestTwoStages.Communication/40 failed at head. 


wujingyue:
```
$ git rev-parse HEAD
fa0cf87dc61c69e21f6e1a86c8bde00b7469df65
```

```
$ NVFUSER_MULTIDEVICE_DEBUG_PRINT=1 NVFUSER_MULTIDEVICE_DEBUG_BARRIER=1 mpirun -np 4 bin/test_multidevice --gtest_filter=Gather*40
...
C++ exception with description "Expected a proper Tensor but got None (or an undefined Tensor in C++) for argument #0 'self'
Exception raised from checked_cast_variable at /opt/pytorch/pytorch/torch/csrc/autograd/VariableTypeManual.cpp:66 (most recent call first):
frame #0: c10::Error::Error(c10::SourceLocation, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) + 0xae (0x7fc6cd015afe in /usr/local/lib/python3.10/dist-packages/torch/lib/libc10.so)
frame #1: c10::detail::torchCheckFail(char const*, char const*, unsigned int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) + 0xf3 (0x7fc6ccfbf591 in /usr/local/lib/python3.10/dist-packages/torch/lib/libc10.so)
frame #2: <unknown function> + 0x5435249 (0x7fc6c1ff0249 in /usr/local/lib/python3.10/dist-packages/torch/lib/libtorch_cpu.so)
frame #3: <unknown function> + 0x432f000 (0x7fc6c0eea000 in /usr/local/lib/python3.10/dist-packages/torch/lib/libtorch_cpu.so)
frame #4: <unknown function> + 0x263f2a2 (0x7fc6bf1fa2a2 in /usr/local/lib/python3.10/dist-packages/torch/lib/libtorch_cpu.so)
frame #5: at::_ops::slice_Tensor::call(at::Tensor const&, long, std::optional<c10::SymInt>, std::optional<c10::SymInt>, c10::SymInt) + 0x271 (0x7fc6bf1b23b1 in /usr/local/lib/python3.10/dist-packages/torch/lib/libtorch_cpu.so)
frame #6: <unknown function> + 0x491c93 (0x5562f0080c93 in bin/test_multidevice)
frame #7: <unknown function> + 0x49249b (0x5562f008149b in bin/test_multidevice)
[...]

PR #2122, opened 2024-04-22
Fix empty tensor when printing comms

samnordmann:
Fixes issue https://github.com/NVIDIA/Fuser/issues/2121

### DIFF
--- a/csrc/multidevice/executor.cpp
+++ b/csrc/multidevice/executor.cpp
@@ -255,7 +255,8 @@ std::ostream& MultiDeviceExecutor::print() {
   int communication_counter = 0;
   for (auto group : workspace.group_run_order) {
     if (is_resharding_[group]) {
-      debug() << "Communication " << communication_counter << ": " << group->exprs().at(0) << "}\n";
+      debug() << "Communication " << communication_counter << ": "
+              << group->exprs().at(0) << "}\n";
       communication_counter++;
     } else {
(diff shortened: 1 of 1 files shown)


samnordmann:
!build --dist

csrc/multidevice/executor.cpp
```diff
@@ -255,12 +255,8 @@ std::ostream& MultiDeviceExecutor::print() {
...
   for (auto group : workspace.group_run_order) {
     if (is_resharding_[group]) {
-      debug() << "Communication " << communication_counter << ":{\n";
-      for (const auto& comm :
-           lowerCommunication(comm_.deviceId(), group->exprs().at(0), {}, {})) {
-        debug() << comm->toString(2) << "\n";
-      }
-      debug() << "}\n";
+      debug() << "Communication " << communication_counter << ": "
+              << group->exprs().at(0) << "}\n";
```
wujingyue, line 259: Does this assume a communication group have only one `Expr` so printing expr 0 is sufficient?

wujingyue, line 259:
```suggestion
              << group->exprs().at(0) << "\n";
```

wujingyue (approved):

csrc/multidevice/executor.cpp
samnordmann: correct. This assumption was holding also before the present patch

PR #2123, opened 2024-04-22
Keep actual tensors in `getMaybeHeuristicsFor`

### LINKED ISSUE #2118: Expose pointer alignment info to scheduler heuristics
Refernce: https://github.com/NVIDIA/Fuser/pull/2105#discussion_r1572736106_
 
As stated there, if we provide input tensors whose first element is not aligned at 16 bytes, we cannot detect that in our heuristics. We should remedy this somehow. One approach would be to leave the original inputs in the `runtime_info` and then replace them with metadata args instead after scheduling.

jacobhinkle:
In #1649, a change to `FusionKernelRuntime::getMaybeHeuristicsFor` was made which replaced all inputs with "metadata-only" tensors. These tensors have the same size, stride, and dtype as the actual inputs, but their data pointer is always `nullptr` and they do not allocate any memory. This interferes with `runtime_info.ptrOf` as seen in #2118. Instead, in this PR I revert that part of #1649. The `ArgumentManager` that holds the actual `at::Tensor`s is destroyed at the end of `getMaybeHeuristicsFor`, as is the `PrecomputedValues`. I don't think there are any dangling references to the actual tensors in the args, so I _think_ this should not leak device memory: cc @rdspring1 to confirm.

See the included passing tests, which fail on main.

Fixes #2118.


jacobhinkle:
!build --diff-bench

jacobhinkle:
!build

jacobhinkle:
!build

rdspring1:
You may hit an OOM memory in one of the benchmark tests.

**Alternative Idea:**
`ArgumentManager` destroys `at::Tensor` args to save memory once they're used. We're using this at [compile-time](https://github.com/NVIDIA/Fuser/blob/main/csrc/kernel_cache.cpp#L1245-L1256) and [runtime](https://github.com/NVIDIA/Fuser/blob/main/csrc/kernel_cache.cpp#L1420-L1435). At compile-time, this isn't useful because all intermediate tensors are metadata only and we destroy the required input tensors. If you separate those cases, you could pass a `KernelArgumentHolder` reference and avoid the extra copy.

jacobhinkle:
> You may hit an OOM memory in one of the benchmark tests.

Do you know which structure is holding onto the `at::Tensor`? It seems like `ArgumentManager` and `PrecomputedValues` are the only variables that could hold onto it and they should not live longer than `getMaybeHeuristicsFor`.

> **Alternative Idea:** `ArgumentManager` destroys `at::Tensor` args to save memory once they're used. We're using this at [compile-time](https://github.com/NVIDIA/Fuser/blob/main/csrc/kernel_cache.cpp#L1245-L1256) and [runtime](https://github.com/NVIDIA/Fuser/blob/main/csrc/kernel_cache.cpp#L1420-L1435). At compile-time, this isn't useful because all intermediate tensors are metadata only and we destroy the required input tensors. If you separate those cases, you could pass a `KernelArgumentHolder` reference and avoid the extra copy.

Ah I see. I think the copy of `at::Tensor` is light-weight though isn't it (I could be wrong)?

jacobhinkle:
Given the test failures, it looks like the reduction scheduler might need some finessing.

jacobhinkle:
!build

csrc/kernel_cache.cpp
```diff
@@ -1610,15 +1610,14 @@ std::optional<FusionKernelRuntime::HeuristicsPtr> FusionKernelRuntime::
...
 
-  // Store metadata copy of arguments for ArgumentManager
-  KernelArgumentHolder args_metadata = copyMetadataArg(args);
+  KernelArgumentHolder args_copy(args);
 
   // ArgumentManager manipulates the KernelArgumentHolder argument.
-  // We make another metadata copy for the PrecomputedValues.
-  KernelArgumentHolder complete_fusion_metadata_args(args_metadata);
+  // We make a copy for the PrecomputedValues.
+  KernelArgumentHolder complete_fusion_args(args_copy);
```
wujingyue, line 1617: Can someone teach me why this copy is required? The above comment seems to describe the copy at Line 1613 not the copy here.

wujingyue: Sorry I'm still missing the context here. 

Am I right that, with this PR, `args`, `args_copy` and `complete_fusion_args` all contain actual tensors not metadata? I understood that `args_copy` is to allow `args_manager` to modify the passed-in arguments. But how does the code break if Line 1644 use `args` instead of `complete_fusion_args`?

jacobhinkle: Yeah I think that works fine. `args_copy` is needed since `args_manager` needs a mutable reference but `PrecomputedValues::bindValues` can take the const reference `args`. I just pushed a change to remove `complete_fusion_args` and rename `args_copy` to `mut_args`.

csrc/scheduler/vectorize_helper.cpp
```diff
@@ -840,6 +840,14 @@ int64_t getVectorizationFactor(
         max_vec_size,
         SchedulerRuntimeInfo::max_alignment_size_in_byte / dtype_size);
 
+    // Handle misaligned inputs or outputs. Usually tensors are fully aligned at
```
naoyam, line 843: Isn't the code below also for checking alignment?

jacobhinkle: Oh you're right. Not sure how I missed that. I'll strip out these changes to only the part that keeps the actual tensors in `getMaybeHeuristicsFor`.

jacobhinkle:
!build

csrc/scheduler/utils.h
```diff
@@ -42,12 +42,24 @@ constexpr int64_t y_grid_limit = 65535;
...
+  // For example
+  //   n               = b101101000
+  //           n - 1   = b101100111
+  //        ~ (n - 1)  = b010011000
+  //   n & (~ (n - 1)) = b000001000
+  // The key is that subtracting one flips all trailing 0s as well as the least
+  // significant 1, so all of the other bits will fail the &, leaving
+  // only that 1.
+  return n & (~(n - 1));
 }
```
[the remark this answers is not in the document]
jacobhinkle: This change wasn't needed. I reverted it. It could probably go in a separate cleanup PR.

PR #2126, opened 2024-04-22
Move code from SetUp to constructor.

wujingyue:
https://google.github.io/googletest/faq.html#CtorVsSetUp

### DIFF
--- a/tests/cpp/multidevice.cpp
+++ b/tests/cpp/multidevice.cpp
@@ -20,19 +20,11 @@ MultiDeviceTest::MultiDeviceTest() {
   communicator = getOrCreateCommunicator();
   tensor_options =
       at::TensorOptions().dtype(at::kFloat).device(communicator->device());
-  if (getNvFuserEnv("MULTIDEVICE_DEBUG_PRINT")) {
-    debug_print = true;
-  }
-  if (getNvFuserEnv("MULTIDEVICE_DEBUG_BARRIER") &&
-      communicator->is_available()) {
-    do_barrier_at_test = true;
-  }
-  if (getNvFuserEnv("MULTIDEVICE_DISABLE_SKIP")) {
-    disable_skip = true;
-  }
-  if (!disable_skip && !communicator->is_available()) {
-    GTEST_SKIP() << "This test needs an available communicator.";
-  }
+  debug_print = getNvFuserEnv("MULTIDEVICE_DEBUG_PRINT") != nullptr;
+  do_barrier_at_test =
+      (getNvFuserEnv("MULTIDEVICE_DEBUG_BARRIER") != nullptr &&
+       communicator->is_available());
+  disable_skip = getNvFuserEnv("MULTIDEVICE_DISABLE_SKIP") != nullptr;
 }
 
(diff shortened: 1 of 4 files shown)


wujingyue:
!build --dist

wujingyue:
!build --dist

wujingyue:
I'll take a look at the failed distributed tests tomorrow. It's unfortunate that these tests currently require 4 GPUs, which I often don't have access to.

wujingyue:
!build --dist

wujingyue:
!build --dist

wujingyue:
CI errors seem irrelevant to this PR. One is a numerical mismatch that seems flaky, and the other seems to be a timeout in benchmark. cc @xwang233

tests/cpp/multidevice.cpp
```diff
@@ -16,32 +16,52 @@
...
-    disable_skip = true;
+  debug_print = getNvFuserEnv("MULTIDEVICE_DEBUG_PRINT") != nullptr;
+  do_barrier_at_test =
+      (getNvFuserEnv("MULTIDEVICE_DEBUG_BARRIER") != nullptr &&
+       communicator->is_available());
+  disable_skip = getNvFuserEnv("MULTIDEVICE_DISABLE_SKIP") != nullptr;
+}
+
+MultiDeviceTest::~MultiDeviceTest() {
+  if (do_barrier_at_test && communicator->is_available()) {
```
cowanmeg, line 31: I think we want to keep the optional barrier in TearDown because after each test we want to ensure all devices completed before moving onto the next test. Then, in the destructor, we want a non-optional barrier. 
wdyt?

tests/cpp/multidevice.cpp
```diff
@@ -16,32 +16,52 @@
...
+  auto it = std::find(devices.begin(), devices.end(), deviceId);
+  if (it != devices.end()) {
+    i = std::distance(devices.begin(), it);
   }
-  NVFuserTest::TearDown();
+  return tensor.slice(sharded_dim, i, i + 1).contiguous();
+}
+
+/*static*/ Communicator* MultiDeviceTest::getOrCreateCommunicator() {
+  static Communicator* communicator = new Communicator();
```
cowanmeg, line 63: Does this create a new communicator each time the function is called?

wujingyue: No. The keyword `static` guarantees it's computed only the first time the function is called.

tests/cpp/multidevice.cpp
wujingyue: https://google.github.io/googletest/faq.html#CtorVsSetUp

> For each TEST_F, GoogleTest will create a fresh test fixture object, immediately call SetUp(), run the test body, call TearDown(), and then delete the test fixture object.

I know it's a bit counter intuitive but the destructor is actually called around the same time as TearDown.

cowanmeg: Got it! Thanks for the explanation!

cowanmeg (approved):

wujingyue:
!build --dist

PR #2129, opened 2024-04-23
Add sanitize link option to final object, too.

tfogal:
CMake does not forward private flags.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -294,12 +294,13 @@ target_link_libraries(codegen_internal PUBLIC
   dl
 )
 
+add_library(nvfuser_codegen SHARED $<TARGET_OBJECTS:codegen_internal>)
 if(NVFUSER_BUILD_WITH_ASAN)
   target_compile_options(codegen_internal PRIVATE -fsanitize=address)
   target_link_options(codegen_internal PRIVATE -fsanitize=address)
+  target_link_options(nvfuser_codegen PRIVATE -fsanitize=address)
 endif()
 
-add_library(nvfuser_codegen SHARED $<TARGET_OBJECTS:codegen_internal>)
 target_include_directories(nvfuser_codegen PUBLIC
   "$<BUILD_INTERFACE:${NVFUSER_SRCS_DIR}>"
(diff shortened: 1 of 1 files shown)


tfogal:
!build

xwang233 (approved):

CMakeLists.txt
```diff
@@ -294,12 +294,13 @@ target_link_libraries(codegen_internal PUBLIC
   dl
 )
 
+add_library(nvfuser_codegen SHARED $<TARGET_OBJECTS:codegen_internal>)
 if(NVFUSER_BUILD_WITH_ASAN)
   target_compile_options(codegen_internal PRIVATE -fsanitize=address)
   target_link_options(codegen_internal PRIVATE -fsanitize=address)
+  target_link_options(nvfuser_codegen PRIVATE -fsanitize=address)
```
jjsjann123, line 301: nitpick question, should we change ASAN link option to be public on `codegen_internal` instead?

jjsjann123 (approved):
minor question

CMakeLists.txt
tfogal: yeah I like that idea better, actually

tfogal: edit: wait, no: making it `PUBLIC` would ensure it propagates when one uses `target_link_libraries`, but the `add_library` calls actually use the `$<TARGET_OBJECTS:...>` syntax, which I do not think propagates properties the way we'd want here.

tfogal: double-edit (after I've merged, of course): we should probably use `PUBLIC` anyway, though, in case we decide to move away from `$<TARGET_OBJECTS:...>` sometime down the road. another PR...

jjsjann123: > edit: wait, no: making it `PUBLIC` would ensure it propagates when one uses `target_link_libraries`, but the `add_library` calls we have actually use the `$<TARGET_OBJECTS:...>` to get the objects (and thus never do `target_link_libraries`), which I do not think propagates properties the way we'd want here.

Got'ya Thanks for explaining this.

PR #2134, opened 2024-04-23
Fix the box and tile example to be consistent with our indexing

zasdfgbnm:
per the feedback from @jjsjann123


jjsjann123 (approved):
Thanks for updating this.

`+4,129 -3,598` interesting how svg translate to LOC. Nice way to boost your stats. :laughing:

PR #2136, opened 2024-04-23
Rewrite argument building to avoid allocs.

### LINKED ISSUE #2135: Reducing unaccounted for launch latency by removing allocations
We should endeavor to 2--3us launch latencies, which should be achievable even with streams (i.e. without moving to cuda graphs).

Below is a number of examples of long latencies in a run of [thunder's `examples/llama2.c/`](https://github.com/Lightning-AI/lightning-thunder/tree/e0ab64867a5be914d0548c195a3f850a76c8c397/examples/llama2.c) port. Here we regularly spend 20--30us launching a kernel.

FusionExecutorrunFusion
FusionExecutorrunFusion
FusionExecutorrunFusion
FusionExecutorrunFusion

One can see that much of the latency comes from ExpressionEvaluator:

training (36 370 s)

But some simple math suggests something more going on. Within `ExecutorRunFusion::GetArgsBuffers`, `ExpressionEvaluator::evaluate` takes 0.460 + 0.504 + 0.388 + 4.117 + 0.576 + 0.459 == 6.504us, or ~54% of the 12.147 GetArgsBuffers time. The other 46% of time is unaccounted for.

*This* issue is about that unaccounted for time (we might consider a separate issue specifically looking at `ExpressionEvaluator` latency). I believe the current unaccounted for time to be an artifact of how we build our `cuLaunchKernel` arguments array.

runtime/tensor.cu is our `Tensor` type, that we use for passing all tensors to the kernel. This is a struct consisting of a data pointer, shape, and strides. To build our `cuLaunchKernel` argument array, then, each tensor argument needs to mimic the layout of this struct. To do this, we create a version of that struct, convert that version to raw bytes, and then append that sequence of bytes to the argument array. Due to implementation details here, we end up creating + resizing multiple `std::vector`s. This necessitates repeated calls to the allocator, which I believe is a significant portion of the unaccounted for time.

One fix would be to pre-create the memory space for <base address, shape array, strides array> (the data fields of our runtime/tensor.cu `Tensor` type), and merely copy the new data into that array. This would obviate the need for many of the allocations.

tfogal:
This PR attempts to pre-allocate space for `Tensor` arguments, as considered in #2135. The key insight is that, while tensor shapes are dynamic, the rank of a tensor is not. This allows us to pre-build the memory space for passing a kernel to the GPU, avoiding numerous allocations on subsequent calls.

For the 4 kernels mentioned in #2135, this improves launch latency by 4.4%, 7.5%, 4.4%, and 16%, respectively. It should be an across-the-board improvement (i.e. no regressions). For comparison to the case highlighted in #2135, the overall `GetArgsBuffers` call drops from 12.147us to 9.512us (~20%):

step 3  45 493 ms

Whether or not this improvement matters depends significantly on the network itself. In many training scenarios, we can already launch quickly enough that the subsequent kernel is enqueued before a current kernel finishes executing. In these scenarios, improving launch latency does not improve overall performance.

After this patch, the ExpressionEvaluator in GetArgsBuffers takes
0.604 + 0.459 + 0.457 + 4.276 + 0.609 + 0.353 == 6.758us, or ~71% of the 9.512us GetArgsBuffers time. This means we are starting to hit diminishing returns from addressing anything sans ExpressionEvaluator.

Fixes #2135.


tfogal:
!build

tfogal:
The CI failure is real:

```
+ python tests/python/test_python_frontend.py -v --save-xml
Running tests...
----------------------------------------------------------------------
00:13:48 Test results will be stored in test-reports/python-unittest/tests.python.test_python_frontend
  test_addcmul (__main__.TestNvFuserFrontend) ... [a4u8g-0069:17800:0:17800] Caught signal 11 (Segmentation fault: address not mapped to object at address (nil))
==== backtrace (tid:  17800) ====
 1 0x000000000042308d nvfuser::FusionExecutor::recomputeArgs()  :0
 2 0x000000000043343e nvfuser::FusionExecutor::runFusion()  ???:0
```

What's happening is that the `entry.args` array is 0 elements but the `recomputeArgs` method is expecting it to be 3 elements.

I happened to notice that this works fine if we comment out the `serde_check` decorator in test_python_frontend.py. @rdspring1 educated me that we actually have a similar but not exactly the same path for deserialization, and what's assuredly happening is that [when I deserialize the ExecutorEntry](https://github.com/NVIDIA/Fuser/blob/dad05c3f5fc6eee85057380cdd0643ee84bb0475/csrc/executor.cpp#L2358), I am not resetting the `entry.args` arrays to be properly sized. This makes a subsequent `runFusion` *think* that the ExecutorEntry is good to go, but in reality it was never sized for the arguments.

As Ryan mentioned, we probably do not want to serialize `ExecutorEntry::args` or `ExecutorEntry::arg_ptrs`. Those are pointers to memory and so they're nonsensical to serialize. Right now we call `computeArgs` (which sizes `::args` and `::arg_ptrs`) only when first initializing the ExecutorEntry; we should insert a call to `computeArgs` on the `::deserialize` path as well.

Huge thanks to @rdspring1, not sure I would have figured this out on my own!

tfogal:
!build

csrc/executor.cpp
```diff
@@ -1868,14 +1999,8 @@ std::vector<at::Tensor> FusionExecutor::runFusion(
...
   }
 
-  std::vector<std::vector<std::byte>> arg_buffers;
-  {
-    FUSER_PERF_SCOPE("ExecutorRunFusion::GetArgsBuffers");
-    arg_buffers.reserve(kernel()->parameters().size());
-    for (auto v : kernel()->parameters()) {
-      arg_buffers.emplace_back(
-          getKernelArgument(expr_eval, v, kernel()->indexType()));
-    }
```
tfogal, line 1878: As discussed interactively, this was created (allocated) and dropped (deallocated) every time we `runFusion`, before. Part of the speedup here is not doing so many dynamic allocations to create the same array.

csrc/executor.cpp
```diff
@@ -1905,11 +2030,7 @@ std::vector<at::Tensor> FusionExecutor::runFusion(
 
     ensureAvailableDynamicSmemSize(executor_entry->launch_params.smem());
 
-    std::vector<void*> arg_buffer_ptrs;
-    arg_buffer_ptrs.reserve(arg_buffers.size());
-    for (auto& arg_buffer : arg_buffers) {
-      arg_buffer_ptrs.push_back(arg_buffer.data());
-    }
```
tfogal, line 1912: These are now stored in the `ExecutorEntry`, as `arg_ptrs`

csrc/executor.cpp
```diff
@@ -1606,6 +1607,137 @@ void FusionExecutor::initializeExecutorEntry(
...
+      // in the base address, shape, and stride arrays to cover whatever new
+      // tensors we got this round.
+      TensorView* mtv = dynamic_cast<TensorView*>(params[p]);
+      const Val* mdexpr = IrBuilder::metadataExpr(mtv);
+      const PolymorphicValue& tmd = expr_eval.evaluate(mdexpr);
+      const size_t idx_type_size =
+          PrimDataType::Int == idx_type ? sizeof(int64_t) : sizeof(int32_t);
+      fill_gpu_ptr(entry, tmd, p, idx_type_size);
+    } else {
+      entry.args[p] = getKernelArgument(expr_eval, params[p], idx_type);
```
tfogal, line 1735: We could maybe do better for the non-tensor arguments, too. `getKernelArgument` is what is slow, due to dynamic allocations.

Such arguments appear to be somewhat rare, though, and I think the scope of this patch is already larger than I'd like it to be.

csrc/executor.cpp
```diff
@@ -1606,6 +1607,137 @@ void FusionExecutor::initializeExecutorEntry(
...
+      // GPU tensors are not passed directly: instead we pass a Tensor<type,
+      // rank, rank> struct. The pointer and dimensions are dynamic, but the
+      // types and ranks are actually static (changing the rank or the types
+      // would need to be done via a new FusionDefinition). As such, we created
+      // the Tensor<t, r, r> struct during ::computeArgs, and here we just fill
+      // in the base address, shape, and stride arrays to cover whatever new
+      // tensors we got this round.
+      TensorView* mtv = dynamic_cast<TensorView*>(params[p]);
+      const Val* mdexpr = IrBuilder::metadataExpr(mtv);
+      const PolymorphicValue& tmd = expr_eval.evaluate(mdexpr);
```
tfogal, line 1730: In the future, we might consider trying to get these 3 lines off the critical path.

In this path, we know we have a tensor and we know we need the 3 fields for that tensor: data, shape, and strides. So ideally we'd just build a:

```c++
struct whatever_t
  void* data;
  IdxType* shape;
  IdxType* strides;
};
```

But we have to go all the way through `PolymorphicValue` and `TensorMetaData` structs, and construct a temporary `Val` here just to get those 3 fields. It would not only be faster to go directly to that struct, it would also be simpler / easier to understand. Maybe a follow-on patch...

csrc/executor.cpp
```diff
@@ -1868,14 +1999,8 @@ std::vector<at::Tensor> FusionExecutor::runFusion(
...
 
-  std::vector<std::vector<std::byte>> arg_buffers;
-  {
-    FUSER_PERF_SCOPE("ExecutorRunFusion::GetArgsBuffers");
-    arg_buffers.reserve(kernel()->parameters().size());
-    for (auto v : kernel()->parameters()) {
-      arg_buffers.emplace_back(
-          getKernelArgument(expr_eval, v, kernel()->indexType()));
-    }
+  if (executor_entry->args.empty()) {
```
tfogal, line 2002: @rdspring1 note that I didn't change the deserialization path; instead I modified this line to just ask if the relevant field was initialized (before, when hitting the issue we discussed, I had kept track of whether `initializeExecutorEntry` was called and use that as the conditional).

Something deep in the `ExpressionEvaluator` was not initialized. It was probably that, at the time of `::deserialize`, we don't have access to the `KernelArgumentHolder` and so I couldn't `ExpressionEvaluator::bind` the inputs. Or maybe it was the `FusionExecutor::kernel()` itself wasn't initialized?

I didn't spend too much time trying to figure it out; it doesn't seem like a terrible thing that the first launch after deserialization is slower than subsequent launches. This is the same for non-deserialization cases anyway.

Of potential note is that even in the normal/non-deserialized case, the `computeArgs` call needs to be this late. Something is getting initialized before this that is critical for us to identify the argument information.

csrc/executor.cpp
naoyam: @zasdfgbnm for viz

csrc/executor.cpp
rdspring1: Serialization associates a metadata `KernelArgumentHolder` for each `FusionKernelRuntime`. We don't have full arguments available during deserialization.

rdspring1: IIUC, you could fill the `ExpressionEvaluator` with the metadata `KernelArgumentHolder` and run `computeArgs`.

csrc/executor.cpp
```diff
@@ -965,7 +966,7 @@ at::Tensor allocateOutput(
     const FusionExecutor::GlobalBufferInfo& out_info,
     const AliasInfo& alias_info,
     const c10::Device& device,
-    ExpressionEvaluator& ee) {
+    const ExpressionEvaluator& ee) {
```
wujingyue, line 969: The evaluate method on a `const ExpressionEvaluator&` is different from that on a `ExpressionEvaluator&`. See https://github.com/NVIDIA/Fuser/blob/8ea9064c147756391d554d1bb736618ecfd9b21f/csrc/expr_evaluator.h#L62 and https://github.com/NVIDIA/Fuser/blob/8ea9064c147756391d554d1bb736618ecfd9b21f/csrc/expr_evaluator.h#L54

IIUC from @zasdfgbnm the former doesn't cache the computed polymorphicvalue (the final one or the intermediate ones), so it can be slower. @zasdfgbnm , what's your opinion on this?

zasdfgbnm: Jingyue is right that the `const` version of this method does not cache intermediate or final result. But is caching important here for this specific function? My understanding is, this function only uses EE for ATen evaluating the fusion outputs marked as alias, and these tensors will only be used to return to the user but nothing else?

csrc/executor.cpp
```diff
@@ -1606,6 +1607,137 @@ void FusionExecutor::initializeExecutorEntry(
...
+/// @param idx the index into entry.args and related parallel arrays in the
+///            entry.
+/// @param idx_type_size generally sizeof(int32_t) or sizeof(int64_t); used for
+///                      computing how large the arrays to copy are.
+static void fill_gpu_ptr(
+    FusionExecutor::ExecutorEntry& entry,
+    const PolymorphicValue& tmd,
+    size_t idx,
+    size_t idx_type_size) {
+  void* data = tmd->*&TensorMetaData::data;
```
wujingyue, line 1632: Maybe a question to @zasdfgbnm: any chance to simplify the API? It seems that something like `pv.as<TensorMetaData>().data` would be more straightforward to many C++ engineers. What's the reason behind the current API?

csrc/executor.cpp
```diff
@@ -1606,6 +1607,137 @@ void FusionExecutor::initializeExecutorEntry(
...
+      TMD_ARRAY_REF(tmd, TensorMetaData::logical_size);
+  const c10::IntArrayRef& strides =
+      TMD_ARRAY_REF(tmd, TensorMetaData::alloc_stride);
+#undef TMD_ARRAY_REF
+
+  // These are the three offsets we need to copy into.
+  std::array<std::byte*, 3> offsets = {
+      entry.args[idx].data(), // data ptr
+      entry.args[idx].data() + sizeof(void*), // shape array
+      entry.args[idx].data() + sizeof(void*) + shape.size() * idx_type_size,
```
wujingyue, line 1652:
```suggestion
      entry.args[idx].data() + sizeof(void*) + shape.size() * idx_type_size,  // strides
```

wujingyue (approved):

csrc/executor.cpp
zasdfgbnm: https://github.com/NVIDIA/Fuser/blob/1e4f744c366f3637c02fda8a197bfe8b38ae77c2/csrc/ir/container.cpp#L318-L325
FYI: tensor metadata is not a temporary Val, it is cached. But still, I agree that skipping these three lines can be faster if the only purpose of getting the tensor metadata is to fill arguments. For other cases, for example when TMA is used, tensor metadata will be evaluated anyway because it is used to create the TMA descriptor.

csrc/executor.cpp
wujingyue: https://github.com/NVIDIA/Fuser/blob/8ea9064c147756391d554d1bb736618ecfd9b21f/csrc/executor.cpp#L1050-L1062 can grow into a pathological case where we can end up evaluating `n` output tensors `O(n^2)` times. It's probably a rare case but kind of hard to diagnose when it shows up as a performance problem. Wdyt?

csrc/executor.cpp
zasdfgbnm: `pv.as<TensorMetaData>().data` requires `TensorMetaData` to be known at the definition of `PolymorphicValue`. Basically, you will need something like:

```C++
using PolymorphicValue = dynamic_type::DynamicType<
    dynamic_type::Containers<std::vector>,
    TensorMetaData,
    ....>;
```

`TensorMetaData` is not the only struct we support, and we do not want to manually specify every struct in the definition of `PolymorphicValue`, instead we want to be able to define an arbitrary struct and let `PolymorphicValue` hold it.

csrc/executor.cpp
zasdfgbnm: That is a good point! Due to this, I would suggest keeping it as a non-constant reference.

csrc/executor.cpp
tfogal: I think I'd need full args. I tried to bind kernel()->inputs(), but that died somewhere deep with an assert that something was a std::monostate instead of a real object.

csrc/executor.cpp
tfogal: Ack, thanks guys. I will change it back to a mutable version.

csrc/executor.cpp
```diff
@@ -1606,6 +1607,137 @@ void FusionExecutor::initializeExecutorEntry(
...
+/// Tensor<t,r,r> during ::computeArgs, and then in this function we just
+/// update that memory with the current values for the tensor's base address,
+/// shape, and strides.
+///
+/// @param entry the entry we have previously setup for this fusion
+/// @param idx the index into entry.args and related parallel arrays in the
+///            entry.
+/// @param idx_type_size generally sizeof(int32_t) or sizeof(int64_t); used for
+///                      computing how large the arrays to copy are.
+static void fill_gpu_ptr(
```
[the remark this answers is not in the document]
tfogal: Kernels can actually take CPU tensors as well, at least according to getKA and specifically polymorphicValueToBytes.

Granted the only CPU "tensors" supported are scalars, but nonetheless...

`TensorMetaData` is a real object in our system (actually something that's used inside this function), and so I hesitate to use that name because that's not what's happening here. Maybe `tensorArgMD`?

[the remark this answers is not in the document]
wujingyue: You mean something like fillTensorArgMetadata? Yeah, LGTM!

tfogal:
!build

tfogal:
I believe I have addressed all comments, and I just merged in main. This is ready for re-review.

Please let me know if I've missed anything. Relatedly, please use 'Resolve conversation' if you feel a discussion has concluded.

tfogal:
@parthmannan @mpatel31415 now that this is merged i'd be curious if there's impact on previously-host-latency-bound scenarios. If you're re-running any such things and can compare to older runs, I'd love to hear if this moves the needle.

PR #2137, opened 2024-04-24
Add examples in Figure 8 and Figure 9 of `tma.md` as unit tests

zasdfgbnm:
I do need to change the sizes in example (a), (c), (d) because when I wrote these examples, I forgot that smem addresses must be 128 byte aligned, and the original numbers does not satisfy this property.

TODO: add Figure 10 as tests as well


zasdfgbnm:
!build

tests/cpp/test_memory.cpp
```diff
@@ -1570,6 +1570,395 @@ TEST_F(TMACompileTimeInvalidTest, SwizzleBulkWithNonBulk) {
           "Unsupported expression between the allocation domain and the partitioned IterDomains:")));
 }
 
+// Tests for the examples in doc/dev/tma.md
+class TMADocTest : public TMATest {};
+
+TEST_F(TMADocTest, Figure8a) {
+  GTEST_SKIP() << "TODO: add check for this invalid case.";
```
zasdfgbnm, line 1577: Currently, there is no smem schedule check. So wrong schedule will lead to misaligned address error or silent wrong result at runtime.

Actually, validating the smem schedule is not easy. I don't have a good idea on how to do it yet.

naoyam (approved):
LGTM. Let's also add references from the doc to these unit tests.

zasdfgbnm:
> Let's also add references from the doc to these unit tests.

Sure, let me merge this as-is, and add the reference in my next PR (the one for Figure 10)

PR #2138, opened 2024-04-24
Add support for 0-dim bias argument in Linear

protonu:
Extending the linear API to support a 0-dim bias.

### DIFF
--- a/csrc/scheduler/mma_utils.cpp
+++ b/csrc/scheduler/mma_utils.cpp
@@ -1041,6 +1041,10 @@ inline void resolveTvToMatmulDomainsMapping(
     IterDomain* k,
     const ComputeAtMap& ca_map) {
   for (const auto tv : tensors) {
+    // This ensures all inputs are added to the deps_map.
+    // There could be inputs such as a bias with 0-dim.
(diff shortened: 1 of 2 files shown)


protonu:
!build

csrc/scheduler/mma_utils.cpp
```diff
@@ -1041,6 +1041,10 @@ inline void resolveTvToMatmulDomainsMapping(
     IterDomain* k,
     const ComputeAtMap& ca_map) {
   for (const auto tv : tensors) {
+    // This ensures all inputs are added to the deps_map.
+    // There could be inputs such as a bias with 0-dim.
+    // Which would otherwise skipped.
```
jacobhinkle, line 1046:
```suggestion
    // There could be inputs such as a zero-dimensional bias which
    // would otherwise be skipped.
```

jacobhinkle (approved):
LGTM assuming this doesn't break any tests.

protonu:
!build

Priya2698 (approved):
LGTM.
We could a CPP test as well but it is not necessary, the python tests alone is fine too.

PR #2139, opened 2024-04-24
Fixed torch profiler API changes in python benchmarks

xwang233:
Re: https://github.com/pytorch/pytorch/pull/123247

This PR changed `self_cuda_time_total` to `self_device_time_total` in API calls.

### DIFF
--- a/benchmarks/python/core.py
+++ b/benchmarks/python/core.py
@@ -237,7 +237,8 @@ class NVFBenchmark:
         elapsed_cuda_time = (
             sum(
                 [
-                    event.self_device_time_total
+                    # Re: torch profiler API changes in https://github.com/pytorch/pytorch/pull/123247
+                    (event.self_device_time_total if hasattr(event, 'self_device_time_total') else event.self_cuda_time_total)
                     for event in prof_averages
                     if event.device_type == DeviceType.CUDA
                 ]
--- a/benchmarks/python/core.py
+++ b/benchmarks/python/core.py
@@ -237,7 +237,7 @@ class NVFBenchmark:
         elapsed_cuda_time = (
             sum(
                 [
-                    event.self_cuda_time_total
+                    event.self_device_time_total
                     for event in prof_averages
                     if event.device_type == DeviceType.CUDA
                 ]
--- a/benchmarks/python/core.py
+++ b/benchmarks/python/core.py
@@ -238,7 +238,11 @@ class NVFBenchmark:
             sum(
                 [
                     # Re: torch profiler API changes in https://github.com/pytorch/pytorch/pull/123247
-                    (event.self_device_time_total if hasattr(event, 'self_device_time_total') else event.self_cuda_time_total)
+                    (
+                        event.self_device_time_total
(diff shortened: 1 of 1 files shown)


xwang233:
!build --pybench

Priya2698 (approved):
Thanks for updating!
Do we need to have different API based on torch version? In this PR: https://github.com/pytorch/pytorch/pull/123247, I don't see any version update though, so this should be sufficient.

benchmarks/python/core.py
```diff
@@ -237,7 +237,7 @@ def _get_kernel_time(
         elapsed_cuda_time = (
             sum(
                 [
-                    event.self_cuda_time_total
+                    event.self_device_time_total
```
jjsjann123, line 240: I think you'll need to guard this with `NVF_TORCH_VERSION_GREATER`. Otherwise we won't be able to build with stable torch release

jjsjann123: wait... this is python. wasn't paying attention to that.

But how did it work through CI? we did test against stable torch version right. Is benchmark not part of CI on older torch version?

xwang233: For torch stable build & test in CI, we run python tests but not python benchmarks.

Priya2698: Should we add the check for torch version then?

jjsjann123: I think we should. Otherwise we won't be able to run benchmark against old torch version.

xwang233: `torch==2.3` is still using `self_cuda_time_total`. Will add a version guard for torch 2.4.

xwang233:
!build --pybench

Priya2698:
> !build --pybench

Is this running benchmarking? Otherwise this run won't really check this API usage

xwang233:
> > !build --pybench
> 
> Is this running benchmarking? Otherwise this run won't really check this API usage

It runs pytest with `--disable-benchmarking`

xwang233:
!build

Priya2698:
> > > !build --pybench
> > 
> > 
> > Is this running benchmarking? Otherwise this run won't really check this API usage
> 
> It runs pytest with `--disable-benchmarking`

This API change will be exercised only if we enable benchmarking and not during validation which is the case on the CI. Generally, we don't need to benchmark on the CI but PRs like these cannot be tested since the torch.profiler will not be used.

Priya2698 (approved):
This should avoid API failures for now, although seems like this PR was closed: https://github.com/pytorch/pytorch/pull/123247. We can revisit and add torch version guards instead of attribute checks once the torch API is final.

jjsjann123:
> This should avoid API failures for now, although seems like this PR was closed: [pytorch/pytorch#123247](https://github.com/pytorch/pytorch/pull/123247). We can revisit and add torch version guards instead of attribute checks once the torch API is final.

Just to clarify that we shouldn't need to revisit this unless upstream renames the API to something else.

That upstream PR is merged and Xiao's patch looks pretty solid to me. There's no need to version guard anything, since attribute query should work just fine.

PR #2140, opened 2024-04-24
[matmul] Merge batch dims with BIDx tile dim

### LINKED ISSUE #1726: Split-K bmm serializes entire kernel
For a batch matmul (bmm), we have a kernel that looks like this:
```
FOR n in batch
  FOR ... // main loop n
    ...
  ENDFOR // main loop n
  FOR elt // epilogue/store result n
    ...
  ENDFOR // epilogue/store result n
ENDFOR // batch
```
Without spit-K, we parallelize the outer loop with `BIDz`, but for split-K, we use `BIDz` for the split-K dimension. In that case, the outer loop (batch dimension) is no longer trivial. This means we wind up with this situation for split-K bmm:
```
blockSerializeWait()
FOR n in batch
  FOR ... // main loop n
    ...
  ENDFOR // main loop n
  FOR elt // epilogue/store result n
    ...
    serialGridReductionStep(...)
    ...
  ENDFOR // epilogue/store result n
ENDFOR // batch
blockSerializeRelease()
```
The serialization means that each block in the segment will wait for all previous blocks to run to completion before beginning, which is clearly inefficient. Instead, we would like to have this:
```
FOR n in batch
  FOR ... // main loop n
    ...
  ENDFOR // main loop n
  blockSerializeWait()
  FOR elt // epilogue/store result n
    ...
    serialGridReductionStep(...)
    ...
  ENDFOR // epilogue/store result n
  blockSerializeRelease()
ENDFOR // batch
```
This way, we allow each main loop to be parallelized across blocks while still serializing the split-K reduction.

### Approach

Currently we simply search for serial grid reductions and place the syncs around the outer loop. Instead, we can let scheduling determine where to sync. To do this, instead of `ParallelType::Serial` we can use `ParallelType::Group` for the inner loops that we want to serialize (also allowing `ParalleliType::Vectorize`). The matmul scheduler can then simply leave the batch dimension alone to avoid serializing it.

We should take care to ensure that serial grid reduction is "persistent": the syncs should always reset the semaphore to zero, and the first block will actually wait on the second batch iteration, etc.

### LINKED ISSUE #1899: Re-use smem in batch split-K matmuls
When we have batch dimensions and we use split-k, we generate a kernel like this:
FOR n in batch
  FOR ... // main loop
    ...
  ENDFOR // main loop
  FOR // store to gmem
    ...
  ENDFOR // store to gmem
ENDFOR // batch
```
If we also use smem epilogue unswizzling, this becomes:
```
FOR n in batch
  FOR ... // main loop
    ...
  ENDFOR // main loop
  FOR // unswizzle
    ...
  ENDFOR // unswizzle
  FOR // store to gmem
    ...
  ENDFOR // store to gmem
ENDFOR // batch
```
To conserve smem and increase occupancy, we ordinarily would try to re-use the smem used to stage the operands A and B for the unswizzling loop. However, we currently only support re-using smem in this way if the _outer_ lifetimes do not overlap. Since the batch loop is present in these cases, the outer lifetimes of all smem buffers is the entire kernel.

Related to #1726 since that involves the loop structure of batch split-K.

# Proposed Approach

We should analyze outer smem aliasing by finding the innermost scope containing all smem buffers in the fusion. Then we can perform our usual outer aliasing lifetime analysis and if there are any outer we should ensure a sync exists before the outer `ENDFOR` position. This way we can insert a block sync at the end of the batch loop, enabling us to "outer alias" the operand smem with the epilogue smem without indexing collisions.

jacobhinkle:
Previously, we parallelized batch dimension as BIDz except when split-K was enabled in which case we looped over the batch dim.That is suboptimal for split-K, see #1726 and #1899. This PR removes those loops in all matmul kernels.

Instead of looping over batch dimensions, this PR merges batch dims with either the Mo or No dimensions (the outer tile dimensions), depending on cta order. We choose the dim to merge with that is parallelized as BIDx, so that we can support larger batch dims (BIDx is limited to 2^31-1 while other dims are limited to 2^16-1).

Note that this changes the CTA order for non-splitK bmms, which were previously parallelized BIDz and are now an outer dim merged into BIDx (i.e. inserted between BIDx and BIDy). If that's not ideal we could potentially merge Mo and No and parallelize those as BIDx, then leave BIDy and BIDz for split-K and batch dim, respectively.

Fixes #1726. Fixes #1899

### DIFF
--- a/csrc/scheduler/matmul.cpp
+++ b/csrc/scheduler/matmul.cpp
@@ -923,11 +923,11 @@ void scheduleMatmul(Fusion* fusion, const MatmulParams& params) {
       mma_result->nDims() == 3 || mma_result->nDims() == 4,
       "Currently, we only support B, M, N and K being a single dimension.",
       " More general tensor contraction is not supported yet.");
-  const int64_t num_batch_dims = mma_result->nDims() - 3;
+  int64_t num_batch_dims = mma_result->nDims() - 3;
 
   // [... M,N,K]
   mma_utils::makeTile(mma_result, gemm_tile.cta_tile.toVector());
-  // [..., Mo, No, Ko, Mi, Ni, Ki]
+  // [BMo, No, Ko, Mi, Ni, Ki]
 
   // Unswizzle mma result in shared memory
   // Note that if we are using split-K, we will set up this buffer after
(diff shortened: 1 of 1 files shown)


jacobhinkle:
!build --diff

jacobhinkle:
!build --matmul-bench

jacobhinkle:
!build --matmul-bench

jacobhinkle:
!build --matmul-bench

jacobhinkle:
!build --matmul-bench

PR #2147, opened 2024-04-24
Bail out when the catted dimension is an outer-split factor.

### LINKED ISSUE #2142: Inconsistent producer and consumer in MoveSplitCatTest.MultiplePairs
With the current main branch, the `MoveSplitCatTest.MultiplePairs` test produces four segments, and in the last segment, there are these three tensors:

```
T23_l[ iblockIdx.x123{( ceilDiv(( ceilDiv(( ceilDiv(( 6 * ( ceilDiv(( 6 * 4 ), 6) ) ), 128) ), 1) ), 1) )}, iUS124{1}, iS122{1}, ithreadIdx.x120{128} ] ca_pos( 4 ) produce
_pos( 4 ) = view( T14_l[ iblockIdx.x130{( ceilDiv(( ceilDiv(( ceilDiv(( 6 * ( ceilDiv(( 6 * 4 ), 6) ) ), 128) ), 1) ), 1) )}, iUS131{1}, iS129{1}, ithreadIdx.x127{128} ] ca_pos( 4 ) produce_pos( 4 ) )
T24_l[ iblockIdx.x116{( ceilDiv(( ceilDiv(( ceilDiv(( 12 * 2 ), 128) ), 1) ), 1) )}, iUS117{1}, iS115{1}, ithreadIdx.x113{128} ] ca_pos( 4 ) produce_pos( 4 )
   = Set( T23_l[ iblockIdx.x123{( ceilDiv(( ceilDiv(( ceilDiv(( 6 * ( ceilDiv(( 6 * 4 ), 6) ) ), 128) ), 1) ), 1) )}, iUS124{1}, iS122{1}, ithreadIdx.x120{128} ] ca_pos( 4 ) produce_pos( 4 ), cache_op=Streaming )
T21_g[ iblockIdx.x109{( ceilDiv(( ceilDiv(( ceilDiv(( 12 * 2 ), 128) ), 1) ), 1) )}, iUS110{1}, iS108{1}, ithreadIdx.x106{128} ] ca_pos( 2 ) produce_pos( 4 )
   = Set( T24_l[ iblockIdx.x116{( ceilDiv(( ceilDiv(( ceilDiv(( 12 * 2 ), 128) ), 1) ), 1) )}, iUS117{1}, iS115{1}, ithreadIdx.x113{128} ] ca_pos( 4 ) produce_pos( 4 ), cache_op=Streaming )
```

`T14` is reshaped to `T23`, which is then copied to `T24`, and then into T21, which is a global-memory output. So, the rfactor domain of `T23` should match with the root domain of `T24` (and also that of `T21`). However, that doesn't seem to be the case.

```
T23_l[ iblockIdx.x123{( ceilDiv(( ceilDiv(( ceilDiv(( 6 * ( ceilDiv(( 6 * 4 ), 6) ) ), 128) ), 1) ), 1) )}, iUS124{1}, iS122{1}, ithreadIdx.x120{128} ] ca_pos( 4 ) produce_pos( 4 )
 root domain : (iS78{6}rf, iS79{4}rf)
  Merge: iS78{6}rf and iS79{4}rf -> iS80{( 6 * 4 )}rf
  Outer split: iS80{( 6 * 4 )}rf by factor 6 -> iS81{6}rf, iS82{( ceilDiv(( 6 * 4 ), 6) )}rf, start offset: 0, stop offset: 0
 rfactor domain : (iS81{6}rf, iS82{( ceilDiv(( 6 * 4 ), 6) )}rf)
```
```
T24_l[ iblockIdx.x116{( ceilDiv(( ceilDiv(( ceilDiv(( 12 * 2 ), 128) ), 1) ), 1) )}, iUS117{1}, iS115{1}, ithreadIdx.x113{128} ] ca_pos( 4 ) produce_pos( 4 )
 root domain : (iS74{12}, iS75{2})
```

The rfactor shape of `T23` is `[6, 4]`, whereas the root shape of `T24` is `[12, 2]`. 

The test does pass, but I assume it's just luck.

wujingyue:
Fix #2142.

### DIFF
--- a/csrc/preseg_passes/move_split_cat.cpp
+++ b/csrc/preseg_passes/move_split_cat.cpp
@@ -99,16 +99,30 @@ class CancelSplitCat {
   // Returns 0 because `slice`'s dimension 0 is the outer dimension.
 
   // Example 4:
-  //   t = reshape(slice, {6}, {2, 3})
+  //   t = reshape(slice, {6}, {2, 3})  // inner split by 3
   //   out = cat({t, ...}, 0}
   //
-  // Returns 0 because `out`'s dimension 0 is the outer dimension.
+  // Returns 0 because `t` is an inner split and `out`'s dimension 0 is the
+  // outer dimension. See #2142 for why checking the split is inner/outer.
   //
   // Example 5:
+  //   t = reshape(slice, {6}, {2, 3})  // outer split by 2
+  //   out = cat({t, ...}, 0}
+  //
+  // Returns -1 because `t` is an outer split. See #2142 for why checking the
+  // split is inner/outer.
+  //
+  // Example 6:
   //   t = reshape(slice, {6}, {2, 3})
   //   out = cat({t, ...}, 1}
   //
-  // Returns -1 because `out`'s dimension 1 is the inner dimension.
+  // Returns -1 because `out`'s dimension 1 is the inner dimension. Consider
+  //   in = arange(12)  # [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11]
+  //   x, y = in.split([6, 6]) # [0, 1, 2, 3, 4, 5], [6, 7, 8, 9, 10, 11]
+  //   x = x.view([2, 3])  # [[0, 1, 2], [3, 4, 5]]
+  //   y = y.view([2, 3])  # [[6, 7, 8], [9, 10, 11]]
+  //   out = cat([x, y], axis=1)  # [[0, 1, 2, 6, 7, 8], [3, 4, 5, 9, 10, 11]]
+  // It's impossible to set strides to make `out` a view of `in`.
   int64_t computeCatAxisAfterZipping(
       const std::vector<IterDomain*>& slice_rfactor,
       IterDomain* cat_id);
@@ -307,7 +321,13 @@ int64_t CancelSplitCat::computeCatAxisAfterZipping(
       Expr* def = defining_group->front();
 
       if (Split* split = dynamic_cast<Split*>(def)) {
-        if (exact_graph.toGroup(split->outer()) == cat_group) {
+        // Check `split` is an inner split to avoid #2142. If we allow an outer
+        // split, `replayExprWithNewInputs` will incorrectly use the same split
+        // factor as the extent of the horizontally merged dimension. We could
+        // instead make `replayExprWithNewInputs` smarter, but I haven't given
+        // that alternative enough thought.
+        if (exact_graph.toGroup(split->outer()) == cat_group &&
+            split->innerSplit()) {
           return exact_graph.toGroup(split->in());
         }
         return nullptr;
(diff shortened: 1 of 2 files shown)


wujingyue:
!build

naoyam:
IIRC, all splits in reshape transformations are outer. Does this mean we can't have a reshape with a split?

jacobhinkle:
Could you instead convert them to inner splits if the inner dimension is const scalar?

wujingyue:
> IIRC, all splits in reshape transformations are outer. Does this mean we can't have a reshape with a split?

Yes. However, are you sure all splits are outer even for [dynamic reshapes](https://github.com/NVIDIA/Fuser/blob/72bc243dfa5e4a7130d80e04d76fed8b4cf5f842/csrc/ops/alias.h#L42)? Can't nvFuser reshape a variable-size vector to a matrix of a fixed width?

naoyam:
I think dynamic reshapes just do the same reshape analysis at the run time, so I think all splits are output. It is possible to try to use inner splits, though.

naoyam:
I also am not sure if outer splits are the root cause. It seems to me we would need to update the input and output shapes of the reshape since the reshape now deals with the concatenated tensor, whereas the original reshape is only with one slice of the tensor.

wujingyue:
> It seems to me we would need to update the input and output shapes of the reshape since the reshape now deals with the concatenated tensor, whereas the original reshape is only with one slice of the tensor.

Yes, and that's done by replayExprWithNewInput. It replays the same **transforms** (therefore the same split factor) of the reshape on a **different** tensor shape. I think doing so is safe for reshapes because (1) reshapes (i.e. ViewOps) don't change the number of elements so splits are even, and (2) the distribution rule applies. For example, `s0*f + s1*f = (s0+s1)*f` for merge and `s0/f + s1/f = (s0+s1)/f` for inner split while the catted dimension is the outer dimension. (In fact, the optimization doesn't kick in at all when the catted dimension is the inner dimension of the split but for a different reason). 

The outer split is problematic for the replay when the catted dimension is also the outer dimension, because the split factor would have to change. E.g. 

```
6 =(outer split by 2)=> 2, 3
12 =(outer split by 4)=> 4, 3
```

These two reshapes could still be horizontally merged but the new split factor would have to be 6. Or as @jacobhinkle mentioned we could change the reshapes to do inner split by 3.

wujingyue:
> If reshape only did inner splits then it would simplify this a lot wouldn't it? 

Yes -- that would simplify **this case** a lot. I'll defer that decision to you and @naoyam . If you think that's the right way, there's also the sub-decision to make reshapes do inner splits across the board (e.g. done by concretization) or case by case (e.g. done by passes like move_split_cat where it knows inner split is better).

naoyam:
> > If reshape only did inner splits then it would simplify this a lot wouldn't it?
> 
> Yes -- that would simplify **this case** a lot. I'll defer that decision to you and @naoyam . If you think that's the right way, there's also the sub-decision to make reshapes do inner splits across the board (e.g. done by concretization) or case by case (e.g. done by passes like move_split_cat where it knows inner split is better).

I don't think there's any specific reason to do inner or outer. @rdspring1, I think you did the original reshape analysis. Do you remember if there's any specific reason to prefer one over the other?

naoyam (approved):
LGTM

rdspring1:
The reshape analysis moves from left to right, so we create new dimensions from outer to inner.
Also, our `TensorDomains` are represented by `std::vector`, so it seems natural to create new `iterDomains` in this order.
It's been a while, so I don't remember a stronger reason than this.

PR #2160, opened 2024-04-29
print out tv domains instead of pointer address

liqiangxl:
Before: `Un-mapped multi-reduction: 0x7f99ee928700 0x7f99ee870d00`
After: `Un-mapped multi-reduction: T43_g[ rS122{56}, rS286{128}, iS124{1024} ] T55_g[ iS154{56}, iS155{128}, rS156{1024} ]`


jjsjann123 (approved):
Thanks for the correction. :bow:

PR #2162, opened 2024-04-30
Reduce number of inputs

Priya2698:
Adding `NVFUSER_DISABLE=kernel_reuse` increases the resource usage by 3x. Reducing the number of inputs (by 4x)  to avoid timeout on CI.


xwang233:
!build --pybench

xwang233:
!build --pybench

benchmarks/python/global_params.py
```diff
@@ -73,7 +73,7 @@ def generate_input_sizes(dims: Union[int, List] = 2) -> List[Tuple]:
         if dim == 2:
             input_ranges = []
 
-            step_size = 256
+            step_size = 1024
```
naoyam, line 76: IIUC, this step size means the problem sizes would be `x + 1024 * i`, where `x` is the minimum size and `i` is increased from 0 until `x + 1024 * i` reaches the maximum size. Is that correct?

Another way of generating problem sizes would be `x + some_factor ^ i`, e.g., `2, 4, 8, 16, 32, ...`.

Priya2698:
Closing as not planned.

PR #2163, opened 2024-04-30
Don't reuse segmented kernel runtimes if we can avoid segmentation

jacobhinkle:
By default, `FusionExecutorCache` uses the following policy for re-using compiled fusions, given a set of input shapes:
1. If we have encountered these input shapes previously, we use the same compiled fusion that was used last time we were passed those shapes.
2. If these new input shapes lead to a new concretization, we must compile a new `FusionKernelRuntime`, which means segmenting the fusion.
3. We check all previously compiled `FusionKernelRuntime`s for the given concretization. For each of those runtimes we look at each segment and compute the heuristic for that segment given the new input shapes. If the heuristic changed (not just the launch params) then we cannot reuse the runtime.

In #563 we added an option to disable kernel reuse in case 3 above. This is because we encountered a fusion and a set of inputs that _could_ have been scheduled unsegmented, but for which there existed a previously-compiled fusion that was segmented but that we reused because it returned the same heuristics for each segment for the new inputs leading to a suboptimal fusion.

This PR adds the failing case above as a test. It also modifies the `FusionExecutorCache` such that we add an additional condition:
4. If we are able to compile a fusion without segmentation, we always refuse to re-use a kernel that is segmented.

This PR addresses the original motivating case for #563. However, note that it is not sufficient to guarantee that `FusionExecutorCache` will always compile optimal kernels regardless of execution order. For example, if the test case above is just one part of a larger fusion that always needs to be segmented, we might still re-use a suboptimal kernel. A more complete solution would involve the following:
1. During segmentation, we would record the sequence of merges performed. In particular, we only need to save the failed merges, and only those which failed because of runtime information (i.e. `canScheduleCompileTime` need not be repeatedly checked).
2. Before re-use, we should check that all of those failed merges still fail. If any do not, then that is an opportunity for increased fusion, so we should not reuse the segmented fusion.

### DIFF
--- a/csrc/kernel_cache.cpp
+++ b/csrc/kernel_cache.cpp
@@ -779,10 +779,22 @@ FusionKernelRuntime* FusionExecutorCache::getKernelRuntimeFor(
   // that whenever we encounter a new set of input shapes we segment and compile
   // a new FusionKernelRuntime.
   if (!isOptionDisabled(DisableOption::KernelReuse)) {
+    SchedulerRuntimeInfo runtime_info(
+        fusion(),
+        args,
+        /*precomputed_values=*/nullptr,
+        /*all_tvs=*/{},
+        forced_index_type);
+    bool should_not_be_segmented =
+        SchedulerEntry::proposeHeuristics(fusion(), runtime_info).has_value();
     auto reuse_it = std::find_if(
         kernel_runtimes.begin(),
         kernel_runtimes.end(),
-        [&args, &new_heuristics, &forced_index_type](auto& kernel_runtime) {
+        [should_not_be_segmented, &args, &new_heuristics, &forced_index_type](
+            auto& kernel_runtime) {
+          if (should_not_be_segmented && kernel_runtime->isSegmented()) {
+            return false;
+          }
           auto maybe_heuristics =
               kernel_runtime->getMaybeHeuristicsFor(args, forced_index_type);
           if (!maybe_heuristics.has_value()) {
--- a/tests/cpp/test_gpu_outer_reduction.cpp
+++ b/tests/cpp/test_gpu_outer_reduction.cpp
@@ -2305,4 +2305,87 @@ TEST_F(OuterReductionTest, OuterReductionMagicScheduler) {
   }
 }
 
+// Test that executing with inputs that require segmentation does not cause a
+// second execution to be segmented if it can avoid it.
+TEST_F(OuterReductionTest, KernelReuse) {
+  // DisableOption::KernelReuse would cause this test to pass even if it
+  // shouldn't, so we disable it here.
+  DisableOptionsGuard og;
+  DisableOptionsGuard::getCurOptions().unset(DisableOption::KernelReuse);
+  ASSERT_FALSE(isOptionDisabled(DisableOption::KernelReuse));
+
+  DataType dtype = DataType::Half;
+  int64_t reduction_axis = 0;
+
+  std::unique_ptr<Fusion> fusion_ptr = std::make_unique<Fusion>();
+  Fusion* fusion = fusion_ptr.get();
+  FusionGuard fg(fusion);
+
+  // setup fusion
+  auto grad_output = makeContigTensor(2, dtype);
+  auto output = makeContigTensor(2, dtype);
+  auto input = makeContigTensor(2, dtype);
+  fusion->addInput(grad_output);
+  fusion->addInput(output);
+  fusion->addInput(input);
+
+  if (dtype == DataType::Half) {
+    grad_output = castOp(DataType::Float, grad_output);
+    output = castOp(DataType::Float, output);
+    input = castOp(DataType::Float, input);
+  }
+
+  auto grad_input = softmax_backward(grad_output, output, reduction_axis);
+
+  if (dtype == DataType::Half) {
+    grad_input = castOp(DataType::Half, grad_input);
+  }
+
+  fusion->addOutput(grad_input);
+
+  auto options =
+      at::TensorOptions().dtype(data_type_to_aten(dtype)).device(at::kCUDA, 0);
+
+  FusionExecutorCache executor_cache(std::move(fusion_ptr));
+
+  auto run = [reduction_axis, &options, &executor_cache](
+                 int64_t reduction_size, int64_t iter_size) {
+    at::Tensor input =
+        (reduction_axis ? at::randn({iter_size, reduction_size}, options)
+                        : at::randn({reduction_size, iter_size}, options));
+
+    at::Tensor grad_output =
+        (reduction_axis ? at::randn({iter_size, reduction_size}, options)
+                        : at::randn({reduction_size, iter_size}, options));
+
+    at::Tensor output =
+        (reduction_axis ? at::randn({iter_size, reduction_size}, options)
+                        : at::randn({reduction_size, iter_size}, options));
+
+    std::vector<c10::IValue> aten_inputs({grad_output, output, input});
+
+    auto cg_outputs = executor_cache.runFusionWithInputs(aten_inputs);
+  };
+
+  {
+    int64_t reduction_size = 2048;
+    int64_t iter_size = 16384;
+    run(reduction_size, iter_size);
+    auto runtime = executor_cache.getMostRecentKernelRuntime();
+    ASSERT_NE(runtime, nullptr);
+    EXPECT_TRUE(runtime->isSegmented());
+    EXPECT_EQ(executor_cache.getKernelRuntimes().begin()->second.size(), 1);
+  }
+
+  {
+    int64_t reduction_size = 512;
+    int64_t iter_size = 16384;
+    run(reduction_size, iter_size);
+    auto runtime = executor_cache.getMostRecentKernelRuntime();
+    ASSERT_NE(runtime, nullptr);
+    EXPECT_FALSE(runtime->isSegmented());
+    EXPECT_EQ(executor_cache.getKernelRuntimes().begin()->second.size(), 2);
+  }
+}
+
 } // namespace nvfuser


jacobhinkle:
!build --diff --diff-bench

jacobhinkle:
!build --diff --diff-bench

xwang233:
!build --pybench

csrc/kernel_cache.cpp
```diff
@@ -779,10 +779,31 @@ FusionKernelRuntime* FusionExecutorCache::getKernelRuntimeFor(
...
+            // whether we can schedule the complete fusion without segmenting.
+            std::unique_ptr<Fusion> conc_fusion = std::make_unique<Fusion>(
+                *(kernel_runtime->fusionSegments()->completeFusion()));
+            SchedulerRuntimeInfo runtime_info(
+                conc_fusion.get(),
+                args,
+                /*precomputed_values=*/nullptr,
+                /*all_tvs=*/{},
+                forced_index_type);
+            should_not_be_segmented = SchedulerEntry::proposeHeuristics(
```
liqiangxl, line 799: `SchedulerEntry::proposeHeuristics` returns a `std::optional<ScheduleHeuristic>` , can we assign it to `std::optional<bool>`?

csrc/kernel_cache.cpp
```diff
@@ -779,10 +779,31 @@ FusionKernelRuntime* FusionExecutorCache::getKernelRuntimeFor(
...
+                                          conc_fusion.get(), runtime_info)
+                                          .has_value();
+          }
+          if (should_not_be_segmented.value() &&
+              kernel_runtime->isSegmented()) {
+            return false;
+          }
           auto maybe_heuristics =
```
liqiangxl, line 807: After found a heuristic for this complete fusion, can we directly use it?

csrc/kernel_cache.cpp
```diff
@@ -779,10 +779,31 @@ FusionKernelRuntime* FusionExecutorCache::getKernelRuntimeFor(
   // that whenever we encounter a new set of input shapes we segment and compile
   // a new FusionKernelRuntime.
   if (!isOptionDisabled(DisableOption::KernelReuse)) {
+    std::optional<bool> should_not_be_segmented = std::nullopt;
     auto reuse_it = std::find_if(
         kernel_runtimes.begin(),
         kernel_runtimes.end(),
-        [&args, &new_heuristics, &forced_index_type](auto& kernel_runtime) {
+        [&should_not_be_segmented, &args, &new_heuristics, &forced_index_type](
```
liqiangxl, line 786: can you comment about the general logic here? The logic in my mind is (1) search for unsegmented runtime cache, (2) if doesn't exist try to schedule as unsegmented, (3) if can't, search for segmented runtime.

csrc/kernel_cache.cpp
jacobhinkle: Currently I'm just checking that it has a value, and the actual heuristic will be found again below when defining `maybe_heuristics` on an unsegmented `FusionKernelRuntime`.

csrc/kernel_cache.cpp
jacobhinkle: Added some comments to the code. This block is getting the first `FusionKernelRuntime` with the same concretization and for which we use the same heuristics in each segment. That is what happens, but it happens all at once here:
1. If `should_not_be_segmented` is not yet set, initialize it by re-using the concretized fusion in `kernel_runtime`. This is done to prevent needing to perform a redundant concretization before entering this loop.
2. If `should_not_be_segmented` is true, then filter out segmented `FusionKernelRuntime`s (this implements your (1) and (2).
3. If `should_not_be_segmented` is false, include this in the search. It will be returned as `*reuse_it` if all segment heuristics match.

jacobhinkle:
It looks like this has a big impact on perf and it's not always positive. Looking at the codediffs for the C++ normalization benchmarks, there are a lot of cases where this change causes a previously segmented kernel to become unsegmented. I would not have expected that since on `main` we are using an option guard to enforce `DisableOption::KernelReuse`. Also, even though kernels are fused it seems to often hurt performance?
image

image

image

image

I'll try and figure out what's happening locally.

jacobhinkle:
> It looks like this has a big impact on perf and it's not always positive. Looking at the codediffs for the C++ normalization benchmarks, there are a lot of cases where this change causes a previously segmented kernel to become unsegmented. I would not have expected that since on `main` we are using an option guard to enforce `DisableOption::KernelReuse`. Also, even though kernels are fused it seems to often hurt performance? image
> 
> image
> 
> image
> 
> image
> 
> I'll try and figure out what's happening locally.

Actually this might be an issue with `diff_report.py` since it is also showing a bunch of removed tests.

liqiangxl:
> It looks like this has a big impact on perf and it's not always positive. Looking at the codediffs for the C++ normalization benchmarks, there are a lot of cases where this change causes a previously segmented kernel to become unsegmented. I would not have expected that since on `main` we are using an option guard to enforce `DisableOption::KernelReuse`. Also, even though kernels are fused it seems to often hurt performance? image
> 
> image
> 
> image
> 
> image
> 
> I'll try and figure out what's happening locally.

Are these figures generated by code diff? That may not be reliable since it only does one run without repeat. Check locally sounds a good idea!

jacobhinkle:
> Are these figures generated by code diff? That may not be reliable since it only does one run without repeat. Check locally sounds a good idea!

Not only that but when we have parse errors the correspondence between tests is often broken so its comparing apples and oranges. That's what is happening here so these speedup numbers are useless.

jacobhinkle:
With #2186 I re-ran the report on the artifact from above (so using the same benchmarking run). Originally the plot looked like this:
image
But now it looks more reasonable:
image
As @liqiangxl mentioned these benchmark numbers are not definitive since we only use a small number of iterations. Note that apparently there are now many cases where we can safely re-use kernels so the normalization benchmarks should run quite a bit faster once we merge this. I'd expect similar in the python benchmarks.

naoyam:
Do you have a list of cases where this change results in different scheduling?

jacobhinkle:
> Do you have a list of cases where this change results in different scheduling?

I'm getting that now. I couldn't extract it from the previous run since we don't print the kernel when we reuse it. I modified the code to actually print reused kernels with a different filename, which I think will let us compare the kernel code between `NVFUSER_DISABLE=kernel_reuse` and the default behavior in this PR.

jacobhinkle:
> > Do you have a list of cases where this change results in different scheduling?
> 
> I'm getting that now. I couldn't extract it from the previous run since we don't print the kernel when we reuse it. I modified the code to actually print reused kernels with a different filename, which I think will let us compare the kernel code between `NVFUSER_DISABLE=kernel_reuse` and the default behavior in this PR.

I am trying to verify that we don't have any regressions but this is tricky. I pushed a commit which we might revert. It outputs cuda and ptx for reused segments so that we can compare against behavior on `main`. I am currently seeing some reordering of reused segmented kernels, which leads to reported diffs even though the segmentation is the same. I'll run codediff with this then post-process the results to try and address that.

jacobhinkle:
!build --diff-bench

naoyam:
Can't we just add some debug prints when a reuse occurs with this change but not with the current main? I suppose the number of cases is small, so we could just take a look at each individual case.

csrc/kernel_cache.cpp
jacobhinkle: Update: this is actually not getting the heuristic for the unsegmented fusion. It is merely running `canSchedule` on the fusion and returning the first `ScheduleHeuristic` (enum) that succeeds. Later we run `kernel_runtime->getMaybeHeuristicsFor` which will run `SegmentedGroup::getMaybeSchedulerEntry` to actually get a config. It's possible to refactor this to reuse that call so that we don't need to re-run the heuristic on unsegmented runtimes, but I think it would duplicate more of the segmentation machinery.

jacobhinkle:
> I suppose the number of cases is small, so we could just take a look at each individual case.

On main we are recompiling for each input size, so there are many cases where we are able to reuse. These correspond to the cases missed by the plots above since previously those benchmarks were shown as removed. For the shard shown in the plots in this comment the x axis goes from ~50 to ~925 so about 95% of cases appear to have been reused in that shard: https://github.com/NVIDIA/Fuser/issues/2163#issuecomment-2090586746.

jacobhinkle:
In local testing I haven't seen any diffs for unsegmented benchmarks but I haven't run a complete benchmarking run since it takes a very long time to run on a single machine. I am therefore pretty certain it is safe to reuse unsegmented kernels. There is still the possibility that a segmented fusion would be suboptimal, so we could disable reuse of segmented runtimes using an env var, or repurpose `NVFUSER_DISABLE=kernel_reuse` to allow reusing unsegmented kernels.

naoyam:
> > I suppose the number of cases is small, so we could just take a look at each individual case.
> 
> On main we are recompiling for each input size, so there are many cases where we are able to reuse. These correspond to the cases missed by the plots above since previously those benchmarks were shown as removed. For the shard shown in the plots in this comment the x axis goes from ~50 to ~925 so about 95% of cases appear to have been reused in that shard: [#2163 (comment)](https://github.com/NVIDIA/Fuser/pull/2163#issuecomment-2090586746).

Sorry, I phrased my question incorrectly. What I meant to say was if this change is not expected to use different scheduled kernels than simply disabling reuse, I thought maybe we could do something like:

```
if (reuse-due-to-this-change) {
  if (heuristics-without-reuse != reused-heuristics) {
    printf("scheduling change detected");
  }
}
```

jacobhinkle:
> Sorry, I phrased my question incorrectly. What I meant to say was if this change is not expected to use different scheduled kernels than simply disabling reuse, I thought maybe we could do something like:

Oh I see. To do so, I will need to create a new FusionKernelRuntime and segment then compare the heuristics to the runtime that would be reused. I will give it a shot.

jacobhinkle:
@naoyam that was a much better way to find mismatched kernels. Running locally I found this:
```
NvFuserScheduler_LayerNorm_BWD_fp16___GRAPH/NvFuserScheduler_LayerNorm_BWD_fp16/160/2/manual_time             76.9 us          132 us         9091 bytes_per_second=31.8323M/s Red On Fastest Dim // Persistent Kernel // Project Persistent Buffers //  // Iteration Domain: multiple reductions per block / split grid dimension /  // Inner Reduction Domain: persistent batch - 1 / vectorize / factor 2/Launch_Parameters[block(1/1/1)/grid(1/168/1)/4]
NvFuserScheduler_LayerNorm_BWD_fp16___GRAPH/NvFuserScheduler_LayerNorm_BWD_fp16/320/2/manual_time             78.0 us          133 us         8972 bytes_per_second=62.6669M/s Red On Fastest Dim // Persistent Kernel // Project Persistent Buffers //  // Iteration Domain: multiple reductions per block / split grid dimension /  // Inner Reduction Domain: persistent batch - 1 / vectorize / factor 2/Launch_Parameters[block(1/2/1)/grid(1/168/1)/8]
NvFuserScheduler_LayerNorm_BWD_fp16___GRAPH/NvFuserScheduler_LayerNorm_BWD_fp16/160/8/manual_time             45.6 us         99.3 us        15353 bytes_per_second=174.625M/s Red On Fastest Dim // Persistent Kernel // Project Persistent Buffers //  // Iteration Domain: multiple reductions per block / split grid dimension /  // Inner Reduction Domain: persistent batch - 1 / vectorize / factor 8/Launch_Parameters[block(1/2/1)/grid(1/84/1)/8]
NvFuserScheduler_LayerNorm_BWD_fp16___GRAPH/NvFuserScheduler_LayerNorm_BWD_fp16/320/8/manual_time             45.6 us          100 us        15361 bytes_per_second=348.78M/s Red On Fastest Dim // Persistent Kernel // Project Persistent Buffers //  // Iteration Domain: multiple reductions per block / split grid dimension /  // Inner Reduction Domain: persistent batch - 1 / vectorize / factor 8/Launch_Parameters[block(1/4/1)/grid(1/84/1)/16]
NvFuserScheduler_LayerNorm_BWD_fp16___GRAPH/NvFuserScheduler_LayerNorm_BWD_fp16/160/64/manual_time            16.5 us         70.5 us        42541 bytes_per_second=3.52441G/s Red On Fastest Dim // Persistent Kernel // Project Persistent Buffers //  // Iteration Domain: multiple reductions per block / split grid dimension /  // Inner Reduction Domain: persistent batch - 1 / vectorize / factor 8/Launch_Parameters[block(1/2/8)/grid(1/84/1)/64]
NvFuserScheduler_LayerNorm_BWD_fp16___GRAPH/NvFuserScheduler_LayerNorm_BWD_fp16/320/64/manual_time            17.5 us         72.0 us        39997 bytes_per_second=6.61482G/s Red On Fastest Dim // Persistent Kernel // Project Persistent Buffers //  // Iteration Domain: multiple reductions per block / split grid dimension /  // Inner Reduction Domain: persistent batch - 1 / vectorize / factor 8/Launch_Parameters[block(1/4/8)/grid(1/84/1)/128]
NvFuserScheduler_LayerNorm_BWD_fp16___GRAPH/NvFuserScheduler_LayerNorm_BWD_fp16/160/512/manual_time           15.3 us         74.1 us        46109 bytes_per_second=29.9947G/s Red On Fastest Dim // Persistent Kernel // Project Persistent Buffers //  // Iteration Domain: multiple reductions per block / split grid dimension /  // Inner Reduction Domain: pad to warp / persistent batch - 1 / vectorize / factor 8/Launch_Parameters[block(1/4/64)/grid(1/84/1)/1024]
NvFuserScheduler_LayerNorm_BWD_fp16___GRAPH/NvFuserScheduler_LayerNorm_BWD_fp16/320/512/manual_time           16.1 us         70.6 us        43065 bytes_per_second=56.9569G/s Red On Fastest Dim // Persistent Kernel // Project Persistent Buffers //  // Iteration Domain: multiple reductions per block / split grid dimension /  // Inner Reduction Domain: pad to warp / persistent batch - 1 / vectorize / factor 8/Launch_Parameters[block(1/4/64)/grid(1/84/1)/1024]
NvFuserScheduler_LayerNorm_BWD_fp16___GRAPH/NvFuserScheduler_LayerNorm_BWD_fp16/160/4096/manual_time          28.3 us         82.0 us        24711 bytes_per_second=129.917G/s Red On Fastest Dim // Persistent Kernel // Project Persistent Buffers //  // Iteration Domain: split grid dimension /  // Inner Reduction Domain: persistent batch - 4 / vectorize / factor 8/Launch_Parameters[block(1/8/16)/grid(1/84/1)/512]
NvFuserScheduler_LayerNorm_BWD_fp16___GRAPH/NvFuserScheduler_LayerNorm_BWD_fp16/320/4096/manual_time          36.9 us         90.4 us        18978 bytes_per_second=199.088G/s Red On Fastest Dim // Persistent Kernel // Project Persistent Buffers //  // Iteration Domain: split grid dimension /  // Inner Reduction Domain: persistent batch - 4 / vectorize / factor 8/Launch_Parameters[block(1/8/16)/grid(1/84/1)/512]
NvFuserScheduler_LayerNorm_BWD_fp16___GRAPH/NvFuserScheduler_LayerNorm_BWD_fp16/160/32768/manual_time          357 us          402 us         1966 bytes_per_second=82.4602G/s
NvFuserScheduler_LayerNorm_BWD_fp16___GRAPH/NvFuserScheduler_LayerNorm_BWD_fp16/320/32768/manual_time          698 us          743 us         1005 bytes_per_second=84.1626G/s
NvFuserScheduler_LayerNorm_BWD_fp16___GRAPH/NvFuserScheduler_LayerNorm_BWD_fp16/160/262144/manual_time        2830 us         2877 us          247 bytes_per_second=83.1763G/s
NvFuserScheduler_LayerNorm_BWD_fp16___GRAPH/NvFuserScheduler_LayerNorm_BWD_fp16/320/262144/manual_time        5873 us         6038 us          124 bytes_per_second=79.9837G/s
NvFuserScheduler_LayerNorm_BWD_fp16___GRAPH/NvFuserScheduler_LayerNorm_BWD_fp16/160/1048576/manual_time      11934 us        12010 us           53 bytes_per_second=78.8841G/s
NvFuserScheduler_LayerNorm_BWD_fp16___GRAPH/NvFuserScheduler_LayerNorm_BWD_fp16/320/1048576/manual_time      25637 us        25862 us           30 bytes_per_second=73.2879G/s
NvFuserScheduler_LayerNorm_BWD_fp16___GRAPH/NvFuserScheduler_LayerNorm_BWD_fp16/2/32768/manual_time           43.5 us         87.9 us        15178 bytes_per_second=11.2143G/s
NvFuserScheduler_LayerNorm_BWD_fp16___GRAPH/NvFuserScheduler_LayerNorm_BWD_fp16/8/32768/manual_time           45.4 us         99.6 us        15802 bytes_per_second=34.9608G/s
Found mismatched parameters! Desired params:

===== Pointwise Parameters ========
Tag: Pointwise heuristics Pointwise Characteristics:
 Gridx: 1 BlckY: 1 BlckX: 128
2D Schedule
  Bcast break point: 1
Vectorize, Factor: 4


===== Reduction Parameters ========

Red On Slow Dim

Iteration Domain: blockIdx.x / threadIdx.x / multiple reductions per block /
Inner Reduction Domain: unroll / factor 8
Launch Parameters: BlockDim.x = 392, BlockDim.y = 1, BlockDim.z = -1, GridDim.x = -1, GridDim.y = -1, GridDim.z = -1, Smem Size = 0
Compile Parameters: index_type = int, maxrregcount = 255, enable_magic_zero = 1, enable_ptxas_verbose = 0

===== Reduction Parameters ========
Tag: 2D Register Inner Persistent Heuristic.

Red On Fastest Dim
Persistent Kernel
Batches per block: 8

Iteration Domain: blockIdx.x /
Inner Reduction Domain: cross block - threadIdx.x /  pad to warp / persistent batch - 8 / vectorize / factor 4
Launch Parameters: BlockDim.x = -1, BlockDim.y = 1, BlockDim.z = -1, GridDim.x = -1, GridDim.y = -1, GridDim.z = -1, Smem Size = 0
Compile Parameters: index_type = int, maxrregcount = 64, enable_magic_zero = 1, enable_ptxas_verbose = 0

Reusing params (THESE DON'T MATCH THE ABOVE PARAMS):

===== Pointwise Parameters ========
Tag: Pointwise heuristics Pointwise Characteristics:
 Gridx: 1 BlckY: 1 BlckX: 128
2D Schedule
  Bcast break point: 1
Vectorize, Factor: 4


===== Reduction Parameters ========

Red On Slow Dim

Iteration Domain: blockIdx.x / threadIdx.x / multiple reductions per block /
Inner Reduction Domain: unroll / factor 8
Launch Parameters: BlockDim.x = 256, BlockDim.y = 1, BlockDim.z = -1, GridDim.x = -1, GridDim.y = -1, GridDim.z = -1, Smem Size = 0
Compile Parameters: index_type = int, maxrregcount = 255, enable_magic_zero = 1, enable_ptxas_verbose = 0

===== Reduction Parameters ========

Red On Fastest Dim

Iteration Domain: blockIdx.y /
Inner Reduction Domain: cross block - threadIdx.x /  pad to warp / cross grid - blockIdx.x / split grid dim / split grid dimension / vectorize / factor 4
Launch Parameters: BlockDim.x = 224, BlockDim.y = -1, BlockDim.z = -1, GridDim.x = 10, GridDim.y = -1, GridDim.z = -1, Smem Size = 0
Compile Parameters: index_type = int, maxrregcount = 255, enable_magic_zero = 1, enable_ptxas_verbose = 0

```
Note that batches per block is missing in the reused heuristic params.

jacobhinkle:
!build --pybench

jacobhinkle:
Python benchmarks are hitting OOM it seems. I ran layer norm forward locally and I see more use than with BWD:
```
FusionExecutorCache Reuse stats:
  Unique inputs seen: 69
  Successfully reused runtimes: 57 (82.6087%) 
    segmented: 14 (20.2899% of total)
    unsegmented: 43 (62.3188% of total)
  Failed to reuse runtimes: 0 (0%)
    segmented: 0 (0% of total) 
    unsegmented: 0 (0% of total) 
NvFuserScheduler_LayerNorm_fp32___GRAPH/NvFuserScheduler_LayerNorm_fp32/16384/1600/manual_time
```

jacobhinkle (commented):
I pushed some changes that implement a complete solution. We now check not only whether the fusion is segmented or not, but whether all the segments chose the same heuristics and params.

jacobhinkle:
!build

jacobhinkle:
!build

csrc/fusion_segmenter.h
```diff
@@ -334,7 +334,7 @@ class SegmentedFusion {
 
   //! Is the fusion segmented?
   bool isSegmented() const {
-    return !groups_.empty();
+    return groups_.size() > 1;
```
jacobhinkle, line 337: `groups_.size() == 1` means unsegmented. This bug was causing test failures.

jacobhinkle:
!build

jacobhinkle:
!build

PR #2165, opened 2024-04-30
Add nvfuser pythonlib search paths to RPATH.

### LINKED ISSUE #2144: Change in ORIGIN after hiding symbols
It seems that #1324 or a nearby commit may have inadvertently changed the lookup paths for certain libraries, making it so that libnvfuser_codegen.so could not be located. Internal apex norm tests started giving:

```
ImportError: libnvfuser_codegen.so: cannot open shared object file: No such file or directory
```

A workaround is to copy the `libnvfuser_codegen.so` library into the `torch/lib` subdirectory.

tfogal:
Fixes #2144.


tfogal:
!build

tfogal:
cc @nWEIdia

PR #2168, opened 2024-05-01
Allocation order refactor

jjsjann123:
refactored allocation order inference pass:
  * Instead of per operation propagation rule, we are now using IdModel mapping to map allocation domain of reference tensor to rfactor domain of target tensor.
  * Updated the inference API to allow specified sources and destinations for the propagation.
  ```
void inferenceAllocationOrder(
    Fusion* fusion,
    const std::vector<TensorView*>& srcs,
    const std::vector<TensorView*>& dsts);
  ```

  * The propagation tried to keep the memory format of `dsts` closer to the `srcs` to simplify scheduling as well as facilitate vectorization. It works roughly as:
    * For each entry `dst`, among all its producers in `srcs`, we'll find the one with the most loop iter domain in its allocation domain as the reference `ref`
    * We try to map each iter domain in `dst`'s rfactor domain to `ref`'s allocation order domain and push those as the inner dimension in `dst`'s new allocation domain, while pushing unmapped iter domains as outer dimensions.
    * I have to put in a WAR for the mapping logic for now, since reduction scheduler is struggling with permuted output. See issue #2202. The WAR is simply to preserve the existing position of reduction iter domain in rfactor the same as it would be in its new allocation domain. This WAR is supposed to be removed at a later point once we fixed reduction scheduler. I kept both code path in the PR for easier future cleanup.


tests/cpp/test_allocation_order_inference.cpp
```diff
@@ -124,83 +132,51 @@ TEST_F(AllocationOrderInferenceTest, BinaryOpPropagation) {
...
-        preseg_passes::inferenceAllocationOrder(&fusion);
-    EXPECT_THAT(inferred_layout.at(tv2), ElementsAre(0, 2, 1, 3));
-    EXPECT_TRUE(inferred_layout.count(tv3) == 0);
-    EXPECT_TRUE(inferred_layout.count(tv4) == 0);
+    preseg_passes::inferenceAllocationOrder(&fusion);
+    EXPECT_THAT(getAllocationDomainPermutation(tv2), ElementsAre(3, 1, 0, 2));
+    EXPECT_THAT(getAllocationDomainPermutation(tv3), ElementsAre(3, 1, 0, 2));
   }
+  // TODO: open an issue. seems to hit an assert in IdModel(&fusion)
+  // {
```
jjsjann123, line 140: Note to myself. verify this with ToT main. I'm guessing it's just some idmodel config that I wasn't using properly.

jjsjann123:
local test failures are encountered here:
```
[  FAILED  ] 9 tests, listed below:
[  FAILED  ] IndexingOpTest.TakeAlongAxisIntermediateTensorTranspose1_CUDA
[  FAILED  ] TransposeTest.FusionTransposeSelfMapping
[  FAILED  ] TransposeTest.TransposeAggregatedVectorizationWidth
[  FAILED  ] TransposeTest.TransposeSplitAggregatedVectorizationWidth
[  FAILED  ] MoveSplitCatTest.Noncancellable_SomeButNotAllArePermuted
```

jjsjann123:
!build

jjsjann123:
> MoveSplitCatTest.Noncancellable_SomeButNotAllArePermuted

^^^ Note to myself. This is the case where self mapping is encountered.

jjsjann123:
muhahaha local test passed. :partying_face:

jjsjann123:
I'll wait until CI passed before start cleaning up the PR for review.

jjsjann123:
`NVFuserTest.FusionReductionWithTrivialReduction_CUDA` fails on V100 with wrong result. This one is a bit concerning. 
```
00:04:20 C++ exception with description "aten_output_tensor.allclose( fusion_output_tensor.to(aten_output_tensor.dtype()), tolerance_values.second, tolerance_values.first, true) INTERNAL ASSERT FAILED at "/opt/pytorch/nvfuser/tests/cpp/validator.h":119, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. 
00:04:20 
00:04:20 Validation error in output 1 on line 5239 in file /opt/pytorch/nvfuser/tests/cpp/test_gpu1.cpp.
00:04:20   Detected abs error of: 38.3021
00:04:20     absolute tolerance was set to 1.30594e-05
00:04:20     and relative tolerance set to 1.30594e-07
```

a log of python instance norm is failing and complaining about wrong vectorization. feels like that's something that we'd want to fix before merging this PR. :cry: 
`tests/python/test_normalization.py::test_instance_norm[bfloat16-True-True-True-channels_last-1-7-3-True] - RuntimeError: The contiguity of a broadcast/reduction dimension must be None. The contiguity of a non-broadcast/reduction dimension must be true/false`

test_matmul failure. I vaguely remember seeing some notes mentioning about vectorization support in matmul. I'll double check if this is expected. I'll likely just fix the test by specifying output allocation domain and then just open an issue to track it.
`00:01:48 [ RUN      ] GPUTTensorCoreTest.FusionAmpereSwizzle_CUDA
00:01:48 unknown file: Failure
00:01:48 C++ exception with description "Vectorized dim for consumer has to be from a contiguous inner most position. tv: T3_g[ iblockIdx.x67{( ceilDiv(( (( (( getMetaData(T0) )).logical_size ))[0] ), 128) )}, iblockIdx.y69{( ceilDiv(( (( (( getMetaData(T1) )).logical_size ))[2] ), 128) )}, ithreadIdx.z182{( ceilDiv(128, 64) )}, ithreadIdx.y184{( ceilDiv(128, 64) )}, iS186{( ceilDiv(64, 16) )}, iS188{( ceilDiv(64, 8) )}, ithreadIdx.x199{( ( ( ceilDiv(( ceilDiv(16, 8) ), 2) ) * 8 ) * ( ceilDiv(8, 2) ) )}, iS195{( ceilDiv(8, 8) )}, iS193{2}, iV198{2} ] ca_pos( 9 ) produce_pos( 2 ), allocation domain: iS216{( (( (( getMetaData(T1) )).logical_size ))[2] )}, iS215{( (( (( getMetaData(T0) )).logical_size ))[0] )}, vectorized id: iS216{( (( (( getMetaData(T1) )).logical_size ))[2] )}, innermost id: iS215{( (( (( getMetaData(T0) )).logical_size ))[0] )}, contiguity: t`

jjsjann123:
!build

jjsjann123:
CI passed! :clap:

jjsjann123:
!build

tests/cpp/test_allocation_order_inference.cpp
```diff
@@ -124,82 +135,9 @@ TEST_F(AllocationOrderInferenceTest, BinaryOpPropagation) {
...
-    auto tv0 = makeSymbolicTensor({-1, -1, 1, 1});
-    fusion.addInput(tv0);
-    auto tv1 = makeSymbolicTensor({-1, -1, 1, 1});
-    fusion.addInput(tv1);
-    // tv2 should have allocation order from tv0
-    auto tv2 = add(tv0, tv1);
-    fusion.addOutput(tv2);
-
-    // reshape propagation is not supported yet
-    auto tv3 = reshape(
```
jjsjann123, line 180: This test has a reshape. Before concretization, it triggers an assert in ComputeAtMap constructor. We are removing this test for that.

Note to myself, the error message is bad. should we check for symbolic dimensions as a hint for Fusion before concretization and throw a meaningful error message instead of assert? cc'ing @jacobhinkle

tests/cpp/test_allocation_order_inference.cpp
```diff
@@ -124,82 +135,9 @@ TEST_F(AllocationOrderInferenceTest, BinaryOpPropagation) {
...
-    FusionGuard fg(&fusion);
-
-    // Testing propagation between two tensors
-    // tv0 and tv1 has the same number of non-broadcast iter domains, so lhs
-    // operand would propagate its allocation order.
-    auto tv0 = makeSymbolicTensor({-1, -1, 1, 1});
-    fusion.addInput(tv0);
-    auto tv1 = makeSymbolicTensor({-1, -1, 1, 1});
-    fusion.addInput(tv1);
-    // tv2 should have allocation order from tv0
```
jjsjann123, line 144: removing this test case since it's now considered ambiguous.
I can probably also keep it an assert that target is not populated with stride order.

jjsjann123:
Looks like the simplified merge logic isn't totally right. I might have to revert back to complicated logic.

logging failing tests below
```
00:20:03 [  FAILED  ] NVFuserTest.DynamicSqueezeTrivialWelford
00:20:03 [  FAILED  ] AliasTest.NotAllOutputsAlias_Reduction
00:20:03 [  FAILED  ] ResizeTest.PadScheduler4

00:09:17 [2024-05-04 00:01:00] thunder/tests/test_grad.py::test_phantom_grad_vs_torch_consistency_einsum_nvfuser_cuda_float32 <- thunder/tests/framework.py [31mFAILED[0m
00:09:21 [2024-05-04 00:01:04] thunder/tests/test_grad.py::test_phantom_grad_vs_torch_consistency_einsum_nvfuser_cuda_float64 <- thunder/tests/framework.py [31mFAILED[0m
```

jjsjann123:
!build

jjsjann123:
!build

jjsjann123:
!build

jjsjann123:
!build

jjsjann123:
!build

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -15,442 +16,274 @@ namespace nvfuser::preseg_passes {
...
+
+  // map target rfactor domain into ref's allocation domain
+  std::vector<IterDomain*> mapped_id_vec;
+  std::unordered_set<IterDomain*> mapped_id_set;
+
+  // logic to preserve reduction iter domain in target to WAR issue #2202
+#if true
+  for (auto* ref_id : ref_alloc_domain) {
+    for (auto* id : target_rfactor_domain) {
+      // sharp-edges 0
```
jjsjann123, line 88: This is necessary to pass
```
00:21:19 [  FAILED  ] 2 tests, listed below:
00:21:19 [  FAILED  ] AliasTest.NotAllOutputsAlias_Reduction
00:21:19 [  FAILED  ] ResizeTest.PadScheduler4
```

Alternatively, we can also try to use compute at :: INNSERMOST mapping instead of exact mapping in IdModel. But eventually we need to fix https://github.com/NVIDIA/Fuser/issues/2202

jjsjann123:
!build

tests/cpp/test_allocation_order_inference.cpp
jacobhinkle: I think it might be best to just manually concretize the fusion before running the pass in this test, mimicking what would happen in a real run. To do that you can use this pattern:
https://github.com/NVIDIA/Fuser/blob/4e8356915fc6ffa823cf97fed4249cc00f317fdf/tests/cpp/test_dynamic_transform.cpp#L71-L72

jjsjann123: Got ya. I'll give this a try. (note that the existing test might still need to be updated just so that we can have a dominating tensor)

jjsjann123:
!build

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -15,442 +16,279 @@ namespace nvfuser::preseg_passes {
...
+void inferenceAllocationOrder(
+    Fusion* fusion,
+    const std::vector<TensorView*>& srcs,
+    const std::vector<TensorView*>& dsts) {
+  // build IdModel, setting allow_self_mapping to avoid assert
+  // even though we do NOT populate allocation order where self_mapping is
+  // present
+  auto id_model =
+      IdModel(fusion, /*build_graphs=*/true, /*allow_self_mapping=*/true);
+  const ValGraph& exact_graph = id_model.idGraph(IdMappingMode::EXACT);
```
zasdfgbnm, line 210: Should we use permissive graph?

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -15,442 +16,279 @@ namespace nvfuser::preseg_passes {
...
-  std::vector<IterDomain*> allocation_domain(rank, nullptr);
-  // specify allocation domain with dimension per allocation order.
-  for (auto i : c10::irange(rank)) {
-    allocation_domain[i] = rfactor_dom.at(alloc_order.at(i));
-  }
-
-  return allocation_domain;
+// counting the number of non-broadcast & non-reduction iter domains in tv's
+// allocation domain.
+size_t countLoopIterDomains(const TensorView* tv) {
```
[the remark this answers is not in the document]
jjsjann123: New name suggestion are more than welcomed.

I'm just trying to describe iter domain that's not a broadcast nor a reduction... So I thought about calling it a `IterationIterDomain` but that sounded weird and I thought `Loop` is a better choice.
How about `NonTriviaIterDomain`?

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -15,442 +16,279 @@ namespace nvfuser::preseg_passes {
...
+  std::vector<IterDomain*> ref_alloc_domain = ref->getMaybeAllocationDomain();
+  const std::vector<IterDomain*>& target_rfactor_domain =
+      target->getMaybeRFactorDomain();
+
+  // map target rfactor domain into ref's allocation domain
+  std::vector<IterDomain*> mapped_id_vec;
+  std::unordered_set<IterDomain*> mapped_id_set;
+
+  // logic to preserve reduction iter domain in target to WAR issue #2202
+#if true
```
wujingyue, line 85: Does `#if true` actually work? I thought `true` would need to be defined in macro. 

(I think you intended to remove this)

wujingyue: I missed the comment -- you do intend to leave it.

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -15,442 +16,279 @@ namespace nvfuser::preseg_passes {
...
+//   unmapped_ids_vec {iS3[i3], iS4[i4], iS6[i5], ir8[1]}
+// 3. append mapped_id_vec at the end of unmapped_id_vec.
+//   target_alloc_domain
+//   {iS3[i3], iS4[i4], iS6[i5], ir8[1], ir5[i1], iS7[i2]}
+void AllocationOrderMapping(
+    const IdModel& id_model,
+    const TensorView* ref,
+    TensorView* target) {
+  const DisjointSets<Val*>& val_sets =
+      id_model.idGraph(IdMappingMode::EXACT).disjointValSets();
```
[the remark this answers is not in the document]
jjsjann123: My motivation is that we do NOT want to map broadcast IDs. I think we should avoid mapping trivial iterdomain. (be it broadcast or reduction).

A concrete (likely limited) example:
If we are trying to map TV0 [i0, i1, b2, i3] to TV1 [i0, i1, i4, i3]. where broadcast `b2` would map to `i4`

if TV0 has an allocation domain as {i0, i1, b2, i3}, if we do NOT map broadcast iterdomain, then the target TV1 would have an allocation domain of {i4, i0, i1, i3}, which has a higher upper bound for vectorization for TV0 {i0*i1*i3} (talking about a point-wise scheduler).
otherwise if we map broadcast, the allocation order would be {i0, i1, i4, i3}. TV0 would only have {i3} as the upper bound for vectorization.

csrc/preseg_passes/allocation_order_inference.cpp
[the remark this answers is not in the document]
zasdfgbnm: Honestly I don't know. I am expecting someone will suggest a better name :-D

csrc/preseg_passes/allocation_order_inference.cpp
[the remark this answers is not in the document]
zasdfgbnm: Hmm, if you are talking about pointwise fusion, wouldn't TV0 be vectorizable to `i3` already? For example, if we have
```
T0[1024, 1024, 1, 3] + T1[1024, 1024, 4, 3]
```
We can not vectorize 4 for T0 due to loop promotion.

zasdfgbnm:
> !build

I suggest using `!build --diff`

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -15,442 +16,279 @@ namespace nvfuser::preseg_passes {
...
-    if (iter->second.empty()) {
-      // We still want to ensure that we propagate empty allocation order if
-      // there's no candidate with a non-empty allocation order
-      if (src == nullptr) {
-        src = tv;
+  }
+#else
+  // mapping id between ref's allocation domain to target's rfactor domain
+  for (auto* ref_id : ref_alloc_domain) {
+    for (auto* id : target_rfactor_domain) {
```
zasdfgbnm, line 140: Instead of using two nested for loop, would it be better to use a ValGroup->IterDomain map mapping the group of each ID in `target_rfactor_domain` to the ID itself?

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -15,442 +16,279 @@ namespace nvfuser::preseg_passes {
...
-        constructAllocationDomain(out_tv, mapped_entry->second), true);
-  }
+  // mark input TensorViews as propagation sources
+  auto input_tvs = ir_utils::filterByType<TensorView>(fusion->inputs());
+  std::vector<TensorView*> srcs(input_tvs.begin(), input_tvs.end());
+  // mark output TensorViews as propagation destinations
+  auto output_tvs = ir_utils::filterByType<TensorView>(fusion->outputs());
+  std::vector<TensorView*> dsts(output_tvs.begin(), output_tvs.end());
+  // propagate allocation domain from sources to destinations
+  inferenceAllocationOrder(fusion, srcs, dsts);
```
wujingyue, line 291: Given `srcs` and `dsts` can be determined by reading `fusion`, why do we pass them as arguments? Because you need to customize that in tests?

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -15,442 +16,279 @@ namespace nvfuser::preseg_passes {
...
+  std::vector<IterDomain*> target_alloc_domain = target_rfactor_domain;
+  // removing mapped ids.
+  auto unmapped_ids_vec_end = std::remove_if(
+      target_alloc_domain.begin(),
+      target_alloc_domain.end(),
+      [&mapped_id_set](IterDomain* it) {
+        return mapped_id_set.count(it) != 0;
+      });
+  // appending mapped ids at the end of target_alloc_domain.
+  std::copy(mapped_id_vec.begin(), mapped_id_vec.end(), unmapped_ids_vec_end);
```
zasdfgbnm, line 157: What if we have `reshape`?

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -15,442 +16,279 @@ namespace nvfuser::preseg_passes {
...
-//   domain that's a permutation of its corresponding rfactor domain and record
-//   it as the allocation order of the tensor;
-//   2. Traverse the fusion IR, propagate allocation order and record results in
-//   alloc_order_map.
-std::unordered_map<const TensorView*, AllocationOrder> inferenceAllocationOrder(
-    Fusion* fusion) {
-  std::unordered_map<const TensorView*, AllocationOrder> alloc_order_map;
+    // high water mark for candidate of ref.
+    size_t non_bc_high_water_mark = 0;
+    for (const auto& iter : loop_iter_count) {
```
wujingyue, line 240: For determinism, can you change this to `for each source` and `loop_iter_count.at`? As is, I suspect ties may lead to different reference tensors being picked, making debugging hard.

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -15,442 +16,279 @@ namespace nvfuser::preseg_passes {
...
+      if (iter.second == non_bc_high_water_mark && ref != nullptr) {
+        // ensure that there's no ambiguity on permutation mapping from multiple
+        // references. we need both ref candidates to have the same mapping on
+        // allocation domain
+        for (auto i : c10::irange(ref->nDims())) {
+          if (!val_sets.permissiveAreMapped(
+                  ref->getMaybeAllocationDomain()[i],
+                  iter.first->getMaybeAllocationDomain()[i])) {
+            // reset ref to nullptr, while keeping the iterdomain count high
+            // water mark. No propagatoin will occur unless we found another ref
```
wujingyue, line 266:
```suggestion
            // water mark. No propagation will occur unless we found another ref
```

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -15,442 +16,279 @@ namespace nvfuser::preseg_passes {
...
+      // new candidate found, update ref and high water mark
+      if (iter.second > non_bc_high_water_mark) {
+        ref = iter.first;
+        non_bc_high_water_mark = iter.second;
+        continue;
+      }
+      // found multiple candidate with the same iterdomain count
+      if (iter.second == non_bc_high_water_mark && ref != nullptr) {
+        // ensure that there's no ambiguity on permutation mapping from multiple
+        // references. we need both ref candidates to have the same mapping on
```
wujingyue, line 259: Is this new logic? How do you compare this with picking an arbitrary `ref` among them when there's a tie like this.

tests/cpp/test_allocation_order_inference.cpp
```diff
@@ -25,6 +25,16 @@ using testing::ElementsAre;
...
 using AllocationOrderInferenceTest = NVFuserTest;
 
+std::vector<int64_t> getAllocationDomainPermutation(TensorView* tv) {
+  std::optional<std::vector<int64_t>> permutation =
+      ir_utils::computePermutation(
+          tv->getMaybeRFactorDomain(), tv->getMaybeAllocationDomain());
+  if (permutation.has_value()) {
+    return permutation.value();
+  }
+  return {};
```
wujingyue, line 35: Wdyt about
```suggestion
  return *permutation;
```
for YAGNI and less ambiguity on what an empty order means. As is, it means either not a permutation or a valid permutation on a rank-0 tensor.

wujingyue (commented):
I'm still reviewing AllocationOrderMapping. Other parts LGTM with comments.

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -15,442 +16,279 @@ namespace nvfuser::preseg_passes {
...
+//     {iS3[i3], iS4[i4], ir5[i1], iS6[i5], iS7[i2], ir8[1]}
 //
-//   // create AllocationOrderInferencer
-//   AllocationOrderInferencer infer(alloc_order_map);
-//   // propagates AllocationOrder from entries already in alloc_order_map
-//   infer.traverse(fusion);
-//   // all tensor that's propagated successfully will have their allocation
-//   // order in alloc_order_map
+// 1. we project iter domains from targets' rfactor domain which has an exact
+// map to ref's allocation domain. (sharp-edge 0: we exlucde mapping from
```
wujingyue, line 43:
```suggestion
// map to ref's allocation domain. (sharp-edge 0: we exclude mapping from
```

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -15,442 +16,279 @@ namespace nvfuser::preseg_passes {
...
+// 1. we project iter domains from targets' rfactor domain which has an exact
+// map to ref's allocation domain.
+//   mapped_id_vec {ir5[i1], iS7[i2]}
+// 2. remove all projected iter domains from target's rfactor
+// domain:
+//   unmapped_ids_vec {iS3[i3], iS4[i4], iS6[i5], ir8[1]}
+// 3. append mapped_id_vec at the end of unmapped_id_vec.
+//   target_alloc_domain
+//   {iS3[i3], iS4[i4], iS6[i5], ir8[1], ir5[i1], iS7[i2]}
+void AllocationOrderMapping(
```
wujingyue, line 69:
```suggestion
void mapAllocationDomain(
```

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -15,442 +16,279 @@ namespace nvfuser::preseg_passes {
...
-//   // create AllocationOrderInferencer
-//   AllocationOrderInferencer infer(alloc_order_map);
-//   // propagates AllocationOrder from entries already in alloc_order_map
-//   infer.traverse(fusion);
-//   // all tensor that's propagated successfully will have their allocation
-//   // order in alloc_order_map
+// 1. we project iter domains from targets' rfactor domain which has an exact
+// map to ref's allocation domain. (sharp-edge 0: we exlucde mapping from
+// iteration id on ref to reduction id on target to avoid unnecessary
+// re-ordering which exposes issue 2202).
```
wujingyue, line 45:
```suggestion
// re-ordering which exposes #2202).
```

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -15,442 +16,279 @@ namespace nvfuser::preseg_passes {
...
+  const DisjointSets<Val*>& val_sets =
+      id_model.idGraph(IdMappingMode::EXACT).disjointValSets();
+
+  std::vector<IterDomain*> ref_alloc_domain = ref->getMaybeAllocationDomain();
+  const std::vector<IterDomain*>& target_rfactor_domain =
+      target->getMaybeRFactorDomain();
+
+  // map target rfactor domain into ref's allocation domain
+  std::vector<IterDomain*> mapped_id_vec;
+  std::unordered_set<IterDomain*> mapped_id_set;
```
wujingyue, line 82: Can https://github.com/NVIDIA/Fuser/blob/2b115a1ec25ada00332d1e7f754dc1a16db66bf2/csrc/disjoint_set.h#L42 help here? So you don't need to keep two data structures in sync.

jjsjann123:
> If we infer a global order of all `ValGroup`s based on all inputs, and make this information available to all schedulers, could that be a solution to #2198?

For #2198, yes, scheduler/segmenter can use this within the fusion segment to infer tentative allocation order on segment boundaries.

I think we can take a short-cut for matmul/linear. We can add inputs to matmul/linear ops in our `dsts` vector and infer their allocation domain from in the global fusion IR. Hopefully that's still going to be a right decision to make?! :crossed_fingers: Open to discussion on this as well

csrc/preseg_passes/allocation_order_inference.cpp
jjsjann123: not just in tests, but also as a utility thing to use. i.e. we would also want to include intermediate tensor. Immediate usage is for inputs to matmul/linear. See #2198 and https://github.com/NVIDIA/Fuser/pull/2168#issuecomment-2099086932

csrc/preseg_passes/allocation_order_inference.cpp
wujingyue: I just didn't know `#if true` was a thing -- I saw `#if 1` much more often. After searching, I think `true` is indeed specially preprocessed, so we should be good!

csrc/preseg_passes/allocation_order_inference.cpp
```diff
@@ -15,442 +16,279 @@ namespace nvfuser::preseg_passes {
...
-    auto* tv = dynamic_cast<TensorView*>(val);
-    // skip non TensorView entry
-    if (tv == nullptr) {
+  // removing mapped ids and reduction ids to create unmapped_ids_vec.
+  // This means for the rest of ids in target_rfactor_domain that's not in
+  // mapped_id_set, they are either 1. a reduction domain, or; 2. in
+  // [unmapped_ids_vec.begin(), unmapped_ids_vec_end) This ensures that
+  // sharp-edges 1's loop would reconstruct a permutation of the
+  // target_rfactor_domain, hence a valid allocation domain for target.
+  std::vector<IterDomain*> unmapped_ids_vec = target_rfactor_domain;
```
wujingyue, line 108:
```suggestion
  std::vector<IterDomain*> unmapped_ids = target_rfactor_domain;
```
since you don't have a `_set` version.

wujingyue (approved):

csrc/preseg_passes/allocation_order_inference.cpp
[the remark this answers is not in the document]
jjsjann123: > I am expecting someone will suggest a better name :-D

Well I'll keep it as-is till that promised day comes.

csrc/preseg_passes/allocation_order_inference.cpp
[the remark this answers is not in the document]
jjsjann123: I'm assuming ` loop promotion` means inlining?!

> T0[1024, 1024, 1, 3] + T1[1024, 1024, 4, 3]

This is a bad example. :) Because T1 has more LoopIterDomain so it would dominate the output stride order.

I'm going to change it to > T2[1025, 1024, 4, 3] = T0[1025, 1024, 1, 3] + T1[1, 1, 4, 1]

if we propagate T0's allocation domain to T2 while ignoring broadcast, We'll have T2's allocation domain being
```T2_alloc = [4, 1025, 1024, 3]```

So scheduling will reorder with `T2[4, 1025, 1024, 3]` which will propagate to `T0[1, 1025, 1024, 3]`. giving it's allocation domain `T0_alloc[1025, 1024, 1, 3]`, shouldn't we be able to figure out that we [1025, 1024, 3] is contiguous and we should be able to vectorize on the factor of [1025*1024*3]?

File: csrc/preseg_passes/allocation_order_inference.cpp (hunk 11 above)
Comment (jjsjann123) [reply csrc/preseg_passes/allocation_order_inference.cpp:157]: This is intentionally left out.

iiuc, reshape that does merge/split/resize wouldn't map to anything and those new dimensions would just be pushed to outer dimensions, which might not be the best decision.
The question is also, how are we treating reshape in IdModel? I tried to use `ComputeAtMap`'s INNERMOST mapping, I think could be one of the things for us to use here. where we'll be able to map through a reshape.

i.e. if we have `i0` -> `i0/4, 4`, we'll be mapping `i0` to `4` here, I think that aligns with what we are trying to do here on mapping the order of inner dimensions from source to destination.
How does that sound like? cc'ing @naoyam 

Having said that, I'm a little bit scared about pushing for that at this moment. reduction scheduler is already struggling with keeping up, exposing more propagation at the time might expose more issues. Which we would want to get there but maybe not yet.

File: csrc/preseg_passes/allocation_order_inference.cpp (hunk 8 above)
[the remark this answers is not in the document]
Comment (zasdfgbnm) [reply csrc/preseg_passes/allocation_order_inference.cpp:74]: Oh, I see what you mean. You are right. Thanks for the clarification!

**Comment by zasdfgbnm** [comment]
> I think we can take a short-cut for matmul/linear. We can add inputs to matmul/linear ops in our `dsts` vector and infer their allocation domain from in the global fusion IR. Hopefully that's still going to be a right decision to make?! 🤞 Open to discussion on this as well

Sounds like an interesting approach. Let's try that (in a separate PR).

File: csrc/preseg_passes/allocation_order_inference.cpp (hunk 11 above)
Comment (wujingyue) [reply csrc/preseg_passes/allocation_order_inference.cpp:157]: I think you'd need to reason about merge/split outside IdModel, e.g., https://github.com/NVIDIA/Fuser/blob/2b115a1ec25ada00332d1e7f754dc1a16db66bf2/csrc/preseg_passes/move_split_cat.cpp#L323-L338. And the logic would be similar to https://github.com/NVIDIA/Fuser/blob/2b115a1ec25ada00332d1e7f754dc1a16db66bf2/csrc/alias_analysis.cpp#L163, although it doesn't use IdModel.

Comment (jjsjann123) [reply csrc/preseg_passes/allocation_order_inference.cpp:157]: yeah, that's basically the ValGraph? I'll just need to do find the path between two ValGroups and reason about that.

**Comment by jjsjann123** [comment]
> > I think we can take a short-cut for matmul/linear. We can add inputs to matmul/linear ops in our `dsts` vector and infer their allocation domain from in the global fusion IR. Hopefully that's still going to be a right decision to make?! 🤞 Open to discussion on this as well
> 
> Sounds like an interesting approach. Let's try that (in a separate PR).

SGTM

**Comment by jacobhinkle** [comment]
> If we infer a global order of all `ValGroup`s based on all inputs, and make this information available to all schedulers, could that be a solution to #2198?

Doesn't transpose mean we cannot have a total order on `ValGroup`s that places innermost dims properly everywhere?

File: csrc/preseg_passes/allocation_order_inference.cpp (hunk 11 above)
Comment (jacobhinkle) [reply csrc/preseg_passes/allocation_order_inference.cpp:157]: IIUC there's a partial order of IterDomains in the reference tv's allocation domain. We can extend that to a partial order of `ValGroup`s in a straightforward way. We would like to extend that to a partial order on the ValGraph using lexicographic order i.e. each node of the graph orders its children. Then use that ordering to sort the consumer allocation domain. Some rules on the ValGraph would be that `a < c && b < c` implies `Merge(a, b) < c`, `a < b && a < c` implies `a < Merge(b, c)`, and `b, c = Split(a)` implies `b < c`. There are upstream implications implied by the contrapositives and transitivity of `<` as well: `b, c = Split(a) && b < d && c < d` implies `a < d` and `c = Merge(a, b) && c < d` implies `a < d && b < d` for example. Other expressions like `Resize` have no implications; e.g. `b = Resize(a)` implies nothing about `b < c` given either `a < c` or `c < a`. You'd start with the orders implied by the reference tensor and all the split outputs, then you'd complete the partial order by propagating these rules. Does that sound right?

File: csrc/preseg_passes/allocation_order_inference.cpp (hunk 12 above)
Comment (jjsjann123) [reply csrc/preseg_passes/allocation_order_inference.cpp:240]: > As is, I suspect ties may lead to different reference tensors being picked, making debugging hard.

I think you see it below that we requires all `tie` to have identical mapping, hence the order of how we run into ties doesn't matter.

File: csrc/preseg_passes/allocation_order_inference.cpp (hunk 14 above)
Comment (jjsjann123) [reply csrc/preseg_passes/allocation_order_inference.cpp:259]: The old logic does propagation only through binary/ternary op, which I think is relatively easy to justify when we make an arbitrary choice.

With IdModel mapping, we are not following how the operands are *joined* together. I'm a bit hesitant to arbitrarily break the tie and I lean towards being on the conservative side.

File: csrc/preseg_passes/allocation_order_inference.cpp [hunk 21]
```diff
@@ -15,442 +16,279 @@ namespace nvfuser::preseg_passes {
...
+
+  std::vector<IterDomain*> ref_alloc_domain = ref->getMaybeAllocationDomain();
+  const std::vector<IterDomain*>& target_rfactor_domain =
+      target->getMaybeRFactorDomain();
+
+  // map target rfactor domain into ref's allocation domain
+  std::vector<IterDomain*> mapped_id_vec;
+  std::unordered_set<IterDomain*> mapped_id_set;
+
+  // logic to preserve reduction iter domain in target to WAR issue #2202
```
naoyam, line 84: Is this going to be removed once #2202 is addressed?

csrc/preseg_passes/allocation_order_inference.cpp
wujingyue: > we requires all tie to have identical mapping

Got it but that's subject to change, right? If/when that happens, I'm afraid nobody will remember to change this loop to be deterministic :) 

Also, given loop_iter_count is only used once, you may want to countLoopIterDomains ionside the loop. 

```
for (TensorView* src : srcs) {
  if (src has self mapping) {
    continue;
  }
  int64_t loop_iter_count = countLoopIterDomains(...);
  ...
}
```

Side note: I don't think it'll be useful here but structured binding can avoid having to write `iter.first` and `iter.second`, which are not named or typed well. E.g. 

```
for (const auto& [src,  count] : loop_iter_count) {
```

csrc/preseg_passes/allocation_order_inference.cpp
jjsjann123: yes that's the plan.

csrc/preseg_passes/allocation_order_inference.cpp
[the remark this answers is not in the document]
jjsjann123: This is just counting `iter domain` that's not a broadcast nor a reduction, so not specific to allocation domain. But I agree with you that it's a bad name...

Should I just switch to `NonTrivialIterDomain` until we came up with a better name for these?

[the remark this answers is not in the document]
naoyam: Yes, non-trivial sounds less confusing to me.

csrc/preseg_passes/allocation_order_inference.cpp
jjsjann123: I think @jacobhinkle 's reasoning here is that we should try to establish many-to-many mapping and a order of ids in the many-to-many mapping and propagate that through the graph.

That's more than what I was thinking, but sounds like a reasonable suggestion as well.

jjsjann123:
!bulid --diff

jjsjann123:
!build --diff

jjsjann123:
jit_thunder_tests on A100 seems to be reporting kernel with wrong arch.
`00:00:36 FATAL: kernel `fmha_cutlassF_f32_aligned_64x64_rf_sm75` is for sm75-sm80, but was built for sm70`

Don't think that one is coming from this PR and I don't see a CI nightly with thunder failure. cc'ing @xwang233

xwang233:
> jit_thunder_tests on A100 seems to be reporting kernel with wrong arch. `00:00:36 FATAL: kernel `fmha_cutlassF_f32_aligned_64x64_rf_sm75` is for sm75-sm80, but was built for sm70`
> 
> Don't think that one is coming from this PR and I don't see a CI nightly with thunder failure. cc'ing @xwang233

Somehow that test job got a T4 GPU instead of A100. Need to investigate that.

csrc/preseg_passes/allocation_order_inference.cpp
jjsjann123: Recording the alternatives in issue #2235 .

jjsjann123:
!build

jjsjann123:
I addressed all issues in comment. I'll be merging the PR after CI clears again since I have already got a stamp earlier.

Please do block merge if you have further concerns. Regarding comments on propagation rule for reshape. Let's move the discussion here #2235 .

jjsjann123:
Thunder failure isn't related. I'm merging this.

PR #2170, opened 2024-05-01
allow iteration grouped reduction for IterType::GatherScatter

liqiangxl:
when a tensor is created by `torch_gather`, its iter domains are markd as `IterType::GatherScatter`, the current iteration grouped reduction can group the `IterType::GatherScatter` domain similar to regular `IterType::Iteration`.


liqiangxl:
!build

liqiangxl:
this old friend... `tests.python.pytest_ops.test_correctness_div_complex64`

naoyam (approved):
LGTM

csrc/device_lower/validation.cpp
```diff
@@ -1137,9 +1137,11 @@ void validateAndConvertIterDomainGrouping(Fusion* fusion) {
       // Remember if a grouped ID is found
       is_grouped = true;
 
-      // Grouping only makes sense for the normal iteration type
+      // Grouping only makes sense for the normal iteration or gather scatter
+      // type
       NVF_CHECK(
-          id->getIterType() == IterType::Iteration,
+          id->getIterType() == IterType::Iteration ||
+              id->getIterType() == IterType::GatherScatter,
```
jjsjann123, line 1144: naive question:

Before the code change, we are just going to hit a negative assert here. Does this mean we might as well check that the itertype is compatible when we assign ParallelType::Group to an id?

jjsjann123 (approved):
:+1:

csrc/device_lower/validation.cpp
liqiangxl: that's the job of `void IterDomain::parallelize(ParallelType t)`. Before this PR, the error msg comes from there.

jjsjann123: Got you. I didn't realize IterDomain::parallelize is defined inside `csrc/ir/nodes.cpp`. :face_with_head_bandage:

PR #2172, opened 2024-05-01
Separate the lowering and the execution of a communication.

wujingyue:
This PR prepares for integrating Communication/Communicator into FusionExecutor. The lowering of a communication will be done by `FusionExecutor::compileFusion`, which is done <= once per shape and where sometimes only meta tensors are available. The execution will be done by `::runFusion`, which is on the critical path.

### DIFF
--- a/csrc/multidevice/communication.cpp
+++ b/csrc/multidevice/communication.cpp
@@ -27,7 +27,7 @@ void post_common(Communication& self, Communicator& comm) {
 }
 
 inline void doLocalCopy(const at::Tensor& dst, const at::Tensor& src) {
-  dst.copy_(src, /*non_blocking=*/true);
+  dst.view_as(src).copy_(src, /*non_blocking=*/true);
 }
 
 template <typename T>
@@ -268,7 +268,11 @@ c10::intrusive_ptr<c10d::Work> ReduceScatter::post(
 
   std::vector<std::vector<at::Tensor>> input_tensors(1);
   // FIXME: use split
-  for (auto i : c10::irange(params_.team.size())) {
+  int64_t team_size = params_.team.size();
+  int64_t scattered_dim_size = input_tensor.size(params_.scattered_axis);
+  NVF_ERROR(
+      scattered_dim_size == team_size, scattered_dim_size, " vs ", team_size);
+  for (auto i : c10::irange(team_size)) {
     input_tensors[0].push_back(
         input_tensor.slice(params_.scattered_axis, i, i + 1));
   }
--- a/csrc/multidevice/communication.cpp
+++ b/csrc/multidevice/communication.cpp
@@ -14,7 +14,6 @@
 namespace nvfuser {
 namespace {
 
-// FIXME: rename this.
 void post_common(Communication& self, Communicator& comm) {
   NVF_ERROR(
       std::find(
@@ -105,7 +104,9 @@ c10::intrusive_ptr<c10d::Work> Broadcast::post(
 
   if (comm.deviceId() == params_.root) {
     if (params_.is_root_in_mesh) {
-      // Do a local copy and the subsequent broadcast will be in place.
+      // Do a local copy and the subsequent broadcast will be in place. Consider
+      // ProcessGroupNCCL::_broadcast_oop so ncclBroadcast doesn't wait for the
+      // local copy to complete.
       doLocalCopy(output_tensor, input_tensor);
     } else {
       // `output_tensor` isn't allocated for this device.
--- a/csrc/multidevice/communication.cpp
+++ b/csrc/multidevice/communication.cpp
@@ -14,6 +14,34 @@
 namespace nvfuser {
 namespace {
 
+inline void assertBufferCount(
+    const std::vector<at::Tensor>& bufs,
+    size_t count) {
+  NVF_ERROR(
+      bufs.size() == count,
+      "there must be ",
+      count,
+      " buffer(s), but ",
+      bufs.size(),
+      " were given");
+}
+
+inline void assertBuffersHaveSameSize(
+    const std::vector<at::Tensor>& bufs1,
+    const std::vector<at::Tensor>& bufs2) {
+  if (bufs1.empty() && bufs2.empty()) {
+    return;
+  }
+  const auto numel = (bufs1.empty() ? bufs2 : bufs1).at(0).numel();
+  for (const auto& bufs : {bufs1, bufs2}) {
+    for (const auto& buf : bufs) {
+      NVF_ERROR(
+          buf.numel() == numel,
+          "all buffers must have the same number of elements");
+    }
+  }
+}
+
 void post_common(Communication& self, Communicator& comm) {
   NVF_ERROR(
       std::find(
@@ -133,6 +161,12 @@ c10::intrusive_ptr<c10d::Work> Gather::post(
   post_common(*this, comm);
 
   if (comm.deviceId() == params_.root && !params_.is_root_in_mesh) {
+    // This is likely a suboptimal way to allocate tensors for nccl. To benefit
+    // from zero copy
+    // (https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/bufferreg.html),
+    // tensors for nccl should be `ncclMemAlloc`ed and be `ncclCommRegister`ed.
+    // https://github.com/pytorch/pytorch/issues/124807 is one proposal trying
+    // to partially address this problem.
     input_tensor = at::empty_like(output_tensor.slice(0, 0, 1));
   }
   std::vector<at::Tensor> input_tensors({input_tensor});
@@ -144,14 +178,16 @@ c10::intrusive_ptr<c10d::Work> Gather::post(
     for (auto i : c10::irange(params_.team.size())) {
       if (root_relative_index_ == static_cast<DeviceIdxType>(i) &&
           !params_.is_root_in_mesh) {
-        output_tensors.front().push_back(input_tensor);
+        output_tensors[0].push_back(input_tensor);
         continue;
       }
-      output_tensors.front().push_back(output_tensor.slice(0, j, j + 1));
+      output_tensors[0].push_back(output_tensor.slice(0, j, j + 1));
       j++;
     }
   }
 
+  assertBufferCount(output_tensors[0], params_.team.size());
+  assertBuffersHaveSameSize(input_tensors, output_tensors[0]);
   return comm.getBackendForTeam(params_.team, backend)
       ->gather(
           output_tensors, input_tensors, {.rootRank = root_relative_index_});
@@ -168,12 +204,11 @@ c10::intrusive_ptr<c10d::Work> Allgather::post(
   post_common(*this, comm);
 
   std::vector<at::Tensor> input_tensors({input_tensor});
-
-  // FIXME: use split
   std::vector<std::vector<at::Tensor>> output_tensors(1);
-  for (auto i : c10::irange(params_.team.size())) {
-    output_tensors[0].push_back(output_tensor.slice(0, i, i + 1));
-  }
+  output_tensors[0] = at::split(output_tensor, /*split_size=*/1, /*dim=*/0);
+
+  assertBufferCount(output_tensors[0], params_.team.size());
+  assertBuffersHaveSameSize(input_tensors, output_tensors[0]);
   return comm.getBackendForTeam(params_.team, backend)
       ->allgather(output_tensors, input_tensors, {});
 }
@@ -207,6 +242,8 @@ c10::intrusive_ptr<c10d::Work> Scatter::post(
     }
   }
 
+  assertBufferCount(input_tensors[0], params_.team.size());
+  assertBuffersHaveSameSize(input_tensors[0], output_tensors);
   return comm.getBackendForTeam(params_.team, backend)
       ->scatter(
           output_tensors, input_tensors, {.rootRank = root_relative_index_});
@@ -273,21 +310,16 @@ c10::intrusive_ptr<c10d::Work> ReduceScatter::post(
   post_common(*this, comm);
 
   std::vector<std::vector<at::Tensor>> input_tensors(1);
-  // FIXME: use split
   NVF_ERROR(
       params_.scattered_axis >= 0,
       "scattered_axis is expected to be non-negative: ",
       params_.scattered_axis)
-  const auto scattered_dim_size = input_tensor.size(params_.scattered_axis);
-  const auto team_size = static_cast<int64_t>(params_.team.size());
-  NVF_ERROR(
-      scattered_dim_size == team_size, scattered_dim_size, " vs ", team_size);
-  for (auto i : c10::irange(team_size)) {
-    input_tensors[0].push_back(
-        input_tensor.slice(params_.scattered_axis, i, i + 1));
-  }
+  input_tensors[0] =
+      at::split(input_tensor, /*split_size=*/1, /*dim=*/params_.scattered_axis);
 
   std::vector<at::Tensor> output_tensors({output_tensor});
+
+  assertBufferCount(input_tensors[0], params_.team.size());
   return comm.getBackendForTeam(params_.team, backend)
       ->reduce_scatter(
           output_tensors, input_tensors, {.reduceOp = params_.redOp});
--- a/csrc/multidevice/communication.cpp
+++ b/csrc/multidevice/communication.cpp
@@ -14,33 +14,8 @@
 namespace nvfuser {
 namespace {
 
-inline void assertBufferCount(
-    const std::vector<at::Tensor>& bufs,
-    size_t count) {
-  NVF_ERROR(
-      bufs.size() == count,
-      "there must be ",
-      count,
-      " buffer(s), but ",
-      bufs.size(),
-      " were given");
-}
-
-inline void assertBuffersHaveSameSize(
-    const std::vector<at::Tensor>& bufs1,
-    const std::vector<at::Tensor>& bufs2) {
-  if (bufs1.empty() && bufs2.empty()) {
-    return;
-  }
-  auto sizes = (bufs1.empty() ? bufs2 : bufs1).at(0).sizes();
-  for (auto& bufs : {bufs1, bufs2}) {
-    for (auto& buf : bufs) {
-      NVF_ERROR(buf.sizes() == sizes, "all buffers must have the same size");
-    }
-  }
-}
-
-inline void post_common(Communication& self, Communicator& comm) {
+// FIXME: rename this.
+void post_common(Communication& self, Communicator& comm) {
   NVF_ERROR(
       std::find(
           self.params().team.begin(),
@@ -52,7 +27,25 @@ inline void post_common(Communication& self, Communicator& comm) {
 }
 
 inline void doLocalCopy(const at::Tensor& dst, const at::Tensor& src) {
-  dst.copy_(src, /* non-blocking */ true);
+  dst.copy_(src, /*non_blocking=*/true);
+}
+
+template <typename T>
+T getInitialValue(c10d::ReduceOp::RedOpType op) {
+  // TODO: add other ops
+  switch (op) {
+    case c10d::ReduceOp::RedOpType::SUM:
+      return 0;
+    case c10d::ReduceOp::RedOpType::PRODUCT:
+      return 1;
+    case c10d::ReduceOp::RedOpType::MAX:
+      return std::numeric_limits<T>::min();
+    case c10d::ReduceOp::RedOpType::MIN:
+      return std::numeric_limits<T>::max();
+    default:
+      NVF_ERROR(false, "unsupported reduction op type");
+      return 0;
+  }
 }
 
 } // namespace
@@ -61,7 +54,6 @@ Communication::Communication(CommParams params, std::string name, bool has_root)
     : params_(std::move(params)),
       collective_type_(std::move(name)),
       has_root_(has_root) {
-  assertBuffersHaveSameSize(params_.src_bufs, params_.dst_bufs);
   NVF_ERROR(
       std::unique(params_.team.begin(), params_.team.end()) ==
           params_.team.end(),
@@ -97,11 +89,6 @@ std::string Communication::toString(int indent) const {
     ss << r << ", ";
   }
   ss << indent1 << "}\n";
-  ss << indent1 << "src_bufs: {";
-  for (auto& t : params_.src_bufs) {
-    ss << "\n" << t;
-  }
-  ss << "\n" << indent1 << "}\n";
   ss << ext_indent << "}";
 
   return ss.str();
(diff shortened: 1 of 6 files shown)


wujingyue:
!build --dist

wujingyue:
!build --dist

wujingyue:
!build --dist

wujingyue:
!build --dist

csrc/multidevice/communication.cpp
```diff
@@ -111,209 +98,225 @@ Broadcast::Broadcast(CommParams params) : Communication(params, "broadcast") {}
...
-                  ->allgather(buf_list, params_.src_bufs, {});
-  params_.dst_bufs = std::move(buf_list.back());
-  return work;
-}
 
-Scatter::Scatter(CommParams params) : Communication(params, "scatter") {
-  assertBufferCount(params_.dst_bufs, 1);
+  std::vector<at::Tensor> input_tensors({input_tensor});
+
+  // FIXME: use split
```
samnordmann, line 171: what do you mean by "use split"? (for my own understanding)

csrc/multidevice/communication.cpp
```diff
@@ -111,209 +98,225 @@ Broadcast::Broadcast(CommParams params) : Communication(params, "broadcast") {}
...
+    tensor = input_tensor;
   }
-  post_common(*this, comm);
-  auto& buf =
-      (comm.deviceId() == params_.root) ? params_.dst_bufs : params_.src_bufs;
+  std::vector<at::Tensor> tensors({tensor});
+
   c10d::ReduceOptions options = {
       .reduceOp = params_.redOp, .rootRank = root_relative_index_};
-  auto team_backend = comm.getBackendForTeam(params_.team, backend);
```
samnordmann, line 231: why removing this?

csrc/multidevice/communication.cpp
```diff
@@ -14,33 +14,8 @@
...
-  NVF_ERROR(
-      bufs.size() == count,
-      "there must be ",
-      count,
-      " buffer(s), but ",
-      bufs.size(),
-      " were given");
-}
-
-inline void assertBuffersHaveSameSize(
```
samnordmann, line 29: as discussed orally we should keep these checks

csrc/multidevice/communication.cpp
```diff
@@ -111,209 +98,225 @@ Broadcast::Broadcast(CommParams params) : Communication(params, "broadcast") {}
...
-    assertBufferCount(params_.src_bufs, 0);
-    assertBufferCount(params_.dst_bufs, 1);
+  if (params_.team.size() == 1) {
+    doLocalCopy(output_tensor, input_tensor);
+    return nullptr;
   }
 
+  std::vector<at::Tensor> tensors(
+      {comm.deviceId() == params_.root ? input_tensor : output_tensor});
   return comm.sendRecv(
```
samnordmann, line 315: the method `sendRecv` should eventually be removed from the class `Communicator`, as done in the other PR https://github.com/NVIDIA/Fuser/pull/2185

csrc/multidevice/communication.h
```diff
@@ -15,43 +15,39 @@
...
 struct CommParams {
   DeviceIdxType root = -1;
-  std::vector<at::Tensor> src_bufs;
-  std::vector<at::Tensor> dst_bufs;
-  Team team; // should not have duplicate
+  bool is_root_in_mesh = true;
+  Team team; // should not have duplicates and should contain both the root and
+             // the mesh
   c10d::ReduceOp::RedOpType redOp = c10d::ReduceOp::RedOpType::UNUSED;
+  int64_t scattered_axis = -1;
```
samnordmann, line 26: It looks a bit "inconsistent" to have `scattered axis` here.

1) Why Reduce scatter needs to treated differently than other collecitves?
2) It seems to be implicitely assumed here that, for all collective, only the outer most axis can be communicated. On the one hand, this is true for now, but this will change soon. On the other hand, if we assume it for all collective, we should  consistently make the assumption for reduce scatter that the scattered axis is the second axis
3) The struct field name should reflect that it only deals with reduce scatter.
4) In general the "Communication" IR should take the TensorView pointers as input. The sharding information is all embedded there through the axis parallel types 

I understand though that some of the above comments are not due to this pr and apply to the current code.
I'm all in for simple temporary solution, so the current one might be good enough -- but I'm happy to discuss in order to align on the direction

csrc/multidevice/communication.h
```diff
@@ -15,43 +15,39 @@
...
-non-blocking. The communication can be posted multiple times.
-It is assumed that the current device_index (given by
-communicator.deviceId()) belongs to the team of the communication,
-otherwise an error is thrown.
-
-NOTE: pytorch's NCCL process group API needs <team_size> buffers on root for
-scatter/gather operation.
-*/
-
+// The class "Communication" represents a MPI-style communication
```
samnordmann, line 29: what motivates this change? The multiline comment is nicer imo, and used in other places of the codebase, e.g., https://github.com/NVIDIA/Fuser/blob/d12b786dd7782c0ea5d087c36fd0497ad92ff7dd/csrc/ir/base_nodes.h#L31

Also, did you change the content or only replace the multi-line comments with the inline comments?

csrc/multidevice/communication.h
```diff
@@ -81,6 +79,7 @@ class Communication {
  private:
   // used for printing
   std::string collective_type_;
+  // FIXME: this seems to be redundant with `root_relative_index_`.
```
samnordmann, line 82: you're right, however, conceptually, "has_root_" is a compile time constant (at the program level, IOW, it is a `constexpr`), whereas `root_relative_index_` is a runtime variable

csrc/multidevice/communication.h
```diff
@@ -15,43 +15,39 @@
...
-  construction a communication
-*/
+// This struct gathers all the parameters necessary for the
+// construction a communication
 struct CommParams {
   DeviceIdxType root = -1;
-  std::vector<at::Tensor> src_bufs;
-  std::vector<at::Tensor> dst_bufs;
-  Team team; // should not have duplicate
+  bool is_root_in_mesh = true;
```
samnordmann, line 22: this shouldn't be a param's field that can be set, this field is entirely deduced from `root` and `team` members. There shouldn't be such a member in the struct

csrc/multidevice/lower_communication.cpp
```diff
@@ -295,43 +204,24 @@ CommParams createParamsForReduce(
...
-    params.dst_bufs = {output_tensor};
-    // The reduce semantics imposes the root to be both
-    // sender and receiver. If the root is not in the mesh, we thus
-    // have to artificially make it send and receive a dummy buffer
-    if (!is_root_in_mesh) {
-      at::Tensor dummy = createDummyTensor(output_tensor, op_type);
-      params.src_bufs.push_back(dummy);
-    }
-  }
+  // FIXME: we may want to store sharded_dim to params for speed.
```
samnordmann, line 217: I am not sure to understand this FIXME, can you please explain?
In general, the `Communication` class should store pointers to I/O TensorView, from which all the sharding info can be infered

tests/cpp/test_multidevice_communications.cpp
```diff
@@ -23,11 +23,11 @@ class CommunicationTest
   void SetUp() override;
 
   void validate(at::Tensor obtained, at::Tensor expected);
-  void resetDstBuffers();
 
   static constexpr DeviceIdxType root = 0;
   static constexpr int tensor_size = 1024;
   static constexpr int number_of_repetitions = 8;
+  // TODO: test other reduction op types using subTest.
```
samnordmann, line 30: what is subTest? (for my own understanding)

tests/cpp/test_multidevice_communications.cpp
```diff
@@ -58,121 +58,96 @@ void CommunicationTest::validate(at::Tensor obtained, at::Tensor expected) {
...
   params.team = all_ranks;
-  params.src_bufs = {at::empty(tensor_size, tensor_options)};
-  if (communicator->deviceId() == root) {
-    for (int64_t i = 0; i < communicator->size(); i++) {
-      params.dst_bufs.push_back(at::empty(tensor_size, tensor_options));
-    }
-  }
   auto communication = Gather(params);
 
-  for (int j : c10::irange(number_of_repetitions)) {
```
samnordmann, line 78: I would personally revert this change
The implementation was made so that we keep the same src/dst buffers when repeating the collectives. This has the following advantages:
- We test/benefit  backend's buffer registration cache. This is essential for benchmarks
- We save allocation time when testing
- The tests are "harden" this wew really test that we can have many inflights collectives on the same buffers. This emulates more accurately the type of Workload we are targeting

csrc/multidevice/communication.h
wujingyue: >  this field is entirely deduced from root and team members

`team` always contains `root` IIUC. Do you mean to let params store root and **mesh** instead?

csrc/multidevice/lower_communication.cpp
wujingyue: My bad -- this FIXME is outdated and should be removed. 

> In general, the Communication class should store pointers to I/O TensorView, from which all the sharding info can be infered

TensorView is available at compileFusion time, so there's a benefit of precompute things over there. However, I haven't timed this so I'm happy to revisit the decision later.

tests/cpp/test_multidevice_communications.cpp
wujingyue: Makes sense!

csrc/multidevice/communication.cpp
```diff
@@ -111,209 +98,225 @@ Broadcast::Broadcast(CommParams params) : Communication(params, "broadcast") {}
...
-  // ProcessGroup API
-  std::vector<std::vector<at::Tensor>> buf_list;
-  if (comm.deviceId() == params_.root) {
-    assertBufferCount(params_.dst_bufs, params_.team.size());
-    buf_list = {std::move(params_.dst_bufs)};
-  } else {
-    assertBufferCount(params_.dst_bufs, 0);
+
+  if (comm.deviceId() == params_.root && !params_.is_root_in_mesh) {
+    input_tensor = at::empty_like(output_tensor.slice(0, 0, 1));
```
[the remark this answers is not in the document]
wujingyue: Good question! 

1. Allocation is done via PyTorch's cached allocator, so it might not be as expensive as many people thought. At this moment, I've never seen it stick out in real traces. If it indeed sticks out in the future, it will probably be a problem for single-device as well and I'll hope to solve the bigger problem. For example, we could pre-allocate for the max sizes and place some restrictions on dynamic shapes. 
2. With Jie's work on dynamic shapes, shapes won't always be available during the "compile" time. 
3. IIUC, is_root_in_mesh==false in Gather (and several other places in this file) is rare in real workloads. Do you remember why we added them in the first place other than for completeness?

csrc/multidevice/communication.h
wujingyue: I had the same question, but I chose to not pulling too many changes in one PR :) 

I got the impression that there's a nccl/ProcessGroup limitation so the scattered axis has to be outermost so the scattered axis can simply be hardcoded? 

If that limitation doesn't exist or isn't fundamental, then should we support non-outermost reduce axis (which is currently hardcoded to be 0)?

csrc/multidevice/communication.h
wujingyue: To make style consistent in one codebase. Yes, I'm sure `/* */` is used in some other files but they appear much less often than `//` when I search both in our codebase. Do you have a technical problem (e.g. IDE couldn't line wrap correctly) with `//`? 

> Also, did you change the content or only replace the multi-line comments with the inline comments?

No. Sorry, I should have called that out to save you some review time.

csrc/multidevice/communication.cpp
wujingyue: `at::split` instead of a for-loop of `at::slice`. But I'll be aware of your other comment on device order in mesh (I haven't got a chance to look further yet).

csrc/multidevice/communication.h
cowanmeg: The scattered axis just has to be the outermost _allocated_ axis. Since the reduction axis isn't allocated in the output, the scattered axis can be either 0 or 1 depending on the reduction axis's placement. 

We could force it to always be the outermost axis in the RFactor domain, but that introduces unnecessary permutes

csrc/multidevice/communication.cpp
```diff
@@ -111,209 +98,225 @@ Broadcast::Broadcast(CommParams params) : Communication(params, "broadcast") {}
...
-  auto work =
-      comm.getBackendForTeam(params_.team, backend)
-          ->scatter(
-              params_.dst_bufs, buf_list, {.rootRank = root_relative_index_});
+  std::vector<at::Tensor> output_tensors({output_tensor});
+
+  std::vector<std::vector<at::Tensor>> input_tensors;
   if (comm.deviceId() == params_.root) {
-    params_.src_bufs = std::move(buf_list.back());
+    input_tensors.resize(1);
```
cowanmeg, line 196: Why is input_tensor sized to 1 instead of `params_.team.size()`?

wujingyue: `input_tensors` passed to `Backend::scatter` is a **vector of vectors** of tensors. In our current code, the first dimension is always 1. 

That being said, do you happen to know the reason behind that ProcessGroup API? Is that to allow multiple gathers to run in parallel? But I'd expect that to have to take multiple streams...

cowanmeg: Right....that argument decision confuses me everytime I see it...
I have no idea why they chose that? Maybe it's to merge multiple small scatter calls into a big one?

wujingyue: https://github.com/pytorch/pytorch/blob/abdd569e412c4af5957b10ff8039511f8175986e/torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp#L3381 is interesting... 

🤷

csrc/multidevice/communication.cpp
```diff
@@ -111,209 +98,225 @@ Broadcast::Broadcast(CommParams params) : Communication(params, "broadcast") {}
...
-          ->gather(
-              buf_list, params_.src_bufs, {.rootRank = root_relative_index_});
+  std::vector<at::Tensor> input_tensors({input_tensor});
+
+  std::vector<std::vector<at::Tensor>> output_tensors;
   if (comm.deviceId() == params_.root) {
-    params_.dst_bufs = std::move(buf_list.back());
+    output_tensors.resize(1);
+    int64_t j = 0;
+    for (auto i : c10::irange(params_.team.size())) {
```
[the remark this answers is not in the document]
wujingyue: I think you are right, and I'm surprised https://github.com/NVIDIA/Fuser/blob/4e8356915fc6ffa823cf97fed4249cc00f317fdf/tests/cpp/test_multidevice_pipeline.cpp#L218 doesn't fail. 

That being said, this appears to be a pre-existing issue according to https://github.com/NVIDIA/Fuser/blob/4e8356915fc6ffa823cf97fed4249cc00f317fdf/csrc/multidevice/lower_communication.cpp#L121. Am I right?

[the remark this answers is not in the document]
cowanmeg: I think the current code works correctly. The iteration is over the device mesh index so even if the mesh is {1,0} it will be ok! Perhaps just renaming i to mesh_idx will clear things up?

csrc/multidevice/communication.h
wujingyue: I'll remove this FIXME. `has_root_` will be gone with #2185.

csrc/multidevice/communication.cpp
```diff
@@ -111,209 +98,225 @@ Broadcast::Broadcast(CommParams params) : Communication(params, "broadcast") {}
...
+    at::Tensor output_tensor,
     std::optional<CommunicatorBackend> backend) {
   post_common(*this, comm);
 
   if (comm.deviceId() == params_.root) {
-    assertBufferCount(params_.src_bufs, 1);
-    if (params_.dst_bufs.size() == 1) {
-      doLocalCopy(params_.dst_bufs.at(0), params_.src_bufs.at(0));
+    if (params_.is_root_in_mesh) {
+      // Do a local copy and the subsequent broadcast will be in place.
```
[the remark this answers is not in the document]
wujingyue: I don't think I've changed the behavior, but correct me if I'm wrong. 

> I think we shouldn't handle the copy ourselves in this case

I'm happy to learn how to avoid the copy with the current c10d::Backend::broadcast API. Am I missing a broadcast_oop somewhere?

csrc/multidevice/communication.cpp
wujingyue: I'll add it back in a follow-up PR. I found that removing reduce_oop helped me a lot understand the code, so my first version was made that way. However, I do understand the speed benefit of using OOP reduction when it's available.

tests/cpp/test_multidevice_communications.cpp
samnordmann: > Done -- the tests now reuse input/output buffers.

Thx!


> I'd love to learn more about that.

Buffer registration can come into play in several places. For example, in low level [RDMA infiniband verbs](https://www.rdmamojo.com/2012/09/07/ibv_reg_mr/), the buffers needs to be registered to the NIC before rdma can be posted. E.g., UCX (p2p library used by ucc and possibly nccl) uses such [registration cache](https://github.com/openucx/ucx/blob/e11197ce7e6d0d892a9f96300c789fa036aad6c8/docs/source/glossaries.rst?plain=1#L69)

> According to https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/bufferreg.html, there is a zero-copy speed benefit when the application uses ncclMemAlloc'ed buffers. Is "buffer registration cache" made for that?

I think this case is different. It is related to switch-based reduction, and ncclMemAlloc guarantees some memory alignment on the buffers (and probably also register them).

> Which "backend" exactly does the "caching", ProcessGroup or NCCL?

NCCL or UCC, or even lower level. There is no such cache at the PG level that I'm aware of.

samnordmann (commented):
Looks good, thx! I only left minor comments

csrc/multidevice/communication.h
wujingyue: I'll think about it. One benefit of separating compilation from execution (and thus PR) is to avoid doing too much at runFusion time, e.g., having to analyze TVs and extract information needed for execution.

tests/cpp/test_multidevice_communications.cpp
wujingyue: I'm not as an expert as you about RDMA and NIC. What's the implication of "buffer registration cache" for nvFuser or any DL compiler that tries to efficiently transfer data between multiple GPUs? 

I believe we already agree that using ncclMemAlloc'ed buffers or using slices within a larger ncclMemAlloc'ed region with proper alignment will speed things up. nvFuser has done this yet because it uses `at::empty`, but I believe this has to be solved. 

In addition to that, do you see extra benefit (and if so how much) of reusing the same base address and size iteation-to-iteration for an nccl call like allgather? The name "buffer registration cache" seems to say yes, but I'm not sure.

csrc/multidevice/communication.cpp
```diff
@@ -111,209 +98,225 @@ Broadcast::Broadcast(CommParams params) : Communication(params, "broadcast") {}
...
-          {.rootRank = root_relative_index_});
+      ->broadcast(tensors, {.rootRank = root_relative_index_});
 }
 
-Gather::Gather(CommParams params) : Communication(params, "gather") {
-  assertBufferCount(params_.src_bufs, 1);
-}
+Gather::Gather(CommParams params) : Communication(params, "gather") {}
 
 c10::intrusive_ptr<c10d::Work> Gather::post(
```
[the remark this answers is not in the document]
wujingyue: I'm confused. Something has to call PG with concrete `at::Tensor`s at run time. Who will be that caller if not Collective::post?

csrc/multidevice/communication.cpp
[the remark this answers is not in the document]
wujingyue: > Imagine we execute the same fusion in a loop, then there is no need to reallocate the buffers at each iteration

and the same input shape, yes. I think https://github.com/pytorch/pytorch/issues/124807 (driven by NVIDIA folks) is trying to solve that without requiring too many changes in DL compilers like nvFuser. 

> where do we need to allocate in nvFuser workflow?

There's also allocation of intermediate global buffers: https://github.com/NVIDIA/Fuser/blob/2dbf1d0a36cdd947f2f65094f7713e3fefba0329/csrc/executor.cpp#L1994-L2018

Anyhow, the problem I was trying to say is that nvFuser currently allocates for the exact shape but we will want to reuse the same compiled kernel for different shapes. Because of this gap, if we want the allocation (even though via the caching allocator) to only occur once, we'll have to either reuse the same allocation for different input shapes (e.g. by allocating for the max) or use a different caching granularity for allocation than for kernels. Either way comes with its limitations, but these tradeoffs should definitely be considered.

tests/cpp/test_multidevice_communications.cpp
wujingyue: FYI, @Aidyn-A pointed me to a [proposal](https://github.com/pytorch/pytorch/issues/124807) that he and his teammates are working on, which I think will help solve the problem without too many changes to nvFuser's allocation mechanism. But let's revisit this topic once more progress is made there.

cowanmeg (approved):
Overall, I think the PR moves in the right direction separating compile time and run time logic. We should discuss offline allocation optimizations, especially in relation to whether it can be pre-computed and relations with NCCL user buffers, etc, but that is well beyond the scope of the PR!

I think beside cleaning up the fixme and adding the buffer size checks to post, I am good with these changes!

cowanmeg:
> Overall, I think the PR moves in the right direction separating compile time and run time logic. We should discuss offline allocation optimizations, especially in relation to whether it can be pre-computed and relations with NCCL user buffers, etc, but that is well beyond the scope of the PR!
> 
> I think beside cleaning up the fixme and adding the buffer size checks to post, I am good with these changes!

I also am unsure which PR is easier to merge first, this one or https://github.com/NVIDIA/Fuser/pull/2185

csrc/multidevice/communication.cpp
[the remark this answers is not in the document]
samnordmann: > Nice! Is [in-place broadcast](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/api/colls.html#ncclbroadcast) still preferred to out-of-place when we don't need to local copy?

As far as NCCL is concerned, the answer is "no" according to that comment saying that [inplace bcast is deprecated](https://github.com/NVIDIA/nccl/blob/ab2b89c4c339bd7f816fbc114a4b05d386b66290/src/collectives.cc#L75) (because it duplicates functionality). But in practice in shouldn't make any difference.

For UCC, there is no such an "out-of-place broadcast" (because this primitive doesn't exists in MPI standard), so we'll need to stick to (inplace) broadcast anyway.

> Currently, input_tensor and output_tensor have different buffers, but we can change nvFuser to make them alias by calling `Fusion::aliasOutputToInput`, if the input is not used elsewhere e.g. saved for backprop.

I see, yes that could be very interesting.
BTw, I'd love to know more about the status of Aliasing. Can we Alias any two given intermediate tensors, or does it need to be Inputs and outputs? also, is there an IR symbolizing aliasing relationship, or do we plan to add such an IR?


Im resolving this conversation since it is not related to the pr anymore

csrc/multidevice/communication.cpp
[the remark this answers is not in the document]
samnordmann: Thanks for the references and explanations!

> and the same input shape, yes. I think https://github.com/pytorch/pytorch/issues/124807 (driven by NVIDIA folks) is trying to solve that without requiring too many changes in DL compilers like nvFuser.

nice

> There's also allocation of intermediate global buffers:

I see, thanks!

> Anyhow, the problem I was trying to say is that nvFuser currently allocates for the exact shape but we will want to reuse the same compiled kernel for different shapes. Because of this gap, if we want the allocation (even though via the caching allocator) to only occur once, we'll have to either reuse the same allocation for different input shapes (e.g. by allocating for the max) or use a different caching granularity for allocation than for kernels. Either way comes with its limitations, but these tradeoffs should definitely be considered.

got it!

csrc/multidevice/communication.h
samnordmann: > I'll think about it. One benefit of separating compilation from execution (and thus PR) is to avoid doing too much at runFusion time, e.g., having to analyze TVs and extract information needed for execution.

Agreed. And let's stick with that for this pr!

On the other hand (for the future) I also see motivation for storing the TV and Expr ptrs:
- it contains the full info 1) which is useful for printing e.g. IO tensors, and, 2) it can be needed in the future when the lowering becomes more fleshed up
- it is more clean/concise/structured way to encode the symbolic Communication

What we could do to achieve best of both world is to store (non-mutable) pointers to IO TensorViews and Expr and precompute as class private data whats needed at instantiation.

csrc/multidevice/communication.cpp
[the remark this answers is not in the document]
samnordmann: My point is, in terms of interface and code organization purely, the function `post` shouldn't be member of the class `Collective` imo, because in general we want to separate symbolic IRs and runtime execution.
That's mostly incidental and not so important. Also, this is already what I did in this other pr https://github.com/NVIDIA/Fuser/pull/2185. So let me resolve the conversation

samnordmann (approved):

csrc/multidevice/communication.h
wujingyue: With #2185 , `Communication` will be an `Expr` which automatically contains input and output TensorView. That should address your concern?

wujingyue: I gave up on the team=>mesh change for this PR. It would cause even more divergence with #2185 , so we'll do that in a separate PR if needed.

csrc/multidevice/communication.cpp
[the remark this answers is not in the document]
wujingyue: Got it!

wujingyue:
!build --dist

csrc/multidevice/communication.h
samnordmann: > With #2185 , `Communication` will be an `Expr` which automatically contains input and output TensorView. That should address your concern?

Yes! That would be a good way of doing it. If we follow this path, I think we can completely get rid of the struct `CommunicationParams` (we then only need to pass to the constructor the ReductionOpType, or a pointer to the set/reduction Expr*)

PR #2173, opened 2024-05-01
Globally enable IdModel

zasdfgbnm:
Per the title.

## Why?

I am looking at refactoring the indexing of TMA into two parts: one part that does the analysis to get TMA domain, tile IterDomain, etc., and another part that uses the result of part 1 to do indexing. This split allows us to do more advanced lowering, for example, implementing the strategy described at [TMA Modeling In Depth](https://github.com/NVIDIA/Fuser/blob/main/doc/reading/tma-modeling-in-depth.md).

When I am working on this refactor, I realize that I need to replay the producer as consumer for TMA load, so that I can propagate from gmem's allocation domain to consumer's TMA domain. `BestEffortReplay` could do the job of course, but considering the current status of `IdModel`, I believe it's better to just not use `BestEffortReplay` and base this on `IdModel`.

## Impact

This will increase compilation time for all fusions, I did not measure how much. But considering that we have to do this step to enable incremental development, I believe this PR would be fine.

Measured a single test:

Before:
```
[       OK ] GPUTTensorCoreTest.FusionAmpereMatmul_CUDA (3916 ms)
```

After:
```
[       OK ] GPUTTensorCoreTest.FusionAmpereMatmul_CUDA (3998 ms)
```

About 2% slow down.

### DIFF
--- a/csrc/device_lower/lower2device.cpp
+++ b/csrc/device_lower/lower2device.cpp
@@ -381,19 +381,16 @@ void GpuLower::analysis(Fusion* fusion) {
   replaceSymbolicSizes(fusion_);
   dumpExprsIfEnabled(fusion_->exprs(), "replaceSymbolicSizes");
 
-  // Build what's refered to as the compute at map. This map contains the
-  // mappings of all iteration domains across the fusion. There are three types
-  // of mappings Permissive, Exact, and Loop, see compute_at_map.h/cpp for more
-  // information.
+  // Build what's refered to as the IterDomain Model. This data structure
+  // contains the mappings of all iteration domains across the fusion, and their
+  // graph structure. There are multiple types of mappings: Permissive, Exact,
+  // Loop, etc. See the id_model directory for more information.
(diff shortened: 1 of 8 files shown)


zasdfgbnm:
!build

zasdfgbnm:
!build

zasdfgbnm:
!build

naoyam (approved):
Functionality wise I don't see any issue, so gave a stamp. The only concern is longer lowering time, which may cause CI timeouts more frequently. Let's wait for @Priya2698 to set up host-side benchmarking to understand its impact.

zasdfgbnm:
close in favor of https://github.com/NVIDIA/Fuser/pull/2389

naoyam:
#2389 still requires opt-in.

There are still failures with some fusions, which I need to look into. (For example, #2253)

zasdfgbnm:
> #2389 still requires opt-in

Oh, I see. But still, let's close this PR, because it is conflicting with main branch badly.

Issue #2149, opened 2024-04-26
Investigate adding IR Nodes for `linear` and `matmul`


kevinstephano:
It was noticed that there is a lot of complexity in pattern matching operations and determining if the set of pattern matched nodes equal a `linear` operation or `matmul` operation prior to asking Expression Evaluator to process the operation.  Therefore, one idea to explore is to add IR nodes for the operations and decompose them later when they specifically come from `ops.linear` and `ops.matmul`.

Priya2698:
PR #2240, #2209 and #2175 add the new IR nodes.

PR #2175, opened 2024-05-01
Add a new ATen Matmul IR node

Priya2698:
Issue #2149.

Adds a new MatmulOp IR node that has the same functionality as `torch.matmul`.

1. Both tensors are 1D: Dot product is returned (`sum(mul(a, b)`), without creating a `MatmulOp` node.
2. One of the tensors is 1D: `[M, K] x [K] -> [M] / [K] x [K, N] -> [N]`
3. Both tensors are 2D: `[M, K] x [K, N] -> [M, N]`
4. Both tensors are atleast 1D and one of the tensors is > 2D: `[B, M, K] x [K, N] -> [B, M, N] `

`csrc/ops/utils/mapMatmulOpIterDomains` defines the logic to map the input operands to the ouput. This is used to create the new output `TensorView` for the `MatmulOp` using the input iterdomains and in `PairwiseRootDomainMap` to accurately map the MatmulOp input/outputs. This is required since the inputs are no longer broadcasted affecting the alignment of the inputs with the output.


csrc/ops/arith.cpp
```diff
@@ -2684,4 +2684,57 @@ TensorView* tensor(Val* val) {
...
+
+  NVF_ERROR(ndims_a>=1 && ndims_b>= 1);
+
+  std::vector<IterDomain*> new_domain;
+  auto higher_dim_domain = ndims_a >= ndims_b ? orig_domain_a : orig_domain_b;  
+  new_domain.reserve(higher_dim_domain.size());
+
+  if (ndims_a > 2 || ndims_b > 2) {
+    for (auto inx: c10::irange(higher_dim_domain.size() - 2)) {
+      new_domain.push_back(IterDomainBuilder(higher_dim_domain[inx]).resetSchedulingParams().build());
```
Priya2698, line 2707: Q: We need to check broadcastability of inputs. I was doing it using a check on the extents -- either one of the extents is 1 or the extents are the same. This does not work for symbolic inputs? Do we rely on thunder to reject input sizes that are incompatible?

csrc/ops/arith.cpp
```diff
@@ -2684,4 +2684,57 @@ TensorView* tensor(Val* val) {
...
+
+  TensorDomain* td = IrBuilder::create<TensorDomain>(
+      new_domain, TensorDomain::getContiguityFilledWith(new_domain, true));
+
+  return IrBuilder::create<TensorView>(td, tv_a->dtype());
+}
+
+} // namespace
+
+TensorView* eagerMatmul(
```
Priya2698, line 2731: Q:  Named it `eagerMatmul` to avoid clash with existing `matmul`. What should we name these new APIs? Preferably a consistent naming scheme across IR node/ API / scheduler. For instance: prefix `ATen`/`eager`/`exprEval`.

csrc/root_domain_map.cpp
```diff
@@ -121,6 +121,34 @@ std::unordered_map<IterDomain*, IterDomain*> PairwiseRootDomainMap::map(
   const auto producer_root =
       TensorDomain::noReductions(producer->maybeRFactor());
   const auto& consumer_root = consumer->root();
+
+  // Map MatmulOp extents in reverse
+  if (consumer_tv_->definition()->isA<MatmulOp>()) {
```
Priya2698, line 126: Q: Do we need refactor this special handling in a separate override method, for instance?

csrc/scheduler/expr_eval_sched.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include <scheduler/registry_utils.h>
+
+namespace nvfuser {
+
+template <typename... Args>
+void vlog(const Args&... args) {
+  scheduler_debug_utils::log("[Expression Evaluator Scheduler] ", args...);
+}
+
+ExprEvalScheduler::ExprEvalScheduler(
```
Priya2698, line 21: Same question as above about naming scheme.

tests/cpp/test_matmul_aten_evaluation.cpp
```diff
@@ -393,4 +393,61 @@ TEST_F(MatmulATenEvaluationTest, LinearWithBias) {
 
   EXPECT_TRUE(at::allclose(out[0], out_ref));
 }
+
+TEST_F(MatmulATenEvaluationTest, MatmulNodeConcrete) {
+  auto fusion = std::make_unique<Fusion>();
+  FusionGuard fg(fusion.get());
+  std::vector<int64_t> a_shape{512, 32, 64}, b_shape{32, 1, 64, 128};
```
Priya2698, line 400: Q: Do you have an example of parametrizing over sets of vectors using `TEST_P`. I was not able to make that work to test multiple input shape combinations.

csrc/root_domain_map.cpp
```diff
@@ -121,6 +121,34 @@ std::unordered_map<IterDomain*, IterDomain*> PairwiseRootDomainMap::map(
...
+    int ritc = consumer_root.size() - 1;
+    int ritp = producer_root.size() - 1;
+
+    while (ritc >=0 && ritp >= 0) {
+      IterDomain* producer_id = producer_root.at(ritp);
+      IterDomain* consumer_id = consumer_root.at(ritc);
+      if (!producer_id->extent()->sameAs(consumer_id->extent())){
```
Priya2698, line 133: This is to avoid mapping K to either M/N in the consumer. Tests pass for symbolic shapes, however, how are extents checked for symbolic shapes? Have we already concretized at this point?

csrc/ops/arith.cpp
jacobhinkle: The output IterDomain should be symbolic in that case. Dynamic fusion concretization should set the IterType automatically. This should work pretty much just like how a binary op works.

csrc/ops/arith.cpp
jacobhinkle: I actually think it should just be `matmul`. We might translate these to nvfuser kernels after all...

jacobhinkle: What is the case for keeping the original `matmul` and `linear`?

Priya2698: There is an existing `matmul` here: https://github.com/NVIDIA/Fuser/blob/bfc8a1f2e4c39ba8e89ed1d4ee78b2940fbe27b3/csrc/ops/composite.cpp#L60. We can remove/rename that.

csrc/scheduler/expr_eval_sched.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include <scheduler/debug_utils.h>
+#include <scheduler/expr_eval_sched.h>
+#include <scheduler/registry_utils.h>
+
+namespace nvfuser {
+
+template <typename... Args>
+void vlog(const Args&... args) {
+  scheduler_debug_utils::log("[Expression Evaluator Scheduler] ", args...);
+}
```
jacobhinkle, line 19: I think you could use a variadic macro here. I dont think there's any functional difference so its up to you but I could be wrong.

tests/cpp/test_matmul_aten_evaluation.cpp
jacobhinkle: I do not but maybe the guru @wujingyue does.

jacobhinkle (commented):
First pass. Havent looked in detail at the mapping code yet.

tests/cpp/test_matmul_aten_evaluation.cpp
wujingyue: testing::Values?

Priya2698: Yes, I was trying something like `testing::Values ((a_shape1, b_shape1) , (a_shape2, b_shape2))` but was running into errors. I'm looking into the documentation to see if there's any restriction on value types.

wujingyue: I don't know what `(a_shape1, b_shape1)` means in C++. You might want to use `{}` or `std::make_pair` instead.

Priya2698: I am using `std::make_tuple{{b, m, k}, {k, n}}` for input values.

csrc/ir/internal_nodes.h
```diff
@@ -2255,4 +2255,37 @@ class NVF_API CatOp : public Expr {
...
+class NVF_API MatmulOp : public Expr {
+ public:
+  using Expr::Expr;
+
+  MatmulOp(IrBuilderPasskey, Val* out, Val* in_a, Val* in_b);
+
+  NVFUSER_DECLARE_CLONE_AND_CREATE
+
+  const char* getOpString() const override {
+    return "LinearOp";
```
jacobhinkle, line 2268: Is this a `F.linear` op or a `torch.matmul` op? This line should match the class name and I think we should have separate ops and nodes for each of `linear` and `matmul`. This PR should only introduce one of these ops so we can inspect the mapping rules one op at a time.

csrc/ops/arith.cpp
jacobhinkle: Yes and that `matmul` should create a `MatmulOp` node. The code that's in there now could be converted to a utility in the future for decomposing a `MatmulOp`; that could be used in the nvfuser matmul scheduler.

csrc/ops/arith.cpp
jacobhinkle: Specifically, we usually use this function to create a new output domain for things like binary ops: https://github.com/NVIDIA/Fuser/blob/378a8386b082456666b3949349ac3640ed7db5ff/csrc/ops/utils.cpp#L183.
I think we should refactor this to create a utility that takes a vector of aligned `IterDomain`s and produces a new output `IterDomain`. `newOutputDomain` would create one of these vectors for each dimension using values from each `TensorView` input. In your case you would call the new function directly with the pair of corresponding axes.

jacobhinkle: If this PR is working with concrete inputs now but not symbolic, then we can do that refactor in a separate PR. We should just guard the `matmul` function with a check that none of the input axes `isSymbolic()` until that is addressed/tested.

csrc/root_domain_map.cpp
jacobhinkle: Possibly, but I don't think that's necessary since `PairwiseRootDomainMap` is not subclassed. Its parent class is `RootDomainMap` which does not need this functionality. Hypothetically could publicly inherit `OptOutConstDispatch` possibly and call `dispatch(consumer_tv_->definition())` here. However, we wouldn't know which input and output are this particular producer and consumer. So we'd wind up reinventing a bunch of stuff I think. Since it's not needed anywhere but here, I think it's safe to keep these `if`s here and just make them as tidy as possible.

csrc/root_domain_map.cpp
jacobhinkle: I don't think this is safe to do. We might have a different extent here when the inputs are `Symbolic`. Instead, I think we need to use the same logic as in the `matmul` op here to determine whether these axes correspond. If possible, it would be nice to factor out that logic from the op so that it can be used here.

csrc/scheduler/expr_eval_sched.h
```diff
@@ -0,0 +1,66 @@
...
+#include <scheduler/heuristic.h>
+#include <scheduler/registry.h>
+
+namespace nvfuser {
+
+class Fusion;
+class SchedulerRuntimeInfo;
+class HeuristicSummary;
+
+//! ExprEval scheduler represents the case where we allocate outputs directly using EE. No code is generated.
```
jacobhinkle, line 19: Can we save the new scheduler for a separate PR which we will review as soon as the new node is merged?

jacobhinkle: Instead, can we have a very simple test that creates a `FusionExecutor` with this op and evaluates it with ATen. This will let us test the node independent of scheduler changes.

tests/cpp/test_matmul_aten_evaluation.cpp
jacobhinkle: Is the problem that there are a variable number of dimensions in each test case?

tests/cpp/test_matmul_aten_evaluation.cpp
```diff
@@ -393,4 +393,61 @@ TEST_F(MatmulATenEvaluationTest, LinearWithBias) {
...
+  fusion->addInput(tv0);
+  fusion->addInput(tv1);
+  fusion->addOutput(tv2);
+
+  at::Tensor t0 = at::randn(a_shape, at::kHalf).cuda();
+  at::Tensor t1 = at::randn(b_shape, at::kHalf).cuda();
+  at::Tensor out_ref = at::matmul(t0, t1);
+
+  FusionExecutorCache fec(std::move(fusion));
+  auto out = fec.runFusionWithInputs({t0, t1});
```
jacobhinkle, line 415: Let's strip out the scheduler changes as mentioned above and change these tests to directly build a `FusionExecutor` to test that evaluation is correct for all supported input patterns.

csrc/ops/arith.cpp
Priya2698: I will rename once we are ready to swap the backend of the python API to use the new nodes. Now that matmul is enabled in Thunder, renaming will break things there.

csrc/ops/arith.cpp
Priya2698: > Specifically, we usually use this function to create a new output domain for things like binary ops:
> 
> https://github.com/NVIDIA/Fuser/blob/378a8386b082456666b3949349ac3640ed7db5ff/csrc/ops/utils.cpp#L183
> 
> .
> I think we should refactor this to create a utility that takes a vector of aligned `IterDomain`s and produces a new output `IterDomain`. `newOutputDomain` would create one of these vectors for each dimension using values from each `TensorView` input. In your case you would call the new function directly with the pair of corresponding axes.

I am not clear on this. Could you give me a small example on what the inputs will be?

Priya2698: Do you mean something like `IterDomain* newIterDomain (const IterDomain* id)`?

csrc/ir/internal_nodes.h
```diff
@@ -2255,4 +2255,37 @@ class NVF_API CatOp : public Expr {
   Val* getPred(int input_idx) const;
 };
 
+//! Matmul Operator to be expression evaluated without decomposition.
+class NVF_API MatmulOp : public Expr {
```
wujingyue, line 2259: https://github.com/NVIDIA/Fuser/blob/main/doc/dev/visibility.md#should-i-mark-my-new-method-or-class-as-nvf_api

```suggestion
class MatmulOp : public Expr {
```

tests/cpp/test_matmul_aten_evaluation.cpp
```diff
@@ -393,4 +408,92 @@ TEST_F(MatmulATenEvaluationTest, LinearWithBias) {
 
   EXPECT_TRUE(at::allclose(out[0], out_ref));
 }
+
+TEST_P(ATenNodesParametrizedTest, MatmulNodeConcrete) {
+  auto fusion = std::make_unique<Fusion>();
+  FusionGuard fg(fusion.get());
+
+  std::vector<int64_t> a_shape = std::get<0>(GetParam());
+  std::vector<int64_t> b_shape = std::get<1>(GetParam());
```
wujingyue, line 417:
```suggestion
  const auto& [a_shape, b_shape] = GetParam();
```

C++17 calls this structured binding declaration fyi.

tests/cpp/test_matmul_aten_evaluation.cpp
```diff
@@ -31,6 +31,21 @@ class MatmulATenEvaluationTest : public NVFuserTest {
       optimization_guard_;
 };
 
+using MatmulNodeParamType =
+    std::tuple<std::vector<int64_t>, std::vector<int64_t>>;
```
wujingyue, line 35: Nit:

```suggestion
using Sizes = std::vector<int64_t>;
using MatmulNodeParamType = std::tuple<Sizes, Sizes>;
```
to save you some typing.

csrc/ops/arith.cpp
Priya2698: I think I got your point. Will ping you once I finish the refactor.

jacobhinkle: Currently, `newOutputDomain` takes a vector of `TensorView`s and assumes the axes are aligned in the trivial way, but that will need to be more flexible for your case. So I think you'd want something like `IterDomain* newIterDomain(const std::vector<IterDomain*> ids)` and the argument would be all of the mapped producer IDs.

csrc/ops/arith.cpp
jacobhinkle: Sounds good. Re thunder: they use the python API anyway so changes to these function names don't necessarily impact them, as long as the python frontend `matmul` op is updated accordingly.

csrc/ops/arith.cpp
```diff
@@ -2684,4 +2684,54 @@ TensorView* tensor(Val* val) {
...
+} // namespace
+
+
+// TODO (Priya): This will be renamed to matmul once we are ready to modify the python API backend.
+// Keeping separate for now, to avoid breaking tests in Thunder.
+TensorView* eagerMatmul(
+    TensorView* tv_a,
+    TensorView* tv_b) {
+  NVF_CHECK(tv_a->getDataType().value() == tv_b->getDataType().value());
+  TensorView* out = newForMatmul(tv_a, tv_b);
```
jacobhinkle, line 2732: In a followup PR I think we should avoid creating this node for the following two special cases:
1. Both inputs are 1D. We should `return sum(mul(tv_a, tv_b));`.
2. Both inputs are `IterType::Broadcast` in the K dimension i.e. `tv_a [M, 1]` and `tv_b [1, N]`. In this case the output is a simple outer product `return mul(tv_a, tv_b);`.

Avoiding `MatmulOp`s means it will be easier for us to fuse into the reduction or pointwise schedulers.

csrc/ops/utils.cpp
```diff
@@ -174,111 +174,111 @@ IterType promoteIterType(IterType type1, IterType type2) {
...
     }
+  return out_domain;
+}
+
+// Adding these pragmas since gcc-12.2.1
+// incorrectly reports a warning with the use of evaluate
+#if defined(__GNUC__) && !defined(__clang__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wfree-nonheap-object"
+#endif
```
jacobhinkle, line 257: CC @protonu. Please check that this GCC check is in the right place. Specifically, should it instead be at `outIterDomain()`? I checked the original PR https://github.com/NVIDIA/Fuser/issues/643 but I didn't find any mention of this code so I'm not sure what the original problem was.

csrc/ops/utils.h
```diff
@@ -33,6 +33,8 @@ Val* newScalar(ValType vtype, DataType dtype);
 
 IterType promoteIterType(IterType type1, IterType type2);
 
+IterDomain* outIterDomain(const std::vector<IterDomain*>& ids);
```
jacobhinkle, line 36: Should we call it `newOutputIterDomain` instead?

csrc/root_domain_map.cpp
```diff
@@ -121,6 +121,37 @@ std::unordered_map<IterDomain*, IterDomain*> PairwiseRootDomainMap::map(
   const auto producer_root =
       TensorDomain::noReductions(producer->maybeRFactor());
   const auto& consumer_root = consumer->root();
+
+  // Map MatmulOp extents in reverse
```
jacobhinkle, line 125: Please place a description of the logic in a comment here. You may want to refer to https://pytorch.org/docs/stable/generated/torch.matmul.html. Ideally a reader should be able to quickly read this and confirm that it implements the torch.matmul interface.

tests/cpp/test_matmul_aten_evaluation.cpp
```diff
@@ -31,6 +31,21 @@ class MatmulATenEvaluationTest : public NVFuserTest {
...
+
+class ATenNodesParametrizedTest
+    : public NVFuserFixtureParamTest<MatmulNodeParamType> {
+ protected:
+  // Allocation order set by the pass breaks matmul tests
+  // see issue https://github.com/NVIDIA/Fuser/issues/1810
+  ATenNodesParametrizedTest() : optimization_guard_(false) {}
+
+ private:
```
jacobhinkle, line 44: Since this depends on aten evaluation, should these tests also add a `DisableOptionsGuard` that ensures `matmul_expr_eval` is unset?

csrc/ops/utils.cpp
Priya2698: Based on the comment about issues with using `evaluate`, it should be around the new `outIterDomain` function.

tests/cpp/test_matmul_aten_evaluation.cpp
Priya2698: I was going to remove setting that flag within the matmul scheduler once I added the ATen scheduler in the next PR.

I don't see a reason to keep this option anymore.

For these tests in particular, they would be unaffected since the outputs are marked for expression evaluator in the ATen scheduler anyway, skipping compilation and would behave the same way as the existing tests (using pattern matching)

csrc/ops/arith.cpp
Priya2698: Added the first case to this PR.

csrc/ops/utils.cpp
```diff
@@ -174,12 +174,128 @@ IterType promoteIterType(IterType type1, IterType type2) {
...
+//! input iterdomain corresponding to the output iterdomain at that index.
+//! If the element is nullptr, there is no mapping between input-output at that
+//! index. Based on the input dimensions following cases are possible:
+//! 1. A/B is 1D: [M, K] x [K] -> [M]
+//! Mapping A: {id_M}, Mapping B: {nullptr}
+//! 2. A and B are 2D: [M, K] x [K, N] -> [M, N]
+//! Mapping A: {id_M, nullptr}, Mapping B: {nullptr, id_N}
+//! 3. A/B are atleast 1D and one of them is > 2D: [B, M, K] x [K, N] -> [B, M,
+//! N] Mapping A: {id_B, id_M, nullptr}, Mapping B: {nullptr, nullptr, id_N}
+std::vector<IterDomain*> mapMatmulOpIterDomains(
```
Priya2698, line 188: Should this be in `csrc/utils`?

csrc/root_domain_map.cpp
Priya2698: Refactored the mapping logic to a separate function. Let me know if the new comment is clear about the functionality

csrc/ops/arith.cpp
Priya2698: In these cases, do we add a `castOp` as well? `castOp` is needed to mimic `torch.matmul` behavior.

Priya2698:
!build

csrc/ops/arith.cpp
jacobhinkle: No, I think we should just create the output as half-precision instead of `Float`.

jacobhinkle: Oh I see what you mean. Yeah I think at first we'd need to insert the cast.

csrc/ops/utils.cpp
```diff
@@ -174,12 +174,128 @@ IterType promoteIterType(IterType type1, IterType type2) {
   }
 }
 
+//! Maps the input iterdomains to the output of MatmulOp.
+//! Returns a vector where each element is the
+//! input iterdomain corresponding to the output iterdomain at that index.
```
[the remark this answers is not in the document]
Priya2698: The entries are still the input iterdomain, just aligned to the output.
For eg: Consider `[B, M, K] x [K, N] -> [B, M, N]`. For A: the function returns `{id_B, id_M, nullptr}`. For B: the function returns `{nullptr, nullptr, id_N}`. This allows us to identify which of the input iterdomains correspond to a particular index in the output.

csrc/ops/utils.h
```diff
@@ -33,6 +33,14 @@ Val* newScalar(ValType vtype, DataType dtype);
 
 IterType promoteIterType(IterType type1, IterType type2);
 
+//! Maps the input iterdomains to the output of MatmulOp.
+std::vector<IterDomain*> mapMatmulOpIterDomains(
+    const std::vector<IterDomain*>& input_domain,
+    bool is_lhs,
+    size_t out_size);
```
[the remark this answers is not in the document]
Priya2698: Yes.

csrc/ops/utils.cpp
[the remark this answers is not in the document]
Priya2698: The use was to avoid copying this logic into `root_domain_map`. This way, I can use the output of this function to map producer-consumer for `MatmulOp`. Does this clarify the use of the function?

tests/cpp/test_matmul_aten_evaluation.cpp
Priya2698: Right, we can reuse this option to disable the scheduler. I can add that in the scheduler PR: #2209 (Draft)

csrc/ops/utils.cpp
Priya2698: We can move it to `csrc/scheduler/matmul_utils.cpp` since this function is used to create the output for `MatmulOp` as well.

csrc/ops/utils.cpp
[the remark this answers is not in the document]
Priya2698: We could rename it to `alignMatmulOpInputToOutput` if `mapping` is creating a misunderstanding about the functionality.

csrc/ops/composite.cpp
```diff
@@ -299,5 +299,62 @@ TensorView* view_as_real(TensorView* x) {
...
+  NVF_ERROR(ndims_a >= 1 && ndims_b >= 1);
+
+  auto ndims_out = std::max(ndims_a, ndims_b);
+  if (std::min(ndims_a, ndims_b) == 1) {
+    ndims_out = std::max(ndims_a, ndims_b) - 1;
+  }
+
+  std::vector<IterDomain*> out_domain(ndims_out, nullptr);
+
+  const auto& mapping_a =
```
wujingyue, line 323: Can you spell out the type so it's easier to review?

csrc/ops/composite.cpp
```diff
@@ -299,5 +299,62 @@ TensorView* view_as_real(TensorView* x) {
...
+
+  auto ndims_a = orig_domain_a.size();
+  auto ndims_b = orig_domain_b.size();
+
+  NVF_ERROR(ndims_a >= 1 && ndims_b >= 1);
+
+  auto ndims_out = std::max(ndims_a, ndims_b);
+  if (std::min(ndims_a, ndims_b) == 1) {
+    ndims_out = std::max(ndims_a, ndims_b) - 1;
+  }
```
wujingyue, line 319: Can you clarify? It's not immediately clear to me.

tests/cpp/test_matmul_aten_evaluation.cpp
```diff
@@ -393,4 +408,90 @@ TEST_F(MatmulATenEvaluationTest, LinearWithBias) {
...
+}
+
+constexpr int64_t b = 128, m = 64, k = 32, n = 16;
+
+INSTANTIATE_TEST_SUITE_P(
+    ,
+    ATenNodesParametrizedTest,
+    testing::Values(
+        std::make_tuple(
+            std::vector<int64_t>({m, k}),
```
wujingyue, line 479:
```suggestion
            Sizes({m, k}),
```

for less typing, since you already defined `Sizes`.

csrc/ops/utils.h
```diff
@@ -33,6 +33,14 @@ Val* newScalar(ValType vtype, DataType dtype);
...
+
+IterDomain* newOutputIterDomain(const std::vector<IterDomain*>& ids);
```
wujingyue, line 42: What's this function for?

csrc/ops/utils.cpp
```diff
@@ -174,12 +174,128 @@ IterType promoteIterType(IterType type1, IterType type2) {
...
+  for (auto out_inx = (int64_t)out_size - 1, inp_inx = inp_size - 1; inp_inx >= 0;
+       inp_inx--) {
+    if (inp_inx != kpos) {
+      mapping[out_inx] = input_domain[inp_inx];
+      out_inx--;
+    } else if (inp_size <= (int64_t)out_size) {
+      out_inx--;
+    }
+  }
+  return mapping;
```
wujingyue, line 213: Do you expect out_inx to reach 0 at this point?

csrc/ops/utils.cpp
```diff
@@ -174,12 +174,128 @@ IterType promoteIterType(IterType type1, IterType type2) {
...
+  if (inp_size == 1) {
+    // Only reduction axis {K}
+    return mapping;
+  }
+  // Input A to matmul: {*, M, K}
+  // Input B to matmul: {*, K, N} / {K} (ndims = 1)
+  auto kpos = is_lhs ? inp_size - 1 : inp_size - 2;
+
+  // If A/B is 1D, out_size < inp_size.
+  for (auto out_inx = (int64_t)out_size - 1, inp_inx = inp_size - 1; inp_inx >= 0;
```
wujingyue, line 204:
```suggestion
  for (auto out_index = (int64_t)out_size - 1, inp_index = inp_size - 1; inp_index >= 0;
```

Or `idx`. `inx` is a much less common acronym.

csrc/ops/utils.cpp
```diff
@@ -174,12 +174,128 @@ IterType promoteIterType(IterType type1, IterType type2) {
...
+//! index. Based on the input dimensions following cases are possible:
+//! 1. A/B is 1D: [M, K] x [K] -> [M]
+//! Mapping A: {id_M}, Mapping B: {nullptr}
+//! 2. A and B are 2D: [M, K] x [K, N] -> [M, N]
+//! Mapping A: {id_M, nullptr}, Mapping B: {nullptr, id_N}
+//! 3. A/B are atleast 1D and one of them is > 2D: [B, M, K] x [K, N] -> [B, M,
+//! N] Mapping A: {id_B, id_M, nullptr}, Mapping B: {nullptr, nullptr, id_N}
+std::vector<IterDomain*> mapMatmulOpIterDomains(
+    const std::vector<IterDomain*>& input_domain,
+    bool is_lhs,
```
wujingyue, line 190: I'd use MmaOperand::A or ::B instead: https://github.com/NVIDIA/Fuser/blob/3598b4df9a2de7f9e202f93fb6f836c9168a402b/csrc/mma_type.h#L208

1. Enum is logically closer to what you are really trying to represent here. 
2. For better type safety. 
3. For better readability on the caller side. You'd need to write `/*is_lhs=*/true` and `/*is_rhs=*/false` in the caller to achieve similar readability; otherwise, I've no idea what `true/false` refers to at the first glance. 
4. Flexibility **if** you want this to be used for linear which takes three operands or for the destination matrix. But be aware of YAGNI. 

I was trying to search for some articles talking about enum vs boolean. https://betterprogramming.pub/dont-use-boolean-arguments-use-enums-c7cd7ab1876a seems to be a good reference.

wujingyue (commented):
I made some drive-by comments, but I'll really defer approval to others because I'm not a matmul expert and I have a bit too many PRs to review this week :)

csrc/ops/composite.cpp
Priya2698: Added comments. Matmul output size is same as the higher dimensional input size if both A/B > 1D. If one of the inputs is 1D, the output size is 1 less than the higher dimensional input size, since either M/N axis will be missing in the output. For eg, consider: `[M, K] x [K] -> [M]`

csrc/ops/utils.cpp
```diff
@@ -174,12 +174,128 @@ IterType promoteIterType(IterType type1, IterType type2) {
   }
 }
 
+//! For MatmulOp, the input iterdomains at a given index do not necessarily map to the output iterdomain at that index
```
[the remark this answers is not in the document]
Priya2698: It is the rfactor domain where it exists or the root domain otherwise.

csrc/ops/utils.cpp
Priya2698: Not necessarily. For instance: `[B, M, K] x [K, N] -> [B, M, N]`. The output index will not reach 0 for input B.

csrc/ops/utils.h
Priya2698: The `newOutputDomain` takes a vector of tensorviews and assumes they are all aligned to create the output tensorview. Since in the case of matmul, the inputs are not aligned to the output and the mapping is more complex, I refactored this part of the code from `newOutputDomain` into a separate `newOutputIterDomain` that takes a vector of aligned input iterdomains to create the output iterdomain.

In `newOutputDomain` (which we use in binary ops), we simply call this function for every set of input iterdomains at any axis. For `MatmulOp`, we specifically identify which input iterdomains correspond to a particular output iterdomain and use this function instead of using `newOutputDomain`.

Priya2698:
> I made some drive-by comments, but I'll really defer approval to others because I'm not a matmul expert and I have a bit too many PRs to review this week :)

Thanks @wujingyue, for the helpful comments.

csrc/ops/arith.cpp
Priya2698: Added the first case to this PR.

csrc/ops/utils.cpp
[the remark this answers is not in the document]
jacobhinkle: I tend to use //! for doxygen but I confess I haven't looked at the doxygen output in over a year. I think you can also use /// for that. My editor colors these differently from //.

tests/cpp/test_matmul_aten_evaluation.cpp
```diff
@@ -393,4 +408,79 @@ TEST_F(MatmulATenEvaluationTest, LinearWithBias) {
...
+    ,
+    ATenNodesParametrizedTest,
+    testing::Values(
+        std::make_tuple(Sizes({k}), Sizes({k})),
+        std::make_tuple(Sizes({m, k}), Sizes({k, n})),
+        std::make_tuple(Sizes({k}), Sizes({k, n})),
+        std::make_tuple(Sizes({m, k}), Sizes({k})),
+        std::make_tuple(Sizes({k}), Sizes({b, k, n})),
+        std::make_tuple(Sizes({b, m, k}), Sizes({k})),
+        std::make_tuple(Sizes({b, 1, m, k}), Sizes({b, k, n}))));
```
jacobhinkle, line 484: Is it true that we can accept any combination of A and B where A is one of `{k}, {m, k}, {b, m, k}, {b, 1, m, k}` and B is one of `{k}, {k, n}, {b, k, n}`? If so maybe we could parametrize each of those separately and hit all combos.

jacobhinkle (approved):
LGTM other than some minor comments

csrc/ops/composite.cpp
```diff
@@ -299,5 +299,73 @@ TensorView* view_as_real(TensorView* x) {
...
+  return IrBuilder::create<TensorView>(td, tv_a->dtype());
+}
+
+} // namespace
+
+// TODO (Priya): This will be renamed to matmul once we are ready to modify the
+// python API backend. Keeping separate for now, to avoid breaking tests in
+// Thunder.
+TensorView* eagerMatmul(TensorView* tv_a, TensorView* tv_b) {
+  NVF_CHECK(tv_a->dtype() == tv_b->dtype());
```
jacobhinkle, line 357: This condition is not mentioned in https://pytorch.org/docs/stable/generated/torch.matmul.html although `torch.matmul` fails with this error if you try different types:
> RuntimeError: expected mat1 and mat2 to have the same dtype, but got: c10::Half != float
Maybe we could put a comment there that this restriction might possibly be lifted in the future.

tests/cpp/test_matmul_aten_evaluation.cpp
```diff
@@ -393,4 +408,79 @@ TEST_F(MatmulATenEvaluationTest, LinearWithBias) {
...
+  fusion->addInput(tv1);
+  fusion->addOutput(tv2);
+
+  at::Tensor t0 = at::randn(a_shape, at::kHalf).cuda();
+  at::Tensor t1 = at::randn(b_shape, at::kHalf).cuda();
+  at::Tensor out_ref = at::matmul(t0, t1);
+
+  FusionExecutor fe;
+  fusion->aliasOutputToInput(
+      fusion->outputs()[0], /*input=*/nullptr, AllocationType::Evaluate);
```
jjsjann123, line 432: This is a strange API for marking something as AllocationType::Evaluate...

tests/cpp/test_matmul_aten_evaluation.cpp
jjsjann123: Out of curiosity. What would happen if k/m/n happen to be 1?

csrc/ops/composite.cpp
```diff
@@ -299,5 +299,73 @@ TensorView* view_as_real(TensorView* x) {
...
+      input_ids.emplace_back(mapping_a[idx]);
+    }
+    if (mapping_b[idx] != nullptr) {
+      input_ids.emplace_back(mapping_b[idx]);
+    }
+    out_domain[idx] = ops::newOutputIterDomain(input_ids);
+  }
+
+  TensorDomain* td = IrBuilder::create<TensorDomain>(
+      out_domain, TensorDomain::getContiguityFilledWith(out_domain, true));
```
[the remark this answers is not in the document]
jacobhinkle: Isn't that quote showing it is false by default?

csrc/ops/utils.cpp
```diff
@@ -174,12 +174,131 @@ IterType promoteIterType(IterType type1, IterType type2) {
...
+      "Unexpected input type.");
+  std::vector<IterDomain*> mapping(out_size, nullptr);
+  auto inp_size = (int64_t)input_domain.size();
+
+  if (inp_size == 1) {
+    // Only reduction axis {K}
+    return mapping;
+  }
+  // Input A to matmul: {*, M, K}
+  // Input B to matmul: {*, K, N} / {K} (ndims = 1)
```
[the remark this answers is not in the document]
Priya2698: Yes, I can modify the comment if it is misleading -- the intention was to show the different position of K in N based on dimension.

csrc/ops/utils.h
```diff
@@ -33,6 +34,25 @@ Val* newScalar(ValType vtype, DataType dtype);
 
 IterType promoteIterType(IterType type1, IterType type2);
 
+// For MatmulOp, the input iterdomains at a given index do not necessarily map
+// to the output iterdomain at that index This function aligns the input
+// iterdomain to the output and returns a vector where each element is the input
+// iterdomain corresponding to the output iterdomain at that index. If the
+// element is nullptr, there is no mapping between input-output at that index.
+// Based on the input dimensions following cases are possible:
+// 1. A/B is 1D: [M, K] x [K] -> [M]
```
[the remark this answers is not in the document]
Priya2698: Yes, I was giving an example per condition, but I see the confusion. Will modify

tests/cpp/test_matmul_aten_evaluation.cpp
Priya2698: > Out of curiosity. What would happen if k/m/n happen to be 1?

Do you mean if the output shape is different? -- It will be the same as any other case.

Or how we intend to handle those cases?
@jacobhinkle pointed out we could special-case these cases though to increase opportunity for fusion. (For eg: For `[M, 1] x [1, N]`, we can simply return the outer product without creating the `MatmulOp` node)

tests/cpp/test_matmul_aten_evaluation.cpp
Priya2698: We used the existing framework for aliasing, hence, some of the API names may be weird.
I'll make a note of this to modify in a future cleanup PR.

tests/cpp/test_matmul_aten_evaluation.cpp
```diff
@@ -393,4 +408,79 @@ TEST_F(MatmulATenEvaluationTest, LinearWithBias) {
...
+}
+
+TEST_P(ATenNodesParametrizedTest, MatmulNodeSymbolic) {
+  auto fusion = std::make_unique<Fusion>();
+  FusionGuard fg(fusion.get());
+
+  const auto& [a_shape, b_shape] = GetParam();
+
+  auto tv0 = makeSymbolicTensor(a_shape.size(), DataType::Half);
+  auto tv1 = makeSymbolicTensor(b_shape.size(), DataType::Half);
```
jacobhinkle, line 449:
```suggestion
  auto tv0 = makeSymbolicTensor(a_shape, DataType::Half);
  auto tv1 = makeSymbolicTensor(b_shape, DataType::Half);
```
This will ensure the tensors are defined in the fusion almost the way they would be using `fd.from_pytorch` in case the shapes contain `1`s. That could be useful here since you might translate some ops to non-`MatmulOp` if they are trivial. To be more precise though, this will still not declare them as contiguous. For that you might want to do
```c++
  auto tv0 = TensorViewBuilder().shape(sym_shape).dtype(DataType::Half).contiguity(true).build();
```
where `sym_shape` is a vector of `1` and `-1`.

Priya2698:
!build

tests/cpp/test_matmul_aten_evaluation.cpp
jacobhinkle: > For `[M, 1] x [1, N]`

This is the `K=1` case but I think Jie is asking what about M and N, which we aren't testing here. I would also add that we're not really testing this case properly either since we are creating the input tensors with `makeSymbolicTensor(a_shape.size())` which will make all dimensions `IterType::Iteration`. I suggested a code change that I think will address that, then we can add shape combos that have 1s in each position.

tests/cpp/test_matmul_aten_evaluation.cpp
Priya2698:
```
To be more precise though, this will still not declare them as contiguous
```

Why do we need this?

tests/cpp/test_matmul_aten_evaluation.cpp
Priya2698: I added cases for M=1/N=1. At present, they will behave the same way as when M/N > 1.

Priya2698:
!build

tests/cpp/test_matmul_aten_evaluation.cpp
jacobhinkle: However if in the future we try to evaluate this test by forcing the nvfuser matmul scheduler we might use an inefficient kernel since we won't be able to tell the inputs are contiguous. It's not a high priority, but it would be nice to have a test utility that could take an `at::Tensor` and create a `TensorView*` that matches it just like how we do in `fd.from_pytorch`.

tests/cpp/test_matmul_aten_evaluation.cpp
Priya2698: >  I would also add that we're not really testing this case properly either since we are creating the input tensors with `makeSymbolicTensor(a_shape.size())` which will make all dimensions `IterType::Iteration`. 

`makeSymbolicTensor(a_shape)` -> Does this mark dimensions as Broadcast if the extent is 1?

Priya2698:
!build

PR #2176, opened 2024-05-01
Fix schedule and lowering of Hopper MMAs

zasdfgbnm:
This PR fixes the schedule and lowering of Hopper MMAs. Please see my own review on this PR for detail.
After https://github.com/NVIDIA/Fuser/pull/2194, the schedule and lowering of MMA finally makes sense now.

Note that MMA is still limited to have only one tile in smem. Supporting multiple tiles is the next step.

### DIFF
--- a/csrc/device_lower/pass/index.cpp
+++ b/csrc/device_lower/pass/index.cpp
@@ -1584,15 +1584,22 @@ void IndexLowering::handle(const MmaOp* mma) {
     auto tv = mma->inA()->as<TensorView>();
     auto base_addr = IrBuilder::baseAddressExpr(tv);
     auto swizzle = getSwizzleMode(tv);
-    int64_t stride_bytes = core_matrix_outer_size *
-        getBytesFromSwizzle(swizzle); // swizzle period in bytes
     int64_t leading_bytes = core_matrix_outer_size *
+        getBytesFromSwizzle(swizzle); // swizzle period in bytes
+    int64_t stride_bytes = core_matrix_outer_size *
         /*number of core matrices, rounded up to handle padding */
         roundUpToMultiple(getM(mma->macro()) * /*bytes per item*/ 2L,
                           getBytesFromSwizzle(swizzle));
-    if (swizzle != MmaInputSmemSwizzle::None) {
-      // TODO: why???!!!
-      std::swap(leading_bytes, stride_bytes);
+    if (swizzle == MmaInputSmemSwizzle::None) {
+      if (mma->layout() == MmaLayout::NT || mma->layout() == MmaLayout::NN) {
+        std::swap(leading_bytes, stride_bytes);
+      } else {
+        stride_bytes = core_matrix_outer_size *
+            /*number of core matrices, rounded up to handle padding */
+            roundUpToMultiple(
+                           getK(mma->macro()) * /*bytes per item*/ 2L,
+                           getBytesFromSwizzle(swizzle));
+      }
     }
     auto matrix_desc = constructMatrixDescriptor(
         base_addr,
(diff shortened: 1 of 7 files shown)


csrc/device_lower/pass/index.cpp
```diff
@@ -1583,13 +1584,19 @@ void IndexLowering::handle(const MmaOp* mma) {
     auto tv = mma->inA()->as<TensorView>();
     auto base_addr = IrBuilder::baseAddressExpr(tv);
     auto swizzle = getSwizzleMode(tv);
-    int64_t stride_bytes =
-        8L * getBytesFromSwizzle(swizzle); // swizzle period in bytes
-    int64_t leading_bytes = /*8x8 items each core matrix*/ 64L *
-        /*number of core matrices*/ (getM(mma->macro()) / 8L) *
```
zasdfgbnm, line 1589: Depending on the layout, the inner can be either M or K. It makes no sense to have just M here.

csrc/device_lower/pass/index.cpp
```diff
@@ -1612,16 +1619,19 @@ void IndexLowering::handle(const MmaOp* mma) {
...
-    int64_t leading_bytes = /*8x8 items each core matrix*/ 64L *
+    int64_t leading_bytes = core_matrix_outer_size *
+        getBytesFromSwizzle(swizzle); // swizzle period in bytes
+    int64_t inner_size =
+        (mma->layout() == MmaLayout::TN || mma->layout() == MmaLayout::NN)
+        ? getK(mma->macro())
+        : getN(mma->macro());
+    int64_t stride_bytes = core_matrix_outer_size *
         /*number of core matrices, rounded up to handle padding */
-        roundUpToMultiple(getN(mma->macro()) / 8L,
```
zasdfgbnm, line 1619: Similar to https://github.com/NVIDIA/Fuser/pull/2176#discussion_r1593194403

csrc/device_lower/pass/index.cpp
```diff
@@ -1612,16 +1619,19 @@ void IndexLowering::handle(const MmaOp* mma) {
...
-                          getBytesFromSwizzle(swizzle) / 16L) *
-        /*bytes per item*/ 2L;
-    if (swizzle != MmaInputSmemSwizzle::None &&
-        (mma->layout() == MmaLayout::TT || mma->layout() == MmaLayout::TN)) {
```
zasdfgbnm, line 1623: B's layout depend on whether A is transposed or not? This is definitely nonsense.

csrc/scheduler/mma_utils.cpp
```diff
@@ -853,12 +853,7 @@ void WarpMmaSwizzler::scheduleOperandRead(TensorView* tv, MmaOperand operand) {
 // https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#asynchronous-warpgroup-level-matrix-shared-memory-layout-swizzling-modes
 void WarpMmaSwizzler::scheduleOperandRead(
     TensorView* tv,
-    MmaInputSmemSwizzle swizzle,
-    bool transpose,
-    bool transpose2) {
-  if (transpose) {
-    tv->reorder({{-2, -1}});
-  }
```
zasdfgbnm, line 861: When this function is called, the leaf should have already have the correct order. It makes no sense to do any transpose here.

csrc/scheduler/mma_utils.cpp
```diff
@@ -868,16 +863,9 @@ void WarpMmaSwizzler::scheduleOperandRead(
     // [Ko, K8, Mo, M8]
     tv->reorder({{-2, -3}});
     // [Ko, Mo, K8, M8]
-    if (transpose2) {
```
zasdfgbnm, line 871: `transpose2` is definitely nonsense.

zasdfgbnm:
!build

naoyam (approved):
LGTM

csrc/device_lower/pass/index.cpp
```diff
@@ -1583,13 +1584,19 @@ void IndexLowering::handle(const MmaOp* mma) {
...
+        (mma->layout() == MmaLayout::TT || mma->layout() == MmaLayout::TN)
+        ? getK(mma->macro())
+        : getM(mma->macro());
+    int64_t stride_bytes = core_matrix_outer_size *
+        /*number of core matrices, rounded up to handle padding */
+        roundUpToMultiple(inner_size * /*bytes per item*/ 2L,
+                          getBytesFromSwizzle(swizzle));
+    if (swizzle == MmaInputSmemSwizzle::None &&
+        (mma->layout() == MmaLayout::NT || mma->layout() == MmaLayout::NN)) {
+      // tnspA and tnspB is ignored for NoSwizzle mode
```
jacobhinkle, line 1599: I don't understand this comment. Doesn't the condition above mean that the transpose of A is not ignored in this case?

zasdfgbnm: These two strides specifies "where to find next core matrix", for the cases where the swizzle mode is not none, which stride to use depend on the value of `tnspA` and `tnspB`. Something like:
```C++
if (tnspB) {
  address = i * stride0 + j * stride1;
} else {
  address = i * stride1 + j * stride0;
}
```
For swizzle mode none, there is no such `if`, it is just
```C++
address = i * stride0 + j * stride1;
```

PR #2180, opened 2024-05-02
Replace NVF_ERROR with EXPECT_TRUE.

wujingyue:
When the check fails, the latter fails gracefully and doesn't stop other tests from running.


wujingyue:
!build

samnordmann (approved):

PR #2181, opened 2024-05-02
DID parallelization of leaf IDs - local compute

cowanmeg:
First steps in allowing DID parallelization of a leaf IterDomain*.  This PR updates tensor binding, tensor metadata, and sharded tensor allocation to enable single device computation with DID parallelized leaf IterDomains when allocation domain is manually set. Follow-up PRs will updates will update collective communication lowering and provide automation. 

*Restrictions - we can later relax these, but for simplicity of generating correct and/or performant code we have these restrictions:
1. DID parallelization of leaf ID is restricted to a split RFactor ID. This is to ensure that only one logical/RFactor ID is sharded.
2. DID parallelization is restricted to the outer split axis. This ensures shards map to a contiguous range of elements in the logical axis, as opposed to an interleaved set of elements. 
3. DID parallelized iter domain extent equals the number of devices sharded over. Ensures that each device in the mesh axis receives a shard of the tensor.
4. Extent of logical/RFactor iterdomain is evenly divided by the number of devices. Ensures that each shard is equal sized. 

Changes:
Sharded tensor representation:
```
// D = number of devices in device mesh x-axis. 
TensorView* tv = makeTensor(2);
tv->split(0, D, inner=false);
tv->axis(0)->parallelize(DIDx);
```
RFactor: [i0, i1] - holds extents of the logical / unsharded tensor
Leaf: [DIDx(D), i0/D, i1]
Allocation: [DIDx(D), i0/D, i1] -> Ignoring the DID parallelized IDs will map to the sharded tensor. 
Unit tests explicitly set the allocation domain of a sharded tensor to the leaf domain. TODO: later PRs will automate this.

Tensor metadata is updated so that logical_size calculates and stores the _unsharded_ tensor sizes. TODO: The logical_stride still reflects the _sharded_ tensor...


cowanmeg:
Note: stacked on top of https://github.com/NVIDIA/Fuser/pull/1987

PR #2185, opened 2024-05-02
Make Communications IRs inheriting from Expr.

samnordmann:
This PR makes the "Communication" class a proper IR inheriting from Expr. This patch is needed for implementing Host Irs. It is also one step towards making the Communications (and more generally the multidevice module) fully symbolic.

By the way, we proceed to a couple of refactoring, and remove `Communicator::sendRecv` method.

Remarks:
1) This IR will be used for now in the context of Host Irs. Later, they could also serve as kernel IR (backed by device-side communication APIs at runtime).
2) Before this patch, we had a base class `Communication` and one derived class per collective type (Allgather, Allreduce, Broadcast, etc.). Now, there is only the class `Communication`, and the collective type is encoded through an enum class `CollectiveType` added to the parameter member `CommParams`
3) the `Communication::post` method was replaced by a standalone function `postCollective`. The motivation is to scoop out the runtime execution from the symbolic representation of the collective.
4) Note that step 3) is only implented halfway here since a `Collective` is instantiated with concrete device Idx and concrete at::Tensor, while it should be instantiated with symbolic representations and binded to actual device indices and Aten buffers at runtime. This will be added in a future PR.

CI: https://gitlab-master.nvidia.com/dl/pytorch/fuser-gh-mirror/-/pipelines/14923305

### DIFF
--- a/csrc/multidevice/communication.cpp
+++ b/csrc/multidevice/communication.cpp
@@ -14,6 +14,37 @@
 #include <utils.h>
 
 namespace nvfuser {
+
+std::ostream& operator<<(std::ostream& os, const CommunicationType& type) {
+  switch (type) {
+    case CommunicationType::Gather:
+      os << "Gather";
+      break;
+    case CommunicationType::Allgather:
+      os << "Allgather";
+      break;
+    case CommunicationType::Scatter:
+      os << "Scatter";
+      break;
+    case CommunicationType::Reduce:
+      os << "Reduce";
+      break;
+    case CommunicationType::Allreduce:
+      os << "Allreduce";
+      break;
+    case CommunicationType::ReduceScatter:
+      os << "ReduceScatter";
+      break;
+    case CommunicationType::Broadcast:
+      os << "Broadcast";
+      break;
+    case CommunicationType::SendRecv:
+      os << "SendRecv";
+      break;
+  }
+  return os;
+}
+
 namespace {
 
 inline void assertBufferCount(
@@ -59,37 +90,14 @@ inline bool isReduction(CommunicationType type) {
       type == CommunicationType::ReduceScatter;
 }
 
-inline std::string typeToString(CommunicationType type) {
-  switch (type) {
-    case CommunicationType::Gather:
-      return "Gather";
-    case CommunicationType::Allgather:
-      return "Allgather";
-    case CommunicationType::Scatter:
-      return "Scatter";
-    case CommunicationType::Reduce:
-      return "Reduce";
-    case CommunicationType::Allreduce:
-      return "Allreduce";
-    case CommunicationType::ReduceScatter:
-      return "ReduceScatter";
-    case CommunicationType::Broadcast:
-      return "Broadcast";
-    case CommunicationType::SendRecv:
-      return "SendRecv";
-    default:
-      NVF_ERROR(false);
-      return "";
-  }
-}
-
 inline void assertValid(
     const CommParams& params,
     const DeviceIdxType my_device_index) {
   assertBuffersHaveSameSize(params.src_bufs, params.dst_bufs);
+  std::unordered_set<DeviceIdxType> team_without_duplicates(
+      params.team.begin(), params.team.end());
   NVF_ERROR(
-      std::adjacent_find(params.team.cbegin(), params.team.cend()) ==
-          params.team.cend(),
+      team_without_duplicates.size() == params.team.size(),
       "the communication must not involve the same device more than once");
   NVF_ERROR(!params.team.empty(), "the team size must be greater than 0");
   NVF_ERROR(
@@ -181,7 +189,7 @@ std::string Communication::toString(int indent) const {
   std::string indent1 = ext_indent + "  ";
   std::string indent2 = ext_indent + "    ";
 
-  ss << ext_indent << "Communication " << typeToString(params_.type) << ": {\n";
+  ss << ext_indent << "Communication " << params_.type << ": {\n";
 
   if (hasRoot(params_.type)) {
     ss << indent1 << "root: " << params_.root << ",\n";
@@ -221,7 +229,7 @@ bool Communication::sameAs(const Statement* other) const {
       p1.team == p2.team && (!isReduction(p1.type) || p1.redOp == p2.redOp));
 }
 
-c10::intrusive_ptr<c10d::Work> postCommunication(
+c10::intrusive_ptr<c10d::Work> postSingleCommunication(
     Communication* communication,
     DeviceIdxType my_device_index,
     c10::intrusive_ptr<c10d::Backend> backend) {
@@ -324,7 +332,7 @@ c10::intrusive_ptr<c10d::Work> postCommunication(
       }
     }
     default:
-      NVF_ERROR(false, "Wrong communication type: ", typeToString(params.type));
+      NVF_ERROR(false, "Wrong communication type: ", params.type);
       return nullptr;
   }
 }
--- a/csrc/multidevice/communication.cpp
+++ b/csrc/multidevice/communication.cpp
@@ -168,7 +168,7 @@ inline void assertValid(
 // pytorch's process group expects the root to be specified
 // as an integer between 0 and world_size-1. We choose it to be
 // the device's relative index within the team
-DeviceIdxType getRootRelativeIndex(const CommParams& params) {
+int64_t getRootRelativeIndex(const CommParams& params) {
   auto it = std::find(params.team.begin(), params.team.end(), params.root);
   return std::distance(params.team.begin(), it);
 }
--- a/csrc/multidevice/communication.cpp
+++ b/csrc/multidevice/communication.cpp
@@ -5,6 +5,8 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 // clang-format on
+#include <ir/cloner.h>
+#include <ir/printer.h>
 #include <multidevice/communication.h>
 #if defined(USE_C10D_NCCL)
 #include <torch/csrc/distributed/c10d/ProcessGroupNCCL.hpp>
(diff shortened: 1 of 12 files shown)


csrc/multidevice/communication.cpp
```diff
@@ -40,56 +42,148 @@ inline void assertBuffersHaveSameSize(
...
+// the device's relative index within the team
+DeviceIdxType getRootRelativeIndex(const CommParams& params) {
+  auto it = std::find(params.team.begin(), params.team.end(), params.root);
+  return std::distance(params.team.begin(), it);
 }
 
+} // namespace
+
+Communication::Communication(IrBuilderPasskey passkey, CommParams params)
+    : Expr(passkey), params_(std::move(params)) {}
```
naoyam, line 171: FYI, here's how we enforce certain IR nodes only appearing in the Kernel IR:

https://github.com/NVIDIA/Fuser/blob/main/csrc/kernel_ir.cpp#L834-L836

Again I don't think things like this are important ATM.

samnordmann: Thanks! I will add it, but not for now since Communications are registered in a normal fusion right now, inside `MultiDeviceExecutor`

csrc/multidevice/communication.cpp
```diff
@@ -40,56 +42,148 @@ inline void assertBuffersHaveSameSize(
...
+      type == CommunicationType::SendRecv;
+}
+
+inline bool isReduction(CommunicationType type) {
+  return type == CommunicationType::Reduce ||
+      type == CommunicationType::Allreduce ||
+      type == CommunicationType::ReduceScatter;
+}
+
+inline std::string typeToString(CommunicationType type) {
```
wujingyue, line 62: I think it's sufficient to directly define an `operator<<` for `CommunicationType` without a typeToString helper, e.g., https://github.com/NVIDIA/Fuser/blob/4e8356915fc6ffa823cf97fed4249cc00f317fdf/csrc/type.cpp#L1231. `NVF_ERROR` is able to take a `operator<<`able type without using an explicit typeToString function.

csrc/multidevice/communication.cpp
```diff
@@ -40,56 +42,148 @@ inline void assertBuffersHaveSameSize(
...
+      return "Allreduce";
+    case CommunicationType::ReduceScatter:
+      return "ReduceScatter";
+    case CommunicationType::Broadcast:
+      return "Broadcast";
+    case CommunicationType::SendRecv:
+      return "SendRecv";
+    default:
+      NVF_ERROR(false);
+      return "";
```
wujingyue, line 82:
```suggestion
```

csrc/multidevice/communication.cpp
```diff
@@ -40,56 +42,148 @@ inline void assertBuffersHaveSameSize(
...
-      has_root_(has_root) {
-  assertBuffersHaveSameSize(params_.src_bufs, params_.dst_bufs);
+inline void assertValid(
+    const CommParams& params,
+    const DeviceIdxType my_device_index) {
+  assertBuffersHaveSameSize(params.src_bufs, params.dst_bufs);
   NVF_ERROR(
-      std::unique(params_.team.begin(), params_.team.end()) ==
-          params_.team.end(),
+      std::adjacent_find(params.team.cbegin(), params.team.cend()) ==
```
wujingyue, line 91: Question: this assumes params.team is sorted. Is that guaranteed?

csrc/multidevice/communication.h
```diff
@@ -218,12 +136,42 @@ case of a local copy)
   - If team is of size 2, the unique non-root have no src buffer and one dst
 buffer
 */
-class SendRecv : public Communication {
+
+class NVF_API Communication : public Expr {
```
wujingyue, line 140:
```suggestion
class Communication : public Expr {
```

https://github.com/NVIDIA/Fuser/blob/main/doc/dev/visibility.md#should-i-mark-my-new-method-or-class-as-nvf_api

csrc/multidevice/communication.cpp
```diff
@@ -107,214 +201,132 @@ std::string Communication::toString(int indent) const {
...
-  // This is used to change the representation of the buffers to match c10d
-  // ProcessGroup API
-  std::vector<std::vector<at::Tensor>> buf_list;
-  if (comm.deviceId() == params_.root) {
-    assertBufferCount(params_.dst_bufs, params_.team.size());
-    buf_list = {std::move(params_.dst_bufs)};
-  } else {
-    assertBufferCount(params_.dst_bufs, 0);
+// TODO add checking symbolic representation of src and dst buffers
+bool Communication::sameAs(const Statement* other) const {
```
wujingyue, line 209: I failed to see why we couldn't reuse the "default" implementation: https://github.com/NVIDIA/Fuser/blob/4e8356915fc6ffa823cf97fed4249cc00f317fdf/csrc/ir/base_nodes.cpp#L342. Can you clarify?

csrc/multidevice/executor.cpp
```diff
@@ -149,7 +149,7 @@ void MultiDeviceExecutor::postKernel(
   }
 }
 
-void MultiDeviceExecutor::postCommunication(SegmentedGroup* group) {
+void MultiDeviceExecutor::postResharding(SegmentedGroup* group) {
```
wujingyue, line 152: Can you move this to a different PR so merging is easier? I have https://github.com/NVIDIA/Fuser/pull/2172 to merge in conflict with this, so any help to make both PRs incremental is going to be useful!

wujingyue (commented):
LGTM in general! I made a few comments to address before I merge this.

csrc/multidevice/communication.cpp
samnordmann: that's a mistake on my side, let me fix that

csrc/multidevice/communication.cpp
samnordmann: Two Communications could be "Expr::sameAs" without being the same, for example if one is Allgather and the other Allreduce. Right?
My goal here is to mimick what's done for other IRs, such as https://github.com/NVIDIA/Fuser/blob/4e8356915fc6ffa823cf97fed4249cc00f317fdf/csrc/ir/nodes.cpp#L2674

But I don't actually use sameAs -- I simply thought implementing it was necessary for matching the IR specs.

So I'm open to any suggestion about this implementation, including removing it

csrc/multidevice/executor.cpp
samnordmann: I understand, np. Just be aware that without it there is a collision in the names: `postResharding` as a standalone function and `MultiDeviceExecutor::postResharding`

csrc/multidevice/communication.cpp
samnordmann: I reverted to `std::unique` but notice that I need to copy the team because `unique` doesn't preserve const-ness. Let me know if you prefer another implementation. The goal is simply to check that there are no duplicates

csrc/multidevice/communication.cpp
```diff
@@ -40,56 +73,125 @@ inline void assertBuffersHaveSameSize(
...
-Communication::Communication(CommParams params, std::string name, bool has_root)
-    : params_(std::move(params)),
-      collective_type_(std::move(name)),
-      has_root_(has_root) {
-  assertBuffersHaveSameSize(params_.src_bufs, params_.dst_bufs);
+inline void assertValid(
+    const CommParams& params,
+    const DeviceIdxType my_device_index) {
+  assertBuffersHaveSameSize(params.src_bufs, params.dst_bufs);
+  std::unordered_set<DeviceIdxType> team_without_duplicates(
```
cowanmeg, line 97: Is DeviceIdxType the device id or the index of a device in the Mesh/Team? I know it's not caused by this PR, but I think DeviceIdxType is used as both and can cause confusion.
For example, the later function `DeviceIdxType getRootRelativeIndex` uses a DeviceIdxType root to calculate the relative index which is also a DeviceIdxType. 

wdyt about something like a `DeviceId` and `DeviceIndex`?

DeviceId root = 3;
std::vector<DeviceId> team = {1, 3, 10};
DeviceIndex root_index = getRootRelativeIndex(...) // 1

cowanmeg: Also, since this isn't directly related to the PR, we can address this later

csrc/multidevice/communication.cpp
```diff
@@ -107,214 +209,132 @@ std::string Communication::toString(int indent) const {
...
+  return (
+      p1.type == p2.type && (!hasRoot(p1.type) || p1.root == p2.root) &&
+      p1.team == p2.team && (!isReduction(p1.type) || p1.redOp == p2.redOp));
 }
 
-c10::intrusive_ptr<c10d::Work> Scatter::post(
-    Communicator& comm,
-    std::optional<CommunicatorBackend> backend) {
-  post_common(*this, comm);
+c10::intrusive_ptr<c10d::Work> postSingleCommunication(
```
cowanmeg, line 232: Just confirming, this is moving all the <Collective>::post logic into `postSingleCommunication` with the logic remaining the same?
Minus all validation checks which got moved into their own function.

cowanmeg (approved):
LGTM assuming that the post and validate logic is just moved around! If there is changes in those let me know so I can look more closely. 

Also it is great to see Send/Recv finally removed from communicator!

csrc/multidevice/communication.cpp
wujingyue: > Is DeviceIdxType the device id or the index of a device in the Mesh/Team?

That would definitely be confusing. Can you remind me when we use it to mean the latter? I only saw it used to mean the former.

csrc/multidevice/communication.cpp
samnordmann: Yes, there is no functional change in this PR, only some refactor and reorganization. The checks have all been moved to an auxiliary function `assertValid`

csrc/multidevice/communication.cpp
samnordmann: Very good point. I fixed that by changing the return type of `getRootRelativeIndex`. I think this is the only relevant place

> Can you remind me when we use it to mean the latter? I only saw it used to mean the former.

the "root" passed to process group needs to be the relative index of the deviceId within the Team. Does it make sense ?

samnordmann:
!build --dist

csrc/multidevice/communication.cpp
wujingyue: `Expr::sameAs` calls `Expr::sameOp`, which checks equality of all attributes. I believe the communication type is one of the attributes.

wujingyue (approved):
LGTM! I'll try to merge this. I may need to split this up into multiple PRs, but will let you know.

wujingyue:
I'm in the process of resolving conflicts. I should be able to merge this tomorrow or Monday.

csrc/multidevice/communication.cpp
samnordmann: Correct me if Im wrong, but in the current implementation it is not an attribute. The reason is that, IIUC, all attributes need to be `Statement*`s, and `CommunicationParams` is not.

samnordmann:
> I'm in the process of resolving conflicts. I should be able to merge this tomorrow or Monday.

Thank you! I'm waiting for this one to be merge before going on with host ir dev. Let me know if I can help!

csrc/multidevice/communication.cpp
wujingyue: There is addDataAttribute as used in https://github.com/NVIDIA/Fuser/blob/730abb552e5e95ac3476a2a6e9b60bfe114de2c0/csrc/ir/nodes.cpp#L2423-L2424. But the PR as is is already hard to merge, I'll clean that up in a separate PR.

wujingyue:
!build --dist

PR #2186, opened 2024-05-02
[codediff] Include c++ benchmarks with no kernels

jacobhinkle:
This gracefully handles cases where we are re-using kernels, which I think will fix the faulty diffs in #2163.

Previously when a benchmark did not result in a compiled kernel, we did not record the benchmark at all. That means many tests would appear as removed.


jacobhinkle:
!build --diff-bench

jacobhinkle:
!build --diff-bench

xwang233 (approved):

PR #2191, opened 2024-05-02
mark used scheduler in python benchmarks

liqiangxl:
Marke the used scheduler in python benchmarks, so when a scheduler is changed, we can select which benchmark to run.
For example, this PR marked 4 cases as inner_outer_persistent.
If `inner_outer_persistent` scheduler is changed, we can select the corresponding benchmarks with `pytest /opt/pytorch/nvfuser/benchmarks/python/  -m inner_outer_persistent`

### DIFF
--- a/benchmarks/python/conftest.py
+++ b/benchmarks/python/conftest.py
@@ -70,7 +70,9 @@ def pytest_configure(config):
     BENCHMARK_CONFIG["warmup_rounds"] = int(
         config.getoption("--benchmark-warmup-rounds")
     )
-
+    config.addinivalue_line(
+        "markers", "inner_outer_persistent: mark tests using inner_outer_persistent scheduler."
+    )
 
 def pytest_collection_modifyitems(session, config, items):
     """
--- a/benchmarks/python/conftest.py
+++ b/benchmarks/python/conftest.py
@@ -72,7 +72,7 @@ def pytest_configure(config):
     )
     config.addinivalue_line(
         "markers",
-        "inner_outer_persistent: mark tests using inner_outer_persistent scheduler.",
+        "inner_outer_persistent: mark tests using inner_outer_persistent scheduler if not being segmented.",
     )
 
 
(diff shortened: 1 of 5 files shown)


naoyam:
Is this something like tagging each benchmark with a scheduler name? If so, can you be more specific on what each tag should mean and how it should be used?

liqiangxl:
> Is this something like tagging each benchmark with a scheduler name? If so, can you be more specific on what each tag should mean and how it should be used?

yes. For example, in this PR, tag 'inner_outer_persistent' means the corresponding test is primarily using the innerOuter persistent scheduler if not segmented due to input size. 
we should use it to select the corresponding benchmark cases when innerOuter persistent scheudler is changed.

naoyam:
Thanks. That seems useful. Should we add the tag to all the other benchmarks as well? 

Please update the benchmark doc as well: https://github.com/NVIDIA/Fuser/wiki/nvFuser-Python-Benchmarks

Do you have any feedback, @Priya2698?

benchmarks/python/conftest.py
```diff
@@ -70,6 +70,10 @@ def pytest_configure(config):
     BENCHMARK_CONFIG["warmup_rounds"] = int(
         config.getoption("--benchmark-warmup-rounds")
     )
+    config.addinivalue_line(
+        "markers",
+        "inner_outer_persistent: mark tests using inner_outer_persistent scheduler.",
```
Priya2698, line 75: Can we add `if not segmented due to input size` to the description for clarity.

Priya2698:
> Thanks. That seems useful. Should we add the tag to all the other benchmarks as well?
> 
> Please update the benchmark doc as well: https://github.com/NVIDIA/Fuser/wiki/nvFuser-Python-Benchmarks
> 
> Do you have any feedback, @Priya2698?

This looks very useful. We can add these tags to all benchmarks. Thanks, @liqiangxl

PR #2117, opened 2024-04-19
Remove MultiDeviceEnvironment for simplicity.

wujingyue:

PR #2192, opened 2024-05-02
Make it easier to `gdb` an `mpirun`.

wujingyue:
This implements https://www.sci.utah.edu/~tfogal/academic/Fogal-ParallelDebugging.pdf, authored by @tfogal.

I tested this manually.

stdout from the test being debugged:
```
Process 16662 is waiting for debugger. To continue debugging, start gdb, `attach 16662`, `set var waiting=false`, and `fini`.
Process 16662 finished waiting.
```

stdout from gdb:
```
(gdb) attach 16662
Attaching to program: /opt/pytorch/nvfuser/build/test_multidevice, process 16662
[New LWP 16667]
`/opt/pytorch/nvfuser/build/test_multidevice' has changed; re-reading symbols.
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/x86_64-linux-gnu/libthread_db.so.1".
0x000056388b5e7fc5 in nvfuser::MultiDeviceTest::waitForDebugger (this=0x7f7e2c41d800) at /opt/pytorch/nvfuser/tests/cpp/multidevice.cpp:66
66          while (waiting) { // Please change `waiting` in the debugger.
(gdb) set var waiting=false
(gdb) fini
Run till exit from #0  0x000056388b5e7fc5 in nvfuser::MultiDeviceTest::waitForDebugger (this=0x7f7e2c41d800) at /opt/pytorch/nvfuser/tests/cpp/multidevice.cpp:66
[New Thread 0x7f7e2bd84000 (LWP 16706)]
[New Thread 0x7f7e20f7b000 (LWP 16707)]
[New Thread 0x7f7e0bf7b000 (LWP 16708)]
[New Thread 0x7f7e0b77a000 (LWP 16709)]
[New Thread 0x7f7e02fff000 (LWP 16710)]
[New Thread 0x7f7df0fde000 (LWP 16711)]
[New Thread 0x7f7db4fff000 (LWP 16713)]
[New Thread 0x7f7db47fe000 (LWP 16719)]
[New Thread 0x7f7db38fc000 (LWP 16722)]
[Thread 0x7f7df0fde000 (LWP 16711) exited]
nvfuser::MultiDeviceTest::MultiDeviceTest (this=0x7f7e2c41d800) at /opt/pytorch/nvfuser/tests/cpp/multidevice.cpp:37
37      }
```

### DIFF
--- a/tests/cpp/multidevice.cpp
+++ b/tests/cpp/multidevice.cpp
@@ -30,16 +30,21 @@ MultiDeviceTest::MultiDeviceTest() {
        communicator->is_available());
   disable_skip = getNvFuserEnv("MULTIDEVICE_DISABLE_SKIP") != nullptr;
 
-  static std::once_flag once;
-  std::call_once(once, [this]() {
-    // Catch exceptions so call_once always flips `once` and executes this
-    // functor only once.
-    try {
-      waitForDebugger();
-    } catch (const std::exception& e) {
-      TORCH_WARN("Failed to wait for debugger: ", e.what());
-    }
-  });
+  char* rank_to_debug_str = getNvFuserEnv("MULTIDEVICE_WAIT_DEBUGGER_AT_RANK");
+  if (rank_to_debug_str != nullptr) {
+    const DeviceIdxType rank_to_debug = std::stol(rank_to_debug_str);
+
+    static std::once_flag once;
+    std::call_once(once, [&]() {
+      // Catch exceptions so call_once always flips `once` and executes this
+      // functor only once.
+      try {
+        waitForDebuggerAtRank(rank_to_debug);
+      } catch (const std::exception& e) {
+        TORCH_WARN("Failed to wait for debugger: ", e.what());
+      }
+    });
+  }
 }
 
 MultiDeviceTest::~MultiDeviceTest() {
@@ -48,22 +53,16 @@ MultiDeviceTest::~MultiDeviceTest() {
   }
 }
 
-void MultiDeviceTest::waitForDebugger() {
-  char* rank_to_debug_str = getNvFuserEnv("MULTIDEVICE_WAIT_DEBUGGER_AT_RANK");
-  if (rank_to_debug_str == nullptr) {
-    return;
-  }
-
-  const DeviceIdxType rank_to_debug = std::stol(rank_to_debug_str);
+void MultiDeviceTest::waitForDebuggerAtRank(const DeviceIdxType rank) {
   NVF_CHECK(
-      rank_to_debug >= 0 && rank_to_debug < communicator->size(),
-      "rank_to_debug=",
-      rank_to_debug,
+      rank >= 0 && rank < communicator->size(),
+      "rank=",
+      rank,
       " must be in the range of [0,",
       communicator->size(),
       ").");
 
-  if (communicator->deviceId() == rank_to_debug) {
+  if (communicator->deviceId() == rank) {
     volatile bool waiting = true;
     auto pid = getpid();
     std::cerr << "Process " << pid
@@ -74,7 +73,10 @@ void MultiDeviceTest::waitForDebugger() {
     }
     std::cerr << "Process " << getpid() << " finished waiting." << std::endl;
   }
-  communicator->barrier();
+
+  if (communicator->is_available()) {
+    communicator->barrier();
+  }
 }
 
 void MultiDeviceTest::SetUp() {
--- a/tests/cpp/multidevice.cpp
+++ b/tests/cpp/multidevice.cpp
@@ -5,6 +5,9 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 // clang-format on
+#include <sys/types.h>
+#include <unistd.h>
+
 #include <fusion_segmenter.h>
 #include <ir/all_nodes.h>
 #include <multidevice/utils.h>
@@ -25,12 +28,46 @@ MultiDeviceTest::MultiDeviceTest() {
       (getNvFuserEnv("MULTIDEVICE_DEBUG_BARRIER") != nullptr &&
        communicator->is_available());
   disable_skip = getNvFuserEnv("MULTIDEVICE_DISABLE_SKIP") != nullptr;
+
+  // Currently, we attempt to wait for debugger before every test, which is
+  // fine because we usually debug one test at a time using --gtest_filter. As
+  // a potential improvement, call it at the test program level:
+  // https://github.com/google/googletest/blob/main/docs/advanced.md#global-set-up-and-tear-down
+  waitForDebugger();
 }
 
 MultiDeviceTest::~MultiDeviceTest() {
   if (do_barrier_at_test && communicator->is_available()) {
     communicator->barrier();
   }
+}
+
+void MultiDeviceTest::waitForDebugger() {
+  char* rank_to_debug_str = getNvFuserEnv("MPI_DEBUG_RANK");
+  if (rank_to_debug_str == nullptr) {
+    return;
+  }
+  const DeviceIdxType rank_to_debug = std::stol(rank_to_debug_str);
+  NVF_CHECK(
+      rank_to_debug >= 0 && rank_to_debug < communicator->size(),
+      "rank_to_debug=",
+      rank_to_debug,
+      " must be in the range of [0,",
+      communicator->size(),
+      ").");
+
+  if (communicator->deviceId() == rank_to_debug) {
+    volatile bool waiting = true;
+    auto pid = getpid();
+    std::cerr << "Process " << pid
+              << " is waiting for the debugger. To continue debugging, "
+              << "start gdb, `attach " << pid
+              << "`, `set var waiting=false`, and `fini`." << std::endl;
+    while (waiting) { // Please change `waiting` in the debugger.
+    }
+    std::cerr << "Process " << getpid() << " finished waiting." << std::endl;
+  }
+  communicator->barrier();
 }
 
 void MultiDeviceTest::SetUp() {
--- a/tests/cpp/multidevice.cpp
+++ b/tests/cpp/multidevice.cpp
@@ -7,6 +7,7 @@
 // clang-format on
 #include <sys/types.h>
 #include <unistd.h>
+#include <mutex>
 
 #include <fusion_segmenter.h>
 #include <ir/all_nodes.h>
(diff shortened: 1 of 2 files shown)


tests/cpp/multidevice.cpp
```diff
@@ -25,6 +28,12 @@ MultiDeviceTest::MultiDeviceTest() {
       (getNvFuserEnv("MULTIDEVICE_DEBUG_BARRIER") != nullptr &&
        communicator->is_available());
   disable_skip = getNvFuserEnv("MULTIDEVICE_DISABLE_SKIP") != nullptr;
+
+  // Currently, we attempt to wait for debugger before every test, which is
+  // fine because we usually debug one test at a time using --gtest_filter. As
+  // a potential improvement, call it at the test program level:
+  // https://github.com/google/googletest/blob/main/docs/advanced.md#global-set-up-and-tear-down
```
tfogal, line 35: Thank you for calling this out with a comment.

This seems like it would be better, in particular because we already have the `do_barrier_at_test` variable. Presumably somebody added that variable because sometimes we want that barrier and sometimes we don't.

But a difficulty is that one can't `waitForDebugger` until the process has gotten through `MPI_Init`. Is that guaranteed the SetUpTestSuite time?

Anyway, I'd like to see one or both of:
* If we want to go this route, some comments / a discussion (with @samnordmann and @cowanmeg) about whether we should remove the `do_barrier_at_test` logic (since this effectively forces it on).
* A switch to the SetUpTestSuite once-per-suite and not once-per-test approach referenced in the comment

tfogal (approved):
> This implements https://arxiv.org/pdf/2205.05198, authored by @tfogal.

Not sure I can take credit for reducing activation recomputation 😁

tests/cpp/multidevice.cpp
wujingyue: Great suggestions -- thank you! 

> since this effectively forces it on

To be more accurate, this effectively forces it on when NVFUSER_MPI_DEBUG_RANK is specified; otherwise, waitForDebugger returns immediately. IIUC, for debugging, it's better to turn on the barrier anyway. 

> A switch to the SetUpTestSuite once-per-suite and not once-per-test approach referenced in the comment

Let me look into how difficult this is.

samnordmann: > This seems like it would be better, in particular because we already have the `do_barrier_at_test` variable. Presumably somebody added that variable because sometimes we want that barrier and sometimes we don't.

> If we want to go this route, some comments / a discussion (with @samnordmann and @cowanmeg) about whether we should remove the do_barrier_at_test logic (since this effectively forces it on).

Why is it a problem to combine the barrier and the present feature?

By the way, I recall that the barrier is non-blocking from the host perspective. Nccl implements it as an Allreduce.

> But a difficulty is that one can't waitForDebugger until the process has gotten through MPI_Init. Is that guaranteed by SetUpTestSuite time?

There is no `MPI_Init` in the program. However I don't think this is a problem: when `waitForDebugger` is called, the `communicator` is already instantiated (through `getOrCreateCommunicator`), so the backend is already available and we can post a barrier.

> > A switch to the SetUpTestSuite once-per-suite and not once-per-test approach referenced in the comment

> Let me look into how difficult this is.

Idea: this could probably be achieved with gtest's environment (that has been removed recently https://github.com/NVIDIA/Fuser/pull/2117), during environment's setup.

tests/cpp/multidevice.cpp
```diff
@@ -33,6 +42,34 @@ MultiDeviceTest::~MultiDeviceTest() {
...
+              << "start gdb, `attach " << pid
+              << "`, `set var waiting=false`, and `fini`." << std::endl;
+    while (waiting) { // Please change `waiting` in the debugger.
+    }
+    std::cerr << "Process " << getpid() << " finished waiting." << std::endl;
+  }
+  communicator->barrier();
+}
+
 void MultiDeviceTest::SetUp() {
```
samnordmann, line 73: I probably missed it from a previous PR, but why do we now have a class constructor and a `SetUp` function? I think it would be more readable to keep everything in the same place.

tests/cpp/multidevice.cpp
```diff
@@ -33,6 +42,34 @@ MultiDeviceTest::~MultiDeviceTest() {
...
+    auto pid = getpid();
+    std::cerr << "Process " << pid
+              << " is waiting for the debugger. To continue debugging, "
+              << "start gdb, `attach " << pid
+              << "`, `set var waiting=false`, and `fini`." << std::endl;
+    while (waiting) { // Please change `waiting` in the debugger.
+    }
+    std::cerr << "Process " << getpid() << " finished waiting." << std::endl;
+  }
+  communicator->barrier();
```
samnordmann, line 70: We probably need to check whether the communicator `is_available()` before calling the barrier

samnordmann: Also, probably not for this PR but, I think we should remove the method `barrier` from the communicator class. It was initially implemented there for convenience but I think that this interface is not needed anymore

tests/cpp/multidevice.cpp
```diff
@@ -33,6 +42,34 @@ MultiDeviceTest::~MultiDeviceTest() {
   }
 }
 
+void MultiDeviceTest::waitForDebugger() {
+  char* rank_to_debug_str = getNvFuserEnv("MPI_DEBUG_RANK");
```
samnordmann, line 46: what do you think of having a way to make all ranks wait? E.g. by setting the variable to -1

samnordmann: I am not sure the word "MPI" should appear explicitely here, since we don't exactly rely on MPI. Also, I point out that the other options in the multidevice test start with the prefix `MULTIDEVICE_`.
Maybe something like `MULTIDEVICE_WAIT_DEBUGGER_AT_RANK`?

samnordmann (commented):
This is good to have, thanks!

tests/cpp/multidevice.cpp
```diff
@@ -33,6 +42,34 @@ MultiDeviceTest::~MultiDeviceTest() {
...
+  if (rank_to_debug_str == nullptr) {
```
samnordmann, line 47: nit: I Ieave it up to you but I would find it more readable to have a test surrounding the call of `waitForDebugger`, i.e., inside test SetUp
```
if (getNvFuserEnv("MPI_DEBUG_RANK")) {
  waitForDebugger();
}
```

tests/cpp/multidevice.cpp
tfogal: > Great suggestions -- thank you!
> 
> > since this effectively forces it on
> 
> To be more accurate, this effectively forces it on when NVFUSER_MPI_DEBUG_RANK is specified; otherwise, waitForDebugger returns immediately. IIUC, for debugging, it's better to turn on the barrier anyway.

Sorry, the "it" I used was indeed ambiguous :-)

I meant that this basically forces `do_barrier_at_test` on. But I see now that I misread, and it never reaches the `communication->barrier()` when the env var is unset, because of an early exit.

> Why is it a problem to combine the barrier and the present feature?

It's not, given my misread of the code.

tests/cpp/multidevice.cpp
wujingyue: gtest [advocates](https://google.github.io/googletest/faq.html#CtorVsSetUp) for constructors for good reasons. It doesn't explicitly speak about both-constructor-and-setup (good question!), but AFAIU the pros and cons mentioned there do still apply. So I changed SetUp do GTEST_SKIP only because that cannot be done in a constructor, and moved the rest to the constructor.

tests/cpp/multidevice.cpp
wujingyue: > we should remove the method barrier from the communicator class

Interesting -- what do you suggest is the replacement? (This PR does need a way to block all processes until gdb is attached to the process to debug).

tests/cpp/multidevice.cpp
wujingyue: > Maybe something like MULTIDEVICE_WAIT_DEBUGGER_AT_RANK?

SG! 

> what do you think of having a way to make all ranks wait? E.g. by setting the variable to -1

I don't need that now but I'm happy to see future PRs addressing that! I'd also consider to specify a subset of ranks (e.g. "1,2,4") and to specify all ranks as a special string (e.g. "all").

tests/cpp/multidevice.cpp
cowanmeg: See my comment here (https://github.com/NVIDIA/Fuser/pull/2196#issuecomment-2093323035), but having barrier is very useful for testing/debugging, especially in the PipelineTestTwoStages.

cowanmeg: We call `wait` on `barrier`'s work object, so I believe the CPU process will be blocked until the barrier completes? https://github.com/NVIDIA/Fuser/blob/e86bab86cdcd503e72cc9e67f9b3f49d5926cc39/csrc/multidevice/communicator.h#L85

This was from ProcessGroupNCCL's barrier documentation, all the unit test needs is stream synchronization so we should be fine?
> ProcessGroupNCCL now relies on stream synchronization instead of device synchronization to block the CPU. Thus, please do not assume that barrier() would perform a device synchronization.

tests/cpp/multidevice.cpp
wujingyue: I changed the code to call waitForDebugger only once by using std::call_once.

wujingyue:
!build --dist

tests/cpp/multidevice.cpp
samnordmann: > gtest [advocates](https://google.github.io/googletest/faq.html#CtorVsSetUp) for constructors for good reasons. It doesn't explicitly speak about both-constructor-and-setup (good question!), but AFAIU the pros and cons mentioned there do still apply.

IIUC, the two reasons why they advocate for constructors/destructors over SetUp/TearDown  are 1) using `const` and 2) not having to call parent's class SetUp/TearDown.
But in the present case, none of these motivations applies, right?

tests/cpp/multidevice.cpp
samnordmann: My comment is very incidental and minor -- I was simply suggesting a simplification of the interface by removing it from the class `Communicator`, since it is implemented easily https://github.com/NVIDIA/Fuser/blob/4e8356915fc6ffa823cf97fed4249cc00f317fdf/csrc/multidevice/communicator.h#L86

But again, that's something minor that we can do later!

samnordmann (approved):

tests/cpp/multidevice.cpp
wujingyue: > using const

[These fields](https://github.com/NVIDIA/Fuser/blob/4e8356915fc6ffa823cf97fed4249cc00f317fdf/tests/cpp/multidevice.h#L36-L46) in MultiDeviceTest should have been constant. 

> not having to call parent's class SetUp/TearDown.

You are right in this particular instance. It's still a good practice to minimize SetUp/TearDown because people might not understand this intricacy before copying the code and tend to overuse SetUp. (It's especially tempting to overuse SetUp because people tend to incorrectly think SetUp is per test while the constructor is per test fixture)

cowanmeg (approved):

PR #2194, opened 2024-05-03
Fix printed PTX for hopper mma

zasdfgbnm:
Printed code:
```diff
4,5c4,5
<     "  setp.ne.b32 p0, %5, 0;\n"
<     "  wgmma.mma_async.sync.aligned.m64n8k16.f32.bf16.bf16 {%0, %1, %2, %3}, {%4, %5, %6, %7}, %8, p0, %9, %10, %11;\n"
---
>     "  setp.ne.b32 p0, %9, 0;\n"
>     "  wgmma.mma_async.sync.aligned.m64n8k16.f32.bf16.bf16 {%0, %1, %2, %3}, {%4, %5, %6, %7}, %8, p0, %10, %11, %12;\n"
```

As a result of this bug, the `transB` of the RS variant of mmas is always `1`, and the `transA` of the SS variant of mmas is also always 1, and the `transB` is `transA`. Because previously the mma tests are green, this means that their schedules are wrong but consistent with this wrong PTX, and we need to fix them. This is left as a future task.

### DIFF
--- a/csrc/codegen.cpp
+++ b/csrc/codegen.cpp
@@ -3053,17 +3053,21 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
       code_ << "\"{\\n\"\n";
       int64_t boolean_counter = 0;
       int64_t counter = 0;
-      for (auto input : asm_->inputs()) {
-        if (input->dtype() == DataType::Bool) {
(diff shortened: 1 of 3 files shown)


zasdfgbnm:
cc: @drzejan2

zasdfgbnm:
!build

naoyam (approved):
LGTM

PR #2195, opened 2024-05-03
Specify supported gcc versions as the officially supported versions

zasdfgbnm:
GCC officially provides a list of versions being supported. We should use the same list. This list will be a guidance for our decisions on things like, should we upgrade C++20?

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,10 +38,23 @@ PyPI: [https://pypi.org/project/nvfuser/](https://pypi.org/search/?q=nvfuser)
 Docs: https://github.com/NVIDIA/Fuser/wiki
 
 Supported compilers:
-- gcc 11.4+
-- clang14+
+
+**GCC:**
+
+We support all "supported releases" of gcc as specified in [the official site](https://gcc.gnu.org/).
+Currently, they are:
+
+- gcc 11.4
+- gcc 12.3
+- gcc 13.2
+- gcc 14.1
+
+**Clang:**
+
+- clang 14+
 
 Supported C++ standard:
+
 - C++17
 - C++20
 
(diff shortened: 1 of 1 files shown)


zasdfgbnm:
I am waiting for this list to change from `gcc 12.3` to `gcc 12.4` (not released yet) so we can continue our C++20 upgrade plan.

jjsjann123 (approved):
stamping for doc!

Out of my own curiosity, have we actually verified all the gcc versions listed here?

And are we going to add CI for each gcc/clang version, rather than relying on our developer to maintain and check...

README.md
```diff
@@ -38,10 +38,23 @@ PyPI: [https://pypi.org/project/nvfuser/](https://pypi.org/search/?q=nvfuser)
 Docs: https://github.com/NVIDIA/Fuser/wiki
 
 Supported compilers:
-- gcc 11.4+
-- clang14+
+
+**GCC:**
+
+We support all "supported releases" of gcc as specified in [the official site](https://gcc.gnu.org/).
+Currently, they are:
```
wujingyue, line 45:
```suggestion
As of 5/3/2024, they are:
```

wujingyue (approved):

zasdfgbnm:
> Out of my own curiosity, have we actually verified all the gcc versions listed here?

I personally used all of these versions, so at least at the time when I was using it, it worked. Don't know how it does for now.

> And are we going to add CI for each gcc/clang version, rather than relying on our developer to maintain and check...

It would be good to have one. Not urgent though.

I mostly consider this as a guidance when we have a conflict: for example, if at some point, we want to use some language feature and it fails on gcc X, should we claim this gcc is not supported and ask people to upgrade/downgrade, or should we give up moving forward with this language feature?

PR #2196, opened 2024-05-03
Do barrier at test unconditionally.

wujingyue:
On dgx1v-loki-21, the wall-time difference is merely 132 seconds vs 136 seconds.

### DIFF
--- a/tests/cpp/multidevice.cpp
+++ b/tests/cpp/multidevice.cpp
@@ -21,14 +21,11 @@ MultiDeviceTest::MultiDeviceTest() {
   tensor_options =
       at::TensorOptions().dtype(at::kFloat).device(communicator->device());
   debug_print = getNvFuserEnv("MULTIDEVICE_DEBUG_PRINT") != nullptr;
-  do_barrier_at_test =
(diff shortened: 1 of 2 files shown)


samnordmann:
> 2\. It's useful to have the ability to test having many "inflight" collectives

nevermind, this is not relevant. I recall that "barrier" is non blocking from the CPU side.
However, because of that, I am not sure adding the barrier makes debugging easier (if that was the motivation)

cowanmeg:
> > 2. It's useful to have the ability to test having many "inflight" collectives
> 
> nevermind, this is not relevant. I recall that "barrier" is non blocking from the CPU side. However, because of that, I am not sure adding the barrier makes debugging easier (if that was the motivation)

The barrier synchronizes the GPUs (I forgot what flags were set in ProcesGroup if barrier was async or not wrt CPU). We have had test failures where a single GPU hangs/errors out while the others GPU finish. This causes a problem because the other GPUs eventually timeout waiting for the hung GPU in a _different_ test. If they all synchronize at the barrier then the faulty test case will show up as failed on all processes.

cowanmeg (approved):

samnordmann (approved):

PR #2200, opened 2024-05-04
Fix TensorView::clearReductionIterDomains specifies wrong contiguity flag

jjsjann123:
`TensorView::clearReductionIterDomains` constructs new contiguity vector on root domain, causing validation assert when output's allocation domain is not identical to its root domain.

This PR patches the logic and adds a test.

### DIFF
--- a/csrc/tensor_view.cpp
+++ b/csrc/tensor_view.cpp
@@ -1198,8 +1198,8 @@ void TensorView::clearReductionIterDomains() {
       getLeafDomain() == getRootDomain(),
       "should not call clearReductionIterDomains on already transformed TensorDomains");
 
-  std::vector<IterDomain*> root = getRootDomain();
-  std::vector<IterDomain*> alloc = getMaybeAllocationDomain();
+  const std::vector<IterDomain*>& root = getRootDomain();
+  const std::vector<IterDomain*>& alloc = getMaybeAllocationDomain();
 
   NVF_ERROR(
       std::is_permutation(root.begin(), root.end(), alloc.begin(), alloc.end()),
@@ -1208,13 +1208,13 @@ void TensorView::clearReductionIterDomains() {
   std::vector<IterDomain*> new_root;
   std::vector<IterDomain*> new_alloc;
   std::vector<std::optional<bool>> new_contig;
-  for (const auto i : c10::irange(getRootDomain().size())) {
-    auto root_i = getRootDomain().at(i);
+  for (const auto i : c10::irange(root.size())) {
+    auto root_i = root.at(i);
     if (!root_i->isReduction()) {
       new_root.push_back(root_i);
     }
     // contig flag is specified for on allocation domain
-    auto alloc_i = getAllocationDomain().at(i);
+    auto alloc_i = alloc.at(i);
     if (!alloc_i->isReduction()) {
       new_alloc.push_back(alloc_i);
       new_contig.push_back(domain()->contiguity().at(i));
--- a/csrc/tensor_view.cpp
+++ b/csrc/tensor_view.cpp
@@ -1198,17 +1198,40 @@ void TensorView::clearReductionIterDomains() {
       getLeafDomain() == getRootDomain(),
       "should not call clearReductionIterDomains on already transformed TensorDomains");
 
-  std::vector<IterDomain*> new_root;
-  std::vector<std::optional<bool>> new_contig;
-  for (const auto i : c10::irange(getRootDomain().size())) {
-    auto root_i = getRootDomain().at(i);
-    if (!root_i->isReduction()) {
-      new_root.push_back(root_i);
-      new_contig.push_back(domain()->contiguity().at(i));
+  if (domain()->hasAllocation()) {
+    NVF_ERROR(
+        std::is_permutation(getLeafDomain().begin(), getRootDomain().end(),
+                            getAllocationDomain().begin(), getAllocationDomain().end(),
+        "should not call clearReductionIterDomains on transformed allocation domain");
+    std::vector<IterDomain*> new_root;
+    std::vector<IterDomain*> new_alloc;
+    std::vector<std::optional<bool>> new_contig;
+    for (const auto i : c10::irange(getRootDomain().size())) {
+      auto root_i = getRootDomain().at(i);
+      if (!root_i->isReduction()) {
+        new_root.push_back(root_i);
+      }
+      auto alloc_i = getRootDomain().at(i);
+      if (!alloc_i->isReduction()) {
+        new_alloc.push_back(root_i);
+        new_contig.push_back(domain()->contiguity().at(i));
+      }
     }
-  }
 
-  setDomain(IrBuilder::create<TensorDomain>(container(), new_root, new_contig));
+    setDomain(IrBuilder::create<TensorDomain>(container(), new_root, {}, new_alloc, {}, new_contig));
+  } else {
+    std::vector<IterDomain*> new_root;
+    std::vector<std::optional<bool>> new_contig;
+    for (const auto i : c10::irange(getRootDomain().size())) {
+      auto root_i = getRootDomain().at(i);
+      if (!root_i->isReduction()) {
+        new_root.push_back(root_i);
+        new_contig.push_back(domain()->contiguity().at(i));
+      }
+    }
+
+    setDomain(IrBuilder::create<TensorDomain>(container(), new_root, new_contig));
+  }
 }
 
 void TensorView::doubleBuffer() {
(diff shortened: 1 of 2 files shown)


jjsjann123:
!build

jjsjann123:
!build

jjsjann123:
!build

jjsjann123:
patching issue exposed in #2168

jacobhinkle (commented):
Looks good to me, but I want to make sure I understand. Allocation domain can contain any type of `IterDomain` including `IterType::Reduction` and `IterType::Broadcast`. Contiguity refers specifically to `TensorDomain::noReductions(tv->getAllocationDomain())`. Is there a reason we use `std::nullopt` for broadcast contiguity whereas we don't represent reduction axes at all?

tests/cpp/test_allocation_domain.cpp
```diff
@@ -1423,4 +1423,19 @@ TEST_F(AllocationDomainTest, ReductionVectorization) {
...
+  auto tv0 = TensorViewBuilder()
+                 .ndims(3)
+                 .shape({-1, 1, -1})
+                 .contiguity({true, std::nullopt, true})
+                 .build();
+  auto tv1 = sum(tv0, {0});
+  tv1->setAllocationDomain(
+      {tv1->axis(0), tv1->axis(2), tv1->axis(1)},
+      {std::nullopt, true, std::nullopt});
+  tv1->clearReductionIterDomains();
```
wujingyue, line 1438: Is it possible to come up with a better test? I assume the user of nvFuser API (Python or csrc/ops) never calls clearReductionIterDomains. So the test as is seems to be testing an internal implementation that's subject to change, not an API behavior.

csrc/tensor_view.cpp
```diff
@@ -1198,17 +1198,43 @@ void TensorView::clearReductionIterDomains() {
       getLeafDomain() == getRootDomain(),
       "should not call clearReductionIterDomains on already transformed TensorDomains");
 
+  std::vector<IterDomain*> root = getRootDomain();
+  std::vector<IterDomain*> alloc = getMaybeAllocationDomain();
```
wujingyue, line 1202: IIUC const& should be good enough. The code below doesn't modify these two variables.

csrc/tensor_view.cpp
```diff
@@ -1198,17 +1198,43 @@ void TensorView::clearReductionIterDomains() {
...
+
+  NVF_ERROR(
+      std::is_permutation(root.begin(), root.end(), alloc.begin(), alloc.end()),
+      "should not call clearReductionIterDomains on transformed allocation domain");
+
   std::vector<IterDomain*> new_root;
+  std::vector<IterDomain*> new_alloc;
   std::vector<std::optional<bool>> new_contig;
   for (const auto i : c10::irange(getRootDomain().size())) {
```
wujingyue, line 1211:
```suggestion
  for (const auto i : c10::irange(root.size())) {
```

csrc/tensor_view.cpp
```diff
@@ -1198,17 +1198,43 @@ void TensorView::clearReductionIterDomains() {
...
     auto root_i = getRootDomain().at(i);
     if (!root_i->isReduction()) {
       new_root.push_back(root_i);
+    }
+    // contig flag is specified for on allocation domain
+    auto alloc_i = getAllocationDomain().at(i);
```
wujingyue, line 1217: Is allocation domain always specified?

csrc/tensor_view.cpp
naoyam, line 1202: nit:
```suggestion
  const std::vector<IterDomain*>& root = getRootDomain();
  const std::vector<IterDomain*>& alloc = getMaybeAllocationDomain();
```

tests/cpp/test_allocation_domain.cpp
naoyam, line 1438: Shouldn't this do some sort of validation?

csrc/tensor_view.cpp
jjsjann123: Good catch. This one should use `alloc.at(i)` instead.

tests/cpp/test_allocation_domain.cpp
jjsjann123: will do

jjsjann123: > Is it possible to come up with a better test?

I like having a tight-fitting test to go with a patch PR.

`TensorView::clearReductionIterDomain` is a public API. So I think it's fair to have this as a *test that there's no assert* thing. But I agree that it's not that interesting a thing to put here. But as Naoya suggested below, if I put validation here, then it's a bad test on asserting internal implementation.

So I agree with @wujingyue here, I'll dig up the other PR for an end-2-end test that goes through `FusionExecutorCache`.

jjsjann123:
> Contiguity refers specifically to `TensorDomain::noReductions(tv->getAllocationDomain())`. Is there a reason we use `std::nullopt` for broadcast contiguity whereas we don't represent reduction axes at all?

I don't think our protocol trates broadcast and reduction differently regarding contiguity. https://github.com/NVIDIA/Fuser/blob/2dbf1d0a36cdd947f2f65094f7713e3fefba0329/csrc/ir/nodes.cpp#L3201-L3219

tests/cpp/test_allocation_domain.cpp
naoyam: I'm not sure if I understand here correctly, but why not just validating the root and allocation domains are set as intended?

jacobhinkle:
> I don't think our protocol trates broadcast and reduction differently regarding contiguity.

You're right. I was confused but I understand now. They're both treated the same as you say. This code is skipping reduction axes in the new contiguity because this method is meant to strip out all the reduction domains.

csrc/tensor_view.cpp
```diff
@@ -1198,17 +1198,43 @@ void TensorView::clearReductionIterDomains() {
       getLeafDomain() == getRootDomain(),
       "should not call clearReductionIterDomains on already transformed TensorDomains");
 
+  const std::vector<IterDomain*>& root = getRootDomain();
+  const std::vector<IterDomain*>& alloc = getMaybeAllocationDomain();
+
+  NVF_ERROR(
+      std::is_permutation(root.begin(), root.end(), alloc.begin(), alloc.end()),
+      "should not call clearReductionIterDomains on transformed allocation domain");
```
wujingyue, line 1206: Won't this break too soon with @zasdfgbnm 's TMA work?

wujingyue: This may just be my ignorance on when clearReductionIterDomain is called.

jjsjann123: I think the assumption in this API call is that

`"should not call clearReductionIterDomains on already transformed TensorDomains"`

It does have an existing check on `leafdomain` being the same as with `rootdomain`.. which I think hints that there's no TMADomain neither.

wujingyue: Got it. It's currently used only on outputs of fusion segments, for which the assumption holds.

tests/cpp/test_allocation_domain.cpp
jjsjann123: > I'm not sure if I understand here correctly, but why not just validating the root and allocation domains are set as intended?

I think for this specific API, it's a fairly reasonable validation. i.e. Reduction should be dropped and contiguity flags updated as well. @wujingyue was arguing that `clearReductionIterDomains` is not a publicly exposed API. So the test isn't meaningful. Does adding the validation suggested by @naoyam make more sense for us to keep the simple test?

jjsjann123:
!build

tests/cpp/test_allocation_domain.cpp
wujingyue: I'm OK with the test as is. My only complaint was/is that this test can be fragile *if* we later want to use a different way to process the output IterDomains for Welford. We might not even have a counterpart API to test against.

wujingyue (approved):

tests/cpp/test_allocation_domain.cpp
jjsjann123: > My only complaint was/is that this test can be fragile if we later want to use a different way to process the output IterDomains for Welford. We might not even have a counterpart API to test against.

Yeah that part is understood and I agree that we should have better test coverage for reduction/normalization scheduler in general.

The issue patched in this PR is just one of the issues discovered in the refactor work #2168.
I have #2202 to track patching allocation domain support in reduction scheduler, which I think would be a better place to put functional end-2-end tests with the scheduler.

jjsjann123:
merging with green CI. Reviewer's concern on test case was tracked in issue #2202

PR #2203, opened 2024-05-05
ensure coalesced gmem access when each thread has 32 bytes

liqiangxl:
In iter grouped grid reduction,threads with `TIDy == 0` in each block, write block reduction results to global memory for final grid reduction. 
When vectorized by 8, each thread writes 8 x sizeof(fp32) = 32 bytes, which exceeds the max vect write width of 16 bytes. In current main branch, the write is split into 2 vectoried writes without changing the data layout in global memory. It leads to a problem similar to using AoS. As shown in the figure, to serve these 256 bytes of data, the main branch needs 4 transactions but if we change to the bottom version, only needs 2 transactions.
image

After this change, the gmem store transaction is reduced but **no apparent influence on performance** since the memory bandwidth is dominated by the loading of the input tensor. The cache also helps the transaction in the approach used in the current main branch. ncu shows the store transactions is reduced by half.
image

### DIFF
--- a/runtime/grid_reduction.cu
+++ b/runtime/grid_reduction.cu
@@ -667,6 +667,36 @@ __device__ void serialReductionStep(
   }
 }
 
+// check required transactions based on data type and vectorization factor
+// ensure each thread in each transaction has no more than 16 bytes which
+// is the maximum allowed vectorization width.
+template <typename T, int vec_size>
+constexpr __device__ int getTransactions() {
+  constexpr int total_bytes = vec_size * sizeof(T);
+  return total_bytes <= 16 ? 1 : total_bytes / 16;
+}
+
+template <typename T, int vec_size>
+constexpr __device__ int getElementsPerTransaction() {
+  return vec_size * sizeof(T) <= 16 ? vec_size : 16 / sizeof(T);
+}
+
+// calculate elements per section
+__inline__ __device__ nvfuser_index_t getElementsPerSection(
+    nvfuser_index_t row_len,
+    nvfuser_index_t col_len,
+    nvfuser_index_t elements_per_thread) {
+  return row_len * col_len * elements_per_thread;
+}
+
+// calculate offset within a section
+__inline__ __device__ nvfuser_index_t getOffsetWithinSection(
+    nvfuser_index_t row_len,
+    nvfuser_index_t row_id,
+    nvfuser_index_t col_id,
+    nvfuser_index_t elements_per_thread) {
+  return (row_id * row_len + col_id) * elements_per_thread;
+}
 // vectorized reduction
 template <
     bool X_THREAD,
@@ -683,7 +713,7 @@ __device__ void iterGroupedGridReduceLastBlock(
         grid_reduction_segment_size, // Number of reductions across
                                      // grid reduce dimensions
     const nvfuser_index_t
-        block_reduction_segment_size, // Number of reductions across the block
+        block_segment_size, // Number of reductions across the block
     Func reduction_op,
     T* shared_buf,
     bool write_pred,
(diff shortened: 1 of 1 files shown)


liqiangxl:
!build

liqiangxl:
!build --pybench

runtime/grid_reduction.cu
```diff
@@ -857,6 +878,30 @@ __device__ void iterGroupedGridReduce(
...
             threadIdx, blockDim);
 
+    // Max vectorized load/store size is 16 bytes, if each thread has more than
+    // 16 bytes, split into multiple sections to ensure each thread occupies
+    // only 16 bytes at most. For example, if each thread has 8 fp32 which
+    // occupies 32 bytes, split into 2 sections, in each secdtion each thread
+    // holds 4 fp32 or 16 bytes. Thread-0 processes elements [0,7], the first 4
+    // elements [0,3] are stored in the first section and the last 4 elements
+    // [4,7] are stored in the 2nd section. The data layout in gmem is:
+    //         |-----------section 1-----------|-----------section 2-----------|
```
jjsjann123, line 888: IIUC, section 1 is the first load and section 2 is the second.
`gmem_offset_inter` is the stride for each load (length of a section).
`gmem_offset_intra` is the offset to the beginning of the section.

jjsjann123: It's not a comment just trying to verify I was reading the code correctly.

jjsjann123 (commented):
Code change looks straight forward. Thanks for catching the issue.

cc'ing @naoyam for a quick sanity check

runtime/grid_reduction.cu
```diff
@@ -725,6 +725,30 @@ __device__ void iterGroupedGridReduceLastBlock(
...
+  // 2nd section. The data layout in gmem is:
+  //         |-----------section 1-----------|-----------section 2-----------|
+  // TIDx:   |000|001|002|003|004|005|006|007|000|001|002|003|004|005|006|007|
+  // GMEM:   |000|016|032|048|064|080|096|112|128|144|160|176|192|208|224|240|
+  // Element:|000|008|016|024|032|040|048|056|004|012|020|028|036|044|052|060|
+  // This layout ensures coalesced access to gmem and each transaction loads 128
+  // bytes.
+  constexpr int total_bytes = vec_size * sizeof(T);
+  constexpr int n_loads = total_bytes <= 16 ? 1 : total_bytes / 16;
+  constexpr int n_elements = total_bytes <= 16 ? vec_size : 16 / sizeof(T);
```
naoyam, line 744:
```suggestion
  constexpr int n_elements_per_transaction = total_bytes <= 16 ? vec_size : 16 / sizeof(T);
```

runtime/grid_reduction.cu
```diff
@@ -725,6 +725,30 @@ __device__ void iterGroupedGridReduceLastBlock(
...
+  // stored in the first section and the last 4 elements [4,7] are stored in the
+  // 2nd section. The data layout in gmem is:
+  //         |-----------section 1-----------|-----------section 2-----------|
+  // TIDx:   |000|001|002|003|004|005|006|007|000|001|002|003|004|005|006|007|
+  // GMEM:   |000|016|032|048|064|080|096|112|128|144|160|176|192|208|224|240|
+  // Element:|000|008|016|024|032|040|048|056|004|012|020|028|036|044|052|060|
+  // This layout ensures coalesced access to gmem and each transaction loads 128
+  // bytes.
+  constexpr int total_bytes = vec_size * sizeof(T);
+  constexpr int n_loads = total_bytes <= 16 ? 1 : total_bytes / 16;
```
naoyam, line 743:
```suggestion
  constexpr int n_transactions = total_bytes <= 16 ? 1 : total_bytes / 16;
```

runtime/grid_reduction.cu
```diff
@@ -857,6 +878,30 @@ __device__ void iterGroupedGridReduce(
         index_utils::maskedOffset<!X_THREAD, !Y_THREAD, !Z_THREAD>(
             threadIdx, blockDim);
 
+    // Max vectorized load/store size is 16 bytes, if each thread has more than
```
naoyam, line 881: Let's make a constexpr function calculate total_bytes, n_loads and n_elements. They are the same here and the lastblock version, right?

runtime/grid_reduction.cu
```diff
@@ -867,15 +912,12 @@ __device__ void iterGroupedGridReduce(
     auto work_buf_offset =
         block_offset * (block_reduction_segment_size * grid_segment_size) +
         block_reduction_segment_size * idx_in_grid_segment + thread_offset;
-    work_buf_offset *= vec_size;
+    unsigned int gmem_offset_intra = work_buf_offset * n_elements;
```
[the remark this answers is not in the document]
liqiangxl: I refactored the code about indexing into a small util fun for clarity.

runtime/grid_reduction.cu
liqiangxl: yes, you are reading the code correctly. I renamed these two variables to more explicitly express their function.
`gmem_offset_inter` --> `elements_per_section`
`gmem_offset_intra` --> `offset_in_section`

runtime/grid_reduction.cu
liqiangxl: yes, added two constexpr functions to calculate n_loads and n_elements seperatelly. I can't directly use std::pair it is not in our runtime.

runtime/grid_reduction.cu
```diff
@@ -683,7 +713,7 @@ __device__ void iterGroupedGridReduceLastBlock(
         grid_reduction_segment_size, // Number of reductions across
                                      // grid reduce dimensions
     const nvfuser_index_t
-        block_reduction_segment_size, // Number of reductions across the block
```
liqiangxl, line 686: In the code, `grid_segment_size` represents `number of reductions in the grid, e.g. gdimx`
`grid_reduction_segment_size` represents `Number of reductions across grid reduce dim, e.g. gdimy`
similarly, we should use `block_segment_size` to represent `Number of reductions across the block, e.g. bdimx`. 
so here, this `block_reduction_segment_size` is renamed to `block_segment_size`, just a rename, no change in logic.

liqiangxl:
!build --pybench

naoyam:
LGTM. I'll let @jjsjann123 to give a stamp. 

The Python benchmark failures are due to timeouts? CC: @xwang233

xwang233:
Yes, there seems to be timeouts on python benchmarks. Let me fix this by using runners with more GPUs.

liqiangxl:
> Yes, there seems to be timeouts on python benchmarks. Let me fix this by using runners with more GPUs.

Thanks. Is the validation tolerance issue fixed on these python benchmarks?

jjsjann123:
Note to myself. I'll do another sanity check later

jjsjann123:
!build

jjsjann123:
failure seems to be unrelated. `ModuleNotFoundError: No module named 'flax'`
Looks like CI container is missing python module.

jjsjann123 (approved):
LGTM

jjsjann123:
There seems to be something fishy about the transformer engine installation in our container. cc'ing @xwang233 
I also run into similar issue running some lightning tests locally that I have to install transformer engine myself. (I think it's complaining about missing transformer_engine_jax)

I'm not sure what in that test triggers the whole import and why it's not happening for others.. :shrug:

PR #2204, opened 2024-05-05
ensure coalesced access to shared memory in block reduction

liqiangxl:
**Issue:**  smem_offset is the offset into shared memory for the current thread. It is calculated as:
` offset = reduction_idx * reduction_size + reduction_tid`.
For outer reduction where TIDy is in the reduction dimension and TIDx is in the iteration dimension and TIDz is not used.
We have `reduction_tid = TIDy` and `reduction_idx = TIDx`. The offset leads to stride access to shared memory since it equivalent to: `offset = TIDx * blockDim.y + TIDy`
**Fix:** To avoid this, we should use the offset based on the indexing of threads within a block.
`offset= threadIdx.x + threadIdx.y * blockDim.x + threadIdx.z * blockDim.x * blockDim.y;`
After this change, the index of the reduction peer element is no longer a simple offset with a stride of 1, it depends on the reduction dim. See code change for details. 

**Results:** 
(1) removes bank conflicts. baseline is main branch, 99% of bank conflicts are removed, smem access efficiency
 `requests  /  wave fronts= 16000 / 16189 = 99%`
image
(2) This fix achieved 10% performance increase for outer reduction with small inputs compared with main branch. For large inputs, the iter grouped version is used, will be addressed in another PR.
image

### DIFF
--- a/runtime/block_reduction.cu
+++ b/runtime/block_reduction.cu
@@ -48,8 +48,21 @@ __device__ void blockReduce(
       index_utils::maskedOffset<!X_REDUCE, !Y_REDUCE, !Z_REDUCE>(
           threadIdx, blockDim);
 
+  // smem_offset is the offset into shared memory for the current thread.
+  // To ensure coalesced access to shared memory, we need to ensure
+  // each transaction is accessing a contiguous block of 128 bytes.
+  // For outer reduction where TIDy is in the reduction dimension and TIDx
+  // is in the iteration dimension and TIDz is not used. We have
+  // reduction_tid = TIDy and reduction_idx = TIDx. If we directly use the
+  // offset based on reduction_tid and reduction_idx, we will have stride
+  // access to shared memory. For example:
+  // offset = reduction_idx * reduction_size + reduction_tid
+  //        = TIDx * blockDim.y + TIDy
+  // To avoid this, we should always use the offset based on the indexing of
+  // threads within a block.
   // Offset into smem for the current thread
-  unsigned int smem_offset = reduction_idx * reduction_size + reduction_tid;
+  unsigned int smem_offset = threadIdx.x + threadIdx.y * blockDim.x +
+      threadIdx.z * blockDim.x * blockDim.y;
 
   // Initialize shared memory
   if (read_pred) {
--- a/runtime/block_reduction.cu
+++ b/runtime/block_reduction.cu
@@ -75,6 +75,7 @@ __device__ void blockReduce(
   // column (with the same TIDx) are considered peers of each other. The
   // distance between an element and its nearest peer is blockDim.x.
   constexpr int num_redu_dims = (int)X_REDUCE + (int)Y_REDUCE + (int)Z_REDUCE;
+  constexpr bool xz_reduce = (num_redu_dims == 2 && !Y_REDUCE);
   // reduction in 3 dimensions, XYZ, stride is 1
   unsigned int peer_stride = 1;
   if (num_redu_dims == 1) {
@@ -91,15 +92,15 @@ __device__ void blockReduce(
     // Reduction in 2 dimensions, only one dimension is not reduced, !X, !Y, !Z
     // If !Z_REDUCE, merge XY, reducing neighbor cols, peer_stride is 1
     // If !X_REDUCE, merge ZY, reducing neighbor rows, peer_stride is blockDim.x
-    // If !Y_REDUCE, if blockDim.y == 1, merge XZ, peer_stride is 1, otherwise
-    //               different strides for different reduction stages, change
-    //               smem layout based on reduction_idx and reduction_tid, may
-    //               have stride access and bank conflict, rare case, may happen
-    //               for batch norm doing multiple reductions per block.
+    // If !Y_REDUCE, if blockDim.y == 1, merge XZ, peer_stride is 1.
+    // otherwise, needs carefully calculate offset to the reduction peer:
+    // (1) redu_offset = reduction_tid + tree_fold_factor
+    // (2) idz = redu_offset / blockDim.x
+    // (3) idx = redu_offset % blockDim.x
+    // (4) smem_offset = idx + threadIdx.y * blockDim.x + idz * blockDim.x *
+    // blockDim.y
     if (!Y_REDUCE) {
-      smem_offset = blockDim.y > 1
-          ? reduction_idx * reduction_size + reduction_tid
-          : smem_offset;
+      peer_stride = 1;
     } else {
       peer_stride = !Z_REDUCE ? 1 : blockDim.x;
     }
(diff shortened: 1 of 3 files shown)


liqiangxl:
!build

liqiangxl:
build --pybench

liqiangxl:
!build --pybench

runtime/block_reduction.cu
```diff
@@ -48,8 +48,58 @@ __device__ void blockReduce(
...
+  //        = TIDx * blockDim.y + TIDy
+  // To avoid this, we should always use the offset based on the indexing of
+  // threads within a block.
   // Offset into smem for the current thread
-  unsigned int smem_offset = reduction_idx * reduction_size + reduction_tid;
+  unsigned int smem_offset = threadIdx.x + threadIdx.y * blockDim.x +
+      threadIdx.z * blockDim.x * blockDim.y;
+
+  // The peer stride is the distance between the current element and its
+  // reduction peer, it depends on the reduction dimension.
```
naoyam, line 72: What does the reduction peer mean?

runtime/block_reduction.cu
```diff
@@ -48,8 +48,58 @@ __device__ void blockReduce(
...
+                           : blockDim.x * blockDim.y;
+  } else if (num_redu_dims == 2) {
+    // Reduction in 2 dimensions, only one dimension is not reduced, !X, !Y, !Z
+    // If !Z_REDUCE, merge XY, reducing neighbor cols, peer_stride is 1
+    // If !X_REDUCE, merge ZY, reducing neighbor rows, peer_stride is blockDim.x
+    // If !Y_REDUCE, if blockDim.y == 1, merge XZ, peer_stride is 1, otherwise
+    //               different strides for different reduction stages, change
+    //               smem layout based on reduction_idx and reduction_tid, may
+    //               have stride access and bank conflict, rare case, may happen
+    //               for batch norm doing multiple reductions per block.
```
naoyam, line 94: Not quite sure what's intended here. Why changing the layout?

runtime/block_reduction.cu
```diff
@@ -63,14 +113,17 @@ __device__ void blockReduce(
   int np2 = 1 << (31 - __clz(reduction_size));
 
   if (reduction_tid < np2 && reduction_tid + np2 < reduction_size) {
-    reduction_op(shared_mem[smem_offset], shared_mem[smem_offset + np2]);
+    reduction_op(
+        shared_mem[smem_offset], shared_mem[smem_offset + np2 * peer_stride]);
```
jjsjann123, line 117: Is the new layout always a net-gain?
`peer_stride != 0` here would mess with the shared load in the tree reduction `shared_mem[smem_offset + np2 * peer_stride]`. With outer reduction, `peer_stride == blockDim.x`, I'm surprised it removed all the bank conflict....

runtime/block_reduction.cu
jjsjann123: IIUC, I think that's just threads participating in reduction. since with the new layout, it's no longer packing all the reduction element at the shared_mem[0, reduction_size].

runtime/block_reduction.cu
liqiangxl: added a figure for reference:
image

runtime/block_reduction.cu
liqiangxl: These two smem load are 2 sepearte transactions. added a figure for reference:
image

runtime/block_reduction.cu
```diff
@@ -48,29 +48,103 @@ __device__ void blockReduce(
...
   if (reduction_tid < np2 && reduction_tid + np2 < reduction_size) {
-    reduction_op(shared_mem[smem_offset], shared_mem[smem_offset + np2]);
+    int peer_offset = smem_offset + np2 * peer_stride;
+    if constexpr (xz_reduce) {
+      if (blockDim.y > 1) {
+        int redu_offset = reduction_tid + np2;
+        int idz = redu_offset / blockDim.x;
+        int idx = redu_offset % blockDim.x;
+        peer_offset =
+            idx + threadIdx.y * blockDim.x + idz * blockDim.x * blockDim.y;
```
jjsjann123, line 127: We have `reduction_tid = threadIdx.z * blockDimx + threadIdx.x`.
and that gives us
`idz = threadIdx.z + (threadIdx.x + np2) / blockDimx`. This non-linear part I feel could lead to bank-conflict?

Does block reduction by itself go through `xz_reduce`? or is it only accessible through grid reduction?

jjsjann123: Actually, let me try to run some concrete number with this one. :expressionless:

liqiangxl: This block reduction supports reduce in X/Y/Z dims, so there are 2^3 = 8 configurations. Assume at least 1 reduction dimension, there are 8 - 1 = 7 different actual configurations. It can do `xz_reduce` by itself, the newly added test `BlockReduction3D` is testing this `xz_reduce`.

I did a solid example for your reference:
image
Color represents different warps, at each stage, different threads within the same warp are accesssing different banks. For example, in the 1st loop,  in the 1st transaction, T0 is loading T0, T1 is loading T1, ..., Tn is loading Tn (where n<128), then in the 2nd transaction, T0 is loading T128, T1 is loading T129, ..., Tn is loading Tn+128 (where n< 128). In the last loop, T0 is loading T0 & T1, T8 is loading T8 & T9, 4 active threads per warp, each thread is loading a different bank. The issue is we have too many idle threads in the last several loops instead of bank conflicts.

liqiangxl: Double checked with ncu using test `BlockReduction3D` with Z,Y,X= [4,8,8], there is no bank conflict.
image

liqiangxl:
As described in the PR description, **there are still minor bank conflicts during grid reduction**. This issue appears to be related to **function calls**. Specifically, the kernel function invokes the grid reduction function, which is **not inlined**. I conducted a preliminary test using the following code::
```
__device__ __noinline__ void dev_func_use_smem(int* reg_val, int tid, int *s)
{
  s[tid] = reg_val[0];
  s[tid] = s[tid] + 1;
  reg_val[0] = s[tid];
  __syncthreads();
}

__global__ void use_smem(int *d, int n)
{
  extern __shared__ int s[];
  int tid = threadIdx.x + threadIdx.y * blockDim.x;
  int self[1];
  self[0] = d[tid];
  dev_func_use_smem(self, tid, s);
  d[tid] = self[0];
}
```
The baseline is without `__noinline__`, so the function call is inlined by the compiler and there is no bank conflict. However, when adding this  `__noinline__` keyword, the compiler won’t inline the device function and there are 7 bank conflicts reported by ncu.
image

runtime/block_reduction.cu
jjsjann123: Thanks for walking through the example.

> idx = (threadIdx.x + np2) % blockDim.x
> peer_offset = idx + threadIdx.y * blockDim.x + idz * blockDim.x * blockDim.y;
> shared_mem[peer_offset]

You are right about the bank conflict part. mod by blockDim.x should be safe, since there's a single tidx in the indexing part... Sorry about the noise.

jjsjann123:
Sounds like we are suspecting that `block_reduction` is getting inlined fine.
```
codegen_kernel(...) {
  blockReduce(...);
}
```

But this pattern is not.
```
void gridReduce(...) {
  blockReduce(...);
}

codegen_kernel(...) {
  grid_reduction(...);
}
```

I'm wondering if the decision is based on the length of the function or just a two level function call. Out of curiosity, does `__forceinline__` work for template and would it be helpful?

liqiangxl:
> __forceinline__



> Sounds like we are suspecting that `block_reduction` is getting inlined fine.
> 
> ```
> codegen_kernel(...) {
>   blockReduce(...);
> }
> ```
> 
> But this pattern is not.
> 
> ```
> void gridReduce(...) {
>   blockReduce(...);
> }
> 
> codegen_kernel(...) {
>   grid_reduction(...);
> }
> ```
> 
> I'm wondering if the decision is based on the length of the function or just a two level function call. Out of curiosity, does `__forceinline__` work for template and would it be helpful?

I double checked several cases and our `grid reduction` is actually inlined! Maybe I was looking at a different kernel. Sorry for the confusion. I checked with Nsight compute people, they told me that the bank conflicts reported by ncu inlcudes two types: (1) access the same bank in shared memory and (2) access the same bank in L1 cache & shared memory.

jjsjann123:
> I double checked several cases and our `grid reduction` is actually inlined! Maybe I was looking at a different kernel. Sorry for the confusion. I checked with Nsight compute people, they told me that the bank conflicts reported by ncu inlcudes two types: (1) access the same bank in shared memory and (2) access the same bank in L1 cache & shared memory.

by "`grid reduction` is acutally inlined!", you mean both the gridReduce and blockReduce are inlined properly, so function invocation is basically out of the window in that scenario?

liqiangxl:
> > I double checked several cases and our `grid reduction` is actually inlined! Maybe I was looking at a different kernel. Sorry for the confusion. I checked with Nsight compute people, they told me that the bank conflicts reported by ncu inlcudes two types: (1) access the same bank in shared memory and (2) access the same bank in L1 cache & shared memory.
> 
> by "`grid reduction` is acutally inlined!", you mean both the gridReduce and blockReduce are inlined properly, so function invocation is basically out of the window in that scenario?

correct. I didn't see funciton call in the SASS code.

liqiangxl:
The key is `the bank conflicts reported by ncu inlcudes two types: (1) access the same bank in shared memory and (2) access the same bank in L1 cache & shared memory.` we can remove (1) but not (2) since L1 is not programmable.

jjsjann123 (commented):
Thanks for following up on the investigation on bank conflict. It's still concerning the root cause of the bank conflict. since we can't/haven't confirm it to be from l1 cache. Can we open an issue for that at least we have it on track.

Meanwhile, this PR looks good to me for improving memory access in shared memory.
I'll let @naoyam to stamp.

jjsjann123 (approved):
Stamping after @naoyam 's approval from offline conversation.

jjsjann123:
!build

naoyam (approved):
LGTM

PR #2205, opened 2024-05-05
remove bank conflict in iter grouped block reduction

liqiangxl:
**Issues:** 
(1) similar to #2203, when vectorized by 8, each thread holds 32 bytes of data, the write to smem is not coalesced
(2) similar to #2204, calculating smem offset using `reduction_tid` and `reduction_idx` leads to stride access for outer reduction.
(3) r/w to smem during the reduction process is not vectorized.

**Fix:** 
(1) similar to #2203, when vectorized by 8, split data into two sections. Each thread in each section has 16 bytes.
(2) similar to #2204, calculating smem offset using index of threads within a block.
(3) ensure vectorized r/w to smem during the reduction process.

**Other minor changes:**
renamed function name from `blockIterGroupedReduce` to `blockIterGroupedYdimReduce` since it is only used in outer reduction where Y dim is used for reduction. Also added check to codegen.

**Investigation of other reduction method:** 
also tried using warp reduction, the kernel latency is not changed much becuase the global memory access is the dominating factor instead of the reduction runtime function. See the [doc](https://docs.google.com/document/d/1IxbiylbeVz6FQ50viOTSn2-YKd5VBSdjmjYWbsTKYTg/edit#heading=h.16fz6k676m4k) for details.

**Results:**
(1) gelu bwd, speedup of this branch over main branch, larger is better. regression is due to the reduce of occupancy from 62% to 47% due to the increased register usage. 
image
(2) outer reduction, similarly, the regression is due to reduced occupancy comes from increased register usage.
image

**Why register usage slightly increased in some cases:**
(1) Ncu analysis shows at case 2048 x 20480, max live register increased from 46 to 49, the register assignd to each thread increased from 48 to 54. Seems like 48 is a critical value, if it needs more than 48 will bump up to the next value. 
(2) Even I remove optimization (1) and (3) and just keep optimization (2) `calculating smem offset using index of threads within a block`,  the register usage still increased from 48 to 54. Further check of PTX and SASS code showed the compiler is automatically vectorizing one of the shared memory load (compiler team says this is expected and is done by LSV, LoadStoreVectorizer).
(3) These small regressions will be fixed in the heuristic revision PR.

schedule paras at case 2048 x 20480:
```
Iteration Domain: blockIdx.x / threadIdx.x / multiple reductions per block / vectorize / factor 2
Inner Reduction Domain: cross block - threadIdx.y / cross grid - blockIdx.y / split grid dim / split grid dimension / unroll / factor 4
Launch Parameters: BlockDim.x = 160, BlockDim.y = 2, BlockDim.z = -1, GridDim.x = -1, GridDim.y = 16, GridDim.z = -1, Smem Size = 0
Compile Parameters: index_type = NotSet, maxrregcount = 255, enable_magic_zero = 1, enable_ptxas_verbose = 0
```


runtime/block_reduction.cu
```diff
@@ -215,96 +215,143 @@ __device__ void blockIterGroupedReduce(
   static_assert(
       N == 2 || N == 4 || N == 8 || N == 16,
       "N should be a valid vectorization factor, one of (2, 4, 8, 16)!");
+  static_assert(
+      !X_REDUCE && Y_REDUCE && !Z_REDUCE, "Only support reduction in Y dim!");
```
jjsjann123, line 219: nitpick here, if this is a static_assert and the logic below is only applied to this case, maybe we should put the assert in `csrc/codegen.cpp::genIterGroupedBlockReduction` and rename this runtime helper?

runtime/block_reduction.cu
```diff
@@ -215,96 +215,143 @@ __device__ void blockIterGroupedReduce(
...
   // Reduce down to nearest power of 2 for the tree reduction:
-  int np2 = 1 << (31 - __clz(reduction_size));
-
   // Perform parallel reduction for each element in the array
+  int np2 = 1 << (31 - __clz(reduction_size));
   if (reduction_tid < np2 && reduction_tid + np2 < reduction_size) {
+    // vectorized load from smem to regs
+    T self[N];
+    T peer[N];
+#pragma unroll
```
jjsjann123, line 260: out of curiosity, do we have any idea how the perf looks like without the `vectorized load from smem to regs`? I'm assuming that you've compared the two and it's winning against the occupancy cause regression?

jjsjann123 (commented):
Naive question:

Given the description here `Ncu analysis shows at case 2048 x 20480, max live register increased from 46 to 49, the register assignd to each thread increased from 48 to 54.`

That doesn't look like a lot of extra register usage.

Though I'm not sure which schedule params would translate to the `N` used in the template.
```
Iteration Domain: blockIdx.x / threadIdx.x / multiple reductions per block / vectorize / factor 2
Inner Reduction Domain: cross block - threadIdx.y / cross grid - blockIdx.y / split grid dim / split grid dimension / unroll / factor 4
Launch Parameters: BlockDim.x = 160, BlockDim.y = 2, BlockDim.z = -1, GridDim.x = -1, GridDim.y = 16, GridDim.z = -1, Smem Size = 0
Compile Parameters: index_type = NotSet, maxrregcount = 255, enable_magic_zero = 1, enable_ptxas_verbose = 0
```

Looking at the kernel, we are trying to maintain (mostly) two arrays of size N with type T. I thought that's going to be more register usage from the code change.

liqiangxl:
> Naive question:
> 
> Given the description here `Ncu analysis shows at case 2048 x 20480, max live register increased from 46 to 49, the register assignd to each thread increased from 48 to 54.`
> 
> That doesn't look like a lot of extra register usage.
> 
> Though I'm not sure which schedule params would translate to the `N` used in the template.
> 
> ```
> Iteration Domain: blockIdx.x / threadIdx.x / multiple reductions per block / vectorize / factor 2
> Inner Reduction Domain: cross block - threadIdx.y / cross grid - blockIdx.y / split grid dim / split grid dimension / unroll / factor 4
> Launch Parameters: BlockDim.x = 160, BlockDim.y = 2, BlockDim.z = -1, GridDim.x = -1, GridDim.y = 16, GridDim.z = -1, Smem Size = 0
> Compile Parameters: index_type = NotSet, maxrregcount = 255, enable_magic_zero = 1, enable_ptxas_verbose = 0
> ```
> 
> Looking at the kernel, we are trying to maintain (mostly) two arrays of size N with type T. I thought that's going to be more register usage from the code change.

`N` corresponds to vectorization factor. The register change is small becuase the original code is unrolled. So, even without vectorization, if the compiler wont' reuse the register we still needs `N` registers to temperarally store the data loaded from shared memory. The attached figure is the PTX code corresponding to 
```
5322	#pragma unroll
5323	    for (int j = 0; j < elements_per_load; ++j) {
5324	        reduction_op(shared_mem[self_offset + j], shared_mem[peer_offset + j]);
5325	      }

```
in this case `elements_per_load = 2`. Two things are interesting:
(1) the load from shared memory `self_offset + j` is automatically vectorized (L504). This is a little surprise.
(2) then two loads from `peer_offset + j` and two adds, two stores to `self_offset + j`

image

For case with vectorized load, the cuda code is long but PTX code is short and clean. Just load self, load peer, sum and store.
image

runtime/block_reduction.cu
liqiangxl: changed accordingly.

runtime/block_reduction.cu
liqiangxl: (1) no noticable change. It's not the dominating factor for the kernel. Also the register usage is not changed without `vectorized load from smem to regs`
(2) the occupancy caused regression will be solved in the heuristic revision PR.

jjsjann123: Follow up Q0: I think in our analysis, we were factoring the regression from the side effect of increasing register usage. Does this mean that the regression is not observed with the version without vectorized load?

Follow up Q1: regarding `the occupancy caused regression will be solved in the heuristic revision PR.` IIUC, the vectorization version is really accompanying the further heuristic change PR. I'm not suggesting that you make the change right now, but I'm wondering if you think we can have a different break down of PRs? I felt it's a bit hard to justify the vectorization change here with only regression but no noticeable benefits at the time.

Thanks for looking into this... it's so hard breaking heuristic changes into smaller PRs and I appreciate your effort.

jjsjann123: sorry I missed the comment below.

I think you are saying that the vectorized ld/st for shared mem isn't contributing much to the increased register usage, but rather from our unrolling. I'm guessing my suggestion above just doesn't make sense then, if the non-vectorized version still suffers from the lowered occupancy.

liqiangxl: I further investigated this issue by only keep fix-2 `(2) similar to https://github.com/NVIDIA/Fuser/pull/2204, calculating smem offset using index of threads within a block.` see this [commit](https://github.com/NVIDIA/Fuser/pull/2205/commits/1aa09d935578c11833cdb0791a3df3f01b965ef7)

Both PTX and SASS code showed vectorized load from shared memory (note that, in this version the cuda code is not vectorized, looks like compiler vectorized by itself!, there are 2 loads, it only vectorized one of them) and increased register usage and lead to regression similar to the version with all 3 fixes. It seems that there is not much we can do to avoid this vectorization and register increase in this PR.

naoyam: Do we want to avoid vectorization? Try inline PTX.

jjsjann123:
QQ regarding this part:
> (2) Even I remove optimization (1) and (3) and just keep optimization (2) calculating smem offset using index of threads within a block, the register usage still increased from 48 to 54.

I thought when we try to revert (1) and (3), we were seeing that perf regression was gone when benchmark against this PR?

liqiangxl:
> QQ regarding this part:
> 
> > (2) Even I remove optimization (1) and (3) and just keep optimization (2) calculating smem offset using index of threads within a block, the register usage still increased from 48 to 54.
> 
> I thought when we try to revert (1) and (3), we were seeing that perf regression was gone when benchmark against this PR?

That result is not correct. When I revert (1) and (3) I didn't make the code correct. After the correction, there is no performance change due to the auto vectorization added by LoadStoreVectorizer.

runtime/block_reduction.cu
liqiangxl: Thanks Naoya. That's a good idea! 
Here I was trying to do a step-by-step implementation and avoid this small regression in the first step. And we needs to add back vectorization in following steps since it removes bank conflicts.

Maybe we can just keep the current implementation with vectorization (so we don't need to add this inline PTX and remove it later). The current small regression will be solved in the heuristic revision. After the heuristic revision, the achieved occupancy is >=50% and performance is higher than the current main branch. what do you think?

runtime/block_reduction.cu
```diff
@@ -191,19 +191,12 @@ __device__ void blockReduce(
...
 template <
-    bool X_REDUCE,
-    bool Y_REDUCE,
-    bool Z_REDUCE,
     bool Aligned,
     int N, // Number of elements per input array
     typename T,
     typename Func>
-__device__ void blockIterGroupedReduce(
+__device__ void blockIterGroupedYdimReduce(
```
naoyam, line 199: Do we have shmoo unit tests for this?

naoyam:
I added a couple of questions to the doc.

liqiangxl:
!build

liqiangxl:
The failed case in `nvfuser-ci/jit_nvfuser_bench_17_8/8 ` with msg `Error from segmentation group 2: c_ids.size() == 1 INTERNAL ASSERT FAILED at "/opt/pytorch/nvfuser/csrc/id_model/id_model.cpp":560,` seems to be caused by the accidently enalbed ID model. Can't reproduce locally after the latest main branch is merged into this branch. @naoyam

liqiangxl:
Checked cpp benchmarks on H100, there are two regressions due to the increase of register usage:
```
NvFuserScheduler_BiasDropoutAddLayernormBwd1_fp32/512/128/1024/manual_time
NvFuserScheduler_BiasDropoutAddLayernormBwd1_tf32/512/128/864/manual_time
Fusion in these cases:
Inputs:
  T0_g[ iS0{i0}, iS1{i2}, iS2{i3} ], float
  T1_g[ iS27{i0}, iS28{i2}, iS29{i3} ], float
  T2_g[ iS30{i0}, iS31{i2}, bS8{1} ], float
  T3_g[ iS35{i0}, iS36{i2}, bS11{1} ], float
Outputs:
  T6_g[ rS18{i0}, rS19{i2}, iS20{i3} ], float
  T8_g[ rS24{i0}, rS25{i2}, iS26{i3} ], float
  T5_g[ iS37{i0}, iS38{i2}, iS39{i3} ], float

%kernel_math {
T6_g[ rS18{i0}, rS19{i2}, iS20{i3} ]
   = reduction( T0_g[ iS0{i0}, iS1{i2}, iS2{i3} ], op = add, initial value = float(0), allreduce = false )
T4_l[ iS32{i0}, iS33{i2}, iS34{i3} ]
   = T1_g[ iS27{i0}, iS28{i2}, iS29{i3} ]
   - T2_g[ iS30{i0}, iS31{i2}, bS8{1} ];
T5_g[ iS37{i0}, iS38{i2}, iS39{i3} ]
   = T4_l[ iS32{i0}, iS33{i2}, iS34{i3} ]
   * T3_g[ iS35{i0}, iS36{i2}, bS11{1} ];
T7_l[ iS21{i0}, iS22{i2}, iS23{i3} ]
   = T0_g[ iS0{i0}, iS1{i2}, iS2{i3} ]
   * T5_g[ iS37{i0}, iS38{i2}, iS39{i3} ];
T8_g[ rS24{i0}, rS25{i2}, iS26{i3} ]
   = reduction( T7_l[ iS21{i0}, iS22{i2}, iS23{i3} ], op = add, initial value = float(0), allreduce = false )
}
```
register usage:
```
With this PR: 70 registers per thread, occupancy = 25%
Before this PR: 64 registers per thread, occupancy = 50%
```
Performance:
Case | Main Branch (GB/s) | This PR-2205 (GB/s) | After Heuristic Revision, PR-2281(GB/s)
-- | -- | -- | --
512/128/1024 | 2570 | 2200 | 2670
512/128/864 | 2296 | 1974 | 2403

</div></b>

liqiangxl:
Checked cpp benchmarks on A100, no regression detected.

jjsjann123 (approved):
Thanks for double checking the recovered performance from regression caused by register usage increase. Looking forward to #2281

jjsjann123:
!build

naoyam (approved):
LGTM

PR #2155, opened 2024-04-27
Remove NVFUSER_DISTRIBUTED. 

wujingyue:
@xwang233 discussed this in the nvFuser-MultiGPU chatroom. At this moment, supporting non-distributed build of pytorch isn't worth the cost of additional CI and using knobs like `NVFUSER_DISTRIBUTED` or `USE_DISTRIBUTED`. 

This PR reverts https://github.com/NVIDIA/Fuser/pull/1711

Feel free to revert this PR if non-distributed build becomes important.

PR #2207, opened 2024-05-06
Revert "Remove NVFUSER_DISTRIBUTED.  (#2155)"

Aidyn-A:
This PR reverts #2155.

The reason for that is we are hitting the following error on Jetson:
```
  File "/opt/pytorch/fuser/setup.py", line 423, in <module>
    main()
  File "/opt/pytorch/fuser/setup.py", line 365, in main
    cmake()
  File "/opt/pytorch/fuser/setup.py", line 303, in cmake
    raise RuntimeError(
RuntimeError: nvFuser requires PyTorch to be built with USE_DISTRIBUTED on.
```

cc @wujingyue @xwang233

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,13 @@ set(NVFUSER_THIRD_PARTY_DIR "${NVFUSER_ROOT}/third_party")
 option(NVFUSER_STANDALONE_BUILD_WITH_UCC "" OFF)
 option(NVFUSER_BUILD_WITH_ASAN "Build nvFuser with asan" OFF)
 
+include(CMakeDependentOption)
+cmake_dependent_option(NVFUSER_DISTRIBUTED "" ON "USE_DISTRIBUTED" OFF)
+if (NVFUSER_DISTRIBUTED)
+  add_compile_definitions(NVFUSER_DISTRIBUTED)
+endif()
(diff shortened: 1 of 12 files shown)


xwang233:
I suggest we use a warning or an optional env var to bypass this check, instead of reverting the entire PR.

xwang233:
@Aidyn-A  can you try if commenting out the check resolves the build issue on Jetson?

wujingyue:
> @Aidyn-A can you try if commenting out the check resolves the build issue on Jetson?

It'll likely cause a linker issue -- declarations in c10d headers are used but not defined. Won't it? 

That being said, I think it's safer and quicker to revert the PR before trying out other solutions.

Aidyn-A:
Yeah, bypassing the check indeed causes linking failures:
```
communicator.cpp:(.text+0x308): undefined reference to `c10d::TCPStore::TCPStore(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, c10d::TCPStoreOptions const&)'
communicator.cpp:(.text+0xfd0): undefined reference to `c10d::PrefixStore::PrefixStore(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, c10::intrusive_ptr<c10d::Store, c10::detail::intrusive_target_default_null_type<c10d::Store> >)'
collect2: error: ld returned 1 exit status
```

wujingyue:
!build --dist

PR #2208, opened 2024-05-06
Unit test for scheduling a Mma op where input B has an allocation domain.

protonu:
As a first step to supporting allocation domain in our matmul scheduler, this unit test demonstrates how we can use reordering to extend scheduling a Mma op when the input Tensorview (B) has an allocation domain.

The next PR on top this will extend the scheduler to support a strided input B in a unit test that schedules a matmul.

### DIFF
--- a/tests/cpp/test_mma.cpp
+++ b/tests/cpp/test_mma.cpp
@@ -130,19 +130,8 @@ std::vector<at::Tensor> scheduleCompileAndRun(
             tvb->getMaybeRFactorDomain(), tvb->getAllocationDomain())
             .value();
 
-    // We have to convert the above permutation to a map of old2new.
-    // For RFactor [M, K, N] -> Alloc: [M, N, K], the map will be
-    // {{0, 0}, {1, 2}, {2, 1}}.
-    std::unordered_map<int64_t, int64_t> reorder_map;
-    int64_t idx = 0;
-    std::transform(
-        b_permutation.begin(),
-        b_permutation.end(),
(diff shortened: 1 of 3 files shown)


jacobhinkle:
Strided here means transposed but still contiguous right? Specifically you are setting the stride order on the input and verifying that we compute the matmul properly.

protonu:
> Strided here means transposed but still contiguous right? Specifically you are setting the stride order on the input and verifying that we compute the matmul properly.

Yes, that's correct. I updated the title to just mention an allocation domain.

tests/cpp/test_mma.cpp
```diff
@@ -85,43 +88,79 @@ class MmaTest : public NVFuserFixtureParamTest<MmaTestParams> {
...
+    std::unordered_map<int64_t, int64_t> reorder_map;
+    int64_t idx = 0;
+    std::transform(
+        b_permutation.begin(),
+        b_permutation.end(),
+        std::inserter(reorder_map, reorder_map.end()),
+        [&idx](const int64_t v) { return std::make_pair(idx++, v); });
+
+    // Reorder the ouput of Mma.
+    tv2->reorder(reorder_map);
```
zasdfgbnm, line 145: Should we just add a `TensorView::reorder(b_permutation)`?

zasdfgbnm (approved):

protonu:
!build

zasdfgbnm (approved):

Issue #2092, opened 2024-04-17
Additional features to fill out `ops.matmul`


Priya2698:
We currently only support 2D inputs for matmul. Extend implementation to support additional cases:

1. Both A and B are 1D (`[M,] x [M,]`)
2. A is 1D and B is 2D (`[M,] x [M, N]`)
4. A is 2D and B is 1D (`[M, N] x [N,]`)
5. A and B are atleast 1D with one of the matrices having more than 2-dimensions (`[B, M, N] x [N,]`).

Torch reference: https://pytorch.org/docs/stable/generated/torch.matmul.html
Thunder reference: https://github.com/Lightning-AI/lightning-thunder/blob/a28575345fcdc18bf4b9163dfb239195dca9f34d/thunder/tests/opinfos.py#L5299

jacobhinkle:
We should also make sure we support more than one batch dim, including for example batch dims between the M and K dims of `A`. Since there's a potential to implement these differently based on `nDims()` we should also be sure to add a benchmark (single input size is fine) comparing bmm with `batch_size=1` to a 2D matmul of the same size as a kind of perf sanity check.

Priya2698:
This issue will be now resolved through work on new IR nodes: https://github.com/NVIDIA/Fuser/issues/2149

Priya2698:
PR #2175 and #2209 add support for all cases accepted by `torch.matmul`.

PR #2209, opened 2024-05-07
ATen scheduler for the new `Matmul/LinearOp` IR nodes

Priya2698:
1. Adds a new scheduler -- `ExprEvalScheduler` that accepts the MatmulOp and LinearOp (next PR) for ATen evaluation.
2. Modify the matmul input generator to test for all cases supported by Thunder.
3. The `eagerMatmul` API is renamed and replaces the existing `matmul` API. `fd.ops.matmul` now creates a `MatmulOp` (except in a few special cases such as scalar dot product, for eg: `[K] x [K]`.

Issue #2149, #2092.


csrc/scheduler/expr_eval_sched.h
```diff
@@ -0,0 +1,66 @@
...
+ private:
+  void computeHeuristics(
+      Fusion* fusion,
+      SchedulerRuntimeInfo& runtime_info,
+      HeuristicSummary* data_cache = nullptr);
+};
+
+//! Provides a dummy heuristic type to ensure
+//!  unified interface on ExprEval scheduler.
+class ExprEvalHeuristic : public HeuristicParams {
```
jacobhinkle, line 50: I'm not sure what the comment here means.

csrc/scheduler/expr_eval_sched.h
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  void schedule(Fusion* fusion) override;
+
+ private:
+  void computeHeuristics(
+      Fusion* fusion,
+      SchedulerRuntimeInfo& runtime_info,
+      HeuristicSummary* data_cache = nullptr);
+};
```
jacobhinkle, line 46: I don't think this is needed.

csrc/scheduler/expr_eval_sched.h
jacobhinkle: Can you just use `HeuristicParams` in place of `ExprEvalHeuristic` or is the problem that that class is pure virtual. You _could_ consider just providing trivial defaults for those virtual methods then you wouldn't need this class.

csrc/scheduler/expr_eval_sched.h
```diff
@@ -0,0 +1,66 @@
...
+  using HeuristicParams::HeuristicParams;
+
+  size_t hash() const override {
+    return 0;
+  }
+  std::shared_ptr<HeuristicParams> clone() const override {
+    return std::make_shared<ExprEvalHeuristic>();
+  }
+  bool sameAs(const std::shared_ptr<HeuristicParams>& other) const override {
+    auto other_casted = std::dynamic_pointer_cast<ReductionParams>(other);
```
jacobhinkle, line 61: Could do this to avoid creating a new `shared_ptr`
```suggestion
    auto* other_casted = dynamic_cast<ReductionParams*>(other.get());
```
Also, `HeuristicParams` is a `PolymorphicBase` so you can use
```c++
  if (!other->isA<ReductionParams>()) {
    return false;
  }
  return other->cparams == cparams;
}
```

jacobhinkle (commented):
Looking great. Sprinkle in a few tests once you merge #2175 and we'll be on our way.

Priya2698:
!build

csrc/ops/composite.h
```diff
@@ -81,6 +71,6 @@ TensorView* leaky_relu(TensorView* x, Val* negative_slope);
 
 NVF_API TensorView* view_as_real(TensorView* x);
 
-TensorView* eagerMatmul(TensorView* tv_a, TensorView* tv_b);
+TensorView* matmul(TensorView* tv_a, TensorView* tv_b);
```
jacobhinkle, line 74: I suggest keeping the documentation in place.
```suggestion
// Matmul function which takes in tensors with the shapes
// A[M,K] B[K,N], but the tensors may have different layouts
// via strides. All restrictions from the matmul APIs also
// apply here.
TensorView* matmul(TensorView* tv_a, TensorView* tv_b);
```

csrc/root_domain_map.cpp
```diff
@@ -150,14 +180,18 @@ std::unordered_map<IterDomain*, IterDomain*> PairwiseRootDomainMap::map(
...
-    const std::vector<IterDomain*>& aligned_producer_id =
+    const std::vector<IterDomain*>& aligned_producer_ids =
         ops::mapMatmulOpIterDomains(producer_root, input_role, out_size);
 
     for (auto inx : c10::irange(out_size)) {
-      IterDomain* map_key_id = aligned_producer_id.at(inx);
-      IterDomain* map_value_id = consumer_root.at(inx);
-      updatePairwiseRootDomainMap(map_key_id, map_value_id);
+      IterDomain* producer_id = aligned_producer_ids.at(inx);
+      IterDomain* consumer_id = consumer_root.at(inx);
```
jacobhinkle, line 188: Good change. These names are much clearer.

csrc/root_domain_map.cpp
```diff
@@ -171,8 +205,6 @@ std::unordered_map<IterDomain*, IterDomain*> PairwiseRootDomainMap::map(
     // 2. IDs that may have different extents (e.g., non indexed
     //  domains of torch_gather)
     // 3. Squeeze and unsqueeze
-    // 4. Broadcast and non broadcast
-    // 5. Symbolic ID with different extent from other ID
```
jacobhinkle, line 175: Why are these being removed? Doesn't this cause breakage?

csrc/scheduler/expr_eval_sched.cpp
```diff
@@ -0,0 +1,47 @@
...
+          heuristicType(), "Fusion must contain a single expression of type MatmulOp");
+  return false;
+}
+
+bool ExprEvalScheduler::canScheduleRunTime(
+    Fusion* fusion,
+    SchedulerRuntimeInfo& runtime_info,
+    HeuristicSummary* data_cache) {
+  return true;
+}
```
jacobhinkle, line 40: nit: I would place this in the header.

csrc/scheduler/expr_eval_sched.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include <scheduler/registry_utils.h>
+
+namespace nvfuser {
+
+ExprEvalScheduler::ExprEvalScheduler(
+    Fusion* fusion,
+    SchedulerRuntimeInfo& runtime_info,
+    HeuristicSummary* data_cache)
+    : SchedulerEntry(heuristicType()) {
+  params_ = std::make_shared<HeuristicParams>("", runtime_info.getIndexType());
```
jacobhinkle, line 21: Do you need this line? Also: I would place this in the header.

csrc/scheduler/heuristic.h
```diff
@@ -25,11 +25,20 @@ class HeuristicParams : public PolymorphicBase {
...
 
-  virtual size_t hash() const = 0;
-
-  virtual bool sameAs(const std::shared_ptr<HeuristicParams>& other) const = 0;
+  virtual size_t hash() const {
+    return 0;
+  };
+
+  virtual bool sameAs(const std::shared_ptr<HeuristicParams>& other) const {
+    if (!other->isA<HeuristicParams>()) {
```
jacobhinkle, line 33: You might want `other->isStrictlyA<HeuristicParams>()` to do a strict check. Otherwise this will be false unless other is `nullptr`.

csrc/scheduler/heuristic_types.h
```diff
@@ -55,11 +55,13 @@ enum class ScheduleHeuristic {
...
   OuterPersistent,
-  Transpose
+  Transpose,
+  ExprEval
 };
 
 //! Define a schedule table to loop over all the heuristics in priority order.
-constexpr std::array<ScheduleHeuristic, 8> all_heuristics_in_priority_order = {
+constexpr std::array<ScheduleHeuristic, 9> all_heuristics_in_priority_order = {
+    ScheduleHeuristic::ExprEval,
```
jacobhinkle, line 64: Should `NoOp` come before `ExprEval`?

tests/python/pytest_input_generators.py
```diff
@@ -1487,8 +1487,37 @@ def vector_at_error_generator(
...
+    cases = (
+        ((K,), (K,)),
+        ((K,), (K, N)),
+        ((M, K), (K,)),
+        ((K,), (B, K, N)),
+        ((B, M, K), (K,)),
+        ((M, K), (K, N)),
+        ((B, M, K), (B, K, N)),
+        ((B, B, M, K), (B, B, K, N)),
+    )
```
jacobhinkle, line 1515: Can we sprinkle in some broadcast dims?

version.txt
```diff
@@ -1 +1 @@
-0.2.3
+0.2.4
```
jacobhinkle, line 1: :+1:

jacobhinkle (changes requested):
Looks good overall. I am just slightly concerned about removing stuff from `PairwiseRootDomainMap::map`.

csrc/root_domain_map.cpp
Priya2698: I moved them within `updatePairwiseDomainMap` since that will be a common check for any special cases. The other 3 conditions were op-specific (select, torch gather etc).

csrc/ops/composite.h
Priya2698: Added a comment detailing the functionality. The above comment is incorrect since we support all sizes now.

csrc/root_domain_map.cpp
jacobhinkle: Ah ok of course. I missed that this is called in those cases now too.

csrc/scheduler/expr_eval_sched.cpp
jacobhinkle: Interesting. I wonder if you could just force `DataType::Int` for the index type always with this scheduler. Note I'm not exactly sure how to do that and it might be more work than this single line.

jacobhinkle: Nevermind, I think this line is fine. You _could_ ignore the runtime_info index type and force int64, but that's orthogonal to whether this line should exist. I think you're good to go.

jacobhinkle (approved):
LGTM after tests pass and you add some broadcasts in test.

csrc/scheduler/heuristic_types.h
Priya2698: Some cases get accepted by NoOp scheduler, which is why I prioritized ExprEval scheduler.

We may need to change the heuristics of `NoOp` if we want to switch the order.

Priya2698:
!build

jjsjann123:
:rocket:

csrc/scheduler/heuristic_types.h
jacobhinkle: Oh! Thanks for mentioning that. Does `NoOp` scheduler accept the cases where you have a single scalar output? Because it seems to me that it would do so based on this code: https://github.com/NVIDIA/Fuser/blob/8c18701e252f279292f9ae1c926b7e919b3ebd17/csrc/fusion.cpp#L341-L359
We should add a special case for zero-dimensional outputs there! I'll send a PR in a moment that does that and flips NoOp to be first again.

jacobhinkle: https://github.com/NVIDIA/Fuser/issues/2241

Priya2698:
```
[  FAILED  ] 6 tests, listed below:
[  FAILED  ] ATenNodesParametrizedTest.MatmulNodeConcrete/2, where GetParam() = ({ 32 }, { 32, 1 })
[  FAILED  ] ATenNodesParametrizedTest.MatmulNodeConcrete/8, where GetParam() = ({ 1, 32 }, { 32 })
[  FAILED  ] ATenNodesParametrizedTest.MatmulNodeConcrete/10, where GetParam() = ({ 1, 32 }, { 32, 1 })
[  FAILED  ] ATenNodesParametrizedTest.MatmulNodeSymbolic/2, where GetParam() = ({ 32 }, { 32, 1 })
[  FAILED  ] ATenNodesParametrizedTest.MatmulNodeSymbolic/8, where GetParam() = ({ 1, 32 }, { 32 })
[  FAILED  ] ATenNodesParametrizedTest.MatmulNodeSymbolic/10, where GetParam() = ({ 1, 32 }, { 32, 1 })
```

It is likely because there no reductions identified since we use ATen, and all the dimensions in the output are broadcast dimensions. So the cases where M/N = 1 get picked by `NoOp`

PR #2210, opened 2024-05-07
Modifying unit test to check if we can manually schedule a matmul where input B has an allocation domain

protonu:
This PR modifies an existing unit test which was scheduling a matmul.
The modified runs the test for the cases:
1.  Input B has the rfactor domain order [N, K]. This was the original test.
2.  Input B has the rfactor domain order [K, N] and allocation domain order [N, K].

The next PR on this stack will modify the matmul scheduler to handle inputs where the input B carries an allocation domain.

### DIFF
--- a/tests/cpp/test_gpu_tensorcore.cpp
+++ b/tests/cpp/test_gpu_tensorcore.cpp
@@ -1671,175 +1671,215 @@ TEST_F(GPUTTensorCoreTest, FusionAmpereViewMatmulTN_CUDA) {
 }
 
 // Test an end-to-end matmul case with swizzled smem
-// data layout.
+// data layout. We run this for two typed in input B tensor.
+// First case B has the rfactor domain [N, K] and the second
+// case, rfactor [K, N] and allocation domain [N, K].
 TEST_F(GPUTTensorCoreTest, FusionAmpereMatmulTNSwizzled_CUDA) {
-  NVFUSER_TEST_CUDA_ARCH_GUARD(8, 0);
+  auto scheduleCompileAndRun = [](bool use_mkn_dim_order) {
+    NVFUSER_TEST_CUDA_ARCH_GUARD(8, 0);
 
-  Fusion fusion;
-  FusionGuard fg(&fusion);
+    Fusion fusion;
+    FusionGuard fg(&fusion);
 
-  int M = 257, N = 511, K = 136;
+    int M = 257, N = 511, K = 136;
 
-  MatMulTileOptions gemm_tile;
-  gemm_tile.cta_tile = GemmTile(128, 128, 32);
-  gemm_tile.warp_tile = GemmTile(64, 64, 32);
-  gemm_tile.instruction_tile = GemmTile(16, 8, 16);
+    MatMulTileOptions gemm_tile;
+    gemm_tile.cta_tile = GemmTile(128, 128, 32);
+    gemm_tile.warp_tile = GemmTile(64, 64, 32);
+    gemm_tile.instruction_tile = GemmTile(16, 8, 16);
 
-  // [M,K]
-  auto tv0 = makeContigTensor(2, DataType::Half);
-  // [N,K]
-  auto tv1 = makeContigTensor(2, DataType::Half);
-  fusion.addInput(tv0);
-  fusion.addInput(tv1);
+    // [M,K]
+    auto tv0 = makeContigTensor(2, DataType::Half);
+    // [N,K] or [K, N] if use_mkn_dim_order is true
+    auto tv1 = makeContigTensor(2, DataType::Half);
+    // Propagate the allocation domain post-broadcast.
+    if (use_mkn_dim_order) {
+      tv1->setAllocationDomain({tv1->axis(1), tv1->axis(0)}, true);
+    }
+    fusion.addInput(tv0);
+    fusion.addInput(tv1);
 
-  // [M,N,K]
-  auto tv0b = broadcast(tv0, {false, true, false});
-  auto tv1b = broadcast(tv1, {true, false, false});
+    // [M,N,K] or [M, K, N] if use_mkn_dim_order is true
+    auto tv0b = use_mkn_dim_order ? broadcast(tv0, {false, false, true})
+                                  : broadcast(tv0, {false, true, false});
+    auto tv1b = broadcast(tv1, {true, false, false});
 
-  auto tv2 = fusedMultiplySum(tv0b, tv1b, {2});
+    // propagate the allocation domain in case there's one.
+    if (use_mkn_dim_order) {
+      tv1b->setAllocationDomain(
+          {tv1b->axis(0), tv1b->axis(2), tv1b->axis(1)}, true);
+    }
 
-  fusion.addOutput(tv2);
+    // [M, K, N] or [M, N, K]
+    auto tv2 = use_mkn_dim_order ? fusedMultiplySum(tv0b, tv1b, {1})
+                                 : fusedMultiplySum(tv0b, tv1b, {2});
 
-  auto mma_ops = ir_utils::getOpsOfType<MmaOp>(&fusion);
-  NVF_CHECK(
-      1 == mma_ops.size(),
-      "Invalid number of MmaOp instances in fusion definition, expected 1, got ",
-      mma_ops.size());
-  mma_ops.front()->setMacro(MmaMacro::Turing_16_8_16);
+    fusion.addOutput(tv2);
 
-  auto tv0cw = tv0->cacheAfter(LoadStoreOpType::CpAsync);
-  auto tv0cr = tv0cw->cacheAfter(LoadStoreOpType::LdMatrix);
-  auto tv1cw = tv1->cacheAfter(LoadStoreOpType::CpAsync);
-  auto tv1cr = tv1cw->cacheAfter(LoadStoreOpType::LdMatrix);
-  auto tv2c = tv2->cacheBefore();
+    auto mma_ops = ir_utils::getOpsOfType<MmaOp>(&fusion);
+    NVF_CHECK(
+        1 == mma_ops.size(),
+        "Invalid number of MmaOp instances in fusion definition, expected 1, got ",
+        mma_ops.size());
+    mma_ops.front()->setMacro(MmaMacro::Turing_16_8_16);
 
-  // Make a CTA tile
-  // ------------------------------------------------------------------
-  // [M,N]
-  tv2->split(-2, gemm_tile.cta_tile.m);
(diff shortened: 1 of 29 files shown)


protonu:
!build

jacobhinkle:
Should we create a new unit test for this instead of removing the existing one?

jacobhinkle:
> so that we can run a matmul scheduler

I understand now what's going on, but I don't think this description is accurate. The matmul scheduler is not used at all in this test. The test is manually scheduling a fusion where instead of preexisting `BroadcastOp` we have allocation domains on the inputs. This could be useful for determining how to insert the needed broadcasts before putting that into the matmul scheduler in a followup PR. Is that accurate?

tests/cpp/test_gpu_tensorcore.cpp
```diff
@@ -1671,175 +1671,215 @@ TEST_F(GPUTTensorCoreTest, FusionAmpereViewMatmulTN_CUDA) {
...
+    auto tv0b = use_mkn_dim_order ? broadcast(tv0, {false, false, true})
+                                  : broadcast(tv0, {false, true, false});
+    auto tv1b = broadcast(tv1, {true, false, false});
 
-  tv1cr->applyMmaSwizzle(MmaOperand::B);
-  tv1b->applyMmaSwizzle(MmaOperand::B);
+    // propagate the allocation domain in case there's one.
+    if (use_mkn_dim_order) {
+      tv1b->setAllocationDomain(
+          {tv1b->axis(0), tv1b->axis(2), tv1b->axis(1)}, true);
```
jacobhinkle, line 1710: Are we able to ignore allocation domain from this point forward and it propagates correctly? Would you mind posting the final `fusion->printTransforms()` and comparing to that in the original test?

tests/cpp/test_gpu_tensorcore.cpp
```diff
@@ -1671,175 +1671,215 @@ TEST_F(GPUTTensorCoreTest, FusionAmpereViewMatmulTN_CUDA) {
...
+    tv2->axis(1)->parallelize(ParallelType::BIDy);
+    tv2->axis(2)->parallelize(ParallelType::TIDz);
+    tv2->axis(3)->parallelize(ParallelType::TIDy);
+
+    tv0cw->doubleBuffer();
+    tv1cw->doubleBuffer();
+    tv0cr->doubleBuffer();
+    tv1cr->doubleBuffer();
+
+    // Create input for either [M, K] @ [N, K] or [M, K] @ [K, N] (with strides)
```
jacobhinkle, line 1858: I think it's a bad habit to write `[M, K] @ [N, K]` as the `@` symbol really has no meaning other than mathematical matrix multiply. Instead you could write something like
> Create input for operand B having stride one in either the K or N dimension

protonu:
> Should we create a new unit test for this instead of removing the existing one?

This doesn't remove an existing test, just modifies it. I didn't want to create a brand new test since I felt that would lead to a mix of more code and refactoring the existing test to get some reuse out of the code

protonu:
> > so that we can run a matmul scheduler
> 
> I understand now what's going on, but I don't think this description is accurate. The matmul scheduler is not used at all in this test. The test is manually scheduling a fusion where instead of preexisting `BroadcastOp` we have allocation domains on the inputs. This could be useful for determining how to insert the needed broadcasts before putting that into the matmul scheduler in a followup PR. Is that accurate?

Sorry for the confusion, and yes, I was using the body of the unit test as a proxy for matmul scheduling.

tests/cpp/test_gpu_tensorcore.cpp
protonu: I am not quite sure what you mean here. 
We create a permutation vector based on the allocation domain for `tv1b` and use it to reorder the output of the mma op. The changes to the output are then propagated back up the fusion.

tests/cpp/test_gpu_tensorcore.cpp
```diff
@@ -1671,175 +1671,215 @@ TEST_F(GPUTTensorCoreTest, FusionAmpereViewMatmulTN_CUDA) {
...
+    auto options = at::TensorOptions().dtype(at::kHalf).device(at::kCUDA, 0);
+    auto t0 = at::randn({M, K}, options);
+    auto t1 = use_mkn_dim_order ? at::randn({K, N}, options)
+                                : at::randn({N, K}, options);
+    if (use_mkn_dim_order) {
+      t1 = t1.as_strided(t1.sizes(), {1, K});
```
jjsjann123, line 1864: nitpick, this can be merged with line 1861. no need for another conditional.

tests/cpp/test_gpu_tensorcore.cpp
jjsjann123: What does this allocation domain supposed to do? how should we check it in the test?

tests/cpp/test_gpu_tensorcore.cpp
```diff
@@ -1671,175 +1671,215 @@ TEST_F(GPUTTensorCoreTest, FusionAmpereViewMatmulTN_CUDA) {
...
+    tv1->computeAt(tv2, 2);
+
+    // Order K
+    //  0   1    2   3     4    5
+    // [Mo,No, M128, N128, Ko, K32]
+    tv2c->split(-1, gemm_tile.cta_tile.k);
+    tv2c->reorder({{2, 3}, {3, 4}, {4, 2}});
+
+    //  0   1  2   3     4    5
+    // [Mo,No, Ko M128, N128, K32]
```
jjsjann123, line 1764: nitpick
```suggestion
    // [Mo,No, Ko, M128, N128, K32]
```

jjsjann123 (commented):
Question on test, how would we validate that proper kernel it picked with different permutation?

tests/cpp/test_gpu_tensorcore.cpp
protonu: I was trying to do that, but ran out of ideas to do it cleanly.  I can add it when building torch options, but that would repeating the use of the options builder a couple of times. What would you suggest?

protonu:
> Question on test, how would we validate that proper kernel it picked with different permutation?

I added checks on the last couple of commits. Please let me if this is what you meant by the question:

If we use the `use_mkn_dim_order_flag:
  check b after broadcast has allocation domain, but a, does not
  get the permutation vector based on broadcasted b's allocation domain, and check if the permutation vector is what we expected.
  
  
Please let me know if you meant something different.

tests/cpp/test_gpu_tensorcore.cpp
jjsjann123: I was merely suggesting either with
```suggestion
    auto t1 = use_mkn_dim_order ? at::randn({K, N}, options).as_strided({K, N}, {1, K})
                                : at::randn({N, K}, options);
```

alternatively, the other case is really just a permuted tensor?
```
t1 = at::randn({N, K}, options);
if (use_mkn_dim_order) {
  t1 = t1.permute({1, 0});
}
```

tests/cpp/test_gpu_tensorcore.cpp
```diff
@@ -1671,175 +1671,215 @@ TEST_F(GPUTTensorCoreTest, FusionAmpereViewMatmulTN_CUDA) {
...
+    auto tv0cr = tv0cw->cacheAfter(LoadStoreOpType::LdMatrix);
+    auto tv1cw = tv1->cacheAfter(LoadStoreOpType::CpAsync);
+    auto tv1cr = tv1cw->cacheAfter(LoadStoreOpType::LdMatrix);
+
+    // In the case that input B has an (transposed [M, N, K] ) allocation domain
+    // and it's rfactor domain in of the order [M, K, N], reorder it to the
+    // expected [M, N, K] and propagate it backwards.
+    if (use_mkn_dim_order) {
+      tv2->reorder({{1, 2}, {2, 1}});
+      scheduler_utils::BoundedDirectionalTransformPropagator::backward(
```
jjsjann123, line 1759: > We create a permutation vector based on the allocation domain for tv1b and use it to reorder the output of the mma op. The changes to the output are then propagated back up the fusion.

I just realized that this is done explicitly in this test.

I think this is where @jacobhinkle was asking. i.e. What does `backward` propagation is supposed to do? are we expecting this backward propagation to change any allocation domain and have we verified that it's still the same test?

My question here is, should we validate the propagation? I'm happy with it as-is.

tests/cpp/test_gpu_tensorcore.cpp
```diff
@@ -1671,175 +1671,236 @@ TEST_F(GPUTTensorCoreTest, FusionAmpereViewMatmulTN_CUDA) {
...
+      auto perm =
+          ir_utils::computePermutation(
+              tv1b->getMaybeRFactorDomain(), tv1b->getAllocationDomain())
+              .value();
+
+      NVF_CHECK(perm.size() == 3, "The permutation vector size is incorrect");
+      // We reorder from [M, K, N] -> [M, N, K], perm:[0, 2, 1]
+      NVF_CHECK(
+          perm.at(0) == 0 && perm.at(1) == 2 && perm.at(2) == 1,
+          "permutation vector is incorrect");
```
jjsjann123, line 1755:
```suggestion
  EXPECT_THAT(perm, ElementsAre(0, 2, 1));
```

I think this is easier to read. You might want to add `using testing::ElementsAre;` at the beginning if it's not already there.

tests/cpp/test_gpu_tensorcore.cpp
jjsjann123: I think this question still holds, is this propagation supposed to change anything and should we validate that?

jjsjann123: NVM, I think Prontonu just mentioned this is used to propagate the re-order. So I guess the scheduling following afterwards is already validating the correctness.

jjsjann123 (commented):
LGTM.

I'll let @jacobhinkle to confirm he get an answer for his question and stamp

jjsjann123:
> The next PR on this stack will modify the matmul scheduler to handle inputs where the input B carries an allocation domain.

BTW, I don't see a stack yet. :stuck_out_tongue_winking_eye:

tests/cpp/test_gpu_tensorcore.cpp
protonu: Yes, the propagation should change the order of the leaf nodes from the output of the mma op all the way to the inputs of the fusion. I can add checks for that.

protonu:
Thanks for reviews - I think I'll close this review given our design for handling allocation domain has evolved quite a bit from the way I was thinking about things here.

PR #2214, opened 2024-05-07
Temporary hopper case for getMmaOp

cowanmeg:
Temporarily have Hopper use Ampere matmul schedules in getMmaOp. 
This is so that distributed matmul tests can pass CI on H100s.


cowanmeg:
!build --dist

zasdfgbnm (approved):

naoyam:
!build

PR #2219, opened 2024-05-08
Skip tests using FusionProfiler when compute-sanitizer is used

naoyam:
`FusionProfiler` and compute-sanitizer both use CUPTI, and running multiple CUPTI tools is not supported. Disabling two tests using FusionProfiler when compute-sanitizer is detected. Since using compute-sanitizer is supplemental and there are only two of affected tests, just disabling seems fine to me.

I didn't know about `dl_iterate_phdr` but just learned from Perplexity.

### DIFF
--- a/csrc/sys_utils.cpp
+++ b/csrc/sys_utils.cpp
@@ -192,7 +192,7 @@ bool detectComputeSanitizer() {
 
 } // namespace nvfuser
 
-#else
+#else // #if defined(__linux__)
 
 namespace nvfuser {
 
@@ -220,4 +220,4 @@ bool detectComputeSanitizer() {
 
 } // namespace nvfuser
 
-#endif
+#endif // #if defined(__linux__)
--- a/csrc/sys_utils.cpp
+++ b/csrc/sys_utils.cpp
@@ -5,6 +5,13 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 // clang-format on
+#if defined(__linux__) && !defined(_GNU_SOURCE)
+// dl_iterate_phdr is only defined when _GNU_SOURCE is defined. The
+// macro needs to be defined before any header file is included. See
+// the man page for more info.
+#define _GNU_SOURCE
+#endif
+
 #include <exceptions.h>
 #include <executor_utils.h>
 #include <sys_utils.h>
@@ -18,16 +25,11 @@
 #include <vector>
 
 #include <dlfcn.h>
+#include <link.h>
 #include <sys/types.h>
 #include <unistd.h>
 #include <cstdio>
 #include <cstdlib>
-// dl_iterate_phdr is only defined when _GNU_SOURCE is defined. See
-// the man page for more info.
-#ifndef _GNU_SOURCE
-#define _GNU_SOURCE
-#endif
-#include <link.h>
 
 namespace nvfuser {
 
--- a/csrc/sys_utils.cpp
+++ b/csrc/sys_utils.cpp
@@ -22,6 +22,8 @@
 #include <unistd.h>
 #include <cstdio>
 #include <cstdlib>
+// dl_iterate_phdr is only defined when _GNU_SOURCE is defined. See
+// the man page for more info.
 #ifndef _GNU_SOURCE
 #define _GNU_SOURCE
 #endif
--- a/csrc/sys_utils.cpp
+++ b/csrc/sys_utils.cpp
@@ -22,6 +22,10 @@
 #include <unistd.h>
 #include <cstdio>
 #include <cstdlib>
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
+#include <link.h>
 
 namespace nvfuser {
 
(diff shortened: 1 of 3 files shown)


naoyam:
!build

csrc/sys_utils.h
```diff
@@ -34,4 +34,7 @@ class LibraryLoader : public NonCopyable {
   void* handle_ = nullptr;
 };
 
+// Return true if compute-sanitizer is attached
+NVF_API bool detectComputeSanitizer();
```
wujingyue, line 38:
```suggestion
bool detectComputeSanitizer();
```

https://github.com/NVIDIA/Fuser/blob/main/doc/dev/visibility.md#should-i-mark-my-new-method-or-class-as-nvf_api

csrc/sys_utils.cpp
```diff
@@ -22,6 +22,10 @@
 #include <unistd.h>
 #include <cstdio>
 #include <cstdlib>
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
```
wujingyue, line 27: Can you code-comment why?

csrc/sys_utils.cpp
```diff
@@ -162,6 +166,28 @@ LibraryLoader::~LibraryLoader() {
...
+
+} // namespace
+
+bool detectComputeSanitizer() {
+  return dl_iterate_phdr(detectComputeSanitizerCallback, nullptr) != 0;
+}
+
 } // namespace nvfuser
 
 #else
```
wujingyue, line 193:
```suggestion
#else // #if defined(__linux__)
```

csrc/sys_utils.cpp
```diff
@@ -185,6 +211,11 @@ LibraryLoader::~LibraryLoader {
...
 }
 
+bool detectComputeSanitizer() {
+  // Not implemented. Just return false for now.
+  return false;
+}
+
 } // namespace nvfuser
 
 #endif
```
wujingyue, line 221:
```suggestion
#endif // #if defined(__linux__)
```

csrc/sys_utils.cpp
wujingyue: It took me many clicks to find the corresponding if.

csrc/sys_utils.cpp
```diff
@@ -162,6 +166,28 @@ LibraryLoader::~LibraryLoader() {
...
+int detectComputeSanitizerCallback(
+    struct dl_phdr_info* info,
+    size_t size,
+    void* data) {
+  std::string lib_name = info->dlpi_name;
+  if (lib_name.find("compute-sanitizer") != std::string::npos) {
+    return 1;
+  } else {
+    return 0;
+  }
```
wujingyue, line 182:
```suggestion
  return lib_name.find("compute-sanitizer") != std::string::npos;
```

tests/cpp/test_alias.cpp
```diff
@@ -1109,6 +1110,10 @@ TEST_F(AliasTest, ReuseBufferAliasAcrossSegments) {
 }
 
 TEST_F(AliasTest, AliasOnlyKernelsAreNotLaunched) {
+  if (detectComputeSanitizer()) {
+    GTEST_SKIP() << "Skipped as compute-sanitzer conflicts with FusionProfiler";
```
wujingyue, line 1114:
```suggestion
    GTEST_SKIP() << "Skipped because compute-sanitizer is detected, which conflicts with FusionProfiler";
```

tests/cpp/test_alias.cpp
```diff
@@ -1179,6 +1184,10 @@ TEST_F(AliasTest, PerfDebugVerboseWhenSomeKernelsNotLaunched) {
 }
 
 TEST_F(AliasTest, NoKernelsAreLaunched) {
+  if (detectComputeSanitizer()) {
+    GTEST_SKIP() << "Skipped as compute-sanitzer conflicts with FusionProfiler";
```
wujingyue, line 1188: ditto

wujingyue (approved):

csrc/sys_utils.cpp
wujingyue: Will this cause incompatibility for any platforms we support or want to support? cc @xwang233

xwang233: I don't know what `_GNU_SOURCE` does

csrc/sys_utils.h
naoyam: Oh, I thought it's necessary to make it visible to the tests.

csrc/sys_utils.cpp
naoyam: Added a comment

csrc/sys_utils.cpp
naoyam: Hahaha, yeah, that actually did hit me first when testing this new function. Thanks for the suggestion!

csrc/sys_utils.cpp
naoyam: It's a GNU extension, so practically speaking, I'd be really surprised if there's any distribution that doesn't support it.

naoyam:
!build

PR #2220, opened 2024-05-08
IdModel: Step 5 of the loop promotion analysis

naoyam:
This is the final step of the loop promotion analysis. The promotion map is almost completed at Step 3, but some partially inlined domains need one more propagation, which is done by Step 4 and Step 5. Step 5 is mostly just a repeat of Step 3.

This basically concludes the loop promotion analysis, although there are a couple of issues that were found while working on indexing (#2218). Those issues will be addressed as further follow-up PRs.

- Step 1: #1650 
- Step 2: #1777 
- Step 3: #1830 
- Step 4: #2003

### DIFF
--- a/csrc/id_model/id_model.h
+++ b/csrc/id_model/id_model.h
@@ -64,6 +64,19 @@ StatefulInliningInfo buildStatefulInliningInfo(
 // considered the exact same size operating on matching dimensions from the root
 // domain mapping.
 //
+// LOOP mode is important to resolve inlined broadcassts. If we have something
+// like: consumer[i0o, threadIdx.x{i0i}] = producer[i0o,
+// threadIdx.y{i0i}](computeAt = 1) which can easily happen when using shared
+// memory. Loop is actually defined for all iteration domains, and resembles
+// groups of iter domains that are effectively inlined with each other.
+// Therefore iter domain's that are a common dependency of inlined leaf domains
+// may be loop mapped together.
+//
+// Loop promotion is a mechanism by which to capture inlined resolved
+// broadcasts. If a consumer resolves a broadcast of a producer, and the
+// producer's broadcast is inlined (in total or partially). Then the producer's
+// iter domain will be "promoted" to the size of the consumers iter domain.
+//
 // IdMappingMode::EXACT
 //   Don't map any broadcast axes to non-broadcast axes
 //   Do not forward through any broadcast IDs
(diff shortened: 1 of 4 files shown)


naoyam:
!build

zasdfgbnm:
Should this be added back?
https://github.com/NVIDIA/Fuser/pull/1968

naoyam:
> Should this be added back? #1968

Thanks. Added back.

csrc/id_model/id_model.h
```diff
@@ -64,6 +64,19 @@ StatefulInliningInfo buildStatefulInliningInfo(
 // considered the exact same size operating on matching dimensions from the root
 // domain mapping.
 //
+// LOOP mode is important to resolve inlined broadcassts. If we have something
+// like: consumer[i0o, threadIdx.x{i0i}] = producer[i0o,
+// threadIdx.y{i0i}](computeAt = 1) which can easily happen when using shared
+// memory. Loop is actually defined for all iteration domains, and resembles
+// groups of iter domains that are effectively inlined with each other.
+// Therefore iter domain's that are a common dependency of inlined leaf domains
+// may be loop mapped together.
```
zasdfgbnm, line 73: I would be very interested in seeing when the loop promotion we have is strictly required, and when it is just one way to do things. For example, if I have a fusion
```
T0[1, 4]
T1[3, 4] = T0[1, 4]

T1->reorder({{0, 1}});
T1->merge(0);
T1->split(2, inner=false);
propagate;
T0->inlineAt(1);
```
Then for this specific case, loop promotion is not necessary, because `(i, j)` in the leaf domain of T1 is `((i*6+j)%3, (i*6+j)/3)` in T1's root domain. For T0, without loop promotion, `(i, k)` in the leaf domain is `(0, i*2+k)` in T0's root domain. According to Theorem 2.15.1 in https://github.com/NVIDIA/Fuser/blob/main/doc/math/integer-division.md, `(i*6+j)/3 = i*2+j/3`, which has very similar mathematical form as `i*2+k`. And this mathematical similarity tells us that each `i` takes the same slice of T0 and T1, therefore the program is valid even without loop promotion.

zasdfgbnm (approved):

csrc/id_model/id_model.h
naoyam: That's very interesting. 🤯

PR #2222, opened 2024-05-09
Add `TensorView::flatten` as a convenient method for merging multiple dims

zasdfgbnm:


zasdfgbnm:
!build

zasdfgbnm:
!build

csrc/tensor_view.cpp
```diff
@@ -565,6 +565,21 @@ TensorView* TensorView::merge(int64_t axis_o, int64_t axis_i) {
   return this;
 }
 
+TensorView* TensorView::flatten(int64_t from, int64_t to) {
+  NVF_ERROR(nDims() > 0, "Tried to do flatten on a 0-dim TensorView");
+  from = wrapDim(from);
+  to = wrapDim(to);
+  if (from > to) {
+    std::swap(from, to);
+  }
```
wujingyue, line 574: This seems to be unnecessarily permissive. Does torch do this?

wujingyue (approved):

csrc/tensor_view.cpp
zasdfgbnm: No PyTorch doesn't. I don't have any use case that uses this "feature". I can remove it here and if it turns out to be inconvenient, I can add it back later.

zasdfgbnm:
!build

csrc/tensor_view.cpp
```diff
@@ -565,6 +565,19 @@ TensorView* TensorView::merge(int64_t axis_o, int64_t axis_i) {
   return this;
 }
 
+TensorView* TensorView::flatten(int64_t from, int64_t to) {
+  NVF_ERROR(nDims() > 0, "Tried to do flatten on a 0-dim TensorView");
+  from = wrapDim(from);
+  to = wrapDim(to);
+  NVF_CHECK(from <= to, "Invalid flatten range. From: ", from, " To: ", to);
```
jacobhinkle, line 572: Another option is to interpret `from > to` as merging from inner to outer instead of outer to inner as is done here. Is there any reason to prefer one direction over the other?

jacobhinkle (approved):
LGTM

csrc/tensor_view.cpp
zasdfgbnm: I don't think there is any reason to prefer one over another. They are just equivalent. And should ideally generate exactly the same code. So is really no need to "customize" which one to use. We just pick any one to use it.

naoyam: It's not just left to right or right to left, but we may want to prefer merging contiguous domains first to promote simpler indexing exprs.

jacobhinkle: > And should ideally generate exactly the same code

Right, due to expression simplification. I remember discussing this previously. Makes sense not to prefer either direction then.

Issue #2213, opened 2024-05-07
Allow reductions in matmul epilogues


jacobhinkle:
We currently support pointwise epilogues. However, there is an opportunity for us to fuse reductions into our epilogues. For example, adding cross-entropy loss function would allow us to reduce total IO considerably (ignoring the possible need to save matrix outputs for the backwards pass). Pytorch has an open issue requesting such a fused op https://github.com/pytorch/pytorch/issues/124480.

I tried the following straightforward approach:
```c++
      Fusion fusion;
      FusionGuard fg(&fusion);

      auto shapes = matmulAtInputShape3DTuring(-1, -1, -1, layout);

      auto tv0 = makeContigConcreteTensor(shapes.first, DataType::Half);
      auto tv1 = makeContigConcreteTensor(shapes.second, DataType::Half);

      fusion.addInput(tv0);
      fusion.addInput(tv1);

      tv0 = canonicalizeInputToBMNK(tv0, layout, MmaOperand::A);
      tv1 = canonicalizeInputToBMNK(tv1, layout, MmaOperand::B);
      auto tv2 = fusedMultiplySum(tv0, tv1, {-1});

      auto tv3 = makeContigConcreteTensor({M}, DataType::Int32);
      fusion.addInput(tv3);
      auto is_true_label =
          eq(broadcast(tv3, {false, true}),
             broadcast(
                 iota(
                     IrBuilder::create<Val>(N),
                     /*start=*/nullptr,
                     /*step=*/nullptr,
                     DataType::Int32),
                 {true, false}));
      // allreduce log sum exp in N dimension
      auto m = max(tv2, {1});
      auto m_bcast = broadcast(m, {false, true});
      auto sumexp = sum(exp(sub(tv2, m_bcast)), {1});
      auto lse_noadj = log(sumexp);
      auto lse_nobcast = add(lse_noadj, m);
      auto lse = broadcast(lse_nobcast, {false, true});

      // xent = lse - y_i*onehot(L)_i
      auto xent = sum(
          sub(lse, where(is_true_label, tv2, fusion.zeroVal(DataType::Float))),
          {1});

      fusion.addOutput(xent);
```
However, this fails during `scheduleOutputTensor`, which expects tiled 2D outputs:
> C++ exception with description "tile_size_m == gemm_tile.cta_tile.m INTERNAL ASSERT FAILED at "/opt/pytorch/nvfuser/csrc/sc
[...]

Issue #2169, opened 2024-05-01
Analyze matmul epilogue to determine vectorization


jacobhinkle:
PR https://github.com/NVIDIA/Fuser/pull/2105 implements operand analysis to determine vectorization of gmem loads. Quote:

 > Are we supporting a general pointwise epilogue fusion, or we only support biases? For example, if I have a "bias" whose shape is `[N, M]`, I transpose it to `[M, N]` and add it to matmul result, can this be handled? It is very important that what we accept into the scheduler is compatible with what we assume here. Building out a general vectorization analysis should refer to the pointwise scheduler and uses code in https://github.com/NVIDIA/Fuser/blob/main/csrc/scheduler/vectorize_helper.h, and this is what https://github.com/NVIDIA/Fuser/pull/807 is doing.

> Regarding this, I believe we have two options:
1. Make sure that we accept only the very limited cases of epilogue fusion (i.e. just with bias and activation) into the schedule and use this simple analysis.
2. Use `vectorize_helper` to build out a complete analysis for pointwise epilogue like the pointwise scheduler.
Whichever option we take, I don't think that is easy and well tested. For option 1, we need to review the scheduler `canScheduleCompileTime` code and brainstorm more adversarial examples, and for option 2, we need to copy some code from the pointwise scheduler like in https://github.com/NVIDIA/Fuser/pull/807. @drzejan2 do you remember the status of https://github.com/NVIDIA/Fuser/pull/807?
> But anyway, epilogue vectorization is a much more difficult task than A and B. Can we move it to a separate PR?

_Originally posted by @zasdfgbnm in https://github.com/NVIDIA/Fuser/pull/2105#discussion_r1585560159_

Option 1 is tracked in #2167. This issue corresponds to option 2 listed above. We might additionally need to update `MatmulParams::SupportedVectorization` if we support different vectorizations for the different input and output tensors

See #807 and #682

Issue #1707, opened 2024-01-31
Segmenter failed to propose the right segment to MatmulScheduler. 


wujingyue:
This causes a "Can not find a scheduler to schedule fusion segment" error. 

The full test can be found in the `wjy/segment_matmul` branch. I'm copying the gist of it here for convenience: 

```
TensorView* a = makeContigTensor(2, DataType::Half);
fusion->addInput(a);
TensorView* b = makeContigTensor(2, DataType::Half);
fusion->addInput(b);

TensorView* c = matmul(
    a, b, MmaLayout::TT, /*turing_or_later=*/true, /*as_mul_sum=*/false);
c = segment_set(c);
c = castOp(DataType::Half, c);
fusion->addOutput(c);
```

@jacobhinkle created a great illustration explaining what happened for a slightly different[^1] test: 
![image](https://github.com/NVIDIA/Fuser/assets/2772612/578ac0db-da4a-41ec-addb-397713f92d64)

[^1]: The figure uses Sum after SegmentSet and the test above uses Cast. However, it's a minor difference that doesn't affect the diagnosis. 

As designed, the segmenter tries to merge one edge at a time. So, it tried to schedule region 4 and region 5 separately, both of which failed [this check](https://github.com/NVIDIA/Fuser/blob/37f5c39385041e5c365a85b49dbc4af80561b059/csrc/scheduler/matmul_utils.cpp#L288) where MatmulScheduler requires both MmaOp inputs to be broadcast. After those failures, the segmenter didn't even attempt to schedule the union of 4 and 5, which would succeed. 

# To reproduce

```shell
git fetch origin wjy/segment_matmul
git checkout wjy/segment_matmul
_bn && bin/test_matmul --gtest_filter=*SegmentBroadcastMatmul 
```

zasdfgbnm:
Self assign for now as I am working on https://github.com/NVIDIA/Fuser/issues/1628 whose solution would help on this.

wujingyue:
Thanks, @zasdfgbnm, for picking this up! 

[...]

PR #2225, opened 2024-05-09
[WIP] Check that matmul epilogue contains only pointwise ops

### LINKED ISSUE #2167: Restrict matmul scheduler to accepting only segments that are supported
Currently, as noted in https://github.com/NVIDIA/Fuser/pull/2105#discussion_r1585560159, we do not inspect matrix epilogues at all in `canScheduleCompileTime`. We should make explicit which cases we support. I believe we can currently expect to compile any matmul segment whose epilogue consists of pointwise operations and whose "C" type inputs are all fully aligned (can check in `canScheduleRuntime`) and have stride order compatible with the output; i.e. (Batch, M, N). We should try and get creative in testing this assumption.

jacobhinkle:
This PR restricts the accepted matmul segments for the nvfuser matmul scheduler to only those containing pointwise epilogues. Additionally, it rules out cases for which we cannot yet reliably determine epilogue input vectorization due to transposes (TODO, see #2169).

Note that this check can be lifted when more epilogue cases are supported, e.g. #2213.

Fixes #2167.

This is stacked on #2175 and follow-up PR to that introducing LinearOp because currently segmentation fails for matmuls unless the complete fusion can be scheduled (see #1707). The MatmulOp and LinearOp IR nodes remove the need to inspect operand producer branches, so segmentation should work fine once that work is merged. This PR will be marked as draft until then.

### DIFF
--- a/csrc/scheduler/matmul_utils.cpp
+++ b/csrc/scheduler/matmul_utils.cpp
@@ -318,6 +318,7 @@ std::string getMatmulCompileTimeRejectReason(Fusion* fusion) {
   // supported inputs layout
   // 3. Check if fusion represents expressions that are recognized by matmul
   // scheduler.
+  // 4. Check that epilogue contains only pointwise ops.
 
   // #1
   // Initializing the machinery to check if there's a Mul-Sum pair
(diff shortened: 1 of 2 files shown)


jacobhinkle:
This is outdated now

PR #2226, opened 2024-05-09
Support allocation domain in the matmul scheduler

protonu:
image

In this discussion when we say the matmul scheduler, we’ll mean the [scheduleMatmul function.](https://github.com/NVIDIA/Fuser/blob/d214286b2e75f59abeba03d5a06ce9fdb93ddc0d/csrc/scheduler/matmul.cpp#L739)

We will also assume that the implementation of the above is tied to the Matmul and Linear Op recently developed. In this note, we’ll focus on the matmul op. 

As shown in the figure above, we will convert the matmul op, which takes in inputs A and B with root domains [M, K] and [M, K] respectively, to a collection of transpose, broadcasts and Mma ops. Please note that the inputs A and B can both have allocation domains (transposed memory layouts - we do not place restrictions on that).

The task at hand focuses on how this collection of ops is handled in the scheduleMatmul function when the inputs have allocation domains. The table below outlines the configuration of inputs A and B we want to support.

<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-87592f69-7fff-92c4-e002-f34db17ce287"><div dir="ltr" style="margin-left:0pt;" align="center">
Root Domain A | Root Domain B | Allocation Domain A | Allocation Domain B
-- | -- | -- | --
[M, K] | [K, N] | [] | []
[M, K] | [K, N] | [] | [N, K]
[M, K] | [K, N] | [K, M] | []
[M, K] | [K, N] | [K, M] | [N, K]

</div></b>

Discussion with @jacobhinkle  and @zasdfgbnm led us to identifying the following issues to get the above cases running:

- Propagate allocation domain from gmem operands to their smem consumers.
   This was solved by Xiang a while back in commit [PR]. 
   (https://github.com/NVIDIA/Fuser/commit/626a405272263978c86f7c4f7d9898d14f494cf4).
- Update mma_utils::orderTiledConcreteIdAsRoot to use allocation domain instead of root.
- Detect transpose properly in scheduleLdMatrix
   This refers to the bit of code [here.](https://github.com/NVIDIA/Fuser/blob/d214286b2e75f59abeba03d5a06ce9fdb93ddc0d/csrc/scheduler/matmul.cpp#L893-L912)
We have to modify how we set LdMatrix vs. LdMatrix based not only on the comparison between rfactorDomain and root domain, but also takes into account the allocation domains of both shared memory or A and B. What I took this to mean is, for the shared memory TV of B, bcw_smem, since it's transposed, we will compare it's rfactor domain and alloction domain or the root domain (if allocation domain isn't there) to decide whether to transpose. And for input A, acw_smem, we'll compare the allocation domain and the root domain to see if we need to transpose. Please correct me if I have this wrong. 
- Update inline_ptx.cpp to detect when to append .trans instead of using a different LoadStoreOpType (optional)
@jacobhinkle  also pointed out that optionally, we may get rid of the enum LdMatrix/LdMatrixTrans and handle the transpose when lowering. 

In this  PR, I prototyped the steps mentioned in the first three bullets above. 
The code here is able to handle all the cases mentioned in the table above.
In a follow-up PR I can handle LdMatrix/LdMatrixTrans.

### DIFF
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -2481,7 +2481,7 @@ std::string LoadStoreOp::toInlineString(int indent_size) const {
 bool LoadStoreOp::hasInnerTranspose() const {
   if (auto out_tv = dynamic_cast<TensorView*>(out())) {
     return out_tv->hasRFactor() &&
-        out_tv->getRootDomain().back() != out_tv->getRFactorDomain().back();
+        out_tv->getMaybeAllocationDomain().back() != out_tv->getRFactorDomain().back();
   }
   return false;
 }
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -2480,11 +2480,8 @@ std::string LoadStoreOp::toInlineString(int indent_size) const {
 
 bool LoadStoreOp::hasInnerTranspose() const {
   if (auto out_tv = dynamic_cast<TensorView*>(out())) {
-    auto use_root_or_alloc = out_tv->hasAllocation()
-        ? out_tv->getAllocationDomain()
-        : out_tv->getRootDomain();
     return out_tv->hasRFactor() &&
-        use_root_or_alloc.back() != out_tv->getRFactorDomain().back();
+        out_tv->getRootDomain().back() != out_tv->getRFactorDomain().back();
   }
   return false;
 }
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -2480,8 +2480,11 @@ std::string LoadStoreOp::toInlineString(int indent_size) const {
 
 bool LoadStoreOp::hasInnerTranspose() const {
   if (auto out_tv = dynamic_cast<TensorView*>(out())) {
+    auto use_root_or_alloc = out_tv->hasAllocation()
+        ? out_tv->getAllocationDomain()
+        : out_tv->getRootDomain();
     return out_tv->hasRFactor() &&
-        out_tv->getMaybeAllocationDomain().back() != out_tv->getRFactorDomain().back();
+        use_root_or_alloc.back() != out_tv->getRFactorDomain().back();
   }
   return false;
 }
--- a/csrc/scheduler/matmul.cpp
+++ b/csrc/scheduler/matmul.cpp
@@ -734,6 +734,10 @@ void scheduleSplitKSum(
   splitk_sum->axis(-1)->parallelize(ParallelType::Vectorize);
 }
 
+// A LoadMatrixTranspose is needed is there's a rfactor which doesn't
+// match the alloc/root domain. If there's no rfactor, then we
+// need a transpose if the alloc domain and root don't match. When we say
+// match, we refer to the innermost iter domain.
 bool needsTranposedLoad(TensorView* out_tv) {
   if (out_tv->hasRFactor()) {
     auto use_root_or_alloc = out_tv->hasAllocation()
@@ -801,7 +805,6 @@ void scheduleMatmul(Fusion* fusion, const MatmulParams& params) {
   const auto mma_layout_opt = mma->layout();
   NVF_ERROR(
       mma_layout_opt.has_value(), "fusion mma op has undefined input layout");
-  // const auto mma_layout = mma_layout_opt.value();
   const auto fusion_layout = mma_utils::getMmaLayout(fusion);
   NVF_ERROR(fusion_layout.isValid(), fusion_layout.getErrorMsg());
 
@@ -908,7 +911,7 @@ void scheduleMatmul(Fusion* fusion, const MatmulParams& params) {
       *tv_r = ldst->out()->as<TensorView>();
       ldst->setOpType(
           needsTranposedLoad(*tv_r) ? LoadStoreOpType::LdMatrixTranspose
-                                   : LoadStoreOpType::LdMatrix);
+                                    : LoadStoreOpType::LdMatrix);
     } else {
       *tv_r = tv_smem->cacheAfter(
           needsTranposedLoad(tv_smem) ? LoadStoreOpType::LdMatrixTranspose
--- a/csrc/scheduler/matmul.cpp
+++ b/csrc/scheduler/matmul.cpp
@@ -734,6 +734,13 @@ void scheduleSplitKSum(
   splitk_sum->axis(-1)->parallelize(ParallelType::Vectorize);
 }
 
+bool hasInnerTranspose(TensorView* out_tv) {
+  auto use_root_or_alloc = out_tv->hasAllocation()
+      ? out_tv->getAllocationDomain()
+      : out_tv->getRootDomain();
+  return use_root_or_alloc.back() != out_tv->getMaybeRFactorDomain().back();
+}
+
 } // namespace
 
 void scheduleMatmul(Fusion* fusion, const MatmulParams& params) {
@@ -898,7 +905,11 @@ void scheduleMatmul(Fusion* fusion, const MatmulParams& params) {
       ldst->setOpType(LoadStoreOpType::LdMatrix);
     }
   } else {
-    acr = acw_smem->cacheAfter(LoadStoreOpType::LdMatrix);
+    if (hasInnerTranspose(acw_smem)) {
+      acr = acw_smem->cacheAfter(LoadStoreOpType::LdMatrixTranspose);
+    } else {
+      acr = acw_smem->cacheAfter(LoadStoreOpType::LdMatrix);
+    }
   }
   if (auto ldst = dynamic_cast<LoadStoreOp*>(bcw_smem->uses().at(0))) {
     bcr = ldst->out()->as<TensorView>();
@@ -908,7 +919,11 @@ void scheduleMatmul(Fusion* fusion, const MatmulParams& params) {
       ldst->setOpType(LoadStoreOpType::LdMatrix);
     }
   } else {
-    bcr = bcw_smem->cacheAfter(LoadStoreOpType::LdMatrix);
+    if (hasInnerTranspose(bcw_smem)) {
+      bcr = bcw_smem->cacheAfter(LoadStoreOpType::LdMatrixTranspose);
+    } else {
+      bcr = bcw_smem->cacheAfter(LoadStoreOpType::LdMatrix);
+    }
   }
 
   // We remove this test for now and add something if required.
--- a/csrc/scheduler/matmul.cpp
+++ b/csrc/scheduler/matmul.cpp
@@ -734,11 +734,15 @@ void scheduleSplitKSum(
   splitk_sum->axis(-1)->parallelize(ParallelType::Vectorize);
 }
 
-bool hasInnerTranspose(TensorView* out_tv) {
-  auto use_root_or_alloc = out_tv->hasAllocation()
-      ? out_tv->getAllocationDomain()
-      : out_tv->getRootDomain();
-  return use_root_or_alloc.back() != out_tv->getMaybeRFactorDomain().back();
+bool needsTranposedLoad(TensorView* out_tv) {
+  if (out_tv->hasRFactor()) {
+    auto use_root_or_alloc = out_tv->hasAllocation()
+        ? out_tv->getAllocationDomain()
+        : out_tv->getRootDomain();
+    return use_root_or_alloc.back() != out_tv->getRFactorDomain().back();
+  }
+  return out_tv->getRootDomain().back() !=
+      out_tv->getMaybeAllocationDomain().back();
 }
 
 } // namespace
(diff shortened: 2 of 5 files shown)


tests/cpp/test_matmul_scheduler.cpp
```diff
@@ -2333,6 +2333,114 @@ TEST_F(MatmulSchedulerPluginTest, BasicMatmul) {
...
+
+  FusionExecutor fe;
+  fe.compileFusion(fusion.get(), {t0, t1}, LaunchParams(), matmul_cparams);
+
+  auto cg_outputs = fe.runFusion({t0, t1});
+  auto tref = t0.to(at::kFloat).matmul(t1.to(at::kFloat));
+  NVF_CHECK(cg_outputs[0].allclose(tref, 0.0001, 0.0001));
+}
+
+TEST_F(MatmulSchedulerPluginTest, SimpleMatmulWithTransposeAndAlloc) {
```
protonu, line 2389: @jacobhinkle  can you take a look at this test.

tests/cpp/test_matmul_scheduler.cpp
```diff
@@ -2333,6 +2333,278 @@ TEST_F(MatmulSchedulerPluginTest, BasicMatmul) {
...
+  FusionExecutor fe;
+  fe.compileFusion(fusion.get(), {t0, t1}, LaunchParams(), matmul_cparams);
+
+  auto cg_outputs = fe.runFusion({t0, t1});
+  auto tref = t0.to(at::kFloat).matmul(t1.to(at::kFloat));
+  NVF_CHECK(cg_outputs[0].allclose(tref, 0.0001, 0.0001));
+}
+
+TEST_F(MatmulSchedulerPluginTest, SimpleMatmulWithTranspose) {
+  NVFUSER_TEST_CUDA_ARCH_RANGE_GUARD(8, 0, 8, 9);
```
jacobhinkle, line 2445: Is the upper bound inclusive for this guard? I had thought it was exclusive.

tests/cpp/test_matmul_scheduler.cpp
```diff
@@ -2333,6 +2333,278 @@ TEST_F(MatmulSchedulerPluginTest, BasicMatmul) {
...
+  // Input -B.
+  tv1->setAllocationDomain({tv1->axis(1), tv1->axis(0)}, true);
+  fusion->addInput(tv0);
+  fusion->addInput(tv1);
+
+  auto tv1t = transpose(tv1); // This has rfactor: {N, K}
+  tv1t->setAllocationDomain({tv1t->axis(0), tv1t->axis(1)}, true);
+  // [M, N, K]
+  auto tv0b = broadcast(tv0, {false, true, false});
+  auto tv1b = broadcast(tv1t, {true, false, false});
```
jacobhinkle, line 2407: Is allocation domain propagated by broadcast at definition? I.e. inside the op?

jacobhinkle: I don't think it is. Would you mind posting the transforms in the scheduled fusion and also the generated cuda code for one of these tests? If the broadcast actually gets allocated with a transposed allocation domain here then we should see that in those printouts hopefully. In any event, does it hurt to manually propagate alloc dom here?

csrc/ir/nodes.cpp
```diff
@@ -2480,8 +2480,11 @@ std::string LoadStoreOp::toInlineString(int indent_size) const {
 
 bool LoadStoreOp::hasInnerTranspose() const {
   if (auto out_tv = dynamic_cast<TensorView*>(out())) {
+    auto use_root_or_alloc = out_tv->hasAllocation()
+        ? out_tv->getAllocationDomain()
+        : out_tv->getRootDomain();
```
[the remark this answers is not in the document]
protonu: I resolve this conversation since I moved the changes out of this function.

csrc/scheduler/matmul.cpp
```diff
@@ -734,6 +734,13 @@ void scheduleSplitKSum(
   splitk_sum->axis(-1)->parallelize(ParallelType::Vectorize);
 }
 
+bool hasInnerTranspose(TensorView* out_tv) {
+  auto use_root_or_alloc = out_tv->hasAllocation()
+      ? out_tv->getAllocationDomain()
+      : out_tv->getRootDomain();
+  return use_root_or_alloc.back() != out_tv->getMaybeRFactorDomain().back();
+}
+
```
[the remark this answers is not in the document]
protonu: Please take a look at the modified implementation.

tests/cpp/test_matmul_scheduler.cpp
protonu: The allocation domain is not propagated across the broadcast. I'll try out manually propagating it.

protonu:
>QQ: is the allocation domain of A not impacting kernel generation? I thought we are looking at their combination which would be 4 cases.

This does run all the 4 cases. I'll update the comment/description of the PR.

protonu:
!build

csrc/scheduler/matmul.cpp
```diff
@@ -734,6 +734,21 @@ void scheduleSplitKSum(
   splitk_sum->axis(-1)->parallelize(ParallelType::Vectorize);
 }
 
+// A LoadMatrixTranspose is needed is there's a rfactor which doesn't
+// match the alloc/root domain. If there's no rfactor, then we
+// need a transpose if the alloc domain and root don't match. When we say
+// match, we refer to the innermost iter domain.
```
jacobhinkle, line 740: I think this assumes the allocation domain of the producer matches its rfactor domain (and `out_tv`s root).

tests/cpp/test_matmul_scheduler.cpp
```diff
@@ -2832,6 +2832,105 @@ TEST_F(MatmulSchedulerTest, DISABLED_RequireExternalPlugin) {
...
+    fusion->addInput(tv0);
+    fusion->addInput(tv1);
+
+    // This has rfactor: {N, K}
+    auto tv1t = transpose(tv1);
+
+    // Propagate the allocation domain of B accross the transpose.
+    if (tv1->hasAllocation()) {
+      tv1t->setAllocationDomain({tv1t->axis(0), tv1t->axis(1)}, true);
+    }
```
jacobhinkle, line 2869: I think the scheduler should actually be the one setting the allocation domain for this tensor. It would set the allocation domain on the register tensor that is the argument to the MmaOp since that is where the TN requirement comes from, then propagate allocation domain back to the register tensor that gets loaded from smem (i.e. the ldmatrix output).

tests/cpp/test_matmul_scheduler.cpp
```diff
@@ -2832,6 +2832,105 @@ TEST_F(MatmulSchedulerTest, DISABLED_RequireExternalPlugin) {
...
+  DisableOptionsGuard option_guard_;
+};
+
+// This tests fusions where inputs to a Matmul will have the root domains
+// [M, K] and [K, N], and all possible combinations of allocation domains.
+// Please note that inpout in B is transposed prior to creating a Mma op.
+TEST_P(AllocationDomainTest, BasicMatmul) {
+  NVFUSER_TEST_CUDA_ARCH_RANGE_GUARD(7, 5, 9, 0);
+  bool a_has_allocation = std::get<0>(GetParam());
+  bool b_has_allocation = std::get<0>(GetParam());
```
jacobhinkle, line 2857:
```suggestion
  bool a_m_inner = std::get<0>(GetParam());
  bool b_n_inner = std::get<0>(GetParam());
```
Just a suggestion to make it more clear that the params are indicating the memory layout of each operand.

csrc/scheduler/matmul.cpp
```diff
@@ -890,33 +904,21 @@ void scheduleMatmul(Fusion* fusion, const MatmulParams& params) {
...
-  } else {
-    bcr = bcw_smem->cacheAfter(LoadStoreOpType::LdMatrix);
   }
 
-  // For Turing and Ampere, the layout of the MmaOp is always TN
-  NVF_ERROR(
-      mma_layout == MmaLayout::TN,
-      "MMAs in Turing and Ampere are TN only, transpose is handled either "
-      "via ldmatrix.trans for fp16 or explicitly for other types.");
-
```
jacobhinkle, line 919: I think we could use a utility `propagateMmaProducerLayouts(MmaOp* mma_op, const std::vector<TensorView*>& to)` which sets the operand allocation domains for the `MmaOp` based on its macro, then propagates that upstream to the given tensors (inclusive). We'd call it like `propagateMmaProducerLayouts(mma_op, {acr, bcr});` I think.

csrc/scheduler/matmul.cpp
```diff
@@ -734,6 +734,16 @@ void scheduleSplitKSum(
   splitk_sum->axis(-1)->parallelize(ParallelType::Vectorize);
 }
 
+bool needsTranposedLoad(TensorView* producer, TensorView* consumer) {
```
jjsjann123, line 737:
```suggestion
bool needsTransposedLoad(TensorView* producer, TensorView* consumer) {
```

csrc/scheduler/matmul.cpp
```diff
@@ -890,33 +899,24 @@ void scheduleMatmul(Fusion* fusion, const MatmulParams& params) {
...
-  if (auto ldst = dynamic_cast<LoadStoreOp*>(bcw_smem->uses().at(0))) {
-    bcr = ldst->out()->as<TensorView>();
-    if (ldst->hasInnerTranspose()) {
-      ldst->setOpType(LoadStoreOpType::LdMatrixTranspose);
+
+  for (auto [tv_smem, tv_r] :
+       {std::make_pair(acw_smem, &acr), std::make_pair(bcw_smem, &bcr)}) {
+    auto producer = tv_smem;
+    auto consumer = tv_smem->uses().at(0)->output(0)->as<TensorView>();
+    auto toTranspose = needsTranposedLoad(producer, consumer);
```
jjsjann123, line 907:
```suggestion
    auto toTranspose = needsTransposedLoad(producer, consumer);
```

csrc/scheduler/matmul.cpp
```diff
@@ -734,6 +734,16 @@ void scheduleSplitKSum(
   splitk_sum->axis(-1)->parallelize(ParallelType::Vectorize);
 }
 
+bool needsTranposedLoad(TensorView* producer, TensorView* consumer) {
+  const auto map =
+      PairwiseRootDomainMap(producer, consumer).mapProducerToConsumer();
+  auto maybeProducerAlloc = producer->getMaybeAllocationDomain();
+  auto maybeConsumerRFactor = consumer->getMaybeRFactorDomain();
```
jjsjann123, line 741: nitpick: let's assert on `std::is_permutation`. Otherwise we can falsely claim needsTransposedLoad.

csrc/scheduler/matmul.cpp
```diff
@@ -890,33 +899,24 @@ void scheduleMatmul(Fusion* fusion, const MatmulParams& params) {
...
+  for (auto [tv_smem, tv_r] :
+       {std::make_pair(acw_smem, &acr), std::make_pair(bcw_smem, &bcr)}) {
+    auto producer = tv_smem;
+    auto consumer = tv_smem->uses().at(0)->output(0)->as<TensorView>();
+    auto toTranspose = needsTranposedLoad(producer, consumer);
+    if (auto ldst = dynamic_cast<LoadStoreOp*>(tv_smem->uses().at(0))) {
+      *tv_r = ldst->out()->as<TensorView>();
+      ldst->setOpType(
+          toTranspose ? LoadStoreOpType::LdMatrixTranspose
+                      : LoadStoreOpType::LdMatrix);
```
jjsjann123, line 912: Out of curiosity. What's planned to change for the follow up PR when you say?

> In a follow-up PR I can handle LdMatrix/LdMatrixTrans.

csrc/scheduler/mma_utils.h
```diff
@@ -57,11 +57,11 @@ void makeTile(TensorView* tv, std::vector<int64_t> tile_sizes);
 
 //! Order the inner tile dimensions as the original order in
 //!  root domain. Also putting broadcast domains on the left.
```
jjsjann123, line 59: `in the original order in allocation domain` instead?

csrc/scheduler/mma_utils.cpp
```diff
@@ -466,18 +467,18 @@ void orderTiledConcreteIdAsRoot(TensorView* tv) {
       leftmost_pos = i;
       continue;
     }
-    auto maybe_root =
-        getMaybeRootIfInnermostTiled(leaf_id, maybe_rfactor_id_set);
+    auto maybe_alloc_domain =
+        getMaybeAllocationIfInnermostTiled(leaf_id, id_set);
```
jjsjann123, line 471: nitpick: I think the logic here assumes rfactor and allocation are just permutation? Should we add an assert somewhere?

jjsjann123: So I guess if we are doing that somewhere here (or maybe it's already done in an even earlier spot, please ignore the comment on the other place where I asked to assert the permutation.

jjsjann123 (commented):
LGTM. I'll let @jacobhinkle stamp.

csrc/scheduler/matmul.cpp
jacobhinkle: This is referring to us inferring the ptx instruction based on allocation domain. I _think_ we need to look at consumer alloc to consumer root and compose that with producer rfactor to producer alloc. The resulting composition can tell us whether a transpose is needed as it combines both the allocation information and the semantic information in consumer's root to rfactor domain.

jjsjann123: Thanks for filling in that~~ :bow:

protonu:
Thanks for the reviews, but given our design has evolved. I'll create a new PR.

PR #2227, opened 2024-05-09
Kill `MmaOp::layout`

zasdfgbnm:
Besides, using `enum class UnitDim { K, M_or_N };` for representing what is the inner dim is a less confusing term.

### DIFF
--- a/csrc/device_lower/pass/index.cpp
+++ b/csrc/device_lower/pass/index.cpp
@@ -1578,6 +1578,7 @@ void IndexLowering::handle(const MmaOp* mma) {
   constexpr int64_t core_matrix_outer_size = 8;
   Val* a = nullptr;
   Val* b = nullptr;
+  auto layout = lower_utils::getMmaLayout(mma);
   if (mma->inA()->as<TensorView>()->getMemoryType() == MemoryType::Shared) {
     // TODO: This is a temporary solution and only supports a single tile in
     // smem.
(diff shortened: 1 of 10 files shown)


zasdfgbnm:
!build

zasdfgbnm:
!build

zasdfgbnm:
!build

zasdfgbnm:
cc: @drzejan2

csrc/device_lower/utils.h
```diff
@@ -319,6 +319,8 @@ std::vector<Val*> getFusionOutputsRequiringCodegen(Fusion* fusion);
 //! fusion has blockDim.x = 128, this function will return 3 instead of 128.
 Val* getNumThreadsInTensorView(TensorView* tv);
 
+std::array<UnitDim, 2> getMmaLayout(const MmaOp* expr);
```
protonu, line 322: Would it be possible to add a small comment here

csrc/device_lower/pass/index.cpp
```diff
@@ -1578,6 +1578,7 @@ void IndexLowering::handle(const MmaOp* mma) {
   constexpr int64_t core_matrix_outer_size = 8;
   Val* a = nullptr;
   Val* b = nullptr;
+  auto layout = lower_utils::getMmaLayout(mma);
```
jacobhinkle, line 1581: This might be more clear and removes one additional variable named `layout` :-D
```suggestion
  const auto& [unitdim_a, unitdim_b] = lower_utils::getMmaLayout(mma);
```

csrc/mma_type.h
```diff
@@ -203,6 +203,7 @@ constexpr MmaMacroEncode::MmaMacroEncode(MmaMacro macro)
 //! TN : M,K X N,K -> M,N
 //! NN : K,M X N,K -> M,N
 enum class MmaLayout { NT = 0, TT, TN, NN };
+enum class UnitDim { K, M_or_N };
```
jacobhinkle, line 206:
```suggestion

//! Indicates which dimension is innermost in the allocation domain of an operand
enum class UnitDim { K, M_or_N };
```

jacobhinkle (approved):
Do we need to do any validation for Ampere macros that K is the unit dim for each operand?

protonu (approved):

csrc/device_lower/pass/index.cpp
zasdfgbnm: fixed

csrc/device_lower/utils.h
zasdfgbnm: added

zasdfgbnm:
!build

drzejan2:
@zasdfgbnm - thanks for preparing this change!

PR #2229, opened 2024-05-09
Add test that external matmul plugin is loaded properly

jacobhinkle:
This adds a C++ test that checks that an external matmul heuristic plugin is loaded. This is useful in automated jobs where we build a matmul heuristic plugin and run some benchmarks. Without this test it is possible to improperly prepare the environment and wind up running benchmarks with the default core heuristics. Instead we can first run this test and fail the job early if it does not succeed. The test is disabled by default so that it does not cause default builds to fail. To run it use a command sequence like this:
```
export NVFUSER_MATMUL_HEURISTIC_PLUGIN=/path/to/plugin.so
build/test_matmul --gtest_also_run_disabled_tests --gtest_filter='MatmulSchedulerTest.DISABLED_RequireExternalPlugin`
```

### DIFF
--- a/tests/cpp/test_matmul_scheduler.cpp
+++ b/tests/cpp/test_matmul_scheduler.cpp
@@ -2376,6 +2376,23 @@ TEST_F(MatmulSchedulerPluginTest, BasicMatmul) {
       executor_cache.fusion(), outputs, {t0, t1}, {tref}, __LINE__, __FILE__);
 }
 
+// This test can be used to check that an external plugin has been loaded. It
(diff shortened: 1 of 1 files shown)


jacobhinkle:
!build

jacobhinkle:
Building to confirm that the test is disabled by default.

xwang233 (approved):

jacobhinkle:
Test failures are unrelated; `test_matmul` passed despite not having a heuristic plugin. Merging.

PR #2230, opened 2024-05-10
shmoo test of iter grouped grid reduction

liqiangxl:
Issue:
The current Shmoo test of iter grouped grid reduction relies on input sizes, its coverage of heuristic parameters depends on how the heuristics is generated.

Fix:
This PR introduces another Shmoo test that spans 7 heuristic parameters improving the test coverage.


naoyam (approved):
LGTM. Thanks for adding this. Out of curiosity, how long does it take to run this test?

liqiangxl:
> LGTM. Thanks for adding this. Out of curiosity, how long does it take to run this test?

about 1 min on H100, I tried different number of cases to keep it around 1 min.

PR #2234, opened 2024-05-13
Improve perf of `isResharding(Expr*)`

samnordmann:
Reimplement `isResharding(Expr*)` with a pairwise map for better perf. Before this patch, it would use `getTvsWithDifferentSharding` which creates a `computeAtMap`, which is too costly. This patch should address https://github.com/Lightning-AI/lightning-thunder/issues/229


csrc/multidevice/utils.cpp
```diff
@@ -142,19 +142,49 @@ int64_t numDeviceDims(TensorView* tv) {
...
-  for (auto tv : ir_utils::filterByType<TensorView>(expr->outputs())) {
-    tvs.insert(tv);
-  }
-  if (tvs.empty()) {
-    return false;
+  // we don't use getTvsWithDifferentSharding because it creates a computeAtMap,
+  // which is too costly
+  for (auto producer : ir_utils::filterByType<TensorView>(expr->inputs())) {
+    for (auto consumer : ir_utils::filterByType<TensorView>(expr->outputs())) {
+      // exit early in the unsharded case for performance
```
wujingyue, line 149: The loop body has a mix of `continue` and `return true`, which is a bit hard to follow. I'd create a helper function `isResharding(input_tv, output_tv)` and make the loop here simply 

```
for each input_tv {
  for each output_tv {
    if isResharding(input_tv, output_tv)
      return true;
  }
}
return false;
```

csrc/multidevice/utils.cpp
```diff
@@ -142,19 +142,49 @@ int64_t numDeviceDims(TensorView* tv) {
...
+      if (producer->getDeviceMesh().vector() !=
+          consumer->getDeviceMesh().vector()) {
+        return true;
+      }
+
+      // Create a map between producer's and consumer's IterDomains. We iterate
+      // over producer's iterdomain and compare sharding type with consumer's
+      // iterdomain
+      const auto p2c_map =
+          PairwiseRootDomainMap(producer, consumer).mapProducerToConsumer();
```
wujingyue, line 163: This is not equivalent to what's in getTvsWithDifferentSharding. The latter looks at the leaf domain instead of the root domain. While I understand it's probably OK at this moment (given we've been splitting rfactor for DID), I think we'll soon have to fix the discrepancy? Wdyt, @cowanmeg

csrc/multidevice/utils.cpp
```diff
@@ -142,19 +142,49 @@ int64_t numDeviceDims(TensorView* tv) {
...
+            (p_id->isDeviceDim() || c_id->isDeviceDim())) {
+          // Mismatch found
+          return true;
+        }
+      }
+    }
   }
-  auto tv_ref = *tvs.begin();
-  tvs.erase(tv_ref);
-  return !getTvsWithDifferentSharding(tv_ref, tvs).empty();
```
wujingyue, line 157: getTvsWithDifferentSharding has several other uses than isResharding. What about fixing getTvsWithDifferentSharding instead?

samnordmann: `getTvsWithDifferentSharding` inputs tv are meant to be very general, with no a priori assumption on producer/consumer dependency -- on the contrary, in `isResharding` we have the extra info that the tvs we are considering are I/O of the expr.

While it's good to have a general helper function like `getTvsWithDifferentSharding`, it also makes sense to have an optimized implementation for special cases like the one of `isResharding`

So, I agree with you, fixing `getTvsWithDifferentSharding` would be more general, but I am not sure it is possible. Let me know if you think there is a way

csrc/multidevice/utils.cpp
samnordmann: Correct, however I couldn't find any `PairwiseLeafDomainMap`

csrc/multidevice/utils.cpp
samnordmann: The problem I see with this suggestion is that the auxialiary function `isResharding(input_tv, output_tv)` needs to know that I/O are consumer/producer, but there is no way to assert that. Should I write such an auxialiary function relying on this implicit assumption?

wujingyue: > there is no easy way to assert that

Something like https://github.com/NVIDIA/Fuser/blob/b020415858caa38a77b3c309566903f3fbdbac3d/csrc/root_domain_map.cpp#L61-L66, which is already asserted by PairwiseRootDomainMap?

csrc/multidevice/utils.cpp
wujingyue: > however I couldn't find any PairwiseLeafDomainMap

Correct -- there's no PariwiseLeafDomainMap. Unfortunately, we probably have to implement it in C++ for loops and branches. I believe this will be needed when we want to split leaf instead for DID.

csrc/multidevice/utils.cpp
wujingyue: Interesting. All the non-test uses are in https://github.com/NVIDIA/Fuser/blob/b020415858caa38a77b3c309566903f3fbdbac3d/csrc/multidevice/utils.cpp and they all reason about producer/consumer. While there probably is a value of keeping a general helper like `getTvsWithDifferentSharding` (at least the general form seems to be useful for https://github.com/NVIDIA/Fuser/blob/b020415858caa38a77b3c309566903f3fbdbac3d/tests/cpp/test_resharding.cpp#L284), other uses in multidevice/utils.cpp are still worth fixing. One possible way is to create a helper like `isResharding(input_tv, output_tv)` and call that instead.

wujingyue (commented):
(Made some inline comments)

csrc/multidevice/utils.cpp
cowanmeg: We have a similar problem in `getShardingChanges` since we analyze the root domain for the convenience of PairwiseRootDomainMap.
 
Since this change will help lightning perf, I am good with analyzing the root domain for now and we will fix it later in the leaf+DID PRs!

csrc/multidevice/utils.cpp
cowanmeg: We could check if `input_tv` and `output_tv` are producer consumer, and use the quick `PairwiseRootDomainMap` path if true, otherwise use the expensive `ComputeAt` map?

csrc/multidevice/utils.cpp
```diff
@@ -142,19 +142,49 @@ int64_t numDeviceDims(TensorView* tv) {
...
+      // over producer's iterdomain and compare sharding type with consumer's
+      // iterdomain
+      const auto p2c_map =
+          PairwiseRootDomainMap(producer, consumer).mapProducerToConsumer();
+      for (auto p_id :
+           TensorDomain::noReductions(producer->getMaybeRFactorDomain())) {
+        auto p2c_map_it = p2c_map.find(p_id);
+        if (p2c_map_it == p2c_map.end()) {
+          NVF_ERROR(
+              !p_id->isDeviceDim(),
```
cowanmeg, line 169: Isn't this always an error regardless if p_id is device parallelized?

csrc/multidevice/utils.cpp
naoyam: A couple of suggestions:

- Since the problem is `isResharding` is called multiple times at the time of segmentation, it should be possible to do the analysis one and save it for reuse. 
- `ComputeAtMap` and `IdModel` are overkill for this case (although I think we already build one, which may be reusable here). `PairwiseRootDomainMap` and `BestEffortReplay` would be a cheaper alternative. 
- While likely not necessary for multi-device cases yet, this is not a trivial dependency analysis in general, especially with broadcast domains. See, for example, `useSameIndex` in sync_information.cpp. IdModel should greatly simplify the analysis.

wujingyue: > Since this change will help lightning perf, I am good with analyzing the root domain for now and we will fix it later in the leaf+DID PRs!

SGTM! 

For the later fix, I'm OK with either changing isResharding to map leaf domains or caching ComputeAtMap/IdModel. I slightly prefer the former because it's stateless. However, if the implementation gets tricky and duplicates what's already in ComputeAtMap/IdModel, I'd try to cache the analysis result as @naoyam suggested, which seems feasible AFAICS from the users of isResharding.

csrc/multidevice/utils.cpp
samnordmann: Right, nice! I did as you suggested, I introduced a new function `bool haveDifferentShardings(TensorView* producer, TensorView* consumer)` in `multidevice/utils.(h|cpp)` and now `isResharding(Expr*)` calls this function by looping over the expr's I/O

csrc/multidevice/utils.cpp
samnordmann: I introduced a new function `bool haveDifferentShardings(TensorView* producer, TensorView* consumer)` in `multidevice/utils.(h|cpp)` that I use instead of calling `getTvsWithDifferentSharding` everywhere in the source code, so we should be good.

I only let the test in `test_resharding.cpp` calling `getTvsWithDifferentSharding`

> We could check if input_tv and output_tv are producer consumer, and use the quick PairwiseRootDomainMap path if true, otherwise use the expensive ComputeAt map?

I did not implement that for now, we can do it later if we feel the need.

Wdyt ?

csrc/multidevice/utils.cpp
samnordmann: I don't know in general but for the case we test you are right! So I made the change accordingly by asserting that all the IDs are mapped to the consumer

samnordmann:
!build --dist

cowanmeg (approved):

csrc/multidevice/utils.cpp
```diff
@@ -141,20 +141,53 @@ int64_t numDeviceDims(TensorView* tv) {
...
+        p2c_map_it != p2c_map.end(),
+        "the producer ",
+        producer,
+        " has a dimension ",
+        p_id,
+        " that is not mapped to its consumer ",
+        consumer);
+    auto c_id = p2c_map_it->second;
+    if (p_id->getParallelType() != c_id->getParallelType() &&
+        (p_id->isDeviceDim() || c_id->isDeviceDim())) {
```
wujingyue, line 171:
```suggestion
    if (p_id->isDeviceDim != c_id->isDeviceDim) {
```

I think this is equivalent and simpler.

wujingyue (approved):
Thank you!

wujingyue:
!build --dist

csrc/multidevice/utils.cpp
samnordmann: I'd rather keep it the way it was so we won't have to change it when we soon introduce the new `DIDy` parallel type

PR #2236, opened 2024-05-13
Translate MatmulOp and LinearOp

jacobhinkle:
The purpose of this PR is to enable the NVFuser matmul scheduler to operate on the new `MatmulOp` an `LinearOp` nodes. That means the matmul scheduler can optionally schedule segments if they are not supported by ATen's `matmul` and `linear` functions.

Specifically, this PR:

1. Adds `MatmulOp` and `LinearOp` to the set of `MatmulPattern`s detected. Previously only `MmaOp` and combined mul-sum patterns were detected.
4. Enable translation of `MatmulOp` and `LinearOp` to fixed broadcast+MmaOp patterns.
5. Introduces `EnableOption::FuseMatmul` and associated `NVFUSER_ENABLE=fuse_matmul` option to enable the automatic scheduler to accept matmul patterns consisting of a `MatmulOp` or `LinearOp`. By default, the matmul scheduler will not accept segments containing `MatmulOp` or `LinearOp` patterns, meaning all those nodes will be computed using the `ExprEval` scheduler (ATen).

Not all cases that are supported by these IR nodes can be translated to an `MmaOp`. In particular:
1. gemv cases where one operand is 1D are not supported
2. Cases with multiple batch dimensions.
3. Cases where "batch" dimensions must be added by unsqueezing one dimension. Those new dimensions are indistinguishable from e.g. multiple M dimensions which do not yet support.

These correspond to test cases in the two new tests. See the `TODO` comments for descriptions of cases we plan to support but cannot yet translate.

There is also a commented out test case for `LinearOp`: M=N=1 which should not be translated to a `LinearOp` at all.


jacobhinkle:
This test is failing: https://github.com/NVIDIA/Fuser/blob/5b5ec8fc9601fe51d41a09a7524f20e2a6af19fa/tests/cpp/test_combine_mul_sum.cpp#L134-L146

I haven't implemented this condition because I don't think we really want it. Rather, we would prefer to accept patterns with multiple M, N, K, or Batch dims and simply canonicalize them via reorder/merge at the beginning of scheduling. Still, if we do support that then we should update the test to actually check that they are properly scheduled; until then we should probably keep the check strict.

tests/cpp/test_combine_mul_sum.cpp
```diff
@@ -269,6 +269,240 @@ TEST_F(CombineMulSumAsMmaTest, UseMatmulScheduler) {
...
+  run(4, 2, -1, /*transpose_a_alloc=*/false, /*expect_aten_eval=*/false);
+
+  // Bias cases
+  run(2, 2, 0, /*transpose_a_alloc=*/false, /*expect_aten_eval=*/false);
+  run(2, 2, 1, /*transpose_a_alloc=*/false, /*expect_aten_eval=*/false);
+  // Undocumented 2D bias support
+  run(2, 2, 2, /*transpose_a_alloc=*/false, /*expect_aten_eval=*/false);
+
+  run(3, 2, 1, /*transpose_a_alloc=*/false, /*expect_aten_eval=*/false);
+  run(4, 2, 1, /*transpose_a_alloc=*/false, /*expect_aten_eval=*/false);
```
jacobhinkle, line 503: These cases are currently failing with 
> C++ exception with description "kernel->inputs().size() <= args.size() INTERNAL ASSERT FAILED at "/opt/pytorch/nvfuser/csrc/executor_utils.cpp":700, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. KernelArgumentHolder contains less argument than kernel's input.
Exception raised from bindInputs at /opt/pytorch/nvfuser/csrc/executor_utils.cpp:700

jacobhinkle:
!build

tests/cpp/test_combine_mul_sum.cpp
```diff
@@ -269,6 +271,341 @@ TEST_F(CombineMulSumAsMmaTest, UseMatmulScheduler) {
...
+  // TODO: We don't yet support multiple batch dims in matmul scheduler
+  run(4, 2, -1, /*transpose_a_alloc=*/false, /*expect_aten_eval=*/true);
+
+  // Bias cases
+  run(2, 2, 0, /*transpose_a_alloc=*/false, /*expect_aten_eval=*/false);
+  run(2, 2, 1, /*transpose_a_alloc=*/false, /*expect_aten_eval=*/false);
+  // Undocumented 2D bias support
+  // TODO: Currently failing in propagateBoundValuesThroughExactMaps, indicating
+  // possible PairwiseRootDomainMap issue for 2D bias?
+  // run(2, 2, 2, /*transpose_a_alloc=*/false, /*expect_aten_eval=*/false);
```
jacobhinkle, line 601: I'm currently investigating this issue.

Priya2698: We do not support 2D bias. Thunder does not allow 2D bias either.

https://github.com/NVIDIA/Fuser/pull/2240#discussion_r1603803176

Priya2698: We will need to update the mapping if we want to support 2D bias, but that case is tricky, since for this case as well, we have 2D bias but the mapping is not to [M, N]: `A = [2, 4, 6], B [3, 6], Bias = [2*4, 3] `

jacobhinkle: OK I remember that now. In that case, with your permission I'll just add a check to our `linear` op so that we fail earlier, and I'll remove this test case.

jacobhinkle:
!build

csrc/scheduler/matmul_utils.cpp
```diff
@@ -236,12 +236,22 @@ std::string isMatmulFusionDefinitionSupported(
...
     if (entry != roles_map.end()) {
       if (MATMUL_CORE_ROLES_EXPECTED_COUNT == entry->second.size()) {
         tvs_with_roles.insert(entry->second.begin(), entry->second.end());
+        for (TensorView* tv : entry->second) {
+          if (operand_dim == -1) {
+            operand_dim = tv->nDims();
+          } else if (tv->nDims() != operand_dim) {
+            return "All A operands must have the same dimension.";
+          }
+        }
```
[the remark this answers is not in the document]
jacobhinkle: Changed to loop over `{MatmulRole::INPUT_A, MatmulRole::INPUT_B}`.

csrc/scheduler/matmul_utils.cpp
```diff
@@ -481,13 +509,33 @@ std::string getMatmulCompileTimeRejectReason(Fusion* fusion) {
...
+      if (op->isA<MatmulOp>() || op->isA<LinearOp>()) {
+        if (!isOptionEnabled(EnableOption::FuseMatmul)) {
+          // Check for MatmulOp or LinearOp. If found, then only fuse if option
+          // is specified
+          return "MatmulOp and LinearOp fusion is disabled by default. "
+                 "Enable it using NVFUSER_ENABLE=fuse_matmul";
+        }
+        // Refuse patterns containing 1D inputs since these are mat-vec as
+        // opposed to mat-mat products.
+        if (pattern.A->nDims() < 2 || pattern.B->nDims() < 2) {
```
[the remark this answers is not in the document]
jacobhinkle: These check slightly different things so I am a little worried about separating them. I'd like to relax the constraint in `isMatmulFusionDefinitionSupported` to include cases where the operands might have different dims that get broadcasted to and used in the matmul op. However, the check here is about disallowing inner/outer products and mat-vec versions; I think this check will stay for the foreseeable future (unless we start handling mat-vec).

csrc/scheduler/matmul_utils.cpp
```diff
@@ -475,13 +503,33 @@ std::string getMatmulCompileTimeRejectReason(Fusion* fusion) {
...
+          return "Cannot fuse matrix-vector products";
+        }
```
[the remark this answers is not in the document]
jacobhinkle: Added guard for dtype, with test

csrc/scheduler/mma_utils.cpp
```diff
@@ -1530,6 +1627,72 @@ std::unordered_map<ValGroup, MatmulDomain> MatmulPattern::getDimRoles(
...
+      NVF_ERROR(
+          false, "MatmulOp node should not be created when both inputs are 1D");
+    } else if (A->nDims() == 1) {
+      // Missing M dimension
+      dim_roles[exact_graph.toGroup(B->axis(-1))] = MatmulDomain::N;
+      m_and_k_dims = 1;
+    } else if (B->nDims() == 1) {
+      // Missing N dimension
+      dim_roles[exact_graph.toGroup(A->axis(-2))] = MatmulDomain::M;
+      m_and_k_dims = 1;
```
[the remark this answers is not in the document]
jacobhinkle: That's a nice idea. In trying to do this I actually found a logic error in the current approach:

The vector returned by `mapMatmulOpIterDomains` contains `nullptr` when that operand has no dimension to match the output dim. However, that is not quite enough information to determine the role without duplicating the logic above. For example:
```
A [ iS0{b} iS1{M} iS2{K} ]
B [ iS3{K} iS4{N} ]
output [ iS5{b} iS6{M} iS7{N} ]
```
In this case when we get these mappings:
```
mapping_a = [ iS0{b} iS1{M} nullptr ]
mapping_b = [ nullptr nullptr iS4{N} ]
```
In the current approach, `iS5{b}` will be treated as a batch dimension. However, since it is actually being broadcasted in the `B` operand, it is actually another M dimension. Anyway, I think using this approach is nice and simple, and even better, correct.

jacobhinkle:
!build

csrc/scheduler/matmul_utils.cpp
```diff
@@ -236,31 +236,42 @@ std::string isMatmulFusionDefinitionSupported(
...
-    if (entry != roles_map.end()) {
-      if (MATMUL_CORE_ROLES_EXPECTED_COUNT == entry->second.size()) {
-        tvs_with_roles.insert(entry->second.begin(), entry->second.end());
+    for (MatmulRole role : {MatmulRole::INPUT_A, MatmulRole::INPUT_B}) {
+      auto entry = roles_map.find(role);
+      if (entry != roles_map.end()) {
+        if (MATMUL_CORE_ROLES_EXPECTED_COUNT == entry->second.size()) {
+          tvs_with_roles.insert(entry->second.begin(), entry->second.end());
+          for (TensorView* tv : entry->second) {
+            if (operand_dim == -1) {
```
Priya2698, line 252: Just to confirm, we do not support 3D inputs in `LinearOp` then in this PR?

csrc/scheduler/mma_utils.cpp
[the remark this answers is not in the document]
Priya2698: I see. My suggestion was that the role information is embedded in the position of the iterdomain in the mapping output. For eg: `out_size-3, out_size-2, out_size-1` are M, N, K respectively.

Could you clarify which approach is erroneous -- `mappingMatmulOpIterDomain` or this PR? Why should `iS5` be M instead of batch?

tests/cpp/test_combine_mul_sum.cpp
```diff
@@ -269,6 +271,344 @@ TEST_F(CombineMulSumAsMmaTest, UseMatmulScheduler) {
...
+  const int64_t A_dim = std::get<0>(GetParam());
+  const int64_t B_dim = std::get<1>(GetParam());
+  const bool enable_fusion = std::get<2>(GetParam());
+  const bool transpose_a_alloc = std::get<3>(GetParam());
+  const bool expect_segmented = std::get<4>(GetParam());
+  const ScheduleHeuristic expected_heuristic = std::get<5>(GetParam());
+
+  // CombineMulSumAsMmaTest disabled MatmulExprEval, but we need it enabled
+  DisableOptionsGuard dog;
+  DisableOptionsGuard::getCurOptions().unset(DisableOption::MatmulExprEval);
```
Priya2698, line 291: PR #2268 removed this disable option once we separated these schedulers. Both schedulers are on by default.

tests/cpp/test_combine_mul_sum.cpp
```diff
@@ -269,6 +271,344 @@ TEST_F(CombineMulSumAsMmaTest, UseMatmulScheduler) {
...
+
+  EnableOptionsGuard eog;
+  if (enable_fusion) {
+    EnableOptionsGuard::getCurOptions().set(EnableOption::FuseMatmul);
+  } else {
+    EnableOptionsGuard::getCurOptions().unset(EnableOption::FuseMatmul);
```
Priya2698, line 297: Do we need to unset it? The options guard should reset the value after each test, right?

tests/cpp/test_combine_mul_sum.cpp
```diff
@@ -269,6 +271,344 @@ TEST_F(CombineMulSumAsMmaTest, UseMatmulScheduler) {
...
+  const int64_t B_dim = std::get<1>(GetParam());
+  const int64_t bias_dim = std::get<2>(GetParam());
+  const bool enable_fusion = std::get<3>(GetParam());
+  const bool transpose_a_alloc = std::get<4>(GetParam());
+  const bool expect_aten_eval = std::get<5>(GetParam());
+
+  // CombineMulSumAsMmaTest disabled MatmulExprEval, but we need it
+  // enabled
+  DisableOptionsGuard dog;
+  DisableOptionsGuard::getCurOptions().unset(DisableOption::MatmulExprEval);
```
Priya2698, line 453: Same as above, we don't need to unset this anymore.

tests/cpp/test_combine_mul_sum.cpp
```diff
@@ -269,6 +271,344 @@ TEST_F(CombineMulSumAsMmaTest, UseMatmulScheduler) {
...
+        // Enable fusion
+        std::make_tuple(2l, 2l, -1l, true, false, false),
+        // We cannot yet handle allocation domain in matmul scheduler
+        std::make_tuple(2l, 2l, -1l, true, true, true),
+        // We don't fuse 1D inputs
+        std::make_tuple(1l, 2l, -1l, true, false, true),
+        std::make_tuple(2l, 1l, -1l, true, false, true),
+        // TODO: The following currently fails but it should not be translated
+        // to LinearOp to begin with
+        // std::make_tuple(1l, 1l, -1l, true, false, false),
```
Priya2698, line 580: Why does this fail? Based on our offline conversation we decided to pipe everything through ATen. I do need to remove the special cases from MatmulOp.

tests/cpp/test_matmul_scheduler.cpp
```diff
@@ -2685,6 +2685,50 @@ TEST_F(NVFuserTest, SegmentLinearOpPrologue) {
   testValidate(executor_cache.fusion(), outputs, {t0, t1}, __LINE__, __FILE__);
 }
 
+// Test that the matmul scheduler refuses to translate a matmul that is not
+// Half or BFloat16
+TEST_F(NVFuserTest, SegmentMatmulOpUnsupportedDtype) {
```
Priya2698, line 2690: nit: `NVFuserTest`-> `MatmulSchedulerTest`. Same for the two tests above. It helps with filtering.

tests/cpp/test_matmul_scheduler.cpp
```diff
@@ -2699,6 +2743,66 @@ TEST_F(MatmulSchedulerTest, DISABLED_RequireExternalPlugin) {
   MatmulParams params;
 }
 
+// Test that we can segment a Fusion containing two matmuls
+TEST_F(MatmulSchedulerTest, Llama2FFN) {
+  NVFUSER_TEST_CUDA_ARCH_RANGE_GUARD(7, 5, 9, 0);
+
+  for (bool enable_fusion : {false, true}) {
```
Priya2698, line 2750: I would recommend parametrizing this, the parametrization pattern can be used with the `SegmentMatmulOpPrologue` and `SegmentLinearOpPrologue` tests as well to test with all three combinations of schedulers as listed in this comment --
```

// TODO: Once we can control the ExprEval and Matmul schedulers via options, run
// this test with all three combinations (with and without each scheduler, but
// at least one enabled).

File: tests/cpp/test_matmul_scheduler.cpp [hunk 14]
```diff
@@ -2699,6 +2743,66 @@ TEST_F(MatmulSchedulerTest, DISABLED_RequireExternalPlugin) {
...
+    auto t0 = at::randn({M, K}, options);
+    auto t1 = at::randn({K, N}, options);
+    auto t2 = at::randn({K, N}, options);
+    std::vector<c10::IValue> inputs{t0, t1, t2};
+
+    EnableOptionsGuard eog;
+    if (enable_fusion) {
+      EnableOptionsGuard::getCurOptions().set(EnableOption::FuseMatmul);
+    } else {
+      EnableOptionsGuard::getCurOptions().unset(EnableOption::FuseMatmul);
```
Priya2698, line 2786: We shouldn't have to unset this when using the options guard.

csrc/scheduler/matmul_utils.cpp
```diff
@@ -253,6 +263,21 @@ std::string isMatmulFusionDefinitionSupported(
...
+            operand_dim = tv->nDims();
+          } else if (tv->nDims() != operand_dim) {
+            // We cannot always handle differently sized inputs, such as those
+            // we encounter when translating MatmulOp and LinearOp. This is
+            // because in those cases one of the operands will have new
+            // Broadcast dimensions where the other operand has Iteration
+            // batch dimensions, meaning these new dims are actually M or N
+            // dimensions. Multiple M and N dimension support is planned but for
+            // now we must reject these patterns before attempting to translate
+            // them.
```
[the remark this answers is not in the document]
Priya2698: Do we intend to support different sizes of A/B. The checks reject such fusions, but the translation itself makes no such assumption.

tests/cpp/test_matmul_scheduler.cpp
jacobhinkle: This is just there in case it was set as an env var.

tests/cpp/test_matmul_scheduler.cpp
jacobhinkle: Oh that's a good idea

csrc/scheduler/matmul_utils.cpp
jacobhinkle: That's correct. A LinearOp with 3D input [B M K] and 2D weights [N K] would be expanded like input [B M 1 K] weight [1 1 N K] so you can see that the "batch" dimension in this case is really another M dimension, which we do not support (until #2303).

tests/cpp/test_combine_mul_sum.cpp
jacobhinkle: Yes it's on by default so this usually has no effect; it's here just ot ensure that behavior. Just like with the `unset(EnableOption::FuseMatmul)` this was just to help my sanity when running stuff with env vars set.

tests/cpp/test_combine_mul_sum.cpp
jacobhinkle: See other responses. You're correct but this makes it explicit and also overrides env vars that might be present.

tests/cpp/test_combine_mul_sum.cpp
jacobhinkle: The NoOp scheduler accepts the whole fusion, which looks like this:
```
Fusion IR after pre-segmenter optimization passes:
Inputs:
  T0_g[ iS0{i0} ], __half
  T1_g[ iS3{i0} ], __halfIt should be easy to fix the NoOp scheduler to avoid accepting zero-dimensional non-trivial fusions.
Outputs:
  T5_g[ ], __half

%kernel_math {
T2_l[ rS2{i0} ]
   = linear(T0_g[ iS0{i0} ],
            T1_g[ iS3{i0} ]  )
T3_l[ ]
   = __half2float(T2_l[ rS2{i0} ]);
T4_l[ ]
   = sinf(T3_l[ ]);
T5_g[ ]
   = __float2half(T4_l[ ]);
}
```
I just updated the NoOp scheduler to avoid fusions with matmul ops, which fixes this issue. I now check cases like this with and without fusion enabled.

csrc/scheduler/mma_utils.cpp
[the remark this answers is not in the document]
jacobhinkle: As for why `iS5` should not be considered a batch dim. For a generic matmul problem, we have two operands and one output and we define the dimension roles as follows:
- The output has one reduction dimension which we call K. It must be non-broadcast in both operands (or else this is not a matmul).
- There's a dimension that is broadcast in the first operand and not the second. It appears in the output (the broadcast is concretized). This is what we call the M dimension.
- Likewise there's a dimension that is broadcast in the second operand but not the first. This is the N dimension.
- There may be dimensions that are Iteration in both operands as well as in the output. This is a batch dimension.

I think the confusion is coming from the difference between a batch matrix multiply, in which we are given a batch of _independent_ matmul problems, versus batched DL training in which we have a batch of data that forms a matrix and we want to comput a batch of mat-vec problems as a single matmul. In the former case the "batch" dimension is really the M dimension of the first argument (input) while the matrix being multiplied against each row of the input is the same (the weights) and it's not batched.

tests/cpp/test_matmul_scheduler.cpp
jacobhinkle: I added a parametrization for the Llamma2FFN test, but the prologue tests are actually currently failing when fusion is enabled, so I will leave that for a separate PR (that might be fixed by #2309).

jacobhinkle:
!build

jacobhinkle:
!build

Priya2698 (approved):
LGTM. Thanks for addressing the comments.

PR #2232, opened 2024-05-10
IdModel: loop promotion cleanup

naoyam:
No logic change. Mostly mechanical cleanup.

Replaced the test-specific IdModel subclass with a callback interface. The callback interface allows to save all necessary temporary results for validation. No more duplication of `buildLoopPromotionMap`. (Related comment: https://github.com/NVIDIA/Fuser/pull/2220#discussion_r1596296138)

To introduce the callback interface, moved the loop promotion part out of `IdModel` to its own builder class.

PR #2239, opened 2024-05-14
Map sibling domains in the LOOP graph

naoyam:
This was a small oversight. Since sibling tensors always share the loops, their iter domains should always be mapped in the loop graph, no matter how they are inlined. Added a simple repro and a fix.

(Stacked on top of #2232)


naoyam:
!build

naoyam:
!build

zasdfgbnm (approved):

Issue #2132, opened 2024-04-23
Support a 0-dimensional Bias argument in Linear.


protonu:
An initial version of the Linear Python API has been checked in here:
https://github.com/NVIDIA/Fuser/pull/2090

PyTorch supports a 0-dim bias and we should too. 
This work will probably have 2 parts:

1. Ensuring our C++ code can handle a 0-D bias. The segmenter seems to be missing some support here.
2. Extending the Python test to support the 0-dim bias.

protonu:
0-Dimensional Bias argument for Linear (ie broadcast a scalar to the bias) [Protonu]
>> PR: https://github.com/NVIDIA/Fuser/pull/2138

kevinstephano:
![image](https://github.com/NVIDIA/Fuser/assets/8293715/6299802f-f299-47c6-b7ee-f23ff67edd84)

Priya2698:
This issue will be now resolved through work on new IR nodes: https://github.com/NVIDIA/Fuser/issues/2149

Priya2698:
Closed by #2240

PR #2240, opened 2024-05-14
`LinearOp` IR node

Priya2698:
1. Add a `LinearOp` node that has the same functionality as `F.linear`.
2. Update `linear_op_generator` to cover all cases allowed in thunder / `F.linear`.

Issue #2149, #2132.


tests/python/pytest_fusion_definitions.py
```diff
@@ -21,14 +21,18 @@ def parse_inputs_fusion_definition(fd: FusionDefinition, opinfo: OpInfo, *args):
 
     nvf_args = []
 
-    if opinfo.symbolic_parameter_list is None:
-        opinfo.symbolic_parameter_list = [ArgumentType.Symbolic] * len(args)
-    num_symbolic_parameters = len(opinfo.symbolic_parameter_list)
+    symbolic_parameter_list = (
```
Priya2698, line 24: Avoid updating `opinfo.symbolic_parameter_list` directly since ops like `linear` can have variable number of parameters.

Priya2698:
!build

Priya2698:
!build

csrc/ir/internal_nodes.h
```diff
@@ -2288,4 +2288,50 @@ class MatmulOp : public Expr {
       const std::vector<PolymorphicValue>& inputs) const override;
 };
 
+//! Linear Operator to be expression evaluated without decomposition.
```
jacobhinkle, line 2291: You could leave a comment here that points to the torch docs for semantics of this node: https://pytorch.org/docs/stable/generated/torch.nn.functional.linear.html#torch.nn.functional.linear

csrc/root_domain_map.cpp
```diff
@@ -197,6 +197,52 @@ std::unordered_map<IterDomain*, IterDomain*> PairwiseRootDomainMap::map(
...
+      }
+      case MatmulRole::INPUT_C: {
+        // The last dimension of LinearOp is out_features.
+        updatePairwiseRootDomainMap(
+            producer_root.at(0), consumer_root.at(out_size - 1));
+        break;
+      }
+      default:
+        NVF_ERROR("Unexpected input type.");
+    }
```
jacobhinkle, line 242: Is there any value in factoring this out the way you did with `ops::mapMatmulOpIterDomains`?

csrc/ops/composite.cpp
```diff
@@ -54,42 +54,66 @@ TensorView* dropout_backward(TensorView* dy, TensorView* mask, Val* scale) {
...
+} // namespace
+
+TensorView* linear(TensorView* tv_a, TensorView* tv_b, TensorView* bias) {
+  NVF_CHECK(tv_a->nDims() >= 1, "Input A must be atleast 1D.");
+  NVF_CHECK(
+      tv_b->nDims() == 1 || tv_b->nDims() == 2,
+      "Input B must be a 1D / 2D tensor.");
+  NVF_CHECK(
+      tv_b->nDims() == 1 && bias != nullptr,
+      "Input B must be a 2D tensor if bias is present, got 1D.")
```
jacobhinkle, line 107: I think this condition should be `bias == nullptr || (tv_b->nDims() == bias->nDims() + 1)`. Also note though that `nDims()` is the size of the leaf domain which includes reductions. In an unsegmented fusion we might actually have reductions that feed straight to matmul or linear, so the dims should be computed via `TensorDomain::noReductions(tv->getMaybeRFactorDomain()).size()` instead.

tests/cpp/test_matmul_aten_evaluation.cpp
```diff
@@ -482,4 +590,20 @@ INSTANTIATE_TEST_SUITE_P(
...
 
+INSTANTIATE_TEST_SUITE_P(
+    LinearWithoutBias,
+    LinearNodeParametrizedTest,
+    testing::Combine(
+        testing::Values(Sizes({k}), Sizes({m, k}), Sizes({b, m, k})),
+        testing::Values(Sizes({k}), Sizes({n, k})),
+        testing::Values(std::nullopt)));
+
+NTIATE_TEST_SUITE_P(
```
jacobhinkle, line 601: Looks like this line might have been eaten by lintrunner or something.

tests/cpp/test_matmul_aten_evaluation.cpp
```diff
@@ -482,4 +590,20 @@ INSTANTIATE_TEST_SUITE_P(
...
+    LinearWithBias,
+    LinearNodeParametrizedTest,
+    testing::Combine(
+        testing::Values(Sizes({k}), Sizes({m, k}), Sizes({b, m, k})),
+        testing::Values(Sizes({n, k})),
+        testing::Values(Sizes({n}))));
```
jacobhinkle, line 607: Can we add broadcast dims here?

tests/python/pytest_input_generators.py
```diff
@@ -1524,18 +1524,21 @@ def linear_input_generator(
...
-            map(pow, itertools.repeat(step, num_steps), range(1, num_steps + 1))
-        )
+    B = 64
+    M = 512
+    N = 256
+    K = 32
+
+    # Cases without bias
+    shapes_input = ((K), (M, K), (B, M, K))
+    shapes_weight = ((K), (N, K))
```
jacobhinkle, line 1534: Can we test with broadcasts? What about more than one batch dim?

csrc/root_domain_map.cpp
Priya2698: Yes, I'll factor it out now. This was a simple enough mapping so we didn't need the utility necessarily but given that we may make changes to the mapping for K dimension, it is useful to consolidate the changes.

csrc/ops/composite.cpp
Priya2698:
```
>>> a = torch.randn(2, 2, dtype=torch.float16)
>>> b = torch.randn(2, dtype=torch.float16)
>>> bias = torch.tensor(3.14, dtype=torch.float16)
>>> torch.nn.functional.linear(a, b, bias)
  File "<stdin>", line 1, in <module>
RuntimeError: mat2 must be a matrix, got 1-D tensor
```

Bias is only accepted when B (weight) is a matrix.

jacobhinkle: Thanks for checking! That constraint is not mentioned in the docs. Let's put that in a comment that although the docs don't mention it, as of now bias is restricted to the case of 2D weight arguments.

Priya2698: Another case that is supported is a 2D bias although the documents do not mention it (The current mapping will not support this case): `A = [2, 4, 6], B [3, 6], Bias = [2*4, 3] `.

jacobhinkle: Should the bias not be [ 4, 3 ] in that case?

Priya2698: No, that fails actually. It looks for the merge of all the non-reduction dimensions in the first operand.

csrc/ops/composite.cpp
```diff
@@ -54,42 +54,71 @@ TensorView* dropout_backward(TensorView* dy, TensorView* mask, Val* scale) {
...
+  const std::vector<IterDomain*>& mapping_b = ops::mapLinearOpIterDomains(
+      orig_domain_b, MatmulRole::INPUT_B, ndims_out);
+  std::vector<IterDomain*> mapping_bias (ndims_out, nullptr);
+  if (bias != nullptr){
+    auto bias_domain = TensorDomain::noReductions(bias->getMaybeRFactorDomain());
+    mapping_bias = ops::mapLinearOpIterDomains(bias_domain, MatmulRole::INPUT_C, ndims_out);
   }
-  return maybeCastOp(a->getDataType().value(), output);
+
+  std::vector<IterDomain*> out_domain = ops::newOutputDomain({mapping_a, mapping_b, mapping_bias});
```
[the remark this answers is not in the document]
Priya2698: What do you mean by take precedence?
For each position, I am using all the corresponding iterdomains from all inputs to create the output iterdomain. You are correct though, that we can only use mapping_b and it will work.

However, if one is Symbolic and the other Concrete, we can resolve the extents for the output if we use both. Wdyt?

csrc/ops/composite.cpp
```diff
@@ -54,42 +54,71 @@ TensorView* dropout_backward(TensorView* dy, TensorView* mask, Val* scale) {
...
+} // namespace
+
+TensorView* linear(TensorView* tv_a, TensorView* tv_b, TensorView* bias) {
+  auto ndims_a = TensorDomain::noReductions(tv_a->getMaybeRFactorDomain()).size();
+  NVF_CHECK(ndims_a > 0, "Input A must be atleast 1D.");
+  
+  auto ndims_b = TensorDomain::noReductions(tv_b->getMaybeRFactorDomain()).size();
+  NVF_CHECK(ndims_b == 1 || ndims_b == 2, "Input B must be a 1D / 2D tensor.");
+
+  // Note: This constraint is not documented but F.linear errors out if bias is given with 1D weights.
```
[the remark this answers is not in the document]
jjsjann123: https://github.com/pytorch/pytorch/pull/126312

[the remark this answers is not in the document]
jjsjann123: NVM. I don't think we can/should push through expanding the torch implementation. I closed my upstream PR.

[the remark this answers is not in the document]
jjsjann123: https://github.com/pytorch/pytorch/pull/126312#issuecomment-2113528243

csrc/ops/utils.cpp
```diff
@@ -299,6 +338,26 @@ IterDomain* newOutputIterDomain(const std::vector<IterDomain*>& ids) {
...
+      "Tried to create new output Tensorview but received empty list.");
+
+  std::vector<IterDomain*> out_domain(input_ids.front().size(), nullptr);
+
+  for (const auto dim_i : c10::irange(out_domain.size())) {
+    std::vector<IterDomain*> ids_i;
+    ids_i.reserve(input_ids.size());
+    for (auto ids : input_ids) {
+      if (ids[dim_i] != nullptr){
+        ids_i.emplace_back(ids[dim_i]);
```
jjsjann123, line 353: This is what I meant by
> mapping_b will take precedence over mapping_bias.

`mapping_b` and `mapping_bias` would both have a mapping here for `N` dimension. So we are always taking `mapping_b`'s id for that dimension.

`torch.linear` doesn't support bias dominating the out_feature while `weight` has size-1 there, but that's just an implementation detail.

```
>>> x = torch.randn(2, 3)
>>> w = torch.randn(1, 3)
>>> b = torch.randn(5)
>>> torch.nn.functional.linear(x, w, b)
  File "<stdin>", line 1, in <module>
RuntimeError: The expanded size of the tensor (1) must match the existing size (5) at non-singleton dimension 1.  Target sizes: [2, 1].  Tensor sizes: [5]
```

jjsjann123: I think the example I posted should *in theory* work for nvfuser scheduled kernel. But I don't think the code we have is working properly. I think your output iter domain will be created as
`x->axis(0), x->axis(1), w->axis(0)`
but it should have been
`x->axis(0), x->axis(1), b->axis(0)`

Bus since this is a case where pytorch asserts, maybe it doesn't matter. Having said that, we should at least throw an error in this case.

tests/python/pytest_input_generators.py
```diff
@@ -1524,18 +1524,21 @@ def linear_input_generator(
...
-        lhs_shape = (M, K)
-        rhs_shape = (N, K)
-        yield (SampleInput(make_arg(lhs_shape), make_arg(rhs_shape), make_arg((N,))))
+    # Cases with bias
+    shape_weight = (N, K)
+    shapes_bias = ((), (N,))
+    for shape_input, shape_bias in itertools.product(shapes_input, shapes_bias):
+        yield SampleInput(
+            make_arg(shape_input), make_arg(shape_weight), make_arg(shape_bias)
+        )
```
jjsjann123, line 1544: can we do an error_input_generator for linear as well so we have legit error message thrown. Sorry I keep nitpicking on these corner cases.

csrc/ops/utils.cpp
Priya2698: > This is what I meant by
> 
> > mapping_b will take precedence over mapping_bias.
> 
> `mapping_b` and `mapping_bias` would both have a mapping here for `N` dimension. So we are always taking `mapping_b`'s id for that dimension.
> 
> `torch.linear` doesn't support bias dominating the out_feature while `weight` has size-1 there, but that's just an implementation detail.
> 
> ```
> >>> x = torch.randn(2, 3)
> >>> w = torch.randn(1, 3)
> >>> b = torch.randn(5)
> >>> torch.nn.functional.linear(x, w, b)
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> RuntimeError: The expanded size of the tensor (1) must match the existing size (5) at non-singleton dimension 1.  Target sizes: [2, 1].  Tensor sizes: [5]
> ```

Yes, that is correct. Bias can be broadcast but if the weight is broadcast in N dimension, then bias can only be broadcast.
I took both the IDs to create the ID for the `out_features` since if we have a non-broadcast N dimension and weight is symbolic but bias is concrete tensor, then the extent for the N dimension in the output can be resolved.

Priya2698: > I think the example I posted should _in theory_ work for nvfuser scheduled kernel. But I don't think the code we have is working properly. I think your output iter domain will be created as `x->axis(0), x->axis(1), w->axis(0)` but it should have been `x->axis(0), x->axis(1), b->axis(0)`
> 
> Bus since this is a case where pytorch asserts, maybe it doesn't matter. Having said that, we should at least throw an error in this case.

Right, we should assert on both being broadcast if weight has N=1

jjsjann123: Ha, sorry I was not reading this correctly. I didn't notice this is taking all the IDs. sorry about that.

:face_with_head_bandage:

tests/python/pytest_input_generators.py
jjsjann123: I don't have any other concern out side of test coverage for the `unsupported` cases.

I think we want to see exceptions when
`1d weight + bias`
`2d weight (size-1 out feature) + bias (not size-1 out feature)`

csrc/ops/composite.cpp
```diff
@@ -54,42 +54,71 @@ TensorView* dropout_backward(TensorView* dy, TensorView* mask, Val* scale) {
...
+    bias == nullptr || bias->dtype() == tv_a->dtype(),
+    "Expected bias to have the same dtype as A and B, got: ", bias->dtype(), " and ", tv_b->dtype()
+  );
+  // For all other cases, create a new LinearOp
+  TensorView* out = newForLinear(tv_a, tv_b, bias);
+  IrBuilder::create<LinearOp>(out, tv_a, tv_b, bias);
+  return out;
+}
+
+TensorView* linear(TensorView* tv_a, TensorView* tv_b) {
```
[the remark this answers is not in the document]
protonu: bad memories :)

csrc/ir/internal_nodes.h
```diff
@@ -2288,4 +2288,51 @@ class MatmulOp : public Expr {
       const std::vector<PolymorphicValue>& inputs) const override;
 };
 
+// Linear Operator to be expression evaluated without decomposition.
```
protonu, line 2291: I may have this wrong, but later we may decide to evaluate this with nvfuser?
If so, we may not need to mention anything here.

csrc/ops/composite.cpp
```diff
@@ -54,42 +54,73 @@ TensorView* dropout_backward(TensorView* dy, TensorView* mask, Val* scale) {
...
+  const std::vector<IterDomain*>& mapping_b = ops::mapLinearOpIterDomains(
+      orig_domain_b, MatmulRole::INPUT_B, ndims_out);
+  std::vector<IterDomain*> mapping_bias (ndims_out, nullptr);
+  if (bias != nullptr){
+    auto bias_domain = TensorDomain::noReductions(bias->getMaybeRFactorDomain());
+    mapping_bias = ops::mapLinearOpIterDomains(bias_domain, MatmulRole::INPUT_C, ndims_out);
   }
-  return maybeCastOp(a->getDataType().value(), output);
+
+  std::vector<IterDomain*> out_domain = ops::newOutputDomain({mapping_a, mapping_b, mapping_bias});
```
Priya2698, line 83: Adding a new method `ops::newOutputDomain` to isolate most of the mapping logic to `mapLinearOpIterDomains`. 

Since now we have to add a reduction axis, we can remove that as well, and directly use `newOutputIterDomains`.

tests/python/pytest_input_generators.py
Priya2698: IIUC, error_generator will check if we the error message matches.

For the second case though, I am relying on torch to throw the error since we may have symbolic tensors and cannot assert if the extents will be the same.

Priya2698: Oh nvm, I can match the torch errors too.

csrc/ops/composite.cpp
jacobhinkle: See my recent push to #2246. Once that is merged we can do `ops::newOutputIterDomain({mapping_a.back(), mapping_b.back()}, /*force_iter_type=*/IterType::Reductoin)` just like I'm doing with `MatmulOp` there.

csrc/ops/composite.cpp
```diff
@@ -314,12 +316,24 @@ static TensorView* newForMatmul(TensorView* tv_a, TensorView* tv_b) {
...
+  
   const std::vector<IterDomain*>& mapping_a = ops::mapMatmulOpIterDomains(
       orig_domain_a, MatmulRole::INPUT_A, ndims_out);
   const std::vector<IterDomain*>& mapping_b = ops::mapMatmulOpIterDomains(
       orig_domain_b, MatmulRole::INPUT_B, ndims_out);
 
-  std::vector<IterDomain*> out_domain = ops::newOutputDomain({mapping_a, mapping_b});
+  for (auto idx : c10::irange(ndims_out)) {
+    std::vector<IterDomain*> input_ids;
+    input_ids.reserve(2);
```
jacobhinkle: nit: if `input_ids.size()` is not always 2 then just skip reserving.

csrc/ops/composite.h
```diff
@@ -47,10 +47,8 @@ NVF_API LstmResult lstm(
     TensorView* cell_x,
     TensorView* out_x);
 
-// Linear functions which takes in two tensors of shapes A[M,K] and
-// B[N,K]. Takes in a options bias of shape [N] and performs
-// out = A * B_Transpose + bias. The output dtype matches the dtype
-// ofthe inputs which should match.
+// Linear functions which takes in two tensors of shapes A[* , in_features], B[out_features, in_features] / [in_features] and an optional bias of shape [out_features] or 0D scalar.
```
jacobhinkle, line 50: Did this pass lintrunner? It seems like we need to wrap this line.

jacobhinkle: For clarity it might make sense to call the parameters `input`, `weight`, and `bias` here since that is what `F.linear` calls them.

csrc/ops/utils.h
```diff
@@ -51,11 +51,22 @@ std::vector<IterDomain*> mapMatmulOpIterDomains(
     MatmulRole input_role,
     size_t out_size);
 
+// For LinearOp, the output is the same as the first input (A[*, in_features])for all but the last dimension.
+// If the second input is 2D (B[out_features, in_features]), the last dimension of output is out_features.
+// If bias is 1D (bias[out_features]) it maps to the last dimension of the output.
+std::vector<IterDomain*> mapLinearOpIterDomains(
```
jacobhinkle, line 57: Please document the `out_size` option since it is not obvious at first glance why it is needed.

csrc/root_domain_map.cpp
```diff
@@ -197,6 +197,42 @@ std::unordered_map<IterDomain*, IterDomain*> PairwiseRootDomainMap::map(
...
+      IterDomain* producer_id = aligned_producer_ids.at(inx);
+      IterDomain* consumer_id = consumer_root.at(inx);
+      if (producer_id == nullptr) {
+        continue;
+      }
+      updatePairwiseRootDomainMap(producer_id, consumer_id);
+    }
+
+    return dom_map;
+  }
```
jacobhinkle, line 234: Is it worth reusing some of this code between matmul and linear? I am not sure if anything but the last loop can be re-used but it seems like there might be a way since the patterns are so similar.

tests/cpp/test_matmul_aten_evaluation.cpp
```diff
@@ -457,12 +470,107 @@ TEST_P(ATenNodesParametrizedTest, MatmulNodeSymbolic) {
...
+  if (bias_shape.has_value()) {
+    fusion->addInput(bias);
+  }
+  fusion->addOutput(tv2);
+
+  at::Tensor t0 = at::randn(a_shape, at::kHalf).cuda();
+  at::Tensor t1 = at::randn(b_shape, at::kHalf).cuda();
+  std::optional<at::Tensor> bias_opt = std::nullopt;
+  if (bias_shape.has_value()) {
+    bias_opt = bias_shape.value().empty() ? at::scalar_tensor(3.14).to(at::kHalf).cuda() : at::randn(*bias_shape, at::kHalf).cuda();
```
jacobhinkle, line 546: Does the regular randn call not work when `bias_shape` is empty?

tests/cpp/test_matmul_aten_evaluation.cpp
```diff
@@ -457,12 +470,107 @@ TEST_P(ATenNodesParametrizedTest, MatmulNodeSymbolic) {
...
+  }
+  auto tv2 = linear(tv0, tv1, bias);
+
+  fusion->addInput(tv0);
+  fusion->addInput(tv1);
+  if (bias_shape.has_value()) {
+    fusion->addInput(bias);
+  }
+  fusion->addOutput(tv2);
+
```
jacobhinkle, line 493: Please test `IdModel` mapping cases here like we do for `MatmulOp` in #2246

tests/python/test_python_frontend.py
```diff
@@ -2446,7 +2445,7 @@ def fusion_func(
             fd.add_output(t_out)
 
         in_tensors = [inputs_mk_nk, inputs_mk_kn, inputs_km_nk, inputs_km_kn]
-        use_bias = [None, bias0d, bias1d, bias2d]
+        use_bias = [None, bias0d, bias1d]
```
jacobhinkle, line 2448: Does 2D bias not work or are we just not going to support it since it's not mentioned in the `F.linear` docs?

jacobhinkle (commented):
Looking good. I had a few minor questions but the only thing I think is missing is a test for IterDomain mapping like in #2246.

tests/python/test_python_frontend.py
jacobhinkle: Oh I guess this is because we cannot map domains properly in that case? https://github.com/NVIDIA/Fuser/pull/2240#discussion_r1600726831

I wonder if this is a case we will encounter. We should be sure to guard our thunder prim so that it does not send us linears with 2D bias.

csrc/ops/composite.cpp
jacobhinkle: The crazy thing is that this is particular to 3D input. If you have 4D input then you can use a 4D bias without any merged dim as expected, but for some reason 3D input requires 2D bias where the first two dims are merged. This is something we could support if we want to, but since it's undocumented we could also just disallow 2D bias as we have here and prevent thunder from sending us such linears. wdyt @jjsjann123

Priya2698: Thunder does not accept 2D bias: https://github.com/Lightning-AI/lightning-thunder/blob/2fa97d6b1fead1eb9d0dc17703ebb12bbf752bbc/thunder/core/prims.py#L3456

tests/python/test_python_frontend.py
Priya2698: https://github.com/Lightning-AI/lightning-thunder/blob/2fa97d6b1fead1eb9d0dc17703ebb12bbf752bbc/thunder/core/prims.py#L3456. They are not supported by Thunder either.

tests/cpp/test_matmul_aten_evaluation.cpp
Priya2698: Had errors with binding the output of `at::tensor` and `at::randn` for empty bias shape (it was identified as rank 1). `torch.randn(0)` also creates an empty tensor instead of a scalar.

jjsjann123: you should be able to do that.

`torch.randn(0)` will give you an `empty` tensor instead of a scalar tensor. I think you need `torch.randn()`. instead.
Similarly, if you want to use `at::tensor`, make sure you are passing in the value as a scalar rather than a vector.

i.e. do `at::tensor(1.5)` instead of `at::tensor({1.5})`.

jjsjann123: FYI:
```
  auto options = at::TensorOptions()
                     .device(at::kCUDA, 0);
  std::vector<int64_t> shape = {};
  at::Tensor t6 = at::randn(shape, options.dtype(at::kFloat));
  std::cout << t6 << std::endl;
```
```
-0.924662
[ CUDAFloatType{} ]
```
This worked for me so you shouldn't need to worry about the conditional above.

Priya2698: Right thanks, `torch.randn(())` generates a scalar whereas `torch.randn()` gave me an empty tensor giving the impression I needed to use another call

csrc/root_domain_map.cpp
Priya2698: I'll put the last loop in a separate fn atleast. Will see if combining the rest of the snippet is clean enough or not since we will still need to distinguish between matmul/linear

Priya2698:
!build

tests/cpp/test_matmul_aten_evaluation.cpp
```diff
@@ -588,4 +740,43 @@ INSTANTIATE_TEST_SUITE_P(
             Sizes({1, 1}),
             Sizes({b, 1, n}))));
 
+INSTANTIATE_TEST_SUITE_P(
```
Priya2698, line 743: I still need to think about a good naming scheme: Underscore separated shapes looked too verbose and for filtering we will need the exact shapes.

jacobhinkle (approved):
LGTM. Thanks for adding the ID mapping checks.

tests/cpp/test_matmul_aten_evaluation.cpp
```diff
@@ -588,4 +740,43 @@ INSTANTIATE_TEST_SUITE_P(
...
+            Sizes({k}),
+            Sizes({m, k}),
+            Sizes({b, m, k}),
+            Sizes({1, k}),
+            Sizes({b, 1, k})),
+        testing::Values(Sizes({k}), Sizes({n, k}), Sizes({1, k})),
+        testing::Values(std::nullopt)));
+
+INSTANTIATE_TEST_SUITE_P(
+    LinearWithBias,
```
jacobhinkle, line 757: I'm getting an error running `build/test_matmul --gtest_filter='*LinearWithBias'`:
> C++ exception with description "Vectorized dim for producer has to be from a contiguous inner most position. tv: T5_g[ iS99{( ceilDiv(64, 128) )}, iS372{( ceilDiv(128, 64) )}, iS374{( ceilDiv(64, 16) )}, iS378{( ceilDiv(8, 2) )}, iS376{( ceilDiv(16, 8) )}, iS379{2} ], allocation domain: iS13{64}, vectorized id: iS13{64}, innermost id: iS13{64}, contiguity: f
Exception raised from getVectorizedIdInAllocationDomain at /opt/pytorch/nvfuser/csrc/device_lower/validation.cpp:433 (most recent call first):

PR #2243, opened 2024-05-14
Host IRs: add Communications

samnordmann:
## What
This PR introduces the utilization and support of `Communications` as Host IRs. This PR allows host program to be any combination of kernel launches and network communications

## How

The patch mainly deals with HostIrExecutor where we needed to add the support for network communication. The implementation is mainly copy-pasted from `MultiDeviceExecutor`.
At the abstract IR level, we also had to make `PostOnStream` accept `Communication` (whereas before it only accepted `HostUnit`)

## Example
 the file `tests/cpp/test_multidevice_host_ir.cpp` contains an exemple that illustrate a situation summarized as:

```
    tv0: input, sharded across devices on its first dimension
    tv1 = Fusion (tv0), on each device
    tv2 = Allgather(tv1)
    tv2: output
```

the output of `HostIrContainer::print()` is
```
%HostIrContainer {

HOST FUSION (T0_g[ ideviceIdx.x0{8}, iS1{8}, iS2{32} ] (DeviceMesh{0, 1, 2, 3, 4, 5, 6, 7, })) -> (T2_g[ iS6{8}, iS7{8}, iS8{32} ] (DeviceMesh{0, 1, 2, 3, 4, 5, 6, 7, })) :
  PostOnStream the operation :{

%kernel {
T1_g[ ideviceIdx.x3{8}, iS4{8}, iS5{32} ] (DeviceMesh{0, 1, 2, 3, 4, 5, 6, 7, })
   = T0_g[ ideviceIdx.x0{8}, iS1{8}, iS2{32} ] (DeviceMesh{0, 1, 2, 3, 4, 5, 6, 7, })
   + T0_g[ ideviceIdx.x0{8}, iS1{8}, iS2{32} ] (DeviceMesh{0, 1, 2, 3, 4, 5, 6, 7, });
}

}, taking inputs: {      T0_g      [ ideviceIdx.x0{8}, iS1{8}, iS2{32} ] (DeviceMesh{0, 1, 2, 3, 4, 5, 6, 7, })
  } and outputs: {
      T1_g      [ ideviceIdx.x3{8}, iS4{8}, iS5{32} ] (DeviceMesh{0, 1, 2, 3, 4, 5, 6, 7, })
  }
  PostOnStream the operation :{
  Communication Allgather: {
  team: 0 1 2 3 4 5 6 7,
}
}, taking inputs: {      T1_g      [ ideviceIdx.x3{8}, iS4{8}, iS5{32} ] (DeviceMesh{0, 1, 2, 3, 4, 5, 6, 7, })
  } and outputs: {
      T2_g      [ iS6{8}, iS7{8}, iS8{32} ] (DeviceMesh{0, 1, 2, 3, 4, 5, 6, 7, })
  }
END.
```


samnordmann:
!build --dist

csrc/host_ir/host_ir.cpp
```diff
@@ -71,7 +83,9 @@ NVFUSER_DEFINE_CLONE_AND_CREATE(PostOnStream)
 std::string PostOnStream::toString(int indent_size) const {
   int indent_increment = 2;
   std::stringstream ss;
-  indent(ss, indent_size) << "Execute the following Fusion, taking inputs :{\n";
+  indent(ss, indent_size) << "PostOnStream the operation :{\n";
+  indent(ss, indent_size) << hostOpToPost()->toString();
```
wujingyue, line 87:
```suggestion
  indent(ss, indent_size) << hostOpToPost();
```

IIRC, we've defined `operator<<` for Expr* so it's the same as calling toString. But please double check that.

csrc/host_ir/host_ir.cpp
```diff
@@ -71,7 +83,9 @@ NVFUSER_DEFINE_CLONE_AND_CREATE(PostOnStream)
 std::string PostOnStream::toString(int indent_size) const {
   int indent_increment = 2;
   std::stringstream ss;
-  indent(ss, indent_size) << "Execute the following Fusion, taking inputs :{\n";
+  indent(ss, indent_size) << "PostOnStream the operation :{\n";
```
wujingyue, line 86:
```suggestion
  indent(ss, indent_size) << "PostOnStream the operation :{" << std::endl;
```

I understand our codebase has been mixing \n and std::endl. @naoyam , what's your opinion here? Even though we don't serve Window users at this moment, the docker image we publish (including pytorch and apex etc) does support Windows. So a little bit of typing tax seems worthwhile for potential portability.

csrc/host_ir/executor.h
```diff
@@ -47,20 +48,25 @@ class HostIrExecutor final : public OptInDispatch {
  public:
   HostIrExecutor(
       std::unique_ptr<HostIrContainer> container,
+      Communicator* communicator = nullptr,
```
wujingyue, line 51: Question: there's also the approach of making the communicator a singleton that can be accessed globally. Do you prefer one way or the other? What would be the practical limitation if we chose singleton?

wujingyue (commented):
LGTM in general -- I'll wrap up the details later today.

csrc/host_ir/host_ir.cpp
samnordmann: Right! I'll fix that

csrc/host_ir/executor.h
samnordmann: using a singleton sounds good to me!

tests/cpp/test_multidevice_host_ir.cpp
```diff
@@ -0,0 +1,190 @@
...
+* SPDX-FileCopyrightText: Copyright (c) 2023-present NVIDIA CORPORATION & AFFILIATES.
+* All rights reserved.
+* SPDX-License-Identifier: BSD-3-Clause
+*/
+// clang-format on
+#include <gtest/gtest.h>
+
+#include <codegen.h>
+#include <device_lower/lower2device.h>
+#include <disjoint_set.h>
```
wujingyue, line 12: I'd clean up these header files. Many of them don't seem necessary and including them wastes compilation time. There may be some open-source tools to help with this such as IWYU.

tests/cpp/test_multidevice_host_ir.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+namespace hir {
+
+using MultiDeviceHostIrTestParams = std::tuple<bool, bool>;
+
+class MultiDeviceHostIrTest
+    : public MultiDeviceTest,
+      public testing::WithParamInterface<MultiDeviceHostIrTestParams> {};
+
+/*
```
wujingyue, line 53:
```suggestion
//
```
for style consistency across the code base.

wujingyue (approved):

tests/cpp/test_multidevice_host_ir.cpp
samnordmann: I have cleaned it up

samnordmann:
!build --dist

samnordmann:
!build --dist

csrc/host_ir/host_ir.cpp
naoyam: Actually, that overload didn't seem to work consistently and nobody couldn't figure out why. I remember it didn't work for my build (just raw addressed displayed) but worked for someone else. We ended up switching to be more explicit with `toString()`.

csrc/host_ir/host_ir.cpp
naoyam: What we print is mostly for informational debugging messages, so I think it's also good to use `std::endl` as it flushes the stream.

csrc/host_ir/executor.h
cowanmeg: I realize now we use one communicator with multiple process groups...

csrc/host_ir/host_ir.cpp
```diff
@@ -48,42 +48,52 @@ bool HostUnit::sameAs(const Statement* other) const {
...
-      this->outputs().size() == hu->fusion_to_execute()->outputs().size());
-  // TODO: harden the assert checks with smth like
-  // for (int i : c10::irange(inputs.size())) {
-  //     // NVF_ERROR(inputs.at(i)->sameAs(executable_fusion->inputs().at(i)));
-  // }
-  // for (int i : c10::irange(outputs.size())) {
-  //     //
-  //     NVF_ERROR(outputs.at(i)->sameAs(executable_fusion->outputs().at(i)));
-  // }
+  if (host_op->isA<HostUnit>()) {
```
naoyam, line 56: Could you add some more info on what `host_op` is allowed to be? IIUC, at this point, it should be either `HostUnit` or `Communication`?

csrc/host_ir/executor.h
```diff
@@ -47,20 +48,25 @@ class HostIrExecutor final : public OptInDispatch {
...
       HostIrExecutorParams = HostIrExecutorParams());
-  std::vector<at::Tensor> runWithInput(const std::vector<c10::IValue>& inputs);
+  std::vector<at::Tensor> runWithInput(
+      std::unordered_map<Val*, c10::IValue> val_to_IValue);
 
  private:
   using OptInDispatch::handle;
   void handle(PostOnStream* post) override;
+  void postCompute(PostOnStream* post);
+  void postCommunication(PostOnStream* post);
```
naoyam, line 60: nit: I find this naming a little unintuitive. "post" in "postCompute" and "postCommunication" are a verb, but then it also appears as the parameter.  Maybe renaming the parameter to something "host_op"?

csrc/host_ir/executor.h
samnordmann: > I realize now we use one communicator with multiple process groups...

yes, correct! So imo we are good for now

csrc/host_ir/host_ir.cpp
samnordmann: > IIUC, at this point, it should be either HostUnit or Communication?

correct!

> Could you add some more info on what host_op is allowed to be?

I'll add a else branch with a `NVF_ERROR(false,...)` to assert that. I can also add some comment about that where `PostOnStream` is defined

csrc/host_ir/executor.h
samnordmann: right, I see what you mean. Wdyt about naming it `post_host_op`, or `post_ir` ?

csrc/host_ir/host_ir.cpp
samnordmann: ok let me reverse that then, ok?

> Actually, that overload didn't seem to work consistently and nobody couldn't figure out why.

naive guess: maybe because `ir/iostream.h` header is missing sometimes?

wujingyue: > naive guess: maybe because ir/iostream.h header is missing sometimes?

Very likely. https://github.com/NVIDIA/Fuser/blob/f0bbb74a97bfca6c4810473513c6b094391a7be7/csrc/ir/iostream.h#L93

That's why it's important to IWYU, which however lacks a good tool to autocomplete and enforce. For example, when seeing `operator<<(const Statement*)`, a tool wouldn't know whether the intention is to print the address or the content. 

That being said, `operator<<(const Statement&)` may work better because people will have to include the right header to compile. However, the downside is that the input can't be null. 

> ok let me reverse that then, ok?

I prefer including `iostream.h` and saving a bit of typing here and there. But either way SGTM.

csrc/host_ir/host_ir.cpp
samnordmann: Weird, I needed to add surrounding parenthis for it to work...

PR #2246, opened 2024-05-14
[WIP] Add MatmulOp to ir_utils::isTvOp

jacobhinkle:
This also adds testing of exact mapping to the node tests (WIP).

### DIFF
--- a/tests/cpp/test_matmul_aten_evaluation.cpp
+++ b/tests/cpp/test_matmul_aten_evaluation.cpp
@@ -9,6 +9,7 @@
 #include <gtest/gtest.h>
 
 #include <fusion.h>
+#include <id_model/id_model.h>
 #include <mma_type.h>
 #include <ops/all_ops.h>
 #include <preseg_passes/allocation_order_inference.h>
@@ -409,6 +410,41 @@ TEST_F(MatmulATenEvaluationTest, LinearWithBias) {
   EXPECT_TRUE(at::allclose(out[0], out_ref));
 }
 
+// Check that ID exact mapping works as expected
+void checkMatmulOpIdMapping(
+    Fusion* fusion,
+    TensorView* A,
+    TensorView* B,
+    TensorView* output) {
+  IdModel id_model(fusion);
+  const ValGraph& vg = id_model.idGraph(IdMappingMode::EXACT);
+  vg.validateConsistency();
+
+  const auto checkMapped = [&vg](IterDomain* x, IterDomain* y) -> bool {
+    if (!vg.hasGroup(x) || !vg.hasGroup(y)) {
+      return false;
+    }
+    const ValGroup& gx = vg.toGroup(x);
+    const ValGroup& gy = vg.toGroup(y);
+    return gx.get() == gy.get();
+  };
+
+  if (A->nDims() == 2 && B->nDims() == 2) {
+    // [M, K] @ [K, N] = [M, N]
+    ASSERT_EQ(output->nDims(), 2);
+    EXPECT_TRUE(checkMapped(A->axis(0), output->axis(0))); // M
+    EXPECT_TRUE(checkMapped(B->axis(1), output->axis(1))); // N
+    // EXPECT_TRUE(checkMapped(A->axis(1), B->axis(0))); // K
+  } else if (A->nDims() == 1 && B->nDims() == 1) {
+    // [M, K] @ [K, N] = [M, N]
+    EXPECT_EQ(output->nDims(), 0);
+    // EXPECT_TRUE(checkMapped(A->axis(0), B->axis(0))); // K
+  } else {
+    std::cout << "Unhandled set of input dimensions" << std::endl;
+    // EXPECT_TRUE(false);
+  }
+}
+
 TEST_P(ATenNodesParametrizedTest, MatmulNodeConcrete) {
   auto fusion = std::make_unique<Fusion>();
   FusionGuard fg(fusion.get());
@@ -422,6 +458,8 @@ TEST_P(ATenNodesParametrizedTest, MatmulNodeConcrete) {
   fusion->addInput(tv0);
   fusion->addInput(tv1);
   fusion->addOutput(tv2);
+
+  checkMatmulOpIdMapping(fusion.get(), tv0, tv1, tv2);
 
   at::Tensor t0 = at::randn(a_shape, at::kHalf).cuda();
   at::Tensor t1 = at::randn(b_shape, at::kHalf).cuda();
@@ -446,6 +484,8 @@ TEST_P(ATenNodesParametrizedTest, MatmulNodeSymbolic) {
   fusion->addInput(tv0);
   fusion->addInput(tv1);
   fusion->addOutput(tv2);
+
+  checkMatmulOpIdMapping(fusion.get(), tv0, tv1, tv2);
 
   at::Tensor t0 = at::randn(a_shape, at::kHalf).cuda();
   at::Tensor t1 = at::randn(b_shape, at::kHalf).cuda();
--- a/csrc/device_lower/utils.cpp
+++ b/csrc/device_lower/utils.cpp
@@ -149,6 +149,7 @@ bool isTvOp(const Expr* expr) {
           WelfordOp,
           GroupedWelfordOp,
           LoadStoreOp,
+          MatmulOp,
           MmaOp,
           BroadcastOp,
           SqueezeOp,


tests/cpp/test_matmul_aten_evaluation.cpp
```diff
@@ -409,6 +410,41 @@ TEST_F(MatmulATenEvaluationTest, LinearWithBias) {
...
+    const ValGroup& gy = vg.toGroup(y);
+    return gx.get() == gy.get();
+  };
+
+  if (A->nDims() == 2 && B->nDims() == 2) {
+    // [M, K] @ [K, N] = [M, N]
+    ASSERT_EQ(output->nDims(), 2);
+    EXPECT_TRUE(checkMapped(A->axis(0), output->axis(0))); // M
+    EXPECT_TRUE(checkMapped(B->axis(1), output->axis(1))); // N
+    // EXPECT_TRUE(checkMapped(A->axis(1), B->axis(0))); // K
```
jacobhinkle, line 437: Currently failing since we do not map K at all.

csrc/ops/composite.cpp
```diff
@@ -304,6 +304,10 @@ static TensorView* newForMatmul(TensorView* tv_a, TensorView* tv_b) {
     out_domain[idx] = ops::newOutputIterDomain(input_ids);
   }
 
+  out_domain[ndims_out - 1] = IterDomainBuilder(mapping_a.back())
```
jacobhinkle, line 307: I placed the rK dimension last in the output, just because that makes it a bit easier to handle the many cases we can encounter for `matmul`. Note that this does not need to match the position of the K dimension in a translated `MmaOp`.

csrc/ops/composite.cpp
```diff
@@ -331,12 +335,44 @@ TensorView* matmul(TensorView* tv_a, TensorView* tv_b) {
       " and ",
       tv_b->dtype());
 
+  // Check for K=1 i.e. reduction of broadcast. In these cases we don't need a
```
jacobhinkle, line 338: I considered placing this translation to `mul` in another PR, but it is relatively compact and it is difficult to work around if we want to exhaustively check ID mappings with all different combinations of inputs.

csrc/ops/utils.cpp
```diff
@@ -190,29 +190,34 @@ std::vector<IterDomain*> mapMatmulOpIterDomains(
   std::vector<IterDomain*> mapping(out_size, nullptr);
   auto inp_size = (int64_t)input_domain.size();
 
+  // Input A to matmul: {*, M, K}
+  // Input B to matmul: {*, K, N}
+  auto kpos = input_role == MatmulRole::INPUT_A ? inp_size - 1 : inp_size - 2;
```
jacobhinkle, line 195: Changes in this file are to accomodate the new Reduction output domain by mapping it to the k position in each operand.

tests/cpp/test_matmul_aten_evaluation.cpp
```diff
@@ -480,6 +575,17 @@ INSTANTIATE_TEST_SUITE_P(
 INSTANTIATE_TEST_SUITE_P(
     ReductionAxisIsOne,
     ATenNodesParametrizedTest,
-    testing::Values(std::make_tuple(Sizes({m, 1}), Sizes({1, n}))));
+    testing::Combine(
```
jacobhinkle, line 578: Off-topic: It's not needed here, but for new tests I wonder if we could combine these kinds of parametrizations? For example since we use k=32, we could parametrize the test with a boolean flag `k_is_one` and if the flag is true we would replace any 32s in the shapes with 1 before running the test. That would reduce the need to repeat all the inputs for each case, but it makes it harder to filter to only the concrete K tests since the tests are simply numbered. It would be great if the parameters were reflected in the name like they are in pytest...

csrc/ops/composite.cpp
```diff
@@ -276,13 +276,13 @@ static TensorView* newForMatmul(TensorView* tv_a, TensorView* tv_b) {
   auto ndims_b = orig_domain_b.size();
 
   // Matmul output size is same as the higher dimensional input size if both A/B
-  // > 1D.
-  auto ndims_out = std::max(ndims_a, ndims_b);
+  // > 1D, but with 1 additional IterType::Reduction axis rK.
+  auto ndims_out = std::max(ndims_a, ndims_b) + 1;
```
jacobhinkle, line 280: Changes in this file are for adding a new `Reduction` output domain to facilitate easier mapping of K axes.

jacobhinkle:
!build --diff

csrc/ops/composite.cpp
Priya2698: Should we use both, mapping_a and mapping_b here?

csrc/ops/composite.cpp
```diff
@@ -331,12 +335,44 @@ TensorView* matmul(TensorView* tv_a, TensorView* tv_b) {
...
+    return maybeCastOp(tv_a->dtype(), float_result);
+  }
+
   if (tv_a->nDims() == 1 && tv_b->nDims() == 1) {
     // Return the dot product instead of creating the MatmulOp.
     // Cast back the output if needed since torch.matmul maintains input dtype.
     return maybeCastOp(tv_a->dtype(), sum(mul(tv_a, tv_b), {0}));
   }
 
+  if (tv_b->nDims() > 1 && tv_b->axis(-2)->isBroadcast()) {
```
Priya2698, line 369: This condition is already checked above in lines 342-343.

csrc/ops/composite.cpp
jacobhinkle: Good question. I was assuming the K dimension would be `IterType::Iteration` in both operands. They will be exact mapped, so they should be equivalent to one another so I just took the ID from A.

tests/cpp/test_matmul_aten_evaluation.cpp
```diff
@@ -409,6 +410,96 @@ TEST_F(MatmulATenEvaluationTest, LinearWithBias) {
...
+      int64_t i_out = output->nDims() - red_dims - 3 - i;
+      if (i_a >= 0 && !A->axis(i_a)->isBroadcast()) {
+        EXPECT_TRUE(checkMapped(A->axis(i_a), output->axis(i_out)));
+      }
+      if (i_b >= 0 && !B->axis(i_b)->isBroadcast()) {
+        EXPECT_TRUE(checkMapped(B->axis(i_b), output->axis(i_out)));
+      }
+    }
+  } else {
+    std::cerr << "Unhandled set of input dimensions" << std::endl;
```
Priya2698, line 498: nit: Consider using `NVF_ERROR` here.

csrc/ops/composite.cpp
jacobhinkle: Good catch.  Removing.

tests/cpp/test_matmul_aten_evaluation.cpp
jacobhinkle: I wanted to use `EXPECT_TRUE` to prevent stopping the test when we encounter this condition. But `EXPECT_*` does not accept any additional error message. We don't really need to print this though; I could just place it in a comment then the test would still fail if we had some new cases that weren't handled and we would see the comment when we investigate.

tests/cpp/test_matmul_aten_evaluation.cpp
Priya2698: We could add another flag, and replace the value of K within tests. I am not sure if there is a way to replace the values before they are passed to a test: that is, the test directly views K=1.  I'll look into it.

csrc/ops/composite.cpp
Priya2698: What happens if one is `Symbolic` and the other `Concrete` tensor?

csrc/ops/composite.cpp
```diff
@@ -331,6 +335,31 @@ TensorView* matmul(TensorView* tv_a, TensorView* tv_b) {
       " and ",
       tv_b->dtype());
 
+  // Check for K=1 i.e. reduction of broadcast. In these cases we don't need a
+  // matmul so we translate it to a multiplication+cast
```
Priya2698, line 339: Can we have a comment at the `matmul` definition or in the header enumerating the different cases and how they are handled for an easier summary.

csrc/ops/composite.cpp
Priya2698: Thanks for adding these cases!

tests/cpp/test_matmul_aten_evaluation.cpp
Priya2698: Oh right, `EXPECT_TRUE` does not abort. We can stick with this then.

Priya2698 (approved):
LGTM overall.

My main question is around using iterdomains from both A and B for mapping. While they will be exact mapped, what happens we may have one Symbolic and one Concrete tensor (although rare, and we do not test this). The `newOutputIterDomain` call be able to resolve the extents using one of the tensors if we use both. Wdyt?

jacobhinkle:
I agree about symbolic domains and I think we might need to handle this at concretization. Namely if K is a broadcast dim then we should translate to a mul op before segmentation. I'll work on an example for a case like that.

csrc/ops/composite.cpp
jacobhinkle: Good point. I updated `ops::newOutputIterDomain` to have a `std::optional<IterType> force_iter_type` argument that I am now using to create the reduction domain. That way, if the A ID is Symbolic but the B ID is not, then we'll use the B extent.

PR #2247, opened 2024-05-15
Allow FusionExecutorCache to take preallocated outputs

samnordmann:
this patch allows the user to also pass outputs to `FusionExecutorCache::runFusionWithInput`. Before the patch, the outputs would be necessary allocated by the executor.

The core capability is already supported by `FusionExecutor`. This pr only implements plumbing the output from `FusionExecutorCache` down to `FusionExecutor`

### DIFF
--- a/tests/cpp/test_gpu3.cpp
+++ b/tests/cpp/test_gpu3.cpp
@@ -31,6 +31,7 @@
 #include <kernel_ir.h>
 #include <kernel_ir_dispatch.h>
 #include <ops/all_ops.h>
+#include <ops/utils.h>
 #include <root_domain_map.h>
 #include <scheduler/all_schedulers.h>
 #include <scheduler/reduction_utils.h>
@@ -8207,36 +8208,28 @@ TEST_F(NVFuserTest, preallocated_outputs) {
   auto fusion = std::make_unique<Fusion>();
   FusionGuard fg(fusion.get());
 
-  TensorView* tv0 = makeSymbolicTensor(2);
-  TensorView* tv1 = makeSymbolicTensor(1);
-  TensorView* tv2 = makeSymbolicTensor(2);
+  TensorView* tv0 = makeSymbolicTensor(2); // Group 0
+  TensorView* tv1 = add(tv0, IrBuilder::create<Val>(1.0)); // Group 0
+  TensorView* tv2 = ops::newValLike(tv1, tv1->getDataType().value())
+                        ->as<TensorView>(); // Group 1
+  IrBuilder::create<LoadStoreOp>(LoadStoreOpType::SegmenterSet, tv2, tv1);
+  TensorView* tv3 = add(tv2, IrBuilder::create<Val>(1.0)); // Group 1
 
   fusion->addInput(tv0);
-  fusion->addInput(tv1);
-  fusion->addInput(tv2);
+  fusion->addOutput(tv3);
 
-  TensorView* tv3 = add(tv0, IrBuilder::create<Val>(1.0)); // Group 0
-  TensorView* tv4 = max(tv3, {0}); // Group 0
-  TensorView* tv5 = add(tv4, tv1); // Group 0
-  TensorView* tv6 = add(tv5, tv2); // Group 1
-
-  fusion->addOutput(tv6);
-
+  std::vector<int64_t> sizes = {8, 5};
   auto options = at::TensorOptions().dtype(at::kFloat).device(at::kCUDA, 0);
-  at::Tensor t0 = at::randn({8, 5}, options);
-  at::Tensor t1 = at::randn({5}, options);
-  at::Tensor t2 = at::randn({8, 5}, options);
-  at::Tensor t6 = at::empty({8, 5}, options);
+  at::Tensor t0 = at::randn(sizes, options);
+  at::Tensor t3 = at::empty(sizes, options);
 
-  std::vector<c10::IValue> aten_inputs = {t0, t1, t2};
-  std::vector<at::Tensor> preallocated_output = {t6};
+  std::vector<c10::IValue> aten_inputs = {t0};
+  std::vector<at::Tensor> preallocated_output = {t3};
 
   FusionExecutorCache executor_cache(std::move(fusion));
   auto outputs =
       executor_cache.runFusionWithInputs(aten_inputs, preallocated_output);
 
-  EXPECT_TRUE(executor_cache.getMostRecentKernelRuntime()->isSegmented())
-      << "segmentation didn't happen";
   EXPECT_EQ(
       executor_cache.getMostRecentKernelRuntime()
           ->fusionSegments()


samnordmann:
!build

samnordmann:
!build

samnordmann:
!build

cowanmeg (approved):
LGTM! Is this so that the host can control when buffers are allocated?

naoyam (changes requested):
Please add tests.

naoyam:
> Please add tests.

Please make sure to have segmented tests. I'm not sure if a given list of vector for a fusion is properly passed to each segment.

samnordmann:
> LGTM! Is this so that the host can control when buffers are allocated?

Yes! To have a better control on allocation and to be able to reuse pre-allocated buffers in different fusion executions

samnordmann:
> > Please add tests.
> 
> Please make sure to have segmented tests. I'm not sure if a given list of vector for a fusion is properly passed to each segment.

Makes sense, that's what I did. Let me know what you think

samnordmann:
!build

csrc/executor.cpp
```diff
@@ -1908,9 +1908,11 @@ std::vector<at::Tensor> FusionExecutor::runFusion(
   }
 
   NVF_ERROR(validKernelId(), "Invalid kernel id for FusionExecutor.");
-  NVF_ERROR(
-      !args.getCacheId().has_value() || outputs.empty(),
-      "short cut input cache is not compatible with pre-allocated output");
+
+  // short cut input cache is not compatible with pre-allocated output
+  if (!outputs.empty()) {
+    args.getCacheId() = std::nullopt;
```
jacobhinkle, line 1914: Does this mean a new `ExecutorEntry` must be created each time we launch with preallocated outputs? What if we include outputs in `args` or possibly create another `KernelArgumentHolder` that just holds the outputs, then combine the IDs?

jacobhinkle (commented):
This is nice! Could you add a test where incompatible outputs are specified, to check that we are properly validating provided outputs? For example incorrect sizes, discontiguous when should be contiguous, non-trivial stride order.

tests/cpp/test_gpu3.cpp
```diff
@@ -8202,6 +8202,53 @@ TEST_F(NVFuserTest, ReverseMerge) {
...
+  TensorView* tv2 = makeSymbolicTensor(2);
+
+  fusion->addInput(tv0);
+  fusion->addInput(tv1);
+  fusion->addInput(tv2);
+
+  TensorView* tv3 = add(tv0, IrBuilder::create<Val>(1.0)); // Group 0
+  TensorView* tv4 = max(tv3, {0}); // Group 0
+  TensorView* tv5 = add(tv4, tv1); // Group 0
+  TensorView* tv6 = add(tv5, tv2); // Group 1
```
wujingyue, line 8221: I'm sure that you've carefully constructed the example so it triggers segmentation, and thanks for testing that. But segment_set is probably an easier and more future-proof way to trigger segmentation if that's your main concern.

tests/cpp/test_gpu3.cpp
```diff
@@ -8202,6 +8202,53 @@ TEST_F(NVFuserTest, ReverseMerge) {
...
+
+  std::vector<c10::IValue> aten_inputs = {t0, t1, t2};
+  std::vector<at::Tensor> preallocated_output = {t6};
+
+  FusionExecutorCache executor_cache(std::move(fusion));
+  auto outputs =
+      executor_cache.runFusionWithInputs(aten_inputs, preallocated_output);
+
+  EXPECT_TRUE(executor_cache.getMostRecentKernelRuntime()->isSegmented())
+      << "segmentation didn't happen";
```
wujingyue, line 8239: The following EXPECT_EQ subsumes this one.

wujingyue (commented):
Some drive-by comments. Thanks for adding this feature!

tests/cpp/test_gpu3.cpp
samnordmann: right, let me remove it

tests/cpp/test_gpu3.cpp
samnordmann: right, I did as you suggest and simplified the test. Thanks!

csrc/kernel_cache.cpp
```diff
@@ -1470,9 +1487,16 @@ std::unordered_map<Val*, const PolymorphicValue*> FusionKernelRuntime::
     // TODO: currently we are still outputing PyTorch tensors, instead of
     // something abstract. This is quite unsatisfying.
 
+    std::vector<at::Tensor> group_runtime_preallocated_outputs;
+    for (auto output : group_to_run->outputs()) {
+      if (output_tensor_map.count(output)) {
```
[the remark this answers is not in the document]
naoyam: So, how would this work in a fusion like:

- The fusion has two outputs
- A user provides two buffers
- Segmented into two kernels
- The first segment produces one of the two outputs and also an intermediate output that is fed into the next segment
- The second segment produces the another output

In this case, the first segment has two outputs but only one of them is provided by the user. Wouldn't this case result in an error?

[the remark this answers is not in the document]
samnordmann: yes you are right!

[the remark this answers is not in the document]
naoyam: That's fine. Just make sure to add some early error check.

csarofeen:
@jjsjann123 @wujingyue I wonder if this type of functionality could be useful for fusions when we have concat. I was thinking if it would be beneficial to artificially do a "concat" by allocation a contiguous concat tensor and passing portions of non-contiguous tensors to the fusion. This way the concat is done by allocation + strided writes, rather than our current pad/masks approach.

wujingyue:
Taking preallocated outputs is useful for FusionExecutor, which I believe is already implemented but exercised only in unit tests. For example, what you just said is the "unzip+segment" approach in https://github.com/NVIDIA/Fuser/issues/1502, which requires FusionExecutor to take pre-allocated outputs. In addition, we'll probably need this if we want to call cublas/cudnn from a fusion because these math libraries don't allocate outputs by themselves. 

However, I'm uncertain about FusionExecutorCache at this point, which orchestrates the execution of the complete fusion. It could be useful if Thunder wants nvFuser to output to a particular buffer allocated by the downstream executor.

jjsjann123:
> This way the concat is done by allocation + strided writes, rather than our current pad/masks approach.

Agree that's a cleaner end result for codegen. But it is adding complexity to infrastructure code around it. My naive question here is, is `pad` considered a tricky thing to handle in codegen hence we try to avoid it?

samnordmann:
> artificially do a "concat" by allocation a contiguous concat tensor and passing portions of non-contiguous tensors to the fusion. This way the concat is done by allocation + strided writes,

This was exactly my motivation

> Taking preallocated outputs is useful for FusionExecutor, which I believe is already implemented but exercised only in unit tests.

Pre-allocating some but not all the output buffers is not supported, however this is a critical piece to enable that feature, as pointed out by Naoya https://github.com/NVIDIA/Fuser/pull/2247#discussion_r1626422131

PR #2251, opened 2024-05-15
Generate a DOT representation of fusion transform.

naoyam:
This is a simple debugging utility that I started using for IdModel. Thought it may be of useful in general.

I'm not a Graphviz expert but just tried a bunch of things to make it the output look reasonably good to me. Any suggestions would be appreciated.

Here's an example output. The fusion is from one of the reshape shmoo tests.

transform

### DIFF
--- a/csrc/ir/graphviz.cpp
+++ b/csrc/ir/graphviz.cpp
@@ -10,9 +10,11 @@
 #include <fusion.h>
 #include <ir/all_nodes.h>
 #include <ir/builder.h>
+#include <ir/utils.h>
 #include <type.h>
 
 #include <fstream>
+#include <sstream>
 
 namespace nvfuser {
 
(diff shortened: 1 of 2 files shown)


csrc/ir/graphviz.cpp
```diff
@@ -390,4 +392,145 @@ void IrGraphGenerator::handle(const TensorView* tv) {
...
+void TransformToDot::handle(IterDomain* id) {
+  if (printed_vals_.find(id) != printed_vals_.end()) {
+    return;
+  }
+
+  // Use blue for broadcast domains
+  indent() << id->name()
+           << " [fontcolor=" << (id->isBroadcast() ? "blue" : "black")
+           << "];\n";
+
```
jacobhinkle, line 517: In order to enforce vertical alignment, you can use the `rank` property https://graphviz.org/docs/attrs/rank/. Already we have the root domains aligned along the top row by default since they do not have producers. But this can also be enforced via `indent() << "{rank = \"min\"; " << id->name() << " [fontcolor=" ...`. Similarly we could align the loop domains along the bottom using `rank = \"max\"`

jacobhinkle (approved):
This is great! I wonder if it would be possible to visually depict ca_pos. I added a comment of how to vertically align the loop domain. To indicate ca_pos, maybe you could change the style of the inlined loops; e.g. all the inlined loops would be inside green ovals instead of black or something?

csrc/ir/graphviz.cpp
jacobhinkle: Also if there are no `uses()` for an ID then it might be helpful to add `[shape=circle]` to distinguish them from intermediates.

naoyam: What do you mean by vertical alignment?

naoyam:
> This is great! I wonder if it would be possible to visually depict ca_pos. I added a comment of how to vertically align the loop domain. To indicate ca_pos, maybe you could change the style of the inlined loops; e.g. all the inlined loops would be inside green ovals instead of black or something?

Yeah, my usual way of visualizing the position is to have a red bar between loop domains, but I have no idea how to do that. I often add more annotations to graphs (like shown here #2261), so I prefer just a plain simple one. Maybe we could add options, but for now that feels like overengineering.

naoyam:
I'm going to merge this for now as it's good enough for me. Follow-up PRs to improve the visualization is certainly appreciated!

PR #2252, opened 2024-05-15
reduce IdModel graph build in AllocationOrderInference

jjsjann123:


csrc/id_model/id_model.cpp
```diff
@@ -1263,6 +1263,7 @@ Expr* IdModel::addReplayAs(std::vector<IterDomain*> new_inputs, Expr* expr) {
 
   // Create the new expression with provided inputs
   auto replay = ReplayTransform::replayAs(new_inputs, expr);
+  NVF_ERROR(replay != nullptr, "no replay found");
```
jjsjann123, line 1266: cc'ing @naoyam 

Does `buildLoopGraph` have some assumption on what a fusion graph should look like, are we not supposed to build all graph from an unscheduled fusion with IdModel? I can open an issue if that's not the case. This PR is just a quick patch to make our thunder benchmark green again.

jjsjann123:
!build

naoyam:
Can you please give some info about what the issue is?

jjsjann123:
> Can you please give some info about what the issue is?

I have a issue opened on this: https://github.com/NVIDIA/Fuser/issues/2253

csrc/id_model/id_model.cpp
jjsjann123: tagging #2253

naoyam (approved):
LGTM

PR #2255, opened 2024-05-15
Make the naming style consistent.

wujingyue:


tests/cpp/test_multidevice_communications.cpp
```diff
@@ -26,24 +26,24 @@ class CommunicationTest
 
   void validate(at::Tensor obtained, at::Tensor expected);
 
-  static constexpr DeviceIdxType root = 0;
-  static constexpr int tensor_size = 1024;
+  static constexpr DeviceIdxType kRoot = 0;
+  static constexpr int kTensorSize = 1024;
```
wujingyue, line 30: The rest of the codebase doesn't consistently follow this leaking-k naming style either (some files do but most files don't). I chose to make these changes anyway because they make the code easier to read by distinguishing local variables and constant variables.

cowanmeg (approved):
LGTM! For future references, are we using camel case and k-prefix for all static const fields?

samnordmann (approved):
thanks!

wujingyue:
> LGTM! For future references, are we using camel case and k-prefix for all static const fields?

No, as I said in https://github.com/NVIDIA/Fuser/pull/2255#discussion_r1602277510, we are not following that style consistently. 

I think we should -- it would help me understand/review code faster. Btw, I didn't invent this style. It comes from https://google.github.io/styleguide/cppguide.html which nvFuser has been mostly following.

wujingyue:
!build

wujingyue:
!build

PR #2260, opened 2024-05-16
Multi-dimensional DeviceMesh

cowanmeg:
Allows DeviceMesh to be multi-dimensional. Devices are represented as a flattened list of device indices and a separate shape. 

Almost all calls to `vector` have been replaced with `getTeam` which returns the team (list of devices) for a device given the DeviceMesh axis. Note that `getTeam` creates a new vector each call which should be cached.

### DIFF
--- a/csrc/multidevice/device_mesh.cpp
+++ b/csrc/multidevice/device_mesh.cpp
@@ -119,7 +119,7 @@ void DeviceMesh::setDevices(std::vector<DeviceIdxType> devices) {
   return DeviceMesh(devices);
 }
 
-/*static*/ DeviceMesh DeviceMesh::createForShape(std::vector<int64_t>& shape) {
+/*static*/ DeviceMesh DeviceMesh::createForShape(std::vector<int64_t> shape) {
   int64_t num_devices = 1;
   for (auto i : c10::irange(shape.size())) {
     num_devices *= shape[i];
@@ -127,17 +127,6 @@ void DeviceMesh::setDevices(std::vector<DeviceIdxType> devices) {
   std::vector<DeviceIdxType> devices(num_devices);
   std::iota(devices.begin(), devices.end(), 0);
   return DeviceMesh(devices, shape);
-}
-
-/*static*/ DeviceMesh DeviceMesh::createForShape(
-    std::initializer_list<int64_t> shape) {
-  int64_t num_devices = 1;
-  for (auto i : c10::irange(shape.size())) {
-    num_devices *= shape[i];
-  }
-  std::vector<DeviceIdxType> devices(num_devices);
-  std::iota(devices.begin(), devices.end(), 0);
-  return DeviceMesh(devices, std::vector<int64_t>(shape));
 }
 
 std::ostream& operator<<(std::ostream& out, const DeviceMesh& mesh) {
--- a/csrc/multidevice/device_mesh.cpp
+++ b/csrc/multidevice/device_mesh.cpp
@@ -7,7 +7,6 @@
 // clang-format on
 
 #include <multidevice/device_mesh.h>
-#include <iostream>
 
 #include <numeric>
 #include <unordered_set>
--- a/csrc/multidevice/device_mesh.cpp
+++ b/csrc/multidevice/device_mesh.cpp
@@ -23,10 +23,8 @@ DeviceMesh::DeviceMesh(
   if (shape.empty()) {
     shape_ = {(int64_t)devices.size()};
   } else {
-    int64_t num_devices = 1;
-    for (auto i : c10::irange(shape.size())) {
-      num_devices *= shape[i];
-    }
+    int64_t num_devices = std::accumulate(
+        shape.begin(), shape.end(), 1, std::multiplies<int64_t>());
     NVF_ERROR(
         (int64_t)devices.size() == num_devices,
         "Specified a list of device with ",
@@ -43,8 +41,7 @@ DeviceMesh::DeviceMesh(std::initializer_list<DeviceIdxType> devices) {
   setDevices(std::vector<DeviceIdxType>(devices));
 }
 
-std::vector<int64_t> DeviceMesh::getLocalIndices(
-    const DeviceIdxType device) const {
+std::vector<int64_t> DeviceMesh::getIndices(const DeviceIdxType device) const {
   auto global_idx = idxOf(device);
   if (global_idx == -1) {
     return {};
@@ -64,7 +61,7 @@ std::pair<int64_t, int64_t> DeviceMesh::getTeamOffsetStride(
   int64_t offset = 0;
   int64_t stride = 1;
   int64_t accumulated_size = 1;
-  auto indices = getLocalIndices(device);
+  auto indices = getIndices(device);
   NVF_ERROR(!indices.empty(), "Device is not in DeviceMesh");
   for (int64_t i = (int64_t)shape_.size() - 1; i >= 0; i--) {
     if (i > axis) {
@@ -119,10 +116,8 @@ void DeviceMesh::setDevices(std::vector<DeviceIdxType> devices) {
 }
 
 /*static*/ DeviceMesh DeviceMesh::createForShape(std::vector<int64_t> shape) {
-  int64_t num_devices = 1;
-  for (auto i : c10::irange(shape.size())) {
-    num_devices *= shape[i];
-  }
+  int64_t num_devices = std::accumulate(
+      shape.begin(), shape.end(), 1, std::multiplies<int64_t>());
   std::vector<DeviceIdxType> devices(num_devices);
   std::iota(devices.begin(), devices.end(), 0);
   return DeviceMesh(devices, shape);
--- a/csrc/multidevice/device_mesh.cpp
+++ b/csrc/multidevice/device_mesh.cpp
@@ -7,21 +7,98 @@
 // clang-format on
 
 #include <multidevice/device_mesh.h>
+#include <iostream>
 
 #include <numeric>
 #include <unordered_set>
 
 // for operator<<(std::ostream&, const std::vector<T>&)
 #include <c10/util/Logging.h>
+#include <c10/util/irange.h>
 
 namespace nvfuser {
 
-DeviceMesh::DeviceMesh(std::vector<DeviceIdxType> devices) {
+DeviceMesh::DeviceMesh(
+    std::vector<DeviceIdxType> devices,
+    std::vector<int64_t> shape) {
+  if (shape.empty()) {
+    shape_ = {(int64_t)devices.size()};
+  } else {
+    int64_t num_devices = 1;
+    for (auto i : c10::irange(shape.size())) {
+      num_devices *= shape[i];
+    }
+    NVF_ERROR(
+        (int64_t)devices.size() == num_devices,
+        "Specified a list of device with ",
+        devices.size(),
+        " elements ",
+        " but shape contains ",
+        num_devices);
+    shape_ = std::move(shape);
+  }
   setDevices(std::move(devices));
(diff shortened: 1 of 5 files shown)


cowanmeg:
!build --dist

csrc/multidevice/device_mesh.h
```diff
@@ -31,23 +30,41 @@ class DeviceMesh final {
...
+  }
+
+  // Returns the shape of the device mesh
+  const std::vector<int64_t>& shape() const {
+    return shape_;
+  }
+
   // Returns a vector containing the device indices of the mesh
   const std::vector<DeviceIdxType>& vector() const {
     return vector_;
```
cowanmeg, line 70: Unless there is a reason to keep this, I think we can get rid of this function.

csrc/multidevice/device_mesh.h
```diff
@@ -67,20 +85,42 @@ class DeviceMesh final {
...
+  // Returns the max device index in the DeviceMesh.
+  DeviceIdxType maxDeviceIdx() const;
+
+  // Returns devices of this device's team given which axis of the DeviceMesh
+  // Ex: [[0 1 2]
+  //      [3 4 5]]
+  // getTeam(4, 0) = {3, 4, 5}
+  // getTeam(4, 1) = {1, 4}
+  // TODO: calculate this once and save it
+  std::vector<DeviceIdxType> getTeam(DeviceIdxType device, int64_t axis = 0)
```
cowanmeg, line 110: Right now this is parameterized by axis, but we might want to change this to ParallelType, where ParallelType has to be a DID type.

csrc/multidevice/device_mesh.h
```diff
@@ -67,20 +85,42 @@ class DeviceMesh final {
...
+      const;
+
  private:
   void setDevices(std::vector<DeviceIdxType> devices);
+  // Calculates offset and stride used to iterate over
+  std::pair<int64_t, int64_t> getTeamOffsetStride(
```
samnordmann, line 116: this probably doesn't need to be a class method. It can be either inlined where it's used or be an auxiliary function not declared in header

csrc/multidevice/device_mesh.h
```diff
@@ -67,20 +85,42 @@ class DeviceMesh final {
...
+  //      [3 4 5]]
+  // getTeam(4, 0) = {3, 4, 5}
+  // getTeam(4, 1) = {1, 4}
+  // TODO: calculate this once and save it
+  std::vector<DeviceIdxType> getTeam(DeviceIdxType device, int64_t axis = 0)
+      const;
+
  private:
   void setDevices(std::vector<DeviceIdxType> devices);
+  // Calculates offset and stride used to iterate over
```
samnordmann, line 115:
```suggestion
  // Calculates offset and stride used in `getTeam` to iterate over
```

csrc/multidevice/device_mesh.h
samnordmann: agreed

csrc/multidevice/device_mesh.h
samnordmann, line 110: also, we have been using the type `Team` to alias `std::vector<DeviceIdxType>`, I suggest we get rid of this alias -- otherwise we might want this function to use it as return type

csrc/multidevice/device_mesh.h
```diff
@@ -58,7 +75,8 @@ class DeviceMesh final {
     return std::find(vector_.begin(), vector_.end(), device) != vector_.end();
   }
 
-  // Returns the index of device in the mesh, or -1 if device is not present.
+  // Returns the global index of device in the mesh, or -1 if device is not
+  // present
   int64_t idxOf(const DeviceIdxType device) const {
```
samnordmann, line 80: this method might not be needed anymore, right ?

csrc/multidevice/device_mesh.h
```diff
@@ -67,20 +85,42 @@ class DeviceMesh final {
     return -1;
   }
 
-  // Returns the device at a particular index in the mesh
+  // Returns the indices of a multi-dimensional mesh, or an empty vector 
+  // if device is not present
+  std::vector<int64_t> getLocalIndices(const DeviceIdxType device) const;
```
samnordmann, line 90: we probably don't need to expose this method, right?

samnordmann (commented):
Thank you! It looks great to me (despite the numerous but very minor comments)!
I like the idea of exposing `getTeam`, and I think that it should allows to remove most of the other methods

csrc/multidevice/device_mesh.cpp
```diff
@@ -13,17 +13,93 @@
...
+  std::vector<int64_t> indices(shape_.size());
+  int64_t accumulated_size = 1;
+  for (int64_t i = (int64_t)shape_.size() - 1; i >= 0; i--) {
+    indices[i] = (global_idx / accumulated_size) % shape_[i];
+    accumulated_size *= shape_[i];
+  }
+  return indices;
+}
+
+std::pair<int64_t, int64_t> DeviceMesh::getTeamOffsetStride(
```
wujingyue, line 58: Is it crazy to let DeviceMesh store an at::Tensor? It feels like we are re-inventing parts of at::Tensor already.

wujingyue (commented):
Great work! I made some comments and are still reviewing the rest...

csrc/multidevice/device_mesh.h
```diff
@@ -58,7 +75,8 @@ class DeviceMesh final {
     return std::find(vector_.begin(), vector_.end(), device) != vector_.end();
   }
 
-  // Returns the index of device in the mesh, or -1 if device is not present.
+  // Returns the global index of device in the mesh, or -1 if device is not
```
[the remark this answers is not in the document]
cowanmeg: Sure, we can change the name, it's just the index in the flattened array of device.

csrc/multidevice/device_mesh.h
samnordmann: > How about updating the comment to make it more clear this is retrieving the communicator for a device given the axis?

> I'd prefer `getTeam` over `get1DSlice` since the later is describing what the output is rather than how it is implemented.

I would argue that `get1DSlice` refers to what the function *does*, not the implementation. The fact that we later use this 1D slice as a `Team` for a communication is an incidental choice that we can make, but that's not explicit nor straightforward, and actually that's not even always true:
1) in the current code, if the root is not in the mesh we append the root to the 1D slice to create the Team
2) When we will implement 2D, if, for instance, we want to bcast from mesh {0} to mesh {{0,1},{2,3}} we might wanna set the Team to be {0,1,2,3} (and root=0), so it's not a 1D slice

Does it make sense?

> How about updating the comment to make it more clear this is retrieving the communicator for a device given the axis?

I would use the word `Team` rather than `communicator`. I think suggestion has the same problem than before: it implicitely assumes that a Team is always a device mesh's 1D slice. This assumption is not necessary -- if we still decide to make such an assumption, we should state it explicitely in the comment

csrc/multidevice/device_mesh.h
```diff
@@ -31,23 +30,41 @@ class DeviceMesh final {
...
+  // shape.
+  static DeviceMesh createForShape(std::vector<int64_t> shape);
+
+  // Returns the total number of devices in the mesh
   int64_t size() const {
     return static_cast<int64_t>(vector_.size());
   }
 
+  // Returns size of an axis in the mesh
+  int64_t size(int64_t axis) {
```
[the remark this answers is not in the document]
samnordmann: np!

csrc/multidevice/device_mesh.h
samnordmann: yes, my point is that it doesn't even need to be a private member of the class since it doesn't use private data -- it could be only implemented in `device_mesh.cpp` in an anonymous namespace and not declared in the header (or simply inlined since it is used only once)

csrc/multidevice/device_mesh.cpp
```diff
@@ -13,17 +13,93 @@
...
+        devices.size(),
+        " elements ",
+        " but shape contains ",
+        num_devices);
+    shape_ = std::move(shape);
+  }
   setDevices(std::move(devices));
 }
 
 DeviceMesh::DeviceMesh(std::initializer_list<DeviceIdxType> devices) {
```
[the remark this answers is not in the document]
samnordmann: As far as I'm concerned `tv->setDeviceMesh(DeviceMesh({...}))` is ok

csrc/multidevice/device_mesh.h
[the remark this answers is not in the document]
samnordmann: I don't find the word "global" explicit as well, I would say smth like "the index in the flattened list"

csrc/multidevice/device_mesh.cpp
samnordmann: Imo it totally makes sense, and that's actually how I implemented it when I prototyped it a while ago, see here https://github.com/samnordmann/Fuser/blob/ff6347b60a85ab3ded979cd14b5fbbb23a1272a5/csrc/multidevice/device_mesh.h#L59

For now, I'm fine with either implementation, but obviously using `at::Tensor` will be probably more powerful in the long run (at some point we might wanna take multiD slices of mesh for example)

csrc/multidevice/device_mesh.h
samnordmann: actually imo shape() is enough... up to you

csrc/multidevice/device_mesh.h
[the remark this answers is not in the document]
samnordmann: actually imo shape() is enough... up to you

csrc/multidevice/device_mesh.cpp
wujingyue: I haven't tried to benchmark the memory storage overhead of at::Tensor. Meshes are currently embedded in every TensorView, so FusionDefinition size could grow significantly. (Sure, there's always a solution to share mesh across TensorViews, because most of them are indeed identical).

cowanmeg: Yes, we should share meshes! Another thing I was thinking about was having DeviceMesh be an abstract class with two subclasses (1) the current mesh where the user specifies the devices in any order and (2) a “dense” device mesh which requires the devices be sequential.  The argument for the dense device mesh is we wouldn’t need to store the device orderings, just the shape. This might be important if we want to support a large training job with thousands of devices. It would also reduce the lookups to calculate what devices are in a team which would get expensive. 

Granted, the workload could have a communication that requires the whole mesh which would materialize the mesh to create the Team. But, those large workloads are generally designed to only have each device talking to a subset of devices.

csrc/multidevice/device_mesh.h
```diff
@@ -67,20 +85,42 @@ class DeviceMesh final {
...
     return vector_.at(index);
   }
 
   bool operator==(const DeviceMesh& other) const {
-    return vector_ == other.vector();
+    return vector_ == other.vector() && shape_ == other.shape();
   }
 
+  // Returns the max device index in the DeviceMesh.
+  DeviceIdxType maxDeviceIdx() const;
```
wujingyue, line 101: If a method can be done without accesses to private fields, I tend to make it a helper function on the side. This way, we minimize the exposure of private fields so it's easier to change them in the future. 

```suggestion
DeviceIdxType maxDeviceIdx(const DeviceMesh&);
```

wujingyue (approved):
LGTM with comments!

csrc/multidevice/device_mesh.h
cowanmeg: It access private data....

samnordmann (approved):

PR #2265, opened 2024-05-17
Check that device is supported in matmul canScheduleCompileTime

jacobhinkle:
This small PR just ensures we won't accept matmul segments that we cannot schedule.

### DIFF
--- a/csrc/scheduler/matmul_utils.cpp
+++ b/csrc/scheduler/matmul_utils.cpp
@@ -462,8 +462,8 @@ std::string getMatmulCompileTimeRejectReason(Fusion* fusion) {
     const auto device_prop = at::cuda::getCurrentDeviceProperties();
     // Use a dummy problem shape to determine whether this is a supported
     // device.
-    const auto mma_op = getMmaOp(
-        device_prop->major * 10 + device_prop->minor, {128, 128, 128, 1});
+    const auto mma_op =
+        getMmaOp(device_prop->major * 10 + device_prop->minor, {128, 128, 128});
     if (!mma_op.has_value()) {
       return "Unsupported device compute capability";
     }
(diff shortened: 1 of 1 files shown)


csrc/scheduler/matmul_utils.cpp
```diff
@@ -449,13 +449,26 @@ std::string getMatmulRunTimeRejectReason(
...
   // 1. Check if there is exactly one MmaOp or suitable mul sum pair
   // defined in the fusion.
   // 2. Check if inputs to the mma op or mul sum pair match any of
   // supported inputs layout
   // 3. Check if fusion represents expressions that are recognized by matmul
   // scheduler.
 
+  // #0
+  {
+    const auto device_prop = at::cuda::getCurrentDeviceProperties();
```
Priya2698, line 462: We also have this utility here: https://github.com/NVIDIA/Fuser/blob/cdcb379d4e28dcb1359e6c673e5c5b162ece8414/csrc/scheduler/matmul_utils.cpp#L46-L68

This is used in `getMatmulHeuristics`:
https://github.com/NVIDIA/Fuser/blob/cdcb379d4e28dcb1359e6c673e5c5b162ece8414/csrc/scheduler/matmul_utils.cpp#L539-L543

These look like very similar use cases so we should preferably combine them.

Priya2698: Additionally, this check is present in `getMatmulHeuristics` which is called by `computeHeuristics` in the matmul scheduler constructor. Do we need this check additionally in `getCompileTimeRejectReason`. Maybe this check should only be in `getCompileTimeRejectReason` instead of raising an error in `getMatmulHeuristics`.

Priya2698: Ahh right, sorry I missed that is the same function being used.

Priya2698 (approved):

Priya2698:
!build

PR #2266, opened 2024-05-17
Fix: invalid TensorDomain after reordered merge

naoyam:
`TensorDomain::merge(i, j)` used to enforce merging outer and inner domains as the outer and inner input to the `Merge` expr. This changed a while ago to allow reordered merge, but the change was not enough.

This part assumes `axis_i` is larger than `axis_o`, which is not the case when `i` refers to an outer domain like below:

```
  fusion.addInput(tv0);

  auto tv1 = set(tv0);
  fusion.addOutput(tv1);

  tv1->merge(1, 0);
```

### DIFF
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -3663,9 +3663,14 @@ void TensorDomain::merge(int64_t axis_o, int64_t axis_i) {
 
   IterDomain* merged_id = IterDomain::merge(first, second);
 
-  leaf_domain_.erase(leaf_domain_.begin() + axis_i);
-  leaf_domain_.erase(leaf_domain_.begin() + axis_o);
-  leaf_domain_.insert(leaf_domain_.begin() + axis_o, merged_id);
+  // axis_o is the outer input of this merge but does not
+  // automatically mean it's an outer domain in TensorDomain.
+  auto td_outer_pos = axis_o < axis_i ? axis_o : axis_i;
+  auto td_inner_pos = axis_o < axis_i ? axis_i : axis_o;
+
+  leaf_domain_.erase(leaf_domain_.begin() + td_inner_pos);
+  leaf_domain_.erase(leaf_domain_.begin() + td_outer_pos);
+  leaf_domain_.insert(leaf_domain_.begin() + td_outer_pos, merged_id);
   resetDomains();
 }
 
--- a/tests/cpp/test_gpu3.cpp
+++ b/tests/cpp/test_gpu3.cpp
@@ -8174,6 +8174,34 @@ TEST_F(NVFuserTest, BlockReduction3D) {
   }
 }
 
+// Simple test to merge an inner domain as an outer input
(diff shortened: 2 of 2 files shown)


naoyam:
!build

tests/cpp/test_gpu3.cpp
```diff
@@ -8174,6 +8174,34 @@ TEST_F(NVFuserTest, BlockReduction3D) {
...
+  auto tv1 = set(tv0);
+  fusion.addOutput(tv1);
+
+  tv1->merge(1, 0);
+
+  ASSERT_EQ(tv1->nDims(), 1);
+  auto merge = dynamic_cast<Merge*>(tv1->axis(0)->definition());
+  ASSERT_NE(merge, nullptr);
+  ASSERT_EQ(merge->outer(), tv1->getRootDomain().at(1));
+  ASSERT_EQ(merge->inner(), tv1->getRootDomain().at(0));
```
jacobhinkle, line 8194: Nit: these could be EXPECT_EQ

csrc/ir/nodes.cpp
```diff
@@ -3663,9 +3663,14 @@ void TensorDomain::merge(int64_t axis_o, int64_t axis_i) {
...
-  leaf_domain_.erase(leaf_domain_.begin() + axis_o);
-  leaf_domain_.insert(leaf_domain_.begin() + axis_o, merged_id);
+  // axis_o is the outer input of this merge but does not
+  // automatically mean it's an outer domain in TensorDomain.
+  auto td_outer_pos = axis_o < axis_i ? axis_o : axis_i;
+  auto td_inner_pos = axis_o < axis_i ? axis_i : axis_o;
+
+  leaf_domain_.erase(leaf_domain_.begin() + td_inner_pos);
+  leaf_domain_.erase(leaf_domain_.begin() + td_outer_pos);
+  leaf_domain_.insert(leaf_domain_.begin() + td_outer_pos, merged_id);
```
jacobhinkle, line 3673: So we always insert the new domain at the outer loop domain position, not the _given_ "outer" position i.e. the first argument to merge()?

jacobhinkle (approved):
LGTM bc it fixes a real bug. See question about where the merged ID should be placed.

tests/cpp/test_gpu3.cpp
naoyam: It seems to me it'd make little sense to continue this test if this condition is not met, so `ASSERT` seems to make more sense to me.

csrc/ir/nodes.cpp
naoyam: Ah, that's an interesting question. Let me check a few more things.

naoyam: I think either should be fine. The fact that this simple bug hasn't showed up until now means nothing should depend on this behavior, so we could choose whichever as we like. I'd leave it as is.

naoyam:
!build

PR #2267, opened 2024-05-18
Remove MatmulATenEvaluationTest tests

jacobhinkle:
These tests exercise the ATen pattern matching to evaluate MmaOp. They depend on the Matmul scheduler accepting segments containing MmaOp. However, the Matmul scheduler now rejects segments unconditionally on unsupported architectures as of #2265. That means that on unsupported architectures these tests will fail at segmentation, even though we can still evaluate the patterns using ATen. However, we plan to remove the pattern matching ATen evaluation code soon to simplify the codebase, so this PR removes these tests in order to fix CI.


jacobhinkle:
This PR disables the tests. Another option would be to guard them with `NVFUSER_TEST_CUDA_ARCH_RANGE_GUARD(7, 5, 9, 0);` but since we are moving away from the ATen MmaOp codepath anyway I think it's safe to just remove the tests.

jacobhinkle:
!build

naoyam (approved):
LGTM

PR #2268, opened 2024-05-18
Remove pattern matching based ATen evaluation for matmul

Priya2698:
We are removing pattern matching based evaluation since we are now using new IR nodes `MatmulOp` / `LinearOp` (Issue #2149). Removing pattern matching, and setting the disable option to clean the codebase.


Priya2698:
!build

tests/cpp/test_double_buffering.cpp
```diff
@@ -457,7 +457,7 @@ TEST_F(DoubleBufferingTest, SmemBlockGemmCacheDoubleBuffer) {
   auto options = at::TensorOptions().dtype(at::kFloat).device(at::kCUDA, 0);
   at::Tensor t0 = at::randn({M, K}, options);
   at::Tensor t1 = at::randn({K, N}, options);
-  at::Tensor aten_output = matmul(t0.to(at::kDouble), t1.to(at::kDouble));
+  at::Tensor aten_output = at::matmul(t0.to(at::kDouble), t1.to(at::kDouble));
```
jacobhinkle, line 460: How is this working currently? There's no `using namespace at;` anywhere that I can see, and those are definitely `at::Tensor`s.

jacobhinkle (approved):
Nice cleanups!

tests/cpp/test_double_buffering.cpp
Priya2698: It is actually calling ATen matmul:
```
Thread 1 "nvfuser_tests" hit Breakpoint 2, at::matmul (self=..., other=...) at /usr/local/lib/python3.10/dist-packages/torch/include/ATen/ops/matmul.h:26
26      inline at::Tensor matmul(const at::Tensor & self, const at::Tensor & other) {
          return at::_ops::matmul::call(self, other);
```

I wasn't aware it could be used without the namespace either.

PR #2269, opened 2024-05-18
Volatile patch

jjsjann123:
Current reduced precision in our runtime header relies on implicit copy constructor & copy assignment operator.
`T::T(const T&)` and `T& T::operator=(const T&)`.

Unfortunately, this would not handle cases where we have `volatile` instances. For generated code with: `*(volatile __half*)&T10[i3] = T4[0];` or `T4[0] = *(volatile __half*)&T10[i3];`, we'll run into errors in nvrtc
```
error: no operator "=" matches these operands
operand types are: volatile <unnamed>::__half = <unnamed>::__half
```

This PR added overload for volatile instances listed below (Thanks to @jacobhinkle for listing it out):
i.e. this should cover cases 
for copy constructor
`volatile` / `non-volatile`
for copy assignment operator
`non-volatile = non-volatile` / `non-volatile = volatile` / `volatile = non-volatile` / `volatile = volatile`
```
T::T(const T&);
T::T(const T&&);
T::T(const volatile T&);
T::T(const volatile T&&);
void T::operator=(const T&);
void T::operator=(const T&&);
void T::operator=(const volatile T&);
void T::operator=(const volatile T&&);
void T::operator=(const T&) volatile;
void T::operator=(const T&&) volatile;
void T::operator=(const volatile T&) volatile;
void T::operator=(const volatile T&&) volatile;
```

- Note 1:
For each volatile argument, we need overload for `const volatile T&` and `const volatile T&&`, because rvalue can only bind to const non-volatile lvalue reference.
This could happen with a example like this: `T4[0] = predicate ? *(volatile __half*)&T10[i3] : 1.5;` where ternary operator would resolve to an volatile rvalue.
Technically, `const T&` could also be bound by rvalue `T&&`. But due to the overload with `const volatile T&&`, which precedes `const T&`, we decided to also add an overload of `const T&&` to avoid that. Thanks to @naoyam for catching this one.

- Note 2:
I explicitly changed the copy assignment operator to NOT return reference to self.

I'm seeing gcc warning in my toy example when copy assignment returns a volatile lvalue reference when it's trivially used as `*(volatile XXX*)s_ptr = s_instance;`. `implicit dereference will not access object of type `volatile XXX` in statement`.

This is really annoying (for reference about the warning, see this: https://gcc.gnu.org/onlinedocs/gcc-4.0.4/gcc/Volatiles.html)
IIUC, we don't use the return from copy assignment operator anyway. I'm happy to revisit this one.


jjsjann123:
!build

jjsjann123:
Happen to run into this when I was desperately playing with concat

```
import torch
import thunder

def apply_rope(x, y) -> torch.Tensor:
    rotated = torch.cat((-x, y), dim=-1)  # (B, nh, T, hs)
    return rotated

dtype = torch.half
#dtype = torch.float
device = "cuda"
bsz = 2
block_size = 20
n_head = 8
head_size = 8
n_query_groups = 4
rope_n_elem = 8


q = torch.randn([bsz, block_size, (n_head + 2 * n_query_groups) * head_size], device=device, dtype=dtype)

jit_apply_rope = thunder.jit(apply_rope, nv_enable_bookend=False)
print(f"after split q shape: {q.shape} and stride: {q.stride()}")
q = jit_apply_rope(q[..., : 20], q[..., 20 : ])
print(thunder.last_traces(jit_apply_rope)[-1])
```

running with `NVFUSER_ENABLE=memory_promotion python rope_repro_0.py`

jjsjann123:
!build

jjsjann123:
!build

runtime/bf16_support.cu
```diff
@@ -16,6 +16,22 @@ __device__ __inline__ __bfloat __float2bfloat(const float);
...
 
+  __bfloat(const volatile __bfloat& other) {
+    __x = other.__x;
+  }
+
+  __bfloat(const volatile __bfloat&& other) {
+    __x = other.__x;
+  }
+
+  __device__ void operator=(const volatile __bfloat& other) volatile {
```
wujingyue, line 27: For my education, why "volatile" at the end of the line is necessary?

jacobhinkle: I think it's because you'd hit a compile error assigning a volatile object to a non-volatile one unless the method was qualified.

My question is: will this automatically promote `other` to `volatile` in case you do `a = b` and neither are volatile? If so then any assignments of these types will be treated as volatile which seems not right. Would we need another method `__device__ void operator=(const __bfloat& other) {`? My understanding is that if you implemented one of the default operators, then all the default operators would not be created. But since there's no default volatile assignment operator does that apply? Also do you need the vol=nonvol and nonvol=vol cases like `operator=(const __bfloat& other) volatile {`?

jjsjann123: > My understanding is that if you implemented one of the default operators, then all the default operators would not be created.

good catch. I thought I can get away with volatile assignment operator but tried that and seems to be not the case.

so yes I'll need another set of overload for 

novol = vol
vol = novol
__bfloat(const __bfloat&)

jjsjann123:
!build

jjsjann123:
This one should be good for review now :bow:

naoyam:
Could you add a repro as a test?

runtime/bf16_support.cu
naoyam: > I think it's because you'd hit a compile error assigning a volatile object to a non-volatile one unless the method was qualified.

Isn't it because assigning whatever value to a volatile object requires the `volatile` qualifier? At least, that's what I understood from this [doc](https://en.cppreference.com/w/cpp/language/member_functions#Member_functions_with_cv-qualifiers).

naoyam:
Just to confirm, 

```
void T::operator=(const volatile T&& volatile);
```

Did you mean below?

```
void T::operator=(const volatile T&&) volatile;
```

jjsjann123:
> Just to confirm,
> 
> ```
> void T::operator=(const volatile T&& volatile);
> ```
> 
> Did you mean below?
> 
> ```
> void T::operator=(const volatile T&&) volatile;
> ```

Good catch. Updated in comment.

naoyam:
Could somebody explain why this actually means?

```
T::T(const volatile T&&);
```

Or 
```
void T::operator=(const volatile T&&);
```

Probably my understanding of rvalue references needs to be updated, but this would refer to some temporary, which we don't know where it's allocated, then what does `volatile` mean?

naoyam:
> Could somebody explain why this actually means?
> 
> ```
> T::T(const volatile T&&);
> ```
> 
> Or
> 
> ```
> void T::operator=(const volatile T&&);
> ```
> 
> Probably my understanding of rvalue references needs to be updated, but this would refer to some temporary, which we don't know where it's allocated, then what does `volatile` mean?

Would this hit the volatile rvalue overload?

```
volatile __float16 x;
__float16 y;
y = std::move(x);
```

I wonder why we would need a move assignment for this struct, though.

naoyam:
> I'm seeing gcc warning in my toy example when copy assignment returns a volatile lvalue reference when it's trivially used as *(volatile XXX*)s_ptr = s_instance;. implicit dereference will not access object of type volatile XXX in statement.

> This is really annoying (for reference about the warning, see this: https://gcc.gnu.org/onlinedocs/gcc-4.0.4/gcc/Volatiles.html)
> IIUC, we don't use the return from copy assignment operator anyway. I'm happy to revisit this one.

I think this is fine. Could you please leave a code comment why the return type is `void`?

runtime/bf16_support.cu
```diff
@@ -16,6 +16,42 @@ __device__ __inline__ __bfloat __float2bfloat(const float);
...
+
+  __device__ void operator=(const __bfloat& other) {
+    __x = other.__x;
+  }
+
+  __device__ void operator=(const volatile __bfloat& other) {
+    __x = other.__x;
+  }
+
+  __device__ void operator=(const volatile __bfloat&& other) {
```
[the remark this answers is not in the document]
jjsjann123: non-volatile temporary is just an rvalue and it can be bound to `const T&`. 

i.e. if you have
```
void foo(const T& x);
foo(std::move(x)); // This will call into `void foo(const T& x)`
```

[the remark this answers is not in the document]
jjsjann123: Do you think it's better to leave an explicit move here, in case we need to change behavior in the future?

To be fair, @jacobhinkle 's [comment earlier](https://github.com/NVIDIA/Fuser/pull/2269#discussion_r1606698934) regarding the various overload of `volatile` on lhs/rhs operands falls into about the same bucket here. i.e. generated code conceptually isn't any different, but having an explicit overload written out might help to eliminate the confusion and doubts.

runtime/bf16_support.cu
jjsjann123: Yes. `volatile` qualifier at the end marks that `self` pointed to instance is volatile. Hence you get volatile access to things like `(this->)__x`.

jjsjann123:
!build

runtime/bf16_support.cu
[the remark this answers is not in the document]
jjsjann123: Added the overload of `const T&&` to avoid overload resolution going through `const volatile T&&` preceding `const T&`. Detailed put in Note 1 in PR description.
Thanks for catching this one.

jjsjann123:
!build

runtime/fp8_support.cu
```diff
@@ -149,6 +201,58 @@ __device__ __inline__ __e5m2 __float2e5m2(const float);
 struct __align__(1) __e5m2 {
   __e5m2() = default;
 
+  __e5m2(const __e5m2& other) {
```
jacobhinkle, line 204: Off-topic: could `__half`, `__bfloat`, `__e4m3`, and `__e5m2` all be written as instantiations of a template:
```c++
template <typename Underlying>
struct __align__(sizeof(Underlying)) ReducedPrecisionType {
  // ...
 protected:
  Underlying __x;
};
using __half = ReducedPrecisionType<uint16_t>;
using __bfloat = ReducedPrecisionType<uint16_t>;
using __e4m3 = ReducedPrecisionType<uint8_t>;
using __e5m2 = ReducedPrecisionType<uint8_t>;
```

jacobhinkle (commented):
So we now have these methods introduced in this PR:
```c++
  T(const T& other);
  T(const T&& other);
  T(const volatile T& other);
  T(const volatile T&& other);
  __device__ void operator=(const T& other);
  __device__ void operator=(const T&& other);
  __device__ void operator=(const volatile T& other);
  __device__ void operator=(const volatile T&& other);
  __device__ void operator=(const T& other) volatile;
  __device__ void operator=(const T&& other) volatile;
  __device__ void operator=(const volatile T& other) volatile;
  __device__ void operator=(const volatile T&& other) volatile;
```
Looks like a complete list to me but I'll leave approval to the others.

naoyam (approved):
LGTM. Thanks for addressing this issue. I think it's time to refactor these classes with a template but that doesn't need to be done as part of this PR.

jjsjann123:
lol, `torch.cat` can't handle fp8. So I shouldn't use that in the reference... But I do want to test fp8. I guess I'll just manually compute reference.
```
00:32:39 unknown file: Failure
00:32:39 C++ exception with description ""normal_kernel_cuda" not implemented for 'Float8_e4m3fn'
00:32:39 Exception raised from operator() at /opt/pytorch/pytorch/aten/src/ATen/native/cuda/DistributionTemplates.h:437 (most recent call first):
00:32:39 frame #0: c10::Error::Error(c10::SourceLocation, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) + 0xac (0x7fc68a34f45c in /usr/local/lib/python3.10/dist-packages/torch/lib/libc10.so)
```

jjsjann123:
!build

PR #2271, opened 2024-05-20
Set allocation domain of sharded tensor

cowanmeg:
Sets allocation domain of sharded tensors during the pass `propagateShardingsAndSetAllocationDomain`. 
The two passes are merged in attempt to reduce the number of passes over all expressions in the fusion.

Allocation domain is set to the tv's leaf domain. Since presegmentation passes and scheduling occur after the sharding passes, the leaf domain is identical to the rfact domain. After DID parallelization of the leaf domain is allowed the leaf and rfactor domain will not be the same.

This will avoid issues such as https://github.com/NVIDIA/Fuser/pull/2245#issuecomment-2111317055 and allow the `AllocationDomainPass` presegmentation pass on for distributed matmul tests


wujingyue (commented):
Is it safe now to replace https://github.com/NVIDIA/Fuser/pull/2245/files#diff-db5ba7cef14ad9a3c1eaab113a6f0a6f875e92890b34078f8185e0970022ce45R298 with a check for contiguity? Although there's little performance penalty to call `.contiguous` on an already contiguous tensor, leaving it there would give readers the wrong impression that the input slices can be non-contiguous, which IIRC shouldn't happen with this PR.

cowanmeg:
> Is it safe now to replace https://github.com/NVIDIA/Fuser/pull/2245/files#diff-db5ba7cef14ad9a3c1eaab113a6f0a6f875e92890b34078f8185e0970022ce45R298 with a check for contiguity? Although there's little performance penalty to call `.contiguous` on an already contiguous tensor, leaving it there would give readers the wrong impression that the input slices can be non-contiguous, which IIRC shouldn't happen with this PR.

Yup the added `PipelineTwoStage` tests pass when removing the `contiguous` call, but 'PipelineTestStagedReduction.StagedReduction/ReductionOnly' is failing. I would hold off on reviewing until I fix!

csrc/multidevice/utils.cpp
```diff
@@ -317,30 +317,52 @@ void insertReshardingsAfter(Fusion* fusion) {
...
+  // Set global input's allocation domain.
+  // Currently we only propagate shardings from consumer to producer,
+  // so a DeviceMesh is required.
+  for (auto global_input_tv :
+       ir_utils::filterByType<TensorView>(fusion->inputs())) {
+    NVF_ERROR(
+        global_input_tv->hasDeviceMesh(),
+        "Global inputs must be assigned a DeviceMesh ",
+        global_input_tv->toString());
+    setShardedAllocationDomain(global_input_tv);
```
wujingyue, line 338: Can you remind me of the purpose of this PR? IIRC, the problem was that certain resharding operations require some specific allocation order, e.g. the DID dimensions need to be put up front and have to be contiguous, etc. But the changes in this function are a bit overkilling? For example, I don't yet understand why the complete fusion's inputs need to have the same allocation domain as the leaf domain, if they are not resharded immediately.

csrc/multidevice/executor.cpp
```diff
@@ -50,7 +50,7 @@ MultiDeviceExecutor::MultiDeviceExecutor(
     Communicator& comm,
     MultiDeviceExecutorParams params)
     : comm_(comm), params_(params) {
-  propagateShardings(fusion.get());
+  propagateShardingsAndSetAllocationDomain(fusion.get());
```
jjsjann123, line 53: Can we put a comment here saying that `propagateShardingsAndSetAllocationDomain needs to happen before `PreSegmenter` optimization passes? I think it's useful to have that as a note when we later decided to merge the code path.

I'm not sure if that applies to `insertReshardings` and `insertShardedAxisReordering` as well.

csrc/multidevice/utils.cpp
```diff
@@ -317,30 +317,52 @@ void insertReshardingsAfter(Fusion* fusion) {
   }
 }
 
+void setShardedAllocationDomain(TensorView* tv) {
+  if (!tv->hasAllocation()) {
+    tv->setAllocationDomain(tv->getLeafDomain(), tv->getContiguity());
```
jjsjann123, line 322:
```suggestion
    tv->setAllocationDomain(tv->getLeafDomain(), true);
```

tests/cpp/test_multidevice_matmul.cpp
```diff
@@ -37,7 +37,7 @@ namespace nvfuser {
 class DistributedMatmulTest : public MultiDeviceTest {
  protected:
   DistributedMatmulTest()
-      : num_devices_(communicator->size()), optimization_guard_(false) {
```
jjsjann123, line 40: :bow:

jjsjann123: Can you also remove the header include above with `preseg_passes`

csrc/multidevice/utils.cpp
```diff
@@ -317,30 +317,52 @@ void insertReshardingsAfter(Fusion* fusion) {
...
 
-void propagateShardings(Fusion* fusion) {
+void propagateShardingsAndSetAllocationDomain(Fusion* fusion) {
+  // Set global input's allocation domain.
+  // Currently we only propagate shardings from consumer to producer,
+  // so a DeviceMesh is required.
+  for (auto global_input_tv :
+       ir_utils::filterByType<TensorView>(fusion->inputs())) {
+    NVF_ERROR(
+        global_input_tv->hasDeviceMesh(),
```
jjsjann123, line 335: This is just a requirement that we choose to enforce, right? so probably needs to be NVF_CHECK instead.

csrc/multidevice/utils.cpp
jjsjann123: FYI: I don't think we need to set this on inputs, *currently*, propagation only goes from inputs to outputs. but that might change in the future where we would set intermediates as we see fit.
Nevertheless, explicitly marking input might not be necessary, since nobody is supposed to change this.

jjsjann123 (commented):
some nitpick, but overall looks straightforward to me.

I'll let @wujingyue stamp since he seems to have more questions. (feel free to delegate it to me for a more thorough review if you don't have time to wrap it up before your vacation :tropical_drink: )

jjsjann123:
!build

jjsjann123:
FYI, be extra careful with CI for PRs touching `allocation domain`. Some of our schedulers have some sharp corners on this and could give you some unexpected failures.

csrc/multidevice/executor.cpp
cowanmeg: So far applies to all of the (re)sharding passes!

csrc/multidevice/utils.cpp
cowanmeg: Oops, my bad I didn't need to set _all_ the sharded tensor's allocation. It has been updated to just the one's around resharding expressions!

cowanmeg:
Sorry everyone who viewed before the big changes 😓 
The new update only sets the allocation domain of TVs that are in a resharding expression. This fixed most of the tests, but one DistributedMatmulTest.
There are some big changes to contiguity that this doesn't address, namely that DID axes continue to have true/false settings while they shouldn't have a value since they aren't allocated. I'll fix that in a later PR! 
Also, this isn't blocking anything so feel free to review after you get back @wujingyue

tests/cpp/test_multidevice_matmul.cpp
```diff
@@ -109,6 +101,11 @@ TEST_F(DistributedMatmulTest, LayoutTN_NoComms) {
   }
   b->setDeviceMesh(mesh);
 
+  // TODO: If c's allocation domain isn't set, it will fail validation at
+  // csrc/device_lower/validation.cpp:419, Vectorized dim for consumer has to be
+  // from a contiguous inner most position.
+  c->setAllocationDomain(c->getLeafDomain(), true);
```
cowanmeg, line 107: I had to set the allocation domain for this TV so that an error about contiguity wasn't raised. I thought the allocation inferring pass would handle this case, since this test has no communication just sharded computation.

tests/cpp/test_multidevice_sharding.cpp
```diff
@@ -33,8 +33,9 @@ TEST_P(MultideviceShardingTest, UnshardedGlobalInput) {
   input_size[sharded_dim] = num_devices;
   input_size[sharded_output_dim] = num_devices;
 
-  TensorView* tv0 = creates_concrete_tensor ? makeConcreteTensor(input_size)
-                                            : makeSymbolicTensor(4);
+  TensorView* tv0 = creates_concrete_tensor
+      ? makeContigConcreteTensor(input_size)
+      : makeContigTensor(4);
```
cowanmeg, line 38: Since these global inputs are inputs to resharding expressions they should have been made contig from the start.

csrc/multidevice/utils.cpp
```diff
@@ -325,15 +331,15 @@ void propagateShardings(Fusion* fusion) {
     auto outputs = ir_utils::filterByType<TensorView>(expr->outputs());
     TensorView* input_with_mesh = nullptr;
     for (auto tv : inputs) {
-      if (tv->hasDeviceMesh()) {
+      NVF_CHECK(
```
cowanmeg, line 334: (Not related to this PR) cleaning up the check for propagate shardings since shardings are propagated from input to output, this just checks inputs have/don't have a sharding (by checking the DeviceMesh is set)

cowanmeg:
!build --dist

cowanmeg:
!build

csrc/multidevice/utils.cpp
```diff
@@ -375,7 +381,8 @@ void insertShardedAxisReordering(Fusion* fusion) {
     auto [shard_additions, shard_deletions] = getShardingChanges(expr);
     NVF_ERROR(
         shard_additions.size() + shard_deletions.size() <= 1,
-        "Resharding expr can only support one axis")
+        "Resharding expr can only support one axis ",
```
wujingyue, line 384:
```suggestion
        "Resharding expr can only support one axis: ",
```

csrc/multidevice/utils.cpp
```diff
@@ -464,6 +471,27 @@ void insertShardedAxisReordering(Fusion* fusion) {
   }
 }
 
+void setShardedAllocationDomain(Fusion* fusion) {
+  for (auto expr : fusion->exprs()) {
+    if (isResharding(expr)) {
```
wujingyue, line 476:
```suggestion
    if (isResharding(expr)) {
      continue;
    }
```
to reduce indentation.

csrc/multidevice/utils.cpp
```diff
@@ -464,6 +471,27 @@ void insertShardedAxisReordering(Fusion* fusion) {
   }
 }
 
+void setShardedAllocationDomain(Fusion* fusion) {
+  for (auto expr : fusion->exprs()) {
```
wujingyue, line 475:
```suggestion
  for (Expr* expr : fusion->exprs()) {
```

csrc/multidevice/utils.cpp
```diff
@@ -325,15 +331,15 @@ void propagateShardings(Fusion* fusion) {
     auto outputs = ir_utils::filterByType<TensorView>(expr->outputs());
     TensorView* input_with_mesh = nullptr;
     for (auto tv : inputs) {
```
wujingyue, line 333:
```suggestion
    for (TensorView* tv : inputs) {
```

csrc/multidevice/utils.cpp
```diff
@@ -464,6 +471,27 @@ void insertShardedAxisReordering(Fusion* fusion) {
...
 
+void setShardedAllocationDomain(Fusion* fusion) {
+  for (auto expr : fusion->exprs()) {
+    if (isResharding(expr)) {
+      for (auto tv : ir_utils::filterByType<TensorView>(expr->inputs())) {
+        for (auto c : tv->getContiguity()) {
+          if (c.has_value()) {
+            NVF_CHECK(
+                c.value(),
+                "Resharding expression input must be contiguous ",
```
wujingyue, line 482:
```suggestion
                "Resharding expression input must be contiguous: ",
```

csrc/multidevice/utils.cpp
```diff
@@ -464,6 +471,27 @@ void insertShardedAxisReordering(Fusion* fusion) {
   }
 }
 
+void setShardedAllocationDomain(Fusion* fusion) {
+  for (auto expr : fusion->exprs()) {
+    if (isResharding(expr)) {
+      for (auto tv : ir_utils::filterByType<TensorView>(expr->inputs())) {
+        for (auto c : tv->getContiguity()) {
+          if (c.has_value()) {
+            NVF_CHECK(
```
wujingyue, line 480: I'm confused. I thought setShardedAllocationDomain ought to make inputs of resharding exprs contiguous rather than expect them to be contiguous. Am I missing something?

tests/cpp/test_sharding.cpp
```diff
@@ -62,6 +64,41 @@ TEST_F(ShardingTest, PropagateSharding) {
...
+  insertReshardings(&fusion);
+  insertShardedAxisReordering(&fusion);
+  setShardedAllocationDomain(&fusion);
+  for (auto expr : fusion.exprs()) {
+    if (isResharding(expr)) {
+      for (auto tv : ir_utils::filterByType<TensorView>(expr->inputs())) {
+        EXPECT_TRUE(tv->hasAllocation());
+      }
+      for (auto tv : ir_utils::filterByType<TensorView>(expr->outputs())) {
+        EXPECT_TRUE(tv->hasAllocation());
```
wujingyue, line 96: Also check for getContiguity==true?

jjsjann123 (approved):
Looks pretty clean to me now. stamping.

tests/cpp/test_sharding.cpp
```diff
@@ -62,6 +64,41 @@ TEST_F(ShardingTest, PropagateSharding) {
...
+  c->axis(1)->parallelize(ParallelType::DIDx);
+  fusion.addInput(a);
+  fusion.addInput(b);
+  fusion.addOutput(d);
+
+  propagateShardings(&fusion);
+  insertReshardings(&fusion);
+  insertShardedAxisReordering(&fusion);
+  setShardedAllocationDomain(&fusion);
+  for (auto expr : fusion.exprs()) {
```
wujingyue, line 90: How do you like making the check more specific? For example, is it possible to check the following:
* I'd expect there's only one resharding Expr, which is a sum
* I'd also expect the input of that Expr has DID as the first IterDomain in the containing allocation domain.

tests/cpp/test_multidevice_matmul.cpp
jjsjann123: This is interesting.... Likely the allocation order inference is getting some strange mapping on output.

I opened #2276 and will investigate it.

csrc/multidevice/utils.cpp
jjsjann123: ^^^ that doesn't sound right. You cannot *change* contiguity / stride order on inputs.

IIUC, the code here validates that the input entry is contiguous and then later explicitly sets allocation domain on each if they are implicit.

One part I'm not totally sure is, `Resharding expression input must be contiguous `. Should this check also apply to TensorViews which we are not specifying allocation domain?

wujingyue:
> I'll let @wujingyue stamp since he seems to have more questions. (feel free to delegate it to me for a more thorough review if you don't have time to wrap it up before your vacation 🍹 )

I won't have time to take another look before my vacation. As long as you and @jjsjann123 are confident in the change, please merge it without me. Anyhow, this is a strict improvement and my remaining comments are about potentially making allocation domains around resharding more correct.

csrc/multidevice/utils.cpp
samnordmann: ProcessGroup only accept [contiguous tensors](https://github.com/pytorch/pytorch/blob/a00a99e801c14d2b09533e09a9c20a1af473d9ab/torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp#L2168) (because nccl and ucc only deal with contiguous buffers, passed as void pointers). So for now it is reasonable to only support contiguous tensors. Later, we could add support for non-contiguous tensors, but to do that we'll have no choice but to make as many process group call as there a contiguous components.

So imo this "assert" makes sense for this pr and we could remove it later by implementing what I described above

@wujingyue this line
https://github.com/NVIDIA/Fuser/blob/690134dbf2403546978631653ea0f5bfabb352d3/csrc/multidevice/utils.cpp#L393-L395

is a "trick" to allow non-oputermost resharding while avoiding non-contiguous buffers, by reordering the axis to place the sharded axis at outermost position (therefore the actual buffer is contiguous in memory)

tests/cpp/test_multidevice_sharding.cpp
samnordmann: right! well spotted

samnordmann (approved):
Looks good to me! Thanks!

csrc/multidevice/utils.cpp
```diff
@@ -464,6 +471,27 @@ void insertShardedAxisReordering(Fusion* fusion) {
...
+                c.value(),
+                "Resharding expression input must be contiguous ",
+                expr);
+          }
+        }
+        setShardedAllocationDomain(tv);
```
naoyam, line 486: Can somebody please explain why this is necessary? IIUC, we confirm this tensor is contiguous. Isn't that sufficient?

naoyam: So, this is not necessary if the allocation order inference is not done?

csrc/multidevice/utils.cpp
naoyam: I assume Jingyue meant to say:

```suggestion
    if (!isResharding(expr)) {
       continue;
     }
```

csrc/multidevice/utils.cpp
jjsjann123: I think so.

For the record, it's not just allocation order inference, alias passes could also update allocation domain introducing similar issue. See Jingyue's comment in https://github.com/NVIDIA/Fuser/pull/2245#issuecomment-2111317055

cc'ing @cowanmeg for sanity check

cowanmeg:
!build

cowanmeg:
!build

cowanmeg:
I see an unrelated tolerance error in `IndexingOpTest.TorchGatherSumAdd_CUDA`, so will merge this.

PR #2272, opened 2024-05-20
Generalize CombineMulSum as MatmulPatterns

jacobhinkle:
This replaces the `CombineMulSum` class with `MatmulPattern` in the Matmul scheduler. Additionally, we use these matmul patterns to determine the problem layout, IterDomain roles, and TensorView roles. The allocation domain is used to determine the problem layout. The matmul scheduler is updated to reject segments whose input allocation domains are non-trivial (until that is supported eg. by #2226).

Note that this does not add handling of `MatmulOp` and `LinearOp` in the matmul scheduler. That will be done next in #2236 or similar.


jacobhinkle:
!build

csrc/ir/utils.cpp
```diff
@@ -1099,6 +1099,31 @@ int64_t getVectorizeSize(const TensorView* tv) {
   return 1;
 }
 
+bool hasTrivialAllocationDomain(const TensorView* tv) {
```
jacobhinkle, line 1102: The intention of this utility is to generalize `!tv->hasAllocation()` to cases where an allocation domain is provided, but it actually corresponds to the no-reductions rfactor domain (ignoring broadcasts).

csrc/ir/utils.cpp
```diff
@@ -1265,7 +1290,7 @@ MmaOpDetails getMmaOpDetails(
   const auto validateOutputDetails = [](const TensorViewDetails& details,
                                         const std::string& desc) {
     // TODO: revise rules when add support for batch gemms
-    NVF_ERROR(details.bcasts.empty(), desc, ": has broadcast domains.");
+    // NVF_ERROR(details.bcasts.empty(), desc, ": has broadcast domains.");
```
jacobhinkle: Fixes #2273. See test `AmpereMulSumToMatmul_MultipleBroadcasts`

jacobhinkle:
!build

jacobhinkle:
!build

csrc/scheduler/mma_utils.cpp
```diff
@@ -1210,143 +1210,122 @@ RolesMapOpt getTensorsRoles(
...
-        return (a->uses().size() == b->uses().size())
-            ? (a->name() < b->name())
-            : (a->uses().size() > b->uses().size());
-      });
+    // NOTE: depending on fusion definition k domain may appear in the output:
+    //  - for mma_output == fusion output k domain is present
+    //  - for mma_output != fusion output (fusion with epilogue) k domain
+    //    is not present
+    if (has_k || has_unmapped) {
+      // Don't map TVs to output roles if they have unmapped dims, or if they
```
Priya2698, line 1282: Why is a tensor with K dim also skipped?

csrc/scheduler/mma_utils.cpp
```diff
@@ -1210,143 +1210,122 @@ RolesMapOpt getTensorsRoles(
...
-      const auto end = domains.end();
-
-      bool has_m = (end != std::find(begin, end, MatmulDomain::M));
-      bool has_n = (end != std::find(begin, end, MatmulDomain::N));
-
-      // NOTE: depending on fusion definition k domain may appear in the output:
-      //  - for mma_output == fusion output k domain is present
-      //  - for mma_output != fusion output (fusion with epilogue) k domain
-      //    is not present
+    // NOTE: the core fusion output tensors are the ones with m and n
```
Priya2698, line 1287: What do you mean by core fusion output tensors?

csrc/scheduler/mma_utils.h
```diff
@@ -242,34 +244,41 @@ class DataWrapperOpt {
...
-  MulAndSumOps mulsumops;
-  InputsOutputs insouts;
-  Broadcasts bcasts;
+//! This represents a single matmul operation, without a prologue or epilogue.
+//! Each matmul has two inputs which might not be fusion inputs: A and B. It
+//! also has one output, which can be Float or reduced precision. For MatmulOp
+//! and LinearOp, the output is the same dtype as the inputs; so output does not
+//! necessarily correspond to the output of a translated MmaOp and it might not
+//! be a fusion output.
+struct MatmulPattern {
```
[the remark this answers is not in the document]
jacobhinkle: In the next PR it will include `MatmulOp` and `LinearOp` and perform that translation, so it really is meant to generalize those.

csrc/scheduler/mma_utils.h
```diff
@@ -290,13 +299,19 @@ using DependenciesMap = std::map<TensorView*, DomainsDesc>;
 //!  transposition of inputs in mma instructions, while other (e.g. Turing,
 //!  Ampere) the only supported transposition is TN which means that mma
 //!  instruction first input is transposed, the second input is non-transposed.
-NVF_API MatmulProblemLayoutOpt getMmaLayout(
-    Fusion* fusion,
-    const mma_utils::MulSumProperties::InputsOutputs& props);
+NVF_API MatmulProblemLayoutOpt
```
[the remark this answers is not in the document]
jacobhinkle: Good question. It would need to be so if used in C++ tests. I will check and reduce the surface area as much as possible.

csrc/scheduler/mma_utils.cpp
```diff
@@ -1437,103 +1416,209 @@ TensorView* getTensorviewPriorToCast(TensorView* in) {
...
+          has_k = true;
+        }
+      }
+
+      // Remember that we are just gathering the immediate inputs to the
+      // matmul, so there should be no prologue between a, b and the mul/sum.
+
+      // Check that the inputs have broadcasts that are not all in common, i.e.
+      // that there is at least one M and at least one N dimension.
+      TensorView* ltv = getTensorviewPriorToCast(bop->lhs()->as<TensorView>());
```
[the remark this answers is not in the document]
jacobhinkle: Will do. BTW it's not just implicit upcasting from mul. If we have half precision inputs and a chain of operations in Float, then we perform this multiplication, if the segmenter segments just before the mul then it could insert casts at the segmentation edge (once we move past #1889).

csrc/scheduler/mma_utils.cpp
```diff
@@ -1437,103 +1416,209 @@ TensorView* getTensorviewPriorToCast(TensorView* in) {
...
+  id_model.maybeBuildGraph(IdMappingMode::EXACT);
+  const ValGraph& exact_graph = id_model.idGraph(IdMappingMode::EXACT);
+
+  // There are four types of ValGroup involved in a MatmulPattern: M, N, K, and
+  // Batch. These are enumerated in the MatmulDomain enum class. They are
+  // defined by their membership as follows:
+  //   M: present in A and output, but not B
+  //   N: present in B and output, but not A
+  //   K: present in A and B, but not output
+  //   Batch: present in all A, B, and Batch
```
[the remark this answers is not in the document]
jacobhinkle: Good call on the change, that's just a typo; I think I wrote that line right after adding Batch to M, N, K in a few places.

Regarding different numbers of batch dims; I think we should consider what is a batch dim vs an M or N dim. If a dimension exists in the output and in B but not in A, then I claim it is an N dimension: we could add a broadcast dim to A corresponding to that dim and obtain the same result, right?  Similar for M dimensions. In fact that is probably how we'll translate those cases for `MatmulOp` and `LinearOp`. So I think the only _true_ batch dims are concrete and exist in all of A, B, and output. There is a trivial matter of Broadcast batch dims, i.e. when a dim is broadcast in A, B, and output but we could handle that by translating to squeeze+bcast.

csrc/scheduler/mma_utils.cpp
```diff
@@ -1051,14 +1054,13 @@ inline void resolveTvToMatmulDomainsMapping(
 
 } // anonymous namespace
 
-ProblemIterDomainsOpt getProblemIterDomains(
-    const mma_utils::MulSumProperties::InputsOutputs& props) {
+ProblemIterDomainsOpt getProblemIterDomains(const MatmulPattern& pattern) {
```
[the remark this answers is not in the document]
jacobhinkle: We do not need `getProblemIterDomains`. It should be obsolete. I just didn't finish removing it yet. I'll work on it.

csrc/scheduler/mma_utils.cpp
```diff
@@ -1100,105 +1102,103 @@ ProblemIterDomainsOpt getProblemIterDomains(Fusion* fusion) {
...
+  // using DataWrapperOpt<MatmulDomain>(std::move(dom)) leads to a clang-tidy
+  // warning because MatmulDomain is trivially movable. There is only a move
+  // constructor for DataWrapperOpt to prevent inadvertent copying. To avoid
+  // this complication I'm using a simple pair for the lambda's result type.
+  using InnerDomResult = std::pair<MatmulDomain, std::string>;
+  const auto innerDomain = [&roles_map, &group_to_domain, &exact_graph](
+                               MatmulRole role) -> InnerDomResult {
+    const auto role_it = roles_map.find(role);
+    if (role_it == roles_map.end()) {
+      return {MatmulDomain::M, "Could not find role in roles_map"};
```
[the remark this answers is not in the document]
jacobhinkle: You're right and this `InnerDomResult` business is a hack to get around the design of `DataWrapperOpt`. I want to just return the error message in this case, but using `DataWrapperOpt` doesn't work properly in clang (haven't tried gcc) when the wrapped type is trivially copyable, since it balks at using `std::move` on such types.

csrc/scheduler/mma_utils.cpp
jacobhinkle: It is skipped for the purposes of being an output dimension. There can still be output tensors without roles possibly, but they won't be identified as the roles in `MatmulRole`. Specifically, if an _output_ tensor has a K dimension then it is downstream of at least one operand via a path other than the matmul, so it's "extra-curricular" computation unrelated to the matmul.

csrc/scheduler/mma_utils.cpp
jacobhinkle: That comment is left over from the previous implementation. It means that outputs we identify as D must have M and N. Others can possibly exist but are not given roles. Note that we can't currently reduce dimensions in the epilogue but that could be possible some day and we should revisit the notion of output roles at that point (see #2213).

jacobhinkle:
!build

csrc/scheduler/mma_utils.h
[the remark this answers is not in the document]
jacobhinkle: It is used in tests.

jacobhinkle:
!build

csrc/scheduler/matmul_utils.cpp
```diff
@@ -232,12 +229,6 @@ std::string isMatmulFusionDefinitionSupported(
...
   // Fusion topology check
   {
-    const auto& roles_map_opt = mma_utils::getTensorsRoles(fusion, props);
-    if (!roles_map_opt.isValid()) {
-      return roles_map_opt.getErrorMsg();
-    }
-
-    const auto& roles_map = roles_map_opt.getData();
     auto entry = roles_map.find(MatmulRole::INPUT_A);
     std::set<TensorView*> tvs_with_roles;
```
Priya2698, line 233: I know this is not added by this PR, but it doesn't look like we need the actual tensorviews, just the total count.

csrc/scheduler/mma_utils.h
```diff
@@ -290,13 +299,19 @@ using DependenciesMap = std::map<TensorView*, DomainsDesc>;
...
-//! the mma_utils::MulSumProperties::InputsOutputs is extracted from the fusion.
-NVF_API MatmulProblemLayoutOpt getMmaLayout(Fusion* fusion);
+//! the MatmulPattern is extracted from the fusion.
+NVF_API MatmulProblemLayoutOpt getProblemLayout(Fusion* fusion);
+
+//! Determine the problem layout based on allocation domain of inputs. This is
+//! called by the above overloads.
+NVF_API MatmulProblemLayoutOpt getProblemLayout(
+    const IdModel& id_model,
+    const std::unordered_map<ValGroup, MatmulDomain>& group_to_domain,
```
zasdfgbnm, line 313: is `dim_roles` a better name?

zasdfgbnm:
Could you rebase this PR? I see obsolete code removed by https://github.com/NVIDIA/Fuser/pull/2268 in this PR.

csrc/scheduler/mma_utils.cpp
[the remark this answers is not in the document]
zasdfgbnm: This can be removed now.

csrc/scheduler/matmul_utils.cpp
jacobhinkle: The count of the unique tensorview is needed I think. I think this code might be for handling cases where a tensor has multiple roles, for example `A@A`. In that case count is not sufficient to determine whether there are TVs without roles.

csrc/scheduler/mma_utils.h
jacobhinkle: Much better name. I'll make the change.

csrc/scheduler/mma_utils.cpp
```diff
@@ -1210,143 +1162,122 @@ RolesMapOpt getTensorsRoles(
...
+      storage.push_back(tv);
+    }
+  }
 
-      // NOTE: the core fusion output tensors are the ones with m and n
-      //  domains
-      if (has_m && has_n) {
-        storage.push_back(entry.first);
-      }
+  // NOTE: sort output roles in descending order by uses() size, and
```
Priya2698, line 1246: Why is sorting the tvs important -- is there a place where we rely on this ordering to be same everywhere?

csrc/scheduler/mma_utils.h
```diff
@@ -242,34 +244,41 @@ class DataWrapperOpt {
...
+  // representing a matmul in the current Fusion. The definition of this tensor
+  // determines what kind of translation is needed, if any. Possible definition
+  // Expr types are: MmaOp, ReductionOp (for mul-sum patterns), MatmulOp, and
+  // LinearOp.
+  TensorView* output;
+
+  //! If the pattern is not already represented by an MmaOp, for example if
+  //! there is a MatmulOp instead, this function modifies the fusion to insert
+  //! an MmaOp. TensorViews A and B are unchanged, but this->output might be
+  //! updated to reflect the replacement tensor.
```
zasdfgbnm, line 266: > but this->output might be updated to reflect the replacement tensor.

Is it? I don't see it in the definition of `translateToMmaOp`.

csrc/scheduler/mma_utils.cpp
```diff
@@ -1437,103 +1368,195 @@ TensorView* getTensorviewPriorToCast(TensorView* in) {
...
+      const std::vector<IterDomain*>& red_root =
+          rop->out()->as<TensorView>()->getRootDomain();
+      NVF_ERROR(red_root.size() == lrf.size());
+      bool has_m = false, has_n = false;
+      for (size_t i : c10::irange(lrf.size())) {
+        if (lrf[i]->isBroadcast() && !rrf[i]->isBroadcast()) {
+          has_m = true;
+        } else if (!lrf[i]->isBroadcast() && rrf[i]->isBroadcast()) {
+          has_n = true;
+        }
```
zasdfgbnm, line 1450: Looks like this is not worse that what is already in main, but I don't think it makes sense to assume lhs is A and rhs is B. It can be the opposite. For now, can we assume the output's rfactor domain is in the order of `m,n,k`, and use this to infer which is A and which is B here? We can review this piece of code when we support matmul-transpose fusion so that we do not assume output order `m,n,k`.

csrc/scheduler/mma_utils.cpp
```diff
@@ -1051,154 +1054,103 @@ inline void resolveTvToMatmulDomainsMapping(
...
+  const InnerDomResult b_dom_res = innerDomain(MatmulRole::INPUT_B);
+  if (!b_dom_res.second.empty()) {
+    std::string err = b_dom_res.second;
+    return err;
   }
-  if ((mk_found && nk_found) && !(km_found || kn_found)) {
+  const bool kinner_b = b_dom_res.first == MatmulDomain::K;
+
+  if (kinner_a && kinner_b) {
     return MmaLayout::TN;
```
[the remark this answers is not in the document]
jacobhinkle: Yes that would be much clearer I'd be happy to not see "TN" appear anywhere in the codebase :-D

csrc/scheduler/mma_utils.cpp
```diff
@@ -1051,154 +1054,103 @@ inline void resolveTvToMatmulDomainsMapping(
...
 }
 
-MatmulProblemLayoutOpt getMmaLayout(Fusion* fusion) {
-  auto mma_exprs = ir_utils::getOpsOfType<MmaOp>(fusion);
-  if (mma_exprs.size() != 1) {
-    std::stringstream ss;
-    ss << "Invalid number of MmaOp instances in fusion, expected 1, got "
-       << mma_exprs.size();
-    return ss.str();
+MatmulProblemLayoutOpt getProblemLayout(
```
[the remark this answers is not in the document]
jacobhinkle: Good catch! I refactored it as you suggest and removed the `getProblemLayout(Fusion*, const MatmulPattern&)` signature. I don't love the `getProblemLayout(Fusion*)` signature either since we should encourage reusing data when possible instead of rebuilding the entire `IdModel`. I think that signature is only used in tests but it seems to be useful for now so I kept it.

csrc/scheduler/mma_utils.cpp
```diff
@@ -1210,143 +1162,122 @@ RolesMapOpt getTensorsRoles(
...
+      auto it = dim_roles.find(g);
+      if (it == dim_roles.end()) {
+        // tv has an unmapped non-broadcast and non-reduction dimension
+        has_unmapped = true;
         continue;
       }
-      // Bias vectors are assigned to INPUT_C role
-      if (!has_k) {
-        roles_map[MatmulRole::INPUT_C].push_back(entry.first);
+      has_m |= it->second == MatmulDomain::M;
```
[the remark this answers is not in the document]
jacobhinkle: That's correct. Thanks for the suggestion.

csrc/scheduler/mma_utils.cpp
jacobhinkle: Exactly. We sometimes iterate over roles TVs. If we did not sort, then the `->name()`s of introduced `Val`s would be ordered arbitrarily, for example. Maintaining deterministic compiled code is helpful for the codediff tool, and also for keeping our sanity when debugging problem fusions.

csrc/scheduler/mma_utils.h
jacobhinkle: This would be the case for `MatmulOp` which ends with a downcast. If all uses of `output` are upcasts back to `Float` then we can remove the casts. In that case we would update output to point to the Float output since the `Half` output is no longer alive in the fusion.

csrc/scheduler/mma_utils.cpp
jacobhinkle: I agree. Ideally we would use the output allocation domain and IdModel to determine which is A and which is B for this pattern. When I wrote this pass I was not using those yet, but I could plumb them in. Since this takes place before translation, then we would need to rebuild the IdModel after translation.

For now, as you said I will assume the rfactor is the allocation domain for determining A and B, and I'll place a note there that this should be revisited.

jacobhinkle: I just pushed a change that looks at the reduction output root domain to determine N which lets us determine which input is B and which is A.

csrc/scheduler/mma_utils.cpp
```diff
@@ -1437,103 +1368,195 @@ TensorView* getTensorviewPriorToCast(TensorView* in) {
...
+  //   K: present in A and B, but not output
+  //   Batch: present in all A, B, and output
+  // If there are other membership patterns, for example a ValGroup present in
+  // only A, then we should raise an exception here.
+
+  // Indicates whether a ValGroup is present in A (bit 0), B (bit 1), or output
+  // (bit 2)
+  using ValGroupPresence = std::bitset<3>;
+
+  std::unordered_map<ValGroup, ValGroupPresence> present_flags;
```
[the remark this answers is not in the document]
jacobhinkle: Actually I just renamed `ValGroupPresence` to `DimPresence` and updated the comment to no longer mention "membership", since I think that's a little more opaque term than "presence". What we really care about is whether a dimension is present in each tensor so I think that term is clearer.

csrc/scheduler/mma_utils.cpp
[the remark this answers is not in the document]
jacobhinkle: As an intermediate step, I'm now using `UnitDim` in this function now, but I'm still returning `MatmulProblemLayoutOpt`.

csrc/scheduler/mma_utils.cpp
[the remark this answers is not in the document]
jacobhinkle: I just pushed a change that uses a bare `variant<std::string, UnitDim>`.

csrc/ir/utils.cpp
protonu: Thanks! Would it be possible to add this as a comment in the header file.

csrc/ir/utils.cpp
```diff
@@ -1103,6 +1103,31 @@ int64_t getVectorizeSize(const TensorView* tv) {
   return 1;
 }
 
+bool hasTrivialAllocationDomain(const TensorView* tv) {
+  if (!tv->hasAllocation()) {
+    return true;
+  }
+  const std::vector<IterDomain*>& alloc = tv->getMaybeAllocationDomain();
+  const std::vector<IterDomain*>& rf = tv->getMaybeRFactorDomain();
```
protonu, line 1111: Is it possible to get the allocation/rfactor domains, then remove the broadcasts and reductions.
Then we can check if the filter allocation domain and the RFactor domains are permutations or not?

jacobhinkle: In this case, we do not just need it to be a permutation, but actually identical. That's why I am checking sameAs in the loop below while removing broadcast and reduction domains. I could potentially use `TensorDomain::noReductions()` and `TensorDomain::noBroadcasts()` first, but I would still need to loop over the two vectors to call `sameAs`, so I just avoided building the temporaries and fused those loops here.

protonu: In this case do we need to call `sameAs` or will checking equality of pointers be enough?

protonu: If you're not against temporaries, we can remove the loop by first filtering with `noReductions` and `noBroadcasts`, then comparing the filter vectors with std::equal, where `sameAs` can be used as part of lambda.

jacobhinkle: `sameAs` is probably not needed; after all we commonly use `std::is_permutation` to check that allocation domain is just a stride ordering. I'll change to use the helper functions (with temporaries) because I think it'll make it clearer.

zasdfgbnm (approved):

csrc/scheduler/matmul_utils.cpp
```diff
@@ -533,16 +552,21 @@ std::shared_ptr<MatmulParams> getMatmulHeuristics(
...
+  NVF_ERROR(!patterns.empty(), "No matmul patterns were found");
   NVF_ERROR(
-      combiner.isValid(),
-      "There's no (single) mma op or mul-sum op which mma op can replace")
+      patterns.size() == 1,
+      "Only a single matmul pattern can currently be fused");
+  mma_utils::MatmulPattern& pattern = patterns.front();
+
+  // IdModel is used to analyze problem shape & layout
+  IdModel id_model(fusion);
```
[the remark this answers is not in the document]
jacobhinkle: That's a good idea. Currently I build an IdModel in `getMatmulHeuristics` then again in `scheduleMatmul`, and actually I also need to rebuild it after translating `MatmulPattern`s to `MmaOp`s. I think this use case is very similar to the `DomainMap` in the pointwise scheduler that you linked to; we just use `IdModel` instead of `ComputeAtMap`.

Issue #2146, opened 2024-04-24
LayerNorm fusion in Diffusion Transformer has worse performance than expected


parthmannan:
nvFuser generated code for a fusion block present in DiT has worse than expected performance. The subgraph is performing a `LayerNorm + + Mul + Add + Add` computation as shown in the code below. nvFuser code was generated using Lightning Thunder compiler for PyTorch.

```
Thunder Forward performance: 99us
Thunder Backward performance: 1034us

Torch.Compile Forward performance: 115us
Torch.Compile Backward performance: 457us
```

Below is the subgraph highlighted.

![Screenshot 2024-03-20 at 1 38 06 PM](https://github.com/NVIDIA/Fuser/assets/38387286/159a35dd-06e6-440e-a708-02e59d8f0f63)

Reproducible script (requires Thunder installed)
```
import torch
import torch.nn as nn
import torch.nn.functional as F
import thunder

class Net(nn.Module):
    def __init__(
        self,
        config):
        super(Net, self).__init__()
        self.ln = nn.LayerNorm(config['hidden_units'])
        
    def modulate(self, x, scale, shift):
        return x * (1 + scale.unsqueeze(1)) + shift.unsqueeze(1)

    def forward(self, x, scale, shift):
        x   = self.ln(x)
        out = self.modulate(x, scale, shift)
        return out

config = {'hidden_units': 1024, 'seq_len': 1024, 'batch_size': 56}

net = Net(config)
net.cuda()
net.to(dtype=torch.bfloat16)

from thunder.executors.sdpaex import sdpa_ex
executors = [sdpa_ex, thunder.nvfuser_executor, thunder.pytorch_executor]

network_fn = thunder.jit(net, executors=executors)
#network_fn = torch.compile(net)

bench_iters = 10
profile_batch = 5

input_shapes    = [(config['batch_size'], config['seq_len'], config['hidden_units']),
                   (config['batch_size'], config['hidden_units']),
                   (config['batch_size'], config['hidden_units'])]

def generate_io_tensor(net, input_shapes):
    input_tensors = []

    for shape in input_shapes:
        tensor = torch.rand(shape, dtype=torch.bfloat16, requires_grad=True, device='cuda')
        input_tensors.append(tensor)

    target_tensor_size = net(*input_tensors).size()
    target_tensor = torch.rand(target_tensor_size, dtype=torch.bfloat16, device='cuda')

    return input_tensors, target_tensor

for idx in range(bench_iters):
[...]

PR #2274, opened 2024-05-20
start post reduction broadcast backward check from output

liqiangxl:
**Issue:** `hasNonNormalizePostReductionBCast()` checks whether the post reduction broadcast ID is mapped to a reduction input ID. Current check backwards from where the broadcast ID is resolved, e.g. T6 in the following fusion (see newly added test `PostReductionBroadcastCheck`) where `T6[I1,I2] = T4[I1,B] + T5[I1,I2]` and the broadcast ID in `T4[I1,B]` is resolved to `I2`. It backwards following `T6-->T4` and `T6->T5->T1`. Neither of these paths leads to an ID consumed by the reduction, becuase in this case the ID consumed by the reduction is in `T2[I1,I2]`.
image
**Fix:** instead of backwards from the location where the broadcast ID is resolved, we should further move forward along the usage chain and stop at the following conditions:
(1) we reached the output tensor, e.g. usage chain becomes empty.
(2) we reached a broadcast op, which introduces a new broadcast ID and needs another check.
(3) can't be mapped further, e.g. in case `TakeAlongAxisIntermediateTensorNormalization1_CUDA`
For this specific case, stop at `T7[I1,I2] = T6[I1,I2] + T2[I1,I2]`. From T7, the backward search can find the corresponding reduction input ID, which is {I2} in T2.

**Results:**
After this PR, the backward of fusion in [issue-2146](https://github.com/NVIDIA/Fuser/issues/2146) is segmented into a reduction kernel and a inner_outer_persistent kernel.
Other fixes, e.g. remove all the redundant broadcast-squeeze ops also solved this issue because it generates a fusion where the scenario descripted in this PR is avoided during the segment process.


liqiangxl:
!build

rdspring1:
Will this PR fix https://github.com/NVIDIA/Fuser/issues/2046?

liqiangxl:
> Will this PR fix #2046?

Yes. It has the same issue, as shown in the following figure, before this PR it backwards from `T40` and can't find an input to reduction due to  `B27->S25` and `B32->S26`. After this PR, it backwards from `T41` and can find `T13` which is an input to reduction.
image

wujingyue:
LGTM but defer approval to @naoyam. The problem is clear and the fix seems to make sense. This PR will let the normalization scheduler accept more patterns, which can be better (e.g. because of fewer kernel launches) or worse (e.g. because the new patterns could lead to slower one-kernel code than multiple kernels). I'm just not sure if it's a strict improvement.

csrc/scheduler/registry_utils.cpp
```diff
@@ -627,6 +627,42 @@ bool SchedulerTopologyChecker::hasNonNormalizePostReductionBCast(
           continue;
         }
 
+        // Move forward to output to avoid unnecessary segmentation when the
+        // reduction input iteration domain is used after this tensor. See
+        // PostReductionBroadcastCheck test and issue-2146.
+        while (!forward_tv_dep_chain.empty()) {
```
naoyam, line 633: This is within another while loop over `forward_tv_dep_chain`, right? Can we simplify the control flow?

liqiangxl:
!build

csrc/scheduler/registry_utils.cpp
liqiangxl: removed one while loop. The new structure is:
```
 std::vector<std::pair<IterDomain*, IterDomain*>> ids_to_resolve;
 while (!forward_tv_dep_chain.empty()) {
  // Find the id to resolve
  if (ids_to_resolve.empty()) {
    // if stll empty, move forward
    continue;
  }
  // Got at least one id to resolve, try move forward further along the use chain
  if (!forward_tv_dep_chain.empty()) {
    // if can move forward, move forward
    continue;
  }
  // backward resolve

  if (!ids_to_resolve.empty()) {
    // bad case, direct return
    return true;
  }
  // if reach here, ids_to_resolve becomes empty again.
  // start next check along the use chain
 }
```

csrc/scheduler/registry_utils.cpp
```diff
@@ -572,59 +572,96 @@ bool SchedulerTopologyChecker::hasNonNormalizePostReductionBCast(
...
-        if (ids_to_resolve.empty()) {
-          continue;
+        // At this point, we have some ids to resovle.
+        // Check if we can move forward further along the use chain.
+        // This helps gathering more usage info to resolve this id to avoid
+        // unnecessary segmentation when the reduction input iteration domain is
+        // used after this tensor. See PostReductionBroadcastCheck test and
+        // issue-2146. We can move forward when:
+        // (1) The usage chain is not empty.
+        // (2) The consumer is not a broadcastOp, as broadcastOp introduces a
```
naoyam, line 637: Could you give more details why?

csrc/scheduler/registry_utils.cpp
```diff
@@ -572,59 +572,96 @@ bool SchedulerTopologyChecker::hasNonNormalizePostReductionBCast(
...
+        if (!forward_tv_dep_chain.empty()) {
+          auto tmp_producer = forward_running_consumer;
+          auto tmp_consumer = forward_tv_dep_chain.front();
+          if (!tmp_consumer->definition()->isA<BroadcastOp>()) {
+            auto forward_pairwise_root_map =
+                PairwiseRootDomainMap(tmp_producer, tmp_consumer);
+            auto forward_p2c_root_map =
+                forward_pairwise_root_map.mapProducerToConsumer();
+            bool at_leat_one_id_mapped = false;
+            for (size_t entry_i = ids_to_resolve.size(); entry_i > 0;
```
naoyam, line 650: Is there any reason traversing from the back?

csrc/scheduler/registry_utils.cpp
```diff
@@ -572,59 +572,96 @@ bool SchedulerTopologyChecker::hasNonNormalizePostReductionBCast(
...
+              if (forward_p2c_root_map.find(running_id) !=
+                  forward_p2c_root_map.end()) {
+                at_leat_one_id_mapped = true;
+                ids_to_resolve[entry_i - 1] = std::make_pair(
+                    forward_p2c_root_map.at(running_id),
+                    forward_p2c_root_map.at(running_id));
+              }
+            }
+            if (at_leat_one_id_mapped) {
+              continue;
```
naoyam, line 662: Can't this continue even if nothing is mapped?

csrc/scheduler/registry_utils.cpp
```diff
@@ -572,59 +572,96 @@ bool SchedulerTopologyChecker::hasNonNormalizePostReductionBCast(
...
+        // unnecessary segmentation when the reduction input iteration domain is
+        // used after this tensor. See PostReductionBroadcastCheck test and
+        // issue-2146. We can move forward when:
+        // (1) The usage chain is not empty.
+        // (2) The consumer is not a broadcastOp, as broadcastOp introduces a
+        //     new broadcast axis and require a new check, see
+        //     FusionReshapePersistentShmoo.
+        // (3) At least one id is mapped, otherwise can't do backward traversal.
+        if (!forward_tv_dep_chain.empty()) {
+          auto tmp_producer = forward_running_consumer;
```
naoyam, line 642: When this line is visited with non-empty `ids_to_resolve` as part of the while loop of line 586, shouldn't it use `forward_running_consumer` and `forward_running_producer` of lines 587 and 588?

liqiangxl: Here we are trying to moveforward along the use chain. For example when we have
 `T0 (forward_running_producer)-> T1 (forward_running_consumer)  -> T2`
At this point, we checked `T0->T1` and want to check if we may moveforward to `T1->T2`, so `T1` becomes the `tmp_producer ` and `T2` becomes the `tmp_consumer`

csrc/scheduler/registry_utils.cpp
liqiangxl: we can't, otherwise there is no way to transverse back if there is no other links. For example:
image

naoyam: Thanks. Makes sense.

csrc/scheduler/registry_utils.cpp
liqiangxl: The only reason I can think of is we are removing element from the vector, so the safest way is to vist backward. I added a case with multiple bcast ids and did some experiment with forward iterating, it also works fine as long as the loop condition is correctly set, e.g. use 
```
size_t num = ids_to_resolve.size();
for (size_t entry_i = 0; entry_i < num;entry_i++)
```
instead of 
`for (size_t entry_i = 0; entry_i < ids_to_resolve.size();entry_i++)`

csrc/scheduler/registry_utils.cpp
naoyam: Maybe I'm reading the code incorrectly, but what happens if there's also another consumer following `T2`, say `T3`? The pair of `T1` and `T2` is checked here and then the control goes back to the top of the loop body of the while loop starting at line 586, right? Even though there's `T3` left, since `forward_tv_dep_chain` is empty since there's `pop_front`, `T3` is not going to be visited, right?

liqiangxl: we didn't pop front when trying to move forward, it's just a copy of the front element and see wheter we can do it.
```
          auto tmp_producer = forward_running_consumer; (which is T1)
          auto tmp_consumer = forward_tv_dep_chain.front(); (which is T2)
          T2 is NOT popped.
```
In your case, when go back to the top of the loop body of the while loop, 
```
        forward_running_producer = forward_running_consumer; (which is T1)
        forward_running_consumer = forward_tv_dep_chain.front(); (which is T2)
        T2 is popped.
```

naoyam: Oh, I see. It seems this while loop essentially consists of two separate parts. Wouldn't it make more sense to have two while loops, one is to find a resolution, i.e., line 598 to line 628, and another is to move forward further.

csrc/scheduler/registry_utils.cpp
naoyam: Where is the removal?

csrc/scheduler/registry_utils.cpp
liqiangxl: That's what I did at the beginning of this implementation. Let me roll back to the original version.

naoyam: That's not what I meant. The previous version has a nested while loop with the same loop condition. What I'm suggesting is to have two separate while loops.

liqiangxl: oh, got you. let me check how to do that.

liqiangxl: Revised. We need 3 while loops. The current logic is:
```
// check this dep chain
while (!forward_tv_dep_chain.empty()) {
  (1) step-1, Find the id to resolve
  while{}
  (2) step-2, Find the location we need to start the backward traversal resolve
  while{}
  (3) step-3, do backward traversal resolve
  (4) step-4, move to beg of the outer while loop, continue check of the dep chain
}
```

csrc/scheduler/registry_utils.cpp
liqiangxl: [Here](https://github.com/NVIDIA/Fuser/blob/ad7ea4466612cca4b2e8bf9b27e6efeb2bdaa3b1/csrc/scheduler/registry_utils.cpp#L693):
```
                    // If mapped, and producer is a producer of a reduction,
                    // we can resolve this id
                    ids_to_resolve.erase(
                        ids_to_resolve.begin() + (int64_t)entry_i - 1);
```

csrc/scheduler/registry_utils.cpp
```diff
@@ -614,20 +586,111 @@ bool SchedulerTopologyChecker::hasNonNormalizePostReductionBCast(
...
+                  forward_running_producer->getLogicalDomain().end(),
+                  [](IterDomain* id) { return id->isBroadcast(); })) {
+            // If there's no broadcast axes in producer it doesn't need to be
+            // checked
+            continue;
+          }
+
+          // If consumer is before another reduction it doesn't need to be
+          // checked
+          if (pre_reduction_tvs.count(forward_running_consumer)) {
```
naoyam, line 606: When does this happen?

naoyam: Previously, when this condition hits, it exits the while loop. In this version, it just exits the inner while loop and moves on to the step 2. What does the difference mean?

liqiangxl: > When does this happen?

when we have two consecutive reductions, e.g. softmax we have
```
|-------->------|
0 -> r1 -> b2 ->3 ->4 -> r5
```
where tensor `3` is an input to reduction `r5`, since the compile time check ensures `r1` and `r5` are same, so the check is deferred to r5's consumers.

liqiangxl: > Previously, when this condition hits, it exits the while loop. In this version, it just exits the inner while loop and moves on to the step 2. What does the difference mean?

before going to step-2, it is captured by further checks of ids_to_resolve, see [code](https://github.com/NVIDIA/Fuser/blob/3153d3ae74edd6902a63af9f5c6d1e7c1aec1ccb/csrc/scheduler/registry_utils.cpp#L638).
```
        // break from this chain if no id to resolve
        if (ids_to_resolve.empty()) {
          break;
        }
```
and breaks from this chain, so still same to the original version.

Added comments for clarity.

liqiangxl:
!build

csrc/scheduler/registry_utils.cpp
naoyam: I see. Makes sense.

naoyam (approved):
LGTM

PR #2279, opened 2024-05-21
Replace bitwise assignment ops on bools

jacobhinkle:
Thanks to the suggestion by @zasdfgbnm while reviewing #2272, I found some additional cases where we took a short-cut to updating bools using the bitwise assignment ops. This is not ideal since its behavior is undefined (there's no guarantee that the underlying representation of `true` is `b1` and not `b10` or any other non-zero value). More importantly, writing `a |= b` as `a = a || b` allows us to short-circuit if `a == true`. Using bitwise `a |= b`, `b` will always be evaluated.


jacobhinkle:
!build

zasdfgbnm (approved):

PR #2281, opened 2024-05-22
step-1 of revising outer reduction heuristics, revise unroll & vectorization

liqiangxl:
**1. Preliminary**

- **What this PR is about**
**step-1 of revise the heuristics of outer reduction, specifically:**
(a) revise vectorization & unroll
(b) revise number of serial loops on top of unroll & number of target blocks
**step-2 is using block reduction instead of grid reduction for some input sizes. will be added in #2302.**

- **Background context.**
current outer reduction is lower than torch compile in many cases, e.g. tested the following cases
outer dim: 2048 to 32768 step by 2048
Inner dim: 2048 to 32768 step by 2048, additional GPT-2 sizes [768, 1024, 1536, 1600] x 3 and 4.
`32% of these cases have a relative speed lower than 0.9x`. For example at 8192 x 3072, nvFuser takes 34 us while torch.compile takes 30 us (two kernels, total kernel time = 23 us, estimated launch overhead = 7 us).

- **SOL analysis**
(1) Current main branch.
(2) Theoritical SOL.  Latency = IO bytes / hardward bandwidth
(3) Max achieved performance. nsight compute analsys shows the kernel is memory bound, use large vectorization factor and unroll can reduce memory load instructions and hide memory access latency. For a case with reduction dim of 8192 and iteration dim of 3072, to reach SOL, we need to vectorize by 8 and unroll by 16. Instead of doing grid reduction, it uses block reduction using only 48 out of the 132 SMs. 
image
-  **Reasons of the performance gap with SOL**
(1) vectorization is only used when `n_elems * max_input_dtype_size > 64l * 1024l * 1024l`
(2) `vect * unroll` is limited to 8. If vectorized by 8, then `unroll = 1`
(3) Max performance was achieved using only block reduction with 48 blocks, current code uses grid reduction with 8 * SM blocks. The limited usage of vectorization and unroll caused long scoreboard warp stalls. 

**2. Approaches**
- **Technical details**
(1) **Vectorization** reduces number of ld/st instructions. So, start vectorization after bdimx reaches a warp size. 
(2) **Unroll** hide memory access latency. So, increase `vect * unroll` from 8 to 16. It ensures we have at least a unroll of 2 even vectorized by 8 and avoids using too many registers which may lead to low occupancy and reduces performance espacially for cases where computation cost is high, e.g. gelu bwd.
(3) reduce number of **serial iterations** on top of unroll from 4 to 2. This change maintains same workload per thread since the increase of `unroll * vect` increased workload per thread.
(4) use **4 * SM blocks** to reduce communication cost when input elements is less than 64M. Otherwise, still use **8*SM blocks**. This is empirical.

- **Limitations and concerns**
(1)  `vect * unroll` is set to 8. However, the best value varies for different fusions and hardwares. For example for outer reduction without fused ops, to reach SOL, we should be `vect = 8 and unroll = 16` for case `8192 x 3072`. However, this large unroll leads to high register usage and low occupancy which leads to regression for compute bound kernels such as gelu backward.
(2) still have empirical values in the heuristics

**3. Results**
**(1) outer reduction.**
**For the specific case of `8192 x 3072`**
image

**compared with main branch**, 53% of the tested cases achieved a speedup higher than 1.1x.
image

**compared with torch.compile**, 22% of the tested cases achieved a speedup higher than 1.1x.
image

**(2) gelu backward**
image

**(3) CPP benchmarks** 
Several regressions for cases with small reduction dim, this PR prioritized vectorization, it leads to lower unroll, so needs more threads/blocks in the reduction dim, which leads to grid reduction which is slower than the original block reduction. Will be fixed after step-2 where block reduction is prioritized. For example, outer reduction of 8 x 2 Million:
R x I = 8 x 2097152 | vect | unroll | bdimx | bdimy | gdimx | gdimy | Bandwidth, GB/s
-- | -- | -- | -- | -- | -- | -- | -- 
main | 1 | 8 | 512 | 1 | 4096 | 1 | 1600
This PR | 8 | 2 | 512 | 1 | 512 | 2 | 1200

Performance changes:
image


csrc/scheduler/reduction.cpp
```diff
@@ -521,15 +521,32 @@ std::shared_ptr<ReductionParams> outerReductionHeuristic(
...
+  // for memory-bound kernels. However, it increases register pressure and may
+  // lead to lower occupancy which is bad for compute-bound kernels. In most
+  // cases, the scheduler uses 512 threads and to reach an occupancy of 50%,
+  // each thread can use up to 64 registers, here only 8 registers are reserved
+  // for unroll and vectorization. The fused ops can have 48 registers for other
+  // purposes. Test shows it leads to 50% occupancy for outer reduction without
+  // fused ops and 50% occupancy for gelu backward which fused 21 ops including
+  // the expensive tanh op. Further tuning of this heuristic can utilize the
+  // cost of the fused ops.
+  const int64_t buffer_reg_count = 8L;
```
jjsjann123, line 535: nitpick, PR description mentioned about increasing max_unroll from 8 to 16. I imagine that's assuming data type fp16.

I'm curious on how this would translate to fp32 in terms of performance, are all the perf measurement done in fp32?

csrc/scheduler/reduction.cpp
```diff
@@ -581,13 +598,9 @@ std::shared_ptr<ReductionParams> outerReductionHeuristic(
       target_blocks *= 2;
     }
 
-    // Delay increasing unroll until we're at a quarter of the target blocks and
-    // threads
+    // Delay increasing unroll until we more than a wave of blocks
```
jjsjann123, line 601:
```suggestion
    // Delay increasing unroll until we have more than a wave of blocks
```

jjsjann123: naive question. Why do we want to have `more than`, instead of `no less than`?

csrc/scheduler/reduction.cpp
```diff
@@ -521,15 +521,32 @@ std::shared_ptr<ReductionParams> outerReductionHeuristic(
...
       scheduler_utils::lastPow2(
           std::max((int64_t)n_tensor_inputs >> 2, (int64_t)1)));
 
   const int64_t n_elems = total_reduction_numel * total_iteration_numel;
-  const int64_t n_waves = 8;
+  // Try to use 4 * SM blocks to reduce communication cost. But still
+  // use 8 * SM blocks if the problem size is large, it increased requested
+  // waves, helps hiding memory latency, if still use 4 * SM blocks, about 5%
+  // regression for compute bound kernels. Not much change for memory bound.
+  const int64_t n_waves = n_elems >= (int64_t)64 * 1024 * 1024 ? 8 : 4;
```
jjsjann123, line 549: nitpick, this is also contradicting PR description, which were saying that we are always targeting 4 waves.

Let's have a loud comment on this number being justified empirically.

jjsjann123 (commented):
Nice speed up!

One issue I want to point out is that, the heat map showing perf comparison with main goes with sizes ranging `[(2048, 32768), (2048, 32768)]` are positive over the board. But we did mention regression on small reduction sizes below in cpp benchmarks.

I know we'll have step two to fix that so I"m not too worired. Just a note that, in the follow up PR, we might want to make sure that we perform perf comparison on main branch before this PR is merged.

Actually. before merging this PR, please put up an issue on perf regression on cpp benchmark and marking the reference commit for comparison. Just so we have a concrete follow up item.

csrc/scheduler/reduction.cpp
liqiangxl: heatmap is based on fp16, cpp benchmark has both fp16 & fp32.

jjsjann123: interesting. I'm wondering if cpp benchmark regressions are just catching the small reduction tensors or is it showing regression on fp32.

Do you think we should get the fp32 benchmark with the heatmap configs just to verify that?

liqiangxl: CPP benchmarks didn't show regression for fp32.
Double checked heatmap test with fp32, also no regression.
**Outer reduction**
image
**Gelu Bwd**
image

csrc/scheduler/reduction.cpp
liqiangxl: revised PR description.

liqiangxl:
> Nice speed up!
> 
> One issue I want to point out is that, the heat map showing perf comparison with main goes with sizes ranging `[(2048, 32768), (2048, 32768)]` are positive over the board. But we did mention regression on small reduction sizes below in cpp benchmarks.
> 
> I know we'll have step two to fix that so I"m not too worired. Just a note that, in the follow up PR, we might want to make sure that we perform perf comparison on main branch before this PR is merged.
> 
> Actually. before merging this PR, please put up an issue on perf regression on cpp benchmark and marking the reference commit for comparison. Just so we have a concrete follow up item.

added issue https://github.com/NVIDIA/Fuser/issues/2396

csrc/scheduler/reduction.cpp
```diff
@@ -708,15 +706,14 @@ std::shared_ptr<ReductionParams> outerReductionHeuristic(
 
   // Try to hit a wave by going cross reduction
   grdim = std::min(rDimAvail(), ceilDiv(device_multiprocessor_count, gidim));
-
-  // // Extend to go to target blocks, but keep 16 iterations per thread
+  // // Extend to go to target blocks
```
jjsjann123, line 709: nitpick: is it intended to have `// //` here?

jjsjann123 (approved):
LGTM. Thanks for the extensive benchmarking. :bow:

csrc/scheduler/reduction.cpp
liqiangxl: no, that's typo.

liqiangxl:
!build

liqiangxl:
!build

csarofeen:
@liqiangxl what GPU was used for the performance measurements? How does this change in heuristics change performance on other GPUs? Do we ever check GPU performance on anything other than A100 or H100?

csarofeen:
It’s also unclear to me if both A100 and H100 were checked, which would always be important to do.

liqiangxl:
> It’s also unclear to me if both A100 and H100 were checked, which would always be important to do.

The result of A100 is not posted in PR description. For this case it is similar to H100 where some regressions are found when reduction dim is small and will be fixed in [step-2](#2302 ) see [doc](https://docs.google.com/document/d/12yBhqBHlZFKyxjgXqHjXxOfzZjsjXySHvwR0_8LsfZg/edit#heading=h.nuw7dwswjq47).
image

liqiangxl:
> @liqiangxl what GPU was used for the performance measurements? How does this change in heuristics change performance on other GPUs? Do we ever check GPU performance on anything other than A100 or H100?

Only checked H100 & A100. Do you have a list of other GPUs we should also check? I can add them to the benchmark when reivising heurisitcs.

liqiangxl:
> @liqiangxl what GPU was used for the performance measurements? How does this change in heuristics change performance on other GPUs? Do we ever check GPU performance on anything other than A100 or H100?

Sorry for editing your original comment by accident. I was supposed to do 'Quote reply'

Issue #2199, opened 2024-05-03
Write a sharded transformer block in nvFuser API. 


wujingyue:
This is to unblock @cowanmeg and @samnordmann 's distributed matmul experiments. 

I'll start with the tensor parallelism proposed by [the original Megatron-LM paper](https://arxiv.org/pdf/1909.08053). 
1. Only MHA and MLP are sharded. 
2. Activations are sharded in 2D, batch and hidden. However, the batch dimension sharding is just for data parallelism and the dimension is never resharded. 
3. Weights are sharded in 1D, the hidden dimension.

wujingyue:
Note to myself: I'll first try to get a single-device nvFuser python definition from Thunder, and then we can manually shard it using nvFuser's API.

@Priya2698 pointed  me to the nv_enable_linear flag (https://github.com/Lightning-AI/lightning-thunder/blob/90a0f4c0d0a90d1e94684a847f3adfe2230985b4/thunder/tests/test_nvfuser.py#L875) that I'll need to turn on to enable prims.linear via nvFuser. I'll probably need to nv_enable_bookend=False as well.

wujingyue:
Note to myself: I'll start with the following benchmark

```
$ pytest thunder/benchmarks/targets.py -k test_nanogpt_block_fwd[thunder] -s
```
which exercises one transformer layer in nanoGPT: 
https://github.com/Lightning-AI/lightning-thunder/blob/cab020881765594fd9552d4deb8cc4e0f64410d2/thunder/tests/nanogpt_model.py#L132-L143

wujingyue:
cc @Priya2698 

`a.ndim==2` is the first check that failed. Here's how you can reproduce the problem: 

With the following patch 

```
diff --git a/thunder/executors/nvfuserex_impl.py b/thunder/executors/nvfuserex_impl.py
index c955da06..4767ab9c 100644
--- a/thunder/executors/nvfuserex_impl.py
+++ b/thunder/executors/nvfuserex_impl.py
@@ -2201,6 +2201,7 @@ def _linear_check(a: TensorProxy, b: TensorProxy, bias: TensorProxy | None) -> b
         return False

     enable_linear: None | bool = get_compile_option("nv_enable_linear", "Enable nvFuser matmul.")
+    enable_linear = True
     if not enable_linear:
         return False
     # Verify linear inputs and bias (optional) are supported tensors.
[...]

PR #2282, opened 2024-05-22
[User Schedule] Add print functions for TensorViews and User Scheduled Fusions

rdspring1:
This PR adds two functions for debugging user scheduled fusions in python frontend.

`std::string FusionDefinition::userScheduleIr()` returns the user scheduled fusion IR.

Usage:
```python
class SomeFusion(FusionDefinition):
    def definition(self):
        # create some fusion

    def schedule(self):
        print(fd.sched.user_schedule_ir())
```

`to_string(python_frontend::Tensor)` returns the string representation for its corresponding TensorView.

Usage:
```python
print(fd.sched.to_string(some_tensor))
```

### PR List:
1. #2282 **<<<  YOU ARE HERE**
2. #2283
3. #2284
4. #2285
5. #2286
6. #2287
7. #2288


rdspring1:
!build

naoyam (commented):
This seems really useful. Thanks! I'll let others chime in on the Python interface itself, but could you also create some documentation? Do we already have a doc for the Python interface? If so, I suppose that should be updated.

wujingyue:
I'm behind for many things this week, so sorry for not being able to review this series of PRs. I think the line of the work (especially https://github.com/NVIDIA/Fuser/pull/2285) could potentially help @cowanmeg manually shard the Transformer block (https://github.com/NVIDIA/Fuser/issues/2199) more easily.

csrc/python_frontend/fusion_definition.cpp
```diff
@@ -228,6 +228,14 @@ std::string FusionDefinition::fusionIr() {
   return ss.str();
 }
 
+std::string FusionDefinition::userScheduleIr() {
+  NVF_CHECK(id().has_value(), "Invalid fusion definition!");
+  NVF_CHECK(user_sched_ != nullptr, "UserSchedule is not valid!");
```
jjsjann123, line 233: Maybe we can relax the check here a bit. i.e. return an error message when user_schedule is not available, instead of an assert.

csrc/python_frontend/python_bindings.cpp
```diff
@@ -593,6 +593,10 @@ void initNvFuserPythonBindings(PyObject* module) {
           "_fusion_ir",
           [](FusionDefinition& self) { return self.fusionIr(); },
           py::return_value_policy::reference)
+      .def(
+          "_user_schedule_ir",
```
jjsjann123, line 597: IIUC, userScheduleIr can be safely called after a single execution with a custom schedule.

But we can also safely call it outside of `def schedule(self):` as well, my question is, is that an intended usage? Otherwise, if we are only supposed to use it inside `schedule`, we might as well keep this method under `nvf_sched` instead?

I'm not pushing strongly against this, but if we decide to keep the API as-is, I'd highly recommend relaxing the NVF_CHECK above with proper warning/recommendation on proper usage.

rdspring1: It could be useful to print fusion IR outside of the `schedule` function. I'll keep it and replace the NVF_CHECK with a string message.

csrc/python_frontend/python_bindings.cpp
```diff
@@ -2777,6 +2781,20 @@ void initNvFuserPythonBindings(PyObject* module) {
   py::class_<FusionDefinition::SchedOperators> nvf_sched(
       fusion_def, "SchedOperators");
   nvf_sched.def(py::init<FusionDefinition*>());
+  nvf_sched.def(
+      "to_string",
```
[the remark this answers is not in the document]
rdspring1: We wrap the print functions for `FusionDefinition` in `nvfuser/__init__.py` like https://github.com/NVIDIA/Fuser/blob/main/nvfuser/__init__.py#L246-L253. I'm not sure if we need the `_` if it is a public function for this case.

csrc/python_frontend/fusion_definition.cpp
```diff
@@ -228,6 +228,18 @@ std::string FusionDefinition::fusionIr() {
   return ss.str();
 }
 
+std::string FusionDefinition::userScheduleIr() {
+  NVF_CHECK(id().has_value(), "Invalid fusion definition!");
```
jjsjann123, line 232: IIUC, vvv
```
  class SomeFusion(FusionDefinition):
    def definition(self):
        # create some fusion
    def schedule(self):
        print(self._user_schedule_ir())

  fd = SomeFusion()
  fd.user_schedule_ir() # it would assert here with `Invalid fusion definition`?
```

one more nitpicking, this really isn't your problem, but we can probably use a better error message and probably a warning rather than an assert. I think all the other FusionDefinition methods defined in `nvfuser/__init__.py` suffers the same.

Keep it as-is in this PR. But I think we can use a clean up PR afterwards. Wdyt @kevinstephano

jjsjann123 (approved):
LGTM, some minor comments on user experience but let's clean up in following PRs.

jjsjann123 (changes requested):
One last question: What's the story with version bump?

Since we are only adding new APIs here, I'm OK with keeping the version as-is and bump then towards the end of the stacked PRs. (and pretend that they do not exist).

One other thing is, let's add a test as well.

csrc/python_frontend/python_bindings.cpp
rdspring1: I moved the method to `nvf_sched` since `user_sched_` is reset after `finalize_schedule` is called and is not available outside of `def schedule`.

rdspring1:
!build

rdspring1:
I updated the version and add some tests.

rdspring1:
!build

jjsjann123 (approved):

PR #2287, opened 2024-05-22
[User Schedule] Support filtering tensors according to some predicate function

rdspring1:
Summary: Use python built-in `filter` function instead of mapping CPP filter functions directly to python-frontend.

Notable Changes:
##### Track all `Tensors` in `FusionDefinition`
  * Create `std::vector<Tensor> all_tensors_`
  * Add `Tensor` to vector in `FusionDefinition::defineTensor``
  * In python, expose through `FusionDefinition.get_all_tensors()`
  * **Limitation:** Tensors created by CPP operations can be hidden from `FusionDefinition`

##### Create filter functions for individual TensorViews
  * `isReduction` and `equality` operator for Scalar, Tensor, and Vector

##### Usage: Run `filter(filter_fn, FusionDefinition.get_all_tensors())`
```python
class SomeFusion(FusionDefinition):
    def definition(self):
         # create some fusion

    def schedule(self):
        # Find all reduction tensors in fusion
        reduction_tvs_generator = filter(self.sched.is_reduction, self.sched.get_all_tensors())
```


rdspring1:
!build

rdspring1:
!build

csrc/python_frontend/fusion_definition.h
```diff
@@ -209,6 +252,9 @@ class NVF_API FusionDefinition : public FusionState {
   Fusion* prev_fusion_;
   //! Data member for holding the current user schedule object
   UserSchedule* user_sched_;
+  //! A vector of all Tensors in FusionState. It is a subset of
+  //! recording_state_.
+  std::vector<Tensor> all_tensors_;
```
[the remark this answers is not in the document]
jjsjann123: It also felt messy to keep an `all_tensors_` in `FusionDefinition`, while we have `recording_state_` in `FusionState`.

csrc/python_frontend/fusion_definition.cpp
```diff
@@ -403,6 +404,10 @@ State FusionDefinition::recordingState(size_t index) const {
   return recording_state_.at(index);
 }
 
+const std::vector<Tensor>& FusionDefinition::getAllTensors() const {
+  return all_tensors_;
```
jjsjann123, line 408: I would strongly encourage against creating `all_tensors_` container.

You should be able to just filter recording_state_ to find indices to `TensorView` entries in `fusion_state_`. We can then construct `Tensor` with the `Tensor(index, tv->nDims(), this)`.

csrc/python_frontend/python_bindings.cpp
```diff
@@ -503,13 +505,15 @@ void initNvFuserPythonBindings(PyObject* module) {
   tensor_class.def("_get_fusion_definition", [](Tensor& self) {
     return self.fusion_definition;
   });
+  tensor_class.def(pybind11::self == pybind11::self);
```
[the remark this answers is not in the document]
rdspring1: I added `operator!=`

tests/python/test_schedule_ops.py
```diff
@@ -442,6 +442,80 @@ def schedule(self):
...
+                # NOTE Manually broadcast because fusion definition cannot access hidden reduction tensor view.
+                self.bcast_sum1 = fd.ops.broadcast(self.sum1, [False, True])
+                self.var = fd.ops.div(self.bcast_sum1, self.norm_const)
+
+                self.t0_diff = fd.ops.sub(self.t0, self.mean)
+                self.var_eps = fd.ops.sqrt(fd.ops.add(self.var, self.s0))
+                self.t0_norm = fd.ops.div(self.t0_diff, self.var_eps)
+                self.add_output(self.t0_norm)
+
+            def schedule(self):
```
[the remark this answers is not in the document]
rdspring1: Renamed `tv` suffix to `tensor`

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

csrc/python_frontend/python_bindings.cpp
```diff
@@ -3149,6 +3157,31 @@ void initNvFuserPythonBindings(PyObject* module) {
...
+      [](FusionDefinition::SchedOperators& self, Tensor tensor) {
+        NVF_CHECK(
+            self.validUse(),
+            "Attempting to use a SchedOperators Op prior to definition!");
+        // Determine if tensor is a result from a reduction operation.
+        FusionDefinition* fd = self.fusion_definition;
+        TensorView* tv =
+            fd->getFusionState(tensor.index)->template as<TensorView>();
+        return (
+            !tv->isFusionInput() &&
```
jjsjann123, line 3178: out of curiosity, here we do not have segmentation, so technically, we should never see inputs with reduction id?

csrc/python_frontend/python_bindings.cpp
```diff
@@ -3149,6 +3157,31 @@ void initNvFuserPythonBindings(PyObject* module) {
...
+            std::any_of(
+                tv->getLeafDomain().begin(),
+                tv->getLeafDomain().end(),
```
jjsjann123, line 3181: nitpick: we don't care about loop structure here right, this should be rfactor/logical domain instead.

jjsjann123 (approved):
minor comments. LGTM

csrc/python_frontend/python_bindings.cpp
rdspring1: Switched to logical domain

rdspring1: Note: The leaf domain is what schedule_utils is using.

csrc/python_frontend/python_bindings.cpp
rdspring1: I'm interested in adding segmentation. I'll leave it here to avoid missing it in the future.

csrc/python_frontend/fusion_definition.h
```diff
@@ -45,6 +45,25 @@ struct Tensor {
     return index;
   }
 
+  bool operator==(const Tensor& other) const {
```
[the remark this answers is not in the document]
rdspring1: In `FusionDefinition::tensors()`, we're creating new Tensors with the same state. Checking only Tensor addresses could create duplicates.

rdspring1:
!build

csrc/python_frontend/fusion_definition.h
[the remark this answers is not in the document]
jjsjann123: > Checking only Tensor addresses could create duplicates.

that's not what I was suggesting though. It's just a short-cut if we have the same address then we don't need to check for each field. Otherwise, the existing logic will kick in to detect different Tensors pointing to the same TensorView/State.

But it's not a big issue.

PR #2288, opened 2024-05-22
Select CompileParams for TMA operations automatically at compile-time

rdspring1:
When using TMA operations, we must use 32-bit indexing and disable magic zero support. Currently, we manually pass the correct `CompileParams` to `FusionExecutor::compileFusion`. This PR automatically updates the `CompileParams` if TMA operations are detected during compilation. It also includes a basic normalization kernel using TMA operations in the python frontend.

### DIFF
--- a/csrc/executor.cpp
+++ b/csrc/executor.cpp
@@ -357,7 +357,8 @@ void FusionExecutor::compileFusion(
     NVF_ERROR(
         !has_cp_async_bulk ||
             (compile_params.index_type.value() == PrimDataType::Int32),
-        "Compilation with int64 is requested but int32 is required because of TMA operations");
+        "Compilation with int64 is requested but int32 is required because ",
+        "of TMA operations.");
 
   } else if (arg_index_type == PrimDataType::Int) {
     // If the given compile option doesn't specify the index type, and
@@ -366,6 +367,10 @@ void FusionExecutor::compileFusion(
     // it's safe to use 32-bit for the whole kernel, so unless it's
     // specified through CompileParams, we do not use 32-bit indexing.
     compile_params.index_type = arg_index_type;
+    NVF_ERROR(
+        !has_cp_async_bulk,
+        "Compilation with int64 is required based on input arguments, but ",
+        "int32 is required because of TMA operations.");
   } else if (has_cp_async_bulk) {
     // TMA operations require 32-bit indexing.
     compile_params.index_type = PrimDataType::Int32;
--- a/csrc/executor.cpp
+++ b/csrc/executor.cpp
@@ -270,7 +270,7 @@ void FusionExecutor::compileFusion(
     return;
   }
 
-  const auto& exprs = fusion->exprs();
+  const std::vector<Expr*>& exprs = fusion->exprs();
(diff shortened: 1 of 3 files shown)


rdspring1:
!build

csrc/python_frontend/fusion_definition.cpp
```diff
@@ -145,11 +146,16 @@ void FusionDefinition::finalizeSchedule(
...
   if (multidevice_executor_ == nullptr) {
+    // NOTE: TMA operations require 32-bit indexing and magic zero support to be
+    // disabled.
     user_sched_->executor->compileFusion(
         user_sched_->schedule.get(),
         inputs,
         user_sched_->fusion_id_,
-        user_sched_->device_id_);
+        user_sched_->device_id_,
+        (uses_tma_ops)
```
jjsjann123, line 156: this feels a bit hacky here.

IIUC, we should be able to traverse the fusion IR `user_sched_->schedule` looking for `LoadStoreOpType::CpAsyncBulkTensorTile` to figure out that we should have use the modified compile param?

jjsjann123: Maybe it's better to do that inside compileFusion instead?

jjsjann123 (commented):
> When using TMA operations, we need to use 32-bit indexing and to disable magic zero support in the CompileParams

I have a question regarding this statement above. If this is a general thing, we probably want to have it done automatically in `compileFusion`, instead of require this as a user flag to enable it.

If it's not a general thing, plumbing it through the API seems to be a bad idea to begin with...

csrc/python_frontend/fusion_definition.cpp
rdspring1: We can update `CompileParam` in `compileFusion` in a separate PR and leave this one open as an example for now.

jjsjann123: Sounds good. I don't have any more concerns regarding this PR. But let's chat offline and discuss how we would want to merge this stack.

jjsjann123:
I think this is the last PR in the stack, let me know when it's ready to be reviewed.

rdspring1:
!build

rdspring1:
!build

csrc/executor.cpp
```diff
@@ -343,13 +352,21 @@ void FusionExecutor::compileFusion(
...
+            (compile_params.index_type.value() == PrimDataType::Int32),
+        "Compilation with int64 is requested but int32 is required because of TMA operations");
+
   } else if (arg_index_type == PrimDataType::Int) {
     // If the given compile option doesn't specify the index type, and
     // the arguments require 64-bit indexing, we need to use 64-bit
     // indexing. Note that if the arg type is 32-bit, it doesn't mean
     // it's safe to use 32-bit for the whole kernel, so unless it's
     // specified through CompileParams, we do not use 32-bit indexing.
     compile_params.index_type = arg_index_type;
```
zasdfgbnm, line 366: Should we add a check `NVF_ERROR(! has_cp_async_bulk_` here?

rdspring1:
!build

zasdfgbnm (approved):

PR #2280, opened 2024-05-22
change from at::ones to at::randn in test of OuterReductionMagicScheduler

liqiangxl:
should use `at::randn` instead of `at::ones` to generate inputs for tests

PR #2289, opened 2024-05-22
Remove MatmulRole::OUTPUT_AUX

jacobhinkle:
We already have `MatmulRole::OUTPUT_D`. We populate both of these by reverse sorting all outputs that are downstream of the `MmaOp` first by number of uses then by `name()`. The first in that reverse sorted list is `OUTPUT_D` and all others are `OUTPUT_AUX`. After this PR, there is only the `OUTPUT_D` list but it might have more than one element: namely the former `OUTPUT_AUX` tensors are now appended in the same order to the `OUTPUT_D` list in the roles map.

I believe the main reason to reverse sort by uses is to find a reference tensor for reverse propagating transforms to epilogue inputs here: https://github.com/NVIDIA/Fuser/blob/0892405e5396820ffc4a300cba916cd6cb65e6be/csrc/scheduler/matmul.cpp#L660-L672
There should be no change to this section since we have not changed the first element of the `OUTPUT_D` vector.

### DIFF
--- a/csrc/scheduler/mma_utils.cpp
+++ b/csrc/scheduler/mma_utils.cpp
@@ -1288,13 +1288,7 @@ RolesMapOpt getTensorsRoles(
     });
 
     if (!storage.empty()) {
-      // NOTE: currently, we pick as a reference tensor one with `m` and `n`
-      //       IterDomains and the most uses
-      auto pos = storage.begin();
-      roles_map[MatmulRole::OUTPUT_D].push_back(*pos);
-      for (++pos; pos != storage.end(); ++pos) {
-        roles_map[MatmulRole::OUTPUT_AUX].push_back(*pos);
-      }
+      roles_map[MatmulRole::OUTPUT_D] = storage;
     }
   };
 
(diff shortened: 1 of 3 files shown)


jacobhinkle:
I think it would be a good idea for a follow up PR to give more descriptive names instead of A, B, C, D. For example
```
enum class MatmulRole { Operand, EpilogueInput, Output };
```
I don't think there's much reason to treat operands differently since we wind up scheduling them the same way anyway.

jacobhinkle:
!build --diff

zasdfgbnm (approved):

jacobhinkle:
Thunder test failure is unrelated and due to some upstream changes in TE and setuptools (cf. https://github.com/NVIDIA/Fuser/pull/2280#issuecomment-2125658106).

csrc/scheduler/mma_utils.cpp
```diff
@@ -1231,27 +1231,12 @@ RolesMapOpt getTensorRoles(
...
 
-  // NOTE: sort output roles in descending order by uses() size, and
-  //  if equal then by name() to ensure the stable ordering of tensor
-  //  views in collections assigned to the supported roles
-  std::sort(storage.begin(), storage.end(), [](TensorView* a, TensorView* b) {
-    return (a->uses().size() == b->uses().size())
-        ? (a->name() < b->name())
-        : (a->uses().size() > b->uses().size());
-  });
-
```
jacobhinkle, line 1242: I realized this is redundant since we now do this for all roles, not just outputs.

drzejan2:
@jacobhinkle - thanks for preparing this change, and I agree with you regarding renaming the roles, they shouldn't be as strictly bound to naming conventions used by BLAS libraries, we should be more generic.

PR #2290, opened 2024-05-22
Remove unnecessary parameter in buildStatefulInliningInfo

zasdfgbnm:


zasdfgbnm:
!build

csrc/id_model/id_model.cpp
```diff
@@ -554,8 +553,8 @@ StatefulInliningInfo buildStatefulInliningInfo(
         auto consumer_tv_i = consumer_tvs.vector().at(i);
         auto all_consumer_i_ids = ir_utils::allIDsOf(consumer_tv_i);
 
-        auto sibling_map =
-            exact_graph.buildMapBetween(all_consumer_ids, all_consumer_i_ids);
+        auto sibling_map = permissive_graph.buildMapBetween(
```
zasdfgbnm, line 556: I don't think exact graph and permissive graph makes any difference in siblings.

zasdfgbnm:
!build

zasdfgbnm:
So building the loop graph does not anything about exact graph. Exact graph is related only for loop promotion?

naoyam (approved):
Makes sense. Thanks for the cleanup!

Issue #1964, opened 2024-03-19
Lowering failure with Fusion involving squeeze of broadcast


jacobhinkle:
```c++
TEST_F(NVFuserTest, FOO) {
  std::unique_ptr<Fusion> fusion_ptr = std::make_unique<Fusion>();
  auto fusion = fusion_ptr.get();
  FusionGuard fg(fusion);

  DataType input_dtype = DataType::Float;
  auto tv0 = makeContigTensor(3, input_dtype);
  auto tv1 = makeContigTensor(1, input_dtype);
  auto tv2 = makeContigTensor(3, input_dtype);
  fusion->addInput(tv0);
  fusion->addInput(tv1);
  fusion->addInput(tv2);

  auto tv3 = broadcast(tv1, {true, false, true});
  auto tv4 = set(tv3);
  auto s = IrBuilder::create<Val>(3);
  auto tv5 = expand(tv4, {s, s, s});
  auto tv6 = mul(tv5, tv2);
  auto tv7 = sum(tv6, {0, 2});
  auto tv8 = mul(tv0, tv2);
  auto tv9 = sum(tv8, {0, 2});
  auto tv10 = broadcast(tv9, {true, false, true});
  auto tv11 = set(tv10);
  auto tv12 = squeeze(tv11, {true, false, true}, false);
  auto tv13 = mul(tv12, tv1);

  fusion->addOutput(tv7);
  fusion->addOutput(tv13);

  auto options =
      at::TensorOptions().dtype(at::kFloat).device(at::kCUDA, 0);
  auto t0 = at::randn({3, 3, 3}, options);
  auto t1 = at::randn({3}, options);
  auto t2 = at::randn({3, 3, 3}, options);
  std::vector<c10::IValue> inputs{t0, t1, t2};

  FusionExecutorCache fec(std::move(fusion_ptr));
  auto outputs = fec.runFusionWithInputs(inputs);
  testValidate(fec.fusion(), outputs, inputs, __LINE__, __FILE__);
}
```
This fails with
> C++ exception with description "!concrete_to_loop.count(concrete_loop_id) INTERNAL ASSERT FAILED at "/opt/pytorch/nvfuser/csrc/device_lower/analysis/index_compute.cpp":638, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. Unsupported loop structure. Two loops are mapped together.bS31{1} and bS28{1}
[...]

Issue #1778, opened 2024-02-16
[Proposal] Remove Broadcast IterType


jacobhinkle:
(This issue might be updated as our discussion evolves)

# Context

Broadcast IterDomains signify that a future operation might perform broadcasting involving this dimension. However, broadcasting is fundamentally a behavior of an operation, not an attribute of a piece of data. In fact it is the _concretization_ of a broadcast that matters. The reason it is attached to tensors is inherited from numpy and PyTorch, which need a convenient way to specify that an op should do broadcasting. In an eager system, implementing broadcast semantics can be done locally to each op. However, shadowing that system in nvFuser means that we carry around Broadcast IterDomains that require a lot of special handling. Furthermore, some operations like `broadcast`, `squeeze`, and `expand` are really no-ops, but representing them in our IR can lead to unintended data transfer.

# Summary of Proposal

In a side conversation, @zasdfgbnm said this (see #1628):
1. What is important is the concretization of broadcasting. All the history of broadcasting IterDomains (IDs) before concretization are not relevant, and I am not interested in inspecting it. Especially, I don’t care which broadcasting ID is mapped to which broadcasting ID in root domain map.
2. Broadcasting IterDomain are just auxiliary placeholders. They should be capable of being created and annihilated arbitrarily.

Indeed, we should be able to move all broadcasts to the point just before the broadcast is concretized. This proposal goes slightly further and just attaches the broadcast concretization info to the concretizing op.

This proposal seeks to remove `IterType::Broadcast` as well as `BroadcastOp`, `SqueezeOp`, `ExpandOp`, and their related C++ and Python frontend operations. They will be replaced by an attribute in `Expr` that tracks which axes in each input are broadcasts concretized by that `Expr`. Broadcast concretization will need to be explicitly declared during fusion definition.
![image](https://github.com/NVIDIA/Fuser/assets/1454944/3b33ad06-d1f7-4463-8779-d144c1b79d61)

To work an actual example from the recent issue #1757 :
![image](https://github.com/NVIDIA/Fuser/assets/1454944/5ca45229-0f89-42d8-9581-e3cd9a733bba)

# Details of Approach

## Tentative Plan

Here is a sketch of how we could proceed. Each stage is likely somewhat involved, but any conceptual misunderstandings should be encountered in the first stages. The later stages will involve updating our frontend so we can coordinate with integration for a safe transition to the new interface.

[...]

Issue #340, opened 2023-05-15
Eliminate unresolved broadcasts


jacobhinkle:
While implementing concretization of dynamic resizes (#258), it occurred to me that we should handle this particular type of concretization more generally. The only reason resize is a dynamic operation is that the output might have an extent equal to 1. We currently mark `IterDomain`s with extent 1 as `Broadcast` so that they maintain the ability to broadcast when they're used in later expressions. However, we don't actually check that they ever _are_ used this way. That means that if users pass inputs that sometimes have size one in some dimension (such as a batch or channel dimension) they will not be able to re-use kernels for size >1 even when that should be possible. In the short term, this also means that we cannot use `cat` along size-1 dimensions (#224).

One reason not to eliminate broadcasts too hastily is that one might add new expressions to the `Fusion` right up until the point that it's lowered. Even if we don't expect that to happen, until the concretization phase was introduced in #24 it was not clear at which point we even should attempt to clear out unresolved broadcasts.

Now that we have a concretization stage and will soon have an "initial info" building stage of concretization (#244), there is a natural place to put this analysis.

# Propagation of broadcast removal

When an `IterDomain`, `bS0`, with `IterType` equal to `Broadcast` is used in an operation in a position that aligns with another `IterDomain`, `iS1`, of type `Iteration`, we say that this "resolves" the broadcast `bS0`. This is an important event since it means we cannot use identical indexing for `iS1` and `bS0`, and we instead need to collapse the `iS1` index to 0 when using it in place of `bS0`. Note that `bS0` (or its `TensorView`) may appear in multiple `Expr`s, so there may be zero or multiple resolutions of a given broadcast.

If there are _zero_ resolutions of `bS0`, this means that any `Expr`s it appeared in contained other expressions that aligned perfectly; i.e. there either were no other expressions, they were unary, or other arguments also had broadcast `IterDomain`s in the aligned position. In this case, the behavior is no different from if the `IterType` of `bS0` had been `Iteration` to begin with.

## Aside: connecting the ID subgraph

To better facilitate analyzing broadcasts, it might be useful to explicitly connect the `IterDomain` subgraph of the `Fusion`. Note that of course the `Fusion` contains enough information to do the analysis as-is, but it is not very ergonomic. To see this, consider a simple unary op like `neg`. Consider the following `Fusion`:
```c++
  auto tv0 = makeConcreteTensor({3, 5});
  fusion->addInput(tv0);
[...]

PR #2291, opened 2024-05-23
remove redundant bcast-squeeze and squeeze-bcast ops

liqiangxl:
**Purpose:** add a preseg pass to remove redundant bcast-squeeze and squeeze-bcast ops
**Algorithm:**
(1) find all broadcast + squeeze patterns
(2) update fusion by replacing squeeze output with broadcast input
(3) find all squeeze + broadcast patterns
(4) update fusion by replacing broadcast output with squeeze input
In the code we don't need to explicitly do the "remove", we just replace the vals and the corresponding broadcast and squzze ops become dead code in the fusion.
**Requirement:**
(1) broadcast and squeeze have the same dim flags.
(2) broadcast output is only consumed by squeeze op or squeeze output is only consumed by broadcast.
**Unit Tests:**
Test | Pattern | Results | Purpose
-- | -- | -- | --
FusionRemoveBcastSqueeze1 | y = broadcast(x); z = squeeze(y); | remove both | basic test
FusionRemoveBcastSqueeze2 | y = broadcast(x); z = squeeze(y); apply to different dims | remove neighter | different dims
FusionRemoveBcastSqueeze3 | y = broadcast(x); z = squeeze(y);  y is output | remove squeeze | output
FusionRemoveBcastSqueeze4 | y = broadcast(x); z = squeeze(y);  z is output | remove both (x becomes the new output) | output
FusionRemoveBcastSqueeze5 | y = broadcast(x); z = squeeze(y);  x is input | remove both | input
FusionRemoveSqueezeBcast1 | y = squeeze(x); z = broadcast(y); | remove both | basic test
FusionRemoveSqueezeBcast2 | y = squeeze(x); z = broadcast(y); z is output | remove both | output
FusionRemoveSqueezeBcast3 | y = squeeze(x); z = broadcast(y); y is output | remove broadcast | output
FusionRemoveSqueezeBcastSqueezeBcast | y = squeeze(x); z = broadcast(y); m = squeeze(z); n = broadcast(m); | remove all | s-b-s-b
FusionRemoveSqueezeBcastBcastSqueeze | y = squeeze(x); z = broadcast(y); m = broadcast(z); n = squeeze(m); | remove all | s-b-b-s

**Limitations:**
(1) No ops are allowed between squeeze and broadcast ops.
(2) No ops are allowed between broadcast and  squeeze ops.
 
Needs to remove these limitations in a following PR to solve #1964 
The issue #2146 doesn't require this PR, it can be solved by #2274

### DIFF
--- a/csrc/preseg_passes/remove_bcast_squeeze.cpp
+++ b/csrc/preseg_passes/remove_bcast_squeeze.cpp
@@ -17,27 +17,22 @@ namespace {
 // Remove broadcast-squeeze and squeeze-broadcast patterns
 // TODO: still remove when have intermediate ops between broadcast and squeeze
 void removeBcastSqueeze(Fusion* fusion) {
-  std::unordered_map<Val*, Val*> replacement_map;
-  for (auto expr : fusion->exprs()) {
+  // Iterate backwards over fusion expressions.
+  // This will ensure we don't process expressions that are no longer valid
+  // after replacement.
+  auto exprs = fusion->exprs();
+  for (auto it = std::rbegin(exprs); it != std::rend(exprs); it++) {
+    Expr* expr = *it;
     // step-1: find and remove broadcast + squeeze pattern
     // before: Y = broadcast(X); Z = squeeze(Y);  M = someOp(Z)
     // after : M = someOp(X)
     // conditions: (1) broadcast and squeeze have the same dim flags
-    // special case: if Z is a fusion output, replace the output with X
-    if (auto sop = dynamic_cast<SqueezeOp*>(expr)) {
-      if (auto bcast = dynamic_cast<BroadcastOp*>(sop->in()->definition())) {
-        if (bcast->getBroadcastDimFlags() == sop->getSqueezeDimFlags()) {
-          if (sop->out()->isFusionOutput()) {
-            fusion->replaceOutput(sop->out(), bcast->in());
-          }
-          // if already have a -> b in replacement map and comes a new pair of c
-          // -> a, instead of add c -> a to the map, add c -> b to avoid two hop
-          // replacement. see test BcastSqueezeBcastSqueeze.
-          auto new_val = bcast->in();
-          if (replacement_map.count(new_val)) {
-            new_val = replacement_map[new_val];
-          }
-          replacement_map.insert({sop->out(), new_val});
+    if (auto squeeze = dynamic_cast<SqueezeOp*>(expr)) {
+      if (auto bcast =
+              dynamic_cast<BroadcastOp*>(squeeze->in()->definition())) {
+        if (bcast->getBroadcastDimFlags() == squeeze->getSqueezeDimFlags()) {
+          ir_utils::replaceValInAllExprInputsAndFusionOutputs(
+              squeeze->out(), bcast->in());
         }
       }
     }
@@ -46,28 +41,14 @@ void removeBcastSqueeze(Fusion* fusion) {
     // before: Y = squeeze(X); Z = broadcast(Y);  M = someOp(Z)
     // after : M = someOp(X)
     // conditions: (1) broadcast and squeeze have the same dim flags
-    // special case: if Z is a fusion output, replace the output with X
     if (auto bcast = dynamic_cast<BroadcastOp*>(expr)) {
-      if (auto sop = dynamic_cast<SqueezeOp*>(bcast->in()->definition())) {
-        if (bcast->getBroadcastDimFlags() == sop->getSqueezeDimFlags()) {
-          if (bcast->out()->isFusionOutput()) {
-            fusion->replaceOutput(bcast->out(), sop->in());
-          }
-          // if already have a -> b in replacement map and comes a new pair of c
-          // -> a, instead of add c -> a to the map, add c -> b to avoid two hop
-          // replacement. see test SqueezeBcastSqueezeBcast
-          auto new_val = sop->in();
-          if (replacement_map.count(new_val)) {
-            new_val = replacement_map[new_val];
-          }
-          replacement_map.insert({bcast->out(), new_val});
+      if (auto squeeze = dynamic_cast<SqueezeOp*>(bcast->in()->definition())) {
+        if (bcast->getBroadcastDimFlags() == squeeze->getSqueezeDimFlags()) {
+          ir_utils::replaceValInAllExprInputsAndFusionOutputs(
+              bcast->out(), squeeze->in());
         }
       }
     }
-  }
-  // Replace if there is any match
-  if (!replacement_map.empty()) {
-    ir_utils::replaceValue(fusion, replacement_map);
   }
 }
 


liqiangxl:
!build --diff

tests/cpp/test_preseg_passes.cpp
```diff
@@ -628,4 +628,258 @@ TEST_F(NVFuserTest, FusionRemoveEmptyMatmul_CUDA) {
...
+  }
+  return IrBuilder::create<TensorView>(
+      IrBuilder::create<TensorDomain>(
+          out_domain, TensorDomain::getContiguityFilledWith(out_domain, true)),
+      input_dtype);
+}
+
+} // namespace
+
+TEST_F(NVFuserTest, FusionRemoveBcastSqueeze1) {
```
wujingyue, line 668:
```suggestion
using RemoveBcastSqueezeTest = NVFuserTest;

TEST_F(RemoveBcastSqueezeTest, FusionRemoveBcastSqueeze1) {
```

tests/cpp/test_preseg_passes.cpp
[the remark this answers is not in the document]
liqiangxl: changed.

tests/cpp/test_preseg_passes.cpp
liqiangxl: added

csrc/preseg_passes/remove_bcast_squeeze.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include <debug.h>
+#include <ir/utils.h>
+#include <options.h>
+#include <preseg_passes/remove_bcast_squeeze.h>
+#include <root_domain_map.h>
+
+namespace nvfuser::preseg_passes {
+
+namespace {
+inline bool isFusionOutput(Fusion* fusion, Val* val) {
```
wujingyue, line 17: Have you considered https://github.com/NVIDIA/Fuser/blob/1e48b3752f8cfc976c83df5f31b6ddc2b12f6fc6/csrc/ir/base_nodes.h#L368?

csrc/preseg_passes/remove_bcast_squeeze.cpp
```diff
@@ -0,0 +1,95 @@
...
+              fusion->replaceOutput(sop->out(), bcast->in());
+            }
+            replacement_map.insert({sop->out(), bcast->in()});
+          }
+        }
+      }
+    }
+    // Replace if there is any match
+    if (!replacement_map.empty()) {
+      ir_utils::replaceValue(fusion, replacement_map);
```
wujingyue, line 48: Have you considered ir_utils::replaceValInAllExprInputsAndFusionOutputs?

csrc/preseg_passes/remove_bcast_squeeze.cpp
```diff
@@ -0,0 +1,95 @@
...
+    }
+  }
+}
+
+} // namespace
+
+void RemoveBcastSqueeze::runPass(Fusion* fusion) {
+  if (isDebugDumpEnabled(DebugDumpOption::PreSegmenterLogging)) {
+    debug() << "Fusion before RemoveBcastSqueeze:" << std::endl;
```
wujingyue, line 83: Not for this PR, but what do you think about letting https://github.com/NVIDIA/Fuser/blob/1e48b3752f8cfc976c83df5f31b6ddc2b12f6fc6/csrc/preseg_passes/pre_segmenter.cpp#L21 to print fusions after each pass? This way, we don't need to duplicate the printing in each individual pass.

tests/cpp/test_remove_bcast_squeeze.cpp
```diff
@@ -0,0 +1,281 @@
...
+  auto tv2 = broadcast(tv1, {false, false, true});
+  auto tv3 = squeeze(tv2, std::vector<bool>{false, false, true});
+  auto tv4 = set(tv3);
+  fusion->addOutput(tv4);
+
+  // preseg_passes should remove both broadcast and squeeze
+  preseg_passes::OptimizationPass<preseg_passes::PreSegmenter>::runPass(
+      fusion.get());
+  auto [has_bcast, has_squeeze] = hasBcastSqueeze(fusion.get());
+  ASSERT_FALSE(has_bcast);
```
wujingyue, line 79:
```suggestion
  EXPECT_FALSE(has_bcast);
```

tests/cpp/test_remove_bcast_squeeze.cpp
```diff
@@ -0,0 +1,281 @@
...
+  ASSERT_FALSE(has_squeeze);
```
wujingyue, line 80:
```suggestion
  EXPECT_FALSE(has_squeeze);
```

tests/cpp/test_remove_bcast_squeeze.cpp
wujingyue: https://google.github.io/googletest/primer.html#assertions

This way, the following code can still run regardless of this check to discover multiple failures at a time.

tests/cpp/test_remove_bcast_squeeze.cpp
```diff
@@ -0,0 +1,281 @@
...
+#include <tests/cpp/validator.h>
+
+#include <torch/torch.h>
+
+namespace nvfuser {
+
+using RemoveBcastSqueezeTest = NVFuserTest;
+
+namespace {
+std::pair<bool, bool> hasBcastSqueeze(Fusion* fusion) {
```
wujingyue, line 27: How about removing this helper in favor of https://github.com/NVIDIA/Fuser/blob/1e48b3752f8cfc976c83df5f31b6ddc2b12f6fc6/csrc/ir/utils.h#L601?

tests/cpp/test_remove_bcast_squeeze.cpp
```diff
@@ -0,0 +1,281 @@
...
+  auto tv0 = makeContigTensor(2, input_dtype);
+  fusion->addInput(tv0);
+  auto tv1 = set(tv0);
+  auto tv2 = broadcast(tv1, {false, false, true, true});
+  auto tv3 = squeeze(tv2, std::vector<bool>{false, false, false, true});
+  auto tv4 = set(tv3);
+  fusion->addOutput(tv4);
+
+  // preseg_passes shouldn't remove either broadcast or squeeze
+  // becuase broadcast dim doesn't match with squeeze dim
```
jacobhinkle, line 96: Should we consider still doing a replacement in cases like this where the flag patterns overlap but are not equal? In this case we would replace
```c++
auto tv2 = broadcast(tv1, {false, false, true, true});
auto tv3 = squeeze(tv2, std::vector<bool>{false, false, false, true});
```
with
```c++
auto tv2 = broadcast(tv1, {false, false, true, false});
```
and that would still remove a bcast+squeeze on the inner dim.

jacobhinkle (approved):
LGTM. I am wondering if there is a more general approach where we would try and move broadcast & squeeze operations in a predictable way such that we could avoid cases where simple ops are caught between these ops and prevent their removal. Broadcasts are not really needed before the point where the broadcast IterDomain is "concretized"; i.e. where it is used in a multi-input Expr aligned with an Iteration IterDomain. We can detect those concretization points and place new `BroadcastOp`s just at those locations. Note that squeeze would not be considered a concretizing op so we could probably remove any squeeze ops that do not operate on inputs. I wonder if that kind of approach would handle the cases covered by this PR. See #1778 and #340.

Also I'm just tagging #1778 as it's a related issue

csrc/preseg_passes/remove_bcast_squeeze.cpp
liqiangxl: Good point. we can do a small refactor after this PR.

csrc/preseg_passes/pre_segmenter.cpp
```diff
@@ -30,6 +31,7 @@ namespace nvfuser::preseg_passes {
   OptimizationPass<MarkAliasesPreparePass>::runPass(fusion);
   OptimizationPass<ExactMappedExtentSubstitutionPass>::runPass(fusion);
   OptimizationPass<AllocationDomainPass>::runPass(fusion);
+  OptimizationPass<RemoveBcastSqueeze>::runPass(fusion);
```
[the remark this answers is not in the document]
liqiangxl: I agree. I was thinking put this pass as the first pass, then I thought we may need a discussion on how to order these passes. It seems that the current order is based on the time they are being added.

tests/cpp/test_remove_bcast_squeeze.cpp
liqiangxl: changed to use `hasOpsOfType`

csrc/preseg_passes/remove_bcast_squeeze.cpp
liqiangxl: changed.

csrc/preseg_passes/remove_bcast_squeeze.cpp
liqiangxl: changed to use `ir_utils::replaceValInAllExprInputsAndFusionOutputs` and merged two loops into one.

csrc/preseg_passes/remove_bcast_squeeze.cpp
```diff
@@ -0,0 +1,95 @@
...
+    }
+  }
+  // step-2: find and remove squeeze + broadcast pattern
+  // before: Y = squeeze(X); Z = broadcast(Y);  M = someOp(Z)
+  // after : M = someOp(X)
+  // conditions: (1) broadcast and squeeze have the same dim flags
+  //             (2) Y is only consumed by Z (has only one consumer)
+  // special case: if Z is a fusion output, replace the output with X
+  {
+    std::unordered_map<Val*, Val*> replacement_map;
```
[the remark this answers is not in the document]
liqiangxl: changed to use ir_utils::replaceValInAllExprInputsAndFusionOutputs and merged two loops into one.

liqiangxl:
> (1)  avoid cases where simple ops are caught between these ops and prevent their removal

A following PR is planned to enable this.

> (2) Broadcasts are not really needed before the point where the broadcast IterDomain is "concretized"; i.e. where it is used in a 
multi-input Expr aligned with an Iteration IterDomain.

These broadcasts may not be considered as redundant and this seems to be another task, something like `removeAndReplaceBcast`?

> (3) We can detect those concretization points and place new `BroadcastOp`s just at those locations. Note that squeeze would not be considered a concretizing op so we could probably remove any squeeze ops that do not operate on inputs. I wonder if that kind of approach would handle the cases covered by this PR. See #1778 and #340.

I think it can do the same job plus additional remove & replace of bcast. We can achieve the same result with two steps:
(1) remvoe redundant bcast-squeeze
(2) remove and replace bcast

csrc/preseg_passes/pre_segmenter.cpp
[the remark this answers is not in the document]
jacobhinkle: Yeah, most passes are independent, so appending to the end makes sense as a default. In this case the new pass probably won't affect others but I don't think it would hurt to put it further up like in first position as you suggest.

tests/cpp/test_remove_bcast_squeeze.cpp
liqiangxl: after this there is still a broadcast dim, may not bring much benefit. so it's not removed in this PR but we can reconsider later.

liqiangxl:
!build

csrc/preseg_passes/remove_bcast_squeeze.cpp
```diff
@@ -0,0 +1,72 @@
...
+    // before: Y = broadcast(X); Z = squeeze(Y);  M = someOp(Z)
+    // after : M = someOp(X)
+    // conditions: (1) broadcast and squeeze have the same dim flags
+    //             (2) Y is only consumed by Z (has only one consumer)
+    // special case: if Z is a fusion output, replace the output with X
+    if (auto squeeze = dynamic_cast<SqueezeOp*>(expr)) {
+      if (auto bcast =
+              dynamic_cast<BroadcastOp*>(squeeze->in()->definition())) {
+        if (bcast->getBroadcastDimFlags() == squeeze->getSqueezeDimFlags() &&
+            bcast->out()->uses().size() == 1) {
```
wujingyue, line 31: I don't think this check is needed. Even when `bcast->out()` has multiple uses, we can still replace `squeeze->out()` without affecting other paths. If you agree on that, you may also want to add one more unit test for that case.

csrc/preseg_passes/remove_bcast_squeeze.cpp
```diff
@@ -0,0 +1,72 @@
...
+    // step-2: find and remove squeeze + broadcast pattern
+    // before: Y = squeeze(X); Z = broadcast(Y);  M = someOp(Z)
+    // after : M = someOp(X)
+    // conditions: (1) broadcast and squeeze have the same dim flags
+    //             (2) Y is only consumed by Z (has only one consumer)
+    // special case: if Z is a fusion output, replace the output with X
+    if (auto bcast = dynamic_cast<BroadcastOp*>(expr)) {
+      if (auto squeeze = dynamic_cast<SqueezeOp*>(bcast->in()->definition())) {
+        if (bcast->getBroadcastDimFlags() == squeeze->getSqueezeDimFlags() &&
+            squeeze->out()->uses().size() == 1) {
```
wujingyue, line 47: ditto

tests/cpp/test_remove_bcast_squeeze.cpp
```diff
@@ -0,0 +1,322 @@
...
+  auto tv5 = set(tv4);
+  fusion->addOutput(tv5);
+
+  // preseg_passes should remove the first broadcast and squeeze
+  preseg_passes::OptimizationPass<preseg_passes::PreSegmenter>::runPass(
+      fusion.get());
+  bool has_bcast = ir_utils::hasOpsOfType<BroadcastOp>(fusion.get());
+  bool has_squeeze = ir_utils::hasOpsOfType<SqueezeOp>(fusion.get());
+  EXPECT_TRUE(has_bcast);
+  EXPECT_FALSE(has_squeeze);
```
wujingyue, line 297:
```suggestion
  EXPECT_TRUE(ir_utils::hasOpsOfType<BroadcastOp>(fusion.get()));
  EXPECT_FALSE(ir_utils::hasOpsOfType<SqueezeOp>(fusion.get()));
```

The call to hasOpsOfType is pretty clear about its intention, so a local variable `has_bcast` or `has_squeeze` doesn't seem worthwhile.

wujingyue (approved):

csrc/preseg_passes/remove_bcast_squeeze.cpp
liqiangxl: agree. added `BcastSqueezeMultipleUses`

csrc/preseg_passes/remove_bcast_squeeze.cpp
liqiangxl: revised.

tests/cpp/test_remove_bcast_squeeze.cpp
liqiangxl: revised.

csrc/preseg_passes/remove_bcast_squeeze.cpp
liqiangxl: I need to roll back to use `ir_utils::replaceValue(fusion, replacement_map);` to fix the failed test case.
I added a cpp unit to reproduce the failed python test, `BcastSqueezeBcastSqueeze`.
In the following fusion, 
```
  auto tv2 = broadcast(tv1, {false, false, true});
  auto tv3 = squeeze(tv2, std::vector<bool>{false, false, true});
  auto tv4 = broadcast(tv3, {false, false, true});
  auto tv5 = squeeze(tv4, std::vector<bool>{false, false, true});
  auto tv6 = add(tv3, tv5);
```
(1) first step: replace `tv3` with `tv1` 
(2) then, replace `tv5` with `tv1`, 
(3) then, the segment error happens when dynamic_cast expr `auto tv6 = add(tv3, tv5)`
```
Caught signal 11 (Segmentation fault: address not mapped to object at address 0xfffffffffffffff0)
==== backtrace (tid: 488215) ====
 1 0x00000000000acc9d __dynamic_cast()  ???:0
 2 0x00000000006a6e82 nvfuser::preseg_passes::RemoveBcastSqueeze::runPass()  :0
```
It looks like `replaceValInExprInputs` invalidates provided Expr.
```
//! Replaces reference Val with substitute in all Expr inputs and attributes.
//! Warning: Invalidates provided Expr.
//! Warning: Removes connection of reference through provided Expr.
//! Warning: Creates new Expr defining substitute.
NVF_API Expr* replaceValInExprInputs(
    Expr* expr,
    Val* reference,
    Val* substitute);
```

wujingyue: Thanks -- that's something I overlooked. Can you try this? https://github.com/NVIDIA/Fuser/blob/25044927da7b5f4a0cf3fb16ee716dc0403b46b5/csrc/multidevice/utils.cpp#L279-L283

liqiangxl: smart solution!

liqiangxl:
!build

liqiangxl:
`jit_nvfuser_bench_17_TNVF 1/8` and `jit_nvfuser_bench_17_TNVF 2/8 ` are doing matmul bench tests, got msg: `There has been a timeout failure or the job got stuck. Check your timeout limits or try again`

naoyam:
Squeeze followed by broadcast is not no-op in terms of domain mappings. When squeezed, the squeeze input domain is not mapped with any of the domains following the operation, which is not the case if squeeze is removed. 

When should we want to remove the operation?

wujingyue:
What was the problem? Although removing squeeze+broadcast reduces the number of ops, it extends the life cycle of the IterDomain and therefore makes IdModel analysis harder?

naoyam:
Yes, though the analysis itself won't be affected.

Originally, the squeezed broadcast domain is not mapped with the broadcast domain added by the removed broadcast op, so there are basically two groups of broadcast domains. The op removal will join them.

There's no correctness issue with the removal, but since the two groups will be joined, for example, think about a fusion like:

```
auto tv0 = some_1D_input;
auto tv1 = broadcast(tv0, {false, true});
auto tv2 = some_unary_op(tv1)
auto tv3 = squeeze(tv2, {1});
auto tv4 = broadcast(tv3, {false, true});
auto tv5 = some_2D_input;
auto tv6 = add(tv4, tv5);
fusion.addOutput(tv6);
```

Here, the broadcast domain added by the first broadcast op is not concretized, so it's safe to remove, although it isn't removed yet. The second broadcast domain is concretized as it's used with a concrete 2D tensor, so we can't remove it.

With this preseg pass, the fusion would look like:

```
auto tv0 = some_1D_input;
auto tv1 = broadcast(tv0, {false, true});
auto tv2 = some_unary_op(tv1)
auto tv5 = some_2D_input;
auto tv6 = add(tv2, tv5);
fusion.addOutput(tv6);
```

Now there's only one concretized broadcast domain. 

In this simple case, I don't think there's any material difference between the two versions, but I'm feeling that mapping previously-unmapped broadcast domains may have some unwanted side effects.

So, back to my original question, why do we want to have this removal pass? The broadcast-then-squeeze patten is fairly obvious, but I'm not sure about the squeeze-then-broadcast pattern.

wujingyue:
> So, back to my original question, why do we want to have this removal pass? The broadcast-then-squeeze patten is fairly obvious, but I'm not sure about the squeeze-then-broadcast pattern.

According to https://github.com/NVIDIA/Fuser/issues/2146#issuecomment-2108942505, some squeeze-then-broadcast patterns made hasNonNormalizePostReductionBCast true and therefore were rejected by `inner_persistent`. This broke monotonicity and led to over segmentation. Btw, this is similar to failure case 2 in [this slide](https://docs.google.com/presentation/d/1eBhMQR74QfLc5NxWCWZ5HSNW6Qz19DsxFmSWVGQgP6E/edit#slide=id.g27fdd03dade_0_114).

naoyam:
@liqiangxl Is this still the issue with #2274?

PR #2293, opened 2024-05-23
Do not assume mul-sum pattern inputs are TensorView

jacobhinkle:
This was a change I made to handle casts that wound up breaking some tests and benchmarks in #2272, leading to dynamic cast errors or segfaults. The solution is to test the type of the left and right hand sides before processing the pattern matching.


jacobhinkle:
!build

jacobhinkle:
I'm looking into the multidevice failures.

zasdfgbnm (approved):

jacobhinkle:
Merging for now to mostly unblock CI then I will address the multidevice problem in another PR.

Issue #2278, opened 2024-05-21
Investigate a new SDPA IR node


Priya2698:
Add a new IR node for SDPA that is currently not supported within nvFuser. 
CC: @cowanmeg @kevinstephano

Priya2698:
Summarizing discussion from today's meeting and an offline discussion with @jjsjann123:

For training, we need two `SdpaOpFwd` and `SdpaOpBwd` nodes that respectively use `at::_scaled_dot_product_flash_attention` and `at::_scaled_dot_product_flash_attention_backward`. PR #2294 currently uses `at::scaled_dot_product_attention` that does not return any intermediate values to be stored for backward and is an inference-only node. We can merge this with `SdpaOpFwd` and potentially have a different API if we don't want to return all the outputs.

Priya2698:
References:
1. Torch implementation: https://github.com/pytorch/pytorch/blob/a6b994ed5467d4df8320cbae51cba6a98ffb139c/aten/src/ATen/native/transformers/attention.cpp#L665-L706
https://github.com/pytorch/pytorch/blob/a6b994ed5467d4df8320cbae51cba6a98ffb139c/tools/autograd/derivatives.yaml#L2806-L2829
2. Thunder SDPA executor: https://github.com/Lightning-AI/lightning-thunder/blob/2fa5cab33c88c7cc49aafbe8368bb421c9514ca3/thunder/executors/sdpaex.py#L493-L517

IvanYashchuk:
> which signature is the right one to target here?  PyTorch itself has so many different variants and each has a different signature:
https://github.com/pytorch/pytorch/blob/a6b994ed5467d4df8320cbae51cba6a98ffb139c/aten/src/ATen/native/transformers/attention.cpp#L665-L706
https://github.com/pytorch/pytorch/blob/a6b994ed5467d4df8320cbae51cba6a98ffb139c/tools/autograd/derivatives.yaml#L2806-L2829

There's no right or wrong signature, it depends on how you want to do backward computation and that would dictate the output signature for the forward function. You need to make the decision yourself what needs to be stashed for backward or could be recomputed. If you want to have fallbacks to ATen then there's no other choice than directly mimicking ATen's function signatures. It's important to remember that Flash Attention doesn't work for all input cases, the "memory efficient" one also doesn't work for all input cases.

Is flash attention kernel representable in nvFuser primitives?

jjsjann123:
[...]

PR #2294, opened 2024-05-23
Add a new `SdpaFwdOp` IR node for Flash Attention

Priya2698:
Issue #2278.
~This PR adds the new node with same functionality as `torch.nn.functional.scaled_dot_product_attention`, and enables scheduling it through `ExprEvalScheduler`.~

Based on the PR discussions, this PR is repurposed to introduce a new IR node `SdpaFwdOp` for scaled dot product flash attention forward (see #2278 for details).
This PR does not include changes to the scheduler.

The next PRs will:
1. Add mapping for producer-consumer in root_domain_map and enable this op in `ExprEvalScheduler`.
2. Python API
3. Add a node for backward pass similar to this forward node.

After the completion of these tasks, we also aim at introducing Memory Efficient Attention.

### DIFF
--- a/csrc/ir/internal_nodes.h
+++ b/csrc/ir/internal_nodes.h
@@ -2205,14 +2205,16 @@ class SdpaOp : public Expr {
  public:
   using Expr::Expr;
 
-  SdpaOp(IrBuilderPasskey, Val* out,
-    Val* query,
-    Val* key,
-    Val* value,
-    Val* attn_mask,
-    double dropout_p,
-    bool is_causal,
-    std::optional<double> scale);
+  SdpaOp(
+      IrBuilderPasskey,
+      Val* out,
+      Val* query,
+      Val* key,
+      Val* value,
+      Val* attn_mask,
+      double dropout_p,
+      bool is_causal,
+      std::optional<double> scale);
 
   NVFUSER_DECLARE_CLONE_AND_CREATE
 
@@ -2240,7 +2242,7 @@ class SdpaOp : public Expr {
   }
 
   Val* attn_mask() const {
-    if (has_mask()){
+    if (has_mask()) {
       return input(3);
     }
     return nullptr;
@@ -2250,7 +2252,7 @@ class SdpaOp : public Expr {
     return attribute<double>(0);
   }
 
-  bool is_causal() const{
+  bool is_causal() const {
     return attribute<bool>(1);
   }
   std::optional<double> scale() const {
@@ -2263,11 +2265,11 @@ class SdpaOp : public Expr {
   std::vector<PolymorphicValue> evaluate(
       const ExpressionEvaluator& ee,
       const std::vector<PolymorphicValue>& inputs) const override;
-  
-  private:
-    bool has_mask() const {
-      return inputs().size() == 4;
-    }
+
+ private:
+  bool has_mask() const {
+    return inputs().size() == 4;
+  }
 };
 
 } // namespace nvfuser
--- a/csrc/ir/internal_nodes.h
+++ b/csrc/ir/internal_nodes.h
@@ -2209,14 +2209,21 @@ class SdpaOp : public Expr {
 
   SdpaOp(
       IrBuilderPasskey,
-      Val* out,
+      TensorView* output,
+      TensorView* log_sumexp,
+      // TensorView* cum_seq_q,
+      // TensorView* cum_seq_k,
+      Val* query_seq_len,
+      Val* key_seq_len,
+      TensorView* philox_seed,
+      TensorView* philox_offset,
+      TensorView* debug_attn_mask,
       Val* query,
       Val* key,
       Val* value,
-      Val* attn_mask,
-      double dropout_p,
-      bool is_causal,
-      std::optional<double> scale);
+      Val* dropout_p,
+      Val* is_causal,
+      Val* scale);
 
   NVFUSER_DECLARE_CLONE_AND_CREATE
 
@@ -2224,12 +2231,12 @@ class SdpaOp : public Expr {
     return "SdpaOp";
   }
 
-  std::string toString(int indent_size = 0) const override;
+  // std::string toString(int indent_size = 0) const override;
   std::string toInlineString(int indent_size = 0) const override;
 
-  Val* out() const {
-    return output(0);
-  }
+  // Val* out() const {
+  //   return output(0);
+  // }
 
   Val* query() const {
     return input(0);
@@ -2243,35 +2250,23 @@ class SdpaOp : public Expr {
     return input(2);
   }
 
-  Val* attn_mask() const {
-    if (has_mask()) {
-      return input(3);
-    }
-    return nullptr;
-  }
+  // double dropout_p() const {
+  //   return attribute<double>(0);
+  // }
 
-  double dropout_p() const {
-    return attribute<double>(0);
-  }
-
-  bool is_causal() const {
-    return attribute<bool>(1);
-  }
-  std::optional<double> scale() const {
-    if (attributes().size() == 3) {
-      return attribute<double>(2);
-    }
-    return std::nullopt;
-  }
+  // bool is_causal() const {
+  //   return attribute<bool>(1);
+  // }
+  // std::optional<double> scale() const {
+  //   if (attributes().size() == 3) {
+  //     return attribute<double>(2);
+  //   }
+  //   return std::nullopt;
+  // }
 
   std::vector<PolymorphicValue> evaluate(
       const ExpressionEvaluator& ee,
       const std::vector<PolymorphicValue>& inputs) const override;
-
- private:
-  bool has_mask() const {
-    return inputs().size() == 4;
-  }
 };
 
 } // namespace nvfuser
--- a/csrc/ir/internal_nodes.h
+++ b/csrc/ir/internal_nodes.h
@@ -2203,16 +2203,16 @@ class LinearOp : public Expr {
 
 // SDPA node with same functionality as F.scaled_dot_product_attention
 // https://pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html
-class SdpaOp : public Expr {
+class SdpaFwdOp : public Expr {
  public:
   using Expr::Expr;
 
-  SdpaOp(
+  SdpaFwdOp(
       IrBuilderPasskey,
       TensorView* output,
       TensorView* log_sumexp,
-      // TensorView* cum_seq_q,
-      // TensorView* cum_seq_k,
+      TensorView* cum_seq_q,
+      TensorView* cum_seq_k,
       Val* query_seq_len,
       Val* key_seq_len,
       TensorView* philox_seed,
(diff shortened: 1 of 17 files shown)


csrc/ir/internal_nodes.h
```diff
@@ -2327,4 +2327,77 @@ class LinearOp : public Expr {
...
+  using Expr::Expr;
+
+  SdpaOp(
+      IrBuilderPasskey,
+      Val* out,
+      Val* query,
+      Val* key,
+      Val* value,
+      Val* attn_mask,
+      double dropout_p,
```
jacobhinkle, line 2343: Should we instead let `dropout_p` be a `Val*`?

csrc/ir/internal_nodes.h
```diff
@@ -2327,4 +2327,77 @@ class LinearOp : public Expr {
...
+      bool is_causal,
+      std::optional<double> scale);
```
jacobhinkle, line 2345: Same. I think this should be a scalar `Val*`, since it will commonly be a function of the input dimensions.

tests/cpp/utils.h
```diff
@@ -683,4 +684,6 @@ at::Tensor atBiasEpilogue(const at::Tensor& tensor, const at::Tensor& bias);
 // Get the number of SMs on the current device
 int64_t getNumSMs();
 
+const bool checkMapped(const ValGraph& vg, IterDomain* x, IterDomain* y);
```
jjsjann123, line 687: nitpick: a const value doesn't do anything on a function's return

csrc/ir/internal_nodes.h
jjsjann123: I agree with @jacobhinkle here.

Since we already allow things like `attn_mask` to be dynamic, we might as well also extend that to all arguments. Including `is_causal`.

IIUC, given current implementation, if we are going to want to schedule and codegen sdpa in the future, we'll have to concretize on attn_mask anyway, since the computational pattern would change based on whether or not we'll have an attn_mask. Why stop there and not extend this all the way to other arguments?

csrc/ir/internal_nodes.h
```diff
@@ -2327,4 +2327,77 @@ class LinearOp : public Expr {
   }
 };
 
+// SDPA node with same functionality as F.scaled_dot_product_attention
+// https://pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html
```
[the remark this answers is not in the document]
Priya2698: Thanks for raising this point. The current `SdpaOp` node will work only for inference.
For training: we would need to create two separate ops for forward and backward with different function signatures.

Another issue that @wujingyue pointed out is, that the current ATen evaluation does not look at whether the inputs are sharded which would need some modifications to the `evaluate` method. This is broader issue, since it applies to all other ops as well but particularly to `MatmulOp` and `LinearOp`, where we use the ATen fallback path primarily..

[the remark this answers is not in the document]
jjsjann123: Just want to point out that, for our ExprEvaluator, returning the intermediates (residuals?!) for backward isn't a big deal, but if we want to decompose and schedule it, we might want to be more careful about it.

Ideally speaking, nvfuser should be able to trim un-used outputs. So maybe we should switch, but which signature is the right one to target here? PyTorch itself has so many different variants and each has a different signature:
https://github.com/pytorch/pytorch/blob/a6b994ed5467d4df8320cbae51cba6a98ffb139c/aten/src/ATen/native/transformers/attention.cpp#L665-L706
https://github.com/pytorch/pytorch/blob/a6b994ed5467d4df8320cbae51cba6a98ffb139c/tools/autograd/derivatives.yaml#L2806-L2829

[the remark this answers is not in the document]
Priya2698: I summarized the current plan we settled on here: https://github.com/NVIDIA/Fuser/issues/2278

Priya2698:
!build

csrc/ir/nodes.cpp
```diff
@@ -4336,4 +4336,135 @@ std::vector<PolymorphicValue> LinearOp::evaluate(
...
+  indent(ss, indent_size) << attn_out()->toString() << "\n";
+  indent(ss, indent_size + 1) << " = sdpa(" << query()->toString() << ",\n";
+  indent(ss, indent_size + 1) << "          " << key()->toString() << ",\n";
+  indent(ss, indent_size + 1) << "          " << value()->toString() << ",\n";
+  indent(ss, indent_size + 1)
+      << "          dropout_p = " << dropout_p() << ",\n";
+  indent(ss, indent_size + 1) << "          is_causal = " << is_causal();
+  if (scale() != nullptr) {
+    indent(ss, indent_size + 1) << ",\n          scale = " << scale();
+  }
```
jacobhinkle, line 4390:
```suggestion
  indent(ss, indent_size + 1)
      << "          dropout_p = " << dropout_p()->toInlineString() << ",\n";
  indent(ss, indent_size + 1) << "          is_causal = " << is_causal()->toInlineString();
  if (scale() != nullptr) {
    indent(ss, indent_size + 1) << ",\n          scale = " << scale()->toInlineString();
  }
```
It's only safe to use `toInlineString()` here if these vals cannot be `TensorView`. It seems like they should always be scalar to me. Can you confirm? If so, should we have some code in the ctor that checks that I couldn't for example pass a `TensorView` for `is_causal`?

csrc/ir/nodes.cpp
```diff
@@ -4336,4 +4336,135 @@ std::vector<PolymorphicValue> LinearOp::evaluate(
...
+      [output,
+       log_sumexp,
+       cum_seq_q,
+       cum_seq_k,
+       query_seq_len,
+       key_seq_len,
+       philox_seed,
+       philox_offset,
+       debug_attn_mask] =
+          at::_scaled_dot_product_flash_attention(
```
jacobhinkle, line 4442: Do we have a plan for handling the other execution modes? I'm assuming there's no higher level op in ATen and that that logic is in the python code for PyTorch?

jacobhinkle (commented):
Is it correct to say that before accepting this in ExprEval scheduler, we need to handle the other execution modes?

Priya2698:
> Is it correct to say that before accepting this in ExprEval scheduler, we need to handle the other execution modes?

No. I mainly separated them to handle the mapping, and the ID graphs workarounds separately than the node to reduce the scope of this PR.

As we discussed in today's meeting, at the moment we only plan on supporting Flash Attention to support multi-GPU development. Once we support Flash Attention, we can revisit, if we need to add Memory Efficient Attention as well. There could be a few ways:
1. Plumbing down the backend info from Thunder and using that within our nodes: While the two implementations have different function signatures, there are overlaps and hence, one possibility is to use a superset of the inputs and outputs. The other design here would be distinct nodes for each implementation.
2. We make the decision about the backend within nvFuser using the same logic as ATen/Thunder. See: https://github.com/Lightning-AI/lightning-thunder/blob/9f0c50cc6df187cf5fd2e31240690fe2b5e9ccc1/thunder/executors/sdpaex.py#L618-L680

csrc/ir/nodes.cpp
Priya2698: What do you mean by higher level op? `at::scaled_dot_product_attention` has the backend logic within its implementation and switches between different modes. We cannot use this implementation since we need the intermediates from these underlying functions for the backward pass.

jacobhinkle:
> > Is it correct to say that before accepting this in ExprEval scheduler, we need to handle the other execution modes?
> 
> No. I mainly separated them to handle the mapping, and the ID graphs workarounds separately than the node to reduce the scope of this PR.
> 
> As we discussed in today's meeting, at the moment we only plan on supporting Flash Attention to support multi-GPU development. Once we support Flash Attention, we can revisit, if we need to add Memory Efficient Attention as well. There could be a few ways:
> 1. Plumbing down the backend info from Thunder and using that within our nodes: While the two implementations have different function signatures, there are overlaps and hence, one possibility is to use a superset of the inputs and outputs. The other design here would be distinct nodes for each implementation.
> 2. We make the decision about the backend within nvFuser using the same logic as ATen/Thunder. See: https://github.com/Lightning-AI/lightning-thunder/blob/9f0c50cc6df187cf5fd2e31240690fe2b5e9ccc1/thunder/executors/sdpaex.py#L618-L680

I dont understand how to do partial support. If we are given inputs that we cannot evaluate with flashattention, what will happen?

Priya2698:
> > > Is it correct to say that before accepting this in ExprEval scheduler, we need to handle the other execution modes?
> > 
> > 
> > No. I mainly separated them to handle the mapping, and the ID graphs workarounds separately than the node to reduce the scope of this PR.
> > As we discussed in today's meeting, at the moment we only plan on supporting Flash Attention to support multi-GPU development. Once we support Flash Attention, we can revisit, if we need to add Memory Efficient Attention as well. There could be a few ways:
> > 
> > 1. Plumbing down the backend info from Thunder and using that within our nodes: While the two implementations have different function signatures, there are overlaps and hence, one possibility is to use a superset of the inputs and outputs. The other design here would be distinct nodes for each implementation.
> > 2. We make the decision about the backend within nvFuser using the same logic as ATen/Thunder. See: https://github.com/Lightning-AI/lightning-thunder/blob/9f0c50cc6df187cf5fd2e31240690fe2b5e9ccc1/thunder/executors/sdpaex.py#L618-L680
> 
> I dont understand how to do partial support. If we are given inputs that we cannot evaluate with flashattention, what will happen?

We will only accept the op if the backend identified is Flash Attention in Thunder: https://github.com/Lightning-AI/lightning-thunder/blob/9f0c50cc6df187cf5fd2e31240690fe2b5e9ccc1/thunder/executors/sdpaex.py#L618-L680. 

Do you think this will not be sufficient?

csrc/ir/nodes.cpp
jacobhinkle: > We cannot use this implementation since we need the intermediates from these underlying functions for the backward pass.

Ah ok. It looks like that logic is available in this function https://github.com/pytorch/pytorch/blob/c3949b20a14875bf22934246404eb3cb497fb9cc/aten/src/ATen/native/transformers/sdp_utils_cpp.cpp#L60
Could we use `sdp::select_sdp_backend_cpp` here in the future to dispatch to the other modes directly?

jacobhinkle:
> Do you think this will not be sufficient?

Makes sense to me. The logic is taking place in thunder to dispatch to flash attention, which seems fine.

csrc/ir/nodes.cpp
Priya2698: Yes, that is the second method we could use if we don't want to make the decision within Thunder.

naoyam:
Is anyone still reviewing this PR? @jacobhinkle?

jacobhinkle (approved):
LGTM once this is addressed: https://github.com/NVIDIA/Fuser/pull/2294#discussion_r1626546770

Priya2698:
!build

Priya2698:
!build

naoyam:
!build

Priya2698:
!build

Priya2698:
Restarted the failed tests: https://gitlab-master.nvidia.com/dl/pytorch/fuser-gh-mirror/-/pipelines/15652567

Priya2698:
!build

Priya2698:
The failing tests look unrelated.

Issue #2253, opened 2024-05-15
IdModel buildAllGraphs() fails in presegment pass


jjsjann123:
Not sure if this is a real issue or just a mis-use of IdModel.

Here's the repro script vvv (likely it'll be rendered as obsolete after #2252)

Basically passing the fusion from this program below to `IdModel` construct triggers an assert during `buildLoopGraph()`. I can help getting a cpp test if this turns out to be a real issue. cc'ing @naoyam 
```
import torch
from nvfuser import FusionDefinition, DataType

def nvfuser_fusion_id1(fd : FusionDefinition) -> None :
    S0 = fd.define_scalar(None, dtype=DataType.Int)
    T1 = fd.define_tensor(shape=[-1, -1, -1, -1], contiguity=[None, None, True, True], dtype=DataType.BFloat16, is_cpu=False, strid
e_order=[3, 2, 1, 0])
    T2 = fd.define_tensor(shape=[-1, -1, -1, -1], contiguity=[True, True, True, True], dtype=DataType.BFloat16, is_cpu=False, strid
e_order=[3, 2, 1, 0])
    T3 = fd.define_tensor(shape=[-1, -1, -1, -1], contiguity=[True, True, True, True], dtype=DataType.BFloat16, is_cpu=False, stride_order=[3, 2, 1, 0])
    T4 = fd.define_tensor(shape=[-1, -1, -1, -1], contiguity=[None, None, True, True], dtype=DataType.BFloat16, is_cpu=False, stride_order=[3, 2, 1, 0])
    T5 = fd.define_tensor(shape=[-1, -1, 1, -1, -1], contiguity=[True, None, True, True, True], dtype=DataType.BFloat16, is_cpu=False, stride_order=[4, 1, 3, 2, 0])
    T6 = fd.ops.cast(T1, dtype=DataType.Float)
    T7 = fd.ops.cast(T4, dtype=DataType.Float)
    S8 = fd.define_scalar(0, dtype=DataType.Int)
    S9 = fd.define_scalar(32, dtype=DataType.Int)
    S10 = fd.define_scalar(32, dtype=DataType.Int)
    S11 = fd.define_scalar(4096, dtype=DataType.Int)
    S12 = fd.define_scalar(0, dtype=DataType.Int)
    V13 = fd.define_vector([S9, S10, S11, S12], dtype=DataType.Int)
    T14 = fd.ops.full(shape=V13, fill_value=S8, dtype=DataType.BFloat16)
[...]

PR #2298, opened 2024-05-24
Loop promotion fix for loop groups with view rf domains

### LINKED ISSUE #2296: Failure to find the right loop promotion with rfactor domains
Finding the right promotion of a loop group may fail when rfactor domains are involved.

```
 auto tv0 = makeConcreteTensor({5});
  fusion.addInput(tv0);
  auto tv1 = makeConcreteTensor({5, 2});
  fusion.addInput(tv1);

  auto tv2 = set(tv0);
  auto tv3 = broadcast(tv2, {false, true});
  auto tv4 = add(tv3, tv1);
  auto tv5 = reshape(tv4, {5, 2}, {10});
  fusion.addOutput(tv5);

  tv4->merge(0);
  tv3->merge(0);

  inlineMost();
```
```
TransformPrinter :
T0_g[ iS0{5} ]
 root domain : (iS0{5})
 contiguity: f
 leaf domain : (iS0{5})
T2_l[ iS3{5} ] ca_pos( 1 )
 root domain : (iS3{5})
 contiguity: t
 leaf domain : (iS3{5})
T3_l[ iS14{( 5 * 1 )} ] ca_pos( 1 ) produce_pos( 1 )
 root domain : (iS4{5}, bS5{1})
 contiguity: t n
  Merge: iS4{5} and bS5{1} -> iS14{( 5 * 1 )}
 leaf domain : (iS14{( 5 * 1 )})
T1_g[ iS1{5}, iS2{2} ]
 root domain : (iS1{5}, iS2{2})
 contiguity: f f
 leaf domain : (iS1{5}, iS2{2})
T4_l[ iS13{( 5 * 2 )} ] ca_pos( 1 ) produce_pos( 1 )
 root domain : (iS6{5}, iS7{2})
 contiguity: t t
  Merge: iS6{5} and iS7{2} -> iS13{( 5 * 2 )}
 leaf domain : (iS13{( 5 * 2 )})
T5_g[ iS12{( 5 * 2 )}rf ] ca_pos( 1 ) produce_pos( 1 )
 root domain : (iS10{5}rf, iS11{2}rf)
  Merge: iS10{5}rf and iS11{2}rf -> iS12{( 5 * 2 )}rf
 rfactor domain : (iS12{( 5 * 2 )}rf)
 contiguity: t
 leaf domain : (iS12{( 5 * 2 )}rf)
}
```

Here's what the transformation looks like:

transform
(The colors indicate exact groupings)

There's only a single loop group in this case: `idg{3 4 5 6 7 10 11 12 13 14}`

Clearly, `12` is the right promotion domain, but the current logic fails to find it because when that loop group is looked at by `findPromotionOfLoopGroup`, none of the domains is identified as a representative of all the domains in the loop group. The primary reason is because `computeCoveredGroups` [stops traversal at view rfactor domains](https://github.com/NVIDIA/Fuser/blob/main/csrc/id_model/loop_promotion.cpp#L588-L594), so the covered groups of `12` is just its exact group, i.e., `idg{12, 13}`. However, since we enumerate the covered exact groups of all of the domains of the loop group, which includes the covered groups of domains such as `14`, whose covered groups include `idg{0, 1, 3, 4, 6, 10}`, the true promotion domain, `12`, fails to "cover" all the domains in this loop group.

The idea behind the processing of view rfactor domains in `computeCoveredGroups` is that since when we do reshape transformations, all of the broadcast domains in the input tensor are first squeezed, so as long as a view rfactor domain is covered, all of the domains proceeding the rfactor domain should also be covered. While that is true, this is not enough for `findPromotionOfLoopGroup` in a case like the above. More specifically, the exact group of `12` does cover `14`, but `computeCoveredGroups` doesn't convey that information to `findPromotionOfLoopGroups`.

A straightforward fix would be not stopping the analysis at view rfactor domains.

naoyam:
Fixes #2296 

In addition to stopping at view rfactor domains as mentioned in the issue page, I also put in a quick shortcut for the case where a view rfactor domain is a terminal domain. In that case, it's guaranteed that the domain is the right promotion domain. This is a shortcut and not strictly necessary for the correctness of the analysis, but I think it's simple enough to have as a quick optimization.

### DIFF
--- a/csrc/device_lower/lower2device.cpp
+++ b/csrc/device_lower/lower2device.cpp
@@ -391,7 +391,7 @@ void GpuLower::analysis(Fusion* fusion) {
   // functionality should be affected. New IterDomains may be created,
   // so it is expected that generated code may use diffrent variable
   // names
-  if (isOptionEnabled(EnableOption::IdModel)) {
+  if (true || isOptionEnabled(EnableOption::IdModel)) {
     IdModel id_model(fusion_);
   }
 
--- a/csrc/id_model/loop_promotion.cpp
+++ b/csrc/id_model/loop_promotion.cpp
@@ -585,14 +585,6 @@ std::unordered_map<ValGroup, ValGroups> computeCoveredGroups(
       covered_ids[id_group] = {id_group};
     }
 
-    // Initialize rfactor groups
-    if (std::any_of(id_group->begin(), id_group->end(), [&](Val* id) {
-          return view_rfactor_ids.find(id->as<IterDomain>()) !=
-              view_rfactor_ids.end();
-        })) {
-      covered_ids[id_group] = {id_group};
-    }
-
     // Initialize broadcast groups to empty since broadcast domains
     // don't matter for indexing
     if (std::any_of(id_group->begin(), id_group->end(), [&](Val* id) {
@@ -668,9 +660,11 @@ IterDomain* LoopPromotionMapBuilder::findPromotionOfLoopGroup(
   // Grab all the (potentially promoted) terminal iter domains in this group.
   // Save the exact group and the iter domain in this vector.
   std::vector<std::pair<ValGroup, IterDomain*>> exact_promoted_terminal_ids;
-  for (auto loop_id : *loop_group) {
+  for (Val* loop_group_val : *loop_group) {
+    auto loop_id = loop_group_val->as<IterDomain>();
+
     // If not a terminal id in the group skip
-    if (!terminal_loop_ids.has(loop_id->as<IterDomain>())) {
+    if (!terminal_loop_ids.has(loop_id)) {
       continue;
     }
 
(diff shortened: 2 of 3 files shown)


naoyam:
!build

csrc/id_model/loop_promotion.cpp
```diff
@@ -682,6 +676,16 @@ IterDomain* LoopPromotionMapBuilder::findPromotionOfLoopGroup(
       continue;
     }
 
+    // If this domain is a view rfactor domain and a terminal domain,
+    // it is guaranteed to represent this loop group because all the
+    // domains merged into this loop_id must be non-broadcast
+    // domains. A concrete example can be found in test
```
zasdfgbnm, line 682: > because all the domains merged into this loop_id must be non-broadcast domains

Could you elaborate more in this comment? 

> when we do reshape transformations, all of the broadcast domains in the input tensor are first squeezed

naoyam: When reshaping a tensor like `{10, 1, 4}` to `{40}`, the input tensor is first squeezed to `{10, 4}`, and the actual reshape op would be `{10, 4}` to `{40}`, so we would never create a view rfactor domain of `{10*1*4}`.

zasdfgbnm: I understand. I meant that we should add this description to the code.

csrc/device_lower/lower2device.cpp
```diff
@@ -391,7 +391,7 @@ void GpuLower::analysis(Fusion* fusion) {
   // functionality should be affected. New IterDomains may be created,
   // so it is expected that generated code may use diffrent variable
   // names
-  if (isOptionEnabled(EnableOption::IdModel)) {
```
jjsjann123, line 394: @naoyam Is this change intended? #2253 is now showing up in thunder's benchmark CI.

cc'ing @IvanYashchuk

IvanYashchuk: Yep, my investigation also led to this PR. I haven't taken a look at our nightly reports and we need to check whether this problem was caught there.

PR #2299, opened 2024-05-24
Remove shift

naoyam:
Removed `shift`-related code, including `ShiftOp` and `GatherOp` IR nodes and their tests.

The existing implementation is not used and is quite dated. If these ops would ever need to be supported, we would want to think about these ops as IterDomain ops.


naoyam:
!build

naoyam:
!build

naoyam:
!build --diff --diff-bench --pybench

naoyam:
!build --diff --diff-bench --pybench

naoyam:
The codegen diffs seem to be just variable name changes.

zasdfgbnm (approved):

PR #2300, opened 2024-05-24
Compute matmul dim roles with no-devices leaf domain

jacobhinkle:
This fixes a bug introduced by #2272 in `test_multidevice` where we reject a matmul segment shaped like `[iDIDxMo, iMi, bN, iK]` for having too many M dimensions. Locally this still has a single M dimension so it is valid. This PR ignores device dims for the purposes of computing tensor roles and problem shape.

Further issues we should look into:
1. As mentioned in #2272 we should proceed to handle multiple M, N, K, and Batch dimensions, although in this case the restriction was useful for surfacing this bug.
2. Even if the matmul scheduler is completely broken or disabled, the _reduction_ scheduler should have been able to schedule this fusion. However, it identified the reduction tensor as `isResharding` and removed it from the `reduction_tvs` list, causing a failure in `scheduleReduction`. We should clean up that check to be able to schedule this type of fusion as a reduction.
3. Inside the matmul scheduler we call `canonicalizeMmaTvOrdering` which I believe still uses rfactor domain to determine domain ordering. Instead this should be updated to use dim roles that are already computed from the `MatmulPattern`.
4. The rfactor domain is often used for scheduling utilities to inspect the logical size of tensors. However, because multidevice scheduling modifies the leaf domain before segmentation, we should probably audit our schedulers to ensure they use the leaf domain and ignore device dims where necessary.
5. I should also not forget to rerun `!build` before merging PRs :sweat_smile:

### DIFF
--- a/csrc/scheduler/mma_utils.cpp
+++ b/csrc/scheduler/mma_utils.cpp
@@ -1170,10 +1170,10 @@ RolesMapOpt getTensorRoles(
 
   const auto findDims = [&dim_roles, &exact_graph](TensorView* tv) {
     DimPresence has;
-    for (IterDomain* id :
-         TensorDomain::noReductions(tv->getMaybeRFactorDomain())) {
-      if (id->isBroadcast()) {
-        // Broadcast domains won't exact map to concrete domains so skip them
+    for (IterDomain* id : TensorDomain::noReductions(tv->getLeafDomain())) {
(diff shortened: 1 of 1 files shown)


jacobhinkle:
!build --diff

cowanmeg (approved):
Thank you!

samnordmann (approved):

PR #2301, opened 2024-05-24
add shmoo tests of iter grouped block reduction

liqiangxl:
The newly added test takes about 35s on H100.
```
[ RUN      ] OuterReductionTest.IterGroupedBlockReductionShmooTests
[       OK ] OuterReductionTest.IterGroupedBlockReductionShmooTests (35704 ms)
```

Code change:
(1) The lambda test function in `IterGroupedGridReduction` is moved to a separate function to support both grid reduction & block reduction.
(2) Added `IterGroupedBlockReductionShmooTests`


liqiangxl:
!build

PR #2302, opened 2024-05-24
step-2 of revising outer reduction heuristics, use block reduction if SM usage is high

liqiangxl:
**Following** #2281 
**Issue**: block reduction has lower communication overhead and should be prioritized over grid reduction.
**Fix:** : Try to use block reduction, **see [doc](https://docs.google.com/document/d/12yBhqBHlZFKyxjgXqHjXxOfzZjsjXySHvwR0_8LsfZg/edit?usp=sharing) for details.**
**1. changes in this PR**
- Added function `maybeBlockOuterReduction` as a short-cut to prioritize using block reduction in outer reduction.
- [mechanical change] moved some code into `heuristicParaToSchedulerPara` for code reuse.
- create struct `outerReduHeuristicParas` to facilitate code reuse.

**2. block reduction heuristics**
**When to use block reduction?**

- Empirically set, reduction dim <= 1024, or
- Empirically set, SM usage >= 90%

**Iteration dim:** vect * bdimx * gdimx

- vectorization = 8, when fusion has only 1 output or small reduction reduce vectorization to increase SM usage.
- Bdimx = varies bdimx from 8 up to 1024, but ensure all SMs are used.
- gdimx = N / (bdimx * vect)

**Reduction dim:** unroll * bdimy * serial
Once iteration dim paras are set, reduction dim paras can be easily derived using:

- Unroll = max unroll / vect
- Bdimy = threads per block / bdimx
- Serial = reduction elements / (unroll * bdimy)

**3. Performance:**
**(1) CPP benchmarks** for outer reduction & gelu backward, **baseline is main branch before #2281 is merged, ensure the previous regressions in #2281 is solved**.
image
Regressions and why:
Cases with very large reduction dim and very small iteration dim. 
The revised version uses `redu unroll = 8` while the main branch uses `redu unroll = 4`. occupancy reduced from 100% to 75%
image

**(2) Heat map** test of outer reduction, **baseline is #2281, shows speedup of block reduction over grid reduction**. 
image

**(3) Heat map** test of gelu backward, **baseline is #2281, shows speedup of block reduction over grid reduction**. 
image


csrc/scheduler/reduction.cpp
```diff
@@ -507,29 +507,270 @@ std::shared_ptr<ReductionParams> innerReductionHeuristic(
...
+      buffer_reg_count * scheduler_utils::bytes_per_register /
+          (int64_t)max_input_dtype_size,
       // Reduce unrolling if we have many inputs, start reduction at 4 inputs
       scheduler_utils::lastPow2(
           std::max((int64_t)n_tensor_inputs >> 2, (int64_t)1)));
 
+  // Try block reduction
+  const int64_t max_threads_per_block = (int64_t)dev_prop->maxThreadsPerBlock;
+  std::optional<outerReduHeuristicParas> maybe_blk_hp =
+      maybeBlockOuterReduction(
```
jjsjann123, line 755: Note for myself: condition check to verify if we use block reduction.

csrc/scheduler/reduction.cpp
```diff
@@ -507,29 +507,270 @@ std::shared_ptr<ReductionParams> innerReductionHeuristic(
...
+          (int64_t)vectorize_factor,
+          device_multiprocessor_count,
+          max_unroll,
+          max_threads_per_block,
+          n_tensor_outputs,
+          total_iteration_numel,
+          total_reduction_numel);
+  if (maybe_blk_hp.has_value()) {
+    return heuristicParaToSchedulerPara(maybe_blk_hp.value());
+  }
```
jjsjann123, line 765: Note for myself: If the condition returns false here, I think we'll just go along with the old logic on block reduction...

csrc/scheduler/reduction.cpp
```diff
@@ -507,29 +507,270 @@ std::shared_ptr<ReductionParams> innerReductionHeuristic(
...
+  outerReduHeuristicParas(
+      int64_t total_iteration_numel,
+      int64_t total_reduction_numel)
+      : total_iteration_numel(total_iteration_numel),
+        total_reduction_numel(total_reduction_numel) {}
+  // iteration dim paras
+  // iteration elements = iter_unroll * bdimx * gdimx
+  int64_t iter_unroll_factor = 1;
+  int64_t bdimx = 1;
+  int64_t gidim = 1;
```
jjsjann123, line 520: is `gidim` supposed to be `gdimx`? code comment seems to be inconsistent with the variable name here.

csrc/scheduler/reduction.cpp
```diff
@@ -507,29 +507,270 @@ std::shared_ptr<ReductionParams> innerReductionHeuristic(
...
+  // reduction dim paras
+  // reduction elments = redu_unroll * bdimy * gdimy * redu_serial
+  int64_t redu_unroll_factor = 1;
+  int64_t bdimy = 1;
+  int64_t grdim = 1;
```
jjsjann123, line 525: ditto

csrc/scheduler/reduction.cpp
```diff
@@ -507,29 +507,270 @@ std::shared_ptr<ReductionParams> innerReductionHeuristic(
...
+       << "vectorize_factor: " << iter_unroll_factor << "\n"
+       << "redu_unroll_factor: " << redu_unroll_factor << "\n"
+       << "grid(" << gidim << ", " << grdim << ", 1)"
+       << "\n"
+       << "block(" << bdimx << ", " << bdimy << ", 1)" << std::endl;
+    return ss.str();
+  }
+};
+
+std::shared_ptr<ReductionParams> heuristicParaToSchedulerPara(
```
jjsjann123, line 554: If I understand correctly, this is just a mechanical change. This function is the old logic moved to a separate function so we can re-use it for block reduction short-cut.

jjsjann123 (commented):
Haven't finished yet. still need to review the actual changes in `maybeBlockOuterReduction`

csrc/scheduler/reduction.cpp
liqiangxl: yes, it is shared by the newly added block reduction heuristics in this PR and the original old heuristics.

csrc/scheduler/reduction.cpp
liqiangxl: revised.

csrc/scheduler/reduction.cpp
```diff
@@ -507,19 +507,224 @@ std::shared_ptr<ReductionParams> innerReductionHeuristic(
...
+  // Step-1, set iteration dim
+  // (1) start with bdimx = 8, gidim = 1, iter_unroll = 1
+  hp.bdimx = std::min(8L, hp.iDimAvail());
+  hp.gidim = 1;
+  hp.iter_unroll_factor = 1;
+
+  // (2) increase iter_unroll to its maximum following two rules:
+  // (2.1) ensure divisible split
+  // (2.2) leave enough blocks to saturate the device.
+  // Test on H100 shows the smallest iteration dim we can fully vectorize is
```
liqiangxl, line 647: see the [doc](https://docs.google.com/document/d/12yBhqBHlZFKyxjgXqHjXxOfzZjsjXySHvwR0_8LsfZg/edit#heading=h.qt779utn5lyh) for why 50% is used.

csrc/scheduler/reduction.cpp
```diff
@@ -507,19 +507,224 @@ std::shared_ptr<ReductionParams> innerReductionHeuristic(
...
+  int64_t sm_count_pow2 = scheduler_utils::lastPow2(sm_count);
+  while (hp.bdimy * hp.bdimx * 2 <= max_threads_per_block &&
+         hp.gidim / 2 >= sm_count_pow2) {
+    hp.bdimx *= 2;
+    hp.gidim /= 2;
+  }
+
+  // (2) Good heuristic if we have enough blocks to saturate the device or it's
+  // a small reduction. This cut-off is empirical based on H100.
+  const float min_sm_efficiency = 0.9f;
```
liqiangxl, line 704: see the [doc](https://docs.google.com/document/d/12yBhqBHlZFKyxjgXqHjXxOfzZjsjXySHvwR0_8LsfZg/edit#heading=h.rpnkwq2ts553) for why 0.9 is used.

csrc/scheduler/reduction.cpp
```diff
@@ -507,19 +507,224 @@ std::shared_ptr<ReductionParams> innerReductionHeuristic(
...
+  // (2.2) leave enough blocks to saturate the device.
+  // Test on H100 shows the smallest iteration dim we can fully vectorize is
+  // 4800, which corresponds to 75 or 56.8 % of SMs. Here 50% is used.
+  const int64_t min_blocks_fully_vectorize = sm_count / 2;
+  int64_t max_iter_unroll = vectorize_factor;
+  // only allow this adjustment when there is only one output tensor
+  // or it's a small reduction where block reduction is enforced.
+  // When has multiple outputs, reduce vectorization leads to regression,
+  // see cpp benchmark of gelu backward which has 2 outputs ( does't seem like a
+  // common usage since thunder.jit generated fusion has only 1).
```
liqiangxl, line 655: regression is attached in the [doc](https://docs.google.com/document/d/12yBhqBHlZFKyxjgXqHjXxOfzZjsjXySHvwR0_8LsfZg/edit#heading=h.yailf9wrbndl).

jjsjann123: Sorry that I might not be reading this right.

> When has multiple outputs, reduce vectorization leads to regression

Is this specific to the *number of outputs*, or rather `size of output buffer`? Since I assume your thunder repro were only outputting the reduction output?

csrc/scheduler/reduction.cpp
```diff
@@ -507,19 +507,224 @@ std::shared_ptr<ReductionParams> innerReductionHeuristic(
...
+         hp.gidim / 2 >= sm_count_pow2) {
+    hp.bdimx *= 2;
+    hp.gidim /= 2;
+  }
+
+  // (2) Good heuristic if we have enough blocks to saturate the device or it's
+  // a small reduction. This cut-off is empirical based on A100 and H100.
+  const float min_sm_efficiency = 0.9f;
+  float f_wave = (float)hp.gidim / (float)sm_count;
+  float sm_efficiency = f_wave / std::ceil(f_wave);
```
jjsjann123, line 706: Want to point out here `sm_efficiency` is different from sm usage efficiency in the doc?

`float sm_efficiency = f_wave / std::ceil(f_wave)`
Given `gidim < sm_count`. These two are the same.

I think you added the `ceil` part here just to handle when `gidim > sm_count` and you are trying to penalize when we don't have a full wave?

csrc/scheduler/reduction.cpp
```diff
@@ -507,19 +507,224 @@ std::shared_ptr<ReductionParams> innerReductionHeuristic(
   return rparams;
 }
 
+struct outerReduHeuristicParas {
```
jjsjann123, line 510: nitpick: maybe `outerReduHeuristicParas` -> `OuterReduHeuristicParas`

csrc/scheduler/reduction.cpp
```diff
@@ -507,40 +507,251 @@ std::shared_ptr<ReductionParams> innerReductionHeuristic(
...
+  // (1) revisit bdimx just in case bdimy doesn't take all the left threads
+  while (hp.bdimy * hp.bdimx * 2 <= max_threads_per_block &&
+         hp.gidim / 2 >= sm_count_pow2) {
+    hp.bdimx *= 2;
+    hp.gidim /= 2;
+  }
+  return hp;
+}
+
+outerReduHeuristicParas getGridOuterReduction(
```
liqiangxl, line 741: No algorithm change. Just move related code to this function.

csrc/scheduler/reduction.cpp
liqiangxl: `f_wave` is the required wave in theory, e.g. 2.7, `std::ceil(f_wave)` is the physically conducted waves, `which is 3 if f_wave = 2.7`. So we use all SMs 3 times but only 2.7 times are required. So the efficiency is `2.7/3.0 = 0.9`

liqiangxl:
@naoyam and @jjsjann123 
The heuristics is revised based on offline discussions. see sections-5&6 in [doc](https://docs.google.com/document/d/12yBhqBHlZFKyxjgXqHjXxOfzZjsjXySHvwR0_8LsfZg/edit#heading=h.463lwwbkdyzy) for details. The major changes are:

1. avoid using block reduction if the reduction size is larger than 16K
2. Both block reduction and grid reduction heuristics are generated and compared based on number of used blocks.

liqiangxl:
@naoyam and @jjsjann123
The heuristics is further revised based on offline discussions. see section-7 in [doc](https://docs.google.com/document/d/12yBhqBHlZFKyxjgXqHjXxOfzZjsjXySHvwR0_8LsfZg/edit#heading=h.7vnw09pg42fy) for details. The major changes are:
**Both block reduction and grid reduction heuristics are generated and compared based on number of used blocks and vectorization factor.**

jjsjann123:
sorry for the delayed response here. somehow I missed the email update on the PR. :bow:

csrc/scheduler/reduction.cpp
```diff
@@ -507,40 +507,253 @@ std::shared_ptr<ReductionParams> innerReductionHeuristic(
...
+  // (2.1) ensure divisible split
+  // (2.2) leave enough blocks to saturate the device.
+  // For example on any GPU with more than 32 SMs
+  // bdimx-vect-gidim = 8-1-32  = 256
+  // bdimx-vect-gidim = 8-2-32  = 512
+  // bdimx-vect-gidim = 8-2-64  = 1024
+  // bdimx-vect-gidim = 8-4-64  = 2048
+  // bdimx-vect-gidim = 8-4-128 = 4096
+  // bdimx-vect-gidim = 8-8-128 = 8192
+  int64_t max_iter_unroll = vectorize_factor;
```
jjsjann123, line 693: nitpick, we can just use vectorize_factor below. Is this renaming trying to make it easier to read? maybe consider renaming the argument of the function.

jjsjann123 (commented):
Looks good to me in general.

I think we somewhat agree'd on this one in our last meeting. Vaguely remember there's a question about the specific cases where we have grid reduction using less blocks, don't remember if we have resolved that already.

jjsjann123:
Overall look good to me. I'll let @naoyam do a final sanity check and stamp. :bow:

jjsjann123:
> @naoyam and @jjsjann123 The heuristics is further revised based on offline discussions. see section-7 in [doc](https://docs.google.com/document/d/12yBhqBHlZFKyxjgXqHjXxOfzZjsjXySHvwR0_8LsfZg/edit#heading=h.7vnw09pg42fy) for details. The major changes are: **Both block reduction and grid reduction heuristics are generated and compared based on number of used blocks and vectorization factor.**

I do want to point out that in the heat map shown in the doc, in this version of heuristics, we are missing out some sections with small inner dim where block reduction could be beneficial.

IIUC, that's a design decision we made hoping for a simpler heuristics.

liqiangxl:
> Looks good to me in general.
> 
> I think we somewhat agree'd on this one in our last meeting. Vaguely remember there's a question about the specific cases where we have grid reduction using less blocks, don't remember if we have resolved that already.

Here are some cases from cpp benchmarks where block reduction uses more SMs than grid reduction on A100. Grid reduction may actually degrade to block reduction when gdimy = 1, e.g. in case-2.
**case-1: outer reduction 512 x 128**, block reduction uses 16 blocks, grid reduction uses 4 x 2 blocks.
```
block reduction heuristics:
===== Outer Reduction Stats ========
total_reduction_numel: 512
total_iteration_numel: 128
vectorize_factor: 1
redu_unroll_factor: 16
grid(16, 1, 1)
block(8, 32, 1)

grid reduction heuristics:
===== Outer Reduction Stats ========
total_reduction_numel: 512
total_iteration_numel: 128
vectorize_factor: 2
redu_unroll_factor: 8
grid(4, 2, 1)
block(16, 16, 1)
```

**case-2: outer reduction 16 x 32768,** block reduction uses 86 blocks while grid reduciton uses 64 blocks (also a block reduction)
```
block reduction heuristics:
===== Outer Reduction Stats ========
total_reduction_numel: 16
total_iteration_numel: 32768
vectorize_factor: 4
redu_unroll_factor: 2
grid(86, 1, 1)
block(96, 8, 1)

grid reduction heuristics:
===== Outer Reduction Stats ========
total_reduction_numel: 16
total_iteration_numel: 32768
vectorize_factor: 4
redu_unroll_factor: 2
grid(64, 1, 1)
block(128, 4, 1)
```

csrc/scheduler/reduction.cpp
```diff
@@ -507,40 +507,253 @@ std::shared_ptr<ReductionParams> innerReductionHeuristic(
...
+       << "grid(" << gidim << ", " << grdim << ", 1)"
+       << "\n"
+       << "block(" << bdimx << ", " << bdimy << ", 1)" << std::endl;
+    return ss.str();
+  }
+
+  // compare block reduction with grid reduction
+  bool isBetterThan(const OuterReduHeuristicParas& grid_hp, int64_t sm_count)
+      const {
+    NVF_ERROR(
```
naoyam, line 556: Also assert that `grid_hp.grdim > 1`

csrc/scheduler/reduction.cpp
```diff
@@ -507,40 +507,253 @@ std::shared_ptr<ReductionParams> innerReductionHeuristic(
...
+    if (gidim * grdim >= grid_hp.gidim * grid_hp.grdim) {
+      return true;
+    }
+
+    // use grid reduction
+    return false;
+  }
+};
+
+std::shared_ptr<ReductionParams> heuristicParaToSchedulerPara(
```
naoyam, line 587: It's generally abbreviated as "params" not "paras" in our codebase. Please use the consistent naming.

csrc/scheduler/reduction.cpp
```diff
@@ -507,40 +507,253 @@ std::shared_ptr<ReductionParams> innerReductionHeuristic(
...
+       << "vectorize_factor: " << iter_unroll_factor << "\n"
+       << "redu_unroll_factor: " << redu_unroll_factor << "\n"
+       << "grid(" << gidim << ", " << grdim << ", 1)"
+       << "\n"
+       << "block(" << bdimx << ", " << bdimy << ", 1)" << std::endl;
+    return ss.str();
+  }
+
+  // compare block reduction with grid reduction
+  bool isBetterThan(const OuterReduHeuristicParas& grid_hp, int64_t sm_count)
```
naoyam, line 554:
```suggestion
  static bool isBetterThan(const OuterReduHeuristicParas& block_hp, const OuterReduHeuristicParas& grid_hp, int64_t sm_count)
```

csrc/scheduler/reduction.cpp
```diff
@@ -507,40 +507,253 @@ std::shared_ptr<ReductionParams> innerReductionHeuristic(
   return rparams;
 }
 
-std::shared_ptr<ReductionParams> outerReductionHeuristic(
+struct OuterReduHeuristicParas {
```
naoyam, line 510:
```suggestion
struct OuterReductionHeuristicParams {
```

csrc/scheduler/reduction.cpp
```diff
@@ -507,40 +507,253 @@ std::shared_ptr<ReductionParams> innerReductionHeuristic(
...
+  }
+
+  // compare block reduction with grid reduction
+  bool isBetterThan(const OuterReduHeuristicParas& grid_hp, int64_t sm_count)
+      const {
+    NVF_ERROR(
+        grdim == 1,
+        "Only support compare block reduction heuristic with grid reduction not vice versa");
+
+    // use block reduction if its SM usage >= 90% and its iter_unroll_factor is
```
naoyam, line 560: Please leave some more detailed background on why these conditions should be used. For example, if this was the first time I saw this logic, I'd wonder why the SM efficiency and unrolling factor of the block reduction were considered.

csrc/scheduler/reduction.cpp
```diff
@@ -507,40 +507,253 @@ std::shared_ptr<ReductionParams> innerReductionHeuristic(
...
+    // use a lower threshold. For computation bound (e.g. gelu bwd), relaxing
+    // the threshold leads to regression.
+    float f_wave = (float)gidim / (float)sm_count;
+    float sm_efficiency = f_wave / std::ceil(f_wave);
+    if (sm_efficiency >= 0.9f &&
+        iter_unroll_factor >= grid_hp.iter_unroll_factor) {
+      return true;
+    }
+
+    // prefer block reduction if it uses same or more blocks than grid
```
naoyam, line 575: If the two options used the same parameters except for the choice of block or grid reductions, the grid reduction should never use fewer blocks. Please put some comments why that's not the case.

naoyam:
> > Looks good to me in general.
> > I think we somewhat agree'd on this one in our last meeting. Vaguely remember there's a question about the specific cases where we have grid reduction using less blocks, don't remember if we have resolved that already.
> 
> Here are some cases from cpp benchmarks where block reduction uses more SMs than grid reduction on A100. Grid reduction may actually degrade to block reduction when gdimy = 1, e.g. in case-2. **case-1: outer reduction 512 x 128**, block reduction uses 16 blocks, grid reduction uses 4 x 2 blocks.
> 
> ```
> block reduction heuristics:
> ===== Outer Reduction Stats ========
> total_reduction_numel: 512
> total_iteration_numel: 128
> vectorize_factor: 1
> redu_unroll_factor: 16
> grid(16, 1, 1)
> block(8, 32, 1)
> 
> grid reduction heuristics:
> ===== Outer Reduction Stats ========
> total_reduction_numel: 512
> total_iteration_numel: 128
> vectorize_factor: 2
> redu_unroll_factor: 8
> grid(4, 2, 1)
> block(16, 16, 1)
> ```

This seems to indicate we should change the grid reduction heuristics. It should not vectorize if it the number of blocks is such small and thus can't saturate the device. That would put both heuristics to use the same number of blocks.

> 
> **case-2: outer reduction 16 x 32768,** block reduction uses 86 blocks while grid reduciton uses 64 blocks (also a block reduction)
> 
> ```
> block reduction heuristics:
> ===== Outer Reduction Stats ========
> total_reduction_numel: 16
> total_iteration_numel: 32768
> vectorize_factor: 4
> redu_unroll_factor: 2
> grid(86, 1, 1)
> block(96, 8, 1)
> 
> grid reduction heuristics:
> ===== Outer Reduction Stats ========
> total_reduction_numel: 16
> total_iteration_numel: 32768
> vectorize_factor: 4
> redu_unroll_factor: 2
> grid(64, 1, 1)
> block(128, 4, 1)
> ```

In the case of the grid reduction, `redu_unroll_factor * gridDim.y * blockDim.y = 8`. Shouldn't this be equal to `total_reduction_numel`?

csrc/scheduler/reduction.cpp
```diff
@@ -507,40 +507,253 @@ std::shared_ptr<ReductionParams> innerReductionHeuristic(
...
+
+OuterReduHeuristicParas getGridOuterReduction(
     const int64_t total_reduction_numel,
     const int64_t total_iteration_numel,
     const int64_t n_tensor_inputs,
     const int64_t max_input_dtype_size,
+    const int64_t max_unroll,
     const size_t vectorize_factor) {
-  // WARNING: Current device for codegen may not be the target device
+  auto dev_prop = at::cuda::getCurrentDeviceProperties();
```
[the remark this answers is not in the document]
liqiangxl: revised.

csrc/scheduler/reduction.cpp
```diff
@@ -763,81 +976,71 @@ std::shared_ptr<ReductionParams> outerReductionHeuristic(
...
-      rparams->multiple_reds_per_blk ? bdimy : LaunchParams::UNINITIALIZED_VAL,
-      LaunchParams::UNINITIALIZED_VAL);
+  // block reduction heuristic
+  auto block_hp = getBlockOuterReduction(
+      (int64_t)vectorize_factor,
+      device_multiprocessor_count,
+      max_unroll,
+      (int64_t)dev_prop->maxThreadsPerBlock,
+      total_iteration_numel,
+      total_reduction_numel);
```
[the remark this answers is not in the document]
liqiangxl: (1) revised function paras
(2) collected required properties and passed to `getGridOuterReduction` and `getBlockOuterReduction`

csrc/scheduler/reduction.cpp
liqiangxl: just to keep consistent with other schedulers, will revise when refactor/clean up reduction para names.

csrc/scheduler/reduction.cpp
liqiangxl: will add this assert after `getGridOuterReduction` is revised. Currently, grid reduction may use 1 block in reduction dim and degrade to block reduction.

csrc/scheduler/reduction.cpp
liqiangxl: revised comments:
```
  // use block reduction if its SM usage >= 90% and its iter_unroll_factor is
  // equal or larger than grid reduction. These two conditions ensure high SM
  // usage and efficient global memory access. The corresponding block reduction
  // avoids the overhead of inter-block data exchange through global memory.
  // It is faster than grid reduction even not all SMs are used.
  // TODO: if we know the fusion is memory bound (e.g. pure reduction), we can
  // use a lower threshold. For computation bound (e.g. gelu bwd), relaxing
  // the threshold leads to regression.
```

csrc/scheduler/reduction.cpp
liqiangxl: revised comments:
```
  // prefer block reduction if it uses same or more blocks than grid
  // reduction. this is to avoid bad grid reduction heuristic that uses very
  // few blocks. Current grid reduction heuristic may use less blocks when input
  // size is small since it uses more threads per block and prioritize
  // vectorization.
  // TODO: ensure grid reduction uses enough blocks to saturate the device.
```

liqiangxl:
> In the case of the grid reduction, `redu_unroll_factor * gridDim.y * blockDim.y = 8`. Shouldn't this be equal to `total_reduction_numel`?

Current grid reduction heuristic leaves at least 2 serial loop on top of unrolls to balance workload per block and number of blocks. It was `4` but we changed to `2` in the first step of heuristic revision where `unroll * vect` is changed from 8 to 16 for fp16. This change ensures workload per thread remains unchanged. (4 x 8 = 2 x 16)

liqiangxl:
> This seems to indicate we should change the grid reduction heuristics. It should not vectorize if it the number of blocks is such small and thus can't saturate the device. That would put both heuristics to use the same number of blocks.

I'll twick the grid reduction heuristics in a separate PR to prioritize using all SMs.

liqiangxl:
Further twick of grid reduction is added in [PR-2681](https://github.com/NVIDIA/Fuser/pull/2681)

csrc/scheduler/reduction.cpp
```diff
@@ -507,40 +507,255 @@ std::shared_ptr<ReductionParams> innerReductionHeuristic(
...
+    return true;
+  }
+
+  // prefer block reduction if it uses same or more blocks than grid
+  // reduction. this is to avoid bad grid reduction heuristic that uses very
+  // few blocks. Current grid reduction heuristic may use less blocks when input
+  // size is small since it uses more threads per block and prioritize
+  // vectorization.
+  // TODO: ensure grid reduction uses enough blocks to saturate the device.
+  if (block_hp.gidim * block_hp.grdim >= grid_hp.gidim * grid_hp.grdim) {
```
liqiangxl, line 583: add more comments.

liqiangxl:
!build

csrc/scheduler/reduction.cpp
naoyam: Please at least mention:

- This is really a tiny problem. The performance impact is most likely in a range of a few us in <10us kernels
- This condition is a WAR. Ideally, the grid reduction heuristics should be improved, but given that the impact is likely negligible, we decided to do this quick adjustment.

liqiangxl: Thanks. I'll add these comments after CI is done.

naoyam (approved):
LGTM

PR #2303, opened 2024-05-24
Infer matmul dimension roles to compute vectorization

### LINKED ISSUE #2169: Analyze matmul epilogue to determine vectorization

jacobhinkle:
This PR does the following:
1. Rename `RolesMap` to `TensorRolesMap` and introduce `DimRolesMap` which is a mapping from `ValGroup` to `MatmulDomain`.
2. Compute a canonical dim ordering on `ValGroup`s based on allocation domains of inputs and outputs. This is used to compute vectorization properly but can be used for canonicalization of loop domains in scheduleMatmul in a future PR.
3. Properly infer vectorization for every operand, epilogue input, and output based on canonical dim ordering.

This is in preparation for further generalization to accomodate multiple MmaOps in a single Fusion.

Fixes #2169.


jacobhinkle:
!build

jacobhinkle:
!build

jacobhinkle:
!build

jacobhinkle:
!build --diff

jacobhinkle:
I think the codediffs are just due to the uncommented tests.

zasdfgbnm:
What if my mma output is (M=1024, N=1024), and my epilogue is:
```
T1[1024, 128, 2, 2', 2"] = view(mma_output);
T2[1024, 128, 2', 2, 2"] = transpose(T1);
T3[1024, 1024] = view(T2);
output T3
```

jacobhinkle:
> What if my mma output is (M=1024, N=1024), and my epilogue is:
> 
> ```
> T1[1024, 128, 2, 2', 2"] = view(mma_output);
> T2[1024, 128, 2', 2, 2"] = transpose(T1);
> T3[1024, 1024] = view(T2);
> output T3
> ```

In that case, I would expect T3 to not be mapped as an `OUTPUT_D` tensor, in which case we currently would refuse to segment this because we do not allow tensors without known roles. In the future, we could try and allow things like this by loosening the restriction that output dims must exact map to M, N, K or Batch dims. Then we could do something like what @protonu did for #2315; Use unmapped dims to traverse backward through producers to find mapped dimensions and determine vectorization. In that case we could accept since the inner dimension of T3 corresponds to a shuffled N dimension, and we'd select vectorization width of 2 for T3. This particular example would also require us to accept epilogues with `ViewOp` in them, which we currently block, but it's at least feasible I think.

zasdfgbnm:
Oh, I see. We are effectively rejecting view ops inside the fusion. And this rejection saves us from having to use the vectorization helper.

csrc/scheduler/mma_utils.cpp
```diff
@@ -1731,6 +1729,96 @@ std::unordered_map<ValGroup, MatmulDomain> MatmulPattern::getDimRoles(
   return dim_roles;
 }
 
+std::vector<ValGroup> canonicalDimOrdering(
+    const mma_utils::TensorRolesMap& tensor_roles,
+    const mma_utils::DimRolesMap& dim_roles,
+    const ValGraph& exact_graph) {
```
zasdfgbnm, line 1735: In the exact graph, broadcasting IDs are not mapped to concrete IDs. So how will the result of this function look like, will it be like `(batch, broadcasting M, non-broadcasting M, broadcasting N, non-broadcasting N, K)`? Could you write you reply to this question as a comment in the code?

csrc/scheduler/matmul_utils.cpp
```diff
@@ -314,144 +314,302 @@ std::string isMatmulFusionDefinitionSupported(
...
+      const mma_utils::DimRolesMap& dim_roles,
+      const ValGraph& exact_graph,
+      SchedulerRuntimeInfo& runtime_info)
+      : runtime_info_(runtime_info),
+        tensor_roles_(tensor_roles),
+        dim_roles_(dim_roles),
+        exact_graph_(exact_graph) {
+    dim_ordering_ =
+        mma_utils::canonicalDimOrdering(tensor_roles, dim_roles_, exact_graph_);
   }
```
zasdfgbnm, line 330: nit:
```suggestion
        exact_graph_(exact_graph),
    dim_ordering_ (
        mma_utils::canonicalDimOrdering(tensor_roles, dim_roles_, exact_graph_))
  {}
```

csrc/scheduler/matmul_utils.cpp
```diff
@@ -314,144 +314,302 @@ std::string isMatmulFusionDefinitionSupported(
...
+
+  // Note this is non-const because we use runtime_info_.expressionEvaluator()
+  std::pair<std::vector<int64_t>, std::vector<int64_t>> getSizesAndStrides(
+      TensorView* tv) {
+    if (tv->isFusionInput()) {
+      return {
+          runtime_info_.getInputAllocationSizes(tv),
+          runtime_info_.getInputAllocationStrides(tv)};
+    }
+    // For non-inputs, compute sizes using ExpressionEvaluator, then compute
```
zasdfgbnm, line 377: Am I understanding correctly that we can do the following?
```C++
NVF_ERROR(tv->isFusionOutput());
```

csrc/scheduler/matmul_utils.cpp
```diff
@@ -314,144 +314,302 @@ std::string isMatmulFusionDefinitionSupported(
...
+    // strides based on allocation domain, assuming full contiguity regardless
+    // of how it is marked in the TensorView.
```
zasdfgbnm, line 379: Could you explain more (in the code as comment) why assuming full contiguity?

csrc/scheduler/matmul_utils.cpp
```diff
@@ -314,144 +314,302 @@ std::string isMatmulFusionDefinitionSupported(
...
+      }
+
+      ValGroup g = exact_graph_.toGroup(id);
+      MatmulDomain dim_role = dimRole(g);
+      if (dim_role == MatmulDomain::Batch) {
+        // We cannot vectorize in batch dimensions
+        break;
+      }
+      if (!vec_dim_role.has_value()) {
+        vec_dim_role = dim_role;
```
zasdfgbnm, line 516: should we do a `break;` after this line?

csrc/scheduler/matmul_utils.cpp
```diff
@@ -314,144 +314,302 @@ std::string isMatmulFusionDefinitionSupported(
...
+
+    // Extract dims with this role in the canonical ordering
+    std::vector<ValGroup> ordered_inner_dims;
+    for (const ValGroup& og : dim_ordering_) {
+      if (dimRole(og) == vec_dim_role.value()) {
+        ordered_inner_dims.push_back(og);
+      }
+    }
+
+    return std::min(vec_size, innerDimsVectorization(tv, ordered_inner_dims));
```
zasdfgbnm, line 532: Should we also check the strides of inner dim is 1, and the non-inner dim? Merged inner dims are considered as a single inner dim.

csrc/scheduler/matmul_utils.cpp
```diff
@@ -314,144 +314,302 @@ std::string isMatmulFusionDefinitionSupported(
...
+      // grid.
+      return 1l;
+    }
+
+    // Match the innermost dimensions above to contiguous innermost dims in tv
+    // from inner to outer. Determine supported vectorization based on product
+    // of matching sizes along with all outer strides.
+    const auto innerMostVec = [&](TensorView* tv) {
+      return std::min(
+          ptrAndDTypeVec(tv), innerDimsVectorization(tv, inner_nonk_dims));
```
zasdfgbnm, line 570: Should we exclude batch from the argument of `innerDimsVectorization`?

zasdfgbnm: Also, should we check strides?

csrc/scheduler/matmul_utils.cpp
```diff
@@ -314,144 +314,302 @@ std::string isMatmulFusionDefinitionSupported(
...
+    };
+
+    const auto d_it = tensor_roles_.find(MatmulRole::OUTPUT_D);
```
zasdfgbnm, line 573: Could you remind me that, are C and D all fusion outputs? How are C and D different? Could we just consider them as the same role, as we are computing their vectorization factor equally?

csrc/scheduler/mma_utils.h
```diff
@@ -370,6 +374,25 @@ int64_t computeExpectedSharedMemoryUsage(
 //!  Z = complex<double>
 char dtypeToChar(const DataType& dtype);
 
+//! Get a total ordering of dimensions for known tensors. All dims of a
+//! particular DimRole are adjacent in the output. We then set the order as
+//! follows:
+//! 1. Batch dimensions go first
+//! 2. K dimensions are innermost
+//! 3. M or N can be innermost, depending on the first output's allocation
```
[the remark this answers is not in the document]
jacobhinkle: > Oh, N don't have to be innermost?! Are we already supporting matmul-transpose fusion? In that case, we need to flip A and B based on the math $(AB)^T = B^T A^T$. Could you add a test for that if you are supporting it (let's do this in a separate PR).

That's the intention but i havent done the swap of the MmaOp inputs yet

csrc/scheduler/mma_utils.cpp
```diff
@@ -1171,7 +1171,7 @@ RolesMapOpt getTensorRoles(
 
   const auto findDims = [&dim_roles, &exact_graph](TensorView* tv) {
     DimPresence has;
-    for (IterDomain* id : TensorDomain::noReductions(tv->getLeafDomain())) {
+    for (IterDomain* id : TensorDomain::noReductions(tv->getRFactorDomain())) {
```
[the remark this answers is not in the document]
jacobhinkle: Correct. It should make no difference in the usual case.

csrc/scheduler/mma_utils.cpp
jacobhinkle: Good question. I have so far only used it to determine which concrete dims are innermost in which case we can ignore broadcasts. However, when we actually use this to do the reordering it will be important so I will change it to use the permissive map, and add a comment explaining this.

csrc/scheduler/matmul_utils.cpp
jacobhinkle: `INPUT_C` is a fusion input (bias) and `OUTPUT_D` is an output. I think it would be good to abandon the letter-based naming and switch to `OPERAND`, `EPILOGUE_INPUT` and `OUTPUT`.

csrc/scheduler/matmul_utils.cpp
jacobhinkle: Good idea. We should only be calling this for inputs or outputs, so I will add this assertion.

csrc/scheduler/matmul_utils.cpp
jacobhinkle: I changed this to respect the contiguity flags. Originally, I was mimicking how `FusionExecutor` allocates outputs: as fully contiguous regardless of these flags. However, we are now actually able to pass preallocated outputs (soon to be fully plumbed in, see #2247). The tensors we pass should be compatible with the allocation domain and contiguity marked in the `TensorView`, but are otherwise unconstrained. I updated the code to choose odd strides for discontiguous dimensions in case of preallocated outputs. We don't have a way to inspect the pointer alignment of such preallocated outputs but I put another comment in the code that this should be checked.

csrc/scheduler/matmul_utils.cpp
jacobhinkle: Yes!

jacobhinkle: Yes.

csrc/scheduler/matmul_utils.cpp
jacobhinkle: Strides are checked in `innerDimsVectorization`. We actually only vectorize dims if the `TensorView` has those dimensions marked contiguous, but we do check that the strides of all those vectorized dims match the expected contiguous values. This includes the case of a single inner-most dim, in which case we check that the stride is 1: https://github.com/NVIDIA/Fuser/blob/6f9f79c68d931734485e1e873b2ee2b5c80dc552/csrc/scheduler/matmul_utils.cpp#L476-L484

csrc/scheduler/matmul_utils.cpp
jacobhinkle: > Should we exclude batch from the argument of `innerDimsVectorization`?

Just above this we return 1 if `inner_nonk_role` is Batch. The `inner_nonk_dims` all have this role, and since it's not K, then all `inner_nonk_dims` must be M or N for this call.

> Also, should we check strides?

Strides are checked in `innerDimsVectorization` but not beyond dimensions marked contiguous in the `TensorView`. See my reply above.

csrc/scheduler/mma_utils.h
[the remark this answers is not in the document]
jacobhinkle: I could assert that N is innermost for now, or perhaps I could put a check in `isMatmulFusionDefinitionSupported` until this is supported. Asserting is probably fine though since the most common paths will all have N innermost. Since we don't support transposes in the epilogue I think we'd need to manually set the allocation domain on the output to specify a column-major output.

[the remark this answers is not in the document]
jacobhinkle: Added assertion that N is innermost. Will revisit in another PR.

csrc/scheduler/matmul_utils.cpp
zasdfgbnm: For the dims not contiguous, should we use its stride value to `std::min` the vectorization size? For example, what if a dimension has stride 1331? This case, we can not vectorize because the address will not always be aligned.

jacobhinkle:
!build

csrc/scheduler/matmul_utils.cpp
zasdfgbnm: > Just above this we return 1 if `inner_nonk_role` is Batch. The `inner_nonk_dims` all have this role, and since it's not K, then all `inner_nonk_dims` must be M or N for this call.

I see, somehow I missed that the `inner_nonk_dims` contain dims of only a single role.


> Strides are checked in `innerDimsVectorization` but not beyond dimensions marked contiguous in the `TensorView`. See my reply above.

See my reply above. I believe we need to use these strides to calculate vector size.

jacobhinkle: > See my reply above. I believe we need to use these strides to calculate vector size.

Oh OK I see what you're saying. Yes, `innerDimsVectorization` does check the outer strides. How it works is we look through the provided dimensions and find the most of those that we can vectorize. We look at the combined size to compute an initial vectorization. Then we start at that position and look at all outer strides starting at that location, which might potentially reduce the vectorization width if they are not multiples of the vectorization.

csrc/scheduler/matmul_utils.cpp
jacobhinkle: Outer dims are handled just as you say later on in this function: https://github.com/NVIDIA/Fuser/blob/6f9f79c68d931734485e1e873b2ee2b5c80dc552/csrc/scheduler/matmul_utils.cpp#L503-L511

csrc/scheduler/matmul_utils.cpp
zasdfgbnm: Oh, I see. I missed that. Find it now.

zasdfgbnm (approved):

PR #2304, opened 2024-05-24
remove canOmitStopPredicate

naoyam:


naoyam:
!build --diff --diff-bench --pybench

csrc/index_compute.cpp
```diff
@@ -2998,90 +2998,6 @@ std::pair<Val*, Val*> getStartAndStopOffsets(
   return {start_offset, stop_offset};
 }
 
-bool canOmitStopPredicate(
-    Val* stop_index,
-    Val* stop_offset,
-    IterDomain* contig_id) {
-  bool index_simple = stop_index->definition() == nullptr;
```
naoyam, line 3005: This logic is actually no longer correct. It used to be a valid assumption. If a `Val` has no definition, it means the corresponding domain (i.e., `contig_id`) has no merge/split, thus the index is guaranteed to be less than the extent, except for a few conditions below. However, the expression simplification seems to create a `Val` with no definition even for non-simple domains (please confirm, @zasdfgbnm), so this assumption doesn't seem to hold anymore.

Luckily, I think this hasn't shown up as a real problem due to a bug at line 3063, where the condition should be `> 0` instead of `>= 0`. Due to that, even when `index_simple` is incorrectly identified as `true`, this function itself never returns `true`. 

Since the same predicate removal can be done with the expression simplification, this function itself is no longer necessary, so I think just removing it would make most sense.

zasdfgbnm: > However, the expression simplification seems to create a `Val` with no definition even for non-simple domains (please confirm, @zasdfgbnm), so this assumption doesn't seem to hold anymore.

I agree. Whether a `Val` has definition or not should not be used as a criteria for whether the ID is transformed.  The pattern-match based simplifier `SimplifyingIrBuilder` can make something that "should" have definition not having it, for example `x + 0 -> x`. On the other hand, I looked at the code above, seems that we are setting the offset as `--extent`, which should be simplified as `extent`. However, if we fail to simplify it for some reason, this result of this condition will also change.

zasdfgbnm (approved):

PR #2309, opened 2024-05-28
Handling allocation domain of the input TensorViews in the matmul scheduler 

protonu:
In this PR we extend the matmul scheduler to support inputs with allocation domains.

To the fusion (with inputs tv_a and tv_b), we add two LoadStoreOps to both inputs.
The first Op corresponds to a load to shared memory, where we propagate the allocation domain. The second op corresponds to reading to registers, where we don't propagate the allocation domain since the scheduler takes charge of setting the allocation domain in the registers. Based on the difference in the (maybe)allocation domain of the producer and consumer of the second LoadStoreOp, we may do transposed load when reading to registers.

image

See also https://github.com/NVIDIA/Fuser/pull/2315.

### DIFF
--- a/csrc/scheduler/matmul.cpp
+++ b/csrc/scheduler/matmul.cpp
@@ -734,11 +734,15 @@ void scheduleSplitKSum(
   splitk_sum->axis(-1)->parallelize(ParallelType::Vectorize);
 }
 
-bool hasInnerTranspose(TensorView* out_tv) {
-  auto use_root_or_alloc = out_tv->hasAllocation()
-      ? out_tv->getAllocationDomain()
-      : out_tv->getRootDomain();
-  return use_root_or_alloc.back() != out_tv->getMaybeRFactorDomain().back();
+bool needsTranposedLoad(TensorView* out_tv) {
+  if (out_tv->hasRFactor()) {
+    auto use_root_or_alloc = out_tv->hasAllocation()
+        ? out_tv->getAllocationDomain()
+        : out_tv->getRootDomain();
+    return use_root_or_alloc.back() != out_tv->getRFactorDomain().back();
+  }
+  return out_tv->getRootDomain().back() !=
+      out_tv->getMaybeAllocationDomain().back();
 }
 
 } // namespace
@@ -895,24 +899,18 @@ void scheduleMatmul(Fusion* fusion, const MatmulParams& params) {
   bcw_smem->definition()->as<LoadStoreOp>()->setCacheOp(cache_op_b);
   NVF_ERROR(acw_smem->uses().size() == 1);
   NVF_ERROR(bcw_smem->uses().size() == 1);
-  if (auto ldst = dynamic_cast<LoadStoreOp*>(acw_smem->uses().at(0))) {
-    acr = ldst->out()->as<TensorView>();
-    ldst->setOpType(LoadStoreOpType::LdMatrix);
-  } else {
-    if (hasInnerTranspose(acw_smem)) {
-      acr = acw_smem->cacheAfter(LoadStoreOpType::LdMatrixTranspose);
+
+  for (auto [tv_smem, tv_r] :
+       {std::make_pair(acw_smem, &acr), std::make_pair(bcw_smem, &bcr)}) {
+    if (auto ldst = dynamic_cast<LoadStoreOp*>(tv_smem->uses().at(0))) {
+      *tv_r = ldst->out()->as<TensorView>();
+      ldst->setOpType(
+          needsTranposedLoad(*tv_r) ? LoadStoreOpType::LdMatrixTranspose
+                                   : LoadStoreOpType::LdMatrix);
     } else {
-      acr = acw_smem->cacheAfter(LoadStoreOpType::LdMatrix);
-    }
-  }
-  if (auto ldst = dynamic_cast<LoadStoreOp*>(bcw_smem->uses().at(0))) {
-    bcr = ldst->out()->as<TensorView>();
-    ldst->setOpType(LoadStoreOpType::LdMatrix);
-  } else {
-    if (hasInnerTranspose(bcw_smem)) {
-      bcr = bcw_smem->cacheAfter(LoadStoreOpType::LdMatrixTranspose);
-    } else {
-      bcr = bcw_smem->cacheAfter(LoadStoreOpType::LdMatrix);
+      *tv_r = tv_smem->cacheAfter(
+          needsTranposedLoad(tv_smem) ? LoadStoreOpType::LdMatrixTranspose
+                                      : LoadStoreOpType::LdMatrix);
     }
   }
 
--- a/csrc/scheduler/matmul.cpp
+++ b/csrc/scheduler/matmul.cpp
@@ -734,6 +734,13 @@ void scheduleSplitKSum(
   splitk_sum->axis(-1)->parallelize(ParallelType::Vectorize);
 }
 
+bool hasInnerTranspose(TensorView* out_tv) {
+  auto use_root_or_alloc = out_tv->hasAllocation()
+      ? out_tv->getAllocationDomain()
+      : out_tv->getRootDomain();
+  return use_root_or_alloc.back() != out_tv->getMaybeRFactorDomain().back();
+}
+
 } // namespace
 
 void scheduleMatmul(Fusion* fusion, const MatmulParams& params) {
@@ -892,13 +899,21 @@ void scheduleMatmul(Fusion* fusion, const MatmulParams& params) {
     acr = ldst->out()->as<TensorView>();
     ldst->setOpType(LoadStoreOpType::LdMatrix);
   } else {
-    acr = acw_smem->cacheAfter(LoadStoreOpType::LdMatrix);
+    if (hasInnerTranspose(acw_smem)) {
+      acr = acw_smem->cacheAfter(LoadStoreOpType::LdMatrixTranspose);
+    } else {
+      acr = acw_smem->cacheAfter(LoadStoreOpType::LdMatrix);
+    }
   }
   if (auto ldst = dynamic_cast<LoadStoreOp*>(bcw_smem->uses().at(0))) {
     bcr = ldst->out()->as<TensorView>();
     ldst->setOpType(LoadStoreOpType::LdMatrix);
   } else {
-    bcr = bcw_smem->cacheAfter(LoadStoreOpType::LdMatrix);
+    if (hasInnerTranspose(bcw_smem)) {
+      bcr = bcw_smem->cacheAfter(LoadStoreOpType::LdMatrixTranspose);
+    } else {
+      bcr = bcw_smem->cacheAfter(LoadStoreOpType::LdMatrix);
+    }
   }
 
   // Make a CTA tile
--- a/csrc/scheduler/matmul.cpp
+++ b/csrc/scheduler/matmul.cpp
@@ -734,19 +734,14 @@ void scheduleSplitKSum(
   splitk_sum->axis(-1)->parallelize(ParallelType::Vectorize);
 }
 
-// A LoadMatrixTranspose is needed is there's a rfactor which doesn't
-// match the alloc/root domain. If there's no rfactor, then we
-// need a transpose if the alloc domain and root don't match. When we say
-// match, we refer to the innermost iter domain.
-bool needsTranposedLoad(TensorView* out_tv) {
-  if (out_tv->hasRFactor()) {
-    auto use_root_or_alloc = out_tv->hasAllocation()
-        ? out_tv->getAllocationDomain()
-        : out_tv->getRootDomain();
-    return use_root_or_alloc.back() != out_tv->getRFactorDomain().back();
-  }
-  return out_tv->getRootDomain().back() !=
-      out_tv->getMaybeAllocationDomain().back();
+bool needsTranposedLoad(TensorView* producer, TensorView* consumer) {
+  const auto map =
+      PairwiseRootDomainMap(producer, consumer).mapProducerToConsumer();
+  auto maybeProducerAlloc = producer->getMaybeAllocationDomain();
+  auto maybeConsumerRFactor = consumer->getMaybeRFactorDomain();
+  auto prodToconsumerAllocDomInner =
+      map.find(maybeProducerAlloc.back())->second;
+  return maybeConsumerRFactor.back() != prodToconsumerAllocDomInner;
 }
 
 } // namespace
@@ -905,15 +900,18 @@ void scheduleMatmul(Fusion* fusion, const MatmulParams& params) {
 
   for (auto [tv_smem, tv_r] :
        {std::make_pair(acw_smem, &acr), std::make_pair(bcw_smem, &bcr)}) {
+    auto producer = tv_smem;
+    auto consumer = tv_smem->uses().at(0)->output(0)->as<TensorView>();
+    auto toTranspose = needsTranposedLoad(producer, consumer);
     if (auto ldst = dynamic_cast<LoadStoreOp*>(tv_smem->uses().at(0))) {
       *tv_r = ldst->out()->as<TensorView>();
       ldst->setOpType(
-          needsTranposedLoad(*tv_r) ? LoadStoreOpType::LdMatrixTranspose
-                                    : LoadStoreOpType::LdMatrix);
+          toTranspose ? LoadStoreOpType::LdMatrixTranspose
+                      : LoadStoreOpType::LdMatrix);
     } else {
       *tv_r = tv_smem->cacheAfter(
-          needsTranposedLoad(tv_smem) ? LoadStoreOpType::LdMatrixTranspose
-                                      : LoadStoreOpType::LdMatrix);
+          toTranspose ? LoadStoreOpType::LdMatrixTranspose
+                      : LoadStoreOpType::LdMatrix);
     }
   }
 
--- a/csrc/scheduler/matmul.cpp
+++ b/csrc/scheduler/matmul.cpp
@@ -785,13 +785,6 @@ void scheduleMatmul(Fusion* fusion, const MatmulParams& params) {
 
   // Collect mma swizzle info
   auto mma = mma_ops.front();
-  const auto mma_layout_opt = mma->layout();
-  NVF_ERROR(
-      mma_layout_opt.has_value(), "fusion mma op has undefined input layout");
-  const auto fusion_layout = mma_utils::getMmaLayout(fusion);
-  NVF_ERROR(fusion_layout.isValid(), fusion_layout.getErrorMsg());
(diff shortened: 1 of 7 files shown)


csrc/scheduler/mma_utils.cpp
```diff
@@ -703,11 +704,51 @@ std::unordered_set<IterDomain*> getMmaDomainSet(
...
 
 void WarpMmaSwizzler::scheduleLdMatrix(TensorView* tv, MmaOperand operand) {
-  bool transpose = tv->definition()->as<LoadStoreOp>()->opType() ==
-      LoadStoreOpType::LdMatrixTranspose;
+  bool transpose = !isConsumerAllocationInnerIDProducerAllocationInnerID(tv);
+  // We don't have to set the Op type here if we removed LdMatrix and LdMatrix
+  // and handle the decision to transpose or not in codegen.
+  tv->definition()->as<LoadStoreOp>()->setOpType(
+      transpose ? LoadStoreOpType::LdMatrixTranspose
+                : LoadStoreOpType::LdMatrix);
```
zasdfgbnm, line 751: Should we just remove `LdMatrixTranspose` and only keep `LdMatrix`, and in lowering, we use `isConsumerAllocationInnerIDProducerAllocationInnerID` to decide whether `LdMatrix` should be lowered as a `ldmatrix` or `ldmatrix.trans`? I think this task itself can be peeled out from this PR and merged independently as a separate PR.

csrc/scheduler/mma_utils.cpp
```diff
@@ -703,11 +704,51 @@ std::unordered_set<IterDomain*> getMmaDomainSet(
   return {mma_domains.begin(), mma_domains.end()};
 }
 
+bool isConsumerAllocationInnerIDProducerAllocationInnerID(
+    TensorView* consumer) {
+  // We are only looking at the allocation domains accross
+  // the producer and consumer of a LoadStoreOp
+  auto LdStOp = consumer->definition();
```
jacobhinkle, line 711: You can use `consumer->definition()-as<LoadStoreOp>()`. Also nit: we usually use lower snake case for variable names.

csrc/scheduler/mma_utils.cpp
```diff
@@ -703,11 +704,51 @@ std::unordered_set<IterDomain*> getMmaDomainSet(
...
+  NVF_CHECK(dynamic_cast<LoadStoreOp*>(LdStOp));
```
jacobhinkle, line 712: Does the definition need to be a LoadStoreOp for this utility?

csrc/scheduler/mma_utils.cpp
protonu: I'll try and send a stacked PR on top of this.

csrc/scheduler/mma_utils.cpp
protonu: That's a really good point - I don't think this I have to limit it to a LoadStoreOp. I'll make that change.

csrc/scheduler/mma_utils.cpp
protonu: I'll remove the requirement for a LoadStoreOp here.

csrc/scheduler/mma_utils.cpp
zasdfgbnm: I think it should be on the bottom of this PR, where without this PR, the producer "AllocationInnerID" will be just the producer rfactor inner id.

zasdfgbnm: Basically I am looking for splitting this PR as smaller PRs, and I believe this part is a good choice to get out.

csrc/scheduler/mma_utils.cpp
```diff
@@ -703,11 +704,51 @@ std::unordered_set<IterDomain*> getMmaDomainSet(
...
+
+  // Get all the IDs from the innermost ID of the allocation domain of
+  // the consumer to the root domain of the consumer.
+  auto vals = DependencyCheck::getAllValsBetween(
+      {consumer->getRootDomain().begin(), consumer->getRootDomain().end()},
+      {consumer->getMaybeAllocationDomain().back()});
+
+  auto ids = ir_utils::filterByType<IterDomain>(vals);
+  auto idsOnPath = std::vector<IterDomain*>(ids.begin(), ids.end());
```
jacobhinkle, line 721: I don't think this line is needed is it? Just use `ids` instead of `idsOnPath`. Also a nit: const on all these variables.

csrc/scheduler/mma_utils.cpp
```diff
@@ -703,11 +704,51 @@ std::unordered_set<IterDomain*> getMmaDomainSet(
...
+  // We are only looking at the allocation domains accross
+  // the producer and consumer of a LoadStoreOp
+  auto LdStOp = consumer->definition();
+  NVF_CHECK(dynamic_cast<LoadStoreOp*>(LdStOp));
+
+  // Get all the IDs from the innermost ID of the allocation domain of
+  // the consumer to the root domain of the consumer.
+  auto vals = DependencyCheck::getAllValsBetween(
+      {consumer->getRootDomain().begin(), consumer->getRootDomain().end()},
+      {consumer->getMaybeAllocationDomain().back()});
```
jacobhinkle, line 718: Why is this needed. Can you not just use `idInConsumerRoot = consumer->getMaybeAllocationDomain().back()`?

csrc/scheduler/mma_utils.cpp
```diff
@@ -703,11 +704,51 @@ std::unordered_set<IterDomain*> getMmaDomainSet(
...
+    TensorView* consumer) {
+  // We are only looking at the allocation domains accross
+  // the producer and consumer of a LoadStoreOp
+  auto LdStOp = consumer->definition();
+  NVF_CHECK(dynamic_cast<LoadStoreOp*>(LdStOp));
+
+  // Get all the IDs from the innermost ID of the allocation domain of
+  // the consumer to the root domain of the consumer.
+  auto vals = DependencyCheck::getAllValsBetween(
+      {consumer->getRootDomain().begin(), consumer->getRootDomain().end()},
```
jacobhinkle, line 717: Do you need to filter out broadcast and reduction domains?

csrc/scheduler/matmul.cpp
```diff
@@ -880,25 +881,17 @@ void scheduleMatmul(Fusion* fusion, const MatmulParams& params) {
...
-  } else {
-    acr = acw_smem->cacheAfter(LoadStoreOpType::LdMatrix);
-  }
-  if (auto ldst = dynamic_cast<LoadStoreOp*>(bcw_smem->uses().at(0))) {
-    bcr = ldst->out()->as<TensorView>();
-    if (ldst->hasInnerTranspose()) {
-      ldst->setOpType(LoadStoreOpType::LdMatrixTranspose);
+
+  for (auto [tv_smem, tv_r] :
+       {std::make_pair(acw_smem, &acr), std::make_pair(bcw_smem, &bcr)}) {
```
jacobhinkle, line 886: Why the `&`s here?

jacobhinkle: Oh I see `tv_r` is a `TensorView**`. This might be clearer if you made a lambda and added a couple comments.

csrc/scheduler/matmul.cpp
```diff
@@ -880,25 +881,17 @@ void scheduleMatmul(Fusion* fusion, const MatmulParams& params) {
...
+    if (auto ldst = dynamic_cast<LoadStoreOp*>(tv_smem->uses().at(0))) {
+      *tv_r = ldst->out()->as<TensorView>();
     } else {
-      ldst->setOpType(LoadStoreOpType::LdMatrix);
+      *tv_r = tv_smem->cacheAfter(
+          LoadStoreOpType::Set,
+          CacheOp::Unspecified,
+          false /* propagate allocation*/);
```
jacobhinkle, line 893: Passing on a tip I got from @wujingyue: if you use `/*propagate_allocation_domain=*/false` instead then `clang-tidy` will check that the variable name matches the comment.

jacobhinkle: Please add comments about the allocation domains of `tv_smem` and `tv_r`, and why you need to not propagate allocation domain at this location.

tests/cpp/test_matmul_scheduler.cpp
```diff
@@ -2699,6 +2699,369 @@ TEST_F(MatmulSchedulerTest, DISABLED_RequireExternalPlugin) {
   MatmulParams params;
 }
 
+class AllocationDomainTest
+    : public NVFuserFixtureParamTest<std::tuple<bool, bool>> {
+ protected:
+  // Allocation order set by the pass breaks matmul tests
+  // see issue https://github.com/NVIDIA/Fuser/issues/1810
+  AllocationDomainTest() : optimization_guard_(false) {
+    DisableOptionsGuard::getCurOptions().set(DisableOption::MatmulExprEval);
```
jacobhinkle, line 2708: Is this still necessary?

tests/cpp/test_matmul_scheduler.cpp
```diff
@@ -2699,6 +2699,369 @@ TEST_F(MatmulSchedulerTest, DISABLED_RequireExternalPlugin) {
...
+
+    // [M, N, K]
+    auto tv0b = broadcast(tv0, {false, true, false});
+    auto tv1b = broadcast(tv1t, {true, false, false});
+
+    auto tv2 = fusedMultiplySum(tv0b, tv1b, {2});
+    fusion->addOutput(tv2);
+  };
+
+  auto schedule = [](Fusion* fusion) {
```
jacobhinkle, line 2741: Is there a particular reason to use lambdas for the definition and scheduling here?

tests/cpp/test_matmul_scheduler.cpp
```diff
@@ -2699,6 +2699,369 @@ TEST_F(MatmulSchedulerTest, DISABLED_RequireExternalPlugin) {
...
+  DisableOptionsGuard option_guard_;
+};
+
+// This tests fusions where inputs to a Matmul will have the root domains
+// [M, K] and [K, N], and all possible combinations of allocation domains.
+// Please note that inpout in B is transposed prior to creating a Mma op.
+TEST_P(AllocationDomainTest, BasicMatmul) {
+  NVFUSER_TEST_CUDA_ARCH_RANGE_GUARD(7, 5, 9, 0);
+  bool a_m_inner = std::get<0>(GetParam());
+  bool b_k_inner = std::get<0>(GetParam());
```
jacobhinkle, line 2724: Should this be the following?
```suggestion
  bool b_k_inner = std::get<1>(GetParam());
```

tests/cpp/test_matmul_scheduler.cpp
```diff
@@ -2699,6 +2699,369 @@ TEST_F(MatmulSchedulerTest, DISABLED_RequireExternalPlugin) {
...
+}
+
+// Matmul test for Ampere MMA: across supported layouts
+TEST_P(AllocationDomainTest, FusionAmpereMatmul_CUDA) {
+  // Keep multiples of 8 to keep vectorizable.
+  int M = 512, N = 128, K = 256;
+
+  NVFUSER_TEST_CUDA_ARCH_RANGE_GUARD(7, 5, 9, 0);
+  bool a_m_inner = std::get<0>(GetParam());
+  bool b_k_inner = std::get<0>(GetParam());
```
jacobhinkle, line 2796:
```suggestion
  bool b_k_inner = std::get<1>(GetParam());
```

tests/cpp/test_matmul_scheduler.cpp
```diff
@@ -2699,6 +2699,369 @@ TEST_F(MatmulSchedulerTest, DISABLED_RequireExternalPlugin) {
...
+  NVF_CHECK(cg_outputs[0].allclose(tref, 0.0001, 0.0001));
+}
+
+TEST_P(AllocationDomainTest, FusionAmpereMatmulWithPrologue) {
+  // Keep multiples of 8 to keep vectorizable.
+  int M = 512, N = 128, K = 256;
+
+  NVFUSER_TEST_CUDA_ARCH_RANGE_GUARD(7, 5, 9, 0);
+  bool a_m_inner = std::get<0>(GetParam());
+  bool b_k_inner = std::get<0>(GetParam());
```
jacobhinkle, line 2864:
```suggestion
  bool b_k_inner = std::get<1>(GetParam());
```

tests/cpp/test_matmul_scheduler.cpp
```diff
@@ -2699,6 +2699,369 @@ TEST_F(MatmulSchedulerTest, DISABLED_RequireExternalPlugin) {
...
+}
+
+// Matmul test for Ampere MMA: across supported layouts
+TEST_P(AllocationDomainTest, FusionAmpereMatmulNoTranspose) {
+  // Keep multiples of 8 to keep vectorizable.
+  int M = 512, N = 128, K = 256;
+
+  NVFUSER_TEST_CUDA_ARCH_RANGE_GUARD(7, 5, 9, 0);
+  bool a_m_inner = std::get<0>(GetParam());
+  bool b_k_inner = std::get<0>(GetParam());
```
jacobhinkle, line 2934:
```suggestion
  bool b_k_inner = std::get<1>(GetParam());
```

tests/cpp/test_matmul_scheduler.cpp
```diff
@@ -2699,6 +2699,369 @@ TEST_F(MatmulSchedulerTest, DISABLED_RequireExternalPlugin) {
...
+}
+
+// Matmul test for Ampere MMA: across supported layouts
+TEST_P(AllocationDomainTest, FusionAmpereMatmulPrologueNoTranspose) {
+  // Keep multiples of 8 to keep vectorizable.
+  int M = 512, N = 128, K = 256;
+
+  NVFUSER_TEST_CUDA_ARCH_RANGE_GUARD(7, 5, 9, 0);
+  bool a_m_inner = std::get<0>(GetParam());
+  bool b_k_inner = std::get<0>(GetParam());
```
jacobhinkle, line 2999:
```suggestion
  bool b_k_inner = std::get<1>(GetParam());
```

jacobhinkle (commented):
It seems there are a few changes here:
- Don't propagate allocation domain when using `cacheAfter` on smem buffers, since we need the loaded register buffers to have allocation domains matching their root domains.
- Change `scheduleLdMatrix` to check consumer/producer innermost allocation ID to see whether transpose is needed. Previously this was signalled by the `LoadStoreOpType` on that op.
- Use allocation domain instead of root domain for `orderTiledConcreteIdAsRoot`. This is called only on shared memory TVs, and we now have possibly non-trivial allocation domains on those tensors.

I think this generally seems fine. I do have a question that we can address in the future: how should we handle cases where there is a transposed operand which has a prologue that comes _before_ the transpose? It seems that we still rely on using the smem->register load for transposing but in such a case that will come _after_ the prologue.

csrc/scheduler/mma_utils.cpp
protonu: In the case of the matmul scheduler and the cache read TensorViews such as `bcr`/`acr`, the allocation domains are complicated (based on the leaf domains). The innermost ID from the allocation domain here is something that was generated by scheduling and ID (`k`). I traverse back to that ID here and then map it back to the producer.
 Let me know if you think this isn't correct.

csrc/scheduler/mma_utils.cpp
protonu: I sort of based on this code here:
https://github.com/NVIDIA/Fuser/blob/5e0c89b289edcf52c4162713ee498ea86ce80564/csrc/ir/utils.cpp#L738-L748
`  // Filter so we only have iteration domains (ignore Ints used in split)
  auto all_ids = ir_utils::filterByType<IterDomain>(all_vals);
  return std::vector<IterDomain*>(all_ids.begin(), all_ids.end());`

csrc/scheduler/mma_utils.cpp
protonu: I sort of based it on this:

https://github.com/NVIDIA/Fuser/blob/5e0c89b289edcf52c4162713ee498ea86ce80564/csrc/ir/utils.cpp#L738-L748


  // Filter so we only have iteration domains (ignore Ints used in split)
  auto all_ids = ir_utils::filterByType<IterDomain>(all_vals);
  return std::vector<IterDomain*>(all_ids.begin(), all_ids.end());

csrc/scheduler/mma_utils.cpp
jacobhinkle: I see. I want to understand this a bit better.

This is called from here https://github.com/NVIDIA/Fuser/blob/5e0c89b289edcf52c4162713ee498ea86ce80564/csrc/scheduler/matmul.cpp#L1083-L1086
so the allocation domain of `acr` is a split/reorder of its logical domain in this case. So we might have something like the following:
```c++
acr = acw_smem->cacheAfter();
// acw_smem [ alloc: iS0{K} iS1{M} ]
//   iS2{Ko}, iS3{Ki} = split(iS0{K}, Ki)
//   iS4{Mo}, iS5{Mi} = split(iS1{M}, Mi}
// acr [ alloc: iS4{Mo} iS2{Ko} iS5{Mi} iS3{Ki} ]
```
If we supported multiple M dimensions then we would likely merge before this stage, so we would have
```c++
acr = acw_smem->cacheAfter();
// acw_smem [ alloc: iS0{M1} iS1{K} iS2{M2} ]
//   iS3{Ko}, iS4{Ki} = split(iS1{K}, Ki)
//   iS5{M1 * M2} = merge(iS0{M1}, iS2{M2})
//   iS6{Mo}, iS7{Mi} = split(iS5{M1 * M2}, Mi}
// acr [ alloc: iS6{Mo} iS3{Ko} iS7{Mi} iS4{Ki} ]
```
In this case it matters that the merge is in this order. If we did `iS5 = merge(iS2, iS0)` instead then I think we would not consider the allocation domain to be transposed? So should we actually traverse the IterDomain expressions from producer allocation to consumer allocation to determine whether the consumer's allocation domains are the innermost splits of (a possible inner merge of) the innermost producer allocation dimension?

Would it be simpler to just record the loop domain at the call site then actually set it after calling `ldMatrix`?

protonu: @jacobhinkle, I'd like to better understand your comments. 
I created a PR (https://github.com/NVIDIA/Fuser/pull/2315/) to remove LdMatrixTranspose. 
Can we move this conversation to that PR?

zasdfgbnm:
This PR needs rebase so that changes in https://github.com/NVIDIA/Fuser/pull/2315 is excluded from the diff of this PR.

protonu:
!build

protonu:
!build

protonu:
!build

zasdfgbnm (approved):

PR #2313, opened 2024-05-29
Loop promotion fix for #2261

### LINKED ISSUE #2261: Invalid loop promotion due to propagation of incomplete promotion
```
t0: [i0, i1, i2]
t1: [i3, i0, i1, i2, i4]

t2 = t0
t3 = broadcast(t2, {true, false, false, false, true})
t4 = t3 + t1

t4->merge(-2, -1)->merge(-2, -1)->merge(-2, -1)->split(0, 4)
propagateFrom(t4)

for t in {t2, t3, t4}:
  t->inlineAt(1)
```

transform
(The colors indicate loop grouping)

In this case, domain `38` should be promoted to a domain that is exact mapped with domain `26`. However, it results in being promoted to domain `65`, which is generated from domain `61` as shown below:

```
Split: iS61{( 1 * ( i0 * ( i2 * ( i3 * i8 ) ) ) )} by factor 4 -> iS64{( ceilDiv(( 1 * ( i0 * ( i2 * ( i3 * i8 ) ) ) ), 4) )}, iS65{4}, start offset: 0, stop offset: 0
```

Notice that the extent of domain `61` has `1 * (i0...`, meaning it's not exact mapped with domain `24`.

This incorrect promotion is due to the second propagation over an IEL graph (i.e., Step 4), which starts with the first results of the loop graph promotion produced by Step 3. At Step 3, the loop graph that contains domain `36` is promoted to `24`, which is correct, but there's also a separate loop group of `{10 14 15 19 20 21 33}`, which is promoted to `21`. 

```
idg{11 16 23 24 35 36 40} -> 24
idg{10 14 15 19 20 21 33} -> 21
```

The promotion of the loop group of `{10 14 15 19 20 21 33}` gets propagated in Step 4 to promote `36` to `61`, and then `38` to `65`. While `36` has the initial loop promotion to `24`, [it isn't used as the propagation through the IEL graph has precedence](https://github.com/NVIDIA/Fuser/blob/main/csrc/id_model/id_model.cpp#L1118-L1137). 

Here's the Step-4 results:

```
idg{38} -> 65
idg{37} -> 64
idg{36} -> 61
idg{42} -> 59
idg{41} -> 58
idg{35} -> 23
idg{40} -> 23
idg{34} -> 22
idg{39} -> 22
```

Essentially, propagation of the loop promotion of `{10 14 15 19 20 21 33}` shouldn't be done. This loop group itself doesn't correspond to any meaningful actual loop but seems like an incomplete group. 

Changing the [precedence](https://github.com/NVIDIA/Fuser/blob/main/csrc/id_model/id_model.cpp#L1118-L1137) doesn't solve the issue as there's indeed a case where the IEL propagation needs to be used instead of the Step-3 loop promotion results.

Fundamentally, when propagating the Step-3 results, we should only do so with the loop groups that already have valid loop promotions. Groups like `{10 14 15 19 20 21 33}` should not join the Step-4 propagation anymore. This could be done by filtering what to propagate at Step 4. Remember that Steps 4 and 5 are for finding promotions of partially inlined groups, so by starting the Step-4 propagation with the expr groups that introduce partial inlining,  e.g., the expr group of the splits of domains `{44 40 36 30 24}`, the issue should be avoided. As a side effect, it should also work as an optimization since the propagations in Steps 4 and 5 are minimized.

naoyam:
Fixes #2261 

Instead of blindly propagating the Step-3 results, just propagate those that matter for partially inlined groups. This fixes the issue of #2261 as well as works as a shortcut if there's no partially inlined groups.

### DIFF
--- a/csrc/id_model/loop_promotion.cpp
+++ b/csrc/id_model/loop_promotion.cpp
@@ -27,6 +27,60 @@ const ValGraph& LoopPromotionMapBuilder::idGraph(IdMappingMode mode) const {
   return id_model_.idGraph(mode);
 }
 
+std::unordered_map<ValGroup, IterDomain*> LoopPromotionMapBuilder::
+    getLoopPromotionsToPropagateAgain(
+        const std::unordered_map<ValGroup, IterDomain*>&
+            initial_loop_promotion_map,
+        const ValGraph& loop_graph) const {
+  std::unordered_map<ValGroup, IterDomain*> loop_promotion_map_to_propagate;
+
+  for (const auto& map_kv : initial_loop_promotion_map) {
+    const auto& loop_group = map_kv.first;
+    const auto& promotion = map_kv.second;
+
+    // If it's promoted to the exactly mapped domain, should not
+    // need further propagation
+    if (std::all_of(
+            loop_group->begin(),
+            loop_group->end(),
+            [&](Val* loop_group_val) -> bool {
+              return idGraph(IdMappingMode::EXACT)
+                  .disjointValSets()
+                  .strictAreMapped(loop_group_val, promotion);
+            })) {
+      continue;
+    }
+
+    const ExprGroups& uses = loop_graph.getUses(loop_group);
+    if (uses.empty()) {
+      continue;
+    }
+    const int expected_num_consumer_loop_group_count_if_fully_inlined =
+        (int)uses.front()->front()->outputs().size();
+
+    // Should not cause partial inline. This should also filter out
+    // circular output edges due to broadcast merge.
+    if (expected_num_consumer_loop_group_count_if_fully_inlined == 1) {
+      continue;
+    }
+
+    ValGroups consumer_loop_groups;
+    for (const ExprGroup& use : loop_graph.getUses(loop_group)) {
+      std::vector<ValGroup> output_loop_groups = loop_graph.outputGroups(use);
+      consumer_loop_groups.pushBack(output_loop_groups);
+    }
+    if (consumer_loop_groups.size() ==
+        expected_num_consumer_loop_group_count_if_fully_inlined) {
+      // Not propagating further as it's not partial inline
+      continue;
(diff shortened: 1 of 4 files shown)


naoyam:
!build

naoyam:
@zasdfgbnm `IdModelTest.LoopPromotion*` need to be updated but this is generally ready for review. I only updated `LoopPromotion4` and `LoopPromotion5` so far (as well as the corresponding part in the doc); will update the rest as well.

csrc/id_model/loop_promotion.cpp
```diff
@@ -27,6 +27,65 @@ const ValGraph& LoopPromotionMapBuilder::idGraph(IdMappingMode mode) const {
...
+    ValGroups consumer_loop_groups;
+    for (const ExprGroup& use : loop_graph.getUses(loop_group)) {
+      std::vector<ValGroup> output_loop_groups = loop_graph.outputGroups(use);
+      consumer_loop_groups.pushBack(output_loop_groups);
+    }
+
+    if (consumer_loop_groups.size() ==
+        expected_num_consumer_loop_group_count_if_fully_inlined) {
+      continue;
+    }
```
zasdfgbnm, line 81: Am I understanding correct that we can `NVF_ERROR(consumer_loop_groups.size() >= expected_num_consumer_loop_group_count_if_fully_inlined)`?

zasdfgbnm (approved):

naoyam:
!build

csrc/id_model/loop_promotion.cpp
naoyam: I think so. If the two outputs of a split are involved in broadcast forwarding, they could be grouped together, so if that happens, there should be only one output loop group, but I don't think that can happen. Let me insert the assertion.

naoyam:
!build

PR #2314, opened 2024-05-29
Add `DynamicType::dispatch` as a general mechanism for dispatching function calls

zasdfgbnm:
Please see the update in the `README.md` for description. 

This PR also updates `DynamicType`'s conversion operator to use this mechanism for testing purposes. In a followup PR, more operators will be changed to use this mechanism.

### DIFF
--- a/lib/dynamic_type/test/dispatch.cpp
+++ b/lib/dynamic_type/test/dispatch.cpp
@@ -1,0 +1,180 @@
+// clang-format off
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2023-present NVIDIA CORPORATION & AFFILIATES.
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+// clang-format on
+
+#include <gmock/gmock-matchers.h>
+#include <gtest/gtest.h>
+
+#include "dynamic_type/dynamic_type.h"
+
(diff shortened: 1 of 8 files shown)


zasdfgbnm:
!build
!build

zasdfgbnm:
!build

lib/dynamic_type/test/dispatch.cpp
```diff
@@ -0,0 +1,166 @@
...
+}
+
+TEST_F(DynamicTypeTest, DispatchArgumentPerfectForwarding) {
+  using IntOrStr = DynamicType<NoContainers, int64_t, std::string>;
+  struct NonCopyable {
+    NonCopyable() = default;
+    NonCopyable(const NonCopyable&) = delete;
+  };
+  auto concat =
+      [](std::string& result, NonCopyable&& nc, auto x, auto y, std::string b) {
```
naoyam, line 66: What is the reason of having the `NonCopyable` parameter? It seems it's not used.

zasdfgbnm: It is just testing that the perfect forwarding is working correctly, because otherwise, we will likely try to create a copy of `nc`, which will lead to an error.

naoyam (approved):
LGTM

Issue #2258, opened 2024-05-16
fp8 recipe support with full reduction


jjsjann123:
fp8 recipe requires `amax` called on output tensor for delayed scaling.

An example can be seen here: https://github.com/jjsjann123/triton_minitu/blob/60b30e46f015fcdf26b6dcf90f875428b799c0b3/nvfuser/bias_add_ln_nvfuser.py#L72-L74

The example is a biasAddLayerNorm layer. It has poor performance due to the full reduction leading to bad segmentation, since the `amax` call cannot be merged into partial reduction in layernorm

We can improve performance by manually breaking `amax` into two reduction operations: https://github.com/jjsjann123/triton_minitu/blob/bbfc7df6f341c68fef52fe3a1a34c3e9c824b6b3/nvfuser/bias_add_ln_nvfuser.py#L75-L90

Given that this is going to be a common pattern. We might want to work it into a ***preseg optimization pass*** to automatically break full reduction if there's horizontal partial reduction.

The original issue was discussed here: https://github.com/NVIDIA/Fuser/issues/1660#issuecomment-2037946622

rdspring1:
I am, but I decided against using a pre-segmentation pass for this.

I was going to segmentation in the python frontend to handle this.

e.g., Manually segment specific layer-norm fusions with fp8 amax reduction rather than detecting a pattern, change the definition, then apply segmentation.

rdspring1:
FP8 with single amax scale factor is lower priority than mxfp8 and nvfp4 for blackwell.

PR #2318, opened 2024-05-29
[WIP - CI Testing Only] Fix Issue 2258

rdspring1:
This PR fixes issue #2258


rdspring1:
!build

jjsjann123:
Is this PR ready for review? Not rushing it, asking since Kevin mentioned about it in the sync meeting :stuck_out_tongue_winking_eye:

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
The latency with `FactorAmaxReduction` enabled is about ~67% faster than the original fusion. 

With `FactorAmaxReduction`, the total latency is 0.02752 ms.
```
============== Profile Kernel ==================
name: nvfuser_inner_persistent_f0_c1_r0_g1, schedule: inner_persistent, segment_id: 1, device: 0, stream: 7
compile time: 284.34 ms, grid: [512, 1, 1], block: [384, 1, 1], registers: 56
input size: 25239556 bytes, output size: 37750784 bytes, time: 0.024160 ms
Effective Bandwidth: 2607.22 GB/s, Peak Bandwidth: 77.773484%

name: nvfuser_reduction_f0_c1_r0_g0, schedule: reduction, segment_id: 0, device: 0, stream: 7
compile time: 105.31 ms, grid: [16, 1, 1], block: [8, 1, 1], registers: 30
input size: 2052 bytes, output size: 4 bytes, time: 0.003360 ms
Effective Bandwidth: 0.61 GB/s, Peak Bandwidth: 0.018253%
==========================================
```

Without `FactorAmaxReduction`, the total latency is 0.04592 ms.
```
============== Profile Kernel ==============
name: nvfuser_inner_persistent_f0_c1_r0_g0, schedule: inner_persistent, segment_id: 0, device: 0, stream: 7
compile time: 207.91 ms, grid: [512, 1, 1], block: [384, 1, 1], registers: 56
input size: 25190400 bytes, output size: 50333696 bytes, time: 0.021824 ms
Effective Bandwidth: 3460.60 GB/s, Peak Bandwidth: 103.229949%

name: nvfuser_reduction_f0_c1_r0_g1, schedule: reduction, segment_id: 1, device: 0, stream: 7
compile time: 211.44 ms, grid: [660, 1, 1], block: [512, 1, 1], registers: 40
input size: 25217032 bytes, output size: 12582916 bytes, time: 0.024096 ms
Effective Bandwidth: 1568.72 GB/s, Peak Bandwidth: 46.795143%
==========================================
```

rdspring1:
`FusionFactorAmaxHorizontalMultiplePartial` represents a worst-case scenario where we factor the amax reduction, but it is not fused with the other segments. The four segment version created with the `FactorAmaxReduction` pass is 27% slower than the original.

Four segments are created with `FactorAmaxReduction` with a total latency of 22.92 us.
Three segments are created without `FactorAmaxReduction` with a total latency of 17.92 us.

tests/cpp/test_preseg_passes.cpp
```diff
@@ -642,4 +642,491 @@ TEST_F(PresegTest, ReplaceOutput) {
...
+
+  // Add Inner and Outer Partial Reductions together
+  TensorView* tv2 = sum(tv0, {1}, /*keepdim=*/true);
+  TensorView* tv3 = sum(tv0, {0}, /*keepdim=*/false);
+  TensorView* tv4 = add(tv2, tv3);
+  fusion.addOutput(tv4);
+
+  // Full Amax Reduction
+  TensorView* tv5 = abs(tv4);
+  TensorView* tv6 = max(tv5, {0, 1}, /*keepdim=*/false);
```
jjsjann123, line 884: brainstorm here.

From this patten, looks like we cannot really merge the partial reduction if there's a divergence in the reduction axes in its immediate dominating reductions.

Should we rethink this criteria:
> 3. Select upstream, partial reduction.
> Ideally, we would pick the partial reduction that lets us save the smallest partial Amax reduction. However, in this pre-segmentation pass, we do not know how the fusion could be segmented nor the size of the tensors. Given this lack of information, we select the partial reduction with shortest dependency chain with Amax reduction.

i.e. maybe we should back out from this optimization unless all of its immediate dominating reductions all share identical reduction axes?

csrc/kernel_cache.cpp
```diff
@@ -1242,6 +1242,19 @@ std::vector<at::Tensor> FusionKernelRuntime::runKernelWithInput(
   return outputs;
 }
 
+std::vector<std::unique_ptr<Fusion>> FusionKernelRuntime::getFusionSegments() {
```
jjsjann123, line 1245: Out of curiosity, is the natural order in fusion segments not stable enough for your validation?

i.e. `this->fusionSegments()->groups()`.

A nitpick, but if this is only used in test validation, I'm leaning towards not adding an API in FusionKernelRuntime.

tests/cpp/test_preseg_passes.cpp
```diff
@@ -642,4 +642,491 @@ TEST_F(PresegTest, ReplaceOutput) {
...
+    int64_t fusion_index,
+    int64_t number_of_outputs_in_fusion,
+    int64_t number_of_iterdomains,
+    int64_t expected_number_of_reduction_axes) {
+  // Check number of fusion segments
+  std::vector<std::unique_ptr<Fusion>> segments = runtime.getFusionSegments();
+  EXPECT_EQ(segments.size(), number_of_segments);
+
+  // Check number of outputs in selected segment
+  Fusion* selected_fusion = segments.at(fusion_index).get();
```
jjsjann123, line 658: follow the previous nitpick, we are also only using the segments `Fusion` at fusion_index. So we don't really need to construct everything.

tests/cpp/test_preseg_passes.cpp
```diff
@@ -642,4 +642,491 @@ TEST_F(PresegTest, ReplaceOutput) {
...
+  auto outputs = runtime.runWithInputs(args);
+
+  // Expected result of factorAmaxReduction pass:
+  // * Four segments are created because a partial reduction and a full
+  // reduction cannot be in the same fusion.
+  // * The first two reductions cannot be fused together because they do not
+  // share a common reduction axis.
+  // * Neither of the first two reduction segments can be fused with the third
+  // segment because the reduction scheduler does not support the pointwise
+  // epilogue.
```
jjsjann123, line 906: I think here by `pointwise epilogue.`, you mean broadcast to another TensorView in the epilogue?

Ideally we should fix the scheduler, but adding the logic into the optimization pass might be an easier target.

csrc/preseg_passes/factor_amax_reduction.cpp
```diff
@@ -0,0 +1,508 @@
...
+    return false;
+  }
+
+  return true;
+}
+
+// Detect amax pattern using a finite state machine.
+//
+// There are six states in FSM.
+// Start, Broadcast, Cast, Max-Reduction, Pass, Fail
```
jjsjann123, line 191:
```suggestion
// Start, Broadcast, Cast, Max-Reduction, Abs, Fail
```

csrc/preseg_passes/factor_amax_reduction.cpp
```diff
@@ -0,0 +1,508 @@
...
+TensorView* detectAmaxPattern(TensorView* tv) {
+  enum State { Start, Invalid, Abs, Broadcast, Cast, MaxReduction };
+
+  // Create state transition map
+  std::unordered_map<State, std::vector<State>> valid_states;
+  valid_states.emplace(
+      Start, std::vector<State>({Broadcast, Cast, MaxReduction}));
+  valid_states.emplace(Broadcast, std::vector<State>({MaxReduction}));
+  valid_states.emplace(Cast, std::vector<State>({Broadcast, MaxReduction}));
+  valid_states.emplace(MaxReduction, std::vector<State>({Abs}));
```
jjsjann123, line 228: nitpick: valid_states can be a `static const`

jjsjann123 (commented):
I need to go through the optimization logic. I think we should be on the cautious side and our goal here should be ensure that decomposed partial reduction can be merged with its producer reductions.

csrc/preseg_passes/factor_amax_reduction.h
```diff
@@ -0,0 +1,67 @@
...
+//  * The reduction is an output tensor.
+//  * It has a maximum reduction definition.
+//  * Its producer is absolute unary operation.
+// 3. Select upstream, partial reduction.
+//  * Ideally, we would pick the partial reduction that lets us save the
+//    smallest partial amax reduction. However, in this presegmentation pass,
+//    we do not know how the fusion could be segmented nor the size of the
+//    tensors.
+//  * Given this lack of information, we select the partial reduction with
+//    shortest dependency chain with amax reduction.
```
jjsjann123, line 46: I think this optimization should prioritize having the partial amax reduction to be fused with its dominating reduction.

csrc/preseg_passes/factor_amax_reduction.cpp
```diff
@@ -0,0 +1,508 @@
...
+  NVF_ERROR(
+      max_reduction_tv == nullptr ||
+      findReductionDefinition(max_reduction_tv, BinaryOpType::Max));
+  return max_reduction_tv;
+}
+
+// Select a partial reduction TensorView to factor the amax reduction.
+// Given the lack of information in the presegmentation pass, we select the
+// partial reduction with shortest dependency chain with amax reduction.
+TensorView* findUpstreamReduction(Fusion* fusion, TensorView* amax_reduction) {
```
jjsjann123, line 311: I think we need to pay more attention to the topology here. Once we found the closest reduction producer, I think we need to ensure that no TV will introduce more concrete IterDomain into the chain on the way to the full reduction.

```
TV1 = reducction(TV0)
TVn = follow_up_operations(TV0, TV1, ..., possibly other TVs as well)
amax(TVn);
```

Does it make sense to require that TVn's logical domain being a subset of the logical domain of TV0. So we know that TVn would be a valid reference tensor for the reduction?

This doesn't feel optimal, I'm wondering if it would help us get rid of the regresssion.

tests/cpp/test_preseg_passes.cpp
rdspring1:
```
**Segmenter** Considering fusion:
T2_l[ iS2{i0}, rS3{i2} ]
   = reduction( T0_g[ iS0{i0}, iS1{i2} ], op = add, initial value = float(0), allreduce = false )
T3_g[ iS4{i0}, bS5{1} ]
   = broadcast( T2_l[ iS2{i0}, rS3{i2} ] )
T6_g[ iS10{i0}, iS11{i2} ]
   = T3_g[ iS4{i0}, bS5{1} ]
   + T5_g[ bS8{1}, iS9{i2} ];

Scheduler _reduction_ ***rejected*** because : has unsupported post reduction fusion
Scheduler _inner_persistent_ ***rejected*** because : no persistent buffer identified
```
This ^^^ is the fusion pattern I was referring to. 

1. Both sum reductions cannot be merged because they have different reduction axes, so two reduction fusions are created.
2. The pointwise addition cannot be fused to either of the reduction fusions because of unsupported epilogue fusion.

csrc/preseg_passes/factor_amax_reduction.cpp
rdspring1: I think if there is a broadcast post-reduction, we need to check for a persistent pattern. Otherwise, the reduction scheduler will reject this pattern.

rdspring1:
!build

jjsjann123:
I'm seeing action on this PR again.

Let me know when it's good for another round of review. :bow:

rdspring1:
Not prioritizing fp8 without block-scaling.

PR #2321, opened 2024-05-30
Add `AbstractTensor` to conveniently schedule `IterDomain`s and `ValGroup`s.

zasdfgbnm:
Please start reviewing this PR from the comment in `csrc/domain_view.h`.

This PR only supports `merge`, other operations will be added in separate PRs.

Suggestions for naming are welcome.


zasdfgbnm:
!build

zasdfgbnm:
!build

csrc/domain_view.cpp
```diff
@@ -0,0 +1,138 @@
...
+      NVF_CHECK(
+          lhs.graph == rhs.graph,
+          "Can not merge ValGroups of different graph.");
+      auto graph = lhs.graph;
+      // If there is already an existing merge in the ValGraph, just use it.
+      auto lhs_uses = graph->getUses(lhs.group);
+      for (const ExprGroup& use : lhs_uses) {
+        auto input_groups = graph->inputGroups(use);
+        NVF_ERROR(input_groups.size() == 2);
+        if (input_groups == std::vector<ValGroup>{lhs.group, rhs.group}) {
```
zasdfgbnm, line 56: Note to myself: should I also check that the ExprGroup is a Merge?

csrc/domain_view.h
```diff
@@ -0,0 +1,142 @@
...
+//   IDLOView v({ValGroupAndItsGraph{g0, &g}, id1});
+//   v.merge(0);
+// This is also equivalent to Example 5. You will get [g01].
+
+namespace nvfuser {
+
+struct ValGroupAndItsGraph {
+  ValGroup group;
+  ValGraph* graph;
+};
```
jacobhinkle, line 108: This would be useful to have in `val_graph.h` as whenever we hold a `ValGroup` we probably also need its graph.

csrc/domain_view.h
```diff
@@ -0,0 +1,142 @@
...
+
+using IterDomainLikeObject = dynamic_type::DynamicType<
+    dynamic_type::Containers<std::vector>,
+    IterDomain*,
+    ValGroupAndItsGraph>;
+
+using IDLO = IterDomainLikeObject;
+
+struct IterDomainLikeObjectView {
```
jacobhinkle, line 117: Could we call this a `TensorDomainLikeObject`?

csrc/domain_view.h
```diff
@@ -0,0 +1,142 @@
...
+// This is also equivalent to Example 5. You will get [g01].
+
+namespace nvfuser {
+
+struct ValGroupAndItsGraph {
+  ValGroup group;
+  ValGraph* graph;
+};
+
+using IterDomainLikeObject = dynamic_type::DynamicType<
```
jacobhinkle, line 110: Would `AbstractIterDomain` be better? Or would that be confusing since it is not a base class of `IterDomain`?

csrc/domain_view.h
```diff
@@ -0,0 +1,142 @@
...
+
+  // TODO: split is not implemented yet
+  void split(int64_t axis, Val* factor, bool inner_split = true);
+  void split(int64_t axis, int64_t factor, bool inner_split = true);
+
+  void merge(int64_t axis_o, int64_t axis_i);
+  void merge(int64_t axis) {
+    merge(axis, axis + 1);
+  }
+};
```
jacobhinkle, line 138: For another PR: `toString()` and `printTransforms()` would be especially useful. Since `ValGroup`s aren't `Val`s I guess those methods could just introduce some names for them to use in the printout and then list the names of their IDs below.

jacobhinkle (commented):
This is exciting!

csrc/domain_view.cpp
```diff
@@ -0,0 +1,138 @@
...
+        std::is_same_v<R, ValGroupAndItsGraph>) {
+      NVF_CHECK(
+          lhs.graph == rhs.graph,
+          "Can not merge ValGroups of different graph.");
+      auto graph = lhs.graph;
+      // If there is already an existing merge in the ValGraph, just use it.
+      auto lhs_uses = graph->getUses(lhs.group);
+      for (const ExprGroup& use : lhs_uses) {
+        auto input_groups = graph->inputGroups(use);
+        NVF_ERROR(input_groups.size() == 2);
```
naoyam, line 55: Can you assume this?

zasdfgbnm: No, we can not. I have to check that it is a group of merges before checking this.

zasdfgbnm: fixed

csrc/domain_view.h
zasdfgbnm: moved

csrc/domain_view.h
```diff
@@ -0,0 +1,137 @@
...
+#include <ir/internal_base_nodes.h>
+#include <val_graph.h>
+
+#ifndef DYNAMIC_TYPE_CHECK
+#define DYNAMIC_TYPE_CHECK NVF_ERROR
+#endif
+
+#include <dynamic_type/dynamic_type.h>
+
+// IterDomainLikeObjectView, or IDLOView in short, is a convenient helper class
```
naoyam, line 20: What's the motivation of allowing just `IterDomain*` and a group of `IterDomain*`?

zasdfgbnm: I think it is mostly for "I want to merge this single ValGroup with multiple ValGroups separately". For example, if I want to manually create an exact graph, dim0 of all tensors are mapped together, dim1 are not because some are broadcasting some are concrete. To schedule, likely I will need to merge this single dim0 with all these different dim1s. For the IterDomain*, it is less interesting, but why not :-D.

naoyam:
I think I have two general comments, one on the interface and another on directions. For the interface, I wonder if the definition of the group merge should be separated from the dispatching interface. I understand that `DispatchMerge` is a convenient mechanism and it's nice to have. However, I think it would make things easier to maintain if the logic of merging groups is separated from the dispatching mechanism. Probably just having separate function like `merge(ValGroup, ValGroup)` and use it from `DispatchMerge`. What do you think?

On the directions, could you give some more concrete roadmap for this interface? We had an offline discussion along this line, but I don't recall if we reached any definitive conclusion regarding what we should do.

zasdfgbnm:
> Probably just having separate function like `merge(ValGroup, ValGroup)` and use it from `DispatchMerge`. What do you think?

I think this makes sense.

> On the directions, could you give some more concrete roadmap for this interface? We had an offline discussion along this line, but I don't recall if we reached any definitive conclusion regarding what we should do.

It's true that we do not have definitive conclusion, so I will not have a long-term roadmap until we have one definitive conclusion.

But in the near future, this will mostly serve as a convenient utility for writing unit tests. For example, I want to create a tensor domain whose leaf, root, allocation domains are separate transformations in a unit test. There is no linear structure, so our current scheduling primitives like `TensorView::merge` will not work. But with this, I can easily create one:

```C++
// Creating consumer:
IDLOView v_rfactor(clone(producer->gerRFactorDomain()));
IDLOView v_common = v_rfactor;
v_common.merge(0);
v_common.split(1, 234);
IDLOView v_leaf = v_common;
v_leaf.split(3, 45);
v_leaf.merge(6);
IDLOView v_alloc = v_common;
v_alloc.split(7, 89);
v_alloc.merge(10);
consumer_td = IrBuilder::create<TensorDomain>(
    {}, // root
    v_rfactor.as<IterDomain*>(), // rfactor
    v_leaf.as<IterDomain*>(), // rfactor
    v_alloc.as<IterDomain*>(), // rfactor
)
```

zasdfgbnm:
> Probably just having separate function like `merge(ValGroup, ValGroup)` and use it from `DispatchMerge`. What do you think?

Done

csrc/domain_view.h
zasdfgbnm: For `toString`, sure thing. For `printTransforms`, I don't think it is possible. Unlike TensorDomain, here we do not keep the history. We only keep the "current domain", so it is impossible to print transformations...

naoyam:
> > Probably just having separate function like `merge(ValGroup, ValGroup)` and use it from `DispatchMerge`. What do you think?
> 
> I think this makes sense.
> 
> > On the directions, could you give some more concrete roadmap for this interface? We had an offline discussion along this line, but I don't recall if we reached any definitive conclusion regarding what we should do.
> 
> It's true that we do not have definitive conclusion, so I will not have a long-term roadmap until we have one definitive conclusion.
> 
> But in the near future, this will mostly serve as a convenient utility for writing unit tests. For example, I want to create a tensor domain whose leaf, root, allocation domains are separate transformations in a unit test. There is no linear structure, so our current scheduling primitives like `TensorView::merge` will not work. But with this, I can easily create one:
> 
> ```c++
> // Creating consumer:
> IDLOView v_rfactor(clone(producer->gerRFactorDomain()));
> IDLOView v_common = v_rfactor;
> v_common.merge(0);
> v_common.split(1, 234);
> IDLOView v_leaf = v_common;
> v_leaf.split(3, 45);
> v_leaf.merge(6);
> IDLOView v_alloc = v_common;
> v_alloc.split(7, 89);
> v_alloc.merge(10);
> consumer_td = IrBuilder::create<TensorDomain>(
>     {}, // root
>     v_rfactor.as<IterDomain*>(), // rfactor
>     v_leaf.as<IterDomain*>(), // leaf
>     v_alloc.as<IterDomain*>(), // allocation
> );
> ```
> 
> I am expecting that `IDLOView` will be "the interface" if we want to schedule different domains of a tensor in a non-linear way in the future.

I see. For scheduling, do you think vectors of `IterDomain*` would be used instead of vectors of `ValGroup`?

zasdfgbnm:
> I see. For scheduling, do you think vectors of `IterDomain*` would be used instead of vectors of `ValGroup`?

Yes, at least for now. I added `ValGroups` just because it is not hard to add, and would be a convenient tool for prototyping ideas.

csrc/id_model/utils.h
naoyam, line 1: nit: "utils.h" sounds a little too generic to me. Maybe `schedule.h` or `transform.h`?

naoyam (approved):
LGTM.

zasdfgbnm:
> LGTM.

Any suggestions on names? I feel like I disliked both my naming and Jacob's recommendation. But I have no better idea...

naoyam:
Good naming seems difficult for this case. It can be just an iterdomain or a group of iterdomains, and I don't have any idea what would be a good name for it. `IDLO` doesn't seem to work very well for me as I still need to think about what it meant, but I understand `IterDomainLikeObject` is tedious to type. Just `IdGroup` may be fine even when it represents jus `IterDomain*`, but it may be confusing as `ValGroup` is always a group of `Val*`.

zasdfgbnm:
Renamed to `AbstractId` and `AbstractTensor` which seems to be a good balance of being meaningful while being easy to type. Let me know if you dislike this name.

PR #2322, opened 2024-05-31
Fix coverage with multiple expr groups producing the same val group

naoyam:
Previously, in `computeCoveredGroups`, when `covered_ids` has already a mapping, the additional covered groups are simply ignored. This could cause the loop promotion analysis to fail to find any promotion. See the added test for a concrete example. That repro is rather synthetic, but the same issue also happens with some of the reshape shmoo tests.


naoyam:
!build

csrc/id_model/loop_promotion.cpp
```diff
@@ -693,6 +693,13 @@ std::unordered_map<ValGroup, ValGroups> computeCoveredGroups(
       // Don't overwrite initialized cases due to rfactor markings.
       if (covered_ids.find(output_group) == covered_ids.end()) {
         covered_ids[output_group] = covered;
+      } else {
+        // An exact group may have multiple exact expr groups and may
+        // have different coverage groups depending on the expr
+        // groups. For example, this can happen with reshape or
+        // resize. See test LoopPromotionCoverage for a concrete
+        // example.
+        covered_ids[output_group].pushBack(covered);
```
zasdfgbnm, line 702: Can this be simplified to just a single:
```C++
for (const ValGroup& output_group : graph.outputGroups(exact_expr)) {
  covered_ids[output_group].pushBack(covered);
}
```

csrc/id_model/loop_promotion.cpp
```diff
@@ -693,6 +693,13 @@ std::unordered_map<ValGroup, ValGroups> computeCoveredGroups(
       // Don't overwrite initialized cases due to rfactor markings.
       if (covered_ids.find(output_group) == covered_ids.end()) {
         covered_ids[output_group] = covered;
+      } else {
```
zasdfgbnm, line 696: Is `view_rfactor_ids` still used?

naoyam: Good point. Removed.

zasdfgbnm (approved):

PR #2325, opened 2024-05-31
Convert from dynamic type to static type early

zasdfgbnm:
When talking about design decisions in `DynamicType`, there are some subtle details that is hard to tell which is the best decision. For example, `at::Tensor == at::Tensor` returns an `at::Tensor`, but some C++ code assumes `operator==` always returns `bool`. For this case, should we make `PolymorphicValue == PolymorphicValue` always return `bool` and disable `at::Tensor == at::Tensor`, or make it return `PolymorphicValue` which is a superset of `bool` and can be casted to `bool`? There is no clear answer to this question. None is perfect, it's just a matter of choosing which bad you want to take. So far, I believe the decision is based on whatever the behavior is that makes `DynamicType` codebase simple.

Recently, I added `DynamicType::dispatch`, which is primarily for supporting https://github.com/NVIDIA/Fuser/pull/2321, but I also noticed that this is a powerful tool that can be used to cleanup implementations of `DynamicType`'s operators.
To be able to more thoroughly cleaning things up, some decisions like above could change. Again, I believe this change is neither an improvement nor a regression on the behavior. It is just a different behavior.

One pitfall for using `PolymorphicValue` is, its `operator&&` and `operator||` are no longer short-circuiting. This is a fundamental limitation of C++ operator overloading, and to the best of my knowledge, there is no workaround for it. This means, if you write code like below:
```C++
pv.hasValue() && f(pv)
```
what you are thinking is likely
```C++
if (pv.hasValue()) {
  return (bool)f(pv);
} else {
  return false;
}
```
but what you get will be
```C++
PolymorphicValue cond1 = pv.hasValue();
PolymorphicValue cond2 = f(pv);
return (bool)(cond1 && cond2);
```
That is, `f(pv)` is always evaluated and likely this evaluation will leads to errors. To get the correct behavior, you will have to use
```C++
pv.hasValue() && (bool)(f(pv))
```
so that the `operator&&` is the `operator&&` of `bool`, not of `PolymorphicValue`.

In the past, I changed a batch of patterns like this during the enablement of `PolymorphicValue`, but there are still patterns like above remain unchanged because it was saved by the fact that the comparison operator of `PolymorphicValue` returning `bool`. If I changed the comparison operators to return `PolymorphicValue`, these uses will break too.

This PR further changes codes like above, so that we always use `operator&&` of bool regardless of the return type of the comparison operators.

### DIFF
--- a/csrc/compute_at_map.cpp
+++ b/csrc/compute_at_map.cpp
@@ -153,11 +153,13 @@ bool IterDomainGraph::exprsMap(
 
     auto extent_0_match = extent_0o->sameAs(extent_1o) ||
         (extent_0o->isConstInt() && extent_1o->isConstInt() &&
-         extent_0o->evaluate() == extent_1o->evaluate());
+         extent_0o->evaluate().as<int64_t>() ==
+             extent_1o->evaluate().as<int64_t>());
 
     auto extent_1_match = extent_0i->sameAs(extent_1i) ||
         (extent_0i->isConstInt() && extent_1i->isConstInt() &&
-         extent_0i->evaluate() == extent_1i->evaluate());
+         extent_0i->evaluate().as<int64_t>() ==
+             extent_1i->evaluate().as<int64_t>());
 
     if (!(extent_0_match || extent_1_match)) {
       return false;
--- a/csrc/device_lower/analysis/predicate_elimination.cpp
+++ b/csrc/device_lower/analysis/predicate_elimination.cpp
@@ -172,7 +172,8 @@ class ProducerConsumerPairAnalyzer : public OptOutDispatch {
 
     auto in_extent = split->in()->extent();
 
-    if (!in_extent->isConstInt() || ((in_extent->evaluate() % factor) != 0)) {
+    if (!in_extent->isConstInt() ||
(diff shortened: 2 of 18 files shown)


zasdfgbnm:
!build

zasdfgbnm:
!build

zasdfgbnm:
!build

zasdfgbnm:
!build

csrc/compute_at_map.cpp
```diff
@@ -153,11 +153,13 @@ bool IterDomainGraph::exprsMap(
 
     auto extent_0_match = extent_0o->sameAs(extent_1o) ||
         (extent_0o->isConstInt() && extent_1o->isConstInt() &&
-         extent_0o->evaluate() == extent_1o->evaluate());
+         extent_0o->evaluate().as<int64_t>() ==
```
naoyam, line 156: One question that's still unclear to me: Is this `as<T>()` strictly required?

zasdfgbnm: Not strictly `as<T>()`. Another alternative is `(extent_0o->evaluate() == extent_1o->evaluate())->as<bool>()`. But we have to make sure what the `&&` get is a `bool` instead of a `PolymorphicValue`.

Issue #2317, opened 2024-05-29
Merging IterDomains requires that their iteration types match.


wujingyue:
Check out `wjy/linear` and run `NVFUSER_DISABLE=parallel_compile python repro.py`. 

```
Traceback (most recent call last):
  File "/opt/pytorch/nvfuser/nvfuser/__init__.py", line 139, in execute
    result = self._execute(
RuntimeError: Merging IterDomains requires that their iteration types match. Outer: iS284{( ceilDiv(1600, 32) )}, Inner: rS257{i6}
Exception raised from merge at /opt/pytorch/nvfuser/csrc/ir/nodes.cpp:2565 (most recent call first):
frame #0: nvfuser::nvfCheckFail(char const*, char const*, unsigned int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) + 0xf3 (0x7f7d3928490d in /opt/pytorch/nvfuser/nvfuser/_C.cpython-310-x86_64-linux-gnu.so)
frame #1: nvfuser::IterDomain::merge(nvfuser::IterDomain*, nvfuser::IterDomain*, bool) + 0x38a (0x7f7d396e94ea in /opt/pytorch/nvfuser/nvfuser/_C.cpython-310-x86_64-linux-gnu.so)
frame #2: <unknown function> + 0x58364d (0x7f7d396e964d in /opt/pytorch/nvfuser/nvfuser/_C.cpython-310-x86_64-linux-gnu.so)
frame #3: nvfuser::TensorView::merge(long, long) + 0x106 (0x7f7d3989c796 in /opt/pytorch/nvfuser/nvfuser/_C.cpython-310-x86_64-linux-gnu.so)
frame #4: nvfuser::scheduleTranspose(nvfuser::Fusion*, nvfuser::TransposeParams) + 0xe9f (0x7f7d39866e2f in /opt/pytorch/nvfuser/nvfuser/_C.cpython-310-x86_64-linux-gnu.so)
frame #5: <unknown function> + 0x703551 (0x7f7d39869551 in /opt/pytorch/nvfuser/nvfuser/_C.cpython-310-x86_64-linux-gnu.so)
frame #6: <unknown function> + 0x5c4166 (0x7f7d3972a166 in /opt/pytorch/nvfuser/nvfuser/_C.cpython-310-x86_64-linux-gnu.so)
[...]

PR #2326, opened 2024-06-01
Transpose Scheduler: Ignore reduction axis in merge

Priya2698:
Issue #2317.

The issue arises in the following lines for reference 1: `[I0, I1, I2, r3]`:

After tiling:
```
  reference1->split(inner_most_pos1_in_ref1, params.tile_size1);
  reference1->reorder({{inner_most_pos1_in_ref1 + 1, -1}});
  reference1->split(inner_most_pos2_in_ref1, params.tile_size2);
  reference1->reorder({{inner_most_pos2_in_ref1 + 1, -1}});
```
Reference 1 is: [I0, I1/tile1, I2/tile2, r3, tile1, tile2]
```
 // Merge remaining dimensions
  int64_t lhs_i = -1;
  for (int64_t i = reference1->nDims() - 2; i > 0; i--) {
    auto axis_i = i - 1;
    if (lhs_i == -1) {
      lhs_i = axis_i;
    } else {
      reference1->merge(axis_i, lhs_i);
      lhs_i = axis_i;
    }
```
This tries to merge a reduction iterdomain with iteration type iterdomain.

This PR ignored the reduction axis when merging all non-tile dimensions.

### DIFF
--- a/csrc/scheduler/transpose.cpp
+++ b/csrc/scheduler/transpose.cpp
@@ -1222,10 +1222,10 @@ void scheduleTranspose(Fusion* fusion, TransposeParams params) {
   // Merge remaining dimensions ignoring reduction axes.
   int64_t rhs_i = reference1->nDims() - 3;
   for (int64_t lhs_i = reference1->nDims() - 4; lhs_i >= 0; lhs_i--) {
-    if (reference1->axis(lhs_i)->isReduction()){
+    if (reference1->axis(lhs_i)->isReduction()) {
       continue;
     }
-    if (reference1->axis(rhs_i)->isReduction()){
+    if (reference1->axis(rhs_i)->isReduction()) {
       rhs_i = lhs_i;
       continue;
     }
--- a/csrc/scheduler/transpose.cpp
+++ b/csrc/scheduler/transpose.cpp
@@ -1219,17 +1219,20 @@ void scheduleTranspose(Fusion* fusion, TransposeParams params) {
   reference1->reorder({{inner_most_pos2_in_ref1 + 1, -1}});
   // [..., I1/tile1, .., I2/tile2, ..., tile1, tile2]
 
-  // Merge remaining dimensions
-  int64_t lhs_i = -1;
-  for (int64_t i = reference1->nDims() - 2; i > 0; i--) {
-    auto axis_i = i - 1;
-    if (lhs_i == -1) {
-      lhs_i = axis_i;
-    } else {
-      reference1->merge(axis_i, lhs_i);
-      lhs_i = axis_i;
+  // Merge remaining dimensions ignoring reduction axes.
+  int64_t rhs_i = reference1->nDims() - 3;
+  for (int64_t lhs_i = reference1->nDims() - 4; lhs_i >= 0; lhs_i--) {
+    if (reference1->axis(lhs_i)->isReduction()){
+      continue;
     }
+    if (reference1->axis(rhs_i)->isReduction()){
+      rhs_i = lhs_i;
+      continue;
+    }
+    reference1->merge(lhs_i, rhs_i);
+    rhs_i = lhs_i;
   }
+
   reference1->split(0, 1);
   // [merged_dim, 1, tile1, tile2]
 
--- a/csrc/scheduler/transpose.cpp
+++ b/csrc/scheduler/transpose.cpp
@@ -1241,8 +1241,8 @@ void scheduleTranspose(Fusion* fusion, TransposeParams params) {
   // [r.., merged_dim, 1, tile1, tile2]
 
   // parallelize non-tile dimensions
-  reference1->axis(1)->parallelize(ParallelType::Unswitch);
-  reference1->axis(0)->parallelize(ParallelType::BIDx);
+  reference1->axis(rhs_i + 1)->parallelize(ParallelType::Unswitch);
+  reference1->axis(rhs_i)->parallelize(ParallelType::BIDx);
   // [BIDx, Unswitch, tile1, tile2]
 
   // Propagate transformations so far to the entire DAG
--- a/csrc/scheduler/transpose.cpp
+++ b/csrc/scheduler/transpose.cpp
@@ -1237,8 +1237,8 @@ void scheduleTranspose(Fusion* fusion, TransposeParams params) {
     rhs_i = lhs_i;
   }
 
-  reference1->split(0, 1);
-  // [merged_dim, 1, tile1, tile2]
+  reference1->split(rhs_i, 1);
+  // [r.., merged_dim, 1, tile1, tile2]
 
   // parallelize non-tile dimensions
   reference1->axis(1)->parallelize(ParallelType::Unswitch);
--- a/csrc/scheduler/transpose.cpp
+++ b/csrc/scheduler/transpose.cpp
@@ -1219,7 +1219,11 @@ void scheduleTranspose(Fusion* fusion, TransposeParams params) {
   reference1->reorder({{inner_most_pos2_in_ref1 + 1, -1}});
   // [..., I1/tile1, .., I2/tile2, ..., tile1, tile2]
 
-  // Merge remaining dimensions ignoring reduction axes.
+  // Merge remaining dimensions ignoring reduction axes (See Issue #2317)
+  // The reduction axes cannot be at any position.
+  // For example: [i0, r1, i1, r2, i2] after tiling is [i0, r1, i1/tile1, r2,
+  // i2/tile2, tile1, tile2] The following code merges all the outer iterdomains
+  // as: [i0 * i1/tile1 * i2/tile2, r1, r2, tile1, tile2]
   int64_t rhs_i = reference1->nDims() - 3;
   for (int64_t lhs_i = reference1->nDims() - 4; lhs_i >= 0; lhs_i--) {
     if (reference1->axis(lhs_i)->isReduction()) {
--- a/tests/python/test_python_frontend.py
+++ b/tests/python/test_python_frontend.py
@@ -4017,6 +4017,21 @@ class TestNvFuserFrontend(TestCase):
             fd.add_output(T101)
 
         nvf_out, _ = self.exec_nvfuser(fusion_func, inputs)
+    
+    # See https://github.com/NVIDIA/Fuser/issues/2317
+    def test_reduction_transpose_sched_issue2317(self):
+        inputs = [
+            torch.randn((1600,), dtype=torch.bfloat16, device='cuda:0').as_strided((1600,), (1,)),
+            torch.randn((2560000,), dtype=torch.bfloat16, device='cuda:0').as_strided((1600, 1600), (1600, 1)),
+            torch.randn((1600,), dtype=torch.bfloat16, device='cuda:0').as_strided((1600,), (1,)),
+            torch.randn((1600,), dtype=torch.bfloat16, device='cuda:0').as_strided((1600,), (1,)),
+            torch.randn((6400,), dtype=torch.bfloat16, device='cuda:0').as_strided((6400,), (1,)),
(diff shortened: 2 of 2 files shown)


Priya2698:
!build

Priya2698:
!build

Priya2698:
!build

Priya2698:
!build

tests/python/test_python_frontend.py
```diff
@@ -4016,6 +4016,187 @@ def fusion_func(fd: FusionDefinition) -> None:
 
         nvf_out, _ = self.exec_nvfuser(fusion_func, inputs)
 
+    # See https://github.com/NVIDIA/Fuser/issues/2317
+    @unittest.skipIf(is_pre_ampere(), "Only supported on Ampere and newer devices.")
+    def test_reduction_transpose_sched_issue2317(self):
```
wujingyue, line 4021: Is it possible to create a smaller, more targeted unit test? I understand that the test as is reproduces the issue, but given the size it'll be hard to debug when it fails.

Priya2698: Let me give it a shot.

csrc/scheduler/transpose.cpp
```diff
@@ -1219,17 +1219,20 @@ void scheduleTranspose(Fusion* fusion, TransposeParams params) {
...
+      continue;
+    }
+    if (reference1->axis(rhs_i)->isReduction()) {
+      rhs_i = lhs_i;
+      continue;
     }
+    reference1->merge(lhs_i, rhs_i);
+    rhs_i = lhs_i;
   }
+
```
jacobhinkle, line 1235: It took me working through a couple examples to see that this handles cases like `[ i r i r r tile1 tile2]` and properly merges the two outer iteration axes. If possible, some explanation in the comments would be helpful.

tests/python/test_python_frontend.py
jacobhinkle: I agree. Ideally we would just have a fusion with a sum along some axes, and a transpose. It would be nice to exercise multiple reduction axes as well given that the reduction skipping code is slightly tricky.

csrc/scheduler/transpose.cpp
Priya2698: Let me know if the code comment clarifies the loop.

tests/python/test_python_frontend.py
Priya2698: I added a smaller repro, using `linear`. I ran into another issue when using sum:

```python
import torch

    T0 = fd.define_tensor(shape=[-1, -1, -1, -1, -1], contiguity=[True, True, True, True, True], dtype=DataType.BFloat16, is_cpu=False, stride_order=[4, 3, 2, 1, 0])
    T1 = fd.define_tensor(shape=[-1, -1, -1, -1], contiguity=[True, True, True, True], dtype=DataType.BFloat16, is_cpu=False, stride_order=[3, 2, 1, 0])
    T2 = fd.ops.permute(T0, dims=[0, 1, 3, 2, 4])
    T3 = fd.ops.stride_order(T2, stride_order=[4, 3, 2, 1, 0])
    V4 = fd.ops.shape(T1)
    T5 = fd.ops.reshape(T3, new_shape=V4)
    T6 = fd.ops.sum(T5, dims=[1], keepdim=False, dtype=DataType.Null)
    T7 = fd.ops.sum(T1, dims=[1], keepdim=False, dtype=DataType.Null)
    T8 = fd.ops.add(T6, T7)
    fd.add_output(T8)

    nvfuser_fusion_id0(fd)

inputs = [
    torch.randn((104857600,), dtype=torch.bfloat16, device='cuda:0').as_strided((16, 32, 25, 128, 64), (6553600, 204800, 8192, 64, 1)),
    torch.randn((104857600,), dtype=torch.bfloat16, device='cuda:0').as_strided((16, 32, 128, 1600), (6553600, 204800, 1600, 1)),
]
fd.execute(inputs)

  File "/opt/pytorch/nvfuser/nvfuser/__init__.py", line 139, in execute
    result = self._execute(
RuntimeError: false INTERNAL ASSERT FAILED at "/opt/pytorch/nvfuser/csrc/executor_utils.cpp":829, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. 
```

csrc/scheduler/transpose.cpp
```diff
@@ -1219,17 +1219,24 @@ void scheduleTranspose(Fusion* fusion, TransposeParams params) {
...
-      lhs_i = axis_i;
-    } else {
-      reference1->merge(axis_i, lhs_i);
-      lhs_i = axis_i;
+  // Merge remaining dimensions ignoring reduction axes (See Issue #2317)
+  // The reduction axes cannot be at any position.
+  // For example: [i0, r1, i1, r2, i2] after tiling is [i0, r1, i1/tile1, r2,
+  // i2/tile2, tile1, tile2] The following code merges all the outer iterdomains
+  // as: [i0 * i1/tile1 * i2/tile2, r1, r2, tile1, tile2]
+  int64_t rhs_i = reference1->nDims() - 3;
```
jjsjann123, line 1227: neat. :clap:

csrc/scheduler/transpose.cpp
```diff
@@ -1219,17 +1219,24 @@ void scheduleTranspose(Fusion* fusion, TransposeParams params) {
...
-    if (lhs_i == -1) {
-      lhs_i = axis_i;
-    } else {
-      reference1->merge(axis_i, lhs_i);
-      lhs_i = axis_i;
+  // Merge remaining dimensions ignoring reduction axes (See Issue #2317)
+  // The reduction axes cannot be at any position.
+  // For example: [i0, r1, i1, r2, i2] after tiling is [i0, r1, i1/tile1, r2,
+  // i2/tile2, tile1, tile2] The following code merges all the outer iterdomains
+  // as: [i0 * i1/tile1 * i2/tile2, r1, r2, tile1, tile2]
```
jjsjann123, line 1226: one quick question, I think the scheduler assumed to have merged non-reduction domain on the outer most?

```
  reference1->split(0, 1);
  // [merged_dim, 1, tile1, tile2]
```

But if there is a reduction id on the outer most, we are screwed here right?

tests/python/test_python_frontend.py
Priya2698: Ahh the repro is not correct. Let me fix it.

csrc/scheduler/transpose.cpp
Priya2698: You are right. A cleaner fix would be to move the reduction dimensions before this using `moveReductionsOut` for both references. Then we have established a protocol on where the reduction dimensions will be.
That will also simplify this logic since non-reduction domains are inner to all reductions.

What do you think?

Another thing I am trying to verify is if the same thing happens for broadcasr dimension.

Priya2698:
```
reference1->axis(1)->parallelize(ParallelType::Unswitch);
reference1->axis(0)->parallelize(ParallelType::BIDx);
```

This will need to be modified to find the non-tile dimensions which will be after the reduction domains.

I think this approach might need more careful changes but would be less error-prone.

jjsjann123: > A cleaner fix would be to move the reduction dimensions before this using moveReductionsOut for both references. Then we have established a protocol on where the reduction dimensions will be.
That will also simplify this logic since non-reduction domains are inner to all reductions.
What do you think?

except you still need to change that later in `Step 4: Schedule group1`. since now inner most are not the tile ids.

I don't have a strong preference here, we can either do a re-order, or just change the parallelize call below to use the right id (non-reduction).

> Another thing I am trying to verify is if the same thing happens for broadcasr dimension.

good catch!

Priya2698: > > A cleaner fix would be to move the reduction dimensions before this using moveReductionsOut for both references. Then we have established a protocol on where the reduction dimensions will be.
> > That will also simplify this logic since non-reduction domains are inner to all reductions.
> > What do you think?
> 
> except you still need to change that later in `Step 4: Schedule group1`. since now inner most are not the tile ids.
> 
> I don't have a strong preference here, we can either do a re-order, or just change the parallelize call below to use the right id (non-reduction).


The outermost will be reduction domains. The inner most should still be tile. My suggestion was basically to do `moveReductionsOut` here instead of later Step 3 and Step 4.

jjsjann123: > The outermost will be reduction domains. The inner most should still be tile.

sorry I was mistaken. Yeah that sounds good to me.

tests/python/test_python_frontend.py
jacobhinkle: It looks like we're not promoting to float in that sum case. Would you mind posting that as a separate issue?

tests/python/test_python_frontend.py
```diff
@@ -4016,6 +4016,32 @@ def fusion_func(fd: FusionDefinition) -> None:
...
+            T16 = fd.ops.reshape(T11, new_shape=T1.shape())
+            T17 = fd.ops.linear(T16, T2)
+            T33 = fd.ops.add(T17, T1)
+
+            T33 = fd.ops.cast(T33, dtype=DataType.BFloat16)
+            T34 = fd.ops.linear(T33, T2)
+            T35 = fd.ops.add(T34, T33)
+            fd.add_output(T35)
+        nvf_out, _ = self.exec_nvfuser(partial(fusion_func, inputs=inputs), inputs)
+
```
Priya2698, line 4044: @jacobhinkle @jjsjann123 I have a smaller repro for the original issue.
I am working on test cases to exercise:
1. Multiple reduction axes in the input to transpose scheduler
2. Reduction axis at front in transpose scheduler

For these cases, I will experiment with C++ tests to mimic these conditions.

Priya2698:
!build

csrc/scheduler/transpose.cpp
```diff
@@ -1219,19 +1219,26 @@ void scheduleTranspose(Fusion* fusion, TransposeParams params) {
...
+      continue;
     }
+    reference1->merge(lhs_i, rhs_i);
+    rhs_i = lhs_i;
   }
-  reference1->split(0, 1);
-  // [merged_dim, 1, tile1, tile2]
+
+  reference1->split(rhs_i, 1);
+  // [r.., merged_dim, 1, tile1, tile2]
```
jjsjann123, line 1241: If I read the logic right, this is more likely to be
// [r.., merged_dim, 1, r..., tile1, tile2]

So I think the line below when we do unswitch, we should have been doing this instead vvv.
```
  // parallelize non-tile dimensions
  reference1->axis(rhs_i+1)->parallelize(ParallelType::Unswitch);
  reference1->axis(rhs_i)->parallelize(ParallelType::BIDx);
```

jjsjann123: I see later uses of reference1 happens after `moveReductionsOut`. So I think we should be good with it.

jjsjann123 (approved):
It's really unfortunate that our CI isn't working right now. But we don't have a test case for the tricky reduction id anyway.

The PR looks good to me, I'm going to trust you on updating parallelize unswitch and bidx per comment. Stamping to unblock

csrc/scheduler/transpose.cpp
Priya2698: > ```
>   // parallelize non-tile dimensions
>   reference1->axis(rhs_i+1)->parallelize(ParallelType::Unswitch);
>   reference1->axis(rhs_i)->parallelize(ParallelType::BIDx);
> ```

Oh right! Thanks for catching that.

csrc/scheduler/transpose.cpp
jacobhinkle: The comment helps in that it shows that we intend to cover this case, and we can verify that the code does that. It's just a slightly tricky function to write, but I think this is fine. Thanks.

PR #2328, opened 2024-06-03
Host irs: use `Communicator` directly instead of wrapped inside `PostOnStream`

samnordmann:
Bootstrapping for simplifying the design. With this patch, the `Communication` class is used to post a collective, without the need of wrapping it inside a `PostOnStream` Ir.

There is no functional change, this is only a minor cleanup I would like to upstream before introducing the "collective wait" IR.

To achieve this, we need to add I/O TVs to the `Communication`'s constructor, and use those I/O directly. This patch aligns with the discussion from https://github.com/NVIDIA/Fuser/pull/2250 and should lead to further simplification in `MultiDeviceExecutor`, that I would like to leave to another PR for avoiding conflicts.

Note that the `Communication` class is used in two places: Host IRs and `MultiDeviceExecutor`. Since the former case should replace the latter eventually, we maintain for now these two usages. In `MultiDeviceExecutor`, we don't set a `Communication`'s I/O, that's why those arguments are made optional in the constructor

### DIFF
--- a/csrc/host_ir/executor.cpp
+++ b/csrc/host_ir/executor.cpp
@@ -46,10 +46,8 @@ void HostIrExecutor::handle(PostOnStream* post_ir) {
   Expr* op = post_ir->hostOpToPost();
   if (op->isA<HostUnit>()) {
     postCompute(post_ir);
-  } else if (op->isA<Communication>()) {
-    postCommunication(post_ir);
   } else {
-    NVF_ERROR(false, "The op cannot be posted on a stream: ", op);
+    dispatch(op);
   }
 }
 
@@ -100,15 +98,10 @@ void HostIrExecutor::postCompute(PostOnStream* post_ir) {
   }
 }
 
-void HostIrExecutor::postCommunication(PostOnStream* post_ir) {
+void HostIrExecutor::handle(Communication* communication) {
   NVF_ERROR(
       communicator_ != nullptr && communicator_->is_available(),
       "A valid communicator must be provided");
-  NVF_ERROR(
-      post_ir->hostOpToPost()->isA<Communication>(),
-      "op must be a Communication: ",
-      post_ir->hostOpToPost());
-  auto communication = post_ir->hostOpToPost()->as<Communication>();
   NVF_ERROR(
       std::find(
           communication->params().team.begin(),
@@ -118,8 +111,8 @@ void HostIrExecutor::postCommunication(PostOnStream* post_ir) {
       communicator_->deviceId(),
       " must be present in the communication's team");
 
-  auto input_val = post_ir->inputs().at(0);
-  auto output_val = post_ir->outputs().at(0);
+  auto input_val = communication->inputs().at(0);
+  auto output_val = communication->outputs().at(0);
   at::Tensor input_tensor;
   if (val_to_IValue_.find(input_val) != val_to_IValue_.end()) {
     input_tensor = val_to_IValue_.at(input_val).toTensor();
--- a/csrc/host_ir/host_ir.cpp
+++ b/csrc/host_ir/host_ir.cpp
@@ -57,7 +57,7 @@ PostOnStream::PostOnStream(
       this,
       "must be registered in a HostIrContainer");
   NVF_ERROR(
-      (host_op->isOneOf<HostUnit, Communication>()), "wrong host op type");
+      (host_op->isA<HostUnit>()), "wrong host op type");
   if (host_op->isA<HostUnit>()) {
     NVF_ERROR(
         this->inputs().size() ==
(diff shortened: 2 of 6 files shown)


samnordmann:
!build --dist

samnordmann:
!build

tests/cpp/test_multidevice_host_ir.cpp
```diff
@@ -91,30 +82,35 @@ TEST_P(MultiDeviceHostIrTest, SingleFusionSingleComm) {
...
       compute_outputs);
-  std::vector<Val*> communication_inputs = {tv1};
-  std::vector<Val*> communication_outputs = {tv2};
-  auto post_communication = IrBuilder::create<PostOnStream>(
+  // [Step 5)b.] Create Communication Ir representing executing the Fusion
+  TensorView* communication_inputs = tv1->as<TensorView>();
+  TensorView* communication_outputs = tv2->as<TensorView>();
+  CommParams comm_params{
+      .type = CommunicationType::Allgather,
+      .root = 0,
```
wujingyue, line 98: Allgather shouldn't have a root. Correct?

tests/cpp/test_multidevice_host_ir.cpp
```diff
@@ -91,30 +82,35 @@ TEST_P(MultiDeviceHostIrTest, SingleFusionSingleComm) {
...
       static_cast<IrContainer*>(hic.get()),
       hu,
       compute_inputs,
       compute_outputs);
-  std::vector<Val*> communication_inputs = {tv1};
-  std::vector<Val*> communication_outputs = {tv2};
-  auto post_communication = IrBuilder::create<PostOnStream>(
+  // [Step 5)b.] Create Communication Ir representing executing the Fusion
+  TensorView* communication_inputs = tv1->as<TensorView>();
+  TensorView* communication_outputs = tv2->as<TensorView>();
```
wujingyue, line 95:
```suggestion
  auto* communication_input = tv1->as<TensorView>();
  auto* communication_output = tv2->as<TensorView>();
```

wujingyue: Note: I replaced inputs/outputs with input/output because there's atm only one input.

csrc/host_ir/executor.cpp
```diff
@@ -46,10 +46,8 @@ void HostIrExecutor::handle(PostOnStream* post_ir) {
   Expr* op = post_ir->hostOpToPost();
   if (op->isA<HostUnit>()) {
     postCompute(post_ir);
-  } else if (op->isA<Communication>()) {
-    postCommunication(post_ir);
   } else {
-    NVF_ERROR(false, "The op cannot be posted on a stream: ", op);
+    dispatch(op);
```
wujingyue, line 50: This is a change of behavior. Did you intend to do this in another PR?

csrc/host_ir/executor.cpp
```diff
@@ -118,8 +111,8 @@ void HostIrExecutor::postCommunication(PostOnStream* post_ir) {
       communicator_->deviceId(),
       " must be present in the communication's team");
 
-  auto input_val = post_ir->inputs().at(0);
-  auto output_val = post_ir->outputs().at(0);
+  auto input_val = communication->inputs().at(0);
+  auto output_val = communication->outputs().at(0);
```
wujingyue, line 115:
```suggestion
  Val* input_val = communication->input(0);
  Val* output_val = communication->output(0);
```

csrc/host_ir/host_ir.cpp
```diff
@@ -56,8 +56,7 @@ PostOnStream::PostOnStream(
       passkey.ir_container_->isA<hir::HostIrContainer>(), // NOLINT
       this,
       "must be registered in a HostIrContainer");
-  NVF_ERROR(
-      (host_op->isOneOf<HostUnit, Communication>()), "wrong host op type");
+  NVF_ERROR((host_op->isA<HostUnit>()), "wrong host op type");
```
wujingyue, line 59: Can you also print out the host_op so the error message is more useful?

wujingyue (commented):
Nice -- so I don't need to make the same change!

csrc/host_ir/executor.cpp
samnordmann: I don't see any change in behavior, am I missing smth?

wujingyue: You are right -- there's no behavior change. There's a subtle difference that the new code won't check the type of the host op in PostOnStream, which I think is fine.

samnordmann:
!build

csrc/host_ir/executor.cpp
samnordmann: IIUC the type of the host op is still checked, inside ´OptInDispatch::dispatch'

wujingyue:
!build

PR #2332, opened 2024-06-03
iter grouped multiple reductions

liqiangxl:
**Issue:** when have multiple outer reductions in a fusion, if vectorized, only the first reduction tv is grouped.
**Fix:** ensure all reductions are grouped
**Influence:** about 1.1x speedup for case `NvFuserScheduler_BiasDropoutAddLayernormBwd1_tf32/1024/128/864/manual_time`, the name is confusing, it is just a simple fusion with 2 reductions. Before this PR, the first reduction is using iterGroupedGridReduction while the 2nd reduction uses regular grid reduction. After this PR, both reductions use iterGroupedGridReduction.
```
Inputs:
  T0_g[ iS0{i0}, iS1{i2}, iS2{i3} ], float
  T1_g[ iS27{i0}, iS28{i2}, iS29{i3} ], float
  T2_g[ iS30{i0}, iS31{i2}, bS8{1} ], float
  T3_g[ iS35{i0}, iS36{i2}, bS11{1} ], float
Outputs:
  T6_g[ rS18{i0}, rS19{i2}, iS20{i3} ], float
  T8_g[ rS24{i0}, rS25{i2}, iS26{i3} ], float
  T5_g[ iS37{i0}, iS38{i2}, iS39{i3} ], float

%kernel_math {
T6_g[ rS18{i0}, rS19{i2}, iS20{i3} ]
   = reduction( T0_g[ iS0{i0}, iS1{i2}, iS2{i3} ], op = add, initial value = float(0), allreduce = false )
T4_l[ iS32{i0}, iS33{i2}, iS34{i3} ]
   = T1_g[ iS27{i0}, iS28{i2}, iS29{i3} ]
   - T2_g[ iS30{i0}, iS31{i2}, bS8{1} ];
T5_g[ iS37{i0}, iS38{i2}, iS39{i3} ]
   = T4_l[ iS32{i0}, iS33{i2}, iS34{i3} ]
   * T3_g[ iS35{i0}, iS36{i2}, bS11{1} ];
T7_l[ iS21{i0}, iS22{i2}, iS23{i3} ]
   = T0_g[ iS0{i0}, iS1{i2}, iS2{i3} ]
   * T5_g[ iS37{i0}, iS38{i2}, iS39{i3} ];
T8_g[ rS24{i0}, rS25{i2}, iS26{i3} ]
   = reduction( T7_l[ iS21{i0}, iS22{i2}, iS23{i3} ], op = add, initial value = float(0), allreduce = false )
}
```

### DIFF
--- a/csrc/scheduler/reduction_utils.cpp
+++ b/csrc/scheduler/reduction_utils.cpp
@@ -323,44 +323,6 @@ std::vector<int64_t> addBackBroadcasts(
   return axes;
 }
 
-// Check if a reduction is effectively an allreduce.
-bool isGridAllreduce(TensorView* reduction_tv) {
-  // Only Local tensor is converted to allreduce
-  if (reduction_tv->getMemoryType() != MemoryType::Local) {
-    return false;
-  }
-
-  // Collect all reduction parallel types
-  ParallelTypeBitmap reduction_parallel_types;
-  std::for_each(
-      reduction_tv->getLeafDomain().begin(),
-      reduction_tv->getLeafDomain().end(),
-      [&](auto id) {
-        if (id->isReduction() &&
-            isParallelTypeBlockDim(id->getParallelType())) {
-          reduction_parallel_types.set(id->getParallelType());
-        }
-      });
-
-  // If any of the reduction parallel types is used to parallelize
-  // the broadcast, it will be converted to an allreduce reduction expr
-  for (auto bcast_expr :
-       ir_utils::filterByType<BroadcastOp>(reduction_tv->uses())) {
-    auto bcast_tv = bcast_expr->out()->as<TensorView>();
-    if (std::any_of(
-            bcast_tv->getLeafDomain().begin(),
-            bcast_tv->getLeafDomain().end(),
-            [&](auto bcast_id) {
-              auto pt = bcast_id->getParallelType();
-              return isParallelTypeBlockDim(pt) &&
-                  reduction_parallel_types.get(pt);
-            })) {
-      return true;
-    }
-  }
-  return false;
-}
-
 void multiReductionInliner(
     Fusion* fusion,
     TensorView* reduction_tv,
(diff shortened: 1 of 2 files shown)


liqiangxl:
!build --diff

csrc/scheduler/reduction_utils.cpp
```diff
@@ -556,7 +556,8 @@ void propagateParallelization(
         std::back_inserter(allreduce_tvs),
         [&](auto tv) {
           return reduction_tv != tv &&
-              reduction_scheduler_utils::isGridAllreduce(tv);
+              (reduction_scheduler_utils::isGridAllreduce(tv) ||
```
jjsjann123, line 559: Naive question, could you elaborate why we should propagate the ParallelType::Group to `isGridAllreduce` tensors?

liqiangxl: Looking at the code, I think we don't need this `isGridAllreduce` check. I don't know why it was there, maybe it was used because we will use grouped reduction if the fusion has GridAllReduce. I deleted it and let's see how CI goes.

liqiangxl:
!build

csrc/scheduler/reduction_utils.cpp
naoyam: I think the reason was grouping was only supported in certain cases. Looks like the condition of applying grouping is now changed to follow the `use_grouped_reduction`. What does this change mean? Is there any case where previously grouping was used but not after this PR?

liqiangxl: Previously (before adding iter grouped outer reducion), the grouped reduction is only allowed for outer-grid-persistent kernels. After adding `iter grouped outer reducion` we extends to non-persistent outer grid/block reduction. There is no change where `previously grouping was used but not after this PR`.

I think these code assumes that for fusion  with `GridAllReduce`  Tvs, we will always do grouped reduciton, which is true as in the current `schedulePersistentKernel`, we have:
```
  const bool is_outer_grid_persistence = rparams.persistent_kernel &&
      rparams.cross_grid_inner_reduction && !rparams.fastest_dim;
  reduction_scheduler_utils::multiReductionInliner(
      fusion,
      reduction_tvs[0],
      reference_tv,
      unroll,
      vectorize,
      is_outer_grid_persistence,
      reduction_tvs,
      cached_inputs,
      cached_outputs,
      dummy_outputs);
```
The para `is_outer_grid_persistence` corresponds to `use_grouped_reduction`. Maybe we should rename it.

naoyam (approved):
LGTM. Did you also do the diff check?

liqiangxl:
!build --diff

liqiangxl:
> LGTM. Did you also do the diff check?

Not after removing `GridAllReduce` checks. Just started one. In theory, there should be no change since it's propagated from the reference reduction tv. If there is no `parallelType::Group`, there is no way to create a  `parallelType::Group` for other tvs.

PR #2333, opened 2024-06-03
Rename rfactor domain as logical domain

zasdfgbnm:


csrc/ir/internal_base_nodes.h
```diff
@@ -553,7 +553,7 @@ class TensorDomain : public Val {
     return !allocation_domain_.empty();
   }
 
-  // Returns if rfactor domain only consists of id's of iter type.
+  // Returns if logical domain only consists of id's of iter type.
   bool hasViewLikeRFactor() const;
```
zasdfgbnm, line 557: I am not renaming this in this PR. I do not have a good name for it, especially, I don't think `hasViewLikeLogical` is not a good name. I will leave it like this. Likely, in the future, I will rename it as `hasViewLikeProducerProjection` together with renaming the root domain as producer projection.

zasdfgbnm:
!build --diff --diff-bench

zasdfgbnm:
!build

naoyam (approved):
Thanks for the cleanup! LGTM.

PR #2336, opened 2024-06-04
Create Python Bindings for FusionProfile Object and `FusionDefinition` methods for usage

kevinstephano:
This PR does not include any special printing of the `FusionProfile` like is found if you use `NVFUSER_PROF=print`.  It only always the user to profile and query the `FusionProfile` and `KernelProfile` objects to keep this PR simple.

Additional Code Cleanup found in this PR:
- I rearranged some of the `FusionDefinition.execute()` parameters such that all parameters except for `inputs` are `kwargs` and they are consistent in parameter order across python and C++.
- Removed CPP printing methods from `FusionProfiler` I had originally envisioned printing multiple `FusionDefinition`s from python similarly to how the `NVFUSER_PROF=print` does it from C++ but I decided that was a bad idea.  I scaled back the ambition to producing a profile per `FusionDefinition` to keep it simple and rely on  a `FusionProfile` dunder to create a pretty print (not part of this PR) to remove the need for special handling of an ENV variable from Python.
- Removed an unused include for `complex` in the python_bindings.cpp

Example usage:

```
import torch
from nvfuser import FusionDefinition

inputs = [
        torch.randn(3072, 2048, device='cuda'),
        torch.randn(3072, 2048, device='cuda'),
        ]

    t1 = fd.from_pytorch(inputs[1])
    t2 = t0 + t1
    t3 = fd.ops.sum(t2, dim=-1)
    t4 = fd.ops.sum(t3, dim=-1)
    fd.add_output(t4)

for _ in range(20):
    out = fd.execute(inputs, profile=True)

print("Segments:", fd.profile().segments)
print("Host Time:", fd.profile().host_time_ms, "ms")
for idx in range(0, fd.profile().segments) :
    print("Seg Time:", idx, fd.profile().kernel_profiles[idx].time_ms, "ms")
```
Example Output:
```
Segments: 2
Host Time: 0.058019 ms
Seg Time: 0 0.005728 ms
Seg Time: 1 0.030208 ms
```


kevinstephano:
!build

kevinstephano:
!build

csrc/python_frontend/python_bindings.cpp
```diff
@@ -488,6 +488,80 @@ void initNvFuserPythonBindings(PyObject* module) {
         return ss.str();
       });
 
+  //! KernelProfiles are encapsulated in FusionProfiles where each KP
+  //! is associated with a segment.
+  py::class_<KernelProfile> kernel_prof(nvfuser, "KernelProfile");
```
Priya2698, line 493: Can we also had the scheduler heuristic used per segment?

kevinstephano: Yes, it will be a separate PR to add a new field.

kevinstephano:
!build

csrc/python_frontend/fusion_definition.h
```diff
@@ -128,9 +128,10 @@ class NVF_API FusionDefinition : public FusionState {
   //! Executes a fusion if a valid definition or cache lookup occurred prior
   NVF_API std::vector<at::Tensor> execute(
       const at::ArrayRef<c10::IValue>& inputs,
+      std::optional<int8_t> device,
```
Priya2698, line 131: nit: Keep the arg names between header and source file.
```Suggestion
      std::optional<int8_t> selected_device,

File: csrc/python_frontend/python_bindings.cpp [hunk 3]
```diff
@@ -488,6 +488,80 @@ void initNvFuserPythonBindings(PyObject* module) {
...
+  kernel_prof.def_property_readonly(
```
Priya2698, line 494: Is there a code comment/ wiki somewhere which lists the descriptions of these field? While some are self-explanatory, others like `stream`, `correlation_id` are not straightforward.

Priya2698 (approved):
LGTM. It would be good to have a list/description of all the properties of the profiler object somewhere.

PR #2343, opened 2024-06-05
Cleanup unary op with dispatch

zasdfgbnm:


lib/dynamic_type/test/unary_ops.cpp
```diff
@@ -21,43 +21,67 @@
...
 TEST_UNARY_OP(Negative, -, Int);
 TEST_UNARY_OP(BinaryNot, ~, Int);
-TEST_UNARY_OP(LogicalNot, !, Bool);
-
 #undef TEST_UNARY_OP
 
+TEST_F(DynamicTypeTest, LogicalNot) {
+  static_assert(!opcheck<DoubleInt64Bool>);
+  static_assert(!opcheck<DoubleInt64BoolVec>);
+  static_assert(std::is_same_v<decltype(!DoubleInt64Bool(2L)), bool>);
```
zasdfgbnm, line 45: This now returns `bool` because `!double`, `!int64_t` and `!bool` are all bool, so the return value does not have to be dynamic.

zasdfgbnm:
!build

naoyam (approved):
LGTM

PR #2345, opened 2024-06-05
Don't invalidate existing ValGroup when not necessary

zasdfgbnm:
Currently, functions in `id_model/schedule.{h,cpp}` will invalidate the input `ValGroup` because it will create a new `ValGroup` with a new ID and map that new group with the input group. This is annoying and unnecessarily expensive. This PR changes the behavior by just appending the new ID into the existing `ValGroup`.

### DIFF
--- a/csrc/disjoint_set.h
+++ b/csrc/disjoint_set.h
@@ -282,8 +282,8 @@ class VectorOfUniqueEntries {
 template <typename T, typename Hash = std::hash<T>>
 class DisjointSets {
  public:
-  using DisjointSetMap = std::
-      unordered_map<T, std::shared_ptr<VectorOfUniqueEntries<T, Hash>>, Hash>;
+  using DisjointSet = std::shared_ptr<VectorOfUniqueEntries<T, Hash>>;
+  using DisjointSetMap = std::unordered_map<T, DisjointSet, Hash>;
 
   DisjointSets() = default;
 
@@ -309,8 +309,7 @@ class DisjointSets {
 
   // Warning: returned values should never be modified. This accessor isn't
   // strictly safe as VectorOfUniqueEntries is not returned as a const.
-  const std::vector<std::shared_ptr<VectorOfUniqueEntries<T, Hash>>>&
-  disjointSets() const {
+  const std::vector<DisjointSet>& disjointSets() const {
     return disjoint_sets_;
   }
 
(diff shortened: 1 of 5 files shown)


zasdfgbnm:
!build

zasdfgbnm:
`mapVals` always destroy the existing ValGroups and create a new ValGroup with all values.

naoyam:
I see. One concern I have is this is correct right now, but if we added something more to `initializeVal`, we would also need to make sure that's done in `appendToGroup` as well. If that's only one, I think that'd be manageable, but we should probably not have too many of these additional interfaces. What do you think?

zasdfgbnm:
> I see. One concern I have is this is correct right now, but if we added something more to `initializeVal`, we would also need to make sure that's done in `appendToGroup` as well. If that's only one, I think that'd be manageable, but we should probably not have too many of these additional interfaces. What do you think?

I do not foresee more interfaces.

Also, I looked at the existing use of `initializeVal` in `id_model.cpp`, primarily there are two patterns:
1. Add a new Val to the graph on its own group.
2. Add a new Val to the graph to an existing group.

The case 2 is what I am optimizing in this PR, and it is also used for loop promotion:
https://github.com/NVIDIA/Fuser/blob/ad183eeadd613bf36c6032ced5553d1bd3542767/csrc/id_model/id_model.cpp#L755-L756

So I think these two cases are both common use cases of adding new Vals.

Regarding interface design, do you think we should rename `appendToGroup` to `initializeVal` while keeping its signature, and move their code near each other, so that `initializeVal` is the only name that can add new Vals to the graph? With the current naming, they look totally unrelated, but actually they are pretty related, and are often used together.

naoyam:
> > I see. One concern I have is this is correct right now, but if we added something more to `initializeVal`, we would also need to make sure that's done in `appendToGroup` as well. If that's only one, I think that'd be manageable, but we should probably not have too many of these additional interfaces. What do you think?
> 
> I do not foresee more interfaces.
> 
> Also, I looked at the existing use of `initializeVal` in `id_model.cpp`, primarily there are two patterns:
> 
> 1. Add a new Val to the graph on its own group.
> 2. Add a new Val to the graph to an existing group.
> 
> The case 2 is what I am optimizing in this PR, and it is also used for loop promotion:
> 
> https://github.com/NVIDIA/Fuser/blob/ad183eeadd613bf36c6032ced5553d1bd3542767/csrc/id_model/id_model.cpp#L755-L756
> 
> So I think these two cases are both common use cases of adding new Vals.
> 

Agreed.

> Regarding interface design, do you think we should rename `appendToGroup` to `initializeVal` while keeping its signature, and move their code near each other, so that `initializeVal` is the only name that can add new Vals to the graph? With the current naming, they look totally unrelated, but actually they are pretty related, and are often used together.

Yeah, I think `initializeVal` would make more sense.

zasdfgbnm:
renamed, and updated the comment

naoyam (approved):
LGTM. Thanks for the change. Makes me much more confident.

PR #2353, opened 2024-06-06
IdModel-based indexing: broadcast indexing

naoyam:
Stacked on top of #2344. Adds support of broadcast indexing with loop promotion. The main change is just the use of promoted domains in loop and allocation domains.


naoyam:
CC: @zasdfgbnm

csrc/id_model/indexing.cpp
```diff
@@ -339,6 +372,27 @@ void TensorIndexer::buildLoopIndexMap() {
   }
 }
 
+bool TensorIndexer::shouldUseZeroIndex(const ValGroup& loop_group) const {
+  // All loops in this set are non-parallel, non-concretized broadcast
```
jacobhinkle, line 376: So if all the axes are broadcast then we should use 0 index, or if the promoted loop has extent 1 (and is not partial). How does "non-parallel" affect this check?

csrc/id_model/indexing.cpp
```diff
@@ -399,10 +480,18 @@ Val* TensorIndexer::getLinearIndex(TensorView* tv, const Expr* expr) {
   return index;
 }
 
+// Get the loop domains of a given expr, which are (potentially
+// promoted) loop domains of the consumer tensor.
 std::vector<IterDomain*> TensorIndexer::getLoopDomains(const Expr* expr) const {
   // Assume consumer-based indexing. Needs to revisit for ops like
   // scatter
-  return ir_utils::getTvOutput(expr)->getLeafDomain();
+  auto loop_domains = ir_utils::getTvOutput(expr)->getLeafDomain();
```
jacobhinkle, line 488: Another place where we assume all outputs have same domain; in this case leaf domain.

csrc/id_model/indexing.h
```diff
@@ -82,6 +82,11 @@ class TensorIndexer {
       const Expr* expr,
       const std::vector<IterDomain*>& index_domains) const;
 
+  // Check if the loop index of a a loop group should be always
```
jacobhinkle, line 85:
```suggestion
  // Check if the loop index of a loop group should be always
```

tests/cpp/test_indexing.cpp
```diff
@@ -21,6 +21,8 @@
 #include <ops/all_ops.h>
 #include <scheduler/utils.h>
 
+#include <functional>
```
jacobhinkle, line 24: `std::forward` is defined in `<utility>` so `<functional>` might not be needed.

tests/cpp/test_indexing.cpp
```diff
@@ -35,6 +37,26 @@ std::vector<Val*> getLoopIndices(TensorView* tv, const TensorIndexer& indexer) {
   return loop_indices;
 }
 
+template <typename... Args>
+Val* addExpr(Args&&... args) {
+  return SimplifyingIrBuilder::addExpr(std::forward<Args>(args)...);
```
jacobhinkle, line 42: These do help readability. If you're planning to do a lot of structural checking of indices we could consider using [user-defined literals](https://en.cppreference.com/w/cpp/language/user_literal) like what @zasdfgbnm used in `test_expr_simplifier.cpp`.

jacobhinkle (approved):
LGTM

csrc/id_model/indexing.cpp
```diff
@@ -339,6 +372,27 @@ void TensorIndexer::buildLoopIndexMap() {
...
+bool TensorIndexer::shouldUseZeroIndex(const ValGroup& loop_group) const {
+  // All loops in this set are non-parallel, non-concretized broadcast
+  //  iterdomains, their "index variable" should be zero. This
+  //  condition should be included in the next triviality check, but
+  //  just checking isBroadcast should be more efficient.
+  if (std::all_of(loop_group->begin(), loop_group->end(), [](Val* val) {
+        return val->as<IterDomain>()->isBroadcast();
+      })) {
+    return true;
+  }
```
zasdfgbnm, line 384: Is this sufficient?
```C++
  auto leaf_id =
      getLoopPromotion(loop_group->front()->as<IterDomain>(), id_model_);
  leaf_id->isBroadcast();
```

csrc/id_model/indexing.cpp
```diff
@@ -339,6 +372,27 @@ void TensorIndexer::buildLoopIndexMap() {
...
+
+  // Trivial loop
+  auto leaf_id =
+      getLoopPromotion(loop_group->front()->as<IterDomain>(), id_model_);
+  if (!leaf_id->maybePartial() && simplifyExpr(leaf_id->extent())->isOneInt()) {
```
zasdfgbnm, line 389: Do we support partial IterDomains?

csrc/id_model/indexing.cpp
```diff
@@ -120,6 +137,20 @@ std::tuple<std::vector<IterDomain*>, std::vector<Val*>> getAllocationDomains(
...
+  // for now.
+  for (auto& allocation_domain : allocation_domains) {
+    bool is_loop = std::find(
+                       tv->getLeafDomain().begin(),
+                       tv->getLeafDomain().end(),
+                       allocation_domain) != tv->getLeafDomain().end();
+    if (!is_loop) {
+      continue;
+    }
+    allocation_domain = getLoopPromotion(allocation_domain, id_model);
```
zasdfgbnm, line 151: `isPartitionedLoop` uses the parallel type of `id` instead of the `getParallelType` of the loop group of `id`, can this be a problem? Similarly, in line 117 above, we are also using `loop_id->getParallelType()` instead of the parallel type of the loop group. IIRC, if the loop promotion id is replayed, we will not set its parallelization type.

csrc/id_model/indexing.cpp
naoyam: Ah, that comment should have been removed. I had some other code above this line. Thanks for catching it.

csrc/id_model/indexing.cpp
zasdfgbnm: If I have
```
smem_tv[b, I1] ca_pos(1)
tv[TIDx, I1] = smem_tv[b, I1] + concrete_tv[I0, I1]
```
then should `smem_tv` be allocated as `[I0, I1]` or `[I1]`?

How about 
```
smem_tv[bTIDx, I1] ca_pos(1)
tv[TIDx, I1] = smem_tv[bTIDx, I1] + concrete_tv[I0, I1]
```
?

csrc/id_model/indexing.cpp
naoyam: No. I think I just put this mostly by following what we have in `kir::ForLoop::isTrivial` before the shift removal. Now that it's removed, this seems more confusing than necessary. I'll remove it.

csrc/id_model/indexing.cpp
naoyam: Yes. I think lifting that restriction is quite challenging. We would need to change many things, including expression sorting, loop generation etc.

csrc/id_model/indexing.cpp
```diff
@@ -120,6 +137,20 @@ std::tuple<std::vector<IterDomain*>, std::vector<Val*>> getAllocationDomains(
     }
   }
 
+  // Loop promotion may affect allocations. Promotions of intermediate
+  // domains may not be defined correctly. Only consider loop domains
+  // for now.
```
zasdfgbnm, line 142: image
What if we have the above schedule? How should we handle this? Should we just raise an error, or allocate the tv with broadcasting as `I0*I1`, or allocating it as `5`?

csrc/id_model/indexing.cpp
naoyam: Thank you. That's a real problem. We should either always look at a loop group or just propagate parallel types to all IDs, including promotion domains. I think the latter is a simpler solution. Will work on it.

naoyam: > If I have
> 
> ```
> smem_tv[b, I1] ca_pos(1)
> tv[TIDx, I1] = smem_tv[b, I1] + concrete_tv[I0, I1]
> ```
> 
> then should `smem_tv` be allocated as `[I0, I1]` or `[I1]`?
> 
> How about
> 
> ```
> smem_tv[bTIDx, I1] ca_pos(1)
> tv[TIDx, I1] = smem_tv[bTIDx, I1] + concrete_tv[I0, I1]
> ```
> 
> ?

In both cases, the allocation size should be `[I0, I1]`, right?

naoyam: Maybe not. In the first case, `[I1]` should be enough as long as a proper predicate is added.

zasdfgbnm: I think for both case, `[I1]` is enough. I think it's just a matter of whether `[I1]` or `[I0, I1]` is easier for analysis.

naoyam: I think what we currently do is we don't inline broadcast domains like `I0` of `smem_tv`. Let me check.

naoyam: This is what I was referring to:

https://github.com/NVIDIA/Fuser/blob/main/csrc/tensor_view.cpp#L167-L170

But that only applies to innermost broadcast domains, so in this case it is inlined.

csrc/id_model/indexing.cpp
naoyam: When the allocation domain is not the loop domain, we don't have any logic other than fully allocating `bxI1`. If we want to just allocate `5`, we could do it by setting the allocation domain as the loop domain.

I wonder what the domains and parallelization would look like with TMA. 

I believe this is more of a question on what the allocation domain should be. I think that ideally `getAllocationDomains` here should be just a trivial function call to `tv->getAllocationDomain` and each tensor should always have its correct allocation domains.

csrc/id_model/indexing.cpp
naoyam: The current lowering only allocates `[I1]`. More specifically, when a domain is a broadcast domain, it's not allocated even when it's promoted. I'd keep this behavior as is.

naoyam: Looking at the code again, this should be already what happens (with #2371). Pure broadcast domains should be filtered out `!mayRequireAllocation`. I'll add a test.

naoyam: Added.

csrc/id_model/indexing.cpp
zasdfgbnm: > I believe this is more of a question on what the allocation domain should be.

I agree. I think we need a restriction that, each ID in the allocation domain must be either fully inlined or fully not inlined. It can not have an ID coming from a merge of an inlined ID with an uninlined ID, or an ID who is the parent of a split whose outer is inlined but inner not.

> I wonder what the domains and parallelization would look like with TMA.

In my mental model, it is similar to above: IDs in the allocation domain must be either a tile or a non-tile, it can not be a mix of both. However, in practice, for now, even if we have a mix, the code should still work (thanks to some existing hack in our system? and the same hack will also make the above example just allocate 5 today?)

naoyam: > > I believe this is more of a question on what the allocation domain should be.
> 
> I agree. I think we need a restriction that, each ID in the allocation domain must be either fully inlined or fully not inlined. It can not have an ID coming from a merge of an inlined ID with an uninlined ID, or an ID who is the parent of a split whose outer is inlined but inner not.

Yeah, I think you could say that the mechanism of promotion is making partially inlined domains fully inlined. In the above case, the innermost `5` domain of the broadcast tensor is promoted to the innermost `5` domain of the non-broadcast tensor, meaning it's effectively fully inlined.

I think we have a reasonable understanding of promotions of loop domains. Can we propagate promotions to allocation domains that are not between logical and loop? I guess we also have a similar problem of propagating parallel types from loop domains to allocation domains.
> 
> > I wonder what the domains and parallelization would look like with TMA.
> 
> In my mental model, it is similar to above: IDs in the allocation domain must be either a tile or a non-tile, it can not be a mix of both. However, in practice, for now, even if we have a mix, the code should still work (thanks to some existing hack in our system? and the same hack will also make the above example just allocate 5 today?)

As long as the allocation domain of the tensor is just the loop domain, indexing is trivial thanks to the loop promotion. How it's implemented in the current main branch isn't that different conceptually, but the implementation is quite convoluted since, for example, it always traverses back to logical domains, whereas in this case we can just directly index the loop (=allocation) domains.

Issue #2346, opened 2024-06-05
CpAsync with shared memory predicate is not permitted


liqiangxl:
// A cp.async op would need a predicate for either the global
    //  input or its shared mem output, or both.
    // Due to the WAR discussed in [Predicate Inversion for CpAsync],
    //  we currently cannot support use cases where both the gmem read
    //  and the smem write need to be predicated.
    // Adding a check here would make the exclusion of such case as precise as
    //  possible and avoid duplication of predicateSharedMemAccess
    //  logic. But this part along with [Predicate Inversion for CpAsync]
    //  should be cleaned up all together when we extend predicate/masking
    //  logic to cover this usage.
    NVF_ERROR(
        !(ir_utils::isCpAsyncOp(expr) && needs_predicate_smem_access),
        "predicate removal: unsupported use case of cp.async");

Why?
see a detailed [explanation](https://github.com/NVIDIA/Fuser/pull/2339#discussion_r1626517112)

Test to reproduce:
```
TEST_F(NVFuserTest, FusionCpAsyncPredicateError) {
  Fusion fusion;
  FusionGuard fg(&fusion);

  int m = 33, n = 48;
  TensorView* tv0 = makeContigTensor(2);
  fusion.addInput(tv0);
  auto tv1 = exp(tv0);
  fusion.addOutput(tv1);

  auto tvs = tv0->cacheAfter(LoadStoreOpType::CpAsync);
  tvs->setMemoryType(MemoryType::Shared);

  tvs->split(-1, 4);
  tvs->axis(-1)->parallelize(ParallelType::Vectorize);
  tvs->axis(-2)->parallelize(ParallelType::TIDx);
  tvs->axis(-3)->parallelize(ParallelType::BIDx);

  tv1->axis(-1)->parallelize(ParallelType::TIDx);
  tv1->axis(-2)->parallelize(ParallelType::BIDx);

  auto options = at::TensorOptions().dtype(at::kFloat).device(at::kCUDA, 0);
  at::Tensor t0 = at::randn({m, n}, options);

  FusionExecutor fe;
  EXPECT_THAT(
      [&]() { fe.compileFusion(&fusion, {t0}); },
      ::testing::ThrowsMessage<nvfuser::nvfError>(
          ::testing::HasSubstr("unsupported use case of cp.async")));
}
```

PR #2357, opened 2024-06-06
prioritize moving buffer used by outer bcast

liqiangxl:
**Issue:** innerOuter scheduler moves buffer from register to shared memory to save register usage. However, regular load from global memory to shared memory still requires temp usage of registers. This increased register pressure and caused performance drops.

**Fix:** Prioritize moving buffers used by outer broadcast tensors to shared memory. Because: 
(1) They are reused in every iteration of the outer loop, has lower IO. 
(2) Load occurs before the outer loop. Temporary register usage won't increase register pressure since the loop is the high-pressure region. 

**Other options:**
These two additional options can be used with the fix in this PR at the same time.
(1) use cpAsync which directly goes from gmem to smem. Paused due to predicate resitrictions see issue #2346 
(2) further reduce register usage to avoid spills. Will investigate later

**Performance:**
(1) RMS Norm Bwd, NvFuserScheduler_RMSNorm_BWD_fp32, Batch size = 2112 (compare this PR & main branch)
image

(2) All RMS Norm Bwd & LN Bwd in cpp benchmarks (compare this PR & main branch)
No regression, speedup can be up to 1.4x
image


liqiangxl:
The failed case is `IndexingTest.SimplePointwise1` and fixed in main.

naoyam:
Have you done the adjustment of the spill-vs-shared factors? If not, let's do that first.

liqiangxl:
> Have you done the adjustment of the spill-vs-shared factors? If not, let's do that first.

Not yet. Let me change this to draft and check that.

jjsjann123:
> Have you done the adjustment of the spill-vs-shared factors? If not, let's do that first.

For my own education, this is just adjusting
`int64_t available_regs = scheduler_utils::register_file_size_full;`
to be something like 
`int64_t available_regs = static_cast<int64_t>(scheduler_utils::register_file_size_full * magic_factor);`?

And we maybe want to do a coarse sweep to pick the magic_factor?

naoyam:
> > Have you done the adjustment of the spill-vs-shared factors? If not, let's do that first.
> 
> For my own education, this is just adjusting `int64_t available_regs = scheduler_utils::register_file_size_full;` to be something like `int64_t available_regs = static_cast<int64_t>(scheduler_utils::register_file_size_full * magic_factor);`?
> 
> And we maybe want to do a coarse sweep to pick the magic_factor?

Yes, but the current is:

```
int64_t available_regs = scheduler_utils::register_file_size_full * magic_factor;
```

where `magic_factor > 1`

liqiangxl:
> > > Have you done the adjustment of the spill-vs-shared factors? If not, let's do that first.
> > 
> > 
> > For my own education, this is just adjusting `int64_t available_regs = scheduler_utils::register_file_size_full;` to be something like `int64_t available_regs = static_cast<int64_t>(scheduler_utils::register_file_size_full * magic_factor);`?
> > And we maybe want to do a coarse sweep to pick the magic_factor?
> 
> Yes, but the current is:
> 
> ```
> int64_t available_regs = scheduler_utils::register_file_size_full * magic_factor;
> ```
> 
> where `magic_factor > 1`

I summarized results using different `magic_factors`, see the [doc](https://docs.google.com/document/d/1XljtycM4aLN_1x_wa5EKf4NgrbX9J1maoi2xqWRaOTM/edit?usp=sharing).

liqiangxl:
@naoyam and @jjsjann123 , as discussed in last sync meeting, I added the explore of different allowed regiser count into this PR since it is just 2 lines of code change and directly related to inner-outer scheduler.

jjsjann123:
> (1) from 23K to 24K, this PR moves 3 buffers to smem while main branch moves 2 buffers. This can be solved by revised heuristics.

QQ: what's the heuristics change to address this?

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -317,6 +317,42 @@ struct PersistentBufferStorageParams {
   bool project_to_input = false;
 };
 
+// Prioritize moving buffers used by outer broadcast tensors to shared memory
+// because:
+// (1) They are reused in every iteration of the outer loop, has lower IO.
+// (2) Load occurs before the outer loop. Temporary register usage won't
+//     increase register pressure since the loop is the high-pressure region.
```
jjsjann123, line 324: > since the loop is the high-pressure region.

Just to confirm that I read this correctly: bcast_buffer is loaded before the outer loop, which impacts register usage and that's why we are moving it to shared memory.

jjsjann123:
I also have a question regarding our benchmark. I thought we have a performance regression on hidden size 28k-32k.Yet I'm not seeing much regression in the end of the doc for our cpp benchmark. (And RMS norm bwd is not showing any regression neither.)

csrc/scheduler/normalization_inner_outer.cpp
jjsjann123: Note to myself. offline discussion answered that my understanding is correct. I'll take another look at the code comment.

liqiangxl:
> I also have a question regarding our benchmark. I thought we have a performance regression on hidden size 28k-32k.Yet I'm not seeing much regression in the end of the doc for our cpp benchmark. (And RMS norm bwd is not showing any regression neither.)

(1) LN bwd in cpp benchmark doesn't cover hidden size 28k-32k
(2) RMS norm bwd doesn't have `bias`, it has only 1 outer redction instead of 2, so uses 10 bytes per elment instead of 14, so only need to move 2 buffers to smem when hidden size <= 37K. Futher increase hidden size may see similar regressions.

liqiangxl:
> > (1) from 23K to 24K, this PR moves 3 buffers to smem while main branch moves 2 buffers. This can be solved by revised heuristics.
> 
> QQ: what's the heuristics change to address this?

revise threads per block & persistent batch size, e.g. at 24K threads per block increased from 256 to 512 and persistent batch size decreased from 12 to 6 (not in this PR). The same change may also speedup cases with 64K registers but not investigated since we want to avoid register spills.

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -317,6 +317,42 @@ struct PersistentBufferStorageParams {
...
+  std::stable_sort(
+      idxs.begin(), idxs.end(), [&is_used_by_outer_bcast](int i, int j) {
+        return is_used_by_outer_bcast[i] && !is_used_by_outer_bcast[j];
+      });
+  std::vector<TensorView*> sorted_candidate_tvs;
+  sorted_candidate_tvs.reserve(n_buffer_inputs);
+  for (auto idx : idxs) {
+    sorted_candidate_tvs.emplace_back(projectable_buffer_inputs.at(idx));
+  }
+  return sorted_candidate_tvs;
```
jjsjann123, line 353: nitpick on coding here: if we have `std::unordered_map<TensorView*, bool> is_used_by_outer_bcast` instead of a vector. I think we can directly sort `std::vector<TensorView*>` without having idxs and map it back?

jjsjann123 (approved):
sorry this falls off my radar. < 50 LOC PR shouldn't have taken a month for review. :bow: 

stamping since the PR looks good to my eyes. If we have questions regarding the sorting function, we can further discuss in follow up issue/PR.

jjsjann123:
CI failure likely is from upstream. I think this one: https://github.com/pytorch/pytorch/pull/129505#issuecomment-2223570906

csrc/scheduler/normalization_inner_outer.cpp
liqiangxl: good point.

liqiangxl: Revised. Also post A100 results.
image

jjsjann123:
!build

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -317,6 +317,36 @@ struct PersistentBufferStorageParams {
...
+
+  // sort based on [is_used_by_outer_bcast]
+  std::vector<TensorView*> sorted_buffer = projectable_buffer_inputs;
+  std::sort(
+      sorted_buffer.begin(),
+      sorted_buffer.end(),
+      [&](TensorView* a, TensorView* b) {
+        return is_used_by_outer_bcast[a] && !is_used_by_outer_bcast[b];
+      });
+  return sorted_buffer;
```
jjsjann123, line 347: awesome. much easier to look at for my monkey brain :smile:

liqiangxl:
!build

liqiangxl:
!build

liqiangxl:
`nvfuser-ci/jit_python_bc_advisory_17_A100 — Failing after 37 minutes https://nv/e2E/101797414` is a know issue, see #2617

PR #2360, opened 2024-06-06
Tensor parallel MLP 

cowanmeg:
Manually sharded tensor parallel multilayer perception layer.

Input is manually translated and sharded mlp layer taken from nanoGPT. See https://github.com/NVIDIA/Fuser/issues/2199 for where we get the initial compute trace.


cowanmeg:
Some followup not addressed by this PR:
(1) Linear node: Need to update to allow 2D local weight tensors (3D+ with device dims) https://github.com/NVIDIA/Fuser/blob/beb22870927e9d903b238092365f6de55c3e914b/csrc/ops/composite.cpp#L102
(2) Symbolic TVs:  This error is encountered with symbolic TVs C++ exception with description "ext_opt.hasValue() INTERNAL ASSERT FAILED at "csrc/dynamic_transform.cpp":276, Could not evaluate dynamic extent: i3
Exception raised from DynamicTransformConcretizationInfo at
csrc/dynamic_transform.cpp:276
(3) Improved sharding propagation: TVs where sharding propagation breaks down are (1) broadcasts where a device dim axis is broadcasted (2) `rand_like` operator because it creates a fresh new TV. The current pass assumes that one of the inputs is already sharded and propagates from producer to consumer. To support this we need to propagate up from the consumer.

tests/cpp/test_multidevice_matmul.cpp
```diff
@@ -404,4 +404,167 @@ TEST_F(DistributedMatmulTest, Matmul_LayoutNT_ReduceScatter) {
...
+  TensorView* matmul2 = sum(local_matmul2, {0}); // Allreduce
+  TensorView* bcast_bias = broadcast(b1, {true, false});
+  TensorView* linear2 = add(matmul2, bcast_bias);
+
+  // Dropout
+  // Note: Propagation breaks at rand_like because it creates a fresh TV.
+  // Temporarily this prevents us from using dropout composite node.
+  TensorView* linear2_ = castOp(DataType::Float, linear2);
+  const double kProb = 0.1;
+  const double kScale = 1.0 / (1.0 - kProb);
```
wujingyue, line 465:
```suggestion
  constexpr double kProb = 0.1;
  constexpr double kScale = 1.0 / (1.0 - kProb);
```

tests/cpp/test_multidevice_matmul.cpp
```diff
@@ -404,4 +404,167 @@ TEST_F(DistributedMatmulTest, Matmul_LayoutNT_ReduceScatter) {
...
+TEST_F(DistributedMatmulTest, MLP_Layer) {
+  std::unique_ptr<Fusion> fusion = std::make_unique<Fusion>();
+  FusionGuard fg(fusion.get());
+  auto mesh = DeviceMesh::createForNumDevices(communicator_->size());
+
+  int64_t sb = 64; // sequence * batch
+  int64_t h = 128;
+  int64_t h4 = 4 * h;
+
+  // TODO: error with dynamic shape
```
wujingyue, line 417: Can you clarify this? Are you saying the following code would fail if changed to makeContigTensor?

tests/cpp/test_multidevice_matmul.cpp
```diff
@@ -404,4 +404,167 @@ TEST_F(DistributedMatmulTest, Matmul_LayoutNT_ReduceScatter) {
...
+  TensorView* linear_int3 = sum(linear_int2, {-1});
+  TensorView* linear_int4 = broadcast(b0, {false, true, false});
+  TensorView* linear1 = add(linear_int3, linear_int4);
+
+  TensorView* linear1_ = castOp(DataType::Float, linear1);
+  TensorView* gelu = tanh_gelu(linear1_);
+  TensorView* gelu_ = castOp(DataType::BFloat16, gelu);
+
+  // Linear #2
+  gelu_ = segment_set(gelu_);
```
wujingyue, line 450: Can you comment on why this is needed?

tests/cpp/test_multidevice_matmul.cpp
```diff
@@ -404,4 +404,167 @@ TEST_F(DistributedMatmulTest, Matmul_LayoutNT_ReduceScatter) {
...
+  fusion->addInput(b0);
+  fusion->addInput(w1);
+  fusion->addInput(b1);
+
+  // Linear #1
+  // Notes: Manually breaking down the linear layer into nvfuser primitives
+  TensorView* linear_int0 = broadcast(x, {true, false, true, false});
+  TensorView* linear_int1 = broadcast(w0, {false, true, false, false});
+  TensorView* linear_int2 = mul(linear_int0, linear_int1);
+  TensorView* linear_int3 = sum(linear_int2, {-1});
```
wujingyue, line 441: Use matmul or linear? cc @Priya2698

wujingyue (commented):
Almost LGTM. Nice work!

tests/cpp/test_multidevice_matmul.cpp
cowanmeg: Without the segment set, the reduction scheduler gets called instead of the matmul scheduler. Can add a comment in the code as well.

tests/cpp/test_multidevice_matmul.cpp
cowanmeg: The first linear layer can be replaced with a linear op because there is no communication in between the bias add and local matmul. The second linear layer will have to be a matmul since the pattern is matmul + allreduce + bias add. 
I do have tests with replacing the second linear layer with matmul, but it was generating very high error. My assumption is that the `testValidate` isn't meant to be used when chaining multiple matmuls together? I also fiddled around with the datatypes of the matmul (i.e. bfloat, float, double) but the error was still high.

tests/cpp/test_multidevice_matmul.cpp
cowanmeg: btw, IIUC we should use `matmul` op for both nvfuser matmul and aten matmul in the future and this will address the segmentation issue we see here. We will just need to update our resharding passes to handle matmul and linear ops appropriately.

csrc/multidevice/utils.cpp
```diff
@@ -524,9 +527,14 @@ std::set<DeviceIdxType> involvedDevices(Expr* expr) {
   std::set<DeviceIdxType> ret;
   for (const auto& tvs : {expr->inputs(), expr->outputs()}) {
     for (auto val : tvs) {
-      NVF_ERROR(val->isA<TensorView>(), "Val is not a TensorView");
+      if (!val->isA<TensorView>()) {
+        continue;
+      }
```
wujingyue, line 532: I think we can simplify this to

```
for (const auto& tvs : {filterByType<TensorView>(expr->inputs()), filterByType<TensorView>(expr->outputs())}) {
  for (auto* tv : tvs) {
```

tests/cpp/test_multidevice_matmul.cpp
wujingyue: > matmul + allreduce + bias add

Is it crazy to do `matmul + add half bias + all_reduce`? This may be slightly faster, but we can explore that later. 

> it was generating very high error

Does the second linear layer in the PR as is map to pointwise+reduction or aten matmul? If the former, we'll need to fix it soon because pointwise+reduction is likely slower. If the latter, I'll be curious why aten matmul via mul+sum is low error than aten matmul via matmul.

tests/cpp/test_multidevice_matmul.cpp
wujingyue: > Can add a comment in the code as well.

That'd always be helpful. Thank you! 

> update our resharding passes to handle matmul and linear ops appropriately

I'm surprised using matmul/linear changes resharding at all because they are done locally. What do you mean?

cowanmeg: Ahh I should clarify, it would be in the `insertResharding` pass which automatically adds `set` operations where necessary. The work around is very simple,  we just add the `set` manually.

tests/cpp/test_multidevice_matmul.cpp
wujingyue: > matmul + allreduce + bias add

Also, mainstream (e.g. llama) doesn't seem to use bias. So it might not be relevant. 

> Currently, the second linear layer uses nvfuser matmul.

SG. In the next PR, let's try to use ATen matmul. This will be how the integration looks like and give the best performance at this moment. Even if the error is slightly higher than the threshold, I'd accept it because (1) we know the error is localized to matmul, and (2) checking in the code allows us to investigate in parallel.

tests/cpp/test_multidevice_matmul.cpp
wujingyue: SG. I'm fine leaving the segment_set as is. However, in the next (or next next :) PR, we should improve insertResharding to avoid the segment_set. Thunder or nvFuser's Python API doesn't have segment_set and it's too low-level for the framework to add segment_set properly anyway. How does that sound?

cowanmeg: If we use matmul instead of the broadcast+mul-sum no segment_set is needed, so in the next PR we will have a version with no segment set :)

cowanmeg:
@wujingyue - I added the MLP test with aten matmul. Note, that the tolerance is bumped up to a bit to pass validation. 

Validation error in output 0 (linear1) on line 583 in file /tests/cpp/test_multidevice_matmul.cpp.
  Detected abs error of: 0.122498
    absolute tolerance was set to 0.005
    and relative tolerance set to 5e-05


Validation error in output 2 (linear2) on line 583 in file tests/cpp/test_multidevice_matmul.cpp.
  Detected abs error of: 4.08847
    absolute tolerance was set to 2
    and relative tolerance set to 0.02

wujingyue (approved):

cowanmeg:
!build

tests/cpp/test_multidevice_matmul.cpp
```diff
@@ -404,4 +405,191 @@ TEST_F(DistributedMatmulTest, Matmul_LayoutNT_ReduceScatter) {
...
+  fusion->addInput(x);
+  fusion->addInput(w0);
+  fusion->addInput(b0);
+  fusion->addInput(w1);
+  fusion->addInput(b1);
+
+  // Linear #1
+  TensorView* matmul1;
+  if (use_aten_matmul) {
+    // TODO: use linear op instead
```
wujingyue, line 441: cc @Priya2698 

@cowanmeg reminded me of a practical limitation: currently, we split rfactor for DID, so w0 is 3D. `linear` (just as `torch.linear`) doesn't take a 3D weight. This limitation will eventually go away when split leaf instead of rfactor, but will exist likely for the rest of the year.

tests/cpp/test_multidevice_matmul.cpp
wujingyue: I think it's another instance of #2462. Please revisit when it's fixed.

cowanmeg: We are using FusionExecutorCache so sadly that did not fix the error. I did narrow down what is causing the error and opened an issue https://github.com/NVIDIA/Fuser/issues/2481

PR #2361, opened 2024-06-06
Add scheduling support for `SDPAOp`

Priya2698:
Stacked on #2294.

1. Adds the producer-consumer mapping to root domain map.
2. Adds `SDPAOp` to `ExprEvalScheduler`.
3. Modifies `ExprEvalSched::canSchedule` to skip computeAt checks and only use the compile time check since expression evaluator scheduler will only accept segments with a single expression of type MatmulOp / LinearOp / SdpaOp.,


Priya2698:
!build

csrc/root_domain_map.cpp
```diff
@@ -229,6 +229,23 @@ std::unordered_map<IterDomain*, IterDomain*> PairwiseRootDomainMap::map(
     return dom_map;
   }
 
+  if (SdpaFwdOp* op = dynamic_cast<SdpaFwdOp*>(consumer_tv_->definition())){
+    // Map N, H from any input (query/key/value)
```
naoyam, line 233: Can you please add comments like the one at line 219 for `LinearOp`?

naoyam (commented):
Is the root map change required for the scheduling support?

Priya2698:
> Is the root map change required for the scheduling support?

Yes. Otherwise, we noticed with the earlier Matmul and LinearOps, that the `exactMappedExtentSubstitution` presegmentation pass runs into error since the producer and consumers are not trivially aligned. I will verify that the error persists if this change is removed for SDPA as well.

jacobhinkle:
> Is the root map change required for the scheduling support?

I suspect there might be problems with `ExpressionEvaluator::propagateBoundValuesThroughExactMaps` if we did not map any IterDomains for this node (or MatmulOp` or `LinearOp`), but maybe it would be OK since we always re-use in the input extents directly so no mapping is needed?

naoyam:
Thanks. Looks good to me. I'll let @jacobhinkle give a stamp.

Priya2698:
> > Is the root map change required for the scheduling support?
> 
> I suspect there might be problems with `ExpressionEvaluator::propagateBoundValuesThroughExactMaps` if we did not map any IterDomains for this node (or MatmulOp`or`LinearOp`), but maybe it would be OK since we always re-use in the input extents directly so no mapping is needed?

It would still not work since, the third dimension (L/S) only maps from query to output.

```
C++ exception with description "known_size == this_size INTERNAL ASSERT FAILED at "/opt/pytorch/nvfuser/csrc/expr_evaluator.cpp":302, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. Conflicting sizes: 128, 64
Exception raised from propagateBoundValuesThroughExactMaps at /opt/pytorch/nvfuser/csrc/expr_evaluator.cpp:302 (most recent call first):
```

Priya2698:
!build

csrc/root_domain_map.cpp
```diff
@@ -229,6 +229,30 @@ std::unordered_map<IterDomain*, IterDomain*> PairwiseRootDomainMap::map(
...
+      if (idx < 2) {
+        updatePairwiseRootDomainMap(
+            producer_logical.at(idx), consumer_root.at(idx));
+      }
+      // Map L, E from query and value respectively
+      if (idx == 2 && producer_tv_->sameAs(op->query())) {
+        updatePairwiseRootDomainMap(
+            producer_logical.at(2), consumer_root.at(2));
+      }
+      if (idx == 3 && producer_tv_->sameAs(op->value())) {
```
jacobhinkle, line 248:
```suggestion
      // Map Ev from value to output
      if (idx == 3 && producer_tv_->sameAs(op->value())) {
```

csrc/root_domain_map.cpp
```diff
@@ -229,6 +229,30 @@ std::unordered_map<IterDomain*, IterDomain*> PairwiseRootDomainMap::map(
...
   }
 
+  if (SdpaFwdOp* op = dynamic_cast<SdpaFwdOp*>(consumer_tv_->definition())) {
+    // Consumers: output = [N, H, L, Ev], logsumexp = [N, H, L], cum_seq_q/k =
+    // [N + 1,] Producers: query = [N, H, L, E], key = [N, H, S, E], value = [N,
+    // H, S, Ev]
+
+    // Map N, H from any input (query/key/value)
+    for (auto idx : c10::irange(consumer_root.size())) {
+      if (idx < 2) {
```
jacobhinkle, line 239: For `idx == 0` and `consumer_tv_->sameAs(op->output(2)) || consumer_tv_->sameAs(op->output(3))`, these should not map since the extents differ by 1. I would just put a check before this that consumer is the output or logsumexp. Btw you might want to add accessors like `op->logSumExp()` to make this new condition more readable.

tests/cpp/test_sdpa_node.cpp
```diff
@@ -9,25 +9,36 @@
...
 namespace nvfuser {
 
-using SDPATest = NVFuserTest;
+class SDPATest : public NVFuserTest {
+ protected:
+  SDPATest() : optimization_guard_(false), allocation_order_guard_(false) {}
+
+ private:
+  // Note: `MoveSpliCat` and `AllocationDomain` preseg passes use ID model.
+  // `SdpaOp` currently does not work with ID model since it requires all
```
jacobhinkle, line 28:
```suggestion
  // Note: `MoveSplitCat` and `AllocationDomain` preseg passes use ID model.
  // `SdpaFwdOp` currently does not work with ID model since it requires all
```

jacobhinkle (commented):
Test looks good. My comments are mostly minor

csrc/root_domain_map.cpp
```diff
@@ -229,6 +229,30 @@ std::unordered_map<IterDomain*, IterDomain*> PairwiseRootDomainMap::map(
     return dom_map;
   }
 
+  if (SdpaFwdOp* op = dynamic_cast<SdpaFwdOp*>(consumer_tv_->definition())) {
+    // Consumers: output = [N, H, L, Ev], logsumexp = [N, H, L], cum_seq_q/k =
+    // [N + 1,] Producers: query = [N, H, L, E], key = [N, H, S, E], value = [N,
+    // H, S, Ev]
```
jacobhinkle, line 235: Readability suggestion:
```suggestion
    // Producers:
    //   query = [N, H, L, E]
    //   key = [N, H, S, E]
    //   value = [N, H, S, Ev]
    // Consumers:
    //   output = [N, H, L, Ev]
    //   logsumexp = [N, H, L]
    //   cum_seq_q/k = [N + 1]
```

csrc/root_domain_map.cpp
Priya2698: Even if I use `resize`? 
I saw an error that there was no mapped iterdomain from producer.

jacobhinkle: Oh ok so you are mapping the consumer root then you have an rfactor domain for that consumer which uses a Resize?

jacobhinkle: That is right then.

Priya2698:
!build

Priya2698:
!build

PR #2238, opened 2024-05-13
[WIP] IdModel-based indexing

naoyam:
(Rebased #2218)

PR #2364, opened 2024-06-07
Support indexing of DIDx parallelized tensors

naoyam:
Stacked on top of #2353 

Small changes to allow indexing of tensors with DIDx domains.

CC: @zasdfgbnm @cowanmeg @samnordmann @wujingyue


csrc/id_model/indexing.cpp
```diff
@@ -361,7 +361,12 @@ void TensorIndexer::buildLoopIndexMap() {
       ParallelType ptype = getParallelType(loop_group);
       if (isParallelTypeThread(ptype)) {
         loop_index = NamedScalar::getParallelIndex(ptype);
-      } else if (shouldUseZeroIndex(loop_group)) {
+      } else if (
+          // TODO: Cleanup needed. ir_utils::isMemoryPartitionedAcross
+          // should be used, but that means we would need to consider
+          // multiple outputs with different memory types, though it
+          // should be uncommon in practice.
+          shouldUseZeroIndex(loop_group) || isParallelTypeDeviceDim(ptype)) {
```
jacobhinkle, line 369: Could `isParallelTypeDeviceDim(ptype)` go inside `shouldUseZeroIndex`? If any ID in the group is parallelized DID then the loop must be trivial right?

jacobhinkle (approved):
LGTM. Ack about using `isMemoryPartitionedAccess`, but I think we're alread assuming the sibling outputs share the same for-loops (see line 347), so maybe we should assert that all siblings have same memory type and number of leaf IDs.

wujingyue:
Not having to split the logical shape for DID is wonderful. For my education, what are the next steps so we can benefit from this work? I assume this PR fixes IdModel to allow leaf-only DID split, but none/few schedulers use IdModel.

naoyam:
> Not having to split the logical shape for DID is wonderful. For my education, what are the next steps so we can benefit from this work? I assume this PR fixes IdModel to allow leaf-only DID split, but none/few schedulers use IdModel.

I'll soon have a PR to integrate this new indexer into lowering. Something like this:

https://github.com/NVIDIA/Fuser/pull/2238/files#diff-625d71418720e0d8f49be94352457734eea3d6b372a44e53b1afd4484aad3d20R1632-R1643

wujingyue:
> I'll soon have a PR to integrate this new indexer into lowering. Something like this:
> 
> https://github.com/NVIDIA/Fuser/pull/2238/files#diff-625d71418720e0d8f49be94352457734eea3d6b372a44e53b1afd4484aad3d20R1632-R1643

Makes sense for device lowering. My concern was about the schedulers not yet using IdModel. Is IdModel required to allow the schedulers to handle leaf-only DID split?

naoyam:
> > I'll soon have a PR to integrate this new indexer into lowering. Something like this:
> > https://github.com/NVIDIA/Fuser/pull/2238/files#diff-625d71418720e0d8f49be94352457734eea3d6b372a44e53b1afd4484aad3d20R1632-R1643
> 
> Makes sense for device lowering. My concern was about the schedulers not yet using IdModel. Is IdModel required to allow the schedulers to handle leaf-only DID split?

At this moment, no, I don't think so.

PR #2365, opened 2024-06-07
Test error fix

naoyam:
Previously an exception was thrown by ComputeAtMap when the validation of IdModel was done. Now that the validation is disabled by default, the error message changed.

### DIFF
--- a/tests/cpp/test_id_model.cpp
+++ b/tests/cpp/test_id_model.cpp
@@ -42,7 +42,7 @@ TEST_F(IdModelTest, DetectSelfMapping) {
   EXPECT_THAT(
       [&]() { IdModel id_model(&fusion, /*build_graphs=*/true); },
       ::testing::ThrowsMessage<nvfuser::nvfError>(
-          ::testing::HasSubstr("!hasSelfMapping")));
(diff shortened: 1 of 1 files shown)


naoyam:
!build

tests/cpp/test_id_model.cpp
```diff
@@ -42,7 +42,7 @@ TEST_F(IdModelTest, DetectSelfMapping) {
   EXPECT_THAT(
       [&]() { IdModel id_model(&fusion, /*build_graphs=*/true); },
       ::testing::ThrowsMessage<nvfuser::nvfError>(
-          ::testing::HasSubstr("!hasSelfMapping")));
+          ::testing::HasSubstr("are mapped with each other")));
```
naoyam, line 45: This was supposed to detect an error message coming from IdModel but it was actually we were detecting an error message by ComputeAtMap when the validation was done. Now that the validation is disabled, the error message is indeed from IdModel, which has a different error message.

zasdfgbnm (approved):

naoyam:
There's still another failure of `IndexingTest.SimplePointwise1`, which I can't reproduce locally.

PR #2366, opened 2024-06-07
Sharded pointwise scheduler

cowanmeg:
Makes the pointwise scheduler DID compatible by ignorning DID axes from analysis and scheduling operations. It will schedule the compute of sharded pointwise fusion identically to its equivalent single device fusion.

Modifications:
- Remove DID extents from `getPointwiseHeuristics` so that any size based calculations (e.g. break point) only analyze allocated sizes.
- Inside `schedulePointwise`, DID axes are reordered to the front of the leaf domain. In the 2D scheduler merged axis are adjusted to ignore DID axes.

### DIFF
--- a/tests/cpp/test_pointwise.cpp
+++ b/tests/cpp/test_pointwise.cpp
@@ -493,19 +493,17 @@ class ShardedPointwiseTest
 
 TEST_P(ShardedPointwiseTest, DID_Compatible) {
   auto [use_1D_scheduler, sharded_dim] = GetParam();
-  auto fusion_ptr = std::make_unique<Fusion>();
-  auto fusion = fusion_ptr.get();
-  FusionGuard fg(fusion);
+  Fusion fusion;
+  FusionGuard fg(&fusion);
 
   TensorView* tv0 = makeContigTensor(4);
   TensorView* tv1 = makeContigTensor(2);
-
-  fusion->addInput(tv0);
-  fusion->addInput(tv1);
   auto tv2 = add(tv0, tv0);
   auto tv3 = broadcast(tv1, {true, true, false, false});
   auto tv4 = add(tv2, tv3);
-  fusion->addOutput(tv4);
+  fusion.addInput(tv0);
+  fusion.addInput(tv1);
+  fusion.addOutput(tv4);
 
   DeviceMesh mesh = DeviceMesh::createForNumDevices(4);
   for (TensorView* tv : {tv0, tv2, tv3, tv4}) {
@@ -514,11 +512,8 @@ TEST_P(ShardedPointwiseTest, DID_Compatible) {
   }
   tv1->setDeviceMesh(mesh);
 
-  FusionExecutorCache fec(std::move(fusion_ptr));
-  fec.profile(true);
-
   // Trigger the 1D scheduler by using small input size.
-  std::vector<int64_t> input_size(3);
+  std::vector<int64_t> input_size;
   if (use_1D_scheduler) {
     input_size = {16, 8, 24};
   } else {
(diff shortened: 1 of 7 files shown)


tests/cpp/test_pointwise.cpp
```diff
@@ -478,12 +478,106 @@ TEST_F(PointwiseTest, VIssue1567ectorizationFactorAnalysisCase3) {
...
+  EXPECT_EQ(use_1D_scheduler, params->break_point == 0);
+  testValidate(fusion, cg_outputs, aten_inputs, __LINE__, __FILE__);
+
+  // Create a single device fusion and verify the same scheduling parameters are
+  // used for given the same single device fusion.
+  auto unsharded_fusion_ptr = std::make_unique<Fusion>();
+  auto unsharded_fusion = unsharded_fusion_ptr.get();
+  FusionGuard unsharded_fg(unsharded_fusion);
+
+  TensorView* utv0 = makeContigTensor(3);
```
wujingyue, line 548: Any particular reason to use 3? I think [16,1,8,24] as a reference fusion would work too, and it may simplify the test by only requiring to clone the fusion not building a new fusion from scratch.

samnordmann (approved):
Looks great, thanks!

tests/cpp/test_pointwise.cpp
cowanmeg: I used 3 to keep the number of non-device dim axes the same. This is important just for testing when we check the break_points are the same. 
I agree that right now the test is a bit clunky because we are creating two slightly different fusions (also easy to make a simple mistake when editing), but when we allow DID with split, we can use a single fusion with input size [16, 8, 24] and have one with DID scheduling and one without.

tests/cpp/test_pointwise.cpp
```diff
@@ -478,12 +478,106 @@ TEST_F(PointwiseTest, VIssue1567ectorizationFactorAnalysisCase3) {
...
+
+  DeviceMesh mesh = DeviceMesh::createForNumDevices(4);
+  for (TensorView* tv : {tv0, tv2, tv3, tv4}) {
+    tv->setDeviceMesh(mesh);
+    tv->axis(sharded_dim)->parallelize(ParallelType::DIDx);
+  }
+  tv1->setDeviceMesh(mesh);
+
+  FusionExecutorCache fec(std::move(fusion_ptr));
+  fec.profile(true);
```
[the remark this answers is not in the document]
cowanmeg: Tested and it was not necessary, so removed 517-518

tests/cpp/test_pointwise.cpp
```diff
@@ -478,12 +478,106 @@ TEST_F(PointwiseTest, VIssue1567ectorizationFactorAnalysisCase3) {
...
+
+  // Trigger the 1D scheduler by using small input size.
+  std::vector<int64_t> input_size(3);
+  if (use_1D_scheduler) {
+    input_size = {16, 8, 24};
+  } else {
+    input_size = {1024, 32, 64};
+  }
```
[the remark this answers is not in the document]
wujingyue: > I can probably break this up but there are two main things I want to test

Thanks for clarifying your intention and I like how you proposed to separate the concerns. So how about splitting this test into two? 
1. Create a sharded fusion and an unsharded fusion, getPointwiseHeuristics both with various input sizes, assert the returned PointwiseParams are the same. 
2. Create a sharded fusion, manually construct PointwiseParams with various `break_point`s, compileFusion, runFusion, and testValidate. This only needs one input size IIUC, because we will be manually set `break_point`.

tests/cpp/test_pointwise.cpp
wujingyue: If we were to use `[16,1,8,24]`, `params->break_point` and `unsharded_params->break_point` would be off by one? That would make sense to me. 

Nonetheless, you can still put the creation of the fusion into a helper function to reduce duplication. E.g. 

```
auto create_fusion = [](const std::vector<int64_t>& input_shape) -> std::unique_ptr<Fusion> {
  ...
};
```

wujingyue (approved):
LGTM with comments!

cowanmeg:
!build

cowanmeg:
!build

PR #2367, opened 2024-06-07
(int64_t)0 is interpreted as nullptr with gcc

naoyam:
Fixes a failure of `IndexingTest.SimplePointwise1`.

```
IrBuilder::getItemExpr(
              IrBuilder::getAttrExpr(
                  IrBuilder::metadataExpr(tv0), "alloc_stride"),
              (int64_t)0)));
```

This was fine with clang, but with gcc `(int64_t)0` was interpreted as `nullptr`.


naoyam:
!build

zasdfgbnm (approved):

PR #2368, opened 2024-06-07
Use standard {uniform,normal} if params are 0, 1

jacobhinkle:
This just opportunistically uses our unscaled runtime functions instead of the parametrized ones whenever we detect that the input parameters are constant 0 and 1.

I thought nvcc would remove an identity transformation like `(x * 1 + 0)` automatically but the PTX diff shows 8 removed add(0) instructions in `tests/python/test_python_frontend.py::TestNvFuserFrontend::test_nanogpt_mha_dpa`. This is also reflected in the SASS (for sm80).

### DIFF
--- a/csrc/ops/arith.cpp
+++ b/csrc/ops/arith.cpp
@@ -190,16 +190,7 @@ namespace {
 bool valEqualsInt(Val* val, int64_t i) {
   ExpressionEvaluator ee;
   PolymorphicValue pv = ee.evaluate(val);
-  if (!pv.hasValue()) {
-    return false;
-  }
-  if (pv.is<int64_t>()) {
-    return pv.as<int64_t>() == i;
-  }
-  if (pv.is<double>()) {
-    return pv.as<double>() == (double)i;
-  }
-  return false;
+  return pv.hasValue() && (pv == i).as<bool>();
 }
 
 } // namespace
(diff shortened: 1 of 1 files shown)


jacobhinkle:
!build --diff

jacobhinkle:
Thunder failures are unrelated. The only codediffs are as expected, in `TensorFactoryTest.FactoryBroadcast` and `tests/python/test_python_frontend.py::TestNvFuserFrontend::test_nanogpt_mha_dpa`.

csrc/ops/arith.cpp
```diff
@@ -182,6 +182,28 @@ TensorView* rand(
...
+  if (!pv.hasValue()) {
+    return false;
+  }
+  if (pv.is<int64_t>()) {
+    return pv.as<int64_t>() == i;
+  }
+  if (pv.is<double>()) {
+    return pv.as<double>() == (double)i;
+  }
+  return false;
```
zasdfgbnm, line 202:
```suggestion
  return pv.hasValue() && (pv == i).as<bool>();
```

zasdfgbnm (approved):

csrc/ops/arith.cpp
jacobhinkle: That's awesome. Thanks.

PR #2373, opened 2024-06-09
preseg optimization pass - optimization for `cat`

jjsjann123:
Adding pre-segmentation optimization pass that optimize `cat` (`PadOp` + `CatOp`) to avoid multiple kernels.

1. The mutation tries to push `PadOp` further to its producer as described below. The goal is to either have the pad op directly on inputs to fusion, or to a point where segmentation at pad leaves a no-op fusion segment before pad.
```
// The concept is that the following program:
//   tv1 = UnaryOp(tv0)
//   tv2 = PadOp(tv1)
// can be replayed as the program below:
//   tv0_padded = PadOp(tv0)
//   tv2_new = UnaryOp(tv0_padded)
```

For details on the propagation logic, please see code comment
Note [ PadOp Propagation Rule ]
Note [ Handling TV with Multiple Uses via Frontier ]

2. As a result of moving the `PadOp`, we also need to replace the `CatOp` with a series of binary add, since its inputs are no longer directly produced by `PadOp`.


jjsjann123:
!build

jjsjann123:
!build

csrc/preseg_passes/move_pad.cpp
```diff
@@ -0,0 +1,233 @@
...
+  //   4. maybe also check aliases?!
+  while (!stack.empty()) {
+    Edge edge = stack.top();
+    Expr* def = edge.val()->definition();
+    stack.pop();
+
+    if (def->isA<UnaryOp>()) {
+      auto* uop = def->as<UnaryOp>();
+      // TODO: exception to break propagation.
+      if (candidate_check(uop->in())) {
```
jjsjann123, line 94: need to exclude some operation types here to avoid illegal instructions like division by 0.

csrc/preseg_passes/move_pad.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+    // TODO: modify existing pad_op, when its only consumer is a pad_op
+  }
+
+  // propagate to update TensorProxy
+  // need to follow the reverse order from earlier stack traversal.
+  std::reverse(replay_sequence.begin(), replay_sequence.end());
+  for (Expr* e : replay_sequence) {
+    // TODO extend this for multiple inputs.
+    Expr* padded_e = replayExprWithNewInput(e, replacement_map.at(e->input(0)));
```
jjsjann123, line 170: note for myself. I don't think we need this here but rather a newValLike would suffice.

jjsjann123:
Oops, looks like there's real issue here. Will check back later.

jjsjann123:
!build

jjsjann123:
!build

jjsjann123:
:crossed_fingers:

csrc/preseg_passes/move_pad.cpp
```diff
@@ -0,0 +1,457 @@
...
+    Val* left_pad = merged_pad_widths.at(i * 2);
+    Val* right_pad = merged_pad_widths.at(i * 2 + 1);
+    IterDomain* inp_id = inp_dom.at(i);
+    if (left_pad->isZeroInt() && right_pad->isZeroInt()) {
+      merged_root_ids.push_back(inp_id->cloneWithoutRFactor());
+      merged_logical_ids.push_back(merged_root_ids.back());
+      continue;
+    }
+    // NOTE: nvfuser pad doesn't support negative padding, so we don't have to
+    // worry about it cancelling out.
```
jacobhinkle, line 193: We do support pad by negative amounts, but it just won't happen in the `cat` scenario. See this test: https://github.com/NVIDIA/Fuser/blob/1bfb11a34d31200d7c2bc8a4e008a4279baf160b/tests/cpp/test_resize.cpp#L1666

csrc/preseg_passes/move_pad.cpp
```diff
@@ -0,0 +1,457 @@
...
+  }
+  auto pad_replay_check = [&pad_dependencies](Val* val) {
+    if (!val->isA<TensorView>()) {
+      return false;
+    }
+    if (std::any_of(
+            pad_dependencies.begin(),
+            pad_dependencies.end(),
+            [val](Val* pad_dependency) {
+              return DependencyCheck::isDependencyOf(pad_dependency, val);
```
jacobhinkle, line 251: You could also use `DependencyCheck::getAllDependentVals` for `pad_dependencies` (have to make it an `unordered_set`). That way you can just check inclusion in the resulting set for each `val`.

csrc/preseg_passes/move_pad.cpp
```diff
@@ -0,0 +1,457 @@
...
+  Val* val() const {
+    return expr_->input(index_);
+  }
+};
+
+// The pass assumes propagating PadOp with zero pad. The criteria here for
+// return true is that `unaryOp(0) == 0`
+bool padCompatibleUnaryOp(UnaryOpType t) {
+  switch (t) {
+    case UnaryOpType::Cast:
```
jacobhinkle, line 39: Do we handle this using `SimplifyingIrBuilder::maybeCastExpr(pad_value)` somewhere?

jacobhinkle (commented):
Just a first pass. Haven't read it all yet.

csrc/preseg_passes/move_pad.cpp
jjsjann123: ha, I'll update and add a check for that.
Along with the comment here: https://github.com/NVIDIA/Fuser/blob/1416d8938fc3d3bd019c44cabb45d32d1ef944c3/csrc/ops/alias.cpp#L469-L470

csrc/preseg_passes/move_pad.cpp
jjsjann123: Would you mind elaborate on this one?

For the context, I'm not trying to simplify existing `cast` in the IR. But merely move the `PadOp` from the cast output to its input.

csrc/preseg_passes/move_pad.cpp
jjsjann123: Yeah this doesn't look right when we try to merge multiple `PadOp` together. I'll address it in the follow up PR #2490  where we can actually put in proper tests.

csrc/preseg_passes/pre_segmenter.cpp
```diff
@@ -36,6 +37,14 @@ namespace nvfuser::preseg_passes {
   OptimizationPass<ConsecutiveCastPass>::runPass(fusion);
   OptimizationPass<AddAxiomsPass>::runPass(fusion);
   OptimizationPass<MoveSplitCatPass>::runPass(fusion);
+  // MovePadPass needs to happen before MarkAliasPrepare and after MoveSplitCat
```
wujingyue, line 40: Can you explain why? I think I know the reason but it's good to put things in the code comment for posterity.

wujingyue (commented):
Nice work! I'll be busy with #2395 for the rest of the week. Given the size of the PR, I'll probably have to review this next week. Feel free to merge with others' approval.

csrc/preseg_passes/pre_segmenter.cpp
jjsjann123: > after MoveSplitCat

avoid this pass moving `PadOp` around to break the `MoveSplitCat`.

> before MarkAliasPrepare

This is not necessary for the implementation as-is. i.e. the pass currently won't change anything in the graph that would impact `MarkAliasPrepare`. So having it running before or after `MarkAliasPrepare` does not matter for now.
But if we were ever to add propagation for things like broadcast/expand/squeeze, we would want `MarkAliasPrepare` to also run on the mutated graph, hence it ideally should happen before.

I will add these in the code comment

jjsjann123:
> Given the size of the PR,

Sorry about that one. I tried to break out the `binary` op into a separate PR (along with the test), which could get rid of about ~250 LOC. :bow: 

But I genuine think that would make the review of the initial PR even harder, since I cannot put up example with the proper handling of mutiple uses in examples `PadReplayOnMultipleUsesCase*`.

I'm still open to that idea if we think it's beneficial to keep smaller PRs and reviewers don't mind leaving some code logic to be justified in separate PRs.

naoyam:
What does the performance look like with this optimization?

csrc/preseg_passes/move_pad.cpp
jjsjann123: Good point.

I didn't know that `pad_value` is sensitive with its DataType. I'll look into that.

csrc/preseg_passes/move_pad.cpp
```diff
@@ -0,0 +1,457 @@
...
+
+// replayConcretePad tries to replay a concretized PadOp on pad_tv. This
+// function allows merging multiple consecutive PadOp by stacking their pad
+// widths together as `vec_pad_widths`. This function assumes it has already
+// resolved the output iter_type for each IterDomain and provided as
+// `ref_iter_type`. The function returns the output from the replay PadOp.
+//
+// NOTE: this assumes all vec_pad_widths are positive entries so we don't need
+// to consider accumulating them changing the output iter_type.
+TensorView* replayConcretePad(
```
wujingyue, line 140: Can this be merged into https://github.com/NVIDIA/Fuser/blob/1bd98c19390c1bd3c2e54954b85315f905c0db2f/csrc/transform_replay.h#L296?

csrc/preseg_passes/move_pad.cpp
```diff
@@ -0,0 +1,457 @@
...
+  // reach a Val during the propagation, we add Val entry *temporarily* to the
+  // frontier map and increment its encounter count. We cannot propagate the
+  // PadOp over the producer of Val yet, because it still have other uses that
+  // expects the original Val. Until we see all uses of Val during the
+  // traversal, we know all uses can be replaced with the padded Val and hence
+  // it's safe to propagate `PadOp` across. We remove Val fron `frontier` map
+  // since we can push the replay to its producer. We return `true` to signal
+  // that further propagation is allowed.
+  if (val->isFusionOutput()) {
+    // skipping propagation when target is an output.
```
wujingyue, line 107: Why?

csrc/preseg_passes/move_pad.cpp
```diff
@@ -0,0 +1,457 @@
...
+
+// This pass tries to push `PadOp`, which is part of `cat` (composed of `PadOp`s
+// + `CatOp`), further to its producers to avoid possible segmentation. As a
+// side effect, it replaces the `CatOp` with a series of pointwise add if any
+// mutation has been made for propagating `PadOp`
+void decomposeCatOp(Fusion* fusion) {
+  // TODO: verify that no dead branch is traversed in exprs.
+  std::vector<Expr*> exprs = fusion->exprs();
+
+  for (auto* cat : ir_utils::filterByType<CatOp>(exprs)) {
```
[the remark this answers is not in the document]
wujingyue: To be constructive, I think there's an alternative to initially add all pads from all cats in the work list (whether a stack or a queue) and then push them upwards and combine if necessary. This can be more effective than the approach taken in this PR, and avoid the need for a separate #2490 to merge pads originated from different cats. Wdyt about that?

wujingyue (commented):
> 2\. we also need to replace the `CatOp` with a series of binary add

Are these binary adds show up in the CUDA code? Or is nvFuser smart enough to figure out all operands but one are zero and therefore skip the adds? I'm asking because there may be a penalty if some pads are moved but not far enough (e.g. to fusion inputs) to offset the cost of extra additions.

jjsjann123:
> What does the performance look like with this optimization?

That's a good question.
This PR helps us avoid **some** segmentation. In terms of kernel performance, I'm not seeing much of a difference on a simple example below when the number of concat tensor is small (just three tensor view).

```
import torch

    T0 = fd.define_tensor(shape=[-1, -1, -1], contiguity=[True, False, True], dtype=DataType.Half, is_cpu=False, stride_order=[2, 1, 0])
    T1 = fd.define_tensor(shape=[-1, -1, -1], contiguity=[True, False, True], dtype=DataType.Half, is_cpu=False, stride_order=[2, 1, 0])
    T6 = fd.define_tensor(shape=[-1, -1, -1], contiguity=[True, False, True], dtype=DataType.Half, is_cpu=False, stride_order=[2, 1, 0])
    T2 = fd.ops.cast(T0, dtype=DataType.Float)
    T3 = fd.ops.neg(T2)
    T4 = fd.ops.cast(T3, dtype=DataType.Half)
    T5 = fd.ops.cat([T4, T1, T6], dim=-1)
    fd.add_output(T5)

    nvfuser_fusion_id0(fd)

BATCH_SIZE=10

inputs = [
    torch.randn((BATCH_SIZE*128*20,), dtype=torch.float16, device='cuda:0').as_strided((BATCH_SIZE, 128, 20), (20*128, 20, 1)),
    torch.randn((BATCH_SIZE*128*1024,), dtype=torch.float16, device='cuda:0').as_strided((BATCH_SIZE, 128, 1024), (1024*128, 1024, 1)),
    torch.randn((BATCH_SIZE*128*1024,), dtype=torch.float16, device='cuda:0').as_strided((BATCH_SIZE, 128, 1024), (1024*128, 1024, 1)),
]

for i in range(5):
  fd.execute(inputs)
```

This is measured before the changes on `main`
```
 ** CUDA GPU Kernel Summary (cuda_gpu_kern_sum):

 Time (%)  Total Time (ns)  Instances  Avg (ns)  Med (ns)  Min (ns)  Max (ns)  StdDev (ns)                                                  Name
     65.4            68001          5   13600.2   13600.0     13504     13728         87.7  <unnamed>::nvfuser_pointwise_f0_c1_r0_g0(<unnamed>::Tensor<<unnamed>::__half, (int)3, (int)3>, <unn…
     10.3            10753          5    2150.6    2144.0      2112      2241         53.0  <unnamed>::nvfuser_pointwise_f0_c1_r0_g1(<unnamed>::Tensor<<unnamed>::__half, (int)3, (int)3>, <unn…
```

With this PR I'm seeing a single kernel and about the same time.
```
 ** CUDA GPU Kernel Summary (cuda_gpu_kern_sum):

 Time (%)  Total Time (ns)  Instances  Avg (ns)  Med (ns)  Min (ns)  Max (ns)  StdDev (ns)                                                  Name
     72.6            66689          5   13337.8   13345.0     13184     13408         91.7  <unnamed>::nvfuser_pointwise_f0_c1_r0_g0(<unnamed>::Tensor<<unnamed>::__half, (int)3, (int)3>, <unn…
```

But if I bump the number of concat tensor up, I'm starting to see performance regression from this change. So this sounds like we should have a heuristic on when the optimization should be applied.

jjsjann123:
> > 2. we also need to replace the `CatOp` with a series of binary add
> 
> Are these binary adds show up in the CUDA code? Or is nvFuser smart enough to figure out all operands but one are zero and therefore skip the adds? I'm asking because there may be a penalty if some pads are moved but not far enough (e.g. to fusion inputs) to offset the cost of extra additions.

Yes they do show up and indeed would impact performance if the number of operands in `cat` is large. See my response to @naoyam's earlier question. I'm not sure if tensor sizes would also impact this, or if it's a purely # of operand thing yet.

Let me run a separate measurement on that and come back with some concrete numbers.

csrc/preseg_passes/move_pad.cpp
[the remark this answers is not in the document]
jjsjann123: I agree with you that it's more efficient a pass when we merge `pads` together before we propagate. Especially in the example you posted here.

```
z0 = cat([x, y0])
z1 = cat([x, y1])  // assume shape(y1)==shape(y0)
```

With our existing logic, we can't handle the horizontal cat in current implementation yet. But we will be able to do so if we try to merge `pad` before we push it out, assuming that we can prove they have identical `pad_widths` / `pad_value`.

I think if I swap the logic here to iterate from later nodes to earlier nodes and add a `merge pads` as part of the traversal logic, I can merge the second PR here as well and I can also ignore the tricky cases with a wild `padOp` could have negative padding.

Sounds like a good idea to me.

csrc/preseg_passes/move_pad.cpp
jjsjann123: I'm likely not following the idea here.

I can try to squeeze it inside that as an overload?! I'm not sure this function I have here is a general `replay expression` and feels that it's very specific just how we construct `pad`. Would you mind elaborating on your suggestion? :bow:

wujingyue: Let me rephrase: replayExprWithNewInput was built to replay a unary op on a new input. It does so by replaying the IterDomain transforms on the new root domain. What's special about PadOp that would break this utility?

jjsjann123: > replay a unary op on a new input

I think it's the **replay** in this function name that's confusing :)

The difference here is that `PadOp` is not a unary op. i.e. in this utility function we do expect new `pad_value` as well as stacked pad widths. So we cannot just replay the old transformation on the exprssion.

csrc/preseg_passes/move_pad.cpp
```diff
@@ -0,0 +1,457 @@
...
+  }
+
+  // frontier contains `Val`s that needs to replay pad_op on. The map here
+  // stores the number of encounter for each `Val*` during traversal, the logic
+  // is used to decide whether propagation can cross Val with multiple uses in
+  // the graph. See NOTE [ Handling TV with Multiple Uses via Frontier ]
+  std::unordered_map<Val*, int64_t> frontier;
+  // replay_sequence is used later to create the updated branch with padded
+  // inputs after all `frontier` has been updated with padding.
+  std::vector<Expr*> replay_sequence;
```
wujingyue, line 277: I suspect it's unnecessary to record a **sequence** and replay things in the sequence all at once. Instead, we may be able to simply move `pad` up incrementally, one "swap" (of `pad` and its preceding op) at a time.

csrc/preseg_passes/move_pad.cpp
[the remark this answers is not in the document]
jjsjann123: Realized that this pattern is actually real in qkv_split.

i.e. cos/sin is used by cat on both q and k. i.e. if we need to propagate `pad` back, we need to handle this. Thanks for pointing this one out @wujingyue

csrc/preseg_passes/move_pad.cpp
jjsjann123: if val is a fusion output, I decided to stop the propagation here. Because after the propagation, we'll have `padded_val` instead, which means we have to preserve the old branch to have correct output `val`.

Note, this logic is moved to a new place in the refactor. I'll add a comment there.

jjsjann123:
!build

csrc/preseg_passes/move_pad.cpp
```diff
@@ -0,0 +1,530 @@
...
+    }
+  }
+
+  return true;
+}
+
+// This operation replaces:
+//   CatOp(inputs)
+// with:
+//   BinaryOp(inputs[n-1], BinaryOp(inputs[n-2], BinaryOp(..., BinaryOp[0])...))
```
jacobhinkle, line 88: I think it's worth mentioning in this comment that the binary op applied depends on the DataType and is `add` for numeric types and `logical_or` for boolean.

csrc/preseg_passes/move_pad.cpp
```diff
@@ -0,0 +1,530 @@
...
+Val* replayCatOpWithBinaryOp(
+    const std::vector<Val*>& inputs,
+    DataType data_type) {
```
jacobhinkle, line 91: Could `data_type` be inferred as `inputs.front()->dtype()` instead?

csrc/preseg_passes/move_pad.cpp
```diff
@@ -0,0 +1,530 @@
...
+    case UnaryOpType::Neg:
+    case UnaryOpType::Real:
+    case UnaryOpType::Relu:
+    case UnaryOpType::Round:
+    case UnaryOpType::Sin:
+    case UnaryOpType::Sinh:
+    case UnaryOpType::Sqrt:
+    case UnaryOpType::Tan:
+    case UnaryOpType::Tanh:
+    case UnaryOpType::Trunc:
```
jacobhinkle, line 137: Add `Expm1`, `Log1p`, `Frac`, `Signbit`, and reorder `Silu` to preserve alphabetic order.
```suggestion
    case UnaryOpType::Abs:
    case UnaryOpType::Asin:
    case UnaryOpType::Asinh:
    case UnaryOpType::Atan:
    case UnaryOpType::Atanh:
    case UnaryOpType::Ceil:
    case UnaryOpType::Erf:
    case UnaryOpType::Erfinv:
    case UnaryOpType::Expm1:
    case UnaryOpType::Floor:
    case UnaryOpType::Frac:
    case UnaryOpType::Gelu:
    case UnaryOpType::Imag:
    case UnaryOpType::Log1p:
    case UnaryOpType::Neg:
    case UnaryOpType::Real:
    case UnaryOpType::Relu:
    case UnaryOpType::Round:
    case UnaryOpType::Silu:
    case UnaryOpType::Signbit:
    case UnaryOpType::Sin:
    case UnaryOpType::Sinh:
    case UnaryOpType::Sqrt:
    case UnaryOpType::Tan:
    case UnaryOpType::Tanh:
    case UnaryOpType::Trunc:
```
Not sure what you'd like to do with `IsInf`, `IsNegInf`, `IsPosInf`, and `IsNan`. These take numeric 0 and give bool `false` so I believe it's safe to include them here.

csrc/preseg_passes/move_pad.cpp
```diff
@@ -0,0 +1,530 @@
...
+  }
+}
+
+// The pass assumes propagating PadOp with zero pad. The criteria here for
+// return true is that `binaryOp(0, x) == 0` & `binaryOp(x, 0) == 0`
+bool zeroIsIdentity(BinaryOpType t) {
+  switch (t) {
+    case BinaryOpType::Add:
+    case BinaryOpType::Mul:
+    case BinaryOpType::Sub:
```
jacobhinkle, line 150: This does not hold for `Sub`: `0 - x != x`.

csrc/preseg_passes/move_pad.cpp
```diff
@@ -0,0 +1,530 @@
...
+    case BinaryOpType::BitwiseAnd:
+    case BinaryOpType::LogicalAnd:
```
jacobhinkle, line 152:
```suggestion
    case BinaryOpType::BitwiseOr:
    case BinaryOpType::BitwiseXor:
    case BinaryOpType::LogicalOr:
```

jacobhinkle (commented):
First just a few comments on the unary and binary ops classifications.

tests/cpp/test_move_pad.cpp
```diff
@@ -0,0 +1,437 @@
...
+  auto fusion = std::make_unique<Fusion>();
+  FusionGuard fg(fusion.get());
+
+  TensorView* tv0 = makeContigConcreteTensor({4, 10});
+  TensorView* tv1 = makeContigConcreteTensor({1, 10});
+  // pad on tv5 will propagation back to tv0, since all its uses are traversed
+  //
+  //   tv0 --> tv2 --> tv3 --> tv5
+  //               \        /
+  //                -> tv4 -
```
jacobhinkle, line 169: I'm trying to trace what happens in this example. We start with `pad(tv5)` then we convert that to
```
add(pad(tv3), pad(tv4));
```
Then I think the stack-based approach we use will process `pad(tv4)`, replacing it with `sin(pad(tv2))`, then replacing `pad(tv2)` with `relu(pad(tv0))`. At this point we can't continue but there is another input PadOp on the frontier and that is `pad(neg(pad(tv0)))` (since we have already replaced `tv2`). After one more move the two pads bump into one another; as I understand it this means we "stack" the pads. How are we avoiding doubling the pad widths in this case?

csrc/preseg_passes/move_pad.cpp
```diff
@@ -0,0 +1,530 @@
...
+  // replay `CatOp` with series of BinaryOp instead, since we might have
+  // pushed `PadOp` out and breaking the codegen if `CatOp` remains.
+  Val* res = nullptr;
+  bool is_boolean = isBooleanType(data_type);
+  for (Val* inp : inputs) {
+    if (res == nullptr) {
+      res = inp;
+    } else {
+      if (is_boolean) {
+        res = bitwise_or(res, inp);
```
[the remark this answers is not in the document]
jjsjann123: good catch!

csrc/preseg_passes/move_pad.cpp
jjsjann123: ohohoh good catch again. I need better testing for these....

tests/cpp/test_move_pad.cpp
jjsjann123: Sorry the propagation logic above has a subtle thing I should have highlighted.

we are not supposed to have `pad(neg(pad(tv0)))`.
when we have `add(neg(pad(tv2)), sin(pad(tv2)))`. The next step would try to unify the pad operations on tv2. Look at the use of `isSamePadOp` above. and we'll merge the fork there. Afterwards, the graph becomes:

```
tv0 -> tv2 - (merged_pad) -> tv_new (forked with neg & sin)
```

then we'll propagate the merge pad up to tv0.

We avoid repetitive application of pad on a single tensor with fork by checking the uses of pad->input to be 1 before propagating any pad across (this applies to unary/binary/catOp).

jacobhinkle: How is it that we would get `sin(tv4)` before proceeding though? What stops the frontier propagating past `neg(pad(tv2))`?

jjsjann123: when we walk up the `neg` branch and reached `neg(pad(tv2))`. If the other `sin` branch is not ready yet (i.e. the pad is still down somewhere like `pad(sin(tv2))`.

At line 357, we'll check the `tv2->uses().size()` and if tv2 has more uses than 1, the propagation will abort on `neg` line.
We would only propagate pad after `tv2`, when it walks up the second time from the `sin` branch. At which point, all uses are consumed by the identical `PadOp` and merged as a single use.

jjsjann123: I realized I should mention this in the code comment. I'll put it in `Note [ PadOp Propagation Rule ]`

jjsjann123: Also, please do drop more questions on this if it doesn't sound clear. That'll help me put up a better comment. :bow:

jacobhinkle: Ah that's it! Lines 357 and 387 are the key there. Sounds good. Thanks for the explanation!

csrc/preseg_passes/move_pad.cpp
```diff
@@ -0,0 +1,530 @@
...
+    return true;
+  }
+  for (Val* pad_val : pad->getPadWidths()) {
+    if (!simplifyExpr(
+             SimplifyingIrBuilder::geExpr(pad_val, pad->fusion()->zeroVal()))
+             ->isTrue()) {
+      return false;
+    }
+  }
+  return true;
```
wujingyue, line 45: Nit: there's a thing called `std::all_of` that may simplify the code here.

csrc/preseg_passes/move_pad.cpp
```diff
@@ -0,0 +1,530 @@
...
+}
+
+// check if `use` and `p` are the same pad operatoin.
+bool isSamePadOp(Expr* use, PadOp* p) {
```
wujingyue, line 49: Did you try https://github.com/NVIDIA/Fuser/blob/9f8c99148c2299ff3a22f2e8f2d9a86bc4063355/csrc/ir/base_nodes.cpp#L344?

csrc/preseg_passes/move_pad.cpp
```diff
@@ -0,0 +1,530 @@
...
+  }
+
+  return true;
+}
+
+// This operation replaces:
+//   CatOp(inputs)
+// with:
+//   BinaryOp(inputs[n-1], BinaryOp(inputs[n-2], BinaryOp(..., BinaryOp[0])...))
+Val* replayCatOpWithBinaryOp(
```
wujingyue, line 89:
```suggestion
Val* replaceCatOpWithBinaryOp(
```
as you said in the comment?

csrc/preseg_passes/move_pad.cpp
```diff
@@ -0,0 +1,530 @@
...
+    const std::vector<Val*>& inputs,
+    DataType data_type) {
+  // replay `CatOp` with series of BinaryOp instead, since we might have
+  // pushed `PadOp` out and breaking the codegen if `CatOp` remains.
+  Val* res = nullptr;
+  bool is_boolean = isBooleanType(data_type);
+  for (Val* inp : inputs) {
```
wujingyue, line 96: Nit: assuming `cat` has at least one input, you can simplify the code to something like

```
auto binary_op = is_boolean ? logical_or : add;
Val* res = inputs[0];
for (auto i : c10::irange(1, inputs.size())) {
  res = binary_op(res);
}
return res;
```

csrc/preseg_passes/move_pad.cpp
```diff
@@ -0,0 +1,530 @@
...
+// Given certain constraints on the UnaryOp, we can ensure that the
+// computational output remains the same when we replace all uses of `tv2` with
+// `tv2_new`.
+//
+// This function only propagates simple padding, where its pad value is `zero`
+// (or `false` for boolean) and pad widths are non-negative. This allows us to
+// unconditionally merge neighboring PadOps as a single op. We also restrict
+// propagation on operatoins that can allow PadOp to propagated across. See
+// `zeroIsFixedPoint` and `zeroIsIdentity`.
+void propagatePad(Fusion* fusion) {
```
wujingyue, line 275:
```suggestion
void propagatePads(Fusion* fusion) {
```

csrc/preseg_passes/move_pad.cpp
```diff
@@ -0,0 +1,530 @@
...
+  // propagating PadOp
+  auto exprs = fusion->exprs();
+  auto filtered_pads = ir_utils::filterByType<PadOp>(exprs);
```
wujingyue, line 278:
```suggestion
  auto filtered_pads = ir_utils::filterByType<PadOp>(fusion->exprs());
```

csrc/preseg_passes/move_pad.cpp
```diff
@@ -0,0 +1,530 @@
...
+  std::vector<PadOp*> frontier;
```
wujingyue, line 279:
```suggestion
  std::vector<PadOp*> worklist;
```
Nit: move_split_cat processes one `cat` at a time, and I used `frontier` to mean a line of ops that are control-equivalent to the `cat` op being processed. Here, the pads in the list might not form a "frontier" per say.

csrc/preseg_passes/move_pad.cpp
```diff
@@ -0,0 +1,530 @@
...
+    PadOp* p = frontier.back();
+    frontier.pop_back();
+
+    // if no uses, this has already been short-wired.
+    if (p->out()->uses().empty() && !p->out()->isFusionOutput()) {
+      continue;
+    }
+
+    // unify all consumer pad of tv;
+    TensorView* tv = p->in()->as<TensorView>();
```
wujingyue, line 305:
```suggestion
    auto* tv = p->in()->as<TensorView>();
```

csrc/preseg_passes/move_pad.cpp
```diff
@@ -0,0 +1,530 @@
...
+        continue;
+      }
+      pad_inputs.insert(val);
+    }
+    std::unordered_set<Val*> pad_dependencies =
+        DependencyCheck::getAllDependentVals(pad_inputs);
+    auto pad_replay_check = [&pad_dependencies](Expr* expr) {
+      return std::all_of(
+          expr->inputs().begin(),
+          expr->inputs().end(),
```
wujingyue, line 338: I don't follow this part. IIUC, `pad_replay_check` tries to detect whether swapping `p` and the preceding `expr` will create cycles. However, having both `expr` and `p` use the same `Val*` doesn't necessarily create cycles. `expr->outputs` would be more reasonable to me.

wujingyue (commented):
LGTM overall. I'll review the tests tomorrow.

csrc/preseg_passes/move_pad.cpp
```diff
@@ -0,0 +1,530 @@
...
+          {p->getPadWidths()},
+          TensorDomain::noReductions(
+              p->out()->as<TensorView>()->getLogicalDomain()));
+
+      new_out = ops::newValLike(new_pad_out, uop->out()->getDataType().value());
+      IrBuilder::create<UnaryOp>(uop->getUnaryOpType(), new_out, new_pad_out);
+      // insert new PadOp(s) to frontier;
+      frontier.push_back(new_pad_out->definition()->as<PadOp>());
+      simple_pad_set.insert(new_pad_out->definition()->as<PadOp>());
+    } else if (auto* bop = dynamic_cast<BinaryOp*>(def)) {
```
wujingyue, line 384: I'm seeing a repeated pattern in handling unary, binary, and cat. Have you thought about combining them?

tests/cpp/test_move_pad.cpp
```diff
@@ -0,0 +1,437 @@
...
+  EXPECT_EQ(runtime->fusionSegments()->groups().size(), 2);
+
+  // ensure that we propagate the pad across binary operation and the first
+  // segment is no-op
+  auto scheduler = runtime->schedulerHeuristics()->heuristicsList().at(0).get();
+  auto heuristic = scheduler->heuristic();
+  NVF_CHECK(
+      heuristic == ScheduleHeuristic::NoOp,
+      "Unexpected heuristic: ",
+      heuristic);
```
wujingyue, line 121: We can probably simplify this using UnorderedElementsAre and HeuristicIs, e.g., https://github.com/NVIDIA/Fuser/blob/4314090a6ab68b955220fcced675b5e05aa9cf0e/tests/cpp/test_alias.cpp#L1111-L1115

tests/cpp/test_move_pad.cpp
```diff
@@ -0,0 +1,437 @@
...
+  FusionGuard fg(fusion.get());
+
+  // all pad operations should be merged together
+  TensorView* tv0 = makeContigConcreteTensor({4, 10});
+  TensorView* tv1 =
+      pad(tv0,
+          {IrBuilder::create<Val>(0L),
+           IrBuilder::create<Val>(2L),
+           IrBuilder::create<Val>(1L),
+           IrBuilder::create<Val>(1L)});
```
wujingyue, line 238: (Can come as a separate PR)

It feels worth adding to csrc/ops/alias.h a helper that takes integer pad widths.

tests/cpp/test_move_pad.cpp
```diff
@@ -0,0 +1,437 @@
...
+  at::Tensor t2 = at::randn({2, 10}, options) > 0.5;
+  std::vector<c10::IValue> aten_inputs = {t0, t1, t2};
+
+  FusionExecutorCache fec(std::move(fusion));
+  auto out_tensors = fec.runFusionWithInputs(aten_inputs);
+
+  FusionKernelRuntime* runtime = fec.getMostRecentKernelRuntime();
+  EXPECT_EQ(runtime->fusionSegments()->groups().size(), 1);
+
+  // ExpressionEvaluator is hitting an assert with dynamic value.
```
wujingyue, line 430: File a bug and link it here?

csrc/preseg_passes/move_pad.cpp
```diff
@@ -0,0 +1,530 @@
...
+      frontier.push_back(vals[0]->definition()->as<PadOp>());
+      simple_pad_set.insert(vals[0]->definition()->as<PadOp>());
+      frontier.push_back(vals[1]->definition()->as<PadOp>());
+      simple_pad_set.insert(vals[1]->definition()->as<PadOp>());
+    } else if (auto* pop = dynamic_cast<PadOp*>(def)) {
+      // stop propagation if PadOp `pop` isn't a simple PadOp, since we can only
+      // merge simple PadOp together. Note that we don't need to check the other
+      // uses of `tv` here, since we want to merge the consecutive pads anyway
+      // and it won't interfere the other uses of `tv`.
+      if (simple_pad_set.count(pop) == 0) {
```
wujingyue, line 447: It feels like we've maintained simple_pad_set just for the pad-pad corner case. Unless proven to be expensive, I'd call isSimplePadOp on demand, which is simpler.

wujingyue (approved):

tests/cpp/test_move_pad.cpp
jjsjann123: Priya was kind enough to support it later. I'll open an issue for her to look at.
:bow:

csrc/preseg_passes/move_pad.cpp
jacobhinkle: I think the issue is that while `isSimplePadOp` works as expected on the original pad ops it's not clear that it will still return true for the replacements we make along the way; mainly because we currently pattern match the uses for a `CatOp` which will fail once we shuffle the pad in the producer direction. So as the frontier evolves we need to track that the pads we encounter are either original simple pads or replacements from simple pads. This could use a comment in the code I think.

csrc/preseg_passes/move_pad.cpp
jjsjann123: I don't think that one would work, since it doesn't try to simplify pad widths, but rather rely on things coming from the same definition https://github.com/NVIDIA/Fuser/blob/3233bcef198f2693634a3e1b49a9836bf4d28694/csrc/ir/base_nodes.cpp#L147-L149

csrc/preseg_passes/move_pad.cpp
jjsjann123: I don't think this works, since filterByType returns a view and `fusion->exprs()` returns a new vector.
https://github.com/NVIDIA/Fuser/blob/3233bcef198f2693634a3e1b49a9836bf4d28694/csrc/ir/utils.h#L200-L201

csrc/preseg_passes/move_pad.cpp
jjsjann123: :bow: 

good catch! I messed it up in the refactor. Thanks a lot!
I'll make sure that I create a test for this one.

csrc/preseg_passes/move_pad.cpp
jjsjann123: Yeah Jacob is right. This is the same reason I have this hacky thing inside: https://github.com/NVIDIA/Fuser/blob/7557dfda306899b96e90bc98353cf47de7889619/csrc/preseg_passes/move_pad.cpp#L32-L37

I'll put a comment down where `simple_pad` is as well

csrc/preseg_passes/move_pad.cpp
jjsjann123: sorry I just realized the criteria here wasn't right....

Would you mind having another look at this list? @jacobhinkle

jjsjann123:
!build

csrc/preseg_passes/move_pad.cpp
jacobhinkle: Ah makes sense to update the condition since we pad both inputs. List looks good but I think NE could be added.

csrc/preseg_passes/move_pad.cpp
```diff
@@ -0,0 +1,539 @@
...
+    case BinaryOpType::Max:
+    case BinaryOpType::Min:
+    case BinaryOpType::Mul:
+    case BinaryOpType::Sub:
+    case BinaryOpType::BitwiseAnd:
+    case BinaryOpType::BitwiseOr:
+    case BinaryOpType::GT:
+    case BinaryOpType::LT:
+    case BinaryOpType::LogicalAnd:
+    case BinaryOpType::LogicalOr:
```
jacobhinkle, line 174: Maybe BinaryOpType::Gcd could be added?

jacobhinkle (approved):
LGTM

jjsjann123:
!build

jjsjann123:
thunder error seems to be in main as well. Looks like it's not related to my PR.

jjsjann123:
Tried qkv_split benchmark with this and performance is really bad. :sob: 

Comparing nvfuser+thunder with and without this PR.
```
   26  python benchmark_thunder.py c1b211cd6c49ea6b8354ebaf64145a3461accbce:c1f672e83934ea9d0e9b4645b62b577aa8021fff --filter="thunder]"
   27  python benchmark_thunder.py c1b211cd6c49ea6b8354ebaf64145a3461accbce:9ec11c0f0085bba40567c352bdb077743e9c7881 --filter="thunder]"
```

https://gist.github.com/jjsjann123/f010e80aaf5f033e1ddf604a2e173de4

changes in this PR does help reducing total kernel numbers, but e2e performance regression comes from the regressed kernel performance with pushed out pad+binary_add.

csrc/preseg_passes/move_pad.cpp
jacobhinkle: This seems useful so in another PR it could probably be generalized into an option like `use->sameAs(p, /*shallow=*/false)` or something.

jjsjann123: That sounds reasonable.

I guess we can just iterate through all inputs and check for `!simplifyExpr(SimplifyingIrBuilder::eqExpr(a->input(i), b->input(i)))` for non-TensorView inputs?

Conceptually it felt a bit blurry to me. But I see how it could be useful.

csrc/preseg_passes/move_pad.cpp
jacobhinkle: :eyes:

jjsjann123:
!build

jjsjann123:
!build

jjsjann123:
!build --pybench

jjsjann123:
!build --pybench

jjsjann123:
!build

jjsjann123:
Don't see any odd failure in pybench. I'll merge this PR after CI clears.

jjsjann123:
!build

jjsjann123:
failures are related to distributed stuff. Don't seem to related but I'm gonna try merging main again.

jjsjann123:
!build

jjsjann123:
!build

jjsjann123:
!build

jjsjann123:
!build

jjsjann123:
!build

PR #2376, opened 2024-06-10
Canonicalize matmul dims in scheduleMatmul using dimension roles

jacobhinkle:
This updates `scheduleMatmul` to use the dimension roles introduced in #2303 to reorder and merge dims within each role. That means we can schedule fusions with more than one tensor in each role.

Two tests are included so far:
- MultipleMDims which performs [ M1, M2, K] @ [N, K]. This corresponds to `torch.linear` with 3D input i.e. a Linear layer with two "batch" dimensions.
- MultipleMDimsBatch which performs [M1, B, M2, K] @ [B, N, K]. This shows that M dimensions need not be contiguous. Note that vectorization is unaffected in this example since K is innermost.
I plan to add more tests to explore more scenarios.

Note that we cannot yet handle cases where K is not the innermost dimension, and I have not yet tested cases with M as the innermost output dimension. These will likely require us to modify the fusion definition to place those dimensions in the right spot in the MmaOp.


csrc/ir/utils.h
```diff
@@ -78,16 +78,6 @@ struct MatmulInputs {
...
-//! Matches the following matmul patterns.
-//! Matmul: A x B, alpha * A x B
-//! Matmul + Bias (addmm): A x B + C,  alpha * A x B + C, A x B + beta * C,
-//!   alpha * A x B  + beta * C
-//! Linear: A x B / A x B + C
-//! Assumptions:
-//! 1. For simplicity, we assume the MmaOp to be in the first operand.
-//! 2. For linear ([M, K], [N, K]), alpha, beta parameters are nullptr.
-bool matchMatmulPatterns(const UnaryOp* cast_op, MatmulInputs* matmul_inp);
-
```
jacobhinkle, line 90: This is undefined/uncalled so should've been removed in a previous PR.

csrc/scheduler/matmul_utils.cpp
```diff
@@ -296,6 +270,44 @@ std::string isMatmulFusionDefinitionSupported(
...
+      return "Canonical dim order must be BMNK";
+    }
+    if (role_order.at(role_order.size() - 3) != MatmulDomain::M) {
+      return "Canonical dim order must be BMNK";
+    }
+
+    // Also check that dims within each role are consecutive with one another
+    // for this pattern.
+    // TODO: Lift this requirement by modifying the definition or setting
+    // allocation domains to support this setting in MmaOp
```
jacobhinkle, line 302: I believe this condition is sufficient to avoid the problematic cases like the included test `MultipleNonConsecutiveNDims`.

csrc/scheduler/mma_utils.cpp
```diff
@@ -1859,11 +1876,20 @@ std::vector<ValGroup> canonicalDimOrdering(
...
-      ordering.push_back(*it);
+      const ValGroup& g = *it;
+      if (skip_device_dims && device_dims.has(g)) {
+        continue;
+      }
+      ordering.push_back(g);
     }
   };
+  // Insert the device dims first, then skip them when inserting dims from each
+  // other role
```
jacobhinkle, line 1891: Ensuring device dim groups are placed outside, even if that means we have non-consecutive dims within each role.

csrc/scheduler/matmul.cpp
```diff
@@ -901,10 +901,17 @@ void scheduleMatmul(Fusion* fusion, const MatmulParams& params) {
...
+
   // Make a CTA tile
   // ------------------------------------------------------------------
   // Dimensions ordered as: [ (device dims), (batch dims), M, N, K ]
-  mma_utils::canonicalizeMmaTvOrdering(mma_result);
+  mma_utils::canonicalizeMmaTvOrdering(
+      mma_result,
+      id_model.idGraph(IdMappingMode::PERMISSIVE),
+      id_roles,
+      ordering);
```
jacobhinkle, line 914: We could probably collect the permissive graph, dim roles, tensor roles, and canonical ordering into a single struct, but I worry a little about lifetime since often when we modify the fusion definition we need to rebuild the permissive map which will invalidate the previously saved `ValGroup`s. For now, I'm just trusting the calling scope to manage all that.

tests/cpp/test_gpu_tensorcore.cpp
```diff
@@ -3222,6 +3222,224 @@ TEST_F(GPUTTensorCoreTest, MisalignedVectorization) {
...
+  ASSERT_TRUE(getBankConflictInfo(fe.kernel()).empty());
+  auto cg_outputs = fe.runFusion(inputs);
+  auto Bpermuted = B.permute({{1, 2}}).reshape({N1 * N2, K});
+  auto tref = at::linear(A.to(at::kFloat), Bpermuted.to(at::kFloat))
+                  .reshape({M, N1, N2});
+  NVF_CHECK(cg_outputs[0].allclose(tref, 0.0001, 0.0001));
+}
+
+// This is a tougher test where we insert a batch dim between the M dims
+TEST_F(GPUTTensorCoreTest, MultipleMDimsBatch) {
```
jacobhinkle, line 3392: A batch dim can be inserted between M dims, but not the K dim. This is because the batch dimension gets parallelized, so that M1 and M2 will be contiguous with one another in the smem tensor.

tests/cpp/test_gpu_tensorcore.cpp
```diff
@@ -3222,6 +3222,224 @@ TEST_F(GPUTTensorCoreTest, MisalignedVectorization) {
...
+
+  // Note that M2 is inside K, so this is an NN layout
+  auto tv0 = makeContigTensor(3, DataType::Half); // M1, M2, K
+  auto tv1 = makeContigTensor(2, DataType::Half); // N, K
+
+  fusion.addInput(tv0);
+  fusion.addInput(tv1);
+
+  tv0 = broadcast(tv0, {false, false, true, false});
+  tv1 = broadcast(tv1, {true, true, false, false});
```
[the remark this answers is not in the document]
jacobhinkle: Oops. Error in the comment. This is the consecutive Ms case. I'll rename it as such as well.

csrc/scheduler/matmul_utils.cpp
```diff
@@ -296,6 +270,44 @@ std::string isMatmulFusionDefinitionSupported(
...
+  {
+    std::vector<ValGroup> dim_ordering = mma_utils::canonicalDimOrdering(
+        tensor_roles, id_roles, permissive_graph);
+    VectorOfUniqueEntries<MatmulDomain> role_order;
+    for (const ValGroup& g : dim_ordering) {
+      const auto it = id_roles.find(g);
+      NVF_ERROR(it != id_roles.end());
+      role_order.pushBack(it->second);
+    }
+    NVF_ERROR(
```
zasdfgbnm, line 285: So `scheduleMatmul` can now partially support having multiple M/N dims, but we are still rejecting it in the scheduler for now.

zasdfgbnm: Wait, should this be the following?
```C++
if (role_order.size() != 3 && role_order.size() != 4) {
  return "Expected either….";
}
```

csrc/scheduler/mma_utils.h
```diff
@@ -65,14 +69,13 @@ void makeTile(TensorView* tv, std::vector<int64_t> tile_sizes);
...
-//! Orders the root id ordering of the given tv as
-//! [Device, Batch, Previous Reduction, M, N, K]
-//!  for easier processing of later scheduling steps.
-//!
-//! This matching works on root domain only, and
-//!  will throw if the tv has a loop iterdomain that is
-//!  not a root id.
-void canonicalizeMmaTvOrdering(TensorView* tv);
+//! Orders the leaf ID canonically, and merges dims of the same role
+//! The return value gives the role of each leaf IterDomain in tv.
```
zasdfgbnm, line 73:
```suggestion
//! The return value gives the role of each loop IterDomain in tv.
```

csrc/scheduler/mma_utils.h
```diff
@@ -65,14 +69,13 @@ void makeTile(TensorView* tv, std::vector<int64_t> tile_sizes);
...
+std::vector<MatmulDomain> canonicalizeMmaTvOrdering(
```
zasdfgbnm, line 74: Is the return value of this function used?

zasdfgbnm (approved):

csrc/scheduler/matmul_utils.cpp
jacobhinkle: > Wait, should this be the following?

Ah! Good catch!

jacobhinkle: > So `scheduleMatmul` can now partially support having multiple M/N dims, but we are still rejecting it in the scheduler for now?

This check will not complain if there are multiple M or N dims, since we build `role_order` as the ordering of the _roles_ not all the dimensions that constitute all the roles. For example we might have A[M1, K1, M2] and B[K1, N1] and ordering [M, N, K].

tests/cpp/test_gpu_tensorcore.cpp
```diff
@@ -3222,6 +3222,289 @@ TEST_F(GPUTTensorCoreTest, MisalignedVectorization) {
...
+  NVFUSER_TEST_CUDA_ARCH_COMPILE_CHECK(
+      8, 0, fe.compileFusion(&fusion, inputs, LaunchParams(), matmul_cparams));
+  ASSERT_TRUE(getBankConflictInfo(fe.kernel()).empty());
+  auto cg_outputs = fe.runFusion(inputs);
+  auto Bpermuted = B.permute({{1, 2}}).reshape({N1 * N2, K});
+  auto tref = at::linear(A.to(at::kFloat), Bpermuted.to(at::kFloat))
+                  .reshape({M, N1, N2});
+  NVF_CHECK(cg_outputs[0].allclose(tref, 0.0001, 0.0001));
+}
+
```
zasdfgbnm, line 3455: Do we also need a `MultipleConsecutiveKDims`? Or, maybe combine all these tests together to do a `(B, B, M, M, N, N, K, K)` matmul?

tests/cpp/test_gpu_tensorcore.cpp
```diff
@@ -3222,6 +3222,289 @@ TEST_F(GPUTTensorCoreTest, MisalignedVectorization) {
...
+  FusionGuard fg(&fusion);
+
+  // Note that M2 is inside K, so this is an NN layout
+  auto tv0 = makeContigTensor(3, DataType::Half); // M1, K, M2
+  auto tv1 = makeContigTensor(2, DataType::Half); // N, K
+
+  fusion.addInput(tv0);
+  fusion.addInput(tv1);
+
+  // M1, N, K, M2
```
zasdfgbnm, line 3298: Should this be `M1, N, M2, K`?

csrc/scheduler/mma_utils.h
jacobhinkle: It's not currently used. I was planning to use this in a future PR to determine the ordering of M and N when we apply this to mma_result. That way we could detect an output transpose and modify the MmaOp inputs to do matmul+transpose efficiently.

tests/cpp/test_gpu_tensorcore.cpp
```diff
@@ -3222,6 +3222,289 @@ TEST_F(GPUTTensorCoreTest, MisalignedVectorization) {
...
+  Fusion fusion;
+  FusionGuard fg(&fusion);
+
+  // Note that M2 is inside K, so this is an NN layout
+  auto tv0 = makeContigTensor(4, DataType::Half); // M1, Batch, M2, K
+  auto tv1 = makeContigTensor(3, DataType::Half); // Batch, N, K
+
+  fusion.addInput(tv0);
+  fusion.addInput(tv1);
+
```
zasdfgbnm, line 3469:
```C++
// M1, B, M2, N, K
```

tests/cpp/test_gpu_tensorcore.cpp
jacobhinkle: Yeah they are pretty orthogonal so I think I could just combine them into one.

zasdfgbnm (approved):

csrc/scheduler/matmul_utils.cpp
zasdfgbnm: > > So `scheduleMatmul` can now partially support having multiple M/N dims, but we are still rejecting it in the scheduler for now?
> 
> This check will not complain if there are multiple M or N dims, since we build `role_order` as the ordering of the _roles_ not all the dimensions that constitute all the roles. For example we might have A[M1, K1, M2] and B[K1, N1] and ordering [M, N, K].

But wouldn't the `role_order.size()` check reject the fusion?

jacobhinkle: No, `role_order` would be [M, N, K] in that case. It doesn't hold dimensions, but roles.

tests/cpp/test_gpu_tensorcore.cpp
jacobhinkle: Actually, this is M1, N, K, M2, it's just that I screwed up the definition. I think I will combine this with the other disabled test and have non-consecutive, M, N, and K dims. Similarly I'll combine the enabled consecutive tests

tests/cpp/test_gpu_tensorcore.cpp
jacobhinkle: `MmaOp` currently has a restriction that there can only be a single K dimension. I added a check for this in `isMatmulFusionDefinitionSupported` since such a pattern could be created as a mul-sum; it cannot be created with matmul or linear.

PR #2379, opened 2024-06-10
Allow domains to be decoupled: validateDomainEquivalence

zasdfgbnm:
In the past, it was assumed that domains in a `TensorDomain` has linear structure `root -> logical -> allocation -> leaf`. I am removing this restriction. As the step 0 of the removal of this restriction, I am modifying the `validateDomainEquivalence`, which is called in the constructor of `TensorDomain`, to support arbitrary dependency order.

### DIFF
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -3037,10 +3037,10 @@ TensorDomain::TensorDomain(
                              : std::move(contiguity)) {
   validateContiguity(maybeAllocation(), contiguity_);
 
-  if (!logical_domain_.empty()) {
-    NVF_CHECK(!loop_domain_.empty(), "Root domain is not empty but loop is");
-    ir_utils::validateDomainEquivalence(logical_domain_, loop_domain_);
-  }
+  NVF_CHECK(
+      loop_domain_.empty() == logical_domain_.empty(),
+      "logical domain and loop domain can only be both empty or neither empty");
+  ir_utils::validateDomainEquivalence(logical_domain_, loop_domain_);
 
   // resetDomains initializes other member variables, required by clang-tidy
   resetDomains();
@@ -3061,13 +3061,12 @@ TensorDomain::TensorDomain(
                              : std::move(contiguity)) {
   validateContiguity(maybeAllocation(), contiguity_);
 
+  NVF_CHECK(
+      loop_domain_.empty() == logical_domain_.empty(),
+      "logical domain and loop domain can only be both empty or neither empty");
+  ir_utils::validateDomainEquivalence(logical_domain_, loop_domain_);
   if (!root_domain_.empty()) {
-    NVF_CHECK(!loop_domain_.empty(), "Root domain is not empty but loop is");
-    ir_utils::validateDomainEquivalence(root_domain_, loop_domain_);
-    if (!logical_domain_.empty()) {
-      ir_utils::validateDomainEquivalence(root_domain_, logical_domain_);
-      ir_utils::validateDomainEquivalence(logical_domain_, loop_domain_);
-    }
+    ir_utils::validateDomainEquivalence(logical_domain_, root_domain_);
   }
 
   // resetDomains initializes other member variables, required by clang-tidy
@@ -3091,17 +3090,15 @@ TensorDomain::TensorDomain(
                              : std::move(contiguity)) {
   validateContiguity(maybeAllocation(), contiguity_);
 
+  NVF_CHECK(
+      loop_domain_.empty() == logical_domain_.empty(),
+      "logical domain and loop domain can only be both empty or neither empty");
+  ir_utils::validateDomainEquivalence(logical_domain_, loop_domain_);
   if (!root_domain_.empty()) {
-    NVF_CHECK(!loop_domain_.empty(), "Root domain is not empty but loop is");
-    ir_utils::validateDomainEquivalence(root_domain_, loop_domain_);
-    if (!logical_domain_.empty()) {
-      ir_utils::validateDomainEquivalence(root_domain_, logical_domain_);
-      ir_utils::validateDomainEquivalence(logical_domain_, loop_domain_);
-    }
-    if (!allocation_domain_.empty()) {
-      ir_utils::validateDomainEquivalence(root_domain_, allocation_domain_);
-      ir_utils::validateDomainEquivalence(allocation_domain_, loop_domain_);
-    }
+    ir_utils::validateDomainEquivalence(logical_domain_, root_domain_);
+  }
+  if (!allocation_domain_.empty()) {
+    ir_utils::validateDomainEquivalence(logical_domain_, allocation_domain_);
   }
 
   // resetDomains initializes other member variables, required by clang-tidy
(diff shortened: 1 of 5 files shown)


zasdfgbnm:
!build

zasdfgbnm:
!build

csrc/ir/nodes.cpp
zasdfgbnm, line 1: Just cleanups, should be logically equivalent

zasdfgbnm:
!build

tests/cpp/test_gpu3.cpp
```diff
@@ -8236,6 +8236,57 @@ TEST_F(NVFuserTest, FusionCpAsyncPredicateError) {
...
+  Fusion fusion;
+  FusionGuard fg(&fusion);
+
+  // XX shape structure:
+  //
+  // domain 0: [I0, I1...    I4  I5} domain 1
+  //             \  /         \  /
+  //            merge         merge
+  //             /  \         /  \.
+  // domain 1: {I2  I3    ...I6  I7] domain 0
```
naoyam, line 8249: I'm not sure how to interpret this comment. For example, what does `[I0, I1... I4 I5}` mean?

zasdfgbnm: > For example, what does `[I0, I1... I4 I5}` mean?

This means nothing. The two returned domains are:
```
[I0, I1, I6  I7]
```
and 
```
[I2,  I3, I4  I5]
```

csrc/ir/utils.cpp
```diff
@@ -820,84 +820,36 @@ std::vector<TensorView*> getTVsWithDynamicTransform(Fusion* fusion) {
   return dynamic_tvs;
 }
 
-namespace {
+void validateDomainEquivalence(
```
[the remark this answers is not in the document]
zasdfgbnm: Updated

tests/cpp/test_gpu3.cpp
naoyam: Isn't the second one `[I4, I5, I2, I3]`? Or I'm still missing something important. I still don't know how to interpret this comment.

zasdfgbnm: added a comment about what is domain 0 and what is domain 1

zasdfgbnm: > Isn't the second one `[I4, I5, I2, I3]`? Or I'm still missing something important. I still don't know how to interpret this comment.

I think it is just a matter of whether you read from top to bottom (`[I4, I5, I2, I3]`) or from left to right (`[I2, I3, I4, I5]`). Here I am taking the left to right approach.

naoyam: Hmm, not sure. I thought that the initial domain vector is `[I0, I1, I2, I3]`. And I thought that the two returned domains you referred to were generated by:

```
AbstractTensor dom0({id0, id1, id2, id3});
    AbstractTensor dom1 = dom0;
    dom0.merge(2);
    dom0.split(2, 256);
    dom1.merge(0);
    dom1.split(0, 256);
```

I'm assuming `I0` corresponds to `id0` (and `I1` to `id1`, etc). And also `dom0` in the code corresponds to `domain 0` in the comment. 

Both `domain 0` and `domain 1` start with `[I0, I1, I2, I3]`. `Domain 0` is then transformed with `merge(2)` and `split(2, 256)`, resulting in `[I0, I1, I2*I3/256, 256]`, where `I2*I3/256` and `256` corresponds to `I6` and `I7`, respectively.

Since `dom1` has `merge(0)` followed by `split(0, 256)`, I thought `dom1` should be transformed to `[I0*I1/256, 256, I2, I3]`, and I assumed `I0*I1/256` and `256` correspond to `I4` and `I5`.

Please let me know where I misunderstand.

zasdfgbnm: Oh, I see the confusion. Your understanding is correct. My naming for I2, I3 and I4, I5 was flipped. Updated the comment in the code.

zasdfgbnm:
!build

tests/cpp/test_gpu3.cpp
naoyam: It's clear now. Thanks!

tests/cpp/test_gpu3.cpp
```diff
@@ -8236,6 +8236,59 @@ TEST_F(NVFuserTest, FusionCpAsyncPredicateError) {
...
+    auto id3 = IterDomainBuilder(s3->fusion()->zeroVal(), s3).build();
+    AbstractTensor dom0({id0, id1, id2, id3});
+    AbstractTensor dom1 = dom0;
+    dom0.merge(2);
+    dom0.split(2, 256);
+    dom1.merge(0);
+    dom1.split(0, 256);
+    return std::make_pair(dom0.as<IterDomain*>(), dom1.as<IterDomain*>());
+  };
+  auto [root_xx0, root_xx1] = create_xx_shape_structure();
```
naoyam, line 8269: It seems to me that these domains are not tied to be root domains. They are just two domains derived from the same domain but differently. They are used as part of root, logical and allocation domains. Am I understanding it correctly? If so, I found calling them "root_xx" a little confusing.

csrc/ir/utils.cpp
```diff
@@ -820,84 +820,36 @@ std::vector<TensorView*> getTVsWithDynamicTransform(Fusion* fusion) {
...
-          derived_domain_ == frontier_,
-          "Invalid derived domain. Initial domain: ",
-          toDelimitedString(initial_domain),
-          ". Derived domain: ",
-          toDelimitedString(derived_domain));
-    }
-  };
+  std::vector<Val*> dom0_val(dom0.begin(), dom0.end());
+  std::vector<Val*> dom1_val(dom1.begin(), dom1.end());
+  auto forward_exprs = DependencyCheck::getAllExprsBetween(dom0_set, dom1_val);
```
naoyam, line 847: How does this work when dom0 and dom1 are not directly dependent? For example, suppose we have:

```
Root: [I0]

// split(I0, 4)
dom0: [I0/4, 4]

// split(I0, 5)
dom1: [I0/5, 5]
```

In this case, it seems there's no forward nor backward exprs between `dom0` and `dom1` because they are not directly derived from one to another.

naoyam: As discussed offline, this is not considered, which is fine for the use case of this validation routine.

csrc/ir/utils.h
```diff
@@ -493,23 +493,29 @@ bool hasResizedRfactor(const TensorView* tv);
...
-//! domain.
+//! Another example, if we have
+//!  I0  I1  I2  I3
+//!   \  /    \  /
+//!    I4      I5
+//!   /  \    /  \.
+//!  I6  I7  I8  I9
+//! Then [I0, I1, I8, I9] is equivalent to [I6, I7, I2, I3]. [I0, I1, I2, I3] is
+//! equivalent to [I6, I7, I8, I9]. But [I0, I1, I8, I3] is NOT equivalent to
+//! [I6, I7, I2, I9]
```
naoyam, line 515: Can you please add a comment about a case like `dom0` and `dom1` are both derived from the same domain, which is not supported.

naoyam (approved):
LGTM

csrc/ir/utils.h
zasdfgbnm: added

PR #2380, opened 2024-06-10
Support indexing of vectorized tensors

naoyam:
Just supporting indexing with vectorization. Note that this is still only for indexing used in tensor ops themselves. Predicate indexing will be supported later.

There's only one thing that needs to be done, which is to use zero for a vectorized domain. This could be done by `shouldUseZeroIndex`, but for now I decided to generate an index and then replace the corresponding val with zero. This is because vectorized domains do not always just use zero but `N-1`, where `N` is the extent of the domain, for predication.


tests/cpp/test_indexing.cpp
```diff
@@ -57,6 +57,27 @@ Val* modExpr(Args&&... args) {
   return SimplifyingIrBuilder::modExpr(std::forward<Args>(args)...);
 }
 
+void printAllIndices(
```
naoyam, line 60: Test-internal utility function.

csrc/id_model/indexing.h
```diff
@@ -87,6 +88,23 @@ class TensorIndexer {
...
+  // This replacement map is used to replace a tensor index after an
+  // index map is generated. Since replacment is only done for loop
+  // domains, this could be done as part of getInitialIndexMap. One
+  // reason that we might want to first generate an index and do some
+  // replacements, rather than using final index vals to build the
+  // index map, is that one index map could be used for multiple
+  // indices. For normal tensor indexing, this may not matter, but for
+  // predicate indexing, it needs to generate both start and stop
+  // predicates, and one index map would be sufficient for both
+  // indices by using different replacement maps.
```
jacobhinkle, line 103: Thanks. This comment addresses my first question of why not just set the index to zero before building the map.

csrc/id_model/indexing.cpp
```diff
@@ -509,8 +514,31 @@ IndexingInfo TensorIndexer::computeIndex(
     index_compute.propagate(expr_group, direction);
   }
 
-  IndexingInfo info{traversal_path, index_compute.indexMap()};
+  IndexingInfo info{loop_domains, traversal_path, index_compute.indexMap()};
   return info;
 }
 
+std::unordered_map<Val*, Val*> TensorIndexer::getIndexReplacementMap(
```
jacobhinkle, line 521: I think we should name this `getVectorizedIndexReplacementMap` since we skip non-vectorized groups in line 531.

jacobhinkle (approved):
LGTM

csrc/id_model/indexing.cpp
naoyam: Let me leave it as is. I may add more stuff here.

PR #2384, opened 2024-06-11
Replace `allgather` by `_allgather_base`

samnordmann:
`allgather` primitive in c10d induces extra buffering time to copy out the received tensors into user buffer. It is therefore always preferable to use `_allgather_base`, which does not perform any extra copy at the cost of assuming that the receive buffers are placed contiguously.

Illustration, with `allgather`, you can see the red segment indicating `cudaMemcpyAsync`:
Screenshot 2024-06-11 at 19 10 18

With `_allgather_base`, the `cudaMemcpyAsync` disappear
Screenshot 2024-06-11 at 19 11 21

### DIFF
--- a/csrc/multidevice/communication.cpp
+++ b/csrc/multidevice/communication.cpp
@@ -253,6 +253,11 @@ c10::intrusive_ptr<c10d::Work> postAllgather(
   assertBufferCount(splits, communication->team().size());
   assertBuffersHaveSameSize({input_tensor}, splits);
 
+  // allgather primitive in c10d induces extra buffering time to copy out the
+  // received tensors into user buffer. It is therefore always preferable to use
+  // _allgather_base, which does not perform any extra copy at the cost of
+  // assuming that the receive buffers are placed contiguously. See #2384 for an
+  // illustration.
   return backend->_allgather_base(output_tensor, input_tensor, {});
 }
 
--- a/csrc/multidevice/communication.cpp
+++ b/csrc/multidevice/communication.cpp
@@ -249,7 +249,6 @@ c10::intrusive_ptr<c10d::Work> postAllgather(
     c10d::Backend* backend,
     at::Tensor input_tensor,
     at::Tensor output_tensor) {
-
   auto splits = at::split(output_tensor, /*split_size=*/1, /*dim=*/0);
   assertBufferCount(splits, communication->team().size());
   assertBuffersHaveSameSize({input_tensor}, splits);
--- a/csrc/multidevice/communication.cpp
+++ b/csrc/multidevice/communication.cpp
@@ -249,13 +249,12 @@ c10::intrusive_ptr<c10d::Work> postAllgather(
     c10d::Backend* backend,
     at::Tensor input_tensor,
     at::Tensor output_tensor) {
-  std::vector<at::Tensor> input_tensors({input_tensor});
-  std::vector<std::vector<at::Tensor>> output_tensors(1);
-  output_tensors[0] = at::split(output_tensor, /*split_size=*/1, /*dim=*/0);
 
-  assertBufferCount(output_tensors[0], communication->team().size());
-  assertBuffersHaveSameSize(input_tensors, output_tensors[0]);
-  return backend->allgather(output_tensors, input_tensors, {});
+  auto splits = at::split(output_tensor, /*split_size=*/1, /*dim=*/0);
+  assertBufferCount(splits, communication->team().size());
+  assertBuffersHaveSameSize({input_tensor}, splits);
+
+  return backend->_allgather_base(output_tensor, input_tensor, {});
 }
 
 c10::intrusive_ptr<c10d::Work> postScatter(


samnordmann:
!build

csrc/multidevice/communication.cpp
```diff
@@ -249,13 +249,11 @@ c10::intrusive_ptr<c10d::Work> postAllgather(
...
-  std::vector<std::vector<at::Tensor>> output_tensors(1);
-  output_tensors[0] = at::split(output_tensor, /*split_size=*/1, /*dim=*/0);
+  auto splits = at::split(output_tensor, /*split_size=*/1, /*dim=*/0);
+  assertBufferCount(splits, communication->team().size());
+  assertBuffersHaveSameSize({input_tensor}, splits);
 
-  assertBufferCount(output_tensors[0], communication->team().size());
-  assertBuffersHaveSameSize(input_tensors, output_tensors[0]);
-  return backend->allgather(output_tensors, input_tensors, {});
+  return backend->_allgather_base(output_tensor, input_tensor, {});
```
wujingyue, line 256:
```suggestion
  // allgather primitive in c10d induces extra buffering time to copy out the received tensors into user buffer. It is therefore always preferable to use _allgather_base, which does not perform any extra copy at the cost of assuming that the receive buffers are placed contiguously.
  // See #2384 for an illustration. 
  return backend->_allgather_base(output_tensor, input_tensor, {});
```

wujingyue (approved):

cowanmeg (commented):
LGTM! Also an FYI there is this comment saying _allgather_base will be deprecated in the "near" future https://github.com/pytorch/pytorch/blob/1d233b8f500f3fafb96e944953579a33a3c1c24e/torch/csrc/distributed/c10d/ProcessGroup.hpp#L239

The comment is already 3 years old, so I don't think we need too worry much.

csrc/multidevice/communication.cpp
wujingyue: It'd also be helpful to link to the PyTorch code where this extra buffering happens. I'm a bit surprised that the default API is slow and the fix seems to be a quite straightforward replacement. cc @Aidyn-A and @syed-ahmed to see if it's worth a fix upstream.

samnordmann: > It'd also be helpful to link to the PyTorch code where this extra buffering happens.

The extra copy happens: 
- in nccl: [here](https://github.com/pytorch/pytorch/blob/1edcb31d34ef012d828bb9f39a8aef6020f580b2/torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp#L3589
) 
- in ucc: [here](https://github.com/pytorch/pytorch/blob/1edcb31d34ef012d828bb9f39a8aef6020f580b2/torch/csrc/distributed/c10d/ProcessGroupUCC.cpp#L940). Note that in ucc there is an alternative solution, which is [to use allgatherv](https://github.com/pytorch/pytorch/blob/1edcb31d34ef012d828bb9f39a8aef6020f580b2/torch/csrc/distributed/c10d/ProcessGroupUCC.cpp#L873) with a destination buffer equalling `nullptr`; this avoids the extra copy, however, allgatherv is less efficient than allgather

> I'm a bit surprised that the default API is slow and the fix seems to be a quite straightforward replacement. cc @Aidyn-A and @syed-ahmed to see if it's worth a fix upstream.

I am not sure this deserves a "fix" upstream, it is done like that on purpose, to provide more flexibility to the user.
1) On the one hand, communication backend like ucc and nccl (idk for Gloo) work with raw void pointers, i.e., fully contiguous src and dst buffers. This restriction is necessary for performance.
2) On the other hand, ProcessGroup API allows users to provide a vector of Tensors instead of a raw pointer for one-to-many, many-to-many and many-to-one collective operations. This makes the API very convenient for users who don't need advanced optimization. In other words, ProcessGroup::allgather expects as dst buffers a vector of n buffer of size m, whereas nccl and ucc expect one buffer of size n*m. To bridge this gap, ProcessGroup needs to copy in/out the vector of buffers into one big contiguous chunk before using the backend, hence the extra cudaMemcpyAsync


To provide "best of both worlds" to the advanced users, ProcessGroup exposes `_allgather_base` `_reduce_scatter_base` and `alltoall_base`, which directly use nccl or ucc backend with raw pointers assuming the buffers are fully contiguous.

The other "base" collectives (e.g. scatter_base, gather_base) are missing in the API though.

Therefore, imo, there is no change upstream necessary, however, as far as nvFuser is concerned, we should only use the "base" collective when they are available.

Coming next a PR to patch `reduce_scatter`

samnordmann:
> Pardon me please for being late to the party. @cowanmeg is right, the `_allgather_base` signature is indeed deprecated. The operation went into a public API as [all_gather_into_tensor](https://pytorch.org/docs/stable/distributed.html#torch.distributed.all_gather_into_tensor).

Let me point out that `all_gather_into_tensor` is a python API call only, which calls `_allgather_base` internally
https://github.com/pytorch/pytorch/blob/15ab636007223803b7d77c14b881ea337ce75a32/torch/distributed/distributed_c10d.py#L3200

Aidyn-A:
> > Pardon me please for being late to the party. @cowanmeg is right, the `_allgather_base` signature is indeed deprecated. The operation went into a public API as [all_gather_into_tensor](https://pytorch.org/docs/stable/distributed.html#torch.distributed.all_gather_into_tensor).
> 
> 
> 
> Let me point out that `all_gather_into_tensor` is a python API call only, which calls `_allgather_base` internally
> 
> https://github.com/pytorch/pytorch/blob/15ab636007223803b7d77c14b881ea337ce75a32/torch/distributed/distributed_c10d.py#L3200

Right. I wasn't paying attention to the code language. The good thing is that the existence of public Python API guarantees that the operation will be kept. It can get renamed though...

PR #2385, opened 2024-06-11
add pointwise benchmark

liqiangxl:
**What is in this PR?**
add a simple pointwise benchmark
```
def simple_pointwise_fwd_fn(inputs: list):  # in_tensor
    return torch.mul(inputs[0], inputs[0])
```
**Why need to add this?**
(1) serve as a basline to check the best bandwidth we can expect for a specific problem size
(2) check whether the device is healthy.

**Current performance:** see [link](https://docs.google.com/document/d/1FtIU_s-iw_Q9C7psdJn5FMp9ApKdzqdkJe4p6R9TQkU/edit?usp=sharing).

### DIFF
--- a/benchmarks/python/test_simple_pointwise.py
+++ b/benchmarks/python/test_simple_pointwise.py
@@ -9,34 +9,29 @@ import torch
 from .global_params import generate_input_sizes, FLOAT_DTYPES, PROMOTE_DTYPES
 
 
-def reduction_fusion(
+def simple_pointwise_fusion(
     fd: FusionDefinition,
     dtype: DataType,
-    reduction_axis: int,
 ) -> None:
     T0 = fd.define_tensor(
         shape=[-1, -1], contiguity=[True, True], dtype=dtype, is_cpu=False
     )
     if dtype in PROMOTE_DTYPES:
         T0 = fd.ops.cast(T0, dtype=DataType.Float)
-    T2 = fd.ops.sum(T0, dims=[reduction_axis], keepdim=False)
+    T2 = fd.ops.mul(T0, T0)
     if dtype in PROMOTE_DTYPES:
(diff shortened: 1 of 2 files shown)


benchmarks/python/test_simple_pointwise.py
```diff
@@ -0,0 +1,72 @@
...
+# SPDX-License-Identifier: BSD-3-Clause
+import pytest
+from nvfuser import FusionDefinition, DataType
+from nvfuser.pytorch_utils import torch_dtype_to_nvfuser_dtype
+from .core import run_benchmark, clear_cuda_cache
+import torch
+from .global_params import generate_input_sizes, FLOAT_DTYPES, PROMOTE_DTYPES
+
+
+def simple_pointwise_fusion(
```
Priya2698, line 12: nit:
```Suggestion
def pointwise_mul_fusion(
Similarly for other names in the file

File: benchmarks/python/test_simple_pointwise.py [hunk 2]
```diff
@@ -0,0 +1,72 @@
...
+def test_simple_pointwise_nvf_benchmark(
+    benchmark,
+    size: tuple,
+    dtype: torch.dtype,
+    disable_validation: bool,
+    disable_benchmarking: bool,
+):
+    clear_cuda_cache()
+
+    inputs = [torch.randn(*size, device="cuda", dtype=dtype)]
```
Priya2698, line 42: `*size` -> `size`

Priya2698 (approved):
LGTM. Only minor comments for file and function naming

benchmarks/python/test_simple_pointwise.py
liqiangxl: changed.

Issue #2372, opened 2024-06-08
DIDx aware ATen evaluation for matmul and linear.


Priya2698:
**Motivation:** Using `MatmulOp/LinearOp` in multi-GPU work. 

**Issue:** We currently do not check for the presence of `DID` axis in the inputs to `evaluate` for any of the nodes. 

**Example**: Consider: A (M x K), B (K x N) are sharded on 2-devices, with rank: 0 computing the first M/2 rows of C (MxN):
In the current approach, the inputs to evaluate will be: A = [1 x M/2 x K], B = [2 x K x N/2] (assuming there is an all-gather step to get the complete B matrix so we compute the first half of the output, instead of only the first quadrant). The output will be: C = [2 x M/2 x N/2]. 

Correct computation: reorder and reshape B = [K, N] before matmul to get [1 x M/2 x N].

CC: @wujingyue @cowanmeg. Correct me, if my understanding is incorrect.

cowanmeg:
Yes, that is the right understanding. We can assume the resharding passes will properly add appropriate collectives. One correction is that B's layout will be [K, N] after it is all-gathered. If the tensor is not sharded it won't have an extra dimension. 
A: [1, M/2, K] and B: [K, N] -> C: [1, M/2, N]
So only A, C will have to be reshaped (unsqueezing the DID axes)

1. Reshape input(s) to remove the extra DID dimensions, reorder into ATen format
2. ATen computation
3. Reshape to add DID dimensions back and reorder back to the nvFuser format

cowanmeg:
This unit test is a good one to look at: https://github.com/NVIDIA/Fuser/blob/main/tests/cpp/test_multidevice_matmul.cpp#L69
It doesn't have any communication just the matmul computation.

Priya2698:
> Yes, that is the right understanding. We can assume the resharding passes will properly add appropriate collectives. One correction is that B's layout will be [K, N] after it is all-gathered. If the tensor is not sharded it won't have an extra dimension. A: [1, M/2, K] and B: [K, N] -> C: [1, M/2, N] So only A, C will have to be reshaped (unsqueezing the DID axes)
> 
> 1. Reshape input(s) to remove the extra DID dimensions, reorder into ATen format
> 2. ATen computation

What is the expected output from the ATen computation -- [1, M/2, N] or [M/2, N]?
If A and B are as described (i.e. there is no reshape and reorder needed on B), then the current implementation will output C as [1, M/2, N] for matmul.

[...]

Issue #2354, opened 2024-06-06
`RuntimeError: Stride mismatch with contiguity info.`


crcrpar:
env
- H100
- `torch.__version__ = '2.4.0a0+gitfb696ef'`
- cuda = `12.5`
- `nvfuser.version() = '0.2.6+gita286c31'`

repro

```python
import torch
from nvfuser import FusionDefinition, DataType

def nvfuser_fusion_id0(fd : FusionDefinition) -> None :
    T0 = fd.define_tensor(shape=[-1, -1], contiguity=[True, True], dtype=DataType.Float, is_cpu=False, stride_order=[1, 0])
    T1 = fd.define_tensor(shape=[-1], contiguity=[True], dtype=DataType.Float, is_cpu=False, stride_order=[0])
    T2 = fd.define_tensor(shape=[-1, -1], contiguity=[True, True], dtype=DataType.Float, is_cpu=False, stride_order=[1, 0])
    T3 = fd.define_tensor(shape=[-1, -1, -1], contiguity=[True, True, True], dtype=DataType.Float, is_cpu=False, stride_order=[2, 1, 0])
    T4 = fd.ops.linear(T3, T2, T1)
    S5 = fd.define_scalar(1.41421, dtype=DataType.Double)
    S6 = fd.ops.reciprocal(S5)
    T7 = fd.ops.mul(T4, S6)
    T8 = fd.ops.erf(T7)
    S9 = fd.define_scalar(0.500000, dtype=DataType.Double)
    T10 = fd.ops.mul(S9, T8)
    S11 = fd.define_scalar(0.500000, dtype=DataType.Double)
    T12 = fd.ops.add(S11, T10)
    T13 = fd.ops.mul(T4, T12)
    T14 = fd.ops.linear(T13, T0)
    fd.add_output(T4)
    fd.add_output(T13)
    fd.add_output(T14)

with FusionDefinition() as fd:
    nvfuser_fusion_id0(fd)

inputs = [
    torch.randn((1179648,), dtype=torch.float32, device='cuda:0').as_strided((768, 1536), (1536, 1)),
    torch.randn((1536,), dtype=torch.float32, device='cuda:0').as_strided((1536,), (1,)),
    torch.randn((1179648,), dtype=torch.float32, device='cuda:0').as_strided((1536, 768), (768, 1)),
    torch.randn((6291456,), dtype=torch.float32, device='cuda:0').as_strided((2048, 4, 768), (3072, 768, 1)),
]
[...]

PR #2386, opened 2024-06-11
Use `matmul` in distributed tests

Priya2698:
Issue #2372.
Modifying the tests to use `matmul` in place of `mul-sum`.
Note: `matmul` API requires the logical shapes `[M,K] x [K,N]` and the output has the same dtype as the input.

### DIFF
--- a/tests/cpp/test_multidevice_matmul.cpp
+++ b/tests/cpp/test_multidevice_matmul.cpp
@@ -82,18 +82,14 @@ TEST_F(DistributedMatmulTest, LayoutTN_NoComms) {
 
   TensorView* a = makeContigTensor(3, DataType::Half); // (Mo,Mi,K)
   TensorView* b = makeContigTensor(2, DataType::Half); // (N,K)
-  // TensorView* a_b = broadcast(a, {false, false, true, false}); // (Mo,Mi,b,K)
-  // TensorView* b_b = broadcast(b, {true, true, false, false}); // (b,b,N,K)
-  // TensorView* ab = mul(a_b, b_b); // (Mo,Mi,N,K)
-  // TensorView* c = sum(ab, {-1}); // (Mo,Mi,N,r)
-  TensorView* c = matmul(a, b);
+  TensorView* b_t = transpose(b, 0, 1); // (K,N)
+  TensorView* c = matmul(a, b_t); //(Mo,Mi,N,r)
 
   fusion->addInput(a);
   fusion->addInput(b);
   fusion->addOutput(c);
 
   // Sharding M dimension
-  // auto all_sharded_tvs = {a, a_b, b_b, ab, c};
   auto all_sharded_tvs = {a, c};
   for (auto tv : all_sharded_tvs) {
     tv->axis(0)->parallelize(ParallelType::DIDx);
@@ -141,14 +137,12 @@ TEST_F(DistributedMatmulTest, LayoutTN_Allgather) {
   int Mo = num_devices_;
   int Mi = M / Mo;
   std::vector<int> a_shape = {Mo, Mi, K};
-  std::vector<int> b_shape = {N, K};
+  std::vector<int> b_shape = {N,K};
 
   TensorView* a = makeContigTensor(3, DataType::Half); // (Mo,Mi,K)
   TensorView* b = makeContigTensor(2, DataType::Half); // (N,K)
-  TensorView* a_b = broadcast(a, {false, false, true, false}); // (Mo,Mi,b,K)
-  TensorView* b_b = broadcast(b, {true, true, false, false}); // (b,b,N,K)
-  TensorView* ab = mul(a_b, b_b); // (Mo,Mi,N,K)
-  TensorView* c0 = sum(ab, {-1}); // (Mo,Mi,N,r)
+  TensorView* b_t = transpose(b, 0, 1); // (K,N)
+  TensorView* c0 = matmul(a, b_t); //(Mo,Mi,N,r)
   TensorView* c = set(c0);
 
   fusion->addInput(a);
@@ -156,7 +150,7 @@ TEST_F(DistributedMatmulTest, LayoutTN_Allgather) {
   fusion->addOutput(c);
 
   // Sharding M dimension
-  auto all_sharded_tvs = {a, a_b, b_b, ab, c0};
+  auto all_sharded_tvs = {a, c0};
   for (auto tv : all_sharded_tvs) {
     tv->axis(0)->parallelize(ParallelType::DIDx);
     tv->setDeviceMesh(mesh);
@@ -202,13 +196,9 @@ TEST_F(DistributedMatmulTest, LayoutNT_AllReduce) {
 
   TensorView* a = makeContigTensor(3, DataType::Half); // (Ko,Ki,M)
   TensorView* b = makeContigTensor(3, DataType::Half); // (Ko,Ki,N)
-  // Transpose into TN layout, keep Ko (device axis) as the outermost.
+  // Transpose into TT layout, keep Ko (device axis) as the outermost.
   TensorView* a_t = transpose(a, 1, 2); // (Ko,M,Ki)
-  TensorView* b_t = transpose(b, 1, 2); // (Ko,N,Ki)
-  TensorView* a_b = broadcast(a_t, {false, false, true, false}); // (Ko,M,b,Ki)
-  TensorView* b_b = broadcast(b_t, {false, true, false, false}); // (Ko,b,N,Ki)
-  TensorView* ab = mul(a_b, b_b); // (Ko,M,N,Ki)
-  TensorView* c0 = sum(ab, {-1}); // (Ko,M,N,r)
+  TensorView* c0 = matmul(a_t, b); // (Ko,M,N,r)
   TensorView* c = sum(c0, {0}); // (r,M,N)
 
   fusion->addInput(a);
@@ -216,7 +206,7 @@ TEST_F(DistributedMatmulTest, LayoutNT_AllReduce) {
   fusion->addOutput(c);
 
   // Parallelize K on all inputs and intermediates.
-  auto all_sharded_tvs = {a, b, a_t, b_t, a_b, b_b, ab, c0};
+  auto all_sharded_tvs = {a, b, a_t, c0};
   for (auto tv : all_sharded_tvs) {
     tv->axis(0)->parallelize(ParallelType::DIDx);
     tv->setDeviceMesh(mesh);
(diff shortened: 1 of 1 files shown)


wujingyue (approved):
Is it worth adding a check to make sure ATen kicked in? E.g. https://github.com/NVIDIA/Fuser/blob/7b780e5fceeba2be0a8147c699e010bef83c50be/tests/cpp/test_matmul_aten_evaluation.cpp#L318

tests/cpp/test_multidevice_matmul.cpp
```diff
@@ -247,11 +239,7 @@ TEST_F(DistributedMatmulTest, LayoutNT_ReduceScatter) {
...
   TensorView* a_t = transpose(a, 1, 2); // (Ko, M, Ki)
-  TensorView* b_t = transpose(b, 1, 2); // (Ko, N, Ki)
-  TensorView* a_b = broadcast(a_t, {false, false, true, false}); // (Ko,M,b,Ki)
-  TensorView* b_b = broadcast(b_t, {false, true, false, false}); // (Ko,b,N,Ki)
-  TensorView* ab = mul(a_b, b_b); // (Ko,M,N,Ki)
-  TensorView* c0 = sum(ab, {-1}); // (Ko,M,N,r)
+  TensorView* c0 = matmul(a_t, b); // (Ko,M,N,r)
   c0 = segment_set(c0);
   std::vector<int64_t> orig_size = {K, M, N};
   std::vector<int64_t> new_size = {K, Mo, Mi, N};
```
cowanmeg, line 245: This is actually an error from me....but K should be Ko. 
I'm surprised it compiled and was correct before.

cowanmeg (approved):
Thanks! I'm glad the ATen pathway worked out of the box!
I assume the extra size 1 axes in the aten tensors are being treated as batch dimensions which is why it works out ok right?

wujingyue:
> Thanks! I'm glad the ATen pathway worked out of the box! I assume the extra size 1 axes in the aten tensors are being treated as batch dimensions which is why it works out ok right?

Yes, that is treated as a BatchMatmul, case 5 in https://pytorch.org/docs/stable/generated/torch.matmul.html

Priya2698:
> EXPECT_FALSE(executors.front().hasCompiledKernel());

> Is it worth adding a check to make sure ATen kicked in? E.g.
> 
> https://github.com/NVIDIA/Fuser/blob/7b780e5fceeba2be0a8147c699e010bef83c50be/tests/cpp/test_matmul_aten_evaluation.cpp#L318

How do I access the executors from `MultiDeviceCommunicator`?
For these tests, since there will be multiple segments, we need to test for the second segment/executor (first is no-op for transpose).

Priya2698:
!build

cowanmeg:
Sorry for adding this late - can we add back the original `TEST_F(DistributedMatmulTest, LayoutTN_NoComms)` with mul-sum in addition to the updated one with the matmul node. We should keep one of these tests since they call the MatmulScheduler with sharded TensorViews. 

> How do I access the executors from MultiDeviceCommunicator?
MultiDeviceExecutor doesn't have accessor functions for executors so we would have to add that first. @wujingyue 

Some background:
The MultiDeviceExecutor takes in a Fusion and segments it into compute and communication segments. The compute segments then go through FusionExecutorCache then they go through the standard segmentation.
So we need to do something like get the first "coarse" segment (which will correspond to the local compute at the beginning) then index into the proper segment/executor. 

See here for the internals: 
https://github.com/NVIDIA/Fuser/blob/main/csrc/multidevice/executor.h#L130
The MultiDeviceExecutor can use either FusionExecutorCache or directly use FusionExecutor for compute segments depending on flags (temporary for development purposes). For the DistributedMatmulTest I set the params to use FusionExecutorCache.

wujingyue:
> How do I access the executors from `MultiDeviceExecutor`? 

Good question. We could plumb it through but I don't think it's worth the benefits at this moment.

wujingyue:
> Can we add back the original `TEST_F(DistributedMatmulTest, LayoutTN_NoComms)` with mul-sum in addition to the updated one with the matmul node. We should keep one of these tests since they call the MatmulScheduler with sharded TensorViews.

Good question! 

AFAIK, nvFuser offers three ways (arguably too many, 🤷) to generate a matmul. 
1. [`fusedMultiplySum`](https://github.com/NVIDIA/Fuser/blob/7efce011add02eb6ec2342b074cd4ccac89bc5a2/csrc/ops/arith.h#L708), which takes broadcasted operands and generates an MmaOp (for MatmulScheduler)
2. [`matmul`](https://github.com/NVIDIA/Fuser/blob/b56c3e7960f18e93c2d806d3167c961ebe7f2b20/csrc/ops/composite.h#L77), which takes non-broadcasted operands and sometimes generates a MatmulOp (for ATen).
3. The same `matmul` API, which sometimes generates a decomposed broadcast+mul+sum pattern (e.g. when the inputs are vectors). 

The current tests in this file exercise only (2). 

I agree it's useful to exercise (1). MatmulScheduler is already DID-capable and eventually we do want nvFuser to generate matmul using the scheduler, so it's good to have tests to maintain that feature. However, I'm not aware that nvFuser can reliably turn a decomposed broadcast+mul+sum into an MmaOp/MatmulOp. Correct me if I'm wrong. So we may need to add new tests that use fusedMultiplySum instead. 

I was told by @Priya2698 that (3) is a corner case that will go away. So I'll leave the decision to her whether to cover that as well.

Priya2698:
> 1. [`fusedMultiplySum`](https://github.com/NVIDIA/Fuser/blob/7efce011add02eb6ec2342b074cd4ccac89bc5a2/csrc/ops/arith.h#L708), which takes broadcasted operands and generates an MmaOp (for MatmulScheduler)
> 2. [`matmul`](https://github.com/NVIDIA/Fuser/blob/b56c3e7960f18e93c2d806d3167c961ebe7f2b20/csrc/ops/composite.h#L77), which takes non-broadcasted operands and sometimes generates a MatmulOp (for ATen).
> 3. The same `matmul` API, which sometimes generates a decomposed broadcast+mul+sum pattern (e.g. when the inputs are vectors).
> 
> The current tests in this file exercise only (2).
> 
> I agree it's useful to exercise (1). MatmulScheduler is already DID-capable and eventually we do want nvFuser to generate matmul using the scheduler, so it's good to have tests to maintain that feature. However, I'm not aware that nvFuser can reliably turn a decomposed broadcast+mul+sum into an MmaOp/MatmulOp. Correct me if I'm wrong. So we may need to add new tests that use fusedMultiplySum instead.
> 
> I was told by @Priya2698 that (3) is a corner case that will go away. So I'll leave the decision to her whether to cover that as well.

Yes, (3) will be removed. `matmul` API will always generate the `MatmulOp`.

For now, from the python frontend API, the only way is to call `fd.ops.matmul` or similarly a `fd.ops.mul + fd.ops.sum`. However, thunder will plumb down matmuls using the former.

To exercise the matmul scheduler:
1. Create a `mul-sum` pattern or `fusedMultiplySum` which both convert to `MmaOp`: https://github.com/NVIDIA/Fuser/pull/2272

2. @jacobhinkle added translation for the `MatmulOp` to `MmaOp` for matmul scheduler: https://github.com/NVIDIA/Fuser/pull/2236. Use `NVFUSER_ENABLE=fuse_matmul` which enables the matmul scheduler to take segments with matmul if expression evaluator cannot (for eg: with epilogues). We can also disable expression evaluator using `NVFUSER_DISABLE=matmul_expr_eval` to only use one or the other.

wujingyue:
Thanks for clarifying. I missed https://github.com/NVIDIA/Fuser/blob/0f66dc230bda8603ce21308e2961a85a8f3b757a/csrc/scheduler/mma_utils.cpp#L1462 -- nvFuser does convert a decomposed mul+sum to MmaOp. 

So we have three ways to exercise the matmul scheduler: 
1. `mul-sum`
2. `fusedMultiplySum`
3. `matmul` with `fuse_matmul` enabled. 

Wdyt? (3) seems to me the best option: we'll get `fd.ops.matmul` from Thunder and stick with that for at least O(months). Btw, when are you going to [enable fd.ops.matmul for Thunder](https://github.com/Lightning-AI/lightning-thunder/blob/c21533c12a2a826aee84e011c415b216cb6f779d/thunder/executors/nvfuserex_impl.py#L2252-L2253) by default?

Priya2698:
> Wdyt? (3) seems to me the best option: we'll get `fd.ops.matmul` from Thunder and stick with that for at least O(months).

Yes, we can use the `NVFUSER_ENABLE/DISABLE` options to control which scheduler to use. It would good to have some way to verify the scheduler heuristic used though. There is no need to plumb down a separate option if I can use `MultiDeviceCommunicator->fec_`?

> Btw, when are you going to [enable fd.ops.matmul for Thunder](https://github.com/Lightning-AI/lightning-thunder/blob/c21533c12a2a826aee84e011c415b216cb6f779d/thunder/executors/nvfuserex_impl.py#L2252-L2253) by default?

We are still seeing some issues such as #2354, so it is not enabled by default yet. I plan on running all the thunder benchmarks and tests with this enabled to preemptively identify existing issues first.

cowanmeg:
> MultiDeviceCommunicator->fec_

We can write an accessor function to return the FusionExecutorCache(s) used by the MultiDeviceRuntime. It will be a vector because each segment has its own FusionExecutorCache. In this case since there's only one compute segment it will only have one entry. 

Related translation for the MatmulOp to MmaOp for matmul scheduler we eventually need to update `insertResharding` pass to handle Matmul, Linear, and SDPA ops to insert `set` for non-reduction collectives. It's easy to work around by manually inserting `set` which the tests were already doing.

Priya2698:
> > MultiDeviceCommunicator->fec_
> 
> We can write an accessor function to return the FusionExecutorCache(s) used by the MultiDeviceRuntime. It will be a vector because each segment has its own FusionExecutorCache. In this case since there's only one compute segment it will only have one entry.
> 
> Related translation for the MatmulOp to MmaOp for matmul scheduler we eventually need to update `insertResharding` pass to handle Matmul, Linear, and SDPA ops to insert `set` for non-reduction collectives. It's easy to work around by manually inserting `set` which the tests were already doing.

So currently, what you suggest is retaining one of the original test cases? 
Or do we want to switch to only using `MatmulOp` and use the nvfuser options to pick one of the schedulers?

cowanmeg:
Let's keep one of the original test cases.

Priya2698:
!build

cowanmeg (approved):
LGTM! Thank you!

Priya2698:
!build

PR #2388, opened 2024-06-12
Use indent utility in dynamic_transform/id_model toString functions

jacobhinkle:
This just uses the pre-existing utility in `ir/iostream.h` instead of `ss << indent` style indenting.


wujingyue:
!build

PR #2389, opened 2024-06-12
Enabling Idmodel-based indexing in lowering

naoyam:
Added options to enable both consumer and producer indexing. Specifically, `NVFUSER_ENABLE=id_model(consumer_index)` would use the IdModel-based indexing for all consumer indexing. 

Not all of the C++ tests would yet pass if enabled. For now, just updated the `IndexingTest` tests to validate indexing results obtained through lowering.


csrc/expr_simplifier.cpp
```diff
@@ -2791,6 +2791,9 @@ Val* simplifyExpr(
   std::unique_ptr<std::unordered_set<std::string>> disabled_passes = nullptr;
   if (isOptionDisabled(DisableOption::ExprSimplify)) {
     const auto& v = getDisableOptionArguments(DisableOption::ExprSimplify);
+    if (v.empty()) {
```
naoyam, line 2794: Currently, even if `expr_simplify` is disabled, `flatten` and `unflatten` are done anyway. This makes sure that nothing is really done. Confirmed with @zasdfgbnm

csrc/id_model/indexing.cpp
```diff
@@ -291,7 +292,7 @@ void IdGraphIndexCompute::handle(Merge* merge) {
     auto outer_idx = getIndex(merge->outer());
     auto inner_idx = getIndex(merge->inner());
     auto out_idx = SimplifyingIrBuilder::addExpr(
-        SimplifyingIrBuilder::mulExpr(outer_idx, inner_ext), inner_idx);
+        SimplifyingIrBuilder::mulExpr(inner_ext, outer_idx), inner_idx);
```
naoyam, line 295: Just to follow the current indexing order.

csrc/id_model/indexing.cpp
```diff
@@ -457,11 +477,10 @@ Val* TensorIndexer::getLinearIndex(TensorView* tv, const Expr* expr) const {
   // TODO: Contiguous indexing
   Val* index = tv->fusion()->zeroVal();
   for (const auto i : c10::irange(allocation_domains.size())) {
-    // Traverse from innermost to outermost
```
naoyam, line 460: I thought this was  the order used in the current indexing, but that is not true.

csrc/id_model/indexing.cpp
```diff
@@ -470,15 +489,13 @@ Val* TensorIndexer::getLinearIndex(TensorView* tv, const Expr* expr) const {
         allocation_domain->toString());
     Val* idx = idx_it->second;
     index = SimplifyingIrBuilder::addExpr(
-        index, SimplifyingIrBuilder::mulExpr(idx, stride));
+        index, SimplifyingIrBuilder::mulExpr(stride, idx));
```
naoyam, line 492: Use the same order as the current indexing

csrc/options.cpp
```diff
@@ -296,9 +323,4 @@ bool isProfilerPrintingVerbose() {
       ProfilerOption::PrintVerbose);
 }
 
-const std::vector<std::string>& getDisableOptionArguments(
```
naoyam, line 299: I believe this was just added accidentally.

tests/cpp/test_indexing.cpp
naoyam, line 1: There's no technical change here. Just added a utility class to simplify validation of lowered index vals.

naoyam:
CC: @zasdfgbnm

tests/cpp/test_indexing.cpp
naoyam, line 1: There's no technical change here. Just added a utility class to simplify validation of lowered index vals.

csrc/id_model/indexing.cpp
```diff
@@ -369,7 +370,26 @@ void TensorIndexer::buildLoopIndexMap() {
           shouldUseZeroIndex(loop_group) || isParallelTypeDeviceDim(ptype)) {
         loop_index = fusion->zeroVal();
       } else {
-        loop_index = IrBuilder::create<Val>(DataType::Index);
+        // Until the transition to the IdModel-based indexing is
```
naoyam, line 373: This is necessary for now since loops are generated using the ComputeAtMap indices.

naoyam:
!build

csrc/device_lower/lower2device.cpp
```diff
@@ -395,19 +395,20 @@ void GpuLower::analysis(Fusion* fusion) {
...
-    IdModel id_model(
+    id_model_ = std::make_unique<IdModel>(
         fusion_,
         /*build_graphs=*/true,
         /*allow_self_mapping=*/false,
         /*validate=*/false);
 #else
     // DEBUG build
-    IdModel id_model(
+    id_model_ = std::make_unique<IdModel>(
```
jacobhinkle, line 405: Since we always build the IdModel now, would it make sense to initialize it in the `GpuLower` constructor instead of using a `unique_ptr`?

csrc/options.cpp
```diff
@@ -261,6 +261,15 @@ const std::vector<std::string>& getDebugDumpArguments(DebugDumpOption option) {
   return DebugDumpOptionsGuard::getCurOptions().getArgs(option);
 }
 
+bool hasDebugDumpArgument(DebugDumpOption option, const std::string& arg) {
```
jacobhinkle, line 264: This could be a new method `bool Options::hasArgument(OptionEnum option, const std::string& arg) const` and we can call it here as
```
bool hasDebugDumpArgument(DebugDumpOption option, const std::string& arg) {
  return DebugDumpOptionsGuard::getCurOptions().hasArgument(option, arg);
}
```

jacobhinkle (approved):
LGTM

csrc/device_lower/lower2device.cpp
naoyam: Could you give more details? This `analysis` function is called from the constructor.

naoyam:
!build

csrc/device_lower/lower2device.cpp
jacobhinkle: I meant we could have `IdModel id_model_;` as a member instead of unique_ptr since it doesn't ever need to hold a `nullptr`. We would initialize it in the member initializer list
```c++

GpuLower::GpuLower(Fusion* fusion, const CompileParams& cparams = CompileParams())
    : passes_(/*...*/), cparams_(cparams), id_model_(fusion) {
  // ...
}
```
Is this not possible because we need to create a new copy of the fusion when we create `kernel_`?

naoyam: I think that's possible too. Maybe one reason to prefer the current style is that there are multiple analyses being done in `analysis` and there are some dependencies between them. Putting everything into the constructor initializer may make the code a little uglier? 

At least right now since we also want to make it possible to disable building an IdModel, I think the current style makes more sense.

jacobhinkle: Thanks. I agree let's keep it as is.

naoyam:
There's a real test error:

```
00:35:15 [ RUN      ] IndexingTest.SimplePointwise1
[253e4e27a783:680  :0:680] Caught signal 11 (Segmentation fault: address not mapped to object at address 0x500000047060)
==== backtrace (tid:    680) ====
 1 0x000000000056d8ab nvfuser::IdModel::idGraph()  :0
 2 0x000000000057f995 nvfuser::TensorIndexer::getLoopIndex()  :0
 3 0x0000000000cadd40 nvfuser::(anonymous namespace)::getLoopIndices()  test_indexing.cpp:0
 4 0x0000000000caea85 nvfuser::IndexingTest_SimplePointwise1_Test::TestBody()::GetReference::getLinearIndex()  test_indexing.cpp:0
 5 0x0000000000cb8c08 nvfuser::(anonymous namespace)::IndexValidator<nvfuser::IndexingTest_SimplePointwise1_Test::TestBody()::GetReference>::validate()  test_indexing.cpp:0
```

I was able to reproduce the same error with both gcc and clang for the Release mode. This doesn't seem to happen with the Debug mode.

Here's the stack trace with debug info:

```
(gdb) bt
#0  nvfuser::IdModel::idGraph (this=0xfc0854830c48348, mode=mode@entry=nvfuser::IdMappingMode::LOOP) at /home/nmaruyama/nvfuser/debug2/csrc/id_model/id_model.cpp:152
#1  0x00005555559f42f9 in nvfuser::TensorIndexer::getLoopIndex (this=this@entry=0x7ffff7fce7e9, loop_id=0x555559dcd150) at /usr/include/c++/11/bits/shared_ptr_base.h:1295
#2  0x000055555609c780 in nvfuser::(anonymous namespace)::getLoopIndices (tv=tv@entry=0x555559dceba0, indexer=...)
    at /home/nmaruyama/nvfuser/debug2/tests/cpp/test_indexing.cpp:37
#3  0x000055555609d3bb in GetReference::getLinearIndex (this=<optimized out>, tv=0x555559dd08b0, maybe_consumer=0x555559dceba0)
    at /home/nmaruyama/nvfuser/debug2/tests/cpp/test_indexing.cpp:232
#4  0x00005555560a84dc in nvfuser::(anonymous namespace)::IndexValidator<nvfuser::IndexingTest_SimplePointwise1_Test::TestBody()::GetReference>::validate(nvfuser::kir::TensorIndex *, nvfuser::kir::TensorIndex *) (this=this@entry=0x7fffffffd380, ti=ti@entry=0x555559dfcf90, out_ti=out_ti@entry=0x555559dfe900)
    at /home/nmaruyama/nvfuser/debug2/tests/cpp/test_indexing.cpp:143
```

It seems that when a reference of `IdModel` is accessed through a `TensorIndexer` that originates [here](https://github.com/NVIDIA/Fuser/pull/2389/files#diff-4ce155b5873f7974a9bb52817ba0d401172eb8520092000370c0938daf9eb79cR37), the `IdModel` reference doesn't seem to be valid anymore. The memory location of the `IdModel` object is invalid according to gdb.

After many trials and errors, looks like this change removes the error:

```
diff --git a/tests/cpp/test_indexing.cpp b/tests/cpp/test_indexing.cpp
index 7e1b3594..c7d50516 100644
--- a/tests/cpp/test_indexing.cpp
+++ b/tests/cpp/test_indexing.cpp
@@ -181,8 +181,9 @@ class IndexValidator : public kir::IrVisitor {
     kernel = lower.run();
     testing::internal::GetCapturedStderr();

+    GetReference get_ref(lower.tensorIndexer());
     IndexValidator<GetReference> validator(
-        lower, GetReference(lower.tensorIndexer()));
+        lower, get_ref);
```

I still don't understand why this fixes the error. 🤷

naoyam:
I think that's because the lifetime of `GetReference` is not extended when its const reference is assigned to the class member of `IndexValidator`. https://stackoverflow.com/a/2784304

naoyam:
!build

jacobhinkle:
> I think that's because the lifetime of `GetReference` is not extended when its const reference is assigned to the class member of `IndexValidator`. https://stackoverflow.com/a/2784304

Huh. TIL :smile:

PR #2391, opened 2024-06-12
Add threadfence to grid_sync::blockSerializeRelease

jacobhinkle:
Prior to this, we could potentially have a data race during split-K reduction. This missing fence is somewhat similar to that recently reported in https://github.com/pytorch/pytorch/pull/128455. Our `gridReduce` contains the proper fence already (inside `grid_sync::sync`), so this vulnerability only impacts split-K.

### DIFF
--- a/runtime/grid_sync.cu
+++ b/runtime/grid_sync.cu
@@ -222,6 +222,24 @@ __device__ void blockSerializeRelease(int64_t* semaphore) {
 
   // Block until writes from all threads in this block are visible to all other
   // blocks before releasing semaphore using thread 0.
+  //
+  // Consider this simple example using two blocks:
+  //
+  //   1. Block 1 acquires lock using blockSerializeWait
+  //   2. Block 1 writes values to tensor T3
+  //   3. Block 1 releases lock using blockSerializeRelease
+  //   4. Block 2 acquires lock using blockSerializeWait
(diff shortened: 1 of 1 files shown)


jacobhinkle:
!build

runtime/grid_sync.cu
```diff
@@ -220,7 +220,11 @@ __device__ void blockSerializeRelease(int64_t* semaphore) {
       index_utils::maskedOffset<X_BLOCK, Y_BLOCK, Z_BLOCK>(blockIdx, gridDim);
   bool last_block = block_idx_in_segment == segment_size - 1;
 
+  // Block until writes from all threads in this block are visible to all other
+  // blocks before releasing semaphore using thread 0.
+  __threadfence();
```
naoyam, line 225: Could you explain why this is necessary? An example would be appreciated.

jacobhinkle: Added example in comment.

naoyam: Thanks. Makes sense.

Is __syncthreads still necessary?

jacobhinkle: Yes they are both necessary because __threadfence does not synchronize the threads. It only causes that thread to wait until all its writes are visible globally. We need to sync so that all the threads finish their fences before thread 0 updates the semaphore.

naoyam (approved):
LGTM

PR #2392, opened 2024-06-12
Avoid small denominator in pytest division samples

jacobhinkle:
This PR is meant to fix the errors like this which have plagued our CI for a while now.
```
 ### [jit_python_tests_20_V100] tests.python.pytest_ops.test_correctness_div_complex64

...

E       Mismatched elements: 1 / 1048576 (0.0%)
E       Greatest absolute difference: 0.002183660166338086 at index (785, 78) (up to 0.001 allowed)
```
The problem is that we generate samples uniformly in the range [-9, 9) for both numerator and denominator. We exclude zero itself (`exclude_zero=True`) and we don't add specific small values (`enable_small_value_testing=False`) but we don't currently _exclude_ small sampled values other than zero. This PR just excludes small values (only in the denominator) and uses this sample generator for div and truediv tests that have floating point dtypes.

### DIFF
--- a/tests/python/pytest_input_generators.py
+++ b/tests/python/pytest_input_generators.py
@@ -1574,3 +1574,34 @@ def linear_error_generator(
         yield SampleInput(
             make_arg(shape_input), make_arg(shape_weight), make_arg(shape_bias)
         ), ex_type, ex_str
+
+
+def div_input_generator(
+    op: OpInfo,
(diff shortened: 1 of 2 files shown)


jacobhinkle:
!build

jacobhinkle:
!build

rdspring1 (approved):
LGTM

PR #2394, opened 2024-06-12
Map non-size-one broadcast domains in the Almost Exact graph

### LINKED ISSUE #2393: Almost Exact graph does not map non-size-one broadcast domains
One could say this would be due to creating broadcast domains with non size one, but if a broadcast domain is, for example, split by 4, it would generate two broadcast domains, `b{ceilDiv(1, 4)}` and `b{4}`, both of which do not "appear" to have size one (the former could be simplified to 1, though).

These broadcast domains do not join the almost-exact mappings:

https://github.com/NVIDIA/Fuser/blob/main/csrc/compute_at_map.cpp#L718
https://github.com/NVIDIA/Fuser/blob/main/csrc/id_model/id_model.cpp#L335

That is, if a non-broadcast domain is merged with, e.g., `b{4}`, the merge output would not be mapped with the input non-broadcast domain. 

One of the main current use cases of the almost-exact graph is indexing, where mapping broadcast domains, regardless of extents, do not matter, so it makes sense to handle such non-size-one broadcast domains in the same way as size-one broadcast domains.

naoyam:
Fixes #2393 

Initially modified `ComputeAtMap` but it resulted in lots of test failures. I don't think it's worthwhile investigating them, so leave it as is.


naoyam:
!build

naoyam:
!build --diff

naoyam:
!build --diff

naoyam:
!build --diff

jacobhinkle (approved):
The change seems safe to me. I have a question that would apply before this PR also though: what about expanded broadcasts?  When we split or merge an expanded domain we might have an expanded inner extent that is not 1, but its `extent()` would always be 1.

naoyam:
> The change seems safe to me. I have a question that would apply before this PR also though: what about expanded broadcasts? When we split or merge an expanded domain we might have an expanded inner extent that is not 1, but its `extent()` would always be 1.

Good question. I'm not that confident yet, but for indexing, is there anything that we would need to do differently than non-expanded broadcasts?

PR #2397, opened 2024-06-12
Modify `matmul` API to always create `MatmulOp`

Priya2698:
Removing special handling for some cases in `MatmulOp` to avoid any performance regressions as compared to ATen, particularly for gemmv cases.

### DIFF
--- a/csrc/ops/composite.cpp
+++ b/csrc/ops/composite.cpp
@@ -388,38 +388,7 @@ TensorView* matmul(TensorView* tv_a, TensorView* tv_b) {
       " and ",
       tv_b->dtype());
 
-  // Check for K=1 i.e. reduction of broadcast. In these cases we don't need a
-  // matmul so we translate it to a multiplication+cast
-  auto b_k_axis = tv_b->nDims() == 1 ? -1 : -2;
-  NVF_CHECK(
-      tv_a->axis(-1)->isBroadcast() == tv_b->axis(b_k_axis)->isBroadcast(),
-      "K dimension must be broadcast in both operands or none");
-  if (tv_a->axis(-1)->isBroadcast()) {
-    TensorView* float_result = nullptr;
-    if (tv_a->nDims() == 1 && tv_b->nDims() == 1) {
-      // [1] @ [1] = []
-      float_result =
-          mul(squeeze(tv_a, std::vector<int64_t>{0}),
-              squeeze(tv_b, std::vector<int64_t>{0}));
-    } else if (tv_a->nDims() == 1) {
-      // [1] @ [..., 1, N] = [..., N]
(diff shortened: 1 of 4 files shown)


Priya2698:
!build

csrc/ops/composite.cpp
```diff
@@ -389,38 +389,7 @@ TensorView* matmul(TensorView* tv_a, TensorView* tv_b) {
       " and ",
       tv_b->dtype());
 
-  // Check for K=1 i.e. reduction of broadcast. In these cases we don't need a
```
jacobhinkle, line 392: So we are actually slower than aten for this case? I wonder what their kernel looks like and how we could improve the pointwise scheduler in these simple broadcast cases.

Priya2698: No. 
The dot product is somewhat tricky, nvFuser is a little faster (around 2% for the test cases I tried) for very large K (order of 10^8) and slower for smaller K (order of 10^5). For smaller K, the kernel time is under 10us, and the difference between eager and nvFuser can be upto 25%.

In an earlier discussion, we decided to use `MatmulOp` ubiquitously, since these are corner cases and it would be easier to debug any performance regressions if we have use ATen for all cases.

We can revisit this, if you think we should handle this case separately.
CC: @kevinstephano

Priya2698: Agreed. So far, the models are not using matmul for these cases afaik. If such a case appears, then we could potentially use `fuse_matmul` option here to enable translation.

jacobhinkle: > If such a case appears, then we could potentially use `fuse_matmul` option here to enable translation.

That's not so straightforward though. We would be accepting the segment with the Matmul scheduler then, when we would need it to be accepted by something else like pointwise, transpose, reduction, etc. This is why it's better to translate as early as possible so that we can properly segment. However, as you say, until we find this use case in the wild we should probably not spend too much energy optimizing it. It is, after all, a translation that can also happen further up the stack like in thunder or could be done by the user.

Priya2698: Yes, we cannot do it after segmentation. I was proposing doing that right here, when we are creating a `MatmulOp`. So same as the current special cases guarded by this enable option.

However, for now, let us move with removing these cases and reevaluate when a relevant fusion comes up.

jacobhinkle: > I was proposing doing that right here, when we are creating a `MatmulOp`

I am not sure I understand how that'd be different from the current behavior. The `fuse_matmul` option shouldn't affect whether we do that translation since there would no longer be a `MatmulPattern` in the fusion if you skip the `MatmulOp` for this case. Anyway, as you say let's proceed with removing it :grin: .

Priya2698:
!build

jacobhinkle (approved):

PR #2401, opened 2024-06-13
Comms/compute Overlap with RS-Based pipelining: ATen implementation

samnordmann:
# What

This PR provides an ATen implementation of "GEMM+Reduce-scatter" operation using a reduce-scattered based pipelining overlapping technique, as used in NEMO-Megatron transformer, precisely at the second layer of the MLP.

# Configurations

We provide a way to fully configure the parameters used in the test through environment variables. The available options are:
- The dimensions `M,K,N` of the matrices
- The number of interleaved chunks `S`
- Whether to use several streams
- which communication backend to use, NCCL or UCC

# Detailed walkthrough

The tensor program that we target is the following, assuming a setup with `num_devices` devices:
   - inputs:
       - `A[M,K]` sharded column-wise: dimension `K` is split by the factor `num_devices`, so `A` is viewed as `[M, num_devices, K/num_devices]`, and the allocation size of `A` is `[M, 1, K/num_devices]`
       - `B[K,N]` sharded row-wise: locally of size `[1, K/num_devices, N]`
   - output:
       - `C[M,N]=matmul(A,B)`, sharded on dimension `M`: dimension `M` is split by the factor `num_devices`, so `C` is viewed as `[num_devices, M/num_devices,N]`, and the allocation size of `M` is `[1, M/num_devices,N]`

Up to some broadcast and view ops, a straightforward program to generate the output could be summarized as
```
C_unreduced = pointwise_multiply(A,B)
C_locally_reduce = reduction(C_unreduced, axis=`K/num_devices`, op=sum)
C = reduce_scatter(C_unreduced, op=sum)
```
where `C_unreduced`, `C_locally_reduce`, `C` are all sharded on `num_devices` and :
- `C_unreduced` has unsharded size `[M,num_devices,K/num_devices,N]`,
- `C_locally_reduce` has unsharded size `[M,num_devices,N]`,
- `C` has unsharded size `[num_devices, M/num_devices, N]`

We want to compare this baseline program with one that is functionally identical but achieves more overlap between computations and communications. Our goal is to interlave the comms and compute using a technic called "reduce-scatter based pipelining". To do so, we further split the row dimension `M` with a factor `S` and we apply the comms and compute described earlier in parallel (i.e. on different streams) on each tensors slices accross `S`. More precisely, assuming the following shapes:
    - `A [S, num_devices, M/S, K/num_devices]`, sharded on `num_devices`
    - `B [num_devices, K/num_devices, N]`, sharded on `num_devices`
    - `C [S, num_devices, M/(S*num_devices), N]`, sharded on `num_devices`
the program achieving comms/compute overlap using rs-based pipelining could be summarized as:
```
for (j=0; j<S; j++) {
    setCurrentStream(Stream[j])
    C_unreduced[j] = pointwise_multiply(A[j],B)
    C_locally_reduce[j] = local_reduction(C_unreduced[j], axis=`K/num_devices`, op=sum)
    C[j]=reduce_scatter(C_locally_reduce[j], op=sum)
}
```
where `[j]` refers to taking a slice onto the `S` dimension.

_Remarks:_
  1) it is convenient to have `S` as being the outermost dimension so `C_locally_reduce[j]` is a contiguous buffer, and can therefore be reduce-scattered with a single MPI collective.
  2) The layout needs to match the reduce-scatter semantics, i.e., the first dimension is reduced and the second is scattered. This is why we choose the layouts to be`[S, sharded_axis, M / (...) , ...]`

# NSight profile analysis

### Setup 
We provide a screenshot of the night profile, obtained with the following setup:
- 8 GPUs V100, on one DGX node
- M=N=K=2^10
- S=8
- Multi stream activated
- backend: nccl

### Profile analysis
Comments:
- We can see on the illustration the three types of kernels:
   - nccl Kernels for reduce_scatter
   - pointwise kernels (for the pointwise multiply)
   - reduce kernels (for the local sum reduction)
- (Circle1)The first kernel we see on the screenshot is the first reduce-scatter. The first compute, happening before reduce-scatter, cannot be seen on the screenshot because of the nccl init overhead. Also, the nccl init is blocking, which explains why the other compute are not hidden by nccl init.
- (Circle2) I _think_  that ATen compute takes all the SMs and therefore nccl Kernel is enqueued, hence the non-overlap.
- (Circle3) During reduce-scatter, ATen compute overlaps. We see comm/compute and compute/compute overlap.
- (Circle4) For the same reason as in Circle 2, nccl kernels are enqueued to after the compute is finished
- (Circle4) The comms do not overlap one with another because PG has only one internal stream. Upcoming optimization: instantiating multiple PG backends to achieve comm/comm overlap. Dependency with https://github.com/NVIDIA/Fuser/pull/2402 and an upcoming PR implementing a dedicated Host IR
Screenshot 2024-06-13 at 17 07 17


samnordmann:
!build

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -0,0 +1,283 @@
...
+    // validating data correctness.
+    at::TensorOptions options =
+        at::TensorOptions().dtype(at::kFloat).device(communicator->device());
+    auto ta_unsharded = at::randn(ta_unsharded_sizes, options);
+    auto tb_unsharded = at::randn(tb_unsharded_sizes, options);
+    ta = at::empty(ta_sizes, options);
+    ta.copy_(ta_unsharded.index(
+        {at::indexing::Slice(),
+         at::indexing::Slice(my_device_index, my_device_index + 1),
+         "..."}));
```
wujingyue, line 138: How is it different from `ta_unsharded.slice(1, my_device_index, my_device_index + 1)`, which is simpler?

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -0,0 +1,283 @@
...
+  }
+
+  void computeATen(
+      at::Tensor ta_j,
+      at::Tensor tb_j,
+      at::Tensor tc_unreduced_j,
+      at::Tensor tc_locally_reduced_j) {
+    at::mul_out(tc_unreduced_j, ta_j, tb_j);
+    at::sum_out(tc_locally_reduced_j, tc_unreduced_j, {3});
+  }
```
wujingyue, line 186: Seems like an einsum, which can save tc_unreduced_j, an intermediate tensor.

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -0,0 +1,283 @@
...
+};
+// clang-format off
```
wujingyue, line 188: Thanks for the detailed comment!

wujingyue (commented):
Nice work! I'm sending out the first batch of comments; it would take too long to finish all.

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -0,0 +1,283 @@
...
+  }
+
+  // synchronize default stream with all other streams
+  setCurrentCUDAStream(c10::cuda::getDefaultCUDAStream());
+  for (auto stream : streams) {
+    stream.synchronize();
+  }
+
+  // validation
+  ASSERT_TRUE(tc.allclose(tc_expected, 1e-3, 1e-3))
```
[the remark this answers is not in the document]
samnordmann: right, let me fix

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -0,0 +1,283 @@
+// clang-format off
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2023-present NVIDIA CORPORATION & AFFILIATES.
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+// clang-format on
+#ifdef NVFUSER_DISTRIBUTED
```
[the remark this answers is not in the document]
samnordmann: no problem, I will change the guards

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -0,0 +1,283 @@
...
+    }
+    if (isEnvVariableDefined("LOG2_K")) {
+      K = std::pow(2, parseEnvVariable("LOG2_K"));
+    }
+    if (isEnvVariableDefined("LOG2_N")) {
+      N = std::pow(2, parseEnvVariable("LOG2_N"));
+    }
+    if (isEnvVariableDefined("LOG2_S")) {
+      S = std::pow(2, parseEnvVariable("LOG2_S"));
+    }
```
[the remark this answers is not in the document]
samnordmann: The goal is to be able to set up the parameters from the commandline without the need for recompilation. At this experimental stage, it is really convernient to be able to quickly set the params and loop over them to run experiments and analyze the profile. The same framework will be used for other tests, e.g., overlap implemented through host IRs, and p2p ring algos.

So, keeping a mechanism to set the params at runtime would greatly help me. In the future, once the tests are consolidated, we could select relevant params value and hardcode them into GTEST suite.

About "how" to achieve what I want, parsing env variable like that felt natural, mimicking what is done for other NVFUSER options, including in the test, e.g., https://github.com/NVIDIA/Fuser/blob/9381e6f48f0663faf6d189412247f8336429471f/tests/cpp/test_tutorial.cpp#L27
But let me know if you prefer another implementation.

tests/cpp/test_multidevice_overlap.cpp
samnordmann: let me fix that

tests/cpp/test_multidevice_overlap.cpp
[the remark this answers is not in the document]
wujingyue: I now understand that you want to test parameters beyond those instantiated. 

The most natural way is to create a new binary (e.g. using add_executable in cmake) with its own `main` function, which parses command line arguments using `getopt`. Env vars don't protect you from typos. For example, if I accidentally mistyped the name of an env var in the command line, there wouldn't be any warnings and the program likely just quietly uses the default value. 

If you plan to also test a list of predefined parameters, you can create a test in addition to the binary. Also, you can make a test with a custom `main` function, which decides whether to `RUN_ALL_TESTS` or run a user-specified parameter via command line arguments. FWIW, I sent you https://github.com/NVIDIA/Fuser/pull/2438 as an example of using a custom `main` function.

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -0,0 +1,283 @@
...
+
+class OverlapTest : public MultiDeviceTest {
+ protected:
+  OverlapTestParams params;
+
+  int64_t num_devices;
+  int64_t my_device_index;
+  at::Tensor ta, tb, tc_unreduced, tc_locally_reduced, tc, tc_expected;
+  // stores the backend
+  c10d::Backend* world_communicator;
```
wujingyue, line 97: Style suggestion: I'd put an _ after these names. This way, it's easy for readers to distinguish member variables and local variables. It has helped me a lot in the past.

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -0,0 +1,283 @@
...
+    num_devices = communicator->size();
+    my_device_index = communicator->deviceId();
+    ASSERT_EQ(params.M % (params.S * num_devices), 0);
+    ASSERT_EQ(params.K % num_devices, 0);
+
+    // Setup the world communicators
+    std::vector<int64_t> devices(num_devices);
+    std::iota(devices.begin(), devices.end(), 0);
+    world_communicator = communicator->getBackendForTeam(
+        devices, /*backend=*/params.backend_type);
```
wujingyue, line 112:
```suggestion
        devices, params.backend_type);
```

The name `backend_type` is pretty clear about which argument of getBackendForTeam it corresponds to.

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -0,0 +1,283 @@
...
+  for (auto j : c10::irange(params.S)) {
+    // define the sliced tensors
+    auto ta_j = getSlice(ta, j);
+    auto tc_unreduced_j = getSlice(tc_unreduced, j);
+    auto tc_locally_reduced_j = getSlice(tc_locally_reduced, j);
+    auto tc_j = getSlice(tc, j);
+
+    if (params.use_different_streams) {
+      auto new_stream = c10::cuda::getStreamFromPool(
+          /*isHighPriority=*/true, my_device_index);
```
wujingyue, line 257: Can you comment on the high-priority part?

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -0,0 +1,283 @@
...
+      N = std::pow(2, parseEnvVariable("LOG2_N"));
+    }
+    if (isEnvVariableDefined("LOG2_S")) {
+      S = std::pow(2, parseEnvVariable("LOG2_S"));
+    }
+    if (isEnvVariableDefined("USE_UCC")) {
+      backend_type = CommunicatorBackend::ucc;
+    }
+    if (isEnvVariableDefined("USE_STREAMS")) {
+      use_different_streams = true;
```
wujingyue, line 67: IIUC, "use different streams" is not just "use streams" but also use a different stream per iteration. So I'd reflect that by changing the flag to USE_DIFFERENT_STREAMS to be consistent.

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -0,0 +1,283 @@
...
+    auto tc_unsharded_expected_reshaped = at::reshape(
+        tc_unsharded_expected,
+        {params.S, num_devices, params.M / (params.S * num_devices), params.N});
+    tc_expected = tc_unsharded_expected_reshaped.index(
+        {at::indexing::Slice(),
+         at::indexing::Slice(my_device_index, my_device_index + 1),
+         "..."});
+
+    // Debug print
+    if (!communicator->deviceId() && params.debug_print) {
```
wujingyue, line 164:
```suggestion
    if (communicator->deviceId() == 0 && params.debug_print) {
```
Nit: I tend to avoid using boolean operations on integers directly. It may be less typing but makes the program a bit hard to follow (like 10 seconds in this case for me wondering the type of DeviceId).

tests/cpp/test_multidevice_overlap.cpp
samnordmann: done, well spotted!

tests/cpp/test_multidevice_overlap.cpp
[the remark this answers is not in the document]
samnordmann: Your suggestion makes sense, thank you also for the pointer. I think that adding a new binary is not worth the effort for now, at least in this PR. So I removed parsing the env variables, and the test is just run with default parameters values. I'll see later if I need a new binary indeed.

samnordmann:
!build

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -0,0 +1,242 @@
...
+          /*isHighPriority=*/true, my_device_index_);
+      streams.push_back(new_stream);
+      setCurrentCUDAStream(new_stream);
+    }
+
+    // local compute
+    computeATen(ta_j, tb_, tc_locally_reduced_j);
+    // communication
+    world_communicator_->_reduce_scatter_base(tc_j, tc_locally_reduced_j)
+        ->wait();
```
naoyam, line 227: Is this `wait` necessary?

samnordmann: Yes, it is always necessary to `wait()` after a communication is posted. However, note here that we change stream at each iteration, and so `wait()` is the last kernel posted on the stream. Therfore, the synchronization will occur only after all the kernels and comms are executed, when we synchronize the stream with the main stream, just before the validation.

tests/cpp/test_multidevice_overlap.cpp
samnordmann: To give more details: 
- CUDA provides a notion of stream priority: https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__STREAM.html#group__CUDART__STREAM_1ge2be9e9858849bf62ba4a8b66d1c3540
- Torch have two pools of streams, one with "high priority", the other with "low priority". https://pytorch.org/cppdocs/notes/tensor_cuda_stream.html
- in Fuser, for now, it is not clear how we could leverage the stream priorities, so we decide in the meantime to only use high priority streams.

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+    // Setup the world communicators
+    std::vector<int64_t> devices(num_devices_);
+    std::iota(devices.begin(), devices.end(), 0);
+    world_communicator_ =
+        communicator->getBackendForTeam(devices, params.backend_type);
+
+    // Define I/O and intermediate Tensor shapes
+    // clang-format off
+        std::vector<int64_t> ta_unsharded_sizes       = {params.S, num_devices_, params.M/params.S, params.K/num_devices_, 1       };
```
wujingyue, line 73:
```suggestion
    std::vector<int64_t> ta_unsharded_sizes       = {params.S, num_devices_, params.M/params.S, params.K/num_devices_, 1       };
```

wujingyue: Nit: I don't see much value for `clang-format off`. You probably wanted to align commas, but personally I don't find it as annoying as having to indent manually :)

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -0,0 +1,242 @@
...
+    // expected result locally, hence, this way of doing is convenient for
+    // validating data correctness.
+    at::TensorOptions options =
+        at::TensorOptions().dtype(at::kFloat).device(communicator->device());
+    auto ta_unsharded = at::randn(ta_unsharded_sizes, options);
+    auto tb_unsharded = at::randn(tb_unsharded_sizes, options);
+    ta_ = at::empty(ta_sizes, options);
+    ta_.copy_(ta_unsharded.slice(1, my_device_index_, my_device_index_ + 1));
+    tb_ = at::empty(tb_sizes, options);
+    tb_.copy_(tb_unsharded.slice(1, my_device_index_, my_device_index_ + 1));
```
wujingyue, line 93: Nit: could use getSlice to simplify the code a little bit.

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+  at::Tensor getSlice(at::Tensor t, int64_t j) {
+    return t.slice(0, j, j + 1);
+  }
+
+  void computeATen(
+      at::Tensor ta_j,
+      at::Tensor tb_j,
+      at::Tensor tc_locally_reduced_j) {
+    // we unsqueeze the output tensor to avoid a torch warning:
```
wujingyue, line 132: Note to myself to try out later: I'm surprised this can't be avoided. Maybe tensordot is not as powerful as einsum.

wujingyue (commented):
Almost LGTM. I left myself a TODO to figure out a better way to dot product.

tests/cpp/test_multidevice_overlap.cpp
samnordmann: ok, let me remove it then

tests/cpp/test_multidevice_overlap.cpp
samnordmann: Ok. It works but gives a warning.
Lmk

tests/cpp/test_multidevice_overlap.cpp
samnordmann: getSlice operates on axis 0, that's why I'm not using it here. But let me add an "axis" argument to getSlice, so I can use it everywhere

tests/cpp/test_multidevice_overlap.cpp
wujingyue: Nit: I tend to use default unless the non-default value is proven to be important. In this case, the benefit of using a high-priority stream seems unclear.

samnordmann: alright, that makes sense. Let me fix it

tests/cpp/test_multidevice_overlap.cpp
wujingyue: I pushed a change with an explanatory commit message. It's suboptimal either way so take it or leave it :)

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -0,0 +1,250 @@
...
+        getSlice(tc_unsharded_expected_reshaped, 1, my_device_index_);
+
+    // Debug print
+    if (communicator->deviceId() == 0 && debug_print) {
+      std::cout << params << std::endl;
+      std::cout << "ta_.sizes()=" << ta_.sizes() << std::endl;
+      std::cout << "tb_.sizes()=" << tb_.sizes() << std::endl;
+      std::cout << "tc_locally_reduced_.sizes()=" << tc_locally_reduced_.sizes()
+                << std::endl;
+      std::cout << "tc_.sizes()=" << tc_.sizes() << std::endl;
```
wujingyue, line 128: Debug prints in general go to std::cerr.

wujingyue (approved):

tests/cpp/test_multidevice_overlap.cpp
samnordmann: took it! thanks

tests/cpp/test_multidevice_overlap.cpp
samnordmann: ok, I change to using `debug()`

samnordmann:
!build

tests/cpp/test_multidevice_overlap.cpp
samnordmann: fixed in https://github.com/NVIDIA/Fuser/pull/2475

PR #2402, opened 2024-06-13
Add prefix option to the communicator's backend cache

samnordmann:
# What
Adds a fine grain control of the Communicator's backend cache

# Why
The `Communicator` is used as a singleton instance managing the different backends for communication. It internally holds a cache of the previously created backend to avoid the costly initial overhead of ProcessGroup creation.

One the one hand, it is important that this caching occurs at the Communicator's level and not at the caller site, because the communicator exists globally accross test instances.

On the other hand, in order to achieve communication/communication overlap, one must use multiple backends because each backend has one and only one internal stream on which it posts kernels.

# How
We add an optional `std::string prefix=""` argument to the function `Communicator:: getBackendForTeam` that is prepended to the key used to retrieve a backend from the cache. This allows arbitrary control of the cache from the caller.

### DIFF
--- a/csrc/multidevice/communicator.h
+++ b/csrc/multidevice/communicator.h
@@ -87,8 +87,8 @@ class Communicator {
 
   // returns the backend associated with a team
   // the argument "prefix" is prepended to the key used to retrieve preexisting
-  // backends. This allow us to have a fine-grained control on how the backends
-  // are cached.
+  // backends. Prefix is used to distinguish between different backends with the
+  // same team
   c10d::Backend* getBackendForTeam(
       const Team& team,
       std::optional<CommunicatorBackend> backend,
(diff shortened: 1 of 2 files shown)


csrc/multidevice/communicator.h
```diff
@@ -86,9 +86,13 @@ class Communicator {
   }
 
   // returns the backend associated with a team
+  // the argument "prefix" is prepended to the key used to retrieve preexisting
+  // backends. This allow us to have a fine-grained control on how the backends
+  // are cached.
```
cowanmeg, line 91: nit: prefix is used to distinguish between different backends with the same team.

cowanmeg (approved):
LGTM! Just a minor suggestion on the comment.

csrc/multidevice/communicator.h
```diff
@@ -86,9 +86,13 @@ class Communicator {
...
   c10d::Backend* getBackendForTeam(
       const Team& team,
-      std::optional<CommunicatorBackend> backend);
+      std::optional<CommunicatorBackend> backend,
+      std::string prefix = "");
```
wujingyue, line 95:
```suggestion
      const std::string& prefix = "");
```
to reduce the number of copies if the caller passed in a const ref.

wujingyue (approved):

PR #2399, opened 2024-06-12
Revert "Host irs: add wait IR (#2338)"

wujingyue:
This reverts commit c33e84167df90b51167895098b05abef41af6932.

The commit causes CI failures, which can be reproduced by running

```
_bn && mpirun -np 4 bin/test_multidevice
--gtest_filter=Manual/MultiDeviceHostIrTest.SingleFusionSingleComm/useFusionExecutor*
```

PR #2338, opened 2024-06-04
Host irs: add wait IR

samnordmann:
add an equivalent to `c10d::Work::Wait` primitive as a host IR, to allow asynchronous execution of communications

Assuming https://github.com/NVIDIA/Fuser/pull/2384 is merged, the following illustration proves that we achieve overlap:
Screenshot 2024-06-11 at 19 12 21

This illustration is generated from the test unit `Manual/MultiDeviceHostIrTest.SingleCommTwoFusionAndWait/useFusionExecutorCache_withShardingAnnotations` with 8 devices. In this test, we run the following comms, wait and compute primitive, with no data depedency:
```
compute
communication
wait
compute
communication
compute
wait
compute
```
The first `compute` `communication/wait` triggers some initialization overhead and are not representative. The profile screenshot displays on the last operations: communication, compute, wait, compute, and we can see overlap between the comm and the first compute, while the second compute is defered to after the communication's completion. The behavior is therefore as expected.

PR #2403, opened 2024-06-13
Host irs/add collective wait

samnordmann:
Pick up from #2338 which introduced [a bug](https://gitlab-master.nvidia.com/dl/pytorch/fuser-gh-mirror/-/jobs/97151941#L925) and has been reverted in #2399

We apply the same patch here, plus [an additional commit](https://github.com/NVIDIA/Fuser/commit/d292a5fe9f7d86860aa4f4e9e479dd809a115843) that fixes the bug. The fix is to add a "wait" primitive after the collective is posted in the test `MultiDeviceHostIrTest.SingleFusionSingleComm`.

### DIFF
--- a/csrc/dispatch.h
+++ b/csrc/dispatch.h
@@ -140,7 +140,8 @@ class Val;
 #define DISPATCH_FOR_ALL_HIR_EXPRS(f) \
   f(HostUnit);                        \
   f(PostOnStream);                    \
-  f(SetCurrentStream);
+  f(SetCurrentStream);                \
+  f(Wait);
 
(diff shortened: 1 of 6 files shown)


samnordmann:
!build

wujingyue:
Makes sense. What's the contract? Each `Communication` needs to be followed by a `Wait`?

wujingyue (approved):

samnordmann:
> Makes sense. What's the contract? Each `Communication` needs to be followed by a `Wait`?

yes

Issue #2375, opened 2024-06-10
group norm segmented into pointwise + persistent + pointwise


liqiangxl:
Group norm is calculated as:
```
x0 = [N, C, H, W]
x1 = x0.cast(fp32).reshape(N, C, H, W)  --> (N, G, C/G, H, W) 
x2 = x1 / x1.sum(C/G, H, W)
x3 = x2.reshape(N, G, C/G, H, W) --> (N, C, H, W)
x4 = w*x3 + b
```
Due to the two reshapes, normalization scheduler rejects the unsegmented fusion and then it is segmented into three sub-fusions:
(1) pointwise doing `cast + reshape`
(2) normalization
(3) pointwise doing `reshape + scale & bias`

Reproduce (modified from [apex group norm implementation](https://github.com/NVIDIA/apex/blob/master/apex/contrib/group_norm/group_norm.py) and  [apex group norm test](https://github.com/NVIDIA/apex/blob/7b73b12361068a10b0f44844534613f252a5ea75/apex/contrib/test/group_norm/test_group_norm.py#L38) ):
```
import torch
import thunder

def torch_group_norm(x, g, w, b, eps, act=""):
    xdtype, wdtype = x.dtype, w.dtype
    if xdtype != wdtype:
        x = x.to(dtype=wdtype)
    y = torch.nn.functional.group_norm(x, g, w, b, eps)
    if act in ["silu", "swish"]:
        y = torch.nn.functional.silu(y)
    if xdtype != wdtype and y.dtype != xdtype:
        y = y.to(dtype=xdtype)
    return y

def verify_group_norm(N=32,
                      C=128,
                      H=256,
                      W=256,
                      G=32,
                      xdtype=torch.float16,
                      wdtype=torch.float32,
                      eps=1e-5,
                      memory_format=torch.channels_last,
                      device='cuda',
                      act=""):
    # create data
    x_shape = (N, C, H, W)
    w_shape = (C,)
    weight = torch.rand(w_shape,
                        dtype=wdtype,
                        device='cuda',
                        requires_grad=True)
    bias = torch.rand(w_shape,
                      dtype=wdtype,
                      device='cuda',
                      requires_grad=True)
    x = torch.randn(x_shape, dtype=xdtype, device='cuda')
    x = x.to(memory_format=memory_format)
    x.requires_grad_(True)
    thunder_group_norm = thunder.jit(torch_group_norm)
[...]

Issue #2395, opened 2024-06-12
RuntimeError: producer->getMemoryType() == MemoryType::Global


wujingyue:
This blocks https://github.com/Lightning-AI/lightning-thunder/issues/191. 

# nvFuser reproducer

Check out https://github.com/NVIDIA/Fuser/tree/wjy/memory and run `python repro.py`. 

```py
import torch
from nvfuser import FusionDefinition, DataType

def nvfuser_fusion_id9(fd : FusionDefinition) -> None :
    T0 = fd.define_tensor(shape=[-1, -1, -1], contiguity=[True, True, None], dtype=DataType.Bool, is_cpu=False, stride_order=[2, 1, 0])
    T1 = fd.define_tensor(shape=[-1, -1, -1], contiguity=[True, True, True], dtype=DataType.Float, is_cpu=False, stride_order=[2, 1, 0])
    T2 = fd.define_tensor(shape=[-1, -1], contiguity=[True, True], dtype=DataType.Bool, is_cpu=False, stride_order=[1, 0])
    S3 = fd.define_scalar(16, dtype=DataType.Int)
    S4 = fd.define_scalar(16, dtype=DataType.Int)
    S5 = fd.define_scalar(1, dtype=DataType.Int)
    V6 = fd.define_vector([S3, S4, S5], dtype=DataType.Int)
    T7 = fd.ops.broadcast_in_dim(T2, shape=V6, broadcast_dims=[0, 1])
    S8 = fd.define_scalar(1, dtype=DataType.Int)
    S9 = fd.define_scalar(16, dtype=DataType.Int)
    S10 = fd.define_scalar(1, dtype=DataType.Int)
    S11 = fd.define_scalar(16, dtype=DataType.Int)
    S12 = fd.define_scalar(32, dtype=DataType.Int)
    S13 = fd.define_scalar(1, dtype=DataType.Int)
    V14 = fd.define_vector([S8, S9, S10, S11, S12, S13], dtype=DataType.Int)
    T15 = fd.ops.broadcast_in_dim(T7, shape=V14, broadcast_dims=[1, 3, 5])
    S16 = fd.define_scalar(16, dtype=DataType.Int)
    S17 = fd.define_scalar(16, dtype=DataType.Int)
    S18 = fd.define_scalar(32, dtype=DataType.Int)
    V19 = fd.define_vector([S16, S17, S18], dtype=DataType.Int)
    T20 = fd.ops.reshape(T15, new_shape=V19)
[...]

PR #2405, opened 2024-06-13
Allow output to alias intermediate tensor, step-2 to solve group norm segmentation issue #2375

liqiangxl:
This PR is step-2 to solve #2375 
Allow output to alias intermediate when:
(1) called from pre segment pass
(2) the `op` that preduces `output` interfer with reduction, see `outputInterferingReduction`

After this PR, the last reshape (which produces the ouput tensor) in group norm is changed to a no-op, see newly added test `OutputAliasIntermediate`

Util function `getRepresentativeReductionTv` is added for convenience and will be used to simipify reduction/normalization schedulers.


liqiangxl:
!build

csrc/alias_analysis.cpp
```diff
@@ -429,15 +432,21 @@ bool okToRelayout(
...
     auto [root, preferred_layout] = root_and_layout;
-    // Walks up the `alias_to_source_` chain.
-    while (root != nullptr && !root->isFusionInput() &&
-           !root->isFusionOutput()) {
-      const auto i = alias_to_source_.find(root);
-      root = (i == alias_to_source_.end() ? nullptr : i->second.first);
+    // If allow output to alias intermediate, then we don't need to walk up
+    // the chain to find an input or output root. It changes the last reshape
+    // in group norm to a no-op.
+    if (!allow_output_alias_intermediate) {
```
wujingyue, line 442: Thanks for the patch. Have you read https://github.com/NVIDIA/Fuser/issues/2395#issuecomment-2207043749? I tagged you there but the notification apparently didn't land. Would that solve this groupnorm issue?

liqiangxl: Yes and it looks great. IIUC, step-1 in the proposal is `Run alias analysis as we do today`, however, current `AliasAnalysisResult::finalize` needs to walk up the chain to find an input or output root and it will miss the case in group norm where the output alias an intermediate tensor. That's why I need to add this patch.
Previously, I also need to patch `MarkAliasesPreparePass::runPass` (see [commit](7175220caf004e979a603deb0b78f5e16643bdce) ), but with #2529, that path is no longer required.

liqiangxl:
!build

csrc/alias_analysis.cpp
liqiangxl: > superseded



> > AliasAnalysisResult::finalize needs to walk up the chain to find an input or output root and it will miss the case in group norm where the output alias an intermediate tensor.
> 
> Yes, that's an important implementation detail that needs to be reworked. `alias_to_source_` does keep aliases between intermediate tensors. So I'll have to put `segment_set` based on that without having to find the fusion I/O.
> 
> Overall, I feel lots of this PR is going to be superseded. However, if you need it in a week, I don't mind reviewing it. Wdyt?

Thanks. It would be great to have this PR merged, so we can make group norm works as expected. Assuming it won't cause troubles for your planned works to revise alias analysis.

csrc/alias_analysis.h
```diff
@@ -112,7 +114,9 @@ class AliasAnalysisResult {
 // schedulers. The former, used by MarkAliasesPreparePass, updates layouts to
 // enable aliases; the latter, used by NoOpScheduler, calls
 // Fusion::aliasOutputToInput to mark aliases.
-NVF_API AliasAnalysisResult
-findAliases(Fusion* fusion, bool can_override_empty_allocation_domain = true);
+NVF_API AliasAnalysisResult findAliases(
```
wujingyue, line 117:
```suggestion
AliasAnalysisResult findAliases(
```

csrc/scheduler/no_op.cpp
```diff
@@ -30,8 +30,10 @@ NoOpScheduler::NoOpScheduler(
 
 namespace {
 bool allOutputsArePointerArithmetics(Fusion* fusion) {
-  const AliasAnalysisResult analysis =
-      findAliases(fusion, /*can_override_empty_allocation_domain=*/false);
+  const AliasAnalysisResult analysis = findAliases(
+      fusion,
+      /*can_override_empty_allocation_domain=*/false,
+      /*can_alias_intermediate=*/false);
```
wujingyue, line 36:
```suggestion
      /*allow_output_alias_intermediate=*/false);
```

Side note: Apparently, we haven't enabled https://clang.llvm.org/extra/clang-tidy/checks/readability/named-parameter.html; otherwise you'll see a clang-tidy error. I'll try to add that to our lint config.

csrc/alias_analysis.cpp
```diff
@@ -429,15 +432,21 @@ bool okToRelayout(
...
+    const bool allow_output_alias_intermediate) {
   for (auto [alias, root_and_layout] : alias_to_source_) {
     auto [root, preferred_layout] = root_and_layout;
-    // Walks up the `alias_to_source_` chain.
-    while (root != nullptr && !root->isFusionInput() &&
-           !root->isFusionOutput()) {
-      const auto i = alias_to_source_.find(root);
-      root = (i == alias_to_source_.end() ? nullptr : i->second.first);
+    // If allow output to alias intermediate, then we don't need to walk up
+    // the chain to find an input or output root. It changes the last reshape
```
wujingyue, line 440: I'd hope it's backed by numbers, but a more reasonable approach seems to be walk up until we reach a tensor that are known to be global, e.g., when outputInterferingReduction returns true or, not in this PR, the input of slice or pad.

tests/cpp/test_gpu_view.cpp
```diff
@@ -2344,4 +2344,128 @@ TEST_F(GpuViewTest, ExpandedBroadcast) {
...
+      at::TensorOptions().dtype(data_type_to_aten(dtype)).device(at::kCUDA, 0);
+  auto t0 = at::randn(input_shape, options);
+  auto t1 = t0.to(at::kFloat);
+  auto t2 = t1.sum({-1, -2, -3}).unsqueeze(-1).unsqueeze(-1).unsqueeze(-1);
+  auto t3 = t1 / t2;
+  auto t4 = t3.reshape(output_shape);
+
+  FusionExecutorCache executor_cache(std::move(fusion));
+  auto cg_outputs = executor_cache.runFusionWithInputs({t0});
+  auto seg_groups =
```
wujingyue, line 2453:
```suggestion
  const std::vector<SegmentedGroup*>& seg_groups =
```

tests/cpp/test_gpu_view.cpp
```diff
@@ -2344,4 +2344,128 @@ TEST_F(GpuViewTest, ExpandedBroadcast) {
...
+    if (sg->heuristic() == ScheduleHeuristic::NoOp) {
+      n_no_op++;
+    } else if (sg->heuristic() == ScheduleHeuristic::InnerPersistent) {
+      n_norm++;
+    } else {
+      FAIL() << "Unexpected heuristic: " << sg->heuristic();
+    }
+  }
+  EXPECT_EQ(n_no_op, 1);
+  EXPECT_EQ(n_norm, 1);
```
wujingyue, line 2466: This can be simplified using gmock. Something like: 

```
EXPECT_THAT(groups, Contains(HeuristicIs(NoOp)).Times(1));
EXPECT_THAT(groups, Contains(HeuristicIs(InnerPersistent)).Times(1));
```

See https://github.com/NVIDIA/Fuser/blob/3c97a085d44a4f34c947efd198dda7b8afad3d56/tests/cpp/test_move_split_cat.cpp#L206 and https://github.com/NVIDIA/Fuser/blob/3c97a085d44a4f34c947efd198dda7b8afad3d56/tests/cpp/test_alias.cpp#L1091-L1093.

liqiangxl: New Knowledge Received!

csrc/scheduler/no_op.cpp
wujingyue: Weird -- I no longer understand why this didn't trigger a clang-tidy error. 

Although readability-named-paraemter is not turned on, bugprone-argument-comment is. It serves the same purpose, and indeed kicks in when I mistyped the argument name: https://github.com/NVIDIA/Fuser/pull/2626 and https://github.com/NVIDIA/Fuser/actions/runs/9982268422/job/27587668042?pr=2626.

csrc/alias_analysis.cpp
liqiangxl: A little confusion about `a tensor that are known to be global`. Here I am trying to let the output tensor alias to a tensor that is not in global memory (not input or output, but intermediate), then further code will segment this `output tensor = someOp(an intermediate tensor)` to a no-op.

Here `outputInterferingReduction` depends on the whole `fusion` not a single tensor. So I didn't move it into the walk up process. But it looks like we should also check alias in `alias_to_source_` is actually defined by a view op. Then we don't need to walk up. So the code is changed as:
(1) ` stop_at_view =  may_alias_intermediate && outputInterferingReduction(fusion)` is moved to `finalize()`
(2) only walk up if `if (!isOpsToStop(alias->definition(), stop_at_view))`

liqiangxl:
!build

liqiangxl:
!build

wujingyue (approved):

PR #2410, opened 2024-06-13
Make the `FusionProfiler` work with a kernel Executor instead of requiring segmentation

kevinstephano:
# Summary
This PR is does 3 things:
1. Adds support for reporting the scheduler type in the `FusionProfiler` which includes adding the appropriate python bindings for the new field in the `KernelProfile`.  This is for ease of categorization of benchmarks in CI.
2. It moves the kernel profiling from the Segmentation level to the Fusion Executor level of code so the `FusionExecutor` can be executed on its one as is found in tests and with User Scheduled Fusions.
3. I fixed up the User Schedule profiling in the `FusionDefinition`'s `execute()` method to execute the `FusionProfiler::start` and `FusionProfiler::stop` around the user schedule's `runFusion()`.  This required moving compilation from `FusionDefintion::finalizeDefinition` to `FusionDefinition::execute`.

# TL;DR
## How do you read the `KernelProfile.scheduler` field from python?
Here is a code example:
```python
import torch
from nvfuser import FusionDefinition

inputs = [
    torch.randn((2, 5), dtype=torch.float, device="cuda:0"),
    torch.randn((2, 5), dtype=torch.float, device="cuda:0"),
]

    T0 = fd.from_pytorch(inputs[0])
    T1 = fd.from_pytorch(inputs[1])
    T2 = fd.ops.add(T0, T1)
    fd.add_output(T2)

for _ in range(5):
    fd.execute(inputs, profile=True)
    kps = fd.profile().kernel_profiles
    for kp in kps:
        print(kp.scheduler)
```

## Why didn't the `FusionProfiler` previously work with User Schedules?
1. Previously, the `FusionProfiler` instrumentation location was embedded inside of the `FusionExecutorCache` around segmentation.  Therefore, if a user created a single kernel Fusion and lowered it to a `FusionExecutor`, like is found in tests, the `FusionProfiler` instrumentation would not be triggered.
2. In order to fix the problem from (1.), the `SegmentProfiler` needed to be started and stopped from within the `FusionExecutor`.  This also required moving the capturing of compilation time.
3. Another thing that needed to change for User Schedules was that their compilation needed to be moved into the `execute` method in order to have a way to cleaning `start` and `stop` and the Profiler.
4. Inside the `FusionDefinition`'s C++ `execute()` method, you will find manual instrumentation to do the following that was only previously found in the `FusionExecutorCache` around segmentation:
 a. Start and stop the profiler
 b. Create a single Profiler segment.
 c. Declare the User Scheduler as type: `user`
### Code Example showing the `FusionProfiiler` working with User Schedules:
```python
import torch
from nvfuser import FusionDefinition

inputs = [
    torch.randn((2, 5), dtype=torch.float, device="cuda:0"),
    torch.randn((2, 5), dtype=torch.float, device="cuda:0"),
]

class MyFusion(FusionDefinition):
    def definition(self):
        self.T0 = fd.from_pytorch(inputs[0])
        self.T1 = fd.from_pytorch(inputs[1])
        self.T2 = fd.ops.add(self.T0, self.T1)
        fd.add_output(self.T2)

    def schedule(self):
        pass

fd = MyFusion()
for _ in range(5):
    fd.execute(inputs, profile=True)
    kps = fd.profile().kernel_profiles
    for kp in kps:
        assert kp.scheduler == "user"
```


csrc/executor.cpp
```diff
@@ -1874,6 +1885,14 @@ std::vector<at::Tensor> FusionExecutor::runFusion(
     std::vector<at::Tensor> outputs) {
   FUSER_PERF_SCOPE("FusionExecutor::runFusion");
 
+  if (isProfilerEnabled()) {
+    NVF_CHECK(group_id_ >= 0, "An invalid segment id is passed to FusinProfiler!:", group_id_);
```
rdspring1, line 1889:
```suggestion
    NVF_CHECK(group_id_ >= 0, "An invalid segment id is passed to FusionProfiler!:", group_id_);
```

csrc/executor.cpp
```diff
@@ -261,6 +262,13 @@ void FusionExecutor::compileFusion(
     fusion_ = std::make_unique<Fusion>(*fusion);
     return;
   }
+ 
+  // NOTE: Profiling needs to be started below the isExpressionEvaluated query
+  // given the conditional can exit early from compilation.
+  if (isProfilerEnabled()) {
+    NVF_CHECK(group_id >= 0, "An invalid segment id is passed to FusinProfiler!:", group_id);
```
rdspring1, line 269:
```suggestion
    NVF_CHECK(group_id >= 0, "An invalid segment id is passed to FusionProfiler!:", group_id);
```

csrc/executor.cpp
```diff
@@ -262,6 +263,16 @@ void FusionExecutor::compileFusion(
     return;
   }
 
+  // NOTE: Profiling needs to be started below the isExpressionEvaluated query
+  // given the conditional can exit early from compilation.
```
Priya2698, line 267: What happens when a fusion contains multiple segments with one of them going through expression evaluator?

csrc/kernel_cache.cpp
```diff
@@ -1200,17 +1200,7 @@ std::vector<at::Tensor> FusionKernelRuntime::runKernelWithInput(
...
-    auto& sprof = FusionProfiler::segment(group_id);
-    sprof.inputBytesAccessed(executor.inputBytesProcessed(args));
-    sprof.startKernel(args.getDeviceIndex());
-  }
   auto outputs = executor.runFusion(args, launch_params, compile_params);
-  if (isProfilerEnabled()) {
-    auto& sprof = FusionProfiler::segment(group_id);
-    sprof.stopKernel();
-    sprof.outputBytesAccessed(executor.outputBytesProcessed(outputs));
-  }
```
Priya2698, line 1213: This would capture times for executors which do not have a compiled kernel as well, but it seems we lose this now?

csrc/executor.cpp
kevinstephano: The primary concern of capturing compile time in the `FusionExecutor` is capturing the nvrtc compile time for a code generated kernel given ExpressionEvaluator is not compiling a kernel..  Expression Evaluation would be  captured for a fusion at the `FusionExecutorCache` level.

Priya2698 (approved):
The PR changes LGTM.

My only concern is that larger fusions may include `matmul` and `linear` which will be expression evaluated. These segments will not be timed so we can not directly get the total CUDA time from this profiling.

It may be not as important since our changes should not impact performance of those segments and we can rely on benchmarking for getting the total time measurements.

kevinstephano:
!build

kevinstephano:
!build

kevinstephano:
!build

kevinstephano:
!build

kevinstephano:
!build

PR #2411, opened 2024-06-13
Add another constructor of `Communication` that takes I/O TensorViews instead of the mesh. 

wujingyue:
The other constructor is still used in some tests. 

Communications created by the new constructor must be put in a non-SSA IrContainer. This is because multiple Communications can have the same input and output TensorViews, e.g., Gather with a receiver mesh of size 2.

### DIFF
--- a/csrc/multidevice/communication.cpp
+++ b/csrc/multidevice/communication.cpp
@@ -42,6 +42,8 @@ std::ostream& operator<<(std::ostream& os, const CommunicationType& type) {
     case CommunicationType::SendRecv:
       os << "SendRecv";
       break;
+    default:
+      NVF_ERROR(false, "unrecognized CommunicationType: ", type);
   }
   return os;
 }
@@ -97,16 +99,37 @@ T getInitialValue(c10d::ReduceOp::RedOpType op) {
 }
 
 bool hasRoot(CommunicationType type) {
-  return type == CommunicationType::Gather ||
-      type == CommunicationType::Scatter || type == CommunicationType::Reduce ||
-      type == CommunicationType::Broadcast ||
-      type == CommunicationType::SendRecv;
+  switch (type) {
+    case CommunicationType::Gather:
+    case CommunicationType::Scatter:
+    case CommunicationType::Reduce:
+    case CommunicationType::Broadcast:
+    case CommunicationType::SendRecv:
+      return true;
+    case CommunicationType::Allgather:
+    case CommunicationType::Allreduce:
+    case CommunicationType::ReduceScatter:
+      return false;
+    default:
+      NVF_ERROR(false, "unrecognized CommunicationType: ", type);
+  }
 }
 
 bool isReduction(CommunicationType type) {
-  return type == CommunicationType::Reduce ||
-      type == CommunicationType::Allreduce ||
-      type == CommunicationType::ReduceScatter;
+  switch (type) {
+    case CommunicationType::Reduce:
+    case CommunicationType::Allreduce:
+    case CommunicationType::ReduceScatter:
+      return true;
+    case CommunicationType::Gather:
+    case CommunicationType::Allgather:
+    case CommunicationType::Scatter:
+    case CommunicationType::Broadcast:
+    case CommunicationType::SendRecv:
+      return false;
+    default:
+      NVF_ERROR(false, "unrecognized CommunicationType: ", type);
+  }
 }
 
 } // namespace
@@ -114,15 +137,19 @@ bool isReduction(CommunicationType type) {
 Communication::Communication(
     IrBuilderPasskey passkey,
     CommunicationType type,
-    DeviceMesh mesh,
+    TensorView* out,
+    TensorView* in,
     Team team,
     DeviceIdxType root,
     RedOpType red_op,
-    int64_t scattered_axis,
-    TensorView* input_tv,
-    TensorView* output_tv)
+    int64_t scattered_axis)
     : Expr(passkey) {
-  NVF_ERROR(mesh.size() > 0, "The mesh size must be greater than 0.");
+  NVF_ERROR(
+      in->getDeviceMesh().size() > 0,
+      "The input mesh size must be greater than 0.");
+  NVF_ERROR(
+      out->getDeviceMesh().size() > 0,
+      "The output mesh size must be greater than 0.");
   NVF_ERROR(
       hasRoot(type) == (root >= 0),
       "Root ",
@@ -133,14 +160,9 @@ Communication::Communication(
   NVF_ERROR(
       (type == CommunicationType::ReduceScatter) == (scattered_axis >= 0));
 
-  if (input_tv != nullptr) {
-    addInput(input_tv);
-  }
-  if (output_tv != nullptr) {
-    addOutput(output_tv);
-  }
+  addInput(in);
+  addOutput(out);
   addDataAttribute(type);
-  addDataAttribute(mesh);
   addDataAttribute(team);
   addDataAttribute(root);
   addDataAttribute(red_op);
@@ -168,7 +190,10 @@ std::string Communication::toString(const int indent_size) const {
   if (hasRoot(type())) {
     indent(ss, indent_size + 1) << "root: " << root() << "," << std::endl;
   }
-  indent(ss, indent_size + 1) << "mesh: " << mesh() << "," << std::endl;
+  indent(ss, indent_size + 1)
+      << "sender mesh: " << senderMesh() << "," << std::endl;
+  indent(ss, indent_size + 1)
+      << "receiver mesh: " << receiverMesh() << "," << std::endl;
   indent(ss, indent_size + 1) << "team: " << team() << "," << std::endl;
   indent(ss, indent_size) << "}";
 
@@ -187,7 +212,7 @@ c10::intrusive_ptr<c10d::Work> postBroadcast(
     at::Tensor input_tensor,
     at::Tensor output_tensor) {
   if (my_device_index == communication->root()) {
-    if (communication->isRootInMesh()) {
+    if (communication->receiverMesh().has(communication->root())) {
       // Do a local copy and the subsequent broadcast will be in place. Consider
       // ProcessGroupNCCL::_broadcast_oop so ncclBroadcast doesn't wait for the
       // local copy to complete.
@@ -214,7 +239,7 @@ c10::intrusive_ptr<c10d::Work> postGather(
     at::Tensor input_tensor,
     at::Tensor output_tensor) {
   if (my_device_index == communication->root() &&
-      !communication->isRootInMesh()) {
+      !communication->senderMesh().has(communication->root())) {
     // This is likely a suboptimal way to allocate tensors for nccl. To benefit
     // from zero copy
     // (https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/bufferreg.html),
@@ -232,7 +257,7 @@ c10::intrusive_ptr<c10d::Work> postGather(
     int64_t j = 0;
     for (auto i : c10::irange(communication->team().size())) {
       if (root_relative_index == static_cast<DeviceIdxType>(i) &&
-          !communication->isRootInMesh()) {
+          !communication->senderMesh().has(communication->root())) {
         output_tensors[0].push_back(input_tensor);
         continue;
       }
@@ -273,7 +298,7 @@ c10::intrusive_ptr<c10d::Work> postScatter(
     at::Tensor input_tensor,
     at::Tensor output_tensor) {
   if (my_device_index == communication->root() &&
-      !communication->isRootInMesh()) {
+      !communication->receiverMesh().has(communication->root())) {
     output_tensor = at::empty_like(input_tensor.slice(0, 0, 1));
   }
   std::vector<at::Tensor> output_tensors({output_tensor});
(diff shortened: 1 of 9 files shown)


wujingyue:
!build

wujingyue:
!build

csrc/multidevice/lower_communication.cpp
```diff
@@ -155,11 +157,16 @@ void lowerToBroadcast(
         " vs ",
         receiver_mesh.size());
     for (auto i : c10::irange(sender_mesh.size())) {
-      lowerToBroadcast(
-          my_device_index,
-          sender_mesh.at(i),
-          DeviceMesh({receiver_mesh.at(i)}),
-          comms);
+      const DeviceIdxType sender = sender_mesh.at(i);
```
wujingyue, line 160: I think this change is beneficial but I'm open to ideas. 

Consider a collective send/recv from `mesh{0,1,2,3}` to `mesh{4,5,6,7}` (for pipeline parallelism) or from `mesh{0,1,2,3}` to `mesh{1,2,3,0}` (for p2p ring), that's decomposed into four ncclSend/ncclRecv pairs. 

Currently, send/recv pairs differ in `mesh` and therefore `team`. With this PR, they will differ in `team` but have the same input and output meshes, allowing them to reuse the same input and output TensorViews. 

(These examples also make me think about letting SendRecv take a vector of send/recv pairs. But that's out of the scope of this PR.)

csrc/multidevice/communication.cpp
```diff
@@ -97,32 +99,57 @@ T getInitialValue(c10d::ReduceOp::RedOpType op) {
...
+  }
 }
 
 } // namespace
 
 Communication::Communication(
     IrBuilderPasskey passkey,
     CommunicationType type,
-    DeviceMesh mesh,
+    TensorView* out,
```
wujingyue, line 140: I put `out` before `in` as a convention. See LoadStoreOp and UnaryOp.

tests/cpp/test_multidevice_host_ir.cpp
```diff
@@ -58,9 +58,9 @@ TEST_P(MultiDeviceHostIrTest, SingleFusionSingleComm) {
   fusion->addOutput(tv1_fusion);
 
   DeviceMesh mesh = DeviceMesh::createForNumDevices(communicator_size);
-  if (with_sharding_annotations) {
-    for (auto tv : {tv0_fusion, tv1_fusion}) {
-      tv->setDeviceMesh(mesh);
+  for (auto tv : {tv0_fusion, tv1_fusion}) {
+    tv->setDeviceMesh(mesh);
```
wujingyue, line 62: I changed this for https://github.com/NVIDIA/Fuser/pull/2411#discussion_r1642215029.

tests/cpp/test_multidevice_matmul.cpp
```diff
@@ -253,8 +253,8 @@ TEST_F(DistributedMatmulTest, LayoutNT_ReduceScatter) {
   TensorView* ab = mul(a_b, b_b); // (Ko,M,N,Ki)
   TensorView* c0 = sum(ab, {-1}); // (Ko,M,N,r)
   c0 = segment_set(c0);
-  std::vector<int64_t> orig_size = {K, M, N};
-  std::vector<int64_t> new_size = {K, Mo, Mi, N};
+  std::vector<int64_t> orig_size = {Ko, M, N};
+  std::vector<int64_t> new_size = {Ko, Mo, Mi, N};
```
wujingyue, line 257: This fixes a trivial bug found in https://github.com/NVIDIA/Fuser/pull/2386#discussion_r1635591610, but I'm happy to separate it out if you prefer.

samnordmann (commented):
Thanks for this patch!

While having the ability to provide I/O Tensors to Communication may sounds natural, I'm also concerned because, in principle, the Team should be enough to fully describe a Communication, and therefore we shouldn't rely on the I/O meshes. Currently, we do rely on those meshes, but only in an incidental way, and we should remove this dependency IMO -- I develop this idea in more detail in one of the comment.

I think adding the I/O is nice to express data dependency in the DAG, however, those arguments should be optional -- or, alternatively, Communications should be wrapped inside an op `PostOnStream` that manages the I/O, i.e., we should revert #2328.
This would save a lot of code complexity in the current code base, and also in the present patch (c.f. the tests that need to artificially define I/O TVs).

I am making those points because I'm afraid that the Communication's apparent dependency on the I/O TVs are misleading when it comes to making design decision. I think that the IR `Communication` should be as close possible to the backend API so it matches the generated code better -- adding I/O TVs may add unnecessary complexity, both in usage (artificially create I/O TVs) and in execution/implementation (burry ATen copies and allocation inside collective post)

wujingyue:
!build

csrc/multidevice/lower_communication.cpp
cowanmeg: > (These examples also make me think about letting SendRecv take a vector of send/recv pairs. This way, one SendRev does all and therefore shortens the host IR. But that's out of the scope of this PR.)

Also out of the scope of the PR :) 
Since these programs are SPMD like why not just use an expression to calculate one send/recv pairs instead of expanding out everything? For the pipeline parallelism example 0 sends to 4, 1 to 5 etc. Once a device has been determined to be to be in a pipeline stage then it's guaranteed to be either a sender or receiver for one of many parallel communications.

csrc/multidevice/communication.cpp
```diff
@@ -97,32 +99,57 @@ T getInitialValue(c10d::ReduceOp::RedOpType op) {
...
-    TensorView* output_tv)
+    int64_t scattered_axis)
     : Expr(passkey) {
-  NVF_ERROR(mesh.size() > 0, "The mesh size must be greater than 0.");
+  NVF_ERROR(
+      in->getDeviceMesh().size() > 0,
+      "The input mesh size must be greater than 0.");
+  NVF_ERROR(
+      out->getDeviceMesh().size() > 0,
+      "The output mesh size must be greater than 0.");
```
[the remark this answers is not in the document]
cowanmeg: > We could allow one of the meshes to be empty and inferred from the other, but it feels less explicit and adds complexity to properly infer meshes.

There might be a few edge cases where this could be nice... For example, if a scheduler adds new exprs, it should re-run the sharding propagation pass, but it's easy to forget and we haven't updated all these passes everywhere. I would leave this as an error for now, but something to consider.

cowanmeg (approved):

wujingyue:
!build

wujingyue:
> I think adding the I/O is nice to express data dependency in the DAG, however, those arguments should be optional

I doubt that. Nonetheless, I updated the PR to keep both constructors, which should address your concern now. PTAL!

csrc/multidevice/executor.cpp
```diff
@@ -164,11 +165,21 @@ void MultiDeviceExecutor::postCommunication(SegmentedGroup* group) {
...
+    IrCloner cloner = Fusion::copy(group->getFusion(), container.get());
+    std::vector<Communication*> communications =
+        lowerCommunication(comm_.deviceId(), cloner.clone(expr));
+    for (auto* communication : communications) {
+      container->pushBackTopLevelExprs(communication);
+    }
+  }
 
   // Compute input_tensor and output_tensor.
-  auto input_val = expr->inputs().at(0);
```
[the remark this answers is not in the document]
samnordmann: Oh right, I see.

So then what is the purpose of the communication's I/O, which belong to the HostIrContainer and not the complete fusion?

samnordmann:
> > I think adding the I/O is nice to express data dependency in the DAG, however, those arguments should be optional
> 
> I doubt that. Nonetheless, I updated the PR to keep both constructors, which should address your concern now. PTAL!

To sum up our discussion, we agreed that a communication's execution should rely only on the Team, not on I/O TensorViews nor meshes. So, we should remove the dependency (by not supporting "unorthodox MPI collectives" where the root is not in the mesh, see [here](https://github.com/NVIDIA/Fuser/pull/2411#discussion_r1642574884)), and there should be a `Communication`'s constructor taking only a team, but no mesh nor TV.

In addition to that, you would like to have another Communication's constructor taking not only a Team but also I/O TVs. Those I/O Tvs are used only to express data dependency in the complete fusion. (This constructor should/could eventually be a thin wrapper around the first constructor mentioned above, taking a Team only). Even though there is no usage right now of the I/O TVs, we add this for the future -- I can't really see how it will be useful but I trust you. My only remaining concern is [this comment](https://github.com/NVIDIA/Fuser/pull/2411#discussion_r1645588649): if the I/O of the communication are TVs belonging to the HostIrContainer and not the full fusion, this will not embed the desired data dependency. Wdyt?

csrc/multidevice/executor.cpp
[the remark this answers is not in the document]
wujingyue: As said in https://github.com/NVIDIA/Fuser/pull/2250#issuecomment-2135875908, I plan to put `Communication`s as top-level expressions in a HostIrContainer. So having input/output `Val*`s allows a host IR executor to find their corresponding input/output tensors/buffers.

csrc/multidevice/communication.cpp
```diff
@@ -97,54 +99,101 @@ T getInitialValue(c10d::ReduceOp::RedOpType op) {
...
+      in->getDeviceMesh().size() > 0,
+      "The input mesh size must be greater than 0.");
   NVF_ERROR(
-      (type == CommunicationType::ReduceScatter) == (scattered_axis >= 0));
+      out->getDeviceMesh().size() > 0,
+      "The output mesh size must be greater than 0.");
+
+  addInput(in);
+  addOutput(out);
+  addDataAttribute(type);
```
samnordmann, line 156: why not calling the other constructor here instead of duplicating?

csrc/multidevice/communication.h
```diff
@@ -81,8 +90,21 @@ class Communication : public Expr {
     return attribute<CommunicationType>(0);
   }
 
-  const DeviceMesh& mesh() const {
-    return attribute<DeviceMesh>(1);
+  TensorView* out() const {
```
samnordmann, line 93: `out`, `in` will throw on instances created with the constructor without I/O Tvs, right?

csrc/multidevice/executor.cpp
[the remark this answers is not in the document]
samnordmann: Ok, thanks

samnordmann (approved):
I think there is a slight misunderstanding, sorry if I have been unclear, but let me try to clarify. I have been advocating (e.g. [here](https://github.com/NVIDIA/Fuser/pull/2411#discussion_r1642574884), [here](https://github.com/NVIDIA/Fuser/pull/2411#pullrequestreview-2122474586), [there](https://github.com/NVIDIA/Fuser/pull/2411#issuecomment-2178002950) and orally) for having a constructor taking only the Team, no DeviceMesh nor TVs. However, this requires removing the dependency on the device Mesh, which need additional work, and so we decided to do this in a future PR.

Besides the aforementioned constructor, you would like another constructor taking additional I/O TVs, used to express data dependency in the complete fusion. This makes sense.

But I don't understand why in this PR we have *two* constructors
1) one constructor taking I/O TVs, but no DeviceMesh, and
2) an other taking a DeviceMesh but no I/O TVs.

I don't see the benefit of having 2), and I find 1) always better and more general. So I would be in favor of removing it completely and keeping only 1), as you originally did in the first version of the PR.

I really don't want to slow things down, so I'm approving, please do as you think is best. I'm happy anyway we had those discussions which I feel helped understand the northern light and be on the same page.

csrc/multidevice/communication.h
```diff
@@ -48,21 +49,29 @@ class Communication : public Expr {
...
+  Communication(
+      IrBuilderPasskey passkey,
+      CommunicationType type,
+      DeviceMesh mesh,
+      Team team, // All devices involved in this communication. It must include
+                 // `root`. It can be a subset of `root`+`mesh` in case of 2D
+                 // sharding.
+      DeviceIdxType root = -1,
+      RedOpType red_op = RedOpType::UNUSED,
+      int64_t scattered_axis = -1);
```
naoyam, line 74: Could you comment when to use which version?

wujingyue:
> I don't see the benefit of having 2), and I find 1) always better and more general. So I would be in favor of removing 2) completely and keeping only 1), as you originally did in the first version of the PR.

I bifurcated Communication's constructor because
1. According to https://github.com/NVIDIA/Fuser/pull/2411#pullrequestreview-2122474586, you want the I/O TVs to be optional, so the first version wouldn't work for you. 
2. At this moment, the mesh is still needed for execution, so the constructor needs to take it when the I/O TVs are null. 
3. While I can make one constructor with the I/O TVs and the mesh all being optional, C++ isn't as good as Python at having too many optional arguments, e.g., all arguments before a non-default argument must be explicitly specified. 

That being said, I do agree with you on the end state of removing `mesh` from Communication's constructor, so the above is about order of actions.

csrc/multidevice/communication.cpp
wujingyue: Good point. I tried constructor delegation but couldn't figure out a good way to call `validate` and also check the mesh -- I want both constructors to call the same `validate` but check meshes differently. I'll leave it as is for now, because we'll eventually (or soon?) remove the mesh.

csrc/multidevice/communication.h
wujingyue: Done -- updated the comment.

wujingyue:
!build

PR #2413, opened 2024-06-14
Refactor lowering of TMA

zasdfgbnm:
This PR refactors the lowering of TMA into two stages:
1. An analysis that computes the ValGroups for all the roles (tile ValGroup, box ValGroup, partitioned ValGroup, etc.)
2. An index lowering that uses the result of 1 to get the index.

I intentionally keep the algorithm unchanged. The analysis is almost a copy-paste of the original `getCpAsyncBulkGmemIndex`, except that it is modified to traverse on the exact graph, instead of on the gmem_tv replayed as consumer. In the future, I might do more cleanup to better take advantage of IdModel.

In step 2, we are using the legacy indexing, instead of the IdModel based indexing. I intentionally not change this so that this PR is effectively a no-op. In the future, the legacy indexing will be replaced with IdModel based indexing.

cc: @rdspring1 @protonu

### DIFF
--- a/csrc/device_lower/lower2device.cpp
+++ b/csrc/device_lower/lower2device.cpp
@@ -331,6 +331,42 @@ kir::Kernel* GpuLower::run() {
   return kernel_.get();
 }
 
+bool requiresIdModel(Fusion* fusion) {
+  // TMA requires IdModel
+  for (auto expr : fusion->exprs()) {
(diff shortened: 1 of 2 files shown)


zasdfgbnm:
!build

zasdfgbnm:
!build

naoyam (approved):
LGTM

PR #2415, opened 2024-06-14
Make `TensorIndexer::traversalGraph` public

zasdfgbnm:
When writing analysis, we might store `ValGroup`s instead of `IterDomain`s as the result of the analysis, for the convenience of usage. It could be important to use the `ValGroup`s belonging to the same IdGraph as the indexer. So I make this public so analysis writer could access and write their analysis based on this graph.


zasdfgbnm:
!build

zasdfgbnm:
!build

zasdfgbnm:
!build

naoyam (approved):
LGTM

PR #2417, opened 2024-06-14
Move `replaceValRecursively` to `computeIndex`

zasdfgbnm:
So that the index map is correct and can be used elsewhere.

### DIFF
--- a/csrc/id_model/indexing.cpp
+++ b/csrc/id_model/indexing.cpp
@@ -491,10 +491,7 @@ Val* TensorIndexer::getLinearIndex(TensorView* tv, const Expr* expr) const {
         index, SimplifyingIrBuilder::mulExpr(stride, idx));
   }
 
-  const auto& replacement_map =
-      getIndexReplacementMap(index_info.loop_domains, index_info.index_map);
-
-  return ir_utils::replaceValRecursively(index, replacement_map);
+  return index;
 }
 
 // Get the loop domains of a given expr, which are (potentially
@@ -531,6 +528,14 @@ IndexingInfo TensorIndexer::computeIndex(
   }
 
   IndexingInfo info{loop_domains, traversal_path, index_compute.indexMap()};
+
+  const auto& replacement_map =
+      getIndexReplacementMap(loop_domains, info.index_map);
+
+  for (auto& [k, v] : info.index_map) {
+    v = ir_utils::replaceValRecursively(v, replacement_map);
+  }
+
   return info;
 }
 


zasdfgbnm:
!build

naoyam:
> So that the index map is correct and can be used elsewhere.

Did you mean the current usage of the index map is not correct?

csrc/id_model/indexing.cpp
```diff
@@ -531,6 +528,14 @@ IndexingInfo TensorIndexer::computeIndex(
   }
 
   IndexingInfo info{loop_domains, traversal_path, index_compute.indexMap()};
+
```
naoyam, line 531: Could you give some more info why this replacement should be done here?

I'm still in the middle of refactoring, but my current plan is to have this this function as a common routine for both tensor expr indexing and predicate indexing. `getLinearIndex` is the one for expr indexing, and I'm going to have `getPredicateIndex` for predicates. I was planning to have `computeIndex` just does some common operations. The main difference of `getLinearIndex` and `getPredicateIndex` is the initial index map, so the index replacement would be done differently.

zasdfgbnm:
> Did you mean the current usage of the index map is not correct?

I mean the index itself in the index map is incorrect, because, for example, the index vectorized dims are not set as zero.

> I'm still in the middle of refactoring, but my current plan is to have this this function as a common routine for both tensor expr indexing and predicate indexing. `getLinearIndex` is the one for expr indexing, and I'm going to have `getPredicateIndex` for predicates. I was planning to have `computeIndex` just does some common operations. The main difference of `getLinearIndex` and `getPredicateIndex` is the initial index map, so the index replacement would be done differently.

I see the plan. Then how would you want to implement the per dim indexing?
https://github.com/NVIDIA/Fuser/blob/979336479d632e280cc209101126a88fdff1b7dc/csrc/index_compute.h#L520-L536

I think this is almost what I wanted, except that I need each dim to be specified by myself, instead of hard coding it to the logical domain.

naoyam:
>  Then how would you want to implement the per dim indexing?

Still need to think about it, but it's likely something that would build a vector of indices based on the index map built by `computeIndex`. We should have a parameter specifying which ID to index.

naoyam:
> > > Then how would you want to implement the per dim indexing?
> > 
> > 
> > Still need to think about it, but it's likely something that would build a vector of indices based on the index map built by `computeIndex`. We should have a parameter specifying which ID to index.
> 
> So `computeIndex` will still return the raw unreplaced indices, and the `getPerDimIndex`-like thing will do the replacement? This sounds reasonable to me. I can update this PR to leave `computeIndex` unchanged and add a `getPerDimIndex` (I may not use this exact name).

Likely yes. I'll proceed refactoring and merging the remaining pieces from my prototype branch (#2238). As long as `computeIndex` is unchanged, it should be fine. I don't have a fully settled plan yet, so I may also change my mind but will ask for your review.

zasdfgbnm:
@naoyam I changed this PR to move `computeIndex` back as private, and added a new method `getIndexFor` for computing the index of the given ValGroups. The index replacement is done at `getIndexFor`. `getLinearIndex` is changed to use `getIndexFor` instead of `computeIndex` so that its code is shorter, but this change is not mandatory, and I can revert it back if it does not make sense to do so.

zasdfgbnm:
!build

naoyam (approved):
LGTM. Thanks for accommodating the changes.

PR #2419, opened 2024-06-14
Combine MatmulRole::INPUT_{A,B} into MatmulRole::OPERAND

jacobhinkle:
Also rename `MatmulRole::INPUT_C` to `MatmulRole::EPILOGUE_INPUT` and `MatmulRole::OUTPUT_D` to `MatmulRole::OUTPUT`. This is in preparation for generalizing our scheduler to handle any number of operands and `MatmulPattern`s.

Even when fusing a single `MatmulPattern` we might not have exactly one Fusion input for A or B. We could have zero if we use a factory method to create the operand, or we could have two if we do some pointwise operations in the prologue. I am trying to refactor our code to handle any of these cases gracefully.

Next steps in generalizing the matmul scheduler:

1. Change `MmaLayout` from an enum to a vector of inner allocation dimensions i.e. `std::vector<MatmulDomain>`. After that, the only reference to "A" and "B" will be the inputs to the actual MmaOp.
2. Modify `scheduleMatmul` to schedule all the tensors in `tensor_roles[MatmulRole::OPERAND]` in the same way. This means we will not have separate `acr` and `bcr` tensors, for example.
3. Modify `MatmulParams` and `matmul_heuristic_plugin::KernelConfig` to remove explicit mentions of A and B.
4. Relax assumptions that only two operands exist and remove canSchedule checks so that we accept a sample fusion that uses a factory method to create one or both operands.
5. Enable fusion of multiple MatmulPatterns like the llama2 FFN. By this point we should just need to modify the scheduler to be more careful during propagations, and modify some canSchedule checks.

### DIFF
--- a/csrc/mma_type.h
+++ b/csrc/mma_type.h
@@ -29,16 +29,13 @@ constexpr std::string_view MATMUL_LOG_PREFIX = "[MATMUL DEBUG] ";
 enum class MatmulDomain { M = 0, N, K, Batch };
 
 //! Named descriptors of TensorView roles in fusion
-//!  INPUT_A - a producer of MMA input A
-//!  INPUT_B - a producer of MMA input B
-//!  INPUT_C - a producer of a tensor used in fusion epilogue,
-//!            for example tensor used in beta scaling fusion
-//!  OUTPUT_D - fusion outputs that have the matmul as a dependency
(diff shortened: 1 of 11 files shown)


jacobhinkle:
!build

zasdfgbnm (approved):

PR #2416, opened 2024-06-14
Use `TensorIndexer` for TMA indexing

zasdfgbnm:
Stacked on https://github.com/NVIDIA/Fuser/pull/2413 https://github.com/NVIDIA/Fuser/pull/2414 https://github.com/NVIDIA/Fuser/pull/2415 https://github.com/NVIDIA/Fuser/pull/2417

PR is ready, but not very readable now because it contain diff from other PRs. Needs to merge other PRs first and rebase.

PR #2420, opened 2024-06-14
Simplify handling of "define box by compositing"

zasdfgbnm:
When there is a "define box by compositing", in principle, we should be indexing into the _imagined_ partitioned IterDomain, as shown below"

However, we were not able to do that in the past because indexing only supports backward propagation. With `TensorIndexer` and `AbstractTensor`, we can easily create the _imagined_ partitioned ValGroup and index into it.
So there is no need to use the state machine approach to manually combine indices, instead, the index of the  _imagined_ partitioned ValGroup can be directly used as the index for the TMA PTX instructions.

### DIFF
--- a/csrc/device_lower/analysis/tma.cpp
+++ b/csrc/device_lower/analysis/tma.cpp
@@ -413,6 +413,23 @@ TMAInfo getTMAInfo(LoadStoreOp* ldst) {
         : (!tma_g_to_box_g.count(g) ? C
                                     : (tma_g_to_stride_g.count(g) ? SB : CB));
   };
+  for (auto i : c10::irange((int64_t)tma_domain.size())) {
+    auto t = gtype(i);
+    auto g = tma_domain[i].as<ValGroupAndItsGraph>().group;
+    std::cout << g->toString() << " -> ";
+    if (t == P) {
+      std::cout << "P" << std::endl;
+    }
+    if (t == C) {
+      std::cout << "C" << std::endl;
+    }
+    if (t == SB) {
+      std::cout << "SB" << std::endl;
+    }
+    if (t == CB) {
+      std::cout << "CB" << std::endl;
+    }
+  }
   // merge contiguous C groups and CB groups
   int64_t i = 0;
   while (i < (int64_t)tma_domain.size() - 1) {
--- a/csrc/device_lower/analysis/tma.cpp
+++ b/csrc/device_lower/analysis/tma.cpp
@@ -1,0 +1,532 @@
+// clang-format off
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2023-present NVIDIA CORPORATION & AFFILIATES.
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+// clang-format on
+#include <device_lower/analysis/tma.h>
+#include <device_lower/lower2device.h>
+#include <id_model/id_model.h>
+#include <ir/utils.h>
+#include <val_graph.h>
+#include <val_graph_visitor.h>
+
+#include <list>
+#include <unordered_map>
+#include <vector>
+
+namespace nvfuser {
+
+namespace {
+
+int64_t getCpAsyncBulkTensorSwizzleSize(TensorView* smem_tv) {
+  auto exprs = DependencyCheck::getAllExprsBetween(
+      {smem_tv->getLogicalDomain().begin(), smem_tv->getLogicalDomain().end()},
+      {smem_tv->getMaybeAllocationDomain().begin(),
+       smem_tv->getMaybeAllocationDomain().end()});
+  for (auto expr : exprs) {
+    if (auto s = dynamic_cast<Swizzle*>(expr)) {
+      return s->inX()->extent()->evaluate().as<int64_t>();
+    }
+  }
+  return 1;
+}
+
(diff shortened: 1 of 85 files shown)


zasdfgbnm:
!build

zasdfgbnm:
Stacked on https://github.com/NVIDIA/Fuser/pull/2416, marked as draft to avoid accidental merge.

csrc/device_lower/analysis/tma.h
```diff
@@ -20,97 +20,27 @@ namespace nvfuser {
...
-  Val* elementStride() const override {
-    return fusion->oneVal();
-  }
-};
-
 struct TMADim {
-  ValGroups partitioned;
-  ValGroups coordinate;
-  std::unique_ptr<Box> box;
+  ValGroup tma;
```
zasdfgbnm, line 24:
```suggestion
  ValGroup partitioned;
```

zasdfgbnm:
!build

csrc/id_model/indexing.h
```diff
@@ -43,7 +43,7 @@ struct IndexingInfo {
 // domains may be promoted.
 class TensorIndexer {
  public:
-  TensorIndexer(const IdModel& id_model);
+  TensorIndexer(IdModel& id_model);
```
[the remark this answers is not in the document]
naoyam: Ah, that's because we create (temporary) domains for indexing? Could you leave a comment why it can't be a const ref?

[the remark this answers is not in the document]
zasdfgbnm: > Ah, that's because we create (temporary) domains for indexing?

Yes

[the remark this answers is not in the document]
zasdfgbnm: Comment added.

zasdfgbnm:
!build

csrc/id_model/indexing.h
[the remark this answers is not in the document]
naoyam: I'd say it's because indexing may create new domains and need to update the graph.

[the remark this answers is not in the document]
zasdfgbnm: comment updated

csrc/device_lower/analysis/tma.cpp
```diff
@@ -378,105 +375,96 @@ TMAInfo getTMAInfo(LoadStoreOp* ldst) {
 
   // So far, we have infered the TMA domain. The size of TMA domain is not
   // necessarily the dimensionality of TMA because we support defining box
-  // by compositing. We use a state machine to infer the dimensions of TMA.
-  //
+  // by compositing. We use AbstractTensor to further merge the TMA domain to
```
naoyam, line 378: IIUC, we are creating the true allocation domain dynamically here. If so, why not do so when defining the tensor, not just doing here dynamically? If the tensor has the right allocation domain as defined below, I suppose the indexing should just work.

naoyam: To be clear, I'm not asking for a change. Just want to make sure if what I'm thinking about makes any sense.

csrc/id_model/indexing.h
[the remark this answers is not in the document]
naoyam: Thanks!

csrc/device_lower/analysis/tma.cpp
zasdfgbnm: Yes, we are creating the true allocation domain dynamically here. And you are correct that, if we just define it in the tensor, it will work.

The reason for not defining it in the tensor is exactly the same reason of why supporting this "define box by compositing" at all. It is mostly about "There is this labor that I have to do, should I do it in the scheduler, or should I do it in lowering?" and I choose to do it in lowering. The reason for this decision is because scheduling is already a very difficult problem, so I want to free the scheduler from having to do the book keeping work of TMA.

For example, if we have a tensor `[I0, I1, I2, I3, I4, I5]`, and I want to schedule a pointwise relu kernel. With this "define box by compositing" mechanism, the scheduler author could just think like below:

> I will just tile the tensor on the innermost two dims as `(4, 256)` and the other dims will be automatically handled.

However, without this "define box by compositing" mechanism, the scheduler author still need to analyze the contiguity of the input tensor, and merge contiguous dims together to set the allocation domain of the global tensor, so that we will not use a non-existing 6D TMA. Also, non-permute allocation domains on fusion inputs/outputs could be hard for things like, stride order inference.

zasdfgbnm:
!build

csrc/device_lower/analysis/tma.cpp
naoyam: Thanks. That makes sense.

csrc/id_model/indexing.h
```diff
@@ -59,7 +61,9 @@ class TensorIndexer {
 
   // The AlmostExact graph is used since size-1 splits and merges
   // should not affect actual index exprs.
-  const ValGraph& traversalGraph() const {
+  // Returns non-const reference because indexing may create new domains and
+  // need to update the graph.
+  ValGraph& traversalGraph() const {
```
naoyam, line 66: My horrible C++ understanding is breaking here. It returns a non-const reference but the function itself is const. How is it possible???

naoyam (approved):
LGTM

csrc/id_model/indexing.h
naoyam: Thank you. Makes sense.

PR #2421, opened 2024-06-14
Multidevice Executor add accessor to fusion executor caches

cowanmeg:
Add `MultiDeviceExecutor::getFusionExecutorCaches` which returns a vector of pointers to the multidevice executor's fusion executor caches.

Note, Also renames the field `workspace` to `workspace_` to match naming conventions. 

cc @Priya2698 for https://github.com/NVIDIA/Fuser/pull/2386

### DIFF
--- a/csrc/multidevice/executor.cpp
+++ b/csrc/multidevice/executor.cpp
@@ -79,7 +79,7 @@ MultiDeviceExecutor::MultiDeviceExecutor(
     should_run_[group] = involvedDevices(expr).count(comm_.deviceId());
   }
   // prepare the order in which to launch the kernels/comms
-  prepareRuntimeOrder(staged_fusion_.get(), workspace);
+  prepareRuntimeOrder(staged_fusion_.get(), workspace_);
 
   // Allocator setup
   // vals_to_allocate_ stores the tensors that need to be allocated at runtime,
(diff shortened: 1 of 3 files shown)


cowanmeg:
!build

csrc/multidevice/executor.cpp
```diff
@@ -277,4 +277,18 @@ std::ostream& MultiDeviceExecutor::print() {
...
 }
 
+std::vector<FusionExecutorCache*> MultiDeviceExecutor::
+    getFusionExecutorCaches() {
+  NVF_CHECK(
+      params_.use_fusion_executor_cache,
+      "MultideviceExecutor must be configured to use FusionExecutorCache");
+  std::vector<FusionExecutorCache*> fecs;
+  for (SegmentedGroup* group : workspace_.group_run_order) {
+    if (!is_resharding_.at(group)) {
```
wujingyue, line 287: What do you think about
```suggestion
    if (fec_.count(group) > 0) {
```

samnordmann (commented):
LGTM! One thing: I'm wondering how the caller of `getFusionExecutorCaches` can figure out what is the order of the `FusionExecutorCache` stored in the vector without accessing `workspace_. group_run_order`. Is my comment relevant?

cowanmeg:
> LGTM! One thing: I'm wondering how the caller of `getFusionExecutorCaches` can figure out what is the order of the `FusionExecutorCache` stored in the vector without accessing `workspace_.group_run_order`. Is my comment relevant?

Yup relevant question. I think the only practical use case is for simple tests like these matmul tests where we have a linear sequence of computation so we know the order. The background for this PR is for matmul with ATen execution we want to check that the executor actually goes through ATen (https://github.com/NVIDIA/Fuser/pull/2386#issuecomment-2162265382) but to do that we need to access the FusionExecutorCache.

wujingyue (approved):

Priya2698:
Thanks for this PR -- I added these checks to the multidevice matmul tests!

PR #2422, opened 2024-06-14
Change getProblemLayout to getOperandInnerDims

jacobhinkle:
This is related to #2419 which is working toward generalizing our scheduler to handle any number of operands instead of exactly 2.

In order to keep the change small, I introduced a `MmaLayout getMatmulProblemLayout(Fusion*)` utility in `tests/cpp/utils.cpp`. This should give the same result that `mma_utils::getProblemLayout(fusion)` used to give, but asserting instead of wrapping in a `DataWrapperOpt`, so we don't need to change our supported layouts much. For new code, we could change to support vectors of inner dimension roles instead.

After this PR, `MmaLayout` should only be used in these places
- In `MmaOpUtils::getInputLayout` which finds the layout of the actual inputs to the `MmaOp` (not the Fusion inputs). This is the correct location to use `MmaLayout` since `MmaOp` supports a single A operand and a single B operand.
- In C++ matmul tests, which still use `MmaLayout` to describe the problem layout.
- In C++ matmul benchmarks, which similarly use `MmaLayout`.

### DIFF
--- a/csrc/scheduler/mma_utils.h
+++ b/csrc/scheduler/mma_utils.h
@@ -282,7 +282,10 @@ struct MatmulPattern {
 //! Traverse the fusion to find supported matmul patterns
 std::vector<MatmulPattern> findMatmulPatterns(Fusion* fusion);
 
-using MatmulProblemLayoutOpt = DataWrapperOpt<MmaLayout>;
+//! This is a vector of roles describing the inner dimension of each operand
+using MatmulOperandInnerDims = std::vector<MatmulDomain>;
+
+using MatmulOperandInnerDimsOpt = DataWrapperOpt<MatmulOperandInnerDims>;
 using ProblemIterDomainsOpt = DataWrapperOpt<ProblemIterDomains>;
 using DimRolesMapOpt = DataWrapperOpt<DimRolesMap>;
 using TensorRolesMapOpt = DataWrapperOpt<TensorRolesMap>;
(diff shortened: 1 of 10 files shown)


jacobhinkle:
!build

jacobhinkle:
!build

csrc/scheduler/mma_utils.h
```diff
@@ -282,16 +282,19 @@ struct MatmulPattern {
 //! Traverse the fusion to find supported matmul patterns
 std::vector<MatmulPattern> findMatmulPatterns(Fusion* fusion);
 
-using MatmulProblemLayoutOpt = DataWrapperOpt<MmaLayout>;
+//! This is a vector of roles describing the inner dimension of each operand
+using MatmulOperandInnerDims = std::vector<MatmulDomain>;
```
zasdfgbnm, line 286: Should this be a `std::array<MatmulDomain, 2>`?

zasdfgbnm (approved):

csrc/scheduler/mma_utils.h
jacobhinkle: I am hoping to generalize to allowing any number of operands.

jacobhinkle:
!build

jacobhinkle:
Test failure is truediv_complex 🙄 . I guess we still allow too small denominators in #2392 ; we could clamp the denominator to at least modulus 1.0 instead. Anyway, that error is not related to this PR, so I'm merging.

PR #2423, opened 2024-06-14
Add a note about `mpirun -continuous`.

wujingyue:


tfogal (approved):
SGTM! Thank you for recording debugging hints to our team.

FWIW, I think these hints are actually OpenMPI-specific, though I could be wrong there. Back in the day at least, `mpirun` took a few standardized arguments and then each implementation accepted a number of implementation-defined ones. Maybe all these are standardized now, though.

PR #2424, opened 2024-06-14
Add swizzle to TensorIndexer

zasdfgbnm:


tests/cpp/test_indexing.cpp
```diff
@@ -1076,4 +1082,49 @@ TEST_F(IndexingTest, SimpleVectorize) {
   IndexValidator<GetReference>::validate(&fusion);
 }
 
+TEST_F(IndexingTest, DISABLED_Swizzle) {
```
zasdfgbnm, line 1085: Can not pass yet, requires domains to be decoupled

naoyam (approved):
I assume this is mainly just as a placeholder since the test doesn't pass yet.

zasdfgbnm:
> I assume this is mainly just as a placeholder since the test doesn't pass yet.

yes

PR #2431, opened 2024-06-17
Fix postSendRecv to use relative indices.

wujingyue:
Also updates the test to use the subteam backend to expose the bug.

### DIFF
--- a/tests/cpp/test_multidevice_communications.cpp
+++ b/tests/cpp/test_multidevice_communications.cpp
@@ -186,10 +186,11 @@ TEST_P(CommunicationTest, SendRecv) {
     GTEST_SKIP() << "This test needs at least 2 GPUs and 2 ranks.";
   }
 
-  constexpr DeviceIdxType sender = 0;
-  constexpr DeviceIdxType receiver = 1;
-  if (communicator->deviceId() > 1) {
-    // Only devices 0 and 1 participate.
+  constexpr DeviceIdxType sender = 1;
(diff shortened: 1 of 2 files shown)


wujingyue:
!build

cowanmeg (approved):
Good catch on the relative index errors!

tests/cpp/test_multidevice_communications.cpp
```diff
@@ -246,13 +245,15 @@ TEST_P(CommunicationTest, SendRecvToSelf) {
...
+  c10d::Backend* backend =
+      communicator->getBackendForTeam(communication->team(), GetParam());
   for (auto repetition : c10::irange(kNumRepetitions)) {
     input_tensor.copy_(at::arange(kTensorSize, tensor_options) + repetition);
 
     postSingleCommunication(
         communication,
         communicator->deviceId(),
-        backend_,
+        backend,
```
samnordmann, line 256: both backend are valid, and `backend_` (i.e. the world communicator) is probably more representative of how we are going to use `SendRecv`

samnordmann (approved):
Thanks!

tests/cpp/test_multidevice_communications.cpp
wujingyue: Agreed. I changed the PR description to not say "right" because both are right. This change is only to expose the bug.

wujingyue:
!build

PR #2435, opened 2024-06-18
Warn if Jax is unavailable as reference function

### LINKED ISSUE #2433: Check whether JAX exists for `pytest_ops.py` tests and give warning instead of failing if JAX is not installed
We get scenarios where users do not install JAX but would like to test none JAX tests and makes sure they pass.

Could we give a warning instead of an error?

rdspring1:
This PR fixes #2433


rdspring1:
!build

tests/python/pytest_core.py
```diff
@@ -52,6 +50,12 @@ def __repr__(self):
         return f"[SampleInput args={self.args} kwargs={self.kwargs}]"
 
     def jax(self):
+        from pytest_utils import JAX_AVAILABLE
+
+        if JAX_AVAILABLE:
```
jjsjann123, line 55: nitpick: this condition doesn't do anything here.

i.e. if jax is not available it's still going to fail when we try to use jnp/torch_to_jax_dtype_map below.
Maybe throw an error instead of guard the import here.

jjsjann123 (approved):
LGTM.

Thanks for patching this!

rdspring1:
!build

PR #2437, opened 2024-06-18
allow reduction/normalization schedulers to handle reshape with only split transforms

liqiangxl:
Issue: #2375 
**Fix:** When the reshape transformations from `root domains` to `logical domains` consists only `split`, it is safe for the reduction scheduler to merge any two `reduction domains`. We can skip `reductionInterferingView` check.

**Results:**
With this PR, the reshape of input ([N,C,H,W]  -> [N, G, C/G, H, W]) can be fused with normalization into one kernel.
Group norm splits into 2 kernels instead of 3.

**Following works:**
(1) It also seems safe if the reshape transformations from `root domains` to `logical domains` contains a `merge`, but all of the merged IDs are mapped to reduction dims or iteration dims.
(2) Needs to fix the reshape of output in group norm.

### DIFF
--- a/csrc/scheduler/registry_utils.cpp
+++ b/csrc/scheduler/registry_utils.cpp
@@ -395,9 +395,6 @@ bool reductionInterferingView(
     Fusion* fusion,
     const ComputeAtMap& ca_map,
     TensorView* reduction_reference) {
-    if(std::getenv("DISABLE_VIEW_CHECK") != nullptr){
-      return false;
-    }
   // Make sure the view doesn't interfere with how we'll want to schedule
   // it. If we might want to do a 3D scheduler make sure views are disjoint
   // based on what the 3D scheduler's merges would be.
--- a/csrc/scheduler/registry_utils.cpp
+++ b/csrc/scheduler/registry_utils.cpp
@@ -395,6 +395,8 @@ bool reductionInterferingView(
     Fusion* fusion,
     const ComputeAtMap& ca_map,
     TensorView* reduction_reference) {
+
+  // return false;
   // Make sure the view doesn't interfere with how we'll want to schedule
   // it. If we might want to do a 3D scheduler make sure views are disjoint
   // based on what the 3D scheduler's merges would be.
@@ -413,6 +415,7 @@ bool reductionInterferingView(
   };
 
   std::vector<IterDomain*> dims = reduction_reference->getLogicalDomain();
+  reduction_reference->printTransforms();
 
   // The disjoint groups we need for this scheduler
   std::vector<std::vector<IterDomain*>> groups;
@@ -440,6 +443,11 @@ bool reductionInterferingView(
       processed.emplace(dims[dim_i]);
     }
 
+    std::cout << "========= dimension: " << dimension << " =========\n current_dims:" << std::endl;
+    for (auto id : current_dims) {
+      std::cout << id->toString() << std::endl;
+    }
+
     // Don't add empty group (would happen if it's a 2D scheduler not 3D)
     if (!current_dims.empty()) {
       groups.push_back(current_dims);
@@ -454,6 +462,14 @@ bool reductionInterferingView(
   auto disjoint_rfactor_sets = scheduler_utils::disjointLogicalSets(fusion);
   auto disjoint_set_information = scheduler_utils::getDisjointLogicalSetsOf(
       fusion, reduction_reference, disjoint_rfactor_sets);
+
+  std::cout << "disjointLogicalSets:\n" << disjoint_rfactor_sets.toString() << std::endl;
+  std::cout << "disjoint_set_ids: ";
+  for(auto id : disjoint_set_information.disjoint_set_ids) {
+    std::cout << id << " ";
+  }
+  std::cout << std::endl;
+
 
   // Convert id's in groups to disjoint_set_ids of disjoint_set_information
   std::vector<std::vector<int64_t>> disjoint_groups;
@@ -476,9 +492,15 @@ bool reductionInterferingView(
               (int64_t)disjoint_set_information.disjoint_set_ids.size(),
           "Error computing disjoint group on the logical domain of ",
           reduction_reference->toString());
+      std::cout << "id: " << id->toString() << ", logical_pos: " << logical_pos << std::endl;
       disjoint_id_sets.push_back(
           disjoint_set_information.disjoint_set_ids[logical_pos]);
     }
+    std::cout << "disjoint_id_sets: ";
+    for(auto id : disjoint_id_sets) {
+      std::cout << id << " ";
+    }
+    std::cout << std::endl;
     disjoint_groups.push_back(disjoint_id_sets);
   }
 
@@ -486,6 +508,7 @@ bool reductionInterferingView(
   // will interfere with the schedule. TODO: Make this better complexity,
   // since it should be relatively small int vectors of a small total nDims,
   // not too worried about it now.
+  return false;
 
   for (auto first_dim_i : c10::irange(disjoint_groups.size())) {
     for (auto second_dim_i = first_dim_i + 1;
--- a/csrc/scheduler/registry_utils.cpp
+++ b/csrc/scheduler/registry_utils.cpp
@@ -395,6 +395,9 @@ bool reductionInterferingView(
     Fusion* fusion,
     const ComputeAtMap& ca_map,
     TensorView* reduction_reference) {
+    if(std::getenv("DISABLE_VIEW_CHECK") != nullptr){
+      return false;
+    }
   // Make sure the view doesn't interfere with how we'll want to schedule
   // it. If we might want to do a 3D scheduler make sure views are disjoint
   // based on what the 3D scheduler's merges would be.
--- a/csrc/scheduler/registry_utils.cpp
+++ b/csrc/scheduler/registry_utils.cpp
@@ -389,14 +389,32 @@ bool requiresForwardViewReplay(Fusion* fusion, ComputeAtMap& ca_map) {
   return false;
 }
 
+namespace {
+bool isSplitOnlyReshapeTransform(Fusion* fusion) {
+  const auto& view_ops = ir_utils::getViewOps(fusion);
+  for (auto view_op : view_ops) {
+    auto tv = view_op->out();
+    for (auto expr : StmtSort::getExprsTo(
+             {tv->getLogicalDomain().begin(), tv->getLogicalDomain().end()})) {
+      if (!expr->isA<Split>()) {
+        return false;
+      }
+    }
+  }
+  return true;
+}
+} // namespace
+
 // Returns if view interferes with how we want to treat the reference, being at
 // least a 2D reduction schedule but maybe a 3D reduction schedule.
 bool reductionInterferingView(
     Fusion* fusion,
     const ComputeAtMap& ca_map,
     TensorView* reduction_reference) {
-
-  // return false;
+  // If reshape transform only has split, it shouldn't influence reduction.
+  if (isSplitOnlyReshapeTransform(fusion)) {
+    return false;
+  }
   // Make sure the view doesn't interfere with how we'll want to schedule
   // it. If we might want to do a 3D scheduler make sure views are disjoint
   // based on what the 3D scheduler's merges would be.
@@ -415,7 +433,6 @@ bool reductionInterferingView(
   };
 
   std::vector<IterDomain*> dims = reduction_reference->getLogicalDomain();
-  reduction_reference->printTransforms();
 
   // The disjoint groups we need for this scheduler
   std::vector<std::vector<IterDomain*>> groups;
@@ -443,11 +460,6 @@ bool reductionInterferingView(
       processed.emplace(dims[dim_i]);
     }
 
-    std::cout << "========= dimension: " << dimension << " =========\n current_dims:" << std::endl;
-    for (auto id : current_dims) {
-      std::cout << id->toString() << std::endl;
-    }
-
     // Don't add empty group (would happen if it's a 2D scheduler not 3D)
     if (!current_dims.empty()) {
       groups.push_back(current_dims);
@@ -462,14 +474,6 @@ bool reductionInterferingView(
   auto disjoint_rfactor_sets = scheduler_utils::disjointLogicalSets(fusion);
   auto disjoint_set_information = scheduler_utils::getDisjointLogicalSetsOf(
       fusion, reduction_reference, disjoint_rfactor_sets);
-
-  std::cout << "disjointLogicalSets:\n" << disjoint_rfactor_sets.toString() << std::endl;
-  std::cout << "disjoint_set_ids: ";
-  for(auto id : disjoint_set_information.disjoint_set_ids) {
-    std::cout << id << " ";
-  }
-  std::cout << std::endl;
-
 
   // Convert id's in groups to disjoint_set_ids of disjoint_set_information
   std::vector<std::vector<int64_t>> disjoint_groups;
@@ -492,15 +496,9 @@ bool reductionInterferingView(
               (int64_t)disjoint_set_information.disjoint_set_ids.size(),
           "Error computing disjoint group on the logical domain of ",
           reduction_reference->toString());
-      std::cout << "id: " << id->toString() << ", logical_pos: " << logical_pos << std::endl;
       disjoint_id_sets.push_back(
           disjoint_set_information.disjoint_set_ids[logical_pos]);
     }
-    std::cout << "disjoint_id_sets: ";
-    for(auto id : disjoint_id_sets) {
-      std::cout << id << " ";
-    }
-    std::cout << std::endl;
     disjoint_groups.push_back(disjoint_id_sets);
   }
 
@@ -508,7 +506,6 @@ bool reductionInterferingView(
   // will interfere with the schedule. TODO: Make this better complexity,
   // since it should be relatively small int vectors of a small total nDims,
   // not too worried about it now.
-  return false;
 
   for (auto first_dim_i : c10::irange(disjoint_groups.size())) {
     for (auto second_dim_i = first_dim_i + 1;
--- a/tests/cpp/test_gpu_view.cpp
+++ b/tests/cpp/test_gpu_view.cpp
@@ -2317,196 +2317,58 @@ TEST_F(GpuViewTest, ExpandedBroadcast) {
   testValidate(&fusion, {actual_out_tensor}, {in_tensor}, __LINE__, __FILE__);
 }
 
-// segmented into 3 kernels: pointwise, normalization, and pointwise
-TEST_F(GpuViewTest, GroupNormOriginal) {
-  auto fusion = std::make_unique<Fusion>();
-  FusionGuard fg(fusion.get());
-  const int64_t N = 2, C = 128, H = 16, W = 16, G = 32;
-  const std::vector<int64_t> input_shape = {N, C, H, W};
-  const std::vector<int64_t> group_shape = {N, G, C / G, H, W};
-  const std::vector<int64_t> input_shape_wb = {C};
-  const std::vector<int64_t> group_shape_wb = {G, C / G};
-  DataType dtype = DataType::Half;
-  auto tv0 = makeContigTensor(input_shape.size(), dtype);
-  auto tv1 = makeContigTensor(input_shape_wb.size(), DataType::Float);
-  auto tv2 = makeContigTensor(input_shape_wb.size(), DataType::Float);
-  fusion->addInput(tv0);
-  fusion->addInput(tv1);
-  fusion->addInput(tv2);
-  // pointwise ops, e.g. cast
-  auto tv3 = castOp(DataType::Float, tv0);
-  // reshape from {N, C, H, W} to {N, G, C / G, H, W}
-  auto tv4 = reshape(tv3, input_shape, group_shape);
-  // normalization
-  auto tv5 = sum(tv4, {-1, -2, -3});
-  auto tv6 = broadcast(tv5, {false, false, true, true, true});
-  auto tv7 = div(tv4, tv6);
-  // reshape back to {N, C, H, W}
-  auto tv8 = reshape(tv7, group_shape, input_shape);
-  // pointwise ops, e.g. scale, bias, cast
-  auto tv9 = broadcast(tv1, {true, false, true, true});
-  auto tv10 = broadcast(tv2, {true, false, true, true});
-  auto tv11 = mul(tv8, tv9);
-  auto tv12 = add(tv11, tv10);
-  auto tv13 = castOp(dtype, tv12);
-  fusion->addOutput(tv13);
+TEST_F(GpuViewTest, ReductionReshapeInputNoMergedIds) {
+  auto test = [](const std::vector<int64_t> reduction_axes) {
+    auto fusion = std::make_unique<Fusion>();
+    FusionGuard fg(fusion.get());
+    const int64_t N = 2, C = 128, H = 16, W = 16, G = 32;
+    const std::vector<int64_t> input_shape = {N, C, H, W};
+    const std::vector<int64_t> group_shape = {N, G, C / G, H, W};
+    DataType dtype = DataType::Half;
+    auto tv0 = makeContigTensor(input_shape.size(), dtype);
+    fusion->addInput(tv0);
+    auto tv1 = castOp(DataType::Float, tv0);
+    auto tv2 = reshape(tv1, input_shape, group_shape);
+    auto tv3 = sum(tv2, {reduction_axes});
+    fusion->addOutput(tv3);
 
-  auto options =
-      at::TensorOptions().dtype(data_type_to_aten(dtype)).device(at::kCUDA, 0);
-  auto options_wb = at::TensorOptions()
-                        .dtype(data_type_to_aten(DataType::Float))
-                        .device(at::kCUDA, 0);
-  auto t0 = at::randn(input_shape, options);
-  auto tw = at::randn(input_shape_wb, options_wb);
-  auto tb = at::randn(input_shape_wb, options_wb);
-  auto t1 = t0.reshape(group_shape).to(at::kFloat);
-  auto t2 = t1.sum({-1, -2, -3}).unsqueeze(-1).unsqueeze(-1).unsqueeze(-1);
-  auto t3 = t1 / t2;
-  auto t4 = t3.reshape(input_shape);
-  auto t5 = tw.unsqueeze(0).unsqueeze(-1).unsqueeze(-1);
-  auto t6 = tb.unsqueeze(0).unsqueeze(-1).unsqueeze(-1);
-  auto t7 = t4.mul(t5).add(t6);
+    auto options = at::TensorOptions()
+                       .dtype(data_type_to_aten(dtype))
+                       .device(at::kCUDA, 0);
+    auto t0 = at::randn(input_shape, options);
+    auto t1 = t0.reshape(group_shape).to(at::kFloat);
+    auto ref = t1.sum({reduction_axes});
 
-  FusionExecutorCache executor_cache(std::move(fusion));
-  auto cg_outputs = executor_cache.runFusionWithInputs({t0, tw, tb});
-  // should expect 1 after adding a pre-segment pass to move reshape to input
-  // and output.
-  auto seg_groups =
-      executor_cache.getMostRecentKernelRuntime()->fusionSegments()->groups();
-  int n_pointwise = 0, n_norm = 0, n_other = 0;
-  for (auto sg : seg_groups) {
-    if (sg->heuristic() == ScheduleHeuristic::PointWise) {
-      n_pointwise++;
-    } else if (sg->heuristic() == ScheduleHeuristic::InnerPersistent) {
-      n_norm++;
-    } else {
-      n_other++;
-    }
+    FusionExecutorCache executor_cache(std::move(fusion));
+    auto cg_outputs = executor_cache.runFusionWithInputs({t0});
+    // should have only 1 segment group
+    auto seg_groups =
+        executor_cache.getMostRecentKernelRuntime()->fusionSegments()->groups();
+    EXPECT_EQ(seg_groups.size(), 1);
+    testValidate(
+        executor_cache.fusion(), cg_outputs, {t0}, {ref}, __LINE__, __FILE__);
+  };
+  std::vector<std::vector<int64_t>> test_cases;
+  // dims of reduction input, {N, G, C / G, H, W}
+  //                           0, 1, 2,     3, 4
+  // outer reductions
+  test_cases.push_back({0});
+  test_cases.push_back({0, 1});
+  test_cases.push_back({0, 1, 2});
+  // inner reductions
+  test_cases.push_back({-1});
+  test_cases.push_back({-1, -2});
+  test_cases.push_back({-1, -2, -3});
+  // 3D reductions
+  test_cases.push_back({0, 2, 4});
+  // outer reduction with non-neighbor axes
+  test_cases.push_back({1, 3});
+  for (const auto& reduction_axes : test_cases) {
+    test(reduction_axes);
   }
-  EXPECT_EQ(n_pointwise, 2);
-  EXPECT_EQ(n_norm, 1);
-  EXPECT_EQ(n_other, 0);
-  EXPECT_TRUE(at::allclose(cg_outputs[0].to(at::kFloat), t7, 1e-5, 0.01))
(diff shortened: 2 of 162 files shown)


liqiangxl:
!build

liqiangxl:
!build

csrc/scheduler/registry_utils.cpp
```diff
@@ -389,12 +389,32 @@ bool requiresForwardViewReplay(Fusion* fusion, ComputeAtMap& ca_map) {
...
+      }
+    }
+  }
+  return true;
+}
+} // namespace
+
 // Returns if view interferes with how we want to treat the reference, being at
 // least a 2D reduction schedule but maybe a 3D reduction schedule.
 bool reductionInterferingView(
```
wujingyue, line 410: (Not introduced by this PR)

I wish https://github.com/NVIDIA/Fuser/blob/9c5f006ef99043f7238450c3a96a287687d5144d/csrc/scheduler/reduction.cpp#L926 could be more specific about which View interferes reduction, for debugging. This would require refactoring reductionInterferingView and isSplitOnlyReshapeTransform to generate the error message and propagate it up.

csrc/scheduler/registry_utils.cpp
```diff
@@ -389,12 +389,32 @@ bool requiresForwardViewReplay(Fusion* fusion, ComputeAtMap& ca_map) {
   return false;
 }
 
+namespace {
+bool isSplitOnlyReshapeTransform(Fusion* fusion) {
```
wujingyue, line 393: It may be slightly better to have
```
bool isSplitOnly(ViewOp* view);
```
and change line 415 to something like
```
const auto& view_ops = ir_utils::getViewOps(fusion);
if (std::all_of(view_ops.begin(), view_ops.end(), [](ViewOp* view) { return isSplitOnly(); })) {
  return false;
}
```
This would make reductionInterferingView slightly more complicated than the current PR, but the intention of checking all reshapes would be immediately clear to readers.

tests/cpp/test_gpu_view.cpp
```diff
@@ -2317,4 +2317,89 @@ TEST_F(GpuViewTest, ExpandedBroadcast) {
...
+  // inner reductions
+  test_cases.push_back({-1});
+  test_cases.push_back({-1, -2});
+  test_cases.push_back({-1, -2, -3});
+  // 3D reductions
+  test_cases.push_back({0, 2, 4});
+  // outer reduction with non-neighbor axes
+  test_cases.push_back({1, 3});
+  for (const auto& reduction_axes : test_cases) {
+    test(reduction_axes);
```
wujingyue, line 2367: I recommend we TEST_P instead of looping over test cases in a TEST_F. https://google.github.io/googletest/advanced.html#value-parameterized-tests lists several benefits of the former, FYI.

wujingyue (commented):
Drive-by comments

csrc/scheduler/registry_utils.cpp
liqiangxl: thanks for the suggestion. revised.

tests/cpp/test_gpu_view.cpp
liqiangxl: changed to TEST_P

csrc/scheduler/registry_utils.cpp
liqiangxl: will add in a seperate PR.

tests/cpp/test_gpu_view.cpp
```diff
@@ -2317,4 +2317,86 @@ TEST_F(GpuViewTest, ExpandedBroadcast) {
...
+  DataType dtype = DataType::Half;
+  auto tv0 = makeContigTensor(input_shape.size(), dtype);
+  fusion->addInput(tv0);
+  auto tv1 = castOp(DataType::Float, tv0);
+  auto tv2 = reshape(tv1, input_shape, group_shape);
+  auto tv3 = sum(tv2, {-1, -2, -3});
+  auto tv4 = broadcast(tv3, {false, false, true, true, true});
+  auto tv5 = div(tv2, tv4);
+  auto tv6 = castOp(dtype, tv5);
+  fusion->addOutput(tv6);
```
jjsjann123, line 2383: qq: Is the added broadcast/div/cast supposed to work with an arbitrary reduction axes? I'm wondering why we have two sets of test instead of having this pattern in the above test suite.

jjsjann123 (approved):
LGTM. minor question on tests

tests/cpp/test_gpu_view.cpp
```diff
@@ -2317,4 +2317,86 @@ TEST_F(GpuViewTest, ExpandedBroadcast) {
   testValidate(&fusion, {actual_out_tensor}, {in_tensor}, __LINE__, __FILE__);
 }
 
+using ViewReductionParamType = std::vector<int64_t>;
+class ViewReductionParametrizedTest
```
wujingyue, line 2321: Not a change request, just FYI: you can use the same parameterized test fixture for both TEST_F and TEST_P. E.g. https://github.com/NVIDIA/Fuser/blob/7bc7a08dac0dc7473bc34fd2280b918d82e39ae8/tests/cpp/test_memory.cpp#L45 and https://github.com/NVIDIA/Fuser/blob/7bc7a08dac0dc7473bc34fd2280b918d82e39ae8/tests/cpp/test_memory.cpp#L117

However, in this case, the parameterized test fixture is specific to view reduction, so the change as is is actually better.

wujingyue, line 2321:
```suggestion
class ViewReductionTest
```

tests/cpp/test_gpu_view.cpp
```diff
@@ -2317,4 +2317,86 @@ TEST_F(GpuViewTest, ExpandedBroadcast) {
...
+  // should have only 1 segment group
+  auto seg_groups =
+      executor_cache.getMostRecentKernelRuntime()->fusionSegments()->groups();
+  EXPECT_EQ(seg_groups.size(), 1);
+  testValidate(
+      executor_cache.fusion(), cg_outputs, {t0}, {ref}, __LINE__, __FILE__);
+}
+
+INSTANTIATE_TEST_SUITE_P(
+    ViewReduction,
```
wujingyue, line 2356:
```suggestion
    ,
```

tests/cpp/test_gpu_view.cpp
```diff
@@ -2317,4 +2317,86 @@ TEST_F(GpuViewTest, ExpandedBroadcast) {
   testValidate(&fusion, {actual_out_tensor}, {in_tensor}, __LINE__, __FILE__);
 }
 
+using ViewReductionParamType = std::vector<int64_t>;
```
wujingyue, line 2320:
```suggestion
using Shape = std::vector<int64_t>;
```

wujingyue (approved):

tests/cpp/test_gpu_view.cpp
liqiangxl: changed.

tests/cpp/test_gpu_view.cpp
liqiangxl: changed.

tests/cpp/test_gpu_view.cpp
liqiangxl: changed to `ReductionAxes`

tests/cpp/test_gpu_view.cpp
liqiangxl: refactored to a suite `class ViewNormalizationTest : public NVFuserFixtureParamTest<ReductionAxes> {};`

liqiangxl:
@naoyam `reductionInterferingView` is changed to use `IdModel`, first time to use it, not sure if it is the correct way to use it. can you take a look? Thanks!

csrc/scheduler/registry_utils.cpp
```diff
@@ -449,53 +449,36 @@ bool reductionInterferingView(
...
-    for (auto second_dim_i = first_dim_i + 1;
-         second_dim_i < disjoint_groups.size();
-         ++second_dim_i) {
-      auto first_group = disjoint_groups[first_dim_i];
-      auto second_group = disjoint_groups[second_dim_i];
-      for (auto first_disjoint_id : first_group) {
-        for (auto second_disjoint_id : second_group) {
-          if (first_disjoint_id == second_disjoint_id) {
+  // Loop through all view ops and check if they merge IDs in different groups
+  // Map through: ID --> ValGroup --> Group index
```
jjsjann123, line 468: I think I understand it as that, each group would be a *merged* id for scheduling to handle, so merge on IDs across group would break scheduler later.

Maybe it's just me. I'm wondering if we should put an explicit comment on that here. Since it's easy to mistakenly associate `group_idx` with ValGroup and get very confused (like I was earlier)?

csrc/scheduler/registry_utils.cpp
```diff
@@ -449,53 +449,36 @@ bool reductionInterferingView(
...
+  IdModel id_model(fusion, false, false, false);
+  id_model.buildExactGraph();
+  const auto& id_use_map = id_model.idUses();
+  const ValGraph& exact_graph = id_model.idGraph(IdMappingMode::EXACT);
+
+  // ValGroup to group index
+  std::unordered_map<ValGroup, int> vg_to_group_idx;
+  for (auto group_id : c10::irange(groups.size())) {
+    for (auto id : groups.at(group_id)) {
+      auto vg = exact_graph.toGroup(id);
```
naoyam, line 462: nit: 
```suggestion
      const auto& vg = exact_graph.toGroup(id);
```

csrc/scheduler/registry_utils.cpp
```diff
@@ -449,53 +449,36 @@ bool reductionInterferingView(
...
+    for (auto id : view->out()->getMaybeRootDomain()) {
+      auto uses_it = id_use_map.find(id);
+      if (uses_it == id_use_map.end() || uses_it->second.empty()) {
+        continue;
+      }
+      for (auto expr : uses_it->second) {
+        if (auto merge = dynamic_cast<Merge*>(expr)) {
+          const auto& vg0 = exact_graph.toGroup(merge->inner());
+          const auto& vg1 = exact_graph.toGroup(merge->outer());
+          if (vg_to_group_idx.count(vg0) && vg_to_group_idx.count(vg1) &&
```
naoyam, line 480: IIUC, `vg_to_group_idx` only has mapping for the exact groups of the logical domains of the reduction tensor, i.e., those in `groups`. Would it be enough? For example, suppose there's a merge of two domains that are both mapped to reduction domains of the reduction tensor, but then there's also another merge of the output of the first merge and another domain from the reduction tensor but non reduction, which should be considered interfering, but it seems that's not going to be detected as the output of the first merge doesn't have a mapping in `vg_to_group_idx`.

csrc/scheduler/registry_utils.cpp
liqiangxl: Good point. renamed it to `vect_of_coalesced_ids` and added comments.
```
  // Each vector of IDs is a group of coalesced_ids, they will be merged into
  // one ID during the reduction schedule. If the view op involves merging IDs
  // in different group of coalesced_ids, it would break the reduction
  // scheduler.
  std::vector<std::vector<IterDomain*>> vect_of_coalesced_ids;
```

csrc/scheduler/registry_utils.cpp
liqiangxl: changed

csrc/scheduler/registry_utils.cpp
liqiangxl: You are right. CI detected similar errors and I fixed them by adding newly created IDs from root to logical domains to `vg_to_group_idx`.

liqiangxl:
!build --diff

liqiangxl:
!build --diff

liqiangxl:
@naoyam code diff in `nvfuser-ci/jit_codegen_diff_17_4/4 — Failing (unknown) https://nv/e2E/99868931` is expected since reshape + reduction/normalization is now merged into one kernel instead of two kernels.

csrc/scheduler/registry_utils.cpp
```diff
@@ -442,63 +442,77 @@ bool reductionInterferingView(
...
+    for (auto expr : StmtSort::getExprsTo(
+             {tv->getLogicalDomain().begin(), tv->getLogicalDomain().end()})) {
+      if (auto merge = dynamic_cast<Merge*>(expr)) {
+        const auto& vg0 = exact_graph.toGroup(merge->inner());
+        const auto& vg1 = exact_graph.toGroup(merge->outer());
+        auto it0 = vg_to_coalesced_ids.find(vg0);
+        auto it1 = vg_to_coalesced_ids.find(vg1);
+        if (it0 == vg_to_coalesced_ids.end() ||
+            it1 == vg_to_coalesced_ids.end()) {
+          continue;
```
naoyam, line 487: Is it ok to just ignore this case?

csrc/scheduler/registry_utils.cpp
```diff
@@ -442,63 +442,77 @@ bool reductionInterferingView(
...
+          const auto& vg = exact_graph.toGroup(merge->out());
+          vg_to_coalesced_ids[vg] = it0->second;
+        }
+      } else if (auto split = dynamic_cast<Split*>(expr)) {
+        const auto& vg0 = exact_graph.toGroup(split->in());
+        const auto& vg1 = exact_graph.toGroup(split->outer());
+        const auto& vg2 = exact_graph.toGroup(split->inner());
+        auto it = vg_to_coalesced_ids.find(vg0);
+        if (it == vg_to_coalesced_ids.end()) {
+          continue;
```
naoyam, line 504: same here

csrc/scheduler/registry_utils.cpp
```diff
@@ -442,63 +442,77 @@ bool reductionInterferingView(
...
-         second_dim_i < disjoint_groups.size();
-         ++second_dim_i) {
-      auto first_group = disjoint_groups[first_dim_i];
-      auto second_group = disjoint_groups[second_dim_i];
-      for (auto first_disjoint_id : first_group) {
-        for (auto second_disjoint_id : second_group) {
-          if (first_disjoint_id == second_disjoint_id) {
-            return true;
-          }
+  // (3) Loop through all view ops and check if they merge IDs in different
```
naoyam, line 469: I'm not sure this would work, for example, when there are multiple reshape ops. Suppose the first reshape merges two disjoint groups, so that should be flagged as invalid, but if its inputs have a dependency to the output of another reshape, the second reshape needs to be visited first, otherwise at the time the first reshape is visited, there's no mapping for its inputs, so the condition of the `continue` cases that I commented about would hit.

csrc/scheduler/registry_utils.cpp
```diff
@@ -442,63 +442,77 @@ bool reductionInterferingView(
...
+  // coalesced_ids. If happens, return true.
+  for (auto view : ir_utils::getViewOps(fusion)) {
+    auto tv = view->out();
+    // visit exprs between root and logical domains in topologically sorted
+    // order (can't use id_model.idUses() since it is not sorted).
```
naoyam, line 474: Can you elaborate what you mean here? `idUses` just holds all uses of each ID. What sorting are you thinking about?

csrc/scheduler/registry_utils.cpp
liqiangxl: viewOps are visited in topological order (`getViewOps` uses `fusion->exprs()` which `Return a list of topologically sorted expressions.`), so this shouldn't happen. I added a tests case.

csrc/scheduler/registry_utils.cpp
liqiangxl: Topological order. Extended the comments:
```
    // visit exprs between root and logical domains in topologically sorted
    // order (can't use id_model.idUses() since it is not sorted). This ensures
    // the exprs are visited in the order they are defined, e.g. when we have
    // A -> B -> C, we visit in order A, B, C instead of C, B, A or B, C, A.
```

csrc/scheduler/registry_utils.cpp
liqiangxl: If I do a backward visit of the collected `viewOps`, I do see the error you mensioned.

csrc/scheduler/registry_utils.cpp
liqiangxl: This and the similar one for split only happens when the reshape is before reduction. If reshape is after reduction, its `root domain` is same to reduction tv's logical domain, so all the IDs constructed from root exists in `vg_to_coalesced_ids`.
If reshape is before reduction, we can actually skip all the checks since the reduction tv should have the same logical domain as the reshape output tv. 
I did the following changes:
```
(1) skip all view ops before reduction
(2) change from `continue` to `NVF_ERROR`
```

liqiangxl:
!build

csrc/scheduler/registry_utils.cpp
```diff
@@ -442,63 +462,83 @@ bool reductionInterferingView(
...
-        for (auto second_disjoint_id : second_group) {
-          if (first_disjoint_id == second_disjoint_id) {
-            return true;
-          }
+  // (3) Loop through all view ops and check if they merge IDs in different
+  // coalesced_ids. If happens, return true.
+  for (auto view : view_ops_after_reduction) {
+    auto tv = view->out();
+    // visit exprs between root and logical domains in topologically sorted
+    // order (can't use id_model.idUses() since it is not sorted). This ensures
```
naoyam, line 493: This still sounds strange to me. `idUses` just holds all uses of each domain, and it isn't designed to solve problems like this ordering.

csrc/scheduler/registry_utils.cpp
```diff
@@ -391,14 +391,31 @@ bool requiresForwardViewReplay(Fusion* fusion, ComputeAtMap& ca_map) {
...
-bool reductionInterferingView(
-    Fusion* fusion,
-    const ComputeAtMap& ca_map,
-    TensorView* reduction_reference) {
+bool reductionInterferingView(Fusion* fusion, TensorView* reduction_reference) {
   // Make sure the view doesn't interfere with how we'll want to schedule
   // it. If we might want to do a 3D scheduler make sure views are disjoint
   // based on what the 3D scheduler's merges would be.
 
+  // don't need to check view ops before reduction, e.g. T1 = view(T0),
```
naoyam, line 399: Suppose there're two view ops before a reduction, e.g.:

```
t0: [I0]
// I0 -> I1, I2
t1 = reshape(t0, [I1, I2]);
// I2 -> I3, I4
t2 = reshape(t1, [I1, I3, I4]);
t3 = sum(t2, {-1);
```

And also there's another reshape using `t1`:

```
// [I1, I2] -> I5
t4 = reshape(t1, [I5])
```

The first two reshapes would be just ignored as they are before the reduction. The last reshape would be checked but since nothing would be known about `I2`, it would also be just ignored. However, the last reshape should be considered conflicting.

Am I understanding correctly?

liqiangxl: (1) `requiresForwardViewReplay` captures this case (two uses of I2, `I2 -> I3, I4` and `[I1, I2] -> I5`) before doing `reductionInterferingView`.
Added a test:
```
  auto tv2 = reshape(tv1, {N}, {16, 8});
  auto tv3 = reshape(tv2, {16, 8}, {16, 4, 2});
  auto tv4 = sum(tv3, {-1});
  // ID {8} was used in split, {8} --> {4, 2}
  // Now it is used again in merge, {16} * {8} --> {N}
  // requiresForwardViewReplay() captures this non-consistant uses.
  auto tv5 = reshape(tv2, {16, 8}, {N});
```
(2) slightly modified `reductionInterferingView` to check `view_ops_not_used_by_reduction` to ensure sibling view ops of reduction ops are checked. e.g. (similar to the previous case but changed `tv5 = reshape(tv2)` to `tv5 = reshape(tv3)` )
```
  auto tv2 = reshape(tv1, {N}, {16, 8});
  auto tv3 = reshape(tv2, {16, 8}, {16, 4, 2});
  auto tv4 = sum(tv3, {-1});
  // This reshape merges an Iter dim and a reduction dim
  auto tv5 = reshape(tv3, {16, 4, 2}, {N});
```

csrc/scheduler/registry_utils.cpp
liqiangxl: Got you. removed `idUses` from the comment.
```
    // visit exprs between root and logical domains in topologically sorted
    // order. add the newly created IDs to the same coalesced_ids as the input
    // IDs.
```

csrc/scheduler/registry_utils.cpp
liqiangxl: It seems difficult to mathmatical prove that this method covers all the cases. But if some unsupported corner case happens, it fails early and loudly since `NVF_ERROR` is used to ensure all the IDs exist in `vg_to_coalesced_ids` instead of silently `continue`

naoyam (approved):
LGTM

liqiangxl:
!build

liqiangxl:
!build

liqiangxl:
Validation error in `NVFuserTest.CombinedSchedulerLayerNormBackward_CUDA` is not related to this PR and is fixed in #2561

liqiangxl:
!build

liqiangxl:
!build

csrc/scheduler/registry_utils.cpp
```diff
@@ -442,60 +442,62 @@ bool reductionInterferingView(
...
-          disjoint_set_information.disjoint_set_ids[logical_pos]);
+  // Check if any view op could merge IDs in different coalesced_ids.
+  // (1) Build the exact graph
+  IdModel id_model(fusion, false, false, false);
+  id_model.buildExactGraph();
+  const ValGraph& graph = id_model.idGraph(IdMappingMode::EXACT);
+  auto disjoint_sets = graph.disjointValSets();
+  for (auto view : ir_utils::getViewOps(fusion)) {
+    auto tv = view->out();
+    std::vector<IterDomain*> ids_to_merge;
```
naoyam, line 460: Please add a comment about this vector.

csrc/scheduler/registry_utils.cpp
```diff
@@ -442,60 +442,62 @@ bool reductionInterferingView(
...
+      // For split, directly map the inner and outer IDs leads to false
+      // positive e.g. the frist reshape in group norm, since only merge op can
+      // cause the merge of IDs in different coalesced_ids. However, we need to
+      // keep track of the split IDs and map them with merged IDs when a merge
+      // op is found. Otherwise, we may miss a case like: reshape: [I2, I3] ->
+      // [8, I2/8, I3] -> [8, I2/8*I3], reduction: [I2, rI3]. This view +
+      // reduction is not supported since I2/8*I3 includes both iter and redu
+      // dims. If we don't process split, I2/8 is not mapped with I3.
+      else if (auto split = dynamic_cast<Split*>(expr)) {
+        ids_to_merge.push_back(split->inner());
```
naoyam, line 481: Looks like this part is just accumulating all split domains, and they are merged with the domains of the first merge op. However, these split input and output domains are not guaranteed to have dependencies with the merge input and output domains, right? Aren't we making false groupings as well as missing true groupings?

liqiangxl: You are right. Can we keep track of the split IDs and when a merged id comes from a set of connected splits, map it with all other IDs in the same set?
```
    // keep track of split IDs, each unordered_set in the vector stores
    // connected split IDs. For example, if we have 3 splits:
    // I0 -> I1, I2
    // I3 -> I4, I5
    // I5 -> I6, I7
    // There are 2 groups: [I0, I1, I2] and [I3, I4, I5, I6, I7]
    // when a merge op is found, e.g. I9 = I8 * I7, we need to map
    // I9 with I8 and all the IDs in [I3, I4, I5, I6, I7].
    std::vector<std::unordered_set<IterDomain*>> ids_in_connected_splits;
```

naoyam: I think there are multiple valid ways and I'm not sure what the best way would be. I don't have any particular way in my mind, but what we have is a graph, and what we need to know is if a certain set of graph nodes are disjoint. Traversal routines in val_graph_visitor.h may help.

liqiangxl:
!build

csrc/scheduler/registry_utils.cpp
```diff
@@ -442,60 +442,88 @@ bool reductionInterferingView(
...
+        disjoint_sets.mapEntries(id1, merge->out());
+        disjoint_sets.mapEntries(id2, merge->out());
+        // if merged id comes from a set of connected splits, map it with all
+        // other IDs in the set
+        for (auto merged_id : {id1, id2}) {
+          auto it = std::find_if(
+              ids_in_connected_splits.begin(),
+              ids_in_connected_splits.end(),
+              [&merged_id](const auto& id_set) {
+                return id_set.count(merged_id);
```
naoyam, line 483: I don't think you can just search by IterDomain. Remember we are traversing a graph where each node is a group of iter domains.

liqiangxl: A little confused here. we are traversing from a tv's root domain to its logical domain using `StmtSort::getExprsTo(
             {tv->getLogicalDomain().begin(), tv->getLogicalDomain().end()})`. So each node is an iter domain.
For example, a fusion with
```
  auto tv2 = reshape(tv1, {12, 20 ,3}, {3, 4, 5, 2, 6});
  auto tv3 = sum(tv1, {-1});
```
From tv2's root to its logical domain, we have:
image
In this graph, each node is an iter domain.
```
 root domain : (iS9{i0}rf, iS12{i2}rf, iS17{i3}rf)
  Outer split: iS9{i0}rf by factor 3 -> iS10{3}rf, iS11{( ceilDiv(i0, 3) )}rf
  Outer split: iS12{i2}rf by factor 5 -> iS13{5}rf, iS14{( ceilDiv(i2, 5) )}rf
  Outer split: iS14{( ceilDiv(i2, 5) )}rf by factor 2 -> iS15{2}rf, iS16{( ceilDiv(( ceilDiv(i2, 5) ), 2) )}rf
  Merge: iS16{( ceilDiv(( ceilDiv(i2, 5) ), 2) )}rf and iS17{i3}rf -> iS18{( ( ceilDiv(( ceilDiv(i2, 5) ), 2) ) * i3 )}rf
 logical domain : (iS10{3}rf, iS11{( ceilDiv(i0, 3) )}rf, iS13{5}rf, iS15{2}rf, iS18{( ( ceilDiv(( ceilDiv(i2, 5) ), 2) ) * i3 )}rf)
```
After visiting the first 3 splits, `ids_in_connected_splits` has 2 sets:
```
set1 from i0: iS10{3}rf iS11{( ceilDiv(i0, 3) )}rf iS9{i0}rf 
set2 from i2: iS15{2}rf iS16{( ceilDiv(( ceilDiv(i2, 5) ), 2) )}rf iS13{5}rf iS14{( ceilDiv(i2, 5) )}rf iS12{i2}rf 
```
When visit the merge expr, this line of code will find `iS16{( ceilDiv(( ceilDiv(i2, 5) ), 2) )}rf` from `set2` in `ids_in_connected_splits`. Then, all other IDs in `set2` are mapped with output of the merge expr.

liqiangxl: oh, great point. I addd 3 additional tests testing multiple reshape ops. Revised to use ValGroup.
```
  // Here we can't directly find IterDomain since they may be in different
  // reshape ops. ValGroup groups IDs in different tvs.
  auto findIdSet = [&](IterDomain* id) {
    auto id_vgroup = graph.toGroup(id);
    return std::find_if(
        ids_in_connected_splits.begin(),
        ids_in_connected_splits.end(),
        [&](const std::unordered_set<IterDomain*>& id_set) {
          return std::any_of(
              id_set.begin(), id_set.end(), [&](IterDomain* set_id) {
                return graph.toGroup(set_id) == id_vgroup;
              });
        });
  };
```

liqiangxl: For example when visit the merge of `i2/5/2` in `reshape-2`, the code is able to find `i2/5` and `i2` in `reshape-1`
image

naoyam: Originally you had code to filter out reshapes before reductions. Don't we still want to do that?

liqiangxl: we can add that and also skip the following checks if there is no merge detetected in any of the reshapes. Other than that, does the current method looks good?

naoyam: Yes, let's add it back.

liqiangxl: Thanks! Added.

liqiangxl:
!build

csrc/scheduler/registry_utils.cpp
naoyam: This won't work if there are reshapes both before and after reductions, right?

liqiangxl: I think it still works fine. For example, this fusion has reshape before and after reduction,
image
In this fusion, the first reshape is not processed since it is used by the reduction. When processing the merge op in the 2nd reshape, `i2/5/2` is mapped to `i2/5` and all other `i2/5` domains from `t3, t4, and t5`. The code can still detect the merge of an iteration domain and a reduction domain. Do  you have an example where the code may fail? The tests include several other cases but all work fine.

naoyam: How about this?

```
auto tv0 = [I0, I1]; // input
auto tv1 = reshape [I0, I1] -> [I0/X, X*I1/Y, Y]
auto tv2 = sum(tv1, {1}) // output

auto tv3 = reshape [I0, I1] -> [I0/X, X*I1/Y, Y] // output
```

The reshape for `tv3` is unused by the reduction. The split-followed-by-merge pattern makes all domains grouped together, thus this pattern will be considered interfering. But should it be?

liqiangxl: You are right. This fusion shouldn't be rejected by `reductionInterferingView`. The output of the merge op and one of the reduction tv's root ID are in the same valGroup. It woun't cause the merge of an iter dim with a redu dim when replayed on the reduction tv.
To avoid this, the algorithm should be modified to avoid mapping the merged Ids if its output and one of the reduction tv's root ID are in the same valGroup.

**Code change**
```
  // val group for reduction tv's root IDs
  std::unordered_set<ValGroup> ref_root_groups;
  for (auto id : reduction_reference->getMaybeRootDomain()) {
    ref_root_groups.insert(graph.toGroup(id));
  }

      if (auto merge = dynamic_cast<Merge*>(expr)) {
        // if the output of the merge op and one of the reduction tv's root ID
        // are in the same valGroup. It won't cause the merge of an iter dim
        // with a redu dim when replayed on the reduction tv. We can skip this
        // transform, further transforms using this ID can still be captured and
        // mapped since it it a root ID of the reduction tv. See tests
        // GpuViewTest.SplitMergeReductionSplitMerge and
        // GpuViewTest.SplitMergeReductionSplitMergeSplitMerge
        if (ref_root_groups.count(graph.toGroup(merge->out())) > 0) {
          continue;
        }
```
**New Tests**
(1) GpuViewTest.SplitMergeReductionSplitMerge
```
  // root domain : (i0, i2)
  // logi domain : (3, i0/3, 4, i2/4)
  auto tv2 = reshape(tv1, {12, 20}, {3, 4, 4, 5});
  // root domain : (3, i0/3, 4, i2/4)
  // logi domain : (3, i0/3*4, i2/4)
  auto tv3 = reshape(tv2, {3, 4, 4, 5}, {3, 16, 5});
  // root domain : (3, i0/3*4, i2/4)
  auto tv4 = sum(tv3, {1});
  // root domain : (i0, i2)
  // logi domain : (3, i0/3, 4, i2/4)
  auto tv5 = reshape(tv1, {12, 20}, {3, 4, 4, 5});
  // root domain : (3, i0/3, 4, i2/4)
  // logi domain : (3, i0/3*4, i2/4)
  // merge transform generates `i0/3*4` which is in the same ValGroup as tv4's
  // root ID `i0/3*4`, it is not interfering reduction.
  auto tv6 = reshape(tv5, {3, 4, 4, 5}, {3, 16, 5});
```
Not rejected by `reductionInterferingView`, however, it triggers a bug, see #2593. Marked as skipped.
(2) GpuViewTest.SplitMergeReductionSplitMergeSplitMerge
Modified from `(1)` by adding another split-merge, rejected by `reductionInterferingView` due to the last merge.
image

liqiangxl:
!build

csrc/scheduler/registry_utils.cpp
naoyam: Did you actually change the code? I don't see anything like `ref_root_groups`.

liqiangxl: oops, local change it not pushed to remote. Updated.
since #2593 is fixed, the tes case is not skipped.

PR #2439, opened 2024-06-18
Update batch size for nightly/weekly runs

Priya2698:
This PR updates the batch size to be the same for nightly/weekly runs.
This will allow us to plot S-curves for the nightly results using the torch baselines from the previous weekly run.
We use `batch size = [16, 512, 2048, 8192, 16384]`


Priya2698:
!build --pybench

Priya2698:
!build --pybench

liqiangxl (approved):
The change looks good to me.
Can we fix these py benchmarks (not in this PR)? IIRC, there is a PR/issue tacking about these validation issues.

PR #2440, opened 2024-06-19
host ir: add support for for-loop

samnordmann:
This PR introduces the support for for-loop used as host IRs.

We provide a unit test that reproduces in host IR the program
```
int64_t buf = kInitialValue;
for (int j = kForLoopStart; j < kForLoopStop; j += kForLoopStep) {
   buf += j;
}
```
where buf is the global output.

We reuse the `kir::ForLoop` to avoid code duplication and to reuse in the future the utils already implemented there. I know there is a conflict in the namespace, and we shouldn't use "kir" in "hir". I don't know what's the best practice here: either "kir::ForLoop" and "hir::ForLoop" should inherit from the same base class, or we both IRs should diverge. I see no fundamental reason to diverge for now, though. @wujingyue wdyt?

### DIFF
--- a/csrc/host_ir/executor.cpp
+++ b/csrc/host_ir/executor.cpp
@@ -149,6 +149,22 @@ void HostIrExecutor::handle(Wait* wait) {
   works_.erase(communication);
 }
 
+void HostIrExecutor::handle(kir::ForLoop* for_loop) {
+  NVF_ERROR(for_loop->start()->isConstInt());
+  NVF_ERROR(for_loop->step()->isConstInt());
+  NVF_ERROR(for_loop->stop()->isConstInt());
+  auto start = for_loop->start()->value().as<int64_t>();
+  auto step = for_loop->step()->value().as<int64_t>();
+  auto stop = for_loop->stop()->value().as<int64_t>();
+
+  for (int64_t i= start; i < stop; i += step) {
+    for (int64_t j : c10::irange(for_loop->body().size())) {
+      val_to_IValue_[for_loop->index()] = at::Scalar(i);
+      dispatch(for_loop->body().at(j));
+    }
+  }
+}
+
 } // namespace hir
 
 } // namespace nvfuser
--- a/tests/cpp/test_host_irs.cpp
+++ b/tests/cpp/test_host_irs.cpp
@@ -377,6 +377,94 @@ TEST_P(HostIrTest, ThreeFusions) {
   GTEST_EXPECT_TRUE(torch::allclose(tv2_2_ref, outputs.at(0)));
 }
 
+// This unit test the for-loop IR by implementing a program that could be summarized as
+//   |  int64_t buf = kInitialValue;
+//   |  for (int j = kForLoopStart; j < kForLoopStop; j += kForLoopStep) {
+//   |    buf += j;
+//   |  }
+// where buf is the ouput.
+TEST_P(HostIrTest, ForLoops) {
+  constexpr int64_t kInitialValue = 21;
+  constexpr int64_t kForLoopStart = 1;
+  constexpr int64_t kForLoopStop = 7;
+  constexpr int64_t kForLoopStep = 2;
+
+  auto hic = std::make_unique<HostIrContainer>();
+  FusionGuard::setCurFusion(hic.get());
+
+  auto index = IrBuilder::create<Val>(DataType::Index);
+  auto start = IrBuilder::create<Val>(kForLoopStart, DataType::Index);
+  auto stop = IrBuilder::create<Val>(kForLoopStop, DataType::Index);
+  auto step = IrBuilder::create<Val>(kForLoopStep, DataType::Index);
+  auto for_loop = IrBuilder::create<kir::ForLoop>(
+      /*IterDomain=*/makeContigConcreteTensor({0})->axis(0), // unused
+      index,
+      start,
+      stop,
+      step,
+      /*vectorize=*/false,
+      /*vectorize_shift=*/nullptr,
+      /*unroll_required=*/false,
+      /*double_buffer_loop_stage=*/DoubleBufferLoopStage::NotApplicable);
+
+  auto fusion = std::make_unique<Fusion>();
+  FusionGuard fg(fusion.get());
+
+  auto arange = iota(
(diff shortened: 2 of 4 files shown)


csrc/host_ir/executor.cpp
```diff
@@ -149,6 +149,22 @@ void HostIrExecutor::handle(Wait* wait) {
   works_.erase(communication);
 }
 
+void HostIrExecutor::handle(kir::ForLoop* for_loop) {
```
wujingyue, line 152: @naoyam wdyt about the namespace? As said in the PR description, `kir` isn't appropriate any more since we'll also use it in host IR.

csrc/host_ir/executor.cpp
```diff
@@ -149,6 +149,22 @@ void HostIrExecutor::handle(Wait* wait) {
...
+  NVF_ERROR(for_loop->start()->isConstInt());
+  NVF_ERROR(for_loop->step()->isConstInt());
+  NVF_ERROR(for_loop->stop()->isConstInt());
+  auto start = for_loop->start()->value().as<int64_t>();
+  auto step = for_loop->step()->value().as<int64_t>();
+  auto stop = for_loop->stop()->value().as<int64_t>();
+
+  for (auto i = start; i < stop; i += step) {
+    for (auto j : c10::irange(for_loop->body().size())) {
```
wujingyue, line 161:
```suggestion
    for (Expr* e : for_loop->body().exprs()) {
```

wujingyue: Because `j` is only used to index expressions, the above is slightly simpler.

wujingyue (commented):
Nice! I'll review the test tomorrow.

csrc/host_ir/executor.cpp
samnordmann: right! I fixed it

samnordmann:
!build

csrc/host_ir/executor.cpp
naoyam: How about just moving it to the top-level namespace? It's in the `kir` namespace because it's only valid in the Kernel IR, but now that if it's used also in the Host IR, placing it inside the `kir` namespace doesn't make sense. Perhaps, we should have a namespace like `nvfuser::ir`, but that's a separate discussion.

tests/cpp/test_host_irs.cpp
```diff
@@ -377,6 +377,95 @@ TEST_P(HostIrTest, ThreeFusions) {
...
+
+  hic->addInput(post_on_stream->inputs().at(1));
+  hic->addOutput(post_on_stream->outputs().at(0));
+  hic->pushBackTopLevelExprs(for_loop);
+
+  HostIrExecutorParams params;
+  auto [use_fusion_executor_cache] = GetParam();
+  if (!use_fusion_executor_cache) {
+    GTEST_SKIP()
+        << "not supported for now because of concretization pb, getting the error: dynamic_tvs.empty() INTERNAL ASSERT FAILED at /opt/pytorch/Fuser/csrc/device_lower/validation.cpp:187, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. Tensor with dynamic transform must be concretized before lowering: T1_l[ ?S2{( ( ( -( fmax(0, ( where(( i7 < 0 ), ( i7 + 10 ), i7) )) ) ) + 10 ) + ( ( fmax(( fmax(0, ( where(( i7 < 0 ), ( i7 + 10 ), i7) )) ), ( fmin(10, ( where(( ( i7 + 1 ) < 0 ), ( ( i7 + 1 ) + 10 ), ( i7 + 1 )) )) )) ) - 10 ) )}rf ], T3_g[ ?S4{( ( ( -( fmax(0, ( where(( i7 < 0 ), ( i7 + 10 ), i7) )) ) ) + 10 ) + ( ( fmax(( fmax(0, ( where(( i7 < 0 ), ( i7 + 10 ), i7) )) ), ( fmin(10, ( where(( ( i7 + 1 ) < 0 ), ( ( i7 + 1 ) + 10 ), ( i7 + 1 )) )) )) ) - 10 ) )} ]";
```
wujingyue, line 448: > concretization pb

What's `pb`?

tests/cpp/test_host_irs.cpp
```diff
@@ -377,6 +377,95 @@ TEST_P(HostIrTest, ThreeFusions) {
...
+  }
+  params.use_fusion_executor_cache = use_fusion_executor_cache;
+  HostIrExecutor hie(std::move(hic), nullptr, params);
```
wujingyue, line 451:
```suggestion
  HostIrExecutor hie(std::move(hic), /*communicator=*/nullptr, params);
```

tests/cpp/test_host_irs.cpp
```diff
@@ -377,6 +377,95 @@ TEST_P(HostIrTest, ThreeFusions) {
...
+//   |  }
+// where buf is the ouput.
+TEST_P(HostIrTest, ForLoops) {
+  constexpr int64_t kInitialValue = 21;
+  constexpr int64_t kForLoopStart = 1;
+  constexpr int64_t kForLoopStop = 7;
+  constexpr int64_t kForLoopStep = 2;
+
+  auto hic = std::make_unique<HostIrContainer>();
+  FusionGuard::setCurFusion(hic.get());
```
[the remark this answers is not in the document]
samnordmann: It totally make sense, the only problem here is that I need to use some of the fusion's TV outside the scope, for example, the line `std::vector<Val*> post_on_stream_inputs = {index, ir_cloner.clone(acc_in)};`. So wrapping it inside another scope would add complexity

tests/cpp/test_host_irs.cpp
samnordmann: that stands for "problem" :)
I replace "pb" by "issue"

csrc/host_ir/executor.cpp
samnordmann: Ok, is it ok if I do that in a separate PR once this one is merged?

samnordmann: done in https://github.com/NVIDIA/Fuser/pull/2453

PR #2441, opened 2024-06-21
Add examples of reusing buffers.

wujingyue:


samnordmann (approved):
thanks!

wujingyue:
> nit suggestion: we could add checks like
> 
> https://github.com/NVIDIA/Fuser/blob/9c5f006ef99043f7238450c3a96a287687d5144d/tests/cpp/test_alias.cpp#L1247
> 
> 
> to ensure the buffers are actually reused

That's definitely a valid concern. In practice, 
1. FusionExecutorCache doesn't even return the aliased buffer of type ReuseBuffer. 
2. FusionExecutor, when `outputs` are specified, returns the same vector. So checking `data_ptr` of the return value or checking `tensor.data_ptr()==tensor.data_ptr()` isn't very interesting at this moment.

tests/cpp/test_alias.cpp
```diff
@@ -1049,7 +1049,43 @@ TEST_F(AliasTest, SegmentBoundary) {
...
 
-TEST_F(AliasTest, ReuseBufferAliasAcrossSegments) {
+TEST_F(AliasTest, ReuseBuffer) {
+  auto fusion = std::make_unique<Fusion>();
+  FusionGuard fg(fusion.get());
+  TensorView* in = makeContigTensor(1);
+  TensorView* out = add(in, IrBuilder::create<Val>(1.0));
+  fusion->addInput(in);
+  fusion->addOutput(out);
+  fusion->aliasOutputToInput(out, in, AllocationType::ReuseBuffer);
```
naoyam, line 1059: Naive question: when an output is an alias of an input, doesn't that mean the output always reuse the same buffer as the input?

wujingyue: It used to be -- we only used that to mark in-place updates. I've overloaded the API to mark aliases introduced by `slice`s (which doesn't strictly output the same buffer) and to mark for ATen evaluation. Therefore, a separate `AlloationType::Evaluate`. 

I'm open to ideas to clean this up. In the meantime, I believe we'll be able to explicitly represent buffer allocation and ATen evaluation in host IR, instead of relying on attributes on `Fusion`.

csarofeen:
Would it be too much to consider an approach like this one: https://github.com/NVIDIA/Fuser/blob/main/csrc/device_lower/pass/alias_memory.h when considering where to reuse buffers? Is this a similar problem?

wujingyue:
Yes, it's a similar problem and what you pointed out is a great reference point. 

Currently, when to reuse allocations of global memory is decided manually (e.g. when [building a batch norm](https://github.com/NVIDIA/Fuser/blob/65799ed56f66b671dfbd51336fa302400e2bdf45/csrc/ops/normalization.cpp#L568)) or decided under the hood by PyTorch's caching allocator. 

I haven't looked at how effective these decisions are, but I suspect nvFuser will have to gain more control down the road. One example is to support NCCL's user buffer registration for models of dynamic shapes.

PR #2442, opened 2024-06-23
Check that users don't call `FusionDefinition` methods from uninitialized definitions outside a Python Context Manager or `definition()` method

### LINKED ISSUE #2430: Segfault on FusionDefinition().define_scalar()
I encountered a segfault by chance while trying to query for function signature or docstrings. Here's a minimal repro:
```py
In [1]: from nvfuser import FusionDefinition

In [2]: FusionDefinition().define_scalar()
Segmentation fault (core dumped)
```

jacobhinkle:
Fixes #2430 since we _should_ error here but we _should not_ segfault ever.

### DIFF
--- a/csrc/python_frontend/fusion_definition.cpp
+++ b/csrc/python_frontend/fusion_definition.cpp
@@ -362,7 +362,8 @@ Scalar FusionDefinition::defineScalar() {
 Tensor FusionDefinition::addTensor(TensorView* tv) {
   FUSER_PERF_SCOPE("FusionDefinition::addTensor");
   NVF_CHECK(
-      trie_node_ != nullptr, "addTensor() must be called from an initialized definition via a python context manager or a child class' definition() method.");
+      trie_node_ != nullptr,
+      "addTensor() must be called from an initialized definition via a python context manager or a child class' definition() method.");
   Tensor output = defineTensor(tv->nDims());
   NVF_CHECK(
       output.index == numFusionStates(),
--- a/csrc/python_frontend/fusion_definition.cpp
+++ b/csrc/python_frontend/fusion_definition.cpp
@@ -353,7 +353,7 @@ Scalar FusionDefinition::defineScalar() {
   FUSER_PERF_SCOPE("FusionDefinition::defineScalar");
   NVF_CHECK(
       trie_node_ != nullptr,
-      "defineScalar must be called inside a 'with' block");
+      "define_scalar() must be called from an initialized definition via a python context manager or a child class' definition() method.");
   Scalar out(recording_state_.size(), this);
   recording_state_.emplace_back(out(), serde::StateType::Scalar);
   return out;
@@ -362,7 +362,7 @@ Scalar FusionDefinition::defineScalar() {
 Tensor FusionDefinition::addTensor(TensorView* tv) {
   FUSER_PERF_SCOPE("FusionDefinition::addTensor");
   NVF_CHECK(
-      trie_node_ != nullptr, "AddTensor must be called inside a 'with' block");
+      trie_node_ != nullptr, "addTensor() must be called from an initialized definition via a python context manager or a child class' definition() method.");
   Tensor output = defineTensor(tv->nDims());
   NVF_CHECK(
       output.index == numFusionStates(),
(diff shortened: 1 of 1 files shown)


jacobhinkle:
!build

csrc/python_frontend/fusion_definition.cpp
```diff
@@ -351,13 +351,18 @@ std::optional<size_t> FusionDefinition::id() const {
 
 Scalar FusionDefinition::defineScalar() {
   FUSER_PERF_SCOPE("FusionDefinition::defineScalar");
+  NVF_CHECK(
+      trie_node_ != nullptr,
+      "defineScalar must be called inside a 'with' block");
```
kevinstephano, line 356:
```suggestion
      "define_scalar() must be called from an initialized definition via a python context manager or a child class' definition() method.");
```

csrc/python_frontend/fusion_definition.cpp
```diff
@@ -351,13 +351,18 @@ std::optional<size_t> FusionDefinition::id() const {
...
   Scalar out(recording_state_.size(), this);
   recording_state_.emplace_back(out(), serde::StateType::Scalar);
   return out;
 }
 
 Tensor FusionDefinition::addTensor(TensorView* tv) {
   FUSER_PERF_SCOPE("FusionDefinition::addTensor");
+  NVF_CHECK(
+      trie_node_ != nullptr, "AddTensor must be called inside a 'with' block");
```
kevinstephano, line 365:
```suggestion
      trie_node_ != nullptr, "addTensor() must be called from an initialized definition via a python context manager or a child class' definition() method.");
```

csrc/python_frontend/fusion_definition.cpp
```diff
@@ -368,20 +373,29 @@ Tensor FusionDefinition::addTensor(TensorView* tv) {
 
 Tensor FusionDefinition::defineTensor(size_t dims) {
   FUSER_PERF_SCOPE("FusionDefinition::defineTensor");
+  NVF_CHECK(
+      trie_node_ != nullptr,
+      "defineTensor must be called inside a 'with' block");
```
kevinstephano, line 378:
```suggestion
      "define_tensor() must be called from an initialized definition via a python context manager or a child class' definition() method.");
```

csrc/python_frontend/fusion_definition.cpp
```diff
@@ -368,20 +373,29 @@ Tensor FusionDefinition::addTensor(TensorView* tv) {
...
   Tensor out(recording_state_.size(), dims, this);
   recording_state_.emplace_back(out(), serde::StateType::Tensor);
   return out;
 }
 
 Vector FusionDefinition::defineVector(size_t size) {
   FUSER_PERF_SCOPE("FusionDefinition::defineVector");
+  NVF_CHECK(
+      trie_node_ != nullptr,
+      "defineVector must be called inside a 'with' block");
```
kevinstephano, line 388:
```suggestion
      "define_vector() must be called from an initialized definition via a python context manager or a child class' definition() method.");
```

csrc/python_frontend/fusion_definition.cpp
```diff
@@ -368,20 +373,29 @@ Tensor FusionDefinition::addTensor(TensorView* tv) {
...
   Vector out(recording_state_.size(), size, this);
   recording_state_.emplace_back(out(), serde::StateType::Vector);
   return out;
 }
 
 void FusionDefinition::defineRecord(RecordFunctor* record) {
   FUSER_PERF_SCOPE("FusionDefinition::defineRecord");
+  NVF_CHECK(
+      trie_node_ != nullptr,
+      "defineRecord must be called inside a 'with' block");
```
kevinstephano, line 398:
```suggestion
      "defineRecord() must be called from an initialized definition via a python context manager or a child class' definition() method.");
```

csrc/python_frontend/fusion_definition.cpp
kevinstephano: Note, I am purposely suggesting using the python function name `define_scalar()` to give a better error message given this error is most likely to be seen from Python.

csrc/python_frontend/fusion_definition.cpp
kevinstephano: I don't think this method is actually exposed via python but it doesn't hurt to add the check.

csrc/python_frontend/fusion_definition.cpp
kevinstephano: I don't think this method is actually exposed via python but it doesn't hurt to add the check.

kevinstephano (changes requested):
I just wanted the error messages to be clearer for users.

kevinstephano (approved):
LGTM!

Issue #557, opened 2023-07-04
nvFuser's RNG agrees with PyTorch only on `(numel - 1) // 4 + 1` elements


IvanYashchuk:
Am I correct that nvFuser and PyTorch use the same variant of Philox RNG? 
nvFuser also shares the RNG state with PyTorch so some users might expect the same values to be generated. Is this a reasonable expectation and something we can promise (as long as PyTorch keeps using the same generator as we do)?

I tried exploring whether we're generating the same sequences as PyTorch, and we are not. But there's a pattern that the results match on the `(total_number_of_elements - 1) // 4 + 1` number of elements. So for a size 1 tensor - 1 element is matching, size 5 -> 2 matches, size 9 -> 3 matches, size 13 -> 4 matches, and so on. What's going on here? Can we match PyTorch's behavior or can we make changes to PyTorch so that it matches nvFuser's behavior?

Here's a script that I used:
```py
from nvfuser import FusionDefinition, DataType
import torch
from itertools import product

shape = (13,)
device = "cuda"

# Now let's try what sequence nvfuser will generate
with FusionDefinition() as fd:
    min = fd.define_scalar(0.0)
    max = fd.define_scalar(1.0)
    u = fd.ops.uniform(min, max, [fd.define_scalar(s) for s in shape])
    fd.add_output(u)

# PyTorch Eager and nvFuser use different threading mapping so the position of
# the elements in the output tensor will be different. However, the sorted values
# can be the same.
torch.manual_seed(0)
pt_result = torch.rand(shape, device=device).ravel().sort().values

torch.manual_seed(0)
nvfuser_result = fd.execute([])[0].ravel().sort().values

# count how many nvfuser elements are the same as the pytorch elements
count = sum(x == y for x, y in product(pt_result, nvfuser_result)).item()
print(f"nvfuser and pytorch agree on {count} elements out of {len(pt_result)}")
```

zasdfgbnm:
Your understanding that nvFuser is using the same philox engine as PyTorch, because PyTorch uses curand, and nvFuser's philox engine is copy-pasted from curand. If we do want the nvFuser and PyTorch to generate identical random numbers, I think it makes more sense to modify PyTorch. nvFuser's random number only depend on the logical index of an item in the tensor, not on stride order and schedule. And PyTorch's random number depends on schedule and stride order. For example:
```python
>>> torch.manual_seed(0)
<torch._C.Generator object at 0x7f9da0af4a70>
[...]

PR #2449, opened 2024-06-24
SDPA Bwd node implementation

Priya2698:
1. Add a `SdpaBwdOp` with same functionality as `at::_scaled_dot_product_flash_attention_backward`.
2. Adds handling for undefined tensors: The intermediates `cum_seq_q/k` from the forward pass are undefined tensors except when the input tensors are nested.

This PR does not include scheduling and root-domain mapping.


Priya2698:
!build

csrc/ir/internal_nodes.h
```diff
@@ -2298,4 +2298,150 @@ class SdpaFwdOp : public Expr {
       const std::vector<PolymorphicValue>& inputs) const override;
 };
 
+/*
+SDPA bwd node with same functionality at::_scaled_dot_product_flash_attention
```
jacobhinkle, line 2302:
```suggestion
SDPA bwd node with same functionality at::_scaled_dot_product_flash_attention_backward
```

csrc/ir/internal_nodes.h
```diff
@@ -2298,4 +2298,150 @@ class SdpaFwdOp : public Expr {
...
+      TensorView* grad_key,
+      TensorView* grad_value,
+      TensorView* grad_output,
+      TensorView* query,
+      TensorView* key,
+      TensorView* value,
+      TensorView* output,
+      TensorView* log_sumexp,
+      TensorView* cum_seq_q,
+      TensorView* cum_seq_k,
```
jacobhinkle, line 2349: Will these be `nullptr`? If so then we might segfault since we often assume `Expr::inputs()` returns non-null `Val*` pointers. One such place is in `toString()` below.

tests/cpp/test_sdpa_node.cpp
```diff
@@ -232,4 +229,268 @@ TEST_F(SDPATest, PairwiseRootDomainMap) {
...
+  std::vector<int64_t> k_shape({n, h, s, e});
+  std::vector<int64_t> v_shape({n, h, s, e});
+  std::vector<int64_t> attn_shape({n, h, l, e});
+
+  auto options = at::TensorOptions().dtype(at::kHalf).device(at::kCUDA, 0);
+  at::Tensor q = at::randn(q_shape, options);
+  at::Tensor k = at::randn(k_shape, options);
+  at::Tensor v = at::randn(v_shape, options);
+
+  double dropout_p = 0.0;
```
jacobhinkle, line 379: Might want to set this to a non-zero value to test that it's passed properly.

csrc/ir/internal_nodes.h
jjsjann123: I think we are getting empty tensors in those case. So it still wouldn't be nullptr?

tests/cpp/test_sdpa_node.cpp
```diff
@@ -48,7 +48,7 @@ TEST_F(SDPATest, NonCausalAttnConcrete) {
 
   auto tvq = makeConcreteTensor(q_shape, DataType::Half);
   auto tvk = makeConcreteTensor(k_shape, DataType::Half);
-  auto tvv = makeConcreteTensor(k_shape, DataType::Half);
+  auto tvv = makeConcreteTensor(v_shape, DataType::Half);
```
jjsjann123, line 51: :clap:

tests/cpp/test_sdpa_node.cpp
jjsjann123: since both reference and ExprEval uses the same torch implementation, if we start with the same random seed  we should still be able to compare the two.

On a separate note, I'm now wondering how are we testing nvfuser codegen's random behavior. Since we are not converging on the random number pattern with torch?

jacobhinkle: We do compute the same random numbers as torch, with the exception that changing stride order in pytorch changes the generated values whereas our numbers always match the logical domain. But yes, in this case it is ATen vs ATen so it should be exactly the same.

csrc/ir/internal_nodes.h
jacobhinkle: Ah ok, so we actually are getting tensors they just have size [0] or something?

tests/cpp/test_sdpa_node.cpp
jjsjann123: > We do compute the same random numbers as torch

Thanks a lot! I didn't know that.

I thought they matches by accident with small buffer and when the size goes up, pytorch matches it differently and would give a different pattern. That's probably some out-dated implementation. :)

csrc/ir/internal_nodes.h
jjsjann123: It's more explicit than that. On the cpp side, you can query `tensor.defined()`.

My rusty understanding is that, when a python API gets a None for a tensor input, you'll just have an undefined tensor. Which distinguishes it from actually receiving an empty tensor as input to the argument.

Priya2698: Yes, it is an undefined tensor. It does not have a device, but it is not a `nullptr`.

tests/cpp/test_sdpa_node.cpp
jacobhinkle: For reference: https://github.com/NVIDIA/Fuser/issues/557#issuecomment-1620466166

csrc/ir/internal_nodes.h
jacobhinkle: Interesting! Thanks for teaching me about this. Should we be checking `tensor.defined()` for inputs to `FusionExecutor::runFusion`? I suppose since their device is missing we would get an error about that if we were passed an undefined tensor in a case where we're not just evaluating with ATen.

tests/cpp/test_sdpa_node.cpp
jacobhinkle: We don't test that the numbers match anymore though, so this might no longer be the case and we should not rely on nvfuser computing the same random numbers.

csrc/ir/internal_nodes.h
Priya2698: Should this be done at node level since codegen should not be dealing with undefined tensors anyway? That is, filtering out the undefined tensors or deciding if it needs to get replaced with a tensor of zeros (https://github.com/pytorch/pytorch/issues/41359) based on what is required for kernel generation. Right now, we don't have any concrete examples though for non-ATen-only nodes.

I am checking for `tensor.defined()` in `getCommonCUDADevice` and `validateValWithConcreteValue` which are called within `compileFusion` and `runFusion` (from within `kernelArgumentHolder`).

Priya2698:
!build

csrc/ir/internal_nodes.h
jacobhinkle: I think since we don't have a way to handle undefined tensors at codegen yet (and maybe ever?) that it would be nice if we could error out if one is passed to a non-ATen evaluated segment. I suspect we'd just hit a less-informative error anyway though in such a case, and I'm not sure where we'd need to put that check. Since the approach you have here works for the ATen evaluated case, I think we're good.

Priya2698:
!build

Priya2698:
!build

jacobhinkle (approved):
LGTM

Priya2698:
!build

PR #2453, opened 2024-06-25
Promote `ForLoop` and `Scope` to global namespace

samnordmann:
Since https://github.com/NVIDIA/Fuser/pull/2440 uses `ForLoops` as a host IR, we remove `ForLoop` and `Scope` for the `kir` namespace and place them in the global namespace. See [this thread](https://github.com/NVIDIA/Fuser/pull/2440#discussion_r1650379533).

Addresses by the way an old "TODO":
https://github.com/NVIDIA/Fuser/blob/fd2586911a9748bec05498357598553aa1648100/csrc/kernel_ir.h#L723

Note that there is still an ambiguity that `ForLoop` appears in `csrc/kernel_ir_dispatch.h`. The same problem will arise with `IfElse` when we will use it as a host IR. We can either live with this ambiguity, or we could rename `kernel_ir_dispatch.h/cpp` to simply `ir_visitor.h/cpp`. Lmk


samnordmann:
!build

csrc/ir/internal_nodes.h
```diff
@@ -2298,4 +2298,214 @@ class SdpaFwdOp : public Expr {
...
+//! placed in its body are considered inside the scope of the for loop. In the
+//! future the implementation should look quite different so that we can do
+//! proper dependency annalysis like in Fusion.
+//!
+//! TODO(kir): this is not a real expression
+//!
+//! ForLoop may represent a part of an iteration domain representend
+//! by iter_domain_. In that case, the loop extent field, extent_, may
+//! be smaller than the extent of iter_domain_.
+class NVF_API ForLoop final : public Expr {
```
wujingyue, line 2392:
```suggestion
class ForLoop final : public Expr {
```

csrc/ir/internal_nodes.h
wujingyue, line 1: @naoyam what do you think about putting ForLoop in this file? AFAICT, `csrc/ir/*nodes.h` currently contains SSA nodes used in fusion IR. However, I'm unsure about the value of keeping that convention.

wujingyue (commented):
LGTM. Defer approval to @naoyam

csrc/ir/internal_nodes.h
naoyam: Sounds good to me. I don't think we need to stick to the convention.

naoyam:
> Note that there is still an ambiguity that `ForLoop` appears in `csrc/kernel_ir_dispatch.h`. The same problem will arise with `IfElse` when we will use it as a host IR. We can either live with this ambiguity, or we could rename `kernel_ir_dispatch.h/cpp` to simply `ir_visitor.h/cpp`. Lmk

I'm not sure if that's an issue. `kir::IrVisitor` is meant to work with a KIR program, which may consist of for-loop and if-then-else constructs. It's just now the namespace of the for-loop construct is not in the `kir` namespace.

naoyam (approved):
LGTM too.

samnordmann:
>> Note that there is still an ambiguity that ForLoop appears in csrc/kernel_ir_dispatch.h. The same problem will arise with IfElse when we will use it as a host IR. We can either live with this ambiguity, or we could rename kernel_ir_dispatch.h/cpp to simply ir_visitor.h/cpp. Lmk

> I'm not sure if that's an issue. kir::IrVisitor is meant to work with a KIR program, which may consist of for-loop and if-then-else constructs. It's just now the namespace of the for-loop construct is not in the kir namespace.

Just to clarify I am just saying that the filename is misleading because it contains "kernel_ir", but I also think it is not a big deal, so I am leaving as is for now

PR #2456, opened 2024-06-26
Allow domains to be decoupled: `ir_utils::allIDsOf`

zasdfgbnm:


zasdfgbnm:
!build

csrc/ir/utils.cpp
```diff
@@ -735,14 +735,25 @@ bool isIndexedConsumerID(const TensorView* tv, const IterDomain* id) {
...
   const auto& root_domain = tv->getMaybeRootDomain();
-  const auto& domain = tv->getLoopDomain();
-  // Grab all values in the history of the tensor view's domain
-  auto all_vals = DependencyCheck::getAllValsBetween(
-      {root_domain.begin(), root_domain.end()}, {domain.begin(), domain.end()});
+  const auto& logical_domain = tv->getLogicalDomain();
+  const auto& loop_domain = tv->getLoopDomain();
+  const auto& alloc_domain = tv->getAllocationDomain();
+
+  std::array<const std::vector<IterDomain*>*, 4> domains{
```
naoyam, line 744: nit: since most of the times `alloc_domain` would be the same as `logical_domain`, we may want to do some quick optimization by not doing `getAllValsBetween` multiple times if they are the same domain.

naoyam (approved):
LGTM

zasdfgbnm:
!build

csrc/ir/utils.cpp
zasdfgbnm: added

PR #2457, opened 2024-06-26
Use `HostIrExecutor` in `MultiDeviceExecutor`

samnordmann:
# What
Relies on `HostIrExecutor` to execute the fusion and communications inside `MultiDeviceExecutor`

# Why
- It plugs Host IRs into the pre-existing stack. It gives an example of automatically created Host Ir program from a scheduled Fusion. It also exercises and tests Host IR executor more thouroghly thanks to all the multi device tests.
- It significantly refactors and simplify the code
- I would like to merge it before other infra patches, e.g., concretize Fusion before using FusionExecutor, improving allocation,etc.

# How

The main change occurs in `MultiDeviceExecutor`'s constructor. After segmenting the fusion into a segmented fusion having the compute/comms segment, we create a HostIrContainer that will represent the host program. Each segment of the segmented fusion will be translated into either 1) Communication/Wait IRs, or 2) PostOnStream/HostUnit IRs, depending whether the segment is resharding.

# TODO

The way allocation is handled remains unchanged and is still a bit ugly. Ideally, we need to handle allocation through Host IRs, which will be done in future PRs.


samnordmann:
!build

csrc/host_ir/executor.h
```diff
@@ -53,7 +53,20 @@ class HostIrExecutor final : public OptInDispatch {
       Communicator* communicator = nullptr,
       HostIrExecutorParams = HostIrExecutorParams());
   std::vector<at::Tensor> runWithInput(
-      std::unordered_map<Val*, c10::IValue> val_to_IValue);
+      std::unordered_map<Val*, c10::IValue> val_to_IValue,
+      LaunchParams launch_params = LaunchParams());
```
wujingyue, line 57: What's this for? IIUC, LaunchParams is per-segment, and I'm yet to understand why we need to pass it in to HostIrExecutor::runWithInput, which runs the whole host program spanning through multiple segments.

csrc/multidevice/executor.h
```diff
@@ -69,18 +70,7 @@ namespace nvfuser {
...
-  bool use_fusion_executor_cache = false;
-  // Experimental: whether to apply auto-scheduling in FusionExecutorCache if
-  // use_fusion_executor_cache=true. WAR: temporary hack mainly use for
-  // development
-  bool skip_auto_scheduling = false;
-  // Experimental: whether to cache fusion executor. WAR: avoid recompilation
-  // but implicitely assumes that the input shape don't change over iterations
-  bool cache_fusion_executor = false;
-};
+using MultiDeviceExecutorParams = hir::HostIrExecutorParams;
```
wujingyue, line 73: I'd just replace MultiDeviceExecutorParams. I don't see the value of creating an alias.

samnordmann: agreed

csrc/host_ir/executor.h
wujingyue: > It is copied from what is currently implemented in MultiDeviceExecutor.

Good to know it's an old behavior copied over! 

Can you try your best to put some code comments? I bet LaunchParams being a global instead of a per-segment config will confuse many readers. It's good to understand how it's used today and the limitations.

csrc/multidevice/executor.cpp
```diff
@@ -49,46 +51,95 @@ std::unique_ptr<Fusion> copyFusionAndChangeOutputs(
...
 
-  staged_fusion_ =
-      SegmentCandidateFinder::segment(std::move(fusion), nullptr, options);
-
-  for (auto group : staged_fusion_->groups()) {
+  for (auto group : workspace.group_run_order) {
+    std::vector<Expr*> host_exprs;
     NVF_ERROR(!group->exprs().empty(), "invalid segmentation");
-    is_resharding_[group] = std::any_of(
+    bool is_resharding = std::any_of(
```
wujingyue, line 99:
```suggestion
    const bool is_resharding = std::any_of(
```

wujingyue: Apparently, this can be moved after the involvedDevices check so `is_resharding` gets closer to where it's used for readability.

tests/cpp/multidevice.h
```diff
@@ -60,7 +60,7 @@ class PipelineTest : public MultiDeviceTest {
   std::vector<c10::IValue> unsharded_inputs;
   std::vector<at::Tensor> outputs;
   std::vector<at::Tensor> ref_unsharded_outputs;
-  MultiDeviceExecutorParams multi_device_executor_params;
+  hir::HostIrExecutorParams multi_device_executor_params;
```
wujingyue, line 63:
```suggestion
  hir::HostIrExecutorParams host_ir_executor_params;
```

csrc/multidevice/executor.cpp
```diff
@@ -49,46 +51,95 @@ std::unique_ptr<Fusion> copyFusionAndChangeOutputs(
...
+  }
+  for (auto output : staged_fusion->outputs()) {
+    hic->addOutput(ir_cloner.clone(output));
   }
-  // prepare the order in which to launch the kernels/comms
-  prepareRuntimeOrder(staged_fusion_.get(), workspace_);
+
+  // Create the HostIrExecutor representing the host program
+  host_ir_executor_ =
+      std::make_unique<hir::HostIrExecutor>(std::move(hic), &comm, params);
```
wujingyue, line 134:
```suggestion
      std::make_unique<hir::HostIrExecutor>(std::move(hic), &comm, std::move(params));
```
to avoid an unnecessary copy.

wujingyue (approved):

csrc/multidevice/executor.cpp
samnordmann: That's what I did originally but the linter CI complained
https://github.com/NVIDIA/Fuser/actions/runs/9683775298/job/26720018890

csrc/host_ir/executor.h
samnordmann: > Can you try your best to put some code comments? I bet LaunchParams being a global instead of a per-segment config will confuse many readers. It's good to understand how it's used today and the limitations.

Sure. I'm gonna add comments.
I agree that the current solution is not ideal. We should consider fixing it in a near future. I will keep this in mind.

csrc/multidevice/executor.cpp
samnordmann: done!

samnordmann:
!build

csrc/multidevice/executor.cpp
wujingyue: That complain is actually correct. There has been several places where `const T&` could have been used to save copies, which I'd probably do. I thought `T` plus `std::move` can save copies as well but the clangtidy reminded me that `HostIrExecutorParams` is a trivially copyable type that does not have a move constructor. So `std::move` is as expensive as a copy.

samnordmann:
Recent PR #2476 introduced a bug for the test case `PipelineTestStagedReduction.StagedReduction/ReductionOnly`. This test case is also the only reason why we are passing `LaunchParams` around in the executor, which [legitimately raised confusion](https://github.com/NVIDIA/Fuser/pull/2457#discussion_r1656534279).

Since this test case doesn't add much value and is responsible for a lot of complexity, I decided to remove it, which allows for some simplification. I think it is the best decision for now.
@cowanmeg @wujingyue does it sound good to you?

samnordmann:
!build

wujingyue:
> @cowanmeg @wujingyue does it sound good to you?

SGTM!

PR #2459, opened 2024-06-26
Allow creating broadcast from nowhere

zasdfgbnm:
Make it possible to create a new broadcast during scheduling. This new broadcast neither comes from a `BroadcastOp` nor from a fusion input. It just appears from no where. This PR only adds the minimal support to our system to make the simple pointwise fusion at test `BroadcastFromNowhere` work. So far, we can not inline such fusions. Our sync analysis could not handle this either. I am not working on these two pieces so far, because it is not clear to me whether this will be needed for scheduling hopper MMA.


zasdfgbnm:
!build

csrc/ir/utils.cpp
```diff
@@ -755,6 +755,14 @@ std::vector<IterDomain*> allIDsOf(const TensorView* tv) {
       auto all_vals_01 = DependencyCheck::getAllValsBetween(
           {dom0->begin(), dom0->end()}, {dom1->begin(), dom1->end()});
       all_vals.pushBack(all_vals_01);
+      // Getting all vals is not sufficient, because there might be broadcasting
```
naoyam, line 758: Does this capture new broadcast IDs if they are just inserted but not used?

zasdfgbnm: Good question, let me more thoroughly test it.

csrc/ir/internal_base_nodes.h
```diff
@@ -683,6 +693,7 @@ class TensorDomain : public Val {
   const std::vector<IterDomain*> logical_domain_;
   std::vector<IterDomain*> allocation_domain_;
   std::vector<IterDomain*> loop_domain_;
+  std::vector<IterDomain*> additional_ids_;
```
zasdfgbnm, line 696: I believe this mechanism can be used for fixing:
https://github.com/NVIDIA/Fuser/pull/2379#discussion_r1653381767
But I will leave it as a future work.

csrc/ir/utils.cpp
zasdfgbnm: Added a new test. I do find some issues, so I added a new field `TensorDomain::additional_ids_` to help keep track of all IDs.

csrc/bfs.h
```diff
@@ -145,7 +145,7 @@ class BFS {
 
     while (!allToNodesVisited()) {
       bool something_was_processed = false;
-      std::deque<NodeType> not_ready_;
+      std::deque<NodeType> not_ready;
```
zasdfgbnm, line 148: Just a renaming, no logical change

zasdfgbnm:
@naoyam This is ready for review again

zasdfgbnm:
!build

csrc/ir/interface_nodes.h
```diff
@@ -282,6 +282,9 @@ class NVF_API TensorView : public Val {
       int64_t position,
       ComputeAtMode mode = ComputeAtMode::Standard);
 
+  //! Create a new broadcast IterDoimain with extent one in the loop domain
```
naoyam, line 285: Let's say this is a 0-dim tensor. Would inserting a broadcast result in any semantic change? 

Asserting non 0-dim would be also reasonable.

zasdfgbnm: I think the "semantic change" is "we want to schedule it like a 1D broadcasting tensor"? Allowing 0-dim is aligned with PyTorch's `unsqueeze`:

```python
>>> torch.zeros(()).unsqueeze(0).shape
torch.Size([1])
>>> torch.zeros(()).unsqueeze(-1).shape
torch.Size([1])
```

csrc/ir/utils.h
```diff
@@ -499,6 +499,10 @@ std::vector<TensorView*> getTVsWithDynamicTransform(Fusion* fusion);
 //! Then [I0, I1, I8, I9] is equivalent to [I6, I7, I2, I3]. [I0, I1, I2, I3] is
 //! equivalent to [I6, I7, I8, I9]. But [I0, I1, I8, I3] is NOT equivalent to
 //! [I6, I7, I2, I9]
+//!
+//! Broadcast IterDomains are ignored in this check, because we consider them as
+//! placeholders and allow them to be created (and annihilated?) arbitrarily as
```
zasdfgbnm, line 504: > and annihilated

Looks like the capability of creation and annihilation comes together, especially in the case today that we no longer assume the direction of transformations.

naoyam (approved):
LGTM

zasdfgbnm:
!build

csrc/ir/internal_base_nodes.h
```diff
@@ -619,6 +626,9 @@ class TensorDomain : public Val {
   //! Returns a position of a root domain
   int64_t rootPosOf(IterDomain* id) const;
 
+  //! Create a new broadcast IterDoimain with extent one in the loop domain
```
jacobhinkle, line 629:
```suggestion
  //! Create a new broadcast IterDomain with extent one in the loop domain
```

csrc/ir/interface_nodes.h
jacobhinkle, line 285:
```suggestion
  //! Create a new broadcast IterDomain with extent one in the loop domain
```

jacobhinkle (approved):

PR #2466, opened 2024-06-26
Remove unused method DynamicTransformInitialInfo::operator==

jacobhinkle:
I just noticed a typo in this signature and decided to fix it, then realized it's probably unused so I figured I'd just remove it.

### DIFF
--- a/csrc/dynamic_transform.h
+++ b/csrc/dynamic_transform.h
@@ -31,12 +31,6 @@ class DynamicTransformInitialInfoBuilder;
 //! sizes
 class DynamicTransformInitialInfo {
  public:
-  bool operator==(const DynamicTransformConcretizationInfo& other) const;
-
-  bool operator!=(const DynamicTransformConcretizationInfo& other) const {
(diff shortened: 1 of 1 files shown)


jacobhinkle:
!build

csrc/dynamic_transform.h
```diff
@@ -31,12 +31,6 @@ class DynamicTransformInitialInfoBuilder;
 //! sizes
 class DynamicTransformInitialInfo {
  public:
-  bool operator==(const DynamicTransformConcretizationInfo& other) const;
```
naoyam, line 34: Shouldn't the corresponding implementation be removed as well?

naoyam: Looks like it doesn't exist. 😃

naoyam (approved):

csrc/dynamic_transform.h
jacobhinkle: :-D Oops!

PR #2468, opened 2024-06-26
Adding unit tests to check TMA loads for input B

protonu:
In the new unit test added we load the input B tensor to shared memory using TMA. These tests are for a single tile, where M=64, N=8 ... 256 and K=16. So for the TN layout B: {N, K} TT: {K, N}.  In the following figure we show how we tile the tensor to 4 loads. 

image

The figure below shows the scheduling for the above tiling. This schedule/tiling works for all inputs sizes (and TT and TN)

image

In addition to the above tiling/shedules, we could have scheduled the same shape above to reduce the number of TMA loads to 2 from 4. It's illustrated in the figure below. Please note that it correctly works for cases where we don't need to split the inner dimension. As future work we can try and make it work for all cases.

image

The schedule for the above tiling scheme is show below. This is illustrated in the test `SingleTileWithTMALoadOuterDimNotSplit`
image

### DIFF
--- a/tests/cpp/test_mma.cpp
+++ b/tests/cpp/test_mma.cpp
@@ -441,6 +441,254 @@ TEST_P(HopperRS, SingleTile) {
   EXPECT_TRUE(at::allclose(cg_outputs[0], tref, 1e-5, 1e-5));
 }
 
+void scheduleTMALoadWhereOuterDimIsSplit(
+    TensorView* tv,
+    MmaInputSmemSwizzle swizzle,
+    DataType dtype) {
+  // We move the broadcast dim to be the left most.
+  moveInnerBroadcastLeft(tv);
+
+  if (swizzle == MmaInputSmemSwizzle::None) {
+    // For no-swizzle case, the entire tile are divided into 8x8 core matrices,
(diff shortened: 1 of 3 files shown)


protonu:
!build

tests/cpp/test_mma.cpp
```diff
@@ -441,6 +441,254 @@ TEST_P(HopperRS, SingleTile) {
   EXPECT_TRUE(at::allclose(cg_outputs[0], tref, 1e-5, 1e-5));
 }
 
+void scheduleTMALoadWhereOuterDimIsSplit(
```
zasdfgbnm, line 444: Should we create a member function of TensorView that contains common part of this and `scheduleTMALoadOuterDimNotSplit`? Like `TensorView::applySwizzle(MmaInputSmemSwizzle)`?

zasdfgbnm (approved):

tests/cpp/test_mma.cpp
protonu: That's a good ideas - I'll send a new PR for that.

PR #2474, opened 2024-06-27
merge all params into ReshapeReduction suite

liqiangxl:
**Issue:** Test function `reductionViewAddFusion` still loops over `has_implicit_broadcast = {true, false}`
**Fix:** move `has_implicit_broadcast` from test function to paras
**Further clean up:**
merged `ReshapeBeforeReduction` and `ReshapeAfterReduction` into `ReshapeReduction` by changing the para to:
```
struct ReshapeReductionParam {
  ReshapeExample reshape_example;
  bool has_implicit_broadcast;
  bool reshape_before_reduction;
};
```

### DIFF
--- a/tests/cpp/test_gpu_view.cpp
+++ b/tests/cpp/test_gpu_view.cpp
@@ -403,8 +403,8 @@ TEST_P(ReshapeReduction, FusionReshapeReduction) {
   reductionViewAddFusion(
       input_shape,
       output_shape,
-      has_implicit_broadcast,
-      reshape_before_reduction);
+      param.has_implicit_broadcast,
+      param.reshape_before_reduction);
 }
 
 INSTANTIATE_TEST_SUITE_P(
--- a/tests/cpp/test_gpu_view.cpp
+++ b/tests/cpp/test_gpu_view.cpp
@@ -399,8 +399,6 @@ using ReshapeReduction = NVFuserFixtureParamTest<ReshapeReductionParam>;
 TEST_P(ReshapeReduction, FusionReshapeReduction) {
   const auto& param = GetParam();
   const auto& [input_shape, output_shape] = param.reshape_example;
-  const auto& has_implicit_broadcast = param.has_implicit_broadcast;
-  const auto& reshape_before_reduction = param.reshape_before_reduction;
   maybeClearAllocator(); // Shmoo tests can occupy a lot of memory
   reductionViewAddFusion(
       input_shape,
--- a/tests/cpp/test_gpu_view.cpp
+++ b/tests/cpp/test_gpu_view.cpp
@@ -243,6 +243,7 @@ TEST_F(GpuViewTest, FusionReshapeFailMulitDimInference) {
 void reductionViewAddFusion(
     const std::vector<int64_t>& input_shape,
     const std::vector<int64_t>& output_shape,
+    const bool has_implicit_broadcast,
     bool reshape_before_reduction) {
   constexpr int kReductionAxis = -1;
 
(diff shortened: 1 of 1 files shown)


tests/cpp/test_gpu_view.cpp
```diff
@@ -368,33 +367,54 @@ std::vector<ReshapeExample> reshape_after_reduce_examples = {
...
-using ReshapeAfterReduction = NVFuserFixtureParamTest<ReshapeExample>;
-TEST_P(ReshapeAfterReduction, FusionReshapeAfterReduction) {
-  const auto& [input_shape, output_shape] = GetParam();
+
+using ReshapeReduction = NVFuserFixtureParamTest<ReshapeReductionParam>;
+TEST_P(ReshapeReduction, FusionReshapeReduction) {
+  const auto& param = GetParam();
+  const auto& [input_shape, output_shape] = param.reshape_example;
+  const auto& has_implicit_broadcast = param.has_implicit_broadcast;
+  const auto& reshape_before_reduction = param.reshape_before_reduction;
```
wujingyue, line 403:
```suggestion
```

tests/cpp/test_gpu_view.cpp
```diff
@@ -368,33 +367,54 @@ std::vector<ReshapeExample> reshape_after_reduce_examples = {
...
   maybeClearAllocator(); // Shmoo tests can occupy a lot of memory
   reductionViewAddFusion(
-      input_shape, output_shape, /*reshape_before_reduction=*/false);
+      input_shape,
+      output_shape,
+      has_implicit_broadcast,
+      reshape_before_reduction);
```
wujingyue, line 409:
```suggestion
      param.has_implicit_broadcast,
      param.reshape_before_reduction);
```

tests/cpp/test_gpu_view.cpp
```diff
@@ -368,33 +367,54 @@ std::vector<ReshapeExample> reshape_after_reduce_examples = {
...
     {{1, 7844, 1, 7}, {1, 1961, 4}}};
 
-namespace {
-using ReshapeBeforeReduction = NVFuserFixtureParamTest<ReshapeExample>;
-TEST_P(ReshapeBeforeReduction, FusionReshapeBeforeReduction) {
-  const auto& [input_shape, output_shape] = GetParam();
-  maybeClearAllocator(); // Shmoo tests can occupy a lot of memory
-  reductionViewAddFusion(
-      input_shape, output_shape, /*reshape_before_reduction=*/true);
+namespace ReshapeReduction {
```
wujingyue, line 370:
```suggestion
namespace {
```

Is there a specific reason you gave it a name? An anonymous namespace looks just fine in a cpp file.

tests/cpp/test_gpu_view.cpp
```diff
@@ -368,33 +367,54 @@ std::vector<ReshapeExample> reshape_after_reduce_examples = {
...
+  // For each reshape, test with and without implicit broadcast
+  int total_tests =
+      2 * (all_reshape_examples.size() + reshape_after_reduce_examples.size());
+  std::vector<ReshapeReductionParam> params;
+  params.reserve(total_tests);
+  for (auto reshape_before_reduction : {true, false}) {
+    const auto& examples = reshape_before_reduction
+        ? all_reshape_examples
+        : reshape_after_reduce_examples;
+    for (auto has_implicit_broadcast : {false, true}) {
```
wujingyue, line 388: There's a thing called testing::Combine, which I'd probably use and simplify: https://google.github.io/googletest/reference/testing.html#INSTANTIATE_TEST_SUITE_P

wujingyue (approved):

tests/cpp/test_gpu_view.cpp
liqiangxl: just in case there are multiple anonymous namespaces with same named vars, we have test files close to 10K lines of code.

tests/cpp/test_gpu_view.cpp
liqiangxl: Seems a way to reduce lines of code. This test is not using `all combinations`, e.g. shape of input & output depend on whether reshape is before or after reduction. Will prioritize using this func when the test is a Cartesian product of all paras.

wujingyue: Got it! 

Nit: `has_implicit_broadcast` seems to be orthogonal, so could still benefit. 

Another nit: Instead of 
```
for (cond : {false, true}) {
  examples = cond ? a_examples : b_examples;
  for (example : examples) {
    params.push_back(...);
  }
}
```
I'd flatten it slightly so the code is more direct. 
```
for (example : a_examples) {
  params.push_back(..., /*cond=*/true);
}
for (example : b_examples) {
  params.push_back(..., /*cond=*/false);
}
```

(I'm not requesting changes. Just some nits to keep in mind for future PRs.)

PR #2477, opened 2024-06-27
[WIP] Preseg propagate pad binary op

jjsjann123:


jjsjann123:
This PR doesn't make sense, since with only unary op in the original PR #2373 makes it hard to justify the logic in the optimization pass.

PR #2480, opened 2024-06-27
Add a test for a bank-conflict-free transpose kernel with new swizzle

zasdfgbnm:
And make that test work


zasdfgbnm:
!build

csrc/id_model/indexing.cpp
```diff
@@ -43,30 +43,10 @@ IterDomain* getLoopPromotion(IterDomain* loop_id, const IdModel& id_model) {
   return loop_promotion_map_it->second;
 }
 
-// True if a given domain is a loop doamin of a given tensor and its
-// loop is partitioned with respect to the memory type of the tensor
-bool isPartitionedLoop(TensorView* tv, IterDomain* id) {
```
zasdfgbnm, line 48: Inlined into `getAllocationDomains`

csrc/id_model/indexing.cpp
```diff
@@ -43,30 +43,10 @@ IterDomain* getLoopPromotion(IterDomain* loop_id, const IdModel& id_model) {
...
-  return ir_utils::isMemoryPartitionedAcross(
-      tv->getMemoryType(), id->getParallelType());
-}
-
 bool isSizeOneDomain(IterDomain* id) {
   return id->isBroadcast() || id->isReduction() || id->extent()->isOneInt();
 }
 
-// True if a given domain of a tensor *may* require allocation
-bool mayRequireAllocation(TensorView* tv, IterDomain* id) {
```
zasdfgbnm, line 66: Inlined into `getAllocationDomains`

csrc/id_model/indexing.cpp
```diff
@@ -113,32 +93,33 @@ std::tuple<std::vector<IterDomain*>, std::vector<Val*>> getAllocationDomains(
       allocation_domains = tv->getLogicalDomain();
       contiguity = tv->domain()->contiguity();
     } else {
-      const auto inlining_pos = tv->getComputeAtPosition();
```
zasdfgbnm, line 116: In this function, currently there are multiple check points for which ID to skip. I feel that this is confusing, and indeed, it could miss IDs for some case. For example, if `tv->hasAllocation()` is true, then we will not remove grid parallelized IDs in the allocation domain if `tv` is smem tensor. I refactored this function to have a single check point regarding which ID to skip.

zasdfgbnm:
Note to myself: the following test does not pass for now. Work on it in my next PR.

```C++
TEST_F(SwizzleTest, Transpose2) {
  Fusion fusion;
  FusionGuard fg(&fusion);

  fusion.addInput(tv0);
  auto tv1 = set(tv0);
  auto tv2 = transpose(tv1, 0, 1);
  fusion.addOutput(tv2);
  tv1->setMemoryType(MemoryType::Shared);

  AbstractTensor smem_alloc{tv1->axis(0), tv1->axis(1)};

  std::vector<IterDomain*> dim0{tv1->axis(0), tv2->axis(1)};
  std::vector<IterDomain*> dim1{tv1->axis(1), tv2->axis(0)};

  AbstractTensor loop{dim0, dim1};
  loop.split(1, 32);
  loop.split(0, 32);
  loop.reorder({{1, 2}});
  loop.merge(0);
  // BIDx, 32, 32
  std::swap(loop[1][1], loop[2][1]);
  loop.merge(1);
  loop.split(1, 256);
  // BIDx, 4, TIDx

  for (auto id : loop[0].as<std::vector>()) {
    id->parallelize(ParallelType::BIDx);
  }
  for (auto id : loop[2].as<std::vector>()) {
    id->parallelize(ParallelType::TIDx);
  }

  auto uz = loop.unzip();
  tv1->setLoopDomain(uz[0].as<IterDomain*>());
  tv2->setLoopDomain(uz[1].as<IterDomain*>());

  smem_alloc.split(1, 32);
  smem_alloc.split(0, 32);
  smem_alloc.reorder({{1, 2}});
  smem_alloc.merge(0);
  smem_alloc.swizzle(SwizzleType::XOR, 1, 2);
  tv1->setAllocationDomain(smem_alloc.as<IterDomain*>(), true);

  EXPECT_NE(tv1->getLoopDomain()[0], tv1->getAllocationDomain()[0]);

  inlineMost();

  auto t = at::randn({10240, 10240}, options);
  FusionExecutor fe;
  fe.compileFusion(&fusion, {t});
  EXPECT_TRUE(getBankConflictInfo(fe.kernel()).empty());
  std::vector<at::Tensor> outputs = fe.runFusion({t});
  EXPECT_TRUE(at::equal(t.t(), outputs[0]));
}
```

zasdfgbnm:
!build

tests/cpp/test_swizzle.cpp
```diff
@@ -690,4 +695,56 @@ TEST_F(SwizzleTest, SwizzleInProducerProjection) {
...
+    id->parallelize(ParallelType::BIDx);
+  }
+  for (auto id : loop[2].as<std::vector>()) {
+    id->parallelize(ParallelType::TIDx);
+  }
+
+  auto ub = loop.unzip();
+  tv1->setLoopDomain(ub[0].as<IterDomain*>());
+  tv2->setLoopDomain(ub[1].as<IterDomain*>());
+  tv1->setAllocationDomain(smem_alloc.as<IterDomain*>(), true);
```
naoyam, line 737: Is this the reason of removing the check in `IterDomain::parallelize`? If `setLoopDomain` is done before `id->parallelize`, can we keep the check there?

zasdfgbnm: For this specific PR, yes, we can keep. But I don't think this check makes sense any more. Loop domain are no longer required to be leaf, so it makes no sense to assert that it has no use.

zasdfgbnm: Specifically, if don't parallelize the loop domain as `[4, TIDx{32*32/4}]`, but instead, as `[TIDy{32}, TIDx{32}]`, then this check will fail, even if we set loop domain first.

naoyam: I'd prefer keeping the check as long as we can. When the check was added, we did find a couple of inconsistent scheduling. Am I understanding correctly that we can still keep it for now?

naoyam: I'd be happy to change my mind then!

csrc/id_model/indexing.cpp
```diff
@@ -113,32 +93,33 @@ std::tuple<std::vector<IterDomain*>, std::vector<Val*>> getAllocationDomains(
...
+        ir_utils::isMemoryPartitionedAcross(tv->getMemoryType(), pt)) {
+      non_allocating_ids.insert(id);
+    }
+  }
+  const auto inlining_pos = tv->getComputeAtPosition();
+  if (!tv->isFusionInput() && !tv->isFusionOutput()) {
+    for (const auto i : c10::irange(inlining_pos)) {
+      auto loop_id = tv->getLoopDomain().at(i);
+      auto pt = loop_id->getParallelType();
+      if (!ir_utils::isMemorySharedAcross(tv->getMemoryType(), pt)) {
```
jacobhinkle, line 117: I suggest preserving the comment here.
```suggestion
      // If the position is left of the inlining position, no need to
      // allocate the domain unless it's shared. For example, if this
      // is a Shared tensor and the domain is parallelized with TID,
      // even if it's outside of the CA position, since the domain
      // is shared, it must be allocated.
      if (!ir_utils::isMemorySharedAcross(tv->getMemoryType(), pt)) {
```

jacobhinkle (approved):

zasdfgbnm:
@jacobhinkle @naoyam I rebased this PR, looks like no changes in `TensorIndexer` is required any more.

zasdfgbnm:
!build

naoyam (approved):
LGTM

jacobhinkle (approved):

zasdfgbnm:
!build

PR #2482, opened 2024-06-27
Allow multiple outputs with different root domains for SDPAOp in IDModel

Priya2698:
CC: @naoyam 

This PR allows outputs of the SDPA ops to have different root domains.

### DIFF
--- a/csrc/id_model/id_model.cpp
+++ b/csrc/id_model/id_model.cpp
@@ -319,7 +319,7 @@ void IdModel::buildExactGraph() {
     } else {
       for (auto p_tv : tv_inputs) {
         for (auto c_tv : other_tv_outputs) {
-          auto exact_c2p_root_map = PairwiseRootDomainMap(p_tv, c_tv)
+          auto exact_c2p_root_map = PairwiseLogicalDomainMap(p_tv, c_tv)
                                         .mapBroadcast(false)
                                         .mapConsumerToProducer();
 
@@ -940,7 +940,7 @@ void IdModel::validateAndPropagatePType() {
 class SdpaFwdOp;
 
 bool hasUniformSiblings(Expr* expr) {
-  return !expr->isOneOf<SdpaFwdOp>();
+  return !expr->isOneOf<SdpaFwdOp, SdpaBwdOp>();
 }
 
 } // namespace nvfuser
--- a/csrc/id_model/id_model.cpp
+++ b/csrc/id_model/id_model.cpp
@@ -937,8 +937,6 @@ void IdModel::validateAndPropagatePType() {
   }
 }
 
-class SdpaFwdOp;
-
 bool hasUniformSiblings(Expr* expr) {
   return !expr->isOneOf<SdpaFwdOp, SdpaBwdOp>();
 }
--- a/csrc/id_model/id_model.cpp
+++ b/csrc/id_model/id_model.cpp
@@ -286,22 +286,6 @@ void IdModel::buildExactGraph() {
         all_tv_outputs.begin(), all_tv_outputs.end());
     other_tv_outputs.pop_front();
 
-    for (auto other_tv_output : other_tv_outputs) {
-      // Sibling tv's must be exactly mapped with eachother so simply zip
-      // their loop iter domains.
-
-      NVF_ERROR(
-          other_tv_output->getMaybeRootDomain().size() ==
-              c_tv->getMaybeRootDomain().size(),
-          "Multiple outputs with mismatched TV domains is not supported.");
-
-      for (auto domain_i : c10::irange(c_tv->getMaybeRootDomain().size())) {
-        auto c_id = c_tv->getMaybeRootDomain()[domain_i];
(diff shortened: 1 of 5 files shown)


Priya2698:
!build

csrc/id_model/id_model.cpp
```diff
@@ -922,4 +936,10 @@ void IdModel::validateAndPropagatePType() {
   }
 }
 
+class SdpaFwdOp;
```
Priya2698, line 939: @naoyam Is there any use for this?

Priya2698:
!build

csrc/id_model/id_model.cpp
naoyam: No, I don't think that's necessary.

naoyam (commented):
Can you please add a test to verify the mappings of `SdpaFwdOp` and `SdpaBwdOp` in IdModel? Something like:

```
IdModel id_model(&fusion);
const auto& exact_graph = id_model.idGraph(IdMappingMode::EXACT);
for (auto producer: sdpa_op->inputs()) {
  for (auto consumer: sdpa_op->outputs()) {
        EXPECT_TRUE(exact_graph.distjointValSets().strictAreMapped(producer->getLogialDomain().at(...), consumer->getMaybeRootDomain().at(...));
        ...
   }
}
```

tests/cpp/test_sdpa_node.cpp
```diff
@@ -216,73 +319,6 @@ TEST_F(SDPATest, CausalAttn) {
   validateSdpaFwdOutputs(nvf_out, aten_out);
 }
 
-TEST_F(SDPATest, PairwiseLogicalDomainMap) {
```
Priya2698, line 219: Removing this test, since it is redundant with the IDModel mapping checks included with all unit tests.

Priya2698:
!build

tests/cpp/test_sdpa_node.cpp
```diff
@@ -82,6 +69,116 @@ auto validateSdpaFwdOutputs = [](std::vector<at::Tensor> nvf_out,
...
 
+void checkSdpaFwdMapping(Fusion* fusion, Expr* op) {
+  IdModel id_model(fusion);
+  const ValGraph& vg = id_model.idGraph(IdMappingMode::EXACT);
+  vg.validateConsistency();
+
+  auto sdpa_op = dynamic_cast<SdpaFwdOp*>(op);
+  ASSERT_TRUE(sdpa_op != nullptr);
+
+  /* SdpaFwdOp:
```
naoyam, line 80: Are we checking the S domains of key and value producers are mapped? Also, the E domains of query and key producers?

tests/cpp/test_sdpa_node.cpp
naoyam: I'd leave these since they are not exactly the same. There are some overlap, but seems like a negligible cost given that these properties are really important to make them right.

tests/cpp/test_sdpa_node.cpp
Priya2698: There is no consumer for those, and we don't map the producers in ID model for these two ops.

Priya2698: Should we map them in the ID model then?

naoyam: Yes, let's map them together.

Priya2698: There are two ways:
1. Adding reduction axis for these domains in the output (similar to having K as a reduction axis in matmul, linear, output)
2. Mapping the producers explicitly in `id_model.cpp::buildExactGraph`.

The former avoids special cases in ID model code but adds the complexity of 2 additional reduction axis.

naoyam: Are they reduction domains? If so, #1 seems to make most sense.

Priya2698: @naoyam: Are there any cons to leaving these two domains unmapped?
We ran into an issue when the reduction dimension is 1 in matmul (https://github.com/NVIDIA/Fuser/issues/2532). In this case, we are considering removing the reduction axis from the output.

A reduction axis in the consumer in matmul is required for translating to `MmaOp`. Since SDPA operators would not be nvFuser scheduled, these are not functionally required.

Wdyt?

naoyam: If it's not trivial, don't bother for now. I thought it'd be just adding some things and add additional mappings, which may be useful in some cases. Faithfully modeling domain relationships is always a good thing, but given that there's no known use case, it's fine to leave as a future to-do item. Please make sure to leave an appropriate comment.

Priya2698: I'll go ahead with leaving these unmapped to avoid issues similar to (#2532) with SDPA. Will revisit if any use case comes up after integrating with Thunder.

naoyam (approved):
LGTM

Priya2698:
!build

PR #2483, opened 2024-06-27
Python API for `SdpaFwdOp`

Priya2698:
Python API for `SdpaFwdOp`.

Currently, the tests are located in `test_python_frontend.py` only since`pytest_ops.py` expects same arguments and outputs between ATen reference and nvFuser. Will experiment adding that in a future PR.

Issue #2278


csrc/preseg_passes/pre_segmenter.cpp
```diff
@@ -35,10 +35,10 @@ namespace nvfuser::preseg_passes {
   // removes consecutive cast operations
   OptimizationPass<ConsecutiveCastPass>::runPass(fusion);
   OptimizationPass<AddAxiomsPass>::runPass(fusion);
-  OptimizationPass<MoveSplitCatPass>::runPass(fusion);
+  // OptimizationPass<MoveSplitCatPass>::runPass(fusion);
```
Priya2698, line 38: TODO: Restore before merge.

csrc/preseg_passes/pre_segmenter.cpp
```diff
@@ -35,10 +35,10 @@ namespace nvfuser::preseg_passes {
...
   OptimizationPass<MarkAliasesPreparePass>::runPass(fusion);
   OptimizationPass<ExactMappedExtentSubstitutionPass>::runPass(fusion);
-  OptimizationPass<AllocationDomainPass>::runPass(fusion);
+  // OptimizationPass<AllocationDomainPass>::runPass(fusion);
```
Priya2698, line 41: TODO: Restore before merge.

tests/python/test_python_frontend.py
```diff
@@ -4342,7 +4342,75 @@ def fusion_func(fd: FusionDefinition):
...
+            output = fd.from_pytorch(inputs[4])
+            log_sumexp = fd.from_pytorch(inputs[5])
+            query_seq_len = fd.from_pytorch(inputs[6])
+            key_seq_len = fd.from_pytorch(inputs[7])
+            philox_seed = fd.from_pytorch(inputs[8])
+            philox_offset = fd.from_pytorch(inputs[9])
+            dropout_p = fd.define_scalar(0.0)
+            is_causal = fd.define_scalar(False, dtype=DataType.Bool)
+
+            grad_query, grad_key, grad_value= fd.ops.sdpfa_bwd(grad_output, q, k, v, output, log_sumexp, query_seq_len, key_seq_len, dropout_p, is_causal, philox_seed, philox_offset)
```
Priya2698, line 4406: @rdspring1: Here, using `fd.ops.sdpfa_bwd` does not allow `dropout_p, is_causal` to be None.

Note, that in `SdpaBwdOpRecord`, I skip printing `None` arguments. If I allow these arguments to be printed I run into the following error:
```
FAILED tests/python/test_python_frontend.py::TestNvFuserFrontend::test_sdpa_bwd - NameError: name 'None0' is not defined
```

Priya2698:
!build

tests/python/test_python_frontend.py
```diff
@@ -4458,6 +4457,53 @@ def fusion_func(fd: FusionDefinition):
...
+                    torch.randn((N, H, L, E), dtype=torch.bfloat16, device="cuda:0"),
+                    torch.randn((N, H, S, E), dtype=torch.bfloat16, device="cuda:0"),
+                    torch.randn((N, H, S, E), dtype=torch.bfloat16, device="cuda:0"),
+                ]
+                # Reset the FusionCache or the fusion would not recompile for all subtests, failing checks in exec_nvfuser.
+                # TODO: Try to move this to pytest_ops.py. Currently, it does not work since the API between nvFuser and torch differs.
+                FusionCache.reset()
+                nvf_out, _ = self.exec_nvfuser(fusion_func, [*qkv, is_causal, scale])
+                ref_out = F.scaled_dot_product_attention(
+                    *qkv, is_causal=is_causal, scale=scale
```
jjsjann123, line 4503: nitpick here: dropout default to None here. Is it because we don't reset seed so we cannot test with a real dropout value?

Just curious, we don't have to fix it in this PR. But maybe a follow up issue would be nice.

tests/python/test_python_frontend.py
```diff
@@ -4458,6 +4457,53 @@ def fusion_func(fd: FusionDefinition):
...
+            is_causal = fd.define_scalar(None, dtype=DataType.Bool)
+            scale = fd.define_scalar(None, dtype=DataType.Double)
+            attn, *intermediate_results = fd.ops.sdpfa_fwd(
+                q, k, v, is_causal=is_causal, scale=scale
+            )
+            fd.add_output(attn)
+
+        N, H, L, S, E = 4, 8, 16, 16, 8
+        is_causal_vals = [False, True]
+        scale_vals = [1 / E**0.5, 1e-3]
```
jjsjann123, line 4489: nitpick, can we also test default value for scale/is_causal?

csrc/python_frontend/python_bindings.cpp
```diff
@@ -2911,8 +2911,71 @@ void initNvFuserPythonBindings(PyObject* module) {
...
+
+        auto dropout_p_state = dropout_p.has_value()
+            ? fd->recordingState(dropout_p.value()())
+            : State(0, serde::StateType::None);
+        auto is_causal_state = is_causal.has_value()
+            ? fd->recordingState(is_causal.value()())
+            : State(0, serde::StateType::None);
+        auto scale_state = scale.has_value()
+            ? fd->recordingState(scale.value()())
+            : State(0, serde::StateType::None);
```
jjsjann123, line 2944: Is default 0 the right behavior?

A quick dive into the sdpa_fwd, looks like a none scale triggers a different code path vs some concrete value.

jjsjann123 (commented):
LGTM in general, but I have some concern over the default value. Can we verify the behavior in tests?

tests/python/test_python_frontend.py
Priya2698: Yes, with non-zero dropout, we cannot validate against torch.
What do you mean by resetting the seed?

jjsjann123: I assume that we cannot validate against torch simply because the offset is moved?

i.e. Would this allow us to validate? vvv
```
torch.cuda.manual_seed(0)
ref_out = F.scaled_dot_product_attention(...)
```

csrc/python_frontend/python_bindings.cpp
Priya2698: That is the state index, and not the default value.
I call the C++ API using `nullptr`.

tests/python/test_python_frontend.py
Priya2698: Right, we are not validating the RNG state here so this should work.
Let me try.

csrc/python_frontend/python_bindings.cpp
jjsjann123: :face_with_head_bandage: 

Of course. Sorry about misreading this.....

csrc/python_frontend/python_bindings.cpp
```diff
@@ -2911,8 +2911,71 @@ void initNvFuserPythonBindings(PyObject* module) {
...
+         std::optional<Scalar> is_causal,
+         std::optional<Scalar> scale) -> decltype(auto) {
+        FUSER_PERF_SCOPE("Operators.sdpfa_fwd");
+        NVF_CHECK(
+            self.validUse(), "Attempting to add to a completed definition!");
+        FusionDefinition* fd = self.fusion_definition;
+        size_t ndims = query.dims;
+        Tensor output = fd->defineTensor(ndims);
+        Tensor log_sumexp = fd->defineTensor(ndims - 1);
+        Tensor query_seq_len = fd->defineTensor(0);
```
rdspring1, line 2930: For clarity:
```suggestion
        Tensor query_seq_len = fd->defineTensor(/*dims=*/0);
```

csrc/python_frontend/python_bindings.cpp
rdspring1, line 2944: For clarity:
```suggestion
            : State(/*_index=*/0, /*_stype=*/serde::StateType::None);
```

tests/python/test_python_frontend.py
```diff
@@ -4458,6 +4457,53 @@ def fusion_func(fd: FusionDefinition):
         for i in range(num_out):
             self.assertEqual(nvf_out[i].data_ptr(), inputs[0].data_ptr())
 
+    def test_sdpa_fwd(self):
```
rdspring1, line 4460: Do we need a device check for this function? I believe it is restricted to Ampere or newer.

rdspring1 (commented):
LGTM.

tests/python/test_python_frontend.py
Priya2698: Good catch, that is true. Will add it.

tests/python/test_python_frontend.py
```diff
@@ -4458,6 +4457,140 @@ def fusion_func(fd: FusionDefinition):
         for i in range(num_out):
             self.assertEqual(nvf_out[i].data_ptr(), inputs[0].data_ptr())
 
+    @unittest.skipIf(is_pre_ampere(), "Only supported on Ampere and newer devices.")
+    def test_sdpa_fwd(self):
```
Priya2698, line 4461: I realize this test is long. Once https://github.com/NVIDIA/Fuser/pull/2701 goes through, I intend to move sdpa tests to a separate file.

tests/python/test_python_frontend.py
```diff
@@ -4458,6 +4457,140 @@ def fusion_func(fd: FusionDefinition):
...
+            fd.add_output(attn)
+
+        N, H, L, S, E = 4, 8, 16, 16, 8
+        qkv = [
+            torch.randn((N, H, L, E), dtype=torch.bfloat16, device="cuda:0"),
+            torch.randn((N, H, S, E), dtype=torch.bfloat16, device="cuda:0"),
+            torch.randn((N, H, S, E), dtype=torch.bfloat16, device="cuda:0"),
+        ]
+
+        dropout_vals = [0.0, 0.2]
```
jjsjann123, line 4547: :clap:

jjsjann123: nitpick on tests here.

I think the your default for None is just None?! so we can probably have it here as:
`dropout_vals = [0.0, 0.2, None]`. Same thing for `is_causal_vals` & `scale_vals`.

So we can just have `itertools.product(is_causal_vals, scale_vals, dropout_vals)`.

Sorry if my previous comment is misleading. We don't have to test pybind API behavior, we just need to cover all the code path in our part.

jjsjann123 (approved):
LGTM. A little nitpick on testing scripts.

tests/python/test_python_frontend.py
Priya2698: Yes, I was attempting that, however scalars (defined through `fd.define_scalar`) does not accept `None` values. Is there any other way to represent `None` values  in fusions?

Priya2698: Another way would be to pass these values to fusion definitions, however that triggers re-compilation failing checks in `exec_nvfuser`.

jjsjann123: ah... I see where it's coming from.

Would a partial work there?

```
def fusion_default(fd, has_dropout: bool, ...):

  dropout_p = None
  if has_dropout:
    dropout_p = fd.define_scalar(None, dtype=DataType.Double)
```

so you can have
```
  nvf_out, _ = self.exec_nvfuser(fusion_default_no_dropout = partial(fusion_default, False, ...), ...)
```

This way you can also simplify the script above.

jjsjann123: BTW, I'm not blocking your PR for this.

You mentioned that this test will be moved in follow up PR, so we can have the clean up then.

rdspring1 (approved):
LGTM

tests/python/test_python_frontend.py
```diff
@@ -4458,6 +4457,140 @@ def fusion_func(fd: FusionDefinition):
...
+
+        # is_causal is not set to exercise default values
+        for dropout_p, scale in itertools.product(dropout_vals, scale_vals):
+            with self.subTest(dropout_p=dropout_p, scale=scale):
+                # Reset the FusionCache or the fusion would not recompile for all subtests, failing checks in exec_nvfuser.
+                FusionCache.reset()
+                nvf_out, _ = self.exec_nvfuser(
+                    fusion_default_causal, [*qkv, dropout_p, scale]
+                )
+                torch.manual_seed(0)
```
wujingyue, line 4574: I understand you're trying to make the dropout in ref and that in nvf match. However, a more conventional way of doing that is to use `fork_rng`, e.g., https://github.com/pytorch/pytorch/blob/784a6ec5a30bd2d1831cb4f78183ad51696794e5/torch/jit/_trace.py#L282-L284. `seed` is supposed to be set up once and at the beginning of the program.

jjsjann123: Thanks for pointing out the example as well. :bow:

Priya2698: Thanks! I was not aware of this, updated the tests now.

tests/python/test_python_frontend.py
Priya2698: Thanks, Jie. This was the way to go, the test looks much more concise now.

Priya2698:
!build

tests/python/test_python_frontend.py
```diff
@@ -4458,6 +4457,86 @@ def fusion_func(fd: FusionDefinition):
...
+                # Reset the FusionCache or the fusion would not recompile for all subtests, failing checks in exec_nvfuser.
+                FusionCache.reset()
+                has_dropout = True if dropout_p is not None else False
+                has_causal = True if is_causal is not None else False
+                has_scale = True if scale is not None else False
+                inputs = [*qkv]
+                for param in [dropout_p, is_causal, scale]:
+                    if param is not None:
+                        inputs.append(param)
+                with torch.random.fork_rng(range(torch.cuda.device_count())):
```
wujingyue, line 4520: According to https://pytorch.org/docs/stable/random.html#torch.random.fork_rng, fork_rng by default operates on all devices, so the `range(...)` part is unnecessary.

tests/python/test_python_frontend.py
```diff
@@ -4458,6 +4457,86 @@ def fusion_func(fd: FusionDefinition):
...
+        qkv = [
+            torch.randn((N, H, L, E), dtype=torch.bfloat16, device="cuda:0"),
+            torch.randn((N, H, S, E), dtype=torch.bfloat16, device="cuda:0"),
+            torch.randn((N, H, S, E), dtype=torch.bfloat16, device="cuda:0"),
+        ]
+
+        dropout_vals = [None, 0.0, 0.2]
+        is_causal_vals = [None, True, False]
+        scale_vals = [None, 1 / E**0.5, 1e-3]
+        torch.manual_seed(0)
```
wujingyue, line 4505: Is this necessary and why?

tests/python/test_python_frontend.py
Priya2698: That's strange. I got an error that since there are multiple devices, I need to specify an ID.

tests/python/test_python_frontend.py
Priya2698: 1. Setting torch.manual_seed(0) since it is currently set within `exec_nvfuser`.
Will refactor that to set for the whole class in PR #2701. A constant seed is required for testing with non-zero dropout.

2. Values for dropout, is_causal, scale are to exercise different combinations of these parameters and verify correctness

tests/python/test_python_frontend.py
wujingyue: The doc says
> will emit a warning if your machine has a lot of devices, since this function will run very slowly in that case.
 
Could that be the reason? I've no idea what "a lot" means. 

Anyhow, this test doesn't use multiple GPUs, so it should be enough to specify just torch.cuda.current_device().

tests/python/test_python_frontend.py
wujingyue: > torch.manual_seed(0) is currently set within exec_nvfuser.

I missed that fact -- thanks for clarifying. However, I'm even more confused on how we are using RNG in the test. 
1. This particular `torch.manual_seed(0)` is done before the first subtest. But `exec_nvfuser` is executed before each subtest. There seems to be an asymmetry and I don't understand how they are logically connected. 
2. `fork_rng` should surround either actual or ref, but not both (at least not necessary). It essentially snapshots the state at entry and restores that state at exit. This restore is only needed between actual and ref, but not after them.

wujingyue: I'm unsure about this, but I now start to think it may be too early to use fork_rng before we remove the manual_seed(0) from exec_nvfuser. I suspect how we use RNG today is broken based on the previous comment. Although fork_rng is the right solution, adding that on top of the brokenness **at this very moment** may create more troubles.

wujingyue: (happy to discuss live)

jjsjann123: I *think* the issue here is that we are running two references....

i.e. `exec_nvfuser` runs a reference for serde, while this test runs another reference with torch.

We should probably be using two `fork_rng` inside exec_nvfuser for nvfuser run and the serde reference. That'll render the manual seed here obselete and we can also drop both fork_rng in this test

More than happy to have that in a separate PR. :stuck_out_tongue_winking_eye:

wujingyue: Got it. 

Sorry that I opened the can of worm. I'm now convinced that we'd better attempt fork_rng in a separate PR and only use manual_seed in this PR. The required change for fork_rng is larger than I expected, and a mix of fork_rng and manual_seed as is is confusing.

wujingyue: Notes to myself: currently, we run five implementations of a fusion and expect them to match: 
```
manual_seed(0)
run fusion without serde
manual_seed(0)
run fusion with serde
manual_seed(0)
run captured fusion without serde
manual_seed(0)
run captured fusion with serde
manual_seed(0)
run reference implementation 
```

Priya2698:
!build

Priya2698:
!build

PR #2484, opened 2024-06-27
Fix TMA failure in CI

zasdfgbnm:
This is a bug introduced in https://github.com/NVIDIA/Fuser/commit/5037d8a6a8f2056271282f87a20918142e2e375c but only get exposed by https://github.com/NVIDIA/Fuser/commit/57e3d4cc2d87f420fffb4b8f5a5a8a5bb352396c
cc: @protonu @rdspring1

### DIFF
--- a/csrc/device_lower/analysis/tma.cpp
+++ b/csrc/device_lower/analysis/tma.cpp
@@ -407,10 +407,8 @@ TMAInfo getTMAInfo(LoadStoreOp* ldst) {
                : (tma_g_to_tile_stride_g.count(g) ? SB : CB));
   };
   // merge contiguous C groups and CB groups
-  int64_t i = 0;
-  while (i < (int64_t)tma_domain.size() - 1) {
+  for (int64_t i = 0; i < (int64_t)tma_domain.size() - 1; i++) {
     if (!contiguity[i]) {
-      i++;
       continue;
     }
     bool is_c = (gtype(i) == C && gtype(i + 1) == C);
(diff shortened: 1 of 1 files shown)


csrc/device_lower/analysis/tma.cpp
```diff
@@ -423,12 +421,11 @@ TMAInfo getTMAInfo(LoadStoreOp* ldst) {
...
         tma_g_to_box_g.emplace(g, g);
       }
-    } else {
-      i++;
+      i--;
     }
   }
   // merge contiguous C with SB/CB
-  for (auto i : c10::irange((int64_t)tma_domain.size() - 1)) {
+  for (int64_t i = 0; i < (int64_t)tma_domain.size() - 1; i++) {
```
zasdfgbnm, line 428: `tma_domain.size()` is changing over time, so we can not use `irange`.

rdspring1 (approved):
LGTM.

zasdfgbnm:
!build

csrc/device_lower/analysis/tma.cpp
naoyam: Do you mean things like line 436? Surprised that would impact `irange`.

zasdfgbnm: Yes. `irange` compute the size at the beginning, and use it for the entire loop. It does not dynamically check if size has changed.

naoyam: Ah, I missed you also added `i--`. Now I got it.

naoyam (approved):

Issue #2481, opened 2024-06-27
Concretization error from communication segment followed by compute segment. 


cowanmeg:
I am seeing a concretization error where an IterDomain extent is not bound to a constant. This error only occurs when we have a reduction based collective (like allreduce/reduce scatter) followed by computation in another segment.

An example looks like this: 
---- local matmul ----
T4_g[ ideviceIdx.x10{i0}, iS11{i3}, iS12{i6}, rS13{i2} ] (DeviceMesh{0 1})
   = matmul(T3_l[ ideviceIdx.x7{i0}, iS9{i3}, iS8{i2} ] (DeviceMesh{0 1}),
            T1_g[ ideviceIdx.x3{i4}, iS4{i5}, iS5{i6} ] (DeviceMesh{0 1}))

---- allreduce --- (dispatched to ProcessGroup)
T5_g[ rS14{i0}, iS15{i3}, iS16{i6} ] (DeviceMesh{0 1} = reduce([ideviceIdx.x3{i4}, iS4{i5}, iS5{i6} DeviceMesh{0 1}))

---- bias add --- Note: rS14{i0} is the Val that triggers the error. 
%kernel {
T7_l[ iS19{i3}, iS20{i6} ] (DeviceMesh{0 1})
   = __half2float(T5_g[ **rS14{i0}**, iS15{i3}, iS16{i6} ] (DeviceMesh{0 1}));
T6_l[ bS17{1}, iS18{i7} ] (DeviceMesh{0 1})
   = broadcast( T2_l[ iS6{i7} ] (DeviceMesh{0 1}) )
T8_l[ bS21{1}, iS22{i7} ] (DeviceMesh{0 1})
   = __half2float(T6_l[ bS17{1}, iS18{i7} ] (DeviceMesh{0 1}));
T9_g[ iS23{i3}, iS24{i6} ] (DeviceMesh{0 1})
   = T7_l[ iS19{i3}, iS20{i6} ] (DeviceMesh{0 1})
   + T8_l[ bS21{1}, iS22{i7} ] (DeviceMesh{0 1});

Error:
C++ exception with description "ext_opt.hasValue() INTERNAL ASSERT FAILED at "/csrc/dynamic_transform.cpp":279, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. Could not evaluate dynamic extent: i0
[...]

PR #2487, opened 2024-06-28
Strip reduction axes in segmentation edges

jacobhinkle:
Currently, when we segment at the output of a reduction, the consumer segment will have an input tensor that has a `Reduction` axis in it. This can be problematic; see #2481 and #2317. This PR strips reduction axes from the root and allocation domain in these cases.

### DIFF
--- a/csrc/fusion_segmenter.cpp
+++ b/csrc/fusion_segmenter.cpp
@@ -1914,15 +1914,17 @@ void eraseInputDistinctRootDomains(Fusion* fusion) {
     }
     if (has_reduction) {
       if (new_td->hasAllocation()) {
+        const std::vector<IterDomain*> new_logical =
+            TensorDomain::noReductions(new_td->logical());
         new_td = IrBuilder::create<TensorDomain>(
-            std::vector<IterDomain*>{},
-            TensorDomain::noReductions(new_td->logical()),
-            no_red_alloc,
-            std::vector<IterDomain*>{},
-            no_red_contiguity);
+            /*root_domain=*/std::vector<IterDomain*>{},
+            /*logical_domain=*/new_logical,
+            /*allocation=*/no_red_alloc,
+            /*loop_domain=*/new_logical,
+            /*contiguity=*/no_red_contiguity);
       } else {
-        new_td =
-            IrBuilder::create<TensorDomain>(no_red_alloc, no_red_contiguity);
+        new_td = IrBuilder::create<TensorDomain>(
+            /*logical_domain=*/no_red_alloc, /*contiguity=*/no_red_contiguity);
       }
     }
 
--- a/csrc/fusion_segmenter.cpp
+++ b/csrc/fusion_segmenter.cpp
@@ -1900,31 +1900,27 @@ void eraseInputDistinctRootDomains(Fusion* fusion) {
     }
 
     // Remove reduction domains from new_td
-    std::vector<IterDomain*> no_red_alloc;
-    std::vector<std::optional<bool>> no_red_contiguity;
-    bool has_reduction = false;
-    for (size_t i : c10::irange(new_td->maybeAllocation().size())) {
-      IterDomain* id = new_td->maybeAllocation()[i];
-      if (id->isReduction()) {
-        has_reduction = true;
-        continue;
+    if (new_td->hasReduction()) {
+      std::vector<std::optional<bool>> no_red_contiguity;
+      for (size_t i : c10::irange(new_td->maybeAllocation().size())) {
+        if (new_td->maybeAllocation()[i]->isReduction()) {
+          continue;
+        }
+        no_red_contiguity.push_back(new_td->contiguity()[i]);
       }
-      no_red_alloc.push_back(id);
-      no_red_contiguity.push_back(new_td->contiguity()[i]);
-    }
-    if (has_reduction) {
       if (new_td->hasAllocation()) {
         const std::vector<IterDomain*> new_logical =
             TensorDomain::noReductions(new_td->logical());
         new_td = IrBuilder::create<TensorDomain>(
             /*root_domain=*/std::vector<IterDomain*>{},
             /*logical_domain=*/new_logical,
-            /*allocation=*/no_red_alloc,
+            /*allocation=*/TensorDomain::noReductions(new_td->allocation()),
             /*loop_domain=*/new_logical,
             /*contiguity=*/no_red_contiguity);
       } else {
         new_td = IrBuilder::create<TensorDomain>(
-            /*logical_domain=*/no_red_alloc, /*contiguity=*/no_red_contiguity);
+            /*logical_domain=*/TensorDomain::noReductions(new_td->logical()),
+            /*contiguity=*/no_red_contiguity);
       }
     }
 
--- a/csrc/fusion_segmenter.cpp
+++ b/csrc/fusion_segmenter.cpp
@@ -1822,12 +1822,12 @@ std::string toString(const SegmentedFusion* segmented_fusion) {
   return ss.str();
 }
 
-//! Sets the logical as root and erases logical of all inputs in fusion. Any
+//! Sets the root as logical and erases root of all inputs in fusion. Any
 //! non-constant expressions in those extents are replaced by new scalars with
 //! no definition. These mutations are performed throughout the Fusion so that
 //! downstream expressions dependent on the original inputs' logical extents can
 //! be computed properly.
-void convertInputLogicalsToRoots(Fusion* fusion) {
+void eraseInputDistinctRootDomains(Fusion* fusion) {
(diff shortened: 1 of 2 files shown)


jacobhinkle:
!build --diff

csrc/fusion_segmenter.cpp
```diff
@@ -1822,12 +1822,12 @@ std::string toString(const SegmentedFusion* segmented_fusion) {
...
 }
 
-//! Sets the logical as root and erases logical of all inputs in fusion. Any
+//! Sets the root as logical and erases root of all inputs in fusion. Any
 //! non-constant expressions in those extents are replaced by new scalars with
 //! no definition. These mutations are performed throughout the Fusion so that
 //! downstream expressions dependent on the original inputs' logical extents can
 //! be computed properly.
-void convertInputLogicalsToRoots(Fusion* fusion) {
+void eraseInputDistinctRootDomains(Fusion* fusion) {
```
jacobhinkle, line 1830: I'm open to suggestions on the new name of this function but since we are actually clearing out the root domain now, I think it should be changed.

csrc/fusion_segmenter.cpp
```diff
@@ -1898,6 +1898,36 @@ void convertInputLogicalsToRoots(Fusion* fusion) {
...
+    for (size_t i : c10::irange(new_td->maybeAllocation().size())) {
+      IterDomain* id = new_td->maybeAllocation()[i];
+      if (id->isReduction()) {
+        has_reduction = true;
+        continue;
+      }
+      no_red_alloc.push_back(id);
+      no_red_contiguity.push_back(new_td->contiguity()[i]);
+    }
+    if (has_reduction) {
```
wujingyue, line 1915: Nit: no_red_contiguity is only used when has_reduction is true. So how about

```
if (TensorDomain::hasReduction(new_td)) {
  compute no_red_contiguity
  ...
}
```

wujingyue (commented):
LGTM overall. I assume you'll add a test?

csrc/fusion_segmenter.cpp
```diff
@@ -1822,12 +1822,12 @@ std::string toString(const SegmentedFusion* segmented_fusion) {
   return ss.str();
 }
 
-//! Sets the logical as root and erases logical of all inputs in fusion. Any
+//! Sets the root as logical and erases root of all inputs in fusion. Any
```
jjsjann123, line 1825: :clap:

csrc/fusion_segmenter.cpp
```diff
@@ -1898,6 +1898,36 @@ void convertInputLogicalsToRoots(Fusion* fusion) {
...
+      no_red_contiguity.push_back(new_td->contiguity()[i]);
+    }
+    if (has_reduction) {
+      if (new_td->hasAllocation()) {
+        const std::vector<IterDomain*> new_logical =
+            TensorDomain::noReductions(new_td->logical());
+        new_td = IrBuilder::create<TensorDomain>(
+            /*root_domain=*/std::vector<IterDomain*>{},
+            /*logical_domain=*/new_logical,
+            /*allocation=*/no_red_alloc,
```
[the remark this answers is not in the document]
jjsjann123: I think the trick here is that we need to match no_red_alloc with no_red_contiguity. So we need to know which entry to remove in the contiguity vector.

csrc/fusion_segmenter.cpp
```diff
@@ -1898,6 +1898,36 @@ void convertInputLogicalsToRoots(Fusion* fusion) {
       new_td = IrBuilder::create<TensorDomain>(
           new_logical_domain, tv->domain()->contiguity());
     }
+
+    // Remove reduction domains from new_td
+    std::vector<IterDomain*> no_red_alloc;
+    std::vector<std::optional<bool>> no_red_contiguity;
+    bool has_reduction = false;
```
jjsjann123, line 1905: nitpick: https://github.com/NVIDIA/Fuser/blob/cfc8edbef1709ef92e83d9967ce30ea26bebfa35/csrc/ir/internal_base_nodes.h#L516-L518

csrc/fusion_segmenter.cpp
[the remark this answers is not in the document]
jacobhinkle: It is equivalent to that. I can move the local variable into the `if (has_reduction)` scope.

csrc/fusion_segmenter.cpp
```diff
@@ -1898,6 +1898,36 @@ void convertInputLogicalsToRoots(Fusion* fusion) {
...
+    for (size_t i : c10::irange(new_td->maybeAllocation().size())) {
```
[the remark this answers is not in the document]
jacobhinkle: Good question. Just in case, we can check both `allocation()` and `logical()`.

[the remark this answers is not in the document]
jacobhinkle: On the other hand, just checking `logical()` should be enough to determine if there is any reduction domain I guess, since we'd definitely not expect to find a reduction axis in the allocation domain if there are none in the logical domain.

jacobhinkle:
Thanks for the suggestions, gents. I incorporated them and added a test.

tests/cpp/test_segmentation.cpp
```diff
@@ -601,4 +601,49 @@ TEST_F(SegmentationTest, codeGenSupportedMergeIssue1970) {
...
+  EXPECT_EQ(segmented_fusion->groups().size(), 3);
+
+  for (SegmentedGroup* group : segmented_fusion->groups()) {
+    std::unique_ptr<Fusion> segment_fusion =
+        segmented_fusion->makeFusion(group).second;
+
+    TensorView* segment_input =
+        segment_fusion->inputs().at(0)->as<TensorView>();
+
+    EXPECT_FALSE(segment_input->domain()->hasReduction());
```
jacobhinkle, line 645: This check fails for both segmentation edges if I disable the reduction removal loop.

tests/cpp/test_segmentation.cpp
```diff
@@ -601,4 +601,49 @@ TEST_F(SegmentationTest, codeGenSupportedMergeIssue1970) {
...
+TEST_F(SegmentationTest, EraseReductionsInSegmentationEdges) {
+  auto fusion = std::make_unique<Fusion>();
+  FusionGuard fg(fusion.get());
+
+  auto tv0 = makeContigConcreteTensor({3, 32, 17}, DataType::Float);
+  fusion->addInput(tv0);
+
+  auto* tv1 = sum(tv0, {2});
+  tv1->setAllocationDomain(
+      {tv1->axis(1), tv1->axis(2), tv1->axis(0)}, {true, std::nullopt, true});
```
jacobhinkle, line 615: Exercises the `hasAllocation()` path.

jjsjann123 (approved):
LGTM

jacobhinkle:
!build

tests/cpp/test_segmentation.cpp
```diff
@@ -601,4 +601,49 @@ TEST_F(SegmentationTest, codeGenSupportedMergeIssue1970) {
...
+  auto options = at::TensorOptions().dtype(at::kFloat).device(at::kCUDA, 0);
+  auto in0 = at::randn({3, 32, 17}, options);
+  auto outputs = fec.runFusionWithInputs({in0});
+
+  testValidate(fec.fusion(), outputs, {in0}, __LINE__, __FILE__);
+
+  const FusionKernelRuntime* runtime = fec.getMostRecentKernelRuntime();
+  ASSERT_NE(runtime, nullptr);
+
+  EXPECT_TRUE(runtime->isSegmented());
```
wujingyue, line 632:
```suggestion
```

I believe line 636 implies this.

tests/cpp/test_segmentation.cpp
```diff
@@ -601,4 +601,49 @@ TEST_F(SegmentationTest, codeGenSupportedMergeIssue1970) {
...
+  FusionExecutorCache fec(std::move(fusion));
+
+  auto options = at::TensorOptions().dtype(at::kFloat).device(at::kCUDA, 0);
+  auto in0 = at::randn({3, 32, 17}, options);
+  auto outputs = fec.runFusionWithInputs({in0});
+
+  testValidate(fec.fusion(), outputs, {in0}, __LINE__, __FILE__);
+
+  const FusionKernelRuntime* runtime = fec.getMostRecentKernelRuntime();
+  ASSERT_NE(runtime, nullptr);
```
wujingyue, line 630:
```suggestion
  ASSERT_TRUE(runtime != nullptr);
```

Super nit: The main benefit of `ASSERT_NE` and alike is to print out the runtime values of the two operands. In this case, printing out the exact address of `runtime` doesn't help and is a slight distraction.

wujingyue (approved):

PR #2490, opened 2024-06-28
preseg optimization pass - merge neighboring `PadOp`

jjsjann123:
multiple `cat` would ended up having their `PadOp` being pushed to the same spot giving us a chain of `PadOp`. The added fusion ir pass tries to merge neighboring `PadOp` as one.

### DIFF
--- a/csrc/preseg_passes/move_pad.cpp
+++ b/csrc/preseg_passes/move_pad.cpp
@@ -448,6 +448,7 @@ void decomposeCatOp(Fusion* fusion) {
   }
 }
 
+// This pass merges neighboring `PadOp` when possible (identical pad value).
 void mergeNeighboringPad(Fusion* fusion) {
   std::vector<Expr*> exprs = fusion->exprs();
   // traverse in topo order. We'll merge neighboring pad and replace the uses of


csrc/preseg_passes/move_pad.cpp
```diff
@@ -448,10 +448,59 @@ void decomposeCatOp(Fusion* fusion) {
...
+void mergeNeighboringPad(Fusion* fusion) {
+  std::vector<Expr*> exprs = fusion->exprs();
+  // traverse in topo order. We'll merge neighboring pad and replace the uses of
+  // consumer pad with the merged producer. So it would not interfere traversal.
+  for (auto* producer : ir_utils::filterByType<PadOp>(exprs)) {
+    while (producer) {
+      Val* pad_out = producer->out();
+      if (pad_out->uses().size() != 1 || !pad_out->uses()[0]->isA<PadOp>()) {
+        break;
+      }
```
jacobhinkle, line 461: I think we might also need to prove that each pad width has the same sign in producer and consumer, or at least that the producer sign is positive. Otherwise, we might have something like the following:
```
auto tv1 = pad(tv0, {0, -1}, 0);
auto tv2 = pad(tv1, {0, 1}, 0); // same as tv0 with last element replaced by 0
```
If we're not careful here we would replace this with a pad by `{0, 0}`. This simple example could sneak by us when using symbolic pad widths.

jjsjann123: Thanks for pointing this out (as well as the comment [here](https://github.com/NVIDIA/Fuser/pull/2373/files/8678757de25a003cf13f37bd177b7a8e9381f4cb#r1664482979)). I was fooled by the comment https://github.com/NVIDIA/Fuser/pull/2373#discussion_r1664482979.

```
auto tv1 = pad(tv0, {0, -1}, 0);
auto tv2 = pad(tv1, {0, 1}, 0); // same as tv0 with last element replaced by 0
```
I'm wondering if we can still merge the two pads even with symbolic pad widths. i.e. shouldn't concretization have already inserted logic to ensure that runtime value isn't going to change the semantics? I'll try to chat with you offline on this one.

jacobhinkle: Hmm, but we don't have a single op that can set just the last element to 0, so what would the merged case be in this case? I agree that we can handle symbolic pad widths I think, it just might need some help. You're right that this could potentially become a concretization thing, similar to #511 .

jjsjann123: > but we don't have a single op that can set just the last element to 0

oh sorry I misread that earlier. It's the shrink then pad which alters the value in the output. Yeah we cannot do that. But we would be able to merge the other direction. I'll fix it up.

Thanks a ton for the example. :bow:

jjsjann123:
in the later refactor, it no longer makes sense to keep this PR separate.

But I did address @jacobhinkle's comment in the original PR.

PR #2491, opened 2024-06-28
HostIrContainer: add default stream

samnordmann:
# What
add in `HostIrContainer` an IR representation of the "default stream" as defined by torch

# Why

used in the overlap implementation
https://github.com/NVIDIA/Fuser/blob/3258fa5d3bf9e89c600c1615f6dbc24b00b8dcce/tests/cpp/test_multidevice_overlap.cpp#L220

# How

`IrContainer::oneVal()` is a similar concept as what we try to achieve here. But the implementation there has some complexity
https://github.com/NVIDIA/Fuser/blob/3258fa5d3bf9e89c600c1615f6dbc24b00b8dcce/csrc/ir/container.cpp#L266
in particular, `oneVal` is not built with `IrBuilder::create` as one would expect, though I am not sure to understand the benefit.

Here, I didn't care to reproduce the implementation of `oneVal`, I propose something simpler relying on `IrBuilder::create`. Let me know if you prefer something in the same vein as the implementation of `oneVal`.

### DIFF
--- a/csrc/host_ir/executor.cpp
+++ b/csrc/host_ir/executor.cpp
@@ -27,7 +27,8 @@ HostIrExecutor::HostIrExecutor(
       : 0;
   streams_.insert(
       {container_->getDefaultStream(),
-       c10::cuda::getDefaultCUDAStream(device_index)});
+       c10::cuda::getDefaultCUDAStream(
+           static_cast<c10::DeviceIndex>(device_index))});
 }
 
 std::vector<at::Tensor> HostIrExecutor::runWithInput(
--- a/csrc/host_ir/executor.cpp
+++ b/csrc/host_ir/executor.cpp
@@ -20,7 +20,7 @@ HostIrExecutor::HostIrExecutor(
     HostIrExecutorParams params)
     : container_(std::move(container)),
       communicator_(communicator),
-      params_(std::move(params)) {
+      params_(params) {
   auto device_index =
       (communicator_ != nullptr && communicator_->is_available())
       ? communicator_->deviceId()
--- a/csrc/host_ir/executor.cpp
+++ b/csrc/host_ir/executor.cpp
@@ -20,7 +20,15 @@ HostIrExecutor::HostIrExecutor(
     HostIrExecutorParams params)
     : container_(std::move(container)),
       communicator_(communicator),
-      params_(params) {}
+      params_(params) {
+  auto device_index =
+      (communicator_ != nullptr && communicator_->is_available())
+      ? communicator_->deviceId()
+      : 0;
+  streams_.insert(
+      {container_->getDefaultStream(),
+       c10::cuda::getDefaultCUDAStream(device_index)});
+}
 
 std::vector<at::Tensor> HostIrExecutor::runWithInput(
     std::unordered_map<Val*, c10::IValue> val_to_IValue) {
--- a/csrc/host_ir/executor.cpp
+++ b/csrc/host_ir/executor.cpp
@@ -20,7 +20,7 @@ HostIrExecutor::HostIrExecutor(
     HostIrExecutorParams params)
     : container_(std::move(container)),
       communicator_(communicator),
-      params_(params) {
+      params_(std::move(params)) {
   auto device_index =
       (communicator_ != nullptr && communicator_->is_available())
       ? communicator_->deviceId()
--- a/tests/cpp/test_host_irs.cpp
+++ b/tests/cpp/test_host_irs.cpp
@@ -487,6 +487,33 @@ TEST_F(NVFuserTest, HostIrSetStream) {
       c10::cuda::getDefaultCUDAStream(0), c10::cuda::getCurrentCUDAStream(0));
 }
 
+// The following test simply demonstrate how to change current CUDA stream in
+// the host program
+TEST_F(NVFuserTest, HostIrDefaultStream) {
+  auto change_stream = [](bool use_default_stream) {
+    auto hic = std::make_unique<HostIrContainer>();
+    Stream* stream;
+    if (use_default_stream) {
+      stream = hic->getDefaultStream();
+    } else {
+      stream = IrBuilder::createInContainer<Stream>(hic.get());
+    }
+    auto set_stream =
+        IrBuilder::createInContainer<SetCurrentStream>(hic.get(), stream);
+    hic->pushBackTopLevelExprs(set_stream);
+    HostIrExecutor hie(std::move(hic));
+    hie.runWithInput({});
+  };
+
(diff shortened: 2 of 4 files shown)


samnordmann:
!build

csrc/host_ir/executor.cpp
```diff
@@ -20,7 +20,16 @@ HostIrExecutor::HostIrExecutor(
     HostIrExecutorParams params)
     : container_(std::move(container)),
       communicator_(communicator),
-      params_(params) {}
+      params_(params) {
+  auto device_index =
+      (communicator_ != nullptr && communicator_->is_available())
+      ? communicator_->deviceId()
+      : 0;
```
wujingyue, line 27: This seems a wrong assumption. Even when `communicator_` not being available, the machine may still have multiple devices and the user may want to run the fusion on a particular device (and therefore not always device 0). 

@naoyam what's your suggestion? We could compute the device index from input tensors, similar to FusionExecutorCache. We could also make the device index a config of HostIrExecutor and therefore pass it into the constructor.

wujingyue (commented):
Given the existing SetCurrentStream IR, I've yet to understand the full value of a default stream. It sounds like the writer of host IR can always create a stream and SetCurrentStream and keep track of it so it can be reused. Is the benefit to avoid even having to SetCurrentStream before posting anything?

tests/cpp/test_host_irs.cpp
```diff
@@ -487,6 +487,33 @@ TEST_F(NVFuserTest, HostIrSetStream) {
       c10::cuda::getDefaultCUDAStream(0), c10::cuda::getCurrentCUDAStream(0));
 }
 
+// The following test simply demonstrate how to change current CUDA stream in
+// the host program
+TEST_F(NVFuserTest, HostIrDefaultStream) {
```
wujingyue, line 492: This and the previous test should probably avoid using NVFuserTest as the fixture **name** -- NVFuserTest has been too overloaded despite that many legacy tests in test_gpu*.cpp still use it. 

One possibility is
```
TEST_F(HostIrTest, DefaultStream) {
```
Even though `HostIrTest` has parameters, it's OK to use it in TEST_F as long as GetParam isn't called. 

(You can use a different fixture name that you like. It's really just NVFuserTest I wanted to avoid.)

tests/cpp/test_host_irs.cpp
```diff
@@ -487,6 +487,33 @@ TEST_F(NVFuserTest, HostIrSetStream) {
...
+    }
+    auto set_stream =
+        IrBuilder::createInContainer<SetCurrentStream>(hic.get(), stream);
+    hic->pushBackTopLevelExprs(set_stream);
+    HostIrExecutor hie(std::move(hic));
+    hie.runWithInput({});
+  };
+
+  setCurrentCUDAStream(c10::cuda::getDefaultCUDAStream(0));
+  change_stream(false);
```
wujingyue, line 509:
```suggestion
  change_stream(/*use_default_stream=*/false);
```

tests/cpp/test_host_irs.cpp
```diff
@@ -487,6 +487,33 @@ TEST_F(NVFuserTest, HostIrSetStream) {
...
+  EXPECT_NE(
+      c10::cuda::getDefaultCUDAStream(0), c10::cuda::getCurrentCUDAStream(0));
+  change_stream(true);
```
wujingyue, line 512: ditto

csrc/host_ir/executor.cpp
samnordmann: right, let me know then.

Notice that the same situation is already present in the current code
https://github.com/NVIDIA/Fuser/blob/4e7365d18973819b9859f4b9ac498269257762a2/csrc/host_ir/executor.cpp#L49

tests/cpp/test_host_irs.cpp
samnordmann: > Even though HostIrTest has parameters, it's OK to use it in TEST_F as long as GetParam isn't called.

Doing this, I'm getting an error:

```
In file included from /opt/pytorch/Fuser/third_party/googletest/googletest/include/gtest/gtest-printers.h:122,
                 from /opt/pytorch/Fuser/third_party/googletest/googletest/include/gtest/gtest-matchers.h:49,
                 from /opt/pytorch/Fuser/third_party/googletest/googletest/include/gtest/internal/gtest-death-test-internal.h:47,                                                             from /opt/pytorch/Fuser/third_party/googletest/googletest/include/gtest/gtest-death-test.h:43,
                 from /opt/pytorch/Fuser/third_party/googletest/googletest/include/gtest/gtest.h:65,
                 from /opt/pytorch/Fuser/tests/cpp/test_host_irs.cpp:8:                                                                                                      /opt/pytorch/Fuser/tests/cpp/test_host_irs.cpp:480:1: error: ‘nvfuser::hir::StreamHostIrTestParams_HostIrSetStream_Test::~StreamHostIrTestParams_HostIrSetStream_Test()’ mark
ed ‘override’, but does not override
  480 | TEST_F(StreamHostIrTestParams, HostIrSetStream) {
      | ^~~~~~
```

Let me use a different class.

samnordmann:
!build

samnordmann:
> Given the existing SetCurrentStream IR, I've yet to understand the full value of a default stream. To achieve the same effect, it sounds like the writer of host IR can always create a stream and SetCurrentStream and keep track of it so it can be reused.
> Is the benefit to avoid even having to SetCurrentStream before posting anything?


You are correct. The only benefit is to be able to use torch's default stream, which I am not sure plays a special role besides being set by default.

Tbh, I was under the wrong impression that the default stream represented the user's stream, but I'm realizing this is not accurate. Saving the user stream might make more sense. We should define better what's Fuser's policy about the user's stream.

Let me know if you think it is still worth merging this pr, or if I should make it a draft in the meantime

csrc/host_ir/executor.cpp
```diff
@@ -20,7 +20,16 @@ HostIrExecutor::HostIrExecutor(
     HostIrExecutorParams params)
     : container_(std::move(container)),
       communicator_(communicator),
-      params_(params) {}
+      params_(params) {
```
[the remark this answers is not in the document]
samnordmann: CI complains: 
https://github.com/NVIDIA/Fuser/actions/runs/9743300305/job/26886443345?pr=2491

samnordmann:
!build

wujingyue (approved):

wujingyue:
> Let me know if you think it is still worth merging this pr, or if I should make it a draft in the meantime

As long as you think it's useful for your host IR experiment, let's merge it and revisit the decision later. It seems to be a strict improvement and the change is localized to host IR so the maintenance cost is controlled.

Issue #2463, opened 2024-06-26
change shmoo TEST_F to TEST_P


liqiangxl:
we have many shmoo tests using `TEST_F`, if fails, don't know which input fails. If change to `TEST_P` we can get a clear msg about which input para fails with `where GetParam() = `, e.g. 
```
[  FAILED  ] ReshapeBeforeReduction.FusionReshapeBeforeReduction/29, where GetParam() = ({ 2, 3, 4, 5 }, { 1, 6, 1, 2, 2, 15, 1 }) 
```

CI at #2437 detected a validation error for test `GpuViewTest.FusionReshapeReductionShmoo`, however, it doesn't tell which input para fails and I can't reproduce locally. If change to  `TEST_P` , we can get a better error msg.

PR #2495, opened 2024-06-28
change reductionShmoo tests to TEST_P

liqiangxl:
Part of #2463 
output sample:
```
 ReductionNoODimShmoo.FusionReductionSchedulerNoODimShmoo_CUDA/dtype_std__complex_double__rsize_1048576 (106 ms)
[ RUN      ] ReductionNoODimShmoo.FusionReductionSchedulerNoODimShmoo_CUDA/dtype___bfloat_rsize_1
```


tests/cpp/test_gpu1.cpp
```diff
@@ -5388,180 +5388,169 @@ TEST_F(NVFuserTest, FusionReductionSchedulerMultiDimFastest_CUDA) {
...
-      }
-      if (is_bf16) {
-        tv1_cast = castOp(DataType::BFloat16, tv1);
-      }
+// data type, reduction dim size, iteration dim size, reduction axis
+using ReductionWithIterDimParams =
+    std::tuple<DataType, int64_t, int64_t, int64_t>;
+using ReductionWithIterDimShmoo =
+    NVFuserFixtureParamTest<ReductionWithIterDimParams>;
+TEST_P(ReductionWithIterDimShmoo, FusionReductionSchedulerDimShmoo_CUDA) {
```
jacobhinkle, line 5474: We could probably remove `Shmoo` (and `_CUDA`) from the name. Something like this?
```suggestion
TEST_P(ReductionWithIterDim, Test) {
```

tests/cpp/test_gpu1.cpp
```diff
@@ -5388,180 +5388,169 @@ TEST_F(NVFuserTest, FusionReductionSchedulerMultiDimFastest_CUDA) {
...
+        testing::Values(160, 320), // iteration dimension size
+        testing::Values(0, 1)), // reduction axis
+    [](const testing::TestParamInfo<ReductionWithIterDimParams>& info)
+        -> std::string {
+      std::stringstream ss;
+      ss << "dtype_" << std::get<0>(info.param);
+      ss << "_rsize_" << std::get<1>(info.param);
+      ss << "_isize_" << std::get<2>(info.param);
+      ss << "_raxis_" << std::get<3>(info.param);
+      return sanitizeTestName(ss.str());
```
jacobhinkle, line 5551: I get that it's good practice, but I am curious if it is necessary to sanitize the test name in this case?

tests/cpp/test_gpu1.cpp
```diff
@@ -5388,180 +5388,169 @@ TEST_F(NVFuserTest, FusionReductionSchedulerMultiDimFastest_CUDA) {
...
-  // doing every other power of 2.
-  for (int i = 1; i <= 1024 * 1024; i <<= 2) {
-    red_dims.push_back(i);
+  TensorView* tv1_cast = tv1;
+  if (is_fp16) {
+    tv1_cast = castOp(DataType::Half, tv1);
+  }
+  if (is_bf16) {
+    tv1_cast = castOp(DataType::BFloat16, tv1);
   }
```
jacobhinkle, line 5427: Could you do this?
```suggestion
  TensorView* tv1_cast = SimplifyingIrBuilder::castOp(dtype, tv1);
```

tests/cpp/test_gpu1.cpp
```diff
@@ -5388,180 +5388,169 @@ TEST_F(NVFuserTest, FusionReductionSchedulerMultiDimFastest_CUDA) {
...
-  // Tried to cut down the number iterations with just
-  // doing every other power of 2.
-  for (int64_t i = 1; i <= 1024 * 1024; i <<= 2) {
-    red_dims.push_back(i);
+  TensorView* tv1_cast = tv1;
+  if (is_fp16) {
+    tv1_cast = castOp(DataType::Half, tv1);
+  }
+  if (is_bf16) {
+    tv1_cast = castOp(DataType::BFloat16, tv1);
```
jacobhinkle, line 5506: Same as above.

jacobhinkle: On second thought, I now see that it was like this before this change. Better to not make changes like that in this PR then.

jacobhinkle (approved):
LGTM

tests/cpp/test_gpu1.cpp
liqiangxl: initially I don't have `sanitizeTestName`, then I got an err msg due to `std::complex<float>`, which is not allowed in the name. So I added a function to sanitize the name. It should be safe if only test non-complex floating types.

PR #2500, opened 2024-06-29
Fix for when allocation domains are promoted to reduction domains

naoyam:
For non-global tensors, unless explicitly set, allocation domains are promotion domains of loop domains. Promotion can be a reduction domain even for non-reduction domains as there's no special handling for reduction domains. This can result in incorrect indexing with the current code. Specifically, we skip reduction domains from allocation, however, since we look at promotion domains, they may be reduction domains even though original loop domains are not. This can result in incorrectly skipping indexing. See the added unit test for a concrete example.

The fix in this PR is to use both the original loop domain and its promotion domain when determining the allocation domain of a given tensor to index.

### DIFF
--- a/csrc/id_model/indexing.cpp
+++ b/csrc/id_model/indexing.cpp
@@ -62,19 +62,16 @@ bool isSizeOneDomain(IterDomain* id) {
   return id->isBroadcast() || id->extent()->isOneInt();
 }
 
-// True if a given domain of a tensor *may* require allocation. The
-// optional promotion domain is also considered if given.
-bool mayRequireAllocation(
-    TensorView* tv,
-    IterDomain* id,
-    IterDomain* promotion_id = nullptr) {
+// True if a given domain of a tensor *may* require allocation
+bool mayRequireAllocation(TensorView* tv, IterDomain* id) {
   // Conditions to consider:
-  // - Fully partitioned: Check the original ID, not the promotion ID
-  // - Size one: Check the promotion ID if given, the original ID otherwise.
+  // - Fully partitioned
+  // - Size one: Allocation is done based on the promotion ID, but as
+  // long as the original ID has size one, its allocation should
+  // remain size one.
   // - Reduction: Check the original ID, not the promotion, which may
   //   be a reduction ID even though the original ID is not a reduction
-  return !isPartitionedLoop(tv, id) &&
-      !isSizeOneDomain(promotion_id != nullptr ? promotion_id : id) &&
+  return !isPartitionedLoop(tv, id) && !isSizeOneDomain(id) &&
       !id->isReduction();
 }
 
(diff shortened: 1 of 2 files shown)


naoyam:
!build

csrc/id_model/indexing.cpp
```diff
@@ -172,13 +190,14 @@ std::tuple<std::vector<IterDomain*>, std::vector<Val*>> getAllocationDomains(
...
-          allocation_domains.at(dim)->extent(), cur_contig_stride);
+          promotion_domain->extent(), cur_contig_stride);
     } else {
       // Assume that the tensor should always be a Global memory
       // tensor if it has non-contig allocation domains
       NVF_ERROR(tv->getMemoryType() == MemoryType::Global);
       strides[dim] = getStrideOfGlobalMemoryTensor(tv, (int64_t)dim);
-      cur_contig_stride = strides[dim];
+      cur_contig_stride = SimplifyingIrBuilder::mulExpr(
+          strides[dim], promotion_domain->extent());
```
naoyam, line 200: This is an unrelated bug fix.

naoyam:
CC: @zasdfgbnm

zasdfgbnm (commented):
Should we always not promote to reduction? That is, if promoted to reduction, we either find or create a non-reduction ID as the promotion result.

naoyam:
> Should we always not promote to reduction? That is, if promoted to reduction, we either find or create a non-reduction ID as the promotion result.

That's a good question, and I also considered it as well and decided to keep the loop promotion as is and address this issue as done in this PR. Two primary reasons:

- The fix as is done in this PR is relatively simple.
- The loop promotion is, as its name indicates, primarily about the loop extent of each iter domain, and in that sense non-reduction and reduction domains shouldn't be treated differently. Their allocation domains are different, but loop iteration counts should be the same. If we had something like "allocation promotion", yes, we would not promote non-reduction domains to reduction domains, but for loop promotions, I think the current definition seems to make most sense.

That said, if promotion of non-reduction domains to reductions domains turns out to be too inconvenient, I'd be happy to change the definition of the loop promotion 😃 , but for now I'd like to keep it as is.

csrc/id_model/indexing.cpp
```diff
@@ -59,12 +59,23 @@ bool isPartitionedLoop(TensorView* tv, IterDomain* id) {
...
-  return !isPartitionedLoop(tv, id) && !isSizeOneDomain(id);
+// True if a given domain of a tensor *may* require allocation. The
+// optional promotion domain is also considered if given.
+bool mayRequireAllocation(
+    TensorView* tv,
+    IterDomain* id,
+    IterDomain* promotion_id = nullptr) {
+  // Conditions to consider:
+  // - Fully partitioned: Check the original ID, not the promotion ID
+  // - Size one: Check the promotion ID if given, the original ID otherwise.
```
jacobhinkle, line 73: If `promotion_id` is not size 1 but `id` is size 1, why do we still need to allocate that dimension for `id`?

naoyam: That's a good point. As long as the original domain is size one, I believe we can keep it as is, so its allocation should not need to be expanded. Let me test the change.

naoyam: Cleaned it up. Testing the [tracking branch](#2238 ) also passed.

jacobhinkle (approved):
LGTM

PR #2502, opened 2024-06-29
remove iter gather

liqiangxl:
`IterType::Gather` is no longer used after #2299 , so maybe we should also remove releated code about `IterType::Gather` or is there some special reason to keep it?

### DIFF
--- a/csrc/type.cpp
+++ b/csrc/type.cpp
@@ -795,8 +795,6 @@ static const char* iter_type2string(IterType t) {
       return "r";
     case IterType::Broadcast:
       return "b";
-    case IterType::Gather:
-      return "g";
     case IterType::Stride:
       return "s";
     case IterType::GatherScatter:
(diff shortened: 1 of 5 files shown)


liqiangxl:
!build

naoyam:
Thanks. It should be removed. Can you also remove it from the `IterType` enum?

liqiangxl:
!build

naoyam (approved):
LGTM

csrc/type.cpp
```diff
@@ -795,8 +795,6 @@ static const char* iter_type2string(IterType t) {
       return "r";
     case IterType::Broadcast:
       return "b";
-    case IterType::Gather:
-      return "g";
     case IterType::Stride:
```
zasdfgbnm, line 798: Is `IterType::Stride` still used?

naoyam: Apparently, no. I think we used to have some conv-like tests by using that, like strided pooling. Stride is fairly simple, so I think we could just leave it there for now, but I'd be ok to remove it as well.

Issue #2497, opened 2024-06-28
Main process destroyed cuda before watchdog loop exited, terminating watchdog.


wujingyue:
This warning looks suspicious enough to file an issue. 

To reproduce the warning, `_bn && TORCH_CPP_LOG_LEVEL=INFO TORCH_DISTRIBUTED_DEBUG=INFO mpirun -np 2 bin/test_multidevice --gtest_filter=CommunicationTest.ReduceScatter/NCCL`. 

```
[I628 20:15:42.000370615 ProcessGroupNCCL.cpp:1501] [PG  Rank 1] main process destroyed cuda before watchdog loop exited, terminating watchdog. (Watchdog caught exception: CUDA error: driver shutting down
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1
Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.

Exception raised from c10_cuda_check_implementation at /opt/pytorch/pytorch/c10/cuda/CUDAException.cpp:43 (most recent call first):
frame #0: c10::Error::Error(c10::SourceLocation, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) + 0xac (0x79f03b38950c in /usr/local/lib/python3.10/dist-packages/torch/lib/libc10.so)
frame #1: c10::detail::torchCheckFail(char const*, char const*, unsigned int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) + 0xf3 (0x79f03b331757 in /usr/local/lib/python3.10/dist-packages/torch/lib/libc10.so)
frame #2: c10::cuda::c10_cuda_check_implementation(int, char const*, char const*, int, bool) + 0x3f2 (0x79f03b2b78c2 in /usr/local/lib/python3.10/dist-packages/torch/lib/libc10_cuda.so)
frame #3: c10d::ProcessGroupNCCL::WorkNCCL::finishedGPUExecutionInternal() const + 0x5e (0x79efff87fe8e in /usr/local/lib/python3.10/dist-packages/torch/lib/libtorch_cuda.so)
frame #4: c10d::ProcessGroupNCCL::WorkNCCL::isCompleted() + 0xb0 (0x79efff885130 in /usr/local/lib/python3.10/dist-packages/torch/lib/libtorch_cuda.so)
[...]

PR #2503, opened 2024-06-30
Delete the shared Communicator in TearDownTestSuite.

wujingyue:
Fixes #2497.

PR #2504, opened 2024-07-01
Destruct the singleton Communicator via testing::Environment. 

wujingyue:
Fixes https://github.com/NVIDIA/Fuser/issues/2497.

My other attempt (#2503) failed. There were errors that seem to be related to repeated create/delete the singleton Communicator, which I don't understand. This PR tries to create/delete the Communicator just once, keeping the behavior as close as before.

### DIFF
--- a/csrc/multidevice/communicator.cpp
+++ b/csrc/multidevice/communicator.cpp
@@ -214,6 +214,18 @@ Communicator::Communicator(
 #endif
 }
 
+Communicator::~Communicator() {
+#ifdef NVFUSER_DISTRIBUTED
+  for (auto& [key, backend] : backends_) {
+    // Call shutdown before destructing a ProcessGroupNCCL as instructed by
+    // https://github.com/pytorch/pytorch/blob/e62073d7997c9e63896cb5289ffd0874a8cc1838/torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp#L1164-L1170.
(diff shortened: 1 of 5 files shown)


wujingyue:
!build

csrc/multidevice/communicator.cpp
```diff
@@ -214,6 +214,18 @@ Communicator::Communicator(
 #endif
 }
 
+Communicator::~Communicator() {
+#ifdef NVFUSER_DISTRIBUTED
```
samnordmann, line 218:
```suggestion
#if(defined NVFUSER_DISTRIBUTED && defined USE_C10D_NCCL)
```

samnordmann (approved):
LGTM, Thanks for looking into this!

cowanmeg:
Does this PR supersede https://github.com/NVIDIA/Fuser/pull/2503 or are both still required?

wujingyue:
> Does this PR supersede #2503 or are both still required?

The former, as said in the PR description. I still believe #2503 should work. It currently leads to non-deterministic bugs that I couldn't understand, so I gave it up for now.

csrc/multidevice/communicator.cpp
wujingyue: Good catch!

wujingyue:
@xwang233 let me know how you like the cmake file changes.

wujingyue:
!build

xwang233 (approved):

PR #2469, opened 2024-06-26
Adding a Hopper Matmul test for single tile with A and B read from shared memory (SS)

protonu:
This adds a unit tests for a matmul (single tile) where both inputs are loaded to the shared memory. This builds on the PR (https://github.com/NVIDIA/Fuser/pull/2468)

PR #2508, opened 2024-07-01
[Hopper matmul] Refactor the code which schedules TMA loads for the input Operands

protonu:
This refactors the helpers functions in the unit tests added in the PRs https://github.com/NVIDIA/Fuser/pull/2469 and https://github.com/NVIDIA/Fuser/pull/2468 into functions that can later be used in the matmul scheduler.


protonu:
!build

protonu:
!build

csrc/ir/interface_nodes.h
```diff
@@ -409,6 +409,13 @@ class NVF_API TensorView : public Val {
   void applyMmaSwizzle(MmaOperand operand);
   void applyMmaSwizzle(MmaInputSmemSwizzle swizzle);
 
+  //! Transforms the innermost iterdomains according to the given mma swizzle,
+  //!  this should be used on the tvs that are inputs of a MmaOp and are loaded
+  //!  using TMA.
+  void applyMmaSwizzleForTMALoad(
```
zasdfgbnm, line 415: Could we also unify this function with this?
https://github.com/NVIDIA/Fuser/blob/4478853dd8001e00b99e5c4f489f3d06e4fecd06/tests/cpp/test_memory.cpp#L401-L413

protonu: Please take another look

protonu:
!build

protonu:
!build

csrc/ir/interface_nodes.h
```diff
@@ -409,6 +409,13 @@ class NVF_API TensorView : public Val {
   void applyMmaSwizzle(MmaOperand operand);
   void applyMmaSwizzle(MmaInputSmemSwizzle swizzle);
 
+  //! Transforms the innermost iterdomains according to the given mma swizzle,
+  //!  this should be used on the tvs that are inputs of a MmaOp and are loaded
```
zasdfgbnm, line 413:
```suggestion
  //!  this should be used on the tvs that are inputs of a MmaOp or are loaded
```

csrc/scheduler/mma_utils.cpp
```diff
@@ -837,6 +837,117 @@ void WarpMmaSwizzler::scheduleLdMatrix(TensorView* tv, MmaOperand operand) {
   setWarpMapped(tv, 2);
 }
 
+void WarpMmaSwizzler::parallelizeAsBulkSkippingFirstIDs(
+    TensorView* tv,
+    size_t first_ids_to_skip) {
```
zasdfgbnm, line 842: nit: should we always use `int64_t` for integers in all our functions? clang-tidy loves to complain about mismatching integer types and using the same integer type could help avoiding having to add casts between types.

tests/cpp/test_memory.cpp
```diff
@@ -457,7 +432,10 @@ void scheduleTile(
     // ...]
     tv->axis(0)->parallelize(ParallelType::BIDx);
     if (swizzle != MmaInputSmemSwizzle::None) {
-      scheduleTMASwizzle(tv, swizzle_size);
+      // In our implementation of swizzling, we work on 2D box where the
+      // inner-dim is the size of the swizzle in Bytes (at most).
+      tv->split(-1, tile_sizes[dim - 1]);
+      mma_utils::WarpMmaSwizzler::swizzleTMABox(tv, swizzle, false);
```
zasdfgbnm, line 438: Should we add this as a method of `TensorView`?

csrc/scheduler/mma_utils.cpp
```diff
@@ -837,6 +837,117 @@ void WarpMmaSwizzler::scheduleLdMatrix(TensorView* tv, MmaOperand operand) {
...
+    // [Ko, K8, No, N8]
+    tv->reorder({{-2, -3}});
+    // [Ko, No, K8, N8]
+    num_ids_to_skip += 2;
+  } else {
+    auto dtype = tv->getDataType().value();
+
+    // In the comments below I assume K=16, N=32, swizzle=32, dtype = half.
+    if (split_outer_dim) {
+      // [..., K(16),  N(326)] ->
```
zasdfgbnm, line 917:
```suggestion
      // [..., K(16),  N(32B)] ->
```

csrc/scheduler/mma_utils.cpp
```diff
@@ -837,6 +837,117 @@ void WarpMmaSwizzler::scheduleLdMatrix(TensorView* tv, MmaOperand operand) {
...
+    tv->split(-1, 8);
+    // [Ko, K8, No, N8]
+    tv->reorder({{-2, -3}});
+    // [Ko, No, K8, N8]
+    num_ids_to_skip += 2;
+  } else {
+    auto dtype = tv->getDataType().value();
+
+    // In the comments below I assume K=16, N=32, swizzle=32, dtype = half.
+    if (split_outer_dim) {
```
zasdfgbnm, line 916: Should we always do the split? This way, we will use this flag to control the reorder dim, instead of whether doing split.

protonu: While implementing this I realized this may have some drawback. Say I tile in both cases and reorder based on the boolean flag then the IDs look like:

[... NO(2), KO(2), KI(8), NI(16] {not split}
[... KO(2), NO(2), KI(8), NI(16)] {split}

Now when we then call `swizzleTMABox`, in that function I would need to remove the bit of code:

```
  if (!split_outer_dim) {
    // [..., KO(4), KI(4), NIO(2), NII(8)] -> [..., KOO(2), KOI(2) KI(4),
    // NIO(2), NII(8)]
    split(-4, (getBytesFromSwizzle(swizzle) / 16));
  }
```

That would mean that `swizzleTMABox` would expect the outer dim to have been tiled by 8 by it's users. I am not sure this is always preferable, as it might be nice to have users call this on a TMA box directly without tiling. I don't have strong opinions about this.

zasdfgbnm: Can we move the code that tile 8 into `swizzleTMABox`? Is there a case that we must not tile 8?

zasdfgbnm: This is the schedule of tile 8 vs not tile 8:
image
The only difference is the highlighted area. Even if the user didn't tile 8, we will tile 4 then 2, which according to Theorem 2.1 (Equivalence of Split-Split) in [The Mathematical Theory of IterDomain
](https://github.com/NVIDIA/Fuser/blob/main/doc/reading/iterdomain.md), this is equivalent to the hightlighted area when the user did tile 8. It sounds tedious to use a flag to switch between mathematically equivalent forms.

csrc/scheduler/mma_utils.cpp
```diff
@@ -837,6 +837,78 @@ void WarpMmaSwizzler::scheduleLdMatrix(TensorView* tv, MmaOperand operand) {
...
+    tv->reorder({{-2, -3}});
+
+    // [NO, K, NI] ->
+    // [NO, KO(2), KIO(2), KII(4), NIO(2), NII(8)]
+    tv->swizzleTMABox(swizzle);
+
+    // If the outer dim is split, then we pull out KO to be outside NO
+    // and KO and NO are both not marked bulk parallel, else NO is outer
+    // and only NO is not marked bulk parallel.
+    if (split_outer_dim) {
```
zasdfgbnm, line 896: Should we rename this flag to something like `permute_outer_dim`?

csrc/ir/interface_nodes.h
```diff
@@ -409,6 +409,20 @@ class NVF_API TensorView : public Val {
...
+  //! This doesn't work for the swizzle none mode. For more details
+  //! refer to the figure doc/dev/tma/swizzle.svg
+  void swizzleTMABox(MmaInputSmemSwizzle swizzle);
+
+  //! Transforms the innermost iterdomains according to the given mma swizzle,
+  //!  this should be used on the tvs that are inputs of a MmaOp or are loaded
+  //!  using TMA.
+  void applyMmaSwizzleForTMALoad(
+      MmaInputSmemSwizzle swizzle,
+      bool split_outer_dim = true);
```
zasdfgbnm, line 424:
```suggestion
      bool permute_outer_dim = true);
```

csrc/scheduler/mma_utils.h
```diff
@@ -195,6 +195,36 @@ class WarpMmaSwizzler {
   //! output and mma inputs are scheduled in scheduleOperandRead, which must be
   //! called before this function.
   static void scheduleLdMatrix(TensorView* tv, MmaOperand operand);
+
+  //! Function to schedule the swizzled TMA box.
+  //! This functions works on the assumption that the TMA box is 2D
+  //! and the inner-dimension is less or equal to the swizzle size.
+  //! This doesn't work for the swizzle none mode. For more details
+  //! refer to the figure doc/dev/tma/swizzle.svg
+  static void swizzleTMABox(
```
zasdfgbnm, line 204: Is this still used?

csrc/scheduler/mma_utils.h
```diff
@@ -195,6 +195,36 @@ class WarpMmaSwizzler {
...
+  //! above. Please note that we currently do not fully support not splitting
+  //! the outer dimension. This only works when the inner-dimension is not
+  //! split, that is the inner dim is less or equal to the swizzle size (in
+  //! bytes). The outer dim here refers to the second ID from the end, so for
+  //! the input [B, N, K], N would be outer. Broadcast is always moved
+  //! outermost.
+  static void scheduleTMALoadForMma(
+      TensorView* tv,
+      MmaInputSmemSwizzle swizzle,
+      bool split_outer_dim = true);
```
zasdfgbnm, line 221:
```suggestion
      bool permute_outer_dim = true);
```

csrc/tensor_view.cpp
```diff
@@ -1327,6 +1327,49 @@ void TensorView::applyMmaSwizzle(MmaInputSmemSwizzle swizzle) {
...
+  // [..., KO(2), KIO(2), KII(4), N(16)] ->
+  // [..., KO(2), KIO(2), KII(4), NIO(2), NII(8)]
+  split(-1, (core_matrix_width_bytes / dataTypeSize(dtype)));
+
+  this->swizzle(SwizzleType::XOR, -4, -2);
+}
+
+void TensorView::applyMmaSwizzleForTMALoad(
+    MmaInputSmemSwizzle swizzle,
+    bool split_outer_dim) {
```
zasdfgbnm, line 1361:
```suggestion
    bool permute_outer_dim) {
```

zasdfgbnm (approved):

csrc/scheduler/mma_utils.h
protonu: Thanks, missed pushing a local commit - removed.

protonu:
!build

protonu:
!build

protonu:
!build

protonu:
!build

PR #2510, opened 2024-07-02
patch isZero/isOne

jjsjann123:
1. There's no overload across equal operator for `std::complex<T>, T`
```
auto x = std::complex<float>(0.0, 0.0);
x == 0; // This one triggers an error.
```

So we hit this with our dynamic type thing in `Val::isZero()`. But it magically worked if I change it from an integer to a float. I have no idea why. cc'ing @zasdfgbnm 

2. `at::pad` does not support complex dtype. But nvfuser API does. Added a chain of operation to support it.

### DIFF
--- a/csrc/ir/base_nodes.cpp
+++ b/csrc/ir/base_nodes.cpp
@@ -218,7 +218,7 @@ PolymorphicValue Val::evaluate() {
 }
 
 bool Val::isZero() const {
-  return value().hasValue() && (bool)(value() == 0);
+  return value().hasValue() && (bool)(value() == 0.0);
 }
 
 bool Val::isZeroInt() const {
@@ -227,7 +227,8 @@ bool Val::isZeroInt() const {
 }
 
 bool Val::isOne() const {
-  return value().hasValue() && (bool)(value() == 1);
+  // NOTE: does this work for complex?!
+  return value().hasValue() && (bool)(value() == 1.0);
 }
 
 bool Val::isOneInt() const {
--- a/csrc/ir/base_nodes.cpp
+++ b/csrc/ir/base_nodes.cpp
@@ -227,7 +227,6 @@ bool Val::isZeroInt() const {
 }
 
 bool Val::isOne() const {
-  // NOTE: does this work for complex?!
   return value().hasValue() && (bool)(value() == 1.0);
 }
 
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -3972,7 +3972,8 @@ std::vector<PolymorphicValue> PadOp::evaluate(
   }
 
   if (isComplexType(*out()->getDataType())) {
-    std::complex<double> value = static_cast<std::complex<double>>(inputs.at(1));
+    std::complex<double> value =
+        static_cast<std::complex<double>>(inputs.at(1));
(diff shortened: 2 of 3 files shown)


csrc/ir/base_nodes.cpp
```diff
@@ -218,7 +218,7 @@ PolymorphicValue Val::evaluate() {
 }
 
 bool Val::isZero() const {
-  return value().hasValue() && (bool)(value() == 0);
+  return value().hasValue() && (bool)(value() == 0.0);
```
jjsjann123, line 221: I don't know why this would work but an integer doesn't. I'm more than happy to open an issue to track this one cc'ing @zasdfgbnm

jjsjann123:
!build

zasdfgbnm:
> But it magically worked if I change it from an integer to a float. I have no idea why.

Probably because `operator==` for `(complex, int)` is not defined?

jjsjann123:
> > But it magically worked if I change it from an integer to a float. I have no idea why.
> 
> Probably because `operator==` for `(complex, int)` is not defined?

`operator==` for `(std::complex<int>, int)` is defined. So is `(std::complex<T>, T)`, but not cross product... So I don't know why a float there would work but not an integer.

zasdfgbnm:
> > > But it magically worked if I change it from an integer to a float. I have no idea why.
> > 
> > 
> > Probably because `operator==` for `(complex, int)` is not defined?
> 
> `operator==` for `(std::complex<int>, int)` is defined. So is `(std::complex<T>, T)`, but not cross product... So I don't know why a float there would work but not an integer.

I don't think `float` works. What is working is `double`:

```C++
#include <complex>

std::complex<double> x = 0.0;
auto y = x + 0.0;
auto z = x + 0.0f; // wrong
auto q = x + 0; // wrong
```

csrc/ir/nodes.cpp
```diff
@@ -3972,7 +3971,17 @@ std::vector<PolymorphicValue> PadOp::evaluate(
     pad_widths.push_back(right_pad);
   }
 
-  return {at::pad(in, pad_widths, "constant", value)};
+  if (isComplexType(*out()->getDataType())) {
+    std::complex<double> value = static_cast<std::complex<double>>(inputs.at(1));
+    auto real = at::real(in);
+    auto imag = at::imag(in);
+    auto padded_real = at::pad(real, pad_widths, "constant", value.real());
```
zasdfgbnm, line 3978: How can `at::pad` does not support complex....

zasdfgbnm (approved):

jjsjann123:
> I don't think `float` works. What is working is `double`:

You got me there.

I guess I was confused to why `double` would work for complexFloat. But I realized that we probably only use the higher precision for scalar? i.e. we'll only have `std::complex<double>` but not `std::complex<float>`.

csrc/ir/nodes.cpp
jjsjann123: :shrug:

jjsjann123:
!build

jjsjann123:
!build

jjsjann123:
!build

zasdfgbnm:
> we'll only have `std::complex<double>` but not `std::complex<float>`.

Yes

PR #2512, opened 2024-07-02
[WIP] Apply AbstractTensor schedule to TensorView

jacobhinkle:
While implementing #2458 I wound up reimplementing some of the AbstractTensor logic in order to schedule operands and epilogue in a unified way. Instead, I thought I'd take a crack at implementing a more general approach.

This PR implements `applyAbstractSchedule(const AbstractTensor&, TensorView*)` so that we can create an `AbstractTensor`, schedule it (at least transforms if not parallelization), then apply it without needing a concrete reference tensor.

### DIFF
--- a/csrc/abstract_tensor_schedule.cpp
+++ b/csrc/abstract_tensor_schedule.cpp
@@ -16,19 +16,45 @@ namespace nvfuser {
 
 namespace {
 
+std::string toString(ValGroup g) {
+  std::stringstream ss;
+  ss << "ValGroup[" << (void*)g.get() << "]:";
+  for (Val* v : g->vector()) {
+    ss << " " << v->name();
+  }
+  return ss.str();
+}
+
 class AbstractTensorSchedule : public IterVisitor {
  public:
-  static void apply(const AbstractTensor& abstract, TensorView* concrete) {
-    AbstractTensorSchedule ats(abstract, concrete);
+  static void apply(
+      const AbstractTensor& abstract,
+      TensorView* concrete,
+      ValGraph* graph) {
+    AbstractTensorSchedule ats(abstract, concrete, graph);
     ats.run();
   }
 
  private:
-  AbstractTensorSchedule(const AbstractTensor& abstract, TensorView* concrete)
-      : abstract_(abstract), concrete_(concrete) {}
+  AbstractTensorSchedule(const AbstractTensor& abstract, TensorView* concrete, ValGraph* graph)
+      : abstract_(abstract), concrete_(concrete), graph_(graph) {}
 
   void run() {
+    concrete_->fusion()->printMath(0);
+    concrete_->fusion()->printTransforms();
+    std::cout << "abstract_ = " << std::endl;
+    for (const AbstractId& abs_id : abstract_.domain) {
+      ValGroup g = abstractIdToValGroup(abs_id);
+      std::cout << "  " << toString(g) << std::endl;
+    }
+    std::cout << "concrete_ = " << concrete_->toString() << std::endl;
+    std::cout << "concrete_ids_ after findNearestProducers:" << std::endl;
+    for (auto& [vg, id] : concrete_ids_) {
+      std::cout << "  " << toString(vg) << "  ->  " << id->toString() << std::endl;
+    }
+
     findNearestProducers();
+
 
     // Now, for each ValGroup in abstract, find the closest producer ValGroups
     // with entries in concrete_ids_ and replay the path from them. If none
@@ -47,6 +73,16 @@ class AbstractTensorSchedule : public IterVisitor {
     concrete_->setLoopDomain(loop_domain);
   }
 
+  ValGroup abstractIdToValGroup(const AbstractId& abs_id) {
+    if (abs_id.is<IterDomain*>()) {
+      return graph_->toGroup(abs_id.as<IterDomain*>());
+    } else if (abs_id.is<ValGroupAndItsGraph>()) {
+      NVF_ERROR(graph_ == abs_id.as<ValGroupAndItsGraph>().graph);
+      return abs_id.as<ValGroupAndItsGraph>().group;
+    }
+    NVF_ERROR(false, "AbstractId must be IterDomain* or ValGroupAndItsGraph");
+  }
+
   //! Work backward from each loop IterDomain in concrete. When we find an
   //! IterDomain* with a ValGroup in scheduled_val_groups, we map to it
   //! (including siblings) in concrete_ids_. Assert that we do find a
@@ -58,29 +94,19 @@ class AbstractTensorSchedule : public IterVisitor {
     // TODO: this should probably be a more general utility like DependencyCheck
     // in val_graph_visitor.h
     std::unordered_set<ValGroup> scheduled_val_groups;
-    ValGraph* graph = nullptr;
     std::stack<ValGroup> vg_stack;
     for (const AbstractId& abs_id : abstract_.domain) {
-      NVF_ERROR(
-          abs_id.is<ValGroupAndItsGraph>(),
-          "abstract tensor must contain only ValGroups");
-      const ValGroupAndItsGraph& gg = abs_id.as<ValGroupAndItsGraph>();
-      if (graph == nullptr) {
-        graph = gg.graph;
-      } else {
-        NVF_ERROR(graph == gg.graph);
-      }
-      vg_stack.push(gg.group);
+      vg_stack.push(abstractIdToValGroup(abs_id));
     }
     while (!vg_stack.empty()) {
       ValGroup vg = vg_stack.top();
       vg_stack.pop();
       scheduled_val_groups.insert(vg);
       // TODO: Do we need to check for cycles?
-      for (const ExprGroup& eg : graph->getDefinitions(vg)) {
+      for (const ExprGroup& eg : graph_->getDefinitions(vg)) {
         for (Expr* e : *eg) {
           for (Val* inp : e->inputs()) {
-            ValGroup vg_inp = graph->toGroup(inp);
+            ValGroup vg_inp = graph_->toGroup(inp);
             vg_stack.push(vg_inp);
           }
         }
@@ -96,7 +122,7 @@ class AbstractTensorSchedule : public IterVisitor {
     while (!id_stack.empty()) {
       IterDomain* id = id_stack.top();
       id_stack.pop();
-      ValGroup g = graph->toGroup(id);
+      ValGroup g = graph_->toGroup(id);
       if (scheduled_val_groups.find(g) != scheduled_val_groups.end()) {
         concrete_ids_.emplace(g, id);
         continue;
@@ -115,11 +141,7 @@ class AbstractTensorSchedule : public IterVisitor {
   }
 
   IterDomain* replayAbstractId(AbstractId abs_id) {
-    NVF_ERROR(
-        abs_id.is<ValGroupAndItsGraph>(),
-        "abstract must contain only ValGroups");
-    ValGroup g = abs_id.as<ValGroupAndItsGraph>().group;
-    ValGraph* graph = abs_id.as<ValGroupAndItsGraph>().graph;
+    ValGroup g = abstractIdToValGroup(abs_id);
 
     // This holds ValGroups that we cannot compute from concrete_'s root. When
     // we detect that any inputs of an ExprGroup are in these groups, we know
@@ -129,18 +151,24 @@ class AbstractTensorSchedule : public IterVisitor {
     std::stack<ValGroup> vg_stack({g});
     // Strategy: fill out concrete_ids_ in the direction of g
     auto propagate = [&](ValGroup vg) {
+      std::cout << "propagate " << toString(vg) << std::endl;
       // We need to try and produce an ID for this group, so we look at
       // definitions for the Vals in this group and check whether we have their
       // inputs computed yet (concrete_ids_) and whether they can be computed
       // (uncomputable_groups).
       std::vector<ValGroup> uncomputed_producer_groups;
-      for (const ExprGroup& eg : graph->getDefinitions(vg)) {
+      for (const ExprGroup& eg : graph_->getDefinitions(vg)) {
         for (Expr* e : *eg) {
+          if (!uncomputed_producer_groups.empty()) {
+            // We already have something to compute in the next round, so skip.
+            continue;
+          }
+          std::cout << "  definition " << e->toString() << std::endl;
           std::vector<ValGroup> vg_inps;
           std::vector<IterDomain*> id_inps;
           bool all_inputs_computed = true;
           for (Val* inp : e->inputs()) {
-            ValGroup vg_inp = graph->toGroup(inp);
+            ValGroup vg_inp = graph_->toGroup(inp);
             vg_inps.push_back(vg_inp);
             auto inp_it = concrete_ids_.find(vg_inp);
             if (inp_it != concrete_ids_.end()) {
@@ -156,6 +184,7 @@ class AbstractTensorSchedule : public IterVisitor {
           }
           // some input is not yet computed
           if (!all_inputs_computed) {
+            std::cout << "  some inputs not computed" << std::endl;
             continue;
           }
           // Compute new ID expression
@@ -187,7 +216,7 @@ class AbstractTensorSchedule : public IterVisitor {
           NVF_ERROR(id_expr != nullptr);
           NVF_ERROR(id_expr->outputs().size() == e->outputs().size());
           for (size_t i : c10::irange(e->outputs().size())) {
-            ValGroup vg_outp = graph->toGroup(e->output((int64_t)i));
+            ValGroup vg_outp = graph_->toGroup(e->output((int64_t)i));
             auto* id_outp = id_expr->output((int64_t)i)->as<IterDomain>();
             concrete_ids_.emplace(vg_outp, id_outp);
           }
@@ -196,13 +225,16 @@ class AbstractTensorSchedule : public IterVisitor {
       if (uncomputed_producer_groups.empty()) {
         // All of the defining expressions are proven uncomputable, so mark vg
         // uncomputable
+        std::cout << "  Marking " << toString(vg) << " uncomputable" << std::endl;
         uncomputable_groups.insert(vg);
         return;
       } else {
         // There are some uncomputed producer groups that might be computable,
         // so try again after processing those producer groups
+        std::cout << "  Adding " << toString(vg) << " back to stack beneath producer groups:" << std::endl;
         vg_stack.push(vg);
         for (ValGroup next_vg : uncomputed_producer_groups) {
+          std::cout << "    " << toString(next_vg) << std::endl;
           vg_stack.push(next_vg);
         }
       }
@@ -211,7 +243,9 @@ class AbstractTensorSchedule : public IterVisitor {
     while (!vg_stack.empty()) {
       ValGroup vg = vg_stack.top();
       vg_stack.pop();
+      std::cout << "popped from stack: " << toString(vg) << std::endl;
       if (concrete_ids_.count(vg) != 0) {
+        std::cout << "  concretized already" << std::endl;
         continue;
       }
       propagate(vg);
@@ -227,6 +261,7 @@ class AbstractTensorSchedule : public IterVisitor {
  private:
   const AbstractTensor& abstract_;
   TensorView* concrete_;
+  ValGraph* graph_;
 
   std::unordered_map<ValGroup, IterDomain*> concrete_ids_;
 };
@@ -235,8 +270,9 @@ class AbstractTensorSchedule : public IterVisitor {
 
 void applyAbstractSchedule(
     const AbstractTensor& abstract,
-    TensorView* concrete) {
-  AbstractTensorSchedule::apply(abstract, concrete);
+    TensorView* concrete,
+    ValGraph* graph) {
+  AbstractTensorSchedule::apply(abstract, concrete, graph);
 }
 
 } // namespace nvfuser
--- a/csrc/abstract_tensor_schedule.cpp
+++ b/csrc/abstract_tensor_schedule.cpp
@@ -37,8 +37,10 @@ class AbstractTensorSchedule : public IterVisitor {
     for (const AbstractId& abs_id : abstract_.domain) {
       IterDomain* new_id = replayAbstractId(abs_id);
       if (new_id == nullptr) {
+        std::cout << "new_id = nullptr" << std::endl;
         continue;
       }
+      std::cout << "new_id = " << new_id->toString() << std::endl;
       loop_domain.push_back(new_id);
     }
 
@@ -117,7 +119,109 @@ class AbstractTensorSchedule : public IterVisitor {
         abs_id.is<ValGroupAndItsGraph>(),
         "abstract must contain only ValGroups");
     ValGroup g = abs_id.as<ValGroupAndItsGraph>().group;
-    return nullptr;
+    ValGraph* graph = abs_id.as<ValGroupAndItsGraph>().graph;
+
+    // This holds ValGroups that we cannot compute from concrete_'s root. When
+    // we detect that any inputs of an ExprGroup are in these groups, we know
+    // that we cannot compute that ExprGroup so we should skip it.
+    std::unordered_set<ValGroup> uncomputable_groups;
+
+    std::stack<ValGroup> vg_stack({g});
+    // Strategy: fill out concrete_ids_ in the direction of g
+    auto propagate = [&](ValGroup vg) {
+      // We need to try and produce an ID for this group, so we look at
+      // definitions for the Vals in this group and check whether we have their
+      // inputs computed yet (concrete_ids_) and whether they can be computed
+      // (uncomputable_groups).
+      std::vector<ValGroup> uncomputed_producer_groups;
+      for (const ExprGroup& eg : graph->getDefinitions(vg)) {
+        for (Expr* e : *eg) {
+          std::vector<ValGroup> vg_inps;
+          std::vector<IterDomain*> id_inps;
+          bool all_inputs_computed = true;
+          for (Val* inp : e->inputs()) {
+            ValGroup vg_inp = graph->toGroup(inp);
+            vg_inps.push_back(vg_inp);
+            auto inp_it = concrete_ids_.find(vg_inp);
+            if (inp_it != concrete_ids_.end()) {
+              id_inps.push_back(inp_it->second);
+            } else {
+              all_inputs_computed = false;
+              if (uncomputable_groups.count(vg_inp) == 0) {
+                // Input is not yet proven to be incomputable, so try and
+                // compute it in the next iteration
+                uncomputed_producer_groups.push_back(vg_inp);
+              }
+            }
+          }
+          // some input is not yet computed
+          if (!all_inputs_computed) {
+            continue;
+          }
+          // Compute new ID expression
+          Expr* id_expr = nullptr;
+          if (auto* m = dynamic_cast<Merge*>(e)) {
+            NVF_ERROR(id_inps.size() == 2);
+            auto* new_id = IterDomain::merge(id_inps[0], id_inps[1]);
+            id_expr = new_id->definition();
+          } else if (auto* s = dynamic_cast<Split*>(e)) {
+            NVF_ERROR(id_inps.size() == 1);
+            auto new_ids =
+                IterDomain::split(id_inps[0], s->factor(), s->innerSplit());
+            id_expr = new_ids.first->definition();
+          } else if (auto* s = dynamic_cast<Swizzle*>(e)) {
+            NVF_ERROR(id_inps.size() == 2);
+            auto new_ids =
+                IterDomain::swizzle(s->swizzleType(), id_inps[0], id_inps[1]);
+            id_expr = new_ids.first->definition();
+          } else if (auto* s = dynamic_cast<Swizzle2D*>(e)) {
+            NVF_ERROR(id_inps.size() == 2);
+            auto new_ids = IterDomain::swizzle(
+                s->swizzleType(), id_inps[0], id_inps[1], s->swizzleMode());
+            id_expr = new_ids.first->definition();
+          } else {
+            NVF_ERROR(false, "Unhandled IterDomain expression ", e->toString());
+          }
+          // Update the mapping to point to all of the newly created Expr's
+          // outputs
+          NVF_ERROR(id_expr != nullptr);
+          NVF_ERROR(id_expr->outputs().size() == e->outputs().size());
+          for (size_t i : c10::irange(e->outputs().size())) {
+            ValGroup vg_outp = graph->toGroup(e->output((int64_t)i));
+            auto* id_outp = id_expr->output((int64_t)i)->as<IterDomain>();
+            concrete_ids_.emplace(vg_outp, id_outp);
+          }
+        }
+      }
+      if (uncomputed_producer_groups.empty()) {
+        // All of the defining expressions are proven uncomputable, so mark vg
+        // uncomputable
+        uncomputable_groups.insert(vg);
+        return;
+      } else {
+        // There are some uncomputed producer groups that might be computable,
+        // so try again after processing those producer groups
+        vg_stack.push(vg);
+        for (ValGroup next_vg : uncomputed_producer_groups) {
+          vg_stack.push(next_vg);
+        }
+      }
+    };
+
+    while (!vg_stack.empty()) {
+      ValGroup vg = vg_stack.top();
+      vg_stack.pop();
+      if (concrete_ids_.count(vg) != 0) {
+        continue;
+      }
+      propagate(vg);
+    }
+
+    // Now check whether there is a concrete entry for abs_id (i.e. g). If not,
+    // that means this dimension is missing and should not be included in the
+    // loop domain, so return nullptr.
+    auto it = concrete_ids_.find(g);
+    return it == concrete_ids_.end() ? nullptr : it->second;
   }
 
  private:
@@ -133,101 +237,6 @@ void applyAbstractSchedule(
     const AbstractTensor& abstract,
     TensorView* concrete) {
   AbstractTensorSchedule::apply(abstract, concrete);
-  /*
-  if (concrete->nDims() == 0) {
-    return;
-  }
-
-  VectorOfUniqueEntries<ValGroup> concrete_loop, abstract_loop;
-
-  ValGraph* graph = nullptr;
-  for (const AbstractId& abs_id : abstract.domain) {
-    NVF_ERROR(abs_id.is<ValGroupAndItsGraph>());
-    ValGroupAndItsGraph gg(abs_id.as<ValGroupAndItsGraph>());
-    abstract_loop.pushBack(gg.group);
-    if (graph == nullptr) {
-      graph = gg.graph;
-    } else {
-      NVF_ERROR(graph == gg.graph);
-    }
-  }
-
-  // Mapping from ValGroups to IterDomains corresponding to concrete
-  std::unordered_map<ValGroup, IterDomain*> concrete_iter_map;
-
-  NVF_ERROR(abstract.size() > 0);
-  for (IterDomain* id : concrete->getLoopDomain()) {
-    ValGroup g = graph->toGroup(id);
-    concrete_loop.pushBack(g);
-    concrete_iter_map.emplace(g, id);
-  }
-
-  for (auto [eg, dir] :
-       ValGraphBFS::getExprsBetween(graph, concrete_loop, abstract_loop)) {
-    NVF_ERROR(
-        dir == Direction::Forward,
-        "Only forward paths are allowed between concrete and abstract");
-    NVF_ERROR(!eg->empty());
-    if (auto s = dynamic_cast<Split*>(eg->front())) {
-      ValGroup g_in = graph->toGroup(s->in());
-      ValGroup g_outer = graph->toGroup(s->outer());
-      ValGroup g_inner = graph->toGroup(s->inner());
-
-      auto it = concrete_iter_map.find(g_in);
-      if (it == concrete_iter_map.end()) {
-        continue;
-      }
-      IterDomain* id_in = it->second;
-
-      auto [id_outer, id_inner] = IterDomain::split(id_in, s->factor(),
-  s->innerSplit());
-
-      concrete_iter_map.emplace(g_outer, id_outer);
-      concrete_iter_map.emplace(g_inner, id_inner);
-    } else if (auto m = dynamic_cast<Merge*>(eg->front())) {
-      ValGroup g_outer = graph->toGroup(m->outer());
-      ValGroup g_inner = graph->toGroup(m->inner());
-      ValGroup g_out = graph->toGroup(m->out());
-
-      auto it = concrete_iter_map.find(g_outer);
-      if (it == concrete_iter_map.end()) {
-        continue;
-      }
-      IterDomain* id_outer = it->second;
-      it = concrete_iter_map.find(g_inner);
-      if (it == concrete_iter_map.end()) {
-        continue;
-      }
-      IterDomain* id_inner = it->second;
-
-      IterDomain* id_out = IterDomain::merge(id_outer, id_inner);
-
-      concrete_iter_map.emplace(g_out, id_out);
-    } else if (auto s = dynamic_cast<Swizzle*>(eg->front())) {
-      NVF_ERROR(
-          false,
-          "Expr ",
-          s->toString(),
-          " not yet handled in applyAbstractSchedule");
-    } else if (auto s = dynamic_cast<Swizzle2D*>(eg->front())) {
-      NVF_ERROR(
-          false,
-          "Expr ",
-          s->toString(),
-          " not yet handled in applyAbstractSchedule");
-    }
-  }
-
-  // Now create loop domain using concrete_iter_map
-  std::vector<IterDomain*> new_loop;
-  for (AbstractId abs_id : abstract.domain) {
-    auto it = concrete_iter_map.find(abs_id.as<ValGroupAndItsGraph>().group);
-    if (it != concrete_iter_map.end()) {
-      new_loop.push_back(it->second);
-    }
-  }
-  concrete->setLoopDomain(new_loop);
-  */
 }
 
 } // namespace nvfuser
--- a/csrc/abstract_tensor_schedule.cpp
+++ b/csrc/abstract_tensor_schedule.cpp
@@ -77,7 +77,11 @@ class AbstractTensorSchedule : public IterVisitor {
     if (abs_id.is<IterDomain*>()) {
       return graph_->toGroup(abs_id.as<IterDomain*>());
     } else if (abs_id.is<ValGroupAndItsGraph>()) {
-      NVF_ERROR(graph_ == abs_id.as<ValGroupAndItsGraph>().graph);
+      if (graph_ == nullptr) {
+        graph_ = abs_id.as<ValGroupAndItsGraph>().graph;
+      } else {
+        NVF_ERROR(graph_ == abs_id.as<ValGroupAndItsGraph>().graph);
+      }
       return abs_id.as<ValGroupAndItsGraph>().group;
     }
     NVF_ERROR(false, "AbstractId must be IterDomain* or ValGroupAndItsGraph");
--- a/csrc/abstract_tensor_schedule.cpp
+++ b/csrc/abstract_tensor_schedule.cpp
@@ -1,0 +1,233 @@
+// clang-format off
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2024-present NVIDIA CORPORATION & AFFILIATES.
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+// clang-format on
+
+#include <abstract_tensor.h>
+#include <abstract_tensor_schedule.h>
+#include <ir/internal_base_nodes.h>
+#include <iter_visitor.h>
+#include <val_graph.h>
+
+namespace nvfuser {
+
+namespace {
+
+class AbstractTensorSchedule : public IterVisitor {
+ public:
+  static void apply(const AbstractTensor& abstract, TensorView* concrete) {
+    AbstractTensorSchedule ats(abstract, concrete);
+    ats.run();
+  }
+
+ private:
+  AbstractTensorSchedule(const AbstractTensor& abstract, TensorView* concrete)
+      : abstract_(abstract), concrete_(concrete) {}
+
+  void run() {
+    findNearestProducers();
+
+    // Now, for each ValGroup in abstract, find the closest producer ValGroups
+    // with entries in concrete_ids_ and replay the path from them. If none
+    // exists, then do not include this dimension in the output
+    std::vector<IterDomain*> loop_domain;
+    for (const AbstractId& abs_id : abstract_.domain) {
+      IterDomain* new_id = replayAbstractId(abs_id);
+      if (new_id == nullptr) {
+        continue;
+      }
+      loop_domain.push_back(new_id);
+    }
+
+    concrete_->setLoopDomain(loop_domain);
+  }
+
+  //! Work backward from each loop IterDomain in concrete. When we find an
+  //! IterDomain* with a ValGroup in scheduled_val_groups, we map to it
+  //! (including siblings) in concrete_ids_. Assert that we do find a
+  //! scheduled ValGroup, otherwise we would not know where to place this
+  //! IterDomain.
+  void findNearestProducers() {
+    // Record all ValGroups that are producers of those in abstract_. This lets
+    // us find starting IterDomains for scheduling concrete's loop domain.
+    // TODO: this should probably be a more general utility like DependencyCheck
+    // in val_graph_visitor.h
+    std::unordered_set<ValGroup> scheduled_val_groups;
+    ValGraph* graph = nullptr;
+    std::stack<ValGroup> vg_stack;
+    for (const AbstractId& abs_id : abstract_.domain) {
+      NVF_ERROR(
+          abs_id.is<ValGroupAndItsGraph>(),
+          "abstract tensor must contain only ValGroups");
+      const ValGroupAndItsGraph& gg = abs_id.as<ValGroupAndItsGraph>();
+      if (graph == nullptr) {
+        graph = gg.graph;
+      } else {
+        NVF_ERROR(graph == gg.graph);
+      }
+      vg_stack.push(gg.group);
+    }
+    while (!vg_stack.empty()) {
+      ValGroup vg = vg_stack.top();
+      vg_stack.pop();
+      scheduled_val_groups.insert(vg);
+      // TODO: Do we need to check for cycles?
+      for (const ExprGroup& eg : graph->getDefinitions(vg)) {
+        for (Expr* e : *eg) {
+          for (Val* inp : e->inputs()) {
+            ValGroup vg_inp = graph->toGroup(inp);
+            vg_stack.push(vg_inp);
+          }
+        }
+      }
+    }
+
+    // Now traverse c2p from concrete loop domain, stopping when we find a
+    // scheduled ValGroup
+    std::stack<IterDomain*> id_stack;
+    for (IterDomain* id : concrete_->getLoopDomain()) {
+      id_stack.push(id);
+    }
+    while (!id_stack.empty()) {
+      IterDomain* id = id_stack.top();
+      id_stack.pop();
+      ValGroup g = graph->toGroup(id);
+      if (scheduled_val_groups.find(g) != scheduled_val_groups.end()) {
+        concrete_ids_.emplace(g, id);
+        continue;
+      }
+      NVF_ERROR(
+          id->definition() != nullptr,
+          "Root IterDomain ",
+          id->toString(),
+          " does not appear in the history of any ValGroups in abstract tensor");
+      for (Val* inp : id->definition()->inputs()) {
+        if (auto inp_id = dynamic_cast<IterDomain*>(inp)) {
+          id_stack.push(inp_id);
+        }
+      }
+    }
+  }
+
+  IterDomain* replayAbstractId(AbstractId abs_id) {
+    NVF_ERROR(
+        abs_id.is<ValGroupAndItsGraph>(),
+        "abstract must contain only ValGroups");
+    ValGroup g = abs_id.as<ValGroupAndItsGraph>().group;
+    return nullptr;
+  }
+
+ private:
+  const AbstractTensor& abstract_;
+  TensorView* concrete_;
+
+  std::unordered_map<ValGroup, IterDomain*> concrete_ids_;
+};
+
+} // namespace
+
+void applyAbstractSchedule(
+    const AbstractTensor& abstract,
+    TensorView* concrete) {
+  AbstractTensorSchedule::apply(abstract, concrete);
+  /*
+  if (concrete->nDims() == 0) {
+    return;
+  }
+
+  VectorOfUniqueEntries<ValGroup> concrete_loop, abstract_loop;
+
+  ValGraph* graph = nullptr;
+  for (const AbstractId& abs_id : abstract.domain) {
+    NVF_ERROR(abs_id.is<ValGroupAndItsGraph>());
+    ValGroupAndItsGraph gg(abs_id.as<ValGroupAndItsGraph>());
+    abstract_loop.pushBack(gg.group);
+    if (graph == nullptr) {
+      graph = gg.graph;
+    } else {
+      NVF_ERROR(graph == gg.graph);
+    }
+  }
+
+  // Mapping from ValGroups to IterDomains corresponding to concrete
+  std::unordered_map<ValGroup, IterDomain*> concrete_iter_map;
+
+  NVF_ERROR(abstract.size() > 0);
+  for (IterDomain* id : concrete->getLoopDomain()) {
+    ValGroup g = graph->toGroup(id);
+    concrete_loop.pushBack(g);
+    concrete_iter_map.emplace(g, id);
+  }
+
+  for (auto [eg, dir] :
+       ValGraphBFS::getExprsBetween(graph, concrete_loop, abstract_loop)) {
+    NVF_ERROR(
+        dir == Direction::Forward,
+        "Only forward paths are allowed between concrete and abstract");
+    NVF_ERROR(!eg->empty());
+    if (auto s = dynamic_cast<Split*>(eg->front())) {
+      ValGroup g_in = graph->toGroup(s->in());
+      ValGroup g_outer = graph->toGroup(s->outer());
+      ValGroup g_inner = graph->toGroup(s->inner());
+
+      auto it = concrete_iter_map.find(g_in);
+      if (it == concrete_iter_map.end()) {
+        continue;
+      }
+      IterDomain* id_in = it->second;
+
+      auto [id_outer, id_inner] = IterDomain::split(id_in, s->factor(),
+  s->innerSplit());
+
+      concrete_iter_map.emplace(g_outer, id_outer);
+      concrete_iter_map.emplace(g_inner, id_inner);
+    } else if (auto m = dynamic_cast<Merge*>(eg->front())) {
+      ValGroup g_outer = graph->toGroup(m->outer());
+      ValGroup g_inner = graph->toGroup(m->inner());
+      ValGroup g_out = graph->toGroup(m->out());
+
+      auto it = concrete_iter_map.find(g_outer);
+      if (it == concrete_iter_map.end()) {
+        continue;
+      }
+      IterDomain* id_outer = it->second;
+      it = concrete_iter_map.find(g_inner);
+      if (it == concrete_iter_map.end()) {
+        continue;
+      }
+      IterDomain* id_inner = it->second;
+
+      IterDomain* id_out = IterDomain::merge(id_outer, id_inner);
+
+      concrete_iter_map.emplace(g_out, id_out);
+    } else if (auto s = dynamic_cast<Swizzle*>(eg->front())) {
+      NVF_ERROR(
+          false,
+          "Expr ",
+          s->toString(),
+          " not yet handled in applyAbstractSchedule");
+    } else if (auto s = dynamic_cast<Swizzle2D*>(eg->front())) {
+      NVF_ERROR(
+          false,
+          "Expr ",
+          s->toString(),
+          " not yet handled in applyAbstractSchedule");
+    }
+  }
+
+  // Now create loop domain using concrete_iter_map
+  std::vector<IterDomain*> new_loop;
+  for (AbstractId abs_id : abstract.domain) {
+    auto it = concrete_iter_map.find(abs_id.as<ValGroupAndItsGraph>().group);
+    if (it != concrete_iter_map.end()) {
+      new_loop.push_back(it->second);
+    }
+  }
+  concrete->setLoopDomain(new_loop);
+  */
+}
+
+} // namespace nvfuser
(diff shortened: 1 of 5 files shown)


csrc/abstract_tensor_schedule.cpp
```diff
@@ -0,0 +1,261 @@
...
+            NVF_ERROR(false, "Unhandled IterDomain expression ", e->toString());
+          }
+          // Update the mapping to point to all of the newly created Expr's
+          // outputs
+          NVF_ERROR(id_expr != nullptr);
+          NVF_ERROR(id_expr->outputs().size() == e->outputs().size());
+          for (size_t i : c10::irange(e->outputs().size())) {
+            ValGroup vg_outp = graph_->toGroup(e->output((int64_t)i));
+            auto* id_outp = id_expr->output((int64_t)i)->as<IterDomain>();
+            concrete_ids_.emplace(vg_outp, id_outp);
```
jacobhinkle, line 209: I think we probably also need to append `id_outp` to `vg_outp`.

tests/cpp/test_abstract_tensor.cpp
```diff
@@ -712,4 +717,62 @@ TEST_F(AbstractTensorTest, UnzipBroadcasting) {
...
+  addAbstractAxis(tv1->axis(1));
+
+  abten.merge(0);
+  abten.merge(0);
+  abten.split(0, 128);
+
+  for (TensorView* tv : {tv3, tv4, tv5, tv6}) {
+    applyAbstractSchedule(abten, tv, &graph);
+    tv->axis(-1)->parallelize(ParallelType::TIDx);
+    tv->axis(-2)->parallelize(ParallelType::BIDx);
```
jacobhinkle, line 760: It might be nice to have a version of `scheduler_utils::parallelizeAllLike` that takes an `AbstractTensor` and vector of `ParallelType`s. That way, we wouldn't need to figure out which axes exist in `tv` here, which in general might mean replicating some of the logic in `applyAbstractSchedule`. Even better would be to be able to parallelize each axis as part of the `AbstractTensor` interface e.g.
```c++
abten.parallelize(-1, ParallelType::Vectorize);
// ...
parallelizeAllLike(/*reference=*/abten);
```

csrc/abstract_tensor_schedule.h
```diff
@@ -0,0 +1,23 @@
...
+#include <val_graph.h>
+
+namespace nvfuser {
+
+//! Apply the transformations found in an AbstractTensor to a concrete
+//! TensorView.
+void applyAbstractSchedule(
+    const AbstractTensor& abstract,
+    TensorView* concrete,
+    ValGraph* graph = nullptr);
```
jacobhinkle, line 21: Maybe `applyAbstractTransforms` would be a better name?

csrc/abstract_tensor_schedule.cpp
```diff
@@ -0,0 +1,261 @@
...
+        continue;
+      }
+      loop_domain.push_back(new_id);
+    }
+    concrete_->setLoopDomain(loop_domain);
+  }
+
+  ValGroup abstractIdToValGroup(const AbstractId& abs_id) {
+    if (abs_id.is<IterDomain*>()) {
+      return graph_->toGroup(abs_id.as<IterDomain*>());
```
jacobhinkle, line 55: I originally thought maybe we could make this work for `IterDomain` abstract IDs as well as `ValGroupAndItsGraph`s. However, it's confusing to use this way so I might assert here that `abs_id.is<ValGroupAndItsGraph>()` instead. To see why, consider how you might use this functionality:
```c++
  AbstractTensor abten({tv2->axis(0), tv1->axis(0), tv1->axis(1)});
  abten.merge(0);
  abten.merge(0);
  abten.split(0, 128);
  IdModel id_model(&fusion);
  ValGraph& graph = id_model.idGraph(IdMappingMode::PERMISSIVE);
  applyAbstractSchedule(abten, tv, &graph);
```
The reason we need to build the graph _after_ scheduling `abten` is that the split and merge ops on `AbstractTensor` introduce new `IterDomain`s and `Merge`/`Split` expressions. However, those are actually disconnected from the `Fusion` and they are ignored when building the `ValGraph`. By contrast, if we start with `ValGroup`s in `abten`, then the `AbstractTensor::merge` and `AbstractTensor::split` methods update the graph so that those disconnected IDs and exprs are added to the appropriate (pre-existing) `ValGroup`s. I don't see a good way around this, so for now it is probably best to only support the case where all the `AbstractId`s in `abten` are `ValGroupAndItsGraph`s.

jacobhinkle: Another option would be to fix up `IdModel` to consider "unused" `IterDomain`s and their expressions when building graphs. Currently we use `fusion->exprs()` and look at the `IterDomain`s in their TVs only, but this would let us grab the "unused" `IterDomain` resulting from `AbstractTensor::merge`, for example, as a `ValGroup` when needed. cc @naoyam.

naoyam: If it's clearly defined which domains and exprs should be considered, then I suppose it would be relatively simple. We shouldn't include truly dead domains and exprs, though.

csrc/abstract_tensor_schedule.cpp
```diff
@@ -0,0 +1,261 @@
...
+    std::stack<ValGroup> vg_stack({g});
+    // Strategy: fill out concrete_ids_ in the direction of g
+    auto propagate = [&](ValGroup vg) {
+      // We need to try and produce an ID for this group, so we look at
+      // definitions for the Vals in this group and check whether we have their
+      // inputs computed yet (concrete_ids_) and whether they can be computed
+      // (uncomputable_groups).
+      std::vector<ValGroup> uncomputed_producer_groups;
+      for (const ExprGroup& eg : graph_->getDefinitions(vg)) {
+        for (Expr* e : *eg) {
```
zasdfgbnm, line 144: Why do we need this `for (Expr* e : *eg)`, instead of just use `graph_->inputGroups`?

csrc/abstract_tensor_schedule.cpp
```diff
@@ -0,0 +1,261 @@
...
+    }
+    NVF_ERROR(false, "AbstractId must be IterDomain* or ValGroupAndItsGraph");
+  }
+
+  //! Work backward from each loop IterDomain in concrete. When we find an
+  //! IterDomain* with a ValGroup in scheduled_val_groups, we map to it
+  //! (including siblings) in concrete_ids_. Assert that we do find a
+  //! scheduled ValGroup, otherwise we would not know where to place this
+  //! IterDomain.
+  void findNearestProducers() {
```
zasdfgbnm, line 72: Could you provide an example in the comment?

csrc/abstract_tensor_schedule.cpp
```diff
@@ -0,0 +1,261 @@
...
+      }
+    }
+  }
+
+  IterDomain* replayAbstractId(AbstractId abs_id) {
+    ValGroup g = abstractIdToValGroup(abs_id);
+
+    // This holds ValGroups that we cannot compute from concrete_'s root. When
+    // we detect that any inputs of an ExprGroup are in these groups, we know
+    // that we cannot compute that ExprGroup so we should skip it.
```
zasdfgbnm, line 132: Could you give an example in the comment?

csrc/abstract_tensor_schedule.cpp
jacobhinkle: If we only needed to find all input `ValGroup`s then `inputGroups` would be better. However, we also need to know what kind of `ExprGroup` we have in order to create `id_expr`. To do that I check the type of some of the `Expr`s in the group until I find one all of whose inputs are computable using IDs from the given TV, then I go ahead and evaluate it.

zasdfgbnm: > However, we also need to know what kind of `ExprGroup`

Would `expr_group->front()` sufficient to know the expr type?

jacobhinkle: I am not sure. I _think_ it will always be the case that every `Expr* e` in an `ExprGroup` will have the same set of `ValGroup`s if you look at `e->inputs()` in the permissive map, and all those `Expr`s should also have compatible (proven equal) attributes. I am not sure if that is a fundamental assumption about the graph, or if it is just coincidentally true for the limited set of IterDomain expressions we currently have.

It would be simpler to just handle `expr_group->front()` and we could later try this current approach if it's needed.

csrc/abstract_tensor_schedule.cpp
jacobhinkle: Added

csrc/abstract_tensor_schedule.cpp
jacobhinkle: Added

csrc/abstract_tensor_schedule.cpp
jacobhinkle: I tried to make this work but given the complexity of using `IterDomain`s instead of `ValGroup`s for this purpose, I removed this branch so we now require using `ValGroup`s in the `AbstractTensor`. We could try and enable it in the future if needed.

csrc/abstract_tensor_schedule.h
naoyam: +1. That seems more consistent.

csrc/abstract_tensor_schedule.cpp
zasdfgbnm: We only map exprs that has matching type and attributes:
https://github.com/NVIDIA/Fuser/blob/65799ed56f66b671dfbd51336fa302400e2bdf45/csrc/val_graph.cpp#L338

csrc/abstract_tensor_schedule.h
jacobhinkle: Done

csrc/abstract_tensor_schedule.cpp
```diff
@@ -0,0 +1,319 @@
...
+
+    // Now check whether there is a concrete entry for abs_id (i.e. g). If not,
+    // that means this dimension is missing and should not be included in the
+    // loop domain, so return nullptr.
+    auto it = concrete_ids_.find(g);
+    return it == concrete_ids_.end() ? nullptr : it->second;
+  }
+
+ private:
+  const AbstractTensor& abstract_;
```
naoyam, line 303: nit: could you change the name to something like `abstract_tensor_`? There are just too many "abstract" something...

csrc/abstract_tensor_schedule.cpp
```diff
@@ -0,0 +1,319 @@
...
+  //!       iS2 = merge(iS0, iS1)
+  //!       iS3, iS4 = split(iS2, 16)
+  //!   graph:
+  //!     ValGroup 0: 0 8
+  //!     ValGroup 1: 1 9
+  //!     ValGroup 2: 2 10
+  //!     ValGroup 3: 3
+  //!     ValGroup 4: 4
+  //!     ValGroup 5: 11
+  //!     ValGroup 6: 12
```
naoyam, line 87: What do `11` and `12` refer to?

naoyam: I guess this example has a diverging transform with `iS2` as the common input?

csrc/abstract_tensor_schedule.cpp
```diff
@@ -0,0 +1,319 @@
...
+        // Avoid cycles
+        continue;
+      }
+      for (const ExprGroup& eg : graph_->getDefinitions(vg)) {
+        for (Expr* e : *eg) {
+          for (Val* inp : e->inputs()) {
+            ValGroup vg_inp = graph_->toGroup(inp);
+            vg_stack.push(vg_inp);
+          }
+        }
```
naoyam, line 122: I think this is similar to the case below, but does it need to check all exprs in an ExprGroup?

csrc/abstract_tensor_schedule.cpp
```diff
@@ -0,0 +1,319 @@
...
+  //!     ExprGroup 0: iS11{ceilDiv(i0 * i1, 32)}, iS12{32} = split(iS10, 32)
+  //!   abstract_:
+  //!     {{ValGroup 5, graph}, {ValGroup 6, graph}}
+  //!
+  //! Then findNearestProducers() work backward from the loop domain of
+  //! concrete_ (ValGroups 3 and 4) and will identify ValGroup 2 as a producer
+  //! to abstract_ and map it to the concrete domain iS2{i0}.
+  //!
+  //! Note that any dangling transforms will be discarded; in this example since
+  //! we work from iS2, we will discard the split that produced iS3 and iS4.
```
naoyam, line 97: Why `iS3` and `iS4` are considered dangling? Does this mean that if existing transforms of a TensorView could be altered?

csrc/abstract_tensor_schedule.cpp
```diff
@@ -0,0 +1,319 @@
...
+  //!     ValGroup 3: 3
+  //!     ValGroup 4: 4
+  //!     ValGroup 5: 11
+  //!     ValGroup 6: 12
+  //!     ExprGroup 0: iS11{ceilDiv(i0 * i1, 32)}, iS12{32} = split(iS10, 32)
+  //!   abstract_:
+  //!     {{ValGroup 5, graph}, {ValGroup 6, graph}}
+  //!
+  //! Then findNearestProducers() work backward from the loop domain of
+  //! concrete_ (ValGroups 3 and 4) and will identify ValGroup 2 as a producer
```
naoyam, line 93: This comment is confusing to me. The code below traverses backward from the abstract domains, not the loop domains of the concrete tensor. What am I missing?

csrc/abstract_tensor_schedule.cpp
jacobhinkle: Yes, I use `setLoopDomain` instead of replaying transforms, and I use this function to find the starting point, which might not be IDs in the current loop domain. This means that there could be transforms in the current tensor that get erased.

csrc/abstract_tensor_schedule.cpp
jacobhinkle: OK great. I'll change to only checking the first Expr in the group as you suggested.

csrc/abstract_tensor_schedule.cpp
jacobhinkle: We first traverse `abstract_tensor_` backward to get a collection of all the `ValGroup`s that appear in the history of `abstract_tensor_`. Then we traverse `tv_` to find the latest `IterDomain`s that have a `ValGroup` that was found in the first step: https://github.com/NVIDIA/Fuser/blob/88ababfe91e5ad0d4b28946ae490c43b351f286b/csrc/abstract_tensor_schedule.cpp#L127-L128.

csrc/abstract_tensor_schedule.cpp
jacobhinkle: You're right. I will convert this to just look at the first `Expr` also.

csrc/abstract_tensor_schedule.cpp
jacobhinkle: I'll make this more clear. It was shorthand for `iS11` and `iS12`, similar to how it looks when printing a `ValGraph`.

This example was needlessly confusing since I made iS0 and iS5 mapped with one another, so it's a diverging transform as you say. I changed it to no longer have a diverging transform: the abstract tensor now just represents a merge and split on two axes followed by a merge of two other axes, and I show a concrete TV with three axes so that the last transform is uncomputable.

csrc/abstract_tensor_schedule.cpp
naoyam: In the existing replay API, it isn't allowed to change existing transforms. Why do we need to change that here? Sounds a little dangerous to quietly discard transformations.

jacobhinkle: You're right that this is different from the replay approach. My motivation is to get away from the statefulness of that approach. If we were to maintain transforms then that puts a burden on the schedulers to guarantee which transforms have already been applied to each tensor. Instead, this lets us do that with `AbstractTensor`s instead, which are easier to work with, can be copied, etc. However, I agree that silently discarding transforms is probably not ideal, unless we have a specific use case where it is helpful. So I think I will just assert that all the loop `IterDomain`s are found in the history of the `AbstractTensor`.

csrc/abstract_tensor_schedule.cpp
naoyam: > a concrete TV with three axes

Are you referring to `tv_`? Both root and loop domains seem to be 2D. Which three axes are you referring to?

jacobhinkle: Ah, I meant this reply for the other code comment. In this case you're right that we have a loop transform in `tv_` that differs from what's present in `abstract_tensor_`. As discussed in another comment, I'm going to start asserting that this is not the case, and I will update this comment accordingly. Essentially instead of `findNearestProducers()` we will have something like `checkLoopIsScheduled` that checks that all the loop ValGroups of `tv_` are in the history of `abstract_tensor_`.

csrc/abstract_tensor_schedule.cpp
naoyam: Maybe if existing domains are set as loop or allocation domains, they won't be just discarded, right? 

In any case, I think this behavior would need to be more explicit.

csrc/abstract_tensor_schedule.cpp
naoyam: OK, sounds good. Let me know when it's again ready for review.

csrc/abstract_tensor_schedule.cpp
```diff
@@ -0,0 +1,290 @@
...
+    // Look up each loop IterDomain in tv and assert that it is in
+    // scheduled_val_groups_, then map it.
+    for (IterDomain* loop_id : tv->getLoopDomain()) {
+      ValGroup vg = graph_->toGroup(loop_id);
+      // NOTE: this check guarantees that we will only ever _append_ transforms
+      // to the loop domain of tv. If instead we searched for producer IDs in
+      // the root->loop path of tv, then we could wind up silently erasing
+      // some pre-existing transforms.
+      NVF_CHECK(
+          scheduled_val_groups_.count(vg) > 0,
```
naoyam, line 112: I didn't expect this should be considered an error. We could do some best-effort propagation (I think that's the behavior of the existing transform propagator). I'm not sure which would be better. Since AbstractTensor could be anything, it may make sense to enforce it must have all corresponding domains, and that may be also good for correctness.

csrc/abstract_tensor_schedule.cpp
```diff
@@ -0,0 +1,290 @@
...
+          loop_id->toString(),
+          " that is not described by abstract tensor.");
+      tv_ids.emplace(vg, loop_id);
+    }
+
+    return tv_ids;
+  }
+
+  IterDomain* replayAbstractId(AbstractId abs_id, GroupIdMap& tv_ids) {
+    ValGroup g = abstractIdToValGroup(abs_id);
```
naoyam, line 123: nit: can this be `const &`? It's a shared_ptr, so copying is not as cheap as just copying a raw pointer.

csrc/abstract_tensor_schedule.cpp
```diff
@@ -0,0 +1,290 @@
...
+    //     ValGroup 4: iS4
+    //     ValGroup 5: iS5
+    //     ValGroup 6: iS7
+    //     ValGroup 7: iS8
+    //     ExprGroup 0:
+    //       iS3 = merge(iS0, iS1)
+    //     ExprGroup 1:
+    //       iS4, iS5 = split(iS3, 16)
+    //     ExprGroup 2:
+    //       iS8 = merge(iS6, iS7)
```
naoyam, line 147: Where are `iS6` and `iS7` defined?

csrc/abstract_tensor_schedule.cpp
```diff
@@ -0,0 +1,290 @@
...
+          id_expr = new_ids.first->definition();
+        } else if (auto* s = dynamic_cast<Swizzle2D*>(expr)) {
+          NVF_ERROR(id_inps.size() == 2);
+          auto new_ids = IterDomain::swizzle(
+              s->swizzleType(), id_inps[0], id_inps[1], s->swizzleMode());
+          id_expr = new_ids.first->definition();
+        } else {
+          NVF_ERROR(
+              false, "Unhandled IterDomain expression ", expr->toString());
+        }
```
naoyam, line 231: Can we reuse some of the existing replay APIs? For example, `ReplayTransform` in id_model/transform_replay.h?

csrc/abstract_tensor_schedule.cpp
```diff
@@ -0,0 +1,290 @@
...
+    std::unordered_set<ValGroup> uncomputable_groups;
+
+    std::stack<ValGroup> vg_stack({g});
+    // Strategy: fill out tv_ids in the direction of g
+    auto propagate = [&](ValGroup vg) {
+      // We need to try and produce an ID for this group, so we look at
+      // definitions for the Vals in this group and check whether we have their
+      // inputs computed yet (tv_ids) and whether they can be computed
+      // (uncomputable_groups).
+      std::vector<ValGroup> uncomputed_producer_groups;
```
naoyam, line 169: I'm getting confused here. Could you add a comment explaining what this vector means?

csrc/abstract_tensor_schedule.cpp
```diff
@@ -0,0 +1,290 @@
...
+    //
+    // However, ValGroup 7 is not computable. It is produced by ExprGroup 2
+    // whose producer ValGroups are 2 and 6. ValGroup 2 is computable since iS0
+    // is in tv, however there is no IterDomain in tv that can be
+    // used to represent ValGroup 6 which also has no producer ValGroups.
+    std::unordered_set<ValGroup> uncomputable_groups;
+
+    std::stack<ValGroup> vg_stack({g});
+    // Strategy: fill out tv_ids in the direction of g
+    auto propagate = [&](ValGroup vg) {
```
naoyam, line 164: This lambda seems quite complex. It seems it does multiple things, including finding proper replay inputs, actually replaying domains, etc. I'd suggest rewriting the lambda such that the semantic difference would be more explicitly highlighted.

csrc/abstract_tensor_schedule.cpp
```diff
@@ -0,0 +1,290 @@
...
+    // produced by ExprGroup 1 which itself produced by ExprGroup 0, and
+    // tv includes iS0 and iS1.
+    //
+    // However, ValGroup 7 is not computable. It is produced by ExprGroup 2
+    // whose producer ValGroups are 2 and 6. ValGroup 2 is computable since iS0
+    // is in tv, however there is no IterDomain in tv that can be
+    // used to represent ValGroup 6 which also has no producer ValGroups.
+    std::unordered_set<ValGroup> uncomputable_groups;
+
+    std::stack<ValGroup> vg_stack({g});
```
naoyam, line 162: Would there be a better name? It's a `std::stack` with `ValGroup`, so `vg_stack` is of course fine, but it doesn't help to understand what this stack is used for. At least, a comment would be required.

csrc/abstract_tensor_schedule.cpp
jacobhinkle: > Since AbstractTensor could be anything, it may make sense to enforce it must have all corresponding domains, and that may be also good for correctness.

The line above is checking that the `AbstractTensor` contains all the loop domains of the `TensorView`. Relaxing this requirement is not a bad idea: we would just "ignore" those domains and not schedule them. However, it's unclear where we should place them in the loop domain. For example
```
T0 [ i0 i1 i2 ]
AbstractTensor [ VG{i0 * i2} ]
```
In this case the middle dimension from `T0` is "unscheduled" and the first and last dimensions need to be merged, so where would we put the `i1` dimension in the output? To avoid this ambiguity, I am just forcing us to explicitly include each loop `ValGroup` in the `AbstractTensor` so that this choice is explicit.

csrc/abstract_tensor_schedule.cpp
jacobhinkle: They are defined elsewhere and could be associated with a `TensorView` or not. For example, `AbstractTensor::split` and `AbstractTensor::merge` will create such `IterDomain`s in order to create an `ExprGroup`. I'll update the comment to try and clarify.

csrc/abstract_tensor_schedule.cpp
jacobhinkle: Thanks for the suggestion. I reused `ReplayTransformations` from `transform_iter.h` to avoid this tedious bit of code.

csrc/abstract_tensor_schedule.cpp
jacobhinkle: Explained more in current version

csrc/abstract_tensor_schedule.cpp
jacobhinkle: You're right. I removed the lambda and tried to streamline the body of the stack loop. Moving out `replayIdExpr` helped. I also added some more comments and changed the name of the stack to `eval_stack` since it's the stack of `ValGroup`s to try evaluating next.

csrc/abstract_tensor_schedule.cpp
naoyam: The existing transform replay also has the same problem. IIRC, it just puts the remaining domains at the end of the list, i.e., the innermost position. 

For AbstractTensor, I think avoiding the ambiguity is better.

csrc/abstract_tensor_schedule.cpp
naoyam: I suspected that but wasn't clear if it's just a mistake. Can you please clarify that in the comment?

jacobhinkle: I added this part to the comment: https://github.com/NVIDIA/Fuser/blob/7a51c9a9ee668ade6ccdba92453a34a48729ce6a/csrc/abstract_tensor_schedule.cpp#L172-L185

csrc/abstract_tensor_schedule.cpp
```diff
@@ -0,0 +1,315 @@
...
+    //     ExprGroup 0:
+    //       iS3 = merge(iS0, iS1)
+    //     ExprGroup 1:
+    //       iS4, iS5 = split(iS3, 16)
+    //     ExprGroup 2:
+    //       iS8 = merge(iS6, iS7)
+    //
+    //   abstract_tensor_.domain: ValGroup 4, ValGroup 5, ValGroup 7
+    //
+    //      VG0   VG1
```
naoyam, line 174: I think it'd be helpful to emphasize that this graph represents the transformations embedded in the abstract tensor and that only `VG0`, `VG1` and `VG2` are the ones that already for the tensor. When I first saw it, I mistakenly thought the left tree represented the transformation of the tensor, which isn't correct.

csrc/abstract_tensor_schedule.cpp
```diff
@@ -0,0 +1,315 @@
...
+    // iS1.
+    //
+    // However, ValGroup 7 is not computable. It is produced by ExprGroup 2
+    // whose input ValGroups are 2 and 6. ValGroup 2 is computable since iS2 is
+    // in tv, however there is no IterDomain in tv that can be used to represent
+    // ValGroup 6 which also has no producer ValGroups.
+
+    // Any computable ValGroup should have an entry in computed_ids. If we prove
+    // that the ValGroup is not computable, we insert it here.
+    std::unordered_set<ValGroup> uncomputable_groups;
```
naoyam, line 198: Please also mention like below:
```
// In the case of the above example, VG6 and VG7 would be included.
```

jacobhinkle: Good idea. I'll leave a comment that VG6 would be included here. In fact we don't mark VG7 as uncomputable here; since VG2 is computable even though VG6 is not, we just use the representative from VG2 to represent VG7 i.e. if there is nothing to merge with we just skip merging and use the existing input ID.

naoyam: > we just use the representative from VG2 to represent VG7 i.e. if there is nothing to merge with we just skip merging and use the existing input ID

Can you elaborate why?

jacobhinkle:
I realized when responding to these comments that there might be a problem with how I am currently updating the `ValGraph`. When all the inputs to an `Expr` are computable, I compute a new ID and register it with the `ValGroup`: https://github.com/NVIDIA/Fuser/blob/7a51c9a9ee668ade6ccdba92453a34a48729ce6a/csrc/abstract_tensor_schedule.cpp#L256

The problem is that this "computability" condition is non-recursive: if we have "computed" a merge with a missing dimension by passing through the existing dimension, then that unmerged axis will represent the merged `ValGroup`. It will appear "computed" in downstream expressions. Those expressions might be replayed and their outputs erroneously mapped to the `ValGroup` in the line shown above. So I think I should not do this mapping (i.e. I should remove that line)

My motivation for updating the `ValGraph` in the first place was so that after we apply the transforms we are able to determine which axes from the `AbstractTensor` the loop domains of the `TensorView` now correspond to. I think a cleaner approach might be to simply return a vector of integers indicating, for each loop domain, which `AbstractId` in `AbstractTensor` each axis corresponds to.

csrc/abstract_tensor_schedule.cpp
jacobhinkle: For example if we have
```
tv2[ iS1{i0} bS2{1} ] =  broadcast( tv0[ iS0{i0} ] )
tv3[ iS3{i0} iS4{i1} ] = mul(tv1, tv2)
tv3->merge(0)
```
Now if we propagate from tv3, this should leave `iS0` in place in `tv0`. This is the type of thing I was trying to support while still applying the same `AbstractTensor` all tensors.

naoyam: We only do this forwarding for certain patterns of broadcast domains. See `ForwardingInfo`.

naoyam: Is this forwarding behavior necessary for your matmul case? If not, we could revisit later. We may be able to remove that eventually with #2459.

jacobhinkle: I am not sure I really need it actually.

Specifically for scheduling matmul, we need to do the following:

1. Cache epilogue inputs/outputs
2. cache operands for smem and register loads
3. Apply basic tiling to mma_result (i.e. merge dimensions with same roles, reorder so that roles are in canonical order). This results in something like [B, Mo, No, Ko, Mi, Ni, Ki]
4. Split Ko by splitk_factor: [B, Mo, No, Kf, Kg, Mi, Ni, Ki]
5. rFactor({Kf, Ki}) to create splitk_sum
6. Cache mma_result to smem for smem_epilogue
7. Schedule all tensors. This might involve multiple `applyAbstractTransforms` and `parallelizeAllLike` but it should not introduce new tensors.

Steps 1, 2, 5, and 6 create new tensors. Step 5 must happen after some transforms have been applied. With the current machinery I have steps 1-6 and can progress to step 7 which involves a bunch of substeps.

After I have scheduled `mma_result` in step 3, I identify the Ko and Ki dimensions using its loop `ValGroup`s returned by `graph.toGroup(id)`. After step 5, I rebuild the `IdModel` since there are new tensors, and I update the `AbstractTensor` used in steps 3/4 to point to the new ValGroups/ValGraph. At that point, I hope to be able to infer the role of each loop domain in every tensor in the fusion for step 7.

In step 7, we'll need to at times apply the same transforms to pre- and post-broadcast tensors, for example if we are applying a 1D or 0D bias in the epilogue, or when scheduling the prologue tensors that broadcast a missing M or N dimension on either operand. In those cases, it seemed more convenient to just skip merging the missing dimensions than to manually build a different AbstractTensor for each pattern found in the fusion (not that that couldn't also be done if we need to).

jacobhinkle: > Is this forwarding behavior necessary for your matmul case? If not, we could revisit later. We may be able to remove that eventually with #2459.

I think we can work around it, at the expense of more book-keeping. Currently, I schedule a single `AbstractTensor` to represent tiling then I can apply that to every tensor in the matmul fusion. If we disable forwarding, then we will need a different `AbstractTensor` for each group of tensors with/without some axes to avoid an error.

How about this: we disable automatic forwarding and instead move it to a utility like this:
```c++
//! Forward missing axes in merge and swizzle expressions (i.e. ignore the expression
//! and keep the axis intact). Axis presence is determined based on the provided
//! TensorView. All AbstractIds in abten must be of type ValGroupAndItsGraph.
AbstractTensor forwardMissingAxes(const AbstractTensor& abten, TensorView* tv);
```
This way the behavior is explicit and the scheduler is in charge of guaranteeing this is the desired behavior.

csrc/abstract_tensor_schedule.cpp
jacobhinkle: I expanded on the comment and added another example showing a ValGroup that is safe to ignore since it has no computable producer groups, as well as one that will cause an error since it has multiple producer groups and only some of them are computable.

csrc/abstract_tensor_schedule.cpp
naoyam: Sounds reasonable

naoyam:
Is it ready to review again or are you still making changes?

jacobhinkle:
> Is it ready to review again or are you still making changes?

Ready to review. I have been experimenting with using this for matmul and I think it will work.

csrc/abstract_tensor_schedule.cpp
```diff
@@ -0,0 +1,350 @@
...
+          break;
+        } else {
+          // Some input is not yet computed
+          if (uncomputed_producer_groups.empty() && !id_inps.empty()) {
+            // The only uncomputed producer groups are uncomputable, but some
+            // are computed. Just pass those computed inputs through. Note that
+            // in this case, id_inps.front() will be the representative of vg
+            // even though it is _not_ actually mapped into that ValGroup.
+            NVF_ERROR(id_inps.size() == 1);
+            computed_ids.emplace(vg, id_inps.front());
```
naoyam, line 302: Is this for forwarding? I thought that's disabled by default.

csrc/abstract_tensor_schedule.cpp
```diff
@@ -0,0 +1,350 @@
...
+        }
+        if (!uncomputed_producer_groups.empty()) {
+          // Do not look at other ExprGroups before we try computing these
+          // uncomputed producer groups
+          break;
+        }
+      }
+
+      if (uncomputed_producer_groups.empty()) {
+        // All of the defining expressions are proven uncomputable, so mark vg
```
naoyam, line 314: Why is this the case? We are still within the while loop of `eval_stack`. Suppose `vg` has all inputs computed, so nothing would be added to `uncomputed_producer_groups`. Am I missing something?

csrc/abstract_tensor_schedule.cpp
jacobhinkle: Sorry! I pushed to the wrong remote before asking for review. I just pushed the real change, which changes this to an assert and adds a utility `forwardAroundMissingAxes`, which derives a new AbstractTensor for a particular TV such that merges with missing axes are skipped.

csrc/abstract_tensor_schedule.cpp
jacobhinkle: I rearranged the code slightly to make it more clear what condition I'm asserting on.

csrc/abstract_tensor_schedule.cpp
```diff
@@ -0,0 +1,460 @@
...
+            computed_ids.emplace(vg_outp, id_outp);
+            vg_computed = true;
+          }
+          // No need to look at next ExprGroup
+          break;
+        }
+        if (!unprocessed_producer_groups.empty()) {
+          // Do not look at other ExprGroups before we try computing these
+          // unprocessed producer groups
+          break;
```
naoyam, line 299: This `break` effectively leads to the `else` block at line 312, right? The control flow here is not trivial to follow. Some clarifying comments would be helpful.

csrc/abstract_tensor_schedule.cpp
```diff
@@ -0,0 +1,460 @@
...
+        }
+        NVF_ERROR(
+            !some_inputs_uncomputable || computed_inputs.empty(),
+            "There are both computable and uncomputable producer groups."
+            "Refusing to automatically forward those computed inputs.",
+            " Expr: ",
+            expr->toString(),
+            " Computed ID: ",
+            computed_inputs.front()->toString());
+        if (all_inputs_computed) {
```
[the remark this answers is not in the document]
jacobhinkle: Your understanding is correct. Thanks for the suggestion.

csrc/abstract_tensor_schedule.cpp
```diff
@@ -0,0 +1,460 @@
...
+      // (uncomputable_groups). If an ExprGroup does not have any computable
+      // inputs, we move on to the next ExprGroup. There might be inputs that
+      // have not yet been computed or proven uncomputable; in those cases we
+      // will place them into the vector below so that we can push them onto
+      // the stack so that we can try again.
+      std::vector<ValGroup> unprocessed_producer_groups;
+      // This will be set to true once we have computed an IterDomain for this
+      // ValGroup
+      bool vg_computed = false;
+      for (const ExprGroup& eg : graph_->getDefinitions(vg)) {
```
naoyam, line 247: IIUC, checking multiple definitions would matter because there could be multiple different exprs in some graphs like the Permissive graph. If this graph is the Exact graph, there should be only one ExprGroup, so this loop should not matter. Am I understanding correctly?

csrc/abstract_tensor_schedule.cpp
jacobhinkle: Added more comments to help clarify.

csrc/abstract_tensor_schedule.cpp
```diff
@@ -0,0 +1,474 @@
...
+} // namespace
+
+void applyAbstractTransforms(
+    const AbstractTensor& abstract_tensor,
+    const std::vector<TensorView*>& tvs,
+    ValGraph* graph) {
+  AbstractTensorSchedule::apply(abstract_tensor, tvs, graph);
+}
+
+AbstractTensor forwardAroundMissingAxes(
```
naoyam, line 384: Something still feels off to me. We are using the Permissive graph here, so if we assume that, the graph should already knows about the broadcast forwarding. Here, it seems we are forwarding whatever missing domains. I'm trying to understand why this is necessary. Could you give me a concrete example?

jacobhinkle: I'll add one as a comment, but the problem is not at the site of the broadcast but before it, where the broadcast axis does not yet exist.

naoyam: Actually, ignore the above comment. The new test seems like a good example.

csrc/abstract_tensor_schedule.cpp
```diff
@@ -0,0 +1,474 @@
...
+      // the originals due to previous forwarding.
+      std::vector<ValGroup> computable_output_groups =
+          replayExprGroup(graph, eg, computable_valgroups);
+      NVF_ERROR(computable_output_groups.size() == output_groups.size());
+      for (size_t i : c10::irange(output_groups.size())) {
+        computable_valgroup_map.emplace(
+            output_groups[i], computable_output_groups[i]);
+      }
+    } else if (computable_valgroups.size() == 1 && eg->front()->isA<Merge>()) {
+      // Forward the input group
```
naoyam, line 447: I'm still not comfortable here. Why is it ok to forward whatever missing merge inputs?

tests/cpp/test_abstract_tensor.cpp
```diff
@@ -712,4 +717,63 @@ TEST_F(AbstractTensorTest, UnzipBroadcasting) {
...
+
+  for (TensorView* tv : {tv3, tv4, tv5, tv6}) {
+    AbstractTensor local_abten = forwardAroundMissingAxes(abten, tv);
+    applyAbstractTransforms(local_abten, tv, &graph);
+    tv->axis(-1)->parallelize(ParallelType::TIDx);
+    tv->axis(-2)->parallelize(ParallelType::BIDx);
+  }
+
+  inlineMost();
+
```
naoyam, line 765: Could you also validate the transformation results as well?

csrc/abstract_tensor_schedule.cpp
jacobhinkle: This is the type of behavior I'm trying to enable here:
```c++
  TensorView* tv0 = makeSymbolicTensor(1);
  TensorView* tv1 = makeSymbolicTensor(2);
  fusion.addInput(tv0);
  fusion.addInput(tv1);

  TensorView* tv2 = mul(tv0, IrBuilder::create<Val>(3.0));
  TensorView* tv3 = broadcast(tv2, {false, true});
  TensorView* tv4 = mul(tv3, tv1);

  fusion.addOutput(tv4);

  tv4->merge(0);
  tv4->split(0, 32);
  tv4->axis(-1)->parallelize(ParallelType::TIDx);
  tv4->axis(-2)->parallelize(ParallelType::BIDx);

  scheduler_utils::transformPropagateToAllFrom(tv4, -1);
  scheduler_utils::parallelizeAllLike(tv4);

  inlineMost();

  fusion.printMath();
  fusion.printTransforms();
```
```
Inputs:
  T0_g[ iS19{( ceilDiv(i0, 32) )}, iS20{32} ], float
  T1_g[ iS12{( ceilDiv(( i2 * i3 ), 32) )}, iS13{32} ], float
Outputs:
  T4_g[ iblockIdx.x9{( ceilDiv(( i0 * i3 ), 32) )}, ithreadIdx.x10{32} ] ca_pos( 2 ) produce_pos( 2 ), float

%kernel_math {
T2_l[ iblockIdx.x17{( ceilDiv(i0, 32) )}, ithreadIdx.x18{32} ] ca_pos( 2 )
   = T0_g[ iS19{( ceilDiv(i0, 32) )}, iS20{32} ]
   * double(3);
T3_l[ iblockIdx.x15{( ceilDiv(( i0 * 1 ), 32) )}, ithreadIdx.x16{32} ] ca_pos( 2 ) produce_pos( 2 )
   = broadcast( T2_l[ iblockIdx.x17{( ceilDiv(i0, 32) )}, ithreadIdx.x18{32} ] ca_pos( 2 ) )
T4_g[ iblockIdx.x9{( ceilDiv(( i0 * i3 ), 32) )}, ithreadIdx.x10{32} ] ca_pos( 2 ) produce_pos( 2 )
   = T3_l[ iblockIdx.x15{( ceilDiv(( i0 * 1 ), 32) )}, ithreadIdx.x16{32} ] ca_pos( 2 ) produce_pos( 2 )
   * T1_g[ iS12{( ceilDiv(( i2 * i3 ), 32) )}, iS13{32} ];
}

T0_g[ iS19{( ceilDiv(i0, 32) )}, iS20{32} ]
 logical domain : (iS0{i0})
 contiguity: f
  Split: iS0{i0} by factor 32 -> iS19{( ceilDiv(i0, 32) )}, iS20{32}
 loop domain : (iS19{( ceilDiv(i0, 32) )}, iS20{32})
T2_l[ iblockIdx.x17{( ceilDiv(i0, 32) )}, ithreadIdx.x18{32} ] ca_pos( 2 )
 logical domain : (iS3{i0})
 contiguity: t
  Split: iS3{i0} by factor 32 -> iblockIdx.x17{( ceilDiv(i0, 32) )}, ithreadIdx.x18{32}
 loop domain : (iblockIdx.x17{( ceilDiv(i0, 32) )}, ithreadIdx.x18{32})
T3_l[ iblockIdx.x15{( ceilDiv(( i0 * 1 ), 32) )}, ithreadIdx.x16{32} ] ca_pos( 2 ) produce_pos( 2 )
 logical domain : (iS4{i0}, bS5{1})
 contiguity: t n
  Merge: iS4{i0} and bS5{1} -> iS14{( i0 * 1 )}
  Split: iS14{( i0 * 1 )} by factor 32 -> iblockIdx.x15{( ceilDiv(( i0 * 1 ), 32) )}, ithreadIdx.x16{32}
 loop domain : (iblockIdx.x15{( ceilDiv(( i0 * 1 ), 32) )}, ithreadIdx.x16{32})
T1_g[ iS12{( ceilDiv(( i2 * i3 ), 32) )}, iS13{32} ]
 logical domain : (iS1{i2}, iS2{i3})
 contiguity: f f
  Merge: iS1{i2} and iS2{i3} -> iS11{( i2 * i3 )}
  Split: iS11{( i2 * i3 )} by factor 32 -> iS12{( ceilDiv(( i2 * i3 ), 32) )}, iS13{32}
 loop domain : (iS12{( ceilDiv(( i2 * i3 ), 32) )}, iS13{32})
T4_g[ iblockIdx.x9{( ceilDiv(( i0 * i3 ), 32) )}, ithreadIdx.x10{32} ] ca_pos( 2 ) produce_pos( 2 )
 logical domain : (iS6{i0}, iS7{i3})
 contiguity: t t
  Merge: iS6{i0} and iS7{i3} -> iS8{( i0 * i3 )}
  Split: iS8{( i0 * i3 )} by factor 32 -> iblockIdx.x9{( ceilDiv(( i0 * i3 ), 32) )}, ithreadIdx.x10{32}
 loop domain : (iblockIdx.x9{( ceilDiv(( i0 * i3 ), 32) )}, ithreadIdx.x10{32})
```
Although we start with a merge on `tv4`, we "ignore" it when propagating back to `tv2`, and we still replay the `Split`. The "forwarding around missing axes" is just that bit of using the existing input to stand in for the merged axis.

naoyam: Yes, but that forwarding is not done arbitrary but is only done for some specific cases as defined by `ForwardingInfo`. My concern is in this case it seems it just ignores any missing merge inputs.

jacobhinkle: When I read through the `ForwardingInfo` code [here](https://github.com/NVIDIA/Fuser/blob/main/csrc/transform_iter.cpp#L723-L832) it seemed like it was just doing something equivalent to what I implemented here: it forwards axes that are squeezed or broadcast through Merge expressions.

jacobhinkle: In `ForwardingInfo` the propagation is done only across `BroadcastOp` and `SqueezeOp` expressions. I think this is actually how the Permissive graph is built in `IdModel`, by checking `ForwardingInfo` for each expression. So in that sense the logic is still in effect since "missing" in the `AbstractTensor` case corresponds to "not permissive mapped into the given ValGroup".

naoyam: I think you're right, but it seems this could expand the definition of the Permissive graph. Let's say:

```
// tv0: [I0]
auto tv0 = makeSymbolicTensor(1); // input
// tv1: [I0, I1]
auto tv1 = makeSymbolicTensor(2); // input
// tv0_cache: [I0]
auto tv0_cache = set(tv0);
// tv1_cache: [I0, I1]
auto tv1_cache = set(tv1);
// tv2: [I0, R1]
auto tv2 = sum(tv1_cache, {1});
// tv3: [I0]
auto tv3 = add(tv0_cache, tv2); // output

tv1_cache->merge(0, 1)->split(0, 4);
// tv1_cache: [I0*I1/4, 4]
```

Here, `ForwardingInfo` would not do anything as there's no broadcast, so the `I0*I1` domain of `tv1_cache` would be mapped neither with `I0` or `I1`, which would mean `tv0_cache` wouldn't be inlined into `tv1_cache`. In other words, the current transform propagator wouldn't propagate the merge and split of `tv1_cache` to `tv0_cache`. That would be fine since `tv0_cache` wasn't used for `tv1_cache`.

Now, with this AbstractTensor scheduling, if an AbstractTensor consists of the loop domains of `tv1_cache` and if it's applied to `tv0_cache`, IIUC, it would transform `tv0_cache` by skipping the merge and applying the split. The resulting Loop graph would be also updated accordingly.

To me, this seems like a subtle but important difference compared to the current behavior. I'm not agains the new behavior. I'm just trying to understand the difference as clearly as possible.

Let me know if I'm missing anything.

tests/cpp/test_abstract_tensor.cpp
```diff
@@ -712,4 +717,72 @@ TEST_F(AbstractTensorTest, UnzipBroadcasting) {
...
+  auto tv0 = makeSymbolicTensor(1);
+  auto tv1 = makeSymbolicTensor(2);
+  auto tv2 = makeSymbolicTensor(2);
+  fusion.addInput(tv0);
+  fusion.addInput(tv1);
+  fusion.addInput(tv2);
+
+  // This fusion does not have a reference tensor
+  auto tv3 = set(tv0);
```
jacobhinkle, line 732: I'm currently trying to understand this. If I remove this `set` the test passes. Otherwise, it fails during lowering (the transform checks pass):
> C++ exception with description "alloc_ind != nullptr INTERNAL ASSERT FAILED at "/opt/pytorch/nvfuser/csrc/index_compute.cpp":1818, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. Couldn't find allocation mapping for T1_g[ iS65{( ceilDiv(( ( (( (( getMetaData(T0) )).logical_size ))[0] ) * ( (( (( getMetaData(T1) )).logical_size ))[1] ) ), 128) )}, iS66{128} ] dim: 1 id: iS53{( (( (( getMetaData(T1) )).logical_size ))[1] )}

csrc/abstract_tensor_schedule.cpp
jacobhinkle: I think you're right and reductions are where we might have a difference since that is the other place besides squeeze where we might remove an axis.

It's true that currently I am updating the graph by inserting the new IterDomain when it's created. I can make that optional. I will think more on the effect of examples like this.

In the matmul case I think I don't hit this because we are not merging reduction groups with non reduction groups and propagating past the reduction. In that case I am only using the graph for this purpose so corrupting it is not a big concern. This is one reason I avoid rebuilding the IdModel after starting scheduling (another is that the ExprGroups introduced by scheduling an AbstractTensor are lost when the graph is rebuilt).

naoyam: Note that the above example does not merge reduction and non-reduction domains.

I think it makes most sense to start with the least change from the current behavior and explore extensions as next steps. In particular, can we drop `forwardAroundMissingAxes` for now and would that still be enough for the mamtul fusion?

jacobhinkle: You're right that the example doesn't merge reduction and non-reduction domains directly; I meant that if we used `applyAbstractTransforms` in this case that we would do that merge/split on `tv2` and we would forward around the merge and apply the split on `tv0_cache`/`tv3`. This is indeed a difference with this approach compared to propagation.

BTW I tried your example with propagation and unless you explicitly stop propagating at the reduction it will also error out by trying to merge the reduction/iteration domains in `tv2`. We actually hit the exact same error message in either case:
```c++
TEST_F(AbstractTensorTest, TestApplyScheduling2) {
  Fusion fusion;
  FusionGuard fg(&fusion);

  // tv0: [I0]
  auto tv0 = makeSymbolicTensor(1); // input
  // tv1: [I0, I1]
  auto tv1 = makeSymbolicTensor(2); // input
  fusion.addInput(tv0);
  fusion.addInput(tv1);
  // tv0_cache: [I0]
  auto tv0_cache = set(tv0);
  // tv1_cache: [I0, I1]
  auto tv1_cache = set(tv1);
  // tv2: [I0, R1]
  auto tv2 = sum(tv1_cache, {1});
  // tv3: [I0]
  auto tv3 = add(tv0_cache, tv2); // output
  fusion.addOutput(tv3);

  // I get the same error regardless of this condition:
  // Merging IterDomains requires that their iteration types match. Outer: iS9{i0}, Inner: rS10{i3}
  if (true) {
    tv1_cache->merge(0, 1)->split(0, 4);
    // tv1_cache: [I0*I1/4, 4]

    scheduler_utils::transformPropagateToAllFrom(tv1_cache, -1);

    tv1_cache->axis(-1)->parallelize(ParallelType::TIDx);
    tv1_cache->axis(-2)->parallelize(ParallelType::BIDx);

    scheduler_utils::parallelizeAllLike(tv1_cache);

    fusion.printTransforms();
  } else {
    IdModel id_model(&fusion);
    ValGraph& graph = id_model.idGraph(IdMappingMode::PERMISSIVE);
    AbstractTensor abten;

    abten.domain.reserve(2);
    const auto addAbstractAxis = [&abten, &graph](IterDomain* id) {
      ValGroup g = graph.toGroup(id);
      abten.domain.emplace_back(ValGroupAndItsGraph{g, &graph});
    };
    addAbstractAxis(tv1_cache->axis(0));
    addAbstractAxis(tv1_cache->axis(1));

    abten.merge(0, 1);
    abten.split(0, 4);

    for (TensorView* tv : {tv0_cache, tv1_cache, tv2, tv3}) {
      AbstractTensor local_abten = forwardAroundMissingAxes(abten, tv);
      applyAbstractTransforms(local_abten, tv, &graph);

      // Check that every tensor is scheduled with two loop dimensions and has
      // inner dimension 4 resulting from a Split
      EXPECT_EQ(tv->nDims(), 2);
      EXPECT_TRUE(tv->axis(-1)->extent()->isConst());
      EXPECT_EQ(tv->axis(-1)->extent()->value().as<int64_t>(), 4);
      EXPECT_TRUE(tv->axis(-1)->definition()->isA<Split>());

      tv->axis(-1)->parallelize(ParallelType::TIDx);
      tv->axis(-2)->parallelize(ParallelType::BIDx);
    }
  }

  inlineMost();

  at::Tensor t0 = at::randn({24}, options);
  at::Tensor t1 = at::randn({24, 1024}, options);
  std::vector<c10::IValue> inputs{t0, t1};

  FusionExecutor fe;
  fe.compileFusion(&fusion, inputs);
  auto cg_outputs = fe.runFusion(inputs);

  testValidate(&fusion, cg_outputs, inputs, __LINE__, __FILE__);
}
```

I agree with an incremental approach and I'm fine with taking out `forwardAroundMissingAxes` and looking at it in a separate PR. I do think we'll want to have it eventually though since for example in our existing test we would need to manually schedule three separate `AbstractTensor`s in order to do the scheduling, at which point we might as well just write a function that schedules a concrete `TensorView` and apply that directly instead of using an `AbstractTensor`.

tests/cpp/test_abstract_tensor.cpp
jacobhinkle: I think this failure might show why #2459 is preferable to forwarding through merges around missing axes. `tv3` here is 1D and scheduled with split(128), while the other tensors get merged then split. Probably all tensors need to be scheduling-broadcast to 3D in order to get the indexing right.

jacobhinkle: However, we don't hit this error with a single set+broadcast+add kernel, so it's probably just another bug.

jacobhinkle:
As discussed in today's meeting, I have removed `forwardAroundMissingAxes` from this PR and updated the tests to manually schedule each part of the example fusion separately. I removed the `set` that is causing an indexing issue in the test. Hopefully in the future, perhaps when #2459 or similar is introduced, it will be safe to revisit forwarding. Until then, this will still be useful for the matmul scheduling refactor in #2458

tests/cpp/test_abstract_tensor.cpp
naoyam: What does the final fusion math look like?

jacobhinkle:
```
Inputs:
  T0_g[ iblockIdx.x17{( ceilDiv(i0, 128) )}, ithreadIdx.x18{128} ], float
  T1_g[ iblockIdx.x28{( ceilDiv(( i2 * i3 ), 128) )}, ithreadIdx.x29{128} ], float
  T2_g[ iblockIdx.x43{( ceilDiv(( i4 * i5 ), 128) )}, ithreadIdx.x44{128} ], float
Outputs:
  T5_g[ iblockIdx.x34{( ceilDiv(( i0 * i3 ), 128) )}, ithreadIdx.x35{128} ] ca_pos( 2 ) produce_pos( 2 ), float
  T7_g[ iblockIdx.x49{( ceilDiv(( i4 * i0 ), 128) )}, ithreadIdx.x50{128} ] ca_pos( 2 ) produce_pos( 2 ), float

%kernel_math {
T3_l[ iblockIdx.x19{( ceilDiv(i0, 128) )}, ithreadIdx.x20{128} ] ca_pos( 2 )
   = Set( T0_g[ iblockIdx.x17{( ceilDiv(i0, 128) )}, ithreadIdx.x18{128} ], cache_op=Streaming )
T4_l[ iblockIdx.x31{( ceilDiv(( i0 * 1 ), 128) )}, ithreadIdx.x32{128} ] ca_pos( 2 ) produce_pos( 2 )
   = broadcast( T3_l[ iblockIdx.x19{( ceilDiv(i0, 128) )}, ithreadIdx.x20{128} ] ca_pos( 2 ) )
T5_g[ iblockIdx.x34{( ceilDiv(( i0 * i3 ), 128) )}, ithreadIdx.x35{128} ] ca_pos( 2 ) produce_pos( 2 )
   = T4_l[ iblockIdx.x31{( ceilDiv(( i0 * 1 ), 128) )}, ithreadIdx.x32{128} ] ca_pos( 2 ) produce_pos( 2 )
   + T1_g[ iblockIdx.x28{( ceilDiv(( i2 * i3 ), 128) )}, ithreadIdx.x29{128} ];
T6_l[ iblockIdx.x46{( ceilDiv(( 1 * i0 ), 128) )}, ithreadIdx.x47{128} ] ca_pos( 2 ) produce_pos( 2 )
   = broadcast( T3_l[ iblockIdx.x19{( ceilDiv(i0, 128) )}, ithreadIdx.x20{128} ] ca_pos( 2 ) )
T7_g[ iblockIdx.x49{( ceilDiv(( i4 * i0 ), 128) )}, ithreadIdx.x50{128} ] ca_pos( 2 ) produce_pos( 2 )
   = T6_l[ iblockIdx.x46{( ceilDiv(( 1 * i0 ), 128) )}, ithreadIdx.x47{128} ] ca_pos( 2 ) produce_pos( 2 )
   + T2_g[ iblockIdx.x43{( ceilDiv(( i4 * i5 ), 128) )}, ithreadIdx.x44{128} ];
}
```

naoyam: I'm actually surprised it works when the set isn't there. Looks like all tensors are inlined at position 2, but none of them should be concrete. 

Could you also show the math output without the set?

jacobhinkle: Sure. Without the set:
```
Inputs:
  T0_g[ iblockIdx.x16{( ceilDiv(i0, 128) )}, ithreadIdx.x17{128} ], float
  T1_g[ iblockIdx.x25{( ceilDiv(( i2 * i3 ), 128) )}, ithreadIdx.x26{128} ], float
  T2_g[ iblockIdx.x40{( ceilDiv(( i4 * i5 ), 128) )}, ithreadIdx.x41{128} ], float
Outputs:
  T4_g[ iblockIdx.x31{( ceilDiv(( i0 * i3 ), 128) )}, ithreadIdx.x32{128} ] ca_pos( 2 ) produce_pos( 2 ), float
  T6_g[ iblockIdx.x46{( ceilDiv(( i4 * i0 ), 128) )}, ithreadIdx.x47{128} ] ca_pos( 2 ) produce_pos( 2 ), float

%kernel_math {
T3_l[ iblockIdx.x28{( ceilDiv(( i0 * 1 ), 128) )}, ithreadIdx.x29{128} ] ca_pos( 2 )
   = broadcast( T0_g[ iblockIdx.x16{( ceilDiv(i0, 128) )}, ithreadIdx.x17{128} ] )
T4_g[ iblockIdx.x31{( ceilDiv(( i0 * i3 ), 128) )}, ithreadIdx.x32{128} ] ca_pos( 2 ) produce_pos( 2 )
   = T3_l[ iblockIdx.x28{( ceilDiv(( i0 * 1 ), 128) )}, ithreadIdx.x29{128} ] ca_pos( 2 )
   + T1_g[ iblockIdx.x25{( ceilDiv(( i2 * i3 ), 128) )}, ithreadIdx.x26{128} ];
T5_l[ iblockIdx.x43{( ceilDiv(( 1 * i0 ), 128) )}, ithreadIdx.x44{128} ] ca_pos( 2 )
   = broadcast( T0_g[ iblockIdx.x16{( ceilDiv(i0, 128) )}, ithreadIdx.x17{128} ] )
T6_g[ iblockIdx.x46{( ceilDiv(( i4 * i0 ), 128) )}, ithreadIdx.x47{128} ] ca_pos( 2 ) produce_pos( 2 )
   = T5_l[ iblockIdx.x43{( ceilDiv(( 1 * i0 ), 128) )}, ithreadIdx.x44{128} ] ca_pos( 2 )
   + T2_g[ iblockIdx.x40{( ceilDiv(( i4 * i5 ), 128) )}, ithreadIdx.x41{128} ];
}
```

naoyam: I don't understand how this actually works. There's no single domain that actually can work as a concrete domain, yet all tensors are fully inlined. 

Does compute-sanitizer show anything?

I'd also be interested in the generated kernel *without* parallelization and see what the outer loop looked like. That should tell us what domain is used as the concrete domain.

jacobhinkle: No errors in compute-sanitizer. Without parallelization, we have two separate loop nests:
```c++
__global__ void nvfuser_none_f0_c0_r0_g0(Tensor<float, 1, 1> T0, Tensor<float, 2, 2> T1, Tensor<float, 2, 2> T2, Tensor<float, 2, 2> T4, Tensor<float, 2, 2> T6) {
  NVFUSER_DEFINE_MAGIC_ZERO;
  nvfuser_index_t i0;
  i0 = T2.logical_size[0LL] * T0.logical_size[0LL];
  nvfuser_index_t i1;
  i1 = ceilDiv(i0, 128LL);
  nvfuser_index_t i2;
  i2 = T0.logical_size[0LL] * T1.logical_size[1LL];
  nvfuser_index_t i3;
  i3 = ceilDiv(i2, 128LL);
  nvfuser_index_t i4;
  i4 = T1.logical_size[1LL] * T0.logical_size[0LL];
  #pragma unroll 1
  for(nvfuser_index_t i5 = 0; i5 < i1; ++i5) {
    nvfuser_index_t i6;
    i6 = 128LL * i5;
    #pragma unroll
    for(nvfuser_index_t i7 = 0; i7 < 128LL; ++i7) {
      nvfuser_index_t i8;
      i8 = i6 + (i7 + nvfuser_zero);
      nvfuser_index_t i9;
      i9 = i8 % T0.logical_size[0LL];
      float T5[1LL];
      T5[0LL]
         = T0[(T0.alloc_stride[0LL] * i9)];
      if ((i8 < i0)) {
        T6[i8]
          = T5[0LL]
          + T2[((T2.alloc_stride[0LL] * (i8 / T0.logical_size[0LL])) + (T2.alloc_stride[1LL] * i9))];
      }
    }
    NVFUSER_UPDATE_MAGIC_ZERO;
  }
  #pragma unroll 1
  for(nvfuser_index_t i10 = 0; i10 < i3; ++i10) {
    nvfuser_index_t i11;
    i11 = 128LL * i10;
    #pragma unroll
    for(nvfuser_index_t i12 = 0; i12 < 128LL; ++i12) {
      nvfuser_index_t i13;
      i13 = i11 + (i12 + nvfuser_zero);
      nvfuser_index_t i14;
      i14 = i13 / T1.logical_size[1LL];
      float T3[1LL];
      if ((i13 < i4)) {
        T3[0LL]
           = T0[(T0.alloc_stride[0LL] * i14)];
      }
      if ((i13 < i2)) {
        T4[i13]
          = T3[0LL]
          + T1[((T1.alloc_stride[0LL] * i14) + (T1.alloc_stride[1LL] * (i13 % T1.logical_size[1LL])))];
      }
    }
    NVFUSER_UPDATE_MAGIC_ZERO;
  }
}
```

naoyam: Oh, I now see what's going on. Without the set, there are essentially two disjoint groups of tensors. The only common tensor is t0, but since it's an input, it doesn't matter for inlining.

With the set, however, an intermediate tensor, t3, is added, and that's another common tensor, and since it's not an input, inlining does matter for t3. Since t3 is used by both of the two groups of tensors, it effectively forces the two groups to be inlined together, and that's when the problem of missing concrete domains happens.

This pattern is similar to the fusion in this [test.](https://github.com/NVIDIA/Fuser/blob/main/tests/cpp/test_id_model.cpp#L1225). The fusion doesn't work with the current lowering. Specifically, an incorrect loop and indices would be generated. The loop promotion analysis should capture all the necessary information to fix the problem, but I haven't yet looked into the issue very closely.

I don't think the loop promotion analysis, at least at this point, would work with this fusion with the set. The analysis relies on the broadcast forwarding pattern to infer the correct concrete domain of each loop group, but in this case there's no broadcast.

I think it's possible to extend the loop promotion analysis, but I don't feel that's the right approach.

PR #2515, opened 2024-07-02
Add a theorem about the equivalence of split-split

zasdfgbnm:
This PR proves a theorem

This is needed for matmul scheduling. For example if for the K dimension, the CTA tile is 256, and you are using 64 byte swizzle for A, and 128 byte swizzle for B. Then the schedule of A, B, and the MMA result can be like:

image

or

image

both options has problem.

For option 1, the schedule of A and B matches with the specification in [Introduction to TMA Support in NVFuser](https://github.com/NVIDIA/Fuser/blob/main/doc/dev/tma.md), however, currently we are unable to inline dimension 0 because it does not have exactly the same transformation history.

Option 2 is the opposite: it is trivial to see that dimension 0 can be inlined, but it does not match with the specification in [Introduction to TMA Support in NVFuser](https://github.com/NVIDIA/Fuser/blob/main/doc/dev/tma.md), because according to the specification, this schedule means that we should first view the tensor from `(K,)` as `(K/256, 256)` and treat the `K/256` and `256` as two separate dimensions of TMA, which is not always possible.

The newly proved theorem tells us that, option 1 and option 2 are equivalent. So ideally, whether scheduling as option 1 or option 2 should make no difference. Both options should be capable of inlining, and both should be capable of using the TMA correctly.

In practice, it is not clear to me yet what is the shortest path to solve this problem, but I believe the first step before changing any code is to prove that they are indeed equivalent, so that we do not make any mistake.

In the long term, I believe `IdModel` should be mathematical-equivalence-aware, that is, mathematically equivalent IDs needs to be mapped together as much as possible in all mapping modes. Inlining should be using the mathematical-equivalence-aware IdModel to determine the maximum inlinable position. Our TMA analysis should be capable of applying mathematical-equivalence analysis to find the most-compatible TMA domain.


naoyam (approved):
LGTM. I think we could have another `ValGraph` that uses this information.

jacobhinkle:
This is cool. So inner splits are associative, meaning we could represent them with _multiple_ inner dimensions without loss of generality.

jacobhinkle:
So could we represent inner splits as having multiple ordered inner dimensions? The split op could then add dimensions as needed instead of creating multiple binary splits that need to be matched later.

PR #2516, opened 2024-07-02
Support indexing of unrolled domains

naoyam:
Unrolling affects allocations, so indexing also needs to be aware of unrolled domains. More specifically, the allocation position is the minimum of the CA position and unroll domain. 

Furthermore, for certain tensors, just looking at its loop domains may not be sufficient if the tensor is broadcast and inlined. For example,

```
 auto tv0 = makeContigConcreteTensor({4});
  fusion.addInput(tv0);
  auto tv1 = makeContigConcreteTensor({3, 4});
  fusion.addInput(tv1);

  auto tv2 = set(tv0);
  auto tv3 = broadcast(tv2, {true, false});
  auto tv4 = add(tv3, tv1);
  fusion.addOutput(tv4);

  tv4->split(0, 1);

  TransformPropagator propagator(tv4);
  MaxRootDomainInfoSpanningTree(tv4).traverse(&propagator);

  inlineMost();

  tv4->axis(1)->parallelize(ParallelType::Unroll);
```

```
Inputs:
  T0_g[ iS0{4} ], float
  T1_g[ iS10{( ceilDiv(3, 1) )}, iS11{1}, iS2{4} ], float
Outputs:
  T4_g[ iS8{( ceilDiv(3, 1) )}, iUR9{1}, iS7{4} ] ca_pos( 3 ) produce_pos( 3 ), float

%kernel_math {
T2_l[ iS3{4} ] ca_pos( 1 )
   = Set( T0_g[ iS0{4} ], cache_op=Streaming )
T3_l[ bS12{( ceilDiv(1, 1) )}, bUR13{1}, iS5{4} ] ca_pos( 3 ) produce_pos( 3 )
   = broadcast( T2_l[ iS3{4} ] ca_pos( 1 ) )
T4_g[ iS8{( ceilDiv(3, 1) )}, iUR9{1}, iS7{4} ] ca_pos( 3 ) produce_pos( 3 )
   = T3_l[ bS12{( ceilDiv(1, 1) )}, bUR13{1}, iS5{4} ] ca_pos( 3 ) produce_pos( 3 )
   + T1_g[ iS10{( ceilDiv(3, 1) )}, iS11{1}, iS2{4} ];
}
```

See `InlinedUnroll` for the full detail.

In this case, `tv2` is fully inlined into `tv3` and then `tv4`. Notice that while `tv4` is unrolled, but just looking at `tv2` doesn't allow you to find the unrolling. In other words, the loop domains of `tv2` fail to represent the unrolling.

I think this is yet another instance where we may want to do something like [this](#2459). If we inserted a broadcast domain to `tv2` when it's inlined, that broadcast domain would capture the unrolling.

While I do think broadcast insertion is a better approach, for now, I'm fixing it by checking the for-loops where a tensor is defined. The logic of finding the correct allocation position considering inlined unrolled domains is already [there](https://github.com/NVIDIA/Fuser/blob/main/csrc/device_lower/utils.h#L285), so I'm just using it for now.

Additionally, I moved `getAllocationDomains` to a separate class and changed the overall flow such that before indexing, traversal of KIR exprs must be done to build a map from each tensor to its allocation information. Previously, the allocation information is collected on the fly, but it didn't work well now that for-loops at the time of definitions are required.


naoyam:
!build

csrc/id_model/indexing.cpp
```diff
@@ -97,128 +98,214 @@ Val* getStrideOfGlobalMemoryTensor(TensorView* tv, int64_t alloc_dim) {
...
+      // - Global: logical domains
+      // - Local/Shared: loop domains to the right of the CA position
+      if (tv->getMemoryType() == MemoryType::Global) {
+        allocation_domains = tv->getLogicalDomain();
+        contiguity = tv->domain()->contiguity();
+      } else {
+        // Allocation position is not always the same as the CA
+        // position. See also lower_utils::getAllocInformation.
+        int64_t allocation_pos =
+            lower_utils::getAllocInformation(tv, for_loops).alloc_pos;
```
naoyam, line 196: There should be no logic change except for this.

naoyam:
CC: @zasdfgbnm

naoyam:
@jacobhinkle Let me know if you can review this PR.

csrc/id_model/indexing.cpp
```diff
@@ -97,128 +98,214 @@ Val* getStrideOfGlobalMemoryTensor(TensorView* tv, int64_t alloc_dim) {
...
-    actual_allocation_domains.push_back(promotion_domain);
-    actual_strides.push_back(stride);
+
+    return IndexingAllocationInfo{actual_allocation_domains, actual_strides};
   }
 
-  return {actual_allocation_domains, actual_strides};
-}
+  std::unordered_map<TensorView*, IndexingAllocationInfo> tv_alloc_info_map;
+  std::unordered_set<TensorView*> used_as_producer;
```
jacobhinkle, line 307: Nit: our convention of appending underscore for private members.
```suggestion
  std::unordered_map<TensorView*, IndexingAllocationInfo> tv_alloc_info_map_;
  std::unordered_set<TensorView*> used_as_producer_;
```

csrc/id_model/indexing.cpp
```diff
@@ -97,128 +98,214 @@ Val* getStrideOfGlobalMemoryTensor(TensorView* tv, int64_t alloc_dim) {
...
-    IterDomain* promotion_domain = nullptr;
-    if (is_loop) {
-      promotion_domain = getLoopPromotion(allocation_domain, id_model);
+  void dispatch(Expr* expr) override {
+    if (ir_utils::isTvOp(expr)) {
+      for (auto out_tv : ir_utils::filterByType<TensorView>(expr->outputs())) {
+        // Note that since we are dealing with a Kernel IR, a single
+        // tensor may show up as consumers multiple times, e.g.,
+        // zero initialization and actual definition. Using the last
+        // expr should give us correct allocation info.
```
jacobhinkle, line 154: Out of curiosity: is there anything special about the last expr? It seems like they would all have the same allocation info.

jacobhinkle (approved):
LGTM. Just minor comments.

csrc/id_model/indexing.cpp
naoyam: They are actually not private. This class was meant to be a relatively simple class, although it isn't anymore.

csrc/id_model/indexing.cpp
naoyam: Here's the lowering result from the `InlinedUnroll` test:

```
FOR i70 in iS8{( ceilDiv(3, 1) )}:
  T2_l[ iS3{4} ] ca_pos( 1 ) = ALLOCATE(buffer=T2_l[ iS3{4} ] ca_pos( 1 ), mem_type=register, size=4, zero_init=false, resets_to_zero=false)
  FOR i69 in iS7{4}:
    IF Inline true:
      T2_l[i69] view( T2 )
         = Set( 0, cache_op=Streaming )
  T3_l[ bS12{( ceilDiv(1, 1) )}, bUR13{1}, iS5{4} ] ca_pos( 3 ) produce_pos( 3 ) = ALLOCATE(buffer=T3_l[ bS12{( ceilDiv(1, 1) )}, bUR13{1}, iS5{4} ] ca_pos( 3 ) produce_pos( 3 ), mem_type=register, size=4, zero_init=false, resets_to_zero=false)
  IF Unswitch ( ( i70 >= 0 ) && ( i70 < 3 ) ):
    FOR i71 in iUR9{1}:
      FOR i69 in iS7{4}:
        IF Manual true:
          T2_l[i69] view( T2 )
             = Set( T0_g[i69] view( T0 ), cache_op=Streaming )
        IF Manual true:
          T3_l[i69] view( T3 )
             = broadcast( T2_l[i69] view( T2 ) )
        IF Manual true:
          T4_g[( ( i70 * 4 ) + i69 )] view( T4 )
             = T3_l[i69] view( T3 )
             + T1_g[( ( i70 * 4 ) + i69 )] view( T1 );
  ELSE:
    FOR i71 in iUR9{1}:
      FOR i69 in iS7{4}:
        IF Inline ( ( i69 >= 0 ) && ( i69 < 4 ) ):
          T2_l[i69] view( T2 )
             = Set( T0_g[i69] view( T0 ), cache_op=Streaming )
        IF Inline true:
          T3_l[i69] view( T3 )
             = broadcast( T2_l[i69] view( T2 ) )
        IF Inline ( ( ( ( i70 >= 0 ) && ( i70 < 3 ) ) && ( i69 >= 0 ) ) && ( i69 < 4 ) ):
          T4_g[( ( i70 * 4 ) + i69 )] view( T4 )
             = T3_l[i69] view( T3 )
             + T1_g[( ( i70 * 4 ) + i69 )] view( T1 );
```

As you can see, there are three assignments to `T2`. The first one is zero initialization and that has a different set of for-loops than the other two. If we used the loop-nest at the time of the initialization, we would get different allocation iter domains because there's no unrolled loop there.

naoyam: Added more comments

PR #2521, opened 2024-07-03
[Python Frontend] Add Jupyter notebook of creating a user scheduled fusion.

rdspring1:
This PR adds a Jupyter notebook tutorial for creating user-schedules in python frontend. It covers defining, scheduling, and profiling a fusion.


rdspring1:
It is easier to review the jupyter notebook through this link. https://github.com/NVIDIA/Fuser/blob/user_demo/doc/dev/python_scheduling/interactive_scheduling_layer_norm.ipynb

doc/dev/python_scheduling/interactive_scheduling_layer_norm.py
```diff
@@ -0,0 +1,190 @@
...
+class LayerNorm(FusionDefinition):
+    def definition(self):
+        self.t0 = self.from_pytorch(inputs[0])
+        self.s0 = self.define_scalar(1e-6, dtype=DataType.Double)
+        self.norm_const = self.define_scalar(tensor_size, dtype=DataType.Int)
+
+        self.mean_cast = self.ops.cast(self.t0, dtype=DataType.Float)
+        self.sum0 = self.ops.sum(self.mean_cast, dims=[-1])
+        # NOTE Manually broadcast because fusion definition cannot access hidden reduction tensor view.
+        self.bcast_sum0 = self.ops.broadcast(self.sum0, [False, True])
```
jjsjann123, line 40: nitpick, doesn't look like we are using bcast_sum0 in scheduling. Does the comment above still hold?

doc/dev/python_scheduling/interactive_scheduling_layer_norm.py
```diff
@@ -0,0 +1,190 @@
...
+        self.mean = self.ops.div(self.bcast_sum0, self.norm_const)
+
+        self.var_cast = self.ops.cast(self.t0, dtype=DataType.Float)
+        self.diff = self.ops.sub(self.var_cast, self.mean)
+        self.diff_sq = self.ops.mul(self.diff, self.diff)
+        self.sum1 = self.ops.sum(self.diff_sq, dims=[-1])
+        # NOTE Manually broadcast because fusion definition cannot access hidden reduction tensor view.
```
jjsjann123, line 47: ditto

doc/dev/python_scheduling/interactive_scheduling_layer_norm.py
```diff
@@ -0,0 +1,190 @@
...
+# In[5]:
+
+
+# Create user schedule for this input
+# NOTE: Schedules defined by the user for specific input sizes.
+fn._setup_schedule(inputs)
+print(fn._user_schedule_ir())
+
+
+# In[6]:
```
jjsjann123, line 79: Is this how we want to expose our manual schedule in a tutorial?

The existing pattern is to have schedule in the derived class's `schedule` function. i.e. line 80-136 should be moved there.

jjsjann123: ^^^ This is meant as a question rather than a suggestion. i.e. what's the reason that we are promoting a different pattern here? Does that mean we should move towards a different approach for scheduling?

cc'ing @kevinstephano who might have some comment on this.

jjsjann123 (commented):
sorry this falls off my radar earlier.

doc/dev/python_scheduling/interactive_scheduling_layer_norm.py
rdspring1: We don't use `bcast_sum0`  in scheduling, but we use `rfactor` on the reduction `self.sum0`.  `sum` with `keepdim=True` returns `bcast_sum0`.

doc/dev/python_scheduling/interactive_scheduling_layer_norm.py
rdspring1: We need to rfactor `self.sum1`

doc/dev/python_scheduling/interactive_scheduling_layer_norm.py
rdspring1: I imagined using a Jupyter notebook for interactive scheduling. The standard pattern forces you to rerun all cells if you want to make a change.

doc/dev/python_scheduling/interactive_scheduling_layer_norm.py
jjsjann123: Got'ya. I didn't realize this is impacting `fn.sched.tensors())`.

This feels restrictive... i.e. should we try to lift such restrictions in the future? Would we be able to pull of an `fn.sched.all_tvs()` that can iterate through live tvs and add missing entries to fusion_state?

doc/dev/python_scheduling/interactive_scheduling_layer_norm.py
jjsjann123: keep nitpicking on this one. I would argue it probably makes sense to have a `proper` tutorial showing our intended workflow with a single `schedule()` function. :stuck_out_tongue_winking_eye: 

But we don't have to do that in this PR.

jjsjann123 (approved):
Thanks for adding the example.

doc/dev/python_scheduling/interactive_scheduling_layer_norm.py
rdspring1: Probably something to handle when I add support for concretized Fusion.

doc/dev/python_scheduling/interactive_scheduling_layer_norm.py
rdspring1: I created `tutorial_scheduling_layer_norm.py` which uses the `schedule()` function.

PR #2522, opened 2024-07-03
Reorder allocation domains to align with logical domains

naoyam:
I put why this needs to be done [here](https://github.com/NVIDIA/Fuser/pull/2522/files#diff-0c47d4d562dba268498573ced9dd7fc7d9638b5882aca4c4cc20e70a7ccec891R315). For a concrete example, see the added unit test.

Alternatively, we could make loop vectorized domains always placed at the rightmost position. Since the indexing scheme uses the loop domains to infer the allocation domain, that would solve the issue as well. For now, I'm adding this patch so that the new indexer can work with the existing schedulers with no change.

Stacked on top of #2516


naoyam:
CC: @zasdfgbnm

csrc/id_model/indexing.cpp
```diff
@@ -303,6 +312,112 @@ class AllocationDomainSetup : private kir::IrVisitor {
     return IndexingAllocationInfo{actual_allocation_domains, actual_strides};
   }
 
+  // Reorder non-logical allocation domains to follow the ordering of
+  // the logical domain. This is necessary when an allocation domain
+  // includes a vectorized loop iter domain since it must be at the
+  // innermost position but that may not be the case in the loop
+  // domain. Not strictly necessary otherwise, but this should also
```
jacobhinkle, line 319: If a non-innermost loop domain is vectorized is that treated as innermost for inlining? That is, it cannot inline further than the vectorized domain right? I wonder in what cases we would want to have a vectorized loop iterdomain that is not inner-most.

csrc/id_model/indexing.cpp
```diff
@@ -303,6 +312,112 @@ class AllocationDomainSetup : private kir::IrVisitor {
...
+      const std::vector<IterDomain*>& allocation_domains) const {
+    // Don't change the set allocation domain. Ignore allocation
+    // domains of non-global tensors. For example,
+    // AliasTest.NotAllOutputAlias_Reduction has a tensor, tv6, that
+    // is a Local tensor with CA position of 4 but has an allocation
+    // domain that's just a permutation of its logical domain. Such
+    // invalid allocations need to be ignored.
+    if (tv->hasAllocation() && tv->getMemoryType() == MemoryType::Global) {
+      return std::nullopt;
+    }
```
jacobhinkle, line 335: > Ignore allocation domains of non-global tensors.

Is this because global stores must always already have their vectorization axes innermost, or is there another reason?

csrc/id_model/indexing.cpp
naoyam: > it cannot inline further than the vectorized domain right?

Right, I forgot that effect to inlining. So, if a non-innermost loop domain is vectorized, it's still vectorized properly, and the maximum inlining position is limited by the position of the vectorized domain. If we changed the scheduler to always place vectorized domains at the innermost loop domain position, it would also change the inlining positions. I'm not sure if that would have any performance implication in practice.

> I wonder in what cases we would want to have a vectorized loop iterdomain that is not inner-most.

I don't know. Maybe that's just for simplicity of scheduling?

csrc/id_model/indexing.cpp
naoyam: What we are doing here is inferring the right allocation domain of a given tensor. If a tensor already has a set allocation domain, I thought no inference should be done and just the set allocation domain should be used. So, originally, I didn't have `tv->getMemoryType() == MemoryType::Global`, but whenever a tensor has an allocation domain, no reordering should be done.

However, sometimes I found Local/Shared tensors also have allocation domains, and some of which don't seem to make sense, so this condition is just to ignore set allocation domains for non-global tensors.

This is an inference that I'm trying to match what the existing indexing implicitly assumes. It's not ideal because there's some ambiguity. There's also another "patch" that I'm going to create a PR for, but ideally, I think we should set the correct allocation domain for all tensors more explicitly. 

I consider this problem of figuring out the right allocation domain is a separate issue from indexing. I think we should refactor the allocation lowering pass such that each tensor gets the right allocation domain. These patches are not ideal, but for the time being, they should get us feature parity with the existing indexer.

naoyam:
!build

naoyam:
Pinging @jacobhinkle

jacobhinkle (approved):
LGTM. Thanks for the explanations.

PR #2523, opened 2024-07-03
Just change the ordering of multiplications when linearizing indices

naoyam:
Just mechanical changes to follow the ordering in the current indexing.

#2522


jacobhinkle (approved):

naoyam:
Marked it as draft to avoid merge.

PR #2524, opened 2024-07-03
Update pairwise mapping for `SDPAFwdOp` outputs

Priya2698:
This PR:
- modifies the variables `cum_seq_q/k` to have root_domain as `{iN}` and logical domain as `{iN+1}`.
- updates the tests to validate all inputs from the forward pass. This excludes the scalar outputs `query_seq_len/key_seq_len` which will added in the follow-up PR.

### DIFF
--- a/csrc/ops/composite.cpp
+++ b/csrc/ops/composite.cpp
@@ -495,8 +495,11 @@ SdpfaFwdResult sdpfa_fwd(
   // Scalar tensors of int64_t dtype.
   TensorView* philox_seed = TensorViewBuilder().dtype(DataType::Int).build();
   TensorView* philox_offset = TensorViewBuilder().dtype(DataType::Int).build();
+
+  // Thunder metadata represents debug_attn_mask of type int64_t, although the debug_attn_mask is of query.dtype.
+  // Since we use return_debug_mask=false in the internal flash attention call, this is a scalar zero tensor.
   TensorView* debug_attn_mask =
-      TensorViewBuilder().dtype(DataType::Int).build();
+      TensorViewBuilder().dtype(query->dtype()).build();
 
   // Set default values for dropout_p (0.0), is_causal(false)
   if (dropout_p == nullptr) {
--- a/csrc/ops/composite.cpp
+++ b/csrc/ops/composite.cpp
@@ -471,11 +471,12 @@ SdpfaFwdResult sdpfa_fwd(
 
   // Create a new Tensorview for cum_seq_q, cum_seq_k of shape (N + 1)
   auto newForCumulativeSeq = [&]() -> TensorView* {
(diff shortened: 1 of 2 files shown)


Priya2698:
!build

csrc/ops/composite.cpp
```diff
@@ -471,16 +471,20 @@ SdpfaFwdResult sdpfa_fwd(
 
   // Create a new Tensorview for cum_seq_q, cum_seq_k of shape (N + 1)
   auto newForCumulativeSeq = [&]() -> TensorView* {
-    IterDomain* batch_id = ops::newOutputIterDomain({query_domain.front()});
-    batch_id = IterDomain::resize(
+    IterDomain* batch_id = ops::newOutputIterDomain({query_domain.front(), key_domain.front(), value_domain.front()});
```
naoyam, line 474: Why is this change?

`key_domain.front(), value_domain.front()}`

Priya2698: `ops::newOutputIterDomain` can use a series of iterdomains to construct a new iterdomain. If any of these are non-symbolic, the extents are set using that iterdomain. This is useful in resolving broadcast iterdomains.
In practice, this maybe unlikely so using just `query_domain.front()` is sufficient too.

This is similar to the construction of attention output and log_sumexp in L446-447, L461-462 which use the iterdomains corresponding to `N` from all inputs (q/k/v).

This was missing for `cum_seq_q/k`, so updating now.

naoyam (approved):
LGTM

tests/cpp/test_sdpa_node.cpp
```diff
@@ -188,44 +223,54 @@ TEST_F(SDPATest, PairwiseRootDomainMap) {
...
-  fusion->addOutput(tvattn.output);
-
-  // Verify mapping between Q,K,V and attention output
+  addSdpaFwdOutputs(fusion.get(), output);
+
+  // Verify mapping between Q,K,V and all output
+  // Producers:
+  //   query = [N, H, L, E]
+  //   key = [N, H, S, E]
+  //   value = [N, H, S, Ev]
```
jjsjann123, line 239: nitpick: Looks like we support broadcasting on the batch dimension from the code changes above? do we actually have a test case for that where one of the entries have a broadcast batch dimension?

jjsjann123 (approved):

tests/cpp/test_sdpa_node.cpp
Priya2698: Yes, broadcast is allowed on N, H.
I have a separate branch adding some checks enforcing `E = Ev` and the corresponding mapping, I will include it there.

Thanks for bringing this up.

jjsjann123: Feel free to add that in a separate PR if that works better~

Priya2698:
!build

PR #2526, opened 2024-07-03
Check for presence of ExprEval-only nodes early in canSchedule to avoid segmentation errors.

### LINKED ISSUE #2517: SDPA node segmentation error
Short example of SDPA node with expressions before and after that trigger in error building computeAtMap (https://github.com/NVIDIA/Fuser/blob/031dc0bdfa0488adfc9a4e47a69c7dd250d8d9a8/csrc/compute_at_map.cpp#L387).

To generate the error comment out the changes in registry.cpp and the segment_set in the test added here: https://github.com/NVIDIA/Fuser/compare/main...cowanmeg:Fuser:sdpa?expand=1

Even with segment_sets surrounding the SDPA inputs and outputs, the segmenter tries to find a scheduler for SPDA + SegmenterSet (example below) which triggers the scheduler's `checkCanSchedule` to error out while building the computeAt map.  The little hack I put in place in registry.cpp skips over any Fusion with SDPA + other expressions (technically we don't require segment sets with this hack...)

```
%kernel {
T8_g[ iS32{16}, iS33{32}, iS34{64}, iS35{64} ]
   = sdpa(T7_g[ iS28{16}, iS29{32}, iS30{64}, iS31{64} ],
            T1_g[ iS4{16}, iS5{32}, iS6{128}, iS7{64} ],
            T2_g[ iS8{16}, iS9{32}, iS10{128}, iS11{64} ],
            dropout_p = double(0),
            is_causal = false  )
T15_g[ iS43{16}, iS44{32}, iS45{64}, iS46{64} ]
   = SegmenterSet( T8_g[ iS32{16}, iS33{32}, iS34{64}, iS35{64} ] )
```

cc @Priya2698

Priya2698:
Resolves #2517.
`SdpaFwdOp` is not compatiable with computeAt map since the outputs of the op do not all have the same root domain.
For larger fusions containing `SdpaFwdOp`, `checkCanSchedule` runs into error due to this.

### DIFF
--- a/csrc/scheduler/registry.cpp
+++ b/csrc/scheduler/registry.cpp
@@ -170,17 +170,27 @@ bool checkCanSchedule(
     Fusion* fusion,
     SchedulerRuntimeInfo& runtime_info,
     HeuristicSummary* data_cache = nullptr) {
-
-  // ExprEval scheduler only requires `canScheduleCompileTime` check and should not use this fn.
-  // The following checks build the computeAt map that do not work with SDPAOp.
+  // ExprEval scheduler only requires `canScheduleCompileTime` check and should
+  // not use this fn. The following checks build the computeAt map that do not
+  // work with SDPAOp.
   NVF_ERROR(SchedulerType::heuristicType() != ScheduleHeuristic::ExprEval);
-  
+
   FusionGuard fg(fusion);
 
-  // Fusions with `MatmulOp/LinearOp/SdpaOp` are only accepted in `ExprEval` scheduler, all other schedulers should reject them.
-  if (ir_utils::hasAnyMatmulOps(fusion)) {
+  // Fusions with `MatmulOp/LinearOp/SdpaOp` are only accepted in `ExprEval`
+  // scheduler, all other schedulers should reject them.
+  if (ir_utils::hasOpsOfType<MatmulOp, LinearOp, SdpaFwdOp>(fusion)) {
     scheduler_debug_utils::canScheduleRejectReason(
-        SchedulerType::heuristicType(), "Matmul ops are not supported");
+        SchedulerType::heuristicType(),
+        "ExprEval-only nodes (MatmulOp/LinearOp/SdpaOps) are not supported.");
+    return false;
+  }
+
+  // Fusions with `MmaOp` are only accepted by Matmul scheduler.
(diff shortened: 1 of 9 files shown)


Priya2698:
!build

csrc/scheduler/registry.cpp
```diff
@@ -170,7 +170,20 @@ bool checkCanSchedule(
...
     SchedulerRuntimeInfo& runtime_info,
     HeuristicSummary* data_cache = nullptr) {
+
+  // ExprEval scheduler only requires `canScheduleCompileTime` check and should not use this fn.
+  // The following checks build the computeAt map that do not work with SDPAOp.
+  NVF_ERROR(SchedulerType::heuristicType() != ScheduleHeuristic::ExprEval);
+  
   FusionGuard fg(fusion);
+
+  // Fusions with `MatmulOp/LinearOp/SdpaOp` are only accepted in `ExprEval` scheduler, all other schedulers should reject them.
```
jacobhinkle, line 180: What about matmul scheduler?

Priya2698: You're right, I missed this function has `MmaOp` as well.
Let me correct it, thanks for catching this.

Priya2698: I will separate these _ExprEval_ ops from `MmaOp` and check them separately.

Priya2698:
!build

cowanmeg (approved):
LGTM! Thanks for getting a fix out so quick!

Priya2698:
!build

csrc/scheduler/registry.cpp
Priya2698: Yes, the updated changes include that.

tests/cpp/test_sdpa_node.cpp
```diff
@@ -495,4 +495,56 @@ TEST_F(SDPATest, NonCausalAttnSymbolicBwd) {
       __LINE__,
       __FILE__);
 }
+
+TEST_F(SDPATest, AttnProgram) {
```
jacobhinkle, line 499: Can you link to #2517 in a comment here?

jacobhinkle (approved):
LGTM! This is a better approach than what I did which involved changing every other scheduler's `canScheduleCompileTime` manually :-D.

PR #2528, opened 2024-07-03
TMA doc: Define box by rotation

zasdfgbnm:
Add a new way of defining box: `Define box by rotation`, and organize it together with the `Define box by compositing` as `Define box by mathematical equivalence`.

For the motivation of adding this new way, see the description of: https://github.com/NVIDIA/Fuser/pull/2515


zasdfgbnm:
TODO: in a separate PR, rename all the "root domain" as "logical domain".

rdspring1:
Do we prefer option 2 from #2528 because we can inline dimension 0 and use "Define box by rotation" to get the TMA descriptor?

zasdfgbnm:
> Do we prefer option 2 from #2528 because we can inline dimension 0 and use "Define box by rotation" to get the TMA descriptor?

Yes, I believe this is the quickest solution in short term. In the long term, I think both option 1 and option 2 should work.

naoyam:
> For the motivation of adding this new way, see the description of: #2515

I get this is possible, but still it's unclear to me when it would be useful. Do you have a specific example using the rotation method?

zasdfgbnm:
> > For the motivation of adding this new way, see the description of: #2515
> 
> I get this is possible, but still it's unclear to me when it would be useful. Do you have a specific example using the rotation method?

In the following example:

image

for both A and B, the box are defined by rotation.

naoyam:
Oh, I see. Makes sense.

naoyam (approved):
LGTM

Issue #670, opened 2023-08-01
Build out `Expr::evaluate` for operators, and clean up unit tests


zasdfgbnm:
During the nvFuser development process, one of our routine tasks is to create a fusion and then compare the output of our codegen with the ATen version of the fusion. Manually writing an ATen implementation can be both challenging and time-consuming. As an example, the test provided below requires 23 lines of carefully constructed ATen code:

```C++
// extracted from NVFuserTest.FusionLayerNormFusedOpsRedundantCast_CUDA
TEST_F(NVFuserTest, ExampleTest) {
  std::unique_ptr<Fusion> fusion_ptr = std::make_unique<Fusion>();
  auto fusion = fusion_ptr.get();
  FusionGuard fg(fusion);

  const float kEps = 1e-5;
  const int batch_size = 2048 * 8;
  const int hidden_size = 20480;
  {
    DataType dtype = DataType::Half;
    auto tv0 = makeContigTensor(1, dtype);
    auto tv1 = makeContigTensor(2, dtype);
    auto tv2 = makeContigTensor(1, dtype);
    auto tv3 = makeContigTensor(1, dtype);
    auto tv4 = makeContigTensor(1, dtype);

    fusion->addInput(tv0);
    fusion->addInput(tv1);
    fusion->addInput(tv2);
    fusion->addInput(tv3);
    fusion->addInput(tv4);
    auto tv5 = broadcast(tv0, {true, false});
    auto tv6 = castOp(DataType::Float, tv1);
    auto tv7 = castOp(DataType::Float, tv5);
    auto tv8 = add(tv6, tv7);
    auto tv9 = castOp(DataType::Half, tv8);
    auto tv10 = broadcast(tv2, {true, false});
    auto tv11 = castOp(DataType::Float, tv9);
    auto tv12 = castOp(DataType::Float, tv10);
    auto tv13 = add(tv11, tv12);
    auto tv14 = castOp(DataType::Half, tv13);
    auto tv15 = castOp(DataType::Float, tv14);
    auto tv16 = variance(tv15, {1}, false, false);
    auto tv17 = broadcast(tv16, {false, true});
    auto tv18 = sum(tv15, {1}, false);
    auto tv19 = broadcast(tv18, {false, true});

    nvfuser::Val* num_features = IrBuilder::create<Val>(1.0);
    num_features = mul(num_features, tv0->getLeafDomain()[0]->extent());
    auto s20 = num_features;

    auto s21 = reciprocal(s20);
[...]

PR #2530, opened 2024-07-04
Adding Signbit & Fmod in ExprEval

jjsjann123:
Signbit & Fmod is used to produce scalar value to reshape, which requires ExpressionEvaluator to run it with scalar value.

Note: Also added the rule in NaiveValueMachine used by PrecomputedValues. But currently those are not being tests.


csrc/ir/nodes.cpp
```diff
@@ -464,6 +464,17 @@ std::vector<PolymorphicValue> UnaryOp::evaluate(
     case UnaryOpType::Sin:
       return {in.as<at::Tensor>().sin()};
       break;
+    case UnaryOpType::Signbit:
```
jjsjann123, line 467: Question for @Priya2698 .

Are we supposed to not have the full coverage on operations from expr eval ?

tests/python/test_python_frontend.py
```diff
@@ -4344,5 +4344,38 @@ def fusion_func(fd: FusionDefinition):
...
+            S10 = fd.define_scalar(0, dtype=DataType.Int)
+            S11 = fd.ops.ne(S9, S10)
+            S12 = fd.ops.bitwise_and(S6, S11)
+            S13 = fd.define_scalar(192, dtype=DataType.Int)
+            S14 = fd.ops.reciprocal(S3)
+            S15 = fd.ops.mul(S13, S14)
+            S16 = fd.ops.cast(S12, dtype=DataType.Int)
+            S17 = fd.ops.sub(S15, S16)
+            V18 = fd.define_vector([S0, S17], dtype=DataType.Int)
+            T19 = fd.ops.reshape(T1, new_shape=V18)
```
jjsjann123, line 4373: `reshape` concretization here calls evaluate on the chain of operations that produces V18.

signbit/fmod are the two that is missing from what we support, hence I needed to add those. My naive question is, what's the criteria for adding something to what expr eval?

csrc/ir/nodes.cpp
Priya2698: We do intend on having full coverage.
When resolving Issue #670, we targeted developing for all operators seen in test cases. As we add more test cases, and missing operators are seen, we should implement evaluate methods for them.

Feel free to open and assign these issues to me, to address missing operators.

tests/python/test_python_frontend.py
jacobhinkle: I think we'd like to have evaluate implemented for everything. Note that there is also `NaiveValueMachine::runBinaryOp` and friends in `evaluator_common.cpp`.

jjsjann123: Looks like I can just add an entry there. But I'm wondering where can I add a test for the entry in `NaiveValueMachine::runBinaryOp`?

jacobhinkle: Good question. If a `PrecomputedValues` is bound to an `ExpressionEvaluator` then you can run `expr_eval.precomputedValues()->evaluate()` as is done here: https://github.com/NVIDIA/Fuser/blob/65799ed56f66b671dfbd51336fa302400e2bdf45/csrc/executor.cpp#L1210. That will run `PrecomputedValues::run()`. It looks like our tests don't actually exercise this very much; the `TernaryOp` test tests that a value can be searched in `PrecomputedValues` but I don't believe that tests the `PrecomputedValues::run()` method: https://github.com/NVIDIA/Fuser/blob/65799ed56f66b671dfbd51336fa302400e2bdf45/tests/cpp/test_evaluator.cpp#L559-L560. That could be added in another PR I think.

jjsjann123: I'm going to leave `adding tests for PrecomputedValues` in another PR.

But looks like it makes sense to keep the implementation consistent between the two. So I'll at least add r signbit & fmod and cpp tests for that.

Thanks for pointing this one out. :bow:

jjsjann123: I just realized that `PrecomputedValues` are only used for ID extent() here: https://github.com/NVIDIA/Fuser/blob/6c6f3a40e09e6f8bece80b9b79c543945846c71b/csrc/evaluator_common.cpp#L93-L117

I don't see an easy way to get that with `Signbit` or `fmod`. Would you mind if we leave the testing part out? :stuck_out_tongue_closed_eyes:

jjsjann123:
!build

csrc/ir/nodes.cpp
jjsjann123: I'm not trying to push this out as high priority. I think it's reasonably fine to patch it as issues surface.

jjsjann123:
The signal I'm getting is that this is a step towards the right direction. Marked this PR as ready for review.

Priya2698 (approved):
LGTM. 
Just a minor question about coverage of float type.

csrc/polymorphic_value.h
```diff
@@ -279,6 +279,62 @@ inline bool isSame(const PolymorphicValue& a, const PolymorphicValue& b) {
   return a == b;
 }
 
+inline PolymorphicValue signbit(const PolymorphicValue& a) {
+  if (a.is<int64_t>()) {
+    return PolymorphicValue(std::signbit(a.as<int64_t>()));
+  }
+  if (a.is<double>()) {
+    return PolymorphicValue(std::signbit(a.as<double>()));
+  }
```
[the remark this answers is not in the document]
jjsjann123: IIUC, we don't have `float` for scalars. Same as `int` vs `int64_t`.

jjsjann123:
tests clears despite some numerical issue. I'll merge after CI clear.

jjsjann123:
!build

jjsjann123:
!build

jjsjann123:
This one is low risk and it has previously passed CI earlier. I'm merging it as-is.

PR #2531, opened 2024-07-04
Return scalars as CPU scalar tensors in `SdpaFwdOp`

Priya2698:
Issue #2278 
nvFuser only allows for TensorView fusion outputs.

`SdpaFwdOp` returns two scalar outputs -- `query_seq_len` and `key_seq_len`. They are returned as CPU scalar tensors of Long type to allow adding them as fusion outputs.
`SdpaBwdOp` is modified to accept these variables as scalar tensors for parity between the two ops. Coverting the scalar tensors to integers for use in backward pass is done within `SdpaBwdOp` node.


csrc/ir/nodes.cpp
```diff
@@ -4378,15 +4378,14 @@ std::vector<PolymorphicValue> SdpaFwdOp::evaluate(
...
-  // Polymorphic Value
+  // Query and key seq len are of type c10::SymInt -> convert them to CPU scalar tensors to support adding them as fusion outputs.
   return {
       output,
       log_sumexp,
       cum_seq_q,
       cum_seq_k,
-      *query_seq_len.maybe_as_int(),
-      *key_seq_len.maybe_as_int(),
+      at::scalar_tensor(*query_seq_len.maybe_as_int(), at::dtype(at::kLong)),
```
Priya2698, line 4387: `at::tensor` creates a 1D tensor, so using `scalar_tensor`.

tests/cpp/test_sdpa_node.cpp
```diff
@@ -610,4 +632,115 @@ TEST_F(SDPATest, AttnProgram) {
...
+        log_sumexp,
+        cum_seq_q,
+        cum_seq_k,
+        /*max_q=*/*query_seq_len.maybe_as_int(),
+        /*max_k=*/*key_seq_len.maybe_as_int(),
+        /*dropout_p=*/0.0,
+        /*is_causal=*/false,
+        philox_seed,
+        philox_offset,
+        /*scale=*/scale);
```
Priya2698, line 736: I am not able to validate the results against using `at::scaled_dot_product_attention` and `out.backward()` instead of the separate calls. Will look into it.

Priya2698:
@jjsjann123 @wujingyue: Before I clean up this PR, I want to get feedback on the changes.

The scope of changes with wrapping these as scalar tensors is localized to the nodes unlike supporting scalar outputs in `runFusion` and elsewhere. I prefer moving with this approach. If there is any other case other than SDPA where scalar outputs are seen, the second approach on allowing scalar outputs can be pursued.

Wdyt?

wujingyue:
> @jjsjann123 @wujingyue: Before I clean up this PR, I want to get feedback on the changes.
> 
> The scope of changes with wrapping these as scalar tensors is localized to the nodes unlike supporting scalar outputs in `runFusion` and elsewhere. I prefer moving with this approach. If there is any other case other than SDPA where scalar outputs are seen, the second approach on allowing scalar outputs can be pursued.
> 
> Wdyt?

SGTM!

Priya2698:
!build

tests/cpp/test_sdpa_node.cpp
```diff
@@ -610,4 +634,91 @@ TEST_F(SDPATest, AttnProgram) {
   EXPECT_TRUE(at::allclose(out[0], expected_out));
 }
 
+TEST_F(SDPATest, AttnFwdBwd) {
+  NVFUSER_TEST_CUDA_ARCH_GUARD(8, 0);
```
wujingyue, line 638: What's the problem with Volta? `at::scaled_dot_product_attention` doesn't like it?

tests/cpp/test_sdpa_node.cpp
```diff
@@ -610,4 +634,91 @@ TEST_F(SDPATest, AttnProgram) {
...
+
+  auto fusion = std::make_unique<Fusion>();
```
wujingyue, line 640:
```suggestion
  Fusion fusion;
```
seems enough. `unique_ptr` would be needed for things like FusionExecutorCache, which claims ownership.

wujingyue (approved):

tests/cpp/test_sdpa_node.cpp
Priya2698: Flash Attention requires Ampere and above.

tests/cpp/test_sdpa_node.cpp
Priya2698: Once I add scheduling support for bwd pass, I will switch to FusionExecutorCache, so I used unique_ptr.

csrc/ir/nodes.cpp
jjsjann123: Please keep the code as-is, scalar_tensor is easier to read.

Meanwhile, `at::tensor(*query_seq_len.maybe_as_int(), at::dtype(at::kLong))` should give you the same scalar tensor.

csrc/ir/nodes.cpp
```diff
@@ -4839,11 +4839,9 @@ std::vector<PolymorphicValue> SdpaBwdOp::evaluate(
   // Backward tensor inputs: grad_input, query, key, value, output, logsumexp,
   // cum_seq_q/k
   std::vector<at::Tensor> bwd_inputs;
-  for (auto idx : c10::irange(8)) {
+  for (auto idx : c10::irange(10)) {
```
jjsjann123, line 4842: nitpick. The comment above should be updated
`grad_output`, ... , `max_q/k`

jjsjann123 (approved):
LGTM

csrc/ir/nodes.cpp
Priya2698: I was getting a 1D tensor instead of 0-dim, will check again.
I agree with you that `scalar_tensor` is more clear.

csrc/ir/nodes.cpp
Priya2698: Missed this comment, fixed this in the other PR.

PR #2533, opened 2024-07-04
`HostIrExecutor`: use `ExpressionEvaluator` instead of `unordered_map<Val*, c10::IValue>`

samnordmann:
# What

Before this patch, `HostIrExecutor` was storing tensors in a map `unordered_map<Val*, c10::IValue>`. We replace this map with using `ExpressionEvaluator`.

# Why
- Makes the code base more uniform by using a preexisting class. `ExpressionEvaluator` exactly serves our need
- Can leverage the `evaluate` method that is able to execute some arithmetic and ATen ops to evaluate tensors. This will be used, e.g., for `SliceOp`, see #2535

# How

The way of doing is pretty straightforward:
- instead of assigning a value to `unordered_map<Val*, c10::IValue>`, we use `ExpressionEvaluator::bind`
- instead of retrieving a value from `unordered_map<Val*, c10::IValue>` we use `ExpressionEvaluator::evaluate`


samnordmann:
!build

csrc/host_ir/executor.cpp
```diff
@@ -34,7 +34,9 @@ HostIrExecutor::HostIrExecutor(
 std::vector<at::Tensor> HostIrExecutor::runWithInput(
     std::unordered_map<Val*, c10::IValue> val_to_IValue) {
   // process input values
-  val_to_IValue_ = std::move(val_to_IValue);
+  for (auto it : val_to_IValue) {
```
wujingyue, line 37:
```suggestion
  for (const auto& [val, ivalue] : val_to_IValue) {
```

wujingyue: This avoids an indirection via `it`.

wujingyue (approved):

samnordmann:
in the commit ["retrieve only known values from expression evaluator"](https://github.com/NVIDIA/Fuser/pull/2533/commits/6672696f0684b9bdabffa9046b959a0153a9bac9), I introduced a small behavior change.

With this patch, when retrieving a Tensor for using it as input to a kernel or a Communication, or to collect global output, we first check that the tensor has already been computed by using the method `ExpressionEvaluator::isKnown`. This behavior is more strict than before, and ensures we do not implicitly hide host operations when using `ExpressionEvaluator::evaluate`.

samnordmann:
!build

csrc/host_ir/executor.cpp
```diff
@@ -14,6 +14,32 @@ namespace nvfuser {
 
 namespace hir {
 
+namespace {
+
+at::Tensor getKnownTensorOrEmptyTensor(
```
wujingyue, line 19:
```suggestion
at::Tensor getKnownTensorOrUndefined(
```

Naming nit: The word "empty" may confuse readers to think it's returned by `torch.empty`. In fact, `at::Tensor` creates an undefined tensor (i.e., `tensor.defined() == false`), whereas `torch.empty` creates an uninitialized but defined (with a valid shape/stride etc) tensor.

wujingyue (approved):
Thanks for the cleanup!

Issue #2532, opened 2024-07-04
an error seemingly related to matmul


crcrpar:
nvfuser commit: d75fc93
cuda: 12.5

```python
import torch
from nvfuser import FusionDefinition, DataType

def nvfuser_fusion_id1(fd : FusionDefinition) -> None :
    T0 = fd.define_tensor(shape=[-1, -1, 1], contiguity=[True, None, True], dtype=DataType.Float, is_cpu=False, stride_order=[2, 0, 1])
    T1 = fd.define_tensor(shape=[-1, 1, -1], contiguity=[True, None, True], dtype=DataType.Float, is_cpu=False, stride_order=[2, 1, 0])
    T2 = fd.ops.sum(T1, dims=[0, 1], keepdim=False, dtype=DataType.Null)
    T3 = fd.ops.matmul(T0, T1)
    T4 = fd.ops.sum(T3, dims=[0], keepdim=False, dtype=DataType.Null)
    fd.add_output(T2)
    fd.add_output(T4)

with FusionDefinition() as fd:
    nvfuser_fusion_id1(fd)

inputs = [
    torch.randn((262400,), dtype=torch.float32, device='cuda:0').as_strided((1025, 256, 1), (256, 1, 256)),
    torch.randn((1049600,), dtype=torch.float32, device='cuda:0').as_strided((1025, 1, 1024), (1024, 1024, 1)),
]
fd.execute(inputs)
```

```console
RuntimeError: Expected T4_g[ iS103{( ceilDiv(( ceilDiv(( ceilDiv(( ( (( (( getMetaData(T4) )).logical_size ))[1] ) * ( ( (( (( getMetaData(T4) )).logical_size ))[2] ) * 1 ) ), 4) ), blockDim.x) ), 1) )}, iS102{blockDim.x}, iS112{( ceilDiv(( ceilDiv(( ceilDiv(( ceilDiv(( (( (( getMetaData(T4) )).logical_size ))[0] ), blockDim.y) ), 2) ), 1) ), gridDim.y) )}, iS104{1}, iS100{4}, iS111{gridDim.y}, iS106{blockDim.y}, iS110{1}, iS108{2} ] to be bound to a tensor of rank 4, but got a tensor of rank 3
Exception raised from validateValWithConcreteValue at /opt/pytorch/nvfuser/csrc/expr_evaluator.cpp:38 (most recent call first):
[...]

PR #2534, opened 2024-07-05
Check ATen-evaluated tensor sizes

jacobhinkle:
This would have surfaced #2532 sooner.

### DIFF
--- a/csrc/executor.cpp
+++ b/csrc/executor.cpp
@@ -1889,6 +1889,29 @@ std::vector<at::Tensor> FusionExecutor::evaluateFusionOutputs(
     for (const auto& out_val : fusion()->outputs()) {
       auto out_tensor =
           expr_eval.evaluate(out_val->as<TensorView>()).as<at::Tensor>();
+      if (auto out_tv = dynamic_cast<TensorView*>(out_val)) {
+        const std::vector<IterDomain*>& tv_logical =
+            TensorDomain::noReductions(out_tv->getLogicalDomain());
+        NVF_ERROR(
+            out_tensor.dim() == tv_logical.size(),
+            "ATen-evaluated tensor with shape ",
+            out_tensor.sizes(),
+            " does not match expected dimension of TensorView ",
+            out_tv->toString());
+        for (size_t i : c10::irange(tv_logical.size())) {
+          IterDomain* id = tv_logical[i];
+          if (id->isBroadcast() && !id->hasExpandedExtent()) {
+            NVF_ERROR(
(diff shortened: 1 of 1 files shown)


jacobhinkle:
Note that some failures are expected due to #2532.

jacobhinkle:
!build

csrc/executor.cpp
```diff
@@ -1889,6 +1889,29 @@ std::vector<at::Tensor> FusionExecutor::evaluateFusionOutputs(
...
+          if (id->isBroadcast() && !id->hasExpandedExtent()) {
+            NVF_ERROR(
+                out_tensor.size(i) == 1,
+                "Expected size-1 axis at position ",
+                i,
+                " of ATen-evaluated tensor ",
+                out_tv->toString(),
+                " but found size ",
+                out_tensor.size(i));
+          }
```
jacobhinkle, line 1912: I guess instead of this we could just evaluate the `getMaybeExpandedExtent` for each axis and check that the output size matches. Also, we could wrap this check in `#ifndef NDEBUG`; wdyt @wujingyue ?

wujingyue: I like catching errors earlier. Unless it's proven to be expensive, I'd turn it on unconditionally (thus no `#ifndef`).

Question (maybe to @zasdfgbnm): the added checks seem to be a subset of https://github.com/NVIDIA/Fuser/blob/bad998ae277ffc2f43fdc28dca07d01d737a1623/csrc/expr_evaluator.cpp#L99, which is only triggered by `ExpressionEvaluator::bind` at this moment. Is it a bad idea to bind outputs? In the launch-a-kernel path of FusionExecutor::runFusion, `allocateOutputs` binds outputs anyway: https://github.com/NVIDIA/Fuser/blob/bad998ae277ffc2f43fdc28dca07d01d737a1623/csrc/executor.cpp#L1111.

wujingyue (approved):

csrc/executor.cpp
zasdfgbnm: > Is it a bad idea to bind outputs?

I don't see any problem for doing so.

jacobhinkle: Thanks guys. Updated to bind the output to `expr_eval`.

Priya2698 (approved):

jacobhinkle:
!build

jacobhinkle:
This is ready to merge but it causes CI failures due to #2532. Will merge once that is fixed.

jacobhinkle:
Closing in favor of #2737

PR #2536, opened 2024-07-05
Patch for allocation domains of shmem tensors used for transpose

naoyam:
(stacked on #2523)

The new indexing framework starts with identifying (or inferring) allocation domains of each tensor. The logic is similar to what's done in the allocation lowering pass, and indeed some of the logic is reused in indexing. 

One case where the inference fails to identify is the required order of allocation domains. One instance is addressed in #2522. Another instance is when a shared memory is used as a staging buffer for storing a transposed tensor. Please see the code comment of `patchAllocationOfTransposedSmemTensor`. 

While this should be good enough for now, I'm wondering if we should actually set the allocation domain of the staging tensor explicitly. When scheduling the fusion, we know what the allocation domain should look like. What's done here is to infer the domain from the fusion pattern. This approach doesn't require any change in the scheduler but the inference may not be always correct. 

Let me know what you think, @zasdfgbnm


zasdfgbnm:
My understanding of the problem:

The legacy indexing does the following when indexing `t2_s` as producer:

1. When allocating, I treat the loop domain as allocation domain `[I0/32a, I1/32b, 32a*32b/4, 4]`.
2. When  indexing, I treat the logical domain as allocation domain: `[I1, I0]`.
  2.1. although `[I1, I0]` are allocation domain, their extents are not the original extents in the `IterDomain` objects. Instead, the "real" extents of both `I1` and `I0` are 32.

This behavior is not the desired behavior we want to have in `TensorIndexer`, and is not easily mimic-able either. Is my understanding correct?

Have you considered another option to hack?

Create a:
```
I0'{32} = resize(I0, 0, 32 - extent)
I1'{32} = resize(I1, 0, 32 - extent)
```
and make `[I1', I0']` the allocation domain? This sounds like the closest to the legacy code.

Basically, what we need to do is to pull in the extent inference code in legacy indexing, and use the resized logical domain instead of the loop domain as the allocation domain.

naoyam:
> My understanding of the problem:
> 
> The legacy indexing does the following when indexing `t2_s` as producer:
> 
> 1. When allocating, I treat the loop domain as allocation domain `[I0/32a, I1/32b, 32a*32b/4, 4]`.
> 2. When  indexing, I treat the logical domain as allocation domain: `[I1, I0]`.
>    2.1. although `[I1, I0]` are allocation domain, their extents are not the original extents in the `IterDomain` objects. Instead, the "real" extents of both `I1` and `I0` are 32.

Correct. The allocation domain is always assumed to be the same as the logical domain, so `[I1, I0]` in this case. And we propagate extent mappings from loop domains by considering that inlined domains should be effectively size one (i.e., should not contribute to allocation sizes).

> 
> This behavior is not the desired behavior we want to have in `TensorIndexer`, and is not easily mimic-able either. Is my understanding correct?

We do not consider `[I1, I0]` as the allocation domain anymore since it just isnt'. So, mimicking doesn't seem to be the right option.

> 
> Have you considered another option to hack?
> 
> Create a:
> 
> ```
> I0'{32} = resize(I0, 0, 32 - extent)
> I1'{32} = resize(I1, 0, 32 - extent)
> ```
> 
> and make `[I1', I0']` the allocation domain? This sounds like the closest to the legacy code.
> 
> Basically, what we need to do is to pull in the extent inference code in legacy indexing, and use the resized logical domain instead of the loop domain as the allocation domain.

I'm not sure why resize? I think what we should really do would be:

```
t2->setAllocationDomain([32b, 32a]);
```

when `t2` is scheduled, and no inference should be needed in lowering.

zasdfgbnm:
> I'm not sure why resize?

I just feel that resize is a magic hack that fix most allocation domain issue, and is relatively easy to implement. My hope is this hack will magically save us time from looking into many failing tests. If you don't think this will save time, then I do not see any value on having this hack. When talking about hack, the only thing I care is how much time it takes to implement this hack, and how much time it will take to remove this hack. I generally won't think like: "does this hack align better with how we understand nvFuser than the other hack?"

I agree that this hack is not the final solution we want to go with. And I agree that `t2->setAllocationDomain([32b, 32a]);` is a strictly better solution.

Regarding the allocation domain, I believe both lowering and scheduler should try to be a good citizen and make the other's life easier. I believe we do need to have some simple non-hacky rules to infer allocation domain when it is empty. Which rule we should have is not clear to me now. A simple thing that comes to my mind is to traverse along the shortest path from this tensor's loop domain to all consumer's loop domains, and select something on that shortest path that matches with certain constraints (for example, vectorization must be the innermost in allocation domain). On the other hand, schedulers should also consider setting the correct allocation domain as its job, and try to do it on every tensor.

Regarding this PR, since you already have a easily-removable hack that make things work, I do not see any reason for not accepting it.

zasdfgbnm (approved):

naoyam:
> > I'm not sure why resize?
> 
> I just feel that resize is a magic hack that fix most allocation domain issue, and is relatively easy to implement. My hope is this hack will magically save us time from looking into many failing tests. If you don't think this will save time, then I do not see any value on having this hack. When talking about hack, the only thing I care is how much time it takes to implement this hack, and how much time it will take to remove this hack. I generally won't think like: "does this hack align better with how we understand nvFuser than the other hack?"

Sounds good. The workaround of this PR at least seems sufficient for the C++ tests. May need further tweaks, but for now this seems good enough.

> 
> I agree that this hack is not the final solution we want to go with. And I agree that `t2->setAllocationDomain([32b, 32a]);` is a strictly better solution.
> 
> Regarding the allocation domain, I believe both lowering and scheduler should try to be a good citizen and make the other's life easier. I believe we do need to have some simple non-hacky rules to infer allocation domain when it is empty. Which rule we should have is not clear to me now. A simple thing that comes to my mind is to traverse along the shortest path from this tensor's loop domain to all consumer's loop domains, and select something on that shortest path that matches with certain constraints (for example, vectorization must be the innermost in allocation domain). On the other hand, schedulers should also consider setting the correct allocation domain as its job, and try to do it on every tensor.
> 

Agreed. Inference seems tricky, so whenever possible, explicit setting by the schedulers should help.

> Regarding this PR, since you already have a easily-removable hack that make things work, I do not see any reason for not accepting it.

👍

PR #2539, opened 2024-07-06
Clean Up Kernel Measurement Instrumentation and Consolidate on `FusionProfiler` Usage

kevinstephano:
# Summary

Removes redundant logic to measure kernel times and effective bandwidth in `FusionExecutor` and enables the simplification of the `FusionExecutor` code.

# TL;DR

* Removes the option `EffectiveBandwidth` as a separate debug flag.  The Effective Bandwidth can be found in the `FusionProfile` as `FusionProfile::effective_bandwidth_gbs`.
* Removes the redundant kernel measurement code found in `FusionExecutor` that was triggered by the setting of the data member `measure_kernel_time_`.
* Removes a lot of complicated logic found the `FusionExecutor` header file to calculate the input and output bytes that is not necessary for the `FusionProfiler`.
* Switches tests to use the `FusionProfiler` to measure single kernel time and fusion time.
* Switches CPP benchmarks to use the `FusionProfiler`'s `kernel_time_ms`.


kevinstephano:
!build

jacobhinkle (approved):
Looks like a good cleanup to me

rdspring1 (approved):
LGTM

csrc/kernel_cache.cpp
```diff
@@ -1208,37 +1199,6 @@ std::vector<at::Tensor> FusionKernelRuntime::runKernelWithInput(
   }
   auto outputs = executor.runFusion(args, launch_params, compile_params);
 
-  // Accumulate the kernel time of each segment
-  kernel_time_ms_ += executor.kernelTimeMs();
-
-  // Print relevant information all at once for easy debuging of perf
-  if (isDebugDumpEnabled(DebugDumpOption::PerfDebugVerbose) &&
```
Priya2698, line 1215: Is the option `DebugDumpOption::PerfDebugVerbose` redundant as well?
Then we should remove it as well.

User although may still like to print out this information at once.

kevinstephano: `PerfDebugVerbose` allows for debugging a range of things like segmentation and occupancy beyond what is captured in the profiler.  I was only concerned with cleaning up the redundancy I had made in order to keep kernel time and effective bandwidth measurements working while I was developing.

kevinstephano:
!build

Issue #2473, opened 2024-06-27
reduce inter-segment I/O using rematerialization


liqiangxl:
**Motivation:**
Seems like we need an IO-aware segmenter to reduce number of tensors between different segments. A real example is from #2146, where the tensor is segmented into 2 kernels with 3 inter-segment tensors.
```
T12_g[ iS35{56}, iS36{1024}, iS259{1024} ] float
T14_g[ iS41{56}, iS42{1024}, bS43{1 ex 1024} ] float
T15_g[ iS44{56}, iS45{1024}, iS264{1024} ] float
```
These tensors can be calculated pointwisely from other inputs in segmentation-2. In other words, they can be re-calculated in segmentation-2 instead of being written out in segmentation-1 and read back in segmentation-2.

**Potential fix:**
instead of greedily merge as many exprs as possible, may also check the influecne on IO bytes. In other words, may change the target from minimize number of segments to minimize total IO bytes.

**Example from #2146**
```
Segmented_Fusion Dump: -- fusion segments:
Segmented_Fusion{ 
groups: 
g{0, 1, 2, 3, 5, 6, 8, 9, 10, 11, 12, 19, 20, 21, 22, 23, 24, 25, 26, 28, 29, 30, 31, 32, 33, 34, 35, 123, 124}

g{0, 9, 13, 14, 15, 16, 18, 19, 27, 36, 37, 38, 39, 40, 41, 42, 43, 44, 45, 46, 47, 48, 49, 50, 51, 52, 53, 58, 60, 63, 65, 66, 67, 69, 70, 71, 72, 73, 74, 75, 76, 125, 126, 127, 128, 129, 130}

edges: 
e{ g{0, 1, 2, 3, 5, 6, 8, 9, 10, 11, 12, 19, 20, 21, 22, 23, 24, 25, 26, 28, 29, 30, 31, 32, 33, 34, 35, 123, 124}
 -> g{0, 9, 13, 14, 15, 16, 18, 19, 27, 36, 37, 38, 39, 40, 41, 42, 43, 44, 45, 46, 47, 48, 49, 50, 51, 52, 53, 58, 60, 63, 65, 66, 67, 69, 70, 71, 72, 73, 74, 75, 76, 125, 126, 127, 128, 129, 130}
(T12_g[ iS35{56}, iS36{1024}, iS259{1024} ]) }

[...]

Issue #2418, opened 2024-06-14
Data-driven heuristics


jacobhinkle:
This is a medium- to long-term proposal to modify our heuristics system to enable some machine-specific information to be used to choose heuristic parameters.

The goal is to have a python script that is run once on the user's machine whenever nvfuser is updated to create a file holding a collection of measured named variables: things like "matmul.splitk.elements_reduced_per_second", "normalization.additional_registers" etc. These are things that are very difficult to model accurately and are much more reliably set through experimentation, which is what the script would be doing.

# Background

Currently most of our heuristics do not directly use measurements of machine performance. Instead, they are rule-based and are developed through trial and error on actual machines. Special casing for different architectures in heuristic code is necessary at times based on constraints in hardware or software, but AFAIK we don't currently write explicit parametrized timing models for any of our heuristics. To do so, we need some model parameters that are not trivial derive by querying the cuda device properties.

We currently have a plugin interface for the matmul heuristic #2049. This is a matmul scheduler-specific approach that allows a user to provide a C++ library at runtime that will take a problem description and previous heuristic and return an updated heuristic. The heuristic is a kernel configuration struct that can be converted to a `MatmulParams` object. Technically a user could use this to recreate something like the system described in this proposal, but it would be specific to the matmul scheduler, and we should facilitate doing this more easily for all our schedulers.

# Proposal

I propose that we create a container class `MachineInfo` that behaves very much like `std::unordered_map<std::string, PolymorphicValue>` (except we really only need `bool`, `int64_t`, and `double` so we could probably wrap a smaller `DynamicType`). This would hold measured information that can be referred to in scheduler heuristics. For example, inside a normalization heuristic we might be interested in the bandwidth from global to shared memory at a given alignment so we would query it like this:
```c++
  double load_bw_gbps = machine_info.query<double>("normalization.load_bandwidth_gbps.align8", /*default=*/1300.0);
```
In order to populate a value like this, we would curate a collection of Fusions to execute and measure their timings in order to estimate the variables of interest in a regression approach. I propose that we do that in a Python script that we provide in `tools`.

The `MachineInfo` object would be serialized and loaded automatically. It could be edited by the user manually but most often it would be filled by a measurement script that would run a bunch of trial problems in order to estimate the values. We should save the nvfuser and other software versions along with device information in the serialized file and we should warn users when it is out of date, as the measured values might change based on changes to our codebase or to the installed NVRTC or driver versions.

[...]

PR #2542, opened 2024-07-07
[Python Frontend] Use heuristic schedulers in user-defined schedules

rdspring1:
This PR allows using the heuristic schedulers in the user-defined schedules.

```python
# Return if we can use specified scheduler with given fusion.
# An error message string is generated to explain why the specified Scheduler cannot be used for a given fusion.
is_schedulable, error_msg = fd.sched.can_schedule(ScheduleHeuristic heuristic)

# Return all available schedulers for a given fusion in priority order.
available_schedules = fd.sched.find_compatible_schedulers()

# Schedules the fusion using the specified scheduler
fd.sched.schedule(ScheduleHeuristic heuristic)
```

### Motivation

1. This PR is a first step to modifying heuristic parameters from the python frontend. See #2418.
2. Each `UserSchedule` maps to a single `FusionExecutor`, which bypasses the nvfuser segmentation algorithm. The user can manually segment a fusion and apply the desired heuristic scheduler.
5. Understand which schedulers are available for a given fusion. See #2473.


rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

csrc/python_frontend/python_bindings.cpp
```diff
@@ -445,6 +474,17 @@ void initNvFuserPythonBindings(PyObject* module) {
       .value("shared", MemoryType::Shared)
       .value("global", MemoryType::Global);
 
+  //! Scheduler Type for scheduling
+  py::enum_<ScheduleHeuristic>(nvfuser, "SchedulerHeuristic")
+      .value("no_op", ScheduleHeuristic::NoOp)
+      .value("pointwise", ScheduleHeuristic::PointWise)
+      .value("reduction", ScheduleHeuristic::Reduction)
```
jacobhinkle, line 481:
```suggestion
      .value("matmul", ScheduleHeuristic::Matmul)
      .value("reduction", ScheduleHeuristic::Reduction)
```

csrc/python_frontend/python_bindings.cpp
```diff
@@ -3266,5 +3306,60 @@ void initNvFuserPythonBindings(PyObject* module) {
...
+      "can_schedule",
+      [](FusionDefinition::SchedOperators& self,
+         const ScheduleHeuristic& heuristic) {
+        NVF_CHECK(
+            self.validUse(),
+            "Attempting to use a SchedOperators Op prior to definition!");
+        return canSchedule(self.fusion_definition->userSchedule(), heuristic);
+      },
+      py::arg("heuristic"));
+  nvf_sched.def("which_schedulers", [](FusionDefinition::SchedOperators& self) {
```
jacobhinkle, line 3319: The name is not very clear to me. Something like `fd.sched.supported_schedulers()` might be better?

csrc/python_frontend/python_bindings.cpp
jacobhinkle: This would allow us to fuse matmuls manually even without the `NVFUSER_ENABLE=fuse_matmul` option.

jacobhinkle (commented):
This looks good to me! I'll leave to others to approve.

csrc/python_frontend/python_bindings.cpp
rdspring1: I changed it to `find_compatible_schedulers`

rdspring1:
!build

csrc/python_frontend/python_bindings.cpp
```diff
@@ -218,6 +221,32 @@ struct DimInfo {
...
+      sched->schedule != nullptr,
+      "Requires Fusion to use heuristic schedulers");
+  NVF_ERROR(
+      sched->runtime_info != nullptr,
+      "Requires SchedulerRuntimeInfo to use heuristic schedulers");
+
+  // Enable collection of messages from canScheduleRejectReason
+  DebugDumpOptionsGuard debug_dump_options_guard;
+  DebugDumpOptionsGuard::getCurOptions().set(
+      DebugDumpOption::FusionSegmenterLog);
```
jjsjann123, line 239: Is this a debug thing that we forgot to remove?

jjsjann123: NVM, realized that FusionSegmenterLog is actually just a bad name for the scheduler rejection.

csrc/python_frontend/python_bindings.cpp
```diff
@@ -3266,5 +3308,63 @@ void initNvFuserPythonBindings(PyObject* module) {
...
+
+        Fusion* fusion = self.fusion_definition->userSchedule()->schedule.get();
+        NVF_ERROR(
+            fusion != nullptr, "Requires Fusion to use heuristic schedulers");
+
+        SchedulerRuntimeInfo* runtime_info =
+            self.fusion_definition->userSchedule()->runtime_info.get();
+        NVF_ERROR(
+            runtime_info != nullptr,
+            "Requires SchedulerRuntimeInfo to use heuristic schedulers");
```
jjsjann123, line 3335: nitpick. we are using similar checks in `canSchedule`.
```
  NVF_ERROR(
      sched->schedule != nullptr,
      "Requires Fusion to use heuristic schedulers");
  NVF_ERROR(
      sched->runtime_info != nullptr,
      "Requires SchedulerRuntimeInfo to use heuristic schedulers");
```

Maybe put this as a method in `UserSchedule`

csrc/python_frontend/python_bindings.cpp
```diff
@@ -3266,5 +3308,63 @@ void initNvFuserPythonBindings(PyObject* module) {
...
+      "schedule",
+      [](FusionDefinition::SchedOperators& self,
+         const ScheduleHeuristic& heuristic) {
+        NVF_CHECK(
+            self.validUse(),
+            "Attempting to use a SchedOperators Op prior to definition!");
+        UserSchedule* sched = self.fusion_definition->userSchedule();
+        NVF_CHECK(
+            sched->heuristic_scheduler == nullptr,
+            "Heuristic Scheduler is already defined for this UserSchedule");
```
jjsjann123, line 3359: nitpick: would be nice to have the stages of a UserSchedule be exposed as a method, rather than query on each field being initialized.

i.e. `canBeScheduled` (when we check for fusion/runtime info),  `hasBeenScheduled` (when we check for heuristic_schedule).

jjsjann123 (commented):
LGTM. I'll take another pass at the tests.

Have some nitpick on some code clean up~~

csrc/python_frontend/python_bindings.cpp
rdspring1: I added separate get functions for `Fusion*` and `SchedulerRuntimeInfo*` that includes these checks respectively.

csrc/python_frontend/python_bindings.cpp
rdspring1: I created `UserSchedule::scheduleWithHeuristic` to build the heuristic and schedule the fusion.

tests/python/test_schedule_ops.py
```diff
@@ -22,6 +35,29 @@ def is_pre_volta():
     return prop.major < 7
 
 
+# A helper function to test heuristic schedulers with user schedules
+def _apply_scheduler_helper(schedule, selected_heuristic):
+    # Check that only selected heuristic is available as a scheduler
+    available_heuristics = schedule.find_compatible_schedulers()
+    assert available_heuristics == [selected_heuristic]
```
jjsjann123, line 42: I think we need this
```suggestion
    assert set(available_heuristics) == set(selected_heuristic)
```

tests/python/test_schedule_ops.py
```diff
@@ -22,6 +35,29 @@ def is_pre_volta():
...
+
+    # Double-check with can_schedule
+    status, _ = schedule.can_schedule(selected_heuristic)
+    assert status
+
+    # Check that the other schedulers are not compatible with this fusion
+    assert all(
+        [
+            not schedule.can_schedule(h)[0]
```
jjsjann123, line 51: Is this expected?

On the cpp side, I think we can have multiple executor being able to compile the same fusion. i.e. no-op and pointwise have overlap when alias analysis is run successfully.
We cannot run preseg passes yet, so I don't have a good counter example right now. But I'm curious what's your motivation to assert here.

jjsjann123 (approved):
LGTM

some minor nitpick on tests and future looking behaviors

tests/python/test_schedule_ops.py
rdspring1: The implicit assumption is that only one scheduler will work in the `auto_scheduler` tests. `_apply_scheduler_helper` checks for this specifically.

rdspring1:
!build

PR #2529, opened 2024-07-03
Segment out meta ops from a fusion. 

wujingyue:
We currently do that to some extent by adding `segment_set` before aliased outputs. This PR modifies and extends the logic for aliased inputs as well. 

Fixes #2395

PR #2548, opened 2024-07-08
Fix a bug in replaceOutput.

wujingyue:
See code comments.

### DIFF
--- a/csrc/fusion.cpp
+++ b/csrc/fusion.cpp
@@ -320,7 +320,11 @@ void Fusion::replaceOutput(Val* output, Val* replacement) {
     }
     if (output->getValType().value() == ValType::TensorView) {
       output->setIsFusionOutput(false);
-      output->as<TensorView>()->setMemoryType(MemoryType::Local);
+      // If `output` is both an input and an output before the replacement,
+      // don't localize it.
+      if (!output->isFusionInput()) {
+        output->as<TensorView>()->setMemoryType(MemoryType::Local);
+      }
     }
     // Mark uses invalid so that they will be reset next time uses() is called
     invalidateTvUses();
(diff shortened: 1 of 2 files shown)


tests/cpp/test_preseg_passes.cpp
wujingyue, line 1: Lots of cosmetic changes to this file, but the real change which adds a test is at the end.

csrc/fusion.cpp
```diff
@@ -320,7 +320,11 @@ void Fusion::replaceOutput(Val* output, Val* replacement) {
     }
     if (output->getValType().value() == ValType::TensorView) {
       output->setIsFusionOutput(false);
-      output->as<TensorView>()->setMemoryType(MemoryType::Local);
+      // If `output` is both an input and an output before the replacement,
+      // don't localize it.
+      if (!output->isFusionInput()) {
+        output->as<TensorView>()->setMemoryType(MemoryType::Local);
```
naoyam, line 326: Just for extra safety, could you also add an assertion in `setMemoryType` that tensors must not be an input nor an output?

naoyam (approved):
LGTM. Thanks for the fix.

csrc/fusion.cpp
wujingyue: The assertion already exists (https://github.com/NVIDIA/Fuser/blob/6c6f3a40e09e6f8bece80b9b79c543945846c71b/csrc/tensor_view.cpp#L1227-L1231) and was indeed triggered when I worked on https://github.com/NVIDIA/Fuser/pull/2529. 

The assertion says "if I/O, then assert global", which is equivalent to "if not global, then assert not I/O".

wujingyue:
!build

PR #2550, opened 2024-07-09
Support resize

naoyam:
(Stacked on top of #2536)

Resize needs a little tweak when finding an indexing path. See IndexingTraversal.


csrc/id_model/indexing_traversal.h
```diff
@@ -0,0 +1,92 @@
...
+
+namespace nvfuser {
+
+// BFS traversal for indexing. The only difference with the default
+// ValGraphBFS is that for indexing there must be a special care taken
+// when resize is involved since there can be multiple paths and
+// there's only one correct path. Specifically, any resize expr group
+// node must appear in the root-logical path of the consumer
+// tensor. Otherwise, resize nodes should be ignored. See
+// IndexingTest.ResizePath for a concret example.
```
jacobhinkle, line 21:
```suggestion
// IndexingTest.ResizePath for a concrete example.
```

csrc/id_model/indexing_traversal.h
```diff
@@ -0,0 +1,92 @@
...
+      const Expr* expr,
+      const ValGraph& graph,
+      std::vector<GroupType> from_groups,
+      std::vector<GroupType> to_groups)
+      : ValGraphBFS(graph, from_groups, to_groups) {
+    auto consumer_tv = ir_utils::getTvOutput(expr);
+    NVF_ERROR(consumer_tv != nullptr);
+    if (consumer_tv->hasRoot()) {
+      // Remember the resize exprs appearing in the consumer
+      // tensro. These resize exprs are the only ones that should be
```
jacobhinkle, line 34:
```suggestion
      // tensor. These resize exprs are the only ones that should be
```

jacobhinkle (approved):
LGTM other than the one redundant block of code. I like this approach of using a hook to pretend some nodes don't exist during the BFS.

PR #2559, opened 2024-07-10
Host benchmarking 

Priya2698:
1. Adds a fusion profiler based timer to measure the host timings for nvFuser.
2. To measure for uncached host overhead, the `setup` function resets `FusionCache` and initializes a new FusionDefinition

### DIFF
--- a/benchmarks/python/core.py
+++ b/benchmarks/python/core.py
@@ -357,7 +357,7 @@ def run_benchmark(
         benchmark_fn: Target function
         inputs: Inputs to the target function
         iobytes (Optional): When given, IO bytes computation is skipped
-                and this is used to compute the metrics.
+                and this is used to compute SOL and bandwidth.
         device (Optional): Default: CUDA, Possible values: ["cuda", "host"].
             Using device="host" is only allowed with nvFuser FusionDefinition.
         fusion_fn (Optional): Must be provided if device = "host".
--- a/benchmarks/python/core.py
+++ b/benchmarks/python/core.py
@@ -9,7 +9,7 @@ from torch.autograd import DeviceType
 from torch.profiler import profile, ProfilerActivity
 from typing import List, Callable, Union, Tuple
 import numpy as np
-from nvfuser import FusionDefinition
+from nvfuser import FusionDefinition, FusionCache
 
 
 def get_device_properties() -> Tuple[int, float]:
@@ -171,7 +171,7 @@ class NVFBenchmark:
     torchprofiler-based timer and metric computation.
     """
 
-    def __init__(self, benchmark_fixture, device: str ="cuda", fd: FusionDefinition = None, precision: float = 1e-6):
+    def __init__(self, benchmark_fixture, device: str ="cuda", precision: float = 1e-6):
         """
         Arguments:
             benchmark_fixture: pytest-benchmark fixture passed to every
@@ -186,7 +186,7 @@ class NVFBenchmark:
         """
 
         self.device = device 
-        self.fd = fd
+        self.fd = None # Set through setup() for host benchmarking.
         self.benchmark = benchmark_fixture
 
         if (device == "cuda"):
@@ -205,7 +205,10 @@ class NVFBenchmark:
         
         # Global montonic clock
         self.current_time = 0.0
-        breakpoint()
+
+        # Specifies if the timer in host measurement is called at the start/finish of execution.
+        # Timings are measured at the end of execution.
+        self.execution_start = True
 
     def __call__(self, function_to_benchmark: Callable, *args, **kwargs):
         return self.benchmark(function_to_benchmark, *args, **kwargs)
@@ -238,10 +241,12 @@ class NVFBenchmark:
         return self.current_time
     
     def fusionprofile_timer(self) -> float:
-        breakpoint()
-        profile = self.fd.profile()
-        elapsed_host_time = profile.host_time_ms / 1e3
-        self._increment_global_time(elapsed_host_time)
+        if (not self.execution_start):
+            profile = self.fd.profile()
+            elapsed_host_time = profile.host_time_ms / 1e3
+            self._increment_global_time(elapsed_host_time)
+        self.execution_start =  not self.execution_start
+        return self.current_time
 
     def _get_kernel_time(
         self, prof_averages: torch.autograd.profiler_util.EventList
@@ -334,11 +339,11 @@ BENCHMARK_CONFIG = {"rounds": 1, "warmup_rounds": 0}
 
 def run_benchmark(
     benchmark: pytest_benchmark.fixture.BenchmarkFixture,
-    benchmark_fn: Callable,
+    benchmark_fn: Callable | None,
     inputs: Union[torch.Tensor, List],
     iobytes: int = None,
     device: str = "cuda",
-    fd: FusionDefinition = None,
+    fusion_fn: Callable = None,
 ) -> Union[torch.Tensor, List]:
     """
     Benchmarks the target function using torchprofiler and stores metrics as extra information.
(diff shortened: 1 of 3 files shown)


benchmarks/python/core.py
```diff
@@ -330,22 +356,59 @@ def run_benchmark(
...
+    ], f"Unsupported device type: {device}. Use one of cuda/host."
+
+    if device == "host":
+        assert fusion_fn is not None and benchmark_fn is None
+
+    nvf_benchmark = NVFBenchmark(benchmark, device=device)
+
+    def host_benchmark_fn(inputs, fd):
+        # Set the fd variable used to query the profile object
+        nvf_benchmark.fd = fd
```
Priya2698, line 384: This is required in L249 to query `fd.profile()`

benchmarks/python/core.py
```diff
@@ -330,22 +356,59 @@ def run_benchmark(
...
+        return fd.execute(inputs, profile=True)
+
     def setup():
         clear_l2_cache()
+
+        if device == "host":
+            # Reset the FusionCache to measure host overhead correctly.
+            FusionCache.reset()
+            with FusionDefinition() as fd:
+                fusion_fn(fd)
```
Priya2698, line 394: Since FusionCache.reset() needs to be done before the `with FusionDefinition()` block, the setup reinitializes fd at each round.

benchmarks/python/core.py
```diff
@@ -330,22 +356,59 @@ def run_benchmark(
...
+        "host",
+    ], f"Unsupported device type: {device}. Use one of cuda/host."
+
+    if device == "host":
+        assert fusion_fn is not None and benchmark_fn is None
+
+    nvf_benchmark = NVFBenchmark(benchmark, device=device)
+
+    def host_benchmark_fn(inputs, fd):
+        # Set the fd variable used to query the profile object
```
Priya2698, line 383: A wrapper is required around `fd`, since it is reinitialized in `setup` at each round.

benchmarks/python/core.py
```diff
@@ -170,7 +171,9 @@ class NVFBenchmark:
     torchprofiler-based timer and metric computation.
     """
 
-    def __init__(self, benchmark_fixture, precision: float = 1e-6):
+    def __init__(
+        self, benchmark_fixture, device: str = "cuda", precision: float = 1e-6
+    ):
```
jacobhinkle, line 176: Nit: please add description for device argument to docstring. In particular, what values are expected?

benchmarks/python/core.py
```diff
@@ -228,6 +244,14 @@ def torchprofile_timer(self) -> float:
             self.prof.start()
         return self.current_time
 
+    def fusionprofile_timer(self) -> float:
+        if not self.execution_start:
+            profile = self.fd.profile()
+            elapsed_host_time = profile.host_time_ms / 1e3
+            self._increment_global_time(elapsed_host_time)
+        self.execution_start = not self.execution_start
```
jacobhinkle, line 252: Can you explain in a comment why we need to toggle `self.execution_start` here?

benchmarks/python/core.py
jacobhinkle: Can you put that in a code comment?

benchmarks/python/test_many_pointwise_ops.py
```diff
@@ -0,0 +1,64 @@
...
+        a = fd.ops.add(x, y)
+
+    if dtype in PROMOTE_DTYPES:
+        a = fd.ops.cast(a, dtype=dtype)
+    fd.add_output(a)
+
+
+# NOTE: num_iters restricted due to issue #1234.
+@pytest.mark.parametrize("num_iters", [2, 4, 8, 16])
+def test_pointwise_ops_benchmark(
```
jacobhinkle, line 34: I wonder if there is a way to automatically modify the test name if we detect that `device="host"`. Barring that, we should just take care that the test name reflects that this is compile time benchmarking.
```suggestion
def test_pointwise_ops_compile_time_benchmark(
```

benchmarks/python/core.py
jacobhinkle: Also, I think calling this argument `device` is confusing. The test will still run on `cuda` device irrespective of this argument. Instead, I think it would be clearer to have something like `measure_execution_only: bool = True` or `measure_compile_time: bool = False`.

jacobhinkle (commented):
I have a general question in addition to these more specific comments:

On `main`, what happens if we write a benchmark like this?
```python
@pytest.mark.parametrize("size", generate_input_sizes(dims=2))
@pytest.mark.parametrize("dtype", FLOAT_DTYPES)
@pytest.mark.parametrize("reduction_axis", [0, 1])
def test_foo(
    benchmark,
    size: tuple,
    dtype: torch.dtype,
    reduction_axis: int,
    disable_validation: bool,
    disable_benchmarking: bool,
    ):
    clear_cuda_cache()

    inputs = [torch.randn(size, device="cuda", dtype=dtype)]

        softmax_fwd_fusion(fd, torch_dtype_to_nvfuser_dtype(dtype), reduction_axis)

    if not disable_validation:
        eager_output = softmax_fwd_fn([inputs[0], reduction_axis])
        fd.validate(inputs, [eager_output])

    if not disable_benchmarking:
        # The following would do _execution_ time measurement
        #run_benchmark(benchmark, fd.execute, inputs)

        # Instead, we wish to do compile+execution time measurement
        def benchmark_fn(*args, **kwargs):
            FusionCache.reset()
                softmax_fwd_fusion(fd, torch_dtype_to_nvfuser_dtype(dtype), reduction_axis)
            return fd.execute(*args, **kwargs)

        run_benchmark(benchmark, benchmark_fn, inputs)
```
That is to say, what limitation in the `benchmark_fn` approach does this new machinery overcome?

benchmarks/python/core.py
Priya2698: Will add a description. I like the idea of using `measure_compile_time` instead of `device`.
Detailed comments are in `run_benchmark` since that is the user-facing function.

benchmarks/python/core.py
Priya2698: `fd.profile()` cannot be queried before `fd.execute`.
Pytest has the following structure when running the benchmark:
```
timer() # Cannot query fd.profile() here. It will raise an error
output = benchmark_fn()
timer() # Query fd.profile() here and increment clock.
```
The alternative is to use a `try-except` block.

Priya2698:
> I have a general question in addition to these more specific comments:
> 
> On `main`, what happens if we write a benchmark like this?
> 
> ```python
> @pytest.mark.parametrize("size", generate_input_sizes(dims=2))
> @pytest.mark.parametrize("dtype", FLOAT_DTYPES)
> @pytest.mark.parametrize("reduction_axis", [0, 1])
> def test_softmax_fwd_compile_time(
>     benchmark,
>     size: tuple,
>     dtype: torch.dtype,
>     reduction_axis: int,
>     disable_validation: bool,
>     disable_benchmarking: bool,
>     ):
>     clear_cuda_cache()
> 
>     inputs = [torch.randn(size, device="cuda", dtype=dtype)]
> 
>     with FusionDefinition() as fd:
>         softmax_fwd_fusion(fd, torch_dtype_to_nvfuser_dtype(dtype), reduction_axis)
> 
>     if not disable_validation:
>         eager_output = softmax_fwd_fn([inputs[0], reduction_axis])
>         fd.validate(inputs, [eager_output])
> 
>     if not disable_benchmarking:
>         # The following would do _execution_ time measurement
>         #run_benchmark(benchmark, fd.execute, inputs)
> 
>         # Instead, we wish to do compile+execution time measurement
>         def benchmark_fn(*args, **kwargs):
>             FusionCache.reset()
>             with FusionDefinition() as fd:
>                 softmax_fwd_fusion(fd, torch_dtype_to_nvfuser_dtype(dtype), reduction_axis)
>             return fd.execute(*args, **kwargs)
> 
>         run_benchmark(benchmark, benchmark_fn, inputs)
> ```
> 
> That is to say, what limitation in the `benchmark_fn` approach does this new machinery overcome?
> 
> It seems like we just need a new option that changes from using device timing via CUPTI to using wall time.

I need access to the `fd` object within `NVFBenchmark` object to query the profile. Which is why I had to move the initialization within setup.

jacobhinkle:
> I need access to the `fd` object within `NVFBenchmark` object to query the profile.

Ah right. I forgot that you're using `fd.profile` to get the CPU timing. Since it's just host timing, what if you just optionally used python to do the timing?

Priya2698:
> > I need access to the `fd` object within `NVFBenchmark` object to query the profile.
> 
> Ah right. I forgot that you're using `fd.profile` to get the CPU timing. Since it's just host timing, what if you just optionally used python to do the timing?

That would be wall-clock time, not necessarily just the host time.

naoyam (commented):
Thanks for adding the host benchmarking. Looks generally good to me. I'll let @jacobhinkle give a stamp.

jacobhinkle:
> That would be wall-clock time, not necessarily just the host time.

Is the difference that "host time" is the sum across threads during parallel compilation which wall clock would not do?

Priya2698:
> > That would be wall-clock time, not necessarily just the host time.
> 
> Is the difference that "host time" is the sum across threads during parallel compilation which wall clock would not do?

That is one way it will be different. I will let @kevinstephano confirm if this is how compilation time is measured.
In general, wall clock time may also capture GPU-only times as well. So the wall clock time may always differ from host times.

kevinstephano:
`host_time` is not a summation of parallel compilations unless the compilations are done serially.  It wraps around the `FusionExecutorCache::runWithInputs()`.  It is effectively wall clock time where wall clock time on the CPU may not include the kernel time.  `CuEvtTime` has the same timer positions as `host_time` except we put events on the GPU so it is effectively `host_time` plus `kernel_time`.

Priya2698:
In any event, the kernel time is likely much smaller than the wall time.

In this particular fusion, yes, the main overhead is host time and kernel time is very small.
I do not see how wall time will always be a good approximation for the host time in larger fusions as well.

jacobhinkle (approved):
LGTM. Thanks for the explanations

Issue #2560, opened 2024-07-10
CombinedSchedulerLayerNormBackward_CUDA may have validation error 


liqiangxl:
After #2511 `CombinedSchedulerLayerNormBackward_CUDA` may have validation error but not always fails.
```
00:35:31 [ RUN      ] NVFuserTest.CombinedSchedulerLayerNormBackward_CUDA
To reproduce: NVFUSER_TEST_RANDOM_SEED=1720577119 NVFUSER_TEST_ATEN_RANDOM_SEED=0 nvfuser_tests --gtest_filter='NVFuserTest.CombinedSchedulerLayerNormBackward_CUDA'
00:35:38 unknown file: Failure
00:35:38 C++ exception with description "aten_output_tensor.allclose( fusion_output_tensor.to(aten_output_tensor.dtype()), tolerance_values.second, tolerance_values.first, true) INTERNAL ASSERT FAILED at "/opt/pytorch/nvfuser/tests/cpp/validator.h":119, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. 
00:35:38 
00:35:38 Validation error in output 0 on line 171 in file /opt/pytorch/nvfuser/tests/cpp/test_combined_inner_outer_reduction.cpp.
00:35:38   Detected abs error of: 7.15256e-06
00:35:38     absolute tolerance was set to 2.23704e-06
00:35:38     and relative tolerance set to 2.23704e-08
```

PR #2561, opened 2024-07-10
change CombinedSchedulerTest to TEST_P

liqiangxl:
Part of #2463 
Also includes a trivial fix on persistent batch size, it should be <= inner dim size.
Helps to figure out why sometimes CI fails #2560 

No validation errors in local tests on H100.


liqiangxl:
!build

tests/cpp/test_combined_inner_outer_reduction.cpp
```diff
@@ -34,255 +34,129 @@ namespace nvfuser {
...
+       std::get<2>(aten_gradients)},
+      __LINE__,
+      __FILE__);
 }
+INSTANTIATE_TEST_SUITE_P(
+    ,
+    CombinedSchedulerTest,
+    ::testing::Combine(
+        // aten doesn't support complex data types
+        testing::ValuesIn(getFloatingDataTypes(/*with complex type ?*/ false)),
```
wujingyue, line 138:
```suggestion
        testing::ValuesIn(getFloatingDataTypes(/*include_complex=*/false)),
```

tests/cpp/test_combined_inner_outer_reduction.cpp
```diff
@@ -34,255 +34,129 @@ namespace nvfuser {
...
+
+  auto maybe_fp16_options =
+      at::TensorOptions().dtype(data_type_to_aten(dtype)).device(at::kCUDA, 0);
+
+  at::Tensor aten_grad_out = at::randn(input_shape, maybe_fp16_options);
+  at::Tensor aten_input = at::randn(input_shape, maybe_fp16_options);
+  at::Tensor aten_weight = at::randn(norm_shape, maybe_fp16_options);
+  at::Tensor aten_bias = at::randn(norm_shape, maybe_fp16_options);
+
+  const float kEps = 1e-5;
```
wujingyue, line 101:
```suggestion
  constexpr float kEps = 1e-5;
```

tests/cpp/test_combined_inner_outer_reduction.cpp
```diff
@@ -34,255 +34,129 @@ namespace nvfuser {
...
+  auto [dtype, batch_size, hidden_size] = GetParam();
+
+  std::vector<int64_t> norm_shape{hidden_size};
+  std::vector<int64_t> input_shape{batch_size, hidden_size};
+  std::vector<int64_t> outer_shape{batch_size, 1};
+  bool fp16_or_bf16 = dtype == DataType::Half || dtype == DataType::BFloat16;
+  auto grad_out = makeContigTensor(input_shape.size(), dtype);
+  auto input = makeContigTensor(input_shape.size(), dtype);
+  auto mean =
+      makeConcreteTensor(outer_shape, fp16_or_bf16 ? DataType::Float : dtype);
```
wujingyue, line 53: `mean` and `rstd` can't be bfloat? Why not?

tests/cpp/test_combined_inner_outer_reduction.cpp
```diff
@@ -34,255 +34,129 @@ namespace nvfuser {
...
-        std::cout << "passed, shape= " << sdim0.str() << str1 << std::endl;
-      }
-      return std::make_tuple(-1.0f, -1.0f);
-    }
-  };
+  fusion.addOutput(layer_norm_results.grad_input);
+  fusion.addOutput(layer_norm_results.grad_weight);
+  fusion.addOutput(layer_norm_results.grad_bias);
+
+  auto maybe_fp16_options =
```
wujingyue, line 93:
```suggestion
  auto options =
```
I don't see the value of mentioning `maybe_fp16`. Obviously, `dtype` can be fp16 and `options` just follow whatever dtype that's passed in.

tests/cpp/test_combined_inner_outer_reduction.cpp
```diff
@@ -34,255 +34,129 @@ namespace nvfuser {
...
-        c10::optional<at::Tensor>(aten_weight),
-        c10::optional<at::Tensor>(aten_bias),
-        {true, true, true});
+// tuple of data type, batch size (outer dim), hidden size (inner dim)
+using CombinedSchedulerParams = std::tuple<DataType, int64_t, int64_t>;
+using CombinedSchedulerTest = NVFuserFixtureParamTest<CombinedSchedulerParams>;
+TEST_P(CombinedSchedulerTest, LayerNormBackward) {
+  std::unique_ptr<Fusion> fusion_ptr = std::make_unique<Fusion>();
+  Fusion& fusion = *fusion_ptr.get();
+  FusionGuard fg(&fusion);
```
wujingyue, line 43:
```suggestion
  auto fusion = std::make_unique<Fusion>();
  FusionGuard fg(fusion.get());
```
Then replace `fusion.` in this test with `fusion->`. 

Having two variables `fusion_ptr` and `fusion` in this case provide little value. So I prefer using one variable for simplicity -- readers don't even need to think about the difference between the two variables.

tests/cpp/test_combined_inner_outer_reduction.cpp
liqiangxl: if other inputs are `fp16_or_bf16`, `mean` and `rstd` are `float`, otherwise they stay the same as other inputs, e.g. when dtype is `double`.

liqiangxl: They can be any data type but in the current implementation of layer norm, they are saved as `float` if inputs are `half or bfloat`

wujingyue: Got it. I think the code in this function can be simplified by introducing a helper:
```
// Returns float if the input data type is lower than that; otherwise, returns the input data type as is. 
DataType floatOrHigher(DataType);
```

Then you can say 

```
mean = makeConcreteTensor(..., floatOrHigher(dtype));
```
and
```
auto cast_to_float_or_higher = [](TensorView* in) -> TensorView* {
  return maybeCastOp(floatOrHigher(in->dtype()), in);
};
grad_out = cast_to_float_or_higher(grad_out);
...
```

wujingyue (approved):

tests/cpp/test_combined_inner_outer_reduction.cpp
liqiangxl: Good point. I believe we have many similar cases in our test suites. will check and simplify them together in a separate PR.

PR #2562, opened 2024-07-10
Use `_reduce_scatter_base` instead of `reduce_scatter` when possible 

samnordmann:
`reduce_scatter` primitive in c10d induces extra buffering time to copy the user's input tensors to an internal source buffer. It is therefore always preferable to use `_reduce_scatter_base` (which does not perform any extra copy) when the the tensors are stored contiguously (e.g., when `scattered_axis==0`).

Note however than only `nccl` supports `_reduce_scatter_base`, not `ucc`.

See also, for a similar PR about allgather:
- #2384


samnordmann:
!build

csrc/multidevice/communication.cpp
```diff
@@ -398,12 +398,19 @@ c10::intrusive_ptr<c10d::Work> postReduceScatter(
...
     at::Tensor input_tensor,
     at::Tensor output_tensor) {
-  std::vector<std::vector<at::Tensor>> input_tensors(1);
   const auto scattered_axis = communication->scatteredAxis();
   NVF_ERROR(
       scattered_axis >= 0,
       "scattered_axis is expected to be non-negative: ",
-      scattered_axis)
+      scattered_axis);
+#if defined(NVFUSER_DISTRIBUTED) && defined(USE_C10D_NCCL)
```
wujingyue, line 406:
```suggestion
// reduce_scatter primitive in c10d induces extra buffering time to copy the user's input tensors to an internal source buffer. It is therefore always preferable to use _reduce_scatter_base (which does not perform any extra copy) when the tensors are stored contiguously (i.e., when scattered_axis==0).
// Note however than only nccl supports _reduce_scatter_base, not ucc.
#if defined(NVFUSER_DISTRIBUTED) && defined(USE_C10D_NCCL)
```

wujingyue: Nice comment in the PR description! It's even nicer when sitting with the code.

wujingyue (approved):

cowanmeg (approved):

Issue #2563, opened 2024-07-10
DID parallelization of loop domain


cowanmeg:
Currently, DID parallelization is only supported on the logical domain which requires us to manually add extra axes in the logical domain when parallelizing Fusions across multiple devices. 

This has caused some PRs to introduce temporary hacks to work around logical DID axes or code that needs to be modified which is tracked here.

wujingyue:
As we discussed in the meeting, here's a breakdown of the task: 
- [x] https://github.com/NVIDIA/Fuser/issues/3282. ETC: 3 weeks. Assigned to @wujingyue 
- [x] Fix communication lowering to support DID split. ETC: 2 weeks. Assigned to @wujingyue 
- [x] Fix ATen-evaluated nodes, e.g. linear, matmul and sdpa. ETC: 2 weeks
- [ ] Fix all schedulers. According to [the nsys traces](https://docs.google.com/spreadsheets/d/1XvKcvhrIqmxsoQMErH8-pcWVBOwlmbHNJlfQBS4NeVM/edit?gid=0#gid=0) @cowanmeg collected, we'll need to fix at least pointwise, *_persistent (aka normalization), transpose and reduction. FWIW, forward needs pointwise and persistent only. ETC: 1 month
- [x] Fix sharding propagation to propagate DID splits, e.g., using TransformPropagator. ETC: 1 month 
- [x] Fix other sharding-related pre-segmenter passes, e.g., insert_reshardings and reorder_sharded_axis. One example is to decompose a DID-involving reduction (e.g. Reduction, Matmul and Linear) into multiple ops. This is conceptually similar to the existing reduction rfactoring. ETC:  1 month
- [ ] Remove tests that rely on logical split of DID and clean up dead code. ETC: 2 weeks

These subtasks are roughly execution ordered. #3282 pretty much blocks everything else, so it's currently my top priority. After that's done, some other tasks (e.g. 2-4) can be parallelized as long as we limit the scope of testing. 

cc @naoyam, @cowanmeg and @samnordmann to take a look

cowanmeg:
Yes, I actually experimented a bit with that in this very old and out of date PR, https://github.com/NVIDIA/Fuser/pull/2181
When we explicitly set the allocation domain of the sharded tensors to the loop domain (before the schedulers applied intra-device parallelization) allocation worked.

wujingyue:
> It's probably necessary to update the allocation logic of gmem tensors.

[...]

PR #2565, opened 2024-07-10
Sharded SDPAFwdOp

cowanmeg:
Temporarily add support for a sharded forward scaled dot product attention.
Currently, we only support DID parallelization on the logical domain which requires us to split and parallelize an axis at the logical level see https://github.com/NVIDIA/Fuser/issues/2563. This is a hack until we support DID parallelization on the loop domain after which this PR can be reverted.

Restrictions:
1. q,k,v inputs are manually sharded _before_ the SDPAFwdOp is created. We cannot rely on sharding propagation or sharding after the Fusion is created, because the dimension checks are called when the op is created. 
2. Only the head dimension is sharded and all inputs and outputs have either a sharded head dimension or unshaded. 
3. DID axis is the outermost axis. This is because during evaluation if we see 5 dimensions, it is assumed the first is the DID axis and is appropriately squeezed from the inputs and unsqueezed onto the outputs.

### DIFF
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -4351,10 +4351,11 @@ std::vector<PolymorphicValue> SdpaFwdOp::evaluate(
     return padded_inp;
   };
 
-  // Temporary hack to handle sharding the head dimension
-  // on the logical domain.
+  // Temporary handling of DID parallelization see
+  // https://github.com/NVIDIA/Fuser/issues/2563
   bool handle_device_dim = false;
   if (query.dim() == 5) {
+    NVF_CHECK(key.dim() == 5 && value.dim() == 5);
     handle_device_dim = true;
     query = query.squeeze(0);
     key = key.squeeze(0);
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -4391,15 +4391,14 @@ std::vector<PolymorphicValue> SdpaFwdOp::evaluate(
               /*return_debug_mask=*/false,
               scale);
 
-  // Add back the device dim axis for outputs with a head dimension.
-  if (handle_device_dim) {
-    output = output.unsqueeze(0);
-    log_sumexp = log_sumexp.unsqueeze(0);
-  }
-
   // If the inputs were padded, slice the output to restore the original size
   if (output.sizes()[3] != last_dim_size) {
     output = output.slice(-1, 0, last_dim_size);
+  }
+
+  // Add back the device dim axis for output.
+  if (handle_device_dim) {
+    output = output.unsqueeze(0);
   }
 
   // Query and key seq len are of type c10::SymInt -> convert them to CPU scalar
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -4351,6 +4351,16 @@ std::vector<PolymorphicValue> SdpaFwdOp::evaluate(
     return padded_inp;
   };
 
+  // Temporary hack to handle sharding the head dimension
+  // on the logical domain.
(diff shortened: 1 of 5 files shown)


csrc/ir/nodes.cpp
```diff
@@ -4380,6 +4391,12 @@ std::vector<PolymorphicValue> SdpaFwdOp::evaluate(
               /*return_debug_mask=*/false,
               scale);
 
+  // Add back the device dim axis for outputs with a head dimension.
+  if (handle_device_dim) {
+    output = output.unsqueeze(0);
+    log_sumexp = log_sumexp.unsqueeze(0);
+  }
+
```
Priya2698, line 4399: This should be after L4401-4402 since dim=3 is assumed to be last dimension. Or `output.sizes()[3]` should be changed to `output.sizes()[-1]`

csrc/ir/nodes.cpp
```diff
@@ -4380,6 +4391,12 @@ std::vector<PolymorphicValue> SdpaFwdOp::evaluate(
               /*return_debug_mask=*/false,
               scale);
 
+  // Add back the device dim axis for outputs with a head dimension.
+  if (handle_device_dim) {
+    output = output.unsqueeze(0);
+    log_sumexp = log_sumexp.unsqueeze(0);
```
Priya2698, line 4397: Why do we need the DID axis in `log_sumexp`?

cowanmeg: Good question, we only need it for the backwards, so we don't need to track the DID axis. I'll remove it!

csrc/root_domain_map.cpp
```diff
@@ -227,32 +227,39 @@ std::unordered_map<IterDomain*, IterDomain*> PairwiseRootDomainMap::map(
...
+    //   key = [DIDx(D)?, N, H, S, E]
+    //   value = [DIDx(D)?, N, H, S, Ev]
     // Consumers:
-    //   output = [N, H, L, Ev]
-    //   logsumexp = [N, H, L]
+    //   output = [DIDx(D)?, N, H, L, Ev]
+    //   logsumexp = [DIDx(D)?, N, H, L]
     //   cum_seq_q/k = [N + 1]
 
+    unsigned int num_device_dim = producer_logical.at(0)->isDeviceDim() ? 1 : 0;
```
wujingyue, line 241:
```suggestion
    size_t num_device_dim = producer_logical.at(0)->isDeviceDim() ? 1 : 0;
```

It's better for num_device_dim and idx to have the same type so comparisons are unambiguous.

csrc/root_domain_map.cpp
```diff
@@ -227,32 +227,39 @@ std::unordered_map<IterDomain*, IterDomain*> PairwiseRootDomainMap::map(
...
       }
       // Map Ev from value to output
-      if (idx == 3 && producer_tv_->sameAs(op->value())) {
+      if (idx == (3 + num_device_dim) && producer_tv_->sameAs(op->value())) {
         updatePairwiseRootDomainMap(
-            producer_logical.at(3), consumer_root.at(3));
+            producer_logical.at(4), consumer_root.at(4));
       }
     }
+    // Map D from any input (query/key/value) to output, logsumexp only.
```
wujingyue, line 259: Still needed given https://github.com/NVIDIA/Fuser/pull/2565/files#r1672758030?

wujingyue (approved):
Nice! I'd wait for Priya's approval.

cowanmeg:
!build

Priya2698 (approved):
LGTM.

cowanmeg:
Note: failing tests are unrelated and caused by `'nlohmann/json.hpp' file not found`

Issue #2564, opened 2024-07-10
nvfuser python API should be able to forward aliased output


jjsjann123:
```
import torch
from nvfuser import FusionDefinition, DataType

def nvfuser_fusion_id0(fd : FusionDefinition) -> None :
    T0 = fd.define_tensor(shape=[-1, -1, -1, -1], contiguity=[True, True, True, True], dtype=DataType.Float, is_cpu=False, stride_order=[3, 2, 1, 0])
    S1 = fd.define_scalar(0.00000, dtype=DataType.Double)
    T2 = fd.ops.gt(T0, S1)
    S3 = fd.define_scalar(0.00000, dtype=DataType.Double)
    T4 = fd.ops.where(T2, T0, S3)
    fd.add_output(T4, T0)
    fd.add_output(T4)
 
with FusionDefinition() as fd:
    nvfuser_fusion_id0(fd)
 
inputs = [
    torch.randn((200,), dtype=torch.float32, device='cuda:0').as_strided((2, 4, 5, 5), (100, 25, 5, 1)),
]
 
print(fd.execute(inputs))
```

Today nvfuser is not returning `T4` in the program, even though the application specifically asked for it.

jjsjann123:
I thought someone from lightning has filed an issue on this but I cannot find it. linking the issue where it strikes again: https://github.com/Lightning-AI/lightning-thunder/pull/749

PR #2566, opened 2024-07-10
fixing output aliased tv

jjsjann123:
Fixes: #2564 

allows nvfuser to return aliased tv output.
The updated behavior is that, when `addOutput` tries to return an aliased output, we just need to remove the existing entry in `outputs_`, if it exists.

The updated API involves:
```
  addOutput(Val* output, bool hide_output = false);
  aliasOutputToInput(Val* output, Val* input);
```

The expected behavior:
1. when user calls `aliasOutputToInput`, the intention is just to insert an inplace update. Hence we are not trying to add `output` to the actual return of the fusion.

2. when user calls `addOutput(output, hide_output)`.
2.1 if hide_output == false (default), we expect the output to be reflected as part of the functional return. we will update io_alias_[output].hide_output to false if it exists; we'll also remove the existing entry of `output` in `outputs_` (since it's only used for inplace update), and appending output to the right position.
2.2 if hide_output == true, we know that we are only adding output for alias, so we'll just append it as-is.


csrc/fusion.cpp
```diff
@@ -259,26 +259,28 @@ void Fusion::addInput(Val* input) {
...
 
-void Fusion::addOutput(Val* output) {
-  // We currently don't support explicitly outputing aliased inputs. This is
-  // because they are already marked as output for in-place update. It's tricky
-  // to allow marking them explicitly as real output, since that requires us to
-  // register/identify output not only by `Val*` pointer, but also by indices;
-  // it also requires us to magically arrange `outputs_` entries in proper order
-  // ^^^ this doesn't look intuitive on `outputs_` in fusion.
-  // I think we can solve this by marking addOutput on io_alias_ keys after
-  // fusion is fully defined. Tracking this in #1488
```
jjsjann123, line 270: #1488 doesn't seem to be related to this.

tests/python/test_python_frontend.py
```diff
@@ -4432,6 +4432,28 @@ def fusion_func(fd: FusionDefinition):
...
+        ]
+
+        def fusion_func(fd: FusionDefinition):
+            T0 = fd.define_tensor(shape=[-1, -1, -1, -1], contiguity=[True, True, True, True], dtype=DataType.Float, is_cpu=False, stride_order=[3, 2, 1, 0])
+            S1 = fd.define_scalar(0.00000, dtype=DataType.Double)
+            T2 = fd.ops.gt(T0, S1)
+            S3 = fd.define_scalar(0.00000, dtype=DataType.Double)
+            T4 = fd.ops.where(T2, T0, S3)
+            fd.add_output(T4)
+            fd.add_output(T4, T0)
```
jjsjann123, line 4447: I think our python API looks bad here (I'm the one to be blamed).

We should have really renamed this to `aliased_update` or something like that.

We can revisit this later. For the time being, this fixes the thunder issue that @crcrpar hit in https://github.com/Lightning-AI/lightning-thunder/pull/749

jjsjann123:
!build

csrc/fusion.cpp
Priya2698: Can you explain how the concerns in this comment are addressed:
```
  // since that requires us to
  // register/identify output not only by `Val*` pointer, but also by indices;
  // it also requires us to magically arrange `outputs_` entries in proper order
  // ^^^ this doesn't look intuitive on `outputs_` in fusion.
```
I do not have the context behind re-arranging outputs.

wujingyue: I believe it refers to https://github.com/csarofeen/pytorch/issues/1488

csrc/fusion.h
```diff
@@ -147,7 +147,7 @@ class NVF_API Fusion : public IrContainer {
   void addInput(Val* input);
 
   //! Register output as an output of the fusion
-  void addOutput(Val* output);
+  void addOutput(Val* output, bool hide_output = false);
```
wujingyue, line 150: How do you expect this to collaborate with Fusion::aliasOutputToInput? Fusion::aliasOutputToInput changes the hide_output bit in io_alias_.

csrc/fusion.cpp
jjsjann123: Ha good catch @wujingyue totally forgot about that repo. :face_exhaling:

jjsjann123: Sorry @Priya2698 , I pinged you here because #1488 was under your name.
Updated the PR description.

csrc/fusion.h
jjsjann123: It does account for existing output though.

```
  io_alias_[output] = AliasInfo{
      .type = type,
      .aliased_io = input,
      .hide_output = !output->isFusionOutput()};

  if (!output->isFusionOutput()) {
    addOutput(output, true);
  }
```

I think I could add a couple more assert in `addOutput`.

aliasOutputToInput only needed to have the update in the output, I think the added logic PR ensures that, while still maintaining us having the *right* output entries in `outputs_`

jjsjann123: I do want to point out that `removeOutput` is not considering aliases yet. It's not too tricky to patch that. But we are also not using `removeOutput` in python API at the point. I can add that as a separate issue for follow up PR if we want to fix that.

csrc/fusion.cpp
```diff
@@ -259,26 +259,28 @@ void Fusion::addInput(Val* input) {
...
   output->as<TensorView>()->setMemoryType(MemoryType::Global);
 
+  // special handling for returning aliased output. We just need to remove its
+  // existing entry in the outputs_ used for inplace update
+  if (!hide_output) {
+    if (io_alias_.count(output) != 0) {
+      // if previous output is only added for aliasing purpose, we should remove
+      // the previous entry and add a new one
+      if (io_alias_[output].hide_output) {
+        removeOutput(output);
```
jjsjann123, line 277: Here we can assert that there's only one entry of output in `outputs_`.

jjsjann123:
Oops. looks like there's a real failure with instance_norm.... Let me take a look there.

csrc/fusion.h
wujingyue: After reading https://github.com/NVIDIA/Fuser/issues/2564 (which I believe is the motivation for this PR), there's apparently a simpler fix that updates https://github.com/NVIDIA/Fuser/blob/baea1bb3b935f71cb1dead5737c6e6c572238393/csrc/python_frontend/fusion_record.h#L1397 to call addOutput before aliasOutputToInput when the user doesn't want to hide an output alias. Note that whether to hide_output is determined here: https://github.com/NVIDIA/Fuser/blob/baea1bb3b935f71cb1dead5737c6e6c572238393/csrc/fusion.cpp#L825. Wdyt?

jjsjann123:
!build

csrc/fusion_segmenter.cpp
```diff
@@ -2523,7 +2523,7 @@ class FusionSegmentGuard : public NonCopyable {
     }
 
     for (auto old_out : old_outputs_) {
-      fusion_->addOutput(old_out);
+      fusion_->addOutput(old_out, fusion_->getOutputAlias(old_out).hide_output);
```
jjsjann123, line 2526: I'm glad we caught this one.

The previous bug was that completeFusion after segmentation were artificially changing the hide_output from FusionSegmentGuard.

jjsjann123: alternatively, we should also have used `fusion_` instead of `fusion` here: https://github.com/NVIDIA/Fuser/blob/baea1bb3b935f71cb1dead5737c6e6c572238393/csrc/kernel_cache.cpp#L602

I'll clean up this part of the code in a follow up PR to remove the permutation stuff on IO.

jjsjann123: added a specific repro for the failing case.

jjsjann123:
!build

csrc/fusion.h
jjsjann123: That'll work. But that makes integration code in thunder a bit awkward.

We parse a program linearly today. `copy_` is translated to `aliasOutputToInput`. So if we traverse a program topologically, we'll always write `aliasOutputToInput` before we call `addOutput`.

Granted that this can also be worked around by special casing for `copy_` in integration, but I prefer to not have integration (or nvfuser users) to live by those restrictions.

jjsjann123:
recording offline conversation.

@wujingyue suggested moving main's `addOutput` to `addOutputHelper`. Reusing that code in this PR's `addOutput` and drop the `hide_output` flag.

That sounds like a much cleaner API. I'll refactor.

jjsjann123:
!build

jjsjann123:
ready for a round 2 @wujingyue

crcrpar:
great work. locally I checked this fixed the linked failure in thunder

csrc/fusion.cpp
```diff
@@ -824,9 +828,9 @@ void Fusion::aliasOutputToInput(
       .aliased_io = input,
       .hide_output = !output->isFusionOutput()};
```
wujingyue, line 829: Can we enforce aliasOutputToInput to happen before addOutput (if addOutput is used at all for unhiding the output)? This would simplify the logic here to something like
```suggestion
      .hide_output = true};
  addOutputHelper(output);
```
and possibly other functions.

jjsjann123: I'm not a big fan on adding restriction on python API. i.e. less restriction == less customer facing doc and more freedom on integration. 

But we can throw an informative error message on that, so I guess this is still a fair conversation. But how strong are you on this? I'm trying to push it back unless you have a firm stance :stuck_out_tongue_winking_eye:

jjsjann123: > and possibly other functions.

My argument is that, the suggested change won't actually simplify the other function by much. It's actually the subsequent `addOutput` which requires updating the `hide_output` field and removing the existing entry in outputs_.

wujingyue: SGTM as is. 

Overall, I'm unhappy with aliasOutputToInput and addOutput having overlapping responsibilities -- both may call addOutputHelper and may modify io_alias_. It's not immediately clear to the user of nvFuser how many times `out` is added in the following cases. 

```
aliasOutputToInput(out, ...);
```
=> once

```
aliasOutputToInput(out, ...);
addOutput(out);
```
=> once

```
addOutput(out);
aliasOutputToInput(out, ...);
```
=> once

```
addOutput(out);
aliasOutputToInput(out, ...);
addOutput(out);
```
=> twice

and things become even more complicated with removeOutput. 

I don't have an easy solution. Just for the record, we may consider to add a finalization phase (if we don't have already?) when building a fusion. Then, change code to do the following
1. aliasOutputToInput only marks `io_alias_` without changing the output list or modifying hide_output
2. addOutput only adds output to list without modifying io_alias_
3. During finalization: 
   ```
   for each key in io_alias_:
     if key is an output: 
       io_alias_[key].hide = false
     else:
       io_alias_[key].hide = true
       addOutput
   // This way, each key in io_alias_ are guaranteed to be an output for implementation purpose. 
   ```

This way, aliasOutputToInput and addOutput have non-overlapping responsibilities. The dirty implementation detail -- an alias has to be an output -- is handled by finalization, just once and at the end.

csrc/fusion.h
```diff
@@ -146,6 +146,9 @@ class NVF_API Fusion : public IrContainer {
   //! Register input as an input of the fusion
   void addInput(Val* input);
 
+  //! Add output to outputs_ without modifying hide_output
+  void addOutputHelper(Val* output);
```
wujingyue, line 150:
```suggestion
  void addOutputInternal(Val* output);
```
just so readers know it's an internal method that should be used with care and shouldn't be exposed to the Python API.

csrc/fusion.cpp
```diff
@@ -285,6 +273,22 @@ void Fusion::addOutput(Val* output) {
   all_tv_uses_valid_ = false;
 }
 
+void Fusion::addOutput(Val* output) {
+  // special handling for returning aliased output. We just need to remove its
+  // existing entry in the outputs_ used for inplace update
+  if (io_alias_.count(output) != 0) {
+    // if previous output is only added for aliasing purpose, we should remove
+    // the previous entry and add a new one
```
wujingyue, line 281:
```suggestion
    // the previous entry and add a new one. Otherwise, it may be positioned wrong in the output list. 
```

wujingyue (approved):

csrc/fusion.h
jjsjann123: Will do. Since I'll also need to update a few other places. I'll manually change this.

jjsjann123:
!build

jjsjann123:
!build

csrc/fusion.cpp
jjsjann123: I think this was resolved during our offline conversation.

I agree that we could use some re-naming here to avoid confusion.

`addOutput(out);` should be the only API that defines a returned value in fusion.
`aliasOutputToInput(out, ...);` should be renamed to something appropriately now since it doesn't define a returned value, though the implementation requires something added to `outputs_`.

`finalization phase` is a reasonable alternative in implementation. Although the logic in this PR is sufficient to resolve that when the APIs are called, and we do not have to defer that at the end.

PR #2568, opened 2024-07-10
Use `force_iter_type` in broadcast cases

Priya2698:
Fixes Issue #2532.

### DIFF
--- a/csrc/ops/utils.cpp
+++ b/csrc/ops/utils.cpp
@@ -350,11 +350,14 @@ IterDomain* newOutputIterDomain(
             .iter_type(iter_type.value())
             .build();
   } else {
+    if (!iter_type.has_value()){
+      iter_type = IterType::Broadcast;
+    }
     out_domain = IterDomainBuilder(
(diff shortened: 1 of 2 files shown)


Priya2698:
!build

csrc/ops/utils.cpp
```diff
@@ -350,11 +350,14 @@ IterDomain* newOutputIterDomain(
             .iter_type(iter_type.value())
             .build();
   } else {
+    if (!iter_type.has_value()){
+      iter_type = IterType::Broadcast;
+    }
```
jacobhinkle, line 355: Can you say more about when we might hit a case where iter_type is not null or broadcast?

Priya2698: Will add a comment.
Right now, I am checking if this fix works generally, and it looks like we cannot have reduction axis with extent 1. Will look at alternative solutions.

Priya2698:
Closing this, as reduction axis cannot have extent 1.

Issue #2348, opened 2024-06-05
[Python Benchmarks] Add a mode to selectively run a few inputs for functionality check


Priya2698:
Aim:
1. Add a mode `--run-subset` which samples a very small subset of input sizes to run. This allows us to run the benchmark suite quickly for a functionality check when making any code changes.
2. Optional: Add a flag to control how many inputs are sampled.

CC: @naoyam

PR #2569, opened 2024-07-10
Add option to benchmark a subset of inputs

Priya2698:
A subset of inputs can be run for each benchmark using `--benchmark-num-inputs=num_inputs`.
Allows for quick functionality test. By default, the entire suite will be run.

Closes Issue #2348.


naoyam:
Could you also update the doc with some example usage?

Priya2698:
> Could you also update the doc with some example usage?

Updated the wiki.

naoyam (approved):
LGTM

PR #2570, opened 2024-07-10
Move IdGraphIndexCompute to its own file

naoyam:
Just refactoring. No change other than moving `IdGraphIndexCompute` to its own file.


csrc/id_model/id_model_index_compute.cpp
naoyam, line 1: The file name looks redundant, but since there are already two `index_compute.cpp`, I prefer to use this name.

jacobhinkle (approved):

PR #2572, opened 2024-07-10
Add support for "Define box by rotation"

zasdfgbnm:


zasdfgbnm:
!build

tests/cpp/test_memory.cpp
```diff
@@ -797,19 +797,19 @@ TEST_F(TMAIndexingTest, Advanced) {
       LoadStoreOpType::CpAsyncBulkTensorTile);
 
   for (auto tv : {tv1, tv2}) {
-    // View the 8D tensor as 5D
+    // View the 8D tensor as 4D
```
zasdfgbnm, line 800: There is no change in the schedule for this test. The code is changed only because of "define box by rotation", the schedule is interpreted differently, so I change the order of code to make have a better match with how we understand the schedule.

zasdfgbnm:
!build

naoyam (approved):
LGTM

zasdfgbnm:
!build

zasdfgbnm:
!build

PR #2580, opened 2024-07-11
Add a script to run Thunder/nvFuser benchmarks.

wujingyue:
The script can run multiple settings and dump results to JSON files, so `pytest-benchmark compare` can easily display the difference.


jjsjann123 (approved):
awesome utility to use in our container!

:clap:

PR #2581, opened 2024-07-11
[Python Frontend] Add cache_fork to user-scheduling

rdspring1:
This PR adds `cache_fork` operation to python frontend. It is used for intermediate tensors in a fusion that are saved as an output.


rdspring1:
!build

jjsjann123 (approved):
:shipit:

PR #2583, opened 2024-07-12
skip further checks in `reductionInterferingView` if non of the view ops has ID merges

liqiangxl:
**Background**
This PR is separated from  #2437 by removing all commits using IdModels. I separate it out so we can merge it in the main branch and moving to step-2 of the group norm issue.
The work to rewrite `reductionInterferingView` using IdModels will be continued in #2437.

**---- PR description---**
**Issue:** https://github.com/NVIDIA/Fuser/issues/2375
**Fix:** When the reshape transformations from root domains to logical domains consists only split, it is safe for the reduction scheduler to merge any two reduction domains. We can skip reductionInterferingView check.

**Results:**
With this PR, the reshape of input ([N,C,H,W] -> [N, G, C/G, H, W]) can be fused with normalization into one kernel.
Group norm splits into 2 kernels instead of 3.

Following works:
(1) It also seems safe if the reshape transformations from root domains to logical domains contains a merge, but all of the merged IDs are mapped to reduction dims or iteration dims. Continued in #2437


liqiangxl:
!build

liqiangxl:
!build

liqiangxl:
!build

csrc/scheduler/registry_utils.cpp
```diff
@@ -389,12 +389,35 @@ bool requiresForwardViewReplay(Fusion* fusion, ComputeAtMap& ca_map) {
   return false;
 }
 
+namespace {
+
+bool isSplitOnly(ViewOp* view_op) {
+  for (auto expr : StmtSort::getExprsTo(
```
naoyam, line 395: This is fine now but may not be in the future. See #2612. Let's explicitly use the root domains.

naoyam (approved):
LGTM

PR #2546, opened 2024-07-08
Host irs: add support for `MatmulOp`

samnordmann:
Adds support for `MatmulOp` in `HostIrExecutor`

The implementation makes it very easy to support new ops that `ExpressionEvaluator` can handle.

Stacked on top of 
- #2533
- #2535

PR #2535, opened 2024-07-05
Host IRs: support `SliceOp`

samnordmann:
# What

add support for `SliceOp` in `HostIrExecutor`.

~~Stacked on top of #2533~~

# Why

Used in reduce-scatter overlap technic 
https://github.com/NVIDIA/Fuser/blob/bad998ae277ffc2f43fdc28dca07d01d737a1623/tests/cpp/test_multidevice_overlap.cpp#L129

# How 

We rely on `ExpressionEvaluator::evaluate` to execute the slice operation on the at::Tensor.

The `SliceOp` needs to be explicitly added to `HostIrExecutor::top_level_exprs_`, as illustrated in the provided unit test.

PR #2584, opened 2024-07-12
Host irs: Reduce-Scatter based pipelining

samnordmann:
We provide a full Host IR implementation of "Reduce-Scatter based pipelining" technique for achieving overlap between comms and compute. An Aten implementation has already been provided in
- #2401

This Host Ir implementation aims to reproduce the same behavior, up to some minor details.

By the way, we added the function `void ExpressionEvaluator::invalidate(Val*)` which removes values from the cache `known_values_`. This is necessary when using `ExpressionEvaluator` in a non-SSA context, which is the case here because of the for-loop.

Stacked on top of:
- #2535
- #2546

TODOs:
- Add multi-streaming
- Make sure there is no stream synchronization issue
- Make sure allocation (especially `tc_locally_reduced`) is not an issue. Try to solve it with aliasing
- Provide NSight profile and analysis


naoyam:
Did you forget to attach a trace result?

csrc/host_ir/executor.cpp
```diff
@@ -195,8 +195,13 @@ void HostIrExecutor::handle(ForLoop* for_loop) {
   auto stop = for_loop->stop()->value().as<int64_t>();
 
   for (auto i = start; i < stop; i += step) {
+    // invalidate i and its consumers before binding
```
wujingyue, line 198: As mentioned offline, I think this will break for cases like
```
sum = zero tensor
for i: 
  sum += Slice(input, i)
```
because the code would invalidate the accumulator when entering each iteration. 

One possible solution is to force-evaluate top-level expressions in execution order, simply using ExpressionEvaluator as a map.

csrc/ir/utils.cpp
```diff
@@ -301,6 +301,16 @@ std::vector<Val*> consumerValsOf(const Val* val) {
   return uniqueEntries<Val>(consumer_vals);
 }
 
+// Return all consumers of val
+std::unordered_set<Val*> allConsumerValsOf(const Val* val) {
```
wujingyue, line 305: I'd expect this to be implemented using some utilities from StmtSort. @naoyam might know. 

I think the current implementation is suboptimal in that it may `allConsumerValsOf` the same Val* multiple times. But I'm happy to accept the function as is if you can make it a separate host IR utility instead of an ir_util.

csrc/host_ir/executor.cpp
samnordmann: > As mentioned offline, I think this will break for cases like
> 
> ```
> sum = zero tensor
> for i: 
>   sum += Slice(input, i)
> ```
>
> because the code would invalidate the accumulator when entering each iteration.
> 

1. I fail to see how the case you mention is related to the present pr. We do not need such accumulation mechanism, so the current implementation does not face this issue. I suggest to save this concern for a future pr.
2. In general in Fuser, supporting this type of accumulation needs *aliasing*, as can be seen in [this example](https://github.com/NVIDIA/Fuser/pull/2441/files). Indeed, we need a way to differentiate the consumer and the producer in the DAG, which is hidden by using the "`+=`" operator in your example. To make this idea more precise, I imagine in the future a program like
```
sum_in = zero tensor
sum_out = aliasOf(sum_in)
for i: 
  sum_out = add(sum_in, Slice(input, i))
```
This formulation is compatible with the invalidation of the producers of `i` (which are `sum_out`, `Slice(input, i)`, but not `sum_in`), provided we do "alias lookup" for inferring the location of the output buffer, and not "evaluate".

_Incidental note: the above formulation points to the fact that `sum_out` plays two different roles: it represents the result of the compute (obtained through "evaluate"), but it also indicates what buffers stores the results (obtained through aliasing). I don't think there a general treatment of this in Fuser. I don't have a cristal clear idea on how it should be done but in general we should differentiate the definition of a TensorView in the DAG and how to access the buffer associated with it_

> One possible solution is to force-evaluate top-level expressions in execution order, simply using ExpressionEvaluator as a map.

Sorry I don't understand your suggestion. In the current example, there is one single top level expression: the for-loop.

csrc/ir/utils.cpp
samnordmann: > I'd expect this to be implemented using some utilities from StmtSort. @naoyam might know.

Ok, makes sense. FWIW, `StmtSort` is called through `allConsumerValsOf` > `consumerValsOf`.

> I think the current implementation is suboptimal in that it may `allConsumerValsOf` the same Val* multiple times.

Correct! Let me fix that.

> But I'm happy to accept the function as is if you can make it a separate host IR utility instead of an ir_util.

Ok, np, I can move it away from ir_utils, I just thought it could help others

samnordmann:
> Did you forget to attach a trace result?

I added it now

csrc/host_ir/executor.cpp
wujingyue: I was thinking about something like
```
void HostIrExecutor::handle(ForLoop* for_loop) {
  ...
  for (i = start; i < stop; i += step) {
    ee_.bind(for_loop->index(), i);
    for (Expr* expr : for_loop->body().exprs()) {
      ee_.invalidate(expr's outputs);
      dispatch(expr);
    }
  }
}
```
without traversing def-use chains.

wujingyue: > the above formulation points to the fact that sum_out plays two different roles: it represents both the result of the compute and what buffer stores the results.

Yeah, that's a good idea. Real definitions are either an allocation or an ATen (or whatever library we may use in the future) call that implicitly allocates the result tensor.

csrc/host_ir/executor.cpp
```diff
@@ -186,6 +186,36 @@ void HostIrExecutor::handle(Wait* wait) {
   works_.erase(communication);
 }
 
+namespace {
+
+void allConsumerValsOf_helper(
```
wujingyue, line 191:
```suggestion
void allConsumerValsOfHelper(
```

wujingyue: for style consistency

wujingyue (approved):

csrc/host_ir/executor.cpp
samnordmann: > I was thinking about something like
> 
> ```
> void HostIrExecutor::handle(ForLoop* for_loop) {
>   ...
>   for (i = start; i < stop; i += step) {
>     ee_.bind(for_loop->index(), i);
>     for (Expr* expr : for_loop->body().exprs()) {
>       ee_.invalidate(expr's outputs);
>       dispatch(expr);
>     }
>   }
> }
> ```
> 
> without traversing def-use chains.

I see. Unfortunately, this won't work. The problem is that there might be some intermediate computations between binding the for-loop index `j` and for-loop's body's expressions. More precisely, in our example, the first expression in the for-loop is slicing `tva_` on range `j`, `j+1`, to produce `tva_j`. If we invalidate only `j` and `tva_j`, as suggested in your comment, we won't invalidate `j+1` (which is produced by `add(j, hic->oneVal())`. In reality, there are many of those intermediate computes of indices and predicates.

Besides, doing as you suggest does not help in the situation of buffer reuse in accumulation problems, that you mentioned in the first place, so I fail to see the benefit of it.

_Incidental remark: I want to try switching to `SelectOp` instead of `SliceOp` in the hope it will make the arithmetic simpler which could benefit the semantic but also the performance...._

csrc/host_ir/executor.cpp
samnordmann: ok!

samnordmann:
## Analysis of the NSight profiles, and related TODOs for future PRs:

cc @naoyam @wujingyue

I can see three items from the analysis of the NSight profile. Each one of those items is actionable and will be addressed separately in future PRs:
 
1. An undesired `cudaMalloc`, is present in both the ATen and the HostIr implementation. It is most probably caused by `at::matmul_out` (I will double-check). We should remove it.
2. In HostIr, a second `cudaMalloc` is visible. This one was expected: it comes from the fact that we execute matmul through `ExprEvaluator` --> `at::matmul`, and not `matmul_out`, therefore a new output buffer is allocated to produce the result. We do not use `at::matmul_out` because of `ExpressionEvaluator`'s design which does not allow output buffers to be passed. This can be simply patched for HostIr/MatMulOp by bypassing `ExpressionEvaluator` completely, but IMO we should consider a general patch on `ExpressionEvaluator`. It is related to a broader issue, e.g. #2247.
3. In the HostIr profile, `ExpressionEvaluator::evaluate` causes a non-negligible overhead for computing slices, which is nothing more than an alias op. I need to investigate this and try to alleviate the overhead.

PR #2585, opened 2024-07-12
[Tutorial Multidevice]: communicator, DeviceMesh, Simple pipelining

samnordmann:
Add first multidevice tutorials!

TODO next: introduction of `ParallelType:DIDx` and simple resharding ops.


tests/cpp/test_multidevice_tutorial.cpp
```diff
@@ -0,0 +1,367 @@
...
+// The first object we need to introduce is the class `Communicator` which is
+// a convenience that is nvFuser's interface to runtime distributed setting
+// and backend. Communicator's setup is expansive (because all IP addresses
+// need to be exchanged to all ranks) and occupies a port. Therefore, we
+// instantiate one Communicator globally and use it accross all tests. There
+// is no benefit in having multiple communicator, and this class is thought to
+// have a singleton instance. This communicator can be accessed through
+// MultiDeviceTest::communicator_
+TEST_F(MultiDeviceTutorial, CommunicatorAndC10d) {
+  Communicator* communicator = communicator_;
```
naoyam, line 36: As a tutorial, it would be helpful to mention this `communicator_` is defined and setup by `MultiDeviceTest`.

tests/cpp/test_multidevice_tutorial.cpp
```diff
@@ -0,0 +1,367 @@
...
+
+  // To check if distributed setting is available:
+  if (communicator->is_available() == false) {
+    GTEST_SKIP() << "distributed setting not available";
```
naoyam, line 40: Could this be potentially a common error message users might encounter? Could we give users some more actionable error messages?

tests/cpp/test_multidevice_tutorial.cpp
```diff
@@ -0,0 +1,367 @@
...
+    verbose_ = getNvFuserEnv("TUTORIAL_VERBOSE");
+  }
+
+ protected:
+  static bool verbose_;
+};
+
+bool MultiDeviceTutorial::verbose_ = false;
+
+// The first object we need to introduce is the class `Communicator` which is
```
naoyam, line 27: Do we also want to mention how to run this tutorial tests?

tests/cpp/test_multidevice_tutorial.cpp
```diff
@@ -0,0 +1,367 @@
...
+  }
+  // During instantiation, the Communicator will parse information from the
+  // environment variable. Setting those env vars properly is what is requested
+  // from the user for multidevice to be available in nvFuser. Those environment
+  // variables are set automatically in the case of a single-node system where
+  // "mpirun" (from open MPI) is used to launch the application. The actual list
+  // of required environment variables are:
+  //  - Set automatically by mpirun (so, no need to manually set them if using
+  //  mpirun):
+  //      - OMPI_COMM_WORLD_RANK
```
naoyam, line 50: Can we use other MPI implementations that may not define `OMPI_` environment variables?

tests/cpp/test_multidevice_tutorial.cpp
```diff
@@ -0,0 +1,367 @@
...
+  // RUNTIME
+  // Set up the input
+  constexpr int64_t tensor_size = 128;
+  const c10::TensorOptions tensor_options =
+      at::TensorOptions().device(communicator->device()).dtype(at::kFloat);
+  // each rank allocate a tensor a on different device
+  at::Tensor input = at::randn({tensor_size}, tensor_options);
+  {
+    // EXECUTION
+    MultiDeviceExecutor multidevice_executor(
```
naoyam, line 219: I think it'd be helpful to mention that a single `MultiDeviceExecutor` manages a single device, i.e., a single CPU process manages a single device, which may not be obvious for users.

tests/cpp/test_multidevice_tutorial.cpp
```diff
@@ -0,0 +1,367 @@
...
+  // SINGLE DEVICE EXECUTION
+  // This indicates that tv0 and tv1 are present on device 0 only
+  tv0->setDeviceMesh(mesh_zero);
+  tv1->setDeviceMesh(mesh_zero);
+  // RUNTIME
+  {
+    // EXECUTION
+    MultiDeviceExecutor multidevice_executor(
+        std::make_unique<Fusion>(fusion), *communicator);
+    // here, the compute is done on device 0 only. Other devices don't even read
```
naoyam, line 244: Then, can we use a dummy, unallocated tensor for non-zero devices?

tests/cpp/test_multidevice_tutorial.cpp
```diff
@@ -0,0 +1,367 @@
...
+// Stages.
+TEST_F(MultiDeviceTutorial, SimplePipelining) {
+  constexpr int64_t tensor_size = 128;
+
+  Communicator* const communicator = communicator_;
+  if (communicator->is_available() == false || communicator->size() < 2) {
+    GTEST_SKIP() << "distributed setting is not available";
+  }
+  if (communicator->deviceId() > 1) { // we only need 2 devices for this test
+    return;
```
naoyam, line 271: Is this necessary? What would happen if the other processes kept running the rest of the test?

naoyam (approved):
This is really helpful. Thanks. Just added some minor comments and suggestions.

tests/cpp/test_multidevice_tutorial.cpp
samnordmann: ok, done!

tests/cpp/test_multidevice_tutorial.cpp
samnordmann: I added the following comment at the top of the first unit test:
```
// To run those tests, allocate a node with at least 2 GPUs and run:
//
// mpirun -np 2 -x TUTORIAL_VERBOSE=1 tutorial_multidevice
```

tests/cpp/test_multidevice_tutorial.cpp
samnordmann: We actually also check the same variable without the `OMPI_COMM_` prefix, check https://github.com/NVIDIA/Fuser/blob/6866230b4c4417ffefed5de8338badc4e966766b/csrc/multidevice/communicator.cpp#L59
This is supposed to be how torchrun sets env vars, so we theoretically also support torch run, and I can't forsee any issue using this alternative ; but since we don't test torchrun, it is safer to always use mpirun.

Anyway I though it was not worth mentionning this alternative in the tutorial, but lmk if you think otherwise

tests/cpp/test_multidevice_tutorial.cpp
samnordmann: ok, I added a comment

tests/cpp/test_multidevice_tutorial.cpp
samnordmann: > Then, can we use a dummy, unallocated tensor for non-zero devices?

No, we cannot, because the shape of the input is used to infer subsequent tensors' shape, therefore, we still need to give each device inputs with valid shapes.

This shall probably be refined in the future, but that's how we do it currently.

I added a comment.

tests/cpp/test_multidevice_tutorial.cpp
samnordmann: it is not necessary. I used it to make a less verbose print but I can move this predicate to where printing happens.

tests/cpp/test_multidevice_tutorial.cpp
samnordmann: I moved this check to the class's setup and provided a more actionavle error message.

tests/cpp/test_multidevice_tutorial.cpp
naoyam: Ah, I remember that's still a limitation. Thanks for expanding the comment. It's important to be clear on such limitations.

tests/cpp/test_multidevice_tutorial.cpp
naoyam: I was thinking of other MPI implementations such as MPICH since (annoyingly and IIRC) these MPI implementation failed to agree on environment names and thus use slightly different ones.

It's ok to just support OpenMPI but it should be mentioned somewhere.

samnordmann: The exact requirement is to have those envs set properly. One sufficient (but not necessary) condition is to use ompi -- if using something else, one needs to make sure those env vars are still setup properly.

I think it is clearly conveyed in the present comment but lmk your thoughts

samnordmann:
@cowanmeg @wujingyue lmk if you have comments

wujingyue (commented):
LGTM overall -- I'll try to review the details by the end of this week!

CMakeLists.txt
```diff
@@ -616,6 +616,14 @@ if(BUILD_TEST)
   add_test_without_main(test_multidevice "${MULTIDEVICE_TEST_SRCS}" "")
   list(APPEND TEST_BINARIES test_multidevice)
 
+  set(MULTIDEVICE_TUTORIAL_SRCS)
+  list(APPEND MULTIDEVICE_TUTORIAL_SRCS
+    ${NVFUSER_ROOT}/tests/cpp/multidevice.cpp
+    ${NVFUSER_ROOT}/tests/cpp/test_multidevice_tutorial.cpp
```
wujingyue, line 622:
```suggestion
    ${NVFUSER_ROOT}/tests/cpp/test_multidevice_executor_tutorial.cpp
```
Nit: at some point, I'll try to add a tutorial for FusionExecutorCache with multidevice, so I slightly prefer naming it the MultiDeviceExecutor tutorial. But we could always rename things on demand.

samnordmann: I prefer calling it "test_multidevice_tutorial". I am not focussing on the multidevice executor in the test.

I suggest when FusionExecutorCache is ready we will use it those tests

cowanmeg (approved):
LGTM! small comments

tests/cpp/test_multidevice_tutorial.cpp
```diff
@@ -0,0 +1,379 @@
...
+  static bool verbose_;
+};
+
+bool MultiDeviceTutorial::verbose_ = false;
+
+// To run those tests, allocate a node with n>1 GPUs and run:
+//
+// mpirun -np n -x TUTORIAL_VERBOSE=1 tutorial_multidevice
+//
+// We use a SPMD paradigm, where each host process manages one and only device.
```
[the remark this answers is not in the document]
samnordmann: Oh right! But we still are in a SPMD model right? IIUC, one the one hand we are SPMD, and on the other hand one-process = one device, right?

What would you write as a comment here?

tests/cpp/test_multidevice_tutorial.cpp
```diff
@@ -0,0 +1,379 @@
...
+
+  // The first scheduling piece we need is the DeviceMesh. A device mesh in
+  // itself is nothing but a multi-dimensional array of integers representing
+  // device indices. In this tutorial, we will focus on 1D Meshes, so a
+  // DeviceMesh is nothing but a vector of indices. For example, we can consider
+  // the DeviceMesh comprised of the device index "0" only
+  DeviceMesh mesh_zero({0});
+  // or the mesh with device "1" only
+  DeviceMesh mesh_one({1});
+  // We can also consider the mesh of device indices "0" "2"
```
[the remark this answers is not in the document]
samnordmann: right! Will do

samnordmann:
@wujingyue kind reminder. I have a follow-up tutorial PR after this one.

tests/cpp/test_multidevice_tutorial.cpp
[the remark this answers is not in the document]
cowanmeg: Yes! we are SPMD because each device executes the same program! I would just update to something like: We use a SPMD paradigm, where each host process manages one and only device and each device executes the same program.

tests/cpp/test_multidevice_tutorial.cpp
```diff
@@ -0,0 +1,381 @@
...
+
+class MultiDeviceTutorial : public MultiDeviceTest {
+ protected:
+  static void SetUpTestSuite() {
+    verbose_ = getNvFuserEnv("TUTORIAL_VERBOSE");
+  }
+
+  void SetUp() {
+    MultiDeviceTest::SetUp();
+    if (communicator_->is_available() == false) {
```
wujingyue, line 23:
```suggestion
    if (!communicator_->is_available()) {
```

tests/cpp/test_multidevice_tutorial.cpp
```diff
@@ -0,0 +1,381 @@
...
+  // we support go through pytorch Process Group (abbreviated as PG), and are
+  // UCC PG and NCCL PG. We can request the communicator to return a
+  // c10::Backend by specifying:
+  // - the Team, aka, the group of ranks that this backend will serve
+  std::vector<int64_t> all_devices(communicator->size());
+  std::iota(
+      all_devices.begin(),
+      all_devices.end(),
+      0); // all_devices = [0,1,..., communicator->size()-1]
+  Team& team = all_devices;
```
wujingyue, line 103:
```suggestion
  const Team& team = all_devices;
```

tests/cpp/test_multidevice_tutorial.cpp
```diff
@@ -0,0 +1,381 @@
...
+  // - (optionally) the backend type, between UCC and NCCL
+  CommunicatorBackend backend_type =
```
wujingyue, line 105:
```suggestion
  auto backend_type =
```

tests/cpp/test_multidevice_tutorial.cpp
```diff
@@ -0,0 +1,381 @@
...
+      CommunicatorBackend::nccl; // or CommunicatorBackend::ucc
+  // the backend_type is an optional argument. By default it will choose nccl if
+  // available, ucc otherwise. We can check that the requested backend is indeed
+  // available
+  if (communicator->isBackendAvailable(backend_type) == false) {
```
wujingyue, line 110:
```suggestion
  if (!communicator->isBackendAvailable(backend_type)) {
```

tests/cpp/test_multidevice_tutorial.cpp
```diff
@@ -0,0 +1,381 @@
...
+
+  // The backend can be retrieved from the the communicator, which manages a
+  // cache
+  c10d::Backend* backend = communicator->getBackendForTeam(team, backend_type);
+
+  // The c10d backend can then be used to execute collectives. This is used
+  // internally in nvFuser to execute collectives, but this is typically too low
+  // level to be exposed to the user. Let us show for convenience how this torch
+  // library is used on a simple example, e.g., "allreduce" a single buffer
+  constexpr int64_t tensor_length = 128;
```
wujingyue, line 122:
```suggestion
  constexpr int64_t kTensorLength = 128;
```

tests/cpp/test_multidevice_tutorial.cpp
```diff
@@ -0,0 +1,381 @@
...
+  }
+}
+
+// Let us now complexify the above example by showing how setting the tensor's
+// device mesh can encode pipelining using multiple devices. Let us define a
+// Fusion which will be executed in a two-stage pipeline fashion. Executing this
+// pipeling will require a network communication between the two pipeline
+// Stages.
+TEST_F(MultiDeviceTutorial, SimplePipelining) {
+  Communicator* const communicator = communicator_;
```
wujingyue, line 285:
```suggestion
```
I'd just use `communicator_`. The one-character saving isn't worth the potential confusion whether `communicator` is the underlying MultiDeviceTest::communicator_.

tests/cpp/test_multidevice_tutorial.cpp
```diff
@@ -0,0 +1,381 @@
...
+
+  // SINGLE DEVICE EXECUTION
+  // This indicates that tv0 and tv1 are present on device 0 only
+  tv0->setDeviceMesh(mesh_zero);
+  tv1->setDeviceMesh(mesh_zero);
+  // RUNTIME
+  {
+    // EXECUTION
+    MultiDeviceExecutor multidevice_executor(
+        std::make_unique<Fusion>(fusion), *communicator);
```
wujingyue, line 262: Make `fusion` a unique_ptr and do
```suggestion
        std::move(fusion), *communicator);
```
This saves a fusion copy.

samnordmann: Right, well spotted! However, this can't work as is because `fusion` is of non-pointer type `Fusion`. So I'll leave it as is for now, despite the unnecessary fusion copy. ok? lmk

wujingyue: That's why I said make fusion an uniqueptr. Would that not work?

samnordmann: https://github.com/NVIDIA/Fuser/pull/2733

PR #2586, opened 2024-07-12
set vectorization factor for each IO tensor

liqiangxl:
**Issue**: current max allowed vectorization factor is set to `min(vect factor of all IO tensors)`. When IO has both fp16 and fp32 tensors (usually happens when segmented), the `fp16` tensors are also vectorized by `4`, although they can be vectorized by `8`.
**Fix**:  This PR extends `getVectorizationFactor` to also return a  `tv_to_vectorization_factor` map, it stores the max allowed vectorization factor for each IO tensor and then the scheduler can set different vectorization factors for different IO tensors.

**Specific changes in this PR**:  
(1) `getVectorizationFactor` is extended to also return `tv_to_vectorization_factor` map
(2)  code processing vectorizations in `propagateParallelization` is moved to `getUnrolledOrVectorizedInputsOutputs`. For each IO tensor, if the vectorization factor in heuristic is larger than the allowed factor in `tv_to_vectorization_factor` map, a furter split is added. For example, a fusion with input `T0, fp16` and `T1, fp32` , we have `T0 : 8` and `T1 : 4` in the map, if the heuristic uses vectorization factor of 8, last dim of T1 is further splitted & parallelized as `S{8/4}, V{4}`
(3) `cacheAndForkOutputs` is revised to return vecto of `std::tuple<cached output, forked output, old output>` instead of `std::pair<cached output, forked output>`. The newly added `old output` is required becase the `tv_to_vectorization_factor` map  stores the old output not the newly forked ones.
(4) `reduction` and `normalization` schedulers are changed to use the max vectorization factor, other schedulers are not changed.

**Tests**: `test_vectorization.cpp` is added
Three Tests, `InnerPersistent_fp16_32_64`, `OuterPersistent_fp16_32_64`,  and  `InnerOuterPersistent_fp16_32_64`, all are:
```
  // Fusion with 3 inputs and 3 outputs
  // inputs: half, float, double
  // outputs: half, float, double
  // all are vectorized as 16 bytes, so the vectorization factors are 8, 4,
  // and 2.
```

**Benchmarks**:  
(1) original fusion in #2146 
```
(1) If enable this feature, inner outer normalization reduced from 325 us to 300 us
(2) If enable this feature, outer reduction kernel increased from 268 us to 340 us. May further reduce to 270 us if the max `vect*unroll` is adjustedf rom 16 to 8 since there are 4 inputs in this sub-fusion.
```


naoyam:
> last dim of T1 is further splitted & parallelized as S{8/4}, V{4}

Doesn't this mean each memory transaction will be strided? What does the performance look like?

liqiangxl:
> > last dim of T1 is further splitted & parallelized as S{8/4}, V{4}
> 
> Doesn't this mean each memory transaction will be strided? What does the performance look like?

(1) Yes, it is strided, similar to dumping tmp results of fp32 to gmem in iter gouped grid reduction vectorized by 8 in  #2203 
(2) We don't have benchmarks designed for this case, working on this. Generally, it should depends on the fusion, e.g. if there is only 1 fp32 IO tensor and many other fp16 IO tensors, using a vectorization of 8 is probabally better than 4. However if there is only 1 fp16 but many fp32 IO tensors, it probabally better to use vectorization 4. It should be part of the heuristics.

liqiangxl:
Checked performance in #2146:
```
(1) If enable this feature, inner outer normalization reduced from 325 us to 300 us
(2) If enable this feature, outer reduction kernel increased from 268 us to 340 us.
```
Seems depends on fusion and should leave it to the heuristics.
More benchmarks will be added later.

PR #2589, opened 2024-07-12
SDPA Bwd Mapping

Priya2698:
This PR:

1. Removes `cum_seq_q/k` from `SdpaFwdOp` outputs and `SdpaBwdOp` inputs. These are undefined tensors for non-nested inputs which are not supported as fusion inputs. The nvFuser API now differs from torch/Thunder.
2. Adds pairwise domain mapping and scheduling support for `SdpaBwdOp`.
3. Allows tensor views marked as CPU scalars to be meta tensors. Detailed reasoning in code comments.

### DIFF
--- a/csrc/expr_evaluator.cpp
+++ b/csrc/expr_evaluator.cpp
@@ -54,7 +54,9 @@ void validateValWithConcreteValue(
         ", but got a tensor of dtype ",
         actual_dtype);
     // Intermediate tensorviews marked as CPU scalars will be created as meta
-    // tensors during compilation.
+    // tensors during compilation. For example, for fusions containing SDPA fwd
+    // and bwd, some outputs of the fwd op (philox seed, philox offset) are CPU
+    // scalars.
     if (tv->isCpuScalar()) {
       NVF_CHECK(
           is_cpu_scalar(t) || is_meta_scalar(t),
--- a/csrc/expr_evaluator.cpp
+++ b/csrc/expr_evaluator.cpp
@@ -55,10 +55,10 @@ void validateValWithConcreteValue(
         actual_dtype);
     if (tv->isCpuScalar()) {
       NVF_CHECK(
-          t.is_meta() || is_cpu_scalar(t),
+          is_cpu_scalar(t) || is_meta_scalar(t),
           "Expected ",
           tv->toString(),
-          " to be bound to a CPU scalar tensor "
+          " to be bound to a CPU or meta scalar tensor "
           ", but got a tensor on device ",
           t.device(),
           " with ",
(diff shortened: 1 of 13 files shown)


Priya2698:
!build

Priya2698:
!build

csrc/expr_evaluator.cpp
```diff
@@ -55,10 +55,10 @@ void validateValWithConcreteValue(
         actual_dtype);
     if (tv->isCpuScalar()) {
       NVF_CHECK(
-          is_cpu_scalar(t),
+          is_cpu_scalar(t) || is_meta_scalar(t),
```
Priya2698, line 58: If an intermediate tensorview is CPU scalar, during compilation it is still created as a meta tensor since we do not yet have an actual tensor corresponding to that tensorview.

This occurs in fusions containing SDPA fwd and bwd ops.  The outputs from fwd ops -- `query_seq_len, key_seq_len, philox seed, philox offset` are CPU scalars.

If these are not marked as CPU scalars when writing the fusion, it will lead to an error in binding once the tensors have been computed. If they are marked as CPU scalars, then in the unmodified version, we run into an error in compilation since these intermediate tensors are created as meta tensors.

Priya2698:
!build

Priya2698:
!build

Priya2698:
!build

csrc/expr_evaluator.cpp
jacobhinkle: Can you summarize that in a comment at this location?

csrc/utils.cpp
```diff
@@ -117,6 +117,10 @@ bool is_cpu_scalar(const c10::TensorType& tensor_type) {
       opt_numel.value() == 1;
 }
 
+bool is_meta_scalar(const at::Tensor& tensor) {
+  return tensor.device().is_meta() && tensor.numel() == 1 && tensor.dim() == 0;
```
jacobhinkle, line 121: Is it not enough to check `tensor.device().is_meta()`?

csrc/utils.h
```diff
@@ -60,6 +60,8 @@ bool is_zero_sized_tensor(const std::shared_ptr<c10::TensorType>& tensor_type);
 bool is_cpu_scalar(const at::Tensor& tensor);
 bool is_cpu_scalar(const c10::TensorType& tensor_type);
 
+bool is_meta_scalar(const at::Tensor& tensor);
```
jacobhinkle, line 63: A short comment here might be helpful as it is not immediately obvious what a "meta scalar" is.

jacobhinkle (approved):
LGTM. Minor suggestions.

csrc/utils.cpp
jjsjann123: we do still want to verify that it's a scalar tensor, that's what the validation should do.

jacobhinkle: Oh duh. Of course! Thanks.

csrc/utils.h
jacobhinkle: Feel free to ignore this, as it's essentially just a scalar `at::Tensor` that `.is_meta()` :-)

PR #2605, opened 2024-07-16
Unit test to try schedule the store of the Mma output from shared to global memory using TMA.

protonu:


protonu:
!build

protonu:
!build

protonu:
!build

csrc/tensor_view.cpp
```diff
@@ -1370,6 +1370,36 @@ void TensorView::applyMmaSwizzleForTMALoad(
       this, swizzle, permute_outer_dim);
 }
 
+void TensorView::scheduleMmaTMAStore() {
```
zasdfgbnm, line 1373: Should we not make this a member function of `TensorView`? Unlike swizzleTMABox, which is generally useful for all TMA operations of arbitrary box size, this is too specificly tied to the 2D tensor of matmul.

csrc/tensor_view.cpp
```diff
@@ -1370,6 +1370,36 @@ void TensorView::applyMmaSwizzleForTMALoad(
...
+      definition()
+              ->as<LoadStoreOp>()
+              ->in()
+              ->as<TensorView>()
+              ->getMemoryType() == MemoryType::Shared,
+      "Producer should be in shared memory");
+
+  // [M(m), N(n)] -> [MO(1), MI(m), NO(1), NI(n)]
+  split(-2, axis(-2)->extent());
+  split(-1, axis(-1)->extent());
```
zasdfgbnm, line 1397: Why not just use `getM(macro)` and `getN(macro)`?

csrc/tensor_view.cpp
protonu: That makes sense - I was conflicted about putting it here. Would it make sense to put it in mma_utils?

zasdfgbnm: > Would it make sense to put it in mma_utils?

That sounds good

csrc/scheduler/mma_utils.cpp
```diff
@@ -1166,6 +1166,42 @@ inline void resolveTvToMatmulDimRolesMapping(
...
+
+  NVF_ERROR(
+      tv->definition()
+              ->as<LoadStoreOp>()
+              ->in()
+              ->as<TensorView>()
+              ->getMemoryType() == MemoryType::Shared,
+      "Producer should be in shared memory");
+
+  auto mma_ops = ir_utils::getOpsOfType<MmaOp>(tv->fusion());
```
zasdfgbnm, line 1191: Instead of finding it from the fusion, could we explicitly pass M and N into this function? @jacobhinkle is working on fusing multiple matmuls, so in the future, we should no longer assume there is only one matmul in the fusion.

zasdfgbnm (approved):

protonu:
!build

protonu:
!build

PR #2609, opened 2024-07-16
Avoid Reduction/Broadcast domains in ValGraph scheduling

jacobhinkle:
Currently, in `id_model/schedule.cpp` if a requested `ExprGroup` doesn't already exist we select the first element in each producer `ValGroup`, clone it without rfactor, then apply IterDomain methods `split`, `merge`, or `swizzle`, and map the result as a new consumer `ValGroup`. See #2321.

Selecting the first element of the `ValGroup` is problematic though since it might mean we try to merge an `IterType::Reduction` domain with an `IterType::Iteration` domain. Instead, this PR introduces a utility `IterDomain* representativeId(const ValGroup& vg)` which picks the "best-behaved" `IterDomain` from the group. It errors if there are no `IterDomain`s in the `ValGroup`.

I created the utility to avoid the Iteration/Reduction merge problem, but it is useful in other contexts also such as trying to extract a constant extent from a `ValGroup` where there might be some mapped symbolic extents, etc.

### DIFF
--- a/csrc/id_model/schedule.cpp
+++ b/csrc/id_model/schedule.cpp
@@ -10,10 +10,48 @@
 
 namespace nvfuser {
 
+IterDomain* representativeId(const ValGroup& vg) {
+  IterDomain* rep = nullptr;
+
+  auto preferNewIterType = [&rep](IterDomain* new_id) {
+    if (rep->isReduction()) {
+      return new_id->isBroadcast() || new_id->isIteration();
+    } else if (rep->isBroadcast()) {
+      return new_id->isIteration();
+    } else if (rep->isIteration()) {
+      return false;
+    } else {
+      // Prefer anything else to a non-iter, non-bcast, non-reduction ID
+      return true;
+    }
+  };
+
+  auto preferNewExtent = [&rep](IterDomain* new_id) {
+    if (rep->hasExpandedExtent() && !new_id->hasExpandedExtent()) {
+      // Prefer non-expanded dimensions
+      return true;
+    }
+    // Prefer constant extent IDs to symbolic
+    return !rep->getMaybeExpandedExtent()->isConstInt() &&
+        new_id->getMaybeExpandedExtent()->isConstInt();
+  };
+
+  for (Val* v : *vg) {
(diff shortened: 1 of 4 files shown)


jacobhinkle:
!build

csrc/id_model/schedule.cpp
```diff
@@ -10,10 +10,48 @@
 
 namespace nvfuser {
 
+IterDomain* representativeId(const ValGroup& vg) {
```
zasdfgbnm, line 13: As a better engineering improvement, should we change (not in this PR) https://github.com/NVIDIA/Fuser/blob/2e293a32017d11aa872a7ec1a362e7fc9640f156/csrc/id_model/id_model.cpp#L743-L765
to use this function, or change `IdModel::addReplayAs` a whole as using utilities in schedule.{h,cpp}? cc @naoyam

csrc/id_model/schedule.cpp
```diff
@@ -10,10 +10,48 @@
...
+    } else if (rep->isIteration()) {
+      return false;
+    } else {
+      // Prefer anything else to a non-iter, non-bcast, non-reduction ID
+      return true;
+    }
+  };
+
+  auto preferNewExtent = [&rep](IterDomain* new_id) {
+    if (rep->hasExpandedExtent() && !new_id->hasExpandedExtent()) {
```
zasdfgbnm, line 30: If we have:

ID1: extent 1, expanded extent 10
ID2: extent 1, no expanded extent

are we preferring ID2?

zasdfgbnm (approved):

csrc/id_model/schedule.cpp
jacobhinkle: Good point, I might have had this backwards. If there are only `Broadcast` domains in the group and one is expanded, the expanded one carries more information. At first I was viewing this as "which IterDomain is most compatible with transforms" and thought expanded extents could potentially cause problems. But I think any such problems have been fixed for a while and it's better to return those expanded IDs so that we could inspect their extents. I'll reverse this condition.

PR #2612, opened 2024-07-16
Rename `rfactor` -> `producer projection` or `logical`

zasdfgbnm:


zasdfgbnm:
!build

zasdfgbnm:
!build

naoyam (approved):
The name change makes sense, but I wonder if there's any alternative name that's as short as "root" and "rfactor".

zasdfgbnm:
> The name change makes sense, but I wonder if there's any alternative name that's as short as "root" and "rfactor".

I can not think of any good alternative name for "root domain", but for this "is_rfactor", other options could be "is_semantical" or "is_math", denoting that this ID defines "what to compute", instead of "how to compute". What is your thought? @naoyam

PR #2614, opened 2024-07-16
Initial support of indexing for predicates

naoyam:
Adding an initial support of indexing for predicates. Similar to tensor indexing, `NVFUSER_ENABLE=id_model(inline_predicate)` switches to the new indexer for inline predicates. 

The overall algorithm is quite similar to tensor indexing. The main difference is that, instead of indexing allocation domains, predicates need to index the logical domains of the consumer tensor of the expression to predicate. See `getPredicateDomains` in predicate_indexing.cpp. 

I only added two unit tests, partly because I'm lazy but also this only works with very limited cases yet. Unswich/unroll will make the logic a little more complicated, which will be done in next PRs.

### DIFF
--- a/csrc/id_model/indexing.cpp
+++ b/csrc/id_model/indexing.cpp
@@ -1023,7 +1023,6 @@ std::vector<PredicateInfo> TensorIndexer::getInlinePredicates(
 
   std::vector<PredicateInfo> info_vec;
   info_vec.reserve(predicate_domains.size());
-  std::unordered_set<ValGroup> already_indexed_domains;
 
   // Follow the same approach as Index::getReferenceRootPredicates.
   for (const auto& predicate_domain : predicate_domains) {
--- a/csrc/id_model/indexing.cpp
+++ b/csrc/id_model/indexing.cpp
@@ -14,6 +14,7 @@
 #include <id_model/id_model_index_compute.h>
 #include <id_model/indexing.h>
 #include <id_model/indexing_utils.h>
+#include <id_model/predicate_indexing.h>
 #include <id_model/to_string.h>
 #include <index_compute.h>
 #include <ir/builder.h>
@@ -29,23 +30,6 @@
 namespace nvfuser {
 
 namespace {
-
-// Get the promotion domain of a given loop domain.
-IterDomain* getLoopPromotion(IterDomain* loop_id, const IdModel& id_model) {
-  const auto& loop_graph = id_model.idGraph(IdMappingMode::LOOP);
-  const auto& loop_promotion_map = id_model.loopPromotionMap();
-  const auto& loop_group = loop_graph.toGroup(loop_id);
-
-  auto loop_promotion_map_it = loop_promotion_map.find(loop_group);
-  NVF_ERROR(
-      loop_promotion_map_it != loop_promotion_map.end(),
-      "No loop promotion found: ",
-      loop_id->toString(),
-      ". Loop group: ",
-      nvfuser::toString(loop_group));
-
-  return loop_promotion_map_it->second;
-}
 
 // True if a given domain is a loop domain of a given tensor and its
 // loop is partitioned with respect to the memory type of the tensor
@@ -318,7 +302,8 @@ class AllocationDomainSetup : private kir::IrVisitor {
                          allocation_domain) != tv->getLoopDomain().end();
       IterDomain* promotion_domain = nullptr;
       if (is_loop) {
-        promotion_domain = getLoopPromotion(allocation_domain, id_model);
+        promotion_domain =
+            indexing_utils::getLoopPromotion(allocation_domain, id_model);
       } else {
         promotion_domain = allocation_domain;
       }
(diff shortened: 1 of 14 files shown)


naoyam:
CC: @zasdfgbnm

csrc/id_model/indexing.h
```diff
@@ -84,6 +84,15 @@ class TensorIndexer {
   // Traverse exprs and set allocation info for each tensor
   void setupAllocationDomains(const std::vector<Expr*>& exprs);
 
+  // Get the list of predicates of a given tensor appearing in a given
+  // expr as a consumer. Each predicate corresponds ot a domain of the
+  // tensor, which is by default one of the logical domain but can be
+  // an intermediate domain with contiguous indexing.
```
jacobhinkle, line 90:
```suggestion
  // Get the list of predicates of a given tensor appearing in a given
  // expr as a consumer. Each predicate corresponds to a domain of the
  // tensor, which is by default one of the logical domains but can be
  // an intermediate domain with contiguous indexing.
```

csrc/id_model/predicate_indexing.h
```diff
@@ -0,0 +1,22 @@
...
+// clang-format on
+#pragma once
+
+#include <device_lower/lower2device.h>
+#include <device_lower/utils.h>
+#include <id_model/id_model.h>
+
+namespace nvfuser {
+
+// Get the domains to prediate for a given tensor used as a consumer
```
jacobhinkle, line 16:
```suggestion
// Get the domains to predicate for a given tensor used as a consumer
```

csrc/id_model/indexing.cpp
```diff
@@ -1022,4 +1007,59 @@ void TensorIndexer::setupAllocationDomains(const std::vector<Expr*>& exprs) {
...
+      getPredicateDomains(tv, expr);
+
+  const IndexingInfo& index_info = computeIndex(
+      expr, traversalGraph().toGroups(predicate_domains), for_loops);
+
+  const auto& index_map = index_info.index_map;
+
+  std::vector<PredicateInfo> info_vec;
+  info_vec.reserve(predicate_domains.size());
+  std::unordered_set<ValGroup> already_indexed_domains;
```
jacobhinkle, line 1026: Is this unused?

csrc/id_model/predicate_indexing.cpp
```diff
@@ -0,0 +1,56 @@
...
+  predicate_domains.erase(
+      std::remove_if(
+          predicate_domains.begin(),
+          predicate_domains.end(),
+          [](IterDomain* id) -> bool { return id->isBroadcast(); }),
+      predicate_domains.end());
+
+  // If this is an expr initializing a buffer for a reduction, the
+  // reduction domains do not need to be predicated. In fact, if it's
+  // a Local tensor, no predicate is necessary at all
```
jacobhinkle, line 39: Why are predicates not needed for Local tensors only? Rather, if there is a reduction domain on a Shared or Global tensor, I don't yet see why they would need predication.

tests/cpp/test_indexing.cpp
```diff
@@ -232,6 +252,98 @@ class IndexValidator : public kir::IrVisitor {
...
+    validate(out_ti, inline_predicate->predicate()->value());
+
+    get_ref_.clearForLoops();
+    get_ref_.clearCircularBufferInfo();
+  }
+
+  void validate(kir::TensorIndex* ti, Val* actual) {
+    TensorView* tv = ti->view();
+    Val* ref = get_ref_.getPredicate(tv);
+    if (ref != nullptr) {
```
jacobhinkle, line 302: Should this be `ASSERT_TRUE(ref != nullptr)`? Otherwise if we inserted a predicate that should not be there it would silently pass.

tests/cpp/test_indexing.cpp
```diff
@@ -2291,4 +2403,140 @@ TEST_F(IndexingTest, CircularBuffering2) {
...
+          }
+        }
+        case 2: {
+          // T2_l[ iS6{i0}, rS8{4}rf, iS9{( ceilDiv(i2, 4) )}rf ]
+          //
+          // The initialization block should not be predicated at all.
+          if (is_init) {
+            return tv->fusion()->trueVal();
+          } else {
+            // Predicaing the logical domains can result in wrong
```
jacobhinkle, line 2516:
```suggestion
            // Predicating the logical domains can result in wrong
```

jacobhinkle (approved):
Mostly just a comment about why we don't always skip Reduction domains in initialization.

csrc/id_model/indexing.cpp
naoyam: Ah, that'll be eventually used but not in this PR yet. Just forgot to remove it from #2238 when I stripped out except for the bare minimum code. Removed.

csrc/id_model/predicate_indexing.cpp
naoyam: If it's a global tensor, since we usually allocate a global memory tensor with its logical domain, the loop domain may be larger than the logical domain if it has non-divisible splits. So, there may be multiple threads initializing the same location if no predication is done.

I think the same issue could happen with shared memory as well. The shape of blockDim and gridDim should be the largest of all the parallelized domains, which may be larger than the parallelized domains of the shared memory tensor. In that case, proper predicates would be needed.

This shouldn't matter in practice, though, since we shouldn't reduce to shared or global memory tensors directly.

tests/cpp/test_indexing.cpp
naoyam: This is intentional. The `GetReference` of each test should return a reference Val for any tensor that should be validated. This is convenient since not all tensors need to be verified, and for those `GetReference` can just return nullptr.

csrc/id_model/predicate_indexing.cpp
jacobhinkle: Ah ok. That makes sense. Thanks for the explanation.

PR #2575, opened 2024-07-11
FusionExecutorCache runs a communication-only fusion. 

wujingyue:
This implements Step 2 of [this plan](https://docs.google.com/document/d/1Wvh_vJYmA_LETfVNovb8mXPK95zmPOTUccauF-DX2h8/edit#heading=h.j8ae9cdb7l0y).

PR #2616, opened 2024-07-16
Make Communicator a singleton. 

wujingyue:
As a follow-up to https://github.com/NVIDIA/Fuser/pull/2575#discussion_r1676563522. 

As the comment says, it's not the best practice but acceptable -- the lifetime concern is localized to the cleanup method, which can be educated.

### DIFF
--- a/csrc/multidevice/communicator.cpp
+++ b/csrc/multidevice/communicator.cpp
@@ -179,6 +179,7 @@ Communicator::Communicator(
       master_port_(0),
       ucc_available_(false),
       nccl_available_(false) {
+  std::cerr << "Communicator constructed" << std::endl;
   // retrieves rank and communicator size
   is_available_ = parseEnv(
       rank_, size_, local_rank_, local_size_, master_addr_, master_port_);
@@ -215,6 +216,7 @@ Communicator::Communicator(
 }
 
 Communicator::~Communicator() {
+  std::cerr << "Communicator destroyed" << std::endl;
 #if defined(NVFUSER_DISTRIBUTED) && defined(USE_C10D_NCCL)
   for (auto& [key, backend] : backends_) {
     // Call shutdown before destructing a ProcessGroupNCCL as instructed by
--- a/csrc/multidevice/communicator.cpp
+++ b/csrc/multidevice/communicator.cpp
@@ -179,7 +179,6 @@ Communicator::Communicator(
       master_port_(0),
       ucc_available_(false),
       nccl_available_(false) {
-  std::cerr << "Communicator::Communicator" << std::endl;
   // retrieves rank and communicator size
   is_available_ = parseEnv(
       rank_, size_, local_rank_, local_size_, master_addr_, master_port_);
@@ -216,8 +215,6 @@ Communicator::Communicator(
 }
 
 void Communicator::cleanup() {
-  std::cerr << "Communicator::cleanup" << std::endl;
-
   store_ = nullptr;
 
 #if defined(NVFUSER_DISTRIBUTED) && defined(USE_C10D_NCCL)
--- a/csrc/multidevice/communicator.cpp
+++ b/csrc/multidevice/communicator.cpp
@@ -217,7 +217,7 @@ Communicator::Communicator(
 void Communicator::cleanup() {
   if (!is_available_) {
     TORCH_WARN(
-        "The singleton Communicator wasn't available before the cleanup. "
+        "The singleton Communicator isn't available. "
         "This is likely because Communicator::cleanup was called more than "
         "once or the instance wasn't successfully initialized.");
     return;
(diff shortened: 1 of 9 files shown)


wujingyue:
!build

csrc/multidevice/communicator.cpp
```diff
@@ -214,7 +214,9 @@ Communicator::Communicator(
 #endif
 }
 
-Communicator::~Communicator() {
+void Communicator::cleanup() {
+  store_ = nullptr;
```
jjsjann123, line 218: once we clear `store_` I don't think we currently have logic to restore it afterwards. Does that mean we need to check for `cleared()` status in `getBackendForTeam` to avoid accessing nullptr `store_`?

csrc/python_frontend/python_bindings_extension.cpp
```diff
@@ -10,5 +10,9 @@
 
 PYBIND11_MODULE(EXTENSION_NAME, m) {
   m.doc() = "nvfuser C API python binding"; // optional module docstring
+
   nvfuser::python_frontend::initNvFuserPythonBindings(m.ptr());
+
+  auto cleanup = []() -> void { nvfuser::python_frontend::cleanup(); };
+  m.add_object("_cleanup", py::capsule(cleanup));
```
jjsjann123, line 17: Please bear with my ignorance.

https://pybind11.readthedocs.io/en/stable/advanced/misc.html#module-destructors
vaguely remember the issue we had with clean up is about the nondeterministic destruction order here, so this reads like our selection to have communicator clean up to be called early is a conscious choice? Can we have a comment on that if so?

csrc/multidevice/communicator.h
```diff
@@ -126,6 +139,10 @@ class Communicator {
   }
 
  private:
+  Communicator(
+      CommunicatorBackend backend = comm_backend_default,
+      RankType server_local_rank = comm_server_local_rank_default);
```
jjsjann123, line 144: nitpick: I feel at this point, we are not using the arguments for the constructor any more.

samnordmann: Indeed we always use default values currently. I am not opposed to removing those arguments if you think that's preferable. Anyway, IMO, the Communicator's interface's final state will very much depend on what will be the multi-device API with frontend -- for now, I think it is too early to have a precise idea about that

csrc/multidevice/communicator.cpp
```diff
@@ -214,7 +214,9 @@ Communicator::Communicator(
 #endif
 }
 
-Communicator::~Communicator() {
+void Communicator::cleanup() {
```
samnordmann, line 217: Why  does it make a difference to name the method `cleanup` instead of being the destructor? As far as I can see we could use the destructor in place of cleanup, without any change of behavior, we would just use `delete Communicator::getInstance();` instead of `Communicator::getInstance().cleanup();`

If I'm missing something and we can't use the destructor, then it's maybe worth adding in the destructor a check that cleanup has been called before, and print a warning otherwise.

csrc/multidevice/communicator.h
wujingyue: > nitpick: I feel at this point, we are not using the arguments for the constructor any more.

Good point. I'll give that a shot in a separate PR. 

Singleton implies that there can only be one instance, and therefore any configs are essentially global. So likely getInstance or the constructor would have to parse flags and decide the config. getInstance is a slightly better place to do so because if later we decide to allow multiple Communicator instances we don't have to change the constructor.

csrc/multidevice/communicator.cpp
wujingyue: It's harder to detect an accidental double deletion -- it triggers an undefined behavior in C++ so the program can behave in an arbitrarily weird way. 

Communicator::cleanup gives us a better control. For example, in the new commit, I added a is_available check to at least warn about double cleanup. 

> print a warning otherwise

Good point. Even better, the new commit did `~Communicator() = delete`.

csrc/python_frontend/python_bindings_extension.cpp
wujingyue: My hope was that Communicator::cleanup is called **before** the cuda driver shutdown, which I believe happens during `atexit`. Therefore, in C++ unit tests, I'm calling it in testing::Environment::TearDown which happens before the end of main and therefore before `atexit`. 

I'm unsure when the module destructor is called (before `main` ends or after). It didn't trigger the crash at least when I ran `mpi_test.py`, so I'm guessing before. 

If your concern was about

> This approach has the potential downside that instances of classes exposed within the module may still be alive when the cleanup callback is invoked

I don't know enough to answer that... Yes, it looks scary but I'm not aware of a better option. Can you advise?

jjsjann123: > I don't know enough to answer that...

That makes the two of us. I thought there's a reason behind our choice between the two alternative and would like to have that in comments. I think among the alternatives listed in that doc, the one we went with is the better option since it has an earlier trigger.

I don't think further action is required at this point.

csrc/multidevice/communicator.cpp
jjsjann123: Probably didn't make the comment clear.
I *think* we need to assert on `is_available` at the beginning of `getBackendForTeam` to detect pre-mature cleanup (or accidental cleanup from somewhere else).

But I'm happy to see that in the follow up PR along with the API change.

jjsjann123 (approved):

csrc/multidevice/communicator.cpp
wujingyue: Added a check in getBackendForTeam. I'm trying to avoid having to add an availability check in every API -- that wouldn't scale. Adding to getBackendForTeam makes sense because it's one method that's used by many APIs.

PR #2618, opened 2024-07-17
Refactor `ValGraphBFS` to support traversing both ValGraphs and IR nodes

zasdfgbnm:
Rename `ValGraphBFS` as `BFS`, move it to `bfs.h`, and refactor it as a class template to make it support traversing both ValGraphs and IR nodes. Now there is both a `ValGraphBFS` and an `IRBFS`. They are subclass of `BFS` with different template arguments. `ValGraphBFS` is used to traverse ValGraph, and `IRBFS` is used to traverse IR nodes.

This PR also use `IRBFS` to rewrite the current `ir_utils::allIDsOf` to make sure all IDs are obtained when domains are not on a linear path. Also, it is used to modify `validateDomainEquivalence` to remove the restrictions the previous implementation has.


csrc/bfs.h
zasdfgbnm, line 1: This file is a copy-paste of `ValGraphBFS` with only minor changes

zasdfgbnm:
!build

zasdfgbnm:
!build

zasdfgbnm:
looks like there is a bug, not ready for review yet

zasdfgbnm:
!build

zasdfgbnm:
> looks like there is a bug, not ready for review yet

Fixed, should be ready now.

zasdfgbnm:
!build

zasdfgbnm:
!build

tests/cpp/test_gpu3.cpp
```diff
@@ -8296,7 +8296,52 @@ TEST_F(NVFuserTest, DecoupledDomains) {
...
+    all_ids.insert(domain[3].as<IterDomain*>());
+    domain.merge(0);
+    all_ids.insert(domain[0].as<IterDomain*>());
+    domain.split(0, 256);
+    all_ids.insert(domain[0].as<IterDomain*>());
+    all_ids.insert(domain[1].as<IterDomain*>());
+    return domain.as<IterDomain*>();
+  };
+
+  auto logical_domain = schedule();
```
naoyam, line 8335: Am I understanding correctly that this part just creates four different set of domains that all have the equivalent transformations? It would be great if you could add a short description on what behavior is tested.

csrc/ir/utils.cpp
```diff
@@ -855,14 +826,12 @@ void validateDomainEquivalence(
       "Duplicated entry is detected in dom1: ",
       toDelimitedString(dom1));
 
-  std::vector<Val*> dom0_val(dom0.begin(), dom0.end());
-  std::vector<Val*> dom1_val(dom1.begin(), dom1.end());
-  auto forward_exprs = DependencyCheck::getAllExprsBetween(dom0_set, dom1_val);
-  auto backward_exprs = DependencyCheck::getAllExprsBetween(dom1_set, dom0_val);
+  auto exprs = IRBFS::getExprsBetween(
```
naoyam, line 829: This is nice

csrc/val_graph_visitor.h
```diff
@@ -120,139 +120,80 @@ class ValGraphStmtSort : public ValGraphVisitor {
...
-enum class Direction { Forward, Backward, Undefined };
-
-std::ostream& operator<<(std::ostream&, const Direction);
+class ValGraphDefinitions {
+  const ValGraph& graph_;
 
-using ExprPath = std::vector<std::pair<ExprGroup, Direction>>;
+ public:
+  ValGraphDefinitions(const ValGraph& graph) : graph_(graph) {}
+  decltype(auto) operator()(const ValGroup& val_group) const {
```
naoyam, line 128: Why doesn't just `auto` instead of `decltype(auto)` work here?

csrc/ir/nodes.cpp
```diff
@@ -3650,6 +3650,61 @@ void TensorDomain::setAllocationDomain(
   contiguity_ = std::move(new_contiguity);
 }
 
+std::vector<IterDomain*> TensorDomain::allIDs() const {
+  std::array<const std::vector<IterDomain*>*, 4> all_domains = {
```
naoyam, line 3654: nit: these four sets of domains would commonly have some redundancy. For example, root and logical would be the same for many tensors. Doing some filtering seems like a quick optimization so that we won't do `getExprsBetween` with `logical` and `allocation`, and also `root` and `allocation`.

csrc/val_graph_visitor.h
zasdfgbnm: `decltype(auto)` faithfully preserves the declare type, `auto` does not. For example:

```C++
int x;

const int& a() {
    return x;
}

auto b() { // returns int
    return a();
}

decltype(auto) c() { // return const int&
    return a();
}

int main() {
    auto d = a();  // int
    decltype(auto) e = a();  // const int&
}
```

naoyam: TIL. Thanks

naoyam (approved):
LGTM

zasdfgbnm:
!build

tests/cpp/test_gpu3.cpp
zasdfgbnm: Yes, correct. Comment added.

csrc/ir/nodes.cpp
zasdfgbnm: Isn't the
```C++
      if (all_domains[j]->empty()) {
        continue;
      }
```
sufficient?

naoyam: You're right. Missed that.

PR #2621, opened 2024-07-17
Host irs: add multistreaming to rs based pipelining

samnordmann:
# What

Adds multi-streaming in the host IR implementation of Reduce-scatter based pipelining that is introduced in
- #2584

It addresses item 4 of this [TODO list](https://github.com/NVIDIA/Fuser/pull/2584#issuecomment-2238833771)

Stacked on top of:
- #2665
# Why

Multi-streaming is central for achieving overlap between comms and compute

# How

We need to change stream at each iteration of the for-loop. To do so, we need a way to refer to the `Stream` using the dynamic value of the for-loop's index. Since this dynamic value is only available at runtime, we make a `Stream` point to a `Val`, which gonna be "expression evaluated" at runtime. This is the only way I can figure out how to refer to a stream dynamically.

### Stacked on top of
- #2584


csrc/host_ir/executor.cpp
```diff
@@ -79,6 +79,11 @@ std::vector<at::Tensor> HostIrExecutor::runWithInput(
 
 void HostIrExecutor::handle(SetCurrentStream* set_current_stream) {
   Stream* stream = set_current_stream->stream();
+  if (Val* index = stream->index(); index != nullptr) {
+    auto value = expr_evaluator_.evaluate(index);
+    NVF_ERROR(value.hasValue() && value.is<int64_t>());
+    stream = (Stream*)(value.as<int64_t>());
```
samnordmann, line 85: This is not very clean but that's the simplest way I can think of to achieve using either a `Stream*` or an `int64_t` as a key to the hashtable `streams_`. Maybe using `std::variants` or `dynamic_type` with a custom hash would be preferable, but more cumbersome, so lmk what you think is best

samnordmann:
!build

samnordmann:
!build

csrc/host_ir/host_ir.h
```diff
@@ -117,13 +117,23 @@ class PostOnStream : public Expr {
 
 class Stream : public Val {
  public:
-  Stream(IrBuilderPasskey passkey);
+  // if index is provided, the IR represents the streams whose index is the
+  // dynamic value of that index. Otherwise, it statically represents a new
+  // Stream.
+  Stream(IrBuilderPasskey passkey, Val* index = nullptr);
   Stream(const Stream* src, IrCloner* ir_cloner);
   bool sameAs(const Statement* other) const override;
```
wujingyue, line 125: Should `sameAs` now return true if indices are the same and are not null?

csrc/host_ir/executor.h
```diff
@@ -86,7 +86,8 @@ class HostIrExecutor final : public OptInDispatch {
   // Cache Fusions, FusionExecutors
   std::unordered_map<HostUnit*, FusionExecutor> fe_;
   std::unordered_map<HostUnit*, FusionExecutorCache> fec_;
-  std::unordered_map<Stream*, c10::cuda::CUDAStream> streams_;
+  using StreamKey = std::variant<int64_t, Stream*>;
```
wujingyue, line 89: I'm personally not a big fan of std::variant because it often ends up mixing two things in one data structure. In this particular case, it's probably easier to understand and debug to keep the original `Stream*=>CUDAStream` mapping and add another `std::unordered_map<int64_t,Stream*>` or `std::vector<Stream*>` (if indices are compact) for only the named streams.

wujingyue (commented):
LGTM otherwise

csrc/host_ir/host_ir.h
samnordmann: fixed

csrc/host_ir/executor.h
samnordmann: I understand, however I personally prefer using it here. I applied your suggestion in [the last commit](https://github.com/NVIDIA/Fuser/pull/2621/commits/1d7be069fcffbc789f1dd2411ae21b7bf4441056), but I would personally revert as it makes the code more cumbersome and duplicated.

Also, even though I get your point, we are already relying a lot on `std::variant`, through dynamic type lib and `PolymorphicValue`

Let me know whether to keep the last commit.

csrc/host_ir/executor.h
```diff
@@ -87,7 +87,8 @@ class HostIrExecutor final : public OptInDispatch {
   // Cache Fusions, FusionExecutors
   std::unordered_map<HostUnit*, FusionExecutor> fe_;
   std::unordered_map<HostUnit*, FusionExecutorCache> fec_;
-  std::unordered_map<Stream*, c10::cuda::CUDAStream> streams_;
+  std::unordered_map<Stream*, c10::cuda::CUDAStream> streams_by_pointer_;
+  std::unordered_map<int64_t, c10::cuda::CUDAStream> streams_by_index_;
```
wujingyue, line 91: Sorry if https://github.com/NVIDIA/Fuser/pull/2621#discussion_r1692592230 wasn't clear. I meant for it to be a `std::unordered_map<int64_t, Stream*>`. This way, we'll still be able to locate `Stream*`s by indices.

csrc/host_ir/executor.h
```diff
@@ -87,7 +87,8 @@ class HostIrExecutor final : public OptInDispatch {
   // Cache Fusions, FusionExecutors
   std::unordered_map<HostUnit*, FusionExecutor> fe_;
   std::unordered_map<HostUnit*, FusionExecutorCache> fec_;
-  std::unordered_map<Stream*, c10::cuda::CUDAStream> streams_;
+  std::unordered_map<Stream*, c10::cuda::CUDAStream> streams_by_pointer_;
```
wujingyue, line 90:
```suggestion
  std::unordered_map<Stream*, c10::cuda::CUDAStream> stream_to_cuda_stream_;
```
This makes clear to the readers what keys/values are, without having to let them find the definition.

csrc/host_ir/executor.h
wujingyue: Yes, PolymorphicValue uses std::variant and it comes with a cost. For example, I found it hard to gdb the value and had to deal with dynamic types in a statically-typed language. 

Unlike dynamic_type, I failed to see large enough benefits to justify the cost. So I'm inclined to avoid std::variant.

csrc/host_ir/executor.h
samnordmann: Sorry, can you clarify what you suggest? I am not sure to understand.

csrc/host_ir/executor.h
samnordmann: The cost is the diff of this commit: https://github.com/NVIDIA/Fuser/pull/2621/commits/3f84fa6fafa8d074931dbe99f6e819772bff462b

The code is more cumbersome, and duplicated in many points. Also, IMO, harder to debug.

Conversely, I don't see the benefit of not using `std::variant` in the present case. Even when using gdb, having to deal with one key and one hashtable is easier IMO, even if the key is of variant type (but with one single print we can see what type and value it is, without having to guess ti what hashtable the key belongs)

csrc/host_ir/executor.h
wujingyue: I now understood in your implementation there can be multiple `Stream*`s mapping to the same index, so a map from index to a `Stream*` doesn't make sense here. I ran into a different mental model where `Stream*`s are cached, and therefore confused myself.

wujingyue: That also justifies the use of std::variant, so reverting 3f84fa6fafa8d074931dbe99f6e819772bff462b sounds good!

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -308,6 +308,13 @@ TEST_F(OverlapTest, ReduceScatterBasedPipeliningHostIrImplementation) {
   for (Expr* expr : loop_body) {
     for_loop->body().push_back(expr);
   }
+  if (params.use_different_streams) {
+    for_loop->body().insert(
+        /*pos=*/0,
+        IrBuilder::create<hir::SetCurrentStream>(
+            IrBuilder::create<hir::Stream>(j)));
```
wujingyue, line 315: I'd add a test to exercise that the same integer maps to the same CUDA stream. This test doesn't do that IIUC. Removing `j` would have the same effect -- creating a new stream each time the program enters the loop body.

wujingyue (approved):

csrc/host_ir/executor.h
samnordmann: ok, sounds good

tests/cpp/test_multidevice_overlap.cpp
samnordmann: > I'd add a test to exercise that the same integer maps to the same CUDA stream. This test doesn't do that IIUC. 

Correct, I can add a test for that, Let me just clarify that the overlap technic doesn't need this feature.

> Removing `j` would have the same effect -- creating a new stream each time the program enters the loop body.

No, only one stream would be created, on the first iteration, corresponding to the one and only `Stream*` present in the fusion. Then, on each iteration, the same cuda stream would be set, over and over again.

wujingyue: > Then, on each iteration, the same cuda stream would be set, over and over again.

Yes, that's right -- I missed that.

PR #2625, opened 2024-07-17
Make nvfuser.__version__ less unknown when "git" cannot find the commit id 

nWEIdia:
Rely on the NVFUSER_VERSION environment variable to set the gitSHA, when "git" command fails to find the commit sha, e.g. in an environment where "git" is not found.

### DIFF
--- a/tools/gen_nvfuser_version.py
+++ b/tools/gen_nvfuser_version.py
@@ -11,6 +11,7 @@ nvfuser_root = Path(__file__).parent.parent
 
 # note that this root currently is still part of pytorch.
 def get_sha() -> str:
+    import os
     # assume the $NVFUSER_VERSION is in sha form
     nvfuser_version = os.environ.get('NVFUSER_VERSION')
     try:
--- a/tools/gen_nvfuser_version.py
+++ b/tools/gen_nvfuser_version.py
@@ -11,10 +11,6 @@ nvfuser_root = Path(__file__).parent.parent
 
 # note that this root currently is still part of pytorch.
 def get_sha() -> str:
-    import os
-
-    # assume the $NVFUSER_VERSION is in sha form
-    nvfuser_version = os.environ.get("NVFUSER_VERSION")
     try:
         return (
             subprocess.check_output(["git", "rev-parse", "HEAD"], cwd=nvfuser_root)
@@ -22,7 +18,11 @@ def get_sha() -> str:
             .strip()
         )
     except Exception:
-        return nvfuser_version if nvfuser_version is not None else UNKNOWN
+        import os
+        # assume the $NVFUSER_VERSION is in sha form
+        if nvfuser_version := os.environ.get("NVFUSER_VERSION"):
+            return nvfuser_version
+        return UNKNOWN
 
 
 def get_version() -> str:
--- a/tools/gen_nvfuser_version.py
+++ b/tools/gen_nvfuser_version.py
@@ -12,8 +12,9 @@ nvfuser_root = Path(__file__).parent.parent
 # note that this root currently is still part of pytorch.
 def get_sha() -> str:
     import os
+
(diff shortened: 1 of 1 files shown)


xwang233 (approved):
I was looking at pytorch's version logic here https://github.com/pytorch/pytorch/blob/main/tools/generate_torch_version.py# but that looks overcomplicated.

If the logic and precedence in this PR works for you, go for it.

tools/gen_nvfuser_version.py
```diff
@@ -11,14 +11,18 @@
...
+    nvfuser_version = os.environ.get("NVFUSER_VERSION")
     try:
         return (
             subprocess.check_output(["git", "rev-parse", "HEAD"], cwd=nvfuser_root)
             .decode("ascii")
             .strip()
         )
     except Exception:
-        return UNKNOWN
+        return nvfuser_version if nvfuser_version is not None else UNKNOWN
```
wujingyue, line 25: The priority of sources isn't clear at the first glance. Consider the following: 

```
completed = subprocess.run(..., check=False, capture_output=True)
if completed.returncode == 0:
  return completed.stdout.decode(...)

if nvfuser_version := os.environ.get("NVFUSER_VERSION"):
  assert nvfuser_version is indeed in sha form
  return nvfuser_version

return UNKNOWN
```

wujingyue: This way, it's immediately clear to readers that the function prioritizes the git report over the env var.

wujingyue (approved):

tools/gen_nvfuser_version.py
```diff
@@ -11,14 +11,18 @@
 
 # note that this root currently is still part of pytorch.
 def get_sha() -> str:
+    import os
+
+    # assume the $NVFUSER_VERSION is in sha form
```
wujingyue, line 16: Is it easy to check that? It seems quite easy to miss this and pass something like 0.2.3 instead.

xwang233: I don't think we have to assert it's in SHA format? Pytorch doesn't "assert" that either.

nWEIdia: Not quite straightforward to me, but I will give it a try.

tools/gen_nvfuser_version.py
```diff
@@ -11,14 +11,18 @@
...
+    nvfuser_version = os.environ.get("NVFUSER_VERSION")
```
jjsjann123, line 17: nitpick: It's probably just a bad variable name that we used in our script. But using `NVFUSER_VERSION` as sha sounds strange...

Is this just the timestamp that we attach in our nightly release? should we rename this maybe?

jjsjann123: I think an alternative is in the `get_version()` function below, when `get_sha()` returns UNKNOWN, we can use the environment variable in its absence. That's probably an easier justification.

jjsjann123 (approved):
Just some nitpick for the implementation.

I don't want to block if this PR is fixing an existing issue. So stamping for now.

tools/gen_nvfuser_version.py
nWEIdia: Agree that the variable name is very confusing to assume an sha format. The environment we have is that git will not work, but NVFUSER_VERSION (this could probably be renamed) inherited the approximate SHA from build pipeline.

tools/gen_nvfuser_version.py
nWEIdia: I would be surprised if someone has set NVFUSER_VERSION environment variable while trying to build nvfuser from source. 
Yeah Xiao is right, I am trying to do a version name override. By default, I would get version+gitUnknown when building nvfuser from an environment that git is not available.

PR #2628, opened 2024-07-17
Promote hoisting of vectorize predicates

naoyam:
Currently, we generate predicates for vectorization as:

```
// 4-way vectorization
if (idx(i + 3) < N) {
  local_tensor[0] = loadGlobalToLocal(gmem_tensor[idx(i)]);
}
```

where `idx(x)` returns an index of the tensor with the vectorized loop index as `x`. The predicate is to make sure that the access of the 4-way vectorized load is within the extent. 

This should be equivalent to:

```
// 4-way vectorization
if (idx(i) < N) {
  local_tensor[0] = loadGlobalToLocal(gmem_tensor[idx(i)]);
}
```

Notice the predicate is the same as the index for the load access. I belive that these two patterns should be equivalent since we assert that the vectorized domain is divisible by the vectorization factor.

The latter pattern is likely to promote more effective hoisting as the indices of the predicate and the tensor access are the same. Also, I think we can simplify the predicate logic for vectorization.

I tested the change with all the C++ tests and the python benchmarks using compute-sanitizer on H100. No invalid memory access happened, although sanitizer may not be perfect and also some of the matmul tests couldn't run due to out-of-memory errors (likely because sanitizer needs more memory).

Anyone has any concern?

### DIFF
--- a/csrc/device_lower/analysis/index_compute.cpp
+++ b/csrc/device_lower/analysis/index_compute.cpp
@@ -383,8 +383,14 @@ IndexingParameters getPredicateInitialIndexParameters(
       } else {
         // Similar to the above, loop_id()->extent() is
         // used here instead of loop->stop(). See the above comment.
-        loop_to_ind_map[loop] = SimplifyingIrBuilder::subExpr(
-            loop_id->extent(), GpuLower::current()->kernel()->oneVal());
+        if (unswitch_pred) {
+          loop_to_ind_map[loop] = SimplifyingIrBuilder::subExpr(
+              loop_id->extent(), GpuLower::current()->kernel()->oneVal());
+        } else {
+          // For vectorize, zero should be fine as well and that would
+          // promote better index hoisting
+          loop_to_ind_map[loop] = GpuLower::current()->kernel()->zeroVal();
+        }
       }
 
       // When predicating a loop at the maximum end, predicate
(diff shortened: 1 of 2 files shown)


naoyam:
!build --pybench

naoyam:
!build

zasdfgbnm (approved):
I think this makes sense. I don't think `idx(i + 3) < length` is more correct than `idx(i) < length`. If the schedule is correct, whatever the schedule is, these two predicates should be equivalent. If the schedule is incorrect, then `idx(i + 3) < length` does not provide additional safety over `idx(i) < length`.

naoyam:
!build

naoyam:
@liqiangxl What's the repro command to get the performance regression?

liqiangxl:
> @liqiangxl What's the repro command to get the performance regression?

For a single case you can use 
```
NVFUSER_PROF=print NVFUSER_DUMP=cuda_to_file,ptxas_verbose,scheduler_params,launch_param pytest /opt/pytorch/nvfuser/benchmarks/python/test_layernorm_bwd.py -k "float16 and 16384_16384" -v -s  --disable-benchmarking 2>&1 |tee 1.log
```
For multiple cases, I ran it locally and upload to dashboard and compare with nightly runs.
Or you can also use `!build --pybench` to launch python benchmark and the CI will share a link to visualize the comparison of PR performance vs nightly run. But only 1/6 of all the cases are benchmarked.

PR #2635, opened 2024-07-18
FusionExecutorCache runs sharding-related preseg passes. 

wujingyue:


wujingyue:
!build

wujingyue:
!build

wujingyue:
!build

wujingyue:
!build

csrc/evaluator_common.cpp
```diff
@@ -334,15 +334,20 @@ void PrecomputedValues::bindTensorMetaData(
...
   for (const auto dim : c10::irange(logical_domain.size())) {
-    auto value = tensor.size((int64_t)dim);
-    if (logical_domain[dim]->hasExpandedExtent()) {
-      auto extent = logical_domain[dim]->extent();
-      auto expanded_extent = logical_domain[dim]->expandedExtent();
+    IterDomain* id = logical_domain[dim];
+    auto dim_size = tensor.size(static_cast<int64_t>(dim));
+    if (id->isDeviceDim()) {
+      dim_size = tv->getDeviceMesh().size();
+    }
```
wujingyue, line 341: This is the only non-cosmetic change in this file. Bind the unsharded size when the dim is sharded.

csrc/executor_kernel_arg.h
```diff
@@ -65,6 +65,14 @@ class KernelArgumentHolder {
     return arguments_.at(ind).get();
   };
 
+  auto begin() const {
+    return arguments_.begin();
+  }
+
+  auto end() const {
+    return arguments_.end();
+  }
```
wujingyue, line 74: This is to allow using range-based for-loops on a non-const KernelArgumentHolder.

csrc/preseg_passes/insert_reshardings.cpp
```diff
@@ -49,7 +49,11 @@ void insertReshardingsBefore(Fusion* fusion) {
       continue;
     }
 
-    auto output = expr->outputs().at(0)->as<TensorView>();
+    if (!expr->output(0)->isA<TensorView>()) {
+      continue;
+    }
+    auto output = expr->output(0)->as<TensorView>();
```
wujingyue, line 55: An Expr can produce scalar outputs. Since we run InsertReshardings unconditionally, many existing tests would fail without this change.

csrc/preseg_passes/propagate_shardings.cpp
```diff
@@ -19,20 +19,13 @@ void PropagateShardingsPass::runPass(Fusion* fusion) {
   for (Expr* expr : fusion->exprs()) {
     const auto& inputs = ir_utils::filterByType<TensorView>(expr->inputs());
     const auto& outputs = ir_utils::filterByType<TensorView>(expr->outputs());
-    if (inputs.empty()) {
+
+    auto i = std::find_if(
```
wujingyue, line 23: Single-device fusions don't have DeviceMeshes so the old code needs to be fixed. I moved the validation to the end.

csrc/tensor_metadata.cpp
```diff
@@ -343,20 +363,30 @@ std::vector<PolymorphicValue> GetMetaData::evaluate(
...
       std::get<PrimDataType>(aten_to_data_type(input.scalar_type()).type);
   metadata->data = input.data_ptr();
-  // If tensor is sharded then logical_size and logical_stride will
-  // refer to size and stride of the sharded tensor.
-  metadata->logical_size = input.sizes();
-  metadata->logical_stride = input.strides();
+
+  if (isSharded(tv)) {
+    auto [unsharded_sizes, unsharded_strides] =
+        unshardedSizesAndStrides(tv, input.sizes(), input.strides());
```
wujingyue, line 369: This partially reverts https://github.com/NVIDIA/Fuser/pull/1932.

tests/cpp/test_multidevice_sharding.cpp
```diff
@@ -186,47 +186,6 @@ TEST_F(MultideviceShardingTest, LayerNorm) {
       __FILE__);
 }
 
-TEST_F(MultideviceShardingTest, ReduceScatter_Allgather) {
```
wujingyue, line 189: Moved to test_multidevice_lower_communication

csrc/executor_kernel_arg.h
wujingyue: Separated to https://github.com/NVIDIA/Fuser/pull/2780.

csrc/preseg_passes/insert_reshardings.cpp
wujingyue: I'll try to separate changes in this file to another PR with tests. Will keep you posted...

wujingyue:
I moved many changes to https://github.com/NVIDIA/Fuser/pull/2783. I'll rebase once that's merged so this PR is easier to review.

tests/cpp/test_multidevice_sharding.cpp
```diff
@@ -196,50 +212,33 @@ TEST_F(MultideviceShardingTest, ReduceScatter_Allgather) {
...
+      fec.fusion(),
+      {out_tensor},
       {in_tensor},
-      {unsharded_in_tensor.sum(0)},
+      {expected_out_tensor},
       __LINE__,
       __FILE__);
 }
 
-INSTANTIATE_TEST_SUITE_P(
```
wujingyue, line 228: Cosmetic change: I moved this closer to the `TEST_P`s.

wujingyue:
> I moved many changes to #2783. I'll rebase once that's merged so this PR is easier to review.

Done -- this is ready for review.

wujingyue:
!build

csrc/evaluator_common.cpp
wujingyue: No, this would break. We are lacking APIs for multi-dimensional meshes, which https://github.com/NVIDIA/Fuser/pull/2260 tried to add and got deprioritized. However, I see your concern that this `getDeviceMesh().size()` call will **quietly** the program when we start to support multi-dimensional meshes. I don't have a great solution for that. 

cc @cowanmeg

cowanmeg: I think there's aspects of that PR we can pull out to make the DeviceMesh a bit more future proof. 
In the future we need something like getDeviceMesh.size(DIDx|DIDy|DIDz) that returns the size of the mesh in one dimension. 
We can pre-emptively parameterize DeviceMesh::size with parallel type, then we only have to update size's definition when multi-dimensional meshes are supported.

csrc/tensor_metadata.cpp
```diff
@@ -323,30 +323,63 @@ inferAndValidateAllocationSizesAndStrides(
...
+namespace {
+std::pair<std::vector<int64_t>, std::vector<int64_t>> unshardedSizesAndStrides(
+    TensorView* tv,
+    c10::IntArrayRef sizes,
+    c10::IntArrayRef strides) {
+  std::vector<int64_t> unsharded_sizes(sizes.size());
+  std::vector<int64_t> unsharded_strides(strides.size());
+  for (const auto i : c10::irange(sizes.size())) {
+    if (tv->getLogicalDomain()[i]->isDeviceDim()) {
+      unsharded_sizes[i] = tv->getDeviceMesh().size();
```
naoyam, line 335: I think this is another one that would break with multidimensional device meshes.

csrc/evaluator_common.cpp
naoyam: Something like fail-safe assertions would be enough for now, but I think we should make sure to avoid silent failures.

samnordmann: > Something like fail-safe assertions would be enough for now, but I think we should make sure to avoid silent failures.

agreed

> In the future we need something like getDeviceMesh.size(DIDx|DIDy|DIDz) that returns the size of the mesh in one dimension. We can pre-emptively parameterize DeviceMesh::size with parallel type, then we only have to update size's definition when multi-dimensional meshes are supported.

Other solution: `DeviceMesh::size` could return a tuple of ints. This has the advantage of not involving the parallel types and let us have an arbitrary number of dimensions for DeviceMeshes

wujingyue:
!build

naoyam (approved):
LGTM. Thanks for the update.

csrc/tensor_metadata.cpp
```diff
@@ -323,30 +323,64 @@ inferAndValidateAllocationSizesAndStrides(
...
+  std::vector<int64_t> unsharded_sizes(sizes.size());
+  std::vector<int64_t> unsharded_strides(strides.size());
+  for (const auto i : c10::irange(sizes.size())) {
+    IterDomain* id = tv->getLogicalDomain()[i];
+    if (id->isDeviceDim()) {
+      unsharded_sizes[i] = tv->getDeviceMesh().size(id->getParallelType());
+      // This probably doesn't matter in practice unless a kernel accidentally
+      // tries to access the data on another rank. To be safe, set the stride
+      // to zero, analogous to an expanded broadcast dimension.
+      unsharded_strides[i] = 0;
```
cowanmeg, line 340: Using 0 is a good idea. I think it makes it more explicit the axis isn't allocated.

cowanmeg (approved):
Sorry for the delay reviewing, LGTM!

PR #2639, opened 2024-07-19
MarkAliasesPrepare attempts to put segment_set after intermediate tensors.

### LINKED ISSUE #2599: Suboptimal segmentation for RoPE
# TL;DR

the RoPE module in Llama-2 (and possibly other configs as well) is over-segmented into six pointwise kernels. Two should suffice. This issue blocks https://github.com/Lightning-AI/lightning-thunder/pull/731, although we don't necessarily have to reach two to unblock. 

# To reproduce

```
$ pwd
/opt/pytorch/lightning-thunder
$ git checkout wjy/bookend
$ NVFUSER_DUMP=segmented_fusion pytest thunder/benchmarks/targets.py -k test_litgpt_qkv_split_rope[Llama-2-13b-hf-forward-bs2-thunder] -s | grep 'g{(pointwise' | wc -l
6
```

# Problem

The following figure shows the current suboptimal segmentation

RoPE

The highlighted ops form the three pointwise kernels for Q. There's a similar pattern for K that's omitted for simplicity, so there are in total six pointwise kernels. The remaining, meta ops are expectedly segmented out as "no-op" regions. 

Note: For now, I omitted the `Pad`s before a `Cat` for simplicity. Soonish, https://github.com/NVIDIA/Fuser/pull/2373, pending, may move `Pad`s upstream, affecting segmentation. 

There are two main reasons we reached this state. 
1. The segmenter enforces a segment boundary before `Slice` and `Pad` (and thus before `Cat`). 
2. The `Squeeze` and the red `ToFloat` were merged too early. This made it impossible to keep merging the red and the green segment, because `Squeeze` is a transitive input to the `Cat` and the merge would lead to a cycle. Several sub-reasons contributed to that order: 
   - The merging runs [in topological order](https://github.com/NVIDIA/Fuser/blob/83e6de805e42b3063adc20eefcd02cfd5e8c1959/csrc/fusion_segmenter.cpp#L3733-L3734). 
   - The segmenter merges one edge at a time. When attempting to merge into the green `+` its operands, the segmenter only merged the right operand instead of both. 
   - A newly merged segment is pushed to [the end](https://github.com/NVIDIA/Fuser/blob/83e6de805e42b3063adc20eefcd02cfd5e8c1959/csrc/fusion_segmenter.cpp#L745) of the worklist.

wujingyue:
# Summary

This PR improves segmentation around meta ops for #2599. It mimics Thunder's bookend optimization, although only the output end is implemented at this moment. Although the implementation isn't complete, I believe it's an improvement overall. It segments out meta ops more aggressively than before, making non-meta ops easier to schedule. 

# Potential problems

1. `segment_set` enforces the segmenter to give up scheduling the complete fusion and instead trying to merge from singletons. In theory, merging should be able to reach a "local maxima". But in practice it's not always as powerful as expected, e.g., https://github.com/NVIDIA/Fuser/pull/2639/files#diff-cf6d314e82c49485e9039e700a04a678904e5c669abcd1610ade897ca8f57c78R1472. 
2. MarkAliasesPreparePass may set an suboptimal layout, because it runs before `AllocationDomainPass` and reasons about only meta ops. For example, 
   ```
   // `in` has a column-major input matrix. 
   x = PointwiseUnary(in);
   out = Transpose(x);
   ```
   MarkAliasesPreparePass will make `x` row-major, the default layout, and `out` column-major. Although `Transpose` is made no-op, `PointwiseUnary` becomes a pointwise kernel with transposition rather than a streamline pointwise kernel. I think this can be fixed by combining the allocation-domain part of MarkAliasesPreparePass into AllocationDomainPass and of course running AllocationDomainPass before. 

# Performance comparison with TOT

https://gist.github.com/wujingyue/afdc3e89e693ff6bbc18271c96409b94 is the performance comparison before the PR and after. This is collected by running the following commands. 

```
$ python tools/benchmark_thunder.py --storage ~/workspace --filter='test_litgpt_qkv_split_rope[phi-2-backward-bs1-thunder] or test_litgpt_gelu[phi-2-backward-bs1-thunder] or test_batch_norm[backward-thunder]' main:main main:bug2599
$ pytest-benchmark --storage=$HOME/workspace compare 0011 0012 --group-by=name
```

Most benchmarks are neutral, which is as expected because bookend is still on. Several benchmarks (namely, `test_litgpt_qkv_split_rope[phi-2-backward-bs1-thunder]`, `test_litgpt_gelu[phi-2-backward-bs1-thunder]` and `test_batch_norm[backward-thunder]`) showed some regression. However, the regression disappeared the second time I ran the benchmarks, so it's likely a noise.

### DIFF
--- a/csrc/preseg_passes/mark_aliases_prepare.cpp
+++ b/csrc/preseg_passes/mark_aliases_prepare.cpp
@@ -7,12 +7,33 @@
 // clang-format on
 #include <alias_analysis.h>
 #include <debug.h>
+#include <ir/iostream.h> // for operator<<(ostream&, TensorView*)
 #include <ir/utils.h>
 #include <ops/alias.h>
 #include <options.h>
 #include <preseg_passes/mark_aliases_prepare.h>
 
 namespace nvfuser::preseg_passes {
+
+namespace {
+
+std::pair<TensorView*, Expr*> findUseToSegment(
+    TensorView* out,
+    const AliasAnalysisResult& analysis,
+    const std::unordered_set<Expr*>& used_by_non_aliases) {
+  Expr* user = nullptr;
+  while (true) {
+    Expr* def = out->definition();
+    if (analysis.getNearestAliasedIo(out) == nullptr ||
+        used_by_non_aliases.count(def)) {
+      return {out, user};
+    }
+    out = def->input(0)->as<TensorView>();
+    user = def;
+  }
+}
+
+} // namespace
 
 void MarkAliasesPreparePass::runPass(Fusion* fusion) {
   const AliasAnalysisResult analysis =
@@ -50,82 +71,78 @@ void MarkAliasesPreparePass::runPass(Fusion* fusion) {
     }
   }
 
-  std::vector<Val*> non_alias_outs;
-  non_alias_outs.reserve(fusion->outputs().size());
-  // Fusion inputs/outputs that are (1) aliased by another output and (2) not
-  // an alias itself. Code will later add `segment_set` around them so aliases
-  // are separated from non-aliases and more likely to be accepted by the no-op
-  // scheduler. See AliasTest.OutputAliasesAnotherOutput and
-  // AliasTest.SegmentMetaOps.
-  //
-  // This algorithm is suboptimal in many cases. See
-  // https://github.com/NVIDIA/Fuser/issues/2395#issuecomment-2207043749 for a
-  // proposal that may work better.
-  std::unordered_set<TensorView*> aliased_ios;
-  for (TensorView* out_tv :
-       ir_utils::filterByType<TensorView>(fusion->outputs())) {
-    if (TensorView* aliased_io = analysis.getNearestAliasedIo(out_tv)) {
-      if (analysis.getNearestAliasedIo(aliased_io) == nullptr) {
-        aliased_ios.insert(aliased_io);
+  auto used_by_non_aliases = [&]() -> std::unordered_set<Expr*> {
+    std::vector<Val*> non_aliases;
+    for (TensorView* tv : ir_utils::allTvs(fusion)) {
+      if (analysis.getNearestAliasedIo(tv) == nullptr) {
+        non_aliases.push_back(tv);
       }
-    } else {
-      non_alias_outs.push_back(out_tv);
+    }
+    // Mark all expressions that are (transitively) used by non-alias outputs.
+    std::vector<Expr*> used_by_non_aliases = StmtSort::getExprsTo(non_aliases);
+    return {used_by_non_aliases.begin(), used_by_non_aliases.end()};
+  }();
+
+  std::vector<std::pair<TensorView*, Expr*>> uses_to_segment;
+  uses_to_segment.reserve(fusion->outputs().size());
+  for (auto* out : ir_utils::filterByType<TensorView>(fusion->outputs())) {
+    std::pair<TensorView*, Expr*> use_to_segment =
+        findUseToSegment(out, analysis, used_by_non_aliases);
+    if (use_to_segment.first != out) {
+      uses_to_segment.push_back(use_to_segment);
+    }
+  }
+  std::sort(uses_to_segment.begin(), uses_to_segment.end());
+  uses_to_segment.erase(
+      std::unique(uses_to_segment.begin(), uses_to_segment.end()),
+      uses_to_segment.end());
+
+  if (isDebugDumpEnabled(DebugDumpOption::PreSegmenterLogging)) {
+    for (const auto& [use_of, user] : uses_to_segment) {
+      debug() << "Will put a segment_set at " << user << std::endl;
     }
   }
 
-  // Mark all expressions that are (transitively) used by non-alias outputs.
-  auto used_by_non_aliases =
-      [](const std::vector<Expr*>& exprs) -> std::unordered_set<Expr*> {
-    return {exprs.begin(), exprs.end()};
-  }(StmtSort::getExprsTo(non_alias_outs));
+  auto i = uses_to_segment.begin();
+  while (i != uses_to_segment.end()) {
+    TensorView* use_of = i->first;
+    auto j = i;
+    while (j != uses_to_segment.end() && j->first == use_of) {
+      j++;
+    }
 
-  for (TensorView* aliased_io : aliased_ios) {
-    // Divide the users of aliased_io into two groups according to
-    // `used_by_non_aliases`.
-    std::vector<Expr*> users_used_by_non_aliases;
-    std::vector<Expr*> users_used_only_by_aliases;
-    for (Expr* e : aliased_io->uses()) {
-      if (used_by_non_aliases.count(e)) {
-        users_used_by_non_aliases.push_back(e);
+    auto insert_segment_set = [&]() {
+      if (static_cast<size_t>(std::distance(i, j)) == use_of->uses().size()) {
+        // Put a `segment_set` before `use_of`.
+        if (use_of->isFusionInput()) {
+          // A `segment_set` before a fusion input is useless.
+          return;
+        }
+        // Rarely, if `aliased_io` is already defined by `segment_set`, don't
+        // create another `segment_set`.
+        if (LoadStoreOp* def =
+                dynamic_cast<LoadStoreOp*>(use_of->definition())) {
+          if (def->opType() == LoadStoreOpType::SegmenterSet) {
+            return;
+          }
+        }
+        use_of->cacheBefore(LoadStoreOpType::SegmenterSet);
       } else {
-        users_used_only_by_aliases.push_back(e);
-      }
-    }
-
-    // If all users are used by non-alias outputs, do nothing. Adding
-    // segment_set around unlikely creates more no-op regions.
-    if (users_used_only_by_aliases.empty()) {
-      continue;
-    }
-
-    // If all users are used by aliases, put a `segment_set` before it.
-    if (users_used_by_non_aliases.empty()) {
-      if (aliased_io->isFusionInput()) {
-        // A `segment_set` before a fusion input is useless.
-        continue;
-      }
-      // Rarely, if `aliased_io` is already defined by `segment_set`, don't
-      // create another `segment_set`.
-      if (LoadStoreOp* def =
-              dynamic_cast<LoadStoreOp*>(aliased_io->definition())) {
-        if (def->opType() == LoadStoreOpType::SegmenterSet) {
-          continue;
+        // Some users of `use_of` are used by non-aliases, and some are used
+        // only by aliases. We put a `segment_set` after `use_of` and redirect
+        // the latter group to use the `segment_set`.
+        TensorView* copy = segment_set(use_of);
+        std::for_each(i, j, [&](const std::pair<TensorView*, Expr*>& use) {
+          ir_utils::replaceValInExprInputs(use.second, use_of, copy);
+        });
+        if (use_of->isFusionOutput()) {
+          fusion->replaceOutput(use_of, copy);
         }
       }
-      aliased_io->cacheBefore(LoadStoreOpType::SegmenterSet);
-      continue;
-    }
+    };
+    insert_segment_set();
 
-    // Some users of `aliased_io` are used by non-aliases, and some are used
-    // only by aliases. We put a `segment_set` after `aliased_io` and redirect
-    // the latter group to use the `segment_set`.
-    TensorView* copy = segment_set(aliased_io);
-    for (Expr* e : users_used_only_by_aliases) {
-      ir_utils::replaceValInExprInputs(e, aliased_io, copy);
-    }
-    if (aliased_io->isFusionOutput()) {
-      fusion->replaceOutput(aliased_io, copy);
-    }
+    i = j;
   }
 
   if (isDebugDumpEnabled(DebugDumpOption::PreSegmenterLogging)) {
--- a/csrc/preseg_passes/mark_aliases_prepare.cpp
+++ b/csrc/preseg_passes/mark_aliases_prepare.cpp
@@ -33,6 +33,15 @@ std::pair<TensorView*, Expr*> findUseToSegment(
   }
 }
 
+bool isSegmentSet(Expr* e) {
+  if (LoadStoreOp* ldst = dynamic_cast<LoadStoreOp*>(e)) {
+    if (ldst->opType() == LoadStoreOpType::SegmenterSet) {
+      return true;
+    }
+  }
+  return false;
+}
+
 } // namespace
 
 void MarkAliasesPreparePass::runPass(Fusion* fusion) {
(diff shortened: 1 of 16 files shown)


wujingyue:
!build

jjsjann123:
it's conflicting with @liqiangxl 's earlier PR :stuck_out_tongue: 

Do you mind a rebase :bow:

wujingyue:
!build

wujingyue:
> Do you mind a rebase 🙇

Done! cc @jjsjann123

wujingyue:
!build

wujingyue:
!build

tests/cpp/test_gpu_view.cpp
```diff
@@ -2379,6 +2398,63 @@ TEST_F(GpuViewTest, SplitMergePointwiseSplitMerge) {
   testValidate(executor_cache.fusion(), {cg_outputs}, {t0}, __LINE__, __FILE__);
 }
 
+// segmented into 2 kernels: pointwise and reduction
+TEST_F(GpuViewTest, GroupNormOriginal) {
```
wujingyue, line 2402: Copied from https://github.com/NVIDIA/Fuser/pull/2405.

tests/cpp/test_gpu_view.cpp
```diff
@@ -1447,61 +1447,77 @@ TEST_F(GpuViewTest, FusionSumViewSchedule) {
 
 // Make sure matching reshapes are segmented into the same kernel
 TEST_F(GpuViewTest, FusionReshapeMagicSchedule1) {
-  auto fusion_ptr = std::make_unique<Fusion>();
-  Fusion& fusion = *fusion_ptr.get();
-  FusionGuard fg(&fusion);
+  auto fusion = std::make_unique<Fusion>();
+  FusionGuard fg(fusion.get());
```
wujingyue, line 1451: Sorry for mixing in some cosmetic changes like this. Happy to revert if they are annoying enough.

tests/cpp/test_resize.cpp
```diff
@@ -2000,7 +2001,8 @@ TEST_F(ResizeTest, ResizeReshapeAndSlice) {
       tv1,
       {{IrBuilder::create<Val>(0L), IrBuilder::create<Val>(2L)},
        {IrBuilder::create<Val>(0L), IrBuilder::create<Val>(2L)}});
-  fusion->addOutput(tv2);
+  auto tv3 = add(tv2, tv2);
```
wujingyue, line 2004: So MarkAliasesPreparePass doesn't kick in. Other changes in the file are minor cleanups.

tests/cpp/test_remove_bcast_squeeze.cpp
```diff
@@ -78,7 +78,7 @@ TEST_F(RemoveBcastSqueezeTest, BcastSqueezeMultipleUses) {
   auto tv1 = set(tv0);
   auto tv2 = broadcast(tv1, is_broadcast_dim);
   auto tv3 = squeeze(tv2, is_broadcast_dim);
-  auto tv4 = set(tv3);
+  auto tv4 = add(tv3, tv3);
```
wujingyue, line 81: Without this change, `squeeze` will be bookend'ed.

liqiangxl:
Can we do a `--diff` to check generated code change?

benchmarks/cpp/matmul.cpp
```diff
@@ -61,6 +61,7 @@ void setupMatmul(Fusion* fusion, MmaLayout layout, MatmulParams params) {
 
   fusion->addOutput(d);
 
+  preseg_passes::OptimizationPass<preseg_passes::PreSegmenter>::runPass(fusion);
```
wujingyue, line 64: This change has been submitted separately as https://github.com/NVIDIA/Fuser/pull/2680.

jjsjann123: should we change base to main? since #2680 is merged.

surprisingly that didn't happen automatically.

jjsjann123: wait. I was mistaken. The base of this PR is on the revert.... NVM

csrc/preseg_passes/mark_aliases_prepare.cpp
```diff
@@ -50,86 +110,100 @@ void MarkAliasesPreparePass::runPass(Fusion* fusion) {
...
-        // A `segment_set` before a fusion input is useless.
-        continue;
-      }
-      // Rarely, if `aliased_io` is already defined by `segment_set`, don't
-      // create another `segment_set`.
-      if (LoadStoreOp* def =
-              dynamic_cast<LoadStoreOp*>(aliased_io->definition())) {
-        if (def->opType() == LoadStoreOpType::SegmenterSet) {
-          continue;
+    auto insert_segment_set = [&]() {
```
jjsjann123, line 163: dumb question, this lambda is not used anywhere else. Does lambda bring anything to the table?

csrc/preseg_passes/mark_aliases_prepare.cpp
```diff
@@ -50,86 +110,100 @@ void MarkAliasesPreparePass::runPass(Fusion* fusion) {
...
-    for (Expr* e : users_used_only_by_aliases) {
-      ir_utils::replaceValInExprInputs(e, aliased_io, copy);
-    }
-    if (aliased_io->isFusionOutput()) {
-      fusion->replaceOutput(aliased_io, copy);
-    }
+      // If all aliasing users are `segment_set`, don't create another
+      // `segment_set`.
+      if (std::all_of(
+              i, j, [](const Use& use) { return isSegmentSet(use.user); })) {
```
jjsjann123, line 186: this condition is `all_of`. which means, if there's only a subset of `use.user` being segmenter_set, we are still going to put another segmenter_set before them.

jjsjann123 (commented):
haven't finished the code yet. Some preliminary comment.

csrc/scheduler/mark_aliases.cpp
```diff
@@ -33,21 +33,22 @@ void markAliases(Fusion* fusion) {
 
   for (TensorView* out :
        ir_utils::filterByType<TensorView>(fusion->outputs())) {
-    if (fusion->getOutputAlias(out).type != AllocationType::New) {
+    // AllocationType::ReuseBuffer requires the output to be updated in place
+    // so it can't be computed as an alias.
+    if (fusion->getOutputAlias(out).type == AllocationType::ReuseBuffer) {
```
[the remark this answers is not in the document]
wujingyue: Probably doesn't make a difference in practice. AllocationType::Evaluate is used for either aliases or ATen fallback. The latter is picked up by the ExprEvalScheduler, which doesn't call this function. I made the change only because it's easier to explain why we should skip ReuseBuffer.

csrc/preseg_passes/mark_aliases_prepare.cpp
wujingyue: So I can early-return inside the lambda without exiting MarkAliasesPreparePass::runPass, the outer function. Alternatively, I could use nested ifs or say
```
if (...) {
  i = j;
  continue;
}
```
which I'm not a big fan of.

csrc/preseg_passes/mark_aliases_prepare.cpp
wujingyue: Yes and that's suboptimal. I think the best fix to that is: still create another segment_set here but short circuit the double segment-set in the following std::for_each. I'll give it a shot tomorrow.

tests/cpp/test_alias.cpp
```diff
@@ -1391,26 +1484,176 @@ TEST_F(AliasTest, SegmentMetaOps) {
...
+  for (int i = 0; i < 3; i++) {
+    in_tensors.push_back(at::randn({b, s, h * f}).cuda());
+  }
+  std::vector<at::Tensor> out_tensors = fec.runFusionWithInputs(in_tensors);
+  testValidate(fec.fusion(), out_tensors, in_tensors, __LINE__, __FILE__);
+
+  EXPECT_TRUE(out_tensors[2].is_alias_of(out_tensors[1]));
+}
+
+TEST_F(AliasTest, Bookend_Issue2375) {
```
jjsjann123, line 1624: cc'ing @liqiangxl

jjsjann123 (approved):
LGTM.

Thanks for modifying existing tests to work with the new analysis. That looked painful to go over each.

csrc/preseg_passes/mark_aliases_prepare.cpp
```diff
@@ -7,13 +7,64 @@
...
+std::unordered_set<Expr*> exprsUsedByNonAliases(
+    const AliasAnalysisResult& analysis,
+    Fusion* fusion) {
+  std::vector<Val*> non_aliases;
+  for (TensorView* tv : ir_utils::allTvs(fusion)) {
+    if (analysis.getRoot(tv) == nullptr) {
+      non_aliases.push_back(tv);
+    }
+  }
+  std::vector<Expr*> used_by_non_aliases = StmtSort::getExprsTo(non_aliases);
```
liqiangxl, line 62: IIUC, this `used_by_non_aliases` means `exprs_producing_non_alias`.

csrc/preseg_passes/mark_aliases_prepare.cpp
```diff
@@ -50,86 +101,101 @@ void MarkAliasesPreparePass::runPass(Fusion* fusion) {
...
-      if (used_by_non_aliases.count(e)) {
-        users_used_by_non_aliases.push_back(e);
-      } else {
-        users_used_only_by_aliases.push_back(e);
-      }
+  // The remaining are optimizations to reduce the number of `segment_set`s
+  // inserted.
+  //
+  // Group `uses_to_segment` by `use_of` and remove duplicates.
+  std::sort(uses_to_segment.begin(), uses_to_segment.end());
```
liqiangxl, line 135: we usually use `stable_sort` in nvFuser for reproductivity.

csrc/preseg_passes/mark_aliases_prepare.cpp
```diff
@@ -50,86 +101,101 @@ void MarkAliasesPreparePass::runPass(Fusion* fusion) {
...
+  }
 
-    // If all users are used by non-alias outputs, do nothing. Adding
-    // segment_set around unlikely creates more no-op regions.
-    if (users_used_only_by_aliases.empty()) {
-      continue;
+  auto i = uses_to_segment.begin();
+  while (i != uses_to_segment.end()) {
+    TensorView* use_of = i->use_of;
+    auto j = i;
```
liqiangxl, line 149: I feel we can start from `j = i + 1`

csrc/preseg_passes/mark_aliases_prepare.cpp
```diff
@@ -50,86 +101,101 @@ void MarkAliasesPreparePass::runPass(Fusion* fusion) {
...
+    for (const auto& use : uses_to_segment) {
+      debug() << "Will put a segment_set at " << use.user << std::endl;
     }
+  }
 
-    // If all users are used by non-alias outputs, do nothing. Adding
-    // segment_set around unlikely creates more no-op regions.
-    if (users_used_only_by_aliases.empty()) {
-      continue;
+  auto i = uses_to_segment.begin();
```
liqiangxl, line 146: can we change `i` and `j` to more meaningful names? and add comments to these two while loops.

tests/cpp/test_gather.cpp
```diff
@@ -1101,6 +1102,9 @@ TEST_F(ScatterGatherTest, TakeAlongAxisIntermediateTensorTranspose2) {
 // transpose the dimension produced by take_along_axis. Currently not
 // supported by the transpose scheduler
 TEST_F(ScatterGatherTest, TakeAlongAxisIntermediateTensorTranspose3) {
+  preseg_passes::OptimizationPassGuard<preseg_passes::MarkAliasesPreparePass>
+      optimization_guard(false);
```
liqiangxl, line 1106: can we add a comment about why disabling?

tests/cpp/test_resize.cpp
liqiangxl: can we add this comment to the code?

liqiangxl (approved):
LGTM. Thanks for adding this powerful feature!

liqiangxl:
(1) We may need to check the influence on the CPU overhead of segmenter since `segment_set enforces the segmenter to give up scheduling the complete fusion and instead trying to merge from singletons.`
(2) When merged into main branch, needs to check code diff to avoid over segmentation similar to several tests modified in this PR.

csrc/preseg_passes/mark_aliases_prepare.cpp
wujingyue: I'll rename this to `depended_by_non_aliases`. I wanted to emphasize on transitivity, because it's not only the direct user that I care about.

csrc/preseg_passes/mark_aliases_prepare.cpp
wujingyue: It makes no difference because the pairs `<use_of,user>` in `uses_to_segment` are distinct after the call to `std::unique`. 

But I got your concern on non-determinism, of which the root cause is sorting by pointer values. What sort keys would you use? Names?

csrc/preseg_passes/mark_aliases_prepare.cpp
wujingyue: Got it. Will change the loop from while-do to do-while, which achieves the same effect of one less iteration.

csrc/preseg_passes/mark_aliases_prepare.cpp
wujingyue: FYI, I'll make insert_segment_set a real function. The capture list is much shorter than I expected.

csrc/preseg_passes/mark_aliases_prepare.cpp
wujingyue: After thinking more about this, the logic as is is probably correct because `segment_set` is supposed to block merging the edges. Consider
```
in -> (add) -> t0 -> (segment_set) -> t1 -> (reshape) -> out0
               |
               +--> (permute) -> out1
               +--> (permute) -> out2
```

IIUC, it's incorrect to redirect the `permutes` to `t1` like the following. 
```
in -> (add) -> t0 -> (segment_set) -> t1 -> (reshape) -> out0
                                      |
                                      +--> (permute) -> out1
                                      +--> (permute) -> out2
```
This is because the original `segment_set` intends to separate the `reshape` and the two `permute`s, but the new `segment_set` doesn't enforce that intention. Instead, the correct outcome should be
```
in -> (add) -> t0 -> (segment_set) -> t2 -> (segment_set) -> t1 -> (reshape) -> out0
                                      |
                                      +--> (permute) -> out1
                                      +--> (permute) -> out2
```

csrc/preseg_passes/mark_aliases_prepare.cpp
liqiangxl: I think `name()` should work.

tests/cpp/test_alias.cpp
```diff
@@ -964,6 +1008,42 @@ TEST_F(AliasTest, DoNotOverSegment) {
 
   FusionKernelRuntime* runtime = fec.getMostRecentKernelRuntime();
   EXPECT_FALSE(runtime->isSegmented());
+
+  // permute_out should be recognized as an alias of add_out. However, the
+  // pointwise scheduler `cacheFork`s outputs
+  // (https://github.com/NVIDIA/Fuser/blob/de488172d280d033511dfc0939e2207f87c1b1cb/csrc/scheduler/pointwise.cpp#L540),
+  // making permute_out no longer a transitive user of add_out.
+  // EXPECT_TRUE(out_tensors[1].is_alias_of(out_tensors[0]));
```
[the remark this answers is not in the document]
wujingyue: That's right...

wujingyue:
!build --diff

wujingyue:
> When merged into main branch, needs to check code diff to avoid over segmentation similar to several tests modified in this PR.

I don't understand the result of codegen_diff any more. E.g., https://gitlab-master.nvidia.com/dl/pytorch/fuser-gh-mirror/-/jobs/103497030/viewer failed but I couldn't find which test case failed and what was the diff.

wujingyue:
!build --diff

PR #2640, opened 2024-07-19
Support predicate indexing for unswitch/unroll/vectorize

naoyam:
Adding support of generating predicates for unswitch/unroll and vectorize.

These predicates are generated by `UnswitchPredicate`. Unswitch and unroll are effectively the same for predication. Vectorize is also mostly the same with only one difference, which is that the predicate of the vectorized loop itself is altered.

For example, if a tensor has `[I0, I1, I2]` and `I1` is unswitched, the code structure would look like:

```
for i0 in I0:
  if idx(i0, I1 - 1, I2 - 1) < N:
    for i1 in I1:
       for i2 in I2:
           do_something
       end-for
     end-for
  else
    for i1 in I1:
       for i2 in I2:
           if idx(i0, i1, i2) < N:
              do_something
           end-if
       end-for
     end-for
  end-if
end-for
```

Here, `idx` is a function to produce the linearized index using the loop indices, and `N` is the extent of the domain. Notice that both `I1` and `I2` loops are predicated with their maximum index values.
 
If `I1` is vectorized:

```
for i0 in I0:
  for i2 in I2:
    if idx(i0, I1 - 1, i2) < N:
      do_something
    end-if
   end-for
end-for
```

I tried to reproduce the same predicates as much as possible with the current lowering. For example, while working on this, I realized it should be also fine to use `0` instead of `I - 1` for vectorized domains since splits involved in vectorized domains are guaranteed to be divisible. For example, this should be equivalent to the above example:

```
for i0 in I0:
  for i2 in I2:
    if idx(i0, 0, i2) < N:
      do_something
    end-if
   end-for
end-for
```

Using zero would probably make the expression simplifier work more effectively, so that's probably a better option for performance. I'll look into it as a separate PR (#2628).

### DIFF
--- a/csrc/device_lower/analysis/index_compute.cpp
+++ b/csrc/device_lower/analysis/index_compute.cpp
@@ -238,59 +238,6 @@ IndexingParameters getNonGlobalInitialIndexParameters(
   return index_parameters;
 }
 
-// Return true if it is sufficient to predicate the end of the loop
-// iteration. An aligned vectorized loop is one example where it is
-// guaranteed to be valid by the validation checks. More generally,
-// the divisible split set is used to find such loops. The divisible
-// split set contains splits used in view transformations as well as
-// those whose output domains are vectorized. View transformations
-// guarantee that any split involved is divisible, whereas
-// vectorization only guarantees that the overall root extent is
-// divisible by the split factor. Thus, if a loop IterDomain is
-// an output of a split included in the divisible view splits, we can
-// just predicate the end of the loop iteration. If a loop IterDomain
-// is an output of a divisible split due to vectorization, it is only
-// valid when the loop IterDomain is mapped with the vectorized inner
-// output IterDomain. If it is mapped with an outer IterDomain, since
-// the split input IterDomain may be an output IterDomain of a
-// non-divisible split, we still need to predicate each loop iteration
-// value.
-bool predicateAtEnd(ForLoop* loop) {
-  auto loop_id = loop->iter_domain();
-  auto split = dynamic_cast<Split*>(loop_id->definition());
-  if (split == nullptr) {
-    return false;
-  }
-
-  bool is_divisible = GpuLower::current()->divisibleSplitSet().count(split) > 0;
-
-  if (!is_divisible) {
-    return false;
-  }
-
-  // Find the other output of the split
-  auto other_out_id =
-      split->inner() == loop_id ? split->outer() : split->inner();
-
-  // If the other output is mapped with a vectorized IterDomain,
-  // this IterDomain needs to be predicated at each iteration point.
-  const auto& other_id_exact_set = GpuLower::current()
-                                       ->caMap()
-                                       ->getIdSets(IdMappingMode::EXACT)
-                                       .getDisjointSetOf(other_out_id);
-
-  if (std::any_of(
-          other_id_exact_set.begin(), other_id_exact_set.end(), [](auto id) {
-            return id->getParallelType() == ParallelType::Vectorize;
-          })) {
-    return false;
-  }
-
-  // Now it is either loop_id is mapped with a vectorized IterDomain
-  // or it's an output of view transformations.
-  return true;
-}
-
 // Check if this loop is actually unswitched, meaning an initial index
 // of the maximum value from a non-size-one range is used.
 bool trackUnswitchedDomain(ForLoop* loop) {
@@ -379,8 +326,8 @@ IndexingParameters getPredicateInitialIndexParameters(
       within_unswitch = loop == unswitch_or_vec_loop;
     }
 
-    bool predicate_at_end =
-        within_unswitch || loop == unswitch_or_vec_loop || predicateAtEnd(loop);
+    bool predicate_at_end = within_unswitch || loop == unswitch_or_vec_loop ||
+        lower_utils::predicateAtEnd(loop);
 
     if (predicate_at_end) {
       // Rely on the reference to check broadcasting. The for loop could be
--- a/csrc/device_lower/utils.cpp
+++ b/csrc/device_lower/utils.cpp
@@ -942,23 +942,6 @@ bool isReductionInitExpr(const Expr* expr) {
   return true;
 }
 
-// Return true if it is sufficient to predicate the end of the loop
-// iteration. An aligned vectorized loop is one example where it is
-// guaranteed to be valid by the validation checks. More generally,
-// the divisible split set is used to find such loops. The divisible
-// split set contains splits used in view transformations as well as
-// those whose output domains are vectorized. View transformations
-// guarantee that any split involved is divisible, whereas
-// vectorization only guarantees that the overall root extent is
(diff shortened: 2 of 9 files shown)


csrc/device_lower/analysis/index_compute.cpp
```diff
@@ -238,59 +238,6 @@ IndexingParameters getNonGlobalInitialIndexParameters(
   return index_parameters;
 }
 
-// Return true if it is sufficient to predicate the end of the loop
```
naoyam, line 241: Moved to device_lower/utils

csrc/device_lower/utils.cpp
```diff
@@ -942,6 +942,59 @@ bool isReductionInitExpr(const Expr* expr) {
   return true;
 }
 
+// Return true if it is sufficient to predicate the end of the loop
```
naoyam, line 945: Moved from analysis/index_compute

csrc/id_model/indexing.h
```diff
@@ -88,10 +88,16 @@ class TensorIndexer {
   // expr as a consumer. Each predicate corresponds to a domain of the
   // tensor, which is by default one of the logical domains but can be
   // an intermediate domain with contiguous indexing.
-  std::vector<PredicateInfo> getInlinePredicates(
+  //
```
naoyam, line 91: Extended to support unswitch/unroll/vectorize

csrc/id_model/predicate_indexing.cpp
```diff
@@ -53,4 +53,104 @@ std::vector<IterDomain*> getPredicateDomains(
...
+                fl->simplifiedStop(), fl->fusion()->oneVal());
+    } else {
+      return nullptr;
+    }
+  };
+
+  // Inspect the for-loops from outer to inner and keep track of
+  // unswitching since it affects all inner loops
+  bool within_unswitch = false;
+  for (const auto fl : for_loops) {
```
naoyam, line 100: Pretty much the same logic as `getPredicateInitialIndexParameters` in analysis/index_compute.cpp

tests/cpp/test_indexing.cpp
```diff
@@ -119,7 +119,15 @@ class AbstractGetReference {
...
-  virtual Val* getPredicate(TensorView* tv) const {
+  // Returns the inline predicate of a given tensor.
+  virtual Val* getInlinePredicate(TensorView* tv) const {
+    return nullptr;
+  }
+
+  // Returns the outer predicate of a given tensor. This only matters
+  // when tv is unswitched or unrolled. Note that if it's vectorized,
+  // the predicate is still inlined.
+  virtual Val* getOuterPredicate(TensorView* tv) const {
```
naoyam, line 130: Overrides this function when validating unswitch predicates

naoyam:
!build

csrc/id_model/predicate_indexing.cpp
```diff
@@ -53,4 +53,104 @@ std::vector<IterDomain*> getPredicateDomains(
...
+
+  // For an iter domain of index i, it is valid to use N-1 instead of
+  // i, where N is the extent of the iter domain if either of the
+  // following conditions is satisfied:
+  //
+  // - Vectorized
+  // - predicateAtEnd returns true
+  // - Within an unswitch/unroll loop
+  //
+  // Use N-1 instead of i but not when it's thread paralellized so
```
jacobhinkle, line 74:
```suggestion
  // Use N-1 instead of i but not when it's thread parallelized so
```

csrc/device_lower/utils.cpp
```diff
@@ -942,6 +942,59 @@ bool isReductionInitExpr(const Expr* expr) {
...
+// vectorization only guarantees that the overall root extent is
+// divisible by the split factor. Thus, if a loop IterDomain is
+// an output of a split included in the divisible view splits, we can
+// just predicate the end of the loop iteration. If a loop IterDomain
+// is an output of a divisible split due to vectorization, it is only
+// valid when the loop IterDomain is mapped with the vectorized inner
+// output IterDomain. If it is mapped with an outer IterDomain, since
+// the split input IterDomain may be an output IterDomain of a
+// non-divisible split, we still need to predicate each loop iteration
+// value.
```
jacobhinkle, line 961: Nit: the above comment is a duplicate of the one found in the header so we could probably just keep the one in the header.

csrc/device_lower/utils.cpp
```diff
@@ -942,6 +942,59 @@ bool isReductionInitExpr(const Expr* expr) {
...
+bool predicateAtEnd(ForLoop* loop) {
+  auto loop_id = loop->iter_domain();
+  auto split = dynamic_cast<Split*>(loop_id->definition());
+  if (split == nullptr) {
+    return false;
+  }
```
jacobhinkle, line 967: I realize this is pre-existing code, but wouldn't it also be possible to have a divisible split followed by a merge? For example:
```
tv[ i0, 32 ]
tv->split(1, 8); // [i0, 4, 8]
tv->split(2, 4); // [i0, 4, 2, 4]
tv->merge(1, 3); // [i0, 16 , 2]
tv->axis(0)->parallelize(ParallelType::Unswitch);
```
Now the both of the dimensions in `tv` divide evenly into the logical dimension of `tv`, but only one of them is a direct split output, so as I understand it that means only the inner loop will have `predicateAtEnd` return `true`. Is this correct, and is this the desired behavior?

csrc/id_model/predicate_indexing.cpp
```diff
@@ -53,4 +53,104 @@ std::vector<IterDomain*> getPredicateDomains(
...
+    if (loop_index->isConst()) {
+      continue;
+    }
+
+    Val* replacement = loop_index;
+
+    // Trivial loop. Note that not all trivial loops should just use
+    // the start index for predication. For example, a vectorized loop
+    // is trivial, but its predicate should use `vec_factor - 1` as
+    // its index. This is taken care after this.
```
jacobhinkle, line 137: This note about vectorized loops seems to contradict the comment at lines 79-81 that vectorized loops are already proven divisible so we might as well use 0 for their predicates.

jacobhinkle (commented):
I think I understand the logic. I just have a couple of clarifying questions.

csrc/device_lower/utils.cpp
naoyam: I think you're right. This check won't detect such patterns. Not ideal, but this was sufficient when I was working on some normalization or reduction fusions. You're more than welcome to extend it 😁

csrc/id_model/predicate_indexing.cpp
naoyam: Right, that's what I meant here:

> Using zero would probably make the expression simplifier work more effectively, so that's probably a better option for performance. I'll look into it as a separate PR (https://github.com/NVIDIA/Fuser/pull/2628).

Right now, we are predicating with `N - 1`, but `0` should be fine as well. In this PR, I intend to keep the current behavior.

jacobhinkle (approved):

PR #2644, opened 2024-07-19
Print repro when error occurs at definition

jacobhinkle:
We currently check for errors during `fd.execute()` and we print a useful repro. However, some errors are caught during definition and currently we just print the error without a repro. Instead, this PR just copies some of the printing code to the `__exit__` method of `FusionDefinition` so that we get a nice error message e.g.
```

    T0 = fd.define_tensor(shape=[-1, -1, -1], contiguity=[True, True, None], dtype=DataType.Float, is_cpu=False, stride_order=[2, 1, 0])
    T1 = fd.ops.where(T0, T0, T0)

    nvfuser_fusion_id0(fd)

```


nvfuser/__init__.py
```diff
@@ -58,7 +58,25 @@ def __enter__(self):
         return self._setup_definition()
 
     def __exit__(self, type, value, traceback):
-        self._finalize_definition()
+        try:
+            self._finalize_definition()
+        except Exception as err:
+            msg = (
```
jjsjann123, line 64: nitpick, can we have this part in a separate question and reuse it with the section in execute? just so that we don't have duplicated strings. :stuck_out_tongue_winking_eye:

jjsjann123 (approved):
sorry but nitpick is all I do.
Stamp for now and I'm going to trust your to follow up on the clean up.

PR #2656, opened 2024-07-22
Overlap: replace at::matmul_out by mul_out+sum_out in ATen implementation of rs-based pipelining

samnordmann:
`at::matmul_out` seems to allocate intermediate buffer, resulting in cudaMalloc with significant overheads in the execution trace. To have better control on allocation, we used those two unfused ATen ops, i.e., pointiwise multiply and sum reduction.

Stacked on top of
- #2665.

### DIFF
--- a/tests/cpp/test_multidevice_overlap.cpp
+++ b/tests/cpp/test_multidevice_overlap.cpp
@@ -54,7 +54,7 @@ class OverlapTest : public MultiDeviceTest {
   int64_t num_devices_;
   int64_t my_device_index_;
   std::vector<int64_t> all_devices_;
-  at::Tensor ta_, tb_, tc_locally_reduced_, tc_, tc_expected_;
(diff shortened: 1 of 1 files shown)


samnordmann:
!build

naoyam:
@samnordmann Replacing MMA with mul-sum doesn't seem ideal. I think we should continue investigating why cudaMalloc is involved. Can you create an issue with a repro?

samnordmann:
> @samnordmann Replacing MMA with mul-sum doesn't seem ideal. I think we should continue investigating why cudaMalloc is involved. Can you create an issue with a repro?

Here is the issue:
https://github.com/NVIDIA/Fuser/issues/2673

Let me know if you want a more minimal reproducer which is fully torch contained

samnordmann:
We will keep using `matmul_out` according to the results of #2673

Issue #2658, opened 2024-07-22
forward inputs in codegen create pointwise kernel for no-op fusion segments.


jjsjann123:
Run into this issue in https://github.com/NVIDIA/Fuser/pull/2630

A repro on main branch:
```
TEST_F(SegmentationTest, Playground) {
  auto fusion = std::make_unique<Fusion>();
  FusionGuard fg(fusion.get());
 
  TensorView* vanilla_in = makeContigConcreteTensor({2, 3});
  TensorView* in = relu(vanilla_in);
  TensorView* permute_out = permute(in, {1, 0});
  TensorView* compute_out = mul(in, in);
  compute_out = add(compute_out, in);
  fusion->addInput(vanilla_in);
  fusion->addOutput(in);
  fusion->addOutput(permute_out);
  fusion->addOutput(compute_out);
 
  FusionExecutorCache fec(std::move(fusion));
  at::Tensor in_tensor = at::randn({2, 3}).cuda();
  std::vector<at::Tensor> out_tensors = fec.runFusionWithInputs({in_tensor});
  testValidate(fec.fusion(), out_tensors, {in_tensor}, __LINE__, __FILE__);
}
```

fusion fed to segmenter.
```
Segmented_Fusion Dump: -- Re-written complete fusion:{
Inputs:
  T0_g[ iS0{2}, iS1{3} ], float
Outputs:
  T5_g[ iS10{2}, iS11{3} ], float
  T2_g[ iS5{3}, iS4{2} ], float
  T4_g[ iS8{2}, iS9{3} ], float

%kernel_math {
T1_g[ iS2{2}, iS3{3} ]
   = relu(T0_g[ iS0{2}, iS1{3} ]);
T5_g[ iS10{2}, iS11{3} ]
   = SegmenterSet( T1_g[ iS2{2}, iS3{3} ] )
T2_g[ iS5{3}, iS4{2} ]
   = Set.Permute( T5_g[ iS10{2}, iS11{3} ], cache_op=Streaming )
T3_g[ iS6{2}, iS7{3} ]
   = T1_g[ iS2{2}, iS3{3} ]
   * T1_g[ iS2{2}, iS3{3} ];
T4_g[ iS8{2}, iS9{3} ]
   = T3_g[ iS6{2}, iS7{3} ]
   + T1_g[ iS2{2}, iS3{3} ];
} // %kernel_math

```

[...]

PR #2659, opened 2024-07-22
fusion segmenter forward input to exclude uop used by SegmenterSet

jjsjann123:
Fixes: #2658

forwarded unary operation on SegmenterSet could potentially put computation ops in no-op segments.

This PR prevents any uop with output consumed by SegmenterSet to be forwarded.


jjsjann123:
!build

jjsjann123:
Not sure who's a good reviewer for segmenter related issues.

jjsjann123:
!build

tests/cpp/test_segmentation.cpp
```diff
@@ -702,4 +704,29 @@ TEST_F(SegmentationTest, MultipleSegmentSetsInOneSegment) {
...
+  // Disable mark aliases prepare pass, which might insert more segment_set
+  preseg_passes::OptimizationPassGuard<preseg_passes::MarkAliasesPreparePass>
+      optimization_guard(false);
+
+  auto fusion = std::make_unique<Fusion>();
+  FusionGuard fg(fusion.get());
+
+  TensorView* vanilla_in = makeContigConcreteTensor({2, 3});
+  TensorView* in = relu(vanilla_in);
+  TensorView* seg_in = segment_set(in);
```
jjsjann123, line 717: In issue #2658 , we ended up forwarding the `relu` op into `segment_set`. Ended up having two pointwise kernel.

jacobhinkle (approved):
LGTM. This seems safe to me. I am not sure it will 100% prevent this issues, since we could also naturally segment at a location where a NoOp segment then gets some forwarded uops. I think ideally we would actually include the forwarded ops in the canSchedule checks also.

jjsjann123:
> we could also naturally segment at a location where a NoOp segment then gets some forwarded uops.

That sounds more like a segmenter issue in general. i.e. when should we try to get a no-op segment. Our current greedy segmenter doesn't consider that at all.

Hence @wujingyue 's mark alias pass tries to use `segmenter_set` to force a segment to facilitate that. At least with this PR, that attempt is not going to be undo by forwarded inputs. :crossed_fingers:

PR #2660, opened 2024-07-22
Add a predicate to main loop of circular buffering

naoyam:
Example with double buffering:

```
for int i = 0; i < N; ++i
    if (idx(i+1) < logical_size && i + 1 < N)
```

This is only done when no epilogue is generated. In that case, this predicate is not necessary.


naoyam:
See #2663 as well.

naoyam:
Closing as it's not strictly necessary

PR #2661, opened 2024-07-22
Fix epilogue loop generation

naoyam:
Note that CudaCodeGenerator currently always starts a loop with 0 even if ForLoop::start_ is non-zero. I think this change is safe since only use case of non-zero start should be the epilogue loop of circular buffering. The concern of degenerate loop should not be applicable.

### DIFF
--- a/csrc/device_lower/pass/circular_buffer.h
+++ b/csrc/device_lower/pass/circular_buffer.h
@@ -217,7 +217,7 @@ class CircularBufferInfo {
 
   //! Get the number of circular buffer stages for the given axis,
   //!  the number of stages will be 2 in the case of circular buffer loop.
-  unsigned int getStageDepthFor(IterDomain* circular_buffered_id) const;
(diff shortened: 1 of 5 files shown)


naoyam:
!build --diff

csrc/device_lower/pass/circular_buffer.h
```diff
@@ -217,7 +217,7 @@ class CircularBufferInfo {
 
   //! Get the number of circular buffer stages for the given axis,
   //!  the number of stages will be 2 in the case of circular buffer loop.
-  unsigned int getStageDepthFor(IterDomain* circular_buffered_id);
+  unsigned int getStageDepthFor(IterDomain* circular_buffered_id) const;
```
rdspring1, line 220: nitpick: change from `unsigned int` to `int64_t`

rdspring1 (approved):
LGTM

naoyam:
!build

naoyam:
!build

PR #2662, opened 2024-07-22
Revert "Host irs: Reduce-Scatter based pipelining (#2584)"

jjsjann123:
This reverts commit b62292bb18364075dc9bcffa2a174c55c28a2645.

Make CI green again!
Feel free to discard the revert. I'm happy with disabling the tests if we think that's a better way.
Would be even better if we could work out a fix.

### DIFF
--- a/csrc/expr_evaluator.h
+++ b/csrc/expr_evaluator.h
@@ -70,10 +70,6 @@ class ExpressionEvaluator {
     return known_values_.count(value) > 0;
   }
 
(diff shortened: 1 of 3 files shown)


jjsjann123:
!build

samnordmann:
#2665 should fix the CI! sorry about that. I think we can close this PR

naoyam:
Merged the fix. Closing.

PR #2663, opened 2024-07-22
Always generate epilogue

naoyam:
Stacked on top of #2661.

We currently only generate circular buffer epilogue loops when the producer is in global memory. This PR changes we always generate epilogue.

This is not strictly necessary for correctness but avoids extra memory accesses that won't be used. 

Alternatively, we could add an extra predicate in the main loop. See #2660 as well.

Overall, I think this approach is simpler and I don't see any performance concern compared to #2660.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -271,7 +271,7 @@ add_library(codegen_internal OBJECT ${NVFUSER_SRCS})
 if(NOT MSVC)
   target_compile_options(codegen_internal PRIVATE
     -Wall -Wno-unused-function
-    -Werror
+    # -Werror
   )
 endif()
 
--- a/csrc/codegen.cpp
+++ b/csrc/codegen.cpp
@@ -2895,20 +2895,9 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
       indent() << "#pragma unroll 1\n";
     }
 
-    indent() << "for(nvfuser_index_t " << gen_index;
-    if (loop->iter_domain()->isParallelized()) {
-      code_ << " = " << gen_start << "; ";
-    } else {
-      // Do not start at  the start of the ID when not parallelized. Instead,
-      // start at 0. Predicates will protect buffers between 0 and ID->start(),
-      // however if we started at ID->start and extent == ID->start, we could
-      // have a "degenerate" loop (loop with no iterations). It may not be an
-      // issue to have a 0-sized loop, but all potential consequences haven't
-      // been covered. One example is WAR analysis which could incorrectly think
-      // a barrier inside a 0-sized loop actually provides protection.
-      code_ << " = 0; ";
-    }
-    code_ << gen_index << " < " << gen_stop << "; " << step_code.str() << ") ";
+    indent() << "for(nvfuser_index_t " << gen_index << " = " << gen_start
+             << "; " << gen_index << " < " << gen_stop << "; "
+             << step_code.str() << ") ";
     startBlock(true);
     kir::ConstIrVisitor::handle(loop);
     endBlock();
--- a/csrc/device_lower/pass/circular_buffer.cpp
+++ b/csrc/device_lower/pass/circular_buffer.cpp
@@ -214,7 +214,9 @@ class CircularBufferLoopCloner : public kir::IrVisitor {
         loop_type_ == CircularBufferLoopStage::Main &&
         requireEpilogue(circular_buffer_load_exprs_)) {
       stop = IrBuilder::subExpr(
-          circular_buffer_loop_->stop(), gpu_lower->kernel()->oneVal());
+          circular_buffer_loop_->stop(),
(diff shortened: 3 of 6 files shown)


naoyam:
!build

zasdfgbnm:
Is this related to https://github.com/NVIDIA/Fuser/pull/2008? cc: @jacobhinkle

naoyam:
> Is this related to #2008? cc: @jacobhinkle

Hmm, does that mean we don't want to have epilogue loops in some cases?

jacobhinkle:
I dont think it's necessarily _bad_ to have an epilogue in cp.async cases, maybe other than code size. With an epilogue we would not need to drain the leftover jobs as was added in #2008.

naoyam:
The test modified in #2008 doesn't seem to fail with this PR change. I suppose it's because the epilogue loop is actually empty. Am I understanding correctly? Do we have some simpler tests, preferably without loop rotation?

naoyam:
> The test modified in #2008 doesn't seem to fail with this PR change. I suppose it's because the epilogue loop is actually empty. Am I understanding correctly? Do we have some simpler tests, preferably without loop rotation?

Ah, no, actually they are failing. Will look into them.

jacobhinkle:
> > The test modified in #2008 doesn't seem to fail with this PR change. I suppose it's because the epilogue loop is actually empty. Am I understanding correctly? Do we have some simpler tests, preferably without loop rotation?
> 
> Ah, no, actually they are failing. Will look into them.

I remember the issue with using an epilogue loop for these `cp.async` waits: see https://github.com/NVIDIA/Fuser/pull/2005#discussion_r1540236954. The problem is that you need to specify the number of groups left in the wait as a constant, and even an unrolled loop variable cannot be used there. Because of that I'm not sure there's any good way to do this without manually replicating (unrolling) the epilogue loop, which would probably introduce more bugs.

naoyam:
> > > The test modified in #2008 doesn't seem to fail with this PR change. I suppose it's because the epilogue loop is actually empty. Am I understanding correctly? Do we have some simpler tests, preferably without loop rotation?
> > 
> > 
> > Ah, no, actually they are failing. Will look into them.
> 
> I remember the issue with using an epilogue loop for these `cp.async` waits: see [#2005 (comment)](https://github.com/NVIDIA/Fuser/pull/2005#discussion_r1540236954). The problem is that you need to specify the number of groups left in the wait as a constant, and even an unrolled loop variable cannot be used there. Because of that I'm not sure there's any good way to do this without manually replicating (unrolling) the epilogue loop, which would probably introduce more bugs.

So, for `cp.async`, it seems the best practice is to not generate epilogue and keep the current codegen as is. Is that what you think?

jacobhinkle:
> So, for cp.async, it seems the best practice is to not generate epilogue and keep the current codegen as is. Is that what you think?

Yes I think so, only because I could not figure out a way around that inline ptx limitation so that we could have a different call in each epilogue iteration.

rdspring1:
@jacobhinkle What about this for selecting `cp.async` at runtime? 

Take a loop at the PTX in https://ce.nvidia.com/z/oKxhxc.
If you apply a constraint to the index for `cp_async_wait_group_read`, it only runs through a subset of instructions.
e.g., `(i % 3)+2` => only `[2, 5]` are in the foo function.


```cpp
inline __device__ void cp_async_wait_group_read(int n)
{
  // https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#data-movement-and-conversion-instructions-cp-async-bulk-wait-group
  if (n == 0) { asm volatile("cp.async.wait_group.read 0; \n" ::: "memory"); }
  if (n == 1) { asm volatile("cp.async.wait_group.read 1; \n" ::: "memory"); }
  if (n == 2) { asm volatile("cp.async.wait_group.read 2; \n" ::: "memory"); }
  if (n == 3) { asm volatile("cp.async.wait_group.read 3; \n" ::: "memory"); }
  if (n == 4) { asm volatile("cp.async.wait_group.read 4; \n" ::: "memory"); }
  if (n == 5) { asm volatile("cp.async.wait_group.read 5; \n" ::: "memory"); }
  assert(n >= 0 && n <= 5);
}
```

jacobhinkle:
> @jacobhinkle What about this for selecting `cp.async` at runtime?

I think this is essentially the same as having a switch statement. That is indeed the alternative if we need to have an epilogue I believe.

> Take a loop at the PTX in https://ce.nvidia.com/z/oKxhxc. If you apply a constraint to the index for `cp_async_wait_group_read`, it only runs through a subset of instructions. e.g., `(i % 3)+2` => only `[2, 5]` are in the foo function.

That's good to know. I didn't know it would be able to prune dead branches like that. That means we could actually have a pretty high number of hard-coded cases and there would be no runtime penalty.

naoyam:
I'm closing this for now as it's still unclear if we could workaround the performance concerns. I wanted to do this to simplify predicate indexing but not strictly necessary.

PR #2666, opened 2024-07-23
Multidevice executor: Allow non-tv and multi-output ops

cowanmeg:
Modifies the `MultiDeviceExecutor` so that it does not error out on non-TV expressions and non-resharding multi-output expressions. 
In the future, we can extend the resharding logic to handle multi-output expressions with resharding, but this PR does not add this.


tests/cpp/test_multidevice_sharding.cpp
```diff
@@ -106,6 +106,87 @@ TEST_P(MultideviceShardingTest, ShardGlobalInput) {
       runtime.completeFusion(), outputs, inputs, {x1, x2}, __LINE__, __FILE__);
 }
 
+TEST_F(MultideviceShardingTest, Slice) {
```
cowanmeg, line 109: These tests are a bit overkill... Slice introduces non-tv ops for indexing and normalization is a multi-output expression. 
They will also get tested in the transformer attention unit test when it can be merged, which would contain these expressions.

samnordmann (approved):
LGTM! Thanks

csrc/multidevice/utils.cpp
```diff
@@ -168,6 +168,9 @@ bool haveDifferentShardings(TensorView* producer, TensorView* consumer) {
 }
 
 bool isResharding(Expr* expr) {
+  if (!ir_utils::isTvOp(expr)) {
+    return false;
```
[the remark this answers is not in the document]
cowanmeg: It's not precisely SliceOp, but the index computation used by Slice. For example: `i13 = fmin(i0, 5)`

This should be correct, because any expression that doesn't have a TV can't be resharding, since it only operates on scalars. 
It looks like almost all expressions with at least one input or output Tv is a TvOp, but there might be a few edge cases since the function checks the exact type of Op. https://github.com/NVIDIA/Fuser/blob/25a3426f1672d1a81db8faf272f5ba18463282db/csrc/device_lower/utils.cpp#L129
From this list I can't tell what ops aren't included, but potentially the ops not in the list would cause an issue...

cowanmeg:
!build

csrc/multidevice/lower_communication.cpp
```diff
@@ -293,11 +293,9 @@ std::vector<Communication*> lowerCommunication(Expr* c) {
...
 
 bool isLowerableToCommunication(Expr* expr) {
-  NVF_ERROR(
-      ir_utils::isTvOp(expr),
-      "Non-tv op is not supported yet: ",
-      expr->toString());
-  if (expr->isA<ReductionOp>()) {
+  if (!ir_utils::isTvOp(expr)) {
+    return false;
+  } else if (expr->isA<ReductionOp>()) {
```
wujingyue, line 298:
```suggestion
  }
  
  if (expr->isA<ReductionOp>()) {
```

csrc/preseg_passes/insert_reshardings.cpp
```diff
@@ -33,13 +35,19 @@ void insertReshardingsBefore(Fusion* fusion) {
...
     }
-    NVF_ERROR(
-        ir_utils::isTvOp(expr),
-        "Non-tv op is not supported yet: ",
-        expr->toString());
-    NVF_ERROR(
-        expr->outputs().size() == 1,
-        "multi-output expressions are not supported");
+
+    // Verify that multi-op expression requires no resharding.
```
wujingyue, line 39:
```suggestion
    // Verify that multi-output expression requires no resharding.
```

wujingyue (approved):

csrc/multidevice/utils.cpp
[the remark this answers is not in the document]
samnordmann: got it, nvm! I had a misconception of what `TvOp` was

Issue #2671, opened 2024-07-24
why consumer's logical domain doesn't equal producer's logical domain after fusion concretization


liqiangxl:
Given the following fusion, save it as `bug.py`
```
import torch
from nvfuser import FusionDefinition, DataType

def nvfuser_fusion_id0(fd : FusionDefinition) -> None :
    T0 = fd.define_tensor(shape=[-1], contiguity=[True], dtype=DataType.Half, is_cpu=False)
    T1 = fd.define_tensor(shape=[-1], contiguity=[True], dtype=DataType.Half, is_cpu=False)
    S2 = fd.define_scalar(32, dtype=DataType.Int)
    S3 = fd.ops.size(T0, dim=0)
    S4 = fd.ops.div(S3, S2)
    V5 = fd.define_vector([S2, S4], dtype=DataType.Int)
    T6 = fd.ops.reshape(T0, new_shape=V5)
    T7 = fd.ops.cast(T6, dtype=DataType.Float)
    T7 = fd.ops.segment_set(T7)
    T9 = fd.ops.reshape(T1, new_shape=V5)
    T10 = fd.ops.cast(T9, dtype=DataType.Float)
    T11 = fd.ops.mul(T7, T10)
    fd.add_output(T11)

with FusionDefinition() as fd:
    nvfuser_fusion_id0(fd)

inputs = [
    torch.randn((128,), dtype=torch.half, device='cuda:0').as_strided((128,), (1,)),
    torch.randn((128,), dtype=torch.half, device='cuda:0').as_strided((128,), (1,)),
]

fd.execute(inputs)

```
using the current main branch and run with `NVFUSER_DISABLE=parallel_compile NVFUSER_DUMP=fusion_ir_preseg,segmented_fusion,pre_segmenter_logging,fusion_ir_concretized python bug.py  2>&1 |tee bug.log`
The concretized Fusion is:
```
%kernel {
T8_l[ iS18{32}rf, iS19{( ceilDiv(i0, 32) )}rf ] = view( T0_g[ iS0{i0} ] )
T3_l[ iS25{32}, iS26{( ceilDiv(i2, 32) )} ]
   = __half2float(T8_l[ iS18{32}rf, iS19{( ceilDiv(i0, 32) )}rf ]);
T4_l[ iS28{32}, iS29{( ceilDiv(i2, 32) )} ]
   = SegmenterSet( T3_l[ iS25{32}, iS26{( ceilDiv(i2, 32) )} ] )
[...]

PR #2672, opened 2024-07-24
fix replace symbolic size

liqiangxl:
**See issue #2671 for details.**
**Issue**
Got the following err for a fusion with two reshapes segmented into 2 segments.
```
 dom0 and dom1 are not equal. dom0: iS22{32}rf, iS109{( (( (( getMetaData(T4) )).logical_size ))[1] )}rf. dom1: iS108{( (( (( getMetaData(T1) )).logical_size ))[0] )}rf
```
Err comes from `replaceSymbolicSizes`, for the 2nd segment with 2 inputs:
```
T1_g[ iS1{i2} ] __half
T4_g[ iS28{32}, iS29{( ceilDiv(i2, 32) )} ] float
```
`i2` is replaced to `T1.size(0)`
`ceilDiv(i2, 32)` is replaced to `T4.size(1)`
however, nvFuser doesn't know the relationship between `T1.size(0)` and `T4.size(1)`.
**Fix**
remove  `ceilDiv(i2, 32)` from the initial replacement map , scine it is derived from `i2`, we don't need to replace it with `T4.size(1)`.

### DIFF
--- a/benchmarks/cpp/matmul.cpp
+++ b/benchmarks/cpp/matmul.cpp
@@ -61,6 +61,8 @@ void setupMatmul(Fusion* fusion, MmaLayout layout, MatmulParams params) {
 
   fusion->addOutput(d);
 
+  preseg_passes::OptimizationPass<preseg_passes::PreSegmenter>::runPass(fusion);
+
   scheduleMatmul(fusion, params);
 }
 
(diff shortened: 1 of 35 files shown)


liqiangxl:
!build --diff

liqiangxl:
!build --diff

jacobhinkle:
Actually I think this is a sign of a little bit more serious problem with `OptOutMutator`. It looks like the replacement map is fine and we should be able to use those replacements. However, in the tensor in question, we have a root domain with a single axis and that is split to form the logical domain. When we do the replacement, the replaced root is correct, and the replaced logical domains are _almost_ right. However, while the first replacement `IterDomain` has a `Split` as its definition, the second one has a nullptr. I suspect this is because we do not treat siblings properly in OptOutMutator, so I think a proper fix would probably involve some changes to `mutator.cpp`. I can look at this in a little more detail and recommend a more specific fix.

PR #2675, opened 2024-07-24
add group norm benchmark

liqiangxl:
add group norm benchmark, NCHW -->NG(C/G)HW
**input sizes:** `generate_input_sizes(dims=4)`
**n_groups:** 
```
    # start num_groups from 1 and increase to 32 at max
    num_groups = 1
    while num_groups * 2 <= 32 and C % (num_groups * 2) == 0:
        num_groups *= 2
```
TODO:
add channel last


benchmarks/python/test_groupnorm_fwd.py
```diff
@@ -0,0 +1,145 @@
...
+    clear_cuda_cache()
+
+    N, C, H, W = size
+    x = torch.randn(size, device="cuda", dtype=dtype)
+    weight = torch.randn(C, device="cuda", dtype=dtype)
+    bias = torch.randn(C, device="cuda", dtype=dtype)
+    # start num_groups from 1 and increase to 32 at max
+    num_groups = 1
+    while num_groups * 2 <= 32 and C % (num_groups * 2) == 0:
+        num_groups *= 2
```
Priya2698, line 108: Can you add a comment here, explaining the reasoning behind this?

benchmarks/python/test_groupnorm_fwd.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def groupnorm_fwd(inputs: list):  # [in_tensor, weights, bias, n_groups]
+    return torch.nn.functional.group_norm(
+        inputs[0], num_groups=inputs[3], weight=inputs[1], bias=inputs[2], eps=1e-05
+    )
+
+
+@pytest.mark.parametrize("size", generate_input_sizes(dims=4))
+@pytest.mark.parametrize("dtype", FLOAT_DTYPES)
```
Priya2698, line 90: Is `channels_last` going to be part of this PR or will be added later?

benchmarks/python/test_groupnorm_fwd.py
```diff
@@ -0,0 +1,145 @@
...
+def test_groupnorm_fwd_nvf_benchmark(
```
Priya2698, line 91: Should we add a marker here for the scheduler too?

liqiangxl: added

benchmarks/python/test_groupnorm_fwd.py
liqiangxl: in a different PR.

Priya2698 (approved):
LGTM.
Please add a comment on how the number of groups is chosen for the fusions as well.

Issue #2623, opened 2024-07-17
DistributedTransformerTest.MLP_Layer/3 fails with 6 processes. 


wujingyue:
```
$ _bn && mpirun -np 6 bin/test_multidevice --gtest_filter=DistributedTransformerTest.MLP_Layer/3
```

```
unknown file: Failure
C++ exception with description "Expected H % D == 0 to be true, but got false.
Exception raised from DistributedTransformerTest at /opt/pytorch/nvfuser/tests/cpp/test_multidevice_transformer.cpp:38 (most recent call first):
frame #0: <unknown function> + 0x9b531 (0x5634c93b8531 in bin/test_multidevice)
frame #1: <unknown function> + 0x7a4b0b (0x5634c9ac1b0b in bin/test_multidevice)
frame #2: <unknown function> + 0x7f78e1 (0x5634c9b148e1 in bin/test_multidevice)
frame #3: <unknown function> + 0x7e36a4 (0x5634c9b006a4 in bin/test_multidevice)
frame #4: <unknown function> + 0x7e461b (0x5634c9b0161b in bin/test_multidevice)
frame #5: <unknown function> + 0x7ecc74 (0x5634c9b09c74 in bin/test_multidevice)
frame #6: <unknown function> + 0x7e3c75 (0x5634c9b00c75 in bin/test_multidevice)
frame #7: <unknown function> + 0x11b7cc (0x5634c94387cc in bin/test_multidevice)
frame #8: <unknown function> + 0x29d90 (0x7fc960df7d90 in /usr/lib/x86_64-linux-gnu/libc.so.6)
frame #9: __libc_start_main + 0x80 (0x7fc960df7e40 in /usr/lib/x86_64-linux-gnu/libc.so.6)
frame #10: <unknown function> + 0x11f365 (0x5634c943c365 in bin/test_multidevice)
" thrown in the test fixture's constructor.
```

cowanmeg:
I'll parameterize the input sizes so they scale better with number of gpus. Thanks for catching!

PR #2590, opened 2024-07-13
Tensor parallel multiheaded attention

cowanmeg:
Unit test for tensor parallel multiheaded attention block.

PR #2676, opened 2024-07-24
Backward Tensor parallel MLP

cowanmeg:
Adds backward MLP test. 
Removed decomposed matmul MLP forward test since Attention (https://github.com/NVIDIA/Fuser/pull/2590) always uses ATen matmul and the differences in type promotion made it difficult to maintain a single mlp reference copy. 

Other changes:
- Update transpose scheduler to not merge/split DID axes. 
- Update weight/bias initialization to be scaled by .02 to reduce error. 
- Parameter sizes updated to match GPT sizes (https://github.com/Lightning-AI/lightning-thunder/blob/5fc67dcba844554c8a2390ef8775594e61f18737/thunder/tests/nanogpt_model.py#L268). However, if number heads isn't evenly divisible by number of devices, we scale number of heads and embedding size so that the test can run to avoid: https://github.com/NVIDIA/Fuser/issues/2623
- Refactor MLP forward tests into helper functions


csrc/multidevice/lower_communication.cpp
```diff
@@ -293,11 +293,9 @@ std::vector<Communication*> lowerCommunication(Expr* c) {
 }
 
 bool isLowerableToCommunication(Expr* expr) {
-  NVF_ERROR(
-      ir_utils::isTvOp(expr),
-      "Non-tv op is not supported yet: ",
-      expr->toString());
-  if (expr->isA<ReductionOp>()) {
+  if (!ir_utils::isTvOp(expr)) {
```
wujingyue, line 296: Can you rebase? There are some changes like this that belong to the other PR.

cowanmeg: I thought I could merge the previous PR in before people started reviewing 😂

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -126,134 +132,305 @@ void validate(
 }
 } // namespace
 
-TEST_P(DistributedTransformerTest, MLP_Layer) {
-  auto [use_aten_matmul, dtype] = GetParam();
-  std::unique_ptr<Fusion> fusion = std::make_unique<Fusion>();
-  FusionGuard fg(fusion.get());
-  auto mesh = DeviceMesh::createForNumDevices(communicator_->size());
-  at::ScalarType at_dtype = data_type_to_aten(dtype);
+std::vector<at::Tensor> reference_mlp(
```
cowanmeg, line 135: This is a copy and paste of the old mlp aten reference.

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -126,134 +132,305 @@ void validate(
...
+      matmul1_grad_w,
+      matmul1_grad_b,
+      gelu_grad,
+      matmul0_grad_w,
+      matmul0_grad_b,
+      matmul0_grad};
+  return grads;
+}
+
+std::vector<TensorView*> mlp(
```
cowanmeg, line 190: This is almost a copy and paste of the nvfuser mlp definition BUT the option to have matmuls decomposed into bcast+mul-sum is removed. If we want to keep those tests I can add back

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -126,134 +132,305 @@ void validate(
...
+      matmul1_grad_w,
+      matmul1_grad_b,
+      gelu_grad,
+      matmul0_grad_w,
+      matmul0_grad_b,
+      matmul0_grad_x};
+  return outputs;
+}
+
+TEST_P(DistributedTransformerTest, MLP_Layer) {
```
cowanmeg, line 311: Same as old MLP_Layer just calling the helper functions

samnordmann (approved):
It looks great, thank you very much!

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -26,17 +28,26 @@
 
 namespace nvfuser {
 
+int64_t D = 1, B = 2, E = 768, H = 12, S = 128;
```
wujingyue, line 31: While I understand you made them global so DistributedTransformerTest can modify their values, these global variables have external linkage and can conflict with variables of the same name in other files. AFAICS, they are only used in DistributedTransformerTest, so I'd make them class members.

tests/cpp/test_multidevice_transformer.cpp
wujingyue, line 135: Put this (and several other methods) in an anonymous namespace? Otherwise, they have external linkage, which unnecessarily slows down linking and can run into conflicts.

wujingyue (commented):
LGTM overall. I've yet to check the math of MLP backprop...

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -126,134 +132,305 @@ void validate(
...
-  auto mesh = DeviceMesh::createForNumDevices(communicator_->size());
-  at::ScalarType at_dtype = data_type_to_aten(dtype);
+std::vector<at::Tensor> reference_mlp(
+    at::Tensor x,
+    at::Tensor w0,
+    at::Tensor b0,
+    at::Tensor w1,
+    at::Tensor b1,
+    at::ScalarType at_dtype) {
+  at::manual_seed(0);
```
wujingyue, line 142: Is https://github.com/NVIDIA/Fuser/blob/88f6a6295fc652f10b4c86416744c6b69eb275cc/tests/cpp/utils.h#L500 not good enough?

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -126,134 +132,305 @@ void validate(
...
+  auto linear1 = at::matmul(x, w0).add(b0).to(at::kFloat);
```
wujingyue, line 143:
```suggestion
  auto linear0 = at::matmul(x, w0).add(b0).to(at::kFloat);
```
since you call the params `w0` and `b0`

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -126,134 +132,305 @@ void validate(
...
+  auto gelu = at::gelu(linear1, "tanh");
+  auto linear2 = at::matmul(gelu.to(at_dtype), w1).add(b1).to(at::kFloat);
```
wujingyue, line 145:
```suggestion
  auto linear1 = at::matmul(gelu.to(at_dtype), w1).add(b1).to(at::kFloat);
```

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -126,134 +132,305 @@ void validate(
...
+  MultiDeviceExecutor runtime(
+      std::move(fusion), *communicator_, executor_params_);
+  auto outputs = runtime.runWithInput(inputs);
+  validate(expected_outputs, outputs);
+}
+
+TEST_P(DistributedTransformerTest, MLP_Backward) {
+  auto dtype = GetParam();
+  at::ScalarType at_dtype = data_type_to_aten(dtype);
+  std::unique_ptr<Fusion> fusion = std::make_unique<Fusion>();
```
wujingyue, line 371:
```suggestion
  auto fusion = std::make_unique<Fusion>();
```

tests/cpp/test_multidevice_transformer.cpp
cowanmeg: Is moving the globals into an anonymous namespace sufficient to avoid the global linkage? I avoided putting them as class members so that the helper functions don't have to use `DistributedTransformerTest::D`. This is purely cosmetic though so it's not a must have

wujingyue: > Is moving the globals into an anonymous namespace sufficient to avoid the global linkage? 

Yes. However, even if they are in an anonymous namespace, they are still globals that can potentially be modified anywhere in the file, making code harder to reason about. 

After reading the PR again, do we really need to modify E and H? We could gtest_skip when they are not divisible by D. We could also fix H-per-device (so it's a constant) and compute H as `num_devices * H-per-device` etc.

cowanmeg: Yes, I think it's fine to just skip the tests when E isn't divisible by D. In most cases this is fine (even numbers up to 8) that way everything but D is a constant

wujingyue (approved):
LGTM with comments

cowanmeg:
!build

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -26,21 +28,26 @@
 
 namespace nvfuser {
 
+namespace {
```
wujingyue, line 31: constexpr are internal linkage by default, so I'd remove the anonymous namespace.

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -50,21 +57,9 @@ class DistributedTransformerTest
...
-  // `SdpaFwdOp` currently does not work with ID model since it requires all
-  // sibling outputs to have the same root domain.
-  //  This will be modified in a future PR.
-  preseg_passes::OptimizationPassGuard<preseg_passes::MoveSplitCatPass>
-      optimization_guard_;
-  preseg_passes::OptimizationPassGuard<preseg_passes::AllocationDomainPass>
-      allocation_order_guard_;
+
+ public:
+  const int64_t D; // number of devices
```
wujingyue, line 62:
```suggestion
  const int64_t num_devices_;
```

cowanmeg: I think I'll leave this as D, just for when Attention layer PR gets merged - there are a lot of a reshapes/slices, so using `D` makes it a bit easier to read: https://github.com/NVIDIA/Fuser/pull/2590/files#diff-4e10caf37fae9be5bfb0bc52891227c5ed5141656d0396d5c2a60e06dd1dd1ebR227

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -65,6 +73,8 @@ class DistributedTransformerTest
       optimization_guard_;
   preseg_passes::OptimizationPassGuard<preseg_passes::AllocationDomainPass>
       allocation_order_guard_;
+  preseg_passes::OptimizationPassGuard<preseg_passes::MarkAliasesPreparePass>
```
cowanmeg, line 76: FYI @wujingyue - https://github.com/NVIDIA/Fuser/pull/2639 was causing issues with transpose. It might be as simple as propagating shardings after the alias pass. I will get a small reproducible test.

cowanmeg:
!build

tests/cpp/test_multidevice_transformer.cpp
wujingyue: Sorry I wasn't clear. Normally, I'd expect the random seed to be set only once and at the beginning of the program, so https://github.com/NVIDIA/Fuser/blob/88f6a6295fc652f10b4c86416744c6b69eb275cc/tests/cpp/utils.h#L500 makes sense. I've yet to understand why we need to call `at::manual_seed` at all in individual tests.

cowanmeg: Ahh, it's to ensure the same seed is used for the reference and nvfuser program for dropout.

tests/cpp/test_multidevice_transformer.cpp
wujingyue: > The fix might be as simple as propagating shardings after the alias pass. 

I don't see how that's possible with MultiDeviceExecutor. Yes with FusionExecutorCache. Nonetheless, I'll wait for your repro to understand the problem better.

cowanmeg:
unrelated error in CI, so pushing.

tests/cpp/test_multidevice_transformer.cpp
wujingyue: Got it!

Solution-wise, there's the seed and there's the state. The common pattern I've seen is to `torch.random.fork_rng` the reference implementation, e.g., https://github.com/pytorch/pytorch/blob/53a5e0f1a8ef5087eeb57deb46f4b0c488fdbcc8/torch/jit/_trace.py#L282-L284. This way, the ref and the actual start with the same state without reseting the seed, which is supposed to be set once. Doing this in C++ may be more cumbersome. I haven't tried this, but I saw get_state and set_state get exposed
https://github.com/pytorch/pytorch/blob/53a5e0f1a8ef5087eeb57deb46f4b0c488fdbcc8/aten/src/ATen/cuda/CUDAGeneratorImpl.cpp#L323 and can be accessed via https://github.com/pytorch/pytorch/blob/53a5e0f1a8ef5087eeb57deb46f4b0c488fdbcc8/torch/csrc/api/src/cuda.cpp#L32.

cowanmeg: Dropout exposes the philox seed and the tests do set it (https://github.com/NVIDIA/Fuser/blob/main/tests/cpp/test_multidevice_transformer.cpp#L92)....so this test might be ok without reseting the seed. 
I will experiment!

wujingyue: > I will experiment!

Thanks! 

https://github.com/NVIDIA/Fuser/blob/main/tests/cpp/test_multidevice_transformer.cpp#L92-L94 looks suspicious. Instead of passing in `seed=0,offset=0`, are we supposed to pass in `Generator::seed()` and `Generator::get_offset()`?

cowanmeg: Yes, I think that is correct. Looks like the defualt is seed is 0 which is why the code ran correctly.

PR #2677, opened 2024-07-24
Predicate indexing for circular buffering

### LINKED ISSUE #2159: Invalid unswitch predicates with double buffering
Looks like we are not generating correct predicates when unswitching double buffered loops. The generated code with `DoubleBufferingTest.DoubleBuffer5` seems incorrect. Here's a simplified repro based on that test:

```
TEST_F(DoubleBufferingTest, UnswitchRepro) {
  Fusion fusion;
  FusionGuard fg(&fusion);

  auto tv0 = makeContigTensor(1);
  fusion.addInput(tv0);

  auto tv1 = set(tv0);
  auto tv2 = add(tv1, IrBuilder::create<Val>(1.0));
  fusion.addOutput(tv2);

  tv2->split(-1, 8);
  tv2->split(0, 1, false);
  TransformPropagatorWithCheck propagator(tv2);
  MaxRootDomainInfoSpanningTree(tv2).traverse(&propagator);

  tv1->inlineAt(2);

  tv2->axis(0)->parallelize(ParallelType::Unswitch);
  scheduler_utils::parallelizeAllLike(tv2);

  tv1->doubleBuffer();

  auto t0 = at::randn({16}, options);

  FusionExecutor fe;
  fe.compileFusion(&fusion, {t0});
  auto cg_outputs = fe.runFusion({t0});

  auto ref = t0 + 1;

  testValidate(&fusion, cg_outputs, {t0}, {ref}, __LINE__, __FILE__);
}
```

Here's the unswitched branch of the generated code:

```
 nvfuser_index_t i0;
  i0 = ceilDiv(T0.logical_size[0LL], 8LL);
 if (((-1LL + (8LL * i0)) < T0.logical_size[0LL])) {
    float T1[16LL];
    #pragma unroll
    for(nvfuser_index_t i1 = 0; i1 < 8LL; ++i1) {
      T1[i1]
         = T0[(i1 + nvfuser_zero)];
    }
    NVFUSER_UPDATE_MAGIC_ZERO;
    #pragma unroll 1
    for(nvfuser_index_t i2 = 0; i2 < i0; ++i2) {
      nvfuser_index_t i3;
      i3 = 8LL * i2;
      nvfuser_index_t i4;
      i4 = 8LL + i3;
      nvfuser_index_t i5;
      i5 = 8LL * ((1LL + i2) % 2LL);
      nvfuser_index_t i6;
      i6 = 8LL * (i2 % 2LL);
      #pragma unroll
      for(nvfuser_index_t i1 = 0; i1 < 8LL; ++i1) {
        T1[(i5 + i1)]
           = T0[(i4 + (i1 + nvfuser_zero))];
      }
      NVFUSER_UPDATE_MAGIC_ZERO;
      #pragma unroll
      for(nvfuser_index_t i7 = 0; i7 < 8LL; ++i7) {
        T2[(i3 + (i7 + nvfuser_zero))]
          = T1[(i6 + i7)]
          + (float) 1.00000000000000000e+00;
      }
      NVFUSER_UPDATE_MAGIC_ZERO;
    }
```

The test doesn't fail, but the read from `T0` can overrun the buffer. Indeed, running this test with compute-sanitizer does fail at that point.

I haven't checked if this also applies to circular buffering, but I suspect that's highly likely.

naoyam:
Adding support of predicate indexing with circular buffering.

Circular buffering itself doesn't need many changes, but circular buffering and unswitch/unroll is a bit more complicated. There's an existing [bug](#2159) as well, which is fixed here. 

#2663 could simplify this PR but we probably don't want to enforce epilogue generation. This PR doesn't rely on it.

Fixes #2159


naoyam:
!build

naoyam:
CC: @zasdfgbnm, @rdspring1

naoyam:
!build

naoyam:
!build

naoyam:
Pinging @jacobhinkle

csrc/id_model/utils.h
```diff
@@ -0,0 +1,62 @@
...
+    opts.insert(IdModelEnableOption::UnswitchPredicate);
+  }
+
+  return opts;
+}
+
+inline bool isIdModelOptionEnabled(IdModelEnableOption option) {
+  const auto opts = getIdModelEnabledOptions();
+  return opts.find(option) != opts.end();
+}
```
jacobhinkle, line 60: Nit. You may have other intended uses of this utility that would make use of `getIdModelEnabledOptions()` directly. If not, then you could instead replace this with a `switch (option)` in `isIdModelOptionEnabled`.

csrc/id_model/predicate_indexing.cpp
```diff
@@ -94,6 +94,37 @@ std::unordered_map<Val*, Val*> getPredicateIndexReplacementMap(
...
+    auto circular_buffer_axis =
+        GpuLower::current()->circularBufferInfo().getCircularBufferAxis(tv);
+    if (circular_buffer_axis == nullptr ||
+        !id_model.idGraph(IdMappingMode::LOOP)
+             .disjointValSets()
+             .strictAreMapped(fl->iter_domain(), circular_buffer_axis)) {
+      return nullptr;
+    }
+
+    // Epilog should not hit this part
```
jacobhinkle, line 112: Is this specifically because there is now a different `IterDomain` mapped to each `ForLoop` that are not Loop mapped with one another and `getCircularBufferAxis(tv)` returns the `IterDomain` of the main loop only?

csrc/id_model/predicate_indexing.cpp
```diff
@@ -94,6 +94,37 @@ std::unordered_map<Val*, Val*> getPredicateIndexReplacementMap(
     }
   };
 
+  // If the tensor is circular buffered and the given for-loop is the
+  // main loop of circular buffering, increment the index by
+  // (number_of_stages - 1) since the main loop has a read that is
+  // (number_of_stages - 1) elements ahead.
+  auto replace_for_circular_buffering = [&](ForLoop* fl,
+                                            Val* original_index) -> Val* {
+    auto circular_buffer_axis =
```
jacobhinkle, line 103: Nit: probably a good idea to check this here like in `getCircularBufferLoopStage`:
```suggestion
    NVF_ERROR(
        GpuLower::hasCurrent(),
        "Circular buffering info of GpuLower is required but GpuLower is missing");
    auto circular_buffer_axis =
```

jacobhinkle (approved):
LGTM. Tests make sense.

csrc/id_model/utils.h
naoyam: I don't have any particular plan for this except that it should be retired once everything about IdModel is enabled by default, so I'll leave it as is.

csrc/id_model/predicate_indexing.cpp
naoyam: I'll expand the comment here, but that's just because predication of an expr is based on the consumer. If the control reaches here, it means the tensor is circular buffered. In the epilogue loop, the circular buffer tensor should never appear as a consumer but only as a producer. So, what this part means is that this tensor is a consumer tensor and circular buffered, so the loop stage should never be epilogue.

jacobhinkle: Ah ok. That makes sense. Thanks for the explanation.

Issue #2241, opened 2024-05-14
NoOp scheduler claims zero-dimensional non-trivial fusions


jacobhinkle:
I'm not sure if this is intended behavior, but it led to some small confusion leading to reordering the `ExpressionEvaluator` scheduler before NoOp in #2209. I'm posting this issue since it is confusing for a NoOp scheduler to compute an op.

The following test "fails":
```c++
TEST_F(NVFuserTest, ZeroDimAdd) {
  auto fusion = std::make_unique<Fusion>();
  FusionGuard fg(fusion.get());

  auto tv0 = makeSymbolicTensor(0, DataType::Float);
  auto tv1 = makeSymbolicTensor(0, DataType::Float);
  auto tv2 = add(tv0, tv1);

  fusion->addInput(tv0);
  fusion->addInput(tv1);
  fusion->addOutput(tv2);

  at::Tensor t0 = at::randn({}, at::kFloat).cuda();
  at::Tensor t1 = at::randn({}, at::kFloat).cuda();
  at::Tensor out_ref = t0 + t1;

  FusionExecutorCache fec(std::move(fusion));
  auto out = fec.runFusionWithInputs({t0, t1});

  const FusionKernelRuntime* runtime = fec.getMostRecentKernelRuntime();
  EXPECT_FALSE(runtime->isSegmented());
  const std::vector<FusionExecutor>& executors = runtime->executors();
  EXPECT_EQ(executors.size(), 1);
  // Verify that fusion compilation was not skipped.
  EXPECT_TRUE(executors.front().hasCompiledKernel());

  EXPECT_EQ(
      runtime->schedulerHeuristics()->heuristicsList().front()->heuristic(),
      ScheduleHeuristic::PointWise);
  /*
Expected equality of these values:
  runtime->schedulerHeuristics()->heuristicsList().front()->heuristic()
    Which is: no_op
  ScheduleHeuristic::PointWise
    Which is: pointwise
  */

  EXPECT_TRUE(at::allclose(out[0], out_ref));
}
```
Even though the no-op scheduler claims the segment, it does result in a working fusion and the result is correct. Here is the kernel:
```c++
__global__ void nvfuser_no_op_f0_c1_r0_g0(Tensor<float, 0, 0> T0, Tensor<float, 0, 0> T1, Tensor<float, 0, 0> T2) {
  T2[0]
    = T0[0]
    + T1[0];
}
```
[...]

PR #2679, opened 2024-07-25
The no-op scheduler clears memory space.

wujingyue:
I hit a serde error when I was working on #2639. This PR fixes that error. 

As a known limitation of serde, https://github.com/NVIDIA/Fuser/blob/446e24737cae95154aca5126a038a30b636ad512/csrc/kernel_cache.cpp#L1053-L1080 runs different code paths for default compilation and for deserialization. This usually leads to the same segmented fusion, but the added test case triggers an actual difference if without this PR. The default compilation path calls `findSegments`, which sets global memory type for attempted segment boundaries (via addInput/addOutput and removeInput/removeOutput). The deserialization path however doesn't call `findSegments`, so attempted segment boundaries remain local. As a result, `kernel->summary().global_allocations` is non-empty for default compilation and empty for deserialization. 

This PR resets memory space of intermediate tensors to local, and therefore makes the segmented fusion between the two paths the same.

### DIFF
--- a/csrc/scheduler/no_op.cpp
+++ b/csrc/scheduler/no_op.cpp
@@ -118,6 +118,11 @@ void NoOpScheduler::schedule(Fusion* fusion) {
     return;
   }
 
+  // Make sure we don't have global memory set on intermediate tensors from
+  // fusion segmentation. Otherwise, the generated kernel may unnecessarily
+  // access intermediate buffers. See NoOpTest.ExpandedReduction.
+  scheduler_utils::clearMemorySpace(fusion);
+
   markAliases(fusion);
 }
 
--- a/tests/cpp/test_no_op.cpp
+++ b/tests/cpp/test_no_op.cpp
@@ -8,6 +8,7 @@
 #include <vector>
 
 #include <gmock/gmock-matchers.h>
+#include <gmock/gmock-more-matchers.h>
 #include <gtest/gtest.h>
 
 #include <fusion.h>
@@ -18,6 +19,9 @@
 namespace nvfuser {
 
 using NoOpTest = NVFuserTest;
+
+using testing::IsEmpty;
+using testing::UnorderedElementsAre;
 
 // Simple test case exercising the null scheduler path.
 TEST_F(NoOpTest, FusionNullScheduler) {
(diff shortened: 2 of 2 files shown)


wujingyue:
!build

wujingyue:
For @rdspring1, this is the real root cause behind the serde issue I talked to you about.

tests/cpp/test_no_op.cpp
```diff
@@ -200,4 +204,33 @@ TEST_F(NoOpTest, View) {
...
+                       .ndims(2)
+                       .dtype(DataType::Float)
+                       .contiguity({std::nullopt, std::nullopt})
+                       .shape({2, 3})
+                       .expanded({true, true})
+                       .build();
+  fusion->addInput(in);
+  TensorView* out = sum(in, {0});
+  out = segment_set(out);
+  fusion->addOutput(out);
```
jacobhinkle, line 221: Shouldn't this fusion simplify to something like this?
```c++
auto out = mul(2, squeeze(in, {0}));
```
In such case shouldn't this be accepted by `Pointwise` instead of `NoOp`? Or is it smart enough to figure out that this is actually a zero-dimensional problem instead of 1-dimensional? Does `NVFUSER_DUMP=cuda_kernel` show a non-trivial kernel? If so is this similar to #2241?

wujingyue: > Shouldn't this fusion simplify to something like this?

Yes

> In such case shouldn't this be accepted by Pointwise instead of NoOp? 

Maybe -- I was surprised to see it no-op too. The test case managed to pass the test [here](https://github.com/NVIDIA/Fuser/blob/446e24737cae95154aca5126a038a30b636ad512/csrc/scheduler/no_op.cpp#L83), so it seems to work as intended. The CUDA kernel generated is the following and yes it looks similar to #2241. 

```
[ RUN      ] NoOpTest.ExpandedReduction

======= Codegen output for kernel: nvfuser_no_op_f0_c1_r0_g0 =======

__global__ void nvfuser_no_op_f0_c1_r0_g0(Tensor<float, 2, 2> T0, Tensor<float, 1, 1> T3) {
  float T1[1];
  T1[0] = 0;
  T1[0]
     = T0[0];
  float T2[1];
  T2[0]
    = T1[0]
    * 2.000000000e+00f;
  T3[0]
     = T2[0];
}

======================================

[       OK ] NoOpTest.ExpandedReduction (164 ms)
```

csrc/scheduler/no_op.cpp
```diff
@@ -118,6 +118,11 @@ void NoOpScheduler::schedule(Fusion* fusion) {
     return;
   }
 
+  // Make sure we don't have global memory set on intermediate tensors from
```
rdspring1, line 121: I think `scheduler_utils::clearMemorySpace` is run for all schedulers in their schedule function. e.g., https://github.com/NVIDIA/Fuser/blob/main/csrc/scheduler/pointwise.cpp#L532-L534

rdspring1 (approved):
LGTM.

Maybe we should run `scheduler_utils::clearMemorySpace` at the end of `SegmentedFusion::makeFusion` to avoid this kind of issue?

csrc/scheduler/no_op.cpp
wujingyue: That's right. Sorry if I missed anything -- are you requesting a change in the comment?

wujingyue:
> Maybe we should run `scheduler_utils::clearMemorySpace` at the end of `SegmentedFusion::makeFusion` to avoid this kind of issue?

That's another possible solution. 

On the other hand, the real culprit is asymmetry between addInput/addOutput and removeInput/removeOutput during https://github.com/NVIDIA/Fuser/blob/eda52d39bbadcd47f70808901c699c7afa461414/csrc/fusion_segmenter.cpp#L2477-L2500. addInput/addOutput unconditionally sets global and removeInput/removeOutput does not reset. So yet another possible solution is to let FusionSegmentGuard remember the old "state". 

I'm having a hard time justifying which one is the right solution...

csrc/scheduler/no_op.cpp
rdspring1: No. I was just thinking that `scheduler_utils::clearMemorySpace` seems like a segmentation utility rather than a scheduler one.

jacobhinkle:
> . addInput/addOutput unconditionally sets global and removeInput/removeOutput does not reset. So yet another possible solution is to let FusionSegmentGuard remember the old "state".
> 
> I'm having a hard time justifying which one is the right solution...

Just my .02 to complicate things further :smile: : it's not just memory space that we might want to reset. Currently we don't remove reduction domains of segment inputs or remove producer->logical (formerly known as root->rfactor) transforms in FusionSegmentGuard, even though we do that once segmentation is over and we're ready to schedule. It would be nice if the canSchedule checks could all see the segment pretty much exactly as it will be segmented (Also with forwarded inputs?)

PR #2680, opened 2024-07-25
Run `preseg` before `schedule`.

wujingyue:
This fixes a potential problem that's exposed by #2639, and is AFAIK the right order of actions. We almost never run preseg passes after scheduling, except for multi-GPU.

I could combine this #2639 so the fix is shipped with a regression test. However, the fix seems to be straightforward enough to stand alone, making #2639 a bit smaller. I don't have a strong opinion -- let me know if you prefer the other way.


wujingyue:
!build

jacobhinkle (approved):
LGTM. Seems safer to keep this as a separate PR since it is so small in scope and only affects C++ matmul benchmarks.

PR #2684, opened 2024-07-25
Add an example to demonstrate schedule mismatching reshape

zasdfgbnm:
Now, domains are no longer required to have a nice linear transformation from root domain to loop domain. We can take advantage of that and schedule reshape.

### DIFF
--- a/tests/cpp/test_gpu_view.cpp
+++ b/tests/cpp/test_gpu_view.cpp
@@ -9,6 +9,7 @@
 #include <gmock/gmock-matchers.h>
 #include <gtest/gtest.h>
 
+#include <abstract_tensor.h>
 #include <codegen.h>
 #include <device_lower/analysis/divisible_split.h>
 #include <device_lower/lower2device.h>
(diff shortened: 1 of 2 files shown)


zasdfgbnm:
!build

tests/cpp/test_gpu_view.cpp
```diff
@@ -2619,4 +2620,78 @@ TEST_F(GpuViewTest, GroupNorm) {
...
+  auto tv3 = reshape(tv1, {2, 3, 5}, {2, 3 * 5});
+  auto tv4 = cos(tv2);
+  auto tv5 = exp(tv3);
+  fusion.addOutput(tv4);
+  fusion.addOutput(tv5);
+
+  // schedule all tensors like tv2 and tv4's logical domain
+  std::vector<IterDomain*> tv5_root{
+      // Topological root of tv5, not the root domain of tv5.
+      // TODO: rename root domain as producer domain
```
zasdfgbnm, line 2644: And stop using the word "root" to refer to producers in our communication😉

zasdfgbnm:
!build

tests/cpp/test_gpu_view.cpp
```diff
@@ -2619,4 +2620,87 @@ TEST_F(GpuViewTest, GroupNorm) {
...
+  // TensorView* tv0 = makeContigTensor(3, DataType::Float);
+  fusion.addInput(tv0);
+  auto tv1 = sin(tv0);
+  auto tv2 = reshape(tv1, {2, 3, 5}, {2 * 3, 5});
+  auto tv3 = reshape(tv1, {2, 3, 5}, {2, 3 * 5});
+  auto tv4 = cos(tv2);
+  auto tv5 = exp(tv3);
+  fusion.addOutput(tv4);
+  fusion.addOutput(tv5);
+
```
naoyam, line 2640: Could you give some general description of the scheduling approach? I'm lost at line 2649.

zasdfgbnm: Added

zasdfgbnm:
!build

tests/cpp/test_gpu_view.cpp
```diff
@@ -2619,4 +2620,140 @@ TEST_F(GpuViewTest, GroupNorm) {
...
+  // Before schedule, tv5 is:
+  //
+  //  logical domain: [I0, I1]
+  //
+  // Now, we want to make tv5 as:
+  //
+  //                    I2   I3
+  //                      \ /
+  //  logical domain: [I0, I1]
```
naoyam, line 2660: Why is this necessary? Doesn't the exact graph already have this connection?

zasdfgbnm: Without this, how do you schedule the loop domain of tv5?

naoyam: Ah, but we still need to create an actual transformation, so that's why we want to have this expr for tv5?

naoyam: > Without this, how do you schedule the loop domain of tv5?

For some reason, I was thinking about something like using the loop domain of tv4 directly. I think we chatted a little about this line of ideas like using AbstractTensor to represent loop domains without actually scheduling each tensor.

As an incremental step, this PR makes a lot more sense.

naoyam (approved):
Makes sense now. Wow, this is really exciting. We know this should work, but really interesting demonstration!

tests/cpp/test_gpu_view.cpp
zasdfgbnm: > Ah, but we still need to create an actual transformation, so that's why we want to have this expr for tv5?

Yes

zasdfgbnm: > For some reason, I was thinking about something like using the loop domain of tv4 directly.

Yes, I believe this is what we will do in the long term, because this is the only way I can think of automating the schedule of reshape. But in the short term, as you pointed out, making this work is a good incremental step to take.

PR #687, opened 2023-08-04
Fix unswitch predicate with merge inner path propagation

naoyam:
Fixes #681 and #667

The issue of #681 is due to a bug in generating predicates for unswitched (or unrolled) domains. What unswitch does is, for example, when we have a kernel like:

```
for (i = 0; i < N; ++i) {
   if (PRED(i) < ExtentOfSomeRootDomain && ...) {
      t1[i] = t0[i];
   }
}
```

where `PRED` is a function to produce a predicate based on loop index `i`. With unswitch, this kernel would look like:

```
if (PRED(N-1) < ExtentOfSomeRootDomain && ...) {
  for (i = 0; i < N; ++i) {
      t1[i] = t0[i];
  }
} else {
  for (i = 0; i < N; ++i) {
     if (PRED(i) < ExtentOfSomeRootDomain && ...) {
        t1[i] = t0[i];
     }
  }
}
```

That is, if the maximum of `PRED(i)` with `0 <= i < N` is less than the extent, we can safely eliminate the predicate from the loop body, presumably increasing the performance. An important assumption here is that the maximum of `PRED(i)` is `PRED(N-1)`.

The issue of #681 is that the assumption is not always valid when there's a merge operation between predicated domains to leaf domains because of the modulo used to propagate predicate indices through the merge inner path. For example, think about a 2D tensor of dimensions `[3, 16]`.

```
[3, 16]
// split 16 by 5 
-> [3, 4, 5]
// merge 3 and 4
-> [3 * 4, 5]
// split 12 by 3
-> [4, 3, 5]
```

And assume that only the innermost two leaf domains are unswitched. The initial index of the unswitched domains would be `2` and `4`. The generated code would look like (for simplicity contig predication is ignored):

```
for (i = 0; i < 4 ++i) {
  if ((i * 3 + 2) / 4 < 3 && ((i * 3 + 2) % 4) * 5 + 4 < 16) {
    for (j = 0; j < 3; ++j) {
      for (k = 0; k < 5; ++k) {
        some_math_op_using_the_tensor;
      }
    }
  } else {
    for (j = 0; j < 3; ++j) {
      for (k = 0; k < 5; ++k) {
        if ((i * 3 + j) / 4 < 3 && ((i * 3 + j) % 4) * 5 + k < 16) {
          some_math_op_using_the_tensor;
        }
      }
    }
  }
}    
```

[...]

Issue #681, opened 2023-08-02
Wrong unswitch predicate 


naoyam:
This fusion fails at the validation:

```
TEST_F(NVFuserTest, Issue667Repro_CUDA) {
  Fusion fusion;
  FusionGuard fg(&fusion);

  auto tv0 = makeSymbolicTensor(2);
  fusion.addInput(tv0);

  auto tv1 = sum(tv0, {0, 1});
  fusion.addOutput(tv1);

  // [i0, i1]
  tv1->split(1, 4);
  // [i0, i1/4, 4]
  tv1->merge(0);
  // [i0*i1/4, 4]
  tv1->split(0, 4);
  // [i0*i1/4/4, 4, 4]
  tv1->split(0, 1);
  // [i0*i1/4/4, 1, 4, 4]

  tv1->axis(1)->parallelize(ParallelType::Unswitch);

  fusion.printKernel();

  auto options = at::TensorOptions().dtype(at::kFloat).device(at::kCUDA, 0);
  at::Tensor t0 = at::randn({4, 10}, options);
  std::vector<c10::IValue> aten_inputs = {t0};

  FusionExecutor fe;
  fe.compileFusion(&fusion, aten_inputs);
  auto outputs = fe.runFusion(aten_inputs);

  auto ref = t0.to(at::kDouble).sum();

  testValidate(
      &fusion, outputs, aten_inputs, {ref}, __LINE__, __FILE__);
}
```

I believe this is a simplified repro of the same issue as #667.

The issue is the unswitch. If it's disabled, the validation passes.

Here's the generated kernel. I disabled index hoisting and predicate elimination.
```
__global__ void kernel1(Tensor<float, 2, 2> T0, Tensor<float, 0, 0> T1) {
  NVFUSER_DEFINE_MAGIC_ZERO;
  T1[0] = 0.00000000000000000e+00;
  #pragma unroll 1
  for(nvfuser_index_t i0 = 0; i0 < (ceilDiv((T0.logical_size[0] * (ceilDiv(T0.logical_size[1], 4))), 4)); ++i0) {
    if ((((3 + (4 * i0)) < (T0.logical_size[0] * (ceilDiv(T0.logical_size[1], 4)))) && ((3 + (4 * ((3 + (4 * i0)) % (ceilDiv(T0.logical_size[1], 4))))) < T0.logical_size[1]))) {
      #pragma unroll
      for(nvfuser_index_t i1 = 0; i1 < 4; ++i1) {
[...]

PR #2689, opened 2024-07-25
Make sure unswitch predicates propagate maximum (or minimum) indices through merge-inner or split-outer domains

naoyam:
(Stacked on top of #2677)

The original issue is #681. It was addressed in #687.

This PR is NOT as comprehensive as #687, but my gut feeling is that this should be good enough, in particular since contig indexing would avoid backward traversals through merge in many cases. 

I'll do final more comprehensive comparison with the legacy indexing once contig indexing is done.

Since the original PR and issue were reviewed by @zasdfgbnm, could you please review this too?


tests/cpp/test_gpu3.cpp
```diff
@@ -7535,42 +7535,6 @@ TEST_F(NVFuserTest, SymbolicOneBroadcasting) {
   EXPECT_TRUE(tv->axis(0)->isBroadcast());
 }
 
-// Repro of unswitch predicate issue #681
```
naoyam, line 7538: Just moved to test_indexing.cpp

naoyam:
!build

zasdfgbnm (approved):

csrc/id_model/indexing_utils.h
```diff
@@ -55,5 +55,27 @@ inline IterDomain* getLoopPromotion(
   return loop_promotion_map_it->second;
 }
 
+// Check if unswitching a given for-loop actually matters. For example,
```
naoyam, line 58: @zasdfgbnm Forgot to add this code. I already merged this but let me know if you have any concern.

PR #2698, opened 2024-07-26
Move validateSegmentatation from tests/cpp/utils.h to tests/cpp/validator.h.

wujingyue:
I'm unsure about the difference between tests/cpp/utils.h and tests/cpp/validator.h. AFAICS, tests/cpp/validator.h is only used by tests not benchmarks, so it seems to be the right home for validateSegmentation. 

This PR comes with some other fixes I had to make along the way: 
1. Move non-trivial implementation to cpp files: https://google.github.io/styleguide/cppguide.html#Inline_Functions. 
2. Add missing `#pragma once`s. 
3. Move definitions of external methods in validation_utils.h out of anonymous namespaces.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -236,6 +236,7 @@ list(APPEND NVFUSER_SRCS
   ${NVFUSER_SRCS_DIR}/utils.cpp
   ${NVFUSER_SRCS_DIR}/val_graph.cpp
   ${NVFUSER_SRCS_DIR}/val_graph_visitor.cpp
+  ${NVFUSER_SRCS_DIR}/validator_utils.cpp
 )
 
 # We don't link CUPTI for MSVC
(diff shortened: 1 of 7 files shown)


wujingyue:
!build

Priya2698:
IIRC `validator_utils.h` was created when we moved `getTolerances` and related code outside of the `tests/validator.cpp` directory to allow its use in `fd.validate`. 
We can consider merging `tests/utils` and `tests/validator.cpp` now, if that makes it simpler to use within tests. Earlier, `tests/validator.cpp` also contained the setup code being used by `testValidate` and was a much larger file.

Priya2698 (approved):
We can consider merging `tests/utils` and `tests/validator.cpp`. I don't have a preference at the moment, so LGTM as is.

PR #2703, opened 2024-07-28
Validate all circular buffers at runtime

### LINKED ISSUE #2699: Add runtime check for circular buffer correctness
We generate incorrect results if the size of the pipeline is greater than the size of the circular buffering for-loop.

For example, say that the pipeline size is 4 and circular buffering loop size is 2.
prologue loop is (0, S-1) => (0, 3)
main loop is (0, N-(S-1)) => (0, -1)
epilogue loop is (N-(S-1), N) => (-1, 2)

The proposed fix is to have a runtime check to determine that the entire pipeline is filled at least once.

rdspring1:
This PR fixes #2699. It adds `validateCircularBuffering` to `FusionExecutor::initializeExecutorEntry`. The check ensures that every circular buffer TensorView is filled completely at least once.

Details:

- Add `CircularBufferInfo` from lowering analysis to `kir::KernelSummary`
- Create runtime check to guarantee that the entire pipelined is filled in `executor_utils.cpp`
- Use `limitCircularBufferingSmemOperands` to modify matmul heuristics so that the pipeline is always filled. In the matmul scheduler, the K dimension is the circular buffer axis for A and B operands. Therefore, the number of stages is `min(desired_stage, ceilDiv(ceilDiv(K, splitk_factor), cta_dim.k)`.

### DIFF
--- a/csrc/device_lower/analysis/circular_buffer.cpp
+++ b/csrc/device_lower/analysis/circular_buffer.cpp
@@ -62,10 +62,6 @@ int64_t getCircularBufferAxisPosition(const TensorView* tv) {
   return valid_pos;
 }
 
-IterDomain* getCircularBufferAxis(const TensorView* tv) {
-  return tv->axis(getCircularBufferAxisPosition(tv));
-}
-
 // Initial inspection of a fusion to find and validate circular buffered tensors
 class CircularBufferFusionInspector : private IterVisitor {
  public:
@@ -317,6 +313,10 @@ std::vector<const TensorView*> CircularBufferInfo::getCircularBufferTvs()
         return pair.first;
       });
   return keys;
+}
+
+IterDomain* getCircularBufferAxis(const TensorView* tv) {
+  return tv->axis(getCircularBufferAxisPosition(tv));
 }
 
 } // namespace nvfuser
--- a/csrc/device_lower/analysis/circular_buffer.cpp
+++ b/csrc/device_lower/analysis/circular_buffer.cpp
@@ -49,7 +49,7 @@ int64_t getCircularBufferAxisPosition(const TensorView* tv) {
       "Valid circular buffer axis not found due to Unroll. ",
       tv->toString());
 
-  int64_t valid_pos = tv->getLoopDomain().size();
+  int64_t valid_pos = (int64_t)tv->getLoopDomain().size();
   // Skip parallelized or broadcast axes
   for (int64_t i = unroll_or_ca_pos - 1; i >= 0; --i) {
     auto pt = tv->axis(i)->getParallelType();
--- a/csrc/device_lower/analysis/circular_buffer.cpp
+++ b/csrc/device_lower/analysis/circular_buffer.cpp
@@ -49,7 +49,7 @@ int64_t getCircularBufferAxisPosition(const TensorView* tv) {
       "Valid circular buffer axis not found due to Unroll. ",
       tv->toString());
 
-  int64_t valid_pos = -1;
+  int64_t valid_pos = tv->getLoopDomain().size();
   // Skip parallelized or broadcast axes
   for (int64_t i = unroll_or_ca_pos - 1; i >= 0; --i) {
     auto pt = tv->axis(i)->getParallelType();
@@ -58,7 +58,6 @@ int64_t getCircularBufferAxisPosition(const TensorView* tv) {
       break;
     }
   }
-
   return valid_pos;
 }
 
@@ -308,7 +307,11 @@ std::vector<const TensorView*> CircularBufferInfo::getCircularBufferTvs()
 }
 
 IterDomain* getCircularBufferAxis(const TensorView* tv) {
-  return tv->axis(getCircularBufferAxisPosition(tv));
+  int64_t cb_axis = getCircularBufferAxisPosition(tv);
+  if (cb_axis == (int64_t)tv->getLoopDomain().size()) {
+    return nullptr;
+  }
+  return tv->axis(cb_axis);
 }
 
 } // namespace nvfuser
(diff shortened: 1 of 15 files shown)


rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

csrc/device_lower/analysis/circular_buffer.h
```diff
@@ -29,8 +31,13 @@ class CircularBufferInfo {
 
   void setCircularBufferAxis(const TensorView* tv, IterDomain* id);
 
+  IterDomain* getCircularBufferAxis(const TensorView* tv) const;
```
naoyam, line 34: Can we remove the non-const version? Looks like a duplication.

csrc/device_lower/analysis/circular_buffer.cpp
```diff
@@ -59,19 +59,9 @@ int64_t getCircularBufferAxisPosition(const TensorView* tv) {
     }
   }
 
-  NVF_ERROR(
```
naoyam, line 62: Why is this removed?

naoyam (commented):
LGTM. Just left a couple of minor comments.

@zasdfgbnm, please review the matmul scheduling part.

csrc/device_lower/analysis/circular_buffer.cpp
rdspring1: I moved the `NVF_ERROR` to [validateCircularBufferedTensor](https://github.com/NVIDIA/Fuser/pull/2703/files#diff-189d7a00cf165617e4bf50aca92e4ab4db0d44213f349a3c4c90ab49f8581484R100-R104).

I was thinking that a user would call `getCircularBufferAxis` to evaluate its extent during scheduling. The user is evaluating if they should apply circular buffering, so we shouldn't throw an assertion in this case.

rdspring1:
!build

csrc/device_lower/analysis/circular_buffer.cpp
naoyam: Hmm, if `getCircularBufferAxisPosition` returns -1, doesn't it get interpreted as the innermost position?

rdspring1:
!build

csrc/device_lower/analysis/circular_buffer.cpp
rdspring1: Ah, I forgot about that. I changed `getCircularBufferAxisPosition` to return `getLoopDomain.size()`. `getCircularBufferAxis` will return `nullptr` if it sees `getLoopDomain.size()` as the axis position.

rdspring1:
!build

csrc/scheduler/matmul.cpp
```diff
@@ -1304,8 +1305,21 @@ void scheduleMatmul(Fusion* fusion, const MatmulParams& params) {
...
+        acr_axis->extent()->evaluate() >= number_of_stages) {
+      acr->circularBuffer(number_of_stages);
+    }
+
+    IterDomain* bcr_axis = getCircularBufferAxis(bcr);
+    NVF_ERROR(bcr_axis != nullptr);
+    if (bcr_axis->extent()->isConstScalar() &&
+        bcr_axis->extent()->evaluate() >= number_of_stages) {
+      bcr->circularBuffer(number_of_stages);
+    }
```
jacobhinkle, line 1322: Nit: you could create a lambda to avoid repeating this part.

jacobhinkle: Instead of an if statement, should we just assert that these are constant? That could be done by just removing the `isConstScalar` check in the condition.

rdspring1:
!build

csrc/scheduler/matmul_utils.cpp
```diff
@@ -126,11 +126,19 @@ inline bool initCoreHeuristics(
...
-      constexpr int stages = 3;
-
-      params->circular_buffer_options.circular_buffer_smem_write = true;
+      // Find the number of circular buffer stages so that the entire pipeline
+      // is filled.
+      //
+      // The axes of the mma tensorviews are permuted to [B, M, N, K],
+      // so K / cta_tile_k is the circular buffer axis for both operands.
+      int64_t k_stages =
+          ceilDiv(problem_shape[(size_t)MatmulDimRole::K], cta_tile.k);
```
zasdfgbnm, line 135: @jacobhinkle Is this correct if we have split-k?

jacobhinkle: No, I believe the numerator should be `ceilDiv(problem_shape[(size_t)MatmulDimRole::K], params.splitk_factor)` here. Good catch.

rdspring1:
!build

csrc/scheduler/matmul_utils.cpp
rdspring1: I added this ^^^ to `limitCircularBufferingSmemOperands`, which updates the parameters from `matmul_heuristic_plugin` and `initCoreHeuristics`.

Can `splitk_factor > 1` in the `initCoreHeuristics`?

rdspring1:
!build --matmul-bench

rdspring1:
!build --matmul-bench

csrc/scheduler/matmul_utils.cpp
jacobhinkle: It should not be `>1` in `initCoreHeuristics` at this time.

zasdfgbnm (approved):
LGTM now

PR #2704, opened 2024-07-29
Cleanup `swizzleSharedMemory`

zasdfgbnm:
Smem epilogue tensor is now using TensorIndexer and new swizzle, so there is no need to schedule an "empty" swizzle just to make IDs match. The problem of self mapping still exist, but it can be easily workaround by a much simpler transformation.

### DIFF
--- a/csrc/scheduler/matmul.cpp
+++ b/csrc/scheduler/matmul.cpp
@@ -117,43 +117,29 @@ inline void checkConcreteStaticDim(IterDomain* id) {
 //!  for matmul mainloop and epilogue.
 //! The shared mem data layout is always 2D currently, and this utility
 //!  function assumes that the shared_mem_tv has the following structure:
-//!  [tile_row, tile_col, ***skip***] where the parameter `skip` is the number
-//!  of reduction domains to be skipped. The IDs of tile_row and tile_col are
-//!  the ones being swizzled.
-//! If the input tensorview is not stored in shared memory, the function will
-//! skip the actual swizzle. This is used to help the domain mapping between
-//! mma_result and the epilogue tensor.
+//!  [tile_row, tile_col]
 //! Returns the domain with swizzle. For the case of legacy swizzle, this
 //! domain must be set as loop domain. For the case of new swizzle, this domain
 //! must be set as allocation domain.
 template <bool legacy = true>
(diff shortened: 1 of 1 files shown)


zasdfgbnm:
!build

zasdfgbnm:
!build --diff

csrc/scheduler/matmul.cpp
```diff
@@ -1021,16 +1001,14 @@ void scheduleMatmul(Fusion* fusion, const MatmulParams& params) {
...
-    mma_result->setLoopDomain(swizzled_dom.as<IterDomain*>());
+  if (params.use_smem_epilogue && params.splitk_factor != 1) {
+    // TODO:
+    // This is a workaround for a problem that different dimensions in the loop
+    // domain are mapped in the loop graph of IdModel due to the mapping of
+    // compliment IDs. We should remove forwarding completely, and remove this
+    // workaround.
+    mma_result->split(-2, 1);
+    mma_result->merge(-3);
   }
```
jacobhinkle, line 1012: If I understand correctly, this is splitting the second to last dimension (`Ni`) and merging it back, so that it does not get loop mapped with `smem_epilogue`?

jacobhinkle (approved):
The main change LGTM. I am not sure I understand what goes wrong without the small fix for smem_epilogue+splitk.

csrc/scheduler/matmul.cpp
zasdfgbnm: Yes, this is splitting the `Ni` and merging it back. The purpose is not to prevent it from being mapped to `smem_epilogue`.  But to prevent different dimensions of the loop domain of the bias tensor being mapped together. Generally, when we have a loop domain of a tensor as `[I1, I2, I3]`, then `I1`, `I2` and `I3` should never be mapped with each other in the loop graph, because by definition, they are different loops. However, due to the mapping of compliment IDs (which is a workaround introduced to fix loop promotion by mapping non-equivalent IDs together), these  IDs might be mapped together, which is clearly a mistake. I did not study in more detail why mapping compliment IDs could cause trouble in the bias tensor, but whatever reason it is, I believe forwarding and mapping compliment IDs is just fundamental wrong, and any effort that tries to "improve" it is a waste of time. So I decided to save my time and pick whatever workaround that takes me least amount of time to implement.

PR #2709, opened 2024-07-29
Some refactor of matmul schedule

zasdfgbnm:
- Remove `TensorView::is_mma_swizzled_`. This flag was designed to prevent further schedule of loop domain, which I don't think makes sense any more because we no longer have a root->loop linear transformation history.
- Rename `WarpMmaSwizzler` as `MmaSwizzler`. Mma swizzling is no longer limited to a warp, but can also be a warp group.
- Rename `scheduleMmaWarpOutput` as `scheduleMmaOutputAllocation`. The thing that this function schedules is the allocation domain. The loop domain may be the same as allocation domain, or may not (for the planned hopper).
- Make `scheduleMmaOutputAllocation` accept an `AbstractTensor` and return an `AbstractTensor`.
- Stop using `applyMmaSwizzle` to schedule matmul output. We should prefer using `scheduleMmaOutputAllocation` directly.


zasdfgbnm:
!build

csrc/mma_type.h
```diff
@@ -210,7 +210,7 @@ enum class MmaLayout { NT = 0, TT, TN, NN };
 enum class UnitDim { K, M_or_N };
 
 //! Utility to annotate which input of mma this option struct describes
-enum class MmaOperand { Accumulator = 0, A, B };
+enum class MmaOperand { A, B };
```
jacobhinkle, line 213: Off-topic: could we instead just use `MatmulTensorRole`?

csrc/scheduler/mma_utils.cpp
```diff
@@ -1029,63 +1023,62 @@ void WarpMmaSwizzler::scheduleOperandRead(
...
   // Hopper one is the most general one. For earlier architectures, we will have
   // some dimensions with size 1 after split, this is fine.
   // Memory format for hopper mma:
   // https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#wgmma-64n16-d
 
   // Assume last 2 dims, for example [M64, N24] or [M64, N24, R]
-  NVF_ERROR(tv->nDims() >= 2);
-  bool is_mma_output = tv->definition()->isA<MmaOp>();
+  NVF_ERROR(t.size() >= 2);
+  bool has_reduction = t[-1]->isReduction();
```
jacobhinkle, line 1035: Should this be an assert now? My understanding is that we no longer apply this to anything but `mma_result`.

jacobhinkle (approved):
LGTM

csrc/scheduler/mma_utils.cpp
zasdfgbnm: No, today, we still apply this function to the consumer of `mma_result` in some unit tests. I believe these tests do so because this is just a convenient utility to manually schedule some tensor for prototyping purposes. For example, if I have a fusion:

```C++
T4_l = mma(...);
T5_g = set(T4_l);
```

Then I can just call this function on `T5_g` to schedule it like `T4_l`. For this PR, let's preserve this behavior. But in the future, I am not a fan of supporting both variants. So I don't mind create a separate PR to clean this up.

csrc/mma_type.h
zasdfgbnm: I think we should. Let's create a separate PR to clean this up after this PR is being merged.

csrc/scheduler/mma_utils.cpp
zasdfgbnm: For non-reduction tensor, I believe we can just push a placeholder as the reduction dimension: https://github.com/NVIDIA/Fuser/pull/2725

PR #2691, opened 2024-07-26
Predication of non-divisible splits

naoyam:
(Stacked on #2677)

Basically just porting what's already done with the current predication

PR #2712, opened 2024-07-29
Fix finding right non-divisible splits to predicate

naoyam:
This is a bug fix for #2691, which isn't sufficient when broadcast tensors are involved.

### DIFF
--- a/csrc/id_model/indexing.cpp
+++ b/csrc/id_model/indexing.cpp
@@ -1168,7 +1168,7 @@ std::vector<PredicateInfo> TensorIndexer::getPredicates(
   if (!lower_utils::isReductionInitExpr(expr)) {
     for (const auto& [eg, direction] : index_info.traversal_path) {
       // NOTE: Fundamentally, the problem of non divisiblity should be
-      // checked while traversing the indexng path. Currently, it uses
+      // checked while traversing the indexing path. Currently, it uses
       // the information gathered in a tensor-by-tensor basis. This
       // should be fine currently, but may not work if, e.g., the
       // indexing path involved both backward and forward traversals.
--- a/csrc/id_model/indexing.cpp
+++ b/csrc/id_model/indexing.cpp
@@ -976,9 +976,9 @@ IndexingInfo TensorIndexer::computeIndex(
 
   // Initialize the loop dependency mappings
   for (const auto& loop_domain : loop_domains) {
-    const auto& traversal_graph_group =
-        traversalGraph().toGroup(loop_domain);
-    const auto& loop_graph_group = id_model_.idGraph(IdMappingMode::LOOP).toGroup(loop_domain);
+    const auto& traversal_graph_group = traversalGraph().toGroup(loop_domain);
+    const auto& loop_graph_group =
+        id_model_.idGraph(IdMappingMode::LOOP).toGroup(loop_domain);
     loop_group_dependencies[traversal_graph_group].pushBack(loop_graph_group);
   }
 
--- a/csrc/id_model/indexing.cpp
+++ b/csrc/id_model/indexing.cpp
@@ -968,11 +968,38 @@ IndexingInfo TensorIndexer::computeIndex(
 
   IdGraphIndexCompute index_compute(traversalGraph(), initial_index_map);
 
-  for (const auto& [expr_group, direction] : traversal_path) {
-    index_compute.propagate(expr_group, direction);
+  // In addition to indeices themselves, compute the dependency of
+  // each ID group to loop ID groups
+  std::unordered_map<ValGroup, ValGroups> loop_group_dependencies;
+
+  // Initialize the loop dependency mappings
+  for (const auto& loop_group : loop_groups) {
+    loop_group_dependencies[loop_group].pushBack(loop_group);
   }
 
-  IndexingInfo info{loop_domains, traversal_path, index_compute.indexMap()};
+  for (const auto& [expr_group, direction] : traversal_path) {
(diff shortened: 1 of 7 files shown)


naoyam:
!build

tests/cpp/test_indexing.cpp
naoyam, line 1: A lot of changes in this file are because I realized `IdModel` is also required to generate correct reference results. The main change is the newly added test.

csrc/id_model/indexing.cpp
```diff
@@ -1125,9 +1125,20 @@ std::vector<PredicateInfo> TensorIndexer::getPredicates(
   // If this is a reduction init expr, then no need to take care of
   // non divisible splits
   if (!lower_utils::isReductionInitExpr(expr)) {
-    for (const PredicateDomainInfo& pred_info :
-         non_divisible_split_predicates) {
-      IterDomain* non_divisible_domain = pred_info.id;
+    for (const auto& [eg, direction] : index_info.traversal_path) {
```
naoyam, line 1128: `pred_info` has all the non-divisible splits of the consumer tensor, but they may not be the ones that should be added here since the actual indexing path may not include them. See the added test. The broadcast tensor, `tv2`, has a non-divisible split, but the real domain that needs to be predicated is the one showing up in its indexing path, which is the split in the `tv3` tensor.

tests/cpp/test_indexing.cpp
```diff
@@ -4081,4 +4126,178 @@ TEST_P(PredicateIndexingTest, UnswitchPredicateIssueRepro681) {
...
+          pred, ltExpr(logical_id0_stop_idx, ref_logical_id0->extent()));
+
+      // threadIdx.y * 16 + threadIdx.x < 100
+      pred = andExpr(
+          pred,
+          ltExpr(
+              non_divisible_domain_stop_idx,
+              non_divisible_split_to_predicate->in()->extent()));
+
+      // TODO: Consolidating unswitch predicates isn't
```
naoyam, line 4261: This is another oversight I've just realized. Will address in a follow-up PR.

naoyam:
Didn't realize the CI failures. They seem to be due to the consolidation problem. Moved this PR back to draft.

naoyam:
!build

csrc/predicate_compute.cpp
```diff
@@ -255,19 +255,53 @@ UnswitchPredicateKey::UnswitchPredicateKey()
...
   }
 
   std::vector<Val*> all_parallelized_consumer_loop_ids;
-  std::copy_if(
-      consumer_tv->getLoopDomain().begin(),
-      consumer_tv->getLoopDomain().end(),
-      std::back_inserter(all_parallelized_consumer_loop_ids),
-      [](IterDomain* x) { return isParallelTypeThread(x->getParallelType()); });
+
+  // Identify which parallel type is used for which loop domain for
```
naoyam, line 270: Added this fix. `UnswitchPredicateKey` was mainly designed for supporting shift predicates. I'll clean do some cleanup later.

naoyam:
> Didn't realize the CI failures. They seem to be due to the consolidation problem. Moved this PR back to draft.

Added a fix for `UnswitchPredicateKey`. It's a bit ugly WAR as I don't want to break the existing usage of the class.

tests/cpp/test_indexing.cpp
```diff
@@ -4081,4 +4126,285 @@ TEST_P(PredicateIndexingTest, UnswitchPredicateIssueRepro681) {
...
+
+  // All tensors are unswitched but with different TID
+  // parallelization. tv2 and tv4 are parallelized as [TIDy, TIDx],
+  // whereas tv3 is as [TIDx,TIDy]. These two domains are both
+  // unswitched, so the unswitch predicate should include both
+  // patterns: one with TIDy and TIDx and another with TIDx and
+  // TIDy. Specifically, the unswitch predicate should consist of:
+  //
+  // (bidx * 8 + tidy) * 16 + tidx
+  // (bidx * 8 + tidx) * 16 + tidy
```
jacobhinkle, line 4336:
```suggestion
  // (bidx * 8 + tidy) * 16 + tidx < i0
  // (bidx * 8 + tidx) * 16 + tidy < i0
```

csrc/id_model/predicate_indexing.h
```diff
@@ -35,4 +35,70 @@ std::unordered_map<Val*, Val*> getPredicateIndexReplacementMap(
...
+  //
+  // - Predicate for the outer logical domain
+  // - Predicate for the inner logical domain
+  // - Predicate for the non-divisible split
+  //
+  // Here, the last two predicates are redundant since both of them
+  // guard the index with respect to the domain of extent 8, which is
+  // redundant. This is a bit annonying but should have no actual
+  // impact as the redundancy should be removed by the expression
+  // simplifier.
```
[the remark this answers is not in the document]
naoyam: For `t0`, since its logical domain is the one before the reshape, we would generate:

```
i2 + i1 * 3 + i0 * (ceilDiv(8, 3) * 3) < tv0.logical_shape[0]
```
and because of the ceilDiv, we would also generate:
```
i2 + i1 * 3 < 8
```

The latter is the non-divisible predicate for `t0`. 

For `t1`, since its logical domain is `[2, 8]`, we would generate:

```
i0 < 2
i2 + i1 * 3 < 8
```
And due to the ceilDiv, we would also generate:
```
i2 + i1 * 3 < 8
```

This is also interesting because reshape is always divisible, we can predicate `[2, 8]` for `t0` too. That would probably help index hoisting.

PR #2717, opened 2024-07-30
benchmarking thunder.jit, group norm

liqiangxl:


benchmarks/python/test_groupnorm_fwd.py
```diff
@@ -121,6 +127,24 @@ def test_groupnorm_fwd_nvf_benchmark(
...
+    dtype: torch.dtype,
+):
+    clear_cuda_cache()
+    N, C, H, W = size
+    x = torch.randn(size, device="cuda", dtype=dtype, requires_grad=True)
+    weight = torch.randn(C, device="cuda", dtype=dtype, requires_grad=True)
+    bias = torch.randn(C, device="cuda", dtype=dtype, requires_grad=True)
+    num_groups = get_n_goups(C)
+    # thunder compiled model
+    groupnorm_fwd_jit = thunder.jit(groupnorm_fwd, nv_enable_bookend=False)
```
IvanYashchuk, line 144: Do you want to test group norm inference forward or training forward? The difference between the two is in the number of saved intermediates. `disable_torch_autograd_support=True` would stop Thunder from implicitly generating the backward trace and the forward will be the inference forward.

thunder.jit's default executors might change, I suggest using an explicit list of executors:
```suggestion
    from thunder.executors.nvfuserex import nvfuserex
    groupnorm_fwd_jit = thunder.jit(groupnorm_fwd, nv_enable_bookend=False, executors=[nvfuserex])
```

liqiangxl: Good point! Changed to explicitly using nvFuser.

In this case, the output seems to be controlled only by `requires_grad=False/True` and `disable_torch_autograd_support` doesn't seem to influence the output. I tested the following 4 cases:
- case-1, `requires_grad=True for x, weight, bias ` and `disable_torch_autograd_support=False` -->3 outputs, `groupNorm(x), mean(x), rsqrt(var)`
- case-2, `requires_grad=False for x, weight, bias ` and `disable_torch_autograd_support=False` --> 1 output, `groupNorm(x)`
- case-3, `requires_grad=True for x, weight, bias ` and `disable_torch_autograd_support=True`--> 3 outputs, `groupNorm(x), mean(x), rsqrt(var)` (same to case-1)
- case-4, `requires_grad=False for x, weight, bias ` and `disable_torch_autograd_support=True`--> 1 output, `groupNorm(x)`(same to case-2)

The additional output of ` mean(x), rsqrt(var)` won't cause a large influence on performance, but we should be consistant for different tests and we can also make it a test para `@pytest.mark.parametrize("require_grad", [False, True])`. Will check with the team about which version to use. @kevinstephano

liqiangxl: talked with @kevinstephano, we want to prioritize training forward

liqiangxl:
> Lets disable thunder benchmarks by default, similar to `compile` and `eager`. The current plan is to run thunder benchmarks locally for comparison.
> 
> We can use something like this: https://docs.pytest.org/en/latest/example/simple.html#control-skipping-of-tests-according-to-command-line-option.
> 
> It seems that `pytest.config` is deprecated but another option is: (https://stackoverflow.com/questions/51883573/using-a-command-line-option-in-a-pytest-skip-if-condition)
> 
> ```
> @pytest.mark.skipif(not _pytest.config.get_config().getoption('--benchmark-thunder'), reason="--need --benchmark-thunder option to run")
> ```

I added `--benchmark-thunder` following `--benchmark-eager`, I think we can change them at the same time if new option is prefered.

benchmarks/python/test_groupnorm_fwd.py
```diff
@@ -6,9 +6,23 @@
 from nvfuser.pytorch_utils import torch_dtype_to_nvfuser_dtype
 from .core import run_benchmark, clear_cuda_cache, clear_dynamo_cache
 import torch
+import thunder
+from thunder.executors.nvfuserex import nvfuserex
 from .global_params import generate_input_sizes, FLOAT_DTYPES, PROMOTE_DTYPES
 
 
+def get_n_goups(C):
```
Priya2698, line 14:
```Suggestion
def get_n_groups(C):

File: benchmarks/python/conftest.py [hunk 3]
```diff
@@ -96,6 +100,7 @@ def pytest_collection_modifyitems(session, config, items):
         compile = true: torch.compile benchmark
     """
     run_eager = config.getoption("--benchmark-eager")
+    run_thunder = config.getoption("--benchmark-thunder")
```
Priya2698, line 103: This alone will not filter out the `thunder` benchmarks. The code following this removes the tests corresponding to that option.

By default, we are still running both.

liqiangxl: added and tested with `pytest ./benchmarks/python/test_groupnorm_fwd.py -s -v -k "bfloat16 and 512_2_4_4"`
Get:
```
[2024-08-09 08:08:45] benchmarks/python/test_groupnorm_fwd.py::test_groupnorm_fwd_nvf_benchmark[dtype=torch.bfloat16-size=[512_2_4_4]] PASSED
[2024-08-09 08:08:48] benchmarks/python/test_groupnorm_fwd.py::test_groupnorm_fwd_thunder_benchmark[dtype=torch.bfloat16-size=[512_2_4_4]] SKIPPED (need --benchmark-thunder option to run)
[2024-08-09 08:08:48] benchmarks/python/test_groupnorm_fwd.py::test_groupnorm_fwd_baseline_benchmark[dtype=torch.bfloat16-size=[512_2_4_4]-eager] SKIPPED (need --benchmark-eager option to run)
[2024-08-09 08:08:48] benchmarks/python/test_groupnorm_fwd.py::test_groupnorm_fwd_baseline_benchmark[dtype=torch.bfloat16-size=[512_2_4_4]-compile] SKIPPED (need --benchmark-torchcompile option to run)
```

benchmarks/python/test_groupnorm_fwd.py
liqiangxl: changed.

Priya2698 (approved):
LGTM.
Sorry about the delay, I thought I stamped it already.

benchmarks/python/conftest.py
```diff
@@ -121,3 +126,9 @@ def pytest_collection_modifyitems(session, config, items):
                 and item.callspec.params["compile"]
             ):
                 item.add_marker(skip_torchcompile)
+
+    if not run_thunder:
+        skip_thunder = pytest.mark.skip(reason="need --benchmark-thunder option to run")
+        for item in items:
+            if "thunder_benchmark" in item.nodeid:
```
Priya2698, line 133: I would filter only on `thunder` to avoid issues from ordering in the name.

Priya2698 (changes requested):

Priya2698 (approved):
Thanks for making the changes.
LGTM.

PR #2719, opened 2024-07-30
Introduce multi matmul scheduler

jacobhinkle:
This is the first in a series of PRs that represents a refactor of the matmul scheduler to increase the flexibility in the number of operands and matmul patterns we can schedule in a single kernel. This PR:
- Introduces the `MultipleMatmulScheduler` class which will eventually implement and extend the functionality in the current `scheduleMatmul`.
- Introduces a new option `EnableOption::FuseMultipleMatmuls` which enables this scheduler in place of the default matmul scheduler.
- Implements some basic functionality in `MultipleMatmulScheduler`. Enough to translate matmul patterns and define most of the relevant tensors (`splitk_sum` and `smem_epilogue` will be defined and all tensors will be scheduled in later PRs).

In the next PRs I will
- Schedule the prologue and introduce parametrized tests to check that the scheduled fusion matches the current scheduler
- Schedule epilogue tensors without split-K and smem_epilogue
- Schedule split-K sum
- Schedule smem_epilogue
- Update the heuristic to adjust parameters when multiple matmul patterns are detected. This is needed to ensure we don't use register/smem buffers that are too large.
- Update canSchedule checks to accept multiple matmuls when the multi matmul scheduler is enabled


jacobhinkle:
!build --matmul-bench

csrc/scheduler/multi_matmul.cpp
```diff
@@ -0,0 +1,421 @@
...
+// by building AbstractTensors for each tensor category; these are held in
+// MultipleMatmulScheduler::schedules_.
+class MultipleMatmulScheduler {
+ public:
+  MultipleMatmulScheduler(Fusion* fusion, const MatmulParams& params)
+      : fusion_(fusion),
+        params_(params),
+        id_model_(fusion, /*build_graphs=*/false) {}
+
+  void run() {
```
zasdfgbnm, line 92: Nice to see that matmul scheduler finally becomes modular.

csrc/scheduler/multi_matmul.cpp
```diff
@@ -0,0 +1,421 @@
...
+    for (size_t i : c10::irange(orig_logical.size())) {
+      ValGroup vg = graph_->toGroup(orig_logical[i]);
+      graph_->initializeVal(cache_logical[i], vg);
+    }
+
+    return c;
+  }
+
+  void scheduleMmaResults() {
+    NVF_ERROR(false, "scheduleMmaResults is not yet implemented");
```
zasdfgbnm, line 366: Thank you for making this PR small

csrc/scheduler/multi_matmul.cpp
```diff
@@ -0,0 +1,421 @@
...
+        smem_operands[i]->definition()->as<LoadStoreOp>()->setCacheOp(cache_op);
+        if (smem_operands[i]->uses().size() > 1) {
+          // There can be multiple uses for example if we have A @ B1 + A @ B2
+          // then A will be cached to smem then it might be loaded into two
+          // separate register buffers, one for each mma. Instead, we will load
+          // it once into registers then re-use the register buffer for both
+          // mmas.
+          cacheAfter(smem_operands[i]);
+        }
+        NVF_ERROR(smem_operands[i]->uses().size() == 1);
```
zasdfgbnm, line 289: Should we move this to `addSetsForCacheReads`? I believe the
```C++
if (auto ldst = dynamic_cast<LoadStoreOp*>(tv_smem->uses().at(0))) {
  ...
} else {
  ...
}
```
in `addSetsForCacheReads` will automatically handle the case where there are multiple uses, and only use a single ldmatrix?

csrc/scheduler/multi_matmul.cpp
```diff
@@ -0,0 +1,421 @@
...
+
+      // Setup accumulator register.
+      mma_results_.push_back(mma->out()->as<TensorView>());
+    }
+
+    LoadStoreOpType load_op = params_.async_gmem_load_operands
+        ? LoadStoreOpType::CpAsync
+        : LoadStoreOpType::Set;
+
+    auto cacheOperandsToSmem = [&](const std::vector<TensorView*>& operands,
```
zasdfgbnm, line 252: Should we make this a member function?

csrc/scheduler/multi_matmul.cpp
```diff
@@ -0,0 +1,421 @@
...
+    cacheOperandsToSmem(as_, acw_smems_, params_.supported_vec_size.a);
+    cacheOperandsToSmem(bs_, bcw_smems_, params_.supported_vec_size.b);
+
+    // We add two LoadStore operators to the inputs of our fusions. The first
+    // one is for a read from global memory and the second one (below) is for a
+    // cache read. As an optimizaton, we avoid adding an operator if there's an
+    // existing LoadStoreOp present. Please note that for the second LoadStore
+    // we don't propagate the allocation domain, since the scheduler sets the
+    // allocation domain in the registers.
+    auto addSetsForCacheReads = [&](const std::vector<TensorView*>& tv_smems,
```
zasdfgbnm, line 302: Should we make this a member function?

csrc/scheduler/multi_matmul.cpp
```diff
@@ -0,0 +1,421 @@
...
+      LoadStoreOpType op_type = LoadStoreOpType::Set,
+      CacheOp cache_op = CacheOp::AllLevels,
+      bool propagate_allocation_domain = false) {
+    const std::vector<IterDomain*> orig_alloc =
+        orig->getMaybeAllocationDomain();
+
+    TensorView* c =
+        orig->cacheAfter(op_type, cache_op, propagate_allocation_domain);
+
+    if (propagate_allocation_domain) {
```
zasdfgbnm, line 339: Why not always add allocation domain to ValGraph? Or, maybe better, could we just add `tv->domain()->allIDs()`?

csrc/scheduler/multi_matmul.cpp
```diff
@@ -0,0 +1,421 @@
...
+        TensorDomain::noReductions(orig->getLogicalDomain());
+    const std::vector<IterDomain*> cache_logical = c->getLogicalDomain();
+    // in split-K we do rFactor which gives us a full = sum(partial)
+    // where partial has root domain that matches the logical domain of the
+    // original tensor. The logical domain contains Iteration transforms of the
+    // Reduction axis in the original mma output.
+    NVF_ERROR(orig_logical.size() == cache_logical.size());
+    for (size_t i : c10::irange(orig_logical.size())) {
+      ValGroup vg = graph_->toGroup(orig_logical[i]);
+      graph_->initializeVal(cache_logical[i], vg);
```
zasdfgbnm, line 359: Is this sufficient? Should we map these domains with its producer and consumer?

zasdfgbnm (approved):
Left some comments, none of them are blocking.

csrc/scheduler/multi_matmul.cpp
jacobhinkle: Yeah that's a good point. I figured if we didn't propagate allocation domain then it would be empty anyway. But it could contain transforms so we should initialize the whole path not just the IDs in the alloc domain itself. I _think_ we should have the same ordering of IDs when using `orig->allIDs()` and `c->allIDs()`. I need that correspondence in order to add them to the correct group. I will give this a try (replacing both these loops with one over `allIDs`) with a note about the correspondence.

csrc/scheduler/multi_matmul.cpp
```diff
@@ -0,0 +1,426 @@
...
+      }
+    }
+
+    const std::vector<IterDomain*> orig_logical =
+        TensorDomain::noReductions(orig->getLogicalDomain());
+    const std::vector<IterDomain*> cache_logical = c->getLogicalDomain();
+    // in split-K we do rFactor which gives us a full = sum(partial)
+    // where partial has root domain that matches the logical domain of the
+    // original tensor. The logical domain contains Iteration transforms of the
+    // Reduction axis in the original mma output.
```
jacobhinkle, line 361: @zasdfgbnm this code comment is relevant to your suggestion to use `tv->allIDs()` and map all the corresponding `ValGroup`s. If we have an rfactored partial reduction it can have `Iteration` transforms of a `Reduction` axis and vice versa. Something like this happens for split-K with smem epilogue since we place the smem tensor between the partial and full reduction tensors using `partial->cacheAfter()`:
```
original mma_result: T12_l[ iS194{( ceilDiv(i0, 256) )}, iS196{( ceilDiv(i4, 128) )}, iS200{2}rf, rS201{( ceilDiv(( ceilDiv(i2, 32) ), 2) )}rf, iS195{256}, iS197{128}, rS199{32}rf ]
 root domain : (iS191{i0}, iS192{i4}, rS193{i2}rf)
  Split: rS193{i2}rf by factor 32 -> iS198{( ceilDiv(i2, 32) )}rf, rS199{32}rf
  Outer split: iS198{( ceilDiv(i2, 32) )}rf by factor 2 -> iS200{2}rf, rS201{( ceilDiv(( ceilDiv(i2, 32) ), 2) )}rf
 logical domain : (iS191{i0}, iS192{i4}, iS200{2}rf, rS201{( ceilDiv(( ceilDiv(i2, 32) ), 2) )}rf, rS199{32}rf)
 contiguity: t t t n n
  Split: iS191{i0} by factor 256 -> iS194{( ceilDiv(i0, 256) )}, iS195{256}
  Split: iS192{i4} by factor 128 -> iS196{( ceilDiv(i4, 128) )}, iS197{128}
 loop domain : (iS194{( ceilDiv(i0, 256) )}, iS196{( ceilDiv(i4, 128) )}, iS200{2}rf, rS201{( ceilDiv(( ceilDiv(i2, 32) ), 2) )}rf, iS195{256}, iS197{128}, rS199{32}rf)
 mma_result after cacheAfter(): T12_l[ iS194{( ceilDiv(i0, 256) )}, iS196{( ceilDiv(i4, 128) )}, iS200{2}rf, rS201{( ceilDiv(( ceilDiv(i2, 32) ), 2) )}rf, iS195{256}, iS197{128}, rS199{32}rf ]
 root domain : (iS191{i0}, iS192{i4}, rS193{i2}rf)
  Split: rS193{i2}rf by factor 32 -> iS198{( ceilDiv(i2, 32) )}rf, rS199{32}rf
  Outer split: iS198{( ceilDiv(i2, 32) )}rf by factor 2 -> iS200{2}rf, rS201{( ceilDiv(( ceilDiv(i2, 32) ), 2) )}rf
 logical domain : (iS191{i0}, iS192{i4}, iS200{2}rf, rS201{( ceilDiv(( ceilDiv(i2, 32) ), 2) )}rf, rS199{32}rf)
 contiguity: t t t n n
  Split: iS191{i0} by factor 256 -> iS194{( ceilDiv(i0, 256) )}, iS195{256}
  Split: iS192{i4} by factor 128 -> iS196{( ceilDiv(i4, 128) )}, iS197{128}
 loop domain : (iS194{( ceilDiv(i0, 256) )}, iS196{( ceilDiv(i4, 128) )}, iS200{2}rf, rS201{( ceilDiv(( ceilDiv(i2, 32) ), 2) )}rf, iS195{256}, iS197{128}, rS199{32}rf)
cache after cacheAfter(): T13_l[ iS212{( ceilDiv(i0, 256) )}, iS214{( ceilDiv(i4, 128) )}, iS211{2}, iS213{256}, iS215{128} ]
 logical domain : (iS209{i0}, iS210{i4}, iS211{2})
 contiguity: t t t
  Split: iS209{i0} by factor 256 -> iS212{( ceilDiv(i0, 256) )}, iS213{256}
  Split: iS210{i4} by factor 128 -> iS214{( ceilDiv(i4, 128) )}, iS215{128}
 loop domain : (iS212{( ceilDiv(i0, 256) )}, iS214{( ceilDiv(i4, 128) )}, iS211{2}, iS213{256}, iS215{128})
```
Determining the corresponding IDs to do the mapping here is tricky since any of the original reduction domains are missing in the cache tensor, and because the pre-cache tensor has root->logical transforms while the cache tensor does not.

Note that I only need this graph in order to determine the leaf and allocation dim roles, so if the other IDs are not mapped it is not really a problem for me. However, a more general way to incrementally update the IdModel's graphs during scheduling (creation of tensors with cache* and rFactor) would be generally useful it seems.

csrc/scheduler/multi_matmul.cpp
jacobhinkle: Here I am just adding the ID to the `ValGroup` in the producer. Those should already be mapped appropriately with the consumer if the graph is up to date before this point I think.

zasdfgbnm: Oh, I see

csrc/scheduler/multi_matmul.cpp
zasdfgbnm: So you will not use the same IdModel for transform propagation?

jacobhinkle: Currently I'm not using IdModel to propagate transforms anywhere; this is because of the questions that came up around forwarding around merges with missing axes. You're right though that in the future it would be nice to be able to keep the permissive graph intact (or something like it like EXACT+concretized bcasts as we discussed at the on-site meeting recently). To do that, I think it would probably be best to actually do the mapping at the TV level, using some `IdModel` method that would update all of the graphs. Then we would just build the IdModel once, modify its Exact graph to map concretized broadcasts into their Iteration groups, then whenever we create a new TV it would be updated just like the exact graph, which I think would give us all these intermediate ID mappings in the natural way.

csrc/scheduler/multi_matmul.cpp
jacobhinkle: Removed here and added a uses().size() check in `addSetsForCacheReads`.

PR #2723, opened 2024-07-30
Mostly mechanical cleanup

naoyam:
Also added a disable option to disable contiguous indexing for testing

### DIFF
--- a/csrc/contiguity.cpp
+++ b/csrc/contiguity.cpp
@@ -48,8 +48,7 @@ OrderedIdInformation::OrderedIdInformation(
 
 bool OrderedIdInformation::checkExclusivelyConsumesAllocs(IterDomain* id) {
   NVF_ERROR(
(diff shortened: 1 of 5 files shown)


naoyam:
!build --diff

naoyam:
!build --diff

zasdfgbnm (approved):

naoyam:
The codegen diff results are unexpected. I suspect they are false alarms but will wait for @jacobhinkle's update.

jacobhinkle:
> The codegen diff results are unexpected. I suspect they are false alarms but will wait for @jacobhinkle's update.

I'm looking into this now with @xwang233 . Until we have a fix, you can run codediff locally by checkout out the branch (it must be clean) and running `tools/codediff/compare_codegen.sh`. With no arguments it will run all the tests.

PR #2725, opened 2024-07-30
Handle placeholders in AbstractTensor

zasdfgbnm:
```C++
// AbstractId in AbstractTensor can be place holders std::monostate{}. For
// example:
//
// Example 10:
//   AbstractTensor v({{}, {}}); // [null, null]
//   v.split(0, 2); // [null, null, null]
//   v.merge(0); // [null, null]
//   v.swizzle(SwizzleType::XOR, 0, 1); // [null, null]
```

### DIFF
--- a/csrc/abstract_tensor.cpp
+++ b/csrc/abstract_tensor.cpp
@@ -52,15 +52,22 @@ struct DispatchSplit {
 
 } // namespace
 
-void AbstractTensor::split(int64_t axis, Val* factor, bool inner_split) {
+AbstractTensor& AbstractTensor::split(
+    int64_t axis,
+    Val* factor,
+    bool inner_split) {
   axis = wrapDim(axis, (int64_t)domain.size());
   auto [outer, inner] =
       AbstractId::dispatch(DispatchSplit{}, domain[axis], factor, inner_split);
   std::swap(domain[axis], inner);
   domain.insert(domain.begin() + axis, outer);
+  return *this;
 }
 
-void AbstractTensor::split(int64_t axis, int64_t factor, bool inner_split) {
+AbstractTensor& AbstractTensor::split(
+    int64_t axis,
+    int64_t factor,
+    bool inner_split) {
   return split(
       axis, IrBuilder::create<Val>(factor, DataType::Index), inner_split);
 }
@@ -136,7 +143,7 @@ struct DispatchMerge {
 
 } // namespace
 
-void AbstractTensor::merge(int64_t axis_o, int64_t axis_i) {
+AbstractTensor& AbstractTensor::merge(int64_t axis_o, int64_t axis_i) {
   axis_o = wrapDim(axis_o, (int64_t)domain.size());
   axis_i = wrapDim(axis_i, (int64_t)domain.size());
 
@@ -149,9 +156,10 @@ void AbstractTensor::merge(int64_t axis_o, int64_t axis_i) {
 
   domain.erase(domain.begin() + domain_inner_pos);
   std::swap(domain[domain_outer_pos], output);
+  return *this;
 }
 
-void AbstractTensor::reorder(
+AbstractTensor& AbstractTensor::reorder(
     const std::unordered_map<int64_t, int64_t>& old2new) {
   NVF_ERROR(
       !domain.empty() || old2new.empty(), "Tried to reorder a 0-dim domain");
@@ -166,10 +174,12 @@ void AbstractTensor::reorder(
       [this](int64_t i) { return domain[i]; });
 
   domain = std::move(reordered_domain);
+  return *this;
 }
 
 // old2new[index] = permutation[index]
-void AbstractTensor::reorder(const std::vector<int64_t>& permutation) {
+AbstractTensor& AbstractTensor::reorder(
+    const std::vector<int64_t>& permutation) {
   std::unordered_map<int64_t, int64_t> reorder_map;
   int64_t idx = 0;
   std::transform(
@@ -180,7 +190,7 @@ void AbstractTensor::reorder(const std::vector<int64_t>& permutation) {
   return reorder(reorder_map);
 }
 
-void AbstractTensor::flatten(int64_t from, int64_t to) {
+AbstractTensor& AbstractTensor::flatten(int64_t from, int64_t to) {
   NVF_ERROR(!domain.empty(), "Tried to do flatten on a 0-dim domains");
   from = wrapDim(from, (int64_t)domain.size());
   to = wrapDim(to, (int64_t)domain.size());
@@ -190,6 +200,7 @@ void AbstractTensor::flatten(int64_t from, int64_t to) {
     (void)_;
     merge(from);
   }
+  return *this;
 }
 
 namespace {
@@ -370,7 +381,10 @@ struct DispatchLegacySwizzle {
 
 } // namespace
 
-void AbstractTensor::swizzle(SwizzleType swizzle_type, int64_t x, int64_t y) {
+AbstractTensor& AbstractTensor::swizzle(
+    SwizzleType swizzle_type,
+    int64_t x,
+    int64_t y) {
   x = wrapDim(x, (int64_t)domain.size());
   y = wrapDim(y, (int64_t)domain.size());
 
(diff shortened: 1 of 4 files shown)


zasdfgbnm:
!build

naoyam:
I'm not sure what this would enable. Could you give more context?

jacobhinkle (commented):
What is the usefulness of placeholders? Is there any promotion possible, e.g. what happens if we merge a placeholder with another kind of AbstractId?

zasdfgbnm:
@naoyam @jacobhinkle I believe this is mostly useful for scheduling reductions. For example, if I have a fusion:

```C++
T1[I1, I2] = set(T0);
T2[I3, r4] = sum(T1[I1, I2]);
T3[I5] = set(T2[I3, r4])
```
Then I can make my abstract tensor like this:

```
dim 0: I1, I3, I5
dim 1: I2, r4, null
```
so I can schedule all the tensors together.

zasdfgbnm:
!build

naoyam:
> @naoyam @jacobhinkle I believe this is mostly useful for scheduling reductions. For example, if I have a fusion:
> 
> ```c++
> T1[I1, I2] = set(T0);
> T2[I3, r4] = sum(T1[I1, I2]);
> T3[I5] = set(T2[I3, r4])
> ```
> 
> Then I can make my abstract tensor like this:
> 
> ```
> dim 0: I1, I3, I5
> dim 1: I2, r4, null
> ```
> 
> so I can schedule all the tensors together.

Hmm, I'm still lost. Could you give a full example?

zasdfgbnm:
> Hmm, I'm still lost. Could you give a full example?

Something like this?

```C++
TEST_F(Tutorial, Reduction) {
  Fusion fusion;
  FusionGuard fg(&fusion);

  TensorView* tv0 = makeSymbolicTensor(4);
  fusion.addInput(tv0);
  auto tv1 = set(tv0);
  auto tv2 = sum(tv1, {2, 3});
  auto tv3 = set(tv2);
  fusion.addOutput(tv3);

  AbstractTensor t(
      {{tv1->axis(0), tv2->axis(0), tv3->axis(0)},
       {tv1->axis(1), tv2->axis(1), tv3->axis(1)},
       {tv1->axis(2), tv2->axis(2), {}},
       {tv1->axis(3), tv2->axis(3), {}}});
  t.merge(0);
  t.merge(1);
  for (auto id : t[0].as<std::vector>()) {
    id->parallelize(ParallelType::BIDx);
  }
  for (auto id : t[1].as<std::vector>()) {
    if (id.hasValue()) {
      id->parallelize(ParallelType::TIDx);
    }
  }
  auto uz = t.unzip();
  tv1->setLoopDomain(uz[0].as<IterDomain*>());
  tv2->setLoopDomain(uz[1].as<IterDomain*>());
  tv3->setLoopDomain(uz[2].strip().as<IterDomain*>());

  at::Tensor t0 = at::randn({100, 100, 10, 10}, options);
  std::vector<c10::IValue> aten_inputs = {t0};

  FusionExecutor fe;
  fe.compileFusion(&fusion);
  std::vector<at::Tensor> outputs = fe.runFusion(aten_inputs);
  testValidate(&fusion, outputs, aten_inputs, __LINE__, __FILE__);
}
```

zasdfgbnm:
!build

naoyam:
> > Hmm, I'm still lost. Could you give a full example?
> 
> Something like this?
> 
> ```c++
>   AbstractTensor t(
>       {{tv1->axis(0), tv2->axis(0), tv3->axis(0)},
>        {tv1->axis(1), tv2->axis(1), tv3->axis(1)},
>        {tv1->axis(2), tv2->axis(2), {}},
>        {tv1->axis(3), tv2->axis(3), {}}});
> ```

Why not just:

```
   AbstractTensor t(
       {{tv1->axis(0), tv2->axis(0), tv3->axis(0)},
        {tv1->axis(1), tv2->axis(1), tv3->axis(1)},
        {tv1->axis(2), tv2->axis(2)},
        {tv1->axis(3), tv2->axis(3)}});
```

Does each dimension of an AbstractTensor need to have the same number of axes?

zasdfgbnm:
> Does each dimension of an AbstractTensor need to have the same number of axes?

Good question. Generally, no, but `unzip` does require dims to have equal number of axes.

Another motivation for adding this feature is, in the future, I expect there will be more and more utility functions to take an `AbstractTensor`, do some scheduling, and return the result `AbstractTensor`. And I feel that, in order to implement these utility functions, it would be convenient to assume the same structure of the input, and if some ID does not exist, the caller of the utility function could just put a placeholder. See: https://github.com/NVIDIA/Fuser/pull/2726.

zasdfgbnm:
> So this is something akin to best-effort replay. I thought that was the idea but I didn't see a case in `DispatchMerge` where only one of the arguments was `std::monostate` so I wasn't sure if that was an intended use case.

Because so far, the only use case that I can think of is to hold the place of a missing reduction, and merging a reduction with a non-reduction is not allowed, so here merging a null with a non-null is considered an error. In general, I also believe merging null and non-null is more likely to be a bug rather than an intention. So I would prefer to throw an error, rather than trying to accept it.

naoyam:
> but unzip does require dims to have equal number of axes.

Could you expand why?

zasdfgbnm:
> Could you expand why?

I think it is because, when the size mismatches, what the correct result should be can be confusing? For example, if you have

```
t = {{id0, id1, id2}, {id3}, {id4, id5, id6}}
```

then what should `t.unzip()` get? There are multiple options:

Option 1: trim to the shortest (similar to python's approach, where `zip` itself is used to do unzip)

```
{id0, id3, id4}
```

Option 2: leave the short ones empty

```
{id0, id3, id4}
{id1, id5}
{id2, id6}
```

Option 3: broadcast

```
{id0, id3, id4}
{id1, id3, id5}
{id2, id3, id6}
```

Option 4: use place holders

```
{id0, id3, id4}
{id1, null, id5}
{id2, null, id6}
```

I don't think there is one option clearly better than other options. And when I am facing such problem, I would prefer to just throwing an error and consider it as unsupported, and leave it to the user to use `null` to pad to the same sizes.

naoyam:
I think I'm not against having this as one mechanism, but I'm not that confident as I'm not yet sure what the scheduling should look like. 

One alternative approach might be that, once an `AbstractTensor` is scheduled (with no placeholder), it might make sense to design some interface to automatically propagate the schedule to each tensor without unzipping the abstract tensor. For example,

```
replay(t, tv3)
```

would propagate the transformation history starting from `tv3->axis(0), tv3->axis(1)`.

zasdfgbnm:
> but I'm not that confident as I'm not yet sure what the scheduling should look like.

Not confident about what? About whether this will be useful, whether this will be bug-prone, or whether we should use this as a canonical way of scheduling?

The target for this feature in my mind is the following senario: if I have a utility function `AbstractTensor scheduleXXX(AbstractTensor t)`, which is originally written to schedule the reduction tensor in a scheduler. But in a unit test that I am working on prototyping, I want use that for scheduling consumers of a reduction. Then I can just quickly insert some placeholders for the reduction axes and use that `scheduleXXX`.

This senario is pretty common in matmul, where we use `scheduleMmaOutputAllocation` for scheduling the result of mma, and use the same function for scheduling the output of mma (the gmem tensor) in unit tests for convenient;

Yes, replay is an alternative. We already have `scheduler_utils::BoundedDirectionalTransformPropagator`, which I believe is totally sufficient to do the work I want. The reason I dislike replay is, it has many parameters and configuration options, and I always have to spend time finding what these parameters are, and what is the value I want to fill in. On the other hand, when I am using this placeholder mechanism, likely I will be already familiar with this `scheduleXXX`, taking advantage of the placeholder mechanism takes less mental effort the studying again what the parameters are.

naoyam:
Summary of offline conversation:

These AbstractTensor PRs are mostly for quickly developing tests. So, they are meant to be used while fully aware of what's going to happen, rather than automating scheduling of complex fusions. Naoya is going to take a quick look again and approve them as they are just a utility interface. Naoya also mentioned to Xiang that these background context would have helped us more quickly review these PRs.

naoyam (approved):
LGTM

zasdfgbnm:
> Summary of offline conversation:
> 
> These AbstractTensor PRs are mostly for quickly developing tests. So, they are meant to be used while fully aware of what's going to happen, rather than automating scheduling of complex fusions. Naoya is going to take a quick look again and approve them as they are just a utility interface. Naoya also mentioned to Xiang that these background context would have helped us more quickly review these PRs.

Thanks for the summary.

PR #2727, opened 2024-07-31
Python API for `SdpaBwdOp`

Priya2698:
Python API for `SdpaBwdOp`. Changes include:
1. A new record type `SdpaBwdRecordType` and corresponding deserializer.
2. Setting default values for `dropout_p` and `is_causal`.
3. Tests for backward pass and complete SDPA fwd-bwd program for all supported parameter configs.

Issue #2278


Priya2698:
!build

Priya2698:
!build

rdspring1 (commented):
LGTM. I just commented on some nitpicks.

tests/python/test_python_frontend.py
```diff
@@ -4584,6 +4584,303 @@ def fusion_func(
...
+        v = torch.randn((N, H, S, E), dtype=torch.bfloat16, device="cuda:0")
+
+        dropout_vals = [None, 0.0, 0.2]
+        is_causal_vals = [None, True, False]
+        scale_vals = [None, 1 / E**0.5, 1e-3]
+
+        def fusion_func(
+            fd: FusionDefinition, has_dropout: bool, has_causal: bool, has_scale: bool
+        ):
+            grad_output = fd.define_tensor(
```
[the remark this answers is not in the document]
Priya2698: Yes, I thought this might be more clear to users in understanding what the different inputs look like -- given that two of them are CPU scalars as well.

csrc/ops/composite.cpp
```diff
@@ -584,6 +584,15 @@ SdpfaBwdResult sdpfa_bwd(
   NVF_CHECK(
       !scale || scale->isScalar(), "Expected scale to be a scalar double.");
 
+  // Set default values for dropout_p (0.0), is_causal(false)
+  if (dropout_p == nullptr) {
+    dropout_p = IrBuilder::create<Val>(0.0, DataType::Double);
+  }
+
+  if (is_causal == nullptr) {
+    is_causal = IrBuilder::create<Val>(false, DataType::Bool);
```
jjsjann123, line 593: nitpick we have zeroVal and falseVal for this.
https://github.com/NVIDIA/Fuser/blob/862dfd18a03b48617c4789b7507455eb996fb5d3/csrc/ir/container.h#L134

csrc/python_frontend/python_bindings.cpp
```diff
@@ -2911,6 +2911,70 @@ void initNvFuserPythonBindings(PyObject* module) {
...
+      },
+      py::arg("grad_output"),
+      py::arg("query"),
+      py::arg("key"),
+      py::arg("value"),
+      py::arg("output"),
+      py::arg("log_sumexp"),
+      py::arg("dropout_p").none(true) = py::none(),
+      py::arg("is_causal").none(true) = py::none(),
+      py::arg("philox_seed"),
```
jjsjann123, line 2973: naive question, we can have `philox_seed` here following something like `dropout_p` where it has a default value? can we actually skip `philox_seed` and what is given to it then?

jjsjann123 (commented):
:clap:

jjsjann123 (approved):
meant to stamp earlier.

csrc/python_frontend/python_bindings.cpp
Priya2698: Can you elaborate with an example of input set?

Philox seed is not optional, so that always needs to have a valid value. For `dropout_p`, we accept a `None` value, and in the C++ API, assign it the default value.

jjsjann123: > Philox seed is not optional

That's what I understand as well. I'm just a bit confused here in the pybind signature we defined for the function. Since `dropout_p` and `is_causal` is given a default value `py::none()`, but `philox_seed` is not. Does this make `philox_seed` a kw_only argument in the signature? But that's not what we used in tests, since these are given as positional args...

Out of curiosity. 
I'm wondering if it makes our lives easier to just remove the default arguments for `dropout_p/is_causal`.

Priya2698: Merging this PR as-is.

I agree with your point, so will change it in a separate PR.

PR #2728, opened 2024-07-31
Add `AbstractTensor::parallelize`

zasdfgbnm:


zasdfgbnm:
!build

csrc/abstract_tensor.cpp
```diff
@@ -413,4 +413,41 @@ std::vector<AbstractTensor> AbstractTensor::unzip() const {
...
+      // we are disabling it for now to avoid unexpected behavior. If in the
+      // future, it does turn out that we need to commonly manually construct a
+      // loop graph, we can turn this on.
+      NVF_ERROR(false, "Parallelizing ValGroup is not supported");
+    } else if constexpr (std::is_same_v<IN, std::vector<AbstractId>>) {
+      for (auto& aid : in) {
+        AbstractId::dispatch((*this), parallel_type, aid);
+      }
+    } else {
+      NVF_CHECK(false, "Unsupported type in AbstractTensor::split");
```
jacobhinkle, line 438:
```suggestion
      NVF_CHECK(false, "Unsupported type in AbstractTensor::parallelize");
```

csrc/abstract_tensor.cpp
```diff
@@ -413,4 +413,41 @@ std::vector<AbstractTensor> AbstractTensor::unzip() const {
...
+      return;
+    } else if constexpr (std::is_same_v<IN, IterDomain*>) {
+      return in->parallelize(parallel_type);
+    } else if constexpr (std::is_same_v<IN, ValGroupAndItsGraph>) {
+      // TODO: instead of throwing an error, should we just parallelize all IDs
+      // in the ValGroup? This behavior only makes sense for the loop graph, so
+      // we are disabling it for now to avoid unexpected behavior. If in the
+      // future, it does turn out that we need to commonly manually construct a
+      // loop graph, we can turn this on.
+      NVF_ERROR(false, "Parallelizing ValGroup is not supported");
```
jacobhinkle, line 432: Regarding the TODO: Since we might not have "replayed" all the tensors yet, I think it might make more sense to keep parallelization type with the `ValGroup` so that we can apply it to a tensor later after replaying.

jacobhinkle (approved):
LGTM. This is going to be great!

csrc/abstract_tensor.cpp
zasdfgbnm: What do you mean by "keep parallelization type with the `ValGroup`"? Something like keeping a `std::unordered_map<ValGroup, ParallelType>`?

csrc/abstract_tensor.cpp
```diff
@@ -413,4 +413,41 @@ std::vector<AbstractTensor> AbstractTensor::unzip() const {
...
+struct DispatchParallelize {
+  template <typename INPUT>
+  void operator()(ParallelType parallel_type, INPUT&& in) const {
+    using IN = std::decay_t<INPUT>;
+    if constexpr (std::is_same_v<IN, std::monostate>) {
+      return;
+    } else if constexpr (std::is_same_v<IN, IterDomain*>) {
+      return in->parallelize(parallel_type);
+    } else if constexpr (std::is_same_v<IN, ValGroupAndItsGraph>) {
+      // TODO: instead of throwing an error, should we just parallelize all IDs
```
naoyam, line 427: Would it make more sense to consider this as the responsibility of the user? I suppose that when using `AbstractTensor`, you'd be aware that each member of the abstract tensor was a ValGroup, so parallelization of the abstract tensor would result in parallelizing all IDs in the group seems to be an intuitive result.

zasdfgbnm: Sounds reasonable.

naoyam (approved):
LGTM

csrc/abstract_tensor.cpp
jacobhinkle: Yeah, something like that, or `ValGroupWithItsGraphAndParallelType` :smile: .

PR #2729, opened 2024-07-31
Add `AbstractTensor::{zip,addRow}`

zasdfgbnm:


zasdfgbnm:
!build

zasdfgbnm:
!build

csrc/abstract_tensor.h
```diff
@@ -127,6 +127,25 @@ using AbstractId = dynamic_type::DynamicType<
...
+//
+// Besides, you can stack another AbstractTensor into the current
+// AbstractTensor. For example:
+//
+// Example 11:
+//   IterDomain *id0, *id1, *id2, *id3, *id4, *id5;
+//   AbstractTensor v0({{id0, id1}, {id2, id3}});
+//   AbstractTensor v1({id4, id5});
+//   v0.stack(v1);
+// Then v0 will be [{id0, id1, id4}, {id2, id3, id5}].
```
jacobhinkle, line 148: Is `zip` just a special case of `stack` where all the input `AbstractId`s are non-vectors? I wonder if we could just implement a general stack that handles both vector and non-vector `AbstractId`s in each position and then alias it as `zip` for ergonomics.

zasdfgbnm: Good question. I think the primary concern I have is, "handles both vector and non-vector" is not very well defined. For example, if I have two abstract tensors:
```
t0 = {id0, {id1, id2}}
t1 = {id3, id4}
```
then what should the stack of `t0` with `t1` defined? Should the result be

```
{{id0, id3}, {id1, id2, id4}}
```

or 

```
{{id0, id3}, {{id1, id2}, id4}}
```

csrc/abstract_tensor.h
```diff
@@ -207,6 +226,17 @@ struct AbstractTensor {
   // AbstractTensor is [dim0={id0, id1}, dim1={id2, id3}], then the return value
   // will be {AbstractTensor{id0, id2}, AbstractTensor{id1, id3}}.
   std::vector<AbstractTensor> unzip() const;
+
+  // Zip multiple AbstractTensors into a single AbstractTensor. For example, if
+  // the input is {AbstractTensor{id0, id2}, AbstractTensor{id1, id3}}, then the
+  // return value will be [dim0={id0, id1}, dim1={id2, id3}].
+  static AbstractTensor zip(std::vector<AbstractTensor> tensors);
+
+  // Stack another AbstractTensor into the current AbstractTensor. For example,
```
naoyam, line 235: This definition of stack sounds like concatenation?

naoyam (approved):
LGTM. I only left a comment on naming.

csrc/abstract_tensor.h
zasdfgbnm: It is neither like stack nor like concat... 

Stack is like below:
```
{id0, id1}
{id2, id3}
->
{{id0, id2}, {id1, id3}}
```
so it has the same meaning as `zip`.

Concat is like below:
```
{{id0}, {id1}}
{{id2}, {id3}}
->
{{id0, id2}, {id1, id3}}
```

But what this function is doing is:
```
{{id0}, {id1}}
{id2, id3}
->
{{id0, id2}, {id1, id3}}
```

Probably I should just rename it as `addRow`.

zasdfgbnm: renamed

PR #2730, opened 2024-07-31
Allow `TensorView::broadcast` to specify extent, and make `BroadcastFromNowhere` more thorough and fix bugs

zasdfgbnm:


csrc/ir/interface_nodes.h
```diff
@@ -283,7 +283,8 @@ class NVF_API TensorView : public Val {
       ComputeAtMode mode = ComputeAtMode::Standard);
 
   //! Create a new broadcast IterDomain with extent one in the loop domain
```
zasdfgbnm, line 285:
```suggestion
  //! Create a new broadcast IterDomain with the given extent in the loop domain
```

csrc/ir/internal_base_nodes.h
```diff
@@ -599,7 +599,7 @@ class TensorDomain : public Val {
   int64_t rootPosOf(IterDomain* id) const;
 
   //! Create a new broadcast IterDomain with extent one in the loop domain
```
zasdfgbnm, line 601:
```suggestion
  //! Create a new broadcast IterDomain with the givenextent in the loop domain
```

tests/cpp/test_gpu3.cpp
```diff
@@ -8171,95 +8171,171 @@ TEST_F(NVFuserTest, BroadcastFromNowhere) {
...
+      tv0->domain()->additionalIDs());
+  ir_utils::validateDomainEquivalence(
+      tv0->getLogicalDomain(),
+      tv0->getLoopDomain(),
+      tv0->domain()->additionalIDs());
 
-  tv0->broadcast(2);
+  tv0->broadcast(2, 2); // [b, I, b]
   EXPECT_TRUE(tv0->axis(2)->isBroadcast());
+  EXPECT_EQ(tv0->axis(2)->extent()->value(), 2);
```
jacobhinkle, line 8205: Is there a reason this option should affect `extent()` and not `expandedExtent()`?

zasdfgbnm: There is no reason. I don't think we have a clear definition on when to use extent and when to use expanded extent. I don't see any reason why expand is not implemented as just using extent, instead of create a new field called "expanded extent".

So far, expanded extent attaches to `ExpandOp`, and if you do `split(b1, 5)`, you will get a broadcasting ID whose extent is 5. For this case, because it is a scheduling activity, I feel that it looks more like the split case.

naoyam:
When do we need to use an non-size-one extent?

zasdfgbnm:
> When do we need to use an non-size-one extent?

The loop domain will have a size 128 broadcasting:
image

naoyam (approved):
I'm still unclear why a size matters for a broadcast domain, but I assume this is how mma is going to be scheduled. I'd like to revisit later, but for now LGTM.

zasdfgbnm:
> I'm still unclear why a size matters for a broadcast domain, but I assume this is how mma is going to be scheduled. I'd like to revisit later, but for now LGTM.

We can definitely revisit this. The current design is, this newly created broadcast does nothing more than being a label saying that "this expr needs 128 threads". It is newly created, and never further transformed, so it is just ignored by most analysis.

tests/cpp/test_gpu3.cpp
jacobhinkle: > I don't see any reason why expand is not implemented as just using extent, instead of create a new field called "expanded extent".

Yeah it's a common point of confusion. I don't see any reason we couldn't get rid of `expandedExtent`. It's still useful to have `hasExpandedExtent()` but I guess that could be implemented as `!extent()->sameAs(fusion()->oneVal())`.

PR #2732, opened 2024-07-31
Split getReproErrorString in two

riccardofelluga:
By splitting this helper function devs can get the repro by using `getReproString`.

### DIFF
--- a/nvfuser/__init__.py
+++ b/nvfuser/__init__.py
@@ -64,14 +64,10 @@ class FusionDefinition(_C._FusionDefinition):
             logger.exception(self.getReproErrorString("defining"))
             raise
 
-    def getReproErrorString(self, section: str, inputs: list | None = None):
+    def getReproString(self, inputs: list | None = None) -> str:
         msg = (
-            f"An error occurred while {section} nvFuser FusionDefinition {self.id()}.\n"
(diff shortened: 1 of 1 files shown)


jacobhinkle (approved):
Good idea. Thanks for adding this.

nvfuser/__init__.py
```diff
@@ -64,14 +64,10 @@ def __exit__(self, type, value, traceback):
...
-    def getReproErrorString(self, section: str, inputs: list | None = None):
+    def getReproString(self, inputs: list | None = None) -> str:
         msg = (
-            f"An error occurred while {section} nvFuser FusionDefinition {self.id()}.\n"
-            "If you believe this is a bug or need assistance, please file an issue at "
-            "https://github.com/NVIDIA/Fuser/issues/new\n"
-            f"Here's a script to reproduce the error:\n"
-            "```python\n"
-            "# CUDA devices:\n"
+        "```python\n"
```
jacobhinkle, line 69: Would it make more sense to keep this line and the closing markdown line in `getReproErrorString` so that `getReproString` returns valid python code?

riccardofelluga: Ah true! that's a good idea!

PR #2734, opened 2024-07-31
newValLike tries to propagate sharding. 

### LINKED ISSUE #2721: MarkAliasPrepare does not preserve shardings
Error from `MarkAliasPrepare` pass (https://github.com/NVIDIA/Fuser/pull/2639) with multidevice execution:

Repro test:
```
TEST_F(MultiDeviceTest, Repro) {
  FusionGuard fg(fusion.get());
  auto mesh = DeviceMesh::createForNumDevices(communicator_->size());

  TensorView* x = makeContigTensor(2);
  TensorView* w = makeContigTensor(3);
  fusion->addInput(x);
  fusion->addInput(w);

  TensorView* w_t = transpose(w, 1, 2);
  TensorView* mm = matmul(x, w_t);
  TensorView* mm_t = transpose(mm, 1, 2);
  fusion->addOutput(mm_t);

  for (auto tv : {x}) {
    tv->setDeviceMesh(mesh);
  }
  for (auto tv : {w, w_t, mm, mm_t}) {
    tv->setDeviceMesh(mesh);
  }

  const auto options =
      at::TensorOptions().device(communicator_->device());
  auto x_ = at::randn({12, 48}, options);
  auto w0_ = at::randn({1, 36, 48}, options);

  hir::HostIrExecutorParams executor_params{
      .use_fusion_executor_cache = true,
      .skip_auto_scheduling = false,
      .cache_fusion_executor = false};
  MultiDeviceExecutor runtime(
      std::move(fusion), *communicator_, executor_params);
  runtime.runWithInput({x_, w0_});
}
```

The error comes from csrc/multidevice/lower_communication.cpp:lowerToScatter, where we have an erroneous scatter generated from a TV without a DeviceMesh, which should not happen.
```
Lower to scatter T4_g[ ideviceIdx.x12{i3}, iS14{i4}, iS13{i0} ] (DeviceMesh{0 1})
   = Set.Permute( T5_g[ iS20{i3}, iS21{i0}, iS22{i4} ], cache_op=Streaming )
```
Ultimately, an error is generated when trying to look up a device in a non-existent DeviceMesh
Error from segmentation group 3: vector::_M_range_check: __n (which is 0) >= this->size() (which is 0)

The error looks like it comes from the introduction of the segment_set after before the transpose which doesn't preserve the shardings.

host_ir dump:
```
%kernel {
T2_l[ ideviceIdx.x5{i3}, iS7{i5}, iS6{i4} ] (DeviceMesh{0 1})
   = Set.Permute( T1_g[ ideviceIdx.x2{i3}, iS3{i4}, iS4{i5} ] (DeviceMesh{0 1}), cache_op=Streaming )
T3_l[ ideviceIdx.x8{i3}, iS9{i0}, iS10{i4}, rS11{i2} ] (DeviceMesh{0 1})
   = matmul(T0_g[ iS0{i0}, iS1{i2} ] (DeviceMesh{0 1}),
            T2_l[ ideviceIdx.x5{i3}, iS7{i5}, iS6{i4} ] (DeviceMesh{0 1}))
**T4_g[ ideviceIdx.x12{i3}, iS14{i4}, iS13{i0} ] (DeviceMesh{0 1})
   = Set.Permute( T3_l[ ideviceIdx.x8{i3}, iS9{i0}, iS10{i4}, rS11{i2} ] (DeviceMesh{0 1}), cache_op=Streaming )**
} // %kernel
```

fusion_ir_preseg dump:
```
%kernel_math {
T2_l[ ideviceIdx.x5{i3}, iS24{i2}, iS6{i4} ] (DeviceMesh{0 1})
   = Set.Permute( T1_g[ ideviceIdx.x2{i3}, iS3{i4}, iS23{i2} ] (DeviceMesh{0 1}), cache_op=Streaming )
T3_l[ ideviceIdx.x8{i3}, iS9{i0}, iS10{i4}, rS11{i2} ] (DeviceMesh{0 1})
   = matmul(T0_g[ iS0{i0}, iS1{i2} ] (DeviceMesh{0 1}),
            T2_l[ ideviceIdx.x5{i3}, iS24{i2}, iS6{i4} ] (DeviceMesh{0 1}))
**T5_l[ iS20{i3}, iS21{i0}, iS22{i4} ]
   = SegmenterSet( T3_l[ ideviceIdx.x8{i3}, iS9{i0}, iS10{i4}, rS11{i2} ] (DeviceMesh{0 1}) )
T4_g[ ideviceIdx.x12{i3}, iS14{i4}, iS13{i0} ] (DeviceMesh{0 1})
   = Set.Permute( T5_l[ iS20{i3}, iS21{i0}, iS22{i4} ], cache_op=Streaming )**
} // %kernel_math 
```

wujingyue:
Fixes #2721. 

#2721 uncovers a deeper problem: the existing preseg passes are not aware of shardings, and therefore generate TensorViews that have no meshes or have wrong DIDs, breaking communication lowering. 

The current flow is:
1. The user annotates meshes and DIDs to some TensorViews. 
2. MultiDeviceExecutor runs PropagateShardingsPass to ensure all TensorViews have a mesh and propagate DIDs to more tensors than annotated. 
3. FusionExecutorCache runs the preseg passes. MarkAliasesPrepare is the one in question, but the problem can happen to any preseg pass. 

This PR implements one solution, although I'm all ears for alternatives. It extends `newValLike` so most `csrc/ops` APIs will attempt to propagate shardings from inputs to outputs. This minimizes the changes to the passes. However, the order of creating ops and setting meshes/paralleltypes will affect the semantics. For example, in
```cpp
TensorView* in = ...;
in->setDeviceMesh(...);
in->axis(...)->parallelize(DID);
TensorView* out = set(in);
``` 
`in` and `out` will have the same sharding. 

However, in
```cpp
TensorView* in = ...;
TensorView* out = set(in);
in->setDeviceMesh(...);
in->axis(...)->parallelize(DID);
```
`out` won't be sharded at all. 

The latter pattern is expected to be used in the Thunder integration -- build the ops before annotate shardings. The former pattern is expected to be used in internal transformation passes who intend to propagate the input sharding to the newly added op.

### DIFF
--- a/csrc/ops/utils.cpp
+++ b/csrc/ops/utils.cpp
@@ -415,20 +415,13 @@ TensorView* newOutputTV(const std::vector<Val*>& vals, DataType dtype) {
       dtype);
 
   DeviceMesh new_mesh;
+  // Find the first input that has a mesh. This seems arbitrary, but is at this
+  // moment safest because it's consistent with PropagateShardingsPass.
   for (auto* tv : ir_utils::filterByType<TensorView>(vals)) {
     const DeviceMesh& mesh = tv->getDeviceMesh();
-    if (mesh.empty()) {
-      continue;
-    }
-    if (new_mesh.empty()) {
+    if (!mesh.empty()) {
       new_mesh = mesh;
-    } else {
-      NVF_ERROR(
-          new_mesh == mesh,
-          "Inconsistent device mesh among inputs: ",
-          new_mesh,
-          " vs ",
-          mesh);
+      break;
     }
   }
   new_out->setDeviceMesh(new_mesh);
--- a/csrc/ops/utils.cpp
+++ b/csrc/ops/utils.cpp
@@ -418,9 +418,8 @@ TensorView* newOutputTV(const std::vector<Val*>& vals, DataType dtype) {
   // Find the first input that has a mesh. This seems arbitrary, but is at this
   // moment safest because it's consistent with PropagateShardingsPass.
   for (auto* tv : ir_utils::filterByType<TensorView>(vals)) {
-    const DeviceMesh& mesh = tv->getDeviceMesh();
-    if (!mesh.empty()) {
-      new_mesh = mesh;
+    if (tv->hasDeviceMesh()) {
+      new_mesh = tv->getDeviceMesh();
       break;
     }
   }
--- a/csrc/ops/utils.cpp
+++ b/csrc/ops/utils.cpp
@@ -316,6 +316,11 @@ IterDomain* newOutputIterDomain(
       continue;
     }
 
+    NVF_ERROR(
+        id->getParallelType() == ParallelType::Serial ||
+            isParallelTypeDeviceDim(id->getParallelType()),
+        id->getParallelType(),
+        " is not expected when building ops.");
     parallel_type = promoteParallelType(parallel_type, id->getParallelType());
 
     if (extent_is_from_symbolic && !id->isSymbolic()) {
--- a/csrc/ops/utils.cpp
+++ b/csrc/ops/utils.cpp
@@ -5,12 +5,13 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 // clang-format on
-#include <ir/builder.h>
-#include <ops/arith.h>
-#include <ops/utils.h>
-
 #include <algorithm>
 #include <limits>
+
+#include <ir/builder.h>
+#include <ir/utils.h>
+#include <ops/arith.h>
+#include <ops/utils.h>
 
 namespace nvfuser {
 namespace ops {
@@ -263,6 +264,22 @@ std::vector<IterDomain*> mapLinearOpIterDomains(
   return mapping;
 }
 
+namespace {
+ParallelType promoteParallelType(ParallelType a, ParallelType b) {
+  if (a == b) {
+    return a;
+  }
+  NVF_ERROR(
+      a == ParallelType::Serial || b == ParallelType::Serial,
+      "Doesn't know how to resolve ",
+      a,
+      " and ",
+      b,
+      " at this moment.");
+  return a == ParallelType::Serial ? b : a;
+}
+} // namespace
+
 // Adding these pragmas since gcc-12.2.1
 // incorrectly reports a warning with the use of evaluate
 #if defined(__GNUC__) && !defined(__clang__)
@@ -282,19 +299,15 @@ IterDomain* newOutputIterDomain(
   Val* extent_val = nullptr;
   bool extent_is_from_symbolic = true;
   Val* expanded_extent_val = nullptr;
+  auto parallel_type = ParallelType::Serial;
   std::optional<IterType> iter_type = std::nullopt;
 
-  std::vector<IterDomain*> ids;
-  ids.reserve(input_ids.size());
+  for (auto id : input_ids) {
+    // Filter out any nullptrs
+    if (id == nullptr) {
+      continue;
+    }
 
-  // Filter out any nullptrs
-  std::copy_if(
-      input_ids.begin(),
-      input_ids.end(),
-      std::back_inserter(ids),
-      [](IterDomain* id) { return id != nullptr; });
-
-  for (auto id : ids) {
     if (id->isBroadcast()) {
       if (id->hasExpandedExtent()) {
         expanded_extent_val =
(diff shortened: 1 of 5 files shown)


wujingyue:
!build --pybench

wujingyue:
!build

wujingyue:
!build

wujingyue:
This fails `ReshardingTest.Detection` as expected because the ops APIs start to aggressively propagate shardings. Before I go ahead fixing the tests, I'd love to hear your thoughts on the general direction.

csrc/ops/utils.cpp
```diff
@@ -282,26 +299,25 @@ IterDomain* newOutputIterDomain(
...
-  for (auto id : ids) {
     if (id->isBroadcast()) {
       if (id->hasExpandedExtent()) {
         expanded_extent_val =
             promoteSize(expanded_extent_val, id->expandedExtent());
       }
       continue;
     }
+
+    parallel_type = promoteParallelType(parallel_type, id->getParallelType());
```
naoyam, line 319: I wonder if we should assert that the parallel type of `id` should be either `Serial` or `DID`, just in case.

naoyam (approved):
LGTM

csrc/ops/utils.cpp
```diff
@@ -391,10 +409,24 @@ std::vector<IterDomain*> newOutputDomain(const std::vector<Val*>& vals) {
 
 TensorView* newOutputTV(const std::vector<Val*>& vals, DataType dtype) {
   auto out_domain = newOutputDomain(vals);
-  return IrBuilder::create<TensorView>(
+  auto* new_out = IrBuilder::create<TensorView>(
       IrBuilder::create<TensorDomain>(
           out_domain, TensorDomain::getContiguityFilledWith(out_domain, true)),
       dtype);
+
+  DeviceMesh new_mesh;
```
cowanmeg, line 417: I wonder if we can get into an issue where we introduced a resharding expression (which would later get an explicit `set` operation by insertResharding preseg pass), but it's probably dependent on where newOutputDomain is called.

cowanmeg (commented):
Do you know which test cases are failing in ReshardingTest.Detection? I wonder if some of the logically redundant reshardings are getting optimized away.

wujingyue:
> Do you know which test cases are failing in ReshardingTest.Detection? I wonder if some of the logically redundant reshardings are getting optimized away.

I'll get back to that tomorrow. Two main reasons are:
1. With this PR, `add` attempts to propagate shardings and doesn't know how to proceed when seeing different device meshes on the two inputs. 
2. With this PR, `set` propagates DID and therefore parallelizes the output TV. Without, `set` doesn't propagate DID and therefore leaves the output replicated.

csrc/ops/utils.cpp
wujingyue: Yes, with this PR, `set` will try to propagate the input sharding to the output. I expect this to be worked around by
1. Not use `set`, and instead create the TensorDomain/TensorView by hand and then the op. 
2. Use `set` but remember to overwrite the DIDs. 

Overall, I don't think this is a perfect solution, and I'm all ears for alternatives.

wujingyue:
!build

tests/cpp/test_resharding.cpp
```diff
@@ -82,97 +82,98 @@ TEST_F(ReshardingTest, Detection) {
 
   TensorView* tv0 = makeContigTensor(3);
   fusion->addInput(tv0);
+  TensorView* tv1 = set(tv0);
+  TensorView* tv2 = set(tv1); // resharding
```
samnordmann, line 86: Why moving all those lines at the top? The problem is that the comment "resharding"/"not resharding" are based on the device mesh, which is now set far for the line

cowanmeg: I think it might be the ordering issue Jingyue brought up in the meeting (short example in the meeting notes)

csrc/ops/utils.cpp
cowanmeg: Yeah, I think we won't solve this until sharding propagation is revamped into something more intelligent. I think this is a good temp solution until we come up with something else.

tests/cpp/test_resharding.cpp
wujingyue: Yes, I mentioned that in the description as well. But Sam's concern is still valid that the comments and the shardings will be too far from each other. I'm not quite happy with that either. 

I can make some changes to the test: 
1. still shard immediately after the op (and therefore the comment), but the annotation will be based on the already propagated shardings. E.g. [reset](https://github.com/NVIDIA/Fuser/pull/2744/files#diff-c5175dc597d3e4dd05ea38630935938b69c33328284a9e3b689a60ccaaef5b7bR204-R205) a parallel type to `Serial` for all gather. 
2. Split the test so we have only one op per test. This is more typing but follows the best testing practice to keep tests independent.

wujingyue:
!build

tests/cpp/test_resharding.cpp
wujingyue: I merged the PR with ReshardingTest.Detection split into one op per test.

PR #2735, opened 2024-07-31
Fix random seed in multidevice mlp test

cowanmeg:
Only use at::manual_seed(getATenRandomSeed()) to control random seed in reference and nvFuser programs.

### DIFF
--- a/tests/cpp/test_multidevice_transformer.cpp
+++ b/tests/cpp/test_multidevice_transformer.cpp
@@ -81,17 +81,11 @@ namespace {
 TensorView* replicated_dropout(
     TensorView* x,
     const double kProb,
-    DeviceMesh mesh,
-    int64_t seed_offset=0) {
-  // Need to modify two things before we can use the existing dropout function
-  // in composite.cpp (1) Sharding propagation breaks at rand_like because it
-  // creates a fresh TV. (2) The philox seed and offset must be set to ensure
-  // the masks are identical across processes.
+    DeviceMesh mesh) {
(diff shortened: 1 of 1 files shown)


cowanmeg:
!build

wujingyue (commented):
This is a strict improvement, so LGTM. However, according to https://github.com/NVIDIA/Fuser/pull/2676#discussion_r1697697414 and the discussion around it, isn't the right fix to fork the state so we don't need to reset the state? IIUC, `at::manual_seed(...)` resets the state regardless of which seed to use.

cowanmeg:
I think fork is also changing the state since it returns the offset back to where it began. Here we reset the seed to the default to force the offset to 0. We could read the offset before the reference and set the offset back.

wujingyue:
> I think fork is also changing the state since it returns the offset back to where it began. 

Yes, there's still a difference between resetting the state to the beginning of the scope and resetting the state all the way to the beginning of the program. I'll need to think more about it -- I've yet to understand whether that's just a difference or has larger implication.

cowanmeg:
For actual training, I would say resetting to 0 would introduce correlation so there is larger implications.

wujingyue (approved):

wujingyue:
This is good enough for now so LGTM! 

fork_rng is worth considering. It resets the state for a smaller scope. It starts to make a difference when we run several fusions in a test and might not want their RNGs to correlate. 

Currently, fork_rng doesn't exist in the C++ API, so it needs extra plumbing to be able to use it.

PR #2737, opened 2024-07-31
Remove reduction axis when `K=1`

Priya2698:
When K=1, we will be mapping a reduction axis to broadcast axis in `LinearOp/MatmulOp`. This is not allowed in nvFuser. This was first noticed in Issue #2532 but exists generally.

PR #2534 is included in this PR for checking `LinearOp`.

### DIFF
--- a/csrc/executor.cpp
+++ b/csrc/executor.cpp
@@ -1861,6 +1861,7 @@ std::vector<at::Tensor> FusionExecutor::evaluateFusionOutputs(
     for (const auto& out_val : fusion()->outputs()) {
       auto out_tensor =
           expr_eval.evaluate(out_val->as<TensorView>()).as<at::Tensor>();
+      expr_eval.bind(out_val, out_tensor);
       outputs.emplace_back(out_tensor);
     }
   }
--- a/csrc/logical_domain_map.cpp
+++ b/csrc/logical_domain_map.cpp
@@ -213,7 +213,7 @@ std::unordered_map<IterDomain*, IterDomain*> PairwiseLogicalDomainMap::map(
     } else {
       NVF_ERROR(false, "Producer did not match any LinearOp input.")
     }
-    
+
     bool k_bcast = op->inA()->as<TensorView>()->axis(-1)->isBroadcast();
     // LinearOp:
     // inputs (0) = {*, in_features}
@@ -222,7 +222,8 @@ std::unordered_map<IterDomain*, IterDomain*> PairwiseLogicalDomainMap::map(
     // output = {*, out_features} / {*}
 
     const std::vector<IterDomain*>& aligned_producer_ids =
-        ops::mapLinearOpIterDomains(producer_logical, input_position, out_size, k_bcast);
+        ops::mapLinearOpIterDomains(
+            producer_logical, input_position, out_size, k_bcast);
     pairwiseMapAllIds(aligned_producer_ids, consumer_root);
     return dom_map;
   }
--- a/csrc/logical_domain_map.cpp
+++ b/csrc/logical_domain_map.cpp
@@ -213,7 +213,8 @@ std::unordered_map<IterDomain*, IterDomain*> PairwiseLogicalDomainMap::map(
     } else {
       NVF_ERROR(false, "Producer did not match any LinearOp input.")
     }
-
+    
+    bool k_bcast = op->inA()->as<TensorView>()->axis(-1)->isBroadcast();
     // LinearOp:
     // inputs (0) = {*, in_features}
     // weight (1) = {out_features, in_features} / {in_features}
@@ -221,7 +222,7 @@ std::unordered_map<IterDomain*, IterDomain*> PairwiseLogicalDomainMap::map(
     // output = {*, out_features} / {*}
 
     const std::vector<IterDomain*>& aligned_producer_ids =
-        ops::mapLinearOpIterDomains(producer_logical, input_position, out_size);
+        ops::mapLinearOpIterDomains(producer_logical, input_position, out_size, k_bcast);
     pairwiseMapAllIds(aligned_producer_ids, consumer_root);
     return dom_map;
   }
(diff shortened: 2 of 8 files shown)


Priya2698:
!build

csrc/executor.cpp
```diff
@@ -1861,6 +1861,7 @@ std::vector<at::Tensor> FusionExecutor::evaluateFusionOutputs(
     for (const auto& out_val : fusion()->outputs()) {
       auto out_tensor =
           expr_eval.evaluate(out_val->as<TensorView>()).as<at::Tensor>();
+      expr_eval.bind(out_val, out_tensor);
```
jacobhinkle, line 1864: Doesn't this happen automatically when you call  `expr_eval.evaluate`?

csrc/logical_domain_map.cpp
```diff
@@ -214,14 +214,16 @@ std::unordered_map<IterDomain*, IterDomain*> PairwiseLogicalDomainMap::map(
...
     // LinearOp:
     // inputs (0) = {*, in_features}
     // weight (1) = {out_features, in_features} / {in_features}
     // bias (2) = {out_features} / {}
     // output = {*, out_features} / {*}
 
     const std::vector<IterDomain*>& aligned_producer_ids =
-        ops::mapLinearOpIterDomains(producer_logical, input_position, out_size);
+        ops::mapLinearOpIterDomains(
+            producer_logical, input_position, out_size, k_bcast);
```
jacobhinkle, line 226: Is something similar also needed for `MatmulOp` and `Sdpa*Op`?

csrc/scheduler/mma_utils.cpp
```diff
@@ -1810,13 +1810,14 @@ DimRolesMap MatmulPattern::getDimRoles(IdModel& id_model) const {
 
   } else if (output->definition()->isA<LinearOp>()) {
     const std::vector<IterDomain*>& out_logical = output->getLogicalDomain();
+    bool k_bcast = A->getLogicalDomain().back()->isBroadcast();
```
jacobhinkle, line 1813: This is good for now, but in the future I think we might want to generalize this pattern matching so that this doesn't even look like a matmul pattern at this point. I am thinking we could have a general stage where we find patterns, possibly in the unsegmented fusion, and those can be of various flavors, like pointwise, matmul, etc. Each of those pattern matchers can then do its own replacements inside the `schedule*` functions (or before). That way we could have a fusion that had a broadcast K like this _inside_ a fusion that has a _real_ matmul and still be able to schedule it as a single-matmul fusion; the pointwise pattern would translate to a broadcasted mul and the matmul pattern would be translated to an `MmaOp`. Anyway, as I said this is probably fine for now but I'm somewhat sure we'll currently have trouble generating a kernel for broadcast K so I'll probably need to check for this condition in the matmul scheduler's `canScheduleCompileTime`.

jacobhinkle (commented):
Makes sense to me. It looks like there are some relevant CI failures.

csrc/logical_domain_map.cpp
```diff
@@ -214,14 +214,16 @@ std::unordered_map<IterDomain*, IterDomain*> PairwiseLogicalDomainMap::map(
       NVF_ERROR(false, "Producer did not match any LinearOp input.")
     }
 
+    bool k_bcast = op->inA()->as<TensorView>()->axis(-1)->isBroadcast();
```
jjsjann123, line 217: naive question, do we not allow broadcast k between `inA` and `inB`? Hence here we only need to check for one but not the other.

jacobhinkle: We should not allow that. I'm not sure if it's checked yet in the op.

csrc/executor.cpp
Priya2698: No, binding does not happen automatically. `validateValWithConcrete` is only triggered when `bind` is called. (Discussion: https://github.com/NVIDIA/Fuser/pull/2534#discussion_r1667013063).

I'll close out your PR #2534 after this is merged.

csrc/scheduler/mma_utils.cpp
Priya2698: Do you want to open an issue to keep track of this case for later?

csrc/ops/composite.cpp
```diff
@@ -64,32 +64,40 @@ static TensorView* newForLinear(
   auto input_domain = TensorDomain::noReductions(input->getLogicalDomain());
   auto weight_domain = TensorDomain::noReductions(weight->getLogicalDomain());
 
+  // Output has a reduction axis rK if K is not bcast
```
jacobhinkle, line 67: I think we could insert a check that `input_domain.back()->isBroadcast() == weight_domain.back()->isBroadcast()` (similar for the other ops). h/t @jjsjann123.

csrc/logical_domain_map.cpp
Priya2698: I'll add a check, thanks for catching it @jjsjann123. I erroneously recalled already having that check.

Priya2698:
CI errors in SDPA will be resolved after https://github.com/NVIDIA/Fuser/pull/2743.

csrc/scheduler/mma_utils.cpp
jacobhinkle: #2748

Priya2698:
!build

csrc/executor.cpp
jacobhinkle: Oh right. Haha I was forgetting my own PR. After you merge this PR, we can merge #2534 and remove this line as part of it.

jacobhinkle (approved):

csrc/executor.cpp
Priya2698: I dont't understand. Why do we need to remove this line?

jacobhinkle: Oh, nevermind. This _is_ the line from #2534. You're right; sorry about the confusion. Yes we can go ahead and close out #2534.

Issue #2673, opened 2024-07-24
`cudaMalloc` when using `at::matmul_out`


samnordmann:
# Description
Nsight profiles reveal that using `matmul_out` introduces allocations of new buffers. Our conclusion is:
- `matmul_out` triggers allocation whereas it shouldn't, at least when using multiple streams
- This allocation might be hidden by torch's allocator cache, which doesn't work well with streams (see, e.g., [here](https://dev-discuss.pytorch.org/t/fsdp-cudacachingallocator-an-outsider-newb-perspective/1486))

Even if the allocator cache can save us in some places, I see no reason why `matmul_out` should allocate.

Also, relying on the allocator is risky in my opinion, especially because we are beyond what is supported by the cache, especially when we use different slices on the same buffer in different streams, as done in the overlap technic of #2401. 

# Experiment

We provide a reproducer in 
- #2682 

that performs `I=8` matrix multiply in a for-loop. The parameters of our experiments are:
- the matrix dimensions `M=2^8`, `N=2^8`, `K=2^8`
- setup: H100 80GB. We also ran the same experiment on V100 16GB and got similar result, so we omit this case.
- `compute_mode`:
    - `matmul`: use `at::matmul`
    - `matmul_out`: use `at::matmul_out` on preallocated output
    - `unfused`: decompose the matmul into pointiwise multiply `at::mul_out` and sum reduction `at::sum_out` on preallocated buffers
- `stream_mode`
    - `no_streams`: use only the default streams
    - `post_on_different_streams`: posts the `at::` compute operation on different streams, but allocate the buffers on the default stream. **This is probably what is closest to our use case** from #2401 where we allocate one big buffer and handle each slice on a different stream
    - `allocate_and_post_on_different_streams`: allocates and posts the `at::` compute operations on different streams
- `enable_allocator_cache`: through setting or unsetting the environment variable `PYTORCH_NO_CUDA_MEMORY_CACHING` 

# 1. `stream_mode=no_streams`
## 1.a. `compute_mode==matmul_out`
Whether we enable allocation cache or not, we do not see any cudaMalloc:
![LOG2_M=8__LOG2_N=8__LOG2_K=8__STREAM_MODE=no_streams__COMPUTE_MODE=matmul_out__ENABLE_MEMORY_CACHE=0](https://github.com/user-attachments/assets/15698310-86b3-443a-a82b-eb9b8fbcd0d2)

## 1.b. `compute_mode==matmul`

[...]

PR #2738, opened 2024-08-01
Host irs/overlap: add `number_of_streams` and `number_of_iterations`

samnordmann:
# What

This patch adds features and improve several aspects the reduce-scatter-based pipelining overlap experiments. We patch both the ATen and the host IR implementation.

### adds a parameter`number_of_streams`: 
For each slice of the tensor, we change the stream in a round-robin fashion. Before this patch, we used as many streams as there were slices (aka `params.S`, the extent of the outermost dimension). We observe in #2673 (see in particular [this message](https://github.com/NVIDIA/Fuser/issues/2673#issuecomment-2260235042)) that we want to:
   - have at least one warm-up iteration
   - keep the number of streams relatively low:
       - `at::matmul_out` (and, in general, pytorch allocator cache) allocates buffers per stream.
       - We only need a couple of streams to achieve overlap. I think using two or three streams should be sufficient (**TODO**: check what is done in Megatron)

Doing so, **we eliminate the undesired `cudaMalloc`** described [here](https://github.com/NVIDIA/Fuser/pull/2584#issuecomment-2238833771) **and achieve overlap between comms and compute**. See screenshots below.

### adds a parameter `number_of_iterations`
For repeating the whole process of instantiating the inputs with new random values and posting the compute/comms. We validate data corectness only on the last iteration. The goal is to have one or more warmup iterations before analyzing the profile and perf.

### Improve memory footprint:
- `tc_locally_reduced_` outermost axis is chosen to be smaller than the number of streams
- unsharded tensors `ta_unsharded_`, `tb_unsharded_` are large tensors only used for validation, so we allocate them on the CPU memory.

# TODO 
- use several PG for comms/comms overlap.
- Run at bigger scale, on H100
- Lookup what params are used in TE

# Profile analysis

### Settings:
8*A100 40g, luna node

Params:
- `M=N=K=2^10`

Other parameters as in the file.
- `S=8`
- `number_of_streams=3`
- `number_of_iterations=4`

## ATen implementation
`OverlapTest.ReduceScatterBasedPipeliningATenImplementation`

In this first zoomed-out screenshot, we can see four "large blocks" separated by the green `cudaStreamSynchronize`, corresponding to the parameter`number_of_iteration=4`. We observe that the first `ncclDevKernel_ReduceScatter` is longer than the other -- probably due to a reallocation of the input at each iteration (which we tried to avoid though...).
Screenshot 2024-08-02 at 13 22 52

In the second screenshot, we zoom inside one iteration. We highlight the `Kernel2`(`cutlass::Kernel2<cutlass_80_tensorop_s1688gemm_64x64_16x6_nn_align4>`) which corresponds to the `matmul_out` operation. The three first matmul are executed before and during the first reduce_scatter, corresponding to the parameter `number_of_streams=3`.
Screenshot 2024-08-02 at 13 25 20

If we zoom further on the right part, we clearly see that over is achieved between comms and compute.
Screenshot 2024-08-02 at 13 27 18

## Host IR implementation

The same behavior is observed in the Host IR implementation `OverlapTest.ReduceScatterBasedPipeliningHostIrImplementation`:

Screenshot 2024-08-02 at 13 28 16

### DIFF
--- a/tests/cpp/test_multidevice_overlap.cpp
+++ b/tests/cpp/test_multidevice_overlap.cpp
@@ -94,7 +94,7 @@ class OverlapTest : public MultiDeviceTest {
         num_devices_, params.K / num_devices_, params.N};
     std::vector<int64_t> tb_sizes = {params.K / num_devices_, params.N};
     std::vector<int64_t> tc_locally_reduced_sizes = {
-        params.S, params.M / params.S, params.N};
+        std::min(params.S, params.number_of_streams), params.M / params.S, params.N}; // we need at most `number_of_streams` slices
     std::vector<int64_t> tc_sizes = {
         params.S, params.M / (params.S * num_devices_), params.N};
 
@@ -218,12 +218,14 @@ TEST_F(OverlapTest, ReduceScatterBasedPipeliningATenImplementation) {
     allocateIO();
 
     for (auto j : c10::irange(params.S)) {
+      int64_t stream_index = j % streams.size();
+      setCurrentCUDAStream(streams.at(stream_index));
+
       // define the sliced tensors
       auto ta_j = ta_.select(0, j);
-      auto tc_locally_reduced_j = tc_locally_reduced_.select(0, j);
+      auto tc_locally_reduced_j = tc_locally_reduced_.select(0, stream_index);
       auto tc_j = tc_.select(0, j);
 
-      setCurrentCUDAStream(streams.at(j % streams.size()));
 
       // local compute
       torch::matmul_out(tc_locally_reduced_j, ta_j, tb_);
--- a/tests/cpp/test_multidevice_overlap.cpp
+++ b/tests/cpp/test_multidevice_overlap.cpp
@@ -20,9 +20,9 @@ namespace nvfuser {
 
 struct OverlapTestParams {
   // Tensors sizes
-  int64_t LOG2_M = (std::getenv("LOG2_M"))? std::atoi(std::getenv("LOG2_M")) : 10;
-  int64_t LOG2_K = (std::getenv("LOG2_K"))? std::atoi(std::getenv("LOG2_K")) : 10;
-  int64_t LOG2_N = (std::getenv("LOG2_N"))? std::atoi(std::getenv("LOG2_N")) : 10;
+  int64_t LOG2_M = (std::getenv("NVFUSER_OVERLAP_LOG2_M"))? std::atoi(std::getenv("NVFUSER_OVERLAP_LOG2_M")) : 10;
+  int64_t LOG2_K = (std::getenv("NVFUSER_OVERLAP_LOG2_K"))? std::atoi(std::getenv("NVFUSER_OVERLAP_LOG2_K")) : 10;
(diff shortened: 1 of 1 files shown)


samnordmann:
!build

samnordmann:
!build

xwang233:
ICYMI, the thunder failure is about `thunder.tests.test_nvfuser_remat.test_find_cut_dropout_nvfuser_cuda_None`.

I've skipped the `test_nvfuser_cse` in thunder, which was abount an environment issue for bitsandbytes.

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -196,36 +211,35 @@ class OverlapTest : public MultiDeviceTest {
...
-    auto tc_j = tc_.select(0, j);
-
-    if (params.use_different_streams) {
-      auto new_stream = c10::cuda::getStreamFromPool(
-          /*isHighPriority=*/false, my_device_index_);
-      streams.push_back(new_stream);
-      setCurrentCUDAStream(new_stream);
+  std::generate_n(
+      std::back_inserter(streams),
+      params.number_of_streams,
```
wujingyue, line 216: AFAICS from https://github.com/NVIDIA/Fuser/blob/346e51c285f2d511cc2a9455d46dd0c97f5787d2/tests/cpp/test_multidevice_overlap.cpp#L52, `params` is fixed to the default value. Therefore, the test never explores other values unless someone manually changes the source code? 

(Not a big deal; I'm just verifying my understanding because my first impression was "why not TEST_P")

samnordmann: > the test never explores other values unless someone manually changes the source code?

That is correct, for now

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -91,49 +101,54 @@ class OverlapTest : public MultiDeviceTest {
...
-    auto tb_unsharded = at::randn(tb_unsharded_sizes, options);
-    ta_ = at::empty(ta_sizes, options);
-    ta_.copy_(ta_unsharded.select(1, my_device_index_));
-    tb_ = at::empty(tb_sizes, options);
-    tb_.copy_(tb_unsharded.select(0, my_device_index_));
+    auto cpu_options = at::TensorOptions().dtype(at::kFloat);
+    at::TensorOptions gpu_options = cpu_options.device(communicator_->device());
+
+    ta_unsharded_ = at::empty(ta_unsharded_sizes, cpu_options);
+    tb_unsharded_ = at::empty(tb_unsharded_sizes, cpu_options);
```
wujingyue, line 108: What was wrong with using gpu_options? It's not immediately clear to me, so some clarification in code comment would help.

wujingyue (approved):

tests/cpp/test_multidevice_overlap.cpp
samnordmann: The reason is what I wrote in the PR description, under the section "Improve memory footprint": those tensors are large and only used for validating data correctness, so we shouldn't use device memory for them. Let me add a comment in the code.

wujingyue:
!build

Issue #1758, opened 2024-02-13
Add a validation pass


tfogal:
We should have a validation pass to ensure that the input program makes sense. The following invalid program tries to add Float and BFloat16 tensors together, which is invalid because our semantics say the user must cast first:

```python
from nvfuser import FusionDefinition, DataType

def nvfuser_fusion_id1(fd : FusionDefinition) -> None :
                T0 = fd.define_tensor(shape=[-1,-1], contiguity=[True,True], dtype=DataType.Float, is_cpu=False, stride_order=[1,0])
                T1 = fd.define_tensor(shape=[-1,-1], contiguity=[True,True], dtype=DataType.BFloat16, is_cpu=False, stride_order=[1,0])
                T2 = fd.ops.add(T0, T1)
                fd.add_output(T2)

with FusionDefinition() as fd:
    nvfuser_fusion_id1(fd)

import torch
inputs = [
    torch.randn((4,), dtype=torch.float32, device='cuda:0').as_strided((2, 2), (2, 1)),
    torch.randn((4,), dtype=torch.float32, device='cuda:0').as_strided((2, 2), (2, 1)),
]   
fd.execute(inputs)
```

When executed, the process dies with:

```
RuntimeError: Expected T1_g[ iS2{i3}, iS3{i4} ] to be bound to a tensor of dtype __bfloat, but got a tensor of dtype float
```

We should fail more gracefully:
* we should not tear down the whole process when we fail; report an error back up instead
*  the user can't do anything with references like `iS2{i3}`; error messages more at the level of the program are what they would want

Incomplete list of what we should consider in a validation pass:
- [ ] Check operand types
- [ ] Ensure support matches the GPU targeted: we can't create bf16 kernels on volta hardware
- [ ] Ensure support matches the GPU targeted: we can't create fp8 kernels on ampere hardware

naoyam:
I think we automatically insert a necessary cast operation. The error in this case is about the `T1` input being `float32`. 

For the fp type support, that should be relatively trivial. 

More generally, though, it's a hard problem to report a meaning error message in Python from a C++ error. Is anybody familiar with general design guidelines? For example, should raising more specific exceptions help? What should be the protocol of error reporting look like between C++ and Python? Currently, the C++ side just uses `NVF_ERROR`, which throws an `nvfuser::nvfError` exception. Maybe we should have some more subclasses of this exception, like `nvfSchedulingError`, `nvfLoweringError`?

[...]

PR #2739, opened 2024-08-01
Improve error message with mismatched valtype/dtype

### LINKED ISSUE #2670: Expected ... to be bound to an at::Tensor, but got l
```
  File "/home/tfogal/env/lib/python3.10/site-packages/nvfuser/__init__.py", line 145, in execute
    result = self._execute(
RuntimeError: Expected T0_g[ iS0{i1}, iS1{i2}, iS2{i3}, iS3{i4} ] to be bound to an at::Tensor, but got l
Exception raised from validateValWithConcreteValue at /tmp/pip-req-build-fk_mwug5/csrc/expr_evaluator.cpp:29 (most recent call first):
frame #0: nvfuser::nvfCheckFail(char const*, char const*, unsigned int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) + 0xf3 (0x74a9a5f92f9b in /home/tfogal/env/lib/python3.10/site-packages/nvfuser/_C.cpython-310-x86_64-linux-gnu.so)
frame #1: nvfuser::ExpressionEvaluator::bind_(nvfuser::Val const*, dynamic_type::DynamicType<dynamic_type::Containers<std::vector>, nvfuser::StructHandle, nvfuser::Pointer, nvfuser::Opaque, at::Tensor, std::complex<double>, double, long, bool>, bool) + 0x188a (0x74a9a632786a in /home/tfogal/env/lib/python3.10/site-packages/nvfuser/_C.cpython-310-x86_64-linux-gnu.so)
frame #2: nvfuser::executor_utils::bindInputs(nvfuser::KernelArgumentHolder const&, nvfuser::Fusion*) + 0x287 (0x74a9a6316da7 in /home/tfogal/env/lib/python3.10/site-packages/nvfuser/_C.cpython-310-x86_64-linux-gnu.so)
frame #3: <unknown function> + 0x69dee4 (0x74a9a6503ee4 in /home/tfogal/env/lib/python3.10/site-packages/nvfuser/_C.cpython-310-x86_64-linux-gnu.so)
frame #4: nvfuser::FusionExecutorCache::runFusionWithInputs(c10::ArrayRef<c10::IValue> const&, std::optional<nvfuser::PrimDataType>, std::optional<signed char>) + 0x1e7 (0x74a9a6505827 in /home/tfogal/env/lib/python3.10/site-packages/nvfuser/_C.cpython-310-x86_64-linux-gnu.so)
```

Simplified reproducer:
```python
import torch

def a(FusionDefinition):
    b.define_tensor([])
with FusionDefinition() as b:
    a(b)
c = [0]
b.execute(c)
```

[The original, larger program that failed](https://github.com/user-attachments/files/16365137/failure.py.txt)

jacobhinkle:
We already check types when binding arguments, but now we explicitly state which arguments mismatch, and in which positions.

Fixes #2670 by improving the error message. Before this change the error message was
```
Expected T0_g[ iS0{i1} ] to be bound to an at::Tensor, but got l
```
After this PR, the error is
```
Expected input 0, T0_g[ iS0{i0} ], to be an at::Tensor but got scalar 2
```

### DIFF
--- a/csrc/expr_evaluator.cpp
+++ b/csrc/expr_evaluator.cpp
@@ -14,13 +14,39 @@
 #include <ir/iostream.h>
 #include <ir/utils.h>
 #include <logical_domain_map.h>
+#include <polymorphic_value.h>
 
+#include <cxxabi.h>
 #include <functional>
 #include <iostream>
 
 namespace nvfuser {
 
 namespace {
+
+// Given a value, if it is not a fusion input, return the empty string. If it is
+// a fusion input return a string like "input 2 ". This helper is used to
+// provide more informative error messages when a malformed input is received.
+std::string getInputPosString(const Val* val) {
+  if (!val->isFusionInput()) {
+    return "";
+  }
+  // Get position
+  const std::vector<Val*>& inputs = val->fusion()->inputs();
+  int64_t pos = -1;
+  for (size_t i : c10::irange(inputs.size())) {
+    if (inputs[i] == val) {
+      pos = (int64_t)i;
+      break;
+    }
+  }
+  NVF_ERROR(
+      pos != -1,
+      "val->isFusionInput() is true but val cannot be found in fusion inputs: ",
+      val->toString());
+  std::stringstream ss;
+  return "input " + std::to_string(pos) + ", ";
+}
 
 void validateValWithConcreteValue(
     const Val* value,
@@ -29,17 +55,19 @@ void validateValWithConcreteValue(
     NVF_CHECK(
         concrete_value.is<at::Tensor>(),
         "Expected ",
+        getInputPosString(tv),
         tv->toString(),
-        " to be bound to an at::Tensor, but got ",
-        concrete_value.type().name());
+        ", to be an at::Tensor but got scalar ",
+        concrete_value);
     const auto& t = concrete_value.as<at::Tensor>();
     auto expect_dim =
         (int64_t)TensorDomain::noReductions(tv->getLogicalDomain()).size();
     NVF_CHECK(
         t.dim() == expect_dim,
         "Expected ",
+        getInputPosString(tv),
         tv->toString(),
-        " to be bound to a tensor of rank ",
+        ", to be bound to a tensor of rank ",
         expect_dim,
         ", but got a tensor of rank ",
         t.dim());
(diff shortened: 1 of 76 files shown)


jacobhinkle:
!build

tfogal (approved):
That error message is a great improvement! Thank you.

jacobhinkle:
This is not an ideal fix since it replicates (poorly) the validation that already exists in `ExpressionEvaluator::bind`. I did it this way to be able to provide the argument number, which we lose once we get to the `bind` call. I'll try again to just improve the error message locally within `bind` and omit the argument number for now. In the future, it might be nice to be able to declare error contexts with guards so that additional info is printed. For example in `bindInputs` we could have:
```c++
for (const auto i : c10::irange(inputs.size())) {
  ErrorContext e("Error binding input " + std::to_string(i));
  expr_eval.bind(inputs[i], *args[i], true);
}
```
Then if `NVF_CHECK` or `NVF_ERROR` fails inside the lifetime of `e`, we would get an error message like `Error binding input 2: <actual error>`. That's a more general problem that we should handle separately (I'll post an issue for it). For now I will just try and clean up the `bind` error messages and revert the current changes.

jacobhinkle:
I fixed this up by determining the input position inside the `bind_` call when a check fails. I also added some cases to the test to check that our error messages work properly for other kinds of malformed inputs.

csrc/expr_evaluator.cpp
```diff
@@ -14,32 +14,60 @@
 #include <ir/iostream.h>
 #include <ir/utils.h>
 #include <logical_domain_map.h>
+#include <polymorphic_value.h>
 
+#include <cxxabi.h>
```
tfogal, line 19: Is this header file used / needed? maybe it was initially there for debugging?

tfogal (approved):
Re-+1'ing the new changes.

But I also want to plug #1758: I'd really encourage an up-front pass to validate as soon as we can. As we found with latency work (#2136), ExprEvaluator is on the critical path, and so I worry about adding more checks in there.

Admittedly--these checks *have* to be on the critical path (since we can only validate when we see what inputs we actually receive). But it would _probably_ behoove us to do an up-front check for UX purposes and avoid checks inside each Expr evaluation.

PR #2740, opened 2024-08-01
[codediff] Gracefully handle starting from detached HEAD

jacobhinkle:
Currently if you are on a detached HEAD (i.e. if you have run `git checkout <SHA>` then `compare_codegen.sh` will fail with `fatal: ref HEAD is not a symbolic ref`. This PR makes it safe to run in such a scenario.

NOTE: I am sometimes seeing new git branches introduced at cleanup named like the SHA of the commit and the compare commit. This has happened previously and I think it is unrelated to this change. Since it only affects cleanup, I would like to merge this PR so that we can improve `!build --diff` in the CI and look into the branch issue later.

### DIFF
--- a/tools/codediff/compare_codegen.sh
+++ b/tools/codediff/compare_codegen.sh
@@ -107,7 +107,8 @@ fi
 # save current commit and current head so we can switch back to branch
 currentcommit=$(git describe --always --long)
 origcommit=$currentcommit
-orighead=$(git symbolic-ref --short HEAD)
+# When head is detached, symbolic-ref will fail. Then we just use the commit SHA.
+orighead=$(git symbolic-ref --short HEAD || git rev-parse HEAD)
 
 comparecommit=$(git describe --always --long "$comparetoref")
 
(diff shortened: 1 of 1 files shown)


tools/codediff/compare_codegen.sh
```diff
@@ -121,7 +122,7 @@ scriptdir=$(mktemp -d -t codediffXXXXXX)
 cp -r "$nvfuserdir/tools/codediff/"* "$scriptdir/"
 
 movecudafiles() {
-    find . -maxdepth 1 -name '__tmp_kernel*.cu' -o -name '__tmp_kernel*.ptx' -exec mv '{}' "$1" \;
+    find . -maxdepth 1 \( -name '__tmp_kernel*.cu' -o -name '__tmp_kernel*.ptx' \) -exec mv '{}' "$1" \;
```
jacobhinkle, line 125: Unrelated fix for a bug found by `shellcheck`.

xwang233 (approved):
LGTM. Thanks for the fix.

Issue #2741, opened 2024-08-01
Performance of innerOuterPersistent scheduler is very sensitive to register usage and spills


liqiangxl:
Noticed a performance regression of layer norm backward from July 19 to July 26 (no CI data from 22, 23, 24, 25) on H100 with hidden size around 15K to 16K. See [dashboard](http://nv/eh3). SOL droped from 65% to 45%

Further check found regression happened after #2628, the code change is minor and only slightly changed register spills

Before, SOL = 70%, register spills
```
64 bytes stack frame, 64 bytes spill stores, 128 bytes spill loads
```
After, SOL = 45%, register spills
```
64 bytes stack frame, 60 bytes spill stores, 208 bytes spill loads
```

liqiangxl:
attached code from main branch and main branch without  promote hoisting of vectorize predicates.
[high_perf.txt](https://github.com/user-attachments/files/16461323/high_perf.txt)
[low_perf.txt](https://github.com/user-attachments/files/16461324/low_perf.txt)

liqiangxl:
Generated with:
```
NVFUSER_PROF=print NVFUSER_DUMP=cuda_to_file,ptxas_verbose,scheduler_params,launch_param pytest /opt/pytorch/nvfuser/benchmarks/python/test_layernorm_bwd.py -k "float16 and 16384_16384" -v -s  --disable-benchmarking 2>&1 |tee 1.log
```

liqiangxl:
The possible reason of regression may because of compiler optimization from cuda to ptx code.
The slow version has 2500 more lines of PTX code.
**In the fast version** , #2742 ,`b47` is used in both `if` and `else` branches. Compiler can optimize out the `else` branch, a very short PTX code generated for else branch.

![image](https://github.com/user-attachments/assets/eed22c04-2d4d-4eba-a68c-ee85f6e3da42)


**In the slow version  (current main)**, `b47` is only used in the `else` branch.  The compiler generates PTX code for the else branch. That’s why we see 2500 more lines of PTX code.

![image](https://github.com/user-attachments/assets/24f89218-14e5-4fd0-94b9-2e1e2d0ead7d)

**PTX compare:**
![image](https://github.com/user-attachments/assets/846021dc-4b63-4870-b0ef-2e2df6478bcd)


[slow.ptx.txt](https://github.com/user-attachments/files/16564730/slow.ptx.txt)
[...]

PR #2742, opened 2024-08-01
Revert "Promote hoisting of vectorize predicates"

liqiangxl:
check issue #2741 

Regression is recovered after revert, see [dashboard](http://nv/eh4).


naoyam:
I thank both you and Xiao for finding the issue, but I don't think just reverting the PR is the right fix. Let's first create an issue.

naoyam:
Sorry I missed there's already an issue.

naoyam:
!build

naoyam (approved):

PR #2743, opened 2024-08-01
Remove `query/key_seq_len, debug_attn_mask` in `SdpaFwdOp`

Priya2698:
1. Since, we have deviated from torch API, making the following changes to have a leaner API:

-   Removes `debug_attn_mask` which is a 1D zero-tensor since `return_debug_mask=false` in `SdpaFwdOp`.
-   Removes `query_seq_len, key_seq_len`: These variables can be computed in the backward pass for non-nested tensors.

2. Adds `DIDx` axis to `log_sumexp` in `SdpaFwdOp::evaluate` and `logical_domain_mapping`. The `DIDx` axis is assumed when the tensorview is created but was not added everywhere. This removes discrepancy in different parts of the code. This will be revisited after work on sharded `SdpaBwdOp`. The `DIDx` can be removed from `log_sumexp` if not required.

### DIFF
--- a/csrc/ir/internal_nodes.h
+++ b/csrc/ir/internal_nodes.h
@@ -2237,11 +2237,8 @@ class SdpaFwdOp : public Expr {
       IrBuilderPasskey,
       TensorView* output,
       TensorView* log_sumexp,
-      TensorView* query_seq_len,
-      TensorView* key_seq_len,
       TensorView* philox_seed,
       TensorView* philox_offset,
-      TensorView* debug_attn_mask,
       Val* query,
       Val* key,
       Val* value,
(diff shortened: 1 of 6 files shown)


Priya2698:
!build

jjsjann123 (commented):
API changes looks mechanical to me.

> Adds DIDx axis to log_sumexp in SdpaFwdOp::evaluate and logical_domain_mapping... This will be revisited after work on sharded SdpaBwdOp

This change looks a bit suspicious to me, but I don't have a good understanding on that topic. I'll let @cowanmeg review that part and stamp.

cowanmeg:
> Adds DIDx axis to log_sumexp in SdpaFwdOp::evaluate

This is correct! If H is sharded on the inputs, it we expect a sharded output _and_ log_sumexp. 
The change probably looks suspicious since it's a temporary hack :P until we don't need to split logical dimensions for DID parallelization.

cowanmeg (approved):

jacobhinkle (approved):
LGTM

PR #2749, opened 2024-08-02
Check captured python definition in pytest framework.

rdspring1:
The string representation of a `FusionDefinition` is an executable function. Previously, this feature was only checked in `test_python_frontend.py`. This PR adds that check to the pytest framework.

Details:
- Created `utils.py` to hold helper functions
- Create `check_captured_python_definition` test helper function to enforce that the string representation runs the same as the original `FusionDefinition`
- Move cuda device checks to `utils.py`

### DIFF
--- a/tests/python/utils.py
+++ b/tests/python/utils.py
@@ -6,7 +6,9 @@
 import torch
 from torch.testing._internal.common_utils import TEST_WITH_ROCM
 from torch.testing._internal.jit_utils import RUN_CUDA
-from nvfuser import FusionDefinition
+
+# flake8: noqa
+from nvfuser import FusionDefinition, DataType
 
 RUN_NVFUSER = RUN_CUDA and not TEST_WITH_ROCM
 
--- a/tests/python/utils.py
+++ b/tests/python/utils.py
@@ -4,7 +4,29 @@
 # Owner(s): ["module: nvfuser"]
 
 import torch
-from nvfuser import (FusionDefinition, DataType)
+from nvfuser import FusionDefinition, DataType
+
+
+def is_pre_volta():
+    if not RUN_NVFUSER:
+        return False
+    prop = torch.cuda.get_device_properties(torch.cuda.current_device())
+    return prop.major < 7
+
+
+def is_pre_ampere():
+    if not RUN_NVFUSER:
+        return False
+    prop = torch.cuda.get_device_properties(torch.cuda.current_device())
+    return prop.major < 8
+
+
+def is_pre_hopper():
+    if not RUN_NVFUSER:
+        return False
+    prop = torch.cuda.get_device_properties(torch.cuda.current_device())
+    return prop.major < 9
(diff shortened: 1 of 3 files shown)


rdspring1:
!build

tests/python/utils.py
```diff
@@ -0,0 +1,67 @@
...
+# All rights reserved.
+# SPDX-License-Identifier: BSD-3-Clause
+# Owner(s): ["module: nvfuser"]
+
+import torch
+from torch.testing._internal.common_utils import TEST_WITH_ROCM
+from torch.testing._internal.jit_utils import RUN_CUDA
+
+# flake8: noqa
+from nvfuser import FusionDefinition, DataType
```
rdspring1, line 11: `flake8` complains about `DataType` being unused in this file but is necessary to run captured fusion definition.

tests/python/utils.py
```diff
@@ -0,0 +1,67 @@
...
+
+RUN_NVFUSER = RUN_CUDA and not TEST_WITH_ROCM
+
+
+def is_pre_volta():
+    if not RUN_NVFUSER:
+        return False
```
Priya2698, line 18: Should this be inside these functions? We have a separate `unittest.skip` decorator for this variable

tests/python/utils.py
```diff
@@ -0,0 +1,67 @@
...
+    prop = torch.cuda.get_device_properties(torch.cuda.current_device())
+    return prop.major < 7
+
+
+def is_pre_ampere():
+    if not RUN_NVFUSER:
+        return False
```
Priya2698, line 25: Same as above.

tests/python/utils.py
rdspring1: I'm not sure if that is true for the pytest framework.

tests/python/utils.py
```diff
@@ -0,0 +1,67 @@
...
+# Owner(s): ["module: nvfuser"]
+
+import torch
+from torch.testing._internal.common_utils import TEST_WITH_ROCM
+from torch.testing._internal.jit_utils import RUN_CUDA
+
+# flake8: noqa
+from nvfuser import FusionDefinition, DataType
+
+RUN_NVFUSER = RUN_CUDA and not TEST_WITH_ROCM
```
jjsjann123, line 13: do we still need these? I'm pretty sure these are coming from the torchscript tests in the old days.

tests/python/utils.py
jjsjann123: let's leave this as a comment int he code.

jjsjann123 (approved):
I think we can remove RUN_CUDA and TEST_WITH_ROCM these days.

Stamping since the overall change looks pretty safe.

Priya2698 (approved):
LGTM. Thanks for expanding this functionality.

rdspring1:
!build

PR #2701, opened 2024-07-27
Refactor `test_python_frontend.py`

Priya2698:
`test_python_frontend.py` is now 4k+ lines which is tedious to work with.

This PR:
1. Moves `exec_nvfuser` into a separate `class NVFuserTest (TestCase)` which can be used as the base class of any nvfuser test class that needs to verify serialization/deserialization.
2. Removes `setUpModule` into `setup_class` of `NVFuserTest` to avoid adding it in every file. This method will run `atexit_serde_check` to enable default serialization at the end of the test suite.
~3. As a starting point, moving matmul and linear tests into a separate file.~ To be done in separate PR.
4. Replaces `unittest` with `pytest` for consistency.

PR #2753, opened 2024-08-02
Add serialization check to the pytest framework.

rdspring1:
This PR adds a serialization check to the pytest opinfo framework.

Details:
- The pytest framework `serde_check` only checks per-operator and per-dtype. It doesn't check every single sample input. We check more input variations in the pytest benchmarks than `test_python_frontend.py`. This is a timesaving measure.
- Creates `atexit_serde_create_op_test` pseudo decorator to test automatic serialization. Serializing error test cases embeds errors in the serialized binary. Normally, we call `FusionCache.reset()` to clear the cache after running an error test. In the pytest framework, the error tests are separate from the correctness tests. We use this decorator to apply this to the correctness tests.
- The pytest framework and `test_python_frontend.py` use different arguments. `basic_serde_check` is the common part between them. It serializes the cache, deletes it, and then deserialized to recreate the cache.

### DIFF
--- a/tests/python/pytest_ops.py
+++ b/tests/python/pytest_ops.py
@@ -10,7 +10,7 @@ from copy import deepcopy
 
 from benchmarks.python.core import clear_cuda_cache
 from pytest_fusion_definitions import default_fd_fn, parse_inputs_fusion_definition
-from pytest_framework import create_op_test
+from pytest_framework import create_op_test, atexit_serde_create_op_test
 from pytest_core import ReferenceType, OpInfo, SampleInput
 from pytest_opinfos import opinfos
 from pytest_utils import ArgumentType, is_tensor, requiresJAX
@@ -199,7 +199,9 @@ def serde_test_fn(op: OpInfo, dtype: torch.dtype):
             return result
 
 
-@create_op_test(tuple(op for op in opinfos if op.sample_input_generator is not None))
+@atexit_serde_create_op_test(
+    tuple(op for op in opinfos if op.sample_input_generator is not None)
+)
 def test_correctness(op: OpInfo, dtype: torch.dtype):
     return serde_test_fn(op, dtype)
 
--- a/tests/python/pytest_ops.py
+++ b/tests/python/pytest_ops.py
@@ -18,7 +18,33 @@ from typing import Callable
 
 from nvfuser import FusionCache, FusionDefinition
 
-from utils import check_captured_python_definition
+from utils import check_captured_python_definition, debug_serde, basic_serde_check
+
+
+def serde_check(test_fn: Callable):
+    """
+    A decorator to verify that serialization works with the given exec_nvfuser
+    function. Currently, it uses serialization to rebuild the FusionCache
+    structure.
+    """
+
+    def inner_fn(*args, **kwargs):
+        # NOTE: For debug purposes, clear FusionCache before running first test
+        # so the behavior is more deterministic (PR #1848).
+        if debug_serde:
+            FusionCache.reset()
+            assert FusionCache.get().num_fusions() == 0
+
+        # Populate FusionCache
+        test_fn(*args, **kwargs)
+
+        # Serialize and Deserialize FusionCache
+        basic_serde_check()
+
+        # Run test with repopulated FusionCache
+        return test_fn(*args, **kwargs)
+
+    return inner_fn
 
 
 def parse_args_fusion_execution(opinfo: OpInfo, *args):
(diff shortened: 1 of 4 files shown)


rdspring1:
!build

Priya2698:
I am unclear on why we have different flavors of the `serde` decorators. Currently, we have:
- `serde_check`
- `atexit_serde_create_optest`
- `basic_serde_check`,
Can you update the PR description with some details on the design choices regarding this?

rdspring1:
For `serde_check` and `basic_serde_check`:
The pytest framework and test_python_frontend.py use different arguments for testing, so we need specific `serde_check` decorators for both frameworks. `basic_serde_check` is the common part between them. It serializes the cache, deletes it, and then deserialized to recreate the cache.

For `atexit_serde_create_optest`:
This pseudo-decorator enables automatic serialization upon program exit and tests deserializing the default workspace upon creating the tests. Serializing error test cases corrupts the serialized binary. We call `FusionCache.reset()` to clear the cache after running an error test in `test_python_frontend.py'. In the pytest framework, the error tests are separate from the correctness tests. We only apply this decorator to the correctness tests to avoid calling FusionCache.reset().

jjsjann123 (approved):
Seems to be a small code change enabling serde stuff in pytest, while the rest is just mechanical changes.
Thanks for the comment in code. :bow:

Priya2698:
> For `serde_check` and `basic_serde_check`:
> 
> * The pytest framework and test_python_frontend.py use different arguments for testing, so we need specific `serde_check` decorators for both frameworks. `basic_serde_check` is the common part between them. It serializes the cache, deletes it, and then deserialized to recreate the cache.

Can we add this in a comment to the code? 

> 
> For `atexit_serde_create_optest`:
> 
> * This pseudo-decorator enables automatic serialization upon program exit and tests deserializing the default workspace upon creating the tests. Serializing error test cases corrupts the serialized binary. We call `FusionCache.reset()` to clear the cache after running an error test in `test_python_frontend.py'. In the pytest framework, the error tests are separate from the correctness tests. We only apply this decorator to the correctness tests to avoid calling FusionCache.reset().

Can this replace `setUpModule`? The functionality seems to be identical.

tests/python/pytest_ops.py
```diff
@@ -10,15 +10,42 @@
...
 from pytest_utils import ArgumentType, is_tensor, requiresJAX
 from typing import Callable
 
 from nvfuser import FusionCache, FusionDefinition
 
-from utils import check_captured_python_definition
+from utils import check_captured_python_definition, debug_serde, basic_serde_check
+
+
+def serde_check(test_fn: Callable):
```
Priya2698, line 24: Can we rename this, maybe `serde_check_ops`.

In PR #2701, I will split the `test_python_frontend.py` file, so it would be easier in avoiding naming conflicts

rdspring1:
> For serde_check and basic_serde_check:
> 
> The pytest framework and test_python_frontend.py use different arguments for testing, so we need specific serde_check decorators for both frameworks. basic_serde_check is the common part between them. It serializes the cache, deletes it, and then deserialized to recreate the cache.
> Can we add this in a comment to the code?

It is already added for `basic_serde_check` in tests/python/utils.py.

> For atexit_serde_create_optest:
> 
> This pseudo-decorator enables automatic serialization upon program exit and tests deserializing the default workspace upon creating the tests. Serializing error test cases corrupts the serialized binary. We call FusionCache.reset() to clear the cache after running an error test in `test_python_frontend.py'. In the pytest framework, the error tests are separate from the correctness tests. We only apply this decorator to the correctness tests to avoid calling FusionCache.reset().
> Can this replace setUpModule? The functionality seems to be identical.

👍🏼

rdspring1:
!build

tests/python/test_python_frontend.py
```diff
@@ -126,6 +82,12 @@ def inner_fn(*args, **kwargs):
     return inner_fn
 
 
+def setUpModule():
+    from utils import atexit_serde_check
+
+    atexit_serde_check()
+
```
Priya2698, line 89: Can we use this as a decorator directly?

rdspring1: `atexit_serde_check` isn't a decorator. It is a regular function. `atexit_serde_create_op_test` is a pseudo-decorator because it creates many functions from a single function.

Priya2698: Right, sorry, I misread.
A follow-up question then is: Can we use `atexit_serde_check` directly? Why is the wrapper around it necessary?

rdspring1: `setUpModule` is a unittest fixture that runs once before running any tests in `TestNvFuserFrontend`. We only want to call `atexit_serde_check` once.

Issue #2755, opened 2024-08-02
makeInitialSchedulerEntry requires more than the current segment's inputs. 


wujingyue:
# Repro

This is a follow-up to https://github.com/NVIDIA/Fuser/pull/1649#discussion_r1535920468. 

I created a simple repro in wjy/input, which you can run by
```
$ git fetch origin wjy/input
$ git checkout wjy/input
$ NVFUSER_DUMP=segmented_fusion python tests/python/test_python_frontend.py -k 2755
```

The test: 
```python
    def test_issue2755(self):
        def fusion_func(fd: FusionDefinition) -> None:
            t0 = fd.define_tensor(shape=[-1])
            t1 = fd.ops.slice(
                t0,
                start_indices=[0],
                end_indices=[5],
            )
            t2 = fd.ops.neg(t1)
            t3 = fd.ops.slice(
                t2,
                start_indices=[0],
                end_indices=[2],
            )
            t4 = fd.ops.neg(t3)
            fd.add_output(t4)

        inputs = [torch.randn((10,), dtype=torch.float32, device="cuda:0")]
        self.exec_nvfuser(fusion_func, inputs)
```

Error message
```
RuntimeError: inner_size_opt.hasValue() INTERNAL ASSERT FAILED at "/opt/pytorch/nvfuser/csrc/scheduler/vectorize_helper.cpp":861, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. Vectorization heuristic could not evaluate inner most size: i116
Exception raised from getVectorizationFactor at /opt/pytorch/nvfuser/csrc/scheduler/vectorize_helper.cpp:861 (most recent call first):
```

The segmented fusion:
```
Segmented_Fusion Dump: -- Re-written complete fusion:{
Inputs:
  T0_g[ iS0{i0} ], float
Outputs:
  T4_g[ iS11{( fmax(0, ( fmin(( fmax(0, ( fmin(i0, 5) )) ), 2) )) )} ], float

%kernel_math {
i10 = fmin(i0, 5);
i12 = fmax(0, i10);
T1_l[ iS7{( fmax(0, ( fmin(i0, 5) )) )}rf ]
   = slice( T0_g[ iS0{i0} ], { {0, i12, 1} } )
T2_g[ iS9{( fmax(0, ( fmin(i0, 5) )) )} ]
   = -T1_l[ iS7{( fmax(0, ( fmin(i0, 5) )) )}rf ];
[...]

PR #2761, opened 2024-08-05
Add a regression test for #2755.

wujingyue:
This is similar to test_issue1277 but much simplified.


tests/python/test_python_frontend.py
```diff
@@ -3123,6 +3123,26 @@ def fusion_func(fd: FusionDefinition) -> None:
         torch_ref = inputs[0] * (inputs[1] * inputs[2]).unsqueeze(-1)
         self.assertEqual(nvf_out[0], torch_ref)
 
+    def test_issue2755(self):
```
rdspring1, line 3126: Can we add a small TODO that we expect this test to fail with `wjy/input` but pass on ToT main?

rdspring1 (approved):
LGTM

Issue #2056, opened 2024-04-08
FusionDefinition cannot handle inputs on new device.


jjsjann123:
`FusionDefinition` with manual `schedule` function breaks device handling.

The code logic doesn't look right. We are accessing out of bound entry here with the device id:
https://github.com/NVIDIA/Fuser/blob/618c4b79a053692bf83ebb32aa19dcc907ebb534/csrc/python_frontend/fusion_definition.cpp#L182-L183

```
Thread 1 "pt_main_thread" hit Catchpoint 1 (exception thrown), 0x00007ffff77fe4a1 in __cxa_throw () from /lib/x86_64-linux-gnu/libstdc++.so.6
(gdb) bt
#0  0x00007ffff77fe4a1 in __cxa_throw () from /lib/x86_64-linux-gnu/libstdc++.so.6
#1  0x00007ffff77f54cd in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#2  0x00007ffe86b0e430 in nvfuser::python_frontend::FusionCache::queryUserSchedule(nvfuser::python_frontend::FusionSchedules const*, unsigned long, int) const () from /opt/pytorch/nvfuser/nvfuser/_C.cpython-310-x86_64-linux-gnu.so
#3  0x00007ffe86b2266e in nvfuser::python_frontend::FusionDefinition::execute(c10::ArrayRef<c10::IValue> const&, bool, bool, std::optional<signed char>) const () from /opt/pytorch/nvfuser/nvfuser/_C.cpython-310-x86_64-linux-gnu.so
#4  0x00007ffe864e4645 in nvfuser::python_frontend::initNvFuserPythonBindings(_object*)::{lambda(nvfuser::python_frontend::FusionDefinition&, pybind11::iterable const&, bool, std::optional<long>, bool)#17}::operator()(nvfuser::python_frontend::FusionDefinition&, pybind11::iterable const&, bool, std::optional<long>, bool) const [clone .constprop.0] [clone .isra.0] ()

```

repro script vvv:
```
import torch
from nvfuser import FusionDefinition, DataType

inputs = [
    torch.randn(8, 8, 8, dtype=torch.float64, device='cuda:0'),
]

class FDScheduler(FusionDefinition):
    def definition(self):
        self.t0 = self.from_pytorch(inputs[0])
        self.t1 = self.ops.sum(self.t0, dim=-1)
        self.add_output(self.t1)

    # if we comment out manual scheduling, the program runs.
[...]

PR #2769, opened 2024-08-07
Run user-scheduled fusions with input arguments on different devices.

rdspring1:
This PR fixes issue #2056.

In the python frontend, user scheduled fusions are cached based on their input arguments and cuda device. Currently, we cannot run a user-scheduled fusion with inputs on a multiple device.

Proposed fix:
1. Check if a user-schedule fusion exists for given input arguments and device. The associated functions are `FusionCache::existsUserSchedule` and `FusionDefinition::existsSchedule`.
2. If not, create a new user schedule.

Secondary Issues:
- We are now calling `setupSchedule` and `finalizeSchedule` for each new `UserSchedule`. During scheduling, new tensors are created and added to the `FusionDefinition`. These schedule-only tensors must be removed in `setupSchedule` prior to scheduling the fusion.
- To address this, we track the number of recording states in prescheduled fusion. Then, pop the schedule-only states in `setupSchedule`.

### DIFF
--- a/csrc/python_frontend/fusion_definition.cpp
+++ b/csrc/python_frontend/fusion_definition.cpp
@@ -194,6 +194,8 @@ void FusionDefinition::setupSchedule(const at::ArrayRef<c10::IValue>& inputs) {
   // members that represent tensors would refer to the IR objects in the
   // original and not the copy needed for scheduling.
   buildFusionIr(user_sched_->schedule.get());
+  NVF_ERROR(num_recording_states_presched_ == 0);
+  num_recording_states_presched_ = recording_state_.size();
 
   // Add TensorViews created by composite operations to Python FusionDefinition.
   findHiddenTensorViews(user_sched_->schedule.get());
@@ -240,6 +242,15 @@ void FusionDefinition::finalizeSchedule(
   user_sched_->runtime_info.reset();
   prev_fusion_ = nullptr;
   user_sched_ = nullptr;
+
+  // Scheduling the fusion can add states to recording_state.
+  // Remove the schedule-only states.
+  size_t num_states_to_remove =
+      recording_state_.size() - num_recording_states_presched_;
+  for (size_t _ : c10::irange(num_states_to_remove)) {
+    recording_state_.pop_back();
+  }
+  num_recording_states_presched_ = 0;
 }
 
 void FusionDefinition::print(std::ostream& os) const {
--- a/csrc/python_frontend/fusion_definition.cpp
+++ b/csrc/python_frontend/fusion_definition.cpp
@@ -169,6 +169,16 @@ void FusionDefinition::updateSymbolicStates(
   }
 }
 
+bool FusionDefinition::existsSchedule(const at::ArrayRef<c10::IValue>& inputs) {
+  FUSER_PERF_SCOPE("FusionDefinition::existsSchedule");
+  NVF_CHECK(id().has_value(), "FusionDefinition definition does not exist!");
+  FusionSchedules* scheds = fusionCache()->queryFusionSchedules(id().value());
+  int8_t device = getCommonDeviceCUDA(inputs);
+  NVF_CHECK(
+      inputs.empty() || device > -1, "Inputs are not all on the same device!");
+  return fusionCache()->existsUserSchedule(scheds, inputs, device);
+}
+
 void FusionDefinition::setupSchedule(const at::ArrayRef<c10::IValue>& inputs) {
   FUSER_PERF_SCOPE("FusionDefinition::setupSchedule");
   NVF_CHECK(id().has_value(), "FusionDefinition definition does not exist!");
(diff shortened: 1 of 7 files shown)


rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

csrc/python_frontend/python_bindings.cpp
```diff
@@ -651,6 +651,15 @@ void initNvFuserPythonBindings(PyObject* module) {
             // Mark the end of a definition
             inst::Trace::instance()->endEvent(nullptr);
           })
+      .def(
+          "_exists_schedule",
```
jjsjann123, line 655: nitpick `_exist_schedule`, since we have `_setup_schedule`

csrc/python_frontend/python_bindings.cpp
```diff
@@ -651,6 +651,15 @@ void initNvFuserPythonBindings(PyObject* module) {
...
+          [](FusionDefinition& self, const py::iterable& iter) {
+            std::vector<c10::IValue> inputs;
+            for (py::handle obj : iter) {
+              inputs.push_back(torch::jit::toIValue(obj, c10::AnyType::get()));
+            }
+            self.existsSchedule(inputs);
```
jjsjann123, line 661: ditto `existSchedule`

nvfuser/__init__.py
```diff
@@ -202,10 +201,17 @@ def execute(
...
 
-        # If schedule is defined by child class, make a schedule for inputs
-        if func_based_def and (super(type(self), self).schedule != self.schedule):
+        # If schedule is defined by child class and schedule is not defined for
+        # inputs, make a schedule.
+        is_fusion_definition_child_class = (
+            type(self) != FusionDefinition
+        ) and issubclass(type(self), FusionDefinition)
+        defined_schedule = (
+            is_fusion_definition_child_class
```
jjsjann123, line 211: naive question, why do we need to check for `is_fusion_definition_child_class` here?

i.e. if we have a derived `schedule` method but went through the `execute` method here, wasn't that always true?

tests/python/test_schedule_ops.py
```diff
@@ -1102,6 +1102,41 @@ def schedule(self):
         torch_ref = torch.abs(inputs[0]).reshape(inputs[1].shape) + inputs[1]
         self.assertEqual(nvf_out[0], torch_ref)
 
+    def test_inputs_with_different_devices(self):
```
jjsjann123, line 1105:
```suggestion
@unittest.skipIf(torch.cuda.device_count() < 2, "more than 1 GPU required")
    def test_inputs_with_different_devices(self):
```

csrc/python_frontend/fusion_definition.cpp
```diff
@@ -105,6 +105,9 @@ void FusionDefinition::finalizeDefinition() {
     trie_node_ = child_node.value();
     fusion_id_ = std::optional<size_t>(trie_node_->fusion_id);
   }
+
+  NVF_ERROR(num_recording_states_presched_ == 0);
```
jjsjann123, line 109: nitpick, std::nullopt might read easier here. Also an error message would be nice.

csrc/python_frontend/fusion_definition.cpp
```diff
@@ -231,6 +243,14 @@ void FusionDefinition::finalizeSchedule(
...
   prev_fusion_ = nullptr;
   user_sched_ = nullptr;
+
+  // Scheduling the fusion can add states to recording_state.
+  // Remove the schedule-only states.
+  size_t num_states_to_remove =
+      recording_state_.size() - num_recording_states_presched_;
+  for (size_t rnd = 0; rnd < num_states_to_remove; ++rnd) {
+    recording_state_.pop_back();
+  }
```
jjsjann123, line 253: another naive question: should this be done in `setupSchedule` instead?

My understanding is that, if I'm playing with this without using a context manager for scheduling. I could potentially create new tensor with things like

`t = fd.sched.rfactor(...)`

If I manually call `fd._finalize_schedule()`, the `t` would vanish afterwards, even though I still have a handle in python.

jjsjann123 (commented):
Looks great! thanks for getting to this one! I have some minor comments/questions.

nvfuser/__init__.py
rdspring1: We have the context manager option that won't create a schedule method. We
```python
  # create a fusion
out = fd.execute(inputs, device=device)
```

rdspring1: I think we should simplify this check to `hasattr(self, "schedule") and callable(self.schedule)` and remove `schedule` from `FusionDefinition`

rdspring1:
!build

csrc/python_frontend/fusion_definition.cpp
rdspring1: I added an error message.

rdspring1: It is a part of FusionDefinition state, so I don't see it as `std::optional`.

csrc/python_frontend/fusion_definition.cpp
rdspring1: Ah. Python where you can access anything, anywhere, anytime.

rdspring1: Okay. Moved that to `setupSchedule`. I also removed resetting `user_sched_` to `nullptr` in `finalizeSchedule`, so users can print user schedule afterwards.

rdspring1:
!build

nvfuser/__init__.py
jjsjann123: Got ya. Thanks for showing that example.

jjsjann123 (approved):
LGTM

PR #2714, opened 2024-07-30
Re-write replaceSymbolicSizes using IdModel

jacobhinkle:
This uses IdModel to implement `replaceSymbolicSizes`. Extents are replaced with a single representative from their exact graph ValGroup with the following precedence:
1. Constants are preferred
2. If no constants exist, prefer the extents of fusion inputs.
3. Ties are broken by choosing the scalar with the smallest `name()`.

Fixes #2702. Fixes #2766

PR #2771, opened 2024-08-07
Always use TIDy in block outer reduction

### LINKED ISSUE #2770: Promote reuse over specialization for small reductions
https://github.com/NVIDIA/Fuser/blob/6d280618a31bbaf8c39770b69ac4b82bd67c6dcd/csrc/scheduler/reduction.cpp#L607

When the reduction size is small, we may end up having `bdimy` of just 1, which disables using `TIDy`. I think we should just use `TIDy` always so that we could reuse scheduled kernels as much as possible.

This showed up in the repro of #2766. While it's not the root case, changing it to use `TIDy` always works around the problem. I'll have a true fix for the issue later.

naoyam:
Fixes #2770


naoyam:
!build --diff

naoyam:
Issue fixed by #2714. Not necessary anymore.

PR #2776, opened 2024-08-07
remove no longer required runtime check of persistent batch size in innerOuter scheduler

liqiangxl:
After adding `getPersistentBufferStorageParams()`, the buffer size is correctly set and no longer require additional checks of whether we can get a "reasonable" persistent batch size.

This PR changes `getOptionalInnerOuterPersistentBufferBatches` to `getBufferBatchSizeAndThreadsPerBlock`, as suggested by the name change, `optional` is deleted indicating it always return a pair of {BufferBatchSize, ThreadsPerBlock}


liqiangxl:
!build --diff-bench

liqiangxl:
code diff comes from cases with hidden size can't be vectorized, e.g. `16384 x 15057`, before this PR they were segmented, with this PR they are using one kernel.

liqiangxl:
Since code diff indicates some code changes for ln bwd, further checked perf changes. Result shows increased perf due to the fusion is no longer segmented.
image

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -573,28 +564,7 @@ getOptionalInnerOuterPersistentBufferBatches(
...
-  // bytes stack frame, 192 bytes spill stores, 384 bytes spill loads.
-  // (3) shape= 16384 x 20480, 730 GB/s time_us mean(var)= 2766.64 (12.3883),
-  // 320 bytes stack frame, 320 bytes spill stores, 640 bytes spill loads. As a
-  // ref, the segmented version takes time_us mean(var)= 2841.91 (5.20231)
-  // without considering the overhead of fusion segmentation.
-  // (4) Disable this optimization if vectorize_factor is 1 due to high register
-  // usage in cases can't be vectorized.
-  const int64_t batch_max_reg_spill =
-      vectorize_factor > 1 ? batch_max + 3 : batch_max;
-  if (ignore_register_size_limit || inner_batch <= batch_max_reg_spill) {
```
jjsjann123, line 593: my follow up question here is that, we are dropping the check on batch_max_reg_spill here and apparently this code path is stil hit in some benchmark.

how comfortable are we with the blast radius here? :wink:

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -647,22 +617,6 @@ bool InnerOuterPersistentKernelScheduler::canScheduleRunTime(
...
-           properties.total_iteration_numel,
-           buffer_params.regs_buffer_size,
-           (int64_t)vectorize_factor,
-           warp_size,
-           false)
-           .first.has_value()) {
-    scheduler_debug_utils::canScheduleRejectReason(
-        heuristicType(),
-        "Required batch number is larger than available batch number! Will cause register spills!");
-    return false;
```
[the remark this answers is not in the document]
liqiangxl: yes.

csrc/scheduler/normalization_inner_outer.cpp
liqiangxl: Previously many cases may hit this check, but after adding `getPersistentBufferStorageParams()`, buffer size is correctly calculated for both fp16 & fp32 (previously, buffer size for fp32 is under estimated since we enforce project to inputs during schedule),  only cases can't be vectorized may hit this removed check. Should be fine.

PR #2781, opened 2024-08-07
Clean up ExprEvalScheduler::canScheduleCompileTime.

wujingyue:


wujingyue:
!build

csrc/scheduler/expr_eval_sched.cpp
```diff
@@ -19,21 +19,26 @@ bool ExprEvalScheduler::canScheduleCompileTime(Fusion* fusion) {
...
       scheduler_debug_utils::canScheduleRejectReason(
           heuristicType(),
           "Matmul ATen evaluation was disabled by NVFUSER_DISABLE=matmul_expr_eval");
+      return false;
     }
-  } else {
-    scheduler_debug_utils::canScheduleRejectReason(
-        heuristicType(),
-        "Fusion must contain only a single expression of type MatmulOp/LinearOp/SdpaFwdOp/SdpaBwdOp");
+    return true;
```
Priya2698, line 36: The logic seems inverted here -- `if (!isOptionDisabled(DisableOption::MatmulExprEval))` should return `false`.

csrc/scheduler/expr_eval_sched.cpp
```diff
@@ -19,21 +19,26 @@ bool ExprEvalScheduler::canScheduleCompileTime(Fusion* fusion) {
   if (exprs.size() != 1) {
     scheduler_debug_utils::canScheduleRejectReason(
         heuristicType(), "Fusion must contain only a single expression.");
```
jacobhinkle, line 21: Out of curiosity, where does the requirement that there is only a single expression come from?

Priya2698: This is a remnant of the design choice we made earlier during implementation of `matmul` and `linear`.

We can revisit and lift this restriction -- I don't see any reason why we cannot support it.
Will test with a few cases.

wujingyue:
!build

csrc/scheduler/expr_eval_sched.cpp
wujingyue: Thanks!

Priya2698 (approved):
LGTM.

PR #2793, opened 2024-08-15
Use FusionExecutorCache in test_multidevice_sharding.cpp.

wujingyue:
This partially implements Step 5 of [this plan](https://docs.google.com/document/d/1Wvh_vJYmA_LETfVNovb8mXPK95zmPOTUccauF-DX2h8/edit#heading=h.lg68cxhmn0y7).

### DIFF
--- a/tests/cpp/test_multidevice_sharding.cpp
+++ b/tests/cpp/test_multidevice_sharding.cpp
@@ -5,10 +5,10 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 // clang-format on
-#include <fusion.h>
 #include <gtest/gtest.h>
-#include <multidevice/executor.h>
-#include <multidevice/utils.h>
+
+#include <fusion.h>
+#include <kernel_cache.h>
 #include <ops/all_ops.h>
 #include <tests/cpp/multidevice.h>
 #include <tests/cpp/validator.h>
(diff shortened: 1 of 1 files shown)


wujingyue:
!build

samnordmann (approved):
LGTM, thanks!

Nevertheless, I'm concerned that completely replacing `MultiDeviceExecutor` by `FusionExecutorCache` will not test the implementation of `MultiDeviceExecutor` anymore, in particular the segmentation + the whole Host Ir stack.

Is there a plan to use Host IR in FusionExecutorCache? Or can we plan to keep testing MultiDeviceExecutor in at least part of the multidevice tests?

Wdyt ?

wujingyue:
!build

wujingyue:
> Is there a plan to use Host IR in FusionExecutorCache? Or can we plan to keep testing MultiDeviceExecutor in at least part of the multidevice tests?

Good point -- I've been debating with myself on this as well. 

The former is the right fix. It simplifies so we can remove MultiDeviceExecutor and several knobs (e.g. for inter-device segmentation and for disabling auto scheduling). My intuition says it won't be too complicated unless we try to also support decomposition. 

If we do the latter instead, what tests would you like to keep to exercise MultiDeviceExecutor?

samnordmann:
> > Is there a plan to use Host IR in FusionExecutorCache? Or can we plan to keep testing MultiDeviceExecutor in at least part of the multidevice tests?
> 
> Good point -- I've been debating with myself on this as well.
> 
> The former is the right fix. It simplifies so we can remove MultiDeviceExecutor and several knobs (e.g. for inter-device segmentation and for disabling auto scheduling). My intuition says it won't be too complicated unless we try to also support decomposition.
> 
> If we do the latter instead, what tests would you like to keep to exercise MultiDeviceExecutor?

Ok! Using Host IR in FusionExecutorCache would be great indeed. I agree it shouldn't be too hard... If we manage to add this relatively soon, then we shouldn't worry about removing `MultiDeviceExecutor`. Otherwise, I would probably keep testing `MultiDeviceExecutor` through the tests in `csrc/multidevice/communicator.cpp`, which range over a variety of scenarios.

PR #2802, opened 2024-08-19
Fix index replacement for circular buffering with trivial main loop.

rdspring1:
## Problem ##
We encounter an indexing issue when the main loop is trivial with TMA circular buffering.

Fusion Details:
- Inner Reduction on 2D tensor
- Number of Stages: 4
- Tensor Size: (128, 128)

NOTE: This TMA load is for the last stage in the pipeline. The y-axis index is not shifted by `(num_stages - 1)`.
**Incorrect indexing:**
```cpp
  nvfuser_index_t i5;
  i5 = 4 * ((nvfuser_index_t)blockIdx.x);
  // tma load for last stage in main loop
  Hopper::cpAsyncBulkTensorTileG2S((Hopper::CpAsyncBulkTensorTileG2SIndex<2>{ 
                ptr4, (Array<nvfuser_index_t, 2, 1>{(256 * i17), i5}), toSmem((&T3[i19])) }), (i8 + (1024 * i17)));
```

**Correct indexing:**
```cpp
  nvfuser_index_t i5;
  i5 = 4 * ((nvfuser_index_t)blockIdx.x);
  nvfuser_index_t i8;
  i8 = 3 + i5;
  // tma load for last stage in main loop
  Hopper::cpAsyncBulkTensorTileG2S((Hopper::CpAsyncBulkTensorTileG2SIndex<2>{ 
                ptr4, (Array<nvfuser_index_t, 2, 1>{(256 * i18), i8}), toSmem((&T3[i20])) }), (i9 + (1024 * i18)));
```

## Explanation ##
Since main loop is trivial, `getLoopIndexOfCircularBufferLoop` returns the start value of for-loop.
In `IdGraphIndexCompute`, we apply `SimplifyingIrBuilder` that removes the trivial addition with zero start value.
The `getIndexReplacementMap` function adds an entry to replace the loop index with offset for circular buffer load.
However, we fail to replace the loop index while processing `ir_utils::replaceValRecursively` because the loop index does not exist. 

## Proposed Solution ##
1. Always return index of for-loop in `getLoopIndexOfCircularBufferLoop`. 
2. In getIndexReplacementMap, replace index with zero start-value. 
3. Apply `simplifyExpr` to remove trivial computation.

Essentially, the fix is avoid simplifying too early, so we can replace it with the correctly offset index.

## PR side-effect ## 
Expression simplification is disabled in `IndexValidator` and `PredicateIndexValidator`. 
Trivial computations appear in the reference `getLinearIndex` and `getInlinePredicate`.

## Cuda Kernel References ##
- Run `./build/nvfuser_tests --gtest_filter='*CircularBufferingTest*Reduction/12'` on `tma_circular_buffering` branch at 56f18831.
1. [fail_reduction_cb4_128_128.txt](https://github.com/user-attachments/files/16665425/fail_reduction_cb4_128_128.txt)
2.  [pass_reduction_cb4_128_128.txt](https://github.com/user-attachments/files/16665427/pass_reduction_cb4_128_128.txt)

### DIFF
--- a/csrc/id_model/indexing.cpp
+++ b/csrc/id_model/indexing.cpp
@@ -884,8 +884,8 @@ std::vector<Val*> TensorIndexer::getIndexFor(
     auto it = info.index_map.find(g);
     NVF_ERROR(
         it != info.index_map.end(), "Index not found for ", g->toString());
-    result.push_back(
-        ir_utils::replaceValRecursively(it->second, replacement_map));
+    result.push_back(simplifyExpr(
+        ir_utils::replaceValRecursively(it->second, replacement_map)));
   }
   return result;
 }
@@ -1133,9 +1133,10 @@ std::vector<PredicateInfo> TensorIndexer::getPredicates(
         predicate_domain->toString());
 
     Val* idx = idx_it->second;
-    Val* start_idx =
-        ir_utils::replaceValRecursively(idx, replacement_map_start);
-    Val* stop_idx = ir_utils::replaceValRecursively(idx, replacement_map_stop);
+    Val* start_idx = simplifyExpr(
+        ir_utils::replaceValRecursively(idx, replacement_map_start));
+    Val* stop_idx = simplifyExpr(
+        ir_utils::replaceValRecursively(idx, replacement_map_stop));
 
     // Generate predicates as follows:
     //
@@ -1151,12 +1152,10 @@ std::vector<PredicateInfo> TensorIndexer::getPredicates(
     info.loop_stage_ = loop_stage;
 
     info.start_predicate_ = SimplifyingIrBuilder::geExpr(
-        SimplifyingIrBuilder::addExpr(simplifyExpr(start_idx), zero_val),
-        zero_val);
+        SimplifyingIrBuilder::addExpr(start_idx, zero_val), zero_val);
 
     info.stop_predicate_ = SimplifyingIrBuilder::ltExpr(
-        SimplifyingIrBuilder::addExpr(
-            simplifyExpr(stop_idx), info.stop_offset_),
+        SimplifyingIrBuilder::addExpr(stop_idx, info.stop_offset_),
         predicate_domain->extent());
 
     info.predicated_domains_ = {predicate_domain};
--- a/csrc/id_model/indexing.cpp
+++ b/csrc/id_model/indexing.cpp
@@ -884,8 +884,8 @@ std::vector<Val*> TensorIndexer::getIndexFor(
     auto it = info.index_map.find(g);
     NVF_ERROR(
         it != info.index_map.end(), "Index not found for ", g->toString());
-    result.push_back(simplifyExpr(
-        ir_utils::replaceValRecursively(it->second, replacement_map)));
+    result.push_back(
+        ir_utils::replaceValRecursively(it->second, replacement_map));
   }
   return result;
 }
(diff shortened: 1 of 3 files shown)


csrc/id_model/indexing.cpp
```diff
@@ -884,8 +884,8 @@ std::vector<Val*> TensorIndexer::getIndexFor(
     auto it = info.index_map.find(g);
     NVF_ERROR(
         it != info.index_map.end(), "Index not found for ", g->toString());
-    result.push_back(
-        ir_utils::replaceValRecursively(it->second, replacement_map));
+    result.push_back(simplifyExpr(
```
naoyam, line 887: Is this `simplifyExpr` strictly required?

rdspring1: I looked at the cuda kernels from the modified tests and there wasn't a difference when I removed `simplifyExpr`. The expressions are likely simplified by scalar hoist pass.

It isn't necessary for the TMA circular buffering tests.

naoyam: Thanks. That was what I thought - since these vals are simplified later, it should not be necessary to run `simplifyExpr` here, which can be costly unlike `SimplifyingIrBuilder`.

naoyam: Let's remove the `simplifyExpr`.

csrc/id_model/indexing.cpp
```diff
@@ -1144,7 +1152,7 @@ std::vector<PredicateInfo> TensorIndexer::getPredicates(
     info.loop_stage_ = loop_stage;
 
     info.start_predicate_ = SimplifyingIrBuilder::geExpr(
-        SimplifyingIrBuilder::addExpr(start_idx, info.start_offset_), zero_val);
+        SimplifyingIrBuilder::addExpr(start_idx, zero_val), zero_val);
```
naoyam, line 1155: Why is this change?

csrc/id_model/indexing.cpp
```diff
@@ -1126,9 +1133,10 @@ std::vector<PredicateInfo> TensorIndexer::getPredicates(
         predicate_domain->toString());
 
     Val* idx = idx_it->second;
-    Val* start_idx =
-        ir_utils::replaceValRecursively(idx, replacement_map_start);
-    Val* stop_idx = ir_utils::replaceValRecursively(idx, replacement_map_stop);
+    Val* start_idx = simplifyExpr(
```
naoyam, line 1136: Similarly, are these `simplifyExpr` necessary?

csrc/id_model/indexing.cpp
```diff
@@ -1042,12 +1042,19 @@ std::unordered_map<Val*, Val*> TensorIndexer::getIndexReplacementMap(
...
                       expr, for_loop, id_model_)) {
             replacement_index = SimplifyingIrBuilder::addExpr(
-                replacement_index != nullptr ? replacement_index : cur_index,
+                for_loop->isTrivial() ? for_loop->start() : base_index,
                 circular_buffer_offset);
+          } else if (
+              for_loop->circularBufferLoopStage() !=
+              CircularBufferLoopStage::NotApplicable) {
+            replacement_index =
+                for_loop->isTrivial() ? for_loop->start() : base_index;
```
naoyam, line 1057: Why is this necessary?

csrc/id_model/indexing.cpp
```diff
@@ -1042,12 +1042,19 @@ std::unordered_map<Val*, Val*> TensorIndexer::getIndexReplacementMap(
...
                   getLoopIndexOffsetForProducerOfCircularBuffer(
                       expr, for_loop, id_model_)) {
             replacement_index = SimplifyingIrBuilder::addExpr(
-                replacement_index != nullptr ? replacement_index : cur_index,
+                for_loop->isTrivial() ? for_loop->start() : base_index,
                 circular_buffer_offset);
+          } else if (
+              for_loop->circularBufferLoopStage() !=
+              CircularBufferLoopStage::NotApplicable) {
+            replacement_index =
```
naoyam, line 1056: Isn't this replacement necessary always, no matter if it's for consumer or producer indexing?

csrc/id_model/indexing.cpp
rdspring1: I'll remove them.

csrc/id_model/indexing.cpp
rdspring1: Not necessary. It was a debugging change.

rdspring1:
!build

csrc/id_model/indexing.cpp
naoyam: If so, I wonder why nothing breaks without the replacement? Since the loop is trivial, no loop is generated, but without the replacement, the index keeps using the loop index, which should result in a compilation error, no?

rdspring1: I updated the PR to always do the replacement.

rdspring1: Since we're only doing circular buffering with TMA loads, it is always the producer index into global memory.
The consumer index for shared memory always starts from the base index.

rdspring1: Since id-model indexing supports TMA circular buffers, I'll refactor some of my manual indexing.

naoyam (approved):
LGTM

PR #2808, opened 2024-08-19
Preserve allocation domain when inserting segment_set.

wujingyue:
For #2599.

### DIFF
--- a/csrc/preseg_passes/mark_aliases_prepare.cpp
+++ b/csrc/preseg_passes/mark_aliases_prepare.cpp
@@ -12,6 +12,7 @@
 #include <ops/alias.h>
 #include <options.h>
 #include <preseg_passes/mark_aliases_prepare.h>
+#include <transform_replay.h>
 
 namespace nvfuser::preseg_passes {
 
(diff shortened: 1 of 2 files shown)


wujingyue:
!build

csrc/preseg_passes/mark_aliases_prepare.cpp
```diff
@@ -99,6 +100,16 @@ void insertSegmentSetAfter(
...
+  // Inherit the allocation domain from `use_of`. This is important to pass
+  // AliasTest.Bookend_SegmentSetPreservesAllocation.
+  TensorDomain* replayed_domain =
+      TransformReplay::replayCasP(
+          copy, use_of, -1, TransformReplayOptions().replayAllocation())
+          .first;
+  if (replayed_domain->hasAllocation()) {
+    copy->setAllocationDomain(
+        replayed_domain->allocation(), replayed_domain->contiguity());
+  }
```
jjsjann123, line 112: so previously when we forgot to update the allocation domain on the intermediate `copy` here, we ended up with two copies?!

jjsjann123 (approved):
:clap: 
Good catch!

csrc/preseg_passes/mark_aliases_prepare.cpp
wujingyue: In some cases, yes. markAliases treats an empty allocation domain as default.

PR #2810, opened 2024-08-20
Print StructHandle in toString(PolymorphicValue)

jacobhinkle:
Previously, attempts to print a `PolymorphicValue` containing a `StructHandle` would encounter the error
> Can not print N7nvfuser12StructHandleE : incompatible type
Exception raised from operator<< at /opt/pytorch/nvfuser/lib/dynamic_type/src/dynamic_type/dynamic_type.h:809

This would be encountered, for example, when binding tensor metadata in a PrecomputedValues then printing it (see updated test).

This PR addresses this by printing structs with their type name and fields e.g.
```
StructHandle<A>{x=2788, y=2.71828}
```


jacobhinkle:
!build

csrc/polymorphic_value.h
```diff
@@ -225,20 +225,7 @@ using PolymorphicValue = dynamic_type::DynamicType<
...
-       << "stride=" << t.strides() << ", dtype=" << t.dtype()
-       << ", device=" << t.device() << ", data_ptr=" << t.data_ptr() << ")";
-  } else if (v.is<std::monostate>()) {
-    ss << "std::monostate";
-  } else {
-    ss << v;
-  }
-  return ss.str();
-}
+NVF_API std::string toString(const PolymorphicValue& v);
```
wujingyue, line 228:
```suggestion
std::string toString(const PolymorphicValue& v);
```
Not sure if we need NVF_API. If all tests pass without this, I'd remove.

csrc/polymorphic_value.cpp
```diff
@@ -0,0 +1,49 @@
...
+  } else if (v.is<StructHandle>()) {
+    const StructHandle& hdl = v.as<StructHandle>();
+    StructType type = (v->*&StructHandle::type)();
+    ss << "StructHandle<" << type.name << ">{";
+    bool first = true;
+    for (size_t i : c10::irange(type.fields.size())) {
+      if (!first) {
+        ss << ", ";
+      }
+      first = false;
```
wujingyue, line 35:
```suggestion
      if (first) {
        first = false;
      } else {
        ss << ", ";
      }
```

csrc/polymorphic_value.cpp
```diff
@@ -0,0 +1,49 @@
...
+      std::string fieldname = type.fields.at(i).name;
```
wujingyue, line 36:
```suggestion
      const std::string& fieldname = type.fields.at(i).name;
```

wujingyue (approved):
LGTM but I'd let @zasdfgbnm double check.

tests/cpp/test_evaluator.cpp
```diff
@@ -478,6 +479,18 @@ TEST_F(ExprEvalTest, TensorMetaData) {
...
+  DebugStreamGuard dsg(ss);
+
+  evaluator.print();
+  {
+    // Now bind a PrecomputedValues and print
+    PrecomputedValues pv(&fusion);
+    evaluator.bindPrecomputedValues(&pv);
+    pv.bindInputs(KernelArgumentHolder::createKernelArgumentHolder({a}));
+    evaluator.print();
+  }
```
wujingyue, line 493: Is this just to verify things don't crash when printing? FWIW, we can https://github.com/NVIDIA/Fuser/blob/522779f20cf2cc06fde579dab1758a9560c1721b/tests/cpp/test_gpu3.cpp#L6827 to capture stdout if you intended to check the printed strings as well.

jacobhinkle: Yes, I just wanted to check that we do not hit an error when printing. But now that you mention it I will use this part to check that we are printing the variables we expect to be bound.

csrc/evaluator_common.cpp
```diff
@@ -222,8 +224,8 @@ void PrecomputedValues::print() const {
   debug() << "Precomputed Values:\n";
   for (auto i : c10::irange(symbols_.size())) {
     if (defined_[i]) {
-      debug() << symbols_[i]->toInlineString() << " = " << values_[i]
-              << std::endl;
+      debug() << symbols_[i]->toInlineString() << " = "
+              << PolymorphicValue_functions::toString(values_[i]) << std::endl;
```
jacobhinkle, line 228: Instead of using `operator<<` directly for `PolymorphicValue`, we need to use `toString`.

csrc/evaluator_common.cpp
```diff
@@ -103,6 +103,8 @@ std::vector<Val*> collectRuntimeUsedValues(Fusion* fusion) {
         ret.push_back(id->expandedExtent());
       }
     }
+    Val* metadata = fusion->metadataOf(tv);
+    ret.push_back(metadata);
```
jacobhinkle, line 107: We were not binding metadata to `PrecomputedValues`, which I think is unexpected. This fixes that (see change to `TensorMetaData` test).

jacobhinkle: On second thought, I'm splitting this into another small PR, with the accompanying test from #2714.

jacobhinkle: See #2812

PR #2812, opened 2024-08-20
Bind TensorMetaData to PrecomputedValues

jacobhinkle:
This was found when implementing #2714. We currently do not bind TensorMetaData for input tensors to PrecomputedValues. This means we cannot evaluate expressions that contain them, which can lead to errors. This PR binds these metadata structs, which I think is the expected behavior.


jacobhinkle:
!build

zasdfgbnm (approved):

jacobhinkle:
I believe the test failures are due to printing, which is fixed in #2810

jacobhinkle:
!build

jacobhinkle:
!build

jacobhinkle:
I am currently seeing failures since `NaiveValueMachine` is being asked to compute ops it had previously not needed to. I think the fix is simply to implement the missing ops, since this should be a more efficient path so the fact it's getting more exercise is probably a good thing.

jacobhinkle:
!build

jacobhinkle:
!build

jacobhinkle:
!build

jacobhinkle:
!build

PR #2813, opened 2024-08-20
cache query patch

jjsjann123:
`_exist_schedule` forgot to return the value, ended up with each execution iteration re-schedule the same fusion.

existing python tests needs to clear cache. Since we use input look up for cache key, which doesn't distinguish scalar value changes.

### DIFF
--- a/csrc/python_frontend/python_bindings.cpp
+++ b/csrc/python_frontend/python_bindings.cpp
@@ -658,7 +658,7 @@ void initNvFuserPythonBindings(PyObject* module) {
             for (py::handle obj : iter) {
               inputs.push_back(torch::jit::toIValue(obj, c10::AnyType::get()));
             }
-            self.existSchedule(inputs);
+            return self.existSchedule(inputs);
           })
       .def(
           "_setup_schedule",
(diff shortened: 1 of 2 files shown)


jjsjann123:
!build

rdspring1 (approved):
LGTM.

Does this get rid of the `Warning: You are overwriting the current user schedule for a definition!` message?

jjsjann123:
> LGTM.
> 
> Does this get rid of the `Warning: You are overwriting the current user schedule for a definition!` message?

Yep it did. I only notice this after seeing that warning and got suspicious about it.

jjsjann123:
cc'ing @xwang233 could you remind me of what's running inside the `jit_python_bc_advisory` tests? Is it taking the cpp code base and running with testing scripts from main?

That would explain the failure....

xwang233:
> cc'ing @xwang233 could you remind me of what's running inside the `jit_python_bc_advisory` tests? Is it taking the cpp code base and running with testing scripts from main?
> 
> That would explain the failure....

`jit_python_bc_advisory` runs python script from the merge base of your PR and main, on your PR build, to detect any python API changes. If there are changes, you should modify the version.txt file and bc_advisory job will be skipped.

PR #2818, opened 2024-08-20
Add `chunk` to `ops/alias.h` for convenience.

wujingyue:


wujingyue:
!build

csrc/ops/alias.cpp
```diff
@@ -840,4 +840,29 @@ TensorView* slice(
   return slice(inp, slices);
 }
 
+std::vector<TensorView*> split(
+    TensorView* in,
+    int64_t dim,
+    const int64_t num_slices) {
```
jacobhinkle, line 846: Should we assert that `num_slices > 0` here?

jacobhinkle (commented):
Seems useful. However, I would suggest naming it [chunk](https://pytorch.org/docs/stable/generated/torch.chunk.html) or possibly [unstack](https://numpy.org/doc/stable/reference/generated/numpy.unstack.html#numpy.unstack) to avoid confusion.

csrc/ops/alias.cpp
```diff
@@ -840,4 +840,29 @@ TensorView* slice(
...
+  const auto in_logical = TensorDomain::noReductions(in->getLogicalDomain());
+  const auto num_dims = static_cast<int64_t>(in_logical.size());
+  dim = wrapDim(dim, num_dims);
+  Val* dim_size = in_logical[dim]->extent();
+  Val* slice_size = SimplifyingIrBuilder::divExpr(
+      dim_size, IrBuilder::create<Val>(num_slices));
```
jacobhinkle, line 852: Should this be a `ceilDiv` expression instead? Consider splitting an axis of size 11 into 6 chunks. We would have 5 chunks of size 2 and one of size 1 using `ceilDiv`. Using `divExpr` we would have 5  chunks of size 1 and one of size 6. 

Using `ceilDiv` I think it's not always possible to return the requested number of chunks. The example here is illustrative: https://pytorch.org/docs/stable/generated/torch.chunk.html. In the last example there they are splitting an axis of size 13 into "6" chunks. If we use `divExpr` instead of `ceilDiv` here, then we'd use 5 chunks of size 2 then one of size _3_. However, using ceilDiv instead we'd have 4 chunks of size 3 then one of size 1.

I think it's probably fine to use `div` but we should clearly state the behavior for non-divisible chunking in the docs.

wujingyue: I wasn't aware of `torch.chunk` -- thank you! 

Yes, that's indeed closest to what I've implemented. I intended to support divisible chunking only. But when the size is symbolic it's impossible to assert evenness or to avoid chunks of negative size. So I used `divExpr` here and documented the behavior in `alias.h`. 

If I use `ceilDiv` here, will nvFuser silently fail on negative-size slices?

wujingyue:
> Seems useful. However, I would suggest naming it [chunk](https://pytorch.org/docs/stable/generated/torch.chunk.html) or possibly [unstack](https://numpy.org/doc/stable/reference/generated/numpy.unstack.html#numpy.unstack) to avoid confusion.

I'll consider `chunk` definitely. `unstack` is however different enough to confuse readers. It splits by the exact dimension size and the slices are squeezed.

csrc/ops/alias.cpp
jacobhinkle: If negative-size slices are requested, you will be given zero-size slices i.e. empty tensors. I think going with `divExpr` makes sense let's just be clear to mention that the last tensor might have _more_ elements than all the others.

zasdfgbnm:
We do have a [torch.split](https://pytorch.org/docs/stable/generated/torch.split.html), the interface is slightly different from `torch.chunk`. I suggest we have both.

csrc/ops/alias.cpp
wujingyue: FWIW, https://github.com/NVIDIA/Fuser/blob/1906d4f6eeca870e029e5ce8a646699b75043d68/csrc/dynamic_transform.cpp#L409-L414 errors out a negative size, which is good -- better than a silent error.

csrc/ir/builder.cpp
```diff
@@ -413,6 +413,12 @@ Val* SimplifyingIrBuilder::divExpr(Val* lhs, Val* rhs) {
   if (rhs->isOneInt()) {
     return lhs;
   }
+
+  if (lhs->isConst() && rhs->isConst()) {
+    return IrBuilder::create<Val>(
+        lhs->value() / rhs->value(), promoteType(lhs->dtype(), rhs->dtype()));
+  }
+
```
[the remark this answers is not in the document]
wujingyue: I no longer need this change and reverted it. What you said makes sense. I think it's actually a bug that SimpilfyingIrBuilder doesn't promote the dtype in this case.

wujingyue:
!build

csrc/ops/alias.cpp
jacobhinkle: Hmmmm. I thought we handled this by massaging the resize expressions themselves.  Essentially any integers are valid inputs to slice, and they get clipped appropriately, so I wonder why we are hitting a negative extent at this check...

naoyam:
I'm wondering if we should also have a new IR node for this operation. The reason is that in patterns like RoPE, we have:

```
t1, t2 = chunk(t0, 2)
t3 = concat(t2, t1)
```

From this, it's clear that `t0` and `t3` correspond to the same logical space, though rotated, which would be a valuable property for scheduling and lowering. However, if implemented slice, it would look like:

```
t1 = slice(t0, {0, t0/2})
t2 = slice(t0, {t0/2, t0});
t3 = concat(t2, t1)
```

Now, it's less clear that `t3` should be mapped with `t1`. I think it's feasible to recover that information, but why not just avoid losing the information from the beginning by explicitly representing it?

I'm not requesting any change, but just wanted to share what I'm thinking about.

wujingyue:
> I'm not requesting any change, but just wanted to share what I'm thinking about.

Even if we have that op (we can even make it as powerful as `split` which doesn't have to be even), nothing prevents model writers from using `slice`. So I'd rather focus our energy on improving how nvFuser deals with Resize.

wujingyue:
!build

wujingyue:
cc @xwang233 FYI, nvfuser-ci results didn't show up in "checks". (I was still able to see the results after I click into nvfuser-ci, so not a big deal.)

xwang233:
> cc @xwang233 FYI, nvfuser-ci results didn't show up in "checks". (I was still able to see the results after I click into nvfuser-ci, so not a big deal.)

Thanks for the ping. There were some failures in the reporting script. Will take a look.

jacobhinkle (approved):
LGTM

Issue #2202, opened 2024-05-05
Reduction scheduler does not handle allocation domain properly and trigger assert when reduction output has specified allocation domain


jjsjann123:
Found an repro here:
```
TEST_F(AllocationDomainTest, ReductionOnPermutedReshape) {
  auto fusion = std::make_unique<Fusion>();
  FusionGuard fg(fusion.get());

  TensorView* in = makeContigConcreteTensor({16, 12, 128, 192});
  in->setAllocationDomain(
      {in->axis(0), in->axis(2), in->axis(1), in->axis(3)}, true);
  TensorView* dqkv = permute(in, {0, 2, 1, 3});
  dqkv = reshape(dqkv, {16, 128, 12, 192}, {16, 128, 2304});
  TensorView* sum_out = sum(dqkv, {0, 1});
  sum_out->setAllocationDomain(
      {sum_out->axis(2), sum_out->axis(0), sum_out->axis(1)}, true);

  fusion->addInput(in);
  fusion->addOutput(sum_out);

  FusionExecutorCache fec(std::move(fusion));
  at::Tensor in_tensor =
      at::randn({16 * 12 * 128 * 192})
          .cuda()
          .as_strided({16, 12, 128, 192}, {128 * 12 * 192, 192, 12 * 192, 1});
  std::vector<at::Tensor> out_tensors = fec.runFusionWithInputs({in_tensor});
  testValidate(fec.fusion(), out_tensors, {in_tensor}, __LINE__, __FILE__);
}
```

I tried to patch it with just fixing the reduction properties: #2201 , but that hit another issue with vectorization.

I realized that the reduction scheduler assumes the inputs to reduction operation to have the same allocation domain as with its output, which is not true.
IIUC, inner reduction means that the reduction dimension contains the inner most iter domain on inputs, rather than its output.

jjsjann123:
Adding another python example show up in CI:

```
import torch
from nvfuser import FusionDefinition, DataType

def nvfuser_fusion_id33(fd : FusionDefinition) -> None :
    T0 = fd.define_tensor(shape=[1, -1, -1, -1], contiguity=[None, True, True, True], dtype=DataType.Float, is_cpu=False, stride_order=[3, 2, 1, 0])
    T1 = fd.define_tensor(shape=[1, 1, -1], contiguity=[None, None, True], dtype=DataType.Float, is_cpu=False, stride_order=[2, 1, 0])
[...]

PR #2819, opened 2024-08-21
set reduction type based on alloc domain (Step-1 of alloc domain support for reduction scheduler)

liqiangxl:
Step-1 to solve issue #2202 
set reduction type based on alloc domain


liqiangxl:
!build

PR #2822, opened 2024-08-21
[stmatrix] Store a 8x8 matrix with hardcoded indices

protonu:
This is an initial implementation of [stmatrix](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html?highlight=stmatrix#warp-level-matrix-store-instruction-stmatrix), where we test it on a 8x8 matrix. We do not schedule the consumer of stmatrix as yet - instead we hardcode indices. 

The generated code is [here.](https://gist.github.com/protonu/7599c6ef4a3f41f9ab5de8332788fe41)


protonu:
!build

csrc/device_lower/pass/index.cpp
```diff
@@ -1507,16 +1507,41 @@ void IndexLowering::handle(const LoadStoreOp* ldst) {
...
+
+      // T_shared[toSmem(T_shared) + 16 * tidx.x]
+      auto out_index = IrBuilder::addExpr(
+          IrBuilder::baseAddressExpr(dynamic_cast<TensorView*>(ldst->out())),
+          IrBuilder::mulExpr(
+              IrBuilder::create<Val>(16, DataType::Index),
+              IrBuilder::create<NamedScalar>("threadIdx.x", DataType::Index)));
+
+      out = IrBuilder::create<kir::TensorIndex>(
+          dynamic_cast<TensorView*>(ldst->out()), out_index, DataType::Null);
```
zasdfgbnm, line 1535: nit:
```suggestion
          dynamic_cast<TensorView*>(ldst->out()), out_index);
```

tests/cpp/test_memory.cpp
```diff
@@ -2504,6 +2504,46 @@ TEST_P(LdMatrixTest, Regular) {
...
+  int sizeN = 8;
+
+  auto tv0 = makeContigConcreteTensor({sizeM, sizeN}, DataType::Half);
+  fusion.addInput(tv0);
+  // tv0 (global) -> tv1 (shared)
+  auto tv1 = set(tv0);
+  tv1->setMemoryType(MemoryType::Shared);
+  auto tv2 = set(tv1);
+  // tv1 (shared) -> tv2 (registers)
+  tv2->definition()->as<LoadStoreOp>()->setOpType(LoadStoreOpType::LdMatrix);
```
zasdfgbnm, line 2525: Instead of using ldmatrix, could we use plain set op to load data directly from gmem to register? The reason for doing so is, I believe we should make unit test really "unit". For example, if there is a bug in the implementation of `ldmatrix`,  this test can still pass as long as long as it is wrong the same way as ldmatrix.

protonu: Thanks! Please take another look at the test.

protonu:
!build

csrc/device_lower/pass/index.cpp
```diff
@@ -1507,16 +1507,41 @@ void IndexLowering::handle(const LoadStoreOp* ldst) {
           std::make_shared<DataType>(DataType::UInt32),
           (size_t)ir_utils::getVectorizeSize(ldst->out()->as<TensorView>()) /
               2};
+    } else if (ir_utils::isStMatrixOp(ldst)) {
+      as_type = ArrayType{
+          std::make_shared<DataType>(DataType::UInt32),
+          (size_t)ir_utils::getVectorizeSize(ldst->in()->as<TensorView>()) / 2};
```
zasdfgbnm, line 1513: I believe I understand why you write this part like this: you need to know the load size to determine whether it is `.x1`, `.x2`, or `.x4`. But I don't think the code here is correct. What we are checking here is the vectorization factor of `ldst->in()`, which is how we write the register tensor, instead of how we read register. What we actually do is to check the consumer's vectorization factor, for this case, the smem tv's vectorization factor. But unfortunately, in this PR, we are not scheduling it at all. How about we just hardcode this value again, instead of using a wrong value?

tests/cpp/test_memory.cpp
```diff
@@ -2504,6 +2506,74 @@ TEST_P(LdMatrixTest, Regular) {
...
+    // [8, 8] -> [8, 4, 2]
+    tv2->split(-1, factor, false);
+    // [8, 4, 2] -> [4, 8, 2]
+    tv2->reorder({{-2, -3}, {-3, -2}});
+    // [32, 4]
+    tv2->merge(-3);
+  }
+  tv2->axis(-2)->parallelize(ParallelType::TIDx);
+  // We do this only because codegen depends on this
+  tv2->axis(-1)->parallelize(ParallelType::Vectorize);
```
zasdfgbnm, line 2563: Is this part needed? Because we are not using ldmatrix to load, so we can just use a very simple loop domain to load, instead of scheduling the same way as ldmatrix? Why not just make the loop domain and allocation domain the same?

protonu:
!build

zasdfgbnm (approved):

protonu:
!build

PR #2773, opened 2024-08-07
[Tracking] TMA Circular Buffering

rdspring1:
## Summary ##
This PR adds support for TMA circular buffering. It is based on #1484. 

## Description ##
- In the allocation pass, create shared memory allocations and operations around `LoadStoreOp` expression.
- In the circular buffer pass, clone operations to create the pre-prologue, prologue, main, epilogue, and post-epilogue for-loops.
- Pre-Prologue allocates share memory and initializes mbarriers.
- Prologue copies only the load operations.
- Main loop copies the load and computation operations and adds `arrive_expected_tx` for next stage and `mbarrier_wait` for current stage.
- Epilogue copies only the computation operations and adds `mbarrier_wait` for remaining stages in the pipeline.
- Post-Epilogue invalidated mbarriers.

## Lowering Details ##
Description of changes in lowering passes.
- `Prologue`, `Main`, and `Epilogue` loops are created by `TmaCircularBufferLoopCloner` which is a child class of `CircularBufferLoopCloner`.
- `PrePrologue` and `PostEpilogue` loops are created by `createCpAsyncBulkFixtures`.
- The `cuTensorMapEncodeTiled ` restricts the size of each box dimension to be `<= 256`. You need to launch multiple load operations to load larger tiles. 
- We only allocate `mbarriers` for each stage, so the `expected_transaction` bytes is multiplied by the number of TMA loads per stage.
- The for-loop cloner must account for the nested for-loop structure used to launch multiple TMA loads before adding the `mbarrier_wait` for the stage.

<details>
<summary>Allocation Pass</summary>

```cpp
// Created tokens, mbarriers, init, and inval operations in allocation pass.
for (circular_buffer_loop) {
    __shared__ tokens[num_stages];
    __shared__ mbarrier[num_stages];
    init(mbarrier);
    cp.async.bulk(data, mbarrier);
    inval(mbarrier);
}
```
</details>

## Loop Structure ##
Description of for-loop structure for circular buffering.

<details>
<summary>Overview Circular Buffer Structure:</summary>

### Pre-prologue loop: ###
- Allocate shared memory for mbarriers and mbarrier tokens
- Initialize mbarrier for all stages

### Prologue loop: ###
- if selected_thread:
  - Issue cp async bulks for all but last stage

### Main loop: ###
- if selected_thread:
  - Issue next cp async bulk for available stage
- All threads wait until tma operation arrives
- Copy body without
  - shared memory allocations
  - mbarrier_init exprs
[...]

PR #2824, opened 2024-08-21
Allocation changes for TMA Circular Buffering

rdspring1:
## Summary ##
It is the changes to the allocation lowering pass from https://github.com/NVIDIA/Fuser/pull/2773.

### GpuLower ###
- `ldst_mbarrier_token_map_` maps `LoadStoreOp` to mbarrier tokens, which are represented as `TensorView` of number of pipeline stages.
- `mbarrier_token_smem_alloc_set_` tracks the `kir::Allocate` expressions for the mbarriers and their tokens.
- `ldst_mbarrier_index_map_` maps the cloned `LoadStoreOp` in the prologue and main loops to their indexed mbarrier.

### Allocation ###
- In the allocation pass, create shared memory allocations and operations around `LoadStoreOp` expression.

```cpp
// Created tokens, mbarriers, init, and inval operations in allocation pass.
for (circular_buffer_loop) {
    __shared__ int64_t tokens[num_stages];
    __shared__ int64_t mbarrier[num_stages];
    init(mbarrier);
    cp.async.bulk(data, mbarrier);
    inval(mbarrier);
}
```

## AliasMemory ##
- The mbarrier and its token are mapped together. The token is the mbarrier state of the last phase. For simplicity, mark token liveness when mbarrier is initialized and invalidated.
- Apply `markWrite` for mbarrier and its token when the expression is `MBarrierInit`
- Apply `markRead` for mbarrier and its token when the expression is `MBarrierInvalidate`

### DIFF
--- a/csrc/device_lower/lower2device.h
+++ b/csrc/device_lower/lower2device.h
@@ -226,6 +226,32 @@ class GpuLower : public NonCopyable {
     return ldst_mbarrier_map_;
   }
 
+  std::unordered_map<const Expr*, TensorView*>& ldstMBarrierTokenMap() {
+    return ldst_mbarrier_token_map_;
+  }
+
+  const std::unordered_map<const Expr*, TensorView*>& ldstMBarrierTokenMap()
+      const {
+    return ldst_mbarrier_token_map_;
+  }
+
+  std::unordered_set<const Expr*>& mBarrierTokenSmemAllocSet() {
+    return mbarrier_token_smem_alloc_set_;
+  }
+
+  const std::unordered_set<const Expr*>& mBarrierTokenSmemAllocSet() const {
+    return mbarrier_token_smem_alloc_set_;
+  }
+
+  std::unordered_map<const Expr*, kir::TensorIndex*>& ldstMBarrierIndexMap() {
+    return ldst_mbarrier_index_map_;
+  }
+
+  const std::unordered_map<const Expr*, kir::TensorIndex*>&
+  ldstMBarrierIndexMap() const {
+    return ldst_mbarrier_index_map_;
+  }
+
   bool isNvFuserZeroEnabled() {
     if (isOptionDisabled(DisableOption::MagicZero)) {
       return false;
@@ -358,6 +384,19 @@ class GpuLower : public NonCopyable {
   //! "extent mod split_factor == 0" and an error message for divisibility check
   //! for vectorization.
   std::vector<std::pair<const Val*, std::string>> validations_;
+
+  // Keep track of placeholders for tokens returned by arrive/expected tx
+  // mbarrier operations for each load/store operation that requires such
+  // synchronization
+  std::unordered_map<const Expr*, TensorView*> ldst_mbarrier_token_map_;
+
+  // Collection of kir::Allocate for smem buffers used for mbarrier and token
+  // objects from cpAsyncBulk synchronization
+  std::unordered_set<const Expr*> mbarrier_token_smem_alloc_set_;
+
+  // Keep track what mbarrier object is used in load/store operation that
+  // requires such synchronization, required by indexing pass
+  std::unordered_map<const Expr*, kir::TensorIndex*> ldst_mbarrier_index_map_;
 
   Fusion* fusion_ = nullptr;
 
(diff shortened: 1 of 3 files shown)


rdspring1:
!build

rdspring1:
!build

csrc/device_lower/lower2device.h
```diff
@@ -359,6 +385,19 @@ class GpuLower : public NonCopyable {
...
+  // synchronization
+  std::unordered_map<const Expr*, TensorView*> ldst_mbarrier_token_map_;
+
+  // Collection of kir::Allocate for smem buffers used for mbarrier and token
+  // objects from cpAsyncBulk synchronization
+  std::unordered_set<const Expr*> mbarrier_token_smem_alloc_set_;
+
+  // Keep track what mbarrier object is used in load/store operation that
+  // requires such synchronization, required by indexing pass
+  std::unordered_map<const Expr*, kir::TensorIndex*> ldst_mbarrier_index_map_;
```
zasdfgbnm, line 399: Should we put them together as a `struct MBarrierInfo`?

csrc/device_lower/pass/allocation.cpp
```diff
@@ -345,6 +347,16 @@ class AllocationInserter : public kir::ExprMutator {
       auto alloc_expr = createAllocExpr(allocation, is_output);
       auto init_expr = createInitExpr(allocation, init);
 
+      // Find the largest circular buffer depth; Used for tma buffer allocation
```
zasdfgbnm, line 350: Should this be: `// Check that all circular buffer depth match`

csrc/device_lower/pass/allocation.cpp
```diff
@@ -378,33 +390,96 @@ class AllocationInserter : public kir::ExprMutator {
...
+        // add them here to handle shared memory correctly in alias memory pass.
+        TensorView* mbarrier =
+            TensorViewBuilder()
+                .shape(std::vector<int64_t>{circular_buffer_depth})
+                .dtype(DataType::UInt)
+                .contiguity(true)
+                .build();
+        mbarrier->setMemoryType(MemoryType::Shared);
+
+        // The wait condition for mbarrier is a single thread and the expected
```
zasdfgbnm, line 406: Instead of assuming there will be only a single thread issuing TMA load, should we check the parallelization of `expr->output(0)` to determine that value?

zasdfgbnm:
> * In the allocation pass, create shared memory allocations and operations around `LoadStoreOp` expression.
> 
> ```c++
> // Created tokens, mbarriers, init, and inval operations in allocation pass.
> for (circular_buffer_loop) {
>     __shared__ tokens[num_stages];
>     __shared__ mbarrier[num_stages];
>     init(mbarrier);
>     cp.async.bulk(data, mbarrier);
>     inval(mbarrier);
> }
> ```

Could you add this as a comment in the code?

csrc/device_lower/pass/allocation.cpp
rdspring1: I added the check but didn't updated the comment.

csrc/device_lower/pass/allocation.cpp
rdspring1: The number of threads needs to a constant when we call `arrive_expected_tx` in the cuda kernel. The number of threads can be a runtime value, so defaulting to a single thread is the simplest thing.

rdspring1: I'm using a regular `block_sync` for synchronizing the CTA rather than the `mbarrier` because of this compile-time constraint.

naoyam (commented):
LGTM. I defer to @zasdfgbnm.

csrc/device_lower/pass/allocation.cpp
zasdfgbnm: I still don't understand. Could you elaborate more on:
> The number of threads needs to a constant when we call arrive_expected_tx in the cuda kernel.

rdspring1: > The number of threads needs to a constant when we call arrive_expected_tx in the cuda kernel.

I was wrong about this ^^^.

I think your proposal is this:
```cpp
kir::MBarrierInit* mbarrier_init = IrBuilder::create<kir::MBarrierInit>(
    mbarrier,
    simplifyExpr(SimplifyingIrBuilder::maybeCastExpr(
        DataType::UInt32,
        lower_utils::getNumThreadsInTensorView(
            expr->output(0)->as<TensorView>()))));
```
I'm fine with this change.

rdspring1:
!build

zasdfgbnm (approved):
Stamping. I don't think there is anything in this PR that can not be resolved in a separate PR.

rdspring1:
!build

PR #2826, opened 2024-08-21
Sharded SDPA backwards support

cowanmeg:
Adds temporary support for sharded backwards scaled dot product attention. Until https://github.com/NVIDIA/Fuser/issues/2563 is completed. 

Similar to https://github.com/NVIDIA/Fuser/pull/2565
Similar restrictions:
1. All necessary sharded inputs are manually sharded before the SDPABwdOp is created. We cannot rely on sharding propagation or sharding after the Fusion is created, because the dimension checks are called when the op is created.
2. Only the head dimension is sharded and all inputs and outputs have either a sharded head dimension or unshaded.
3. DID axis is the outermost axis. This is because during evaluation if we see 5 dimensions, it is assumed the first is the DID axis and is appropriately squeezed from the inputs and unsqueezed onto the outputs.


csrc/ir/nodes.cpp
```diff
@@ -4859,9 +4859,16 @@ std::vector<PolymorphicValue> SdpaBwdOp::evaluate(
     const std::vector<PolymorphicValue>& inputs) const {
   // Backward tensor inputs: grad_input, query, key, value, output, logsumexp,
   // max_q/k
+  // Temporary handling of DID parallelization see
+  // https://github.com/NVIDIA/Fuser/issues/2563
+  bool handle_device_dim = inputs.at(1).as<at::Tensor>().dim() == 5;
```
wujingyue, line 4864:
```suggestion
  bool first_dim_is_did = inputs.at(1).as<at::Tensor>().dim() == 5;
```
may be clearer.

Priya2698:
> * All necessary sharded inputs are manually sharded before the SDPABwdOp is created. We cannot rely on sharding propagation or sharding after the Fusion is created, because the dimension checks are called when the op is created.

Do you think the location of these checks needs to be changed or conditioned on sharding to avoid manual sharding?

csrc/ir/nodes.cpp
```diff
@@ -4859,9 +4859,16 @@ std::vector<PolymorphicValue> SdpaBwdOp::evaluate(
...
   std::vector<at::Tensor> bwd_inputs;
   for (auto idx : c10::irange(6)) {
-    bwd_inputs.emplace_back(inputs.at(idx).as<at::Tensor>());
+    auto in_tensor = inputs.at(idx).as<at::Tensor>();
```
Priya2698, line 4867: Can we add a comment here mentioning the removal of DID axis, if present, prior to computation.

csrc/logical_domain_map.cpp
```diff
@@ -284,19 +284,25 @@ std::unordered_map<IterDomain*, IterDomain*> PairwiseLogicalDomainMap::map(
     bool consumer_has_e = consumer_tv_->sameAs(op->grad_query()) ||
         consumer_tv_->sameAs(op->grad_key());
 
+    size_t num_device_dim =
```
Priya2698, line 287: Can we update the comments in L266 - 275,  to mention a conditional DID axis.

Priya2698 (approved):
The updated `SdpaBwdOp::evaluate` and logical domain mapping look right to me.
I left some minor feedback around adding comments.

I will defer to @wujingyue to verify the changes in the preseg_pass.

csrc/ir/nodes.cpp
```diff
@@ -4859,9 +4859,16 @@ std::vector<PolymorphicValue> SdpaBwdOp::evaluate(
...
+    if (handle_device_dim) {
+      in_tensor = in_tensor.squeeze(0);
+    }
+    bwd_inputs.emplace_back(in_tensor);
```
wujingyue, line 4871: I know this didn't come from you, but would you mind changing it to `push_back`? https://abseil.io/tips/112

csrc/logical_domain_map.cpp
```diff
@@ -284,19 +284,25 @@ std::unordered_map<IterDomain*, IterDomain*> PairwiseLogicalDomainMap::map(
...
+        producer_logical.size() > 0 && producer_logical.at(0)->isDeviceDim()
+        ? 1
+        : 0;
```
wujingyue, line 290:
```suggestion
    NVF_ERROR(producer_logical.size() > 0);
    size_t num_device_dim = producer_logical.at(0)->isDeviceDim();
```

csrc/logical_domain_map.cpp
```diff
@@ -284,19 +284,25 @@ std::unordered_map<IterDomain*, IterDomain*> PairwiseLogicalDomainMap::map(
...
     for (auto idx : c10::irange(producer_logical.size())) {
-      if (idx < 2) {
+      if (idx < (2 + num_device_dim)) {
```
wujingyue, line 292:
```suggestion
      if (idx < 2 + num_device_dim) {
```

csrc/logical_domain_map.cpp
```diff
@@ -284,19 +284,25 @@ std::unordered_map<IterDomain*, IterDomain*> PairwiseLogicalDomainMap::map(
...
         // Map N, H from all producers to consumers
         updatePairwiseLogicalDomainMap(
             producer_logical.at(idx), consumer_root.at(idx));
-      } else if (idx == 2 && (producer_has_s == consumer_has_s)) {
+      } else if (
+          idx == (2 + num_device_dim) && (producer_has_s == consumer_has_s)) {
```
wujingyue, line 297: Parens are not used consistently. I'd do
```suggestion
          idx == 2 + num_device_dim && producer_has_s == consumer_has_s) {
```
or if you much prefer parens

```
          (idx == 2 + num_device_dim) && (producer_has_s == consumer_has_s)) {
```

csrc/logical_domain_map.cpp
```diff
@@ -284,19 +284,25 @@ std::unordered_map<IterDomain*, IterDomain*> PairwiseLogicalDomainMap::map(
...
         // producer/consumer[2] = L/S
         updatePairwiseLogicalDomainMap(
-            producer_logical.at(2), consumer_root.at(2));
-      } else if (idx == 3 && (producer_has_e == consumer_has_e)) {
+            producer_logical.at(idx), consumer_root.at(idx));
+      } else if (
+          idx == (3 + num_device_dim) && (producer_has_e == consumer_has_e)) {
```
wujingyue, line 302: ditto

csrc/multidevice/utils.cpp
```diff
@@ -140,6 +140,10 @@ int64_t numDeviceDims(const TensorView* tv) {
 bool haveDifferentShardings(
     const TensorView* producer,
     const TensorView* consumer) {
+  // cpu scalars are not required to have a mesh
+  if (producer->isCpuScalar() || consumer->isCpuScalar()) {
```
wujingyue, line 144: Is it possible to have `producer` being a sharded tensor while `consumer` being a CPU scalar? What would be the expected behavior in that case?

csrc/ops/composite.cpp
```diff
@@ -565,15 +565,40 @@ SdpfaBwdResult sdpfa_bwd(
...
+        "Only suport DID parallelization on outermost axis");
+    NVF_CHECK(
+        value_domain[0]->isDeviceDim(),
+        "Only suport DID parallelization on outermost axis");
+    NVF_CHECK(
+        grad_output->axis(0)->isDeviceDim(),
+        "Only suport DID parallelization on outermost axis");
+    NVF_CHECK(
+        output->axis(0)->isDeviceDim(),
+        "Only suport DID parallelization on outermost axis");
```
wujingyue, line 586:
```suggestion
    auto check_first_is_did = [](const std::vector<IterDomain*>& ids) -> void {
      NVF_CHECK(ids[0]->isDeviceDim(), "Only support DID parallelization on outermost axis");
    };
    check_first_is_did(query_domain);
    check_first_is_did(key_domain);
    check_first_is_did(value_domain);
    check_first_is_did(grad_output->getLogicalDomain());
    check_first_is_did(output->getLogicalDomain());
```

csrc/preseg_passes/propagate_shardings.cpp
```diff
@@ -50,16 +50,16 @@ void PropagateShardingsPass::runPass(Fusion* fusion) {
...
-    } else {
+    } else if (!tv->isCpuScalar()) {
       update_if_null(tv_without_mesh, tv);
     }
   }
   NVF_CHECK(
       tv_with_mesh == nullptr || tv_without_mesh == nullptr,
       "Found ",
-      tv_with_mesh,
+      tv_with_mesh->toString(),
```
wujingyue, line 60: Instead of this, include `<ir/iostream.h>`.

csrc/preseg_passes/propagate_shardings.cpp
```diff
@@ -50,16 +50,16 @@ void PropagateShardingsPass::runPass(Fusion* fusion) {
...
       " assigned a mesh and ",
-      tv_without_mesh,
+      tv_without_mesh->toString(),
```
wujingyue, line 62: ditto

csrc/logical_domain_map.cpp
Priya2698: The producer_size can be 0, for `philox_seed/offset` which are scalar tensors.
We just skip mapping for them, it is not an error.

tests/cpp/test_sdpa_node.cpp
```diff
@@ -826,6 +830,137 @@ TEST_F(SDPATest, Sharded_SdpaFwd) {
...
+  std::vector<int64_t> kv_shape({d, n, h / d, s, e});
+  std::vector<int64_t> attn_shape({d, n, h / d, l, e});
+
+  auto options = at::TensorOptions().dtype(at::kHalf).device(at::kCUDA, 0);
+  at::Tensor q = at::randn({n, h / d, l, e}, options);
+  at::Tensor k = at::randn({n, h / d, s, e}, options);
+  at::Tensor v = at::randn({n, h / d, s, e}, options);
+
+  double dropout_p = 0.2;
+  bool is_causal = false;
```
wujingyue, line 852:
```suggestion
  constexpr double dropout_p = 0.2;
  constexpr bool is_causal = false;
```

tests/cpp/test_sdpa_node.cpp
```diff
@@ -826,6 +830,137 @@ TEST_F(SDPATest, Sharded_SdpaFwd) {
   validateSdpaFwdOutputs(nvf_out, aten_out);
 }
 
+// TODO: Remove/update when https://github.com/NVIDIA/Fuser/issues/2563 is
+// resolved.
+TEST_F(SDPATest, Sharded_SdpaBwd) {
+  NVFUSER_TEST_CUDA_ARCH_GUARD(8, 0);
+  at::manual_seed(0);
```
wujingyue, line 837: Is this for dropout? I don't see the random seed gets reset in this test, so I'm confused about its purpose.

cowanmeg:
> > * All necessary sharded inputs are manually sharded before the SDPABwdOp is created. We cannot rely on sharding propagation or sharding after the Fusion is created, because the dimension checks are called when the op is created.
> 
> Do you think the location of these checks needs to be changed or conditioned on sharding to avoid manual sharding?

This is just temporary! We have the restriction that the sharded dimension has to be split at the logical level, but we have a todo task to allow sharding a loop iterdomain. When we finish that we can actually remove all this :D

csrc/multidevice/utils.cpp
cowanmeg: Hm I don't think we have any exprs that produce a CPU scalar, so I'll just check if the producer is a CPU scalar

tests/cpp/test_sdpa_node.cpp
cowanmeg: Ooops not needed, was doing too much copy paste

csrc/ir/nodes.cpp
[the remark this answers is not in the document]
cowanmeg: Changed to check if the query input is sharded on the first axis

cowanmeg:
!build

csrc/multidevice/utils.cpp
wujingyue: Doesn't SpdaFwdOp produce CPU scalars for seed and offset?

csrc/preseg_passes/propagate_shardings.cpp
```diff
@@ -50,7 +51,7 @@ void PropagateShardingsPass::runPass(Fusion* fusion) {
 
     if (tv->hasDeviceMesh()) {
       update_if_null(tv_with_mesh, tv);
-    } else {
+    } else if (!tv->isCpuScalar()) {
```
wujingyue, line 54: I'd change this to
```
if (tv->isCpuScalar()) {
  continue;
}

if (tv->hasDeviceMesh()) {
  update_if_null(...);
} else {
  update_if_null(...);
}
```
Same logic but slightly easier to follow -- CPU scalars are skipped as a corner case at the beginning.

wujingyue (approved):

csrc/multidevice/utils.cpp
Priya2698: > Hm I don't think we have any exprs that produce a CPU scalar, so I'll just check if the producer is a CPU scalar

`SdpaFwdOp` has philox seed and philox offset as CPU scalar consumers.

cowanmeg: ahh yes, you're right - changing back to the original

cowanmeg:
!build

PR #2827, opened 2024-08-21
Tensor parallel backwards MHA

cowanmeg:
Manually sharded backwards pass of multiheaded attention layer.

Backward pass is manually translated from thunder trace which performs activation recomputation of the first linear layer and formation of the QKV vectors. SDPA outputs are saved as inputs for the backwards pass.


cowanmeg:
!build

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -175,6 +175,100 @@ std::vector<at::Tensor> reference_mlp_backwards(
...
+    at::Tensor grad,
+    at::Tensor x,
+    at::Tensor mask,
+    at::Tensor w0,
+    at::Tensor b0,
+    at::Tensor w1,
+    at::ScalarType at_dtype) {
+  // recompute up to sdpa
+  auto m = at::matmul(x, w0).add(b0).view({B, S, 3 * E});
+  auto qkv_vec = m.split(E, 2);
```
wujingyue, line 188:
```suggestion
  auto qkv_vec = m.split(E, /*dim=*/-1);
```

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -175,6 +175,100 @@ std::vector<at::Tensor> reference_mlp_backwards(
...
+  for (auto i = 0; i < 3; i++) {
+    qkv_vec[i] =
+        qkv_vec[i].reshape({B, S, H, E / H}).transpose(2, 1).to(at_dtype);
```
wujingyue, line 191:
```suggestion
        qkv_vec[i].reshape({B, S, H, E / H}).transpose(1, 2).to(at_dtype);
```

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -295,55 +448,165 @@ std::vector<TensorView*> mlp_backwards(
...
+    // Explicitly shard qkv before calling SDPA node
+    for (auto tv : {tv_slice, tv_reshape, tv_trans, tv_cast}) {
+      tv->setDeviceMesh(mesh);
+      tv->axis(0)->parallelize(ParallelType::DIDx);
+    }
+  }
+
+  // Backwards
+  constexpr double kScale = 1.0 / (1.0 - kDropoutProb);
+  Val* dscale = IrBuilder::create<Val>(kScale);
```
wujingyue, line 532:
```suggestion
  auto* dscale = IrBuilder::create<Val>(kScale);
```

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -295,55 +448,165 @@ std::vector<TensorView*> mlp_backwards(
...
+  TensorView* dropout_grad = dropout_backward(grad, mask, dscale);
```
wujingyue, line 533: What does `d` in `dscale` stand for? If it means, I'd say `dropout_scale` for clarity.

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -535,6 +798,111 @@ TEST_P(DistributedTransformerTest, MLP_Backward) {
...
+      {tvw1, tvw0, tvb0},
+      {tvouts[1],
+       tvouts[2],
+       tvouts[3],
+       tvouts[4],
+       tvouts[5],
+       tvouts[6],
+       tvouts[7]},
+      tvw0);
+  shardBetween({tvx, tvmask, tvgrad}, {tvouts[0], tvouts[8]}, tvx);
```
wujingyue, line 860: Can you code-comment on why it's sharded this way? Some illustrative ASCII art may help.

cowanmeg: Comment added. The propagation is relatively simple, just have to cover all the outputs so that's why the list is long.

cowanmeg: + reduced the set in `to` since some were not leaves of the DAG and were covered by other TVs.

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -175,6 +175,100 @@ std::vector<at::Tensor> reference_mlp_backwards(
...
+  auto dropout_grad =
+      at::native_dropout_backward(grad, mask, 1.0 / (1.0 - kDropoutProb));
+  auto dropout_grad_q = dropout_grad.to(at_dtype);
+  auto matmul1_grad_x = at::matmul(dropout_grad_q, w1.transpose(0, 1));
+  // sdpa output: B, H, S, E/H
+  auto sdpa_output_reshape = sdpa_output.transpose(1, 2).view({B * S, E});
+  auto matmul1_grad_w =
+      at::matmul(dropout_grad_q.transpose(0, 1), sdpa_output_reshape)
+          .transpose(0, 1);
+  auto matmul1_grad_b = at::sum(dropout_grad, {0});
```
wujingyue, line 225:
```suggestion
  auto linear_grad_b = at::sum(dropout_grad, {0});
```
I slightly prefer calling it linear because matmul, strictly speaking, doesn't take a bias.

wujingyue (commented):
More to come...

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -274,6 +368,65 @@ std::vector<TensorView*> mha(
...
+// weight format: [DID(D), i1, i2/D] dtype
+// grad format: [DID(D) i0, i2/D] float
+// outputs: grad_x [i0, i1] dtype
+// grad_w [DID i1, i2/D] dtype
+// grad_b [DID i2/2] float
+LinearBackwardsResult linear_backwards(
+    TensorView* x,
+    TensorView* w,
+    TensorView* grad,
+    DataType dtype) {
```
wujingyue, line 389: Can we use `x->dtype()` instead? (One less parameter to worry about)

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -274,6 +368,65 @@ std::vector<TensorView*> mha(
...
+  TensorView* grad_q;
+  if (grad->dtype() == dtype) {
+    grad_q = grad;
+    grad = castOp(DataType::Float, grad_q);
+  } else {
+    grad_q = castOp(dtype, grad);
+  }
```
wujingyue, line 396:
```suggestion
  TensorView* grad_q = maybeCastOp(dtype, grad);
```
and replace `sum(grad, ...)` later with `sum(maybeCastOp(Float, grad))`.

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -175,6 +175,100 @@ std::vector<at::Tensor> reference_mlp_backwards(
...
+              v,
+              /*dropout_p=*/kSdpaProb,
+              /*is_causal=*/true,
+              /*return_debug_mask=*/false,
+              /*scale=*/kSdpaScale);
+
+  // backwards pass
+  auto dropout_grad =
+      at::native_dropout_backward(grad, mask, 1.0 / (1.0 - kDropoutProb));
+  auto dropout_grad_q = dropout_grad.to(at_dtype);
```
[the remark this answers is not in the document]
cowanmeg: correct

tests/cpp/test_multidevice_transformer.cpp
wujingyue: Btw, can `grad` ever be lower than float? Apparently, `mlp` and `reference_mlp` both output float.

cowanmeg: The incoming grads get quantized to half/bfloat16 for the backwards linear operations since the weights and bias are in lower precision.

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -293,57 +446,167 @@ std::vector<TensorView*> mlp_backwards(
...
+    TensorView* w1,
+    TensorView* mask,
+    TensorView* sdpa_output,
+    TensorView* sdpa_log_sumexp,
+    TensorView* sdpa_seed,
+    TensorView* sdpa_offset,
+    TensorView* grad,
+    const DeviceMesh& mesh,
+    DataType dtype) {
+  // Recompute: Linear 1, QKV vectors
```
wujingyue, line 507:
```suggestion
  // Recompute: Linear 0, QKV vectors
```
It looks like you are recomputing linear 0 instead. Correct?

tests/cpp/test_multidevice_transformer.cpp
cowanmeg: yes. will update all the helpers!

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -535,6 +798,106 @@ TEST_P(DistributedTransformerTest, MLP_Backward) {
...
+      shardTensor(reference_outs[10].view({E, 3, E}), 2, mesh)
+          .view({1, E, 3 * E / D}),
+      shardTensor(reference_outs[11].view({3, E}), 1, mesh)
+          .view({1, 3 * E / D}),
+      reference_outs[12]};
+
+  MultiDeviceExecutor runtime(
+      std::move(fusion), *communicator_, executor_params_);
+  at::manual_seed(getATenRandomSeed());
+  runtime.runWithInput(inputs);
```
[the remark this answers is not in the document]
cowanmeg: ooops that is a typo

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -293,57 +446,167 @@ std::vector<TensorView*> mlp_backwards(
...
+    TensorView* sdpa_log_sumexp,
+    TensorView* sdpa_seed,
+    TensorView* sdpa_offset,
+    TensorView* grad,
+    const DeviceMesh& mesh,
+    DataType dtype) {
+  // Recompute: Linear 1, QKV vectors
+  TensorView* mm = matmul(x, w0);
+  TensorView* proj_bias_bcast = broadcast(b0, {false, true, false});
+  TensorView* qkv1 = add(mm, proj_bias_bcast);
```
wujingyue, line 510: Style nit: you can save some typing by not having to give everything a name. E.g. 
```
  TensorView* mm0 = matmul(x, w0);
  TensorView* qkv = add(mm0, broadcast(b0, {false, true, false}));
  qkv = reshape(qkv, ...);
```

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -293,57 +446,167 @@ std::vector<TensorView*> mlp_backwards(
...
+  // Forming the q,k,v vectors:
+  auto D = w0->axis(0)->extent()->value().as<int64_t>();
```
wujingyue, line 512:
```suggestion
  const auto D = w0->axis(0)->extent()->value().as<int64_t>();
```

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -293,57 +446,167 @@ std::vector<TensorView*> mlp_backwards(
...
+  TensorView* qkv = reshape(qkv1, {D, B * S, 3 * E / D}, {D, B, S, 3 * E / D});
+  std::vector<TensorView*> qkv_reshaped = {};
```
wujingyue, line 514:
```suggestion
  std::vector<TensorView*> qkv_reshaped;
```

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -293,57 +446,167 @@ std::vector<TensorView*> mlp_backwards(
...
-      matmul0_grad_w,
-      matmul0_grad_b,
-      matmul0_grad_x};
+      linear0_grads.grad_w,
+      linear0_grads.grad_b,
+      linear0_grads.grad_x};
   return outputs;
 }
+
+std::vector<TensorView*> mha_backwards(
```
wujingyue, line 494: I remember you said this is different from https://github.com/NVIDIA/Fuser/blob/1158543c9bd16190fe5820b41a7f4c7e891bc5d9/benchmarks/python/test_transformer.py#L413 in which values got recomputed. Can you code-comment that for posterity?

tests/cpp/test_multidevice_transformer.cpp
wujingyue: I understood that -- that's what `grad_q` is for. But `grad` itself, the input of this function, is always float. Correct? If so, you can simply say `sum(grad)` somewhere below.

cowanmeg: No, it depends on the backwards operator that generates the incoming grad.
For MLP the incoming grad is float since it was generated from gelu.
For MHA the incoming grad is half since it was generated from sdpa.

wujingyue: Got it!

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -274,6 +368,65 @@ std::vector<TensorView*> mha(
...
+// we do not have logically split rfactor domain.
+struct LinearBackwardsResult {
+  TensorView* grad_x;
+  TensorView* grad_w;
+  TensorView* grad_b;
+};
+
+// x format: [i0, i1] dtype
+// weight format: [DID(D), i1, i2/D] dtype
+// grad format: [DID(D) i0, i2/D] float
```
wujingyue, line 381: Since `grad` can be lower precision, would you mind updating the comment?

tests/cpp/test_multidevice_transformer.cpp
cowanmeg: updated to linear1 which is more consistent with the nvFuser program. Thanks for pointing out the inconsistency! The naming is a bit all over the place... I'll clean up the other functions in a different PR since they also have this issue 😅

tests/cpp/test_multidevice_transformer.cpp
cowanmeg: That will be for the full backward test! Will make certain there is a comment for that.

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -169,14 +169,103 @@ std::vector<at::Tensor> reference_mlp_backwards(
...
+    at::Tensor mask,
+    at::Tensor w0,
+    at::Tensor b0,
+    at::Tensor w1) {
+  auto at_dtype = w0.dtype();
+  // recompute up to sdpa
+  auto linear0 = at::matmul(x, w0).add(b0).view({B, S, 3 * E});
+  auto qkv = linear0.split(E, /*dim=*/-1);
+  for (auto i = 0; i < 3; i++) {
+    qkv[i] = qkv[i].reshape({B, S, H, E / H}).transpose(2, 1).to(at_dtype);
```
wujingyue, line 184:
```suggestion
    qkv[i] = qkv[i].reshape({B, S, H, E / H}).transpose(1, 2).to(at_dtype);
```

tests/cpp/test_multidevice_transformer.cpp
wujingyue: I don't think this has been resolved yet. Even though `grad`'s dtype is not fixed, we can still simplify the code using maybeCastOp as mentioned in the first comment in this thread.

wujingyue (approved):

tests/cpp/test_multidevice_transformer.cpp
cowanmeg: sure i can change it to use two maybes instead...

PR #2828, opened 2024-08-21
Allow creating CPU scalar tensors using `fd.from_pytorch`

Priya2698:
This PR sets the `is_cpu` flag in `define_tensor` instead of raising an error when `fd.from_pytorch` is called for CPU scalar tensors.

### DIFF
--- a/tests/python/test_pointwise.py
+++ b/tests/python/test_pointwise.py
@@ -69,8 +69,10 @@ def test_issue_2395():
 
 # Tests that CPU scalar tensor can be instantiated using fd.from_pytorch
 def test_cpu_add():
-    inputs = [torch.tensor(2.0, device="cpu", dtype=torch.float),
-              torch.randn(3, device="cuda", dtype=torch.float)]
+    inputs = [
+        torch.tensor(2.0, device="cpu", dtype=torch.float),
+        torch.randn(3, device="cuda", dtype=torch.float),
+    ]
 
     def fusion_func(fd: FusionDefinition):
         t0 = fd.from_pytorch(inputs[0])
--- a/tests/python/test_pointwise.py
+++ b/tests/python/test_pointwise.py
@@ -69,15 +69,16 @@ def test_issue_2395():
 
 # Tests that CPU scalar tensor can be instantiated using fd.from_pytorch
 def test_cpu_add():
-    inputs = [torch.tensor(2.0, device="cpu", dtype=torch.float)]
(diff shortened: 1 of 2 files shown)


Priya2698:
!build

tests/python/test_pointwise.py
```diff
@@ -65,3 +65,19 @@ def create_fusion(fd: FusionDefinition) -> None:
...
+    def fusion_func(fd: FusionDefinition):
+        t0 = fd.from_pytorch(inputs[0])
+        s0 = fd.define_scalar(3.0)
+        t1 = fd.ops.add(t0, s0)
+        fd.add_output(t1)
+
+    with FusionDefinition() as fd:
+        fusion_func(fd)
+    nvf_out = fd.execute(inputs)
+    torch.testing.assert_close(nvf_out[0].cpu(), inputs[0] + 3.0)
```
jjsjann123, line 83: why do we need to call `cpu()` on nvf_out[0]. Are we producing output on cuda device by default? that doesn't feel right here.

Priya2698: Yes, it is a cuda tensor.
Through the codegen path, we do not generate CPU scalar tensors so I would expect this.

jjsjann123: sorry this falls off my radar earlier. Thanks for requesting review again. :bow: 

So this is supposed to be an error coming from nvfuser I think, since we are being wild on producing output Tensor on an arbitrary device. Can you open an issue with this test as a repro, when the PR is merged?

Meanwhile, for this test, I suggest that we modify the test as a scalar cpu tensor add a gpu non-scalar tensor. Which should produce an output on gpu as expected.

jjsjann123 (approved):
minor comment about the test. Code change looks good to me.

Priya2698:
!build

tests/python/test_pointwise.py
Priya2698: > So this is supposed to be an error coming from nvfuser I think, since we are being wild on producing output Tensor on an arbitrary device. Can you open an issue with this test as a repro, when the PR is merged?

Why should this be an error instead of on CPU? Maybe we should be using the expression evaluator here?

jjsjann123: Yeah, if we can generate a tensor on cpu that'll be great.

By throwing an error I was just referring to `we should throw an error if nvfuser generate an output tensor on gpu **instead** of cpu`

Priya2698:
!build

Priya2698:
!build

PR #2831, opened 2024-08-21
Add `fd.execute()` kwargs for generating a Repro Script

kevinstephano:
This PR is a cleanup and extension of what @riccardofelluga contributed by adding the `fd.getReproString()` and `fd.getReproErrorString()` APIs.

# Contributions
* Renamed Repro script functions to match `FusionDefinition` naming style.
  * `fd.getReproString()` -> `fd.repro_script_for()`
  * `fd.getReproErrorString` -> `fd._repro_error_str()` : This is not a public function because it is not intended to be externally used.
* Added the function `fd.last_repro_script()` which returns a string for the last set of inputs seen by the `fd.execute()` function.  
  * This is intended to be used in tandem with Thunder to print out repro scripts by setting a kwarg during a Thunder run. 
  * Requires setting the kwarg `fd.execute(inputs, save_repro_inputs=True)`.
* Added 2 `fd.execute()` kwargs
  * `fd.execute(inputs, print_repro=True)` will print a repro script to `stdout`.
  * `fd.execute(inputs, save_repro_inputs=True)` saves the inputs associated with last run.  The function `fd.last_repro_script()` needs to be called to get the resulting string.
* NOTE: `fd.last_repro_script()` can only be queried once after setting `fd.execute(inputs, save_repro_inputs=True)` such that the `FusionDefinition` does not indefinitely hold onto tensors. 

# Examples

## `FusionDefinition`

```python
inputs = [
    torch.randn(4, 4, device="cuda:0"),
    torch.randn(4, 4, device="cuda:0"),
]

    t1 = fd.from_pytorch(inputs[1])
    t2 = fd.ops.mul(t0, t1)
    fd.add_output(t2)
```
##  Kwarg: `print_repro=True` prints script to `stdout`

```python
fd.execute(inputs, print_repro=True)
```

## Kwarg: `save_repro_inputs=True` plus `last_repro_script()` query

```python
fd.execute(inputs, save_repro_inputs=True)
print(fd.last_repro_script())
```

## Expected Output
```
# CUDA devices:
#  0: NVIDIA A100 80GB PCIe
# torch version: 2.5.0a0+git25d5a81
# cuda version: 12.6
# nvfuser version: 0.2.10+gitb2e3f89
import torch

    T0 = fd.define_tensor(shape=[-1, -1], contiguity=[True, True], dtype=DataType.Float, is_cpu=False, stride_order=[1, 0])
    T1 = fd.define_tensor(shape=[-1, -1], contiguity=[True, True], dtype=DataType.Float, is_cpu=False, stride_order=[1, 0])
    T2 = fd.ops.mul(T0, T1)
    fd.add_output(T2)

    nvfuser_fusion_id0(fd)

inputs = [
    torch.randn(16, dtype=torch.float32, device='cuda:0').as_strided((4, 4), (4, 1)),
    torch.randn(16, dtype=torch.float32, device='cuda:0').as_strided((4, 4), (4, 1)),
]
fd.execute(inputs)
```


kevinstephano:
!build

nvfuser/__init__.py
```diff
@@ -378,6 +324,81 @@ def profile(self):
 
         return fp
 
+    def last_repro_script(self) -> str:
+        assert (
+            self.inputs is not None
+        ), "fd.last_repro_script() cannot provide a repro because fd.execute(inputs, save_repro_state=True) was not set!"
```
jacobhinkle, line 330: It might be good to include a note in the error message like "fd.last_repro_script() may only be called once after each call to fd.execute()".

nvfuser/__init__.py
```diff
@@ -216,6 +157,9 @@ def execute(
             self.schedule()
             self._finalize_schedule(inputs)
 
+        if save_repro_inputs:
+            self.inputs = inputs
```
jacobhinkle, line 161: Since we only require sizes, strides, device, and dtype, I wonder if we could extract that and keep it without needing to release it. This would be almost like a meta tensor, but it would also keep track of the device. In fact could we just use a utility like this:
```python
def convert_inputs_to_meta(inputs: list) -> list:
    return [(inp.to('meta'), inp.device)
            if isinstance(inp, torch.Tensor)
            else inp
            for inp in inputs]
```
Then we would just need to check for tuple inputs in line 353.

tests/python/test_python_frontend.py
```diff
@@ -4280,3 +4282,95 @@ def fusion_func(fd: FusionDefinition):
...
+        comp_str = test_stdout.getvalue()
+        comp_str = comp_str.strip()
+        comp_lines = comp_str.split("\n")
+
+        assert len(expected_lines) == len(
+            comp_lines
+        ), "Reproduction script does not have the expected number of lines! Expected: {} Actual: {}".format(
+            len(expected_lines), len(comp_lines)
+        )
+        # Don't compare the first 5 lines as they may be device and cuda specific
```
jacobhinkle, line 4347: Nit: could instead use `re.match` in order to handle the case where multiple devices are available, in which case there will be more than 5 comment lines.

jacobhinkle (approved):
Looks great. Just a couple suggestions.

tests/python/test_python_frontend.py
jjsjann123: further nitpick, might as well do ast.parse -> ast.dump and then just compare that. But that might be an overkill, and that still won't save us from variable renaming and stuff like that....

jjsjann123: I think I'm just saying that I like Jacob's suggestion above.

nvfuser/__init__.py
jjsjann123: I think we can manually do fake tensor here.

https://pytorch.org/docs/stable/torch.compiler_fake_tensor.html

```
>>> x = torch.randn(2, 3, 4, 5, device="cuda")

>>> from torch._subclasses.fake_tensor import FakeTensorMode
>>> fake_mode = FakeTensorMode()
>>> fake_x = fake_mode.from_tensor(x)
>>> del x

>>> fake_x
FakeTensor(..., device='cuda:0', size=(2, 3, 4, 5))
```

jjsjann123 (approved):
nitpick on holding onto tensor just for repro part, since that's just for printing a repro, feels like it's a tricky to get that a pleasant experience.

Stamping since everything else looks good to me.

nvfuser/__init__.py
jacobhinkle: Oh this is way better! TIL

nvfuser/__init__.py
kevinstephano: This is not true anymore as I used fake tensors as @jjsjann123  suggested.

nvfuser/__init__.py
kevinstephano: I implemented @jjsjann123 suggestion to use Fake Tensors to grab the inputs.

tests/python/test_python_frontend.py
kevinstephano: I fixed this by not including the starter header in the comparison by looking for comments.

kevinstephano:
!build

nvfuser/__init__.py
```diff
@@ -189,10 +126,14 @@ def execute(
                 must either tell NVFuser where to run the resulting kernel, or
                 let it default to 0. Note that passing this option providing
                 and input tensors that lie on another device is an error.
+            print_repro (bool): Prints a reproduction script to stdout.
+            profile (bool): Captures a CUPTI based profile of a fusion.
```
jjsjann123, line 130: nitpick: order here isn't right. The two entry should be below `capture_debug_output`

tests/python/test_python_frontend.py
```diff
@@ -4276,6 +4278,112 @@ def fusion_func(fd: FusionDefinition):
...
+        assert (
+            expected_lines[5:] == comp_lines[start_line:]
+        ), "Reproduction script does not match expected output!"
+
+        # Test fd.execute(inputs, save_repro_inputs=True) ; fd.last_repro_script()
```
jjsjann123, line 4359: nitpick: is this just test code?

tests/python/test_python_frontend.py
```diff
@@ -4276,6 +4278,112 @@ def fusion_func(fd: FusionDefinition):
...
+            fd.add_output(t2)
+
+        with FusionDefinition() as fd:
+            fusion_func(fd)
+
+        # Strip white space before and after script
+        expected_str = expected_str.strip()
+        expected_lines = expected_str.split("\n")
+
+        # Test fd.execute(inputs, print_repro=True)
```
jjsjann123, line 4325: ditto

jjsjann123 (approved):
another nitpick, can we also bump up the version.txt since this is an API change. :pray:

kevinstephano:
!build

kevinstephano:
!build

kevinstephano:
!build

kevinstephano:
The tests that are failing are Thunder tests that need to be version gated by version `2.14`.  The gating will happen with this [Thunder PR](https://github.com/Lightning-AI/lightning-thunder/pull/1231).

kevinstephano:
!build

kevinstephano:
!build

kevinstephano:
!build

PR #2835, opened 2024-08-22
Concretize resize

jacobhinkle:


jacobhinkle:
!build --diff-bench

jacobhinkle:
!build --diff

jacobhinkle:
I'm investigating the test failure in `test_host_ir --gtest_filter=HostIrTest.ForLoops/useFusionExecutor`.

jacobhinkle:
The latest pushed change fixes the broken test but note that this exposes a bug in `HostIrExecutor`'s concretization method when `params.use_fusion_executor_cache = false`. That bug is not specific to `Resize` ops: I think if we did a reshape based on the `ForLoop` index then we would see a similar error for example. I will post that as a separate issue soon.

jacobhinkle:
Note that at some point in the past week we have gone from 35 us on main to about 13 us for the repro in #2795. This PR still brings this down to 8 us. I haven't bisected yet to see where that speedup came from; I just noticed it this evening.

naoyam:
IIUC, this means that resized domains would be effectively treated as static? Most likely, the expansion factors of resize would depend on the actual extent of the resized domain, like in the RoPE case, so if the expansion factors were treated as static, the domain itself would be effectively static. Is my understanding correct?

csrc/mutator.cpp
```diff
@@ -183,22 +187,56 @@ Expr* OptOutMutator::mutateExpr(
...
   for (auto input : op->inputs()) {
-    mutated_inputs.emplace_back(replace_inputs ? maybeMutated(input) : input);
+    if (replace_inputs) {
+      Val* mut_inp = maybeMutated(input);
+      if (mut_inp != input &&
+          std::find_if(
+              op->outputs().begin(), op->outputs().end(), [&](Val* const outp) {
+                return (replace_outputs ? maybeMutated(outp) : outp) == mut_inp;
+              }) == op->outputs().end()) {
+        // skip using mutated input if is one of the inputs.
```
jjsjann123, line 214:
```suggestion
        // skip using mutated input if is one of the outputs.
```

csrc/mutator.cpp
```diff
@@ -183,22 +187,56 @@ Expr* OptOutMutator::mutateExpr(
...
+        Val* mut_attr = maybeMutated(attr_val);
+        if (mut_attr != attr_val &&
+            std::find_if(
+                op->outputs().begin(),
+                op->outputs().end(),
+                [&](Val* const outp) {
+                  return (replace_outputs ? maybeMutated(outp) : outp) ==
+                      mut_attr;
+                }) == op->outputs().end()) {
+          // skip using mutated input if is one of the inputs.
```
jjsjann123, line 235:
```suggestion
          // skip using mutated attr if is one of the outputs.
```

csrc/mutator.cpp
```diff
@@ -84,7 +84,8 @@ void OptOutMutator::mutate(Val* s) {}
 
 void OptOutMutator::mutate(NamedScalar* ns) {}
 
-void OptOutMutator::mutate(IterDomain* id) {
+void OptOutMutator::mutate(IterDomain* orig_id) {
```
jjsjann123, line 87: sorry I'm not following the code change here.

Reads like this is an orthogonal issue in mutator that's patched in this PR? My usual ask is, can we have an accompanying test / or a small segment in the PR description to help me understand it? :bow:

jjsjann123 (commented):
still looking at the actual transform.

jacobhinkle:
> IIUC, this means that resized domains would be effectively treated as static? Most likely, the expansion factors of resize would depend on the actual extent of the resized domain, like in the RoPE case, so if the expansion factors were treated as static, the domain itself would be effectively static. Is my understanding correct?
> 
> 


Yes this is correct, resized IterDomains are treated as if they were static with static expands.

csrc/mutator.cpp
jacobhinkle: Good point. I will add that.

jjsjann123:
> > IIUC, this means that resized domains would be effectively treated as static? Most likely, the expansion factors of resize would depend on the actual extent of the resized domain, like in the RoPE case, so if the expansion factors were treated as static, the domain itself would be effectively static. Is my understanding correct?
> 
> Yes this is correct, resized IterDomains are treated as if they were static with static expands.

nitpick:
Here we are using `sameAs` to check if a resize is needed in slice. Should/could we try to evaluate a equal operator between `range.stop` and `inp_root_size`?https://github.com/NVIDIA/Fuser/blob/13fba3b62c7b2230405c4b240b4bdaca6a363a1c/csrc/ops/alias.cpp#L786-L787

csrc/dynamic_transform.cpp
```diff
@@ -930,21 +954,134 @@ void DynamicTransformConcretizer::concretizeReshape() {
...
         id->isRFactorProduct(),
         iter_type);
 
-    registerConcretization(id, new_id);
+    registerConcretization(orig_id, new_id);
+
+    // Concretize the output shape which is constant
+    if (!orig_id->extent()->sameAs(new_id->extent())) {
+      registerConcretization(orig_id->extent(), new_id->extent());
+    }
```
jjsjann123, line 1084: Why are we not handling expanded extent here?

i.e. doesn't slice an expanded dimension preserve the expanded field here, which means we need to register that in concretization as well?

jjsjann123 (commented):
Looks neat!

Please ping me again when mutator piece is good for another round of review.

csrc/dynamic_transform.cpp
jacobhinkle: It's a good point that we should handle expanded extents on the _inputs_. Ideally we would need to handle output expanded extent so I will update the PR. However, note that currently I believe if you resize an expanded ID, the result is `Iteration`. This is because we do not check that the pad values are negative, so we might be doing an actual padding of that dimension. In case the pad values are constant we could do that check and sometimes preserve expanded broadcasts in slice.

csrc/dynamic_transform.cpp
```diff
@@ -396,27 +403,40 @@ void DynamicTransformConcretizationInfo::analyzeResizes(
...
+      // If this option is disabled, it means we will only concretize IterType,
+      // not the extents of Resize ops. In such case we standardize the
+      // concretization info so that the resize_extents_ entries are equivalent
+      // to saving IterType. This is so that we avoid unnecessary cache misses
+      // in cases where the IterType does not change but the specific extents or
+      // expand values do change.
+      resize_extents_.emplace_back(
+          id_index,
+          ConcreteResize{
+              2, input_extent + left_expand + right_expand == 1 ? -1 : 1, 0});
```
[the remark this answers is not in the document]
jacobhinkle: Good point. I think instead an input extent of -1 should be used, which will never conflict with an actual case. I can then use -1,1,1 to represent Broadcast and -1, 2, 2 to represent Iteration.

tests/cpp/test_dynamic_transform.cpp
```diff
@@ -931,6 +931,10 @@ void reductionDynamicPadAddFusion(
     auto pad_widths = std::get<1>(inv);
     auto expect_miss = std::get<2>(inv);
 
+    // Since we concretize the inputs and expands for each Resize, we always
+    // expect a cache miss unless we have re-used the exact same values.
+    expect_miss = true;
```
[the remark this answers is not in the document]
jacobhinkle: I switched this PR to opt-in, so I am reverting the changes to this test.

PR #2836, opened 2024-08-22
Sharding propagation utility functions

cowanmeg:
Add a utility function `shardBetween` that shards all tensorviews between reference tensorviews and boundary tensorview like the reference(s), so that we can control shardingPropagation boundaries. 

Updates Transformer tests to disable the current sharding propagation pass and manually control sharding propagation boundaries with `shardBetween`. This let's us use the minimal number of manual sharding annotations (inputs, outputs, and reshardings).


cowanmeg:
fyi, found an issue with forward transformer block where shardings weren't propagated correctly, so moving to draft until I fix it.

csrc/multidevice/utils.cpp
```diff
@@ -227,6 +227,19 @@ void shardAllLike(TensorView* ref, std::vector<TensorView*> tvs) {
...
+void shardBetween(
+    std::vector<TensorView*> ref_tvs,
+    const std::unordered_set<TensorView*>& boundary_tvs) {
+  auto tvs_between = scheduler_utils::getAllTvsFrom(ref_tvs, boundary_tvs);
+  std::vector<TensorView*> tvs_to_shard;
+  std::copy_if(
+      tvs_between.begin(),
+      tvs_between.end(),
+      std::back_inserter(tvs_to_shard),
+      [](TensorView* tv) { return !tv->hasDeviceMesh(); });
```
wujingyue, line 239: Use `std::remove_if` and `erase`.

wujingyue, line 239: Use `std::mem_fn` which is more direct. Example: https://github.com/NVIDIA/Fuser/blob/58aa9759ffbe322fa3b9425859a44be27876d8c9/csrc/preseg_passes/propagate_shardings.cpp#L25

csrc/multidevice/utils.h
```diff
@@ -86,8 +86,15 @@ bool haveDifferentShardings(
 // Returns whether a resharding expr reshards an inner axis
 bool isInnerResharding(Expr* expr);
 
+// Shards all tensors in tvs like reference
 void shardAllLike(TensorView* ref, std::vector<TensorView*> tvs);
 
+// Shards all tensors within a boundary like ref_tvs, but does not change the
+// sharding of tvs already sharded.
+void shardBetween(
+    std::vector<TensorView*> ref_tvs,
```
wujingyue, line 95: The behavior that it only picks `ref_tvs[0]` as the reference tensor is worth calling out. (I understand other tensors are used as the starting points of the traversal). 

Alternatively, we can make it `shardBetween(starts, ends, ref_tv)`. It's slightly more typing but less surprising.

csrc/multidevice/utils.cpp
```diff
@@ -227,6 +227,19 @@ void shardAllLike(TensorView* ref, std::vector<TensorView*> tvs) {
   }
 }
 
+void shardBetween(
+    std::vector<TensorView*> ref_tvs,
```
wujingyue, line 231:
```suggestion
    const std::vector<TensorView*>& ref_tvs,
```
In most cases, this saves a copy or is as fast.

csrc/multidevice/utils.h
cowanmeg: I wanted to add a check that all the tensors in ref_tv are sharded consistently.  

I think IdModel could be used for this? We can check if a there is a DID iterdomain then they should all be mapped.

cowanmeg:
Note: don't review since I'm just committing an update that needs to be cleaned up

csrc/multidevice/utils.cpp
```diff
@@ -177,6 +177,21 @@ bool isResharding(const Expr* expr) {
     return false;
   }
 
+  // Reduction over a sharded dimension.
```
cowanmeg, line 180: The sharding propagation pass will propagate DIDx through an AllReduce
DIDx(i0), i1 -> r(DIDx(i0), i1. Since `isResharding` is used to detect when an expr is lowered to a collective, the AllReduce would not get identified without explicitly checking for a reduction of a DID axis.

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -395,6 +374,8 @@ TEST_P(DistributedTransformerTest, MLP_Layer) {
   for (TensorView* tv : tvsout) {
     fusion->addOutput(tv);
   }
+  shardBetween({tvw0, tvb0, tvw1}, {tvsout[3]}, tvw0);
+  shardBetween({tvx, tvb1}, {tvsout[3]}, tvx);
```
cowanmeg, line 378: Is there a simple way to assert that the final fusion contains only 1 allreduce? I can see from HostIR dump that 1 allreduce as expected is there, but not sure how I can get access this info easily.

csrc/multidevice/utils.cpp
```diff
@@ -177,6 +177,21 @@ bool isResharding(const Expr* expr) {
...
+  if (expr->isA<ReductionOp>()) {
+    // auto in = expr->as<ReductionOp>()->in()->as<TensorView>();
+    auto out = expr->as<ReductionOp>()->out()->as<TensorView>();
```
wujingyue, line 183:
```suggestion
    auto out* = expr->as<ReductionOp>()->out()->as<TensorView>();
```

csrc/multidevice/utils.cpp
```diff
@@ -177,6 +177,21 @@ bool isResharding(const Expr* expr) {
     return false;
   }
 
+  // Reduction over a sharded dimension.
+  if (expr->isA<ReductionOp>()) {
+    // auto in = expr->as<ReductionOp>()->in()->as<TensorView>();
```
wujingyue, line 182:
```suggestion
```

csrc/multidevice/utils.cpp
```diff
@@ -177,6 +177,21 @@ bool isResharding(const Expr* expr) {
...
+    auto out = expr->as<ReductionOp>()->out()->as<TensorView>();
+    std::vector<IterDomain*> reduction_axis;
+    std::copy_if(
+        out->getLogicalDomain().begin(),
+        out->getLogicalDomain().end(),
+        std::back_inserter(reduction_axis),
+        [](IterDomain* id) { return id->isReduction(); });
+    if (reduction_axis[0]->isDeviceDim()) {
+      return true;
+    }
```
wujingyue, line 192:
```suggestion
    if (std::any_of(out_logical.begin(), out_logical.end(), std::mem_fn(IterDomain::isReduction)) {
      return true;
    }
```
It's not doing exactly the same thing but I think it's simpler and more general -- it seems unnecessary to only check whether the first reduction axis is DID.

wujingyue (commented):
Still reviewing the rest...

csrc/multidevice/utils.cpp
wujingyue: cc @naoyam 

This is related to our broadcast/DID discussion last week. I've yet to understand the difference between `r(DIDx{i0})` and `r(i0)`. 
1. Are they semantically equivalent? 
2. If so, should nvFuser support both or only a canonical form (can be either of them)?

naoyam: First of all, one caveat is that I remember there's a slight semantic difference of interpreting parallel types between DID and BID/TID. I don't exactly remember what it is, though. Let me check, but @samnordmann, I remember I told you. Do you remember?

So, suppose we have `r(TIDx{i0})` and `r(i0)`. They are different that:

- Only the former instructs nvFuser to parallelize the reduction. In this case, the `blockReduce` device function defined in one of the runtime files will be used.
- The latter indicates serial execution. Semantically, it should be valid if only one of the threads, e.g., tid==0, runs the reduction. I think by default we just let all threads run the reduction redundantly. There's a predicate generation pass to limit the redundant execution, so that may affect.
- In the latter case, since it's executed by just a single thread, all the elements must be accessible by the thread. If the input is parallelized by TIDx, it means the input must be on shared memory. A syncthreads will be automatically inserted.

BID is also the same.

cowanmeg: We've had many conversation around this.

`DIDx(i0) -> ri0` vs. `DIDx(i0) -> r(DIDx(i0))`
Since we don't codegen the collectives, we can't control how the parallelization occurs, so the two would both get lowered to an (all)reduce collective. In a sense the DID parallelization is only conveying the state of the axis.


ReduceScatter is not represented well with `r(DIDx(i0))` unfortunately... 
Currently we look for a pattern like this: `[DIDx(i0), i1] -> [r(i0), DIDx(i1)]`

If we use `r(DIDx(i0))` then we get: `[DIDx(i0), i1] -> [r(DIDx(i0)), DIDx(i1)]`
We have 2 DIDx axis, which is not allowed, but in this case the `r(DIDx(i0))` is only conveying the reduction happened on a sharded axis.


There was another discussion with @naoyam where we discussed that ideally we should represent reduce scatter as two primitives (reduce + scatter), but since it would complicate the communication lowering pass we did not do this.

csrc/multidevice/utils.h
```diff
@@ -86,7 +86,30 @@ bool haveDifferentShardings(
...
+        ParallelType::Serial,
+        ParallelType::DIDx});
+
+// Shards all TVs from the output of from expressions to the outputs of the to
+// expressions like the reference TV.
+void shardBetween(
+    const std::vector<Expr*>& from,
+    const std::vector<Expr*>& to,
+    TensorView* ref,
+    const std::unordered_set<ParallelType>& parallel_types = {
```
wujingyue, line 103: Do you plan to customize this? It's always DIDx currently, and I'd expect this to be all DIDs for the future.

csrc/multidevice/utils.h
```diff
@@ -86,7 +86,30 @@ bool haveDifferentShardings(
 // Returns whether a resharding expr reshards an inner axis
 bool isInnerResharding(Expr* expr);
 
-void shardAllLike(TensorView* ref, std::vector<TensorView*> tvs);
+// Shards all tensors in tvs like reference
+void shardAllLike(
+    TensorView* ref,
+    std::vector<TensorView*> tvs,
+    const std::unordered_set<ParallelType>& parallel_types = {
+        ParallelType::Serial,
```
wujingyue, line 94: Why Serial? IIRC, everything starts with Serial and parallelAllLike will make some of them DIDx.

tests/cpp/test_multidevice_transformer.cpp
wujingyue: https://github.com/NVIDIA/Fuser/blob/cf141ed738b77b08e7c3bfe5e8b1311873c725e5/tests/cpp/test_multidevice_lower_communication.cpp#L25 is the simplest way I can think of at this moment.

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -371,6 +348,8 @@ std::vector<TensorView*> mlp_backwards(
 } // namespace
 
 TEST_P(DistributedTransformerTest, MLP_Layer) {
+  preseg_passes::OptimizationPassGuard<preseg_passes::PropagateShardingsPass>
```
wujingyue, line 351: Why? We'd better test the real behavior without turning any knobs. I guess in this case PropagateShardingPass failed to process MatmulOp? If so, can you file a bug and `TODO(bug number)`?

csrc/multidevice/utils.h
cowanmeg: For now it's just DIDx, but later we might want one TV to be the ref for DIDx and another for DIDy. 
Since it's not necessary now, I'm open to removing it and revisiting later.

csrc/multidevice/utils.h
cowanmeg: I recall we added serial for cases in the reorder sharding axis pass. It might be good to revisit that and see if we should be using shardAllLike in that case.

tests/cpp/test_multidevice_transformer.cpp
cowanmeg: Since the manual sharding propagations covered all the tensors, it's actually fine to let the current pass run since it will do nothing. I left it more as a sanity check that the manual propagations were doing what I thought.

csrc/multidevice/utils.cpp
wujingyue: Thank you both! 

> Since we don't codegen the collectives, we can't control how the parallelization occurs, so the two would both get lowered to an (all)reduce collective

That's a great summary. 

Back to my earlier question: 
1. `DIDx(i0) -> ri0` and `DIDx(i0) -> r(DIDx(i0))` are indeed semantically equivalent, although not the case for TID/BID which affects codegen. 
2. For the time being, we need the former to avoid two `DIDx`s for ReduceScatter. Also, [newForReduction](https://github.com/NVIDIA/Fuser/blob/2edebb60130b6f17c5b328279b41d35c15a21fbd/csrc/ops/arith.cpp#L1197-L1203), which I changed weeks ago, doesn't propagate DIDs to reduction dimensions and therefore implements the former. (It's easy to change that either way.) So should we stick with only the former (for other ops) or should we support both?

tests/cpp/test_multidevice_transformer.cpp
wujingyue: Got it! I'd let the pass run because it's closer to the real integration. The user (likely Thunder) is expected to annotate shardings but not expected to turn off individual passes.

csrc/multidevice/utils.cpp
cowanmeg: Hmm.. I don't have strong opinions, but maybe we should stick to the former to stay consistent?

csrc/multidevice/utils.h
wujingyue: Got it. I'd go YAGNI if (I think so) it leads to less code.

csrc/multidevice/utils.cpp
naoyam: > DIDx(i0) -> ri0 vs. DIDx(i0) -> r(DIDx(i0))
> Since we don't codegen the collectives, we can't control how the parallelization occurs, so the two would both 
> get lowered to an (all)reduce collective. In a sense the DID parallelization is only conveying the state of the axis.

> DIDx(i0) -> ri0 and DIDx(i0) -> r(DIDx(i0)) are indeed semantically equivalent

I don't follow what you meant. In the former case, the output is not parallelized, so it indicates the operation does not need to be parallelized. How they are interpreted has some flexibility within the semantics, so they could result in the same execution, but that doesn't mean there no semantic difference. 

If the output is not parallelized, all of the below cases should be valid:

1. All of the GPUs execute `allreduce` collectively. All GPUs will have the valid reduction result. This is what we do now with DID.
2. All of the GPUs execute `reduce` collectively. Only one of the GPUs will have the valid result. Can be much more efficient than case 1.
3. It is only executed by one of the GPUs. The rest do nothing. This is not possible with distributed tensors.

In TID/BID, when the output is not parallelized, it is interpreted as case 3. If the input is TID parallelized, it must be placed on the shared memory or global memory. When it's parallelized, it's interpreted as case 2.

The case 1 behavior is actually `reduction + broadcast`. The reduction removes the reduced iter domain, so it no longer indicates nothing about the sharding, so a reduction itself does not indicate the behavior like case 1 but needs a broadcast. For example,

```
// t0: [DID(i0)]
t1 = sum(t0, {0})
t1->axis(0)->parallelize(DID)
// t1: [DID(r0)]
t2 = set(t1)
// t2: [] 
```

`t2` is a scalar tensor. It doesn't need to be sharded over DID.  The above fusion only instructs a reduction to a device. That means `MPI_Reduce` is enough.

`MPI_Allreduce` corresponds to below:

```
// t0: [DID(i0)]
t1 = sum(t0, {0})
t1->axis(0)->parallelize(DID)
// t1: [DID(r0)]
t2 = broadcast(t1, {true})
t2->axis(0)->parallelize(DID)
// t2: [DID(b0)] 
```

This pattern appears in normalization kernels, for which we have a lowering [pass](https://github.com/NVIDIA/Fuser/blob/main/csrc/device_lower/analysis/fused_reduction.cpp) that converts `reduce + broadcast` to `allreduce`. 

To be clear, I'm not saying the above behavior is absolutely right, but it's the behavior of the TID/BID parallelization and all of the existing analyses enabling the two parallel types have been built upon the behavior. For prototyping, I don't think we should care much, however, what concerns me most is, when integrating DID more deeply, these semantic differences might bite us in an unexpected, hard-to-understand way.

cowanmeg: > case 2: All of the GPUs execute reduce collectively. Only one of the GPUs will have the valid result. Can be much more efficient than case 1.

We also need to remember the device mesh is involved which I left out since we are mostly dealing with same device mesh. So in this example if the device mesh between the input and output is the same, then we all the GPUs must have the final result which would make case 2. If the output mesh is a single device (e.g. something like mesh moves from {0,1,2} -> {0} then we can use a single Reduce.

naoyam: >  then we all the GPUs must have the final result 

Why?

wujingyue: > > then we all the GPUs must have the final result

> Why?

Because otherwise nvFuser wouldn't know which GPU holds the result. 

Below is my understanding of the semantics of a **TensorView**, which agrees with @cowanmeg . 
1. `r(DID{i0}), mesh={0}` means the reduction result is only valid on GPU 0. 
2. `r(i0), mesh={0}` means the same as 1. 
3. `r(DID{i0}), mesh={0,1,2,3}` means the reduction result is replicated on all four GPUs. Assume `i0=4` for now. 
4. `r(i0), mesh={0,1,2,3}` means the same as 3. 

That being said, I think you are raising a good question of whether an IterDomain is parallelized on DID may affect the semantics of the **op**. For example, we can use `DID{i0} -> r(i0)` to represent an `Allreduce`, and `DID{i0} -> r(DID{i0})` a `Reduce+Broadcast`. If so, nvFuser will have to support both. However, there's also the approach to represent `Reduce+Broadcast` as 
```
DID{i0}, mesh={0,1,2,3} -> [reduction] r(i0), mesh={0}
r(i0), mesh={0} -> [set] -> (the IterDomain no longer exists in the output), mesh={0,1,2,3}
```
which I believe nvFuser already supports today.

naoyam: > Because otherwise nvFuser wouldn't know which GPU holds the result.

Hmm, in the case of TID/BID, it's just tid==0 or bid==0. In the case of DID, since we specify the head device, doesn't we know where it is?

wujingyue: > head device

I haven't heard about that notion, but we certainly could assume the first device in the mesh is the head device. 

Let's talk about this tomorrow? I believe the semantics you proposed can work. If we go with that, code that currently generates `b(DIDx{i0})` or `r(DIDx{i0})` will need to be changed to generate `b(i0)` or `r(i0)` instead.

wujingyue: (In case it wasn't clear, this discussion has no intention to block the PR.)

naoyam: Happy to discuss tomorrow. For the record, the head device refers to the `root` in `ncclReduce`, for example. https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/api/colls.html#ncclreduce

cowanmeg: Even if we arbitrarily made the first device the root...

t0[DIDx(i0), i1] -> t1[r(i0), i1] -> t2[i1] -> t3[i1] all with mesh - {0,1,2}
t1 would only be reduced on gpu 0. How would t3 know if it is valid on the entire mesh or just gpu 0 without traversing back up to the root tensor looking for all reductions and seeing if it was an allreduce or reduce?

cowanmeg: also back to the PR...just to make sure we are on the same page for short term:
DIDx(i0) mesh -> r(i0) mesh --> same mesh lowered to AllReduce. 
DIDx(i0) mesh0 -> r(i0) mesh1 --> different mesh lowered to Reduce(s) where mesh1 makes up the roots

This means sharding propagation will not propagate DIDx through a reduction.

wujingyue: > also back to the PR...just to make sure we are on the same page for short term:
DIDx(i0) mesh -> r(i0) mesh --> same mesh lowered to AllReduce.
DIDx(i0) mesh0 -> r(i0) mesh1 --> different mesh lowered to Reduce(s) where mesh1 makes up the roots

Yes. I believe that also matches our current implementation of communication lowering. 

> This means sharding propagation will not propagate DIDx through a reduction.

That seems safer -- it won't run into the arguable case of `r(DID{i0})`. 

As an incidental note, https://github.com/NVIDIA/Fuser/blob/13fba3b62c7b2230405c4b240b4bdaca6a363a1c/tests/cpp/test_multidevice_matmul.cpp#L87 generates `b(DID{i0})`, which is as arguable.

cowanmeg: haha yes, I actually remember specifically doing that to prevent an resharding expression from getting added (in this case it would insert an unecessary allgather)

csrc/multidevice/utils.cpp
```diff
@@ -234,41 +223,62 @@ void shardAllLike(TensorView* ref, std::vector<TensorView*> tvs) {
...
+    TensorView* ref) {
+  auto tvs_between = scheduler_utils::getAllTvsFrom(from_tvs, boundary_tvs);
+  // std::vector<TensorView*> tvs;
+  // std::copy_if(tvs_between.begin(), tvs_between.end(),
+  // std::back_inserter(tvs), [](TensorView* tv) {return !isDIDReduction(tv);});
+  shardAllLike(ref, {tvs_between.begin(), tvs_between.end()});
+
+  // Remove DID parallelizations on reduction axes.
+  for (auto tv : tvs_between) {
+    for (auto id : tv->getLogicalDomain()) {
```
wujingyue, line 243:
```suggestion
    for (IterDomain* id : tv->domain()->getAllIDs()) {
```
so it's still correct when we split loop instead.

csrc/multidevice/utils.cpp
```diff
@@ -234,41 +223,62 @@ void shardAllLike(TensorView* ref, std::vector<TensorView*> tvs) {
...
+    const std::unordered_set<TensorView*>& boundary_tvs,
+    TensorView* ref) {
+  auto tvs_between = scheduler_utils::getAllTvsFrom(from_tvs, boundary_tvs);
+  // std::vector<TensorView*> tvs;
+  // std::copy_if(tvs_between.begin(), tvs_between.end(),
+  // std::back_inserter(tvs), [](TensorView* tv) {return !isDIDReduction(tv);});
+  shardAllLike(ref, {tvs_between.begin(), tvs_between.end()});
+
+  // Remove DID parallelizations on reduction axes.
+  for (auto tv : tvs_between) {
```
wujingyue, line 242:
```suggestion
  for (auto* tv : tvs_between) {
```

csrc/multidevice/utils.cpp
```diff
@@ -234,41 +223,62 @@ void shardAllLike(TensorView* ref, std::vector<TensorView*> tvs) {
...
+        ref, tvs, {ParallelType::DIDx, ParallelType::Serial});
+  }
+}
+
+namespace {
+void propagateShardings(
+    const std::vector<TensorView*>& from_tvs,
+    const std::unordered_set<TensorView*>& boundary_tvs,
+    TensorView* ref) {
+  auto tvs_between = scheduler_utils::getAllTvsFrom(from_tvs, boundary_tvs);
```
wujingyue, line 235:
```suggestion
  std::unordered_set<TensorView*> tvs_between = scheduler_utils::getAllTvsFrom(from_tvs, boundary_tvs);
```

csrc/multidevice/utils.cpp
```diff
@@ -234,41 +223,62 @@ void shardAllLike(TensorView* ref, std::vector<TensorView*> tvs) {
...
+  // std::vector<TensorView*> tvs;
+  // std::copy_if(tvs_between.begin(), tvs_between.end(),
+  // std::back_inserter(tvs), [](TensorView* tv) {return !isDIDReduction(tv);});
```
wujingyue, line 238:
```suggestion
```

csrc/multidevice/utils.h
```diff
@@ -86,8 +86,22 @@ bool haveDifferentShardings(
...
+// expressions like the reference TV.
+void shardBetween(
+    const std::vector<Expr*>& from,
+    const std::vector<Expr*>& to,
+    TensorView* ref);
+
+// Shards all TVs between from_tvs and to_tvs like the reference TV.
+void shardBetween(
+    const std::vector<TensorView*>& from_tvs,
+    const std::vector<TensorView*>& to_tvs,
```
wujingyue, line 102:
```suggestion
    const std::vector<TensorView*>& from,
    const std::vector<TensorView*>& to,
```
The type is already clear that these arguments must be TVs.

csrc/multidevice/utils.h
```diff
@@ -86,8 +86,22 @@ bool haveDifferentShardings(
...
 void shardAllLike(TensorView* ref, std::vector<TensorView*> tvs);
 
+// Shards all TVs from the output of from expressions to the outputs of the to
+// expressions like the reference TV.
+void shardBetween(
+    const std::vector<Expr*>& from,
+    const std::vector<Expr*>& to,
+    TensorView* ref);
+
+// Shards all TVs between from_tvs and to_tvs like the reference TV.
```
wujingyue, line 99: "between" isn't accurate -- as you said in the .cpp comment, the function processes a superset of the between expressions. I think it's worth calling out the behavior in the .h comment instead of .cpp.

csrc/multidevice/utils.cpp
```diff
@@ -227,6 +227,83 @@ void shardAllLike(TensorView* ref, std::vector<TensorView*> tvs) {
...
+    }
+  }
+}
+} // namespace
+
+void shardBetween(
+    const std::vector<Expr*>& from,
+    const std::vector<Expr*>& to,
+    TensorView* ref) {
+  std::vector<TensorView*> from_tvs;
```
wujingyue, line 256: I think this can be implemented as
```
from_tvs = output tensorviews of from;
to_tvs = output tensorviews of to;
shardBetween(from_tvs, to_tvs, ref);
```
This way, it's simpler and leaves the slightly error-prone logic of grabbing the input tensorviews of `from` only in the other helper.

wujingyue (commented):
LGTM overall

csrc/multidevice/utils.cpp
cowanmeg: `getAllTvsFrom` adds all reachable TVs, so we have to add the inputs of `from` exprs into the boundary.

csrc/multidevice/utils.h
cowanmeg: how about `shardFrom`?

wujingyue: I think that's a better name. However, even with that, shardFrom makes me think it's similar to getAllTvsFrom while it's not. In getAllTvsFrom, cutoff_tv_set needs to surround from_tvs, and shardFrom's `to` doesn't need to. I think that behavior is still worth calling out in the header comment.

csrc/multidevice/utils.cpp
wujingyue: I understood that, and that's why I suggested call the TV version of `shardBetween` and not `propagateShardings` directly. Am I missing something?

cowanmeg: Sorry, I got confused with the different names and thought you proposed removing the inputs of `from`, you're suggestion should work

csrc/multidevice/utils.h
cowanmeg: How about `propagateShardings` then? With the updated comment of the scope of TVs that get affected, this will clear up some confusion with `getTvsBetween` and `getAllTvsFrom` behavior?

wujingyue: Do you mean renaming this function to `propagateShardings`? That makes sense. I just don't know how to deal with the existing `propagateShardings` :) 

Sorry, I don't want to bikeshed too much on naming. I wasn't opposed to `shardBetween` -- it is probably still the closest to the actual behavior. It was just the fact that the result TVs are not strictly "between" `from` and `to` that made me think some header comments are worthwhile.

cowanmeg:
!build

cowanmeg:
!build

csrc/multidevice/utils.h
cowanmeg: sure I changed it back to `shardBetween` to avoid confusion with the propagateSharding pass. Comment has been updated as well.

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -600,6 +587,10 @@ TEST_P(DistributedTransformerTest, Forward) {
     tv->setDeviceMesh(mesh);
   }
 
+  shardBetween({mha_in->definition()}, {mha_out->definition()}, mha_w0);
+  shardBetween({mlp_in->definition()}, {mlp_out->definition()}, mlp_w0);
```
wujingyue, line 591:
```suggestion
  shardBetween({mha_in}, {mha_out}, mha_w0);
  shardBetween({mlp_in}, {mlp_out}, mlp_w0);
```
is this any different? If it's indeed equivalent, we could simply remove the Expr-version of shardBetween.

wujingyue (approved):

tests/cpp/test_multidevice_transformer.cpp
cowanmeg: There is an equivalent version, it's just sometimes easier to think of it through expression boundaries. I can remove if that helps merge the PR.

wujingyue: > it's just sometimes easier to think of it through expression boundaries. 

Why? Because the TensorView version has a weird from-is-inclusive-but-to-is-exclusive semantics? 

> I can remove if that helps merge the PR.

Up to you -- you've been on the front line of this and it helps to make yourself comfortable!

cowanmeg: It makes it slightly easier to get boundaries correct. For example, if you have any expression with multiple inputs like c = matmul(a, b) and b is part of the `to` tensorviews but not a, then the propagation will visit a's and its producers, which may or may not be expected. 

Using the expression version helps to mimic the between behavior, though it doesn't help with non-linear regions like residuals.

PR #2837, opened 2024-08-23
Remove truediv operation

rdspring1:
This PR removes the `truediv` operation, which mapped to `torch.true_divide` and `torch.div(a, b, rounding_mode='none')`. 

Why?
- It is difficult to distinguish `truediv` from `div` in the Fusion IR.
- Thunder `prim.div` maps to the `div` operator. The other types of `div` are implement through this prim.

### DIFF
--- a/csrc/python_frontend/python_bindings.cpp
+++ b/csrc/python_frontend/python_bindings.cpp
@@ -2056,14 +2056,19 @@ void initNvFuserPythonBindings(PyObject* module) {
       "__lshift__", "bitwise_left_shift", bitwise_left_shift)
   NVFUSER_PYTHON_BINDING_BINARY_OP_SPECIAL(
       "__rshift__", "bitwise_right_shift", bitwise_right_shift)
-  // In PyTorch, __div__ (//) and __truediv__ (/) are different.
-  // When applied to integer-dtype arguments, they do as expected, returning
-  // integer and float outputs, respectively. When applied to two floating-type
-  // arguments, they return the floor of division for // and plain division for
-  // /. When applied to mixed types, the types are promoted, so the
-  // floating-point behavior is returned.
-  // Our div operator matches the __truediv__ behavior, so we do not implement
-  // __div__.
+  // In python, __truediv__ (/) always returns a float regardless of whether
+  // the input arguments are float or integer. __truediv__ (/) corresponds with
+  // pytorch torch.true_divide(a, b). The __div__ operator is depreciated in
+  // python 3.
+  //
+  // In nvfuser, truediv function has the same semantics as python's
+  // operator __truediv__ (/). The div function truncates the result instead
+  // of promoting it to float, so it has the same semantics as the C++'s (/)
+  // operator. In pytorch, torch.div(a, b, rounding_mode='trunc') corresponds
+  // C-style integer division.
+  //
+  // Hence, in the python frontend, the __truediv__ (/) python operator maps to
+  // trunc division.
   NVFUSER_PYTHON_BINDING_BINARY_OP_SPECIAL("__truediv__", "div", div)
 #undef NVFUSER_PYTHON_BINDING_BINARY_OP_SPECIAL
 


rdspring1:
!build

csrc/ir/nodes.cpp
```diff
@@ -593,6 +593,7 @@ std::vector<PolymorphicValue> BinaryOp::evaluate(
       return {lhs * rhs};
       break;
     case BinaryOpType::Div:
+    case BinaryOpType::Truediv:
```
zasdfgbnm, line 596: I don't think this will work. Isn't `2 truediv 5 = 0.4`, instead of `0`?

rdspring1: Can we do this?
```cpp
    case BinaryOpType::Truediv:
      return {
          SimplifyingIrBuilder::maybeCastExpr(DataType::Float, lhs) /
          SimplifyingIrBuilder::maybeCastExpr(DataType::Float, rhs)};
      break;
````

Comment (rdspring1) [reply csrc/ir/nodes.cpp:596]: IIRC, this is what we do in the `truediv` composite op.

**Comment by jacobhinkle** [comment]
What exactly is the semantics of our `div` meant to be? For ints it is truncdiv (C++ int division). If either arg is float, then it is truediv. My point is that we do not actually implement python `div` (floordiv), so what is the value in separately supporting truediv?

**Comment by rdspring1** [comment]
```
// Like torch.div with rounding_mode=None
NVFUSER_DEFINE_BINARY_FLOAT_OP(truediv, Div)
BinaryOpType::op_type, v1, v2, TypePromotion::float_op_config);
```

Are you proposing to just remove `truediv`?

**Comment by jacobhinkle** [comment]
> Are you proposing to just remove `truediv`?

I'm not sure what I'm proposing. You're right that in the kernel our `Div` will cast int inputs to float, but that's not true in our evaluator. That's probably just a bug in the evaluator, but I think that behavior is also used in scalar arithmetic, e.g. in #2818.

I guess for ultimate clarity we should have all these separate nodes/ops: truediv, truncdiv, floordiv, and ceildiv and nothing just called `div`.

**Comment by rdspring1** [comment]
!build

File: csrc/ir/nodes.cpp (hunk 1 above)
Comment (rdspring1) [reply csrc/ir/nodes.cpp]: I just removed the changes to `evaluate`.

**Comment by rdspring1** [comment]
I believe `truediv` isn't used by Thunder, so maybe we can remove it for clarity. For now, I removed `truediv` from expression evaluator.

**Review by jjsjann123 — COMMENTED** [review]
> This is used for translating CPP Fusion to python FusionDefinition

What's the usage here? Can you add an example to show that?

I'm also not sure if we actually do the right behavior here. i.e. we promote Tensors to floating point before div, but do we do the same for scalars? I share the concern @jacobhinkle has in his comment. But I don't think we have to resolve everything in this PR. Hence the ask of an example of what the problem we want to patch.

BTW, our python binding is binding `__truediv__` to `div`. Might want to change that as well, since I think that's also relevant here?

The division handling in nvfuser is a concerning question. We should open and track that with another issue. Does that sound good to you @jacobhinkle ?

**Comment by jacobhinkle** [comment]
> Does that sound good to you @jacobhinkle ?

:100:

**Comment by rdspring1** [comment]
> What's the usage here? Can you add an example to show that?

It is this series of PRs.
- https://github.com/NVIDIA/Fuser/pull/2841
- https://github.com/NVIDIA/Fuser/pull/2844

**Comment by rdspring1** [comment]
@jjsjann123 Why not get rid of `truediv`? It isn't used in Thunder.

**Comment by rdspring1** [comment]
Do we need to support `truediv`, `truncdiv`, `floordiv`? Thunder already handles this.

**Comment by jjsjann123** [comment]
> What exactly is the semantics of our `div` meant to be? For ints it is truncdiv (C++ int division). If either arg is float, then it is truediv. My point is that we do not actually implement python `div` (floordiv), so what is the value in separately supporting truediv?

Looks like thunder prim is only has c++ style `div` and `py_floordiv` (which we are not claiming anyway).
I don't think it makes sense to have others in API, since we do not have correct support at this moment.

**Comment by rdspring1** [comment]
!build

**Comment by rdspring1** [comment]
!build

**Comment by rdspring1** [comment]
!build

**Comment by rdspring1** [comment]
@jjsjann123 I changed this PR to remove the `truediv` operation.

File: csrc/ops/arith.h [hunk 2]
```diff
@@ -452,12 +452,6 @@ NVF_API Val* atan2(Val* v1, Val* v2);
 NVF_API TensorView* atan2(TensorView* v1, Val* v2);
 NVF_API TensorView* atan2(Val* v1, TensorView* v2);
 NVF_API TensorView* atan2(TensorView* v1, TensorView* v2);
-// truediv: promote to float for integer division, has the same semantics as the
-// python's operator /
-NVF_API Val* truediv(Val* v1, Val* v2);
-NVF_API TensorView* truediv(TensorView* v1, Val* v2);
-NVF_API TensorView* truediv(Val* v1, TensorView* v2);
-NVF_API TensorView* truediv(TensorView* v1, TensorView* v2);
```
jjsjann123, line 460: Naive question, we don't think we need/should remove the c++ truediv API. Since these does float promotion, so it is a truediv in nature? Even though it's not a BinaryOpType::TrueDiv.

jjsjann123 (commented):
I think I see why you want to have a `BinaryOpType::Truediv`. in #2841, this is just so we have a clean 1-to-1 translation for the truediv part?

I'm not sure how big of a deal it is to have that 1-to-1 translation. We don't have to keep a `truediv` binary op, having it implicitly handled by floating point promotion followed by a c++ style div should be good enough.

*not* having truediv is somewhat an inconvenience IMHO. thunder doesn't have to use it, but we still have our python API where I can define fusion definition manually. So I tend to think maybe we want to keep it, but I don't have a strong opinion there neither.

jjsjann123 (commented):
BTW, this isn't necessarily caused by your PR. I'm glad you are exposing these issues. :heart: 

Two things that I do have a stronger opinion on:
1. In this PR, the BinaryOpType::Truediv isn't doing true div, so that doesn't feel right.
2. Not in your code, but our python binding has
```
1555   // In PyTorch, __div__ (//) and __truediv__ (/) are different.
1556   // When applied to integer-dtype arguments, they do as expected, returning
1557   // integer and float outputs, respectively. When applied to two floating-type
1558   // arguments, they return the floor of division for // and plain division for
1559   // /. When applied to mixed types, the types are promoted, so the
1560   // floating-point behavior is returned.
1561   // Our div operator matches the __truediv__ behavior, so we do not implement
1562   // __div__.
1563   NVFUSER_PYTHON_BINDING_BINARY_OP_SPECIAL("__truediv__", "div", div)
```

This is just totally wrong and we should fix it.

PR #2841, opened 2024-08-24
Create CPP Fusion to Python FusionDefinition framework

rdspring1:
## Goal ##
Translate from CPP `Fusion` to Python `FusionDefinition`

## Why ##
- Access fusions in python frontend after running `Presegmentation` and `Segmentation` passes.  

## Overview ##

- This PR creates `void translate(Fusion* fusion, FusionDefinition* fd)`, which translates the CPP fusion's expressions and values into corresponding `RecordFunctors` for `FusionDefinition`.
- `FusionDefinition::clone(FusionDefinition& other)` acts like a copy constructor. It translates the prescheduled fusion from `other` to the blank `FusionDefinition`. Only used for testing purposes.
- It adds `check_cpp_translation` check to python testing, which checks that the translated CPP fusion is functionally equivalent to original `FusionDefinition`.
- Currently, only the binary operations are supported with `FusionTranslation`. The remaining operations will be added in future PRs.

### DIFF
--- a/csrc/ir/internal_base_nodes.h
+++ b/csrc/ir/internal_base_nodes.h
@@ -475,6 +475,10 @@ class TensorDomain : public Val {
     return contiguity_;
   }
 
+  const std::vector<int64_t>& strideOrder() const {
+    return stride_order_;
+  }
+
   NVF_API void setContiguity(const std::vector<std::optional<bool>>& contig);
 
   std::string getContiguityString() const {
@@ -672,6 +676,7 @@ class TensorDomain : public Val {
   std::vector<IterDomain*> no_bcast_domain_;
   std::vector<IterDomain*> no_reduction_domain_;
   std::vector<std::optional<bool>> contiguity_;
+  std::vector<int64_t> stride_order_;
   bool has_reduction_ = false;
 };
 
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -2954,7 +2954,8 @@ TensorDomain::TensorDomain(
       loop_domain_(logical_domain_),
       contiguity_(
           contiguity.empty() ? getContiguityFilledWith(maybeAllocation(), false)
-                             : std::move(contiguity)) {
+                             : std::move(contiguity)),
+      stride_order_(stride_order) {
   // setting the proper allocation domain
   if (!stride_order.empty()) {
     auto rank = logical_domain_.size();
--- a/csrc/python_frontend/fusion_definition.cpp
+++ b/csrc/python_frontend/fusion_definition.cpp
@@ -13,6 +13,7 @@
 #include <options.h>
 #include <python_frontend/fusion_cache.h>
 #include <python_frontend/fusion_definition.h>
+#include <python_frontend/translation.h>
 #include <scheduler/heuristic_types.h>
 #include <utils.h>
 #include <validator_utils.h>
@@ -203,14 +204,15 @@ void FusionDefinition::setupSchedule(const at::ArrayRef<c10::IValue>& inputs) {
 
   user_sched_ = fusionCache()->createUserSchedule(scheds, inputs, device);
 
-  // Building a new Fusion container for scheduling with definition such that
-  // the definition's tensor data members refer to the corresponding IR objects
-  // needed for scheduling. A simple copy of the container would mean the data
-  // members that represent tensors would refer to the IR objects in the
-  // original and not the copy needed for scheduling.
+  // Building a new Fusion container for scheduling with definition such
+  // that the definition's tensor data members refer to the corresponding IR
+  // objects needed for scheduling. A simple copy of the container would
+  // mean the data members that represent tensors would refer to the IR
+  // objects in the original and not the copy needed for scheduling.
   buildFusionIr(user_sched_->schedule.get());
 
-  // Add TensorViews created by composite operations to Python FusionDefinition.
+  // Add TensorViews created by composite operations to Python
+  // FusionDefinition.
   findHiddenTensorViews(user_sched_->schedule.get());
 
   KernelArgumentHolder args =
@@ -357,9 +359,9 @@ std::vector<at::Tensor> FusionDefinition::execute(
     }
   }
 
-  // when `!override_user_schedule == true`, it *could* have produced an output
-  // already at this point and we would not want to overwrite generated output
-  // through user scheduled kernel.
+  // when `!override_user_schedule == true`, it *could* have produced an
+  // output already at this point and we would not want to overwrite
+  // generated output through user scheduled kernel.
   if (outputs.empty()) {
     outputs = scheds->auto_gen_schedules->runFusionWithInputs(
         inputs, std::nullopt, selected_device);
@@ -621,6 +623,14 @@ std::vector<Tensor> FusionDefinition::tensors() {
 std::vector<std::pair<double, double>> FusionDefinition::getValTolerances(
     const at::ArrayRef<c10::IValue>& inputs) {
   return get_val_constants(preschedFusion(), inputs);
+}
+
+void FusionDefinition::clone(FusionDefinition& other) {
+  NVF_ERROR(!completed(), "Expected an incomplete definition before cloning!");
+  NVF_ERROR(
+      other.completed(),
+      "Expected incoming FusionDefinition to be omplete before cloning!");
+  return translate(other.preschedFusion(), this);
 }
 
 } // namespace nvfuser::python_frontend
--- a/csrc/python_frontend/fusion_definition.h
+++ b/csrc/python_frontend/fusion_definition.h
@@ -165,6 +165,12 @@ class NVF_API FusionDefinition : public FusionState {
   FusionDefinition(FusionDefinition&& fd) = delete;
   FusionDefinition& operator=(const FusionDefinition& fd) = delete;
   FusionDefinition& operator=(FusionDefinition&& fd) = delete;
+  ~FusionDefinition() override = default;
+
+  //! Copy definition from other FusionDefintion's presched CPP fusion.
+  //! Primarily for testing purposes to check that the translation from CPP
+  //! fusion is correct.
+  void clone(FusionDefinition& other);
 
   //! Enter Python Context Manager -- Reset trie for new cache lookup
   NVF_API FusionDefinition* setupDefinition();
--- a/csrc/python_frontend/python_bindings.cpp
+++ b/csrc/python_frontend/python_bindings.cpp
@@ -704,6 +704,11 @@ void initNvFuserPythonBindings(PyObject* module) {
             return ss.str();
           })
       .def(
+          "_clone",
+          [](FusionDefinition& self, FusionDefinition& other) {
+            self.clone(other);
+          })
+      .def(
           "_execute",
           [](FusionDefinition& self,
              const py::iterable& iter,
--- a/csrc/python_frontend/translation.cpp
+++ b/csrc/python_frontend/translation.cpp
@@ -1,0 +1,252 @@
+// clang-format off
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2023-present NVIDIA CORPORATION & AFFILIATES.
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+// clang-format on
+//
+#include <dispatch.h>
+#include <python_frontend/translation.h>
+#include <python_frontend/translation_utils.h>
+
+#include <vector>
+
+namespace nvfuser::python_frontend {
+namespace {
+class FusionTranslator : public OptInConstDispatch {
+ public:
+  static void translate(Fusion* fusion, FusionDefinition* fd) {
+    FusionTranslator translator(fusion, fd);
+    translator.translate();
+  }
+
+ private:
+  FusionTranslator(Fusion* fusion, FusionDefinition* fd)
+      : fusion_(fusion), fd_(fd) {}
+
+  bool checkExpressionDependencies(Expr* e) {
+    return std::all_of(
+        e->inputs().begin(), e->inputs().end(), [&](const Val* v) {
+          return map_val_to_fd_index_.count(v) > 0;
+        });
+  }
+
+  void translate() {
+    fd_->setupDefinition();
+
+    std::deque<nvfuser::Expr*> to_visit;
+
+    // Add Fusion inputs to FusionDefinition
+    for (nvfuser::Val* v : fusion_->inputs()) {
+      OptOutConstDispatch::dispatch(v);
+
+      // Add uses for input value to to_visit
+      for (Expr* e : v->uses()) {
+        to_visit.push_back(e);
+      }
+    }
+
+    // Topological search of Fusion expressions
+    std::unordered_set<nvfuser::Expr*> visited;
+    while (!to_visit.empty()) {
+      Expr* e = to_visit.front();
+      to_visit.pop_front();
+
+      // short-circuit: skip if already visited
+      if (visited.count(e) > 0) {
+        continue;
+      }
+
+      // short-circuit: add to back of stack if not all of the expression's
+      // inputs are available.
+      if (!checkExpressionDependencies(e)) {
+        to_visit.push_back(e);
+        continue;
+      }
+
+      visited.insert(e);
+
+      // Create RecordFunctor given inputs, outputs, and attributes.
+      OptOutConstDispatch::dispatch(e);
+
+      // Add output uses to to_visit
+      for (Val* v : e->outputs()) {
+        for (Expr* e : v->uses()) {
+          to_visit.push_back(e);
+        }
+      }
+    }
+
+    // Outputs and Aliasing
+    for (nvfuser::Val* v : fusion_->outputs()) {
+      // Handle only TensorViews
+      NVF_ERROR(v->isA<TensorView>());
+      handleOutput(v->as<TensorView>());
+    }
+
+    fd_->finalizeDefinition();
+  }
+
+  void handle(const Val* v) final {
+    Scalar output = fd_->defineScalar();
+    fd_->defineRecord(new ScalarRecord(
+        {fd_->recordingState(output())},
+        v->value(),
+        std::get<PrimDataType>(v->dtype().type)));
+    map_val_to_fd_index_.emplace(v, output());
+  }
+
+  void handle(const TensorView* tv) final {
+    Tensor output = fd_->defineTensor(tv->nDims());
+
+    std::vector<int64_t> shape;
+    std::transform(
+        tv->domain()->logical().begin(),
+        tv->domain()->logical().end(),
+        std::back_inserter(shape),
+        [](IterDomain* id) {
+          return (id->extent()->isConstScalar())
+              ? id->extent()->evaluate().as<int64_t>()
+              : -1;
+        });
+
+    fd_->defineRecord(new TensorRecord(
+        {fd_->recordingState(output())},
+        shape,
+        tv->domain()->contiguity(),
+        std::get<PrimDataType>(tv->dtype().type),
+        tv->isCpuScalar(),
+        tv->domain()->strideOrder()));
+
+    map_val_to_fd_index_.emplace(tv, output());
+  }
+
+  void handleOutput(const TensorView* tv) {
+    // Add fusion outputs to FusionDefinition
+    size_t output_index = map_val_to_fd_index_.at(tv);
+    fd_->defineRecord(new OutputRecord<TensorView>(
+        {fd_->recordingState(output_index)}, serde::RecordType::OutputTv));
+  }
+
+  void handle(const BroadcastOp* bcast_op) final {
+    Tensor output =
+        fd_->defineTensor(bcast_op->out()->as<TensorView>()->nDims());
+    fd_->defineRecord(new BroadcastOpRecord(
+        {fd_->recordingState(map_val_to_fd_index_.at(bcast_op->in()))},
+        {fd_->recordingState(output())},
+        "ops.broadcast",
+        bcast_op->getBroadcastDimFlags()));
+    map_val_to_fd_index_.emplace(bcast_op->out(), output());
+  }
+
+  template <typename ExprType, typename ResultType, typename... ArgTypes>
+  void handleOpRecord(
+      const Expr* e,
+      serde::RecordType record_type,
+      ResultType result,
+      ArgTypes... args) {
+    NVF_ERROR(e->isA<ExprType>());
+    std::vector<State> argument_states;
+    std::transform(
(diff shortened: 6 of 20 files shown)


rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

tests/python/utils.py
```diff
@@ -240,6 +240,37 @@ def check_captured_python_definition(reference_outputs, fd, inputs, device=None)
...
 
+# Run original FusionDefinition
+# Clone FusionDefinition
+# Run cloned python definition
+# Check that the result of cloned python definition matches original results
+def check_cpp_translation(reference_outputs, fd, inputs, device=None):
+    try:
+        torch.manual_seed(0)
+        cloned_fd = fd.clone()
+        cloned_outputs = cloned_fd.execute(inputs, device=device)
```
jacobhinkle, line 251: Maybe we should print the fusion IR also and compare the output?

tests/python/test_python_frontend.py
```diff
@@ -701,7 +701,7 @@ def fusion_func(fd: FusionDefinition):
             fd.add_output(t1, alias_input=t0)
             fd.add_output(t2)
 
-        out_tensors, _ = self.exec_nvfuser(fusion_func, in_tensors)
+        out_tensors, _ = self.exec_nvfuser(fusion_func, in_tensors, is_clonable=True)
```
jacobhinkle, line 704: Should we just make `is_clonable=True` the default and mark the ones that currently are not clonable?

csrc/serde/fusion_record.cpp
```diff
@@ -812,6 +812,8 @@ void RecordFunctorFactory::setupFunctionMaps() {
   NVFUSER_BINARY_TV_OP("bitwise_and", bitwise_and)
   NVFUSER_BINARY_TV_OP("bitwise_or", bitwise_or)
   NVFUSER_BINARY_TV_OP("bitwise_xor", bitwise_xor)
+  NVFUSER_BINARY_TV_OP("logical_and", logical_and)
+  NVFUSER_BINARY_TV_OP("logical_or", logical_or)
```
jacobhinkle, line 816: Does the absence of these mean that previously we could not serialize `logical_and` and `logical_or`?

csrc/ir/internal_base_nodes.h
```diff
@@ -475,6 +475,8 @@ class TensorDomain : public Val {
     return contiguity_;
   }
 
+  std::vector<int64_t> strideOrder() const;
```
jacobhinkle, line 478: Please provide a comment here telling us the meaning of the returned ints.

csrc/python_frontend/translation.cpp
```diff
@@ -0,0 +1,524 @@
...
+//
+// Rather than create a new FusionDefinition from the CPP Fusion, we add
+// RecordFunctors to a blank FusionDefinition. This is a design decision because
+// of the FusionDefinition python class, which inherits from the
+// _C._FusionDefinition class created by pybind11. It is easier to operate on
+// the child class directly than to create a new child instance from parent
+// instance.
+class FusionTranslator : public OptInConstDispatch {
+ public:
+  static std::unordered_map<const nvfuser::Val*, size_t> translate(
```
jacobhinkle, line 31: A comment would be helpful here for understanding the return value.

csrc/python_frontend/translation.cpp
```diff
@@ -0,0 +1,524 @@
...
+          "Cycle detected: None of the expressions can be processed!");
+
+      // short-circuit: skip if already visited
+      if (visited.count(e) > 0) {
+        continue;
+      }
+
+      // short-circuit: skip Split and Merge expressions created by Reshape
+      // short-circuit: skip Resize expressions created by Slice
+      if (e->isA<Split>() || e->isA<Merge>() || e->isA<Resize>()) {
```
jacobhinkle, line 175: Do we already check somewhere that no TVs have loop domains, so that it's safe to only check these definitional transforms i.e. that we can skip `Swizzle`?

csrc/python_frontend/translation.cpp
```diff
@@ -0,0 +1,524 @@
...
+    // Handle only Cast operation
+    NVF_ERROR(uop->getUnaryOpType() == UnaryOpType::Cast);
+    handleCastOp(uop);
+  }
+
+  // Map cast UnaryOp to CastOpRecord
+  void handleCastOp(const Expr* op) {
+    bool is_cast_op = op->isA<UnaryOp>() &&
+        op->as<UnaryOp>()->getUnaryOpType() == UnaryOpType::Cast;
+    NVF_ERROR(op->isA<LoadStoreOp>() || is_cast_op);
```
jacobhinkle, line 416: Why is `LoadStoreOp` accepted here?

csrc/python_frontend/translation.cpp
```diff
@@ -0,0 +1,524 @@
...
+          bop,
+          serde::RecordType::Binary_VAL,
+          bop->out(),
+          bop->lhs(),
+          bop->rhs());
+    }
+  }
+
+  // Add Broadcast operation to FusionDefinition
+  void handle(const BroadcastOp* bcast_op) final {
```
jacobhinkle, line 495: I think we might actually need to do some pattern matching here since `broadcast_in_dim` translates to broadcast+expand. You could always just match the `ExpandOp` individually but then if you actually input a `broadcast_in_dim` from python and did a round trip you would wind up with _two_ `broadcast_in_dim` calls I think.

jacobhinkle (commented):
I really like this idea. It will also let us get repros for individual segments when they fail. I have a few comments so far. I also am slightly worried about how we will keep this in sync with the frontend ops as we add more of those; when we add an op we'll have at least these three files to update. Maybe there's a way to colocate `translation*` stuff in `python_bindings.cpp`?

rdspring1:
!build

tests/python/utils.py
rdspring1: I was only after functional equivalence for now. Perfect string matching can be tedious.

tests/python/test_python_frontend.py
rdspring1: With this PR, there are roughly 78 failing and 35 passing. Most of the 35 passing tests overlap with the binary opinfo tests. I need a test specific for alias testing.

csrc/python_frontend/translation.cpp
rdspring1: Future PRs will add the `void handle(const ExpandOp* eop)` for expand operations. To keep the visitor pattern clean, it would be easier to add an `ExpandOpRecord` and `expand` to the python-frontend.

jacobhinkle: > To keep the visitor pattern clean, it would be easier to add an `ExpandOpRecord` and `expand` to the python-frontend.

I agree.

csrc/python_frontend/translation.cpp
rdspring1: If the dtype between input and output are the same, `castOp` creates a `LoadStoreOp`. See https://github.com/NVIDIA/Fuser/blob/main/csrc/ops/arith.cpp#L30-L34.

If we see a `LoadStoreOp` without permutation, we map back to a `castOp`.

rdspring1: `Broadcast` also creates a `set` if it is a no-op.

jacobhinkle: I see. So we never map back to `fd.ops.set` then. I wonder if we should just skip translating trivial sets unless they are `segment_set`.

tests/python/utils.py
rdspring1: The python and cpp APIs are not a perfect 1:1 match. String matching would be useful for the second round of PRs.

rdspring1:
> How we will keep this in sync with the frontend ops?

Eventually, every test that goes through `exec_nvfuser` will check if it is cloneable.
The `FusionTranslate` class inherits from `OptInConstDispatch`.
When a new operation is added and its `handle` function is not defined, the `clone` function will automatically fail.

rdspring1:
!build

rdspring1:
!build

csrc/python_frontend/translation.cpp
rdspring1: I added a check to compare the loop domain against the logical domain.

https://github.com/NVIDIA/Fuser/pull/2841/files#diff-22914f1fdfa0d5e8983f85f18b2232e2eeadb6df2c28fdb4e7c6ee113d991328R202-R209

csrc/ir/nodes.cpp
```diff
@@ -3235,6 +3235,29 @@ void TensorDomain::setContiguity(
...
+  for (size_t logical_idx : c10::irange(logical_domain_.size())) {
+    IterDomain* logical_id = logical_domain_.at(logical_idx);
+    auto alloc_iter = std::find(
+        allocation_domain_.begin(), allocation_domain_.end(), logical_id);
+    NVF_ERROR(
+        alloc_iter != allocation_domain_.end(),
+        "Unable to find logical IterDomain in allocation domain.");
+    int64_t alloc_idx = std::distance(allocation_domain_.begin(), alloc_iter);
+    stride_order.push_back((int64_t)logical_domain_.size() - 1 - alloc_idx);
+  }
```
jjsjann123, line 3256: :cry: the `stride_order` things really annoys me, otherwise we could just use the `computePermutation` in ir/utils.h

csrc/python_frontend/fusion_definition.cpp
```diff
@@ -624,4 +626,12 @@ std::vector<std::pair<double, double>> FusionDefinition::getValTolerances(
   return get_val_constants(preschedFusion(), inputs);
 }
 
+void FusionDefinition::clone(FusionDefinition& other) {
+  NVF_CHECK(id().has_value(), "FusionDefinition definition does not exist!");
```
jjsjann123, line 630: nitpick, `id().has_value()` is just a `complete()` here. I think it looks cleaner to keep it consistent with the `!other.complete()`.

csrc/python_frontend/python_bindings.cpp
```diff
@@ -703,6 +703,11 @@ void initNvFuserPythonBindings(PyObject* module) {
             self.print(ss);
             return ss.str();
           })
+      .def(
+          "_clone",
+          [](FusionDefinition& self, FusionDefinition& other) {
+            self.clone(other);
```
jjsjann123, line 709: any reason for this indirection here? if we are just calling `self.clone` anyway.

tests/python/utils.py
```diff
@@ -366,21 +397,33 @@ def setup_class(cls):
...
+        device=None,
+        is_clonable=False,
     ):
         fc = FusionCache.get()
         before_fusions = fc.num_fusions()
         # Copy inputs because aliased outputs can modify inputs when running
         # FusionDefinition
-        inputs_cap = deepcopy(inputs)
+        inputs_captured = deepcopy(inputs)
+        inputs_cloned = deepcopy(inputs)
```
jjsjann123, line 413: nitpick, this deepcopy is only needed with `is_clonable==True`

csrc/python_frontend/python_bindings.cpp
jjsjann123: sorry this is just meant to be a nitpicking here.

If clone is indeed just a debug function, let's not put it inside FusionDefinition, but rather as a function directly here.

csrc/python_frontend/fusion_definition.cpp
```diff
@@ -624,4 +626,12 @@ std::vector<std::pair<double, double>> FusionDefinition::getValTolerances(
...
+  NVF_ERROR(
+      !other.completed(),
+      "Expected an incomplete definition before translation.");
+  translate(preschedFusion(), &other);
```
jjsjann123, line 634: I'm now questioning why `FusionDefinition::preschedFusion()` is a private method. That definition is just calling all public methods.
I'm leaning towards having this debug function by itself.

csrc/python_frontend/translation.cpp
```diff
@@ -0,0 +1,555 @@
...
+          }
+          // An alias output can also be returned as a fusion output
+          // if it is already aliased or if not hide_output
+          if (num_visited > 0 || !alias_info.hide_output) {
+            handleOutput(v->as<TensorView>());
+          }
+          break;
+        }
+        default:
+          NVF_ERROR(false, "Unsupported AllocationType");
```
jjsjann123, line 260: I see Evaluate is not supported.

Don't have to add it here in this PR, I'm wondering if we can actually just ignore that.. i.e. Evaluate is supposed to be added in our transformation, since we do not have it exposed at python API where user can annotate which output gets executed by ExprEvaluator.

csrc/python_frontend/translation.cpp
```diff
@@ -0,0 +1,555 @@
...
+  // Returns a map from the values in the CPP fusion to its corresponding
+  // FusionDefinition State index.
+  //
+  // Why?
+  // For segmentation, we divide the original FusionDefinition into its
+  // segments. Each segment has a separate index namespace. To run a segment,
+  // we need to pass outputs from prior segments as this segment's input
+  // arguments. The original FusionDefinition coordinates this argument passing.
+  // The map returned by this function is used to a global mapping from the
+  // original FusionDefinition's indicies to this segment's indicies.
```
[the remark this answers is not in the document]
rdspring1: Yes, it is used in the segmentation PRs.

csrc/python_frontend/python_bindings.cpp
rdspring1: I'm not sure what you mean. Can you add a code snippet example?

rdspring1:
!build

csrc/python_frontend/python_bindings.cpp
jjsjann123: Apology for vague comment.

So instead of having clone as a member function inside `FusionDefinition`. I think we should push this out as a function in `python_bindings.cpp`, if it's meant to be just a debug function.

```
void FusionDefinition::clone(FusionDefinition& other) {
  NVF_CHECK(completed(), "FusionDefinition definition does not exist!");
  NVF_ERROR(
      !other.completed(),
      "Expected an incomplete definition before translation.");
  translate(preschedFusion(), &other);
}
```

i.e.

```
void clone(FusionDefinition& from, FusionDefinition& to) {
  NVF_CHECK(from.completed(), "FusionDefinition definition does not exist!");
  NVF_ERROR(
      !to.completed(),
      "Expected an incomplete definition before translation.");
  // Something like this would work? vvv
  translate(fusionCache()->queryFusionSchedules(from.id())->preschedFusion(), &to);
}
```

jjsjann123:
I think one question I would want to resolve first is, are we pushing for having more segmentation stuff plumbed back to python side? Is this going to be expanded further to support a manual scheduling with segmented FusionDefinition, or are we just going for a smaller repro python script?

> I also am slightly worried about how we will keep this in sync with the frontend ops as we add more of those; when we add an op we'll have at least these three files to update. Maybe there's a way to colocate translation* stuff in python_bindings.cpp?

@jacobhinkle's questions/suggestions are great. I'm not pushing against this PR. I'm fine with us plumbing the suggested recommendation in follow up PRs if we are moving forward with this. What do you think? cc'ing @jacobhinkle @kevinstephano

csrc/python_frontend/translation.cpp
```diff
@@ -0,0 +1,555 @@
...
+      OptOutConstDispatch::dispatch(v);
+    }
+
+    // Gather all expressions in CPP Fusion.
+    const std::vector<nvfuser::Expr*> fusion_exprs = fusion_->exprs();
+    std::deque<nvfuser::Expr*> to_visit(
+        fusion_exprs.begin(), fusion_exprs.end());
+
+    // Scalar expressions are not handled by usedMathVals, so gather them
+    // manually.
```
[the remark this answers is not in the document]
jacobhinkle: Yeah it seems like `fusion_->exprs()` should already have all those scalar expressions in topological order to me since it calls `StmtSort::getExprs()`.

rdspring1:
> I think one question I would want to resolve first is, are we pushing for having more segmentation stuff plumbed back to python side? Is this going to be expanded further to support a manual scheduling with segmented FusionDefinition, or are we just going for a smaller repro python script?

In my personal branch, I've already expanded the feature to support manual scheduling of segmented fusions.

rdspring1:
!build

csrc/python_frontend/translation.cpp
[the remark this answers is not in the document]
jjsjann123: cc'ing @rdspring1 in case you missed this one.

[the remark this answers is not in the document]
rdspring1: The scalar expressions are defined for the extents of the IterDomains for a `reshape` TensorView's logical domain.
`StmtSort::getExprs()` only traverses a value's inputs. 

`TV1 [id0, id1, (id2 * id3)] = view(TV0[id0, id1, id2, id3])` --- We don't store the scalar expressions in the `ViewOp` itself, so there is nothing to gather with `StmtSort::getExprs`

csrc/python_frontend/fusion_definition.h
```diff
@@ -228,6 +228,9 @@ class NVF_API FusionDefinition : public FusionState {
     return id().has_value();
   }
 
+  //! Return a prescheduled Fusion object
```
rdspring1, line 231: @jjsjann123 I made `preschedFusion` public.

rdspring1:
@jacobhinkle @jacobhinkle https://github.com/NVIDIA/Fuser/pull/3025 contains the complete draft of changes to handle segmentation in python frontend.

Once you can get the FusionDefinition for the segments, it is straight-forward to run them in topological order.
See  [segment](https://github.com/NVIDIA/Fuser/blob/user_sched_segmentation/nvfuser/__init__.py#L57-L78) and [_execute_segment](https://github.com/NVIDIA/Fuser/blob/user_sched_segmentation/nvfuser/__init__.py#L162-L205) for details.

rdspring1:
!build

rdspring1:
!build

csrc/python_frontend/translation_utils.h
```diff
@@ -0,0 +1,116 @@
...
+#include <ir/all_nodes.h>
+#include <ops/all_ops.h>
+
+namespace nvfuser::python_frontend {
+
+// Get std::function for BinaryOp
+template <typename ResultType, typename... ArgTypes>
+std::function<ResultType(ArgTypes...)> getFunction(const BinaryOp* bop) {
+  auto get_std_function = [](ResultType (*fn)(ArgTypes...)) {
+    return static_cast<ResultType (*)(ArgTypes...)>(fn);
```
jjsjann123, line 18: Do we need to have a static_cast here? I see that we are using `get_std_function` for overload resolution, but I don't see why we need a cast to the same function pointer?

jjsjann123: another nitpick, `get_std_function` sounds like we are getting std::..., but this is just our ops/arith.h stuff right?

csrc/python_frontend/translation.cpp
```diff
@@ -0,0 +1,560 @@
...
+  // python_frontend
+
+  // A generic function to map UnaryOp, BinaryOp, and TernaryOp to
+  // python_frontend OpRecord
+  template <typename ExprType, typename ResultType, typename... ArgTypes>
+  void handleOpRecord(
+      const Expr* e,
+      serde::RecordType record_type,
+      ResultType result,
+      ArgTypes... args) {
```
jjsjann123, line 421: Note for myself, I see are using `ArgTypes...` just for the template, but `args` aren't really used.

csrc/python_frontend/translation.cpp
```diff
@@ -0,0 +1,560 @@
...
+        [&](auto arg) {
+          return fd_->recordingState(map_val_to_fd_index_.at(arg));
+        });
+
+    fd_->defineRecord(new OpRecord<ResultType, ArgTypes...>(
+        argument_states,
+        {fd_->recordingState(map_val_to_fd_index_.at(result))},
+        "ops." + getString(e->as<ExprType>()),
+        record_type,
+        getFunction<ResultType, ArgTypes...>(e->as<ExprType>())));
```
jjsjann123, line 437: If I'm being a language police, I would say that the template `getFunction` seems a bit too complicated. I think you can construct a function type here with `ResultType(*)(ArgTypes...)` and use static_cast along to resolve the overload.

jjsjann123 (approved):
some minor comment/nitpicking on the template part. I'm fine with the code in this PR.

stamping to unblock progress.

csrc/python_frontend/translation_utils.cpp
```diff
@@ -0,0 +1,107 @@
...
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+// clang-format on
+//
+#include <python_frontend/translation_utils.h>
+
+namespace nvfuser::python_frontend {
+
+std::string getString(const BinaryOp* bop) {
```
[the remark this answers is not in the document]
rdspring1: `binary_op_type2string` doesn't always return the same string as `getString`, which corresponds with the names in `python_bindings.cpp`.
e.g., `BinaryOpType::Eq` return "equals" for `binary_op_type2string` vs "eq" in `getString`.

csrc/python_frontend/translation.cpp
rdspring1: Iterating over variadic templates seemed overly complicated.

csrc/python_frontend/translation_utils.h
rdspring1: I was seeing an error before but it works now.
I renamed `get_std_function` to `wrap_function`.

rdspring1:
!build

csrc/python_frontend/translation.cpp
rdspring1: Returning a function pointer has ugly syntax, so I went with returning `std::function`.

csrc/python_frontend/translation_utils.cpp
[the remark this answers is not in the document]
jjsjann123: Ha, that's subtle. Thanks for pointing that out. 🙇 

I'm wondering if `binary_op_type2string` is used anywhere in codegen? I guess not since there are things like `greaterThanOrEqual`...
Does it make sense to update `binary_op_type2string` to match `python_binding.cpp`'s string, which is customer facing already.

cc'ing @naoyam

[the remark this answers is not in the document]
jjsjann123: BTW, I'm not blocking merging the PR on this one. We can resolve it later if we decide to merge the two.

PR #2842, opened 2024-08-24
patch atexit_serde_create_op_test

rdspring1:
* Failed to run correctness tests because the scope member was not initialized correctly.


rdspring1:
!build

rdspring1:
!build

tests/python/pytest_framework.py
```diff
@@ -74,4 +74,11 @@ def __init__(self, opinfos, *, scope=None):
 
         atexit_serde_check()
 
-        create_op_test.__init__(self, opinfos, scope=scope)
+        # Replicate create_op_test.__init__ to have correct scope
```
jjsjann123, line 77: to make this slightly more future proof. should we still be calling `super().__init__` but with the updated `scope` arg instead?

jjsjann123: admittedly, it doesn't make much of a difference given the code as-is.

jjsjann123 (approved):
LGTM with minor nitpick

tests/python/pytest_framework.py
rdspring1: But that adds to the scope if you call another function.

jjsjann123: sorry, I meant that we keep the part about resolving the scope here, but just pass the resolved scope. So we get to re-use the `self.opinfos = opinfos` in the super methods.

Again, this is really just nitpicking... It's your call :laughing:

rdspring1:
!build

tests/python/pytest_framework.py
rdspring1: Since I needed to rebase, I added this change.

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
@jjsjann123 I removed the per-op serialization check because it was making the CI take too long.

Currently, the python tests take about an hour. With the serialization check, it is over 1 hour, 45 minutes.

We'll just run the serialization check at exit for the opinfo tests in the CI.

jjsjann123:
> Currently, the python tests take about an hour. With the serialization check, it is over 1 hour, 45 minutes.
> 
> We'll just run the serialization check at exit for the opinfo tests in the CI.

naive question, are the two checks identical?
I thought the atexit stuff only checks for serialization. It doesn't attempt to deserialize and verify the test like `serde_check_ops` does.

If so, should we at least run serde_check_ops in our nightly/weekly runs to ensure coverage?

rdspring1:
You're correct. `atexit` only does the serialization part in the CI. `serde_check_ops` does serialization, deserialization, and verify. We have the `serde_check_ops` in `test_python_frontend`, so we have daily coverage there.

The opinfo tests covers each op individually and handles more inputs, so it is slower. The opinfos are useful for development or debugging because every op is isolated.

> If so, should we at least run serde_check_ops in our nightly/weekly runs to ensure coverage?

That is a good idea. If you have unlimited time it doesn't hurt to run them, but `test_python_frontend` provides good quality assurance.

PR #2845, opened 2024-08-26
Run test_host_ir and test_multidevice in manual_ci.sh

jacobhinkle:
This small PR just updates `manual_ci.sh` to have more complete coverage for local testing.


manual_ci.sh
```diff
@@ -21,7 +21,11 @@ run_test './bin/lib/dynamic_type/test_dynamic_type_17'
 run_test './bin/lib/dynamic_type/test_dynamic_type_20'
 run_test './bin/nvfuser_tests'
 run_test './bin/test_rng'
-# run_test './bin/test_multidevice'
+run_test './bin/test_host_ir'
+if type -p mpirun > /dev/null
```
jjsjann123, line 25: out of curiosity, why do we care about having `-p` here?

jjsjann123 (approved):

manual_ci.sh
jacobhinkle: No strong reason, but `type -p` will always write to stdout either the path to the executable or nothing whereas `type` alone will write to stdout if the argument is a command and it will write an error message to stderr if it's not found. So `type -p $CMD > /dev/null` lets you avoid redirecting stderr along with stdout. I will say I had to look this up though as it's one of those things I do mindlessly :sweat_smile:

PR #2852, opened 2024-08-27
Clean up PadOp index lowering

naoyam:


naoyam:
!build

csrc/id_model/indexing.cpp
```diff
@@ -1131,9 +1131,6 @@ std::vector<PredicateInfo> TensorIndexer::getPredicates(
           /*is_start_predicate=*/false,
           /*unswitched_loop=*/unswitched_loop);
 
-  const std::vector<PredicateDomainInfo> non_divisible_split_predicates =
```
naoyam, line 1134: It was just unused

tests/cpp/test_resize.cpp
```diff
@@ -23,7 +23,7 @@
 
 namespace nvfuser {
 
-using ResizeTest = NVFuserTest;
+using ResizeTest = NVFuserFixtureParamTest<bool>;
```
naoyam, line 26: Pad unit tests run both the legacy and new indexers. Will change the slice and cat tests later.

csrc/device_lower/pass/index.cpp
```diff
@@ -1872,8 +1872,8 @@ Val* IndexLowering::getIterationIndexForBroadcast(
 
 void IndexLowering::handle(const PadOp* pad) {
   // Convert to a where op as:
-  // consumer[consumer_idx] = (producer_idx >= 0 && producer_idx <
-  //                           producer_extent) ?
+  // consumer[consumer_idx] = (consumer_idx >= left_pad && consumer_idx <
```
naoyam, line 1875: Don't remember why I originally used the producer logical indices. The override index isn't necessary if the consumer indices are used.

csrc/device_lower/pass/index.cpp
```diff
@@ -1887,37 +1887,27 @@ void IndexLowering::handle(const PadOp* pad) {
 
   const auto pad_val = pad->value();
 
-  std::unordered_map<IterDomain*, Val*> override_index;
```
naoyam, line 1890: This was necessary because a broadcast domain gets 0 as its index, no matter what the index of the corresponding consumer domain is.

naoyam:
!build

naoyam:
!build

tests/cpp/test_resize.cpp
```diff
@@ -50,6 +56,13 @@ TEST_F(ResizeTest, Pad1) {
   auto t0 = at::randn(shape, options);
   std::vector<c10::IValue> aten_inputs({t0});
 
+  EnableOptionsGuard enable_options_guard;
+  if (GetParam()) {
+    EnableOptionsGuard::getCurOptions().set(EnableOption::IdModel, {"all"});
+  } else {
+    EnableOptionsGuard::getCurOptions().unset(EnableOption::IdModel);
+  }
```
jacobhinkle, line 64: We could place this in `ResizeTest::SetUp()`.

csrc/device_lower/pass/index.cpp
jacobhinkle: Ah yes, this was needed because of using producer index in the pad predicate. Now that we are using consumer index, it's not needed. That's great. I believe `getIterationIndexForBroadcast` above can now be removed.

jacobhinkle (approved):
LGTM. See comments.

PR #2815, opened 2024-08-20
MarkAliasesPrepare applies bookend from inputs as well as outputs. 

wujingyue:
As a follow-up to #2639. #2639 bookends from the outputs, and this PR bookends from the inputs. 

Fixes #2599. 
Fixes #2577. 

Benchmark results are mostly neutral. `test_nanogpt_layer_norm` got slower because of the added host latency. This is expected because, compared with #2639, latency of ops bookended from the inputs is more likely to appear on the critical path. See benchmark details below. 

```
$ nvidia-smi -L
GPU 0: NVIDIA A100 80GB PCIe (UUID: GPU-d9e8abeb-4f1a-5cd0-f825-a45f7ea57875)
$ python tools/benchmark_thunder.py --storage ~/workspace --sync main:main main:wjy/bookend
$ pytest-benchmark --storage ~/workspace compare 0027 0028 --group-by name
```

https://gist.github.com/wujingyue/fc11f9725bc510827d9c53061ca05898

0027 -- without the PR
0028 -- with the PR

Below are the nvprof traces of `test_nanogpt_layer_norm[inference-thunder]` without and with this PR. 

Without the PR: 

without-2815

With the PR:

with-2815

PR #2854, opened 2024-08-27
Fix the computation of input_discontig_strides_.

wujingyue:


wujingyue:
!build

csrc/scheduler/registry.cpp
```diff
@@ -81,14 +81,16 @@ SchedulerRuntimeInfo::SchedulerRuntimeInfo(
...
       input_discontig_strides_[fusion_inp] = {};
-      int64_t dims = (int64_t)alloc_strides.size();
+      auto dims = static_cast<int64_t>(alloc_strides.size());
       int64_t expected_stride = 1;
       for (int64_t dim = dims - 1; dim >= 0; dim--) {
         auto size = alloc_sizes.at(dim);
-        if (size <= 1) {
+        auto stride = alloc_strides.at(dim);
+        // Skip broadcast dimensions because they don't affect contiguity.
+        if (size <= 1 || stride == 0) {
```
wujingyue, line 90: @jjsjann123 would you mind clarifying https://github.com/NVIDIA/Fuser/pull/2815/commits/3ff26bccab7fdcfb333a5043f8e16d19d98284d6#r1731934823? I don't follow the difference between isBroadcast and checking the size and the stride. I expect enough validations elsewhere to ensure a broadcast dimension must be size-1 and/or stride-0.

jjsjann123: > I expect enough validations elsewhere

That's a fair ask. Looks like we are safe from the python side. https://github.com/NVIDIA/Fuser/blob/742f24c9c00c8a25d631751e7c8a05a2fa675e55/csrc/python_frontend/python_bindings.cpp#L233-L235

earlier I was concerned about the c++ API. I think user could present a set of inputs with stride == 0 without marking the dimension as broadcast. Semantically, they are not promising a dimension is broadcast, though the execution input could be interpreted as such.
So feels like the broadcast definition is a more robust check, instead of checking the explicit size/stride, which is only retrieved from the runtime inputs.

jjsjann123: I don't have a super strong stance on this one from the python use case. I'm happy with putting a NOTE in the comment and proceed forward with this as-is.

This does feel a bit scary on the c++ side though....

jjsjann123: BTW, not saying it's your modification that introduced the problem, the old code is using `size <= 1` here, so the crime has already been committed. :laughing:

jjsjann123 (approved):
I'm worried about the c++ API's safety here. But I don't want to block this PR for that.

I'll try to play with a repro and see if the c++ API can be broken and follow up with an update (either a clear signal or an nvfuser issue). Let's merge this PR as-is since it resolves the regression for you.

jjsjann123:
For my own reference:
```
TEST_F(NVFuserTest, Playground) {
  FusionGuard fg(fusion.get());

  auto tv0 = TensorViewBuilder().ndims(3).shape({-1, -1, -1}).contiguity({false, false, true}).build();
  fusion->addInput(tv0);

  auto tv1 = relu(tv0);
  fusion->addOutput(tv1);

  FusionExecutorCache executor_cache(std::move(fusion));

  // at::Tensor t0 = at::randn({1024, 4, 128}, options);
  at::Tensor t0 = at::randn({1024*128*4}, options).as_strided({1024, 4, 128}, {128*4, 0, 1});
  auto cg_outputs = executor_cache.runFusionWithInputs({t0});
  testValidate(executor_cache.fusion(), cg_outputs, {t0}, __LINE__, __FILE__);
  
  at::Tensor t0_v2 = at::randn({1024*4*129*2}, options).as_strided({1024, 4, 128}, {4*128, 129, 1});
  auto outs2 = executor_cache.runFusionWithInputs({t0_v2});
  testValidate(executor_cache.fusion(), outs2, {t0_v2}, __LINE__, __FILE__);
}
```

PR #2856, opened 2024-08-27
Add Reduction, Squeeze, Reshape, and Expand to FusionTranslation

rdspring1:
This PR adds support for Reduction, Squeeze, Reshape, and Expand operations.

Details:
- Add `squeeze_expanded` field to `SqueezeOpRecord`. The flag is enabled by default during translation to avoid saving an extra field.
- Depends on #3142


rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

csrc/python_frontend/translation_utils.cpp
```diff
@@ -277,4 +277,52 @@ GET_FUNCTION_TERNARY_SPECIALIZATION_DEFINITION(
     Val*)
 GET_FUNCTION_TERNARY_SPECIALIZATION_DEFINITION(Val*, Val*, Val*, Val*)
 
+std::string getString(const ReductionOp* rop) {
+  switch (rop->getReductionOpType()) {
+    case BinaryOpType::Add:
+      return "sum";
+      break;
+    case BinaryOpType::Mul:
+      return "mul";
```
jjsjann123, line 286:
```suggestion
      return "prod";
```

csrc/python_frontend/translation.cpp
```diff
@@ -635,6 +709,55 @@ class FusionTranslator : public OptInConstDispatch {
...
+    }
+
+    // Always squeeze_expanded dimensions
+    Tensor output = fd_->defineTensor(sop->out()->as<TensorView>()->nDims());
+    map_val_to_fd_index_.emplace(sop->out(), output());
+    fd_->defineRecord(new SqueezeOpRecord(
+        {fd_->recordingState(map_val_to_fd_index_.at(sop->in()))},
+        {fd_->recordingState(output())},
+        squeeze_dims,
+        /*squeeze_expanded=*/true));
```
jjsjann123, line 729: This feels a bit strange here.

SqueezeOp doesn't have a field for `squeeze_expanded`. It reads like we added this field here just to avoid an assert during the translation.

Is there a reason that we wanted to disable squeezing expanded dimensions in the first place?

rdspring1:
!build

jjsjann123 (commented):
Let's bump the version number here since we are changing existing API.

I don't know what's the reason behind the behavior on disable squeeze expanded by default. If we change that behavior I think we would be able to get away without the plumbing on squeeze

csrc/python_frontend/fusion_record.h
```diff
@@ -678,28 +678,31 @@ struct SqueezeOpRecord : RecordFunctor {
...
+  //! | squeeze_expanded? | Squeeze Dim hash        |
   size_t hash() const final {
     auto result = RecordFunctor::hash();
     size_t squeeze_dims_hash = 0;
     for (auto dim : dims_) {
       squeeze_dims_hash ^= static_cast<size_t>(dim);
     }
-    result = result | (squeeze_dims_hash & 0xffffffff);
+    result = result | (squeeze_dims_hash & 0x7fffffff);
+    result |= ((static_cast<size_t>(squeeze_expanded_) & 0x1) << 31);
```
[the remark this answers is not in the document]
rdspring1: No good reason.

csrc/python_frontend/translation.cpp
rdspring1: IIUC, for reduction operations, we want to squeeze expanded dimensions to avoid actually computation.

rdspring1: Alternative: Check if the squeeze occurs on any expanded dimensions. Then map to a reduction operation. You wouldn't know which reduction operation `(sum, prod, min, max)` was originally specified.

jjsjann123: > for reduction operations, we want to squeeze expanded dimensions to avoid actually computation.

I see what you mean now. So we still want to disable squeeze on expanded dimensions in general, since we want to raise exception when that occurs, similar to how we treat squeeze on non-broadcast dimensions.

I'm fine with the approach here as-is. Since this is only used as a check in the operation. Let's put in a comment here mentioning why it's ok to have this field default-on

csrc/python_frontend/fusion_record.h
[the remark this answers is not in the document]
rdspring1: I just followed the same pattern as other bool parameters.

e.g., `is_cpu_` for `TensorRecord`
```cpp
  //! Child specific hash function in lower 32 bits.
  //! |  31  | 30 --- 24 | 23 --------- 12 | 11 ------------------------  0 |
  //! | CPU? | Dtype     | Symbolic Sizes  | Contiguous Info & stride_order |
    result |= ((static_cast<size_t>(is_cpu_) & 0x1) << 31);
```

jjsjann123 (approved):
LGTM

rdspring1:
!build

PR #2857, opened 2024-08-27
Move `clear_cuda_cache` to `nvfuser` module

Priya2698:
This PR moves the utility function `clear_cuda_cache` from `benchmarks` to `nvfuser` module. This function is now also used in the `tests` module which causes some import discrepancies. nvFuser wheel does not bundle `benchmarks` module, and hence, cannot run `test_ops.py` directly there (CC: @xwang233).

### DIFF
--- a/nvfuser/pytorch_utils.py
+++ b/nvfuser/pytorch_utils.py
@@ -62,9 +62,6 @@ def get_device_properties() -> Tuple[int, float]:
     """
     Computes device properties using ctypes and cuda.
     Note: Consider using CUDA-Python when CUDA support >= 12.0.
-    Loading libraries will raise errors on non-CUDA machines, so this
-    function should be called within other functions where this info is needed
-    instead of having a global variable.
     """
     libnames = ("libcuda.so", "libcuda.dylib", "nvcuda.dll", "cuda.dll")
     for libname in libnames:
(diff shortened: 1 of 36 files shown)


xwang233 (commented):
thanks for the fix, looks good, let's do a `!build --pybench` when you're ready

Priya2698:
!build --pybench

Priya2698:
!build --pybench

nvfuser/pytorch_utils.py
```diff
@@ -54,3 +56,124 @@ def patch_codegen_so():
         os.path.join(src_dir, "libnvfuser_codegen.so"),
         os.path.join(dst_dir, "libnvfuser_codegen.so"),
     )
+
+
+def get_device_properties() -> Tuple[int, float]:
```
xwang233, line 61: I don't know how heavy this function is. If we will call `clear_cuda_cache` many times, shall we add a `@functools.lru_cache(None)` to this function?

xwang233 (approved):
Thanks for working on this. This generally looks good to me.

nvfuser/pytorch_utils.py
Priya2698: That's an execellent point. It will be called a high number of times -- which is why I initially had it as a global variable computed once.
Where can I see the benchmark wall time -- that can give us an estimate how huge the overhead is.

Priya2698:
!build --pybench

PR #2860, opened 2024-08-27
Contiguous predicate indexing

naoyam:
This is a follow-up to #2752 and concludes contiguous indexing support in the new indexing system. The previous only supported tensor indexing. This one extends that for predicate indexing. The primary differences with the tensor indexing case are:

- In predicate indexing, all logical domains can be considered contiguous. Ordering of merge doesn't matter, either. This is also the same with the current indexer
- However, contiguity analysis needs to take non-divisible splits into consideration as it's invalid to do contiguous indexing through non-divisible splits. This is also the same with the current indexer.


naoyam:
!build

csrc/id_model/indexing.cpp
```diff
@@ -316,6 +316,8 @@ class AllocationDomainSetup : private kir::IrVisitor {
       }
     }
 
+    NVF_ERROR(allocation_domains.size() == contiguity.size());
```
naoyam, line 319: Not related. Just for extra safety.

csrc/id_model/indexing.cpp
```diff
@@ -402,6 +404,9 @@ class AllocationDomainSetup : private kir::IrVisitor {
       actual_contiguity.push_back(contig.value());
     }
 
+    NVF_ERROR(actual_allocation_domains.size() == actual_strides.size());
```
naoyam, line 407: Not related. Just for extra safety.

csrc/id_model/indexing.cpp
```diff
@@ -756,8 +761,7 @@ TensorIndexer::TensorIndexer(IdModel& id_model) : id_model_(id_model) {
 
   if (isDebugDumpEnabled(DebugDumpOption::IndexingVerbose)) {
     std::ofstream ofs("indexing_traversal_graph.dot", std::ofstream::trunc);
-    auto dot_string =
-        id_model_.idGraph(IdMappingMode::ALMOSTEXACT).toGraphvizDotGraph();
+    auto dot_string = traversalGraph().toGraphvizDotGraph();
```
naoyam, line 764: Just cleanup

csrc/id_model/indexing.cpp
```diff
@@ -937,22 +941,24 @@ Val* TensorIndexer::getLinearIndex(
       getContigIndexFor(expr, as_consumer, alloc_info, for_loops);
 
   // Linearize the indices with strides.
-  Val* index = tv->fusion()->zeroVal();
+  Val* linear_index = tv->fusion()->zeroVal();
```
naoyam, line 944: Just cleanup

naoyam:
!build

csarofeen:
@naoyam looks like this is pending a review, is there someone else that can review? Or do we need to ping Jacob on this?

naoyam:
Pinging @jacobhinkle

csrc/id_model/contiguity.cpp
```diff
@@ -123,6 +151,26 @@ void ContigIDGroups::handle(Merge* merge, Direction direction) {
   contig_ids_.emplace(graph_.toGroup(merge->out()));
 }
 
+// Avoid contiguous indexing if going through non-divisible
+// splits. Not all non-divisible splits need specific predicates, so
+// this condition could be relaxed.
+void ContigIDGroups::handle(Split* split, Direction direction) {
+  if (direction == Direction::Forward) {
```
jacobhinkle, line 158: Why is this only applied to the forward direction?

tests/cpp/test_indexing.cpp
```diff
@@ -4850,6 +4851,233 @@ TEST_F(ContigIndexingTest, ConcretizedBroadcastMerge) {
...
+
+  // Merge first one of the split outputs with the innermost. Since
+  // the split for the reshape is guaranteed to be divisible, the
+  // output domain of the last merge should be the one to predicate
+  tv1->merge(1, 2);
+  tv1->merge(0, 1);
+
+  // While the transformations are the same as tv1, nvFuser doesn't
+  // know the corresponding split is actually divisible, so it's
+  // considered non-divisible. The resulting predicaion should be done
```
jacobhinkle, line 5006:
```suggestion
  // considered non-divisible. The resulting predication should be done
```

tests/cpp/test_indexing.cpp
```diff
@@ -4850,6 +4851,233 @@ TEST_F(ContigIndexingTest, ConcretizedBroadcastMerge) {
...
+  // with each of the two logical domains.
```
jacobhinkle, line 5007: Off-topic but it's too bad we can't attest that the splits done during a reshape are always divisible. Actually I guess the existence of a split along a root->logical path in a `ViewOp` consumer could actually be used to determine that the `ExprGroup` of the split is divisible...

jacobhinkle (approved):
Sorry for the delay

csrc/id_model/contiguity.cpp
naoyam: The problem with split is traversing backward may exceed the extent of the split input. That's why forward splits are checked here (note that this analysis looks at the indexing traversal in a reverse order). When traversing a forward split, there is no possibility of the two output domains to get indices exceeding their respective extents, so that's why only Forward is sufficient here.

jacobhinkle: Makes sense. Thanks for the explanation.

tests/cpp/test_indexing.cpp
naoyam: We do use that information by collecting all splits used in reshape transformations. Here, it is a rather synthetic case. The reshape split is:

```
Outer split: iS17{( (( (( getMetaData(T0) )).logical_size ))[0] )}rf by factor 2 -> iS5{2}rf, iS6{( ceilDiv(( (( (( getMetaData(T0) )).logical_size ))[0] ), 2) )}rf
```

The split of tv2 is:

```
Split: iS7{i0} by factor 5 -> iS11{( ceilDiv(i0, 5) )}, iS12{5}
```

As you can see, they are effectively the same operation, but our analysis is not comprehensive enough to detect the equality. I'm not sure if it's worthwhile to extend the divisibility analysis unless there's any actual use case.

jacobhinkle: Ah that's right. The `getAllDivisibleSplits` lowering pass does what I was saying and fills the `divisible_splits_` set that's used here. However as you say proving a divisible outer split is equal to the corresponding inner split is another matter.

naoyam: Added a clarifying comment.

PR #2864, opened 2024-08-28
calculate SOL using 1G = 1e9 instead of 1024^3

liqiangxl:


xwang233 (approved):
LGTM

naoyam (changes requested):
As I mentioned offline, I don't think this is the right fix.

xwang233 (changes requested):
we shouldn't modify metrics in place while still using the same name

[placeholder for request changes]

Priya2698:
@liqiangxl is out of office, but if we have come to a conclusion we can merge this PR. From the offline discussion, it seemed we will have the existing field, and add the `Bandwidth_v2` as a new field that uses 1e9. @naoyam, @xwang233 do you agree with this?

naoyam:
I'm good as long as we eventually remove `Bandwidth_v2` and just keep `Bandwidth` using `1e9`.

Priya2698:
> I'm good as long as we eventually remove `Bandwidth_v2` and just keep `Bandwidth` using `1e9`.

@xwang233, if we choose to have these two variables now, then we will eventually still modify the metric computation while keeping the same name. `Bandwidth_v2` is a transition period (a month?) and finally `Bandwidth_v2 -> Bandwidth`. Do you agree with this plan?

xwang233:
I can't agree to post data with different meanings into database with same name.

naoyam (approved):
LGTM

xwang233 (approved):

liqiangxl:
!build

PR #2867, opened 2024-08-28
Remove isProducerOf/isConsumerOf

csarofeen:
Remove isProducerOf/isConsumerOf, replace the one instance. Original definitions were backwards.


naoyam (approved):
Thanks for the cleanup

PR #2872, opened 2024-08-30
Cache exact map to speed up ExpressionEvaluator ctor

jacobhinkle:
This just precomputes an ExactLogicalDomainMap for the dynamic fusion in FusionExecutorCache, so that we don't need to re-build that map every time we call `propagateBoundValuesThroughExactMaps`. This speeds up the dynamic shape overhead due to building
`DynamicTransformConcretizationInfo` considerably.

### DIFF
--- a/csrc/dynamic_transform.cpp
+++ b/csrc/dynamic_transform.cpp
@@ -13,6 +13,7 @@
 #include <fusion.h>
 #include <ir/cloner.h>
 #include <ir/utils.h>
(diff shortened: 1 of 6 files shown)


jacobhinkle:
!build

csarofeen (approved):
LGTM

csarofeen:
Please update description with the case it improves and what the perf improvement is.

PR #2873, opened 2024-08-30
Cache `ir_utils::allTvs` as part of Fusion

jacobhinkle:
This caches `ir_utils::allTvs(fusion)` as `fusion->allTvs()`. The cache is automatically invalidated whenever the TV graph topology changes; this mechanism is the same one used to recompute `Expr` uses automatically.

### DIFF
--- a/csrc/fusion.cpp
+++ b/csrc/fusion.cpp
@@ -94,7 +94,7 @@ IrCloner Fusion::copy(const Fusion* from, Fusion* to) {
         .hide_output = alias_info.hide_output};
   }
 
-  to->all_tv_uses_valid_ = from->all_tv_uses_valid_;
+  to->all_tvs_and_uses_valid_ = from->all_tvs_and_uses_valid_;
   // This should never be true on copy, but copying for completeness.
   to->is_during_update_uses_ = from->is_during_update_uses_;
 
@@ -176,7 +176,7 @@ void Fusion::clear() noexcept {
   managed_data_.clear();
   managed_named_data_.clear();
 
-  all_tv_uses_valid_ = false;
+  all_tvs_and_uses_valid_ = false;
   is_during_update_uses_ = false;
 }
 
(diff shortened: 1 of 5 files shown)


jacobhinkle:
!build

csarofeen:
Topo sort test was unhappy because it uses an unregistered expression and this approach triggers a traversal that needs dispatch. trying again.

csrc/fusion.cpp
```diff
@@ -116,6 +116,14 @@ IrCloner Fusion::copy(const Fusion* from, Fusion* to) {
 
   to->expected_dynamic_smem_bytes_ = from->expected_dynamic_smem_bytes_;
 
+  if (from->all_tvs_ptr_ != nullptr) {
```
naoyam, line 119: nit: avoid unnecessary copy of a stale vector
```suggestion
  if (from->all_tvs_and_uses_valid_) {
```

csrc/fusion.h
```diff
@@ -451,13 +455,14 @@ class NVF_API Fusion : public IrContainer {
...
-  //! Val::uses()
-  void resetTvUses();
+  //! Val::uses(). Also populate the all_tvs_ entry with all the used TVs.
+  void resetAllTvsAndUses();
 
   //! Declare that TensorView uses need to be updated (but don't actually do
   //! the update).
   void invalidateTvUses() {
-    all_tv_uses_valid_ = false;
+    all_tvs_and_uses_valid_ = false;
```
naoyam, line 464: This is an issue with the existing code, but this doesn't seem to be consistently used. We should just either remove it or rename it to something `invalidateAllTvsAndUses` and use it consistently.

naoyam:
!build

csarofeen:
!build

naoyam (approved):
LGTM

naoyam:
!build

csarofeen:
@naoyam let's hold off on merging this. I replaced all the uses of allTvs in the codebase and am now getting a segfault. There must still be something wrong with it.

csarofeen:
> Do we still need to keep ir_utils::allTvs?

I'll cleanup in a follow up.

csarofeen:
!build

PR #2874, opened 2024-08-30
Remove dead code in sync_information.cpp.

wujingyue:


zasdfgbnm (approved):
Surprised that clang-tidy didn't catch this automatically.

Issue #2851, opened 2024-08-26
Improving Scheduling and Lowering of Slice and Concat (Part 1)


naoyam:
This is the overall tracking issue. See the corresponding [design doc](https://docs.google.com/document/d/1VYOYpiAOpchm9aBLZeOiguO1EDRJ2XNvxzjux4ejOHA/edit?usp=sharing) (NVIDIA internal)

naoyam:
Closing as the initial version got merged.

Issue #2850, opened 2024-08-26
Refactor the RAW sync analysis with Idmodel


naoyam:
[`SyncMap`](https://github.com/NVIDIA/Fuser/blob/main/csrc/device_lower/analysis/sync_information.h#L19) analyzes each producer-consumer pair and inserts necessary synchronization primitives such as `__syncthreads()`. It employs an indexing-like analysis to reduce false positives, i.e., [`useSameIndex`](https://github.com/NVIDIA/Fuser/blob/main/csrc/device_lower/analysis/sync_information.cpp#L233). 

It should be possible to greatly simplify the analysis with IdModel. Specifically, I believe that what is necessary is just:

- Take parallelized loop iter domains of the producer and grab their AlmostExact groups. Use the loop promotion map if promoted.
- Take parallelized loop iter domains of the consumer and grab their AlmostExact groups. Use the loop promotion map if promoted.
- If the loop group of a parallelized loop group of the producer is also a parallelized loop group of the consumer with the same parallel type, there's no cross-thread dependency. Nothing needs to be done.
- If not, assert that the dependency can be resolved with a synchronization. If yes, insert the synchronization primitive. If not, throw an error.

Specifically, I think what matters is just the loop domains of a given producer and a consumer. As long as they are the same, i.e., mapped in the `Exact` or `AlmostExact` graph using the same parallel type, no synchronization should be necessary. And, if not, there's a cross-thread data dependency, which either means insertion of a sync primitive such as `__syncthreads` or just an error (for example, we don't support a Local producer and a Local consumer use different TID types). The allocation domains, for example, shouldn't affect anything as they just define the layout of each tensor.

Am I missing anything? Any feedback would be appreciated.

naoyam:
Fixed by #2875

PR #2875, opened 2024-08-30
Use loop promotion and indexing traversal graph to find mismatched parallelization

naoyam:
See #2850
Stacked on #2901 

The old code is still used by default. With `NVFUSER_ENABLE=id_model`, the new analysis is used. It's also used for tensors with non-conventional domains.

This is required for #2851. It also enables previously disabled parallelization of the mismatching reshape test from #2684.

I validated the change by comparing the results between the existing and new analyses with all the tests and benchmarks. The only mismatch was with the mismatching reshape test, for which the existing analysis declared a sync is required, whereas the new one correctly recognizes there's no cross-thread dependency.

### DIFF
--- a/csrc/device_lower/analysis/sync_information.cpp
+++ b/csrc/device_lower/analysis/sync_information.cpp
@@ -7,6 +7,7 @@
 // clang-format on
 #include <device_lower/analysis/index_compute.h>
 #include <device_lower/lower2device.h>
+#include <id_model/indexing.h>
 #include <id_model/indexing_utils.h>
 #include <instrumentation.h>
 #include <ir/utils.h>
@@ -543,19 +544,22 @@ SyncMap::SyncMap(Fusion* fusion) {
 
         ProducerConsumerIndexingInfoCache indexing_info(producer, consumer);
 
-        const auto p2c_map_no_forwarding =
-            BestEffortReplay(
-                consumer->getLoopDomain(),
-                producer->getLoopDomain(),
-                PairwiseLogicalDomainMap(producer, consumer)
-                    .mapProducerToConsumer(),
-                /*replay_forward_id_map=*/{},
-                /*target_forward_id_map=*/{},
-                /*skip_replay_swizzle=*/false,
-                /*skip_target_swizzle=*/false,
-                /*skip_resize=*/false,
-                /*error_on_failure=*/false)
-                .getReplay();
+        // P2C map is required when using the IdModel-based analysis
+        const std::unordered_map<IterDomain*, IterDomain*>
+            p2c_map_no_forwarding = GpuLower::current()->hasIdModel()
+            ? BestEffortReplay(
+                  consumer->getLoopDomain(),
+                  producer->getLoopDomain(),
+                  PairwiseLogicalDomainMap(producer, consumer)
+                      .mapProducerToConsumer(),
+                  /*replay_forward_id_map=*/{},
+                  /*target_forward_id_map=*/{},
+                  /*skip_replay_swizzle=*/false,
+                  /*skip_target_swizzle=*/false,
+                  /*skip_resize=*/false,
+                  /*error_on_failure=*/false)
+                  .getReplay()
+            : std::unordered_map<IterDomain*, IterDomain*>{};
 
         // At this point each parallel type that's present in the consumer or
         // the producer will be present in their corresponding `_parallel_ids`
@@ -679,13 +683,8 @@ SyncMap::SyncMap(Fusion* fusion) {
             continue;
           }
 
-          // Check if the IdModel-based analysis matches with the
-          // existing analysis
-          bool requires_sync = true;
-
-          IterDomain* p_loop = nullptr;
-          IterDomain* c_loop = nullptr;
-
+          // Use the IdModel loop promotion when available. This is
+          // required for tensors with non-trivial loop domains
           if (GpuLower::current()->hasIdModel()) {
             if (producer_ptype == consumer_ptype) {
               // If p_id and c_id are mapped in BestEffortReplay with
@@ -693,7 +692,7 @@ SyncMap::SyncMap(Fusion* fusion) {
               // synchronization.
               if (auto it = p2c_map_no_forwarding.find(p_id);
                   it != p2c_map_no_forwarding.end() && it->second == c_id) {
-                requires_sync = false;
+                continue;
               } else {
                 // Even if not mapped in BestEffortReplay, inlining
                 // may effectively make the producer and consumer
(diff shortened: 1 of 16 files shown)


naoyam:
!build --diff

naoyam:
!build --diff

naoyam:
!build --diff-bench

naoyam:
!build --diff-bench

naoyam:
!build --pybench

naoyam:
@zasdfgbnm Could you take a look and let me know what you think? I refactored the sync analysis using the loop promotion results. For now, to make it easier to verify the results, both the existing and new analyses are used and their results are compared. If they don't agree, it's a signal that we should look into, so it'll throw an exception.

To do the refactoring, I needed `BestEffortReplay` that's truly "best effort". The current version checks if the root-to-logical expressions of a consumer tensor always appear in a corresponding producer tensor. I changed the check optional because all we need is to find out if a producer ID and a consumer ID are derived from mapped root/logical domains with mapped exprs. Alternatively, we could use the "broadcast" graph that would map broadcast and non-broadcast domains in addition to the Exact graph.

Let me know what you think. If it makes sense, I'll prepare a real PR for review.

naoyam:
!build --pybench

csrc/device_lower/analysis/sync_information.cpp
```diff
@@ -662,16 +679,67 @@ SyncMap::SyncMap(Fusion* fusion) {
...
+          IterDomain* c_loop = nullptr;
+
+          if (GpuLower::current()->hasIdModel()) {
+            if (producer_ptype == consumer_ptype) {
+              // If p_id and c_id are mapped in BestEffortReplay with
+              // no broadcast forwarding, they should not require any
+              // synchronization.
+              if (auto it = p2c_map_no_forwarding.find(p_id);
+                  it != p2c_map_no_forwarding.end() && it->second == c_id) {
+                requires_sync = false;
```
zasdfgbnm, line 696: Why do we need this branch? Isn't the second branch itself sufficient?

naoyam: That was my initial thought, but it's actually not enough since the AlmostExact graph doesn't map broadcast and non-broadcast domains. For example, just a simple fusion like this would fail:

```
// t0: [i0]
// t1: [b1]
t2 = t0 + t1
// t2: [i2]
```

Suppose nothing is inlined, so there's no promotion. Since `b1` and `i2` are not mapped, this would result in requiring a synchronization, which isn't the case. `t2` and `t1` use different indices but no sync is needed.

That's why I also mentioned about the broadcast graph. If it were used instead of the AlmostExact graph, that would be sufficient as well.

I think in this case `BestEffortReplay` would be good enough.

zasdfgbnm: If t1 is on smem, and both b1 and i2 are parallelized on TIDx, then we wouldn't need sync? For this case will 1 thread write smem, or all threads write it? Whichever it is, don't we need to wait until all writes are complete before reading?

naoyam: In that case, TIDx would be marked as a redundant parallel type and only one thread should write to smem. You're right a sync would be required, which should be taken care at lines around 580. The way how it's handled doesn't seem ideal to me, but I'm trying to make an incremental improvement.

csrc/device_lower/analysis/sync_information.cpp
```diff
@@ -662,16 +679,67 @@ SyncMap::SyncMap(Fusion* fusion) {
...
+                // cross-thread dependency
+                const auto& id_model = GpuLower::current()->idModel();
+                auto producer_loop_id =
+                    indexing_utils::getLoopPromotion(p_id, id_model);
+                p_loop = producer_loop_id;
+                auto consumer_loop_id =
+                    indexing_utils::getLoopPromotion(c_id, id_model);
+                c_loop = consumer_loop_id;
+                const auto& indexing_traveral_graph =
+                    id_model.idGraph(IdMappingMode::ALMOSTEXACT);
```
zasdfgbnm, line 712: nit: should this be `tensor_indexer.traversalGraph()`? I know they are the same today, but in the future, if we use IEL for indexing, should we use IEL or almost exact?

naoyam: Yes. Since right now it's also allowed to enable just IdModel without enabling the indexer, so `tensor_indexer` may not exist. I'll add some static function to `TensorIndexer`.

csrc/device_lower/analysis/sync_information.cpp
```diff
@@ -492,7 +495,7 @@ SyncMap::SyncMap(Fusion* fusion) {
           producer_redundant_types & (~producer_redundant_use_types);
 
       for (const auto producer_i : c10::irange(producer->nDims())) {
-        auto producer_axis = producer->axis(producer_i);
+        auto producer_axis = producer->getLoopDomain().at(producer_i);
```
naoyam, line 498: Not necessary, but just make it explicit that we are looking at the loop domain

csrc/device_lower/analysis/sync_information.cpp
```diff
@@ -653,6 +673,7 @@ SyncMap::SyncMap(Fusion* fusion) {
                      producer->getLogicalDomain(), {p_id})
                      .empty()) {
               raw_dims.set(producer_ptype);
+              continue;
```
naoyam, line 676: Not related, but I believe this was just missing

naoyam:
!build

zasdfgbnm (approved):

naoyam:
!build

naoyam:
@zasdfgbnm Could you please review again? Looks like my test was not comprehensive and I missed a simple case where a sync is required. I added a test to reproduce it. Specifically, mapping with `BestEffortReplay` doesn't tell no synchronization if a producer domain includes a broadcast and a non broadcast. 

Case 1: 
Producer loop ID: broadcast (which may be produced by merging multiple broadcast domains)
Consumer loop ID: non-broadcast
-> They are not exactly mapped but sync is not necessary as discussed above. 

Case 2:
Producer loop ID: non-broadcast
Consumer loop ID: non-broadcast
-> No sync if they are exactly mapped. This case is covered by [the promotion check](https://github.com/NVIDIA/Fuser/pull/2875/files#diff-1610b83eb56354800656aef7092fde223cb917ad477369b9984ecb516a737823R700).

Case 3:
Producer loop ID: non-broadcast
Consumer loop ID: non-broadcast
-> Sync required if they are not exactly mapped, even when they are mapped by the best effort replay. (See the new test for a simple repro)

The previous version missed case 3. 

I updated the code and did the validation by comparing the new and old analyses. See #2907 for the actual validation.

zasdfgbnm (approved):

PR #2877, opened 2024-08-30
Fix the handling of `b(DID)`.

wujingyue:
According to what we [discussed](https://docs.google.com/document/d/1Lrx_lt09O-POJfIvzTrq-OnfREkYVIbk0ihJzw8amu0/edit#heading=h.366wkzawvgie) on Thursday.

### DIFF
--- a/csrc/evaluator_common.cpp
+++ b/csrc/evaluator_common.cpp
@@ -348,19 +348,21 @@ void PrecomputedValues::bindTensorMetaData(
 
   for (const auto dim : c10::irange(logical_domain.size())) {
     IterDomain* id = logical_domain[dim];
-    auto dim_size = tensor.size(static_cast<int64_t>(dim));
-    if (id->isDeviceDim()) {
-      dim_size = tv->getDeviceMesh().size(id->getParallelType());
-    }
-
-    if (id->hasExpandedExtent()) {
-      Val* extent = id->extent();
-      Val* expanded_extent = id->expandedExtent();
-      bindValue(extent->evaluatorIndex(), 1L);
-      bindValue(expanded_extent->evaluatorIndex(), dim_size);
+    const auto dim_size = tensor.size(static_cast<int64_t>(dim));
+    if (id->isBroadcast()) {
+      // DIDs are ignored.
+      bindValue(id->extent()->evaluatorIndex(), 1L);
+      if (id->hasExpandedExtent()) {
+        bindValue(id->expandedExtent()->evaluatorIndex(), dim_size);
+      }
     } else {
-      Val* extent = id->extent();
-      bindValue(extent->evaluatorIndex(), dim_size);
+      if (id->isDeviceDim()) {
+        bindValue(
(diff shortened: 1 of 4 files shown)


wujingyue:
!build

csrc/ops/utils.cpp
```diff
@@ -323,6 +323,13 @@ IterDomain* newOutputIterDomain(
       continue;
     }
 
+    NVF_ERROR(
```
naoyam, line 326: Was this moved due to the change of the interpretation? Or just for clarity?

tests/cpp/test_multidevice_sharding.cpp
```diff
@@ -230,4 +230,55 @@ TEST_F(MultideviceShardingTest, Issue2758) {
       __FILE__);
 }
 
+TEST_F(MultideviceShardingTest, Broadcast) {
```
naoyam, line 233: Could you add some comments here? For example, what is this test meant to verify? How would this test fail (or not?) if the change of this PR were not done?

csrc/ops/utils.cpp
wujingyue: Without this change, pointwise ops like `set` don't propagate the DID on a broadcast dimension from input to output. Although `b(DID)` is semantically the same as without DID, I think it's better to get the input and the output to agree.

tests/cpp/test_multidevice_sharding.cpp
wujingyue: Done.

naoyam: So, before this PR, if this test is executed with, say, 4 GPUs, the extent of `in->getLogicalDomain()[0]->extent()` and `out->getLogicalDomain()[0]->extent()` would be bound to 4? I'm not sure what would happen then, but I understand that isn't what it should be.

wujingyue: > would be bound to 4?

... would be attempted to, and then hit something like "Tried to bind to a constant value: 1 as 4". Overall, we haven't tested a multidevice fusion with broadcasted inputs. Therefore, the bug and this PR.

naoyam (approved):
LGTM

csrc/evaluator_common.cpp
```diff
@@ -348,19 +348,21 @@ void PrecomputedValues::bindTensorMetaData(
...
-    }
-
-    if (id->hasExpandedExtent()) {
-      Val* extent = id->extent();
-      Val* expanded_extent = id->expandedExtent();
-      bindValue(extent->evaluatorIndex(), 1L);
-      bindValue(expanded_extent->evaluatorIndex(), dim_size);
+    const auto dim_size = tensor.size(static_cast<int64_t>(dim));
+    if (id->isBroadcast()) {
+      // DIDs are ignored.
```
naoyam, line 353: Could you leave a comment why it's ignored? Doesn't need to be here.

wujingyue:
!build

PR #2881, opened 2024-08-31
Improve Dynamic Shape Latency for Inner Normalizations

csarofeen:
On the problem of interest:

With this branch: across 10 executions 133 us per execution
Main: across 10 executions 456 us per execution

Jacob mentioned there might be some failures with his PRs, so need to test and check. I did validate the case I was looking at.


csarofeen:
CC @jacobhinkle

csarofeen:
Everything important from here has been pulled in and merged

PR #2891, opened 2024-09-03
Fix MLP's reference implementation.

wujingyue:
This will make the reference implementation and the actual close enough to pass under testValidate, when `-np 1`. I'm looking into `-np 2` or more as part of #2890. I suspect for that we'll have to in addition simulate the behavior of sharding. However, this change is a strict improvement and seems to be standalone enough to get merged first.


wujingyue:
!build

cowanmeg (approved):

PR #2893, opened 2024-09-03
Skip op checks in checkCanSchedule if we are given a data_cache

jacobhinkle:
During scheduling, no data cache is provided when calling `checkCanSchedule`. However, when getting heuristics for a segment, specifically when we are checking whether we can re-use a compiled runtime for new input shapes, we do pass a `data_cache` so that we can re-use some objects for scheduling. In these cases, we have already passed the compile time checks so the check is redundant. In this PR, we avoid the scheduler-agnostic compile time checks added in #2526 when the `data_cache` is given.


jacobhinkle:
!build

csarofeen:
CC @liqiangxl

PR #2896, opened 2024-09-03
Print sizes when mismatch.

wujingyue:


wujingyue:
!build

Priya2698 (approved):

tests/cpp/validator.cpp
```diff
@@ -97,7 +97,10 @@ void testValidate(
                 static_cast<int64_t>(
                     TensorDomain::noReductions(out_tv->getLogicalDomain())
                         .size()),
-        "Dimensionality mismatch in outputs.");
+        "Dimensionality mismatch in outputs: ",
+        aten_output_tensor.sizes(),
```
Priya2698, line 101: nit: Can we do `Expected ..., got ...`, so that we don't have to remember which one is reference.

wujingyue: Here's a sample error message:
```
C++ exception with description "aten_output_tensor.dim() == fusion_output_tensor.dim() && fusion_outputs[i].dim() == static_cast<int64_t>( TensorDomain::noReductions(out_tv->getLogicalDomain()) .size()) INTERNAL ASSERT FAILED at "/opt/pytorch/nvfuser/tests/cpp/validator.cpp":94, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. Dimensionality mismatch in outputs: [1, 2, 128, 1536] vs [1, 256, 1536]
```
The message is pretty clear already about what's LHS and what's RHS. So I'll stick with the current format for conciseness. 


`NVF_ERROR(==)` takes an LHS and an RHS,

PR #2897, opened 2024-09-04
Further WAR for indexing resize

naoyam:
Design doc: https://docs.google.com/document/d/1IJK1WdS3Lgq9mrXRJauBfICfoPXzeUN9Peu0axvZnAs/edit

This is a workaround necessary for #2851. There's more promising idea discussed in the above design doc, but for the problem at hand right now, this WAR is good enough.

### DIFF
--- a/csrc/id_model/indexing_traversal.cpp
+++ b/csrc/id_model/indexing_traversal.cpp
@@ -25,7 +25,7 @@ IndexingTraversal::IndexingTraversal(
   // This is a WAR for cases like
   // ResizeTest.SliceScheduledLikeProducer. Alternatively, we could
   // have a separate graph for indexing that does not map producer and
-  // consumers in non-unary ops. See PR #???.
+  // consumers in non-unary ops. See PR #2897.
   auto all_ids = consumer_tv->domain()->allIDs();
   std::unordered_set<IterDomain*> all_id_set(all_ids.begin(), all_ids.end());
   for (auto id : all_ids) {
--- a/csrc/id_model/indexing_traversal.cpp
+++ b/csrc/id_model/indexing_traversal.cpp
@@ -1,0 +1,46 @@
+// clang-format off
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2023-present NVIDIA CORPORATION & AFFILIATES.
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+// clang-format on
+#include <id_model/indexing_traversal.h>
+
+namespace nvfuser {
+
+IndexingTraversal::IndexingTraversal(
+    const Expr* expr,
+    const ValGraph& graph,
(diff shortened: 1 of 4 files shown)


tests/cpp/test_resize.cpp
```diff
@@ -3677,4 +3677,61 @@ TEST_F(ResizeTest, Chunk_Uneven) {
...
+      tv2_loop_id,
+      IrBuilder::create<Val>(-1, DataType::Index),
+      IrBuilder::create<Val>(-1, DataType::Index));
+
+  tv2->setLoopDomain({tv2_loop_id});
+
+  for (auto tv : {tv1, tv2}) {
+    tv->split(0, 32);
+    // RAW sync analysis not updated yet
+#if 0
```
naoyam, line 3714: This will be enabled once the sync analysis PR is done

naoyam:
!build

csrc/id_model/indexing_traversal.cpp
```diff
@@ -0,0 +1,46 @@
...
+    auto resize = dynamic_cast<Resize*>(id->definition());
+    if (resize == nullptr) {
+      continue;
+    }
+    auto resize_in = resize->in();
+    if (all_id_set.find(resize_in) == all_id_set.end()) {
+      // ths resize must not be part of the exprs involved for
+      // the domains of consumer_tv
+      continue;
+    }
```
jacobhinkle, line 41: It's good to be safe here, but I wonder how this condition could fire. When would we have a `Resize` whose output is in the `TensorDomain` but its input is not?

tests/cpp/test_resize.cpp
```diff
@@ -3677,4 +3677,64 @@ TEST_F(ResizeTest, Chunk_Uneven) {
...
+    tv->axis(0)->parallelize(ParallelType::BIDx);
+    tv->axis(1)->parallelize(ParallelType::TIDx);
+#endif
+  }
+
+  fusion.print();
+  fusion.printKernel();
```
jacobhinkle, line 3724: We can remove these debug prints before merging.

jacobhinkle (approved):
LGTM

csrc/id_model/indexing_traversal.cpp
naoyam: It's unlikely, but for example, I'm replacing the definition of an ID with a resize:

https://github.com/NVIDIA/Fuser/pull/2898/files#diff-98e2e8f5cb3849584331987b87d7cc114cb88cfddddd2f739ca8b452d92f4eceR3866-R3870

In this case the resize is in the path between the root and loop domains. Technically, it's a valid operation to change the loop domain again, and suppose its loop domain is changed back to its logical. The root domain would still keep the resize as its definition, but it is no longer part of the valid operations of the tensor, so it would need to be filtered out. Alternatively, we could delete the resize definition when the loop domain is changed.

PR #2900, opened 2024-09-04
Add script to generate val consts

Priya2698:
Adds script to reproduce the computation that was used when generating the validation tolerances: 
https://github.com/NVIDIA/Fuser/blob/892b7ac5646b3b873f13f2b2ea1db15fa9b9effb/csrc/validator_utils.h#L25-L46

### DIFF
--- a/tools/generate_validation_tolerances.py
+++ b/tools/generate_validation_tolerances.py
@@ -42,6 +42,8 @@ if __name__ == "__main__":
     assert torch.cuda.is_available(), "A CUDA device is required."
 
     device = torch.cuda.get_device_name(torch.cuda.current_device())
+    # Comparison is made against the next higher precision,
+    # since it seems sufficient and avoids incurring extra memory.
     dtype_to_ref_dtypes = {
         torch.bfloat16: torch.float32,
         torch.float16: torch.float32,
--- a/tools/generate_validation_tolerances.py
+++ b/tools/generate_validation_tolerances.py
@@ -10,7 +10,7 @@ The computation (randn + mul + sum) is repeated a large number of times.
 The base tolerance is the max error seen when comparing the desired precision type with the next higher precision.
 The final tolerances are computed by adding a safety factor of 3/4 to this base tolerance.
 
-The output of this file is a .npy file corresponding to each dtype, containing a {size: base_tolerances} dict.
+The output of this file is a .npy file corresponding to each dtype, containing a {size: safety_factor * base_tolerances} dict.
 
 Note: The script may take a long time for a high number of num_iters, so it may be useful to only run the script for one dtype at a time. Modify `dtype_to_ref_dtypes` accordingly.
 
@@ -18,7 +18,7 @@ To run: python tools/generate_validation_tolerances.py
 To load the files:
     ```
     import numpy as np
-    base_tol_dict = np.load(file_name.npy, allow_pickle=True).item()
+    tol_dict = np.load(file_name.npy, allow_pickle=True).item()
     ```
 """
 
@@ -50,15 +50,21 @@ if __name__ == "__main__":
         torch.float32: torch.float64,
     }
 
+    # The original safety factor used is not known, but reverse-engineering estimates it between 3-5.
+    safety_factor = 4
+
     for dtype, ref_dtype in dtype_to_ref_dtypes.items():
         tolerances = {}
         for size in sizes:
-            print(f"size:{size}")
-            val_tolerance = torch.finfo(torch.double).min
+            print(f"Computing validation constant for size:{size}")
+            base_tolerance = torch.finfo(torch.double).min
             for i in range(num_iters):
-                val_tolerance = max(
-                    val_tolerance, compute_max_error(size, dtype, ref_dtype)
+                base_tolerance = max(
+                    base_tolerance, compute_max_error(size, dtype, ref_dtype)
                 )
-            tolerances[size] = val_tolerance
+            tolerances[size] = safety_factor * base_tolerance
         date = datetime.now().strftime("%Y%m%d%H%M%S")
         np.save(f"validation_consts_{dtype}_{device}_{date}.npy", tolerances)
+        print(
+            f"dtype:{dtype} device:{device} safety_factor:{safety_factor}\n{tolerances}\n"
+        )
--- a/tools/generate_validation_tolerances.py
+++ b/tools/generate_validation_tolerances.py
@@ -1,0 +1,61 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-present NVIDIA CORPORATION & AFFILIATES.
+# All rights reserved.
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""
+This script reproduces the computation used for generating the validation tolerances in csrc/validator_utils.h:
+https://github.com/NVIDIA/Fuser/blob/892b7ac5646b3b873f13f2b2ea1db15fa9b9effb/csrc/validator_utils.h#L25-L46
+
+The computation (randn + mul + sum) is repeated a large number of times.
+The base tolerance is the max error seen when comparing the desired precision type with the next higher precision.
+The final tolerances are computed by adding a safety factor of 3/4 to this base tolerance.
+
+The output of this file is a .npy file corresponding to each dtype, containing a {size: base_tolerances} dict.
+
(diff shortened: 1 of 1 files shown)


tools/generate_validation_tolerances.py
```diff
@@ -0,0 +1,61 @@
...
+    max_error = out.sub(ref).abs().max().to(torch.double).item()
+    return max_error
+
+
+assert torch.cuda.is_available(), "A CUDA device is required."
+
+device = torch.cuda.get_device_name(torch.cuda.current_device())
+dtype_to_ref_dtypes = {
+    torch.bfloat16: torch.float32,
+    torch.float16: torch.float32,
```
wujingyue, line 46: Should their ref dtypes be float64 as well?

naoyam (approved):
LGTM

tools/generate_validation_tolerances.py
```diff
@@ -0,0 +1,61 @@
...
+    for size in sizes:
+        print(f"size:{size}")
+        val_tolerance = torch.finfo(torch.double).min
+        for i in range(num_iters):
+            val_tolerance = max(
+                val_tolerance, compute_max_error(size, dtype, ref_dtype)
+            )
+        tolerances[size] = val_tolerance
+    date = datetime.now().strftime("%Y%m%d%H%M%S")
+    np.save(f"validation_consts_{dtype}_{device}_{date}.npy", tolerances)
```
wujingyue, line 61: Move this into main

```
if __name__ == "__main__":
  <code>
```

tools/generate_validation_tolerances.py
```diff
@@ -0,0 +1,61 @@
...
+
+"""
+This script reproduces the computation used for generating the validation tolerances in csrc/validator_utils.h:
+https://github.com/NVIDIA/Fuser/blob/892b7ac5646b3b873f13f2b2ea1db15fa9b9effb/csrc/validator_utils.h#L25-L46
+
+The computation (randn + mul + sum) is repeated a large number of times.
+The base tolerance is the max error seen when comparing the desired precision type with the next higher precision.
+The final tolerances are computed by adding a safety factor of 3/4 to this base tolerance.
+
+The output of this file is a .npy file corresponding to each dtype, containing a {size: base_tolerances} dict.
```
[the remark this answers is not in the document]
Priya2698: I prefered storing it as a file, since I was experiementing with 1) Safety factors 2) Graphs comparing different devices and existing constants. This is an experimentation script, so I much prefer storing the data. The script also runs for a large time -- so I do not work with it interactively.

tools/generate_validation_tolerances.py
Priya2698: @csarofeen used the next higher precision for the existing tolerances.
We can pivot now. The motivation then was, that it seemed sufficient to not incur extra memory.

tools/generate_validation_tolerances.py
```diff
@@ -0,0 +1,62 @@
...
+    ref = (a.to(ref_dtype) * b.to(ref_dtype)).sum()
+    max_error = out.sub(ref).abs().max().to(torch.double).item()
+    return max_error
+
+
+if __name__ == "__main__":
+    assert torch.cuda.is_available(), "A CUDA device is required."
+
+    device = torch.cuda.get_device_name(torch.cuda.current_device())
+    dtype_to_ref_dtypes = {
```
wujingyue, line 45:
```suggestion
    # Use the next higher precision for the existing tolerances.
It seems sufficient to not incur extra memory. 
    dtype_to_ref_dtypes = {
```

wujingyue (approved):

tools/generate_validation_tolerances.py
```diff
@@ -0,0 +1,64 @@
...
+# All rights reserved.
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""
+This script reproduces the computation used for generating the validation tolerances in csrc/validator_utils.h:
+https://github.com/NVIDIA/Fuser/blob/892b7ac5646b3b873f13f2b2ea1db15fa9b9effb/csrc/validator_utils.h#L25-L46
+
+The computation (randn + mul + sum) is repeated a large number of times.
+The base tolerance is the max error seen when comparing the desired precision type with the next higher precision.
+The final tolerances are computed by adding a safety factor of 3/4 to this base tolerance.
```
wujingyue, line 11:
```suggestion
```
I believe this talks about https://github.com/NVIDIA/Fuser/blob/d5b2a245cf5a5b3d66b931f9054f14cad3416bdc/csrc/validator_utils.cpp#L23 and is probably outdated. Even if it's up to date, it can easily get outdated because this comment is too far from validator_utils.cpp and nobody will remember to update here when code changes. Therefore, I recommended we remove this comment or make it vague (and less likely to get outdated) by saying the final tolerances will be computed by getTolerances based on this.

Priya2698: No, this is not what is being referred to. This script is solely to generate the constants listed in `csrc/validator_utils.h`: https://github.com/NVIDIA/Fuser/blob/892b7ac5646b3b873f13f2b2ea1db15fa9b9effb/csrc/validator_utils.h#L25-L46. This is unrelated to `getTolerance` logic.

If the max error seen is `x`, then the constant listed in the file are `safety_factor * x`.
`getTolerances` does an analysis to figure out what is the reduction size.

wujingyue: Who does `safety_factor * x`? I failed to find it in this script. Yet another script?

tools/generate_validation_tolerances.py
```diff
@@ -0,0 +1,64 @@
...
+    import numpy as np
+    base_tol_dict = np.load(file_name.npy, allow_pickle=True).item()
+    ```
+"""
+
+import numpy as np
+import torch
+from datetime import datetime
+
+sizes = [2**i for i in range(2, 22)]  # {4, 2097152}
```
wujingyue, line 29: Question: IIRC, the reduction size is computed with respect to fusion inputs, so it tends to grow very fast. For a two-layer MLP (https://github.com/NVIDIA/Fuser/pull/2905), the reduction size of the second linear output is already `4 * hidden_size * hidden_size`, close to 2M. I'd imagine the whole transformer block will generate an even larger reduction size, much larger than the max size specified here. What's the implication of getTolerance querying a size larger than the max?

tools/generate_validation_tolerances.py
Priya2698: I can include it in this script as well. The reason for keeping it separate was, because, I don't actually know the safety factor used at the time of generating these numbers (it is not documented). When I reverse-engineered it, it would fall between 3-5.

I was using this script as follows:
- Generate the base tolerances
- Interactively experiment with different safety factors and run the tests to see how they impact validation.

Note that this script is a reproduction based on the comments to generate these validation constants for experimentation (I believe we do not have the actual script that was used). The numbers in this file have not been changed.

tools/generate_validation_tolerances.py
Priya2698: It would use twice the max error in the list(https://github.com/NVIDIA/Fuser/blob/1158543c9bd16190fe5820b41a7f4c7e891bc5d9/csrc/validator_utils.cpp#L134-L138). So far, for the unit tests, these existing sizes have been sufficient.
If we have very few examples with larger (~2M) reduction sizes, it would be simpler to set a threshold manually. If we have several such examples, we may benefit from having more cases.

tools/generate_validation_tolerances.py
wujingyue: Got it -- thanks for clarification. 

I'm all for checking in everything needed to reproduce the tolerances. That's the only way to avoid tribal knowledge. I understood that the **current** tolerances in `validator_utils.h` are no longer reproducible -- we lost the script. But let's try our best to avoid that for the **future** ones. 

For this PR, if you think a safety factor will be necessary to generate the tolerances to be directly used by validator_utils.h, code that up in the script and check it in. Even if you are not sure what's the right value, put your best guess and iterate.

tools/generate_validation_tolerances.py
wujingyue: > If we have several such examples, we may benefit from having more cases.

In case you are looking for concrete use cases, I believe https://github.com/NVIDIA/Fuser/blob/main/tests/cpp/test_multidevice_transformer.cpp#L569-L578 will give you reduction sizes much larger than 2M, when you change validation to use testValidate.

cowanmeg: The other issue with MLP is we have compounding error from the running consecutive ops

PR #2901, opened 2024-09-04
Allow silently ignore missing root-to-logical ops in BestEffortReplay

naoyam:
Extracted from and required for #2875


naoyam:
!build

zasdfgbnm (approved):

PR #2905, opened 2024-09-04
Fix reference implementations and improve validation. 

wujingyue:
1. Fix some reference implementations to cast in the same way as nvFuser. 
2. Allow `validate` to take tolerances. 
3. Refine the tolerances so the tests pass up to 6 GPUs.

### DIFF
--- a/csrc/ops/composite.cpp
+++ b/csrc/ops/composite.cpp
@@ -276,24 +276,23 @@ TensorView* tanh_gelu_backward(TensorView* dy, TensorView* x) {
 
   auto x_sq = mul(x, x);
   auto x_cube = mul(x, x_sq);
+  auto beta = IrBuilder::create<Val>(kBeta);
+  auto kappa = IrBuilder::create<Val>(kKappa);
+  auto one = IrBuilder::create<Val>(1.0);
+  auto half = IrBuilder::create<Val>(0.5);
 
-  auto inner_1 = mul(IrBuilder::create<Val>(kKappa), x_cube);
-  auto inner_2 = add(x, inner_1);
-  auto inner_3 = mul(IrBuilder::create<Val>(kBeta), inner_2);
-  auto tanh_inner = tanh(inner_3);
+  auto inner = mul(beta, add(x, mul(kappa, x_cube)));
+  auto tanh_inner = tanh(inner);
 
-  auto left = mul(IrBuilder::create<Val>(0.5), x);
-  auto right = add(IrBuilder::create<Val>(1.0), tanh_inner);
+  auto left = mul(half, x);
+  auto right = add(one, tanh_inner);
 
-  auto left_derivative = mul(IrBuilder::create<Val>(0.5), right);
+  auto left_derivative = mul(half, right);
 
-  auto tanh_inner_sq = mul(tanh_inner, tanh_inner);
-  auto tanh_derivative = sub(IrBuilder::create<Val>(1.0), tanh_inner_sq);
-
-  auto constant_mul_x_sq =
-      mul(IrBuilder::create<Val>(kBeta * 3 * kKappa), x_sq);
-  auto inner_derivative = add(IrBuilder::create<Val>(kBeta), constant_mul_x_sq);
-  auto right_derivative = mul(left, mul(tanh_derivative, inner_derivative));
+  auto tanh_derivative = sub(one, mul(tanh_inner, tanh_inner));
+  auto inner_derivative =
+      mul(beta, add(one, mul(mul(IrBuilder::create<Val>(3.0), kappa), x_sq)));
+  auto right_derivative = mul(mul(left, tanh_derivative), inner_derivative);
 
   auto dx = mul(dy, add(left_derivative, right_derivative));
   return dx;
--- a/tests/cpp/test_multidevice_transformer.cpp
+++ b/tests/cpp/test_multidevice_transformer.cpp
@@ -103,7 +103,8 @@ void validate(
     };
 
     EXPECT_TRUE(outputs[i].allclose(expected_outputs[i], rtol, atol))
-        << "Output " << i << " mismatches:" << std::endl
+        << "Output " << i << " mismatches with rtol " << rtol << " and atol "
+        << atol << ":" << std::endl
         << generate_comparison_details(
                expected_outputs[i], outputs[i], rtol, atol);
   }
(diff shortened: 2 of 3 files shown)


wujingyue:
!build

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -104,10 +106,9 @@ std::vector<at::Tensor> reference_mlp(
     at::Tensor w1,
     at::Tensor b1,
     at::ScalarType at_dtype) {
-  auto linear0 = at::matmul(x, w0).to(at::kFloat) + b0.to(at::kFloat);
-  auto gelu = at::gelu(linear0, "tanh");
-  auto linear1 =
-      at::matmul(gelu.to(at_dtype), w1).to(at::kFloat) + b1.to(at::kFloat);
+  auto linear0 = at::matmul(x, w0).to(at::kFloat) + b0;
```
wujingyue, line 109: A no-op change. `+` in PyTorch promotes data types.

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -104,10 +106,9 @@ std::vector<at::Tensor> reference_mlp(
...
+  auto gelu = at::gelu(linear0, "tanh").to(at_dtype);
```
wujingyue, line 110: This is necessary for `at::allclose` to pass because `mlp` generates the output of `at_dtype`.

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -401,11 +402,21 @@ TEST_P(DistributedTransformerTest, MLP_Layer) {
...
       reference_outs[3]};
 
-  MultiDeviceExecutor runtime(
-      std::move(fusion), *communicator_, executor_params_);
+  FusionExecutorCache fec(std::move(fusion));
   at::manual_seed(getATenRandomSeed());
-  auto outputs = runtime.runWithInput(inputs);
-  validate(expected_outputs, outputs);
+  auto outputs = fec.runFusionWithInputs(inputs);
+  ASSERT_EQ(outputs.size(), expected_outputs.size());
```
cowanmeg, line 408: Why not just use the existing validate and change the tolerance?

wujingyue: I'll try to do that later. This PR only fixes the ref of MLP, and `validate` is used by other tests as well.

cowanmeg: you could parameterize it with the lower atol. I imagine if we want tighter bounds for all the tests each one will have it's own value that is hard to predict without manually trying

cowanmeg:
!build

csrc/ops/composite.cpp
wujingyue, line 1: I suspected some inaccuracy and made these changes. Although they turned out not to affect accuracy, I left them in the PR as general cleanups.

wujingyue:
!build

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -56,33 +57,38 @@ class DistributedTransformerTest
...
     // not showing which output was problematic.
-    ASSERT_EQ(out[i].dtype(), expected_out[i].dtype())
+    ASSERT_EQ(outputs[i].dtype(), expected_outputs[i].dtype())
         << "Output " << i << " has a mismatching data type.";
 
     // Note: Scaling tolerance up since the error accumulates across ops
     // BFloat16 error is quite high, but the program has been verified with
     // double precision to be logically correct.
-    const double atol = 0.075 * (i + 1);
-    const double rtol = 1.6e-2;
```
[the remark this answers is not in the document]
cowanmeg: It would also be beneficial to keep some form of default value, since tests are still getting added this adds to a lot of effort to get a simple example working.

csrc/ops/composite.cpp
```diff
@@ -276,24 +276,23 @@ TensorView* tanh_gelu_backward(TensorView* dy, TensorView* x) {
...
+  auto inner = mul(beta, add(x, mul(kappa, x_cube)));
+  auto tanh_inner = tanh(inner);
 
-  auto left_derivative = mul(IrBuilder::create<Val>(0.5), right);
+  auto left = mul(half, x);
+  auto right = add(one, tanh_inner);
 
-  auto tanh_inner_sq = mul(tanh_inner, tanh_inner);
-  auto tanh_derivative = sub(IrBuilder::create<Val>(1.0), tanh_inner_sq);
+  auto left_derivative = mul(half, right);
```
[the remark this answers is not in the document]
wujingyue: It's not necessary and I'm happy to revert it. I kept it for only two reasons: 
1. Code becomes shorter. 
2. Code matches [the PyTorch implementation](https://github.com/pytorch/pytorch/blob/e72ed4717eab2c3d6c54a93537f5b298f33dff21/aten/src/ATen/native/cuda/ActivationGeluKernel.cu#L49-L66) more closely, which I hypothesized would help accuracy.

tests/cpp/test_multidevice_transformer.cpp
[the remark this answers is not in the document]
wujingyue: I removed it because I found I ended up always setting rtol to 0 :) 

I'd love to keep some default value to, as you said, minimize the effort to get a simple example working. What would be a good default value? The old `atol=0.075*(i+1)` is too relaxed after I fixed the reference implementation. The `*(i+1)` part is also problematic because the output number doesn't necessarily match the layer number. 

How about this: 
1. Hardcode rtol to 0.016. I'll try to finetune this value a little bit -- 0.016 sounds like a large **default** rtol to start with. 
2. Still require the caller to provide a list of per-output atols, because it's hard to fine a good default and each test seems to require something different.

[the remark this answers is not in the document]
cowanmeg: It was supposed to have been taken from Pytorch. https://github.com/pytorch/pytorch/blob/042f2f7746a064f1527d95d1f1d712b4f0b34186/test/test_transformers.py#L85
But should be .0126.

wujingyue:
!build

tests/cpp/test_multidevice_transformer.cpp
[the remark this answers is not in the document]
wujingyue: Done -- added default rtols.

cowanmeg (approved):

PR #415, opened 2023-05-27
Fix vectorization analysis

zasdfgbnm:
This PR is stacked on https://github.com/NVIDIA/Fuser/pull/414, please review that PR first.

Fixes: https://github.com/NVIDIA/Fuser/issues/413. I checked kernels in benchmarks, no vectorization factor change.

I think the entire equation used to compute PE through merge and split are wrong. I rewrite it. For example in `I1{4}*I2{6}`, if the PE of `I1*I2` is `10`, then `I2` shouldn't have full PE `6` even if `10 > 6`, instead, it should have partial PE `gcd(10, 6) = 2`.

Some changes:
- class `ProjectedExtent` is removed. Now we are using a single `Val*` to store PE.
- `propagateExtentSplitBackward` and `propagateExtentMergeForward` are replaced with `combinePE`.
- `propagateExtentMergeBackward` and `propagateExtentSplitForward` are replace with `distributePE`.
- `getAllVectorizedMapsOf` is renamed as `getTvToContigMergeOfInnerSizeMap`, with some refactor
- `getContigVectorSizesOf` is renamed as `getContigMergeOfInnerSize`, it now returns a product of `Val`s instead of vector of `Val`s.

PR #2911, opened 2024-09-05
Add utility `lower_utils::proveLinearAndGetStride`

zasdfgbnm:
Add utility `lower_utils::proveLinearAndGetStride`, which proves that the provided `linear_g` is linear with respect to the given `domain`, and returns the linear coefficient (stride). This utility will be used in the lowering of mma to fill in the "Leading dimension byte offset" and "Stride dimension byte offset" in the matrix descriptor.

See:
https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#asynchronous-warpgroup-level-matrix-shared-memory-layout-matrix-descriptor
https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#strides


csrc/utils.h
```diff
@@ -594,4 +594,8 @@ T pow(T a, T b) {
   }
 }
 
+template <typename T>
+using MaybeUniqueOwningPtr = dynamic_type::
+    DynamicType<dynamic_type::NoContainers, T*, std::unique_ptr<T>>;
+
```
zasdfgbnm, line 600: End up not using this, but seems useful to just leave it there.

csrc/val_graph_visitor.h
```diff
@@ -183,16 +183,22 @@ class ValGraphBFS : public BFS<
...
   // given graph. Dependency between vals and exprs must be satisfied.
   // It is an error if no valid path is found.
+  static ExprPath getExprsBetween(
+      const ValGraph& graph,
+      std::vector<NodeType> from,
+      std::vector<NodeType> to) {
+    ValGraphBFS bfs(graph, std::move(from), std::move(to));
+    bfs.traverse();
+    return bfs.getShortestExprPath();
+  }
```
zasdfgbnm, line 193: End up not using, but feel that it is helpful to keep.

zasdfgbnm:
!build

csarofeen:
Please add a description and motivation for this PR. "See code" isn't a good PR description.

zasdfgbnm:
> Please add a description and motivation for this PR. "See code" isn't a good PR description.

Added

csrc/device_lower/utils.cpp
```diff
@@ -1013,6 +1016,552 @@ bool predicateAtEnd(ForLoop* loop) {
...
+//
+// We use a dynamic type to be able to represent all these cases.
+template <typename AbstractValGroup>
+struct PartOf {
+  std::shared_ptr<AbstractValGroup> group;
+  Val* inner_extent = nullptr;
+  Val* selected_extent = nullptr;
+};
+
+using AbstractValGroup = dynamic_type::DynamicType<
```
jacobhinkle, line 1102: Would `ValGroupRegion` be a clearer name?

csrc/device_lower/utils.cpp
```diff
@@ -1013,6 +1016,552 @@ bool predicateAtEnd(ForLoop* loop) {
...
 
+// Implementation of:
+//   Val* proveLinearAndGetStride(
+//       const ValGraph& id_graph,
+//       const ValGroup& linear_g,
+//       const ValGroups& domain);
+//
+// The idea is to propagate from linear_g to domain, keep track of how linear_g
+// lives in domain. For example, let's consider the following two schedules
+// (same example as the NVFuserTest.ProveLinearAndGetStride test):
```
jacobhinkle, line 1027: Perhaps it would help to define "linear" in this context first, in terms of `IterDomain`s instead of indices. I believe it's similar in spirit to stride compatibility if domain represents the size of a strided tensor. In that case, `linear_g` being linear in `domain` would correspond to the condition that there exists a "view" of the strided tensor such that `linear_g` is one of the new strided axes. That is, it's possible to represent the index for `linear_g` as a combination of the indices for `domain`. Is this correct?

csrc/device_lower/utils.cpp
```diff
@@ -1013,6 +1016,552 @@ bool predicateAtEnd(ForLoop* loop) {
...
+}
+
+std::string print(const std::monostate&) {
+  return "std::monostate";
+}
+
+std::string print(const AbstractValGroup& group) {
+  return AbstractValGroup::dispatch(
+      [&](const auto& group) { return print(group); }, group);
+}
```
jacobhinkle, line 1143: Should these be `toString` instead of `print`?

csrc/device_lower/utils.cpp
```diff
@@ -1013,6 +1016,552 @@ bool predicateAtEnd(ForLoop* loop) {
...
+
+bool related(const std::monostate& current, const ValGroup& to) {
+  return false;
+}
+
+bool related(const AbstractValGroup& current, const ValGroup& to) {
+  return AbstractValGroup::dispatch(
+      [&](const auto& current) { return related(current, to); }, current);
+}
```
jacobhinkle, line 1169: Did you need to forward-declare this in order for this dispatch call to compile?

zasdfgbnm: Forward declare is not for this function body to compile, but instead, is for the definition of
```C++
bool related(const PartOf<AbstractValGroup>& current, const ValGroup& to)
```
and
```C++
bool related(const std::deque<AbstractValGroup>& current, const ValGroup& to)
```
to compile because it uses `bool related(const AbstractValGroup& current, const ValGroup& to)` in its function body.

zasdfgbnm: in this context, the type of `current` will be one of `const PartOf<AbstractValGroup>&`, or `const std::deque<AbstractValGroup>&`, or `std::monostate`, so this `related(current, to)` will use the above specialized versions, instead of the abstract version `bool related(const AbstractValGroup& current, const ValGroup& to)`.

csrc/device_lower/utils.cpp
zasdfgbnm: renamed

csrc/device_lower/utils.cpp
zasdfgbnm: renamed

zasdfgbnm: No, I have to change it back. `toString` is too heavily overloaded, and C++ have trouble finding the correct one.

csrc/device_lower/utils.cpp
zasdfgbnm: > Perhaps it would help to define "linear" in this context first, in terms of `IterDomain`s instead of indices. I believe it's similar in spirit to stride compatibility if domain represents the size of a strided tensor. In that case, `linear_g` being linear in `domain` would correspond to the condition that there exists a "view" of the strided tensor such that `linear_g` is one of the new strided axes.

Yes.

> That is, it's possible to represent the index for `linear_g` as a combination of the indices for `domain`.

Isn't this always possible regardless of if linear or not?

zasdfgbnm: added something to `utils.h`.

jacobhinkle: > Isn't this always possible regardless of if linear or not?

Not as a linear combination with integer coefficients, right?

csrc/device_lower/utils.cpp
jacobhinkle: Got it. Thanks.

csrc/device_lower/utils.cpp
```diff
@@ -1013,6 +1016,552 @@ bool predicateAtEnd(ForLoop* loop) {
...
+//       /       \  /       \.
+//      /         \/        64.
+//     /          /\       /  \.
+//    /          /  \     /    \.
+//   16        [2]   8   8      8
+//                    \ /
+//                    xor
+//                    / \.
+//                   8   8
+//
```
naoyam, line 1045: Correct me if this understanding is wrong:

```
Inner split of I0 by 128 -> I0o, 128_1
Inner split of I1 by 128 -> I1o, 128_2
Inner split of 128_1 by 8 -> 16, 8_1
Inner split of 128_2 by 64 -> [2], 64
Inner split of 64 by 8 -> 8_2, 8_3
Xor swizzle of 8_1 and 8_2 -> 8_4, 8_5
```

(Used `_` to denote different IDs with the same extent)

naoyam:
> This utility will be used in the lowering of mma to fill in the "Leading dimension byte offset" and "Stride dimension byte offset" in the matrix descriptor.

Do you already have this code? That would be helpful to understand the context more clearly.

csrc/device_lower/utils.cpp
zasdfgbnm: > Not as a linear combination with integer coefficients, right?

This isn't the case for linear as well. For example, if `linear_g` is `I0`, and `domain` is `[I0*4]`, assume the index of `I0*4` is `i`, then the index of `I0` is `i / 4`, which is not a linear function of `i`. I think it is more important that `i` is a linear function of `i0`, not the opposite.

csrc/device_lower/utils.cpp
zasdfgbnm: Yes, exactly

zasdfgbnm:
> Do you already have this code? That would be helpful to understand the context more clearly.

Not yet

csrc/device_lower/utils.cpp
jacobhinkle: If `linear_g` is `I0` and `domain` is `[I0/4, 4]` a divisible split, then would we also consider `linear_g` to be linear in `domain`? Or similarly, if `linear_g` is `I0*4` and `domain` is `[I0, 4]`?

zasdfgbnm: I think for both case, yes.

naoyam:
Could you list up all conditions that would violate linearity?

naoyam:
Another question: Is the stride obtained by the tensor indexer different from the stride returned by this function?

zasdfgbnm:
> Could you list up all conditions that would violate linearity?

For example, if `linear_g` is `I0`, but domain is `I0/4, I1, 4`, then it will not be linear. I believe this is the most common case that is not linear. Another example is: if `linear_g` is `I0`, but domain is `I0/4, 4/32, 32`.

> Another question: Is the stride obtained by the tensor indexer different from the stride returned by this function?

`TensorIndexer` never returns any stride, it can only return the full index. But if it is indeed linear, and the expr simplifier is able to simplify the expression as linear, then the stride in the full index will match with what this function returns.

This utility is going to be used as follows: For example, for the `M64_N256_K16` macro, the loop domain will have three IDs `64, 256, 16` all parallelized on `ParallelType::Mma`. To index A, assuming no swizzle, I will have `(64, 16)` split as `(64/8, 8, 16/8, 8)`, and use this utility to compute the stride of the `64/8` and the `16/8`, these strides will be used to fill the matrix descriptor.

naoyam:
> > Could you list up all conditions that would violate linearity?
> 
> For example, if `linear_g` is `I0`, but `domain` is `I0/4, I1, 4`, then it will not be linear. I believe this is the most common case that is not linear. Another example is: if `linear_g` is `I0`, but `domain` is `I0/4, 4/32, 32`. A third example: `linear_g` is `I0`, but `domain` is `I0/4, xor(4), ...`. 

Sounds like all of the above cases are because of non-divisible splits? 

> In general, I believe there are infinite number of cases that is not linear.

Of course, but I suppose the number of conditions is limited, right? And I suppose what this PR is trying to do is to detect those conditions, so before reading through the code, I'd like to understand what the underlying logic is.

> > Another question: Is the stride obtained by the tensor indexer different from the stride returned by this function?
> 
> `TensorIndexer` never returns any stride, it can only return the full index. But if it is indeed linear, and the expr simplifier is able to simplify the expression as linear, then the stride in the full index will match with what this function returns.
> 

You're right, and I'm sorry that I should have phrased the question differently. I was thinking that, since `TensorIndexer` also computes strides to get linear indices, we could add an API to return non linearized indices with strides.

> This utility is going to be used as follows: For example, for the `M64_N256_K16` macro, the loop domain will have three IDs `64, 256, 16` all parallelized on `ParallelType::Mma`. To index A, assuming no swizzle, I will have `(64, 16)` split as `(64/8, 8, 16/8, 8)`, and use this utility to compute the stride of the `64/8` and the `16/8`, these strides will be used to fill the matrix descriptor.

In this case, the allocation domain of matrix A is `(64/8, 8, 16/8, 8)`? If so, can we calculate the stride of each domain just by iterating from the innermost, i.e., `[128, 16, 8, 1]`?

zasdfgbnm:
> Sounds like all of the above cases are because of non-divisible splits?

No. For example, for the first case, even if the `I0/4` is divisible, the result is still non-linear. Same for the third example.

> Of course, but I suppose the number of conditions is limited, right? And I suppose what this PR is trying to do is to detect those conditions

I don't think the number of non-linear conditions is limited, and I don't think we can just enumerate all these conditions. Proving linearity is a very difficult problem and is far more complicated than just detecting a few patterns. It is as difficult as the vectorization helper: https://github.com/NVIDIA/Fuser/blob/main/csrc/scheduler/vectorize_helper.h

For example, if I have:
```C++
[linear_g] 6 -> split -> (2, 3) -> reorder -> (3, 2) -> merge -> 6 -> split -> (2, 3) -> reorder -> (3, 2) -> merge -> 6 [domain]
```
then it is not linear.

But if instead we have

```C++
[linear_g] 6 -> split -> (2, 3) -> reorder -> (3, 2) -> merge -> 6 -> split -> (3, 2) -> reorder -> (2, 3) -> merge -> 6 [domain]
```
then it is linear

> I was thinking that, since `TensorIndexer` also computes strides to get linear indices, we could add an API to return non linearized indices with strides.

That would mean that we will be relying too much on expression simplifier, which is a restriction that I don't want to have. I am trying to keep expr simplifier an opt-in feature so we can disable it as needed for debugging. This PR also uses expr simplifier, but for the matmul case, constant folding is sufficient, which is not the case if we "just use `TensorIndexer`".

For example, if `linear_g` is `I0`, and `domain` is `[I0/4, 4]` (the split is divisible), if the index of `I0` is `i`, then the domain's index is `i/4*4 + i%4`, this form is not naturally a linear function. We only know it is linear if we simplify expression using Euclid's Division Lemma and simplify it as `i`.

> In this case, the allocation domain of matrix A is `(64/8, 8, 16/8, 8)`? If so, can we calculate the stride of each domain just by iterating from the innermost, i.e., `[128, 16, 8, 1]`?

No, it can be `(64/8, 16/8, 8, 8)`, but not necessarily as easy as this. It could be, for example, `(128/8, 32/8, 8, 8)`, where the 64 is the inner of 128, and the 16 is the inner of 32, and we should be able to know that.

naoyam:
> > Sounds like all of the above cases are because of non-divisible splits?
> 
> No. For example, for the first case, even if the `I0/4` is divisible, the result is still non-linear. Same for the third example.
> 
> > Of course, but I suppose the number of conditions is limited, right? And I suppose what this PR is trying to do is to detect those conditions
> 
> I don't think the number of non-linear conditions is limited, and I don't think we can just enumerate all these conditions. Proving linearity is a very difficult problem and is far more complicated than just detecting a few patterns. It is as difficult as the vectorization helper: https://github.com/NVIDIA/Fuser/blob/main/csrc/scheduler/vectorize_helper.h
> 

I think I finally am getting some understanding of the linearity. Is this analysis effectively the same as the vectorization analysis? In the case of the vectorization analysis, we do the analysis with a symbolic vectorization factor and try to find the maximum factor since the schedule of tensors is not yet determined. Here, we already have transformed tensors and the "factor", i.e., `linear_g`, and try to see if the "factor" corresponds to a linear subregion in `domain`. Does this make sense?

zasdfgbnm:
> I think I finally am getting some understanding of the linearity. Is this analysis effectively the same as the vectorization analysis? In the case of the vectorization analysis, we do the analysis with a symbolic vectorization factor and try to find the maximum factor since the schedule of tensors is not yet determined. Here, we already have transformed tensors and the "factor", i.e., `linear_g`, and try to see if the "factor" corresponds to a linear subregion in `domain`. Does this make sense?

Yes, correct. I would consider this new utility `lower_utils::proveLinearAndGetStride` as a validation of vectorization during lowering. IIRC, we disabled some validation on vectorization during lowering when we implemented the vectorization helper. I indeed plan to spend some of my time checking the disabled validation and see if we can replace it with something like:
```C++
auto stride = lower_utils::proveLinearAndGetStride(id_graph, vectorized_id, allocation_domain);
NVF_CHECK(stride != nullptr, "invalid vectorization");
GpuLower::current()->validate(SimplifyingIrBuilder::eqExpr(stride, 1), "invalid vectorization");
```
but this is mostly for finding low-hanging fruit. If it eventually turns out that we need to largely expand `proveLinearAndGetStride` to make it more comprehensive, then likely I will not continue adding this validation.

naoyam:
> > I think I finally am getting some understanding of the linearity. Is this analysis effectively the same as the vectorization analysis? In the case of the vectorization analysis, we do the analysis with a symbolic vectorization factor and try to find the maximum factor since the schedule of tensors is not yet determined. Here, we already have transformed tensors and the "factor", i.e., `linear_g`, and try to see if the "factor" corresponds to a linear subregion in `domain`. Does this make sense?
> 
> Yes, correct. I would consider this new utility `lower_utils::proveLinearAndGetStride` as a validation of vectorization during lowering. IIRC, we disabled some validation on vectorization during lowering when we implemented the vectorization helper. I indeed plan to spend some of my time checking the disabled validation and see if we can replace it with something like:
> 
> ```c++
> auto stride = lower_utils::proveLinearAndGetStride(id_graph, vectorized_id, allocation_domain);
> NVF_CHECK(stride != nullptr, "invalid vectorization");
> GpuLower::current()->validate(SimplifyingIrBuilder::eqExpr(stride, 1), "invalid vectorization");
> ```
> 
> but this is mostly for finding low-hanging fruit. If it eventually turns out that we need to largely expand `proveLinearAndGetStride` to make it more comprehensive, then likely I will not continue adding this validation.

Thanks! I'm finally ready to review the actual code. 

The relevance with the vectorization analysis was pretty helpful for me. I think that's worth mentioning in the comment. Also, do we want to say "projection" as it is used in the vectorization analysis? I think that well describes what this code is trying to find.

csrc/device_lower/utils.cpp
```diff
@@ -1013,6 +1016,552 @@ bool predicateAtEnd(ForLoop* loop) {
...
+
+// From the above example, we can see that how linear_g lives in domain could be
+// complicated. It can be, for example:
+//   1. linear_g is equivalent to a single ValGroup in domain.
+//   2. linear_g is the inner of a ValGroup in domain.
+//   3. linear_g is the outer of a ValGroup in domain.
+//   4. linear_g is the middle of a ValGroup in domain, where on the right,
+//      there is a 2.
+//   5. linear_g is split as g1, g2, g3 in domain.
+//   6. linear_g is split as the inner 2 of g1, g2, and the outer 4 of g3.
```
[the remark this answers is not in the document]
zasdfgbnm: It does not refer to anything in particular. It just refers to some ValGroups that I give them this name. For example, `linear_g is split as g1, g2, g3 in domain.` could mean something like:
```
 linear_g
 /    \
g1   g23
     / \
   g2   g3
```

zasdfgbnm:
> Also, do we want to say "projection" as it is used in the vectorization analysis?

I think I can rename `ValGroupRegion` as `LinearGroupProjection`.

zasdfgbnm:
> The relevance with the vectorization analysis was pretty helpful for me. I think that's worth mentioning in the comment.

Rephrased the comment before `namespace {` in https://github.com/NVIDIA/Fuser/pull/2911/commits/e8cb5e3d2f7c879ba9920d78bbe3b5aac8fbe8d3. I believe comparing it with vectorization helper and use the term "projection" makes the idea much clearer. Thank you for the suggestion.

csrc/device_lower/utils.cpp
```diff
@@ -1013,6 +1016,552 @@ bool predicateAtEnd(ForLoop* loop) {
...
+//
+// We use a dynamic type to be able to represent all these cases.
+template <typename ValGroupRegion>
+struct PartOf {
+  std::shared_ptr<ValGroupRegion> group;
+  Val* inner_extent = nullptr;
```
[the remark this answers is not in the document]
zasdfgbnm: Added a comment in code before this:
```
  // The extent on the inner of linear_g in group. nullptr means 1, that is,
  // linear_g is the inner of group.
```

csrc/device_lower/utils.cpp
```diff
@@ -1013,6 +1016,552 @@ bool predicateAtEnd(ForLoop* loop) {
...
+  Val* selected_extent = nullptr;
```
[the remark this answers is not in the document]
zasdfgbnm: Added a code comment before it:
```
// The extent of linear_g.
```

csrc/device_lower/utils.cpp
```diff
@@ -1013,6 +1016,552 @@ bool predicateAtEnd(ForLoop* loop) {
...
+    // If linear_g is merged with another group, then part of the merged group
+    // represents linear_g.
+    NVF_ERROR(from.size() == 2);
+    NVF_ERROR(to.size() == 1);
+    NVF_ERROR(from.front() == current || from.back() == current);
+    return PartOf<ValGroupRegion>{
+        std::make_shared<ValGroupRegion>(to.front()),
+        /*inner_extent=*/from.front() == current
+            ? from.back()->front()->as<IterDomain>()->extent()
+            : nullptr,
```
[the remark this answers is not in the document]
zasdfgbnm: See https://github.com/NVIDIA/Fuser/pull/2911#discussion_r1755147121

csrc/device_lower/utils.cpp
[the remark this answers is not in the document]
naoyam: I see. It's still unclear to me what this means:

> linear_g is split as the inner 2 of g1, g2, and the outer 4 of g3.

What's "inner 2 of g1, g2"?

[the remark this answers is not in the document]
zasdfgbnm: Something like this:
```C++
 x  linear_g{2}
  \ /
  g12
  / \
g1  g2
```

csrc/device_lower/utils.cpp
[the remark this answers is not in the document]
zasdfgbnm: The extent refers the size on the inner of `linear_g`, not the innermost dim in `group`. If `linear_g` is the innermost, then there is nothing on the inner of `linear_g`, so it is one.

csrc/device_lower/utils.cpp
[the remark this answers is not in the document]
naoyam: I see. Maybe rephrase to something like this?

```
linear_g, which is an extent-2 domain, is merged as an inner domain and then split to g1 and g2.
```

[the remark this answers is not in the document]
zasdfgbnm: Added detailed explanation for each case as code comment in https://github.com/NVIDIA/Fuser/pull/2911/commits/8585a878ae833fb48e4fe6fe2eabd09620c0a21d

csrc/device_lower/utils.cpp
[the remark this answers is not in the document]
naoyam: Oh, I see. Is it the same as the stride then?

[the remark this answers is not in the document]
zasdfgbnm: Yes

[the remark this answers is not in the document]
zasdfgbnm: If you believe `stride` is a better name, I am happy to change it.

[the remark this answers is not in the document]
naoyam: Either is probably fine. It's just helpful to have `stride` mentioned as well.

csrc/device_lower/utils.cpp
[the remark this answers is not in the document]
naoyam: Hmm, what does "selected" imply then?

csrc/device_lower/utils.cpp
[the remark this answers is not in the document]
naoyam: Now I see it. Thanks.

csrc/device_lower/utils.cpp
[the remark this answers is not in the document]
zasdfgbnm: I view the structure of `group` as:
```
.--------------------------.
| outer | linear_g | inner |
'--------------------------'
```
and "selected" just refers to `linear_g`.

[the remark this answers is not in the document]
naoyam: I see. 

IIUC, `linear_g` won't change throughout this analysis. Why each `PartOf` needs to have its extent?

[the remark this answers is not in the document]
zasdfgbnm: You are right that `selected_extent` will never change. It is just a convenient mechanism to carry this information over.

csrc/device_lower/utils.cpp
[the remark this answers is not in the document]
zasdfgbnm: updated the comment in code.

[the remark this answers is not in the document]
naoyam: Much easier to follow. Thanks!

csrc/device_lower/utils.cpp
```diff
@@ -1013,6 +1016,618 @@ bool predicateAtEnd(ForLoop* loop) {
...
+    const ValGraph& id_graph,
+    const ExprGroup& eg,
+    Direction direction) {
+  auto from = fromGroups(id_graph, eg, direction);
+  auto to = toGroups(id_graph, eg, direction);
+  if (from.size() == 1) {
+    NVF_ERROR(to.size() == 2);
+    NVF_ERROR(related(*current.group, from.front()));
+    auto group = propagate(*current.group, id_graph, eg, direction);
+    if (!group.hasValue()) {
```
naoyam, line 1304: When does this happen? IIUC, we are propagating the group of `current`, which is not `linear_g`.

csrc/device_lower/utils.cpp
```diff
@@ -1013,6 +1016,618 @@ bool predicateAtEnd(ForLoop* loop) {
...
+      return {};
+    }
+    // The result is actually just the
+    //   PartOf{group, current.inner_extent, current.selected_extent}
+    // However, we want to simplify the result. For example, if the inner_extent
+    // is a multiple of the extent of the last item in group, we can simplify
+    // the inner_extent and group by canceling the last item in group.
+    Val* new_inner_extent = current.inner_extent;
```
naoyam, line 1312: This `new_inner_extent` represents the inner extent of `linear_g` in `group`, right? Why doesn't the inner extent change? For example, if the expr is a merge, and `current.group` corresponds to the outer of the merge, the new group would be the output of the merge, and the inner extent of `linear_g` in the output domain seems to be `current.inner_extent * merge_inner->extent()`. What am I missing?

naoyam (commented):
Slowly moving forward...

csrc/device_lower/utils.cpp
zasdfgbnm: Added a comment in code:
```C++
      // Propagation of group may fail. For example, if the group before
      // propagation is [4, 3], and eg is split(3, 2), then the end result
      // would be like [4, 2, 2]. However, because the split is indivisible,
      // this makes group discontiguous. That is, "a composition of 4 and 3"
      // is not a "composition of 4, 2, and 2". In this case, the return value
      // of `propagate` will be std::monostate, indicating that the linear
      // proving fails. For this case, we just propagate the failure.
```

csrc/device_lower/utils.cpp
naoyam: Hmm, it isn't just this assignment, but the comment above also says:

```
 // The result is actually just the
 //   PartOf{group, current.inner_extent, current.selected_extent}
```

I assume `group` here refers to the result of propagation to `*current.group` at line 1303. Per the comment, I assume the rest of the code in this function is not required for correctness, so it seems to indicate `new_inner_extent` is indeed `current.inner_extent`. Please let me know where I misunderstood.

csrc/device_lower/utils.cpp
```diff
@@ -1013,6 +1016,625 @@ bool predicateAtEnd(ForLoop* loop) {
...
+    // common factors. The simplified group will be updated in the variable
+    // `group`, the extent of the group after simplification is stored in
+    // `group_extent`, and the new inner_extent after simplification is stored
+    // in `new_inner_extent`.
+    if (group.is<Composition>()) {
+      auto dq = group.as<Composition>();
+      // If the new_inner_extent is divisible by the extent of the last item
+      // in dq, we can simplify new_inner_extent and dq by canceling dq's last
+      // items. For example, if new_inner_extent is 6 and dq is [5, 3], we can
+      // simplify them to 2 and [5].
```
naoyam, line 1331: I'm puzzled here as I'm not sure why this is always valid simplification. 

IIUC, `dq: [5, 3], new_inner_extent: 6` means that there's a 2D tile of shape `5x3` (outer x inner). Since `new_inner_extent` is 6, the projection to this 2D tile is the `[2:, :]` subregion. I'm not sure why the inner axis can be dropped. Suppose the next expression where this projection is propagated along is `split the inner domain by 2`. If `[5, 3]` were simplified to `[5]`, I'm not sure how the next expression should be interpreted as there's no longer the inner domain of extent 3.

csrc/device_lower/utils.cpp
```diff
@@ -1013,6 +1016,625 @@ bool predicateAtEnd(ForLoop* loop) {
...
+      return related(g, from.front());
+    }));
+    // Split one group in current into two groups.
+    bool may_be_indivisible_split = eg->front()->isA<Split>() &&
+        !simplifyExpr(eg->front()->as<Split>()->isDivisible())->isTrue();
+    Composition<LinearGroupProjection> result;
+    bool first = true;
+    for (const auto& g : current) {
+      if (g.is<ValGroup>() && g.as<ValGroup>() == from.front()) {
+        if (may_be_indivisible_split && !first) {
```
naoyam, line 1449: Why `!first`? The first item of the composition correspond to the outermost dimension, right? Why can it be ignored?

csrc/device_lower/utils.cpp
zasdfgbnm: Oh, I see. I think I understand your question now. 

> For example, if the expr is a merge, and `current.group` corresponds to the outer of the merge, the new group would be the output of the merge, and the inner extent of `linear_g` in the output domain seems to be `current.inner_extent * merge_inner->extent()`.

This logic is inside the `if (from.size() == 1) {` clause, so if the expr is a merge, then it must be a backward propagation of merge. So `current.group` must be the output of the merge. After propagation, the new group will be a composition of the outer and inner of the merge.

> Per the comment, I assume the rest of the code in this function is not required for correctness, so it seems to indicate `new_inner_extent` is indeed `current.inner_extent`.

Oh, yes, that's right. For example, before the propagation, we have `group = 6` and `inner_extent = 3`. Assume expr is 6 = merge(2, 3), then after propagation, at this point, we will just have "part of the composition of [2, 3], where on the right of linear_g, there is a 3", which can be simplified as "2" later, but for now, we just have the unsimplified case.

csrc/device_lower/utils.cpp
zasdfgbnm: > the projection to this 2D tile is the `[2:, :]` subregion

Do you mean `[::2, 0]`?
```
>>> t = torch.arange(15).view(5, 3)
tensor([[ 0,  1,  2],
        [ 3,  4,  5],
        [ 6,  7,  8],
        [ 9, 10, 11],
        [12, 13, 14]])
>>> t[::2, :]
tensor([[ 0,  1,  2],
        [ 6,  7,  8],
        [12, 13, 14]])
```

> Suppose the next expression where this projection is propagated along is `split the inner domain by 2`. If `[5, 3]` were simplified to `[5]`, I'm not sure how the next expression should be interpreted as there's no longer the inner domain of extent 3.

`[5, 3]` is simplified as `[5]`, and `new_inner_extent` will become 2. When we see `split the inner domain by 2`, we will get
```C++
PartOf{group=[3, 2], inner_extent=2}
```
which will further be simplified as
```C++
PartOf{group=[3], inner_extent=1}
```
which will further be simplified as `[3]`, so `linear_g` is projected as the outer of the split.

This result makes sense, because, for example, if the `[5, 3]` is:
```
 0  1  2
 3  4  5
 6  7  8
 9 10 11
12 13 14
```

Then `PartOf{group=[5, 3], inner_extent=6}` means pad and view [5, 3] as [3, 6]:
```
 0  1  2  3  4  5
 6  7  8  9 10 11
12 13 14  _  _  _
```

Canceling the 3 just means: let's always consider each 3 items as a whole. Let's use `()` to denote a whole item that can not be further divided, then cancelling the 3 in `[5, 3]` as `[5]` means considering the above `[5, 3]` as

```
x1 = ( 0  1  2)
x2 = ( 3  4  5)
x3 = ( 6  7  8)
x4 = ( 9 10 11)
x5 = (12 13 14)
```

and `PartOf{group=[5], inner_extent=2}` means pad and view the [5] as [3, 2]:

```
x1 x2
x3 x4
x5  _
```

Splitting the [5] as [3, 2] means pad and view the same way as above, and `PartOf{group=[3, 2], inner_extent=2}` means viewing the above [3, 2] tensor as is because the inner is already 2.

Cancelling the 2 in `[3, 2]` as [3] means considering every two items as a whole, that is:

```
y1 = (x1 x2)
y2 = (x3 x4)
y3 = (x5  _)
```

csrc/device_lower/utils.cpp
zasdfgbnm: @naoyam Looks like in order to update this, I need to do a refactor to separate simplification with propagation. I believe this will also make the code easier to read. I will let you know when it is ready to review again.

zasdfgbnm:
@naoyam This PR is ready for review again. Highlighted changes:

- Rename `LinearGroupProjection` -> `Projection`
- Added a paragraph in comment suggesting that we should view `Projection` as a formal language where different types in the dynamic type represent different node types in the abstract syntax tree(AST). Updated other comments mentioning "abstract syntax tree". I believe this view point is helpful for understanding the algorithm.
- Added a new function `simplify` for traversing the AST, identifying specific patterns, and simplify identified patterns. The function `simplify` is very similar to a compiler where each pass find and replace a specific pattern in the AST.
- `propagate` now doesn't do any simplification. It focuses on generating the correct AST, and invoke `simplify` to simplify it.
- Rename `PartOf::group` -> `PartOf::what`.

I believe after the change, the code is much more modular and easier to understand now.

csrc/device_lower/utils.cpp
```diff
@@ -1013,6 +1016,775 @@ bool predicateAtEnd(ForLoop* loop) {
...
+PartOf<Projection> mergeParts(const PartOf<Projection>& part) {
+  // Combine PartOf(what=PartOf(what=x), ...) into PartOf(what=x, ...).
+  //
+  // Example:
+  // PartOf{
+  //   what=PartOf{what=24, inner_extent=2, selected_extent=12},
+  //   inner_extent=3,
+  //   selected_extent=2}
+  // =>
+  // PartOf{what=24, inner_extent=6, selected_extent=2}
```
naoyam, line 1388: To add a little more context here, `PartOf{what=24, inner_extent=2, selected_extent=12}` can be generated by merging `linear_g` of extent 12 and a domain of extent 2, i.e., `merge(linear_g, [2]) -> PartOf{what=24, inner_extent=2, selected_extent=12}`. 

Then, what propagation would create this?

```
   PartOf{
     what=PartOf{what=24, inner_extent=2, selected_extent=12},
     inner_extent=3,
     selected_extent=2}
```

csrc/device_lower/utils.cpp
```diff
@@ -1013,6 +1016,775 @@ bool predicateAtEnd(ForLoop* loop) {
...
+  return group;
+}
+
+PartOf<Projection> cancelCommonFactors(const PartOf<Projection>& part) {
+  // If `what` is a composition and inner_extent is a multiple of the extent of
+  // the last items in `what`, we can simplify the inner_extent and `what` by
+  // canceling the last items in `what`.
+  //
+  // Example:
+  // PartOf{what=[5,3,2], inner_extent=42} => PartOf{what=[5], inner_extent=7}
```
naoyam, line 1301: Could you also mention how `PartOf{what=[5,3,2], inner_extent=42}` can be generated?

zasdfgbnm: Screenshot_20240918_093052_Samsung Notes

zasdfgbnm: This is not a very good example. A better example could be
```
PartOf{what=[7,3,2], inner_extent=30}
```

csrc/device_lower/utils.cpp
```diff
@@ -1013,6 +1016,775 @@ bool predicateAtEnd(ForLoop* loop) {
...
+  Val* required_extent =
+      SimplifyingIrBuilder::mulExpr(part.selected_extent, part.inner_extent);
+
+  Val* what_extent = nullptr;
+  int64_t count = 0;
+  while (count < dq.size()) {
+    count++;
+    const auto& item = dq.at(dq.size() - count);
+    what_extent = SimplifyingIrBuilder::mulExpr(what_extent, extent(item));
+    if (simplifyExpr(IrBuilder::isDivisibleExpr(what_extent, required_extent))
```
naoyam, line 1358: Can this be just `what_extent >= required_extent`?

csrc/device_lower/utils.cpp
```diff
@@ -1013,6 +1016,775 @@ bool predicateAtEnd(ForLoop* loop) {
...
+      },
+      proj);
+}
+
+} // namespace
+
+Val* proveLinearAndGetStride(
+    const ValGraph& id_graph,
+    const ValGroup& linear_g,
+    const ValGroups& domain) {
```
naoyam, line 1760: Just in case:

```suggestion
    const ValGroups& domain) {
    FusionGuard fg(linear_g->front()->fusion());
```

csrc/device_lower/utils.cpp
naoyam: Hmm, I'm confused. `PartOf{what=24, inner_extent=2, selected_extent=12}` represents the `24` node. Then, IIUC, `PartOf{ what=PartOf{what=24, inner_extent=2, selected_extent=12},...` should be generated by propagating through either merge or split, but your diagram indicates something propagated along the `12` node. What am I missing?

csrc/device_lower/utils.cpp
naoyam: Do you mean a domain of extent 30 is split twice? I think I understand each split would generate `[1, 42], inner_extent=42` and `[5, 3, 2]`, but I don't follow why the could be combined.

csrc/device_lower/utils.cpp
zasdfgbnm: I believe no, for similar reason as https://github.com/NVIDIA/Fuser/pull/415 where we changed `>=` to `gcd` in vectorization helper.

csrc/device_lower/utils.cpp
zasdfgbnm: Yes, they split twice. They could combine because, for example, the `[1, 42]` is the loop domain, and the `[5, 3, 2]` is allocation domain.

csrc/device_lower/utils.cpp
zasdfgbnm: When we propagate from the red 2 to the 24, we first propagate through the merge that generates 6, at this point, we will get:
```
PartOf{what=6, inner_extent=3, selected_extent=2}
```
The next step is to propagate through the merge that generates the 12, and we will replace the `6` with
```
PartOf{what=12, inner_extent=nullptr, selected_extent=6}
```
which we get:
```
PartOf{what=PartOf{what=12, inner_extent=nullptr, selected_extent=6}, inner_extent=3, selected_extent=2}
```
which will be simplified as
```
PartOf{what=12, inner_extent=3, selected_extent=2}
```
Then we will propagate through the merge that generates the 24, which will replace the `12` with
```
PartOf{what=24, inner_extent=2, selected_extent=12}
```
and get
```
PartOf{what=PartOf{what=24, inner_extent=2, selected_extent=12}, inner_extent=3, selected_extent=2}
```

naoyam: Thanks a lot. This makes sense to me. Can you add this as a code comment as well? I think it's very helpful to understand the logic.

csrc/device_lower/utils.cpp
naoyam: Thanks. Can you please add this as a comment?

csrc/device_lower/utils.cpp
naoyam: Hmm, I'm not sure why they could be combined. I thought this analysis is meant to understand what portion (subregion) an ID corresponds to in a domain. If that's the case, I don't understand what the combined domain of a loop domain and an allocation domain would mean.

zasdfgbnm: Oh, I think it is because the above transformation is mathematically equivalent to:
Screenshot_20240918_122642_Samsung Notes

naoyam: Actually, I think I understand what you mean. 

1. Initial domain: `[1]`
2. Backward split by 42 -> `[30], inner_extent=42, selected_extent=1` 
3. Forward split by 6 -> `[5, 6], inner_extent=42, selected_extent=1`
4. Forward split by 2 -> `[5, 3, 2], inner_extent=42, selected_extent=1`

Am I right?

zasdfgbnm: Yes, right. I think I finally understand your question after you answered it yourself 😂.

naoyam: So, about this simplification, I suppose this is the same simplification we discussed earlier (`[5, 3]` -> `[5]`). In the case of `[5, 3, 2], inner_extent=42` -> `[5], inner_extent=7`, this means that we view the inner two IDs of `[3, 2]` as a non-modifiable unit, so it should be effectively equivalent to `[5], inner_extent=7`. However, what would happen if there's also an expr to traverse that uses `[3, 2]`? For example, let's say there's a split of `2` to `[1, 3]`. Since `what` no longer has `[2]`, this propagation would be just a no-op? Or, what about, e.g., a merge of `3` as the outer and `5` as the inner?

csrc/device_lower/utils.cpp
```diff
@@ -1013,6 +1016,775 @@ bool predicateAtEnd(ForLoop* loop) {
...
+// inner of [2] in the 256, we know that the [2] will be fully covered by the
+// inner 2 of the 4 of the output of the split. So, we know that, when the [2]
+// projects to the 4, it projects to the inner 2 of the 4. Now we have finished
+// propagation and reached domain. Because [2] is the inner of the 4, and the 4
+// is linear in domain, so we have proved that [2] is linear in domain. In
+// domain, the domains on the right of the 4 are 8, 8, 8, so the stride is 8*8*8
+// = 512.
+namespace {
+
+// From the above example, we can see that how linear_g lives in domain could be
```
naoyam, line 1089: General question: how is reordering handled? Since there's no "reorder" op, the ValGraph traversal itself doesn't seem to account for effects by reordering.

csrc/device_lower/utils.cpp
zasdfgbnm: This is a really good question. I love this question because it really touches the mathematical foundation of the proving we implemented.

When we start from `linear_g` and propagate to `5, 3, 2`, mathematically, we are proving that `linear_g ~ f(5, 3, 2)`, where `f` is a specific way to operates on its parameters, and `~` denote "equivalent" (e.g. `6 ~ [3, 2]` if you have `6 = merge(3, 2)`).

When we run the simplification, we simplify `f(5, 3, 2)` into `g(5)`, where `g` is another specific way to operate on its parameters. The process of simplifying `f(5, 3, 2)` into `g(5)` is just like in a math class, we simplify $f(x, y, z) = z + x^2 + 0 \cdot y - z$ as $g(x) = x^2$, where `y` and `z` is canceled means the value of `f(x, y, z)` does not depend on the value of `y` and `z`. Similarly, for this case, simplifying `f(5, 3, 2)` into `g(5)` means that, the index of `f(5, 3, 2)` does not depend on the index of `3` and `2`.

When we further propagate from `g(5)` to `domain`, mathematically, we are proving that `g(5) ~ h(x1, x2, ...)` where `x1`, `x2`, ... are ValGroups in `domain`. Note that this process only depend on the exprs between `5` and `domain`, nothing else. The split of `2` into `[1, 3]` is not on this path, so it will be unrelated. Merging `3` with `5` is related, but the only information that is related here is "3 is not a parameter of `g`", it does not matter where this `3` comes from and whether it is connected to `linear_g` or not, because here, we are just focusing on the subproblem of proving `g(5) ~ h(x1, x2, ...)` and `linear_g` is unrelated to this subproblem.

csrc/device_lower/utils.cpp
zasdfgbnm: Reorder does not need to be handled. Just like we do not need to handle reorder in the vectorization helper. However, we do check the orders when we see a merge, and we we run `proveLinearAndGetStrideAfterPropagation`(basically, any function that uses `search` checks the order).

csrc/device_lower/utils.cpp
zasdfgbnm: Added to the comment before `propagate`. Also, I moved the definition of `simplify` below `proveLinearAndGetStrideAfterPropagation` so that the order the reader read comments is more understandable.

zasdfgbnm: I also added a "Proof of correctness" to the comment of `cancelCommonFactors`

csrc/device_lower/utils.cpp
zasdfgbnm: Added a "Proof of correctness" for `trimRedundant`.

csrc/device_lower/utils.cpp
naoyam: So, IIUC, what you're saying is that no matter what other exprs are traversed, if it uses anything other than `5`, it should be just ignored as it shouldn't matter. Is my understanding correct?

naoyam: I think what I missed was this part:

> inner_extent is a multiple of the extent of the last items in `what`,

So, because of this condition, the canceled factor should not matter for the remaining traversal. 

It finally makes sense to me. Thanks for the updated comment. That helped a lot.

naoyam (approved):
LGTM. Thanks for all the discussions!

Issue #2888, opened 2024-09-03
Warning: Unroll required but not possible.


wujingyue:
I saw this warning every time I ran `mpirun -np 2 bin/test_multidevice --gtest_filter=DistributedTransformerTest*`. 

Not with `-np 1` though.

wujingyue:
cc @cowanmeg

wujingyue:
Thank you!

PR #2912, opened 2024-09-05
Ignore DID loop ids in unroll

cowanmeg:
Fixes false warning from https://github.com/NVIDIA/Fuser/issues/2888

### DIFF
--- a/csrc/index_compute.cpp
+++ b/csrc/index_compute.cpp
@@ -1214,7 +1214,9 @@ void ensureStaticIndexing(
       continue;
     }
     IterDomain* loop_id = loop->iter_domain();
-    if (loop->vectorize() || loop_id->isThread() || loop_id->isDeviceDim()) {
+    if (loop->vectorize() ||
+        nvfuser::ir_utils::isMemoryPartitionedAcross(
+            tv->getMemoryType(), loop_id->getParallelType())) {
       continue;
     }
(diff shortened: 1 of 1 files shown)


csrc/index_compute.cpp
```diff
@@ -1214,7 +1214,7 @@ void ensureStaticIndexing(
       continue;
     }
     IterDomain* loop_id = loop->iter_domain();
-    if (loop->vectorize() || loop_id->isThread()) {
+    if (loop->vectorize() || loop_id->isThread() || loop_id->isDeviceDim()) {
```
naoyam, line 1217: Can you please change this line to:

```suggestion
    if (loop->vectorize() || isMemoryPartitionedAcross(tv->getMemoryType, loop_id->getParallelType()) {
```

Basically, what we are doing here is to figure if each loop contributes to the index of the given tensor. If so, the loop needs to be fully unrolled to make the index static. That's important since dynamic indices would prevent the use of registers.

This particular line is just filtering out loops that won't matter. For register tensors, TID/BID parallelized loops won't matter as registers are thread local. That's obviously the case too for DID. 

`isMemoryPartitionedAcross` is a utility function we recently added to make this logic a little easier to change and extend. If we ever add another memory type, this would help reduce code changes.

naoyam (approved):
LGTM

cowanmeg:
!build

PR #2913, opened 2024-09-05
Multi-matmul scheduler: add test and schedule smem operand store

jacobhinkle:
This PR follows up from #2719 toward the eventual goal of merging #2458. Here I am introducing a parametrized test suite that I will use to check that the generated code matches the current scheduler for a variety of scenarios. I also am scheduling the first consumer tensor in the fusion: the smem store of operands. This shows how we can selectively check that the schedule is correct before we have everything implemented, since the test is only checking these tensors and nothing else in the fusion.

In the next PRs, I will schedule the rest of the prologue, as well as the mma result and the epilogue tensors. In each case, I will change the test to check more and more of the fusion for correctness.

This PR introduces `AbstractMatmuTensor`, an `AbstractTensor` for which each dimension can be tagged with a role. This lets us track the roles of each dimension during abstract tensor scheduling. This is only currently used in `blockTileTensors` and `mma_utils::makeTile`, but in the future we can imagine using it more.


jacobhinkle:
!build

jacobhinkle:
!build

csrc/scheduler/multi_matmul.cpp
```diff
@@ -29,7 +30,443 @@ namespace nvfuser {
...
+//! Automatically generates the shared memory swizzled data layout
+//!  for matmul mainloop and epilogue.
+//! The shared mem data layout is always 2D currently, and this utility
+//!  function assumes that the shared_mem_tv has the following structure:
+//!  [tile_row, tile_col]
+//! Returns the domain with swizzle. For the case of legacy swizzle, this
+//! domain must be set as loop domain. For the case of new swizzle, this domain
+//! must be set as allocation domain.
+template <bool legacy = true>
+AbstractTensor swizzleSharedMemory(TensorView* shared_mem_tv) {
```
jacobhinkle, line 68: Copied straight from matmul.cpp

jacobhinkle:
!build

tests/cpp/test_multi_matmul_scheduler.cpp
```diff
@@ -0,0 +1,620 @@
...
+    scheduleMultipleMatmuls(&new_fusion, params);
+
+    // find tensors to compare
+    auto getTensorsToCompare = [](Fusion* fusion) {
+      std::vector<TensorView*> tvs;
+
+      for (Val* v : fusion->inputs()) {
+        EXPECT_TRUE(v->isA<TensorView>());
+        // Take the consumer of each input, which is the smem store
+        tvs.push_back(v->uses().at(0)->output(0)->as<TensorView>());
```
jacobhinkle, line 368: Currently this just checks the smem store. But `compareTvs` is recursive so we can change this to add the smem loads next, then the MmaOp outputs, and eventually the fusion outputs, in order to check every tensor in the fusion.

csrc/id_model/schedule.cpp
zasdfgbnm, line 1: Could you move the change of this file to a separate PR?

tests/cpp/test_multi_matmul_scheduler.cpp
```diff
@@ -0,0 +1,625 @@
...
+        EXPECT_FALSE(tv_inp_new->isFusionInput());
+        compareTVs(tv_inp_orig, tv_inp_new);
+      }
+    }
+
+    // Print suffix whenever a test fails in recursive call
+    ASSERT_FALSE(testing::Test::HasFailure()) << suffix;
+  }
+
+  void compareSchedules() {
```
zasdfgbnm, line 350: What is the future plan of this check? I don't think in the long term we want to keep the legacy scheduler just for testing purposes. Will we completely removing those tests?

jacobhinkle: Yes. Once we remove the old scheduler we will remove this file.

csrc/scheduler/multi_matmul.cpp
```diff
@@ -239,10 +670,81 @@ class MultipleMatmulScheduler {
   // Currently the support is for a, b, c and d as fusion inputs/outputs
   //  aka. no prolog fusion yet.
   void defineOperandCaches() {
-    cacheOperandsToSmem(as_, acw_smems_, params_.supported_vec_size.a);
-    addSetsForCacheReads(acw_smems_, acrs_);
+    LoadStoreOpType load_op = params_.async_gmem_load_operands
+        ? LoadStoreOpType::CpAsync
+        : LoadStoreOpType::Set;
+
+    auto cacheOperandsToSmem = [&](const std::vector<TensorView*>& operands,
```
zasdfgbnm, line 677: We changed this to a member function in https://github.com/NVIDIA/Fuser/pull/2719, but it is changed back as a lambda here. Is this intentional?

jacobhinkle: Oh thanks for noticing. I just botched the merge. I will update it to reflect the change on main.

csrc/abstract_tensor.h
zasdfgbnm, line 1: Could you also move the change here into a separate PR? (different PR from `schedule.cpp`)

csrc/abstract_tensor.h
```diff
@@ -338,4 +351,120 @@ struct AbstractTensor {
...
+
+  using AbstractTensor::split;
+
+  AbstractTensor& split(int64_t axis, Val* factor, bool inner_split = true)
+      override {
+    axis = wrapDim(axis, (int64_t)domain.size());
+    // copy tags from original axis
+    tags.insert(tags.begin() + axis + 1, tags.at(axis));
+    return AbstractTensor::split(axis, factor, inner_split);
+  }
```
zasdfgbnm, line 404: Looks like this class has a lot of code duplication with `AbstractTensor`. Does it make sense to refactor the code like below?

```C++
template <typename Info>
struct AbstractTensorWithInfo {
  std::vector<std::pair<AbstractId, Info>> domain;
  ...
  AbstractTensor& AbstractTensor::merge(int64_t axis_o, int64_t axis_i) {
    wrap dim
    auto [id_o, info_o] = ...;
    auto [id_i, info_i] = ...;
    AbstractId result_id = do_the_merge(id_o, id_i);
    Info result_info = Info::merge(info_o, info_i);
    auto result = std::make_pair(result_id, result_info);
    // write result to outer
  }
  ...
};

struct NoInfo {
  static NoInfo merge(NoInfo, NoInfo) { return {}; }
  ...
} ; 

using AbstractTensor = AbstractTensorWithInfo<NoInfo>;

template <typename Tag>
struct TagInfo {
  std::unordered_set<Tag> tags;
  static TagInfo merge(TagInfo, TagInfo) { return ...; }
  ...
};

template <typename Tag>
using TaggedAbstractTensor = AbstractTensorWithInfo<TagInfo<Tag>>;
```

csrc/id_model/schedule.cpp
jacobhinkle: #2919

jacobhinkle:
!build

jacobhinkle:
!build

tests/cpp/test_multi_matmul_scheduler.cpp
```diff
@@ -0,0 +1,609 @@
...
+                        : at::randn({K, N}, options);
+    if (broadcasted) {
+      t0 = t0.unsqueeze(-1);
+      t1 = t1.unsqueeze(0);
+    }
+    return {t0, t1};
+  }
+
+  // Recursively compare scalar values
+  void compareScalars(Val* v_orig, Val* v_new) {
```
zasdfgbnm, line 162: Can this be simplified as:
```C++
return cloner_->clone(v_orig)->sameAs(v_new);
```
If the above isn't just working, then don't bother investing more time to make it work. This is a test that will throw out in the future, so I don't really care about having a clean and elegant implementation.

zasdfgbnm (approved):

csrc/scheduler/multi_matmul.cpp
```diff
@@ -326,6 +757,102 @@ class MultipleMatmulScheduler {
     }
   }
 
+  //! Rebuilds IdModel, then updates all ValGroups in abstract tensors to refer
+  //! to the new IdModel. This is necessary whenever we perform an operation
+  //! that creates a new TensorView, such as caching or rFactor
+  void updateIdModel() {
+    // Build new IdModel
+    IdModel new_id_model(fusion_, /*build_graphs=*/false);
+    new_id_model.buildPermissiveGraph();
```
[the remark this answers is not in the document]
jacobhinkle: It should. I will need to change that in the other spots in mma_utils as well so I was planning to follow this up with a small PR switching to Broadcast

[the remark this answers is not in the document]
zasdfgbnm: It's fine to land this first and make this change in a followup.

csrc/scheduler/multi_matmul.cpp
```diff
@@ -98,25 +534,22 @@ class MultipleMatmulScheduler {
     // and dimension roles for all tensors in the fusion
     findPatterns();
     translatePatterns();
-    // translatePatterns changes the TensorView graph, so we build the IdModel
-    // afterward
-    buildIdModel();
     findRoles();
```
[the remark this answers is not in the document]
jacobhinkle: It is used by `findRoles()` but it is updated at the end of `translatePatterns()` now. I moved it there to make the logic in this scope a bit clearer.

tests/cpp/test_multi_matmul_scheduler.cpp
jacobhinkle: This would work if we had cloned the _scheduled_ fusion, but here I have cloned the _unscheduled_ fusion then scheduled them both. That means that some of these scalars are new in both the original and the clone. When I clone those over using `cloner_->clone(v_orig)` we wind up with a new _undefined_ scalar, which fails the `sameAs` check if the original had a definition.

PR #2914, opened 2024-09-06
Fix typo in NoOpHeuristic::sameAs

jacobhinkle:
This was probably a cut/paste error. The result though is that comparing two `NoOpHeuristic`s will currently always return false, which leads to a recompile of the entire segmented fusion if it contains a single NoOp segment.


jacobhinkle:
!build

PR #2919, opened 2024-09-09
Scheduling tool: representativeId(const ValGroup&)

jacobhinkle:
This introduces a tool to get a representative IterDomain for a ValGroup. A ValGroup might have IterDomains with different IterTypes. For example, the Exact graph maps Iteration domains with Reduction domains, and the permissive graph maps Broadcast domains with Iteration when the broadcasts are concretized.

The purpose of `representativeId` is to find an `IterDomain` that is most useful for scheduling. In particular, it is an ID that is easiest to merge with other IDs. Since we restrict merging Reduction domains with non-Reduction domains, we give the lowest priority to Reduction domains in the ValGroup; we only return a Reduction domain if there is no other choice. Other than that, we prefer Iteration to Broadcast, we prefer unexpanded Broadcasts to expanded ones, and we prefer IDs with constant extents to those with symbolic extents.

### DIFF
--- a/csrc/id_model/schedule.cpp
+++ b/csrc/id_model/schedule.cpp
@@ -26,24 +26,16 @@ IterDomain* representativeId(const ValGroup& vg) {
     }
   };
 
-  auto preferNewExtent = [&rep](IterDomain* new_id) {
-    if (rep->hasExpandedExtent() && !new_id->hasExpandedExtent()) {
-      // Prefer non-expanded dimensions
-      return true;
-    }
-    // Prefer constant extent IDs to symbolic
-    return !rep->getMaybeExpandedExtent()->isConstInt() &&
-        new_id->getMaybeExpandedExtent()->isConstInt();
-  };
-
   for (Val* v : *vg) {
     if (auto id = dynamic_cast<IterDomain*>(v); id &&
         (rep == nullptr || preferNewIterType(id) ||
-         (id->getIterType() == rep->getIterType() && preferNewExtent(id)))) {
+         (id->getIterType() == rep->getIterType() &&
+          id->name() < rep->name()))) {
       rep = id;
       continue;
     }
   }
+
   NVF_ERROR(rep != nullptr);
   return rep;
 }
--- a/csrc/id_model/schedule.cpp
+++ b/csrc/id_model/schedule.cpp
@@ -10,10 +10,48 @@
 
 namespace nvfuser {
 
+IterDomain* representativeId(const ValGroup& vg) {
+  IterDomain* rep = nullptr;
+
+  auto preferNewIterType = [&rep](IterDomain* new_id) {
+    if (rep->isReduction()) {
+      return new_id->isBroadcast() || new_id->isIteration();
+    } else if (rep->isBroadcast()) {
+      return new_id->isIteration();
+    } else if (rep->isIteration()) {
+      return false;
+    } else {
+      // Prefer anything else to a non-iter, non-bcast, non-reduction ID
+      return true;
+    }
+  };
+
+  auto preferNewExtent = [&rep](IterDomain* new_id) {
+    if (rep->hasExpandedExtent() && !new_id->hasExpandedExtent()) {
+      // Prefer non-expanded dimensions
+      return true;
+    }
+    // Prefer constant extent IDs to symbolic
+    return !rep->getMaybeExpandedExtent()->isConstInt() &&
+        new_id->getMaybeExpandedExtent()->isConstInt();
+  };
+
+  for (Val* v : *vg) {
+    if (auto id = dynamic_cast<IterDomain*>(v); id &&
+        (rep == nullptr || preferNewIterType(id) ||
+         (id->getIterType() == rep->getIterType() && preferNewExtent(id)))) {
+      rep = id;
+      continue;
+    }
+  }
+  NVF_ERROR(rep != nullptr);
+  return rep;
+}
+
 ValGroup merge(ValGraph* graph, const ValGroup& g0, const ValGroup& g1) {
   NVF_ERROR(!g0->empty() && !g1->empty(), "ValGroup can not be empty");
-  auto g0_id = g0->front()->as<IterDomain>();
-  auto g1_id = g1->front()->as<IterDomain>();
+  auto g0_id = representativeId(g0);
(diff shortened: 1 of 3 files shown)


zasdfgbnm (approved):

tests/cpp/test_id_model.cpp
```diff
@@ -2614,4 +2615,33 @@ TEST_F(IdModelTest, ParallelTypePropagation) {
...
+  auto tv1 = makeConcreteTensor({-1, -1});
+  fusion.addInput(tv0);
+  fusion.addInput(tv1);
+
+  auto tv2 = add(tv0, tv1);
+  auto tv3 = sum(tv2, {0, 1});
+  fusion.addOutput(tv3);
+
+  IdModel id_model(&fusion);
+  const ValGraph& graph = id_model.idGraph(IdMappingMode::PERMISSIVE);
```
naoyam, line 2632: I'm thinking about removing the PERMISSIVE graph. Any particular reason to use it?

csrc/id_model/schedule.cpp
```diff
@@ -10,10 +10,48 @@
...
+      return new_id->isIteration();
+    } else if (rep->isIteration()) {
+      return false;
+    } else {
+      // Prefer anything else to a non-iter, non-bcast, non-reduction ID
+      return true;
+    }
+  };
+
+  auto preferNewExtent = [&rep](IterDomain* new_id) {
```
naoyam, line 29: Unless this graph is Exact, loop promotion man need to be used as well, but I'm not really sure which graph is intended to be used. Maybe only the Exact graph?

naoyam (changes requested):

tests/cpp/test_id_model.cpp
jacobhinkle: Right. I actually just wanted a graph with multiple IterTypes in it. I could instead just build a graph manually.

csrc/id_model/schedule.cpp
jacobhinkle: For AbstractTensor scheduling, we'll use something similar to Exact, but with concretized broadcasts mapped. I don't think we need to consider loop promotion though since we are just looking for a concrete IterDomain to use in an `IterDomain::merge` operation, for example.

naoyam: Hmm, actually, why do we need this? I understand the preference of IterTypes. Why does the extent also matter?

jacobhinkle: It probably doesn't matter much and I might have been overthinking things. I was thinking for example that when we merge ValGroups, we create a new output IterDomain that we can then potentially map other IterDomains into. Assuming those mapped IterDomains have symbolic extents, we would prefer that the original merged ID has a constant extent so that the ValGroup has a constant extent in at least one of its members. If we used representatives with symbolic extents when some constant ones exist, then that would not be the case: we'd have all symbolic extents in the merged ValGroup.

naoyam: My brain is not working today. Could you give a concrete example?

jacobhinkle: Suppose we have a Fusion with the following TVs (among others):

```
tv3[ i5, i6 ]
tv4[ 3, 4 ]
tv5 = tv3 + tv4
```
So `i5` is mapped to 3 and `i6` is mapped to 4.
```
AbstractTensor at = [ {i5, 3, ...}  {i6, 4, ...} ]
at.merge(0)
```
This merge step is done by using representative IDs for both `ValGroup`s and merging them to form a new ID, and then edges are added to the `ValGraph` representing the merge. If we use the first IDs in each group instead of the ones with constant extent, then the new `ValGroup` contains `{ i5*i6 }` instead of `{ 12 }`.

Now we apply this to `tv3`, but not `tv4`. This means the loop domain of `tv3` becomes `i5*i6`. We don't currently re-use any of the graphs we build for further steps in scheduling/lowering, but _if we did_ then we would already have a "nice" ID mapped to that loop domain, so we could infer that its extent is actually 12.

Note that this wouldn't have any effect right now since we create new IdModel during lowering anyway, so I'm fine with removing this part of the PR to improve clarity if you would like me to.

tests/cpp/test_id_model.cpp
jacobhinkle: Test was changed to build graph directly instead.

csrc/id_model/schedule.cpp
naoyam: Oh, somehow I was thinking about AbstractTensor with IterDomains, not ValGroups, so when you have a merge like:

```
AbstractTensor at = [ {i5, 3, ...}  {i6, 4, ...} ]
at.merge(0)
```

I was expecting to get:

```
[{i5 * i6}, {3 * 4}, ...]
```

naoyam: Now I understand that extents do matter. I still don't know what this would mean with any graph except for Exact.

In Permissive, a ValGroup may have IDs with different extents, and all of them may be non-broadcast. What is representative then? 

I feel that this would be only well defined for the Exact graph, where extents are all guaranteed to be the same. If that's the case, choosing which extent to use shouldn't matter.

jacobhinkle: > I feel that this would be only well defined for the Exact graph, where extents are all guaranteed to be the same. If that's the case, choosing which extent to use shouldn't matter.

Good point. I think I will just remove all the extent stuff so that the code is simply choosing based on IterType.

naoyam: Note that what loop promotion tries to analyze is exactly this problem, so that's why I wondered if loop promotion should also be used here.

I think it's also fine to keep this almost as is, as long as some safeguards are added. For example, if two different symbolic extents are given, we could simply throw an error. Depending on what you're planning to use this, that may be good enough to move forward for now. If that's the case, I'd be happy to give a stamp.

jacobhinkle: > Depending on what you're planning to use this, that may be good enough to move forward for now.

All I actually need from this PR is to be sure that it's possible to merge two ValGroups even if they include a mix of Reduction and Iteration domains, which the `IterType` checks will ensure. I think we should probably wait for more motivation to implement extent checks/preferences.

PR #2920, opened 2024-09-09
Benchmark `broadcast + add`

Priya2698:
Ports [CPP benchmark](https://github.com/NVIDIA/Fuser/blob/main/benchmarks/cpp/broadcast.cpp) to python.
The fusion definition corresponds to:
```
def bcast_add(x, bias, bcast_dim):
    return x + bias.unsqueeze(bcast_dim)
```
This benchmark uses 2D inputs. The parameters here are:
`bcast_axis`: outer or inner (0 / 1), specifies which axis in bias is broadcasted.
`contiguous`: True or False, specifies if `x` is contiguous in memory.

### DIFF
--- a/benchmarks/python/test_broadcast_add_fwd.py
+++ b/benchmarks/python/test_broadcast_add_fwd.py
@@ -25,7 +25,11 @@ def bcast_add_fusion(
         stride_order=[1, 0],
     )
 
-    T3 = fd.ops.broadcast_in_dim(T0, shape=T1.shape(), broadcast_dims=[1 - bcast_axis])
+    bcast_shape = [T1.size(0), T1.size(1)]
+    bcast_shape[bcast_axis] = 1
+
+    T2 = fd.ops.broadcast_in_dim(T0, shape=bcast_shape, broadcast_dims=[1 - bcast_axis])
+    T3 = fd.ops.broadcast_in_dim(T2, shape=T1.shape(), broadcast_dims=[0, 1])
 
     if dtype in PROMOTE_DTYPES:
         T1 = fd.ops.cast(T1, dtype=DataType.Float)
--- a/benchmarks/python/test_broadcast_add_fwd.py
+++ b/benchmarks/python/test_broadcast_add_fwd.py
@@ -1,0 +1,94 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-present NVIDIA CORPORATION & AFFILIATES.
+# All rights reserved.
+# SPDX-License-Identifier: BSD-3-Clause
+import pytest
+from nvfuser import FusionDefinition, DataType
+from nvfuser.pytorch_utils import torch_dtype_to_nvfuser_dtype, clear_cuda_cache
+from .core import run_benchmark, clear_dynamo_cache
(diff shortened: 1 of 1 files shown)


benchmarks/python/test_broadcast_add_fwd.py
```diff
@@ -0,0 +1,98 @@
...
+        dtype=dtype,
+        is_cpu=False,
+        stride_order=[1, 0],
+    )
+
+    bcast_shape = [T1.size(0), T1.size(1)]
+    bcast_shape[bcast_axis] = 1
+
+    T2 = fd.ops.broadcast_in_dim(T0, shape=bcast_shape, broadcast_dims=[1 - bcast_axis])
+    T3 = fd.ops.broadcast_in_dim(T2, shape=T1.shape(), broadcast_dims=[0, 1])
```
naoyam, line 32: Why are there two broadcast ops? Is the second one necessary?

benchmarks/python/test_broadcast_add_fwd.py
```diff
@@ -0,0 +1,98 @@
...
+def test_bcast_add_baseline_benchmark(
+    benchmark,
+    size: tuple,
+    dtype: torch.dtype,
+    bcast_axis: int,
+    compile: bool,
+):
+    clear_cuda_cache()
+    if compile:
+        clear_dynamo_cache()
```
naoyam, line 89: nit: is there anything like a hook where we can register some common operations before each benchmarking? I assume we need to do these clear ops with all benchmarks. I'd expect testing/benchmarking frameworks would have some mechanism to simplify such common operations.

benchmarks/python/test_broadcast_add_fwd.py
Priya2698: No, we can combine them into a separate op.
But I kept the fusion definition consistent with what Thunder generated.

For outer broadcast, if the fn is `x + bias`, then Thunder only produces one bcast op.
The use of `unsqueeze` generates 2 broadcast ops.

benchmarks/python/test_broadcast_add_fwd.py
Priya2698: Good point, I'll look into it.

benchmarks/python/test_broadcast_add_fwd.py
naoyam: OK, can you just leave a comment about that?

Priya2698:
CC: @xwang233 for additional parameters.

csarofeen:
LGTM pending Naoya's comments. Thanks!

benchmarks/python/test_broadcast_add_fwd.py
Priya2698: Added a comment.

benchmarks/python/test_broadcast_add_fwd.py
Priya2698: I'll pursue this is in a separate PR, since it will be a module-wide change.

naoyam (approved):
LGTM

PR #2921, opened 2024-09-09
Benchmark `split + silu + mul` fusion

Priya2698:
The fusion definition corresponds to:
```
def split_silu(xy): 
    x, y = torch.chunk(inputs[0], 2, -1)
    return torch.nn.functional.silu(x) * y
```

### DIFF
--- a/benchmarks/python/test_split_silu_bwd.py
+++ b/benchmarks/python/test_split_silu_bwd.py
@@ -1,0 +1,118 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-present NVIDIA CORPORATION & AFFILIATES.
+# All rights reserved.
+# SPDX-License-Identifier: BSD-3-Clause
+import pytest
+from nvfuser import FusionDefinition, DataType
+from nvfuser.pytorch_utils import torch_dtype_to_nvfuser_dtype, clear_cuda_cache
+from .core import run_benchmark, clear_dynamo_cache, unary_bwd_torch
+import torch
+from .global_params import generate_input_sizes, FLOAT_DTYPES, PROMOTE_DTYPES
+import numpy as np
+
+
+def split_silu_bwd_fusion(fd: FusionDefinition, dtype: DataType, size: tuple):
+    T0 = fd.define_tensor(
+        shape=[-1, -1],
+        contiguity=[True, True],
+        dtype=dtype,
+        is_cpu=False,
+        stride_order=[1, 0],
+    )
+    T1 = fd.define_tensor(
+        shape=[-1, -1],
+        contiguity=[True, True],
+        dtype=dtype,
+        is_cpu=False,
+        stride_order=[1, 0],
+    )
+    S0, S1 = size
+    T2 = fd.ops.slice(
+        T0, start_indices=[0, 0], end_indices=[S0, S1 // 2], strides=[1, 1]
+    )
+    T3 = fd.ops.slice(
+        T0, start_indices=[0, S1 // 2], end_indices=[S0, S1], strides=[1, 1]
+    )
+
+    if dtype in PROMOTE_DTYPES:
+        T1 = fd.ops.cast(T1, dtype=DataType.Float)
+        T2 = fd.ops.cast(T2, dtype=DataType.Float)
+        T3 = fd.ops.cast(T3, dtype=DataType.Float)
+
+    T4 = fd.ops.neg(T2)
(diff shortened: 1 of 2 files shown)


benchmarks/python/test_split_silu_fwd.py
```diff
@@ -0,0 +1,88 @@
...
+    T3 = fd.ops.neg(T1)
+    T4 = fd.ops.exp(T3)
+    S5 = fd.define_scalar(1.00000, dtype=DataType.Double)
+    T6 = fd.ops.add(S5, T4)
+    T7 = fd.ops.reciprocal(T6)
+    T8 = fd.ops.mul(T1, T7)
+    T9 = fd.ops.mul(T8, T2)
+    if dtype in PROMOTE_DTYPES:
+        T9 = fd.ops.cast(T9, dtype=dtype)
+    fd.add_output(T9)
```
naoyam, line 39: @IvanYashchuk Could you confirm if this matches with the case you're looking at?

benchmarks/python/test_split_silu_bwd.py
```diff
@@ -0,0 +1,118 @@
...
+    T18 = fd.ops.neg(T17)
+    T19 = fd.ops.add(T12, T18)
+    S20 = fd.define_scalar(0.00000, dtype=DataType.Double)
+    T21 = fd.ops.pad(T11, [S1 // 2, 0, 0, 0], S20)
+    S22 = fd.define_scalar(0.00000, dtype=DataType.Double)
+    T23 = fd.ops.pad(T19, [0, S1 // 2, 0, 0], S22)
+    T24 = fd.ops.add(T21, T23)
+    if dtype in PROMOTE_DTYPES:
+        T24 = fd.ops.cast(T24, dtype=dtype)
+    fd.add_output(T24)
```
naoyam, line 64: @IvanYashchuk Could you confirm if this matches with the case you're looking at?

naoyam:
@Priya2698 Thanks for the PR. Have you checked the performance? Does it match with what Ivan reported?

benchmarks/python/test_split_silu_bwd.py
IvanYashchuk: This case is slightly different because, currently, Thunder produces different backward definitions depending on whether `split` or `chunk` was used. Here no `cat` is present (backward of `slice` is `pad`). For `split` there's an explicit rule that backward of `split` is `cat`. I suggest adding in addition to this definition also the one with `cat`. It's unclear what's easier for nvFuser to support performantly. The other backward variant can be generated by using `x, y = torch.split(inputs[0], inputs[0].shape[-1] // 2, -1)` instead of `x, y = torch.chunk(inputs[0], 2, -1)` in forward.

Priya2698:
Fwd benchmark: http://nv/eka
Bwd benchmark: http://nv/ekb

I can see the regression against torchcompile for forward benchmark, but not backward.

I will add the `torch.split` version for backward and re-run.

benchmarks/python/test_split_silu_bwd.py
Priya2698: > This case is slightly different because, currently, Thunder produces different backward definitions depending on whether `split` or `chunk` was used. Here no `cat` is present (backward of `slice` is `pad`). For `split` there's an explicit rule that backward of `split` is `cat`. I suggest adding in addition to this definition also the one with `cat`. It's unclear what's easier for nvFuser to support performantly. The other backward variant can be generated by using `x, y = torch.split(inputs[0], inputs[0].shape[-1] // 2, -1)` instead of `x, y = torch.chunk(inputs[0], 2, -1)` in forward.

If we are adding another variant, we should do the same for fwd. My concern is that the baselines (torch.compile and eager) may have different performances (will have to verify) for the two variants even if the fusion definition is the same.
Wdyt? @naoyam @IvanYashchuk -- is there a preference for only selecting one? The upside of not adding an additional forward benchmark would be saved time on CI and minimize redundancy in nvfuser benchmarking.  
Else, we can have essentially have 2 benchmarks, for `fwd` and `bwd` each.

naoyam: Can we have two as separate benchmarks but each with a smaller set of input sizes so that the overall workload would be roughly the same?

Priya2698:
Update:

I noticed some discrepancy in torch.compile benchmarks  (PR #3300) -- I will be re-running these benchmarks and update the results here.

PR #2922, opened 2024-09-09
TP Backward Transformer

cowanmeg:
This differs slightly from the backward transfromer trace, because it caches the second linear output from the multiheaded attention layer. The thunder trace recomputes this, but the second linear layer requires an allreduce, which would push the implementation to use 3 AllReduces instead of 2 in the backward pass. 

Note: Renaming of variables to stay consistent with others.


tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -118,20 +118,20 @@ std::vector<at::Tensor> reference_mha(
     at::Tensor w1,
     at::Tensor b1) {
   auto at_dtype = w0.dtype();
-  auto m = (at::matmul(x, w0).to(at::kFloat) + b0).view({B, S, 3 * E});
-  auto qkv_vec = m.split(E, 2);
+  auto linear0 = (at::matmul(x, w0).to(at::kFloat) + b0).view({B, S, 3 * E});
```
cowanmeg, line 121: Note, this is just renaming to be consistent.

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -270,34 +270,32 @@ std::vector<TensorView*> mlp(
     TensorView* w1,
     TensorView* b1,
     const DeviceMesh& mesh) {
-  DataType dtype = w0->dtype();
-  // Linear #1
-  TensorView* matmul1 = matmul(x, w0);
-  TensorView* b0_bcast = broadcast(b0, {false, true, false});
-  TensorView* linear1 = add(matmul1, b0_bcast);
+  const DataType dtype = w0->dtype();
```
cowanmeg, line 273: More renaming

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -307,59 +305,55 @@ std::vector<TensorView*> mha(
     TensorView* w1,
     TensorView* b1,
     const DeviceMesh& mesh) {
-  DataType dtype = w0->dtype();
-  // Linear 1
-  TensorView* mm = matmul(x, w0);
-  TensorView* proj_bias_bcast = broadcast(b0, {false, true, false});
-  TensorView* qkv1 = add(mm, proj_bias_bcast);
+  const auto D = w0->axis(0)->extent()->value().as<int64_t>();
```
cowanmeg, line 308: More renaming + removing some explicit shardings when forming q, k, v vectors

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -50,6 +50,7 @@ class DistributedTransformerTest
     if (!deviceMajorMinorCheck(8)) {
       GTEST_SKIP() << "Distributed transformer tests require Ampere or newer";
     }
+    at::globalContext().setAllowBF16ReductionCuBLAS(false);
```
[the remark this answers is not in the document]
cowanmeg: ahh shoot, that should be gone. I was observing error differences by setting that flag.

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -118,20 +119,20 @@ std::vector<at::Tensor> reference_mha(
...
     at::Tensor b1) {
   auto at_dtype = w0.dtype();
-  auto m = (at::matmul(x, w0).to(at::kFloat) + b0).view({B, S, 3 * E});
-  auto qkv_vec = m.split(E, 2);
+  auto linear0 = (at::matmul(x, w0).to(at::kFloat) + b0).view({B, S, 3 * E});
+  auto qkv = linear0.split(E, 2);
   for (auto i = 0; i < 3; i++) {
-    qkv_vec[i] =
-        qkv_vec[i].reshape({B, S, H, E / H}).transpose(1, 2).to(at_dtype);
+    qkv[i] = qkv[i].reshape({B, S, H, E / H}).transpose(2, 1).to(at_dtype);
```
wujingyue, line 125:
```suggestion
    qkv[i] = qkv[i].reshape({B, S, H, E / H}).transpose(1, 2).to(at_dtype);
```

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -270,34 +271,63 @@ std::vector<TensorView*> mlp(
...
+      reshape(linear0, {D, B * S, 3 * E / D}, {D, B, S, 3 * E / D});
+  std::vector<TensorView*> qkv;
+  for (auto i : c10::irange(3)) {
+    TensorView* tv_reshaped =
+        slice(qkv_cat, {0, 0, 0, E / D * i}, {D, B, S, E / D * (i + 1)});
+    tv_reshaped =
+        reshape(tv_reshaped, {D, B, S, E / D}, {D, B, S, H / D, E / H});
+    tv_reshaped = castOp(dtype, transpose(tv_reshaped, 2, 3));
+    // TODO: this might not be needed Explicitly shard q, k, and v before
+    // calling SDPA node
```
wujingyue, line 324:
```suggestion
    // Explicitly shard q, k, and v before
    // calling SDPA node. TODO: this might not be needed. 
```

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -270,34 +271,63 @@ std::vector<TensorView*> mlp(
...
+  // Recompute: linear 0, q, k, and v
+  TensorView* matmul0 = matmul(x, w0);
+  TensorView* linear0 = add(matmul0, broadcast(b0, {false, true, false}));
+  // Forming the q,k,v:
+  TensorView* qkv_cat =
+      reshape(linear0, {D, B * S, 3 * E / D}, {D, B, S, 3 * E / D});
+  std::vector<TensorView*> qkv;
+  for (auto i : c10::irange(3)) {
+    TensorView* tv_reshaped =
+        slice(qkv_cat, {0, 0, 0, E / D * i}, {D, B, S, E / D * (i + 1)});
```
wujingyue, line 319: I think you can simplify this using https://github.com/NVIDIA/Fuser/blob/d9a50596a92e098b154ea3248ecaf29d07e822ad/csrc/ops/alias.cpp#L847

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -270,34 +271,63 @@ std::vector<TensorView*> mlp(
...
 
-  return {linear1, gelu, linear2, dropout_result};
+  return {linear0, gelu, linear1, dropout_result};
+}
+
+std::vector<TensorView*> mha_qkv(
+    TensorView* x,
+    TensorView* w0,
+    TensorView* b0,
+    TensorView* w1,
```
wujingyue, line 306: w1 doesn't seem to be used anywhere in this function.

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -984,6 +998,275 @@ TEST_P(DistributedTransformerTest, Forward) {
...
+  at::ScalarType at_dtype = data_type_to_aten(dtype);
+  auto fusion = std::make_unique<Fusion>();
+  FusionGuard fg(fusion.get());
+  const auto mesh = DeviceMesh::createForNumDevices(D);
+  constexpr float kEps = 1e-5;
+  auto eps = IrBuilder::create<Val>(kEps);
+  std::vector<int64_t> norm_shape{E}; // 768
+
+  // Note: Deviate from the thunder trace with layer norm recomputation
+  // Thunder saves intermediate values that are not exposed by ATen
```
wujingyue, line 1012: > not exposed by ATen

Can you be more specific about this? ATen should have enough low-level APIs to rebuild the Thunder trace. I guess you meant to say these intermediate values are not exposed by certain ATen APIs, making it hard to build the reference implementation. Correct?

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -270,34 +271,63 @@ std::vector<TensorView*> mlp(
...
+
+std::vector<TensorView*> mha_qkv(
+    TensorView* x,
+    TensorView* w0,
+    TensorView* b0,
+    TensorView* w1,
+    const DeviceMesh& mesh) {
+  DataType dtype = w0->dtype();
+  const auto D = w0->axis(0)->extent()->value().as<int64_t>();
+  // Recompute: linear 0, q, k, and v
```
wujingyue, line 310:
```suggestion
```
I don't find this comment relevant any more after you moved it into this helper function. The word "recompute" only makes sense under the context of backprop.

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -984,6 +998,275 @@ TEST_P(DistributedTransformerTest, Forward) {
...
+  // The thunder trace recompute mha linear1, but this would result in 3
+  // AllReduces in the backwards pass.
+  if (mha_linear1 == nullptr) {
+    TensorView* sdpa_transpose = transpose(mha_sdpa_out, 2, 3);
+    TensorView* sdpa_reshape =
+        reshape(sdpa_transpose, {D, B, S, H / D, E / H}, {D, B * S, E / D});
+    TensorView* mha_local_matmul1 = matmul(sdpa_reshape, mha_w1);
+    TensorView* mha_matmul1 = sum(mha_local_matmul1, {0}); // allreduce
+    mha_linear1 = add(mha_matmul1, broadcast(mha_b1, {true, false}));
+  }
```
wujingyue, line 1081:
```suggestion
#if 0
  // The thunder trace recomputes mha linear1, but this would result in 3
  // AllReduces in the backwards pass.
  if (mha_linear1 == nullptr) {
    TensorView* sdpa_transpose = transpose(mha_sdpa_out, 2, 3);
    TensorView* sdpa_reshape =
        reshape(sdpa_transpose, {D, B, S, H / D, E / H}, {D, B * S, E / D});
    TensorView* mha_local_matmul1 = matmul(sdpa_reshape, mha_w1);
    TensorView* mha_matmul1 = sum(mha_local_matmul1, {0}); // allreduce
    mha_linear1 = add(mha_matmul1, broadcast(mha_b1, {true, false}));
  }
#endif
```

wujingyue: `mha_linear1 == nullptr` never happens in this test, so I suggest we make this a comment or completely remove this. It's hard to keep untested code working.

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -984,6 +998,275 @@ TEST_P(DistributedTransformerTest, Forward) {
...
+  Val* scale = IrBuilder::create<Val>(1.0 / (1.0 - kDropoutProb));
```
wujingyue, line 1082:
```suggestion
  Val* dropout_scale = IrBuilder::create<Val>(1.0 / (1.0 - kDropoutProb));
```
so the readers immediately know what this scale is for.

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -984,6 +998,275 @@ TEST_P(DistributedTransformerTest, Forward) {
...
+  // Use input mha mask to implement dropout
+  auto mha_out = mul(mha_linear1, mha_mask);
+  mha_out = mul(mha_out, scale);
+  auto resid_0 = add(x, mha_out);
+  auto ln_1 = layer_norm(resid_0, norm_shape, ln1_w, ln1_b, eps);
+  auto mlp_in = castOp(dtype, ln_1.output);
+  // Note: We only use linear0 and gelu outputs from the mlp forward pass.
+  auto mlp_tensors = mlp(mlp_in, mlp_w0, mlp_b0, mlp_w1, mlp_b1, mesh);
```
wujingyue, line 1090:
```suggestion
  auto mlp_tvs = mlp(mlp_in, mlp_w0, mlp_b0, mlp_w1, mlp_b1, mesh);
```

wujingyue, line 1090: I like how you made LinearBackwardsResult to encapsulate things into a struct with named fields. You may want to consider doing that for MLP results as well, so (1) the readers are clear about the semantics of `mlp_tensors[0]` and `mlp_tensors[1]` and (2) it's harder to make mistakes if we change `mlp`'s return values.

wujingyue: (Of course, this can come in a separate PR)

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -984,6 +998,275 @@ TEST_P(DistributedTransformerTest, Forward) {
...
+      mha_in,
+      mha_w0,
+      mha_b0,
+      mha_w1,
+      mha_mask,
+      mha_sdpa_out,
+      mha_sdpa_log_sumexp,
+      mha_sdpa_seed,
+      mha_sdpa_offset,
+      ln1_grads.grad_input,
```
wujingyue, line 1122: Don't we need to add `grad` on top of this? In the forward pass, the result of MHA is forked to `ln1` and the `add` at the end. So in backprop I expect the two corresponding gradients to be added.

tests/cpp/test_multidevice_transformer.cpp
cowanmeg: Yes, I was thinking about doing that for all the helpers. Probably for a later PR since this one is getting quite big!

tests/cpp/test_multidevice_transformer.cpp
cowanmeg: oops, I thought I had removed this. I was debating making the test parameterized to be closer to Thunder and recompute mha_linear1, but maybe a follow-up

tests/cpp/test_multidevice_transformer.cpp
cowanmeg: Disregard! The ATen implementation actually does expose the intermediates we need, they were just named slightly different form nvfuser! Will fix.

tests/cpp/test_multidevice_transformer.cpp
cowanmeg: let me check the thunder trace more closely, I might have missed that.

cowanmeg: Thanks! I was missing both the residual adds in the back prop.
Also the forward pass had the wrong tensors added in the residuals which is also fixed.

cowanmeg:
!build

samnordmann (approved):
LGTM! :)

tests/cpp/test_multidevice_transformer.cpp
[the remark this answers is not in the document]
cowanmeg: Not needed! This is a flag for controlling whether a bf16 matmul reduction is in bf16 or float. I was playing around with the setting to see if the errors changed, but the default is false, so the accuracy didn't change.

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -431,21 +444,63 @@ LinearBackwardsResult sharded_linear_backwards(
...
+  std::vector<bool> outer_broadcast_mask(kNumberOfDims, false);
+  for (const auto idx : c10::irange(kOuterNumDims)) {
+    outer_broadcast_mask[idx] = true;
+  }
+
+  auto x_sub_mean = sub(x, mean_bcast);
+  auto y = mul(x_sub_mean, invstd);
+
+  // Optional: norm * weight
+  if (weight != nullptr) {
```
wujingyue, line 468: AFAICS, all callers pass in a non-null weight and bias, so this check can be removed for simplicity.

wujingyue (approved):

PR #2833, opened 2024-08-22
Add TMA support for circular buffering pass

rdspring1:
## Summary ##
This PR adds support for TMA circular buffering. It is stacked on #2824 and #2825.
Tracking branch: https://github.com/NVIDIA/Fuser/pull/2773

## Description ##
- The Pre-Prologue and Post-Epilogue loops are created in the allocation pass.
- Pre-Prologue loop allocates share memory and initializes mbarriers, while Post-Epilogue loop invalidates mbarriers.
- In the circular buffer pass, `CloneTmaCircularBufferLoopAndInsertSync` clones operations and inserts mbarrier synchronization logic to create the prologue, main, and epilogue for-loops.
- Prologue copies only the load operations. `arriveExpectTx` and `arrive` expressions are created for cpAsyncBulk load operations.
- Main loop copies the load and computation operations, adds `arriveExpectedTx` and `arrive` for next stage, and calls `mbarrierWait` for current stage.
- Epilogue copies only the computation operations and adds `mbarrierWait` for remaining stages in the pipeline.

## Lowering Details ##
Description of changes in lowering passes.
- `Prologue`, `Main`, and `Epilogue` loops are created by `CloneTmaCircularBufferLoopAndInsertSync` which is a child class of `CircularBufferLoopCloner`.
- `PrePrologue` and `PostEpilogue` loops are created in the allocation pass.
- The `cuTensorMapEncodeTiled ` restricts the size of each box dimension to be `<= 256`. You will need to launch multiple load operations to load larger tiles. 
- We only allocate `mbarriers` for each stage, so the `expected_transaction` bytes is multiplied by the number of TMA loads per stage.
- The for-loop cloner must account for the nested for-loop structure used to launch multiple TMA loads before adding the `mbarrier_wait` for the stage.

## Loop Structure ##
Description of for-loop structure for circular buffering.

<details>
<summary>Overview Circular Buffer Structure:</summary>

### Pre-prologue loop: ###
- Allocate shared memory for mbarriers and mbarrier tokens
- Initialize mbarrier for all stages

### Prologue loop: ###
- if selected_thread:
  - Issue cp async bulks for all but last stage

### Main loop: ###
- if selected_thread:
  - Issue next cp async bulk for available stage
- All threads wait until tma operation arrives
- Copy body without
  - shared memory allocations
  - mbarrier_init exprs
  - mbarrier_inval exprs

### Epilogue loop: ###
- All threads wait until tma operation arrives
- Copy body without
  - shared memory allocations
[...]

PR #2923, opened 2024-09-09
Create ElectSync predicate type

rdspring1:
## Summary
The PR creates the `ElectSync` predicate type to select a single thread in an if-then-else block.

## Why?
The standard predicate `threadIdx.x == 0 && threadIdx.y == 0 && threadIdx.z == 0` selects a specific thread from a CTA, which can create a peeling loop. The `elect.sync` ptx will select an arbitrary thread from the a given warp.

## Lowering Details

1. Create `ElectSync` PredicateType
2. Add `ElectSync` as a unary op.
3. Enforce that `blockDim.x` has at least one warp because the default `membermask` is `0xFFFFFFFF`.

## NvFuser's ElectSync Predicate
```cpp
int warp_idx_zero = threadIdx.x < 32 && threadIdx.y == 0 && threadIdx.z == 0;
int lane_predicate = hopper::electSync(/*membermask=*/0xFFFFFFFF);
if (warp_idx_zero && lane_predicate) {
  // Do Something
}
```
Note: It could be more efficient to use `__shfl_sync` to get `warp_idx` like CUTLASS.

## How CUTLASS selects a leader thread to issue TMA instructions
```cpp
int warp_idx = __shfl_sync(0xffffffff, threadIdx.x / 32, 0);
int lane_predicate = hopper::electSync(/*membermask=*/0xFFFFFFFF);

if ((warp_idx == 0) && lane_predicate) {
  // Issue Tma Descriptor Prefetch from a single thread
}
```


rdspring1:
!build

csrc/device_lower/utils.h
```diff
@@ -143,6 +143,9 @@ bool isCpAsyncBulk(const Expr* expr);
 //! Short-cut for detecting initialization for cpAsync op.
 bool isCpAsyncInit(const Expr* expr);
 
+//! Returns true if the value is a kir::Predicate with PredicateType::ElectSync
+bool isElectSyncPredicate(const Val* v);
```
naoyam, line 147: nit: looks like it's only used once. Is it common enough to have it here?

naoyam:
Why not exactly following what CUTLAS does?

naoyam:
I'd drop `Sync` from the name. It implies the warp synchronous instruction execution. Here, it's just electing one thread from a block. Maybe `ElectThread`?

I don't think it's necessary at this moment, but to make it more future proof, we may also want to consider how this predicate could be extended to support electing one execution entity within each parallel type. Again, just a thought, not important right now.

rdspring1:
> Why not exactly following what CUTLASS does?

Cutlass already knows which dimensions of the block are zero, so they can skip the extra predicates. 

> consider how this predicate could be extended to support electing one execution entity within each parallel type.

I agree. Cutlass does select warps based on `threadIdx.x` and `threadIdx.y`.

> Change name to ElectThread?

I don't have a strong opinion about changing the name to `ElectThread`. However, all threads within the member-mask are synchronized, so there is some synchronization with this PTX instruction.

> The mandatory .sync qualifier indicates that elect causes the executing thread to wait until all threads in the member-mask execute the elect instruction before resuming execution.

Reference: https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#parallel-synchronization-and-communication-instructions-elect-sync

zasdfgbnm:
I think my biggest concern is, I don't think `ElectSync` is a separate predicate type. I would rather consider it as an optimization on existing predicate types. For example, if BDIMx is 32, and there is an IterDomain `I0{5}` parallelized in TIDx, then we will generate predicate `threadIdx.x < 5`, which can not be optimized as elec sync. Similarly, if `I0`'s extent is `1`, instead of `5`, then we will have `threadIdx.x < 1`, which can be optimized as elect sync.

The reason I dislike having this `ElectSync` as a separate predicate type is because it effectively skips all our existing analyses on predication and just hard code the "select one thread", which would generate silent wrong result if the schedule does not imply "only one thread does the work". In the above example, if the extent of `I0` is 5, then using the elec sync predicate type is just silently wrong.

I believe a better way to do this is: run through all the existing predicate analyses we already have in our system. If these analyses indicates that only one thread is selected, then optimize it as elec-sync.

What do you think?

naoyam:
I generally agree with @zasdfgbnm. Ideally, we should automatically use `elect_sync` whenever possible. We have "thread predicates", which indicates which threads should be allowed to execute. Maybe we could use it to pipe it through.

That said, I think at this point we should prioritize get things done as quickly as possible, so I'd prefer to have something working first and then reconsider the design.

naoyam:
> I don't have a strong opinion about changing the name to `ElectThread`. However, all threads within the member-mask are synchronized, so there is some synchronization with this PTX instruction.

That's true. Actually, I don't this is worth worrying too much. It's just a Kernel IR type, so not exposed to the user, so never mind.

zasdfgbnm:
> That said, I think at this point we should prioritize get things done as quickly as possible, so I'd prefer to have something working first and then reconsider the design.

I agree with this, but I am still feeling uncomfortable with silent wrong results.

Today, TMA, just like any other expressions in nvFuser, fully support being launched by multiple threads. It will generate predicates like `if (threadIdx.x < 5)` if the TIDx parallelized ID has extent of 5. And this behavior is well documented in our documents [tma.md](https://github.com/NVIDIA/Fuser/blob/main/doc/dev/tma.md) and [test_tutorial.cpp](https://github.com/NVIDIA/Fuser/blob/145d15a5883b0ebc7a590c392ef2549d3dccf409/tests/cpp/test_tutorial.cpp#L942-L950).

But suddenly we introduced this arbitrary limitation that does not match with what has been documented and could provide silent wrong result. In the long term, I believe we should have as little special case as possible, which means, TMA should continue support being launched by multiple threads, and an arbitrary `threadIdx.x < 1` should be capable of being optimized as elect sync. But in the short term, I am not suggesting that this is the most important thing, but I still believe at least what is documented should match with what is implemented, which should also match with what is validated.

rdspring1:
> We should at least add some validation that the schedule is compatible with this predicate type, and update the document about the limitation.

We can have a check so that we're not launching multiple threads with any block parallel dimension.

rdspring1:
Since TMA is an async operation, I do wonder if

```cpp
if (threadIdx.x < 5) {
 // launch tma simultaneously
}
```
is better than 

```cpp
for (size_t i : irange(5)) {
  if (warp_idx == 0 && electSync()) {
    // launch tma individually
  }
}
```
Or vice versa? Probably not much difference.

naoyam:
I just realized this was extracted from #2833, which I haven't yet looked at. This PR itself doesn't have any use of the new predicate type, so it's hard to discuss how safe it would be. Let me review #2833 first.

csrc/device_lower/utils.h
rdspring1: Not at the moment. I'll remove it.

zasdfgbnm:
> Since TMA is an async operation, I do wonder if
> 
> ```c++
> if (threadIdx.x < 5) {
>  // launch tma simultaneously
> }
> ```
> 
> is better than
> 
> ```c++
> for (size_t i : irange(5)) {
>   if (warp_idx == 0 && electSync()) {
>     // launch tma individually
>   }
> }
> ```
> 
> Or vice versa? Probably not much difference.
> 
> This ^^^ isn't related to the PR. Just general curiosity.

I don't know honestly. I don't think there will be any first-order difference. But I won't be surprised if there are some second-order effects. For example, one variant uses more registers than the other, and this extra register bring the occupancy down from 2 to 1. Just want to have this flexibility to easily experiment different things.

zasdfgbnm:
> We can have a check so that we're not launching multiple threads with any block parallel dimension.

Yeah, adding a check is sufficient for now. Please also add a warning message to the doc mentioning that if you do circular buffering, you must launch TMA using one thread.

rdspring1:
!build

rdspring1:
I added the compatibility check for the `ElectSync` predicate.

Summary: In predicate lowering pass, for any If-Then-Else expression with `ElectSync` predicate, check that all output TensorViews do not have any block parallelization.

Nothing uses the `ElectSync` predicate in this PR, so I added the test to #2833. See `TEST_F(NVFuserTest, ElectSyncCompatibility)` in [f29aa22](https://github.com/NVIDIA/Fuser/pull/2833/commits/f29aa22f460bf75a7de5c010d17644c368b56fd3).

doc/dev/tma.md
```diff
@@ -349,6 +349,11 @@ the TMA domain can be completely inferred from the schedule.
 > We do not have validation on shared memory schedule yet.
 > If you scheduled something invalid, likely you will see misaligned address error or silent wrong result.
 
+> [!WARNING]
```
rdspring1, line 352: Perhaps there should be a circular buffering section but I added the warning here for now.

csrc/device_lower/pass/predicate.cpp
```diff
@@ -42,10 +42,54 @@ class ConditionalFromPredicateModifier : public kir::ExprMutator {
...
+    auto check_scope_compatibility = [](Scope& scope) {
+      for (Expr* expr : scope.exprs()) {
+        // Thread predicates are generated based on the expression's outputs
+        for (Val* val : expr->outputs()) {
+          // short-circuit
+          if (!val->isA<kir::TensorIndex>()) {
+            continue;
+          }
+          // Check that none of the IterDomains in TensorView are parallelized
+          // with a block dimension.
```
zasdfgbnm, line 62: I think I understand what you are saying. But do we have a clear definition of terminology about what a "block dimension" refers to? TIDx or BIDx?

csrc/fusion_executor/executor.cpp
```diff
@@ -1577,6 +1577,19 @@ void FusionExecutor::initializeExecutorEntry(
 
   executor_utils::validateCircularBuffering(kernel(), expr_eval);
 
+  // Check that a full warp exists in blockDim.x if the kernel contains
+  // ElectSync predicate.
+  bool has_elect_sync_predicate =
```
zasdfgbnm, line 1582: I am not very familiar with `FusionExecutor`, but my understanding is `initializeExecutorEntry` will only be invoked at the first time when we run a fusion. Which means, what could happen is, the first time when we run a fusion, a good BIDx value is provided. But later, we get a bad BIDx, but we will not run this check again. Should we make `has_elect_sync_predicate` a kernel summary (so that we don't have to run this full fusion traversal on `fusion()->vals()` over and over again), and move this check to `runFusion`?

zasdfgbnm (approved):

csrc/fusion_executor/executor.cpp
jacobhinkle: Note that `fusion()->vals()` is fast as it just returns a reference to the unordered_set of all vals registered in the `IrContainer`; it doesn't do a traversal or topo sort like `usedMathVals()` does. That also means it looks for an `ElectSync` predicate even if it's no longer "alive" in the fusion. See #2949.

csrc/device_lower/pass/predicate.cpp
rdspring1: Personally, I define block dimensions as `TIDx, TIDy, TIDz` and grid dimensions as `BIDx, BIDy, BIDz`.

I'll change `block` to `thread` in the section and refer to the parallel dimension specifically.

csrc/fusion_executor/executor.cpp
jacobhinkle: Wouldn't BDimx only change if different input sizes were given? I believe in this case, we call `initializeExecutorEntry` again.

rdspring1: `initializeExecutorEntry` is called whenever we see a new set of inputs.
Reference: https://github.com/NVIDIA/Fuser/blob/main/csrc/fusion_executor/executor.cpp#L1220-L1234

rdspring1: I'll move `has_elect_sync_predicate` to the kernel summary.

rdspring1:
!build

rdspring1:
!build

PR #2926, opened 2024-09-10
Modifying option for nvrtc

jjsjann123:
Adding `NVFUSER_ENABLE=kernel_debug` to enable debug option `-G` in nvrtc;
Moving `DebugDumpOption::DebugInfo` to `EnableOption::KernelLineInfo`.

### DIFF
--- a/csrc/fusion_executor/executor.cpp
+++ b/csrc/fusion_executor/executor.cpp
@@ -179,16 +179,16 @@ std::string FusionExecutor::getStructuredCode(
             << code << "\n======================================\n\n";
   }
   if (isDebugDumpEnabled(DebugDumpOption::CudaToFile) ||
-      isDebugDumpEnabled(DebugDumpOption::DebugInfo)) {
+      isOptionEnabled(EnableOption::KernelLineInfo))
     std::stringstream file_name;
-    file_name << "__tmp_kernel_" << kernel_id_ << ".cu";
-    debug() << "PRINTING: " << file_name.str() << std::endl;
-    std::ofstream out(file_name.str());
-    out << code << std::endl;
-    out.close();
-  }
+  file_name << "__tmp_kernel_" << kernel_id_ << ".cu";
+  debug() << "PRINTING: " << file_name.str() << std::endl;
+  std::ofstream out(file_name.str());
+  out << code << std::endl;
+  out.close();
+}
 
-  return code;
+return code;
 }
 
 std::string FusionExecutor::getStructuredCode() const {
--- a/csrc/fusion_executor/executor.cpp
+++ b/csrc/fusion_executor/executor.cpp
@@ -179,16 +179,16 @@ std::string FusionExecutor::getStructuredCode(
             << code << "\n======================================\n\n";
   }
   if (isDebugDumpEnabled(DebugDumpOption::CudaToFile) ||
-      isOptionEnabled(EnableOption::KernelLineInfo))
+      isOptionEnabled(EnableOption::KernelLineInfo)) {
     std::stringstream file_name;
-  file_name << "__tmp_kernel_" << kernel_id_ << ".cu";
-  debug() << "PRINTING: " << file_name.str() << std::endl;
-  std::ofstream out(file_name.str());
-  out << code << std::endl;
-  out.close();
-}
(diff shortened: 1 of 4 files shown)


jacobhinkle:
## Option naming

We already have `-lineinfo` which can be enabled using `NVFUSER_DUMP=debug_info`. That is similar to `-G` which also disables optimizations and is called `NVFUSER_ENABLE=jit_debug` in this PR. I would suggest we try to disambiguate these two. I think `debug_info` is not clear, while neither is the cmdline option `-lineinfo` unfortunately.

`NVFUSER_DUMP=debug_info` is mostly useful for profiling so we can see code locations with `ncu`. However, to add to the confusion we also have `NVFUSER_ENABLE=kernel_profile` which enables _intra-kernel_ profiling markers, and `NVFUSER_PROF=...` which does CUPTI profiling of kernels as well as host latency.

So I think we have four related utilities that use three different env vars. One of the utilities is clearly for functional debugging (this PR), while the others are mostly for profiling and perf debugging. I think we should rename the `NVFUSER_DUMP=debug_info` option as `NVFUSER_ENABLE=kernel_lineinfo` (it affects the generated kernel so I think it should be an `EnableOption` instead of a dump option) and as a separate issue we might want to make sure `NVFUSER_ENABLE=kernel_profile` is not too confusing: e.g. the `NVFUSER_PROF` profiler uses a structure called `KernelProfile`.

naoyam:
Do we need `-G` rather than just `--lineinfo`? The former also disables optimization.

naoyam:
Yes, that's what I'm asking about. For sanitizer, I think `--lineinfo` is sufficient. That said, `-G` may also be useful for actual debugging, so I'm fine to add that too.

csrc/fusion_executor/executor.cpp
```diff
@@ -179,7 +179,7 @@ std::string FusionExecutor::getStructuredCode(
             << code << "\n======================================\n\n";
   }
   if (isDebugDumpEnabled(DebugDumpOption::CudaToFile) ||
-      isDebugDumpEnabled(DebugDumpOption::DebugInfo)) {
+      isOptionEnabled(EnableOption::KernelLineInfo)) {
```
naoyam, line 182: It seems to me a little unexpected that an enable option dumps a file. Do we need this?

jjsjann123: I agree and I'm more than happy to remove this one. tagging @zasdfgbnm to see if there's a reason that we were dumping the cuda source in the first place.

naoyam (approved):
Thanks for the cleanup!

csrc/fusion_executor/executor.cpp
naoyam: > I agree and I'm more than happy to remove this one. tagging @zasdfgbnm to see if there's a reason that we were dumping the cuda source in the first place.

I think the original intention was indeed dumping some useful info for analyses using ncu. This can be done `NVFUSER_ENABLE=kernel_lineinfo NVFUSER_DUMP=cuda_to_file`. Now that we have (too) many options, I'd prefer each option as simple as possible.

jjsjann123: > It's probably there because the line info is pretty useless without the dumped kernel. If you only provide `NVFUSER_DUMP=debug_info` and not `cuda_to_file`, then `ncu-ui` will not show the source (it gives a file not found when you try to look at the source and asks where the file is).

:exploding_head: That's a real surprise... I though I've been using the lineinfo on my local machine when I copy over just the profile file without the actual cuda source.

zasdfgbnm: The discussion here covers all the information. I have nothing more to add, and I am OK with the change. Just a small request that we should document how to use these flags in our wiki page.

jjsjann123: good point. I'll remember to do that before merging this one.

jjsjann123: Updated in https://github.com/NVIDIA/Fuser/wiki/Developer-guide#profile-kernels

jjsjann123: FYI, you don't have to manually copy the cuda source. you can opt in with `--import-source 1` in ncu to import the source into the report, so you don't need to manually resolve that in ncu-ui.

Issue #2929, opened 2024-09-11
mpi_test.py stopped working.


wujingyue:
To reproduce, 

```
$ mpirun -np 2 pytest tests/python/mpi_test.py

_____________________________________________________________________________________ ERROR collecting tests/python/mpi_test.py ______________________________________________________________________________________
tests/python/mpi_test.py:43: in <module>
    o = fn.execute(inputs)
nvfuser/__init__.py:215: in execute
    self._setup_schedule(inputs)
E   RuntimeError:  INTERNAL ASSERT FAILED at "/opt/pytorch/nvfuser/csrc/fusion_executor/executor_utils.cpp":731, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. When trying to run the provided host program, there was an error with the provided input 0. Provided input was:
E     Tensor(sizes=[1, 4], stride=[4, 1], dtype=float, device=cuda:0, data_ptr=0x7d88c0e00000)
E     Fusion input is:
E     T0_g_float[ iS0{2}, iS1{4} ]
E     Expr eval provided the error:
E   """Tried to bind to a constant value: 2 as 1"""
E
E   Exception raised from bindInputs at /opt/pytorch/nvfuser/csrc/fusion_executor/executor_utils.cpp:731 (most recent call first):
E   frame #0: nvfuser::nvfCheckFail(char const*, char const*, unsigned int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) + 0xf3 (0x7d88d979b17f in /opt/pytorch/nvfuser/nvfuser/_C.cpython-310-x86_64-linux-gnu.so)
E   frame #1: nvfuser::nvfErrorFail(char const*, char const*, unsigned int, char const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) + 0x53 (0x7d88d9b00223 in /opt/pytorch/nvfuser/nvfuser/_C.cpython-310-x86_64-linux-gnu.so)
E   frame #2: <unknown function> + 0x13ce26 (0x7d88d97a6e26 in /opt/pytorch/nvfuser/nvfuser/_C.cpython-310-x86_64-linux-gnu.so)
E   frame #3: <unknown function> + 0x45668c (0x7d88d9ac068c in /opt/pytorch/nvfuser/nvfuser/_C.cpython-310-x86_64-linux-gnu.so)
[...]

PR #2931, opened 2024-09-11
Add test_multidevice.py.

wujingyue:
This PR does the following: 
1. Move the only test in mpi_test.py to test_multidevice.py, so it will be [picked up](https://gitlab-master.nvidia.com/dl/pytorch/fuser-gh-mirror/-/blob/github-ci/jit/jit.yml#L93. ) by CI. 
2. Because the test requires MPI, decorate the test with `@pytest.mark.mpi` whose dependencies was added in [this PR](https://gitlab-master.nvidia.com/dl/pytorch/update-scripts/-/merge_requests/132). This way, the test will only be triggered by `mpirun ... pytest ... --with-mpi`. 
3. The test however needs to be disabled due to #2929. 
4. Add another test that exercises just MPI, and fix a bug triggered by this test.


wujingyue:
!build

wujingyue:
!build

wujingyue:
!build

csrc/multidevice/communicator.cpp
```diff
@@ -212,14 +212,17 @@ Communicator::Communicator(
...
+  NVF_CHECK(
+      !cleaned_up,
+      "The singleton Communicator has already been cleaned up. This is "
+      "likely because Communicator::cleanup was called more than once");
   cleaned_up = true;
 
+  // Without this, the TCPStore server can be cleaned up before TCPStore
+  // clients are created, causing an hang. This happened with
+  // test_multidevice.py::test_sizes_and_ranks.
+  barrier();
```
jjsjann123, line 224: :clap:

csrc/multidevice/communicator.cpp
```diff
@@ -212,14 +212,17 @@ Communicator::Communicator(
 
 void Communicator::cleanup() {
   static bool cleaned_up = false;
-  if (cleaned_up) {
```
jjsjann123, line 215: QQ: I forgot why we didn't have this as a check earlier but rather just a warning.

jjsjann123: Anyway, glad to see it converted to an error instead.

jjsjann123 (approved):
LGTM.

QQ: do we need to add another entry in CI script to actually pick up mpi test?

Looking at this https://pytest-mpi.readthedocs.io/en/latest/usage.html

Seems like it needs to run pytest under MPI, so we need to change the CI script as well?

wujingyue:
Yes. Haven't done that yet

csrc/multidevice/communicator.cpp
wujingyue: I probably should have. I didn't think about it too much because the rest of the function was safe enough to run again. However, with this PR and thus the added call to barrier, `cleanup` won't be able to run twice.

PR #2932, opened 2024-09-11
Slice python api

jjsjann123:


rdspring1:
Do you have any plan to update the composite operation API at https://github.com/NVIDIA/Fuser/blob/main/csrc/ops/alias.h#L119-L136?

It only takes `std::vector<int64_t>` for the slice range instead of `std::vector<Val*>`.

jjsjann123:
> Do you have any plan to update the composite operation API at https://github.com/NVIDIA/Fuser/blob/main/csrc/ops/alias.h#L119-L136?
> 
> It only takes `std::vector<int64_t>` for the slice range instead of `std::vector<Val*>`.

Nope. I can just use the very first overload with `std::vector<Slice>`, that would support dynamic use case as well.

jjsjann123:
!build

csrc/python_frontend/fusion_record.h
```diff
@@ -366,6 +366,57 @@ struct OpRecord : RecordFunctor {
...
+    const std::vector<Val*>& stride =
+        fd.getFusionStateVector(args_.at(3).index);
+    std::vector<Slice> vec_slice;
+    for (const auto idx : c10::irange(arg->nDims())) {
+      // NOTE: there's an extra move, we can use emplace_back if we go write
+      // some constructors for Slice.
+      Val* start_idx = start.at(idx);
+      Val* end_idx = end.at(idx);
+      Val* stride_idx = stride.at(idx);
+      NVF_CHECK(
```
jjsjann123, line 398: These checks are now inside `finalizeDefinition` instead of on the constructor. Hence exposes the issue on our FusionCache #2953

tests/python/test_python_frontend.py
```diff
@@ -4325,6 +4325,38 @@ def fusion_func(fd: FusionDefinition) -> None:
...
+                dtype=DataType.Float,
+                is_cpu=False,
+                stride_order=[2, 1, 0],
+            )
+            T1 = fd.ops.slice(
+                T0, start_indices=offset, end_indices=(2, 5, 10), strides=(1, 1, 1)
+            )
+            fd.add_output(T1)
+            V_start = fd.define_vector(offset)
+            V_end = T0.shape()
```
jacobhinkle, line 4346: Off-topic: would it be better to bind `Tensor.shape` using [def_property](https://pybind11.readthedocs.io/en/stable/classes.html#instance-and-static-fields) so we could use `T0.shape` instead of `T0.shape()`?

csrc/python_frontend/fusion_record.h
```diff
@@ -366,6 +366,57 @@ struct OpRecord : RecordFunctor {
...
+  void operator()(FusionState& fd) final {
+    TensorView* arg = fd.getFusionState(args_.at(0).index)->as<TensorView>();
+    const std::vector<Val*>& start = fd.getFusionStateVector(args_.at(1).index);
+    const std::vector<Val*>& end = fd.getFusionStateVector(args_.at(2).index);
+    const std::vector<Val*>& stride =
+        fd.getFusionStateVector(args_.at(3).index);
+    std::vector<Slice> vec_slice;
+    for (const auto idx : c10::irange(arg->nDims())) {
+      // NOTE: there's an extra move, we can use emplace_back if we go write
+      // some constructors for Slice.
```
jacobhinkle, line 394: You could also do the following
```c++
vec_slice.emplace_back();
vec_slice.back().start = start_idx;
vec_slice.back().end = end_idx;
vec_slice.back().stride = stride_idx;
```
But I agree adding a constructor for `Slice` and using `vec_slice.emplace_back(start_idx, end_idx, stride_idx)` is simplest and clearest...

csrc/python_frontend/python_bindings.cpp
```diff
@@ -235,6 +241,77 @@ struct DimInfo {
   }
 };
 
+template <class ShapeType>
+Tensor slice_fn(
+    FusionDefinition::Operators& self,
+    Tensor arg,
+    ShapeType start,
+    ShapeType end,
+    std::optional<ShapeType> strides) {
```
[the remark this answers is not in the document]
jjsjann123: That's enough for thunder. For reference https://github.com/Lightning-AI/lightning-thunder/blob/9580411e1dcec34ce0226a5a688ed65c32422b8b/thunder/executors/nvfuserex_impl.py#L1253

We can control how the API is used. That's my excuse of not plumbing these extra APIs.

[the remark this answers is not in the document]
jjsjann123: I'm leaning towards leaving the API as-is for a somewhat stable API...

csrc/python_frontend/fusion_record.h
jjsjann123: I'm leaning towards leaving this as-is since it's just a move. I can have a follow up clean up if you think this bothers too much :laughing:

tests/python/test_python_frontend.py
jjsjann123: I don't have a strong opinion one way or the other. :stuck_out_tongue_winking_eye:

jjsjann123:
!build

tests/python/test_python_frontend.py
jacobhinkle: Me either, so let's leave it as is :-)

csrc/python_frontend/fusion_record.h
jacobhinkle: > leaving this as-is

All good with me

jjsjann123:
!build

jjsjann123:
!build

jjsjann123:
!build

jjsjann123:
!build

jjsjann123:
!build

PR #2933, opened 2024-09-11
Introduce TaggedAbstractTensor

jacobhinkle:
This generalizes `AbstractTensor` to the templated struct `AbstractTensorWithInfo<Info>` and introduces a special case subclass `TaggedAbstractTensor<Tag>`. This can be used by passing an enum class for `Tag`, and holds an `unordered_set<Tag>` for each dimension. Merging and swizzling unions these sets, and split duplicates the set.

Note that a lot of code had to be moved out of the cpp into the header because of templatization. However, there are no changes to the `Dispatch*` classes. The `AbstractTensorWithInfo` methods like split, merge, swizzle, etc. are just changed to add calls to `Info::merge`.

Related to #2913, which specializes this as `using AbstractMatmulTensor = TaggedAbstractTensor<MatmulDimRole>`.


jacobhinkle:
!build

csrc/abstract_tensor.h
```diff
@@ -253,70 +580,262 @@ struct AbstractTensor {
...
+    return false;
   }
 
   template <typename T>
   bool operator!=(T&& t) const {
     return !operator==(std::forward<T>(t));
   }
 
-  AbstractTensor& parallelize(int64_t axis, ParallelType parallel_type);
+  AbstractTensorWithInfo<Info>& parallelize(
```
zasdfgbnm, line 616: nit: I believe `AbstractTensorWithInfo& parallelize(` should be sufficient. C++ will automatically know that `AbstractTensorWithInfo` refers to `AbstractTensorWithInfo<Info>` in this context.

csrc/abstract_tensor.h
```diff
@@ -253,70 +580,262 @@ struct AbstractTensor {
...
+    auto domain_inner_pos = axis_o < axis_i ? axis_i : axis_o;
+
+    domain.erase(domain.begin() + domain_inner_pos);
+    std::swap(domain[domain_outer_pos], output);
 
-  AbstractTensor& split(int64_t axis, Val* factor, bool inner_split = true);
-  AbstractTensor& split(int64_t axis, int64_t factor, bool inner_split = true);
+    info[domain_outer_pos] =
+        Info::merge(info[domain_outer_pos], info[domain_inner_pos]);
+    info.resize(info.size() - 1);
```
zasdfgbnm, line 663: What happens with this resize? Does it drop the last item (which is not what we wanted)?

csrc/abstract_tensor.h
```diff
@@ -253,70 +580,262 @@ struct AbstractTensor {
...
+      int64_t axis,
+      Val* factor,
+      bool inner_split = true) {
+    axis = wrapDim(axis, (int64_t)domain.size());
+    auto [outer, inner] = AbstractId::dispatch(
+        DispatchSplit{}, domain[axis], factor, inner_split);
+    std::swap(domain[axis], inner);
+    domain.insert(domain.begin() + axis, outer);
+
+    info.insert(info.begin() + axis, info[axis]);
```
zasdfgbnm, line 634: Do we need a `Info::split`?

csrc/abstract_tensor.h
```diff
@@ -332,10 +851,131 @@ struct AbstractTensor {
...
+      std::vector<std::unordered_set<Tag>> tag_sets)
+      : AbstractTensorWithInfo<TagSetInfo<Tag>>(domain) {
+    NVF_ERROR(
+        this->info.size() == tag_sets.size(),
+        "Mismatch in length between provided domain and tag sets");
+    size_t i = 0;
+    for (const auto& s : tag_sets) {
+      this->info[i++].tags.insert(s.begin(), s.end());
+    }
+  }
```
zasdfgbnm, line 931: nit: If we add a new ctor `TagSetInfo(std::unordered_set<Tag>)`, then we can simplify this as:
```suggestion
  TaggedAbstractTensor(
      std::vector<AbstractId> domain,
      std::vector<std::unordered_set<Tag>> tag_sets)
      : AbstractTensorWithInfo<TagSetInfo<Tag>>(domain, {tag_sets.begin(), tag_sets.end()}) {}
```

csrc/abstract_tensor.h
jacobhinkle: Ah of course. Switched to `erase` instead and added a test. Thanks for noticing!

csrc/abstract_tensor.h
jacobhinkle: It is easy to add that, but since generic copy is sufficient for the matmul case I didn't add an interface for all of the scheduling primitives. However, I think it's probably clearer to do so. Especially since that gives me a place to document what we do for swizzle (treated same as merge in `TagSetInfo`).

csrc/abstract_tensor.h
jacobhinkle: Great suggestion. That did clean up the `TaggedAbstractTensor` constructors.

csrc/abstract_tensor.h
```diff
@@ -253,70 +580,260 @@ struct AbstractTensor {
...
+            "Can not unzip an AbstractTensor with different sizes in its domains.");
+      }
+    }
+
+    // unzip the AbstractTensor, broadcast the non-vector items
+    result.resize(size);
+    for (const auto& aid : domain) {
+      for (auto i : c10::irange(size)) {
+        if (!aid.is<std::vector>()) {
+          result[i].domain.emplace_back(aid);
```
zasdfgbnm, line 801: should we do a ``result[i].info.emplace_back(info);``?

csrc/abstract_tensor.h
```diff
@@ -253,70 +580,260 @@ struct AbstractTensor {
...
+    }
+
+    result.domain.reserve(tensors[0].domain.size());
+    for (auto i : c10::irange(tensors[0].domain.size())) {
+      std::vector<AbstractId> domain;
+      domain.reserve(tensors.size());
+      for (auto& tensor : tensors) {
+        domain.emplace_back(std::move(tensor.domain[i]));
+      }
+      result.domain.emplace_back(std::move(domain));
```
zasdfgbnm, line 832: Should we do a
```C++
NVF_CHECK(std::all_of(tensors.begin(), tensors.end(), [](auto t) { return t.info[i] == tensors[0].info[i]; });
result.info.emplace_back(std::move(tensors[0].info[i]));
```

csrc/abstract_tensor.h
```diff
@@ -332,10 +849,139 @@ struct AbstractTensor {
...
+        "Can not add a new row with different number of domains.");
+    NVF_CHECK(
+        std::all_of(
+            domain.begin(),
+            domain.end(),
+            [](const AbstractId& aid) { return aid.is<std::vector>(); }),
+        "Can not add a new row to an AbstractTensor with non-vector domains.")
+
+    for (auto i : c10::irange(domain.size())) {
+      domain[i].as<std::vector>().emplace_back(std::move(tensor.domain[i]));
```
zasdfgbnm, line 864: Should we do a
```C++
NVF_CHECK(return info[i] == tensor.info[i]);
result.info.emplace_back(std::move(tensor.info[i]));
```

csrc/abstract_tensor.h
```diff
@@ -332,10 +849,139 @@ struct AbstractTensor {
...
 
   // Remove all the null elements.
-  AbstractTensor& strip();
+  AbstractTensorWithInfo& strip() {
+    AbstractTensorWithInfo result;
+    for (const auto& aid : domain) {
+      if (aid.hasValue()) {
+        result.pushBack(aid);
+      }
+    }
```
zasdfgbnm, line 876: Should this be something like:
```C++
    for (auto& [aid, info] : domainAndInfo()) {
      if (aid.hasValue()) {
        result.pushBack(std::move(aid), std::move(info));
      }
    }
```

csrc/abstract_tensor.h
```diff
@@ -332,10 +849,139 @@ struct AbstractTensor {
...
+using AbstractTensor = AbstractTensorWithInfo<EmptyInfo>;
+
+//! This is a holds a comparable type of "tag". This version is strict in the
+//! sense that we check that the tags match whenever we do a merge instead of
+//! allowing mismatches.
+template <typename Tag>
+struct TagSetInfo {
+  std::unordered_set<Tag> tags;
+
+  TagSetInfo(std::unordered_set<Tag> tags_) : tags(tags_) {}
```
zasdfgbnm, line 911: nit:
```suggestion
  TagSetInfo(std::unordered_set<Tag> tags_={}) : tags(std::move(tags_)) {}
```

csrc/abstract_tensor.h
```diff
@@ -332,10 +849,139 @@ struct AbstractTensor {
...
+  bool operator==(const EmptyInfo& t) const {
+    return true;
+  }
+};
+
+using AbstractTensor = AbstractTensorWithInfo<EmptyInfo>;
+
+//! This is a holds a comparable type of "tag". This version is strict in the
+//! sense that we check that the tags match whenever we do a merge instead of
+//! allowing mismatches.
```
zasdfgbnm, line 906: Where is the code that does the check:
> This version is strict in the sense that we check that the tags match whenever we do a merge instead of allowing mismatches.

csrc/abstract_tensor.h
```diff
@@ -332,10 +849,139 @@ struct AbstractTensor {
...
+template <typename Tag>
+struct TagSetInfo {
+  std::unordered_set<Tag> tags;
+
+  TagSetInfo(std::unordered_set<Tag> tags_) : tags(tags_) {}
+
+  static TagSetInfo<Tag> merge(
+      const TagSetInfo<Tag>& a,
+      const TagSetInfo<Tag>& b) {
```
zasdfgbnm, line 915: nit:
```suggestion
  static TagSetInfo merge(
      const TagSetInfo& a,
      const TagSetInfo& b) {
```

csrc/abstract_tensor.h
```diff
@@ -332,10 +849,139 @@ struct AbstractTensor {
...
+
+  //! Swizzling mixes the tags so here we re-use merge and duplicate the result
+  static std::pair<TagSetInfo<Tag>, TagSetInfo<Tag>> swizzle(
+      const TagSetInfo<Tag>& a,
+      const TagSetInfo<Tag>& b) {
+    TagSetInfo<Tag> merged_info = merge(a, b);
+    return {merged_info, merged_info};
+  }
+
+  bool operator==(TagSetInfo<Tag>&& t) const {
```
zasdfgbnm, line 934:
```suggestion
  bool operator==(const TagSetInfo<Tag>& t) const {
```

csrc/abstract_tensor.h
```diff
@@ -332,10 +849,139 @@ struct AbstractTensor {
...
+
+//! This is a special case of AbstractTensorWithInfo which propagates a set of
+//! tags for each axis. The tags can be any hashable type with equality;
+//! typically Tag would be an enum.
+template <typename Tag>
+struct TaggedAbstractTensor : AbstractTensorWithInfo<TagSetInfo<Tag>> {
+  TaggedAbstractTensor(
+      std::vector<AbstractId> domain,
+      const std::vector<std::unordered_set<Tag>>& tag_sets)
+      : AbstractTensorWithInfo<TagSetInfo<Tag>>(
```
zasdfgbnm, line 947: nit:
```suggestion
      : AbstractTensorWithInfo(
```

csrc/abstract_tensor.h
jacobhinkle: Error in the comment. It should say we just union the sets upon merge or swizzle. Will update.

csrc/abstract_tensor.h
jacobhinkle: Yes to the first line, but we don't need to modify `info`, just check that it matches the new row.

csrc/abstract_tensor.h
```diff
@@ -332,10 +849,139 @@ struct AbstractTensor {
...
+  TaggedAbstractTensor(
+      const std::vector<IterDomain*>& domain,
+      const std::vector<std::unordered_set<Tag>>& tag_sets)
+      : AbstractTensorWithInfo<TagSetInfo<Tag>>(
+            domain,
+            {tag_sets.begin(), tag_sets.end()}) {}
+  TaggedAbstractTensor(
+      std::initializer_list<AbstractId> domain,
+      std::initializer_list<std::initializer_list<Tag>> tag_sets)
+      : AbstractTensorWithInfo<TagSetInfo<Tag>>(
```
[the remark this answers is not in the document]
jacobhinkle: I thought I could get rid of these also but clang complains if I even remove `Tag` here. :man_shrugging:

csrc/abstract_tensor.h
```diff
@@ -332,10 +849,139 @@ struct AbstractTensor {
...
+            domain,
+            {tag_sets.begin(), tag_sets.end()}) {}
```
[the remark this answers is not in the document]
jacobhinkle: This actually didn't work for me because although there is an `unordered_set` ctor for `TagSetInfo`, there's no `vector<unordered_set<Tag>>` ctor for `vector<TagSetInfo>`.

tests/cpp/test_abstract_tensor.cpp
```diff
@@ -796,4 +796,71 @@ TEST_F(AbstractTensorTest, AddRow) {
...
+    EXPECT_TRUE(v0.hasTag(-1, TestTag::A));
+    EXPECT_EQ(v0.getTag(1), TestTag::A);
+
+    v0.merge(0);
+
+    EXPECT_EQ(v0.getTag(0), TestTag::A);
+  }
+
+  {
+    // Mismatched tags should throw an error due to strict tag matching
```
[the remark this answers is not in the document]
jacobhinkle: Removed this stale comment.

jacobhinkle:
!build

csrc/abstract_tensor.h
```diff
@@ -613,6 +613,10 @@ struct AbstractTensorWithInfo {
   bool operator==(T&& t) const {
     if constexpr (std::is_same_v<AbstractTensorWithInfo, std::decay_t<T>>) {
       return domain == t.domain && info == t.info;
+    } else if constexpr (std::is_same_v<
+                             std::vector<AbstractId>,
+                             std::decay_t<T>>) {
+      return domain == t;
```
jacobhinkle, line 619: This fixes a failure in the `Reorder` test where we were no longer automatically casting `vector<AbstractId>` to `AbstractTensorWithInfo` so this returned false even with a matching domain provided.

jacobhinkle: To avoid this kind of silent error again, should we just remove the template `T` here and implement this only for `AbstractTensorWithInfo` and `std::vector<AbstractId>`?

zasdfgbnm: Makes sense to me.

tests/cpp/test_abstract_tensor.cpp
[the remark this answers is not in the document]
zasdfgbnm: Looks like this comment is not removed

zasdfgbnm (approved):

tests/cpp/test_abstract_tensor.cpp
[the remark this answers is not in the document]
jacobhinkle: Done now.

Issue #2930, opened 2024-09-11
`FusionReductionWithTrivialReduction_CUDA` fails with compute-sanitizer


xwang233:
repro (pjnl-20240910):
```
NVFUSER_ENABLE=kernel_debug PYTORCH_NO_CUDA_MEMORY_CACHING=1 compute-sanitizer bin/nvfuser_tests --gtest_filter='*FusionReductionWithTrivialReduction_CUDA*'
```

sample stacktrace:
```
========= Invalid __local__ write of size 16 bytes
=========     at void <unnamed>::loadGeneric<float, (int)4>(T1 *, T1 *)+0x830 in /opt/pytorch/nvfuser/__tmp_kernel_reduction_f0_c1_r0_g5.cu:2623
=========     by thread (3,4,0) in block (71,0,0)
=========     Address 0xfffa94 is misaligned
=========     Saved host backtrace up to driver entry point at kernel launch time
=========     Host Frame: [0x2dcdbf]
=========                in /usr/local/cuda/compat/lib.real/libcuda.so.1
=========     Host Frame:nvfuser::FusionExecutor::runFusion(nvfuser::KernelArgumentHolder&, nvfuser::LaunchParams const&, nvfuser::CompileParams, std::vector<at::Tensor, std::allocator<at::Tensor> >) [0x55f85e]
=========                in /opt/pytorch/nvfuser/bin/nvfuser_tests
=========     Host Frame:nvfuser::FusionKernelRuntime::runKernelWithInput(nvfuser::KernelArgumentHolder&, nvfuser::SegmentedGroup*) [0x721c9d]
=========                in /opt/pytorch/nvfuser/bin/nvfuser_tests
=========     Host Frame:nvfuser::FusionKernelRuntime::runSegmentsWithInputs(nvfuser::KernelArgumentHolder&) [0x72676c]
=========                in /opt/pytorch/nvfuser/bin/nvfuser_tests
=========     Host Frame:nvfuser::FusionKernelRuntime::runWithInputs(nvfuser::KernelArgumentHolder&) [0x726c49]
=========                in /opt/pytorch/nvfuser/bin/nvfuser_tests
```

line of code at `/opt/pytorch/nvfuser/__tmp_kernel_reduction_f0_c1_r0_g5.cu:2623`:
```cuda
template <typename scalar_t, int vec_size>
__device__ void loadGeneric(scalar_t* to, scalar_t* from) {
  // It would be really nice to use memcpy here, but one example was failing
[...]

PR #665, opened 2023-07-31
Alias allocations with same size elements despite different dtypes

jacobhinkle:
Currently the `reuseMemoryAllocations` lowering pass reuses `TensorView`s if they have compatible shapes and vectorization, and the same dtype. This PR replaces the dtype condition with a check that the size of each element is the same.

It is somewhat rare in our test suite that we find opportunities to re-use one buffer with another of a different type, but it does happen. For example, one case in `FusionWelfordShmoo_CUDA` corresponding to `testWelford(DataType::ComplexFloat, 1, 320, 256);` currently shows:
```
ptxas info    : Used 61 registers, 16 bytes smem, 464 bytes cmem[0]
```
With the current PR, this becomes
```
ptxas info    : Used 59 registers, 16 bytes smem, 464 bytes cmem[0]
```
since we re-use two register `TensorView`s that change from `int64_t` to `std::complex<float>`.

Re-using shared memory in these cases is likely always beneficial, but the effect of explicitly re-using registers is hard to analyze since the compiler pipeline is so complicated. It is likely rarely harmful, so this behavior is on by default. However, it can be disabled using `DisableOption::ReuseMismatchedTypeRegisters` or `NVFUSER_DISABLE=reuse_mismatched_type_registers`.

PR #2934, opened 2024-09-11
use aligned array for iter grouped reduction inputs

liqiangxl:
Fix https://github.com/NVIDIA/Fuser/issues/2930

Using aligned array of registers when needs to vectorized data transfer between registers and shared memory.


liqiangxl:
!build

liqiangxl:
!build

liqiangxl:
!build

liqiangxl:
Failed tests do not seem to be related to this PR. `thunder.tests.test_grad.test_vjp_correctness_sdpa_manual_grad_forward_scaled_dot_product_attention_nvfuser_cuda_thunder.dtypes.bfloat16`

csrc/codegen.cpp
```diff
@@ -1846,14 +1846,8 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
...
     ArgumentBuilder func_args(block_nest_level_ + 1, kTab);
-    auto output_tv = output->view();
-    auto va = kernel_->summary().vectorized_accesses;
-    if (va.find(output_tv) != va.end()) {
-      func_args.arg(genVariableName(output) + ".array");
-    } else {
-      func_args.arg(genVariableName(output));
-    }
-    func_args.arg(genVariableName(input));
+    func_args.arg(genVariableNameConvertAlignedArray(output));
```
naoyam, line 1849: Is this just a cleanup?

csrc/codegen.cpp
```diff
@@ -2762,14 +2756,8 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
...
     ArgumentBuilder func_args;
-    auto output_tv = output->view();
-    auto va = kernel_->summary().vectorized_accesses;
-    if (va.find(output_tv) != va.end()) {
-      func_args.arg(genVariableName(output) + ".array");
-    } else {
-      func_args.arg(genVariableName(output));
-    }
-    func_args.arg(genVariableName(input));
+    func_args.arg(genVariableNameConvertAlignedArray(output->view()));
```
naoyam, line 2759: Same here

csrc/codegen.cpp
```diff
@@ -2995,9 +2987,17 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
...
           auto va = kernel_->summary().vectorized_accesses;
+          unsigned int vect_factor = 0;
           if (va.find(tv) != va.end()) {
+            vect_factor = va.at(tv);
+          } else if (
+              kernel_->summary().num_grouped_iterations > 1 &&
+              ir_utils::isConsumedByIterGroupedReduction(tv)) {
+            vect_factor = kernel_->summary().num_grouped_iterations;
+          }
+          if (vect_factor > 0) {
```
naoyam, line 2998: When can this be 0?

csrc/codegen.cpp
```diff
@@ -2995,9 +2987,17 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
           break;
         case MemoryType::Local: {
           auto va = kernel_->summary().vectorized_accesses;
+          unsigned int vect_factor = 0;
           if (va.find(tv) != va.end()) {
+            vect_factor = va.at(tv);
+          } else if (
+              kernel_->summary().num_grouped_iterations > 1 &&
+              ir_utils::isConsumedByIterGroupedReduction(tv)) {
+            vect_factor = kernel_->summary().num_grouped_iterations;
```
naoyam, line 2996: In general, `Kernel` should already have the correct vectorization factor and `CudaCodeGen` should just be a straightforward printer. Is there any specific reason for this case?

naoyam (commented):
Only about codegen.cpp for now

csrc/codegen.cpp
liqiangxl: Yes. `genVariableNameConvertAlignedArray` detects whether we should append `.array` or not.

csrc/codegen.cpp
liqiangxl: yes, just a clean up, no logical change.

csrc/codegen.cpp
liqiangxl: `Kernel` only has vectorized factor for global memory access, not for shared memory access.I added comments for clarity.
```
Should use aligned array of registers when:
(1) vectorized ld/st with global memory, tv exists in kernel summary vectorized_accesses.
(2) vectorized ld/st with shared memory, tv is input to iteration grouped reduction and vectorized in runtime function.
```
Another option, is adding these tvs with vectorized shared memory access to `kernel_->summary().vectorized_accesses`.

csrc/codegen.cpp
liqiangxl: It is initilized to 0, so stays at 0 when the tv is not vectorized with gmem or smem.

csrc/codegen.cpp
naoyam: It seems strange to me that the kernel summary only has info about global memory tensors but not shared memory. Is there any reason for that?

liqiangxl: > It seems strange to me that the kernel summary only has info about global memory tensors but not shared memory. Is there any reason for that?

To be more precise, should be `kernel summary doesn't have info about vectorized access in runtime functions, e.g. blockIterGroupedYdimReduce()`

naoyam: I understand that. Does that matter?

liqiangxl:
!build

liqiangxl:
!build

csrc/codegen.cpp
liqiangxl: Yes, that matters. If a tv is consumbed by a `IterGroupedReduction`, then it is a input to runtime function `blockIterGroupedYdimReduce` and will be vectorized loaded to smem within that runtime function, so we need to allocate it as aligned array.

naoyam: Yes, I understand that. Does that matter for my question for this?

> It seems strange to me that the kernel summary only has info about global memory tensors but not shared memory. Is there any reason for that?

liqiangxl: oh, I mean this statement is not true. Kernel summary has info about vectorized r/w for both global and shared memories, if these vectorized r/w were generated by the scheduler.
It does't have the info about vectorized r/w implemented in the runtime functions.

naoyam: I know that. I'm asking why.

liqiangxl: Becase it doesn't check the runtime functions. These runtime functions are just `strings` when generating kernel.

naoyam: It isn't necessary to do so, right? Your PR doesn't do that either. You check some condition in CudaCodeGen. All I'm asking is why it cannot be done when generating the kernel summary. Let me say this again:

> In general, Kernel should already have the correct vectorization factor and CudaCodeGen should just be a straightforward printer.

liqiangxl: So you are suggesting doing option-2?

> Another option, is adding these tvs with vectorized shared memory access to kernel_->summary().vectorized_accesses.

No problem. we can do that.

liqiangxl:
!build --diff

liqiangxl:
Revised to use option-2, where the tvs used in iter gouped reductions are added to `kernel_->summary().vectorized_accesses.`
Step-1, mark `ParallelType::Group` as a special vectorization.
```
      // ParallelType::Group is used for both iteration grouped reduction and
      // fused reductions where 2 reductions are fused into one. Vectorized
      // access to shared memory only exists in the first case.
      if (ptype == ParallelType::Group) {
        auto def = tv->definition();
        auto grop = dynamic_cast<GroupedReductionOp*>(def);
        if (grop->numHorizontallyGroupedExprs() == 1) {
          has_grouped_vectorize_dim = true;
        }
      }
```
Step-2, call `VectorizeValidator` where tv and its producer is added to `vectorizedSetInfo` and `vectorizedAccesses`
 ```
    if (has_vectorize_dim || has_misaligned_vectorize_dim ||
        has_grouped_vectorize_dim) {
      VectorizeValidator::validate(tv);
    }
```

**Comment by liqiangxl** [comment]
!build --diff

**Comment by liqiangxl** [comment]
!build --diff

**Comment by liqiangxl** [comment]
!build --diff

**Comment by liqiangxl** [comment]
@naoyam , I think it's ready for another round of review. codediff is expected, distributd failed tests are not related to this PR.

File: csrc/codegen.cpp [hunk 5]
```diff
@@ -2972,6 +2960,10 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
           aligned_array_of_regs_.insert(tv);
         }
       }
+      // tv is aligned if alias is aligned
```
naoyam, line 2963: I'm not sure what we are doing here, not just the change by this PR. When we have an `Allocate` node, does it automatically mean if it's an alias of another tensor but with a different type? It seems that's what is indicated by line 2956 since it does `reinterpret_cast` to an Array type. Is this really safe? Don't we need to check the type of the original allocation of the alias tensor?

naoyam: Looks like that was originally introduced in #665. What do you think, @jacobhinkle?

csrc/codegen.cpp
```diff
@@ -2972,6 +2960,10 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
...
+      if (aligned_array_of_regs_.count(alias_tv) > 0) {
+        aligned_array_of_regs_.insert(tv);
```
naoyam, line 2965: What is this variable used for? It seems it's only used by `genVariableNameConvertAlinedArray`, which is only used by `generateGroupedGridAllReduceWelfordOuter`. I wonder if this change matters for this PR. Also, it doesn't seem to be used consistently. Am I understanding correctly?

csrc/fusion_executor/executor_utils.cpp
```diff
@@ -299,7 +299,8 @@ std::unique_ptr<caching::VectorizedTensorInfo> getVectorizedTensorValidationInfo
 
     auto vector_dim = vector_info.vectorized_loop_id;
     const auto is_aligned =
-        vector_dim->getParallelType() == ParallelType::Vectorize;
+        vector_dim->getParallelType() == ParallelType::Vectorize ||
+        vector_dim->getParallelType() == ParallelType::Group;
```
naoyam, line 303: Why is this change necessary?

csrc/codegen.cpp
naoyam: Never mind, using it was part of the PR changes.

naoyam:
I'm not sure why we also need to change the executor as well as the vectorization validator. I thought what we are missing is using aligned arrays inside some of the device functions and they are just local temporary arrays, so those are not something we would need to validate, right? What are we validating then?

csrc/codegen.cpp
jacobhinkle: > Is this really safe? Don't we need to check the type of the original allocation of the alias tensor?

Isn't that safe if the two types have the same `sizeof`? That is what #665 is doing.

naoyam: Sorry, my question was incomplete. If the original tensor is defined as a non-aligned tensor, is it safe to reinterpret-cast it to an aligned array type? Isn't that what could happen?

jacobhinkle: That's right if the original tensor had a different alignment from the new tensor. So the danger here would be for example if the original tensor had vectorized access of width 2 then the second tensor which tries to alias it has same sizeof(dtype) but it has a vectorized access of width 8. I didn't think about that in #665 but yes I think we should either guard against that when we're setting up the alias or accomodate the alias vectorized accesses when we codegen the original allocation.

naoyam: Yeah, as far as I can see, the original tensor doesn't even seem to be guaranteed to be aligned at any size.

This doesn't need to be fixed in this PR, but please create an issue.

jacobhinkle: Actually, I forgot about this since I haven't looked at the aliasing code in a while, but we already do this analysis to guarantee the alias vectorizations are at most the same width as the original: https://github.com/NVIDIA/Fuser/blob/d610ce84c25307b9926df1a92e9f928b8e59369c/csrc/device_lower/pass/alias_memory.cpp#L1243-L1247

naoyam: Oh, I see. Glad it'a a false alarm.

naoyam: Actually, I wonder if it should specify the alignment size with `reinterpret_cast`. 

```
<< " = *reinterpret_cast<Array<" << buffer_dtype << ", "
<< genInline(size) << ">*>(&" << genVariableName(alias_tv)
```

If I read this correctly, it just uses an Array type with no alignment requirement, for example, `Array<float, 8>`. The default alignment size is 1, so it seems this would tell the compiler that we are using a non-aligned address with vector loads and stores. The address is indeed aligned, so it should have no problem unless the compiler does something when a given address is marked not aligned properly. I think it'd be safer to use a proper aligned type always even when we know it's properly aligned.

naoyam, line 2963: Why is this? I understand this is just a naming difference, but whether the original allocation is aligned or not shouldn't matter for the aliasing tensor, right? For example, this `tv` can be a tensor with no alignment requirement, right?

csrc/fusion_executor/executor_utils.cpp
liqiangxl: Because grouped reduction tv is added to `vectorized_set_info`,  if not aligned, must be fusion input or output.
If don't directly using `VectorizeValidator::validate(tv)`, won't be added to `vectorized_set_info` and this change is not required.

liqiangxl:
!build

csrc/codegen.cpp
jacobhinkle: Yeah, it's probably not a bad idea to add it: https://github.com/NVIDIA/Fuser/pull/3084

liqiangxl: On the other hand, `T29` does't have to be aligned. We can use dynamic cast to remove this. But then we need to change the code of alias allocation.

csrc/device_lower/validation.cpp
```diff
@@ -598,7 +598,25 @@ void validateAndCollectVectorizeInfo(Fusion* fusion) {
             "Only allow misaligned vectorization between global and local memory.");
         has_misaligned_vectorize_dim = true;
       }
+
+      // ParallelType::Group is used for both reduction & normalization.
```
naoyam, line 602: Is this really necessary? Shared memory is always aligned. If the producer is in global memory, then that should be a temporary work buffer, so it should be always a contiguous, aligned buffer.

liqiangxl: I mean between register and shared memory, so we need to ensure the registers are aligned when doing vectorized read/write.

liqiangxl: In this case, `T13 = GroupedReductionOf(T14)`, `T14 (registers)` should be aligned becase runtime funciton uses vectorized copy from `T14(registers)` to shared memory.
```
T13_l_float[ iblockIdx.x68{( ceilDiv(( ceilDiv(( ceilDiv(i2, 4) ), blockDim.x) ), 1) )}, ithreadIdx.x67{blockDim.x}, iUS69{1}, iG65{4}, rthreadIdx.y62{blockDim.y} ] ca_pos( 3 ) produce_pos( 2 )
   = reduction( T14_l_float[ iblockIdx.x54{( ceilDiv(( ceilDiv(( ceilDiv(i2, 4) ), blockDim.x) ), 1) )}, ithreadIdx.x53{blockDim.x}, rS60{( ceilDiv(( ceilDiv(( ceilDiv(i1, blockDim.y) ), 2) ), 1) )}rf, iUS55{1}, iS51{4}, ithreadIdx.y57{blockDim.y}rf, rUS61{1}rf, rS59{2}rf ] ca_pos( 2 ) produce_pos( 8 ), op = add, initial value = float(0), allreduce = false )
```

naoyam: I think I got it. The comment seems wrong, though.

```
since they are register arrays defined in runtime function
```

This `producer_tv` is not defined in the runtime functions, right?

liqiangxl: That's right. `producer_tv` is not defined in the runtime functions, it is just passed to the runtime function.

liqiangxl: comment is revised as 
```
      // ParallelType::Group is used for both reduction and normalization.
      // In grouped outer reduction, the runtime function uses vectorized data
      // transfers between registers and shared memory. The producer tensor is
      // stored in registers and loaded into shared memory in a vectorized
      // manner, so we add it to the vectorizedAccesses map to ensure register
      // alignment.
```

naoyam (approved):
LGTM. Thanks for the fix.

liqiangxl:
!build

Issue #2902, opened 2024-09-04
Extend the inlining analysis to allow inlining with non-conventional loop domains


naoyam:
The inlining analysis such as `MaxPosCalculator` was designed with an implicit assumption that loop domains are derived from logical domains, which is not always the case anymore. For example, finding non-inlinable domains would break as it assumes the logical-to-loop dependency (https://github.com/NVIDIA/Fuser/blob/main/csrc/inlining.cpp#L83).

Design doc: https://docs.google.com/document/d/1lws9xtAz9epfTFZxDCvZlsofGUCS2ECbpsgpyvMPGCQ/edit

PR #2937, opened 2024-09-13
Allow loop domains to have extra IDs

naoyam:
Relaxes the validation of `setLoopDomain`. All IDs of the logical domain, except for broadcast, must be reachable from the loop domain, but the opposite doesn't need to be true. 

The main use case is to set a concrete ID for a broadcast logical ID when inlined into tensors with concrete IDs only.

```
// [i0, i1]
  fusion.addInput(tv0);
  // [i0]
  auto tv1 = makeSymbolicTensor(1);
  fusion.addInput(tv1);

  // [i0]
  auto tv2 = set(tv1);
  // [i0, b1]
  auto tv3 = broadcast(tv2, {false, true});
  // [i0, i1]
  auto tv4 = add(tv0, tv3);
  fusion.addOutput(tv4);
```

In this fusion, when inlining tv2 and tv3 to tv4, it effectively have tv2 and tv3 get the same loop domain as `tv4`. That can be explicitly done by something like this (almost):

```
 // Set the loop domain of tv2 the same as tv4. The new loop domain
  // includes an ID that is not reachable from tv2 logical domain
  tv2->setLoopDomain(
      {tv2->getLogicalDomain().at(0),
       tv4->getLogicalDomain().at(1)->cloneWithoutRFactor()});

  // Same for tv3
  tv3->setLoopDomain(
      {tv3->getLogicalDomain().at(0),
       tv4->getLogicalDomain().at(1)->cloneWithoutRFactor()});
```

This is not complete yet as the clone of tv4 logical ID is not mapped with the tv4 logical ID itself, but assuming that's somehow done, inlining would be trivial as all tensors use the mapped loop domains.

Note that modifying the loop domain like this way is still considered experimental, and in a long term we would probably need a validation that is a little more restrictive.

Not strictly required but related to #2851 and #2902.


naoyam:
!build

naoyam:
!build

csrc/ir/utils.cpp
```diff
@@ -892,39 +892,70 @@ void validateDomainEquivalence(
       std::any_of(frontier.begin(), frontier.end(), is_symb);
   bool dom1_has_symbolic =
       std::any_of(dom1_set.begin(), dom1_set.end(), is_symb);
+
+  CompareDomainResult result;
+
+  auto check_ids = [](const auto& ids_to_check,
```
zasdfgbnm, line 898: Could you add a comment describing what this function does?

csrc/ir/utils.cpp
```diff
@@ -892,39 +892,70 @@ void validateDomainEquivalence(
...
+        continue;
+      }
+      if (!target_set.count(id)) {
+        // not found in target, which means either:
+        //
+        // 1. id is unreachable from target_set, or
+        // 2. id is reachable from target_set but was erased from
+        // target_set as it was used as an input in the traversal.
+        //
+        // The second case means id is redudant
```
zasdfgbnm, line 914: Is this added primarily to check that the loop domain could have IDs not covered by the logical domain, but these additional IDs must be disconnected, instead of being a duplicated ID.

zasdfgbnm (approved):

csrc/ir/utils.cpp
naoyam: Right. I was wondering if that check would need to be done by this function itself since it seems like it's more about the validity of a loop domain, which should not include any duplicated IDs, but detection of duplication is also done by the original function, I decided to keep it detectable.

csrc/ir/utils.cpp
naoyam: Added

PR #2944, opened 2024-09-16
Initialize size_ and local_size_ to 1.

wujingyue:
This partially fixes the bug that Sam found in
https://github.com/NVIDIA/Fuser/pull/2804/files#r1761633706. One is also a more logical initial value when the communicator is unavailable -- the test is being run on single node and single GPU.


wujingyue:
!build

samnordmann (approved):

Issue #2899, opened 2024-09-04
Distributed transformer tests require number of devices evenly divide E.


wujingyue:
The [actual code](https://github.com/NVIDIA/Fuser/blob/892b7ac5646b3b873f13f2b2ea1db15fa9b9effb/tests/cpp/test_multidevice_transformer.cpp#L46-L49) however checks `H % D`. I think it makes sense to move this check to individual test cases. For example, MLP_Layer tests doesn't involve `H` at all and can check `4*E % D` instead.

PR #2945, opened 2024-09-16
Update multidevice transformer test skip conditions

cowanmeg:
Fix https://github.com/NVIDIA/Fuser/issues/2899

### DIFF
--- a/tests/cpp/test_multidevice_transformer.cpp
+++ b/tests/cpp/test_multidevice_transformer.cpp
@@ -35,9 +35,9 @@ class DistributedTransformerTest
 
   void SetUp() {
     MultiDeviceTest::SetUp();
-    if (H % D != 0) {
+    if ((4 * E) % D != 0) {
       GTEST_SKIP()
-          << "Distributed transformer tests require number of devices evenly divide E ";
+          << "Distributed transformer tests require number of devices evenly divide 4*E";
     }
     if (!deviceMajorMinorCheck(8)) {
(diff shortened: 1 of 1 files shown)


tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -35,9 +35,9 @@ class DistributedTransformerTest
 
   void SetUp() {
     MultiDeviceTest::SetUp();
-    if (H % D != 0) {
+    if ((4 * E) % D != 0) {
```
wujingyue, line 38: Is this required for all the tests here or just the MLP tests? I think MHA is fine without this requirement.

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -673,6 +673,9 @@ TEST_P(DistributedTransformerTest, MLP_Layer) {
 }
 
 TEST_P(DistributedTransformerTest, MultiheadAttention) {
+  if (H % D != 0) {
+    GTEST_SKIP() << "Requires number of devices evenly divide H";
```
wujingyue, line 677: Can you print out the value of H and of D to make the message more useful?

tests/cpp/test_multidevice_transformer.cpp
cowanmeg: Hmm just MLP, I'll move it there

wujingyue (approved):

PR #2948, opened 2024-09-17
Avoid using FusionProfiler if compute-sanitizer is used

naoyam:
Both use CUPTI, multiple of which cannot coexist

### DIFF
--- a/tests/cpp/test_fusion_profiler.cpp
+++ b/tests/cpp/test_fusion_profiler.cpp
@@ -16,6 +16,7 @@
 #include <inlining.h>
 #include <kernel_cache.h>
 #include <ops/all_ops.h>
+#include <sys_utils.h>
 #include <tests/cpp/utils.h>
 #include <tests/cpp/validator.h>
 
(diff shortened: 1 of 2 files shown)


naoyam:
!build

tests/cpp/test_fusion_profiler.cpp
```diff
@@ -25,6 +26,11 @@ class FusionProfilerTest : public NVFuserTest {
  protected:
   void SetUp() override {
     NVFuserTest::SetUp();
+
+    if (detectComputeSanitizer()) {
```
naoyam, line 30: Completely disables the FusionProfiler tests when compute-sanitizer is used. This should be fine as we should also run the tests without compute-sanitizer.

tests/cpp/test_matmul.cpp
```diff
@@ -442,9 +445,13 @@ TEST_P(MatmulTestWithLayout, AmpereSwizzle) {
     NVF_CHECK(gdimx == expected_gdimx);
     NVF_CHECK(gdimy == expected_gdimy);
 
-    FusionProfiler::stop();
-    runtime = FusionProfiler::profile().kernel_time_ms;
-    ProfilerOptionsGuard::getCurOptions().unset(ProfilerOption::Enable);
+    if (!detectComputeSanitizer()) {
```
naoyam, line 448: Here, FusionProfiler is used for timing measurements, which doesn't mean anything with compute-sanitizer

kevinstephano (approved):
LGTM

PR #2950, opened 2024-09-17
ring-based reduce-scatter pipelining, ATen implementation

samnordmann:
ATen implementation of a GEMM+Reduce-scatter operation, with a decomposition of the Reduce-scatter into a ring algorithm. This reproduces a technique present in TransformerEngine for achieving comm/compute overlap in Megatron.

# Experiment
The following nSight profile clearly shows that we achieve overlap:
Screenshot 2024-09-18 at 12 24 07
setup: DGX 8*V100 32GB
params: backend=NCCL with coalescence, M=K=N=2048, S=8, number_of_streams=3

# Arbitrary number of steps

The algorithm we provide is slightly more general than the classical case, as it allows decomposing reduce-scatter into an arbitrary large number of steps, not only `num_devices_` steps as in the classical algorithm. More precisely, the parameter `S` (which stands for the number of steps, meaning, the number of interleaved comms and compute), which is classically equal to `num_devices_` for the ring algorithm, is only assumed to be a **multiple of `num_devices_`** in our version. This is an important parameter as it gives more flexibility for the size of interleaved chunks and therefore could lead to better overlap -- but thorough perf analysis remains to be done. If `S>num_devices_`, only a fraction `S/num_devices_` of the buffer is computed and communicated to the peers at each iteration.

# Coalescing
## NCCL
ProcessGroupNCCL provides two methods `startCoalescing` and `endCoalescing`, which internally correspond to `ncclGroupStart` and `ncclGroupEnd`, see [doc here](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/groups.html). Those calls group p2p calls that need to be progressed together -- one global work handle returned by `endCoalescing` needs to be progressed. This has the following main advantages
- calls are progressed concurrently
- since NICs are two-sided, a send and a recv calls need to be coalesced to achieve full BW
- If not coalesced, we can easily reach a deadlock if the send/recv pairs are not ordered correctly. E.g., 
rank0:
```
send to rank 1
recv from rank 1 
```
rank1:
```
send to rank 0
recv from rank 0 
```
This situation created a deadlock because no rank can receive before it has sent.

It is in general preferable to coalesce send/recv calls. The only drawback is that we don't have a fine-grain control on synchronicity, in other words, we can only synchronize with the bulked communication as a whole.

### Remark:
note that NCCL doesn't support tag in send/recv.

## UCC

ProcessGroupUCC does not implement coalesced groups for now. It should not be a problem to achieve full bidirectional BW with two send/recv though. However, having more than two ops in a batch will be suboptimal. Adding UCC coalescing was discussed but not added to POR for now because lacking a good use case.

# Reducing memory foot print

Further optimizations, that we leave as TODO, are possible to reduce the footprint of the work buffers `src_buffer_` and `dst_buffer_`:
- One slice can be shared between src and dst, corresponding to the "local" send/recv at the last iteration
- Probably possible to reduce the outermost dimension with the number of streams.


samnordmann:
!build

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -332,7 +374,104 @@ TEST_F(OverlapTest, ReduceScatterBasedPipeliningHostIrImplementation) {
...
+
+class RingBasedOverlapTest : public OverlapTest {
+ protected:
+  at::Tensor src_buffer_, dst_buffer_;
+  void SetUp() override {
+    OverlapTest::SetUp();
+
+    std::vector<int64_t> buffer_sizes = {
+        params.S, params.M / params.S, params.N};
+    src_buffer_ = at::empty(buffer_sizes, gpu_options_);
```
samnordmann, line 387: further optimizations are possible here to reduce the footprint of the src and dst buffers:
- One slice can be shared between src and dst, corresponding to the "local" send/recv at the last iteration
- Probably possible to reduce the outermost dimension with the number of streams.

Imo it would be reasonable to keep those optimizations for later when we have proper perf analysis

samnordmann:
!build

samnordmann:
I am seeing issues with UCC, both locally and in [the CI](https://gitlab-master.nvidia.com/dl/pytorch/fuser-gh-mirror/-/jobs/111751169). I need to investigate. In the meantime, I disable UCC from being tested

samnordmann:
!build

cowanmeg:
Not needed for the PR, but can we run this with larger sizes and compare it against matmul+reduce scatter baseline? I'm curious if we can see any time savings

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -18,6 +18,28 @@
 
 namespace nvfuser {
 
+namespace {
+
+std::vector<c10::cuda::CUDAStream> CreateStreams(
```
wujingyue, line 23:
```suggestion
std::vector<c10::cuda::CUDAStream> createStreams(
```
for consistency with the rest of the code base.

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -18,6 +18,28 @@
...
+  std::vector<c10::cuda::CUDAStream> streams;
+  std::generate_n(
+      std::back_inserter(streams), number_of_streams, [my_device_index]() {
+        return c10::cuda::getStreamFromPool(
+            /*isHighPriority=*/false, my_device_index);
+      });
+  return streams;
+}
+
+void SynchronizeStreams(const std::vector<c10::cuda::CUDAStream>& streams) {
```
wujingyue, line 35: ditto

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -332,7 +374,151 @@ TEST_F(OverlapTest, ReduceScatterBasedPipeliningHostIrImplementation) {
 
     hie.runWithInput(std::move(inputs));
   }
-  validate();
 }
 
+using RingBasedOverlapTestParams = std::tuple<
+    bool, // use_coalescence
```
wujingyue, line 380: Thanks for the nice explanation in the PR description. It would be really great if the comment stays with the code. It's much harder for readers to `git blame`  a PR that made a particular change.

wujingyue, line 380: For this particular ring-based implementation, isn't there a strong preference to use coalescing? If so, I'd rather remove the code for not coalescing and simplify.

wujingyue (commented):
Thanks for the PR! I'm sending out some initial comments before reviewing the rest.

tests/cpp/test_multidevice_overlap.cpp
samnordmann: Sure, let me add a comment

samnordmann: Thinking about it, I think it would make more sense to add a detailed comment where we add the HostIr for coalesced group, in a PR that I will open soon. Does that sound good to you ?

samnordmann: > For this particular ring-based implementation, isn't there a strong preference to use coalescing? If so, I'd rather remove the code for not coalescing and simplify.

In theory, yes, it is correct. However, ProcessGroupUCC doesn't implement coalescing, so we need an alternative for this backend, hence the branching. UCC has great advantages for P2P, e.g., regarding SM occupation, that is why it is important to support this backend eventually.

samnordmann: would you like me to remove it in the meantime, and add it separately when we add support to UCC ?

wujingyue: > would you like me to remove it in the meantime, and add it separately when we add UCC support ?

Yes, that sounds the best to me.

wujingyue, line 380:
```suggestion
    // Whether to coalesce P2P communication. ProcessGroupNCCL provides two methods startCoalescing and endCoalescing, which internally correspond to ncclGroupStart and ncclGroupEnd, see [doc here](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/groups.html). Those calls group p2p calls that need to be progressed together -- one global work handle returned by endCoalescing needs to be progressed. It is in general preferable to coalesce send/recv calls. The only drawback is that we don't have a fine-grain control on synchronicity, in other words, we can only synchronize with the bulked communication as a whole. ProcessGroupUCC does not implement coalesced groups for now. It should not be a problem to achieve full bidirectional BW with two send/recv though. However, having more than two ops in a batch will be suboptimal. Adding UCC coalescing was discussed but not added to POR for now because lacking a good use case.
    bool,
```

wujingyue: I'm unsure how relevant this is, because you may want to remove this knob for now. Anyhow, I copied some of what you said in the description in another comment. If you decide to keep the knob, I'd keep the comment with the code.

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -102,28 +121,21 @@ class OverlapTest : public MultiDeviceTest {
...
     tb_unsharded_ = at::empty(tb_unsharded_sizes, cpu_options);
-    ta_ = at::empty(ta_sizes, gpu_options);
-    tb_ = at::empty(tb_sizes, gpu_options);
-
-    // We pre-allocate the output and some intermediate buffers so we do not
-    // rely on torch allocator, which do not behave well with multi-stream
-    // programming.
-    tc_locally_reduced_ = at::empty(tc_locally_reduced_sizes, gpu_options);
-    tc_ = at::empty(tc_sizes, gpu_options);
+    ta_ = at::empty(ta_sizes, gpu_options_);
```
wujingyue, line 131: Note to myself: ta_ is column-wise sharded and row-wise chunked.

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -332,7 +374,151 @@ TEST_F(OverlapTest, ReduceScatterBasedPipeliningHostIrImplementation) {
...
+      {number_of_steps_per_ring,
+       number_of_rings,
+       params.M / params.S,
+       params.K / num_devices_});
+  auto src_buffer_reshaped = at::reshape(
+      src_buffer_,
+      {number_of_steps_per_ring,
+       number_of_rings,
+       params.M / params.S,
+       params.N});
```
wujingyue, line 438: This is as large as the whole output matrix? Probably OK for the first version. However, I believe there's a way to reuse chunks by limiting the number of streams and putting extra synchronization, e.g., https://docs.google.com/document/d/1BK_xEolbJDEky5trLW8jbuNHMX-v2QSY40O8782lYQ8/edit#heading=h.g5s0igtg8my1

samnordmann: Yes, this is the note I wrote at the end of the PR description under the header "Reducing memory foot print"

tests/cpp/test_multidevice_overlap.cpp
samnordmann: done

samnordmann: I removed the knob

samnordmann:
> I am seeing issues with UCC, both locally and in [the CI](https://gitlab-master.nvidia.com/dl/pytorch/fuser-gh-mirror/-/jobs/111751169). I need to investigate. In the meantime, I disable UCC from being tested

Issue has been filled and assigned to UCX team: https://redmine.mellanox.com/issues/4088373

samnordmann:
!build

wujingyue (approved):

PR #2953, opened 2024-09-17
Fusion cache error cache

jjsjann123:
adding an entry in TrieNode to store exception message.

This allows us to cache exception thrown during `FusionDefinition::finalizeDefinition()`, so later attempts of construction with the same FusionDefinition wouldn't silently skip fusion IR construction and crash with cached incomplete fusion definition.

e.g. in the script below, assuming the fusionDefinition fails inside `finalizeDefinition()`. The first iteration would throw in the try block, but the second iteration would execute fine without any assertion, since it goes through this [short-cut](https://github.com/NVIDIA/Fuser/blob/adbbc75e58e6c53c606e90c8bc64f020b4b9df85/csrc/python_frontend/fusion_definition.cpp#L101-L107).
```
for i in range(2):
    try:
            fusion_func(fd)
    except:
        pass
```

tl;dr: After this PR, the above script would throw the same exception in every iteration.


jjsjann123:
!build

csrc/python_frontend/fusion_definition.cpp
```diff
@@ -87,13 +87,19 @@ void FusionDefinition::finalizeDefinition() {
...
-      print(debug());
-    }
+      if (isDebugDumpEnabled(DebugDumpOption::PythonDefinition)) {
+        print(debug());
+      }
 
-    buildFusionIr(preschedFusion());
+      buildFusionIr(preschedFusion());
+    } catch (const std::exception& e) {
+      trie_node_->setException(e.what());
```
jjsjann123, line 99: issue happening after `fusionCache()->createChild` wouldn't be visible by fusion cache, if the exception is suppressed by python side.

Hence we explicitly set the exception message on the terminal node, so we'll be able to throw the same exception again when user tries to create the same fusion entry.

rdspring1 (commented):
I see how this is useful.

Why can't the program crash when we fail to make a valid FusionDefinition?

Do we need to guard every call to `buildFusionIr`?

1. [FusionCache::serialization](https://github.com/NVIDIA/Fuser/blob/main/csrc/python_frontend/fusion_cache.cpp#L744)
2. [FusionDefinition::finalizeDefinition](https://github.com/NVIDIA/Fuser/blob/main/csrc/python_frontend/fusion_cache.cpp#L744)
3. [FusionDefinition::setupSchedule](https://github.com/NVIDIA/Fuser/blob/main/csrc/python_frontend/fusion_definition.cpp#L210)

csrc/python_frontend/fusion_definition.cpp
jacobhinkle: Can you put something to that effect in a comment at this location?

jacobhinkle (commented):
LGTM. Will let others stamp.

jjsjann123:
> Why can't the program crash when we fail to make a valid FusionDefinition?
> 
> Do we need to guard every call to `buildFusionIr`?

No crashing is fine. The only reason that I'm patching this instance is because we use it in our python test.
https://github.com/NVIDIA/Fuser/blob/21ab9ab22f19f328f8ec673728007f5d1dd00bff/tests/python/test_ops.py#L275-L278

jjsjann123:
Oops... looks like I indeed breaks serde test.... Let me take another look at that.

jjsjann123:
!build

jjsjann123:
Tagging @rdspring1  CI looks clean now. (please ignore the distributed tests) that's patched in https://github.com/NVIDIA/Fuser/pull/2959

rdspring1 (approved):

PR #2936, opened 2024-09-12
[WIP] IdModel-based inlining

naoyam:

PR #2961, opened 2024-09-18
Minor cleanup & fix

naoyam:
Encountered while working on #2936

### DIFF
--- a/csrc/ir/internal_base_nodes.h
+++ b/csrc/ir/internal_base_nodes.h
@@ -529,10 +529,19 @@ class TensorDomain : public Val {
     return root_domain_.empty() ? logical_domain_ : root_domain_;
   };
 
+  bool isRoot(const IterDomain* id) const {
+    return hasRoot() &&
+        std::find(root().begin(), root().end(), id) != root().end();
+  }
+
   // The output logical domain.
   const std::vector<IterDomain*>& logical() const {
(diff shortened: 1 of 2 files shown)


csrc/logical_domain_map.cpp
```diff
@@ -693,11 +693,11 @@ bool ComputeAtLogicalDomainMap::canMap(
     const TensorDomain* td_b,
     const IterDomain* id_b) const {
   NVF_ERROR(
-      id_a->definition() == nullptr || id_a->isRFactorProduct(),
```
naoyam, line 696: Previously we assumed a root domain has no definition, which may not be the case anymore.

csrc/ir/internal_base_nodes.h
```diff
@@ -529,22 +529,41 @@ class TensorDomain : public Val {
     return root_domain_.empty() ? logical_domain_ : root_domain_;
   };
 
+  bool isRoot(const IterDomain* id) const {
```
jacobhinkle, line 532: We might want to include a comment here indicating that this is always false if there is no root domain, i.e. this is not the same as checking `getMaybeRootDomain()` for `id`. Same goes for loop and allocation.

jacobhinkle (approved):
LGTM

naoyam:
!build

PR #2963, opened 2024-09-18
Append to AbstractTensor's info whenever we append to domain

jacobhinkle:
This fixes failures in `TMAIndexingTest` tests on Hopper that result in a segfault due to unchecked accesses that result when `domain` has entries but `info` is empty in `AbstractTensorWithInfo`. Note that another approach would be to switch to a class instead of a struct and add accessors that guarantee we maintain the same length for each of these vectors.

This change also adds more assertions in scheduling methods of AbstractTensorWithInfo to check that the `domain` and `info` attributes maintain the same size.

### DIFF
--- a/csrc/device_lower/analysis/tma.cpp
+++ b/csrc/device_lower/analysis/tma.cpp
@@ -816,6 +816,7 @@ class DomainMerger {
     for (auto& item : raw_tma_domain) {
       domain_.domain.emplace_back(
           ValGroupAndItsGraph{std::move(std::get<0>(item)), &id_graph});
+      domain_.info.emplace_back();
       contiguity_and_stride_.emplace_back(std::get<1>(item), std::get<2>(item));
     }
   }
--- a/csrc/abstract_tensor.h
+++ b/csrc/abstract_tensor.h
@@ -634,6 +634,8 @@ struct AbstractTensorWithInfo {
       int64_t axis,
       Val* factor,
       bool inner_split = true) {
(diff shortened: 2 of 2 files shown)


jacobhinkle:
!build

jjsjann123 (approved):
stamp for clean CI.

jacobhinkle:
> stamp for clean CI.

:bowing_man:

csrc/device_lower/analysis/tma.cpp
```diff
@@ -816,6 +816,7 @@ class DomainMerger {
     for (auto& item : raw_tma_domain) {
       domain_.domain.emplace_back(
           ValGroupAndItsGraph{std::move(std::get<0>(item)), &id_graph});
+      domain_.info.emplace_back();
```
naoyam, line 819: This is quite easy to miss. I think we should disallow directly changing `domain` and `info` since they need to be consistent.

jacobhinkle: I agree. Shall I merge this to clear CI then put the change from struct to class in another PR?

naoyam: Let's do that in another PR.

jacobhinkle: See #2967

jjsjann123:
Note for myself. We aren't sure why we don't see the exact segfault error message in the log, though there is a hint on a possible segfault.

I'll ask Xiao when he gets back from vacation.

PR #2964, opened 2024-09-19
Extend MaxPosCalculator::isAllowedID to support loop domains that are not derived from logical domains

naoyam:
Stacked on top of #2957.
Part of #2902 

For this, the only required change is just using `BFS` instead of `IterVisitor`


csrc/inlining.h
```diff
@@ -29,6 +29,7 @@ class MaxPosCalculator {
   // map to all its consumer TVs.
   void buildUnmappableDims(bool compute_at_only);
 
+ public:
```
naoyam, line 32: Just to make it testable

tests/cpp/test_gpu3.cpp
```diff
@@ -3603,166 +3603,6 @@ TEST_F(
   testValidate(&fusion, cg_outputs, aten_inputs, __LINE__, __FILE__);
 }
 
-TEST_F(NVFuserTest, FusionInliningMismatchedDims1_CUDA) {
```
naoyam, line 3606: Moved to test_inlining.cpp

tests/cpp/test_inlining.cpp
```diff
@@ -0,0 +1,325 @@
...
+
+  EXPECT_EQ(
+      TransformReplay::getMatchedLeafPosWithoutReplayPasC(tv0, tv1, 3), 3);
+  EXPECT_EQ(
+      TransformReplay::getMatchedLeafPosWithoutReplayCasP(tv1, tv0, 3), 3);
+  EXPECT_EQ(
+      TransformReplay::getMatchedLeafPosWithoutReplayPasC(tv1, tv2, 3), 3);
+  EXPECT_EQ(
+      TransformReplay::getMatchedLeafPosWithoutReplayCasP(tv2, tv1, 3), 3);
+}
```
naoyam, line 184: Moved from test_gpu3.cpp with no change except for using `EXPECT_` instead of `NVF_ERROR`.

tests/cpp/test_inlining.cpp
```diff
@@ -0,0 +1,325 @@
...
+    // Persistent domain
+    EXPECT_FALSE(isAllowedID(tv1, 2)) << tv1->getLoopDomain().at(2)->toString();
+  }
+
+  // tv2
+  {
+    std::vector<IterDomain*> loop_domain{
+        tv2->getLogicalDomain().at(0),
+        tv1->getLoopDomain().at(1)->cloneWithoutRFactor(),
+        tv1->getLoopDomain().at(2)->cloneWithoutRFactor()};
```
zasdfgbnm, line 250: Do we need to change the `IterType` to `Reduction`?

tests/cpp/test_inlining.cpp
```diff
@@ -0,0 +1,325 @@
...
+        tv1->getLoopDomain().at(0)->cloneWithoutRFactor(),
+        tv1->getLoopDomain().at(1)->cloneWithoutRFactor(),
+        tv1->getLoopDomain().at(2)->cloneWithoutRFactor()};
+    IrBuilder::create<Merge>(
+        tv3->getLogicalDomain().at(0), loop_domain[1], loop_domain[2]);
+    tv3->setLoopDomain(loop_domain);
+    for (const auto i : c10::irange(3)) {
+      EXPECT_TRUE(isAllowedID(tv3, i))
+          << tv3->getLoopDomain().at(i)->toString();
+    }
```
zasdfgbnm, line 274: If I have
```
producer: [I0, I1]
```
and
```
consumer: [I0*I1]
```
Then can we inline `producer` at 1?

tests/cpp/test_inlining.cpp
naoyam: That's a good point. Yes, I think so. We should actually probably do some check when doing `setLoopDomain`.

tests/cpp/test_inlining.cpp
naoyam: No, the producer also needs the merge of `I0` and `I1`.

Note that `isAllowedID` doesn't consider producer-consumer relationships. It's only about just the given ID itself, except for the persistence constraint by `ComputeAtLogicalDomainMap`.

tests/cpp/test_inlining.cpp
naoyam: Changed. Also #2974

zasdfgbnm (approved):

PR #2965, opened 2024-09-19
only do grid split when needed

liqiangxl:
**Issue:** In inner outer persistent scheduler, the last step is doing an outer reduction, the inner dim is parallelized by `vectorization`, `bimdx`, and `gdimy`. Current main branch always do three splits using  `vectorization`, `bdimx`, and `gdimy`, however, the last split is not needed if `vectorization * bdimx * gdimy >= inner dim`, for example:
```
T0 
logical domain : (iS264{gridDim.y}, iS265{i1})
 contiguity: t t
  Split: iS265{i1} by factor 4
  Split: iS997{( ceilDiv(i1, 4) )} by factor blockDim.x 
  Split: iS999{( ceilDiv(( ceilDiv(i1, 4) ), blockDim.x) )} by factor gridDim.y
```
The last split is redundant if `4 * blockDim.x * gridDim.y >= i1`
**Fix:**
Only split when `vectorization * bdimx * gdimy < inner dim`
**Influence:**
Removing this extra split saves one loop in the generated code.
Performance is increased in some cases but decreased in other cases, all changes are within 10%. see [dashboard](http://nv/ekP).


liqiangxl:
!build --diff --pybench

liqiangxl:
Hi @jjsjann123 , can you help taking a look of [this](https://gitlab-master.nvidia.com/dl/pytorch/fuser-gh-mirror/-/jobs/111975020) CI fail? It didn't find any test error but exit with 
```
Cleaning up project directory and file based variables
00:00
ERROR: Job failed: exit status 1
```

jjsjann123:
> Hi @jjsjann123 , can you help taking a look of [this](https://gitlab-master.nvidia.com/dl/pytorch/fuser-gh-mirror/-/jobs/111975020) CI fail? It didn't find any test error but exit with
> 
> ```
> Cleaning up project directory and file based variables
> 00:00
> ERROR: Job failed: exit status 1
> ```

Since it mentioned about segfault and it's hopper matmul tests... I'm guessing it's fixed by this PR: https://github.com/NVIDIA/Fuser/pull/2963

jjsjann123:
!build --diff --pybench

liqiangxl:
!build

csrc/scheduler/reduction_heuristic.h
```diff
@@ -125,6 +125,7 @@ class ReductionParams : public HeuristicParams {
   bool tidx_for_outer_reduction = false;
   // pad outer reduction to warp
   bool pad_outer_reduction_to_warp = false;
+  bool combined_split_grid_inner_dim = false;
```
jjsjann123, line 128: nitpick: let's put a comment here on what this field does and why.

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -872,8 +872,14 @@ std::shared_ptr<ReductionParams> innerOuterPersistentHeuristic(
...
     }
     rparams->block_dim_iter_dom = ParallelType::TIDy;
+    rparams->combined_split_grid_inner_dim =
+        iop.vectorization_factor_outer * iop.bdimy * iop.gdimy <
+        inner_dim_numel;
   } else {
     rparams->block_dim_inner_reduction_extra = ParallelType::TIDy;
+    rparams->combined_split_grid_inner_dim =
+        iop.vectorization_factor_outer * iop.bdimx * iop.gdimy <
+        inner_dim_numel;
```
jjsjann123, line 882: nitpic: IIUC, the above block is identical for the if-else-blocks. maybe just move it out so we don't have redundancy

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -1128,8 +1135,10 @@ void scheduleReductionCombinedOuter(
         outer_reduction_tv->axis(axisID--)->parallelize(ParallelType::TIDx);
       }
 
-      outer_reduction_tv->split(
-          axisID, NamedScalar::getParallelDim(ParallelType::BIDy));
+      if (rparams.combined_split_grid_inner_dim) {
+        outer_reduction_tv->split(
+            axisID, NamedScalar::getParallelDim(ParallelType::BIDy));
+      }
```
jjsjann123, line 1141: ditto

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -1103,9 +1109,10 @@ void scheduleReductionCombinedOuter(
             axisID, NamedScalar::getParallelDim(ParallelType::TIDx));
         outer_reduction_tv->axis(axisID--)->parallelize(ParallelType::TIDx);
       }
-
-      outer_reduction_tv->split(
-          axisID, NamedScalar::getParallelDim(ParallelType::BIDy));
+      if (rparams.combined_split_grid_inner_dim) {
```
jjsjann123, line 1112: naive question. Even though we are skipping the split, I thought we still would need to specify the current IterDomain with `ParallelType::BIDy`?

csrc/scheduler/normalization_inner_outer.cpp
liqiangxl: They are slightly different:
```
if(){
vect * bdimy * gdimy < inner_dim
}else{
vect * bdimx * gdimy < innder_dim
}
```

csrc/scheduler/normalization_inner_outer.cpp
liqiangxl: Yes, we always specify it as `ParallelType::BIDy` using `outer_reduction_tv->axis(axisID--)->parallelize(ParallelType::BIDy);`

liqiangxl: `outer_reduction_tv->axis(axisID--)->parallelize(ParallelType::BIDy);` is out of the `if` statement.

csrc/scheduler/normalization_inner_outer.cpp
liqiangxl: same as previous

csrc/scheduler/reduction_heuristic.h
liqiangxl: added

liqiangxl:
!build --diff

liqiangxl:
!build --diff

csrc/scheduler/normalization_inner_outer.cpp
jjsjann123: ah, sorry about that. 🙇

csrc/scheduler/normalization_inner_outer.cpp
jjsjann123: But this one is the same as line 837-841.

(Note, I'm only nitpicking, we don't have to change it. I wanted to point it out in case there's some logic issue here.)

jjsjann123 (commented):
LGTM, a minor question/comment.

jjsjann123:
> Removing this extra split saves one loop in the generated code.
Performance is increased in some cases but decreased in other cases, all changes are within 10%. 

what's with the perf regression here? Are those just small kernels with fluctuation?
If we are just removing one trivial loop, what's the reason for a potential negative perf impact?

csrc/scheduler/normalization_inner_outer.cpp
liqiangxl: Thanks for pointing it out. Yes, they are same and the logic is fine. I repeated them in the if-else branches simplily becuase I want to keep both the code block in if and else branches as complete schedule proess.
**current approach**
```
if(multiple reductions per block){
    schedule approach-1
} else {
    schedule approach-2
}
```
There are some common schedules (e.g. using BIDy) in approach-1 and approach-2. However, it is not prefered to split it out and using the following code:
**Other options:**
```
if(multiple reductions per block){
    part of schedule approach-1
} else {
    part of schedule approach-2
}
common schedule processes shared by approach-1 & approach-2
```

liqiangxl:
> > Removing this extra split saves one loop in the generated code.
> > Performance is increased in some cases but decreased in other cases, all changes are within 10%.
> 
> what's with the perf regression here? Are those just small kernels with fluctuation? If we are just removing one trivial loop, what's the reason for a potential negative perf impact?

Some are from large cases and repeatable, e.g. 16384 x 23040 reduced from 47% SOL to 42% SOL. Not sure why, we do saved one loop in both CUDA & PTX codes.
image
image
PTX info is also same
```
ptxas info    : Used 255 registers, used 1 barriers, 16 bytes smem
```
We can leave this PR open and recheck after warp reduction & heuristics.

jjsjann123 (approved):
hmmm. that perf regression is indeed strange. I think it's interesting to look at it but I'm not sure about what priority we should give this. (Yet another weird compiler behavior?)

Since you did verify the behavior in generated kernel, I'm stamping it to unblock you.

liqiangxl:
!build

liqiangxl:
!build

PR #2966, opened 2024-09-19
warp reduction in x and y dims

liqiangxl:
**Issue**
The first part of innerOuter persistent kernel is an inner reduction, the inner dim is parallelized by `vectorization`, `bdimx`, `bdimy`, and `persistent`. Both `bdimx` and `bdimy` are used becuase we want to re-use `bdimy` to parallel the outer dim and re-use `bdimx` to parallel the inner dim in the second part of the kernel. However, it keeps us away from using warp reduction since the current runtime function only supports reduction in `bimdx`.
**Fix**
(1) Add another warp reduction runtime function to support reduction in x and y dims.
(2) `bdimx` and `bdimy` are explicitly set to static and passed to warp reduction as template paras. 
(3) we are launching a cooperative kernel, `gdimy` is also static
**Influence**
(1) For a case with `vectorization=8`, `bdimx=64`, `bdimy=4`, and `persistent=13` inner reduction dim is scheduled as:
```
rS195{13}, rUS197{1}, rthreadIdx.z198{( ceilDiv(( ceilDiv(( ceilDiv(( ceilDiv(( ceilDiv(i1, 8) ), 64) ), 4) ), 13) ), 1) )}, rthreadIdx.y194{4}, rthreadIdx.x192{64}, rV190{8}
```
The heuristic ensures `bdimz = inner dim size / vectorization / bdimx / bdimy / persistent == 1`, this gives us a static warp reduction across bdimx and bdimy.
**Performance:
A100 layer norm backward:**
image

**A100 [RMS norm backward](http://nv/ekR):**
image

**[H100 layer norm backward](http://nv/ekS)** 
Local run, not included in CI.
image

**Other hardware:**
Local run, not included in CI.
[link](https://docs.google.com/document/d/14GFuGodBatD-g7_Ws--4QEjPZGAAPOPZOhjrETX-gX8/edit#heading=h.2owuywwe5bv4)
**Other options**
(1) we can set `bdimy` to dynamic, then no need to use `bdimz` and can save one split. Then, to use warp reduction, needs to pad `bdimy` and ensure `bdimx * padded_bdimy % 32 == 0`. Aslo needs to revise the outer reduction part of the kernel, it was assuming `bdimy` is static.


liqiangxl:
!build --diff --pybench

liqiangxl:
!build

csrc/device_lower/utils.cpp
```diff
@@ -345,17 +349,36 @@ std::optional<IterDomain*> getMaybeWarpReductionDim(
...
 
-  if (reduction_on_xdim->hasPaddingToMultipleOfWarp()) {
-    return std::optional<IterDomain*>(reduction_on_xdim);
-  }
+  // reduction only in xdim.
+  if (!reduction_on_ydim && !reduction_on_zdim) {
+    std::pair<IterDomain*, IterDomain*> reduction_dims =
+        std::make_pair(reduction_on_xdim, nullptr);
+    if (reduction_on_xdim->hasPaddingToMultipleOfWarp()) {
+      return std::optional<std::pair<IterDomain*, IterDomain*>>(reduction_dims);
```
jjsjann123, line 357: nitpick: `return std::make_pair(reduction_on_xdim, nullptr)`

csrc/device_lower/utils.cpp
```diff
@@ -345,17 +349,36 @@ std::optional<IterDomain*> getMaybeWarpReductionDim(
...
+    }
 
-  if (reduction_on_xdim->extent()->isConstInt()) {
-    auto extent_value = reduction_on_xdim->extent()->evaluate();
-    if (extent_value % at::cuda::warp_size() == 0) {
-      return std::optional<IterDomain*>(reduction_on_xdim);
+    if (reduction_on_xdim->extent()->isConstInt()) {
+      auto extent_value = reduction_on_xdim->extent()->evaluate();
+      if (extent_value % at::cuda::warp_size() == 0) {
+        return std::optional<std::pair<IterDomain*, IterDomain*>>(
```
jjsjann123, line 363: ditto

csrc/device_lower/utils.cpp
```diff
@@ -345,17 +349,36 @@ std::optional<IterDomain*> getMaybeWarpReductionDim(
...
+    // special case used in innerOuter scheduler where bdimx and bdimy are
+    // constants bdimz is always 1.
+    if (reduction_on_xdim->extent()->isConstInt() &&
+        reduction_on_ydim->extent()->isConstInt()) {
+      auto extent_x_value = reduction_on_xdim->extent()->evaluate();
+      auto extent_y_value = reduction_on_ydim->extent()->evaluate();
+      if ((extent_x_value * extent_y_value) % at::cuda::warp_size() == 0) {
+        std::pair<IterDomain*, IterDomain*> reduction_dims =
+            std::make_pair(reduction_on_xdim, reduction_on_ydim);
+        return std::optional<std::pair<IterDomain*, IterDomain*>>(
```
jjsjann123, line 377: ditto

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -611,6 +612,13 @@ std::unique_ptr<ReductionParams> innerOuterPersistentHeuristic(
     rparams->block_dim_iter_dom = ParallelType::TIDy;
   } else {
     rparams->block_dim_inner_reduction_extra = ParallelType::TIDy;
+    rparams->static_bdimx = true;
+    rparams->static_bdimy = true;
```
jjsjann123, line 616: I think here we need to have all block dimensions be static, since `bdimz == 1`?

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -631,13 +639,14 @@ std::unique_ptr<ReductionParams> innerOuterPersistentHeuristic(
...
   rparams->grid_dim_iter_dom = ParallelType::BIDy;
+
   rparams->lparams = LaunchParams(
       LaunchParams::UNINITIALIZED_VAL,
       iop.gdimy,
       LaunchParams::UNINITIALIZED_VAL,
       iop.bdimx,
       iop.bdimy,
-      LaunchParams::UNINITIALIZED_VAL);
+      iop.bdimz);
```
jjsjann123, line 649: Naive question regarding iop.bdimz. Why do we add this in heuristic if it's always going to be 1?

runtime/warp.cu
```diff
@@ -88,4 +88,58 @@ __device__ void warpReduceTIDX(
   }
 }
 
+template <int BDIMX, int BDIMY, bool Aligned, typename T, typename Func>
```
jjsjann123, line 91: Why do we need to have this template function with a slight different implementation for inter-warp reduction?

csrc/scheduler/normalization_inner_outer.cpp
liqiangxl: It is redundant since we already have `NVF_ERROR(iop.bdimz == 1, "bdimz must be 1.");`
Removed this change.

runtime/warp.cu
liqiangxl: The existing version `warpReduceTIDX` is doing reduction in X dim, Y dim is used for iteration. It requires `bdimx % 32 ==0`
This new version `warpReduceTIDXY` is doing reduction in X & Y dims, it requires `(bdimx * bdimy) % 32 ==0`. So they are different. Also `bdimx` and `bdimy` are static values, then we can use template vars.

csrc/device_lower/utils.cpp
liqiangxl: revised to
```
      return std::optional<std::pair<IterDomain*, IterDomain*>>(
          std::make_pair(reduction_on_xdim, nullptr));
```
I think we still want to keep this `std::optional` since the function is `getMaybeWarpReductionDim`, otherwise, we need to remove `Maybe` in the function name and make other changes in its callers.

csrc/device_lower/utils.cpp
liqiangxl: same

csrc/device_lower/utils.cpp
jjsjann123: No I wasn't suggesting to change the signature. It should automatically convert that to an std::optional. Is there any benefit of having that explicit?

runtime/warp.cu
jjsjann123: ah, my bad. I didn't realized that in warpReduceTIDX, we are still using threadIdx.y. Earlier I was suggesting that these two should be merged together, but that doesn't look like worth it.

csrc/scheduler/normalization_inner_outer.cpp
jjsjann123: Ha, I see it now. Thanks for bearing with me here.

jjsjann123:
The code change looks good to me.
One last question regarding the performance, with A100 layer norm backward, what's worst regression we are looking at? The axis doesn't give that.

csrc/device_lower/utils.cpp
liqiangxl: got you. I didn't realize it can auto convert to `std::optional`, changed.

liqiangxl:
> The code change looks good to me. One last question regarding the performance, with A100 layer norm backward, what's worst regression we are looking at? The axis doesn't give that.

double checked, it is 0.86x, [link](http://nv/elq).
image

jjsjann123 (approved):
LGTM

PR #2967, opened 2024-09-19
Convert AbstractTensor from struct to class

jacobhinkle:
`AbstractTensor` now has many methods and as of #2933 it is subclassed into `TaggedAbstractTensor` which has additional methods. Also, there is no longer just a single vector attribute `domain` but also `info`. Instead of requiring the user to manage those vectors and keep them the same length, this PR makes them `protected` members of a class and uses the existing accessors to adjust them, as suggested in https://github.com/NVIDIA/Fuser/pull/2963#discussion_r1765951724.

### DIFF
--- a/csrc/abstract_tensor.h
+++ b/csrc/abstract_tensor.h
@@ -476,31 +476,29 @@ struct DispatchParallelize {
 //   auto vv = v.strip(); // [id0]
 
 template <typename Info>
-struct AbstractTensorWithInfo {
-  std::vector<AbstractId> domain;
-  std::vector<Info> info;
-
+class AbstractTensorWithInfo {
+ public:
   AbstractTensorWithInfo() = default;
 
   // These constructors use default info
   AbstractTensorWithInfo(std::vector<AbstractId> domain)
-      : domain(std::move(domain)), info(domain.size()) {}
+      : domain_(std::move(domain)), info_(domain_.size()) {}
   AbstractTensorWithInfo(const std::vector<IterDomain*>& domain)
-      : domain(domain.begin(), domain.end()), info(domain.size()) {}
+      : domain_(domain.begin(), domain.end()), info_(domain_.size()) {}
   AbstractTensorWithInfo(std::initializer_list<AbstractId> domain)
-      : domain(domain), info(domain.size()) {}
+      : domain_(domain), info_(domain_.size()) {}
 
   // These constructors use provided info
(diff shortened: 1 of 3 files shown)


jacobhinkle:
!build

jacobhinkle:
!build

zasdfgbnm:
Is the build failure related?

jacobhinkle:
> Is the build failure related?

I am not sure yet, but I doubt it. I am also seeing this failure on #2913 which doesn't include this change.

csrc/abstract_tensor.h
```diff
@@ -1000,8 +1005,13 @@ struct TaggedAbstractTensor : AbstractTensorWithInfo<TagSetInfo<Tag>> {
...
   const std::unordered_set<Tag>& getTags(int64_t i) const {
-    i = wrapDim(i, (int64_t)this->info.size());
-    return this->info[i].tags;
+    i = wrapDim(i, (int64_t)this->size());
+    return this->info_[i].tags;
+  }
+
+  std::unordered_set<Tag>& getTags(int64_t i) {
+    i = wrapDim(i, (int64_t)this->size());
+    return this->info_[i].tags;
```
jacobhinkle, line 1014: I added this non-const accessor so that we can manually insert and erase tags for a single axis.

zasdfgbnm:
I am surprised that this PR just work, because I remembered that there are more direct uses of `.domain` and some requires none-trivial change.

jacobhinkle:
> I am surprised that this PR just work, because I remembered that there are more direct uses of `.domain` and some requires none-trivial change.

Yeah I thought it was going to be tougher too but the only places `AbstractTensor` is currently used is in matmul utilities and TMA analysis:
- `MmaSwizzler::scheduleMmaOutputAllocation`
- `swizzleSharedMemory`
- `collapse_tma_domain::DomainMerger`

Only `DomainMerger` was tinkering with `.domain` directly. Otherwise, we do make use of `AbstractTensor` in some tests outside of `test_abstract_tensor.cpp` but those are not directly manipulating `.domain` either.

zasdfgbnm (commented):
LGTM, just want to see CI green for double confirming.

jacobhinkle:
!build

jacobhinkle:
Rebuilding following merging of #2969

zasdfgbnm (approved):

PR #2968, opened 2024-09-19
only set maxrregcount when want to use less than 255 registers

liqiangxl:
No actual perf change, [A100](http://nv/ek_), [H100]"http://nv/ek("
Why want to make this change?
See [doc](https://docs.google.com/document/d/14GFuGodBatD-g7_Ws--4QEjPZGAAPOPZOhjrETX-gX8/edit#heading=h.rxgxv47beek8).


liqiangxl:
!build --pybench

jjsjann123 (approved):
We use to explicitly set max_register to encourage register usage. But looks like it doesn't cause any regression for that use case any more.

Not opposing updating this.

naoyam:
Have you also run the diff check to see actual differences of register counts?

PR #2970, opened 2024-09-19
Host irs: add p2p comms

samnordmann:
# What
Add a new communication IR called `P2PCommunication` representing the one-sided operations of `send` and `recv`.

# Why

## Used in RS+GEMM Host IR implementation
The Reduce-Scatter+GEMM P2P pipelining has been implemented in ATen in 
#2950. The host IR implementation needs this new IR to express the send/recv, see #3060

##  Why not using the pre-existing `Communication` class

A `Communication` IR represents in nvFuser a MPI-style collective (e.g. allgather, reduce-scatter, etc.). For completeness, we also allowed this class to represent a two-sided `SendRecv` operation, grouping together the send and recv calls, even though the semantics of `SendRecv` does not perfectly matches the one of MPI collective. Here is a summary of the main differences
- `send` and `recv` are one-sided communications, contrarily to collectives.
- In contrast with other MPI-collectives, the world communicator (i.e. the `c10d::Backend` initialized with all ranks) can be used to perform the communication. In contrast, an MPI-collective requires instantiating a new `c10d::Backend` for each new `Team`. 
- `SendRecv` result in two different backend call (i.e. `c10d::Backend::send` or `c10d::Backend::recv`) depending on the role sender/receiver. In contrast, all rank participating to a collective (e.g. `allgather`) call the same primitive (`c10d::Backend::allgather`). Therefore, having two different IRs for sending and receiving matches the ATen code better.
- c10d::Backend has different signatures between send/recv on the one hand and other collectives on the other hand: 
    - `send` and `recv` take _one peer rank_ and _one buffer_
    - Collectives take a _team of peer ranks, a src and a dst buffer_.

In addition, for the present use case, we think it is more convenient to add new IRs, `Send` and `Recv`, rather than using the existing `SendRecv` IR. Here are the main reasons:

- **Need for a fully symbolic IR**. The `Communication` IR is, for now, not fully symbolic, since the `root` and `Team` are passed as concrete values and not symbolic `Val`s. This does not fit the need of The Reduce-Scatter+GEMM P2P #2950 since, there, the send/recv peer depends on the index of a symbolic ForLoop.
- Systematically grouping Send and Recv together is not desirable because we want to be able to typically synchronize differently on those two calls


samnordmann:
!build

csrc/multidevice/communication.cpp
```diff
@@ -210,6 +210,53 @@ std::string Communication::toInlineString(int indent_size) const {
...
+
+P2PCommunication::P2PCommunication(
+    IrBuilderPasskey passkey,
+    P2PCommunicationType type,
+    TensorView* buffer,
+    Val* peer,
+    Val* tag)
+    : Expr(passkey) {
+  if (tag == nullptr) {
+    tag = passkey.ir_container_->zeroVal(); // NOLINT
```
wujingyue, line 235: What did lint complain about? I'd specify the type instead of the generic NOLINT.

csrc/multidevice/communication.cpp
```diff
@@ -210,6 +210,53 @@ std::string Communication::toInlineString(int indent_size) const {
...
+  addAttribute(tag);
+}
+
+NVFUSER_DEFINE_CLONE_AND_CREATE(P2PCommunication)
+
+std::string P2PCommunication::toString(const int indent_size) const {
+  std::stringstream ss;
+  indent(ss, indent_size) << "P2PCommunication " << name() << " ("
+                          << "type=" << type() << ", "
+                          << " buffer=" << buffer() << ", "
```
wujingyue, line 250:
```suggestion
                          << "buffer=" << buffer() << ", "
```

tests/cpp/test_multidevice_host_ir.cpp
```diff
@@ -240,6 +240,61 @@ INSTANTIATE_TEST_SUITE_P(
       return s;
     });
 
+class P2PCommHostIrTest : public MultiDeviceTest {};
```
wujingyue, line 243:
```suggestion
using P2PCommHostIrTest = MultiDeviceTest;
```

csrc/multidevice/communication.cpp
```diff
@@ -210,6 +210,53 @@ std::string Communication::toInlineString(int indent_size) const {
...
+  }
+  return os;
+}
+
+P2PCommunication::P2PCommunication(
+    IrBuilderPasskey passkey,
+    P2PCommunicationType type,
+    TensorView* buffer,
+    Val* peer,
+    Val* tag)
```
wujingyue, line 232: What's tag for?

wujingyue (commented):
Is there a reason to keep SendRecv? I didn't like it either because it tries to shoehorn a p2p communication into collective concepts like root and team. 

That being said, there's https://openxla.org/xla/operation_semantics#collectivepermute that implements a "collective" p2p which is widely applicable, e.g., for ring-based communication. Wonder your thoughts on that.

csrc/multidevice/communication.cpp
samnordmann: tag is to differentiate between multiple sends and recv between the same peers. So that if one rank makes two sends to another, the send can be identified by tag-matching.

However, NCCL doesn't support tags. So I would be ok to remove it for now, but still, we might use it in the future.

wujingyue: Thanks! I recommend we remove it for simplicity.

samnordmann: ok, done

csrc/multidevice/communication.cpp
samnordmann: since I removed tags, this is no longer relevant

csrc/host_ir/host_ir.cpp
```diff
@@ -174,12 +175,16 @@ bool SetCurrentStream::sameAs(const Statement* other) const {
   return false;
 }
 
-Wait::Wait(IrBuilderPasskey passkey, Communication* communication)
-    : Expr(passkey, {}, {}, {communication}) {
+Wait::Wait(IrBuilderPasskey passkey, Expr* expr)
+    : Expr(passkey, {}, {}, {expr}) {
   NVF_ERROR(
       passkey.ir_container_->isA<hir::HostIrContainer>(), // NOLINT
```
wujingyue, line 181: I believe this NOLINT will be removed after rebase, but please double check.

csrc/host_ir/host_ir.cpp
```diff
@@ -174,12 +175,16 @@ bool SetCurrentStream::sameAs(const Statement* other) const {
...
       this,
       "must be registered in a HostIrContainer");
+  NVF_ERROR(
+      (expr->isOneOf<Communication, P2PCommunication>()),
```
wujingyue, line 185: Not for this PR, but will you prefer having Communication, CollectiveCommunication (a subclass), and P2PCommunication? This seems to be more structured but adds one level of inheritance hierarchy, potentially making code less trackable.

wujingyue (approved):
Looks great!

csrc/multidevice/communication.h
```diff
@@ -111,6 +111,46 @@ class Communication : public Expr {
   void validate();
 };
 
+enum class P2PCommunicationType { send, recv };
```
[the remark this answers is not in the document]
samnordmann: I was thinking of https://github.com/NVIDIA/Fuser/blob/main/csrc/multidevice/communication.h#L26-L35

Let me change it to `SEND` and `RECV` then, which I find more natural.

csrc/host_ir/host_ir.cpp
samnordmann: it's ok

samnordmann:
> Is there a reason to keep SendRecv? I didn't like it either because it tries to shoehorn a p2p communication into collective concepts like root and team.
> 
> That being said, there's https://openxla.org/xla/operation_semantics#collectivepermute that implements a "collective" p2p which is widely applicable, e.g., for ring-based communication. Wonder your thoughts on that.


I don't like it much easier, for the same reason. I have no strong opinion. SendRecv has a slightly different point-of-view, since it is purely SPMD, while Send and Recv are not. We could imagine to keep SendRecv but to lower to p2p-Send or p2p-recv. Anyway, I have no strong reason to keep SendRecv for now.

About CollectivePermute, yes I saw that a while ago, I don't have particular thoughts on it. It is a very flexible API, but I don't see any particular use case for us at the moment.

samnordmann:
!build

naoyam:
https://github.com/NVIDIA/Fuser/actions/runs/11179177285

PR #2973, opened 2024-09-19
Make HeuristicSummary a lazy cache

csarofeen:
Currently, HeuristicSummary runs all the scheduler analyses to produce the cache it also validates all the information exists once doing so. I simply converted it to be lazy, populating on first hit and reusing if entries exist. This makes it easier to update and use correctly.

I tested it locally on a dynamic shape case and from profiling it seems to be caching as it's supposed to (latency remained consistent before and after).


csarofeen:
!build

csrc/scheduler/compile_time_info.h
```diff
@@ -219,26 +219,16 @@ class CompileTimeInfoBase : public PolymorphicBase {
...
-//!   type to simplify data validation.
-//!  HeuristicSummary has two modes of operation:
-//!  - when in `recording` mode, the information is not available
-//!     in the cache and entries can be written and stored.
-//!  - when not in `recording` mode, compiled-time data has
-//!     been stored in this cache and the entries can be accessed
-//!!    but new entries can no longer be inserted.
+//! Note: Do NOT export this class. MSVC issue with exported class that contains
+//! std::vector<unique_ptr<xxx>>: https://godbolt.org/z/3E4e8T1P1
+//! Compile-time information cache for `canSchedule` and `getHeuristics`
```
naoyam, line 224: nit: Not directly related to the changes in this PR, but I always find this name a little confusing. It's a caching system for some fusion properties that are used to generate heuristics. "HeuristicSummary" seems to indicate it's a summary of heuristics.

naoyam (approved):
LGTM.

csarofeen:
!build

csarofeen:
!build

PR #2977, opened 2024-09-20
Add an experimental option to map cloned IterDomains in the Exact graph

naoyam:
The primary use case is when setting a loop domain of a tensor that would be involved in broadcast forwarding in the conventional inlining approach. For example, think of a fusion like below:

```
t0 = [i0]
t1 = [i0, i1]
t2 = broadcast(t0, {false, true}); // [i0, b2]
t3 = t1 + t2
```

To inline all tensors together, t0 and t2 would be inlined with broadcast forwarding.

Suppose a loop domain of t3 would be set for all tensors, t0, for example, would have a loop domain of: {t0->axis(0), clone of t3->axis(1)}. In this case, however, the clone is almost an exact copy of the t3 ID, there's no use or def to define its exact mapping with the original t3 ID. This is problematic when inlining t0 since the inlining rule would use the Exact (or Broadcast) graph to determine the validity of inlining, but for the cloned ID of t0, since there's no def or use, it would have no mapping with any of the other IDs, which means the inlining would fail.

The mapping option added in this PR workarounds the issue. While I am not completely confident about forcing certain ID mappings, I think it's a reasonable compromise between safety and flexibility. Besides, the mapping is currently only allowed if two IDs are the same per IterDomain::sameAs.

### DIFF
--- a/csrc/id_model/id_model.cpp
+++ b/csrc/id_model/id_model.cpp
@@ -298,7 +298,7 @@ ValGraph& IdModel::buildExactGraph() {
       for (auto c_id :
            getSortedKeys(exact_c2p_logical_map, Statement::lessThan)) {
         auto p_id = exact_c2p_logical_map.at(c_id);
-        idGraph(IdMappingMode::EXACT).mapVals(c_id, p_id);
+        graph.mapVals(c_id, p_id);
       }
     }
 
@@ -312,7 +312,7 @@ ValGraph& IdModel::buildExactGraph() {
         for (auto domain_i : c10::irange(c_tv->getMaybeRootDomain().size())) {
           auto c_id = c_tv->getMaybeRootDomain()[domain_i];
           auto o_id = other_tv_output->getMaybeRootDomain()[domain_i];
-          idGraph(IdMappingMode::EXACT).mapVals(o_id, c_id);
+          graph.mapVals(o_id, c_id);
         }
       }
     } else {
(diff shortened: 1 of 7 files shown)


naoyam:
!build

csrc/id_model/id_model.cpp
```diff
@@ -323,19 +325,42 @@ ValGraph& IdModel::buildExactGraph() {
...
+  // appear in this IdModel and when they are the same (per sameAs).
+  if (!tv_exprs_.empty()) {
+    Fusion* fusion = tv_exprs_.front()->fusion();
+    if (fusion->hasRegisteredExactMappings()) {
+      DisjointSets<IterDomain*> additional_mappings =
+          fusion->registeredExactMappings();
+      for (const auto& disjoint_set : additional_mappings.disjointSets()) {
+        auto num_ids = disjoint_set->size();
+        for (const auto i : c10::irange(num_ids)) {
+          for (const auto j : c10::irange(i + 1, num_ids)) {
```
zasdfgbnm, line 348: nit: Do I have to map all the pairs? Can I just `for (const auto i : c10::irange(1, num_ids))` and in the loop, just map `disjoint_set[i]` with `disjoint_set[0]` ?

csrc/ir/cloner.h
```diff
@@ -93,6 +94,27 @@ class IrCloner {
...
+  template <typename T, typename Hash = std::hash<T>>
+  DisjointSets<T, Hash> clone(const DisjointSets<T, Hash>& disjoint_sets) {
+    DisjointSets<T, Hash> cloned_disjoint_sets;
+    for (const auto& original_set : disjoint_sets.disjointSets()) {
+      NVF_ERROR(!original_set->empty());
+      bool first = true;
+      for (const auto& val : *original_set) {
+        typename DisjointSets<T, Hash>::DisjointSet new_set;
+        if (first) {
+          auto it = cloned_disjoint_sets.initializeSet(val).first;
```
zasdfgbnm, line 106: Do I need to use `clone(val)` here?

csrc/ir/cloner.h
```diff
@@ -93,6 +94,27 @@ class IrCloner {
...
+          new_set = it->second;
+          first = false;
+        } else {
+          cloned_disjoint_sets.appendToSet(val, new_set);
```
zasdfgbnm, line 110: Same here, do I need to use `clone(val)` here?

csrc/id_model/id_model.cpp
naoyam: You're right. That should be sufficient. Will update.

zasdfgbnm (approved):

PR #2980, opened 2024-09-20
Allow profiling a benchmark iteration.

wujingyue:
See code comments for details.

### DIFF
--- a/tests/python/test_transformer_engine.py
+++ b/tests/python/test_transformer_engine.py
@@ -24,16 +24,11 @@ class ComputeType(Enum):
 
 
 # This benchmark is instrumented with cudaProfilerStart/Stop. Therefore, one
-# can collect stats of a non-warmup benchmark iteration using
+# can collect stats of the first few non-warmup benchmark iterations using
 #
 # ```bash
-# mpirun -np <processes> nsys profile --capture-range=cudaProfilerApi --capture-range-end=stop pytest tests/python/test_transformer_engine.py -k <filter> --only-mpi
+# mpirun -np <processes> nsys profile --capture-range=cudaProfilerApi --capture-range-end=repeat:<iterations> pytest tests/python/test_transformer_engine.py -k <filter> --only-mpi
 # ```
-#
-# Because of `--capture-range-end=stop`, the above command only profiles the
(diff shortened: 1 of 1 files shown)


wujingyue:
!build

tests/python/test_transformer_engine.py
```diff
@@ -68,28 +81,58 @@ def test_transformer_layer(multidevice_test, benchmark, compute_type):
...
         case ComputeType.FORWARD:
 
-            def benchmark_fn():
-                return transformer_layer(x)
+            def benchmark_fn(profile):
+                if profile:
+                    torch.cuda.cudart().cudaProfilerStart()
+
+                y = transformer_layer(x)
+                torch.cuda.synchronize()
```
wujingyue, line 89: I totally forgot about having to synchronize. The report timing makes much more sense now. Phew!

tests/python/test_transformer_engine.py
```diff
@@ -21,6 +23,17 @@ class ComputeType(Enum):
...
 
+# This benchmark is instrumented with cudaProfilerStart/Stop. Therefore, one
+# can collect stats of a non-warmup benchmark iteration using
+#
+# ```bash
+# mpirun -np <processes> nsys profile --capture-range=cudaProfilerApi --capture-range-end=stop pytest tests/python/test_transformer_engine.py -k <filter> --only-mpi
+# ```
+#
+# Because of `--capture-range-end=stop`, the above command only profiles the
+# **first** iteration. This should be enough because timing changes very little
```
cowanmeg, line 34: Might want to change to capture at least 2 iterations. In general the first iteration can be significantly slower if the gpus start at different times because the allreduce synchronizes them and nsys profiling overhead.

cowanmeg (approved):

tests/python/test_transformer_engine.py
wujingyue: I meant to say the first non-warmup iteration. Is that good enough?

cowanmeg: No, this is separate from warmup. It's overhead caused from starting the profiler

wujingyue: Done -- thanks!

PR #2983, opened 2024-09-21
Extend MaxPosCalculartor::getMaxProducerPosFromConsumer to support non-conventional loop domains

naoyam:
Part of #2902. See the linked design doc for the background context.

~Stacked on #2984~
Stacked on #2987 

The IdModel-based analysis is required for non-conventional loop domains as the dependency from the logical to loop domains may not be just one way from logical to loop. However, the IdModel-based analysis does not support broadcast forwarding, so we can't just completely switch to the new method.

For now, I think it makes most sense to use the existing method whenever logical and loop domains are generated in the conventional scheduling primitives such as split and merge. The new method is only used otherwise.

I also considered implementing the broadcast forwarding in the IdModel-based approach as well, but it doesn't seem to me worthwhile doing so at this moment. I thought maybe using the Permissive graph would make it trivial, but there seem to be (not a small number of) subtle corner cases of mappings that may not be considered in the Permissive graph, e.g., if the indexed domains should be mapped or not. 

It should be certainly possible to extend the Permissive graph to address these corner cases, but another factor I think we should consider is that the broadcast forwarding itself may not be necessary with the setLoopDomain-based approach. We should probably set the right loop domain from the beginning instead of trying to make different loop domains to be inlinable.

### DIFF
--- a/csrc/inlining.cpp
+++ b/csrc/inlining.cpp
@@ -5,7 +5,6 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 // clang-format on
-#include <device_lower/utils.h>
 #include <id_model/utils.h>
 #include <inlining.h>
 #include <ir/utils.h>
@@ -167,14 +166,14 @@ size_t MaxPosCalculator::getMaxProducerPosFromConsumer(
   // TODO: Consider caching these properties in TensorView as they
   // could only change with setLoopDomain
   const bool may_need_forwarding =
-      lower_utils::hasRootToLoopLinearTransformations(producer) &&
+      ir_utils::hasRootToLoopLinearTransformations(producer) &&
       !ir_utils::compareDomains(
            producer->getLoopDomain(),
            producer->getLogicalDomain(),
            /*additional_ids=*/{},
            /*ignore_broadcast=*/false)
            .dom0_has_unreachable_ids &&
-      lower_utils::hasRootToLoopLinearTransformations(consumer) &&
+      ir_utils::hasRootToLoopLinearTransformations(consumer) &&
       !ir_utils::compareDomains(
            consumer->getLoopDomain(),
            consumer->getLogicalDomain(),
--- a/csrc/inlining.cpp
+++ b/csrc/inlining.cpp
@@ -200,8 +200,6 @@ size_t MaxPosCalculator::getMaxProducerPosFromConsumer(
     }
     return producer->nDims();
   } else {
-    std::cerr << "Use idmodel-based method: " << producer->toString() << ", "
-              << consumer->toString() << "\n";
     auto consumer_it = consumer->getLoopDomain().begin();
     for (const auto producer_pos : c10::irange(producer->nDims())) {
       auto p_id = producer->getLoopDomain().at(producer_pos);
--- a/csrc/inlining.cpp
+++ b/csrc/inlining.cpp
@@ -169,11 +169,17 @@ size_t MaxPosCalculator::getMaxProducerPosFromConsumer(
   const bool may_need_forwarding =
       lower_utils::hasRootToLoopLinearTransformations(producer) &&
       !ir_utils::compareDomains(
-           producer->getLoopDomain(), producer->getLogicalDomain())
+           producer->getLoopDomain(),
+           producer->getLogicalDomain(),
+           /*additional_ids=*/{},
+           /*ignore_broadcast=*/false)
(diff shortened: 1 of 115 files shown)


naoyam:
!build

csrc/inlining.cpp
```diff
@@ -131,29 +143,93 @@ size_t MaxPosCalculator::getMaxPosSelf(
...
+  // at the same time allow inlinig with explicitly set loop domains,
+  // the legacy method is used whenever both the producer and consumer
+  // have loop domains that are derived from their logical domains
+  // with no redundancy. Otherwise, the IdModel-based method is used.
+
+  // TODO: Consider caching these properties in TensorView as they
+  // could only change with setLoopDomain
+  const bool may_need_forwarding =
+      lower_utils::hasRootToLoopLinearTransformations(producer) &&
+      !ir_utils::compareDomains(
```
naoyam, line 171: side note: here, the order of `compreDomains` needs to be `loop` and `logical`. It doesn't seem to work with `logical and `loop`. I'll fix it in a separate PR.

naoyam:
!build

csrc/inlining.h
```diff
@@ -55,7 +62,7 @@ class MaxPosCalculator {
   size_t getMaxProducerPosFromConsumer(
       TensorView* producer,
       TensorView* consumer,
-      bool best_effort) const;
+      bool best_effort);
```
naoyam, line 65: Needed to drop `const` as `inliningGraph` lazily builds a graph

csrc/inlining.cpp
```diff
@@ -131,29 +143,91 @@ size_t MaxPosCalculator::getMaxPosSelf(
...
+              consumer, producer, producer_pos + 1) < 0) {
+        return producer_pos;
+      }
+      auto map_it = p2c_replay_map.find(producer->axis(producer_pos));
+      if (map_it != p2c_replay_map.end()) {
+        auto c_id = map_it->second;
+        if (!isAllowedID(c_id, consumer, best_effort, true, false, true)) {
+          return producer_pos;
+        }
+      }
```
naoyam, line 199: This part is the same as before.

csrc/inlining.cpp
naoyam: Issue: #2985

naoyam:
!build

csrc/logical_domain_map.cpp
```diff
@@ -742,7 +742,7 @@ bool ComputeAtLogicalDomainMap::canMap(
     const TensorDomain* td_b,
     const IterDomain* id_b) const {
   NVF_ERROR(
-      id_b->definition() == nullptr || id_b->isRFactorProduct(),
+      td_b->isLogical(id_b) || td_b->isRoot(id_b),
```
naoyam, line 745: Bug fix. Related: https://github.com/NVIDIA/Fuser/pull/2961#discussion_r1765815241

csrc/ir/utils.cpp
```diff
@@ -797,7 +797,8 @@ std::vector<TensorView*> getTVsWithDynamicTransform(Fusion* fusion) {
 CompareDomainResult compareDomains(
     std::vector<IterDomain*> dom0,
     const std::vector<IterDomain*>& dom1,
-    const std::vector<IterDomain*>& additional_ids) {
+    const std::vector<IterDomain*>& additional_ids,
+    bool ignore_broadcast) {
```
naoyam, line 801: Broadcast IDs also need to be considered to detect extra IDs added by `setLoopDomain`.

zasdfgbnm (approved):

csrc/inlining.cpp
```diff
@@ -131,29 +143,97 @@ size_t MaxPosCalculator::getMaxPosSelf(
...
+  // At this moment, in order to keep the existing behavior as is and
+  // at the same time allow inlinig with explicitly set loop domains,
+  // the legacy method is used whenever both the producer and consumer
+  // have loop domains that are derived from their logical domains
+  // with no redundancy. Otherwise, the IdModel-based method is used.
+
+  // TODO: Consider caching these properties in TensorView as they
+  // could only change with setLoopDomain
+  const bool may_need_forwarding =
+      lower_utils::hasRootToLoopLinearTransformations(producer) &&
```
zasdfgbnm, line 170: Should we make this `ir_utils` instead of `lower_utils`? It's strange that inlining uses lower utils.

PR #2984, opened 2024-09-21
Fix TensorDomain::allIDs when loop domain has extra IDs

naoyam:
This is a follow-up to #2937, which allowed the loop domain of a tensor to have extra IDs (e.g., for inlining a 1D tensor to a 2D tensor).

When working on #2983, I realized `TensorDomain::allIDs` has a problem of finding all IDs with extra loop IDs. As mentioned in the added code comment, the issue is primarily because `IRBFS::getExprsBetween` is asymmetric with respect to its two domain parameters. See the added test for a simple concrete example.


naoyam:
!build

zasdfgbnm (approved):

PR #2990, opened 2024-09-23
Fix compile error in multi-matmul scheduler

jacobhinkle:
`main` is currently not compiling due to the changes introduced in #2913. There was a change on main that I didn't see between when CI passed and when I merged. This PR fixes that.


jacobhinkle:
!build

naoyam (approved):

PR #2991, opened 2024-09-23
Host irs: Coalescing

samnordmann:
# What

add Host IRs representing `c10d::Backend::startCoalescing()` and `c10d::Backend::endCoalescing()`.

Dependency on:

- [x] #2970

For `ProcessGroupNCCL` the methods `startCoalescing` and `endCoalescing` correspond to `ncclGroupStart` and `ncclGroupEnd` respectively, see [doc here](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/groups.html). Those calls group p2p calls that need to be progressed together -- one global work handle returned by `endCoalescing` needs to be progressed. This has the following main advantages
- calls are progressed concurrently
- since NICs are two-sided, a send and a recv calls need to be coalesced to achieve full BW
- If not coalesced, we can easily reach a deadlock if the send/recv pairs are not ordered correctly. E.g., 
rank0:
```
send to rank 1
recv from rank 1 
```
rank1:
```
send to rank 0
recv from rank 0 
```
This situation created a deadlock because no rank can receive before it has sent.

It is in general preferable to coalesce send/recv calls. The only drawback is that we don't have a fine-grain control on synchronicity, in other words, we can only synchronize with the grouped communication at once.

**Remark:** `ProcessGroupUCC` does not implement coalesced groups for now


csrc/host_ir/host_ir.cpp
```diff
@@ -237,6 +237,56 @@ bool Synchronize::sameAs(const Statement* other) const {
...
+  std::stringstream ss;
+  indent(ss, indent_size) << "StartCoalescing" << std::endl;
+  return ss.str();
+}
+
+std::string StartCoalescing::toInlineString(int indent_size) const {
+  NVF_CHECK(false, "Cannot be printed inline");
+}
+
+// TODO: implement
```
wujingyue, line 260: Doesn't [the default one](https://github.com/NVIDIA/Fuser/blob/2c8a7c657857912c1fefe821ee37d3b18118b8f1/csrc/ir/base_nodes.cpp#L331) just work?

wujingyue (approved):

csrc/host_ir/host_ir.cpp
samnordmann: You are right, I removed those methods.

BTW, should we consider adding comparison of attributes in the default sameAs? If yes, we could probably remove a lot of overriding definitions

samnordmann:
!build

samnordmann:
!build

csrc/host_ir/host_ir.cpp
wujingyue: Check sameOp. I believe that compares attributes but I'm afk to double check

PR #2992, opened 2024-09-23
Multiple Matmul Scheduler: schedule prologue

jacobhinkle:
This is the next PR in a series leading up to #2458. The previous PR 2913 scheduled the smem store and added the test framework. This PR schedules the smem load and the prologue, i.e. everything up to the mma instruction.

The next PRs in the series will do the following:
- Schedule the mma op result
- Schedule the epilogue including the split-K tensor
- Add test with horizontal fusion of matmuls (#2458)
- Switch to multiple matmul scheduler, i.e. remove old scheduleMatmul()
- Update matmul heuristic to account for horizontal fusion
- Update canSchedule checks to accept horizontally fused matmul segments


csrc/mma_type.cpp
```diff
@@ -142,4 +142,19 @@ size_t hash(const MatMulTileOptions& opts) {
...
+      return "M";
+    case MatmulDimRole::N:
+      return "N";
+    case MatmulDimRole::K:
+      return "K";
+  }
+  // Unreachable
+  return "Unrecognized role";
+}
+
```
jacobhinkle, line 159: Not strictly needed but helpful for debugging.

jacobhinkle:
!build

csrc/scheduler/mma_utils.cpp
```diff
@@ -1210,6 +1382,24 @@ std::vector<MatmulDimRole> canonicalizeMmaTvOrdering(
   return roles;
 }
 
+void mergeCanonicalAbstractTensor(AbstractMatmulTensor& abstract_tensor) {
```
jacobhinkle, line 1385: This is currently unused but will be useful for doing `AbstractTensor` scheduling in the future.

csrc/scheduler/mma_utils.h
```diff
@@ -61,7 +70,24 @@ NVF_API void scheduleWarpTileWithNoReduction(
...
+    const GemmTile& tile_sizes,
+    const std::vector<MatmulDimRole>& axis_roles);
+
+//! We model each dimension of every tensor in the Fusion with ID roles
+//! described by MatmulDimRole.
+using AbstractMatmulTensor = TaggedAbstractTensor<MatmulDimRole>;
+
+//! Abstract version of the above utility. Schedules the provided
+//! AbstractTensor and returns the resulting ID roles.
+void makeTile(
```
zasdfgbnm, line 88: Looks like a duplicate due to resolve conflict

csrc/scheduler/mma_utils.h
```diff
@@ -96,6 +122,13 @@ std::vector<MatmulDimRole> canonicalizeMmaTvOrdering(
     const DimRolesMap& dim_roles,
     const std::vector<ValGroup>& ordering);
 
+//! Given a TensorView matching the canonicalDimOrdering, schedule it by
+//! merging dimensions with matching roles.
+void mergeAxesWithSameRole(
```
zasdfgbnm, line 127: Is this an artifact of resolve conflict?

csrc/scheduler/mma_utils.cpp
```diff
@@ -281,6 +281,143 @@ void scheduleWarpTileWithReduction(TensorView* tv, MatMulTileOptions tile) {
...
+      k_dim != -1,
+      "scheduleWarpTile should only be called on an operand with K dimension but found ",
+      tv->toString());
+
+  mma_utils::checkDimSize(tv, inner_dims, required_inner_dim_sizes);
+
+  //     m_dim  n_dim k_dim
+  //[...     M,     N,    K]
+  // Distribute warp tile:
+  if (m_dim != -1) {
```
zasdfgbnm, line 341: Does `-1` mean "the last dim", or "not found"? We might want to use `std::optional<int64_t> m_dim` for better readability here.

csrc/scheduler/mma_utils.cpp
```diff
@@ -281,6 +281,143 @@ void scheduleWarpTileWithReduction(TensorView* tv, MatMulTileOptions tile) {
...
+  if (n_dim != -1) {
+    tv->split(n_dim, warp_tile.n);
+    if (m_dim > n_dim) {
+      m_dim++;
+    }
+    if (k_dim > m_dim) {
+      k_dim++;
+    }
+  }
+
```
zasdfgbnm, line 359: Is `warp_tile.k` always equal to `cta_tile.k`? It would be good to add a `NVF_ERROR` here for better readability.

csrc/scheduler/mma_utils.cpp
```diff
@@ -281,6 +281,143 @@ void scheduleWarpTileWithReduction(TensorView* tv, MatMulTileOptions tile) {
...
+    }
+    if (n_dim > k_dim) {
+      n_dim++;
+    }
+  }
+
+  //  m_dim  m_dim+1  m_dim+2  n_dim  n_dim+1  n_dim+2  k_dim  k_dim+1
+  // [  Mwo       Mw       Mi    Nwo       Nw       Ni    Kwo       Ki]
+
+  // Nowwe reorder. The new_order will be
```
zasdfgbnm, line 393: nit: `Now we`

csrc/scheduler/multi_matmul.cpp
```diff
@@ -1006,16 +1006,179 @@ class MultipleMatmulScheduler {
...
+    // Hold this vector so we can use it as a boundary to propagate backward
+    // from each mma input.
+    std::vector<TensorView*> all_smem_stores = acw_smems_;
+    all_smem_stores.insert(
+        all_smem_stores.end(), bcw_smems_.begin(), bcw_smems_.end());
+
+    // Now for each operand, we load from smem to registers and compute a
+    // prologue (generally) in registers. We typically refer to the register
+    // buffer that is loaded from operand A's smem buffer using ldmatrix as
+    // "acr". This is the beginning of the register epilogue region for that
```
zasdfgbnm, line 1100: Do you mean `register prologue region`?

csrc/scheduler/multi_matmul.cpp
```diff
@@ -1006,16 +1006,179 @@ class MultipleMatmulScheduler {
...
+      }
+
+      scheduleMmaOperands(mma_inputs, operand_type);
+
+      // Propagate backward from all mma_results to smem_stores
+
+      for (TensorView* mma_input : mma_inputs) {
+        // Schedule mma_input, since we know it has the broadcast dimension M or
+        // N, whereas the smem read might not
+        if (isTuring(params_->mma_macro) || isAmpere(params_->mma_macro)) {
```
zasdfgbnm, line 1139: Is this always true?

jacobhinkle:
!build

csrc/scheduler/multi_matmul.cpp
```diff
@@ -1006,16 +1006,179 @@ class MultipleMatmulScheduler {
     scheduleBranch(bs_, bcw_smems_, params_->supported_vec_size.b);
   }
 
+  // Schedule acr and bcr
+  void scheduleLdMatrixLoads() {
```
zasdfgbnm, line 1010: Is this used?

csrc/scheduler/mma_utils.cpp
jacobhinkle: Yes. That check is at line 294.

csrc/scheduler/mma_utils.cpp
jacobhinkle: Sounds good. It means "not found", but you're right it's more readable to use optionals.

csrc/scheduler/multi_matmul.cpp
jacobhinkle: Yes, typo. Thanks for noticing.

csrc/scheduler/multi_matmul.cpp
jacobhinkle: Ah you're right. It is no longer used. I'll remove it.

csrc/scheduler/multi_matmul.cpp
jacobhinkle: I am not sure we always need to do this, but this is in the current scheduler. If you mean is the comment accurate I think so: `mma_input` is an input to the `MmaOp` so it should have an inner broadcast dim.

csrc/scheduler/mma_utils.cpp
```diff
@@ -281,6 +281,143 @@ void scheduleWarpTileWithReduction(TensorView* tv, MatMulTileOptions tile) {
   // [Kwo Mwo Nwo Mw Nw Mi Ni Ki]
 }
 
+void scheduleWarpTile(
```
zasdfgbnm, line 284: Actually, is this still used after the removal of `scheduleLdMatrixLoads`?

csrc/scheduler/multi_matmul.cpp
zasdfgbnm: I think this `if` existed for the purpose of dispatching Turing/Ampere vs Volta. But support for Volta has been dropped, so I don't think we should keep it anymore.

csrc/scheduler/mma_utils.cpp
jacobhinkle: You're right. I reverted a change that was using this. Now we don't use it. We might use it in the future but I'll remove it until then.

csrc/scheduler/multi_matmul.cpp
jacobhinkle: Removed the condition.

csrc/scheduler/multi_matmul.cpp
```diff
@@ -1006,16 +1006,146 @@ class MultipleMatmulScheduler {
...
+      operand->axis((int64_t)merged_roles.size() + num_splitk_dims_ + 2)
+          ->parallelize(ParallelType::TIDy);
+    }
+  }
+
+  // MmaOperand contains only A and B. If tvs are outputs (i.e. not operands),
+  // then operand_type should be std::nullopt.
   void scheduleMmaResults() {
-    NVF_THROW("scheduleMmaResults is not yet implemented");
+    NVF_THROW("scheduleEpilogue is not yet implemented");
```
zasdfgbnm, line 1053: Why this change? Will this function be renamed in the future?

csrc/scheduler/multi_matmul.cpp
```diff
@@ -1006,16 +1006,146 @@ class MultipleMatmulScheduler {
...
+      std::unordered_set<TensorView*> smem_load_mma_inputs;
+      for (TensorView* smem_load : smem_loads) {
+        // Insert only if smem_load is also in mma_inputs
+        bool is_mma_input =
+            std::find(mma_inputs.begin(), mma_inputs.end(), smem_load) !=
+            mma_inputs.end();
+        if (is_mma_input) {
+          smem_load_mma_inputs.insert(smem_load);
+        }
+        if (!is_mma_input) {
```
zasdfgbnm, line 1127: This logic looks odd to me. I believe it is reqiured because in `applyMmaSwizzle` and `scheduleOperandRead`, there are logics that does things differently for input vs non-input. Could we refactor `applyMmaSwizzle` to make all of `applyMmaSwizzle`, `scheduleOperandRead` and this function simpler? Let's do this in a separate PR. This  PR as-is looks good.

zasdfgbnm (approved):

csrc/scheduler/multi_matmul.cpp
jacobhinkle: Cut/paste error! Will correct it.

csrc/scheduler/multi_matmul.cpp
jacobhinkle: I agree. This felt ugly to write, but it corresponds to the condition here https://github.com/NVIDIA/Fuser/blob/027cd0f5b5e723ab5c57c75da9ec8852e0e23285/csrc/scheduler/matmul.cpp#L1114
I'll look at cleaning this up in a separate PR, as you suggest.

PR #2996, opened 2024-09-23
Only the first local rank prints segmented fusions.

wujingyue:
Similar to #2895.

### DIFF
--- a/csrc/fusion_segmenter.cpp
+++ b/csrc/fusion_segmenter.cpp
@@ -14,7 +14,8 @@
 #include <ir/graphviz.h>
 #include <ir/iostream.h>
 #include <ir/utils.h>
-#include <multidevice/utils.h>
+#include <multidevice/communicator.h>
+#include <multidevice/utils.h> // For isResharding
 #include <ops/arith.h>
 #include <options.h>
 #include <scheduler/debug_utils.h>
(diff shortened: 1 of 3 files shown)


wujingyue:
!build

jacobhinkle:
Would it make sense to set `debug()` to a trivial stream if rank is not zero? Otherwise we might need to do this for every debug option separately wouldn't we?

wujingyue:
> Would it make sense to set `debug()` to a trivial stream if rank is not zero? Otherwise we might need to do this for every debug option separately wouldn't we?

Different ranks may diverge during execution. For example, https://github.com/NVIDIA/Fuser/blob/cd1e4f560ff3c7fa6fd56ed9a7f70a74d7914340/csrc/fusion_executor/executor.cpp#L1364-L1365 may print out different profiles. 

However, I share the same concern. How about something like `debug(/*only_first_local_rank=*/true)` so the caller can choose whether all ranks should print?

wujingyue:
> How about something like debug(/*only_first_local_rank=*/true) so the caller can choose whether all ranks should print?

This doesn't work out of the box because `Fusion::printMath` and `Fusion::printTransforms` use `debug()` directly. DebugStreamGuard may be able to do magic but I've yet to figure it out.

csrc/fusion_segmenter.cpp
```diff
@@ -1969,7 +1970,23 @@ std::pair<IrCloner, std::unique_ptr<Fusion>> SegmentedFusion::makeFusion(
...
 }
 
-std::unique_ptr<SegmentedFusion> SegmentCandidateFinder::segment(
+/*static*/ std::unique_ptr<SegmentedFusion> SegmentCandidateFinder::segment(
+    std::unique_ptr<Fusion> fusion,
+    const KernelArgumentHolder* inputs,
+    const SegmentCandidateFinderOptions& options) {
+  if (isDebugDumpEnabled(DebugDumpOption::FusionSegments)) {
+    const auto& communicator = Communicator::getInstance();
+    if (!communicator.is_available() || communicator.local_rank() == 0) {
```
samnordmann, line 1979: Why using `local_rank`? If we only want only one rank to print, then we need to use `deviceId` instead

samnordmann (commented):
It is sometimes useful to print debug from all ranks, or a certain rank in particular, since, for now, all ranks do not run the same program. So we may not want to impose that rank 0 only can print.

It is relatively easy to separate the different debug dumps of each rank from the command line, with something like
`mpirun -np 8 bash -c './your_mpi_program > logs/output_rank_$OMPI_COMM_WORLD_RANK.log`.

samnordmann:
An alternative would be to prefix each debug print by the rank and PID, as is done in many distributed libraries

wujingyue:
> all ranks do not run the same program and the same segmented fusion

For my education, when do they diverge? I've been expecting fusion IR and segmented fusion to be the same. Host IR may diverge because it's in flux.

samnordmann:
> > all ranks do not run the same program and the same segmented fusion
> 
> For my education, when do they diverge? I've been expecting fusion IR and segmented fusion to be the same. Host IR may diverge because it's in flux.

In case of pipelining, if some rank does not run a stage of the pipeline, that stage will be completely removed from that rank's program, instead of being predicated (as it should be for achieving true SPMD)

PR #2997, opened 2024-09-23
Extend the host latency benchmarking for additional cases.

Priya2698:
We currently measure the first time overhead of a fusion execution. The FusionCache is reset at the start of every benchmarking round to measure this overhead correctly. This work aims at extending the current setup for measuring host latencies in the python benchmarks to cover the following cases:

- Case 1: Compilation through NVRTC (Host time – first time overhead): This measures the first time overhead when executing a fusion definition (current setup)
- Case 2: Cache Hit overhead: This ignores the first time overhead, and measures the host latency when executing the same input multiple times.
- Case 3: Dynamic Shape overhead (not counting first time overhead): This ignores the first time overhead, and measures the host latency when executing different inputs shapes for the same fusion definition.

Screenshot 2024-09-23 at 2 31 37 PM

### DIFF
--- a/benchmarks/python/core.py
+++ b/benchmarks/python/core.py
@@ -267,6 +267,11 @@ def run_benchmark(
     ], f'Unsupported device type: {device.split(":")[0]}. Use one of cuda/host.'
 
     host_bench_mode = None
+
+    # Store warmup rounds locally to modify for host:steady/dynamic cases.
+    global warmup_rounds
+    warmup_rounds = BENCHMARK_CONFIG["warmup_rounds"]
+
     if device.split(":")[0] == "host":
         # Host benchmarking expects a fusion function to generate fusion definitions everytime FusionCache is reset.
         assert fusion_fn is not None and benchmark_fn is None
@@ -286,9 +291,9 @@ def run_benchmark(
         ):
             # By default, warmup_rounds=1. If BENCHMARK_CONFIG['warmup_rounds'] == 0 through --benchmark-warmup-rounds, raise a warning that it was ignored.
             warnings.warn(
-                "--benchmark-warmup-rounds=0 was ignored for host benchmarking. Setting warmup_rounds=1."
+                "--benchmark-warmup-rounds=0 is ignored for host:steady/dynamic benchmarking. Setting warmup_rounds=1."
             )
-            BENCHMARK_CONFIG["warmup_rounds"] = 1
+            warmup_rounds = 1
 
     """
     Setup function: This is called before each benchmarking round. This function is used to:
@@ -355,7 +360,7 @@ def run_benchmark(
         benchmark_fn,
         setup=setup,
         rounds=BENCHMARK_CONFIG["rounds"],
-        warmup_rounds=BENCHMARK_CONFIG["warmup_rounds"],
+        warmup_rounds=warmup_rounds,
     )
 
     if device == "cuda":
--- a/benchmarks/python/core.py
+++ b/benchmarks/python/core.py
@@ -9,6 +9,7 @@ from typing import List, Callable, Union
 import numpy as np
 from nvfuser import FusionDefinition, FusionCache
 from nvfuser.pytorch_utils import DEVICE_PROPERTIES
+from itertools import cycle
 
 # These variables can be overwritten through CLI commands
 # --benchmark-rounds=rounds --benchmark-warmup-rounds=warmup_rounds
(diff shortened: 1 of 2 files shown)


benchmarks/python/core.py
```diff
@@ -259,33 +260,95 @@ def run_benchmark(
...
+        # Set the warmup rounds if required for `steady/dynamic` host latency measurement.
+        if (
+            host_bench_mode in ["steady", "dynamic"]
+            and BENCHMARK_CONFIG["warmup_rounds"] == 0
+        ):
+            # By default, warmup_rounds=1. If BENCHMARK_CONFIG['warmup_rounds'] == 0 through --benchmark-warmup-rounds, raise a warning that it was ignored.
+            warnings.warn(
+                "--benchmark-warmup-rounds=0 was ignored for host benchmarking. Setting warmup_rounds=1."
+            )
+            BENCHMARK_CONFIG["warmup_rounds"] = 1
```
naoyam, line 291: IIUC `BENCHMARK_CONFIG` is like a global configuration shared by multiple benchmark runs. If so, it seems a little odd that it gets modified while executing one instance of benchmarks. What would happen both host and non-host benchmarks were run? Wouldn't that affect non-host benchmarks as well?

Priya2698: This was bug, thanks for pointing it out. It has been resolved now. 

Running the following command:
```
pytest benchmarks/python/test_many_pointwise_ops.py benchmarks/python/test_pointwise_mul.py --benchmark-warmup-rounds=0 --benchmark-num-inputs=1 -s -v
```
where `--benchmark-num-inputs=1` is to only run 1 input size for `test_pointwise_mul.py`.

Screenshot 2024-09-24 at 5 24 40 PM

The output shows that the CLI setting for `warmup_rounds` is adhered to for non-host benchmarking in `test_pointwise_mul.py` as well as `host:compile`.
_(Note: The warmup rounds are not automatically printed on stdout, I inserted a print statement for easier debugging and demonstration.)_

benchmarks/python/core.py
```diff
@@ -259,40 +260,107 @@ def run_benchmark(
...
         "cuda",
         "host",
-    ], f"Unsupported device type: {device}. Use one of cuda/host."
+    ], f'Unsupported device type: {device.split(":")[0]}. Use one of cuda/host.'
 
-    if device == "host":
+    host_bench_mode = None
+
+    # Store warmup rounds locally to modify for host:steady/dynamic cases.
+    global warmup_rounds
```
naoyam, line 272: Why does this need to be `global`?

benchmarks/python/core.py
```diff
@@ -259,40 +260,107 @@ def run_benchmark(
...
+    1. Clear L2 cache.
+    2. For host latency benchmarks, the 3 modes use different setups.
+        a) 'compile': FusionCache is reset at every round to measure the first time overhead before instantiating fd.
+        b) 'steady': Nothing additional is required. The warmup round avoids including the first time overhead in the measurements.
+        c) 'dynamic': We maintain a global counter to track which input is executed. Once all the inputs have been executed once,
+        the FusionCache is reset again and we execute fd for the first input to avoid including the first time compile overhead in the dynamic measurement.
+    """
+
+    # Counter used in `dynamic` host latency benchmarking, unused in other cases.
+    global counter
```
naoyam, line 309: same here

Priya2698: This needs to be declared global to modify within different calls of the `setup` function.

benchmarks/python/core.py
Priya2698: This was a remnant from some code changes I was trying.
Removed it now.

benchmarks/python/core.py
naoyam: IIRC, you don't need to have `global` here since you have `global` inside the `setup` function (at line 336), but I'm not an active user of Python anymore, so maybe I'm wrong.

naoyam (approved):
LGTM

benchmarks/python/core.py
Priya2698: It is needed here since the declaration of `counter` is within the `run_benchmark` function itself, so by default, the scope of `counter=0` without the global declaration will be the outer function and the inner `setup` function cannot modify this local variable of the outer function. If it was defined at the module level, outside of this function, we would not need this `global`

Issue #2664, opened 2024-07-23
Numerically incorrect results with calculation on large arrays


shino16:
The following script gave numerically incorrect values on a small part of the array.
```py
import torch
import thunder


def f(a, b, c):
    c += 1
    return a.add(b, alpha=c)


dim = 2**22
a = torch.arange(dim, dtype=torch.float32, device='cuda')
b = torch.arange(dim, dtype=torch.float32, device='cuda')
c = torch.tensor(1, dtype=torch.float32, device='cuda')

jitted_f = thunder.jit(f)
ret = jitted_f(a, b, c)

expected = torch.arange(dim, dtype=torch.float32, device='cuda') * 3
torch.testing.assert_close(ret, expected)
```
```
AssertionError: Tensor-likes are not close!

Mismatched elements: 6144 / 1048576 (0.6%)
Greatest absolute difference: 980479.0 at index (980479,) (up to 1e-05 allowed)
Greatest relative difference: 0.3333333432674408 at index (568832,) (up to 1.3e-06 allowed)
```

I generated its fusion definition by calling `FusionDefinition.getReproErrorString` manually.
```py
# CUDA devices:
#  0: NVIDIA H100 80GB HBM3
# torch version: 2.5.0a0+git01e41f1
# cuda version: 12.5
# nvfuser version: 0.2.6+gitd75fc93
import torch
from nvfuser import FusionDefinition, DataType

def nvfuser_fusion_id0(fd : FusionDefinition) -> None :
    T0 = fd.define_tensor(shape=[-1], contiguity=[True], dtype=DataType.Float, is_cpu=False, stride_order=[0])
    T1 = fd.define_tensor(shape=[-1], contiguity=[True], dtype=DataType.Float, is_cpu=False, stride_order=[0])
    T2 = fd.define_tensor(shape=[], contiguity=[], dtype=DataType.Float, is_cpu=False)
    S3 = fd.define_scalar(1.00000, dtype=DataType.Double)
    T4 = fd.ops.add(T2, S3)
    S5 = fd.define_scalar(4194304, dtype=DataType.Int)
    V6 = fd.define_vector([S5], dtype=DataType.Int)
    T7 = fd.ops.broadcast_in_dim(T4, shape=V6, broadcast_dims=[])
    T8 = fd.ops.mul(T1, T7)
    T9 = fd.ops.add(T0, T8)
    fd.add_output(T4, T2)
    fd.add_output(T9)

with FusionDefinition() as fd:
[...]

PR #2999, opened 2024-09-23
Presegmentation pass to force segment breaks when inplace update can cause RW race

Priya2698:
Issue #2664

A RW race can occur when an intermediate tensorview is aliased to a fusion input and the intermediate tensorview or the aliased input is a producer/consumer of a broadcast op. This presegmentation pass traverses the fusion to find such inplace updates, and inserts `segmet_set + set` to force the inplace update into a separate copy kernel. This ensures that the write to the fusion input only occurs when all the reads of that fusion input have concluded.


Priya2698:
!build --diff

jacobhinkle:
> A RW race can occur when an intermediate tensorview is aliased to a fusion input and the intermediate tensorview or the aliased input is a producer/consumer of a broadcast op.

Does there need to be any broadcast ops for us to encounter this problem actually?
```python
    T0 = fd.define_tensor(shape=[-1, 1], contiguity=[True], dtype=DataType.Float, is_cpu=False, stride_order=[1, 0])
    T1 = fd.define_tensor(shape=[-1, -1], contiguity=[True], dtype=DataType.Float, is_cpu=False, stride_order=[1, 0])

    T3 = fd.ops.add(T1, T0)  #  [ i0, i1 ]
    T4 = fd.ops.neg(T0)  # [ i0, 1 ]

    fd.add_output(T3)
    fd.add_output(T4, T0)
```

Also, does the aliased input and output actually need to have Broadcast dimensions for this RAW hazard? For example we could have the following:

```python
    T0 = fd.define_tensor(shape=[-1], contiguity=[True], dtype=DataType.Float, is_cpu=False, stride_order=[0])
    T1 = fd.define_tensor(shape=[-1, -1], contiguity=[True], dtype=DataType.Float, is_cpu=False, stride_order=[1, 0])

    T2 = fd.ops.broadcast_in_dim(T0, shape=fd.ops.tensor_sizes(T1), broadcast_dims=[0])  # [ i0, 1 ]
    T3 = fd.ops.add(T1, T2)  #  [ i0, i1 ]
    T4 = fd.ops.neg(T0)  # [ i0 ]

    fd.add_output(T3)
    fd.add_output(T4, T0)
```
I think in a case like this we'll still potentially hit the problem, even though the aliased output itself doesn't involve any broadcasts, because we'd try and schedule the problem as `[ i0, i1 ]`.

I might have an error in my misunderstanding, but if these two cases do lead to the hazard then do we need a full loop promotion analysis using the pointwise scheduler to detect this?

jjsjann123:
Oops. I think the implicit broadcast part is indeed concerning... 

This is from Jacob's example 1.
```
Segmented_Fusion Dump: -- Re-written complete fusion:{
Inputs:
  T0_g_float[ iS0{i0}, bS1{1} ]
  T1_g_float[ iS8{i0}, iS3{i5} ]
Outputs:
  T2_g_float[ iS9{i0}, iS5{i5} ]
  T3_g_float[ iS6{i0}, bS7{1} ]

%kernel_math {
T2_g_float[ iS9{i0}, iS5{i5} ]
   = T1_g_float[ iS8{i0}, iS3{i5} ]
   + T0_g_float[ iS0{i0}, bS1{1} ];
T3_g_float[ iS6{i0}, bS7{1} ]
   = -T0_g_float[ iS0{i0}, bS1{1} ];
} // %kernel_math
```

So even though there's no broadcast op, It's still implied by the binay op.

jjsjann123:
Asking @jacobhinkle 
Re the second issue, 

> the intermediate tensorview or the aliased input is a producer/consumer of a broadcast op

I don't think this is saying `does the aliased input and output actually need to have Broadcast dimensions`. Tried your second example with this PR, it actually does create a segment_set.

FYI, there seems to be something with the segmenter logic again... the segmenter set somehow give us 3 segments, instead of 2 for the second example. :cry:

jjsjann123:
> do we need a full loop promotion analysis using the pointwise scheduler to detect this?

Very interested in this, can you elaborate on this?

Can we run the loop promotion analysis on preseg passes, I thought we need a scheduled fusion to do those. If this can be used to replace the traverse in this PR, that'll be great.

Alternatively, I think we can continue adding `Binary` and `Ternary` ops in the traverse and looking for potential broadcast to catch those... but that does seem like a fragile approach.

Priya2698:
> > A RW race can occur when an intermediate tensorview is aliased to a fusion input and the intermediate tensorview or the aliased input is a producer/consumer of a broadcast op.
> 
> Does there need to be any broadcast ops for us to encounter this problem actually?
> 
> ```python
> def nvfuser_fusion_id0(fd : FusionDefinition) -> None :
>     T0 = fd.define_tensor(shape=[-1, 1], contiguity=[True], dtype=DataType.Float, is_cpu=False, stride_order=[1, 0])
>     T1 = fd.define_tensor(shape=[-1, -1], contiguity=[True], dtype=DataType.Float, is_cpu=False, stride_order=[1, 0])
> 
>     T3 = fd.ops.add(T1, T0)  #  [ i0, i1 ]
>     T4 = fd.ops.neg(T0)  # [ i0, 1 ]
> 
>     fd.add_output(T3)
>     fd.add_output(T4, T0)
> ```

That's a good catch. Implicit broadcasting here will cause a RAW race and the current preseg pass will not detect this as a race condition.


> Also, does the aliased input and output actually need to have Broadcast dimensions for this RAW hazard? For example we could have the following:
> 
> ```python
> def nvfuser_fusion_id0(fd : FusionDefinition) -> None :
>     T0 = fd.define_tensor(shape=[-1], contiguity=[True], dtype=DataType.Float, is_cpu=False, stride_order=[0])
>     T1 = fd.define_tensor(shape=[-1, -1], contiguity=[True], dtype=DataType.Float, is_cpu=False, stride_order=[1, 0])
> 
>     T2 = fd.ops.broadcast_in_dim(T0, shape=fd.ops.tensor_sizes(T1), broadcast_dims=[0])  # [ i0, 1 ]
>     T3 = fd.ops.add(T1, T2)  #  [ i0, i1 ]
>     T4 = fd.ops.neg(T0)  # [ i0 ]
> 
>     fd.add_output(T3)
>     fd.add_output(T4, T0)
> ```
> 
> I think in a case like this we'll still potentially hit the problem, even though the aliased output itself doesn't involve any broadcasts, because we'd try and schedule the problem as `[ i0, i1 ]`.

In this case, T4 does not involve any broadcasts, however T0 (the aliased input) is a producer for the broadcast, so the preseg pass does insert segment set correctly.

> I might have an error in my misunderstanding, but if these two cases do lead to the hazard then do we need a full loop promotion analysis using the pointwise scheduler to detect this?

csrc/preseg_passes/segment_inplace_update.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+  // fusion->exprs() is a topologically sorted list. Filter out the broadcast
+  // ops from the list.
+  auto all_exprs = fusion->exprs();
+  auto all_bcast_ops = ir_utils::filterByType<BroadcastOp>(all_exprs);
+
+  // Traverse and store all direct/indirect consumer tensorviews of these
+  // broadcast nodes If the tensorview has been visited, return --> this means
+  // that we have already traversed that branch
+  std::unordered_set<TensorView*> visited_tvs;
```
Priya2698, line 60: There are multiple ways to do this traversal using the IdModel approach of checking for concretized broadcasts starting from aliased inputs too:
Let number of number of tensorviews = M, aliased_inputs = N in the given fusion.
- For each aliased input, traverse the graph and if any reachable tensorview has a concretized broadcast, the corresponding aliased output/tv is a candidate for inserting a segment set. This requires O(M) space to store visited tensorviews with concretized bcast and O(N) graph traversals.
- Traverse the graph starting from all aliased inputs and only visit each tensorview once, recording which tensorviews have a concretized bcast. After this, propagate this information upwards towards aliased inputs. This ensures, if during the graph traversal for any aliased input, a node was already visited and hence, the traversal did not go further, we correctly mark that aliased input/output for segment set when a concretized bcast is present further down in that branch. This requires O(M) space, and 2 graph traversals.
- The current approach requires O(M) space to record visited tensorviews, and 1 graph traversal.

This motivated using a hybrid solution which uses:
1. IdModel to check for concretized bcast in aliased inputs (the implicit bcast case)
2. Existing solution of traversing starting from bcast ops, since it required minimum number of graph traversals.

By graph traversal -- I am referring to the number of times each node (tv/expr) is visited.

Priya2698:
!build

tests/python/test_pointwise.py
```diff
@@ -84,3 +84,180 @@ def fusion_func(fd: FusionDefinition):
...
+
+    fd.execute(inputs, profile=True)
+    assert fd.profile().segments == 4
+
+    torch.testing.assert_close(inputs[1], ref_out[0])
+    torch.testing.assert_close(inputs[2], ref_out[1])
+
+# Example 4 for Issue 2664: There is no explicit broadcast. However, the aliased input has a broadcast dimension that is concretized in the fusion.
+# T0 has a implicit broadcast which is used in add(T3) and neg (T4). T4 is used to inplace update T0, which causes RW race.
+def test_implicit_bcast_inplace():
```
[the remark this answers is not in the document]
jjsjann123: This is a separate problem and we don't have to resolve it here. Can we have a short issue tracking this tho? 🙇

csrc/preseg_passes/segment_inplace_update.cpp
```diff
@@ -0,0 +1,156 @@
...
+          tvs_to_visit.push_back(input_tv->as<TensorView>());
+        }
+      }
+    }
+  }
+
+  // Use permissive IdModel graph to identify any concretized broadcast
+  // iterdomain in any aliased input.
+  auto id_model =
+      IdModel(fusion, /*build_graphs=*/false, /*allow_self_mapping=*/true);
```
jjsjann123, line 106: another naive question. What's the case with self_mapping here? i.e. is it expected to have self mapping in the graph or it's just hiding some issue that should be tracked and made aware by @naoyam ?

csrc/preseg_passes/segment_inplace_update.cpp
```diff
@@ -0,0 +1,156 @@
...
+        if (!other_id->as<IterDomain>()->isBroadcast()) {
+          return true;
+        }
+      }
+    }
+    return false;
+  };
+
+  // For all aliased tensorviews:
+  // 1) if that tv or the correspodning aliased input is a producer/consumer of
```
jjsjann123, line 133:
```suggestion
  // 1) if that tv or the corresponding aliased input is a producer/consumer of
```

jjsjann123 (approved):
LGTM

csrc/preseg_passes/segment_inplace_update.cpp
```diff
@@ -0,0 +1,156 @@
...
+// that all write operations to the fusion inputs occur after the read
+// operations have completed. See Issue #2664: https://
+// github.com/NVIDIA/Fuser/issues/2664
+namespace {
+void insertSegmentSet(Fusion* fusion) {
+  std::vector<TensorView*> aliased_tvs;
+  std::vector<BroadcastOp*> bcast;
+
+  // Find all tensorviews which are used in inplace updates
+  for (TensorView* tv : fusion->allTvs()) {
```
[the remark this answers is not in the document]
Priya2698: Good catch. `aliasOutputToInput` always adds aliases as outputs.https://github.com/NVIDIA/Fuser/blob/e01e23ef2f3b4d49e05a5e6a1f9c81119385d537/csrc/fusion.cpp#L830-L833

csrc/preseg_passes/segment_inplace_update.cpp
Priya2698: This is not needed here, removed now.

Priya2698:
!build

Priya2698:
For the test: https://github.com/NVIDIA/Fuser/blob/2ec2b926a0c589f7b72bd7a3abce7a49111c5620/tests/cpp/test_alias.cpp#L980-L1029, the fusion is segmented such that the RW race does not occur. However, this preseg pass will still insert a segment set and split the fusion into 3 segments, even though 2 segments is functionally correct.

```
} // {Re-written complete fusion}
Segmented_Fusion Dump: -- fusion segments:
Segmented_Fusion{ 
groups: 
  reduction{0, 1, 2}
  pointwise{3, 4, 5}
edges: 
  e{ reduction{0, 1, 2} -> pointwise{3, 4, 5}(T5_g_float[ iS9{i2} ]) }

group details:
g{(reduction)
group id: 0
inputs:
  T0_g_float[ iS0{i0}, iS1{i2} ] float
  T1_g_float[ iS16{i2} ] float
outputs:
  T5_g_float[ iS9{i2} ] float


T3_l_float[ iS5{i0}, iS6{i2} ]
   = T0_g_float[ iS0{i0}, iS1{i2} ]
   + double(1);
(0)
T4_g_float[ rS7{i0}, iS8{i2} ]
   = reduction( T3_l_float[ iS5{i0}, iS6{i2} ], op = fmax, initial value = double(-inf), allreduce = false )
(1)
T5_g_float[ iS9{i2} ]
   = T4_g_float[ rS7{i0}, iS8{i2} ]
   + T1_g_float[ iS16{i2} ];
(2)
}

g{(pointwise)
group id: 1
inputs:
  T0_g_float[ iS0{i0}, iS1{i2} ] float
  T2_g_float[ iS3{i4}, iS17{i2} ] float
  T5_g_float[ iS9{i2} ] float
outputs:
  T7_g_float[ iS12{i4}, iS13{i2} ] float
  T8_g_float[ iS14{i4}, iS15{i2} ] float


T6_g_float[ bS10{1}, iS11{i2} ]
   = broadcast( T5_g_float[ iS9{i2} ] )
(3)
T7_g_float[ iS12{i4}, iS13{i2} ]
   = T6_g_float[ bS10{1}, iS11{i2} ]
   + T2_g_float[ iS3{i4}, iS17{i2} ];
(4)
T8_g_float[ iS14{i4}, iS15{i2} ]
   = T7_g_float[ iS12{i4}, iS13{i2} ]
   + double(1);
(5)
}
```

Priya2698:
!build

Priya2698:
!build

jjsjann123:
:clap:

PR #3001, opened 2024-09-24
compiler warning: comparison of different signedness

jjsjann123:


CMakeLists.txt
```diff
@@ -278,7 +278,7 @@ add_library(codegen_internal OBJECT ${NVFUSER_SRCS})
 if(NOT MSVC)
   target_compile_options(codegen_internal PRIVATE
     -Wall -Wno-unused-function
-    # -Werror
```
jjsjann123, line 281: I forgot why we turned off this.

jjsjann123:
!build

jacobhinkle (approved):
Don't know why it was turned off either but thanks for catching this

CMakeLists.txt
zasdfgbnm: The manylinux container is using gcc 12.2, which is not supported. This version of gcc have a bug that makes it unable to compile one of my PR with `-Werror`.

zasdfgbnm: Ideally, we should upgrade to gcc 12.4, but I tried that, and it is very hard compile gcc 12.4 in AlmaLinux 8....

jacobhinkle: Maybe we should just turn on `-Werror` for the CI clang-build job

jjsjann123: sounds good. I'll back out from that.

CMakeLists.txt
```diff
@@ -278,6 +278,9 @@ endif()
 add_library(codegen_internal OBJECT ${NVFUSER_SRCS})
 
 if(NOT MSVC)
+  # -Werror is not enabled, because of gcc 12.2 used in manylinux image.
+  # consider enable this when we upgrad. linking comment:
+  # https://github.com/NVIDIA/Fuser/pull/3001#discussion_r1772551266
```
jjsjann123, line 283: revert the changes and added a comment on why we don't have it.

CMakeLists.txt
```diff
@@ -278,6 +278,9 @@ endif()
 add_library(codegen_internal OBJECT ${NVFUSER_SRCS})
 
 if(NOT MSVC)
+  # -Werror is not enabled, because of gcc 12.2 used in manylinux image.
+  # consider enable this when we upgrad. linking comment:
```
jacobhinkle, line 282:
```suggestion
  # consider enable this when we upgrade. linking comment:
```

jacobhinkle (approved):

CMakeLists.txt
zasdfgbnm: BTW, because `-Wall` is already broken on gcc 12.2, we can continue the C++20 migration plan. Because the `-Wall` was the blocker, but now, `-Wall` is disabled due to a failure in C++17, we can just continue ignoring the warnings.

jacobhinkle: Reference: #3666

PR #3002, opened 2024-09-24
Compute strides for Hopper Mma

zasdfgbnm:
Currently, we use hard-coded `leading_bytes` and `stride_bytes` in the main branch. These hard-coded stride values make it only possible for us to use a very specific memory format with no flexibility. This also makes it hard to handle multiple tiles of operands in smem. This PR utilizes the new utility `lower_utils::proveLinearAndGetStride` to compute the stride, instead of hard-coding some value for them.

Hopefully, this is the last piece of Hopper Mma's indexing. With this in, we should be ready to freely schedule MMAs, single or multiple tiles per CTA, depending on the heuristics. However, this PR only tests the changes on existing unit tests, which are all single-tile tests that is compatible with the hard-coded value. I will add more single/multiple tile tests in future PRs.

### DIFF
--- a/csrc/device_lower/pass/index.cpp
+++ b/csrc/device_lower/pass/index.cpp
@@ -1804,6 +1804,37 @@ ValGroup getInnerMmaLoopGroup(TensorView* tv, const MmaOp* mma) {
   return inner;
 }
 
+// Compute the "leading_bytes" in the matrix descriptor of Mma. The leading
+// bytes is the stride of the innermost dimension in the allocation domain of tv
+// considering core matrices. For example, if the tv is [M, K], where K is the
+// inner, then the schedule of the loop domain of the mma output must be
+// something like:
+//      M            K
+//      |            |
+//     ...          ...
+//      |            |
+//     / \.         / \.
+//   ...  m_inst  ...  k_inst
+// where m_inst and k_inst are the instruction tiles of M and K, respectively,
+// that is, the number of items each TensorCore instruction can execute. Both
+// m_inst and k_inst must be parallelized on Mma. The leading_bytes is the
+// stride of the outer (k_inst/swizzle_size) in the allocation domain of tv.
+// That is, if we futher split k_inst as:
+//      M            K
+//      |            |
+//     ...          ...
+//      |            |
+//     / \.         / \.
+//   ...  m_inst  ...  k_inst
+//                     /    \.
+//               linear      swizzle_size
+// Then we would need to prove that `linear` is linear in the allocation domain
+// of tv, and the stride of `linear` is the leading_bytes. This function does
+// the following things:
+// 1. Find k_inst.
+// 2. Split k_inst as above.
+// 3. Prove that `linear` is linear in the allocation domain of tv, and get the
+//    stride of `linear`.
 Val* getInnerStride(TensorView* tv, const MmaOp* mma) {
   auto swizzle = getSwizzleMode(tv);
   auto swizzle_size = getBytesFromSwizzle(swizzle) / dataTypeSize(tv->dtype());
@@ -1823,6 +1854,37 @@ Val* getInnerStride(TensorView* tv, const MmaOp* mma) {
   return SimplifyingIrBuilder::mulExpr(stride, dataTypeSize(tv->dtype()));
 }
 
+// Compute the "stride_bytes" in the matrix descriptor of Mma. The stride
+// bytes is the stride of the outer dimension in the allocation domain of tv
+// considering core matrices. For example, if the tv is [M, K], where K is the
+// inner, then the schedule of the loop domain of the mma output must be
+// something like:
+//      M            K
+//      |            |
+//     ...          ...
+//      |            |
+//     / \.         / \.
+//   ...  m_inst  ...  k_inst
+// where m_inst and k_inst are the instruction tiles of M and K, respectively,
+// that is, the number of items each TensorCore instruction can execute. Both
+// m_inst and k_inst must be parallelized on Mma. The stride_bytes is the
+// stride of the outer (m_inst/8) in the allocation domain of tv.
+// That is, if we futher split m_inst as:
+//       M            K
+//       |            |
+//      ...          ...
+//       |            |
+//      / \.         / \.
+//    ...  m_inst  ...  k_inst
+//         /    \.
+//   linear      8
+// Then we would need to prove that `linear` is linear in the allocation domain
+// of tv, and the stride of `linear` is the stride_bytes. This function does
+// the following things:
+// 1. Find m_inst.
+// 2. Split m_inst as above.
+// 3. Prove that `linear` is linear in the allocation domain of tv, and get the
+//    stride of `linear`.
 Val* getOuterStride(TensorView* tv, const MmaOp* mma) {
   ValGraph& id_graph = GpuLower::current()->tensorIndexer().traversalGraph();
   auto logical_domain =
--- a/csrc/device_lower/pass/index.cpp
+++ b/csrc/device_lower/pass/index.cpp
@@ -1768,13 +1768,15 @@ ValGroup getInnerMmaLoopGroup(TensorView* tv, const MmaOp* mma) {
       "Matmul with all broadcasting dimension is not supported yet.");
 
   auto exprs =
-      ValGraphBFS::getExprsBetween(id_graph, alloc_domain, loop_domain);
-  for (const auto& [expr, direction] : exprs) {
+      ValGraphBFS::getExprsBetween(id_graph, loop_domain, alloc_domain);
+  while (!exprs.empty()) {
+    auto [expr, direction] = exprs.back();
+    exprs.pop_back();
     auto from =
-        (direction == Direction::Forward ? id_graph.inputGroups(expr)
+        (direction == Direction::Backward ? id_graph.inputGroups(expr)
                                          : id_graph.outputGroups(expr));
     auto to =
-        (direction == Direction::Forward ? id_graph.outputGroups(expr)
+        (direction == Direction::Backward ? id_graph.outputGroups(expr)
                                          : id_graph.inputGroups(expr));
     bool in_from = std::find(from.begin(), from.end(), inner) != from.end();
     if (!in_from) {
--- a/csrc/device_lower/pass/index.cpp
+++ b/csrc/device_lower/pass/index.cpp
@@ -1884,7 +1884,7 @@ Val* getOuterStride(TensorView* tv, const MmaOp* mma) {
     return false;
   };
   ValGroup selected = nullptr;
-  for (int64_t i = 0; i < mma_groups.size(); ++i) {
+  for (int64_t i = 0; i < (int64_t)mma_groups.size(); ++i) {
     if (is_projected_to_concrete(i)) {
       NVF_ERROR(
           selected == nullptr,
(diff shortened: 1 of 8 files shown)


zasdfgbnm:
!build

csrc/device_lower/pass/index.cpp
```diff
@@ -1742,10 +1743,240 @@ static MmaInputSmemSwizzle getSwizzleMode(TensorView* tv) {
...
+  // way to the consumer's loop domain, and keep track of the inner dimension.
+  // After propagating, the inner dimension should be a dimension that is
+  // parallelized on Mma.
+  ValGroup inner = nullptr;
+  for (auto it = alloc_domain.rbegin(); it != alloc_domain.rend(); ++it) {
+    if (!(*it)->front()->as<IterDomain>()->isBroadcast()) {
+      inner = *it;
+      break;
+    }
+  }
```
jacobhinkle, line 1766: At line 1752 you have already filtered out broadcasts. That doesn't mean there are no broadcasts in the ValGroup, but it does mean that this allocation domain ID is not Broadcast. So why do we need to skip ValGroups here whose first ID is broadcast?

csrc/device_lower/pass/index.cpp
```diff
@@ -1742,10 +1743,240 @@ static MmaInputSmemSwizzle getSwizzleMode(TensorView* tv) {
...
+  NVF_ERROR(
+      inner != nullptr,
+      "Matmul with all broadcasting dimension is not supported yet.");
+
+  auto exprs =
+      ValGraphBFS::getExprsBetween(id_graph, loop_domain, alloc_domain);
+  while (!exprs.empty()) {
+    auto [expr, direction] = exprs.back();
```
jacobhinkle, line 1774: Why not `for (auto& [expr, direction] : exprs) {` instead?

csrc/device_lower/pass/index.cpp
```diff
@@ -1742,10 +1743,240 @@ static MmaInputSmemSwizzle getSwizzleMode(TensorView* tv) {
...
+//                     /    \.
+//               linear      swizzle_size
+// Then we would need to prove that `linear` is linear in the allocation domain
+// of tv, and the stride of `linear` is the leading_bytes. This function does
+// the following things:
+// 1. Find k_inst.
+// 2. Split k_inst as above.
+// 3. Prove that `linear` is linear in the allocation domain of tv, and get the
+//    stride of `linear`.
+Val* getInnerStride(TensorView* tv, const MmaOp* mma) {
```
jacobhinkle, line 1838: Should we name this `getInnerStrideBytes`?

csrc/device_lower/pass/index.cpp
```diff
@@ -1742,10 +1743,240 @@ static MmaInputSmemSwizzle getSwizzleMode(TensorView* tv) {
...
+//         /    \.
+//   linear      8
+// Then we would need to prove that `linear` is linear in the allocation domain
+// of tv, and the stride of `linear` is the stride_bytes. This function does
+// the following things:
+// 1. Find m_inst.
+// 2. Split m_inst as above.
+// 3. Prove that `linear` is linear in the allocation domain of tv, and get the
+//    stride of `linear`.
+Val* getOuterStride(TensorView* tv, const MmaOp* mma) {
```
jacobhinkle, line 1888: Should we name this `getOuterStrideBytes`?

csrc/device_lower/pass/index.cpp
```diff
@@ -1742,10 +1743,240 @@ static MmaInputSmemSwizzle getSwizzleMode(TensorView* tv) {
...
+                                         : id_graph.inputGroups(expr));
+    for (const auto& g : from) {
+      for (auto& projection_on_logical : projections_on_logical) {
+        if (projection_on_logical.count(g)) {
+          projection_on_logical.erase(g);
+          projection_on_logical.insert(to.begin(), to.end());
+        }
+      }
+    }
+  }
```
jacobhinkle, line 1937: Projecting a subset of a domain onto another domain feels like a utility that would be useful elsewhere. Maybe consider making it a separate utility?

csrc/device_lower/pass/index.cpp
zasdfgbnm: Good catch. I wasn't filtering out broadcasts originally in 1752, so this was needed. I changed 1752 later, but forget to clean this up.

csrc/device_lower/pass/index.cpp
zasdfgbnm: This is because we need to traverse `exprs` in the reverse order. Note that we can not just swap the `loop_domain` with the `alloc_domain` in the above `ValGraphBFS::getExprsBetween`, because `alloc_domain` does not fully cover loop domain, and if we do so, BFS will fail.

csrc/device_lower/pass/index.cpp
zasdfgbnm: Good suggestion!

zasdfgbnm: Renamed

csrc/device_lower/pass/index.cpp
zasdfgbnm: Renamed

csrc/device_lower/pass/index.cpp
zasdfgbnm: Removed

csrc/device_lower/pass/index.cpp
zasdfgbnm: > > because `alloc_domain` does not fully cover loop domain, and if we do so, BFS will fail
> 
> Ahhh OK. That makes sense. It will be nice to move to C++20 for cases like this so we can just do
> 
> ```c++
> for (auto& [expr, direction] : exprs | std::views::reverse) {
> ```

100% agree

csrc/device_lower/pass/index.cpp
zasdfgbnm: changed

zasdfgbnm:
!build

csrc/device_lower/utils.h
```diff
@@ -375,6 +375,13 @@ Val* proveLinearAndGetStride(
     const ValGroup& linear_g,
     const ValGroups& domain);
 
+// Given `from`, project it to `to`. This function will return a subset of `to`
```
naoyam, line 378: nit: should this be in this util file? Why not `val_graph_visitor.h`?

csrc/device_lower/utils.cpp
```diff
@@ -1933,6 +1933,42 @@ Val* proveLinearAndGetStride(
   return proveLinearAndGetStrideAfterPropagation(frontier, domain);
 }
 
+std::unordered_set<ValGroup> projectTo(
```
naoyam, line 1936: It seems this is quite similar to `getReachableValsFrom`, which I recently added. 

https://github.com/NVIDIA/Fuser/blob/main/csrc/iter_visitor.h#L616

csrc/device_lower/pass/index.cpp
naoyam: Why not just `ValGraphBFS::getExprsBetween(id_graph, alloc_domain, loop_domain);`?

csrc/device_lower/utils.cpp
```diff
@@ -1933,6 +1933,42 @@ Val* proveLinearAndGetStride(
...
+  }
+  // Remove items that are not in `to`. This could happen if `from` is not
+  // connected to `to`.
+  for (auto it = projection.begin(); it != projection.end();) {
+    if (!to.has(*it)) {
+      it = projection.erase(it);
+    } else {
+      ++it;
+    }
+  }
```
jacobhinkle, line 1968: C++20 will make this nicer too:
```c++
std::erase_if(projection, [&to] (const ValGroup& vg) { return !to.has(vg); });
```

tests/cpp/utils.h
```diff
@@ -555,9 +555,9 @@ class NVFuserTest : public ::testing::Test {
 class HopperBase : public NVFuserTest {
  protected:
   void SetUp() override {
-    if (cudaArchGuardShouldSkip(9, 0)) {
-      GTEST_SKIP() << "skipping tests on pre-Hopper GPUs";
-    }
+    // if (cudaArchGuardShouldSkip(9, 0)) {
```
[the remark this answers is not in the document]
zasdfgbnm: .... I enabled this for the convenience of debugging on a non-Hopper machine, but forget to uncomment it. Will uncomment.

But.... Why there is no failure in CI?

csrc/device_lower/pass/index.cpp
zasdfgbnm: This does not work. Loop domain is bigger than the allocation domain and may contain IDs not in the allocation. This way, BFS will fail.

tests/cpp/test_mma.cpp
```diff
@@ -388,8 +388,10 @@ TEST_P(HopperRS, SingleTile) {
   {
     auto s = mma_utils::MmaSwizzler::scheduleMmaOutputAllocation(
         tv2c->getLoopDomain());
-    tv2c->setLoopDomain(s.as<IterDomain*>());
```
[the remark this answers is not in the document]
zasdfgbnm: As we previously discussed in [Representing ldmatrix](https://docs.google.com/document/d/1C7FJW_ySGgBLCYuZR2xM67SQJGIxlHwACpnSweImX78/edit?usp=sharing), we need both the read domain and write domain for MmaOp. And the read and write domain will not match. So, there is no correct answer of what the loop domain of the mma output should be unless we implement that document.

Previously, we are scheduling the loop domain as the write domain, which allows us to have multiple tiles in register (but we are not using this capability because we always just have one tile in all unit tests). But because this PR starts to infer instead of hard-code the strides, we really need to have a correct read domain, so I decide to change the loop domain from write domain to read domain. With this change, we will be able to support multiple tiles in smem, but not in registers. But for now, we can always make sure there is only one tile in register by inlining most.

csrc/device_lower/utils.cpp
zasdfgbnm: Let me try if I can just use `getReachableValsFrom`, if yes, I will remove this utility, if not, I will move it to `val_graph_visitor.h`.

tests/cpp/utils.h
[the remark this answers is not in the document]
naoyam: It actually seems the A100 binary tests did fail. It's just not shown in the github page.

csrc/device_lower/utils.cpp
zasdfgbnm: `getReachableValsFrom` worked. Removed `projectTo`.

csrc/device_lower/utils.h
zasdfgbnm: This function is removed

naoyam (approved):
LGTM

tests/cpp/test_mma.cpp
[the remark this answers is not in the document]
zasdfgbnm: Added

zasdfgbnm:
!build

csrc/device_lower/utils.cpp
zasdfgbnm: cc: @naoyam Looks like `getReachableValsFrom` does not work for multi-tile tests.

naoyam: Do you know why?

zasdfgbnm: Didn't look deep, but I believe the only difference between `getReachableValsFrom` and `projectTo` is the traversal direction. I believe generally it is most safe if `from` fully covers `to`. If not, `require_all_to_visited=true` may make some case work, but seems that it is not guaranteed.

PR #3003, opened 2024-09-24
Support non-conventional loop domains in TensorView::updateMaxProducerPosition.

naoyam:
Follow up to #2983.

`TensorView::updateMaxProducerPosition` is used to set the max producer position when `inlineAt` is used. With this PR, `inlineMost` should work with `setLoopDomain`.

This should be the last remaining piece to enable inlineMost with loop domains set by `setLoopDomain`. There are a couple more issues to address in lowering, though.


naoyam:
!build

naoyam:
!build

naoyam:
!build

tests/cpp/test_inlining.cpp
```diff
@@ -591,4 +590,192 @@ TEST_F(InliningTest, GetMaxPosAllNormalization) {
...
+  auto tv2 = set(tv1);
+  auto tv3 = broadcast(tv2, {false, true});
+  auto tv4 = add(tv0, tv3);
+  fusion.addOutput(tv4);
+
+  {
+    std::vector<IterDomain*> loop_domain{
+        tv2->getLogicalDomain().at(0),
+        tv3->getLogicalDomain().at(1)->cloneWithoutRFactor(true)};
+    tv2->setLoopDomain(loop_domain);
```
zasdfgbnm, line 613: So in the loop domain, tv2 has a new broadcast mapped to tv3's broadcast. Is this intentional? Isn't the plan to only have concrete IDs in the loop domain?

naoyam: Yes, I originally thought we could just have concrete domains only, but I realized using broadcast domains could save allocation sizes in some cases.

For example, let's say we have a fusion like:

```
t0 = [i0, i1]
t1 = [i0]

t2 = broadcast(t1, {false, true}); // [i0, b2]
t3 = t0 + t1
```

Suppose t1 gets the loop domain of t3 (or t0), i.e., `[i0, i1]`. Suppose `t1->split(1, 4); t1->inlineAt(1);`, i.e., `[i0, i1/4, 4] ca_pos(1)`. Its allocation size would be `i1/4 * 4`.

Suppose t1 gets the loop domain of t2, i.e., `[i0, b2]`. Suppose the same schedule and inlining, i.e., `[i0, b2/4, 4] ca_pos(1)`. Its allocation size would be just 1 as the two innermost dimensions are still just broadcast.

So, generally, I think missing domains should always be broadcast domains, and instead of allowing `setLoopDomain` to have arbitrary extra domains, it may make more sense to only allow `TensorDomain::broadcast`.

zasdfgbnm: I see. But doesn't this depend on how we "infer" allocation domain? I think for this case, we do want to allocate 1, instead of `i1/4*4`, and the question is mostly "how?". A very simple algorithm for inferring allocation domain is: the allocation domain is the part of loop domain that is connected to logical domain. Using this algorithm, because non of the `i1/4` or `4` is connected to logical domain, we will just ignore it, and we will have allocation domain `[]` (`i0` is not included due to compute at), which is allocation size 1.

I personally do not have a strong preference on whether we should only allowing adding broadcasts, and continue "inferring" allocation domain naively, or we should make loop domain faithfully represent the extent of the loop and inferring allocation domain more smartly. I slightly prefer the latter, but not strong. I do have a strong opinion that we should remove permissive graph and only keeping broadcast graph, but I believe this is unrelated here.

zasdfgbnm (approved):

tests/cpp/test_inlining.cpp
naoyam: > the question is mostly "how?"

Right, we don't need to have the allocation domain the same as the loop domain, but that's the most straightforward way right now, because it doesn't require any analysis of parallelization and computeAt position from the loop domain to the allocation domain. That's the only reason I took this approach.

PR #3010, opened 2024-09-24
Short-circut magic zero pass completely when not needed

zasdfgbnm:
I encountered an indexing error in `protectPredicateIndexWithMagicZero` in one of my matmul unit tests. Matmul doesn't use magic zero, so just completely short-circuit it when not needed.


zasdfgbnm:
!build

naoyam (approved):
LGTM.

Curious what error you had. `TensorIndexer` doesn't implement magic zero yet, so I suppose you shouldn't hit this path as long as `TensorIndexer` is used.

zasdfgbnm:
> LGTM.
> 
> Curious what error you had. `TensorIndexer` doesn't implement magic zero yet, so I suppose you shouldn't hit this path as long as `TensorIndexer` is used.

I think it is failing in a tensor that is not using tensor indexer. For matmul, we are only using tensor indexer for specific tensors. The error message is something like "unable to get index for X", where X is some broadcast ID.

PR #3016, opened 2024-09-25
Return SchedulerType::None in proposeHeuristics and tryMerge instead of optional.

csarofeen:
Just a nit as we have `SchedulerType::None` we can use that to signal a scheduler was not found for a group in tryMerge and proposeHeuristic.


csarofeen:
!build

PR #3017, opened 2024-09-25
Add option to configure the install directory in `setup.py`

samnordmann:
# What

Add in `setup.py` an option `install-dir=<absolute path>` to specify the install directory. If not provided, the install dir defaults to `./nvfuser`, thus the current default behavior remains unchanged.

# Why

- Useful option in itself
- My use case: I develop by  mounting the source repo inside a docker. Since NFS is very slow, I want to restrict the write/read to the mounted repo to the bare minimum, i.e., reading the source files. I am therefore already building fuser on a local mount, using the option `--build-dir`. But I found out that changing also the install dir speeds up further the installation
- I hope this could help reduce the NFS usage, as [was requested](https://nvidia.slack.com/archives/C8JCM3DPY/p1727281989644789?thread_ts=1727238691.063249&cid=C8JCM3DPY) to @cowanmeg and myself

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,6 +38,9 @@
 #   --build-dir=<ABSOLUTE PATH>
 #     Specify in which directory to build nvfuser. If not specified, the default build directory is "./build".
 #
+#   --install-dir=<ABSOLUTE PATH>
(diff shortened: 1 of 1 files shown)


samnordmann:
!build

xwang233 (approved):

PR #3018, opened 2024-09-25
[WIP] Accept multi-matmul fusions for automatic scheduling

jacobhinkle:
Stacked on #2458

This updates the Matmul scheduler to accept fusions that have multiple matmul patterns if they pass a topology check. These conditions are checked:
- No matmul pattern can depend on the output of another matmul pattern
- Dimension roles must be compatible among all matmul patterns. That means a dimension must not appear as an M dimension in one pattern and an N dimension in another pattern, for example.
- The above condition implies that tensor roles are also compatible: an A operand in one pattern cannot appear as a B operand in another pattern for example.

Each `mma_utils::MatmulPattern` is mapped to its global memory operand inputs and patterns that share at least one of these operands are collected together. These collections will be computed together in a "main loop"; these are described by `mma_utils::MainLoopDescriptor` which holds the number of A  and B operands as well as the number of matmul patterns computed in that loop i.e. the number of register result buffers needed. The fusion topology is then described by a `mma_utils::MatmulFusionTopology` struct which just holds a vector of main loop descriptors.

This PR introduces those types along with `mma_utils::computeMatmulFusionTopology()`, and runs that in `getMatmulCompileTimeRejectReason`. Additionally, it updates `getMatmulHeuristics` to accomodate the multiple matmul case such that we do not run out of smem or register space.

### TODO
- Finish updating the matmul heuristics
- Update `mma_utils::generateSharedMemoryEpilogueHeuristics` to account for topology
- Update `MatmulParams` to accomodate different vectorization sizes for each operand in each main loop
- Plumb changes to `MatmulParams` to heuristic plugin so that we can make adjustments in the heuristic plugin instead of the post hoc approach here.
- Add more tests. Currently I'm just looking at the Llama2FFN test.

### DIFF
--- a/csrc/scheduler/matmul_utils.cpp
+++ b/csrc/scheduler/matmul_utils.cpp
@@ -907,6 +907,56 @@ std::unique_ptr<MatmulParams> getMatmulHeuristics(
     NVF_ERROR(status, "Initialization of core part of heuristics failed.");
   }
 
+  if (patterns.size() > 2) {
+    // Currently, both the default heuristic and plugins cannot account for
+    // multiple matmuls being present in a single fusion, so they assume there
+    // is only one. This can lead to errors due to register and shared memory
+    // limitation unless we adjust the tile sizes.
+    // TODO: provide number of matmuls and basic topology to heuristic instead
+
+    // [Adjusting CTA tile to account for multiple MmaOps]
+    // If we have nm mma operations na A operands and nb B operands the smem
+    // usage of the operand loads is:
+    //   (na*(cm*ck) + nb*(cn*ck))*num_stages
+    // where cm, cn, ck is the size of the CTA tile.
+    // If we use smem for the epilogue unswizzling, we have the following total
+    // smem use (assuming smem reuse):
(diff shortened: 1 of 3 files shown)


tests/cpp/test_matmul_scheduler.cpp
```diff
@@ -2689,11 +2689,16 @@ TEST_P(MatmulFusionTest, Llama2FFN) {
   const FusionKernelRuntime* runtime =
       executor_cache.getMostRecentKernelRuntime();
 
-  EXPECT_TRUE(runtime->isSegmented());
-
   if (fusion_enabled) {
-    EXPECT_EQ(runtime->fusionSegments()->groups().size(), 2);
+    if (isOptionEnabled(EnableOption::FuseMultipleMatmuls)) {
```
jacobhinkle, line 2693: Running `NVFUSER_ENABLE=fuse_multiple_matmuls test_matmul --gtest_filter=MatmulFusionTest.Llama2FFN/fuse` will result in a kernel similar to that of `test_matmul --gtest_filter=MatmulSchedulerTest.Llama2FFNNoSegmentation`

PR #3025, opened 2024-09-26
Enable segmentation support to python frontend user-scheduling

rdspring1:
## Overview:
The original `FusionDefinition` stores the sequence of sub-fusions and acts as an argument manager. It gathers the input arguments before running the sub-fusion and stores its results. To perform this function, it uses a map from the segment index space to the original index space. This mapping was generated while creating the python definition for each sub-fusion.

## Changes in this PR

This PR implements `_execute_segments ` function for user-scheduler segmentation. It is the third PR in a stack, preceded by https://github.com/NVIDIA/Fuser/pull/3334 and https://github.com/NVIDIA/Fuser/pull/3335.

1. Implement `_execute_segments ` function in `nvfuser/__init__.py` to orchestrate segments in original fusion.
2. Add `supports_segmentation flag` to `exec_nvfuser`, so segmentation testing is enabled by default for all python tests.

```python
                          contiguity=[True, True],
                          dtype=DataType.Float,
                          is_cpu=False)
                          contiguity=[True, True],
                          dtype=DataType.Float,
                          is_cpu=False)
    T4 = fd.ops.add(T1, T3)
    fd.add_output(T4)
```

#### `map_original_fid_to_value` state: 6 entries
| Original Index | Description |
| -----------------| --------------- |
| 0 | The first tensor argument for the original fusion. |
| 1 | The second tensor argument for the original fusion. |
| -1 | Extent of axis 0 for first tensor argument. |
| -2 | Extent of axis 1 for first tensor argument. |
| -3 | Extent of axis 0 for second tensor argument. |
| -4 | Extent of axis 1 for second tensor argument. |

* Omit extents [-1, -4] from table in future steps because they are not necessary for these segments.

#### `map_original_fid_to_value` state: 6 entries
| Original Index | Description |
| -----------------| --------------- |
| 1 | The second tensor argument for the original fusion. |
| 3 | The broadcasted, reduction tensor, which is the output from the first segment. |

*  Removed the entry for `T0` because the first tensor argument is not required for second segment.
*  Added the entry for `T3`, which is the output from the first segment.

#### `map_original_fid_to_value` state: 5 entries
| Original Index | Description |
| -----------------| --------------- |
| 4 | The pointwise addition, which is the output for the original fusion. |

*  Removed the entries for `T1` and `T3` because they are not necessary anymore.
*  Added the entry for `T4`, which is the output from the second segment.

### Step 4 after running all segments.
* Return `T4` from `map_original_fid_to_value` as the result for the original fusion.


rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!test

rdspring1:
!test

rdspring1:
!test

tests/python/utils.py
```diff
@@ -242,13 +242,24 @@ def check_captured_python_definition(reference_outputs, fd, inputs, device=None)
...
     try:
         torch.manual_seed(0)
+
+        # Clone
         cloned_fd = FusionDefinition()
         clone(fd, cloned_fd)
+
+        # Segment
+        if supports_segmentation:
+            cloned_fd.segment(inputs)
```
jjsjann123, line 260: what's the expected workflow in the finished form?

Looks like `segment` should always be called and check if segmentation happens? hint: maybe we want to name it as `_segment` so it's not intended to be exposed as a user API?

jjsjann123: I'm wondering if we should also consider having `supports_segmentation` as an optional kwargs in `FusionDefinition.execute`

rdspring1:
!test

tests/python/utils.py
jjsjann123: > Looks like segment should always be called and check if segmentation happens? hint: maybe we want to name it as _segment so it's not intended to be exposed as a user API?

btw, I'm not saying that we should refactor that in this PR.

I understand that the sole purpose of segment usage is for debugging. I think we can simplify that and have it done automatically. We can discuss this later maybe.

rdspring1:
```python
  # define fusion
fd.segment(inputs)  # manually segment fusion
fd.execute(inputs) # run fusion
```
This ^^^ is the current workflow. The user calls `segment` and runs `execute`.

rdspring1: Your suggestion is:
```python
  # define fusion
fd.execute(inputs, segment_fusion=True) # run `fusion` and `segment` in `execute`
```

jjsjann123: Sorta.

My question is, when would we expect someone to use `segment_fusion=True`? If the expectation is just for getting a smaller repro, we can automate this. i.e. user don't need to enable python segmentation when they execute, we can do that in the try/catch block and minimize the repro there.

But if we plan to have people eagerly segment and play with scheduling for each segments, that's a different story.

rdspring1: It is the latter for me. I'd like to replace badly segmented fusions with a manually crafted solution. Other ideas is to prototype new segmentation algorithms in python.

rdspring1:
!test

csrc/python_frontend/segmentation.cpp
```diff
@@ -35,31 +35,52 @@ int64_t SegmentationState::setupSegmentation(
   IrCloner original_to_cloned_map =
       Fusion::copy(fusion, cloned_original_fusion_.get());
 
+  // Step 2) Given the map_presched_value_to_original_python_index AND the
```
rdspring1, line 38: Originally, this step mapped from `prescheduled_fusion` to `cloned_concretized_fusion`, which causes issue with DEBUG build.

The fix is to separate the step into a transform from `prescheduled_fusion` to `cloned_fusion`, apply concretization, then transform from `cloned_fusion` to `cloned_concretized_fusion`.

rdspring1:
!test

rdspring1:
!test

csrc/python_frontend/segmentation.cpp
jjsjann123: QQ: what's the issue you see on DEBUG build? Are we accessing dangling pointers or something?

jjsjann123: i.e. if there's some check that we only enabled in DEBUG build, we might want to also include that in our CI tests.

jjsjann123 (approved):
LGTM.

csrc/python_frontend/segmentation.cpp
rdspring1: @xwang233 I believe we run the DEBUG build in the CI tests. The CI found the failure for me.

It found an issue through `as<Val>()` where dynamic cast is used.

rdspring1:
!test

csrc/python_frontend/segmentation.cpp
xwang233: we don't run DEBUG build but we run `--explicit-error-check` build and all tests are executed on builds with explicit error check

jjsjann123: nvm, we are all good then. Thanks for bringing this up @rdspring1

rdspring1:
!test

PR #3028, opened 2024-09-26
Indexing test cleanup

naoyam:
Renamed test_gpu_indexing.cpp to test_indexing_advanced.cpp. Changed the tests to exercise both the legacy and new indexers.

Added several tests originally developed for IdModel (#32). Some of them are disabled as they are not yet supported.

### DIFF
--- a/csrc/device_lower/lower2device.cpp
+++ b/csrc/device_lower/lower2device.cpp
@@ -417,21 +417,11 @@ void GpuLower::analysis(Fusion* fusion) {
   // names
   if (this->requiresIdModel() || isOptionEnabled(EnableOption::IdModel)) {
     // Enable validation in the DEBUG build mode
-#ifdef NDEBUG
-    // Not DEBUG build
     id_model_ = std::make_unique<IdModel>(
         fusion_,
         /*build_graphs=*/true,
         /*allow_self_mapping=*/false,
         /*validate=*/false);
-#else
-    // DEBUG build
-    id_model_ = std::make_unique<IdModel>(
-        fusion_,
-        /*build_graphs=*/true,
-        /*allow_self_mapping=*/false,
-        /*validate=*/true);
-#endif
     id_model_->validateAndPropagatePType();
   }
 
--- a/tests/cpp/test_gpu3.cpp
+++ b/tests/cpp/test_gpu3.cpp
@@ -3765,39 +3765,6 @@ TEST_F(NVFuserTest, FusionScheduleTransposeRepro1_CUDA) {
   testValidate(&fusion, cg_outputs, {input0, input1}, __LINE__, __FILE__);
 }
 
-// Repro for issue #1873
-TEST_F(NVFuserTest, FusionInlineBroadcastIndexing0_CUDA) {
-  Fusion fusion;
-  FusionGuard fg(&fusion);
-
-  auto tv0 = makeContigTensor(1);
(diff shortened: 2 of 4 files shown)


tests/cpp/test_gpu3.cpp
```diff
@@ -3765,39 +3765,6 @@ TEST_F(NVFuserTest, FusionScheduleTransposeRepro1_CUDA) {
   testValidate(&fusion, cg_outputs, {input0, input1}, __LINE__, __FILE__);
 }
 
-// Repro for issue #1873
```
naoyam, line 3768: Moved to test_indexing_advanced.cpp

csrc/device_lower/lower2device.cpp
```diff
@@ -417,21 +417,11 @@ void GpuLower::analysis(Fusion* fusion) {
...
   if (this->requiresIdModel() || isOptionEnabled(EnableOption::IdModel)) {
     // Enable validation in the DEBUG build mode
-#ifdef NDEBUG
-    // Not DEBUG build
     id_model_ = std::make_unique<IdModel>(
         fusion_,
         /*build_graphs=*/true,
         /*allow_self_mapping=*/false,
         /*validate=*/false);
-#else
```
naoyam, line 427: Disabled the validation with ComputeAtMap since we now deal with non-conventional graph structures, so `ComputeAtMap` does not always work.

naoyam:
!build

tests/cpp/test_indexing_advanced.cpp
```diff
@@ -753,4 +816,326 @@ TEST_F(NVFuserTest, FusionIndexing18_CUDA) {
...
+  fe.compileFusion(&fusion, inputs);
+  auto outputs = fe.runFusion(inputs);
+
+  testValidate(&fusion, outputs, inputs, __LINE__, __FILE__);
+}
+
+// Create a case where we're missing a valid concrete id so the compute at map
+// processing will fail. We need to be able to create the concrete ID not just
+// look for one.
+TEST_F(AdvancedIndexingIdModelTest, 20) {
```
jacobhinkle, line 864: This example seems similar to this example we have looked at previously: https://github.com/NVIDIA/Fuser/pull/2512/files#diff-1e72aab35f54e7aaec2696293a478093301cc0a000d77fba8975cc8571740b84R731-R736. I am wondering why it fails, i.e. which dimension is missing a concrete ID?

jacobhinkle (approved):
LGTM. Just one question about the test that's failing ComputeAtMap

tests/cpp/test_indexing_advanced.cpp
naoyam: This fusion has two paths from `tv1`. One is `{tv2, tv3, tv4, tv5`}, and another is `{tv6, tv7, tv8, tv9}`. The first group has `[7, 11, 1]`, while the second has `[7, 1, 13]`. These two groups are inlined together at line 906, so the inlined loops need `[7, 11, 13]`, however, none of the inlined tensors has the complete shape, and `ComputeAtMap` just fails as it can't find a valid concrete domain (at least, it isn't a silent failure).

IdModel's loop promotion can handle this inlining pattern as it automatically create a valid concrete domain. However, since we haven't completely replaced ComputeAtMap with IdModel, the lowering doesn't work yet.

PR #3034, opened 2024-09-26
Use more detailed broadcast-squeeze analysis in preseg pass

### LINKED ISSUE #3027: Dynamic allocations in local memory error
When running the following snippet, nvFuser fails with error:

```shell
INTERNAL ASSERT FAILED at "/opt/pytorch/nvfuser/csrc/fusion_executor/executor.cpp":427, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. Allocations must be based on constant integers for local memory. However, found: T35_l_float[ bblockIdx.x417{( ceilDiv(( ceilDiv(( ceilDiv(1, 4) ), blockDim.x) ), 1) )}, bthreadIdx.x416{blockDim.x}, iS423{( ceilDiv(( ceilDiv(( ceilDiv(4096, blockDim.y) ), 2) ), 1) )}, bUS418{1}, bS414{4}, ithreadIdx.y420{blockDim.y}, iUS424{1}, iS422{2}, bS97{1} ] produce_pos( 8 ),  have dynamic allocations but are placed in local memory.
```

Repro snippet:

```python
# CUDA devices:
#  0: NVIDIA RTX 6000 Ada Generation
# torch version: 2.4.0+cu121
# cuda version: 12.1
# nvfuser version: 0.2.13+git9d66566
import torch

    T0 = fd.define_tensor(shape=[1, 4096, 4096], contiguity=[None, True, True], dtype=DataType.BFloat16, is_cpu=False, stride_order=[2, 1, 0])
    T1 = fd.define_tensor(shape=[1, 4096, 4096], contiguity=[None, True, True], dtype=DataType.BFloat16, is_cpu=False, stride_order=[2, 1, 0])
    T2 = fd.define_tensor(shape=[4096], contiguity=[True], dtype=DataType.BFloat16, is_cpu=False, stride_order=[0])
    T3 = fd.define_tensor(shape=[1, 4096, 4096], contiguity=[None, True, True], dtype=DataType.BFloat16, is_cpu=False, stride_order=[2, 1, 0])
    T4 = fd.ops.cast(T0, dtype=DataType.Float)
    T5 = fd.ops.cast(T1, dtype=DataType.Float)
    T6 = fd.ops.add(T4, T5)
    T7 = fd.ops.mul(T6, T6)
    T8 = fd.ops.sum(T7, dims=[2], keepdim=False, dtype=DataType.Null)
    T13 = fd.ops.broadcast_in_dim(T8, shape=[1, 4096, 1], broadcast_dims=[0, 1])
    S14 = fd.define_scalar(4096.00, dtype=DataType.Double)
    S15 = fd.ops.reciprocal(S14)
    T16 = fd.ops.mul(T13, S15)
    S17 = fd.define_scalar(1.00000e-05, dtype=DataType.Double)
    T18 = fd.ops.add(T16, S17)
    T19 = fd.ops.rsqrt(T18)
    S20 = fd.define_scalar(3.00000, dtype=DataType.Double)
    T21 = fd.ops.pow(T19, S20)
    T26 = fd.ops.broadcast_in_dim(T19, shape=[1, 4096, 4096], broadcast_dims=[0, 1, 2])
    T27 = fd.ops.mul(T6, T26)
    T32 = fd.ops.broadcast_in_dim(T2, shape=[1, 4096, 4096], broadcast_dims=[2])
    T33 = fd.ops.cast(T32, dtype=DataType.Float)
    T34 = fd.ops.cast(T3, dtype=DataType.Float)
    T35 = fd.ops.mul(T33, T34)
    T36 = fd.ops.mul(T27, T34)
    T37 = fd.ops.sum(T36, dims=[0, 1], keepdim=False, dtype=DataType.Null)
    T38 = fd.ops.cast(T37, dtype=DataType.BFloat16)
    T39 = fd.ops.mul(T26, T35)
    T40 = fd.ops.mul(T6, T35)
    T41 = fd.ops.sum(T40, dims=[0, 2], keepdim=False, dtype=DataType.Null)
    T46 = fd.ops.broadcast_in_dim(T41, shape=[1, 4096, 1], broadcast_dims=[1])
    S47 = fd.define_scalar(-0.500000, dtype=DataType.Double)
    T48 = fd.ops.mul(S47, T46)
    T49 = fd.ops.mul(T48, T21)
    S50 = fd.define_scalar(4096.00, dtype=DataType.Double)
    S51 = fd.ops.reciprocal(S50)
    T52 = fd.ops.mul(T49, S51)
    T53 = fd.ops.sum(T52, dims=[0, 2], keepdim=False, dtype=DataType.Null)
    T57 = fd.ops.broadcast_in_dim(T53, shape=[1, 4096], broadcast_dims=[1])
    T62 = fd.ops.broadcast_in_dim(T57, shape=[1, 4096, 1], broadcast_dims=[0, 1])
    T67 = fd.ops.broadcast_in_dim(T62, shape=[1, 4096, 4096], broadcast_dims=[0, 1, 2])
    T68 = fd.ops.mul(T6, T67)
    T69 = fd.ops.add(T39, T68)
    T70 = fd.ops.add(T69, T68)
    T71 = fd.ops.cast(T70, dtype=DataType.BFloat16)
    fd.add_output(T38)
    fd.add_output(T71)

    nvfuser_fusion_id0(fd)

inputs = [
    torch.randn(16777216, dtype=torch.bfloat16, device='cuda:0').as_strided((1, 4096, 4096), (16777216, 4096, 1)),
    torch.randn(16777216, dtype=torch.bfloat16, device='cuda:0').as_strided((1, 4096, 4096), (16777216, 4096, 1)),
    torch.randn(4096, dtype=torch.bfloat16, device='cuda:0').as_strided((4096,), (1,)),
    torch.randn(16777216, dtype=torch.bfloat16, device='cuda:0').as_strided((1, 4096, 4096), (16777216, 4096, 1)),
]
fd.execute(inputs)
```

jacobhinkle:
Our current broadcast+squeeze simplification pass (introduced in #2291) only does replacements if the broadcast flags exactly match the squeeze flags for neighboring `BroadcastOp` and `SqueezeOp` IR nodes. However, there are additional simplifications that can be made, and missing these can interfere with ID mapping leading to issues with inlining, for example: see comment here: https://github.com/NVIDIA/Fuser/issues/3027#issuecomment-2377501846.

This PR enables us to do simplifications where only some of the axes are bcast+squeezed. It also lets us remove intermediate `set` ops that could otherwise interfere with this analysis.

Fixes #3027

### Limitation with respect to expand

When we use `broadcast_in_dim` with new sizes other than 1, we insert two operations: a `BroadcastOp` and an `ExpandOp`. If we use that repeatedly we can have broadcast-expand-broadcast-expand patterns. We currently cannot collapse that into a single broadcast-expand, so if there were a squeeze preceding the first broadcast, we would not be able to simplify it:
```c++
fusion->addInput(tv0); // [ iS0{i0}  bS1{1} ]
auto tv1 = squeeze(tv0); // [ iS2{i0} ]
auto tv2 = broadcast(tv1, {true, false }); // [ bS3{1} iS4{i0} ]
auto tv3 = expand(tv2, {i1, i0}); // [ bS5{i1} iS6{i0} ]
auto tv4 = broadcast(tv3, {false, false, true}); // [ bS7{i1} iS8{i0} bS9{1} ]
auto tv5 = expand(tv4, {i1, i0, i2}); // [ bS10{i1} iS11{i0} bS12{i2} ]
```
Ideally we would be able to rearrange the expand-broadcast-expand pattern so that we could combine the two broadcasts, which would let us get rid of the squeeze. We can't currently do this because this pass still ignores expand, but it might be a useful optimization for a future PR.

### DIFF
--- a/csrc/preseg_passes/remove_bcast_squeeze.cpp
+++ b/csrc/preseg_passes/remove_bcast_squeeze.cpp
@@ -163,7 +163,7 @@ AxisOps composeOps(const AxisOps& prev, const AxisOps& next) {
     // to if it's not a broadcast. If it is a broadcast, then we'll unzip some
     // of these squeezes.
     while (next_pos < next.size() && next[next_pos] == AxisOp::BROADCAST) {
-      if (ops.back() == AxisOp::SQUEEZE) {
+      if (!ops.empty() && ops.back() == AxisOp::SQUEEZE) {
         ops.back() = AxisOp::PRESERVE;
       } else {
         ops.push_back(AxisOp::BROADCAST);
--- a/csrc/preseg_passes/remove_bcast_squeeze.cpp
+++ b/csrc/preseg_passes/remove_bcast_squeeze.cpp
@@ -8,6 +8,7 @@
 #include <debug.h>
 #include <ir/utils.h>
 #include <logical_domain_map.h>
+#include <multidevice/utils.h>
 #include <ops/alias.h>
 #include <ops/arith.h>
 #include <options.h>
@@ -60,22 +61,6 @@ AxisOps squeezeToAxisOps(SqueezeOp* squeeze) {
   return ops;
 }
 
-//! Check whether the loop domains are the same size and have the same
-//! parallelization, ignoring reduction dimensions.
-bool hasCompatibleParallelization(TensorView* orig_tv, TensorView* new_tv) {
-  std::vector<IterDomain*> orig_domain =
-      TensorDomain::noReductions(orig_tv->getLoopDomain());
-  std::vector<IterDomain*> new_domain =
-      TensorDomain::noReductions(new_tv->getLoopDomain());
-  NVF_ERROR(orig_domain.size() == new_domain.size());
-  for (size_t i : c10::irange(orig_domain.size())) {
-    if (orig_domain[i]->getParallelType() != new_domain[i]->getParallelType()) {
-      return false;
-    }
-  }
-  return true;
-}
-
 //! Checks whether this is a simple Set of a TensorView. If not, then this might
 //! represent a scalar set, or a segment_set.
 bool isSimpleTVSet(Expr* expr) {
@@ -88,15 +73,16 @@ bool isSimpleTVSet(Expr* expr) {
   return ldst->opType() == LoadStoreOpType::Set && in_tv != nullptr &&
       out_tv != nullptr
       // The hasRoot() check is to prevent picking up Set.Permute ops here
-      && !ldst->out()->as<TensorView>()->hasRoot()
-      // A set operation that changes parallelization is not considered trivial
-      && hasCompatibleParallelization(in_tv, out_tv);
+      && !ldst->out()->as<TensorView>()->hasRoot();
 }
 
 //! This defines the types of operations that are eligible for simplification in
 //! this pass.
 bool isReplaceableExpr(Expr* expr) {
   if (expr == nullptr) {
+    return false;
+  }
+  if (isResharding(expr)) {
     return false;
   }
   return expr->isA<BroadcastOp>() || expr->isA<SqueezeOp>() ||
@@ -248,13 +234,6 @@ TensorView* maybeDoReplacement(TensorView* orig) {
       case AxisOp::PRESERVE:
         // This is equivalent to a set Op
         replacement = input_tv;
-        // Check that parallelization is consistent for replacement. We do not
-        // want to alter the parallelization of input_tv here; we only
-        // parallelize _new_ TVs. If the parallelization is inconsistent, we
-        // simply refuse to make the replacement.
-        if (!hasCompatibleParallelization(orig, replacement)) {
-          return orig;
-        }
         break;
       case AxisOp::SQUEEZE:
         replacement = squeeze(input_tv, nonPreservedDims(simplified_ops));
--- a/csrc/preseg_passes/remove_bcast_squeeze.cpp
+++ b/csrc/preseg_passes/remove_bcast_squeeze.cpp
@@ -205,10 +205,6 @@ TensorView* maybeDoReplacement(TensorView* orig) {
     return orig;
   }
 
-  std::cout << "Checking replaceable pair of exprs:" << std::endl;
-  std::cout << "  first=" << first->toString() << std::endl;
-  std::cout << "  second=" << second->toString() << std::endl;
-
   AxisOps first_ops = exprToAxisOps(first);
   AxisOps second_ops = exprToAxisOps(second);
   AxisOps simplified_ops = composeOps(first_ops, second_ops);
--- a/csrc/preseg_passes/remove_bcast_squeeze.cpp
+++ b/csrc/preseg_passes/remove_bcast_squeeze.cpp
@@ -168,11 +168,12 @@ std::vector<bool> nonPreservedDims(const AxisOps& ops) {
 //! descriptor of their composition
 AxisOps composeOps(const AxisOps& prev, const AxisOps& next) {
   size_t prev_pos = 0, next_pos = 0;
+  size_t prev_size = prev.size(), next_size = next.size();
   AxisOps ops;
   while (true) {
     // If there are previous squeezes, insert them since they don't affect next
     // position
-    while (prev_pos < prev.size() && prev[prev_pos] == AxisOp::SQUEEZE) {
+    while (prev_pos < prev_size && prev[prev_pos] == AxisOp::SQUEEZE) {
       ops.push_back(AxisOp::SQUEEZE);
       prev_pos++;
     }
@@ -180,7 +181,7 @@ AxisOps composeOps(const AxisOps& prev, const AxisOps& next) {
     // prev[prev_pos] now gives the provenance of the axis that next_pos points
     // to if it's not a broadcast. If it is a broadcast, then we'll unzip some
     // of these squeezes.
-    while (next_pos < next.size() && next[next_pos] == AxisOp::BROADCAST) {
+    while (next_pos < next_size && next[next_pos] == AxisOp::BROADCAST) {
       if (!ops.empty() && ops.back() == AxisOp::SQUEEZE) {
         // This is a "squeeze then broadcast" pattern
         ops.back() = AxisOp::PRESERVE;
@@ -190,9 +191,9 @@ AxisOps composeOps(const AxisOps& prev, const AxisOps& next) {
       next_pos++;
     }
 
-    if (prev_pos >= prev.size() || next_pos >= next.size()) {
+    if (prev_pos >= prev_size || next_pos >= next_size) {
       NVF_ERROR(
-          prev_pos >= prev.size() && next_pos >= next.size(),
+          prev_pos >= prev_size && next_pos >= next_size,
           "Failed to align some ops");
       break;
     }
--- a/csrc/preseg_passes/remove_bcast_squeeze.cpp
+++ b/csrc/preseg_passes/remove_bcast_squeeze.cpp
@@ -179,64 +179,58 @@ AxisOps composeOps(const AxisOps& prev, const AxisOps& next) {
   // produce an output axis.
   //
   // For example:
-  //
   //   prev = [ S S P B P P S ]
   //   next = [ B P P S P B B ]
   //
   // The input has dimension 6, the intermediate tensor has dimension 4, and the
-  // output has dimension 6. The composition is equivalent to [ S P P P B ]. We
-  // perform 4 main loop iterations corresponding to the four intermediate
+  // output has dimension 6. The composition is equivalent to [ S P P P P P B ].
+  // We perform 4 main loop iterations corresponding to the four intermediate
   // dimensions.
   //
   //   ops = []
   //   prev = [ (prev_pos)S S P B P P S ]
   //   next = [ (next_pos)B P P S P B B ]
   //
+  //   (run processUnalignedOps()): // one cancelled SQUEEZE becomes PRESERVE
+  //   ops = [ S P ]
+  //   prev = [ S S (prev_pos)P B P P S ]
+  //   next = [ B (next_pos)P P S P B B ]
+  //
   //   First main loop iteration:
-  //
-  //     (run processUnalignedOps()):
-  //
-  //     // one cancelled SQUEEZE becomes PRESERVE
-  //     ops = [ S P ]
-  //     prev = [ S S (prev_pos)P B P P S ]
-  //     next = [ B (next_pos)P P S P B B ]
-  //
   //     (process aligned ops "preserve then preserve")
   //     ops = [ S P P ]
   //     prev = [ S S P (prev_pos)B P P S ]
   //     next = [ B P (next_pos)P S P B B ]
   //
-  //   Second main loop iteration:
-  //
   //     (run processUnalignedOps()) // no change
   //
+  //   Second main loop iteration:
   //     (process aligned ops "broadcast then preserve")
   //     ops = [ S P P B ]
   //     prev = [ S S P B (prev_pos)P P S ]
   //     next = [ B P P (next_pos)S P B B ]
   //
-  //   Third main loop iteration:
   //     (run processUnalignedOps()) // no change
   //
+  //   Third main loop iteration:
   //     (process aligned ops "preserve then squeeze")
   //     // Note that we would normally insert SQUEEZE resulting in
-  //        ops = [ S P P B S ], but instead we remove the
+  //        ops = [ S P P B S ], but instead we change the
   //        previously-inserted BROADCAST to PRESERVE in pushOp(AxisOp::SQUEEZE)
-  //     ops = [ S P P ]
+  //     ops = [ S P P P ]
   //     prev = [ S S P B P (prev_pos)P S ]
   //     next = [ B P P S (next_pos)P B B ]
   //
-  //   Third main loop iteration:
   //     (run processUnalignedOps()) // no change
   //
+  //   Fourth main loop iteration:
   //     (process aligned ops "preserve then preserve")
-  //     ops = [ S P P P ]
+  //     ops = [ S P P P P ]
   //     prev = [ S S P B P P (prev_pos)S ]
   //     next = [ B P P S P (next_pos)B B ]
   //
-  //   Post-processing:
-  //     (run processUnalignedOps())
-  //     ops = [ S P P P B ]
+  //     (run processUnalignedOps()) // one cancelled SQUEEZE becomes PRESERVE
+  //     ops = [ S P P P P P B ]
   //     prev = [ S S P B P S (prev_pos) ]
   //     next = [ B P P S B B (next_pos) ]
   size_t prev_pos = 0, next_pos = 0;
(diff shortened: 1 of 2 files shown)


jacobhinkle:
I think this is a promising PR, but unfortunately it doesn't fully address #3027. We replace the broadcast+squeeze with a single squeeze:
```
T24_l_float[ bblockIdx.x462{( ceilDiv(( ceilDiv(( ceilDiv(1, 8) ), blockDim.x) ), 1) )}, bthreadIdx.x461{blockDim.x}, iS471{( ceilDiv(( ceilDiv(( ceilDiv(( ceilDiv(4096, blockDim.y) ), 2) ), 1) ), gridDim.y) )}, bUS463{1}, bS459{8}, iblockIdx.y470{gridDim.y}, ithreadIdx.y465{blockDim.y}, iUS469{1}, iS467{2}, bS68{1} ] produce_pos( 9 )
   = T23_l_float[ bblockIdx.x392{( ceilDiv(( ceilDiv(( ceilDiv(1, 8) ), blockDim.x) ), 1) )}, bthreadIdx.x391{blockDim.x}, iS401{( ceilDiv(( ceilDiv(( ceilDiv(( ceilDiv(4096, blockDim.y) ), 2) ), 1) ), gridDim.y) )}, bUS393{1}, bS389{8}, iblockIdx.y400{gridDim.y}, ithreadIdx.y395{blockDim.y}, iUS399{1}, iS397{2}, bS65{1} ] ca_pos( 9 ) produce_pos( 9 )
   * d102;
T37_l_float[ iS737{( ceilDiv(( ceilDiv(( ceilDiv(( ceilDiv(4096, blockDim.y) ), 2) ), 1) ), gridDim.y) )}, iblockIdx.y736{gridDim.y}, ithreadIdx.y731{blockDim.y}, iUS735{1}, iS733{2}, bS103{1} ] ca_pos( 5 )
   = squeeze( T24_l_float[ bblockIdx.x462{( ceilDiv(( ceilDiv(( ceilDiv(1, 8) ), blockDim.x) ), 1) )}, bthreadIdx.x461{blockDim.x}, iS471{( ceilDiv(( ceilDiv(( ceilDiv(( ceilDiv(4096, blockDim.y) ), 2) ), 1) ), gridDim.y) )}, bUS463{1}, bS459{8}, iblockIdx.y470{gridDim.y}, ithreadIdx.y465{blockDim.y}, iUS469{1}, iS467{2}, bS68{1} ] produce_pos( 9 ) )
```
As we can see, T37 is not inlined still...

jacobhinkle:
> We replace the broadcast+squeeze with a single squeeze:

As @liqiangxl pointed out, this is not sufficient in this case. There is an additional Broadcast after the squeeze and we are missing that combination, so I think the key to fixing #3027 is still to clean up this PR so that our bcast+squeeze combination pass is as powerful as possible.

jacobhinkle:
!build --diff

csrc/preseg_passes/remove_bcast_squeeze.cpp
```diff
@@ -8,45 +8,242 @@
...
-    if (auto bcast = dynamic_cast<BroadcastOp*>(expr)) {
-      if (auto squeeze = dynamic_cast<SqueezeOp*>(bcast->in()->definition())) {
-        if (bcast->getBroadcastDimFlags() == squeeze->getSqueezeDimFlags()) {
-          ir_utils::replaceValInAllExprInputsAndFusionOutputs(
-              bcast->out(), squeeze->in());
-        }
+    if (isReplaceableExpr(expr)) {
+      Expr* prev_expr = expr->input(0)->definition();
+      if (isReplaceableExpr(prev_expr)) {
+        maybeDoReplacement(prev_expr, expr);
```
jacobhinkle, line 246: I don't like this recursion, but I wanted to check that this fixes our issue. It is needed, but I'd rather just stop using `fusion->exprs()` at the level up and instead manage our own stack-based DFS from the outputs at that level.

jacobhinkle:
!build --diff

jacobhinkle:
!build --diff

jacobhinkle:
Seeing an occasional segfault in this pass, for instance in the thunder test failure. Investigating...

jacobhinkle:
!build --diff

jacobhinkle:
!build --diff

jacobhinkle:
The failure in `binary_tests_*_1/3` is real. The distributed test `DistributedTransformerTest.Backward/__bfloat` fails to segment when we make the following replacement in the IR:
```diff
 T62_l_float[ iS179{i21}, iS180{256}, iS181{i23} ] (DeviceMesh{0})
    = __bfloat2float(T60_l___bfloat[ iS172{i21}, iS173{256}, iS174{i23}, rS175{768} ] (DeviceMesh{0}));
 T202_l_float[ ideviceIdx.x603{i21}, iS604{256}, iS605{i23} ] (DeviceMesh{0})
    = Set( T62_l_float[ iS179{i21}, iS180{256}, iS181{i23} ] (DeviceMesh{0}), cache_op=Streaming )
 T61_l___bfloat[ ideviceIdx.x662{i21}, bS177{1}, iS663{i23} ] (DeviceMesh{0})
    = broadcast( T6_g___bfloat[ ideviceIdx.x660{i21}, iS661{i23} ] (DeviceMesh{0}) )
-T199_l___bfloat[ iS664{i21}, bS595{1}, iS665{i23} ] (DeviceMesh{0})
-   = Set( T61_l___bfloat[ ideviceIdx.x662{i21}, bS177{1}, iS663{i23} ] (DeviceMesh{0}), cache_op=Streaming )
 T63_l_float[ iS666{i21}, bS183{1}, iS667{i23} ] (DeviceMesh{0})
-   = __bfloat2float(T199_l___bfloat[ iS664{i21}, bS595{1}, iS665{i23} ] (DeviceMesh{0}));
+   = __bfloat2float(T61_l___bfloat[ ideviceIdx.x662{i21}, bS177{1}, iS663{i23} ] (DeviceMesh{0}));
 T201_l_float[ ideviceIdx.x668{i21}, bS601{1}, iS669{i23} ] (DeviceMesh{0})
    = Set( T63_l_float[ iS666{i21}, bS183{1}, iS667{i23} ] (DeviceMesh{0}), cache_op=Streaming )
 T64_l_float[ ideviceIdx.x185{i21}, iS186{256}, iS187{i23} ] (DeviceMesh{0})
    = T202_l_float[ ideviceIdx.x603{i21}, iS604{256}, iS605{i23} ] (DeviceMesh{0})
    + T201_l_float[ ideviceIdx.x668{i21}, bS601{1}, iS669{i23} ] (DeviceMesh{0});
 T206_l_float[ iS615{i21}, iS616{256}, iS617{i23} ] (DeviceMesh{0})
    = Set( T64_l_float[ ideviceIdx.x185{i21}, iS186{256}, iS187{i23} ] (DeviceMesh{0}), cache_op=Streaming )
```
I think this means we need to make sure that device parallelization is preserved in the replacement.

jacobhinkle:
!build

jacobhinkle:
Distributed test failures are unrelated. I think this is ready for review now.

csrc/preseg_passes/remove_bcast_squeeze.cpp
```diff
@@ -8,46 +8,296 @@
...
+  }
+  NVF_ERROR(replacement != orig, "Expected non-trivial replacement");
+  ir_utils::replaceValInAllExprInputsAndFusionOutputs(orig, replacement);
+  std::vector<IterDomain*> old_loop =
+      TensorDomain::noReductions(orig->getLoopDomain());
+  std::vector<IterDomain*> new_loop =
+      TensorDomain::noReductions(replacement->getLoopDomain());
+  NVF_ERROR(new_loop.size() == old_loop.size());
+  for (size_t i : c10::irange(old_loop.size())) {
+    if (old_loop[i]->isParallelized()) {
```
liqiangxl, line 271: Here in pre-segment pass, device parallelize is the only parallel type we have. Can we just use parallel type device and add NVF to ensure no other unexpected parallel types?

liqiangxl (commented):
LGTM. Leaves some minor comments. Do we have some theory for the algorithm in `AxisOps composeOps(const AxisOps& prev, const AxisOps& next)` ? If not, we may need to add more comments and tests.

csrc/preseg_passes/remove_bcast_squeeze.cpp
```diff
@@ -8,46 +8,296 @@
...
+    TensorView* input_tv = first->input(0)->as<TensorView>();
+    switch (simple_op_type_opt.value()) {
+      case AxisOp::PRESERVE:
+        // This is equivalent to a set Op
+        replacement = input_tv;
+        // Check that parallelization is consistent for replacement. We do not
+        // want to alter the parallelization of input_tv here; we only
+        // parallelize _new_ TVs. If the parallelization is inconsistent, we
+        // simply refuse to make the replacement.
+        if (!hasCompatibleParallelization(orig, replacement)) {
```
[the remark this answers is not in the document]
jacobhinkle: We do not. Previously I was not performing this check inside `isSimpleTVset`. Thanks for pointing this out.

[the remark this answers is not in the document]
jacobhinkle: Actually, I do need to keep it here. If we simplify a "broadcast then squeeze" or "squeeze then broadcast" we might actually hit this path. In that case, it's possible that the input to the first op has different parallelization than the output to the second op. I assume we should skip the replacement in that setting...

csrc/preseg_passes/remove_bcast_squeeze.cpp
```diff
@@ -8,46 +8,296 @@
...
+  if (ldst == nullptr) {
+    return false;
+  }
+  auto in_tv = dynamic_cast<TensorView*>(ldst->in());
+  auto out_tv = dynamic_cast<TensorView*>(ldst->out());
+  return ldst->opType() == LoadStoreOpType::Set && in_tv != nullptr &&
+      out_tv != nullptr
+      // The hasRoot() check is to prevent picking up Set.Permute ops here
+      && !ldst->out()->as<TensorView>()->hasRoot()
+      // A set operation that changes parallelization is not considered trivial
```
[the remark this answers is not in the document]
jacobhinkle: These are actually "resharding" set ops inserted by this pass: https://github.com/NVIDIA/Fuser/blob/4cbd7a4978f23376f9d6dee2efcf981737880a56/csrc/preseg_passes/insert_reshardings.h#L15-L17 E.g. `T199_l` in the example above was a resharding set op that we should not be removing.

csrc/preseg_passes/remove_bcast_squeeze.cpp
```diff
@@ -8,46 +8,314 @@
...
+  ops.reserve(flags.size());
+  for (bool flag : flags) {
+    ops.push_back(flag ? AxisOp::SQUEEZE : AxisOp::PRESERVE);
+  }
+  return ops;
+}
+
+//! Check whether the loop domains are the same size and have the same
+//! parallelization, ignoring reduction dimensions.
+bool hasCompatibleParallelization(TensorView* orig_tv, TensorView* new_tv) {
```
wujingyue, line 65: There's a utility called `isResharding` that you can leverage.

jacobhinkle: Awesome. Thanks for pointing me to this.

csrc/preseg_passes/remove_bcast_squeeze.cpp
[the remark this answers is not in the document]
jacobhinkle: Thanks to @wujingyue's tip, I cleaned this up in d928305c by removing this check, and removing `hasCompatibleParallelization`. In `isReplaceableExpr`, I check `isResharding(expr)` which is a clearer way to avoid replacing reshardings.

jacobhinkle:
> Do we have some theory for the algorithm in `AxisOps composeOps(const AxisOps& prev, const AxisOps& next)` ?

Not anything written down anywhere. I will expand on it in the comments.

csrc/preseg_passes/remove_bcast_squeeze.cpp
```diff
@@ -8,46 +8,410 @@
...
+  flags.reserve(ops.size());
+  for (size_t i : c10::irange(ops.size())) {
+    flags.push_back(ops[i] != AxisOp::PRESERVE);
+  }
+  return flags;
+}
+
+//! Given a descriptors of two sequences of broadcast+squeeze ops, return a
+//! descriptor of their composition
+AxisOps composeOps(const AxisOps& prev, const AxisOps& next) {
```
jacobhinkle, line 158: Rewrote this and explained with an example in comment.

liqiangxl:
!build
seems like needs build to let `nvfuser-ci/build_image_17 ` go

liqiangxl:
!build

jacobhinkle:
gg

> !build seems like needs build to let `nvfuser-ci/build_image_17 ` go

Yeah we need to do `!build` for final merge now, but we can merge as soon as the build job is done

csrc/preseg_passes/remove_bcast_squeeze.cpp
```diff
@@ -165,51 +154,167 @@ std::vector<bool> nonPreservedDims(const AxisOps& ops) {
...
+  // results in an intermediate axis that is consumed in the next op. This
+  // implies that the previous op is not a SQUEEZE and the next op is not a
+  // BROADCAST. If either is PRESERVE, then we can simplify the composition by
+  // simply using the other op. The only case left is then "broadcast then
+  // squeeze", which we ignore since it does not consume an input axis or
+  // produce an output axis.
+  //
+  // For example:
+  //
+  //   prev = [ S S P B P P S ]
```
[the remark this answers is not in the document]
jacobhinkle: Not in a single Expr, but if we used this to compose more than two expressions at once then we could. It seemed about the same difficulty to assume that as it was to just treat the general case as I've done here, so I went with the latter.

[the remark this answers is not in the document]
liqiangxl: `next = [ B P P S P B B ]` doesn't seem valid in this case. For example
```
input = [1, 1, 32, 64, 128, 1]
after prev ops we get:
intermediate = [32, 1, 64, 128]
then apply next ops:
[1, 32, 1, S(corresponds to 128, not valid)]
```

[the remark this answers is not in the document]
jacobhinkle: Preserve does not imply that these are Iteration axes, just that they are not squeezed. I think the squeeze is of 64 in your example, but if the input is `[1, 1, 32, 1, 128, 1]` instead then there's no issue.

csrc/preseg_passes/remove_bcast_squeeze.cpp
```diff
@@ -165,51 +154,167 @@ std::vector<bool> nonPreservedDims(const AxisOps& ops) {
...
+  //     (run processUnalignedOps()) // no change
+  //
+  //     (process aligned ops "preserve then preserve")
+  //     ops = [ S P P P ]
+  //     prev = [ S S P B P P (prev_pos)S ]
+  //     next = [ B P P S P (next_pos)B B ]
+  //
+  //   Post-processing:
+  //     (run processUnalignedOps())
+  //     ops = [ S P P P B ]
```
liqiangxl, line 239: seems should be `ops = [ S P P P P B ]`, where the last `P` comes from `S` in `prev` and `B` in `next`

jacobhinkle: Oh you're right that this is wrong.  The output also should be 6D... Let me take another shot at this. I'll ping you when it's ready for another look.

jacobhinkle: I found the error. When I push a squeeze onto `ops` and `ops.back() == SQUEEZE`, we should convert `ops.back()` to `PRESERVE` instead of removing it. This is now fixed.

liqiangxl (approved):
LGTM.
Did a mannual check of the example.
 if we take `tv0 = [1, 1, 32, 1, 128, 1]` as input, after `prev = [ S S P B P P S ]` we have `tv1 = [32, 1, 1, 128]`; then after `next = [ B P P S P B B ]` we have `tv2= [1, 32, 1, 128, 1, 1]`.
If we pass `tv0 = [1, 1, 32, 1, 128, 1]`  to the composed ops of `ops = [ S P P P P P B ]`, we have `tv3= [1, 32, 1, 128, 1, 1]`
'tv3' is same to `tv2`

jacobhinkle:
!build

PR #3044, opened 2024-09-27
Transformer test updates

cowanmeg:
Update to make transformer test match reference:
(1) In the forward test, layernorm weights and biases are included. They are initialized to one and zero respectively. Names of variables are 0-indexed now to match other tests. 
(2) In the backward test, a sharding propagation was accidentally removed causing extra communication and is added back.

### DIFF
--- a/tests/cpp/test_multidevice_transformer.cpp
+++ b/tests/cpp/test_multidevice_transformer.cpp
@@ -1006,14 +1006,14 @@ TEST_P(DistributedTransformerTest, Forward) {
   const auto options =
       at::TensorOptions().dtype(at_dtype).device(communicator_->device());
   auto x_ = at::randn({B * S, E}, options).to(at::kFloat);
-  auto ln0_w_ = at::ones(E, options).to(at::kFloat);
-  auto ln0_b_ = at::zeros(E, options).to(at::kFloat);
+  auto ln0_w_ = at::randn(E, options).to(at::kFloat);
+  auto ln0_b_ = at::randn(E, options).to(at::kFloat);
   auto mha_w0_ = at::randn({E, 3 * E}, options) * kParamScale;
   auto mha_b0_ = at::randn({3 * E}, options) * kParamScale;
   auto mha_w1_ = at::randn({E, E}, options) * kParamScale;
   auto mha_b1_ = at::randn({E}, options) * kParamScale;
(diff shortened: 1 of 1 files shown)


tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -1000,50 +1006,51 @@ TEST_P(DistributedTransformerTest, Forward) {
   const auto options =
       at::TensorOptions().dtype(at_dtype).device(communicator_->device());
   auto x_ = at::randn({B * S, E}, options).to(at::kFloat);
+  auto ln0_w_ = at::ones(E, options).to(at::kFloat);
+  auto ln0_b_ = at::zeros(E, options).to(at::kFloat);
```
wujingyue, line 1010: Can this hide bugs? For example, a buggy implementation might totally forget to multiply the weight and add the bias, but will pass the comparison. 

I think a better fix is to still use random weight and bias but pass them into the reference implementation.

wujingyue (commented):
LGTM otherwise

tests/cpp/test_multidevice_transformer.cpp
cowanmeg: Sure, this is just to match standard initialization since I'm not sure what a good range is to avoid tuning tolerances

wujingyue: I'd try randn as before and check whether the test will pass. It should because this is a strict improvement. 

That said, there are several places where we don't know what a good range/distribution the activations/weights should be. That's a bigger issue we can postpone to after October.

wujingyue (approved):

cowanmeg:
!build

wujingyue:
There's a real number mismatch in CI. PTAL before merging.

cowanmeg:
> There's a real number mismatch in CI. PTAL before merging.

Thanks for the reminder. I typically check before I merge regardless :)

wujingyue:
> I typically check before I merge regardless :)

Thanks you for that!

samnordmann (approved):

cowanmeg:
build!

cowanmeg:
!build

cowanmeg:
Note: not seeing any tolerance failures with the higher tolerances added. Failure is the same NCCL teardown issue.

Issue #2962, opened 2024-09-18
Failed to update DistributedTransformerTest to use GPT-3 sizes.


wujingyue:
For a fair comparison with the baseline in test_transformer_engine.py. 

I tried the following briefly and hit big number mismatches. 

```patch
diff --git a/tests/cpp/test_multidevice_transformer.cpp b/tests/cpp/test_multidevice_transformer.cpp
index af1187d3..2c922418 100644
--- a/tests/cpp/test_multidevice_transformer.cpp
+++ b/tests/cpp/test_multidevice_transformer.cpp
@@ -17,7 +17,7 @@

 namespace nvfuser {

-constexpr int64_t B = 2, E = 768, H = 16, S = 128;
+constexpr int64_t B = 1, E = 12288, H = 96, S = 2048;
 // Note parameters scaled by kParamScale following weight initialization
 // recommendations:
 // https://huggingface.co/docs/transformers/en/model_doc/gpt2#transformers.GPT2Config.initializer_range
```

```
$ _bn && mpirun -np 4 bin/test_multidevice --gtest_filter=DistributedTransformerTest.Forward/__half

Note: Google Test filter = DistributedTransformerTest.Forward/__half
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from DistributedTransformerTest
[ RUN      ] DistributedTransformerTest.Forward/__half
Note: Google Test filter = DistributedTransformerTest.Forward/__half
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from DistributedTransformerTest
[ RUN      ] DistributedTransformerTest.Forward/__half
Note: Google Test filter = DistributedTransformerTest.Forward/__half
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from DistributedTransformerTest
[ RUN      ] DistributedTransformerTest.Forward/__half
Note: Google Test filter = DistributedTransformerTest.Forward/__half
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from DistributedTransformerTest
[ RUN      ] DistributedTransformerTest.Forward/__half
/opt/pytorch/nvfuser/tests/cpp/test_multidevice_transformer.cpp:107: Failure
Value of: outputs[i].allclose(expected_outputs[i], rtol, atol)
  Actual: false
Expected: true
Output 1 mismatches with atol 0.01:
[...]

PR #3045, opened 2024-09-27
Partially sharded transformer layer forward using Python API. 

wujingyue:
For #2199. 

This PR only shards the MLP. MHA will come in a separate PR (#3115) to keep changes small and incremental.


wujingyue:
!build

wujingyue:
!build

wujingyue:
!build

wujingyue:
!build

wujingyue:
!build

tests/python/test_multidevice.py
```diff
@@ -112,3 +113,362 @@ def multidevice_schedule(self):
...
+            S24,
+            S25,
+            shape=[b, s, h],
+            rng_seed=S3,
+            rng_offset=S2,
+            dtype=DataType.BFloat16,
+        )
+        T31 = self.ops.cast(self.input, dtype=DataType.Float)
+        S32 = self.define_scalar(0.900000, dtype=DataType.Double)
+        T33 = self.ops.lt(T23, S32)
```
samnordmann, line 236: forgive my ignorance, but I'm suprised by this `lt` op. Why is it present here?

tests/python/test_multidevice.py
```diff
@@ -112,3 +113,362 @@ def multidevice_schedule(self):
...
+        T30 = self.ops.uniform(
+            S24,
+            S25,
+            shape=[b, s, h],
+            rng_seed=S3,
+            rng_offset=S2,
+            dtype=DataType.BFloat16,
+        )
+        T31 = self.ops.cast(self.input, dtype=DataType.Float)
+        S32 = self.define_scalar(0.900000, dtype=DataType.Double)
```
samnordmann, line 235: Shouldn't we make `0.900000` a variable ?

tests/python/test_multidevice.py
```diff
@@ -112,3 +113,362 @@ def multidevice_schedule(self):
...
+        T33 = self.ops.lt(T23, S32)
+        S34 = self.define_scalar(0.900000, dtype=DataType.Double)
```
samnordmann, line 237: `S32` and S34` seem redundant

tests/python/test_multidevice.py
```diff
@@ -112,3 +113,362 @@ def multidevice_schedule(self):
...
+            self.layernorm0_weight, shape=[b, s, h], broadcast_dims=[2]
+        )
+        T68 = self.ops.cast(T67, dtype=DataType.Float)
+        T69 = self.ops.mul(T62, T68)
+        T74 = self.ops.broadcast_in_dim(
+            self.layernorm0_bias, shape=[b, s, h], broadcast_dims=[2]
+        )
+        T75 = self.ops.cast(T74, dtype=DataType.Float)
+        T76 = self.ops.add(T69, T75)
+        T77 = self.ops.cast(T76, dtype=DataType.BFloat16)
```
samnordmann, line 259: I understand this is a decomposed layer norm. Could LayerNorm be grouped in a separate function for readability?

tests/python/test_multidevice.py
```diff
@@ -112,3 +113,362 @@ def multidevice_schedule(self):
...
+            end_indices=[b, s, h * 3],
+            strides=[1, 1, 1],
+        )
+        T123 = self.ops.reshape(T104, new_shape=[b, s, a, h // a])
+        T124 = self.ops.permute(T123, dims=[0, 2, 1, 3])
+        T130 = self.ops.reshape(T91, new_shape=[b, s, a, h // a])
+        T131 = self.ops.permute(T130, dims=[0, 2, 1, 3])
+        T137 = self.ops.reshape(T117, new_shape=[b, s, a, h // a])
+        T138 = self.ops.permute(T137, dims=[0, 2, 1, 3])
+        S139 = self.define_scalar(0.100000, dtype=DataType.Double)
```
samnordmann, line 285: Can we use informative name like rather than litteral value, like in https://github.com/NVIDIA/Fuser/blob/30bf8e51ebeb79f9840b3377059905d72dae9f3b/tests/cpp/test_multidevice_transformer.cpp#L27C58-L27C67

tests/python/test_multidevice.py
```diff
@@ -112,3 +113,362 @@ def multidevice_schedule(self):
...
+        S140 = self.define_scalar(True, dtype=DataType.Bool)
+        T141, T142, T143, T144 = self.ops.sdpfa_fwd(T131, T124, T138, S139, S140, None)
+        T145 = self.ops.permute(T141, dims=[0, 2, 1, 3])
+        T146 = self.ops.stride_order(T145, stride_order=[3, 2, 1, 0])
+        T151 = self.ops.reshape(T146, new_shape=[b, s, h])
+        T152 = self.ops.linear(T151, self.mha_linear1_weight, self.mha_linear1_bias)
+        T153 = self.ops.cast(T152, dtype=DataType.Float)
+        T154 = self.ops.cast(T33, dtype=DataType.Float)
+        T155 = self.ops.mul(T153, T154)
+        S156 = self.define_scalar(1.11111, dtype=DataType.Double)
```
samnordmann, line 295: ditto, here and in other places

tests/python/test_multidevice.py
```diff
@@ -112,3 +113,362 @@ def multidevice_schedule(self):
...
+            self.layernorm1_weight, shape=[b, s, h], broadcast_dims=[2]
+        )
+        T191 = self.ops.cast(T190, dtype=DataType.Float)
+        T192 = self.ops.mul(T185, T191)
+        T197 = self.ops.broadcast_in_dim(
+            self.layernorm1_bias, shape=[b, s, h], broadcast_dims=[2]
+        )
+        T198 = self.ops.cast(T197, dtype=DataType.Float)
+        T199 = self.ops.add(T192, T198)
+        T200 = self.ops.cast(T199, dtype=DataType.BFloat16)
```
samnordmann, line 322: here we could reduce code duplication by factoring out the LayerNorm implementation

tests/python/test_multidevice.py
```diff
@@ -112,3 +113,362 @@ def multidevice_schedule(self):
...
+        T206 = self.ops.mul(S205, T204)
+        T207 = self.ops.add(T202, T206)
+        S208 = self.define_scalar(0.797885, dtype=DataType.Double)
+        T209 = self.ops.mul(S208, T207)
+        T210 = self.ops.tanh(T209)
+        S211 = self.define_scalar(0.500000, dtype=DataType.Double)
+        T212 = self.ops.mul(S211, T202)
+        S213 = self.define_scalar(1.00000, dtype=DataType.Double)
+        T214 = self.ops.add(S213, T210)
+        T215 = self.ops.mul(T212, T214)
```
samnordmann, line 337: please write the definition of GELU as a separate function

tests/python/test_multidevice.py
```diff
@@ -112,3 +113,362 @@ def multidevice_schedule(self):
...
+        T216 = self.ops.cast(T215, dtype=DataType.BFloat16)
+        # T217 = self.ops.linear(T216, self.mlp_linear1_weight, self.mlp_linear1_bias)
```
samnordmann, line 339: what is the issue with using Linear layer here?

tests/python/test_multidevice.py
```diff
@@ -112,3 +113,362 @@ def multidevice_schedule(self):
...
+
+    mlp_linear0_weight = torch.randn(
+        d, h * 4 // d, h, dtype=torch.bfloat16, device="cuda"
+    )
+    mlp_linear0_bias = torch.randn(d, h * 4 // d, dtype=torch.bfloat16, device="cuda")
+    mlp_linear1_weight = torch.randn(
+        d, h, h * 4 // d, dtype=torch.bfloat16, device="cuda"
+    )
+    ins = [
+        29,
```
samnordmann, line 425: can you add with inlined comment what each input corresponds to?

tests/python/test_multidevice.py
```diff
@@ -112,3 +113,362 @@ def multidevice_schedule(self):
...
+        )
+
+    torch.cuda.set_device(mpi_test.local_rank)
+
+    mlp_linear0_weight = torch.randn(
+        d, h * 4 // d, h, dtype=torch.bfloat16, device="cuda"
+    )
+    mlp_linear0_bias = torch.randn(d, h * 4 // d, dtype=torch.bfloat16, device="cuda")
+    mlp_linear1_weight = torch.randn(
+        d, h, h * 4 // d, dtype=torch.bfloat16, device="cuda"
```
samnordmann, line 422: The full unsharded tensors are only used for validation and thus should be allocated on CPU, copying then only the right slice to GPU. This will drastically reduce the memory footprint and allow testing with larger shapes.

tests/python/test_multidevice.py
```diff
@@ -112,3 +113,362 @@ def multidevice_schedule(self):
...
+    rank = mpi_test.rank
+
+    b, s, a, h = 2, 2048, 96, 12288
+
+    if h * 4 % d != 0:
+        pytest.skip(
+            f"We only support even split, so {h} * 4 has to be divisible by {d}."
+        )
+
+    torch.cuda.set_device(mpi_test.local_rank)
```
samnordmann, line 415: This type of call, if necessary, should be performed in the backend, probably in `Communicator`'s constructor.

tests/python/test_multidevice.py
```diff
@@ -112,3 +113,362 @@ def multidevice_schedule(self):
...
+        self.sched.parallelize(self.mlp_linear0_bias, 0, nvfuser.ParallelType.mesh_x)
+        self.sched.parallelize(self.mlp_linear1_weight, 0, nvfuser.ParallelType.mesh_x)
+
+
+@pytest.mark.skipif(
+    utils.is_pre_ampere(),
+    reason="Flash Attention is only supported on Ampere and newer devices.",
+)
+@pytest.mark.mpi
+def test_transformer_forward(mpi_test):
```
samnordmann, line 404: can you remind me: are those units automatically tested by CI?

samnordmann (commented):
LGTM! Thanks! I only left some minor comments

tests/python/test_multidevice.py
wujingyue: Good question here and several other questions about readability down below. 

As mentioned [here](https://github.com/NVIDIA/Fuser/pull/3045/files#diff-31dc564b0e6d2c5e614a5ee718088699ed239fb7cbe8c19a4bf7802ec355a7c8R118-R121), this fusion definition is a port from the nvFuser fusion definition generated by Thunder. So I was trying to balance between making the code readable (e.g. giving constants names and extracting common stuff into functions) and minimizing manual changes when next time I update the test due to changes in Thunder. 

I'm on the fence and would love to hear arguments either way.

tests/python/test_multidevice.py
wujingyue: ditto

tests/python/test_multidevice.py
wujingyue: ditto

tests/python/test_multidevice.py
wujingyue: ditto

tests/python/test_multidevice.py
wujingyue: Good idea!

tests/python/test_multidevice.py
```diff
@@ -112,3 +113,362 @@ def multidevice_schedule(self):
...
+# 2. Split device dimensions and parallelize them.
+# 3. Decompose the second linear layer in MLP so the matmul result can be allreduced.
+class TransformerForwardFusion(FusionDefinition):
+    def __init__(self, num_devices, batch, sequence, head, hidden):
+        super().__init__()
+        self._num_devices = num_devices
+        self._batch = batch
+        self._sequence = sequence
+        self._head = head
+        self._hidden = hidden
```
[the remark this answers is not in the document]
wujingyue: This is done by the calling test so I can pytest.skip there: https://github.com/NVIDIA/Fuser/pull/3045/files#diff-31dc564b0e6d2c5e614a5ee718088699ed239fb7cbe8c19a4bf7802ec355a7c8R410-R413

tests/python/test_multidevice.py
wujingyue: ditto

tests/python/test_multidevice.py
wujingyue: ditto

tests/python/test_multidevice.py
wujingyue: https://github.com/NVIDIA/Fuser/pull/3127

wujingyue:
!build

tests/python/test_multidevice.py
samnordmann: Ok, I did not realize that. I am personally in favor of making the code readable, because 1) we shouldn't expect fundamental change from thunder, and 2) imo we don't necessarily need to track in this unit test all small changes from thunder: this test is useful because it is _one_ implementation of transformer, and it's ok if it differs from the actual thunder generation.

Not having DropOut, LayerNorm, GeLu scooped out as auxiliary functions makes imo the code error-prone, not flexible, and not readable.

I would, at least, add comments delimiting the different primitives.

This being said, I might be missing the actual goal of this unit test, so feel free to do as you think is best

samnordmann: Ok, I did not realize that. I am personally in favor of making the code readable, because 1) we shouldn't expect fundamental change from thunder, and 2) imo we don't necessarily need to track in this unit test all small changes from thunder: this test is useful because it is _one_ implementation of transformer, and it's ok if it differs from the actual thunder generation.

Not having DropOut, LayerNorm, GeLu scooped out as auxiliary functions makes imo the code error-prone, not flexible, and not readable.

I would, at least, add comments delimiting the different primitives.

This being said, I might be missing the actual goal of this unit test, so feel free to do as you think is best

tests/python/test_multidevice.py
samnordmann: I see, it is done in the computation of `T155` way below. Thanks!

tests/python/test_multidevice.py
samnordmann: what about that ?

tests/python/test_multidevice.py
samnordmann: is it really the same consideration as in https://github.com/NVIDIA/Fuser/pull/3045/files#r1790935495 ?

tests/python/test_multidevice.py
wujingyue: I commented in the code.

tests/python/test_multidevice.py
wujingyue: I agree with most of your points. I recommend we merge this and discuss the plan after @cowanmeg is back. While I love cleanups, I'll feel more comfortable doing so if I know they'll stay. 

The purpose of this test is for performance comparison. test_transformer_engine is the baseline, SOTA. This test is to reflect what Thunder/nvFuser will generate. "Will" because we've yet to integrate multi-GPU nvFuser into Thunder. For that, I'd like it to be as close as possible to what Thunder will generate. When the integration is ready, we can/should remove this test, and build and track the benchmark in Thunder instead. 

That said, there's a different purpose that you are probably alluding to: build a best performing sharded transformer layer using nvFuser's API, regardless of the framework. There's definitely value to that, and in that case I'm all for optimizing readability. However, there's DistributedTransformerTest.Forward/Backward that serves the same purpose. We should discuss whether/how the two should co-exist. 

An incidental note: due to #2962, it's currently hard to verify any cleanup is behavior neutral.

tests/python/test_multidevice.py
wujingyue: I added a code comment to point readers to their meanings. I could add inlined comments for each argument, but I would have a hard time making sure they stay consistent with those in the definition. So I prefer adding a pointer so there's only one source of truth.

samnordmann: thanks!

tests/python/test_multidevice.py
samnordmann: Alright

samnordmann (approved):

PR #3054, opened 2024-09-30
Backprop meshes in addition to forward. 

wujingyue:
See code comments for details.

### DIFF
--- a/csrc/preseg_passes/propagate_shardings.cpp
+++ b/csrc/preseg_passes/propagate_shardings.cpp
@@ -71,7 +71,8 @@ void PropagateShardingsPass::runPass(Fusion* fusion) {
     shardAllLike(input_with_mesh, outputs_without_mesh);
   }
 
-  // Back-propagate device meshes. This is needed in addition to the forward
+  // Back-propagate device meshes. This makes sure all TensorViews have a mesh
+  // if any of them has one. This is needed in addition to the forward
   // propagation for ops that don't take any TensorView operands, e.g.,
   // `uniform` used in dropout. See MultiDeviceTest.BackpropMeshes for an
   // example.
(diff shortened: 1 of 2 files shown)


wujingyue:
!build

csrc/preseg_passes/propagate_shardings.cpp
```diff
@@ -30,15 +29,40 @@ void PropagateShardingsPass::runPass(Fusion* fusion) {
...
-    for (TensorView* tv : outputs) {
+    for (auto* tv : outputs) {
       if (!tv->hasDeviceMesh()) {
         outputs_without_mesh.push_back(tv);
       }
     }
     shardAllLike(input_with_mesh, outputs_without_mesh);
   }
 
+  // Back-propagate device meshes. This is needed in addition to the forward
```
cowanmeg, line 42: This will cause some issues where a DID parallelization needed to be back-propagated as well. Since dropout is replicated, we don't see this but this could cause lots of confusion if we only propagate the mesh.

wujingyue: That's a valid concern. I could manually set meshes for all "tensor factory" calls -- only `uniform` is used in the transformer layer. 

I wonder your thoughts on a "good default". For UX, I'm OK with having to set meshes for fusion inputs and outputs. However, having to do for intermediate tensors isn't ideal. In addition to the extra code the user has to write, these tensors might not always be exposed to Thunder. `uniform` happens to be because it's a one-to-one mapping, but a Thunder prim can well be converted to a tensor factory call followed by an elementwise...

cowanmeg (approved):

wujingyue:
!build

PR #3059, opened 2024-09-30
Host IRs: Reduce, View, and a generic default support through Expression Evaluator

samnordmann:
# What

This PR adds to `HostIrExecutor` a default generic way to handle expression which are supported by `ExpressionEvaluator`, i.e., which implement an `Expr::evaluate` method.

This mode of execution is already exercised for `SliceOp`, `SelectOp` and `MatmulOp`. In this PR, we add tests for `ViewOp` and `ReduceOp`.

# Why

- #3060 requires a `ReduceOp`
- The code gets refactored and `HostIrExecutor` gets more capable

# How

We make `HostIrExecutor` inherit from `OptOutDispatch` (instead of `OptInDispatch`) and use `OptOutDispatch::unhandled` to implement the generic handling through `ExpressionEvaluator`


samnordmann:
!build

wujingyue (approved):
Good idea!

wujingyue:
!build

PR #3061, opened 2024-09-30
Revert "only set maxrregcount when want to use less than 255 registers"

liqiangxl:
Reverts NVIDIA/Fuser#2968
Check code diff.


liqiangxl:
!build -diff

liqiangxl:
!build  --diff --diff-bench

PR #3062, opened 2024-09-30
Revert "only set maxrregcount when want to use less than 255 registers"

liqiangxl:
Reverts NVIDIA/Fuser#2968 to check register usage changes.
Not intend to do the actual revert, just use this PR to  check register usage changes

### DIFF
--- a/csrc/fusion_executor/executor_utils.cpp
+++ b/csrc/fusion_executor/executor_utils.cpp
@@ -811,10 +811,8 @@ std::optional<int64_t> getMaxRegCount(
     max_register = env_max_reg_count;
   }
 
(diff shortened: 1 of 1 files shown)


liqiangxl:
!build --diff --diff-bench

liqiangxl:
**changes if revert**:
tests/python/test_schedule_ops.py::TestScheduleOps::test_reduction_factor_op `registers: 168→ 254 `
image

So, the original PR reduced register usage in most cases. In several cases, it increased register usage (e.g. 30->32, 14->20), these changes won't decrease occupancy.

naoyam:
Thanks for checking!

PR #3060, opened 2024-09-30
RS+GEMM p2p pipelining: host IR implementation

samnordmann:
# What
adds a Host IR implementation of the RS+GEMM p2p pipelining as was already implemented in ATen in #2950

Nsight profile obtained on 8*H100 DGX, M=2^8, K=N=2^17, S=8, number_of_streams=3
Screenshot 2024-10-14 at 15 18 02

Dependency on:

- [x] #2970
- [x] #2991
- [x] #3058
- [x] #3059

TODOs:
- Add UCC support: Support not coalesced p2p calls
- Benchmark

PR #3063, opened 2024-09-30
Host IRs: support `IfThenElse`

samnordmann:
# What
- `HostIrExecutor` support for `IfThenElse`
- `HostIrExecutor` support scalar inputs. This was needed for testing, were we pass a `bool` as input.

We might wanna move `IfThenElse` and `Predicate` nodes out of the `kir` namespace, as we previously did for ForLoop #2440. Lmk what you think.

# Why
- It will be needed in the future for adding UCC support in Host IR RS+GEMM p2p pipelining https://github.com/NVIDIA/Fuser/pull/3060
- Another application would be to add predicate surrounding the compute and comms in case of Pipeline parallelism. Today, we are simply removing in each nodes the parts of the graph that should not be executed, thus resulting in a non SPMD program


csrc/fusion_executor/executor_kernel_arg.h
```diff
@@ -135,4 +135,6 @@ std::vector<std::byte> getKernelArgument(
     Val* parameter,
     PrimDataType index_type);
 
+PolymorphicValue IValueToPolymorphicValue(const c10::IValue& val);
```
samnordmann, line 138: should this util function be moved to another file?

wujingyue: cc @zasdfgbnm -- I don't mind either way.

wujingyue (approved):

samnordmann:
!build

PR #3064, opened 2024-09-30
Allgather-based pipelining ATen Implementation

nsarka:
From the document with some tasks Sam prepared for me: https://docs.google.com/document/d/1Oj-EgUyhgfpxDFkfF43Q4pIe_4gtJzXOSUIgjYEI7wU/edit#heading=h.t098gw7m1xok

This PR implements Allgather-based pipelining

### DIFF
--- a/tests/cpp/test_multidevice_overlap.cpp
+++ b/tests/cpp/test_multidevice_overlap.cpp
@@ -592,7 +592,6 @@ class AGOverlapTest : public MultiDeviceTest {
 };
 // This test implements an allgather-based pipelining overlapping technique,
 // similar to the above reduce-scattered based pipelining overlapping technique
-// clang-format on
 TEST_F(AGOverlapTest, AllgatherBasedPipeliningATenImplementation) {
   std::vector<c10::cuda::CUDAStream> streams;
   std::generate_n(
--- a/tests/cpp/test_multidevice_overlap.cpp
+++ b/tests/cpp/test_multidevice_overlap.cpp
@@ -545,7 +545,7 @@ class AGOverlapTest : public MultiDeviceTest {
     auto cpu_options = at::TensorOptions().dtype(at::kFloat);
     at::TensorOptions gpu_options = cpu_options.device(communicator_->device());
 
-    ta_ = at::empty(ta_sizes, gpu_options);
+    ta_ = at::empty(ta_sizes, cpu_options);
     tb_ = at::empty(tb_sizes, gpu_options);
     tc_ = at::empty(tc_sizes, gpu_options);
     ta_sharded_ = at::empty(ta_sharded_sizes, gpu_options);
@@ -560,15 +560,15 @@ class AGOverlapTest : public MultiDeviceTest {
   }
 
   void initializeIO() {
-    ta.uniform_();
-    tb.uniform_();
+    ta_.uniform_();
+    tb_.uniform_();
     ta_sharded_.copy_(ta_.select(0, my_device_index_));
-    tb_sharded_.copy_(tb_.select(0, my_device_index_));
   }
 
   void validate() {
     // compute the expected output for data correctness validation
-    auto tc_expected_ = torch::matmul(ta_, tb_);
+    auto tb_cpu = tb_.to(torch::kCPU);
+    auto tc_expected_ = torch::matmul(ta_, tb_cpu); // ta already on cpu
     EXPECT_TRUE(tc_.to(torch::kCPU).allclose(tc_expected_, 1e-1, 1e-1))
         << "Unexpected results, obtained:" << tc_
         << "\n expected: " << tc_expected_;
@@ -615,15 +615,11 @@ TEST_F(AGOverlapTest, AllgatherBasedPipeliningATenImplementation) {
       int64_t stream_index = j % streams.size();
       setCurrentCUDAStream(streams.at(stream_index));
 
-      // define the sliced tensors
-      auto ta_j = ta_.select(0, j);
-      auto tc_j = tc_.select(0, j);
-
       // local compute
-      torch::matmul_out(tc_j, ta_j, tb_);
+      torch::matmul_out(tc_sharded_, ta_sharded_, tb_);
 
       // communication
-      world_communicator_->_allgather_base(tc_, tc_j, tc_j)
+      world_communicator_->_allgather_base(tc_, tc_sharded_)
           ->wait();
     }
   }
--- a/tests/cpp/test_multidevice_overlap.cpp
+++ b/tests/cpp/test_multidevice_overlap.cpp
@@ -495,15 +495,15 @@ TEST_F(
   }
 }
 
-class AGOverlapTest : public MultiDeviceTest {
+class AllgatherOverlapTest : public MultiDeviceTest {
  protected:
   OverlapTestParams params;
 
   int64_t num_devices_;
   int64_t my_device_index_;
   std::vector<int64_t> all_devices_;
-  at::Tensor ta_, tb_, tc_;
-  at::Tensor ta_sharded_, tc_sharded_;
+  at::Tensor ta_unsharded_, tb_unsharded_, tc_unsharded_;
+  at::Tensor ta_, tc_;
   // stores the backend
   c10d::Backend* world_communicator_;
 
@@ -511,11 +511,11 @@ class AGOverlapTest : public MultiDeviceTest {
   // A has shape (S, sharded(D), M/(S*D), K)
   // B(K,N)
   // C has shape (S, M/S, N)
+  std::vector<int64_t> ta_unsharded_sizes;
+  std::vector<int64_t> tb_unsharded_sizes;
+  std::vector<int64_t> tc_unsharded_sizes;
   std::vector<int64_t> ta_sizes;
-  std::vector<int64_t> tb_sizes;
   std::vector<int64_t> tc_sizes;
-  std::vector<int64_t> ta_sharded_sizes;
-  std::vector<int64_t> tc_sharded_sizes;
   std::vector<int64_t> tc_j_sizes;
 
   void SetUp() {
@@ -523,8 +523,7 @@ class AGOverlapTest : public MultiDeviceTest {
 
     num_devices_ = communicator_->size();
     my_device_index_ = communicator_->deviceId();
-    ASSERT_EQ(params.M % num_devices_, 0);
-    ASSERT_EQ(params.S % num_devices_, 0);
+    ASSERT_EQ(params.M  % (num_devices_ * params.S), 0);
 
     // Setup the world communicators
     std::vector<int64_t> devices(num_devices_);
@@ -538,14 +537,14 @@ class AGOverlapTest : public MultiDeviceTest {
       debug() << params << std::endl;
     }
 
-    ta_sizes = std::vector<int64_t>{
+    ta_unsharded_sizes = std::vector<int64_t>{
         params.S, num_devices_, params.M / (num_devices_ * params.S), params.K};
-    tb_sizes = std::vector<int64_t>{params.K, params.N};
-    tc_sizes = std::vector<int64_t>{
+    tb_unsharded_sizes = std::vector<int64_t>{params.K, params.N};
+    tc_unsharded_sizes = std::vector<int64_t>{
         params.S, num_devices_, params.M / (num_devices_ * params.S), params.N};
-    ta_sharded_sizes = std::vector<int64_t>{
+    ta_sizes = std::vector<int64_t>{
         params.S, params.M / (num_devices_ * params.S), params.K};
-    tc_sharded_sizes = std::vector<int64_t>{
+    tc_sizes = std::vector<int64_t>{
         params.S, params.M / (num_devices_ * params.S), params.N};
     tc_j_sizes = std::vector<int64_t>{
         params.M / params.S, params.N};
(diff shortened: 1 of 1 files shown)


wujingyue:
!build

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -472,4 +472,136 @@ TEST_F(
...
+    // compute the expected output for data correctness validation
+    auto tb_cpu = tb_.to(torch::kCPU);
+    auto ta_cpu = ta_.to(torch::kCPU);
+    auto tc_expected_ = torch::matmul(ta_cpu, tb_cpu);
+    EXPECT_TRUE(tc_.to(torch::kCPU).allclose(tc_expected_, 1e-1, 1e-1))
+        << "Unexpected results, obtained:" << tc_
+        << "\n expected: " << tc_expected_;
+  }
+};
+// clang-format off
```
wujingyue, line 570:
```suggestion
```

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -472,4 +472,136 @@ TEST_F(
...
+// This test implements an allgather-based pipelining overlapping technique,
+// similar to the above reduce-scattered based pipelining overlapping technique
+// clang-format on
```
wujingyue, line 573:
```suggestion
```

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -472,4 +472,136 @@ TEST_F(
...
+  std::vector<int64_t> all_devices_;
+  at::Tensor ta_, tb_, tc_;
+  at::Tensor ta_sharded_, tc_sharded_;
+  // stores the backend
+  c10d::Backend* world_communicator_;
+
+  // Define I/O and intermediate Tensor shapes
+  // A has shape (S, sharded(n), M/(S*n), K)
+  // B(K,N)
+  // C has shape (S, M/S, N)
```
wujingyue, line 490:
```suggestion
  // A has shape (S, sharded(D), M/(S*D), K)
  // B(K,N)
  // C has shape (S, M/S, N)
```
Nit: I assume `n` is the number of devices? I changed it to D to avoid confusion with the other `N`, but I'm open to other ways to avoid confusion.

wujingyue, line 490: This doesn't look right. I believe `C` should be sharded as well.

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -472,4 +472,136 @@ TEST_F(
...
+    world_communicator_ =
+        communicator_->getBackendForTeam(all_devices_, params.backend_type);
+
+    // Debug print
+    if (communicator_->deviceId() == 0 && debug_print) {
+      debug() << params << std::endl;
+    }
+
+    ta_sizes = std::vector<int64_t>{
+        params.S, num_devices_, params.M / (num_devices_ * params.S), params.K};
```
wujingyue, line 519: Can you point me to the source of this algorithm? What you try to implement is apparently different from [what I had in mind](https://docs.google.com/document/d/1BK_xEolbJDEky5trLW8jbuNHMX-v2QSY40O8782lYQ8/edit#heading=h.ao128ya3cf28). I don't have evidence showing which is better, but I suspect host-sharding the M dimension (in addition to device-sharding it already) may lead to wide submatrices that are less compute friendly. So I'm curious what people do in practice in e.g. TransformerEngine.

tests/cpp/test_multidevice_overlap.cpp
wujingyue: I don't follow you. But this comment is at least different from [the code](https://github.com/NVIDIA/Fuser/pull/3064/files#diff-30df6421558f87ef0024b01f11752c35d3d68b80a9e6e0ec0fd49de535acb91aR521-R522). It would make sense if it says `(S, sharded(n), M/(S*n), N)`. Am I missing something?

tests/cpp/test_multidevice_overlap.cpp
nsarka: There is no source, this is just what Sam and I put together from our meeting together: https://docs.google.com/document/d/1Fzr9Zs2Dqfj3e4yR8LKxFrRqC1EkMUfQczJMYQGJQUI/edit#heading=h.8niucpl6b50

Would it make sense for me to implement the algorithm you had in mind using ATen after this one is merged? I believe the main difference is that yours device shards along rows of A and host-shards along columns of B, where mine does both on A and fully replicates B.

tests/cpp/test_multidevice_overlap.cpp
wujingyue: I'll read it again. I've been incorrectly assuming B is the tensor to allgather and A and C are sharded.

tests/cpp/test_multidevice_overlap.cpp
wujingyue: > I believe the main difference is that yours device shards along rows of A and host-shards along columns of B, where mine does both on A and fully replicates B.

There's that, but I don't worry about that difference too much. The main difference to me is that your current implementation host-shards A's rows and we could alternatively host-shard A's columns (while A remains to be device-sharded row-wise). 

Yes, a follow-up sounds good to me. It's certainly shape dependent.

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -472,4 +472,134 @@ TEST_F(
...
+  std::vector<int64_t> tc_sharded_sizes;
+  std::vector<int64_t> tc_j_sizes;
+
+  void SetUp() {
+    MultiDeviceTest::SetUp();
+
+    num_devices_ = communicator_->size();
+    my_device_index_ = communicator_->deviceId();
+    ASSERT_EQ(params.M % num_devices_, 0);
+    ASSERT_EQ(params.S % num_devices_, 0);
```
wujingyue, line 504: Why is this required?

wujingyue (commented):
LGTM otherwise

tests/cpp/test_multidevice_overlap.cpp
samnordmann: Tiling the reduced dimension is definitely possible, but it is not what is done in Transformer sequence parallel

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -472,4 +472,134 @@ TEST_F(
...
 
+class AGOverlapTest : public MultiDeviceTest {
+ protected:
+  OverlapTestParams params;
+
+  int64_t num_devices_;
+  int64_t my_device_index_;
+  std::vector<int64_t> all_devices_;
+  at::Tensor ta_, tb_, tc_;
+  at::Tensor ta_sharded_, tc_sharded_;
```
samnordmann, line 483: nit: Since A is sharded, and B and C are unsharded, I would use the notations `ta_unsharded_` and `tc_sharded_` in contrast with `ta_, tb_, tc_`

tests/cpp/test_multidevice_overlap.cpp
samnordmann: indeed, doesn't seem required. However we should make sure that `params.M  % (num_devices_ * params.S) == 0`

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -472,4 +472,134 @@ TEST_F(
...
+
+    // Set up input tensors. We create the full unsharded tensors and define the
+    // actual input as the shard corresponding to the current device. Having the
+    // full unsharded input on each rank makes it possible to compute the
+    // expected result locally, hence, this way of doing is convenient for
+    // validating data correctness.
+    auto cpu_options = at::TensorOptions().dtype(at::kFloat);
+    at::TensorOptions gpu_options = cpu_options.device(communicator_->device());
+
+    ta_ = at::empty(ta_sizes, gpu_options);
```
samnordmann, line 538: `ta_unsharded_` should be allocated on cpu

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -472,4 +472,134 @@ TEST_F(
...
+      auto ta_sharded_j = ta_sharded_.select(0, j); // params.M / (num_devices_ * params.S), params.K
+      auto tc_sharded_j = tc_sharded_.select(0, j); // params.M / (num_devices_ * params.S), params.N
+      torch::matmul_out(tc_sharded_j, ta_sharded_j, tb_); // params.M / (num_devices_ * params.S), params.N
+
+      // communication
+      auto tc_j = tc_.select(0, j); // num_devices_, params.M / (num_devices_ * params.S), params.N
+      world_communicator_->_allgather_base(tc_j, tc_sharded_j) // num_devices_, params.M / (num_devices_ * params.S), params.N
+          ->wait();
+    }
+
```
samnordmann, line 600: we should `synchronizeStreams` here

samnordmann (commented):
Looks good, thanks! I only left a couple of minor comments

tests/cpp/test_multidevice_overlap.cpp
wujingyue: Let me try again. I was talking about host-sharding the reduced dimension, which I believe is what's done and/or what can be done for sequence parallel. For example:

Assume batch size is 1. 

```
input: [s_d, h]. the output of the second layernorm, which is sequence parallel
weight: [h, 4h_d]. the weight of the first linear layer in MLP, which is hidden dimension parallel
output: [s, 4h_d]

input' = allgather(input)
output = matmul(input', weight)
```

This can be host-sharded to S iterations. The first iteration is

```
output = 0
for i in range(S):
  input' = allgather(input[s_d, h/S*i:h/S*(i+1)])
  output' = matmul(input', weight[h/S*i:h/S*(i+1)])
  output += output'
```

In GPT-3, s=2048 and h=12288. So I think host-sharding h is preferable.

samnordmann: I was not understanding correctly. You are not talking about sharding but about host-tiling those axis, nvm.

So, it is _possible_ to host-tile either the non-contracting or the contracting axis. It is important to support both eventually. The same holds for GEMM+RS. I suggest to merge this PR and to implement the other one later.

We started with tiling the non-contracting axis because **1) it is what is done in TE** and 2) it is simpler: when tiling the contracting dimension, we need to tile both the input and the weight, and to accumulate the result.

_Another practical point to keep in mind: the host-parallelilzed axis needs general to be at the outermost position in the allocation domain, otherwise the `allgather` is not possible as-is. The layout we chose is thus slightly more straightforward as it saves a transposition._

Thanks for the concrete dimension lengths, that's really helpful. I hear you: it is in general better to tile large dimension. However, I would be careful about what's best for e2e performance. For example, one element in favor of tiling the non-contracting axis: tiling the contracting axis requires a global synchronization at the end of the AG+GEMM, while the alternative potentially allows to push the synchronization point further, even possibly at the very end of the MLP, if I'm not mistaken.

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -495,4 +495,134 @@ TEST_F(
...
+    initializeIO();
+
+    for (auto j : c10::irange(params.S)) {
+      int64_t stream_index = j % streams.size();
+      setCurrentCUDAStream(streams.at(stream_index));
+
+      // local compute
+      auto ta_sharded_j = ta_sharded_.select(0, j); // params.M / (num_devices_ * params.S), params.K
+      auto tc_sharded_j = tc_sharded_.select(0, j); // params.M / (num_devices_ * params.S), params.N
+      torch::matmul_out(tc_sharded_j, ta_sharded_j, tb_); // params.M / (num_devices_ * params.S), params.N
```
samnordmann, line 616: Can we permute `matmul` and `allgather`? I mean, to first allgather and then matmul. I sorry if I advised otherwise before.

I understand both solutions are possible, but since what we target is Megatron parallelization, we should stick to it and first study the AG+GEMM. Sorry for the trouble

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -495,4 +495,130 @@ TEST_F(
...
+  void initializeIO() {
+    ta_unsharded_.uniform_();
+    tb_unsharded_.uniform_();
+    ta_.copy_(ta_unsharded_.select(1, my_device_index_));
+  }
+
+  void validate() {
+    // compute the expected output for data correctness validation
+    auto tb_unsharded_cpu = tb_unsharded_.to(torch::kCPU);
+    auto tc_unsharded_expected_ = torch::matmul(ta_unsharded_, tb_unsharded_cpu);
```
wujingyue, line 579:
```suggestion
    auto tc_unsharded_expected_ = torch::matmul(ta_unsharded_, tb_unsharded_.cpu());
```

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -495,4 +495,130 @@ TEST_F(
...
+    EXPECT_TRUE(tc_unsharded_.to(torch::kCPU).allclose(tc_unsharded_expected_, 1e-1, 1e-1))
```
wujingyue, line 580:
```suggestion
    EXPECT_TRUE(tc_unsharded_.cpu().allclose(tc_unsharded_expected_, 1e-1, 1e-1))
```

wujingyue (approved):

wujingyue:
LGTM. Can you fix lint and `!build` to verify the unit tests?

tests/cpp/test_multidevice_overlap.cpp
```diff
@@ -495,4 +495,130 @@ TEST_F(
...
+      auto ta_allgathered_j = ta_allgathered_.select(0, j);
+      world_communicator_->_allgather_base(ta_allgathered_j, ta_j) // num_devices_, params.M / (num_devices_ * params.S), params.K
+          ->wait();
+
+      // local compute
+      auto tc_j = tc_unsharded_.select(0, j); // num_devices_, params.M / (num_devices_ * params.S), params.N
+      torch::matmul_out(tc_j, ta_allgathered_j, tb_unsharded_); // num_devices_, params.M / (num_devices_ * params.S), params.N
+    }
+
+    for (const auto& stream : streams) {
```
samnordmann, line 616: nit: you can use the function `synchronizeStreams(streams)` defined at the top of this file

samnordmann (approved):

nsarka:
> LGTM. Can you fix lint and `!build` to verify the unit tests?

Yes! Updated.

wujingyue:
!build

PR #3043, opened 2024-09-27
Always use scheduleMultipleMatmuls

jacobhinkle:
This removes the old matmul scheduler and switches to only using `scheduleMultipleMatmuls` instead. The tests for compatibility with the old scheduler are removed.

I have left the `NVFUSER_ENABLE=fuse_multiple_matmuls` option alive so that we can opt into accepting multiple matmuls in a fusion segment for now.

PR #3065, opened 2024-09-30
Multi-matmul scheduler: Skip device dims when counting local batch dims

jacobhinkle:
This patches a small bug that incorrectly counted dimensions. We use a different system (`MatmulDimRolesMap`) to detect device and batch dimensions now, and it was identifying some device dims as batch dimensions. Although these dimensions _are_ batch dimensions in the batch matmul sense, they are not _local_ batch dimensions, i.e. they are not realized on each rank, so we skip them when computing `num_local_batch_dims_`. By contrast, in the existing scheduler we simply look for a device dim then subtract that plus 3 from `mma_result->nDims()`, which is not as flexible for future supported use cases with multiple M, N, or K dimensions.

This is necessary to get #3043 working.

### DIFF
--- a/tests/cpp/test_multi_matmul_scheduler.cpp
+++ b/tests/cpp/test_multi_matmul_scheduler.cpp
@@ -563,6 +563,40 @@ TEST_P(MultiMatmulSchedulerMatchTest, BatchMatmul) {
   fusion->addOutput(tv2);
 }
 
+// Copied from DistributedMatmulTest.MulSum_LayoutTN_NoComms
+TEST_P(MultiMatmulSchedulerMatchTest, MultiDeviceMulSum) {
+  int num_devices = 1;
+  auto mesh = DeviceMesh::createForNumDevices(num_devices);
+  int M = 256, N = 64, K = 64;
+  int Mo = num_devices;
+  int Mi = M / Mo;
+  std::vector<int> a_shape = {Mo, Mi, K};
+  std::vector<int> b_shape = {N, K};
+
+  TensorView* a = makeContigTensor(3, DataType::Half); // (Mo,Mi,K)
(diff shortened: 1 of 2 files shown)


jacobhinkle:
!build

jacobhinkle:
Confirmed in https://github.com/NVIDIA/Fuser/pull/3043#issuecomment-2383626752 of #3043 that this fixes only outstanding incompatibility between the new scheduler and the original, so #3043 will be ready once this PR is merged.

tests/cpp/test_multi_matmul_scheduler.cpp
```diff
@@ -563,6 +563,40 @@ TEST_P(MultiMatmulSchedulerMatchTest, BatchMatmul) {
...
+  for (auto tv : all_sharded_tvs) {
+    tv->axis(0)->parallelize(ParallelType::DIDx);
+    tv->setDeviceMesh(mesh);
+  }
+  b->setDeviceMesh(mesh);
+  // TODO: If c's allocation domain isn't set, it will fail validation at
+  // csrc/device_lower/validation.cpp:419, Vectorized dim for consumer has to be
+  // from a contiguous inner most position.
+  c->setAllocationDomain(c->getLoopDomain(), true);
+}
```
zasdfgbnm, line 598: Are we calling any scheduler?

zasdfgbnm (approved):

tests/cpp/test_multi_matmul_scheduler.cpp
jacobhinkle: It's a little bit naughty, but I'm running `compareSchedules` inside `MultiMatmulSchedulerMatchTest::TearDown()`, so it does actually get scheduled, similar to the other cases.

PR #3077, opened 2024-10-01
Cleanup `applyMmaSwizzleForTMALoad`

zasdfgbnm:
In the past, we had a `permute_outer_dim` parameter for `applyMmaSwizzleForTMALoad` because strides in the Mma Op's matrix descriptor were hard-coded. With hardcoded strides, we had zero flexibility, and had to use multiple TMA to load smem buffer into a very specific format. After https://github.com/NVIDIA/Fuser/pull/3002, the strides were computed based on the schedule, so we do have the flexibility to choose a memory layout that uses less TMA instructions to load.


tests/cpp/test_mma.cpp
```diff
@@ -902,91 +902,6 @@ TEST_P(HopperRS, SingleTileWithTMALoadStore) {
   EXPECT_TRUE(at::allclose(cg_outputs[0], tref, 1e-5, 1e-5));
 }
 
-TEST_P(HopperRS, SingleTileWithTMALoadOuterDimNotSplit) {
```
zasdfgbnm, line 905: Now, all the tests have "OuterDimNotSplit", so there is no need to have a separate `OuterDimNotSplit` test.

zasdfgbnm:
!build

zasdfgbnm:
!build

protonu (approved):
LGTM!

Issue #3072, opened 2024-10-01
RoPE-like slice-concat patterns may result in cyclic exact graphs


naoyam:
Example (same as #3068):

```
t0: [i0]
t1 = t0 // [i1]
t2 = t1 // [i2]
t3 = t2[0:2] // [i3]
t4 = pad(t23 {0, i0 - 2}) // [i4]
t5 = t1 + t4 // [i5]
```

![Image](https://github.com/user-attachments/assets/963b11ed-f009-419a-a488-bef357c57a7a)

The exact graph groups `i0`, `i1`, `i2`, `i4` and `i5` together as they are just used together with the pointwise ops. As you can see in the above diagram, this results in a cycle between the two groups.

Being acyclic is one of the basic assumptions we have in the loop promotion analysis. For example, `ValGraphStmtSort` won't work with cyclic graphs, which means, for example, `computeCoveredGroups` wont't work since it relies on `ValGraphStmtSort`. The propagation of promotion info through the IEL graph (intersection of exact and loop graphs) is also done with `ValGraphStmtSort`, so that may not work either, although IEL might not necessarily be cyclic even when the exact graph is cyclic.

jacobhinkle:
Can't you also have cycles in the ValGraph with reshape?
```
tv0 // [ iS0{12} ]
tv1 = reshape(tv0, {4, 3})  // [ iS1{4}, iS2{3} ]
tv2 = reshape(tv1, {12})  // [ iS3{12} ]
tv3 = add(tv0, tv2)  // [ iS4{12} ]
```
IDs {0, 3, 4} are exact mapped, and there's a split from that to {1} and {2} and a merge back.

naoyam:
That is true too. It's just we haven't been hit by actual use cases (yet...).

PR #3079, opened 2024-10-01
Loop promotion analysis WAR/shortcut and cycle check

naoyam:
To workaround the cyclic graph issue (#3072) as much as possible, this PR skips the full analysis when possible. When skipped, the alternative analysis is just simply picking the first ID, so it shouldn't have any constraint with the graph structure. It can be also considered a shortcut as it would skip the analysis if not necessary. 

Also added a check of existence of cycles in a graph as it isn't supported.


naoyam:
!build

naoyam:
!build

naoyam:
!build

csrc/device_lower/utils.h
```diff
@@ -16,7 +16,6 @@
 #include <kernel_ir.h>
 #include <parallel_type_bitmap.h>
 #include <val_graph.h>
-#include <val_graph_visitor.h>
```
naoyam, line 19: Cleaned up the usage of this header file as I noticed just changing this file required recompilation of more than 160 files..

naoyam:
Hmm, I didn't expect this would cause any error. Making this draft for now.

naoyam:
!build

csrc/id_model/loop_promotion.cpp
```diff
@@ -898,4 +932,60 @@ std::unordered_map<ValGroup, IterDomain*> LoopPromotionMapBuilder::get(
...
+      if ((is_const && !this_is_const) ||
+          (is_const == this_is_const && this_num_exprs >= num_exprs)) {
+        continue;
+      }
+
+      // Update the current promotion
+      is_const = this_is_const;
+      promotion = loop_id;
+      num_exprs = this_num_exprs;
+    }
```
naoyam, line 981: This part is completely optional. Just make it easier to my eyes when looking at debug print and also it avoids this test error. https://github.com/NVIDIA/Fuser/pull/3079#issuecomment-2387571701

zasdfgbnm (approved):

naoyam:
!build

PR #3084, opened 2024-10-02
Specify different-dtype alias TV alignment

jacobhinkle:
See https://github.com/NVIDIA/Fuser/pull/2934#discussion_r1783671457

PR #665 allowed us to re-use allocations that have different dtypes. We already check that our aliased tensors do not have vectorized accesses larger than those of the original tensors. However, when we have different dtypes we `reinterpret_cast` it to a different `Array` type. Previously we did not specify any alignment in that type's template args, meaning it assumed an alignment of size 1. Since the actual addresses will all still be aligned this does not caused misaligned accesses at runtime. This PR sets the template arg for alignment to be that of the vectorized access width for the alias tensor, so that the compiler could hypothetically do some optimizations knowing the address is aligned.

### DIFF
--- a/csrc/codegen.cpp
+++ b/csrc/codegen.cpp
@@ -2964,10 +2964,13 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
       } else {
         indent() << "// Alias Allocation (changing dtype) - "
                  << alloc->memoryType() << "\n";
+        auto va = kernel_->summary().vectorized_accesses;
+        auto it = va.find(tv);
+        int64_t alias_alignment = it == va.end() ? 1 : it->second;
         indent() << "auto " << genVariableName(tv)
                  << " = *reinterpret_cast<Array<" << buffer_dtype << ", "
-                 << genInline(size) << ">*>(&" << genVariableName(alias_tv)
-                 << ");\n";
+                 << genInline(size) << ", " << alias_alignment << ">*>(&"
+                 << genVariableName(alias_tv) << ");\n";
         if (alloc->memoryType() == MemoryType::Local) {
           aligned_array_of_regs_.insert(tv);
         }


jacobhinkle:
!build

csrc/codegen.cpp
```diff
@@ -2964,10 +2964,13 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
       } else {
         indent() << "// Alias Allocation (changing dtype) - "
                  << alloc->memoryType() << "\n";
+        auto va = kernel_->summary().vectorized_accesses;
+        auto it = va.find(tv);
```
jacobhinkle, line 2968: We could instead use the original TV, whose vectorized accesses are guaranteed to be at least as large as this :man_shrugging:

naoyam: That may not work since the actual alignment size is `sizeof(buffer_dtype) * alias_alignment`. If we used the alignment of the original tensor, the alignment size would become larger if the size of the data type of the original tensor was smaller.

naoyam (approved):
LGTM

csrc/codegen.cpp
jacobhinkle: This is currently only used by "inner aliasing" in which the size of the dtypes must match, as well as the parallelization and shapes. I'll leave it though for the reason you stated since in the future we might possibly allow mismatched shapes and dtype sizes subject to some other constraint.

naoyam: OK, I see. I think what's more important here would be we should make `CudaCodeGen` as trivial as possible and try to minimize the coupling with the lowering passes. It's difficult to keep track of all the assumptions made in the prior phases of code translations. For this case particular, since the lowering passes the alignment size, that should be just used as is. If some other valid value should be used instead, that decision should be made by the lowering and be recorded in the kernel summary.

Issue #3068, opened 2024-09-30
Inlining RoPE-like patterns require an analysis akin to ComputeAtLogicalDomainMap


naoyam:
In RoPE-like fusions, where a domain is sliced and then padded back to the original domain, inlining seems to need to consider a constraint that is similar to the persistent constraint in normalization fusions. 

Simplified example:

```
t0: [i0]
t1 = t0 // [i1]
t2 = t1 // [i2]
t3 = t2[0:2] // [i3]
t4 = pad(t23 {0, i0 - 2}) // [i4]
t5 = t1 + t4 // [i5]
```
![Image](https://github.com/user-attachments/assets/7291afdd-4010-418e-8d16-a676079a00ee)

Here, `i2` cannot be inlined into `i3` since the extent of `i2` is larger than that of `i3`. That also means `i1` cannot be inlined into `i2` since if that's done, `i5` would be also pulled into the same inlined loop, which would then mean `i3` and `i4` would need to be pulled together. Since `i2` and `i3` cannot be inlined together, that inling pattern is invalid.

This is quite similar to the inlining constraint due to the reduction-broadcast pattern in normalization. I think ideally we should generalize the analysis to consider constraints like this case.

zasdfgbnm:
Should we also handle something like this:
```
t0: [i0]
t1 = t0 // [i1]
t2 = t1 // [i2]
t3 = t2 // [i3/2, 2]
t4 = t3 // [i4]
t5 = t1 + t4 // [i5]
```
we can not inline `t1` at 1, although `i2` is mapped with `i5`.

naoyam:
I think that's true in our current inlining system. In the original design of `computeAt`, however, the split of `t3` would be propagated across the fusion to make them inlinable. 

In general, I think that the analysis of inlinability needs to be a global analysis. `ComputeAtLogicalDomainMap` does that for the reduction-broadcast pattern, but that's not the only case that affects inlinability.

PR #3036, opened 2024-09-26
Generate loops based on the IdModel Loop graph (promotion)

naoyam:
This PR introduces an option to use loop promotion IDs when generating `ForLoop` IR nodes. Consequently, loop index variables are also maintained by `IdModel`, not by `ComputeAtMap`. 

This is required when loop domains are not directly derived from logical domains as the ComputeAtMap loop map does not support such loop domains.

For now, I decided to add yet another IdModel option (`IdModelEnableOption::Loop`) for more controlled, incremental progress. It is automatically enabled in most of the tests in `test_indexing_advanced.cpp`.

Major changes are:

- Concrete loop domain: it's either the promotion ID (when `IdModelEnableOption::Loop` is on) or the concrete ID in `ComputeAtMap`
- Loop index variable: it's either the variable managed by `IdModel` (when `IdModelEnableOption::Loop` is on) or `ComputeAtMap`

This is part of #2851.

PR #3086, opened 2024-10-02
Loop graph with set loop domain

naoyam:
Depends on #3036 and #3079
Part of #2851 

I'm still experimenting with IdModel setLoopDomain, but this PR at least
allows for correct loop graph and promotion maps to be generated as long
as no forwarding-like scheduling is involved in loop domains that are
not fully derived from logical domains.

The existing slice tests with setLoopDomain are extended with inlineMost
since it's now fully working.

Note that there is still an issue of inlining analysis (#3068).
Addressing it is not in the scope of this PR.


naoyam:
!build

zasdfgbnm:
@naoyam Could you resolve conflict? Looks like this PR contains diff from other PRs.

tests/cpp/test_resize.cpp
```diff
@@ -3681,6 +3681,9 @@ TEST_F(ResizeTest, SliceScheduledLikeProducer) {
 
   std::vector<int64_t> shape({100});
 
+  EnableOptionsGuard enable_options_guard;
```
naoyam, line 3684: Just moved up as this option needs to be enabled before using `inlineMost`

tests/cpp/test_resize.cpp
naoyam, line 1: The only change in this file is effectively just adding `inlineMost`

naoyam:
!build

naoyam:
@zasdfgbnm This is ready for review

zasdfgbnm (approved):

PR #3088, opened 2024-10-03
Skip broadcast when computing TMA's expect bytes in circular buffering pass

zasdfgbnm:
Looks like instead of using the promoted loop, we should just skip broadcast completely.

### DIFF
--- a/csrc/device_lower/pass/circular_buffer.cpp
+++ b/csrc/device_lower/pass/circular_buffer.cpp
@@ -155,7 +155,7 @@ class CircularBufferLoopCloner : public kir::IrVisitor {
   }
 
   void handle(kir::IfThenElse* ite) final {
-    NVF_THROW("No IfThenElse should exist yet");
+    NVF_THROW("No IfThenElse should exist yet:\n", ite->toString());
   }
 
   void dispatch(Expr* expr) override {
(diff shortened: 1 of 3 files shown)


zasdfgbnm:
!build

rdspring1 (approved):
LGTM.

Rather than excluding broadcast, do you think we should only accept the `IterDomain` if it is serial or `ParallelType::Bulk`?

Also, isn't the extent of a broadcast dimension 1, so `expected_bytes` would stay the same. I suppose it could be expanded.

zasdfgbnm:
> Rather than excluding broadcast, do you think we should only accept the `IterDomain` if it is serial or `ParallelType::Bulk`?

Good question. I believe the behavior should be like this:
| IterType v \ ParallelType > | Serial | Bulk  | TID                   | BID, DID | Others |
|-----------------------------|--------|-------|-----------------------|----------|--------|
| Iterate                     | count  | count | error unless extent=1 | ignore   | error  |
| Broadcast                   | ignore | error | ignore                | ignore   | error  |
| Others                      | error  | error | error                 | error    | error  |

We only count Serial Iterate and Bulk Iterate. The most concerning missing check is Broadcast Bulk, which we will accept it and get undefined behavior, and it is easy to make mistakes like this (for example, `tv->axis(I_used_a_wrong_number_here)->parallelize(Bulk)`.

PR #3090, opened 2024-10-03
Allow to print null.

wujingyue:
This is useful when printing a vector of `Statement*` where some elements are null. For example,
https://github.com/NVIDIA/Fuser/blob/1aa6ed072d890d75b9f2114a2c00f918d4ab75b4/csrc/ops/composite.cpp#L80-L81.


wujingyue:
!build

naoyam (approved):

PR #3099, opened 2024-10-03
Vectorization validation fix

naoyam:
This is a follow-up PR to #3110.

Vectorized IDs may be found in loop domains that are not directly derived from logical domains, thus requiring the IdModel-based analysis.

All tests passed, so we could switch to this analysis entirely once IdModel is enabled by default.


naoyam:
!build

naoyam:
!build

csrc/device_lower/validation.cpp
```diff
@@ -288,8 +288,7 @@ class VectorizeValidator : public OptInDispatch {
     } else if (s->inner() == vectorized_id_) {
       vectorized_id_ = s->in();
     }
-    domains_.insert(s->outer());
-    domains_.insert(s->inner());
```
naoyam, line 292: I believe these two lines were wrong

naoyam:
!build

naoyam:
!build

csrc/device_lower/validation.cpp
```diff
@@ -361,6 +363,110 @@ class VectorizeValidator : public OptInDispatch {
...
+      const auto& outputs = dir == Direction::Forward
+          ? graph.outputGroups(expr_g)
+          : graph.inputGroups(expr_g);
+
+      if (expr->isOneOf<Swizzle, Swizzle2D>()) {
+        // Not supported
+        if (std::find(inputs.begin(), inputs.end(), cur_group) !=
+                inputs.end() ||
+            std::find(outputs.begin(), outputs.end(), cur_group) !=
+                outputs.end()) {
```
zasdfgbnm, line 413: Do we need to check outputs?

zasdfgbnm (approved):

csrc/device_lower/validation.cpp
naoyam: I don't think so, but that's what the existing code does (line 321)

naoyam:
!build

PR #2339, opened 2024-06-04
async copy save registers

liqiangxl:
**Issue** : Current CpAsync doesn't allow IfThenElse predicate, becuase for matmul we need to use inline predicate in cpAsync to ensure out of boundary positions are filled with 0. However, for non-matmul cases, out of boundary positions are not used and we can use the IfThenElse predicate.
**Fix** : Allow using IfThenElse predicate for CpAsync if there is no matmul in the fusion.
**Influence** : After this change, we can use CpAsync in normalization kernels with shared memory persistent. It reduces register usage/spills since it bypass L1/SMEM when copying from gmem to smem.

PR #3102, opened 2024-10-03
check shared memory predicate in matmul tests

liqiangxl:
Adding check to matmul tests since shared memory access in matmul kernels shouldn't be predicated.
Helps https://github.com/NVIDIA/Fuser/pull/2339

### DIFF
--- a/tests/cpp/utils.h
+++ b/tests/cpp/utils.h
@@ -176,10 +176,12 @@ class PredicatedChecker : public kir::IrVisitor {
     return checker.predicated_ite_;
   }
 
-  static bool isSharedMemoryPredicatedByIfThenElse(kir::Kernel* kernel) {
+  // If CpAsync from gmem to smem, then loaded from smem to registers using
+  // ldmatrix, then it is used in mma and should not use if-then-else predicate.
+  // If just CpAsync from gmem to smem, without further copy to register, then
+  // it is not used in mma and can use if-then-else predicate.
+  static bool isCpAsyncMmaPredicatedByIfThenElse(kir::Kernel* kernel) {
     for (auto tv : kernel->allTvs()) {
-      // If a tv is CpAsync copied to shared memory, then LdMatrix copied to
-      // registers, should use inline predicate instead of if-then-else
       if (tv->definition() != nullptr &&
           ir_utils::isCpAsyncOp(tv->definition())) {
         const auto& consumers = ir_utils::consumerTvsOf(tv);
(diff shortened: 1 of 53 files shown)


tests/cpp/utils.h
```diff
@@ -169,6 +169,22 @@ class PredicatedChecker : public kir::IrVisitor {
...
 
+  static bool isPredicatedByIfThenElse(
+      StmtNameType tv_name,
+      kir::Kernel* kernel) {
+    PredicatedChecker checker(tv_name, kernel->topLevelExprs());
+    return checker.predicated_ite_;
+  }
+  static bool isSharedMemoryPredicatedByIfThenElse(kir::Kernel* kernel) {
+    for (auto tv : kernel->allTvs()) {
+      if (tv->getMemoryType() == MemoryType::Shared &&
```
zasdfgbnm, line 180: Should we only check smem operands? smem epilogue does not use cp.async, so it should not have this limitation.

liqiangxl: changed to `If a tv is CpAsync copied to shared memory, then LdMatrix copied to registers, it is used in mma and should use inline predicate instead of if-then-else`

tests/cpp/utils.h
```diff
@@ -169,6 +169,31 @@ class PredicatedChecker : public kir::IrVisitor {
...
   }
 
+  static bool isPredicatedByIfThenElse(
+      StmtNameType tv_name,
+      kir::Kernel* kernel) {
+    PredicatedChecker checker(tv_name, kernel->topLevelExprs());
+    return checker.predicated_ite_;
+  }
+
+  static bool isSharedMemoryPredicatedByIfThenElse(kir::Kernel* kernel) {
```
zasdfgbnm, line 179: Should we update the name for the latest change?

tests/cpp/utils.h
```diff
@@ -169,6 +169,31 @@ class PredicatedChecker : public kir::IrVisitor {
...
+    for (auto tv : kernel->allTvs()) {
+      // If a tv is CpAsync copied to shared memory, then LdMatrix copied to
+      // registers, should use inline predicate instead of if-then-else
+      if (tv->definition() != nullptr &&
+          ir_utils::isCpAsyncOp(tv->definition())) {
+        const auto& consumers = ir_utils::consumerTvsOf(tv);
+        if (std::any_of(
+                consumers.begin(), consumers.end(), [&](TensorView* tv) {
+                  return ir_utils::isLdMatrixOp(tv->definition());
```
zasdfgbnm, line 188: Why checking ldmatrix?

tests/cpp/utils.h
liqiangxl: changed to `isCpAsyncMmaPredicatedByIfThenElse` and added comments.

zasdfgbnm (approved):
Could you also update the title of this PR?

liqiangxl:
!build

PR #3104, opened 2024-10-03
enable circular buffer

zasdfgbnm:


zasdfgbnm:
I added a `__nanosleep(100000);` to `mbarrier::wait`, then the failing test will go back to work again. So I can conclude that the failure is due to data race.

PR #3107, opened 2024-10-04
Scheduling loop domains

### LINKED ISSUE #3134: Utilities to generate loop domains
Currently manually created. Needs some helper tools
https://github.com/NVIDIA/Fuser/blob/8046359bcc226c613c140fa2762a49fe5c0d7fea/tests/cpp/test_resize.cpp#L4339-L4360

naoyam:
Part of #2851 
Closes #3134 

Adds a new scheduler utility function, `scheduler_utils::scheduleLoopDomainsLike(std::vector<TensorView*> tvs, std::vector<IterDomain*> reference_loop_domain)`. This works in some sense in a similar manner as the existing transform propagator but is more flexible as it does not assume the forward dependency from root/logical to loop domains. The primary motivation is to simplify setting up the loop domains of tensors in RoPE-like fusions. See, for example, the [`SliceSliceConcatConcat` test](https://github.com/NVIDIA/Fuser/blob/8046359bcc226c613c140fa2762a49fe5c0d7fea/tests/cpp/test_resize.cpp#L4131). With this PR, all of the manual calls of `setLoopDomain` are eliminated.

While the actual implementation resulted in a little convoluted, the underlying algorithm is quite simple. When setting up the loop domain of a given tensor, we first find a path from the reference loop domain to the root domain of the tensor (not logical domain. See https://github.com/NVIDIA/Fuser/pull/3107/files#diff-078826ec098dc190fe9aef215a8130a7f5d268dc36f9537b2bb9e6f255d589d6R2939-R2957). Then, new IDs are created by cloning the reference IDs and then the path is replayed using the new cloned IDs as the inputs. The new IDs are then set as the new loop domain of the tensor.

The above algorithm actually has one problem with path finding since the shortest ValGraph path may not be a valid path within a single TensorDomain. That's because while a ValGroup can have multiple definitions, that's still not allowed for IterDomain. To avoid this problem, I took a two-step traversal, where a backward-only traversal is first done from the reference domain and then a forward-only traversal is done to a target tensor. See https://github.com/NVIDIA/Fuser/pull/3107/files#diff-078826ec098dc190fe9aef215a8130a7f5d268dc36f9537b2bb9e6f255d589d6R2915 for more details.

To do forward or backward only traversals, I added an option to `BFS` to specify what directions can be taken. See the new test in test_bfs.cpp for a concrete example case.


naoyam:
!build

naoyam:
!build

csrc/device_lower/pass/loop_rotation.h
```diff
@@ -9,7 +9,6 @@
 
 #include <exceptions.h>
 #include <ir/all_nodes.h>
-#include <runtime/executor_params.h>
```
naoyam, line 12: Just unused

tests/cpp/test_bfs.cpp
naoyam, line 1: Just moved from the other files here, except for the last test

tests/cpp/test_bfs.cpp
```diff
@@ -0,0 +1,469 @@
...
+      {tv1->axis(0)});
+
+  std::vector<Val*> ref{
+      tv1->getLogicalDomain().at(1), tv1->getLogicalDomain().at(3)};
+
+  EXPECT_EQ(all_deps, ref);
+}
+
+// Test directed getExprsBetween
+TEST_F(BFSTest, TraversalDirection) {
```
naoyam, line 396: This is the only new test in this file

tests/cpp/test_gpu3.cpp
```diff
@@ -8819,159 +8819,6 @@ TEST_F(NVFuserTest, RAWSync) {
           "Producer is required to be in Global or Shared Memory based on parallelization strategy. RAW flags: (threadIdx.x)")));
 }
 
-// Testing IRBFS::getReachableValsFrom with a resize fusion
```
naoyam, line 8822: Moved

naoyam:
!build

csrc/scheduler/utils.cpp
```diff
@@ -2656,6 +2659,367 @@ void moveNonConcretizedBroadcastInnermost(
...
+
+    // Clone outputs if not found
+    for (const auto& output_g : output_groups) {
+      if (all_id_groups.has(output_g)) {
+        continue;
+      }
+
+      // No need to force exact mapping since this clone is going to
+      // be connected with tv
+      auto clone = output_g->front()->as<IterDomain>()->cloneWithoutRFactor();
```
zasdfgbnm, line 2904: Should we use this?
https://github.com/NVIDIA/Fuser/blob/2980c3c25999bbb94d39a79044fcbf968fa78b1f/csrc/id_model/schedule.h#L14-L19

csrc/scheduler/utils.cpp
```diff
@@ -2656,6 +2659,367 @@ void moveNonConcretizedBroadcastInnermost(
...
+// See LoopDomainSchedulingTest.ReshapeTraversalDirection for a
+// concrete example.
+ValGraphBFS::ExprPath LoopDomainScheduler::getReplayPath(TensorView* tv) const {
+  // Find the path to the root domain of the tensor. It is important
+  // to use the root domain if available since there can be multiple
+  // forward paths to the logical domain in the ValGraph. For example,
+  //
+  // t0 = [i0]
+  // t1 = reshape(t0, {i0}, [i0/4, 4})
+  // t2 = reshape(t1, {i0/4, 4], {i0})
```
zasdfgbnm, line 2945: nit:
```suggestion
  // t1 = reshape(t0, {i0}, {i0/4, 4})
  // t2 = reshape(t1, {i0/4, 4}, {i0})
```

zasdfgbnm (approved):
Excellent!

tests/cpp/test_loop_domain_scheduling.cpp
```diff
@@ -0,0 +1,248 @@
...
+  auto tv5 = set(tv4);
+  auto tv6 = reshape(tv5, {6, 2}, {12});
+
+  auto tv7 = reshape(tv1, {12}, {4, 3});
+  auto tv8 = reshape(tv7, {4, 3}, {12});
+
+  auto tv9 = add(tv6, tv8);
+  fusion.addOutput(tv9);
+
+  // Consider a case where the loop domain of tv4 is scheduled using
```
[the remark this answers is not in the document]
naoyam: That is correct. Fixed. Thanks.

csrc/scheduler/utils.cpp
naoyam: That's a good idea. Then, this should never pick reduction IDs over non-reduction IDs.

tests/cpp/test_loop_domain_scheduling.cpp
```diff
@@ -0,0 +1,248 @@
...
+// Iter domain cannot have multiple definitions, whereas ValGroup can.
+// This means that the traversal path in ValGraph may not be a valid
+// history to construct within a tensor domain. For example, a path
+// with a forward expr followed by a backward expr is invalid since it
+// requires the output of the forward expr to have both of the exprs
+// as its definitions. This test constructs a fusion where the
+// ValGraph shortest path of two tensors results in the invalid
+// pattern. The loop domain scheduler should be able to find a
+// non-shortest but valid path.
+TEST_F(LoopDomainSchedulingTest, ReshapeTraversalDirection) {
```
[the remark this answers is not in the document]
naoyam: Yes, it works. Added a new test: https://github.com/NVIDIA/Fuser/pull/3107/files#diff-add3baa66fa88dd28b1baec00ec023373d88630908bff6583a0a4d61379e17cbR250

naoyam:
!build

PR #3108, opened 2024-10-04
Fix detection of root domains

naoyam:
`Expr::definition() == nullptr` is no longer a valid check to detect root domains.


naoyam:
!build --diff

zasdfgbnm (approved):

PR #3110, opened 2024-10-04
A little code refactoring and fix

naoyam:
Extracted from #3099

### DIFF
--- a/csrc/device_lower/validation.cpp
+++ b/csrc/device_lower/validation.cpp
@@ -20,7 +20,6 @@
 #include <type.h>
 
 #include <ATen/cuda/CUDAContext.h>
-#include <limits>
 
 namespace nvfuser {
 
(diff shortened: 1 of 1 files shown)


csrc/device_lower/validation.cpp
```diff
@@ -198,13 +197,13 @@ namespace {
 // Check contiguity for all allocation domains associated with Misaligned
 // Vectorize ParallelType
 void checkContiguity(
-    const std::unordered_set<IterDomain*>& domains,
+    const std::unordered_set<IterDomain*>& dep_alloc_ids,
```
naoyam, line 200: Previously, `domains` include all IDs visited from the vectorized loop ID to the corresponding allocation IDs. Since only the allocation IDs are used here, `dep_alloc_ids` only includes the allocation IDs that are reached from the vectorized loop ID.

csrc/device_lower/validation.cpp
```diff
@@ -288,8 +288,7 @@ class VectorizeValidator : public OptInDispatch {
     } else if (s->inner() == vectorized_id_) {
       vectorized_id_ = s->in();
     }
-    domains_.insert(s->outer());
-    domains_.insert(s->inner());
+    domains_.insert(s->in());
```
naoyam, line 291: I believe this was a bug

csrc/device_lower/validation.cpp
```diff
@@ -353,19 +351,32 @@ class VectorizeValidator : public OptInDispatch {
...
+    for (auto alloc : tv->getMaybeAllocationDomain()) {
+      if (validator.domains_.find(alloc) != validator.domains_.end()) {
+        dep_alloc_ids.emplace(alloc);
+      }
+    }
+
+    return {validator.vectorized_id_, dep_alloc_ids};
+  }
+
+  static void validateAllocationVectorizedId(
```
naoyam, line 364: Separated finding the relevant allocation IDs and validating them. The first part will be also extended with IdModel in #3099

naoyam:
!build --diff

zasdfgbnm (approved):
I am wondering if we can just replace this entire thing with `proveLinearAndGetStride`

naoyam:
The binary test error should be fixed by #3111

naoyam:
> I am wondering if we can just replace this entire thing with `proveLinearAndGetStride`

I think so.

PR #3113, opened 2024-10-04
save cuda file when kernel line info is enabled

liqiangxl:
1. When profiling with `ncu --import-source yes`, it's necessary to enable `NVFUSER_DUMP=cuda_to_file` to ensure the correct source code is loaded into the profile results.
2. This PR automatically dumps the CUDA code to a file when `NVFUSER_ENABLE=kernel_lineinfo` is enabled, as this option is commonly used in debugging or profiling scenarios.
3. Additionally, the code responsible for dumping the CUDA source has been moved outside of getStructuredCode(), ensuring that the correct file name is saved when the kernel is executed with `NVFUSER_EXTERNAL_SRC`.
4. These changes are for convenience purpose.

### DIFF
--- a/csrc/runtime/executor.cpp
+++ b/csrc/runtime/executor.cpp
@@ -379,7 +379,7 @@ void FusionExecutor::compileFusion(
     file_name << "__tmp_kernel_" << kernel_id_ << ".cu";
     debug() << "PRINTING: " << file_name.str() << std::endl;
     std::ofstream out(file_name.str());
-    out << code << std::endl;
+    out << structured_code << std::endl;
     out.close();
   }
 
--- a/csrc/runtime/executor.cpp
+++ b/csrc/runtime/executor.cpp
@@ -169,14 +169,6 @@ std::string FusionExecutor::getStructuredCode(
             << " =======\n\n"
             << code << "\n======================================\n\n";
   }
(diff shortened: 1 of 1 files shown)


liqiangxl:
!build

naoyam:
I think we used to do this but disabled it as it was confusing. Check some of the recent PRs.

liqiangxl:
> I think we used to do this but disabled it as it was confusing. Check some of the recent PRs.

this one https://github.com/NVIDIA/Fuser/pull/2926? Looks like were disscussing about debug/profile tags.
Here the key is dump the file, then `ncu` can pick and save the correct source file. I met a scenario where I have a `__tmp_kernel_xxx.cu` file in my folder, then I run `ncu --import-source yes` , it saves the existing old file instead of the actual generated kernel .

liqiangxl:
Another scenario is if I ran with `NVFUSER_EXTERNAL_SRC=a.cu ncu --import-source yes`, ncu will complain there is no file called `__tmp_kernel_xxx.cu`. Still needs to run with `NVFUSER_DUMP=cuda_to_file`.

I am fine without these changes and just needs to remember always use `NVFUSER_DUMP=cuda_to_file ncu --import-source yes`

naoyam:
https://github.com/NVIDIA/Fuser/pull/2926#discussion_r1752182124

PR #3073, opened 2024-10-01
Allow linear to take a >2D weight and a >1D bias.

wujingyue:
As long as the extra dimensions are DID-parallel.

This allows a distributed transformer layer to use linear (instead of matmul+add) for speed and will simplify the pending #3045. 

To avoid ambiguity, this PR also removes the support for 1D weight and 0D bias; otherwise, it's unclear whether a 2D weight is one device dimension plus a non-device or two non-devices. This support can be added back by changing the thunder-to-nvFuser bridge to convert a 1D/0D linear to unsqueeze followed by a 2D/1D linear followed by a squeeze.

PR #3114, opened 2024-10-04
Fix fd.ops.linear to compute the right output rank.

wujingyue:
This PR fixes a bug introduced in #3073. This bug causes `nvFuser.Tensor` to have a different rank than the corresponding `TensorView`. This didn't trigger any test failure until I wrote a more complicated test that `slice`s the output of a linear. 

Question for @rdspring1 and/or @kevinstephano: shouldn't this bug be caught earlier? I guess when the Python frontend finalizes the definition it should have checked the output `nvFuser.Tensor`s are consistent with the output `TensorView`s. Wdyt?

### DIFF
--- a/csrc/ops/composite.cpp
+++ b/csrc/ops/composite.cpp
@@ -72,8 +72,10 @@ TensorView* newForLinear(
   bool k_bcast = input_domain.back()->isBroadcast();
   size_t red_dims = k_bcast ? 0 : 1;
 
-  // Linear: a = {*, in_features}, b = {out_features, in_features} /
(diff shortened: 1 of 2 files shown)


wujingyue:
!build

rdspring1 (approved):
LGTM.

To compare a `Tensor` and its `TensorView`, you'd have to construct the cpp `Fusion` by calling `fd.execute`. Then, for each tensor, get its `TensorView` via `getFusionState(size_t index)`, then compare their dimensions.

PR #3115, opened 2024-10-05
Shard MHA. 

wujingyue:
As a follow-up to #3045. 

For #2199.

### DIFF
--- a/tests/python/test_multidevice.py
+++ b/tests/python/test_multidevice.py
@@ -502,7 +502,7 @@ def test_transformer_forward(mpi_test):
     d = mpi_test.size
     rank = mpi_test.rank
 
-    b, s, h, e = 2, 2048, 96, 12288
+    b, s, h, e = 1, 2048, 96, 12288
 
     assert (
         e % h == 0
--- a/tests/python/test_multidevice.py
+++ b/tests/python/test_multidevice.py
@@ -234,17 +234,17 @@ class TransformerForwardFusion(FusionDefinition):
             dtype=DataType.BFloat16,
         )
         self.mha_linear0_weight = self.define_tensor(
-            shape=[e * 3, e],
+            shape=[d, e * 3 // d, e],
             contiguity=True,
             dtype=DataType.BFloat16,
         )
         self.mha_linear0_bias = self.define_tensor(
-            shape=[e * 3],
+            shape=[d, e * 3 // d],
             contiguity=True,
             dtype=DataType.BFloat16,
         )
         self.mha_linear1_weight = self.define_tensor(
-            shape=[e, e],
+            shape=[d, e, e // d],
             contiguity=True,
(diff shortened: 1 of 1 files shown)


wujingyue:
!build

wujingyue:
!build

tests/python/test_multidevice.py
```diff
@@ -332,35 +332,49 @@ def definition(self) -> None:
...
+        #                   [d,b,s,e/d]        [d,e,e/d]                 [e]
+        T152_local_matmul = self.ops.matmul(
+            T151,
+            self.ops.broadcast_in_dim(
+                self.ops.permute(self.mha_linear1_weight, [0, 2, 1]),
+                [d, 1, e // d, e],
+                [0, 2, 3],
+            ),
+        )
+        T152_matmul = self.ops.sum(T152_local_matmul, [0])  # allreduce
```
samnordmann, line 372: I wonder what would happen currently if we do not decompose the matmul and the allreduce...

tests/python/test_multidevice.py
```diff
@@ -332,35 +332,49 @@ def definition(self) -> None:
...
+            start_indices=[0, 0, 0, e * 2 // d],
+            end_indices=[d, b, s, e * 3 // d],
         )
-        T123 = self.ops.reshape(T104, new_shape=[b, s, h, e // h])
-        T124 = self.ops.permute(T123, dims=[0, 2, 1, 3])
-        T130 = self.ops.reshape(T91, new_shape=[b, s, h, e // h])
-        T131 = self.ops.permute(T130, dims=[0, 2, 1, 3])
-        T137 = self.ops.reshape(T117, new_shape=[b, s, h, e // h])
-        T138 = self.ops.permute(T137, dims=[0, 2, 1, 3])
+        T123 = self.ops.reshape(T104, new_shape=[d, b, s, h // d, e // h])
```
samnordmann, line 348: IIUC, we pass from shape `[d, b, s, e//d]` to `[d, b, s, h//d, e//h]`. Nothing illegal about it but it looks surprising to me so I just want to make sure

samnordmann (approved):
LGTM

tests/python/test_multidevice.py
wujingyue: I double checked -- it looks right. MHA is head parallel according to Figure 3b in https://arxiv.org/pdf/1909.08053.

tests/python/test_multidevice.py
wujingyue: The first thing that'll break is that `linear` will produce a wrong shape. linear, as is implemented today, will output a tensor of rank `input_rank + weight_rank - 2 = 5`. However, we want the shape to be `[d,b,s,e]` and thus 4D.

wujingyue:
!build

PR #3116, opened 2024-10-05
Fix sdpfa_fwd to not assume the presence of DIDs.

wujingyue:
Similar to #3073, `sdpfa_fwd` shouldn't assume DIDs are available at definition time. Instead, treat extra preceding dimensions as batch at definition time and check they are device parallel at evaluation time. 

This is required to land #3115.

### DIFF
--- a/csrc/ops/composite.cpp
+++ b/csrc/ops/composite.cpp
@@ -455,8 +455,9 @@ SdpfaFwdResult sdpfa_fwd(
       " vs ",
       value_domain);
   NVF_CHECK(
-      query_domain.size() >= 4,
-      "Expect Q/K/V to be at least 4D: ",
+      query_domain.size() == 4 || query_domain.size() == 5,
+      "Expect Q/K/V to be either 4D or 5D. If 5D, the first dimension is "
+      "expected to be device parallel during expression evaluation: ",
       query_domain);
 
   NVF_CHECK(
--- a/csrc/ops/composite.cpp
+++ b/csrc/ops/composite.cpp
@@ -444,35 +444,20 @@ SdpfaFwdResult sdpfa_fwd(
   auto query_domain = TensorDomain::noReductions(query->getLogicalDomain());
   auto key_domain = TensorDomain::noReductions(key->getLogicalDomain());
   auto value_domain = TensorDomain::noReductions(value->getLogicalDomain());
-
-  // Temporary handling of DID parallelization see
-  // https://github.com/NVIDIA/Fuser/issues/2563
-  bool has_device_dim = (query_domain.size() == 5);
-  if (has_device_dim) {
-    NVF_CHECK(
-        query_domain[0]->isDeviceDim(),
-        "Only support DID parallelization on outermost axis");
-    NVF_CHECK(
-        key_domain[0]->isDeviceDim(),
-        "Only support DID parallelization on outermost axis");
-    NVF_CHECK(
-        value_domain[0]->isDeviceDim(),
-        "Only support DID parallelization on outermost axis");
-  }
-
-  auto concrete_query_size = TensorDomain::noDevices(query_domain).size();
-  auto concrete_key_size = TensorDomain::noDevices(key_domain).size();
-  auto concrete_value_size = TensorDomain::noDevices(value_domain).size();
-
   NVF_CHECK(
-      concrete_query_size == 4 && concrete_key_size == 4 &&
-          concrete_value_size == 4,
-      "Expected query, key, and value to be 4D but got: ",
-      concrete_query_size,
-      " ",
-      concrete_key_size,
-      " ,and ",
-      concrete_value_size);
+      query_domain.size() == key_domain.size(),
+      query_domain,
+      " vs ",
+      key_domain);
+  NVF_CHECK(
+      query_domain.size() == value_domain.size(),
+      query_domain,
+      " vs ",
+      value_domain);
+  NVF_CHECK(
+      query_domain.size() >= 4,
+      "Expect Q/K/V to be at least 4D: ",
+      query_domain);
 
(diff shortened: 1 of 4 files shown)


wujingyue:
!build

wujingyue:
!build

csrc/ops/composite.cpp
```diff
@@ -444,35 +444,20 @@ SdpfaFwdResult sdpfa_fwd(
...
+      query_domain,
+      " vs ",
+      key_domain);
+  NVF_CHECK(
+      query_domain.size() == value_domain.size(),
+      query_domain,
+      " vs ",
+      value_domain);
+  NVF_CHECK(
+      query_domain.size() >= 4,
```
samnordmann, line 458: should we also check that it is `<=5` ?

tests/python/test_multidevice.py
```diff
@@ -112,3 +114,81 @@ def multidevice_schedule(self):
...
+    d, b, s, a, h = mpi_test.size, 2, 1024, 12, 768
+
+    if a % d != 0:
+        pytest.skip(f"We only support even split, so {a} has to be divisible by {d}.")
+
+    class Model(FusionDefinition):
+        def definition(self) -> None:
+            self.q = self.define_tensor(
+                shape=[d, -1, -1, -1, -1],
+                contiguity=[True if d > 1 else None, True, True, True, True],
```
samnordmann, line 134: why do we need this conditional here?

tests/python/test_multidevice.py
```diff
@@ -112,3 +114,81 @@ def multidevice_schedule(self):
...
+            attn = sdpa_result[0]
+            self.add_output(attn)
+
+        def multidevice_schedule(self) -> None:
+            mesh = self.sched._create_device_mesh(range(d))
+            for t in [self.q, self.k, self.v]:
+                self.sched._set_device_mesh(t, mesh)
+                self.sched.parallelize(t, 0, nvfuser.ParallelType.mesh_x)
+
+    torch.cuda.set_device(mpi_test.local_rank)
```
samnordmann, line 161: is it needed?

tests/python/test_multidevice.py
```diff
@@ -112,3 +114,81 @@ def multidevice_schedule(self):
...
+    torch.manual_seed(0)
```
samnordmann, line 162: shouldn't we make the seed manual for all tests, and put this line in the test's constructor ?

tests/python/test_multidevice.py
```diff
@@ -112,3 +114,81 @@ def multidevice_schedule(self):
...
+    q, k, v = [
+        torch.randn(b, a, s, h // a, dtype=torch.bfloat16, device="cuda")
+        for _ in range(3)
+    ]
+
+    with torch.random.fork_rng(
```
samnordmann, line 168: I might be missing some subtle considerations with the RNG, but would like to understand it better. Why is this fork needed here?

samnordmann (approved):
LTGM

tests/python/test_multidevice.py
wujingyue: https://github.com/NVIDIA/Fuser/blob/61a77e0a64d5bc446ba1c009f04a19204a28eab2/csrc/ir/nodes.cpp#L2922-L2925 would fail without this. 

However, I agree this makes `define_tensor` a bit cumbersome to use. I guess this hasn't been too annoying because `define_tensor` is usually called by Thunder-generated code and/or called with a constant shape. @jjsjann123 What do you think about this?

tests/python/test_multidevice.py
wujingyue: Yes and cc @Priya2698 who briefly mentioned how this should be done which I forgot :) We do have https://github.com/NVIDIA/Fuser/blob/61a77e0a64d5bc446ba1c009f04a19204a28eab2/tests/python/utils.py#L387 a "fixture" class used by many Python tests, so we could set seed there. However, I'm under the impression that we are migrating from unittest to pytest, so this will have to be done differently.

tests/python/test_multidevice.py
wujingyue: I did this so q, k, and v are created on a wrong device. But I'm definitely open to better ways of doing this.

tests/python/test_multidevice.py
wujingyue: Removed. It's not needed after I set dropout_p to 0. It would be needed otherwise to ensure the dropout in nvFuser's implementation and the reference use the same seed.

wujingyue: Thanks for catching that!

wujingyue:
!build

tests/python/test_multidevice.py
wujingyue: https://github.com/NVIDIA/Fuser/pull/3127

tests/python/test_multidevice.py
jjsjann123: Yeah this is indeed unwieldy. I think we can add another API replacing `std::vector<std::optional<bool>> contiguity` with `bool default_contiguous`.

Alternatively, currently behavior is that `if contiguity.empty()`, we default to have everything as non-contiguous. We can change this to have everything as contiguous and you wouldn't need to provide a contiguity arg to this function. 

But I'm not a big fan of altering existing behavior. Leaning towards adding another API.

tests/python/test_multidevice.py
Priya2698: We are using this with `pytest`: https://github.com/NVIDIA/Fuser/blob/e01e23ef2f3b4d49e05a5e6a1f9c81119385d537/tests/python/test_python_frontend.py#L41-L42.

What do you mean by doing it differently for pytest?

PR #3126, opened 2024-10-07
Line numbers have been incorrect

naoyam:
Just noticed the line numbers of the trace messages are off by one. ~I think it's good enough to just check the existence of the file name with `:`.~

`DEBUG_PRINT_SCOPE`, `DEBUG_LOG` and `RECORD_AND_RETURN` are now enabled with Release.


naoyam:
!build

wujingyue:
The test was able to pass despite wrong line numbers? Isn't that an issue with the test?

naoyam:
We haven't seen any error since they are only run with Debug build.

@zasdfgbnm I believe these tests were added by you. Why are they skipped with Release?

wujingyue:
> We haven't seen any error since they are only run with Debug build.

Indeed. https://github.com/NVIDIA/Fuser/blob/33514147cd868519508ef80b07cc450c6c3e278d/csrc/utils.h#L446

wujingyue:
> I think it's good enough to just check the existence of the file name with :.

I'd keep the line numbers. Printing correct line numbers is important for debugging and therefore worth testing. I think the real problem is that these tests aren't enabled by default. If they had been enabled, it would have been easy to catch and fix any line number mismatches.

zasdfgbnm:
> > I think it's good enough to just check the existence of the file name with :.
> 
> I'd keep the line numbers. Printing correct line numbers is important for debugging and therefore worth testing. I think the real problem is that these tests aren't enabled by default. If they had been enabled, it would have been easy to catch and fix any line number mismatches.

Agreed. I would recommend just globally enable `DEBUG_PRINT_SCOPE_NAME`. I remember tested that the perf impact is negilible.

naoyam:
!build

zasdfgbnm (approved):

naoyam:
> > > I think it's good enough to just check the existence of the file name with :.
> > 
> > 
> > I'd keep the line numbers. Printing correct line numbers is important for debugging and therefore worth testing. I think the real problem is that these tests aren't enabled by default. If they had been enabled, it would have been easy to catch and fix any line number mismatches.
> 
> Agreed. I would recommend just globally enable `DEBUG_PRINT_SCOPE_NAME`. I remember tested that the perf impact is negilible.

Done. I also did quick cleanups.

PR #3127, opened 2024-10-07
Call torch.cuda.set_device in the mpi_test fixture.

wujingyue:


wujingyue:
!build

samnordmann:
Why not calling this the cpp `Communicator`'s constructor?

wujingyue:
> Why not calling this in the cpp `Communicator`'s constructor?

Maybe later in a separate PR. I wanted to keep blast radius small especially for something that changes a global state. Calling torch.cuda.set_device in Communicator affects the behavior of all tensor factory APIs inside and **outside** nvFuser. 

Also, I'm waiting for a repro from @cowanmeg. I believe she saw nvFuser didn't synchronize the right device. Understanding that will give me more confidence about the potential changes in nvFuser internals.

samnordmann:
> > Why not calling this in the cpp `Communicator`'s constructor?
> 
> Maybe later in a separate PR. I wanted to keep blast radius small especially for something that changes a global state.

OK, np

>Calling torch.cuda.set_device in Communicator affects the behavior of all tensor factory APIs inside and **outside** nvFuser.

Right, I didnt realized the "outside", but we need to be careful. However, the present patch also will change the current device, also outside nvFuser.

> Also, I'm waiting for a repro from @cowanmeg. I believe she saw nvFuser didn't synchronize the right device. Understanding that will give me more confidence about the potential changes in nvFuser internals.


Ok. FWIW, what we can observe if calling a barrier as first collective: 
```
[W1008 11:01:30.790625840 ProcessGroupNCCL.cpp:4121] [PG ID 0 PG GUID  Rank 7]  using GPU 7 to perform barrier as devices used by this process are currently unknown. This can potentially cause a hang if this rank to GPU mapping is incorrect.Specify device_ids in barrier() to force use of a particular device,or call init_process_group() with a device_id
```

wujingyue:
> However, the present patch also will change the current device, also outside nvFuser.

I don't think so -- this PR changed only the test fixture. Unless code outside runs nvFuser's tests, but why would they?

samnordmann:
> Unless code outside runs nvFuser's tests

I was thinking of this case, but nevermind!

samnordmann (commented):
From the comment I understand that this patch aims at setting the default device on which the tensors are allocated. However, calling `set_device` does way more than that, and might impact the whole behaviour. Hence I am not sure this the right way to achieve the goal.

Having a different logic/behaviour for this test specifically doesn't sound ideal... This refers to [this comment](https://github.com/NVIDIA/Fuser/pull/3143#pullrequestreview-2356909013).

PR #3131, opened 2024-10-08
Support using stmatrix to store any shape in registers by tiling to multiple calls to stmatrix for 8x8, 16x8 and 16x16 tiles.

protonu:
This PR adds support to store any 2D piece of memory in registers to shared memory using stmatrix.
We tile the 2D memory in registers and store it using multiple calls to stmatrix inside a for-loop. 

For example the 2D memory [32, 32] can be tiled to 16 stmatrix calls (.x1 8x8) or 8 calls (.x2 16x8) or 4 calls (.x4 16x16).
In this PR, we start with a 2D memory say [32, 64], tile it to [(2, 16), (4, 8) ] -> [4, 2, 16, 8] -> [8, 16, 8] -- That is 8 calls to stmatrix of type .x2 which stores 16x8 tiles.

For the input [32, 64] once we have transformed to [8, 16, 8], we swizzle the memory in registers. We then parallelize it such that that inner-most dims (after merging) is vectorized (with vector length being the same as the elements being store per thread).

indexing

After swizzling and parallelizing we generate indices for the output of the stmatrix instruction in a hardcoded manner. The indexing logic has two parts.

1. Computing the index of the tile.
indexing

2. Computing the index inside the tile. The final index is the sum of the two.
inner-tile

### DIFF
--- a/csrc/device_lower/pass/index.cpp
+++ b/csrc/device_lower/pass/index.cpp
@@ -1581,62 +1581,112 @@ static inline DataType getMmaOutType(TensorView* mma_out) {
 }
 
 namespace {
-std::pair<Val*, Val*> hardCodedIndexGenerationForStMatrix(
+// Helper function to compute the index of the output of stmatrix (in shared
+// memory). We have two special cases (1) Handle tiles of size 8x8 and 16x8 and
+// (2) tiles of size 16x16. Each of these two cases has further two special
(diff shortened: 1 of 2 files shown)


protonu:
!build

protonu:
!build

zasdfgbnm (approved):

PR #3135, opened 2024-10-08
Update `HopperMatmulTest.HSH_NT_128BSwizzle`

zasdfgbnm:
Previously, this test was assuming that, each CTA only has a single tile of MMA. This PR removes this assumption so that it is more convenient to try different things.


zasdfgbnm:
!build

naoyam (approved):

PR #3136, opened 2024-10-08
Don't use `fence.proxy.async`

zasdfgbnm:
According to: https://docs.nvidia.com/cuda/parallel-thread-execution/#async-proxy, mbarrier wait already guaranteed correct ordering, so this sync is not needed:

> The completion of a `cp{.reduce}.async.bulk` operation is followed by an implicit generic-async proxy fence. So the result of the asynchronous operation is made visible to the generic proxy as soon as its completion is observed. Async-group OR mbarrier-based completion mechanism must be used to wait for the completion of the `cp{.reduce}.async.bulk` instructions.


zasdfgbnm:
!build

naoyam (approved):
Almost 50%!

zasdfgbnm:
!build

zasdfgbnm:
!build

PR #3139, opened 2024-10-08
Only do `wgmma.fence.sync.aligned` once

zasdfgbnm:
Currently, the code for Hopper matmul kernel looks like below
```C++
accumulator = 0
for (prologue) {
  TMA load A;
  TMA load B;
}
for (main) {
  TMA load A;
  TMA load B;
  mbarrier wait;
  wgmma.fence;
  wgmma;
}
for (epilogue) {
  mbarrier wait;
  wgmma.fence;
  wgmma;
}
```
where we do the fencing at  every iteration. This is not optimal because, according to https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#asynchronous-warpgroup-level-matrix-instructions-wgmma-fence:

> The wgmma.fence instruction must be issued by all warps of the warpgroup at the following locations:
> 
> - Before the first `wgmma.mma_async` operation in a warpgroup.
>
> - Between a register access by a thread in the warpgroup and any `wgmma.mma_async` instruction that accesses the same registers, either as accumulator or input register containing fragments of matrix A, except when these are accumulator register accesses across multiple `wgmma.mma_async` instructions of the same shape. In the latter case, an ordering guarantee is provided by default.

Because the order of `wgmma.mma_async` instructions of the same shape is automatically guaranteed, there is no need to do a fence at every iteration.

This PR changes the generated code into:
```C++
accumulator = 0
wgmma.fence;
for (prologue) {
  TMA load A;
  TMA load B;
}
for (main) {
  TMA load A;
  TMA load B;
  mbarrier wait;
  wgmma;
}
for (epilogue) {
  mbarrier wait;
  wgmma;
}
```
which in theory should be better.

There is no visible perf change for this PR.

### DIFF
--- a/csrc/device_lower/pass/allocation.cpp
+++ b/csrc/device_lower/pass/allocation.cpp
@@ -567,6 +567,20 @@ class AllocationInserter : public kir::ExprMutator {
             ? nullptr
             : &allocation.init_for_loop->body();
         registerInsertBefore(allocation.init_place_before, init_expr, scope);
+
+        // Needs to do a wgmma.fence.sync.aligned so that the initial values of
+        // the accumulator are visible to TensorCore.
+        if (auto mma = dynamic_cast<MmaOp*>(expr)) {
+          if (isHopper(mma->macro())) {
+            auto wgmma_fence = IrBuilder::create<kir::Asm>(
+                "wgmma.fence.sync.aligned",
+                std::vector<Val*>{},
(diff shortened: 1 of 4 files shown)


zasdfgbnm:
!build

zasdfgbnm:
!build

zasdfgbnm:
!build

naoyam (approved):

PR #3142, opened 2024-10-09
Add expand operation to python frontend

rdspring1:
* Currently, the python frontend represents `broadcast_in_dims` as a `broadcast` and `expand` composite operation. 
* For cpp to python translation, when we encounter an `Expand` expression, we must create a `BroadcastInDimsOp`.
* This `BroadcastInDimsOp` creates a redundant `broadcast` because the original `broadcast` was handled separately from the `expand`. 
* By adding the `expand` operation to the python frontend, we can create a one-to-one mapping in the cpp to python translation.


rdspring1:
!build

jjsjann123 (approved):
LGTM.

PR #3143, opened 2024-10-09
Call init/destroy_process_group once. 

### LINKED ISSUE #3129: test_transformer_engine:test_transformer_layer[backward] hangs.
It has been disabled by https://github.com/NVIDIA/Fuser/pull/3119, but should be fixed properly.

wujingyue:
Fixes #3129. 

This is done by using fixtures. 

It's unclear why this avoids #3129. However, this appears to be the best practice according to https://pytorch.org/docs/stable/distributed.html#shutdown, and is more efficient.

### DIFF
--- a/tests/python/mpi_fixtures.py
+++ b/tests/python/mpi_fixtures.py
@@ -34,7 +34,7 @@ class MpiTest:
         self._communicator.barrier()
 
 
-@pytest.fixture
+@pytest.fixture(scope="session")
 def mpi_test():
     fixture = MpiTest()
     yield fixture
--- a/tests/python/test_transformer_engine.py
+++ b/tests/python/test_transformer_engine.py
@@ -2,7 +2,6 @@
 # All rights reserved.
 # SPDX-License-Identifier: BSD-3-Clause
 
-import os
 import pytest
 import torch
 import torch.distributed as dist
--- a/tests/python/test_transformer_engine.py
+++ b/tests/python/test_transformer_engine.py
@@ -50,7 +50,7 @@ def setup_process_group(mpi_test) -> None:
     [ComputeType.FORWARD, ComputeType.BACKWARD],
     ids=["forward", "backward"],
 )
-def test_transformer_layer(mpi_test, setup_process_group, benchmark, compute_type):
+def test_transformer_layer(setup_process_group, benchmark, compute_type):
     # Hyperparameters for GPT-3
     hidden_size = 12288
     num_heads = 96
@@ -59,8 +59,8 @@ def test_transformer_layer(mpi_test, setup_process_group, benchmark, compute_typ
     sequence_length = 2048
     dtype = torch.bfloat16
 
-    size = mpi_test.size
-    rank = mpi_test.rank
+    size = dist.get_world_size()
+    rank = dist.get_rank()
 
     torch.cuda.set_device(rank)
 
--- a/tests/python/test_transformer_engine.py
+++ b/tests/python/test_transformer_engine.py
@@ -25,12 +25,10 @@ class ComputeType(Enum):
 
 @pytest.fixture(scope="module")
 def setup_process_group(mpi_test) -> None:
-    os.environ["MASTER_ADDR"] = "localhost"
     # The default port as used by https://github.com/pytorch/pytorch/blob/45a8b5682eb69d865cbf68c7f2f689b56b4efd53/torch/csrc/distributed/c10d/TCPStore.hpp#L51.
-    os.environ["MASTER_PORT"] = "29500"
     dist.init_process_group(
         backend="nccl",
-        init_method="env://",
+        init_method="tcp://localhost:29500",
         world_size=mpi_test.size,
         rank=mpi_test.rank,
     )
--- a/tests/python/test_transformer_engine.py
+++ b/tests/python/test_transformer_engine.py
@@ -23,6 +23,21 @@ class ComputeType(Enum):
     BACKWARD = auto()
 
 
+@pytest.fixture(scope="module")
+def process_group(mpi_test) -> None:
+    os.environ["MASTER_ADDR"] = "localhost"
+    # The default port as used by https://github.com/pytorch/pytorch/blob/45a8b5682eb69d865cbf68c7f2f689b56b4efd53/torch/csrc/distributed/c10d/TCPStore.hpp#L51.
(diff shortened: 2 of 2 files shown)


tests/python/mpi_fixtures.py
```diff
@@ -34,7 +34,7 @@ def barrier(self):
         self._communicator.barrier()
 
 
-@pytest.fixture
+@pytest.fixture(scope="session")
```
wujingyue, line 37: Because setup_process_group, has module scope, uses mpi_test, mpi_test must have module or session scope.

wujingyue:
!build

tests/python/test_transformer_engine.py
```diff
@@ -51,15 +63,6 @@ def test_transformer_layer(mpi_test, benchmark, compute_type):
...
-    os.environ["MASTER_ADDR"] = "localhost"
-    # The default port as used by https://github.com/pytorch/pytorch/blob/45a8b5682eb69d865cbf68c7f2f689b56b4efd53/torch/csrc/distributed/c10d/TCPStore.hpp#L51.
-    os.environ["MASTER_PORT"] = "29500"
-    dist.init_process_group(
-        backend="nccl",
-        init_method="env://",
-        world_size=size,
-        rank=rank,
-    )
     tp_group = dist.new_group()
```
samnordmann, line 66: why do we need to create a new group here? The default pg should work as intended

tests/python/test_transformer_engine.py
```diff
@@ -23,6 +23,21 @@ class ComputeType(Enum):
     BACKWARD = auto()
 
 
+@pytest.fixture(scope="module")
+def setup_process_group(mpi_test) -> None:
```
samnordmann, line 27: maybe move to `mpi_fixtures.py`?

tests/python/test_transformer_engine.py
```diff
@@ -32,13 +47,10 @@ class ComputeType(Enum):
 @pytest.mark.mpi
 @pytest.mark.parametrize(
```
samnordmann, line 48: I am not sure but it might work to specify `scope=function` here and remove the scope arguments from the two fixtures

tests/python/test_transformer_engine.py
```diff
@@ -23,6 +23,21 @@ class ComputeType(Enum):
...
 
 
+@pytest.fixture(scope="module")
+def setup_process_group(mpi_test) -> None:
+    os.environ["MASTER_ADDR"] = "localhost"
+    # The default port as used by https://github.com/pytorch/pytorch/blob/45a8b5682eb69d865cbf68c7f2f689b56b4efd53/torch/csrc/distributed/c10d/TCPStore.hpp#L51.
+    os.environ["MASTER_PORT"] = "29500"
+    dist.init_process_group(
+        backend="nccl",
+        init_method="env://",
```
samnordmann, line 33: Address and Port should be read from env variables, like in: https://github.com/NVIDIA/Fuser/blob/d9cee3852dffab13bed7e4e1034f42971c5e2b0f/csrc/multidevice/communicator.cpp#L102

in particular, setting address as "localhost" excludes the possibility of running the test on a multi-node setup.

Btw, if, for some reason, you still want to use `localhost:29500`, then instead of overwrite env vars you can do:

```suggestion
    # The default port as used by https://github.com/pytorch/pytorch/blob/45a8b5682eb69d865cbf68c7f2f689b56b4efd53/torch/csrc/distributed/c10d/TCPStore.hpp#L51.
    dist.init_process_group(
        backend="nccl",
        init_method="tcp://localhost:29500",
```

tests/python/test_transformer_engine.py
```diff
@@ -51,15 +63,6 @@ def test_transformer_layer(mpi_test, benchmark, compute_type):
     rank = mpi_test.rank
```
samnordmann, line 63: `rank` and `size` should be retrieved from the pg backend

samnordmann (commented):
Why not using `Communicator` for those tests? The behavior of this test differs from all the other tests, which is confusing and add significant technical debt. The present present pr is an example of the technical debt incurred by managing another pg.

Using `Communicator` would reuse and exercise our infrastructure, have a cheaper overhead, and would uniformize the behavior of this test with the others.

For example, with the way the test is implemented today: we need OMPI to be installed, `mpirun` to be the launcher, cannot run it on a multi-node system, cannot set the port, etc.

wujingyue:
> Why not using Communicator for those tests? 

Good question! 

test_transformer_engine.py in particular is an exception. It serves as a performance baseline and doesn't need to depend on nvFuser (and thus `Communicator`). That lack of dependency helped me once when I gave this as a repro to the TE team. So, strictly speaking, test_transformer_engine.py doesn't even need to be in the nvFuser repo. I put it here because it's the repo we've been collaborating on and it's easy to refer to the baseline when it's in the same repo. 

If by "those tests" you mean other Python distributed tests, that's a good idea. I'm unsure how to implement this. We'd need to expose Communicator via nvFuser's Python API, and we'd need to figure out a way to easily filter in/out distributed/non-distributed tests. I'm happy to hear your thoughts on that.

tests/python/test_transformer_engine.py
wujingyue: You are right. I will need that when working on 2D sharding, but I'm happy to YAGNI for now.

tests/python/test_transformer_engine.py
wujingyue: I'm not sure I understand your suggestion. AFAIK, `scope` is an attribute of a fixture, not of a test.

tests/python/test_transformer_engine.py
wujingyue: TIL!

wujingyue:
!build

wujingyue:
!build

wujingyue:
I'm too busy with #2199 and the Communicator change will have to come after that. Nonetheless, this PR alone fixes a bug and restores the performance baseline, so I recommend we merge this and clean up the mpi_test fixture after October.

samnordmann:
> If by "those tests" you mean other Python distributed tests, that's a good idea. I'm running into some problems with mpi4py that can be solved by this.

I am actually talking about both this tests and other tests, but I agree (and didn't realized) that it is probably less important for the present tests.

samnordmann (approved):
sorry for the delay!

Issue #3137, opened 2024-10-08
Tracking perf optimization of `HopperMatmulTest.HSH_NT_128BSwizzle` for problem size `(M=16384, N=16384, K=1024)`, CTA tile size `(64, 256)`


zasdfgbnm:
Tracking the perf of a specific shape in this specific unit test using `nsys nvprof`.

TOT main branch (as measured [here](https://github.com/NVIDIA/Fuser/issues/3137#issuecomment-2428295264)):

Perf:

```C++
 Time (%)  Total Time (ns)  Instances  Avg (ns)   Med (ns)   Min (ns)  Max (ns)  StdDev (ns)                                                  Name

 --------  ---------------  ---------  ---------  ---------  --------  --------  -----------  ----------------------------------------------------------------------------------------------------
     19.1          1309169          1  1309169.0  1309169.0   1309169   1309169          0.0  <unnamed>::nvfuser_none_f0_c0_r0_g0(<unnamed>::Tensor<<unnamed>::__half, (int)3, (int)3>, <unnamed>…
     11.3           776119          1   776119.0   776119.0    776119    776119          0.0  nvjet_hsh_192x192_64x3_2x1_v_bz_coopB_NTN
```

Perf nvFuser/cuBLAS: `59.28%`.

zasdfgbnm:
Initial perf:
```C++
 Time (%)  Total Time (ns)  Instances  Avg (ns)   Med (ns)   Min (ns)  Max (ns)  StdDev (ns)                                                  Name

 --------  ---------------  ---------  ---------  ---------  --------  --------  -----------  ----------------------------------------------------------------------------------------------------
     34.9          2974662          1  2974662.0  2974662.0   2974662   2974662          0.0  <unnamed>::nvfuser_none_f0_c0_r0_g0(<unnamed>::Tensor<<unnamed>::__half, (int)3, (int)3>, <unnamed>…
      9.2           780729          1   780729.0   780729.0    780729    780729          0.0  nvjet_hsh_192x192_64x3_2x1_v_bz_coopB_NTN
```

Perf nvFuser/cuBLAS: `26%`

zasdfgbnm:
After https://github.com/NVIDIA/Fuser/pull/3136:

```C++
 Time (%)  Total Time (ns)  Instances  Avg (ns)   Med (ns)   Min (ns)  Max (ns)  StdDev (ns)                                                  Name

[...]

PR #3146, opened 2024-10-09
A few cleanups

zasdfgbnm:
- add new IR node `kir::FenceAsyncProxy` and `kir::WgMmaFence`, which will be lowered as `fence.proxy.async` and `wgmma.fence.sync.aligned` in `inline_ptx.cpp`. While it is possible to just create the `kir::Asm` nodes in the passes, I believe it's better to have a separate IR node so that other passes will not need to worry about details of the PTX instruction.
- In the past, syncs for TMA store and MMA are inserted in `index.cpp` and `inline_ptx.cpp`, which is not optimal. I moved the insertion code to `insert_syncs.cpp`.
- For Hopper MMA whose inputs are not guarded by mbarrier, insert fences before each mma instruction. These fences were removed in https://github.com/NVIDIA/Fuser/pull/3139 and https://github.com/NVIDIA/Fuser/pull/3136, but I believe we should add them back for this edge case, because there was no mbarrier guaranteeing the ordering. This is just an edge case, for real kernel, we will only use the SS variant and operands will always be guarded by mbarrier.

This is just a cleanup, there is no change in the perf of https://github.com/NVIDIA/Fuser/issues/3137

### DIFF
--- a/csrc/device_lower/pass/allocation.cpp
+++ b/csrc/device_lower/pass/allocation.cpp
@@ -568,16 +568,21 @@ class AllocationInserter : public kir::ExprMutator {
             : &allocation.init_for_loop->body();
         registerInsertBefore(allocation.init_place_before, init_expr, scope);
 
-        // Needs to do a wgmma.fence.sync.aligned so that the initial values of
-        // the accumulator are visible to TensorCore.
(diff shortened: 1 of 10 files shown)


naoyam (approved):
LGTM

zasdfgbnm:
!build

PR #3147, opened 2024-10-09
use static bdimx

liqiangxl:
1.05x to 1.10x speedup for bdimx padded cases, e.g. `7424 = 8 x 4 x 232 (padded to 256)` , see [dashboard](http://nv/elH).


liqiangxl:
!build --diff --pybench

PR #3150, opened 2024-10-10
Enable matmul operands with non-trivial alloc domains

jacobhinkle:
Previously we rejected matmul segments with non-trivial allocation domains. However, I believe we properly handle allocation domain now with the latest changes to the matmul scheduler. I am enabling this now so that we can run all the python benchmarks.

Note that we still reject fusions with non-trivial output allocation domains. That support is likely also working but requires more testing before we enable it I believe.

### DIFF
--- a/csrc/scheduler/runtime_info.cpp
+++ b/csrc/scheduler/runtime_info.cpp
@@ -62,18 +62,10 @@ SchedulerRuntimeInfo::SchedulerRuntimeInfo(
       if (alloc_perm_opt.has_value()) {
         // Save the strides in order of allocation domain in case the
         // allocation domain is a permutation of RFactor domain
-        std::vector<int64_t> orig_sizes = alloc_sizes.vec();
-        std::vector<int64_t> orig_strides = alloc_strides.vec();
-        std::vector<int64_t> ordered_sizes, ordered_strides;
-        ordered_sizes.reserve(orig_sizes.size());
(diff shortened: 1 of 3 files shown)


jacobhinkle:
!build

csrc/scheduler/runtime_info.cpp
```diff
@@ -62,18 +62,10 @@ SchedulerRuntimeInfo::SchedulerRuntimeInfo(
...
-        std::vector<int64_t> ordered_sizes, ordered_strides;
-        ordered_sizes.reserve(orig_sizes.size());
-        ordered_strides.reserve(orig_strides.size());
-        NVF_ERROR(orig_strides.size() == alloc_perm_opt->size());
-        for (int64_t i : alloc_perm_opt.value()) {
-          ordered_sizes.push_back(orig_sizes[i]);
-          ordered_strides.push_back(orig_strides[i]);
-        }
-        input_sizes_[fusion_inp] = std::move(ordered_sizes);
-        input_strides_elements_[fusion_inp] = std::move(ordered_strides);
```
jacobhinkle, line 76: Previously I thought we needed to permute the sizes and strides to match the allocation domain. However, it seems that these are already permuted when we grab them at line 52.

jacobhinkle:
!build

zasdfgbnm (approved):

PR #3151, opened 2024-10-10
Move scheduling-related files to scheduler/tools

naoyam:
No code change. Just moved utility-like files for scheduling to scheduler/tools. Specifically, these files are moved:

- inlining.h
- abstract_tensor.h
- maxinfo_propagator.h
- LoopDomainScheduler from scheduler/utils.h to scheduler/tools/loop_domain_scheduler.h

I also considered moving `id_model/schedule.h`, but it seems to define more generic operations, so didn't touch it.

Also added a namespace, `scheduler_tools`.


naoyam:
!build

zasdfgbnm (approved):

PR #3160, opened 2024-10-10
Remove some Hopper tests

zasdfgbnm:
There are lots of Hopper matmul tests. These tests comes from incremental development: every time we add something new feature, we add a new test for it. While having these tests made a lot of sense in incremental development, we do not have to keep all of them after things are developed. For example, if we have:
```
Test A: feature 1
Test B: feature 1 + feature 2
Test C: feature 1 + feature 2 + feature 3
```
While having all the above tests does help us quickly narrow down which part is wrong if failures happen, we do not have unlimited resources to run all of them. In this case, we can delete Test B because C already covers all the features, although C is not as unit as B. I would still keep A as it is very unit for feature 1.

The latency of Hopper CI is becoming extremely long recently. It is pretty common to wait >12h, even >24h to get a result. This is unacceptable. This PR removes some tests while still making sure that pieces covered by the removed tests are covered somewhere else.


tests/cpp/test_mma.cpp
```diff
@@ -400,209 +400,6 @@ TEST_P(HopperRS, SingleTile) {
...
-//   store accumulator to gmem;
-//
-// For TT layout where the inner dimension is not a reduction:
-//   load operand B from gmem to smem;
-//   for i in tiles:
-//     load operand A from gmem to register;
-//     accumulator = 0;
-//     accumulator += A * B;
-//     store accumulator to gmem;
-TEST_P(HopperRS, FullSwizzle) {
```
zasdfgbnm, line 431: Should be covered by `MultipleTile` test

tests/cpp/test_mma.cpp
```diff
@@ -697,91 +494,6 @@ TEST_P(HopperRS, SingleTileTransposed) {
   EXPECT_TRUE(at::allclose(cg_outputs[0], tref, 1e-5, 1e-5));
 }
 
-TEST_P(HopperRS, SingleTileWithTMALoad) {
-  Fusion fusion;
```
zasdfgbnm, line 701: We already have test that uses TMA for both load and store. There is no need for a test that only has load but not store.

tests/cpp/test_mma.cpp
```diff
@@ -1313,184 +1025,6 @@ TEST_P(HopperSS, SingleTileTransposed) {
   EXPECT_TRUE(at::allclose(cg_outputs[0], tref, 1e-5, 1e-5));
 }
 
-// See the note in HopperRS.FullSwizzle for the explanation of this test.
-TEST_P(HopperSS, FullSwizzle) {
```
zasdfgbnm, line 1317: Covered by `MultipleTile` tests

tests/cpp/test_mma.cpp
```diff
@@ -400,388 +400,6 @@ TEST_P(HopperRS, SingleTile) {
...
-      inputs.first.squeeze().to(at::kFloat),
-      inputs.second.squeeze().to(at::kFloat),
-      layout);
-  EXPECT_TRUE(at::allclose(cg_outputs[0], tref, 1e-5, 1e-5));
-}
-
-// Same as SingleTile, except that the core matrices of B are stored
-// in a transposed way in smem. This is useful for testing if we are
-// inferring strides of core matrices correctly.
-TEST_P(HopperRS, SingleTileTransposed) {
```
zasdfgbnm, line 609: Covered by `SS` variant of `SingleTileTransposed`

tests/cpp/test_mma.cpp
```diff
@@ -854,164 +472,23 @@ TEST_P(HopperRS, SingleTileWithTMALoadStore) {
...
-
-  auto cg_outputs = fe.runFusion({inputs.first, inputs.second});
-  auto tref = atMatmul(
-      inputs.first.squeeze().to(at::kFloat),
-      inputs.second.squeeze().to(at::kFloat),
-      layout);
-  EXPECT_TRUE(at::allclose(cg_outputs[0], tref, 1e-5, 1e-5));
-}
-
-TEST_P(HopperRS, MultipleTile) {
```
zasdfgbnm, line 881: Covered by `SS` variant of `MultipleTile`

tests/cpp/test_mma.cpp
```diff
@@ -1313,303 +790,6 @@ TEST_P(HopperSS, SingleTileTransposed) {
...
-      &fusion, {inputs.first, inputs.second}, LaunchParams(), matmul_cparams);
-  auto cg_outputs = fe.runFusion({inputs.first, inputs.second});
-  auto tref = atMatmul(
-      inputs.first.squeeze().to(at::kFloat),
-      inputs.second.squeeze().to(at::kFloat),
-      layout);
-  EXPECT_TRUE(at::allclose(cg_outputs[0], tref, 1e-5, 1e-5));
-}
-
-TEST_P(HopperSS, SingleTileWithTMALoad) {
```
zasdfgbnm, line 1494: We already have a test that uses TMA for both load and store. Although that test is `RS`, because the memory layout of A and B are the same, testing B should pretty confidently mean that A will just work.

zasdfgbnm:
!build

jacobhinkle (approved):

PR #3162, opened 2024-10-11
Add select to FusionTranslation.

rdspring1:
This PR adds support for `select` operation to FusionTranslation.

Motivation:
- If the `index` argument to `index_select` is a scalar TensorView, a `SelectOp` expression is created in the Fusion.
- Rather than map `SelectOp` to an `IndexSelectOpRecord`, this PR creates the `SelectOpRecord` for a one-to-one mapping.
- It also adds the `select` operation to python_frontend. It is functionally equivalent to [torch.select](https://pytorch.org/docs/stable/generated/torch.select.html).


rdspring1:
!build

rdspring1:
!build

jacobhinkle (commented):
LGTM but I'll let @jjsjann123 give approval since this is modifying our python interface.

rdspring1:
!build

csrc/python_frontend/fusion_record.h
```diff
@@ -1820,6 +1820,58 @@ struct IndexSelectOpRecord : RecordFunctor {
   int64_t dim_;
 };
 
+struct SelectOpRecord : RecordFunctor {
```
jjsjann123, line 1823: nitpick. I wish we were a bit smarter and would have `IndexSelectopRecord` be general enough to be reused here.
But we don't have to worry about that for now. Can you please add a `TODO` here for merging indexing operations?

jjsjann123 (approved):
:clap:

rdspring1:
!build

PR #3164, opened 2024-10-11
move more buffers to smem

liqiangxl:
**In this PR**
(1) Move more buffer from registers to shared memory until their sizes are similar (For RMS norm, it has 6 bytes of cached input buffer and 8 bytes internal outer reduction buffer,equivalent to move all cached input buffers. For RMS norm, it has 6 bytes of cached input buffer and 4 bytes internal outer reduction buffer, only 4 bytes of cached input buffer is moved to shared memory)
(2) revise heuristics to set threads per block to pow2, e.g. 128 and 256. 
**Performance**
(1) Layer Norm Backward
image
(2) RMS Norm Backward (batch size = 8K, for 16K hidden size is limited to about 36K where 3 smem buffer is not triggered yet)
In the first section, only 2 tvs are moved to smem, if moving all 3 tvs to smem the smem buffer size is larger than register buffer size and causes regression, e.g. perf reduced from 80% SOL to 50% SOL at hidden size 25600.
image

**Why**
(1) Async copy to smem is more efficient than regular copy, https://github.com/NVIDIA/Fuser/pull/3132 shows perf improvment when moving 3 buffers to register but not much change when only moves 1 or 2 buffers.
(2) After step-(1), layer norm backward has regression from 16K to 19K, see blue squares in the perf figure.
For case at 18K, achieved 70% SOL with 192 threads, the parallelizaiton is:
```
inner reduction inner dim: 18432 = 8(vect) x 48 (bdimx) x 4 (bdimy) x 12 (persistent)
inner reduction outer dim: 16384 = 132 (gdimy) x serial
outer reduction inner dim: 18432 = 4(vect) x 48 (bdimx) x 96 (gdimy)
outer reduciton outer dim: 132 = 4 (bdimy) x 33 (serial)
```
after revision, achieved 78% SOL with 256 threads, the parallelizaiton is:
```
inner reduction inner dim: 18432 = 8(vect) x 64 (bdimx) x 4 (bdimy) x 9 (persistent)
inner reduction outer dim: 16384 = 132 (gdimy) x serial
outer reduction inner dim: 18432 = 4(vect) x 64 (bdimx) x 72 (gdimy)
outer reduciton outer dim: 132 = 4 (bdimy) x 33 (serial)
```
The increased perf comes from higher occupancy (6 warps --> 8 warps) and lower workload per thread (12 batches --> 9 batches).

### DIFF
--- a/csrc/scheduler/normalization_inner_outer.cpp
+++ b/csrc/scheduler/normalization_inner_outer.cpp
@@ -343,7 +343,8 @@ PersistentBufferStorageParams getPersistentBufferStorageParams(
 std::pair<int64_t, int64_t> getBufferBatchSizeAndThreadsPerBlock(
     const int64_t inner_dim_numel,
     const int64_t outer_dim_numel,
-    const int64_t persistent_buffer_size,
+    const int64_t register_buffer_size,
+    const int64_t shared_memory_buffer_size,
     const int64_t vectorize_factor,
     const int64_t warp_size) {
   // if inner_dim_numel <= 1024, we are doing multiple reductions per block
@@ -384,7 +385,7 @@ std::pair<int64_t, int64_t> getBufferBatchSizeAndThreadsPerBlock(
   // of inputs and outputs.
   auto getMaximumInnerOuterPersistentBufferBatch = [&]() -> int64_t {
     int64_t register_per_batch = ceilDiv(
-        persistent_buffer_size / inner_dim_numel * vectorize_factor,
+        register_buffer_size / inner_dim_numel * vectorize_factor,
         scheduler_utils::bytes_per_register);
     return scheduler_utils::safeDiv(
         scheduler_utils::max_registers_per_thread -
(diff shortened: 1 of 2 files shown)


csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -394,12 +408,20 @@ std::pair<int64_t, int64_t> getBufferBatchSizeAndThreadsPerBlock(
...
-    threads_per_block += warp_size;
-    inner_batch = ceilDiv(after_vectorization, threads_per_block);
+  if (shared_memory_buffer_size > 0) {
+    while (inner_batch > batch_max &&
+           threads_per_block * 2 <= threads_per_block_max &&
+           ceilDiv(after_vectorization, threads_per_block * 2) >= batch_min) {
+      threads_per_block *= 2;
+      inner_batch = ceilDiv(after_vectorization, threads_per_block);
+    }
+  }else{
```
liqiangxl, line 418: didn't change heuristics for cases without smem persistent, leave it for following heuristics tunning works.

liqiangxl:
> **In this PR** (1) Move more buffer from registers to shared memory until their sizes are similar (equivalent to move all cached input buffers for layer norm bwd with fp16, after move, it has 8 bytes register buffer & 6 bytes smem buffer) (2) revise heuristics to set threads per block to pow2, e.g. 128 and 256. **Performance** image
> 
> **Why** (1) Async copy to smem is more efficient than regular copy, #3132 shows perf improvment when moving 3 buffers to register but not much change when only moves 1 or 2 buffers. (2) After step-(1), some regression happend from 16K to 19K, see blue squares in the perf figure. For case at 18K, achieved 70% SOL with 192 threads, the parallelizaiton is:
> 
> ```
> inner reduction inner dim: 18432 = 8(vect) x 48 (bdimx) x 4 (bdimy) x 12 (persistent)
> inner reduction outer dim: 16384 = 132 (gdimy) x serial
> outer reduction inner dim: 18432 = 4(vect) x 48 (bdimx) x 96 (gdimy)
> outer reduciton outer dim: 132 = 4 (bdimy) x 33 (serial)
> ```
> 
> after revision, achieved 78% SOL with 256 threads, the parallelizaiton is:
> 
> ```
> inner reduction inner dim: 18432 = 8(vect) x 64 (bdimx) x 4 (bdimy) x 9 (persistent)
> inner reduction outer dim: 16384 = 132 (gdimy) x serial
> outer reduction inner dim: 18432 = 4(vect) x 64 (bdimx) x 72 (gdimy)
> outer reduciton outer dim: 132 = 4 (bdimy) x 33 (serial)
> ```
> 
> The increased perf comes from higher occupancy (6 warps --> 8 warps) and lower workload per thread (12 batches --> 9 batches).

liqiangxl:
!build --pybench

liqiangxl:
The changes in this PR are merged in https://github.com/NVIDIA/Fuser/pull/3175

PR #3167, opened 2024-10-11
Compare cparams in MatmulParams::sameAs

### LINKED ISSUE #3157: Illegal memory access error in python matmul benchmark with nvfuser scheduler
To reproduce on main, use this command with an A100 80GB PCIe:
```
NVFUSER_ENABLE=fuse_matmul NVFUSER_DISABLE=matmul_expr_eval pytest benchmarks/python/test_matmul.py -vs
```
Either with or without our internal heuristic plugin, we have a lot of successes, then eventually we start hitting failures like 
`RuntimeError: CUDA error: an illegal memory access was encountered` at the test ` benchmarks/python/test_matmul.py::test_matmul_nvf_benchmark[dtype=torch.float16-328_302328_22072_TN]`. Using `CUDA_LAUNCH_BLOCKING=1` confirms that it always happens at this test. Running this test directly does not lead to error. `compute-sanitizer` shows invalid reads at the `cp.async` of the A operand before the main loop, i.e. in the circular buffer prologue where we are doing the initial fill of the circular buffer.

These are the scheduler params and kernel args for this failing test:
```

MMA macro: Ampere_16_8_16
CircularBufferOptions:
  circular_buffer_smem_write: true
  circular_buffer_smem_read: true
  smem_circular_buffer_stage: 5
SupportedVectorization:
  a: 8
  b: 8
  epilogue: 8
MatMulTileOptions: instruction tile [16, 8, 16], warp tile [64, 64, 32], CTA tile [128, 128, 32]
Async global mem load: true
Indexing mode: int64_t
Tile rastrization order: column-major
Grid swizzle factor: 1
Use shared memory epilogue: 1
Promote re-use of prologue shared memory: 1
Split-K factor: 1

Arguments for kernel2:
Inputs:
  Tensor(sizes=[328, 22072], stride=[22072, 1], dtype=c10::Half, device=cuda:0, data_ptr=0x7f0a06000000)
  Tensor(sizes=[22072, 302328], stride=[1, 22072], dtype=c10::Half, device=cuda:0, data_ptr=0x7f06ba000000)
Outputs:
  Half [328, 302328] (strides = [302328, 1], address = 0x7f0a12c00000)
Intermediate global buffers:
```

This is unrelated to the recent PR #3150, as you can see by adding `-k TT` (this means default strides are used for both operands).

You can add `-x` to fail on the first error, or omit it to see that eventually we start hitting this error and never stop.

Removing `NVFUSER_DISABLE=matmul_expr_eval`, we do not hit this error.

jacobhinkle:
A change of index_type caused the IMA encountered in #3157. This happened because an earlier fusion was compiled with the same matmul params but Int32 index type, then it was re-used when Int was needed, causing wraparound and an illegal index.

Fixes #3157

### DIFF
--- a/csrc/scheduler/matmul_heuristic.h
+++ b/csrc/scheduler/matmul_heuristic.h
@@ -22,7 +22,7 @@ namespace nvfuser {
 class MatmulParams : public HeuristicParams {
  public:
   MatmulParams()
-      : HeuristicParams(SchedulerType::Matmul), supported_vec_size(){};
+      : HeuristicParams(SchedulerType::Matmul), supported_vec_size() {};
   //! A list of possible strategies used to define along which axis
   //!  parallelization will be done.
   enum class TileRasterizationOrder { RowMajor = 0, ColumnMajor = 1 };
--- a/csrc/scheduler/matmul_heuristic.h
+++ b/csrc/scheduler/matmul_heuristic.h
@@ -238,7 +238,8 @@ class MatmulParams : public HeuristicParams {
       return false;
     }
 
-    return other->mma_macro == mma_macro &&
+    return other->cparams == cparams &&
+        other->mma_macro == mma_macro &&
         other->async_gmem_load_operands == async_gmem_load_operands &&
         other->rotate_ldmatrix_out_of_main_loop ==
         rotate_ldmatrix_out_of_main_loop &&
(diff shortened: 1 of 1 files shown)


jacobhinkle:
I checked all the other HeuristicParams subclasses and none of them have this deficiency.

jacobhinkle:
!build

zasdfgbnm (approved):

jacobhinkle:
!build

PR #3172, opened 2024-10-11
Allow both load and compute to be pipelined for circular buffering

zasdfgbnm:
Today, a circular-buffered Hopper MMA kernel looks like below (assuming 4 stages):

```python
# prologue
for i in range(3):
    load to A_buffer[i];
    load to B_buffer[i];
# main loop
for i in range(N):
    load to A_buffer[(i + 3) % 4];
    load to B_buffer[(i + 3) % 4];
    wait for A_buffer[i % 4] and B_buffer[i % 4] to be ready;
    mma(A_buffer[i % 4], B_buffer[i % 4]);
    wait for mma to complete
# epilogue
for i in range(N-3, N):
    wait for A_buffer[i % 4] and B_buffer[i % 4] to be ready;
    mma(A_buffer[i % 4], B_buffer[i % 4]);
    wait for mma to complete
```

Note that the `wait for mma to complete` in the main loop is mandatory, because at the next iteration, TMA will write to the operands of the mma operation in the current iteration, and the wait in the end of the current iteration is to avoid WAR hazard. Due to this wait, we are actually using mma as a synchronous op because we sync immediately after the mma instruction being fired. Not taking advantage of the async nature of mma means that it could be hard to make the TensorCore full.

This PR adds an additional parameter `prefetch` to have better control of our circular buffering behavior. Currently, we always allocate `stage * tile_size` buffer size, and prefetch `stage - 1` tiles in the prologue. With this PR, we still allocate `stage * tile_size` buffer size, but we will prefetch `prefetch` tiles in the prologue. `prefetch` can range from `0` to `stage - 1`. If `prefetch == 0`, this means we do not prefetch at all. `prefetch == stage - 1` would reproduce the current behavior, therefore is the default value. For the case where `prefetch < stage - 1`, the operand of the MMA in each iteration will not be overwritten by the next iteration, therefore no WAR hazard, and we can defer the wait of mma a few iterations away, allowing MMA to be pipelined.

I would consider `prefetch` as a slide bar between `0` and `stage - 1`, adjusting the partition of the entire `stage * tile_size` into "buffers to remove WAR hazard" and "buffers to remove RAW hazard". Sliding to the left most means "the entire buffer is used for removing WAR", and sliding to the right most means "the entire buffer is used for removing RAW". During schedule, we should adjust this slide bar to achieve the best performance.

I haven't looked deep into the impact on Hopper matmul perf. Just did a quick experiment, and I was able to push our perf to 58% of cuBLAS. Please note that this PR alone will not benefit Hopper matmul perf, because we always generate `wgmma.wait_group 0` (which means to wait untill 0 pending mma ops). A followup-PR will update this value according to `prefetch` and `stage`.

### DIFF
--- a/csrc/device_lower/analysis/circular_buffer.cpp
+++ b/csrc/device_lower/analysis/circular_buffer.cpp
@@ -204,21 +204,21 @@ void CircularBufferInfo::setStageDepthAndPrefetchDistance(
     int64_t prefetch_distance) {
   auto concrete_loop_id = lower_utils::getConcreteLoopID(id);
 
-  auto maybe_exisiting_depth_it =
+  auto maybe_existing_depth_it =
       circular_buffer_options_.find(concrete_loop_id);
-  if (maybe_exisiting_depth_it == circular_buffer_options_.end()) {
+  if (maybe_existing_depth_it == circular_buffer_options_.end()) {
     circular_buffer_options_[concrete_loop_id].stage = stage_depth;
     circular_buffer_options_[concrete_loop_id].prefetch = prefetch_distance;
   } else {
     NVF_ERROR(
-        stage_depth == maybe_exisiting_depth_it->second.stage &&
-            prefetch_distance == maybe_exisiting_depth_it->second.prefetch,
+        stage_depth == maybe_existing_depth_it->second.stage &&
+            prefetch_distance == maybe_existing_depth_it->second.prefetch,
         "Unsupported multiple depth/prefetch pipelining, was set to (",
-        maybe_exisiting_depth_it->second.stage,
+        maybe_existing_depth_it->second.stage,
         ",",
-        maybe_exisiting_depth_it->second.prefetch,
+        maybe_existing_depth_it->second.prefetch,
         ") by ",
-        maybe_exisiting_depth_it->first->toString(),
+        maybe_existing_depth_it->first->toString(),
         " and then set to (",
         stage_depth,
         ",",
--- a/csrc/device_lower/analysis/circular_buffer.cpp
+++ b/csrc/device_lower/analysis/circular_buffer.cpp
@@ -194,25 +194,36 @@ void CircularBufferInfo::setCircularBufferAxis(
     IterDomain* axis) {
   getTvInfo(tv).circular_buffer_axis = axis;
   // Set and validate the new stage depth.
-  setStageDepth(axis, tv->circularBufferDepth());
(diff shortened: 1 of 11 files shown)


zasdfgbnm:
!build

csrc/device_lower/analysis/circular_buffer.cpp
```diff
@@ -194,25 +194,36 @@ void CircularBufferInfo::setCircularBufferAxis(
...
+void CircularBufferInfo::setStageDepthAndPrefetchDistance(
+    IterDomain* id,
+    int64_t stage_depth,
+    int64_t prefetch_distance) {
   auto concrete_loop_id = lower_utils::getConcreteLoopID(id);
 
-  auto maybe_exisiting_depth_it = stage_depth_.find(concrete_loop_id);
-  if (maybe_exisiting_depth_it == stage_depth_.end()) {
-    stage_depth_[concrete_loop_id] = stage_depth;
+  auto maybe_exisiting_depth_it =
```
rdspring1, line 207:
```suggestion
  auto maybe_existing_depth_it =
```

tests/cpp/test_circular_buffering.cpp
```diff
@@ -800,24 +804,43 @@ TEST_P(CircularBufferingTest, NoSync) {
 }
 
 // Test circular buffer from 2 to 10 stages
+auto StagesAndPrefetches() {
+  std::vector<StageAndPrefetch> values;
+  for (int64_t i : c10::irange(2, 10)) {
+    for (int64_t j : c10::irange(i)) {
```
rdspring1, line 810: Should we test some negative prefetch values?

csrc/device_lower/analysis/circular_buffer.cpp
zasdfgbnm: fixed

tests/cpp/test_circular_buffering.cpp
zasdfgbnm: Good idea. Added.

rdspring1 (commented):
LGTM. I'll let @jacobhinkle and @naoyam have a chance to review the PR.

naoyam (approved):
LGTM. Having the prefetch parameter seems to make sense.

jacobhinkle (approved):
LGTM

zasdfgbnm:
!build

PR #3174, opened 2024-10-12
Skip python matmul benchmarks that fail due to OOM

jacobhinkle:
I'm adding this skip so that we can soon expect to run python matmul benchmarks without any failures. Some of these tests use extreme sizes needing over 200GB of memory. Instead of removing these tests, I try to run all tests and skip the ones that fail due to torch OOM error.


jacobhinkle:
!diff --matmul-bench

zasdfgbnm (approved):

jacobhinkle:
!build --matmul-bench

jacobhinkle:
!build --matmul-bench

PR #3175, opened 2024-10-13
move all persistent cached inputs to shared memory even there are enough registers

liqiangxl:
**Issue**:
In the current innerOuter persistent scheduler, we prioritize using registers to store all persistent buffers. Shared memory is utilized for persistence only when there are insufficient registers, such as when the hidden size exceeds 16K in layer norm backward. Achieving optimal performance requires careful tuning of heuristics to prevent register spills. Performance is highly sensitive to register usage and spilling, where even a small increase in register spills can result in a 20% performance drop or improvement, as seen in issue #2741. This sensitivity makes it challenging to develop robust, high-performance heuristics.

**In this PR**:
This PR demonstrates that using CpAsync, shared memory persistence can achieve performance levels comparable to, and somestimes better than, register-based persistence at various hidden sizes, without fine-tuning heuristics.
**Three major changes: (can further split into 3 PRs, all included in this one to simplify benchmarks)**
**(1) Moved all persistent cached inputs to shared memory (if device smem size is large enough)**
Previously, shared memory is utilized for persistence only when there are insufficient registers. This PR tries to move all the cached inputs to shared memory even there are enough registers to store all of them.

**(2) Implemented vectorized loads from shared memory to registers to avoid bank conflicts.**
For example, in test `PersistentBufferTest.SmemPersistent2DReduction`, there are bank conflicts in computing `T5_l = T4_s / T2_l` since each thread visits four 32-bytes consecutive elements in `T4_s`  because `T4_s` is vectorized by 4 during `T4_s = CpAsync(T0_g)`
Before this PR:
```
Inputs:
  T0_g_float[ iS0{i0}, iS1{i2}, iS2{i3}, iS3{i4} ]
Outputs:
  T3_g_float[ iS20{i0}, iS21{i2}, iS22{i3}, iS23{i4} ]

%kernel_math {
T4_s_float[ iS16{i0}, iS17{i2}, iS18{i3}, iS19{i4} ]
   = CpAsync( T0_g_float[ iS0{i0}, iS1{i2}, iS2{i3}, iS3{i4} ] )
T1_l_float[ iS4{i0}, rS5{i2}, rS6{i3}, rS7{i4} ]
   = reduction( T4_s_float[ iS16{i0}, iS17{i2}, iS18{i3}, iS19{i4} ], op = add, initial value = float(0), allreduce = false )
T2_l_float[ iS8{i0}, bS9{1}, bS10{1}, bS11{1} ]
   = broadcast( T1_l_float[ iS4{i0}, rS5{i2}, rS6{i3}, rS7{i4} ] )
T5_l_float[ iS12{i0}, iS13{i2}, iS14{i3}, iS15{i4} ]
   = T4_s_float[ iS16{i0}, iS17{i2}, iS18{i3}, iS19{i4} ]
   / T2_l_float[ iS8{i0}, bS9{1}, bS10{1}, bS11{1} ];
T3_g_float[ iS20{i0}, iS21{i2}, iS22{i3}, iS23{i4} ]
   = Set( T5_l_float[ iS12{i0}, iS13{i2}, iS14{i3}, iS15{i4} ], cache_op=Streaming )
} // %kernel_math 

```
Aftert this PR:
```
Inputs:
  T0_g_float[ iS0{i0}, iS1{i2}, iS2{i3}, iS3{i4} ]
Outputs:
  T3_g_float[ iS20{i0}, iS21{i2}, iS22{i3}, iS23{i4} ]

%kernel_math {
T4_s_float[ iS16{i0}, iS17{i2}, iS18{i3}, iS19{i4} ]
   = CpAsync( T0_g_float[ iS0{i0}, iS1{i2}, iS2{i3}, iS3{i4} ] )
T7_l_float[ iS28{i0}, iS29{i2}, iS30{i3}, iS31{i4} ]
   = Set( T4_s_float[ iS16{i0}, iS17{i2}, iS18{i3}, iS19{i4} ], cache_op=Streaming )
T6_l_float[ iS24{i0}, iS25{i2}, iS26{i3}, iS27{i4} ]
   = Set( T4_s_float[ iS16{i0}, iS17{i2}, iS18{i3}, iS19{i4} ], cache_op=Streaming )
T1_l_float[ iS4{i0}, rS5{i2}, rS6{i3}, rS7{i4} ]
   = reduction( T6_l_float[ iS24{i0}, iS25{i2}, iS26{i3}, iS27{i4} ], op = add, initial value = float(0), allreduce = false )
T2_l_float[ iS8{i0}, bS9{1}, bS10{1}, bS11{1} ]
   = broadcast( T1_l_float[ iS4{i0}, rS5{i2}, rS6{i3}, rS7{i4} ] )
T5_l_float[ iS12{i0}, iS13{i2}, iS14{i3}, iS15{i4} ]
   = T7_l_float[ iS28{i0}, iS29{i2}, iS30{i3}, iS31{i4} ]
   / T2_l_float[ iS8{i0}, bS9{1}, bS10{1}, bS11{1} ];
T3_g_float[ iS20{i0}, iS21{i2}, iS22{i3}, iS23{i4} ]
   = Set( T5_l_float[ iS12{i0}, iS13{i2}, iS14{i3}, iS15{i4} ], cache_op=Streaming )
} // %kernel_math 
```
The difference is additional loading from shared memory to registers `T7_l_float` and `T6_l_float` these loadings are then vectorized by the scheduler.
The generated code is:
```
      Array<float, 4LL, 4> T6;
      T6.set(float(0.000000000e+00f));
      loadGeneric<float, 4>( &T6[0LL],  &T4[i14]);
      #pragma unroll
      for(nvfuser_index_t i16 = 0LL; i16 < 4LL; ++i16) {
        T8[0LL]
          = T8[0LL]
          + T6[i16];
      }
```
(3) Small changes to heuristic. (a) Each block uses 128 or 256 threads, it leads to high occupancy and better performance, see https://github.com/NVIDIA/Fuser/pull/3164. (b) Made a small adjustment to the heuristics to avoid using excessively large persistent batch sizes. After moving all cached inputs to shared memory, the remaining register buffer size is 4 bytes per element for RMS norm bwd. This allows for a larger persistent batch size without causing register spills, but leads to fewer threads per block, e.g. at 18K it uses 128 threads, each processes 8 x 18 elements. To avoid this, the max allowed persistent batch size is empirically set to 14 (will be removed in a following PR where a sort method is added to select the best heuristic paras).

**Performance**:
(1) Layer Norm Backward
image
(2) RMS Norm backward
image

**Why**:
(1) speedup is due to increased occupancy, e.g. ln bwd at hidden size of 8192, increased from 8 to 12 warps per sm.
(2) speedup is due to reduced register pressure, removed register spills, and allows larger persistent batch size.
(2) Regression at small hidden sizes is because register pressure is low and the original occupancy is already high, e.g. layer norm backward max regression happened at hidden size 1536, occupancy increased from 20 to 28 warps per sm but perf dropped from 65% SOL to 60% SOL. We can avoid these regressions by further tunning when should start using shared memory persistent buffers.


liqiangxl:
!build --pybench

liqiangxl:
!build

liqiangxl:
!build

liqiangxl:
!build

naoyam:
It is unclear to me how each of the changes is related to each other. I think it would be helpful that some of the obvious fixes are addressed first. For example, I assume avoiding bank conflicts should be always beneficial and it shouldn't depend on the other changes. If so, please address it in a separate PR and make sure to include proper unit tests.

Also, please doublecheck the perf graphs as they appear inconsistent. For example, why RMS norm doesn't have opt-1? Please use the same colors for the same versions. The layernorm results seem to indicate opt-1 is the best one, which is supposed to suffer from bank conflicts. Is that true?

liqiangxl:
> It is unclear to me how each of the changes is related to each other. I think it would be helpful that some of the obvious fixes are addressed first. For example, I assume avoiding bank conflicts should be always beneficial and it shouldn't depend on the other changes. If so, please address it in a separate PR and make sure to include proper unit tests.
> 
> Also, please doublecheck the perf graphs as they appear inconsistent. For example, why RMS norm doesn't have opt-1? Please use the same colors for the same versions. The layernorm results seem to indicate opt-1 is the best one, which is supposed to suffer from bank conflicts. Is that true?

Thanks for the feedback. I probabally should make this PR as an issue to discuss proposed solutions. These changes can be seperated, not closely linked, there were put together to reduce number of benchmarks. **(1) Removing bank conflicts is now moved to PR-https://github.com/NVIDIA/Fuser/pull/3186 we can discuss other changes after step-(1) is done.**

(2) `why RMS norm doesn't have opt-1?` It is merged with `opt-2` and not benchmarked in each step. 
(3) You are right, the table in layernorm results is not correct. opt-1 is blue, opt-2 is yellow, opt-3 is green. (The figure is corrected now, the order in the legend confused me.)

PR #354, opened 2023-05-17
Add dynamic shape enabling operations: `ops.shape`, `ops.size`, and `ops.at` to the  `FusionDefinition`

kevinstephano:
This PR is adding the following operations to the FusionDefinition:
1. `fd.ops.shape(Tensor)`: returns a `Vector` of symbolic tensor dimension sizes.  Includes a `Tensor` method (ie `T1.shape()`).
2. `fd.ops.size(Tensor, dim)`: returns a symbolic size for a given `Tensor` dimension.  Includes a `Tensor` method (ie `T1.size(0)`).
3. `fd.ops.at(Vector,  index)`: returns a symbolic value for a given `Vector` index.  Includes a `Vector` method `__getitem__` that supports bracket indexing (ie `V1[0]`).

This PR also includes:

- Opinfo error testing of the 3 new functions.
- Arith functions in the code generator for each function

PR #3179, opened 2024-10-14
Deduplicate tensor_sizes and shape.

wujingyue:
They have the same implementation -- not sure why they have co-existed. I prefer `shape` because it's shorter. To reduce blast radius, I only removed `tensor_sizes` from the C++ APIs. The python API, `fd.ops.tensor_sizes`, remains the same.

### DIFF
--- a/csrc/ops/arith.cpp
+++ b/csrc/ops/arith.cpp
@@ -1738,17 +1738,6 @@ TensorView* expand_as(TensorView* inp, TensorView* other) {
   return out_tensor;
 }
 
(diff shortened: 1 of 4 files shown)


wujingyue:
!build

naoyam (approved):
LGTM.

Just in case, pinging @kevinstephano since it seems that `tensor_sizes` existed before `shape`, which was added in #354 by @kevinstephano.

wujingyue:
!build

PR #3182, opened 2024-10-14
Remove bit_cast to convert MmaMacroEncode to uint64_t

jacobhinkle:
This standardizes how we convert `MmaMacroEncode` to `uint64_t`. Doing this with a bitcast actually gives us UB because we are not guaranteed to have the given order for struct fields defined as bitfields. I first noticed this was happening when using a `libnvfuser.so` compiled as C++20 with another project linking against that was compiled using C++17. In that case, the compilation and linking worked properly, but we wound up with garbage for the `MmaMacro`. I think this PR will address this kind of issue.

### DIFF
--- a/csrc/mma_type.h
+++ b/csrc/mma_type.h
@@ -16,9 +16,6 @@
 #include <cstring>
 #include <ostream>
 
-#if IS_CPP20
-#include <bit>
-#endif
 #include <cstdint>
 
 namespace nvfuser {
--- a/csrc/mma_type.h
+++ b/csrc/mma_type.h
@@ -98,12 +98,18 @@ struct MmaMacroEncode {
         (uint64_t)k;
   }
 
-  constexpr operator MmaMacro();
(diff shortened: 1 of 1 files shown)


jacobhinkle:
!build

zasdfgbnm (commented):
Do we also want to remove `MmaMacroEncode::operator MmaMacro` and the constructor?

jacobhinkle:
> Do we also want to remove `MmaMacroEncode::operator MmaMacro` and the constructor?

Removing them would require some unergonomic changes to other parts of the code, so I just modified them by removing the `#if`.

jacobhinkle:
!build

jacobhinkle:
At this point the only use of `IS_CPP20` is here: https://github.com/NVIDIA/Fuser/blob/55971f8811243822983edf8510b3fa26524d1f15/tests/cpp/test_matmul.cpp#L484-L488

jacobhinkle:
!build

PR #3177, opened 2024-10-14
Add a multi-GPU test for transformer backward.

wujingyue:
All tensors are replicated to all devices to start with. Future PRs will try to shard them. 

For #2199.

PR #3184, opened 2024-10-15
Update the forward pass based on a more recent Thunder version. 

wujingyue:
More tensors are saved, i.e., are returned as an output. This new version is also consistent with #3177, which tries to add the backprop. 

The previous version came from benchmarks/python/test_transformer.py, which has been outdated. Down the road, I plan to remove benchmarks/python/test_transformer.py and maintain only one version.

### DIFF
--- a/tests/python/test_multidevice.py
+++ b/tests/python/test_multidevice.py
@@ -195,6 +195,11 @@ def test_sdpa(mpi_test):
 # 2. Split device dimensions and parallelize them.
 # 3. Decompose the second linear layer in MLP so the matmul result can be allreduced.
 # 4. Rename the receiving variables for `define_scalar`s and `define_tensor`s for readability.
+
+
+# torch version: 2.6.0a0+git0eba7e5
+# cuda version: 12.6
+# nvfuser version: 0.2.15+gitf3a2087
 class TransformerForwardFusion(FusionDefinition):
     def __init__(self, num_devices, batch, sequence, head, hidden):
         super().__init__()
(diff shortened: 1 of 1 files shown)


wujingyue:
!build

cowanmeg:
Can you summarize what the differences with this newer version is? I assume thunder is going to keep updating their passes, so it's infeasible that we keep manually updating this and re-sharding

wujingyue:
> Can you summarize what the differences with this newer version is? 

I briefly mentioned that in the PR description: 
> Fewer tensors are saved, i.e., are returned as an output. This new version is also consistent with https://github.com/NVIDIA/Fuser/pull/3177, which tries to add the backprop. 

Let me know if you'd like more details. 

> I assume thunder is going to keep updating their passes, so it's not a good use of our time to keep manually updating this and re-sharding

Good point! 

I did this mainly to keep it consistent with backprop (#3177). Otherwise, the two benchmarks won't be realistic -- forward would have saved tensors that are never read by backprop. 

I see two options down the road. 
1. If the purpose of these benchmarks is to reflect Thunder, we'll keep them in sync periodically, e.g., once per quarter. 
2. If the purpose is instead to reflect the best of nvFuser, let's try to optimize the two fusions for the best performance. Deciding what tensors to save vs recompute will be part of that. We don't have to get the best version immediately -- we can always start with one implementation and keep iterating.

cowanmeg:
Ahh got it, I didn't realize this was necessary for the updated backprop PR. Thanks for clarifying!

cowanmeg (approved):

wujingyue:
!build

PR #3187, opened 2024-10-15
TMA store to global memory after stmatrix

protonu:
Modified the stmatrix to store to global memory using TMA store.

### DIFF
--- a/tests/cpp/test_memory.cpp
+++ b/tests/cpp/test_memory.cpp
@@ -2719,6 +2719,9 @@ TEST_P(StMatrixTest, Regular) {
   tv2->setMemoryType(MemoryType::Shared);
   // tv2 (shared) -> tv3(global)
   auto tv3 = set(tv2);
+  // TMA store back to global memory.
+  tv3->definition()->as<LoadStoreOp>()->setOpType(
+      LoadStoreOpType::CpAsyncBulkTensorTile);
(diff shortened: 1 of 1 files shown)


protonu:
!build

zasdfgbnm:
Could we keep this unit test unit? We can add a separate test with stmatrix + TMA, and keep this test as small as just stmatrix. One interesting thing to study in that separate stmatrix test is, are we currently accessing smem with bank conflict? If yes, is our `getBankConflictInfo` capable of catching it? And how to swizzle it in a way that is consistent with TMA?

protonu:
@zasdfgbnm  Thanks - great points! I was trying to not increase the number of tests we have. How about I add this to the single tile Mma test; where we can use `stmatrix` in addition to TMA load/store. Then as a next step I could investigate bank conflicts.

zasdfgbnm:
> @zasdfgbnm Thanks - great points! I was trying to not increase the number of tests we have. How about I add this to the single tile Mma test; where we can use `stmatrix` in addition to TMA load/store. Then as a next step I could investigate bank conflicts.

Sounds reasonable to me. Make sure don't add a new MMA test because it takes really. Could we add it to `SingleTileWithTMALoadStore`?

protonu:
This work is no longer needed - we'll create a separate test with mma + stmatrix + tma

PR #3196, opened 2024-10-16
clean propagate parallel

liqiangxl:


liqiangxl:
!build

liqiangxl:
!build

naoyam:
Can you please add more detailed comments to those functions? Please describe each individual parameter. It can be very brief, but descriptions about what each parameter is supposed to do would be also helpful to understand what the function is supposed to do.

liqiangxl:
> Can you please add more detailed comments to those functions? Please describe each individual parameter. It can be very brief, but descriptions about what each parameter is supposed to do would be also helpful to understand what the function is supposed to do.

Good point! Added in the code.

csrc/scheduler/reduction_utils.h
```diff
@@ -56,22 +56,53 @@ void propagateRFactor(
...
+//   is_unroll_or_vectorization: Indicates if unroll or vectorization is used in
+//                               the scheduler.
+//
+//   reduction_tvs: All reduction TensorViews in the fusion. May add grouped
+//                  parallelization.
+//
+//   unroll_vectorizable_cached_tvs: Cached TensorViews that are both unrollable
+//                                   and vectorizable.
+//
+//   selected_tvs: TensorViews selected for parallelization.
```
naoyam, line 98: Mention what the default value means.

csrc/scheduler/reduction_utils.cpp
```diff
@@ -426,135 +428,152 @@ void propagateRFactor(
...
+          cached_smem->definition()->toString());
+      are_unrolled.emplace(cached_smem);
+    }
+  }
+
+  return are_unrolled;
+}
+
+namespace {
+void reductionTvUnrollVectorizationGroup(
```
naoyam, line 493: Unclear to me what this function name means. Again, please don't forget adding comments, especially for functions with this many parameters.

csrc/scheduler/reduction_utils.h
```diff
@@ -56,22 +56,53 @@ void propagateRFactor(
...
+// Get all cached input/output and shared memory TensorViews that are
+// vectorizable and unrollable.
+//
+// Parameters:
+//   reference_tv: TensorView created during RFactor, used to find vectorizable
+//                 TensorViews.
+//   is_vectorize: Indicates if vectorization is applied in the scheduler.
+//   cached_inputs: Inputs cached in registers or shared memory.
+//   cached_outputs: Outputs cached in registers.
+//   cached_smem_buffer: Shared memory persistent buffers cached in registers.
```
naoyam, line 68: Is this tensor cached in registers? Isn't it just cached in shared memory?

csrc/scheduler/reduction_utils.cpp
```diff
@@ -426,135 +428,152 @@ void propagateRFactor(
   }
 }
 
-void propagateParallelization(
-    Fusion* fusion,
-    TensorView* reduction_tv,
+std::unordered_set<TensorView*> getUnrollVectorizableCachedTvs(
```
naoyam, line 431:
```suggestion
std::unordered_set<TensorView*> getCachedTvsToUnrollOrVectorize(
```

csrc/scheduler/reduction_utils.cpp
liqiangxl: renamed to `clearUnrollVectorizationAddGroupReduction()`. oopps, forgot about this internal function, added comments:
```
// Clears unroll or vectorization parallelization for reduction_tv and
// reference_tv if they shouldn't be unrolled or vectorized.
// When group reduction is used, convert vectorization to group parallelization
// and propagate group parallelization to other reduction tvs.

// Parameters:
//   reduction_tv: The reduction TensorView being scheduled and parallelized.
//                 Needs to clear its vectorization or convert to grouped
//                 reduction.

//   reference_tv: The reference TensorView being scheduled and parallelized,
//                 Needs to clear its vectorization.

//   use_grouped_reduction: Indicates if group reduction is used in the
//                          scheduler.

//   reduction_tvs: All reduction TensorViews in the fusion. May add grouped
//                  parallelization.
//
//   unroll_vectorizable_cached_tvs: Cached TensorViews that are both unrollable
//                                   and vectorizable.
```

csrc/scheduler/reduction_utils.h
liqiangxl: added, means all tvs are selected.

csrc/scheduler/reduction_utils.h
liqiangxl: Yes, they are in registers. The name follows `cached_inputs`. For example, `cached_inputs` means input tvs cached in registers or shared memory. So here `cached_smem_buffer` means smem_buffer cached in registers.
It's a little bit confusing. I modified to `smem_consumers`
```
//   smem_consumers: Consumers of shared memory persistent buffers, they are
//                   register cached Tvs after the shared memory tv.
```

csrc/scheduler/reduction_utils.cpp
liqiangxl: manually applied.

liqiangxl:
!build

csrc/scheduler/reduction_utils.h
naoyam: Still surprised. Are you saying there's a tensor stored in shared memory and we make a cache of the tensor in registers? Why?

liqiangxl: We want to vectorized load them from smem to registers to avoid bank conflicts. [Here](https://github.com/NVIDIA/Fuser/blob/973205ef6cc279c692434bded3899ebb7160594b/csrc/scheduler/normalization_utils.cpp#L1267) is the code.

naoyam: It shouldn't be called a cache as there's no reuse.

liqiangxl: renamed to `smem_consumers`

liqiangxl:
!build

csrc/scheduler/reduction_utils.cpp
```diff
@@ -426,17 +424,142 @@ void propagateRFactor(
...
+void clearUnrollVectorizationAddGroupReduction(
+    TensorView* reduction_tv,
+    TensorView* reference_tv,
+    const bool use_grouped_reduction,
+    const std::vector<TensorView*>& reduction_tvs,
+    const std::unordered_set<TensorView*>& unroll_vectorizable_cached_tvs) {
+  std::vector<TensorView*> rfactor_and_reduction_tvs = {
+      reference_tv, reduction_tv};
+  for (auto tv : rfactor_and_reduction_tvs) {
+    if (unroll_vectorizable_cached_tvs.count(tv) == 0) {
```
naoyam, line 519:
```suggestion
    if (unroll_vectorizable_cached_tvs.count(tv) != 0) {
      continue;
    }
```

naoyam (approved):
LGTM

liqiangxl:
!build

liqiangxl:
!build

PR #3197, opened 2024-10-16
Fix Issue 3190

### LINKED ISSUE #3190: Errors during _finalize_definition don't indicate the problematic op.
This made efforts like #3184 much harder. I had to examine all reshapes one by one and manually inferred the shapes. 

https://github.com/NVIDIA/Fuser/commit/88fb1626d13b19bd5041c23b240381d2f7ad7850 added a simplified example. It's easy to spot the error in that example because there's only one reshape and the shapes are constant. However, it's much harder to deal with large fusions like the one in #3184.

rdspring1:
This PR adds which `RecordFunctor` is causing the nvfuser failure to the exception message.

Fixes #3190

### DIFF
--- a/csrc/python_frontend/fusion_state.cpp
+++ b/csrc/python_frontend/fusion_state.cpp
@@ -95,7 +95,19 @@ void FusionState::buildFusionIr(Fusion* fusion) {
   auto fusion_guard = FusionGuard(fusion);
   for (auto& record : recording_) {
     auto functor = record.get();
-    (*functor)(*this);
+    try {
+      (*functor)(*this);
+    } catch (const std::exception& e) {
+      std::stringstream ss;
+      record->print(ss);
+
+      NVF_ERROR(
+          false,
+          "Detected exception while building Fusion Ir. The failing RecordFunctor is ",
+          ss.str(),
+          "\nNvFuser error message is printed below.\n",
+          e.what());
+    }
   }
 }
 
--- a/csrc/python_frontend/fusion_state.cpp
+++ b/csrc/python_frontend/fusion_state.cpp
@@ -101,8 +101,7 @@ void FusionState::buildFusionIr(Fusion* fusion) {
       std::stringstream ss;
       record->print(ss);
 
-      NVF_ERROR(
-          false,
+      NVF_THROW(
           "\nDetected exception while building Fusion Ir. The failing RecordFunctor is: ",
           ss.str(),
           "\nNvFuser error message is: ",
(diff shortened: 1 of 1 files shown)


rdspring1:
!build

csrc/python_frontend/fusion_state.cpp
```diff
@@ -95,7 +95,19 @@ void FusionState::buildFusionIr(Fusion* fusion) {
...
     auto functor = record.get();
-    (*functor)(*this);
+    try {
+      (*functor)(*this);
+    } catch (const std::exception& e) {
+      std::stringstream ss;
+      record->print(ss);
+
+      NVF_ERROR(
+          false,
```
wujingyue, line 105:
```suggestion
      NVF_THROW(
```

wujingyue:
This is indeed an improvement but doesn't quite fix the issue. For example, for https://github.com/NVIDIA/Fuser/commit/88fb1626d13b19bd5041c23b240381d2f7ad7850, it prints
```
Detected exception while building Fusion Ir. The failing RecordFunctor is: T3 = fd.ops.reshape(T0, new_shape=[5])
```
However, `T3` and `T0` don't occur in the source code. Therefore, when the fusion definition has many reshapes of similar shapes, it's still hard to figure out which. Any ideas?

wujingyue (approved):

rdspring1:
```python
def nvfuser_incomplete_fusion(fd : FusionDefinition) -> None :
    T0 = fd.define_tensor(shape=[2, 3], contiguity=[False, False], dtype=DataType.Float, is_cpu=False)
    T3 = fd.ops.reshape(T0, new_shape=[5])
    fd.add_output(T3)
```
`T0` and `T3` exist in the incomplete definition. It is printed during failure.

rdspring1:
pybind11 doesn't have any knowledge of the variable names in the original script nor does the `RecordFunctor`.

rdspring1:
```python
import inspect

def get_variable_name(variable):
    frame = inspect.currentframe().f_back
    for name, value in frame.f_locals.items():
        if value is variable:
            return name
```

As you create a `FusionDefinition`, you can get its variable name in python and store that information in `RecordFunctor`.
The print function would use the original variable name if it exists in `RecordFunctor`.

rdspring1:
!build

wujingyue:
> ```python
> def nvfuser_incomplete_fusion(fd : FusionDefinition) -> None :
>     T0 = fd.define_tensor(shape=[2, 3], contiguity=[False, False], dtype=DataType.Float, is_cpu=False)
>     T3 = fd.ops.reshape(T0, new_shape=[5])
>     fd.add_output(T3)
> ```
> 
> `T0` and `T3` exist in the incomplete definition. It is printed during failure.

Got it! I missed the `nvfuser_incomplete_fusion` printed out.

PR #3200, opened 2024-10-16
Transformer benchmark

cowanmeg:
(1) Moves C++ Transformer definition from `test_multidevice_transfromer.cpp` to `multidevice_transfromer.h/cpp`. So that we can call the same transformer definition in the tests for validation (with no SDPA dropout, smaller parameters) and in the benchmark (with GPT3 parameter sizes).

### DIFF
--- a/tests/cpp/multidevice_transformer.cpp
+++ b/tests/cpp/multidevice_transformer.cpp
@@ -1,0 +1,616 @@
+// clang-format off
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2023-present NVIDIA CORPORATION & AFFILIATES.
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+// clang-format on
+#include <vector>
+
+#include <fusion.h>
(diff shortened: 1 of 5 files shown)


tests/cpp/multidevice_transformer.h
```diff
@@ -0,0 +1,89 @@
...
+ * AFFILIATES. All rights reserved. SPDX-License-Identifier: BSD-3-Clause
+ */
+// clang-format on
+#pragma once
+
+#include <stdint.h>
+
+#include <runtime/fusion_executor_cache.h>
+
+namespace nvfuser {
```
cowanmeg, line 12: This is just copy and pasted from tests/cpp/multidevice_transformer.cpp
No need to read the actual code, aside from the class structure.

tests/cpp/multidevice_transformer.cpp
```diff
@@ -0,0 +1,616 @@
+// clang-format off
```
cowanmeg, line 1: This is copied pasted from test_multidevice_transformer.cpp

PR #3201, opened 2024-10-16
Update multidevice transformer to use LinearOp

cowanmeg:
Multiple updates to C++ transformer tests
(1) Updates MLP and MHA layer to use LinearOp instead of Matmul+add when possible. Change the layout of weights to be in a transposed format to reduce transposes.
(2) Removes activation rematerializations of gemms in the backward tests.
(3) Also ups tolerances of failing tests.

### DIFF
--- a/tests/cpp/test_multidevice_transformer.cpp
+++ b/tests/cpp/test_multidevice_transformer.cpp
@@ -52,6 +52,10 @@ void validate(
     const std::vector<at::Tensor>& expected_outputs,
     const std::vector<at::Tensor>& outputs,
     const std::vector<double>& atols) {
+  for (auto i : c10::irange(expected_outputs.size())) {
+    std::cout << i << " " << outputs[i].sizes() << " " << expected_outputs[i].sizes() << std::endl;
+  }
+
   using testing::SizeIs;
   const auto num_outputs = outputs.size();
   ASSERT_THAT(expected_outputs, SizeIs(num_outputs));
@@ -119,9 +123,9 @@ std::vector<at::Tensor> reference_mlp(
     at::Tensor w1,
     at::Tensor b1) {
   auto at_dtype = w0.dtype();
-  auto linear0 = at::matmul(x, w0).to(at::kFloat) + b0;
+  auto linear0 = at::linear(x, w0, b0);
   auto gelu = at::gelu(linear0, "tanh").to(at_dtype);
-  auto linear1 = at::matmul(gelu, w1).to(at::kFloat) + b1;
+  auto linear1 = at::linear(gelu, w1, b1).to(at::kFloat);
   auto [dropout, mask] = at::native_dropout(linear1, kDropoutProb, true);
   return {linear0, gelu, linear1, dropout, mask};
 }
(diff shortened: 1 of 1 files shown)


tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -345,19 +315,17 @@ std::vector<TensorView*> mha(
     const DeviceMesh& mesh) {
   const auto D = w0->axis(0)->extent()->value().as<int64_t>();
   auto dtype = w0->dtype();
-  // Linear 0
-  TensorView* matmul0 = matmul(x, w0);
-  TensorView* linear0 = add(matmul0, broadcast(b0, {false, true, false}));
+
+  TensorView* linear0 = linear(x, w0, b0);
+  TensorView* linear0_temp = castOp(DataType::Float, linear0);
```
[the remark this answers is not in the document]
wujingyue: Thanks -- can you put this in code comment so it's easier to discover?

wujingyue:
> Updates MLP and MHA layer to use LinearOp

I remember you found that doing so [slows down](https://docs.google.com/document/d/1Lrx_lt09O-POJfIvzTrq-OnfREkYVIbk0ihJzw8amu0/edit?tab=t.0#bookmark=id.7vru8n2rs3ln) forward. Is that still the case?

cowanmeg:
Yes a slight slowdown in forward, but it removes transposes from the backward pass, so not really a perf hit

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -1120,72 +1086,65 @@ TEST_P(DistributedTransformerTest, Backward) {
   fusion->addInput(ln0_b);
   fusion->addInput(ln0_mean);
   fusion->addInput(ln0_rstd);
+  fusion->addInput(mha_linear0);
   fusion->addInput(mha_linear1);
+  fusion->addInput(mlp_linear0);
 
   // Recomputation: Recompute, mha linear0, qkv, mlp linear0, and mlp gelu.
```
wujingyue, line 1093: I believe this needs to be updated, e.g., mha_linear0 and mlp_linear0 will no longer be recomputed.

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -1095,17 +1062,16 @@ TEST_P(DistributedTransformerTest, Backward) {
   TensorView* ln0_b = makeContigTensor(1);
   TensorView* ln0_mean = makeConcreteTensor({B * S, 1});
   TensorView* ln0_rstd = makeConcreteTensor({B * S, 1});
+  TensorView* mha_linear0 = makeContigTensor(3, dtype);
   TensorView* mha_linear1 = makeContigTensor(2);
```
wujingyue, line 1066: Should this be dtype as well?

wujingyue (commented):
LGTM otherwise! Look forward to the updated timing!

tests/cpp/test_multidevice_transformer.cpp
cowanmeg: Ideally, yes mha_linear1 would also be dtype. But, we compute mha_linear1 as: local matmul (bfloat16) + allreduce (bfloat16) + sum (**float**) because nvfuser doesn't support bfloat16 reduction. 

I will open an issue about supporting bfloat16 reduction, since this is a pattern we will see. We can either (1) keep mha_linear1 as float to avoid the cast from bfloat16->float or (2) change the dtype and add the casts.

wujingyue: nvFuser doesn't support bfloat16 reduction **in codegen**. But I believe it supports bfloat16 reduction when it's **lowered to communication**. 

For example, https://github.com/NVIDIA/Fuser/blob/31d3a629bb954cec421404fea79a8aba362fb961/tests/python/test_multidevice.py#L423 has been using bfloat16 reduction just fine. I further confirmed this by running

```bash
$ git checkout wjy/shard
$ _bn && mpirun -np 2 -x NVFUSER_DUMP=fusion_ir_preseg pytest tests/python/test_multidevice.py -k 'backward' --only-mpi -s > /tmp/stdout
$ grep -B 1 -a -e 'reduction.*bfloat' /tmp/stdout
T65_l___bfloat[ rS212{2}, iS213{2048}, iS214{12288} ] (DeviceMesh{0 1})
   = reduction( T64_l___bfloat[ ideviceIdx.x208{2}, iS209{2048}, iS210{12288}, rS211{24576} ] (DeviceMesh{0 1}), op = add, initial value = __bfloat(0), allreduce = false )
--
T155_l___bfloat[ rS536{2}, iS537{2048}, iS538{12288} ] (DeviceMesh{0 1})
   = reduction( T154_l___bfloat[ ideviceIdx.x532{2}, iS533{2048}, iS534{12288}, rS535{18432} ] (DeviceMesh{0 1}), op = add, initial value = __bfloat(0), allreduce = false )
```

(I'm totally fine with pursing that in a separate PR)

wujingyue (approved):

samnordmann (approved):

tests/cpp/test_multidevice_transformer.cpp
cowanmeg: My bad, it's not the reduction that's the issue (this is in bfloat16 as you mentioned) it's the bias add.

cowanmeg:
!build

cowanmeg:
!build

cowanmeg:
!build

Issue #2979, opened 2024-09-20
Self-mapping error when compiling 1D bias linear fusion with smem-epilogue and no split-K


jacobhinkle:
The following test fails currently:
```c++
TEST_F(MatmulSchedulerTest, SelfMappingErrorSmemEpilogue1dBias) {
  NVFUSER_TEST_CUDA_ARCH_RANGE_GUARD(7, 5, 9, 0);

  Fusion fusion_obj;
  Fusion* fusion = &fusion_obj;
  FusionGuard fg(fusion);

  auto tv0 = makeContigTensor(2, DataType::Half);
  auto tv1 = makeContigTensor(2, DataType::Half);
  fusion->addInput(tv0);
  fusion->addInput(tv1);

  auto tv2 = makeContigTensor(1, DataType::Half);
  fusion->addInput(tv2);
  TensorView* tv3 = linear(tv0, tv1, tv2);

  fusion->addOutput(tv3);

  MatMulTileOptions gemm_tile;
  gemm_tile.cta_tile = GemmTile(64, 128, 32);
  gemm_tile.warp_tile = GemmTile(32, 32, 32);
  gemm_tile.instruction_tile = GemmTile(16, 8, 16);

  MatmulParams params;
  params.supported_vec_size = {8, 8, 4};
  params.mma_macro = MmaMacro::Ampere_16_8_16;
  params.tile_sizes = gemm_tile;
  params.async_gmem_load_operands = true;
  params.circular_buffer_options.circular_buffer_smem_write = true;
  params.circular_buffer_options.circular_buffer_smem_read = true;
  params.circular_buffer_options.smem_circular_buffer_stage = 2;
  mma_utils::MmaDataTypes data_types = {
      DataType::Half, DataType::Half, DataType::Float};
  std::tie(params.use_smem_epilogue, params.promote_prologue_smem_reuse) =
      mma_utils::generateSharedMemoryEpilogueHeuristics(
          gemm_tile,
          params.circular_buffer_options.smem_circular_buffer_stage,
          data_types,
          /*ignore_occupancy_drop=*/true);
  scheduleMatmul(fusion, params);

  auto options = at::TensorOptions().dtype(at::kHalf).device(at::kCUDA, 0);
  int64_t M = 256, N = 128, K = 64;
  at::Tensor t0 = at::randn({M, K}, options);
  at::Tensor t1 = at::randn({N, K}, options);
  at::Tensor t2 = at::randn({K}, options);
[...]

PR #3203, opened 2024-10-17
Run with and without half-precision matmul reduction in benchmark

jacobhinkle:
This changes the python matmul benchmark to run four times as many tests:
- We parametrize by reduction in float or in fp16/bf16, which is enabled by default in PyTorch.
- We parametrize by `eager`. If this is true we directly compute `torch.matmul` without involving nvFuser. Otherwise we use nvFuser. This lets us compute baselines in the same run as we compute the nvFuser result instead of needing to re-run the benchmark with different environment variables as we previously had to.

nvFuser does not support split-K in reduced precision (see #1719), so we skip these cases for now.

### DIFF
--- a/benchmarks/python/core.py
+++ b/benchmarks/python/core.py
@@ -129,15 +129,18 @@ class NVFBenchmark:
         """
         try:
             self.prof.stop()
-            prof_averages = self.prof.key_averages()
-            elapsed_cuda_time = self._get_kernel_time(prof_averages)
-            self._increment_global_time(elapsed_cuda_time)
-            # Clear the internal profiler object to avoid accumulating function events and then restart the profiler
-            # See PR: https://github.com/pytorch/pytorch/pull/125510
-            self.prof.profiler = None
-            self.prof.start()
         except AssertionError:
             self.prof.start()
+            return self.current_time
+
+        prof_averages = self.prof.key_averages()
+        elapsed_cuda_time = self._get_kernel_time(prof_averages)
+        self._increment_global_time(elapsed_cuda_time)
+        # Clear the internal profiler object to avoid accumulating function events and then restart the profiler
+        # See PR: https://github.com/pytorch/pytorch/pull/125510
+        self.prof.profiler = None
+        self.prof.start()
+
         return self.current_time
 
     def fusionprofile_timer(self) -> float:
@@ -157,22 +160,20 @@ class NVFBenchmark:
         Returns:
             time_value: Elapsed CUDA time in seconds.
         """
-        elapsed_cuda_time = (
-            sum(
-                [
-                    # Re: torch profiler API changes in https://github.com/pytorch/pytorch/pull/123247
-                    (
-                        event.self_device_time_total
-                        if hasattr(event, "self_device_time_total")
-                        else event.self_cuda_time_total
-                    )
-                    for event in prof_averages
-                    if event.device_type == DeviceType.CUDA
-                ]
+        elapsed_cuda_time = 0
+        has_cuda_event = False
+        for event in prof_averages:
+            if event.device_type != DeviceType.CUDA:
+                continue
+            has_cuda_event = True
+            # Re: torch profiler API changes in https://github.com/pytorch/pytorch/pull/123247
+            elapsed_cuda_time = (
+                elapsed_cuda_time + event.self_device_time_total
+                if hasattr(event, "self_device_time_total")
+                else event.self_cuda_time_total
             )
-            / 1e6
-        )
-        return elapsed_cuda_time
+        assert has_cuda_event, "No CUDA events found"
+        return elapsed_cuda_time / 1e6
 
     def _increment_global_time(self, elapsed_time: float) -> None:
         self.current_time += elapsed_time
--- a/benchmarks/python/core.py
+++ b/benchmarks/python/core.py
@@ -127,11 +127,11 @@ class NVFBenchmark:
         Returns:
             self.current_time: Global monotonic clock variable
         """
-        try:
-            self.prof.stop()
-        except AssertionError:
+        if self.prof.profiler is None:
             self.prof.start()
             return self.current_time
+
+        self.prof.stop()
 
         prof_averages = self.prof.key_averages()
         elapsed_cuda_time = self._get_kernel_time(prof_averages)
@@ -161,7 +161,7 @@ class NVFBenchmark:
             time_value: Elapsed CUDA time in seconds.
         """
         elapsed_cuda_time = 0
-        has_cuda_event = False
+        self.has_cuda_event = False
         for event in prof_averages:
             if event.device_type != DeviceType.CUDA:
                 continue
@@ -172,7 +172,6 @@ class NVFBenchmark:
                 if hasattr(event, "self_device_time_total")
                 else event.self_cuda_time_total
             )
-        assert has_cuda_event, "No CUDA events found"
         return elapsed_cuda_time / 1e6
 
     def _increment_global_time(self, elapsed_time: float) -> None:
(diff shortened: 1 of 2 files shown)


jacobhinkle:
!build --matmul-bench

Priya2698:
Do you think, we should separate out torch baselines from nvfuser benchmark to also have a marker for the default pytorch performance?

jacobhinkle:
> Do you think, we should separate out torch baselines from nvfuser benchmark to also have a marker for the default pytorch performance?

Yeah we could do that in order to track the effect of reduced precision reduction on perf in our baseline. In the separate torch baselines we would use `torch.matmul` directly instead of running through nvFuser right? That could work, but I think ideally we will start adding benchmarks involving fusions as well, in which case the point of comparison is to create an epilogue kernel but compute the matmul using aten, i.e. our default nvfuser path. I currently am running the "baseline" by running without env vars, then running the nvfuser version using `NVFUSER_ENABLE=fuse_matmul NVFUSER_DISABLE=matmul_expr_eval`, and a heuristic plugin.

> Like you said, pytorch for half reduction will be faster but less accurate, so I am not sure if we should be using default pytorch behavior for comparison (which is what nvfuser will be compared against by users), or use single precision for comparing with same settings.

My .02 is that users already have a knob to control this (the ones i'm turning in this PR), so we should merge something like #1719 and check for that value in ATen to decide the reduction dtype in our heuristic.

Priya2698:
> > Do you think, we should separate out torch baselines from nvfuser benchmark to also have a marker for the default pytorch performance?
> 
> Yeah we could do that in order to track the effect of reduced precision reduction on perf in our baseline. In the separate torch baselines we would use `torch.matmul` directly instead of running through nvFuser right? 

Yes -- running `torch.matmul` directly.

> My .02 is that users already have a knob to control this (the ones i'm turning in this PR), so we should merge something like #1719 and check for that value in ATen to decide the reduction dtype in our heuristic.

This looks more robust and we would cover all comparison metrics.

These are not blockers for this PR, so approving it. Once we have the matmul benchmarks in our CI, we should revisit this.

benchmarks/python/test_matmul.py
```diff
@@ -40,6 +40,11 @@ def test_matmul_nvf_benchmark(
 
     clear_cuda_cache()
 
+    # disable half-precision split-K reduction in cuBLAS since we do not
+    # support this in nvFuser
+    torch.backends.cuda.matmul.allow_fp16_reduced_precision_reduction = False
+    torch.backends.cuda.matmul.allow_bf16_reduced_precision_reduction = False
+
```
Priya2698, line 47: Can we add a comment on the other option of using PR #1719 to turn these knobs

Priya2698 (dismissed):
LGTM.

jacobhinkle:
> Looks like there are CI errors for matmul/linear translation tests.

Yes but the CI errors are not blocking. Those were there before this PR. What happens is the `--matmul-bench` option runs `test_matmul` with our internal heuristic plugin enabled, and there are some unrelated bugs in that plugin that are providing invalid configs currently.

Priya2698 (approved):
LGTM

Priya2698:
> > Looks like there are CI errors for matmul/linear translation tests.
> 
> Yes but the CI errors are not blocking. Those were there before this PR. What happens is the `--matmul-bench` option runs `test_matmul` with our internal heuristic plugin enabled, and there are some unrelated bugs in that plugin that are providing invalid configs currently.

Is there a tracking issue?

jacobhinkle:
> Is there a tracking issue?

#3213 and #2979

jacobhinkle:
In the latest version I am computing baselines in the same test (when `eager == True`), and I also parametrize by whether we reduce in half or full precision. This lets us lay everything out at once:
image
Thanks for the suggestion @Priya2698

Priya2698:
@jacobhinkle is this PR ready for review or do you first want to merge PR #3252

jacobhinkle:
> @jacobhinkle is this PR ready for review or do you first want to merge PR #3252

It is ready. I think the two PRs are pretty much independent.

benchmarks/python/core.py
```diff
@@ -157,22 +160,19 @@ def _get_kernel_time(
...
+        self.has_cuda_event = False
+        for event in prof_averages:
+            if event.device_type != DeviceType.CUDA:
+                continue
+            has_cuda_event = True
+            # Re: torch profiler API changes in https://github.com/pytorch/pytorch/pull/123247
+            elapsed_cuda_time = (
+                elapsed_cuda_time + event.self_device_time_total
+                if hasattr(event, "self_device_time_total")
+                else event.self_cuda_time_total
```
Priya2698, line 173: We are currently seeing tests with no CUDA events -- which should not be happening. Rather than returning 0, we may want to add an assert here `NVF_CHECK(has_cuda_event)` to identify these unexpected behaviors.

benchmarks/python/core.py
```diff
@@ -222,14 +222,15 @@ def set_metrics(
...
         self.benchmark.extra_info["IOBytes"] = iobytes
-        bandwidth_bps = (
-            iobytes * self.benchmark.stats["rounds"]
-        ) / self.benchmark.stats["total"]
-        self.benchmark.extra_info["Bandwidth (Bps)"] = bandwidth_bps
-        self.benchmark.extra_info["Bandwidth (GBps)"] = bandwidth_bps / 1e9
-        self.benchmark.extra_info["% Peak Bandwidth (SOL)"] = (
-            100 * (bandwidth_bps / 1e9) / PEAK_BANDWIDTH_GBPS
-        )
+        if self.has_cuda_event:
```
Priya2698, line 225: Similar to above, this feels like it will lead to silent errors. Benchmarking inaccurate readings is quite tedious given the number of inputs we run. I would make this an assert instead.

benchmarks/python/test_matmul.py
```diff
@@ -25,21 +25,32 @@ def load_matmul_problems():
...
+@pytest.mark.parametrize("eager", [False, True], ids=["nvfuser", "eager"])
+@pytest.mark.parametrize("dtype", [torch.float16, torch.bfloat16], ids=["fp16", "bf16"])
 @pytest.mark.parametrize(
-    "config", load_matmul_problems(), ids=lambda val: "_".join(str(v) for v in val)
+    "config", load_matmul_problems(), ids=lambda val: "-".join(str(v) for v in val)
 )
-@pytest.mark.parametrize("dtype", [torch.float16, torch.bfloat16])
 def test_matmul_nvf_benchmark(
     benchmark,
+    eager: bool,
```
Priya2698, line 36: If we separate `eager` benchmark into its own benchmark, they will not run by default (https://github.com/NVIDIA/Fuser/blob/5a6c90b17c950a922697ca8f7955f8c1c3cb55ed/benchmarks/python/conftest.py#L107-L116). Right now, we will run both nvfuser (which is ATen at this time) and eager in every run.

We can separate out the common code into a utility function and call them from `test_matmul_nvf_benchmark` and `test_matmul_baseline_benchmark`.

benchmarks/python/core.py
jacobhinkle: Yeah, I will add that assert back. I had removed the assert here because I was trying to get automatic calibration working. Since I never got that to work properly anyway and actually reflect the CUPTI precision properly, I think it's fine to just assert that if we're using this timer then there should always be some CUDA events between observations.

benchmarks/python/test_matmul.py
jacobhinkle: When I split the test into two tests there will be no difference between eager and compile, but my understanding is that we need to have both of those to trigger this code path.

jacobhinkle: I guess I can just put that `compile` option in, since eventually we will probably extend this to cover some epilogue cases, and multi-matmul cases

Priya2698: The second function will have the compile parameter with just the `[False]` value to exercise the eager benchmark and it will not be run by default.

>there will be no difference between eager and compile, but my understanding is that we need to have both of those to trigger this code path.

I am not sure what you mean here -- trigger which code path?

Priya2698: The `compile` parameter is what is required to skip eager by default.

jacobhinkle: Ah right. Gotcha. I will make that change tonight/AM eastern

jacobhinkle: I just pushed the change splitting the baseline into a separate test that can be enabled with `--benchmark-eager`.

jacobhinkle:
!build

benchmarks/python/test_matmul.py
```diff
@@ -25,21 +25,73 @@ def load_matmul_problems():
...
+        a = torch.randn(m, k, device="cuda", dtype=dtype)
+        b = torch.randn(k, n, device="cuda", dtype=dtype)
+
+        if layout == "NT" or layout == "NN":
+            a = a.as_strided(size=[m, k], stride=[1, m])
+        if layout == "TN" or layout == "NN":
+            b = b.as_strided(size=[k, n], stride=[1, k])
+
+        # NOTE: we never need to validate eager, as it is our baseline
+        if not disable_benchmarking:
```
Priya2698, line 60: This conditional is not needed. This flag is for nvfuser benchmarks. I'll make a note to rename this. If --benchmark-eager is used we always run this benchmark.

Priya2698 (approved):
LGTM apart from the disable-benchmarking flag in the eager benchmark that needs to be removed.

Thanks for the changes!

jacobhinkle:
!build

Issue #2499, opened 2024-06-28
Some ranks failed to wait for the barrier.


wujingyue:
I managed to narrow down a non-deterministic barrier hang into the following reliable reproducer. 
```
TEST_P(CommunicationTest, Barrier) {
  const auto device_id = communicator_->deviceId();
  for (int i = 0; i < 4; i++) {
    sleep(device_id);
    LOG(INFO) << "[Rank " << device_id << "] enter barrier " << i;
    communicator_->barrier();
    LOG(INFO) << "[Rank " << device_id << "] exit barrier " << i;
  }
}
```

To run this, check out `wjy/barrier` and `_bn && python setup.py develop --debug && TORCH_CPP_LOG_LEVEL=INFO mpirun -np 3 bin/test_multidevice --gtest_filter=CommunicationTest.Barrier/NCCL`. Note two GPUs weren't enough to reproduce this, and therefore `-np 3`. 

I printed out the time that each rank enters and exits the barrier: 

| timestamp | rank 0 | rank 1 | rank 2 |
| --- | --- | --- | --- |
| :27       | enter 0|        | |
| :28       |        | enter 0| |
| :29       |        |        | enter 0 |
| :30       | exit 0 | exit 0 | exit 0 |
| :30       | enter 1|        | |
| :31       |        | enter 1| |
| :31       |        | exit 1 | |
| :32       |        | enter 2| enter 1 |
| :32       | exit 1 | exit 2 | exit 1 |
| :32       | enter 2|        | |
| :33       |        | enter 3| |
| :33       |        | exit 3 | |
| :34       |        |        | enter 2 |
| :34       | exit  2|        | exit 2 |
| :34       | enter 3|        | |
| :36       |        |        | enter 3 |
| :36       |        |        | exit 3 |

Rank 0 hangs and eventually times out. 

Observations: Rank 0 and rank 2 exits the same barrier at the same time, which is expected behavior. However, rank 1 is range free and doesn't synchronize except for the first iteration. 

FWIW, some warnings got printed out at the beginning mentioning a potential hang. However, the guessed mapping seems to be correct so I doubt that's the reason. 

```
[I628 23:23:24.590086547 ProcessGroupNCCL.cpp:3905] [PG  Rank 0]  using GPU 0 to perform barrier as devices used by this process are currently unknown. This can potentially cause a hang if this rank to GPU mapping is incorrect.Specify device_ids in barrier() to force use of a particular device.
[...]

PR #3206, opened 2024-10-17
Print duration as seconds for better error message.

wujingyue:
The test failed
[once](https://gitlab-master.nvidia.com/dl/pytorch/fuser-gh-mirror/-/jobs/116949080/viewer), but it's unclear from the error message how off the expectations are.

### DIFF
--- a/tests/cpp/test_multidevice_communicator.cpp
+++ b/tests/cpp/test_multidevice_communicator.cpp
@@ -23,8 +23,6 @@ void PrintTo(const std::chrono::duration<double>& d, ostream* os) {
 } // namespace chrono
 } // namespace std
 
-namespace {} // namespace
-
 namespace nvfuser {
 
 using testing::PrintToString;
--- a/tests/cpp/test_multidevice_communicator.cpp
+++ b/tests/cpp/test_multidevice_communicator.cpp
@@ -6,13 +6,28 @@
  */
 // clang-format on
 #include <chrono>
+#include <ostream>
 
+#include <gmock/gmock-matchers.h>
 #include <gtest/gtest.h>
 
 #include <multidevice/communicator.h>
 #include <tests/cpp/multidevice.h>
 
+namespace std {
+namespace chrono {
+// Without this, EXPECT_* would print duration as bytes.
+void PrintTo(const std::chrono::duration<double>& d, ostream* os) {
+  *os << d.count() << " seconds";
+}
+} // namespace chrono
+} // namespace std
+
+namespace {} // namespace
+
 namespace nvfuser {
+
+using testing::PrintToString;
 
 class CommunicatorTest
     : public MultiDeviceTest,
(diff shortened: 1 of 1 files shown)


wujingyue:
!build

samnordmann (commented):
1) I am looking for the first time at how this test is implemented and I don't think it is relevant. Passing/failing the test is neither a necessary nor sufficient condition for the barrier's success/failure. The fact that it failed [here](https://gitlab-master.nvidia.com/dl/pytorch/fuser-gh-mirror/-/jobs/116949080/viewer) proves this point. The test will probably fail more frequently as we move to larger multi-nodes systems. I suggest to completely remove this test, which only adds maintenance debts without testing what it is intended to.
2) Isn't there a more native way to print time_point than manually converting it to double? I'm surprised there is no `<<` operator for this type.
3) If the unit matters, then we should make it explicit in the error message: even with this patch the reader won't be able to know that it is in seconds
4) Shouldn't we make the comparison as time_point, before conversion to double, which is only here for pretty printing?

wujingyue:
> Passing/failing the test is neither a necessary nor sufficient condition for the barrier's success/failure.

I don't think I follow this. Anyhow, for the context, this was added as a regression test for https://github.com/pytorch/pytorch/pull/129908. 

> Casting to double seems like a heavy hammer here.

Agreed -- I hope C++20 will make this easier: https://en.cppreference.com/w/cpp/chrono/duration/operator_ltlt.

wujingyue:
> What I mean is that the unit test does not test the barrier collective

I don't follow that. This test verifies https://github.com/NVIDIA/Fuser/issues/2499, a bug in "the barrier collective" fixed by 
@eqy's https://github.com/pytorch/pytorch/pull/129908, doesn't show up again.

samnordmann:
> > What I mean is that the unit test does not test the barrier collective
> 
> I don't follow that. This test verifies #2499, a bug in "the barrier collective" fixed by @eqy's [pytorch/pytorch#129908](https://github.com/pytorch/pytorch/pull/129908), doesn't show up again.

yes, but it is only incidental

naoyam:
@samnordmann 

> yes, but it is only incidental

The original issue of some rank missing a proper synchronization doesn't seem just incidental, so the test itself seems to make sense to me. 

This particular failure seems to be due to how we verify the barrier behavior, which may not be very robust as Sam pointed out, but for now if it works on the system scale we usually test and benchmark, I'd say it's good enough.

One way to improve the robustness may be something like: let rank0 spin wait for 1ms (or some reasonably longer time than actual barrier latency) and call `barrier`, while the other ranks immediately call `barrier`. The barrier latency should be larger than 1ms on all ranks. This way, missing a proper synchronization should be always visible no matter how noisy the actual barrier implementation would be.

samnordmann (approved):
Thanks for addressing the comments!

wujingyue:
!build

PR #3209, opened 2024-10-17
Fix sdpfa_bwd to not assume the presence of DIDs

wujingyue:
Similar to https://github.com/NVIDIA/Fuser/pull/3116.

### DIFF
--- a/csrc/ops/composite.cpp
+++ b/csrc/ops/composite.cpp
@@ -534,6 +534,20 @@ SdpfaFwdResult sdpfa_fwd(
   return {output, log_sumexp, philox_seed, philox_offset};
 }
 
+namespace {
+template <typename T>
+void checkAllEqual(std::initializer_list<T> elements) {
+  for (const auto& element : elements) {
+    NVF_CHECK(
+        element == *elements.begin(),
+        "Expected all elements to be equal, but found ",
+        element,
+        " and ",
+        *elements.begin());
+  }
+}
+} // namespace
+
 SdpfaBwdResult sdpfa_bwd(
     TensorView* grad_output,
     TensorView* query,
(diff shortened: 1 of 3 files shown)


wujingyue:
!build

csrc/ops/composite.cpp
```diff
@@ -546,48 +560,39 @@ SdpfaBwdResult sdpfa_bwd(
...
-  NVF_CHECK(
-      query->dtype() == key->dtype() && query->dtype() == value->dtype(),
-      "Expected query, key, and value to have the same dtype but got: ",
-      query->dtype(),
-      " ",
-      key->dtype(),
-      " ,and ",
-      value->dtype());
-
+  checkAllEqual(
```
wujingyue, line 563: This and the checkAllEqual at line 576 will harden the checks. Currently, the code checks only Q, K and V.

csrc/ops/composite.cpp
```diff
@@ -534,6 +534,20 @@ SdpfaFwdResult sdpfa_fwd(
   return {output, log_sumexp, philox_seed, philox_offset};
 }
 
+namespace {
+template <typename T>
+void checkAllEqual(std::initializer_list<T> elements) {
```
Priya2698, line 539: Nice template, we can propagate this to `sdpfa_fwd` above as well (does not have to be this PR)

Priya2698 (approved):
It should be possible to use https://github.com/NVIDIA/Fuser/blob/aec6797d6386cbfe02c1b5405b7e0a9439915295/tests/python/utils.py#L387 here as noted in comment [here](https://github.com/NVIDIA/Fuser/pull/3116#discussion_r1790715297). See `test_python_frontend.py` for usage example.
I don't think this switch is a blocker for this PR if you decide to go this route and can be done separately.

LGTM once the CI error is resolved.

wujingyue:
> It should be possible to use
> 
> https://github.com/NVIDIA/Fuser/blob/aec6797d6386cbfe02c1b5405b7e0a9439915295/tests/python/utils.py#L387

Alas, this doesn't quite work out, and I created https://github.com/NVIDIA/Fuser/issues/3214.

csrc/ops/composite.cpp
wujingyue: Done!

wujingyue:
!build

PR #3215, opened 2024-10-18
Simplify cat implementation

naoyam:
`cat` is translated to CUDA code with a if-then-else block:

```
if (i < input0_ext) {
  out[idx] = input0[idx];
} else if (i < input0_ext + input1_ext) {
  out[idx] = input1[idx];
} else if ...
```

This is correct, but I believe this can be simplified to just `out[idx] = input0[idx] + input1[idx] + ...` since all of the inputs are padded by zero, so this result should be equivalent.

Since `+` is not defined for some low precision types, bitwise-or is instead used when addition is not available.

On A100, this simplification yielded about 5% perf improvement of a RoPE module.

I didn't add any specific test since I don't know if any new test would be beneficial. Half, bfloat16 and fp8 types are tested by `ResizeTest.CatMemoryPromotionReducedFloating`.

### DIFF
--- a/runtime/fp8_support.cu
+++ b/runtime/fp8_support.cu
@@ -73,12 +73,11 @@ struct __align__(1) __e4m3 {
     __x = __double2e4m3(f).__x;
   }
 
-  __device__ __e4m3(const int x): __x(x) {}
+  __device__ __e4m3(const int x) : __x(x) {}
 
-  __device__ __e4m3(const uint8_t x): __x(x) {}
+  __device__ __e4m3(const uint8_t x) : __x(x) {}
 
-  __device__ __e4m3(const uint16_t x): __x(x) {}
-
+  __device__ __e4m3(const uint16_t x) : __x(x) {}
 
   __device__ uint8_t raw() const {
     return __x;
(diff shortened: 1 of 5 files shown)


runtime/fp8_support.cu
```diff
@@ -195,8 +207,19 @@ __device__ __inline__ __bfloat __e4m32bfloat(const __e4m3 h) {
   return val;
 }
 
+__device__ __inline__ __e4m3 operator|(const __e4m3 x, const __e4m3 y) {
```
naoyam, line 210: Not exactly sure if this is reasonable. I don't know why `memcpy` is extensively used for these fp8 types. 

Pinging @jjsjann123

naoyam:
!build

runtime/fp8_support.cu
jjsjann123: This does look cleaner :)

jacobhinkle (approved):
LGTM. I wonder if we could handle all types (other than ComplexDouble) by bitcasting to an integer of the right size then doing bitwise or. For this we would need Int8 and Int16, but we would not need support in the runtime files for bitwise ops on floats then.

runtime/fp8_support.cu
naoyam: I thought there may be some magic with `memcpy`.

naoyam:
> LGTM. I wonder if we could handle all types (other than ComplexDouble) by bitcasting to an integer of the right size then doing bitwise or. For this we would need Int8 and Int16, but we would not need support in the runtime files for bitwise ops on floats then.

Yeah, I did think about it, but I found it's just easier to handle these low precision types separately than adding the new integer types, at least for now.

PR #3217, opened 2024-10-18
always prioritize using smem to store inner persistent buffers

liqiangxl:
**Issue:** In the current innerOuter persistent scheduler, inner persistent buffers are used for inner normalization, outer persistent buffers are used for outer reduction. Inner persistent buffers are usually cached input tvs and loaded from global memory to shared memory using `CpAsync`, outer persistent buffers are used to accumulating outer reduction results. These buffers are stored in registers and moved to shared memory only when there are not sufficient registers for all of them. This leads to high register pressure and performance is highly sensitive to register usage and spilling, where even a small increase in register spills can result in a 20% performance drop or improvement, as seen in issue https://github.com/NVIDIA/Fuser/issues/2741. This sensitivity makes it challenging to develop robust, high-performance heuristics.

**Fix:**   Put all inner persistent buffers to shared memory since loading from global memory uses async copy which bypasses L1 cache and won't use tmp registers. Outer reduction buffers are used to accumulate partial results of the outer reduction. They  are not loaded from global memory and requires frequent read/write. So, they are always stored in registers.

**Results:**
image

**Following works**
https://github.com/NVIDIA/Fuser/pull/3223 (Not rebased yet), Heuristic revision to fix regressions.


liqiangxl:
!build

liqiangxl:
Performance regression is further fixed in #3221 and #3223

naoyam:
IIUC, suppose a fusion has two buffers. If the combined size is small enough, only registers are used, but if it exceeds the register limi, both of them are moved to shmem because using shmem is more efficient than registers. 

Is my understanding correct? If so, why not using shmem from the beginning and only use registers when necessary?

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -284,21 +286,25 @@ PersistentBufferStorageParams getPersistentBufferStorageParams(
...
     // If both the register buffer size and shared memory buffer size are within
     // the allowable limit, we found a good configuration. Record the number of
-    // buffers to be moved to shared memory and break the loop.
-    if (buffer_params.regs_buffer_size <= available_regs &&
-        buffer_params.smem_buffer_size <= available_smem) {
+    // buffers to be moved to shared memory and continue checking if we can move
+    // more. Moving more buffers to shared memory reduces the usage of
+    // registers, the comipler can uses more register to optimize the kernel for
+    // better instruction level parallelism and leads to better performance.
+    if (regs_buffer_size <= available_regs &&
```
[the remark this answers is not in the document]
liqiangxl: It is still necessary. We begin with `{regs_buffer = N, smem_buffer = 0}`. In each iteration, one buffer is moved from registers to shared memory, resulting in `{regs_buffer = N-x, smem_buffer = x}`. We must check the conditions `{N-x <= available_regs}` and `{x <= available_smem}`. For example, when `x > available_smem`, we should stop moving buffers from registers to shared memory. The loop continues until no more buffers can be moved to shared memory. Otherwise, we may encounter invalid states, such as `{regs_buffer > available_regs, smem_buffer > available_smem}` or {regs_buffer < available_regs, smem_buffer > available_smem}.

In summary, this process achieves two goals:
(1) Ensures that as many register buffers as possible are moved to shared memory.
(2) Ensures there are enough registers and shared memory to store the buffers.

Also, as you suggested, we may start with `{regs_buffer = 0, smem_buffer = N}`, and move from smem to registers. This seems simplier, let me give it a shot.

liqiangxl:
> IIUC, suppose a fusion has two buffers. If the combined size is small enough, only registers are used, but if it exceeds the register limi, both of them are moved to shmem because using shmem is more efficient than registers.
> 
> Is my understanding correct? If so, why not using shmem from the beginning and only use registers when necessary?

Yes, your understanding is correct and we can use shmem from the beginning and only use registers when necessary. The results should be same to https://github.com/NVIDIA/Fuser/pull/3221 where smem is prioritized over register. The logic is simplier. Thanks for this great suggestion.

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -238,33 +240,53 @@ PersistentBufferStorageParams getPersistentBufferStorageParams(
...
   }
 
-  // (3) Relocate buffers to shared memory until the buffer size in registers is
-  // within the allowable limit.
+  // Moving outer reduction buffer to shared memory is not considered yet,
+  // set to false if the outer reduction buffer size exceeds the register size.
+  if (buffer_params.regs_buffer_size > available_regs) {
+    buffer_params.has_enough_regs_and_smem = false;
+    return buffer_params;
+  }
```
liqiangxl, line 277: This `if` branch will be removed in #3221. It should be removed with a slight change of the heuristics to avoid large regressions.

liqiangxl:
Current logic is:
(1) Put all buffers in registers
(2) If not enough registers, put all cached input buffers to shared memory and outer reduction buffers to registers
(3) If not enough registers, return false, will be segmented
(4) if not enough shared memory, move cached input buffers from shared memory to registers.

liqiangxl:
Achieved same results using the new logic (green cross markers).
image

liqiangxl:
!build --pybench

naoyam:
> Current logic is: (1) Put all buffers in registers (2) If not enough registers, put all cached input buffers to shared memory and outer reduction buffers to registers (3) If not enough registers, return false, will be segmented (4) if not enough shared memory, move cached input buffers from shared memory to registers.

I'm a bit confused since you seemed to agree with my suggestion. https://github.com/NVIDIA/Fuser/pull/3217#issuecomment-2429588644

Why put all buffers in registers? Why not starting all of them in shmem and move to registers as needed?

liqiangxl:
> > Current logic is: (1) Put all buffers in registers (2) If not enough registers, put all cached input buffers to shared memory and outer reduction buffers to registers (3) If not enough registers, return false, will be segmented (4) if not enough shared memory, move cached input buffers from shared memory to registers.
> 
> I'm a bit confused since you seemed to agree with my suggestion. [#3217 (comment)](https://github.com/NVIDIA/Fuser/pull/3217#issuecomment-2429588644)
> 
> Why put all buffers in registers? Why not starting all of them in shmem and move to registers as needed?

`starting all of them in shmem and move to registers` is done in PR https://github.com/NVIDIA/Fuser/pull/3221.The reason is that after move, the regiser usage is low, max allowed persistent batch size without causing register spills is increased, the current heuristics always use the largest possible persistent batch size which leads to lower performance in some hidden sizes. In PR-3221, it is done with a change of heuristics to avoid large regressions.

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -238,80 +240,108 @@ PersistentBufferStorageParams getPersistentBufferStorageParams(
...
-  // (2) If the available register is larger than current register buffer size,
-  // no need to move buffers to shared memory, return early.
-  if (buffer_params.regs_buffer_size <= available_regs) {
+  // (1) Try to put all the buffers into registers.
+  if (total_buffer_size <= available_regs) {
+    buffer_params.regs_buffer_size = total_buffer_size;
+    buffer_params.smem_buffer_size = 0;
     buffer_params.has_enough_regs_and_smem = true;
     return buffer_params;
   }
```
liqiangxl, line 249: This `if` branch will be removed in #3221. It should be removed with a slight change of the heuristics to avoid large regressions.

naoyam:
> > > Current logic is: (1) Put all buffers in registers (2) If not enough registers, put all cached input buffers to shared memory and outer reduction buffers to registers (3) If not enough registers, return false, will be segmented (4) if not enough shared memory, move cached input buffers from shared memory to registers.
> > 
> > 
> > I'm a bit confused since you seemed to agree with my suggestion. [#3217 (comment)](https://github.com/NVIDIA/Fuser/pull/3217#issuecomment-2429588644)
> > Why put all buffers in registers? Why not starting all of them in shmem and move to registers as needed?
> 
> `starting all of them in shmem and move to registers` is done in PR #3221 reason is that after move, the regiser usage is low, max allowed persistent batch size without causing register spills is increased, the current heuristics always use the largest possible persistent batch size which leads to lower performance in some hidden sizes. In PR-3221, it is done with a change of heuristics to avoid large regressions.

Still unclear what you mean. In #3221, I still see code like:

```
// (1) init the buffer_params by putting all the persistent tensors in
  // registers
```
https://github.com/NVIDIA/Fuser/pull/3221/files#diff-1b9c932dfa154070bef4f9c75d30bb331ff35abd54063e020bf9f35d901e224aR241

liqiangxl:
> > > > Current logic is: (1) Put all buffers in registers (2) If not enough registers, put all cached input buffers to shared memory and outer reduction buffers to registers (3) If not enough registers, return false, will be segmented (4) if not enough shared memory, move cached input buffers from shared memory to registers.
> > > 
> > > 
> > > I'm a bit confused since you seemed to agree with my suggestion. [#3217 (comment)](https://github.com/NVIDIA/Fuser/pull/3217#issuecomment-2429588644)
> > > Why put all buffers in registers? Why not starting all of them in shmem and move to registers as needed?
> > 
> > 
> > `starting all of them in shmem and move to registers` is done in PR #3221 reason is that after move, the regiser usage is low, max allowed persistent batch size without causing register spills is increased, the current heuristics always use the largest possible persistent batch size which leads to lower performance in some hidden sizes. In PR-3221, it is done with a change of heuristics to avoid large regressions.
> 
> Still unclear what you mean. In #3221, I still see code like:
> 
> ```
> // (1) init the buffer_params by putting all the persistent tensors in
>   // registers
> ```
> 
> https://github.com/NVIDIA/Fuser/pull/3221/files#diff-1b9c932dfa154070bef4f9c75d30bb331ff35abd54063e020bf9f35d901e224aR241

Oh, my bad, that PR was not updated yesterday, it was too late in the night.

After changing to your suggested approach, I feel it is more appropriate to re-organise the PRs into the following two PRs. 
(1) This one, PR-3217, always use shared memory to store persistent buffers. Will cause some regressions, see the following figure.
image

(2) https://github.com/NVIDIA/Fuser/pull/3223 (Not rebased yet), Heuristic revision to fix regressions.

We no longer needs #3221, it is splited into two parts, the first part is merged into PR-3217 (this PR, already done) and the 2nd part can be merged into PR-3223 (not done yet).

csrc/scheduler/normalization_inner_outer.cpp
```diff
@@ -238,80 +232,100 @@ PersistentBufferStorageParams getPersistentBufferStorageParams(
...
-  // (3.2) Before this loop, all buffers are in registers.
-  // Try to move buffer from register to shared memroy.
-  // After one buffer is moved to shared memory, the buffer size in registers
-  // and shared memory are updated accordingly. Break if required register and
-  // shared memory are lower than limit or shared memory exceeds the limit.
-  int64_t n_smem_buffer = -1;
+  // (2) Now, shared memory is overused, move some buffers to registers.
+  // (2.1) Sort the candidate persistent buffers
+  if (buffer_params.project_to_input) {
+    buffers = sortProjectableBufferInputs(buffers, outer_broadcast_tvs);
```
naoyam, line 291: Could you remind me why sorting is done only when projected to inputs?

naoyam:
Can you please add some unit tests?

csrc/scheduler/normalization_inner_outer.cpp
liqiangxl: (1) This condition is not unclear. It should be revised to if(!outer_broadcast_tvs.empty()), making it clear that sorting is only needed when outer broadcast tensors are present, as the sorting depends on whether a buffer is a producer of the outer broadcast tensor.
(2) IIRC, it was used to indicate the presence of an outer broadcast tensor in the fusion. When there is an outer broadcast tensor, this scheduler projects the buffer to the inputs, even if it results in a larger buffer size. 
(3) I double-checked the effect of with or without projecting to inputs for fp32 after adding async copy. The conclusion remains the same: projecting to inputs is faster, even with a larger buffer size (3 fp32 when projecting, versus 2 fp32 when not) unless it leads to segmentation (we can patch this latter).
image

liqiangxl:
> Can you please add some unit tests?


Needs some code movement to expose `getPersistentBufferStorageParams`, so I made a separate PR  https://github.com/NVIDIA/Fuser/pull/3264. After review, that PR will be merged into this one and together goes to the main branch. Similarly, we can also merge https://github.com/NVIDIA/Fuser/pull/3223 into this one so the regressions are avoided in the main branch.

naoyam (approved):
LGTM. Just in case, please check the CI again as the code seems to have changed since the last time the CI is done.

liqiangxl:
> LGTM. Just in case, please check the CI again as the code seems to have changed since the last time the CI is done.

Yes. CI did found two hidden bugs:
(1) async copy of tensor with data type `bool`, fixed in https://github.com/NVIDIA/Fuser/pull/3274
(2) vectorization factor of shared memory consumers, fixed in https://github.com/NVIDIA/Fuser/pull/3271

liqiangxl:
!test --pybench

liqiangxl:
!test --pybench

PR #3226, opened 2024-10-18
should unroll the outer dim for 2D pointwiser scheduler 

liqiangxl:
**Issue:** The 2D pointwise scheduler is used to save loaded data size when there are outer broadcast TensorViews. It should unroll along the outer dimension to reuse data across different rows. The previous implementation mistakenly unrolled the inner dimension.

**Fix:**
(1) Fix unroll dim
(2) Test number of blocks is reduced by unroll factor

### DIFF
--- a/csrc/scheduler/pointwise.cpp
+++ b/csrc/scheduler/pointwise.cpp
@@ -781,15 +781,15 @@ void schedulePointwise(Fusion* fusion, const PointwiseParams* pparams) {
           unswitch_pos = 3;
         }
       } else {
-        // [BIDx | BIDy | Unswitch, Unroll, TIDx, vect]
+        // [BIDx | BIDy | Unswitch, Unroll, TIDx, Vect]
         reference_tv->axis(0)->parallelize(ParallelType::BIDx);
         if (pparams->split_grid_y_dim) {
-          // [BIDx | i-remainder, BIDy{65535} | Unswitch, Unroll, TIDx, vect]
+          // [BIDx | i-remainder, BIDy{65535} | Unswitch, Unroll, TIDx, Vect]
           reference_tv->split(1, 65535);
           reference_tv->axis(2)->parallelize(ParallelType::BIDy);
           unswitch_pos = 4;
         } else {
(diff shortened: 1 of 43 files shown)


csrc/scheduler/pointwise.cpp
```diff
@@ -704,16 +704,19 @@ void schedulePointwise(Fusion* fusion, const PointwiseParams* pparams) {
       if (pparams->vectorization_factor > 1) {
         reference_tv->split(1, pparams->vectorization_factor);
       }
-      // [outer | inner/vect, vect]
+      // [outer | i-remainder, Vect]
       reference_tv->split(1, NamedScalar::getParallelDim(ParallelType::TIDx));
-      // [outer | inner/vect/TIDx, TIDx, vect]
-      reference_tv->split(1, pparams->unroll_factor);
```
liqiangxl, line 710: Here, shoud unroll outer dim.

rdspring1 (approved):
LGTM.

Should we add a comment for the `unroll_factor` argument?
https://github.com/NVIDIA/Fuser/blob/main/csrc/scheduler/pointwise_heuristic.h#L44-L45

liqiangxl:
> LGTM.
> 
> Should we add a comment for the `unroll_factor` argument? https://github.com/NVIDIA/Fuser/blob/main/csrc/scheduler/pointwise_heuristic.h#L44-L45

added
```
  // Unroll on top of vectorization
  // In the 2D scheduler, unroll the outer dimension to reuse loaded data across
  // rows, reducing loaded bytes by the unroll factor.
```

liqiangxl:
!build

liqiangxl:
> **Issue:** The 2D pointwise scheduler is used to save loaded data size when there are outer broadcast TensorViews. It should unroll along the outer dimension to reuse data across different rows. The previous implementation mistakenly unrolled the inner dimension.
> 
> **Fix:** (1) Fix unroll dim (2) Test number of blocks is reduced by unroll factor

Actually, the original implementation should be used when we have inner dim broadcast. The best position to unroll depends on the actual problem and should be set by the heuristics.

PR #3230, opened 2024-10-19
Host benchmarking for many-segment fusion

Priya2698:
This adds host benchmarking for a fusion with 11 segments aiming at measuring host latency for highly-segmented fusions. We use `segment_set` to create multiple segments with pointwise ops.
Screenshot 2024-10-21 at 2 43 30 PM

### DIFF
--- a/benchmarks/python/test_many_segment_host.py
+++ b/benchmarks/python/test_many_segment_host.py
@@ -1,0 +1,79 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-present NVIDIA CORPORATION & AFFILIATES.
+# All rights reserved.
+# SPDX-License-Identifier: BSD-3-Clause
+import pytest
+from nvfuser import FusionDefinition, DataType
+from nvfuser.pytorch_utils import clear_cuda_cache
+from .core import run_benchmark
+import torch
+
+
+def many_segment_fusion(fd: FusionDefinition) -> None:
+    x = fd.define_tensor(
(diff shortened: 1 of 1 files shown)


naoyam:
Just want to make sure, when a fusion has multiple segments, the reported host latency includes everything involved in executing all the segments including CPU times, right?

Priya2698:
> Just want to make sure, when a fusion has multiple segments, the reported host latency includes everything involved in executing all the segments including CPU times, right?

Yes, this is the total CPU overhead.

benchmarks/python/test_many_segment_host.py
```diff
@@ -0,0 +1,79 @@
...
+    y = fd.define_tensor(
+        shape=[-1], contiguity=[True], dtype=DataType.Half, is_cpu=False
+    )
+
+    x = fd.ops.cast(x, dtype=DataType.Float)
+    y = fd.ops.cast(y, dtype=DataType.Float)
+
+    a = fd.ops.add(x, y)
+    # Generate multiple segments using segment_set
+    for _ in range(10):
```
naoyam, line 24: Just curious, why 10?

Priya2698: It is an arbitrary choice. I wanted to create enough number of segments to capture the overhead well while the benchmark also runs in a reasonable time.

naoyam (approved):
LGTM

Priya2698:
Will be closed in favor of #3307 once it is merged.

Priya2698:
Closing in favor of #3307

PR #3232, opened 2024-10-19
Shard backprop. 

wujingyue:


wujingyue:
!build

wujingyue:
!build

csrc/ir/nodes.cpp
wujingyue, line 1: I'm happy to factor this change into a separate PR; I chose to leave it here because it's small and doesn't change behavior.

csrc/python_frontend/fusion_record.h
wujingyue, line 1: ditto

wujingyue:
!build

tests/python/test_multidevice.py
```diff
@@ -800,30 +816,30 @@ def definition(self) -> None:
...
             contiguity=True,
             dtype=DataType.BFloat16,
         )
         self.sdpa_out = self.define_tensor(
-            shape=[b, h, s, e // h],
+            shape=[d, b, h // d, s, e // h],
             contiguity=True,
             dtype=DataType.BFloat16,
-            stride_order=[3, 1, 2, 0],
+            stride_order=[4, 3, 1, 2, 0],
```
cowanmeg, line 832: I probably missed this from a past PR, but why does sdpa's output need to have the stride order explictly defined? This looks like what would be the default stride order

wujingyue: I'm pretty sure it's a Thunder artifact. The `.contiguous()` call [here](https://github.com/Lightning-AI/lightning-thunder/blob/79e59d0c5c5f8aa8ef80eb31f3fe918466d64c1c/thunder/tests/nanogpt_model.py#L107) enforces the strider order and is translated into a `prims.stride_order`:
https://gist.github.com/wujingyue/b111aa8b8d92067fc6004f5d0488dd27#file-forward-L54. For some reasons, rematerialization chose to save [`y`](https://gist.github.com/wujingyue/b111aa8b8d92067fc6004f5d0488dd27#file-forward-L52), which is in BHSE format, instead of the output of `stride_order`, which is in BSHE format.

wujingyue:
!build

PR #3233, opened 2024-10-19
Fix fd.ops.matmul's ndim computation.

wujingyue:


wujingyue:
!build

csrc/python_frontend/python_bindings.cpp
```diff
@@ -1520,7 +1520,8 @@ void initNvFuserPythonBindings(PyObject* module) {
         NVF_CHECK(                                                             \
             self.validUse(), "Attempting to add to a completed definition!");  \
         FusionDefinition* fd = self.fusion_definition;                         \
-        Tensor output = fd->defineTensor(arg1.dims);                           \
+        /* ops.matmul can do batch matmul. */                                  \
+        Tensor output = fd->defineTensor(std::max(arg1.dims, arg2.dims));      \
```
jjsjann123, line 1524: I'm a bit unsure about this one. I *think* we need to map this behavior https://github.com/NVIDIA/Fuser/blob/3d54c06926ce167f201c16fd3899a940c0f54f6b/csrc/ops/composite.cpp#L367-L376

jjsjann123 (approved):
I'm stamping this for now, since it's net improvement.

csrc/python_frontend/python_bindings.cpp
wujingyue: Thanks -- I'll make it consistent with https://pytorch.org/docs/stable/generated/torch.matmul.html. 

Is Tensor.dims supposed to include reduction dimensions? Seems no according to https://github.com/NVIDIA/Fuser/blob/530c62db573efdc688e91e7f1dac506bc2c3b4a3/csrc/python_frontend/python_bindings.cpp#L1482; otherwise, we are computing wrong for all unary ops.

wujingyue: Done -- PTAAL

wujingyue:
!build

csrc/python_frontend/python_bindings.cpp
jjsjann123: ah, sorry I got confused about the define tensor part. Yeah we don't need to worry about the reduction axes, since we are not dealing with `TensorView`, but `Tensor` directly.
Thanks for fixing the 1d case. :bow:

PR #3237, opened 2024-10-19
Add cat operation to fusion translation

rdspring1:
This PR adds support for cat in FusionTranslation.

Details:
* `cat` operation applies `pad`. During translation, the `pad` expression is handled separately. This PR adds a short-circuit path to `cat` that is enabled via the `manual_padding` argument.


rdspring1:
!build

rdspring1:
!build

tests/python/utils.py
```diff
@@ -406,7 +404,7 @@ def exec_nvfuser(
         *,
         new_fusion_expected=True,
         device=None,
-        is_clonable=False,
+        is_clonable=True,
```
jjsjann123, line 407: :clap:

csrc/ops/alias.cpp
```diff
@@ -629,6 +629,21 @@ TensorView* cat(
     return set(inputs.at(0));
   }
 
+  // short-circuit: If all inputs are already padded, assume correctness and cat
+  // them without padding. FusionTranslation adds the padOp for each tensor
+  // separately.
+  // TODO add correctness check against desired padding.
+  bool all_padded =
```
jjsjann123, line 636: > assume correctness and cat
  // them without padding. FusionTranslation adds the padOp for each tensor
  // separately.
  // TODO add correctness check against desired padding.

I think we needed to have the correctness checking here.
How are we going to identify cases where we explicitly pad a tensor before concat?

rdspring1:
!build

csrc/ops/alias.cpp
rdspring1: I'm not sure how to check if two values are the same at the fusion definition without expression evaluator.

My only idea is to add a boolean flag for the translation, so this short circuit only applies then.

rdspring1: I decided to detect if a `CatOp` follows a `PadOp` during translation. In this case, skip the `PadOp` so it is like the original script. Now there is no changes to `csrc/ops/alias.cpp`.

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

tests/python/utils.py
rdspring1: I reverted this. I still need to add `Matmul`, `SpdaFwd`, and `SpdaBwd`

rdspring1:
!build

tests/python/utils.py
rdspring1: See https://github.com/NVIDIA/Fuser/pull/3240

csrc/ops/alias.cpp
jjsjann123: sorry this falls off my radar. I'll take another look shortly

rdspring1:
@jjsjann123 I added a boolean flag argument called `manual_padding` to the cat composite op. If `manual_padding == true`, we skip the pad step in the cat op. It is only enabled when the `CatOp` expression is translated from cpp to python.

**Why?** 
1. I couldn't find a better correctness check.
2. If you skip the `PadOp`, you can get unused intermediates in the translated `FusionDefinition`. It seems cleaner to have the translation match the cpp fusion exactly.

rdspring1:
> What specifically are we thinking could be incorrect? 

Nothing in particular but it would be nice to check that the pad on the inputs matches the pad the CatOp was expected to generate.

> The unused intermediates would be the scalars that appear as the PadOps attributes right? If needed, this could probably be avoided by using a simple custom stack-based traversal instead of StmtSort.

Yes. I wanted to avoid writing a special case for this.

csrc/ops/alias.cpp
```diff
@@ -629,6 +630,23 @@ TensorView* cat(
...
+  // them without padding. FusionTranslation adds the padOp for each tensor
+  // separately.
+  bool all_padded =
+      std::all_of(inputs.begin(), inputs.end(), [](TensorView* tv) {
+        return tv->definition()->isA<PadOp>();
+      });
+  NVF_ERROR(
+      !manual_padding || all_padded,
+      "Expected all inputs to be padded when manual_padding is True.");
+  if (all_padded) {
```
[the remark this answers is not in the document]
jjsjann123: nitpick: we should move the above check inside `if (manual_padding)` as well

jjsjann123 (commented):
Can we add a test in `test_python_frontend.py` where we explicitly pad a bunch of inputs before cat them together?

jjsjann123:
> What specifically are we thinking could be incorrect? The CatOp inputs are in the correct order, so if we just strip off the PadOps we would have the right order of arguments.

we are making changes on our `cat` api directly.
I think if we have a fusion definition, where all inputs are explicitly padded for something unrelated to the cat, without a safety check, we'll incorrectly skip the pad needed for cat. Hence the ask about a safety check on whether or not it's safe to skip the pad.

But a `manual_padding` flag works as well, so I'm happy with this.

csrc/serde/fusion_cache.fbs
```diff
@@ -255,6 +256,12 @@ table BroadcastInDim {
   broadcast_dims: [long];
 }
 
+// Data for CatOpRecord
+table Cat {
+  dim: long;
```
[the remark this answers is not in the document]
rdspring1: No, It reuse a different table. Tracking a single dimension is common across RecordFunctors.

naoyam:
How to know if a certain subset of ops corresponds to a cat?

rdspring1:
!build

rdspring1:
@jjsjann123 I added `test_pad_prior_cat` to test padding tensors before cat operation.

tests/python/test_python_frontend.py
```diff
@@ -1906,12 +1906,36 @@ def fusion_func(fd: FusionDefinition):
...
         self.assertEqual(torch.cat([inputs[0], inputs[2]], dim=0), nvf_out[1])
         # self.assertEqual(torch.cat([inputs[0], inputs[3]], dim=0), nvf_out[2])
 
+    def test_pad_prior_cat(self):
+        inputs = [
+            torch.randn(2, 4, device="cuda"),
+            torch.randn(3, 3, device="cuda"),
+        ]
+
+        def fusion_func(fd: FusionDefinition):
```
jjsjann123, line 1921: :clap:

jjsjann123 (approved):
My issues are addressed. :ship:

csrc/python_frontend/python_bindings.cpp
```diff
@@ -2665,7 +2665,8 @@ void initNvFuserPythonBindings(PyObject* module) {
       "cat",
       [](FusionDefinition::Operators& self,
          std::vector<Tensor> tensors,
-         int64_t dim) -> Tensor {
+         int64_t dim,
+         bool manual_padding) -> Tensor {
```
jjsjann123, line 2669: Do we need this? Shouldn't python API to always have the vanilla `cat` that lowers to `pad + cat`?

Follow up with Naoya's question
> How to know if a certain subset of ops corresponds to a cat?

I think the case we are making here is that, we assume that `cat` is always lowered to `PadOp + CatOp`. So any `PadOp` followed by `CatOp` is considered as `corresponds to a cat`.

We just needed a different API to skip adding the `PadOp`, since when we do cpp->python translation, we've already explicitly added the PadOp in the fusion.

rdspring1: The translated python definition will have a CatOpRecord that needs print a python string with `manual_padding` set to true.

csrc/python_frontend/python_bindings.cpp
```diff
@@ -2678,11 +2679,15 @@ void initNvFuserPythonBindings(PyObject* module) {
           tensor_states.push_back(fd->recordingState(t()));
         }
         self.fusion_definition->defineRecord(new CatOpRecord(
-            tensor_states, {fd->recordingState(output())}, dim));
+            tensor_states,
+            {fd->recordingState(output())},
+            dim,
+            manual_padding));
```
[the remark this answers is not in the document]
rdspring1: You need it to print the translated python definition, right?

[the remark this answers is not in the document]
jjsjann123: Sorry I wasn't being clear here.

Yeah we need to keep manual_padding in CatOpRecord, yes. As well as inside `alias.cpp`. But *not* in python binding. we can just pass `true` here and we don't need to have manual_padding at line 2669.

The only time where we would want to pass a `manual_padding=true` is inside the translation function, and we already do that. I don't see why we want to expose it at python API.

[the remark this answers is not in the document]
rdspring1:
```python
cloned_fd = FusionDefinition()
clone(fd, cloned_fd)
print(cloned_fd)
# ^^^ The print cloned_fd definition should have `manual_padding=true`. The python API should reflect what the RecordFunctor is doing.
```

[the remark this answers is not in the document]
jjsjann123: ah, good catch. I forgot about that part. :face_with_head_bandage: 
Sorry about the distraction. :bow:

csrc/ops/alias.cpp
```diff
@@ -629,6 +630,24 @@ TensorView* cat(
     return set(inputs.at(0));
   }
 
+  // short-circuit: If manual_padding is true, check that all inputs are already
+  // padded. Assume the padding is correct and create the catOp immediately.
+  // Primarily used for FusionTranslation, which adds the padOp for each tensor
+  // separately.
```
jjsjann123, line 636: Thanks for the comment as well. :+1:

jacobhinkle:
> How to know if a certain subset of ops corresponds to a cat?

Yeah I guess it is a bit of an awkward pattern to match:
```
d = cat([a, b, c, d], dim=0)

   a       b      c        d
   |       |      |        |
 PadOp   PadOp   PadOp   PadOp
   |       |      |        |
   |       |      |       /
   |       |       OR
   |       |      |
   |       |     /
   |        OR
   |       |
   |      /
    OR
   |
   d
```
The pad values must all be zero and then the unpadded regions must partition dimension 0. Probably proving the partitioning part is the hardest condition to check.

PR #3238, opened 2024-10-20
Fix torch allocator clearing cache on every benchmark

jacobhinkle:
This speeds up running the benchmarks by quite a bit since malloc is so slow.

### DIFF
--- a/nvfuser/pytorch_utils.py
+++ b/nvfuser/pytorch_utils.py
@@ -173,9 +173,6 @@ def clear_cuda_cache() -> None:
     """
     Utility function to clear CUDA cache before running a test.
     """
-    if (
-        torch.cuda.memory_allocated()
-        or torch.cuda.memory_reserved() > 0.8 * DEVICE_PROPERTIES["gpu_gmem_bytes"]
-    ):
+    if torch.cuda.memory_reserved() > 0.8 * DEVICE_PROPERTIES["gpu_gmem_bytes"]:
         gc.collect()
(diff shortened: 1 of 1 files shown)


jacobhinkle:
!build

jacobhinkle:
!build --matmul-bench

nvfuser/pytorch_utils.py
```diff
@@ -173,9 +173,6 @@ def clear_cuda_cache() -> None:
     """
     Utility function to clear CUDA cache before running a test.
     """
-    if (
-        torch.cuda.memory_allocated()
```
Priya2698, line 177: Can we safely remove this? 
Shouldn't we always clear the L2 cache if there is memory allocated at the beginning to benchmark each round from the same memory state?

I was under the impression that this just queries the device statistics, how does this use malloc? Can you point me to any reference on its working?

jacobhinkle: As mentioned [here](https://pytorch.org/docs/stable/notes/cuda.html#cuda-memory-management), "allocated" refers to only the memory occupied by tensors currently, but deleted tensors' memory may still remain allocated to the pytorch cache. That unused but still allocated memory is included along with the allocated memory in the "reserved" amount. I don't think we care what is allocated, and in particular having this here means that we will clear the cache _every_ time this is called if there is even a single tensor still in memory somewhere i.e. not yet garbage collected.

jacobhinkle:
Closing in favor of #3252

PR #3240, opened 2024-10-20
Add support for Matmul and SDPA to fusion translation

rdspring1:
This PR add translation for `MatmulOp`, `SdpaFwdOp`, and `SdpaBwdOp` to fusion translation. It also enables `is_clonable` by default.


rdspring1:
!build

tests/python/test_matmul.py
```diff
@@ -201,3 +201,4 @@ def fusion_func(fd: FusionDefinition) -> None:
         ]
         outputs, _ = self.exec_nvfuser(fusion_func, inputs)
         assert outputs[0].ndim == 3
+        self.exec_nvfuser(fusion_func, inputs)
```
jjsjann123: what is this trying to do?

tests/python/utils.py
```diff
@@ -404,7 +404,7 @@ def exec_nvfuser(
         *,
         new_fusion_expected=True,
         device=None,
-        is_clonable=False,
+        is_clonable=True,
```
jjsjann123, line 407: :clap:

tests/python/test_matmul.py
Priya2698, line 204: Line 202 already calls this.

Priya2698:
Thanks for adding these translations for the new operators, I wasn't aware of it.
What are these used for?

rdspring1:
!build

tests/python/utils.py
Priya2698, line 407: What is the motivation for switching this default value -- is it just so we run the cloned fusion check for all tests instead of a few selected ones?

rdspring1: Yes, all the tests are clonable after this PR.

rdspring1:
There isn't a way to manipulate the `FusionSegments` created by the segmentation algorithm in python.
My idea is to run segmentation, translate the segments back to the python frontend, and to compose a complex `FusionDefinition` as a sequence of segments.

jjsjann123 (approved):
:shipit:

Priya2698:
> There isn't a way to manipulate the `FusionSegments` created by the segmentation algorithm in python. My idea is to run segmentation, translate the segments back to the python frontend, and to compose a complex `FusionDefinition` as a sequence of segments.

Got it, thanks.
Do you intend on adding tests to exercise these translations?

tests/python/test_matmul.py
rdspring1: This feels like leftover from a bad rebase.

rdspring1:
> Do you intend on adding tests to exercise these translations?

The baseline test is to translate the original CPP fusion back to a python definition. The results between the original and clone should match.

Priya2698 (approved):
LGTM

rdspring1:
There is also the use case of translating all of the fusions in the cpp tests to the python frontend.

PR #3242, opened 2024-10-21
Fix a multi-GPU-related bug in remove_bcast_squeeze. 

wujingyue:


wujingyue:
!build

csrc/preseg_passes/remove_bcast_squeeze.cpp
```diff
@@ -362,18 +362,46 @@ TensorView* maybeDoReplacement(TensorView* orig) {
...
-      // In particular, we might have a Device dimension parallelized for the
-      // output, which we need to preserve.
-      new_loop[i]->parallelize(old_loop[i]->getParallelType());
+      needs_resharding = true;
+      break;
     }
   }
-  return replacement;
+
+  if (needs_resharding) {
```
wujingyue, line 398: I'm also fine with simply bailing out and thus keeping the squeeze-broadcast pair. I send this version because it's slightly more powerful.

jacobhinkle: This makes sense to me. If we bail then we would leave the squeeze+bcast sequence shown in your comment above, which could lead to some of the errors we have seen previously, so I prefer to leave this as you've written it here.

csrc/preseg_passes/remove_bcast_squeeze.cpp
```diff
@@ -362,18 +362,46 @@ TensorView* maybeDoReplacement(TensorView* orig) {
...
       " has different dimension than original ",
       orig->toString());
+  // Replacing `orig` with `replacement` can introduce resharding on the
+  // consumer of `orig`, e.g., the `cast` in the following example.
+  //
+  // [bDID{1}, i0]: replacement
+  //      |
+  //      | squeeze
+  //      v
+  //     [i0]
```
jacobhinkle, line 372: Just so I understand correctly, this actually is a resharding op, but it is not skipped here because we don't check SqueezeOp or BroadcastOp for resharding. Is that right?

jacobhinkle (approved):
Thank you for pushing this I had not considered DID on broadcast axes in the last of these PRs.

wujingyue:
!build

wujingyue:
!build

csrc/preseg_passes/remove_bcast_squeeze.cpp
wujingyue: > this actually is a resharding op

It's arguable. I tend to think it's not resharding because, assuming same mesh, broadcast dimensions are size-1 and they are replicated regardless of whether DID or not. 

Note: remove_bcast_squeeze correctly checks for isResharding when picking candidates: https://github.com/NVIDIA/Fuser/blob/63cc45c47f1e96946cb952e3923bd076867a8015/csrc/preseg_passes/remove_bcast_squeeze.cpp#L85-L87. But, for the reason that I said, https://github.com/NVIDIA/Fuser/blob/63cc45c47f1e96946cb952e3923bd076867a8015/csrc/multidevice/utils.cpp#L162-L165 doesn't treat as resharding a removal or a creation of a broadcast dimension.

PR #3243, opened 2024-10-21
Create HopperMultiMatmulScheduler

rdspring1:
This PR creates `HopperMultiMatmulScheduler`.

## Changes:
* `scheduleMultipleMatmuls` in `multi_matmul.h` selects between `AmpereMultiMatmulScheduler` and `HopperMultiMatmulScheduler` based on device compute capability.
* Rename `MultiMatmulScheduler` to `AmpereMatmulScheduler`. Move it from `multi_matmul.cpp` to `ampere_multi_matmul.h`
* Create `HopperMultiMatmulScheduler` in `hopper_multi_matmul.h`
* Update arch guard in `test_matmul_scheduler.cpp`

## Why?
Pros:
- For deprecation, we could remove lines in `MatmulScheduler` and delete old files.
- Assume uniform compute capability in each `MultipleMatmulScheduler`.
- All reusable code would go matmul_utils.h and mma_util.h.

Cons:
- Older generations would have fewer capabilities, since most development would go to the current generation.
- More lines of code.
- `MatmulParams` could contain parameters for different generations.


rdspring1:
!build

rdspring1:
!build

csrc/scheduler/hopper_multi_matmul.cpp
```diff
@@ -0,0 +1,1524 @@
...
+  IterDomain* id = representativeId(abs_id);
+  NVF_ERROR(
+      !id->isBroadcast() && !id->isReduction(),
+      "no support for reduction or broadcast domains, but got ",
+      id->toString());
+  NVF_ERROR(
+      id->extent()->isConstInt(),
+      "swizzled dimension's extend must be known during scheduling, got ",
+      id->toString());
+}
```
jacobhinkle, line 58: If this is not needed for TMA swizzle, we can remove it. Otherwise it can go to `mma_utils.cpp`

csrc/scheduler/hopper_multi_matmul.cpp
```diff
@@ -0,0 +1,1524 @@
...
+//! Automatically generates the shared memory swizzled data layout
+//!  for matmul mainloop and epilogue.
+//! The shared mem data layout is always 2D currently, and this utility
+//!  function assumes that the shared_mem_tv has the following structure:
+//!  [tile_row, tile_col]
+//! Returns the domain with swizzle. For the case of legacy swizzle, this
+//! domain must be set as loop domain. For the case of new swizzle, this domain
+//! must be set as allocation domain.
+template <bool legacy = true>
+AbstractTensor swizzleSharedMemory(TensorView* shared_mem_tv) {
```
jacobhinkle, line 69: Maybe we should put a `// TODO: Remove this` note here since we will probably not need it or it will change enough to deserve a new name once we implement TMA scheduling

csrc/scheduler/hopper_multi_matmul.cpp
```diff
@@ -0,0 +1,1524 @@
...
+
+  // Cache and fork outputs
+  cached_outputs_ =
+      scheduler_utils::cacheAndForkOutputs(fusion_, /*unroll=*/true);
+}
+
+void HopperMultipleMatmulScheduler::findPatterns() {
+  patterns_ = mma_utils::findMatmulPatterns(fusion_);
+  NVF_ERROR(!patterns_.empty(), "No matmul patterns were found");
+}
```
jacobhinkle, line 528: Stuff like this should probably just go into a base class which we subclass for each new arch. What do you think?

jacobhinkle (commented):
I like this. I think there is probably enough in common to justify having a base class that is parent to both the Ampere and Hopper classes, where we can hold common methods in `multi_matmul.cpp`.

csrc/scheduler/hopper_multi_matmul.cpp
rdspring1: I think the general swizzle logic can go in `mma_utils.cpp` because we need swizzle support for regular smem epilogue.

csrc/scheduler/hopper_multi_matmul.cpp
rdspring1: I was using some of the logic to determine if swizzle is necessary and which swizzle to apply with TMA.

csrc/scheduler/hopper_multi_matmul.cpp
rdspring1: Make sense. Should we make this refactor and consolidation in a separate PR so the changes are more clear.

rdspring1:
!build

rdspring1:
I created https://github.com/NVIDIA/Fuser/issues/3266 to hold all refactoring changes to the `MultiMatmulScheduler`.

jacobhinkle (approved):
Approving in light of #3266

PR #3247, opened 2024-10-21
Update print message for cycle detection.

rdspring1:
When a cycle occurs in the fusion, a segmentation fault can occur when printing the statements in the cycle. In this case, the developer doesn't get any information and must use `gdb` to diagnose the segmentation fault. This PR adds a warning message to print that a cycle is detected before printing the cycle's statements.


rdspring1:
!build

jjsjann123 (approved):

PR #3248, opened 2024-10-21
Use float instead of double

naoyam:
This doesn't fix anything, but for low precision types, use float instead of double to represent immediate
values in generated CUDA code.

This change does make some difference in PTX, although I haven't checked
the final SASS.

For example, when initializing an `__e4m3` value with 0, the generated
code looks like:

```
__e4m3 x = 0.0;
```

Or for a pad op:

```
__e4m3 x = (cond) ? y[...] : 0.0;
```

This PR changes the generated code to be:

```
__e4m3 x = 0.0f;
```

and

```
__e4m3 x = (cond) ? y[...] : 0.0f;
```

The corresponding PTX has a 64-bit register used for initializing `x`
with 0 without the `f` suffix, whereas only a 32-bit register is used
with `f`. The difference may not matter in the final SASS as it may be
optimized away, but I think this makes more sense.


naoyam:
!build

jacobhinkle (approved):
LGTM

PR #3255, opened 2024-10-22
Gather data for pointwise fusions

rdspring1:
This PR creates a script to collect data for pointwise fusions.

## Why?
* Previous approaches were limited to training a model on the performance metrics of a single fusion.
* Training more complex models will require more data. It is convenient to separate the data collection and training steps.

## Overview
1. Run through a shmoo of pointwise fusion configurations.
2. Apply different vectorization and unroll factor settings.
3. Profile several metrics for each fusion and scheduler combination.
4. Create Pandas DataFrame
5. Save results to csv file.

## What Fusions?
You can view pointwise fusions in terms of vertical and horizontal components. The vertical component is the number and type of operations in the fusion. The horizontal component is the number, size and type of dimensions for the tensors.

### Vertical Components
1. Number of operations
2. Number of MUFU operations
3. Ordering of MUFU operations

### Horizontal Components
1. Number of broadcast dimensions
2. Ordering of broadcast dimensions

### Input Metrics
1. Shapes of Input Tensors
2. Shapes of Output Tensors
3. Number of Operations
4. Ordering of MUFU operations
5. Broadcast Multiples
6. Vectorization Factor
7. Unroll Factor

### Output Metrics
1. Number of Registers
2. Shared memory usage - static and dynamic
3. Grid shape
4. Block shape
5. Effective Bandwidth (GB/s)
6. Kernel Runtime (ms)


doc/dev/python_scheduling/collect_pointwise_data.py
```diff
@@ -0,0 +1,253 @@
...
+                output = fd.from_pytorch(input_tensors[input_tensor_idx])
+                input_tensor_idx += 1
+
+            if idx in mufu_indices:
+                output = fd.ops.exp(output)
+            else:
+                other = fd.from_pytorch(input_tensors[input_tensor_idx])
+                input_tensor_idx += 1
+                output = fd.ops.add(output, other)
+        fd.add_output(output)
```
jjsjann123, line 133: should we explore more topology rather than always chained operations?

rdspring1: Good idea. At the moment, I was working backwards from the information available to the Pointwise scheduler. It doesn't compute any metrics about the fusion's DAG.

rdspring1: Is a more complex topology relevant for pointwise scheduler? e.g., The DAG is flattened to a linear sequence of operations, so it does not affect the arithmetic intensity of the Fusion.

rdspring1:
!build

liqiangxl:
Do you have some demo results using this script?

doc/dev/python_scheduling/collect_pointwise_data.py
```diff
@@ -0,0 +1,378 @@
...
+# All rights reserved.
+# SPDX-License-Identifier: BSD-3-Clause
+# Owner(s): ["module: nvfuser"]
+
+import torch
+import itertools
+import numpy as np
+from nvfuser import FusionCache, FusionDefinition, SchedulerType
+
+# ============================ Description ============================
```
Priya2698, line 11: Can we also add a comment on how to run this script, and the output it will generate?
Have you used this script to generate data for autotuning?

PR #3258, opened 2024-10-23
Add manual_normalization to slice operation.

rdspring1:
The PR adds the `manual_normalization` boolean flag to the slice composite operation.

Why?
* During translation, the `normalize_slice_range` lambda is applied to the Slice arguments redundantly.
* In the worst case, the redundant normalization step can cause a segmentation fault.
* This ^^^ happened in https://github.com/NVIDIA/Fuser/pull/3025.


rdspring1:
!build

csrc/ops/alias.cpp
```diff
@@ -744,7 +747,8 @@ TensorView* slice(TensorView* inp, const std::vector<Slice>& ranges) {
       ", Expected: ",
       ndims);
 
-  const auto normalize_slice_range = [](Slice range, Val* extent) -> Slice {
+  const auto normalize_slice_range =
+      [](Slice range, Val* extent, bool manual_normalization) -> Slice {
```
jacobhinkle, line 751: Can you just capture `manual_normalization` instead of passing it explicitly?

csrc/python_frontend/translation.cpp
```diff
@@ -29,6 +29,27 @@ namespace {
 // _C._FusionDefinition class created by pybind11. It is easier to operate on
 // the child class directly than to create a new child instance from parent
 // instance.
+//
+// How to add support for an expression not yet overriden by FusionTranslator?
```
jacobhinkle, line 33: Thanks for adding this

jacobhinkle (approved):
It's a shame that we need to do these sorts of tricks. I understand that the redundant normalization means we won't build exactly the same IR as we originally had, but I would still not expect that to cause any errors. Do you have a repro for the segfault you encountered?

rdspring1:
The segmentation fault occurred in `test_reshape_squeeze_concretization`. It is caused by a stack overflow.

### Why?
- It is a specific combination with slice and view.
- The slice operation normalizes the range, which becomes its output TensorView's extents.
- The view operation manipulates the slice's TensorDomain to create a new TensorView.

Now, add segmentation, which clones the fusion to create sub-segments. This compounds the redundant normalization computation. I didn't figure out how things went from mildly annoying to infinite stack overflow. There were only like 2-3 extra cloning steps.

csrc/ops/alias.cpp
rdspring1: Sure. I tend to capture when it become inconvenient to pass something as an argument.

rdspring1:
> It's a shame that we need to do these sorts of tricks. 

It'd be nice to prove that two DAGs are functionally equivalent.
- Compare sequence of operations
- For full fusions, you can compare IdGraphs for equivalence.

csrc/ops/alias.h
```diff
@@ -126,7 +126,10 @@ NVF_API TensorView* cat(
 //! ranges parameter. Stepping must be one at this moment. The semantics of
 //! slicing with negative values and values >= extent follow those of numpy and
 //! PyTorch.
-NVF_API TensorView* slice(TensorView* inp, const std::vector<Slice>& ranges);
+NVF_API TensorView* slice(
+    TensorView* inp,
+    const std::vector<Slice>& ranges,
+    bool manual_normalization = false);
```
jjsjann123, line 132: nitpick: would appreciate a comment on what `manaul_normalization` does.

csrc/serde/fusion_cache.fbs
```diff
@@ -303,9 +303,7 @@ table Size {
 
 // Data for SliceOpRecord
 table Slice {
-  start_indices: [long];
-  end_indices:[long];
-  strides: [long];
+  manual_normalization: bool;
```
jjsjann123, line 306: oops. thanks for cleaning up this one.

jjsjann123 (approved):
I'm also quite curious on how we are getting an infinite stack overflow.

rdspring1:
!build

PR #3053, opened 2024-09-29
When computing isResharding, ignore parallel types for broadcasts. 

wujingyue:
This is a follow-up to #2877. The fix wasn't complete, causing some intra-device set to be mistaken as a communication. 

This is required to land #3045.

PR #3259, opened 2024-10-23
Sequence Parallel MLP

cowanmeg:
Sequence parallel MLP test

Kernel dump:
```
void at::native::vectorized_elementwise_kernel<(int)4, at::native::FillFunctor<c10::BFloat16>, at::detail::Array<char *, (int)1>>(int, T2, T3)
ncclDevKernel_AllGather_RING_LL(ncclDevKernelArgsStorage<(unsigned long)4096>)
[Linear0] ampere_bf16_s16816gemm_bf16_64x64_sliced1x2_ldg8_relu_f2f_stages_64x6_tn
void at::native::vectorized_elementwise_kernel<(int)4, at::native::FillFunctor<c10::BFloat16>, at::detail::Array<char *, (int)1>>(int, T2, T3)
[GeLU] <unnamed>::nvfuser_pointwise_f0_c1_r0_g7(<unnamed>::Tensor<<unnamed>::__bfloat, (int)3, (int)3>, <unnamed>::Tensor<<unnamed>::__bfloat, (int)3, (int)3>)
[Linear1-matmul]ampere_bf16_s16816gemm_bf16_64x64_sliced1x2_ldg8_f2f_stages_64x6_tn
void at::native::vectorized_elementwise_kernel<(int)4, at::native::FillFunctor<c10::BFloat16>, at::detail::Array<char *, (int)1>>(int, T2, T3)
[Linear1-reduction]ncclDevKernel_ReduceScatter_Sum_bf16_RING_LL(ncclDevKernelArgsStorage<(unsigned long)4096>)
void at::native::vectorized_elementwise_kernel<(int)4, at::native::FillFunctor<float>, at::detail::Array<char *, (int)1>>(int, T2, T3)
void at::native::vectorized_elementwise_kernel<(int)4, at::native::FillFunctor<float>, at::detail::Array<char *, (int)1>>(int, T2, T3)
[Linear1-bias add, dropout]<unnamed>::nvfuser_pointwise_f0_c1_r0_g8(<unnamed>::Tensor<<unnamed>::__bfloat, (int)1, (int)1>, <unnamed>::Tensor<<unnamed>::__bfloat, (int)3, (int)3>, long long *, long long, long long *, long long, <unnamed>::Tensor<float, (int)3, (int)3>, <unnamed>::Tensor<float, (int)3, (int)3>)
```


csrc/multidevice/utils.cpp
```diff
@@ -165,6 +165,13 @@ bool haveDifferentShardings(
       continue;
     }
 
+    if (p_id->isBroadcast()) {
```
wujingyue, line 168: Did you `!build`? I attempted to make a similar change months aog and found it broke some other unit tests.

cowanmeg:
!build

csrc/multidevice/utils.cpp
wujingyue: https://github.com/NVIDIA/Fuser/pull/3053 is the PR I was talking about. Unfortunately, it's too old to keep the test history; otherwise, I could figure out which test(s) failed.

cowanmeg: Hmm it is failing the single device multidevice tests... unfortunately there's two issues (1) size 1 axes are treated as broadcast regardless if it's the fake DID axis or an actual broadcast. (2) Reshapes that insert 'D' axes intended to be sharded and/or reduced are optimized away when D = 1.

cowanmeg: (1) isn't an issue for this PR, but it does come up with a sharded normalization op. (2) is unrelated to this particular line

cowanmeg:
!build

csrc/multidevice/utils.cpp
cowanmeg: Removed this. Note, that (2) is still an issue in this test where an extra logical device axis is reshaped for a ReduceScatter that gets optimized away when D=1. 
Since this is an artifact of DID parallelizations only on the logical domain, I decided to skip this test when D=1 and left a TODO to revisit when DID loop parallelization is finished.

wujingyue: > Reshapes that insert 'D' axes intended to be sharded and/or reduced are optimized away when D = 1.

Can you clarify the problem? IIUC, such a reshape will be converted into a broadcast. Similarly, a reshape that deletes `D` will be converted into a squeeze when `D=1`. In either case, the reshape isn't treated as resharding.

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -278,29 +278,65 @@ std::vector<TensorView*> mlp(
...
+    TensorView* mask = nullptr) {
   const DataType dtype = w0->dtype();
-  // // Linear 0
+
+  if (sequence_parallel) {
+    // Input arrives sharded and must be allgathered back
+    x->setDeviceMesh(mesh);
+    x->axis(0)->parallelize(ParallelType::DIDx);
+    x = set(x); // allgather
+    x->setDeviceMesh(mesh);
```
wujingyue, line 291: This line isn't needed. `set` propagates the input mesh.

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -278,29 +278,65 @@ std::vector<TensorView*> mlp(
...
+    local_matmul1 = reshape(local_matmul1, {D, B * S, E}, {D, D, B * S / D, E});
+  }
+  TensorView* matmul1 = sum(local_matmul1, {0}); // Allreduce or Reduce scatter
+  TensorView* linear1 = add(matmul1, broadcast(b1, {true, true, false}));
   // Dropout
   Val* prob = IrBuilder::create<Val>(1.0 - kDropoutProb);
   Val* scale = IrBuilder::create<Val>(1.0 / (1.0 - kDropoutProb));
-  auto dropout_result = dropout(linear1, prob, scale).output;
+  if (mask == nullptr) {
+    auto rand_vals = rand_like(linear1);
```
wujingyue, line 316: (not required for this PR) 

Should we also feed seed and offset? rand_like can already take that. It's in general better to expose rng outside so it's easier to make things deterministic yet random. I think this is helpful for implementing sharded dropout where each rank has to be randomized differently.

csrc/multidevice/utils.cpp
cowanmeg: The output of the local matmul (partially reduced):`[DIDx(D), i0, i1]`. In preparation for reduce scatter where i0 becomes the new sharded axis we have to reshape into `[DIDx(D), D, i0/D, i1]`. After this reshape the non-sharded `D` was optimized away. I tried turning off preseg passes remove bcast squeeze but a different error occurred.

tests/cpp/test_multidevice_transformer.cpp
cowanmeg: Yes, that would be better. I did attempt setting the offset and seed in the nvfuser fusion, but I couldn't figure out exactly how to set the parameters to match ATen's dropout. 

I assumed that ATen's dropout would advance the offset by the size of the input, but it only advanceded the offset by 1 for each call to dropout. 
Some pytorch code showing this
```
>>> x = torch.rand([5])
tensor([0.4963, 0.7682, 0.0885, 0.1320, 0.3074])
>>> torch.nn.functional.dropout(x)
tensor([0.0000, 1.5364, 0.1770, 0.2641, 0.6148])
>>> torch.rand([5])
tensor([0.4901, 0.8964, 0.4556, 0.6323, 0.3489])

>>> torch.manual_seed(0)
<torch._C.Generator object at 0x7f819c426830>
>>> torch.rand([10])
tensor([0.4963, 0.7682, 0.0885, 0.1320, 0.3074, 0.6341, 0.4901, 0.8964, 0.4556,
        0.6323])
The dropout only caused 0.6341 to be skipped
  ```

File: tests/cpp/test_multidevice_transformer.cpp [hunk 4]
```diff
@@ -278,29 +278,65 @@ std::vector<TensorView*> mlp(
...
-  // // Linear 0
+
+  if (sequence_parallel) {
+    // Input arrives sharded and must be allgathered back
+    x->setDeviceMesh(mesh);
+    x->axis(0)->parallelize(ParallelType::DIDx);
+    x = set(x); // allgather
+    x->setDeviceMesh(mesh);
+    x->axis(0)->parallelize(ParallelType::Serial);
+    // Reshape back to 2D for linearOp
```
samnordmann, line 293: is there a plan to patch LinearOp to accept nD tensors ?

samnordmann (approved):
Looks great! thannks

tests/cpp/test_multidevice_transformer.cpp
cowanmeg: Good question, the ATen interface can accept an arbitrary number of input dimensions for activations. @Priya2698 would know better. 

I might be mixing up the restriction with weights and activations...

tests/cpp/test_multidevice_transformer.cpp
wujingyue: Got it! I saw Thunder solve this by using a lower-level op: https://github.com/Lightning-AI/lightning-thunder/blob/1575797a8d21e0b845e188ddd6af305be6457b1c/thunder/executors/torchex.py#L187. Certainly not ideal because have to manually decompose. But worth noting when we really need this.

tests/cpp/test_multidevice_transformer.cpp
wujingyue: LinearOp supports nD input activations. >2D weights however will be squeezed and results will be unsqueezed -- preceding dimensions are treated as device dimensions. According to the code [here](https://github.com/NVIDIA/Fuser/blob/628a47e33ac418dc7b2a3a79a01f52174c2fe325/csrc/ir/nodes.cpp#L4379-L4394), I think it should work out of the box without "reshaping back to 2D".

wujingyue (commented):
LG in general. I'm taking a final round to review the model implementation.

tests/cpp/test_multidevice_transformer.cpp
Priya2698: Yes. Activations can have any number of dimensions. Weights can be 1D or 2D.
@wujingyue removed 1D weights for ease of DID implementation of linear.

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -669,6 +704,93 @@ TEST_P(DistributedTransformerTest, MLP_Layer) {
...
+    GTEST_SKIP() << "Requires number of devices=" << D
+                 << " evenly divide B*S=" << B * S;
+  }
+  DataType dtype = GetParam();
+  at::ScalarType at_dtype = data_type_to_aten(dtype);
+  auto fusion = std::make_unique<Fusion>();
+  FusionGuard fg(fusion.get());
+  const auto mesh = DeviceMesh::createForNumDevices(D);
+
+  TensorView* x = makeContigConcreteTensor({D, B * S / D, E}, dtype);
```
wujingyue, line 728: (Not required for this PR)

Did you try `{D, B, S/D, E}`, which is more readable and closer to how the model is built conceptually?

wujingyue (approved):

tests/cpp/test_multidevice_transformer.cpp
cowanmeg: I did try this, but since the sequence dimension isn't the outer-most dimension it breaks the analysis in the pointwise scheduler that assumes DID axes are outermost axes. I didn't bother modifying that scheduler since it will have to be fixed for loop + DID parallelization.

cowanmeg:
!build

PR #3260, opened 2024-10-23
[codediff] Handle timestamps in pytest verbose mode

jacobhinkle:
Our improper handling of pytest log output led to failures in codediff CI jobs in the past day or so. I removed the splitting-based parser code and improved the regex approach. I tested with the artifact from a recent failing job and it worked.

This also fixes a tricky bug where we were observing different numbers of kernels in tests that did not have any code changes that would cause such behavior. The problem was that the FusionCache was serialized to `/tmp/nvfuser_kernel_db` on the first run of the script then it was deserialized on the next run. This means the second run would show fewer new kernels being compiled since there would be some cache hits. Now I make sure to clear `/tmp/nvfuser_kernel_db` inside `run_command.sh` and reinstate it after.

### DIFF
--- a/tools/codediff/diff_report.py
+++ b/tools/codediff/diff_report.py
@@ -391,7 +391,14 @@ class LogParserPyTest(LogParser):
 
         self.itemlist_re = re.compile(r"Running \d+ items in this shard: (.*)$")
 
-        self.wildcard_testname_re = re.compile(r"^(?P<testname>\S+::\S+) (?P<line>.*)$")
+        # match lines like these:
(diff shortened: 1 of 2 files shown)


jacobhinkle:
!build --diff

jacobhinkle:
!build --diff-bench

xwang233 (approved):

naoyam:
!build --diff

naoyam:
!build --diff

jacobhinkle:
!build --diff

jacobhinkle:
!build --diff

PR #3261, opened 2024-10-23
Lowering vectorized pad

jjsjann123:
Added support for lowering TernaryOp:where with vectorization factor.

i.e.
```
predicate
  ? loadGlobalToLocal<...>(&dst[0], &src[i_src])
  : dst.set(0.0f) 
```

Currently this can only be done via manual scheduling. The follow up PR on vectorization analysis will make this automatically applied in PR #3321


jjsjann123:
!build

jjsjann123:
!build

csrc/codegen.cpp
```diff
@@ -1001,6 +1001,94 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
   }
 
   void handle(const TernaryOp* top) final {
+    // Get vectorization information
```
jjsjann123, line 1004: Note for myself. merge this with LoadStoreOp

csrc/device_lower/pass/index.cpp
```diff
@@ -2306,9 +2306,22 @@ void IndexLowering::handle(const PadOp* pad) {
 
   pred = GpuLower::current()->commonScalarMap().hoistScalar(pred, for_loops_);
 
+#if 0
+  const auto prev_scope = active_scope_;
```
jjsjann123, line 2310: remove commented code and open an issue to track this as alternative for when we needed to lift the ternary into an if statement later.

csrc/codegen.cpp
```diff
@@ -402,6 +402,52 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
     }
   }
 
+  void generateVectorizedLdSt(Val* in, Val* out, CacheOp cache_op) {
```
jjsjann123, line 405: mechanical change, this is lifted from `void handle(const LoadStoreOp* ldst) final`, so it can be shared with TernaryOp handling.

jjsjann123:
!test

jjsjann123:
!test

tests/cpp/test_resize.cpp
```diff
@@ -4041,4 +4041,37 @@ TEST_F(ResizeTest, SliceSliceConcatConcat) {
   NVF_CHECK(ref.equal(cg_outputs[0]));
 }
 
+// manual scheduling that should have vectorized load on padded inputs.
+TEST_F(ResizeTest, VectorizePadLowering) {
```
zasdfgbnm, line 4045: Should we have a test for vectorizing where without using pad?

jjsjann123: good call. almost forgot that we have `where` directly :face_with_head_bandage:

csrc/codegen.cpp
```diff
@@ -1001,6 +1051,50 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
...
+        // TODO: should we have the option to specify cache level?
+        generateVectorizedLdSt(
+            top->in2(), top->out(), CacheOp::AllLevels, vector_word_size);
+        code_ << "\n";
+
+        if (out_tv->getMemoryType() == MemoryType::Local &&
+            !out_tv->isCircularBuffered()) {
+          // Vectorized initialization, explicit type conversion is needed for
+          // complex numbers
+          indent() << kTab << ": " << genVariableName(out_tv) << ".set("
```
naoyam, line 1082: I'm a little confused here. Isn't this `.set` the one defined here?

https://github.com/NVIDIA/Fuser/blob/main/runtime/array.cu#L13-L18

If so, it doesn't return anything, so I thought it shouldn't compile, does it?

csrc/codegen.cpp
```diff
@@ -1001,6 +1051,50 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
   }
 
   void handle(const TernaryOp* top) final {
+    // Get vectorization information
```
naoyam, line 1054: Can you add some comments about the expectation? IIUC, only `in2` is allowed to be vectorized, but technically speaking, it should be possible to have vectorized loads in both `in2` and `in3`, right? Not sure if it's worthwhile to allow that as well, although the required change seems minimal.

csrc/codegen.cpp
```diff
@@ -1001,6 +1051,50 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
...
+            !out_tv->isCircularBuffered()) {
+          // Vectorized initialization, explicit type conversion is needed for
+          // complex numbers
+          indent() << kTab << ": " << genVariableName(out_tv) << ".set("
+                   << genCall(out_tv->dtype(), gen(top->in3())) << ");\n";
+        } else {
+          // Note: currently arraySet option is not vectorized, so it will
+          //  rely on auto vectorization pass of cuda compiler.
+          indent() << kTab << ": "
+                   << "arraySet<" << out_tv->getDataType().value() << ", "
```
naoyam, line 1088: Same here as well

csrc/codegen.cpp
jjsjann123: Yes we can have in2 / in3 as TensorViews, I'm trying to add that since @zasdfgbnm mentioned about having a `where` test.

csrc/codegen.cpp
jjsjann123: > it doesn't return anything

no it doesn't. But we don't need to return anything, since we are setting outputs directly. We are also doing the same thing in the second operand when predicate == true.

```
predicate
  ? loadGlobalToLocal<...>(&dst[0], &src[i_src])
  : dst.set(0.0f) 
```
^^^ So dst is used as the target, we don't need to return anything from the ternary op.

jjsjann123:
!test

naoyam (approved):
LGTM

PR #3263, opened 2024-10-24
Create dispatch system for executors

csarofeen:
Separate out `ExprEvalExecutor` and `HostIrExecutor` from what's now called `KernelExecutor`. Create a dispatch system for them as compile and run are simpler for the former two.

Also renamed instances of `FusionExecutorCache` to `executor_cache`, `KernelExecutor` to `ke`, `ExprEvalExecutor` to `eee`, and `HostIrExecutor` to `hire`. It makes this PR large, but was critical to refactor all the instances of these classes.

For review focus on the following files:
csrc/host_ir/executor.[cpp,h]
csrc/runtime/executor.[cpp,h]
csrc/runtime/executor_abstract.h
csrc/runtime/executor_dispatch.[cpp,h]
csrc/runtime/fusion_executor_cache.cpp
csrc/runtime/fusion_kernel_runtime.[cpp,h]

Remaining files are just renaming. I would break this into multiple PRs, but it would be difficult to do at this point.


csrc/fusion.cpp
```diff
@@ -845,13 +845,6 @@ bool Fusion::hasDynamicTransform() {
   return !ir_utils::getTVsWithDynamicTransform(this).empty();
 }
```
csarofeen, line 847: Just moved this function to executor.cpp as it wasn't used anywhere else.

csrc/fusion_profiler.cpp
```diff
@@ -326,17 +326,15 @@ SegmentProfiler::SegmentProfiler(uint32_t id, bool cupti_disabled)
       output_bytes_(0),
       kernel_profile_state_(ProfilerState::Ready) {}
 
-void SegmentProfiler::startCompile(int device) {
-  device_ = device;
+void SegmentProfiler::startCompile() {
```
csarofeen, line 329: Separated out set device as a separate function. KernelExecutor knows device on compilation since runtime information is needed for it, the other executors set it in run.

csrc/fusion_profiler.h
```diff
@@ -22,6 +22,7 @@ namespace nvfuser {
 //! \enum ProfilerState
 //! \brief An enum used to represent the state of a profiling state machine
 enum class ProfilerState {
+  None,
```
csarofeen, line 25: Just added this to initialize the state on construction.

csrc/instrumentation.h
```diff
@@ -82,10 +82,12 @@ class TraceScope : public NonCopyable {
  public:
   explicit TraceScope(const char* event_name) : event_name_(event_name) {
     Trace::instance()->beginEvent(event_name_);
+    // std::cout << "Enter: " << event_name_ << std::endl;
```
csarofeen, line 85: Cleanup.

csrc/python_frontend/fusion_definition.cpp
```diff
@@ -345,10 +345,12 @@ std::vector<at::Tensor> FusionDefinition::execute(
         // Manual schedule
         if (!user_sched.executor->isCompiled()) {
           user_sched.executor->compileFusion(
-              user_sched.scheduled_fusion.get(),
-              inputs,
-              user_sched.fusion_id_,
-              user_sched.device_id_);
+              user_sched.scheduled_fusion.get(), inputs
```
csarofeen, line 348: Need Ryan's advice here.

csrc/runtime/executor.cpp
```diff
@@ -181,57 +201,231 @@ std::string FusionExecutor::getStructuredCode(
...
+    std::vector<GlobalBufferInfo> output_info = getBufferInfos(
+        expr_eval, PrimDataType::Int, host_ir_container_->outputs());
+    outputs = allocateOutputs(
+        host_ir_container_.get(),
+        output_info,
+        c10::Device(c10::DeviceType::CUDA, args.getDeviceIndex()),
+        expr_eval);
+  }
+
+  // TODO: If outputs are provided validate they're the correct size
```
csarofeen, line 305: Check with Sam and Jingyue if we should do this.

csrc/runtime/executor.cpp
```diff
@@ -1161,50 +1323,16 @@ std::vector<at::Tensor> FusionExecutor::runFusion(
...
-
-  if (isExpressionEvaluated(fusion())) {
-    FUSER_PERF_SCOPE("FusionExecutor::runFusion::evaluate_with_ExprEval");
-    outputs = evaluateFusionOutputs(outputs, expr_eval);
-    if (isProfilerEnabled()) {
-      auto& sprof = FusionProfiler::segment(group_id_);
-      sprof.stopKernel();
-      sprof.outputBytesAccessed(outputBytesProcessed(outputs));
+  // Temporary for refactoring as future users should use a dispatch or check
+  // isSupported before trying to compile
```
csarofeen, line 1327: TODO: Cleanup

csrc/runtime/executor.cpp
```diff
@@ -136,12 +136,32 @@ std::string getStructuredCodeFromExternalFiles(const int64_t fusion_id) {
   buffer << cuda_src.rdbuf();
   return buffer.str();
 }
-} // namespace
 
-FusionExecutor::FusionExecutor()
-    : communicator_(&Communicator::getInstance()) {}
+int64_t inputBytesProcessed(const KernelArgumentHolder& args) {
```
csarofeen, line 140: Code motion.

csrc/runtime/executor.cpp
```diff
@@ -136,12 +136,32 @@ std::string getStructuredCodeFromExternalFiles(const int64_t fusion_id) {
...
+    if (args[i]->is<at::Tensor>()) {
+      auto t = args[i]->as<at::Tensor>();
+      num_bytes += static_cast<int64_t>(t.storage().nbytes());
+    }
+  }
+  return num_bytes;
+}
 
-std::unique_ptr<PrecomputedValues>& FusionExecutor::
+int64_t outputBytesProcessed(const std::vector<at::Tensor>& outputs) {
```
csarofeen, line 152: code motion

csrc/runtime/executor.cpp
```diff
@@ -181,57 +201,231 @@ std::string FusionExecutor::getStructuredCode(
...
-    return;
   }
+  if (isProfilerEnabled()) {
+    FusionProfiler::segment(group_id_).stopKernel();
+    FusionProfiler::segment(group_id_).setDevice(args.getDeviceIndex());
+  }
+  return outputs;
+}
+
+void KernelExecutor::compileFusion(
```
csarofeen, line 397: TODO: Rename

csrc/runtime/executor.cpp
```diff
@@ -1098,51 +1294,17 @@ int64_t FusionExecutor::ensureAvailableDynamicSmemSize(
...
-  auto i =
-      std::find(fusion->outputs().begin(), fusion->outputs().end(), fusion_out);
-  NVF_ERROR(i != fusion->outputs().end());
-  auto index = std::distance(fusion->outputs().begin(), i);
-  return out_tensors[index];
-}
-} // namespace
-
-std::vector<at::Tensor> FusionExecutor::runFusion(
+std::vector<at::Tensor> KernelExecutor::runFusion(
```
csarofeen, line 1302: TODO: Rename

csrc/runtime/executor.cpp
```diff
@@ -1435,36 +1566,12 @@ std::vector<at::Tensor> FusionExecutor::runFusion(
...
-  for (auto i : c10::irange(args.size())) {
-    if (args[i]->is<at::Tensor>()) {
-      auto t = args[i]->as<at::Tensor>();
-      num_bytes += static_cast<int64_t>(t.storage().nbytes());
-    }
-  }
-  return num_bytes;
-}
-
-int64_t FusionExecutor::outputBytesProcessed(
```
csarofeen, line 1450: Code Motion

csrc/runtime/executor_abstract.cpp
```diff
@@ -0,0 +1,15 @@
+// clang-format off
```
csarofeen, line 1: TODO: Remove file, not used.

csrc/runtime/executor.h
```diff
@@ -494,19 +531,6 @@ class FusionExecutor : public NonCopyable {
   // TensorViews actually used in the kernel.
   std::vector<TensorView*> used_tvs_;
 
-  // ID of fusion in python frontend fusion cache, which maps to a single
```
csarofeen, line 497: Moved to ExecutorAbstract

csrc/runtime/fusion_executor_cache.cpp
```diff
@@ -56,22 +56,17 @@ std::vector<at::Tensor> FusionExecutorCache::runFusionWithInputs(
   if (isProfilerEnabled()) {
     FusionProfiler::start(!isProfilerEnabledWithCupti());
   }
```
csarofeen, line 58: TODO: Revert whitespace changes

csrc/runtime/fusion_kernel_runtime.h
```diff
@@ -215,7 +215,7 @@ class FusionKernelRuntime {
 
   //! something to do with parallel compilation, not sure what it's actually
   //! being used to protect.
-  std::mutex mutex_;
+  mutable std::mutex mutex_;
```
csarofeen, line 218: mutable allows to lock the mutex in a const member function

tests/cpp/test_alias.cpp
```diff
@@ -316,40 +329,36 @@ TEST_F(AliasTest, DuplicateOutputsSegmentedFusion) {
...
+          ->groups()
+          .size(),
+      2)
       << "segmentation didn't happen as expected";
 }
 
 namespace {
 
-// Returns the only executor in the most recent runtime.
-const FusionExecutor& onlyExecutorInMostRecentRuntime(
```
csarofeen, line 338: moved to utils

tests/cpp/test_tutorial.cpp
```diff
@@ -80,93 +80,94 @@ TEST_F(Tutorial, Memcpy) {
   auto options = at::TensorOptions().dtype(at::kFloat).device(at::kCUDA, 0);
   at::Tensor t0 = at::randn({32, 32}, options);
   std::vector<c10::IValue> aten_inputs = {t0};
+  {
```
csarofeen, line 83: TODO: Did something strange happen here or is this just the linter?

csrc/runtime/executor_kernel_arg.h
```diff
@@ -135,4 +135,8 @@ std::vector<std::byte> getKernelArgument(
     Val* parameter,
     PrimDataType index_type);
 
+int64_t inputBytesProcessed(const KernelArgumentHolder& args);
```
csarofeen, line 138: Code motion, used by runtim/executor and host_ir/executor

csrc/runtime/executor_kernel_arg.cpp
```diff
@@ -372,4 +372,27 @@ std::vector<std::byte> getKernelArgument(
   return polymorphicValueToBytes(pv, parameter->dtype(), index_type);
 }
 
+int64_t inputBytesProcessed(const KernelArgumentHolder& args) {
```
csarofeen, line 375: code motion

csrc/runtime/executor.h
```diff
@@ -231,12 +238,6 @@ class FusionExecutor : public NonCopyable {
   int getKernelRegisterSpills() const {
     return compiled_kernel_->register_spills;
   }
-  //! Returns the input bytes accessed for a kernel
```
csarofeen, line 234: Code motion, moved to executor_kernel_arg.[cpp,h]

tests/cpp/test_tutorial.cpp
csarofeen: I scoped it to reuse the instance name "ke"

csarofeen:
!build

csarofeen:
!test

csarofeen:
!test

csarofeen:
!test

csrc/runtime/fusion_kernel_runtime.cpp
```diff
@@ -142,28 +144,44 @@ FusionKernelRuntime::FusionKernelRuntime(
 }
 
 void FusionKernelRuntime::evictCache(size_t input_id) {
-  for (auto& fe : executors_) {
-    fe.evictCache(input_id);
+  NVF_CHECK(
```
csarofeen, line 147: Should this be removed?

csarofeen:
@wujingyue and @samnordmann for the host_ir files
@naoyam for the rest of the PR

csrc/python_frontend/fusion_definition.cpp
csarofeen: @rdspring1 another place I could use your help, please see the comment below.

wujingyue (commented):
Can you clarify the motivation? 

> Separate out ExprEvalExecutor and HostIRExecutor from what's now called KernelExecutor.

This looks good -- it refactors quite some complexity in FusionExecutor. 

> Create a dispatch system for them as compile and run are simpler for the former two.

This seems to suggest a follow-up change. What will this "dispatch system" enable?

csarofeen:
Dispatch system just keeps logic simple within KernelRuntime as it may need to run one of three executors, so it doesn't need to directly implement the logic to figure out which one and how to run each. It's all in executor_dispatch.[cpp,h].

csrc/runtime/executor_abstract.h
```diff
@@ -0,0 +1,49 @@
...
+#include <runtime/executor_kernel_arg.h>
+#include <runtime/executor_params.h>
+#include <runtime/executor_utils.h>
+#include <scheduler/scheduler_types.h>
+
+namespace nvfuser {
+
+class ExecutorAbstract : public PolymorphicBase, public NonCopyable {
+ public:
+  //   // NVF_API was added for nvfuser_extension. See examples/sinh_extension.
```
wujingyue, line 20:
```suggestion
```
as long as CI passes without NVF_API

csrc/host_ir/executor.cpp
```diff
@@ -7,13 +7,157 @@
...
+    int64_t fusion_id,
+    int64_t concrete_id,
+    int64_t runtime_id,
+    int64_t group_id)
+    : ExecutorAbstract(fusion_id, concrete_id, runtime_id, group_id),
+      communicator_(&Communicator::getInstance()) {}
+
+bool HostIrExecutor::supported(Fusion* fusion) {
+  FUSER_PERF_SCOPE("HostIrExecutor::supported");
+  if (fusion->isA<hir::HostIrContainer>()) {
```
wujingyue, line 35: Do you expect ::supported to be called both pre and post compilation? Why is post needed? I presume after compilation we get a list/DAG of ExecutorDispatch so we can simply call ExecutorDispatch::run for each.

csarofeen:
!test

naoyam:
Pinging @samnordmann. I'll try to land the PR, but would be great if any concern would be resolved first if @wujingyue and @samnordmann had any.

csrc/host_ir/executor.cpp
csarofeen: I'm having trouble linking your comment to this line of code. I believe this is here because some of the tests are directly creating a HostIrContainer, and sending that to execution. So, for those cases I wanted to make sure the HostIrExecutor was selected. Maybe I misunderstood and this isn't how the execution path should go, though all the tests are working now (barring maybe one that's rerunning after merging main in).

csarofeen:
@wujingyue 
Re:
> Can you clarify the motivation?

The reason I wanted to separate this out is in a big part because I want to keep host execution logic separated from running a kernel. It's reasonable for HostIrContainers to contain "HostUnits" (which really seem like KernelUnits), but it didn't make sense for me that the Host execution goes through the execution path that runs the Kernels.

Ideally, I'd like to get KernelExecution down to the point where the only thing it actually runs is the kernel, and everything else ends up being in the host program. This seemed to be a step on that path.

csrc/runtime/executor_dispatch.h
```diff
@@ -0,0 +1,57 @@
...
+  static void compile(
+      std::unique_ptr<ExecutorAbstract>& executor,
+      Fusion* fusion);
+
+  static void compile(
+      std::unique_ptr<ExecutorAbstract>& executor,
+      Fusion* fusion,
+      const KernelArgumentHolder& args,
+      const LaunchParams& launch_constraints,
+      CompileParams compile_params,
```
wujingyue, line 37:
```suggestion
      const CompileParams& compile_params,
```

csrc/runtime/executor_dispatch.h
```diff
@@ -0,0 +1,57 @@
...
+  static std::vector<at::Tensor> run(
+      std::unique_ptr<ExecutorAbstract>& executor,
+      KernelArgumentHolder& args,
+      std::vector<at::Tensor> outputs = {});
+
+  static std::vector<at::Tensor> run(
+      std::unique_ptr<ExecutorAbstract>& executor,
+      KernelArgumentHolder& args,
+      const LaunchParams& launch_constraints = LaunchParams(),
+      CompileParams compile_params = CompileParams(),
```
wujingyue, line 51: I'm unsure if both are needed for compile and run. Conceptually, CompileParams should be used for compile only, and LaunchParams for run.

csrc/host_ir/executor.cpp
wujingyue: > I believe this is here because some of the tests are directly creating a HostIrContainer, and sending that to execution.

This is normal. However, I'd expect HostIrEvaluator to pick up that HostIrContainer, and give HostUnit in PostOnStream to ExecutorAbstract. That HostUnit, IIRC, contains a plain fusion instead of another HostIrContainer. 

(great to know tests are passing; happy with addressing this in a separate PR)

cc @samnordmann

wujingyue (approved):

samnordmann:
> @wujingyue Re:
> 
> > Can you clarify the motivation?
> It's reasonable for HostIrContainers to contain "HostUnits" (which really seem like KernelUnits)

Agreed, up to the fact that for now a HostUnit (which is really a Fusion) can be further segmented into several kernels (if the HostIrExecutor has the option to use FusionExecutorCache). This being said, we might want to remove that in the future.

> but it didn't make sense for me that the Host execution goes through the execution path that runs the Kernels.

IIUC I agree. Is it correct to rephrase it as "KernelExecutor shouldn't call HostIrExecutor"? Are you suggesting to have a strict ordering 
FusionExecutorCache > HostIrExecutor > KernelExecutor,
where "x>y" means "x should be able to call y but not conversely"? 


> Ideally, I'd like to get KernelExecution::run down to the point where the only thing it actually runs is the kernel, and everything else ends up being in the host program. This seemed to be a step on that path.
> 
> This also seemed on path to have HostIr containing multiple "HostUnit"s that it's coordinating between.

Agreed!

csrc/host_ir/executor.h
```diff
@@ -13,30 +13,59 @@
...
+  NVF_API std::vector<at::Tensor> run(
+      KernelArgumentHolder& args,
+      std::vector<at::Tensor> outputs = {});
+
+  const std::unique_ptr<hir::HostIrContainer>& hostContainer() {
+    return host_ir_container_;
+  }
+
+ private:
+  std::unique_ptr<hir::HostIrContainer> host_ir_container_;
```
samnordmann, line 47: nit:
```suggestion
  std::unique_ptr<hir::HostIrContainer> host_ir_container_ = nullptr;
```
since `isCompiled` relies on this

csrc/host_ir/executor.h
```diff
@@ -13,30 +13,59 @@
...
+
+  void compile(Fusion* fusion);
+
+  bool isCompiled() const override;
+
+  NVF_API std::vector<at::Tensor> run(
+      KernelArgumentHolder& args,
+      std::vector<at::Tensor> outputs = {});
+
+  const std::unique_ptr<hir::HostIrContainer>& hostContainer() {
```
samnordmann, line 42: Is this method needed? IIUC the host container shouldn't completely private.

Anyway, I would suggest
```suggestion
  const hir::HostIrContainer* hostContainer() {
```

@wujingyue for confirmation

csrc/host_ir/executor.cpp
```diff
@@ -7,13 +7,157 @@
...
+  NVF_ERROR(
+      supported(fusion),
+      "HostIrExecutor does not support the Fusion provided.");
+  if (isProfilerEnabled()) {
+    FusionProfiler::segment(group_id_).startCompile();
+  }
+
+  if (fusion->isA<hir::HostIrContainer>()) {
+    host_ir_container_ = std::make_unique<hir::HostIrContainer>();
+    IrCloner cloner = Fusion::copy(fusion, host_ir_container_.get());
```
samnordmann, line 66: those two lines belong to if and else branches and could be move outside the block

csrc/host_ir/executor.cpp
```diff
@@ -7,13 +7,157 @@
...
+    for (auto expr : fusion->as<hir::HostIrContainer>()->topLevelExprs()) {
+      host_ir_container_->pushBackTopLevelExprs(cloner.clone(expr));
+    }
+  } else {
```
samnordmann, line 70: Is this else branch exercised anywhere? In my opinion it could be removed

csrc/host_ir/executor.cpp
samnordmann, line 66: Can't we avoid the copy here? Why taking `HostIrContainer*` pointer as input and then copy it to a `unique_ptr`?

csrc/host_ir/executor.cpp
```diff
@@ -7,13 +7,157 @@
...
+    FusionProfiler::segment(group_id_).stopCompile();
+  }
+}
+
+bool HostIrExecutor::isCompiled() const {
+  return host_ir_container_ != nullptr;
+}
+
+namespace {
+// Host IR specific function, returns the at:Tensor (ordered list) associated
```
samnordmann, line 93: In general, I found that passing the tensors through `unordered_map<Val*, at::Tensor>` is way more powerful and straightforward than dealing with `std::vector<at::Tensor>`, because with the latter we always need to find the relative index and loose track of intermediate tensors. In addition, I don't see any benefit in keeping an order in the inputs.

A map would allow to get rid of this type of helper function.

Also, I think the inputs, outputs and intermediate tensors should undergo an uniform treatment inside the Executor. Ideally, we want to allow the user to provide some preallocated output or intermediate buffer, or at least to handle allocations of those buffers independently of the execution.

csrc/host_ir/executor.cpp
```diff
@@ -7,13 +7,157 @@
...
+  return false;
+}
+
+void HostIrExecutor::compile(Fusion* fusion) {
+  FUSER_PERF_SCOPE("HostIrExecutor::compile");
+  NVF_ERROR(
+      supported(fusion),
+      "HostIrExecutor does not support the Fusion provided.");
+  if (isProfilerEnabled()) {
+    FusionProfiler::segment(group_id_).startCompile();
```
samnordmann, line 61: I am not sure what this means... Has it been tested?

csrc/host_ir/executor.cpp
```diff
@@ -7,13 +7,157 @@
...
+    NVF_CHECK(
+        group_id_ >= 0,
+        "An invalid segment id is passed to FusionProfiler!:",
+        group_id_);
+    SegmentProfiler& sprof = FusionProfiler::segment(group_id_);
+    sprof.inputBytesAccessed(computeBytes(args));
+    sprof.scheduler(toString(SchedulerType::ExprEval));
+    sprof.startKernel();
+  }
+  NVF_ERROR(host_ir_container_, "Need to compile before you can run.");
```
samnordmann, line 121: nit
```suggestion
  NVF_ERROR(host_ir_container_ != nullptr, "Need to compile before you can run.");
```

csrc/host_ir/executor.cpp
```diff
@@ -7,13 +7,157 @@
...
+  // Bind fusion inputs
+  auto expr_eval = executor_utils::bindInputs(args, host_ir_container_.get());
+
+  if (outputs.empty()) {
```
samnordmann, line 125: Why do we need to allocate anything in the HostIrExecutor?

csrc/host_ir/executor.cpp
```diff
@@ -7,13 +7,157 @@
...
+    std::vector<GlobalBufferInfo> output_info = getBufferInfos(
+        expr_eval, PrimDataType::Int, host_ir_container_->outputs());
+    outputs = allocateOutputs(
+        host_ir_container_.get(),
+        output_info,
+        c10::Device(c10::DeviceType::CUDA, args.getDeviceIndex()),
+        expr_eval);
+  }
+
+  // TODO: If outputs are provided validate they're the correct size
```
samnordmann, line 135: shouldn't this be done by `ExpressionEvaluator::bind` ?

samnordmann (commented):
1) IIUC, `HostIrExecutor` is not called in any test, but only `HostIrEvaluator`, right? If so, why keeping it?

2) I don't understand the implementation of `HostIrExecutor` and what it is supposed to do. In my mind it should just be a thin wrapper around `HostIrInterpreter`. Can you help me understand? I see some duplicate logic in the HostIr Executor/Interpreter

csrc/runtime/executor_abstract.h
csarofeen: I need to recheck this, I'm guessing the comment can be removed, as I imagine the extension would use KernelExecutor not ExecutorAbstract.

csrc/runtime/executor_dispatch.cpp
```diff
@@ -0,0 +1,141 @@
...
+    return eee->isCompiled();
+  }
+  if (auto ke = dynamic_cast<KernelExecutor*>(executor.get())) {
+    return ke->isCompiled();
+  }
+  NVF_THROW("Unsupported Executor detected.");
+}
+
+std::vector<at::Tensor> ExecutorDispatch::run(
+    std::unique_ptr<ExecutorAbstract>& executor,
```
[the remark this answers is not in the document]
csarofeen: I'll revisit, I think I tried and decided for some reason it wasn't working out well, but I can't recall why. So will try again.

csrc/runtime/executor_dispatch.h
csarofeen: Good point, will double check. This might have been blind copy-paste.

csarofeen:
@wujingyue 

> This is normal. However, I'd expect HostIrEvaluator to pick up that HostIrContainer, and give HostUnit in PostOnStream to ExecutorAbstract. That HostUnit, IIRC, contains a plain fusion instead of another HostIrContainer.

I'll take another look quickly to understand if this was a needed change for the tests. If not I can revert and we can follow up. If it was, I'd like to leave it as is and follow up either way, though there's another project I want to do first so it could sit there as is for a month.

csarofeen:
@samnordmann

> IIUC I agree. Is it correct to rephrase it as "KernelExecutor shouldn't call HostIrExecutor"? Are you suggesting to have a strict ordering
> FusionExecutorCache > HostIrExecutor > KernelExecutor,
> where "x>y" means "x should be able to call y but not conversely"?

I'm not suggesting a strict ordering, though I think that could be logical in the longer term. I mostly wanted the HostIrExecutor logic out of what's now called KernelExecutor. That way we can evolve them separately into potentially the structure you recommend (potentially as in I think that's the right structure, I'm just not 100% until we draft it).

csrc/runtime/executor_abstract.h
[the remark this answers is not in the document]
csarofeen: I think that's right, will double check, closing as duplicate Jingyue also mentioned this.

csrc/host_ir/executor.h
```diff
@@ -91,11 +120,11 @@ class HostIrExecutor final : public OptOutDispatch {
...
   Communicator* communicator_;
-  HostIrExecutorParams params_;
+  HostIrEvaluatorParams params_;
   // Stores concrete computed values
   ExpressionEvaluator expr_evaluator_;
-  // Cache Fusions, FusionExecutors
-  std::unordered_map<HostUnit*, FusionExecutor> fe_;
+  // Cache Fusions, KernelExecutors
+  std::unordered_map<HostUnit*, std::unique_ptr<ExecutorAbstract>> executors_;
   std::unordered_map<HostUnit*, FusionExecutorCache> fec_;
```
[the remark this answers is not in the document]
csarofeen: That's a good point, I'd like to refactor our caching system which would hit FusionExecutorCache, KernelRuntime, and KernelExecutor. I'll look at having FusionExecutorCache inherit from AbstractExecutor when considering what to do there.

csrc/host_ir/executor.cpp
```diff
@@ -7,13 +7,157 @@
...
+            [](Expr* e) {
+              return isResharding(e) && isLowerableToCommunication(e);
+            }),
+        "Could not execute fusion as all expressions in a host IR container must be communication based at this point.");
+    return true;
+  }
+  return false;
+}
+
+void HostIrExecutor::compile(Fusion* fusion) {
```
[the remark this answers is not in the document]
csarofeen: Yes, it will be called through the FusionExecutorCache path just as compile is called for the other executors.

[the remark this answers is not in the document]
csarofeen: It does very little today, but that would be the point at which we could lower the fusion into a HostIr program.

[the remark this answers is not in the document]
csarofeen: i.e. it should eventually call a `HostLower` just like `KernelExecutor::Compile` calls `GPULower`.

csarofeen:
@samnordmann 

> That HostUnit, IIRC, contains a plain fusion instead of another HostIrContainer.
> 
> That is correct. Although, I had in mind to let in the future a HostUnit contain a HostIrContainer as well, but it''s not strictly needed for now

Cool, let's talk about that later, as you mentioned maybe we want a one-way structure to keep things simpler.

csrc/host_ir/executor.h
csarofeen: I don't recall writing this function, I wonder if it was just copy-pasted from `FusionExecutor` maybe. I'll need to check.

csrc/host_ir/executor.cpp
csarofeen: They couldn't so I duplicated them. I'll revisit to explain why.

csarofeen: I think I more or less copy-pasted this code, will double check.

csrc/host_ir/executor.cpp
wujingyue: This comes from https://github.com/NVIDIA/Fuser/blob/ea9c9135dc15fb118e660a9d28480c2942f89859/csrc/runtime/executor.cpp#L1178-L1205. In the current FusionExecutorCache integration, each segment is compiled/run via an ExecutorAbstract. A resharding Expr forms a no-op segment and, with this PR, is picked up by HostIrExecutor. 

I understand this won't be the end state. Your statement makes sense for HostIrEvaluator.

csrc/host_ir/executor.cpp
csarofeen: It's just a profiling mechanism to time the compilation with FusionProfiler. CC @kevinstephano to review, it's used for the FusionExecutor compilation path which had HostIrExecutor logic in it, so it's unlikely to regress or fail.

csrc/host_ir/executor.cpp
csarofeen: It's basically doing this on the fusion path, and it's to take ownership of the IR which I believe it should.

csrc/host_ir/executor.cpp
wujingyue: I believe this comes from https://github.com/NVIDIA/Fuser/blob/ea9c9135dc15fb118e660a9d28480c2942f89859/csrc/runtime/executor.cpp#L211-L225, and is tested by most of the multi-GPU tests, which use FusionExecutorCache. 

However, I agree with you that it could be removed in the future, after the "Host IR lowering for non-overlapping integrated to FusionExecutorCache" milestone.

csarofeen: Isn't this the common path?
As it exists today:
https://github.com/NVIDIA/Fuser/blob/main/csrc/runtime/executor.cpp#L211-L225

csrc/host_ir/executor.cpp
csarofeen: This follows the explicit contract with user's/integration. This contact is followed by KernelExecutor, FusionExecutorCache, and KernelRuntime, so I'm trying to stay consistent.

csarofeen: I agree for inputs/outputs, but why would we do so for intermediate buffers? Who would provide those allocations?

csrc/host_ir/executor.cpp
csarofeen: It's a unique_ptr not a raw pointer so checking it in a conditional is valid.

csrc/host_ir/executor.cpp
csarofeen: For inputs it would throw an error, for outputs it might show up during evaluation and I'm uncertain if it would be an easy to debug error message. Ideally in the future, we'll have allocations as explicit statements that could be easily checked.

csrc/host_ir/executor.h
csarofeen: I did write this function, and I think it was in the spirit of KernelExecutor being able to return its Fusion:
https://github.com/NVIDIA/Fuser/blob/main/csrc/runtime/executor.h#L195
or Kernel:
https://github.com/NVIDIA/Fuser/blob/main/csrc/runtime/executor.h#L190

I should make the function const. These functions are typically used in testing and I imagine would serve a similar purpose here.

csrc/host_ir/executor.h
csarofeen: I'll change it in isCompiled as for unique_ptr's we usually just condition on the object itself to know if it has been initialized.

csrc/host_ir/executor.cpp
csarofeen: Race conflict with Jinygue 😆

csarofeen:
Adding @rdspring1 to review changes to serialization of FusionExecutor.

csrc/host_ir/executor.h
samnordmann: As long as it is `const` I guess it is ok

csrc/host_ir/executor.cpp
samnordmann: > It's basically doing this on the fusion path, and it's to take ownership of the IR which I believe it should.

In this case it would preferable imo to take a `unique_ptr<Fusion>` as input and remove the copy, so that the function's caller is free to either copy or move the ownership

csrc/host_ir/executor.cpp
samnordmann: > This follows the explicit contract with user's/integration. This contact is followed by KernelExecutor, FusionExecutorCache, and KernelRuntime, so I'm trying to stay consistent.

This contract makes things much harder than they should in my opinion. What I am proposing is to change this contract to pass `unordered_map<Val*, at::Tensor>` instead of `std::vector<at::Tensor>`. I see no advantage of `std::vector` version, where we constantly need to retrieve the corresponding Val associated with the concrete tensor. 
This being said, nothing prevents us to have an overloaded user facing entrypoint taking `std::vector`, that would only be thin wrapper around the one taking `unorderd_map`.  

> I agree for inputs/outputs, but why would we do so for intermediate buffers? Who would provide those allocations?

The motivation behind this proposal is to have a uniform and more flexible treatment of allocation and pre-allocated buffers accross all level of execution:
- We might want to only pre-allocate some outputs or intermediate buffers and not all of them, and there is no way to indicate this when passing `std::vector<at::Tensor>`
- When dealing with a hierarchical DAG, with several segmentations, some intermediate tensors become I/O of sub-fusion
- When dealing with inter-device communication we need to pre-allocate recv buffers ; but those can be intermediate buffers
- In host pipelines, we sometimes need to pre-allocate the output and write to it slice by slice.

csrc/host_ir/executor.cpp
csarofeen: I think you're right Jingyue, I must have been mistaken when adding this, it looks like no tests rely on it. Must have been my misinterpretation of what was happening in the host_ir tests. Will revert and double check the tests pass.

samnordmann (commented):
I'm realizing that most of my concerns deal with code that has just been copy-pasted, and so might not be directly related to this PR. Let me try to express them here anyway, though I'm ok to address them later in a separate PR if that makes upstreaming easier.

The execution flow I have in mind can be summarized (after many simplifications, e.g., discarding fusion definition and scheduling) as follow:
a) **Entrypoint:** user call `FusionExecutorCache` on the Fusion
b) **Lower to HostIr**: "Host" Segmentation and creation of a `HostIrContainer` representing the host program. In the simple single-device case, the Host program only consists of a linear list of kernels.
c) **Host Execution** through `HostIrExecutor` (called `HostIrEvaluator` in this PR) that is repsonsible for executing the `HostIrContainer` produced at the previous step.
d) **GPU execution**: `HostIrExecutor` may call `FusionExecutor` (called `KernelExecutor` in this PR) for lowering, codegen-ing and executing kernels, or call another backend like `ATen`.

One key point of this flow is the strict ordering described [here](https://github.com/NVIDIA/Fuser/pull/3263#issuecomment-2458151841) ; in particular, `FusionExecutorCache` never calls `KernelExecutor` directly, but only through `HostIrExecutor`

As a consequence, what is implemented under `HostIrExecutor` in this PR is not relevant imo, and seems like a too cheap way to integrate HostIR in the flow:
1) `HostIrExecutor` do not call `HostIrEvaluator` nor any other Host Ir utility. It duplicates all the logic, for execution, allocation, handling I/O, etc. There should only be one Host Ir Executor, not `HostIrExecutor` and `HostIrEvaluator`.
2) `HostIrExecutor` as it is implemented in this PR is very limited as it only ops that are pure communications. This won't support comms/compute pipelines...
3) We should separate Host Ir lowering, compilation, and execution. However, here, what should be called "lowering" is implemented inside `HostIrExecutor::compile`. I have already implemented what I think should be the lowering pass from `Fusion` to `HostIrContainer`: see `MultiDeviceExecutor`'s conctructor here https://github.com/NVIDIA/Fuser/blob/main/csrc/multidevice/executor.cpp#L82-L180. This logic should be scooped at and called inside `FusionExecutorCache` to implement step "b)" above.

Wdyt?

csrc/runtime/executor_dispatch.cpp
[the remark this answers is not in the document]
csarofeen: Changed

csrc/runtime/executor_dispatch.h
csarofeen: done

csarofeen:
!test

csrc/host_ir/executor.cpp
samnordmann: FWIW, binding inputs and potential other preallocated buffers holds in one line in HostIrExecutor, :
https://github.com/NVIDIA/Fuser/blob/967c2824cfdf967901706d7aee1f57a9ab91f261/csrc/host_ir/executor.cpp#L69

csrc/runtime/fusion_kernel_runtime.cpp
```diff
@@ -142,28 +144,44 @@ FusionKernelRuntime::FusionKernelRuntime(
...
     flatbuffers::FlatBufferBuilder& builder) const {
   // See table definition for FusionKernelRuntime in serde/fusion_cache.fbs
 
-  // 1. Serialize FusionExecutor objects
-  std::vector<flatbuffers::Offset<serde::FusionExecutor>> executors_fb;
+  // This check needs to be disabled though it seems strange to me that we
+  // would/could/should serialize KRTs that aren't compiled. For now the pytorch
+  // tests require it.
+  //
+  // NVF_CHECK(
```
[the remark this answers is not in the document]
rdspring1: I enabled the check and the tests in `test/python/test_python_frontend.py` run correctly.

csrc/host_ir/executor.cpp
csarofeen: You're absolutely right. I remember duplicating this because it wasn't building, but taking them out works fine now, so I don't know what was going on then. Thanks.

csrc/host_ir/executor.cpp
samnordmann: It should throw for both inputs and outputs (and also any intermediate buffers). If we first bind the inputs, then the shape of the output is known to ExpressionEvaluator (i.e., can be evaluated), thus, when binding preallocated outputs buffers will assert that the extents are matching

samnordmann:
> I don't think it's necessary to segment twice, for host and for device. As implemented today, using SchedulerEntry::canSchedule*, one pass is sufficient to propose which parts of the complete fusion (e.g. resharding expressions) to run on host and the kernel boundaries. One detailed challenge, regardless of one-pass or two-pass, is a sum along a host dimension. It can't run entirely on host (the addition would be super slow) or entirely on device (would disable overlapping). Host IR lowering has to turn that into a for-loop of additions and fuse the addition into the previous kernel.

It is probably ok for now, but I am unsure about that in the long run. We might need hierarchical scheduling and segmentation at some point. For example, if we want to pipeline Allgather + GEMM for fine grain overlap, we might want to schedule the Host IR program leaving the GEMM as a "bulked" HostUnit; but then scheduling the GEMM might need a further segmentation into several kernels. Does it make sense?

samnordmann:
> One detailed challenge, regardless of one-pass or two-pass, is a sum along a host dimension. It can't run entirely on host (the addition would be super slow) or entirely on device (would disable overlapping). Host IR lowering has to turn that into a for-loop of additions and fuse the addition into the previous kernel.

I am not sure to understand why there is specific challenge here. Don't we just need to accumulate accros the host for-loop iteration? HostIr Executor can support that through a Host Unit, or we could also easily add support for `at::sum_out` and `at::add_out`

naoyam:
@csarofeen I merged #3349. I did `git merge -s ours`, so nothing should be overwritten by the merge.

naoyam:
!test

naoyam:
!test

wujingyue:
> For example, if we want to pipeline Allgather + GEMM for fine grain overlap, we might want to schedule the Host IR program leaving the GEMM as a "bulked" HostUnit; but then scheduling the GEMM might need a further segmentation into several kernels. Does it make sense?

Maybe -- I'm unsure what it buys us to "leave the GEMM as a bulked HostUnit". Before host IR lowering, we already know which IterDomains in loop domains are host-parallelized. So I believe host IR lowering can run a compute-at-map analysis to generate a host for-loop regardless of how many HostUnits/segments the loop body contains.

wujingyue:
> HostIr Executor can support that through a HostUnit with aliased I/O

Yes. I was talking about the mechanism to generate that HostUnit/Fusion with I/O aliases. It's [easy](https://github.com/NVIDIA/Fuser/blob/f5b903e94c6ac3bd12914214a1e3be7509423e1b/csrc/device_lower/pass/index.cpp#L604) to device-lower a sum into a loop of additions, because it's put in one IrContainer, the kernel. However, host IR lowering has to deal with multiple containers. (I wouldn't be surprised at all if you know how to implement this -- I am just unsure myself)

> or we could also easily add support for at::sum_out and at::add_out

Yes but suboptimal -- inputs to `sum_out` and `add_out` would have to be materialized to global memory.

samnordmann:
> > For example, if we want to pipeline Allgather + GEMM for fine grain overlap, we might want to schedule the Host IR program leaving the GEMM as a "bulked" HostUnit; but then scheduling the GEMM might need a further segmentation into several kernels. Does it make sense?
> 
> Maybe -- I'm unsure what it buys us to "leave the GEMM as a bulked HostUnit". Before host IR lowering, we already know which IterDomains in loop domains are host-parallelized. So I believe host IR lowering can run a compute-at-map analysis to generate a host for-loop regardless of how many HostUnits/segments the loop body contains.

I mean that the scheduler will be applied in a hierarchical way. We will need to "host" schedule "AG+GEMM" on the one hand, and also to schedule the GEMM on the other hand. So the two scheduler need to be kind of composable, and be applied one after the other on overlapping segments. As long as segmentation and scheduling are tied together, this also holds for segmentation. 


> > HostIr Executor can support that through a HostUnit with aliased I/O
> 
> Yes. I was talking about the mechanism to generate that HostUnit/Fusion with I/O aliases. It's [easy](https://github.com/NVIDIA/Fuser/blob/f5b903e94c6ac3bd12914214a1e3be7509423e1b/csrc/device_lower/pass/index.cpp#L604) to device-lower a sum into a loop of additions, because it's put in one IrContainer, the kernel. However, host IR lowering has to deal with multiple containers. (I wouldn't be surprised at all if you know how to implement this -- I am just unsure myself)
> 
> > or we could also easily add support for at::sum_out and at::add_out
> 
> Yes but suboptimal -- inputs to `sum_out` and `add_out` would have to be materialized to global memory.

By definition, if it's a host operation, we will need to produce the data on global memory. That is true for any host op, also in the case it is a HostUnit with aliased I/O. So I am not sure to understand what you are suggesting.

There is no Host lowering today so creating this HostUnit with aliased I/O is not implemented, however, it doesn't seem hard to implement, unless I'm missing something...

wujingyue:
> I think that the HostUnit alternative will not achieve anything more than what add_out and sum_out can achieve. 

I think the HostUnit alternative gives fewer global reads/writes when the addition can be fused to the preceding kernel. 

```
c = sum_H(a*b)  # H means along a host-parallel dimension
```

With the HostUnit alternative, each iteration reads `a`, reads `b`, reads `c`, computes `c+a*b` and writes that as the updated `c`. 

With `sum_out` or `add_out`, each iteration reads `a`, reads `b`, writes `a*b`, reads `a*b`, reads `c`, computes `c+a*b`, and writes the updated `c`.

samnordmann:
> > I think that the HostUnit alternative will not achieve anything more than what add_out and sum_out can achieve.
> 
> I think the HostUnit alternative gives fewer global reads/writes when the addition can be fused to the preceding kernel.
> 
> ```
> c = sum_H(a*b)  # H means along a host-parallel dimension
> ```
> 
> With the HostUnit alternative, each iteration reads `a`, reads `b`, reads `c`, computes `c+a*b` and writes that as the updated `c`.
> 
> With `sum_out` or `add_out`, each iteration reads `a`, reads `b`, writes `a*b`, reads `a*b`, reads `c`, computes `c+a*b`, and writes the updated `c`.

However here the accumulation across iteration is still done on a globally allocated buffer (`c`).

I think what you are describing here corresponds to fusing operations within the host for-loop's body, say into one HostUnit, which gives the classical benefit of fusing kernels. However, my point is that fusing across iterations is not possible, and the I/O of the for-loop's body (here, `a`, `b`, and `c`, but not `a*b` indeed) must be on global memory.

wujingyue:
> However, my point is that fusing across iterations is not possible

Agreed! I wasn't trying to argue about that.

csarofeen:
@samnordmann @wujingyue this conversation seems pretty great, it'd be wonderful if you could capture it in a design doc.

csrc/python_frontend/fusion_definition.cpp
naoyam: @rdspring1 Could you take a look here?

csrc/runtime/executor.cpp
```diff
@@ -185,53 +178,96 @@ std::string KernelExecutor::getStructuredCode() const {
...
+bool ExprEvalExecutor::isCompiled() const {
+  return fusion_ != nullptr;
+}
+
+std::vector<at::Tensor> ExprEvalExecutor::run(
+    KernelArgumentHolder& args,
+    std::vector<at::Tensor> outputs) {
+  FUSER_PERF_SCOPE("ExprEvalExecutor::run");
+
+  if (isProfilerEnabled()) {
```
naoyam, line 212: @csarofeen Don't we need to set the current device here like line 242?

csrc/runtime/executor.cpp
```diff
@@ -185,53 +178,96 @@ std::string KernelExecutor::getStructuredCode() const {
...
+        auto out_tensor =
+            expr_eval.evaluate(out_val->as<TensorView>()).as<at::Tensor>();
+        expr_eval.bind(out_val, out_tensor);
+        outputs.emplace_back(out_tensor);
+      }
+    }
+  }
+  if (isProfilerEnabled()) {
+    FusionProfiler::segment(group_id_).stopKernel();
+    FusionProfiler::segment(group_id_).setDevice(args.getDeviceIndex());
```
naoyam, line 242: @csarofeen  Why is `setDevice` done after the kernel execution?

naoyam:
@Priya2698 Could you check the benchmark profiling with this PR? There should be no performance change, but since there's a lot of code changes, we should make sure everything works as expected.

Priya2698:
> @Priya2698 Could you check the benchmark profiling with this PR? There should be no performance change, but since there's a lot of code changes, we should make sure everything works as expected.

Are you interested in a complete sweep or only the host benchmarking?

Priya2698:
> > @Priya2698 Could you check the benchmark profiling with this PR? There should be no performance change, but since there's a lot of code changes, we should make sure everything works as expected.
> 
Are you interested in a complete sweep or only the host benchmarking? We can run a complete sweep on the CI.
CC: @xwang233

naoyam:
> > > @Priya2698 Could you check the benchmark profiling with this PR? There should be no performance change, but since there's a lot of code changes, we should make sure everything works as expected.
> 
> Are you interested in a complete sweep or only the host benchmarking? We can run a complete sweep on the CI. CC: @xwang233

Please do a complete sweep just in case. Either A100 or H100. Not necessary to check both.

Priya2698:
> > > > @Priya2698 Could you check the benchmark profiling with this PR? There should be no performance change, but since there's a lot of code changes, we should make sure everything works as expected.
> > 
> > 
> > Are you interested in a complete sweep or only the host benchmarking? We can run a complete sweep on the CI. CC: @xwang233
> 
> Please do a complete sweep just in case. Either A100 or H100. Not necessary to check both.

Got it, we will need to use CI resources then since the runs time out due to dlcluster time limits.
@xwang233 will be able to help. We can run preferably run on A100 due to more availability.

xwang233:
!test --pybench-full

csrc/runtime/executor.cpp
csarofeen: Fusion Profiler is just a logging system, it coordinates/accumulates information based on group_id_. As long as we set device correctly once it's fine.

csrc/runtime/executor.cpp
csarofeen: No particular reason, either way should be functional.

csrc/runtime/executor.cpp
naoyam: I understand that, but when this `run` function is first called, is it guaranteed to have the profiler have the correct device already set?

naoyam:
@Priya2698 @xwang233 Could you please let us know how the benchmarking went?

Priya2698:
> @Priya2698 @xwang233 Could you please let us know how the benchmarking went?

Comparing the pipeline results to the latest nvfuser nightly results does not show any major difference: http://nv/eno
Please verify against nightly results, since weekly has additional inputs.

naoyam:
Thanks. I don't see anything suspicious either.

Where can we find host benchmark results?

xwang233:
> Thanks. I don't see anything suspicious either.
> 
> Where can we find host benchmark results?

host latency benchmarks for this PR were executed with kernel_reuse disabled, so dynamic results might be inaccurate

Priya2698:
I am seeing similar results for all cases except `test_many_segments_host_benchmark` for `dynamic`. This PR does not contain changes from PR #3388 so this is expected. Apart from that, I do not see any deviations.

On `A100-asus`:

**Branch: main**
Screenshot 2024-11-13 at 7 56 18 PM
Screenshot 2024-11-13 at 7 56 38 PM
Screenshot 2024-11-13 at 7 57 16 PM


**Branch: Current PR**
Screenshot 2024-11-13 at 8 07 58 PM
Screenshot 2024-11-13 at 8 08 26 PM
Screenshot 2024-11-13 at 8 08 55 PM

naoyam:
@csarofeen Everything looks good. I'll run the final CI just in case.

naoyam:
!test

naoyam (approved):

csrc/runtime/executor.cpp
naoyam: `setDevice` just records which device is used, so it doesn't matter when it's called. I thought it does something more than that.

Issue #3022, opened 2024-09-26
Enable codegen for matmul/linear in nvfuser through Thunder


Priya2698:
This task will involve development within nvFuser and Thunder.

**Current status:**
In Thunder, we have the compile flags `nvf_enable_matmul` and `nvf_enable_linear` to allow nvfuser to optionally accept `matmul/linear` operators. By default, these are executed through ATen via our `ExprEvalScheduler`.

**Next steps:**
Within nvfuser, we have the ability to fuse matmul/linear operations in certain cases using `NVF_ENABLE=fuse_matmul` and disable ATen evaluation through `NVF_DISABLE=matmul_expr_eval`. For experimentation directly through Thunder, we want to allow setting these options from within Thunder.

Related efforts:
1. https://github.com/NVIDIA/Fuser/pull/1905

- Similar to this approach, we can specifically add additional arguments in `fd.execute` to control the above two global enums and exercise different combinations of `Matmul/ExprEvalScheduler`.

2. https://github.com/NVIDIA/Fuser/pull/2077

- More generally, we can expose these global enums through the python interface of `fd.execute`.

In both these cases, within Thunder, we can add the ability to pass additional flags to the nvfuser executor to set these options.

_Self-note: Follow-up with a design doc after preliminary discussion._

CC: @jacobhinkle @kevinstephano

crcrpar:
> Within nvfuser, we have the ability to fuse matmul/linear operations in certain cases using NVF_ENABLE=fuse_matmul and disable ATen evaluation through NVF_DISABLE=matmul_expr_eval.

Would this option have an effect on both matmul and linear?
If so, I would find it reasonable to unify the two kwarg `nv_enable` options in Thunder into one.

Priya2698:
> > Within nvfuser, we have the ability to fuse matmul/linear operations in certain cases using NVF_ENABLE=fuse_matmul and disable ATen evaluation through NVF_DISABLE=matmul_expr_eval.
> 
> Would this option have an effect on both matmul and linear?

Yes, this impacts both linear and matmul since they use the same schedulers.

>If so, I would find it reasonable to unify the two kwarg `nv_enable` options in Thunder into one.

That makes sense, we can have `nv_enable_matmul` and the default value can be `False` for `matmul`, and `True` for `linear`. Open to any better naming scheme.

PR #3270, opened 2024-10-24
Add `enable_options` and `disable_options` to `fd.execute`

Priya2698:
This PR adds `enable_options` and `disable_options` to `fd.execute` to allow setting them through the python frontend in lieu of the environment variables.
This work will be used to allow enabling nvfuser matmul codegen from within Thunder.

Inspired by @jacobhinkle's PR #1905!

Tracking Issue: #3022


Priya2698:
!build

jacobhinkle:
Thanks for taking the initiative with this! We will also need to update FusionExecutorCache to be option-aware so that we don't re-use kernels compiled with different options.

Priya2698:
> Thanks for taking the initiative with this! We will also need to update FusionExecutorCache to be option-aware so that we don't re-use kernels compiled with different options. That was part of the idea with #2077: separating options (features) into ones that affect each level of caching or not.

Got it -- we can add that support separately, what do you think? That will allow development on https://github.com/NVIDIA/Fuser/issues/3022. IIUC, we are currently missing this for the environment variables based approach as well so we may be reusing kernels which used different env variables?

jacobhinkle:
> IIUC, we are currently missing this for the environment variables based approach as well so we may be reusing kernels which used different env variables?

I guess this is true: a process can change its env vars at any time and we will not recognize that change after the first time we look up an option. However that is off-label use. If we are providing a kwarg to `execute` then a user would probably assume it's fully supported.

I wonder if we could do something quickly like use a separate `FusionCache` for each set of options we receive in the frontend. If no options are provided we'd use the default. That could be done entirely in python.

Priya2698:
> I wonder if we could do something quickly like use a separate `FusionCache` for each set of options we receive in the frontend. If no options are provided we'd use the default. That could be done entirely in python.

What do you mean by using separate `FusionCache`?

jacobhinkle:
> . The options are passed in the `execute` method so they are not available when the fusion definition is initialized.

Hmmm. Yeah I guess you're right. `FusionCache` is not the right place to do this. Instead we might do it by having a separate `FusionExecutorCache` might work: https://github.com/NVIDIA/Fuser/blob/5db18de5d8dafc6a8f1e76909d73799753fe7a5e/csrc/python_frontend/fusion_definition.cpp#L387-L390
Here `scheds->auto_gen_schedules` is a `unique_ptr<FusionExecutorCache>`. This is fine as a default but when features are supplied, we should create a new cache and use it instead I think.

nvfuser/__init__.py
```diff
@@ -177,13 +185,21 @@ def execute(
...
+                )
+
             results = self._execute(
                 inputs,
                 device=device,
                 override_user_schedule=override_user_schedule,
                 capture_debug_output=capture_debug_output,
                 profile=profile,
+                _enable_options=_enable_options,
+                _disable_options=_disable_options,
```
Priya2698, line 202: As discussed offline, moving forward with renaming the kwargs to `_enable/_disable_options`. I added a warning about the possibility of kernel reuse so that we are aware of this when utilizing this feature.

Priya2698:
!build

nvfuser/__init__.py
```diff
@@ -120,6 +123,11 @@ def execute(
             profile (bool): Captures a CUPTI based profile of a fusion.
             save_repro_inputs (bool): Saves the inputs for last_repro_script() to
                 provide a provide a reproduction script.
+            _enable_options/_disable_options (list): NVFUSER_ENABLE/DISABLE options to use.
+                This is an alternative to enviroment variables.
```
rdspring1, line 127:
```suggestion
                This is an alternative to environment variables.
```

Priya2698:
!build

jacobhinkle (approved):
LGTM. Thanks for making it clear that this is a somewhat unsafe option at least until we figure out the caching problem.

Priya2698:
!test

Priya2698:
!test

Priya2698:
!test

PR #3271, opened 2024-10-25
check vectorization factor of shared memory consumers to avoid illegal vectorization size

liqiangxl:
**Issue** InnerOuter persistent scheduler uses shared memory to store persistent buffers, the data flow is `input in gmem ---> async copy to smem --> vectorized load to registers (smem consumers)`, the `-->` are simply `LoadStoreOp` and same vectorization factors of these two copies are used. [CI](https://nv/e2E/118278383) found a case where the shared memory persistent buffers have a data type of fp32 while the inputs are fp16 (when there are view ops, project to inputs is not used). The vectorization factor is set to 8 and caused 32 bytes vectorization when loading from shared memory to registers.

**Changes**:
(1)  Added code to handle the vectorization of smem consumers. Add an additional split if `smem --> regs` copy leads to vectorization larger than 16 bytes.
(2) Added a test

**Results**: Ensure vectorizations are <= 16 bytes.

**Following works**
See issue https://github.com/NVIDIA/Fuser/issues/3272


liqiangxl:
!build

liqiangxl:
!build

csrc/scheduler/reduction_utils.cpp
```diff
@@ -1009,5 +1003,30 @@ std::ostream& operator<<(std::ostream& os, ReductionType reduction_type) {
...
+  // When this happens, there are two additional optimizations should be done:
+  // (1) writing to smem should be vectorized.
+  // (2) when n_loads > 1, still has bank conflicts.
+  // See test SharedMemoryPersistentVectFactor.
+  for (auto tv : smem_consumers) {
+    NVF_ERROR(
+        tv->definition()->isA<LoadStoreOp>(),
+        "smem consumers should be LoadStoreOp. Got: ",
+        tv->definition()->toString());
+    auto innermost_extent = tv->axis(-1)->extent()->evaluate().as<int64_t>();
```
liqiangxl, line 1020: Will add a check to ensure `innermost_extent == vectorization factor set for cached input`

csrc/scheduler/reduction_utils.cpp
```diff
@@ -1009,5 +1003,30 @@ std::ostream& operator<<(std::ostream& os, ReductionType reduction_type) {
   return os;
 }
 
+void sharedMemoryConsumerVectorization(
+    std::vector<TensorView*>& smem_consumers) {
+  // Vectorization of smem consumers, they were created with cacheAfter().
```
naoyam, line 1008: This comment seems to largely overlap with the comment of the header file. Have some comment that's important for the user of this function in the header. The implementation file should be mostly about implementation details.

Looks like both the header file and this file only discusses implementation details. See my comment at the header file as well.

csrc/scheduler/reduction_utils.h
```diff
@@ -139,5 +136,16 @@ std::string toString(ReductionType reduction_type);
 ReductionType getReductionType(Fusion* fusion);
 ReductionType getReductionType(const std::vector<TensorView*>& reduction_tvs);
 
+// Vectorization of smem consumers, they were created with cacheAfter() and
```
naoyam, line 139: Please mention what this function does and how it should be used. All mentioned here are some implementation limitations, but it's unclear what the expected result is by using this function.

csrc/scheduler/reduction_utils.cpp
naoyam: Can you please elaborate? Why not added yet?

naoyam (commented):
The fix itself makes sense.

csrc/scheduler/reduction_utils.cpp
liqiangxl: added. I was waiting for the CI to complete, didn't want to re-start it due to a small change,

csrc/scheduler/reduction_utils.cpp
liqiangxl: Thanks for the suggestion. Revised and proposed to follow the same style in the future.
For declaration in header file:
1. Use Doxygen-style comments before the function declaration.
2. Focus on describing the interface - what the function does, parameters, return value.
For implemenation in cpp file:
1. explaining the implementation details focus on `why` if `what` is simple and clear.

liqiangxl:
!build

naoyam (approved):
LGTM

liqiangxl:
!build

liqiangxl:
!build

liqiangxl:
Revised to ensure the correct axis is used.
```
    // non-concretized broadcast domains are moved to the innermost before
    // transform propagation, should skip these axes.
    int64_t vect_axis_pos = -1;
    while (tv->axis(vect_axis_pos)->isBroadcast()) {
      vect_axis_pos --;
      NVF_ERROR(
          vect_axis_pos + tv->nDims() >= 0,
          "Out of bound access when visiting dim ",
          vect_axis_pos,
          " in Tv: ",
          tv->toString());
    }
```

liqiangxl:
!build

Issue #3273, opened 2024-10-25
PTX code for async copy of bool type is not correctly generated


liqiangxl:
Test `DistributedTransformerTest.Backward/__bfloat` has bool type tensor, when shared memory persistent is used with async copy, it triggers a bug. A mini-reproduce is as follows:
```
TEST_F(NVFuserTest, CpAsyncDataTypeBool) {
  Fusion fusion;
  FusionGuard fg(&fusion);
  auto dtype = DataType::Bool;
  int m = 33, n = 128;
  auto tv0 = makeContigConcreteTensor({m, n}, dtype);
  fusion.addInput(tv0);
  auto tv1 = set(tv0);
  tv1->setMemoryType(MemoryType::Shared);
  tv1->definition()->as<LoadStoreOp>()->setOpType(
      LoadStoreOpType::CpAsync);
  tv1->definition()->as<LoadStoreOp>()->setCacheOp(CacheOp::Unspecified);
  auto tv2 = castOp(DataType::Int32, tv1);
  fusion.addOutput(tv2);
  
  for(auto tv : {tv0, tv1, tv2}){
    tv->split(1, 4);
  }
  for(auto tv : {tv0, tv1, tv2}){
    tv->axis(0)->parallelize(ParallelType::BIDx);
    tv->axis(1)->parallelize(ParallelType::TIDx);
  }
  tv1->axis(2)->parallelize(ParallelType::Vectorize);

  inlineMost();

  auto at_dtype = data_type_to_aten(dtype);
  auto options = at::TensorOptions().dtype(at_dtype).device(at::kCUDA, 0);
  // randn deosn't support bool type
  at::Tensor t0 = at::ones({m, n}, options);

  FusionExecutor fe;

  // requires ampere+ GPU
  if (!deviceMajorMinorCheck(8)) {
    ASSERT_THAT(
        [&]() { fe.compileFusion(&fusion, {t0}); },
        testing::ThrowsMessage<nvfuser::nvfError>(testing::HasSubstr(
            "Reason: LoadStoreOpType::CpAsync requires Ampere")));
    GTEST_SKIP() << "skipping tests on pre-AMPERE GPUs";
  } else {
    fe.compileFusion(&fusion, {t0});
  }
  fe.compileFusion(&fusion, {t0});
  auto cg_outputs = fe.runFusion({t0});

  testValidate(&fusion, cg_outputs, {t0}, __LINE__, __FILE__);
}
```

The generated code is:
```
[...]

PR #3274, opened 2024-10-25
avoid treating pointer to bool as bool when handling kir::asm in codegen

liqiangxl:
**Issue:**  https://github.com/NVIDIA/Fuser/issues/3273
The original failure happened in test `DistributedTransformerTest.Backward/__bfloat` when shared memory persistent is used with async copy (after https://github.com/NVIDIA/Fuser/pull/3217, not merged yet). The reason is pointer to bool was treated as bool when handling kir::asm in codegen 

**Fix:** If pointer, reture pointer type not the type it points to

**Results:** Added a unit test, error is fixed.

### DIFF
--- a/csrc/codegen.cpp
+++ b/csrc/codegen.cpp
@@ -3077,6 +3077,15 @@ class CudaKernelGenerator : private kir::ConstIrVisitor {
       indent();
     }
 
+    auto getTypeOrIndexType = [](Val* value) {
+      if (auto ti = dynamic_cast<kir::TensorIndex*>(value)) {
+        if (isPointerType(ti->index()->dtype())) {
+          return ti->index()->dtype();
+        }
+      }
+      return value->dtype();
+    };
+
     if (asm_->hasBooleanInput()) {
       code_ << "\"{\\n\"\n";
       int64_t boolean_counter = 0;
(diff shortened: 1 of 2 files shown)


liqiangxl:
!build

zasdfgbnm:
Could you elaborate more on the problem? Why only pointer of bool is a problem, not pointer to others?

liqiangxl:
> Could you elaborate more on the problem? Why only pointer of bool is a problem, not pointer to others?

Because `handle(const kir::Asm* asm_)` in codegen has special treatment for bool type, e.g.
```
          if (val->dtype() == DataType::Bool) {
            indent() << "\"  .reg .pred p" << boolean_counter << "; \\n\"\n";
            indent() << "\"  setp.ne.b32 p" << boolean_counter << ", %"
                     << counter << ", 0;\\n\"\n";
            boolean_counter++;
          }
```
and convert to uint32, e.g. 
```
             if (register_->dtype() == DataType::Bool) {
                code_ << "(uint32_t)(";
              }
```

Before Fix:
```
  asm volatile(
    "{\n"
    "  .reg .pred p0; \n"
    "  setp.ne.b32 p0, %0, 0;\n"
    "  .reg .pred p1; \n"
    "  setp.ne.b32 p1, %1, 0;\n"
    "  .reg .pred p2; \n"
    "  setp.ne.b32 p2, %3, 0;\n"
    "  cp.async.ca.shared.global [%0], [%1], %2, p0;\n"
    "}\n"
    :
    :"r"((uint32_t)((uint32_t)((toSmem(T1) + i0)))),
     "l"((uint32_t)(((T0.data + i0) + i1))),
     "n"(4LL),
     "r"((uint32_t)((!b3)))
  );
```
After fix:
```
  asm volatile(
    "{\n"
    "  .reg .pred p0; \n"
    "  setp.ne.b32 p0, %3, 0;\n"
    "  cp.async.ca.shared.global [%0], [%1], %2, p0;\n"
    "}\n"
    :
    :"r"((uint32_t)((toSmem(T1) + i0))),
     "l"(((T0.data + i0) + i1)),
     "n"(4LL),
     "r"((uint32_t)((!b3)))
  );
```

zasdfgbnm:
> Because `handle(const kir::Asm* asm_)` in codegen has special treatment for bool type

Oh, I see. I thought it was a dispatch on all dtypes. Didn't realize that it is only for bool.

zasdfgbnm (approved):

PR #3280, opened 2024-10-25
Export DEBUG_SERDE=true in compare_codegen.sh

jacobhinkle:
I am trying to fix some false positives in the codediff script


jacobhinkle:
!build --diff

PR #3284, opened 2024-10-25
Allgather with DID loop split

wujingyue:
Another baby step towards #2563


csrc/multidevice/utils.cpp
```diff
@@ -121,48 +121,77 @@ bool isSharded(const TensorView* tv) {
...
-    const TensorView* tv,
-    c10::IntArrayRef sizes) {
-  std::vector<int64_t> unsharded_sizes = sizes.vec();
-
-  for (IterDomain* alloc_id : tv->getMaybeAllocationDomain()) {
-    const ParallelType parallel_type = alloc_id->getParallelType();
+namespace {
+// Collect device-parallel IterDomains in `domain` and return them as a
+// ParallelType-to-IterDomain map.
+std::unordered_map<ParallelType, IterDomain*> mapDeviceParallelTypeToId(
```
wujingyue, line 127: Code motion and rename

csrc/multidevice/utils.cpp
```diff
@@ -121,48 +121,77 @@ bool isSharded(const TensorView* tv) {
...
+        "Found multiple loop IterDomains with the same parallel type (",
+        parallel_type,
+        "): ",
+        toDelimitedString(domain));
+  }
+  return parallel_type_to_id;
+}
+} // namespace
+
+int64_t getShardedAxis(const TensorView* tv, const ParallelType parallel_type) {
```
wujingyue, line 148: This is fixed to traverse from allocation back to logical.

csrc/multidevice/utils.cpp
```diff
@@ -121,48 +121,77 @@ bool isSharded(const TensorView* tv) {
...
+
+  // Count the number of non-reduction IterDomains before `iter`. Reduction
+  // IterDomains are not materialized in the at::Tensor's shape.
+  return std::count_if(
+      tv->getLogicalDomain().begin(), iter, [](IterDomain* id) -> bool {
+        return !id->isReduction();
+      });
+}
+
+std::vector<int64_t> unshardedSizes(
```
wujingyue, line 184: This is refactored to reuse getShardedAxis.

tests/cpp/multidevice.cpp
```diff
@@ -144,13 +145,7 @@ at::Tensor MultiDeviceTest::shardTensor(
   auto stride = extent / nslices;
   // TODO: returning slice 0 temporarily when device is not in the mesh.
   i = (i < 0) ? 0 : i;
-  auto slice = tensor.slice(axis, i * stride, (i + 1) * stride).contiguous();
-  // Temporary until https://github.com/NVIDIA/Fuser/issues/2563. Adds DIDx
-  // axis in front representing the sharded extent of the tensor.
```
wujingyue, line 149: Because loop-split cases don't need the unsqueeze, I moved the unsqueeze to the users, all of which are in distributed transformer tests. I don't feel the extra typing too bothersome, but I could make `shardTensor(...).unsqueeze(0)` a separate helper with a different name.

wujingyue:
!test

wujingyue:
!test

csrc/multidevice/utils.cpp
```diff
@@ -121,48 +121,77 @@ bool isSharded(const TensorView* tv) {
...
+  if (alloc_id == nullptr) {
+    return -1;
   }
 
+  const auto inputs = IterVisitor::getInputsTo(
+      {alloc_id},
+      {tv->getLogicalDomain().begin(), tv->getLogicalDomain().end()});
+  NVF_ERROR(
+      !inputs.empty(),
+      "IterVisitor::getInputsTo shouldn't return empty unless `of` is empty.");
```
samnordmann, line 161: what do you mean by `of` here?

samnordmann:
```suggestion
      "IterDomain ", alloc_id, " should be produced from the tv's logical domain ", tv->getLogicalDomain());
```

csrc/multidevice/utils.cpp
```diff
@@ -121,48 +121,77 @@ bool isSharded(const TensorView* tv) {
...
+  NVF_ERROR(
+      inputs.size() == 1,
+      "Failed to find the single logical input to ",
+      alloc_id,
+      ". This is likely because there's a Merge expression from logical to allocation, which isn't supported. Inputs are: ",
+      toDelimitedString(inputs));
+
+  const auto iter = std::find(
+      tv->getLogicalDomain().begin(), tv->getLogicalDomain().end(), inputs[0]);
+  NVF_ERROR(
```
samnordmann, line 171: I am not sure to understand why this check is needed. Isn't it true that by assumption what is returned by getInputsTo is an element of tv->getLogicalDomain()?

samnordmann: Also I am not sure what is meant by "dominate" in the error message

csrc/multidevice/utils.h
```diff
@@ -123,9 +123,9 @@ int64_t requestedNumberOfDevices(Fusion*);
 void unshard(Fusion*);
 void unshard(TensorView*);
 
-// Returns the index of the a sharded axis if none return -1.
-// TODO: Assumes no merges/splits on sharded axis.
-int64_t getShardedAxis(TensorView*);
+// Returns the index of the sharded logical axis corresponding to
+// `parallel_type`. If `tv` isn't sharded on the parallel type, returns -1.
```
samnordmann, line 127: IIUC, when sharding a split and/or transposed loop's domain axis, then this function returns the (unique) logical domain's axis that produces that axis.

I find the comment misleading here, and would suggest to make it more explicit, e.g,
```suggestion
// Returns the index of the logical axis which is the producer of the loop axis sharded with
// `parallel_type`. If `tv` isn't sharded on the parallel type, returns -1.
```

In general, I am skeptical about how this function can be conveniently used without returning also the information on how to transform this logical axis to the logical sharded axis...

csrc/multidevice/communication.cpp
```diff
@@ -325,8 +325,8 @@ c10::intrusive_ptr<c10d::Work> postAllgather(
     c10d::Backend* backend,
     at::Tensor input_tensor,
     at::Tensor output_tensor) {
-  auto splits = at::split(output_tensor, /*split_size=*/1, /*dim=*/0);
-  assertBufferCount(splits, communication->team().size());
+  auto splits =
```
samnordmann, line 328: what is the motivation for this change?

csrc/multidevice/lower_communication.cpp
```diff
@@ -196,7 +196,7 @@ void lowerToReduceScatter(
     std::vector<Communication*>& comms) {
   const DeviceMesh& mesh = input_tv->getDeviceMesh();
   auto reduction_axis = output_tv->getReductionAxis().value();
-  auto scattered_axis = getShardedAxis(output_tv);
+  auto scattered_axis = getShardedAxis(output_tv, ParallelType::DIDx);
```
samnordmann, line 199: ok, however if the sharded dimension is split, then `scatted_axis` is not valid here, right?

tests/cpp/test_multidevice_lower_communication.cpp
```diff
@@ -202,6 +202,38 @@ TEST_F(LowerCollectiveTest, Allgather) {
...
+  in->setAllocationDomain(in->getLoopDomain(), true);
+
+  out->split(0, num_devices, /*inner_split=*/false);
+  out->setAllocationDomain(out->getLoopDomain(), true);
+
+  at::Tensor unsharded_tensor =
+      at::randn({num_devices * kTensorSize}, at::kFloat);
+  at::Tensor in_tensor =
+      shardTensor(unsharded_tensor, in).to(communicator_->device());
+
```
samnordmann, line 229:
```suggestion

std::vector<int64_t> ref_int_tensor_shape = {kTensorSize};
EXPECT_EQ(in_tensor.sizes(), ref_int_tensor_shape);

```

samnordmann: I don't see how shardTensor is correct here if it never play the split backwards... But I might be missing something.

samnordmann (commented):
Thank you for this pr! I am still trying to fully understand how the logical works, but let me post a series of minor comments in the meantime.

tests/cpp/multidevice.cpp
cowanmeg: Sounds good! This will also help flag tests that will eventually be cleaned up.

tests/cpp/test_multidevice_lower_communication.cpp
```diff
@@ -202,6 +202,38 @@ TEST_F(LowerCollectiveTest, Allgather) {
...
+  FusionExecutorCache fec(std::move(fusion));
+  at::Tensor out_tensor = fec.runFusionWithInputs({in_tensor})[0];
+  assertIsCompiledToHostIrContainer(fec);
+
+  EXPECT_TRUE(at::equal(out_tensor.cpu(), unsharded_tensor));
```
cowanmeg, line 234: Why not use `validate` here?

cowanmeg: I noticed allgather's lowering was not changed...I'm a bit surprised it didn't need any modifications for inputs with DID loop split! I might have missed a few earlier PRs though

naoyam: > Why not use `validate` here?

Since `validate` allows for (small) differences, if two tensors are supposed to be exactly the same, just using the simpler validation method, i.e., `at::equal`, would be more preferable.

wujingyue: > I'm a bit surprised it didn't need any modifications for inputs with DID loop split!

Whether we call lowerToAllGather depends on I/O meshes and whether I/O is sharded: https://github.com/NVIDIA/Fuser/blob/67127c94a43e23f673cab05f769269f38e4fb043/csrc/multidevice/lower_communication.cpp#L285. isSharded have been reading the allocation domain ince #3444. 

That being said, I think this PR as is is a bit too permissive and may lower a `set` to `Allgather` without properly checking its allocation domain. For example, https://github.com/NVIDIA/Fuser/blob/67127c94a43e23f673cab05f769269f38e4fb043/csrc/multidevice/utils.cpp#L77 reads root and logical and needs to be updated. I'll try to fix that.

samnordmann:
> Thanks for the review! I think there are two problems with the PR as is:
> 
> 1. `shardTensor` may slice wrong numbers. For example, if an inner split is DID'ed, the slicing needs to be strided per the outer split.
> 2. nvFuser doesn't error out when Allgather is not along the outermost allocated dimension. This was guaranteed by ReorderShardedAxisPass by checking isInnerResharding. However, getShardingChanges, one of its dependencies, hasn't been updated to read loop/allocation: https://github.com/NVIDIA/Fuser/blob/67127c94a43e23f673cab05f769269f38e4fb043/csrc/multidevice/utils.cpp#L77


I see, thanks for the clarifications. Is it still ready for review or would you like to fix the points you mentioned above?

On my side, I understand why the PR works as is, but I am a little bit concerned because, in my opinion, it relies on a weak contract. To be more precise, the tests only work here because, when splitting the innermost loop axis and sharding one split, the allgather will still operate on innermost dimension (index 0), which has the same index as the logical axis used to produce that loop axis. The fact that this works feels a bit incidental and not robust, in particular it should fail for any variants like, merging axis, transposition (even on non-sharded axis), multidimensional Parallelization (e.g. DIDy), etc. Wdyt?

wujingyue:
> Is it still ready for review or would you like to fix the points you mentioned above?

I'll try to fix these points. I'd like nvFuser to at least error out on cases it doesn't support. This way, when we trigger a known limitation in the future, it'll show up as an exception instead of silently generating wrong numbers.

csrc/multidevice/utils.cpp
wujingyue: Re dominate: https://en.wikipedia.org/wiki/Dominator_(graph_theory) and I extended the concept to a set of nodes dominating another set. 

Re the check: I heard from @naoyam that logical won't always dominate allocation with "the new indexing system".

csrc/multidevice/utils.cpp
wujingyue: Code replaced.

csrc/multidevice/communication.cpp
wujingyue: split takes the size of each split and tensor_split takes the number of splits. Previously, it was easier to use split because the split size must be 1. With this PR, it's easier to use tensor_split because the split size is not necessarily 1 but the number of splits must be the team size.

csrc/multidevice/lower_communication.cpp
wujingyue: I can't think of an immediate problem and https://github.com/NVIDIA/Fuser/pull/3504 apparently works fine. Could be incidental and I'm happy to hear what you think is problematic.

tests/cpp/test_multidevice_lower_communication.cpp
wujingyue: Re the suggested change: I manually checked the shape is as expected. I added some extra unit tests for shardTensor alone, so we don't have to verify it here.

wujingyue: I made a couple of changes to address the problems I said in https://github.com/NVIDIA/Fuser/pull/3284#discussion_r1868843000. 
1. https://github.com/NVIDIA/Fuser/pull/3284/commits/7cf238402fdbd0656f26577088128418e7433f6d. It's an overkill but will probably be OK for quite some time. I had a hard time finding a concrete use case that has to mix DID and host ID within one logical dimension. I agree that to properly support inner splits we'll need to "replay the split backwards". It's not a trivial change anyhow so I'll postpone it to a separate PR. 
2. I wrote https://github.com/NVIDIA/Fuser/pull/3531 to harden runtime checks for allgather and added to this PR one more allgather test (Allgather_LoopSplit_Noncontiguous). These extra checks will fire when we trigger some most common limitations before properly fixing ReorderShardedAxisPass, which will take several decent-size PRs.

tests/cpp/test_multidevice_lower_communication.cpp
wujingyue: > That being said, I think this PR as is is a bit too permissive and may lower a set to Allgather without properly checking its allocation domain.

I tried to address this in https://github.com/NVIDIA/Fuser/pull/3284#discussion_r1870826379.

csrc/multidevice/utils.cpp
```diff
@@ -121,48 +121,103 @@ bool isSharded(const TensorView* tv) {
...
-    const auto index = std::count_if(
-        tv->getLogicalDomain().begin(), iter, [](IterDomain* id) -> bool {
-          return !id->isReduction();
-        });
-    unsharded_sizes.at(index) *= tv->getDeviceMesh().size(parallel_type);
+        def != nullptr,
+        "Failed to find a non-reduction logical IterDomain that produces ",
+        alloc_id);
+    if (auto* split = dynamic_cast<Split*>(def)) {
+      // FIXME: comment
```
wujingyue, line 182: Note to myself: explain the problem with inner splits before merging the PR.

tests/cpp/test_multidevice_lower_communication.cpp
wujingyue: > I had a hard time finding a concrete use case that has to mix DID and host ID within one logical dimension. 

In fact, there's https://github.com/NVIDIA/Fuser/blob/64bc5604f5b14a4052bb579426133c028662a934/tests/cpp/test_multidevice_overlap.cpp#L681. So I'll try to file a feature request after this PR.

csrc/multidevice/utils.h
```diff
@@ -123,9 +123,10 @@ int64_t requestedNumberOfDevices(Fusion*);
 void unshard(Fusion*);
 void unshard(TensorView*);
 
-// Returns the index of the a sharded axis if none return -1.
-// TODO: Assumes no merges/splits on sharded axis.
-int64_t getShardedAxis(TensorView*);
+// Returns the index of the sharded logical axis that produces the allocation
+// IterDomain sharded on `parallel_type`. If `tv` isn't sharded on the parallel
+// type, returns -1.
+int64_t getShardedAxis(const TensorView* tv, ParallelType parallel_type);
```
naoyam, line 129: nit:
```suggestion
int64_t getShardedLogicalAxis(const TensorView* tv, ParallelType parallel_type);
```

naoyam: Can you give some more details about what this function is supposed to return? As mentioned in the implementation, there's no single logical ID if the parallelized ID is a merge output. I don't have a good overall picture, so it's unclear to me what the context is.

samnordmann: +1
I am also unsure about how this function is supposed to be used.

csrc/multidevice/lower_communication.cpp
wujingyue: In your case, 
```
tv1:
  logical: [r{i0}, i{i1}]
  allocation: [r{i0}, i{i1/d}, iDID{d}]
```

getShardedLogicalAxis will return 0, the tensor axis being sharded. This is correct because the output at::Tensor for will be of shape [i1/d] and indeed axis 0 is the sharded dimension. 

Caveat: With https://github.com/NVIDIA/Fuser/commit/7cf238402fdbd0656f26577088128418e7433f6d, DID'ing an inner split is disallowed by code. So the above case will actually throw an exception. But what I said should be correct after we lift that limitation.

csrc/multidevice/utils.h
wujingyue: Done!

csrc/multidevice/utils.cpp
```diff
@@ -179,7 +181,35 @@ int64_t getShardedAxis(const TensorView* tv, const ParallelType parallel_type) {
...
+      //   t->split(0, 3);
+      //   t->axis(0)->parallelize(Host);
+      //   t->axis(1)->parallelize(DIDx);
+      //   // [iHost{2}, iDIDx{3}, i{5}]
+      //
+      // Given an unsharded at::Tensor of shape [30], we'll first replay the
+      // splits using `torch.view` to get a tensor of shape [2,3,5]. Then, we
+      // `torch.slice` axis 1 for DIDx to get a tensor of shape [2,1,5]. Then,
+      // we rewind the splits (and therefore apply merging) using
+      // `torch.reshape` to get a sharded tensor of shape [10].
```
naoyam, line 212: Makes sense. Erroring out also seems reasonable.

naoyam (approved):
LGTM

csrc/multidevice/utils.cpp
wujingyue: FWIW, as said in https://github.com/NVIDIA/Fuser/pull/3284#discussion_r1871622953, there is a valid and concrete use case for this. So we'll fix it sooner or later.

wujingyue:
!test

PR #3288, opened 2024-10-27
fix padded bdimx to use warp reduction in inner reduction scheduler

liqiangxl:
Simple fix to padded bdimx in inner reduction scheduler.
**Performance changes:**
**(1) H100**
5 lines corresponds to batch size of 16, 512, 2048, 8192, 16384
image
**(2) A100**
image

### DIFF
--- a/csrc/scheduler/reduction.cpp
+++ b/csrc/scheduler/reduction.cpp
@@ -388,9 +388,6 @@ std::unique_ptr<ReductionParams> innerReductionHeuristic(
   bool pad_bdimx = bdimx > 16 &&
       bdimx * bdimy <
           (int64_t)at::cuda::getCurrentDeviceProperties()->maxThreadsPerBlock;
-  // If barely just covering reduction dim, don't pad to the next warp
-  pad_bdimx = pad_bdimx &&
-      bdimx * inner_reduction_unroll_factor != inner_most_dimension_numel;
   rparams->pad_inner_reduction_to_warp = pad_bdimx;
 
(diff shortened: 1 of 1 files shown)


liqiangxl:
!build

csrc/scheduler/reduction.cpp
```diff
@@ -388,9 +388,6 @@ std::unique_ptr<ReductionParams> innerReductionHeuristic(
   bool pad_bdimx = bdimx > 16 &&
       bdimx * bdimy <
           (int64_t)at::cuda::getCurrentDeviceProperties()->maxThreadsPerBlock;
-  // If barely just covering reduction dim, don't pad to the next warp
-  pad_bdimx = pad_bdimx &&
-      bdimx * inner_reduction_unroll_factor != inner_most_dimension_numel;
```
jjsjann123, line 393: the comment looks a bit strange here.

`bdimx * inner_reduction_unroll_factor != inner_most_dimension_numel;` == `barely just covering reduction dim`?

jjsjann123: nvm. I think the idea here is that we always want to pad to facilitate warp reduction, hence removing this... I'm curious why we choose not to pad earlier when the problem size is small.

jjsjann123 (approved):

csrc/scheduler/reduction.cpp
liqiangxl: I don't know why. Seems a history thing, maybe older generations of card, at least on A100 we still gains speedup.

PR #3289, opened 2024-10-27
Add information for coordinating segments in python frontend.

rdspring1:
# Overview
This PR adds information necessary for coordinating segments in the python frontend. Changes pulled from https://github.com/NVIDIA/Fuser/pull/3025.

## PR Details
* Track the fusion state ids for the inputs, outputs, and extents of a Fusion. Inputs and extents are used to gather tensor arguments and scalars to run a fusion segment, while the outputs are employed to store results between segments.
* A map from a CPP value to its corresponding fusion state id, which is needed to map values from original fusion to its segmented fusions.

## Implementation Details

- `FusionState` is a lightweight representation of a CPP `Fusion`. 
- When calling `buildFusionIr`, a CPP `Fusion` is created from the Python `FusionDefinition`. At this point, the `FusionState` creates a mapping from CPP `Fusion` to its `State` objects.
- However, the `FusionState` is temporary and the CPP `Fusion` is cached in `FusionCache`. The information linking the CPP `Fusion` and Python `FusionDefinition` is stored in `FusionCache`.
- When we create a new `FusionState`, we look for a cached CPP `Fusion`. If it exists, we restore the mapping from the data stored in `FusionSchedules`.


rdspring1:
!build

csrc/python_frontend/fusion_state.cpp
```diff
@@ -206,4 +232,60 @@ void FusionState::aliasOutputToInput(Val* output, Val* input) {
...
+
+  std::vector<Val*> extents;
+  for (Val* v : fusion->inputs()) {
+    // short-circuit: skip if not TensorView
+    if (!v->isA<TensorView>()) {
+      continue;
+    }
+    TensorView* tv = v->as<TensorView>();
+    std::vector<IterDomain*> logical_dom =
+        TensorDomain::noReductions(tv->getLogicalDomain());
```
jjsjann123, line 263: @wujingyue is trying to change how we bind IO buffers to kernels. i.e. we might rethink which domain and how we are going to use here.

Not proposing any change, just trying to raise awareness.

csrc/python_frontend/fusion_state.cpp
```diff
@@ -206,4 +232,60 @@ void FusionState::aliasOutputToInput(Val* output, Val* input) {
...
+void FusionState::addExtents() {
+  NVF_CHECK(fusion_ != nullptr, "Fusion is undefined.");
+
+  // The size of the tensor dimensions can be used as an input of the
+  // segments. NvFuser does not support returning scalar values. Segmentation
+  // must pass those sizes as segment arguments manually.
+  std::vector<Val*> extents = getExtents(fusion_);
+  for (Val* extent : extents) {
+    int64_t num_extents = (int64_t)extents_fid_.size();
+    int64_t extent_fid = -num_extents - 1;
```
jjsjann123, line 282: is this a negative index?

csrc/python_frontend/fusion_state.cpp
```diff
@@ -206,4 +232,60 @@ void FusionState::aliasOutputToInput(Val* output, Val* input) {
...
+    extents_fid_.push_back(extent_fid);
+    // The extent can already exist in the fusion. However, since scalars cannot
+    // be passed between segments, always overwrited existing fids. The original
+    // fusion definition will provide scalar extents.
+    map_value_to_fid_[extent] = extent_fid;
```
jjsjann123, line 287: I'm a bit lost here.

iiuc, the map_value_to_fid_ on other values are mapped from the Val* to their index field in `FusionState`. Here looks like we are trying to create a the same thing for each TensorView's logical domain. Where are we creating the python container for that?

rdspring1:
!build

jjsjann123 (commented):
sorry for the delayed review.

csrc/python_frontend/fusion_state.cpp
rdspring1: All scalars, vectors, and tensors use positive indices. The extents do not exist in the `FusionState`, so I used the negative numbers exclusively for the extent scalars.

The extents are the size of `iterDomain` in CPP fusion. We don't track those in `FusionDefinition` but they can become input arguments to a `FusionDefinition` after segmentation.

csrc/python_frontend/fusion_state.cpp
rdspring1: I'm not exposing the TensorView's logical domain to the python frontend, but I am tracking it in the `FusionState`. We may have to pass the scalar extents of the TensorView's logical domain as an input argument to a fusion segment.

Priya2698:
Is it possible to add a test demonstrating what new information the `FusionState` stores and its link to the `FusionCache`?

csrc/python_frontend/fusion_cache.h
```diff
@@ -107,6 +107,14 @@ struct FusionSchedules {
   std::mutex scheds_lock;
   //! ID of fusion in python frontend fusion cache
   int64_t fusion_id_ = -1;
+  //! Input arguments for FusionState
```
Priya2698, line 110: nit: 
```Suggestion
  //! Fusion IDs of input arguments for FusionState

Similarly for others

File: csrc/python_frontend/fusion_state.cpp (hunk 2 above)
Comment (jjsjann123) [reply csrc/python_frontend/fusion_state.cpp:282]: so the negative number here is just an initialization? does the number carry any meaning or does a global `-1` would do it just fine?
sorry I might miss the part where extents_fid_ is being used.

Comment (rdspring1) [reply csrc/python_frontend/fusion_state.cpp:282]: There is an ordering component to the extent index.

It is used for the same purpose as collecting the extents in `prepareRuntimeOrder`.
https://github.com/NVIDIA/Fuser/blob/main/csrc/runtime/fusion_cache_utils.cpp#L199-L208

We're mapping the tensor sizes to the extents like so https://github.com/NVIDIA/Fuser/pull/3025/files#diff-e512bea3b02f75ab1e81b759562879c5867e6e863679d6e7696fa34087dc3dc9R98-R100.

Comment (Priya2698) [reply csrc/python_frontend/fusion_state.cpp:282]: Can we add this in a comment listing the use of negative indices to avoid conflict with other indices.

Comment (rdspring1) [reply csrc/python_frontend/fusion_state.cpp:285]: Added comment.

File: tests/python/test_python_frontend.py [hunk 5]
```diff
@@ -4600,3 +4600,34 @@ def fusion_func(fd: FusionDefinition) -> None:
...
+
+        with FusionDefinition() as fd:
+            fusion_func(fd)
+
+        nvf_out1 = fd.execute(inputs)
+        self.assertEqual(eager_out, nvf_out1[0])
+
+        self.assertEqual(fd.inputs(), [0, 1])
+        self.assertEqual(fd.outputs(), [5])
+        self.assertEqual(fd.extents(), [idx for idx in range(-1, -7, -1)])
```
Priya2698, line 4633: Can you add a comment mapping the IDs to the variables in fusion_func -- it will be easier to follow?

Priya2698:
LGTM overall. I'll let @jjsjann123 stamp it.
Do you have a document describing the intended flow of information between `FusionCache`, `FusionState`, `FusionSchedules`, `FusionDefinition` that will be built for Issue #3025?

Priya2698 (approved):

csrc/python_frontend/fusion_state.h
```diff
@@ -79,12 +79,24 @@ class FusionState {
...
+  NVF_API const std::unordered_map<const Val*, int64_t>& getValueMap() const;
+  //! Get indicies for the inputs of FusionState
+  NVF_API const std::vector<int64_t>& inputs() const;
+  //! Get indicies for the outputs of FusionState
+  NVF_API const std::vector<int64_t>& outputs() const;
+  //! Get indicies for the extents of TensorView inputs of FusionState
+  NVF_API const std::vector<int64_t>& extents() const;
+
+  //! Add extents of TensorView inputs to FusionState
+  NVF_API void addExtents();
```
jjsjann123, line 98: nitpick: does this need to be a public method? looks like it's only used in `FusionState::buildFusionIr`

csrc/python_frontend/fusion_state.cpp
jjsjann123: Got'ya. It's hard to figure out the necessity without looking at the actual use. We can keep it as-is and revisit in follow up PRs.

jjsjann123 (approved):
:shipit:

rdspring1:
!build

rdspring1:
## Summary
* A `FusionDefinition` holds a series of `RecordFunctors`. When building the `FusionDefinition`, we traverse the `Trie` in the `FusionCache`. Upon reaching the `EndRecord`, create the CPP `Fusion` using the `RecordFunctors`. Also, generate the mappings from `Scalar`, `Vector`, and `Tensor` states to CPP `Val`.

* Since the `FusionDefinition` is temporary, store any information in the `FusionSchedules` associated with this `FusionDefinition`. The `FusionSchedules` is associated with the `EndRecord` leaf in the `FusionCache`.

* Now, say we have a new `FusionDefinition` python object with the same definition. When building the `FusionDefinition`, traverse the `Trie` in the `FusionCache` again. The CPP `Fusion` already exists, so we load the CPP `Fusion` to the `FusionDefinition` along with the stored information in `FusionSchedules`.

## Information Flow
* `FusionCache` holds `FusionSchedules` in the leaves of Trie.
* `FusionSchedules` holds `Fusion` and other information.
* `FusionDefinition` creates CPP Fusion associated with `FusionCache`.
* `FusionDefinition` contains CPP `Fusion` and mappings from python frontend and CPP `Fusion`.
* `FusionState` is the parent class of `FusionDefinition`.

@Priya2698

PR #3294, opened 2024-10-28
Only the TMA thread arrive

zasdfgbnm:
Previously:
```C++
if (elect-sync) {
  arriveExpectTx
  TMA
} else {
  arrive
}
```
Now:
```C++
if (elect-sync) {
  arriveExpectTx
  TMA
}
```

I am very surprised that this fixes all the latencies introduced in the elect-sync fix https://github.com/NVIDIA/Fuser/pull/3295, and even better! But in general, we should sync as less as possible, and avoid unnecessary wait, so I think this PR makes sense.

Perf:
```C++
 Time (%)  Total Time (ns)  Instances  Avg (ns)  Med (ns)  Min (ns)  Max (ns)  StdDev (ns)
 Name
     39.0           172735          1  172735.0  172735.0    172735    172735          0.0  <unnamed>::nvfuser_none_f0_c0_r0_g0(<unnamed>::Tensor<<unnamed>::__half, (int)3, (int)3>, <unnamed>…
     20.0            88768          1   88768.0   88768.0     88768     88768          0.0  nvjet_hsh_256x128_64x4_1x2_h_bz_coopA_NTT
```

Perf nvFuser/cuBLAS: `51.4%`.

### DIFF
--- a/csrc/device_lower/pass/allocation.cpp
+++ b/csrc/device_lower/pass/allocation.cpp
@@ -67,25 +67,6 @@ Expr* initializeMbarrier(
   kir::TensorIndex* stage_mbarrier =
       IrBuilder::create<kir::TensorIndex>(all_mbarriers, loop->index());
 
-  // Get all threads in CTA
-  Val* bdimx =
-      GpuLower::current()->parallelDimensionMap().get(ParallelType::TIDx);
-  Val* bdimy =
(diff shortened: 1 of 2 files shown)


zasdfgbnm:
!build

naoyam (approved):
LGTM

rdspring1 (approved):

PR #3300, opened 2024-10-29
Compile forward function in baseline backward benchmarks

Priya2698:
Currently, the backward benchmark for `torch.compile` are compiling the `output.backward(grads)` function.
This PR changes the benchmarks to compile the corresponding forward function. 

The benchmarking results are at: http://nv/emR
Earlier, I was observing similar performance for eager and torchcompile for these backward benchmarks.
The files `*_bwd.py` are changed in this PR. We observe improved performance with torchcompile as compared to eager now. Note, that, the results from weekly will have more data points than the CI pipeline.

**Question**: `test_gelu_bwd_reduction.py` reduces the gradient of the input. I am not sure what would be the right function to compile here :/

Thanks to @xwang233 for help with running the CI pipeline!

### DIFF
--- a/benchmarks/python/normalization.py
+++ b/benchmarks/python/normalization.py
@@ -489,12 +489,15 @@ def norm_bwd_baseline_benchmark(
         grads = grads.to(memory_format=torch.channels_last)
 
     norm_fwd_fn = batchnorm_fwd_fn if norm == "batch_norm" else instancenorm_fwd_fn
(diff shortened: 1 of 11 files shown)


xwang233:
!build

Priya2698:
!build

liqiangxl (approved):
LGTM.

PR #3303, opened 2024-10-29
Change where we print python repros to allow us to print repros prior to segfaults

kevinstephano:
This is just a simple code reordering to put the point at which I print a repro from `fd.execute()` after the pybind function `_execute()` to before the function to allow us to print repros prior to segfaults.


kevinstephano:
!build

kevinstephano:
!build

kevinstephano:
!build

PR #3304, opened 2024-10-29
Disable nvfusertest_serde_check if DEBUG_SERDE=disable

### LINKED ISSUE #3265: Python test "at-exit" serialization has effect on repeat testing runs
Recently we noticed that codediff in CI was showing diffs for branches that had trivial changes. The reason turned out to be that running the python tests a second time on the same machine can give slightly different results. This is because of the default saving of FusionCache to a temporary file upon program exit. This is documented here:
https://github.com/NVIDIA/Fuser/blob/85c22a2af3fc756d9b301c93ac29b2bba8abf2af/tests/python/utils.py#L275-L287

When the test is run a second time, the serialized FusionCache is loaded, after which some FusionDefinitions will become cache hits whereas they were cache misses on the first invocation of the test suite. I worked around this by just backing up the fusion cache and restoring it in #3260, but setting `DEBUG_SERDE=true` is an alternative.

I have not observed this yet but I am slightly worried that we might hit this during development. If we are making changes to the backend and we're running a python test repeatedly, is it possible that we will load the stale version of the fusion instead of the new one? Is `DEBUG_SERDE` the best interface: should we rename it, or add this as a pytest cmdline arg? Also, it's important to test this functionality which is fundamental to serde, but should we disable it by default and enable it only for the CI jobs? I'm opening this issue for discussion. cc @rdspring1

### LINKED ISSUE #3283: Non-deterministic codegen with some of the Python tests
The diff test started to keep failing with some of the Python tests. For example:

https://dl.gitlab-master-pages.nvidia.com/pytorch/fuser-gh-mirror//nvfuser_github_ci/codegen_diff_p19742638_j118568737_1729887841468788908_codediff_896a28ad_b9203e1c_custom_command_20241025_131551.html

Ran `NVFUSER_DUMP=cuda_to_file pytest -v -k 'test_prim_layer_norm_fwd` 10 times, and here are the number of kernels generated per each run:

```
6
5
6
4
6
5
5
6
5
4
```

CC: @jacobhinkle, @xwang233 
Related: #3260 #3280 #3256

jacobhinkle:
This is another attempt to fix the codediff CI job

Fixes #3265. Fixes #3283.

### DIFF
--- a/tests/python/utils.py
+++ b/tests/python/utils.py
@@ -288,9 +288,10 @@ def check_cpp_translation(reference_outputs, fd, inputs, device=None):
 # DEBUG_SERDE=disable
 # 1) It disables the @nvfusertest_serde_check decorator. This disables checking
 # that serde round-trips preserve the definition during testing.
-env_var_debug_serde = os.getenv("DEBUG_SERDE").lower()
+env_var_debug_serde = os.getenv("DEBUG_SERDE", "").lower()
 debug_serde: bool = env_var_debug_serde == "true"
 disable_serde: bool = env_var_debug_serde == "disable"
+del env_var_debug_serde
 
 
 # The pytest framework and test_python_frontend.py use different arguments for
--- a/tests/python/utils.py
+++ b/tests/python/utils.py
@@ -289,7 +289,7 @@ def check_cpp_translation(reference_outputs, fd, inputs, device=None):
 # 1) It disables the @nvfusertest_serde_check decorator. This disables checking
 # that serde round-trips preserve the definition during testing.
 env_var_debug_serde = os.getenv("DEBUG_SERDE", "").lower()
-debug_serde: bool = env_var_debug_serde == "true"
+debug_serde: bool = env_var_debug_serde == "debug"
 disable_serde: bool = env_var_debug_serde == "disable"
 del env_var_debug_serde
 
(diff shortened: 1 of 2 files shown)


jacobhinkle:
!build --diff

rdspring1:
I would disable serde if `DEBUG_SERDE=disable`. It seems clearer this way.

```python
env_var_debug_serde = os.getenv("DEBUG_SERDE")
debug_serde: bool = env_var_debug_serde == "debug"
disable_serde: bool = env_var_debug_serde == "disable"
```

jacobhinkle:
!build --diff

jacobhinkle:
!build --diff

jacobhinkle:
!build --diff

rdspring1 (approved):
LGTM. Thanks!

tests/python/utils.py
```diff
@@ -283,8 +284,14 @@ def check_cpp_translation(reference_outputs, fd, inputs, device=None):
...
-env_var_debug_serde = os.getenv("DEBUG_SERDE")
-debug_serde: bool = env_var_debug_serde in ("true", "1")
+#
+# DEBUG_SERDE=disable
+# 1) It disables the @nvfusertest_serde_check decorator. This disables checking
+# that serde round-trips preserve the definition during testing.
+env_var_debug_serde = os.getenv("DEBUG_SERDE", "").lower()
+debug_serde: bool = env_var_debug_serde == "debug"
+disable_serde: bool = env_var_debug_serde == "disable"
+del env_var_debug_serde
```
rdspring1, line 294: 👍🏼

jacobhinkle:
I _think_ the codediff false positives in this case are because we have disabled the serde check in this PR commit but not on merge-base. I am hoping that if we merge this then try again in another commit this might finally be addressed.

PR #3309, opened 2024-10-30
Use SimplifyingIrBuilder for split and merge extents

naoyam:
Currently, when we split an iter domain, we get `ceilDiv(i0, s)`, where `i0` is the extent of the input iter domain and `s` is the split factor. With this PR, it may no longer be always `ceilDiv` when it's simple enough that `SimplyfingIrBuilder` can understand. One common example is split by 1, which is often done for unswitch. We would just get `i0` instead of `ceilDiv(i0, 1)`. Same for `merge`.

This should most likely have no impact on actual kernel performances as many of them are simplified and hoisted anyway. However, `printMath` could be greatly simplified especially when `resize` is used with static shapes.

On the other hand, as mentioned https://github.com/NVIDIA/Fuser/pull/3309#discussion_r1824753545, we would not be able to infer what ID operations were used by just looking at the extent. I personally that wouldn't be too concerning, and looks like @jacobhinkle and @zasdfgbnm also have the same opinion.

### DIFF
--- a/csrc/device_lower/pass/replace_size.cpp
+++ b/csrc/device_lower/pass/replace_size.cpp
@@ -115,8 +115,6 @@ std::unordered_map<Val*, Val*> getSimplificationMap(Fusion* fusion) {
     }
     NVF_ERROR(rep != nullptr);
     Val* rep_ext = rep->extent();
-    NVF_ERROR(!ir_utils::isRecursivelyDefined(rep_ext));
-
     for (Val* v : *group) {
       auto* id = v->as<IterDomain>();
       Val* ext = id->extent();
@@ -126,7 +124,6 @@ std::unordered_map<Val*, Val*> getSimplificationMap(Fusion* fusion) {
       }
     }
   }
-
   return simplification_map;
 }
 
@@ -135,10 +132,6 @@ std::unordered_map<Val*, Val*> getSimplificationMap(Fusion* fusion) {
 void replaceSymbolicSizes(Fusion* fusion) {
   FUSER_PERF_SCOPE("GpuLower::Lower::replaceSymbolicSizes");
   std::unordered_map<Val*, Val*> tensor_dim_map;
-
-  // std::cout << "replaceSymbolicSizes\n";
-  // fusion->print();
-  // std::cout << std::endl;
 
   // Grab inputs and outputs
   std::vector<TensorView*> inputs_and_outputs;
--- a/csrc/device_lower/pass/replace_size.cpp
+++ b/csrc/device_lower/pass/replace_size.cpp
@@ -59,11 +59,14 @@ std::unordered_map<Val*, Val*> getSimplificationMap(Fusion* fusion) {
     // 1. Constant ints. These might be non-immediate constants
     // 2. Extents of input TVs.
     // 3. Extents of non-input TVs.
-    // Within these three classes, we find the IterDomain with the smallest
-    // name().
+    // Within these three classes, we find the IterDomain with the
+    // smallest name(). For case 3, we also prefer the IterDomain with
+    // the simplest extent, which has the smallest number of defining
+    // expessions.
     bool group_is_const = false;
     IterDomain* rep = nullptr;
     bool rep_is_input_id = false;
+    int64_t rep_num_defs = 0;
     std::unordered_set<Val*> dynamic_scalars;
     for (Val* v : *group) {
       auto* id = dynamic_cast<IterDomain*>(v);
@@ -81,6 +84,10 @@ std::unordered_map<Val*, Val*> getSimplificationMap(Fusion* fusion) {
         rep = id;
         rep_is_input_id = is_input_id;
         group_is_const = ext_is_const;
+        // If neigher const nor input, record the number of exprs
+        if (!ext_is_const && !is_input_id) {
+          rep_num_defs = ir_utils::getOperationCount(id->extent());
+        }
         continue;
       }
 
@@ -106,9 +113,11 @@ std::unordered_map<Val*, Val*> getSimplificationMap(Fusion* fusion) {
         if (group_is_const || rep_is_input_id) {
           continue;
         }
-        if (id->name() < rep->name()) {
+        auto num_defs = ir_utils::getOperationCount(id->extent());
+        if (num_defs < rep_num_defs || id->name() < rep->name()) {
           rep = id;
           rep_is_input_id = is_input_id;
+          rep_num_defs = num_defs;
           continue;
         }
       }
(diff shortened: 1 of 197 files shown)


naoyam:
!build

csrc/ir/nodes.cpp
```diff
@@ -2625,10 +2627,14 @@ std::pair<IterDomain*, IterDomain*> IterDomain::split(
       factor->isIntegralScalar(), "Cannot split by non-integer value ", factor);
 
   // outer loop size
-  Val* remainder = ceilDiv(in->extent(), factor);
+  Val* remainder = getenv("SPLIT")
+      ? SimplifyingIrBuilder::ceilDivExpr(in->extent(), factor)
```
naoyam, line 2631: I think I figured it out why this resulted in so many segfaults. Assuming there's no functional regression, is there any concern with this change? From my quick experiments, slice and concat with static shapes seem to be greatly simplified from the beginning, which is quite convenient for debugging.

However, it could be also considered less useful as you can't figure out how an iter domain is constructed by just looking at its extent. For example, `iS{ceilDiv(32, 4)}` would become just `iS{8}`. With the current representation, it's clear that it's an output of a size-32 domain split by a factor of 4, but that's not immediately visible just from the extent notation since it'd be just `8`. 

Overall, I'd prefer the simplification since it would also help avoid accumulating integer expressions around reshape. Wdyt? @zasdfgbnm @jacobhinkle

jacobhinkle: > I think I figured it out why this resulted in so many segfaults. Assuming there's no functional regression, is there any concern with this change? 

I don't think so. We shouldn't be pattern matching on the extent formula anywhere to my knowledge.
 
> Overall, I'd prefer the simplification since it would also help avoid accumulating integer expressions around static resize. Wdyt? @zasdfgbnm @jacobhinkle

+1 I'm for simplification. I agree somewhat about the utility in scanning through the fusion math to check for split/merge but that also sometimes makes it tough to even see how many dimensions there are in a scheduled tensor.

zasdfgbnm: In my experience, I often feel not simplifying could be a problem. For example, in my mind, I can clearly remember that I am scheduling as `[..., 2, ....]`, but I can not remember and I don't really care how I get this `2` (for example, `split(split(I0, 128).outer, 2).inner` and `split(split(I0, 256).inner, 128).outer` is mathematically equivalent, so I don't care which one we are using). When I am looking for the `2` in the loop domain, it is most convenient that it is just `2`, but sometimes it can be complicated as `ceilDiv(ceilDiv(128, 8), 8)`, which takes me a few seconds to figure out if it is `2` or not.

naoyam:
Part of the failures were because `SimplifyingIrBuilder` was only used in `IterDomain::split` and `IterDomain::merge`, but not in transform_rfactor and transform_replay, where `Split` and `Merge` are directly created without using the `IterDomain` functions. This meant that when we have a split of `i0` by 1, for example, sometimes it's simplified to just `i0` but not always. Since they are still exactly mapped, an exact group has something like `{i0, ceilDiv(i0, 1), ...}`. Suppose `i1 = ceilDiv(i0, 1)`, we get an equality relationship of `i0 == i1`, which is used in `replaceSymbolicSizes`.  In this case, it can result in a recursive definition as follows.  The definition of `i1`,

```
i1 = ceilDiv(i0, 1)
```

is going to be examined by the mutator. While we disable replacing outputs, the use of `i0` in `ceilDiv` may be replaced by `i1` since the replacement map may have `i0 -> i1`. As a result, the definition of `i1` becomes:

```
i1 = ceilDiv(i1, 1)
```

To avoid this, vals should never be replaced with dependent vals. We should probably add some assertions.

naoyam:
!test

naoyam:
Stacked on top of #3344, although nothing depends on #3344.

naoyam:
Forgot to update reference code for checking generated code. Will ping once done.

naoyam:
!test

naoyam:
!test

naoyam:
!test

csrc/device_lower/pass/replace_size.cpp
```diff
@@ -114,7 +114,8 @@ std::unordered_map<Val*, Val*> getSimplificationMap(Fusion* fusion) {
           continue;
         }
         auto num_defs = ir_utils::getOperationCount(id->extent());
-        if (num_defs < rep_num_defs || id->name() < rep->name()) {
+        if (num_defs < rep_num_defs ||
+            (num_defs == rep_num_defs && id->name() < rep->name())) {
```
naoyam, line 118: This is a bug fix.

csrc/ir/nodes.cpp
```diff
@@ -28,7 +28,6 @@
 #include <complex>
 #include <iterator>
 #include <numeric>
-#include <regex>
```
naoyam, line 31: unused

tests/cpp/test_gpu3.cpp
```diff
@@ -8943,11 +8943,14 @@ TEST_F(NVFuserTest, ReplaceSymbolicSizesPreferSimplerExtents) {
   fusion.addInput(tv0);
   auto tv1 = makeSymbolicTensor(2);
   fusion.addInput(tv1);
-  auto i0 = IrBuilder::create<Val>(DataType::Index);
-  fusion.addInput(i0);
 
-  auto tv2 = reshape(tv0, {i0});
-  auto tv3 = reshape(tv1, {i0});
+  auto tv2 = reshape(
```
naoyam, line 8947: This happened to work due to the bug in replaceSymbolicSizes, but it should have been written this way from the beginning.

tests/cpp/test_indexing.cpp
naoyam, line 1: All the changes in this file are due to the simplification by `SimplifyingIrBuilder`.

csrc/index_compute.cpp
```diff
@@ -753,9 +753,8 @@ class UpdateLeafIndices : public IterVisitor {
     // Nothing need to be done when mappings for the output axes
     // already exist.
     if (index_map_.find(outer_id) != index_map_.end()) {
-      NVF_ERROR(
```
naoyam, line 756: This condition is not guaranteed to hold, and it gets exposed with more AlmostExact mappings. The code itself is considered (almost) obsolete.

naoyam:
Ready to review again. @jacobhinkle @zasdfgbnm

jacobhinkle (approved):
LGTM

naoyam:
!test

PR #3310, opened 2024-10-30
Add TT, TN, NT, NN tests for HopperMultipleMatmulScheduler

rdspring1:
This PR creates four tests for the `HopperMultiMatmulScheduler`. Each tests covers a different matmul layout - TT, TN, NT, and NN where the input arguments are already broadcasted.

### DIFF
--- a/tests/cpp/test_matmul_scheduler.cpp
+++ b/tests/cpp/test_matmul_scheduler.cpp
@@ -127,7 +127,7 @@ void checkUnsegmentedVectorization(
 //   D = (A x B) + bias
 //  Target architectures: Turing, Ampere
 TEST_P(PrecisionParametrizedTest, EpilogueBias) {
-  NVFUSER_TEST_CUDA_ARCH_RANGE_GUARD(7, 5, 10, 0);
+  NVFUSER_TEST_CUDA_ARCH_RANGE_GUARD(7, 5, 9, 0);
   const auto layout = MmaLayout::TT;
 
   static TestCaseErrorThresholds errs = {
@@ -227,7 +227,7 @@ TEST_P(PrecisionParametrizedTest, EpilogueBias) {
 //   D = relu(A x B)
 //  Target architectures: Turing, Ampere
 TEST_P(PrecisionParametrizedTest, EpilogueRelu) {
-  NVFUSER_TEST_CUDA_ARCH_RANGE_GUARD(7, 5, 10, 0);
+  NVFUSER_TEST_CUDA_ARCH_RANGE_GUARD(7, 5, 9, 0);
   const auto layout = MmaLayout::TT;
 
   static TestCaseErrorThresholds errs = {
@@ -312,7 +312,7 @@ TEST_P(PrecisionParametrizedTest, EpilogueRelu) {
 //  Target architectures: Ampere
 TEST_P(PrecisionParametrizedTest, EpilogueBiasRelu) {
   // NOTE: test skips Turing arch, the relative error was too big
-  NVFUSER_TEST_CUDA_ARCH_RANGE_GUARD(7, 5, 10, 0);
+  NVFUSER_TEST_CUDA_ARCH_RANGE_GUARD(7, 5, 9, 0);
   const auto layout = MmaLayout::TT;
 
   static TestCaseErrorThresholds errs = {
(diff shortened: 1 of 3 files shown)


rdspring1:
!build

tests/cpp/test_matmul_scheduler.cpp
```diff
@@ -3075,4 +3075,43 @@ TEST_F(MatmulSchedulerTest, OperandOrderIssue2434) {
...
+TEST_F(MatmulSchedulerTest, HSH_NT_128BSwizzle) {
+  NVFUSER_TEST_CUDA_ARCH_RANGE_GUARD(8, 0, 10, 0);
+  auto fusion = std::make_unique<Fusion>();
+  FusionGuard fg(fusion.get());
+
+  const auto dtype = DataType::Half;
+  constexpr auto layout = MmaLayout::NT; // [K, M] x [K, N] -> [M, N]
+
+  auto tv0 = makeContigConcreteTensor({-1, -1, 1}, dtype);
+  auto tv1 = makeContigConcreteTensor({-1, 1, -1}, dtype);
```
jacobhinkle, line 3087: I think we should start with MNK order instead of KMN.

tests/cpp/test_matmul_scheduler.cpp
```diff
@@ -3075,4 +3075,43 @@ TEST_F(MatmulSchedulerTest, OperandOrderIssue2434) {
...
+  fusion->addInput(tv0);
+  fusion->addInput(tv1);
+
+  auto tv2 = fusedMultiplySum(tv0, tv1, {0});
+
+  // Reorder the accumulator as [M, N, K]
+  // [K, M, N] -> [M, N, K]
+  tv2->reorder({{-3, -1}});
+  tv2->commitLeafToLogical();
```
jacobhinkle, line 3096: This should not be necessary I think if we use MNK order to begin with.

tests/cpp/test_matmul_scheduler.cpp
```diff
@@ -3075,4 +3075,43 @@ TEST_F(MatmulSchedulerTest, OperandOrderIssue2434) {
...
+
+  auto tv3 = castOp(DataType::Half, tv2);
+
+  fusion->addOutput(tv3);
+
+  NVF_CHECK(
+      1 == ir_utils::getOpsOfType<MmaOp>(fusion.get()).size(),
+      "matmul fusion must have at least one MmaOp");
+  FusionExecutorCache executor_cache(std::move(fusion));
```
jacobhinkle, line 3105: I think we should directly schedule and run using a `FusionExecutor`, so that we can control the scheduling params in this first test.

rdspring1:
I created TN test with `MNK` ordering, added custom `MatmulParams`, and kept the original NT tests because two tests are better than one.

rdspring1:
!build

tests/cpp/test_matmul_scheduler.cpp
jacobhinkle: Since this needs to be done at definition, should we do this when translating `MatmulOp` and `LinearOp` too? Maybe we should we just set this root->logical transform in the `fusedMultiplyAdd` function? cc @zasdfgbnm

rdspring1:
!build

rdspring1:
!build

tests/cpp/test_matmul_scheduler.cpp
```diff
@@ -3075,4 +3075,298 @@ TEST_F(MatmulSchedulerTest, OperandOrderIssue2434) {
...
+TEST_F(MatmulSchedulerTest, HSH_TT) {
+  NVFUSER_TEST_CUDA_ARCH_RANGE_GUARD(9, 0, 10, 0);
+  auto fusion = std::make_unique<Fusion>();
+  FusionGuard fg(fusion.get());
+
+  const auto dtype = DataType::Half;
+  constexpr auto layout = MmaLayout::TT;
+
+  auto tv0 = makeContigConcreteTensor({-1, -1, 1}, dtype); // A [M, K, b]
+  auto tv1 = makeContigConcreteTensor({1, -1, -1}, dtype); // B [b, K, N]
```
jacobhinkle, line 3087: In this case the MmaOp input order is MKN, and the output gets reordered with a root->logical reordering to MNK. In the TN case there's no such reordering because the logical order of inputs is MNK. Note that we can also have allocation domains set on the inputs. Maybe we could parametrize all the combinations i.e. the orders of the allocation and logical domains of the inputs?

jacobhinkle (approved):
LGTM, although it is a fair amount of code duplication that could be helped with parametrization. Also just a note that we could also test allocation domain here.

tests/cpp/test_matmul_scheduler.cpp
rdspring1: > Maybe we could parametrize all the combinations i.e. the orders of the allocation and logical domains of the inputs?

When the allocation and logical domains are different, would the input operand be not contiguous, concrete tensors?

jacobhinkle: They can be contiguous, concrete, and have permutated allocation domain. Contiguity is _with respect to_ allocation domain, so e.g. a tensor of logical shape [5, 7] and stride [7, 1] is contiguous, but so is one with logical shape [5, 7] and stride [1, 5]. The latter would correspond to having a swapped allocation domain in nvFuser.

PR #3312, opened 2024-10-30
Add script to check for non-determinism

jacobhinkle:
For example, try running the following commands:
```bash
rm -rf /tmp/nvfuser_kernel_db
tools/check_determinism.sh -- pytest -vs tests/python/test_ops.py::test_correctness_var_mean_float64
# This fails with a message like
# 10845c10845
# < __global__ void nvfuser_inner_persistent_f7_c1_r0_g2(Tensor<double, 1, 1> T0, Tensor<double, 0, 0> T8, Tensor<double, 0, 0> T7) {
# ---
# > __global__ void nvfuser_inner_persistent_f7_c1_r0_g2(Tensor<double, 1, 1> T0, Tensor<double, 0, 0> T7, Tensor<double, 0, 0> T8) {
# 10897c10897
# <     T8[0]
# ---
# >     T7[0]
# 10923c10923
# <     T7[0]
# ---
# >     T8[0]
# Diff of __tmp_kernel_inner_persistent_f7_c1_r0_g2.cu from rep 1 to rep 5 (above) is non-zero


rm -rf /tmp/nvfuser_kernel_db
export DEBUG_SERDE=disable
tools/check_determinism.sh -- pytest -vs tests/python/test_ops.py::test_correctness_var_mean_float64
# This succeeds
```
Note that this script will delete any existing *.cu files in the current directory.

### DIFF
--- a/tools/check_determinism.sh
+++ b/tools/check_determinism.sh
@@ -1,0 +1,98 @@
+#!/bin/bash
+
+set -e
+set -o pipefail
+
+usage() {
+  echo "Usage: $0 [-h] [-n NUMREPS=10] -- command to run]"
+}
+
+
+while getopts "n:h" arg
+do
+  case $arg in
+    n)
+      NUMREPS=$OPTARG
+      shift
+      ;;
+    h | ?)
+      usage
+      exit 1
+      ;;
+  esac
+done
+# getopts stops parsing if it sees "--". We can detect that case and record command
+while [[ $# -gt 0 ]]
+do
+  if [[ "$1" == "--" ]]
+  then
+    shift
+    break
+  fi
+  shift
+done
+CMD=$*
+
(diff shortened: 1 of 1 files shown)


rdspring1 (approved):
Uninformed approval, since I'm not a bash expert.

## My Summary
1. Get command.
2. Remove cuda files
3. Run command
4. Save cuda kernels to separate directories
5. Run diff between kernels
6. Fail if new kernel does not exist in first set of kernels or there is a difference between kernels
7. Succeed if all kernels match.

jacobhinkle:
> ## My Summary

You got it. Just a simple script I've found useful when behavior seems weird. BTW @rdspring1 do you have any ideas what might be going on here and how we can fix this for our CI codediff jobs?

jacobhinkle:
!build

PR #3301, opened 2024-10-29
Use deep evaluation of extents in remove_empty pass

jacobhinkle:
For dynamic fusions, we detect empty tensors and set their extents to immediate constant 0. Later, in the remove_empty preseg pass, we do a shallow check that extents are empty so that we can simplify the fusion. When the fusion is not dynamic there is no concretization step where we would do this extent replacement, so we might have constant 0 extents that are compound scalars. This caused us to miss some empty tensors in #3292, particularly one of the inputs to a `cat`.

This PR:
- Uses a deep evaluation of each `getMaybeExpandedExtent()` to determine if an axis is empty
- Adds an ExpressionEvaluator field to `EmptyTensorRemover` to avoid repeating the deep evaluation when possible. This won't help prevent repeated evaluation of symbolic extents; we could track those in an `unordered_set` potentially instead.

Fixes #3292

Issue #3292, opened 2024-10-28
INTERNAL ASSERT FAILED (Vectorized accesses cannot be inline with computation)


t-vi:
This is from the PyTorch CI, but I can also reproduce with a self-built NVFuser:

Internal assert:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/dist-packages/nvfuser/__init__.py", line 181, in execute
    results = self._execute(
              ^^^^^^^^^^^^^^
RuntimeError:  INTERNAL ASSERT FAILED at "/home/tv/data/firma/grid/thunder/Fuser/csrc/device_lower/validation.cpp":799, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. Vectorized accesses cannot be inline with computation: T70_l_float[ iblockIdx.x595{( ceilDiv(( ceilDiv(( 5 * 64 ), 4) ), blockDim.x) )}, iblockIdx.y564{( ceilDiv(( 5 * 7 ), 1) )}, iUS565{1}, iV560{4}, ithreadIdx.x562{blockDim.x} ] ca_pos( 3 )
   = pad( T16_g_float[ iS594{( ceilDiv(( ceilDiv(( 5 * 64 ), 4) ), blockDim.x) )}, iS572{( ceilDiv(( 5 * 7 ), 1) )}, iS573{1}, iS568{4}, iS570{blockDim.x} ], {0, 0, 0, 0, 0, 0, 0, i660} )

Exception raised from validateAndCollectVectorizeInfo at /home/tv/data/firma/grid/thunder/Fuser/csrc/device_lower/validation.cpp:799 (most recent call first):
frame #0: nvfuser::nvfCheckFail(char const*, char const*, unsigned int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) + 0xca (0x7f267251572c in /usr/local/lib/python3.11/dist-packages/nvfuser/_C.cpython-311-x86_64-linux-gnu.so)
frame #1: nvfuser::nvfErrorFail(char const*, char const*, unsigned int, char const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) + 0x3e (0x7f267289b31e in /usr/local/lib/python3.11/dist-packages/nvfuser/_C.cpython-311-x86_64-linux-gnu.so)
[...]

PR #3313, opened 2024-10-30
Empty cat at definition

jacobhinkle:
In #3292, a `PadOp` is created for a pad of two inputs, one of which has zero size in the cat dimension. This caused an error when we replaced the empty input with a `FullOp` output. That is addressed in the remove_empty pass by PR #3301. This PR aims to additionally simplify the fusion definition when we have concrete sizes. In particular, arguments to `pad` and `cat` are inspected for empty  dimensions and if found, we avoid using `PadOp` or `CatOp` unless necessary. The behavior is demonstrated in the included `CatOfEmpty` test:
> Cat of tensors where some non-cat dimensions are zero results in a FullOp of the proper size
Cat of tensors that are all empty in cat dim results in LoadStoreOp of the first input
Cat of tensors, some empty and some not, results in ignoring the empty inputs. In this case that leaves only a set op
Cat of tensors, some empty and some not, results in ignoring the empty inputs. In this case that leaves us with a CatOp with two inputs

### DIFF
--- a/csrc/ops/alias.cpp
+++ b/csrc/ops/alias.cpp
@@ -5,6 +5,7 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 // clang-format on
+#include <expr_evaluator.h>
 #include <expr_simplifier.h>
 #include <ir/builder.h>
 #include <ir/utils.h>
@@ -13,6 +14,7 @@
 #include <ops/utils.h>
 #include <transform_view.h>
 #include <type_promotion.h>
+#include "polymorphic_value.h"
 
 namespace nvfuser {
 
@@ -479,6 +481,21 @@ bool hasSimilarDtype(DataType base, DataType dt) {
   NVF_THROW("Unrecognized base dtype.");
 }
 
+Val* zeroForDtype(DataType dtype) {
+  // Create a zero of the appropriate type
+  if (isComplexType(dtype)) {
+    return IrBuilder::create<Val>(std::complex<double>(0), dtype);
+  } else if (isFloatingPointType(dtype)) {
+    return IrBuilder::create<Val>(0.0, dtype);
+  } else if (isBooleanType(dtype)) {
+    return IrBuilder::create<Val>(false, dtype);
+  } else {
+    return IrBuilder::create<Val>(0L, dtype);
+  }
+  NVF_THROW("Unsupported dtype in zeroForDtype: ", dtype);
+  return nullptr;
+}
+
 // Padding widths are assumed to be non-negative. Currently there's no
 // validation.
 TensorView* pad(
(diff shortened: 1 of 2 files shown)


jacobhinkle:
!build --diff

csrc/ops/alias.cpp
```diff
@@ -616,16 +652,52 @@ TensorView* cat(
...
+    // size of the first original input except in the cat dim which is the sum
+    // of the extents of `inputs` in that dimension.
+    std::vector<Val*> shape(ndims, nullptr);
+    for (const std::vector<IterDomain*>& inp_dom : inp_doms) {
+      for (size_t dim : c10::irange(ndims)) {
+        Val* extent = inp_dom.at(dim)->getMaybeExpandedExtent();
+        if (dim == cat_dim) {
+          shape[dim] = SimplifyingIrBuilder::addExpr(shape[dim], extent);
+        } else if (shape[dim] == nullptr) {
+          shape[dim] = extent;
```
naoyam, line 693: I'm not sure what the correct behavior should be here. Can we just use the first extent for a non-cat dimension? Suppose the first cat input has size zero for that dimension, the output of the cat would also have zero for the dimension. Would that be expected?

jacobhinkle: In that case it should be caught by the `inputs.empty()` condition at line 678, since all of the inputs should be empty. Actually though, I just realized that assumes that if one of the inputs has constant size 0 in one dimension that all the other inputs will have constant size zero in the same dimension. If some are symbolic this definition should be proof that they're equal by exact mapping, but they won't be detected at line 660. I guess what I should do instead is fire this condition if _any_ input has a zero non-cat dimension.

jacobhinkle:
The failure in jit_binary_tests_17_A100_1/3 is real. I have a fix but am waiting to push until the codediff is done.

csrc/ops/alias.cpp
naoyam: > if one of the inputs has constant size 0 in one dimension that all the other inputs will have constant size zero in the same dimension.

Yeah, this was what I was thinking about. I'm not sure if that's actually allowed. Is it?

jacobhinkle: We could have something like this I think:
```
tv0[ 2, 0 ]
tv1[ 3, i0 ]
tv2 = cat({tv0, tv1}, /*axis=*/0)
```
In this case we would normally exact map `tv0->axis(1)` with `tv1->axis(1)` so that `i0` must be 0. But now since we'll just translate this to `full({5, 0})` I don't think there's any such constraint, so maybe it would be legal to pass in a `tv1` with shape (3,2) and we would not hit an error.

naoyam: OK, makes sense.

jacobhinkle:
!test --diff

PR #3314, opened 2024-10-30
Use a single elect sync ite for all trasactions

zasdfgbnm:
Before:

```C++
if (elect-sync) {
  arrive-expect-tx1;
  tma1;
}
if (elect-sync) {
  arrive-expect-tx2;
  tma2;
}
```
After:
```
if (elect-sync) {
  arrive-expect-tx1;
  tma1;
  arrive-expect-tx2;
  tma2;
}
```

Perf:
```markdown
 Time (%)  Total Time (ns)  Instances  Avg (ns)  Med (ns)  Min (ns)  Max (ns)  StdDev (ns)                                                  Name

     36.0           151775          1  151775.0  151775.0    151775    151775          0.0  <unnamed>::nvfuser_none_f0_c0_r0_g0(<unnamed>::Tensor<<unnamed>::__half, (int)3, (int)3>, <unnamed>…
     20.7            87135          1   87135.0   87135.0     87135     87135          0.0  nvjet_hsh_256x128_64x4_1x2_h_bz_coopA_NTT
```
nvFuser/cuBLAS = `57.4%`.

### DIFF
--- a/csrc/device_lower/pass/circular_buffer.cpp
+++ b/csrc/device_lower/pass/circular_buffer.cpp
@@ -534,10 +534,6 @@ class CloneTmaCircularBufferLoopAndInsertSync
           return out_tv == circular_buffer_tv;
         });
 
-    if (!is_circular_buffer_load_expr) {
-      elect_sync_if_then_else_ = nullptr;
-    }
-
     insertMBarrierWaitBeforeFirstRead(expr);
 
     // Handle Short-Circuit conditions
--- a/csrc/device_lower/pass/circular_buffer.cpp
+++ b/csrc/device_lower/pass/circular_buffer.cpp
@@ -707,9 +707,9 @@ class CloneTmaCircularBufferLoopAndInsertSync
   // If there is already an if-then-else with electSync() predicate, use it.
   // Otherwise, create a new one.
   kir::IfThenElse* getElectSyncIfThenElse() {
-    if (loop_containing_elec_sync_ != for_loop_stack_.back()) {
-      elect_sync_if_then_else_ = nullptr;
-    }
+    // if (loop_containing_elec_sync_ != for_loop_stack_.back()) {
+    //   elect_sync_if_then_else_ = nullptr;
+    // }
     if (elect_sync_if_then_else_ == nullptr) {
       elect_sync_if_then_else_ = IrBuilder::create<kir::IfThenElse>(
           IrBuilder::create<kir::Predicate>(PredicateType::ElectSync));
--- a/csrc/device_lower/pass/circular_buffer.cpp
+++ b/csrc/device_lower/pass/circular_buffer.cpp
@@ -700,6 +700,17 @@ class CloneTmaCircularBufferLoopAndInsertSync
     return wait_exprs;
   }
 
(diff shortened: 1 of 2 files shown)


zasdfgbnm:
!build

csrc/index_compute.cpp
```diff
@@ -2238,6 +2238,7 @@ kir::TensorIndex* Index::getConsumerIndex(
     DataType as_type) {
   Val* index = nullptr;
   if (!ir_utils::hasRootToLoopLinearTransformations(consumer) ||
+      ir_utils::isCpAsyncBulkLoad(consumer->definition()) ||
```
zasdfgbnm, line 2241: Encountered an indexing error. Don't why. Don't care why.

zasdfgbnm:
!build

csrc/index_compute.cpp
rdspring1: For tracking purposes, what was the error?

rdspring1 (commented):
LGTM

csrc/index_compute.cpp
zasdfgbnm: I believe it is this one: https://github.com/NVIDIA/Fuser/blob/bad9e50bc9539054050310f423317b4c1d259c53/csrc/index_compute.cpp#L1966-L1975

csrc/device_lower/pass/circular_buffer.cpp
```diff
@@ -719,16 +730,14 @@ class CloneTmaCircularBufferLoopAndInsertSync
     NVF_ERROR(mbarrier_arrive_tx_ != nullptr);
     NVF_ERROR(expr != nullptr);
 
-    // Create the if-then-else with electSync() predicate for the arrive expect
-    // transaction.
-    kir::IfThenElse* if_expr = IrBuilder::create<kir::IfThenElse>(
-        IrBuilder::create<kir::Predicate>(PredicateType::ElectSync));
+    // Use the if-then-else with electSync() predicate for the arrive expect
+    // and cpAsyncBulk operations.
+    kir::IfThenElse* if_expr = getElectSyncIfThenElse();
```
naoyam, line 735: I wonder if there could be a situation where inserting a new op to the existing `if_expr` could result in an invalid order. For example, suppose the current scope has:

```
some_op1;
if (electSync()) {
  some_op2;
}
some_op3;
```

At this point, if we did `addTmaLoadBlock`, the old `IfThenElse` would be reused, but inserting a new TMA op there would mean it would be executed before `some_op3`, which might be invalid. Could something like this happen?

zasdfgbnm: Yes, this could indeed change the ordering of exprs, but I don't think it will be a problem. Because the elect-sync ITE is only for circular buffer tvs, and within a circular buffer loop, the circular buffer tvs are considered to be the "inputs" of the loop, so we are free to move circular buffer load exprs forward.

csrc/index_compute.cpp
rdspring1: That was my guess. I had to add a similar change to my matmul scheduler branch.

naoyam (approved):
LGTM

PR #3316, opened 2024-10-30
Fix IterDomain::merge with expanded inner input

naoyam:
I believe this is just a trivial bug, and, luckily, I don't think it would actually affect anything. This could matter if an expanded iter domain got merged with a non-broadcast iter domain as part of a reshape op, but reshape converts expanded iter domains to non-broadcast iter domains, so this bug won't matter.

In the case of normal scheduling, whether the output of a merge with an expanded broadcast is still a broadcast or not shouldn't matter, I believe.


naoyam:
!build --diff

jacobhinkle (approved):
LGTM. Thanks for the fix

PR #3319, opened 2024-10-31
Fix `is_clonable` in `test_issue_3292`

rdspring1:
This PR adds `ceilDiv` to serde, which enables `is_clonable` in `test_issue_3292`.


rdspring1:
!build

PR #3320, opened 2024-10-31
Load mma operands to shared memory with TMA

rdspring1:
This PR modifies `schedulePrologues` to use TMA loads to move mma operands to shared memory. Stacked on https://github.com/NVIDIA/Fuser/pull/3324 and https://github.com/NVIDIA/Fuser/pull/3310.

## Details
1. Input operands are loaded into shared memory via `CpAsyncBulkTensorTile` LoadStoreOp.
2. Replace `LdMatrix` operation with basic set.
3. Modified `scheduleOperandSmemStores` to apply swizzling to avoid bank conflicts.
4. Refactor `swizzleSharedMemory` by moving the analysis component to a separate function named `analyzeSwizzleSharedMemory`.
5. Create `tmaSwizzleSharedMemory` function that uses `analyzeSwizzleSharedMemory` and then finds the appropriate tma swizzle format.
6. Disable loop rotation. There is an issue with tma loads and circular buffering. Not sure if loop rotation is required for hopper matmul.
7. Expect hopper matmul tests to give incorrect results.

### DIFF
--- a/csrc/scheduler/hopper_multi_matmul.cpp
+++ b/csrc/scheduler/hopper_multi_matmul.cpp
@@ -31,6 +31,8 @@ namespace nvfuser {
 
 namespace {
 
+// Get gcd between megabanks and row stride AND repeated pattern size.
+// If gcd is 1, then there is no swizzle is necessary to resolve bank conflicts.
 std::pair<int64_t, int64_t> analyzeSwizzleSharedMemory(
     TensorView* shared_mem_tv) {
   NVF_ERROR(shared_mem_tv->getMemoryType() == MemoryType::Shared);
@@ -265,14 +267,11 @@ std::pair<int64_t, int64_t> analyzeSwizzleSharedMemory(
   return {g, repeated_pattern_size};
 }
 
-//! Automatically generates the shared memory swizzled data layout
-//!  for matmul mainloop and epilogue.
-//! The shared mem data layout is always 2D currently, and this utility
-//!  function assumes that the shared_mem_tv has the following structure:
-//!  [tile_row, tile_col]
-//! Returns the domain with swizzle. For the case of legacy swizzle, this
-//! domain must be set as loop domain. For the case of new swizzle, this domain
-//! must be set as allocation domain.
+//! Automatically generates the shared memory swizzled data layout for tma loads
+//! in matmul mainloop. The shared memory data layout is always 2D currently.
+//! This utility function assumes that the shared_mem_tv has the following
+//! structure: [tile_row, tile_col]
+//! Returns which swizzle format to use for mma inputs with tma loads.
 MmaInputSmemSwizzle tmaSwizzleSharedMemory(TensorView* shared_mem_tv) {
   auto&& [g, repeated_pattern_size] = analyzeSwizzleSharedMemory(shared_mem_tv);
 
@@ -297,8 +296,8 @@ MmaInputSmemSwizzle tmaSwizzleSharedMemory(TensorView* shared_mem_tv) {
   return MmaInputSmemSwizzle::B128;
 }
 
-//! Automatically generates the shared memory swizzled data layout
-//!  for matmul mainloop and epilogue.
+//! Automatically generates the shared memory swizzled data layout for matmul
+//! epilogue.
 //! The shared mem data layout is always 2D currently, and this utility
 //!  function assumes that the shared_mem_tv has the following structure:
 //!  [tile_row, tile_col]
@@ -1373,7 +1372,7 @@ void HopperMultipleMatmulScheduler::setUpCircularBuffering() {
   }
 
   /*
-  // Disable loop rotation with tma circular buffering
+  // TODO Investigate. Disable loop rotation with tma circular buffering
   if (params_->circular_buffer_options.circular_buffer_smem_read &&
       params_->circular_buffer_options.circular_buffer_smem_write) {
     // rotate Kg loop
--- a/csrc/scheduler/hopper_multi_matmul.cpp
+++ b/csrc/scheduler/hopper_multi_matmul.cpp
@@ -293,7 +293,25 @@ MmaInputSmemSwizzle tmaSwizzleSharedMemory(TensorView* shared_mem_tv) {
       num_gigarows % num_gigabanks == 0,
       "Requires non-square swizzle, which is not supported yet");
 
-  return MmaInputSmemSwizzle::B128;
+  AbstractTensor swizzle_domain(shared_mem_tv->getLoopDomain());
+  // Extract the constant sizes of the swizzled tile
+  const int64_t inner_dim_size =
+      swizzle_domain[-1]->extent()->evaluate().as<int64_t>();
+
+  auto dtype = shared_mem_tv->getDataType().value();
+  const int64_t B128_elements = 128 / dataTypeSize(dtype);
+  const int64_t B64_elements = 64 / dataTypeSize(dtype);
+  const int64_t B32_elements = 32 / dataTypeSize(dtype);
+
+  if (inner_dim_size >= B128_elements) {
+    return MmaInputSmemSwizzle::B128;
+  } else if (inner_dim_size >= B64_elements) {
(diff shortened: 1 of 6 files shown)


rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!build

rdspring1:
!test

rdspring1:
!test

csrc/scheduler/hopper_multi_matmul.cpp
```diff
@@ -31,18 +31,10 @@ namespace nvfuser {
...
-//!  [tile_row, tile_col]
-//! Returns the domain with swizzle. For the case of legacy swizzle, this
-//! domain must be set as loop domain. For the case of new swizzle, this domain
-//! must be set as allocation domain.
-//!
-//! TODO: Refactor this for TMA loads
-template <bool legacy = true>
-AbstractTensor swizzleSharedMemory(TensorView* shared_mem_tv) {
+// Get gcd between megabanks and row stride AND repeated pattern size.
+// If gcd is 1, then there is no swizzle is necessary to resolve bank conflicts.
```
jacobhinkle, line 35:
```suggestion
// Get gcd between megabanks and row stride AND repeated pattern size.
// If gcd is 1, then no swizzle is necessary to resolve bank conflicts.
```

jacobhinkle: Can we put a comment about what is the return value here?

csrc/scheduler/hopper_multi_matmul.cpp
```diff
@@ -269,6 +261,75 @@ AbstractTensor swizzleSharedMemory(TensorView* shared_mem_tv) {
...
+//! structure: [tile_row, tile_col]
+//! Returns which swizzle format to use for mma inputs with tma loads.
+MmaInputSmemSwizzle tmaSwizzleSharedMemory(TensorView* shared_mem_tv) {
+  auto&& [g, repeated_pattern_size] = analyzeSwizzleSharedMemory(shared_mem_tv);
+
+  if (g == 1) {
+    return MmaInputSmemSwizzle::None; // No need to swizzle in this case.
+  }
+
+  // 128B swizzle results in 8 x 8 matrix given half precision inputs.
```
jacobhinkle, line 282: Are we only using 128B swizzle or do we plan to support the smaller swizzles as well?

csrc/scheduler/hopper_multi_matmul.cpp
```diff
@@ -269,6 +261,75 @@ AbstractTensor swizzleSharedMemory(TensorView* shared_mem_tv) {
...
+//! Automatically generates the shared memory swizzled data layout for matmul
+//! epilogue.
+//! The shared mem data layout is always 2D currently, and this utility
+//!  function assumes that the shared_mem_tv has the following structure:
+//!  [tile_row, tile_col]
+//! Returns the domain with swizzle. For the case of legacy swizzle, this
+//! domain must be set as loop domain. For the case of new swizzle, this domain
+//! must be set as allocation domain.
+template <bool legacy = true>
+AbstractTensor swizzleSharedMemory(TensorView* shared_mem_tv) {
```
jacobhinkle, line 326: The epilogue will also use TMA so can it just use the same tma swizzling function as the operand loads? Can we just remove this?

tests/cpp/test_matmul_scheduler.cpp
```diff
@@ -3148,8 +3145,7 @@ TEST_F(MatmulSchedulerTest, HSH_TT) {
       matmul_cparams);
   auto cg_outputs = fe.runFusion({inputs.first, inputs.second});
   auto tref = atMatmul(inputs.first.squeeze(), inputs.second.squeeze(), layout);
-  EXPECT_TRUE(at::allclose(cg_outputs[0], tref, 1e-5, 1e-5));
-  */
+  EXPECT_FALSE(at::allclose(cg_outputs[0], tref, 1e-5, 1e-5));
```
jacobhinkle, line 3148: We could place a comment at these `EXPECT_FALSE` lines that these are disabled until the mma is scheduled.

tests/cpp/test_matmul.cpp
```diff
@@ -3154,6 +3154,8 @@ TEST_P(MatmulTestWithLayout, AmpereMatmulBatchSplitKBias) {
 
 // Avoid lowering error https://github.com/NVIDIA/Fuser/issues/1808
 TEST_F(MatmulTest, ReproIssue1808) {
+  NVFUSER_TEST_CUDA_ARCH_RANGE_GUARD(8, 0, 9, 0);
```
jacobhinkle, line 3157:
```suggestion
  // TODO: Make these tests work with Hopper as well as Ampere
  NVFUSER_TEST_CUDA_ARCH_RANGE_GUARD(8, 0, 9, 0);
```

csrc/scheduler/hopper_multi_matmul.cpp
```diff
@@ -1332,6 +1396,8 @@ void HopperMultipleMatmulScheduler::setUpCircularBuffering() {
     }
   }
 
+  /*
+  // TODO Investigate. Disable loop rotation with tma circular buffering
```
jacobhinkle, line 1400: My first thought was can we just disable this parameter for loop rotation, but I realized that we do not actually respect the `rotate_ldmatrix_out_of_main_loop` parameter in `MatmulParams`. In fact, I just ran `git log --patch csrc/scheduler/ | grep -C20 rotate_` and sifted through the results. I don't think we've **ever** used that parameter :-D cc @zasdfgbnm .

jacobhinkle (commented):
First pass. Looks good so far. One question: how will we handle partial vectorization?

csrc/scheduler/hopper_multi_matmul.cpp
rdspring1: I figured it would be removed once we use tma for load and stores, but keep it until then.

tests/cpp/test_matmul.cpp
rdspring1: I added this comment to the guards in `test_matmul_scheduler`. That test used specific ampere configuration.

csrc/scheduler/hopper_multi_matmul.cpp
rdspring1: I need to update this comment.

rdspring1:
> how will we handle partial vectorization?

Do you mean when the tensor is not 16B aligned? You can overcopy with TMA, cp.async, or regular LDG + STS.

jacobhinkle:
> > how will we handle partial vectorization?
> 
> Do you mean when the tensor is not 16B aligned? You can overcopy with TMA, cp.async, or regular LDG + STS.

Yeah exactly. So if we had K=60 and that is the inner dimension of each of the operands, in the Ampere scheduler we need to handle them differently when we generate the kernel since we can only do 4-element reads for the cp.async call then in stead of 8-element reads. But I don't see where that kind of alignment analysis comes in when using TMA; will TMA handle misaligned boxes dynamically using the same compiled kernel as for fully-aligned inputs?

rdspring1:
TMA should automatically handle the case when K=60 by filling the out-of-bounds accesses.
If the tensor is not 16B aligned, TMA will fail and you need to use regular set accesses.

jacobhinkle (approved):
LGTM

rdspring1:
!test

jacobhinkle:
Looks like you just need to guard `AmpereMatmulBroadcastBatch`. I noticed I needed this in #3278 but I was too lazy to merge that upstream to this PR for you. https://github.com/NVIDIA/Fuser/pull/3278/files#diff-64fc4e7bfbc5b9f95ac3dc5823bd99b683b048926805c13310ce6a8ef8032289R147-R148

rdspring1:
!test

PR #3321, opened 2024-10-31
Adding resize(PadOp) vectorization analysis

jjsjann123:
Adding **conditional** support of reszie in vectorization analysis. This PR allows vectorized load on `PadOp` directly without using cache load. This PR improves performance of generated kernel.

What's in this PR:
1. Add propagation rule for resize in vectorization analysis. The propagation rule works as:
    i. For supported resize: a). project the resize op to the frontier and clear `(frontier.begin(), resize_position)`; b). add projected extent of the new resize op as `gcd(id_from, resize_op->leftExpand(), resize_op->rightExpand)`
    ii. For unsupported resize: clear `[frontier.begin(), resize_position]`; no behavior change.

2. updating TensorView::cacheAfter to opt-in a set of uses to cache while leaving other uses unchanged. Necessary for cases where inputs are used by PadOp as well as other operation that relies on cached load for vectorization.

Follow up to #3261.
Work for supporting rope performance. [design doc](https://docs.google.com/document/d/1tafRMNIXMmHlIGAiNlaPkYp6mZAzJ2Rh_NtARHbmYNA/edit?disco=AAABYEnV_ZY):


jjsjann123:
!test

jjsjann123:
hmmm. this one isn't functional yet. The vectorize factor isn't computed correctly.

jjsjann123:
!test

csrc/scheduler/vectorize_helper.cpp
```diff
@@ -365,9 +392,48 @@ std::vector<IterDomain*> ContiguousInnerDimensionsMapper::projectId(
...
+      // resize created by PadOp.
+
+      // project resize op to frontier.
+      frontier[pos] = id_to;
+      // clear left of resize, since those are no long contiguous.
+      frontier.erase(frontier.begin(), it);
+
+      if (recording_) {
+        // since this is a pad operation, we don't need to worry about the
+        // offset, but rather the pad_extent, which becomes the real buffer on
```
naoyam, line 419: What do you refer to by "offset"?

csrc/tensor_view.cpp
```diff
@@ -1168,15 +1168,26 @@ TensorView* TensorView::cacheFork() {
...
     CacheOp cache_op,
-    bool propagate_allocation_domain) {
+    bool propagate_allocation_domain,
+    const std::unordered_set<Expr*>& cached_uses) {
   NVF_ERROR(
       !container()->isA<kir::Kernel>(),
       "Function invalid for kernel container.");
   FusionGuard fg(fusion());
 
+  auto target_uses = fusion()->unordered_uses(this);
```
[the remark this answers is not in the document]
jjsjann123: It's used for mostly checks, except
```
  for (auto expr : target_uses) {
    ir_utils::replaceValInExprInputs(expr, this, consumer);
  }
```

I can switch to `uses()` and filter it by uniqueness.

csrc/scheduler/vectorize_helper.cpp
```diff
@@ -365,9 +392,48 @@ std::vector<IterDomain*> ContiguousInnerDimensionsMapper::projectId(
...
+        // output. Hence we do GCD among padded extent as well as extent of the
+        // id_from. Note since we are taking the GCD here, I don't think using
+        // id_from or id_to makes a difference.
+        auto consumer_factor = getProjectedExtent(id_from);
+        auto comp = [](Val* factor, Val* extent) {
+          return SimplifyingIrBuilder::whereExpr(
+              SimplifyingIrBuilder::eqExpr(
+                  extent, extent->container()->zeroVal()),
```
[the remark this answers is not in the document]
jjsjann123: the extent here is the padding extent (left or right). Not the extent of the padded id.

I should add a test for this.

csrc/scheduler/vectorize_helper.cpp
jjsjann123: I was referring to the offset created by slice. i.e. negative resize extent.

csrc/scheduler/vectorize_helper.cpp
```diff
@@ -365,9 +392,48 @@ std::vector<IterDomain*> ContiguousInnerDimensionsMapper::projectId(
     distributePE(merge_or_split);
   };
 
-  auto clear_left_of = [&frontier](IterDomain* id) {
-    auto it = std::find(frontier.begin(), frontier.end(), id);
-    if (it != frontier.end()) {
+  auto propagateResize = [&frontier, this](Resize* resize_op, bool p2c) {
+    // TODO: resize could affect the offset/pad_extent impacting vectorization
+    // factor on the loading inputs. But we shouldn't use that to block the
```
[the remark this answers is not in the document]
jjsjann123: Yes we are using that here. I'm just wondering if that is the right approach supporting slices, since it's going to be more restrictive.
I should remove this comment, since it's only adding confusion here.

csrc/scheduler/vectorize_helper.cpp
```diff
@@ -365,9 +392,48 @@ std::vector<IterDomain*> ContiguousInnerDimensionsMapper::projectId(
...
+        // output. Hence we do GCD among padded extent as well as extent of the
+        // id_from. Note since we are taking the GCD here, I don't think using
+        // id_from or id_to makes a difference.
+        auto consumer_factor = getProjectedExtent(id_from);
+        auto comp = [](Val* factor, Val* extent) {
+          return SimplifyingIrBuilder::whereExpr(
+              SimplifyingIrBuilder::eqExpr(
+                  extent, extent->container()->zeroVal()),
+              factor,
+              SimplifyingIrBuilder::gcdExpr(factor, extent));
```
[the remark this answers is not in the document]
naoyam: > sorry, this is bad naming on my side. `factor` should be positive.
> 
> I think you mean padding extent being negative, i.e. from slicing. Which would be a bit different here.

It's effectively slicing, but `pad` itself allows negative offsets.

csrc/preseg_passes/move_pad.cpp
```diff
@@ -287,7 +287,10 @@ TensorView* replayConcretePad(
 
   auto* new_out = IrBuilder::create<TensorView>(
       IrBuilder::create<TensorDomain>(
-          merged_root_ids, merged_logical_ids, merged_logical_ids),
+          merged_root_ids,
+          merged_logical_ids,
+          merged_logical_ids,
+          TensorDomain::getContiguityFilledWith(merged_logical_ids, true)),
```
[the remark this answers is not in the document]
jjsjann123: Great question. No I don't need this change to pass the added tests. Since we only check contiguity on IO tensors.

But seems like the right thing to do.

csrc/scheduler/vectorize_helper.cpp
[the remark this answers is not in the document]
jjsjann123: good point. Forgot about that when I was only looking at `cat`.

can limit the factor to be 1, effectively disabling it for negative extent. We can address that in the follow up PR when we enable slice. Does that sound reasonable?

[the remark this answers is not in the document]
naoyam: Yes, makes 100% sense. We would just need some check somewhere.

csrc/tensor_view.cpp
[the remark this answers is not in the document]
naoyam: Yes, I'm concerned about that part since `replaceValInExprInputs` would create new IR nodes, and while the final kernel would do the same computations, the generated kernel would have non-deterministic variable names.

csrc/scheduler/vectorize_helper.cpp
naoyam: "Negative resize extent" sounds less ambiguous.

naoyam (commented):
It overall looks good. Just would like a few things I commented about to get addressed.

naoyam:
!test --pybench

naoyam:
Initiated testing with python benchmarks just in case.

jjsjann123:
Thanks, I'll address the issues you brought up as well as running through some real size problem so we get a taste of the perf impact. :bow:

jjsjann123:
!test --pybench

jjsjann123:
!test --pybench

jjsjann123:
Did a quick look at the perf. The end-2-end time looks very noisy. I'm a bit unsure about my measuring script so instead just did a nsys on the kernel time.

On A100 80GB PCIe, peak bandwidth is 2TB/s.
* At bsz 256 with bf16, we are looking at IO size of roughtly 512MB `(256 * 1024 * 16 * 32 * 2 * 2 + 1024 * 8 * 2 * 2) / 1024 / 1024`. `main` branch kernel time is roughly 510ms, achieving ~ 1TB/s. With vectorization, we boost it to 368ms ~ 1.39TB/s.
* At bsz 516 with bf16, `main` gives 1013ms ~ 1.01TB/s. With vectorization 732ms ~ 1.40TB/s

Something like this vvv.
```
import torch
import thunder
def rope_one_entry(x: torch.Tensor, cos: torch.Tensor, sin: torch.Tensor, rope_n_elem: int) -> torch.Tensor:
    x_rope = x[..., : rope_n_elem]
    x1 = x_rope[..., : rope_n_elem // 2]  # (B, nh, T, hs/2)
    x2 = x_rope[..., rope_n_elem // 2 :]  # (B, nh, T, hs/2)
    rotated = torch.cat((-x2, x1), dim=-1)  # (B, nh, T, hs)
    roped = (x_rope * cos) + (rotated * sin)
    [roped.to](http://roped.to/)(dtype=x.dtype)
    return torch.cat((roped, x[..., rope_n_elem :]), dim=-1)
dtype = torch.bfloat16
device = "cuda"
bsz = 256
block_size = 1024
n_head = 16
head_size = 32
n_query_groups = 4
rope_n_elem = 8
WARMPU_ITER = 5
MEASURE_ITER = 20
cos = torch.randn(block_size, rope_n_elem, device=device, dtype=dtype)
sin = torch.randn(block_size, rope_n_elem, device=device, dtype=dtype)
thunder_rope_one = thunder.jit(rope_one_entry, executors=("nvfuser",) , nv_enable_bookend=False)
x = torch.randn([bsz, n_head, block_size, head_size], device=device, dtype=dtype)
# ref full run
o_ref = rope_one_entry(x.float(), cos.float(), sin.float(), rope_n_elem).to(dtype=dtype)
l2_clear_buffer = torch.empty(80, 1024, 1024, dtype=torch.float, device="cuda")
# warm up
for i in range(WARMPU_ITER):
    o = thunder_rope_one(x, cos, sin, rope_n_elem)
# measurement
for i in range(MEASURE_ITER):
    l2_clear_buffer.zero_()
    o = thunder_rope_one(x, cos, sin, rope_n_elem)
assert(o.allclose(o_ref))
```

jjsjann123:
I think review comments have been addressed as well. CI was green before my benchmark. Ready for a final review.

jjsjann123:
!test --pybench

jjsjann123:
build failure seems to be flaky. I re-started the CI on that one and it passed internally.

Unfortunately it didn't update the github status here. Not a big issue but cc'ing @xwang233 in case this is something you are not aware of.

csrc/tensor_view.cpp
```diff
@@ -1168,15 +1168,39 @@ TensorView* TensorView::cacheFork() {
...
 
+  std::vector<Expr*> target_uses;
+  if (!cached_uses.empty()) {
+    std::unordered_set<Expr*> unique_uses = fusion()->unordered_uses(this);
+    target_uses.reserve(cached_uses.size());
+    for (auto use : cached_uses) {
+      NVF_ERROR(
+          unique_uses.count(use),
+          "cached_uses is not among the use of the TensorView");
+      target_uses.push_back(use);
```
naoyam, line 1186: Isn't this ordering still non deterministic? I think the parameter itself needs to be deterministically ordered.

jjsjann123: Good catch. Sorry I missed the `cached_uses`. Let me give it another try.

csrc/tensor_view.cpp
```diff
@@ -1168,15 +1168,36 @@ TensorView* TensorView::cacheFork() {
...
+    std::unordered_set<Expr*> unique_uses = fusion()->unordered_uses(this);
+    for (auto use : cached_uses) {
+      NVF_ERROR(
+          unique_uses.count(use),
+          "cached_uses is not among the use of the TensorView");
+    }
+  } else {
+    // avoid non-determinism and ensure unique
+    std::unordered_set<Expr*> unique_uses;
+    auto this_uses = uses();
```
naoyam, line 1188: Is it possible to have duplicates?

jjsjann123: I don't think so. since we have `uses_` as private and addUse does skip duplicates https://github.com/NVIDIA/Fuser/blob/a5022da00e2e429228e78b309fb78bc3da3cf49c/csrc/ir/base_nodes.cpp#L96-L102

But I'm a bit scared by just leaving it not checked, since it's only a `std::vector`.

naoyam: If so, let's make it an error if a duplicate is found.

jjsjann123: switched. I'll kick off the CI to see if there's any surprises. :crossed_fingers:

jjsjann123:
!test --pybench

naoyam (approved):
LGTM

PR #3323, opened 2024-10-31
Make it more explicit about the rfactor flag propagation

naoyam:
`IterDomain::split` has a default value of `false` for the rfactor flag, which means if an rfactor iter domain is split, it would generate non-rfactor output domains. I'm not sure if that's the right default behavior, so this PR removes the default value and make the behavior more explicit.

Nothing should be affected by this change, as long as all the tests and benchmarks don't fail.

### DIFF
--- a/csrc/ir/internal_base_nodes.h
+++ b/csrc/ir/internal_base_nodes.h
@@ -131,7 +131,25 @@ class NVF_API IterDomain : public Val {
       IterDomain* in,
       Val* factor,
       bool inner_split,
-      bool rfactor_domain = false);
+      bool rfactor_domain);
+
+  // Shortcut of the above split function. Previously, the above split function
+  // had a default value of false for the rfactor_domain flag. The
+  // default value is removed as it may implicitly drop
+  // the rfactor flag when the input is an rfactor iter domain, which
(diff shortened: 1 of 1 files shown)


naoyam:
!build

naoyam:
Failed cases:

```
NvFuserScheduler_Matmul_Manual/nvfuser_splitk_TT/M:1024/N:1024/K:50304/warps:8/stages:3/splitk_factor:2/smem_epilogue:0/manual_time for 1
00:00:10 terminate called after throwing an instance of 'nvfuser::nvfError'
00:00:10   what():   INTERNAL ASSERT FAILED at "/opt/pytorch/nvfuser/csrc/ir/internal_base_nodes.h":147, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. Unexpected input iter domain. Input should not be an rfactor iter domain: rS196{32}rf
00:00:10 Exception raised from split at /opt/pytorch/nvfuser/csrc/ir/internal_base_nodes.h:147 (most recent call first):

naoyam:
Actually, never mind. Rfactor should be only for between root and logical, so in normal cases the rfactor flag should be propagated to outputs.

PR #3275, opened 2024-10-25
add knobs control inner dim unroll and outer dim unroll in pointwise scheduler

liqiangxl:
**What's in this PR?**
(1) Added two knobs to control unroll in inner dim and outer dim for pointwise scheduler
(2) The original untoll knob which applies to outer dim is removed.
(3) Extended test `UnrollOnTopOfVectorize` to test 8 different combinations of `vectorization`, `inner unroll`, and `outer unroll`.
(4) Neither `inner unroll` nor `outer unroll` is used in the heuristics. They are always `1` unless `vectorization == 1`, in that case, `inner unroll` is used.
(5) If `inner or outer unroll factor == 1`, we won't split out an additional domain with size of `1`.

**Why?**
These two knobs allows more performance optimizations, e.g. unroll in different dims based on broadcast dims.

PR #3325, opened 2024-11-01
add knobs control inner dim unroll and outer dim unroll in pointwise scheduler redo pr-3275 to check code changes

liqiangxl:
redo #3275 to check code changes.


liqiangxl:
!test --diff-bench --diff

liqiangxl:
(1) code diff in `nvfuser-ci/jit_codegen_diff_bench_20_10/10` and `
nvfuser-ci/jit_codegen_diff_bench_20_9/10` are due to we deleted the split when unroll factor is 1. Removing one loop, leads to different compute at position and different expression order. This change is not a blocker since it only happens when the input can't be vectorized. I didn't check the performance since it is rare and we have a following PR reivising heuristics.
image

liqiangxl:
(2) code diffs in `nvfuser-ci/jit_codegen_diff_20_5/7` , `nvfuser-ci/jit_codegen_diff_20_6/7 `, and `nvfuser-ci/jit_codegen_diff_20_7/7 `are RNGTest, also because of the input size can't be vectorized and split of unroll = 1 is skipped.

(3) code diffs in `ci/jit_codegen_diff_20_7/7` also has a `DynamicTransformIssue418_CUDA` where output order is switched. can't reproduce in local run.

PR #3327, opened 2024-11-01
Use IterDomain::split and IterDomain::merge

### LINKED ISSUE #2520: refactor handle(Merge*/Split* s) to use IterDomain::merge/split
see the original [comments](https://github.com/NVIDIA/Fuser/pull/2489#discussion_r1664573349).

To reduce the duplication of the logic.

naoyam:
Instead of manually creating output IDs and directly constructing Split and Merge nodes, IterDomain::split and IterDomain::merge are used. This would reduce duplication of logic.

There's some slight change in iteration types of IDs generated when `TensorView::rFactor` is used. See the inline comment.

As of [8321cfd](https://github.com/NVIDIA/Fuser/pull/3327/commits/8321cfdd27ac15e9464f551e0c2ea0e5159357db), everything seems fine.

Closes #2520

### DIFF
--- a/csrc/ir/internal_base_nodes.h
+++ b/csrc/ir/internal_base_nodes.h
@@ -120,18 +120,23 @@ class NVF_API IterDomain : public Val {
   static std::vector<IterDomain*> clone(
       const std::vector<IterDomain*>& domains);
 
-  //! When `rfactor_domain` is true, also set the `is_rfactor_domain_` flag of
-  //! the result IterDomain.
+  //! The optional parameters of rfactor_domain and iter_type can be
+  //! used to override the default behavior.
   static IterDomain* merge(
       IterDomain* outer,
       IterDomain* inner,
-      bool rfactor_domain = false);
+      std::optional<bool> rfactor_domain = std::nullopt,
+      std::optional<IterType> iter_type = std::nullopt);
 
+  //! The optional parameters of rfactor_domain, outer_iter_type and
+  //! inner_iter_type can be used to override the default behavior.
   static std::pair<IterDomain*, IterDomain*> split(
       IterDomain* in,
       Val* factor,
       bool inner_split,
-      bool rfactor_domain = false);
+      std::optional<bool> rfactor_domain = std::nullopt,
+      std::optional<IterType> outer_iter_type = std::nullopt,
+      std::optional<IterType> inner_iter_type = std::nullopt);
 
   //! Resize an IterDomain by expanding both the left and right sides
   //! by given widths. The resulting IterDomain has an extent of
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -2550,7 +2550,8 @@ IterDomain* IterDomain::cloneWithoutRFactor(bool map_with_original) {
 IterDomain* IterDomain::merge(
     IterDomain* outer,
     IterDomain* inner,
-    bool rfactor_domain) {
+    std::optional<bool> rfactor_domain,
+    std::optional<IterType> iter_type) {
   NVF_CHECK(
       outer->isReduction() == inner->isReduction(),
       "Merging IterDomains requires that their iteration types match. ",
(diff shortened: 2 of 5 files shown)


naoyam:
!test --diff

csrc/ir/internal_base_nodes.h
```diff
@@ -120,18 +120,23 @@ class NVF_API IterDomain : public Val {
...
 
-  //! When `rfactor_domain` is true, also set the `is_rfactor_domain_` flag of
-  //! the result IterDomain.
+  //! The optional parameters of rfactor_domain and iter_type can be
+  //! used to override the default behavior.
   static IterDomain* merge(
       IterDomain* outer,
       IterDomain* inner,
-      bool rfactor_domain = false);
+      std::optional<bool> rfactor_domain = std::nullopt,
```
naoyam, line 128: Changed this parameter to `std::optional` so that it's clear if the argument of `false` is intentional.

csrc/ir/nodes.cpp
naoyam, line 1: There should be no change of logic in this file.

naoyam:
!test --diff

csrc/transform_rfactor.cpp
```diff
@@ -119,32 +116,27 @@ class ReplayRFactor : public ReplayTransformations {
...
-                rfactor_axes_.count(s->outer()) ? IterType::Reduction
-                                                : IterType::Iteration)
-            .is_rfactor_domain(static_logical_outputs)
-            .build();
+    // Let IterDomain::split determine the correct IterType, except
+    // when the output is a reduction domain but not part of the
+    // rfactored domains. If it isn't involved in the rfactor, it's no
+    // longer a redunction domain
+    std::optional<IterType> outer_iter_type;
+    if (s->outer()->isReduction() && !rfactor_dep_ids_.count(s->outer())) {
```
naoyam, line 124: Previously, every output iter domain uses `IterType::Iteration` unless it's one of the rfactored loop IDs. This means that, for example:

```
auto t0 = [i0, r1, r2];
t0->merge(1, 2); // [i0, r1*r2]
t0->split(1, 4); // [i0, r1*r2/4, 4]
t1 = t0->rFactor(-1);
```

In this case, `t1` would inherit all the transformations of `t0` and the merge output would not be marked as a reduction domain.

Generally, when a tensor is rfactored, whether it's `Iteration` or `Reduction` seems to be blurred, and there's a case where a reduction iter domain is merged with a non-reduction iter domain. That seems to be one of the reasons `IterType::merge` wasn't used since the function asserts both inputs should have the same iter type.

In this PR, I tried to make it more consistent by looking at all IDs between the logical reduction IDs and the rfactored loop IDs. Any IDs, including the intermediate ones should remain reductions, so there should be no case of merging `Iteration` and `Reduction`.

naoyam: Added a simple repro just in case.

naoyam:
!test

csrc/transform_replay.cpp
```diff
@@ -52,25 +52,20 @@ class ReplaySelf : public ReplayTransformations {
...
-    // Generate the split node
-    IrBuilder::createInContainer<Split>(
-        s->container(), ido, idi, mapped, s->factor(), s->innerSplit());
+    NVF_ERROR(s->outer()->isRFactorProduct() == s->inner()->isRFactorProduct());
+
+    // Due to rfactor transformations, the iter types of the outputs
+    // may not follow the default rule. For example, even if the input
+    // is a reduction iter domain, the outputs may not. To replay the
+    // original split expression, the output iter types need to be
+    // specified explicitly.
```
jacobhinkle, line 61: Thanks for this comment.

tests/cpp/test_gpu3.cpp
```diff
@@ -8874,6 +8874,31 @@ TEST_F(NVFuserTest, CpAsyncDataTypeBool) {
...
+  auto tv1 = sum(tv0, {1, 2});
+  fusion.addOutput(tv1);
+
+  tv1->merge(1, 2);
+  tv1->split(1, 4);
+
+  auto tv2 = tv1->rFactor({-1});
+
+  auto split = dynamic_cast<Split*>(tv2->axis(-1)->definition());
+  ASSERT_NE(split, nullptr);
```
jacobhinkle, line 8896: Can we also do `EXPECT_TRUE(tv2->axis(-1)->isIteration())` to check that altering the intermediate domains to `Reduction` does not interfere with the factoring; i.e. that this is still only a _partial_ sum.

naoyam: Sure, though it's actually `tv2->axis(-1)->isReduction()` and `tv2->axis(-2)->isIteration` since it's the innermost domain that is rfactored. I'll add an assertion.

naoyam:
!build

PR #3330, opened 2024-11-02
split inner reduction heuristics into 2d and 3d heuristics

liqiangxl:
**What's in this PR?**
This PR separates inner reduction heuristics into distinct 2D and 3D heuristic functions.

**Why?**
The 2D and 3D reductions represent different domain structures within the reduction tensor view:

- 2D inner reduction: Domains in the reduction tensor view can be merged into a 2D structure, [I, R].
- 3D inner reduction: Domains can be merged into a 3D structure, [R, I, R].

These two configurations require different parallelization strategies, so keeping them in separate functions enhances maintainability and allows for individual optimization of each heuristic.

**code changes**
The existing `innerReductionHeuristic()` is duplicated as `inner2dReductionHeuristic` and `inner3dReductionHeuristic`, will clean  `inner2dReductionHeuristic` in a separate PR.

### DIFF
--- a/csrc/scheduler/reduction.cpp
+++ b/csrc/scheduler/reduction.cpp
@@ -63,14 +63,13 @@ void reduceProductTo(int64_t& z, int64_t& y, int64_t& x, const int64_t max) {
   }
 }
 
-std::unique_ptr < ReductionParams >
-    2dInnerReductionHeuristic(
-        const int64_t total_reduction_numel,
-        const int64_t total_iteration_numel,
-        const int64_t inner_most_dimension_numel,
-        const int64_t n_tensor_inputs,
-        const int64_t max_input_dtype_size,
-        const size_t vectorize_factor) {
+std::unique_ptr<ReductionParams> inner2dReductionHeuristic(
+    const int64_t total_reduction_numel,
+    const int64_t total_iteration_numel,
+    const int64_t inner_most_dimension_numel,
+    const int64_t n_tensor_inputs,
+    const int64_t max_input_dtype_size,
+    const size_t vectorize_factor) {
   // Set some targets for parallelization
 
   const int64_t n_elems = total_reduction_numel * total_iteration_numel;
@@ -493,7 +492,7 @@ std::unique_ptr < ReductionParams >
                 << (rparams->unroll_factor_inner_reduction > 1) << ", "
                 << rparams->cross_grid_inner_reduction << std::endl;
       }
-      return innerReductionHeuristic(
+      return inner2dReductionHeuristic(
           total_reduction_numel,
           total_iteration_numel,
           total_reduction_numel,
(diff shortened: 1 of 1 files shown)


liqiangxl:
!build

liqiangxl:
!build

csrc/scheduler/reduction.cpp
```diff
@@ -63,7 +63,7 @@ void reduceProductTo(int64_t& z, int64_t& y, int64_t& x, const int64_t max) {
   }
 }
 
-std::unique_ptr<ReductionParams> innerReductionHeuristic(
+std::unique_ptr<ReductionParams> inner2dReductionHeuristic(
```
jjsjann123, line 66: nitpick: this is renamed as a 2d reduction. But in the code comment, we refer to this as 1D schedule, which is a bit confusing.

csrc/scheduler/reduction.cpp
```diff
@@ -492,7 +492,449 @@ std::unique_ptr<ReductionParams> innerReductionHeuristic(
...
+            << "total_iteration_numel: " << total_iteration_numel << "\n"
+            << "vectorize_factor: " << vectorize_factor << "\n"
+            << "n_tensor_inputs: " << n_tensor_inputs << "\n"
+            << "max_input_dtype_size: " << max_input_dtype_size << "\n"
+            << "block(" << bdimx << ", " << bdimy << ", " << bdimz << ")"
+            << std::endl;
+    debug() << rparams->toString() << std::endl;
+  }
+
+  // If 3d, check if it's supported by the scheduler, otherwise force 1D
```
jjsjann123, line 925: oh I think I was mistaken. So if we do return `rparams` at line 947, that's the 1d schedule this comment is referring to?

liqiangxl: Should be 2D, we don't have `1D` scheduler.

liqiangxl:
!build

csrc/scheduler/reduction.cpp
```diff
@@ -63,7 +63,449 @@ void reduceProductTo(int64_t& z, int64_t& y, int64_t& x, const int64_t max) {
...
+
+    rparams->grid_dim_iter_dom = ParallelType::BIDy;
+    if (godim > scheduler_utils::y_grid_limit) {
+      rparams->split_grid_dim_iter_dom_outer = true;
+      gdimy = std::min(godim, scheduler_utils::y_grid_limit);
+    }
+
+  } else {
+    rparams->grid_dim_iter_dom = ParallelType::BIDx;
+    if (gdimx > scheduler_utils::x_grid_limit) {
```
rdspring1, line 443: Isn't `gdimx` set to `LaunchParams::UNINITIALIZED_VAL` here?
Therefore, this if-block is alway False.

jjsjann123: :face_with_head_bandage: Is this supposed to be godim? And since this is just a copy of the old function, we also need to fix the other one then.

PR #3334, opened 2024-11-03
Build Segments for User Schedule Segmentation

rdspring1:
## General Overview of Segmentation:

Segmentation decomposes a fusion into a directed acyclic graph (DAG) of sub-fusions. After applying the segmentation algorithm, we can translate the sub-fusions into their corresponding python definitions. Then, given the fusion's input arguments, the segments are run in the correct order to produce the output results.

The original FusionDefinition stores the sequence of sub-fusions and acts as an argument manager. It gathers the input arguments before running the sub-fusion and stores its results. To perform this function, it requires a map from the segment index space to the original index space. This mapping is generated while creating the python definition for each sub-fusion.

### CPP functions:
Step 1: `setupSegmentation` runs the segmentation algorithm on the CPP Fusion to create the `SegmentedFusion`. Then, sub-fusions are ordered according to their dependencies by the `prepareGroupOrder` function. It returns the number of segments in `SegmentedFusion`.

Step 2: `buildSegment` creates the CPP `Fusion` for a given segment id, translates it to a python `FusionDefinition`, then returns a mapping from the segment fusion state indices to the original fusion state indices.

Step 3: `finalizeSegmentation` destroys any state stored in `FusionDefinition`.

### Python functions:

1. `setupSegmentation`, `buildSegment`, and `finalizeSegmentation` are called together in `FusionDefinition.segment`.
6. If a python `FusionDefinition` has segments, call `_execute_segments` in the `FusionDefinition.execute`. The original `FusionDefinition` acts as argument manager, running the sub-fusions in topological order.

```python
                          contiguity=[True, True],
                          dtype=DataType.Float,
                          is_cpu=False)
                          contiguity=[True, True],
                          dtype=DataType.Float,
                          is_cpu=False)
    T4 = fd.ops.add(T1, T3)
    fd.add_output(T4)
```

### After Segmentation:
- The reduction scheduler does not support fusing any operations with an inner reduction, so the original fusion is divided into two segments. Segment 2 depends on Segment 1, so there is a strict ordering of the segments.
- The first segment contains the reduction and broadcast operations, which corresponds with [T0, T2, T3] in the original fusion. 
- The second segment is the pointwise addition with the broadcasted reduction. It corresponds with [T1, T3, T4] in the original fusion.

**First Segment:**
```python
def nvfuser_fusion_id2(fd : FusionDefinition) -> None :
                         contiguity=[True, True],
                         dtype=DataType.Float,
                         is_cpu=False)
   T1 = fd.ops.sum(T0, dims=[1], keepdim=False, dtype=DataType.Float)
   T2 = fd.ops.broadcast(T1, is_broadcast_dim=[False, True])
   fd.add_output(T2)
```

**Second Segment:**
```python
def nvfuser_fusion_id3(fd : FusionDefinition) -> None :
                         contiguity=[True, True],
                         dtype=DataType.Float,
                         is_cpu=False)
   T1 = fd.define_tensor(shape=[-1, 1],
                         contiguity=[True, None],
                         dtype=DataType.Float,
                         is_cpu=False)
   T2 = fd.ops.add(T0, T1)
   fd.add_output(T2)
```

## Changes in this PR

This PR implements `setupSegmentation` function for user-scheduler segmentation. It is the first PR in a stack, followed by https://github.com/NVIDIA/Fuser/pull/3335 and https://github.com/NVIDIA/Fuser/pull/3025.

1. Create `SegmentationState` class that contains all segmentation logic for python-frontend.
2. All segmentation logic is contained in a separate file - `csrc/python_frontend/segmentation.h`
3. `FusionDefinition` contains an instantiation of `SegmentationState` and exposes its logic in a public interface. This interface is added to the python bindings.
4. Created `test_segmentation_reduction_pointwise_epilogue` to test functionality.

### DIFF
--- a/csrc/python_frontend/fusion_definition.cpp
+++ b/csrc/python_frontend/fusion_definition.cpp
@@ -9,6 +9,7 @@
 #include <fusion_profiler.h>
 #include <instrumentation.h>
 #include <options.h>
+#include <preseg_passes/pre_segmenter.h>
 #include <python_frontend/fusion_cache.h>
 #include <python_frontend/fusion_definition.h>
 #include <python_frontend/translation.h>
@@ -671,6 +672,21 @@ std::vector<Tensor> FusionDefinition::tensors() {
 std::vector<std::pair<double, double>> FusionDefinition::getValTolerances(
     const at::ArrayRef<c10::IValue>& inputs) {
   return get_val_constants(preschedFusion(), inputs);
+}
+
+int64_t FusionDefinition::setupSegmentation(
+    const at::ArrayRef<c10::IValue>& inputs) {
+  NVF_CHECK(id().has_value(), "FusionDefinition definition does not exist!");
+  NVF_ERROR(
+      segmentation_state_ == nullptr, "SegmentationState already exists!");
+  segmentation_state_ = std::make_unique<SegmentationState>();
+  return segmentation_state_->setupSegmentation(
+      preschedFusion(), map_value_to_fid_, inputs);
+}
+
+void FusionDefinition::finalizeSegmentation() {
+  // Destroy SegmentedState
+  segmentation_state_.reset();
 }
 
 } // namespace nvfuser::python_frontend
--- a/csrc/python_frontend/fusion_state.cpp
+++ b/csrc/python_frontend/fusion_state.cpp
@@ -68,6 +68,27 @@ std::ostream& operator<<(std::ostream& os, const State& state) {
   return os;
 }
 
+std::vector<Val*> getExtents(Fusion* fusion) {
+  NVF_CHECK(fusion != nullptr, "Fusion is undefined.");
+
+  std::vector<Val*> extents;
+  for (Val* v : fusion->inputs()) {
+    // short-circuit: skip if not TensorView
+    if (!v->isA<TensorView>()) {
+      continue;
+    }
+    TensorView* tv = v->as<TensorView>();
+    std::vector<IterDomain*> logical_dom =
+        TensorDomain::noReductions(tv->getLogicalDomain());
+    std::transform(
+        logical_dom.begin(),
+        logical_dom.end(),
+        std::back_inserter(extents),
+        [](IterDomain* id) { return id->getMaybeExpandedExtent(); });
+  }
+  return extents;
+}
+
 FusionState::FusionState()
     : end_record_(new EndRecord()),
       recording_(),
@@ -247,27 +268,6 @@ const std::vector<int64_t>& FusionState::outputs() const {
 
 const std::vector<int64_t>& FusionState::extents() const {
   return extents_fid_;
-}
-
-std::vector<Val*> FusionState::getExtents(Fusion* fusion) {
-  NVF_CHECK(fusion != nullptr, "Fusion is undefined.");
-
-  std::vector<Val*> extents;
-  for (Val* v : fusion->inputs()) {
-    // short-circuit: skip if not TensorView
-    if (!v->isA<TensorView>()) {
-      continue;
-    }
-    TensorView* tv = v->as<TensorView>();
-    std::vector<IterDomain*> logical_dom =
-        TensorDomain::noReductions(tv->getLogicalDomain());
-    std::transform(
-        logical_dom.begin(),
-        logical_dom.end(),
-        std::back_inserter(extents),
-        [](IterDomain* id) { return id->getMaybeExpandedExtent(); });
-  }
-  return extents;
 }
 
 void FusionState::addExtents() {
--- a/csrc/python_frontend/segmentation.cpp
+++ b/csrc/python_frontend/segmentation.cpp
@@ -1,0 +1,146 @@
+// clang-format off
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2023-present NVIDIA CORPORATION & AFFILIATES.
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+// clang-format on
+#include <python_frontend/fusion_definition.h>
+#include <python_frontend/translation.h>
+
+namespace nvfuser::python_frontend {
+
+int64_t SegmentationState::setupSegmentation(
+    Fusion* fusion,
+    const std::unordered_map<const Val*, int64_t>& map_value_to_original_fid,
+    const at::ArrayRef<c10::IValue>& inputs) {
+  // Check state
+  NVF_ERROR(fusion != nullptr);
+  NVF_ERROR(segment_fusion_ == nullptr);
+  NVF_ERROR(segmented_fusion_ == nullptr);
+  NVF_ERROR(group_run_order_.empty());
+  NVF_ERROR(map_cloned_value_to_fid_.empty());
+  NVF_ERROR(cloned_extents_.empty());
+
+  int8_t device = getCommonDeviceCUDA(inputs);
+  NVF_CHECK(
+      inputs.empty() || device > -1, "Inputs are not all on the same device!");
+
+  // Step 1) Clone preschedFusion CPP Fusion.
+  segment_fusion_ = std::make_unique<Fusion>();
+
+  // The IRCloner returned by Fusion::copy acts as map from the original fusion
+  // to the cloned fusion.
+  IrCloner original_to_cloned_map = Fusion::copy(fusion, segment_fusion_.get());
+
+  KernelArgumentHolder args =
+      KernelArgumentHolder::createKernelArgumentHolder(inputs, device);
+
+  // Step 2) Concretize fusion with input arguments.
+  std::unordered_map<Val*, Val*> symbolic_to_concrete_map =
+      DynamicTransform::concretizeFusion(segment_fusion_.get(), args);
+
+  // Step 3) Given the map_value_to_original_fid, the IRCloner returned by
+  // Fusion::copy, AND the symbolic_to_concrete map returned by
+  // concretization pass, create a mapping from cloned Vals to original fusion
+  // state indices.
+  std::transform(
+      map_value_to_original_fid.begin(),
+      map_value_to_original_fid.end(),
+      std::inserter(map_cloned_value_to_fid_, map_cloned_value_to_fid_.end()),
+      [&](const auto& item) {
+        const Val* original_value = item.first;
+        int64_t fid = item.second;
+        Val* cloned_val = original_to_cloned_map.clone(original_value);
+        if (symbolic_to_concrete_map.count(cloned_val)) {
+          cloned_val = symbolic_to_concrete_map.at(cloned_val);
+        }
+        return std::make_pair(cloned_val, fid);
+      });
+
+  // Track the extents for input TensorViews in cloned CPP Fusion.
+  cloned_extents_ = getExtents(segment_fusion_.get());
+
+  // Create runtime infomation
+  SchedulerRuntimeInfo runtime_info(
+      segment_fusion_.get(),
+      args,
+      /*precomputed_values=*/nullptr,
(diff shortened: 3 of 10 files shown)


rdspring1:
!test

csrc/python_frontend/fusion_state.cpp
```diff
@@ -68,6 +68,27 @@ std::ostream& operator<<(std::ostream& os, const State& state) {
   return os;
 }
 
+std::vector<Val*> getExtents(Fusion* fusion) {
```
rdspring1, line 71: Moved `getExtents` out of `FusionState`, so it is accessible in `SegmentationState`.

csrc/python_frontend/segmentation.cpp
```diff
@@ -0,0 +1,146 @@
...
+      std::move(segment_fusion_), &args, runtime_info);
+
+  // Get the order for fusion segments
+  prepareGroupOrder();
+
+  // Return the number of segments created by segmentation algorithm.
+  return (int64_t)segmented_fusion_->groups().size();
+}
+
+void SegmentationState::prepareGroupOrder() {
```
rdspring1, line 82: `prepareGroupOrder` is like [prepareRuntimeOrder](https://github.com/NVIDIA/Fuser/blob/main/csrc/runtime/fusion_cache_utils.h#L42-L44).

Differences:
* It does not track `group_extent_binding_order`, which is the pe-determined order to bind tensor input metadata.

csrc/python_frontend/segmentation.h
```diff
@@ -0,0 +1,221 @@
...
+
+#include <python_frontend/fusion_definition.h>
+#include <python_frontend/fusion_state.h>
+#include <runtime/fusion_executor_cache.h>
+
+namespace nvfuser::python_frontend {
+
+class FusionDefinition;
+
+//! Overview:
```
rdspring1, line 18: I'd start reviewing from the description here.

csrc/python_frontend/segmentation.h
```diff
@@ -0,0 +1,221 @@
...
+//! Segmentation decomposes a fusion into a directed acyclic graph (DAG) of
+//! sub-fusions. After applying the segmentation algorithm, we can translate
+//! the sub-fusions into their corresponding python definitions. Then, given the
+//! fusion's input arguments, the segments are run in the correct order to
+//! produce the output results.
+//!
+//! The original FusionDefinition stores the sequence of sub-fusions and acts as
+//! an argument manager. It gathers the input arguments before running the
+//! sub-fusion and stores its results. To perform this function, it requires a
+//! map from the segment index space to the original index space. This mapping
```
Priya2698, line 28: IIUC, the segment index space and original index space refer to the input indices?
Can we specify that since we also have the indices for the segments/sub-fusions too?

csrc/python_frontend/segmentation.h
```diff
@@ -0,0 +1,221 @@
...
+//!     S4 = fd.define_scalar(3, dtype=DataType.Int)
+//!     S5 = fd.define_scalar(100, dtype=DataType.Int32)
+//!     S6 = fd.define_scalar(1, dtype=DataType.Int32)
+//!     T7 = fd.ops.iota(S4, S5, S6, dtype=DataType.Int32)
+//!     fd.add_output(T3)
+//!     fd.add_output(T7)
+//!
+//! After Segmentation:
+//! The original fusion is divided into two segments. There is no dependencies
+//! between either segment so they can run in an order.
```
Priya2698, line 60:
```Suggestion
//! between either segment so they can run in any order.
```

csrc/python_frontend/segmentation.cpp
Priya2698: naive question: Is it possible to just reuse `prepareRuntimeOrder` then?

csrc/python_frontend/segmentation.h
```diff
@@ -0,0 +1,221 @@
...
+  //  8)     Add SegmentedGroup to group_run_order_. Mark all outputs of
+  //         SegmentedGroup as ready.
+  //  9)   End For
+  //  10)  Fail if none of the SegmentedGroups are available to run.
+  //  11) End While
+  void prepareGroupOrder();
+
+ private:
+  // Clone of original fusion for segmentation
+  std::unique_ptr<Fusion> segment_fusion_ = nullptr;
```
Priya2698, line 203: Would `cloned_fusion_` be a better name here, this is confusing since we also have `segmented_fusion_`?

csrc/python_frontend/segmentation.cpp
Priya2698: naive question: Is it possible then to reuse `prepareRuntimeOrder`?

csrc/python_frontend/segmentation.cpp
```diff
@@ -0,0 +1,146 @@
...
+      const std::vector<Val*>& group_outputs = group->outputs();
+      for (size_t group_out_i : c10::irange(group_outputs.size())) {
+        available_input.insert(group_outputs.at(group_out_i));
+      }
+      group_ran[group_i] = true;
+      ran_any_group = true;
+    }
+    NVF_ERROR(
+        ran_any_group,
+        "Failed to run all groups; An error must have occured in segmentation.");
```
Priya2698, line 142:
```Suggestion
"Failed to run any groups; An error must have occured in segmentation.");
```

csrc/python_frontend/segmentation.h
```diff
@@ -0,0 +1,221 @@
...
+  //  1) Clone preschedFusion CPP Fusion.
+  //  2) Concretize fusion using input arguments.
+  //  3) Given the map_value_to_original_fid, the IRCloner returned by
+  //     Fusion::copy, AND symbolic_to_concrete map returned by concretization
+  //     pass, create a mapping from cloned Vals to original fusion state
+  //     indices
+  //  4) Get extents for cloned fusion
+  //  5) Create SchedulerRuntimeInfo
+  //  6) Run segmentation algorithm using cloned fusion, input arguments, and
+  //  scheduler runtime infomation.
```
jjsjann123, line 175:
```suggestion
  //  scheduler runtime information.
```

csrc/python_frontend/segmentation.h
```diff
@@ -0,0 +1,221 @@
...
+  //  4) While not all the SegmentedGroups are run:
+  //  5)   For each SegmentedGroup:
+  //  6)     Skip SegmentedGroup if it is already run
+  //  7)     Skip SegmentedGroup if inputs are not ready
+  //  8)     Add SegmentedGroup to group_run_order_. Mark all outputs of
+  //         SegmentedGroup as ready.
+  //  9)   End For
+  //  10)  Fail if none of the SegmentedGroups are available to run.
+  //  11) End While
+  void prepareGroupOrder();
```
jjsjann123, line 199: nitpick: does `prepareGroupOrder()` belong in the public? Looks like it's only used inside `setupSegmentation`

nvfuser/__init__.py
```diff
@@ -54,6 +54,10 @@ def __init__(self, id=None, max_length=1024):
         super(FusionDefinition, self).__init__(id, max_length)
         self.profiled = False
 
+    def segment(self, inputs):
+        self.num_segments = self._setup_segmentation(inputs)
+        self._finalize_segmentation()
```
jjsjann123, line 59:
```suggestion
        num_segments = self._setup_segmentation(inputs)
        self._finalize_segmentation()
        return num_segments
```
nitpick: since segment isn't stateful. we probably don't want to keep `num_segments` inside self.

tests/python/test_python_frontend.py
```diff
@@ -1244,6 +1244,28 @@ def fusion_func(fd: FusionDefinition):
...
+            fd.add_output(t3)
+
+        nvf_out1, _ = self.exec_nvfuser(fusion_func, inputs)
+        eager_out = torch.sum(inputs[0], dim=-1, keepdim=True) + inputs[1]
+        self.assertEqual(eager_out, nvf_out1[0])
+
+        with FusionDefinition() as fd:
+            fusion_func(fd)
+        fd.segment(inputs)
+        assert fd.num_segments == 2
```
jjsjann123, line 1267:
```suggestion
        assert fd.segment(inputs) == 2
```

tests/python/test_python_frontend.py
```diff
@@ -3881,7 +3903,7 @@ def fusion_func(fd: FusionDefinition) -> None:
 
             fd.add_output(T88)
 
-        nvf_out, _ = self.exec_nvfuser(fusion_func, inputs, is_clonable=True)
```
jjsjann123, line 3884: IIUC, the default behavior is `is_clonable = True`, so this is just a clean up?!

csrc/python_frontend/fusion_definition.cpp
```diff
@@ -673,4 +674,19 @@ std::vector<std::pair<double, double>> FusionDefinition::getValTolerances(
...
+      segmentation_state_ == nullptr, "SegmentationState already exists!");
+  segmentation_state_ = std::make_unique<SegmentationState>();
+  return segmentation_state_->setupSegmentation(
+      preschedFusion(), map_value_to_fid_, inputs);
+}
+
+void FusionDefinition::finalizeSegmentation() {
+  // Destroy SegmentedState
+  segmentation_state_.reset();
+}
```
jjsjann123, line 690: a design question. What's the point of separating this into two steps?

Doesn't look like anything else is happening between `setupSegmentation()` to `finalizeSegmentation()`.

csrc/python_frontend/segmentation.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+void SegmentationState::prepareGroupOrder() {
+  NVF_ERROR(segmented_fusion_ != nullptr);
+
+  // Gather initial inputs for SegmentedFusion.
+  std::unordered_set<Val*> available_input;
+  std::copy(
+      segmented_fusion_->inputs().begin(),
+      segmented_fusion_->inputs().end(),
+      std::inserter(available_input, available_input.end()));
```
jjsjann123, line 90: nitpick: is there any benefit from std::copy other than initializing directly with this?
```suggestion
  std::unordered_set<Val*> available_input(segmented_fusion_->inputs().begin(), segmented_fusion_->inputs().end());
```

nvfuser/__init__.py
rdspring1: `num_segments` is used in the `test_segmentation_reduction_pointwise_epilogue`. We don't have anything else to check in this PR.

rdspring1: Ah, I can just check the return value.

csrc/python_frontend/fusion_definition.cpp
rdspring1: https://github.com/NVIDIA/Fuser/pull/3335 adds the `buildSegmentation()` to translate the CPP segments to their corresponding python definition.

We still need to destroy the `segmentation_state_` in this PR.

csrc/python_frontend/segmentation.h
rdspring1:
``` cpp
//! Each FusionDefinition contains a set of states representing tensors, vectors
//! and scalars. Every state has a unique index, which matches the insertion
//! order of the state in the FusionDefinition. These indices form a linear
//! index space for each FusionDefinition.
```

rdspring1:
!test

csrc/python_frontend/segmentation.cpp
rdspring1: Maybe, but I prefer to keep them separate for now.

This is to be a jumping off point for improving the segmentation algorithm.

Priya2698 (approved):
LGTM.

PR #3335, opened 2024-11-03
Translate segments to python definition

rdspring1:
## Overview:

-  `buildSegment` creates the CPP Fusion for a given segment id, translates it to a python FusionDefinition, then returns a mapping from the segment fusion state indices to the original fusion state indices.
- `FusionDefinition.segment` calls `setupSegmentation`, `buildSegment`, and `finalizeSegmentation` to create python definitions for the sub-fusions and their index mappings.

## Changes in this PR

This PR implements `buildSegment` function for user-scheduler segmentation. It is the second PR in a stack, preceded by https://github.com/NVIDIA/Fuser/pull/3334 and followed by https://github.com/NVIDIA/Fuser/pull/3025.

1. Implement `buildSegment` function in `csrc/python_frontend/segmentation.cpp`.
2. Complete `segment` function in `nvfuser/__init__.py`

```python
                          contiguity=[True, True],
                          dtype=DataType.Float,
                          is_cpu=False)
                          contiguity=[True, True],
                          dtype=DataType.Float,
                          is_cpu=False)
    T4 = fd.ops.add(T1, T3)
    fd.add_output(T4)
```

**After Segmentation:** The reduction scheduler does not support fusing any operations with an inner reduction, so the original fusion is divided into two segments.

## First Segment:
The first segment contains the reduction and broadcast operations, which corresponds with [T0, T2, T3] in the original fusion. Therefore, the segment index to original index map has two entries.

| Segment Index | Original Index | Description |
| -----------------| ---------------  | ------------- |
| T0 | T0 | The first tensor argument for the original fusion. |
| T2 | T3 | The broadcasted, reduction tensor is this segment's output. |

```python
def nvfuser_fusion_id2(fd : FusionDefinition) -> None :
                         contiguity=[True, True],
                         dtype=DataType.Float,
                         is_cpu=False)
   T1 = fd.ops.sum(T0, dims=[1], keepdim=False, dtype=DataType.Float)
   T2 = fd.ops.broadcast(T1, is_broadcast_dim=[False, True])
   fd.add_output(T2)
```
## Second Segment:
The second segment is the pointwise addition with the broadcasted reduction. It corresponds with [T1, T3, T4] in the original fusion.

| Segment Index | Original Index | Description |
| -----------------| ---------------  | ------------- |
| T0 | T1 | The second tensor argument for the original fusion. |
| T1 | T3 | The broadcasted, reduction tensor, which is the output from the first segment. |
| T2 | T4 | The pointwise addition, which is the output for the original fusion. |

```python
def nvfuser_fusion_id3(fd : FusionDefinition) -> None :
                         contiguity=[True, True],
                         dtype=DataType.Float,
                         is_cpu=False)
   T1 = fd.define_tensor(shape=[-1, 1],
                         contiguity=[True, None],
                         dtype=DataType.Float,
                         is_cpu=False)
   T2 = fd.ops.add(T0, T1)
   fd.add_output(T2)
```

### DIFF
--- a/csrc/python_frontend/fusion_definition.cpp
+++ b/csrc/python_frontend/fusion_definition.cpp
@@ -696,6 +696,16 @@ int64_t FusionDefinition::setupSegmentation(
       preschedFusion(), map_value_to_fid_, inputs);
 }
 
+std::unordered_map<int64_t, int64_t> FusionDefinition::buildSegment(
+    FusionDefinition& segment_fd,
+    int64_t segment_id) {
+  NVF_CHECK(id().has_value(), "FusionDefinition does not exist!");
+  NVF_CHECK(
+      segmentation_state_ != nullptr,
+      "Run setupSegmentation first before trying to build segments!");
+  return segmentation_state_->buildSegment(segment_fd, segment_id);
+}
+
 void FusionDefinition::finalizeSegmentation() {
   // Destroy SegmentedState
   segmentation_state_.reset();
--- a/csrc/python_frontend/segmentation.cpp
+++ b/csrc/python_frontend/segmentation.cpp
@@ -12,71 +12,283 @@ namespace nvfuser::python_frontend {
 
 int64_t SegmentationState::setupSegmentation(
     Fusion* fusion,
-    const std::unordered_map<const Val*, int64_t>& map_value_to_original_fid,
+    const std::unordered_map<const Val*, int64_t>&
+        map_presched_value_to_original_python_index,
     const at::ArrayRef<c10::IValue>& inputs) {
   // Check state
   NVF_ERROR(fusion != nullptr);
-  NVF_ERROR(cloned_fusion_ == nullptr);
+  NVF_ERROR(cloned_original_fusion_ == nullptr);
   NVF_ERROR(segmented_fusion_ == nullptr);
   NVF_ERROR(group_run_order_.empty());
-  NVF_ERROR(map_cloned_value_to_fid_.empty());
-  NVF_ERROR(cloned_extents_.empty());
+  NVF_ERROR(map_cloned_concretized_value_to_original_python_index_.empty());
+  NVF_ERROR(cloned_original_extents_.empty());
 
   int8_t device = getCommonDeviceCUDA(inputs);
   NVF_CHECK(
       inputs.empty() || device > -1, "Inputs are not all on the same device!");
 
   // Step 1) Clone preschedFusion CPP Fusion.
-  cloned_fusion_ = std::make_unique<Fusion>();
+  cloned_original_fusion_ = std::make_unique<Fusion>();
 
   // The IRCloner returned by Fusion::copy acts as map from the original fusion
   // to the cloned fusion.
-  IrCloner original_to_cloned_map = Fusion::copy(fusion, cloned_fusion_.get());
+  IrCloner original_to_cloned_map =
+      Fusion::copy(fusion, cloned_original_fusion_.get());
 
   KernelArgumentHolder args =
       KernelArgumentHolder::createKernelArgumentHolder(inputs, device);
 
   // Step 2) Concretize fusion with input arguments.
   std::unordered_map<Val*, Val*> symbolic_to_concrete_map =
-      DynamicTransform::concretizeFusion(cloned_fusion_.get(), args);
+      DynamicTransform::concretizeFusion(cloned_original_fusion_.get(), args);
 
-  // Step 3) Given the map_value_to_original_fid, the IRCloner returned by
-  // Fusion::copy, AND the symbolic_to_concrete map returned by
+  // Step 3) Given the map_presched_value_to_original_python_index, the IRCloner
+  // returned by Fusion::copy, AND the symbolic_to_concrete map returned by
   // concretization pass, create a mapping from cloned Vals to original fusion
   // state indices.
   std::transform(
-      map_value_to_original_fid.begin(),
-      map_value_to_original_fid.end(),
-      std::inserter(map_cloned_value_to_fid_, map_cloned_value_to_fid_.end()),
+      map_presched_value_to_original_python_index.begin(),
+      map_presched_value_to_original_python_index.end(),
+      std::inserter(
+          map_cloned_concretized_value_to_original_python_index_,
+          map_cloned_concretized_value_to_original_python_index_.end()),
       [&](const auto& item) {
         const Val* original_value = item.first;
-        int64_t fid = item.second;
+        int64_t python_index = item.second;
         Val* cloned_val = original_to_cloned_map.clone(original_value);
         if (symbolic_to_concrete_map.count(cloned_val)) {
           cloned_val = symbolic_to_concrete_map.at(cloned_val);
         }
-        return std::make_pair(cloned_val, fid);
+        return std::make_pair(cloned_val, python_index);
       });
 
   // Track the extents for input TensorViews in cloned CPP Fusion.
-  cloned_extents_ = getExtents(cloned_fusion_.get());
+  cloned_original_extents_ = getExtents(cloned_original_fusion_.get());
 
   // Create runtime infomation
   SchedulerRuntimeInfo runtime_info(
-      cloned_fusion_.get(),
+      cloned_original_fusion_.get(),
       args,
       /*precomputed_values=*/nullptr,
-      cloned_fusion_->allTvs());
+      cloned_original_fusion_->allTvs());
 
   // Run segmentation algorithm
   segmented_fusion_ = SegmentCandidateFinder::segment(
-      std::move(cloned_fusion_), &args, runtime_info);
+      std::move(cloned_original_fusion_), &args, runtime_info);
 
   // Get the order for fusion segments
   prepareGroupOrder();
 
   // Return the number of segments created by segmentation algorithm.
   return (int64_t)segmented_fusion_->groups().size();
+}
+
+// setupSegmentation transforms the Prescheduled, Symbolic Fusion to Cloned,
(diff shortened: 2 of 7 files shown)


rdspring1:
!test

Priya2698:
I am seeing changes from PR #3334, can you rebase to only include changes from this PR for easier review?

rdspring1:
!test

jjsjann123:
Oops. I think the merge of #3334 messed up the git history. You might have to resolve the conflicts by hand now.

rdspring1:
!test

rdspring1:
I used `git rebase` to fixed the conflicts.

csrc/python_frontend/segmentation.cpp
```diff
@@ -79,6 +79,167 @@ int64_t SegmentationState::setupSegmentation(
...
+    segment_fid_to_original_fid_map.emplace(
+        segment_fid.at(idx), original_fid.at(idx));
+  }
+
+  // Step 4e) short-circuit: No extra extents required for python definition.
+  if (fusion_segment->inputs().size() == other.inputs().size()) {
+    return segment_fid_to_original_fid_map;
+  }
+
+  // The python segment can require the size of tensor dimensions from original
```
jjsjann123, line 172: I have to say that I am quite lost on how the mapping were done in the following section.

iiuc, there are four things that we are trying to map.

1. original Fusion.
2. Fusion segment.
3. original FusionDefinition.
4. FusionDefinition for the segment that we populate in this function.

I find that the use of `translated_xxx` and `cloned_xxx` in the variable name a bit confusing (this could just be me).

csrc/python_frontend/segmentation.cpp
```diff
@@ -79,6 +79,167 @@ int64_t SegmentationState::setupSegmentation(
...
+    if (map_translated_val_to_segment_fid.count(segment_extent) == 0) {
+      continue;
+    }
+
+    size_t segment_fid = map_translated_val_to_segment_fid.at(segment_extent);
+    segment_to_cloned_extents.emplace(segment_extent, cloned_extent);
+    segment_fid_to_translated_val.emplace(segment_fid, segment_extent);
+  }
+
+  // Step 4h) Find the set difference between all segment input indices and
```
jjsjann123, line 192: I see the point of having the two map above.
But from step 4h) -> 4l), we are just doing a chained mapping. Do we really want to have it done in such an explicit way? other than just do the mapping in

```
  for (int64_t input_fid : other.inputs()) {
    if (segment_fid_to_original_fid_map.count(input_fid) == 0) {
      segment_fid_to_original_fid_map.emplace(input_fid, magic_chained_map(input_fid));
    }
  }
```

csrc/python_frontend/segmentation.cpp
rdspring1: The mapping is from state indices for the segment to state indices of the original fusion.
e.g., From number 4 to number 3.

`translated_xxx` corresponds with the segment CPP fusion. `cloned_xxx` corresponds with the clone of the original CPP fusion.

csrc/python_frontend/segmentation.cpp
rdspring1: The `magic_chained_map` can represented as a function to do the following:
`segment index -> segment val -> cloned val -> original index`

csrc/python_frontend/segmentation.h
```diff
@@ -185,6 +185,43 @@ class SegmentationState {
...
+  //    f) Create a map from segment to cloned extents.
+  //    g) Create a map from segment fusion indices to cloned extents.
+  //    h) Find segment inputs that are missing from segment to original
+  //       indices map.
+  //    i) Get segment Vals for the missing segment fusion indices.
+  //    j) Map segment Vals to cloned Vals.
+  //    k) Map cloned Vals to their corresponding fusion indices.
+  //    l) Add missing mappings to segment to original indices map.
+  //  5) Return the mapping from the segmented FusionDefinition index space to
+  //  original FusionDefinition index space.
```
jjsjann123, line 220: my nitpick on the comment along with the implementation.

It felt quite a bit verbose on how the mapping is done. At least I struggle to follow that. I think a higher level description would be useful.

i.e. how the map goes from `segmented_fusion_definition -> fusion_segment -> original_fusion -> original_fusion_definition`. I think it would also help if we try to rename variables in the implementation to make it easy to see how each map it projecting in the map above.

jjsjann123 (approved):
I don't have a strong opinion against the implementation. But I'm rather nitpicking on the implementation/comments.
So I'm not pushing for fixing everything in this PR, given that it's just an implementation detail that isn't necessarily exposed to the user.

csrc/python_frontend/fusion_definition.cpp
```diff
@@ -701,6 +701,16 @@ int64_t FusionDefinition::setupSegmentation(
       preschedFusion(), map_value_to_fid_, inputs);
 }
 
+std::unordered_map<int64_t, int64_t> FusionDefinition::buildSegment(
+    FusionDefinition& other,
+    int64_t segment_id) {
+  NVF_CHECK(id().has_value(), "FusionDefinition definition does not exist!");
+  NVF_CHECK(
+      segmentation_state_ != nullptr,
+      "Run setupSegmenation first before trying to build segments!");
```
Priya2698, line 710:
```Suggestion
      "Run setupSegmentation first before trying to build segments!");

File: tests/python/test_python_frontend.py [hunk 5]
```diff
@@ -1263,7 +1263,16 @@ def fusion_func(fd: FusionDefinition):
...
         with FusionDefinition() as fd:
             fusion_func(fd)
-        assert fd.segment(inputs) == 2
+
+        # create segments
+        fd.segment(inputs)
+
+        # Each segment has a map from its segment index space to the original
+        # fusion's index space. The original fusion creates two segments.
+        # Check that segment maps match expected behavior.
```
Priya2698, line 1272: Can you list the two segments here, so that it is easier to compare the maps below?

nvfuser/__init__.py
```diff
@@ -56,9 +56,41 @@ def __init__(self, id=None, max_length=1024):
...
+            List[FusionDefinition]: The FusionDefinitions corresponding to the
+            sub-fusion segments of this FusionDefinition.
+        """
         num_segments = self._setup_segmentation(inputs)
+        if num_segments == 1:
+            self._finalize_segmentation()
+            return []
+
+        self.segments = []
+        self.segment_maps = []
```
Priya2698, line 79: `self.segment_maps` is not clear to me on what it maps. Can we rename it to `self.segment_to_original_fid_map`?

csrc/python_frontend/segmentation.cpp
```diff
@@ -79,6 +79,167 @@ int64_t SegmentationState::setupSegmentation(
   return (int64_t)segmented_fusion_->groups().size();
 }
 
+std::unordered_map<int64_t, int64_t> SegmentationState::buildSegment(
+    FusionDefinition& other,
+    int64_t segment_id) {
+  NVF_ERROR(
+      !other.completed(),
+      "Expected an incomplete definition before translation.");
```
Priya2698, line 87: I am missing what `.completed` returns here. Is it that the execution is complete?

csrc/python_frontend/segmentation.h
```diff
@@ -185,6 +185,43 @@ class SegmentationState {
...
+  //  1) Use segment id to get SegmentedGroup from group_run_order_.
+  //  2) Create CPP Fusion for SegmentedGroup.
+  //  * IrCloner acts as a map from fusion segment to the original fusion.
+  //  3) Translate CPP Fusion to Python FusionDefinition
+  //  4) Create map from segment fusion indices to original fusion indices.
+  //    a) Get cloned Vals for SegmentedGroup's inputs and outputs. Map them
+  //       to their original fusion indices.
+  //    b) Map cloned Vals to their segment Vals
+  //    c) Map segment Vals to their fusion indices.
+  //    d) Map original indices to segment indices.
```
Priya2698, line 208: As @jjsjann123 suggested below, this detail will be more helpful in the CPP file to follow the mapping.

nvfuser/__init__.py
```diff
@@ -56,9 +56,41 @@ def __init__(self, id=None, max_length=1024):
...
+        self.last_used_segment = {}
```
Priya2698, line 80: Please add a comment on what this dictionary is storing. The naming is not descriptive of the contents to me.

tests/python/test_python_frontend.py
Priya2698: Additionally, it might be helpful to compare the expected python definitions for the segments similar to `test_repro_script_generation`. Wdyt?

Priya2698 (approved):
LGTM overall.

I mainly have requests around naming and test.
Thanks for the detailed comments, very helpful in understanding your implementation. Great work!

csrc/python_frontend/segmentation.cpp
rdspring1: changed from `other` to `segment_fd`

nvfuser/__init__.py
rdspring1: Added ``` # Track the last segment a value is used as an input ```

nvfuser/__init__.py
rdspring1: changed to `self.segment_index_space_maps`

tests/python/test_python_frontend.py
rdspring1: I figured you'd ask for a string comparison, but they can be brittle and annoying to maintain. We only care about matching correctness which the next PR will enforce.

rdspring1:
!test

rdspring1:
I renamed some variables to make things clearer. I hope it helps!!!

PR #3338, opened 2024-11-04
Sequence Parallel Forward Transformer

cowanmeg:
Sequence parallel forward transformer layer and multi-headed attention tests. 

1. Cleans up sharding annotations in Forward fusion definitions. Only sharding changes and inputs are explicitly sharded.
2. Updates output of mha and mlp to be a struct with named TVs to make code more readable. 
3. Dropout probability is temporarily set to 0. This will be fixed in a later PR to use philox seed and offset with validation.

### DIFF
--- a/tests/cpp/test_multidevice_transformer.cpp
+++ b/tests/cpp/test_multidevice_transformer.cpp
@@ -315,13 +315,13 @@ MLPResult mlp(
     local_matmul1 = reshape(local_matmul1, {D, B * S, E}, {D, D, B * S / D, E});
   }
   TensorView* matmul1 = sum(local_matmul1, {0}); // Allreduce or Reduce scatter
-  std::vector<bool> bcast_mask(matmul1->nDims()-1, true);
-  bcast_mask[matmul1->nDims()-2] = false;
+  std::vector<bool> bcast_mask(matmul1->nDims() - 1, true);
+  bcast_mask[matmul1->nDims() - 2] = false;
   TensorView* linear1 = add(matmul1, broadcast(b1, bcast_mask));
   // Dropout
   Val* prob = IrBuilder::create<Val>(1.0 - kDropoutProb);
   Val* scale = IrBuilder::create<Val>(1.0 / (1.0 - kDropoutProb));
-  TensorView* dropout_result =  dropout(linear1, prob, scale).output;
+  TensorView* dropout_result = dropout(linear1, prob, scale).output;
 
   // Tensor parallel shardings
   for (auto* tv : {w0, b0, w1}) {
@@ -356,7 +356,7 @@ MHAResult mha(
     TensorView* w1,
     TensorView* b1,
     const DeviceMesh& mesh,
-    bool sequence_parallel=false) {
+    bool sequence_parallel = false) {
   const auto D = w0->axis(0)->extent()->value().as<int64_t>();
   auto dtype = w0->dtype();
 
@@ -366,7 +366,8 @@ MHAResult mha(
     x->axis(0)->parallelize(ParallelType::DIDx);
     x = set(x); // allgather
     x->axis(0)->parallelize(ParallelType::Serial);
-    // Reshape is uncessary, it is here to keep shapes with TP and TP+SP the same for validation.
+    // Reshape is uncessary, it is here to keep shapes with TP and TP+SP the
+    // same for validation.
     x = reshape(x, {D, B * S / D, E}, {B * S, E});
   }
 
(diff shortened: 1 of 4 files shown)


cowanmeg:
!build

samnordmann (approved):

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -307,50 +315,62 @@ std::vector<TensorView*> mlp(
     local_matmul1 = reshape(local_matmul1, {D, B * S, E}, {D, D, B * S / D, E});
   }
   TensorView* matmul1 = sum(local_matmul1, {0}); // Allreduce or Reduce scatter
-  TensorView* linear1 = add(matmul1, broadcast(b1, {true, true, false}));
+  std::vector<bool> bcast_mask(matmul1->nDims() - 1, true);
+  bcast_mask[matmul1->nDims() - 2] = false;
```
wujingyue, line 319: Why -2? I could be wrong but I think `matmul1` ends at E not a reduction dimension.

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -307,50 +315,62 @@ std::vector<TensorView*> mlp(
...
-      tv->axis(0)->parallelize(ParallelType::DIDx);
-    }
     matmul1->setDeviceMesh(mesh);
     matmul1->axis(1)->parallelize(ParallelType::DIDx);
-    mask->setDeviceMesh(mesh);
-    mask->axis(0)->parallelize(ParallelType::DIDx);
   }
 
-  return {linear0, gelu, linear1, dropout_result};
+  return {linear0, gelu, matmul1, linear1, dropout_result};
```
wujingyue, line 341: I'm unsure about saving both `matmul1` and `linear1`. Is that a common practice, e.g., done by TE?

wujingyue (commented):
Thanks! I'm still reviewing the MHA part...

tests/cpp/test_multidevice_transformer.cpp
wujingyue: (I wasn't sure because linear1 could be cheaply rematerialized from matmul1 and b1 -- almost the same compute but much less memory)

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -1074,11 +1305,11 @@ TEST_P(DistributedTransformerTest, Forward) {
   auto ln_input = castOp(DataType::Float, x);
   auto ln0 = layer_norm(ln_input, norm_shape, ln0_w, ln0_b, eps);
   auto mha_in = castOp(dtype, ln0.output);
-  auto mha_out = mha(mha_in, mha_w0, mha_b0, mha_w1, mha_b1, mesh)[3];
+  auto mha_out = mha(mha_in, mha_w0, mha_b0, mha_w1, mha_b1, mesh).output;
```
wujingyue, line 1308: 👍

tests/cpp/test_multidevice_transformer.cpp
cowanmeg: The fusion does not output matmul1, it's returned by the helper function for sharding propagation, since matmul1 is a boundary tv for reduce-scatter in sequence parallel.

tests/cpp/test_multidevice_transformer.cpp
```diff
@@ -307,50 +315,62 @@ std::vector<TensorView*> mlp(
     local_matmul1 = reshape(local_matmul1, {D, B * S, E}, {D, D, B * S / D, E});
   }
   TensorView* matmul1 = sum(local_matmul1, {0}); // Allreduce or Reduce scatter
```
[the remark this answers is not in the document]
cowanmeg: See other comment, but the different shape not an issue since matmul1 is not validated against the reference!

tests/cpp/test_multidevice_transformer.cpp
cowanmeg: hmm...yeah that looks like it should be -1...I'm surprised it worked?

cowanmeg: oh ... it's because the number of logical dimensions is matmul1->nDims() -1 because there is one reduction axis... so the index of the last one is -2

wujingyue (approved):

cowanmeg:
!build

liqiangxl:
check `DistributedTransformerTest.MultiheadAttention_SP/__half`
!test

liqiangxl:
!test

PR #3343, opened 2024-11-05
Avoid replacing a Val with a dependent Val

naoyam:
It would result in a recursive definition.


naoyam:
!test

csrc/ir/utils.cpp
```diff
@@ -1204,6 +1204,40 @@ bool isFunctional(const Val* v) {
   return std::all_of(def->inputs().begin(), def->inputs().end(), isFunctional);
 }
 
+bool isRecursivelyDefined(Val* val) {
```
naoyam, line 1207: Ended up not using this, but was useful while debugging.

PR #3344, opened 2024-11-05
Prefer simpler vals in replace sizes

naoyam:
Noticed while working on #3309 that `i0` is replaced with `ceilDiv(i0, 1)`. While it isn't incorrect, it would make generated code look simpler if `ceilDiv(i0, 1)` is replaced with `i0`. 

This PR just changes representative iter domains used for replacing extents. In addition to the existing priority rules, the iter domain with the simplest extent is preferred as the representative ID of a given ID group. The simplicity of extents is just defined based on the number of expressions defining the extent val. So, for example, an iter domain with extent of `i0` should be used as the representative ID instead of iter domains with extent `ceilDiv(i0, 1)`.

There should be no logic change.

### DIFF
--- a/csrc/device_lower/pass/replace_size.cpp
+++ b/csrc/device_lower/pass/replace_size.cpp
@@ -52,25 +52,6 @@ std::unordered_map<Val*, Val*> getSimplificationMap(Fusion* fusion) {
 
   std::unordered_map<Val*, Val*> simplification_map;
 
-  auto get_number_of_defs = [](Val* val) -> int64_t {
-    std::deque<Val*> vals;
-    vals.push_back(val);
-    int64_t num_defs = 0;
-    while (!vals.empty()) {
-      auto v = vals.front();
-      vals.pop_front();
-      auto def = v->definition();
-      if (def == nullptr) {
-        continue;
-      }
-      ++num_defs;
-      for (auto inp : def->inputs()) {
-        vals.push_back(inp);
-      }
-    }
-    return num_defs;
-  };
-
   for (const ValGroup& group : graph.disjointValSets().disjointSets()) {
     // For each ValGroup, find a single extent to use for all extents of
     // IterDomains in the group. These are chosen in descending order of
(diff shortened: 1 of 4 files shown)


naoyam:
!build

csrc/device_lower/pass/replace_size.cpp
```diff
@@ -52,18 +52,40 @@ std::unordered_map<Val*, Val*> getSimplificationMap(Fusion* fusion) {
 
   std::unordered_map<Val*, Val*> simplification_map;
 
+  auto get_number_of_defs = [](Val* val) -> int64_t {
```
jacobhinkle, line 55: This might be useful in other contexts. Maybe we could make it a utility like `ir_utils::getOperationCount(Val*)`

csrc/device_lower/pass/replace_size.cpp
```diff
@@ -52,18 +52,40 @@ std::unordered_map<Val*, Val*> getSimplificationMap(Fusion* fusion) {
...
     // For each ValGroup, find a single extent to use for all extents of
     // IterDomains in the group. These are chosen in descending order of
     // preference:
     // 1. Constant ints. These might be non-immediate constants
     // 2. Extents of input TVs.
     // 3. Extents of non-input TVs.
-    // Within these three classes, we find the IterDomain with the smallest
-    // name().
+    // Within these three classes, we find the IterDomain with the
+    // smallest name(). For case 3, we also prefer the IterDomain with
```
jacobhinkle, line 82: I guess we could also do this for Case 1 just for readability, but it shouldn't affect the compiled kernel so not important..

jacobhinkle (approved):
LGTM

csrc/device_lower/pass/replace_size.cpp
naoyam: Yeah, I didn't do that just because `get_number_of_defs` may not be cheap.

naoyam:
!test

PR #3345, opened 2024-11-05
Ensure expression simplification is enabled in proveLinearAndGetStride

jacobhinkle:
When debugging the matmul scheduler, I noticed that we hit an error when generating Hopper MMA when we use `NVFUSER_DISABLE=expr_simplify`. This PR makes that possible again. Note that the option is disabled only within `proveLinearAndGetStride` so it has no direct effect on index expressions, which is what we want to look at when using that option.


jacobhinkle:
!test

zasdfgbnm (approved):

jacobhinkle:
I think the distributed test failures are benign.

PR #3346, opened 2024-11-05
replaceSymbolicSizes

naoyam:
can't repro but this seems to be the right thing to do

Update: added a standalone repro


naoyam:
!test --diff

naoyam:
!test --diff-bench

naoyam:
@jacobhinkle As I mentioned offline, there's one test that had register usage jump, but I still think this is the right thing to do. I started the diff check with the benchmarks as well.

jacobhinkle (approved):
LGTM

naoyam:
@jacobhinkle No codegen change with the benchmarks.

Issue #2718, opened 2024-07-30
benchmarking thunder.jit


liqiangxl:
Current python bechmark uses a pre-defined fusion definition which was generated by `thunder.jit`, when `thunder.jit` changes, it can't directly refelect the change.
To solve this issue, we are considering directly use `thunder.jit` in python benchmarks. Current approach in #2717 
(1) Add a standalone test function, e.g. `def test_groupnorm_fwd_thunder_benchmark`, its sole purpose is benchmarking `thunder.jit`
(2) It runs nigthly & weekly
(3) may consider retire current pre-defined version ?

Other options:
Add `thunder.jit` to `test_groupnorm_fwd_baseline_benchmark`, e.g. extend it to
`@pytest.mark.parametrize("backend", ["eager", "compile", "thunder"])`
and runs only weekly

Priya2698:
The updated implementation for this task is:
1. Modify `compile` to `executor` which can be `['eager', 'torchcompile', 'thunder'].
2. The `thunder.jit` benchmarks will be appended to the baseline results and run optionally using `--benchmark-thunder` and in the weekly CI.
3. Finally, we will modify the benchmark infrastructure to run Thunder benchmarks by default instead of nvFuser and nvFuser benchmarks if thunder definition is not supported and host benchmarks.

Priya2698:
Closing this as we benchmark `thunder.jit` now.

PR #3350, opened 2024-11-05
Modify the `compile` parameter in baseline benchmarks to `executor`

Priya2698:
This PR is the first step in adding `thunder.jit` benchmarks.
The major change is modifying the `compile` parameter to `executor` with values `eager`, `torchcompile`, `thunder`. This PR does not introduce any new thunder benchmarks (to be done in next PR).

CC: @xwang233 for dashboard changes.

Issue: #2718

### DIFF
--- a/benchmarks/python/conftest.py
+++ b/benchmarks/python/conftest.py
@@ -96,45 +96,36 @@ def pytest_configure(config):
 
 def pytest_collection_modifyitems(session, config, items):
     """
-    The baseline benchmarks use `compile` parameter:
-        compile = false: Eager mode benchmark
-        compile = true: torch.compile benchmark
+    The baseline benchmarks use `executor` parameter with 
+    values ["eager", "torchcompile", "thunder"] that are optionally
+    run using `--benchmark-{executor}` flag. They are skipped by 
+    default.
     """
-    run_eager = config.getoption("--benchmark-eager")
-    run_thunder = config.getoption("--benchmark-thunder")
-    run_torchcompile = config.getoption("--benchmark-torchcompile")
 
     from nvfuser.pytorch_utils import retry_on_oom_or_skip_test
 
+    def get_test_executor(item) -> str | None:
+        if (
+            hasattr(item, "callspec")
+            and "executor" in item.callspec.params
+        ):
+            return item.callspec.params["executor"]
+        return None
+    
+    executors_to_skip = []
+
+    for executor in ["eager", "torchcompile", "thunder"]:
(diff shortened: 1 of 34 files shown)


benchmarks/python/test_broadcast_add_fwd.py
```diff
@@ -112,9 +112,14 @@ def test_bcast_add_baseline_benchmark(
         x = x.t()
     assert x.is_contiguous() == contiguous
 
+    benchmark_fn = {
+        "eager": bcast_add_fwd_fn,
+        "torchcompile": torch.compile(bcast_add_fwd_fn),
+    }
```
naoyam, line 118: Personally I'd avoid using pure strings as much as possible because it'd be difficult to debug when things don't work as intended. For example, I suppose you won't get any error even if you typed "torch.compile". 

Not a request to change anything, but just my two cents.

benchmarks/python/conftest.py
```diff
@@ -96,45 +96,33 @@ def pytest_configure(config):
...
-    run_eager = config.getoption("--benchmark-eager")
-    run_thunder = config.getoption("--benchmark-thunder")
-    run_torchcompile = config.getoption("--benchmark-torchcompile")
 
     from nvfuser.pytorch_utils import retry_on_oom_or_skip_test
 
+    def get_test_executor(item) -> str | None:
+        if hasattr(item, "callspec") and "executor" in item.callspec.params:
+            return item.callspec.params["executor"]
+        return None
```
liqiangxl, line 110: suggest adding a check ensure `executor` is one of `"eager", "torchcompile", "thunder"` to avoid typo, e.g. "torchcompile" -> "torch.compile"

benchmarks/python/test_broadcast_add_fwd.py
liqiangxl: we 've seen a similar typo err where `device` was typed as `degvice`

Priya2698: Agreed, I considered using enums, but the misspelling can cause errors with that too.
Here, if `torchcompile` is spelled as `torch.compile`,  in later lines, `benchmark_fn[executor]` will error out. So combined with the assertion that @liqiangxl suggested in the below comment for `get_test_executor`, we should be able to catch any spelling errors.

Please let me know if you have any suggestions which may be more error-proof.

naoyam (approved):
LGTM

Priya2698:
!buils

Priya2698:
!build

PR #3352, opened 2024-11-05
Move allocOutputSpace to multidevice/executor.cpp

wujingyue:
It's only used there and doesn't need to be exposed via headers.


wujingyue:
!build

zasdfgbnm (approved):

PR #3357, opened 2024-11-06
Set `fusion_id` and `device_id` of `KernelExecutor` in constructor for user scheduling.

rdspring1:
The goal is to set `fusion_id` and `device_id` when creating `KernelExecutor` for `UserSchedule. Previously, it was set during `FusionExecutor::compileFusion`. This PR is stacked on `executor_dispatch`

**Changes to `UserSchedule` cache system:**
**Current:** The map key is the integer value of input arguments. The vector is of size `device id`.
`std::unordered_map<size_t, std::vector<UserSchedule>> user_def_schedules;`

**New:** The key to first map is the integer value of input arguments. The key to second map is of `device`.

**Why?** We can set the the `fusion_id` and `device_id` in the constructor of `UserSchedule` and `KernelExecutor`.

### DIFF
--- a/csrc/python_frontend/fusion_cache.cpp
+++ b/csrc/python_frontend/fusion_cache.cpp
@@ -597,11 +597,6 @@ UserSchedule* FusionCache::createUserSchedule(
   auto& user_scheds = scheds->user_def_schedules;
   auto input_id = user_def_input_encodings_.lookupId(inputs);
 
-  // Create device to UserSchedule map for input arguments id
-  if (user_scheds.count(input_id.id) == 0) {
-    user_scheds[input_id.id] = std::unordered_map<int, UserSchedule>();
-  }
-
   // Create UserSchedule for device
   if (user_scheds[input_id.id].count(device) == 0) {
     user_scheds[input_id.id].emplace(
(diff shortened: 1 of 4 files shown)


rdspring1:
!test

csrc/python_frontend/fusion_cache.cpp
```diff
@@ -591,22 +596,25 @@ UserSchedule* FusionCache::createUserSchedule(
...
   auto& user_scheds = scheds->user_def_schedules;
   auto input_id = user_def_input_encodings_.lookupId(inputs);
-  auto user_sched = user_scheds.find(input_id.id);
-  if (user_sched == user_scheds.end()) {
-    user_scheds[input_id.id] = std::vector<UserSchedule>(device + 1);
+
+  // Create device to UserSchedule map for input arguments id
+  if (user_scheds.count(input_id.id) == 0) {
+    user_scheds[input_id.id] = std::unordered_map<int, UserSchedule>();
+  }
```
jjsjann123, line 603:
```suggestion
```

nitpick: given
```
  std::unordered_map<size_t, std::unordered_map<int, UserSchedule>>
      user_def_schedules;
```
this is just a default behavior. you don't need this.

jjsjann123 (approved):
OOC, Is this a stacked PR? since the target is `executor_dispatch`. Let's put the target PR in the description.

rdspring1:
!test

PR #3361, opened 2024-11-06
Use SimplyfingIrBuilder for circular buffering

naoyam:
Noticed some easy simplification while working on #3309. Not necessary for correctness and the actual perf would probably have no effect ,if any, thanks to expr simplification and index hoisting, but examining generated code could be easier when simplification is possible.

### DIFF
--- a/csrc/ir/nodes.cpp
+++ b/csrc/ir/nodes.cpp
@@ -4881,6 +4881,13 @@ bool ForLoop::isTrivial() const {
     return true;
   }
 
+  if (start()->isConstScalar() && simplifiedStop()->isConstScalar() &&
+      start()->evaluate().as<int64_t>() + 1 ==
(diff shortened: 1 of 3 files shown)


csrc/ir/nodes.cpp
```diff
@@ -4881,6 +4881,13 @@ bool ForLoop::isTrivial() const {
     return true;
   }
 
+  if (start()->isConstScalar() && simplifiedStop()->isConstScalar() &&
```
naoyam, line 4884: Because of the simplification, the pattern previously detected by the above condition, e.g., `for (int i = ceilDiv(32, 8) - 1; i < ceilDiv(32, 8); ++i)` may become `for (int = 3; i < 4; ++i)`, so this condition is necessary to detect that pattern too.

naoyam:
!test

rdspring1 (approved):
Thanks!

Issue #871, opened 2023-09-12
codegen error: index_map.find(alloc_dom[i]) != index_map.end() INTERNAL ASSERT FAILED at "/opt/pytorch/nvfuser/csrc/index_compute.cpp"


jjsjann123:
Repro python script
```
import torch
from nvfuser import FusionDefinition, DataType

def nvfuser_fusion_id0(fd : FusionDefinition) -> None :
    T0 = fd.define_tensor(shape=[1, -1, -1], contiguity=[None, True, True], dtype=DataType.Half, is_cpu=False)
    T6 = fd.ops.broadcast_in_dim(T0, shape=[1, 2, 5, 1], broadcast_dims=[0, 1, 2])
    T13 = fd.ops.broadcast_in_dim(T6, shape=[1, 2, 5, 1, 1], broadcast_dims=[0, 1, 2, 4])
    T14 = fd.ops.slice(T13, start_indices=[0, 0, 0, 0, 0], end_indices=[1, 2, 2, 1, 1], strides=[1, 1, 1, 1, 1])
    T15 = fd.ops.slice(T13, start_indices=[0, 0, 2, 0, 0], end_indices=[1, 2, 5, 1, 1], strides=[1, 1, 1, 1, 1])
    T23 = fd.ops.broadcast_in_dim(T14, shape=[1, 2, 2, 1, 1, 1], broadcast_dims=[0, 1, 2, 4, 5])
    T31 = fd.ops.broadcast_in_dim(T23, shape=[1, 2, 2, 2, 1, 1], broadcast_dims=[0, 1, 2, 3, 4, 5])
    T39 = fd.ops.broadcast_in_dim(T15, shape=[1, 2, 3, 1, 1, 1], broadcast_dims=[0, 1, 2, 4, 5])
    T47 = fd.ops.broadcast_in_dim(T39, shape=[1, 2, 3, 1, 1, 1], broadcast_dims=[0, 1, 2, 3, 4, 5])
    T48 = fd.ops.reshape(T31, new_shape=[1, 2, 4, 1, 1])
    T49 = fd.ops.set(T48)
    T50 = fd.ops.reshape(T47, new_shape=[1, 2, 3, 1, 1])
    T51 = fd.ops.set(T50)
    T52 = fd.ops.cat([T49, T51], dim=2)
    fd.add_output(T52)

with FusionDefinition() as fd:
    nvfuser_fusion_id0(fd)

inputs = [
[...]

PR #3368, opened 2024-11-07
Add NVFUSER_DUMP=python_definition_segments

jacobhinkle:
This adds a new debug dump option that simply prints out all of the segmented fusion segments as python FusionDefinitions at compile time. This is useful for debugging non-segmentation related errors whose repros contain many segments. To do so, when you notice a compile error the printout will tell you which segmented group it was found in. Then run your code again with `NVFUSER_DUMP=python_definition_segments` which will print a definition function for each segmented group using the C++ to python translation that was recently added. This lets you create a more targeted repro by copying that smaller definition into the repro printed in the error message then updating the inputs so that it executes.

### DIFF
--- a/csrc/options.cpp
+++ b/csrc/options.cpp
@@ -135,6 +135,7 @@ std::unordered_map<DebugDumpOption, std::vector<std::string>> Options<
       {"ptx", DebugDumpOption::Ptx},
       {"ptxas_verbose", DebugDumpOption::PrintPtxasLog},
       {"python_definition", DebugDumpOption::PythonDefinition},
+      {"python_definition_segments", DebugDumpOption::PythonDefinitionSegments},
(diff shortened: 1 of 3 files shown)


jacobhinkle:
!build

jacobhinkle:
I used this this morning to narrow down the repro here: https://github.com/NVIDIA/Fuser/issues/871#issuecomment-2462425423.

jjsjann123 (approved):
awesome!

Can we add a short section in https://github.com/NVIDIA/Fuser/wiki/Developer-guide#debug-nvfuser to describe how to use this :bow:

rdspring1 (approved):

jacobhinkle:
> Can we add a short section in https://github.com/NVIDIA/Fuser/wiki/Developer-guide#debug-nvfuser to describe how to use this 🙇

I will do that right now

PR #3371, opened 2024-11-07
Explicit check build flag

jjsjann123:
Adding a build option `--explicit-error-check` in setup.py along with related changes in CMakeLists.txt, so that the python option would set a compile option `NVFUSER_EXPLICIT_ERROR_CHECK` to the project.

This is to be used in our CI runs, where debug build is too slow but we still want things like dynamic_cast to catch errors.

Currently the change enables some runtime check, e.g. dynamic_cast, which was not checked previously in release build and have been hidden in CI.


naoyam:
Why not a little more descriptive name like `--explicit-error-check`?

jjsjann123:
!build

xwang233 (approved):

PR #3384, opened 2024-11-08
Remove debug prints

wujingyue:
Unit tests are supposed to clean themselves. If the purpose is to test toGraphvizDotGraph, I'd keep just that statement without storing the result to a file, the content of which isn't validated anyway.


naoyam (approved):
Sorry, this must be just something I forgot to remove after debugging.

wujingyue:
!test

PR #3388, opened 2024-11-08
Update input sizes for `test_many_segment_benchmark` to ensure kernel reuse

Priya2698:
Updating the sizes for this benchmark since we use different heuristics leading to kernel recompilation on A100 and H100.

Before: `sizes = [4, 8, 16, 32, 64, 128`
Screenshot 2024-11-08 at 2 39 35 PM

Current: `sizes = [5, 7, 9, 11]`
Screenshot 2024-11-08 at 2 39 11 PM

The `dynamic` measurement has lower standard deviation since we reuse kernels for all cases, and the average measurement is ~1.8ms as opposed to ~80ms with the earlier input sizes, with the maximum measurement of ~400ms

### DIFF
--- a/benchmarks/python/test_many_segments_host.py
+++ b/benchmarks/python/test_many_segments_host.py
@@ -30,11 +30,13 @@ def test_many_segment_benchmark(
     disable_validation: bool,
     disable_benchmarking: bool,
 ):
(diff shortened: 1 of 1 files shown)


jacobhinkle (approved):

Priya2698:
!build

PR #3391, opened 2024-11-09
Accept axis mapping when defining MmaOp

### LINKED ISSUE #3372: Enable MmaOp to receive unbroadcasted inputs
This is a proposal to enable MmaOp to receive inputs shaped like [M, K] and [N, K] instead of [M, 1, K] and [1, N, K].

This is an alternative to #3366.

# Motivation

Currently, MmaOp requires at least 3D inputs in which all of the dimensions "line up". That means that M dimensions should be Iteration in the A operand and Broadcast in the B operand. For example:
```
tv_a [ iS0{M}, bS1{1}, iS2{K} ]
tv_b [ bS3{1}, iS4{N}, iS5{K} ]
tv_c [ iS6{M}, iS7{N}, rS8{K} ] = fusedMultiplySum(tv_a, tv_b, axes={-1})
```
This lets us use the default exact domain mapping between operands and MmaOp output to determine the following groups {0, 3, 6}, {1, 4, 7}, {2, 5, 8}.

However, it means that if we are translating a Fusion that has `MatmulOp` or `LinearOp` to use `MmaOp`, we need to introduce `BroadcastOp` nodes, which interferes with the optimal gmem->smem->mma pipeline on Hopper.

# Proposed Approach

In order to avoid needing `BroadcastOp` when our segment inputs do not already have broadcasts, we need to handle cases like this:
```
(MatmulOp translation)
tv_a [ iS0{M}, iS1{K} ]
tv_b [ iS2{K}, iS3{N} ]
tv_c [ iS4{M}, iS5{N}, rS6{K} ] = fusedMultiplySum(tv_a, tv_b, ??)
```
We can no longer just specify some numbered axes to reduce, since the inputs do not have the same number of axes as the output. And even if we can specify it in the op, the IR node will need to hold some more information so that we can perform Exact and Broadcast mapping of IterDomains across the MmaOp.

We now need to specify the axes to reduce as well as the axis correspondences in the inputs. One possibility is to specify the position of the corresponding input axis for each output axis as an axis mapping:
```c++
tv_c = fusedMultiplySum(
    tv_a,
    tv_b,
    /*init=*/nullptr,
    /*axis_mapping=*/{/*a_axes=*/{0, -1, 1}, /*b_axes=*/{-1, 1, 0}});
```
I know this looks a little verbose but remember that we're not going to be creating these by hand very often as our main path will be to receive `MatmulOp` and `LinearOp` from Thunder and that interface will not change.

I propose to do the following:
* Add attributes to the `MmaOp` indicating the roles of dimensions in the `tv_a` and `tv_b` inputs.
* Add a special case in `PairwiseLogicalDomainMap` that will map the output domains to domains in the inputs using the map above. This is similar to what we do for `SdpaFwdOp` and `SdpaBwdOp` currently.
* Update `mma_utils::MatmulPattern::translateToMmaOp` to skip inserting broadcasts and use this interface instead.
* Update the Ampere matmul scheduler to not assume there is a broadcast M or N dimension in the `ab` and `bb` tensors.

Note that I also plan to keep the current interface for `fusedMultiplySum` available, so that we can use broadcasted inputs if we want to. The only caveat with keeping that old behavior around is that it might complicate the changes to the Ampere scheduler.

jacobhinkle:
This keeps the default interface of `fusedMultiplySum` but also adds an option to provide an `MmaOp::AxisMapping` object. This mapping defines, for each output dimension, which axis in each operand (if any) corresponds to that output dimension.

This PR does not alter the behavior of `mma_utils::MatmulPattern::translateToMmaOp` meaning we still have BroadcastOp in translations for Hopper matmuls, but that change should be relatively simpler.

Fixes #3372 

The included test only checks that dimensions are properly mapped in an MmaOp defined without broadcast axes. In followup PRs I plan to do the following:
1. Demonstrate scheduling a Hopper matmul with unbroadcasted inputs manually. This should surface any bugs in the lowering of the MmaOp instruction when broadcasts are absent.
2. Ensure that we don't depend on having broadcast dims in the Hopper matmul scheduler. For example, we will handle this case in `moveInnerBroadcastLeft` and we may also need to adjust the swizzling of the TMA smem load TensorView. At this point we will be able to automatically schedule an `MmaOp` without broadcasted inputs that has been manually defined using our automatic scheduler.
3. Add an option `MatmulPattern::translateToMmaOp(/*avoid_intermediates=*/true)` and enable that in the Hopper matmul scheduler. At this point it will be safe for us to accept `MatmulOp` and `LinearOp` in the Hopper matmul scheduler.


csrc/ir/utils.cpp
```diff
@@ -1265,158 +1265,3 @@ int64_t getOperationCount(Val* val) {
 }
 
 } // namespace nvfuser::ir_utils
-
-namespace nvfuser::MmaOpUtils {
```
jacobhinkle, line 1269: This was only needed for defining `MmaOp::mAxes()` and friends, but:
1. Those methods are never used so I removed them and
2. We can reconstruct that information easily using `mma->axisMapping()`.

jacobhinkle:
!test

jacobhinkle:
!test

jacobhinkle:
!test

jacobhinkle:
!test

jacobhinkle:
!test

jacobhinkle:
!test

jacobhinkle:
After this PR, one thing we can do is specify the dimension order of the output of the `MmaOp` independently from the inputs. When we translate `MatmulOp` and `LinearOp`, the output already has logical order `M, N` and we are free to place `K` wherever we want, so I'll place it last. I think this will let us avoid using `commitLeafToLogical` like is done here: https://github.com/NVIDIA/Fuser/blob/d34553f46e61341f8ac138630273dceb9f6cfbf8/tests/cpp/test_mma.cpp#L603
So in that case we can see how the `AxisMapping` is standing in for a root->logical reordering. Since there is one for each input operand this feels like another nice use case for read/write/compute domains as suggested by @zasdfgbnm for indexing ldmatrix.

csrc/ir/internal_nodes.h
```diff
@@ -1359,16 +1359,35 @@ class GroupedWelfordOp : public Expr {
 class NVF_API MmaOp : public Expr {
  public:
   using AxesData = std::vector<int64_t>;
+  // AxisMapping denotes the pairing of two input dimensions to produce an
+  // output dimension. It holds two vectors of integers indicating the
+  // corresponding position of each output axis in either the A or B input.
+  // Positions are absolute and refer to the noReductions logical domain. NOTE:
+  // -1 indicates that the axis does not exist, so Broadcast and Reduction
+  // dimensions should not have position -1.
```
zasdfgbnm, line 1367: Could you provide an example? For example, what does it mean if we have:
```C++
a_axes = {0, 2, 1, 3}
b_axes = {1, 0 , 3, 2}
```
does it mean `a.axis(0)` is mapped to `b.axis(1)`?

jacobhinkle: I'll put this in a comment but yes, this would mean we would map the following sets:
```
{a.axis(0), b.axis(1), out.axis(0)}
{a.axis(2), b.axis(0), out.axis(1)}
{a.axis(1), b.axis(3), out.axis(2)}
{a.axis(3), b.axis(2), out.axis(3)}
```

csrc/ir/internal_nodes.h
```diff
@@ -1359,16 +1359,59 @@ class GroupedWelfordOp : public Expr {
 class NVF_API MmaOp : public Expr {
  public:
   using AxesData = std::vector<int64_t>;
+  // AxisMapping denotes the pairing of two input dimensions to produce an
+  // output dimension. It holds two vectors of integers indicating the
+  // corresponding position of each output axis in either the A or B input.
+  // Positions are absolute and refer to the noReductions logical domain. NOTE:
```
zasdfgbnm, line 1365: `noReductions logical` does this mean we are ignoring K in the list?

tests/cpp/test_matmul_scheduler.cpp
```diff
@@ -3409,4 +3409,74 @@ TEST_F(MatmulSchedulerTest, HSH_NN) {
...
+  fusion.addOutput(tv2);
+
+  std::vector<mma_utils::MatmulPattern> patterns =
+      mma_utils::findMatmulPatterns(&fusion);
+
+  ASSERT_EQ(patterns.size(), 1);
+  const mma_utils::MatmulPattern& pattern = patterns.front();
+
+  IdModel id_model(&fusion);
+  const ValGraph& permissive_graph =
```
zasdfgbnm, line 3448: Let's use broadcast graph instead of permissive graph.

csrc/ir/internal_nodes.h
zasdfgbnm: From the implementation, it doesn't seem to be ignoring K.

csrc/ops/arith.cpp
```diff
@@ -2087,101 +2088,105 @@ TensorView* viewAsScalar(TensorView* inp) {
...
+    IterDomain* a_id = a_pos == -1 ? nullptr : a_logical[(size_t)a_pos];
+    IterDomain* b_id = b_pos == -1 ? nullptr : b_logical[(size_t)b_pos];
+
+    bool a_concrete = a_id == nullptr ? false : !a_id->isBroadcast();
+    bool b_concrete = b_id == nullptr ? false : !b_id->isBroadcast();
+    // NOTE: we can have !a_concrete && !b_concrete if there are broadcast batch
+    // dims
+
+    // Check for K dimensions
+    bool is_reduction = a_concrete && b_concrete && axes_set.count(i);
```
zasdfgbnm, line 2175: Is it possible to have `a_concrete && !b_concrete && axes_set.count(i)`? Should we error out on this case?

zasdfgbnm (approved):

csrc/ir/internal_nodes.h
jacobhinkle: Confusingly worded. It means reductions are ignored on the _inputs_ in case a reduction output is used in an MmaOp.

tests/cpp/test_matmul_scheduler.cpp
jacobhinkle: I am actually planning to do this, but it means I need to update mma_utils since permissive is still used there, so I will do that in a follow-up.

csrc/ops/arith.cpp
jacobhinkle: Great point. I will validate that any K dimensions are concrete in both inputs.

jacobhinkle:
!build

jacobhinkle:
!build

jacobhinkle:
!build

tests/cpp/test_matmul_scheduler.cpp
jacobhinkle: #3397

PR #3393, opened 2024-11-11
add knob controls unroll on top of vectorization in inner reduction

liqiangxl:
**Feature Added:** Knob for Unrolling on Top of Vectorization in Inner Reduction

Introduced a new knob to control unrolling in conjunction with vectorization for inner reduction loops.
**Note: The knob is currently not integrated into the heuristics and must be manually enabled for use.**
Added a test case to validate the functionality and ensure the knob's proper behavior.


liqiangxl:
!test

csrc/scheduler/reduction_heuristic.h
```diff
@@ -103,6 +103,9 @@ class ReductionParams : public HeuristicParams {
   // Unrolling/Vectorization factor for outer reduction factor
   int64_t unroll_factor_outer_reduction = 1;
 
+  // Extra unroll on top of vectorization
+  int64_t unroll_factor_top_of_vectorization = 1;
```
jjsjann123, line 107: nitpick: I think this belongs to be right beneath the declaration of `vectorize_inner_reduction`.

Should this also be renamed to identify that this is associated with `inner_reduction` in the name? in case we decide to expand similar support for `iter_dom`. i.e. `vectorize_iter_dom` and `unroll_factor_iter_dom`?

tests/cpp/test_reduction_pointwise.cpp
```diff
@@ -126,4 +126,36 @@ TEST_F(PointwiseFusedReductionTest, OuterReductionBroadcast) {
...
+  std::vector<c10::IValue> runtime_inputs({t0});
+
+  // Generate heuristics & enforce unroll on top of vectorization
+  SchedulerRuntimeInfo runtime_info(fusion.get(), runtime_inputs);
+  auto scheduler_instance =
+      SchedulerEntry::makeSchedulerInstance(SchedulerType::Reduction);
+  auto heuristic_params =
+      scheduler_instance->computeHeuristics(fusion.get(), runtime_info);
+  auto rparams = heuristic_params->as<ReductionParams>();
+  rparams->unroll_factor_top_of_vectorization = 2;
```
jjsjann123, line 150: should we validate that we are getting vectorization here?

csrc/scheduler/reduction_heuristic.h
liqiangxl: Good point. Moved to after `unroll_factor_inner_reduction (this is the vectorization factor)`.

For name, we don't need to change, The reason is it only applies to inner dim, only inner dim can be vectorized. It doesn't matter this inner dim is reduction or iteration.
For outer dim, we already have `unroll_factor_iter_dom`, and outer dim can't be vectorized.

tests/cpp/test_reduction_pointwise.cpp
liqiangxl: added `EXPECT_TRUE(rparams->vectorize_inner_reduction);`

liqiangxl:
!test

csrc/scheduler/reduction_heuristic.h
jjsjann123: Got'ya.

jjsjann123 (approved):

PR #3394, opened 2024-11-12
Add thunder benchmarks

Priya2698:
Adds `thunder` as an additional executor to the baseline benchmarks and the corresponding `thunder.jit` function.
The following benchmarks do not have `thunder` benchmark:
1. `instancenorm`: Unsupported operator in Thunder
2. `test_gelu_backward_reduction.py`: `.backward` call is not supported within Thunder definitions. @IvanYashchuk has suggested using explicit backward computation for this case.

Issue #2718

### DIFF
--- a/benchmarks/python/conftest.py
+++ b/benchmarks/python/conftest.py
@@ -2,7 +2,7 @@
 # All rights reserved.
 # SPDX-License-Identifier: BSD-3-Clause
 import pytest
-from .core import BENCHMARK_CONFIG, DEFAULT_EXECUTORS
+from .core import BENCHMARK_CONFIG
 from nvfuser.pytorch_utils import DEVICE_PROPERTIES
 
 
@@ -102,7 +102,7 @@ def pytest_configure(config):
 def pytest_collection_modifyitems(session, config, items):
     """
     The baseline benchmarks use `executor` parameter with
-    values ["eager", "torchcompile", "thunder"] that are optionally
+    values ["eager", "torchcompile", "thunder", "thunder-torchcompile"] that are optionally
     run using `--benchmark-{executor}` flag. They are skipped by
     default.
     """
@@ -122,7 +122,7 @@ def pytest_collection_modifyitems(session, config, items):
 
     executors_to_skip = []
 
-    for executor in DEFAULT_EXECUTORS:
+    for executor in executors:
         if not config.getoption(f"--benchmark-{executor}"):
             executors_to_skip.append(executor)
 
--- a/benchmarks/python/conftest.py
+++ b/benchmarks/python/conftest.py
@@ -2,9 +2,9 @@
 # All rights reserved.
 # SPDX-License-Identifier: BSD-3-Clause
 import pytest
-from .core import BENCHMARK_CONFIG
+from .core import BENCHMARK_CONFIG, DEFAULT_EXECUTORS
 from nvfuser.pytorch_utils import DEVICE_PROPERTIES
-from .global_params import DEFAULT_EXECUTORS
+
 
 def pytest_addoption(parser):
     parser.addoption(
--- a/benchmarks/python/conftest.py
+++ b/benchmarks/python/conftest.py
@@ -4,7 +4,7 @@
 import pytest
 from .core import BENCHMARK_CONFIG
 from nvfuser.pytorch_utils import DEVICE_PROPERTIES
-
+from .global_params import DEFAULT_EXECUTORS
 
 def pytest_addoption(parser):
     parser.addoption(
@@ -122,7 +122,7 @@ def pytest_collection_modifyitems(session, config, items):
 
     executors_to_skip = []
 
-    for executor in executors:
+    for executor in DEFAULT_EXECUTORS:
         if not config.getoption(f"--benchmark-{executor}"):
             executors_to_skip.append(executor)
 
--- a/benchmarks/python/core.py
+++ b/benchmarks/python/core.py
@@ -22,7 +22,8 @@ BENCHMARK_CONFIG = {"rounds": 10, "warmup_rounds": 1, "num_inputs": None}
 L2_CACHE_SIZE = DEVICE_PROPERTIES["gpu_l2_bytes"]
 PEAK_BANDWIDTH_GBPS = DEVICE_PROPERTIES["gpu_peak_bandwidth_gbps"]
 
-DEFAULT_EXECUTORS = ['eager', 'torchcompile', 'thunder']
+DEFAULT_EXECUTORS = ["eager", "torchcompile", "thunder"]
+
 
 def clear_l2_cache() -> None:
     """
@@ -44,7 +45,8 @@ def clear_dynamo_cache() -> None:
 
 
 # Backward function for torch baseline benchmarks.
-def unary_bwd_torch(inputs: List):  # [output, grad_out]
+# The first two inputs are expected to be out and grad_out. The remaining are inputs of the forward pass used to clear grad between subsequent runs to avoid grad accumulation. See setup() in run_benchmark().
+def unary_bwd_torch(inputs: List):  # [output, grad_out, fwd_inputs]
     inputs[0].backward(inputs[1], retain_graph=True)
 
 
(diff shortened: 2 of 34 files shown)


Priya2698:
!build

liqiangxl:
!test --pybench

benchmarks/python/test_gelu_bwd.py
```diff
@@ -109,6 +111,9 @@ def gelu_fwd():
     fwd_fn = {
         "eager": gelu_fwd,
         "torchcompile": torch.compile(gelu_fwd),
+        "thunder": thunder.jit(
+            gelu_fwd, executors=[nvfuserex], nv_enable_bookend=False
+        ),
     }
     outputs = fwd_fn[executor]()
```
jacobhinkle, line 118: Maybe we should abstract this away with a couple new functions in `core.py`. For example instead of listing the executors directly in we could use
```python
@pytest.mark.parametrize('executor', default_executors)
```
Then down here we could use
```python
fwd_fn = with_executor(executor, gelu_fwd)
```
That way we can enable/disable executors in a centralized location.

Priya2698:
!test --pybench-full --dev

Priya2698:
!test --pybench-full --dev

Priya2698:
!test --pybench-full --dev

benchmarks/python/core.py
```diff
@@ -325,6 +327,9 @@ def run_benchmark(
     def setup():
         clear_l2_cache()
         if device == "cuda":
+            for inp in inputs:
+                if isinstance(inp, torch.Tensor):
+                    inp.grad = None
```
jjsjann123, line 332: Thank you for this one. But this is only the cases where input requires gradient. Are we also clearing gradient on parameters?

Priya2698: Do you mean, for instance, weights in layernorm?

Priya2698: Ahh you're right.
Even with clearing the gradients of weights, bias and inputs in backward pass, I think it is still missing some variables/internal states that need to be reset.
The simplest way is to only run 1 round for backward, but I feel that may be noisy, so have trying to make it work for multiple runs.

jjsjann123: Got'ya. No worries. I'm not totally clear what's the protocol in thunder on ownership of parameters, I think it's supposed to be a functional compilation.
So we can still expect that `with_executor` has the chance to extract parameters from nn.Module if it's given for benchmark and we should be able to identify parameter that needs zero_grad. just like an optimizer would.

Priya2698:
!test --pybench-full --dev

Priya2698:
!test --pybench-full --dev

Priya2698:
We still see some performance differences between Thunder and nvfuser manual definitions, for some operators, which may be due to slight differences in the fusion definitions generated. I will look into this operator-wise.

I have compared the measured timings against nsys timings for some operators to verify the accuracy of the benchmark infra. My recommendation is to check in these changes adding Thunder benchmarks, while we investigate the performance gap.

benchmarks/python/test_softmax_bwd.py
```diff
@@ -115,6 +121,6 @@ def test_softmax_bwd_baseline_benchmark(
     run_benchmark(
         benchmark,
         unary_bwd_torch,
-        [outputs, grads],
+        [outputs, grads, *fwd_inputs],
```
jjsjann123, line 124: sneaky! I see `unary_bwd_torch` discards further inputs, but you pass them here to clear the grad?
We can use a comment in unary_bwd_torch on why we are not asserting num of inputs.

jjsjann123 (commented):
I think there're still quite some questions on what's the next step and how to automate things like IOBytes computation and stuff.

But for this PR, things look pretty mechanical, so I'm good with merging it as-is then iron out the remaining issues.

My only concern is, would the disruption rendering our benchmark not reliable until we fix all these issues? and does it cause issues on folks looking at the benchmarks?

Priya2698:
> My only concern is, would the disruption rendering our benchmark not reliable until we fix all these issues? and does it cause issues on folks looking at the benchmarks?

The `thunder.jit` performance numbers match against nsys. They also compare correctly for operators like `scale-bias-relu`, `softmax`, `silu-mul` against the manual nvfuser definitions. But for others like normalization, the performance is different. So my next step is to actually compare the fusion definitions, and account how each difference in the 2 fusion definitions correspond to the performance gap (for instance, the manual fusion definitions have some missing downcasts for intermediate ops, or the ordering of ops etc).

So by performance gap, I imply the latter case. The measurements of the benchmarks should be accurate.
This PR is actually fixing one of the measurement issues existent in the backward benchmarks due to grad accumulation.

Priya2698:
> I think there're still quite some questions on what's the next step and how to automate things like IOBytes computation and stuff.

Agreed! Let me open an issue tracking the IOByte case. I will create an issue for the differences between `thunder.jit` and nvfuser manual definition performance shortly (I want to do some initial analysis).

Priya2698:
!build

Priya2698:
!test --pybench-full --dev

Priya2698:
@jacobhinkle @liqiangxl pinging for review.

We continue to see performance difference between Thunder-nvfuser and nvfuser. The thunder-nvfuser executor will not run by default so we can continue investigating this in parallel. This PR has the fix for bwd grad accumulation issue, so we can merge this. If the consensus is to hold off on adding Thunder-nvfuser benchmarks altogether, I can remove it from the executors list for benchmarks as well. That change is minimal.

@jjsjann123 Can you give me the numbers you expect for RoPE bwd after the grad accumulation issue is resolved -- I can verify the numbers with this PR?

benchmarks/python/core.py
```diff
@@ -23,6 +22,8 @@
 L2_CACHE_SIZE = DEVICE_PROPERTIES["gpu_l2_bytes"]
 PEAK_BANDWIDTH_GBPS = DEVICE_PROPERTIES["gpu_peak_bandwidth_gbps"]
 
+DEFAULT_EXECUTORS = ["eager", "torchcompile", "thunder"]
```
Priya2698, line 25: Maybe this should be named differently since these are not run in nightly, but for most benchmarks, these are the set of executors we execute weekly. We also have `thunder-torchcompile` for RoPE.
Maybe `BASELINE_EXECUTORS` is better, although Thunder is not really a baseline.

Priya2698:
!build

benchmarks/python/conftest.py
```diff
@@ -102,7 +102,7 @@ def pytest_configure(config):
 def pytest_collection_modifyitems(session, config, items):
     """
     The baseline benchmarks use `executor` parameter with
-    values ["eager", "torchcompile", "thunder"] that are optionally
+    values ["eager", "torchcompile", "thunder", "thunder-torchcompile"] that are optionally
```
jjsjann123, line 105: Thanks for adding this in comment. :bow:

benchmarks/python/core.py
jjsjann123: speaking of `Thunder is not really a baseline`. I think @naoyam made a comment about renaming `thunder` to `thunder-nvfuser`, which I think makes sense, since we are fixing the executor to only allow nvfuser in those benchmarks.

I'll try to start a PR renaming that. Does that sound reasonable to you?
cc'ing @xwang233 sorry we might need to also update our nightly/weekly scripts.

jjsjann123 (approved):
Changes looks good. Thanks for updating this.

Let me go verify the kernel time in rope... :bow:

benchmarks/python/core.py
xwang233: not a problem

Priya2698: Yes, I recall that. I can do that in a separate PR so it is strictly a string change. May be easier to review.

Priya2698:
> Changes looks good. Thanks for updating this.
> 
> Let me go verify the kernel time in rope... 🙇

Thanks for the review. I'll merge this once the rope times are verified.

jjsjann123:
Not your problem, but I realized that there's some issue with the rope implementation. Let me patch that on the side.

Priya2698:
> Not your problem, but I realized that there's some issue with the rope implementation. Let me patch that on the side.

Okay, merging this PR then.

PR #3397, opened 2024-11-12
Replace PERMISSIVE graph with BROADCAST for matmul scheduler

jacobhinkle:
The dimension role mapping code in the matmul scheduler was written before the PERMISSIVE graph was rewritten in terms of an intermediate graph called BROADCAST. This PR does not introduce any functional changes but just switches to using this new graph.


jacobhinkle:
!test

zasdfgbnm (approved):

PR #3398, opened 2024-11-12
Change tests to call runPass directly.

wujingyue:
This
1. makes the tests closer to reality because inferenceAllocationOrder is not the only thing that the pass does, and 
2. simplifies the tests so they don't have to specify sources and destinations.


wujingyue:
!test

jjsjann123 (approved):

PR #3401, opened 2024-11-13
Prevent marking inplace update target as candidate for alias analysis

jjsjann123:
We shouldn't inplace-update target as candidate for alias analysis.

Since the update requires a real kernel, marking it as candidate for alias could lead to unnecessary segmentation that cannot be handled by trivially evaluating with meta operations on an existing Tensor.


jjsjann123:
!test

jjsjann123:
The issue was found from here: https://github.com/Lightning-AI/lightning-thunder/issues/1173

jjsjann123:
!test

csrc/alias_analysis.cpp
```diff
@@ -248,6 +248,11 @@ void AliasFinder::handle(const LoadStoreOp* set) {
   }
   TensorView* out = set->out()->as<TensorView>();
 
+  // The set operation is an inplace update and shouldn't be marked as alias op.
+  if (out->fusion()->getOutputAlias(out).type == AllocationType::ReuseBuffer) {
```
wujingyue, line 252: We may want to check this in AliasAnalysisResult::finalize. Keeping this alias pair in `alias_to_source_` is useful if the result of the `set` is also used by other aliasing ops. I believe as long as we don't materialize it in `alias_to_root_` we should be fine. 

https://github.com/NVIDIA/Fuser/blob/e7d551e908fd691e791bc34a966c80bdda34f36b/csrc/alias_analysis.cpp#L452

tests/cpp/test_alias.cpp
```diff
@@ -1545,4 +1545,38 @@ TEST_F(AliasTest, Issue2664) {
...
+      at::randn(in_shape, at::dtype(at::kFloat).device(at::kCUDA, 0));
+  at::Tensor original_tensor = in_tensor.clone();
+  std::vector<at::Tensor> out_tensors =
+      executor_cache.runFusionWithInputs({in_tensor});
+  ASSERT_EQ(out_tensors.size(), 1);
+
+  // Verify inplace update
+  EXPECT_TRUE(out_tensors[0].equal(in_tensor));
+  // Verify no segmentation
+  EXPECT_FALSE(executor_cache.getMostRecentKernelRuntime()->isSegmented())
```
wujingyue, line 1574: https://testing.googleblog.com/2013/08/testing-on-toilet-test-behavior-not.html

I think what we really care about is that the input is correctly updated. Whether the fusion is segmented is an implementation detail that's not observable to the user. Wdyt?

wujingyue, line 1574: Question: does nvFuser support in-place updates (aka input/output alias) in multi-segment fusions? I vaguely remember seeing some segmentation code dealing with that, but I'm not sure.

jjsjann123: I felt like that I needed something to justify the PR.

If we just care about correctness, there's nothing wrong with ToT code as-is. But segmentation could lead to regression (compilation time as well as execution time), which is arguably an observable thing to user. :)

If you think the test is a bit too explicit. I'm more than happy to remove it. I hope to keep it in case we refactor this pass later and might hit the same issue again.

jjsjann123: Yes we do.

@Priya2698 actually added a preseg pass to force segmentation to avoid RW racing, that's probably what you are referring to.

csrc/alias_analysis.cpp
```diff
@@ -248,6 +248,11 @@ void AliasFinder::handle(const LoadStoreOp* set) {
   }
   TensorView* out = set->out()->as<TensorView>();
 
+  // The set operation is an inplace update and shouldn't be marked as alias op.
```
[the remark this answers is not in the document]
jjsjann123: For now I think it's only applicable to `set`.

csrc/alias_analysis.cpp
jjsjann123: Good point. Let me take another look at this later.

tests/cpp/test_alias.cpp
wujingyue: > If we just care about correctness, there's nothing wrong with ToT code as-is. But segmentation could lead to regression (compilation time as well as execution time), which is arguably an observable thing to user. :)

Got it! Then it's certainly fine to keep the segmentation check here.

wujingyue: I was referring to https://github.com/NVIDIA/Fuser/blob/7b9271640b3b918b5cb65d9c68ea2c17d14319d2/csrc/fusion_segmenter.cpp#L355-L356. I was curious when the complete fusion has to be segmented (for acceptance by a scheduler) how nvFuser sets up I/O aliases of sub-fusions.

csrc/alias_analysis.cpp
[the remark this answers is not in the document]
wujingyue: Why not `view`? For example, 

```
out = reshape(in, [6], [2, 3]);
out = add(out, 1)
out = reshape(out, [2, 3], [6]);
aliasOutputToInput(out, in);
```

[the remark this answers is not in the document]
jjsjann123: That's a great example.

It's an implementation thing I think. Yes. in theory we **could** support it. Since it's just a view and we should be able to infer the new stride for `out` when it's an alias of `in`.

But I don't think we do that today. https://github.com/NVIDIA/Fuser/blob/7b9271640b3b918b5cb65d9c68ea2c17d14319d2/csrc/runtime/allocations.cpp#L260-L275
We are blindly using `in` at the location of `out` in the allocation, which requires `out` and `in` to have the same size/stride.

I'll try your example and make sure that an error is thrown for that.

tests/cpp/test_alias.cpp
jjsjann123: I see.

The logic in segmenter is naive. Actually the whole codegen is pretty naive about I/O tensors. i.e. we pretend that we don't know about it at all and we just treat aliases as if they are new buffer.

The only thing the segmenter does is to add the alias-in as an input to the segment that's producing the alias-out. Just so that at runtime, we'll be able to use it for output.

csrc/alias_analysis.cpp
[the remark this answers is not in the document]
wujingyue: > which requires out and in to have the same size/stride.

`in` and `out` in my example do have the same size/stride. There's a *roundtrip* of reshapes.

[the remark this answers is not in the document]
jjsjann123: oh how blind was I :face_with_head_bandage: 

So I think your earlier comment makes sense, if we do this in `AliasAnalysisResult::finalize` and not materialize `alias` when it has a re-use buffer. We are not restricted by the op type.

Thanks for pointing it out. :saluting_face:

tests/cpp/test_alias.cpp
```diff
@@ -1538,4 +1538,86 @@ TEST_F(AliasTest, Issue2664) {
...
+  TensorView* in = makeContigConcreteTensor(in_shape);
+  fusion->addInput(in);
+  TensorView* r_in = reshape(in, in_shape, out_shape);
+  TensorView* out = add(r_in, IrBuilder::create<Val>(3.141));
+  fusion->addOutput(out);
+  // this is an inplace update and shouldn't be segmented into its own kernel by
+  // alias analysis
+  TensorView* r_out = reshape(out, out_shape, in_shape);
+  TensorView* update_input = set(r_out);
+  fusion->aliasOutputToInput(update_input, in, AllocationType::ReuseBuffer);
```
jjsjann123, line 1597: added the case with reshape. Thanks a lot for the example.

tests/cpp/test_alias.cpp
```diff
@@ -1538,4 +1538,86 @@ TEST_F(AliasTest, Issue2664) {
...
+  at::Tensor original_tensor = in_tensor.clone();
+  std::vector<at::Tensor> out_tensors =
+      executor_cache.runFusionWithInputs({in_tensor});
+  ASSERT_EQ(out_tensors.size(), 1);
+
+  // Verify inplace update
+  EXPECT_TRUE(out_tensors[0].equal(in_tensor));
+  // Verify no segmentation
+  EXPECT_FALSE(executor_cache.getMostRecentKernelRuntime()->isSegmented())
+      << "segmentation is not supposed to happen";
```
wujingyue, line 1569:
```suggestion
      ;
```
This error message doesn't add any additional information to what's already inside EXPECT_FALSE. So I'd remove it. The same goes to the next test case.

wujingyue (approved):

wujingyue:
Nit: can you write a better title for the PR than "fixing alias analysis stuff"?

jjsjann123:
!test

PR #3402, opened 2024-11-13
Skip SetUp when the communicator isn't available.

wujingyue:
Although `MultiDeviceTest::SetUp` early returns when the communicator isn't available, it doesn't automagically make the *caller* exit early.


wujingyue:
!test

naoyam (approved):
LGTM

Issue #2853, opened 2024-08-27
Adding CPU scalar tensor to scalar value produces CUDA tensor.


Priya2698:
The following repro produces a CUDA tensor instead of a CPU tensor.

```
def test_cpu_add():
    inputs = [torch.tensor(2.0, device="cpu", dtype=torch.float)]

    def fusion_func(fd: FusionDefinition):
        t0 = fd.from_pytorch(inputs[0])
        s0 = fd.define_scalar(3.0)
        t1 = fd.ops.add(t0, s0)
        fd.add_output(t1)

    with FusionDefinition() as fd:
        fusion_func(fd)
    nvf_out = fd.execute(inputs)
    torch.testing.assert_close(nvf_out[0].cpu(), inputs[0] + 3.0)
```

CC: @jjsjann123

Priya2698:
The issue itself will not be resolved since nvFuser codegen cannot return CPU tensors.
We will look into adding and error message for such conditions if possible.

PR #3403, opened 2024-11-13
Reject fusions with CPU scalar outputs in `KernelExecutor`

Priya2698:
Issue #2853 surfaced a case where adding CPU scalar tensor to scalar value generates a CUDA tensor. This is because nvFuser cannot generate CPU tensors.
This PR adds a check to identify any CPU scalar outputs in `KernelExecutor::supported` to surface such unexpected cases.


csrc/runtime/executor.cpp
```diff
@@ -208,6 +208,15 @@ void KernelExecutor::compile(
     return;
   }
 
+  const bool has_any_cuda_tensor = std::any_of(
```
Priya2698, line 211: Doing this check at the segment level instead of `runFusionWithInputs`, since the entire fusion may have a CUDA tensor input but individual segments may not. See the test case added in the PR for example.

csrc/runtime/executor.cpp
```diff
@@ -208,6 +208,15 @@ void KernelExecutor::compile(
...
+      fusion->inputs().begin(), fusion->inputs().end(), [](Val* inp) {
+        return inp->isA<TensorView>() && !inp->as<TensorView>()->isCpuScalar();
+      });
+
+  NVF_CHECK(
+      has_any_cuda_tensor,
+      "Expected atleast one input to be on DeviceType::CUDA.");
```
jacobhinkle, line 218:
```suggestion
      "Expected at least one input to be on DeviceType::CUDA.");
```
Also, since this is an internal error and not a misspecification of the overall Fusion, maybe we should use `NVF_ERROR` instead.

csrc/runtime/executor.cpp
```diff
@@ -208,6 +208,15 @@ void KernelExecutor::compile(
     return;
   }
 
+  const bool has_any_cuda_tensor = std::any_of(
+      fusion->inputs().begin(), fusion->inputs().end(), [](Val* inp) {
+        return inp->isA<TensorView>() && !inp->as<TensorView>()->isCpuScalar();
```
jacobhinkle, line 213: Unfortunately I think we'll get some false positives here when we use factory functions. What about this fusion?
```python
  tv0 = fd.ops.full([5], dtype=DataType.Float)
  tv1 = fd.ops.neg(tv0)
  fd.add_output(tv1)
```
This fusion has no inputs at all but produces a CUDA tensor. Instead, we should probably propagate `isCpuScalar()` through expressions, and check that there is at least one non-cpuscalar tensor in the fusion in this check.

Priya2698: You're right.
There seem to some other conditions too, let me revise the check.

What do you mean by `propagate isCpuScalar() through expressions`?

jacobhinkle: I mean if we have two cpu scalar tensors and we multiply them, for example, then we get another cpu scalar tensor. Same if we multiply by a non-tensor scalar. But multiplying a CPU scalar tensor by a GPU tensor produces a GPU tensor. I think if we propagate these rules properly than we should only be compiling kernels whenever there is a GPU tensor output.

Priya2698:
!test

csrc/runtime/executor.cpp
```diff
@@ -244,6 +244,50 @@ std::vector<at::Tensor> ExprEvalExecutor::run(
   return outputs;
 }
 
+bool KernelExecutor::supported(Fusion* fusion) {
+  FUSER_PERF_SCOPE("KernelExecutor::supported");
+
+  auto has_only_scalar_inputs = [](Expr* expr) -> bool {
```
jacobhinkle, line 250: Since this function might eventually do stuff other than verify scalar tensors, it might make sense to put most of this implementation into a static function  like `isPureCPUScalarFusion(fusion)`.

jacobhinkle (commented):
Since we could call `setCpuTensor` at any time either during or after the Fusion has been defined, does it make sense to propagate that setting in a preseg pass?

csrc/runtime/executor.cpp
Priya2698: Yes, I agree.
I am still playing around with some other configurations of inputs to make sure these are the only rules.
I'll finalize the PR after that.

jacobhinkle:
> What scenario are you considering which would require a preseg pass propagation?

All three cases above would potentially be handled properly if we propagate this property in the definition ops. For example
```
tv0->setCpuScalar()
tv1 = exp(tv0)
// what does tv1->isCpuScalar() return here?
```
Since we trust that this flag is set properly on fusion inputs, instead of relying on the ops to manage this at definition (and for this flag to survive all our rewriting passes), we could just propagate it at preseg like so:
```c++
void propagateCpuScalars(Fusion* fusion) {
  for (Expr* expr : StmtSort::getExprs(fusion)) {
    bool has_cpu_scalar_input = false;
    bool has_cuda_input = false;
    for (Val* inp : expr->inputs()) {
      if (auto* tv : dynamic_cast<TensorView*>(inp)) {
        if (tv->isCpuScalar()) {
          has_cpu_scalar_input = true;
        } else {
          has_cuda_input = true;
        }
      }
    }
    if (has_cpu_scalar_input && !has_cuda_input) {
      // Mark all TV outputs as CPU scalars
      for (Val* outp : expr->outputs()) {
        if (auto* tv : dynamic_cast<TensorView*>(inp)) {
          tv->setCpuScalar();
        }
      }
    }
  }
}
```

Priya2698:
> > What scenario are you considering which would require a preseg pass propagation?
> 
> All three cases above would potentially be handled properly if we propagate this property in the definition ops. For example
> 
> ```
> tv0->setCpuScalar()
> tv1 = exp(tv0)
> // what does tv1->isCpuScalar() return here?
> ```
> 
> Since we trust that this flag is set properly on fusion inputs, instead of relying on the ops to manage this at definition (and for this flag to survive all our rewriting passes), we could just propagate it at preseg like so:
> 
> ```c++
> void propagateCpuScalars(Fusion* fusion) {
>   for (Expr* expr : StmtSort::getExprs(fusion)) {
>     bool has_cpu_scalar_input = false;
>     bool has_cuda_input = false;
>     for (Val* inp : expr->inputs()) {
>       if (auto* tv : dynamic_cast<TensorView*>(inp)) {
>         if (tv->isCpuScalar()) {
>           has_cpu_scalar_input = true;
>         } else {
>           has_cuda_input = true;
>         }
>       }
>     }
>     if (has_cpu_scalar_input && !has_cuda_input) {
>       // Mark all TV outputs as CPU scalars
>       for (Val* outp : expr->outputs()) {
>         if (auto* tv : dynamic_cast<TensorView*>(inp)) {
>           tv->setCpuScalar();
>         }
>       }
>     }
>   }
> }
> ```

I see. This is a good alternative.
So instead of going through the fusion expressions, like the current approach, we can query this flag for the `fusion->outputs()`. It should work for `SdpaFwdOp` as well, since the relevant outputs are set as CPU scalars by the op. I do not have a preference on which implementation to go with -- from the overhead consideration, both approach should be fine.

>instead of relying on the ops to manage this at definition (and for this flag to survive all our rewriting passes)

If the passes rewrite this flag accidentally, setting the flags for outputs will also be error-prone.
What do you mean by ` relying on the ops to manage this at definition` -- is it that in the example above, the flag for `tv1` will not be set?

Priya2698:
!test

Priya2698:
> > What scenario are you considering which would require a preseg pass propagation?
> 
> All three cases above would potentially be handled properly if we propagate this property in the definition ops. For example
> 
> ```
> tv0->setCpuScalar()
> tv1 = exp(tv0)
> // what does tv1->isCpuScalar() return here?
> ```
> 
> Since we trust that this flag is set properly on fusion inputs, instead of relying on the ops to manage this at definition (and for this flag to survive all our rewriting passes), we could just propagate it at preseg like so:
> 
> ```c++
> void propagateCpuScalars(Fusion* fusion) {
>   for (Expr* expr : StmtSort::getExprs(fusion)) {
>     bool has_cpu_scalar_input = false;
>     bool has_cuda_input = false;
>     for (Val* inp : expr->inputs()) {
>       if (auto* tv : dynamic_cast<TensorView*>(inp)) {
>         if (tv->isCpuScalar()) {
>           has_cpu_scalar_input = true;
>         } else {
>           has_cuda_input = true;
>         }
>       }
>     }
>     if (has_cpu_scalar_input && !has_cuda_input) {
>       // Mark all TV outputs as CPU scalars
>       for (Val* outp : expr->outputs()) {
>         if (auto* tv : dynamic_cast<TensorView*>(inp)) {
>           tv->setCpuScalar();
>         }
>       }
>     }
>   }
> }
> ```

There is an issue in propagating this flag to tensor outputs.
Consider
```
        def fusion_func(fd: FusionDefinition):
            s0 = fd.from_pytorch(inputs[1])
            t1 = fd.ops.add(t0, s0)
            fd.add_output(t1)
```
Here, `t0` will have an implicit broadcast.
```
T2_l_float[ bS1{1} ]
   = broadcast( T0_g_float[ ] )
```
The output of this broadcast will be inferred as CPU scalar, but that is incorrect. As a whole, this fusion produces a CUDA tensor output so is a valid fusion.

Priya2698:
!test

csrc/preseg_passes/propagate_cpu_scalars.cpp
```diff
@@ -0,0 +1,57 @@
...
+    bool has_cuda_input = false;
+
+    for (Val* inp : expr->inputs()) {
+      if (auto* inp_tv = dynamic_cast<TensorView*>(inp)) {
+        if (inp_tv->isCpuScalar()) {
+          has_cpu_scalar_input = true;
+        } else {
+          has_cuda_input = true;
+          // Return early -- found atleast one CUDA input
+          break;
```
jacobhinkle: Good idea to break early here.

jacobhinkle (approved):
LGTM

jacobhinkle:
> The output of this broadcast will be inferred as CPU scalar, but that will raise an error since we do not allow CPU non-scalar tensors. As a whole, this fusion produces a CUDA tensor output so is a valid fusion.

I see, so the failure occurs when `tv->setCpuScalar()` is called in the preseg pass in that case.  Maybe the condition should be `TensorDomain::noBroadcasts(getLogicalDomain()).empty()` instead of `nDims() == 0`?

tests/cpp/test_evaluator.cpp
```diff
@@ -814,4 +814,27 @@ TEST_F(ExprEvalTest, TensorMetadataPrecomputedValues) {
...
+  auto out = add(tv0, s0);
+
+  auto options = at::TensorOptions().dtype(at::kFloat).device(at::kCPU);
+  auto t0 = at::scalar_tensor({3.0}, options);
+
+  ExpressionEvaluator evaluator;
+  evaluator.bind(tv0, t0);
+  at::Tensor out_tensor = evaluator.evaluate(out).as<at::Tensor>();
+  EXPECT_TRUE(out_tensor.is_cpu());
+  debug() << out_tensor << std::endl;
```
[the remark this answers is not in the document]
Priya2698: I removed this example.

Priya2698:
> > The output of this broadcast will be inferred as CPU scalar, but that will raise an error since we do not allow CPU non-scalar tensors. As a whole, this fusion produces a CUDA tensor output so is a valid fusion.
> 
> I see, so the failure occurs when `tv->setCpuScalar()` is called in the preseg pass in that case. Maybe the condition should be `TensorDomain::noBroadcasts(getLogicalDomain()).empty()` instead of `nDims() == 0`?

Yes, that should work most likely.
If we choose to support returning CPU outputs through expression evaluator, we would need the analysis to be within the preseg pass and  the protocol to be modified, i.e. `TensorDomain::noBroadcasts(getLogicalDomain()).empty()` instead of `nDims() == 0`.
I am not sure of the original motivation to allow CPU scalar tensors alone, though and if we can support CPU inputs more generally (currently, there isn't a strong use-case AFAIK)

I'll can this PR as-is and open an issue for the above, wdyt?

Priya2698:
!test

Priya2698:
!test

jacobhinkle:
> I can merge this PR as-is and open an issue for the above, wdyt?

Yes, I think if this issue is not currently causing any breakage then we should do that to get you unblocked.

PR #3410, opened 2024-11-15
Preserve additional IDs when mutating TensorDomains in OptOutMutator

### LINKED ISSUE #3409: Additional IDs are lost when mutating a TensorDomain
When we create a Broadcast IterDomain using `tv->broadcast(pos)`, it does inserts the new ID into the loop domain of `tv`, as well as into another domain called `tv->additionalIDs()`. This additional IDs domain is used to find traversal paths in order to define all ID expressions when building ID graphs. We must ensure that these domains are preserved whenever we clone or mutate a TensorDomain, otherwise this can cause issues when building an ID graph with the new Fusion.

We currently do not preserve the additional IDs when mutating, as seen in this repro:
```c++
TEST_F(NVFuserTest, OptOutMutatorAdditionalBroadcastID) {
  FusionGuard fg(fusion);

  auto tv0 = makeSymbolicTensor(1);
  fusion->addInput(tv0);

  auto tv1 = exp(tv0);

  fusion->addOutput(tv1);

  // We add a broadcast domain bS2{1}. This adds the new Broadcast ID to tv1->domain()->additionalIDs()
  // logical: [ iS1{i0} ]
  // loop: [ iS1{i0}, bS2{1} ]
  // additional IDs: [ bS2{1} ]
  tv1->broadcast(1);
  EXPECT_FALSE(tv1->domain()->additionalIDs().empty());

  // After this split we have
  // logical: [ iS1{i0} ]
  // loop: [ iS1{i0}, bS3{1}, bS4{2} ]
  // additional IDs: [ bS2{1} ]
  tv1->split(1, 2);
  EXPECT_FALSE(tv1->domain()->additionalIDs().empty());

  // Now register a mutation that will alter some IDs in the domain
  OptOutMutator mut;
  mut.registerMutation(tv1->axis(0)->extent(), IrBuilder::create<Val>(DataType::Index));
  TensorDomain* old_tensor_domain = tv1->domain();
  auto all_stmts = StmtSort::getStmts(
      fusion,
      /*traverse_members*/ true,
      /*traverse_attributes*/ true,
      /*traverse_siblings*/ true);
  for (auto stmt : all_stmts) {
    mut.dispatchMutate(stmt);
  }
  EXPECT_TRUE(tv1->domain() != old_tensor_domain) << "Mutation did not change the TensorDomain";

  EXPECT_FALSE(tv1->domain()->additionalIDs().empty())<< "Mutation did not preserve additional IDs";
}
```
I noticed this in a very similar case when working on #3406 .

jacobhinkle:
Fixes #3409 

This PR does the following:
1. Creates `tests/cpp/test_mutator.cpp` and moves the `OptOutMutator` tests from `test_dynamic_transform.cpp` into that file.
2. Adds a new test that repros #3409 
3. Adds an additional argument `const std::vector<IterDomain*>& additional_ids` to one of the `TensorDomain` constructors.
4. Uses that constructor in `OptOutMutator::mutate(TensorDomain*)` to pass a mutated `additional_ids`.

A bit off-topic maybe: implementing this got me thinking more about traversals and whether or not `StmtSort::getStmts(fusion, /*traverse_members=*/true)` will still guarantee full traversals now that the loop domain is not necessarily the set of leaf IterDomains in the TensorDomain:
https://github.com/NVIDIA/Fuser/blob/f1a7e23604dc8639bec488e472909eca8bdbbb31/csrc/iter_visitor.cpp#L71-L74
I think in another PR we might want to modify this code to ensure that we traverse all of `domain->allIDs()` in topological order.


jacobhinkle:
!test

tests/cpp/test_mutator.cpp
```diff
@@ -0,0 +1,149 @@
...
+  KernelExecutor ke;
+  ke.compile(fusion);
+
+  auto outputs = ke.run({3L});
+
+  testValidate(fusion, outputs, {3L}, __LINE__, __FILE__);
+}
+
+// Test that additional IDs are preserved when mutating a TensorView
+TEST_F(NVFuserTest, OptOutMutatorAdditionalBroadcastID) {
```
jacobhinkle, line 105: This is the new test. The other two tests above were copied verbatim.

jacobhinkle:
!build

naoyam (approved):
Thanks for the fix!

Issue #3279, opened 2024-10-25
Tracking perf optimization of `HopperMatmulTest.HSH_NT_128BSwizzle` for problem size `(M=2048, N=2048, K=8192)`, CTA tile size `(128, 256)`


zasdfgbnm:
The CTA tile size `(128, 256)` is a size that can relatively easily achieve high math throughput. The problem size is carefully selected as one full wave.

zasdfgbnm:
Initial perf as measured in https://github.com/NVIDIA/Fuser/pull/3281:
```
 Time (%)  Total Time (ns)  Instances  Avg (ns)  Med (ns)  Min (ns)  Max (ns)  StdDev (ns)                                                  Name

 --------  ---------------  ---------  --------  --------  --------  --------  -----------  ----------------------------------------------------------------------------------------------------
     43.2           205150          1  205150.0  205150.0    205150    205150          0.0  <unnamed>::nvfuser_none_f0_c0_r0_g0(<unnamed>::Tensor<<unnamed>::__half, (int)3, (int)3>, <unnamed>…
     18.5            87550          1   87550.0   87550.0     87550     87550          0.0  nvjet_hsh_256x128_64x4_1x2_h_bz_coopA_NTT
```

nvFuser/cuBLAS = `42.7%`

zasdfgbnm:
There is a perf regression after the fix of elect-sync:
https://github.com/NVIDIA/Fuser/pull/3295

Perf:
```C++
 Time (%)  Total Time (ns)  Instances  Avg (ns)  Med (ns)  Min (ns)  Max (ns)  StdDev (ns)
 Name
 --------  ---------------  ---------  --------  --------  --------  --------  -----------  ----------------------------------------------------------------------------------------------------
     47.8           247326          1  247326.0  247326.0    247326    247326          0.0  <unnamed>::nvfuser_none_f0_c0_r0_g0(<unnamed>::Tensor<<unnamed>::__half, (int)3, (int)3>, <unnamed>…
     17.0            88191          1   88191.0   88191.0     88191     88191          0.0  nvjet_hsh_256x128_64x4_1x2_h_bz_coopA_NTT
```
Perf nvFuser/cuBLAS: `35.6%`

zasdfgbnm:
Perf:
```markdown
 Time (%)  Total Time (ns)  Instances  Avg (ns)  Med (ns)  Min (ns)  Max (ns)  StdDev (ns)                                                  Name

[...]

PR #3411, opened 2024-11-15
Add support for stmatrix in the unit test HopperMatmulTest/HSH_NT_128BSwizzle

protonu:
This demonstrates the use of stmatrix in a multi-tile hopper matmul.


tests/cpp/test_matmul.cpp
```diff
@@ -3706,7 +3708,26 @@ TEST_F(HopperMatmulTest, HSH_NT_128BSwizzle) {
 
   auto tv3 = castOp(DataType::Half, tv2);
 
-  fusion.addOutput(tv3);
+  TensorView* tv4 = nullptr;
+  if (use_smem_epilogue) {
```
jacobhinkle, line 3712: Nit: you can unconditionally addOutput here then if using smem epilogue use cacheBefore to create the set op

tests/cpp/test_matmul.cpp
```diff
@@ -3706,7 +3708,26 @@ TEST_F(HopperMatmulTest, HSH_NT_128BSwizzle) {
...
+    tv4 = set(tv3);
+
+    tv3->setMemoryType(MemoryType::Shared);
+    tv4->definition()->as<LoadStoreOp>()->setOpType(
+        LoadStoreOpType::CpAsyncBulkTensorTile);
+
+    fusion.addOutput(tv4);
+
+    // We'll use stmatrix.x4 to store from reg to shared memory
+    fusion.manage("st_matrix_m_tile", 16);
```
jacobhinkle, line 3723: Just to be clear, does this mean we must use a macro such that getN(macro) is a multiple of 16?

tests/cpp/test_matmul.cpp
```diff
@@ -3775,12 +3800,28 @@ TEST_F(HopperMatmulTest, HSH_NT_128BSwizzle) {
...
+          tv->getLoopDomain());
+      tv->setLoopDomain(s.as<IterDomain*>());
+    }
+  } else {
     auto s = mma_utils::MmaSwizzler::scheduleMmaOutputAllocation(
-        tv->getLoopDomain());
-    tv->setLoopDomain(s.as<IterDomain*>());
+        tv3c->getLoopDomain());
+    tv3c->setLoopDomain(s.as<IterDomain*>());
+    tv3c->setAllocationDomain(s.as<IterDomain*>(), true);
```
jacobhinkle, line 3813: Is this necessary? The loop domain was just set so shouldnt that be the default allocation domain?

tests/cpp/test_matmul.cpp
```diff
@@ -3775,12 +3800,28 @@ TEST_F(HopperMatmulTest, HSH_NT_128BSwizzle) {
...
+
+    // We need to split and merge dimensions which have
+    // been marked parallel/vectorized. So we revert the
+    // inner-most dimension back to serial.
+    tv3->axis(-1)->parallelize(ParallelType::Serial);
```
jacobhinkle, line 3818: How does tv3 look before this line? I expected it to just be block tiled and the outer dims parallelized BIDy and BiDy by the propagation from tv2.

tests/cpp/test_matmul.cpp
protonu: yes -
For the m-dimension this shouldn't be a problem since it's always 64 for Hopper.
For the n-dimension, if it's not a multiple of 16, it has to a multiple of 8, and we'll have to use stmatrix.x2 where st_matrix_n_tile will be 8.

tests/cpp/test_matmul.cpp
protonu: I am not sure what I was doing last night that I saw a vectorized dim there! 
There isn't one - will remove the serial type.

tests/cpp/test_matmul.cpp
protonu: This is require - I don't understand this well, maybe @zasdfgbnm can tell us the exact reason.

csrc/device_lower/pass/index.cpp
zasdfgbnm, line 1: nit: revert unrelated change

tests/cpp/test_matmul.cpp
zasdfgbnm: In the legacy indexer, the default allocation is not the loop domain, it is the `getMaybeAllocationDomain`, which is the logical domain. So the behavior isn't really well-defined in the transition of indexers... My suggestion is, if commenting this out this test still pass, then remove this. Otherwise, keep it.

zasdfgbnm (approved):
Could you also post some perf data here with smem epilogue on? The command I use is posted here: https://github.com/NVIDIA/Fuser/issues/3279

tests/cpp/test_matmul.cpp
protonu: Without it, the test doesn't pass

protonu:
Observations:

1. On running `getBankConflictInfo`, I see both cases (mma with and without a stmatrix) have zero bank conflicts. The codes I used was:
```
  KernelExecutor ke;
  auto launch_constraints = LaunchParams();
  ke.compile(
      &fusion,
      {inputs.first, inputs.second},
      launch_constraints,
      matmul_cparams);

  auto bank_conflict_info =
      getBankConflictInfo(ke.kernel(), launch_constraints);

  if (bank_conflict_info.empty()) {
    debug() << "===== No bank confliction =====" << std::endl;
  } else {
    debug() << "======= Bank confliction =======" << std::endl;
    for (auto info : bank_conflict_info) {
      debug() << "Expr: " << info.first->toString() << std::endl;
      auto conflict = info.second;
      if (conflict.first > 1) {
        debug() << "input conflict: " << conflict.first << " way, ";
      }
      if (conflict.second > 1) {
        debug() << "output conflict: " << conflict.second << " way";
      }
      debug() << std::endl;
    }
    debug() << "================================" << std::endl;
  }
  fusion.printKernel();
```

protonu:
!build

